PR #3108, opened 2026-07-25
fix(notifications): suppress toast for the actively-viewed focused session

### LINKED ISSUE #3107: Suppress in-app notifications for the actively-viewed, focused session
## Summary

In-app notifications fire for a session's `needs_input` transition even when that session is already open and focused in the foreground. If you're actively watching a session, you still get a toast telling you it needs input — which you can already see.

## Screenshot

Image

## Steps to Reproduce

1. Open a session in the terminal and keep it focused/visible.
2. Trigger a `needs_input` transition in that same session.
3. Observe the in-app toast fires anyway.

## Expected

No toast (and arguably no bell-badge increment) for a `needs_input` transition on the session you're currently viewing in a focused window.

## Actual

Toast fires unconditionally on every `needs_input` transition, regardless of which session is active or whether the window is focused.

## Root Cause

Frontend-only. The daemon behaves correctly and its emission must stay unconditional.

- **Backend (correct, do not change):** `backend/internal/lifecycle/manager.go:259-281` emits a `NotificationNeedsInput` intent on every `needs_input` transition. The daemon is headless and serves multiple clients at once (desktop + mobile over LAN + multiple windows), so it can't know which pane the user is looking at. Suppression belongs client-side.
- **Frontend (missing guard):** `frontend/src/renderer/lib/notifications.ts:209-221` — the SSE `notification_created` handler calls `aoBridge.notifications.show()` for every new notification with no check against the active session or window visibility.

## Proposed Fix

Gate `aoBridge.notifications.show()` in `notifications.ts` so it's skipped when **both**:

1. `notification.sessionId === activeSessionId` (session is the one open in the terminal), **and**
2. `document.visibilityState === "visible"` (window is focused).

The visibility pattern is already used in `telemetry.ts:116` and `useDaemonStatus.ts:70`.

**Wiring:** `createNotificationsTransport(queryClient)` (`NotificationCenter.tsx:83`) currently only receives `queryClient` and has no access to active-session state. Thread in a getter:

```ts
export function createNotificationsTransport(
  queryClient,
  getActiveSessionId: () => string | undefined,
)
```

Active session is available as `selection.activeSessionId` (from router `params.sessionId`, used in `Sidebar.tsx:593` and `ShellTopbar.tsx:45`).

## Open Question

Toast suppression is the priority. Whether to also suppress the bell-badge increment (and auto-`markNotificationRead`) for an actively-viewed session is a design call worth confirming before implementing.

## Scope

- One file + its test: `notifications.ts`, `notifications.test.ts`
- Plus the call-site signature change in `NotificationCenter.tsx`
- No backend changes, no API contract changes, no DB migration.

## Labels / Priority

- **Labels:** `bug`, `comp/desktop`
- **Priority:** P2 — annoying UX, not a blocker (workaround: dismiss the toast). Bumps to P1 if it fires in a tight loop.

ikeshavvarshney:
Fixes #3107

## Problem

In-app notifications fire for a session's `needs_input` transition even when that session is already open and focused in the foreground. If you are actively watching a session, you still get a toast telling you it needs input — which you can already see on screen.

## Why the fix belongs in the renderer

The daemon's emission is correct and stays unchanged. `backend/internal/lifecycle/manager.go` emits a `NotificationNeedsInput` intent on every `needs_input` transition, and it has to: the daemon is headless and serves multiple clients at once (desktop, mobile over the LAN listener, multiple windows). It cannot know which pane any given user is looking at, so it must emit unconditionally and let each client decide what to surface.

Suppression is therefore a client-side concern. No backend changes, no API contract changes, no migration.

## Change

**`frontend/src/renderer/lib/notifications.ts`**

`createNotificationsTransport` takes a second, optional `getActiveSessionId: () => string | undefined` parameter. In the `notification_created` SSE handler, the toast is skipped when **both** conditions hold:

1. `notification.sessionId === getActiveSessionId()` — the notification is for the session currently open in the terminal, and
2. `document.visibilityState === "visible"` — the window is actually in the foreground.

Either condition alone is not enough. A background window showing the active session still gets a toast (you are not looking at it), and a visible window on a different session still gets a toast (that session is not the one you can see).

The parameter defaults to `() => undefined`, so any existing call site keeps the previous unconditional behaviour rather than silently suppressing.

**`frontend/src/renderer/components/NotificationCenter.tsx`**

`NotificationRuntime` reads the router's `sessionId` via `useParams({ strict: false })` — the same source `Sidebar.tsx` and `ShellTopbar.tsx` already use for active-session state — and passes it down as a getter backed by a ref.

The ref matters. The SSE transport is established once inside a `useEffect` keyed on `queryClient`; passing the session id by value would either go stale as the user navigates, or force the effect to re-run and tear down/reopen the `EventSource` on every session switch. A ref updated on each render gives the long-lived connection a live read of the current session with no reconnect.

## Deliberately out of scope: the bell badge

The issue raises an open question about whether the bell-badge increment (and an auto-`markNotificationRead`) should also be suppressed for an actively-viewed session. This PR does **not** touch that.

`mergeUnreadNotification` and `mergeRecentNotification` still run for every event, so the badge count and the notification list stay accurate and consistent with what the daemon has recorded. Only the transient toast is suppressed. Auto-marking-read on view is a behavioural design call with its own edge cases (what counts as "seen", what happens across multiple windows), and is better made deliberately in a follow-up than smuggled in here.

## Verification

**Unit tests** — three cases added to `notifications.test.ts`, covering both halves of the guard:

| Scenario | Expected |
| --- | --- |
| Active session + window visible | no toast |
| Different session + window visible | toast fires |
| Active session + window hidden | toast fires |

Each also asserts the notification still lands in the unread cache, pinning the "badge keeps working" behaviour.

**Live run against a real daemon** — built `ao daemon` from source, ran it on an isolated `AO_DATA_DIR` on port 3911, pointed the renderer dev server at it, and drove a real Chromium instance with Playwright on a `/sessions/:sessionId` route. The real (unmocked) transport module was exercised in-page against the real SSE stream URL:

```
stream url: http://127.0.0.1:3911/api/v1/notifications/stream
toast after active+visible : []                            (expected [])
toast after other session  : [ntf_other]                   (expected [ntf_other])
toast after active+hidden  : [ntf_other, ntf_hidden]       (expected [ntf_other, ntf_hidden])
bell/unread cache ids      : [ntf_active, ntf_hidden, ntf_other]  (all three retained)
```

No console errors on either the root or the session route.

**Checks**

- `npx vitest run` over `notifications.test.ts`, `NotificationCenter.test.tsx`, `shell-new-session-shortcut.test.tsx` — 35 passed
- `npm run typecheck` — clean

## Note on formatting

`npx prettier --check` reports style issues on the touched files, but it reports the same on untouched files (for example `src/renderer/lib/format-time.ts`). There is no Prettier config committed in the repo and the existing code is tab-indented, so running `--write` here would have produced formatting churn unrelated to the fix. The new code matches the surrounding style. Happy to reformat if the project wants a Prettier pass.

## Scope

- `frontend/src/renderer/lib/notifications.ts`
- `frontend/src/renderer/lib/notifications.test.ts`
- `frontend/src/renderer/components/NotificationCenter.tsx`

3 files, +50 / -5. No backend, API, or schema changes.


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (changes requested):
Requesting changes for two notification-suppression bugs:

1. `notifications.ts:218` treats `document.visibilityState === "visible"` as focused. In Electron on Windows/Linux, an unfocused or covered window can remain visible, so AO can suppress the toast while the user is in another app. Require actual focus (`document.hasFocus()` or `BrowserWindow.isFocused()`).

2. The guard applies to every notification type. #3107 only asks to suppress `needs_input`, but `ready_to_merge`, `pr_merged`, and `pr_closed_unmerged` toasts are also hidden for active sessions. Gate on `notification.type === "needs_input"` and add regression coverage.

ikeshavvarshney:
Thanks, both were real. Fixed in `71e57a7`.

1. **Visibility vs. focus** — You're right that `visibilityState` was the wrong signal. The guard now requires:

   ```ts
   document.visibilityState === "visible" && document.hasFocus()
   ```

   I reproduced the case you described in a real Chromium window rather than trusting a stub. With a second tab holding focus, the session page reports:

   ```json
   {
     "visibility": "visible",
     "focused": false
   }
   ```

   That's exactly the state that used to swallow the toast. It now fires.

2. **Over-broad guard** — Also correct, and the blast radius was every other type in the enum (`ready_to_merge`, `pr_merged`, `pr_closed_unmerged`). Those report outcomes that aren't visible in the terminal pane, so they always deserve a toast. Suppression is now gated on:

   ```ts
   notification.type === "needs_input"
   ```

While tightening the comparison, I found one more edge: `notification.sessionId === activeSessionId` matched when both were `undefined`, so a notification with no session would have been suppressed on a route without one. It now rejects an empty `sessionId` explicitly.

The logic moved into `suppressToastForFocusedSession()` so the reasoning—why visibility alone is insufficient, and why only `needs_input` qualifies—lives next to the code instead of in the PR description.

On the regression coverage: rather than just adding tests, I verified they actually catch these bugs by reverting the guard to the pre-review logic. Exactly four failed—the visible-but-unfocused case and all three PR-outcome types—and nothing else. So the new cases are pinned to the two defects you found, not passing incidentally.

I also re-ran end-to-end against a real daemon (isolated `AO_DATA_DIR`, real SSE stream) driving the actual module in a browser: **11/11 passing**, including confirmation that the unread cache still retains every notification whether or not the toast was suppressed.

The bell badge is deliberately still untouched—happy to take that on here as well if you'd prefer it in scope.

One unrelated note: `src/main/supervisor-link.test.ts` fails locally for me with `listen EACCES` on its temporary `.sock` files. I confirmed it fails identically on a clean tree, so it's environmental on Windows and not caused by this change.

illegalcall (commented):
Re-reviewed at `71e57a7` against merge-base `0127410b2` in a clean worktree.

## Both prior points are genuinely fixed

1. **`visibilityState` ≠ focused** — the guard now requires `document.visibilityState === "visible" && document.hasFocus()`.
2. **Over-broad guard** — suppression returns early unless `notification.type === "needs_input"`, so `ready_to_merge` / `pr_merged` / `pr_closed_unmerged` still toast.

The self-caught `undefined === undefined` match on a session-less notification is a good catch too.

I checked the regression tests actually pin those defects rather than passing incidentally: reverting `suppressToastForFocusedSession` to the pre-review logic fails exactly four tests — the visible-but-unfocused case and all three PR-outcome types — and nothing else.

**Verification in the worktree:** `tsc --noEmit` clean, full frontend suite `86 files / 1087 tests` passing. CI is green (`test`, `renderer-smoke`, `scan`).

**Base staleness:** `main` has moved 6 commits, including project session tabs (#3138). I checked for a semantic conflict — #3138 adds a `tabOwner` search param, but the route's `sessionId` is still the *displayed* session, and main's only edit to `NotificationCenter.tsx` is the bell icon class near line 172. No textual or logical conflict; a rebase isn't required.

## One remaining gap, same class as the bug just fixed

Being on the session route doesn't mean the agent pane is on screen.

`CenterPane.tsx:213` renders a **single** `TerminalPane` driven by `terminalTarget`. When the user selects a shell-terminal tab (`SessionView.tsx:68`, `kind: "shell"`) or the reviewer terminal, the agent's terminal is swapped out entirely — not shown alongside. In that state the route param still matches, the window is visible and focused, so a `needs_input` toast is suppressed while the prompt is nowhere on screen. `SessionPaneTab` (`CenterPane.tsx:151`) carries no needs-input indicator, so nothing else surfaces it either.

It's narrow — it needs an open shell or reviewer tab in the same session — but it's the same "visible ≠ actually seen" failure mode the review just corrected. Two ways out: thread `terminalTarget.kind === "worker"` into the guard, or put an attention dot on the session tab. Reasonable to defer alongside the bell-badge follow-up, but worth making that a conscious deferral rather than leaving it implicit.

## Non-blocking notes

- `activeSessionIdRef.current = params.sessionId` writes a ref during render. Technically outside strict concurrent-render purity, but the value is render-derived with no tearing consequence, and the effect-based alternative would lag the SSE handler. Fine as-is.
- When the browser panel's `WebContentsView` holds focus, the renderer's `document.hasFocus()` goes false, so a toast fires even though the user is looking at the session. That errs toward showing a toast, which is the safe direction.

Ready to land from my side once the stale changes-request is dismissed.

illegalcall (approved):

illegalcall:
@ikeshavvarshney Please make sure you test and confirm once you verify on nightly.

PR #3114, opened 2026-07-26
feat: box session inspector sections with settings surfaces

Annieeeee11:
## Summary
- Wrap Summary/Reviews inspector sections in settings-row cards (`bg-settings-row`, `rounded-settings-row`) so the right rail matches the boxed inspector refs
- Align inspector typography and borders with the settings theme (`text-settings-muted` / `text-settings-label`, settings input border/bg tokens)
- Collapse the dual `surface` vs plain `Section` variants into one boxed layout (prop kept for call-site compatibility)

## Test plan
- [ ] Open a session and confirm Summary sections (Overview, Completion, etc.) render as boxed cards
- [ ] Switch to Reviews and confirm PR cards / empty states use the settings surface colors
- [ ] Resize the inspector rail narrow and confirm section headers/kv rows still layout cleanly
- [ ] Confirm Terminate / Terminate-on-merge controls still work inside the Completion box

## Before
image

## After
image

### DIFF
--- a/frontend/src/renderer/components/SessionInspector.tsx
+++ b/frontend/src/renderer/components/SessionInspector.tsx
@@ -99,14 +99,14 @@ const inspectorShellClass = "@container/inspector flex h-full min-h-0 flex-col o
 
 const inspectorBodyClass = "min-h-0 flex-1 overflow-y-auto p-3 pb-4 @max-[300px]/inspector:px-2.5";
 
-const inspectorEmptyClass = "text-xs text-muted-foreground leading-normal";
+const inspectorEmptyClass = "text-xs text-settings-muted leading-normal";
 
 const kvRowClass =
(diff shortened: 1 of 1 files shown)


codebanditssss (approved):
lgtm. clean visual refactor and nicely stacked on top of the merged reviews work, so the boxing now covers the reviews sections too. verified every settings token resolves (settings-muted/label, bg-settings-row, rounded-settings-row from --radius-settings-row, tracking-settings-section) and CI is green.

two small non-blocking nits for whenever:
- CompletionControls has a stray extra tab on the canTerminateNow <div> so it's mis-indented vs its siblings. harmless in JSX, just untidy.
- PRSummaryCard/ResumeAgentControl use the arbitrary-value form bg-(--color-bg-settings-input) / border-(--color-border-settings-input) right next to proper utilities like bg-settings-row. those input tokens only have css vars, no utility, so it's a workaround, but a bg-settings-input/border-settings-input utility would keep it consistent.

approving.

PR #3117, opened 2026-07-26
fix(codex): recover sessions stuck in active state

### LINKED ISSUE #3115: bug(codex): session stuck 'working' for days — final turn's stop hook never fired, no idle fallback demotes stale active
## Bug
A codex worker session has shown **working / `activity_state=active`** continuously for 2+ days while the agent sat at a chat prompt doing nothing. The last activity transition ever recorded is an `active` at the start of its final turn; the `stop` hook that should have demoted it to `idle` at turn end never arrived, and nothing else in the system corrects a stale `active`.

**Source:** maintainer local report (desktop app, macOS) | **Date:** 2026-07-25 | **Analyzed against:** `dec77c61b` (local) / `12591e1c6` (origin/main) | **Daemon:** app-owned, :3001
**Confidence:** High on evidence and adapter mechanism; Medium on why this specific `stop` was lost
**Priority:** medium-high — "working" is a trust-critical signal; a permanently-working session masks a blocked fleet (and inverts [#2814](https://github.com/AgentWrapper/agent-orchestrator/issues/2814)'s kimi idle-while-working)

## Evidence (session `agent-orchestrator-39`, codex)
`change_log` shows healthy hook-driven cycles, then a final transition that never closes:

```
10:19:40 active   10:19:53 idle
10:20:15 active   10:20:29 idle
10:21:15 active   10:21:23 idle
10:21:45 active   10:22:05 idle
10:23:28 active   ← LAST EVENT EVER (2026-07-23); still active on 2026-07-25
```

The agent's final turn ended normally from the user's perspective — it printed a clarifying question and returned to the composer prompt (visible in tmux scrollback). AO's own idle detection even fired worker-idle notifications for this session around the same window, yet the persisted activity state stayed `active`, and `ao session get` reports `status: working` two days later.

## Root Cause
`backend/internal/adapters/agent/codex/activity.go` — codex activity is driven by exactly three hook events: `user-prompt-submit`→active, `permission-request`→waiting_input, `stop`→idle. The adapter's own doc comment notes codex has **no SessionEnd/Notification equivalent** and no pre/post-tool-use hooks; runtime exit falls back to the reaper. Consequences:

1. If a single `stop` hook is lost (hook process failed, PATH issue, codex skipped firing it on that turn shape), the session is `active` **forever** — there is no reconciliation that says "no hook events for N minutes + composer prompt visible → demote to idle".
2. The reaper only handles runtime *exit*, not stale *state*; the process is alive, so it never intervenes.
3. Downstream effects compound: `waiting_input`/idle-gated flows behave wrongly for a phantom-active session (e.g. nudge gating in `backend/internal/lifecycle/reactions.go` skips on NeedsInput but not on stale-active; board/status pills show working).

## Fix
Suggested directions:
1. **Staleness reconciliation:** if `activity_state=active` but no hook event has arrived for N minutes, cross-check a cheap liveness/inbox signal (terminal output delta, composer-prompt detection à la `activitystate`) and demote to idle — the generic safety net that would fix this class for every hook-based adapter, not just codex.
2. **Codex-specific:** investigate which turn shapes end without firing the `stop` managed hook (this repro: turn ended with a plain assistant question, no tool use) and harden the hook installation.
3. Emit a diagnostic when a session's last hook event is `active` and older than some threshold — this state is currently invisible until a human squints at the timestamps.

## Impact
- Board/sidebar/status pills show a permanently-working session; the human cannot tell blocked-waiting-on-me from working (this repro sat 2 days awaiting user input the UI never surfaced).
- Automated flows gated on activity state (nudges, idle notifications, completion checks) run against fiction.

## Related
- [#2814](https://github.com/AgentWrapper/agent-orchestrator/issues/2814): kimi inverse (idle while working) — same "hook coverage gaps corrupt activity truth" family; fix PR [#2873](https://github.com/AgentWrapper/agent-orchestrator/pull/2873) in flight
- [#2047](https://github.com/AgentWrapper/agent-orchestrator/issues/2047): claude-code activity misreporting
- [#2802](https://github.com/AgentWrapper/agent-orchestrator/issues/2802): IsAlive pane-exists ≠ agent-alive (liveness-vs-inbox split)
- [#2944](https://github.com/AgentWrapper/agent-orchestrator/issues/2944) / [#2345](https://github.com/AgentWrapper/agent-orchestrator/issues/2345): status presentation locked to PR state

Hardik180704:
Fixes #3115.

## What changed

Codex can occasionally miss its final `stop` hook, which leaves AO showing a session as working even after Codex has returned to the composer.

This adds a small activity observer that:

- checks only stale active sessions whose adapter supports terminal activity detection;
- recognizes Codex's idle composer and footer markers;
- keeps sessions active while the `esc to interrupt` working marker is visible;
- fences terminal observations with the runtime launch and session update timestamp so stale output cannot overwrite a newer prompt;
- uses the normal lifecycle transition, including the durable worker-idle event.

Other agent adapters are left unchanged.

## Verification

- `npm run lint`
- `cd backend && go test -race ./...`
- frontend typecheck and E2E typecheck
- 1,084 frontend Vitest tests
- 20 renderer-smoke Playwright tests
- real tmux + SQLite regression test covering both idle recovery and a genuine long-running Codex turn

### DIFF
--- a/backend/internal/adapters/agent/codex/codex.go
+++ b/backend/internal/adapters/agent/codex/codex.go
@@ -64,6 +64,7 @@ var _ adapters.Adapter = (*Plugin)(nil)
 var _ ports.Agent = (*Plugin)(nil)
 var _ ports.ActiveTurnSteerer = (*Plugin)(nil)
 var _ ports.AgentAuthChecker = (*Plugin)(nil)
+var _ ports.TerminalActivityDetector = (*Plugin)(nil)
 
 // Manifest returns the adapter's static self-description.
 func (p *Plugin) Manifest() adapters.Manifest {
--- a/backend/internal/adapters/agent/codex/terminal_activity.go
+++ b/backend/internal/adapters/agent/codex/terminal_activity.go
@@ -1,0 +1,49 @@
+package codex
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+var codexTerminalEscape = regexp.MustCompile(`\x1b(?:\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]|\][^\x07]*(?:\x07|\x1b\\))`)
+
+// DetectTerminalActivity reports idle only when Codex's composer and footer are visible.
+func (p *Plugin) DetectTerminalActivity(output string) (domain.ActivityState, bool) {
+	lines := terminalLines(output)
+	if len(lines) < 2 {
+		return "", false
+	}
+	start := len(lines) - 12
+	if start < 0 {
+		start = 0
+	}
+	for _, line := range lines[start:] {
+		if strings.Contains(strings.ToLower(line), "esc to interrupt") {
+			return "", false
+		}
+	}
+	for i := len(lines) - 2; i >= start; i-- {
+		if !strings.HasPrefix(strings.TrimSpace(lines[i]), "›") {
+			continue
+		}
+		if strings.Contains(lines[i+1], " · ") {
+			return domain.ActivityIdle, true
+		}
+	}
+	return "", false
+}
+
+func terminalLines(output string) []string {
+	plain := codexTerminalEscape.ReplaceAllString(strings.ReplaceAll(output, "\r", "\n"), "")
+	raw := strings.Split(plain, "\n")
+	lines := raw[:0]
+	for _, line := range raw {
+		line = strings.TrimSpace(line)
+		if line != "" {
+			lines = append(lines, line)
+		}
+	}
+	return lines
+}
--- a/backend/internal/adapters/agent/codex/terminal_activity_test.go
+++ b/backend/internal/adapters/agent/codex/terminal_activity_test.go
@@ -1,0 +1,43 @@
+package codex
+
+import (
+	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+func TestDetectTerminalActivity(t *testing.T) {
+	tests := []struct {
+		name   string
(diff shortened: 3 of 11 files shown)


codebanditssss (approved):
lgtm, this is careful, well-fenced work. went through the observer, the codex marker detection, and the lifecycle fencing.

what stood out as solid:
- the stale-read-vs-fresh-prompt race is double-fenced: LaunchID rejects an older process generation, and the new ExpectedUpdatedAt snapshot-CAS drops the signal if the record moved or is no longer active. the user-prompt-submit -> UpdatedAt bump closes the other direction so a new prompt invalidates an in-flight idle signal. both directions are directly tested.
- detection fails safe: requires composer + footer AND absence of the "esc to interrupt" working marker, so it won't false-idle a live turn. lines[i+1] is bounds-safe and the raw[:0] in-place filter is correct.
- zero blast radius: opt-in via the TerminalActivityDetector interface, codex-only, guarded on staleness + runtime handle. observer is wired and drained on shutdown.
- strong tests including a real tmux+sqlite integration case and a genuine long-running turn.

a few non-blocking notes for whenever:
- the codex markers (the › glyph, "esc to interrupt" / " · ", the 12-line window) are hardcoded, so a future codex UI change would silently stop recovery rather than false-idle. it's validated against real codex today, but a log/metric when detection never fires on stale sessions, or a comment pinning the tested UI version, would help catch drift.
- the user-prompt-submit -> UpdatedAt bump now applies to all adapters, not just codex. it's needed for the fence and tested, just worth confirming it doesn't cause UpdatedAt churn elsewhere.
- trivial: the 12 scan-window could be a named const alongside DefaultOutputLines.

approving. nice work.

PR #3122, opened 2026-07-26
feat(landing): swap in the new aoagents-landing app

codebanditssss:
Swaps frontend/src/landing to the new aoagents-landing app.

Landing app sourced from https://github.com/ronishrohan/aoagents — credit to that repo.


Annieeeee11 (approved):
igtm

PR #3128, opened 2026-07-26
fix(desktop): recover when orchestrator agent is missing

### LINKED ISSUE #3125: bug(desktop): Spawn Orchestrator has no recovery when project has no agent configured
## Summary

A valid project registered without an `orchestrator.agent` appears normally in the desktop app and shows an enabled **Spawn Orchestrator** button. Clicking it fails with the raw daemon error:

```text
spawn: session: agent harness required: configure project orchestrator.agent or pass --harness (AGENT_REQUIRED)
```

The desktop action has no harness input, agent picker, or link to project settings, so the user cannot recover from the surface where the error occurs.

Verified on current `main` at `5a587191e` using the Electron development app and its isolated `~/.ao/dev` state.

## Reproduction

1. Create a Git repository with an initial commit.
2. Register it through the supported CLI without agent flags:

   ```bash
   ao project add --path /path/to/repo --name "No Agent Test"
   ```

3. Open that project in the desktop app.
4. Click **Spawn Orchestrator**.

## Actual behavior

- The button is enabled and looks ready to use.
- The spawn request is sent without a harness.
- The board renders the technical `AGENT_REQUIRED` error below the button.
- Retrying produces the same result.
- The error tells the user to pass `--harness`, although the desktop button provides no way to do that.

## Expected behavior

When the project has no configured orchestrator agent, the desktop should provide a recovery path before sending the spawn request. For example:

- open an agent picker and spawn with the selected installed/authorized harness; or
- open Project Settings focused on the missing orchestrator-agent field; or
- disable the button with a clear action to configure an agent.

Configured projects should keep the current one-click spawn behavior.

## Why this state is valid/reachable

`ao project add` intentionally makes `--worker-agent` and `--orchestrator-agent` optional (`backend/internal/cli/project.go`). This can also affect older or externally registered projects whose config lacks role defaults.

The board path currently calls `spawnOrchestrator(projectId, "board")` without checking `workspace.orchestratorAgent` (`frontend/src/renderer/components/SessionsBoard.tsx`). The failure is only rendered as text after the daemon rejects the request.

## Suggested scope

Frontend-only recovery should be enough; the daemon is correctly rejecting an underspecified spawn.

Add regression coverage for:

- missing `orchestratorAgent`: clicking **Spawn Orchestrator** does not POST an invalid request and opens/directs to an agent-configuration surface;
- configured `orchestratorAgent`: existing one-click spawn still works;
- the same missing-config behavior from every spawn affordance (board empty state, topbar, and sidebar shortcut).

Hardik180704:
## What changed

- redirect unconfigured projects to Project Settings before spawning an orchestrator
- cover the board, top bar, sidebar, and command palette launchers
- keep the existing one-click flow unchanged when an agent is configured or an orchestrator is already running
- add regression coverage for every affected launcher

## Why

Projects added through the CLI can exist without an orchestrator agent. The desktop still showed an enabled spawn action and sent a request that could only fail with `AGENT_REQUIRED`. This gives the user a recovery path instead.

## Checks

- frontend typecheck
- E2E typecheck
- full Vitest suite: 1,081 passed
- renderer smoke: 20 passed
- `git diff --check`

Closes #3125

### DIFF
--- a/frontend/src/renderer/__tests__/integration/board-empty-states.test.tsx
+++ b/frontend/src/renderer/__tests__/integration/board-empty-states.test.tsx
@@ -36,7 +36,7 @@ import { SessionsBoard } from "../../components/SessionsBoard";
 import { ShellProvider, type ShellContextValue } from "../../lib/shell-context";
 import { useUiStore } from "../../stores/ui-store";
 
-type Project = { id: string; name: string; path: string };
+type Project = { id: string; name: string; path: string; orchestratorAgent?: string };
 type Session = Record<string, unknown>;
 
 function respondWith(projects: Project[], sessions: Session[]) {
@@ -47,7 +47,12 @@ function respondWith(projects: Project[], sessions: Session[]) {
 	});
 }
 
-const project: Project = { id: "proj-1", name: "my-app", path: "/repo/my-app" };
+const project: Project = {
+	id: "proj-1",
+	name: "my-app",
+	path: "/repo/my-app",
+	orchestratorAgent: "claude-code",
+};
 
 const workerSession: Session = {
 	id: "sess-1",
@@ -186,6 +191,22 @@ describe("project board with no sessions", () => {
 		await userEvent.click(spawnButton);
 
 		expect(await screen.findByText(/branch is already checked out/)).toBeInTheDocument();
+	});
+
+	it("opens project settings instead of spawning when no orchestrator agent is configured", async () => {
+		const unconfiguredProject = { ...project, orchestratorAgent: undefined };
+		respondWith([unconfiguredProject], []);
+		renderBoard(<SessionsBoard projectId="proj-1" />);
+
+		await screen.findByText("No worker sessions yet");
+		const [spawnButton] = screen.getAllByRole("button", { name: "Spawn Orchestrator" });
+		await userEvent.click(spawnButton);
+
+		expect(navigateMock).toHaveBeenCalledWith({
+			to: "/projects/$projectId/settings",
+			params: { projectId: "proj-1" },
+		});
+		expect(spawnOrchestratorMock).not.toHaveBeenCalled();
 	});
 
 	it("shows the project creation startup error after navigating to the project board", async () => {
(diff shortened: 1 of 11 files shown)


JAYATIAHUJA (changes requested):
Thanks for the focused fix. The primary launcher surfaces now guard the missing orchestrator-agent case, and the added tests cover Command Palette, Sessions Board, Shell Topbar, and Sidebar. I verified the targeted renderer tests and frontend typecheck locally.  Blocking issue: rontend/src/renderer/components/RestoreUnavailableDialog.tsx:25 still calls spawnOrchestrator(session.workspaceId, 'restore_dialog', true) without checking whether the project has a configured orchestratorAgent. That dialog is another orchestrator-spawn entry point, so an archived/unrestorable orchestrator can still hit the same failed/harmful spawn path this PR is trying to prevent. Please guard this path as well, either by passing/resolving the project config and routing to /projects//settings when the orchestrator agent is missing, or by replacing/disabling the recreate action with a settings action. Please also add a focused test for the restore-unavailable dialog path.  

Verification run: 
pm --prefix frontend run test -- CommandPalette.test.tsx 
ShellTopbar.test.tsx S
idebar.test.tsx board-empty-states.test.tsx passed, and 
pm --prefix frontend run typecheck passed. 


Requesting changes until the restore-dialog spawn path is covered.

Hardik180704:
Thanks for catching this, Jayati. Fixed in `b2bd6f723`.

- `RestoreUnavailableDialog` now checks the project config before recreating.
- If the orchestrator agent is missing, it closes the dialog and opens Project Settings without calling the spawn API.
- The configured-agent path still performs the existing clean recreation.
- Added focused tests for both paths.

Local checks are green: 1,083 frontend tests, both typechecks, and 20 renderer-smoke tests passed. Fresh CI is running now. Could you take another look when you get a chance?

JAYATIAHUJA (approved):
The restore-dialog recreation path is now guarded as requested. I rechecked the current PR diff against upstream/main: it now covers the primary launcher surfaces plus RestoreUnavailableDialog, and includes focused restore-dialog coverage for the missing orchestrator-agent case and the configured-agent recreation path.  Verification run locally: 
pm --prefix frontend run test -- RestoreUnavailableDialog.test.tsx CommandPalette.test.tsx ShellTopbar.test.tsx Sidebar.test.tsx board-empty-states.test.tsx passed (5 files, 99 tests), and 
pm --prefix frontend run typecheck passed. Approving.

PR #3129, opened 2026-07-26
fix(linux): match mac shell chrome (framed board + titlebar cluster)

codebanditssss:
linux had its own `□ ← → Board` shell topbar while mac frames the board and keeps the sidebar toggle + history arrows in a fixed titlebar cluster. this makes linux use the mac path so the two look the same.

## what changed
- `hidesShellTopbar()` now covers linux → board/session actions and the Board crumb mount in-panel (framed), not in a separate shell topbar
- `TitlebarNav` renders on linux too, pinned at the sidebar's top-left (no traffic lights to offset past)
- the sidebar reserves the same titlebar band on linux (`underTopbar`)
- dropped the inline linux-only sidebar toggle — the fixed cluster replaces it. this also fixes the collapse-vanish bug: the cluster is fixed, so an off-canvas collapse always leaves a way to expand again

windows is untouched.

## supersedes #3124
that PR fixed the collapse-vanish by switching linux to an icon rail. this takes the fuller "match mac" route the vanish fix falls out of, so #3124 can close if this lands.

## verification
- `tsc --noEmit` clean, full renderer suite green (754 tests); updated `platform.test.ts` for the new linux expectations
- **needs a visual pass on linux**: i can't render the linux electron build here, so the cluster's exact position (`top-1.5 left-1.5`) and the reserved band are matched to the mac path but eyeballed. easy to nudge if the spacing is off.

image
image

### DIFF
--- a/frontend/src/renderer/components/CenterPanelShell.tsx
+++ b/frontend/src/renderer/components/CenterPanelShell.tsx
@@ -1,10 +1,11 @@
 import type { ReactNode } from "react";
 import { cn } from "../lib/utils";
 import { useWindowFullScreen } from "../hooks/useWindowFullScreen";
-import { isMacPlatform } from "../lib/platform";
+import { isLinuxPlatform, isMacPlatform } from "../lib/platform";
 import { useUiStore } from "../stores/ui-store";
 
 const isMac = isMacPlatform();
+const isLinux = isLinuxPlatform();
 
 /**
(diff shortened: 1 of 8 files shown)


Annieeeee11:
please add screenshots

codebanditssss:
> please add screenshots

done

Annieeeee11 (approved):
lgtm

PR #3130, opened 2026-07-26
fix: render bare agent logos on board cards (no tile/border)

codebanditssss:
two related fixes to the agent logos on board task cards.

## 1. render logos bare (no tile/border)
the card avatar wrapped every agent logo in a neutral tile (`rounded-md border border-border bg-raised overflow-hidden`). that boxed in brand marks that already carry their own shape/background (codex, claude, cursor, opencode, …), so they read as a logo-inside-a-box instead of the proper logo.

- `AgentAvatar` now renders the brand logo bare — no tile, border, or background — at `size-icon-xl`, letting each brand's own shape show
- agents with no asset (agy, auggie, autohand, fake) keep a bare uppercase initial, also no box

## 2. correct claude-code logo
`claude-code` shared `claude.svg` (the Anthropic sunburst). it now uses the dedicated **Claude Code** pixel mascot (lobehub `claudecode-color`, `#D97757`) so the card shows the product's own mark. the `claude` alias keeps the sunburst.

files: `AgentAvatar.tsx` + new `assets/agents/claude-code.svg`. used on every board/task card via `SessionsBoard`.

## verification
- `tsc` clean, board + shell-topbar suites green
- rendered real cards under the fake bridge: logos show bare, and a `claude-code` session shows the terracotta pixel mascot

image


Annieeeee11:
please add screenshots

Annieeeee11 (approved):
lgtm

PR #3131, opened 2026-07-26
fix(board): make the kanban grid shrink to fit (no horizontal scroll)

codebanditssss:
minimizing / narrowing the window showed a horizontal scrollbar on the kanban board (see the scroll rail under the columns). the 4-column grid carried a fixed `min-w-[64rem]` (1024px), so any pane narrower than that overflowed and scrolled.

## change
- drop `min-w-[64rem]` and the now-dead `xl:min-w-0` from the grid, and switch the wrapper from `overflow-x-auto` to `overflow-hidden`
- the columns are already `min-w-0` and cards truncate/wrap, so the grid just shares the available width — responsive at any window size, no horizontal scroll

single file: `SessionsBoard.tsx`. also removes a hardcoded arbitrary width value.

## verification
- `tsc` clean, board suite green (28 tests)
- rendered the board at 780px under the fake bridge: measured `scrollWidth <= clientWidth` (no h-scroll) and confirmed the 4 columns fit and shrink cleanly

image

### DIFF
--- a/frontend/src/renderer/components/SessionsBoard.tsx
+++ b/frontend/src/renderer/components/SessionsBoard.tsx
@@ -328,8 +328,11 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
 						spawnError={visibleSpawnError}
 					/>
 				) : (
-					<div className="h-full overflow-x-auto overflow-y-hidden">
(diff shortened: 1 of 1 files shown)


Annieeeee11 (approved):
lgtm

PR #3132, opened 2026-07-26
fix(board): show agent brand logo on archive cards (was text label)

codebanditssss:
the ARCHIVE cards under the board printed the agent as a plain text label ("Claude", "OpenCode", …), while the live task cards now show the agent's brand logo. this makes the archive match.

## change
- both archive layouts (grid + rows) render the same bare `AgentAvatar` brand logo the kanban task cards use, in place of the `agentLabel(...)` text
- removed the now-unused `agentLabel` helper
- updated the archive-card test to assert the logo (by `title=provider`) instead of the "Claude" text

single component: `SessionsBoard.tsx` (+ its test).

## verification
- `tsc` clean, board suite green (28 tests)
- rendered the expanded archive under the fake bridge: grid cards show the bare logo (claude-code mascot, codex, cursor, …) where the text used to be, matching the task cards

before 
image

after 
image

### DIFF
--- a/frontend/src/renderer/components/AgentAvatar.tsx
+++ b/frontend/src/renderer/components/AgentAvatar.tsx
@@ -58,6 +58,10 @@ type AgentAvatarProps = {
  * no tile, border, or background — so each brand's own shape shows (codex,
  * claude, cursor, … carry their own rounded backgrounds). Agents without an
  * asset fall back to a bare initial. Kept small so the title stays the hero.
+ *
+ * The provider is exposed as the accessible name (alt / aria-label), not just a
+ * hover title, so surfaces that show the logo in place of visible agent text —
+ * e.g. the archive cards — still name the agent for screen readers.
  */
 export function AgentAvatar({ provider, className }: AgentAvatarProps) {
 	const logo = LOGOS[provider];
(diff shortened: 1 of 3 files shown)


Hardik180704 (changes requested):
The visual change looks good and the implementation is nicely scoped. One accessibility issue before merge: `AgentAvatar` renders logo images with `alt=""`, so replacing the visible provider text removes the agent name from screen readers. The `title` attribute used by the test is mainly a tooltip and isn't a reliable accessible name.

Could we give the logo/fallback avatar an accessible provider name and update the test to assert that name with `getByRole(...)` instead of `getByTitle(...)`? It would also be good to cover the row layout since both archive layouts changed. Everything else looks good.

codebanditssss:
good catch, fixed. `AgentAvatar` now exposes the provider as the accessible name (`alt` on the logo, `role="img"` + `aria-label` on the initial fallback), so screen readers announce the agent instead of it being decorative. updated the test to assert via `getByRole("img", { name })` instead of `getByTitle`, and added coverage for the row archive layout too. full renderer suite is green.

Hardik180704 (approved):
The accessibility concern is fully addressed now. Both logo and fallback avatars expose the provider name, the test asserts the accessible role/name, and the row archive layout is covered too. I also verified the board suite and frontend typechecks locally, and CI is green. Looks good to merge.

PR #2731, opened 2026-07-15
fix: use context.Background for spawn rollback cleanup

tamish560:
## Summary
Spawn's rollback helpers (rollbackSpawnSeedRow, markSpawnFailedTerminated, markSpawnFailedTerminatedWithoutWorkspace) used the same context passed to Spawn. If that context was cancelled (timeout, user interrupt), the real SQLite store's DeleteSession (which uses BeginTx(ctx) / ExecContext(ctx)) and the lifecycle's MarkTerminated would fail, leaving orphan session rows stuck in "initializing" state with no runtime backing them.

## Changes
- `rollbackSpawnSeedRow`: uses `context.Background()` for `DeleteSession`
- `markSpawnFailedTerminated`: uses `context.Background()` for `MarkTerminated`
- `markSpawnFailedTerminatedWithoutWorkspace`: uses `context.Background()` for `GetSession` and `UpdateSession`
- `destroySpawnWorkspace`: uses `context.Background()` for `DeleteSessionWorktrees`

## Tests
Two new tests in `manager_test.go`:
- `TestSpawn_CleansUpOnCancelledContext`: verifies that when the runtime cancels the context during `Create`, the session row is still cleaned up (deleted or marked terminated)
- `TestSpawn_RollbackWorksOnCancelledContext_Provisioning`: verifies the same guarantee for an earlier error path (workspace creation failure) when the context is already cancelled

Both tests use context-aware store and LCM wrappers (`ctxCancelStore`, `ctxCancelLCM`) that return `ctx.Err()` when the context is cancelled, mirroring the real SQLite store's behavior. Without the fix, `DeleteSession` and `MarkTerminated` fail on the cancelled context, leaving an orphan row.

## Related
- Closes #2713
- Related to #2646 (spawn timeout leaves no-runtime session rows and locked worktrees)

PR #3137, opened 2026-07-26
fix: preserve failed scratch workspaces for retries

### LINKED ISSUE #3135: bug(daemon): failed scratch orchestrator launch leaves workspace state that blocks retries
## Summary

A failed initial orchestrator launch in a scratch project can leave AO-created agent hook files in the managed workspace. The spawn seed row is rolled back, so the next attempt reuses the same session ID and workspace path. Scratch workspace creation then refuses the non-empty directory, making the retry fail with an unrelated internal error.

Reproduced through the Electron app. I also fetched current `upstream/main` at `4736485cf`; the relevant session-manager and scratch-workspace paths are unchanged there.

## Reproduction

1. Start with a scratch project whose configured orchestrator is unavailable locally (Claude Code in this reproduction).
2. Click **Spawn Orchestrator**.
3. The initial launch fails before a runtime is created because the agent binary cannot be resolved.
4. Click **Spawn Orchestrator** again.

## Actual behavior

The retry returns `500 INTERNAL_ERROR`. The daemon reports:

```text
spawn scratch-1: workspace: scratch workspace: path ".../worktrees/scratch/orchestrators/scratch-1" already contains files: workspace: uncommitted changes present
```

The preserved directory contains only the workspace-local Claude hook files created during the failed preparation:

```text
scratch-1/.claude/settings.local.json
scratch-1/.claude/.gitignore
```

The UI exposes only the generic internal error, and subsequent retries remain blocked.

## Why it happens

The pre-launch path installs workspace-local agent hooks before validating the launch binary. On failure, `rollbackPreparedSpawnWorkspace` calls workspace destruction first and only calls agent cleanup if destruction succeeds.

Scratch `Destroy` correctly preserves every non-empty directory as potential user work. Because the AO-created hook files make the directory non-empty, destruction fails and agent cleanup never runs. The seed session row is still deleted, allowing the next spawn to reuse `scratch-1` and collide with the preserved path.

## Expected behavior

A failed pre-runtime launch should leave scratch spawn retryable without risking user files.

Possible safe approaches include:

- remove only agent-owned preparation artifacts before attempting the normal safe scratch destroy; or
- preserve the failed workspace/session identity so the next spawn does not reuse its path.

AO should not force-delete arbitrary non-empty scratch directories.

## Suggested tests

- Use the real scratch workspace adapter with an agent that writes workspace-local hook files, then fail binary validation.
- Verify a second spawn does not fail with `ErrWorkspaceDirty` from the first attempt.
- Verify unrelated user-created files are never deleted during rollback.

## Related

[#2551](https://github.com/AgentWrapper/agent-orchestrator/issues/2551) fixed stale/non-git worktrees during replacement of an existing orchestrator. This reproduction is different: it occurs during rollback of the initial scratch spawn before a live session exists.

I’m willing to work on this once the expected rollback approach is confirmed.

Image

Hardik180704:
## What

Fixes #3135 by keeping a failed scratch session when its non-empty workspace can't be removed, so retries get a fresh session number instead of colliding with the old path.

## Why

Scratch workspaces intentionally refuse to delete non-empty directories. The old rollback still deleted the seed session row, which made the next spawn reuse `scratch-1` and fail with a dirty-workspace error.

## How

- record the preserved workspace path when rollback cannot remove it
- mark that failed session as terminated
- keep the existing delete-and-cleanup behavior when workspace removal succeeds
- verify the next spawn allocates `scratch-2`

No force deletion or fallback cleanup was added, so user files remain protected.

## Testing

- `npm run lint`
- `go test -race ./internal/session_manager -count=1`
- `go build ./...`
- `go vet ./...`

## Checklist

- [x] Branched from `main`
- [x] One focused change linked to #3135
- [x] Follows `AGENTS.md` conventions
- [x] Regression coverage added

### DIFF
--- a/backend/internal/session_manager/manager.go
+++ b/backend/internal/session_manager/manager.go
@@ -350,8 +350,7 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 	// Per-project workspace provisioning: symlink shared files, then run any
 	// post-create commands (e.g. `pnpm install`) before the agent launches.
 	if err := m.provisionWorkspace(ctx, project, ws.Path); err != nil {
-		m.destroySpawnWorkspace(ctx, ws, workspaceProject)
-		m.rollbackSpawnSeedRow(ctx, id)
+		m.rollbackSeedSpawnWorkspace(ctx, rec, ws, workspaceProject, false)
 		return domain.SessionRecord{}, 0, 0, fmt.Errorf("spawn %s: provision: %w", id, err)
 	}
 
@@ -363,8 +362,7 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 	if len(cfg.Attachments) > 0 {
 		refs, err := writeSpawnAttachments(ws.Path, cfg.Attachments)
 		if err != nil {
-			_ = m.workspace.Destroy(ctx, ws)
-			m.rollbackSpawnSeedRow(ctx, id)
+			m.rollbackSeedSpawnWorkspace(ctx, rec, ws, workspaceProject, false)
 			return domain.SessionRecord{}, 0, 0, fmt.Errorf("spawn %s: attachments: %w", id, err)
 		}
 		// Keep the attachments dir out of git status. Best-effort: the images are
@@ -377,16 +375,14 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 
 	agent, ok := m.agents.Agent(cfg.Harness)
 	if !ok {
-		m.destroySpawnWorkspace(ctx, ws, workspaceProject)
-		m.rollbackSpawnSeedRow(ctx, id)
+		m.rollbackSeedSpawnWorkspace(ctx, rec, ws, workspaceProject, false)
 		return domain.SessionRecord{}, 0, 0, fmt.Errorf("spawn %s: no agent adapter for harness %q", id, cfg.Harness)
 	}
 	agentConfig := effectiveAgentConfig(cfg.Kind, project.Config)
 	env := m.runtimeEnv(id, cfg.ProjectID, cfg.IssueID, project.Config.Env)
 	m.augmentAgentRuntimeEnv(agent, env)
 	if err := m.prepareWorkspace(ctx, agent, id, ws.Path, systemPrompt, systemPromptFile, agentConfig, env); err != nil {
-		m.destroySpawnWorkspace(ctx, ws, workspaceProject)
-		m.rollbackSpawnSeedRow(ctx, id)
+		m.rollbackSeedSpawnWorkspace(ctx, rec, ws, workspaceProject, false)
 		return domain.SessionRecord{}, 0, 0, fmt.Errorf("spawn %s: %w", id, err)
 	}
 	launchCfg := ports.LaunchConfig{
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

codebanditssss (commented):
nice fix, and the test is realistic (real scratch workspace + a hook agent that makes the dir non-empty + missing binary, so it exercises the actual validateAgentBinary path). i verified the linchpin: `NextSessionNum` is `SELECT COALESCE(MAX(num),0)+1 FROM sessions WHERE project_id=?`, which ignores `is_terminated`, so keeping the row terminated correctly forces the retry to `scratch-2`. the `maxSessionNumStore` double mirrors that exactly. the refactor is also behavior-preserving on the success path (prepared=false = old early-path rollback, prepared=true = old rollbackPreparedSpawnWorkspace).

things worth addressing:

**uneven preserved-path recording.** only the pre-runtime paths call `preserveFailedSpawnWorkspace`. the two post-runtime paths (MarkSpawned ~451, deliverAfterStartPrompt ~457) still use `rollbackPreparedSpawnWorkspace` + `markSpawnFailedTerminated[WithoutWorkspace]`. they keep the row terminated so `scratch-2` still works, but they don't record the preserved path, and the `...WithoutWorkspace` variant actively clears the workspace metadata even though the dir survives on disk. so a scratch failure after runtime start leaves an orphaned dir with no row pointing at it. the allocation fix is fine everywhere, just flag whether the path recording should be uniform.

**hard conflict with #2731.** both PRs refactor these exact rollback call sites and make different decisions on the deliverAfterStartPrompt path (#2731 adds detached contexts + a wsOK/rtErr branch). whoever merges second needs a real rebase, worth coordinating.

minor:
- `preserveFailedSpawnWorkspace` logs the same warning string for both the GetSession-fail and UpdateSession-fail branches, differentiate them so logs are diagnosable.
- same latent ctx-cancellation fragility #2731 targets: `preserveFailedSpawnWorkspace`/`markSpawnFailedTerminated` use the passed ctx (fine for the scratch scenario since ctx is live there, but if #2731 lands first these should use its detached ctx too).
- the test only covers the validateAgentBinary (prepared=true) path; the prepared=false early paths share the helper so lower-risk but untested.

security-wise no force-delete added, user files stay protected as claimed. one tradeoff: orphaned scratch dirs accumulate on repeated failures, a GC pass could be a follow-up.

net: approve with nits. correct and well-tested. before merge i'd align (or justify) the post-runtime path recording, differentiate the duplicate log line, and coordinate the rebase against #2731.

Hardik180704:
Thanks @codebanditssss — addressed the follow-ups in a8a58062d:

- MarkSpawned and after-start delivery failures now preserve the scratch workspace path when cleanup fails.
- Runtime handles are cleared only when runtime destruction succeeds.
- Rollback warnings now distinguish load, missing-session, and update failures.
- Added regression coverage for both early provisioning failure and after-start delivery failure with non-empty scratch workspaces.

I left the cancellation-context cleanup to #2731 since that PR already owns that behavior and is still open. The branch is rebased on the latest main, and npm run lint, the full session-manager suite, race tests, go build, and go vet all pass locally. GitHub CI is running now.

codebanditssss (approved):
the follow-up commit resolves the review:

- **preserved-path recording is uniform now**: `rollbackPreparedSpawnWorkspace` takes `runtimeDestroyed` + returns a bool and preserves the path on destroy-failure, so the post-runtime paths (MarkSpawned, deliverAfterStartPrompt) record it too, not just the pre-runtime ones.
- the `...WithoutWorkspace` edge is gated correctly: it only clears the workspace metadata when `runtimeDestroyed && workspaceDestroyed`, and `preserveFailedSpawnWorkspace` only clears the runtime handles when `runtimeDestroyed` — so a surviving scratch dir keeps a row pointing at it.
- the duplicate warning is split into three distinct messages.

signature changes are consistent (both callers updated, `rollbackSeedSpawnWorkspace` passes `true` since no runtime exists on the pre-runtime paths). the fix is still verified by the earlier `MissingBinaryPreservesNonEmptyScratchWorkspaceForRetry` test, and `NextSessionNum` (MAX(num)+1 over all rows) still guarantees the retry gets `scratch-2`.

approving.

one coordination note (not code): this and #2731 both add a `runtimeDestroyed`/`wsOK` branch to the same deliverAfterStartPrompt rollback, so they'll still conflict at merge. whichever lands first, the other needs a rebase.

Issue #3113, opened 2026-07-26
[Feature]: Notion like tabs for project sessions


amoreX:
### Problem

Ability for worktrees of one proj to be displayed as tabs as well, for better ux

### Proposed direction

<img width="2160" height="1098" alt="Image" src="https://github.com/user-attachments/assets/5116c772-23df-4061-984a-46d6a4c286a3" />

<img width="1808" height="802" alt="Image" src="https://github.com/user-attachments/assets/83431bb2-458e-47ae-8f79-cfb9172253e6" />

### Before submitting

- [x] I searched existing issues / discussions for something similar
- [x] I am willing to help implement this if maintainers agree on scope

Manyy8:
Hi! I'd like to work on this enhancement if it's available. Could you please assign it to me?

somewherelostt:
Thanks for contributing to Agent Orchestrator.

This issue is being picked up by the current external contributor on-call pair:

- @illegalcall
- @pulkit7070

If someone is already working on this, please continue as usual.
The on-call pair is added for visibility, tracking, and support, not to take over the work.
If you need help with review, direction, reproduction, or next steps, please tag @illegalcall and @pulkit7070 here.

For faster context or live questions, you can also join the AO Discord.

Join the session here:
https://discord.gg/H6ZDcUXmq

Come by if you want to see what is being built, ask questions, or just hang around with the community.

i-trytoohard:
From the NotesBot call on `2026-08-06`, this existing issue appears to be relevant.
Match confidence: `0.58` (fuzzy).

Relevant discussion excerpts:
- - prateek summarized that for orchestrator notifications it is project title, and for session-related notifications it is session title, and said he would check it. ## 🧪 Browser Tabs, File Diff & Dev Suite
- - -Prasad- said session title is used where needed, while for orchestrator-spawn notifications the project title is shown instead. - prateek summarized that for orchestrator notifications it is project title, and for session-related notifications it is session title, and said he would check it. ## 🧪 Browser Tabs, [...]
- - prateek summarized that for orchestrator notifications it is project title, and for session-related notifications it is session title, and said he would check it. ## 🧪 Browser Tabs, File Diff & Dev Suite - .phylolver(vaibhaav) said he is working “in chunks” and currently focused on the Dev Suite, browser tabs, [...]

Source thread: https://discord.com/channels/1476302178913357958/1534979721245364424/1534979721245364424

[...]

PR #3138, opened 2026-07-26
Feat/project session tabs

amoreX:
#3113  
tldr: 
• Persistent, isolated tab layouts per session. 
• Add terminals or live sessions across projects. 
• Compact 5-session launcher with search and scrollable “Show all”. 
• Closing/switching tabs preserves the originating session context.

### DIFF
--- a/frontend/src/renderer/components/CenterPane.test.tsx
+++ b/frontend/src/renderer/components/CenterPane.test.tsx
@@ -18,6 +18,12 @@ const worker = {
 	updatedAt: "2026-06-10T00:00:00Z",
 	prs: [],
 } satisfies WorkspaceSession;
+const secondWorker = {
+	...worker,
+	id: "sess-2",
+	title: "review the change",
+	branch: "ao/sess-2",
+} satisfies WorkspaceSession;
 
 describe("CenterPane toolbar session label", () => {
 	const makeShells = (count: number) =>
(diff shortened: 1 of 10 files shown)


Pulkit7070 (commented):
Reviewed and exercised the tab feature. It is well-built, with careful edge handling.

Verified:
- Unit: CenterPane, SessionView, ui-store, and shell-new-session-shortcut suites pass (56 tests), typecheck clean.
- Playwright against the real renderer: seeding localStorage ao.sessionTabs with {owner: [pinnedSessionId]} and loading the owner session renders the pinned tab (a "Close session tab idle worker" control appears) and the store round-trips, so persistent layouts work end to end. Separately, a persisted tab pointing at a session id that no longer exists is silently filtered out and the view renders without crashing. The Add tab launcher opens with its search field.

Edge cases handled well:
- Stale references. SessionView maps stored tab ids through allSessions.find and filters out anything that does not resolve, plus terminated sessions and the owner itself, so a persisted tab for a deleted or terminated session just disappears rather than rendering a broken tab.
- ui-store persistence is defensive: initialSessionTabs validates the parsed JSON is an object whose values are string arrays and drops malformed entries, addSessionTab dedupes and rejects self-pinning, and removeSessionTab deletes the owner key when its last tab is closed.
- Closing the currently-viewed tab navigates back to the owner session rather than stranding the view, and owner context is preserved via the tabOwner route search param.

One small non-blocking note: stored tab ids that no longer resolve to a live session are filtered at render and only removed from localStorage on an explicit close. So if an owner keeps tabs and a pinned session is deleted elsewhere, that dead id lingers in ao.sessionTabs indefinitely. It is harmless (small strings, filtered on read), but an opportunistic prune of unresolved ids when building projectSessions would keep the persisted map from accumulating dead references over time.

No defects found.

Pulkit7070:
Follow-up: ran a fuller runtime stress pass on the interactive paths, not just persistence. All green against the real renderer.

- Full lifecycle through the actual Add tab launcher: open the menu, select a session, it pins as a tab and writes ao.sessionTabs, then closing it removes the tab and prunes the entry. Consistent across a 5x rapid add/close churn with no crash or leftover state.
- Tab switch preserves owner context end to end: clicking a pinned tab navigates to /projects/<proj>/sessions/<tabSessionId>?tabOwner=<ownerId>, and the owner's tab strip stays intact after the switch. This is the cross-tab navigation and tabOwner round-trip, verified in the browser rather than only read.
- Adding a Terminal from the launcher does not crash the strip.
- Earlier checks still hold: 56 unit tests pass, a persisted tab resolves and renders, and a persisted tab pointing at a nonexistent session is filtered out without crashing.

One honest scope caveat: the browser harness models a single project, so true cross-project pinning (a tab whose session lives in a different project) is covered by the unit test that includes a second project plus the navigation code (params use projectSession.workspaceId, so the target project is threaded correctly), but I did not exercise a two-project layout in the browser. The single-project interactive paths above are fully runtime-verified.

The only follow-up I would still suggest is opportunistic pruning of stored tab ids that no longer resolve, since dead ids otherwise linger in ao.sessionTabs until an explicit close. Not a defect.

Pulkit7070 (approved):

PR #3140, opened 2026-07-26
feat(landing): add 0.11.0 changelog entry + restore the detail route

codebanditssss:
the changelog page showed **"No updates yet."** the site's changelog now builds itself from GitHub Releases, with the curated 0.11.0 entry on top.

## what
- `getChangelogEntries` fetches stable `vX.Y.Z` releases from the GitHub API at build time and merges them with any curated `content/changelog` MDX (curated wins for the same version, so 0.11.0 "The Floor" stays until v0.11.0 is tagged)
- filters out nightlies, per-PR prereleases, package tags, and empty-body releases
- **release bodies render as plain Markdown** (`react-markdown` + `remark-gfm`), never through the MDX compiler — so arbitrary release text (`import`/`export` lines, JSX, `{expressions}`) can't break the static build. curated entries still compile as MDX; a `source` field picks the renderer
- restored the missing `/changelog/[slug]` detail route (`generateStaticParams`); the data fns went async (list page, `[slug]`, sitemap, `changelog.xml` updated)

## refresh mechanism
the changelog is baked in at build, so it updates whenever the landing is **redeployed** (a push to `main`, or a manual dispatch). making a newly published release appear automatically needs a `release: [published]` + `schedule` trigger on `deploy-landing.yml` — that workflow edit can't be pushed from this PR's token (no `workflow` scope), so it's a small follow-up a maintainer applies alongside.

## verification
- `next build` (`output: export`) generates `/changelog` + a detail page per entry (0.11.0 + v0.10.3 → v0.2.0)
- reproduced the reviewer's failing case (`export const metadata = {…}` in a body) through react-markdown: renders clean, text kept literal, no throw

### DIFF
--- a/frontend/src/landing/src/lib/changelog.ts
+++ b/frontend/src/landing/src/lib/changelog.ts
@@ -12,9 +12,9 @@ export {
 
 const CHANGELOG_DIR = path.join(process.cwd(), "content/changelog");
 
-// Releases are pulled from GitHub so the changelog updates itself on each build
-// (the landing redeploys on push to main + a daily schedule). Curated MDX
-// entries in content/changelog still take precedence for the same version.
+// Releases are pulled from GitHub at build time, so the changelog refreshes
+// whenever the landing is redeployed (a push to main, or a manual deploy).
+// Curated MDX entries in content/changelog take precedence for the same version.
 const RELEASES_REPO = "AgentWrapper/agent-orchestrator";
 // Only stable vMAJOR.MINOR.PATCH tags — skips nightlies, per-PR prereleases,
 // and package tags like @composio/ao@x.
@@ -45,6 +45,7 @@ function parseFrontmatter(filePath: string): ChangelogEntry | null {
 			date: dateValue,
 			image: data.image,
 			content,
+			source: "mdx",
 			draft: data.draft === true,
 		};
 	} catch {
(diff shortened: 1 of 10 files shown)


frontend/src/landing/src/lib/changelog.ts
```diff
@@ -35,38 +52,95 @@ function parseFrontmatter(filePath: string): ChangelogEntry | null {
...
 
-export function getChangelogEntry(slug: string): ChangelogEntry | undefined {
-	const filePath = path.join(CHANGELOG_DIR, `${slug}.mdx`);
+// Release bodies are plain GitHub markdown, not MDX. Escape the constructs MDX
+// would otherwise try to evaluate (JSX tags, {expressions}) everywhere except
+// code spans, so an arbitrary release body can never break the build.
+function sanitizeMdx(markdown: string): string {
+	return markdown
+		.split(/(```[\s\S]*?```|`[^`\n]*`)/g)
+		.map((segment, i) => (i % 2 === 1 ? segment : segment.replace(/[<{}]/g, (c) => `\\${c}`)))
```
Hardik180704: **Blocking:** this does not fully protect arbitrary GitHub Markdown from being parsed as MDX. I reproduced a compile failure with the release-body line `export const metadata = { version: "1.0.0" }`: escaping the braces leaves an MDX ESM construct that `compileMDX` rejects. A future stable release containing this kind of text could therefore break the landing deployment. Could we either render fetched release bodies through a Markdown-only path, or sanitize MDX ESM constructs too, with focused tests covering import/export lines, JSX, expressions, and fenced/inline code?

Hardik180704 (dismissed):
Nice direction overall—the curated 0.11.0 page, detail route, feeds, and GitHub-release merge are put together cleanly. I found one release-body case that can still break the static build, though, so I think that needs addressing before merge.

I also noticed the comment says the landing deploy runs daily, but `.github/workflows/deploy-landing.yml` currently only runs for landing-file pushes or manual dispatch. That means a newly published GitHub release will not appear automatically until another landing deployment happens. Please add the release/scheduled trigger as part of this feature, or adjust the behavior/docs so the refresh mechanism is accurate.

codebanditssss:
thanks — both addressed.

**blocking (MDX parsing of release bodies):** fixed structurally rather than by patching the sanitizer. a release body is plain Markdown, so it now renders through a Markdown-only path (`react-markdown` + `remark-gfm`) and never touches the MDX compiler. curated `content/changelog` entries still compile as MDX (they use custom components); entries carry a `source` field to pick the renderer, and `sanitizeMdx` is gone.

i reproduced your exact case to confirm: rendering `export const metadata = { version: "1.0.0" }` (plus `import` lines, raw `<tags>`, `{expressions}`, and fenced code) through react-markdown returns cleanly and keeps the text literal — no throw. the full static export also passes over the real release bodies. since release bodies bypass `compileMDX` entirely, that whole class of failure is gone by construction, which is why i didn't add a sanitizer test suite (there's no sanitizer left, and the landing has no test runner set up).

**deploy trigger accuracy:** you're right, good catch — the comment/description claimed a daily schedule the workflow doesn't have. i've corrected the code comment (it refreshes on the next landing redeploy: a push to main or a manual dispatch). the actual `release: [published]` + `schedule` triggers are the right addition, but that edit to `.github/workflows/deploy-landing.yml` can't be pushed from this PR's token (it lacks `workflow` scope), so it's being applied separately by a maintainer. flagging that so the two land together.

**react-doctor js-combine-iterations:** `getReleaseEntries` now filters + maps in a single `for…of` pass.

JAYATIAHUJA (approved):

PR #3145, opened 2026-07-26
fix: close session lifecycle gaps

### LINKED ISSUE #3087: Follow-up to #2980: remaining session lifecycle correctness gaps
## Origin

This is a follow-up derived from merged PR [#2980 — `fix: unify session state and lifecycle presentation`](https://github.com/AgentWrapper/agent-orchestrator/pull/2980).

The PR's central model is sound: agent activity, terminal-runtime lifetime, durable session termination, and SCM outcome are independent facts. A deeper implementation review found the six correctness gaps below. They remain present after the final PR head and were revalidated against `main` at `f0512c885`.

This issue is not reopening the product decisions made in #2980. It tracks concrete implementation failures and missed consumers that remain after that PR.

## Shared invariant

The implementation should preserve these rules:

- An exited agent is not necessarily a terminated AO session.
- A merged PR does not by itself prove whether the session is alive.
- Once Kill accepts terminal intent, Resume must not resurrect the session.
- External runtime side effects and their durable launch generation must not disagree.
- Independently owned settings must not be overwritten by stale lifecycle snapshots.
- UI chronology and navigation must use the underlying lifetime/timestamp facts rather than construction order or a derived display label.

## Remaining failures

| # | Failure | Impact | Area |
|---|---|---|---|
| 1 | Scratch sessions can never use Resume agent | High | Daemon + desktop |
| 2 | Successful tmux respawn followed by probe failure leaves an unfenced agent running | High | Daemon |
| 3 | Resume races Kill and can undo terminal intent | High | Daemon |
| 4 | Lifecycle activity can silently overwrite the terminate-on-merge preference | Medium | Daemon/storage |
| 5 | Activity history is not actually reverse chronological | Medium | Desktop |
| 6 | Terminated merged sessions are promoted as needing attention | Medium | Desktop |

### 1. Scratch sessions can never use Resume agent

**Expected behavior**

Scratch projects are intentionally branchless. An exited scratch agent with an existing workspace and runtime handle should therefore be resumable without a Git branch.

**Why the code fails**

`DefaultSpawnBranch` deliberately returns an empty branch for scratch projects, but `ResumeAgentWithMode` rejects every empty branch before it loads the project and can apply the same scratch exception already used by Restore.

- [Scratch sessions intentionally get an empty branch](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/session_manager/manager.go#L2049-L2059)
- [Resume treats every empty branch as an incomplete handle](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/session_manager/manager.go#L936-L971)
- [The desktop still offers Resume based on live + exited state](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/components/SessionInspector.tsx#L324-L337)

**Real-world situation**

A first-time user chooses a scratch project to prototype a JSON-to-CSV tool without connecting a Git repository. The agent finishes and exits. Ten minutes later, the user notices quoted commas are handled incorrectly and clicks **Resume agent** to request one follow-up change.

**What the user sees**

The Resume button is present and the files and terminal still exist, but every attempt fails with `ErrIncompleteHandle`. The intentionally branchless session is being treated as corrupted state.

**Fix/proof shape**

- Load the project before validating the branch.
- Require a branch only for non-scratch project kinds, matching Restore.
- Add an exited-scratch Resume test and a controller/UI path test.

### 2. tmux can commit the respawn side effect but report failure

**Expected behavior**

Resume must leave AO in one of two coherent states: either the new agent generation did not start, or it started and the same generation is durably authoritative.

**Why the code fails**

`respawn-pane -k` kills the old pane workload and launches the new one before the separate `IsAlive` verification. If that verification has a transient error, `Restart` returns an empty handle and an error. The caller cancels the pending generation but has no successful handle with which to compensate.

- [`respawn-pane -k` runs before the fallible liveness verification](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/adapters/runtime/tmux/tmux.go#L247-L283)
- [The relaunch path returns immediately on that error without committing the launch ID](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/session_manager/manager.go#L1030-L1063)

**Real-world situation**

A developer clicks **Resume agent** while their laptop is under load. tmux successfully executes `respawn-pane -k`, and the replacement Codex process begins starting, but the following `has-session` command briefly times out or returns a transient tmux error.

**What the user sees**

AO reports “Resume failed” and leaves the session in Exited even though a new agent is running. Its generation-specific hooks are rejected against the old durable generation. Clicking Resume again can replace that running process with yet another generation and discard work it already began.

**Fix/proof shape**

- Preserve ownership after the destructive respawn step.
- Either compensate by destroying the restarted handle, return an explicit “applied but unverified” result for reconciliation, or make verification/rollback part of the runtime contract.
- Add a test where respawn succeeds and only the following liveness probe fails.

### 3. Resume races Kill and can undo terminal intent

**Expected behavior**

Once explicit Kill or terminate-on-merge accepts terminal intent, a concurrent Resume must not clear termination or recreate the runtime.

**Why the code fails**

The per-session Resume exclusion only prevents a second Resume. Kill does not participate in that operation boundary. Both paths can read the same live/exited record, mutate external resources, and later commit conflicting lifecycle states. `MarkSpawned` unconditionally clears `IsTerminated`.

- [Kill has no shared arbitration with Resume](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/session_manager/manager.go#L713-L783)
- [Resume validates once, restarts later, and is guarded only against another Resume](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/session_manager/manager.go#L936-L987)
- [`MarkSpawned` clears termination unconditionally](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/lifecycle/manager.go#L790-L811)

**Real-world situation**

An agent exits just as its PR merges. The user clicks **Resume agent** to request a small follow-up while the enabled terminate-on-merge policy starts Kill. Both requests read the session as live and exited before either completes.

**Failure sequences**

1. Resume starts the replacement runtime.
2. Kill destroys runtime/workspace and marks the row terminated.
3. Resume calls `MarkSpawned`, clears termination, and leaves a live row whose resources are gone.

Or:

1. Kill destroys the old runtime and preserves a dirty workspace.
2. Resume observes no live runtime and creates a replacement.
3. Resume clears termination, defeating explicit or merge-driven cleanup.

**What the user sees**

The session may move to Archive and then reappear, remain live with a missing terminal/worktree, or continue running after the user configured it to terminate.

**Fix/proof shape**

- Share one per-session operation boundary across Resume, Kill, Restore, and replacement teardown, or use a conditional final transition tied to the expected state/generation.
- Make accepted terminal intent monotonic.
- Add deterministic tests for both Resume/Kill interleavings.

### 4. Lifecycle activity can erase the terminate-on-merge preference

**Expected behavior**

Once the merge policy update succeeds, unrelated activity callbacks must not revert it.

**Why the code fails**

The preference setter performs a narrow update, while lifecycle mutations read a complete session and later write every mutable column—including `terminate_on_pr_merge`. The store's write mutex serializes individual writes, but not the lifecycle read and later full-row commit as one atomic operation.

- [Full-row `UpdateSession` includes the policy column](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/storage/sqlite/queries/sessions.sql#L14-L22)
- [The policy setter is a separate narrow update](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/storage/sqlite/queries/sessions.sql#L61-L62)
- [Lifecycle performs a read-modify-full-write sequence](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/backend/internal/lifecycle/manager.go#L195-L212)

**Real-world situation**

While an agent is finishing a tool call, the user enables **Terminate on merge**. An activity callback had already read the row with the preference set to false. The narrow preference update commits true; the callback then commits its older full-row snapshot.

**What the user sees**

The switch may turn on and then silently flip back after CDC refresh. If the user navigates away before noticing, the PR later merges while the runtime and workspace remain alive, contrary to the setting they believed they saved.

**Fix/proof shape**

- Use lifecycle-specific partial update queries, a transaction spanning read/write ownership, or optimistic concurrency with an expected revision.
- Add a deterministic lost-update test that pauses lifecycle after its read, changes the preference, then releases the lifecycle write.

### 5. The Activity timeline is not reverse chronological

**Expected behavior**

Historical milestones should be ordered by their actual event timestamps, with the live current-state row kept separately at the top.

**Why the code fails**

The inspector appends events in category blocks—Created, Draft, Opened, Merged, Done—and then reverses the assembled array. It never sorts by the underlying timestamps.

- [Timeline events are assembled by category](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/components/SessionInspector.tsx#L492-L531)
- [`history.reverse()` substitutes construction order for chronology](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/components/SessionInspector.tsx#L533-L557)

**Real-world situation**

A worker owns a stacked pair of PRs. The root PR merged at 10:00, and the child PR opened at 11:45 after its dependency cleared. At noon, the user opens Activity to understand the sequence.

**What the user sees**

“Merged root PR” can appear above “Opened child PR” even though the opening happened later. The timeline implies the wrong causal sequence and can mislead stacked-PR or CI diagnosis.

**Fix/proof shape**

- Keep the raw timestamp on each historical event.
- Sort descending by parsed time with a deterministic fallback for missing/invalid values.
- Prepend the current untimestamped row after sorting.
- Add tests whose category order deliberately conflicts with timestamp order.

### 6. Terminated merged sessions still appear under Needs attention

**Expected behavior**

A terminated merged session should remain archived/search-only. `merged` describes SCM outcome; `isTerminated` describes lifetime and navigation eligibility.

**Why the code fails**

The command palette promotes every merge-zone session into `attentionSessions` without checking `sessionIsActive`. Its ID is then excluded from the later branch that would mark inactive sessions `searchOnly`.

- [Merge-zone sessions enter attention without a lifetime check](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/lib/command-palette.ts#L156-L170)
- [Only the later non-attention branch applies `searchOnly`](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/lib/command-palette.ts#L183-L188)
- [The existing test claims merged/terminated coverage but asserts only the terminated-status fixture](https://github.com/AgentWrapper/agent-orchestrator/blob/f0512c8857bfa2397703c0773394ad219133222c/frontend/src/renderer/lib/command-palette.test.ts#L168-L197)

This is best classified as a missed consumer exposed by #2980's explicit merged-vs-terminated model; it may predate the PR rather than being newly introduced by it.

**Real-world situation**

A user merges a payment-retry PR, terminates the session, and sees it enter Archive. The next morning they open the command palette to jump to active work.

**What the user sees**

The old terminated session is promoted under **Needs attention** without a search query. Selecting it opens archived context instead of actionable work and crowds genuinely live attention items.

**Fix/proof shape**

- Require `sessionIsActive(session)` in the attention filter.
- Add an explicit `status=merged, isTerminated=true` fixture and assert that it is search-only and absent from default suggestions.

## Completion criteria

- [ ] Exited scratch sessions can resume successfully without a branch.
- [ ] A post-respawn verification error cannot leave an untracked live generation.
- [ ] Kill/terminate-on-merge cannot be undone by an overlapping Resume.
- [ ] Lifecycle writes cannot revert an independently updated merge policy.
- [ ] Historical Activity events are sorted by actual timestamps.
- [ ] Terminated merged sessions stay out of default attention suggestions.
- [ ] Deterministic regression tests cover every failure sequence above.

## Suggested implementation order

1. Resume/Kill arbitration and monotonic terminal intent.
2. tmux restart ownership after partial success.
3. Scratch Resume validation.
4. Policy lost-update protection.
5. Timeline chronology.
6. Command-palette active-session filtering.

Related merged PR: [#2980](https://github.com/AgentWrapper/agent-orchestrator/pull/2980).

saiprasannasastry:
## Summary

- make agent Resume project-aware so branchless Scratch sessions can resume while repository-backed sessions still require a branch
- serialize Resume, Kill, Restore, replacement retirement, and terminal Cleanup per session, with a re-read before Cleanup destroys resources
- fence the durable Resume commit against concurrent terminal intent and launch-generation changes
- make tmux create/restart transitions cancellation-safe: proven post-respawn probe failures retain ownership, while dispatch-timeout uncertainty is separately typed and compensated instead of being falsely adopted
- keep non-atomic runtime replacement (used by Windows ConPTY) inside one bounded, cancellation-detached Destroy/Create transition, while leaving genuine Create failures exited and retryable
- keep `terminate_on_pr_merge` under its dedicated setter so stale generic lifecycle writes cannot revert user policy
- sort activity history by raw timestamps with deterministic invalid/missing fallbacks
- keep terminated sessions out of command-palette attention actions while retaining them in search

## Root cause

Several lifecycle boundaries treated validation, runtime mutation, and persistence as independent operations. Scratch Resume inherited repository-only branch validation; Resume and Kill could cross each other; a successful tmux respawn followed by a transient probe error was reported as total failure; cancellation could split a destructive runtime command from durable adoption or compensation; Cleanup could act on a stale terminated snapshot after Restore had recreated resources; and generic session updates could overwrite independently owned policy state. The renderer also derived timeline ordering and attention actions from presentation-oriented state instead of the underlying lifecycle facts.

## Concurrency and ownership guarantees

- operations for the same session share one ref-counted lock; different session IDs remain independent
- both Resume-first/Kill-second and Kill-first/Resume-second orderings preserve monotonic terminal intent
- Cleanup re-reads the session under that lock, so it cannot destroy a runtime/workspace recreated by a winning Restore
- caller cancellation is honored before a destructive runtime boundary; after dispatch, bounded detached work reaches a coherent adoption or compensation result
- a confirmed tmux respawn plus failed verification is adoptable, but a dispatch-timeout outcome is never treated as proven success and is compensated using the owned handle
- the test-only post-respawn probe runner is mutex-protected, exact-target-specific, and supports interleaved targets without one probe stealing another target's injected failure

## Impact

Scratch sessions can resume reliably, concurrent lifecycle operations preserve terminal monotonicity, Cleanup cannot tear down freshly restored resources, and cancellation or transient tmux failures cannot silently leave an authoritative generation disagreed with runtime reality. Merge policy and renderer state remain consistent across asynchronous updates.

## Validation

- `GOFLAGS=-p=4 npm run lint` — all backend tests and golangci-lint, 0 issues
- `GOFLAGS=-p=4 go test -race ./... -timeout=10m`
- focused concurrency/cancellation suites under `-race -count=50` for Resume/Kill, Cleanup/Restore, fallback Destroy/Create, tmux Create/Restart, compensation, retry, and target-specific probe injection
- real isolated tmux daemon: race-enabled post-respawn failure integration test passed three consecutive runs
- full frontend Vitest suite: 87 files / 1,086 tests
- packaged Electron → real daemon/SQLite → real agent process/hooks → real isolated tmux Resume E2E, passed twice
- `git diff --check`

Fixes #3087

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -275,7 +275,13 @@ func (r *Runtime) Restart(ctx context.Context, handle ports.RuntimeHandle, cfg p
 	}
 	alive, err := r.IsAlive(ctx, handle)
 	if err != nil {
-		return ports.RuntimeHandle{}, fmt.Errorf("tmux runtime: verify restarted session %s: %w", id, err)
+		// respawn-pane -k has already killed the old workload and launched the
+		// replacement. Preserve ownership of that applied side effect so the
+		// session manager can durably adopt the new generation even when this
+		// separate probe fails transiently (including caller cancellation).
+		return handle, &ports.RestartAppliedUnverifiedError{
+			Cause: fmt.Errorf("tmux runtime: verify restarted session %s: %w", id, err),
+		}
 	}
 	if !alive {
 		return ports.RuntimeHandle{}, fmt.Errorf("tmux runtime: session %s exited during restart", id)
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -201,8 +201,27 @@ func (r *Runtime) Create(ctx context.Context, cfg ports.RuntimeConfig) (ports.Ru
 
 	launchCmd := buildLaunchCommand(cfg)
 	args := newSessionArgs(id, cfg.WorkspacePath, r.shell, launchCmd)
-	if _, err := r.run(ctx, args...); err != nil {
+	// new-session is the creation ownership boundary. Honor cancellation before
+	// dispatch, then let the local tmux client finish on its bounded detached
+	// context. If that client times out after dispatch, destroy the exact session
+	// name before returning so a possibly-created process cannot run without a
+	// durable owner.
+	if err := ctx.Err(); err != nil {
 		return ports.RuntimeHandle{}, fmt.Errorf("tmux runtime: create session %s: %w", id, err)
+	}
+	if outcomeUncertain, err := r.runCommitted(ctx, args...); err != nil {
+		cause := fmt.Errorf("tmux runtime: create session %s: %w", id, err)
+		if !outcomeUncertain {
+			return ports.RuntimeHandle{}, cause
+		}
+		handle := ports.RuntimeHandle{ID: id}
+		if destroyErr := r.Destroy(context.WithoutCancel(ctx), handle); destroyErr != nil {
+			return ports.RuntimeHandle{}, errors.Join(
+				cause,
+				fmt.Errorf("tmux runtime: compensate outcome-uncertain create session %s: %w", id, destroyErr),
+			)
+		}
+		return ports.RuntimeHandle{}, fmt.Errorf("tmux runtime: compensated outcome-uncertain create session %s: %w", id, err)
 	}
 	if err := r.verifyPaneWorkingDirectory(ctx, id, cfg.WorkspacePath); err != nil {
 		_ = r.Destroy(context.Background(), ports.RuntimeHandle{ID: id})
@@ -270,15 +289,35 @@ func (r *Runtime) Restart(ctx context.Context, handle ports.RuntimeHandle, cfg p
 	}
 
 	launchCmd := buildLaunchCommand(cfg)
-	if _, err := r.run(ctx, respawnPaneArgs(id, cfg.WorkspacePath, r.shell, launchCmd)...); err != nil {
+	// respawn-pane -k is destructive: once tmux begins it, cancelling the
+	// client can leave us unable to tell whether the server killed the old
+	// workload and launched the replacement. Define the cancellation boundary
+	// immediately before that command and let an accepted operation finish on
+	// its own bounded context. Verification is likewise detached and bounded so
+	// caller cancellation cannot discard a useful definitive result after the
+	// destructive boundary.
+	if err := ctx.Err(); err != nil {
 		return ports.RuntimeHandle{}, fmt.Errorf("tmux runtime: restart session %s: %w", id, err)
 	}
-	alive, err := r.IsAlive(ctx, handle)
+	if outcomeUncertain, err := r.runCommitted(ctx, respawnPaneArgs(id, cfg.WorkspacePath, r.shell, launchCmd)...); err != nil {
+		cause := fmt.Errorf("tmux runtime: restart session %s: %w", id, err)
+		if outcomeUncertain {
+			// The tmux client timed out after dispatch. The server may already
+			// have applied respawn-pane, so conservatively preserve ownership
+			// and require the session manager to compensate it rather than
+			// adopting a generation that may never have started.
+			return handle, &ports.RestartOutcomeUncertainError{Cause: cause}
+		}
+		return ports.RuntimeHandle{}, cause
+	}
+	verifyCtx, cancelVerify := context.WithTimeout(context.WithoutCancel(ctx), r.timeout)
+	defer cancelVerify()
+	alive, err := r.IsAlive(verifyCtx, handle)
 	if err != nil {
 		// respawn-pane -k has already killed the old workload and launched the
 		// replacement. Preserve ownership of that applied side effect so the
 		// session manager can durably adopt the new generation even when this
-		// separate probe fails transiently (including caller cancellation).
+		// separate probe fails transiently.
 		return handle, &ports.RestartAppliedUnverifiedError{
 			Cause: fmt.Errorf("tmux runtime: verify restarted session %s: %w", id, err),
 		}
@@ -549,6 +588,29 @@ func attachEnv(base []string) []string {
 // run wraps runner.Run with a per-call timeout context.
 func (r *Runtime) run(ctx context.Context, args ...string) ([]byte, error) {
 	return r.runCommand(ctx, r.binary, args...)
+}
+
+// runCommitted runs a destructive tmux command after its caller has accepted
+// the cancellation boundary. The command is isolated from caller cancellation
+// but remains bounded by the runtime timeout. A nil runner error wins a
+// simultaneous timeout race because it is definitive evidence that the tmux
+// client completed successfully.
+//
+// A runner error at the internal timeout remains inherently ambiguous: the
+// local tmux client may have timed out before or after the server applied the
+// command. outcomeUncertain tells the caller to retain ownership until it can
+// compensate the possible side effect.
+func (r *Runtime) runCommitted(ctx context.Context, args ...string) (outcomeUncertain bool, err error) {
+	cmdCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), r.timeout)
+	defer cancel()
+	out, err := r.runner.Run(cmdCtx, nil, r.binary, args...)
+	if err == nil {
+		return false, nil
+	}
+	if cmdCtx.Err() != nil {
+		return true, cmdCtx.Err()
+	}
+	return false, commandError{err: err, output: strings.TrimSpace(string(out))}
 }
 
 func (r *Runtime) runCommand(ctx context.Context, name string, args ...string) ([]byte, error) {
--- a/backend/internal/adapters/runtime/tmux/tmux_integration_test.go
+++ b/backend/internal/adapters/runtime/tmux/tmux_integration_test.go
@@ -2,10 +2,13 @@ package tmux
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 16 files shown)


Rishet11:
Closing this because issue #3087 was closed as completed by a maintainer on Aug 23 through other work while this PR remained unreviewed, and the branch has since conflicted with main. The remaining lifecycle work is tracked in the still-open #3987. Thanks for the deep work on resume fencing and tmux cancellation safety.

PR #3146, opened 2026-07-26
fix(landing): recenter hero preview after resize

illegalcall:
## Summary

- Re-derive the untouched hero preview window layout when its container resizes.
- Preserve user-requested drag and resize geometry across responsive viewport changes.
- Port the fix from ronishrohan/aoagents#2 to the moved landing-page source.

## Root cause

The resize observer clamped the last rendered state on every container change. A narrow viewport pinned the preview to the window margin, and returning to desktop grew it from that corner instead of restoring its centered layout.

## Impact

The landing-page hero preview now recenters after a desktop-to-mobile-to-desktop round trip while retaining intentional user positioning after a drag or resize.

## Validation

- `npm run build` in `frontend/src/landing`
- `npx tsc --noEmit` in `frontend/src/landing`
- `git diff --check`


codebanditssss (commented):
lgtm

codebanditssss (approved):

PR #3148, opened 2026-07-26
fix(frontend): remove hidden sidebar controls from tab order

### LINKED ISSUE #3147: bug(frontend): hidden collapsed-sidebar Settings button remains keyboard-focusable
## Description

When the sidebar is expanded, the Settings button from the collapsed sidebar rail remains in the keyboard tab order underneath the visible Settings row.

This creates a duplicate invisible focus stop for keyboard and screen-reader users.

## Steps to reproduce

1. Open the Electron app with the sidebar expanded.
2. Navigate to the visible **Settings** button using the keyboard.
3. Press `Tab` once.
4. Focus moves to the hidden Settings button belonging to the collapsed sidebar rail.

## Expected behavior

Only the currently visible Settings control should be focusable and exposed to assistive technologies.

## Actual behavior

The hidden rail button receives keyboard focus even though it is not the active sidebar variant.

Verified on the native Electron renderer from the latest `main`:

- hidden button has `tabIndex=0`
- `aria-hidden` is not set
- visible text is empty
- CSS has `pointer-events: none`
- pressing `Tab` from the visible Settings row focuses this hidden button

## Possible fix

Remove the inactive sidebar variant from the tab order and accessibility tree, for example with `tabIndex={-1}` and `aria-hidden="true"`, while ensuring those attributes switch correctly when the sidebar collapses. A focused keyboard-accessibility test should cover both expanded and collapsed states.

I am willing to work on this issue.

Hardik180704:
## What changed

- remove inactive sidebar footer controls from keyboard navigation
- hide the inactive expanded/collapsed footer variant from assistive technologies
- apply the same handling to the restart-to-update controls
- add regression coverage for both sidebar states

## Why

When the sidebar was expanded, pressing Tab from the visible Settings row focused the hidden collapsed-rail Settings button. The inactive control stayed at `tabIndex=0` even though it had `pointer-events: none`.

## Verification

- native Electron keyboard verification: Tab skips the hidden Settings control
- `npm run typecheck`
- `npm run typecheck:e2e`
- full frontend suite: 1,087 tests passed
- renderer smoke: 20 tests passed
- focused Sidebar suite: 41 tests passed

Closes #3147

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -230,6 +230,41 @@ afterEach(() => {
 });
 
 describe("Sidebar", () => {
+	it("keeps only the expanded Settings control keyboard-accessible while expanded", () => {
+		renderSidebar();
+
+		const settingsButtons = Array.from(document.querySelectorAll<HTMLButtonElement>('button[aria-label="Settings"]'));
+		const expandedButton = settingsButtons.find((button) => button.textContent?.includes("Settings"));
+		const collapsedButton = settingsButtons.find((button) => !button.textContent?.includes("Settings"));
+
+		expect(settingsButtons).toHaveLength(2);
+		expect(expandedButton).toHaveAttribute("tabindex", "0");
+		expect(expandedButton?.parentElement).not.toHaveAttribute("aria-hidden");
+		expect(collapsedButton).toHaveAttribute("tabindex", "-1");
+		expect(collapsedButton?.closest('[aria-hidden="true"]')).toBeInTheDocument();
+	});
+
+	it("keeps only the collapsed Settings control keyboard-accessible while collapsed", async () => {
+		renderSidebar();
+
+		fireEvent.pointerDown(screen.getByTestId("resize-handle"), { clientX: 240 });
+		fireEvent.pointerMove(window, { clientX: 120 });
+
+		await waitFor(() => {
+			expect(document.querySelector('[data-slot="sidebar"][data-state="collapsed"]')).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


Pulkit7070 (commented):
Reviewed and ran the Sidebar suite (41 tests pass, including the two new cases). The fix is correct and the a11y pattern is right.

- The hidden footer variant (expanded chrome when collapsed, or the icon rail when expanded) now gets aria-hidden plus tabIndex -1 on its controls, so exactly one Settings and one restart-to-update control is tab-reachable per state. That removes the reported bug where Tab from the visible Settings row landed on the hidden collapsed-rail button.
- aria-hidden is set as {isCollapsed || undefined} / {!isCollapsed || undefined}, so it renders aria-hidden="true" only when hidden and omits the attribute otherwise, avoiding aria-hidden="false". Good.
- Importantly this stays aria-hidden compliant: the hidden container has no focusable descendant, because the buttons inside it are set to tabIndex -1 in that state. Combined with the existing pointer-events: none, the hidden variant is fully inert.

One small coverage note, non-blocking: the new tests assert the two Settings buttons, but RestartToUpdateRow/RailButton only render when an update is downloaded, so their tabIndex is not exercised by a test. The logic is the same expression as the Settings buttons, so risk is low, but a case with a downloaded UpdateStatus would lock the restart controls too.

Hardik180704:
Thanks for reviewing and verifying the tests, Pulkit! Since the restart-to-update coverage note is non-blocking, I’ve kept this PR focused on the reported Settings regression. Could you please approve it when you get a chance?

Pulkit7070 (approved):

PR #3150, opened 2026-07-27
feat(command-palette): add PR and review workflow actions 

### LINKED ISSUE #3062: [Feature]: Add command palette actions for PR and review workflows
### Problem

The command palette currently includes:

- Current session navigation
- New task / new project
- Open orchestrator
- Project settings
- Copy branch name for the current session
- PR entries, but selecting a PR currently navigates to the session instead of opening/copying the PR directly

### Proposed direction

Add command palette actions for PR and review workflows:

- `Open PR`
  - Opens the selected PR URL externally.
  - Available for sessions with open PRs.

- `Copy PR URL`
  - Copies the PR URL to clipboard.
  - Useful for sharing review links.

- `Trigger review` / `Re-run review`
  - Starts a review for the current session or selected PR when eligible.
  - Should reuse the existing review trigger endpoint/UI behavior.

- Keep or improve `Copy branch name`
  - This already exists for the current session.
  - Consider making it easier to discover with keywords like `git`, `branch`, `copy`.

## Expected behavior

- Users can press `Ctrl+K` / `Cmd+K`, type `open pr`, `copy pr`, or `review`, and run the action directly.
- Disabled states explain why an action is unavailable, for example:
  - no open PR
  - review already running
  - session is not eligible for review

## Suggested implementation notes

- Extend `CommandAction` in `frontend/src/renderer/lib/command-palette.ts`.
- Add command generation for PR/review actions in `buildCommands`.
- Handle new action kinds in `frontend/src/renderer/components/CommandPalette.tsx`.
- Reuse existing helpers:
  - `aoBridge.clipboard.writeText`
  - `aoBridge.app.openExternal`
  - existing review API endpoints/hooks where available

## Validation

- Add tests in `frontend/src/renderer/lib/command-palette.test.ts`.
- Add action execution tests in `frontend/src/renderer/components/CommandPalette.test.tsx`.

### Before submitting

anmol0b:
## What

Adds command palette actions for PR and review workflows (frontend-only): per-PR `Open PR #N`, `Copy PR URL #N`, and `Run review`/`Re-run review #N` items, plus the shared review-data plumbing that lets the palette show live review eligibility for any session with open PRs.

## Why

Fixes #3062. Today the palette's PR entries only navigate to the session — acting on a PR (opening it in the browser, sharing its URL, triggering an AO review) means leaving the keyboard flow and clicking through the session inspector. This makes the three most common PR/review actions available directly from `Cmd/Ctrl+K`, with disabled states that explain why an action is unavailable.

## How

- New `CommandAction` variants (`open-pr`, `copy-pr-url`, `trigger-review`) in `command-palette.ts`, with three new items emitted per open PR in `buildCommands`. All three are `searchOnly` so the default palette view is unchanged; keywords make `open pr` / `copy pr` / `review` queries surface them.
- Extracted `sessionReviewsQueryOptions` (the `["session-reviews", session.id]` query, including the preview mock) and the eligibility helpers (`openReviewStatesFor`, `reviewIsRunning`, `reviewRunDisabled`, `reviewSessionRunAction`) from `SessionInspector.tsx` into a new `lib/session-reviews.ts`. The palette and the Reviews tab now share one fetch path, one React Query cache key, and one copy of the disabled-state logic — no duplicated eligibility math to drift.
- `CommandPalette.tsx` fetches review states via `useQueries` for sessions with open PRs **only while the palette is open**; sessions already viewed in the inspector reuse cached data. Unloaded/failed review fetches leave the action enabled — the daemon validates on trigger and the palette's existing error banner surfaces the response.
- Disabled states mirror ReviewPanel semantics: session review running → "Review already running"; ineligible or draft PR → "Not eligible for review". The trigger is deliberately session-scoped (same endpoint as the Reviews tab button — one run covers all of a session's eligible open PRs), so multi-PR sessions show one row per PR, all firing the same session trigger with per-PR-accurate labels.
- Trigger success invalidates `["session-reviews", id]` + the workspace query, same as ReviewsView. Also adds `copy` to the Copy branch name keywords (the issue's discoverability ask).

Intentional omissions: no palette `Cancel review` action — the issue specifies a disabled "review already running" state, so cancel stays in the Reviews tab (possible fast-follow). No per-action icons — new items use the existing `prs` group fallback (`GitPullRequest`). No backend/API changes: the trigger/copy/open actions reuse the existing endpoints and bridges as the issue suggested.

## Testing

Frontend:

- `npx vitest run src/renderer/lib/command-palette.test.ts src/renderer/components/CommandPalette.test.tsx src/renderer/components/SessionInspector.test.tsx` — 102 tests pass. New coverage: 9 `buildCommands` tests (per-PR action items + URLs, Run vs Re-run labels, "Review already running" / "Not eligible for review" disabled states, enabled-while-unloaded fallback, multi-PR sessions, merged/closed exclusion, `open pr`/`copy pr`/`review` search discoverability, search-only hiding, `copy` keyword on Copy branch) and 6 `CommandPalette` action tests (Open PR fires `openExternal` + closes, Copy PR URL writes clipboard + closes, Run review POSTs the trigger endpoint with the session id + closes, trigger failure keeps the palette open with the daemon error in the alert banner, disabled running/ineligible items never POST).
- Full renderer suite `npm test` — 1100 passed (87 files), confirming the SessionInspector extraction is behavior-preserving.
- `npm run typecheck` and `npm run typecheck:e2e` — clean.

Screenshots (browser preview, `VITE_NO_ELECTRON=1`, mock data; palette opened with `Cmd+K`):

Before:

03-query-open-pr
04-query-copy-pr
05-query-review

After:

03-query-open-pr
04-query-copy-pr
05-query-review
06-query-re-run-review

Highlights: `open pr` goes from plain `#N` navigate rows to 11 direct `Open PR #N` actions; `copy pr` goes from one accidental title match to 11 `Copy PR URL #N` actions; `review` goes from session-name matches only to per-PR `Run review`/`Re-run review` items with "Not eligible for review" shown on the draft PR.

## Checklist

- [x] Relevant CI checks pass for the area touched (frontend vitest + typecheck)


illegalcall (commented):
Reviewed on a clean worktree at `8871640dd`.

**Verdict: approve with changes.** Best-structured PR of the current batch. Extracting `session-reviews.ts` is a real win — it kills the duplicated `mockReviewsResponse` / `mockReviewTitle` and the eligibility logic that had drifted into `SessionInspector`, and giving the query key one owner is the right call.

Three things I'd want addressed before merge, all in the palette wiring rather than the extraction.

## 1. Fan-out fetch on every palette open (medium)

`useQueries` issues one `GET /api/v1/sessions/{id}/reviews` per session with an open-or-draft PR, **across all projects**, gated only on `isOpen`. The global `staleTime` is 10s (`lib/query-client.ts`), so any palette open more than 10s after the last one refires all of them.

On a busy workspace with 30+ sessions carrying open PRs that's a 30-request burst on ⌘K — on the one surface where latency is most visible to the user. The comment says "sessions already viewed in the inspector reuse the cached data", which is true but only covers the sessions you happen to have opened recently.

Suggest a per-query `staleTime` well above the global (60s+ — review state doesn't move fast), or deferring the fetch until the typed query actually matches a review command.

## 2. The memoization is defeated (medium)

`useQueries` without `combine` returns a **fresh array identity every render**. So:

`reviewQueries` (new array) → `reviewStatesBySessionId` useMemo recomputes → `items` useMemo recomputes → `buildCommands` walks every workspace and session **on every keystroke**.

Both `useMemo`s are currently no-ops. Move the Map construction into `useQueries({ queries, combine })` — that's exactly what `combine` exists for, and it restores the memoization you clearly intended.

## 3. Items mutate under the cursor after open (medium, UX)

On open, `reviewStatesBySessionId` is empty, so every review item renders enabled as `Run review #N`. When the fetches land they can flip to `Re-run review #N`, `Reviewing... #N`, or disabled with `Not eligible for review`.

Because the **title** changes (so the cmdk value changes) and `resolvedValue` at `CommandPalette.tsx:99-101` re-picks the first non-disabled item when the selected one becomes disabled, the highlighted row can jump between opening the palette and pressing Enter. Your own test — "disables the review action with Review already running **once the session review state loads**" — documents exactly this transition.

Worth rendering a stable label until the data lands, so the list doesn't reshuffle under someone who's already typing.

## 4. "Not eligible for review" is per-PR but the action is per-session (low)

`prReviewCommand` marks a PR ineligible when *its own* state is missing or `ineligible`:

```ts
const ineligible = openReviewStates !== undefined && (!prReviewState || prReviewState.status === "ineligible");
```

whereas `reviewRunDisabled` (the Reviews tab) only disables when **every** open PR is ineligible. So in a stacked-PR session where #1 is eligible and #2 isn't, the tab's button is enabled while the palette's "#2" item reads "Not eligible for review" — even though triggering would legitimately run and cover #1.

The doc comment acknowledges the session scoping, but the user-visible string doesn't. Either make the reason say so, or fall back to the session-level rule.

## 5. Draft PRs get the right answer via the wrong path (low)

`openPRs()` includes `state: "draft"` (`types/workspace.ts:206`), but `openReviewStatesFor` filters strictly on `pr.state === "open"`. So a draft's review item ends up disabled because it was **excluded from the open set**, not because its status is `ineligible`. Right outcome today, but fragile if `openReviewStatesFor` ever widens to include drafts — the disable would silently vanish.

Also worth a deliberate decision on whether drafts should get Open / Copy items at all. They currently do.

## 6. Terminated sessions get an enabled "Run review" (low)

`workerSessions()` filters orchestrators, not `sessionIsActive`. The nav item already had this shape, but this is the first **mutating** action offered for a dead session. It'll fail server-side via `writeReviewError` and surface in the banner, so it's noise rather than danger — but it's avoidable noise.

## 7. Unrelated keyword change (nit)

Adding `"copy"` to the Copy-branch keywords (`command-palette.ts:163`) is a genuine improvement with its own test, but it's unrelated to PR/review actions — scope creep against AGENTS.md's "keep every change surgical".

## Notes

Going 1 → 4 items per open PR is fine: three are `searchOnly`, so the empty-query palette is unchanged, and `MAX_SEARCH_RESULTS` caps the rest. Good call.

Tests are strong on both levels — the pure builder and the wired component, including the daemon-error path and both disabled reasons.

illegalcall (approved):
Approving — the extraction into `session-reviews.ts` is a genuine improvement and the behavior is right.

Please still land these three before merge; none changes the design:

1. `useQueries({ queries, combine })` — without `combine` the result array has a fresh identity every render, so both `useMemo`s are no-ops and `buildCommands` re-walks every workspace on each keystroke.
2. A per-query `staleTime` (60s+) — the global is 10s, so a busy workspace re-fires one request per session-with-open-PR on every ⌘K.
3. A stable label until review state loads, so items don't retitle and re-sort under the cursor between opening the palette and pressing Enter.

Details and the four lower-severity items are in my review above.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
From the NotesBot call on `2026-07-27`, this existing pr appears to be relevant.
Match confidence: `0.63` (fuzzy).

Relevant discussion excerpts:
- - **Merge button in Kanban** to merge PRs directly - **Workflow actions** via Command Palette (“Command K”) to take actions on PRs/issues - Investigating a bug raised by Nikhil: review worker receiving “**gibberish content**” when reviewing from GitHub; aditi suspects current flow only uses SCM and doesn’t inject [...]
- - **Workflow actions** via Command Palette (“Command K”) to take actions on PRs/issues - Investigating a bug raised by Nikhil: review worker receiving “**gibberish content**” when reviewing from GitHub; aditi suspects current flow only uses SCM and doesn’t inject GitHub content; investigating source.
- - **Workflow actions** via Command Palette (“Command K”) to take actions on PRs/issues - Investigating a bug raised by Nikhil: review worker receiving “**gibberish content**” when reviewing from GitHub; aditi suspects current flow only uses SCM and doesn’t inject GitHub content; investigating source. - prateek [...]

Source thread: https://discord.com/channels/1476302178913357958/1531352924586774700/1531352924586774700

anmol0b:
> Approving — the extraction into `session-reviews.ts` is a genuine improvement and the behavior is right.
> 
> Please still land these three before merge; none changes the design:
> 
> 1. `useQueries({ queries, combine })` — without `combine` the result array has a fresh identity every render, so both `useMemo`s are no-ops and `buildCommands` re-walks every workspace on each keystroke.
> 2. A per-query `staleTime` (60s+) — the global is 10s, so a busy workspace re-fires one request per session-with-open-PR on every ⌘K.
> 3. A stable label until review state loads, so items don't retitle and re-sort under the cursor between opening the palette and pressing Enter.
> 
> Details and the four lower-severity items are in my review above.



@illegalcall Addressed all three must-fix items in c6b34c69:

useQueries({ queries, combine }), returning a plain Record (not Map) so structural sharing keeps identity stable
Per-query staleTime: 60_000 on the palette's review queries (Reviews tab keeps the global 10s)
Freeze-on-open snapshot so items don't retitle/reshuffle under the cursor mid-session

Also resolved the lower-severity items: kept per-PR disabled semantics, drafts keep Open/Copy, terminated sessions lose only the review trigger (not Open/Copy), kept the copy keyword (it's explicitly called out in #3062).

Merged latest main in 86614cdb to resolve conflicts in CommandPalette.tsx/SessionInspector.tsx — all tests passing (104/104), typecheck clean.

Ready for another look whenever you get a chance.

illegalcall (changes requested):
Re-reviewed `86614cdb8`. The focused 57 tests and frontend typecheck pass, and all three CI checks are green.

Keeping this at request-changes because the third must-fix from the prior review is still present.

The new snapshot is not actually frozen for the lifetime of an open palette. On the first open with pending queries, the effect snapshots the partial/empty result and sets `openSnapshotSettledRef.current = false`; when the queries settle, the later branch calls `setReviewStatesSnapshot(...)` while `isOpen` is still true. That still retitles/disables the row under the cursor. The existing test at `CommandPalette.test.tsx:547` proves the transition: it opens the palette, waits for review state, then expects `Run review #7` to have become `Reviewing... #7` with `Review already running` during the same open.

Please either keep the opening snapshot unchanged until close/reopen, or preserve the visible command identity/title/selection while live state arrives. Add a deferred-query regression test that records the row while open, resolves the query, and asserts it does not move or retitle until the next open.

The other two must-fixes are resolved cleanly: `useQueries({ combine })` restores stable aggregation and the palette queries have a 60s `staleTime`. Removing review triggers for terminated sessions is also good.

Cleanup before the next pass: remove the accidental `// new`, restore the surrounding tab indentation, and remove the trailing whitespace in `command-palette.test.ts:270`.

anmol0b:
> Re-reviewed `86614cdb8`. The focused 57 tests and frontend typecheck pass, and all three CI checks are green.
> 
> Keeping this at request-changes because the third must-fix from the prior review is still present.
> 
> The new snapshot is not actually frozen for the lifetime of an open palette. On the first open with pending queries, the effect snapshots the partial/empty result and sets `openSnapshotSettledRef.current = false`; when the queries settle, the later branch calls `setReviewStatesSnapshot(...)` while `isOpen` is still true. That still retitles/disables the row under the cursor. The existing test at `CommandPalette.test.tsx:547` proves the transition: it opens the palette, waits for review state, then expects `Run review #7` to have become `Reviewing... #7` with `Review already running` during the same open.
> 
> Please either keep the opening snapshot unchanged until close/reopen, or preserve the visible command identity/title/selection while live state arrives. Add a deferred-query regression test that records the row while open, resolves the query, and asserts it does not move or retitle until the next open.
> 
> The other two must-fixes are resolved cleanly: `useQueries({ combine })` restores stable aggregation and the palette queries have a 60s `staleTime`. Removing review triggers for terminated sessions is also good.
> 
> Cleanup before the next pass: remove the accidental `// new`, restore the surrounding tab indentation, and remove the trailing whitespace in `command-palette.test.ts:270`.

Fixed — the snapshot is now a true freeze: captured once on open, never updated again until close/reopen. Removed the "first settle" branch that was still applying live data mid-session.

Also addressed:

- Added the regression test you asked for: opens the palette, lets the review query resolve in the background, and asserts the row does not retitle or disable during that same open.
- Updated the two existing tests that relied on the old mid-session flip — they now close and reopen the palette to observe the resolved state, since that's the only point it's allowed to change.
- Removed the stray // new comment, fixed the indentation, and trimmed the trailing whitespace at command-palette.test.ts:270.

55/55 tests pass, typecheck clean.

One tradeoff worth flagging: on a session's very first open with nothing cached, review items will show the generic enabled "Run review" label for that entire open, even after the fetch resolves — disabled reasons only appear starting the next open. Given the 60s staleTime and the shared cache with the Reviews tab, this should be rare in practice, and the daemon still validates on trigger either way so nothing incorrect can execute. Happy to revisit if you'd rather preserve live disabled-state updates without moving the row (more invasive — would need to decouple cmdk's selection tracking from title changes).

frontend/src/renderer/components/CommandPalette.tsx
```diff
@@ -56,15 +62,56 @@ export function CommandPalette() {
...
+	// already fresh).
+	useEffect(() => {
+		if (!isOpen) {
+			wasOpenRef.current = false;
+			setReviewStatesSnapshot(undefined);
+			return;
+		}
+		if (!wasOpenRef.current) {
+			wasOpenRef.current = true;
+			setReviewStatesSnapshot(reviewQuerySummary.reviewStatesBySessionId);
```
illegalcall: Blocking UX/spec gap: on a cold first open, this captures the empty pre-fetch result and then intentionally never updates it. A session whose review is already running or ineligible therefore shows an enabled generic Run review action for the entire open; the new test at CommandPalette.test.tsx:572 codifies that stale enabled state. The daemon rejects the bad trigger, but #3062 requires the action to be disabled with the reason. Preserve stable command identity/title/order while allowing eligibility/disabled state to update, or wait for the initial review state before exposing the mutating action.

illegalcall (changes requested):
Re-reviewed current head d4c0396. The useQueries combine, 60-second stale time, true no-retitle snapshot, terminated-session action removal, focused tests, typecheck, and all three CI checks are green. The true freeze solves cursor jumping, but it creates a cold-open correctness gap that still blocks approval.

anmol0b:
> Re-reviewed current head [d4c0396](https://github.com/Untrivial-ai/agent-orchestrator/commit/d4c039630b149907e40f991ff0653707053cdb0c). The useQueries combine, 60-second stale time, true no-retitle snapshot, terminated-session action removal, focused tests, typecheck, and all three CI checks are green. The true freeze solves cursor jumping, but it creates a cold-open correctness gap that still blocks approval.

Fixed per your suggestion — went with "wait for the initial review state before exposing the mutating action" rather than decoupling selection from title.

- The review action is now omitted entirely until we actually know the session's eligibility (buildCommands gates on whether reviewStatesBySessionId has resolved for that session yet). Open PR / Copy PR URL still render immediately since they don't depend on review state.
- Once a session's state resolves, it's captured and frozen for the rest of that open — merging in newly-resolved sessions as they land, but never overwriting one already known, so a later poll (e.g. a running review completing) still can't retitle or reorder a row.
- Added a test that exercises the full sequence: open with a pending fetch → confirm the review row is absent while Open PR/Copy PR URL are present → resolve the fetch → confirm the row appears correctly disabled → simulate a later poll resolving to a different state → confirm the row does not change again within the same open.

frontend/src/renderer/components/CommandPalette.tsx
```diff
@@ -56,15 +61,62 @@ export function CommandPalette() {
...
+	// key means sessions already viewed in the inspector reuse the cached data.
+	const reviewQuerySummary = useQueries({
+		queries: sessionsWithOpenPRs.map((session) =>
+			sessionReviewsQueryOptions(session, isOpen, PALETTE_REVIEW_STALE_TIME_MS),
+		),
+		combine: (results) => {
+			const reviewStatesBySessionId: Record<string, PRReviewState[]> = {};
+			results.forEach((result, index) => {
+				const session = sessionsWithOpenPRs[index];
+				if (session && result.data) reviewStatesBySessionId[session.id] = result.data.reviews ?? [];
```
illegalcall: Blocking: this fixes the cold empty-cache case, but stale cached data still enters the frozen snapshot. React Query may expose result.data immediately from cache while a refetch is in progress; this combine includes it without checking freshness/isFetching, and the effect below never overwrites a session once captured. An old eligible value can therefore expose Run review even when the fresh state is running or ineligible. Gate snapshot capture until the fresh fetch settles (or represent freshness explicitly), and add a regression test that seeds stale eligible cache data and completes a refetch with an ineligible state.

illegalcall (changes requested):
Re-reviewed current head 561123398271a7e77e19e8c4c03ebeb256f82017 after the latest author update. The originally reported cold/empty-cache issue is fixed, and that resolved thread should remain resolved.

A separate stale-cache path remains blocking: React Query can synchronously return old review data while it refetches, the combine step treats that old data as resolved, and the per-open snapshot then refuses to replace it. See the inline finding. This can expose an enabled Run review command even though the fresh state is running or otherwise ineligible.

Verification: all 3 GitHub checks are green; frontend typecheck and the two focused Vitest files passed (56 tests). The current tests do not simulate a stale cached value followed by an actual refetch.

Non-blocking follow-ups: make the review-state loading contract explicit instead of overloading an omitted map, update the comments that still say unloaded data renders enabled, and align the per-PR Run review #N label with the session-wide trigger endpoint. The issue also asks for a disabled explanation when no PR is available.

anmol0b:
> Re-reviewed current head [5611233](https://github.com/Untrivial-ai/agent-orchestrator/commit/561123398271a7e77e19e8c4c03ebeb256f82017) after the latest author update. The originally reported cold/empty-cache issue is fixed, and that resolved thread should remain resolved.
> 
> A separate stale-cache path remains blocking: React Query can synchronously return old review data while it refetches, the combine step treats that old data as resolved, and the per-open snapshot then refuses to replace it. See the inline finding. This can expose an enabled Run review command even though the fresh state is running or otherwise ineligible.
> 
> Verification: all 3 GitHub checks are green; frontend typecheck and the two focused Vitest files passed (56 tests). The current tests do not simulate a stale cached value followed by an actual refetch.
> 
> Non-blocking follow-ups: make the review-state loading contract explicit instead of overloading an omitted map, update the comments that still say unloaded data renders enabled, and align the per-PR Run review #N label with the session-wide trigger endpoint. The issue also asks for a disabled explanation when no PR is available.

Fixed — thanks for catching this, it was a real gap. The combine step now excludes a session's data while result.isFetching is true, so a stale cached value returned synchronously during a background refetch is no longer treated as resolved. Only genuinely settled data enters the frozen snapshot now.

Added a regression test that seeds a stale eligible cache entry, opens the palette (triggering a background refetch), confirms the mutating action stays hidden while that refetch is in flight, then resolves it to a running/ineligible state and confirms the row reflects the fresh data correctly.

57/57 tests pass, typecheck clean.

On the four non-blocking follow-ups (explicit loading contract instead of overloading the omitted-map case, stale comments, per-PR vs session-wide label alignment, disabled explanation when no PR exists) — happy to take these as a fast-follow if you'd like them split out, or address them here if you'd rather keep it in one PR.

frontend/src/renderer/components/CommandPalette.tsx
```diff
@@ -56,15 +61,64 @@ export function CommandPalette() {
...
+	// key means sessions already viewed in the inspector reuse the cached data.
+	const reviewQuerySummary = useQueries({
+		queries: sessionsWithOpenPRs.map((session) =>
+			sessionReviewsQueryOptions(session, isOpen, PALETTE_REVIEW_STALE_TIME_MS),
+		),
+		combine: (results) => {
+			const reviewStatesBySessionId: Record<string, PRReviewState[]> = {};
+			results.forEach((result, index) => {
+				const session = sessionsWithOpenPRs[index];
+				if (session && result.data && !result.isFetching) {
```
illegalcall: Blocking: !isFetching handles an in-flight refetch, but not a failed one. After a background refetch errors, TanStack Query retains the stale result.data and sets isFetching=false, so this condition admits that stale eligible state and the effect freezes it. I reproduced the behavior with stale needs_review cache data plus a failed GET: Run review #7 reappears. Require a successful/non-error revalidation before snapshotting (for example also exclude result.isError), and add the failed-refetch regression alongside the success-path test.

illegalcall (changes requested):
Re-reviewed updated head 099f49042ce22eb35cd20032c8c194a298c94a7b. The author correctly fixed the successful background-refetch case: stale data is excluded while isFetching is true, and the new regression covers fresh running data replacing stale needs_review. The prior thread is correctly resolved. Frontend typecheck, both focused Vitest files (57 tests), and all 3 CI checks pass.

One adjacent stale-cache path still blocks approval: when that background refetch fails, TanStack Query preserves the cached data and transitions isFetching to false. The combine condition then admits the stale eligible value into the frozen snapshot. I verified this with a temporary diagnostic test that seeded stale needs_review data, made the refetch return an API error, and observed Run review #7 render after the failed refetch. See the inline finding.

Non-blocking follow-ups remain: make loading/freshness explicit instead of overloading an optional map, update stale comments, clarify the per-PR label for a session-wide action, provide the requested no-open-PR explanation, and clean the trailing whitespace/no-final-newline reported by git diff --check.

anmol0b:
> Re-reviewed updated head [099f490](https://github.com/Untrivial-ai/agent-orchestrator/commit/099f49042ce22eb35cd20032c8c194a298c94a7b). The author correctly fixed the successful background-refetch case: stale data is excluded while isFetching is true, and the new regression covers fresh running data replacing stale needs_review. The prior thread is correctly resolved. Frontend typecheck, both focused Vitest files (57 tests), and all 3 CI checks pass.
> 
> One adjacent stale-cache path still blocks approval: when that background refetch fails, TanStack Query preserves the cached data and transitions isFetching to false. The combine condition then admits the stale eligible value into the frozen snapshot. I verified this with a temporary diagnostic test that seeded stale needs_review data, made the refetch return an API error, and observed Run review #7 render after the failed refetch. See the inline finding.
> 
> Non-blocking follow-ups remain: make loading/freshness explicit instead of overloading an optional map, update stale comments, clarify the per-PR label for a session-wide action, provide the requested no-open-PR explanation, and clean the trailing whitespace/no-final-newline reported by git diff --check.

Fixed — thanks, this was the right next gap to close. Added !result.isError alongside !result.isFetching, so a session only enters the snapshot once its data has arrived from a fetch that neither is in-flight nor ended in error. A failed background refetch on stale cached data no longer gets treated as resolved.

Added a regression test alongside the success-path one: seeds stale eligible cache data, opens the palette (triggering a background refetch), rejects that refetch, and confirms the row stays hidden rather than reappearing with the stale "Run review" state.

58/58 tests pass, typecheck clean.

Given how many edge cases this frozen-snapshot approach keeps surfacing, happy to also tackle the "make loading/freshness explicit" follow-up now rather than later if you'd prefer — it might close off this whole class of stale-vs-fresh ambiguity at the root instead of us finding cases one at a time.

achalbajpai:
Can we nest these three actions under their corresponding PR, similar to the “Needs attention” flow? This would make the hierarchy clearer and keep the palette consistent.  @anmol0b

anmol0b:
> Can we nest these three actions under their corresponding PR, similar to the “Needs attention” flow? This would make the hierarchy clearer and keep the palette consistent. @anmol0b

Good catch on the inconsistency. The flat rows are intentional, though: #3062 specifically asks for open pr / copy pr / review to fire directly from the typed query, and nesting would add a drill-in step to exactly those flows. They're also searchOnly, so they don't clutter the default palette. Happy to add a drill-in on the #42 row in addition (keeping the direct rows) as a follow-up; 

lmk if you'd like me to open a separate issue for that so we can keep this PR focused.

illegalcall:
@achalbajpai Wdyt?

achalbajpai:
I was thinking like how big of a diff for me would be if i had to go inside the nested way
we are having copy ( which is okay , very few times we need to copy a pr number ) , rerun review ( sometime would be there ) and open pr ( the most freq one in this case ) @anmol0b

anmol0b:
> I was thinking like how big of a diff for me would be if i had to go inside the nested way we are having copy ( which is okay , very few times we need to copy a pr number ) , rerun review ( sometime would be there ) and open pr ( the most freq one in this case ) @anmol0b

not as big as this pr, but not tiny either—probably a few hundred lines. it'd involve adding the drill-in view, `buildPRActions`, back navigation, and updating the pr/review tests.

and yeah, i think your read is right. the only action that really gets worse with nesting is open pr—it goes from `type → enter` to `type → enter → enter`. copy and rerun are already one keystroke, so nesting doesn't really improve those.

i'd leave it as-is. the flat rows with `searchOnly` already feel clean and keep everything one keystroke away. if we decide the hierarchy is worth it later, we can always make #42 the drill-in point while keeping the rows flat, so there's no regression.

anmol0b:
@achalbajpai @illegalcall could you take a look at this? if everything looks good, i'll resolve the conflicts and we can merge it.

illegalcall (approved):

Issue #3034, opened 2026-07-23
macOS auto-update fails: ditto errors during ShipIt install (daemon/ao + AppleDouble files missing)


i-trytoohard:
## Bug Report: macOS nightly auto-update fails with `ditto` errors

**Reported by:** @prateek (Discord #bug-triage)
**Affected version:** `v0.10.4-nightly.202607221413` (updating TO this nightly on macOS)
**Platform:** macOS (username `itrytoohard`)

### Error

When updating to the latest nightly via in-app auto-update, the update downloads but fails during the **Squirrel.Mac `ShipIt` installation step**. The UI shows:

```
ditto: ~/Library/Caches/dev.agent-orchestrator.desktop.ShipIt/update.SdDd11x/Agent Orchestrator.app/Contents/Resources/daemon/ao: No such file or directory
ditto: ~/Library/Caches/dev.agent-orchestrator.desktop.ShipIt/update.SdDd11x/Agent Orchestrator.app/Contents/Resources/._icon.png__: No such file or directory
ditto: ~/Library/Caches/dev.agent-orchestrator.desktop.ShipIt/update.SdDd11x/Agent Orchestrator.app/Contents/Resources/._electron.icns__: No such file or directory
ditto: ~/Library/Caches/dev.agent-orchestrator.desktop.ShipIt/update.SdDd11x/Agent Orchestrator.app/Contents/Resources/._app.asar__: No such file or directory
ditto: ~/Library/Caches/dev.agent-orchestrator.desktop.ShipIt/update.SdDd11x/Agent Orchestrator.app/Contents/Resources/._icon.ico__: No such file or directory
```

### Analysis

**The published nightly zip is correct.** I downloaded both `darwin-arm64` and `darwin-x64` zips from the `v0.10.4-nightly.202607221413` release and verified:

- ✅ `Contents/Resources/daemon/ao` is present (22.9 MB arm64 / 24.5 MB x64)
- ✅ `Contents/Resources/icon.png`, `icon.ico`, `app.asar`, `electron.icns` all present
- ✅ `app-update.yml` present
- ✅ Squirrel.framework with `ShipIt` binary bundled

**The problem is in the update extraction/installation step, not the published artifact.** The `ditto` errors originate from Squirrel.Mac's `ShipIt` process trying to copy from the **extracted update cache** (`update.SdDd11x/`), where files that ARE in the published zip are reported as missing.

### Root cause hypothesis

[...]

PR #3151, opened 2026-07-27
fix(release): skip blockmap sidecar for macOS zips

anmol0b:
## What
Short-term fix for #3034: macOS zips no longer get a `.blockmap` sidecar
file generated or referenced in the feed yml.

## Why
Squirrel.Mac's ShipIt install step runs `ditto` against the extracted
update cache, which fails looking for AppleDouble (`._*`) metadata files
that `@electron-forge/maker-zip`'s output never contains (unlike
`electron-builder`'s ditto-based zips). A blockmap-driven differential
update against that mismatched zip format is the likely corruption path
causing the reported `ditto: ... No such file or directory` errors.

## How
- Added `hashFile` — computes the same `{sha512, size}` shape
  `writeBlockmap` returns, but via plain `node:crypto` hashing, with no
  `.blockmap` sidecar written to disk.
- `generateFeeds` now uses `hashFile` for mac zips specifically and keeps
  `writeBlockmap` unchanged for win/linux.
- Since no sidecar exists for mac, `electron-updater`'s `MacUpdater` will
  always fall back to a full zip download instead of attempting a
  blockmap diff — this doesn't fix the underlying zip-format mismatch, but
  removes the specific failure path from this bug report.

## Scope
This is the **short-term** fix the issue proposes. The **long-term** fix —
generating mac zips via `ditto` (matching `electron-builder`'s format,
preserving AppleDouble metadata) — is a bigger change to the packaging
pipeline (`forge.config.ts` / a custom maker or post-package hook) that I
haven't attempted here, since it needs real macOS build/sign/publish
verification to confirm it actually resolves auto-update end-to-end. This
PR only removes the blockmap-sidecar half of the failure; tracking the
`ditto`-based packaging fix as a separate follow-up.

## Testing
- `npx vitest run scripts/feed.test.mjs` — 19 tests pass, including new
  coverage: `hashFile` unit tests (correct sha512/size, no sidecar
  written) and `generateFeeds` regression tests confirming mac skips
  `writeBlockmap` entirely while win/linux are unchanged.
- Full frontend suite `npx vitest run` — 1089 tests pass across 87 files,
  no regressions.
- `grep` confirms no other script/build step depends on `writeBlockmap`
  running unconditionally for every platform.

**Important caveat**: these tests confirm the *logic* — no blockmap is
produced/referenced for mac — but not that a real macOS auto-update now
succeeds end-to-end. That requires a signed build published through the
real release pipeline, which I don't have access to. Flagging for
@illegalcall / @Pulkit7070 (already assigned to #3034) to confirm on an
actual nightly build before considering this fully verified in
production.

Fixes the short-term portion of #3034.

## Checklist


illegalcall (commented):
Reviewed on a clean worktree at `bc79a38d0`.

**Verdict: approve after the newline and doc fixes.** The premise holds up.

## Verified

I checked the mechanism against [electron-updater's `MacUpdater`](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/MacUpdater.ts). It **does** attempt a differential download:

```js
if (canDifferentialDownload()) {
  differentialDownloadFailed = await this.differentialDownloadInstaller(...)
}
if (differentialDownloadFailed) {
  await this.httpExecutor.download(zipFileInfo.url, destinationFile, downloadOptions)
}
```

Combined with `blockmap.mjs`'s existing note that omitting `blockMapSize` forces the sidecar path, mac was on differential and removing the sidecar makes `differentialDownloadInstaller` fail → clean fallback to the full zip. So the fix does what it claims. Worth noting the fallback costs one 404 per update check before it kicks in; harmless, but it'll show up in updater logs.

## 1. Both files are missing a trailing newline (blocking nit)

`\ No newline at end of file` on both `feed.mjs` and `feed.test.mjs`.

## 2. Accurate documentation was deleted (low)

`buildYml`'s comment lost two sentences that are still true of the code directly beneath it:

- the deprecated top-level `path` / `sha512` pointing at `files[0]` — still emitted at `feed.mjs:51-52`
- `important: true` escalating the in-app prompt — still emitted at `feed.mjs:54`

Neither relates to this change. Please restore them. The blank line between `generateFeeds` and the CLI block was also removed — formatting churn AGENTS.md asks to avoid.

## 3. Sibling comment is now stale (low)

`blockmap.mjs:13-16` still says omitting `blockMapSize` "forces the client onto the sidecar differential path on **every platform** (verified against MacUpdater / NsisUpdater / AppImage)". That's the exact statement this PR invalidates for mac. Since it names MacUpdater specifically, it'll mislead the next person.

## 4. `hashFile` reads the entire zip into memory (low)

`readFileSync` on a mac app zip is a few hundred MB in one Buffer; `buildBlockMap` streams. `createReadStream` piped into the hash is three lines and removes an avoidable spike (plus Node's max-Buffer ceiling as a hard limit). Not urgent on current runners, just unnecessary.

## 5. Already-published mac blockmaps aren't removed, and a job re-run re-uploads them (low)

`frontend-release.yml:404-409` does:

```bash
gh release download "$tag" --dir dist --clobber
node scripts/feed.mjs dist "${tag#v}" latest
assets=(dist/latest*.yml dist/*.blockmap)
gh release upload "$tag" "${assets[@]}" --clobber
```

So re-running the feed job against a tag that already carries a mac sidecar pulls it back into `dist/` and re-publishes it. Fresh releases are fine (nothing to download), so this only bites on re-runs — but it's the exact scenario you'd hit while remediating #3034. An `rm -f dist/*darwin*.blockmap` before the glob would make the fix stick.

## 6. The root cause is still a hypothesis

The comment says "the **likely** corruption source", and the PR is honest about being short-term — no objection to that framing. But since the entire value here is whether mac updates now succeed, it'd be worth one end-to-end mac update against a build with no sidecar before merging. Right now the change is reasoned but unobserved.

## Tests

Good. Mocking `writeBlockmap` and asserting both directions — "not called for mac, no sidecar on disk" and "still called twice for win+linux with the right paths" — is the right pair, and exporting `generateFeeds` just for that is a fair trade.

illegalcall (approved):
Approving — I verified the mechanism against electron-updater's `MacUpdater`: it does attempt a differential download, and removing the sidecar makes it fail into the full-zip path cleanly. The fix does what it claims.

Please fix before merge:

1. Trailing newline on both files (`\ No newline at end of file` on `feed.mjs` and `feed.test.mjs`).
2. Restore the two deleted `buildYml` doc sentences — the deprecated top-level `path`/`sha512` and the `important: true` behavior are both still true of the code right below the comment.
3. `blockmap.mjs:13-16` still claims the sidecar differential path holds "on every platform (verified against MacUpdater …)" — now false for mac, and it names MacUpdater specifically.

The `readFileSync` memory point and the `gh release download` re-upload path are in my review above; both are fine to defer. Worth one real mac update test before this goes out, since the root cause is still stated as "likely".

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

anmol0b:
@illegalcall  fixed the pre-merge items.

  Updated this branch to:

  - add trailing newlines to `frontend/scripts/feed.mjs` and `frontend/scripts/feed.test.mjs`
  - restore the `buildYml` documentation for deprecated top-level `path`/`sha512` and `important: true`
  - update the stale `blockmap.mjs` comment so it no longer says the sidecar differential path applies to MacUpdater
  - restore the blank line before the CLI block

  Re-ran:

  `npx vitest run scripts/feed.test.mjs scripts/blockmap.test.mjs -v`

  Result: 2 files passed, 20 tests passed.

illegalcall (approved):
Re-reviewed `1bc5193cb`. All four items are fixed, and the fixes are clean:

- trailing newlines on `feed.mjs` and `feed.test.mjs` ✅
- `buildYml` doc restored — both the deprecated top-level `path`/`sha512` sentence and the `important: true` one ✅
- blank line before the CLI block restored ✅
- `blockmap.mjs` comment corrected ✅ — and better than I asked for: it now scopes the claim to win/linux, names the mac exception, and points at `hashFile`, so the two files explain each other instead of contradicting.

Nothing new introduced; the diff is exactly the requested scope. Approval stands.

The two deferred items from my first pass are still open and still fine to defer — `hashFile`'s `readFileSync` loading the whole zip into one Buffer, and the `gh release download` → re-upload path in `frontend-release.yml:404-409` that will re-publish a stale mac sidecar if the feed job is re-run against a tag that already has one. Neither blocks.

Still worth one real mac update against a sidecar-free build before this ships, since the root cause is stated as "likely" rather than confirmed.

harshitsinghbhandari:
Thanks @anmol0b, and thanks @illegalcall for the two review passes. The change itself is clean and I agree with the mechanism.

Two things before this can merge.

**1. It needs a coordinated change on the release-pipeline side.** Feed generation runs inside our signing pipeline, which asserts an expected number of `.blockmap` files before it publishes. This PR takes mac from 2 blockmaps to 0, so merging it on its own would fail the publish step on the next release. I'm handling that side; this PR just can't land first. Not something you need to change here.

**2. The premise may be stale, and it is worth one test before we give up differential updates.** Since 25 July our published mac zips are produced with `ditto -c -k --keepParent` rather than shipped straight from `maker-zip`. I compared the actual artifacts:

- `v0.10.4-nightly.202607221413` (the build in #3034): 281 entries, **0** AppleDouble `._*` entries
- `v0.10.4-nightly.202607280518` (current): 556 entries, **275** AppleDouble entries

including the exact four files named in the ditto errors in #3034: `Resources/._icon.png`, `._electron.icns`, `._app.asar`, `._icon.ico`. So the format mismatch this PR works around may already be gone, in which case merging it would disable mac differential updates (a full ~106 MB download per update instead of a diff) for no gain.

One end-to-end test settles it: install `v0.10.4-nightly.202607251158` on a Mac and let it auto-update to `v0.10.4-nightly.202607280518`. Both are ditto-built and both have blockmaps. If that update succeeds, #3034 is already fixed and we can close this out. If it fails, we merge this together with the pipeline change.

Holding on that, not on anything wrong with the code.

PR #3155, opened 2026-07-27
fix(crush): forward role-specific model config on launch

### LINKED ISSUE #2900: Crush: decide how to honor role-specific model config on launch

- Original report: #2769

## Decision needed

The **Crush** (Charm) adapter (`crush`, binary `crush`) does not forward `cfg.Config.Model`. An override mechanism exists but neither path fits AO's single `Model` string cleanly, so a decision is needed.

## Findings

- `GetLaunchCommand` (`backend/internal/adapters/agent/crush/crush.go:73-92`) emits only `--cwd` + `--yolo`; `GetRestoreCommand` (`:125-151`) adds `--session`. `hooks.go` merges only `options.context_paths` into `.crush.json` (`mergeCrushContextPath`, `:146-159`). Inherits empty `GetConfigSpec`.
- **Config route:** `.crush.json` has a top-level `models` object with `large`/`small` keys, each `{ "model": "<id>", "provider": "<providerID>" }` — **both fields required**. AO's `agentConfig.model` carries no provider id.
- **Flag route:** `--model`/`-m` is documented only for the non-interactive `crush run`; not confirmed on the interactive root command AO launches (UNKNOWN).
  Sources: https://deepwiki.com/charmbracelet/crush/4.3-model-configuration , https://deepwiki.com/charmbracelet/crush/2.2-cli-usage

## Options

1. **Config route:** extend the AO-managed `.crush.json` (`mergeCrushContextPath`) with `models.large` (and/or `.small`) — but this needs a `provider` value AO doesn't currently model. Would `agentConfig` need a `provider` field?
2. **Flag route:** first verify `crush -m <model>` works on the root interactive command (`crush --help` on the pinned version); if so, a simple flag fix applies.

## References

- Relevant file: `backend/internal/adapters/agent/crush/crush.go`

anmol0b:
## What
Forwards AO's per-role model config (`agentConfig.Model`) into Crush's `.crush.json` on launch, so a configured worker/orchestrator model actually takes effect instead of silently falling back to Crush's own default.

## Why
Fixes #2900.

Every adapter must opt in to forwarding `cfg.Config.Model` to its CLI. Claude, Codex, and Kiro already do this; Crush was one of the adapters that silently dropped it.

## How
The issue needed a decision between two routes:

- **Flag route (rejected):** Crush's `--model`/`-m` flag only exists on the non-interactive `crush run` subcommand — confirmed via Crush's CLI docs (root-only flags are just `--yolo`, `--session`, `--continue`). It's not available on the interactive root command AO actually launches, so a flag fix was never viable.
- **Config route (chosen):** `.crush.json`'s `models.large` selection requires both a `model` id and a `provider` id, but AO's `AgentConfig.Model` is a single bare string with no provider. Rather than adding a `Provider` field to AO's schema (which would ripple through the API, UI, and every other adapter), this scopes the fix to the Crush adapter only: it parses `Model` as the convention `"<provider>/<model-id>"` (e.g. `anthropic/claude-sonnet-4-5`) and merges it into `.crush.json`'s `models.large` in `GetAgentHooks`.

Implementation notes:
- `GetAgentHooks` runs before both Spawn and Restore (via `prepareWorkspace`), so the model override is re-applied on every launch and stays current if the role's config changes.
- The merge preserves any other fields already on `models.large` (e.g. a user's `max_tokens`/`temperature`), and any other top-level `.crush.json` config — it only owns the `model`/`provider` identity keys.
- If `Config.Model` is empty, nothing is touched, so a model the user picked through Crush's own model picker is left alone.
- A model string without a `/` (no provider) returns a clear error rather than silently forwarding something Crush would reject.
- `GetConfigSpec` now declares the `model` field and its required `"<provider>/<model-id>"` format, matching the `claude-code`/`codex`/`kiro` pattern.

**Intentional omission:** this doesn't add a `provider` field to `domain.AgentConfig`/the API/UI. That was one of the two options raised in the issue; I went with the string-convention approach to keep the change scoped to the Crush adapter. Happy to switch to a schema field if maintainers prefer that route instead.

## Testing
Ran the crush adapter's test suite locally:

```bash
go build ./internal/adapters/agent/crush/...
go vet ./internal/adapters/agent/crush/...
gofmt -l internal/adapters/agent/crush/*.go
go test ./internal/adapters/agent/crush/... -v
```

All existing tests pass, plus new coverage:
- `TestGetConfigSpecReportsModelField` — spec declares the `model` field
- `TestGetAgentHooksWritesModelOverride` — provider/model parsed and written into `models.large`
- `TestGetAgentHooksPreservesOtherModelFieldsOnOverride` — unrelated keys (e.g. `max_tokens`) survive the merge
- `TestGetAgentHooksLeavesModelUntouchedWhenNoOverride` — no override means no write, user's own selection untouched
- `TestGetAgentHooksRejectsModelWithoutProvider` — bare model string (no `/`) errors clearly

Screenshot 2026-07-27 at 9 38 37 AM

## Checklist

### DIFF
--- a/backend/internal/adapters/agent/crush/crush.go
+++ b/backend/internal/adapters/agent/crush/crush.go
@@ -54,6 +54,28 @@ func (p *Plugin) Manifest() adapters.Manifest {
 	}
 }
 
+// GetConfigSpec reports the per-project agent config keys Crush understands:
+// a model override. Unlike Claude/Codex (a bare --model flag) or Kiro (a
+// single model key), Crush's .crush.json requires both a provider id and a
+// model id per selection (see mergeCrushModel in hooks.go), so AO's plain
+// Model string must carry the provider too. This declares the expected
+// "<provider>/<model-id>" shape so it is discoverable/validated through the
+// config-spec surface, matching claude-code, codex, and kiro.
+func (p *Plugin) GetConfigSpec(ctx context.Context) (ports.ConfigSpec, error) {
+	if err := ctx.Err(); err != nil {
+		return ports.ConfigSpec{}, err
+	}
+	return ports.ConfigSpec{
+		Fields: []ports.ConfigField{
+			{
+				Key:         "model",
+				Type:        ports.ConfigFieldString,
+				Description: "Model override written into Crush's workspace-local config as the large-model selection. Must be \"<provider>/<model-id>\" (e.g. \"anthropic/claude-sonnet-4-5\"): Crush's config schema requires a provider id alongside the model id, which a bare model string doesn't carry.",
+			},
+		},
+	}, nil
+}
+
 // GetLaunchCommand builds the argv to start an interactive Crush session.
 // Shape:
 //
--- a/backend/internal/adapters/agent/crush/crush_test.go
+++ b/backend/internal/adapters/agent/crush/crush_test.go
@@ -485,15 +485,16 @@ func TestGetAgentHooksGitignoresManagedCrushFiles(t *testing.T) {
 	if strings.Contains(text, "/"+crushConfigFileName) {
 		t.Fatalf(".gitignore should not ignore project config %q:\n%s", crushConfigFileName, text)
 	}
-	data, err = os.ReadFile(filepath.Join(workspace, ".gitignore"))
-	if err != nil {
-		t.Fatalf("read workspace gitignore: %v", err)
-	}
-	text = string(data)
-	for _, want := range []string{hookutil.GitignoreSentinel, "/" + crushConfigFileName} {
-		if !strings.Contains(text, want) {
-			t.Fatalf("workspace .gitignore missing %q:\n%s", want, text)
-		}
+
+	// AO must not write/own a repo-root .gitignore: EnsureWorkspaceGitignore
+	// is a no-op when a root .gitignore already exists without AO's
+	// sentinel, and unconditionally overwrites one that does carry it on
+	// every subsequent call — neither is safe for a directory the user (or
+	// their repo) already owns. .crush.json churn in the agent's worktree
+	// is a known, pre-existing limitation left to a future AddExclude-based
+	// fix (writing to .git/info/exclude instead).
+	if _, err := os.Stat(filepath.Join(workspace, ".gitignore")); !os.IsNotExist(err) {
+		t.Fatalf("workspace root .gitignore should not be created by crush's adapter, stat err = %v", err)
 	}
 }
 
--- a/backend/internal/adapters/agent/crush/crush_test.go
+++ b/backend/internal/adapters/agent/crush/crush_test.go
@@ -348,7 +348,7 @@ func TestGetAgentHooksPreservesOtherModelFieldsOnOverride(t *testing.T) {
 	if err := os.MkdirAll(filepath.Dir(configPath), 0o750); err != nil {
 		t.Fatal(err)
 	}
-	if err := os.WriteFile(configPath, []byte(`{"models":{"large":{"model":"old-model","provider":"old-provider","max_tokens":4096}}}`), 0o600); err != nil {
+	if err := os.WriteFile(configPath, []byte(`{"models":{"large":{"model":"old-model","provider":"old-provider","max_tokens":4096},"small":{"model":"small-model","provider":"small-provider"}},"providers":{"openai":{"base_url":"https://example.test"}}}`), 0o600); err != nil {
 		t.Fatal(err)
 	}
 
@@ -360,12 +360,22 @@ func TestGetAgentHooksPreservesOtherModelFieldsOnOverride(t *testing.T) {
 		t.Fatalf("GetAgentHooks err = %v", err)
 	}
 
-	large := readCrushConfigForTest(t, configPath)["models"].(map[string]any)["large"].(map[string]any)
+	cfg := readCrushConfigForTest(t, configPath)
+	models := cfg["models"].(map[string]any)
+	large := models["large"].(map[string]any)
 	if large["provider"] != "openai" || large["model"] != "gpt-5" {
 		t.Fatalf("models.large = %#v, want provider=openai model=gpt-5", large)
 	}
 	if large["max_tokens"] != float64(4096) {
 		t.Fatalf("models.large.max_tokens was dropped: %#v", large)
+	}
+	small := models["small"].(map[string]any)
+	if small["provider"] != "small-provider" || small["model"] != "small-model" {
+		t.Fatalf("models.small = %#v, want existing small model untouched", small)
+	}
+	providers := cfg["providers"].(map[string]any)
+	if _, ok := providers["openai"].(map[string]any); !ok {
+		t.Fatalf("providers were dropped: %#v", providers)
 	}
 }
 
@@ -392,19 +402,64 @@ func TestGetAgentHooksLeavesModelUntouchedWhenNoOverride(t *testing.T) {
 	}
 }
 
-func TestGetAgentHooksRejectsModelWithoutProvider(t *testing.T) {
+func TestGetAgentHooksInfersProviderForBareModel(t *testing.T) {
 	workspace := t.TempDir()
+	configPath := crushConfigFile(workspace)
+	if err := os.MkdirAll(filepath.Dir(configPath), 0o750); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(configPath, []byte(`{"models":{"large":{"model":"old-model","provider":"anthropic","temperature":0.2}}}`), 0o600); err != nil {
+		t.Fatal(err)
+	}
 
-	err := (&Plugin{}).GetAgentHooks(context.Background(), ports.WorkspaceHookConfig{
+	if err := (&Plugin{}).GetAgentHooks(context.Background(), ports.WorkspaceHookConfig{
 		WorkspacePath: workspace,
 		SystemPrompt:  "AO standing instructions",
 		Config:        ports.AgentConfig{Model: "claude-sonnet-4-5"},
-	})
-	if err == nil {
-		t.Fatal("GetAgentHooks err = nil, want error for a model without a provider prefix")
+	}); err != nil {
+		t.Fatalf("GetAgentHooks err = %v", err)
 	}
-	if !strings.Contains(err.Error(), "provider") {
-		t.Fatalf("error = %v, want it to mention the missing provider", err)
+
+	large := readCrushConfigForTest(t, configPath)["models"].(map[string]any)["large"].(map[string]any)
+	if large["provider"] != "anthropic" || large["model"] != "claude-sonnet-4-5" {
+		t.Fatalf("models.large = %#v, want provider=anthropic model=claude-sonnet-4-5", large)
+	}
+	if large["temperature"] != 0.2 {
+		t.Fatalf("models.large.temperature was dropped: %#v", large)
+	}
+}
+
+func TestGetAgentHooksSkipsBareModelWithoutProvider(t *testing.T) {
+	workspace := t.TempDir()
+
+	if err := (&Plugin{}).GetAgentHooks(context.Background(), ports.WorkspaceHookConfig{
+		WorkspacePath: workspace,
+		SystemPrompt:  "AO standing instructions",
+		Config:        ports.AgentConfig{Model: "claude-sonnet-4-5"},
+	}); err != nil {
+		t.Fatalf("GetAgentHooks err = %v", err)
+	}
+
+	cfg := readCrushConfigForTest(t, crushConfigFile(workspace))
+	if _, ok := cfg["models"]; ok {
+		t.Fatalf("models were written for bare model with no provider to infer: %#v", cfg)
+	}
+}
+
+func TestGetAgentHooksSplitsModelOverrideOnFirstSlash(t *testing.T) {
+	workspace := t.TempDir()
+
+	if err := (&Plugin{}).GetAgentHooks(context.Background(), ports.WorkspaceHookConfig{
+		WorkspacePath: workspace,
+		SystemPrompt:  "AO standing instructions",
+		Config:        ports.AgentConfig{Model: "openrouter/anthropic/claude-x"},
+	}); err != nil {
+		t.Fatalf("GetAgentHooks err = %v", err)
+	}
+
+	large := readCrushConfigForTest(t, crushConfigFile(workspace))["models"].(map[string]any)["large"].(map[string]any)
+	if large["provider"] != "openrouter" || large["model"] != "anthropic/claude-x" {
+		t.Fatalf("models.large = %#v, want provider=openrouter model=anthropic/claude-x", large)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


illegalcall (commented):
Reviewed on a clean worktree at `8a7983e2e`. `go test ./internal/adapters/agent/crush/...` passes locally.

**Verdict: request changes.** The `.crush.json` merge itself is correct — it's the failure mode that needs rethinking.

## Verified

Checked against [crush's `schema.json`](https://raw.githubusercontent.com/charmbracelet/crush/main/schema.json): `models.<type>` requires **both** `model` and `provider`, and carries optional `max_tokens` / `temperature` / `reasoning_effort` / `think` / etc. So the premise is right — a bare model string genuinely can't satisfy crush's schema — and `mergeCrushModel` preserving the sibling fields is exactly correct.

## 1. A bare model value now hard-fails every crush spawn *and* restore (high)

`AgentConfig.Model` is a single shared string, merged project-level → role override in `effectiveAgentConfig` (`session_manager/manager.go:632-643`). Its own doc comment gives a **bare** example (`domain/agentconfig.go:26`: "e.g. claude-opus-4-5"), and every other adapter — droid (#3156), claude-code, codex, kiro — consumes it bare.

So consider a project that sets `agentConfig.model: "claude-opus-4-5"` for its claude worker and runs a **crush orchestrator**. `effectiveAgentConfig` hands crush that bare string, and `applyCrushModelOverride` returns an error, which `prepareWorkspace` (`manager.go:2602-2612`) wraps and returns — aborting `Spawn` (`:384`) and `Restore` (`:1043`) and running `cleanupPreparedAgentWorkspace`:

```
install hooks: crush.GetAgentHooks: model "claude-opus-4-5" must be "<provider>/<model-id>" …
```

Before this PR crush ignored `Model` and started fine. This turns "config silently ignored" into "session will not start" — triggered by config the user set for a *different* agent. That's the part I'd push back on: the failure is remote from the cause, and it happens at spawn time when the user is trying to do something else.

Suggested fix — don't fail the launch:

- skip the override with a `logger.Warn` when there's no provider prefix; **or**
- reuse the provider already present in `.crush.json` when one is set, and only skip when there's nothing to infer.

If you want a hard error, put it in `AgentConfig.Validate()` where the user is actually setting the value and can act on the message — not at spawn.

## 2. The model override sits behind an unrelated early return (medium, latent)

`GetAgentHooks` returns early at `hooks.go:52-54` when the system prompt is empty, and `applyCrushModelOverride` is the **last** statement at `:82`. Today `buildSystemPrompt` always returns non-empty for both `KindWorker` and `KindOrchestrator` (`manager.go:2239-2283`), so it's unreachable — I checked. But the coupling is invisible, and one new session kind or an empty-rules path silently drops the model with no error and no log.

Move `applyCrushModelOverride` above the prompt work. It has nothing to do with the prompt, and the current placement makes it look like it does.

## 3. `.crush.json` isn't gitignored (low, pre-existing but amplified)

It lives at the workspace root; only `.crush/ao-system-prompt.md` is excluded (`hooks.go:79` — `EnsureWorkspaceGitignore(filepath.Dir(promptPath), ...)` writes `.crush/.gitignore`). AO already mutates `.crush.json` via `mergeCrushContextPath`, so this isn't new, but the model write adds churn to a file sitting in the agent's own working tree, where the agent can commit it into its PR. An `AddExclude` for `/.crush.json` would close it.

## 4. The contract is documented only where nobody reads it (low)

The `<provider>/<model-id>` requirement is stated in `GetConfigSpec` — which has **no production caller**. Grepping `backend/` finds 15 adapter implementations, `ports/agent.go:40`, and test doubles; nothing consumes the spec. So the single place that explains the format is never rendered in any UI, CLI, or doc. That's a large part of why finding #1 bites: the user gets an error at spawn time for violating a rule they were never shown.

## 5. `strings.Cut` splits on the first `/` (nit)

`openrouter/anthropic/claude-x` → provider `openrouter`, model `anthropic/claude-x`. That is the correct behavior for openrouter-style ids — please pin it with a test so nobody later "fixes" it to `LastIndex`.

## Tests

Good coverage of the paths you added. Two gaps worth closing:

- assert `models.small` (and other top-level keys) survive the merge — the comment on `mergeCrushModel` promises this but nothing checks it
- the first-slash case from #5

illegalcall (changes requested):
Formalizing the verdict from my detailed review above as **request changes**.

One blocking item:

**A bare `agentConfig.model` now hard-fails every crush spawn and restore.** `AgentConfig.Model` is one shared string merged project-level → role override (`manager.go:632-643`), documented with a bare example (`domain/agentconfig.go:26`), and consumed bare by droid/claude-code/codex/kiro. A project that sets `model: "claude-opus-4-5"` for its claude worker and runs a crush orchestrator now gets `install hooks: crush.GetAgentHooks: model … must be "<provider>/<model-id>"` out of `prepareWorkspace`, which aborts both Spawn and Restore. That's a regression from "config ignored" to "session won't start", triggered by config set for a different agent — and the format requirement is only stated in `GetConfigSpec`, which has no production caller, so the user was never shown the rule they violated.

Please make the missing-provider case non-fatal: skip with a `logger.Warn`, or infer the provider from the existing `.crush.json` when one is set. If you want a hard error, move it to `AgentConfig.Validate()` where the user can act on it.

Also worth fixing in the same pass (not blocking on its own): `applyCrushModelOverride` sits after the empty-system-prompt early return at `hooks.go:52-54`, so it's silently gated on something unrelated. Move it above that block.

The `mergeCrushModel` implementation is correct — I verified the provider requirement and the preserved sibling fields against crush's `schema.json`.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

anmol0b:
> Formalizing the verdict from my detailed review above as **request changes**.
> 
> One blocking item:
> 
> **A bare `agentConfig.model` now hard-fails every crush spawn and restore.** `AgentConfig.Model` is one shared string merged project-level → role override (`manager.go:632-643`), documented with a bare example (`domain/agentconfig.go:26`), and consumed bare by droid/claude-code/codex/kiro. A project that sets `model: "claude-opus-4-5"` for its claude worker and runs a crush orchestrator now gets `install hooks: crush.GetAgentHooks: model … must be "<provider>/<model-id>"` out of `prepareWorkspace`, which aborts both Spawn and Restore. That's a regression from "config ignored" to "session won't start", triggered by config set for a different agent — and the format requirement is only stated in `GetConfigSpec`, which has no production caller, so the user was never shown the rule they violated.
> 
> Please make the missing-provider case non-fatal: skip with a `logger.Warn`, or infer the provider from the existing `.crush.json` when one is set. If you want a hard error, move it to `AgentConfig.Validate()` where the user can act on it.
> 
> Also worth fixing in the same pass (not blocking on its own): `applyCrushModelOverride` sits after the empty-system-prompt early return at `hooks.go:52-54`, so it's silently gated on something unrelated. Move it above that block.
> 
> The `mergeCrushModel` implementation is correct — I verified the provider requirement and the preserved sibling fields against crush's `schema.json`.

Thanks @illegalcall , agreed on the blocking failure mode.

  I updated the branch so a bare `agentConfig.model` no longer aborts Crush spawn/restore. Crush now reuses the existing `models.large.provider` from `.crush.json` when available;
  if there is no provider to infer, it skips the override and logs a warning instead of failing launch.

  I also moved the model override before the system-prompt early return, added coverage for that path, pinned first-slash splitting for OpenRouter-style IDs, and expanded the merge
  tests to ensure `models.small` plus other top-level config survive.

  I also added ignore coverage for AO-managed `.crush.json` writes.

  Re-ran:

  - `go test ./internal/adapters/agent/crush/...`
  - `go build ./internal/adapters/agent/crush/...`
  - `go vet ./internal/adapters/agent/crush/...`
  - `gofmt -l backend/internal/adapters/agent/crush/*.go`

illegalcall (changes requested):
Re-reviewed `b0eca1650`. Thanks for the quick turnaround — **the blocking item is resolved**, and I verified it.

Confirmed fixed:

- **bare model no longer aborts spawn/restore** ✅ — `applyCrushModelOverride` now infers the provider from `models.large.provider` and falls back to warn-and-skip. `slog.Default()` matches existing adapter precedent (`cursor/hooks.go:79`), so no concern there. Edge cases behave: `"foo/"` and `"/bar"` both warn and skip rather than writing a half-selection.
- **moved above the empty-prompt early return** ✅, with `TestGetAgentHooksAppliesModelOverrideWithEmptyPrompt` pinning it
- **first-slash split pinned** ✅ (`openrouter/anthropic/claude-x`)
- **`models.small` + `providers` preservation asserted** ✅

`go test ./internal/adapters/agent/crush/...` passes locally.

---

## One new issue, introduced by the gitignore hunk

Keeping this at request-changes for this only. I'm sorry — my original suggestion said `AddExclude`, and `EnsureWorkspaceGitignore` behaves quite differently at the repo root.

`hookutil.EnsureWorkspaceGitignore(dir, names...)` (`hookutil.go:30-54`) does three things depending on what it finds at `<dir>/.gitignore`:

1. file exists **without** the AO sentinel → `return nil`, no change
2. file absent → create it with sentinel + `/.gitignore` + names
3. file exists **with** the sentinel → **overwrite the whole file** with sentinel + `/.gitignore` + names

**(1) makes the fix a no-op in the common case.** Any repo that already has a root `.gitignore` — essentially every real repo — is skipped. I probed it directly: temp workspace with a pre-existing `.gitignore` containing `node_modules/`, then `GetAgentHooks` with `Config.Model = "anthropic/claude-sonnet-4-5"`:

```
root .gitignore after hooks:
node_modules/
dist/
.crush.json written: true (193 bytes)
RESULT: .crush.json is NOT ignored
```

So the case my original note was about — the model write dirtying `.crush.json` in the agent's own worktree — is unchanged.

**(3) is the part I'd actually block on.** In a repo with no root `.gitignore`, AO now creates one, and because it carries the sentinel, *every subsequent* `GetAgentHooks` call overwrites it wholesale. Anything the user or the agent later adds to that file is silently discarded on the next spawn or restore. That's fine for `.crush/` and the other AO-owned subdirectories — which is what all 14 other call sites target (`cline`, `kilocode`, `copilot`, `cursor`, `agy`, `kiro`, `amp`, `hooksjson`, `vibe`, `kimi`, `opencode`, and crush's own `.crush/` call). This is the first call anywhere in the tree that points at the repo root, and the overwrite semantics weren't designed for a directory the user owns.

Also note the call at `hooks.go:89` is unconditional in the prompt path, so this runs for **every** crush session, not just ones with a model override — the blast radius is all crush users.

Two ways out, either is fine by me:

- **Drop the hunk.** It was a "low" nit in my first pass, not a blocker, and `.crush.json` churn predates this PR via `mergeCrushContextPath`. Shipping the model fix without it is strictly better than shipping a no-op with a root-`.gitignore` side effect.
- **Use `AddExclude`.** `ports.Workspace.AddExclude` writes `.git/info/exclude`, which is never tracked, always effective regardless of existing `.gitignore` content, and never clobbers a user file. `manager.go:370` already uses it for the attachments dir. It'd need plumbing from the session manager since the adapter has no `Workspace` handle — which is itself an argument for option one and a separate follow-up.

Everything else here is good to go; flip this one and I'll approve.

anmol0b:
> Re-reviewed `b0eca1650`. Thanks for the quick turnaround — **the blocking item is resolved**, and I verified it.
> 
> Confirmed fixed:
> 
> * **bare model no longer aborts spawn/restore** ✅ — `applyCrushModelOverride` now infers the provider from `models.large.provider` and falls back to warn-and-skip. `slog.Default()` matches existing adapter precedent (`cursor/hooks.go:79`), so no concern there. Edge cases behave: `"foo/"` and `"/bar"` both warn and skip rather than writing a half-selection.
> * **moved above the empty-prompt early return** ✅, with `TestGetAgentHooksAppliesModelOverrideWithEmptyPrompt` pinning it
> * **first-slash split pinned** ✅ (`openrouter/anthropic/claude-x`)
> * **`models.small` + `providers` preservation asserted** ✅
> 
> `go test ./internal/adapters/agent/crush/...` passes locally.
> 
> ## One new issue, introduced by the gitignore hunk
> Keeping this at request-changes for this only. I'm sorry — my original suggestion said `AddExclude`, and `EnsureWorkspaceGitignore` behaves quite differently at the repo root.
> 
> `hookutil.EnsureWorkspaceGitignore(dir, names...)` (`hookutil.go:30-54`) does three things depending on what it finds at `<dir>/.gitignore`:
> 
> 1. file exists **without** the AO sentinel → `return nil`, no change
> 2. file absent → create it with sentinel + `/.gitignore` + names
> 3. file exists **with** the sentinel → **overwrite the whole file** with sentinel + `/.gitignore` + names
> 
> **(1) makes the fix a no-op in the common case.** Any repo that already has a root `.gitignore` — essentially every real repo — is skipped. I probed it directly: temp workspace with a pre-existing `.gitignore` containing `node_modules/`, then `GetAgentHooks` with `Config.Model = "anthropic/claude-sonnet-4-5"`:
> 
> ```
> root .gitignore after hooks:
> node_modules/
> dist/
> .crush.json written: true (193 bytes)
> RESULT: .crush.json is NOT ignored
> ```
> 
> So the case my original note was about — the model write dirtying `.crush.json` in the agent's own worktree — is unchanged.
> 
> **(3) is the part I'd actually block on.** In a repo with no root `.gitignore`, AO now creates one, and because it carries the sentinel, _every subsequent_ `GetAgentHooks` call overwrites it wholesale. Anything the user or the agent later adds to that file is silently discarded on the next spawn or restore. That's fine for `.crush/` and the other AO-owned subdirectories — which is what all 14 other call sites target (`cline`, `kilocode`, `copilot`, `cursor`, `agy`, `kiro`, `amp`, `hooksjson`, `vibe`, `kimi`, `opencode`, and crush's own `.crush/` call). This is the first call anywhere in the tree that points at the repo root, and the overwrite semantics weren't designed for a directory the user owns.
> 
> Also note the call at `hooks.go:89` is unconditional in the prompt path, so this runs for **every** crush session, not just ones with a model override — the blast radius is all crush users.
> 
> Two ways out, either is fine by me:
> 
> * **Drop the hunk.** It was a "low" nit in my first pass, not a blocker, and `.crush.json` churn predates this PR via `mergeCrushContextPath`. Shipping the model fix without it is strictly better than shipping a no-op with a root-`.gitignore` side effect.
> * **Use `AddExclude`.** `ports.Workspace.AddExclude` writes `.git/info/exclude`, which is never tracked, always effective regardless of existing `.gitignore` content, and never clobbers a user file. `manager.go:370` already uses it for the attachments dir. It'd need plumbing from the session manager since the adapter has no `Workspace` handle — which is itself an argument for option one and a separate follow-up.
> 
> Everything else here is good to go; flip this one and I'll approve.

Thanks for digging into `EnsureWorkspaceGitignore`'s repo-root behavior — good catch, especially the "clobbers on every subsequent call" case.

Went with **option 1 (drop the hunk)**. Pushed in the latest commit:

- Removed both `EnsureWorkspaceGitignore(cfg.WorkspacePath, crushConfigFileName)` calls (the conditional one gated on a model write, and the unconditional one in the prompt path).
- Simplified `applyCrushModelOverride` back to an `error`-only return, since the `bool` it returned had no remaining purpose once the conditional gitignore call was removed.
- Updated `TestGetAgentHooksGitignoresManagedCrushFiles` to assert AO does *not* create a root `.gitignore` anymore, instead of asserting the old (unsafe) behavior.

`.crush/ao-system-prompt.md` is still gitignored as before (that call was never at the repo root and isn't affected). `.crush.json` churn in the agent's own worktree is unchanged from `main` (pre-existing via `mergeCrushContextPath`) — agreed that's a real gap, but per your note it's better scoped as a separate `AddExclude`-based follow-up once there's a way to plumb a `Workspace` handle into the adapter. Happy to file that issue if useful.

`go build`, `go vet`, `gofmt -l`, and `go test ./internal/adapters/agent/crush/... -v` all pass locally (24/24).

illegalcall (approved):
Re-reviewed `7410f9cbb`. Approved.

The sole remaining blocker from the last pass is resolved exactly as requested:

- both unsafe repo-root `EnsureWorkspaceGitignore` calls are gone
- the now-unused boolean return from `applyCrushModelOverride` is removed cleanly
- the regression test asserts the adapter does not create or own the repository root `.gitignore`
- the AO-owned `.crush/ao-system-prompt.md` ignore behavior remains intact

`go test ./internal/adapters/agent/crush/...` and package vet pass locally; all eight CI checks are green. No new issue in the follow-up diff.

The broader `.crush.json` ignore gap remains pre-existing and should be handled separately with a workspace `AddExclude` path; I am not reopening it here after explicitly offering “drop the hunk” as the acceptable remedy.

PR #3156, opened 2026-07-27
fix(droid): forward role-specific model config via settings file

### LINKED ISSUE #2898: Droid: decide whether to honor role-specific model config on launch

- Original report: #2769

## Decision needed

The **Droid** (Factory) adapter (`droid`, binary `droid`) does **not** forward `cfg.Config.Model`, and this is currently **by design** — the package comment states *"Droid resolves its model and other defaults from the user's own settings; only the autonomy level is overridden"* (`droid/droid.go:70-74`, `:149-169`). We need a product/architecture decision on whether AO should override the model for Droid roles.

## Findings

- **No interactive `--model` flag.** `-m, --model <id>` exists only on `droid exec`; the interactive `droid [prompt]` mode AO launches changes model via `/model` or `Ctrl+N` mid-session (Factory CLI reference — https://docs.factory.ai/reference/cli-reference).
- **Model *is* settable via the settings file** AO already writes and passes: top-level `"model": "<id>"` in the JSON blob emitted by `permissionSettingsArgs` (`droid/droid.go:180-198`) and passed with `--settings <path>`. Currently that file is written **only for non-default permission modes** and carries only `autonomyLevel` (Factory settings docs — https://docs.factory.ai/cli/configuration/settings).
- **Upstream risk:** open Factory issues show model-config precedence flakiness (Factory-AI/factory #660, #787), and the docs don't explicitly confirm the top-level `model` in a `--settings` file overrides the session.

## Options

1. **Honor it:** add top-level `"model": strings.TrimSpace(cfg.Config.Model)` to the settings blob and emit `--settings` whenever autonomy **or** model is set. Requires verifying against a live `droid` that `--settings` model actually re-pins the session model.
2. **Keep deferring** to the user's `~/.factory/settings.json` and document that `agentConfig.model` is intentionally a no-op for Droid.

## Suggested fix (only if option 1 chosen)

Extend the settings builder (`permissionSettingsArgs`) with the top-level `model` key; always write the settings file when a model is set; add a `GetConfigSpec` `model` field; add regression tests. **Do not** add a `--model` argv flag — the interactive binary rejects it.

## References

- Relevant file: `backend/internal/adapters/agent/droid/droid.go`

anmol0b:
## What
Droid adapter now forwards `agentConfig.model` into Droid's runtime `--settings`
file instead of silently dropping it.

## Why
Fixes #2898. AO resolves a per-role model into `cfg.Config.Model` and forwards
it to every adapter, but each adapter has to opt in to actually using it.
Claude Code, Codex (#2869), and Kiro already do; Droid didn't — a configured
worker/orchestrator model silently fell back to the CLI's global default.

## How
Interactive `droid` has no `--model` flag (it only exists on `droid exec`), so
model can't be forwarded as an argv flag the way Codex does it. Instead this
folds a top-level `"model"` key into the same runtime settings JSON blob
already used for permission autonomy (`sessionDefaultSettings.autonomyLevel`),
and writes/passes `--settings <path>` whenever *either* value is set (previously
only written for non-default permission modes).

- `permissionSettingsArgs` takes a `model` param, trims it, and sets `model` in
  the settings blob when non-empty.
- `GetLaunchCommand` / `GetRestoreCommand` pass `cfg.Config.Model` through.
- Added `GetConfigSpec` (Droid had none before, so `model` wasn't surfaced as a
  configurable field at all) — mirrors the Codex adapter's spec shape.
- Updated package/function doc comments — the old comments explicitly said
  "only the autonomy level is overridden," which is no longer accurate.

**Intentional omission / open question carried over from the issue:** whether
Droid's `--settings` top-level `model` key actually re-pins the session model
live isn't confirmed by Factory's docs — the issue flags this explicitly as
unverified upstream behavior (see linked Factory issues in #2898). I don't have
a live `droid` binary to confirm against, so this needs a manual smoke test
before merge. If it turns out `--settings`'s `model` key is ignored by Droid in
practice, this PR's mechanism is wrong and option 2 (documented no-op) from the
issue should be taken instead.

## Testing
- `go test ./internal/adapters/agent/droid/...` — all existing tests pass, plus
  new regression tests: model-only settings write, model+autonomy combined in
  one file, and restore forwarding model.
- `go test ./...` — full backend suite green.
- `go vet ./internal/adapters/agent/droid/...` — clean.
- Grepped `docs/`, `backend/`, `frontend/` for the old "Droid resolves its own
  model" language — none found outside the file itself.
- **Not yet done:** manual verification against a live `droid` binary that the
  settings-file `model` key actually changes the session's model.

## Checklist
- [x] Branched from `main`
- [x] One focused change; closes #2898
- [x] `go test ./...` and `go vet` pass for the area touched

### DIFF
--- a/backend/internal/adapters/agent/droid/droid.go
+++ b/backend/internal/adapters/agent/droid/droid.go
@@ -11,11 +11,13 @@
 //
 // Launch uses the interactive `droid [prompt]` command (the prompt is a
 // positional argument). Droid's interactive TUI exposes no per-launch permission
-// flag (--auto / --skip-permissions-unsafe live only on `droid exec`), so AO's
-// graduated permission modes are delivered by writing a process-scoped runtime
-// settings file (sessionDefaultSettings.autonomyLevel) and passing it via the
-// root `--settings <path>` flag. Restore prefers the hook-captured native
-// session id via `-r <id>`.
+// flag (--auto / --skip-permissions-unsafe live only on `droid exec`) and no
+// interactive --model flag either, so AO's graduated permission modes and any
+// role-specific model override are both delivered by writing a process-scoped
+// runtime settings file (sessionDefaultSettings.autonomyLevel and a top-level
+// "model" key) and passing it via the root `--settings <path>` flag. The
+// settings file is written whenever either value is set. Restore prefers the
+// hook-captured native session id via `-r <id>`.
 package droid
 
 import (
@@ -63,6 +65,21 @@ func (p *Plugin) Manifest() adapters.Manifest {
 	}
 }
 
+func (p *Plugin) GetConfigSpec(ctx context.Context) (ports.ConfigSpec, error) {
+	if err := ctx.Err(); err != nil {
+		return ports.ConfigSpec{}, err
+	}
+	return ports.ConfigSpec{
+		Fields: []ports.ConfigField{
+			{
+				Key:         "model",
+				Type:        ports.ConfigFieldString,
+				Description: "Model override written to Droid's runtime settings file.",
+			},
+		},
+	}, nil
+}
+
 // GetLaunchCommand builds the argv to start a new interactive Droid session:
 //
 //	droid [--settings <path>] [--append-system-prompt[-file] <x>] [prompt]
@@ -81,7 +98,7 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 	cmd = make([]string, 0, 6)
 	cmd = append(cmd, binary)
 
-	settingsArgs, err := permissionSettingsArgs(cfg.SessionID, cfg.Permissions)
+	settingsArgs, err := permissionSettingsArgs(cfg.SessionID, cfg.Permissions, cfg.Config.Model)
 	if err != nil {
 		return nil, err
 	}
@@ -122,7 +139,7 @@ func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig)
 
 	cmd = make([]string, 0, 5)
 	cmd = append(cmd, binary)
-	settingsArgs, err := permissionSettingsArgs(cfg.Session.ID, cfg.Permissions)
+	settingsArgs, err := permissionSettingsArgs(cfg.Session.ID, cfg.Permissions, cfg.Config.Model)
 	if err != nil {
 		return nil, false, err
 	}
@@ -177,15 +194,22 @@ func droidAutonomyLevel(mode ports.PermissionMode) string {
 // --skip-permissions-unsafe exist only on `droid exec`), so autonomy must be
 // delivered through settings. The file is written under the OS temp dir, keyed
 // by session id, rather than into the worktree so it never lands in a commit.
-func permissionSettingsArgs(sessionID string, mode ports.PermissionMode) ([]string, error) {
+func permissionSettingsArgs(sessionID string, mode ports.PermissionMode, model string) ([]string, error) {
 	level := droidAutonomyLevel(mode)
-	if level == "" {
+	model = strings.TrimSpace(model)
+	if level == "" && model == "" {
 		return nil, nil
 	}
 
-	blob, err := json.Marshal(map[string]any{
-		"sessionDefaultSettings": map[string]any{"autonomyLevel": level},
-	})
+	settings := map[string]any{}
+	if level != "" {
+		settings["sessionDefaultSettings"] = map[string]any{"autonomyLevel": level}
+	}
+	if model != "" {
+		settings["model"] = model
+	}
+
+	blob, err := json.Marshal(settings)
 	if err != nil {
 		return nil, fmt.Errorf("droid: encode runtime settings: %w", err)
 	}
--- a/backend/internal/adapters/agent/droid/droid.go
+++ b/backend/internal/adapters/agent/droid/droid.go
@@ -66,7 +66,7 @@ func (p *Plugin) Manifest() adapters.Manifest {
 }
 
 // GetConfigSpec reports the per-project agent config keys Droid understands.
-// Model is honored via the runtime --settings file (see permissionSettingsArgs)
+// Model is honored via the runtime --settings file (see runtimeSettingsArgs)
 // rather than an argv flag, since interactive droid has no --model flag.
 func (p *Plugin) GetConfigSpec(ctx context.Context) (ports.ConfigSpec, error) {
 	if err := ctx.Err(); err != nil {
@@ -87,11 +87,12 @@ func (p *Plugin) GetConfigSpec(ctx context.Context) (ports.ConfigSpec, error) {
 //
 //	droid [--settings <path>] [--append-system-prompt[-file] <x>] [prompt]
 //
-// The prompt is delivered as a positional argument (in command). Droid resolves
-// its model and other defaults from the user's own settings; only the autonomy
-// level is overridden, and only for non-default permission modes (see
-// permissionSettingsArgs). System-prompt text/file is appended (not replaced),
-// matching Droid's --append-system-prompt semantics.
+// The prompt is delivered as a positional argument (in command). Both the
+// permission autonomy level and any role-specific model override are written
+// to the runtime --settings file when set (see runtimeSettingsArgs); when
+// neither is set, Droid resolves everything from the user's own settings.
+// System-prompt text/file is appended (not replaced), matching Droid's
+// --append-system-prompt semantics.
 func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (cmd []string, err error) {
 	binary, err := p.droidBinary(ctx)
 	if err != nil {
@@ -101,7 +102,7 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 	cmd = make([]string, 0, 6)
 	cmd = append(cmd, binary)
 
-	settingsArgs, err := permissionSettingsArgs(cfg.SessionID, cfg.Permissions, cfg.Config.Model)
+	settingsArgs, err := runtimeSettingsArgs(cfg.SessionID, cfg.Permissions, cfg.Config.Model)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


illegalcall (commented):
Reviewed on a clean worktree at `532ec9334`. `go test ./internal/adapters/agent/droid/...` passes locally.

**Verdict: approve once the tests land.** Smallest and cleanest of the model-config batch.

## Verified

I checked the load-bearing claim against [Factory's settings docs](https://docs.factory.ai/cli/configuration/settings):

- A top-level `"model"` key does exist, and it takes a **bare** model id — the docs' own example is `{"model": "claude-opus-4-7"}`. That matches AO's `AgentConfig.Model` convention (`domain/agentconfig.go:26` documents it as e.g. `claude-opus-4-5`), so no format translation is needed here.
- `--settings` applies as an **overlay** at org-level precedence, not a replacement. That matters: it means writing `{"model": ...}` alone for a default-permission session does not clobber the user's `~/.factory/settings.json`. Worth recording somewhere, because the whole design rests on it.

## 1. No test covers the new behavior (medium)

The only test touched is the config-spec assertion. Nothing exercises the actual change. Missing cases:

- model set + **default** permissions → `--settings` is emitted, and the file contains `model` with **no** `sessionDefaultSettings` key
- model + non-default permission → both keys present in one file
- `GetRestoreCommand` forwards `cfg.Config.Model`

AGENTS.md asks for tests covering "the user-visible behavior and boundary being changed", and this is the only boundary that changed. `TestGetLaunchCommandBypassWritesSettings` is right there as a template — it already reads the file back and parses it.

## 2. `permissionSettingsArgs` is now misnamed (nit)

It renders permissions *or* model; `runtimeSettingsArgs` would be honest. Its doc comment at `droid.go:172-178` is now wrong in two places:

- "renders a non-default permission mode as a `--settings` argv pair" — it also renders a model
- "The default mode returns nil (no flag, no file)" — no longer true when a model is set

## 3. Wider blast radius than the diff suggests (nit)

Before this PR, `--settings` was only ever passed for non-default permission modes. Now **any** project with `agentConfig.model` set gets `--settings` on every droid launch and restore. The merge semantics above make that safe, but it's a bigger behavioral surface than a 42-line diff implies — worth one line in the PR body so the next person doesn't have to re-derive it.

## Cross-cutting (applies to #3155 / #3167 too)

`GetConfigSpec` has **no production caller**. Grepping `backend/` for it turns up 15 adapter implementations, `ports/agent.go:40`, and test doubles — nothing else. So the `Description` string added here is never rendered in any UI, CLI, or doc. Not a blocker for this PR, but it means the config-spec entries in this batch are documentation that no user will see. Probably worth its own issue to either wire it into project-config validation/UI or stop treating it as a user-facing surface.

Also worth a note in `domain/agentconfig.go`: this PR treats `Model` as a bare id, #3155 makes crush require `<provider>/<model-id>`. Same shared field, two contracts. Whoever wires the next adapter will need to know that.

illegalcall (approved):
Approving. I verified against Factory's settings docs that a top-level `"model"` key exists and takes a bare model id, and that `--settings` applies as an overlay rather than a replacement — so writing `{"model": …}` alone on a default-permission launch won't clobber the user's own settings. The approach is sound.

The one thing I'd like before merge: **a test for the behavior this PR adds.** Right now only the config-spec assertion changed. `TestGetLaunchCommandBypassWritesSettings` is already the right template — it reads the file back and parses it. Worth covering model-only + default permissions (file has `model`, no `sessionDefaultSettings`), both values together, and the restore path.

Also rename `permissionSettingsArgs` / fix its doc comment — it renders permissions *or* model now, and the comment still says the default mode writes no file.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

anmol0b:
> Approving. I verified against Factory's settings docs that a top-level `"model"` key exists and takes a bare model id, and that `--settings` applies as an overlay rather than a replacement — so writing `{"model": …}` alone on a default-permission launch won't clobber the user's own settings. The approach is sound.
> 
> The one thing I'd like before merge: **a test for the behavior this PR adds.** Right now only the config-spec assertion changed. `TestGetLaunchCommandBypassWritesSettings` is already the right template — it reads the file back and parses it. Worth covering model-only + default permissions (file has `model`, no `sessionDefaultSettings`), both values together, and the restore path.
> 
> Also rename `permissionSettingsArgs` / fix its doc comment — it renders permissions _or_ model now, and the comment still says the default mode writes no file.

Pushed the changes @illegalcall asked for in review: renamed
`permissionSettingsArgs` → `runtimeSettingsArgs` (with an updated doc comment
now that it renders permissions and/or model), added the three regression
tests covering model-only + default permissions, model + non-default
permissions, and restore forwarding model, and documented both the
`--settings` overlay semantics and the cross-adapter `Model` format difference
with #3155 for whoever wires the next adapter.

Full suite is green (`go test ./...`, `go vet` clean) and the PR description
is updated to reflect all of this.

One thing genuinely still open, not something I can resolve from here: I don't
have a live `droid` binary, so the manual smoke test confirming the
`--settings` `model` key actually re-pins the session model is still
outstanding. Flagging that clearly so it isn't missed before merge — happy to
run it myself if someone can point me to a way to get a droid binary/license,
otherwise this needs a maintainer or someone with local access to verify
before it lands.

illegalcall (changes requested):
Re-reviewed `b055b9afb`. The requested behavior tests are good: model-only/default permissions, model plus autonomy, and restore forwarding are all covered. The helper rename is honest, package tests/vet pass, and all eight CI checks are green.

One hard repository boundary still blocks this head: `runtimeSettingsArgs` writes AO-managed settings to `os.TempDir()`. AGENTS.md and `docs/architecture.md` require **all app state under `~/.ao` / `AO_DATA_DIR`**. This path existed for non-default permissions, but the PR now expands it to every model-only default-permission launch, so the change materially broadens the violation.

Both `ports.LaunchConfig` and `ports.RestoreConfig` already carry `DataDir`. Thread that into `runtimeSettingsArgs` / `runtimeSettingsPath` and place the file under a deterministic subdirectory such as `<DataDir>/agent-runtime/droid/...`, creating the parent safely. Update the tests to use `t.TempDir()` as `DataDir`.

Also repair the restore doc comment, which now ends mid-sentence at “when the hook-derived”. The live Droid smoke test remains outstanding as the PR body says; unit tests prove JSON/argv construction, not that interactive Droid actually re-pins the model through `--settings`.

anmol0b:
> Re-reviewed `b055b9afb`. The requested behavior tests are good: model-only/default permissions, model plus autonomy, and restore forwarding are all covered. The helper rename is honest, package tests/vet pass, and all eight CI checks are green.
> 
> One hard repository boundary still blocks this head: `runtimeSettingsArgs` writes AO-managed settings to `os.TempDir()`. AGENTS.md and `docs/architecture.md` require **all app state under `~/.ao` / `AO_DATA_DIR`**. This path existed for non-default permissions, but the PR now expands it to every model-only default-permission launch, so the change materially broadens the violation.
> 
> Both `ports.LaunchConfig` and `ports.RestoreConfig` already carry `DataDir`. Thread that into `runtimeSettingsArgs` / `runtimeSettingsPath` and place the file under a deterministic subdirectory such as `<DataDir>/agent-runtime/droid/...`, creating the parent safely. Update the tests to use `t.TempDir()` as `DataDir`.
> 
> Also repair the restore doc comment, which now ends mid-sentence at “when the hook-derived”. The live Droid smoke test remains outstanding as the PR body says; unit tests prove JSON/argv construction, not that interactive Droid actually re-pins the model through `--settings`.

Thanks for catching this. Pushed a fix for the data-dir boundary:

- `runtimeSettingsArgs` / `runtimeSettingsPath` now take `dataDir` and write
  under `<DataDir>/agent-runtime/droid/...` instead of `os.TempDir()`, per
  AGENTS.md / docs/architecture.md. Parent dir is created via `os.MkdirAll`
  before the atomic write.
- Both `GetLaunchCommand` and `GetRestoreCommand` now thread `cfg.DataDir`
  through.
- All four settings-writing tests updated to use `t.TempDir()` as `DataDir`
  (dropped the now-redundant manual `os.Remove` cleanup, since `t.TempDir()`
  handles it automatically).
- Fixed the truncated restore doc comment ("...falls back to fresh launch
  behavior — mirroring the Codex and opencode adapters.").

Full suite green: `go test ./internal/adapters/agent/droid/... -v`,
`go vet`, and `go test ./...` all pass clean.

Agreed the live smoke test is still the one thing unit tests can't prove —
still outstanding, flagging again so it doesn't get lost at merge time.

illegalcall (changes requested):
Re-reviewed current head 2be62f2.

The DataDir boundary fix is correct, the launch/restore regression tests cover the emitted settings JSON and argv, targeted adapter tests/vet pass, and all eight CI checks are green. I found no remaining code-level defect in the implementation.

I am not approving this for merge yet because #2898 explicitly requires live Droid verification. The unit tests prove file construction, not that interactive Droid actually applies the role-specific model on both a fresh launch and a restored session. Please record one live smoke result covering both paths and confirming the active model. Once that evidence is present, I consider this ready to approve.

anmol0b:
> Re-reviewed current head [2be62f2](https://github.com/Untrivial-ai/agent-orchestrator/commit/2be62f240f4e1bfe5858aaf19d64641c39c6297c).
> 
> The DataDir boundary fix is correct, the launch/restore regression tests cover the emitted settings JSON and argv, targeted adapter tests/vet pass, and all eight CI checks are green. I found no remaining code-level defect in the implementation.
> 
> I am not approving this for merge yet because #2898 explicitly requires live Droid verification. The unit tests prove file construction, not that interactive Droid actually applies the role-specific model on both a fresh launch and a restored session. Please record one live smoke result covering both paths and confirming the active model. Once that evidence is present, I consider this ready to approve.

Ran the live smoke test — here's what I found.

**Fresh launch:**
```
droid --settings /tmp/ao-test-settings.json     # contents: {"model": "glm-5.2"}
```
Status bar shows `GLM-5.2 (Droid Core) (High)` as active. This differs from the model active in the session prior to the override (I saw Opus 5 and separately Fable 5 active in earlier sessions before testing this) — so the override is genuinely driving which model launches, not just being ignored in favor of the account default. This confirms `--settings`' top-level `"model"` key is honored on a fresh interactive launch, which is the primary path this PR targets.

**Restore:**
- `droid -r <session-id>` with **no** `--settings` flag already restored as `GLM-5.2` — Droid persists the model in its own per-session state file (`~/.factory/sessions/<project>/<id>.settings.json`) independent of `--settings`.
- `droid --settings /tmp/ao-test-settings.json -r <session-id>` also restored as `GLM-5.2` — no destabilization, but also no evidence `--settings` is doing anything on restore that Droid wasn't already going to do on its own.

**Net:** the launch path is confirmed live and is the one that matters for closing #2898 (a fresh worker/orchestrator session picking up its configured model). The restore path's use of `--settings` appears inert rather than incorrect — happy to open a follow-up note or drop `--settings` from `GetRestoreCommand` if you'd rather not carry dead weight, but didn't want to change behavior beyond what's tested without your input.

Environment: droid v0.185.0, macOS, Droid Core models (Fable 5 / Opus 5 as prior defaults, GLM-5.2 as override).

illegalcall (approved):
Re-reviewed current head 2be62f2 after the requested live verification.

The fresh interactive Droid smoke proves that the top-level model in the process-scoped settings file changes the active launch model (GLM-5.2 instead of the prior account defaults). Restore forwarding is covered by unit tests; the live restore result confirms no destabilization, while Droid's own persisted session model makes isolated restore precedence a follow-up rather than a blocker for #2898's launch requirement.

Fresh verification on this exact head: full backend go test ./... and go vet ./... pass, and all eight GitHub checks are green. Approving.

PR #3157, opened 2026-07-27
add customizable keybindings

aprv10:
## What

Add customizable application keyboard shortcuts while preserving the existing platform defaults.

Users can now:

- Open the keybinding editor from Settings or directly from the `Ctrl/Cmd+/` shortcut reference.
- Replace bindings or add up to two alternatives.
- Search commands by name, category, or key combination.
- Detect conflicts and explicitly reassign bindings.
- Remove or reset individual bindings.
- Reset all shortcuts to their defaults.
- Undo changes from an in-app confirmation toast.

The shortcut reference and sidebar hints display the effective user bindings.

## Why

Application shortcuts were previously hardcoded across the Electron main process and several renderer components. Users could not adapt them to their preferences, and displayed shortcut labels could drift from runtime behavior.

This provides one shared resolution path for built-in defaults and persisted user overrides.

## How

- Added shared default, effective-binding, matching, and display utilities.
- Persisted validated overrides atomically in `~/.ao/keybindings.json`.
- Passed live overrides through Electron shortcut interception, including browser preview web contents.
- Updated renderer-owned commands such as sidebar, command palette, project navigation, and inspector actions to use the shared matcher.
- Added a searchable keybinding editor with recording, conflict handling, alternatives, reset controls, hover tooltips, and toast feedback.
- Kept ordinary terminal and text-editing shortcuts outside the customizable application-command scope.
- Kept project `1–9` navigation fixed because it represents nine indexed actions rather than one rebindable command.

## Testing

- `npm run typecheck`
- `npx vitest run --config vite.renderer.config.ts src/shared/shortcuts.test.ts src/main/new-session-shortcut.test.ts src/main/keybinding-settings.test.ts src/renderer/components/KeyboardShortcutsDialog.test.tsx`
  - 39 tests passed.
- `npx vitest run --config vite.renderer.config.ts src/renderer/components/KeyboardShortcutsDialog.test.tsx`
  - 4 tests passed after adding the direct Customize action.

## Checklist

- [x] Branched from `main`
- [x] One focused change
- [x] Remote CI checks pass

Screenshot 2026-07-27 101053
Screenshot 2026-07-27 101128
Screenshot 2026-07-27 101208
Screenshot 2026-07-27 101257

### DIFF
--- a/frontend/src/main/app-shortcuts.ts
+++ b/frontend/src/main/app-shortcuts.ts
@@ -1,18 +1,14 @@
 import {
 	FOCUS_TERMINAL_SHORTCUT_CHANNEL,
 	KEYBOARD_SHORTCUTS_HELP_CHANNEL,
-	matchesFocusTerminalShortcut,
-	matchesKeyboardShortcutsHelpShortcut,
-	matchesNextSessionShortcut,
-	matchesNewSessionShortcut,
-	matchesNewShellTerminalShortcut,
-	matchesOpenSettingsShortcut,
-	matchesPreviousSessionShortcut,
+	matchesAppShortcut,
 	NEXT_SESSION_SHORTCUT_CHANNEL,
 	NEW_SESSION_SHORTCUT_CHANNEL,
 	NEW_SHELL_TERMINAL_SHORTCUT_CHANNEL,
 	OPEN_SETTINGS_SHORTCUT_CHANNEL,
 	PREVIOUS_SESSION_SHORTCUT_CHANNEL,
+	type AppShortcutId,
+	type KeybindingOverrides,
 	type ShortcutChord,
 } from "../shared/shortcuts";
 
@@ -44,14 +40,24 @@ type ShortcutTargetContents = {
 	send: (channel: string) => void;
 };
 
-const appShortcutChannel = (chord: ShortcutChord, isMac: boolean): string | null => {
-	if (matchesNewSessionShortcut(chord, isMac)) return NEW_SESSION_SHORTCUT_CHANNEL;
-	if (matchesNewShellTerminalShortcut(chord, isMac)) return NEW_SHELL_TERMINAL_SHORTCUT_CHANNEL;
-	if (matchesKeyboardShortcutsHelpShortcut(chord, isMac)) return KEYBOARD_SHORTCUTS_HELP_CHANNEL;
-	if (matchesOpenSettingsShortcut(chord, isMac)) return OPEN_SETTINGS_SHORTCUT_CHANNEL;
-	if (matchesPreviousSessionShortcut(chord, isMac)) return PREVIOUS_SESSION_SHORTCUT_CHANNEL;
-	if (matchesNextSessionShortcut(chord, isMac)) return NEXT_SESSION_SHORTCUT_CHANNEL;
-	if (matchesFocusTerminalShortcut(chord, isMac)) return FOCUS_TERMINAL_SHORTCUT_CHANNEL;
+const mainShortcutChannels: readonly [AppShortcutId, string][] = [
+	["new-session", NEW_SESSION_SHORTCUT_CHANNEL],
+	["new-shell-terminal", NEW_SHELL_TERMINAL_SHORTCUT_CHANNEL],
+	["keyboard-shortcuts", KEYBOARD_SHORTCUTS_HELP_CHANNEL],
+	["open-settings", OPEN_SETTINGS_SHORTCUT_CHANNEL],
+	["previous-session", PREVIOUS_SESSION_SHORTCUT_CHANNEL],
+	["next-session", NEXT_SESSION_SHORTCUT_CHANNEL],
+	["focus-terminal", FOCUS_TERMINAL_SHORTCUT_CHANNEL],
+];
+
+const appShortcutChannel = (
+	chord: ShortcutChord,
+	isMac: boolean,
+	overrides: KeybindingOverrides,
+): string | null => {
+	for (const [id, channel] of mainShortcutChannels) {
+		if (matchesAppShortcut(id, chord, isMac, overrides)) return channel;
+	}
 	return null;
 };
 
@@ -63,6 +69,7 @@ export function attachAppShortcuts(
 	isMac: boolean,
 	target: ShortcutTargetContents,
 	focusTarget = false,
+	getOverrides: () => KeybindingOverrides = () => ({}),
 ): void {
 	contents.on("before-input-event", (event, input) => {
 		if (input.type !== "keyDown" || input.isAutoRepeat) return;
@@ -76,6 +83,7 @@ export function attachAppShortcuts(
 				alt: input.alt,
 			},
 			isMac,
+			getOverrides(),
 		);
 		if (!channel) return;
 
--- a/frontend/src/main/app-shortcuts.ts
+++ b/frontend/src/main/app-shortcuts.ts
@@ -70,9 +70,13 @@ export function attachAppShortcuts(
 	target: ShortcutTargetContents,
 	focusTarget = false,
 	getOverrides: () => KeybindingOverrides = () => ({}),
+	isRecording: () => boolean = () => false,
 ): void {
 	contents.on("before-input-event", (event, input) => {
 		if (input.type !== "keyDown" || input.isAutoRepeat) return;
+		// Let the renderer's capture listener receive application-owned chords
+		// while the user is recording a replacement binding.
+		if (isRecording()) return;
 		const channel = appShortcutChannel(
 			{
 				key: input.key,
--- a/frontend/src/main/keybinding-settings.ts
+++ b/frontend/src/main/keybinding-settings.ts
@@ -1,0 +1,89 @@
+import { mkdir, readFile, rename, writeFile } from "node:fs/promises";
+import path from "node:path";
+import {
+	APP_SHORTCUTS,
+	type AppShortcutId,
+	type KeybindingOverrides,
+	type ShortcutBinding,
+} from "../shared/shortcuts";
+
+export const KEYBINDING_SETTINGS_FILE_NAME = "keybindings.json";
+
+const customizableIds = new Set<AppShortcutId>(
+	APP_SHORTCUTS.filter((shortcut) => shortcut.customizable !== false).map((shortcut) => shortcut.id),
+);
+
+let settingsOperationQueue: Promise<void> = Promise.resolve();
+
+function coerceBinding(raw: unknown): ShortcutBinding | null {
+	if (!raw || typeof raw !== "object") return null;
+	const value = raw as Record<string, unknown>;
+	if (typeof value.key !== "string" || value.key.length === 0 || value.key.length > 32) return null;
+	const code = typeof value.code === "string" && value.code.length > 0 && value.code.length <= 32 ? value.code : undefined;
+	const candidate: ShortcutBinding = {
+		key: value.key,
+		...(code ? { code } : {}),
+		ctrl: value.ctrl === true,
+		meta: value.meta === true,
+		shift: value.shift === true,
+		alt: value.alt === true,
+	};
+	// Application-wide shortcuts must not consume ordinary typing.
+	if (!candidate.ctrl && !candidate.meta && !candidate.alt && !/^F(?:[1-9]|1\d|2[0-4])$/.test(candidate.key)) return null;
+	return candidate;
+}
+
+export function coerceKeybindingOverrides(raw: unknown): KeybindingOverrides {
+	if (!raw || typeof raw !== "object") return {};
+	const source = raw as Record<string, unknown>;
+	const overrides: KeybindingOverrides = {};
+	for (const id of customizableIds) {
+		const rawBindings = source[id];
+		if (!Array.isArray(rawBindings)) continue;
+		const bindings = rawBindings
+			.slice(0, 2)
+			.map(coerceBinding)
+			.filter((candidate): candidate is ShortcutBinding => candidate !== null);
+		overrides[id] = bindings;
+	}
+	return overrides;
+}
+
+async function readUnlocked(stateDir: string): Promise<KeybindingOverrides> {
+	try {
+		const raw = await readFile(path.join(stateDir, KEYBINDING_SETTINGS_FILE_NAME), "utf8");
+		return coerceKeybindingOverrides(JSON.parse(raw));
+	} catch {
+		return {};
+	}
+}
+
+async function writeUnlocked(stateDir: string, overrides: KeybindingOverrides): Promise<KeybindingOverrides> {
+	const next = coerceKeybindingOverrides(overrides);
+	await mkdir(stateDir, { recursive: true, mode: 0o750 });
+	const file = path.join(stateDir, KEYBINDING_SETTINGS_FILE_NAME);
+	const tmp = path.join(stateDir, `.keybindings-${process.pid}-${Date.now()}.json`);
+	await writeFile(tmp, `${JSON.stringify(next, null, 2)}\n`, { mode: 0o600 });
+	await rename(tmp, file);
+	return next;
+}
+
+function runSettingsOperation<T>(operation: () => Promise<T>): Promise<T> {
+	const queued = settingsOperationQueue.then(operation, operation);
+	settingsOperationQueue = queued.then(
+		() => undefined,
+		() => undefined,
+	);
+	return queued;
+}
+
+export async function readKeybindingOverrides(stateDir: string): Promise<KeybindingOverrides> {
+	return readUnlocked(stateDir);
+}
+
+export async function writeKeybindingOverrides(
+	stateDir: string,
+	overrides: KeybindingOverrides,
+): Promise<KeybindingOverrides> {
+	return runSettingsOperation(() => writeUnlocked(stateDir, overrides));
+}
--- a/frontend/src/main/keybinding-settings.ts
+++ b/frontend/src/main/keybinding-settings.ts
@@ -2,6 +2,7 @@ import { mkdir, readFile, rename, writeFile } from "node:fs/promises";
 import path from "node:path";
 import {
 	APP_SHORTCUTS,
+	shortcutBindingValidationError,
 	type AppShortcutId,
 	type KeybindingOverrides,
 	type ShortcutBinding,
@@ -15,7 +16,7 @@ const customizableIds = new Set<AppShortcutId>(
 
 let settingsOperationQueue: Promise<void> = Promise.resolve();
 
-function coerceBinding(raw: unknown): ShortcutBinding | null {
+function coerceBinding(raw: unknown, isMac: boolean): ShortcutBinding | null {
 	if (!raw || typeof raw !== "object") return null;
 	const value = raw as Record<string, unknown>;
 	if (typeof value.key !== "string" || value.key.length === 0 || value.key.length > 32) return null;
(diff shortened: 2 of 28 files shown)


Pulkit7070 (commented):
Reviewed and stress-tested this. It is a large feature but well-architected, and I did not find a defect.

Architecture, which is the part that usually goes wrong for customizable keybindings:
- Single source of truth. The main process owns keybindingOverrides, persisted to ~/.ao/keybindings.json (atomic temp+rename, mode 0600, serialized writes, coerced/validated on read and write). The renderer store is a write-through cache: setBindings/resetBinding/resetAll all round-trip through aoBridge.keybindings.set and adopt the coerced result. There is no second persisted copy in the renderer, so main and renderer cannot drift.
- Main-process shortcut handling reads the overrides through a live getter (() => keybindingOverrides), so a rebind from the renderer immediately affects both renderer keydown matching and main-process/global shortcuts. No sync gap.
- Matching is correct: exact match on all four modifiers (so Ctrl+B does not match Ctrl+Shift+B), a code-vs-key path for physical-key bindings that compares code===code when both are present and only falls back to key otherwise, and a 1-9 special case for go-to-session. Override replaces default via ??, which also makes an empty override a valid unbound state while reset (deleting the key) restores the default.
- Conflict handling refuses to steal a non-customizable binding (toast) and otherwise prompts an explicit reassignment rather than silently double-binding.

Verification:
- Unit: main-process suites 16 pass (keybinding-settings, new-session-shortcut), renderer+shared suites 47 pass (shortcuts, KeyboardShortcutsDialog, SessionView). typecheck clean.
- Playwright against the real renderer: the editor opens from Settings, lists commands, and the search filters the list. The important interactive path works end to end: entering record mode on a command and pressing a chord captures it and persists through the bridge.

Honest scope notes, not defects:
- I did not drive the full conflict-then-reassign flow in the browser (it is covered by the code path above and the unit tests); I only confirmed the simple rebind capture-and-persist live.
- My reset-all assertion did not land cleanly because there are two Reset all controls (the trigger and its confirm), so that was a test-selector issue on my side rather than a product problem; the control is present and confirm-gated in the code.

No blocking issues found.

frontend/src/renderer/components/settings/KeyboardShortcutsSettingsDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+							>
+								<RotateCcw className="size-icon-base" aria-hidden="true" />
+								Reset all
+							</button>
+						)}
+					</div>
+				</DialogContent>
+			</Dialog>
+
+			{conflict ? (
```
illegalcall: **P1 (confidence 9/10): this conflict dialog is unreachable in the built app.**

It is a bare `<div>` rendered as a sibling of the Radix `<Dialog>`, not inside `DialogContent` and not portaled. Radix's `modal` defaults to `true`, and `DismissableLayer` sets `document.body.style.pointerEvents = "none"` while the dialog is open (`@radix-ui/react-dismissable-layer/dist/index.mjs:73`).

Verified with a throwaway jsdom test:

- `document.body.style.pointerEvents === "none"`, and this overlay has no `pointer-events-auto`, so Cancel and Reassign inherit `pointer-events: none`. **Not clickable.**
- It sits under an `aria-hidden="true"` ancestor: `screen.queryByRole("alertdialog")` returns `null` while `document.querySelector('[role="alertdialog"]')` finds it. **Not announced**, and outside Radix's `FocusScope` so **not tabbable**.
- No Escape handler. `setRecording(null)` has already run, so the only way out is closing the whole settings dialog.

The tell is line 456 just below: the toast *does* carry `pointer-events-auto`. This was hit once and patched in one place only.

Fix: reuse `ConfirmDialog`, or at minimum move this inside `DialogContent` and add `pointer-events-auto` plus an Escape handler.

frontend/src/main/app-shortcuts.ts
```diff
@@ -76,6 +83,7 @@ export function attachAppShortcuts(
 				alt: input.alt,
 			},
 			isMac,
+			getOverrides(),
```
illegalcall, line 90: **P1 (confidence 9/10): the recorder can never capture the seven chords this function owns.**

Line 90 calls `event.preventDefault()`, which stops the keydown from reaching the page at all. The recorder listens in the renderer (`KeyboardShortcutsSettingsDialog.tsx:195`), so it never sees these chords. There is no suppression path: grepping `recording|suppress|isRecording` across `main.ts`, `main/app-shortcuts.ts`, and `preload.ts` returns nothing.

While the recorder is armed and the user is looking at the row they want to change:

- `Cmd+/` opens the shortcut reference dialog on top of the settings dialog.
- `Cmd+,` fires open-settings. `Cmd+N` opens the New Task dialog over the settings dialog.
- `Cmd+Shift+T`, `Ctrl+PageUp`/`PageDown`, ``Ctrl+` `` all fire their commands.
- The recorder stays armed, because no candidate ever arrives.

Net effect: you cannot reassign a chord currently owned by a main-process command, and conflict detection is unreachable for exactly the seven commands most likely to collide.

Fix: a `keybindings:setRecording` IPC flipping a `let recordingActive` in main; have `appShortcutChannel` return `null` while true. ~10 lines, plus release on dialog close and window blur.

frontend/src/main/keybinding-settings.ts
```diff
@@ -0,0 +1,89 @@
...
+	const source = raw as Record<string, unknown>;
+	const overrides: KeybindingOverrides = {};
+	for (const id of customizableIds) {
+		const rawBindings = source[id];
+		if (!Array.isArray(rawBindings)) continue;
+		const bindings = rawBindings
+			.slice(0, 2)
+			.map(coerceBinding)
+			.filter((candidate): candidate is ShortcutBinding => candidate !== null);
+		overrides[id] = bindings;
```
illegalcall, line 50: **P2: a corrupt binding silently unassigns the command instead of falling back to its default.**

This line runs even when `bindings` is empty *because every entry failed validation*. `keybinding-settings.test.ts:19` asserts it as intended: `{"open-settings": [{key: "s"}]}` becomes `{"open-settings": []}`.

So one typo in a hand-edited `~/.ao/keybindings.json`, a truncated write, or a binding shape written by a newer build permanently disables that command with no diagnostic, indistinguishable from a deliberate "Unassigned".

Worth distinguishing the two cases:

```ts
const bindings = rawBindings.slice(0, 2).map(coerceBinding).filter(...);
if (rawBindings.length > 0 && bindings.length === 0) continue; // corrupt → fall back to default
overrides[id] = bindings;
```

Keep `[]` only when the raw array was genuinely `[]`.

frontend/src/renderer/components/settings/KeyboardShortcutsSettingsDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+								}}
+							>
+								Reassign
+							</button>
+						</div>
+					</div>
+				</div>
+			) : null}
+
+			{toast ? (
```
illegalcall: **P2: Undo is mouse-only and never announced.**

`pointer-events-auto` here makes clicks work (good catch), but this toast is still outside the Radix portal and under the same `aria-hidden="true"` ancestor as the conflict overlay. So `role="status" aria-live="polite"` never announces, and Tab never reaches the Undo button.

Undo is the only recovery path for an accidental rebind short of Reset all, so keyboard and screen-reader users lose it entirely. Portalling the toast fixes both this and the conflict overlay.

frontend/src/shared/shortcuts.ts
```diff
@@ -115,6 +129,113 @@ export function shortcutKeys(shortcut: ShortcutDefinition, isMac: boolean): read
...
+	if (key === "Up") return "arrowup";
+	if (key === "Down") return "arrowdown";
+	return key.toLowerCase();
+}
+
+export function matchesShortcutBinding(chord: ShortcutChord, candidate: ShortcutBinding): boolean {
+	const keyMatches =
+		candidate.key === "1-9"
+			? /^[1-9]$/.test(chord.key)
+			: candidate.code !== undefined
```
illegalcall: **P3: this makes `code` strictly win, dropping the old fallback for non-US layouts.**

The matcher this replaces was an OR:

```ts
const isBackquote = chord.code === "Backquote" || chord.key === "`";
```

Now when both `candidate.code` and `chord.code` are present, only the codes are compared. On a layout where backtick is not on the `Backquote` physical key, ``Ctrl+` `` regresses from working to not working.

Fix: keep the OR rather than the precedence, i.e. `codeMatches || keyMatches` instead of `code ? codeMatches : keyMatches`.

frontend/src/shared/shortcuts.ts
```diff
@@ -115,6 +129,113 @@ export function shortcutKeys(shortcut: ShortcutDefinition, isMac: boolean): read
...
+): ShortcutBinding => ({
+	key,
+	ctrl: false,
+	meta: false,
+	shift: false,
+	alt: false,
+	...modifiers,
+});
+
+export function defaultShortcutBindings(id: AppShortcutId, isMac: boolean): readonly ShortcutBinding[] {
```
illegalcall, line 116: **P2: this is now the second source of truth for defaults, and the two have already drifted.**

`APP_SHORTCUTS[].mac` / `.windowsLinux` (line 47) and `shortcutKeys` (line 128) have **zero production callers** after this PR; grep only finds `shortcuts.test.ts:132`. Display moved to `shortcutBindingKeys(effectiveShortcutBindings(...))`.

Existing drift:

- `previous-session` mac reads `["⌘", "Alt", "↑"]`, renderer prints `⌘⌥↑`.
- `open-project` reads `"1–9"` (en dash), renderer prints `1-9` (hyphen).

One shared resolution path was the stated goal of the PR. Either delete `mac`/`windowsLinux` + `shortcutKeys` and derive labels from `defaultShortcutBindings`, or add a test asserting the two agree for every id.

frontend/src/renderer/components/settings/KeyboardShortcutsSettingsDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+	mode: RecordingState["mode"];
+};
+type ToastState = {
+	title: string;
+	body?: string;
+	undo?: () => Promise<void>;
+};
+
+const isMac = isMacPlatform();
+const modifierKeys = new Set(["Alt", "Control", "Meta", "Shift"]);
```
illegalcall: **P3: `AltGraph` is missing, so tapping AltGr alone records a binding.**

On Windows/Linux, AltGr reports `ctrlKey && altKey` true with `key === "AltGraph"`. That is not in this set, so line 46's modifier guard passes and `eventBinding` returns `Ctrl+Alt+ALTGRAPH` from a single AltGr tap.

Add `"AltGraph"`, and probably `"CapsLock"`, `"NumLock"`, and `"Dead"` too.

frontend/src/main/keybinding-settings.ts
```diff
@@ -0,0 +1,89 @@
...
+	const candidate: ShortcutBinding = {
+		key: value.key,
+		...(code ? { code } : {}),
+		ctrl: value.ctrl === true,
+		meta: value.meta === true,
+		shift: value.shift === true,
+		alt: value.alt === true,
+	};
+	// Application-wide shortcuts must not consume ordinary typing.
+	if (!candidate.ctrl && !candidate.meta && !candidate.alt && !/^F(?:[1-9]|1\d|2[0-4])$/.test(candidate.key)) return null;
```
illegalcall: **P3: no reserved-chord guard.**

Bare `F1`–`F24` pass this check with no modifier, and nothing blocks `Ctrl+C` / `Ctrl+D` / `Ctrl+Z` either. Assign `F5` to a main-process command and `app-shortcuts.ts:90` will `preventDefault()` it in every terminal, so `htop`, `mc`, and vim F-key bindings die app-wide. Assign `Ctrl+C` and you lose terminal interrupt.

Reset all is a real escape hatch, so this is a foot-gun rather than a trap. But a small blocklist of terminal-critical chords is cheap insurance, and it would pair well with the "keeps ordinary terminal shortcuts out of scope" claim in the description.

frontend/src/renderer/components/settings/KeyboardShortcutsSettingsDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+	binding: ShortcutBinding;
+	mode: RecordingState["mode"];
+};
+type ToastState = {
+	title: string;
+	body?: string;
+	undo?: () => Promise<void>;
+};
+
+const isMac = isMacPlatform();
```
illegalcall: **P3: module-scope platform detection.**

Computing this once at import time makes the whole dialog untestable on the other platform. It is also inconsistent with the sibling `KeyboardShortcutsDialog`, which takes `isMac` as a prop (defaulted from `isMacPlatform()`).

A prop or `useMemo` would match that component and make the recording/conflict paths testable on both platforms, which matters given there are currently no tests for this file.

illegalcall (changes requested):
## Pre-Landing Review: 9 issues (2 critical, 7 informational)

Reviewed on an isolated worktree at `pr-3157`. `npm run typecheck` is clean and the 7 shortcut-related suites pass (115 tests). The shared-resolution refactor is a good shape, and the live-override plumbing through `before-input-event` is the right design; the new `new-session-shortcut.test.ts` case proves overrides apply without reattaching listeners.

But two of the four user-facing capabilities in the description do not work in a real build. Details inline; summary below.

**Scope: CLEAN.** Every changed file traces to the stated intent. State lands at `path.dirname(runFilePath())` = `~/.ao/keybindings.json`, so the `CLAUDE.md` hard rule holds. Nothing is written outside `~/.ao`.

---

### P1 (confidence 9/10) The conflict/reassign dialog is unreachable in the built app

`KeyboardShortcutsSettingsDialog.tsx:410`. The conflict overlay is a bare `<div role="alertdialog">` rendered as a sibling of the Radix `<Dialog>`, not inside `DialogContent` and not portaled. Radix's `modal` defaults to `true`, and `DismissableLayer` sets `document.body.style.pointerEvents = "none"` while open (`@radix-ui/react-dismissable-layer/dist/index.mjs:73`).

Confirmed all three failure modes with a throwaway jsdom test:

- `document.body.style.pointerEvents === "none"` while the settings dialog is open, and the overlay carries no `pointer-events-auto`. Cancel and Reassign inherit `pointer-events: none`, so they are **not clickable**.
- The overlay sits under an `aria-hidden="true"` ancestor. `screen.queryByRole("alertdialog")` returns `null` even though `document.querySelector('[role="alertdialog"]')` finds it. So it is **not announced**, and it is outside Radix's `FocusScope` so it is **not tabbable**.
- No Escape handler of its own. `setRecording(null)` has already run, so the user is left with a modal-looking overlay they can only dismiss by closing the whole settings dialog.

"Detect conflicts and explicitly reassign bindings" is a headline bullet in the description, and it is dead. The tell is line 456: the toast right below the overlay *does* carry `pointer-events-auto`, so this was hit once and patched in one place only.

Fix: reuse the existing `ConfirmDialog`, or at minimum move the overlay inside `DialogContent` and add `pointer-events-auto` plus an Escape handler.

### P1 (confidence 9/10) The recorder cannot capture any of the seven main-process-owned chords

`main/app-shortcuts.ts:90` calls `event.preventDefault()` inside `before-input-event`, which stops the keydown from ever reaching the page. The recorder listens in the renderer (`KeyboardShortcutsSettingsDialog.tsx:195`). There is no suppression path: grepping `recording|suppress|isRecording` across `main.ts`, `main/app-shortcuts.ts`, and `preload.ts` returns nothing.

So while the recorder is armed and the user is looking at the row they want to change:

- `Cmd+/` opens the shortcut reference dialog on top of the settings dialog.
- `Cmd+,` fires open-settings. `Cmd+N` opens the New Task dialog over the settings dialog.
- `Cmd+Shift+T`, `Ctrl+PageUp`/`PageDown`, and ``Ctrl+` `` all fire their commands.
- The recorder stays armed the whole time, because no candidate ever arrives.

You can never reassign a chord currently owned by a main-process command, which also means conflict detection is unreachable for exactly the seven commands most likely to collide. That compounds the P1 above.

Fix: a `keybindings:setRecording` IPC flipping a `let recordingActive` in main; `appShortcutChannel` returns `null` while true. Roughly 10 lines, plus release on dialog close and window blur.

### P2 A corrupt persisted binding silently unassigns the command instead of falling back to default

`main/keybinding-settings.ts:47`. See inline comment.

### P2 Toast Undo is mouse-only and never announced

`KeyboardShortcutsSettingsDialog.tsx:454`. See inline comment.

### P2 Two sources of truth for defaults, with nothing asserting they agree

`shared/shortcuts.ts:47` and `:128`. `APP_SHORTCUTS[].mac` / `.windowsLinux` and `shortcutKeys` now have **zero production callers** (only `shortcuts.test.ts:132`). Display moved to `shortcutBindingKeys(effectiveShortcutBindings(...))`. Every default is now declared twice, and they have already drifted: `previous-session` mac reads `["⌘", "Alt", "↑"]` while the renderer prints `⌘⌥↑`, and `open-project` reads `"1–9"` (en dash) while the renderer prints `1-9` (hyphen).

One shared resolution path was the stated goal. Delete `mac`/`windowsLinux` + `shortcutKeys` and derive labels from `defaultShortcutBindings`. If you would rather keep them, add a test asserting the two agree for every id.

### P3 Non-US layouts lose the code-vs-key fallback for the terminal chord

`shared/shortcuts.ts:194`. See inline comment.

### P3 AltGr is bindable on Windows/Linux

`KeyboardShortcutsSettingsDialog.tsx:42`. See inline comment.

### P3 No reserved-chord guard; bare F-keys are bindable

`main/keybinding-settings.ts:32`. See inline comment.

### P3 `isMac` is computed once at module load

`KeyboardShortcutsSettingsDialog.tsx:41`. See inline comment.

---

### Testing gap

486 lines of new UI with zero tests. Recording, conflict detection, reassignment, add-alternative, remove, per-binding reset, reset-all, and undo are all untested, and two of them are broken. The new tests cover pure functions (`coerceKeybindingOverrides`, `matchesAppShortcut`, live-override read in `attachAppShortcuts`) plus one button click.

A single component test would have caught the conflict dialog in minutes.

### Behavior changes worth adding to the description

1. `toggle-inspector` and `toggle-sidebar` previously accepted `metaKey || ctrlKey` on both platforms; they are now platform-exact. That is why `SessionView.test.tsx` had to flip `metaKey` to `ctrlKey`. Windows/Linux users pressing `Win+Shift+B` lose the inspector toggle.
2. `open-project` on macOS previously also accepted `Ctrl+1-9`; it is now `Cmd` only.
3. Genuine side fix worth calling out: `Cmd+Shift+B` previously toggled the sidebar *and* the inspector at once, because shadcn's listener ignored `shiftKey`. They are now exclusive.

### Verdict

Not landable as-is. The two P1s are about 30 lines of fixes combined.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

aprv10:
@illegalcall 
Thanks for the detailed review. Also, Leaving the current design as is for now; we’ll iterate on the UX in a follow up once the core behavior has settled. I addressed all of the inline comments:

### What changed and why

- Replaced the custom conflict overlay with the shared `ConfirmDialog`.
  - This keeps the confirmation inside Radix’s modal system, so pointer interaction, focus trapping, keyboard navigation, screen-reader visibility, and Escape cancellation work correctly.

- Moved the notification and Undo action inside the keyboard-shortcuts dialog.
  - The previous fixed-position sibling could be hidden by Radix’s accessibility isolation. It is now announced through a live region and its actions remain keyboard-accessible.

- Added a `keybindings:setRecording` bridge between the renderer and main process.
  - Main-process-owned shortcuts were being intercepted before the recorder could see them. Shortcut interception is now temporarily suspended while recording and restored after capture, cancellation, blur, dialog closure, unmount, or renderer termination.

- Changed persisted-binding recovery behavior.
  - An intentional empty array still means “unassigned,” while a non-empty but entirely invalid/corrupted value is omitted so the command safely falls back to its default.

- Restored logical-key fallback alongside physical-code matching.
  - Matching now succeeds when either `key` or `code` matches, preserving shifted-key handling without breaking non-US keyboard layouts.

- Removed the duplicate platform-label/default definitions.
  - `defaultShortcutBindings()` is now the single source of truth used for both runtime behavior and displayed shortcut labels, preventing future drift.

- Ignored AltGr, modifier-only, lock, dead, process, and unidentified key events.
  - These events should never become application shortcuts, particularly AltGr on Windows/Linux where it can appear as Ctrl+Alt.

- Added shared reserved-binding validation.
  - Bare function keys and terminal/OS-critical combinations such as interrupt, EOF, suspend, terminal copy/paste, macOS editing/window commands, and Alt+F4 are rejected because application shortcuts are intercepted before the terminal receives them.

- Made platform detection injectable in the settings dialog.
  - Runtime behavior still defaults to the current OS, but macOS and Windows/Linux behavior can now be tested deterministically.

### Validation

- `npm run typecheck`
- `npm run typecheck:e2e`
- 57 focused shortcut, persistence, preload, and dialog tests
- 32 BrowserView host tests

The new dialog tests cover recording, conflicts, Escape cancellation, reassignment, Undo, AltGr/reserved keys, blur cleanup, add/remove/reset/reset-all, and platform-specific bindings.

illegalcall (approved):

PR #2535, opened 2026-07-08
feat(frontend): migrate renderer to token-based design system

Annieeeee11:
## Summary
Migrates `frontend/src/renderer` to a token-based design system so styling is consistent, themeable, and easier to maintain.
- **Token foundation** — adds `src/styles/tokens.css` as the single source of truth for colors, typography, spacing, motion, icons, and z-index; wires tokens into Tailwind via `@theme` / `@utility` in `styles.css`
- **Legacy cleanup** — removes BEM dashboard chrome from `styles.css` (~1,450 lines deleted); global CSS now only covers xterm, keyframes, session-split, and resize global state
- **Shared components** — introduces `TopbarButton`, `StatusPill`, and `ResizeHandle` to replace ad-hoc BEM classes and inline patterns
- **Component adoption** — updates renderer components and shadcn `ui/*` primitives to use semantic utilities (`text-caption`, `h-control-form`, `size-icon-*`, `z-chrome`, etc.) instead of hardcoded values
- **Theme boot** — `index.html` sets `data-theme` before first paint to avoid FOUC; terminal palette reads `--color-term-*` via `terminal-themes.ts`
- **Docs** — adds `DESIGN_TOKENS.md`, `DESIGN_AUDIT.md`, and `MIGRATION_GAPS.md` (migration marked complete)
**Scope:** renderer only; `src/landing/` intentionally unchanged.
**Diff:** 46 files, ~984 insertions / ~2,122 deletions (net reduction driven by BEM removal).
## Test plan
- [ ] `npm run typecheck` passes
- [ ] `npx vitest run src/renderer` passes
- [ ] `npx vite build` succeeds
- [ ] App loads without theme flash on cold start (dark default, light toggle works)
- [ ] Visual spot-check: sidebar, sessions board, session inspector, topbar, dialogs/sheets, PR table, terminal pane, notification center
- [ ] Resize handles work between panes; no layout regressions on narrow/wide window sizes
- [ ] Status pills and kanban column tones still render correctly (runtime color tones)

PR #3159, opened 2026-07-27
style: tokenize desktop typography system

JAYATIAHUJA:
## What

Tokenizes the desktop renderer typography system around bundled Inter UI fonts and JetBrains Mono terminal/code fonts.

## Why

The app had a mix of system font fallbacks, half-step typography tokens, and Tailwind text aliases that made the UI feel less cohesive across platforms. This gives the renderer a stable font baseline while keeping legacy token aliases available for existing components.

## How

- Adds bundled Inter and JetBrains Mono WOFF2 assets with a font notice.
- Wires `@font-face` declarations into renderer CSS before Tailwind theme token usage.
- Consolidates the core typography ramp into role-based tokens while preserving existing alias tokens/classes.
- Updates the terminal default font comment to match the renamed 12px token.

## Testing

- `npm exec vite -- --config vite.renderer.config.ts build` passed from `frontend/` after sandbox escalation; build output includes the bundled font assets.
- `npm --prefix frontend run test` was attempted twice and failed only in `src/main/supervisor-link.test.ts` with 4 Windows socket `EACCES` failures; 1081 tests passed.
- `npm --prefix frontend run test -- src/renderer` was attempted and timed out with unrelated existing component test failures/EPIPE in this environment, so it is not listed as passing.

## Checklist


Hardik180704 (approved):
Tested this locally in the native Electron app on macOS at both the default window size and the minimum supported 960×640 size. The updated typography loads correctly, and I didn’t find any clipping, overflow, or sizing regressions in the sidebar, top bar, settings, or New Task dialog. Typecheck, frontend tests, renderer smoke tests, and the frontend build also pass. LGTM!

Hardik180704:
macOS sizing screenshots from the local Electron verification at the minimum supported 960×640 window size:

pr3159-mac-dashboard-960
pr3159-mac-new-task-960

Vaibhaav-Tiwari (changes requested):
resolve the connflicts else looks good

JAYATIAHUJA:
Resolved the merge conflict with current upstream/main and pushed the updated branch. The PR is now mergeable and all visible checks are passing; the only remaining block is the stale requested-changes review.

Pulkit7070 (commented):
Reviewed the technical side and it checks out.

- Font wiring: eight @font-face blocks (Inter and JetBrains Mono, weights 400-700), all font-display: swap, all src: url("./assets/fonts/*.woff2") format("woff2"). Relative bundled paths, no CDN, which is the right call for the Electron renderer (offline plus CSP). Vite fingerprints and copies these on build.
- Licensing: the NOTICE file gives correct SIL OFL 1.1 attribution for both families (Fontsource latin-subset woff2, repackaged unmodified). Good.
- Token consolidation is non-breaking. I diffed tokens.css and computed the set difference of removed vs added token names: it is empty, i.e. every token name that changed is re-added (re-valued), and only new role tokens (font-size-lg/micro/title, text-lg/md/title) are introduced. So no token was deleted, which means there are no dangling var(--...) references, matching the "preserving existing alias tokens/classes" claim. Consistent with @Hardik180704's local pass finding no clipping/overflow/sizing regressions.

On the block: the only outstanding review is a CHANGES_REQUESTED asking to resolve conflicts. The branch is now mergeable (it carries a merge from upstream/main at the head), so mergeable = MERGEABLE. That conflict request looks already addressed; it just needs a re-review or dismissal to unblock, there is no code change pending from it.

One note, not a defect: bundling Inter and JetBrains Mono is a visual/design decision that should track DESIGN.md and the agent-orchestrator baseline. That is a look-and-feel call rather than a correctness one, and a maintainer already approved the rendered result, so I am flagging it only for the record, not asking for a change.

No defects found.

frontend/src/styles/tokens.css
```diff
@@ -85,29 +85,33 @@
...
-	/* Scale — dense UI chrome uses half-steps between base sizes */
-	--font-size-2xs: 10.5px; /* uppercase meta, board footers */
-	--font-size-xs: 10px; /* micro labels, tight pills */
-	--font-size-caption: 11px; /* badges, kanban labels */
-	--font-size-sm-md: 11.5px; /* inspector tabs, status pills */
-	--font-size-sm: 12px; /* secondary chrome / Tailwind text-xs */
-	--font-size-md-sm: 12.5px; /* kv rows, PR summaries */
-	--font-size-base: 13px; /* dense controls (board buttons) */
-	--font-size-brand: 16px; /* titlebar / board project name */
+	/* Scale - consolidated IDE-grade ramp (VS Code density: 13px chrome, 14px body). */
```
Pulkit7070, line 96: Minor doc inaccuracy: the comment says VS Code density: 13px chrome, 14px body, but the ramp has no 14px step (micro 10, caption 11, xs 12, sm 13, lg 15, heading 17, title 20). Either add the 14px body token or drop 14px body from the comment so it matches the actual scale.

frontend/src/styles/tokens.css
```diff
@@ -85,29 +85,33 @@
...
-	--font-size-heading-sm: 17px; /* empty-state headings */
-	--font-size-heading: 21px; /* dashboard subhead */
-	--font-size-heading-lg: 22px; /* preview mock headings */
-
+	--font-size-lg: 15px; /* dialog titles, stat values, titlebar */
+	--font-size-heading: 17px; /* section / empty-state headings */
+	--font-size-title: 20px; /* dashboard subhead, page titles */
+
+	/* Backward-compatible aliases for existing component classes. */
+	--font-size-2xs: var(--font-size-caption);
```
Pulkit7070, line 107: These aliases keep the old names but repoint them to new values, so components still using them change size. Tracing the pixels: --font-size-2xs 10.5px to 11px, --font-size-sm-md 11.5px to 11px, --font-size-md-sm 12.5px to 12px, --font-size-brand 16px to 15px, --font-size-heading-lg 22px to 20px (subtitle 15px and heading-sm 17px are unchanged). Also note --font-size-xs itself moved 10px to 12px on line 99, so any direct consumer of --font-size-xs grows 2px while the old 10px semantics move to the new --font-size-micro. None of this is a build break, but please confirm these size shifts are intended rather than an accidental side effect of the rename, since the PR text frames the aliases as compatibility shims.

frontend/src/styles/tokens.css
```diff
@@ -335,7 +339,7 @@
 	--color-text-settings-row: rgb(236 236 236);
 	--color-border-settings: var(--color-border-welcome-panel);
 	--color-border-settings-menu: rgb(66 66 66 / 0.78);
-	--font-size-settings-title: 28px;
+	--font-size-settings-title: var(--font-size-title);
```
Pulkit7070, line 342: Biggest single shift: --font-size-settings-title goes from 28px to var(--font-size-title) = 20px, an 8px / ~29% reduction on the settings title. Please confirm that is intended. @Hardik180704 said settings looked fine in local testing, so it is likely deliberate, but a 29% drop on a title is large enough to call out explicitly.

Pulkit7070 (commented):
Went through the diff line by line. The font wiring, the component token swaps (a coherent topbar densification: text-sm to text-xs, h-control-lg to h-control-form), the design-tokens.ts comment re-point (keeps the JS constant at 12px, now correctly citing --font-size-xs), and the NOTICE are all correct.

One correction to my earlier summary, which the line read makes clear: the consolidation is build-safe (I confirmed no token name is deleted, so no dangling var refs), but the backward-compatible aliases preserve names while CHANGING values. So this does re-size existing components, it is not size-neutral. That is presumably the intent (a consolidated ramp) and @Hardik180704 visually verified it, but it is worth being explicit because a reader could take preserving existing alias tokens to mean sizes are unchanged. Inline comments mark the notable shifts to confirm. No defects, three confirm-intended notes.

frontend/src/styles/tokens.css
JAYATIAHUJA: The 14px body step is present as `--font-size-md: 14px; /* body / Tailwind text-sm */`, with `--text-md` mapped to it in the Tailwind theme. The comment is intentional: 13px is the default chrome density, while 14px remains the body/content token.

frontend/src/styles/tokens.css
JAYATIAHUJA: Confirmed intended. The compatibility shim here is name/API compatibility, not pixel neutrality: existing `text-*`/`var(--font-size-*)` consumers keep resolving, but several aliases intentionally repoint onto the consolidated ramp. The `--font-size-xs` growth and the move of the old 10px semantics to `--font-size-micro` are deliberate parts of the typography tokenization.

frontend/src/styles/tokens.css
JAYATIAHUJA: Confirmed intended. Settings is being pulled onto the shared `--font-size-title` token so the desktop renderer uses one title scale instead of keeping a standalone 28px settings-only title. @Hardik180704 also covered settings in the local Electron sizing pass, so I am keeping this as the intended consolidated hierarchy.

Hardik180704 (approved):
Re-reviewed the current head after the upstream merges. The renderer suite (824 tests), typecheck, and production Vite build pass locally, and the bundled Inter/JetBrains Mono assets are emitted correctly. The token changes are intentional size adjustments with existing names preserved. No new blocking issues found. The older CHANGES_REQUESTED review appears to refer only to conflicts that are now resolved.

JAYATIAHUJA:
Resolved the latest conflict with current upstream/main and pushed the merge commit. The PR is now mergeable again; visible checks are green (Frontend test, renderer-smoke, and gitleaks scan), and I also reran `npm --prefix frontend run typecheck` locally after resolving `ShellTopbar.tsx`.

@Vaibhaav-Tiwari could you please re-review or dismiss the stale changes-requested review when you get a chance?

frontend/src/renderer/styles.css
```diff
@@ -13,22 +70,25 @@
 	--font-sans: var(--font-family-base);
 	--font-mono: var(--font-family-mono);
 
-	/* Typography — align Tailwind text-* with token scale */
-	--text-xs: var(--font-size-sm);
-	--text-sm: var(--font-size-md);
-	--text-base: var(--font-size-md);
+	/* Typography - align Tailwind text-* with the consolidated token scale. */
+	--text-xs: var(--font-size-xs);
+	--text-sm: var(--font-size-sm);
```
Pulkit7070, line 75: Worth calling out explicitly: this remaps Tailwind `text-sm` and `text-base` from 14px to 13px app-wide (both previously resolved to `--font-size-md`=14px, now to `--font-size-sm`=13px). Since `text-sm` is used almost everywhere in a shadcn/Tailwind renderer, this is the broadest-surface change in the PR, and it was not in the earlier itemized shift list (which traced only the `--font-size-*` aliases). Same applies to `--font-size-command-input` going 16px to `--font-size-lg`=15px. Assuming the 13px chrome density is intended, no change needed, just confirming it is deliberate rather than a side effect of the ramp rename.

Rishet11:
The approach here was replaced by the Geist token foundation (#2535 and #3421), the assets/fonts directory is gone from main, and the changes requested in review were never addressed. Closing as stale.

PR #3164, opened 2026-07-27
chore(landing): rename workspace scope to @ao/* and align site metadata

codebanditssss:
Standardizes the landing's internal package scope and brings the site's generated metadata in line with our brand.

## Packages
- rename the internal workspace scope to `@ao/*` — package names in `packages/*`, ~40 imports, and the lockfile; workspaces relinked so it resolves cleanly

## Site metadata & content
- CLI installer (`public/cli/install.sh`) → uses our Homebrew tap
- `llms.ts` (drives `llms.txt` / `agents.md` / agent-card / api-catalog / mcp routes) → our brand, domain, and GitHub; dropped SDK/skills links we don't ship
- marketplace theme copy, analytics-consent cookie name, JSON-LD `sameAs`, mock preview data, `.gitignore` comments

## Renderer/docs
- tidy stale design-reference comments (`styles.css`, `badge`, `button`, an e2e comment, `CLAUDE.md`)

## Note
- `DESIGN.md` still has older design-lineage notes; left as-is pending a decision on whether to keep that history.

## Verification
- landing static export builds (71 pages)


codebanditssss:
## Summary

Aligns the workspace scope and site metadata across the landing package and cleans up stale wording in a few renderer/docs files. No behavior change to the app itself.

**Scope**
- Renames the internal workspace scope to `@ao/*` across the landing package (package manifests + all imports) and the lockfile.
- Updates site metadata and generated routes to point at our own URLs, GitHub org, and Homebrew tap.
- Rewrites the CLI install script to install via our Homebrew tap.
- Refreshes marketplace theme JSON and mock/demo copy to our own naming.
- Neutralizes the design-lineage wording in `DESIGN.md` and a handful of renderer comments/e2e labels; every design spec (palette, type, spacing, layout, motion) and the live "clone agent-orchestrator verbatim" direction are unchanged.

**Verification**
- Full landing build passes (71 pages) after the scope relink.
- No data path to any external service: analytics is env-key driven and stays uninitialized when the key is unset; the only external URL that remained was literal text in generated metadata, never fetched.

59 files, +158 / -280.

Pulkit7070 (commented):
Reviewed. This is a clean, mostly-mechanical branding and rename chore, and I verified the parts that could actually break something.

- Rename completeness: the @superset/* to @ao/* rename is complete. I checked out the head and grepped the landing tree for @superset and found zero remaining references, and package-lock.json is regenerated. So no dangling import should break the workspace resolve. One caveat: I do not see a landing build in the PR's CI checks (doctor, scan, React Doctor are the ones that ran), so the 71-page static export you mention is the local build evidence plus the grep, not a gated check.
- install.sh rewrite reads well: it drops the vendored installer for brew install agentwrapper/tap/agent-orchestrator, and if Homebrew is absent it errors with a clear message and a link to the desktop release. Reasonable scope change.
- Renderer and docs edits are comment-only (removing the Superset and emdash design-lineage notes from CenterPane, badge, button, styles.css, plus the CLAUDE.md and DESIGN.md text neutralization). No logic changes, so nothing to run there.

One specific thing to confirm before this lands, because it is user-facing: the domain. This PR points the landing at aoagents.dev consistently (install.sh header, llms.ts API_URL api.aoagents.dev, docs at aoagents.dev/docs, and the curl -fsSL https://aoagents.dev/cli/install.sh line). But README.md on main uses ao-agents.com (with a hyphen) for the docs and install links. Those cannot both be the live site. Since this change ships a public curl | sh installer and the docs URLs in llms.txt/agents.md, a wrong host means a 404 for anyone who runs it. Please confirm aoagents.dev is the canonical domain; if it is, README should be updated to match in a follow-up, and if it is not, install.sh and llms.ts are pointing at a dead host. The PR is internally consistent, so this is about which domain is real, not about this diff contradicting itself.

No other issues.

codebanditssss:
Thanks for the careful pass, @Pulkit7070 — and good catch on the domain.

Confirmed: **aoagents.dev is the canonical live domain** (it's what the shared brand constants in `packages/shared/src/constants.ts` resolve to, and what this PR is consistent with). So `install.sh` and `llms.ts` are pointing at the right host — nothing to change here.

The hyphenated `ao-agents.com` refs in `README.md` and the docs `.mdx` files (changelog, migration) are the stale ones. Those are out of scope for this rename chore, so I'll clean them up in a separate follow-up PR rather than widen this diff.

On the landing build: agreed it's local evidence (grep + the 71-page static export), not a gated check — noting that for the record.

Pulkit7070:
Re-review: my one open item is resolved.

The domain question I raised is settled. aoagents.dev is confirmed canonical (packages/shared/src/constants.ts defines MARKETING_URL/DOCS_URL/etc. on aoagents.dev), so this PR's use of aoagents.dev in install.sh and llms.ts is correct, not a dead host. The stale ao-agents.com references lived only in the root README and two docs pages, which the follow-up #3165 fixes; the two PRs together retire ao-agents.com entirely and don't overlap.

Everything else still checks out on the current head (bbd2ab5a5): the . to @ao rename is complete (zero leftover references on this branch), the renderer and docs edits are comment-only, and CI is now fully green (test, doctor, scan, renderer-smoke, React Doctor).

Nothing outstanding from me on this PR.

Annieeeee11 (approved):
lgtm

somewherelostt:
Thanks for the cleanup here. One small follow-up I noticed after reviewing the merged diff:

`frontend/src/landing/src/lib/constants.ts` changes `ANALYTICS_CONSENT_KEY`, while `frontend/src/landing/src/instrumentation-client.ts` only checks the new key before disabling analytics capture. That can reset the saved preference for returning visitors who had previously declined analytics, so they may be treated as having no saved preference until they opt out again.

Could we preserve or migrate the previous declined value before capture is allowed? Everything else I checked looked consistent, and the PR checks are green.

PR #3167, opened 2026-07-27
fix(agy): honor cfg.Config.Model via --model flag on launch/restore

### LINKED ISSUE #2901: Agy (Antigravity): verify model-override support and honor role-specific model config

- Original report: #2769

## Decision / verification needed

The **Agy** (Antigravity) adapter (`agy`, binary `agy`) does not forward `cfg.Config.Model`. Whether the `agy` CLI even accepts a model override is **unverified** — no public evidence was found. This issue is to verify and then fix if a mechanism exists.

## Findings

- `GetLaunchCommand` (`backend/internal/adapters/agent/agy/agy.go:65-88`) emits `agy --add-dir <ws> [--dangerously-skip-permissions] [--prompt-interactive <prompt>]`; `GetRestoreCommand` (`:90-116`) uses `--conversation <id>`. No model handling. Inherits empty `GetConfigSpec`.
- The adapter's flags mirror the Claude-Code flag style (`--add-dir`, `--dangerously-skip-permissions`), which *suggests* a `--model` flag may exist by analogy — but this was **not confirmed** in docs.

## Next step

Run `agy --help` (and check Antigravity CLI docs) for a model-selection flag, env var, or config key.
- **If a launch-time `--model` (or similar) exists:** apply the mechanical Codex-style fix (helper appending the flag in launch + restore, plus `GetConfigSpec` `model` field + tests).
- **If only an interactive/config mechanism exists:** treat like the config-route harnesses (kilocode/vibe).
- **If none exists:** close as not-supported (like Amp) and document `agentConfig.model` as a no-op for Agy.

## References

- Relevant file: `backend/internal/adapters/agent/agy/agy.go`

anmol0b:
## What
Forwards the per-role model override (`cfg.Config.Model`) to the `agy` (Antigravity) CLI as a `--model` flag on both session launch and session restore, and declares a `GetConfigSpec` entry for it.

## Why
Fixes #2901.

AO resolves a per-role model and forwards it to every adapter as `cfg.Config.Model`, but each adapter has to opt in to actually passing it to its CLI. The `agy` adapter never did, so a configured worker/orchestrator silently fell back to agy's default model instead of the one set in `agentConfig.model`.

## How
- Verified via `agy --help` that a launch-time flag exists (`--model  Model for the current CLI session`) — this was previously unconfirmed per the issue, so it settles the "does agy even support this" question in favor of applying the mechanical fix.
- Mirrors the pattern from the Codex adapter fix (#2869): an `appendModelFlag(cmd *[]string, cfg ports.AgentConfig)` helper that appends `--model <trimmed value>` only when set, called from both `GetLaunchCommand` and `GetRestoreCommand`.
- Added `GetConfigSpec` (previously inherited empty from `agentbase.Base`), declaring the `model` field with the same shape used by the Codex/Copilot adapters, so the config layer/UI know this adapter now supports a model override.
- Intentional omission worth flagging: unlike the Copilot adapter (#2895), which deliberately does *not* forward `--model` on restore pending verification of `--model` + `--resume` composition, this PR forwards on both launch **and** restore. I found no equivalent warning for `agy` in `--help` or available docs, so I didn't add the same restriction — but a reviewer with a real `agy` install should sanity-check this.

## Testing
```
go build ./...
go test ./internal/adapters/agent/agy/...
golangci-lint run ./internal/adapters/agent/agy/...
```
All pass, 0 lint issues. Covered:
- `TestGetLaunchCommand` — model set, flag appears in the right position.
- `TestGetLaunchCommandNoModel` — model unset, flag omitted entirely.
- `TestGetRestoreCommand` — model set, flag appears alongside `--conversation`.
- `TestGetConfigSpecReportsModelField` — config spec reports the `model` field.

Not yet verified locally: an actual `agy --model <x> --print "..."` run against a real `agy` binary, to confirm the flag changes model selection at runtime rather than just being accepted without erroring.

## Checklist
- [x] Branched from `main` (`fix/agy-model-override`)
- [x] One focused change; links the related issue (#2901)
- [x] Relevant CI checks pass for the area touched (`go`, frontend, etc.) — pending CI run after push

### DIFF
--- a/backend/internal/adapters/agent/agy/agy.go
+++ b/backend/internal/adapters/agent/agy/agy.go
@@ -59,10 +59,28 @@ func (p *Plugin) Manifest() adapters.Manifest {
 	}
 }
 
+// GetConfigSpec reports the per-project agent config keys the Agy (Antigravity)
+// CLI understands: a model override passed via --model. Confirmed via
+// `agy --help`: "--model  Model for the current CLI session".
+func (p *Plugin) GetConfigSpec(ctx context.Context) (ports.ConfigSpec, error) {
+	if err := ctx.Err(); err != nil {
+		return ports.ConfigSpec{}, err
+	}
+	return ports.ConfigSpec{
+		Fields: []ports.ConfigField{
+			{
+				Key:         "model",
+				Type:        ports.ConfigFieldString,
+				Description: "Model override passed to `agy --model` (e.g. gemini-3-pro).",
+			},
+		},
+	}, nil
+}
+
 // GetLaunchCommand builds the argv to start an interactive Agy session.
 // Shape:
 //
-//	agy --add-dir <WorkspacePath> [--dangerously-skip-permissions] [--prompt-interactive <Prompt>]
+//	agy --add-dir <WorkspacePath> [--dangerously-skip-permissions] [--model <Model>] [--prompt-interactive <Prompt>]
 func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (cmd []string, err error) {
 	binary, err := p.agyBinary(ctx)
 	if err != nil {
@@ -79,6 +97,8 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 		cmd = append(cmd, "--dangerously-skip-permissions")
 	}
 
+	appendModelFlag(&cmd, cfg.Config)
+
 	if cfg.Prompt != "" {
 		cmd = append(cmd, "--prompt-interactive", cfg.Prompt)
 	}
@@ -87,7 +107,7 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 }
 
 // GetRestoreCommand rebuilds the argv that continues an existing Agy session:
-// `agy --add-dir <WorkspacePath> [--dangerously-skip-permissions] --conversation <agentSessionId>`.
+// `agy --add-dir <WorkspacePath> [--dangerously-skip-permissions] [--model <Model>] --conversation <agentSessionId>`.
 func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig) (cmd []string, ok bool, err error) {
 	if err := ctx.Err(); err != nil {
 		return nil, false, err
@@ -112,6 +132,8 @@ func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig)
 	if cfg.Permissions == ports.PermissionModeBypassPermissions {
 		cmd = append(cmd, "--dangerously-skip-permissions")
 	}
+
+	appendModelFlag(&cmd, cfg.Config)
 
 	cmd = append(cmd, "--conversation", agentSessionID)
 	return cmd, true, nil
@@ -156,4 +178,15 @@ func (p *Plugin) agyBinary(ctx context.Context) (string, error) {
 	}
 	p.resolvedBinary = binary
 	return binary, nil
+}
+
+// appendModelFlag appends `--model <trimmed>` when cfg.Model is set,
+// mirroring the Codex adapter's pattern (see #2869). A blank or
+// whitespace-only value is omitted so agy falls back to its own default
+// model resolution exactly as an unconfigured launch would. Confirmed via
+// `agy --help`: "--model  Model for the current CLI session".
+func appendModelFlag(cmd *[]string, cfg ports.AgentConfig) {
+	if model := strings.TrimSpace(cfg.Model); model != "" {
+		*cmd = append(*cmd, "--model", model)
+	}
 }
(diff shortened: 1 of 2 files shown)


illegalcall (commented):
Reviewed on a clean worktree at `9386bc5b8`. `go test ./internal/adapters/agent/agy/...` passes locally.

**Verdict: approve.** Cleanest of the three model-config PRs (#3155 crush, #3156 droid, this one). It follows the existing in-repo pattern exactly, the tests cover both directions, and the one real open question is flagged honestly in the PR body.

## The restore-path question — I think you made the right call, and for a better reason than the PR gives

The PR argues "I found no equivalent warning for `agy` in `--help` or available docs, so I didn't add the same restriction." That's an argument from absence. There's a stronger one available:

Copilot's omission (`copilot.go:154-157`, `:351-358`) isn't "we know `--model` breaks `--resume`". Reading the comment, it's "an attempt to verify was **inconclusive** — the test account is on the Copilot Free plan, which only grants `auto` regardless of `--model`, on both launch and resume. That makes the plan itself a confound, not evidence either way." So that restriction is about an experiment that failed to produce a signal, not a known incompatibility.

Meanwhile **codex** — the actual precedent for a verified `--model` flag — calls `appendModelFlag(&cmd, cfg.Config)` on both launch (`codex.go:117`) and restore (`codex.go:159`). Since agy has a real launch-time `--model` and no known resume interaction, following codex rather than copilot is the defensible default. Worth putting that reasoning in the PR body so nobody later "corrects" it to match copilot.

## Verified

- `appendModelFlag(cmd *[]string, cfg ports.AgentConfig)` is byte-for-byte the shape used by codex (`codex.go:377`) and copilot (`copilot.go:359`). Yes, that's now three copies of a five-line helper, but it matches the established per-adapter convention and AGENTS.md says follow existing package boundaries — so I wouldn't extract it in this PR. Might be worth an `agentbase` helper once a fourth adapter needs it.
- The `--model` flag does exist on agy: independent references corroborate it (e.g. [Antigravity CLI guides](https://www.aibuilderclub.com/blog/antigravity-cli-guide)). Google's own [CLI reference](https://antigravity.google/docs/cli/reference) documents slash commands, keybindings and `settings.json` keys but **not** command-line flags, so `agy --help` remains the only authoritative source — which is why your having run it matters. Please keep that quote in the code comment; it's currently the only record.

## 1. Verify the `gemini-3-pro` example (low)

The config-spec description says `(e.g. gemini-3-pro)`. Third-party docs show launch examples in display-name form — `agy --model="Gemini 3.5 Flash"` — and describe the accepted values as "the stable, user-facing model slugs that appear in the `/model` picker". I couldn't confirm which of the two forms `--model` actually takes, or that `gemini-3-pro` is a live slug.

No functional risk either way — Go's `exec` passes argv directly, so a value with spaces needs no quoting and won't break. But since `GetConfigSpec` descriptions are meant to be user-facing guidance, an example that isn't a real accepted value is worse than no example. Please confirm against the `/model` picker on a real install, or drop the parenthetical.

## 2. The runtime verification gap you flagged (low)

Agreed, and I'd keep it visible: `agy --help` accepting `--model` proves the flag parses, not that it changes model selection — and it says nothing about `--model` + `--conversation`. A single `agy --model <x> --print "what model are you?"` plus one resume with a different model would close both. Worth a follow-up issue rather than blocking this, since the fix is mechanical and correct-by-pattern regardless.

## 3. `TestGetLaunchCommand` changed meaning (nit)

The original body moved into the new `TestGetLaunchCommandNoModel`, and `TestGetLaunchCommand` now covers the model case. Fine, but `TestGetLaunchCommandWithModel` / `TestGetLaunchCommandNoModel` would read better as a pair — right now the "default" name is the non-default case.

## 4. `TestGetConfigSpecReportsModelField` asserts the exact Description string (nit)

`reflect.DeepEqual` against the full `ConfigField` means any wording tweak to a user-facing help string fails the test. Asserting `Key` and `Type` (as the crush and droid tests in this batch do) would be less brittle. Minor, and arguably a feature if you want the string pinned.

## Cross-cutting (same note left on #3155 and #3156)

`GetConfigSpec` has **no production caller** — grepping `backend/` finds 15 adapter implementations, `ports/agent.go:40`, and test doubles, nothing else. So the `Description` added here is never rendered anywhere. Not this PR's problem, but it means all three PRs in this batch are writing documentation into a dead surface. Probably worth its own issue.

Also: this PR and #3156 treat `AgentConfig.Model` as a bare id, while #3155 makes crush require `<provider>/<model-id>`. Same shared field, three adapters, two contracts — a note in `domain/agentconfig.go` would save the next person the archaeology.

illegalcall (approved):
Approving — follows the codex pattern exactly, tests cover both directions, and the one open question is flagged honestly in the PR body.

On the restore path: I think you landed correctly, but there's a stronger argument than "no equivalent warning for agy". Copilot's omission isn't a known incompatibility — its own comment says the verification was **inconclusive** because the test account's Free plan forces `auto` regardless of `--model`. Codex, the actual verified precedent, forwards on both launch and restore. Worth putting that in the PR body so nobody later "corrects" this to match copilot.

Two small things, neither blocking: confirm `gemini-3-pro` is a real accepted value before shipping it as the config-spec example (Google's CLI reference documents slash commands and `settings.json` but not flags, and third-party docs show display-name form), and file the runtime-verification follow-up you mention so `--model` + `--conversation` doesn't stay untested indefinitely.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #3168, opened 2026-07-27
test(frontend): reuse sidebar sizing constants

Hardik180704:
## What changed

- reuse the sidebar sizing constants in resize and collapsed-state tests
- derive drag coordinates and expected persisted widths instead of repeating numeric width values
- keep the existing resize behavior and coverage unchanged

## Why

This is a small follow-up to #3148. The tests added there repeated sidebar width assumptions, which could drift if the production sizing values change later.

## Verification

- Sidebar suite: 41 tests passed
- full frontend suite: 1,096 tests passed
- renderer smoke: 20 tests passed
- frontend typecheck
- E2E typecheck
- `git diff --check`


Annieeeee11 (approved):
lgtm

PR #3169, opened 2026-07-27
fix(board): legible agent logos in both themes + decluttered task cards

codebanditssss:
Two board fixes from the design thread.

## 1. Agent logos legible in both themes
Several agent marks ship as a single colour — white-only (opencode, cursor, cline, goose, kilocode) or black-only (copilot) — so they vanished against one theme (white on the light board, black on the dark board). Those marks now render as inline `currentColor` SVGs so they take the board's foreground colour and stay legible in **both** light and dark. Full-colour brand logos are untouched. Also drops the placeholder Copilot **octocat** for its real mark.

## 2. Decluttered task cards
Each column header already names the stage (Working / Needs you / In review / Ready to merge), so the per-card status pill just repeated it ("Working" under Working, "Approved" under Ready to merge, …). Removed it entirely. The card now carries only its own identity + code state — agent, title, branch, PR(s), **diff size**, updated time — in one footer row instead of the old two-row split that left blank gaps.

## Scope
- `AgentAvatar.tsx` — inline currentColor for the 6 mono marks; colour logos unchanged; removes `copilot.png`
- `SessionsBoard.tsx` — card drops the status pill; single meta footer row with PR + diff + time
- `SessionsBoard.test.tsx` — assert cards carry no status pill and still route to the right column

No token or column-header changes. Verifiable via `npm run dev:web` (toggle light/dark).

image


Annieeeee11 (approved):
lgtm

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -780,11 +780,18 @@ function SessionCard({
...
+	// own identity + code state: agent, title, branch, PRs, diff, updated time.
 	const badge = getSessionStatusView(session.status);
 	const issueId = canonicalTrackerIssueId(session.issueId);
 	const branch = session.branch || "";
 	const showBranch = branch !== "" && !sameLabel(branch, session.title) && !sameLabel(branch, session.id);
 	const prSummaries = sessionPRDisplaySummaries(session, useSessionScmSummary(session.id).data);
+	const changed = session.changedFiles ?? [];
+	const additions = changed.reduce((total, file) => total + file.additions, 0);
+	const deletions = changed.reduce((total, file) => total + file.deletions, 0);
+	const showDiff = changed.length > 0 && additions + deletions > 0;
```
codebanditssss, line 794: The new diff summary will not render in the packaged app. `session.changedFiles` is only present in mock data; `fetchWorkspaces()` does not populate it from the production sessions API, so `changed` is always empty. Please derive the totals from `prSummaries`, or add diff data to the sessions API and map it in `useWorkspaceQuery.ts`. Please also add a production-shaped test that verifies the `+additions` / `−deletions` values appear.

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -780,11 +780,18 @@ function SessionCard({
 	onTerminate?: () => void;
 	interactive?: boolean;
 }) {
+	// The column header already names the stage (Working / Needs you / In review /
+	// Ready to merge), so the card never repeats a status pill. It carries only its
+	// own identity + code state: agent, title, branch, PRs, diff, updated time.
```
codebanditssss, line 785: These status labels are not always redundant with the column header. "Needs you" groups Input needed, Exited, No signal, CI failed, and Changes requested, while "In review" also groups multiple distinct states. Removing the label makes users unable to tell why a task needs attention, and removes that information for screen readers. Please retain the specific status on mixed-state columns, visually and accessibly; it can still be omitted where it genuinely duplicates the column, such as Working.

PR #3177, opened 2026-07-27
feat(board): agent mark registry, one card system, and lane/archive chrome

codebanditssss:
Re-lands #3169 (merged then reverted in #3172) and grows well past it. Preview with `npm run dev:web`, toggle light/dark.

Five things, all on the board.

## Agent marks come from a registry

Single-colour brands vanished against one theme. The first pass fixed that by inlining each brand's SVG path as a string in `AgentAvatar` — 76 lines of path data, and the `.svg` files on disk were only half the truth.

`lib/agent-logos.ts` is now the single source: one entry per harness, pointing at the asset and saying how it has to be painted. `AgentAvatar` only renders. Adding a harness is one line in one file.

- **No background plates.** A logo on its own brand-coloured tile fights every surface it lands on, so the plate is stripped from the asset rather than hidden with CSS. `codex` keeps its gradient with the white plate path removed; `aider`, `droid` and `continue` are cleaned in place.
- **Theme-adaptive.** Single-colour marks are drawn as a mask filled with `currentColor`, so one asset stays legible on both boards instead of disappearing into one.
- Plated rasters are replaced with the upstream vectors the existing `codex` and `cursor` marks already came from — no new dependency, the files are vendored.
- `crush` deliberately keeps its background: it's an illustration whose trail bleeds off-frame, not a mark on a plate, so there's nothing to strip without a proper vector.

## Live and archived sessions share one card

The archive drew its own card — status and time on a top row, agent under the title, a different divider inset. Sitting right under the lanes, it read as a different product rather than the same board in a quieter state.

Archived sessions now use the live card's shape. Two things the live card doesn't need are kept: the **status**, because no column header names the stage down there, and **restore**, moved to the corner the live card puts Terminate in.

Task cards gain the copy button the archive already had, and name their agent in the meta row when there's no PR, diff or issue — that row was otherwise empty but for a timestamp across the whole Working lane.

## Lanes stop reserving space they aren't using

`flex-[3]`/`flex-[2]` gave the upper lane 60% of the column whatever it held, so three idle cards left a gap before the Working heading. Lanes are now content-sized under one shared scroller.

The heading also said things twice — it named both halves while the section below repeated one of them, and named an Idle lane even when nothing was idle. It now names only the lane the column starts with.

## Headings read as headings

They looked like another card footer: 10.5px, no separation from the stack under them. Now a band of their own — tint, bottom border, a colour bar in the status hue — with the title at 12px semibold and the count in a chip.

## Archive is capped and resizable

It opened at every session it held and pushed the lanes off screen. Now it opens at ~2 rows and scrolls, with a drag handle on its top edge (pointer and keyboard, persisted). Its grid is pinned `content-start`, because a grid stretches its rows to fill the container by default, which inflated every card as the panel grew instead of leaving the new room empty.

Board scrollbars were invisible: `--color-scrollbar` is `0.12` alpha and shared with the diff/file scrollbars, and `color-mix` with transparent can only *lower* alpha — no percentage could brighten it. The lanes get their own token pair; other scrollbars are untouched.

## Also

- Per-card status pill dropped entirely (author design call), with the status kept in each card's accessible name for screen readers.
- Diff size derives from the PR summaries populated by the SCM API, not the mock-only `session.changedFiles`, so it renders in the packaged app.
- Sidebar search field radius matched to the cards; label uses a new `--font-size-nano` token rather than a hardcoded value.

## Merged with main

Takes main's **grid-only archive** (#3211) — the rows/grid toggle is gone, so the unified card, height cap and drag handle all apply to the grid alone. Keeps main's orchestrator activity indicator (#3200) and daemon startup loader (#3205).

## Checks

34 files, +662/−279. Typecheck clean, **1189 unit tests pass, zero failures**. CI green: `test`, `renderer-smoke`, `scan`.

Test assertions that encoded the old layout (`flex-[3]`, dot classes, `0 working sessions`) were rewritten to the new contract rather than deleted.


codebanditssss:
Note on the status-pill review point (raised on #3169): after design review, the author's call is to **drop the status pill on every card**. The four columns already name the stage, and the mixed-column labels (Changes requested / CI failed / Review pending) read as clutter next to the title. The card keeps its identity + code state — agent, title, branch, PR link, and diff — and the PR/CI lifecycle in the PR line still conveys much of the "why".

The other review point **is** addressed: the diff now derives from the PR summaries (SCM API), not the mock-only `session.changedFiles`, so it renders in the packaged app — covered by a new production-shaped test.

Happy to revisit the accessibility angle (e.g. sr-only status) if reviewers feel strongly.

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -780,11 +780,20 @@ function SessionCard({
 	onTerminate?: () => void;
 	interactive?: boolean;
 }) {
+	// The column header already names the stage (Working / Needs you / In review /
+	// Ready to merge), so the card carries no status pill — only its own identity +
```
nikhilachale, line 849: [P1] I understand the design decision to remove the visible status pill, but the card now also loses that information for screen-reader users. The “Needs you” lane can represent Input needed, Exited, No signal, CI failed, or Changes requested, and the PR line does not communicate all of these states. Please preserve badge.label as visually hidden text (sr-only) or include it in the card’s accessible name. This keeps the cleaner visual design while still explaining why each task needs attention.

codebanditssss: Done in 97c9346 — the status stays visually hidden but is now announced to screen readers: an `sr-only` "Status: {badge.label}" sits after the title, so the accessible name reads e.g. "Resolve reviewer feedback… Status: Changes requested … PR #318 open". Visual design is unchanged (still no pill). Added a test asserting the sr-only status is present while the visible pill is not.

neversettle17-101 (commented):
I found two issues worth addressing:

1. Empty Ready-to-merge columns are labeled as `Merged` in the split-lane header. In `frontend/src/renderer/components/SessionsBoard.tsx`, `SplitLaneColumn` chooses `secondaryTone` whenever `primarySessions.length === 0`. For the merge column, when both ready and merged sessions are empty, that makes the board show `Merged 0` instead of the expected `Ready to merge 0`, so the four columns read Working / Needs you / In review / Merged on common empty-ready boards. Please keep the empty merge lane on the ready/default tone, or pass an explicit empty/default tone per split lane.

2. Persisted archive height can bypass the viewport cap after reload. `initialArchiveHeight()` accepts any stored value above the minimum, while the max cap is only applied when committing a new drag value. If the user later opens the app on a smaller window/display, `style={{ height: resolvedArchiveHeight }}` can restore an over-tall archive and bring back the behavior this PR is trying to avoid. Please clamp the stored value against `archiveMaxHeight()` when reading it, and ideally re-clamp on window resize.

Issue #2631, opened 2026-07-12
SCM PR attribution relies purely on branch-name convention with no cryptographic session tie — collision risk widened by #2623


mukesher4:
## Summary

Found while reviewing #2623 (cross-fork PR attribution fix) locally. Not a regression introduced by that PR's logic, but the PR does widen an existing, pre-#2623 weakness's surface area — filing separately since it's a distinct concern from what #2609/#2623 fix.

**Core issue**: AO's PR-to-session attribution (`discoverNewPRs`/`matchSession` in `backend/internal/observe/scm/observer.go`) has no cryptographic or authoritative link between a PR and the session it gets attributed to. Attribution is entirely a branch-name-prefix match against the session's namespace (`ao/<session-id>/...`), gated only by a repo-level filter (`candidatesForHeadRepo`). Nothing verifies that the branch was actually created by that session's own agent.

## Concrete scenario

Session `agent-orchestrator-6` has namespace `ao/agent-orchestrator-6` (from its working branch `ao/agent-orchestrator-6/root`) and a legitimate PR from a `fork` remote: branch `ao/agent-orchestrator-6/terminal-focus-fix`.

Now suppose a **second, unrelated** PR exists with the exact same branch name text, but living in a different repo — e.g. someone with push access to `origin` itself creates a branch literally named `ao/agent-orchestrator-6/terminal-focus-fix` and opens a PR from it (same-repo, not cross-fork). Scanning `origin`'s PR list finds this PR; `candidatesForHeadRepo` matches it against the `(to=origin, from=origin)` pair (which always exists — same-repo PRs were never the broken case); `matchSession` then matches purely on the branch name text and attributes it to `agent-orchestrator-6`, with zero verification that this worker's own agent created it.

## Confirmed: this predates #2623, but #2623 widens where it can happen

Traced the pre-fix code specifically to verify: `sessionRepo{repo: repo, headRepo: origin}` — when `repo == origin`, this produces exactly `{repo: origin, headRepo: origin}`, the same-repo pair, which existed unmodified in the original, broken code before #2623 touched anything. So a same-repo branch-name collision (as in the scenario above) was **already possible before #2623**.

[...]

PR #3180, opened 2026-07-27
fix(reviews): improve review display

neversettle17-101:
## Summary

- show only the latest available AO review run, using the previous run as the fallback when a current run has not started yet
- avoid showing an expanded `Not run` badge once a PR has been reviewed at least once
- remove the separate pull request reviews block from the Reviews tab

## Tests

- `cd frontend && npm test -- SessionInspector.test.tsx`


AgentWrapper:

AgentWrapper:
Reopening — apologies, this was closed in error during a cleanup of our own orchestrator's sessions. Your PR comes from a different AO install whose session numbering collides with ours on the ao/agent-orchestrator-53/* branch namespace (the exact misattribution problem tracked in #2631). The PR itself is untouched and welcome.

neversettle17-101 (commented):
Needs changes. The validation path looks aligned, but the SCM round-trip dedup is too broad: it keys delivered AO review coverage by author and then removes every non-bot review-thread comment from that author. That can hide later actionable feedback from the same GitHub account on the same PR, including a separate/manual review or a new thread not tied to the delivered review ID. Please narrow the filtering to comments/threads that are actually associated with the remembered review ID, or persist enough per-comment/thread identity to make that distinction.

backend/internal/lifecycle/reactions.go
```diff
@@ -379,6 +392,93 @@ func (m *Manager) notificationIntentForCurrentSCM(ctx context.Context, id domain
...
+	return coverage, nil
+}
+
+func filterCoveredReviewThreads(threads []ports.SCMReviewThreadObservation, coveredAuthors map[string]bool) []ports.SCMReviewThreadObservation {
+	filtered := make([]ports.SCMReviewThreadObservation, 0, len(threads))
+	for _, th := range threads {
+		next := th
+		next.Comments = make([]ports.SCMReviewCommentObservation, 0, len(th.Comments))
+		for _, c := range th.Comments {
+			if !c.IsBot && coveredAuthors[strings.TrimSpace(c.Author)] {
```
neversettle17-101: This removes comments by any author that has one remembered AO review on the PR, not comments that are known to belong to that delivered review. After one delivered review from `alice` is remembered, any later non-bot review-thread comment by `alice` on the same PR is filtered out here; if those are the only unresolved comments, the observation can be downgraded to `ReviewNone` below and the worker never gets that feedback. Please filter by review/comment/thread identity tied to the delivered GitHub review ID instead of by author alone.

neversettle17-101: Addressed in fdbd25586: the SCM round-trip dedup no longer filters by reviewer author. GitHub review-thread comments now carry their owning submitted review ID from `pullRequestReview.id`, and lifecycle suppresses only comments whose `ReviewID` matches a remembered delivered AO GitHub review ID. I also added lifecycle coverage for same-author feedback from a different review staying visible, plus adapter coverage that the GitHub query/parser preserves the comment review ID.

Vaibhaav-Tiwari (changes requested):
1. Dedup probably won’t work in real GitHub reviews.
     The reviewer submits GitHub REST .id, but the SCM observer later compares it to GitHub GraphQL id. Those are different ID types. Result: AO may still re-
     send the same reviewer feedback to the worker after it sees the review from GitHub.

  2. It filters by author too broadly.
     After AO records one delivered review, the new code drops all later review comments from the same author, not just comments from that specific delivered
     review. If the same GitHub account leaves another real comment later, AO may hide that feedback from the worker.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -1096,35 +1100,64 @@ function ReviewPanel({
 	);
 }
 
+function aoReviewMeta(reviewState: PRReviewState): string {
+	if (reviewState.latestRun?.createdAt) {
+		return `#${reviewState.prNumber} · ${formatTimeCompact(reviewState.latestRun.createdAt)}`;
+	}
+	if (reviewVerdict(reviewState).label === "Not run") {
+		return `#${reviewState.prNumber} · Not run`;
```
neversettle17-101, line 1027: When `latestRun` is absent but `previousRun` exists, `AoReviewRow` below displays the previous run verdict/body/link, but this metadata path still returns `#<n> · Not run` because it only checks `latestRun` and the aggregate `reviewVerdict`. In the fallback state the header can say `Not run` while the expanded row shows `Changes requested`/`Approved`, which contradicts the PR goal of using the previous run as the displayed fallback and avoiding the stale Not run state once a review exists. Please base this metadata on the same display run (`latestRun ?? previousRun`) or omit `Not run` when a previous run is being shown.

neversettle17-101 (commented):
Needs changes. The new review display mostly matches the intent, but the row metadata still reports a PR as `Not run` even when the expanded row is intentionally falling back to a previous AO review. That makes the UI show contradictory state for exactly the fallback case this PR adds.

frontend/src/renderer/components/SessionInspector.tsx
neversettle17-101: Addressed in ea3839403: `aoReviewMeta` now uses the same display run as the expanded row (`latestRun ?? previousRun`). If a previous run is being shown as the fallback, the header uses that run timestamp instead of reporting `Not run`; `Not run` is only emitted when there is no display run. Added a SessionInspector regression assertion for the previous-run fallback metadata.

neversettle17-101 (commented):
Ready to merge. I reviewed the updated frontend-only diff for the Reviews tab. The previous-run fallback now uses consistent row metadata, the separate pull request review summaries block has been removed as described, and the tests cover the not-run metadata, fallback display/link, running-state suppression of previous output, and removal of the PR review summaries section. I did not find any blocking correctness or regression issues in the reviewed changes.

Vaibhaav-Tiwari (approved):

PR #3181, opened 2026-07-27
fix(claudecode): remove ~/.claude.json trust entry on workspace teardown

zenprocess:
## What

Adds `CleanupWorkspace` on the claudecode adapter so each spawned worktree's trust entry is removed from `~/.claude.json` when the workspace is actually torn down. The matching add (PreLaunch) was already wired; the symmetric removal was missing.

## Why

Claude Code's PreLaunch records `projects[workspacePath].hasTrustDialogAccepted = true` in `~/.claude.json` so its first-run "trust this folder" dialog does not block the headless spawn. Nothing was removing those entries on teardown.

On a long-running daemon the `projects` map grew without bound (measured: 1,256 entries, of which 1,186 -- 94% -- pointed at directories that no longer existed). `~/.claude.json` therefore grows without bound for the lifetime of the install, and AO is the writer that creates the entries on each spawn, so AO should remove them on teardown. Unbounded growth of a config file the tool writes per session is the bug on its own terms; this PR makes the cleanup symmetric with the registration so the file size stays bounded by the number of currently-live workspaces rather than the total ever-spawned.

A search of the issue tracker did not surface a prior report for this accumulation. The closest existing reports discuss the *missing*-entry side of the same dialog (entries never registered, so the dialog blocks the first spawn). This PR addresses the symmetric, opposite failure: entries registered correctly, never removed. Happy to file a tracking issue first if maintainers prefer.

## How

Mirrors the cursor adapter's existing cleanup template (`backend/internal/adapters/agent/cursor/hooks.go:153-164` and `removeCursorWorkspaceTrust`):

- `Plugin.CleanupWorkspace(ctx, ports.WorkspaceHookConfig) error` -- public entry point, derives the config path via `claudeConfigPath()` and delegates to the helper.
- `removeWorkspaceTrustEntry(cfgPath, workspacePath)` -- reads the config, deletes `projects[workspacePath]`, writes back via the same `os.CreateTemp` + `os.Rename` atomic pattern `ensureWorkspaceTrusted` already uses (`backend/internal/adapters/agent/claudecode/claudecode.go:517-581`), reusing `claudeTrustMu` so a concurrent PreLaunch on another workspace cannot race the removal.
- Missing config file or missing entry is a no-op (no rewrite, mtime preserved) rather than an error, so a torn-down workspace that was never trusted does not surface a spurious failure.

**Why the entry is removed whole, not just the trust flag.** AO is the only writer of the entry -- `ensureWorkspaceTrusted` created it -- so leaving an empty `{hasTrustDialogAccepted: true}` shell behind would still consume a deny-path slot in the sandbox profile and defeat the point of the cleanup.

**Path normalization: not needed on this side, by construction.** The trust entry is keyed by the raw `cfg.WorkspacePath` string in both `ensureWorkspaceTrusted` (PreLaunch) and `removeWorkspaceTrustEntry` (CleanupWorkspace), with no `filepath.Abs` or `filepath.Clean`. The cursor adapter normalizes via `filepath.Abs` on both sides (`cursor/hooks.go:220, 294`) because it derives a *filename* from the path (sanitized into a per-project subdir), so the absolute path is needed for the matching side too. The claudecode adapter stores the path as a map key, used identically on both sides, so normalization would only matter if the same logical workspace could arrive spelled differently on the two paths. It cannot: both call sites hand the SAME `ws.Path` field through `Manager.prepareWorkspace` (`session_manager/manager.go:384, :1043`) and `Manager.cleanupAgentWorkspace` (`session_manager/manager.go:2640-2667`), so the add and remove paths are guaranteed to see the identical string.

## Intentional omissions

**No adapter-level test for the still-present-workspace case.** The guarantee does not live in `removeWorkspaceTrustEntry` -- the helper operates on the JSON file directly and does not consult the filesystem. It lives in `Manager.Kill` (`session_manager/manager.go:770-802`), which calls `cleanupAgentWorkspace` only on the success branches (the `cleaned` and `freed=true` arms). The `ErrWorkspaceDirty` branch (line 788-797 for standalone worktrees, 773-779 for workspace projects) returns `false, nil` early without reaching `cleanupAgentWorkspace`, so a workspace that still has dirty state on disk is never silently removed from the trust list. That wiring is unchanged in this PR and is already exercised by the 191 passing `session_manager` tests. Adding a second adapter-level test for a guarantee enforced by a different layer would duplicate coverage and risk drifting from the real dispatch path.

## Testing

```
cd backend && go build ./... && go test ./internal/adapters/agent/claudecode/... && go test ./internal/session_manager/... && go vet ./...
```

- `go build ./...` -- Success
- `go vet ./...` -- No issues found
- `go test ./internal/adapters/agent/claudecode/...` -- 69 passed (includes 6 new tests)
- `go test ./internal/session_manager/...` -- 191 passed (existing dispatch unchanged; the `workspaceCleaner` call site is wired into the pre-existing `cleanupAgentWorkspace`, not added fresh)

**New tests** (in `backend/internal/adapters/agent/claudecode/claudecode_test.go`, mirroring the existing `ensureWorkspaceTrusted*` fixture shape):

| Test | Asserts |
| --- | --- |
| `TestRemoveWorkspaceTrustEntryRemovesTrackedEntry` | entry removed; unrelated projects + top-level keys preserved |
| `TestRemoveWorkspaceTrustEntryAbsentEntryIsNoOp` | absent entry -> no error, mtime unchanged (no rewrite) |
| `TestRemoveWorkspaceTrustEntryMissingConfigIsNoOp` | missing `~/.claude.json` -> no error, file stays missing |
| `TestRemoveWorkspaceTrustEntryEmptyConfigIsNoOp` | empty file -> no error, mtime unchanged |
| `TestRemoveWorkspaceTrustEntryConcurrencySafety` | 32 concurrent cleanups of one entry leave a valid JSON file with the entry gone |
| `TestTrustEntryRegisterCleanupCycleIsBalanced` | **adversarial**: 25 full `ensureWorkspaceTrusted` -> `removeWorkspaceTrustEntry` cycles using the production helpers (not a hand-seeded map); asserts the projects map returns to its pre-cycle size and member set every iteration -- bounded-but-growing would still leak entries forever, so the assertion is exact-set-equality, not just size floor |

**Dirty-refusal verification.** `Manager.Kill` (`session_manager/manager.go:770-802`) calls `cleanupAgentWorkspace` only on the success branches (the `cleaned` and `freed=true` arms). The `ErrWorkspaceDirty` branch (line 773-779 for workspace projects, line 788-797 for standalone worktrees) returns `false, nil` without calling cleanup. Adapter-side, `removeWorkspaceTrustEntry` operates on the config file directly and does not consult the filesystem, so a still-present workspace directory does not change its behavior -- but the manager-level guarantee means the function is never even called for a dirty workspace.

## Checklist

- [x] Branched from `main` (sibling under session namespace)
- [x] One focused change
- [x] Follows `AGENTS.md` conventions and PR hygiene (conventional commit, narrow tests first)
- [x] Tests added for user-visible behavior
- [x] Relevant CI checks pass for the area touched (`go build`, `go vet`, `go test ./internal/adapters/agent/claudecode/...`, `go test ./internal/session_manager/...`)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/backend/internal/adapters/agent/claudecode/claudecode.go
+++ b/backend/internal/adapters/agent/claudecode/claudecode.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -213,6 +214,41 @@ func (p *Plugin) PreLaunch(ctx context.Context, cfg ports.LaunchConfig) error {
 		return err
 	}
 	return ensureWorkspaceTrusted(cfgPath, cfg.WorkspacePath)
+}
+
+// CleanupWorkspace removes the durable trust entry AO recorded for a session
+// workspace. It is the symmetric counterpart of PreLaunch: PreLaunch adds
+// projects[cfg.WorkspacePath] to ~/.claude.json so Claude Code's trust
+// dialog does not block the headless spawn, and CleanupWorkspace removes it
+// once the workspace is actually torn down.
+//
+// Without this, every spawned worktree leaves a permanent entry behind and
+// ~/.claude.json accumulates indefinitely. Claude Code's sandbox profile
+// builder derives one deny-path per known project dir from this file, so
+// unbounded growth eventually pushes the profile past ARG_MAX and every
+// subsequent Bash spawn fails with E2BIG.
+//
+// The manager invokes CleanupWorkspace only after the workspace Destroy
+// succeeds — its ErrWorkspaceDirty branch returns early without calling it,
+// so a workspace that still has dirty state on disk is never silently
+// removed from the trust list. A missing config or a missing entry is a
+// no-op, not an error, so a torn-down workspace that was never trusted (or
+// was already cleaned up) does not surface a spurious failure.
+func (p *Plugin) CleanupWorkspace(ctx context.Context, cfg ports.WorkspaceHookConfig) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if strings.TrimSpace(cfg.WorkspacePath) == "" {
+		return errors.New("claudecode.CleanupWorkspace: WorkspacePath is required")
+	}
+	cfgPath, err := claudeConfigPath()
+	if err != nil {
+		return err
+	}
+	if err := removeWorkspaceTrustEntry(cfgPath, cfg.WorkspacePath); err != nil {
+		return fmt.Errorf("claudecode.CleanupWorkspace: %w", err)
+	}
+	return nil
 }
 
 // GetRestoreCommand rebuilds the argv that continues an existing Claude Code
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

zenprocess:
Heads up that the three workflow runs on this PR (Go, CLI E2E, gitleaks) are all currently in the `action_required` state — that's why no checks show up and the PR is showing `mergeable_state: blocked`.

`action_required` is the state GitHub assigns to workflows queued for first-time-contributor approval; it clears as soon as an approving maintainer approves the run. Nothing in the diff needs to change for this to resolve — purely a run-approval gate.

zenprocess:
@illegalcall @pulkit7070 — three runs (Go, CLI E2E, gitleaks) still sit in action_required; first-time-contributor approval is the only thing keeping the PR from acquiring a CI signal and clearing mergeable_state. The diff itself doesn't need any further changes for that to resolve — would a maintainer be willing to approve the queued workflow runs so this can be evaluated? Appreciate the on-call coverage on this thread.

zenprocess:
Hey @illegalcall @pulkit7070 — flagging that the branch is 88 commits behind `main`. Once those runs get approved they may surface failures from drift rather than from the change. We are able to rebase and force-push on our side whenever it would help — just say the word.

zenprocess:
Gentle ping for a workflow-approval run, not for review.

This PR appears to be held at `action_required` — check-runs have started but none have been allowed to conclude, which is the first-time-contributor workflow-approval gate rather than anything about the change itself. As far as I can tell that is only clearable by a maintainer.

If a maintainer could approve the workflow run, CI can actually report and this becomes reviewable. If the PR needs a rebase, or is not wanted in its current form, I am glad to do either — I just wanted to check it is not sitting purely on the approval gate unnoticed.

No urgency, and thanks for the project.

i-trytoohard:
Closing as stale: blocked on maintainer workflow approval for three weeks. @team if the workflow change is wanted, say so and we reopen; otherwise this stays closed.

PR #3105, opened 2026-07-25
fix: hold terminal input until a fresh pane's agent TUI is ready

ikeshavvarshney:
## Summary

Fixes #3023 — terminal accepted keystrokes as soon as `tmux attach` succeeded, before the launched agent's TUI had entered raw mode. Fast typing right after opening a fresh session landed as raw, echoed text in the pane's scrollback instead of the agent's input box, and was unrecoverable there.

### Root cause

Attach-readiness and agent-readiness are two different things. `tmux` forwards bytes to the pane's foreground process the moment attach succeeds — regardless of whether that process has actually read stdin yet. A full-screen TUI agent (Claude Code, Codex, ...) can take a moment after launch to enter raw mode, and until it does, the underlying line discipline echoes typed characters straight into the terminal's canonical-mode output.

### Fix

Hold input on a pane's first-ever attach for a short grace window (500ms), reusing the existing `pendingInput` buffering in `attachment.write()` — no new buffering logic needed. The window is tracked per pane id in `Manager` (`inputHoldDeadline`), keyed off the first time this daemon process sees that id, so:

- A pane's genuine first attach (freshly launched agent) waits.
- Any later reattach, tab reopen, or second client viewing the same pane reuses the already-ticking (or already-expired) deadline — never re-delays.
- Fully configurable/disableable via `WithInputGraceWindow` (defaults to 500ms, tests disable it for determinism).
- Output streaming is untouched — only accepting input is delayed, so the user still sees the agent boot in real time.

## Changes

- `backend/internal/terminal/manager.go` — `inputHoldDeadline` first-seen tracking per pane id, `WithInputGraceWindow` option, default 500ms grace window.
- `backend/internal/terminal/attachment.go` — `setPTY` waits out the hold (if any) on a pane's first attach before flipping `inputReady`, cancellable via `ctx` so shutdown/close is never blocked.
- `backend/internal/terminal/manager_test.go` — two new tests (`TestServeFreshPaneHoldsInputUntilGraceWindowElapses`, `TestServeReattachSkipsInputGraceWindow`) plus `WithInputGraceWindow(0)` added to existing timing-sensitive tests to keep them deterministic.

## Test plan

- `go build ./...`
- `go test ./internal/terminal/...` — full suite (33 tests) passes, including the two new grace-window tests.
- `gofmt` clean.
[...]

PR #3103, opened 2026-07-25
feat(desktop): open Option-clicked links in system browser

nightcityblade:
## What

Open web links in the OS default browser when they are clicked with Option/Alt, both in terminal output and across regular app anchors.

## Why

This provides an escape hatch from link behavior inside AO while preserving each surface’s existing plain-click behavior.

Fixes #3101.

## How

Terminal URL and OSC 8 handlers use the modifier to route web URLs through the validated `app.openExternal` bridge. A shared renderer policy handles regular anchors at the shell boundary in the bubble phase, after element handlers have run; it only intercepts unhandled Option/Alt-clicks on external HTTP(S) links. Plain anchor clicks continue through their existing handlers/window-open policy, while terminal plain clicks continue to use the AO Browser panel.

URL parsing is guarded and reads the `href` attribute so SVG and malformed anchors are safe. Bridge failures are caught and logged instead of becoming unhandled rejections.

## Testing

- `npm run frontend:typecheck` — passed
- `npm --prefix frontend run typecheck:e2e` — passed
- focused link, terminal, and shell suites — 63 passed
- full frontend Vitest suite — 88 files / 1,102 tests passed

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; links the related issue when applicable
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense
- [x] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

PR #3185, opened 2026-07-27
fix(terminal): write the initial replay as one chunk behind a cover

### LINKED ISSUE #3160: Terminal visibly auto-scrolls from mid-session position on session open
## Bug Description

When opening any worker session, the terminal briefly shows a mid-session position and then visibly auto-scrolls down to the end. This happens every time a session is opened or re-attached, and feels janky/broken to users.

## Root Cause

xterm.js processes large `write()` data **asynchronously in incremental batches** (~15KB write-buffer chunks), rendering intermediate buffer states between batches. The scrollback/handshake replay arrives as a single burst of PTY bytes but xterm splits it into chunks, rendering each chunk before processing the next. The cursor (and viewport) advances through the replay top-to-bottom across multiple animation frames, producing the visible scroll.

The timeline on first session open:

1. **Frontend opens** (`useTerminalSession.ts:195`): `mux.open()` + `mux.resize()` fire immediately
2. **Backend attaches** (`attachment.go:137`): `src.Attach()` spawns `tmux attach-session` (macOS/Linux) or dials the conpty host (Windows)
3. **Runtime replays scrollback**:
   - **tmux**: `tmux attach-session` replays the pane's full screen state
   - **conpty**: `Ring.Snapshot()` (`conpty/ring.go:61`) returns up to **1000 lines** of stored output, sent as the first `MsgTerminalData` frame (`conpty/host.go:253-264`)
4. **`copyOut()`** (`attachment.go:185`) pumps bytes in 32KB chunks → `onData` → WS frames → frontend `terminal.write(bytes)` (`useTerminalSession.ts:212`)
5. **xterm's write buffer** processes the replay in ~15KB batches via `setTimeout` scheduling. Between batches, the renderer draws the current buffer state — the viewport visibly scrolls from mid-replay to the final bottom position.

**Compound effect**: The onRender convergence loop (`XtermTerminal.tsx:550-571`) fires up to 20 re-fits during attach, each sending a backend `Resize` → SIGWINCH → application repaint. These repaint bytes interleave with the initial replay, extending the scroll duration.

## Why No Existing Mitigation Catches This

- **No visual gate during `connecting`**: `TerminalPane` shows `showEmptyState` only when `!handleId` (no terminal handle). Once a session is selected, xterm is immediately visible and receives replay data with no cover overlay (`TerminalPane.tsx:329`).
- **No batch-write API**: `onData` calls `terminal.write(bytes)` per WS frame — no callback signaling when the initial replay is fully processed.
- **`firstAttach` only gates `clear()`** (`useTerminalSession.ts:115`), not visibility.

## Steps to Reproduce

1. Spawn a worker session and let it run for a few minutes (accumulate scrollback)
2. Switch away to another session or close the pane
3. Open the worker session again
4. Observe: the terminal shows mid-session content, then visibly scrolls down to the bottom

## Expected Behavior

The terminal should appear already-rendered at the final scroll position. No visible mid-session scroll.

## Key Files

| File | Symbol | Role |
|------|--------|------|
| `frontend/src/renderer/hooks/useTerminalSession.ts` | `connect()` L195, `onData` L210 | Attach lifecycle, data handler |
| `frontend/src/renderer/components/TerminalPane.tsx` | `AttachedTerminal` L223, `showEmptyState` L329 | Overlay gating |
| `frontend/src/renderer/components/XtermTerminal.tsx` | mount effect L272, onRender stabilizer L550 | xterm init + fit convergence loop |
| `backend/internal/terminal/attachment.go` | `copyOut()` L185, `setPTY()` L274 | Replay pump, `onOpen` timing |
| `backend/internal/adapters/runtime/conpty/host.go` | `handleConn()` L241 | Scrollback snapshot replay |
| `backend/internal/adapters/runtime/conpty/ring.go` | `Snapshot()` L61 | 1000-line scrollback buffer |

## Environment

- All platforms (macOS/Linux via tmux, Windows via conpty)
- Most noticeable on sessions with substantial scrollback (5+ minutes of agent output)

illegalcall:
## Summary

Fixes #3160 — opening a session showed a mid-session position and then visibly scrolled down to the tail.

The daemon pumps a pane's attach replay in 32KB reads (`attachment.go:182` `copyOut`), so the renderer receives N WebSocket frames and issues N `terminal.write()` calls. xterm parses each write atomically, but the browser paints **between** event-loop turns — so every frame boundary is a painted, further-scrolled state.

### Correction to the issue's stated root cause

The issue attributes this to xterm splitting large writes into ~15KB internal batches. It doesn't do that. In `@xterm/xterm@5.5.0`'s `src/common/input/WriteBuffer.ts`, `_writeBuffer` holds **one entry per `write()` call** and `_innerWrite` parses a whole entry atomically; `WRITE_TIMEOUT_MS = 12` is only checked *after* an entry finishes, to decide whether to yield before the next. There is no size-based splitting.

**The batch granularity is the caller's `write()` calls — i.e. our WS frames.** That distinction is what makes the fix below work.

### Measured

Real xterm 5.5.0 in real Chromium, sampling `buffer.active.viewportY` on every animation frame. Distinct consecutive values = scroll positions actually painted.

```
payload   98,528 bytes (1000 lines)   grid 200×50

  delivery              writes  frames  positions   ingest  max block
  32KB chunks, 0ms gap       4       8          1    129ms      0.1ms
  8KB chunks, 4ms gap       13      12          5    200ms      0.1ms
  4KB chunks, 8ms gap       25      22         15    367ms      0.1ms
  4KB chunks, 16ms gap      25      33         25    550ms      0.1ms
  one concatenated write     1       8          1    133ms      2.2ms
```

Severity is `frame count × inter-frame gap` and nothing else. One `write()` of 98KB painted exactly **1** position — if xterm split at ~15KB it would have painted ~7.

## Fix

Buffer the initial burst, write it once, keep the pane covered until that write is parsed.

- **`XtermTerminal.tsx`** — forward xterm's write callback, which was being dropped (`write: (data) => term.write(data)`). It fires when the chunk is parsed, giving a deterministic reveal point instead of a timer guess.
- **`useTerminalSession.ts`** — the gate opens in `connect()`, **not** on `opened`: the daemon fires `onOpen` from `setPTY` and only then starts `copyOut`, so `attached` precedes the first replay byte and gating on it would uncover a pane that hasn't drawn. A 60ms quiet window ends the burst; a 750ms cap bounds the cover and still flushes as one chunk, so the worst case degrades to a single jump rather than back to the walk. Exposes `replaySettled`.
- **`TerminalPane.tsx`** — blank terminal-coloured cover while buffering, label delayed 120ms so fast opens never flash a loader. Deliberately *not* the existing empty-state card, which would flash "Starting session" on every session switch.
- **Resize re-assert held while a burst is in flight** — the backend answers every resize frame with an explicit SIGWINCH, and that repaint would otherwise land ~250ms later, just after the cover lifts. Held, never dropped: losing it leaves the pane laid out for the old grid until the next real change.

Frontend only. No backend or protocol change, and no overlap with #3105 (all backend).

## Test plan

- `npm test` — **1107 tests / 87 files pass**
- `npm run typecheck` — clean
- `go build ./...`, `go test ./internal/terminal/...` — clean (toolchain bumped to Go 1.26.5 to build the daemon)
- 12 new tests: single-write flush, live passthrough after the gate closes, empty replay, cap behaviour, exit/error mid-replay, reattach re-arming, re-assert deferral, superseded-connection frames

## ⚠️ Known gap — please read before sizing this

I captured **real tmux attach replays** off a live daemon and replayed them through real xterm. They paint **1 position even without this change**:

| capture | bytes | frames | span | positions as-is |
|---|---|---|---|---|
| sparse screen | 1,385 | 4 | 0.3ms | 1 |
| dense colored screen | 21,083 | 23 | 1.2ms | 1 |

Structural reason: `tmux attach-session` replays only the **visible screen**, bounded by rows×cols — it never dumps scrollback. So on macOS/Linux the burst is small and lands inside a single animation frame.

The severe case is the **Windows/conpty** path, where `Ring.Snapshot()` (`conpty/ring.go:11`, `MaxOutputLines = 1000`) genuinely returns 1000 lines as one burst.

Still unmeasured:
- an **agent session** TUI attach (I used a shell terminal — same tmux path, but an agent's screen is denser and it repaints on SIGWINCH)
- the **conpty** path end to end
- behaviour under a loaded renderer

So this change is correct and cheap, but whether it's user-visible on macOS is unconfirmed. Happy to hold for an agent-session capture if reviewers would rather size it first.

## Notes

- Frontend comments throughout still say `zellij attach`; the adapters are `tmux`/`conpty`/`ptyexec`. Left alone here to keep the diff scoped.
- One pre-existing `Viewport.syncScrollArea` → `dimensions` error appears once at app startup. `XtermTerminal.tsx`'s own header documents it ("Writing before layout settles is what crashed xterm's Viewport"). It didn't recur across reloads, and this change strictly *delays* the first buffer write — but I couldn't get a clean baseline A/B, so I'm flagging rather than claiming.

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -259,7 +259,7 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 		},
 		[session?.id],
 	);
-	const { attach, state, error } = useTerminalSession(attachSession, {
+	const { attach, state, error, replaySettled } = useTerminalSession(attachSession, {
 		daemonReady,
 		shellTerminalHandleId,
 		onOutput: watchLinks ? handleOutput : undefined,
@@ -363,6 +363,12 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 
 	const banner = bannerText(state, error);
 	const showEmptyState = !handleId;
+	// Cover xterm while the attachment buffers the initial replay, so the pane
+	// appears already drawn at the tail instead of visibly scrolling down to it
+	// (issue #3160). Deliberately NOT the empty state above: that renders a
+	// centered "Starting session" card, and flashing it on every session switch
+	// would be worse than the scroll it replaces.
+	const showReplayCover = Boolean(handleId) && !replaySettled;
 	const showEndedState = state === "exited" || canRestoreSession;
 	const emptyStateTitle = session ? "Starting session" : "Agent Orchestrator";
 	const emptyStateMessage = session
@@ -406,6 +412,7 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 						</div>
 					</div>
 				)}
+				{showReplayCover && <ReplayCover />}
 				{banner && (
 					<div className="absolute inset-x-3 top-2 rounded-md border border-border bg-surface/95 px-3 py-1.5 font-mono text-caption text-muted-foreground">
 						{banner}
@@ -422,6 +429,25 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 					}}
 				/>
 			)}
+		</div>
+	);
+}
+
+// Blank terminal-coloured cover held over xterm while the initial replay is
+// buffered. A fast open (the common case) shows nothing at all — the label only
+// appears if the wait is long enough to read as a stall rather than a repaint,
+// so normal session switching never flashes a loader.
+const REPLAY_COVER_LABEL_MS = 120;
+
+function ReplayCover() {
+	const [showLabel, setShowLabel] = useState(false);
+	useEffect(() => {
+		const timer = window.setTimeout(() => setShowLabel(true), REPLAY_COVER_LABEL_MS);
+		return () => window.clearTimeout(timer);
+	}, []);
+	return (
+		<div className="absolute inset-0 grid place-items-center bg-terminal" data-testid="terminal-replay-cover">
+			{showLabel && <div className="font-mono text-caption text-terminal-dim">Loading latest output…</div>}
 		</div>
 	);
 }
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -368,7 +368,13 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 	// (issue #3160). Deliberately NOT the empty state above: that renders a
 	// centered "Starting session" card, and flashing it on every session switch
 	// would be worse than the scroll it replaces.
-	const showReplayCover = Boolean(handleId) && !replaySettled;
+	// Only while a replay is actually imminent. Gating on the state as well as
+	// the gate keeps the cover from reappearing over a pane that is visibly
+	// disconnected: an open timeout lifts it, the backoff reconnect would
+	// otherwise pull it straight back down, and the "reattaching" banner already
+	// explains that window better than a blank overlay does.
+	const showReplayCover =
+		Boolean(handleId) && !replaySettled && (state === "connecting" || state === "attached");
 	const showEndedState = state === "exited" || canRestoreSession;
 	const emptyStateTitle = session ? "Starting session" : "Agent Orchestrator";
 	const emptyStateMessage = session
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -446,7 +446,13 @@ function ReplayCover() {
 		return () => window.clearTimeout(timer);
 	}, []);
 	return (
-		<div className="absolute inset-0 grid place-items-center bg-terminal" data-testid="terminal-replay-cover">
+		// pointer-events-none: the cover is purely visual and xterm underneath is
+		// live the whole time, so clicks, selection and wheel must pass through
+		// rather than being swallowed for the length of the gate.
+		<div
+			className="pointer-events-none absolute inset-0 grid place-items-center bg-terminal"
+			data-testid="terminal-replay-cover"
+		>
 			{showLabel && <div className="font-mono text-caption text-terminal-dim">Loading latest output…</div>}
 		</div>
 	);
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -364,10 +364,10 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 	const banner = bannerText(state, error);
 	const showEmptyState = !handleId;
 	// Cover xterm while the attachment buffers the initial replay, so the pane
-	// appears already drawn at the tail instead of visibly scrolling down to it
-	// (issue #3160). Deliberately NOT the empty state above: that renders a
-	// centered "Starting session" card, and flashing it on every session switch
-	// would be worse than the scroll it replaces.
+	// appears already drawn at the tail instead of visibly scrolling down to it.
+	// Deliberately NOT the empty state above: that renders a centered "Starting
+	// session" card, and flashing it on every session switch would be worse than
+	// the scroll it replaces.
 	// Only while a replay is actually imminent. Gating on the state as well as
 	// the gate keeps the cover from reappearing over a pane that is visibly
 	// disconnected: an open timeout lifts it, the backoff reconnect would
--- a/frontend/src/renderer/hooks/useTerminalSession.ts
+++ b/frontend/src/renderer/hooks/useTerminalSession.ts
@@ -29,7 +29,12 @@ export type TerminalUserInputSource = "keyboard" | "paste" | "composition" | "sh
 export type AttachableTerminal = {
 	cols: number;
 	rows: number;
-	write: (data: Uint8Array) => void;
+	/**
+	 * `done` fires once this exact chunk has been parsed into the buffer (xterm's
+	 * own write callback). The attachment uses it to reveal the pane at the
+	 * replay's final scroll position instead of guessing with a timer.
+	 */
+	write: (data: Uint8Array, done?: () => void) => void;
 	writeln: (line: string) => void;
 	/**
 	 * Erase screen + scrollback and home the cursor, preserving terminal modes.
@@ -90,6 +95,22 @@ const RESIZE_DEBOUNCE_MS = 100;
 // explicit SIGWINCH (pty_unix.go), so this re-assert makes the client re-read
 // and re-report its grid; when everything is already in sync it's a no-op.
 const RESIZE_REASSERT_MS = 250;
+// Initial-replay gate (issue #3160). On attach the runtime replays the pane's
+// state, and the daemon pumps it in 32KB reads (attachment.go copyOut) — so the
+// renderer gets N WebSocket frames, N `write()` calls, and N separate event-loop
+// turns. xterm parses each write atomically but the browser paints BETWEEN
+// turns, so every frame boundary is a painted, further-scrolled state: the
+// terminal visibly walks from mid-session down to the tail. Measured on a
+// 1000-line replay: 25 frames at 16ms spacing paint 25 distinct scroll
+// positions; the same bytes as ONE write paint exactly 1, for ~2ms of parse.
+//
+// So the replay burst is buffered and written once, with the pane covered until
+// that write is parsed. QUIET_MS is the no-data gap that ends the burst; CAP_MS
+// bounds the cover if the replay never goes quiet. Hitting the cap still writes
+// what has arrived as a single chunk, so the worst case degrades to one jump
+// rather than back to the walk.
+const REPLAY_QUIET_MS = 60;
+const REPLAY_CAP_MS = 750;
 
 function defaultCreateMux(): TerminalMux {
 	// Resolved per connect, not per hook: a daemon restart can change the port.
@@ -100,6 +121,9 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
 	const queryClient = useQueryClient();
 	const [state, setState] = useState<TerminalSessionState>("idle");
 	const [error, setError] = useState<string | undefined>(undefined);
+	// False only while the initial replay is being buffered — the pane keeps a
+	// cover over xterm until the burst has been written and parsed.
+	const [replaySettled, setReplaySettled] = useState(true);
 
 	const sessionRef = useRef(session);
 	sessionRef.current = session;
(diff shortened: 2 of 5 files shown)


frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -230,11 +379,20 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
...
+					r.replayCapTimer = setTimeout(flushReplay, REPLAY_CAP_MS);
+				}
 			}),
 			mux.onExit(handle, () => {
 				if (!isCurrentAttachment(generation, handle, mux)) return;
 				clearOpenTimer(generation);
 				r.inputReady = false;
+				// Land whatever was buffered before the notice, and lift the cover:
+				// a pane that exits mid-replay must never be left behind it.
+				flushReplay();
```
nikhilachale, line 432: [P2] Replay cover can disappear before xterm finishes parsing. flushReplay() queues an asynchronous xterm write, but exit/error/reconnect immediately changes state, causing TerminalPane to hide the cover before the callback fires. This can briefly expose stale output and reproduce the scrolling artifact. The tests miss it because their fake invokes the callback synchronously.

frontend/src/renderer/components/TerminalPane.tsx
```diff
@@ -363,6 +363,18 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 
 	const banner = bannerText(state, error);
 	const showEmptyState = !handleId;
+	// Cover xterm while the attachment buffers the initial replay, so the pane
+	// appears already drawn at the tail instead of visibly scrolling down to it
+	// (issue #3160). Deliberately NOT the empty state above: that renders a
```
nikhilachale: > Please remove the issue number from this code comment. The surrounding explanation is already self-contained, and implementation comments shouldn’t depend on external issue context.

illegalcall:
## Self-review pass on the replay gate

Reviewed this branch against `origin/main` and pushed fixes in c56ce3f. Full renderer suite green (1155 tests, 89 files), `tsc --noEmit` clean.

No defect was found that produces an incorrect outcome — the strand analysis is complete, and the two paths that are easy to miss (socket dying before `opened`; teardown with no reconnect) are both genuinely handled. What follows is one behavioural gap and two comments that documented mechanisms the code does not have.

### The mechanism this PR relies on is real — but #3160's root cause is wrong

Worth recording, because the PR is unusable if the issue is right. #3160 says xterm "splits it into chunks, rendering each chunk before processing the next" in ~15KB batches. If that were true of xterm 5.5.0, concatenating the frames would change nothing.

It isn't. In the installed `_innerWrite`, the drain loop hands **one whole queue entry** to the parser in a single synchronous call and only then tests its 12ms budget:

```js
for (; buffer.length > offset; ) {
  parse(buffer[offset]);                 // whole entry, uninterrupted
  callbacks[offset](); offset++;
  if (Date.now() - start >= 12) break;   // check is BETWEEN entries
}
if (buffer.length > offset) setTimeout(drain);   // yield -> browser may paint
```

The yield sits between entries, never inside one. So N writes can paint between them; one write cannot, at any size. The PR's own comments describe this correctly — only the issue text is wrong. Also means the 1MB byte cap doesn't reintroduce the scroll; it only bounds main-thread block time, which is what its comment claims.

### 1. A pane that replayed nothing stayed covered for the full cap — fixed

The quiet window was armed only from an incoming data frame, so with no data nothing triggered an early flush and the 750ms cap was the only path. The user got a blank cover for 750ms with `Loading latest output…` from 120ms, on a pane with nothing to show. Before this branch they'd have seen the empty terminal immediately.

The `flushReplay` empty-chunks branch claimed otherwise — "reveal immediately rather than holding the cover for the full cap" — but only reveals immediately *once something calls it*, and here the only caller was the cap. The existing test couldn't catch it: it advanced exactly 750ms and asserted settled, which passes either way.

Fix is `REPLAY_FIRST_BYTE_MS = 250`, armed at `opened` alongside the cap. The important part is that it **lifts the cover without ending the gate**. The naive version — arming the quiet timer at `opened` — is actively harmful: `opened` fires from `setPTY` *before* `copyOut`, so a 60ms timer can expire before the first replay byte exists, flush an empty buffer, clear `replayBuffering`, and leave the real replay to arrive frame by frame with the bug fully intact and no cover over it. Same trap the cap-anchoring comment already warns about. There's a regression test that fails if this ever becomes a flush.

### 2. A test credited the cap for a case the cap cannot cover — fixed

`lifts the cover when the attachment is torn down with no reconnect scheduled` said "The cap timer is what guarantees the cover still lifts — it is deliberately shorter than OPEN_TIMEOUT_MS". That test never emits `opened`, and the cap is armed from `opened`, so it never runs. `teardownMux` is what settles. The implementation comment in `connect()` says this correctly, so the test contradicted the code fifteen lines away — on precisely the risk class this design is most exposed to.

Reworded, and now pins the timing: not settled at 2999ms, settled at `OPEN_TIMEOUT_MS`. That fails if either replay timer is ever moved back to `connect()`.

### 3. The resize deferral's rationale was inverted — comment corrected, behaviour kept

The comment justified deferring the re-assert as avoiding a repaint that "would land just after the cover lifts and flash". Deferring *guarantees* that repaint:

| | Fire mid-burst | Defer to flush (ships) |
|---|---|---|
| Repaint bytes go | into the buffer, absorbed into the single write — invisible | onto the revealed pane — visible |

The real benefit is a **shorter cover**: added frames restart the 60ms quiet window, so not injecting more bytes lets the burst go quiet sooner. That's a good reason, just not the one the comment gave. Behaviour unchanged; comment now states the actual trade.

The latent bug the timer split fixes — debounce and re-assert sharing one slot, letting a stale grid land after a newer one — is real and correctly handled.

### Verified and holding

- **Buffering frame references is safe.** `base64ToBytes` allocates a fresh `Uint8Array` per frame, so nothing is reused underneath.
- **No strand path.** Enumerated all of quiet / cap / byte-cap / exit / error / socket-closed / user-input / teardown; every terminal state has a live settle path.
- **No write to a disposed xterm.** `AttachedTerminal` is keyed by handle, so a session switch remounts the subtree and React runs the parent's flush before the child disposes. The one path unmounting xterm alone is the init-failure early return, where `onError` fires from the `catch` around construction — before `onReady`, so no attachment exists.
- **Byte accounting at the 1MB trip** — tripping chunk is pushed before the check, so it's included: no loss, no double write.
- **Decoder ordering** — joined buffer decoded once, before later live chunks.
- **The measurement claims are internally consistent.** "~2ms parse" and "~44KB/ms" only agree if the burst was ~88KB over ~25 frames, i.e. ~3.5KB each. That follows from `copyOut` forwarding whatever a read yields; 32KB is the ceiling, not the frame size.

### Accepted residuals

- Replays over 1MB keep the walk for the tail once the byte cap trips. Right call — a multi-MB single parse would freeze the main thread.
- One blank frame is possible at the reveal, since the cover's removal goes through React while xterm's paint goes through its own scheduler. Blank, never a scroll.
- The reattach path is uncovered by design (state gate). Still gets the single write, so one jump rather than a walk.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):

frontend/src/renderer/components/TerminalPane.tsx
illegalcall: Done in 3adee5b2a — removed here, and applied the same rule to the gate's constant block in `useTerminalSession.ts`.

I left two references in place, say the word if you want them gone too:

- The two test-suite headers (`useTerminalSession.test.tsx`, `TerminalPane.test.tsx`), where the number reads as provenance for why the suite exists rather than context the code depends on.
- `XtermTerminal.tsx:734`, which already carries the reference on `main` — it shipped there via #3184, so editing it here only adds diff noise for a comment.

frontend/src/renderer/hooks/useTerminalSession.ts
illegalcall: The ordering you describe is real, but I don't think either stated consequence follows. Walking it through:

**On the mechanism** — correct. `onExit`/`onError` call `flushReplay()` and then `transition(...)`, and `showReplayCover` is gated on `connecting || attached`, so the state change drops the cover without waiting for xterm's write callback. Same on the reconnect path via `onConnectionChange("closed")` → `scheduleReattach()` → `reattaching`.

**"Briefly expose stale output"** — there is nothing stale to expose. While the gate is open nothing has been written to xterm yet, and that is the whole point of buffering. On a first attach xterm's buffer is empty; on a reattach `connect()` has already called `terminal.clear()` before arming the gate. So what shows for that frame is a blank grid, which is visually what the cover itself was showing (`bg-terminal`).

**"Reproduce the scrolling artifact"** — this one I'd push back on more firmly. The artifact needs *multiple* writes painting successive scroll positions. The flush is a single `write()` of the joined buffer, and xterm parses one write-buffer entry in one uninterrupted `_action` call — the 12ms yield check sits *between* entries, never inside one. So the flush can produce at most one paint regardless of when its callback fires (two on the exit path, since `writeln("[process exited]")` follows). One jump, never a walk.

So the net effect is the residual I noted in the review comment above: one blank frame is possible at the reveal, because the cover's removal goes through React state while xterm's paint goes through its own render scheduler. That's true on the happy path too, not just on exit/error, and it's blank rather than wrong.

**The part I think is a fair hit**: the fake does invoke `done` synchronously, so the suite genuinely does not exercise callback-after-state-change ordering. Happy to make it async — worth knowing before I do that it isn't a free change, because `onExit`/`onError` currently settle *through* that callback, so tests like `lands buffered output and lifts the cover when the pane exits mid-replay` assert `replaySettled` immediately after `emitExit()` and would need timer advances. That's arguably the suite becoming more honest. Want me to take it?

illegalcall:
## Merged `main` — and a heads-up about how the conflict arose

Resolved in 13d977d1b. Worth flagging why this conflicted at all, because it affects `main` and not just this PR.

**#3184 ("fix(landing): keep the mobile app preview's proportions on small screens") shipped an earlier copy of this branch's replay-gate work to `main`.** That commit (e45092a19) touches six files, and five of them are renderer/terminal files with nothing to do with the landing page:

```
frontend/src/landing/.../MobileAppDemo.tsx            <- the actual fix
frontend/src/renderer/components/TerminalPane.test.tsx
frontend/src/renderer/components/TerminalPane.tsx
frontend/src/renderer/components/XtermTerminal.tsx
frontend/src/renderer/hooks/useTerminalSession.test.tsx
frontend/src/renderer/hooks/useTerminalSession.ts
```

Four of those five are **byte-identical to this branch's first commit** (e765f7d48). `XtermTerminal.tsx` differs only because #3103's Option-click handling landed there separately.

So `main` currently runs the unreviewed first draft of this gate, including the defect the later commits on this branch fixed: **the cap is armed from `connect()` rather than from `opened`**. On a slow attach that spends the whole 750ms cap on the handshake, flushes an empty buffer, and leaves `replayBuffering` false for the rest of the attachment — so the replay then lands frame by frame with the scroll bug fully intact, behind a cover that achieved nothing. `main` also lacks the byte ceiling, the teardown flush, and the split re-assert timer.

Landing this PR fixes all of that on `main`. No action needed here, but the packaging is worth someone's attention — a landing-page PR carrying five terminal files past review is the kind of thing that should probably trip a CODEOWNERS or diff-scope check.

### How I resolved it

- `useTerminalSession.ts`, `TerminalPane.tsx` and both test files: took this branch. Since `main`'s copies are byte-identical to e765f7d48 and that commit is an ancestor here, this branch is a strict superset — nothing from `main` is dropped, and the later commits supersede that snapshot deliberately.
- `XtermTerminal.tsx`: kept git's automatic merge. #3103's `event.altKey` → `openLinkInSystemBrowser` and this branch's `write: (data, done)` forwarding are in different regions; both survive.
- Everything else from `main` came through untouched, including #3184's actual `MobileAppDemo.tsx` change.

Verified on the merge result: frontend typecheck clean, renderer suite **1175 passed across 91 files**. Net diff against `main` is now four files.

PR #3191, opened 2026-07-27
fix: use wait-based conpty pid liveness

### LINKED ISSUE #3080: Windows ConPTY pidAlive returns true for terminated processes (OpenProcess without WaitForSingleObject)
## Summary

`pidAlive` in `backend/internal/adapters/runtime/conpty/pidalive_windows.go` probes process liveness by opening the process handle and returning `true` if `OpenProcess` succeeds — **without ever calling `WaitForSingleObject`**. On Windows a terminated process's kernel object persists until all handles to it are closed, so `OpenProcess` succeeds for recently-dead processes and `pidAlive` reports them as alive.

## Root cause

The current implementation (`pidalive_windows.go:14-21`):

```go
func pidAlive(pid int) bool {
    h, err := windows.OpenProcess(windows.SYNCHRONIZE, false, uint32(pid))
    if err != nil {
        return false
    }
    _ = windows.CloseHandle(h)
    return true // only checks that the process OBJECT exists, not that it's running
}
```

It requests `SYNCHRONIZE` access — the access right needed to call `WaitForSingleObject` on the handle — but never waits. The permission is effectively dead code; the probe degenerates to "does the kernel object still exist?", which is `true` for any terminated-but-not-fully-reaped process.

The **correct pattern already exists** in this repo at `backend/internal/processalive/process_windows.go::Alive`, which is the shared package for exactly this check:

```go
func Alive(pid int) bool {
    handle, err := windows.OpenProcess(windows.SYNCHRONIZE, false, uint32(pid))
    if err != nil {
        return errors.Is(err, windows.ERROR_ACCESS_DENIED) // access denied => exists
    }
    defer windows.CloseHandle(handle)
    status, err := windows.WaitForSingleObject(handle, 0)
    if err != nil {
        return false
    }
    return status == uint32(windows.WAIT_TIMEOUT) // WAIT_TIMEOUT = still running
}
```

`WAIT_OBJECT_0` (signaled) = terminated; `WAIT_TIMEOUT` = still running. The conpty `pidAlive` is also weaker than its own Unix sibling (`pidalive_unix.go`), which explicitly treats zombies as dead.

Two additional gaps vs. `processalive.Alive`:
- No `pid <= 0` guard.
- `ERROR_ACCESS_DENIED` is treated as dead instead of alive (the process exists, just not openable with the requested rights).

## Impact

`pidAlive` is consumed by the conpty runtime shutdown path (`runtime.go:118-127`):

```go
// Poll up to ~500ms (20 x 25ms) for the pty-host pid to exit.
deadline := time.Now().Add(500 * time.Millisecond)
for time.Now().Before(deadline) {
    if !pidAlive(sess.pid) {
        break
    }
    time.Sleep(25 * time.Millisecond)
}
```

The intent is "ask the host to shut down gracefully, detect its exit early, otherwise force-kill after ~500ms." With the bug, the early-exit `break` is effectively dead on Windows — the loop spins the full 20 iterations on **every** session teardown even when the host exits in a few milliseconds. Sessions still terminate correctly (the unconditional force-kill after the loop handles it), so this is a bounded latency/correctness degradation, not a hang — hence **P2**.

## Suggested fix

Delegate to the existing, tested `processalive.Alive`:

```go
//go:build windows

package conpty

import (
    "fmt"
    "os"

    "github.com/aoagents/ao/backend/internal/processalive"
)

func pidAlive(pid int) bool {
    return processalive.Alive(pid)
}
```

…or, if the conpty package wants to keep its own probe for test-injection reasons (`pidAlive` is a package-level var in `ptyregistry`), inline the `WaitForSingleObject(handle, 0)` + `WAIT_TIMEOUT` check and add the `pid <= 0` / `ERROR_ACCESS_DENIED` handling from `processalive.Alive`.

Either way, **do not** rely on `OpenProcess` success alone.

## Environment

- Windows ConPTY runtime adapter (`//go:build windows`)
- Reproducible on any Windows host using the ConPTY runtime; not observable on macOS/Linux (separate `pidalive_unix.go`).

---

Reported by Neo in Discord #bug-triage.

AgentWrapper:
## Summary
- make the Windows ptyregistry PID probe delegate to processalive.Alive so exited process handles are checked with WaitForSingleObject
- add Windows-only regression tests for processalive.Alive, conpty.pidAlive, and ptyregistry.defaultPidAlive against exited-but-unreaped processes

## Root cause
The ptyregistry Windows liveness probe used OpenProcess success as proof that a process was still running. On Windows, a terminated process object can remain openable while handles still exist, so the probe could preserve dead pty-host entries.

Fixes #3080.

## Validation
- go test ./internal/processalive ./internal/adapters/runtime/conpty ./internal/adapters/runtime/conpty/ptyregistry
- GOOS=windows GOARCH=amd64 go test -c -o /tmp/processalive.test.exe ./internal/processalive
- GOOS=windows GOARCH=amd64 go test -c -o /tmp/conpty.test.exe ./internal/adapters/runtime/conpty
- GOOS=windows GOARCH=amd64 go test -c -o /tmp/ptyregistry.test.exe ./internal/adapters/runtime/conpty/ptyregistry
- git diff --check

## Notes
- Full backend go test ./... was attempted; affected packages passed, but the run failed in unrelated existing tests: internal/adapters/agent/kilocode TestAuthStatusUnknownWhenKeyOnlyComesFromInteractiveShell, internal/adapters/runtime/tmux TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell, and several internal/cli hook/spawn tests.

### DIFF
--- a/backend/internal/adapters/runtime/conpty/pidalive_windows_test.go
+++ b/backend/internal/adapters/runtime/conpty/pidalive_windows_test.go
@@ -1,0 +1,41 @@
+//go:build windows
+
+package conpty
+
+import (
+	"os/exec"
+	"testing"
+	"time"
+
+	"golang.org/x/sys/windows"
+)
+
+func TestPidAliveReportsExitedUnreapedProcessAsDead(t *testing.T) {
+	cmd := exec.Command("cmd", "/c", "exit 0")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("start child: %v", err)
+	}
+	defer func() { _ = cmd.Wait() }()
+
+	handle, err := windows.OpenProcess(windows.SYNCHRONIZE, false, uint32(cmd.Process.Pid))
+	if err != nil {
+		t.Fatalf("open child process: %v", err)
+	}
+	defer windows.CloseHandle(handle)
+
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		status, err := windows.WaitForSingleObject(handle, 0)
+		if err != nil {
+			t.Fatalf("wait for child process: %v", err)
+		}
+		if status == uint32(windows.WAIT_OBJECT_0) {
+			if pidAlive(cmd.Process.Pid) {
+				t.Fatal("pidAlive returned true for exited unreaped process")
+			}
+			return
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	t.Fatal("child process did not exit before timeout")
+}
--- a/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows.go
+++ b/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows.go
@@ -3,17 +3,10 @@
 package ptyregistry
 
 import (
-	"golang.org/x/sys/windows"
+	"github.com/aoagents/agent-orchestrator/backend/internal/processalive"
 )
 
-// defaultPidAlive probes PID liveness via OpenProcess. SUCCESS means alive
-// (CloseHandle and return true). ERROR_ACCESS_DENIED mirrors EPERM: the
-// process exists but cannot be queried, so treat as alive.
+// defaultPidAlive probes PID liveness on Windows.
 func defaultPidAlive(pid int) bool {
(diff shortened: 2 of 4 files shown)


AgentWrapper:

PR #3192, opened 2026-07-27
fix: ignore daemon SIGPIPE from closed log pipes

### LINKED ISSUE #3182: Daemon dies of SIGPIPE when the supervisor's stdout pipe closes, leaving a stale running.json
## Bug

When the Electron supervisor goes away (Ctrl-C on `ao start --source` / `npm run dev`, or any hard Electron exit), the daemon's stdout/stderr pipe loses its reader. The next log line the daemon writes gets `EPIPE`, and because Go terminates on `SIGPIPE` for fds 1 and 2 unless a handler is registered, the daemon dies **before** it can run its graceful shutdown. `running.json` is left behind pointing at a dead PID.

The daemon does exit, so this reads as "shutdown worked" — but it exited by signal, not by shutting down.

**Source:** found while verifying the daemon process-tree change for the fork's [xuelongmu/agent-orchestrator#231](https://github.com/xuelongmu/agent-orchestrator/pull/231) | **Analyzed against:** `823c29d2` (`main`)
**Confidence:** High — reproduced in isolation with the exact exit signal.
**Priority:** medium — the daemon still exits and the next start overwrites the run file; the cost is a stale handshake that makes `ao status` and CLI commands aim at a dead daemon until then.

## Reproduction

Isolated repro that removes Electron from the picture entirely — a daemon whose stdout is a pipe, a reader that goes away, then one log-producing request:

```bash
cd backend && go build -o /tmp/ao ./cmd/ao
mkfifo /tmp/aoprobe.pipe
AO_DATA_DIR=/tmp/aoprobe/data AO_RUN_FILE=/tmp/aoprobe/running.json AO_PORT=3099 \
  /tmp/ao daemon > /tmp/aoprobe.pipe 2>&1 &
DAEMON=$!
exec 3< /tmp/aoprobe.pipe
head -c 120 <&3 > /dev/null   # consume startup lines
exec 3<&-                     # reader gone
curl -s -o /dev/null http://127.0.0.1:3099/healthz   # forces an "http request" log line
wait $DAEMON; echo "exit=$?"
```

Observed:

```
daemon exit status: 141   (128+13 = SIGPIPE)
run file after exit: STILL PRESENT
```

Same outcome through the real path on macOS: `ISOLATE_DEV=true ao start --source`, then Ctrl-C. The daemon's log ends mid-stream with no shutdown lines, and `~/.ao/dev/running.json` survives.

## Root Cause

- `backend/internal/daemon/daemon.go:291` — `newLogger()` writes to `os.Stderr`, which is a pipe to Electron for every app-spawned daemon.
- `backend/internal/daemon/daemon.go:87` — `signal.NotifyContext(..., syscall.SIGINT, syscall.SIGTERM)` covers the graceful signals but not `SIGPIPE`.

Per `os/signal`: SIGPIPE is ignored for other fds, but a write to fd 1 or 2 that returns `EPIPE` raises SIGPIPE and takes the default action — terminate — unless the program registered for it via `signal.Notify`. So the first log line after Electron exits kills the daemon, skipping the graceful path that removes `running.json`.

Note the daemon *has* a correct teardown path for exactly this situation: the supervisor socket drops and it self-stops within `DefaultShutdownTimeout`. SIGPIPE just beats it to the punch.

## Fix

Register a discarding SIGPIPE handler on POSIX before serving, so a broken stdout/stderr surfaces as a write error instead of killing the process:

```go
// SIGPIPE on fd 1/2 terminates a Go program by default. The daemon's stderr is a
// pipe to whichever supervisor spawned it; when that supervisor exits, the next
// log line must not kill us before the graceful shutdown runs.
signal.Notify(make(chan os.Signal, 1), syscall.SIGPIPE)
```

Worth pairing with a test that asserts the daemon removes its run file when its stdout reader disappears.

## Impact

- Stale `running.json` after Ctrl-C, so `ao status` and CLI commands target a dead PID/port until something rewrites it.
- Shutdown work the daemon would otherwise do on the graceful path is skipped.
- Masks the real shutdown story in testing: a daemon that "exits promptly" looks correct while actually dying by signal. This is what made the Windows daemon-lifecycle verification ambiguous.

## Related
- [#2242](https://github.com/AgentWrapper/agent-orchestrator/issues/2242): Electron keeps stale project actions enabled after daemon exit (same stale-state-after-exit surface).

AgentWrapper:
## Summary
- register a POSIX SIGPIPE handler when the daemon starts so closed supervisor stdout/stderr pipes cannot kill the process
- keep Windows behavior as a no-op and rely on the existing owned run-file cleanup path during graceful shutdown
- add subprocess and e2e regressions covering closed stderr/stdout pipes and running.json cleanup after supervisor disconnect

Fixes #3182

## Tests
- go test ./internal/daemon -run TestIgnoreBrokenPipeSignalAllowsClosedStdoutStderr -count=1
- go test -tags e2e ./internal/cli -run TestE2E_DaemonSurvivesClosedLogPipeAndCleansRunFile -count=1 -v
- env -u AO_SESSION_ID -u AO_RUNTIME_LAUNCH_ID -u AO_LAUNCH_ID -u AO_PROJECT_ID -u AO_DATA_DIR -u AO_RUN_FILE -u AO_PORT go test ./internal/daemon ./internal/httpd
- env -u AO_SESSION_ID -u AO_RUNTIME_LAUNCH_ID -u AO_LAUNCH_ID -u AO_PROJECT_ID -u AO_DATA_DIR -u AO_RUN_FILE -u AO_PORT go test ./internal/cli
- env -u AO_SESSION_ID -u AO_RUNTIME_LAUNCH_ID -u AO_LAUNCH_ID -u AO_PROJECT_ID -u AO_DATA_DIR -u AO_RUN_FILE -u AO_PORT npm run lint

## Notes
- An initial full backend test run hit a local tmux integration timeout in TestRuntimeIntegration; rerunning that package passed, and npm run lint later passed the full backend test suite and golangci-lint.

### DIFF
--- a/backend/internal/cli/e2e_test.go
+++ b/backend/internal/cli/e2e_test.go
@@ -18,6 +18,7 @@ package cli_test
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"os"
@@ -63,6 +64,20 @@ type env struct {
 func newEnv(t *testing.T) env {
 	t.Helper()
 	dir := t.TempDir()
+	return env{
+		runFile: filepath.Join(dir, "running.json"),
+		dataDir: filepath.Join(dir, "data"),
+		port:    freePort(t),
+	}
+}
+
+func newShortSocketEnv(t *testing.T) env {
+	t.Helper()
+	dir, err := os.MkdirTemp("/tmp", "ao-e2e-")
+	if err != nil {
+		return newEnv(t)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
 	return env{
 		runFile: filepath.Join(dir, "running.json"),
 		dataDir: filepath.Join(dir, "data"),
@@ -162,6 +177,11 @@ func (e env) startDaemon(t *testing.T) {
 		<-waitDone
 	})
 
+	e.waitReady(t)
+}
+
+func (e env) waitReady(t *testing.T) {
+	t.Helper()
 	deadline := time.Now().Add(10 * time.Second)
 	for {
 		out, _ := e.run(t, "status", "--json")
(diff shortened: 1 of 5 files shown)


AgentWrapper:

PR #3193, opened 2026-07-28
fix: ignore daemon sigpipe on closed output

### LINKED ISSUE #3182: Daemon dies of SIGPIPE when the supervisor's stdout pipe closes, leaving a stale running.json

AgentWrapper:
## Summary
- Register a POSIX SIGPIPE handler during daemon startup so closed stdout/stderr pipes return write errors instead of terminating the daemon.
- Keep Windows as an explicit no-op for the platform helper.
- Add regressions for closed stderr survival and the real daemon closed-output plus supervisor-disconnect path that removes `running.json`.

## Root Cause
When Electron or a dev supervisor exits, the daemon can inherit stdout/stderr pipes with no reader. A later log write to fd 1/2 can raise SIGPIPE and terminate the Go process before `httpd.Server.Run` reaches its deferred `running.json` cleanup.

Fixes #3182.

## Tests
- `go test ./internal/daemon -run TestIgnoreBrokenPipeSignalSurvivesClosedStderr`
- `go test -tags e2e ./internal/cli -run TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile -count=1 -v`
- `env -u AO_RUNTIME_LAUNCH_ID -u AO_LAUNCH_ID -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_OWNER -u AO_ISSUE_ID -u AO_DATA_DIR -u AO_RUN_FILE -u AO_PORT go test ./internal/daemon ./internal/cli ./internal/httpd`
- `npm run lint` with AO worker env stripped and isolated `GOLANGCI_LINT_CACHE`

## Notes
The default local golangci-lint cache in this worktree contained stale diagnostics from a deleted sibling worktree, so the final lint run used an isolated cache.

### DIFF
--- a/backend/internal/cli/e2e_sigpipe_unix_test.go
+++ b/backend/internal/cli/e2e_sigpipe_unix_test.go
@@ -1,0 +1,117 @@
+//go:build e2e && !windows
+
+package cli_test
+
+import (
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile(t *testing.T) {
+	dir, err := os.MkdirTemp("/tmp", "ao-sigpipe-")
+	if err != nil {
+		t.Fatalf("mktemp: %v", err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	e := env{
+		runFile: filepath.Join(dir, "running.json"),
+		dataDir: filepath.Join(dir, "data"),
+		port:    freePort(t),
+	}
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("pipe: %v", err)
+	}
+	t.Cleanup(func() { _ = r.Close() })
+	go func() {
+		_, _ = io.Copy(io.Discard, r)
+	}()
+
+	cmd := exec.Command(aoBin, "daemon")
+	cmd.Env = e.environ("")
+	cmd.Stdout = w
+	cmd.Stderr = w
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("spawn ao daemon: %v", err)
+	}
+	_ = w.Close()
+
+	processDone := make(chan struct{})
+	var waitErr error
+	go func() {
+		waitErr = cmd.Wait()
+		close(processDone)
+	}()
+	t.Cleanup(func() {
+		select {
+		case <-processDone:
+			return
+		default:
+		}
+		_ = cmd.Process.Kill()
+		select {
+		case <-processDone:
+		case <-time.After(5 * time.Second):
+			t.Error("daemon process did not exit after kill")
+		}
+	})
+
+	deadline := time.Now().Add(10 * time.Second)
+	for {
+		info, err := os.Stat(e.runFile)
+		if err == nil && info.Size() > 0 {
+			break
+		}
+		if err != nil && !os.IsNotExist(err) {
+			t.Fatalf("stat run-file: %v", err)
+		}
+		select {
+		case <-processDone:
+			t.Fatalf("daemon exited before writing run-file: %v", waitErr)
+		default:
+		}
+		if time.Now().After(deadline) {
+			t.Fatal("daemon did not write run-file within 10s")
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	supervisorConn, err := net.DialTimeout("unix", filepath.Join(filepath.Dir(e.runFile), "supervise.sock"), 2*time.Second)
+	if err != nil {
+		t.Fatalf("connect supervisor socket: %v", err)
+	}
+
+	_ = r.Close()
+
+	resp, err := (&http.Client{Timeout: 2 * time.Second}).Get("http://127.0.0.1:" + strconv.Itoa(e.port) + "/healthz")
+	if err != nil {
+		t.Fatalf("GET /healthz after output reader closed: %v", err)
+	}
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET /healthz = %d, want 200", resp.StatusCode)
+	}
+
+	_ = supervisorConn.Close()
+
+	select {
+	case <-processDone:
+		if waitErr != nil {
+			t.Fatalf("daemon exited via signal/error after output pipe closed: %v", waitErr)
+		}
+	case <-time.After(8 * time.Second):
+		t.Fatal("daemon did not self-stop after supervisor disconnect")
+	}
+
+	if _, err := os.Stat(e.runFile); !os.IsNotExist(err) {
+		t.Fatalf("run-file after closed-pipe shutdown = %v, want missing", err)
+	}
+}
--- a/backend/internal/cli/e2e_sigpipe_unix_test.go
+++ b/backend/internal/cli/e2e_sigpipe_unix_test.go
@@ -4,7 +4,6 @@ package cli_test
 
 import (
 	"io"
-	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


AgentWrapper (commented):
Ready to merge. I reviewed the SIGPIPE handling change and the new daemon/e2e coverage; I did not find any blocking issues.

Tests run:
- `go test ./internal/daemon`
- `go test -tags e2e ./internal/cli -run TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile -count=1`

backend/internal/daemon/sigpipe_unix.go
```diff
@@ -0,0 +1,17 @@
...
+	"os/signal"
+	"syscall"
+)
+
+var sigpipeSink = make(chan os.Signal, 1)
+
+// ignoreBrokenPipeSignal prevents a closed supervisor stdout/stderr pipe from
+// killing the daemon before its graceful shutdown can remove running.json.
+func ignoreBrokenPipeSignal() {
+	signal.Notify(sigpipeSink, syscall.SIGPIPE)
```
AgentWrapper, line 16: Good call keeping this scoped to the daemon startup path with a package-level sink. This prevents the supervisor stdout/stderr close from terminating the process before graceful cleanup can remove `running.json`.

AgentWrapper (commented):
Adding one inline review note as requested; still no changes requested.

backend/internal/daemon/sigpipe_unix_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("pipe: %v", err)
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestIgnoreBrokenPipeSignalSurvivesClosedStderr")
+	cmd.Env = append(os.Environ(), "AO_SIGPIPE_HELPER=1")
+	cmd.Stderr = w
+	if err := cmd.Start(); err != nil {
```
AgentWrapper, line 26: This helper test can pass even if SIGPIPE is not ignored, because the child may write to stderr before the parent closes the read end on line 29. Please make the broken-pipe condition deterministic, e.g. close `r` before starting the helper or add a handshake so the helper only writes after the reader is closed.

backend/internal/cli/e2e_sigpipe_unix_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile(t *testing.T) {
+	dir, err := os.MkdirTemp("/tmp", "ao-sigpipe-")
```
AgentWrapper: Nit: please prefer `t.TempDir()` here like the rest of the e2e helpers instead of hard-coding `/tmp` plus manual cleanup. It keeps the test aligned with package conventions and avoids depending on a writable `/tmp` path.

AgentWrapper (commented):
Needs changes: I found a small test determinism issue worth fixing before merge.

backend/internal/daemon/sigpipe_unix.go
AgentWrapper: Thanks. Kept the handler scoped to daemon startup with the package-level sink.

backend/internal/daemon/sigpipe_unix_test.go
AgentWrapper: Fixed in cdd5109a1: the parent now closes the read end before starting the helper, so the helper stderr write deterministically hits a closed pipe and fails without the SIGPIPE ignore.

backend/internal/cli/e2e_sigpipe_unix_test.go
AgentWrapper: Fixed in cdd5109a1: the e2e now uses t.TempDir(). To avoid Unix socket path-length dependence, it triggers /shutdown after the closed-output log write and still verifies running.json cleanup.

AgentWrapper (commented):
Ready to merge. I reviewed the updated SIGPIPE/run-file cleanup change at cdd5109 and did not find any blocking issues. The follow-up now makes the daemon helper test deterministic by closing the read side before launching the child, and the e2e regression covers closed stdout/stderr followed by a graceful `/shutdown` path.

Tests run against the PR head in a temporary checkout:
- `go test ./internal/daemon`
- `go test -tags e2e ./internal/cli -run TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile -count=1`

AgentWrapper (commented):
Ready to merge for the mirrored/canonical PR as well. I reviewed cdd5109 against the current base and found no changes to request.

Tests run against the PR head in a temporary checkout:
- `go test ./internal/daemon`
- `go test -tags e2e ./internal/cli -run TestE2E_DaemonClosedOutputPipeShutdownRemovesRunFile -count=1`

PR #3195, opened 2026-07-28
fix(topbar): stop worker branch label reading as a button; tooltip for cropped names

### LINKED ISSUE #3173: Session name in topbar is cropped and looks like a button
## Bug

**Reported by:** @aditi (Aditi, core-BE)

The session name / branch identifier displayed at the top of the agent session view is **cropped** on narrower windows, and the **visual design makes it look like a clickable button** when it is not.

## Repro

1. Open a worker session in the desktop app (macOS)
2. Observe the topbar area at the top of the session view — it shows the branch name (e.g. `ao/agent-orchestrator-53/root`) next to the status pill (e.g. "Idle")
3. **Cropping:** When the window is narrow (or the sidebar is expanded), the branch name is truncated with ellipsis because the span uses `truncate` in a `min-w-0` flex container that competes for space with the right-side actions (Kill, Orchestrator, bell, inspector toggle), which are all `shrink-0`.
4. **Button-like appearance:** The branch text sits immediately next to the `StatusPill`, which has rounded corners (`rounded-md`), background fill, and an inset border — making it look like a pill/badge. The proximity of the GitBranch icon + monospace branch text to this pill creates the visual impression that the branch text is a clickable chip or button, when it is actually static non-interactive text.

## Location

**File:** `frontend/src/renderer/components/ShellTopbar.tsx` — lines 143–152 (worker session branch display)

```tsx
<div className="flex min-w-0 items-center gap-3">
    {session?.branch ? (
        <div className="inline-flex min-w-0 items-center gap-1 font-mono text-2xs leading-none text-passive">
            <GitBranch className="size-icon-2xs shrink-0" aria-hidden="true" />
            <span className="truncate">{session.branch}</span>
        </div>
    ) : null}
    {session ? <SessionStatusPill session={session} /> : null}
</div>
```

On macOS, this topbar renders inside `SessionView` (`hidesShellTopbar()` returns `true`), so it appears as a bar at the top of the session detail panel.

The `StatusPill` component (`frontend/src/renderer/components/StatusPill.tsx`) applies a background fill, inset box-shadow border, and `rounded-md` padding — visually identical to a badge/chip.

## Suggested fixes

1. **Cropping:** Show the full branch name in a tooltip on hover when truncated, or display only the last path segment (e.g. `root` instead of `ao/agent-orchestrator-53/root`) with the full path available via hover/title. Alternatively, give the branch container more breathing room by reducing the visual weight of the right-side action cluster.

2. **Button-like appearance:** Add clear visual separation between the static branch text and the status pill (more gap, a divider, or different visual hierarchy) so the branch text does not read as part of an interactive chip. The status pill styling could also be toned down in this context to avoid the badge/chip appearance.

AgentWrapper:
## Summary

Fixes #3173 — the worker session topbar's branch identifier was cropped on narrow windows with no way to recover the full name, and its proximity to the status pill made the static text read as a clickable chip.

- The branch stays plain mono text (no fill/border/hover affordance) and a hairline divider now separates it from the status pill, so the two no longer merge into one button-like chip.
- The name keeps the full row width where space allows; when it is actually cropped (measured with a ResizeObserver on the truncating span), the full branch is recoverable from a shadcn/Radix tooltip. The Tooltip root stays mounted and only its content is gated on the cropped state, so the measured span never remounts out from under its observer.
- `WindowTitlebar.tsx` needs no change — it is the Windows-only menu bar and renders no session name; the session identity lives solely in `ShellTopbar`.

## Tests

- New `ShellTopbar worker branch (#3173)` suite: branch renders as plain text (not a button), divider present/absent with/without a branch, no tooltip while the name fits, tooltip with the full branch when cropped (layout metrics stubbed since jsdom has no layout).
- Full frontend suite: 89 files / 1149 tests pass; `npm run typecheck` clean.
- Verified visually in a browser against the live daemon (vite `dev:web` proxy): full name + divider at wide widths, ellipsis at tight widths, tooltip showing `ao/agent-orchestrator-63/root` on hover.

## Risks / notes

- Existing tests that render `ShellTopbar` directly now wrap it in `TooltipProvider` (the app already provides one at the root route).
- At very tight widths the collapsed-sidebar TitlebarNav overlay can sit above the topbar text in the browser preview harness; that is pre-existing and unrelated to this change.

### DIFF
--- a/frontend/src/renderer/components/ShellTopbar.tsx
+++ b/frontend/src/renderer/components/ShellTopbar.tsx
@@ -373,8 +373,9 @@ function ProjectTerminationFeedback({ projectId }: { projectId: string | undefin
 // window runs out of space); while it is actually cropped, the full branch is
 // recoverable from a tooltip. The Tooltip root stays mounted and only the
 // content is gated on the cropped state, so the measured span never remounts
-// out from under its ResizeObserver.
-export function SessionBranchLabel({ branch }: { branch: string }) {
+// out from under its ResizeObserver. The trigger opts out of the macOS drag
+// region (noDragStyle) — inside it, Electron would swallow the hover.
+function SessionBranchLabel({ branch }: { branch: string }) {
 	const textRef = useRef<HTMLSpanElement | null>(null);
 	const [isCropped, setIsCropped] = useState(false);
 
@@ -391,7 +392,10 @@ export function SessionBranchLabel({ branch }: { branch: string }) {
 	return (
 		<Tooltip>
 			<TooltipTrigger asChild>
-				<div className="inline-flex min-w-0 cursor-default items-center gap-1 font-mono text-2xs leading-none text-passive">
+				<div
+					className="inline-flex min-w-0 cursor-default items-center gap-1 font-mono text-2xs leading-none text-passive"
+					style={noDragStyle}
+				>
 					<GitBranch className="size-icon-2xs shrink-0" aria-hidden="true" />
 					<span ref={textRef} className="truncate">
 						{branch}
--- a/frontend/src/renderer/components/ShellTopbar.tsx
+++ b/frontend/src/renderer/components/ShellTopbar.tsx
@@ -1,7 +1,7 @@
 import { useQueryClient } from "@tanstack/react-query";
 import { useNavigate, useParams } from "@tanstack/react-router";
 import { GitBranch, LayoutDashboard, PanelRightClose, PanelRightOpen, Plus, Trash2 } from "lucide-react";
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { NotificationCenter } from "./NotificationCenter";
 import {
 	findProjectOrchestrator,
@@ -27,6 +27,7 @@ import { isMacPlatform, usesBoardActionsInPanel } from "../lib/platform";
 import { StatusPill } from "./StatusPill";
 import { TopbarButton, TopbarKillError, topbarHeaderClass, topbarProjectLabelClass } from "./TopbarButton";
 import { SessionTerminationPopover } from "./SessionTerminationPopover";
+import { Tooltip, TooltipContent, TooltipTrigger } from "./ui/tooltip";
 
 const isMac = isMacPlatform();
 const boardActionsInPanel = usesBoardActionsInPanel();
@@ -156,10 +157,12 @@ export function ShellTopbar() {
 				) : isSessionRoute ? (
 					<div className="flex min-w-0 items-center gap-3">
 						{session?.branch ? (
-							<div className="inline-flex min-w-0 items-center gap-1 font-mono text-2xs leading-none text-passive">
-								<GitBranch className="size-icon-2xs shrink-0" aria-hidden="true" />
-								<span className="truncate">{session.branch}</span>
-							</div>
+							<>
+								<SessionBranchLabel branch={session.branch} />
+								{/* Hairline divider so the static branch text can't be read as
+								    part of the status pill's chip (#3173). */}
+								<span aria-hidden="true" className="h-3.5 w-px shrink-0 bg-border" />
+							</>
 						) : null}
 						{session ? <SessionStatusPill session={session} /> : null}
 					</div>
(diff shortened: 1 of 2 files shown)


frontend/src/renderer/components/ShellTopbar.tsx
```diff
@@ -364,6 +367,42 @@ function ProjectTerminationFeedback({ projectId }: { projectId: string | undefin
...
+		measure();
+		const observer = new ResizeObserver(measure);
+		observer.observe(el);
+		return () => observer.disconnect();
+	}, [branch]);
+
+	return (
+		<Tooltip>
+			<TooltipTrigger asChild>
+				<div className="inline-flex min-w-0 cursor-default items-center gap-1 font-mono text-2xs leading-none text-passive">
```
AgentWrapper: On macOS this trigger sits inside a drag region: the header applies `WebkitAppRegion: drag` (`dragStyle`), and `SessionView` mounts `ShellTopbar` in-panel on macOS. Electron does not deliver pointer events to elements inside drag regions — which is why every interactive element in this header passes `style={noDragStyle}`. Radix opens the tooltip from pointer events on the trigger, so on macOS the cropped-name tooltip will most likely never open. The `dev:web` browser check cannot catch this because `-webkit-app-region` is inert outside Electron. Suggest `style={noDragStyle}` on this div (matching the other topbar controls) and a quick hover verify in the Electron app on macOS.

AgentWrapper (commented):
## Review

**Requesting changes** for one functional gap in the feature this PR adds; everything else looks solid.

### Main finding

**The cropped-name tooltip is very likely unreachable on macOS.** On macOS the topbar header gets `-webkit-app-region: drag` (`dragStyle` at the top of `ShellTopbar.tsx`), and `SessionView` mounts `ShellTopbar` in-panel there (`hidesShellTopbar()`), so the new branch label renders inside an Electron drag region. Drag regions consume pointer events before they reach the DOM — the existing convention in this header is that every element expecting interaction opts out via `style={noDragStyle}` (all the topbar buttons and the notification bell). The tooltip trigger does not, and Radix opens tooltips from pointer events on the trigger, so on macOS the hover should never fire. Windows/Linux are unaffected (`dragStyle` is undefined there). The visual verification in the PR description ran in a plain browser via `dev:web`, where `-webkit-app-region` is ignored, so it could not have caught this. The fix is one line — `style={noDragStyle}` on the trigger div — plus a hover check in the actual Electron app on macOS.

### Minor, non-blocking

- The trigger div is not focusable, so the recovered name is pointer-only; Radix tooltips normally also open on keyboard focus. This is in tension with the goal of not reading as a control, so leaving it is defensible — flagging it as a conscious trade-off.
- `ResizeObserver` only fires when the span box resizes; a late-loading mono font can change `scrollWidth` without a box resize, leaving `isCropped` stale until the next resize. Edge case, fine to leave.
- `SessionBranchLabel` is exported but nothing imports it (the tests render `ShellTopbar`); the `export` can be dropped.

### Verified locally

- `ShellTopbar.test.tsx`: 32/32 pass, including the five new `#3173` tests; `npm run typecheck` clean.
- Both queued heads (`26e36a98` pre-rebase and `4aa9c81f` post-rebase) carry a byte-identical patch; this review covers both.

The test approach (prototype-level `scrollWidth`/`clientWidth` stubs with restore) is sound, and keeping the Tooltip root mounted so the measured span survives crop-state flips is the right call.

frontend/src/renderer/components/ShellTopbar.tsx
AgentWrapper: Fixed in b19af8473 — the tooltip trigger div now passes `style={noDragStyle}`, matching every other interactive control in this header, and the unused `export` on `SessionBranchLabel` is dropped.

On verification: I attempted a real-input hover check on macOS two ways. (1) A minimal Electron 33 harness (drag-region header, plain child vs `no-drag` child) driven by Playwright showed both children receive pointer events — CDP-injected input bypasses the OS-level drag-region hit test, so it can't reproduce the swallowing. (2) Posting genuine CGEvent mouse moves failed: this session lacks Accessibility permission, so the events are silently dropped (cursor verifiably never moved). A physical-mouse check in the packaged app on macOS therefore still needs a human hand — but the fix is exactly the exemption the demonstrably-working controls beside it (Kill / Orchestrator / bell) rely on. A jsdom unit test can't lock this in either: I probed it, and jsdom's CSSStyleDeclaration drops `-webkit-app-region` entirely (empty style attribute), so there's nothing to assert on.

Also noting the two non-blocking points as conscious trade-offs kept as-is: the trigger stays non-focusable (keyboard focus would give it control-like affordance, the opposite of this fix's goal), and the ResizeObserver/late-font edge case is accepted.

PR #3200, opened 2026-07-28
feat(web): show orchestrator activity on the board

### LINKED ISSUE #3141: feat(web): show a running orchestrator's activity indicator on the board (parity with workers)
## Feature request (UI/UX)

**Wish:** give a running orchestrator session a positive activity indicator on the sessions board / sidebar, comparable to the glowing status dot a worker session shows while it is working.

## Today

- A running **worker** session shows a glowing status dot on the board (the `working`/active tone) while the agent is busy.
- A running **orchestrator** shows **no per-session activity status** on the board. It only surfaces as:
  - A toolbar entry (`Orchestrator` / `Spawn Orchestrator` / `Restarting…` / `Spawning…`) with no live activity state.
  - A health banner that appears only when the orchestrator is `!== "ok"` — i.e. on failure/unhealthy, never a positive "actively working" signal.

By construction the board is worker-centric: the orchestrator is deliberately excluded from the list of sessions that get a `SessionCard` (and therefore the status dot), so it never receives an indicator regardless of its activity state.

## Why it matters

While an orchestrator is actively running there is no at-a-glance signal that it is working (vs. idle vs. done). The user has to open the orchestrator to find out. A small positive indicator — same family of tones the workers already use — would make the orchestrator's state visible from the board without opening it, matching how workers already behave.

## Related (not duplicates)

- #2814 — activity pill shows `Idle` while a worker agent is actively working (Kimi missing mid-turn hooks). That is the *activity-state derivation* layer for workers; this request is about the orchestrator *not being rendered with any indicator at all* on the board.
- #1985 — sidebar dot 5-bucket color scheme. About worker dot semantics; does not cover giving the orchestrator an indicator.
- #1563 — killed orchestrator shown as a stale running card. Opposite direction (stale positive signal); this request is about a *live* positive signal for a genuinely running orchestrator.

## Possible directions (not prescribing the fix)

1. Render a dedicated status indicator beside the orchestrator toolbar entry/banner, driven by the orchestrator's activity/status through the same view-mapping already used for workers (keeps tones/legend consistent).
2. Stop excluding the orchestrator from the board's session list and render a card for it too (with a role badge), giving it the same dot treatment as workers by construction.

Either way the goal is the same: a running orchestrator should have a visible "actively working" signal on the board.

Hardik180704:
Closes #3141

## What

- Shows the running orchestrator's activity beside the board toolbar action.
- Reuses the same activity colors and pulse behavior as worker sessions.
- Covers both the in-panel board toolbar and the shell topbar layouts.

## Why

The orchestrator is intentionally excluded from worker cards, so the project board had no quick way to tell whether it was working, idle, or waiting for input.

## Before

Before: orchestrator button without activity indicator

## After

After: orchestrator button with working activity indicator

## Testing

- `npm --prefix frontend test -- --run src/renderer/components/SessionsBoard.test.tsx src/renderer/components/ShellTopbar.test.tsx` — 54 passed after rebasing onto latest `main`
- Full frontend suite — 1,153 passed
- Renderer smoke — 20 passed

## Checklist

- [x] Branched from latest `main`
- [x] One focused change linked to the issue
- [x] Tests cover both toolbar layouts and active/non-active activity states
- [x] Relevant frontend checks pass

### DIFF
--- a/frontend/src/renderer/components/OrchestratorActivityIndicator.tsx
+++ b/frontend/src/renderer/components/OrchestratorActivityIndicator.tsx
@@ -1,0 +1,16 @@
+import type { WorkspaceSession } from "../types/workspace";
+import { getAgentActivityView, getSessionDotView } from "../lib/session-presentation";
+import { cn } from "../lib/utils";
+
+export function OrchestratorActivityIndicator({ session }: { session: WorkspaceSession }) {
+	const activity = getAgentActivityView(session.activity);
+	const dot = getSessionDotView(session);
+
+	return (
+		<span
+			aria-label={`Orchestrator activity: ${activity.label}`}
+			className={cn("size-dot-sm shrink-0 rounded-full", dot.className)}
+			role="status"
+		/>
+	);
+}
--- a/frontend/src/renderer/components/OrchestratorActivityIndicator.tsx
+++ b/frontend/src/renderer/components/OrchestratorActivityIndicator.tsx
@@ -1,16 +1,9 @@
 import type { WorkspaceSession } from "../types/workspace";
-import { getAgentActivityView, getSessionDotView } from "../lib/session-presentation";
+import { getSessionDotView } from "../lib/session-presentation";
 import { cn } from "../lib/utils";
 
 export function OrchestratorActivityIndicator({ session }: { session: WorkspaceSession }) {
-	const activity = getAgentActivityView(session.activity);
 	const dot = getSessionDotView(session);
 
-	return (
-		<span
-			aria-label={`Orchestrator activity: ${activity.label}`}
-			className={cn("size-dot-sm shrink-0 rounded-full", dot.className)}
-			role="status"
-		/>
-	);
+	return <span aria-hidden="true" className={cn("size-dot-sm shrink-0 rounded-full", dot.className)} />;
 }
--- a/frontend/src/renderer/components/SessionsBoard.test.tsx
+++ b/frontend/src/renderer/components/SessionsBoard.test.tsx
@@ -115,6 +115,46 @@ describe("SessionsBoard", () => {
 		expect(screen.getByRole("button", { name: "New task" })).toBeInTheDocument();
 	});
 
+	it.each([
+		["active", "Working", "bg-status-working", true],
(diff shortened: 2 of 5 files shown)


JAYATIAHUJA (approved):
Reviewed the renderer changes and ran the focused local checks. Typecheck passed, and the SessionsBoard/ShellTopbar test files passed locally. The activity indicator wiring is small and consistent with the existing session presentation helpers.\n\nNon-blocking follow-up: because the indicator renders inside an aria-labeled button, consider folding the activity state into the button accessible label if we want screen readers to announce it reliably. LGTM.

Pulkit7070:
Post-merge review (this landed already; recording the assessment for the record).

Reviewed the diff and verified on current main. Clean feature, no issues.

- Consistency by construction: OrchestratorActivityIndicator renders getSessionDotView(session), the exact helper worker dots use, so the orchestrator's activity color and pulse cannot drift from worker sessions. That is the stated goal and it is achieved by reuse rather than a parallel implementation.
- Accessibility is right: the dot is aria-hidden (decorative) and the activity is conveyed semantically through the button aria-label ("Orchestrator, <activityLabel>"), so it is not color-only for assistive tech. When no orchestrator is running it falls back to "Spawn Orchestrator" with no dot.
- Null-safe: getAgentActivityView(activity?) does activity?.state ?? "unknown" and getSessionDotView uses ?? fallbacks, so an orchestrator with missing activity renders the unknown view rather than crashing. Confirmed in session-presentation.ts.
- Both layouts covered (in-panel board toolbar and shell topbar), and the tests are parameterized across activity states asserting both the aria-label and the dot tone/pulse in each layout. I ran src/renderer/components/SessionsBoard.test.tsx and ShellTopbar.test.tsx on current main: 54 pass.

No defects and nothing that warrants a follow-up.

axisrow:
Thanks for adding this — nice parity improvement, the board was missing orchestrator activity feedback for a while.

Hardik180704:
Thanks, glad this was useful 🙂 The missing activity feedback was confusing on the board, so it’s good to know the indicator is helping.

PR #3202, opened 2026-07-28
feat: add Windows source installer

xuelongmu:
## Summary

- add a native PowerShell installer for Windows source checkouts
- build and install the current checkout by default, with fetching gated behind `-Fetch`
- optionally install frontend dependencies or start the source desktop harness
- preserve PATH ordering and verify that `ao` resolves to the installed binary

## Why

The existing `scripts/daemon-build.sh` flow assumes Bash and does not provide a native Windows install path. Windows also cannot replace a running `ao.exe`, so rebuilding from source required several manual stop, build, copy, PATH, and verification steps.

This installer makes that workflow repeatable without fetching unexpectedly. It gracefully stops a running daemon before replacing the executable, installs to `~/.local/bin` by default, avoids duplicate PATH entries, and fails if the resulting `ao` resolution is not the source-built binary.

## User impact

Windows contributors can now run:

```powershell
& .\scripts\install.ps1
```

They can opt into updating `main` with `-Fetch`, reinstall dependencies with `-InstallDependencies`, or immediately launch the source harness with `-Start`.

## Validation

- parsed `scripts/install.ps1` with the PowerShell AST parser
- `go -C backend build ./...`
- `git diff --check`
- ran `& .\scripts\install.ps1` end-to-end on Windows with a live daemon and restored PTY hosts
- ran `& .\scripts\install.ps1 -Start` through dependency repair and Electron source launch
- verified `ao` resolves to `~/.local/bin/ao.exe`, reports the source `dev` version, and the source app starts a ready daemon from that path on port 3001

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,0 +1,181 @@
+[CmdletBinding()]
+param(
+    [switch]$Fetch,
+    [switch]$InstallDependencies,
+    [switch]$Start,
+    [string]$InstallDirectory = (Join-Path $HOME '.local\bin')
+)
+
+$ErrorActionPreference = 'Stop'
+
+function Invoke-Native {
+    param(
+        [Parameter(Mandatory)]
+        [string]$Command,
+
+        [Parameter(ValueFromRemainingArguments)]
+        [string[]]$Arguments
+    )
+
+    & $Command @Arguments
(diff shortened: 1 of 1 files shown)


xuelongmu:
Closing because this work belongs on the origin fork's main line; superseded by xuelongmu/agent-orchestrator#235.

PR #3205, opened 2026-07-28
feat: add daemon startup loader

nikhilachale:
## Summary

- show an animated AO startup loader inside the main content panel while the desktop daemon and workspace data initialize
- preserve the persisted sidebar preference while startup temporarily controls only the rendered layout
- require an authoritative workspace request after daemon readiness, bypassing stale cache entries and rejecting untrusted API reads instead of caching placeholder data
- keep same-port daemon restarts recoverable and ignore cancelled confirmed fetches without flashing a false workspace error
- keep populated board columns visible when daemon startup has already failed, so diagnostics and restart controls remain usable
- avoid welcome-board chrome and empty-state flashes while startup hydration is still in progress
- keep the rotating loader copy visual-only for assistive technology and use a valid themed wordmark size
- report a durable startup failure after 30 seconds with captured daemon stdout/stderr and launch details
- provide a **Restart daemon** action that stops the old process, preserves restart intent across slow shutdowns, and starts a replacement
- continue readiness discovery after the timeout so a slow daemon can still recover automatically
- keep restart failures actionable by surfacing durable error status and retaining retry controls

## Validation

- `npm run typecheck`
- `npm run typecheck:e2e`
- focused startup/board/workspace regressions: 43 passed
- complete frontend Vitest suite: 1,162 passed
- renderer smoke suite: 20 passed
- rendered the exact worktree in the in-app browser and inspected the board layout

### DIFF
--- a/frontend/src/renderer/components/DaemonStartupLoader.tsx
+++ b/frontend/src/renderer/components/DaemonStartupLoader.tsx
@@ -35,9 +35,9 @@ export function DaemonStartupLoader() {
 				<div className="grid h-28 w-32 place-items-center" aria-hidden="true">
 					<img className="ao-startup-logo h-22 w-25 object-contain" src={aoLogo} alt="" />
 				</div>
-				<p className="mt-5 text-md font-semibold tracking-tight text-foreground">Agent Orchestrator</p>
+				<p className="mt-5 text-base font-semibold tracking-tight text-foreground">Agent Orchestrator</p>
 				<p className="mt-2 min-h-5 text-md-sm text-muted-foreground">
-					<span className="ao-startup-status" key={phrase}>
+					<span aria-hidden="true" className="ao-startup-status" key={phrase}>
 						{phrase}
 					</span>
 				</p>
--- a/frontend/src/renderer/components/DaemonStartupLoader.tsx
+++ b/frontend/src/renderer/components/DaemonStartupLoader.tsx
@@ -27,7 +27,7 @@ export function DaemonStartupLoader() {
 			aria-busy="true"
 			aria-label="Agent Orchestrator is starting"
 			aria-live="polite"
-			className="flex h-full min-h-0 items-center justify-center bg-background text-foreground"
+			className="ao-startup-screen flex items-center justify-center bg-background text-foreground"
 			data-testid="daemon-startup-loader"
 			role="status"
 		>
--- a/frontend/src/renderer/components/DaemonStartupLoader.tsx
+++ b/frontend/src/renderer/components/DaemonStartupLoader.tsx
@@ -1,0 +1,52 @@
+import { useEffect, useState } from "react";
+import aoLogo from "../../../assets/ao-logo.svg";
+
+const STARTUP_PHRASES = [
+	"Starting local services",
+	"Connecting to the daemon",
+	"Loading workspaces",
+	"Preparing your board",
+] as const;
+
+const PHRASE_INTERVAL_MS = 2_200;
+
+export function DaemonStartupLoader() {
+	const [phraseIndex, setPhraseIndex] = useState(0);
+
+	useEffect(() => {
+		const timer = window.setInterval(() => {
+			setPhraseIndex((current) => (current + 1) % STARTUP_PHRASES.length);
+		}, PHRASE_INTERVAL_MS);
+		return () => window.clearInterval(timer);
+	}, []);
+
+	const phrase = STARTUP_PHRASES[phraseIndex];
+
+	return (
+		<div
+			aria-busy="true"
+			aria-label="Agent Orchestrator is starting"
+			aria-live="polite"
+			className="flex h-full min-h-0 items-center justify-center bg-background text-foreground"
+			data-testid="daemon-startup-loader"
+			role="status"
+		>
+			<div className="ao-startup-content flex -translate-y-[3vh] flex-col items-center text-center">
+				<div className="grid h-28 w-32 place-items-center" aria-hidden="true">
+					<img className="ao-startup-logo h-22 w-25 object-contain" src={aoLogo} alt="" />
+				</div>
+				<p className="mt-5 text-md font-semibold tracking-tight text-foreground">Agent Orchestrator</p>
+				<p className="mt-2 min-h-5 text-md-sm text-muted-foreground">
+					<span className="ao-startup-status" key={phrase}>
+						{phrase}
+					</span>
+				</p>
+				<div className="ao-startup-dots mt-3 flex h-4 items-center gap-1.5" aria-hidden="true">
+					<span />
+					<span />
+					<span />
+				</div>
+			</div>
+		</div>
+	);
+}
--- a/frontend/src/renderer/components/DaemonStartupLoader.tsx
+++ b/frontend/src/renderer/components/DaemonStartupLoader.tsx
@@ -27,7 +27,7 @@ export function DaemonStartupLoader() {
 			aria-busy="true"
 			aria-label="Agent Orchestrator is starting"
 			aria-live="polite"
-			className="ao-startup-screen flex items-center justify-center bg-background text-foreground"
+			className="ao-startup-screen flex h-full w-full items-center justify-center bg-background text-foreground"
 			data-testid="daemon-startup-loader"
 			role="status"
 		>
--- a/frontend/src/renderer/components/SessionsBoard.tsx
+++ b/frontend/src/renderer/components/SessionsBoard.tsx
@@ -34,7 +34,7 @@ import {
 import { useSessionScmSummary, type SessionPRSummary } from "../hooks/useSessionScmSummary";
 import { useRestoreSession } from "../hooks/useRestoreSession";
 import { useTerminateSession } from "../hooks/useTerminateSession";
-import { useWorkspaceQuery, workspaceQueryKey } from "../hooks/useWorkspaceQuery";
+import { usesPreviewWorkspaceData, useWorkspaceQuery, workspaceQueryKey } from "../hooks/useWorkspaceQuery";
 import { NotificationCenter } from "./NotificationCenter";
 import { BoardWelcome, ProjectBoardEmpty } from "./BoardEmptyStates";
 import { OrchestratorIcon } from "./icons";
@@ -138,7 +138,7 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
 	// query has resolved, so the welcome never flashes over real data): the
 	// global board teaches the app before any project exists, and a fresh
 	// project board invites the first task instead of showing four zeros.
-	const isDaemonReady = shell ? shell.daemonStatus.state === "ready" : true;
+	const isDaemonReady = usesPreviewWorkspaceData || (shell ? shell.daemonStatus.state === "ready" : true);
 	const workspaceStartupState = shell?.workspaceStartupState ?? "ready";
 	const isLoaded = isDaemonReady && workspaceStartupState === "ready" && workspaceQuery.isSuccess;
 	const showStartup =
--- a/frontend/src/renderer/components/SessionsBoard.tsx
+++ b/frontend/src/renderer/components/SessionsBoard.tsx
@@ -51,6 +51,8 @@ import { useUiStore } from "../stores/ui-store";
 import { RestoreUnavailableDialog } from "./RestoreUnavailableDialog";
 import { Tooltip, TooltipContent, TooltipTrigger } from "./ui/tooltip";
 import { SessionTerminationDialog } from "./SessionTerminationDialog";
+import { DaemonStartupLoader } from "./DaemonStartupLoader";
+import { useShellMaybe } from "../lib/shell-context";
 
 type SessionsBoardProps = {
 	/** When set, the board shows only this project's sessions. */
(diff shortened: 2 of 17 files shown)


nikhilachale:
Follow-up for the renderer-smoke failure: the repeated notifications-stream proxy refusal was non-fatal background noise. The actual regression was that three `dev:web` smoke specs intentionally use built-in mock workspaces without installing a ready daemon bridge, while the new production readiness gate kept that preview mode on the loader.

The follow-up commit bypasses daemon readiness only when `VITE_NO_ELECTRON=1`; packaged Electron still waits for daemon readiness plus a confirmed workspace fetch. Verified locally with the full renderer smoke suite: 20/20 passing while the same proxy warnings remained, plus `npm run typecheck`.

nikhilachale:
https://github.com/user-attachments/assets/8e2f6af9-9042-4194-aa86-282af1dd2b2f

codebanditssss (approved):
lgtm

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -135,7 +139,12 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
 	// query has resolved, so the welcome never flashes over real data): the
 	// global board teaches the app before any project exists, and a fresh
 	// project board invites the first task instead of showing four zeros.
-	const isLoaded = workspaceQuery.isSuccess;
+	const isDaemonReady = usesPreviewWorkspaceData || (shell ? shell.daemonStatus.state === "ready" : true);
```
illegalcall, line 142: `DaemonStatus.state` has four values (`starting | ready | stopped | error`), and `!isDaemonReady` is the first disjunct of `showStartup` on line 145. So the loader renders for `stopped` and `error` too, not just for boot.

`_shell.tsx:172` in this same PR gets this right: `isStartupLoading` includes `!daemonStatus.code`, so a reported failure is not treated as loading and the sidebar stays available. The board doesn't apply that guard, so the two now disagree about what "starting up" means.

Concretely: the daemon dies mid-session and the supervisor publishes `{ state: "stopped", code: "exited" }` (`frontend/src/main.ts`). The shell correctly leaves startup mode, but the content pane keeps cycling "Starting local services / Connecting to the daemon / Loading workspaces / Preparing your board" indefinitely over a board that had real sessions in it. For the codes the new Restart button doesn't cover (`binary_missing`, `datadir_unwritable`, `not_configured`, `identity_mismatch`) there's no way out of it at all.

Reusing the shell's predicate fixes it, either by passing `isStartupLoading` down through the shell context or by applying the same `code` check here:

```ts
const daemonIsFailed = Boolean(shell?.daemonStatus.code);
const showStartup =
    shell !== null &&
    !daemonIsFailed &&
    (!isDaemonReady || workspaceStartupState === "loading" || (!workspaceQuery.isSuccess && !workspaceQuery.isError));
```

Worth a test alongside it: board with `{ state: "stopped", code: "exited" }` plus a populated workspace query should render the columns, not `daemon-startup-loader`. The new case in `board-empty-states.test.tsx` only covers `state: "starting"`.

frontend/src/renderer/routes/_shell.tsx
```diff
@@ -385,6 +454,21 @@ function ShellLayout() {
...
+	useEffect(() => {
+		if (isStartupLoading) {
+			sidebarWasCollapsedForStartupRef.current = true;
+			cancelSidebarPeekClose();
+			setIsSidebarPeekOpen(false);
+			return;
+		}
+		if (!sidebarWasCollapsedForStartupRef.current) return;
+		sidebarWasCollapsedForStartupRef.current = false;
+		if (!isSidebarOpen) toggleSidebar();
```
illegalcall: This overwrites the preference of anyone who deliberately keeps the sidebar collapsed.

`isSidebarOpen` is persisted: `toggleSidebar` writes to `localStorage` under `sidebarStorageKey` (`stores/ui-store.ts:176-178`), and `initialSidebarOpen()` reads it back at launch (`ui-store.ts:144`). So for a user whose saved preference is collapsed, every launch goes:

1. `isStartupLoading` is true, so `sidebarWasCollapsedForStartupRef` is set to true.
2. Startup finishes. The ref is true, and `isSidebarOpen` is `false` — that's their own saved preference, not something startup did.
3. `toggleSidebar()` fires, opens the sidebar, and writes `true` back to localStorage.

The preference is overwritten on disk rather than just for the session, so a collapsed sidebar can never stick across launches.

The root cause is that the ref's name doesn't match what actually happened: nothing collapsed the sidebar. Line 656 only overrides the `open` prop and leaves the store value untouched, so by the time this branch runs there's no way to tell "we hid it for startup" from "the user wants it hidden."

Since line 656 already reverts to `isSidebarOpen || isSidebarPeekOpen` the moment `isStartupLoading` goes false, the restore branch looks unnecessary — dropping lines 467-469 and the ref should give the intended behavior on its own. If an explicit restore is wanted, capture `isSidebarOpen` when startup begins and set it back to that value instead of forcing it open.

Possibly related: `409965f05 Revert "fix: preserve sidebar state during startup"`.

frontend/src/renderer/components/DaemonStartupLoader.tsx
```diff
@@ -0,0 +1,52 @@
...
+			aria-live="polite"
+			className="ao-startup-screen flex h-full w-full items-center justify-center bg-background text-foreground"
+			data-testid="daemon-startup-loader"
+			role="status"
+		>
+			<div className="ao-startup-content flex -translate-y-[3vh] flex-col items-center text-center">
+				<div className="grid h-28 w-32 place-items-center" aria-hidden="true">
+					<img className="ao-startup-logo h-22 w-25 object-contain" src={aoLogo} alt="" />
+				</div>
+				<p className="mt-5 text-md font-semibold tracking-tight text-foreground">Agent Orchestrator</p>
```
illegalcall: `text-md` isn't a utility in this theme, so this element ends up with no font-size at all.

`styles.css` `@theme inline` defines `--text-xs`, `--text-sm`, `--text-base`, `--text-sm-md`, `--text-md-sm`, `--text-control`, `--text-brand` and the heading sizes. There's no `--text-md`, and Tailwind v4 has no `text-md` default either. No other component in `src/renderer` uses the class.

So the wordmark silently falls back to the inherited size. Line 39 just below uses the real `text-md-sm`, which makes this read like a typo of that one.

frontend/src/renderer/components/DaemonStartupLoader.tsx
```diff
@@ -0,0 +1,52 @@
...
+		return () => window.clearInterval(timer);
+	}, []);
+
+	const phrase = STARTUP_PHRASES[phraseIndex];
+
+	return (
+		<div
+			aria-busy="true"
+			aria-label="Agent Orchestrator is starting"
+			aria-live="polite"
```
illegalcall, line 29: `aria-live="polite"` on this container, combined with text that rotates every 2.2s modulo 4, means a screen reader announces all four phrases and then starts over, for as long as the loader stays mounted.

The phrases run on a fixed timer with no relationship to `daemonStatus.state`, so the announcements aren't reporting real progress. And while the board can still reach this component in a failed-daemon state (see the `SessionsBoard.tsx:142` comment), they never stop.

Either `aria-hidden` the rotating span and let the static `aria-label` on this element carry the announcement, or drive the copy from the real daemon status so the sequence is both finite and accurate.

frontend/src/renderer/routes/_shell.tsx
```diff
@@ -369,6 +381,63 @@ function ShellLayout() {
...
+	// ready. React Query keeps polling and the event transport may invalidate
+	// the workspace query later, so let a newer successful result recover the
+	// shell without requiring a daemon restart or port change.
+	useEffect(() => {
+		if (
+			usesPreviewWorkspaceData ||
+			daemonStatus.state !== "ready" ||
+			workspaceStartupState === "ready" ||
+			!workspaceQuery.isSuccess ||
+			workspaceQuery.dataUpdatedAt <= workspaceStartupBaselineRef.current
```
illegalcall, line 432: Two things about this comparison.

`dataUpdatedAt` is a `Date.now()` millisecond stamp, so if the confirmed fetch resolves within the same millisecond as the value already sitting in the cache, this returns early and the loader stays up until the next `refetchInterval` tick (15s) or an event-transport invalidation. Unlikely with two loopback GETs plus the JSON mapping, but the tests step 100 -> 200 and never exercise equality, so nothing would catch it if it did happen.

The larger point: this accepts *any* newer successful stamp, not specifically the fetch the effect above issued. That's correct today only because of two properties that live in other files:

1. `useDaemonStatus` calls `applyDaemonStatus` (which sets the base URL) before `setStatus`, so any refetch starting after the ready render is a real request.
2. The untrusted path in `fetchWorkspaces` is a bare `return []` with no `await` ahead of it, so a placeholder fetch resolves within a microtask and can't still be in flight for `queryClient.fetchQuery` to dedupe into. If it could, that call would resolve `[]`, stamp a newer `dataUpdatedAt`, and flip the state to `ready` with exactly the empty list this PR exists to distrust.

Add a single `await` before the trust check in `fetchWorkspaces` and the original bug comes back silently. Both new shell tests mock `fetchQuery` and `getQueryState`, so neither property is covered.

Cleanest fix is upstream: have `fetchWorkspaces` throw instead of returning `[]` when there's no trusted base URL, which removes the ambiguity this whole mechanism is working around. Failing that, a comment at both sites recording the dependency would at least make it survive the next refactor.

illegalcall (commented):
Reviewed at `367aa87`. Scoping the loader to the content pane and gating `isStartupLoading` on `!daemonStatus.code` both look right. Five things inline.

illegalcall (commented):
Two more from a cross-model pass, both in the same effect. These are about the confirmed fetch not doing what it looks like it does.

frontend/src/renderer/routes/_shell.tsx
```diff
@@ -369,6 +381,63 @@ function ShellLayout() {
...
+			return () => {
+				active = false;
+			};
+		}
+
+		workspaceStartupBaselineRef.current =
+			queryClient.getQueryState(workspaceQueryKey)?.dataUpdatedAt ?? 0;
+		setWorkspaceStartupState("loading");
+		void queryClient
+			.fetchQuery(workspaceQueryOptions)
```
illegalcall: `fetchQuery` respects `staleTime`, and the shared client sets `staleTime: 10_000` (`lib/query-client.ts`). So on the exact path this effect exists to handle, it usually doesn't fetch anything.

Cold start: the route loader at line 49 runs `ensureQueryData` before a port is known, `fetchWorkspaces` returns `[]` through the `hasTrustedApiBaseUrl()` guard, and that lands in the cache at t~0. The daemon reports ready a second or two later, this effect runs, and the cached `[]` is still fresh — so `fetchQuery` returns it without contacting the daemon. `dataUpdatedAt` never moves, and the recovery effect's `dataUpdatedAt > baseline` test can't pass.

What actually unblocks the board today is the agent-catalog effect at line 499, which invalidates `workspaceQueryKey` at line 506 on the same `[daemonStatus.port, daemonStatus.state, queryClient]` change and forces a real refetch. That's an unrelated effect this gate is silently depending on.

And it doesn't always fire. Line 501 early-returns when `agentCatalogPortRef.current === daemonStatus.port`, so a daemon that restarts and comes back on the *same* port skips the invalidation — which is the normal case in dev, where `DEV_DAEMON_PORT` / `AO_PORT` are fixed. In that case this `fetchQuery` short-circuits on the still-fresh cache too, nothing advances `dataUpdatedAt`, and the loader sits until the 15s `refetchInterval` fires.

Opting out of `staleTime` here would make the confirmed fetch actually confirm something:

```ts
void queryClient
    .fetchQuery({ ...workspaceQueryOptions, staleTime: 0 })
    .catch(...)
```

frontend/src/renderer/routes/_shell.tsx
```diff
@@ -369,6 +381,63 @@ function ShellLayout() {
...
+			.catch(() => {
```
illegalcall: This `catch` can't distinguish a failed request from a cancelled one, and something in this same render cancels it.

The agent-catalog effect at line 499 has the identical dependency array (`[daemonStatus.port, daemonStatus.state, queryClient]`) and is declared after this one, so it runs immediately after it on every ready-port change. Line 506 calls `queryClient.invalidateQueries({ queryKey: workspaceQueryKey })`, which defaults to `cancelRefetch: true` — it aborts the in-flight fetch this effect just started and begins a replacement. The cancelled promise rejects, this `catch` fires, and `workspaceStartupState` becomes `"error"` while a perfectly healthy refetch is still in progress. The board then renders "Could not load sessions." (`SessionsBoard.tsx:331`).

The recovery effect heals it once the replacement succeeds, so this surfaces as an error flash rather than a stuck state, but it's a visible one whenever the loopback response is slow enough for the two effects to overlap.

Narrowing the catch would fix it, roughly:

```ts
.catch((error) => {
    if (!active || error instanceof CancelledError) return;
    setWorkspaceStartupState("error");
});
```

Or move the `workspaceQueryKey` invalidation ahead of this fetch instead of after it.

One caveat: the `cancelRefetch: true` default and the rejection-on-cancel behavior are React Query semantics I'm asserting rather than something I ran here, so worth confirming against the installed v5 version before reworking it. The ordering and the invalidation call itself are straight from this file.

nikhilachale:
Addressed all review comments in 176997108:
- Forced the confirmed workspace fetch with staleTime: 0.
- Removed the unrelated workspace invalidation that could cancel the fetch.
- Ignored genuine React Query cancellations to prevent false error flashes.
- Made readiness depend on the confirmed fetch rather than timestamp advancement.
- Untrusted API reads now fail instead of caching a placeholder empty result.
- Added same-port daemon restart coverage.
- Preserved the user’s persisted collapsed-sidebar preference.
- Prevented failed daemon states from hiding an already-populated board.
- Fixed the invalid loader font utility.
- Hid rotating loader phrases from screen readers while retaining the static status announcement.

PR #3207, opened 2026-07-28
feat: add controlled browser preview workflow

nikhilachale:
## Summary

- keep Browser closed by default and treat URLs printed in terminal output as signal-only, showing the Browser indicator without navigating or stealing focus
- preserve explicit terminal-link navigation while safely serializing and coalescing preview requests
- open web links in AO Browser only for active workers; standalone shell, orchestrator, and terminated-session links continue to use the system browser
- extend the terminal-link context menu with Open in system browser and Copy link actions
- refresh explicitly selected static HTML, Markdown, and related assets only after worker activity becomes idle, waits for input, or exits
- never choose a preview target for a fresh worker; selection remains owned by `ao preview`, a managed preview server, or deliberate user navigation
- avoid replayed terminal output opening live previews
- document the rendered-preview workflow for framework projects: use the exact session worktree, start or reuse its development server, wait for a healthy URL, and run `ao preview <url>` near the end of UI work
- add an explicit Browser toolbar action for opening the current page externally
- merge current `main` and resolve the preview poller and Browser panel conflicts

## Review fixes

- guard every poller target mutation and refresh with a URL + target-revision compare-and-set so a stale session snapshot cannot overwrite an explicit user selection
- persist the exact revision produced by a poller-owned missing-file clear, and restore only while the blank target still has that revision
- separate explicit target revisions from static-content refresh revisions; automatic refreshes now reload the Browser's current location instead of navigating back to the selected entrypoint
- replace the per-session 250 ms recursive asset walk with a complete, uncapped fingerprint at a five-second cadence while retaining cheap entry metadata checks on each poll
- preserve current `main`'s explicit-preview ownership instead of reintroducing automatic entrypoint discovery
- restore system-browser fallback for terminals that do not have an AO Browser inspector
- route the Browser toolbar's external action through the shared error-handling policy
- remove a stale Browser layout wrapper left over from the retired viewport selector

## Tests

- `npm run lint` (full backend tests + golangci-lint)
- `cd backend && go test -race ./internal/preview ./internal/service/session ./internal/storage/sqlite/store`
- `cd frontend && npm test`
- regenerated and verified sqlc, OpenAPI, and frontend API artifacts with the repository-pinned generators

## Notes

- framework applications still require their project development server; worker guidance selects the relevant route and invokes `ao preview`
- static file previews are daemon-driven after explicit selection and refreshed only when active worker activity finishes


backend/internal/preview/poller.go
```diff
@@ -141,24 +152,39 @@ func (p *Poller) Poll(ctx context.Context) error {
...
+		current := strings.TrimSpace(sess.Metadata.PreviewURL)
+		// Recursively fingerprint assets only while this exact workspace-owned
+		// static preview is active. Hidden entries and external dev servers do
+		// not need a full workspace walk on every poll.
+		state := stateFor(entry, workspaceOwned && current == target)
+		if seenBefore && previous == state {
+			continue
+		}
+		pending := seenBefore && previous.pending
+		if !p.shouldRefresh(sess, target, seenBefore, workspaceOwned, pending) {
```
whoisasx, line 211: `shouldRefresh` validates ownership against the `ListAllSessions` snapshot, but this `SetPreview` is unconditional. If the user selects a dev-server URL after that snapshot and before this call, the stale static target overwrites the newer explicit selection; the recursive scan widens that window. Please make this update conditional on the expected preview revision and URL, or expose a CAS-style service operation. This race was reproduced in a focused regression probe.

backend/internal/preview/poller.go
```diff
@@ -121,7 +128,11 @@ func (p *Poller) Poll(ctx context.Context) error {
 			// Only restore an entry after the poller itself cleared it because
 			// the selected file temporarily disappeared. A blank fresh session
 			// or an explicit user clear must remain blank.
-			if !seenBefore || !previous.cleared || previous.path == "" {
+			if strings.TrimSpace(sess.Metadata.PreviewURL) != "" ||
```
whoisasx, line 151: After the poller clears a missing file, `entryState` remembers only `cleared`, not the revision produced by that clear. If the user subsequently runs an explicit `ao preview clear`, the URL remains blank while its revision advances; recreating the file still enters this restoration path and resurrects a preview the user deliberately cleared. Record the poller-owned clear revision and restore only while the current blank state still matches it.

backend/internal/preview/tree_state.go
```diff
@@ -0,0 +1,61 @@
...
+
+// staticTreeSignature fingerprints the files served from the selected entry's
+// directory. A preview rooted at dist/index.html watches dist/**, while a root
+// index.html watches the workspace tree. Framework dev-server URLs never reach
+// this path; their own HMR remains authoritative.
+func staticTreeSignature(entry Entry) uint64 {
+	root := filepath.Dir(entry.AbsPath)
+	hash := fnv.New64a()
+	seen := 0
+	_ = filepath.WalkDir(root, func(filePath string, item fs.DirEntry, walkErr error) error {
```
whoisasx, line 17: This recursive walk runs every 250 ms for every active static-preview session, visiting up to 5,000 files each time and doing so sequentially across sessions. Besides significant steady filesystem load, the cutoff means changes after the first 5,000 entries are never detected. Please use a filesystem watcher or a substantially coarser/debounced scan with complete, explicit coverage.

backend/internal/preview/poller.go
```diff
@@ -141,24 +152,39 @@ func (p *Poller) Poll(ctx context.Context) error {
...
+		if seenBefore && previous == state {
+			continue
+		}
+		pending := seenBefore && previous.pending
+		if !p.shouldRefresh(sess, target, seenBefore, workspaceOwned, pending) {
 			p.seen[sess.ID] = state
 			continue
 		}
-		if !p.shouldRefresh(sess, target, seenBefore, workspaceOwned) {
+		automaticRefresh := current == target || restoringCleared || pending
```
whoisasx, line 215: Using `SetPreview` as the refresh signal increments `previewRevision`. The existing Browser view interprets a new revision as target navigation, so a user who deliberately navigated elsewhere inside the panel is forced back to this static entry on the next asset change. Refresh/reload should be separate from target selection, or explicit in-panel navigation must update the ownership generation.

whoisasx (changes requested):
Requesting changes for four correctness and scalability issues in preview ownership and refresh handling. The controlled terminal-link workflow, fresh-worker behavior, external-open support, and prompt/documentation changes otherwise look coherent. Focused backend/frontend tests, race tests, typecheck, and CI pass, but focused regression probes reproduced the ownership failures described inline.

backend/internal/preview/poller.go
nikhilachale: Fixed in a43d25e1f. Poller target updates and refreshes now use service/store compare-and-set operations guarded by the expected preview URL and target revision. A stale snapshot is rejected without touching the newer selection; TestPollerStaleSnapshotCannotReplaceNewExplicitTarget covers the race.

backend/internal/preview/poller.go
nikhilachale: Fixed in a43d25e1f. The clear CAS returns its exact committed revision via SQLite RETURNING, entryState records it, and restoration requires the current blank target to retain that revision. TestPollerDoesNotRestoreAfterExplicitClearAdvancesRevision covers explicit clear after the poller clear.

backend/internal/preview/tree_state.go
nikhilachale: Fixed in a43d25e1f. Recursive asset fingerprints now run at a separate five-second cadence instead of every 250 ms, and the 5,000-file cutoff was removed for complete coverage. Tests cover rate limiting and a changed asset beyond the former cutoff.

backend/internal/preview/poller.go
nikhilachale: Fixed in a43d25e1f. Explicit target selection keeps previewRevision, while automatic static-content changes bump a new previewRefreshRevision. The renderer handles that signal with Browser reload on the current location and does not navigate back to previewUrl; the hook regression test covers in-panel navigation.

nikhilachale:
already done in another PR

PR #2860, opened 2026-07-20
fix(preview): don't auto-preview Markdown docs for brand-new sessions

i-trytoohard:
Fixes #2859

## Summary

A new worker session's in-app browser was opening non-empty — navigated to an arbitrary repo Markdown/HTML file (e.g. `README.md`) instead of starting empty. Reported by @prateek: new sessions showed URLs like `…/sessions/safesplit-10/preview/files/website-setUp-for-links.md`.

**Root cause:** the preview poller (`backend/internal/preview/poller.go`) auto-discovers a workspace file on its very first 250 ms scan and sets `preview_url` for every active worker session. `DiscoverEntry`'s `mostRecentPreviewable` fallback (`entry.go`) returns the most-recently-modified `.md`/`.html` file anywhere in the workspace, so nearly every new session in a Markdown-rich repo matched and got auto-previewed — directly contradicting the expectation that a new session's browser starts empty.

## The fix

Restrict the **never-previewed** (`workspaceOwned == false`) auto-discovery path to real static-frontend entrypoints (`index.html` and its `public/`/`dist/`/`build/` variants). Drop the `.md`/`.html` `mostRecentPreviewable` fallback for that path only.

- New `DiscoverWebEntrypoint(workspacePath)` (`entry.go`) — entrypoint candidates only, no document fallback.
- `DiscoverEntry` is refactored to delegate to `DiscoverWebEntrypoint` first, then `mostRecentPreviewable` — **behavior-preserving** for its other callers (the explicit `ao preview` HTTP path at `sessions.go:733`, and the `entry_test.go` suite).
- In the poller, the fallback after `EntryAtPath` fails now branches on `workspaceOwned`: already-previewed sessions keep full `DiscoverEntry` (so their preview stays fresh as files change, including `.md`), while never-previewed sessions use `DiscoverWebEntrypoint`.

**Behavior after the fix:**
| Session state | Workspace has | Browser on creation |
| --- | --- | --- |
| New (never previewed) | only `.md` docs | **empty** (fixed) |
| New (never previewed) | `dist/index.html` | auto-previews the frontend (preserved) |
| Explicitly previewed (`ao preview X`) | `X` changed | refreshed (unchanged) |
| Explicitly previewed, file deleted | other `.md` | full re-discovery (unchanged) |

This is the "surgical alternative" noted in the issue rather than a blanket opt-in: it fixes both reported cases (`.md` repo docs) while preserving the poller's primary purpose of auto-surfacing built static frontends.

## Test

[...]

PR #3213, opened 2026-07-28
fix(preview): keep poller previews entrypoint-only

Vaibhaav-Tiwari:
@nikhilachale addressed your comment on PR #2860 as it was merged before i could push it

 ## Summary

  - Track preview URLs created automatically by the preview poller separately from explicit user-selected previews.
  - Prevent poller-owned previews from falling back to Markdown/report files after the auto-discovered `index.html` is deleted.
  - Preserve document fallback behavior for explicit workspace previews.
  - Add regression coverage for the two-poll delete case plus a preservation test for explicit preview fallback.

  ## Tests

  - `go test ./internal/preview -count=1`
  - `go test ./internal/httpd/controllers -run SetPreview -count=1`
  - `git diff --check HEAD~1..HEAD`

### DIFF
--- a/backend/internal/preview/poller.go
+++ b/backend/internal/preview/poller.go
@@ -42,6 +42,9 @@ type entryState struct {
 	path    string
 	modUnix int64
 	size    int64
+	// pollerOwned is true when the poller created the current workspace preview
+	// automatically, as opposed to the user explicitly choosing it via ao preview.
+	pollerOwned bool
 	// cleared is set when the poller itself cleared the preview URL because the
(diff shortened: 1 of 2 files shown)


nikhilachale:
@Vaibhaav-Tiwari  i was fixing some changes in the browser part and i already fixed that thats why i merged it 
and i fixed and already raised a PR for that

nikhilachale (commented):
you can check #3207 i already fixed it and also added some more fixes in it

Issue #3203, opened 2026-07-28
Diff viewer shows 0 changes when session branch has committed work (compares vs HEAD instead of merge-base)


i-trytoohard:
## Problem

When an agent creates a PR from a session branch, its changes are committed to the branch. The desktop diff viewer shows **0 files changed** / empty diffs for these committed changes.

## Root cause

All three git operations that power the workspace diff viewer compare against **`HEAD`** (the current commit), not the branch's merge base:

| File | Line | Command | Problem |
|------|------|---------|---------|
| `backend/internal/service/session/workspace_files.go` | 412 | `git status --porcelain` | Compares working tree vs HEAD → clean working tree = no files shown |
| same | 449 | `git diff --numstat HEAD` | Uncommitted changes only → committed changes invisible |
| same | 507 | `git diff --no-ext-diff --unified=3 HEAD` | Same — only uncommitted diff |

Once the agent commits and the working tree is clean, **everything resolves to "unmodified" against HEAD**, so the file list, numstat counts, and per-file diffs all come back empty.

## Expected behavior

The diff viewer should show the **full diff from the merge base** — the point where the session branch diverged from its upstream/base branch (typically `main`). This is the same view GitHub shows on a PR diff tab.

## Suggested approach

1. **Resolve the base ref** for the session branch. Options:
   - The upstream tracking branch (`git rev-parse --abbrev-ref @{upstream}`)
   - The PR base branch (if known from session/PR metadata)
   - Fall back to `origin/main` or a configured default branch
2. **Compute the merge base**: `git merge-base HEAD <base-ref>`
3. **Use the merge-base as the diff target** instead of `HEAD`:
   - `git diff --numstat <merge-base>` for the summary counts
   - `git diff <merge-base>` for the per-file diff
   - `git diff --name-status <merge-base>` for file status

## Environment

- Reproduced in discussion (Maaz + Vaibhaav + Prateek, bug-triage thread)
- Affects all sessions where the agent commits before the user opens the diff viewer

Vaibhaav-Tiwari:
Already on it

somewherelostt:
Thanks for contributing to Agent Orchestrator.

This issue is being picked up by the current external contributor on-call pair:

- @illegalcall
- @pulkit7070

If someone is already working on this, please continue as usual.
The on-call pair is added for visibility, tracking, and support, not to take over the work.
[...]

PR #3190, opened 2026-07-27
fix: correct synthetic new-file diff line count

AgentWrapper:
## Summary
- Drop the sentinel empty split element when building synthetic diffs for untracked added files ending in a newline
- Add regression coverage for new-file diffs with and without a trailing newline, including addition counts matching the hunk

Fixes #2923

## Tests
- cd backend && go test ./internal/service/session
- cd backend && env -u AO_RUNTIME_LAUNCH_ID -u AO_SESSION_ID -u AO_PROJECT_ID go test ./internal/cli
- cd backend && env -u AO_RUNTIME_LAUNCH_ID -u AO_SESSION_ID -u AO_PROJECT_ID go test ./...

## Notes
- A plain `go test ./...` in this AO worker environment failed in `backend/internal/cli` because inherited `AO_RUNTIME_LAUNCH_ID`/`AO_SESSION_ID`/`AO_PROJECT_ID` variables changed CLI test expectations; rerunning with those variables unset passed.

Issue #2923, opened 2026-07-21
[Bug]: New-file diff shows an extra trailing blank line and an inflated added-line count


dobbydobap:
### Summary

The diff shown for a newly-added (untracked) text file in the session workspace file viewer has an off-by-one: the hunk header counts one line too many and an extra empty `+` line is appended, for any file whose content ends in a newline (i.e. almost every file). The added-line count in the diff also disagrees with the addition count shown for the same file in the changed-files list.

### Steps to reproduce

1. In a session workspace, create a new text file that ends with a trailing newline — e.g. contents `hello\nworld\n` — and leave it untracked (don't `git add` it).
2. Open that file in the session's workspace file / changes viewer.
3. Look at the diff and at the file's addition count in the changed-files list.

### Expected behavior

The synthetic diff matches what `git diff` produces for a new file: header `@@ -0,0 +1,2 @@`, exactly two `+` lines (`+hello`, `+world`), and no trailing blank `+` line. The hunk's added-line count equals the addition count shown for the same file in the changed-files list.

### Actual behavior

The diff is:

```
@@ -0,0 +1,3 @@
+hello
+world
+
```

— the header claims **3** added lines, and a phantom empty `+` line is appended. The changed-files list, however, shows **+2** for the same file, so the two counts contradict each other.

### AO version

Source checkout — fork of `AgentWrapper/agent-orchestrator`, `main` @ commit `6e54247`. Not a released binary.

### Environment

Windows 11 (x86-64), Node 24.18, Go 1.25.x. The defect is in server-side Go string handling (`backend/internal/service/session`), so it's platform-independent — reproduces on any OS/arch.

### Likely cause

`syntheticAddedFileDiff` builds the diff for untracked added files with `strings.SplitAfter(content, "\n")` (`backend/internal/service/session/workspace_files.go:341`). For content ending in `\n`, `SplitAfter` returns a trailing empty element (`"hello\nworld\n"` → `["hello\n", "world\n", ""]`). The guard on the next line only drops it when the whole file is empty, so the stray `""` is counted in the hunk header (`:350`) and emitted as a blank `+` line (`:351-357`).

`countUntrackedTextLines` (`:461`) counts additions correctly, which is why the changed-files badge (`+2`) and the diff header (`+3`) disagree for the same file.

### Suggested fix

Drop a single trailing empty element after the split (subsumes the empty-file guard, so empty-file behavior is unchanged):

```go
[...]

Issue #2172, opened 2026-06-24
Support multi-PR reviews per worker session


neversettle17-101:
Migrated from aoagents/ReverbCode/issues/395

Source: https://github.com/aoagents/ReverbCode/issues/395
Original assignees: neversettle17-101

---

## Summary
Extend AO reviews so one worker session can have multiple PRs reviewed by a single shared reviewer terminal. Keep `review` session-level because it represents the reviewer terminal/handle for the worker session. Store PR-specific review state in `review_run`.

Split the work into two PRs:
1. Backend/storage/API/CLI support for multi-PR review execution.
2. Frontend UI for showing eligible and up-to-date PR review state.

## PR 1: Backend Multi-PR Review Support
- Keep `review` as one row per worker session:
  - `session_id` remains unique.
  - `reviewer_handle_id` remains the single reviewer terminal handle.
  - `review.pr_url` should no longer drive behavior; leave it as legacy/deprecated or stop updating it.
- Make `review_run` the PR-scoped record:
  - Always persist `pr_url`.
  - Add a migration replacing uniqueness with `(session_id, pr_url, target_sha)` where `target_sha != "" AND status != "failed"`.
  - Rename store lookup from session/SHA to session/PR/SHA.
- Add a pure planner in `backend/internal/review`:
  - Output per-PR items with `prUrl`, `prNumber`, `targetSha`, `status`, `latestRun`.
  - Status values: `needs_review`, `running`, `up_to_date`, `changes_requested`, `ineligible`.
  - Eligible PRs are open, non-draft, and not already AO-approved for the current `pr_url + head_sha`.
- Update trigger flow:
  - `Trigger(sessionID)` evaluates all worker PRs.
  - Create runs for `needs_review` and current-head `changes_requested`.
  - Reuse/skip `running` and `up_to_date`.
  - Spawn one reviewer terminal if needed, then send one review task per created run to that same terminal.
  - Return batch results with reviewer handle, items, and created runs.
- Update submit flow:
  - Keep `ao review submit --session --run ...`.
  - `runID` identifies the PR through stored `review_run.pr_url`.
  - Worker changes-requested messages include the PR URL and GitHub review id.
- Add API fields in an additive way:
  - `ListReviewsResponse.items`.
  - `TriggerReviewResponse.items`.
  - `TriggerReviewResponse.createdReviews`.
  - Keep `reviewerHandleId` and `reviews` history for compatibility.
[...]

PR #3214, opened 2026-07-28
fix(session): make workspace diffs base-aware

Vaibhaav-Tiwari:
Fixes https://github.com/AgentWrapper/agent-orchestrator/issues/3203

  This PR makes the session Files diff viewer compare against the session branch’s base instead of only comparing against HEAD. Previously, once an agent
  committed its work, the working tree became clean and the desktop diff viewer could show 0 files changed even though the session branch contained changes.

  Changes included:

  - Persist diff_base_sha and diff_base_ref when spawning single-repo sessions.
  - Use the recorded base for workspace file lists, numstat counts, and per-file diffs.
  - Fall back to PR base SHA, then default-branch merge-base, then HEAD only when no base can be resolved.
  - Include committed branch changes, uncommitted changes, deletions, renames, and workspace-project child repo diffs.
  - Return compare metadata and rename previousPath through the daemon API.
  - Update the frontend Files viewer rename display and empty states.
  - Add SQLite migration plus regenerated sqlc, OpenAPI, and frontend API schema artifacts.
  - Add backend/frontend tests covering the diff-base behavior and viewer updates.

  ## Testing

  - npm.cmd run sqlc
  - npm.cmd run api
  - go test ./internal/service/session -run "TestWorkspaceFilesIncludeCommittedBranchDiffAgainstRecordedBase|TestWorkspaceFilesReportRenamesAgainstRecordedBase|
    TestWorkspaceFilesIncludeWorkspaceProjectChildRepoDiffs|TestListWorkspaceFilesReturnsTrackedAndUntrackedStatus|TestGetWorkspaceFileReturnsContentAndDiff"

  - go test ./internal/session_manager -run TestSpawnRecordsDiffBaseForSingleRepoSessions
  - go test ./internal/storage/sqlite/store -run TestSessionPersistsDiffBaseMetadata -count=1
  - go test ./internal/httpd/controllers -run "TestSessionsAPI_ListWorkspaceFiles|TestSessionsAPI_GetWorkspaceFile"
  - npm.cmd run typecheck
  - npm.cmd run test -- SessionFilesView
  - git diff --check


backend/internal/session_manager/manager.go
```diff
@@ -610,6 +613,54 @@ func (m *Manager) createSessionWorkspace(ctx context.Context, project domain.Pro
...
+		if ref == "" {
+			return
+		}
+		if _, ok := seen[ref]; ok {
+			return
+		}
+		seen[ref] = struct{}{}
+		refs = append(refs, ref)
+	}
+	add(defaultBranch)
```
illegalcall: **Base ref candidate order is inverted relative to how the worktree was actually created.**

`gitworktree.baseRefCandidates` (`commands.go:131`) deliberately puts `origin/<default>` **before** `refs/heads/<default>`, and says so in a comment: *"The local head comes after origin/<defaultBranch> so remote-tracking still wins when present."* Its test pins the order: `["origin/feature/test", "origin/main", "refs/heads/main", "feature/test"]`.

So the worktree is created from `origin/main`, but this list resolves local `main` first, and `merge-base HEAD main` uses `refs/heads/main`. Whenever local `main` lags `origin/main` (the normal state, since nobody pulls `main` before every session) the recorded base is stale and the session reports upstream commits as its own work immediately on spawn, with no rebase involved.

Verified against a real clone with local `main` one commit behind:

```
local main : d96e0ec      origin/main : 2a18ab8

merge-base HEAD main         = d96e0ec  ->  A agent.txt  M shared.txt  A upstreamonly.txt
merge-base HEAD origin/main  = 2a18ab8  ->  A agent.txt
```

The agent only wrote `agent.txt`.

Suggested fix, move the local branch to last so this matches `gitworktree`:

```go
	if !strings.HasPrefix(defaultBranch, "origin/") && !strings.HasPrefix(defaultBranch, "refs/") {
		add("origin/" + defaultBranch)
		add("refs/remotes/origin/" + defaultBranch)
	}
	add(defaultBranch)
	return refs
```

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +146,284 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+		if ref == "" {
+			return
+		}
+		if _, ok := seen[ref]; ok {
+			return
+		}
+		seen[ref] = struct{}{}
+		refs = append(refs, ref)
+	}
+	add(defaultBranch)
```
illegalcall: **Same inverted ordering as `spawnDiffBaseRefCandidates`, so the read-time fallback repeats the bug.**

This list also resolves local `<defaultBranch>` before `origin/<defaultBranch>`, which contradicts `gitworktree.baseRefCandidates` (`commands.go:131`). Any session that reaches this fallback path with a stale local `main` gets a merge base behind the commit its worktree was actually created from.

Same fix: move `add(defaultBranch)` after the two `origin/` entries. Both lists need it, since one records the base and this one recovers it.

I applied both reorderings plus the `resolveWorkspaceCompare` change below and the full suite stayed green, including `internal/adapters/workspace/gitworktree` whose `TestBaseRefCandidates` pins the convention.

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +146,284 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+	if c.Mode == WorkspaceCompareBase && strings.TrimSpace(c.BaseSHA) != "" {
+		return c.BaseSHA
+	}
+	return "HEAD"
+}
+
+func (s *Service) resolveWorkspaceCompare(ctx context.Context, rec domain.SessionRecord, root, recordedSHA, recordedRef, defaultBranch string) (workspaceCompareTarget, error) {
+	recordedSHA = strings.TrimSpace(recordedSHA)
+	recordedRef = strings.TrimSpace(recordedRef)
+	if recordedSHA != "" && gitCommitExists(ctx, root, recordedSHA) {
```
illegalcall, line 242: **A recorded merge base goes stale as soon as the branch integrates newer base commits, and `gitCommitExists` cannot detect that.**

`cat-file -e` only proves the object exists. After a rebase onto newer `main`, or a merge of `main` into the session branch, the old merge base is still a reachable, still-valid commit, so this returns early and the live `gitMergeBase` path at line 238 is never reached. Diffing against it then attributes the base branch's own commits to the session.

Repro through this exact code path (rebase onto a `main` that moved):

```
compareMode="base" baseSHA=26f9f84c baseRef="main"
changed files reported: map[README.md:modified agent.go:added mainonly.go:added]
```

The agent only wrote `agent.go`. GitHub's PR diff lists exactly that one file.

Worth noting `merge-base --is-ancestor` does **not** fix this. I checked, it returns 0 in both the rebase and merge cases, because the old merge base is an ancestor of the new base. It has to be recomputed:

```go
	if recordedRef != "" && recordedRef != "HEAD" {
		for _, ref := range workspaceBaseRefCandidates(recordedRef) {
			if sha, ok := gitMergeBase(ctx, root, ref); ok {
				return workspaceCompareTarget{BaseSHA: sha, BaseRef: recordedRef, Mode: WorkspaceCompareBase}, nil
			}
		}
	}
	if recordedSHA != "" && gitCommitExists(ctx, root, recordedSHA) {
		return workspaceCompareTarget{BaseSHA: recordedSHA, BaseRef: recordedRef, Mode: WorkspaceCompareBase}, nil
	}
```

Keeping the stored SHA as the fallback still covers the case that justified storing it (shallow clone, `origin` not fetched). As written though, `DiffBaseSHA` is always populated for new single-repo sessions, so this early return means the live merge-base path at line 238 is effectively dead code.

This change plus the two candidate reorderings passes all five of the new tests in this PR unmodified.

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +146,284 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+		return workspaceCompareTarget{BaseSHA: recordedSHA, BaseRef: recordedRef, Mode: WorkspaceCompareBase}, nil
+	}
+	if s.store != nil {
+		prs, err := s.store.ListPRsBySession(ctx, rec.ID)
```
illegalcall: **This picks whichever PR was updated most recently, which is not a stable base.**

`ListPRsBySession` is `ORDER BY updated_at DESC` (`queries/pr.sql:88`), so `prs[0]` is just the last-touched PR. For a session with more than one PR (the multi-PR sessions from #2172) the comparison base can change on its own whenever PR update ordering changes, with no code or branch change. On a stacked session, picking a child PR's `BaseSHA` compares only against its parent and silently hides everything lower in the stack.

Prefer the root or default-target PR, or add an explicit `ORDER BY created_at ASC` for this use, rather than relying on `updated_at` ordering.

Minor, same block: this runs inside `resolveWorkspaceCompare`, which `listWorkspaceProjectFiles` calls once per child repo, so an N-repo workspace issues N identical queries plus N `cat-file` execs. Worth hoisting the PR lookup out of the per-repo loop.

backend/internal/service/session/workspace_files.go
```diff
@@ -366,32 +723,78 @@ func workspaceGitFiles(ctx context.Context, root string) ([]string, bool, error)
...
+	}
+	statuses := map[string]WorkspaceFileStatus{}
+	previous := map[string]string{}
+	for rel, status := range diffStatuses {
+		statuses[rel] = status
+	}
+	for rel, prev := range diffPrevious {
+		previous[rel] = prev
+	}
+	for rel, status := range statusStatuses {
```
illegalcall, line 900: **`git status` is relative to HEAD, `git diff <base>` is relative to base. Letting status win unconditionally makes the reported status describe the wrong baseline.**

Verified output from this code path, on a branch that committed two new files and then touched them in the worktree:

```
path=added_then_deleted.go    status=deleted    +0/-0
path=added_then_edited.go     status=modified   +3/-0
```

- `added_then_edited.go` was created in a branch commit then edited. It does not exist in base, so it is `added`, not `modified`. The `+3/-0` from numstat already says so, so the row contradicts its own counts.
- `added_then_deleted.go` was created in a branch commit then deleted uncommitted. Net effect versus base is nothing, but it renders as a phantom `deleted` row whose diff body comes back empty.

The underlying point is that `git diff <base>` with no `--cached` already covers committed, staged **and** unstaged changes, so `workspaceStatuses` is only genuinely needed for untracked entries. Restricting the merge to those lets diff-vs-base own tracked-file status and both cases above resolve:

```go
	for rel, status := range statusStatuses {
		if _, tracked := diffStatuses[rel]; tracked {
			continue // base-relative status already correct
		}
		statuses[rel] = status // untracked (??) only
	}
```

Renames followed by an edit are mislabeled the same way.

backend/internal/service/session/workspace_files.go
```diff
@@ -366,32 +723,78 @@ func workspaceGitFiles(ctx context.Context, root string) ([]string, bool, error)
...
 	}
+	for _, part := range parts {
+		addPath(part)
+	}
+	extra := make([]string, 0, len(extraPaths))
+	for rel := range extraPaths {
+		extra = append(extra, rel)
+	}
+	sort.Strings(extra)
+	for _, rel := range extra {
```
illegalcall, line 853: **Changed paths are added after the `ls-files` loop has already spent the cap, so in a large repo the actual changes are the thing that gets dropped.**

`addPath` hard-returns once `len(paths) >= maxWorkspaceFiles` (5000). The `ls-files` loop above runs first and its output is path-sorted, so in a repo with more than 5000 tracked plus untracked files the budget is consumed by alphabetically-early **unmodified** files, and every changed path sorting after them never makes it into the list. The Files tab then shows 5000 unchanged files and none of the session's work, which is the same user-visible symptom as #3203 itself, just gated on repo size.

Fix is to add the changed paths first, since they are the only ones the user opened the tab to see:

```go
	extra := make([]string, 0, len(extraPaths))
	for rel := range extraPaths {
		extra = append(extra, rel)
	}
	sort.Strings(extra)
	for _, rel := range extra {
		addPath(rel)
	}
	for _, part := range parts {
		addPath(part)
	}
```

This repo is 1677 files so it will not reproduce here, but AO points at arbitrary repos.

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +146,284 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+		exclude := []string(nil)
+		if prefix == "" {
+			exclude = childPrefixes
+		}
+		compare, err := s.resolveWorkspaceCompare(ctx, rec, row.WorktreePath, row.BaseSHA, defaultBranch, defaultBranch)
+		if err != nil {
+			return WorkspaceFiles{}, err
+		}
+		compares = append(compares, compare)
+		repoFiles, repoTruncated, err := workspaceFileSummaries(ctx, row.WorktreePath, prefix, exclude, compare)
```
illegalcall, line 406: **The `maxWorkspaceFiles` cap is per repo, not global, so the guard is bypassed in multi-repo workspaces.**

Each call gets its own 5000 budget and results are appended with no aggregate check, while `truncated` is only OR'd across rows. Two child repos with 4000 files each return 8000 entries with `Truncated == false`, so both the response-size guard and the truncation signal to the client are defeated.

Thread the remaining budget through the loop and set `truncated` once the aggregate limit is hit, rather than capping each repo independently.

Note this interacts with the ordering issue on line 749: one bug drops changed files below the cap, the other lets the aggregate exceed it. A single budget-aware pass that prioritizes changed paths fixes both.

backend/internal/service/session/workspace_files.go
```diff
@@ -178,36 +447,124 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+
+func aggregateWorkspaceCompare(compares []workspaceCompareTarget) workspaceCompareTarget {
+	if len(compares) == 0 {
+		return workspaceCompareTarget{Mode: WorkspaceCompareHeadFallback}
+	}
+	mode := WorkspaceCompareBase
+	baseSHA := compares[0].BaseSHA
+	baseRef := compares[0].BaseRef
+	for _, compare := range compares {
+		if compare.Mode != WorkspaceCompareBase {
```
illegalcall, line 628: **One unresolvable child repo flips the whole response to `head_fallback`, contradicting the data that was already computed.**

If any single repo failed to resolve a base, this returns `head_fallback` with an empty SHA. But every repo's files were already diffed against **its own** resolved base at line 327, so the response metadata now describes something different from the file list it ships with. On the client that surfaces as `emptyFilesMessage` claiming "No changes against HEAD." for a set that was in fact compared against base.

Consider reporting per-repo compare state, or at least keeping `mode: base` when at least one repo resolved and signalling the partial failure separately.

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +146,284 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+		}
+		prefix, ok := prefixes[row.RepoName]
+		if !ok {
+			continue
+		}
+		exclude := []string(nil)
+		if prefix == "" {
+			exclude = childPrefixes
+		}
+		compare, err := s.resolveWorkspaceCompare(ctx, rec, row.WorktreePath, row.BaseSHA, defaultBranch, defaultBranch)
```
illegalcall: **`defaultBranch` is being passed in the `recordedRef` parameter slot.**

The signature is `resolveWorkspaceCompare(ctx, rec, root, recordedSHA, recordedRef, defaultBranch)`, so this call sets `recordedRef = defaultBranch`. Child repos then report `CompareBaseRef` as the default branch name while `CompareBaseSHA` is `row.BaseSHA`, the worktree's creation SHA. Those are two different commits, so the API advertises provenance that does not match the SHA it returns.

Line 362 in `getWorkspaceProjectFile` does the same thing.

Either pass `""` for `recordedRef` and let the resolver label it, or record a real ref alongside `BaseSHA` on the worktree row.

illegalcall (commented):
Reviewed this thoroughly against `main`. The design is right, record a base at spawn, resolve it at read time, plumb compare metadata to the client, and the DB / API / sqlc / OpenAPI layers are clean and complete. `go build ./...` and the four relevant test packages pass locally, and the five new tests are real (actual git repos, real assertions) rather than mock theatre.

The findings below are all either reproduced by execution or follow from unambiguous logic. I have left them as inline comments on the specific lines with suggested fixes.

The one I would most want addressed before merge is the base ref candidate ordering (`manager.go:646`, `workspace_files.go:264`). It contradicts `gitworktree.baseRefCandidates`, which intentionally prefers `origin/<default>` over the local head. The result is that a session whose local `main` is not freshly pulled, which is the normal state, misreports upstream commits as its own work from the first render, with no rebase or merge needed. That is the same class of silently-wrong diff this PR set out to fix.

Grouped by theme:

**Base resolution (three separate triggers, one fix family)**
- `manager.go:646` and `workspace_files.go:264`, candidate order inverted versus `gitworktree`
- `workspace_files.go:222`, recorded SHA goes stale after rebase or merge and `cat-file -e` cannot detect it
- `workspace_files.go:226`, PR fallback picks the most recently updated PR, which is not stable

**Status and file listing**
- `workspace_files.go:793`, `git status` (vs HEAD) overriding `git diff` (vs base) mislabels rows
- `workspace_files.go:749`, changed paths appended after the 5000 cap is spent
- `workspace_files.go:327`, the cap is per repo, so multi-repo workspaces bypass it

**Metadata accuracy**
- `workspace_files.go:535`, one unresolved repo poisons the whole response mode
- `workspace_files.go:322` and `:362`, `defaultBranch` passed in the `recordedRef` slot

I verified the base-resolution fix rather than just proposing it. Reordering both candidate lists plus recomputing the merge base live in `resolveWorkspaceCompare` (roughly 12 lines across two files) makes my repros pass and keeps the whole suite green, including `internal/adapters/workspace/gitworktree` whose `TestBaseRefCandidates` pins the ordering convention:

```
ok  internal/service/session                 6.371s
ok  internal/session_manager                 5.938s
ok  internal/httpd/controllers               (cached)
ok  internal/adapters/workspace/gitworktree  13.835s
```

**Test gaps worth closing**, since the base-resolution bugs shipped CI-green:
1. A session whose local default branch lags `origin/<default>` at spawn
2. A branch rebased or merged onto a newer base after spawn
3. `head_fallback` mode, empty `DiffBaseSHA` with no resolvable ref
4. The PR-base fallback branch at lines 226 to 236, currently untested
5. Committed **deletions** against base, the description claims these are handled but nothing asserts it

**Checked and clear**, flagging these explicitly since they were the risky areas:
- `-z` parsing is correct. I dumped real `git diff --name-status --find-renames -z` and `--numstat -z` through `od -c` and traced both parsers. `parseGitStatusRecord` handles the `R080\0old\0new\0` triple, and `workspaceNumstat` correctly handles binary (`-\t-\tpath`), renames (`1\t1\t\0old\0new`) and the binary-rename combination. Careful code.
- No shell injection. `gitWorkspaceOutput` uses an argument array, never a shell, and all pathspecs sit behind `--`.
- Path confinement is intact, `confinedWorkspaceFile` resolves symlinks segment by segment and re-checks containment at each step.
- Migration 0037 is fine. Two ALTERs in one `StatementBegin` block matches existing convention (0004 has 72), `DROP COLUMN` in the down is used by five existing migrations, modernc.org/sqlite v1.51.0 supports it. No backfill needed because empty `diff_base_sha` falls through to the runtime fallback, which is the right call.

**Adjacent, not this PR's fault:** `syntheticAddedFileDiff` is byte-identical to `main`, so #2923 (extra trailing blank line and inflated `@@` count from `SplitAfter` leaving a trailing empty element) is untouched here. Since #3190 closed without merging it is still open, and it is visible in this area, `textLineCount` computes the right count two functions away so the summary row and the diff header currently disagree.

Nothing here is architectural. It is resolution order and baseline mixing, and the fixes are small.

backend/internal/service/session/workspace_files.go
```diff
@@ -129,32 +147,362 @@ func (s *Service) GetWorkspaceFile(ctx context.Context, id domain.SessionID, raw
...
+		}
+		prefix, ok := prefixes[row.RepoName]
+		if !ok {
+			continue
+		}
+		exclude := []string(nil)
+		if prefix == "" {
+			exclude = childPrefixes
+		}
+		compare := resolveWorkspaceCompare(ctx, row.WorktreePath, row.BaseSHA, "", defaultBranch, nil)
```
neversettle17-101: Workspace-project child repos still use only the spawn-time `row.BaseSHA` here, with no recorded ref. That leaves them exposed to the same stale-base issue this PR fixes for single-repo sessions: after a child branch rebases onto or merges a newer base, the old base SHA still exists, so `resolveWorkspaceCompare` keeps diffing against it and can attribute base-branch commits to the session.

Can we give child worktree rows equivalent base-ref/base-branch metadata, or recompute the child repo merge-base from its configured base before falling back to `row.BaseSHA`? The detail path below has the same issue.

Vaibhaav-Tiwari:
@neversettle17-101 pls check now

Hardik180704 (changes requested):
The base-aware diff direction is solid, but this branch needs one update before merge. It adds `0038_add_session_diff_base.sql`, while current `main` already contains migration versions `0038` and `0039`. Merging this as-is would create a duplicate Goose version and fail `TestMigrationVersionsAreUnique`.

Please rebase onto current `main`, renumber this migration to the next available version, regenerate the sqlc/API artifacts as needed, and rerun CI.

Hardik180704 (approved):
Rechecked the latest head after the requested updates. The migration is now correctly numbered as 0040, the base-resolution and workspace-diff review threads have been addressed, generated artifacts are in sync, and all CI checks are green. Looks good to merge from my side.

Issue #3038, opened 2026-07-23
Remove the '[AO] Worker <id> has gone idle' nudge messages sent to the orchestrator


harshitsinghbhandari:
The daemon auto-sends messages like this to the orchestrator session whenever a worker goes idle:

> [AO] Worker agent-orchestrator-69 ("release-arch-spec") has gone idle and may be done. Inspect it with `ao session get ...`, then report its status and any PR to the human. If it needs more work, redirect it with `ao send`.

These are unnecessary and should be removed:

- They re-fire repeatedly for the same worker (the same session gets nudged multiple times, including ones already inspected and reported).
- Idle does not mean done; read-only/research workers idle immediately after reporting, so the nudge adds no signal.
- Each nudge burns an orchestrator turn and context on boilerplate the human can already see in the sidebar (idle state is visible there).

Just drop the nudge. The orchestrator inspects workers on demand; the human sees idle state in the UI.

AgentWrapper:
## New Report — live orchestrator-side evidence: the idle nudge is not just noisy, its delivery logic is buggy in three distinct ways
**Reported by:** maintainer + orchestrator session logs | **Date:** 2026-07-28 | **Daemon:** app-owned :3001, running post-#2836 build (worker_idle_outbox / migration 25 present; note the checkout at `dec77c61b` predates it, so line refs omitted)

Observed from a single orchestrator session's inbox over 24h:

1. **Delivery to the orchestrator AFTER the worker was terminated.** Session `agent-orchestrator-52` was killed at ~its 13th minute; the "[AO] Worker agent-orchestrator-52 has gone idle" nudge arrived *after* the kill (daemon confirmed `terminated: true` on inspection). The outbox appears to enqueue on the idle transition and deliver later with no terminal-state check at delivery time.

2. **Duplicate deliveries with no state change in between.** Sessions `-63` and `-64` (completed workers, status `review_pending`) each produced two identical nudges minutes apart; session `-40` produced at least four over ~3h on 2026-07-27. Either the outbox lacks ack/dedup, or every observation re-queues.

3. **Re-arming on human interaction with long-done workers.** Session `-32` has been finished and `review_pending` for 8 days; the maintainer typed one message into it ('preview localhost:5173'), the activity flapped active→idle, and the orchestrator got nudged about it twice — as if the week-old worker had just finished. A `review_pending`/`done` session going idle after a human poke is not a signal the orchestrator needs.

[...]

PR #3040, opened 2026-07-23
refactor(lifecycle): remove the worker-idle orchestrator nudge

harshitsinghbhandari:
Closes #3038.

## What

Drops the `[AO] Worker <id> has gone idle and may be done. Inspect it with ...` message the daemon auto-sent to the orchestrator session whenever a worker went idle.

Per the issue, the nudge added no signal and cost turns:

- It re-fired repeatedly for the same worker (each idle re-nudged, including ones already inspected).
- Idle does not mean done: read-only/research workers idle immediately after reporting.
- The human already sees idle state in the sidebar, and the orchestrator inspects workers on demand.

## How

The nudge was the sole consumer of the `worker_idle` outbox, so the whole pipeline that existed only to carry it is removed:

- `domain.WorkerIdleEvent`, the store methods (`RecordWorkerIdle`, `ListPendingWorkerIdleEvents*`, `MarkWorkerIdleEventDelivered`), and the sqlc query/model.
- Lifecycle dispatch plumbing: `DispatchPendingWorkerIdleEvents`, `DispatchAllPendingWorkerIdleEvents`, `workerIdleNudgeMessage`, `crossedToIdle`, `orchestratorDispatchTrigger`, `safeToDeliver`, `liveOrchestrator`, and the per-project dispatch locks.
- Daemon wiring: the boot-time redelivery goroutine and the 2-minute recovery sweep.
- Migration `0037` drops the now-unused `worker_idle_events` table.

Idle **detection** on the session row is untouched (the UI still shows idle state); only the orchestrator-directed message goes away. Net: +35 / -1330 lines.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test ./internal/lifecycle/ ./internal/adapters/agent/fake/ ./internal/storage/sqlite/...` pass (the sqlite suite applies migration 0037).
- Pre-existing, unrelated failures confirmed identical on pristine `upstream/main`: macOS `/private/var` symlink path tests (`workspace/scratch`, `daemon` wiring) and the `internal/cli` spawn 404 integration tests.

## Rebase onto main (2026-07-28)

Rebased onto `upstream/main` (a10c98c80) to clear the conflict. Resolutions:

- `daemon/lifecycle_wiring.go`: kept mains new `runtimeReaper` / `activityDone` fields, the activity poller wiring, and `ReconcileRuntime`; dropped only `sweepDone` and the `startWorkerIdleSweep` start/drain.
[...]

PR #3215, opened 2026-07-28
fix(lifecycle): stop duplicate and stale worker-idle nudges to the orchestrator (#3038)

Pulkit7070:
## What

Fixes the noise in the "[AO] Worker … has gone idle" nudges reported in #3038. Instead of removing the worker-idle outbox, this keeps the feature and closes the three bug classes with targeted guards.

## Why (live evidence)

Pulled from a running daemon's `worker_idle_events` table: **88 nudge events, one worker (`zero_to_one-23`) generated 25 of them in a single evening** while it was only ever `idle` (not terminated, not done). Root cause: the outbox enqueued on *every* worker `active→idle`, with no delivery-time recheck.

Three distinct symptoms (matching the issue's follow-up report):
1. **Delivery after kill** — event enqueued on idle, delivered later with no terminal check.
2. **Duplicates** — every re-idle re-enqueued (auto-mode turn boundaries, retries).
3. **Re-arm on human/orchestrator poke** — a done/`review_pending` worker flapping `active→idle` after a message re-nudged as if freshly finished.

## Changes (no schema change)

- **Report-once-per-episode** (`ApplyActivitySignal`): an in-memory `reportedIdle` set enqueues at most one nudge per attention episode. Re-idles are suppressed; a worker entering a real `needs_input`/`blocked` state reopens reporting; spawn/restore and termination reset it. A poke-induced idle never passes through `needs_input`, so it is naturally suppressed — this collapses both the duplicates (#2) and the poke re-arms (#3). In-memory by design, mirroring the existing `flights` map: a daemon restart at worst allows one extra nudge, never a lost completion.
- **Delivery-time recheck** (`DispatchPendingWorkerIdleEvents`): re-reads the worker right before sending and drains the event (marks delivered without sending) when the worker is **terminated** or owns an **open PR** (work already in review and visible to the orchestrator). Fixes #1 and stops nudging review_pending workers.

Net effect: workers still show idle in the sidebar; the orchestrator just stops being auto-poked repeatedly for the same completion.

## Tests

`go test ./internal/lifecycle/` (incl. `-race`) green. Added coverage:
- `TestActivity_WorkerReportedOncePerEpisode` — re-idle produces no second nudge.
- `TestActivity_WorkerReReportsAfterNeedsInput` — a genuine needs-input episode re-arms reporting.
- `TestDispatch_DrainsTerminatedWorkerWithoutNudge` — terminated worker's event is drained, not sent.
- `TestDispatch_OpenPRDrainsMergedPRStillNudges` — open-PR worker drained; merged-PR worker still nudges.

Pre-existing failures in `internal/cli`, `adapters/workspace/scratch`, `adapters/agent/kilocode`, and `daemon` are environment-related (macOS `/private/var` path canonicalization, spawn tests needing a live daemon) and reproduce on a clean baseline; none touch `internal/lifecycle`.

## Alternative considered

Full removal of the outbox (what the issue literally proposes). This PR is the "keep it, make it correct" option. If maintainers prefer removal, that is a larger, cleaner delete (trigger, dispatch, store, table via a new drop-migration, daemon sweep) and can be done instead.

Refs #3038

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -304,6 +304,17 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID, 
 		m.mu.Unlock()
 		return nil
 	}
+	// A new instruction submitted to a worker starts a new completion
+	// generation: its next active->idle is a genuine completion of newly
+	// assigned work, so reopen reporting. Re-arming keys on new work arriving
+	// (user-prompt-submit) — the issue's "completion-generation" — rather than
+	// on incidental needs-input transitions, which are unrelated to whether new
+	// work was assigned (a mid-run permission dialog must not decide it). Placed
+	// before the staleness guards so it lands regardless of same-state early
+	// returns; clearing at worst allows one extra nudge, never a lost one.
+	if rec.Kind == domain.KindWorker && s.Event == "user-prompt-submit" {
+		delete(m.reportedIdle, id)
+	}
 	if s.LaunchID != "" && s.LaunchID != rec.Metadata.RuntimeLaunchID {
 		m.mu.Unlock()
 		return nil
@@ -381,14 +392,11 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID, 
 	// A worker's active->idle transition creates a durable worker_idle event in
 	// the same write as the activity change, so a crash can't persist the idle
 	// state while losing the pending delivery. It is enqueued at most once per
-	// attention episode (see reportedIdle): re-idles and poke-induced idles are
-	// suppressed, and a worker entering a real needs-input state reopens
-	// reporting for its next completion.
+	// completion generation (see reportedIdle): re-idles and poke-induced idles
+	// within the same generation are suppressed, and a new instruction
+	// (user-prompt-submit, handled above) reopens reporting for the next one.
 	var idleEvent *domain.WorkerIdleEvent
-	switch {
-	case next.Kind == domain.KindWorker && !prevState.NeedsInput() && next.Activity.State.NeedsInput():
-		delete(m.reportedIdle, id)
-	case crossedToIdle(prevState, next):
+	if crossedToIdle(prevState, next) {
 		if _, reported := m.reportedIdle[id]; !reported {
 			idleEvent = &domain.WorkerIdleEvent{
 				ID:           uuid.NewString(),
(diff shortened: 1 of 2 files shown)


harshitsinghbhandari:
Actually we would prefer to not have this feature because I usually don't kill worker sessions because I need them frequently. Happy to discuss is it is needed or not. I wouldn't really prefer this. Linking my PR: #3040

backend/internal/lifecycle/manager.go
```diff
@@ -366,15 +380,24 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID,
...
-			ID:           uuid.NewString(),
-			ProjectID:    next.ProjectID,
-			WorkerID:     next.ID,
-			TransitionAt: next.Activity.LastActivityAt,
-			CreatedAt:    now,
+	switch {
+	case next.Kind == domain.KindWorker && !prevState.NeedsInput() && next.Activity.State.NeedsInput():
+		delete(m.reportedIdle, id)
+	case crossedToIdle(prevState, next):
+		if _, reported := m.reportedIdle[id]; !reported {
```
codebanditssss, line 400: This is the report-once gate. Because `reportedIdle` only clears on needs_input / spawn / terminate, a worker that gets new work via `ao send` and completes it never nudges again.

Probe run on this branch vs `origin/main` (idle -> active -> idle 20 min later, no needs_input in between): main = 2 nudges, this branch = 1. Insert a single `blocked` transition mid-run and this branch goes back to 2, so the re-arm depends on whether a permission dialog happened rather than on whether new work arrived.

`s.Event == "user-prompt-submit"` is already available in this function and is a closer proxy for the issue's "completion-generation".

backend/internal/lifecycle/manager.go
```diff
@@ -494,6 +517,20 @@ func (m *Manager) DispatchPendingWorkerIdleEvents(ctx context.Context, project d
...
 	ev := events[0]
+	// Re-check the worker at delivery time, not just at enqueue. A completion
+	// enqueued minutes ago is stale if the worker has since been killed (no one
+	// to inspect) or opened a PR (its work is already in review and visible to
+	// the orchestrator). In both cases the nudge is noise, so drain the event
+	// without sending rather than leave it pending to retry forever.
+	if worker, ok, err := m.store.GetSession(ctx, ev.WorkerID); err != nil {
+		slog.Default().Error("lifecycle: read worker for idle delivery", "worker", ev.WorkerID, "err", err)
+		return
+	} else if !ok || worker.IsTerminated || m.workerHasOpenPR(ctx, ev.WorkerID) {
```
codebanditssss: The `IsTerminated` half of this is right and well tested. The `workerHasOpenPR` half drops the common case: a worker's last act is `gh pr create`, so the PR row exists by dispatch time and the completion is drained. Probe: main nudges, this branch sends nothing for a worker that finished by opening a PR - and the nudge text is "report its status and any PR to the human".

`!Merged && !Closed` also covers draft, ci_failed and changes_requested. If the intent is "already in review", matching the derived status (review_pending / approved / mergeable) would be narrower and closer to what the issue suggested.

backend/internal/lifecycle/manager.go
```diff
@@ -494,6 +517,20 @@ func (m *Manager) DispatchPendingWorkerIdleEvents(ctx context.Context, project d
...
+		if err := m.store.MarkWorkerIdleEventDelivered(ctx, ev.ID, m.clock()); err != nil {
+			slog.Default().Error("lifecycle: drain stale worker idle", "event", ev.ID, "err", err)
+		}
+		return
```
codebanditssss, line 558: Minor: a drain returns without trying the next pending event, and unlike a delivery it produces no orchestrator transition to re-trigger dispatch. So a stale backlog only advances on the 2-minute `workerIdleSweepInterval` sweep, one per cycle (~3h for the 88 events in the description). Looping to the next pending event after a drain would clear it in one pass.

codebanditssss (changes requested):
Nice, careful patch: the locking is right (`reportedIdle` is only touched under `m.mu` in `ApplyActivitySignal`/`mutate`/`MarkSpawned`, and dispatch never touches it), the two switch cases can't shadow each other since `crossedToIdle` requires `next.State == idle`, and there's no schema change. I ran `go test ./internal/lifecycle/ -race` and `go vet` on 52ec83d - both green. The terminated-worker drain at delivery time is exactly what the issue's follow-up asked for.

I did write a couple of probe tests and ran them against both this branch and `origin/main`, and two behaviours change in ways I don't think are intended.

### 1. A follow-up assignment's completion is never reported

Report-once-per-episode only reopens on needs_input / spawn / terminate, so the ordinary orchestrator loop loses its signal:

worker idles -> nudge -> orchestrator `ao send`s more work -> worker works 20 minutes -> worker idles -> **no nudge**.

Probe (same test on both): `main` = 2 nudges, this branch = 1. That second idle is a real completion of newly assigned work, not a re-idle and not a poke.

The related problem is that the re-arm key is unrelated to whether new work arrived. If the same worker happens to hit one permission dialog mid-run (`blocked`), the second completion *does* nudge (probe = 2 on this branch). So whether the orchestrator hears about a finished task depends on whether a permission prompt fired, which is harness-dependent: an auto-approved worker, or an adapter that never emits waiting_input/blocked, gets exactly one nudge for its whole lifetime.

The issue's own suggestion was "dedup by (session, completion-generation)", where the generation advances when new work goes in. `ApplyActivitySignal` already sees `s.Event == "user-prompt-submit"`, which is a much closer proxy for "new instruction arrived" than needs_input is. That does re-arm on a human poke, which is symptom #3, so it needs a call from the maintainer - but keying on needs_input trades a real completion signal for that.

### 2. The open-PR drain swallows the completions that mattered most

The normal AO worker ending is `gh pr create` then idle, so the PR row exists by the time the event is dispatched. Probe: `main` nudges, this branch sends nothing at all for a worker that finished by opening a PR.

The nudge text is literally "report its status **and any PR** to the human", so this guard drops precisely the events whose job was to get the PR reported to the human. `!Merged && !Closed` is also wider than the `review_pending` the issue suggested: it covers draft, ci_failed and changes_requested PRs too. Some of that is handled by the reactions pipeline nudging the worker directly, but the orchestrator-tells-the-human path isn't.

If the intent is "already in review", matching the derived status (review_pending / approved / mergeable) rather than any non-closed PR would be closer, and would leave draft and ci_failed alone.

### 3. Minor: a drained event doesn't advance the queue

A drain `return`s without trying the next pending event, and unlike a delivery it causes no orchestrator state change to re-trigger dispatch. So a stale backlog only advances on the 2-minute `workerIdleSweepInterval` sweep, one event per cycle - the 88-event backlog in your description would take ~3 hours to clear. Continuing to the next pending event after a drain instead of returning would fix it.

### On direction

The issue title asks for removal and the maintainer's follow-up says "if these are kept at all". You flag this in "Alternative considered", which is the right call - worth getting a maintainer decision on keep-vs-remove before reworking the episode key, since a removal makes all of the above moot.

Happy to re-review once the episode key and the open-PR predicate are settled. The delivery-time terminated check and its test can stand as-is either way.

backend/internal/lifecycle/manager.go
Pulkit7070: Fixed in e1e6d723c. Re-arm now keys on the worker's user-prompt-submit (the completion-generation: a new instruction arrived), replacing the needs_input keying. A mid-run blocked no longer re-arms; a new prompt does. Added TestActivity_WorkerReReportsAfterNewPrompt and TestActivity_WorkerBlockedMidRunDoesNotReReport.

backend/internal/lifecycle/manager.go
Pulkit7070: Fixed in e1e6d723c. Dropped the open-PR half of the drain entirely; only terminated workers are drained now, so a worker that finished by opening a PR is still reported (that is exactly what the nudge is for). Removed workerHasOpenPR. Added TestDispatch_OpenPRWorkerStillNudges.

backend/internal/lifecycle/manager.go
Pulkit7070: Fixed in e1e6d723c. Dispatch now continues to the next pending event after a drain instead of returning, so a stale backlog clears in one pass; still at most one live delivery per pass. Added TestDispatch_DrainAdvancesToNextPendingEvent.

Pulkit7070:
> Actually we would prefer to not have this feature because I usually don't kill worker sessions because I need them frequently. Happy to discuss is it is needed or not. I wouldn't really prefer this. Linking my PR: #3040

Makes sense, I'm on board with removing it. The whole outbox existed only to carry this nudge, so    
  dropping it (your #3040) is cleaner than keeping a feature you don't want and guarding around the    
  edge cases. I'll close #3215 in favor of #3040. Happy to review #3040 if that helps.

Pulkit7070:
Makes sense, I'm on board with removing it. The whole outbox existed only to carry this nudge, so dropping it (#3040) is cleaner than keeping a feature you don't want and guarding around the edge cases. Closing this in favor of #3040. Happy to review #3040 if that helps.

AgentWrapper:
@Pulkit7070 why did we close this PR?

PR #3217, opened 2026-07-28
fix(telemetry): validate command-shaped values at PostHog sink

### LINKED ISSUE #3133: Telemetry: validate command-shaped values at the PostHog sink (defense in depth)
Follow-up carved out of #2816, which is being closed as superseded.

## Context

#2829 (merged as `a6d90dfe3`) fixed the reported leak in #2813 at the source: `usageErrorCommand` in `backend/internal/cli/root.go` now walks the live Cobra command tree and stops at the first token that is not a registered subcommand, so raw positional args (URLs, file paths, free text) can no longer reach `command` or `command_path`.

That is the client-side half. #2816 also proposed a **sink-side** defense in depth that is still not on main.

## The gap

The PostHog allowlist gates telemetry *keys*, not *values*. If any future code path writes a command-shaped property, nothing at the sink validates what is actually in it. #2829 fixed today's producer; it did not add a backstop against the next one.

## Proposed change

In the PostHog sink (`sanitizeRemoteValue`), for command-shaped keys (`command`, `command_path`):

- Validate the value against the expected command shape (registered subcommand names joined by spaces).
- If it does not match, replace it with a `sha256:<16 hex>` digest rather than dropping it, so volume and cardinality stay analysable without shipping the raw string.

## Notes

- #2816 has a working implementation of this half if it is useful as a starting point, though it will need rebasing onto post-#2829 main.
- The reviewer on #2816 flagged that there was no end-to-end route test for the sink defense. Worth adding one here.

Hardik180704:
## Summary

- validate `command` and `command_path` values at the PostHog sink
- replace invalid command-shaped values with stable `sha256:<16 hex>` tokens
- keep valid command names and paths unchanged
- cover the sanitizer and the full telemetry route-to-PostHog payload

## Why

The remote allowlist restricted telemetry keys, but any string stored under a command-shaped key could still be forwarded. This adds a sink-side backstop so URLs, local paths, prose, and other unexpected values cannot leave the machine as raw text.

## Tests

- `npm run lint`
- `cd backend && go test -race ./internal/adapters/telemetry ./internal/httpd`

Closes #3133

### DIFF
--- a/backend/internal/adapters/telemetry/posthog.go
+++ b/backend/internal/adapters/telemetry/posthog.go
@@ -25,6 +25,54 @@ import (
 const postHogBufferSize = 128
 const maxCommandShapeLength = 48
 
+var remoteCommandTokens = map[string]struct{}{
+	"add":              {},
+	"agent":            {},
+	"agent-process":    {},
+	"cancel":           {},
+	"claim-pr":         {},
+	"cleanup":          {},
+	"clear":            {},
+	"completion":       {},
+	"daemon":           {},
+	"delete":           {},
+	"dev":              {},
+	"doctor":           {},
+	"execute":          {},
+	"get":              {},
+	"help":             {},
+	"hooks":            {},
+	"import":           {},
+	"import-projects":  {},
+	"kill":             {},
+	"launch":           {},
+	"list":             {},
+	"ls":               {},
+	"merge":            {},
+	"orchestrator":     {},
+	"preview":          {},
+	"pr":               {},
+	"project":          {},
+	"remove":           {},
+	"rename":           {},
+	"resolve-comments": {},
+	"restart":          {},
+	"restore":          {},
+	"review":           {},
+	"rm":               {},
+	"send":             {},
+	"session":          {},
+	"set-config":       {},
+	"spawn":            {},
+	"start":            {},
+	"status":           {},
+	"stop":             {},
+	"submit":           {},
+	"supervise":        {},
+	"trigger":          {},
+	"version":          {},
+}
+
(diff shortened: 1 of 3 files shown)


JAYATIAHUJA (changes requested):
Requesting one change: the sink-side command validation is still too permissive.

`isCommandShaped` currently accepts any short token sequence containing letters/digits/hyphen/underscore, so plain prose or private labels like `customer acme launch`, `secret_project`, or `ao private customer` would still pass through raw as `command` / `command_path`.

Can we tighten this by making `command_path` require the expected AO command form, e.g. start with `ao` and then only known/allowed command tokens, and make `command` a single expected command token? Please add tests for plain alphabetic prose/private labels to ensure they hash instead of exporting raw.

The rest looks close: focused backend telemetry/httpd tests pass locally, and CI is green.

Hardik180704:
@JAYATIAHUJA addressed this in 1c7a3649a. The sink now uses an explicit AO command-token allowlist: `command` must be a single allowed token, and `command_path` must start with `ao` and contain only allowed tokens. I also added regression cases for `customer acme launch`, `secret_project`, and `ao private customer`; all three are hashed instead of exported raw. The full backend suite, focused race tests, and golangci-lint are green locally. Thanks for catching this!

JAYATIAHUJA (approved):
LGTM. Fix commit 1c7a364 addresses the earlier feedback: command is now a single allowlisted token, command_path must start with ao and use only allowed tokens, and prose/private-label values (customer acme launch, secret_project, ao private customer) are covered by regression tests and hash to sha256:<16 hex> instead of exporting raw. CI is green.

PR #3222, opened 2026-07-28
fix: improve daemon failure and light terminal contrast

JAYATIAHUJA:
## What
- Centers the daemon startup failure surface as a modal alert dialog with dismiss, retry, retry-error feedback, and details/copy actions.
- Adds light terminal ANSI tokens and makes xterm's light palette read those tokens directly.
- Normalizes light-mode terminal output so dark TUI panels become tokenized light panels and overly bright prompt/status foregrounds are darkened without corrupting split UTF-8 or non-SGR CSI output.

## Why
- The daemon failure toast could appear awkwardly over settings/content and had no direct recovery action.
- In light mode, Codex/agent terminal output could render black panels or very pale yellow/green prompt text that was difficult to read.

## How
- Uses existing design tokens such as `--color-scrim` and new `--color-term-light-*` terminal palette tokens instead of embedding design colors in renderer TypeScript.
- Keeps terminal color rewriting scoped to light mode and preserves streaming output with a per-terminal `TextDecoder` plus incomplete-SGR buffering.
- Adds modal alert dialog semantics, focus containment, Escape dismissal, and handled retry failures.

before :
Screenshot 2026-07-28 161938
Screenshot 2026-07-28 141648

after : 
Screenshot 2026-07-28 170844

## Testing
- `npm --prefix frontend run test -- src/renderer/components/DaemonFailureBanner.test.tsx src/renderer/components/XtermTerminal.test.tsx src/renderer/lib/terminal-themes.test.ts` passed: 66 tests.
- `npm exec vite -- --config vite.renderer.config.ts build` passed from `frontend/` after sandbox escalation for native Vite/Tailwind dependencies.
- `npm --prefix frontend run test` was run and still has unrelated existing/local-environment failures: 4 `src/main/supervisor-link.test.ts` Windows socket `EACCES` failures and 1 `src/renderer/test/shell-new-session-shortcut.test.tsx` collapsed-sidebar shortcut assertion failure in untouched code; 1171 tests passed.


Hardik180704 (changes requested):
The focused tests and typecheck pass, but the streaming normalizer can corrupt output when the theme changes while it holds a partial UTF-8 character or SGR sequence. I reproduced this by splitting λ across two chunks, sending the first in light mode, switching to dark, and sending the second: the terminal receives �. This happens because the dark-mode branch returns raw bytes while the decoder still owns the prefix. Please preserve or flush buffered state across theme transitions and add a light-to-dark regression test. Also please run formatting around the retryDaemon declaration in DaemonFailureBanner.tsx.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -92,6 +92,137 @@ const SUPPRESS_NATIVE_PASTE_MS = 100;
...
+}
+
+function createTerminalOutputNormalizer(theme: () => Theme): (data: Uint8Array) => Uint8Array | string {
+	const decoder = new TextDecoder();
+	let pendingText = "";
+	return (data) => {
+		if (theme() !== "light") return data;
+		const decoded = pendingText + decoder.decode(data, { stream: true });
+		const { complete, pending } = splitIncompleteSgrTail(decoded);
+		pendingText = pending;
```
Pulkit7070, line 179: Minor edge: the normalizer keeps pendingText across calls to hold an incomplete SGR escape split across chunks, but the early return on the non-light path returns raw data without flushing pendingText. If the theme switches from light to dark at exactly the moment an SGR sequence is split across a chunk boundary, the buffered tail (\x1b[ plus digits) is dropped and the next dark-path write starts mid-sequence, which can briefly mis-render one escape. Cosmetic and rare, but you could flush pendingText once when leaving light mode, or clear it, to be tidy.

frontend/src/renderer/components/DaemonFailureBanner.tsx
```diff
@@ -1,27 +1,81 @@
...
-	if (!status.code || status.state === "ready") return null;
-	return <DaemonFailureContent status={status} />;
+	const [dismissed, setDismissed] = useState(false);
+	const failureKey = `${status.state}:${status.code ?? ""}:${status.message ?? ""}:${status.details ?? ""}`;
+
+	useEffect(() => {
+		setDismissed(false);
+	}, [failureKey]);
+
+	if (!status.code || status.state === "ready" || dismissed) return null;
```
Pulkit7070: Confirm the intended recovery path: dismissing hides the dialog until failureKey changes, but if the daemon stays down with the same failure signature, the retry button is gone until the failure changes. That is fine only if the user has another way back to retry or to see the daemon-down state (e.g. the sidebar daemon status). If dismiss is the only affordance and the failure is persistent, a dismissed user is left with no visible indication or retry. Please confirm there is a secondary entry point, otherwise consider keeping a small persistent indicator after dismiss.

Pulkit7070 (commented):
Reviewed line by line. This is careful work; the two hard parts hold up.

Light-terminal ANSI normalization (XtermTerminal.tsx): the color math is correct. remapLightThemeForeground uses perceptual luminance (0.299/0.587/0.114) and scales channels uniformly to a max luminance, preserving hue; xterm256ColorToRgb implements the standard 6x6x6 cube (steps 0/95/135/175/215/255) and the 232-255 grayscale ramp correctly; dark backgrounds (all channels <= 56) map to the light panel token. The SGR rewriter is the risky part and it is safe: I traced 38;5;N and 38;2;r;g;b, truncated params (38;5 or 38;2;100), and mixed sequences like 1;38;5;82;4, and the isFinite/isInteger guards plus fall-through mean malformed input is passed through unchanged rather than corrupted or throwing. The streaming decoder with splitIncompleteSgrTail correctly buffers an incomplete escape across chunk boundaries.

Daemon failure modal (DaemonFailureBanner.tsx): the alertdialog is sound. The hand-rolled focus trap handles Tab and Shift+Tab wrap in both directions and the initial container-focused case, Escape dismisses, and dismissed resets on failureKey so a new failure re-surfaces.

Two minor, non-blocking notes inline. No defects found.

frontend/src/renderer/components/DaemonFailureBanner.tsx
JAYATIAHUJA: Confirmed. After the merge resolution, the modal keeps the dismiss behavior, and recovery is still available through the daemon startup surface from main. The retry action itself now uses aoBridge.daemon.restart(), matching the current main recovery path, and retry failures are shown inline.

frontend/src/renderer/components/XtermTerminal.tsx
JAYATIAHUJA: Fixed in d0bab3ad. The normalizer now keeps using the streaming TextDecoder after light-mode buffering has started, so a light-to-dark transition flushes any pending SGR tail and preserves split UTF-8 continuations instead of returning raw bytes. Added a regression test for split λ across a light-to-dark theme change.

JAYATIAHUJA:
@Hardik180704 Fixed in d0bab3ad.

- The terminal normalizer now keeps the streaming TextDecoder active after light-mode buffering starts, so light-to-dark transitions preserve split UTF-8 continuations and flush pending SGR text instead of returning raw bytes.
- Added the light-to-dark split lambda regression test.
- Resolved the DaemonFailureBanner merge conflict and formatted the retryDaemon section while wiring retry through aoBridge.daemon.restart().

Verification:
- npm --prefix frontend run typecheck
- npm --prefix frontend run test -- src/renderer/components/DaemonFailureBanner.test.tsx src/renderer/components/XtermTerminal.test.tsx src/renderer/lib/terminal-themes.test.ts
- git diff --check

JAYATIAHUJA:
@Pulkit7070 Addressed both review notes in d0bab3ad.

- The terminal normalizer now preserves buffered decoder state across a light-to-dark theme transition, so split UTF-8 and partial SGR tails are not dropped.
- Added the light-to-dark split lambda regression test.
- Confirmed the dismiss behavior: recovery remains available through the daemon startup surface from main, and retry now uses aoBridge.daemon.restart() with inline retry errors.

Both inline threads are replied to and resolved.

Hardik180704 (approved):
Re-reviewed the latest d0bab3ad changes. The earlier light-to-dark streaming issue is fixed: the decoder now preserves buffered UTF-8/SGR state across the theme transition, and the split-λ regression test covers the failure we reproduced. The daemon retry path also uses aoBridge.daemon.restart() and handles both thrown and returned error states inline.\n\nVerified locally:\n- frontend typecheck\n- 68 focused daemon-banner, xterm, and terminal-theme tests\n\nCI is green and I found no remaining blockers.

frontend/src/styles/tokens.css
```diff
@@ -446,6 +446,26 @@
 	--color-term-bright-cyan: #79d4df;
 	--color-term-bright-white: #f4f5f7;
 
+	/* Light terminal ANSI palette tokens. Kept globally addressable so xterm can
+	   build a light palette even while the document is transitioning themes. */
+	--color-term-light-black: #e5e7eb;
```
Pulkit7070, line 451: In light mode this `--color-term-light-black` (#e5e7eb) feeds xterm's `black` slot (via `--color-term-black` at line 624), and xterm uses that one slot for BOTH SGR 40 backgrounds and SGR 30 foregrounds. So a program that emits ANSI black text on the default background (SGR 30, no background set) now renders near-white #e5e7eb on the #f5f5f4 light terminal background and is unreadable, which runs counter to the contrast fix.

The same token also backs `--color-term-light-panel-bg` (line 467), so it is serving two roles at once. Suggest decoupling them: keep the ANSI black slot dark enough to stay legible as foreground text, and use a separate panel token for the dark-background remap.

i-trytoohard:
Closing as stale: approved 40 days ago but never merged. The contrast fix is good — branch stays up, reopen and it should sail straight through.

PR #3229, opened 2026-07-28
fix: allow project setup inside ancestor repos

### LINKED ISSUE #3228: Project creation offers git init then rejects with PROJECT_PATH_NOT_REPO_ROOT when folder is inside a parent git repo
## Summary

When importing a folder that is **not itself a git repo** but is **nested inside a parent git repo** (e.g., `~/universe` where `~` or some ancestor has a `.git`), the **Project** (single repo) create flow breaks:

1. AO shows the agent selection sheet with the message **"If this folder needs Git setup, AO will initialize it and create the first commit before starting."**
2. User clicks "Create and start"
3. AO immediately fails with:
   > **IMPORT FAILED · PROJECT NOT REGISTERED**
   > Selected folder is inside a Git repository. Select the repository root instead.
   > `(PROJECT_PATH_NOT_REPO_ROOT)`

The frontend offers to init, the backend rejects it. Broken experience.

> **Confirmed by reporter:** The same folder imported as a **Workspace** (instead of Project) works correctly. This is consistent with the root cause below — the workspace path skips the broken frontend detection entirely.

## Reproduction

1. Create a plain folder inside a directory that is itself a git repo (e.g., `~` is a dotfiles repo, and you create `~/universe`)
2. New Project → **Project** → pick the plain folder
3. See "will initialize Git" message + agent sheet
4. Click Create and start
5. Get the PROJECT_PATH_NOT_REPO_ROOT error

## Root Cause

**Frontend/backend git detection mismatch — single-repo path only.**

### Why only Project breaks (not Workspace)

In `CreateProjectFlow.tsx:70`, the git-setup detection only runs for single-repo projects:

```ts
if (path && kind === "single_repo") {    // ← workspace skips this entirely
    const setupCode = await repositorySetupRequired(path);
    setRepositorySetup(setupCode);
}
```

The **Workspace** flow never calls `repositorySetupRequired`, so it never makes the wrong "not a git repo" determination. It goes straight to `onCreateProject`, and the backend's `AddWorkspace` handles plain parent folders by initializing them as a root repo — no `classifyRepositorySetupTarget` rejection.

### The mismatch

**Frontend scan** (`scanGitRepo` — `frontend/src/main.ts:1236`):

When `repoPath/.git` does not exist, the function falls to the catch block (line 1253). It only checks if the folder is a bare repo, then **returns `null`** (line 1270). It does **not** check whether a parent directory is a git repo.

`null` → `repos.length === 0` → `repositorySetupRequired()` returns `"NOT_A_GIT_REPO"` → frontend offers the git-init flow.

**Backend classify** (`classifyRepositorySetupTarget` — `backend/internal/service/project/service.go:332`):

The backend runs `git rev-parse --show-toplevel` from the selected path. Git traverses **up** the directory tree and finds the ancestor repo's root. Since the ancestor root differs from the selected path, it returns `PROJECT_PATH_NOT_REPO_ROOT`.

```go
// service.go:332-341
if top, err := gitOutput(ctx, path, "rev-parse", "--show-toplevel"); err == nil {
    root := normalizeGitReportedPath(path, strings.TrimSpace(top))
    selected := comparablePath(path)
    if !samePath(root, selected) {
        return repositorySetupPlainFolder, apierr.Invalid("PROJECT_PATH_NOT_REPO_ROOT", ...)
    }
}
```

| Layer | Detection method | Result for nested folder |
|---|---|---|
| Frontend `scanGitRepo` | `stat(repoPath/.git)` only | "Not a git repo" (misses parent) |
| Backend `classifyRepositorySetupTarget` | `git rev-parse --show-toplevel` (traverses parents) | "Inside a git repo" (correct) |

## Suggested Fix

**Frontend `scanGitRepo`** (or `repositorySetupRequired` in `CreateProjectFlow.tsx:241`) should run `git rev-parse --show-toplevel` from the selected folder in the no-`.git` catch path. If it succeeds and returns a path different from the selected folder, return an error scan result so the frontend shows `PROJECT_PATH_NOT_REPO_ROOT` **upfront** instead of offering to init.

This mirrors exactly what the backend's `classifyRepositorySetupTarget` does at line 332 — the two layers should agree.

## Environment

- Platform: Desktop (Electron)
- Path in screenshot: `~/universe`

## Reported by

@prateek — Discord #bug-triage

AgentWrapper:
## Summary
- align Project setup with Workspace setup: a selected plain folder can be initialized as its own repo root even when an ancestor directory is already a Git repo
- show a warning in the Project agent sheet when the selected plain folder is inside an existing ancestor Git repo, so accidental subdirectory selections are visible before initialization
- keep unsafe AO-state paths blocked up front and keep direct add validation as NOT_A_GIT_REPO until setup has run
- extract the Electron import-folder scanner into a tested main-process helper

Fixes #3228

## Tests
- go test ./internal/service/project
- npm test -- --run src/main/import-folder-scan.test.ts src/renderer/components/Sidebar.test.tsx
- npm run frontend:typecheck
- env -u AO_DATA_DIR -u AO_ISSUE_ID -u AO_OWNER -u AO_PROJECT_ID -u AO_RUNTIME_LAUNCH_ID -u AO_SESSION_ID -u AO_TELEMETRY_EVENTS -u AO_TELEMETRY_POSTHOG_HOST -u AO_TELEMETRY_POSTHOG_KEY -u AO_TELEMETRY_REMOTE npm run lint

## Notes
- Initial npm run lint failed because AO session environment variables affected existing CLI tests; rerun with AO_* session context cleared passed after clearing stale golangci-lint cache.

### DIFF
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -235,6 +235,9 @@ func (m *Service) Add(ctx context.Context, in AddInput) (Project, error) {
 		return p, nil
 	}
 	if !isGitRepo(path) {
+		if root, ok := gitRepositoryRoot(ctx, path); ok && !samePath(root, comparablePath(path)) {
+			return Project{}, projectPathNotRepoRootError(path, root)
+		}
 		return Project{}, apierr.Invalid("NOT_A_GIT_REPO", "AO needs a Git repository with an initial commit before it can create agent workspaces.", nil)
 	}
 	if !repoHasCommit(ctx, path) {
@@ -329,15 +332,10 @@ func classifyRepositorySetupTarget(ctx context.Context, path string) (repository
 		return repositorySetupUnbornRepo, nil
 	}
 
-	if top, err := gitOutput(ctx, path, "rev-parse", "--show-toplevel"); err == nil {
-		root := normalizeGitReportedPath(path, strings.TrimSpace(top))
+	if root, ok := gitRepositoryRoot(ctx, path); ok {
 		selected := comparablePath(path)
 		if !samePath(root, selected) {
-			return repositorySetupPlainFolder, apierr.Invalid("PROJECT_PATH_NOT_REPO_ROOT", "Selected folder is inside a Git repository. Select the repository root instead.", map[string]any{
-				"path":         path,
-				"repoRoot":     root,
-				"suggestedFix": "Select the repository root folder, then try again.",
-			})
+			return repositorySetupPlainFolder, projectPathNotRepoRootError(path, root)
 		}
 		return repositorySetupPlainFolder, apierr.Invalid("UNSUPPORTED_GIT_REPO", "Selected folder contains an unsupported Git repository layout.", map[string]any{"path": path})
 	}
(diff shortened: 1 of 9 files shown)


AgentWrapper (commented):
Ready to merge. I reviewed the nested project folder preflight/backend setup changes and did not find any blocking issues.

Tests run:
- `cd backend && go test ./internal/service/project`
- `cd frontend && npm test -- import-folder-scan.test.ts Sidebar.test.tsx`

neversettle17-101 (approved):
LGTM , We need to replicate/explore the same flow for workspace registration it can be taken as a separate PR. Created issue here: https://github.com/AgentWrapper/agent-orchestrator/issues/3230

PR #3231, opened 2026-07-28
fix(tmux): complete accepted message delivery after cancellation

### LINKED ISSUE #3221: send: GUI-typed message occasionally sits unsubmitted in worker input until next send flushes both
## Summary

`POST /api/v1/sessions/{id}/send` (and `ao send`) occasionally leaves a message unsubmitted in the worker’s TUI input area. The endpoint returns `200 OK` because `tmux send-keys` exits 0, but the agent never actually receives the prompt — the text sits in the prompt buffer as a draft, indistinguishable from an idle worker. Sending another message via the daemon API then appends to the same draft and triggers a single Enter that submits **both** messages together, looking like "the next send flushed the stuck one and the new one".

A reliable report (3 occurrences on 2026-07-28) had one session sitting idle for ~4 hours waiting on a reply that was never going to come because the prompt was never delivered.

## Repro (best-effort, intermittent)

1. Run the AO desktop app and connect to a worker session running Claude Code (or another TUI-based harness).
2. When the agent is **mid-rendering a long response** (lots of tokens streaming, agent busy), `ao send --session <id> --message "<a non-trivial prompt>"` (or the GUI equivalent) the worker.
3. In a fraction of attempts, the message sits in Claude Code’s prompt input area un-submitted. The `ao send` command returns 0 immediately. No error, no warning.
4. Issue another `ao send` (or any payload via `POST /api/v1/sessions/{id}/send`). The chunks of the second message are appended after the stranded first message; the second message’s trailing Enter submits both messages together.

The race is timing-sensitive (depends on how saturated the TUI’s input buffer is when chunks land), so it is hard to reproduce reliably. The user impact is asymmetric: when it happens, the GUI shows no error and the user has no signal until they happen to send again.

## Root cause

The single-Enter contract still has soft edges that the existing pre-Enter pause + post-Send `confirmActive` rescue loop (introduced by #2342 / #2357) do not fully cover.

`backend/internal/adapters/runtime/tmux/tmux.go:411` `SendMessage`:

- Chunks are written with `r.run(ctx, ...)` over the **caller’s** `ctx`. If the caller’s HTTP request is cancelled mid-paste (renderer re-render, network blip, client disconnect), the chunk loop short-circuits on the cancellation error **before** the post-paste pause / detached Enter can run. Chunks already accepted by tmux remain in the pane as a draft, with no Enter coming to submit them.
- The detached-Enter contract (`enterCtx = context.WithTimeout(context.WithoutCancel(ctx), r.enterDelay+5s)` at line 434) only protects Enter **after** the chunk loop completes. A cancellation that lands inside the chunk loop skips the detached-Enter code entirely.

`backend/internal/session_manager/manager.go:1764` `Manager.Send` + `confirmActive`:

- Pre-Enter pause (`defaultEnterDelay = 300ms`, `tmux.go:31`) is fixed and small. For a large multiline paste into a busy TUI, 300ms can be too short for the TUI to absorb the chunks before Enter arrives — Enter then lands inside the pasted text and submits nothing.
- `confirmActive`’s budget (`sendConfirmAttemptDeadline = 2s`, `sendConfirmMaxAttempts = 3`, `manager.go:210-212`) gives the TUI ~6 seconds total to flip Activity.State to active. For a slow TUI that takes longer to absorb a large paste and start the agent’s user-prompt-submit hook, the rescue loop gives up silently — Send still returns nil.
- `confirmActive` is skipped entirely for harnesses that don’t implement both `EmitsSubmitActivity` AND `EmitsBlockedActivity` (`manager.go:1813` `harnessNudgeSafe` — see `adapters/agent/codex/codex.go:44` `EmitsBlockedActivity = false`). For those harnesses, a lost Enter has no rescue path at all.

The interlocking effect: when a chunked-paste races with a busy TUI, the post-Send confirmation either doesn’t run or runs out of budget, the controller returns `200 OK` to the GUI, and the message is silently stranded in the worker’s prompt input.

## Impact

- **Silent failure.** No signal to the operator (status code 200, no log, no notification). The session just sits idle.
- **Wasted user time.** A user waiting for a reply that will never come (one report: 4 hours).
- **Cross-contamination.** The NEXT send through the same daemon API path writes into the same input box. Its trailing Enter then submits accumulated drafts as one prompt, surprising the user with concatenated messages from prior un-sent sends.

## Suggested fix surface (not a proposed patch yet)

1. **`tmux.SendMessage` chunk-loop context:** once the first chunk has been written successfully, switch the chunk loop and the trailing Enter onto the same `enterCtx` (detached from the caller, bounded by a fixed timeout). This closes the "caller cancelled mid-paste, chunks stranded without Enter" hole.
2. **`sendConfirm` budget:** bump `sendConfirmAttemptDeadline` / `sendConfirmMaxAttempts` so the rescue loop has more headroom for slow TUIs. Optionally make the budget per-prompt-length scaled, not a single fixed budget.
3. **`confirmActive` telemetry:** when the rescue loop exhausts its budget without seeing ActivityActive, emit a structured `WARN` log line (session id, attempt count, observed states) so operators have a signal in telemetry even though Send returns 200.
4. **(Larger; out of scope for a small fix)** Extend `confirmActive` to non-safe harnesses with a guarded "nudge only when state is provably WaitingInput" fallback, so the rescue also works for Codex and similar adapters.

## Related

- Closes #2342 (which added Enter delay + `confirmActive`). This bug is the leftover soft edges of that fix.
- The review-nudge `sendOnce` / `MarkReviewRunDelivered` dedup-stamp path (also called out in #2342) has the same shape: a hung message becomes invisible AND locked out of redelivery.

Hardik180704:
Closes #3221

## What changed

- keep the first tmux message chunk tied to the caller context
- after that first chunk is accepted, send the remaining chunks, paste delay, and Enter using one bounded detached context
- log a warning when the activity-confirmation budget is exhausted
- preserve the existing confirmation budget and safe-harness behavior

## Why

A caller cancellation could arrive after tmux had accepted the first chunk but before the rest of the message and Enter were delivered. That left a partial draft in the pane, and retrying could then double-paste the prompt.

## Verification

- regression test fails against `main` and passes with this fix
- cancellation before the first accepted chunk still aborts
- cancellation after the first accepted chunk completes the remaining chunks and Enter
- confirmation-budget exhaustion emits the expected warning
- `npm run lint`
- `go test -race ./internal/adapters/runtime/tmux ./internal/session_manager`

### Before — current `main`

The caller cancellation reaches the second chunk and leaves the accepted message incomplete.

Issue #3221 regression test failing on main

### After — this fix

Once the first chunk is accepted, the remaining chunks and Enter complete successfully.

Issue #3221 regression test passing with the fix

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -490,9 +490,10 @@ func (r *Runtime) SendMessage(ctx context.Context, handle ports.RuntimeHandle, m
 	}
 	enterCtx := ctx
 	if message != "" {
+		messageChunks := chunks(message, r.chunkSize)
 		sendCtx := ctx
 		var finishCancel context.CancelFunc
-		for i, chunk := range chunks(message, r.chunkSize) {
+		for i, chunk := range messageChunks {
 			if _, err := r.run(sendCtx, sendKeysLiteralArgs(id, chunk)...); err != nil {
 				if finishCancel != nil {
 					finishCancel()
@@ -500,7 +501,8 @@ func (r *Runtime) SendMessage(ctx context.Context, handle ports.RuntimeHandle, m
 				return fmt.Errorf("tmux runtime: send message %s: %w", id, err)
 			}
 			if i == 0 {
-				enterCtx, finishCancel = context.WithTimeout(context.WithoutCancel(ctx), r.enterDelay+5*time.Second)
+				completionBudget := sendCompletionBudget(len(messageChunks), r.timeout, r.enterDelay)
+				enterCtx, finishCancel = context.WithTimeout(context.WithoutCancel(ctx), completionBudget)
 				sendCtx = enterCtx
 			}
 		}
@@ -517,6 +519,9 @@ func (r *Runtime) SendMessage(ctx context.Context, handle ports.RuntimeHandle, m
 		// the Enter are detached from the caller's cancellation (bounded by
 		// their own timeout instead): abandoning mid-pause would strand an
 		// unsubmitted draft that a retried send would then double-paste.
+		// Errors reported by tmux after it accepts a chunk still return to the
+		// caller; they are not retried because AO cannot safely distinguish
+		// whether tmux applied the failed command.
 		if r.enterDelay > 0 {
 			select {
 			case <-enterCtx.Done():
@@ -529,6 +534,10 @@ func (r *Runtime) SendMessage(ctx context.Context, handle ports.RuntimeHandle, m
 		return fmt.Errorf("tmux runtime: send enter %s: %w", id, err)
 	}
 	return nil
+}
+
+func sendCompletionBudget(chunkCount int, commandTimeout, enterDelay time.Duration) time.Duration {
+	return time.Duration(chunkCount)*commandTimeout + enterDelay
 }
 
 // Interrupt sends Ctrl-C to the foreground process without destroying the tmux
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -490,10 +490,22 @@ func (r *Runtime) SendMessage(ctx context.Context, handle ports.RuntimeHandle, m
 	}
 	enterCtx := ctx
 	if message != "" {
-		for _, chunk := range chunks(message, r.chunkSize) {
-			if _, err := r.run(ctx, sendKeysLiteralArgs(id, chunk)...); err != nil {
+		sendCtx := ctx
+		var finishCancel context.CancelFunc
+		for i, chunk := range chunks(message, r.chunkSize) {
+			if _, err := r.run(sendCtx, sendKeysLiteralArgs(id, chunk)...); err != nil {
+				if finishCancel != nil {
+					finishCancel()
+				}
 				return fmt.Errorf("tmux runtime: send message %s: %w", id, err)
 			}
+			if i == 0 {
+				enterCtx, finishCancel = context.WithTimeout(context.WithoutCancel(ctx), r.enterDelay+5*time.Second)
+				sendCtx = enterCtx
+			}
+		}
+		if finishCancel != nil {
+			defer finishCancel()
 		}
 		// Give the target TUI a moment to accept the pasted text before the
 		// trailing Enter, mirroring conpty's ptyInputEnterDelay. Without it a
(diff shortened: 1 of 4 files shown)


Pulkit7070 (commented):
The approach is sound: holding the caller context only until the first chunk is accepted, then finishing on a detached context, closes the window where a cancellation left a partial, unsubmitted draft in the pane. Tests cover both sides of that boundary and the confirmation-budget warning is a good addition.

Two things worth resolving before merge.

**The detached deadline now has to cover the whole remaining message, not just the pause and Enter.** enterCtx is created once at the first chunk with enterDelay + 5s, and from there it bounds every subsequent chunk send as well as the pause and the trailing Enter. Previously that 5s covered only the pause and Enter, while the chunk loop ran on the caller context. A message large enough to span many 16 KB chunks can exhaust that single deadline on a later chunk or on the Enter, and the send then returns an error with a partial draft already in the pane, which is the exact state this change is meant to prevent. Most agent prompts are a single chunk and never reach this, but a large paste can. Consider scaling the budget with the number of chunks still to send, or giving the pause and Enter their own deadline once the last chunk is in, so the Enter is never the call that times out.

**A genuine send failure after the first chunk still leaves a partial draft.** Cancellation is handled, but if run returns a real tmux error on, say, the third chunk, SendMessage returns while chunks one and two are already in the pane, and a caller that retries produces the double paste this change describes. The in-code comment already flags that double-paste risk for the mid-pause case. This may be acceptable and out of scope, so it would help to state which cases the retry path is expected to handle rather than leave it implicit.

Hardik180704:
Thanks @Pulkit7070 — addressed both points.

- The detached completion deadline now scales with the total chunk count (`chunkCount * commandTimeout + enterDelay`), so large multi-chunk pastes retain enough budget for the remaining sends and trailing Enter. I added a regression test covering the scaled budget.
- Genuine tmux errors after an accepted chunk are intentionally not retried because AO cannot safely determine whether tmux applied the failed command. That scope is now documented beside the delivery logic.

I also rebased onto latest `main`. Focused tests, the race detector, the full backend suite, and golangci-lint all pass locally.

i-trytoohard:
From the NotesBot call on `2026-07-28`, this existing pr appears to be relevant.
Match confidence: `0.67` (fuzzy).

Relevant discussion excerpts:
- - Added “live orchestrator activity” to make background orchestration visible in UI. - Fixed PostHog sync issue. - Fixed interrupted tmux message delivery: cancellation should not leave “partially pasted unsubmitted prompt after the first chunk is accepted.”
- - Fixed PostHog sync issue. - Fixed interrupted tmux message delivery: cancellation should not leave “partially pasted unsubmitted prompt after the first chunk is accepted.”
- - Fixed PostHog sync issue. - Fixed interrupted tmux message delivery: cancellation should not leave “partially pasted unsubmitted prompt after the first chunk is accepted.” - Stated intention to focus on “cloud side” going forward; prateek asked to sync next morning to align after prateek’s self-hosting [...]

Source thread: https://discord.com/channels/1476302178913357958/1531716639849185380/1531716639849185380

Pulkit7070:
lgtm

Pulkit7070 (approved):

PR #3233, opened 2026-07-28
fix: polish terminal topbar controls

JAYATIAHUJA:
## What

Polishes the desktop topbar and terminal controls based on visual review feedback.

## Why

The terminal font controls were floating over terminal content, the kill dialog overstated the action as irreversible, and the session branch crumb/kill button styling did not match the surrounding desktop typography and topbar button system.

## How

- Moves terminal display controls into a dedicated tokenized strip above the terminal body, with a subtle divider so terminal content starts below the controls.
- Updates the session branch crumb to use the same title-scale styling as the project label while preserving the branch icon and truncation.
- Makes the Kill button use the same neutral raised topbar button chrome as New task, swaps to the cleaner `Trash` icon, and removes the icon animation/destructive topbar styling.
- Updates the kill confirmation copy to say the session moves to the Archived section below the Kanban board and can be restored later.
- Tokenizes new control sizing through existing utilities (`h-control-sm`, `size-control-sm`, `w-font-size-label`, `size-icon-*`) instead of raw pixel values.

## Testing

- `npm --prefix frontend run test -- src/renderer/components/CenterPane.test.tsx src/renderer/components/ShellTopbar.test.tsx` passed: 38 tests.

before : 
Screenshot 2026-07-28 204651
Screenshot 2026-07-28 204803
Screenshot 2026-07-28 170904

After : 
Screenshot 2026-07-28 205743
Screenshot 2026-07-28 204527

## Checklist


Hardik180704 (approved):
Reviewed the topbar and terminal-control changes along with their tests. The controls now have dedicated layout space, the branch crumb and kill action remain responsive and accessible, and the confirmation copy matches the recoverable archived-session behavior. Focused component tests (38) and typecheck pass locally. No blocking issues found.

frontend/src/renderer/components/ShellTopbar.tsx
```diff
@@ -321,7 +321,7 @@ export function TopbarKillButton({
 					if (!kill.isPending) setConfirmOpen(open);
 				}}
 				title="Kill session?"
-				description={`Are you sure you want to kill "${session.title}"? This stops the agent and tears down its workspace. This cannot be undone.`}
+				description={`Are you sure you want to kill "${session.title}"? This stops the agent and moves the session to the Archived section below the Kanban board, where it can be restored later.`}
```
Pulkit7070: This softens the kill warning in a way that may now understate it. The old copy said it tears down its workspace and cannot be undone; this drops both and says the session moves to Archived where it can be restored later.

The session-restore half is accurate: there is a restore path. But restore is a relaunch, not a state recovery. RestoreMode in session_manager is native / saved_prompt / fresh, i.e. it resumes the agent transcript or relaunches from the saved prompt (or fresh), and Kill is explicitly a worktree-releasing path (the manager comments describe Kill as releasing and ForceDestroying the worktree). So the working tree, including any uncommitted changes, is torn down on kill and is not what restore brings back.

As written, a user could read can be restored later as their workspace and files being preserved, which they are not. Two ways to make it accurate: keep a short note that the working tree is discarded (e.g. its workspace is discarded; the session can be relaunched from Archived), or if Kill already refuses on a dirty worktree, say that instead so the no-data-loss claim is grounded. Please confirm which is true and align the copy. The removal of cannot be undone specifically is the part I would not ship without that confirmation.

Pulkit7070 (commented):
Reviewed line by line. The relocation of the terminal display controls from a floating absolute overlay into a bordered strip above the terminal body is clean and does fix the float-over-content problem (the new test asserts the controls sit outside the tab bar and precede the terminal body). The branch-crumb restyle to topbarProjectLabelClass and the icon size-token swaps (size-5 to size-icon-base, size-icon-md to size-icon-sm) are cosmetic and fine. TopbarButton kill variant losing the destructive red styling is consistent with the copy change below.

One substantive thing to confirm, on the kill dialog copy (inline). The rest is polish; this is the only change that alters a user-facing safety claim.

frontend/src/renderer/components/ShellTopbar.tsx
JAYATIAHUJA: Verified against the lifecycle/workspace behavior: Kill is the worktree-discarding path, while Archived restore is a relaunch/resume path rather than file-state recovery.

Updated the copy to make that explicit:

> This stops the agent, discards its workspace, and moves the session to the Archived section below the Kanban board. It can be relaunched from Archived later.

Also updated the ShellTopbar test assertion and reran:
- `npm --prefix frontend run test -- src/renderer/components/CenterPane.test.tsx src/renderer/components/ShellTopbar.test.tsx` - 38 passed
- `npm --prefix frontend run typecheck` - passed

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -207,55 +207,54 @@ export function CenterPane({
 					<span className="ml-auto truncate font-mono text-xs text-passive">{target.harness}</span>
 				</div>
 			) : null}
-			<div className="relative min-h-0 flex-1">
+			<div
```
Pulkit7070, line 210: `aria-label` on a role-less `div` is not dependably announced by screen readers, so this group name is effectively test-only (getByLabelText finds it, but assistive tech at runtime may not). Since this wraps the font-size and fullscreen controls, add `role="group"` (or `role="toolbar"`) here so the label is actually exposed and the test reflects real semantics.

Rishet11:
The topbar area this polishes has been redesigned three times since: #3556, #3874, and #4348. The controls this PR adjusts no longer exist in that form. Closing as obsolete, and happy to re-review if you rebase onto the current design.

PR #3235, opened 2026-07-28
feat: restore cached workspace state during daemon startup

nikhilachale:
## Summary
- persist the last successful workspace state in the Electron UI cache and restore it before the daemon is ready
- keep cached data visible and read-only until the current daemon successfully hydrates live workspace state
- retain cached boards and project navigation when background refreshes fail
- validate nested cached workspace, session, activity, and pull-request data before restoration

## Validation
- npm run typecheck
- npx vite build --config vite.renderer.config.ts
- full frontend test suite: 1,175 tests verified
- exact-worktree Browser preview verified for cached board and session controls


Rishet11:
Closing as stale. There has been no activity since July 30, the branch conflicts with main, and the PR never received review. Cached workspace restore is not on main, so a rebased refresh would be welcome if this is still pursued.

PR #3251, opened 2026-07-29
feat(cloud): multi-tenant control plane skeleton (phase 1)

AgentWrapper:
## Summary

Draft Phase 1 implementation of AO Cloud's multi-tenant control-plane skeleton.

Architecture decisions referenced from the maintainer-approved 2026-07-28 direction:
- multi-tenant control plane from day one, while reusing this repo's daemon services/API boundaries instead of forking core behavior
- Google OAuth-only identity for v1 with short-lived JWT access tokens and refresh tokens
- users -> orgs -> teams tenancy model with org-scoped domain rows
- managed Postgres state, with the existing AO durable project/session/CDC shape ported behind a tenant-scoped store
- server-side encrypted credential custody through a SecretsManager port and local AES-GCM dev implementation
- usage instrumentation schema only; no pricing or payment flows

## Current progress

- Added backend/cmd/ao-cloud.
- Added backend/internal/cloud packages for auth, tenancy middleware, Postgres storage, and secrets.
- Added Postgres goose migration for control-plane tables plus tenant-scoped AO project/session/CDC tables.
- Added optional request-aware CDC event filtering hook to the existing HTTP event controller.

## Tests

- go test ./internal/cloud/... ./cmd/ao-cloud

## Remaining before ready for review

- Add Postgres integration test proving org isolation across existing project/session API routes.
- Add docs/cloud/control-plane.md local bring-up runbook.
- Rerun go test ./... and fix any failures.


AgentWrapper:
## Adversarial security / tenancy audit

I reviewed the full PR diff plus the surrounding `backend/internal/cloud` and touched `httpd` event code. This is intentionally focused on ways to break tenant isolation and auth, not feature completeness.

### Critical

1. **Unauthenticated device approval can mint tokens for an arbitrary known user id**

   - `backend/internal/cloud/server.go:94-99` registers all `/auth/*` routes outside the tenancy middleware.
   - `backend/internal/cloud/auth/handlers.go:133-142` accepts `userId` directly from the request body for `/auth/device/approve`.
   - `backend/internal/cloud/postgres/auth_store.go:140-145` writes that arbitrary `userID` into `devices.approved_user_id` with no authenticated actor check.
   - `backend/internal/cloud/auth/handlers.go:154-171` then lets the holder of the matching `deviceCode` exchange it for an access token and refresh token for that approved user and all of their orgs.

   **Exploit scenario:** an attacker calls `/auth/device/code` to get their own `{ deviceCode, userCode }`, then POSTs `/auth/device/approve` with `{"userId":"<victim-user-id>","userCode":"<attacker-user-code>"}`. If the victim user id is known or leaked, the attacker can immediately poll `/auth/device/token` with their `deviceCode` and receive bearer/refresh tokens scoped to the victim's org memberships. No victim session, Google login, or org membership check is involved in approval.

   **Minimal fix:** `/auth/device/approve` must authenticate the approving browser/user through Google/session/JWT and derive the user id server-side from that authenticated principal. Do not accept `userId` from the request body. The store update should also require the code to be unapproved (`approved_at IS NULL`) and should return the approved authenticated user only.

### High

2. **Refresh-token rotation is replayable under concurrent use**

   - `backend/internal/cloud/postgres/auth_store.go:104-109` selects a usable refresh token.
   - `backend/internal/cloud/postgres/auth_store.go:116` revokes it in a separate statement without atomically claiming the row.
   - `backend/internal/cloud/auth/handlers.go:76-93` issues a new token pair after `ConsumeRefreshToken` returns `ok`.

   **Exploit scenario:** if a refresh token is stolen, the attacker can race the legitimate client. Under PostgreSQL read-committed semantics, both transactions can read `revoked_at IS NULL` before either update commits; both then update the same row and both receive fresh access and refresh tokens. That defeats one-time refresh-token rotation and lets a replayed token become a persistent session.

   **Minimal fix:** consume refresh tokens with one atomic statement, e.g. `UPDATE api_tokens SET revoked_at = $now WHERE token_hash = $hash AND kind = 'refresh' AND revoked_at IS NULL AND (expires_at IS NULL OR expires_at > $now) RETURNING user_id`, then load the user/orgs only if the update returned a row. `SELECT ... FOR UPDATE` with a recheck before returning `ok` is also acceptable.

### Medium

3. **CDC store has an unscoped fallback that leaks all orgs if any caller misses tenancy context**

   - `backend/internal/cloud/postgres/changelog_store.go:17-35` uses `currentOrg(ctx)` and falls back to `WHERE seq > $1` across the entire `change_log` table when no org scope is present.
   - `backend/internal/cloud/postgres/changelog_store.go:58-66` does the same for `LatestSeq`.
   - Current HTTP wiring mitigates this for `/api/v1/events`: `backend/internal/cloud/server.go:111-118` installs an org filter and mounts `api.Register` under `tenancy.Middleware`.

   **Exploit scenario:** the present cloud HTTP route appears protected, but the store itself does not fail closed. Any future streaming/admin/debug handler, background subscriber, or test helper that calls `store.EventsAfter(context.Background(), ...)` with the cloud store will receive every org's events. Because `change_log` includes project/session ids and event payloads, this is a cross-tenant data leak if one route forgets the middleware or context propagation.

   **Minimal fix:** make the cloud `EventsAfter`/`LatestSeq` store methods require `tenancy.OrgIDFromContext(ctx)` like the project/session/worktree stores. If a truly global admin stream is needed, add a separate explicitly named admin method/interface instead of overloading the tenant store path.

4. **Request-time org authorization is based only on org ids embedded in the access token**

   - `backend/internal/cloud/auth/tokens.go:81-88` embeds `OrgIDs` at issue time.
   - `backend/internal/cloud/tenancy/middleware.go:39-59` verifies the token and checks only `slices.Contains(claims.OrgIDs, orgID)` for `X-AO-Org-ID`.
   - `backend/internal/cloud/postgres/auth_store.go:202-209` lists memberships during token issuance/refresh, but there is no DB membership check in middleware.

   **Exploit scenario:** if a user's org membership is revoked, or if a token was minted with an org membership that should no longer apply, the bearer token remains authorized for that org until access-token expiry. That creates a revocation window for tenant data reads/writes. It is bounded by the current 15-minute default, but it is still not a request-time membership authorization check.

   **Minimal fix:** have the tenancy middleware call an authz store method such as `IsOrgMember(ctx, subject, orgID)` before setting scope, or add token/session versioning that invalidates existing access tokens when membership changes. Keep the store-level `org_id` predicates either way.

5. **JWT and Google ID-token validation lack audience checks**

   - `backend/internal/cloud/auth/tokens.go:119-125` has no `aud` claim, and `backend/internal/cloud/auth/tokens.go:141-165` verifies signature/issuer/expiry but not audience.
   - `backend/internal/cloud/auth/google.go:94-115` relies on Google's `tokeninfo` status and then accepts `sub/email/name`; it does not check the returned `aud` against `GoogleConfig.ClientID` (nor does it explicitly validate issuer/email_verified in this code).

   **Exploit scenario:** an AO Cloud access token minted for another audience/environment can be accepted here if it uses the same issuer/secret. For Google, I am not asserting `tokeninfo` fails open, but this code does not enforce that the ID token belongs to this configured OAuth client, so the guarantee is external and untested.

   **Minimal fix:** add an audience config/claim to AO Cloud JWTs and require it in `VerifyAccessToken`. For Google, either use Google's ID-token verifier library with audience set to `ClientID`, or parse the `tokeninfo` response and require `aud == cfg.ClientID` plus issuer/expiry/email verification as appropriate.

6. **OAuth `state` is generated but never validated**

   - `backend/internal/cloud/auth/handlers.go:45-50` generates a `state` for Google login.
   - `backend/internal/cloud/auth/handlers.go:53-63` handles the callback using only `code`; it does not read or validate `state`.

   **Exploit scenario:** this permits login CSRF/account-confusion flows where a callback initiated for one browser context can be completed without proving it matches the login request that started in that context. This is less direct than the device-code issue because the callback currently returns token JSON, but it is still a broken OAuth invariant.

   **Minimal fix:** persist a one-time state nonce server-side or in a secure/httpOnly SameSite cookie and require callback `state` to match before exchanging the code. Consume the nonce on success/failure.

### Notes / not findings

- I did not find raw SQL injection in the new cloud store code. Queries are parameterized; the concatenated SQL fragments I saw are constant local `WHERE` fragments, not user input.
- Implemented project/session/worktree/PR read paths generally enforce `org_id` at the store/query layer via `tenancy.OrgIDFromContext(ctx)` and `WHERE org_id = $1` predicates. The migration also uses composite `(org_id, id)` primary keys/foreign keys for the tenant-owned domain tables.
- `X-AO-Org-ID` spoofing is blocked in the current middleware for REST/SSE routes because the selected org must be present in the token's org list.
- The current `/api/v1/events` cloud route is mounted under tenancy middleware and also has an event filter; the issue above is specifically the store's unscoped fallback if any caller bypasses that route shape.
- Secrets custody is mostly schema/port skeleton in this PR. The migration stores `ciphertext`, `nonce`, `key_id`, and `algorithm`, and I did not see a plaintext credential API response/log path. However, `backend/internal/cloud/server.go:50` constructs and discards the `SecretsManager`; I do not see credential persistence wired through encryption yet, so encrypt-before-persist is not proven by this PR.

### Verification run

- `cd backend && go test ./internal/cloud/... ./cmd/ao-cloud` passed.

### Verdict

I would not merge this as a multi-tenant control-plane skeleton yet. The project/session store scoping is in the right direction, but the unauthenticated device approval path is a blocking account/org takeover bug, and refresh-token rotation is replayable under race. I would block merge until at least those two auth issues are fixed, and I would strongly prefer the CDC store to fail closed on missing tenant scope before more routes build on it.

AgentWrapper:
Security audit fixes pushed in 7222dba85.

Fixed:
- /auth/device/approve now requires a valid access token, derives the approving user from the token subject, rejects userId in the body, and only approves unapproved device codes.
- Refresh-token rotation is now a single atomic UPDATE ... RETURNING operation, with a concurrency regression test proving exactly one consume wins.
- Tenant CDC reads fail closed without org scope.
- Tenant middleware now checks current org membership in Postgres at request time.
- AO Cloud JWTs now carry and verify aud; Google tokeninfo validation checks aud, issuer, and email_verified.
- OAuth callback now validates an HttpOnly/Secure/SameSite state cookie.
- Credential persistence now goes through an encryption service before writing agent_credentials, with a custody test.

Local verification:
- cd backend && go test ./internal/cloud/...
- npm run lint

CI has started on the pushed commit and is currently pending.

backend/internal/cloud/server.go
```diff
@@ -0,0 +1,160 @@
...
+		envelope.WriteJSON(w, http.StatusOK, map[string]string{"status": "ok", "service": "ao-cloud"})
+	})
+
+	projectSvc := projectsvc.NewWithDeps(projectsvc.Deps{Store: store, DefaultHarness: domain.HarnessCodex})
+	sessionSvc := sessionsvc.NewWithDeps(sessionsvc.Deps{Manager: cloudCommander{}, Store: store})
+	api := httpd.NewAPI(daemonconfig.Config{RequestTimeout: cfg.RequestTimeout}, httpd.APIDeps{
+		Projects: projectSvc,
+		Sessions: sessionSvc,
+		CDC:      store,
+		Events:   cdc.NewBroadcaster(),
```
AgentWrapper: This creates a fresh broadcaster for `/api/v1/events`, but nothing ever tails the Postgres `change_log` and publishes into it. The local daemon calls `cdc.NewPoller(...).SeekToHead(...).Start(...)`; ao-cloud does not, and the Postgres source currently requires request tenancy scope so a normal poller cannot read all orgs either. As a result SSE clients only get the initial replay and then miss live session/project changes until they reconnect with a new `after` cursor. Please wire a cloud CDC poller/source or do not expose the live events endpoint yet.

backend/internal/cloud/server.go
```diff
@@ -0,0 +1,160 @@
...
+			scope, ok := tenancy.ScopeFromContext(r.Context())
+			return ok && scope.OrgID != "" && e.OrgID == scope.OrgID
+		},
+	})
+	r.Group(func(r chi.Router) {
+		r.Use(tenancy.Middleware(issuer, store))
+		if credentialSvc != nil {
+			r.Post("/api/v1/agent-credentials", credentialSvc.CreateHTTP)
+		}
+		api.Register(r)
```
AgentWrapper: Mounting the whole daemon API exposes all write/side-effecting routes in ao-cloud, not just the read paths described for phase 1. For example `POST /api/v1/projects` and `/projects/initialize` run the project service and inspect/mutate the cloud server filesystem, while session/review/shell routes are wired to partial stubs. This is a security and correctness issue for a hosted control plane; please register only the supported read endpoints or return explicit 501s for unsupported cloud operations.

AgentWrapper (commented):
Requesting changes. I found two issues that should be fixed before this is ready.

Tests run: `cd backend && go test ./...`.

backend/internal/cloud/server.go
AgentWrapper: Fixed in 31ab59e96. The shared daemon API still registers the full surface, but ao-cloud now uses a new read-only registration path that mounts only project/session read routes plus /api/v1/events. POST /api/v1/projects, project initialization, session mutation/runtime routes, review/shell/import/push routes, etc. are no longer mounted on ao-cloud. Added TestControlPlaneMountsReadOnlyAPISurface to cover that write routes do not succeed.

Verification: npm run lint.

backend/internal/cloud/server.go
AgentWrapper: Fixed in 31ab59e96. ao-cloud now creates one CDC broadcaster in Run, starts a cdc.Poller against an explicitly named Postgres AdminChangeLogSource, seeks it to head on startup, and passes that broadcaster into the cloud handler. Tenant replay still uses the scoped Store.EventsAfter path, while live fanout is filtered per request org. Added TestControlPlaneEventsReceivesLivePostgresChange to prove /api/v1/events receives a live Postgres change through the poller/broadcaster path.

Verification: npm run lint.

AgentWrapper:
Closing this PR — cloud work is consolidating onto the contributor PR #3426. Keeping the branch intact as a reference implementation: #3251 for the adversarially-audited org-multi-tenant auth model, #3254 for the Daytona runtime adapter + end-to-end learnings. Not deleting the branch.

Issue #2802, opened 2026-07-18
bug(runtime/tmux): IsAlive reports pane-exists as agent-alive, masking crashed agents behind the keep-alive shell


anmol0b:
## Problem

`tmux.Runtime.IsAlive` (`backend/internal/adapters/runtime/tmux/tmux.go`) only
checks whether the tmux **session** exists (`tmux has-session`). It has no way
to tell whether the **agent process** inside the pane is still running.

The launch command built by `buildLaunchCommand` always appends a keep-alive
shell after the agent's argv:

```go
b.WriteString(strings.Join(parts, " "))
// Keep the tmux session alive after the agent exits so the operator can
// inspect the terminal.
b.WriteString(`; exec "${SHELL:-/bin/sh}" -i`)
```

`;` runs unconditionally regardless of exit status, so a crash, an OOM kill,
or a clean exit all leave the pane alive on an interactive shell. From that
point on, `has-session` returns true forever, `IsAlive` reports `true`
forever, and the reaper (`backend/internal/observe/reaper/reaper.go`,
`probeOne`) only ever reports `ProbeAlive` to the lifecycle manager — it has
no signal that the agent itself is gone.

## Impact

- A session whose agent crashes while in a sticky state (`blocked` /
  `waiting_input`) never gets a `ProbeDead` fact, so the lifecycle fix in
  #2749 (for #2745) can't trigger for the tmux runtime — the sticky-state
  reducer fix is correct, but the fact it depends on is never produced for
  most real crashes. See [@neversettle17-101's review comment on
  #2749](https://github.com/AgentWrapper/agent-orchestrator/pull/2749#pullrequestreview-4729018515).
- Same architectural gap as #2089 (filed against the old TypeScript
  implementation's `isAlive()`); it was never fixed for the Go rewrite's
  tmux adapter.
- Any other automation that trusts `IsAlive`/`ProbeAlive` (session restore,
  dashboard status) is misled the same way.

## Repro

1. Spawn a session on the tmux runtime.
2. Find the agent's PID inside the pane and `kill -9` it (or let it crash),
   leaving the pane's shell alive.
3. Call `IsAlive` (or wait for the reaper's next tick) — it still reports
   `true` / `ProbeAlive`, even though the agent is gone.

## Proposed fix

Keep the fix contained to the tmux adapter — no changes needed to
`ports.ProbeResult`, `RuntimeFacts`, the reaper, or lifecycle.

[...]

Issue #2342, opened 2026-07-02
bug(send): ao send hangs on multiline — message sits unsubmitted in worker input


axisrow:
## Summary

`ao send` (orchestrator→worker, and back) intermittently leaves the message **unsubmitted** in the
worker's input field. The orchestrator returns HTTP 200 the moment the `tmux send-keys` commands
exit 0, but for a **large multiline** prompt claude-code's TUI never accepts the trailing `Enter`,
so the text sits as an unsubmitted draft until a human presses Enter. `UserPromptSubmit` therefore
never fires, and the orchestrator cannot tell the worker started — it looks identical to an idle
worker, so it polls endlessly. The inverse (worker→orchestrator) has the same symptom.

Related to #2105, but the root shifted: #2105 was the 0.9.2 bracketed-paste path; the Go rewrite
uses `send-keys -l`, yet the same class of failure persists because **paste and Enter fire
back-to-back with no confirmation that the prompt was accepted**.

## Reproduction (live, logged)

Two confirmed cases on the same machine (macOS, tmux, claude-code workers, `~/.ao/daemon.log`):

| time | session | bytes | HTTP | submitted? |
|---|---|---|---|---|
| 10:10:42 | srouter-38 | **3761** (multiline) | 200 | **no** — hung in input |
| 10:20:30 | srouter-37 | **3397** (multiline) | 200 | **no** — hung in input |
| 10:49:27 | srouter-38 | 482 (short) | 200 | yes |
| 10:49:27 | srouter-37 | 357 (short) | 200 | yes |

A manual `tmux send-keys -t <sess> Enter` seconds later submitted the hung text in both cases.

## Root cause

`SendMessage` (`backend/internal/adapters/runtime/tmux/tmux.go:198`) sends `send-keys -l <chunk>`
×N then immediately `send-keys Enter` — **no delay, no retry, no confirmation**:

- For large multiline input the single Enter is absorbed by the paste; claude-code never emits
  `UserPromptSubmit`, so `DeriveActivityState` never flips the session to `active`
  (`claudecode/activity.go:21` — the only producer of `active`).
- A hung message is observationally identical to a finished/idle worker (`status.go:54`: `no_signal`
  clears after the first signal), so the orchestrator can't distinguish "stuck" from "done".

The conpty runtime (Windows) already inserts delays (`conpty/client.go:36-71`: 15ms between chunks,
300ms before Enter); the tmux runtime does not.

## Suggested fix

[...]

PR #3254, opened 2026-07-29
feat(runtime): Daytona sandbox runtime adapter (phase 2)

AgentWrapper:
Phase 2 of AO Cloud: a Daytona-backed sandbox runtime + workspace adapter, so the unmodified session manager / lifecycle manager can run a cloud session. Runs in parallel with the phase-2 control-plane work — this PR deliberately touches nothing under `backend/cmd/ao-cloud` or `backend/internal/cloud/`.

Design doc (API survey + port mapping + attach story + hooks contract + cost model): `docs/cloud/daytona-runtime.md` — written from verified 2026 Daytona docs, both OpenAPI specs, and the official Go SDK source.

## What's implemented (`backend/internal/adapters/runtime/daytona/`)

- **Thin REST client** (`client.go`): platform API (create/list-by-label/start/stop/delete) + per-sandbox toolbox API (exec, git clone, PTY-over-WebSocket via the repo's existing `coder/websocket`). No new dependency tree; a narrow `Client` interface is the test seam. Live state is found by sandbox label `ao/session=<handle>` — no in-memory maps, so daemon restarts re-adopt cloud sessions like boot reconcile re-adopts tmux ones.
- **Runtime** (`daytona.go`, `attach.go`, `process.go`): tmux **inside** the sandbox, driven over toolbox exec, mirroring the tmux adapter's semantics (launch line with env exports + keep-alive shell, exact-match targets, chunked `send-keys` + delayed Enter #2342, capture-pane scrollback). Implements `ports.Runtime`, `ports.Attacher` (fresh PTY per attach running `tmux attach`; outbound-only, no inbound sandbox ports), `ports.RuntimeRestarter` (wake-from-park rebuilds tmux under the same handle), `ports.SupervisedProcessInspector` (#2802 pane-descendant walk via remote `ps`), and the `runtimeselect.Runtime` union (compile-time assert added).
- **Workspace** (`workspace.go`): sandbox provision (snapshot + boot-env credential injection per the locked custody model) + clone + session-branch checkout; dirty-refusal `Destroy`, `ForceDestroy`, remote `StashUncommitted`/`ApplyPreserved` (same temp-index / cherry-pick sequences as gitworktree), `AddExclude`, and `Park` (Daytona stop).
- **Hooks path** (`internal/cli/client.go`): `ao hooks` (and the whole CLI client) honors `AO_API_BASE` + `AO_API_TOKEN` — inside a sandbox there is no loopback daemon, so activity events post to the injected base with a bearer token. The loopback listener is untouched (still 127.0.0.1, unauthenticated).
- **Idle/cost mapping**: sandbox auto-stop (default 15 min, configurable/disable-able) + explicit `Park`. Parked = `IsAlive true` / supervised-process dead → AO's existing exited→resume flow wakes it via `Restart`. Cost: ~$0.067/h running (default size), ~$0.0003/h parked, $0 archived.

## Tests

- `go build ./...`, `go vet ./...`, full `go test ./...` and `-race` on the touched packages — green (the 5 pre-existing `internal/cli` spawn-test failures under an AO worker shell come from inherited `AO_SESSION_ID` env; they pass with a scrubbed env, unrelated to this PR).
- Contract tests over an in-memory fake Daytona (mirroring the tmux fake-runner seam): launch-line construction (no host-PATH leakage), parked-liveness, never-kill-on-transient probes, #2802 generations, dirty refusal + re-park, preserve-ref stale/clean/conflict mapping, chunked send round-trip, handle parity with `tmux.SessionName`.
- Live smoke (`smoke_live_test.go`), **skipped unless `DAYTONA_API_KEY` is set**: (1) sandbox+exec+PTY WebSocket round-trip, (2) full lifecycle — workspace create (clone) → tmux-hosted process → output capture → message round-trip → park → wake/restart → clean teardown, (3) opt-in real-agent demo via `AO_DAYTONA_AGENT_ARGV` (e.g. `["claude","-p","say hello"]`) with `CLAUDE_CODE_OAUTH_TOKEN` injected.

## Live validation (2026-07-28) — DONE, all green against real Daytona

Run with the maintainer's test org + the pushed `ao-agent-sandbox:dev` snapshot + a real `CLAUDE_CODE_OAUTH_TOKEN` (credentials via env only):

| Test | Result | Proves |
| --- | --- | --- |
| `TestLive_ClientSandboxLifecycle` | **PASS 4.4s** | sandbox create→started ~2–3s, exec, PTY-WebSocket round-trip + resize, delete |
| `TestLive_RuntimeSessionLifecycle` | **PASS 14.4s** | workspace (sandbox+clone) → tmux agent → scrollback → send round-trip → park → wake (same handle) → synchronous teardown |
| `TestLive_RealAgentSession` | **PASS 13.6s** | real Claude Code **authenticated with no browser flow** (boot-env token injection), answered a prompt; keep-alive shell held after exit |

The live runs surfaced three real async-state behaviors (stop/delete return 200 while state lags; list trails GET) — fixed + regression-tested in 475d285f8. Terminal streaming is validated at the port level (PTY WebSocket ↔ `ports.Stream`); in-app rendering awaits phase-3 per-session runtime selection — **merging this PR changes nothing for local installs** (`runtimeselect` still picks tmux/conpty by OS; no `AO_RUNTIME=daytona` gate is included). Confirmed gap: sandbox→daemon activity events need the control plane's reachable `AO_API_BASE` + per-session token (CLI side unit-tested). Full details: `docs/cloud/daytona-runtime.md` §6.

## What I need from you (maintainer) — ✅ both provided; live suite green

1. A Daytona org + API key (https://app.daytona.io/dashboard/keys, scopes: write/delete sandboxes) exported as `DAYTONA_API_KEY` to run the live smoke: `cd backend && go test ./internal/adapters/runtime/daytona/ -run TestLive -v`. ~$200 free credit covers this many times over (each run is a few sandbox-minutes).
2. For the full real-agent demo: a custom snapshot with tmux/git/node + Claude Code (+ the `ao` Linux binary for hooks) — `daytona snapshot create ao-agent --dockerfile …`; I can write that Dockerfile as a follow-up. Until then the lifecycle test best-effort apt-installs tmux into the default snapshot.

## Known gaps / follow-ups (documented in the design doc)

- Control plane must supply: reachable `AO_API_BASE` + per-session token minting/validation, the agent-harness snapshot, repo-URL resolver + git credentials for private repos (the default resolver reads local `origin` for the hybrid local-daemon setup).
- Spawn attachments / per-project provisioning write local files today — cloud sessions need those uploaded via the toolbox fs API (control-plane concern, phase 3).
- Preserve refs live on the sandbox disk; pushing them to `refs/ao/preserved/*` on origin is a follow-up.
- Parked sessions lose tmux scrollback (Daytona stop kills processes); agent transcripts survive on disk and native resume works.

### DIFF
--- a/backend/internal/adapters/runtime/daytona/workspace.go
+++ b/backend/internal/adapters/runtime/daytona/workspace.go
@@ -1,0 +1,594 @@
+package daytona
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os/exec"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// Workspace provisions a session's isolated checkout INSIDE its Daytona
+// sandbox: Create makes the sandbox (from the configured snapshot, with the
+// control plane's boot env injected) and clones the project repo into it;
+// Destroy tears the sandbox down after the same dirty-worktree refusal the
+// local gitworktree adapter enforces. Every git operation runs in the sandbox
+// via the toolbox API, mirroring the local adapter's command sequences.
+type Workspace struct {
+	core
+	opts          WorkspaceOptions
+	createTimeout time.Duration
+	logger        *slog.Logger
+}
+
+var _ ports.Workspace = (*Workspace)(nil)
+
+// RepoRemote tells the workspace adapter where a project's repository lives
+// from the sandbox's point of view, plus clone credentials for private repos.
+type RepoRemote struct {
+	// URL is the clone URL (https). Required.
+	URL string
+	// DefaultBranch is the branch session branches are created from when the
+	// workspace config does not name one.
+	DefaultBranch string
+	// Username/Password authenticate the clone (e.g. "x-access-token" + a
+	// GitHub token). They travel in the toolbox git API body, never on a
+	// command line.
+	Username string
+	Password string
+}
+
+// WorkspaceOptions configures the Daytona Workspace adapter.
+type WorkspaceOptions struct {
+	// Client talks to Daytona. Required.
+	Client Client
+	// Snapshot is the Daytona snapshot sandboxes are created from. It must have
+	// tmux, git, and the agent harness (agent CLIs + `ao` Linux binary)
+	// preinstalled; see docs/cloud/daytona-runtime.md. Required for Create.
+	Snapshot string
+	// Target picks the Daytona region ("us"/"eu"); empty uses the org default.
+	Target string
+	// CPU/MemoryGiB/DiskGiB size the sandbox; zero uses Daytona defaults.
+	CPU       int
+	MemoryGiB int
+	DiskGiB   int
+	// AutoStopMinutes is Daytona's inactivity auto-stop (the park mechanism in
+	// the cost model). 0 applies the adapter default (15); -1 disables
+	// auto-stop entirely.
+	AutoStopMinutes int
+	// AutoArchiveMinutes is how long a stopped sandbox waits before archiving
+	// to object storage; 0 keeps Daytona's default.
+	AutoArchiveMinutes int
+	// BootEnv is injected into the sandbox at creation — this is where the
+	// control plane places agent inference credentials (e.g.
+	// CLAUDE_CODE_OAUTH_TOKEN) and the hooks-path env (AO_API_BASE,
+	// AO_API_TOKEN). See the design doc's control-plane contract.
+	BootEnv map[string]string
+	// WorkspaceRoot is the directory inside the sandbox that holds session
+	// checkouts; default /home/daytona/ao.
+	WorkspaceRoot string
+	// ResolveRepo maps a workspace config onto a clone URL + credentials. The
+	// default resolver reads the `origin` remote of cfg.RepoPath on the daemon
+	// host (the hybrid local-daemon/cloud-runtime setup); the cloud control
+	// plane substitutes its own resolver.
+	ResolveRepo func(ctx context.Context, cfg ports.WorkspaceConfig) (RepoRemote, error)
+	// GitUserName/GitUserEmail configure commit identity inside the sandbox so
+	// agents can commit out of the box. Defaults: "AO Agent" /
+	// "ao-agent@users.noreply.github.com"; set either to "-" to skip.
+	GitUserName  string
+	GitUserEmail string
+	// ExecTimeout / StartTimeout mirror Options. CreateTimeout bounds sandbox
+	// creation (snapshot pull) and the initial clone; default 5m.
+	ExecTimeout   time.Duration
+	StartTimeout  time.Duration
+	CreateTimeout time.Duration
+	// Logger receives provisioning diagnostics; nil defaults to slog.Default().
+	Logger *slog.Logger
+}
+
+const (
+	defaultWorkspaceRoot   = "/home/daytona/ao"
+	defaultAutoStopMinutes = 15
+	defaultGitUserName     = "AO Agent"
+	defaultGitUserEmail    = "ao-agent@users.noreply.github.com"
+)
+
+// NewWorkspace builds the Daytona Workspace adapter.
+func NewWorkspace(opts WorkspaceOptions) (*Workspace, error) {
+	if opts.Client == nil {
+		return nil, errors.New("daytona workspace: client is required")
+	}
+	if err := validateEnvKeys(opts.BootEnv); err != nil {
+		return nil, fmt.Errorf("daytona workspace: boot env: %w", err)
+	}
+	if opts.WorkspaceRoot == "" {
+		opts.WorkspaceRoot = defaultWorkspaceRoot
+	}
+	if opts.AutoStopMinutes == 0 {
+		opts.AutoStopMinutes = defaultAutoStopMinutes
+	}
+	if opts.ResolveRepo == nil {
+		opts.ResolveRepo = resolveRepoFromLocalOrigin
+	}
+	if opts.GitUserName == "" {
+		opts.GitUserName = defaultGitUserName
+	}
+	if opts.GitUserEmail == "" {
+		opts.GitUserEmail = defaultGitUserEmail
+	}
+	execTimeout := opts.ExecTimeout
+	if execTimeout <= 0 {
+		execTimeout = defaultExecTimeout
+	}
+	startTimeout := opts.StartTimeout
+	if startTimeout <= 0 {
+		startTimeout = defaultStartTimeout
+	}
+	createTimeout := opts.CreateTimeout
+	if createTimeout <= 0 {
+		createTimeout = defaultCreateTimeout
+	}
+	logger := opts.Logger
+	if logger == nil {
+		logger = slog.Default()
+	}
+	return &Workspace{
+		core: core{
+			client:       opts.Client,
+			execTimeout:  execTimeout,
+			startTimeout: startTimeout,
+		},
+		opts:          opts,
+		createTimeout: createTimeout,
+		logger:        logger,
+	}, nil
+}
+
+// resolveRepoFromLocalOrigin is the default resolver for the hybrid setup
+// where the daemon runs next to the canonical repo: it reads the repo's
+// `origin` URL on the daemon host. Credentials are left empty (public repos,
+// or a snapshot with ambient git credentials).
+func resolveRepoFromLocalOrigin(ctx context.Context, cfg ports.WorkspaceConfig) (RepoRemote, error) {
+	if cfg.RepoPath == "" {
+		return RepoRemote{}, errors.New("daytona workspace: no repo path to resolve a clone URL from; configure WorkspaceOptions.ResolveRepo")
+	}
+	out, err := exec.CommandContext(ctx, "git", "-C", cfg.RepoPath, "remote", "get-url", "origin").Output()
+	if err != nil {
+		return RepoRemote{}, fmt.Errorf("daytona workspace: resolve origin of %s: %w", cfg.RepoPath, err)
+	}
+	return RepoRemote{URL: strings.TrimSpace(string(out))}, nil
+}
+
+// workspacePath is the checkout directory for a session inside its sandbox.
+func (w *Workspace) workspacePath(name string) string {
+	return strings.TrimRight(w.opts.WorkspaceRoot, "/") + "/" + name
+}
+
+// Create provisions the session's sandbox and clones the repo/branch into it.
+func (w *Workspace) Create(ctx context.Context, cfg ports.WorkspaceConfig) (ports.WorkspaceInfo, error) {
+	name, err := sessionName(cfg.SessionID)
+	if err != nil {
+		return ports.WorkspaceInfo{}, err
+	}
+	if w.opts.Snapshot == "" {
+		return ports.WorkspaceInfo{}, errors.New("daytona workspace: snapshot is required to create sandboxes")
+	}
+	remote, err := w.opts.ResolveRepo(ctx, cfg)
+	if err != nil {
+		return ports.WorkspaceInfo{}, err
+	}
+	if remote.URL == "" {
+		return ports.WorkspaceInfo{}, errors.New("daytona workspace: repo resolver returned an empty clone URL")
+	}
+
+	sb, found, err := w.sandboxForHandle(ctx, name)
+	if err != nil {
+		return ports.WorkspaceInfo{}, err
+	}
+	if !found {
+		sb, err = w.createSandbox(ctx, name, cfg)
+		if err != nil {
+			return ports.WorkspaceInfo{}, err
+		}
+	} else {
+		if sb, err = w.ensureStarted(ctx, sb); err != nil {
+			return ports.WorkspaceInfo{}, err
+		}
+	}
+
+	path := w.workspacePath(name)
+	branch, err := w.provisionCheckout(ctx, sb, path, cfg, remote)
+	if err != nil {
+		// The sandbox without a checkout is useless; tear it down so a retried
+		// spawn does not find a half-provisioned sandbox.
+		if delErr := w.client.DeleteSandbox(context.WithoutCancel(ctx), sb.ID); delErr != nil {
+			w.logger.Warn("daytona workspace: cleanup after failed provision", "sandbox", sb.ID, "error", delErr)
+		}
+		return ports.WorkspaceInfo{}, err
+	}
+	return ports.WorkspaceInfo{
+		Path:      path,
+		Branch:    branch,
+		SessionID: cfg.SessionID,
+		ProjectID: cfg.ProjectID,
+	}, nil
+}
+
+func (w *Workspace) createSandbox(ctx context.Context, name string, cfg ports.WorkspaceConfig) (Sandbox, error) {
+	autoStop := w.opts.AutoStopMinutes
+	if autoStop < 0 {
+		autoStop = 0 // Daytona: 0 disables auto-stop
+	}
+	req := CreateSandboxRequest{
+		Name:     "ao-" + name,
+		Snapshot: w.opts.Snapshot,
+		Env:      w.opts.BootEnv,
+		Labels: map[string]string{
+			LabelSession: name,
+			LabelProject: string(cfg.ProjectID),
+		},
+		Target:           w.opts.Target,
+		CPU:              w.opts.CPU,
+		Memory:           w.opts.MemoryGiB,
+		Disk:             w.opts.DiskGiB,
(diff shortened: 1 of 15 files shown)


AgentWrapper:
Follow-up landed on this branch: `test/daytona-snapshot/Dockerfile` — a custom Daytona snapshot with tmux, git, procps, node + Claude Code (pin via `--build-arg CLAUDE_CODE_VERSION`), and the `ao` Linux binary built from source (CGO-free), under the `daytona` user with home `/home/daytona` to match the adapter's default `WorkspaceRoot`. Build/push/run instructions are in `docs/cloud/daytona-runtime.md` §5a.

Verified locally: the image builds for linux/amd64, all tools resolve (tmux 3.3a, git 2.39, procps ps, node 22, Claude Code 2.1.220, `ao`), and tmux new-session/has-session/capture-pane work inside the container. With this snapshot registered (`daytona snapshot push ao-agent-sandbox:<version>`), the live tests run with `AO_DAYTONA_SNAPSHOT=ao-agent-sandbox:<version>` and skip the runtime tmux apt-install entirely.

AgentWrapper:
**Live validation complete against a real Daytona account** (maintainer-provided test key):

- `TestLive_ClientSandboxLifecycle`: **PASS (6.2s)** — sandbox create, exec, PTY-WebSocket round-trip with resize, delete.
- `TestLive_RuntimeSessionLifecycle`: **PASS (16.2s)** — full exit-criteria flow on the custom snapshot: workspace create (sandbox + clone) → tmux-hosted process → scrollback capture → send round-trip → park (stop) → wake (restart, same handle) → synchronous teardown.
- The snapshot image built from `test/daytona-snapshot/Dockerfile` is pushed and ACTIVE as `ao-agent-sandbox:dev` (0.21 GB compressed; push tip: build with `--provenance=false --sbom=false`, the Daytona CLI chokes on buildx attestation manifests — I'll fold that flag into the docs).

The live runs flushed out three real bugs the fake couldn't catch — all fixed in 475d285f8: Daytona's stop/delete are async (200 while state lags), and the list endpoint trails `GET /sandbox/{id}`, so a freshly-parked sandbox could read `stopping` and the old wake path waited on `started` without ever issuing the start. Park now waits for `stopped`, `ensureStarted` settles transitional states before acting (regression-tested with a state-settling fake), and destroys confirm the sandbox is gone.

Remaining for the real-agent demo (`TestLive_RealAgentSession`): a `CLAUDE_CODE_OAUTH_TOKEN` to inject — the snapshot already has Claude Code 2.1.220 baked in.

AgentWrapper:
**Exit criteria met, live.** `TestLive_RealAgentSession` PASS in 10.0s with the maintainer's Daytona key + a `CLAUDE_CODE_OAUTH_TOKEN` injected as sandbox boot env (custody model as locked — no browser flow):

```
DAYTONA-AGENT-OK

Hello! Good to meet you, I'm ready to help with whatever you're working on.
daytona@53203e96-…:~/ao/ao-smoke-realagentsession-…$
```

A real logged-in Claude Code, in a sandbox provisioned from `ao-agent-sandbox:dev`, launched and read back entirely through `ports.Runtime`/`ports.Workspace` — the trailing shell prompt is the keep-alive contract holding after agent exit. All three live suites now pass (client 6.2s, full lifecycle 16.2s, real agent 10.0s).

AgentWrapper (commented):
**Review of the Daytona runtime + workspace adapter (phase 2). Verdict: requesting changes** — nothing structural; the core adapter is careful and well-tested, but there is a teardown dead-end that leaks billed sandboxes, plus three provisioning edge paths that can silently lose or misplace session state. All are in `workspace.go`; details inline.

What I checked:
- Diffed against the merge-base with main (987f00e6d); read the full adapter, tests, the CLI `AO_API_BASE` change, runtimeselect, and the snapshot Dockerfile.
- Contract parity vs the tmux runtime (launch line, exact-match targets, chunked send-keys + delayed Enter for #2342, the #2802 pane-descendant walk) and vs gitworktree (dirty refusal, temp-index stash / cherry-pick apply sequences, stale semantics) — faithful, with one divergence flagged inline (stale `Destroy`).
- Traced how the session manager consumes the error sentinels these methods return (kill / cleanup / retire paths in `manager.go`) — that is where the stale-`Destroy` dead-end bites.
- `go build ./...`, `go vet`, and `go test` on the touched packages: green (`internal/cli` green with the inherited `AO_SESSION_ID` scrubbed, matching the PR description).

Findings, most severe first (each has an inline comment):
1. `Destroy` on a stale checkout errors forever → unkillable session and a leaked, still-billing sandbox; gitworktree succeeds in the analogous case. This is the one I would fix before the phase-3 wiring lands.
2. The Create reuse path never checks out `cfg.Branch` after a partial provision → session silently runs on the base branch.
3. The provision probe treats transport failures as not-provisioned, and the failure cleanup deletes adopted sandboxes → narrow but destructive path (a checkout with uncommitted work deleted without a stash).
4. The fetch fallback in `provisionCheckout` conflates fetch failure (auth on private repos, transient network) with branch-missing → a restore silently starts from base.
5. (minor) `Restore` maps transport errors to `ErrWorkspaceStale`, unlike the execError-gated stale checks elsewhere.

The hooks/CLI change, attach path, process-table walk, and client are solid; the live-validated async-state handling (settle-then-act in `ensureStarted`, the mid-stop re-read in `IsAlive`) is particularly good. CLI remote-base behavior is covered by a proper test.

backend/internal/adapters/runtime/daytona/workspace.go
```diff
@@ -0,0 +1,599 @@
...
+	return nil
+}
+
+func (w *Workspace) isDirty(ctx context.Context, sandboxID, path string) (bool, error) {
+	out, err := w.exec(ctx, sandboxID, "git -C "+shellQuote(path)+" status --porcelain")
+	if err != nil {
+		var execErr *execError
+		if errors.As(err, &execErr) {
+			// Not a git repo any more: stale, nothing to protect.
+			return false, fmt.Errorf("daytona workspace: dirty check %s: %w: %w", path, ports.ErrWorkspaceStale, err)
```
AgentWrapper, line 549: **Teardown dead-end: a stale checkout makes `Destroy` fail forever, leaking the sandbox.**

`isDirty` maps a stale checkout (path gone / not a repo — e.g. the agent removed the dir, or a half-failed provision whose cleanup delete also failed) to an error, and `Destroy` returns it. The session manager kill path only special-cases `ErrWorkspaceDirty` (`manager.go:873`); an `ErrWorkspaceStale`-wrapped error falls into the generic `kill: workspace:` failure, and every retry fails identically because the checkout never comes back. Result: an unkillable session and a sandbox that keeps billing until auto-archive. The cleanup path (`manager.go:2156`) skips it forever for the same reason.

gitworktree succeeds in the analogous case (unregistered path → prune → removeAll → nil), and the comment here even says stale means nothing to protect — but the code returns the error instead of proceeding. Suggest: when the dirty probe fails with a definitive `execError` (stale), fall through to `deleteAndWait` rather than returning.

backend/internal/adapters/runtime/daytona/workspace.go
```diff
@@ -0,0 +1,599 @@
...
+	}
+	return sb, nil
+}
+
+// provisionCheckout clones the repo (base branch when set) and creates or
+// checks out the session branch, returning the effective branch name. An
+// existing valid checkout is reused (idempotent retry after a partial spawn).
+func (w *Workspace) provisionCheckout(ctx context.Context, sb Sandbox, path string, cfg ports.WorkspaceConfig, remote RepoRemote) (string, error) {
+	// Already provisioned? (`git -C path rev-parse` succeeds only for a repo.)
+	if _, err := w.exec(ctx, sb.ID, "git -C "+shellQuote(path)+" rev-parse --is-inside-work-tree"); err == nil {
```
AgentWrapper: **Reuse path ignores `cfg.Branch`, so a retried provision can leave the session on the base branch.**

If a prior provision died after the clone but before the session-branch checkout (daemon crash — the delete-on-failure cleanup only runs when the failing Create call itself is still executing), this reuse path sees a valid repo and returns `currentBranch()`, i.e. the cloned base branch. The session then silently runs on the base branch and `WorkspaceInfo.Branch` records it as such — nasty to debug downstream (review/PR flows expect the session branch).

Suggest reconciling on reuse: when `cfg.Branch` is set and differs from the current branch, run the same fetch/checkout script as the fresh path.

backend/internal/adapters/runtime/daytona/workspace.go
```diff
@@ -0,0 +1,599 @@
...
+			return ports.WorkspaceInfo{}, err
+		}
+	}
+
+	path := w.workspacePath(name)
+	branch, err := w.provisionCheckout(ctx, sb, path, cfg, remote)
+	if err != nil {
+		// The sandbox without a checkout is useless; tear it down so a retried
+		// spawn does not find a half-provisioned sandbox.
+		if delErr := w.client.DeleteSandbox(context.WithoutCancel(ctx), sb.ID); delErr != nil {
```
AgentWrapper: **Two related hazards on this failure-cleanup path.**

(a) The `rev-parse --is-inside-work-tree` probe above treats *any* error — including toolbox transport failures like the 502s you observed live during state transitions — as not-provisioned. A transient failure against a healthy checkout then proceeds to `GitClone` into a non-empty dir, which fails.

(b) This cleanup then deletes the sandbox unconditionally, even when it was adopted (`found == true`) rather than created by this call.

Combined: a transient probe failure on a retried spawn deletes a sandbox holding a valid checkout — possibly with uncommitted work — without a stash. Narrow, but it is the one destructive path in the adapter. Suggest gating the provisioned-probe on a definitive `execError` (transport errors should abort inconclusively, matching the careful split everywhere else in this package), and only deleting sandboxes this Create call itself created.

backend/internal/adapters/runtime/daytona/workspace.go
```diff
@@ -0,0 +1,599 @@
...
+		// Scratch-style session: stay on the cloned branch.
+		return w.currentBranch(ctx, sb.ID, path)
+	}
+	if err := validateBranchName(branch); err != nil {
+		return "", err
+	}
+	// Prefer resuming an existing remote branch (restore of a prior session);
+	// otherwise create the session branch from the cloned base.
+	script := "cd " + shellQuote(path) + " && " +
+		"if git fetch origin " + shellQuote(branch) + " >/dev/null 2>&1; then " +
```
AgentWrapper: **Fetch failure is conflated with branch-does-not-exist.**

`git fetch origin <branch>` also fails on auth and transient network errors — and for private repos (the primary cloud case per the custody model) the clone credentials traveled in the toolbox API body, so in-sandbox `git fetch` has no credentials at all unless the snapshot carries ambient ones. In those cases a restore/re-provision silently creates a fresh branch from base instead of resuming the remote session branch: prior work is still on the remote, but the session resumes from the wrong state with no signal.

Worth separating missing-ref from failed-fetch (e.g. `git ls-remote --exit-code origin refs/heads/<branch>` distinguishes exit 2 = absent from other failures), or at minimum failing hard when credentials were provided for the clone.

backend/internal/adapters/runtime/daytona/workspace.go
```diff
@@ -0,0 +1,599 @@
...
+	}
+	if !found {
+		return w.Create(ctx, cfg)
+	}
+	if sb, err = w.ensureStarted(ctx, sb); err != nil {
+		return ports.WorkspaceInfo{}, err
+	}
+	path := w.workspacePath(name)
+	if _, err := w.exec(ctx, sb.ID, "git -C "+shellQuote(path)+" rev-parse --is-inside-work-tree"); err != nil {
+		return ports.WorkspaceInfo{}, fmt.Errorf("daytona workspace: restore %s: checkout missing: %w: %w", name, ports.ErrWorkspaceStale, err)
```
AgentWrapper: Minor inconsistency: this maps *any* exec failure — including toolbox transport errors — to `ErrWorkspaceStale`, whereas `StashUncommitted` (exit 3) and `isDirty` gate stale on a definitive `execError`. A transient toolbox hiccup during restore should surface as an inconclusive error, not report the checkout as stale.

backend/internal/adapters/runtime/daytona/workspace.go
AgentWrapper: Fixed in 6d0002458: `Destroy` now treats a stale dirty-probe (`execError` → `ErrWorkspaceStale`) as nothing-to-protect and proceeds to `deleteAndWait`, mirroring gitworktree's unregistered-path success — no more unkillable session / leaked billing sandbox. Transport errors still abort (inconclusive; the checkout might hold work). Regression: `TestWorkspaceDestroyStaleCheckoutStillDeletesSandbox` (+ a subtest pinning that transport errors do NOT delete).

backend/internal/adapters/runtime/daytona/workspace.go
AgentWrapper: Fixed in 6d0002458: the reuse path now goes through `reconcileExistingCheckout`, which compares the current branch to `cfg.Branch` and runs the same session-branch checkout as the fresh path when they differ — a provision that died between clone and checkout no longer leaves the session on base. Regression: `TestWorkspaceCreateReusedCheckoutReconcilesBranch`.

backend/internal/adapters/runtime/daytona/workspace.go
AgentWrapper: Both fixed in 6d0002458: (a) the provisioned-probe only treats a definitive `execError` as not-a-repo; transport failures return a "probe checkout" error without cloning. (b) `Create` tracks `createdHere` and the failure cleanup only deletes sandboxes this call created — adopted sandboxes are never deleted by the failure path, so the destructive window is closed. Regression: `TestWorkspaceCreateProbeTransportErrorAbortsWithoutDeleting` (adopted sandbox survives, no clone attempted) alongside the existing created-here cleanup test.

backend/internal/adapters/runtime/daytona/workspace.go
AgentWrapper: Fixed in 6d0002458 with both halves of your suggestion: `checkoutSessionBranch` decides existence via `git ls-remote --exit-code --heads` — exit 2 (definitively absent) creates from base; exit 0 fetches + checks out (a fetch failure is now a hard error); anything else fails hard as "cannot verify remote branch (network or credentials)". And the clone credentials are now seeded into the sandbox via the toolbox `POST /git/credentials` API (body-only, like the clone), so in-sandbox ls-remote/fetch — and the agent's own pushes — are authenticated for private repos. Regressions: `TestCheckoutSessionBranchSeparatesMissingFromUnverifiable` (hard-fail + credential seeding). Live `TestLive_RuntimeSessionLifecycle` re-run green with the new script (22.6s).

backend/internal/adapters/runtime/daytona/workspace.go
AgentWrapper: Fixed in 6d0002458: `Restore`'s checkout probe now gates `ErrWorkspaceStale` on a definitive `execError` and surfaces transport failures as an inconclusive "probe checkout" error, consistent with `isDirty`/`StashUncommitted`. Regression: `TestWorkspaceRestoreTransportErrorIsNotStale`.

AgentWrapper:
**Review 4803249501 addressed** in 6d0002458 (all five findings, each with a regression test), and the branch is **rebased onto main** with the one conflict resolved — `internal/cli/client.go`'s new `doJSONPathWithHeaders` (from the browser feature) now carries the `AO_API_BASE`/bearer branch inside it, with caller headers applied after the auth header.

1. Stale-checkout `Destroy` proceeds to sandbox deletion (transport errors still abort) — no more unkillable sessions/leaked sandboxes.
2. Reuse path reconciles onto the session branch via the shared checkout helper.
3. Provisioned-probe requires a definitive `execError`; failure cleanup only deletes sandboxes the call itself created.
4. `ls-remote --exit-code` separates branch-missing from unverifiable remote (hard error), and clone credentials are seeded via the toolbox `POST /git/credentials` API so in-sandbox git (incl. agent pushes) authenticates on private repos.
5. `Restore`'s stale mapping is `execError`-gated like `isDirty`/`StashUncommitted`.

Verification: `go build ./...`, vet, gofmt clean; full backend `go test ./...` green (scrubbed env); live `TestLive_RuntimeSessionLifecycle` re-run green (22.6s) against real Daytona with the reworked checkout script. All five threads resolved.

AgentWrapper:
Closing this PR — cloud work is consolidating onto the contributor PR #3426. Keeping the branch intact as a reference implementation: #3251 for the adversarially-audited org-multi-tenant auth model, #3254 for the Daytona runtime adapter + end-to-end learnings. Not deleting the branch.

Issue #3152, opened 2026-07-27
Long-term fix for #3034: replace maker-zip with ditto-based mac packaging


anmol0b:
## Context
Follow-up to #3034. #<short-term-PR-number> ships the short-term fix
(skip the `.blockmap` sidecar for macOS zips, forcing full downloads
instead of a corrupted differential update). This issue tracks the
actual root-cause fix: the macOS zip artifact itself is built in a
format Squirrel.Mac's `ditto`-based install step doesn't fully support.

## Root cause (from #3034)
`@electron-forge/maker-zip` uses a pure-JS zip writer that never
produces AppleDouble (`._*`) companion files or preserves extended
attributes/resource forks — unlike `electron-builder`, which uses
macOS's native `ditto` to build its zips. Squirrel.Mac's ShipIt install
step shells out to `ditto` to extract updates, and `ditto` expects those
companion entries to exist next to their real files. Their absence is
the likely cause of the `ditto: ... No such file or directory` errors
in #3034.

## Proposed direction
Replace `@electron-forge/maker-zip` for the darwin target with a maker
that produces a zip via the same mechanism `electron-builder` uses,
so the artifact format matches what Squirrel.Mac already knows how to
extract.

**Two possible approaches, in order of preference:**

1. **Delegate to `app-builder-lib`** (electron-builder's internals),
   mirroring the existing pattern in `makers/maker-nsis.ts`, which
   already calls `app-builder-lib`'s `buildForge` rather than
   hand-rolling NSIS invocation. Needs research: does `app-builder-lib`
   expose a usable mac zip target that internally uses `ditto`? If so,
   this is the lower-risk option — it reuses code this project already
   trusts for Windows packaging, and may make it possible to safely
   *re-enable* blockmap differential updates for mac (currently
   disabled by #<short-term-PR-number>), since electron-builder's own
   blockmap generator (`blockmap.mjs` already reaches into it) is
   designed for its own zip format.

2. **Hand-roll a `ditto`-based maker** (`execFileSync("ditto", ["-c",
   "-k", "--keepParent", ...])`), if `app-builder-lib` doesn't expose a
   suitable target. Higher risk: reimplements packaging logic instead
   of delegating to a proven library, and needs careful flag selection
   (`--keepParent`, no `--sequesterRsrc`, so AppleDouble entries land
   inline next to their real files — matching the layout in the
   reported ditto errors — rather than in a separate `__MACOSX/`
   folder).

[...]

PR #3261, opened 2026-07-29
docs: explain the macOS Gatekeeper block on the download page

### LINKED ISSUE #3260: macOS first install fails with "developer cannot be verified" when Gatekeeper's ticket lookup is blocked
Reported by a teammate on macOS. A fresh download from https://aoagents.dev/download/ fails to open with:

> "Agent Orchestrator" cannot be opened because the developer cannot be verified.
> macOS cannot verify that this app is free from malware.

Buttons offered are Move to Trash and Cancel. The user retried four times (Finder had reached `Agent Orchestrator 4`) before giving up.

## The artifact is not at fault

Confirmed against the stable arm64 build they downloaded (v0.11.0, matched by bundle size: 269.8 MB decimal, versus 279.3 MB for x64):

```
codesign -dvvv     Developer ID Application: Prateek Karnal (X9LNSP8MC2)
                   flags=0x10000(runtime), Notarization Ticket=stapled
spctl --assess --ignore-cache --no-cache   accepted, source=Notarized Developer ID
syspolicy_check distribution               App passed all pre-distribution checks
```

Also verified with a Chrome-style `com.apple.quarantine` attribute set, and with Gatekeeper caches bypassed so a prior local approval could not mask a failure. All nested code (the Go daemon at `Contents/Resources/daemon/ao`, Electron Framework, all four helper apps) is individually signed with hardened runtime. The same checks pass on v0.11.1 arm64/x64 and on v0.10.4-nightly.202607280518. @harshitsinghbhandari reproduced these independently and also confirmed the alias and versioned assets are byte-identical.

## Cause: unknown, under investigation

**Correction.** This issue originally claimed the cause was Gatekeeper performing an online notarization-ticket lookup against `api.apple-cloudkit.com` that a VPN, corporate firewall, or DNS filter had blocked. That was not supported by the evidence and has been retracted.

The claim rested on `xcrun stapler validate -v` contacting that endpoint. As pointed out in https://github.com/Untrivial-ai/agent-orchestrator/pull/3261, that is stapler's own validation path, not syspolicyd's launch check, and it performs the lookup whether or not a ticket is already stapled. Stapling exists specifically so first launch succeeds without network access. So the observation says nothing about launch behaviour, and the VPN/firewall premise was never confirmed with the affected user.

What is actually established: the published artifact is correctly signed, notarized, and stapled, and it passes policy evaluation on a machine that can open it. Why one machine refused it is not yet known.

## What would settle it

Needed from the affected machine while it is still failing:

- `sw_vers`
- `codesign -dvvv "/path/to/Agent Orchestrator.app"`
- `spctl -a -t exec -vvv --ignore-cache --no-cache "/path/to/Agent Orchestrator.app"`
- `log show --last 10m --predicate 'subsystem == "com.apple.syspolicy"'`

The syspolicy log is the decisive one: it is macOS stating its own reason for refusing the app, which beats every theory in this thread.

## Not the same as #3034 / #3152

https://github.com/Untrivial-ai/agent-orchestrator/issues/3034 and https://github.com/Untrivial-ai/agent-orchestrator/issues/3152 concern Squirrel.Mac ShipIt auto-update on existing installs. This is first install, with no ShipIt or blockmap involved. The stapled ticket is a regular file at `Contents/CodeResources`, so the AppleDouble/xattr gap described in those issues cannot strip it.

## Partial mitigation

https://github.com/Untrivial-ai/agent-orchestrator/pull/3261 adds symptom-only guidance to the download page and README: right-click and Open, plus the diagnostics above so the next person to hit this produces the data instead of working around it. It deliberately does not publish `xattr -dr com.apple.quarantine`, which would tell every visitor to strip quarantine before macOS has evaluated the app and would mask the cause.

That PR does not close this issue. Keep it open until the syspolicy log identifies the cause.

Pulkit7070:
Refs https://github.com/Untrivial-ai/agent-orchestrator/issues/3260

Does **not** close that issue. It mitigates the symptom; the cause is still unknown.

> Note when merging: the first commit message (`355ce6d39`) contains a stale `Closes #3260` line from the original version of this PR. Please drop it from the squash body, or 3260 will auto-close while the cause is still open.

## Problem

A teammate could not install from https://aoagents.dev/download/. macOS refused to open the app with "cannot be opened because the developer cannot be verified", offering only Move to Trash and Cancel. They retried four times (Finder had reached `Agent Orchestrator 4`) before giving up.

The artifact is not at fault. I downloaded the exact build they had (stable v0.11.0 arm64, matched by bundle size at 269.8 MB decimal versus 279.3 MB for x64) and it passes every check, including with a Chrome-style `com.apple.quarantine` attribute set and with Gatekeeper caches bypassed:

```
codesign -dvvv     Developer ID Application: Prateek Karnal (X9LNSP8MC2)
                   flags=0x10000(runtime), Notarization Ticket=stapled
spctl --assess --ignore-cache --no-cache   accepted, source=Notarized Developer ID
syspolicy_check distribution               App passed all pre-distribution checks
codesign --verify --deep --strict          valid on disk, satisfies its Designated Requirement
```

@harshitsinghbhandari reproduced this independently and also confirmed the alias and versioned assets are byte-identical.

**Why one machine refused it is unknown.** An earlier revision of this PR claimed Gatekeeper performs an online ticket lookup against `api.apple-cloudkit.com` that a VPN or DNS filter had blocked. That was wrong and has been retracted: the observation came from `xcrun stapler validate -v`, which is stapler's own validation path and performs that lookup whether or not a ticket is already stapled, so it says nothing about the launch check. Stapling exists so first launch works offline. Details in the review thread below.

What remains true is the part that matters for the page: the build is signed and notarized, so the dialog is not evidence the download is unsafe, and users should not be left concluding it is.

## Change

Documentation only. No packaging, signing, or CI changes.

- `frontend/src/landing/src/app/download/page.tsx`: an "If macOS blocks the app on first launch" section below the Stable/Nightly grids. Symptom-only. States that builds are Developer ID signed and notarized, gives the right-click and Open steps (noting that double-clicking offers only Move to Trash, which is what made the reporter retry instead of getting through), and otherwise asks for diagnostics.
- Second card collects what we need to find the cause: `sw_vers`, `codesign -dvvv`, `spctl -a -t exec -vvv`, and `log show --predicate 'subsystem == "com.apple.syspolicy"'`, with a link to open an issue.
- `README.md`: the same guidance in two sentences, since the README is the other main download entry point.

Deliberately not documented: `xattr -dr com.apple.quarantine`. Publishing it would tell every visitor to strip quarantine from an app before macOS has evaluated it, which is bad advice on a public page independent of this bug, and it would mask the cause. A code comment records that so it does not get added back later. An earlier revision also had a `curl` snippet that avoided the flag rather than clearing it; same objection, also removed.

## Verification

- `tsc --noEmit` on the landing tsconfig: clean.
- Rendered at `http://localhost:3112/download/` and asserted against the server HTML that `xattr`, `com.apple.quarantine`, `curl -L`, `ditto`, and the "could not reach Apple" wording are all absent, and that the right-click steps, diagnostics block, and issue link are present.
- Card classes, theme tokens, and 2-space indentation match the surrounding file. Both `<pre>` blocks are wrapped in `overflow-x-auto` so long commands scroll instead of breaking the page.

Not verified visually: the browser automation extension timed out and no local headless Chrome was available. Content and structure are confirmed by executing, but not pixels. Worth a look at the preview deploy.

## Notes

- Different failure from https://github.com/Untrivial-ai/agent-orchestrator/issues/3034 and https://github.com/Untrivial-ai/agent-orchestrator/issues/3152, which are Squirrel.Mac ShipIt auto-update on existing installs. This is first install: no ShipIt, no blockmap. The stapled ticket is a regular file at `Contents/CodeResources`, so the AppleDouble/xattr gap in those issues cannot strip it.
- Out of scope here: `AgentWrapper/agent-orchestrator` 301-redirects to `Untrivial-ai/agent-orchestrator`, so the stable download links in `constants.ts` and the README take an extra hop. They work because the redirect is followed, but they should be repointed.


harshitsinghbhandari:
Thanks @Pulkit7070. The verification half of this is solid and it matches what I got independently: `v0.11.0` and `v0.11.1` both come back `accepted / source=Notarized Developer ID` with `Notarization Ticket=stapled`, including with a Chrome-style `com.apple.quarantine` attribute set and `--ignore-cache --no-cache`. I also confirmed the alias and versioned assets are byte-identical, so the download page and the Homebrew cask serve exactly the bytes we verified. Good, careful work.

Where I would push back is the cause.

`xcrun stapler validate` hitting `api.apple-cloudkit.com` is stapler's own validation path, not Gatekeeper's launch check. On my run it printed:

```
Downloaded ticket has been stored at file:///var/folders/.../*.ticket
The validate action worked!
```

even though `codesign -dvvv` on the same bundle already reported `Notarization Ticket=stapled` locally. So that network call happens regardless of whether the ticket is stapled, which means it does not show that syspolicyd ignores a present staple at first launch. Stapling exists specifically so first launch works without the network.

The VPN / firewall / DNS-filter premise also has not been confirmed with the affected user. We have asked them for `codesign -dvvv`, `spctl --ignore-cache --no-cache`, `sw_vers`, and the `com.apple.syspolicy` log from the failing machine. That log will give syspolicyd's own stated reason, which beats every theory in this thread, mine included.

Two asks:

1. Please drop the `xattr -dr com.apple.quarantine` snippet. Putting that on the public download page tells every visitor to strip quarantine from a correctly notarized app, and it would mask whatever the real cause turns out to be.
2. Keep the right-click-and-Open guidance. That is genuinely useful for any macOS app and is worth having regardless of how this resolves.

Holding until the data comes back from the affected machine, then we can word the section around the actual cause rather than around a hypothesis. The investigation work here is appreciated and most of it will carry over.

Pulkit7070:
You are right on both counts, and the correction matters more than the wording change. Pushed in 0664c1e83.

On the cause: I used `xcrun stapler validate -v` reaching `api.apple-cloudkit.com` as evidence that syspolicyd does an online lookup at launch. That does not follow. As you say, it is stapler's own validation path and it performs that lookup whether or not a ticket is already stapled, so it says nothing about the launch check. Stapling exists precisely so first launch works offline. I asserted a mechanism I had not verified, and I have removed it from the page, the README, and the issue rather than soften it. The VPN and DNS-filter premise was mine, not the reporter's, and it has not been confirmed with them either.

What I can still stand behind is only the artifact half: the published assets are Developer ID signed, hardened, notarized, and stapled, and they pass `spctl` and `syspolicy_check` including quarantined and cache-bypassed. Why one machine refused it is unknown, and the syspolicy log you have asked for is the right thing to wait on.

On the `xattr` snippet: agreed, and I should not have put it there. Telling every visitor to strip quarantine from an app before macOS has evaluated it is bad advice on a public page regardless of this bug, and it would have hidden the cause. I also removed the `curl` snippet, which had the same effect by avoiding the flag in the first place rather than clearing it after.

The revised section is symptom-only:

- Kept the right-click and Open steps, including that double-clicking offers only Move to Trash, which is what made the reporter retry four times instead of getting through.
- Kept the statement that builds are signed and notarized, since that is measured and it is the part that stops the dialog reading as "this is malware."
- Replaced the workaround card with the diagnostics you asked the reporter for (`sw_vers`, `codesign -dvvv`, `spctl -a -t exec -vvv`, and the `com.apple.syspolicy` log) and a link to open an issue, so the next person to hit this produces the data instead of silently working around it.

A code comment records why `xattr` is deliberately absent, so it does not get helpfully added back later.

Happy to hold this until the affected machine reports back. If the log shows something that makes the right-click guidance wrong too, I would rather rewrite the section again than leave a plausible-sounding explanation on the download page.

i-trytoohard:
Closing as stale: there has been no author activity for five weeks, the branch conflicts with main, and the Gatekeeper documentation it adds is still absent from the download page. A rebased resubmission would be welcome if you're still planning it.

PR #3262, opened 2026-07-29
fix: prevent foreign PR auto-attribution

### LINKED ISSUE #3259: SCM observer auto-attributes foreign PRs to sessions by branch name with no owner check
## Problem

Reported by @aditi: While working on a session (not on any PR), the SCM observer attached another developer's PR to her session. The only reason it matched was the branch name.

This causes incorrect PR-to-session attribution and will produce clashes as the user base grows.

## Root Cause

`discoverNewPRs` in `backend/internal/observe/scm/observer.go` (lines 716–795) attributes PRs to sessions using two signals:

1. **Head-repo (fork) guard** — `candidatesForHeadRepo` narrows candidates to sessions whose push origin matches the PR's `HeadRepo`. This correctly blocks cross-fork misattribution.
2. **Branch-name prefix matching** — `matchSession` + `sessionBranchPrefixes` (lines 823–848) picks the session whose branch is an exact match OR a namespace prefix of the PR's source branch.

There is **no owner/author check**. When multiple users push to the same repo (the common case for a shared monorepo), the fork guard passes for everyone, and any PR whose source branch collides with a session's branch name gets auto-attached.

### Collision surface

The default worker branch is `ao/<namespace>/<session-id>/root`. `sessionBranchPrefixes` expands this to also match `ao/<namespace>/<session-id>/*`, which is how stacked PRs are discovered. But this means any branch under that namespace is fair game. If another developer pushes a branch that happens to match (or a custom `--branch` collides), the PR lands on the wrong session.

The `PR.Author` field (GitHub login of the PR creator) is already populated on every `SCMPRObservation` and persisted to the `PullRequest` row — but it is never consulted during attribution.

### No identity stored

`SessionRecord` has no GitHub-identity field. The daemon authenticates with a GitHub token but never resolves or stores which user the token belongs to, so there is nothing to compare `PR.Author` against.

## Impact

- **Severity:** P1 — incorrect attribution causes confusion, false lifecycle triggers (e.g., merge teardown for a PR the session doesn't own), and noisy dashboards.
- **Component:** comp/daemon (SCM observer)
- **Frequency:** Increases with team size on shared repos.

### Layer 1 — SCM identity gate (primary fix, addresses the root cause)

Resolve the authenticated GitHub user at daemon startup (one-time `GET /user` call) and store it. In `discoverNewPRs`, skip any PR whose `Author` does not match the token user.

- **Why this works:** The agent creates branches and pushes them from the daemon's authenticated account. PRs opened by other users in the same repo will have a different `Author` and be skipped.
- **Fallback:** If the identity can't be resolved (no token, or it's a bot/GitHub App), fall back to current behavior with a debug log. Users can still explicitly claim via `ao pr claim`.
- **Edge case:** A teammate opening a PR for the agent's branch — handled by `ao pr claim` as the explicit override.

### Layer 2 — Exact-match priority (defense in depth)

When multiple sessions match a PR by branch name, prefer exact matches over prefix matches. Only fall back to prefix/namespace matching when no exact match exists. This narrows the collision surface even when Layer 1 falls back.

### Layer 3 — Push-lineage verification (future, strongest)

Record the commit SHA(s) the session actually pushed. During discovery, verify the PR's `HeadSHA` is reachable from the session's known commits. This proves git-level ownership independent of branch names or authors. Higher implementation cost; consider for a follow-up.

## Files

- `backend/internal/observe/scm/observer.go` — `discoverNewPRs`, `matchSession`, `sessionBranchPrefixes`, `candidatesForHeadRepo`
- `backend/internal/domain/session.go` — `SessionRecord` / `SessionMetadata` (needs identity field)
- `backend/internal/adapters/scm/github/` — provider adapter (needs authenticated-user resolution)

Hardik180704:
## What changed

- resolve and cache the human account authenticated with the GitHub provider
- ignore foreign-authored PRs during automatic branch-based discovery
- keep the existing discovery fallback when identity is unavailable or belongs to a bot/app
- prefer true exact branch matches over namespace-prefix matches
- add regression coverage for owner filtering, case-insensitive logins, fallback behavior, caching, and exact-match priority

## Why

The SCM observer previously attributed PRs using only the head repository and branch name. In shared repositories, another developer could create a PR from a colliding branch name and have it attached to the wrong AO session, potentially causing incorrect lifecycle behavior.

Authenticated identity remains provider-scoped and in memory; it is not persisted on sessions. Explicit `ao pr claim` remains the override when another user opens a PR for a session branch.

## Verification

- `go test ./internal/observe/scm ./internal/adapters/scm/github`
- `go test -race ./internal/observe/scm`
- `go test -race ./internal/adapters/scm/github`
- `npm run lint`
- `go build ./...`
- `git diff --check`

Fixes #3259

### DIFF
--- a/backend/internal/observe/scm/observer.go
+++ b/backend/internal/observe/scm/observer.go
@@ -75,6 +75,10 @@ type credentialChecker interface {
 	SCMCredentialsAvailable(ctx context.Context) (bool, error)
 }
 
+type authenticatedIdentityProvider interface {
+	AuthenticatedIdentity(ctx context.Context) (ports.SCMIdentity, error)
+}
+
 // Config holds optional observer knobs. Zero values use production defaults.
 type Config struct {
 	// Tick is the fast PR/CI polling interval. Zero uses DefaultTickInterval.
@@ -714,6 +718,7 @@ func pendingRepoRefreshes(guards map[string]repoGuardState) map[string]bool {
 // NotModified against a known ETag are skipped, since nothing new can have
 // appeared since the last poll.
 func (o *Observer) discoverNewPRs(ctx context.Context, sessionRepos []sessionRepo, subjects map[string]*subject, guards map[string]repoGuardState, now time.Time, markRepoFailed func(ports.SCMRepo)) {
+	identity, identityKnown := o.authenticatedIdentity(ctx)
 	byRepo := map[string][]sessionRepo{}
 	repos := map[string]ports.SCMRepo{}
 	for _, sr := range sessionRepos {
@@ -739,6 +744,9 @@ func (o *Observer) discoverNewPRs(ctx context.Context, sessionRepos []sessionRep
 		}
 		for _, pr := range pulls {
 			if pr.Number <= 0 || pr.SourceBranch == "" {
+				continue
+			}
+			if identityKnown && !strings.EqualFold(strings.TrimSpace(pr.Author), identity.Login) {
 				continue
 			}
 			key := prKey(repo, pr.Number)
@@ -794,6 +802,24 @@ func (o *Observer) discoverNewPRs(ctx context.Context, sessionRepos []sessionRep
 	}
 }
 
+func (o *Observer) authenticatedIdentity(ctx context.Context) (ports.SCMIdentity, bool) {
+	provider, ok := o.provider.(authenticatedIdentityProvider)
+	if !ok {
+		return ports.SCMIdentity{}, false
+	}
+	identity, err := provider.AuthenticatedIdentity(ctx)
+	if err != nil {
+		o.logger.Debug("scm observer: authenticated identity unavailable; preserving branch-based discovery", "err", err)
+		return ports.SCMIdentity{}, false
+	}
+	identity.Login = strings.TrimSpace(identity.Login)
+	if !identity.Human || identity.Login == "" {
+		o.logger.Debug("scm observer: authenticated human identity unavailable; preserving branch-based discovery")
+		return ports.SCMIdentity{}, false
+	}
+	return identity, true
+}
+
 // matchSession picks the session that owns sourceBranch. A session owns the
 // branch when it is an exact match or a stacked descendant ("branch/..."). The
 // default worker branch is a leaf named "<namespace>/root"; for that shape the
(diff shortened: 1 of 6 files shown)


backend/internal/observe/scm/observer.go
```diff
@@ -714,6 +718,7 @@ func pendingRepoRefreshes(guards map[string]repoGuardState) map[string]bool {
 // NotModified against a known ETag are skipped, since nothing new can have
 // appeared since the last poll.
 func (o *Observer) discoverNewPRs(ctx context.Context, sessionRepos []sessionRepo, subjects map[string]*subject, guards map[string]repoGuardState, now time.Time, markRepoFailed func(ports.SCMRepo)) {
+	identity, identityKnown := o.authenticatedIdentity(ctx)
```
neversettle17-101, line 721: Can we make this identity dependency explicit from the SCM wiring boundary instead of having the observer discover it through a provider type assertion? I think the behavior is right for auto-discovery: gate on AO’s active SCM identity, and if a PR was opened/pushed outside AO with a different GitHub account, require explicit `ao pr claim` rather than branch-name auto-attribution.

The change I’d like here is to wire a lazy/cached identity resolver alongside the SCM provider and have the observer use that dependency directly. That keeps daemon startup non-blocking, preserves the current fallback when identity is unavailable or bot/app-backed, and gives us one reusable source for “the GitHub account AO is authenticated as” across observer/doctor/future PR flows.

Hardik180704: Addressed in e2d34eccf. I added a provider-neutral SCMIdentityResolver dependency and now inject it explicitly from daemon SCM wiring instead of discovering it through a provider type assertion. The GitHub provider remains lazy and caches successful identity resolution, while unavailable, empty, or bot/app-backed identities still preserve the existing branch-based fallback. I also added coverage using a resolver separate from the provider to keep this boundary explicit. Full npm run lint and focused race tests are green.

JAYATIAHUJA (approved):
LGTM. Verified the fix end to end:

- Identity resolution is lazy and cached on the provider, failures are not cached (retry next poll), and daemon startup stays non-blocking.
- The previous review feedback is addressed: `IdentityResolver` is now an explicit `ports.SCMIdentityResolver` dependency wired from `daemon/scm_wiring.go` instead of a provider type assertion.
- Author filtering is safe: `ListOpenPRs` populates `Author` from `pull.User.Login`, so known-identity polls won't silently drop the user's own PRs. Comparison is case-insensitive with trimming.
- Fallback behavior on lookup error / bot / app identity is preserved and covered by tests.
- The `matchSession` exact-match-priority fix is a genuine bug fix (namespace-root sessions could previously win over an exact branch owner via longest-prefix) and has regression coverage.

CI is fully green. Approving.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

PR #3266, opened 2026-07-29
fix(assets): tighten ao-logo viewBox to square around artwork

ikeshavvarshney:
The AO logo's viewBox was `150 80 870 760` — a rectangle with uneven padding, leaving the mark off-center wherever it lands in a square slot (favicon, app icon, README 160×160, `size-4` header img).

Computed the exact geometry bbox across all 22 paths (cubic Bézier extrema plus `translate` transforms; fills only, so no stroke bleed): x `170.86–908.09`, y `129.79–789.50` = **737.23 × 659.70**. One zero-area path at `(961.116, 398.09)` renders nothing and was excluded — it had been inflating the apparent width by ~53 units.

New viewBox: `170.86 91.03 737.23 737.23` — square on the wider axis, vertically centered. Horizontally exact-tight; ~38.8 units above and below, the minimum possible for a square since the mark is wider than tall. Square over tight-rect because every consumer slot is square, so the padding is better explicit and centered than left to each renderer to letterbox or stretch.

Path data untouched. One-line change per file; all four copies remain byte-identical to each other.

- `assets/ao-logo.svg`
- `frontend/assets/ao-logo.svg`
- `frontend/src/landing/public/ao-logo.svg`
- `frontend/src/landing/public/favicon.svg`


illegalcall (approved):

illegalcall:
@ikeshavvarshney After this releases there might be alignment issues where these images were used. You should manually verify and confirm that it doesnt cause an issue.
Take help from @codebanditssss

PR #3271, opened 2026-07-29
feat(workspace): warn when creating workspace inside ancestor Git repo

### LINKED ISSUE #3230: Explore Workspace import behavior inside ancestor Git repositories
## Summary

Follow up from #3228 / PR #3229: Project import now detects when the selected folder is a plain directory inside an ancestor Git repository and blocks before offering Git setup. Workspace import uses a different flow and should be explored for the same ancestor-repo case.

This is not confirmed as a user-facing regression yet. The goal is to decide whether Workspace should intentionally allow this, or whether it should enforce the same safety invariant as Project before auto-initializing the workspace parent.

## Current behavior to verify

Workspace import currently appears to differ from Project import:

1. Desktop Workspace selection does not run the new Project-only `projectRepositoryPreflight()` path.
2. The Workspace scan mode scans immediate child directories for repositories; it does not check whether the selected workspace parent itself is inside an ancestor Git repository.
3. Backend `Add(... AsWorkspace: true)` goes through `prepareWorkspaceProject()`.
4. If the selected workspace parent is not considered `isGitRepo(parent)`, backend can initialize that parent as the workspace root, provided it has valid direct child repos.

That may mean a folder like this can be accepted as a Workspace parent:

```text
parent-repo/              # existing ancestor git repository
  universe/               # selected as AO Workspace parent, no .git initially
    api/                  # direct child git repo
    web/                  # direct child git repo
```

## Question

Should AO allow Workspace setup to initialize `universe/.git` when `universe` is inside `parent-repo`, or should Workspace reject this upfront like Project does and ask the user to select the ancestor repo root or move the workspace folder?

## Suggested exploration

- Reproduce Workspace import with a plain workspace parent nested under an ancestor Git repo.
- Confirm whether backend creates a nested `.git` in the selected workspace parent.
- Decide the intended product behavior for Workspace vs Project.
- If the invariant should match Project, add an ancestor-root check to `validateWorkspaceParent()` and mirror it in the desktop Workspace preflight/error UI.
- Add real-git coverage for the chosen behavior.

## Related

- #3228
- PR #3229

vanshx999:
Closes #3230


vanshx999:
This adds ancestor-repo detection to the workspace creation flow. When the user picks a folder that sits inside an existing Git repo, they now see a warning in the agent sheet: "Selected folder is inside an existing Git repository at X. AO will initialize this folder as a separate repository." The workspace still proceeds it's informational, not blocking.
For single-repo imports, this warning already existed (in projectRepositoryPreflight → scanImportFolder project mode). This PR extends it to the workspace path via the same scanImportFolder workspace mode scan, which now runs ancestor detection in parallel with the child directory scan.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Standing review issues at head 670a22018e2ec782a8ab14063a8cf2a2a8450eb3:

1. **P2 — Workspace scan failures can be invisible in Sidebar and Command Palette.** In [CreateProjectFlow.tsx lines 84-99](https://github.com/Untrivial-ai/agent-orchestrator/blob/670a22018e2ec782a8ab14063a8cf2a2a8450eb3/frontend/src/renderer/components/CreateProjectFlow.tsx#L84-L99), a rejected workspace scan only sets local error state. The embedded first-run flow renders that error, but the non-embedded mode dialog does not, selectedPath is never set, and the agent sheet never opens. A removed/unreadable folder, disconnected volume, or IPC failure therefore makes the selection appear to do nothing. Please either route the failure into CreateProjectFolderDialog/render it on every entry point, or decouple the ancestor-warning probe so a diagnostic child scan cannot block selection.

2. **P2 — The required Workspace acceptance coverage is missing.** [import-folder-scan.test.ts](https://github.com/Untrivial-ai/agent-orchestrator/blob/670a22018e2ec782a8ab14063a8cf2a2a8450eb3/frontend/src/main/import-folder-scan.test.ts) only covers ancestor warnings in Project mode. [Sidebar.test.tsx lines 570-690](https://github.com/Untrivial-ai/agent-orchestrator/blob/670a22018e2ec782a8ab14063a8cf2a2a8450eb3/frontend/src/renderer/components/Sidebar.test.tsx#L570-L690) adjusts workspace scan mocks/call counts but never returns setupWarning or asserts the new behavior. Please add scanner cases for a nested plain Workspace parent, a selected repository root, and no ancestor; plus a flow test proving the warning is visible, valid children remain valid, creation submits asWorkspace, and the single-Project initializer is not called. This coverage was explicitly requested in issue #3230's scope update.

3. **P3 — Workspace selection now waits for the complete child-repository diagnostic scan.** [import-folder-scan.ts lines 249-259](https://github.com/Untrivial-ai/agent-orchestrator/blob/670a22018e2ec782a8ab14063a8cf2a2a8450eb3/frontend/src/main/import-folder-scan.ts#L249-L259) runs the ancestor probe and scans up to 200 children, but Promise.all returns only after both finish. The successful selection path consumes setupWarning and discards repos, so large or slow/network-backed folders can leave Opening active unnecessarily. A lightweight ancestor-only preflight, or separating warning and child-diagnostic scans, would avoid that latency.

I consider the first two merge-blocking. The third is a concrete workload-dependent performance regression and should either be addressed here or explicitly accepted.

vanshx999:
Thanks for the thorough review @illegalcall. Latest push 622552f addresses all three issues:

P2 (error handling): Replaced the full scanImportFolder call with a lightweight checkAncestorRepo IPC that only runs git rev-parse --show-toplevel. The call is wrapped in try/catch  if it fails for any reason, the selection proceeds without a warning.

P2 (test coverage): Added 3 scanner tests in import-folder-scan.test.ts covering nested plain workspace parent, git repo root, and no ancestor. Added a flow test in Sidebar.test.tsx that proves the warning renders, valid children remain valid, creation submits asWorkspace: true, and the single-repo initializer is not called. Also fixed comparablePath to resolve macOS symlinks (/var → /private/var).

P3 (performance): The preflight now uses the dedicated checkAncestorRepo bridge instead of scanImportFolder, avoiding the full child directory scan entirely.
All 57 tests pass, typecheck clean.

vanshx999:
@illegalcall Thanks for the e2e bridge addition. Anything else needed before merge?

illegalcall (approved):

PR #3274, opened 2026-07-29
feat: add durable idle-orchestrator re-engagement

### LINKED ISSUE #3245: Idle-orchestrator self-continuation: durable, linked re-engagement on the standing prompt
## Problem (in AO's own terms)

AO's entire orchestrator-coordination mechanism today is "a Kind=orchestrator session plus standing prompts" (verbatim, issue #2764's own problem statement), plus `ao spawn`/`ao send`, which are not durable. A worker-idle-notifies-orchestrator nudge already ships (#3038, currently being asked to be quieted for noise) but nothing analogous runs the other way: when the orchestrator itself goes idle with no new instruction, nothing re-engages it. The only existing "orchestrator gets replaced" behavior is an accidental bug (#2501, open, high-priority): a stale-activity/probe race silently kills a live, mid-turn orchestrator and spawns a fresh one 3 seconds later with zero linkage back to the original and no user notification -- i.e. the closest thing to self-continuation today is an accidental, context-destroying restart, not a designed loop.

## Evidence this costs us specifically

Our own audit found 11 orchestrators and only 3 workers on a 75%-idle host -- exactly the failure mode AO's own tracker has not yet closed, playing out at our operating scale.

## What Paperclip does (design validation, not a spec to copy wholesale)

- Heartbeat model: agents "don't run continuously -- they wake up in heartbeats, check their inbox, do a chunk of work, and stop until the next heartbeat," with wake triggered deterministically by assignment, approval resolution, or schedule. https://docs.paperclip.ing/guides/welcome/glossary/
- Delegation: "CEO monitors progress each heartbeat" and escalates unresolvable blockers to the human. https://docs.paperclip.ing/guides/org/delegation/

## Proposed design (harness-shaped, scoped to a single project/session)

When an orchestrator session's `activity_state` settles to `idle`/`waiting_input` with no pending durable notification, re-inject its own standing prompt (the same one used at session start) -- or a bounded "anything left to do?" check -- instead of doing nothing. This is the orchestrator-facing mirror of the already-shipped worker-idle nudge (#3038), but durable and explicitly linked to the original session, unlike #2501's silent, unlinked replace. File referencing #2501 and #3038 as evidence that this is the missing half of an already-half-built mechanism, not a random new ask.

## Acceptance criteria (for AO maintainers to judge)

An orchestrator session with a standing prompt and no new work re-receives that standing prompt on the next heartbeat/tick, the re-injection is logged and visible to the user, and the original session is never silently terminated/replaced in the process.

## Non-goals

- Not a request for cross-project/cross-repo priority arbitration -- that is ours (ao-company), per #2764's own stated v1 non-goals.
- Not a request to build a full org-chart/CEO-agent model -- must stay session/standing-prompt scoped, or it deviates from AO's harness vision and should not be filed as-is.

Hardik180704:
## What changed

- add durable idle-orchestrator re-engagement state
- re-engage only idle orchestrators after a bounded delay
- use exponential backoff with a hard three-attempt ceiling
- defer automation while the session has an unread human notification
- notify the user when automatic attempts are exhausted
- add `ao orchestrator done --session <id>` for an explicit durable completion outcome

The automated message is clearly marked as daemon-authored and not human authorization. Active, waiting-input, blocked, exited, terminated, worker, and completed sessions are not re-engaged.

## Why

An orchestrator can become idle while concrete assigned work remains. This gives AO a bounded and durable recovery path without treating idle as unlimited permission to continue or creating an infinite nudge loop.

Closes #3245

## Verification

- `npm run lint`
- `go test -race` for all changed backend packages
- full backend race suite; the unrelated Kilo Code timeout passed three isolated reruns
- frontend typecheck and E2E typecheck
- frontend tests: 91 files, 1,188 tests
- `npm run api`
- `npm run sqlc`
- `git diff --check`


JAYATIAHUJA (approved):
LGTM. Reviewed all 28 files; the design is sound and the boundaries are right:

**Core loop** — Durable state in its own table, 3-attempt ceiling with capped exponential backoff, progress-gated reset (post-tool-use), deferral while an unread human notification exists, and a terminal needs-input notification on exhaustion. The automated message is explicitly marked as daemon-authored and non-authorizing, which addresses the safety concern in #3245. Attempt recording is atomic (`UPDATE ... WHERE state='active' RETURNING`), so concurrent Complete/schedule interleavings can't corrupt the count.

**LCM hook placement** — Forwarding happens after `m.mu` is released (no deadlock against the store write path), and crucially the same-state early-return paths forward `post-tool-use` progress too — a steadily working orchestrator emits mostly same-state signals, so progress would otherwise never register. Transition paths pass correct before/after records.

**Storage** — Migration 0038 is the correct next number (no collision with 0037 on main), has a Down, and the partial index on (state, next_attempt_at) matches the due-scan query. Upsert semantics in the SQL exactly match the fake store used in orchestratorloop tests. gen/ is sqlc-regenerated, not hand-edited.

**API/CLI** — New route + DTO + specgen entry, and openapi.yaml/schema.ts are regenerated together (api-drift passes). CLI stays a thin HTTP client, `--session` omission returns usageError (exit 2), and the dto-drift fake is updated.

**Verified against the tree**: `sessionguard.NudgeCoordination`, `observe.StartPollLoop`, and the `post-tool-use` event naming all line up with existing code. All 10 CI checks green.

Two non-blocking observations (no change requested):
- When an attempt is deferred (unread notification or guard suppression), the row stays due and is re-evaluated every 30s tick. That's arguably the right behavior — the nudge fires as soon as the user catches up — but it means one extra read per tick per deferred session.
- If a row is completed between ListDue and RecordOrchestratorReengagementAttempt, the record query returns ErrNoRows and logs an error after the message was already sent. Benign race, log-only.

Approving.

Pulkit7070 (commented):
The durable state machine here is well thought through. The progress_since_attempt flag correctly stops an orchestrator from gaming the ceiling by toggling in and out of idle: a re-idle resets the backoff and attempt count only when real work happened in between, an exhausted row revives only on progress, and a completed one never revives. The ceiling is enforced atomically in the attempt UPDATE, the re-engagement path re-checks that the session is still an idle orchestrator before sending so a session that went active between the due query and the attempt is left alone, EnsureOrchestratorReengagement is insert-if-absent so the per-tick sweep does not reset live schedules, and the message is explicit that it is not human authorization for any pending or destructive decision. Deferring while an unread human notification exists is the right call too.

One thing worth hardening before merge: the terminal human-attention notification is best-effort and not durable. RecordOrchestratorReengagementAttempt commits the transition to exhausted, and only then does the attempt path call Notify. If Notify returns an error, or the daemon restarts between those two steps, the row is already exhausted and ListDueOrchestratorReengagements returns only active rows, so the session is never revisited and the one signal that a human needs to step in is dropped silently. That is exactly the case the notification exists for: the automated attempts are spent and nobody is told. Consider deriving the needs-attention surface from the exhausted state on read, or recording the notification as pending and retrying until delivered, so a single failed or interrupted delivery does not abandon the session quietly.

backend/internal/orchestratorloop/manager.go
```diff
@@ -0,0 +1,259 @@
...
+	next := now.Add(m.backoff(item.AttemptCount + 1))
+	updated, err := m.store.RecordOrchestratorReengagementAttempt(ctx, rec.ID, next, now, m.maxAttempts)
+	if err != nil {
+		return err
+	}
+	m.logger.Info("orchestrator re-engagement sent", "session", rec.ID, "attempt", updated.AttemptCount)
+	if updated.State == domain.OrchestratorReengagementExhausted {
+		m.logger.Warn("orchestrator re-engagement exhausted; human attention required", "session", rec.ID)
+		if m.notifications != nil {
+			return m.notifications.Notify(ctx, ports.NotificationIntent{
```
Pulkit7070: This is the non-durable step. RecordOrchestratorReengagementAttempt on line 200 has already committed the transition to exhausted, and only here do we deliver the human-attention notification. If Notify returns an error, or the daemon restarts between line 200 and this call, the row is already exhausted, and ListDueOrchestratorReengagements returns only active rows, so this session is never revisited and the notification is never retried. The result is a session whose automated attempts are spent with nobody told, which is the exact case this notification exists for. Consider deriving the needs-attention surface from the exhausted state on read, or persisting a pending-notification marker and retrying until delivery succeeds, so a single failed or interrupted Notify does not silently abandon the session.

Pulkit7070 (commented):
Marking the exhaustion-notification durability point on the exact line, per my earlier review.

backend/internal/orchestratorloop/manager.go
Hardik180704: Fixed in bd3c298c3. Exhaustion now persists an attention_notified=false marker, and each loop tick revisits exhausted rows whose terminal notification is still pending. We only mark it delivered after Notify succeeds, so notification errors or a daemon restart leave it eligible for retry. If the daemon stops after the notification row is persisted but before the marker update, the existing unread-notification dedupe makes the retry idempotent and the next pass acknowledges it. I also added manager coverage for failure -> retry -> no further delivery, plus SQLite persistence coverage for pending and acknowledged states. Full npm run lint and focused race tests are green.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/orchestratorloop/manager.go
Pulkit7070: Confirmed resolved by bd3c298c3. attempt() no longer delivers on the exhaustion transition; it just records the exhausted state, and Tick's deliverPendingAttention re-lists state='exhausted' AND attention_notified=0 and delivers via deliverAttention, which calls Notify first and marks attention_notified only on success. A failed Notify or a restart between recording exhaustion and delivery now leaves the row pending and it retries next tick, and the mark-where-not-notified guard keeps it to a single delivery. Gap closed.

PR #3275, opened 2026-07-29
fix: replace terminal tab strip session picker with a new-terminal button

### LINKED ISSUE #3208: Session picker in terminal '+' dropdown is confusing and shows cross-project sessions
## Summary

PR [#3138](https://github.com/AgentWrapper/agent-orchestrator/pull/3138) ("Feat/project session tabs", merged July 27) changed the terminal tab strip's **"+" button** from a direct "new shell" action to a `DropdownMenu`. The dropdown lists all worker sessions alongside the "Terminal" (new shell) option, allowing users to pin Session B's terminal as a tab on Session A's page — which is confusing.

**Reported by:** @prateek

## Decision (from triage thread)

**Remove the session picker from the "+" dropdown entirely.** The "+" should go back to directly creating a new terminal — no dropdown, no extra window.

- Swap the terminal icon with "+"
- No session picker popover
- Make sure the new terminal opens in the **correct worktree** (the current session's worktree)

## Context

The "+" in `CenterPane.tsx` currently opens a `DropdownMenu` (`isTabLauncherOpen`) with:
1. "Terminal" → creates a new standalone shell
2. A searchable list of all sessions (cross-project — see scope bug below) → pins them as terminal tabs

Users expect "+" to **create** a terminal, not browse sessions. The session-pinning entry point via "+" doesn't make sense.

## Scope bug (for reference if session tabs are kept elsewhere)

`availableSessions` in `SessionView.tsx` flattens **all** workspaces, so sessions from other projects appear:

```js
const availableSessions = workspaces.flatMap((workspace) =>
    workerSessions(workspace.sessions).filter((candidate) => candidate.isTerminated !== true),
);
```

## Affected files

- `frontend/src/renderer/components/CenterPane.tsx` — remove the `DropdownMenu`/session picker, restore "+" → direct new shell
- `frontend/src/renderer/components/SessionView.tsx` — remove `availableProjectSessions` prop wiring (and related `addSessionTab`/`removeSessionTab`/`selectProjectSession` if the "+" entry point was the only consumer)
- `frontend/src/renderer/stores/ui-store.ts` — `sessionTabsByOwner` may no longer be needed if the "+" was the only way to populate it

## Repro

1. Open a session in a project
2. Click the "+" button at the end of the terminal tab strip
3. Dropdown shows a session picker with sessions from all projects — should just create a new terminal

ikeshavvarshney:
Fixes #3208

## Problem

PR #3138 turned the "+" at the end of the terminal tab strip into a `DropdownMenu`
listing every worker session, so Session B's terminal could be pinned as a tab on
Session A's page. `availableSessions` in `SessionView.tsx` flattened **all**
workspaces, so sessions from other projects showed up too. Users expect "+" to
create a terminal, not browse sessions.

## Changes

**`CenterPane.tsx`** — the dropdown is gone. The button at the end of the strip is
now a terminal icon (`aria-label="New terminal"`) that calls `onNewShellTerminal`
directly. Dropped with it: `isTabLauncherOpen` / `showAllSessions` / `sessionSearch`
state, `COMPACT_SESSION_LIMIT`, the search `Input`, and the
`projectSessions` / `availableProjectSessions` / `onAddProjectSession` /
`onCloseProjectSession` / `onSelectProjectSession` / `tabOwnerSessionId` props. The
strip now renders the session's own tab plus its shells.

**`SessionView.tsx`** — new shells open with `{ projectId: session?.workspaceId, sessionId }`,
i.e. the session on screen, instead of the pinning owner's ids. The daemon's
`resolveShellTerminalWorkingDir` prefers `SessionWorkspace(sessionID)` over the
project root, so this is what puts the terminal in the right worktree. Shell-tab
filtering keys on `sessionId` for the same reason.

**Dead-code removal** — the picker was the only writer of the pinning state, so
leaving it would strand tabs pinned before the upgrade (they'd rehydrate from
localStorage with no way to remove them). Removed `sessionTabsByOwner`,
`addSessionTab`, `removeSessionTab` and the `ao.sessionTabs` persistence from
`ui-store.ts`; the `?tabOwner=` search param from both session routes and
`_shell.tsx` (the Ctrl+Shift+` handler now scopes to the route's own session).

## Verification

- **Manual, real desktop app.** Clicking the button on a live session opened a shell
  in `~/.ao/dev/data/worktrees/ao/orchestrator/ao-orchestrator` — the session
  worktree, not the project root at `E:\ao`.
- **Daemon contract, live.** `POST /api/v1/shell-terminals` with `{projectId, sessionId}`
  (what the renderer now sends) returns `workingDir` = the session worktree;
  with `{projectId}` alone it returns the project root. That difference is the fix.
- **Renderer, driven in a browser.** Exactly one "New terminal" button, `menu` and
  `menuitem` counts both 0 on click, shell tabs 0 → 1 → 2 and back on close, and a
  second session shows none of the first session's shells.
- **Unit:** typecheck clean; affected suites pass. Full suite last run at 1180 passed /
  4 failed — all in `src/main/supervisor-link.test.ts` (`listen EACCES` on unix
  sockets), pre-existing on Windows and untouched here.
- **e2e:** `shell-terminal-tabs.spec.ts` test 1 passes again (see below).

## Notes for the reviewer

- **`shell-terminal-tabs.spec.ts` was already broken by #3138.** It expects a tab-strip
  button named "New terminal", which #3138 had renamed to "Add tab". This PR restores
  that name, and I fixed the spec's session-tab selector (the old `getByTitle("Session terminal")`
  stopped matching once the label truncates, and a plain role+name match collides with
  the sidebar). It now passes.
- **One test in that file is still red and I left it alone:**
  `opens a terminal from the board, where no session view is mounted` expects a
  board-level "New terminal" button. No such button exists anywhere in the codebase —
  the board's only entry point is Ctrl+Shift+`, which is an Electron menu accelerator
  and unreachable in the browser-based e2e run. Fixing it means either restoring a
  board button or dropping the test; that's a product call, not part of this fix. The
  regression it guarded is covered by `shell-new-session-shortcut.test.tsx`.
- **Unrelated, but worth a follow-up:** `playwright.config.ts` uses
  `VITE_NO_ELECTRON=1 npm run dev:web`, which fails on Windows
  (`'VITE_NO_ELECTRON' is not recognized`). I started the dev server by hand to run e2e.
- The repo has no Prettier config or format script, so I matched the surrounding style
  by hand rather than running a formatter (a default Prettier run reflows these files
  completely).
  
---

cc @illegalcall

### DIFF
--- a/frontend/e2e/shell-terminal-tabs.spec.ts
+++ b/frontend/e2e/shell-terminal-tabs.spec.ts
@@ -12,13 +12,19 @@ test("opens, selects, and closes standalone shell terminals from the tab strip",
 	const closeButtons = page.getByRole("button", { name: /^Close terminal / });
 	const initialCount = await closeButtons.count();
 
-	// The topbar action opens a shell and makes it the active pane.
+	// The button at the end of the tab strip opens a shell and makes it the
+	// active pane. It creates a terminal directly — no menu, no session picker.
 	await page.getByRole("button", { name: "New terminal" }).click();
+	await expect(page.getByRole("menu")).toHaveCount(0);
 	await expect(closeButtons).toHaveCount(initialCount + 1);
 
-	// Selecting the session tab hands the pane back to the agent. Matched by
-	// title, not role-name: the tab's accessible name is the session's title.
-	const sessionTab = page.getByTitle("Session terminal");
+	// Selecting the session tab hands the pane back to the agent. Matched by the
+	// session's own title: the tab's accessible name is that title, and its
+	// title attribute falls back to the label once the strip truncates it.
+	// Scoped to the terminal panel: the sidebar carries the same session name.
+	const sessionTab = page
+		.getByTestId("terminal")
+		.getByRole("button", { name: "Build screenshot-ready dashboard data" });
 	await sessionTab.click();
 	await expect(sessionTab).toHaveAttribute("aria-current", "true");
 
--- a/frontend/src/renderer/components/CenterPane.test.tsx
+++ b/frontend/src/renderer/components/CenterPane.test.tsx
@@ -18,12 +18,6 @@ const worker = {
 	updatedAt: "2026-06-10T00:00:00Z",
 	prs: [],
 } satisfies WorkspaceSession;
-const secondWorker = {
-	...worker,
-	id: "sess-2",
-	title: "review the change",
-	branch: "ao/sess-2",
-} satisfies WorkspaceSession;
 
 describe("CenterPane toolbar session label", () => {
 	const makeShells = (count: number) =>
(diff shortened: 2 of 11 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #3278, opened 2026-07-29
fix: agent logos in the README, sourced from the renderer asset folder

Pulkit7070:
## What was wrong

The Supported Agents list pointed 13 of the 23 icons at `frontend/src/landing/public/docs/logos/`, a directory that contains only 5 files. Those 13 rendered as broken images: grok, droid, crush, qwen, copilot, goose, continue, devin, kimi, kiro, kilocode, vibe, pi. Five more had no icon at all: amp, agy, auggie, cline, autohand.

Fixing the paths surfaced two further problems in the icon set itself: the `copilot` icon was the wrong logo, and several marks are single-color in a way that makes them disappear against one of the two themes.

## One set of logos, not two

All 26 references (23 worker, 3 reviewer) now point at `frontend/src/renderer/assets/agents/`, the folder the board already renders from. An earlier revision of this PR added a second copy under `docs/assets/readme/logos/`; that copy is deleted, because two sets would drift apart the first time a brand mark changed.

## Icon fixes, applied in the renderer folder

The board folder had the same defects the README exposed, so the fixes land there and both surfaces pick them up:

- `copilot` was the GitHub Octocat, not Copilot. Replaced with the Copilot mark already shipped at `app-icons/copilot.svg`, which the landing hero uses.
- `codex` was a white rounded card, so on a light surface the card vanished and only the inner glyph showed. Replaced with the transparent gradient mark.
- `goose` was a pure black glyph and `kilocode` a dark opaque tile, both invisible in the app's dark theme. Replaced with the bare marks from the landing set.
- `cursor`, `opencode` and `cline` were pure white, invisible in the app's light theme and against a light README. Recolored to `#6e7681`, which reads on both grounds.
- `agy`, `auggie` and `autohand` had no asset at all. The board fell back to a lettered tile and the README showed no icon. Added, and wired into `AgentAvatar`.

`AgentAvatar` is updated for the three new marks and for the three files whose extension changed.

## Still carrying a background

`aider` (dark green), `crush` (light lavender), `droid` (black), `grok` (white) and `kiro` (purple) keep their opaque tiles: no background-free version of those marks exists in the landing app-icons, the renderer assets, or the docs site logos. `crush` and `kiro` are brand-colored tiles that read on either ground. `droid`, `grok` and `aider` each clash with one theme. Fixing them needs the official marks from upstream, which is out of scope here.

`devin`, `kimi` and `pi` are dark ink on transparent, so they are weak in the app's dark theme. Pre-existing and likewise unfixable from in-repo assets.

## Verification

- Every `src=` in README.md resolves to a file on disk. 13 did not before.
- Every `AgentAvatar` import path resolves, every file in the renderer folder is imported, and the three new `LOGOS` keys match existing backend adapters (`agy`, `auggie`, `autohand`).
- Rendered the full README locally through marked with GitHub's stylesheet in both themes: 26 logo images present, 0 unresolved image sources.
- Audited all 24 assets programmatically for opaque backgrounds and single-color fills, which is where the list above comes from.
- Not typechecked or built: the frontend dependencies are not installed in this worktree, so the `AgentAvatar` change is verified by resolving import paths rather than by compiling. Worth a CI check.
- Not compared against GitHub's own renderer; browser automation was unavailable in this environment.

## Judgment calls for the reviewer

- Recoloring `cursor`, `opencode` and `cline` to a neutral gray changes the board's appearance in dark mode, where they were previously white. It is the only way one file serves both themes, but it is a deviation from those brands' own monochrome marks.
- The README now references paths inside `frontend/src/renderer/`, which couples the docs to the app source tree. That is the trade for having a single set of icons.


JAYATIAHUJA:
Review result: no functional issues found.

Minor hygiene note: `frontend/src/renderer/assets/agents/autohand.svg` has trailing whitespace on several added lines. `git diff --check 944000c..feae0d8` flags lines 6, 9, 12, 15, 18, 21, 24, 27, 30, 35, 40, 47, 55, 62, 68, and 82. This is not a behavioral bug, but it is easy to clean up and may matter if whitespace checks are enforced.

Checked:
- README image paths all resolve at PR head.
- Deleted renderer PNGs for `copilot`, `goose`, and `kilocode` have no stale renderer/README references.
- `AgentAvatar` imports map to existing assets and provider IDs match backend adapter IDs.
- `npm run typecheck` passes in an isolated PR worktree after installing frontend dependencies.

Verdict: ready to merge after optionally cleaning the SVG trailing whitespace.

nikhilachale (approved):
Resolve the issue @JAYATIAHUJA  added, and also sync up with @codebanditssss and @Annieeeee11 . Keep the logos in one folder only.
also use copilot cli logo instead of github logo 
 I'm giving approval, but do not approve before resolving this.

PR #3284, opened 2026-07-29
feat(topbar): add an Open in editor button beside Kill

Pulkit7070:
## Summary

Adds a split `Open` control to the session topbar, immediately left of Kill,
modelled on the VS Code opener in t3 chat. The main half opens the session's
worktree in the detected editor and focuses the file the agent changed most
recently, so a session that was fixing the download button lands on that file.
The chevron offers a folder-only open and a picker for the other installed
editors.

Detection covers VS Code and its forks (Cursor, Windsurf, Trae, Kiro, Positron,
VSCodium, Insiders), Zed, Sublime Text, and the JetBrains IDEs including Android
Studio and Fleet. Each resolves from PATH, the Homebrew and /usr/local prefixes,
the JetBrains Toolbox scripts directory, and on macOS the app bundle's CLI shim,
since the daemon is not launched from a login shell. The button hides itself when
no editor is found rather than offering an action that cannot succeed.

Xcode is deliberately excluded: its `xed` shim ships with the Command Line Tools
and fails unless `xcode-select` points at Xcode.app, and Xcode opens projects
rather than arbitrary folders.

### New API

- `GET /api/v1/editors`
- `POST /api/v1/sessions/{sessionId}/open-editor`

Path resolution and the launch both happen in the daemon. Worktree paths are
deliberately absent from the session DTOs, so the renderer sends only a session
id and an optional workspace-relative path, and the response carries no absolute
path. A controller test asserts that.

Sessions with no live worktree (never spawned, or cleaned up) fall back to the
project checkout with no file focused.

### Icons

Each editor renders its own mark, tinted where the brand actually has a colour.
VS Code's blue is sampled from its own app icon; the rest are simple-icons'
published hexes. Cursor, Windsurf, Zed and the JetBrains marks are monochrome
brands, so they inherit the text colour and stay legible in both themes.

## Verification

Driven end to end in the running desktop app against real sessions, capturing the
launcher's argv to confirm what was actually opened: newest changed file on a
dirty worktree, folder alone for "Open folder only", the chosen editor rather than
the default from the picker, folder-only on a clean worktree, the project-checkout
fallback for a session whose worktree was deleted, and correct isolation across
projects. Rapid clicks produce a single launch. API abuse probes (path traversal,
absolute path, directory, missing file, unknown editor, unknown session, malformed
body, wrong method) all return the appropriate 4xx with a specific code.

Backend and frontend suites, typecheck, and golangci-lint are clean.

## Risks and follow-ups

- Bundle sub-paths are confirmed against a real install for VS Code, Cursor and
  Zed only; the other seventeen follow vendor documentation. A wrong sub-path
  degrades to "not detected", never to launching the wrong editor, and the PATH
  lookup still finds the editor whenever its shell command is installed.
- `editorResolveError` cannot distinguish "not supported by AO" from "not
  installed", so an unsupported id returns "That editor is not installed". Only
  reachable via the API, since the UI offers only detected editors.
- The empty-detection path (button hidden) is unit-tested but was not driven in
  the app, as it needs a machine with no editor installed.
- Editor choice is not persisted. A setting would be the natural follow-up.

## Screenshots

The control in the session topbar, immediately left of Kill:

Open in editor button in the session topbar

Closer in. The two halves read as one pill, with the joining edge squared and a
divider between the label and the chevron:

Open control, zoomed

The chevron menu: folder-only open, then every editor detected on the machine,
each with its own brand mark:

Open in editor dropdown

Full window for context:

AO with the Open in editor button

Rendered from this branch in a headless browser against the mock demo workspace,
so no real project or session data appears.


JAYATIAHUJA:
Findings:

High: `backend/internal/service/session/open_editor.go:152` auto-selected files bypass the symlink/workspace confinement used for explicit paths. Explicit `path` requests go through `confinedWorkspaceFile`, but the main Open button path uses `mostRecentlyChangedFile` then returns `filepath.Join(root, rel)`. A changed symlink inside the worktree can therefore be selected and handed to the editor, opening a target outside the session workspace. Fix by resolving the auto-picked `rel` through `confinedWorkspaceFile` too, and skip/reject candidates that escape.

Important: `backend/internal/service/editor/editor_test.go:131` and `backend/internal/service/editor/editor_test.go:213` add POSIX-style executable shims that fail on Windows. Running the touched backend tests on this Windows checkout fails all `TestEveryCandidateIsDetectedFromPath/*` cases and `TestOpenPassesTheFolderThenTheFiles` with "executable file not found in %PATH%". The PR claims backend suites are clean, but this new package is not portable. Use OS-specific test shims, `.cmd` launchers on Windows, or skip POSIX launcher assertions on Windows.

Verification run:
- `go test ./internal/service/session ./internal/httpd/controllers ./internal/httpd/apispec/...` passed.
- `go test ./internal/service/editor` failed as above.
- `npm run typecheck` passed.
- `npm run test -- TopbarOpenEditorButton.test.tsx` passed.
- `git diff --check 6b6190d..4fe2939` passed.

Verdict: not ready to merge until the confinement gap is fixed.

backend/internal/service/editor/editor.go
```diff
@@ -0,0 +1,227 @@
...
+// The child is started and released: these launchers hand off to a running
+// instance and exit on their own, and blocking the HTTP handler on an editor
+// cold start would stall the request. Only spawn failures (missing or
+// non-executable launcher) surface as errors.
+func Open(ed Editor, dir string, files ...string) error {
+	if ed.Bin == "" {
+		return ErrUnknownEditor
+	}
+	args := append([]string{dir}, files...)
+	cmd := aoprocess.Command(ed.Bin, args...)
```
nikhilachale: ### 1. Windows launch failure — `editor.go:216`

> **[P1] Handle Windows batch-based editor launchers**
>
> On Windows, `exec.LookPath("code")` resolves VS Code’s `code.cmd`, but `aoprocess.Command` ultimately passes that batch file directly to `CreateProcess`. Windows cannot execute `.cmd` or `.bat` files this way; they must be invoked through `cmd.exe /c` or replaced with a resolved native executable.
>
> This means VS Code is detected and the Open button is displayed, but clicking it returns `EDITOR_LAUNCH_FAILED` on a supported platform. Please add a Windows-specific launcher path and a native Windows test that executes a `.cmd` fixture.

backend/internal/service/session/open_editor.go
```diff
@@ -0,0 +1,207 @@
...
+	// Falling back to the project checkout means there is no session work to
+	// point at, so the editor opens with no files (as requested).
+	if scope != "workspace" {
+		return "", "", nil
+	}
+	rel := mostRecentlyChangedFile(ctx, root)
+	if rel == "" {
+		return "", "", nil
+	}
+	return rel, filepath.Join(root, filepath.FromSlash(rel)), nil
```
nikhilachale: ### 2. Symlink confinement — `open_editor.go:152`

> **[P2] Apply workspace confinement to the automatically selected file**
>
> Explicit paths are validated with `confinedWorkspaceFile`, but the automatically selected file is returned using only `filepath.Join`. If the newest changed entry is a symlink pointing outside the worktree, the editor will follow it and open a file outside the validated workspace.
>
> Please resolve the automatic target through `confinedWorkspaceFile` as well, skipping it or returning an error when it escapes the workspace. A regression test using an untracked symlink to an external file would cover this boundary.

backend/internal/service/editor/editor.go
```diff
@@ -0,0 +1,227 @@
...
+// extraPathDirs are searched after PATH. The daemon is launched by the desktop
+// supervisor rather than a login shell, so it can miss the Homebrew and
+// /usr/local prefixes where these shims usually live. The JetBrains Toolbox
+// scripts directory is the only place its generated launchers exist unless the
+// user has added it to their own PATH.
+var extraPathDirs = []string{
+	"/opt/homebrew/bin",
+	"/usr/local/bin",
+	"/usr/bin",
+	os.ExpandEnv("$HOME/Library/Application Support/JetBrains/Toolbox/scripts"),
```
nikhilachale: ### 3. Repository app-data boundary — `editor.go:135`

> **[P2] Avoid probing OS-default application-data directories**
>
> These fallbacks make the daemon read from `~/Library/Application Support` and `~/.local/share` on every editor detection pass. That conflicts with the repository’s hard rule that AO must not read from or write to OS-default app-data locations.
>
> Please rely on PATH and permitted application-bundle locations, or establish and document an explicit architectural exception before adding these probes.

frontend/src/renderer/hooks/useEditors.ts
```diff
@@ -0,0 +1,69 @@
...
+	/** Workspace-relative file to focus, or "." to open the folder with no files. */
+	path?: string;
+};
+
+// Open a session's workspace in an external editor. The daemon owns both the
+// path resolution and the launch: the renderer never sees a filesystem path.
+export function useOpenInEditor() {
+	return useMutation({
+		mutationFn: async ({ sessionId, projectId, editorId, path }: OpenInEditorInput) => {
+			void captureRendererEvent("ao.renderer.open_in_editor_requested", {
```
nikhilachale, line 42: ### 4. Telemetry sanitization — `useEditors.ts:42`

> **[P3] Allowlist the new editor telemetry properties**
>
> The three `open_in_editor_*` events are not handled by `sanitizeRendererProperties`, so every property supplied here is silently discarded. The events reach PostHog with only the global context; `project_id`, `editor_id`, `target`, `scope`, and `focused_file` are lost.
>
> Please add sanitizer cases that hash `project_id` and allow only the expected enum/boolean values, along with tests covering all three events.

nikhilachale (dismissed):
### Overall review

> The implementation and API shape look solid overall, and the focused backend, API, frontend, and type-check suites pass. I’m requesting changes because the current launcher is non-functional for a standard Windows VS Code installation, and the automatic file-selection path bypasses the confinement applied to explicit paths. The app-data and telemetry findings should also be addressed before merge.

Pulkit7070:
Addressed all four findings, pushed in 7121bf3d5 and 761cd4ee0 (branch was also rebased onto main to clear the merge conflict).

1. **Windows launch failure** (`editor.go:216`): added `launch_windows.go`/`launch_unix.go`. On Windows, a `.cmd`/`.bat` launcher (VS Code's `code.cmd` and every fork's equivalent) now runs through `cmd.exe /d /s /c`, mirroring the existing `previewserver` batch-shim pattern rather than handing the batch file straight to `CreateProcess`. Added `launch_windows_test.go` with a native `.cmd` fixture test, and fixed the two POSIX-only test fixtures (`writeShim`, the `TestOpenPassesTheFolderThenTheFiles` launcher script) that were failing on Windows for the same underlying reason: Go's Windows `FileMode` only sets the executable bit for a handful of recognized extensions, so a bare extensionless fixture file was never detected as executable there. `native (windows-latest)` CI is green on this PR, including the new test.

2. **Symlink confinement** (`open_editor.go:152`): the auto-picked "most recently changed" file now goes through `confinedWorkspaceFile`, same as explicit paths. An escape degrades to "nothing changed" (folder still opens) rather than failing the whole request. Added `TestEditorTargetFileAutoPickSkipsASymlinkEscape`, verified it fails without the fix and passes with it.

3. **App-data boundary** (`editor.go:135`): removed the two JetBrains Toolbox fallback paths under `~/Library/Application Support` and `~/.local/share`. A Toolbox-installed IDE is now only detected once its scripts directory is on PATH.

4. **Telemetry sanitization** (`useEditors.ts:42`): added `sanitizeRendererProperties` cases for all three `open_in_editor_*` events — hashed `project_id`, and a fixed-vocabulary allowlist for `editor_id`/`target`/`scope`/`focused_file` (mirroring the daemon's own candidate id list). Added unit tests for all three events, including a case with a bogus `editor_id` to confirm it's dropped.

Full backend suite, `go vet`/build for both `GOOS=linux` and `GOOS=windows`, golangci-lint, and the full frontend suite all pass locally in addition to CI.

illegalcall:
Implemented and pushed in [`728400ca8`](https://github.com/Untrivial-ai/agent-orchestrator/commit/728400ca8ddfb76ad314eb5ffb09e6c73d0b20bb).

What changed:
- moved native editor/file-manager/terminal launch ownership into Electron main behind a narrow, loopback-only workspace lookup
- removed workspace fallback and most-recent-file guessing; handoff now opens only the live session workspace root
- added persisted preferred-editor selection (Cursor by default), visible no-editor/missing-workspace guidance, and responsive topbar metadata
- added verified launcher failure handling plus Windows batch-shim support
- updated telemetry, API artifacts, translations, and public failure-path coverage

Local verification:
- `env GOMAXPROCS=2 GOFLAGS=-p=2 npm run lint` — passed, 0 issues
- `go test ./...` — passed
- frontend suite — 2,694 passed, 1 skipped
- frontend and e2e typechecks — passed
- production Electron packaging — passed
- real macOS Electron e2e — passed against an isolated daemon, real session/worktree, and native Cursor launch; `preferredEditorId: cursor` persisted only after launch success and Cursor was frontmost

### E2E screenshots

Real-session target menu with Finder, Terminal, Cursor, and VS Code:

Real AO session target menu

Cursor opened the exact generated worktree (`pr3284e2e-3`):

Cursor opened the generated AO worktree

illegalcall (approved):

codebanditssss (approved):

PR #3287, opened 2026-07-29
fix(landing): live section anchors, clean hash navigation, themed scrollbar

### LINKED ISSUE #3269: Landing page: dead anchors, stuck hash, unstyled scrollbar, and placeholder testimonials
## Summary

Four defects found on the marketing site during a pass over the landing page and shared UI package. Two are P2 (broken navigation, credibility), two are P3 (interaction polish, visual polish). All four are confirmed in code with file/line references below.

| # | Issue | Severity | Effort |
|---|-------|----------|--------|
| 1 | Footer "Features" / "Agents" are dead anchors | P2 | Small — add IDs |
| 2 | `#see-it` hash persists in URL after scroll | P3 | Small — router fix |
| 3 | Scrollbar unstyled, clashes with dark theme | P3 | Trivial — CSS |

---

## 1. Footer "Features" and "Agents" are dead anchors — P2

Both footer links point at anchors that do not exist on the page.

| Link | Points to | Reality |
|------|-----------|---------|
| Features | `/#features` (`Footer.tsx:35`) | `FeaturesSection.tsx:29` has no `id` on its `<section>` — only `className` |
| Agents | `/#agents` (`Footer.tsx:36`) | No `id="agents"` exists anywhere on the page — zero matches |

`NAV_ITEMS` in `packages/shared/src/constants.ts:48` also uses `/#features`, so the same broken anchor is live in the nav.

**Proposed fix**
- Add `id="features"` to the `<section>` in `FeaturesSection`.
- For "Agents": there is no agents section on the landing page. Either remove the footer link, or give an existing section an `id="agents"` (`HarnessCoverageDemo:251` already mentions Agents and is a reasonable target).
- Keep `NAV_ITEMS` consistent with whatever anchors end up existing.

**Why P2:** broken navigation, not cosmetic — clicking the link does nothing.

---

## 2. `#see-it` hash stays in the URL after scrolling — P3

The `see-it` section does exist (`VideoSection.tsx:29`). The problem is Next.js App Router default behavior: clicking `/#see-it` sets the hash and scrolls, but the hash is never cleared when the user scrolls away. Clicking "Demo" a second time does nothing, because the browser considers the target already current.

**Repro**
1. Click "Demo" in the nav → scrolls to the video section, URL becomes `/#see-it`.
2. Scroll back to the top.
3. Click "Demo" again → nothing happens.

**Proposed fix** (any one of):
- Add an effect that strips the hash once the scroll settles (e.g. a `usePathname` watcher or scroll-end handler).
- Replace the raw anchor with a click handler that calls `router.push('/')` and then `scrollIntoView` manually.
- `scroll-mt-*` on the section improves the landing offset but does **not** fix the hash behavior on its own.

**Why P3:** first click works; only repeated clicks are affected. No data loss or breakage.

---

## 3. Scrollbar is unstyled and clashes with the dark theme — P3

No scrollbar styling exists anywhere in the codebase:
- `globals.css` imports `packages/ui/src/globals.css`, which contains no `::-webkit-scrollbar` rules.
- The only related utility is `scrollbar-hide`, used to hide scrollbars inside internal demos.
- The main page and docs therefore render the browser default scrollbar, which reads light against the dark site.

Image

**Proposed fix** — add themed scrollbar rules to `packages/ui/src/globals.css`:

```css
::-webkit-scrollbar { width: 10px; height: 10px; }
::-webkit-scrollbar-track { background: transparent; }
::-webkit-scrollbar-thumb {
  background: var(--muted-border); /* or appropriate theme token */
  border-radius: 6px;
  border: 2px solid transparent;
  background-clip: content-box;
}
::-webkit-scrollbar-thumb:hover {
  background: var(--border);
  background-clip: content-box;
}

/* Firefox */
* {
  scrollbar-width: thin;
  scrollbar-color: var(--muted-border) transparent;
}
```

Should use theme tokens so it works in both light and dark modes.

**Why P3:** pure polish, but the default scrollbar makes an otherwise premium-looking dark site feel unfinished.

ikeshavvarshney:
Fixes #3269 .

## What was broken

**1. Dead footer anchors (P2).** `Footer.tsx` linked `/#features` and `/#agents`.
`FeaturesSection` rendered no `id`, and no `#agents` target existed anywhere on the
page — both clicks did nothing.

**2. Stuck `#see-it` hash (P3).** Clicking "Demo" set the hash and scrolled, but the
hash then stayed in the URL forever. Scroll away, click "Demo" again → nothing,
because the browser treats the target as already current.

**3. Scrollbar (P3).** The issue reported the scrollbar as unstyled. The actual cause
is the opposite: `site-theme/tokens.css` was *explicitly hiding* it:

```css
html, body { -ms-overflow-style: none; scrollbar-width: none; }
html::-webkit-scrollbar, body::-webkit-scrollbar { display: none; }
```

So the page had no scroll affordance at all.

## What changed

**Anchors.** `id="features"` on `FeaturesSection`; `id="agents"` on `TrustedBySection`
("Use the agents you already trust" — the actual agents content on the page).

**Hash navigation.** New `HashLink` component. Same-page `/#` clicks call
`scrollIntoView()` directly and never write the hash, so the link behaves identically
on every click. Modified clicks (ctrl/meta/shift/middle), cross-page navigation and
unresolvable targets all fall through to normal `Link` behaviour, so it still degrades
to a real navigation without JS. Used by Footer, DesktopNav and MobileNav — every
`/#` link on the site now goes through one path.

**Header offset, without hardcoding.** Anchors were landing under the fixed 56px header.
Instead of per-section `scroll-mt-*` (which has to be remembered for every new section
and silently rots when the header height changes), a single `--header-height` token now:

- sizes the header itself (`h-(--header-height)`),
- drives `scroll-padding-top` on `<html>`, which offsets both `#hash` landings and `scrollIntoView()`,
- feeds the mobile nav's `top` / `max-height`, which previously duplicated `3.5rem`.

There is no pixel math anywhere in `HashLink` — `scrollIntoView()` is called with no
arguments and the CSS supplies both the offset and the smoothness (with a
`prefers-reduced-motion` opt-out).

**Scrollbar.** The hide-hack is replaced with themed scrollbars declared through both
standard properties (`scrollbar-width` / `scrollbar-color` — Firefox, Chrome 121+) and
`::-webkit-scrollbar*` pseudo-elements (Chrome, Edge, Safari), plus `color-scheme: dark`
so Safari/iOS overlay scrollbars and form controls render dark too. Colours come from new
`--scrollbar-thumb` / `--scrollbar-thumb-hover` tokens alongside the rest of the theme.
The `scrollbar-hide` utility used inside the demos still wins — it sits in the utilities
layer and outranks the bare `*` / `::-webkit-*` base selectors.

image

## Not included

Item 4 (placeholder testimonials) is named in the issue title but has no section in
the body. It is real — `WallOfLoveSection/constants.ts` has three fabricated entries
(`maria-garcia`, `luke-tanaka`, `rachel-kim`) with `url: "#"` and no avatar, rendered as
`<a href="#">` so clicking them jumps to the top of the page. Deliberately left out of
this PR; worth its own issue.

Two other corrections to the issue's analysis, for the record:

- `NAV_ITEMS` in `packages/shared/src/constants.ts` has zero importers — the nav is driven by `PRODUCT_LINKS` / `RESOURCE_LINKS` in `Header/constants.ts`, which never had a Features link. The broken anchor was footer-only. `NAV_ITEMS` is left as-is (dead, but both of its `/#` hrefs now resolve).
- `packages/ui/src/globals.css` is a one-line stub, not the theme entry point. The scrollbar rules belong in `site-theme/tokens.css`.

## Verification

21/21 assertions pass under Playwright/Chromium against `next dev`, plus
`tsc --noEmit` clean:

- `#features`, `#agents`, `#see-it` each resolve to exactly one element.
- Footer Features/Agents and nav Demo all land at `top=56px` — flush under the header.
- Demo click #1 → `scrollY=4579`, URL stays `/` with no hash. Scroll to top, click #2 → `scrollY=4579` again. The reported bug does not reproduce.
- Cross-page `/changelog` → `/#features` lands correctly, and re-clicking from home still scrolls even with the stale hash in the URL.
- Mobile (390×844): nav closes, `body.style.overflow` is restored before the scroll runs, scroll succeeds, no hash.
- Scrollbar measured in headed Chromium: document gutter `10px` (was fully hidden), `.scrollbar-hide` element gutter `0px`.

Firefox and WebKit were not executed — those Playwright binaries aren't installed
locally. Both engines are covered by the declared CSS, but worth a manual look before
merge.

---

cc @codebanditssss


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/landing/src/app/components/HashLink/HashLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * sits in the address bar forever — scroll away, click the same link again, and
+ * the browser considers the target current and does nothing. Here a same-page
+ * click is handled directly: scroll the element into view and let the URL stay
+ * clean, so the link works identically on every click.
+ *
+ * Cross-page navigation (`/docs` -> `/#see-it`) falls through to normal `Link`
+ * behaviour, and so does anything without a resolvable target — the anchor still
+ * degrades to a real navigation with JS off.
+ */
+export function HashLink({ href, onClick, ...props }: LinkProps) {
```
illegalcall, line 22: @ikeshavvarshney I am not totally opposed to this, but can you please investigate if a solution already exists for this? It will cover edge cases. LMK what you think

Annieeeee11:
I don't think we need to make any changes to the Coverage component since Jayati is going to redesign it completely. So I don't think it's worth updating it right now.

ikeshavvarshney:
@Annieeeee11 @illegalcall  I discussed these fixes with khushi 3 days ago, she agreed, that's when I opened this PR after discussion. Just waiting for a review from then.
It's ok, if its not important now as it was earlier.

Annieeeee11:
I dont think so this pr is valid.
@codebanditssss  I guess we should close this.

Rishet11:
Closing this PR because this implementation is no longer the path we are using. The remaining landing-page work is tracked in #4771.

ikeshavvarshney:
@Rishet11 Bro, i guess you just recreated the same issue i created and fixed in this PR.

PR #3289, opened 2026-07-29
fix false-positive claude-code auth detection, document ao daemon

frankstupak:
Found while running v0.11.0 headless on a Linux server.

`claudeLocalAuthStatus` treats a non-empty `userID` in `~/.claude.json` as proof the CLI is authorized. That field sticks around after logout (and exists on machines that never finished login), so a logged-out box shows up as authorized:

```
$ claude auth status
{ "loggedIn": false, "authMethod": "none", "apiProvider": "firstParty" }
$ ao agent ls
claude-code  Claude Code  installed  authorized
```

You then spawn a worker and it dies on its first turn.

The `claude auth status` probe is already the right source of truth; the problem is only the config-file fallback. This change keeps `hasAvailableSubscription` and a populated `oauthAccount` as valid evidence, but a bare `userID` now returns Unknown instead of Authorized, so the probe decides. Added tests for the stale-userID cases plus a fake-binary harness for the probe path.

Second commit fixes docs/cli/README.md, which still says `ao start` starts the daemon. In 0.11 it fetches and opens the desktop app; the headless entrypoint is `ao daemon`, which wasn't documented anywhere. Table updated to match.

`go build ./...` and the claudecode package tests (incl. -race) are green. Full `go test ./...` has 5 pre-existing failures in internal/cli on my machine (daemon test harness 404s); they fail identically on a clean checkout so I left them alone.


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

frankstupak:
Bumping this - it's been open since late July with no review yet. @illegalcall @pulkit7070 any chance one of you can take a look? Happy to split into smaller PRs if that helps review.

Rishet11:
@frankstupak
Closing this PR as stale after 30+ days without author activity and no review movement after your last update.

Issue #3288, opened 2026-07-29
macOS release hardening: e2e update test, verification gate, feed guard, single publisher, and dmg distribution


harshitsinghbhandari:
## Summary

Two separate mac failures over 28-29 Jul burned two days, partly from misdiagnosis:

- A user could not install from the download page ("developer cannot be verified"). The artifact itself was fine. The false trail came from someone running `codesign --verify` on a copy extracted with plain `unzip`, which leaves stray `._*` files inside the bundle and breaks the seal. That produced a convincing but wrong "the signing pipeline corrupts the seal" conclusion, which paused all releases and nearly shipped an `xattr -cr` patch.
- Separately, a user hit the #3034 ShipIt failure doing an in-app update. Differential mac updates have never worked reliably. Mitigated by deleting all darwin `.blockmap` sidecars from the last 15 releases and by PR #3151, which makes full-download-only permanent going forward.

Two corrections to the record, since the wrong lesson is worse than no lesson:

- **Why the `xattr -cr` patch was wrong.** Not because it would have broken a working seal; it would not have. Extended attributes are not part of an Electron bundle's seal (signatures live in the Mach-O binaries and `_CodeSignature/CodeResources`, and the staple ticket is a regular file), and `xattr -cr` is Apple's own documented remedy for xattr detritus (QA1940). The actual reason it was the wrong patch: a client-side `xattr -cr` strips the quarantine flag, which bypasses Gatekeeper and masks whatever the real bug is. Not shipping it was correct, for that reason.
- **The `unzip` breakage is real but artifact-specific.** It holds for our zips because they embed AppleDouble entries per-directory (ditto without `--sequesterRsrc`). A `--sequesterRsrc` zip puts them in a sibling `__MACOSX` directory and can survive `unzip` extraction. So "unzip always breaks the seal" is true of our artifacts, not universally.

This is now a full macOS release-hardening plan, not just the dmg migration. Each workstream below is independently pickupable as its own PR. References: #3034, #3151, #3152. The dmg workstream's full decision record lives at #3267 and is folded in here as workstream 6.

All electron-updater claims below were checked against the published npm tarball of the pinned version (`electron-updater@6.8.9`), not against monorepo `master`, since line positions there drift. All `codesign`/`spctl` claims were checked empirically on macOS 27.0.

## Workstream priority order

0. End-to-end update test, highest priority in the whole plan.
1. Post-staple verification gate, highest priority after the update test.
[...]

Issue #3267, opened 2026-07-29
Plan: move macOS first-install distribution to .dmg, keep zip for auto-update (supersedes #3152)


harshitsinghbhandari:
## Summary

Move macOS distribution to a `.dmg` for first install, while keeping the existing `.zip` publishing for electron-updater's auto-update path. This supersedes #3152. References: #3034, #3151, #3152.

Priorities, in order:
1. First install should have no extraction step, so the class of failure where a third-party unzip tool mishandles AppleDouble entries and breaks the code signature seal cannot happen at all.
2. Electron auto-update keeps working. Hard requirement, no regression.
3. Mac differential updates stay permanently disabled (full download only). Document why, and what evidence would be needed to ever revisit that.

## Why this supersedes #3152, not just extends it

#3152 proposes replacing `@electron-forge/maker-zip` with ditto-based packaging in `forge.config.ts`, on the premise that matching electron-builder's own zip format (via ditto) is what makes mac auto-update reliable. That premise does not hold, on two independent lines of evidence:

- Pulkit7070's own investigation on #3152 found that `app-builder-lib` 26.15.3 does not use `ditto` for its mac zips (it uses `zip -q -r -y`), that electron-builder's real zips contain zero AppleDouble (`._*`) entries, and that adding them "changes nothing measurable." Their read of #3034 is that the actual corruption came from a blockmap differential-download bug, not from zip-format mismatch, and that #3151 (already merged, forces full downloads on mac) already fixes it pending one real auto-update cycle to confirm.
- Reading `app-builder-lib`'s blockmap generator directly (`packages/app-builder-lib/src/targets/blockmap/blockmap.ts`, tag `electron-builder@26.15.3`) confirms this independently: it chunks the file as an opaque byte stream via Rabin fingerprinting. It has no zip-structure awareness at all, so AppleDouble entries were never relevant to blockmap correctness.
[...]

PR #3294, opened 2026-07-29
feat(macos): release hardening (update e2e, verify script, dmg maker, single-publisher rule)

harshitsinghbhandari:
Implements the public-repo workstreams of #3288 (macOS release hardening). Workstream 1 and the workstream 3 publish guard live in the private signing/release pipeline and are not touched here.

One PR by request, kept reviewable as one commit per workstream:

| Commit | Workstream |
| --- | --- |
| `feat(updater): add env-gated e2e sentinel and macOS update test harness` | 0 |
| `feat(scripts): add shared macOS artifact verification script` | 2 |
| `docs(scripts): correct blockmap sidecar model for linux and mac` | 3 (public part) |
| `docs(release): add the single-publisher rule and shared verify step` | 4 |
| `feat(macos): add dmg first-install maker with explicit sign, notarize, staple` | 6 |

## Workstream 0: end-to-end update test

- `frontend/src/main/auto-updater.ts`: an `AO_E2E_UPDATE_SENTINEL` sentinel that writes a marker file once an update is actually staged. It is a complete no-op when the env var is unset, which is always the case in a user-facing build.
- `frontend/scripts/e2e-mac-update.mjs`: the harness. Installs nothing itself; it takes an already-installed N-1 bundle, points it at the real feed, and proves N downloads, stages, installs on quit, and that the relaunched app reports N and is alive.
- `frontend/scripts/e2e-mac-update.test.mjs`: unit tests for the flag contract.
- `.github/workflows/mac-update-e2e.yml`: the macOS CI job.

Three things here are counter-intuitive and are per #3288's verified facts, so flagging them for review:

1. **The sentinel keys off the NATIVE `require("electron").autoUpdater` event, not electron-updater's.** In `MacUpdater.updateDownloaded()`, `dispatchUpdateDownloaded(event)` fires first and only then does `nativeUpdater.checkForUpdates()` kick Squirrel into fetching from the local proxy. Keying on electron-updater's `update-downloaded` means quitting before anything is staged, so nothing installs and the test flaps.
2. **A plain macOS quit installs but does NOT relaunch.** The harness quits the app to trigger the install, then relaunches it itself with `open -a`.
3. **The ShipIt swap has no completion signal**, so the wait is a bounded poll on the installed bundle's `Info.plist` `CFBundleShortVersionString`, never a fixed sleep. Liveness reuses the `running.json` plus loopback `/healthz` pattern from `backend/internal/cli/e2e_test.go`.

The workflow is `workflow_dispatch` plus `workflow_call` only, and is deliberately not wired into any release workflow. Those are paused, and #3288 asks for real per-run cost data before deciding whether nightly-to-nightly coverage is worth macOS runner minutes. The job prints its wall clock to the step summary so that question can be settled with numbers. The header comment shows the two-line snippet to wire it as a stable-promotion gate when that decision is made.

## Workstream 2: shared verify script

`frontend/scripts/verify-mac-artifact.sh <path-to-.zip-or-.app>`, plus a smoke test. Extracts with `ditto -x -k` (never `unzip`), then runs the three checks:

```
codesign --verify --deep --strict
spctl -a -vv -t exec
xcrun stapler validate
```

`-vv` on spctl is mandatory: without it an accepted artifact prints nothing at all and merely exits 0, so any grep for `Notarized Developer ID` can never match. `spctl` alone does not prove stapling either (it can satisfy the check through Apple's online lookup), which is why `stapler validate` is the gate. Exit codes follow the repo convention: 2 for usage errors, 1 for a failed check.

It accepts a `.app` directly as well as a `.zip`, so the update-e2e job can verify what ShipIt actually swapped onto disk with the same script. The private pipeline's post-staple gate (workstream 1) should shell out to this rather than reimplementing the checks.

The workflow 1 gate is the meaningful verification of the core logic; see the honest-coverage note below.

## Workstream 3, public part: blockmap comment drift

`feed.mjs` and `blockmap.mjs` both claimed that omitting `blockMapSize` "forces sidecar differential" on win and linux. True for win, false for linux: `AppImageUpdater` never fetches a sidecar. It uses `FileWithEmbeddedBlockMapDifferentialDownloader`, which reads the blockmap from the AppImage's own tail and hard-requires `blockMapSize`, which `feed.mjs` never writes. So linux differential updates are already off, and every linux `.blockmap` we publish is read by nobody. Both comments now state the real per-platform model.

Also corrected the mac rationale in `hashFile`, which still asserted the AppleDouble zip-format theory as the likely corruption source. #3267 decision 4 disproved that with a production repro against a correctly ditto-built zip carrying its full AppleDouble set. Leaving a known-false rationale in place is the exact failure mode this workstream exists to fix. This is comment-only; no behavior change.

**On dropping linux sidecar generation: kept, deliberately.** Two reasons, both documented in the code. The release publish guard lives in the separate private pipeline and currently asserts on the hardcoded linux sidecar filenames, so deleting the files here reds that guard out of band while another worker is actively rewriting it. And #3288 itself flags this as a behavior change deserving its own decision rather than something to fold in. The dead-weight status is now written down at the generation site so the eventual removal is a two-line change made together with the guard.

## Workstream 4: single publisher, documented

New "Hard rule: exactly one publisher" section at the top of `frontend/docs/desktop-release.md`, with the reason stated plainly: between 22 and 24 Jul two publishers emitted divergent macOS zip formats on alternating days, which made the 28-29 Jul incident evidence unreadable. It also separates approving a release (any approver) from triggering one (conductor only), and points people at the fork dev loop for test builds.

Cross-referenced from `AGENTS.md`'s Distribution section, along with two adjacent rules that kept getting re-derived wrong: use `verify-mac-artifact.sh` rather than hand-typed checks, and macOS ships both a zip and a dmg with the zip publishing forever.

While in `desktop-release.md`, corrected the release verification section: it still claimed every installer ships a `.blockmap` sidecar, which stopped being true for macOS at #3151.

## Workstream 6: macOS dmg distribution

`frontend/makers/maker-dmg.ts` bridges `app-builder-lib`'s `buildForge` with a `mac: ["dmg:<arch>"]` target, mirroring `maker-nsis.ts` closely. Per #3267 decision 1, Forge's own `maker-dmg` (electron-installer-dmg to appdmg) has no notarize or staple code and neither does app-builder-lib's dmg target, so no maker solves signing for us; given that, consistency with the existing bridge wins over adding a second packaging engine for one platform.

- **Strictly additive.** The darwin `maker-zip` entry is untouched, and now carries a comment saying it can never be removed: `MacUpdater.doDownloadUpdate` calls `findFile(files, "zip", ["pkg", "dmg"])` and throws `ERR_UPDATER_ZIP_FILE_NOT_FOUND` when absent. A dmg can never be the auto-update artifact.
- **The dmg container gets its own explicit seal.** A new `hooks.postMake` in `forge.config.ts` calls `sealDmg()`: `codesign --sign --timestamp`, then `xcrun notarytool submit --wait`, then `xcrun stapler staple`. It reuses exactly the credentials `packagerConfig` already consumes (`APPLE_SIGNING_IDENTITY`/`CSC_LINK`, and either `AO_NOTARY_PROFILE` or the `APPLE_API_KEY` trio). No new secrets.
- **`mac.identity` is `null`** so electron-builder can never re-sign the already signed, notarized and stapled `.app` that `packagerConfig` produced before any maker ran.
- **`dmg.sign` is false**, per electron-builder's own warning that signing during dmg creation conflicts with notarization. The container is sealed afterwards instead.
- **No credentials means a warned no-op**, not a build failure, so unsigned local builds and the unsigned desktop-testing pipeline still produce a dmg. When credentials ARE present, any failure throws and fails the build: a silently unsigned dmg published as a release asset is the outcome this work exists to prevent.

**Download links intentionally NOT flipped in this PR.** `DOWNLOAD_URL_MAC_ARM64`/`_X64` and the two macOS README rows still point at the `.zip`. No published release carries a dmg yet, so flipping them at merge time would 404 for every macOS visitor until the next release, and the release workflows are paused. #3267 decision 8 step 6 makes this ordering explicit: flip only after a real release has published both formats and the dmg is verified. There is now a comment at the constants explaining the pending flip and naming the README rows, so the rollout step is discoverable from the code rather than only from the issue. Say the word if you would rather take the flip here and accept the gap.

The private-pipeline handoff for dmg asset passthrough is in #3267 and is owned elsewhere.

## Verified here

- `npm run frontend:typecheck` (`tsc --noEmit`): clean.
- Full frontend vitest suite: 1216 passing, 0 failing. That includes the new `frontend/scripts/verify-mac-artifact.test.mjs` (9), `frontend/scripts/e2e-mac-update.test.mjs` (9) and `frontend/makers/maker-dmg.test.ts` (10).
- `verify-mac-artifact.sh` parses under `bash -n`, is committed executable, and every usage/precondition path exits 2 with the right message. Portability detail: it avoids `mapfile` because macOS ships bash 3.2.
- `.github/workflows/mac-update-e2e.yml` parses as YAML.
- The maker's `buildForge` call shape is asserted against a mocked `app-builder-lib`, and `sealDmg`'s exact command lines are asserted against a stubbed `execFile` for all five credential combinations, including the failure path.

## NOT verified here, and why

Being explicit rather than implying more coverage than exists.

- **Nothing was signed, notarized or stapled.** No Apple credentials in this environment. `sealDmg`'s command construction is unit-tested, but no real `codesign`, `notarytool` or `stapler` invocation was ever made. First real exercise is a signed build.
- **The three verification commands were never run against a real artifact.** Fully mocking `codesign`/`spctl`/`xcrun` needs real signing material, and a mock that always passes proves nothing about a gate. So the smoke test deliberately covers only the usage and precondition paths, which is the half a runner with no signing material can honestly assert. The meaningful verification is the private pipeline's post-staple gate (workstream 1) running this script against real published artifacts. #3288 recommends dry-running it against several past releases before making it blocking.
- **No dmg was actually produced.** `npm run make` does not complete in this worktree: it dies during electron-packager's "Finalizing package" step, before any maker runs. Confirmed environmental rather than caused by this change, since plain `npm run package` on otherwise unmodified code dies at the identical step. So `buildForge` accepting `mac: ["dmg:arm64"]` end to end, and `hdiutil` producing a mountable image, are both unverified and need a macOS build to confirm.
- **The postMake hook has never executed.** Same reason.
- **The update e2e job has never run.** It needs a real macOS runner, a real signed N-1 release with its `agent-orchestrator-darwin-arm64.zip` alias, and a live N feed. None of that exists here. The sentinel's event wiring is per source reading of the pinned `electron-updater@6.8.9` tarball, not observed at runtime. The first real run should be a `workflow_dispatch` against two existing published releases before anyone considers making it a gate.
- **Timeout defaults in the harness (20 min download, 5 min swap, 3 min launch) are estimates**, not measured. All three are flags, so the first real run can tune them.

## Out of scope, confirmed untouched

No changes to the private signing/release pipeline. No release workflow un-paused. No re-enabling of mac differential updates. No broader `AgentWrapper` to `Untrivial-ai` org rename.

Closes nothing on its own; #3288 stays open for workstreams 1 and 3's guard.

Refs #3288, #3267, #3151, #3034.

### DIFF
--- a/.github/workflows/mac-update-e2e.yml
+++ b/.github/workflows/mac-update-e2e.yml
@@ -1,0 +1,139 @@
+name: macOS update e2e
+
+# End-to-end macOS auto-update test (#3288 workstream 0): install version N-1,
+# point it at version N's REAL published feed, and prove the update downloads,
+# stages, installs, and that the relaunched app both reports N and is alive.
+#
+# This is the check that would have caught both #3034 and the 29 Jul incident
+# before any user saw them. The interesting logic lives in
+# frontend/scripts/e2e-mac-update.mjs; this job only supplies a real macOS host
+# and a real signed baseline install.
+#
+# Deliberately NOT wired into any release workflow yet:
+#   - The release workflows are paused, and un-pausing them is a separate
+#     decision that does not belong to this change.
+#   - macOS runner minutes are materially more expensive than Linux, and this
+#     project ships nightlies frequently. #3288 recommends making this a hard
+#     required gate on promotion to the STABLE channel specifically (the exact
+#     transition that failed in production), and treating nightly-to-nightly
+#     coverage as optional pending real per-run cost data. The "cost" step below
+#     prints the wall clock so that question can be settled with numbers.
+#
+# To wire it as a stable-promotion gate later, call it from the release
+# workflow:
+#
+#   update-e2e:
+#     uses: ./.github/workflows/mac-update-e2e.yml
+#     with:
+#       baseline_tag: v<previous stable>
+#       target_version: <new version>
+#
+# Requires: the baseline release's signed macOS zip must be published, and the
+# target version's release + latest-mac.yml feed must already be live, since the
+# app resolves the update through the real GitHub feed.
+
+on:
+  workflow_dispatch:
+    inputs:
+      baseline_tag:
+        description: "Release tag to install as N-1 (e.g. v0.10.3)"
+        required: true
+        type: string
+      target_version:
+        description: "Version the app must end up on, no leading v (e.g. 0.10.4)"
+        required: true
+        type: string
+      channel:
+        description: "Update channel to run the test on"
+        required: false
+        default: latest
+        type: string
+  workflow_call:
+    inputs:
+      baseline_tag:
+        required: true
+        type: string
+      target_version:
+        required: true
+        type: string
+      channel:
+        required: false
+        default: latest
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  update-e2e:
+    # macos-latest is Apple silicon, so the arm64 asset is the one under test.
+    runs-on: macos-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Record start time
+        id: clock
+        run: echo "started=$(date +%s)" >> "$GITHUB_OUTPUT"
+
+      - name: Download the baseline (N-1) macOS build
+        env:
+          GH_TOKEN: ${{ github.token }}
+          BASELINE_TAG: ${{ inputs.baseline_tag }}
+        run: |
+          set -euo pipefail
+          mkdir -p "$RUNNER_TEMP/baseline"
+          # The version-free alias is published for every release (see
+          # frontend/docs/desktop-release.md), so this name is stable across tags.
+          gh release download "$BASELINE_TAG" \
+            --repo "$GITHUB_REPOSITORY" \
+            --pattern 'agent-orchestrator-darwin-arm64.zip' \
+            --dir "$RUNNER_TEMP/baseline"
+
+      - name: Verify the baseline artifact is sealed, notarized and stapled
+        # Same canonical script the release pipeline's post-staple gate uses.
+        # An unsigned or unstapled baseline cannot auto-update, so a failure
+        # here is a real signal, not harness noise.
+        run: frontend/scripts/verify-mac-artifact.sh "$RUNNER_TEMP/baseline/agent-orchestrator-darwin-arm64.zip"
+
+      - name: Install the baseline into /Applications
+        run: |
+          set -euo pipefail
+          rm -rf "/Applications/Agent Orchestrator.app"
+          # ditto, never unzip: plain unzip leaves stray AppleDouble files inside
+          # the bundle and breaks the seal (see verify-mac-artifact.sh).
+          ditto -x -k "$RUNNER_TEMP/baseline/agent-orchestrator-darwin-arm64.zip" /Applications
+          plutil -extract CFBundleShortVersionString raw -o - \
+            "/Applications/Agent Orchestrator.app/Contents/Info.plist"
+
+      - name: Run the update end to end
+        env:
+          TARGET_VERSION: ${{ inputs.target_version }}
+          CHANNEL: ${{ inputs.channel }}
+        run: |
+          set -euo pipefail
+          node frontend/scripts/e2e-mac-update.mjs \
+            --app "/Applications/Agent Orchestrator.app" \
+            --expect-version "$TARGET_VERSION" \
+            --channel "$CHANNEL"
+
+      - name: Verify the installed app after the update
+        # The swap is done by ShipIt, not by our signing pipeline, so re-run the
+        # full trio on what actually ended up on disk.
+        if: always()
+        run: frontend/scripts/verify-mac-artifact.sh "/Applications/Agent Orchestrator.app"
+
+      - name: Report runner cost
+        if: always()
+        env:
+          STARTED: ${{ steps.clock.outputs.started }}
+        run: |
+          set -euo pipefail
+          elapsed=$(( $(date +%s) - STARTED ))
+          # macOS minutes bill at a multiplier; #3288 asks for real numbers before
+          # deciding whether nightly-to-nightly coverage is worth running.
+          echo "macOS update e2e wall clock: $(( elapsed / 60 ))m $(( elapsed % 60 ))s" >> "$GITHUB_STEP_SUMMARY"
--- a/frontend/forge.config.ts
+++ b/frontend/forge.config.ts
@@ -1,6 +1,7 @@
 import type { ForgeConfig } from "@electron-forge/shared-types";
 import { VitePlugin } from "@electron-forge/plugin-vite";
 import MakerNSIS from "./makers/maker-nsis";
+import MakerDMG, { sealDmg } from "./makers/maker-dmg";
 import MakerAppImage from "./makers/maker-appimage";
 import { writeFileSync } from "node:fs";
 
@@ -80,6 +81,22 @@ const config: ForgeConfig = {
 			].join("\n");
 			writeFileSync("app-update.yml", yml);
 		},
+		// The dmg container is NOT signed, notarized or stapled by any maker
+		// (neither Forge's maker-dmg nor app-builder-lib's dmg target does it), and
+		// the .app's own stapled ticket does not propagate through an unsealed
+		// container. So seal it here, after the maker has produced it, reusing the
+		// same credentials packagerConfig already consumes (#3267 decision 3).
+		// The .app inside was already signed + notarized + stapled by
+		// packagerConfig above, before any maker ran; nothing here touches it.
+		postMake: async (_forgeConfig, makeResults) => {
+			for (const result of makeResults) {
+				if (result.platform !== "darwin") continue;
+				for (const artifact of result.artifacts) {
+					if (artifact.endsWith(".dmg")) await sealDmg(artifact);
+				}
+			}
+			return makeResults;
+		},
 	},
 	rebuildConfig: {},
 	makers: [
@@ -97,7 +114,21 @@ const config: ForgeConfig = {
 			},
 			["win32"],
 		),
+		// macOS auto-update artifact. This entry can NEVER be removed:
+		// MacUpdater.doDownloadUpdate looks for a "zip" and explicitly excludes
+		// .pkg/.dmg, throwing ERR_UPDATER_ZIP_FILE_NOT_FOUND otherwise, so the zip
+		// and latest-mac.yml must keep publishing forever (#3267 decision 2).
 		{ name: "@electron-forge/maker-zip", platforms: ["darwin"], config: {} },
+		// macOS FIRST-INSTALL artifact, additive to the zip above: a dmg has no
+		// user-driven extraction step, so a third-party unzip tool can no longer
+		// break the signature seal on the way in (see makers/maker-dmg.ts, #3267).
+		new MakerDMG(
+			{
+				appId: "dev.agent-orchestrator.desktop",
+				productName: "Agent Orchestrator",
+			},
+			["darwin"],
+		),
 		// Linux fetch-and-run artifact for `ao start`: a single self-contained
 		// AppImage the Go bootstrapper downloads and runs directly (see
 		// makers/maker-appimage.ts). The deb/rpm makers below stay for users who
--- a/frontend/makers/maker-dmg.ts
+++ b/frontend/makers/maker-dmg.ts
@@ -1,0 +1,154 @@
+import path from "node:path";
+import { execFile } from "node:child_process";
+import { promisify } from "node:util";
+import { MakerBase, type MakerOptions } from "@electron-forge/maker-base";
+import type { ForgePlatform } from "@electron-forge/shared-types";
+
+const run = promisify(execFile);
+
+// macOS first-install distribution as a .dmg, so a user never runs an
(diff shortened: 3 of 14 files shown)


illegalcall (approved):

frontend/makers/maker-dmg.ts
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	async make({ dir, targetArch, appName }: MakerOptions): Promise<string[]> {
+		const { buildForge } = await import("app-builder-lib");
+		const cfg = this.config ?? {};
+		// Mirror buildForge's own output layout (<dir>/../make) so artifacts land
+		// where Forge's publisher expects them.
+		const output = path.join(path.dirname(path.resolve(dir)), "make");
+		return buildForge(
+			{ dir },
```
illegalcall, line 58: [P1] `dir` is the Forge package directory, not the `.app` path. `buildForge` treats it as `prepackaged`, so this produces `Agent Orchestrator.app/Agent Orchestrator.app`; the outer bundle has no `Contents` and will not launch. Pass `path.join(dir, `${appName}.app`)` and add a layout-level maker test.

.github/workflows/mac-update-e2e.yml
```diff
@@ -0,0 +1,139 @@
...
+          plutil -extract CFBundleShortVersionString raw -o - \
+            "/Applications/Agent Orchestrator.app/Contents/Info.plist"
+
+      - name: Run the update end to end
+        env:
+          TARGET_VERSION: ${{ inputs.target_version }}
+          CHANNEL: ${{ inputs.channel }}
+        run: |
+          set -euo pipefail
+          node frontend/scripts/e2e-mac-update.mjs \
```
illegalcall, line 119: [P1] The process launched here is the published N-1 binary, but the `AO_E2E_UPDATE_SENTINEL` listener is introduced only in this PR. Existing baselines ignore the env var, so the first E2E run waits for the sentinel and never exercises ShipIt. Use a sentinel-capable baseline or observe staging externally, and fail fast when capability is absent.

frontend/makers/maker-dmg.ts
```diff
@@ -0,0 +1,154 @@
...
+		console.warn(`[dmg] no signing material (APPLE_SIGNING_IDENTITY / CSC_LINK); leaving ${dmgPath} unsigned`);
+		return;
+	}
+
+	console.log(`[dmg] signing ${dmgPath}`);
+	await run("codesign", ["--sign", identity, "--timestamp", "--force", dmgPath]);
+
+	const notaryArgs = resolveNotaryArgs(env);
+	if (!notaryArgs) {
+		console.warn(`[dmg] no notarization credentials (AO_NOTARY_PROFILE / APPLE_API_KEY); ${dmgPath} is not notarized`);
```
illegalcall, line 140: [P1] Partial release credentials fail open: signing without notary credentials reaches this warning and returns success, while notary credentials without a signing identity return earlier. That can publish an unsigned or unnotarized DMG. Allow only the zero-credential local case; throw on every partial combination and test both directions.

illegalcall (commented):
Release-hardening review on exact head 032dcdf66be3ee1084f9f37dbce9b28cf8b59f9b. Seven actionable findings follow; this COMMENT review does not alter the existing approval.

frontend/makers/maker-dmg.ts
```diff
@@ -0,0 +1,154 @@
...
+						// maker. `identity: null` guarantees electron-builder never
+						// re-signs it and breaks that seal; we only want its dmg target.
+						identity: null,
+					},
+					dmg: {
+						// Left unsigned on purpose. electron-builder's own docs: "Signing
+						// is not required and will lead to unwanted errors in combination
+						// with notarization requirements." The dmg container is sealed
+						// afterwards by sealDmg() in forge.config.ts's postMake hook.
+						sign: false,
```
illegalcall, line 81: [P2] `dmg-builder@26.15.3` defaults `writeUpdateInfo` on, so this returns a `.dmg.blockmap` generated before postMake changes the DMG bytes. Forge can publish that stale macOS sidecar, violating the full-download-only policy. Set `writeUpdateInfo: false` and assert no DMG blockmap is returned.

frontend/forge.config.ts
```diff
@@ -80,6 +81,22 @@ const config: ForgeConfig = {
...
+		// the .app's own stapled ticket does not propagate through an unsealed
+		// container. So seal it here, after the maker has produced it, reusing the
+		// same credentials packagerConfig already consumes (#3267 decision 3).
+		// The .app inside was already signed + notarized + stapled by
+		// packagerConfig above, before any maker ran; nothing here touches it.
+		postMake: async (_forgeConfig, makeResults) => {
+			for (const result of makeResults) {
+				if (result.platform !== "darwin") continue;
+				for (const artifact of result.artifacts) {
+					if (artifact.endsWith(".dmg")) await sealDmg(artifact);
```
illegalcall, line 95: [P1] This seals the DMG but never runs the canonical post-seal verifier, which currently rejects `.dmg`. A bad signature, Gatekeeper assessment, or ticket state can reach publishing. Extend the verifier for DMGs (`codesign --verify`, `spctl -a -t open --context context:primary-signature`, and `stapler validate`) and invoke it here after `sealDmg`.

frontend/scripts/e2e-mac-update.mjs
```diff
@@ -0,0 +1,257 @@
...
+	rmSync(sentinel, { force: true });
+	rmSync(opts.runFile, { force: true });
+	seedUpdateSettings(opts.stateDir, opts.channel);
+
+	// Launched as a child process rather than via `open` so the sentinel env var
+	// reaches the app. The RELAUNCH below deliberately uses `open -a` instead,
+	// because by then the bundle has been swapped underneath us.
+	console.log(`launching ${opts.app} (channel: ${opts.channel})`);
+	const child = spawn(join(opts.app, "Contents", "MacOS", readExecutableName(opts.app)), [], {
+		env: { ...process.env, AO_E2E_UPDATE_SENTINEL: sentinel },
```
illegalcall, line 189: [P2] Custom `--state-dir` and `--run-file` never reach the app: this launch passes only the sentinel, and the `open` relaunch passes neither `AO_DATA_DIR` nor `AO_RUN_FILE`. The harness seeds and polls different paths and can time out. Propagate both overrides through both launches, constrain the deleted run-file path, and add a behavioral test.

.github/workflows/mac-update-e2e.yml
```diff
@@ -0,0 +1,139 @@
...
+# workflow:
+#
+#   update-e2e:
+#     uses: ./.github/workflows/mac-update-e2e.yml
+#     with:
+#       baseline_tag: v<previous stable>
+#       target_version: <new version>
+#
+# Requires: the baseline release's signed macOS zip must be published, and the
+# target version's release + latest-mac.yml feed must already be live, since the
```
illegalcall, line 32: [P1] This cannot be a stable-promotion gate while N and `latest-mac.yml` must already be public: stable clients can update before the check finishes. Test against an isolated candidate or draft feed and promote the public feed only after success, or document this as post-release monitoring.

harshitsinghbhandari:
Thanks for the review. All seven findings are real and none are dismissed. Merging this as the base and carrying every one of them into an immediate follow-up PR rather than churning this branch, per the requester's call to keep PR count against this repo low.

Threads resolved here only to mark them as triaged into that follow-up; full text is preserved in the follow-up's description. Carried forward:

P1
1. `maker-dmg.ts`: `dir` is the Forge package directory, not the `.app`. `buildForge` treats it as prepackaged, producing a nested `Agent Orchestrator.app/Agent Orchestrator.app`. Pass `path.join(dir, \`${appName}.app\`)` plus a layout-level maker test.
2. `sealDmg`: partial credentials fail open (identity without notary creds warns and returns success). Throw on every partial combination; allow only the zero-credential local case.
3. `forge.config.ts` postMake: seals the dmg but never runs the canonical verifier. Extend `verify-mac-artifact.sh` to accept `.dmg` (with `spctl -a -t open --context context:primary-signature`) and invoke it after `sealDmg`.
4. `mac-update-e2e.yml`: the N-1 baseline predates the sentinel, so the first run waits forever. Needs a sentinel-capable baseline or external staging observation, and a fail-fast capability check.
5. `mac-update-e2e.yml`: cannot be a stable-promotion gate while N and `latest-mac.yml` must already be public. Either test against an isolated candidate feed and promote after success, or document it as post-release monitoring.

P2
6. `dmg-builder@26.15.3` defaults `writeUpdateInfo` on, so a `.dmg.blockmap` is generated before postMake rewrites the dmg bytes. Set `writeUpdateInfo: false` and assert no dmg blockmap is emitted.
7. `e2e-mac-update.mjs`: `--state-dir`/`--run-file` never reach the app (neither launch passes `AO_DATA_DIR`/`AO_RUN_FILE`). Propagate both through both launches and constrain the deleted run-file path.

Worth stating plainly: finding 1 means the dmg produced by this branch would not launch. Nothing publishes a dmg yet (no maker output is wired into a release, and the download links deliberately still point at the zip), so merging does not ship a broken artifact, but the dmg maker should be treated as not usable until the follow-up lands.

PR #3295, opened 2026-07-29
fix(browser): prevent hidden native view overlays

### LINKED ISSUE #3285: bug(browser): native WebContentsView overlays agent session when switching sessions
## Bug

The native browser preview `WebContentsView` renders on top of the active agent session — even when the Browser inspector tab was never clicked. Reported by @neversettle17-101 (Aditi) on Discord with a screenshot showing a different session's content (demo/dashboard-screenshot) overlaid on the "reviewer flow update" session.

Related to #2491 (positioning) but distinct — this is a **visibility/lifecycle** problem, not a coordinate misalignment.

## Repro

1. Have a session with `ao preview <url>` active (e.g. a demo session loading a localhost dev server).
2. Navigate to a different session in the sidebar.
3. The previous session's native browser view paints over the new session's terminal/inspector area.

## Root cause (traced from code)

The native `WebContentsView` is a window-level overlay Chromium composites above the entire DOM. Two independent systems control its visibility — and they are not coordinated:

### 1. Renderer-side (`useBrowserView.ts`)
All lifecycle effects are `useEffect` (lines 229, 274, 483), which run **after the browser paints**. When `sessionId` changes:
- React commits the new session's DOM
- **Browser paints** — the old session's native view is still visible at its last bounds
- `useEffect` cleanup runs → `sendHiddenBounds(oldViewId)` → IPC to main process → view hidden

There is a one-paint-frame window where the old view overlays the new session. If the preview page is a dev server with HMR, repeated navigations can make this flash persistently visible.

### 2. Main-process-side (`browser-view-host.ts`)
`wireNavEvents` calls `entry.view.setVisible?.(true)` on every `did-navigate` event (line 571), and `navigate()` calls it after `loadURL` (line 304). This happens **without notifying the renderer**. If a hidden view's page navigates (HMR, SPA redirect, reload), the main process re-shows it. The renderer has no listener for this on sessions it's no longer managing — the `onNavState` callback bails early when `state.viewId !== viewIdRef.current`.

After `sendHiddenBounds`, the view's bounds are `OFFSCREEN_BOUNDS` (0×0 at −10000,−10000), so `setVisible(true)` alone should not paint. But the one-frame flash from the `useEffect` timing gap is the primary visible symptom.

## Suggested fixes

1. **Use `useLayoutEffect` for the view-hiding cleanup** in `useBrowserView` so the old view is hidden before the browser paints the new session. This eliminates the one-frame flash.

2. **Track a `rendererVisible` flag in the main process** per view entry. `did-navigate` should only call `setVisible(true)` if the renderer has previously indicated the view should be visible (via a `setBounds({ visible: true })` call). This prevents the main process from un-hiding a view the renderer intentionally hid.

3. **Forward `did-navigate` visibility changes to the renderer** so it can re-evaluate whether the view should be shown, even for sessions that are no longer the active route.

## Files

- `frontend/src/renderer/hooks/useBrowserView.ts` — lifecycle effects, `measureAndSend`, `sendHiddenBounds`
- `frontend/src/main/browser-view-host.ts` — `wireNavEvents` (`did-navigate` → `setVisible(true)`), `navigate`, `setBounds`
- `frontend/src/renderer/components/SessionView.tsx` — `useBrowserView` instantiation, `active` flag computation

## Environment

Reported on macOS (Aditi's machine). Issue #2491 reported similar positioning problems on Debian Linux. Likely cross-platform given the architecture is platform-agnostic.

Vaibhaav-Tiwari:
## Summary
- make native browser visibility renderer-owned instead of navigation-owned
- hide the WebContentsView before paint when the browser slot becomes inactive or unmounts
- keep ao preview loading in the background without painting over other sessions

Fixes #3285.

## Tests
- npm.cmd run test -- --pool=threads --maxWorkers=1 src/main/browser-view-host.test.ts src/renderer/hooks/useBrowserView.test.tsx src/renderer/components/SessionView.test.tsx
- npm.cmd run typecheck

## Notes
- Full frontend test suite was attempted twice; it fails in this Windows environment on unrelated supervisor-link named-socket tests with listen EACCES under %LOCALAPPDATA%\\Temp.


frontend/src/main/browser-view-host.ts
```diff
@@ -580,6 +607,9 @@ function wireNavEvents(contents: BrowserWebContents, options: BrowserViewHostOpt
 	contents.on("did-stop-loading", update);
 	contents.on("did-fail-load", (_event, errorCode, errorDescription) => {
 		if (errorCode === -3) return;
+		entry.rendererVisible = false;
```
neversettle17-101: This clears the renderer-owned visibility flag on any non-aborted load failure, which can make an active browser stay hidden after recovery. For example, if a visible preview fails to load and the user hits Reload, or `ao preview` reruns the same URL, `reload()`/`navigate()` can complete without the renderer visibility effect rerunning because `active` and `navState.url` did not change. `applyEntryBounds()` then keeps the view hidden because `rendererVisible` is still false. Consider preserving the renderer visibility state here and only hiding the native view for the failed load, or explicitly re-requesting bounds on recovery; a fail-then-reload/same-URL preview test would cover it.

Vaibhaav-Tiwari: Fixed in 83b481e7a: failed loads now only hide the native view and preserve the renderer-owned session bounds, and renderer navigation commands reapply those bounds on recovery. Added a fail-then-reload regression test alongside the hidden-navigation guard.

Hardik180704 (approved):
Reviewed the latest head. The renderer-owned visibility behavior is correctly scoped to the active browser slot, hidden bounds are sent during unmount/inactive transitions, and the earlier failed-load recovery concern is addressed by preserving session visibility and reapplying bounds on navigation recovery. The regression coverage and CI look good.

PR #3296, opened 2026-07-29
fix(notifications): clickable rows, unseen/unresolved sections, auto-resolution

### LINKED ISSUE #3210: bug(notifications): notification rows not clickable, redundant terminal button, unnecessary mark-as-read mechanism
## Bug

The notification dropdown has three UX issues with click targets, read/unread state, and tab structure:

### 1. Notification rows are not clickable as a whole

Only the title text inside each notification is clickable — the icon, body text, and surrounding row area have no click handler. The entire notification row should be a click target that navigates to the session (as long as the session is alive), matching the user's expectation from the hover highlight (the row already shows a `hover:bg-interactive-hover` background, implying clickability).

**Root cause:** `NotificationItem` (`frontend/src/renderer/components/NotificationCenter.tsx:350-450`) renders the row as a plain `<div>` with `role="listitem"` — no `onClick`. Only the inner title `<a>` (PR-type, line 384) or `<button>` (session-type, line 400) is interactive. The icon container (lines 374-381) and body `<p>` (lines 413-417) are not clickable.

### 2. Redundant terminal button on PR-type notifications

PR-type notifications show a separate `SquareTerminal` button (lines 420-430) that navigates to the session. If the whole notification row is clickable and navigates to the session, this button is redundant and should be removed.

### 3. Rethink "Unread/All" tabs → "Unseen" and "Unresolved" sections

The current "Unread" / "All" tab structure and the manual mark-as-read mechanism are not the right model. `All` notifications are not useful forever — a notification from three days ago about an agent waiting for input is no longer relevant today. The proposed redesign has two sections:

**Section 1 — Unseen notifications.** Totally new notifications the user hasn't looked at yet. Once the user opens the notifications panel and sees them, they clear (auto-resolve). This is the "mark all read on open" behavior — no manual checkbox needed.

**Section 2 — Unresolved notifications.** Notifications that still need resolution. These should resolve **automatically** when the underlying issue is resolved — no manual checkbox:
- If a session is waiting for input → resolves when the user provides the input
- If a PR is waiting on merge → resolves when the PR is merged

We do not need a checkbox to manually resolve notifications. Both the per-item "Mark as read" checkmark (lines 431-446) and the header "Mark all read" button (lines 222-231) should be removed.

## Reproduction

1. Open the AO desktop app with at least one notification (e.g., a PR ready-to-merge or needs-input event)
2. Click the bell icon to open the notification dropdown
3. Try clicking anywhere on a notification row except the title text
4. **Expected:** navigates to the session page
5. **Actual:** nothing happens — only the title text is a clickable link/button

## Root Cause

All three issues are in `frontend/src/renderer/components/NotificationCenter.tsx`:

| Issue | Lines | Problem |
|-------|-------|---------|
| Row not clickable | 367-373 | Container `<div>` has `role="listitem"` but no click handler; only inner title is interactive |
| Redundant terminal button | 420-430 | `SquareTerminal` button duplicates what a clickable row would provide |
| Manual mark-as-read controls | 222-231, 431-446 | Two manual mark-read controls instead of auto-clear-on-open + auto-resolve |
| PR navigation early-return | 66-76 | `openPrimary` returns before session navigation for PR-type |

The backend always populates `target.sessionId` for both session-type and PR-type notifications (`notificationTargetFromRecord` in `backend/internal/httpd/controllers/notifications.go:223`), so the data is available — this is purely a frontend interaction design issue.

The "Unresolved" auto-resolution will require backend support: the notification service (`backend/internal/service/notification/service.go`) needs logic to mark notifications resolved based on lifecycle state changes (session receives input → needs_input notification resolves; PR merged → ready_to_merge notification resolves).

## Fix

1. **Make the whole row clickable:** Move the click handler (`onOpenSession`) to the container `<div>` (or wrap in a `<button>`), calling `openSession` for all notification types. Keep the external-link icon as a visual indicator for PR-type notifications.

2. **Remove the terminal button:** Delete the `SquareTerminal` button block (lines 420-430) and the `SquareTerminal` import.

3. **Replace "Unread/All" tabs with "Unseen/Unresolved" sections:**
   - **Unseen:** notifications clear automatically when the panel is opened (auto-read-on-open)
   - **Unresolved:** notifications auto-resolve when the underlying issue is resolved (session receives input, PR merged, etc.) — no manual checkbox
   - Remove per-item mark-as-read button and header "Mark all read" button

## Impact

- Core notification UX degraded — users must click precisely on the title text to navigate
- The terminal button and mark-as-read controls add visual clutter with no real value
- "All" tab accumulates stale notifications that are no longer relevant
- Manual mark-as-read mechanism adds cognitive overhead for state the system can derive automatically

## Related

- [#2726](https://github.com/AgentWrapper/agent-orchestrator/issues/2726) — UI/UX audit finding AO-09: notification badge should deep-link to session. Overlaps with issue #1.
- [#3033](https://github.com/AgentWrapper/agent-orchestrator/issues/3033) — notification title includes session instance number; counter never resets

## Screenshot

The notification dropdown shows two unread items. Each has a hover background suggesting clickability, but only the underlined title text (PR link / session button) is actually clickable. A separate terminal icon button appears on the PR notification. Each notification has a checkmark button to mark individually read, and the header has a "Mark all read" button.

**Source:** Discord #bug-triage thread | **Reported by:** @prateek, @dhruv-sharma | **Analyzed against:** `90d5b3cae`
**Confidence:** High — traced exact components and line numbers

ikeshavvarshney:
Closes #3210

## What

Three fixes to the notification dropdown.

**1. The whole row is a click target.** Previously only the title text was
clickable, despite the row showing a `hover:bg-interactive-hover` background
that implied otherwise. Clicking anywhere now navigates to the session, by
mouse or keyboard (`role="button"`, Enter/Space).

**2. The redundant terminal button is gone.** With the row navigating to the
session, the `SquareTerminal` button duplicated it. The PR title stays a real
`<a>` with `stopPropagation`, so a PR row still offers both destinations: link
opens the PR, row opens the session.

**3. Unread/All tabs replaced by Unseen/Unresolved, with no manual controls.**
Both mark-as-read affordances (per-item checkmark, header "Mark all read") are
removed.

## How

Notifications now carry two independent axes instead of one read flag:

| Column | Meaning | Written by |
|---|---|---|
| `status` | has the user **seen** it | cleared wholesale when the panel opens |
| `resolved_at` | is the underlying issue still **open** | lifecycle only — never a user action |

They are genuinely independent: a notification you already looked at stays in
Unresolved until the thing it reported goes away.

Auto-resolution triggers:

- session leaves the needs-input family, or terminates → `needs_input` resolves
- PR merges, closes, or stops being mergeable → `ready_to_merge` resolves
- daemon startup re-checks open rows against durable session/PR facts, because
  a transition observed while the daemon was down never reaches lifecycle

`pr_merged` / `pr_closed_unmerged` are excluded from Unresolved — they report
something that already happened, so they are worth seeing once and never worth
resolving.

Dedupe moves from `status = 'unread'` to `status = 'unread' OR resolved_at IS
NULL`. Keying on unread alone would let a seen-but-still-unresolved
notification be duplicated by the next observation of the same fact.

SSE gained a `notification_resolved` event so an open panel drops rows live;
the push dispatcher ignores it (nothing new to buzz a phone about).

## Notes

- Migration `0039_notification_resolution.sql` back-fills pre-existing read
  rows as resolved. They have no live resolution tracking behind them, so
  treating the acknowledgement as the resolution is better than resurrecting
  three-day-old rows into the new section.
- The panel renders a snapshot captured at open, because acknowledging empties
  the unread cache and a row must not vanish under the cursor.
- `PATCH /api/v1/notifications/{id}` is kept for the mobile client; only the
  renderer's use of it is removed.
- API additions: `status=unresolved` filter, `resolvedAt` field,
  `unresolvedCount`. `openapi.yaml` and `schema.ts` regenerated.

## Testing

- Renderer suite: 62 files / 852 tests pass
- Backend: every touched package passes (`notify`, `push`,
  `service/notification`, `lifecycle`, `storage/sqlite/store`,
  `httpd/controllers`)
- `typecheck` and `typecheck:e2e` pass
- `golangci-lint`: no issues in any changed file
- Manually verified in the desktop app against a seeded daemon

Some pre-existing failures on the Windows dev box are unrelated to this change
and touch no files here: 4 `supervisor-link` tests (`EACCES` on a Unix-domain
socket path) and 4 backend packages (`config`, `adapters/agent/fake`,
`adapters/runtime/tmux`, `daemon`) that assume POSIX paths or hit Windows file
locks.


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/lib/notifications.ts
```diff
@@ -111,22 +163,12 @@ function mergeNotificationIntoCache(
...
-	});
-}
-
 export function markAllCachedNotificationsRead(queryClient: QueryClient): void {
 	queryClient.setQueryData<NotificationsCache>(unreadNotificationsQueryKey, (current) => {
 		if (!current) return current;
 		return {
 			pageParams: [""],
-			pages: [{ notifications: [], unreadCount: 0 }],
+			pages: [{ notifications: [], unreadCount: 0, unresolvedCount: current.pages[0]?.unresolvedCount ?? 0 }],
```
illegalcall: **[P1] Preserve unseen pagination before clearing this cache**

This resets the unread infinite query to one empty page and discards its `nextCursor`. Opening the panel snapshots only the pages already loaded, while `POST /read-all` marks every unread row in SQLite as read.

Concrete failure: with 130 unseen `pr_merged` / `pr_closed_unmerged` rows, the UI snapshots the first 100, the server acknowledges all 130, and this callback removes the cursor to the remaining 30. Terminal rows are not part of Unresolved and the All view was removed, so those 30 become permanently inaccessible from the panel.

Please preserve an open-time paging source until the snapshot is complete, or acknowledge explicit IDs / an open-time cutoff rather than every server row. A deterministic regression test should return 100 rows plus a cursor out of 101, complete the mutation, fetch the next page, and assert row 101 remains visible during that panel session.

backend/internal/storage/sqlite/queries/notifications.sql
```diff
@@ -43,10 +65,52 @@ UPDATE notifications
...
+       OR activity_state NOT IN ('waiting_input', 'blocked')
+  )
+RETURNING *;
+
+-- name: ResolveStaleReadyToMergeNotifications :many
+UPDATE notifications
+SET resolved_at = sqlc.arg(resolved_at)
+WHERE type = 'ready_to_merge'
+  AND resolved_at IS NULL
+  AND pr_url IN (SELECT url FROM pr WHERE pr_state <> 'open')
```
illegalcall: **[P2] Reconcile the complete ready-to-merge predicate**

The live predicate also considers draft state, CI, review decision, unresolved human comments, and mergeability. This startup repair resolves only when `pr_state <> 'open'`.

There is a concrete persistence gap: the SCM observer writes new durable PR facts before invoking lifecycle. If the daemon stops after persisting `ci_state = failing` (or another blocker) but before lifecycle resolves the existing notification, restart reconciliation leaves that notification unresolved as long as the PR is still open. A later observation may self-heal it, but while SCM access is unavailable the UI contradicts durable state.

Please mirror the live readiness conditions here (or route both paths through one shared decision). Add real-store reconciliation tests for an open PR with failing CI, requested changes, draft state, unresolved comments, and non-mergeable state.

frontend/src/renderer/components/NotificationCenter.tsx
```diff
@@ -173,11 +182,19 @@ export function NotificationCenter({ style }: NotificationCenterProps) {
 		setPanelOpen(false);
 	};
 
+	// One viewport over both sections. Unseen renders first, so its older pages
+	// are what "more" means until it runs out; then unresolved keeps going.
+	const pagingQuery = unreadQuery.hasNextPage ? unreadQuery : unresolvedQuery;
+	const isLoading =
+		(unreadQuery.isLoading && unseen.length === 0) || (unresolvedQuery.isLoading && unresolved.length === 0);
+	const isError = unreadQuery.isError && unresolvedQuery.isError;
```
illegalcall: **[P2] Surface a one-sided query failure instead of claiming success**

With `&&`, an empty successful Unseen request plus a failed Unresolved request produces `isError = false` and `isEmpty = true`, so the panel renders “You're all caught up” even though unresolved data was never loaded. The mirror case has the same problem.

Please keep any successfully loaded section visible but render a section-level warning/retry for the failed query. Add tests for each direction of partial failure and assert neither path displays the all-caught-up message.

illegalcall (commented):
Reviewed the complete rebased diff at `477037c` and traced the behavior from SQLite through lifecycle, REST/SSE, the React Query cache, and the notification panel.

The rebase resolution itself looks correct and scoped: the regenerated sqlc output retains both main's re-engagement query and this PR's resolution queries, and the notification migration is correctly moved to `0039` to avoid main's existing `0038`.

The two-axis seen/resolved model is sound for ordinary live transitions, but I found three correctness gaps, commented inline:

1. **High:** acknowledging the panel can permanently hide unseen terminal notifications beyond the first 100.
2. **Medium:** startup ready-to-merge reconciliation checks only open/closed PR state, not the complete live readiness predicate.
3. **Medium:** a failure of exactly one section query can be presented as “You're all caught up.”

I recommend fixing the pagination issue before merge because it causes irreversible visibility loss. I also recommend closing the startup reconciliation gap in this PR because restart repair is part of the new auto-resolution contract. The partial-error presentation is smaller, but should at least be covered by a follow-up and regression test.

All current CI checks are green; these are semantic boundaries that the existing tests do not exercise.

ikeshavvarshney:
solving conflicts..

ikeshavvarshney:
Merged latest main in — conflicts are resolved and the PR is mergeable again.

- `gen/notifications.sql.go` was the only conflict. sqlc emits alphabetically,
  so this branch's `Resolve*` functions land directly against the
  `SessionHasUnread` block your revert (#3394) deletes. Resolved by taking this
  branch's side and regenerating; the stale function drops out on its own.
- Migration renumbered to `0040` — `0039` collided with
  `0039_drop_orchestrator_reengagement.sql`. Worth flagging: git merges that one
  silently since the filenames differ, and it only surfaces as a goose
  "duplicate version 39" panic that kills every test opening a store.
- The revert landed cleanly otherwise: `orchestratorloop` is gone, no orphaned
  references left.

Verified on the merged tree: build, vet, gofmt, `tsc`, the renderer suite
(879 tests), and the storage, httpd, notify, service/notification, lifecycle,
push, and domain packages all pass.

illegalcall:
Thanks @ikeshavvarshney 
I am still getting notified when im already on the same terminal session. We need to not show it when we are on the same worker terminal.
image
image

ikeshavvarshney:
> Thanks @ikeshavvarshney I am still getting notified when im already on the same terminal session. We need to not show it when we are on the same worker terminal. 

@illegalcall I think this can be a seperate issue. I have not worked on this, in this PR.

illegalcall:
@ikeshavvarshney A separate issue is fine for the same-session in-app bell behavior. Please file the new issue and link it in this PR so the follow-up is tracked; once it is linked, you are good to start on that separately.

There is one blocking migration issue that does need to be fixed in this PR before merge:

- Current `main` now contains `0040_add_session_diff_base.sql`.
- This PR also contains `0040_notification_resolution.sql`.
- Git reports the branch as mergeable because the filenames differ, but Goose keys migrations by numeric version. The merged tree will therefore contain duplicate version `0040`, which fails at runtime when the migration set is loaded.

Please merge/rebase the latest `main`, rename this PR migration to the next free version (`0041_notification_resolution.sql` against current main), and rerun the SQLite/store tests plus the normal lint command. Please also update the stale migration filename in the PR description. The same-session bell change can stay in the linked follow-up issue, but the migration collision must be resolved here.

AgentWrapper:
@ikeshavvarshney notifications from terminated sesssions should not be clickable, instead they should have a restore agent session button on the right of the notification

Also, can you show what the notification dropdown would look like when there is a mix of read and unread notifications?

Ideally unread notifications should be highlighted visually for users to understand that they are unread.
Instead of haing a separate "Unread" title text.

Annieeeee11:
@AgentWrapper  I'll be merging this PR and taking it from here since @ikeshavvarshney is unavailable today. I'll create a follow-up PR to address the remaining issues.

Annieeeee11 (approved):

PR #3298, opened 2026-07-29
fix(terminal): scope the tab strip to one session, restore one-click new terminal

Pulkit7070:
## Problem

https://github.com/Untrivial-ai/agent-orchestrator/pull/3138 let a session's terminal strip pin tabs for *other* sessions, across projects. Two things came out of that:

- The strip stopped being about one session. Session 1's tab bar could show session 2's terminal, which makes it ambiguous which session you are typing into.
- The everyday action, "open a terminal", moved behind a dropdown whose first item was "Terminal", so one click became two.

## Change

The strip is one session's own again: its tab, then the shells it opened.

- The `+` opens a terminal on a single click. The session launcher, its search box, and the "Show all sessions" list are gone.
- The session's own tab is the first tab and has no close affordance. Ending a session stays the topbar's Kill.
- Removed the cross-session plumbing: the `tabOwner` search param on both session routes, `sessionTabsByOwner` and its localStorage persistence, and the add/close/select project-session props on `CenterPane`. Net 543 deletions against 149 insertions.
- Closing a shell now drops the tab optimistically instead of waiting on the round trip.

## On the close latency

The optimistic update is only half of it. The daemon's DELETE blocks for about 5 seconds before answering, which is a separate bug with its own root cause on the Go side; that fix is https://github.com/Untrivial-ai/agent-orchestrator/pull/3328. This PR makes the tab disappear on click regardless of how long the daemon takes.

## Verification

Drove the renderer in a browser: `+` adds a terminal with no menu in the DOM, right click no longer produces a launcher, the session tab has no close button, and the tab disappears on click. A unit test pins the optimistic behavior properly by holding the DELETE unresolved and asserting the tab has already left the query cache, which the browser run could not prove on its own since the preview build has no daemon.

`npm test` (1184 tests, 92 files) and `npm run typecheck` both pass.

## Pre-existing failures, not addressed here

`frontend/e2e/shell-terminal-tabs.spec.ts` has two failures that #3138 introduced. Confirmed against a detached worktree at `6b6190deb`, so they predate this branch, and they fail at the same points with it applied:

- The board-route test expects a "New terminal" button on `/#/projects/:id`, and nothing renders one there anymore.
- The session-route test gets further with this branch (the button is found and opens a terminal again) but then fails at `getByTitle("Session terminal")`: the session tab only carries that title when its label is not truncated (`CenterPane.tsx`), and the demo label truncates.

Both are worth their own fix. The second overlaps the tab-bar visual pass, which is deliberately not in this PR.

### DIFF
--- a/backend/internal/service/shellterm/service.go
+++ b/backend/internal/service/shellterm/service.go
@@ -233,6 +233,23 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		return ShellTerminal{}, fmt.Errorf("open shell terminal: handle id: %w", err)
 	}
 
+	// Resolve the tab name BEFORE the runtime exists. This reads the store, and
+	// a store failure after Create would strand a PTY that nothing can find:
+	// no row means neither shutdown nor the app-run reaper can ever destroy it.
+	// Doing it here means the only failure that can leave a runtime behind is
+	// the insert below, which already rolls back.
+	//
+	// Counting is safe: the session gate is held for the whole open, so two
+	// concurrent opens for one session cannot pick the same number.
+	title := shellTerminalTitle(workingDir)
+	if in.SessionID != "" {
+		ordinal, err := s.nextSessionTerminalOrdinal(ctx, in.SessionID)
+		if err != nil {
+			return ShellTerminal{}, err
+		}
+		title = sessionShellTerminalTitle(ordinal)
+	}
+
 	// SessionID is the runtime adapters' name for "what to call this PTY"; it
 	// is not a session row and no sessions record is ever created. The
 	// shellterm- prefix keeps the two namespaces disjoint.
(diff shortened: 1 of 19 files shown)


Pulkit7070:
Closing this as superseded — no work is lost.

https://github.com/Untrivial-ai/agent-orchestrator/pull/3275 landed the same cross-session removal this PR was carrying: the dropdown gone, `+` calling `onNewShellTerminal` directly, the same props dropped (`projectSessions`, `availableProjectSessions`, `onAddProjectSession`, `onCloseProjectSession`, `onSelectProjectSession`, `tabOwnerSessionId`), plus `sessionTabsByOwner` and the `tabOwner` search param. That duplication is what put this branch into conflict across five files.

Everything this PR had that `main` still lacks now lives in https://github.com/Untrivial-ai/agent-orchestrator/pull/3374, rebuilt on top of `main` and conflict-free: the optimistic close, `Terminal N` tab naming, and the exit handling.

Worth noting for anyone following the thread: https://github.com/Untrivial-ai/agent-orchestrator/pull/3329 shows as merged but merged into *this* branch rather than `main`, so its content — including all three review fixes from that PR — never reached `main` either. #3374 carries those too.

PR #3301, opened 2026-07-29
fix(macos): address the seven review findings from #3294

harshitsinghbhandari:
Follow-up to merged PR #3294, fixing the seven findings from the post-approval review on that PR. Tracking issue: #3288. Decision record for the dmg work: #3267.

One PR, five commits. Nothing here changes the release workflows, un-pauses anything, re-enables mac differential updates, or flips any download link from `.zip` to `.dmg`.

## P1-1. maker-dmg passed the wrong path to `buildForge`

**Bug.** `buildForge`'s `dir` becomes electron-builder's `prepackaged`, and on macOS that value IS the app path, not the package directory. `macPackager.packMacTargets` does `appPath = prepackaged ?? join(computeAppOutDir(...), "<productFilename>.app")` and hands `appPath` straight to the dmg target, which copies it into the image renamed to `<productFilename>.app` (`dmg.ts` `computeDmgOptions`' default `contents` entry). Passing Forge's package directory therefore copied the entire package directory in under that name, producing `Agent Orchestrator.app/Agent Orchestrator.app`: an outer bundle with no `Contents`, which cannot launch.

**Change.** Pass `path.join(dir, `${appName}.app`)`. Take the output directory from Forge's own `makeDir` instead of re-deriving it from a path that no longer points at the package dir (`buildForge`'s own default is `dirname(dir)/make`, which would now land inside the package directory).

**Why maker-nsis gets away with the bare `dir`.** Checked, not assumed. `winPackager` goes through the generic `platformPackager.pack`, whose `computeAppOutDir` returns `prepackaged` as the app OUT directory, and the NSIS target wants exactly that directory. Windows has no `.app` wrapper to descend into. The asymmetry is in electron-builder, not in these makers. maker-nsis is left alone.

**Verified.** Read against the installed `app-builder-lib@26.15.3` in `node_modules`, plus a layout-level maker test asserting the exact `dir` handed to `buildForge` and that it is not the package directory.

## P1-2. `sealDmg` failed open on partial credentials

**Bug.** A signing identity with no notary credentials logged a warning and returned success, so a signed but unnotarized dmg could reach publishing. Notary credentials with no signing identity returned even earlier and did nothing at all. Either way the build went green with an artifact Gatekeeper would reject.

**Change.** Exactly one incomplete state is tolerated: neither a signing identity nor notary credentials, which stays a warned no-op so unsigned local builds and the unsigned `desktop-testing.yml` pipeline keep producing a dmg. Every partial combination throws, in both directions, including a partial App Store Connect key trio. `sealDmg` now returns whether it actually sealed, which is what gates the new verification step below.

**Verified.** Unit tests for both directions, the partial trio, and the preserved zero-credential no-op.

## P1-3. postMake sealed the dmg but never verified it

**Bug.** `sealDmg` exiting 0 only says three commands ran on the build machine. It does not say Gatekeeper accepts the published bytes with a stapled ticket. A bad signature, assessment or ticket state could reach publishing unseen.

**Change.** `verify-mac-artifact.sh` now accepts a `.dmg` alongside `.zip` and `.app`, and `forge.config.ts`'s postMake runs it immediately after `sealDmg`, only when `sealDmg` reports it actually sealed (an unsigned local build has nothing to verify and must not fail a check it can never pass).

The dmg check is deliberately not the `.app` check. A dmg is not executable code, it is a container Gatekeeper assesses on open, so it uses `spctl -a -vv -t open --context context:primary-signature`, which is what #3267 decision 3 step 4 specifies. `-t exec` would assess the wrong thing. The existing `.zip`/`.app` path is untouched: still `-t exec`, still `ditto -x -k` extraction, never `unzip`. `-vv` stays mandatory on both, since without it an accepted artifact prints nothing and merely exits 0.

**Verified.** Ran the extended script against a real unsigned dmg built with `hdiutil` on macOS 27.0. All three checks dispatch on the dmg branch, `spctl` accepts the new flag combination and returns a real assessment (`rejected`, `source=no usable signature`), and the script exits 1. The smoke test stays runnable with zero signing material.

## P1-4. The N-1 baseline predates the sentinel

**Bug.** The harness waits on `AO_E2E_UPDATE_SENTINEL`, but that listener only exists from #3294 onward. Every currently published release ignores the env var, so the first real run would wait out the full 20 minute download timeout and report a "never staged" failure that looks like a broken update.

**Choice.** Require a sentinel-capable baseline, with a fail-fast capability check. Observing staging externally would mean inferring Squirrel's internal staging state from the filesystem, which is exactly the kind of inference the sentinel exists to replace (see #3288's note on why the native event is the only trustworthy signal).

**Change.** `assertSentinelCapable` probes the packaged bundle for the env var name before anything slow runs and fails in seconds with a message naming the cause and telling you to pick a newer baseline. The probe is a substring scan of `app.asar`, because the name survives bundling as a plain string literal (`auto-updater.ts` reads it as `process.env[E2E_UPDATE_SENTINEL_ENV]`). The workflow calls the same function in its own step, so there is one implementation and an unmistakable red step in the log.

**Verified.** Behavioral tests over synthetic bundles: capable, incapable, and not-a-packaged-bundle.

## P1-5. It cannot be a stable-promotion gate as written

**Bug.** The job needs version N and `latest-mac.yml` to already be public before it can start, so real stable clients can take the update before the check finishes. The header offered it as a stable-promotion gate anyway.

**Choice.** Document it honestly as post-release monitoring and remove the gate framing. The isolated-candidate-feed option is the better end state and the mechanism exists in principle (the `AO_RELEASE_REPO` build-time override, #3267 decision 8 step 3), but pointing this job at a candidate feed requires the private signing/release pipeline to build and publish there, and that repo is explicitly out of scope for this work. Claiming a gate we do not have is worse than not having one, because it invites someone to trust it as a pre-promotion barrier.

**Change.** The header now says post-release monitoring, states what a real gate would require, and says not to re-label it before that isolated feed exists. The "to wire it as a stable-promotion gate later" snippet is gone. No release workflow is wired up or un-paused.

## P2-6. `dmg-builder` emits a `.dmg.blockmap`

**Bug.** `DmgTarget.build` does `this.options.writeUpdateInfo === false ? null : createBlockmap(...)`, so the default is on. The blockmap is computed before postMake's `sealDmg` rewrites the dmg bytes, so Forge could publish a stale macOS sidecar against the permanent full-download-only policy (#3151, #3267 decision 4).

**Change.** `writeUpdateInfo: false` in the maker config, plus a filter on the returned artifact list. That list is the array Forge hands its publisher, so filtering there is what actually keeps a sidecar off a release, and it also covers a caller re-enabling the option through `cfg.dmg`.

Linux `.blockmap` generation is untouched, per the out-of-scope list.

**Verified.** Option name and default read from the installed `dmg-builder@26.15.3`, and `packager.js` confirmed that every emitted artifact file (the blockmap included) lands in `buildResult.artifactPaths`. Tests assert both the flag and that no dmg blockmap survives in the returned artifacts.

## P2-7. State-dir overrides never reached the app

**Bug.** `--state-dir` and `--run-file` were parsed and used by the harness but never handed to the app. The direct-spawn launch passed only the sentinel; the `open -a` relaunch passed nothing deterministic. The harness seeded and polled paths the app never used, so it could only time out.

**Change.** Both launches share one environment carrying `AO_E2E_UPDATE_SENTINEL`, `AO_RUN_FILE` and `AO_DATA_DIR`. The relaunch spawns the swapped bundle's executable directly rather than going through `open`, re-reading the executable name from the swapped bundle so a rename between N-1 and N is picked up.

**Which relaunch mechanism, measured not assumed** (macOS 27.0, minimal probe bundle that writes its environment to a file):

| mechanism | sentinel | AO_RUN_FILE | AO_DATA_DIR |
|---|---|---|---|
| direct spawn | reached | reached | reached |
| `open -a` | not reached | only via the caller's ambient env | only via the caller's ambient env |
| `open --env VAR=value` | reached | reached | reached |

`open --env` works on macOS 27, but it is undocumented in `open`'s usage output on older releases, so relying on it would tie the harness to the runner image's macOS version. Direct spawn is the same mechanism the first launch already uses.

**Two path details that came out of tracing what the app actually does.** Update settings are seeded into `dirname(AO_RUN_FILE)`, because `main.ts`'s `initAutoUpdates` computes `stateDir = path.dirname(runFilePath())` and reads `update-settings.json` from there. Seeding `--state-dir` instead left the first-run dialog up whenever the two diverged, which would have blocked the whole flow. `AO_DATA_DIR` mirrors `backend/internal/config` `resolveDataDir`'s default of `<ao home>/data`.

**Run-file deletion constrained.** The run file is deleted before each launch, so `--run-file` must now be an absolute path ending in `.json` and `--state-dir` must be absolute. `rmSync` without `recursive` already refused directories, so this closes the remaining case: a typo can no longer aim that delete at something unrelated.

**Verified.** Behavioral tests over the launch environment, the derived settings and data directories, the path guards, and that `seedUpdateSettings` writes where the app will look.

## What was verified here, and what was not

Verified locally on macOS 27.0:

- `npm run frontend:typecheck` clean.
- Full frontend vitest suite: 1235 passing, 0 failing.
- The extended `verify-mac-artifact.sh` against a real unsigned `hdiutil`-built dmg (dispatch, flag acceptance, exit 1).
- The `open` vs direct-spawn environment behavior, with a purpose-built probe bundle.
- `app-builder-lib` / `dmg-builder` 26.15.3 behavior by reading the installed sources in `node_modules`, not from memory.

**Not verified here, and a reviewer should treat these as unproven:**

- **No dmg was built.** `npm run make` does not complete in this environment: it dies during electron-packager's "Finalizing package" step, and `npm run package` alone dies at the identical step on unmodified code, so this is environmental and not caused by these changes. P1-1's corrected layout and P2-6's absent blockmap are therefore covered by source reading plus unit tests against a mocked `buildForge`, not by inspecting a real image. The first real macOS build should confirm that the dmg contains one launchable `Agent Orchestrator.app` and that no `.dmg.blockmap` appears among the artifacts.
- **No signed or notarized artifact was checked.** Everything requiring Apple signing material (the accepted-artifact side of `spctl`, `notarytool`, `stapler staple`) is unexercised. The dmg `spctl` invocation was proven to parse and assess, not proven to return `accepted / source=Notarized Developer ID` on a good artifact.
- **The e2e workflow has never run.** It needs a macOS runner, a real signed baseline, and a live target feed. The new capability check and the environment propagation are covered by unit tests only.
- **P1-5 is a documentation fix, not a mechanism fix.** The job still cannot gate a stable promotion. Making it one needs an isolated candidate feed owned by the private release pipeline.

## Commits

1. `fix(makers)`: P1-1, P1-2, P2-6 (all in `maker-dmg.ts`)
2. `feat(scripts)`: P1-3
3. `fix(scripts)`: P2-7 and the harness half of P1-4
4. `docs(ci)`: P1-5 and the CI half of P1-4
5. `refactor(makers)`: drops a single-use local left over from commit 1

Closes nothing. Refs #3288, #3267, #3294.

### DIFF
--- a/.github/workflows/mac-update-e2e.yml
+++ b/.github/workflows/mac-update-e2e.yml
@@ -4,33 +4,49 @@ name: macOS update e2e
 # point it at version N's REAL published feed, and prove the update downloads,
 # stages, installs, and that the relaunched app both reports N and is alive.
 #
-# This is the check that would have caught both #3034 and the 29 Jul incident
-# before any user saw them. The interesting logic lives in
-# frontend/scripts/e2e-mac-update.mjs; this job only supplies a real macOS host
-# and a real signed baseline install.
+# This is the check that would have caught both #3034 and the 29 Jul incident.
+# The interesting logic lives in frontend/scripts/e2e-mac-update.mjs; this job
+# only supplies a real macOS host and a real signed baseline install.
 #
-# Deliberately NOT wired into any release workflow yet:
+# THIS IS POST-RELEASE MONITORING, NOT A PROMOTION GATE. Read that literally
+# before wiring it anywhere.
+#
+# The app resolves the update through the REAL published GitHub feed, so this
+# job cannot start until version N's release and its latest-mac.yml are already
+# public. By then every real stable client can already see and take that update,
+# and the run takes tens of minutes. A check that finishes after the thing it
+# guards has shipped is not a gate; calling it one would be worse than not
+# having it, because it invites someone to trust it as a pre-promotion barrier.
+# So: run it right after a release lands and treat a failure as "pull or hotfix
+# the release you just published", not "block the release you are about to".
+#
+# What it would take to make it a real pre-promotion gate: an isolated candidate
+# feed that real clients never see, with the public feed promoted only after the
+# check passes. The mechanism for that already exists in principle (the
+# AO_RELEASE_REPO build-time override, #3267 decision 8 step 3, which publishes
+# to a non-production repo), but pointing this job at one requires the private
+# signing/release pipeline to build and publish a candidate there. That pipeline
+# is a separate repo and out of scope here, so the honest description today is
+# monitoring. Do not re-label this without that isolated feed existing.
+#
+# Deliberately NOT wired into any release workflow, and not to be wired in as
+# part of this file's current shape:
 #   - The release workflows are paused, and un-pausing them is a separate
 #     decision that does not belong to this change.
 #   - macOS runner minutes are materially more expensive than Linux, and this
-#     project ships nightlies frequently. #3288 recommends making this a hard
-#     required gate on promotion to the STABLE channel specifically (the exact
-#     transition that failed in production), and treating nightly-to-nightly
-#     coverage as optional pending real per-run cost data. The "cost" step below
-#     prints the wall clock so that question can be settled with numbers.
+#     project ships nightlies frequently. #3288 recommends concentrating spend on
+#     promotion to the STABLE channel specifically (the exact transition that
+#     failed in production), and treating nightly-to-nightly coverage as optional
+#     pending real per-run cost data. The "cost" step below prints the wall clock
+#     so that question can be settled with numbers.
 #
-# To wire it as a stable-promotion gate later, call it from the release
-# workflow:
-#
-#   update-e2e:
-#     uses: ./.github/workflows/mac-update-e2e.yml
-#     with:
-#       baseline_tag: v<previous stable>
-#       target_version: <new version>
-#
-# Requires: the baseline release's signed macOS zip must be published, and the
-# target version's release + latest-mac.yml feed must already be live, since the
-# app resolves the update through the real GitHub feed.
+# Requires, both hard:
+#   - The baseline release's signed macOS zip is published, AND that baseline is
+#     new enough to carry the AO_E2E_UPDATE_SENTINEL listener (added by #3288
+#     workstream 0). Older baselines ignore the env var and can only ever time
+#     out. The harness checks this explicitly and fails in seconds; see the
+#     "Check the baseline can signal staging" step below.
+#   - The target version's release and latest-mac.yml feed are already live.
 
 on:
   workflow_dispatch:
@@ -109,6 +125,27 @@ jobs:
           ditto -x -k "$RUNNER_TEMP/baseline/agent-orchestrator-darwin-arm64.zip" /Applications
           plutil -extract CFBundleShortVersionString raw -o - \
             "/Applications/Agent Orchestrator.app/Contents/Info.plist"
+
+      - name: Check the baseline can signal staging
+        # The harness waits on AO_E2E_UPDATE_SENTINEL, and that listener only
+        # exists from #3288 workstream 0 onward. A baseline published before it
+        # ignores the env var entirely, so the run could only ever burn the full
+        # 20 minute download timeout and report a "never staged" failure that
+        # looks like a broken update. Same check the harness makes, hoisted into
+        # its own step so an incapable baseline is unmistakable in the log.
+        run: |
+          set -euo pipefail
+          node --input-type=module -e '
+            import { assertSentinelCapable } from "./frontend/scripts/e2e-mac-update.mjs";
+            const app = "/Applications/Agent Orchestrator.app";
+            try {
+              assertSentinelCapable(app);
+            } catch (err) {
+              console.error(`::error::${err.message}`);
+              process.exit(1);
+            }
+            console.log("baseline supports AO_E2E_UPDATE_SENTINEL");
+          '
 
       - name: Run the update end to end
         env:
--- a/frontend/forge.config.ts
+++ b/frontend/forge.config.ts
@@ -1,7 +1,7 @@
 import type { ForgeConfig } from "@electron-forge/shared-types";
 import { VitePlugin } from "@electron-forge/plugin-vite";
 import MakerNSIS from "./makers/maker-nsis";
-import MakerDMG, { sealDmg } from "./makers/maker-dmg";
+import MakerDMG, { sealDmg, verifyDmg } from "./makers/maker-dmg";
 import MakerAppImage from "./makers/maker-appimage";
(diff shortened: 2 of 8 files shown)


illegalcall (commented):
Review result: **changes requested**. I found two reproducible issues.

### [P1] Reject a partial App Store Connect trio when no signing identity is present

`frontend/makers/maker-dmg.ts:138-178`

The PR promises that every partial credential combination fails closed, but `resolveNotaryArgs` maps both “no notary credentials” and “incomplete API-key trio” to `undefined`. If no signing identity is present either, `sealDmg` enters the deliberate zero-credential branch, returns `false`, and causes `postMake` to skip both sealing and canonical verification.

I reproduced this directly: calling `sealDmg` with only `APPLE_API_KEY_ID` logged the unsigned no-op and resolved `false`. The added partial-trio test always supplies `APPLE_SIGNING_IDENTITY`, so it misses this direction.

Please detect whether *any* of `APPLE_API_KEY`, `APPLE_API_KEY_ID`, or `APPLE_API_ISSUER` is supplied and reject the state unless all three are present. Add coverage for a partial trio without a signing identity.

### [P2] Do not allow `--run-file` to delete an arbitrary JSON file

`frontend/scripts/e2e-mac-update.mjs:105-115,261,296`

The new guard only requires an absolute path ending in `.json`. A path such as `/Users/me/project/package.json` passes parsing and is then unconditionally deleted twice with `rmSync(..., { force: true })`. That does not provide the unrelated-file protection claimed by the comment.

Please constrain deletion to a harness-owned location/basename, or validate that an existing target is an AO run file before removing it. Add a regression test using an unrelated absolute JSON path.

### Verification

- Focused tests: 47/47 passed
- Full frontend Vitest suite: 1,235/1,235 passed
- Frontend and E2E typechecks passed
- Workflow YAML parsed successfully
- `git diff --check` passed
- Current `main` merge simulation is conflict-free

The credential-state issue is the merge-blocking concern. The run-file issue is limited to the internal macOS E2E harness but is still a real destructive-boundary gap.

somewherelostt (approved):

PR #3304, opened 2026-07-29
feat(mobile): Liquid Glass app icon matching the desktop build

Prasad-D-Ware:
## Problem

The mobile icon drew its own rounded rect and blue border ring, which iOS then masked again — double-rounded corners, a stray ring, and ~8% dead margin that made the mascot read small. It was also flat mid-blue, where the Electron icon has a vertical gradient, rim glow and a glowing wand tip.

## What changed

**iOS** now ships an Icon Composer bundle, `assets/AO.icon`, with the creature, wand and tip glow as separate layers, so iOS 26 lights each independently and the tinted / dark / clear appearances resolve correctly. Xcode flattens it for iOS 18 and older.

The background is deliberately a plain **solid** fill. iOS applies its own gradient and specular edge on device; baking either in stacks on top of the system pass and over-brightens the edge (measured: top-edge luminance 160 → 220, bottom 93 → 154).

Composition is **optically centred** on the artwork's centre of mass. Centring the bounding box leaves the creature visibly displaced, because the wand juts up-right while the creature sits low-left.

**Fixed independently of the redesign:**

- `android-icon-foreground.png` — refitted by measured content radius, so the wand tip survives the launcher mask instead of being clipped
- `android-icon-monochrome.png` — new; without it Android 13+ auto-generates a poor themed icon
- `splash-icon.png` 512² → 1024², `favicon.png` 48² → 256²

**Tooling** — assets are generated by `scripts/generate-icons.sh` from `assets/ao-logo.svg` rather than hand-edited. The source SVG is split into wand / creature layers by colour, since the trace draws both inside a single subpath and geometry can't separate them.

## Reviewing this

Use `scripts/preview-icon.sh`, which validates `icon.json` and renders all six iOS appearances via `ictool`. **Do not judge it in the simulator** — the simulator doesn't implement the Liquid Glass icon material and draws every icon flat, Apple's own included. That's documented at the top of the script, since it reliably looks like a bug.

## Verification

- Android verified from a real `expo prebuild -p android`: adaptive icon renders correctly under circle and squircle masks, themed variant included
- iOS verified via `ictool` renders and an iOS 26.5 simulator build
- `generate-icons.sh` re-run reproduces the committed assets


Vaibhaav-Tiwari (approved):
nothing is a blocker as such but i would advise to go through the pointers below for a cleaner code:


  1. Unused import — compose.py imports colorsys at the top but never actually uses it anywhere in that file. Leftover cruft.
  2. Dead function arguments — the flat_icon() function accepts background and alpha_bg as parameters, but the function body ignores both and just does its own
  thing. They do nothing — should be deleted.
  3. Outdated comment/docs — the top of compose.py describes --layers and --flat command-line flags for controlling what gets generated. Those flags don't
  actually exist in the code — the script only accepts --creature, --wand, --out and always generates everything. The docs are lying about how to use the tool.
  4. Unexplained inconsistency — almost everywhere the code sizes the logo using a named constant (LOGO_WIDTH_FRACTION = 0.630), but the splash screen uses a
  random hardcoded 0.62 instead, with no comment explaining why it's different.
  5. New dependency, not documented — this PR is the first time the repo uses Python at all (everything else is JS/TS). It requires Python 3 + the Pillow image
  library, but there's no requirements.txt, no version pinning, and no check that it's installed — if a dev doesn't have Pillow, the build script just crashes
  with a confusing error partway through. The README also wasn't updated to mention any of this.
  6. Fragile cross-package dependency — one of the new scripts (render-svg.mjs) reaches into a different package's node_modules folder
  (frontend/node_modules/playwright) instead of installing its own copy. It's intentional (avoids installing Playwright twice), but it's undocumented as a
  requirement, so it'll break confusingly if someone runs the mobile scripts without frontend set up.

PR #3313, opened 2026-07-29
fix(worktree): recover interrupted initialization husks

### LINKED ISSUE #3220: bug(worktree): spawn can leave an unborn-HEAD worktree husk under load, then stall workers spawned into it
## Summary
Under load, worktree creation during spawn can be interrupted mid-checkout, leaving a worktree directory registered/present on disk but with no completed checkout (effectively an "unborn HEAD" husk — the dir exists, `git worktree add`'s checkout step never finished). A subsequent spawn can be routed into that husk and the worker stalls with no usable working tree.

## Repro
Observed under concurrent spawn load: a worktree directory existed, but checkout had never completed, and a worker spawned into it stalled indefinitely with no usable working tree.

## Why this matters
This is silent and hard to diagnose from the worker side — the worker sees a directory that exists but isn't in a valid git state, with no clear signal that the fix is "the spawn plumbing gave you a broken worktree," not "something is wrong with your task."

## Proposed behavior
Make worktree creation atomic with respect to spawn eligibility:
- Either the worktree is fully created (checkout complete) and only then marked available for a worker to be spawned into, or
- On any failure/interruption mid-creation, the partial worktree is cleaned up immediately (removed + deregistered) rather than left behind as a candidate for a future spawn.

## Scope
Worktree creation path used by `ao spawn` (backend worktree provisioning, create-or-cleanup transaction around `git worktree add`).

Hardik180704:
## What changed

- verify a registered worktree has a valid `HEAD` before AO reuses it
- recover worktrees that Git still marks as `locked initializing`
- clean up attempt-owned initialization husks even when the spawn context was cancelled
- preserve invalid worktrees when AO cannot prove they came from an interrupted initialization

## Why

An interrupted `git worktree add` can leave a registered directory with an unusable `HEAD`. AO previously treated any registered directory as ready, so a later spawn could reuse the husk and leave its worker stalled.

Cleanup is deliberately target-specific: it does not run repository-wide prune, and it never removes a valid, pre-existing, dirty, or otherwise unproven worktree.

## Verification

- `npm run lint`
- `go test ./...`
- `go test -race ./internal/adapters/workspace/gitworktree`
- `go vet ./...`
- real-Git integration test covering interrupted initialization and successful recreation

Closes #3220


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

JAYATIAHUJA (changes requested):
Findings:

- High: `backend/internal/adapters/workspace/gitworktree/workspace.go:699` and `workspace.go:844` treat `locked initializing` plus invalid `HEAD` as enough proof that a worktree can be unlocked and force-removed. That can delete user/agent work. I reproduced this with real git: an initializing-locked worktree with corrupted `HEAD` can still contain untracked files, and the PR's `unlock` plus `worktree remove --force` sequence deletes the directory. The new integration test covers invalid `HEAD`, but not dirty/untracked content. This needs a positive cleanliness/preservation guard before force removal, or it should refuse/preserve when cleanliness cannot be established.

- Medium: `backend/internal/adapters/workspace/gitworktree/workspace.go:636` leaves `Restore` returning an existing registered worktree without checking `HEAD`. `Restore` is used for persisted sessions in `backend/internal/session_manager/manager.go:1511` and `manager.go:1617`, so a daemon restart after interrupted initialization can still hand an invalid worktree path to the runtime as restored. The new recovery is only in `Create`/failed-add cleanup, not this restore path.

Verification:

- `go test ./internal/adapters/workspace/gitworktree` passed on PR head when run outside the sandbox.
- A real-git reproduction confirmed the deletion risk above.

JAYATIAHUJA (commented):
createWorkspaceProjectRepo still bypasses the new interrupted-initialization recovery path. Normal Create/Restore now inspect a registered path, verify HEAD, and remove a locked initializing husk before retrying, but workspace-project creation goes straight from staleRegistrationForPath to ddNewBranchWorktree here. If a workspace-project root or child repo is left as a locked initializing worktree with only the .git file present, 
egisteredWorktreeDirMissing returns false because the directory exists, then git worktree add -b ... <same path> fails with atal: '<path>' already exists instead of recovering. I reproduced this against real git with the same shape used by the new tests. Please route workspace-project repo creation through the same HEAD/initializing cleanup before calling ddNewBranchWorktree, and add coverage for root/child workspace-project repos.

harshitsinghbhandari:
## Correctness review (comment only, not an approval)

Thanks for the careful fail-closed design. The core unborn/broken-HEAD husk path looks right, and the preserve guard for non-husk dirs addresses the earlier deletion concern. I still see residual correctness gaps before this fully matches "init is atomic w.r.t. spawn eligibility."

### 1. Medium: valid HEAD does not mean checkout finished

`existingWorktree` / `Restore` / workspace-project recovery only treat a registered path as broken when `rev-parse HEAD` fails. Real git can leave a still-usable HEAD while checkout is incomplete.

Repro shape:

```bash
git worktree add --no-checkout -b feature /path/to/husk main
# dir often contains only .git
git -C /path/to/husk rev-parse --verify HEAD   # succeeds
git worktree lock --reason initializing /path/to/husk
```

On this PR, that registration is treated as ready and reused. If spawn is interrupted after HEAD is written but before checkout finishes (or in any equivalent "locked initializing + partial/empty tree + valid HEAD" state), AO can still hand a worker a non-usable tree. Issue #3220 is specifically unborn/broken HEAD, so that case is covered; the broader interrupted-init class is not.

Stronger options (pick one, with tests):

- Never reuse a worktree that is still `locked` with reason `initializing`, even when HEAD resolves. Only force-remove when `ensureInitializationHusk` passes; otherwise refuse.
- Or require unlocked + HEAD ok (and maybe a stronger readiness check) before reuse.

### 2. Medium: workspace-project new-branch path skips failed-attempt cleanup

In `createWorkspaceProjectRepo`:

- existing local branch path joins `cleanupFailedInitialization` on add failure
- the common `-b` / `addNewBranchWorktree` path does not

So a cancelled mid-add on the new-branch path can leave an initializing husk until a later call hits the pre-check recovery. Main `addWorktree` cleans both paths. Please route the `-b` failure path through the same cleanup helper (and cover it with a cancelled-context test like `TestAddWorktreeCleansInitializationCreatedByFailedAttempt`).

Also: when HEAD is already valid at `repo.outputPath`, the function still falls through to `worktree add` instead of returning success. That may be pre-existing for workspace projects, but the new HEAD check only handles the error branch, not "already ready."

### 3. Low: concurrent create on the same path

`cleanupFailedInitialization` with `registrationExisted == false` can force-remove another attempt's in-progress husk if two creates race the same path (A mid-init with only `.git`, B fails add and cleans). Session IDs usually unique the path, so this is unlikely in normal spawn, but it is still a check-then-delete race if double-create is possible.

### What looks good

- Parse `LockReason` from porcelain and key off exact `initializing`
- Refuse reuse when HEAD is bad and the worktree is not a proven init husk
- `ensureInitializationHusk` (only `.git` file) before unlock + force-remove
- `WithoutCancel` + timeout so cancelled spawn context still cleans attempt-owned husks on the main `addWorktree` path
- Real-git integration coverage for recover + preserve

### Suggested bar before merge

1. Do not treat `locked initializing` as reusable solely because HEAD resolves.
2. Wire failed-attempt cleanup into the workspace-project `-b` path with a test.
3. Keep the husk-only force-remove policy (that part is correct).

Not approving: the unborn-HEAD fix is in the right direction, but I would not call init atomic w.r.t. spawn eligibility yet.

Hardik180704:
Fixed all the cases you pointed out:

- AO will not reuse a worktree while Git still says it is initializing, even if HEAD works.
- Failed workspace project setup is now cleaned up for the new branch path too.
- Two create requests for the same path cannot clean up each other's work now.
- An already ready workspace project is reused instead of adding it again.

Added tests for these cases too. The race tests, full backend tests, vet and lint are passing.

harshitsinghbhandari:
Re-verified the current HEAD: all three points from the Aug 3 review are addressed (locked-initializing is checked before the HEAD check in both Restore and existingWorktree, the -b failure path now routes through cleanupFailedInitialization, and an already-registered ready workspace repo returns early), and the earlier force-remove concern is gated by the exact .git-only husk check before every force removal. CI is green on the current HEAD. LGTM from my side; the CHANGES_REQUESTED state is stale and just needs a fresh review to clear.

i-trytoohard:
@Hardik180704 Closing as stale: quiet since early August, and the review feedback from @harshitsinghbhandari on Aug 8 is still unaddressed. The branch is safe, so reopen once those fixes are in. Thanks for the PR, @Hardik180704!

PR #3314, opened 2026-07-29
fix: split current telemetry from legacy PostHog firehose

AgentWrapper:
## Summary

- send current PostHog active, CLI, route, load, and renderer error streams under an `ao.v2.*` namespace
- stamp current telemetry with `telemetry_schema_version = 2` and preserve daemon aliases via `legacy_event_name`
- add a PostHog cost-control runbook with exact ingestion drop rules, DAU/dashboard migration queries, and verification queries

## Root cause

Read-only HogQL showed the spend is dominated by old uncapped clients: 5,598,610 events in 30 days, including 2,667,927 `ao.cli.invoked` and 2,553,297 `ao.app.active`. In the last 7 days, legacy `ao hooks` alone produced 962,837 command events and 947,731 CLI active events, with `actor_type = null`, so those cannot be cleanly separated from current telemetry by version adoption alone.

## Impact

The v2 namespace gives PostHog admins a clean current stream so they can drop legacy firehose events without hiding current DAU or current observability. The runbook keeps legacy renderer DAU and low-volume reliability events while blocking the legacy CLI automation paths responsible for the bill. Estimated reduction from the documented rules is 10x+ immediately, with current live data closer to 20x+ on the top streams.

## Validation

- `cd backend && go test ./internal/adapters/telemetry ./internal/httpd`
- `cd frontend && npm run test -- --run src/renderer/lib/telemetry.test.ts`

### DIFF
--- a/backend/internal/adapters/telemetry/posthog.go
+++ b/backend/internal/adapters/telemetry/posthog.go
@@ -24,6 +24,16 @@ import (
 
 const postHogBufferSize = 128
 const maxCommandShapeLength = 48
+const remoteTelemetrySchemaVersion = 2
+
+var remoteEventNameAliases = map[string]string{
+	"ao.app.active":              "ao.v2.app.active",
+	"ao.cli.invoked":             "ao.v2.cli.invoked",
+	"ao.renderer.route_viewed":   "ao.v2.renderer.route_viewed",
+	"ao.renderer.loaded":         "ao.v2.renderer.loaded",
+	"ao.renderer.api_error":      "ao.v2.renderer.api_error",
+	"ao.renderer.daemon_failure": "ao.v2.renderer.daemon_failure",
+}
 
 var remoteCommandTokens = map[string]struct{}{
 	"add":              {},
@@ -246,9 +256,10 @@ func (s *PostHogSink) loop() {
 }
 
 func (s *PostHogSink) send(ev ports.TelemetryEvent) {
+	eventName := remoteEventName(ev.Name)
 	body := map[string]any{
 		"api_key":     s.apiKey,
-		"event":       ev.Name,
+		"event":       eventName,
 		"distinct_id": s.distinctID,
 		"properties":  s.properties(ev),
 		"timestamp":   ev.OccurredAt.UTC().Format(time.RFC3339Nano),
@@ -279,12 +290,16 @@ func (s *PostHogSink) send(ev ports.TelemetryEvent) {
 
 func (s *PostHogSink) properties(ev ports.TelemetryEvent) map[string]any {
 	props := map[string]any{
-		"source": ev.Source,
-		"level":  string(ev.Level),
+		"source":                   ev.Source,
+		"level":                    string(ev.Level),
+		"telemetry_schema_version": remoteTelemetrySchemaVersion,
 		// The distinct ID is a random install ID with no person data behind it,
 		// so skip PostHog person-profile processing: identified events bill at
 		// several times the anonymous rate and the profiles would hold nothing.
 		"$process_person_profile": false,
+	}
+	if remoteEventName(ev.Name) != ev.Name {
+		props["legacy_event_name"] = ev.Name
 	}
 	if ev.RequestID != "" {
 		props["request_id"] = ev.RequestID
@@ -299,6 +314,13 @@ func (s *PostHogSink) properties(ev ports.TelemetryEvent) map[string]any {
 		props[k] = v
 	}
 	return props
+}
+
+func remoteEventName(name string) string {
+	if alias, ok := remoteEventNameAliases[name]; ok {
+		return alias
+	}
+	return name
 }
 
 func sanitizeRemotePayload(name string, payload map[string]any) map[string]any {
--- a/backend/internal/adapters/telemetry/posthog_test.go
+++ b/backend/internal/adapters/telemetry/posthog_test.go
@@ -178,9 +178,18 @@ func TestPostHogSinkSanitizesAppActivePayload(t *testing.T) {
 
 	select {
 	case req := <-requests:
+		if got := req["event"]; got != "ao.v2.app.active" {
+			t.Fatalf("event = %#v, want ao.v2.app.active", got)
+		}
 		props, ok := req["properties"].(map[string]any)
 		if !ok {
 			t.Fatalf("properties type = %T, want map[string]any", req["properties"])
+		}
+		if props["legacy_event_name"] != "ao.app.active" {
+			t.Fatalf("legacy_event_name = %#v, want ao.app.active", props["legacy_event_name"])
+		}
+		if props["telemetry_schema_version"] != float64(2) {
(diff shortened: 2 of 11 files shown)


AgentWrapper (commented):
Ready — no changes requested.

I reviewed the telemetry event-name migration and PostHog cost-control docs. The backend and renderer keep local/internal event names for sanitization/rate-limiting while remapping noisy streams to ao.v2.* for PostHog export, and the schema-version/person-profile safeguards look consistent.

Checks run:
- go test ./internal/adapters/telemetry ./internal/httpd
- npm --prefix frontend run typecheck

Note: npm --prefix frontend test -- telemetry.test.ts did not start in this checkout because Vitest failed loading its config with ERR_REQUIRE_ESM from std-env; typecheck passed.

somewherelostt:
Reviewed for the PostHog spend issue. No blocking code issues from me; this is okay to merge once the required approval is in.

Update: the PostHog-side controls are now enabled, and I re-tested both the code path and live PostHog data.

Verified locally:
- `go test ./internal/cli ./internal/httpd ./internal/adapters/telemetry` passes
- `npm test -- --run src/renderer/lib/telemetry.test.ts` passes: 29 renderer telemetry tests
- renderer/daemon export still maps `ao.app.active` to `ao.v2.app.active`
- routine internal CLI successes are dropped for `ao status`, `ao session ls/get`, `ao project ls/get`, `ao orchestrator ls`, `ao hooks`, and `ao pty-host`
- user/action commands such as `ao spawn` and `ao send` still emit bounded usage telemetry

Verified in live PostHog:
- AO legacy CLI firehose transformation is enabled and runs before GeoIP
- web vitals, console capture, performance capture, heatmaps, and session replay are off
- IP anonymization is on
- last-24h legacy CLI/app-active firehose is effectively quiet: only 1 `ao.cli.invoked` event, from `ao session kill`, and no old polling-command stream
- since the transformation enable time, the old AO CLI/app-active firehose query returns zero rows
- current DAU/WAU migration query still returns active-user numbers using `ao.v2.app.active` plus temporary legacy renderer/user fallbacks

Important: `ao.v2.*` is still not expected in live data yet because this PR has not been merged/released. Saved DAU/WAU insights currently still point at `ao.app.active`; that is okay before release, but after release we should migrate them to include `ao.v2.app.active` with temporary legacy renderer/user fallbacks so DAU/WAU does not drop during rollout.

Cost note: the existing bill is mostly historical usage from before the controls. The CSV shows about 5.4M events / 5.4M identified events and $500 total for June 29 - July 29, with several 300k-470k event days. Live post-control volume is around ~1.2k events/day right now, mostly `$exception`, so new daily spend should be much lower unless a new stream appears. The remaining `$exception` traffic looks like reliability/error-tracking volume, not the old CLI polling firehose.

If we still need failure visibility for internal CLI paths, add a follow-up `ao.v2.cli.failed` event with safe enum-like properties (`command_path`, `error_category`, `error_code`, version/schema), and never raw errors, stacks, local paths, repo URLs, prompts, terminal output, tokens, or request payloads.

Merge/deploy order I would use:
1. Merge this PR.
2. Release/deploy it.
3. Verify `ao.v2.*` is arriving.
4. Migrate DAU/WAU dashboards to `ao.v2.app.active` plus temporary legacy renderer/user CLI fallbacks.
5. Later switch Customer Analytics activity event from `$pageview` to `ao.v2.app.active` once live data confirms the new event is flowing.

somewherelostt (approved):

Pulkit7070:
**Post-merge notes on the telemetry split**

The two-layer drop is the right shape, and the part most likely to have been subtly wrong is correct: `ao hooks` is a single cobra command with two positional args, so `CommandPath()` returns exactly `ao hooks` and the equality check matches. Dropping before the reservoir reservation is also a good call, since blocked commands no longer consume dedup slots.

I built the daemon and pointed `AO_TELEMETRY_POSTHOG_HOST` at a local capture endpoint to observe what actually goes on the wire. Five real `ao hooks claude-code post-tool-use` invocations produced zero events, and `ao status`, `ao project ls`, `ao session ls`, `ao orchestrator ls` were all silent. A control run of `ao version` exported correctly as `ao.v2.cli.invoked` with `legacy_event_name: ao.cli.invoked` and `telemetry_schema_version: 2`, and with `AO_SESSION_ID` unset it also refreshed `ao.v2.app.active` on `channel: cli`. The rename and the drop both work for current clients.

Four things worth following up.

**`docs/telemetry.md` now asserts a state the code does not implement.** The change flips "session recording is enabled for the renderer" to "session recording is disabled by default", but `buildPostHogConfig` is untouched and sets no `disable_session_recording`, while `telemetry.ts` imports `posthog-js/dist/module.full.no-external`, the bundle carrying the replay recorder. Recording remains client-enabled and gated only on the project-side toggle. Root `README.md` was not updated either, so it still says recording is enabled, and the two docs now contradict each other. Replay bills per recording rather than per event, so it sits outside every limiter in `telemetry.ts` and is the one line item that can grow independently of event volume. Either set `disable_session_recording: true` or revert the doc line.

**The "10x+ immediately" estimate is not something this change delivers.** The 962,837 weekly `ao hooks` events come from already-deployed clients, and nothing in the diff reduces their volume. The reduction depends entirely on ingestion drop rules configured in the PostHog UI, which have no representation in the repo, so there is no way to tell from the code whether they were ever enabled. Reading the merge, it looks like the spend problem is closed. Suggest separating what shipped from what still requires manual PostHog configuration, and filing a tracking issue for the latter, otherwise the runbook's estimate reads as a result.

**`isRoutineInternalCLICommand` is duplicated verbatim in `backend/internal/cli/root.go` and `backend/internal/httpd/router.go` with nothing asserting the two lists agree.** Divergence is silent in both directions: adding a command only to the CLI list leaves old clients leaking it, and adding it only to the daemon list drops a command the CLI still counts. Given the deliberate no-import boundary between the two packages, a small shared list would be safer than a copy.

**The gate is stricter than the threat it defends against.** Posting variant command paths to `/internal/telemetry/cli-invoked` as an old or foreign client would, three of six got through: `"ao  hooks"` with a double space, `"AO HOOKS"`, and `"ao hooks claude-code post-tool-use"`. They were then classified `actor_type: user` and so counted as active-user signal, with `command_path` arriving hashed as `sha256:...`. No real client produces those forms and the endpoint is loopback-only, so this is not current volume, but the daemon-side gate exists specifically for clients you do not control. The first two close with a one-line change in both copies:

```go
switch strings.ToLower(strings.Join(strings.Fields(commandPath), " ")) {
```

Related: `cliActorType` falls through to `user` for any unrecognized path, so an unknown command is DAU-eligible by default. For a cost-control change, a conservative fallback would fit better.

**Minor:** the rewrite of `TestCLIInvokedRouteSeparatesAgentAndSystemInvocationsFromActiveUsers` dropped the only coverage of the `"ao hooks" -> "agent"` inference in `cliActorType`. That branch is now unreachable in effect, since `isRoutineInternalCLICommand` returns first, and no test would fail if it were removed.

PR #3324, opened 2026-07-29
fix: improve project removal and session termination UI

### LINKED ISSUE #3318: Termination confirmation modal traps the user (non-dismissable) for the entire synchronous Kill()
## Summary

When terminating a session, the confirmation dialog (`SessionTerminationDialog`) turns into a **non-dismissable modal** that stays open for the entire duration of the backend `Kill()`. The user is trapped staring at a frozen "Terminating…" dialog until the session has been fully torn down on the backend, with no way to cancel or close.

## Repro
1. Open a project board with an active session.
2. Click the **Terminate** (trash) button on a session card.
3. In the confirmation dialog, click **Terminate session**.
4. The dialog switches to a busy "Terminating…" state. **All controls are disabled** (the close `X`, the `Cancel` button, and the confirm button), and Esc / outside-click cannot dismiss it.
5. The modal stays open until the backend has finished destroying the runtime, closing shell terminals, tearing down the worktree, and marking the session terminated — only then does it close.

## Why it happens

**Frontend — the modal is hard-locked while the kill is in flight:**
- `frontend/src/renderer/components/SessionsBoard.tsx:427` renders `SessionTerminationDialog` with `busy={terminateSession.isPending}`.
- `SessionTerminationDialog` → `ConfirmDialog` (`frontend/src/renderer/components/ConfirmDialog.tsx`). The `busy` prop disables **every** button: close `X` (L48), `Cancel` (L74), and confirm (L84).
- The `onOpenChange` handler (`SessionsBoard.tsx:432`) only clears the dialog state when `!terminateSession.isPending` — so even the Radix overlay close attempts are suppressed while pending.
- Result: while `POST /api/v1/sessions/{id}/kill` is in flight there is **no** way to dismiss the modal.

**Backend — `Kill()` is fully synchronous within the HTTP request:**
- `Service.Kill` → `Manager.Kill` (`backend/internal/session_manager/manager.go:803`) does, in a single request:
  - `runtime.Destroy(ctx, handle)` (L824)
  - shell-terminal teardown gate (L835)
  - workspace-project row destruction / `workspace.Destroy` (L856–873)
  - `lcm.MarkTerminated` + cleanup
- The HTTP response does not return until all of that completes, so the mutation stays `isPending` (and the modal locked) for the entire teardown.

The `SessionInspector.tsx:380` path (terminate from the inspector) has the same `ConfirmDialog` + synchronous-mutation lock.

## Impact
Session **termination is a frequent, primary action**, and trapping the user behind a frozen, undissmisable modal every time — for what can be several seconds of teardown — is poor UX. Functionally sessions terminate correctly; this is a blocking-UX problem.

## Suggested direction (not prescriptive)
Make termination **optimistic / non-blocking**: close the modal immediately on confirm and reflect the in-flight state inline on the card (e.g. dim the card + a "Terminating…" badge). The user should never be locked behind a modal for a backend teardown. Alternatively, allow the modal to be dismissed while pending and surface progress elsewhere (e.g. the NotificationCenter / card state).

## Environment
Desktop app (Electron + React renderer), `main` @ `ffecf1ba`.

nikhilachale:
Fixes #3318

## Summary
- replace the centered session termination modal with a compact UI-matched confirmation card anchored below each kill control, with No/Yes actions
- dismiss destructive confirmations immediately and navigate to a stable destination while teardown continues
- return to the projects home card view after confirming project removal, with sidebar progress/error feedback
- scope termination pending and error state to the exact worker session ID, including rapid worker switching
- preserve termination progress and failures on the destination orchestrator topbar or originating board card
- add focused frontend coverage for confirmation placement, navigation, pending-state isolation, and cleanup/error behavior

## Verification
- `npm --prefix frontend test -- src/renderer/components/ConfirmDialog.test.tsx src/renderer/components/Sidebar.test.tsx src/renderer/components/ShellTopbar.test.tsx src/renderer/components/SessionsBoard.test.tsx src/renderer/components/SessionInspector.test.tsx` (150 passed)

## Caveats
- Manual UI verification is left to the requester; no local-browser testing was performed.
- The full renderer suite was not completed locally; the focused suites covering every touched component passed.


frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -468,7 +468,7 @@ function ProjectItem({
 			await onRemoveProject(workspace.id);
 			setConfirmOpen(false);
 			// The route for a removed project no longer resolves; fall back home.
-			if (selection.activeProjectId === workspace.id) selection.goHome();
+			selection.goHome();
```
nikhilachale: **[P1] Project removal still traps the user for the full teardown.** `handleConfirmRemove` still awaits `onRemoveProject` before closing, while `busy={isRemoving}` disables every dialog control and `onOpenChange` rejects dismissal. For a project with multiple sessions this preserves the exact synchronous, non-dismissable flow called out in #3318; this change only navigates home after it finally completes. Please dismiss optimistically and surface removal progress/errors outside the modal.

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -425,11 +425,14 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
...
 			)}
 			<SessionTerminationDialog
-				busy={terminateSession.isPending}
-				error={terminateSession.error instanceof Error ? terminateSession.error.message : null}
-				onConfirm={() => terminationSession && terminateSession.mutate(terminationSession)}
+				onConfirm={() => {
+					if (!terminationSession) return;
+					const target = terminationSession;
+					setTerminationSession(undefined);
+					terminateSession.mutate(target);
```
nikhilachale: **[P1] Preserve pending/error state after dismissing this dialog.** Once `terminationSession` is cleared, the board renders neither `terminateSession.isPending` nor `terminateSession.error`; the unchanged card remains fully enabled until the synchronous request returns. Users can therefore reopen the dialog and issue concurrent kill requests for the same session, and a failed request is completely silent. Please key the inline card/board state by session ID (as the topbar does), disable repeat termination while pending, and expose failures.

frontend/src/renderer/components/ShellTopbar.tsx
```diff
@@ -298,43 +297,37 @@ export function TopbarKillButton({
 }) {
 	const [confirmOpen, setConfirmOpen] = useState(false);
 	const kill = useTerminateSession();
-	const error = kill.error instanceof Error ? kill.error.message : null;
+	const { error, isPending } = useTerminateSessionState(session.id);
+
+	const confirmKill = () => {
+		setConfirmOpen(false);
+		kill.mutate(session);
+		onKilled(session.workspaceId, orchestratorId);
```
nikhilachale, line 312: **[P2] Do not navigate away from the only failure surface without relocating it.** `onKilled` now runs before the kill succeeds. The destination topbar observes the orchestrator ID (or the board renders no terminate error at all), while the mutation error is keyed to the worker ID, so a failed kill becomes invisible and the user is left believing termination succeeded. The inspector path has the same unconditional navigation. Immediate navigation is fine, but progress/failure needs to be surfaced at the destination (for example a notification or project-scoped status).

nikhilachale (commented):

The focused renderer suites pass (147 tests), and session confirmations now dismiss promptly. However, project removal still has the original blocking behavior, and the non-blocking session flows drop essential pending/error feedback.

frontend/src/renderer/components/Sidebar.tsx
nikhilachale: Addressed in 0e43a5a0a. Project removal now closes the confirmation and navigates home before awaiting teardown. The sidebar project row remains mounted to show a non-modal “Removing…” status, keeps duplicate removal disabled, and surfaces any failure inline for retry.

frontend/src/renderer/components/SessionsBoard.tsx
nikhilachale: Addressed in 0e43a5a0a. Each board card now reads termination state from the shared mutation cache by exact session ID. Only the targeted card shows a visible disabled Killing control; other workers stay enabled. Failures render on the affected card, and retry clears only that session’s stale result.

frontend/src/renderer/components/ShellTopbar.tsx
nikhilachale: Addressed in 0e43a5a0a. The project orchestrator topbar now observes project termination mutations and shows the originating worker’s pending/error feedback after immediate navigation. The board fallback shows the same state on the originating card, covering both topbar and inspector initiation paths.

nikhilachale:
Addressed requested changes from review 4813710563 in `0e43a5a0a`: project removal is now non-blocking with sidebar progress/error feedback; board termination pending/error state is keyed per worker; and termination progress/failures remain visible at the orchestrator or board navigation destination. Focused frontend suites pass (150 tests), typecheck passes, and the renderer production build passes.

nikhilachale (commented):
## Review verdict: Approved

The follow-up resolves the prior findings: project removal now dismisses immediately with persistent sidebar progress/error feedback, board termination is keyed per session and prevents duplicate requests, and termination progress/failures remain visible after navigation. I found no remaining correctness issues in the updated diff.

Verification: focused renderer suites pass (150 tests), and `npm run frontend:typecheck` passes.

nikhilachale:
Updated the termination confirmation UI in `bcba50178`: it is now a small popover card anchored directly below the kill/terminate control, using the existing popover surface and compact No/Yes actions. The per-worker pending/error and navigation behavior is unchanged. Focused tests (150), typecheck, and renderer build pass.

nikhilachale (commented):

The anchored confirmation behavior works in the focused suites (150 tests), and typecheck passes. One viewport-placement regression remains: confirmations for cards near the bottom edge can render their actions off-screen.

frontend/src/renderer/components/SessionTerminationPopover.tsx
```diff
@@ -0,0 +1,55 @@
...
+	session?: WorkspaceSession;
+	trigger: ReactElement;
+}) {
+	return (
+		<Popover onOpenChange={onOpenChange} open={open}>
+			<PopoverTrigger asChild>{trigger}</PopoverTrigger>
+			<PopoverContent
+				align="end"
+				aria-label={`Terminate ${session?.title ?? "session"}?`}
+				avoidCollisions={false}
```
nikhilachale: **[P1] Keep collision handling enabled for board cards near the viewport bottom.** This popover is also used by terminate buttons on vertically scrolling board cards. With `avoidCollisions={false}`, Radix disables both its shift and flip middleware, so a trigger in the bottom ~130px of the window positions the confirmation below the viewport and the No/Yes controls become unreachable. This is common for the last card in a lane, which cannot necessarily be scrolled farther upward. Please retain collision handling (the normal case will still use `side=bottom`, while constrained cases can shift or flip) and add an edge-placement test if practical.

nikhilachale: Addressed in 43b94c073. Removed the `avoidCollisions={false}` override so Radix shift/flip positioning is enabled again, and set `collisionPadding={8}` to keep the compact confirmation and its No/Yes actions inside the viewport. The existing placement test still verifies the unconstrained case opens below the kill button.

nikhilachale:
Addressed review 4813885275 in `43b94c073`: restored Radix collision-aware shift/flip placement and added an 8px collision boundary, preventing bottom-edge confirmation actions from rendering off-screen. Focused tests, typecheck, and renderer build pass.

nikhilachale (commented):
## Review verdict: Approved

The follow-up restores Radix collision-aware shift/flip behavior and adds an 8px viewport boundary, resolving the bottom-edge placement issue without changing the preferred below-trigger placement. I found no remaining correctness issues.

Verification: focused renderer suites pass (150 tests), and `npm run frontend:typecheck` passes.

Pulkit7070 (commented):
This is a clean reworking of the termination flow. Tracking termination state in the react-query mutation cache keyed by session id is the right call: it survives the navigation this change introduces, isolates each worker's Killing/error state so rapid switching between workers cannot bleed one session's state into another's topbar, and lets the pending indicator and any failure follow to wherever the user lands. The success path invalidates the workspace and drops the session from the aggregate, so a completed termination does not linger, and clearTerminateSessionState only removes settled mutations so it never drops an in-flight one. The destination logic is right too: the in-session topbar navigates to the orchestrator, or falls back to the project board when there is no orchestrator, while the board card kill stays in place and just shows progress. The popover copy matches the backend: a user kill preserves a dirty worktree rather than force-removing it, so "Changes are preserved" is accurate.

One behavior worth confirming, since it is a deliberate trade-off rather than a bug. Confirming a kill from the in-session topbar navigates immediately, before the kill resolves. If the kill then fails, the user has already been moved to the orchestrator (or board) with the worker still alive, and the failure surfaces at that destination rather than where they acted, with no in-place retry from the session they were on. That is consistent with the dismiss-and-navigate-while-teardown-continues design and the failure is not lost, but please confirm that landing a failed termination on the destination surface, rather than keeping the user on the session to retry, is the intended experience.

Pulkit7070 (approved):

Pulkit7070:
Fallback there lgtm

nikhilachale:
If termination fails:                                                  
                                                                         
  - The confirmation card closes immediately.                            
  - The worker is not marked terminated.                                 
  - Its Kill control becomes enabled again.                              
  - The daemon’s error appears:                                          
      - On the worker’s Kanban card when returning to the board.         
      - In the orchestrator topbar after navigation.                     
                                                                         
  - Retrying Kill clears that worker’s previous error.                   
  - Other workers remain unaffected—the error and state stay scoped to   
    the failed worker session.

PR #3328, opened 2026-07-30
fix(tmux): stop burning the reap grace where pkill cannot match a session

Pulkit7070:
## Problem

Closing a terminal took about 5 seconds. The delay is on the daemon: `Destroy` reaps a pane's leftover background processes synchronously inside the shell-terminal DELETE handler, waiting a 5s grace between SIGTERM and SIGKILL so a backgrounded dev server can release its ports (issue #2523).

That reap uses `pkill -s <sid>` / `pgrep -s <sid>`. **`-s` is a Linux procps extension.** BSD pkill and pgrep reject it:

```
$ pkill -TERM -s 99999; echo $?
-TERM: illegal option -- s
2
$ pgrep -s 99999; echo $?
pgrep: illegal option -- s
2
```

So on macOS every signal and every probe failed. `sessionsHaveProcesses` treats any non-1 exit as "assume survivors", so the probe always reported survivors, the wait always ran to the end, and the SIGKILL that followed was equally impossible. The reap accomplished nothing and cost the full grace on every terminal close and every session destroy.

## Change

- Skip the reap when pkill cannot match by session id at all, rather than waiting out a grace that cannot accomplish anything. Exit 2 is a usage error on both procps and BSD, so it is a portable signal for "this matcher does not exist here".
- Poll for survivors during the grace instead of sleeping through it, so on platforms where `-s` does work, a pane that empties early returns immediately instead of always burning the full 5s.

A process that genuinely needs the time still gets the full grace before SIGKILL, so #2523's behavior on Linux is unchanged.

## Verification

Measured end to end against a local daemon on macOS (`ao daemon` against a scratch `AO_DATA_DIR`, open a shell terminal over the API, time the DELETE):

| | `DELETE /api/v1/shell-terminals/{handleId}` |
|---|---|
| before | 5.06s |
| after | 0.04s |

New unit tests cover the three paths: returns early once the session is empty, still SIGKILLs survivors after the full grace, and does not wait at all when the matcher is unsupported. `go build ./...` is clean and the runtime, shellterm, and httpd package tests pass.

## Follow-up

This leaves macOS without orphan reaping. That is what it had in practice already, since none of the signals were landing, but it means #2523's protection has only ever worked on Linux. Doing it properly on macOS needs a different matcher (session id is not something BSD pkill can match, and process group does not cover a job the shell backgrounded) and is worth its own issue.

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -31,8 +31,15 @@ const (
 	defaultEnterDelay = 300 * time.Millisecond
 	// defaultReapGrace is how long Destroy waits between SIGTERM and SIGKILL when
 	// reaping a pane's leftover background processes, giving them a chance to
-	// exit cleanly (release ports) before being forced (issue #2523).
+	// exit cleanly (release ports) before being forced (issue #2523). It is a
+	// ceiling, not a fixed wait: reapPollInterval decides how soon a pane that
+	// is already empty lets Destroy return.
 	defaultReapGrace = 5 * time.Second
+	// reapPollInterval is how often the reap rechecks for survivors while the
+	// grace runs. A plain shell exits within a tick or two, so Destroy returns
+	// in roughly this long instead of always burning the full grace — which the
+	// DELETE handler blocks on, and the user sees as a tab that will not close.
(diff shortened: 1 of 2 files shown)


nikhilachale (approved):
LGTM ,I found no blocking correctness issues.

PR #3329, opened 2026-07-30
fix(terminal): number a session's terminal tabs and retire them on exit

Pulkit7070:
Stacked on https://github.com/Untrivial-ai/agent-orchestrator/pull/3298 and targets its branch, so the diff here is only the tab naming and exit handling.

## Problem

**Naming.** Every shell in a session starts in that session's worktree, so `filepath.Base(workingDir)` put the identical label on every tab: three terminals in one session all read `ao-22`. The function's own comment claimed the directory "is the only thing that distinguishes one shell pane from another in the UI", which is exactly what it failed to do.

**Exit.** A shell whose PTY ended left a tab that could never be attached again, showing "TERMINAL ENDED" until something else happened to refetch. The daemon already prunes dead shells on list, but nothing told the client — `useShellTerminals.ts` says so outright: "shell terminals only change when this client opens or closes one". They also change when the shell exits on its own.

**Preview mock.** It stamped the project name on every tab, which the daemon never produces. That is why the naming bug looked different from what it actually was, and it makes the browser dev build misleading for anyone demoing from it.

## Change

- A session's tabs are numbered: `Terminal 1`, `Terminal 2`. The session is already the first tab in the strip, so repeating its worktree name on every terminal added nothing.
- Standalone shells (board, `/terminals`) keep the directory name. Those can span projects, so the directory is what tells them apart there.
- The number is **one past the highest still in use**, not a count, so closing `Terminal 1` while `Terminal 2` is open does not hand the next tab a name already on screen. A tab the user renamed no longer parses and stops reserving its number.
- A shell whose PTY exits retires its tab through the existing close path. Shells only: a session pane that ends still has a row, a status, and a restore path, so its tab stays.
- The preview mock now mirrors the daemon on both counts.

## Verification

Driven against a real daemon (isolated `HOME`, scratch data dir, own port) with a real session spawned via `--harness fake`:

```
### session terminals are numbered
  open -> Terminal 1
  open -> Terminal 2
  open -> Terminal 3
  strip: ['Terminal 1', 'Terminal 2', 'Terminal 3']

### close Terminal 1, open again -> must not reissue a live name
  after close: ['Terminal 2', 'Terminal 3']
  next open -> Terminal 4

### standalone open in the same daemon still uses the directory
  -> data
  strip: ['Terminal 2', 'Terminal 3', 'Terminal 4', 'data']
```

Also driven against a real daemon: a shell whose tmux session is killed out from under it disappears from `GET /shell-terminals` (`['data']` then `[]`), which is the daemon half of the exit path.

Renderer driven in a browser: two clicks on `+` give `Terminal 1` and `Terminal 2`, no tab is named after the project, right click produces no picker, and the session tab has no close button. Closing `Terminal 1` and reopening gives `Terminal 3`.

## Not verified by running

The renderer's retire-on-exit needs a live PTY, which only exists in the Electron app. The daemon-side prune and the renderer's reporting logic are each covered, but the two have not been exercised together end to end. Worth a manual check by whoever reviews this.

## Known limitation

Only the attached pane notices an exit. If a background tab's shell exits while you are on another tab, that tab lingers until you select it or open/close something. Fixing it properly wants a daemon push when a PTY dies rather than polling the list.

### DIFF
--- a/backend/internal/service/shellterm/service.go
+++ b/backend/internal/service/shellterm/service.go
@@ -245,6 +245,17 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		return ShellTerminal{}, fmt.Errorf("open shell terminal %s: runtime: %w", handleID, err)
 	}
 
+	title := shellTerminalTitle(workingDir)
+	if in.SessionID != "" {
+		// Safe to count here: the session gate is held for the whole open, so
+		// two concurrent opens for one session cannot pick the same number.
+		ordinal, err := s.nextSessionTerminalOrdinal(ctx, in.SessionID)
+		if err != nil {
+			return ShellTerminal{}, err
+		}
+		title = sessionShellTerminalTitle(ordinal)
+	}
+
 	rec := ShellTerminalRecord{
 		HandleID: handle.ID,
 		// The resolved project, not the requested one: a session-scoped open
@@ -253,7 +264,7 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		ProjectID:  projectID,
 		SessionID:  in.SessionID,
 		WorkingDir: workingDir,
-		Title:      shellTerminalTitle(workingDir),
+		Title:      title,
 		AppRunID:   s.appRunID,
 		CreatedAt:  s.now().UTC(),
 	}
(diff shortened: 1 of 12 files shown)


frontend/src/renderer/components/TerminalPane.tsx
```diff
@@ -275,6 +282,20 @@ function AttachedTerminal({ session, theme, daemonReady, terminalTarget, fontSiz
 		session !== undefined &&
 		!isSessionActive;
 
+	// A shell whose PTY exits leaves a tab that can never be attached again, so
+	// retire it instead of parking a dead pane in the strip. Reported once per
+	// pane: the component is keyed by handle, so a later shell on the same tab
+	// mounts fresh. Only for shells — a session pane that ends still has a row,
+	// a status, and a restore path, so its tab must stay.
+	const reportedShellExitRef = useRef(false);
+	useEffect(() => {
```
nikhilachale, line 319: [P1] An attachment failure can destroy a still-running shell.

it treats every exited state as proof that the underlying shell died and invokes the DELETE close path. However, the terminal backend also emits exited after repeated liveness-probe or attachment errors—even when runtime death was never confirmed. A transient/unknown probe failure can therefore make the renderer call DELETE, which explicitly destroys the potentially live shell and any running command.

nikhilachale (changes requested):

backend/internal/service/shellterm/service.go
```diff
@@ -245,6 +245,17 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		return ShellTerminal{}, fmt.Errorf("open shell terminal %s: runtime: %w", handleID, err)
 	}
 
+	title := shellTerminalTitle(workingDir)
+	if in.SessionID != "" {
+		// Safe to count here: the session gate is held for the whole open, so
+		// two concurrent opens for one session cannot pick the same number.
+		ordinal, err := s.nextSessionTerminalOrdinal(ctx, in.SessionID)
+		if err != nil {
+			return ShellTerminal{}, err
```
nikhilachale: > **[P2] Roll back the runtime when ordinal lookup fails**
>
> At this point `runtime.Create` has already succeeded, but no tracking row has been inserted. Returning directly when `nextSessionTerminalOrdinal` fails leaves an untracked tmux/PTY runtime that shutdown and reaping cannot discover. Please either calculate the ordinal before `runtime.Create`, or destroy `handle` here using the same `context.WithoutCancel(ctx)` rollback used for insert failures. A store-error regression test should also assert that `Destroy` is called.

frontend/src/renderer/components/ShellTerminalsView.tsx
```diff
@@ -107,6 +113,7 @@ export function ShellTerminalsView() {
 					<TerminalPane
 						daemonReady={daemonStatus.state === "ready"}
 						fontSize={12}
+						onShellExited={() => refreshShellTerminals()}
```
nikhilachale: > **[P2] Remount the pane when the daemon retains a shell after an exit hint**
>
> If `exited` resulted from attachment/probe failures while the shell is still alive, this refresh returns the same shell. Because `TerminalPane` remains mounted with the same handle, its `useTerminalSession` state stays `exited` and it never attempts another attachment. The UI remains stuck on “Terminal ended” and advises the user to close a live shell. Please force a fresh pane attachment when the refreshed list retains this handle—similar to how `SessionView` switches away from the shell—and add coverage for the false-exit/live-shell path.

PR #3336, opened 2026-07-30
feat: add inline browser text edits

Vaibhaav-Tiwari:
## Summary

  - Added source-backed inline text editing for the in-app browser.
  - Added POST /api/v1/sessions/{sessionId}/workspace/text-edit to apply browser-selected text edits back into the session workspace.
  - Added browser text-edit mode in the UI, with preload capture/edit overlay and renderer handling for applied, ambiguous, conflict, unsupported, and not-found
    results.

  - Added safety gates so edits only work for the session’s local preview, skip generated/vendor/build files, reject weak/unchanged selections, and protect
    against stale or shifted ambiguous selections.

  - Added tests across backend service/controller and frontend browser panel, hook, preload, and main browser host behavior.

  ## Tests

  - go test ./internal/cli -run TestE2E_SpawnAndProjectAddDTORoundTrip
  - go test ./internal/service/session ./internal/httpd/controllers ./internal/httpd/...
  - npm run frontend:typecheck
  - npm run test -- src/renderer/components/BrowserPanel.test.tsx src/renderer/hooks/useBrowserView.test.tsx src/main/browser-view-host.test.ts src/
    preload.test.ts src/annotate-preload.test.ts

  - git diff --check origin/main..HEAD


backend/internal/service/session/workspace_files.go
```diff
@@ -299,6 +429,363 @@ func scratchWorkspaceFile(root string, id domain.SessionID, rel string) (Workspa
...
+	line := workspaceTextLine(content, index)
+	snippet := workspaceTextSnippet(content, index, len(oldText))
+	if (candidate.Line > 0 && candidate.Line != line) || (candidate.Snippet != "" && candidate.Snippet != snippet) {
+		result.Status = WorkspaceTextEditConflict
+		result.Message = "Selected source text no longer matches"
+		return result, nil
+	}
+	result.Line = line
+	updated := content[:index] + newText + content[index+len(oldText):]
+	if err := writeWorkspaceTextFile(file, updated, info.Mode().Perm()); err != nil {
```
aprv10: **[P1] Prevent this write from overwriting concurrent agent saves**

`content` is a full-file snapshot, so an agent save that lands after the read above but before `writeWorkspaceTextFile` replaces the file is silently discarded instead of returning `conflict`. Concurrent writes are a normal case in an active worker workspace. Please make the replacement conditional on the file still matching the snapshot, or otherwise coordinate the write, before replacing the source file.

frontend/src/main/browser-view-host.ts
```diff
@@ -663,7 +678,32 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
...
+			enabled: input.enabled,
+		});
+	};
+
+	const setTextEditMode = (event: IpcMainInvokeEvent, input: BrowserTextEditModeInput): void => {
+		if (!isRendererOwned(event, input.viewId)) return;
+		const session = entries.get(input.viewId);
+		if (!session) return;
+		const entry = activeEntry(session);
+		entry.textEditEnabled = input.enabled;
```
aprv10, line 716: **[P2] Clear text-edit mode when changing browser tabs**

`textEditEnabled` belongs to the active tab entry, but `activateTab` never cancels it on the previous entry. Repro: enable edit mode on tab A, switch to B, cancel there, then return to A; A's preload overlay is still active and intercepts clicks while the renderer says the mode is off. Please cancel and notify the previous tab during activation, and cover this transition in the host test.

aprv10 (changes requested):
The inline-edit flow has two correctness issues that can either discard concurrent workspace changes or leave a hidden tab intercepting clicks.

Hardik180704 (changes requested):
There are two blockers before this is ready to merge.

First, the branch currently conflicts with `main`, so it needs to be updated and the generated API artifacts revalidated.

Second, `writeWorkspaceTextFileIfUnchanged` still has a check-then-rename race. It rereads the source and compares it with the snapshot, but an active agent can save after that comparison and before `os.Rename`; the rename would then replace the agent's newer content. Please coordinate edits per workspace file (or use another mechanism that closes that window) and add a deterministic concurrent-save regression test.

i-trytoohard:
From the NotesBot call on `2026-07-31`, this existing pr appears to be relevant.
Match confidence: `0.95` (fuzzy).

Relevant discussion excerpts:
- - *prateek:* asked when the file-related issue would be fixed so committed changes and PR diffs would be visible in the UI. - *.phylolver(vaibhaav):* said an approved PR should be merged after resolving conflicts, with 4–5 changes expected by **before 12 AM IST**. - *.phylolver(vaibhaav):* said browser work [...]
- [29:51] .phylolver(vaibhaav): we can— it's not necessary, I think. [29:58] prateek: Also, uh, how can we make this box even simpler? [30:23] .phylolver(vaibhaav): I was saying that I've also added inline text edit. So apart from the browser annotation, you can just highlight the text that you want to change and [...]
- [29:58] prateek: Also, uh, how can we make this box even simpler? [30:23] .phylolver(vaibhaav): I was saying that I've also added inline text edit. So apart from the browser annotation, you can just highlight the text that you want to change and you can edit that directly on the browser.

Source thread: https://discord.com/channels/1476302178913357958/1532816669167063130/1532816669167063130

Vaibhaav-Tiwari:
> There are two blockers before this is ready to merge.
> 
> First, the branch currently conflicts with `main`, so it needs to be updated and the generated API artifacts revalidated.
> 
> Second, `writeWorkspaceTextFileIfUnchanged` still has a check-then-rename race. It rereads the source and compares it with the snapshot, but an active agent can save after that comparison and before `os.Rename`; the rename would then replace the agent's newer content. Please coordinate edits per workspace file (or use another mechanism that closes that window) and add a deterministic concurrent-save regression test.

please see again, i have made the changes

PR #3345, opened 2026-07-30
feat: add --model flag to ao spawn for per-session model override (#3331)

### LINKED ISSUE #3331: feat(spawn): add --model to ao spawn for per-session model override
## What is missing

`ao spawn` can select the harness per session but not the model. Full flag list today:

```
--branch, --claim-pr, --harness/--agent, --issue, --kind, --name,
--no-takeover, --project, --prompt, --skip-agent-check
```

Model is only settable at project scope, via `ao project set-config <id> --model <model>`, which resolves when a session spawns and therefore applies to every session spawned afterward.

## Why this matters

The orchestrating use case wants model per task, not per project. A batch of parallel workers is rarely homogeneous: a two-file CI-config fix and a security-sensitive auth path do not deserve the same model. Since `--harness` is already per-spawn, `--model` is its natural sibling, and today an orchestrator has no way to express "this worker is cheap, that one is not" at spawn time.

Concretely, while orchestrating a multi-batch build I wanted three workers on Sonnet and could only get there by mutating project config, which also silently repointed future orchestrator sessions in the same project.

## Why the current workaround is bad

1. **It is global.** `--model` at project scope hits workers and orchestrators alike unless you hand-write role-scoped `agentConfig` blocks.
2. **It is racy for mixed batches.** Building a mixed-model batch means calling `set-config --model` between spawns. Config is read at spawn time, so interleaved or concurrent spawns can pick up whichever value happens to be current.
3. **It leaves persistent state.** The project is left sitting in whatever model the last spawn needed, which then silently applies to unrelated future sessions.
4. **`set-config` replaces rather than patches.** Callers must re-pass unrelated fields (`--default-branch`, `--worker-agent`, `--orchestrator-agent`) alongside `--model` or drop them. That makes the workaround easy to get wrong in a script.

## Proposed behavior

Add `--model` to `ao spawn`, scoped to the single session being spawned, with no persistent effect on project config.

Suggested precedence, extending the existing chain:

```
spawn --model  >  role agentConfig.model  >  project agentConfig.model  >  user/global default
```

That is a fourth, narrowest layer on the resolution order and should not require changes to how the existing layers merge.

Notes:

- Validation should reject a model the selected harness cannot honor, rather than silently falling back. Several harnesses already have known gaps here (see related issues below), so `--model` should fail loudly on an unsupported combination instead of adding a new silent-swap path.
- `ao session get` should report the model a session actually resolved to, so an orchestrator can verify what it got rather than inferring it from project config.

## Related

- #2999 `feat(config): User-scope merge layer, User to Project to role in effectiveAgentConfig`. Closest neighbor. That issue adds a layer above project scope; this one adds a layer below role scope. Not blocked on it, though the two should agree on precedence.
- #3188 `Per-role Codex reasoning effort, plus a model/reasoning picker in project settings`. Same theme of finer-grained model selection, but project-settings-scoped rather than spawn-scoped.
- #3218 `restore silently swaps an unrecognized model instead of refusing or prompting`. Relevant to the validation note above; `--model` should not add a second silent-swap surface.
- The cluster of "X ignores role-specific model config on launch" issues (#2883, #2884, #2886, #2887, #2891, #2898, #2899, #2902) is why per-harness validation matters here.

## Environment

`ao version`: `dev`

praneshnikhar:
Adds a `--model` flag to `ao spawn` that lets users override the agent model for a specific session (e.g. `--model claude-sonnet-4-20250514`). The model is persisted on the session record so `ao session get` reports what the session resolved to. The resolved model flows into the agent's effectiveAgentConfig, where it takes precedence over project and role config.

Closes #3331

### Changes
- New migration 0039_add_session_model.sql
- sqlc regenerated
- domain.SessionRecord.Model, ports.SpawnConfig.Model
- CLI --model flag threaded through spawnRequest → daemon → controller → SpawnConfig → session_manager
- Resolved model stored on the record after effectiveAgentConfig override
- OpenAPI spec and frontend TS schema regenerated

### DIFF
--- a/backend/internal/cli/spawn.go
+++ b/backend/internal/cli/spawn.go
@@ -31,6 +31,7 @@ type spawnOptions struct {
 	claimPR        string
 	noTakeover     bool
 	skipAgentCheck bool
+	model          string
 }
 
 // spawnRequest mirrors the daemon's SpawnSessionRequest body for
@@ -43,6 +44,7 @@ type spawnRequest struct {
 	Branch      string `json:"branch,omitempty"`
 	Prompt      string `json:"prompt,omitempty"`
 	DisplayName string `json:"displayName"`
+	Model       string `json:"model,omitempty"`
 }
 
 type spawnResult struct {
@@ -126,6 +128,7 @@ func newSpawnCommand(ctx *commandContext) *cobra.Command {
 				Branch:      opts.branch,
 				Prompt:      opts.prompt,
 				DisplayName: name,
+				Model:       opts.model,
 			}
 			var res spawnResult
 			if err := ctx.postJSON(cmd.Context(), "sessions", req, &res); err != nil {
@@ -176,6 +179,7 @@ func newSpawnCommand(ctx *commandContext) *cobra.Command {
 	f.StringVar(&opts.claimPR, "claim-pr", "", "Immediately claim an existing PR for the spawned session")
 	f.BoolVar(&opts.noTakeover, "no-takeover", false, "Refuse if another active session owns the claimed PR (requires --claim-pr)")
 	f.BoolVar(&opts.skipAgentCheck, "skip-agent-check", false, "Skip advisory agent catalog install/auth preflight before spawning")
+	f.StringVar(&opts.model, "model", "", `Agent model override for this session (e.g. "claude-sonnet-4-20250514"). Takes precedence over project and role agentConfig. Unsupported by some harnesses.`)
 	return cmd
 }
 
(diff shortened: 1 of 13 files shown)


Hardik180704 (changes requested):
Thanks for working on this — the overall direction looks good, but I found a few blockers before this is ready:

- The resolved model is persisted, but the restore path still rebuilds agent config only from the current project/role config. If that config changes later, a resumed session can silently switch away from its original per-spawn model. Please apply the persisted session model during restore and cover it with a regression test.
- The issue requires rejecting `--model` when the selected harness cannot honor it. Currently every non-empty value is forwarded, so the documented rejection behavior is not implemented.
- This PR adds no tests for the new behavior. Please cover per-spawn precedence, session persistence/readback, blank model input, unsupported harnesses, and restore behavior.

Also, GitHub currently reports no CI checks for this branch. Once these cases are addressed and the relevant checks are green, I’ll be happy to review again.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@praneshnikhar 
#4102 merged the --model flag for ao spawn, same feature and same files, and issue #3331 was closed completed by it. Closing as superseded.

PR #3357, opened 2026-07-30
feat(reviews): rework the Reviews tab and fix reviewer session handling

codebanditssss:
## Summary

Reworks the session inspector's Reviews tab around what it can actually tell you, and fixes three reviewer-runtime bugs found while testing it against a real repo.

**Reviews tab**

- hide the tab entirely when the session has no pull request, instead of showing an empty shell
- drop the redundant panel heading and the repeated agent role label
- show each reviewer's logo next to its name, sourced from a small harness→logo registry
- keep every past review pass, grouped by the reviewer that ran it, with a tab per reviewer above the PR list
- pick the reviewer agent for a single pass from the panel, reusing the same picker as project settings rather than a second one that could drift
- collapse PR rows by default, including the first one, so a long review doesn't push the rest of the panel off screen
- lock the reviewer picker while a pass is running and say why — multi-harness isn't supported yet
- let each session opt out of handing findings back to its worker agent

**Stale verdicts**

A "changes requested" verdict from before the latest push was rendered as if it still applied. It now reads as a previous verdict, with the commit it was left against.

**Reviewer runtime fixes**

- a terminal that outlived its agent is no longer treated as a live reviewer
- each review pass gets its own agent session id rather than reusing the pane's
- the tmux foreground probe uses `list-panes` instead of `display-message`

## Why

The reviewer's liveness check asked tmux whether the session existed. It always did — a tmux pane survives the process that started it and falls back to a shell. So AO decided the previous reviewer was still running and delivered the next review prompt to it, which meant **typing prompt text into an interactive shell inside the worker's worktree**. Harmless in practice only because the prompt opened with `Read` and the target was a non-executable `.md`.

Fixing liveness surfaced the next layer: claude-code derives a deterministic `--session-id` from the AO session id, and the reviewer handle is stable per worker, so a second pass on the same worker collided with the first — `Session ID … is already in use`. Confirmed by reproducing the UUIDv5 derivation from the handle and matching the id in the error. Passes now carry the run id, so each gets its own session. codex and opencode take no session flag and were never affected.

The probe added for liveness then turned out to be inert. `display-message` resolves an exact session target (`=name`) to the session but reports nothing for a pane-scoped format, so it returned empty on every call — and empty means "cannot tell", which callers treat as "keep the pane". The check kept answering yes. `list-panes` resolves the same target correctly and is what the codebase already uses for pane pids.

## Validation

- `go build ./...` and `go test ./internal/...` — no failures
- `npx tsc --noEmit` and `npx vitest run` — 1191 tests across 91 files pass
- `go generate ./internal/httpd/apispec/...` produces no diff, so the spec and the generated TS client are in sync
- tmux probe behaviour checked against the live server rather than assumed: `display-message` with `=name` returns empty, without it returns `zsh`, `list-panes` returns `zsh` either way
- reviews run end-to-end against a real GitHub repo in the desktop app

Not yet re-run end-to-end after the last commit: the `list-panes` fix is covered by a unit test and verified against live tmux, but a full claude-code pass on a restarted daemon hasn't been done since.

## Follow-ups

- a codex pass was observed posting its review successfully and then reporting `Conversation interrupted` — separate cause, still open


backend/internal/domain/projectconfig.go
```diff
@@ -51,6 +51,13 @@ type ProjectConfig struct {
 	// list falls back to claude-code (see ResolveReviewerHarness).
 	Reviewers []ReviewerConfig `json:"reviewers,omitempty"`
 
+	// ReviewAutoInjectOff stops AO handing a completed review's findings to the
+	// worker agent. Delivery has always been unconditional; this is the opt-out,
+	// so an absent value keeps that behaviour and only an explicit true disables
+	// it. Reviews are still recorded and shown either way — the switch is about
+	// whether the worker gets told.
+	ReviewAutoInjectOff bool `json:"reviewAutoInjectOff,omitempty"`
```
neversettle17-101: auto inject is is a session level config not a project level config. If we are showing it a session level then we are making it a session level config.

backend/internal/service/session/resolve_comments.go
```diff
@@ -0,0 +1,86 @@
...
+// ResolvePRComments resolves review threads on one of a session's pull requests.
+//
+// The PR is addressed by session + number rather than number alone: numbers are
+// only unique within a repository, and AO indexes pull requests by the session
+// that owns them, so a bare number cannot identify one.
+//
+// An empty threadIDs resolves every unresolved human thread on the PR, matching
+// the CLI's "resolve everything" shape. Bot comments are left alone — they are
+// not what a reviewer is waiting on.
+func (s *Service) ResolvePRComments(ctx context.Context, id domain.SessionID, prNumber int, threadIDs []string) (int, error) {
```
neversettle17-101: We don't require an API for this, currently the review agent does it on its own.

neversettle17-101 (changes requested):

backend/internal/domain/projectconfig.go
codebanditssss: Fixed in c49a3da8d. Auto-inject is now a durable session-level opt-out (including SQLite migration, CDC, API contract, and frontend session mapping); it has been removed from ProjectConfig.

backend/internal/service/session/resolve_comments.go
codebanditssss: Fixed in c49a3da8d. Removed the session-scoped resolve-comments service/API and the desktop Resolve button; review agents continue to own thread resolution.

frontend/src/renderer/components/ReviewerSelect.tsx
```diff
@@ -0,0 +1,106 @@
...
+						selected={selected}
+						status={agent.status}
+						statusTone={agent.statusTone}
+						disabled={agent.disabled}
+					/>
+				);
+			}}
+		/>
+	);
+}
```
neversettle17-101, line 109: `git diff --check origin/main...HEAD` reports a whitespace error in this file: `frontend/src/renderer/components/ReviewerSelect.tsx:106: new blank line at EOF`. Please remove the extra blank line so diff/check-style validation passes cleanly.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -771,11 +801,20 @@ function ReviewsView({
 		},
 	});
 	const reviewStates = reviewsQuery.data?.reviews ?? [];
+	const scmSummary = useSessionScmSummary(session.id);
+	const prSummaries = sessionPRDisplaySummaries(session, scmSummary.data);
+	const githubReviews = prSummaries.filter((pr) => pr.state === "open" && (pr.review?.reviews?.length ?? 0) > 0);
```
neversettle17-101: This filters the PR review panel to PRs that have at least one decisive `review.reviews` entry, but unresolved review-thread state is computed from all open PRs. A PR can have unresolved human review comments from a COMMENTED/non-decisive review and no `reviews` entry; in that case `unresolvedTotal` is positive but `GithubReviewPanel` receives no PR and renders “No one has reviewed this pull request yet.” Please include PRs with unresolved comments as well as PRs with decisive review rows.

backend/internal/adapters/scm/github/provider.go
```diff
@@ -767,3 +767,39 @@ func scrubError(err error) string {
...
+const resolveReviewThreadMutation = `mutation($threadId:ID!){
+  resolveReviewThread(input:{threadId:$threadId}){
+    thread{ id isResolved }
+  }
+}`
+
+// ResolveReviewThread resolves a single review thread on a pull request.
+// Resolving an already-resolved thread is a no-op on GitHub's side, so callers
+// may safely retry.
+func (p *Provider) ResolveReviewThread(ctx context.Context, threadID string) error {
```
neversettle17-101: This provider mutation appears to be unused in the final branch: there is no production caller for `ResolveReviewThread`, and the session-scoped resolve-comments route/UI described in the PR body is not present. The only remaining route is the legacy `/prs/{id}/resolve-comments`, which the PR notes still returns 501. Please either wire this through the intended session-scoped route/UI or remove the unused mutation and update the PR description so it does not promise resolving review threads.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -955,6 +1137,44 @@ function ReviewPanel({
...
+	// on each PR. Nesting the tabs inside a PR row meant expanding a row before
+	// you could compare agents at all.
+	const allRuns = [...runsByPR.values()].flat().sort((a, b) => b.createdAt.localeCompare(a.createdAt));
+	const reviewerTabs: string[] = [];
+	for (const run of allRuns) {
+		const name = run.harness || "reviewer";
+		if (!reviewerTabs.includes(name)) reviewerTabs.push(name);
+	}
+	const newestReviewer = reviewerTabs[0] ?? "";
+	const [activeReviewer, setActiveReviewer] = useState(newestReviewer);
```
neversettle17-101: This hook is reached only after the early returns above. On the common Reviews-tab path, the first render can hit `isLoading` and return before this `useState`, then the loaded render reaches this hook, which changes the hook count for `ReviewPanel` and can throw a rendered-more-hooks-than-previous-render error. Please move these reviewer state hooks above the early returns, or split the loaded body into a child component that only mounts after loading.

backend/internal/service/session/pr_summary.go
```diff
@@ -91,6 +91,9 @@ type PRReviewCommentLink struct {
 	URL  string
 	File string
 	Line int
+	// ThreadID is the review thread the comment belongs to, which is what
+	// GitHub's resolve mutation addresses. Several comments can share one.
+	ThreadID string
```
neversettle17-101: If this PR no longer intends to deterministically resolve GitHub review threads from the summary UI/API, please remove `ThreadID` from `PRReviewCommentLink` and the corresponding wire DTO/schema. The only reason this summary needs provider thread IDs is to drive `resolveReviewThread`; for display we already have URL/file/line, and exposing thread IDs without the resolve flow leaks unused provider mutation details into the session PR summary contract.

frontend/src/renderer/components/ReviewerSelect.tsx
codebanditssss: Fixed in d02681645. Removed the extra EOF blank line; both working-tree and branch diff checks now pass.

frontend/src/renderer/components/SessionInspector.tsx
codebanditssss: Fixed in d02681645. The panel now includes open PRs with either decisive review rows or unresolved comments, with a regression test for unresolved-only reviews.

backend/internal/adapters/scm/github/provider.go
codebanditssss: Fixed in d02681645. Removed the unused provider mutation and its tests, and updated the PR description to remove the resolve-thread promise and obsolete follow-up.

frontend/src/renderer/components/SessionInspector.tsx
codebanditssss: Fixed in d02681645. Reviewer state hooks now run before every early return, preserving a stable hook count across loading transitions.

backend/internal/service/session/pr_summary.go
codebanditssss: Fixed in d02681645. Removed thread IDs from the service summary, wire DTO, generated OpenAPI/TypeScript schema, fixtures, and tests.

neversettle17-101 (changes requested):
The end goal for this PR should be : 
1. User should be able to view the review comment summary from AO agent harness in one place
2. User should be able to select agent harness for a particular session without it affecting the other review sessions in the project.
3. Both #1 and #2 should be different.
4. Nail the UI for these things in this PR itself.
All other changes will come in following PRs

backend/internal/cli/dto_drift_e2e_test.go
```diff
@@ -104,6 +104,10 @@ func (f *fakeSessionService) SetTerminateOnPRMerge(context.Context, domain.Sessi
 	return domain.Session{}, nil
 }
 
+func (f *fakeSessionService) SetReviewAutoInject(context.Context, domain.SessionID, bool) (domain.Session, error) {
```
neversettle17-101, line 107: test 1

backend/internal/httpd/apispec/specgen/build.go
```diff
@@ -94,9 +94,14 @@ func Build() ([]byte, error) {
 		}
 		if op.reqBody != nil {
 			// AddReqStructure leaves requestBody.required absent, which
-			// OpenAPI reads as optional. These bodies are mandatory, so force
-			// it — otherwise validators/generators treat the body as skippable.
-			oc.AddReqStructure(op.reqBody, openapi.WithCustomize(markRequestBodyRequired))
+			// OpenAPI reads as optional. Most of these bodies are mandatory, so
```
neversettle17-101, line 97: test 2

neversettle17-101 (changes requested):

PR #3359, opened 2026-07-30
Open first browser content automatically and badge background activity

aprv10:
## What

Make the session Browser’s attention behavior reflect actual browser content and activity:

- Automatically open Browser when the session receives its first real content.
- Respect the user’s tab choice after that first reveal.
- Show the Browser glow when new preview or agent-browser activity occurs while Browser is hidden.
- Clear the glow when Browser becomes visible.
- Reset the lifecycle after the preview is cleared, allowing the next first content to auto-open again.
- Clear stale navigation state when switching sessions.

Users can now continue working in Summary, Reviews, or Files without Browser repeatedly stealing focus, while still receiving a meaningful indicator when background browser work occurs.

## Why

The previous glow was driven by terminal-output URL detection rather than browser state.

Any HTTP(S) URL printed in the terminal—including replayed terminal history—could activate the Browser glow even when the Browser had no page to display. At the same time, a legitimate first preview did not automatically reveal Browser.

This made the indicator misleading: it could appear with nothing to show and did not reliably guide users to actual preview content.

## How

Browser attention is now driven by authoritative session state:

- A non-empty `previewUrl` or browser navigation URL begins a content lifecycle.
- The first content in that lifecycle selects Browser and expands the inspector.
- Later `ao preview` revisions glow when the user is viewing another inspector tab.
- Agent-browser commands also glow while Browser is hidden, including commands that do not navigate.
- Opening Browser consumes the unseen state.
- Clearing all browser content resets both the reveal and unseen state.
- Explicitly clicking a terminal link still opens Browser immediately and persists the selected target.

### Removed

- Removed `detect-urls.ts` and its tests.
- Removed terminal output decoding through `useTerminalSession.onOutput`.
- Removed the `TerminalPane` watcher that treated printed URLs as Browser activity.

These paths existed only to power the passive URL glow. Clickable terminal links use the separate xterm link handler and remain supported.

## Testing

- `npm test -- src/renderer/components/SessionView.test.tsx src/renderer/components/SessionInspector.test.tsx src/renderer/components/TerminalPane.test.tsx src/renderer/hooks/useBrowserView.test.tsx src/renderer/hooks/useTerminalSession.test.tsx`
  - 5 test files passed
  - 158 tests passed
- `npm run typecheck`
  - Passed
- Manually verified the Browser auto-open, background glow, glow clearing, and tab-preservation behavior in the desktop app.

## Checklist

- [x] Branched from `main`
- [x] One focused change
- [x] Tests added or updated for the user-visible behavior
- [x] Relevant local frontend checks pass


frontend/src/renderer/hooks/useBrowserView.ts
```diff
@@ -253,6 +253,8 @@ export function useBrowserView({
 		// ensuring a different worker so equal revision numbers cannot suppress
 		// that worker's own target.
 		previewTriggerRef.current = null;
+		setViewId("");
+		setNavState(EMPTY_NAV_STATE);
```
illegalcall, line 291: This reset occurs in a passive effect, so consumers still render once with the previous session’s `navState`. `SessionView` then interprets that stale URL under the new `sessionId` and reveals Browser for an otherwise empty session. Please key the exposed navigation state to its owning session synchronously, or reject navigation state whose owner does not match the current session.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -240,23 +264,38 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
 	]);
 
-	// Keep the badge honest: clear it whenever the Browser tab is the open, active
-	// view (covers opening the rail while already parked on Browser, which
-	// setInspectorView's own clear does not see).
+	// Agent browser commands are genuine browser activity even when they do not
+	// navigate (fill, click, snapshot, etc.). When Browser is hidden, surface
+	// that activity as unseen rather than reopening the tab.
+	useEffect(() => {
+		if (!hasInspector || !browserView.agentBrowserActive || !hasBrowserContent) return;
```
illegalcall: This `!hasBrowserContent` guard prevents agent-browser activity from glowing on an initial blank target. That contradicts the PR requirement that agent commands—including commands that do not navigate—produce the unseen indicator while Browser is hidden. Activity attention needs to be recorded independently of the content lifecycle.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -184,6 +185,8 @@ export function SessionView({ sessionId }: SessionViewProps) {
 		sessionId: session?.id,
 		navUrl: browserView.navState.url,
 	});
+	const browserUrl = browserView.navState.url.trim();
+	const hasBrowserContent = Boolean(previewUrl || browserUrl);
```
illegalcall: `previewUrl` can remain persisted after termination, while `useBrowserView` deliberately suppresses/destroys terminated previews. Counting it here as real content makes a fresh visit to a terminated session open an empty Browser. Please exclude terminated/suppressed preview state from `hasBrowserContent`.

illegalcall (changes requested):
The overall direction is good, and the existing 158 focused tests plus typecheck pass. However, three behaviors promised by the PR are still broken. I reproduced each with a focused regression test:

1. Switching from a session with browser content to an empty session exposes the previous session’s URL during the destination render, causing the empty session to select Browser and expand its inspector.
2. Agent-browser commands on an empty target do not set the unseen indicator, despite the stated requirement that non-navigation commands glow while Browser is hidden.
3. A terminated session with a persisted `previewUrl` opens Browser even though `useBrowserView` suppresses and destroys that preview, leaving an empty Browser selected.

Please address these cases and retain the corresponding regression tests. The current green CI does not cover them.

Vaibhaav-Tiwari (dismissed):
1. Switching sessions can open Browser by mistake.
     If you move from a session that had browser content to another session with no browser content, the new session may briefly inherit the old browser URL and
     auto-switch to the Browser tab, leaving the user on a blank Browser tab instead of Summary.

  2. Ended sessions can open Browser even though preview is suppressed.
     If a terminated/merged session still has an old previewUrl, this PR may auto-open Browser for it, even though the browser hook intentionally hides old
     preview content for terminated sessions.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari:
Pushed a follow-up commit (0417c2b24) addressing the three regressions flagged in review:

1. **Stale navState leak across session switch** — `useBrowserView`'s `navState`/`viewId` are React state, not derived from `sessionId`, so they kept reporting the departed session's values for one render after `sessionId` changed (the reset only lands via an effect, a render later). `SessionView`'s reveal effect could read that stale render and wrongly auto-open Browser on a session with no content of its own. Fixed by adjusting state during render (React's documented pattern for resetting state on a prop change) instead of only via an effect, so no consumer ever observes a mismatched session's data — even transiently.
2. **Terminated session with a stale `previewUrl` auto-opening Browser** — a terminated session's persisted `previewUrl` is a stale DB fact once `useBrowserView` suppresses/destroys the live preview for it, so it no longer counts toward `hasBrowserContent`.
3. **Agent-browser commands not badging on an empty target** — commands like fill/click/snapshot are real activity even before any page has loaded; removed the `hasBrowserContent`/`browserContentRevealed` gates that were suppressing the badge in that case.

Added regression tests for all three, including one in `useBrowserView.test.tsx` that logs every committed render (not just the settled end value) to actually catch the transient stale-render bug — verified it fails without the fix and passes with it. Full frontend suite + typecheck are clean.

frontend/src/renderer/hooks/useBrowserView.ts
illegalcall: Addressed in 3979f7d3f: useBrowserView now keys exposed navigation/tab/activity state to its owning session, so the first destination render synchronously rejects all state from the departed session while the normal session effect resets the backing state.

frontend/src/renderer/components/SessionView.tsx
illegalcall: Addressed and regression-covered: agent-browser activity is recorded even on an empty target while Browser is hidden. 3979f7d3f also preserves that behavior while clearing the indicator on an explicit preview clear.

frontend/src/renderer/components/SessionView.tsx
illegalcall: Addressed and regression-covered: terminated sessions exclude persisted preview/navigation state from the content lifecycle and ignore agent-browser activity, so they cannot auto-open or badge an empty destroyed Browser.

aprv10:
@Vaibhaav-Tiwari @illegalcall 

I think all the review comments has been addressed. Does this PR require anything else?

illegalcall (approved):

PR #3369, opened 2026-07-30
fix: expand maximized session file diffs

Vaibhaav-Tiwari:
## What

Remove the inner 1200px width cap from the session Files view when it is maximized, so the diff uses the available session-panel width.

## Why

The maximized file diff viewer stayed scoped to the session panel, but its content was still centered inside max-w-[1200px], leaving large useless gutters on wide session panels.

## How

- Reuse the existing isMaximized prop in SessionFilesView.
- Keep the centered max-width wrapper for the normal inspector rail.
- Drop mx-auto max-w-[1200px] only in maximized mode.
- Leave the maximize overlay scoped to the session panel; it does not cover the full AO window.

## Testing

- npm.cmd run test -- SessionFilesView.test.tsx - 12/12 passed
- npm.cmd run typecheck - passed
- npm.cmd run test - attempted; unrelated Windows/env failures in scripts/verify-mac-artifact.test.mjs and src/main/supervisor-link.test.ts

## Checklist

- [x] Branched from main (or continuing an existing PR branch)
- [x] Relevant CI checks pass for the area touched (go, frontend, etc.)

### DIFF
--- a/frontend/src/renderer/components/SessionFilesView.test.tsx
+++ b/frontend/src/renderer/components/SessionFilesView.test.tsx
@@ -278,6 +278,17 @@ describe("SessionFilesView", () => {
 		expect(screen.getByLabelText("Session files").querySelector("header")).toHaveClass("h-11", "px-1.5");
 	});
 
+	it("uses the full session panel width while maximized", async () => {
+		const { unmount } = renderWithQuery(<SessionFilesView onClose={vi.fn()} sessionId="sess-1" />);
+		const railList = await screen.findByRole("list");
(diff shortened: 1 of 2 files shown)


Hardik180704 (approved):
Reviewed the latest change and found no blockers. The maximized Files view now drops only the inner mx-auto/max-w-[1200px] cap, so the diff uses the full session-panel width while the normal inspector rail keeps its existing centered width. The overlay scope, scrolling, and surrounding layout remain unchanged.\n\nVerified locally:\n- frontend typecheck\n- SessionFilesView tests: 12/12 passed\n\nCI is green.

PR #3371, opened 2026-07-30
fix: retain terminal viewport across navigation

illegalcall:
## Summary

- retain each lazily visited terminal renderer for its authoritative logical session plus terminal handle/generation lifetime
- reject stale routed shell/reviewer targets synchronously and replace reviewer renderers when a new run reuses a handle
- park inactive terminals hidden, inert, blurred, and unable to send input or resize while continuing to parse live output
- restore canonical bottom-follow or a marker-backed historical viewport before revealing the retained terminal
- share one browser-to-daemon mux WebSocket across independently owned terminal attachment leases
- preserve exited/error scrollback and Restore controls, badge URLs from hidden sessions, and clean up on authoritative lifecycle boundaries
- add deterministic real-xterm Chromium coverage for six live sessions, frame-sensitive reveal, reconnect, cleanup, scrollback eviction, and handle replacement

## Root cause

After #3185, navigating away still unmounted TerminalPane, xterm, and its attachment. Returning created a fresh attachment, so tmux replayed terminal state into a new buffer. Replay buffering reduced the visible walk but could not preserve the user-scrolled historical viewport, selection, or renderer state. A retained terminal could also expose a stale offscreen DOM scrollbar for one frame unless its canonical xterm buffer position and DOM viewport were synchronized before reveal.

The fix moves renderer and attachment ownership above routed session views. It does not add another timeout or extend the replay cover.

Follow-up to #3185 and #3160.

## User impact

Returning to a previously visited terminal restores the same xterm instance, selection, content anchor, and viewport. Hidden output advances the buffer without forcing a scrolled-up viewport to the bottom; terminals left at the bottom continue following output. Rapid A-B-A switching does not create duplicate writers or expose an intermediate replay frame.

Exact historical-anchor restoration is bounded by xterm's existing 5,000-line scrollback. This patch does not reduce terminal history. If enough hidden output evicts the selected line, return clamps once to the oldest retained line instead of replay-scrolling or jumping repeatedly.

The behavior is provider-independent: Codex, Claude Code, OpenCode/Kilo, reviewer terminals, and standalone shells all use the same terminal ownership layer.

## Validation

- focused Vitest: 5 files, 185 tests passed
- full renderer Vitest: 95 files, 1,291 tests passed
- frontend typecheck and E2E typecheck: passed
- isolated Chromium terminal/smoke run: 6 tests passed
- real-xterm cases cover six retained sessions, hidden/return-time output, bottom and historical anchors, reconnect, resize/fit, rapid A-B-A, empty/short replay, selection, scrollback eviction, generation replacement, one socket/multiple writers, and cleanup
- production macOS arm64 package: passed
- git diff check: passed

## Residual risk

A genuine terminal grid change still relies on the existing resize/SIGWINCH protocol, which has no explicit acknowledgement that a full-screen terminal application completed its repaint. The retained xterm is fitted and synchronized before reveal, but a runtime-specific repaint can arrive afterward. This was not observed in the supplied recordings, Chromium integration tests, or manual validation.


frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -955,7 +959,11 @@ function ReviewPanel({
 	const runAction = reviewSessionRunAction(openReviewStates, isTriggering);
 	const openReviewerTerminal = () => {
 		if (!terminalEnabled) return;
-		onOpenTerminal?.({ handleId: reviewerHandleId, harness });
+		onOpenTerminal?.({
+			generation: latest?.batchId || latest?.id || reviewerHandleId,
```
illegalcall: [P1] Key the reviewer cache from the actual current batch. `openReviewStates` is sorted by PR number, and `latest` above is only the first PR with any `latestRun`. If a lower-numbered PR has completed batch A while a higher-numbered PR was just reviewed in batch B, this sends generation A for the still-current reviewer handle. An existing exited A entry is then reused and the user never sees B, defeating the generation boundary this PR adds. The trigger success path has the same problem when an older run is already running. Select the newly created or newest run across all PR states, for example by `createdAt`, rather than the first array entry.

frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -636,6 +660,25 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
...
+		if (r.resizeTimer) {
+			clearTimeout(r.resizeTimer);
+			r.resizeTimer = null;
+		}
+		if (r.reassertTimer) {
+			clearTimeout(r.reassertTimer);
+			r.reassertTimer = null;
+		}
+		r.replayPendingReassert = null;
+		r.mux.resize(r.handle, cols, rows);
```
illegalcall, line 806: [P1] Do not emit a PTY resize while the entry is still preparing without a repaint barrier. `syncSize` is called from `prepareForActivation`, so this forwards `mux.resize` before the cache marks the terminal visible. The checked-in design explicitly says parked/preparing terminals cannot forward resize and the plan calls this phase transport-inert. More importantly, preparation waits only for local xterm render/RAF work; it never observes the SIGWINCH-driven remote repaint. On a grid change the pane can therefore be revealed with the old TUI frame and repaint afterward, violating the first-correct-frame guarantee. Keep remote resize out of preparation or add an authoritative repaint/ack barrier before reveal.

illegalcall (commented):
## Verdict: Changes requested

The focused renderer tests pass (92 tests) and frontend typecheck passes, but the core lifecycle contract still has two correctness gaps. Reviewer generation can be keyed to an older PR batch, and activation sends a resize while preparing then reveals before any remote repaint is known to have landed. Please address the inline findings before merge.

## Standards

No hard documented-standard violations found.

## Spec

2 findings. The worst issue is that multi-PR reviewer selection can reuse an exited renderer from an older batch.

frontend/src/renderer/components/SessionInspector.tsx
illegalcall: Addressed in 2c5d07851. Reviewer ownership now comes from a daemon-persisted reviewerGeneration stored beside the live handle and advanced only after successful Spawn/Notify; failed preflight, spawn, or notify preserves the previous handle-generation pair. SessionInspector consumes that authoritative field (with newest-createdAt projection only as compatibility fallback), and tests cover multi-PR ordering, omitted projections, generation replacement, and all three launch-failure paths.

frontend/src/renderer/hooks/useTerminalSession.ts
illegalcall: Addressed in 2c5d07851. Activation preparation no longer fits or forwards a PTY resize. It restores/renders the retained viewport while hidden, exposes a synchronized but inert/aria-hidden phase across paint frames, then becomes interactive; any real grid fit happens only in the visible phase. Ordinary retained A-B-A emits no resize. A parked reconnect opens at 0x0 and is promoted once, with a positive grid, only after visible activation. The real-xterm Chromium suite asserts the resize phases and frame-sensitive reveal.

backend/internal/storage/sqlite/migrations/0039_add_review_reviewer_generation.sql
```diff
@@ -0,0 +1,21 @@
...
+
+-- Existing databases did not record handle ownership separately. Preserve the
+-- best available generation so an upgrade does not invalidate every retained
+-- reviewer; all launches after this migration update it atomically.
+UPDATE review
+SET reviewer_generation = COALESCE((
+    SELECT CASE WHEN review_run.batch_id != '' THEN review_run.batch_id ELSE review_run.id END
+    FROM review_run
+    WHERE review_run.session_id = review.session_id
+    ORDER BY review_run.created_at DESC, review_run.id DESC
```
illegalcall: **[P2] Do not infer handle ownership from the newest run**

Review runs are inserted before preflight, Spawn, or Notify, and those failure paths mark the new runs failed without advancing `review.reviewer_handle_id`. An existing database can therefore have a live handle from batch A and a newer failed, never-launched batch B. This unconditional newest-run lookup assigns B as the generation of A on upgrade, so the frontend treats B as the owner of the old pane—the exact ownership advance that the design prohibits after launch failure. Backfill from an identity known to represent the existing handle (for example, a stable legacy generation derived from that handle), rather than from an unverified latest run.

illegalcall (commented):
## Verdict

**Changes requested.** The follow-up fixes the two previously reported runtime issues: reviewer ownership now comes from the separately persisted full-history generation, and activation no longer resizes the PTY before the retained frame is revealed. One upgrade-path ownership bug remains.

## Standards

No standards deviations found. The change uses a new migration, updates the source SQL plus generated sqlc output, and keeps the OpenAPI and frontend schema artifacts synchronized.

## Spec

- **[P2] The migration can assign a failed, never-launched batch to the previous reviewer handle.** See the inline comment.

## Validation

- `go test ./internal/review ./internal/storage/sqlite/store ./internal/httpd/controllers ./internal/cli`
- Focused frontend Vitest suites: 144 tests passed

backend/internal/storage/sqlite/migrations/0039_add_review_reviewer_generation.sql
illegalcall: Addressed in f12ffd4f5. Migration 0039 now derives the opaque legacy generation directly from the existing reviewer_handle_id and never infers ownership from review_run history. A real v38→v39 migration regression seeds an older live reviewer handle plus a newer failed, never-launched batch and verifies that the handle-derived token survives. The next successful Spawn/Notify still replaces the legacy token with the actual launched batch generation.

illegalcall (commented):
## Verdict

**Ready.** The follow-up resolves the remaining upgrade-path ownership bug. Migration 0039 now uses the existing reviewer handle as a stable opaque legacy generation instead of inferring ownership from a newer historical run, and the regression test covers a newer failed, never-launched batch. Successful future Spawn or Notify operations still replace that token with the actual batch generation.

## Standards

No documented-standard violations found. As a non-blocking judgement call, `TerminalPane.tsx` now concentrates cache ownership and lifecycle policy alongside terminal rendering and restore UI (possible Divergent Change); this does not warrant changes in this PR.

## Spec

No remaining spec findings. The two earlier runtime issues remain fixed, and the legacy migration behavior now matches the documented ownership requirement.

## Validation

- `go test ./internal/storage/sqlite ./internal/review ./internal/httpd/controllers ./internal/cli`
- `go test -count=1 ./internal/storage/sqlite -run TestMigration0039BackfillsReviewerGenerationFromExistingHandle`
- Focused frontend Vitest suites: 144 tests passed

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale:
Implemented terminal replay and resize stability fixes:

- Split large first-open terminal replay into 256 KB `xterm.write()` batches to reduce long renderer-blocking tasks.
- Preserve exact PTY output order when live data arrives during replay.
- Safely drain pending replay data before terminal exit or reconnect.
- Wait for fullscreen/layout transitions to settle before fitting xterm.
- Remove duplicate fits and unconditional full-grid refreshes.
- Ignore unchanged terminal dimensions in both Electron and the daemon.
- Prevent secondary viewers from redundantly resizing a shared PTY.
- Fall back to Canvas when the WebGL context is lost.

These changes reduce visible first-open repainting, unnecessary TUI redraws, and renderer stalls. Large or continuously streaming terminals may still show a CPU spike because xterm must parse and render the complete output; batching improves responsiveness but does not eliminate the total parsing cost.

illegalcall (commented):
## Latest-head review (`a320ac84d`)

The resize/layout storm is substantially reduced, but I do not think this is ready to call fully fixed yet.

1. **[P1] The PR is not currently mergeable/test-clean against `main`.** The current Frontend `test` check fails while compiling the GitHub merge candidate: `SessionView.tsx` constructs a shell `TerminalTarget` without the newly required `generation`. `main` has advanced again since that run, and a fresh merge-tree now also reports a content conflict in `_shell.tsx`. Please merge/rebase the latest base and run the frontend checks on the resolved result, not only on the branch head.

2. **[P1] The new orchestrator-worker regression test does not exercise the regression as committed.** `terminal-viewport-retention.spec.ts:401` asks for a button named `Open Project orchestrator`, but the merged UI exposes `Open orchestrator` (topbar) and `Open fake-proj orchestrator` (sidebar), so the test times out before its first switch. After correcting that selector locally, `--repeat-each=3` still failed once because the baseline was captured before the orchestrator's activation/settle resize completed (`Expected 0, Received 1`). The test needs to wait on an explicit fit/activation-settled condition before recording counts.

3. **[P2] A failed fit is still reported as “settled.”** In `XtermTerminal.tsx:575-595`, `flushScheduledFit` catches `fit.fit()` failure but then unconditionally invokes and clears all settled listeners. `prepareForActivation` uses that callback to reveal the terminal. During an unmeasurable or still-moving layout this can reveal a stale/default grid and then perform the later WebGL/PTY resize that this change is intended to avoid. Only resolve activation preparation after a successful measurable fit; keep/retry the listeners after failure (with a separate bounded fallback if needed).

4. **[P2] The regression assertion is too far downstream to establish that the CPU issue is fixed.** Three cycles that only inspect mux opens/resizes cannot catch a reintroduced flex/layout/WebGL storm if daemon deduplication masks the final resize. In an exact same-machine six-cycle A/B comparison:

   - parent `4986af7`: 2.43–2.45 s renderer task time, 264–270 layouts, 6 resize publications per terminal, 2.60 s wall time;
   - fix commit `23eb763`: 1.93–2.19 s renderer task time, 85–87 layouts, 0–1 resize publications, 3.72–3.96 s wall time.

   So the change clearly lowers peak CPU/layout/PTY churn, but every warmed activation now pays the 120 ms quiet window and total switching becomes about 45% slower. Please add a renderer-level performance assertion or instrumentation for layout/fit work and avoid imposing the full quiet delay when the slot is already stable.

Focused validation: backend terminal tests passed; 109 focused renderer unit tests passed; frontend typecheck passes on the head alone; the full terminal viewport E2E file is not green on this head (1/5 passed locally), including the deterministic selector failure above.

backend/internal/terminal/manager.go
```diff
@@ -158,18 +158,23 @@ func (m *Manager) joinTerminal(id string, c *connState, att *attachment, cols, r
...
+		_ = att.resize(s.authRows, s.authCols)
 		c.enqueue(serverMsg{Ch: chTerminal, ID: id, Type: msgResize, Cols: s.authCols, Rows: s.authRows})
 	}
 }
 
 // updateTerminalSize records a client's newly requested grid and reconciles.
-func (m *Manager) updateTerminalSize(id string, c *connState, cols, rows uint16) {
+// Identical normal updates are no-ops; force is reserved for an explicit
+// recovery re-signal when an attach client may have missed the prior SIGWINCH.
+func (m *Manager) updateTerminalSize(id string, c *connState, cols, rows uint16, force bool) {
```
illegalcall, line 177: [P2] Please remove or constrain the new `force` resize escape hatch. No production caller passes `true` (only tests do), yet any mux client—including a secondary viewer—can set it, bypass the identical-grid guard, and make `reconcileLocked` resize every attachment at an unchanged authoritative grid. That recreates the redundant SIGWINCH path this fix is removing and expands the wire contract for a hypothetical recovery flow. If recovery is needed later, model it as an authorized primary-only operation with a real caller; otherwise keep unchanged resizes unconditional no-ops.

nikhilachale (approved):
Tested the changes and it works well

PR #3374, opened 2026-07-30
fix(terminal): number session tabs, retire dead ones, tidy the strip

Pulkit7070:
Replaces https://github.com/Untrivial-ai/agent-orchestrator/pull/3298, which conflicted with `main` and is now largely redundant.

## Why a new branch

https://github.com/Untrivial-ai/agent-orchestrator/pull/3275 landed the same cross-session removal #3298 was carrying: the dropdown gone, `+` calling `onNewShellTerminal` directly, and the same props dropped (`projectSessions`, `availableProjectSessions`, `onAddProjectSession`, `onCloseProjectSession`, `onSelectProjectSession`, `tabOwnerSessionId`), plus `sessionTabsByOwner` and the `tabOwner` search param. Resolving #3298's five conflicting files would have meant merging duplicate code, so this is only the part `main` does not already have, rebuilt on top of it.

It also carries https://github.com/Untrivial-ai/agent-orchestrator/pull/3329, which was merged into #3298's branch rather than `main` and so never reached it, including all three review fixes from that PR.

## Changes

**Naming.** Every shell in a session starts in that session's worktree, so `filepath.Base(workingDir)` put the identical label on every tab — three terminals in one session all read `ao-22`. Session tabs are numbered instead; the session is already the first tab in the strip. Standalone shells (board, `/terminals`) keep the directory name, which is what tells shells from different projects apart there.

The number is **one past the highest still in use**, not a count, so closing `Terminal 1` while `Terminal 2` is open does not hand the next tab a name already on screen. A renamed tab no longer parses and stops reserving its number.

**Exit handling.** A shell whose PTY ended left a tab that could never be attached again. The daemon already prunes dead shells on list, but nothing told the client — the list is only refetched around this client's own open/close.

The pane now reports the exit as a **hint, never a close**. `attachment.fail()` reaches the same `markExited()` after the liveness probe or the attach errors past the retry cap, and the attach loop's own comment says a probe error is not proof of death — destroying on that would kill a live shell and whatever is running in it. So the renderer asks the daemon, whose list prunes only what it can confirm is gone and keeps anything whose probe errored. If the shell is still listed, the attachment is remounted (an attach epoch in `TerminalPane`'s key) so the pane reconnects rather than sitting on "Terminal ended" over a live shell.

**Ordering.** The ordinal is resolved before `runtime.Create`, so a store failure cannot leave a PTY with no row pointing at it — nothing could then find it to reap.

**Strip chrome.** The `TERMINAL` heading is gone; the tabs already say what it is. That exposed a second gap: the scroll chevrons were hidden with `invisible`, which is `visibility: hidden` and still reserves the full box, leaving a dead column before the first tab. They collapse to zero size now and stay mounted so focus is not dropped mid-scroll.

**Preview mock.** It stamped the project name on every tab, which the daemon never produces, making the browser dev build misleading.

## Verification

Driven against a real daemon (isolated `HOME`, scratch data dir, own port) with a real session via `--harness fake`, on this branch's code:

```
### session terminals are numbered
  open -> Terminal 1
  open -> Terminal 2
  open -> Terminal 3

### close Terminal 1, open again -> must not reissue a live name
  after close: ['Terminal 2', 'Terminal 3']
  next open  -> Terminal 4

### standalone open in the same daemon still uses the directory
  -> data
```

Also driven earlier against a real daemon: a shell whose tmux session is killed out from under it disappears from `GET /shell-terminals`, the daemon half of the exit path.

`npm test` 1286 passing, `npm run typecheck` clean, Go build and the shellterm/tmux package tests pass, `gofmt` clean.

The regression tests were each checked to fail without their fix: the false-exit test fails without the epoch bump, and the leaked-runtime test fails with the old `Create`-then-lookup ordering.

## Not verified by running

The renderer's retire-on-exit needs a live PTY, which only exists in the Electron app. The daemon prune and the renderer's reporting are each covered, but not the two joined end to end.

## Known limitation

Only the attached pane notices an exit. A background tab whose shell dies lingers until you select it or open/close something. The real fix is a daemon push when a PTY dies rather than polling the list.

### DIFF
--- a/backend/internal/service/shellterm/service.go
+++ b/backend/internal/service/shellterm/service.go
@@ -233,6 +233,23 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		return ShellTerminal{}, fmt.Errorf("open shell terminal: handle id: %w", err)
 	}
 
+	// Resolve the tab name BEFORE the runtime exists. This reads the store, and
+	// a store failure after Create would strand a PTY that nothing can find:
+	// no row means neither shutdown nor the app-run reaper can ever destroy it.
+	// Doing it here means the only failure that can leave a runtime behind is
+	// the insert below, which already rolls back.
+	//
+	// Counting is safe: the session gate is held for the whole open, so two
+	// concurrent opens for one session cannot pick the same number.
+	title := shellTerminalTitle(workingDir)
+	if in.SessionID != "" {
+		ordinal, err := s.nextSessionTerminalOrdinal(ctx, in.SessionID)
+		if err != nil {
+			return ShellTerminal{}, err
+		}
+		title = sessionShellTerminalTitle(ordinal)
+	}
+
 	// SessionID is the runtime adapters' name for "what to call this PTY"; it
 	// is not a session row and no sessions record is ever created. The
 	// shellterm- prefix keeps the two namespaces disjoint.
@@ -253,7 +270,7 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 		ProjectID:  projectID,
 		SessionID:  in.SessionID,
 		WorkingDir: workingDir,
-		Title:      shellTerminalTitle(workingDir),
+		Title:      title,
 		AppRunID:   s.appRunID,
 		CreatedAt:  s.now().UTC(),
 	}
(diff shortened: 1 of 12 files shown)


Rishet11:
@Pulkit7070 
Closing as superseded. Terminal tab numbering and lifecycle landed through merged #4572 and #4726, which together cover this PR's shellterm and CenterPane changes. The referenced issue #3298 was also closed unmerged. Nothing of this branch remains to merge.

Rishet11:
@Pulkit7070 Hi, thanks for this. Terminal tab numbering did land on main, but through a different implementation than this branch's, and the shell-exit retirement and tab-logo tidy-up here never landed anywhere and aren't tracked by an issue. Given the branch is stale and conflicts with main, I'm closing it, but a fresh smaller PR for the retirement behavior would be welcome.

PR #3381, opened 2026-07-30
feat(review): add Kilo Code reviewer

nikhilachale:
## Summary

- add `kilocode` as a first-class AO reviewer harness, registered independently from worker harnesses
- derive the adapter from the existing OpenCode reviewer design: reuse the Kilo worker command builder, keep AO reviewer panes persistent for follow-up tasks, store standing/task prompts outside the worktree, and apply a read-only inspection/reporting policy
- allow only repository inspection plus the required `gh api` and `ao review submit` reporting commands; preserve AO cancellation and runtime failure behavior
- expose Kilo Code in project configuration and the desktop reviewer selector, with domain, adapter, UI, and documentation coverage

## Official Kilo Code CLI findings

Sources:
- https://kilo.ai/docs/code-with-ai/platforms/cli
- https://kilo.ai/docs/code-with-ai/platforms/cli-reference
- https://kilo.ai/docs/contributing/architecture/cli-runtime
- https://github.com/Kilo-Org/kilocode/blob/main/packages/opencode/package.json

The current package is `@kilocode/cli`. `kilo` is the primary executable and the package also installs `kilocode` as an official alias. Headless automation is `kilo run --auto <message>`; `kilo run` supports `--agent`, `--dir`, and `--format default|json`. Documented headless exits are 0 for success, 1 for initialization/execution failure, and 124 for timeout. Authentication is configured through `/connect`, `kilo auth login`, provider config/environment variables, and inspected with `kilo auth list`.

AO intentionally uses the existing persistent Kilo worker/TUI command path instead of `kilo run --auto`, because reviewer panes receive later review tasks. AO does not parse Kilo JSON events; the reviewer posts through `gh api` and records results through `ao review submit`.

## Tests

- `go test ./internal/adapters/reviewer/kilocode ./internal/adapters/reviewer ./internal/domain ./internal/review` — passed
- `env -u AO_SESSION_ID AO_RUN_FILE=/tmp/ao-agent-orchestrator-195-no-daemon.json AO_DATA_DIR=/tmp/ao-agent-orchestrator-195-data TMPDIR=/tmp go test -p 1 ./...` — passed
- `go test ./internal/adapters/agent/kilocode -run TestAuthStatusUnknownWhenKeyOnlyComesFromInteractiveShell -count=1` — passed
- `npm run typecheck` in `frontend/` — passed
- `npx vitest run --config vite.renderer.config.ts src/renderer/components/ProjectSettingsForm.test.tsx` — passed
- `git diff --check` — passed
- `npm run build` in `frontend/` — not available on current `main`; npm reports `Missing script: build`

## Known limitations

- reviewer auth remains advisory through the existing agent catalog; launch/preflight validates binary availability, while provider/auth failures use the existing runtime failure path
- Kilo headless JSON output is not an AO result contract
- this change does not implement Cursor
- the frontend package currently has no `build` script

### DIFF
--- a/backend/internal/adapters/reviewer/kilocode/kilocode.go
+++ b/backend/internal/adapters/reviewer/kilocode/kilocode.go
@@ -102,6 +102,7 @@ func withReviewerConfig(argv []string, taskPromptRoot, systemPromptFile string) 
 			"git show*":                     "allow",
 			"git status*":                   "allow",
 			"ao review submit *":            "allow",
+			"printf *":                      "allow",
 			"printf * | gh api *":           "allow",
 			"printf * | ao review submit *": "allow",
 		},
--- a/backend/internal/adapters/reviewer/kilocode/kilocode.go
+++ b/backend/internal/adapters/reviewer/kilocode/kilocode.go
@@ -1,0 +1,154 @@
+// Package kilocode adapts the Kilo Code worker agent for code-review sessions.
+package kilocode
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	workeragent "github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/kilocode"
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+const configAssignmentPrefix = "KILO_CONFIG_CONTENT="
+
+// Reviewer is the Kilo Code code-review adapter.
+type Reviewer struct {
+	agent ports.Agent
+}
+
+// New builds the Kilo Code reviewer adapter.
+func New() *Reviewer {
+	return &Reviewer{agent: workeragent.New()}
+}
+
+// Harness identifies this reviewer in the reviewer registry.
+func (r *Reviewer) Harness() domain.ReviewerHarness {
+	return domain.ReviewerKiloCode
+}
+
+var _ ports.Reviewer = (*Reviewer)(nil)
+var _ ports.ReviewerCanceller = (*Reviewer)(nil)
+
+// ReviewCommand launches Kilo Code with the same hidden system-prompt agent as
+// the worker adapter, then replaces its normal permission config with the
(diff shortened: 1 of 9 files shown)


nikhilachale (commented):
## Changes requested

The adapter is registered cleanly and the focused Go tests pass, but the Kilo shell policy blocks the reporting pipelines mandated by the AO task, so the reviewer cannot complete its core contract.

backend/internal/adapters/reviewer/kilocode/kilocode.go
```diff
@@ -0,0 +1,154 @@
...
+		"grep": "allow",
+		"bash": map[string]string{
+			"*":                             "deny",
+			"gh api *":                      "allow",
+			"git diff*":                     "allow",
+			"git log*":                      "allow",
+			"git show*":                     "allow",
+			"git status*":                   "allow",
+			"ao review submit *":            "allow",
+			"printf * | gh api *":           "allow",
```
nikhilachale, line 106: **[P1] Allow the parsed `printf` stage used by reporting**

Kilo evaluates shell pipelines as separate parsed commands and requires every command to be permitted. Both task-mandated reporting commands begin with `printf ... | ...`; although `gh api *` and `ao review submit *` are allowed, the parsed `printf ...` command still matches the catch-all deny. A whole-pipeline glob does not authorize that component, so Kilo will deny the command and cannot post the GitHub review or submit AO bookkeeping. Please allow the narrowly scoped `printf` stage (or use a reporting mechanism without a pipe) and add a regression test that reflects Kilo parsed-command permission behavior.

nikhilachale: Addressed in 0e9e0feaa: the Kilo reviewer bash policy now explicitly allows the parsed `printf *` stage used by both required reporting pipelines, while retaining the catch-all deny. I also added a regression test that evaluates each pipeline stage independently (the `printf`, `gh api`, and `ao review submit` commands) and verifies an unrelated destructive command remains denied.

nikhilachale:
Superseded by #3384, which combines the Cursor, Kilo Code, and Copilot reviewer integrations into one reviewed union without dropping this PRs behavior or tests.

PR #3382, opened 2026-07-30
feat: add Copilot reviewer

nikhilachale:
## Summary
- add GitHub Copilot CLI as a persistent AO reviewer with a narrow read-only command policy
- split Copilot custom-agent profile installation from worker lifecycle hooks
- register Copilot in reviewer resolution and project settings

## Tests
- go test ./internal/adapters/agent/copilot ./internal/adapters/reviewer/copilot ./internal/adapters/reviewer ./internal/domain ./internal/review
- ProjectSettingsForm Vitest: 17 passed
- npm run frontend:typecheck
- golangci-lint v2.12.2
- git diff --check
- Copilot CLI 1.0.76 parser check with the exact allow/deny policy

## Known limitations
- Full npm run lint was attempted repeatedly, but unrelated shared-host integration tests collided with concurrent AO sessions: the fake lifecycle timing test observed scheduler-delayed sleeps, and the tmux integration test reused a globally named session. Focused packages and golangci-lint pass.
- Real security smoke omitted because no designated safe test PR was available.

PR #3380, opened 2026-07-30
feat(review): add Cursor reviewer

nikhilachale:
## Summary

- add Cursor as a first-class AO reviewer harness, following the existing OpenCode-derived adapter/registry contract
- reuse the Cursor worker adapter for `cursor-agent` discovery and launch construction
- run reviews headlessly with `--print --output-format text --mode=ask`; AO continues to provide prompt files and reviewers self-submit through `ao review submit`
- register Cursor in project config validation, reviewer discovery, the desktop reviewer selector, and README documentation

## Design and official Cursor constraints

The adapter follows the OpenCode reviewer shape: a dedicated reviewer wrapper supplies command, message, and cancellation behavior; the central launcher owns workspace/runtime execution and executable preflight; AO-owned system/task files contain the review instructions; and results are recorded through the existing `ao review submit` boundary rather than parsed from terminal output.

Official references reviewed:

- CLI overview/executable: https://docs.cursor.com/en/cli/overview
- headless print mode: https://docs.cursor.com/en/cli/headless
- parameters: https://docs.cursor.com/en/cli/reference/parameters
- authentication (`cursor-agent login/status`, `CURSOR_API_KEY`): https://docs.cursor.com/en/cli/reference/authentication
- output and nonzero failure semantics: https://docs.cursor.com/en/cli/reference/output-format
- config-file permissions: https://docs.cursor.com/cli/reference/permissions
- Ask mode: https://cursor.com/changelog/cli-jan-16-2026
- official repositories: https://github.com/cursor

Cursor exposes permissions through configuration files rather than a per-invocation tool allowlist. This integration uses Ask mode for non-editing review analysis and `--force` so the required git inspection and `gh`/`ao review submit` reporting commands do not wait for interactive approval.

## Integration surfaces

- new `backend/internal/adapters/reviewer/cursor` adapter and focused tests
- reviewer domain vocabulary and registry
- project config validation and worker-harness fallback resolution
- desktop project-settings reviewer selection and test coverage
- README reviewer list

No HTTP DTO changed, so OpenAPI/frontend schema regeneration was not required.

## Verification

- `go test ./internal/adapters/reviewer/... ./internal/domain ./internal/review` — passed
- `go test ./internal/adapters/agent/kilocode -run TestAuthStatusUnknownWhenKeyOnlyComesFromInteractiveShell -count=1` — passed on isolated rerun
- `git diff --check` — passed
[...]

PR #3384, opened 2026-07-31
feat(review): add interactive reviewers for all agent harnesses

nikhilachale:
## Summary

Adds interactive code-review support for every currently supported AO agent harness. Reviewers run in persistent terminal sessions, receive an AO-owned review prompt, inspect the target change, and submit structured findings through the existing review flow.

This PR expands the reviewer catalog from Claude Code, Codex, and OpenCode to **23 reviewer adapters**:

- Agy, Aider, Amp, Auggie, Autohand, Cline, Continue, Copilot, Crush, Cursor
- Devin, Droid, Goose, Grok, Kilo Code, Kimi, Kiro, Pi, Qwen, and Vibe
- Existing Claude Code, Codex, and OpenCode reviewers remain supported

It also exposes the complete catalog in project settings and adds documentation and test coverage for reviewer selection and launch behavior.

## How reviewers work

- Reviewers launch as interactive TUIs in persistent runtime panes.
- AO injects a reviewer-specific system/task prompt that requires read-only analysis and structured review submission.
- Later review tasks reuse a healthy pane; stopped or superseded panes are replaced safely.
- Cancellation uses the adapter's supported interrupt behavior, including native Escape handling where required.
- Reviewer configuration is separate from worker and orchestrator configuration; a project must select the reviewer it wants.

## Reviewer behavior matrix

| Reviewer | Session behavior | How AO constrains the reviewer |
| --- | --- | --- |
| Claude Code | Persistent interactive pane | AO reviewer system prompt through the native Claude adapter; automatic permission mode, so prompt policy is part of the trust boundary |
| Codex | Persistent interactive pane | AO reviewer system prompt through the native Codex adapter, with Codex's native approval/sandbox configuration |
| OpenCode | Persistent interactive pane | AO reviewer system prompt through the native OpenCode adapter; automatic permission mode, so prompt policy is part of the trust boundary |
| Cursor | Persistent interactive pane | Private AO-owned reviewer profile and restrictive allow/deny configuration; no force, YOLO, or headless flags |
| Kilo Code | Persistent interactive pane | Hidden reviewer-agent configuration, isolated project resources, and an explicit read/report-only policy |
| Copilot | Persistent interactive pane | AO-owned profile with granular tool allow/deny rules; broad allow-all flags are not enabled |
| Pi | Persistent interactive pane | Resource discovery and built-in tools disabled; an AO-owned extension exposes bounded reads, fixed Git inspection, and structured submission only |
| Kiro | Persistent interactive pane | Runs from an AO-owned neutral directory with a private reviewer agent and deny-by-default shell policy |
| Amp | Persistent interactive pane | AO-owned settings plus the reviewer system prompt; readiness markers are used before task delivery |
| Aider | **Fresh process per review pass** | AO supplies the system and task prompt files as read-only context; the process is request-scoped |
| Agy | Persistent interactive pane | **Experimental host-trusted** mode; restrictive prompt/CLI setup reduces accidental edits but is not OS isolation |
| Continue | Persistent interactive pane | **Experimental host-trusted** mode; launches in an AO-owned neutral directory and receives the task after readiness |
| Devin | Persistent interactive pane | **Experimental host-trusted** mode; AO injects the reviewer role, but the provider does not expose enforceable cross-platform containment |
| Droid | Persistent interactive pane | **Experimental host-trusted** mode; AO injects the reviewer role, but the provider does not expose enforceable cross-platform containment |
| Goose | Persistent interactive pane | **Experimental host-trusted** mode with a constrained reviewer configuration; not an OS sandbox |
| Kimi | Persistent interactive pane | **Experimental host-trusted** mode; restrictive launch shape and prompt policy, without OS-level isolation |
| Qwen | Persistent interactive pane | **Experimental host-trusted** mode from an AO-owned neutral directory; task is delivered after readiness and Escape cancels the active turn |
| Vibe | Persistent interactive pane | **Experimental host-trusted** plan mode; restrictive mode reduces accidental authority but is not OS isolation |
| Grok | Persistent interactive pane | **Experimental user-approved** mode; AO supplies the reviewer role and keeps native permission prompts enabled |
| Crush | Persistent interactive pane | **Experimental user-approved** mode; AO supplies the reviewer role and keeps native permission prompts enabled |
| Auggie | **Fresh process per review pass** | **Experimental user-approved** mode using native rules/permission prompts; the process lifecycle is request-scoped |
| Cline | Persistent interactive pane | **Experimental user-approved** mode; system instructions are injected without enabling auto-approval |
| Autohand | Persistent interactive pane | **Experimental user-approved** mode; AO supplies the system prompt and leaves native approvals enabled |

All 23 adapters use an interactive terminal UI. “Host-trusted” means the no-edit rule is enforced primarily by prompt and restrictive provider modes, not by an AO-controlled OS sandbox. These reviewers must therefore be selected explicitly.

## Safety model

- Hardened adapters use provider-specific controls such as isolated AO-owned configuration, restricted tools, permission flags, or the secure reviewer gateway.
- Adapters without enforceable CLI containment are clearly treated as **host-trusted/experimental** and require explicit selection. Their prompts prohibit file edits, but prompts are guidance rather than an OS sandbox.
- Reviewer launch fails closed when required binaries, versions, capabilities, or runtime prerequisites are unavailable.
- The launcher verifies readiness before reusing a pane and avoids delivering tasks to stale or superseded sessions.

## Notable infrastructure

- Adds the provider-neutral secure TUI gateway and capability manifests.
- Adds constrained checkout/prompt reads, fixed read-only Git operations, and structured GitHub/AO review submission.
- Extends tmux and ConPTY runtime support for interactive reviewer cancellation and lifecycle handling.
- Adds reviewer availability and trust-state messaging to project settings.

## Verification

- All 11 GitHub CI checks pass on the current head.
- Backend reviewer, review launcher, gateway, domain, controller, and runtime tests pass.
- Frontend reviewer settings tests and TypeScript checks pass.
- Adapter tests cover command shape, prompt delivery, cancellation, readiness, pane reuse, and fail-closed behavior.

These are deterministic unit/integration checks using fake runtimes and command assertions. They do not claim live end-to-end authentication or review execution against every third-party provider.

## Supersedes

- #3380
- #3381
- #3382


nikhilachale:
Addressed both Cursor findings in 4dd6b99a0cbb80f5839cb3acfe71ff7b74c80871:

- `ReviewCommand` now adds `--add-dir <TaskPromptRoot>` before the prompt, with command coverage.
- Cursor now declares its headless pane non-reusable. The launcher replaces the stable pane on a later notification, and the regression test verifies it destroys/recreates the pane without sending the task to the fallback shell.

Focused reviewer/review/domain tests pass, full `npm run lint` passes, and `git diff --check` is clean.

backend/internal/adapters/reviewer/cursor/cursor.go
```diff
@@ -0,0 +1,99 @@
...
+	argv, err := r.agent.GetLaunchCommand(ctx, ports.LaunchConfig{
+		SessionID:     inv.ReviewerID,
+		WorkspacePath: inv.WorkspacePath,
+		Prompt:        prompt,
+		Permissions:   ports.PermissionModeAuto,
+	})
+	if err != nil {
+		return ports.ReviewCommandSpec{}, err
+	}
+	flags := []string{"--print", "--output-format", "text", "--mode=ask"}
```
nikhilachale: **[P1] Trust the workspace for the headless Cursor run**

A project can select Cursor as its reviewer even when the worker uses another harness, so the worker checkout is not guaranteed to have Cursor trust state seeded. This command uses `--print`, which cannot answer the workspace trust prompt, but it does not pass the Cursor CLI `--trust` flag or perform the worker adapter trust prelaunch. On a fresh or previously unseen checkout the reviewer can therefore stall before reading the task. Add `--trust` to this headless invocation and cover it in the argv test.

frontend/src/landing/content/docs/configuration/projects.mdx
```diff
@@ -158,6 +158,22 @@ worker:
...
+## Choose A Reviewer
+
+Reviewer agents are configured independently from worker and orchestrator agents:
+
+```yaml
+reviewers:
+  - harness: kilocode
+```
+
+Built-in reviewer harnesses are `claude-code`, `codex`, `opencode`, and `kilocode`. Kilo Code requires the
```
nikhilachale: **[P2] List every reviewer harness introduced here**

This PR registers six built-in reviewers, but the newly added configuration documentation lists only four and omits `copilot` and `cursor`. Users following this page will incorrectly conclude that those two new options are unsupported. Please keep this list aligned with `domain.AllReviewerHarnesses`.

nikhilachale (commented):
Changes requested. The one-shot respawn and prompt-directory access fixes address the previous findings, but fresh Cursor reviewer launches can still block on workspace trust. The new reviewer documentation also omits two harnesses added by this PR.

Tests run: `cd backend && go test ./internal/adapters/reviewer/... ./internal/review ./internal/domain` (pass).

backend/internal/adapters/reviewer/cursor/cursor.go
nikhilachale: Addressed in dd42b911fca88844da59e18d91cce8fc50224225. The headless Cursor reviewer argv now always includes `--trust`, before `--print`, so a fresh checkout cannot block waiting for an interactive workspace-trust response. `TestReviewCommandBuildsHeadlessAskInvocation` asserts the flag.

frontend/src/landing/content/docs/configuration/projects.mdx
nikhilachale: Addressed in dd42b911fca88844da59e18d91cce8fc50224225. The project configuration documentation now lists all six values from `domain.AllReviewerHarnesses`: `claude-code`, `codex`, `copilot`, `cursor`, `kilocode`, and `opencode`.

nikhilachale:
Replying to review 4824481397: addressed both findings in dd42b911fca88844da59e18d91cce8fc50224225. Fresh headless Cursor reviewer launches now pass `--trust` (with argv coverage), and the reviewer documentation lists all six registered harnesses in domain order. Focused reviewer/review/domain tests pass, full `npm run lint` passes with 0 issues, and `git diff --check` is clean.

nikhilachale:
Updated Cursor reviewer lifecycle in 3853e0cfa90d90bf21b663d0b9af752961572b9e. Cursor now runs as a persistent interactive Ask-mode TUI with `--sandbox enabled`, `--trust`, and exact prompt-root `--add-dir`; print/output flags and one-shot respawn behavior are removed. AO installs an ownership-guarded reviewer profile under its data dir with the requested allow/deny tokens, while relying on Cursor 2.5 Ask-mode read-only sandboxing rather than claiming unsupported git-subcommand permission granularity. Focused backend/reviewer/HTTP tests, ProjectSettingsForm (18 tests), frontend typechecks, full lint, Electron package build, and `git diff --check` pass. The live destructive spike was not run because no disposable safe PR was available.

nikhilachale:
https://github.com/user-attachments/assets/efd151f3-abf4-4ab8-ae0f-0e522f7623c2


https://github.com/user-attachments/assets/0a6b42d6-f775-4144-b6b6-cce6f293dba8

nikhilachale:
Updated the Cursor reviewer again in b5bdafb13c854e3d95bdd7798ad9ae8daae67033 per the selected design. It now uses PermissionModeDefault and a normal persistent interactive TUI with only --trust and conditional --add-dir; all mode, sandbox, force/yolo, print, and output-format flags are absent. PreLaunch writes the exact version-1 permission policy atomically into the stable hashed AO-owned CURSOR_DATA_DIR profile, without touching checkout or user config. The PR body records the installed CLI compatibility caveat that this Cursor version resolves cli-config.json through CURSOR_CONFIG_DIR, which this design intentionally does not set. Focused/full local verification passes.

i-trytoohard:
From the NotesBot call on `2026-07-31`, this existing pr appears to be relevant.
Match confidence: `0.67` (fuzzy).

Relevant discussion excerpts:
- - *Dhruv Sharma:* described e2e/agent testing as a possible way to validate whether review findings are real, but *prateek* pushed to think about this holistically at the PR level rather than only coupling it to reviews. - *Nikhil Achale:* said he added 3 more reviewer adapters — **KiloCode, Cursor, and GitHub [...]
- - *Nikhil Achale:* said he added 3 more reviewer adapters — **KiloCode, Cursor, and GitHub Copilot** — and also worked on agent model selection and bulk scroll improvements. - *prateek:* asked why reviewer support cannot be expanded to all available adapters at once if reviewer spawning is mainly “spawning the [...]
- - *prateek:* said review UX must let users use whichever harness they want and should not require much work. - *Dhruv Sharma:* described e2e/agent testing as a possible way to validate whether review findings are real, but *prateek* pushed to think about this holistically at the PR level rather than only coupling [...]

Source thread: https://discord.com/channels/1476302178913357958/1532816669167063130/1532816669167063130

backend/internal/adapters/reviewer/registry.go
```diff
@@ -23,9 +43,29 @@ type Adapter interface {
 // here (and to domain.AllReviewerHarnesses) to register it.
 func Constructors() []Adapter {
 	return []Adapter{
+		aider.New(),
```
neversettle17-101, line 47: We should also add an API to show available reviewer agent harness -- we can create a followup ticket for the same.

backend/internal/adapters/reviewer/qwen/qwen.go
neversettle17-101, line 1: Qwen reviewer reuse keeps stale gateway authority — ReviewCommand writes   
AO_REVIEW_GATEWAY_MANIFEST once at launch and the manifest contains only the     
current RunID/PR/target SHA, but ReviewMessage just returns the next prompt and  
does not refresh that manifest. Because the launcher reuses panes by default when
an adapter does not implement ReviewerReusePolicy, a later Qwen review can run   
with a stale manifest from the previous task. This can block structured          
submission for the new run or authorize the wrong task context

neversettle17-101 (changes requested):

neversettle17-101 (commented):
Re-reviewed the current head and found two issues:

- **Medium**: Qwen gateway manifest authorizes only the first run in a multi-PR review batch.
  `backend/internal/adapters/reviewer/qwen/qwen.go:99` builds the manifest `Tasks` from only `inv.RunID`, `inv.PRURL`, and `inv.TargetSHA`. But the shared review prompt tells reviewers to complete every PR in `ReviewQueue` and submit one result per run (`backend/internal/review/prompt.go:30`), and the engine passes the full queue into the launch spec (`backend/internal/review/review.go:291`). For Qwen, any queued run after the first is not present in `AO_REVIEW_GATEWAY_MANIFEST`, so gateway-backed submit/review operations for later runs can be rejected or lack authority. Please build Qwen manifest tasks from `inv.ReviewQueue`, falling back to the single invocation, similar to Pi's `activateManifest` in `backend/internal/adapters/reviewer/pi/pi.go:157`, and add a multi-run test.

- **Low**: README lifecycle docs are stale after the Qwen non-reuse fix.
  `README.md:155` says only Aider and Auggie open a fresh process and “the other reviewers reuse their panes.” Current Qwen explicitly returns non-reusable because its manifest is launch-scoped (`backend/internal/adapters/reviewer/qwen/qwen.go:83`). Please update the README to list Qwen with the fresh-process reviewers.

I did not run local tests for this review. Current GitHub checks are green on the PR head.

neversettle17-101:
We also need to implement restore functionality on top of all agent harnesses except claude code, opencode and codex as it will be handled by this PR: https://github.com/Untrivial-ai/agent-orchestrator/pull/3484/changes

neversettle17-101 (commented):
Auggie looks inconsistent with the restore/reuse policy. The README says Aider, Auggie, and Qwen open a fresh process for each pass, and the Auggie reviewer package describes itself as one-shot, but Auggie does not implement `ReviewProcessReusable() false`. Because `Launcher.Reusable` defaults to true when no policy is present, the engine can `Notify` or `Restore` a recorded Auggie reviewer pane instead of forcing a fresh spawn. Either Auggie should opt out of reuse like Aider/Qwen, ideally with an engine test for non-reusable reviewers, or the docs/comment should stop claiming Auggie is fresh-per-pass.

nikhilachale:
Addressed the Auggie reuse feedback in commit `4956f69bb4aebeeab4ae976b34987bf4ced979c9`, which is included in the current head `d067eb2a8079a77e8efa8f82d673c062193bc2e6`.

- Auggie now implements `ReviewerReusePolicy` and returns `false` from `ReviewProcessReusable()`.
- The adapter test asserts Auggie is non-reusable.
- `TestTriggerSpawnsFreshPassForNonReusableReviewer` proves the engine spawns a fresh Auggie process without checking liveness, restoring, or notifying the recorded pane.

Verification: `cd backend && go test -count=1 ./internal/adapters/reviewer/auggie ./internal/review ./internal/adapters/reviewer/qwen ./internal/adapters/reviewer` (pass). The branch was already synchronized with origin; push reports no additional objects required.

backend/internal/review/launcher.go
```diff
@@ -260,6 +443,15 @@ func (l *agentLauncher) Alive(ctx context.Context, handleID string) (bool, error
...
+func (l *agentLauncher) Reusable(harness domain.ReviewerHarness) bool {
+	reviewer, ok := l.reviewers.Reviewer(harness)
+	if !ok {
+		return false
+	}
+	policy, ok := reviewer.(ports.ReviewerReusePolicy)
+	return !ok || policy.ReviewProcessReusable()
+}
+
 func (l *agentLauncher) Cancel(ctx context.Context, handleID string, harness domain.ReviewerHarness) error {
```
neversettle17-101, line 543: This cancel primitive now actively sends Ctrl-C/Escape to the reviewer pane, so the caller needs to avoid invoking it when there are no running review runs. Today `Engine.Cancel` loads `ListRunningReviewRunsBySession` but still calls `launcher.Cancel(...)` even when that slice is empty, which means the CLI/API can interrupt an idle reviewer terminal even though the desktop UI normally hides the action. Please add a `len(running) == 0` guard before calling into the launcher and cover that path with a test.

neversettle17-101 (commented):
Two remaining issues from the re-review:

1. `frontend/src/renderer/hooks/useWorkspaceQuery.ts` still drops every session reviewer override except `claude-code`, `codex`, `opencode`, and `muse`. With the new reviewer set, a session switched to `qwen`, `goose`, `aider`, etc. comes back from `/sessions` with that saved `reviewerHarness`, but the frontend maps it to `undefined`, so the inspector can display/use the project default on the next refetch. Please preserve the full reviewer harness set here, e.g. by validating against the same known reviewer ids used by `ReviewerSelect`, and widen the `WorkspaceSession.reviewerHarness` type accordingly.

2. `backend/internal/review/launcher.go` builds the idle restore invocation without `DataDir`. That value is later passed through `agentrestore.Command` into the wrapped agent restore config, and Muse uses it to compute/install its managed hook path. With it empty, restored Muse reviewer panes can fail restore or stop reporting the native session id needed for later restores. Please set `DataDir: l.dataDir` in the idle invocation and add a restore test that asserts the restored invocation includes it.

backend/internal/domain/session.go
```diff
@@ -67,7 +67,7 @@ type SessionRecord struct {
 	Harness   AgentHarness `json:"harness,omitempty"`
 	// ReviewerHarness is this session's preferred reviewer. Empty delegates to
 	// the project configuration.
-	ReviewerHarness ReviewerHarness `json:"reviewerHarness,omitempty" enum:"claude-code,codex,opencode"`
+	ReviewerHarness ReviewerHarness `json:"reviewerHarness,omitempty" enum:"claude-code,codex,opencode,muse"`
```
neversettle17-101: We would need to add all reviewer agent harness name in the enum

neversettle17-101 (changes requested):

neversettle17-101 (approved):

Issue #3385, opened 2026-07-31
[Feature]: Add adapter-aware worker and orchestrator model selection


nikhilachale:
## Problem

AO lets users choose separate worker and orchestrator agents, but it does not provide a reliable, role-specific way to select the model used by each agent.

Users currently have to know and manually enter adapter-specific model identifiers. The available models differ by agent:

- Claude Code and Codex have their own supported model families and aliases.
- Some agent CLIs expose a non-interactive model-list command.
- Some adapters accept a model override but do not expose a discoverable catalog.
- Amp exposes execution modes (`low`, `medium`, `high`, and `ultra`) rather than model identifiers.

A single hard-coded global list will become stale and cannot accurately represent all adapters. Querying every CLI on every settings render would also be slow and unnecessarily expensive.

## Proposed direction

Add an adapter-aware model selector to Project Settings, with independent values for the worker and orchestrator roles.

AO should use three catalog strategies:

1. **Curated catalog** for agents with stable, documented model aliases or product-specific models.
2. **CLI discovery** when the installed agent exposes a documented, non-interactive model-list command.
3. **Free-text fallback** when AO cannot reliably discover a model list.

Amp should use a constrained mode selector instead of a model input.

### User experience

- Show a model field immediately below the selected worker agent.
- Show a separate model field immediately below the selected orchestrator agent.
- For a catalog, provide searchable suggestions while still allowing custom IDs when supported.
- For adapters without discovery, provide a free-text input.
- For Amp, provide a mode menu with `low`, `medium`, `high`, and `ultra`.
- Selecting the empty/default value should let the agent choose its own default.
- Changing an agent should clear the model/mode previously selected for that role.
- Provide a refresh action for each role's catalog.
- If refresh fails, retain the last successful cached catalog and show a warning.

## Adapter coverage

AO currently ships 23 production agent adapters. The intended initial strategy is:

### Curated catalog

- Claude Code
- Codex

### Mode selector

- Amp

### CLI-discovered catalog

- AGY
- Aider
- Auggie
- Autohand
- Cursor
- Devin
- Grok
- Kimi
- Kiro
- Kilo Code
- OpenCode
- Pi

### Free-text fallback

- Cline
- Continue
- Copilot
- Crush
- Droid
- Goose
- Qwen
- Vibe

Discovery commands should be kept in the backend and executed only with static adapter-defined arguments. A failed or unsupported command must never prevent project settings from being used.

## Backend and cache design

[...]

PR #3386, opened 2026-07-31
feat: add adapter-aware model selection

nikhilachale:
## Summary

Adds adapter-aware model and mode selection for independent worker and orchestrator roles. AO loads catalogs only for agents selected in Project Settings, combines curated values, safe CLI discovery, and declarative local configuration, then stores the normalized result in a project-scoped cache.

Users get a searchable selector when models are known and a custom text field when they are not. Worker and orchestrator choices remain independent.

## User-facing behavior

- Worker and orchestrator each have their own model or mode control.
- Claude Code and Codex use curated model catalogs.
- Amp exposes low, medium, high, and ultra modes.
- Supported CLI-backed agents discover models lazily through documented non-interactive commands.
- Crush, Continue, OpenCode, Qwen, and Kimi also contribute configured/custom models from safe local config parsing.
- Catalogs are searchable by model label or ID and retain a Custom model escape hatch.
- Agents without a reliable catalog use free-text input.
- Refresh failures retain the last useful catalog and show a warning.

## Discovery and caching

AO does not invoke all 23 adapters when settings opens. The frontend requests catalogs only for the currently selected worker and orchestrator agents.

The daemon merges and deduplicates model metadata from these sources:

| Source | Agents/use |
|---|---|
| Curated catalog | Claude Code, Codex |
| Curated mode list | Amp |
| Isolated CLI discovery | Agents with documented non-interactive list commands |
| Declarative config | Crush, Continue, OpenCode, Qwen, Kimi |
| Manual fallback | Agents with no reliable discovery source |

Catalogs are cached in SQLite by agent and project. Cache invalidation accounts for the installed binary version and a non-sensitive config metadata fingerprint. Explicit refresh bypasses the cache.

If CLI discovery fails but configuration yields models, AO returns and caches the partial catalog with a warning. If discovery fails with a previous cache, the previous catalog is returned as stale. If no usable result or cache exists, AO returns the free-text fallback contract.

## Portable configuration discovery

Config candidates follow each agent's documented precedence and are derived from the runtime OS and environment instead of this contributor's machine. Candidate paths are deduplicated and included in cache invalidation.

| Agent | Supported config locations |
|---|---|
| Crush | `CRUSH_GLOBAL_CONFIG`, `CRUSH_GLOBAL_DATA`, XDG/default user config and data, Windows `%LOCALAPPDATA%`, bounded project `crush.json`/`.crush.json`, current `crush.json`, and legacy Catwalk `providers.json`/`CRUSH_DATA_DIR` |
| Continue | OS user home `~/.continue/config.yaml` (plus `.yml` compatibility) and the existing project-local `.continue` compatibility files; Continue does not publish a stable filesystem pointer for the last `/config` selection |
| OpenCode | XDG/default global config, `OPENCODE_CONFIG`, `OPENCODE_CONFIG_DIR`, bounded nested project `opencode.json(c)`, `.opencode/opencode.json(c)`, and Linux/macOS/Windows managed JSON directories |
| Qwen | `QWEN_HOME`, project `.qwen/settings.json`, system defaults/settings on Linux/macOS/Windows, and both `QWEN_CODE_SYSTEM_*_PATH` overrides |
| Kimi | `KIMI_CODE_HOME/config.toml` or the portable `~/.kimi-code/config.toml` default |

The parsers support current and legacy Crush layouts and current/legacy Qwen `modelProviders` layouts, producing model IDs in the format each adapter can launch.

## Configuration security

Configuration discovery reads only bounded adapter-owned paths and extracts model ID, label, provider, and default state.

- JSON, JSONC, YAML, and standard TOML parsing
- 4 MiB maximum file size
- regular files only; symbolic links and unsafe traversal are rejected
- no execution of plugins, commands, imports, substitutions, or file references
- no caching of API keys, tokens, headers, endpoints, credentials, or raw config
- OpenCode CLI discovery always runs with `--pure` so project plugins cannot execute

## Backend and API

- Adds project-scoped agent model catalog persistence using migration 0041.
- Adds `GET /api/v1/agents/{agent}/models`.
- Adds `POST /api/v1/agents/{agent}/models/refresh`.
- Preserves model and Amp mode through domain, API, project configuration, and the CLI hand-mirrored DTO.
- Completes model forwarding for affected launch and restore adapters.
- Keeps legacy project-level model values readable and migrates displayed values to `worker.agentConfig` and `orchestrator.agentConfig` on save without dropping shared permissions.

## Intentional omissions

- No startup refresh of every adapter.
- No provider-account authorization check for individual models.
- No installation of missing agent CLIs.
- No automatic replacement of deprecated saved models.
- No filesystem-wide search or guessed undocumented config paths.
- Continue's saved runtime config selection is not read because Continue does not expose a stable documented local pointer for it; the documented default remains supported.
- Aider, Autohand, and other adapters without safe complete discovery remain free-text.

## Verification

Completed locally:

- focused backend tests for all five config-backed adapters
- race tests for model catalog and agent service
- cross-compilation of model catalog tests for Windows and Linux from macOS
- `golangci-lint` v2.12.2: 0 issues
- full backend package run; unrelated timing-sensitive fake/Kilo/OpenCode/tmux tests failed only under concurrent suite load and passed when rerun independently
- prior API generation, frontend typecheck/tests, renderer smoke, storage migration, and CLI round-trip coverage from this PR

Coverage includes platform and environment path precedence, nested Git-worktree lookup, current/legacy Crush and Qwen parsing, JSONC comments/trailing commas, YAML, TOML escapes, credential exclusion, file-size limits, symlink rejection, source merging, partial fallback, cache invalidation, API/service behavior, searchable filtering, refresh warnings, CLI model/mode round trips, migration behavior, and launch/restore forwarding.


backend/internal/domain/agentconfig.go
```diff
@@ -25,6 +25,9 @@ const (
 type AgentConfig struct {
 	// Model overrides the agent's default model (e.g. claude-opus-4-5).
 	Model string `json:"model,omitempty"`
+	// Mode selects an agent-owned operating mode when the adapter exposes modes
+	// instead of raw model ids (currently Amp: low|medium|high|ultra).
+	Mode string `json:"mode,omitempty"`
```
illegalcall, line 30: **[P1] Preserve `mode` across the CLI's hand-mirrored DTO.** `backend/internal/cli/project.go` still defines `agentConfig` with only `model` and `permissions` (lines 76–80). As a result, `ao project get --json` omits a saved Amp mode, and `ao project set-config --config-json ...` unmarshals then reserializes the full config without it, silently clearing the mode. Please add the `mode` field to the CLI mirror and cover both JSON read and set-config round trips.

backend/internal/adapters/agent/modelcatalog/catalog.go
```diff
@@ -0,0 +1,361 @@
...
+
+type commandSpec struct {
+	args   []string
+	parser func([]byte) ([]ports.AgentModelInfo, error)
+}
+
+var ansiPattern = regexp.MustCompile(`\x1b\[[0-9;]*[[:alpha:]]`)
+
+var commandSpecs = map[string]commandSpec{
+	"opencode": {args: []string{"models"}, parser: parseIDLines},
```
illegalcall: **[P1/security] Do not run OpenCode discovery with external project plugins enabled.** Discovery later sets `cmd.Dir` to the registered project path, and Project Settings automatically starts this query. OpenCode automatically loads JavaScript/TypeScript from `.opencode/plugins/` at startup, and plugin initializers receive Bun's shell API, so viewing settings for an untrusted repository can execute repository-controlled code. Run this discovery in OpenCode's global `--pure` mode (or provide equivalent isolation) and test the safe invocation. Official behavior: https://opencode.ai/docs/plugins/ and https://opencode.ai/docs/cli/#global-flags

backend/internal/service/agent/service.go
```diff
@@ -172,6 +200,121 @@ func (s *Service) Probe(ctx context.Context, agentID string) (ProbeResult, error
...
+	discovered.BinaryVersion = version
+	if discoverErr != nil {
+		if hasCached {
+			cached.Catalog.Stale = true
+			cached.Catalog.Warning = discoverErr.Error()
+			return cached.Catalog, nil
+		}
+		discovered.Stale = true
+		discovered.Warning = discoverErr.Error()
+		return discovered, nil
```
illegalcall: **[P2] Return the specified free-text fallback when discovery fails without a cache.** `Discover` returns `Base(agentID)` on error; for command-backed agents that base still has `selectionMode: "catalog"`, an empty model list, and source `"cli"`. The issue's backend contract explicitly requires `selectionMode: "text"` / custom-value fallback when no previous catalog exists. The frontend currently compensates for the empty list, but other API clients will receive the wrong mode. Convert this no-cache error result to the manual text fallback and add an API/service assertion for it.

frontend/src/renderer/components/ProjectSettingsForm.tsx
```diff
@@ -429,6 +479,191 @@ function SaveChangesFooter({
...
+	const query = useQuery(agentModelsQueryOptions(agentId, projectId));
+	const refreshMutation = useMutation({
+		mutationFn: () => refreshAgentModels(agentId, projectId),
+		onSuccess: (catalog) => queryClient.setQueryData(agentModelsQueryKey(agentId, projectId), catalog),
+	});
+	const catalog: AgentModelCatalog | undefined = query.data;
+	const isMode = catalog?.selectionMode === "mode";
+	const label = `${role} ${isMode ? "mode" : "model"}`;
+	const datalistID = `${role.toLowerCase()}-model-options`;
+	const warning =
```
illegalcall, line 544: **[P2] Surface refresh request failures as the required warning.** This warning only considers the catalog payload and the initial query error; `refreshMutation.error` is ignored. If the refresh POST fails at the transport/controller layer, React Query keeps the last catalog but the UI shows no warning, contrary to issue #3385's “retain the last successful cached catalog and show a warning” requirement. Include the refresh mutation error here and cover the rejected-refresh path.

frontend/src/renderer/components/ProjectSettingsForm.tsx
```diff
@@ -429,6 +479,191 @@ function SaveChangesFooter({
...
+			onModelChange(value === "__default__" ? "" : value);
+		}
+		onModeChange("");
+	};
+	return (
+		<>
+			<SettingsRow icon={Sparkles} label={label}>
+				<div className="flex min-w-0 items-center gap-2">
+					{hasCatalog && !showCustomInput ? (
+						<SettingsOptionMenu
```
illegalcall: **[P2] Catalog suggestions are required to be searchable.** `SettingsOptionMenu` only renders every option in a dropdown; it has no query input or filtering. Agent/provider catalogs can be large, so this does not meet issue #3385's explicit “provide searchable suggestions” UX requirement. Please use a searchable combobox or add filtering to this catalog path, with a test that narrows options by label/ID.

illegalcall (changes requested):
Requesting changes for five concrete correctness/security gaps. Each is anchored inline. CI is green at `a462c39`, but the current behavior can execute project plugin code during settings discovery, drop an Amp mode through the CLI, return the wrong no-cache fallback contract, omit the required searchable catalog UI, and hide refresh transport/server failures.

nikhilachale:
Addressed the latest model-catalog review findings in ccb13e780:

- Replaced the custom Kimi TOML scanner with pelletier/go-toml v2, covering standard TOML escapes and removing the model-section regex/ReDoS surface.
- Added explicit coverage for JSONC line comments ending at EOF.
- Reworked config paths as trusted root + relative path and now uses os.Root traversal, rejects symlink/non-directory components, verifies each opened directory inode, and verifies the final regular-file inode to prevent escape and TOCTOU swaps.
- Added per-adapter resolver serialization across Models, Refresh, and Probe, with concurrent race coverage.
- Cache upsert failures no longer discard a successfully discovered catalog; the response is retained with a user-visible warning.
- Kept the existing model-ID character policy because custom provider IDs are intentionally supported; catalog parsing remains bounded by the 4 MiB input limit and normalized/deduplicated output.

Verification: focused go test -race passed for modelcatalog and service/agent; Windows and Linux cross-compilation passed; go mod tidy -diff is clean; golangci-lint reports no issues. Fresh CI is running.

backend/internal/storage/sqlite/migrations/0047_agent_model_catalog.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
```
illegalcall, line 1: **[P1] Renumber the catalog migration past main's 0041.** Current `main` already has `backend/internal/storage/sqlite/migrations/0041_notification_resolution.sql`. After rebasing, both files have Goose version 41, so the duplicate-version check fails and blocks CI and daemon startup. Renumber this as the next migration, as required by `AGENTS.md`.

backend/internal/service/agent/service.go
```diff
@@ -172,6 +225,217 @@ func (s *Service) Probe(ctx context.Context, agentID string) (ProbeResult, error
...
+		return "", nil
+	}
+	project, ok, err := s.projects.GetProject(ctx, projectID)
+	if err != nil {
+		return "", apierr.Internal("PROJECT_LOAD_FAILED", "Failed to load project")
+	}
+	if !ok || !project.ArchivedAt.IsZero() {
+		return "", apierr.NotFound("PROJECT_NOT_FOUND", "Unknown project")
+	}
+	return project.Path, nil
```
illegalcall: **[P2] Include the project environment in discovery.** When a project sets `OPENCODE_CONFIG`, `QWEN_HOME`, or another supported override in `ProjectConfig.Env`, launched sessions receive that environment, but this lookup discards `project.Config` and discovery uses the daemon's environment. The project-scoped endpoint therefore fingerprints and caches models from different files than the agent will load; pass the project environment into config and CLI discovery.

backend/internal/service/agent/service.go
```diff
@@ -172,6 +225,217 @@ func (s *Service) Probe(ctx context.Context, agentID string) (ProbeResult, error
...
+		}
+		return cached.Catalog, nil
+	}
+
+	discovered, discoverErr := modelcatalog.Discover(ctx, agentID, binary, workingDir)
+	discovered.BinaryVersion = version
+	discovered.ValidatedAt = time.Now().UTC()
+	discovered.RefreshRecommended = false
+	if discoverErr != nil {
+		if len(discovered.Models) > 0 {
```
illegalcall, line 334: **[P2] Preserve a fuller cached catalog on refresh failure.** If CLI discovery transiently fails while config parsing yields any models, this branch runs before `hasCached`, saves the partial config-only result, and returns it. A previously cached full CLI catalog therefore loses all non-configured suggestions after a failed refresh, contrary to the requirement that refresh failures retain the last useful catalog; prefer the useful cache or merge the partial result.

backend/internal/adapters/agent/modelcatalog/config.go
```diff
@@ -0,0 +1,913 @@
...
+		}
+		if !found {
+			continue
+		}
+		parsed, err := spec.parser(data)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("parse %s: %w", filepath.Base(path.name), err))
+			continue
+		}
+		models = append(models, parsed...)
```
illegalcall: **[P2] Apply later config layers before deduplicating models.** For layered configs, every parsed layer is appended, while `normalize` preserves earlier labels/providers and ORs `isDefault`. For example, global and project OpenCode configs selecting different models return both as defaults rather than reflecting the project override. Merge layers using their documented precedence before normalizing.

backend/internal/adapters/agent/modelcatalog/config.go
```diff
@@ -0,0 +1,913 @@
...
+	return provider + "/" + model
+}
+
+func parseContinueConfig(data []byte) ([]ports.AgentModelInfo, error) {
+	var root yaml.Node
+	if err := yaml.Unmarshal(data, &root); err != nil {
+		return nil, err
+	}
+	var models []ports.AgentModelInfo
+	walkYAMLMaps(&root, func(node *yaml.Node) {
```
illegalcall: **[P2] Exclude non-chat Continue models from the selector.** Continue YAML can contain autocomplete, embedding, and reranking models distinguished by `roles`, while legacy files contain similarly shaped maps. This recursive walk accepts every mapping with a `model` field and ignores its section or roles, so unusable non-chat IDs are offered and forwarded to `cn --model`; restrict extraction to chat-capable entries.

backend/internal/httpd/controllers/agents.go
```diff
@@ -29,6 +32,41 @@ func (c *AgentsController) Register(r chi.Router) {
...
+	c.writeModels(w, r, false, false)
+}
+
+func (c *AgentsController) refreshModels(w http.ResponseWriter, r *http.Request) {
+	c.writeModels(w, r, true, r.URL.Query().Get("revalidate") == "true")
+}
+
+func (c *AgentsController) writeModels(w http.ResponseWriter, r *http.Request, refresh, revalidate bool) {
+	if c.Catalog == nil {
+		apispec.NotImplemented(w, r, r.Method, "/api/v1/agents/{agent}/models")
```
illegalcall: **[P2] Use the refresh route in the 501 response.** When the agents dependency is unavailable, POSTing to `/agents/{agent}/models/refresh` asks `apispec.NotImplemented` for the nonexistent operation `POST /api/v1/agents/{agent}/models`. `NotImplemented` consequently panics instead of returning a 501 JSON envelope; use the actual refresh route template, preserving the API error envelope required by `AGENTS.md`.

backend/internal/adapters/agent/modelcatalog/config.go
```diff
@@ -0,0 +1,913 @@
...
+	}
+	for left, right := 0, len(reversed)-1; left < right; left, right = left+1, right-1 {
+		reversed[left], reversed[right] = reversed[right], reversed[left]
+	}
+	return reversed
+}
+
+// ConfigModels reads only declarative, bounded agent configuration files. It
+// never expands references, executes plugins, or retains unrelated fields.
+func ConfigModels(agentID, workingDir string) ([]ports.AgentModelInfo, error) {
```
illegalcall: **[P2] Thread the load context through config-file I/O.** On slow or network-backed filesystems, `ConfigModels` and `ConfigVersion` walk, open, and parse all candidates without the model-load context, so the 30-second timeout can expire while the coalesced goroutine continues blocking work. Pass `context.Context` through these operations, as required for I/O and blocking work by `AGENTS.md`.

illegalcall (commented):
Seven additional correctness and robustness issues found in the model catalog patch:

backend/internal/service/agent/service.go
```diff
@@ -2,21 +2,44 @@ package agent
...
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"sort"
+	"strconv"
 	"sync"
 	"time"
 
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/modelcatalog"
```
illegalcall: **[P2] Keep model discovery behind a port boundary.** This makes the core `service/agent` package depend directly on the concrete `adapters/agent/modelcatalog` implementation and call its discovery, fingerprinting, and fallback helpers. `docs/architecture.md` says core services access external systems through interfaces in `internal/ports`; inject an `AgentModelDiscoverer` (or equivalent narrow port) through `Deps` and wire the concrete adapter in `daemon`.

backend/internal/service/agent/service.go
```diff
@@ -172,6 +225,232 @@ func (s *Service) Probe(ctx context.Context, agentID string) (ProbeResult, error
...
+	}
+	discovery, err := s.projectDiscoveryContext(ctx, projectID)
+	if err != nil {
+		return ports.AgentModelCatalog{}, err
+	}
+	cached, hasCached, err := s.cachedCatalog(ctx, agentID, projectID)
+	if err != nil {
+		return ports.AgentModelCatalog{}, err
+	}
+	if hasCached && mode == modelLoadCached {
```
illegalcall: **[P2] Make binary-version invalidation part of the daemon cache contract.** This cache-hit return happens before the binary is resolved or fingerprinted, so `GET /models` can serve an entry indefinitely after the installed agent changes; only this React frontend's later revalidation POST notices it. Conversely, `modelCatalogDiscoveryTTL` forces rediscovery after six hours even when the binary is unchanged. Issue #3385 requires catalogs to be reused until manual refresh or an installed-version change and to rediscover automatically on that change, so the daemon read path should perform or schedule that version check rather than relying on one client and a time-based expiry.

backend/internal/adapters/agent/modelcatalog/catalog.go
```diff
@@ -0,0 +1,461 @@
...
+	"grok":     {args: []string{"models"}, parser: parseGrokModels},
+	"cursor":   {args: []string{"models"}, parser: parseCursorModels},
+	"agy":      {args: []string{"models"}, parser: parseIDLines},
+	"kilocode": {args: []string{"models"}, parser: parseIDLines},
+	"pi":       {args: []string{"--list-models"}, parser: parsePiModels},
+	"kimi":     {args: []string{"provider", "list", "--json"}, parser: parseJSONModels},
+	"auggie":   {args: []string{"models", "list", "--json"}, parser: parseJSONModels},
+	"devin":    {args: []string{"models", "list", "--format", "json"}, parser: parseJSONModels},
+	"kiro":     {args: []string{"chat", "--list-models", "--format", "json"}, parser: parseJSONModels},
+}
```
illegalcall, line 56: **[P2] Implement the specified Aider and Autohand discovery coverage.** Issue #3385 lists both under CLI-discovered catalogs, but neither has a `commandSpec`, so both silently fall back to free text. Aider documents `--list-models MODEL` (`--models MODEL`) in its official CLI reference: https://aider.chat/docs/config/options.html#--list-models-model. Please add safe discovery/parsing for these adapters, or get the issue scope explicitly changed before treating the omission as intentional.

backend/internal/adapters/agent/aider/aider.go
```diff
@@ -84,6 +98,12 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 	return cmd, nil
 }
 
+func appendModelFlag(cmd *[]string, cfg ports.AgentConfig) {
```
illegalcall: **[P3] Centralize the repeated model-flag plumbing.** This exact `appendModelFlag` implementation, along with nearly identical model `ConfigSpec` construction, is repeated in Aider, Continue, Cursor, Devin, Grok, and OpenCode. This is now real cross-adapter duplication; a small helper in `agentbase` would keep model forwarding consistent without introducing a one-off abstraction.

illegalcall (commented):
Re-review against current `main`: the branch is still conflicting in `backend/internal/adapters/agent/grok/grok.go` and `backend/internal/adapters/agent/registry/registry_test.go`, so it needs a rebase before it can merge. Also, the config-file discovery subsystem materially expands issue #3385 beyond its three requested strategies (curated, CLI discovery, and free text); please confirm that added scope with the maintainers. I left new inline findings below and followed up on the two existing threads whose fixes remain partial.

backend/internal/adapters/agent/modelcatalog/config.go
illegalcall: Thanks for threading `context.Context` into the public config functions. The original timeout failure mode is still only partially addressed: the checks run before or after `Lstat`, `EvalSymlinks`, `Open`, `Stat`, and `ReadAll`, but cannot run while one of those calls is blocked, so the coalesced load can remain in `modelCalls` after its 30-second context expires. Please isolate the blocking scan behind a context select (or otherwise ensure `loadModels` returns at the deadline). Also, `configSpec.paths` still buries the context inside `configPathContext`; per `AGENTS.md`, make it an explicit first argument for this I/O-performing path walk.

backend/internal/service/agent/service.go
illegalcall: The new model-count heuristic does not fully resolve this. `hasCached && len(cached.Models) > len(discovered.Models)` preserves only a *strictly larger* cache. I reproduced an equal-count case where a cached one-model CLI catalog is replaced and persisted by a different one-model config-only result after the binary fails. Model count is not a proxy for “last successful catalog”; on any CLI discovery failure, prefer the existing cache or merge the partial config result into it by ID/source.

illegalcall (approved):

PR #2947, opened 2026-07-21
fix: reduce telemetry event volume

AgentWrapper:
## Summary

Reduces PostHog event volume and separates actual user activity from agent/system background activity:

- persists CLI telemetry reservations under `DataDir` so daemon restarts cannot re-emit every polling command or active heartbeat slot for the same install/day
- classifies CLI invocations as `actor_type=user|agent|system`
- keeps `ao hooks` and commands run with `AO_SESSION_ID` as agent activity, but prevents them from refreshing `ao.app.active` / DAU
- excludes internal system CLI processes such as `ao pty-host`
- changes `ao.app.active` to a bounded heartbeat: at most once per six-hour UTC slot, or four times per day per install/channel
- caps renderer `ao.renderer.route_viewed` to once per coarse surface per UTC day per renderer install
- forces renderer PostHog captures to include `$process_person_profile: false`, including exceptions, and disables Web Vitals capture
- documents the read-only PostHog investigation, product metrics model, top event counts, ratios, and projected savings in `docs/telemetry.md`

## Product metrics model

Current AO telemetry has a stable install ID, not a signed-in account user ID. That means today's DAU/MAU can accurately represent active installs, but not unique people across multiple machines. True user-level new/churn/journey metrics require an explicit stable user identity from a login, license, or workspace account system. That identity should be sent as a first-party AO user ID or one-way hash only after auth/consent; it should not be inferred from machine fingerprints, local paths, git remotes, or repo config.

Active-user events should come from human/product activity only:

- renderer heartbeat when a human uses the desktop app
- user-context CLI commands

`ao.app.active` remains the active-user event for DAU/WAU/MAU. It is now emitted at most once per six-hour UTC slot, so unique-user aggregation stays bounded while arbitrary rolling windows are less likely to undercount someone who is active later in the same UTC day.

These should not drive DAU/MAU:

- `ao hooks` agent callbacks
- CLI commands run inside AO-managed agent sessions (`AO_SESSION_ID` set)
- internal runtime/supervisor commands such as `ao pty-host`, `ao daemon`, `ao start`
- raw polling frequency for read-only state commands

Agent CLI activity remains useful, but as command/agent-activity signal, not active-user signal.

## PostHog investigation: trailing 30 days as of 2026-07-21

Read-only HogQL against project `475752` found 3,203,364 total events. Top events:

| Event | Count | Installs | Events/install |
| --- | ---: | ---: | ---: |
[...]

PR #3388, opened 2026-07-31
fix: harden PostHog telemetry controls

somewherelostt:
## Summary

Addresses Pulkit's post-merge telemetry follow-up notes from #3314 ( https://github.com/Untrivial-ai/agent-orchestrator/pull/3314#issuecomment-5137511549)

- explicitly disables PostHog session recording in renderer config and aligns the README with docs
- moves routine internal CLI command classification into a shared telemetry helper
- normalizes daemon-posted command paths before cost-control classification
- drops odd legacy/internal command shapes like `ao  hooks`, `AO HOOKS`, and `ao hooks claude-code post-tool-use`
- stops treating unknown legacy command paths without an explicit actor type as DAU-eligible user activity
- clarifies that the 10x+ spend estimate requires PostHog-side ingestion controls for already-deployed legacy clients

## Why

#3314 fixed the main PostHog spend issue, but the follow-up review found places where docs and code could drift or where old/foreign loopback callers could bypass the daemon-side routine-command gate through command-path variants.

## Validation

- `go test ./internal/cli ./internal/httpd ./internal/telemetrymeta ./internal/adapters/telemetry`
- `npm test -- --run src/renderer/lib/telemetry.test.ts`
- `npm run typecheck`

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ ao start
 
 ## Telemetry
 
-Agent Orchestrator's Electron renderer sends anonymous usage events to PostHog for reliability and product understanding, and PostHog session recording is enabled with local paths and local URLs redacted before transmission. Set `VITE_AO_POSTHOG_KEY` to an empty string before building to disable transmission. See [docs/telemetry.md](docs/telemetry.md).
+Agent Orchestrator's Electron renderer sends anonymous usage events to PostHog for reliability and product understanding. PostHog session recording is disabled by default; if a time-boxed investigation enables it, local paths and local URLs are redacted before transmission. Set `VITE_AO_POSTHOG_KEY` to an empty string before building to disable transmission. See [docs/telemetry.md](docs/telemetry.md).
 
 ## License
 
--- a/backend/internal/cli/root.go
+++ b/backend/internal/cli/root.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/daemon"
 	aoprocess "github.com/aoagents/agent-orchestrator/backend/internal/process"
 	"github.com/aoagents/agent-orchestrator/backend/internal/processalive"
+	"github.com/aoagents/agent-orchestrator/backend/internal/telemetrymeta"
 )
 
 // Execute runs the ao CLI with process stdio.
@@ -211,8 +212,8 @@ type commandContext struct {
 }
 
 func shouldEmitCLIInvocation(cmd *cobra.Command) bool {
-	commandPath := strings.TrimSpace(cmd.CommandPath())
-	if isRoutineInternalCLICommand(commandPath) {
+	commandPath := telemetrymeta.NormalizeCommandPath(cmd.CommandPath())
+	if telemetrymeta.IsRoutineInternalCLICommand(commandPath) {
 		return false
 	}
 	switch commandPath {
@@ -224,22 +225,6 @@ func shouldEmitCLIInvocation(cmd *cobra.Command) bool {
 		return false
 	default:
 		return true
-	}
-}
-
-func isRoutineInternalCLICommand(commandPath string) bool {
-	switch strings.TrimSpace(commandPath) {
-	case "ao status",
-		"ao session ls",
-		"ao session get",
-		"ao project ls",
-		"ao project get",
-		"ao orchestrator ls",
-		"ao hooks",
-		"ao pty-host":
-		return true
-	default:
-		return false
 	}
 }
 
--- a/backend/internal/cli/root_test.go
+++ b/backend/internal/cli/root_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/daemonmeta"
 	"github.com/aoagents/agent-orchestrator/backend/internal/runfile"
+	"github.com/aoagents/agent-orchestrator/backend/internal/telemetrymeta"
 )
 
 func TestRootHelpDoesNotShowDaemon(t *testing.T) {
@@ -147,6 +148,39 @@ func TestCLIInvocationActorType(t *testing.T) {
 	t.Setenv("AO_SESSION_ID", "ao-session-1")
 	if got := cliInvocationActorType(byName["status"]); got != "agent" {
 		t.Fatalf("status actor with session env = %q, want agent", got)
+	}
+}
+
+func TestTelemetryMetaClassifiesRegisteredCommandPaths(t *testing.T) {
+	systemCommands := map[string]struct{}{
+		"ao agent-process":           {},
+		"ao agent-process supervise": {},
+		"ao completion":              {},
+		"ao daemon":                  {},
+		"ao help":                    {},
+		"ao pty-host":                {},
+		"ao start":                   {},
+	}
+
+	var failures []string
+	var walk func(*cobra.Command)
+	walk = func(cmd *cobra.Command) {
+		for _, child := range cmd.Commands() {
+			path := telemetrymeta.NormalizeCommandPath(child.CommandPath())
+			got := telemetrymeta.CLIActorType("", path)
+			if got == "system" {
+				_, explicitlySystem := systemCommands[path]
+				if !explicitlySystem && !telemetrymeta.IsRoutineInternalCLICommand(path) {
+					failures = append(failures, path)
+				}
+			}
+			walk(child)
+		}
+	}
+	walk(NewRootCommand(Deps{}))
+
+	if len(failures) > 0 {
+		t.Fatalf("actor-less command paths classified as system: %s", strings.Join(failures, ", "))
 	}
 }
 
--- a/backend/internal/httpd/cli_telemetry_reservoir.go
+++ b/backend/internal/httpd/cli_telemetry_reservoir.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/telemetrymeta"
 )
 
 const cliTelemetryStateFile = "telemetry_cli_daily.json"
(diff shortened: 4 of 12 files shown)


illegalcall:
The current application-code changes look sufficient for this PR: command paths are normalized centrally, routine/internal commands (including prefixed hook variants) are dropped by both the CLI and daemon paths, unknown actor-less legacy commands no longer become user activity, session recording is explicitly disabled, and the relevant cases are tested.

There is one remaining mismatch in `docs/posthog-cost-controls.md` that should be corrected before merge. The documented PostHog rule still depends on `actor_type` being missing/null. That preserves the middle-generation leak: those clients populate `actor_type`, but still emit the legacy event names, so automated commands such as `ao hooks`, `ao session ls/get`, and `ao status` are allowed through.

Please update the document as follows.

### 1. Make the ingestion rule command-based, not actor-based

Define the normalized command path as lowercase, trimmed, with repeated whitespace collapsed. Treat a command as routine when the normalized path equals one of the following or begins with that path followed by a space:

- `ao hooks`
- `ao session ls`
- `ao session get`
- `ao project ls`
- `ao project get`
- `ao orchestrator ls`
- `ao status`
- `ao pty-host`

Then document these drop conditions:

```text
Drop ao.cli.invoked when routine_command is true,
regardless of actor_type.

Drop ao.app.active when channel = cli and routine_command is true,
regardless of actor_type.
```

The same routine-command check can be retained as a defensive backstop for the corresponding `ao.v2.*` event names, even though current clients should already suppress them before transmission.

`actor_type` should remain available for segmentation and analysis, but it should not be part of the cost-control predicate. The command describes the activity we want to exclude; the actor field changed between client generations and is therefore not a reliable ingestion boundary.

Examples that should be explicitly covered by the documented rule:

| Event | Command path | Actor | Result |
|---|---|---|---|
| `ao.cli.invoked` | `ao hooks` | `agent` | Drop |
| `ao.cli.invoked` | `AO  HOOKS` | `user` | Drop |
| `ao.cli.invoked` | `ao hooks claude-code post-tool-use` | `user` | Drop |
| `ao.app.active` (`channel=cli`) | `ao session get sess-123` | `user` | Drop |
| `ao.cli.invoked` | `ao spawn` | `user` | Keep |
| `ao.app.active` (`channel=renderer`) | n/a | `renderer` | Keep |

### 2. Correct the historical DAU query

The historical query currently removes routine CLI automation only when `actor_type` is null/empty. Remove that actor condition. In plain language, the query should exclude a CLI active event whenever its normalized command path is routine, regardless of whether the client reported `user`, `agent`, `system`, or no actor.

This changes the query from:

```text
Exclude routine CLI activity only when actor_type is empty
```

to:

```text
Exclude routine CLI activity based on normalized command_path
```

This affects dashboard accuracy only; it does not delete stored events or reduce ingestion costs.

### 3. State the deployment boundary clearly

Please add a short warning/checklist to the document or PR description:

- Merging this PR protects clients that receive the next release.
- Updating this Markdown does not modify the live PostHog project.
- The live PostHog transformation must be updated separately with the same actor-independent command rule to stop already-deployed clients.
- Verify the transformation with the cases above, then check volume shortly after enabling it and again after 24 hours.

No additional updater, source-map, dashboard, spawn, or update-prompt work is requested in this PR. Those should remain separate changes so this PR stays focused on telemetry cost-control hardening.

Pulkit7070:
Reviewed in an isolated worktree with the classifier exercised against the real command tree, and the documented queries run against project 475752. Skipping the actor-versus-command ingestion predicate since illegalcall already covered it.

The structural work here is good. `telemetrymeta` collapses the duplicated gate into one source that both callers use, and the prefix match (`normalized == routine || strings.HasPrefix(normalized, routine+" ")`) is better than the plain normalization I suggested, because it catches `ao hooks claude-code post-tool-use`. The trailing space in the prefix guard is the detail that stops `ao project getx` matching `ao project get`. Normalizing the reservoir key was not in the follow-up notes and is a real catch: without it, variant spellings each claim their own daily slot.

Three things below, one of which I would fix before merge.

## 1. The `cliActorType` user allowlist enumerates leaf command names, not command paths

The list contains 17 entries that are not real command paths: `ao add`, `ao merge`, `ao kill`, `ao rename`, `ao claim-pr`, `ao cancel`, `ao submit`, `ao trigger`, `ao restore`, `ao cleanup`, `ao clear`, `ao delete`, `ao remove`, `ao restart`, `ao launch`, `ao resolve-comments`, `ao import-projects`. Those are the leaf names of `ao project add`, `ao pr merge`, `ao session kill`, `ao review cancel` and so on.

The consequence is the inverse: real paths are missing, and since the default is now `system`, they emit nothing at all. I enumerated the command tree and ran `cliActorType("", path)` for every path on this branch. 46 real paths classify as `system`, including all 28 `ao browser *` commands, `ao pr merge`, `ao preview start|stop|status|clear`, `ao review ls|cancel|trigger`, `ao session claim-pr`, `ao agent ls`, `ao dev`, and `ao orchestrator done`.

Most of that is latent, since current CLIs always populate `actor_type` and short-circuit before the allowlist, and `ao browser` postdates actor stamping (#2947, 2026-07-27) so it can never arrive actor-less. But some of it is live. Querying actor-less CLI events in the current billing period for paths the allowlist does not cover:

| command_path | events | installs |
|---|---:|---:|
| `ao agent ls` | 639 | 98 |
| `ao session claim-pr` | 392 | 29 |
| `ao smoke set` | 264 | 1 |
| `ao project orchestration get` | 46 | 3 |
| `ao project orchestration set` | 44 | 3 |
| `ao dev import-projects` | 37 | 11 |
| `ao handoff` | 33 | 5 |
| `ao smoke list` | 30 | 1 |

`ao agent ls` across 98 installs stops producing telemetry and stops contributing CLI-channel DAU. Modest in volume, but it is silent: a `system` classification emits nothing, so nothing surfaces when the list drifts.

Rather than curating the list by hand, consider a test that walks `NewRootCommand(Deps{}).Commands()` recursively and asserts every `CommandPath()` classifies as something other than `system` unless it is in an explicit named exception set. That would have caught the leaf-name mistake and prevents the list rotting as commands are added.

## 2. The updated DAU query in `docs/posthog-cost-controls.md` does not run

Run against project 475752 it fails with:

```
No scope or CTE available
Tables referenced: events
```

The `WITH <expression> AS normalized_command_path` sits before the `SELECT`, so `properties.command_path` has no table scope to resolve against. Moving the expression into a subquery works. This version returns 40 rows against the real project:

```sql
SELECT day, uniqExact(distinct_id) AS active_installs FROM (
  SELECT distinct_id, toDate(timestamp) AS day,
    lower(trim(replaceRegexpAll(toString(properties.command_path), '\s+', ' '))) AS ncp,
    properties.channel AS channel, properties.actor_type AS actor
  FROM events
  WHERE timestamp >= now() - INTERVAL 90 DAY AND event = 'ao.app.active')
WHERE NOT (
  channel = 'cli' AND (actor IS NULL OR actor = '')
  AND (ncp IN ('ao hooks','ao session ls','ao session get','ao orchestrator ls',
               'ao status','ao project ls','ao project get','ao pty-host')
       OR startsWith(ncp, 'ao hooks ') OR startsWith(ncp, 'ao pty-host ')))
GROUP BY day ORDER BY day DESC
```

## 3. The documented normalization does not match the Go implementation

`trim` runs before the whitespace collapse, so whitespace that is not a literal space survives as a trailing space:

```sql
SELECT lower(replaceRegexpAll(trim(BOTH ' ' FROM 'ao  hooks\t'), '\s+', ' ')),  -- 'ao hooks ' (doc)
       lower(trim(replaceRegexpAll('ao  hooks\t', '\s+', ' ')))                 -- 'ao hooks'  (matches Go)
```

`strings.Fields` in `NormalizeCommandPath` produces `ao hooks`, so the two definitions disagree despite the doc claiming "lowercase with whitespace collapsed". It happens to be masked here because every routine path also has a `startsWith(..., '<path> ')` clause that catches the trailing-space form. Worth swapping the order so the SQL and the Go agree, since the doc is the reference for anyone writing the ingestion rule by hand.

## Notes, not blockers

`go test ./internal/cli` fails 6 tests on this branch (`TestBrowserRequiresSessionAndValidWait` plus 5 spawn tests). They fail identically on `origin/main`, with and without `AO_*` environment variables set, so this is pre-existing rather than introduced here. Mentioning it only because the PR body lists that command under validation.

Separately, one actor-less `command_path` in the data reads `Nyrokme-Agency-Service hooks`, alongside earlier events named `moly-engine.onboarding.*` and `to.onboarding.*`. Those look like rebranded forks still shipping the bundled project key, which means any command-based ingestion rule keyed on the literal `ao hooks` will not match their hook traffic. Out of scope for this PR, but relevant to how the cost-control predicate is written.

somewherelostt:
Addressed the review feedback here. The PR now keeps the PostHog cost control rule command-path based, fixes the historical DAU query, uses real CLI command paths in the actor-less classifier, and has tests covering the command tree plus PostHog autocapture staying off.

Validation:
- go test ./internal/cli ./internal/httpd ./internal/telemetrymeta ./internal/adapters/telemetry
- npm --prefix frontend run test -- --run src/renderer/lib/telemetry.test.ts
- npm run frontend:typecheck

CI is green on 7c53632.

Not doing the broader telemetry work in this PR. Follow-up PRs should cover: coarse GeoIP/location only if we want it, toolkit/adapter cleanup, app_version/ao_version distribution, daemon/error dashboard cleanup, and optional rate-limited ao.v2.cli.failed for real internal CLI failures.

somewherelostt:
@Pulkit7070 @illegalcall final review when you get a chance.

I pushed the follow-up fixes from the review command-path based PostHog controls, fixed historical DAU query, real actor-less CLI command classification, command-tree coverage, and the system-command classification hardening.

There is some related telemetry follow-up work, but I’ll keep it out of this PR and do it separately: version distribution, location/GeoIP, toolkit cleanup, daemon/error dashboards, and optional rate-limited CLI failure telemetry.

illegalcall (approved):

PR #3390, opened 2026-07-31
Smooth file viewer transitions and browser maximize handoff

Vaibhaav-Tiwari:
## Summary
- Adds a GSAP FLIP grow/shrink transition for the File diff viewer when maximizing and restoring, plus a CSS grid-row expand/collapse transition for individual file diffs.
- Keeps the Browser panel maximize/restore as an immediate layout swap. Its native `WebContentsView` cannot participate in DOM transforms, so a captured frame now only covers the native-view bounds handoff and crossfades away once the live view is ready.
- Adds reusable `useFlipTransition` and `useMaximizeTransition` hooks for the file surface, with reduced-motion support and design-system motion tokens.
- Prevents the docked and maximized Files views from mounting simultaneously and competing for the same FLIP target.
- Handles browser handoff edge cases: missing captures, rapid toggles, native-view reveal timing, and session switches both during a held transition and while capture is still pending.
- Documents why browser maximize must remain instant until the native-view constraint changes.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run typecheck:e2e`
- [x] `npx vitest run` — 99 files, 1331 tests passed
- [x] Focused transition and panel suite — 7 files, 164 tests passed
- [x] Manual macOS Electron check at `76dfb1d`: browser/files maximize and restore plus rapid toggling behaved cleanly; screenshots are in the PR discussion
- [ ] Recheck the final immediate-swap browser handoff in Electron at the latest head


Vaibhaav-Tiwari:
@codebanditssss can you please look into the browser resizing? file diff works nicely but the browser is stretching out and shrinking components on minimzining and maximizing. its a small feature, focus should be on smoothness.

Hardik180704 (changes requested):
The implementation and automated coverage look thoughtful, but I don't think we should approve this visual/native-view change without the manual Electron check called out in the PR itself. The latest discussion still mentions browser stretching/shrinking behavior, while the current GSAP revision has not been manually verified in the running app.

Please verify browser maximize/restore, files maximize/restore, rapid toggling, session switching mid-transition, and reduced-motion behavior on macOS, then attach the result or screenshots. I can approve once that visual check is confirmed.

codebanditssss:
ran this locally in the desktop app. the file diff side works nicely, but the browser panel was stretching the page and squeezing its toolbar controls on both maximize and restore. pushed two commits to the branch.

### browser resize (59b5d019f)

root cause is that `.browser-panel` is the Flip target and `scale: false` tweens its real width/height, so every frame is a genuine relayout of the panel's contents. the toolbar is a dense single row, 8 icon buttons plus a `flex-1` URL field, and at docked width those controls are already near their floor, so tweening through that range visibly squeezes and re-spaces them. the snapshot compounds it: at `object-fit: contain` its 100%/100% box changes aspect every frame, so the captured page rescales continuously and grows letterbox bands against `--bg`.

the files view escapes this because it's a vertical list, so reflowing its width just re-wraps text. that's why only one of the two surfaces looks wrong.

rather than try to animate the chrome nicely, the panel now sits the tween out. `useFlipTransition` stamps `data-flipping` on the target for exactly the tween's duration, and the panel's CSS uses that to hide the toolbar and let the viewport fill the whole box, so the growing or shrinking window reveals and clips one static image and nothing reshapes. toolbar fades back in as it lands. snapshot moves to `object-fit: cover` for the popout case, which is full-bleed and uniformly scaled, cropping instead of letterboxing, so it reads as a zoom. tab-switch keeps `contain` since its box shape doesn't change.

the mark is cleared on complete, on interrupt, and on unmount. worth calling out that `kill()` never fires `onComplete`, so without the interrupt path an abandoned node would stay marked and the panel would be permanently chrome-less.

i also corrected the comment above the hook. it said real layout tweening means "nothing but the intended box ever changes shape", which is the assumption the approach was built on and is exactly what breaks here.

### two review items (76dfb1da5)

**the description says this adds `--duration-emphasized`/`--ease-emphasized` tokens, but those don't exist.** `grep -rn emphasized` over the repo returns nothing. the values are hardcoded in `useFlipTransition.ts` as `DEFAULT_DURATION_MS = 320` and `DEFAULT_EASE = "expo.out"`, and DESIGN.md documents them as prose. i added the tokens for real in `tokens.css` and the hook now reads the duration from the document root, with the documented default as fallback. the ease stays a GSAP ease since that's what drives the tween, and `--ease-emphasized` is its cubic-bezier equivalent for the CSS-driven parts.

**the files view mounts twice while maximized.** `SessionView` passes the same `filesMaximize.setNodeRef` to both the rail's `SessionFilesView` and the full-window overlay, and `handleToggleFilesPopOut` forces the inspector to the files tab, so both are mounted at once and both claim that one ref. whichever attaches last silently wins. on restore the overlay unmounts and hands back `null` while the docked one stays mounted, so the animation falls through to an immediate settle. it only works today because `SessionFilesView` uses an inline arrow for `ref`, which React re-attaches every render. that's accidental rather than designed.

`BrowserView` already guards against exactly this (`SessionInspector.tsx`, "must not mount a second BrowserPanelView"). the files tab now does the same and offers a Return to panel action.

### what's good

the `releasing` crossfade is a real race fix and the reasoning holds up, `measureAndSend()`'s IPC round trip and the React state update do land on different ticks. clearing `popoutTransitionRef` on session switch is a genuine catch and it's tested. `inert` plus the grid-row collapse is the right way to keep the diff row mounted. the comments explain why rather than restating the code, which is rarer than it should be.

### verification

`tsc --noEmit` clean, 897/897 renderer tests pass. added tests for the `data-flipping` lifecycle, the duration token, and the files guard, and mutation-tested four of them: breaking the interrupt cleanup, the settle cleanup, the token read, and the files guard each fail a test.

the unchecked manual-verification box in the test plan was the right call to flag. this is what a click-through catches. worth another pass on your side once you pull, particularly whether the toolbar drop-out reads as too abrupt, since the fade-in duration is the knob there.

Hardik180704:
Verified the latest head (`76dfb1d`) in the real Electron app on macOS.

- Browser panel: docked → maximized → restored cleanly; the page keeps its normal layout and I did not see stretching, clipping, or a stale snapshot.
- Rapid maximize/restore toggling settles back into a usable docked state.
- Files panel: maximize/restore also completed cleanly.
- `npm run typecheck` passed.
- Focused transition/panel suite passed: 6 files, 116 tests. This includes the reduced-motion and transition cleanup coverage.

Docked browser:
PR 3390 browser docked

Maximized browser:
PR 3390 browser maximized

Hardik180704 (approved):
Verified the latest head locally in the real Electron app on macOS. Browser and Files maximize/restore behave cleanly, rapid toggling settles correctly, typecheck passes, and the focused suite is green (116 tests). I posted the docked/maximized screenshots in the PR discussion.

codebanditssss:
follow-up, and a correction to my earlier comment: **the browser panel no longer animates its maximize at all.** the approach I described above (`data-flipping`, hiding the toolbar, `object-fit: cover`) is gone. if you read that comment, skip its browser-transition section, the review findings below it still stand.

### why it was removed

I kept fixing artifacts one at a time and each fix exposed the next one:

1. the toolbar squeezed through the tween, since `scale: false` moves real width/height and the toolbar is a dense row already near its floor at docked width
2. fixed that by hiding it, then the page visibly zoomed in and snapped back at the handoff, because the frozen frame was being fitted to a box that resizes for the whole transition while the live view always renders at 1x
3. fixed that by pinning the frame at 1x, and then the panel just grew into empty space, which looked worse than either

underneath all three, the tween relayouts the whole panel subtree every frame for 320ms, which is what actually read as lag. none of my fixes touched that.

the real constraint is structural: browser content is a native `WebContentsView` that Chromium composites above the page. it cannot be transformed, clipped or tweened. so any grow/shrink is a captured bitmap standing in for the page, and a bitmap can't agree with a live view that relayouts at the new size. scale it and it zooms; pin it and you get blank space. there's no version of this that's both smooth and honest without solving the native-view problem first.

so maximize is now an immediate swap. the frozen frame stays, but only for what it's genuinely good at, covering the native view's handoff, which is an IPC round trip that would otherwise flash, and it crossfades out once the live view is up at the new bounds. nothing tweens layout, so there's nothing to lag.

net 280 lines deleted, 76 added. removed everything that only existed for the animation: `panelRef`, `data-flip-id`, the frame pinning, and the `data-flipping` mark. **the file diff viewer keeps its FLIP grow untouched**, it works well there because a vertical list of file cards reflows cleanly under a width sweep. that difference is the whole story of why one of these surfaces looked fine and the other didn't.

DESIGN.md now records the reverted attempt, each variant and how it failed, so this doesn't get retried blind.

### also fixed, still standing from the earlier comment

- **`--duration-emphasized`/`--ease-emphasized` didn't exist** despite the description claiming them. added for real in `tokens.css`; the flip hook reads the duration from the document root instead of holding its own copy.
- **the files view mounted twice while maximized**, rail copy plus overlay, both handed the same maximize-transition node ref. the rail now stands down while maximized with a Return to panel action, matching what `BrowserView` already did.

### verification

`tsc --noEmit` clean, 894/894 renderer tests. mutation-tested the tests I added: the files guard, the duration token read, and the capture-reporting branch each fail a test when the fix is reverted. one earlier mutation of mine passed misleadingly because `SessionView.test.tsx` mocks `useBrowserView` wholesale, so I moved that coverage into `useBrowserView.test.tsx` where it actually exercises the branch.

worth saying plainly: the first three of my commits here were the wrong approach and the last one removes them. happy to squash if you'd rather the branch not carry that history, or to revert my browser commits entirely if you'd prefer to take a different run at it yourself.

illegalcall:
Follow-up pushed in `6d7181942`.

While validating the final immediate-swap implementation, I found a separate handoff race: if the snapshot capture was still pending during a session switch, its completion could apply the old popout request to the incoming session. `SessionView` now invalidates pending requests and settle state across session changes, with a regression test for the pending-capture case. I also removed stale comments that still described the Browser panel as a FLIP target.

Verification:
- `npm run typecheck`
- `npm run typecheck:e2e`
- focused transition/panel suite: 164/164
- full renderer suite: 1331/1331
- GitHub frontend test, renderer smoke, and gitleaks: all green

I refreshed the PR title/body to describe the final architecture accurately. The latest immediate-swap head still deserves the manual Electron click-through called out in the test plan.

illegalcall:
Rebased onto `main` at `42acd419e` and force-pushed the rewritten head as `a8e6bd911`.

The conflicts were in `SessionView` and `SessionView.test.tsx`. The resolution preserves `main`'s `useLayoutEffect` session reset and browser-slot visibility regression coverage alongside this PR's snapshot handoff, maximize tests, and stale popout-request cancellation.

Verification after the rebase:
- focused transition/panel suite: 168/168
- full renderer suite: 1343/1343
- `npm run typecheck`
- `npm run typecheck:e2e`
- GitHub frontend test, renderer smoke, and gitleaks: all green

illegalcall:
### Manual verification — macOS Electron (`a8e6bd911`)

Verified:

- [x] `⌘⇧B` toggles the right inspector sidebar.
- [x] The Files view opens and displays changed files.
- [x] Files maximize grows smoothly without warped or duplicated content.
- [x] Files minimize returns to the inspector layout.
- [x] Individual file diffs expand and collapse smoothly.
- [x] Rapid Files maximize/minimize toggling always settles into a usable state.
- [x] Browser navigation loads the requested URL.
- [x] Browser maximize switches immediately without stretching, toolbar squeezing, a blank frame, or page zoom; the snapshot handoff is acceptable.

Issues still identified:

- [ ] The Files view shows an unwanted scrollbar/scroll behavior. Dont need in file scrolling, overall scroll is enough
- [ ] In the maximized Browser view, the macOS traffic lights cover the Back button, so Back cannot be used.
image

- [ ] The Files toolbar does not need the Refresh and Close (`X`) buttons.
image


- [ ] Restoring Files from maximized/fullscreen to the normal inspector layout feels slow; investigate the minimize path.
- [ ] The Browser action is maximize/minimize, not a pop-out, but the control is labeled **Pop out / Return to panel**. Rename it to **Maximize browser / Minimize browser**.

illegalcall (changes requested):
Added comemnt

illegalcall:
Addressed the requested review feedback in `a40c95db5`:

- Removed the nested per-file diff scroller and height cap; the Files surface now owns scrolling.
- Removed the Files **Refresh** and **Close (X)** toolbar actions.
- Added macOS traffic-light clearance to the maximized Browser toolbar so **Back** remains usable.
- Renamed the Browser action to **Maximize browser / Minimize browser** (including the inspector fallback action).
- Shortened Files restore to the existing `--duration-normal` timing while keeping maximize on `--duration-emphasized`.

Verification:
- `npm test`: 100 files, 1,348 tests passed
- Focused regression suite: 65 tests passed
- `npm run typecheck`
- `npm run typecheck:e2e`

Vaibhaav-Tiwari:
> Addressed the requested review feedback in `a40c95db5`:
> 
> * Removed the nested per-file diff scroller and height cap; the Files surface now owns scrolling.
> * Removed the Files **Refresh** and **Close (X)** toolbar actions.
> * Added macOS traffic-light clearance to the maximized Browser toolbar so **Back** remains usable.
> * Renamed the Browser action to **Maximize browser / Minimize browser** (including the inspector fallback action).
> * Shortened Files restore to the existing `--duration-normal` timing while keeping maximize on `--duration-emphasized`.
> 
> Verification:
> 
> * `npm test`: 100 files, 1,348 tests passed
> * Focused regression suite: 65 tests passed
> * `npm run typecheck`
> * `npm run typecheck:e2e`

can you please tell me the status and state of this pr?

PR #3391, opened 2026-07-31
fix(desktop): coalesce steady-state PTY writes per animation frame to…

### LINKED ISSUE #3309: bug(desktop): terminal panel shows smudged/garbled output and cursor jumps erratically during agent activity
## Bug

The built-in terminal panel intermittently renders **smudged/garbled output** and the **cursor visibly jumps around very fast** while an agent session is active. It's difficult to capture in a static screenshot, but the effect is a flickering, unstable terminal surface during heavy agent output.

## Reproduction

1. Start an agent session (worker or orchestrator) from the desktop app
2. While the agent is actively producing output (reading files, searching, etc.), watch the terminal panel
3. Observe intermittent smudged/garbled text artifacts and the cursor rapidly jumping between positions

## Expected behaviour

Terminal output should render cleanly with the cursor moving smoothly as new lines are written — no visual artifacts or erratic cursor movement.

## Actual behaviour

- **Smudged/garbled output**: text appears momentarily distorted or overlapping
- **Erratic cursor jumping**: the visible cursor blinks/teleports rapidly across the terminal surface during active output

## Environment

- AO desktop app (xterm.js-based terminal panel)
- OS: _(to confirm)_
- AO version: _(to confirm)_

## Screenshots / recordings

**Screenshot** (static — doesn't fully capture the flickering nature):

terminal showing active agent session

**Screen recording**: _(to be attached — see comment below)_

## Suspected area

The terminal panel uses `@xterm/xterm` (v6). Likely candidates:
- **xterm canvas/DOM renderer** struggling to keep up with burst writes from the PTY mux
- **ANSI escape sequence handling** (cursor positioning codes causing visible reflow artifacts mid-render)
- **Debouncing / batch-flush** of terminal writes — if each PTY chunk triggers an immediate xterm render, rapid output can cause the renderer to tear

Related existing issues: #3160 (terminal auto-scroll) — may share root cause in how the terminal surface handles rapid updates.

anmol0b:
## What
Coalesces PTY output writes into `xterm` so bursts of live agent output are painted as a single frame instead of one repaint per WebSocket chunk.

## Why
Fixes #3309. The terminal panel showed smudged/garbled output and an erratically jumping cursor during active agent sessions. The initial-replay gate (#3160) already solves this exact frame-boundary tearing for the first burst after attach — but once that gate closes, every subsequent PTY chunk was written straight through with `terminal.write(bytes)`, one call per WebSocket `data` frame. The browser is free to paint between those writes, so a burst of output (common while an agent is actively reading files, searching, etc.) painted several distinct intermediate states instead of one clean update.

## How
Added a second, lighter coalescing layer in `useTerminalSession.ts` alongside the existing replay buffer:
- `steadyWriteChunks` / `flushSteadyWrite` batch bytes arriving within the same animation frame into a single `term.write()` call, scheduled via `requestAnimationFrame` rather than gated on a quiet window — a live pane shouldn't wait for a lull the way the initial replay can.
- Bounded by `STEADY_WRITE_MAX_BYTES` (256KB) so a never-idle stream still flushes regularly instead of growing the buffer unbounded.
- All teardown/exit/error/reconnect paths (`onExit`, `onError`, `onConnectionChange` closed branch, `teardownMux`) now flush the steady buffer first, so exit/error banner lines never land ahead of output that was still buffered.

Scope is intentionally limited to `useTerminalSession.ts` — `XtermTerminal.tsx` and the mux/backend layers were not touched, since the tearing is a renderer-side batching problem, not a protocol or PTY issue.

## Testing
- `npm --prefix frontend run typecheck` — clean
- `npm --prefix frontend test` — 97 files / 1302 tests passed (including `useTerminalSession.test.tsx`, updated for the new frame-scheduled write timing)
- `npm --prefix frontend run typecheck:e2e` — clean
- `npm --prefix frontend run test:e2e:renderer` — 20/20 `@T0`/`@P0` passed, including `smoke-terminal.spec.ts`

## Checklist
- [x] Branched from `main`
- [x] One focused change; closes #3309
- [x] Relevant CI checks pass for the area touched (frontend)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -563,8 +615,16 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
 					drainPostReplayWrites();
 					return;
 				}
-				terminal.write(bytes);
 				emitOutput(bytes);
```
whoisasx: [P2] `onOutput` now runs before the deferred `terminal.write()`. This changes the documented post-write behavior of this hook and differs from the previous implementation, which wrote the bytes before invoking the observer. Consumers such as the URL watcher may currently tolerate this, but the public hook contract has changed. Please invoke the observer after the terminal write and add a regression test that verifies the ordering.

whoisasx (changes requested):
## Review summary

Changes requested. The steady-state batching approach is reasonable, but the output observer ordering regresses the documented hook contract.

Please address the inline finding and add a regression test covering the ordering between onOutput and terminal.write. The merge conflict is intentionally excluded from this review as requested.

frontend/src/renderer/hooks/useTerminalSession.ts
anmol0b: Thanks for the catch. This is resolved by the merge with `main`: #3359 removed the `onOutput` observer (along with `outputDecoder`/`emitOutput`/`detect-urls.ts`) entirely, so the steady-state path no longer has any observer callback — bytes are batched into `steadyWriteChunks` and flushed to `terminal.write` per animation frame (or synchronously past the byte cap), with nothing running ahead of the write. The onOutput-vs-write ordering contract no longer exists to regress.

I also added a dedicated regression test locking in the new contract: "coalesces a steady-state burst into one ordered write per frame" asserts a same-frame burst is not written synchronously and lands as a single ordered write on the next animation frame. It joins the existing ordering coverage ("preserves wire order when output arrives between bounded replay writes", "streams live output straight through once the gate has closed", "flushes early once the buffered burst passes the byte cap").

Resolving this thread — happy to revisit if you see anything else.

whoisasx (approved):
Approved. The latest main merge removes the obsolete output observer concern, and the added steady-state coalescing test now covers same-frame ordered writes directly. The focused hook suite and frontend typecheck pass, and the branch merges cleanly with current main. Looks good to merge from my side.

i-trytoohard:
@anmol0b Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@anmol0b
Closing this PR as stale after 30+ days open. It was approved but never merged, and the linked [#3309](https://github.com/Untrivial-ai/agent-orchestrator/issues/3309) remains open.

Issue #3378, opened 2026-07-30
`ao session ls` hides orchestrator-kind sessions → operators spawn duplicate orchestrators unknowingly


zenprocess:
## Summary
The session listing (`ao session ls`) omits sessions of kind `orchestrator`. Because the default listing is the primary way an operator checks "what's already running," an orchestrator that already exists is invisible there — so the operator (or an automation) spawns a second orchestrator for the same project. Multiple orchestrators then run against the same worktree/board, competing: duplicate dispatches, conflicting merges, and one orchestrator reaping the other's workers.

## Impact
- Silent duplicate orchestrators per project; wasted spend and contradictory actions.
- Hard to diagnose: the very command used to check state is the one that hides the cause.
- A dead/duplicate orchestrator surfaces later as a confusing "No conversation found with session ID …" when someone clicks through to it.

## Steps to reproduce
1. Start an orchestrator for a project.
2. Run `ao session ls` — the orchestrator is not listed.
3. Lacking any signal that one exists, start another orchestrator for the same project.
4. Observe two orchestrators dispatching/merging against the same target.

## Root cause
The listing filters out the `orchestrator` session kind (only `ao orchestrator ls` shows them). The two surfaces are separate, so the common path gives a false "nothing is running."

## Proposed fix (any one, ideally 1+3)
1. **Include orchestrators in `ao session ls`** with a clear `KIND` column (or a visible `— orchestrators hidden, see \`ao orchestrator ls\`` footer when any exist).
2. **Guard against duplicates at spawn:** when creating an orchestrator for a project that already has a live one, refuse (or require `--force`) with a pointer to the existing one.
3. **Reconcile on spawn:** detect an existing live orchestrator for the same project/worktree and attach/adopt rather than create a second.

## Acceptance
- With a live orchestrator, the default listing makes its existence discoverable (fix 1), **and/or**
- spawning a second orchestrator for the same project is blocked or explicitly warned (fix 2/3),
- covered by a test asserting a second same-project orchestrator spawn is prevented or surfaced.

zenprocess:
Adding detail to this report rather than restating it. Two concrete points that fell out of an investigation of the same failure mode on a fleet install. Take them as supporting evidence for fix points 2/3 in the original report.

**Point A: `ao spawn --kind orchestrator` does not route to the live-orchestrator-aware endpoint.**

[...]

PR #3393, opened 2026-07-31
fix(cli): surface hidden orchestrator sessions

Hardik180704:
Addresses the session-list visibility portion of #3378.

The generic orchestrator spawn path remains tracked in #3378 and will be handled separately.

## What changed

- show a hint when `ao session ls` hides active orchestrator sessions
- point users to `--all` or `ao orchestrator ls`
- expose `hiddenOrchestratorCount` in JSON metadata
- keep the existing `--all` behavior unchanged

## Why

The default session list intentionally filters orchestrators, but it did so silently. That could make the output look like no relevant session was running.

## Checks

- `go test ./internal/cli`
- `go vet ./internal/cli`
- full backend `go test ./...`
- two unrelated auth-probe timeout tests were rerun individually and passed

### DIFF
--- a/backend/internal/cli/session.go
+++ b/backend/internal/cli/session.go
@@ -401,11 +401,12 @@ func (c *commandContext) listSessions(ctx context.Context, cmd *cobra.Command, o
 	sessions := filterAndSortSessions(res.Sessions, opts.all)
 	hiddenTerminatedCount := 0
 	if !opts.includeTerminated {
-		count, err := c.countHiddenTerminated(ctx, opts.project, opts.all)
+		terminatedCount, terminatedOrchestratorCount, err := c.countHiddenTerminated(ctx, opts.project, opts.all)
 		if err != nil {
 			return err
 		}
-		hiddenTerminatedCount = count
+		hiddenTerminatedCount = terminatedCount
+		hiddenOrchestratorCount += terminatedOrchestratorCount
 	}
 	if opts.json {
 		out := sessionListOutput{Data: sessionListEntries(sessions)}
@@ -416,7 +417,7 @@ func (c *commandContext) listSessions(ctx context.Context, cmd *cobra.Command, o
 	return writeSessionList(cmd, sessions, hiddenTerminatedCount, hiddenOrchestratorCount)
 }
 
-func (c *commandContext) countHiddenTerminated(ctx context.Context, project string, includeOrchestrators bool) (int, error) {
+func (c *commandContext) countHiddenTerminated(ctx context.Context, project string, includeOrchestrators bool) (int, int, error) {
 	params := url.Values{}
 	if project != "" {
 		params.Set("project", project)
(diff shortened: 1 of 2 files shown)


zenprocess:
Review on the diff itself (not running it — I can't swap a binary on the install I'm reporting from).

Diff is sound for the visibility half. The `hiddenOrchestratorCount` JSON field plus the `Use --all or `ao orchestrator ls`` hint is a clean small change and the test coverage is right. The work here is good.

Two review points, only because this PR carries the `Closes #3378` keyword and the loop in 3378 is wider than the visibility half:

**1. The hint closes a hole for a human reading a terminal; it does not close the loop.** The duplicates in the wild were not created by humans looking at `ao session ls` — at least not the ones we reported in the 3378 thread. They came from programmatic callers hitting the generic `POST /api/v1/sessions` path that has no live-orchestrator check. On current `main`:

- Generic spawn route: `backend/internal/service/session/service.go:193-215` (`Service.Spawn`) goes straight to the manager; no `List(..., OrchestratorOnly: true, Active: &true)` guard before the insert, and no call to `lockOrchestratorProject`.
- Existing dedup lives in `Service.SpawnOrchestrator` at `service.go:330-364` — runs in the `else` branch after `lockOrchestratorProject(projectID)`, queries active orchestrators, and returns the existing one when `len(existing) > 0`.
- That path is only reachable via `POST /api/v1/orchestrators` (`backend/internal/httpd/controllers/sessions.go:144`).
- `ao spawn --kind orchestrator` POSTs to `/api/v1/sessions` instead (`backend/internal/cli/spawn.go:131`, file-level comment at `:37`: *"POST /api/v1/sessions. The CLI keeps its own copy so it need not import httpd."*).

So after this PR merges, a programmatic caller is still silently producing duplicates; only the human-in-a-terminal path gets the hint. The accumulation loop 3378 describes is therefore not actually closed.

**2. Suggesting either narrowing the closing keyword, or filing the spawn-side half as a follow-up.**

- Option A: change `Closes #3378` to a partial reference (e.g. `"Addresses the visibility half of #3378"`), leave the issue open, and let a follow-up PR (or this one's follow-up) carry the spawn-side guard.
- Option B: keep the closing keyword, file a sibling issue immediately for the spawn-side half with the same `Service.SpawnOrchestrator`-current-shape evidence so it does not get lost.

Whichever you pick, the spawn-side guard really does need a separate PR (or a follow-up commit in this one) to actually close 3378 — the hint alone won't.

**3. One property that any spawn-side guard must preserve**, since this PR is the moment that surface is most likely to get designed: a **terminated** orchestrator must not block a replacement. Current `SpawnOrchestrator` keys on `Active: &true`, which maps to `IsTerminated == false` (`Service.matchesSessionFilter` at `service.go:624-626`). Preserve that. A guard keyed on existence rather than liveness makes replacing a dead orchestrator impossible, which is the most damaging and easiest-to-write-by-accident failure mode in this matrix.

Net: ship this as-is for what it is, and don't let `Closes` absorb a defect this PR does not touch. Their call on A vs B.

Hardik180704:
Thanks for tracing the generic spawn path—you’re right. I verified `SpawnOrchestrator` but missed that `ao spawn --kind orchestrator` uses the generic sessions endpoint, so this PR only fixes the visibility half.

I’ve updated the closing reference to keep #3378 open. I’ll handle the generic spawn-side guard separately, including coverage that active orchestrators are reused or rejected while terminated orchestrators do not block replacement.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

JAYATIAHUJA (commented):
hiddenOrchestratorCount is computed only from the primary sessions response, which is ctive=true by default. That means a terminated orchestrator is still hidden silently: o session ls asks for active sessions, sees no orchestrator to count, then countHiddenTerminated(..., opts.all=false) filters the terminated orchestrator back out of the terminated count as well. In that state the user gets neither the new orchestrator hint nor the terminated hint, even though o orchestrator ls would surface the session. Please include terminated orchestrators in the hidden-orchestrator accounting, or otherwise ensure this default-hidden case gets a hint too.

JAYATIAHUJA (approved):
LGTM. Verified the follow-up commit 6529b9d5 resolves the earlier review comment: countHiddenTerminated now returns the terminated-orchestrator count separately, so a terminated-only orchestrator gets the hidden-orchestrator hint instead of being silently dropped from both counters — and TestSessionList_HintsWhenOnlyTerminatedOrchestratorIsHidden covers exactly that case. Active and terminated responses are disjoint (active=true vs active=false), so no double-counting; --all correctly suppresses the hint. go test ./internal/cli, go vet, and go build all pass locally on the PR head; all CI checks are green.

PR #3394, opened 2026-07-31
revert: remove automated orchestrator re-engagement messages

harshitsinghbhandari:
## What changed

Removes the automated orchestrator re-engagement feature in full by reverting commit 79a70e82f (PR #3274). Orchestrator sessions were receiving the daemon-authored '[AO automated re-engagement]' message repeatedly, 3 to 4 duplicates in a row, and the decision is to drop the feature entirely rather than tune it.

## Provenance

The message was introduced by @Hardik180704 (Hardik Sharma) in PR #3274, 'feat: add durable idle-orchestrator re-engagement', commit 79a70e82f, merged 2026-07-29, closing #3245. Per that PR's description it was meant as a bounded recovery path (exponential backoff, three-attempt ceiling) for orchestrators that go idle while assigned work remains, effectively succeeding the worker-idle nudge that was removed in #3040 (issue #3038). In practice it fires repeatedly at orchestrator sessions.

## Removed

- the orchestratorloop manager (timer loop, idle detection, message send) and its tests
- lifecycle and daemon wiring for the re-engagement manager
- the sessions /complete endpoint, its DTO, controller, and OpenAPI surface
- the ao orchestrator done CLI command: it existed solely to durably suppress re-engagement and had no other consumers (verified by grep across backend, frontend, and docs), so it is removed too
- the orchestrator_reengagement store, sqlc queries, and generated code; the schema is removed by a new migration 0039_drop_orchestrator_reengagement (merged migration 0038 stays untouched per AGENTS.md, mirroring the 0037_drop_worker_idle_outbox precedent)
- the generated frontend API schema entries

No disabled code paths remain: grep for 'reengag', 'orchestratorloop', 're-engagement', 'orchestrator done', and 'CompleteOrchestrator' returns nothing. The only remaining automated nudges are the pre-existing PR-observation nudges in internal/lifecycle/reactions.go, which target worker sessions that own PRs, never the orchestrator.

Refs #3038, #3245.

## Verification

- go build ./... and go vet ./...: pass
- go test -race for all changed packages: pass
- full backend go test ./...: pass except 6 pre-existing environmental failures in internal/cli (spawn/browser tests that hit the live local daemon); the same 6 fail identically on a clean upstream/main checkout
- golangci-lint v2.12.2 with a clean cache: 0 issues
- npm run sqlc and npm run api: zero drift against the reverted generated files
- frontend typecheck and typecheck:e2e: pass
- frontend vitest: 97 files, 1302 tests, all pass


Hardik180704:
Thanks for catching the repeated-message behavior. I’m happy to help investigate the root cause or improve the deduplication logic if the team prefers fixing it over fully removing the feature.

One thing I noticed: this also deletes migration `0038`, which has already been merged into `main`. The repository guidelines say merged migrations should remain immutable and schema removal should use a new migration instead. Could we keep `0038` and add a follow-up migration if those tables need to be removed?

Also, can we confirm whether the intended direction is full removal rather than fixing the duplicate delivery behavior?

harshitsinghbhandari:
Yes, the intended is full removal. because not always an orchestrator is working. i might just want to work on other project and these messages contaminate the context.

zenprocess:
Read the diff; not running it. **Not arguing against the revert** — your call whether the feature stays, and I'm not asking you to reverse it. Offering one observation because we are unusually well placed to see it and nobody else seems to have raised it.

**Hypothesis, clearly unproven:** if a project has accumulated multiple orchestrator sessions — the defect in #3378, which we measured at six orchestrators in one project inside roughly 24 hours — then each idle duplicate is independently eligible for re-engagement. On that reading the timer is not misfiring; it is correctly firing at orchestrators that should never have existed. One project would emit N nearly-simultaneous messages, which would present exactly as "3 to 4 duplicates in a row". The two bugs compound: dup-orchestrators → multi-target re-engagement → looks like a buggy timer.

**Cannot prove this. Have not reproduced it against your build.** It is offered because it is cheap to check and expensive to miss. If the hypothesis is wrong, the revert stands on its own merits and this comment is irrelevant. If it is right, both fixes are needed in the same window — the spawn-side guard on the generic `POST /api/v1/sessions` path (see #3378) plus the revert — and #3038/#3245 do not reopen.

**Cheap check.** Whether the duplicate re-engagement messages observed carried **distinct session ids for the same project**, or were the same session id repeating. If they were distinct ids, that supports the compounding reading. If they were a single id, the hypothesis is wrong, the re-engagement loop itself was the problem, and the revert is the right answer regardless of #3378.

**One scope note, not a complaint.** I verified the revert against the diff and it does take with it `ao orchestrator done` (`backend/internal/cli/orchestrator.go`, 27 deletions) and the sessions `/complete` endpoint (`backend/internal/httpd/controllers/sessions.go` 15 deletions; OpenAPI surface 55 deletions; DTO/controller layer). The PR body notes this; flagging here in case anyone downstream had been planning to land a sibling PR against `ao orchestrator done` or `/complete` — that surface goes with the re-engagement feature.

Not asking for a reversal. Asking only that the duplicate-id check is cheap enough to be worth doing before this lands, in case it changes whether you want a paired fix in the same release.

harshitsinghbhandari:
Good catch on the migration, thank you. You are right that AGENTS.md treats already-merged migrations as immutable regardless of release status. Fixed in f95571bcc: 0038 is restored byte-for-byte and the tables are now dropped by a new 0039_drop_orchestrator_reengagement migration, mirroring the 0037_drop_worker_idle_outbox precedent. sqlc shows no drift and the storage test suite passes through 0039.

On direction: full removal is the maintainer's explicit call. The repeated delivery was the trigger, but the decision was to drop the automated re-engagement path entirely rather than tune its deduplication, so this PR intentionally removes the feature in full (including ao orchestrator done, which had no consumers outside this feature). If the team later wants a recovery mechanism for idle orchestrators, that can be designed fresh with this incident as input.

Hardik180704:
Got it, thanks for confirming. Full removal makes sense if unsolicited re-engagement can contaminate an orchestrator’s context while the user is working elsewhere.

My only remaining concern is migration `0038`, since it has already landed on `main`. Could we keep the existing migration immutable and use a new migration for any schema cleanup?

harshitsinghbhandari:
I have updated the migration to 0039 instead of changing 0038

Hardik180704:
Thanks for addressing the migration concern and updating it to use a new migration.

Hardik180704 (approved):
Reviewed the latest head, including the new `0039` migration. The removal is internally consistent, `0038` remains byte-for-byte unchanged, generated API artifacts match, and the affected backend tests and vet checks pass locally.

One non-blocking note: the PR description says the reverted feature was unreleased, but commit `79a70e82f` is included in `v0.11.2` and `v0.11.2-nightly.202607301613`. Please update that release-context statement so the removal of the briefly released CLI/API surface is documented accurately.

zenprocess:
Confirming the revert was necessary, not catching anything out — but one premise in the PR body is wrong and worth fixing before this is read by others.

Per the compare API: `compare/v0.11.2...79a70e82f` reports `status: behind, ahead_by: 0, behind_by: 25, merge_base_commit: 79a70e82fa455824568e25b9530a8e2241314cec` (which is the commit itself), so `79a70e82f` is an ancestor of `v0.11.2` and that release DOES contain PR #3274. The v0.11.2 release notes list `feat: add durable idle-orchestrator re-engagement (#3274)` by name. That means anyone who upgraded to v0.11.2 already has migration 0038 applied, which makes restoring 0038 byte-for-byte plus dropping the tables via a new 0039 the correct handling — not merely a documentation-rule formality.

The code change in this PR already lands that handling correctly. Only the "unreleased; no tag contains 79a70e82f" sentence in the PR body needs amending — the implementation is right and worth shipping.

PR #3395, opened 2026-07-31
feat(landing): generate Markdown twins for docs

VenkataSakethDakuri:
## Summary

- mark rendered documentation titles, descriptions, and article bodies with stable extraction attributes
- generate an `index.html.md` twin for every exported documentation page during `postbuild`
- extract only documentation content with Cheerio and convert it with `node-html-markdown`
- fail the build when no documentation HTML is available or required content is missing

## Why

The documentation is authored as MDX and statically rendered to HTML. Generating Markdown twins from that final HTML reuses the existing React/MDX component rendering while excluding navigation and other page chrome.

## Validation

- `npm ci --dry-run --ignore-scripts`
- `npm run build`
- production export generated 55 documentation HTML pages and 55 Markdown twins, with zero missing twins
- `git diff --check`

### DIFF
--- a/frontend/src/landing/scripts/generate-markdown-twins.mjs
+++ b/frontend/src/landing/scripts/generate-markdown-twins.mjs
@@ -21,7 +21,16 @@ for (const htmlFile of htmlFiles) {
   const $ = load(await readFile(htmlFile, "utf8"));
   const title = $("[data-doc-title]").first().text().trim();
   const description = $("[data-doc-description]").first().text().trim();
-  const article = $("[data-doc-content]").first().html();
+  const content = $("[data-doc-content]").first();
+
+  content.find("[data-doc-tab-list]").remove();
+  content.find("[data-doc-tab-panel]").each((_, panel) => {
+    const label = $(panel).attr("data-doc-tab-label");
+    if (label) $(panel).prepend($("<p>").append($("<strong>").text(label)));
+    $(panel).removeAttr("hidden");
+  });
+
+  const article = content.html();
 
   if (!title || !description || !article) throw new Error(`Missing documentation content in ${htmlFile}`);
 
--- a/frontend/src/landing/scripts/generate-markdown-twins.mjs
+++ b/frontend/src/landing/scripts/generate-markdown-twins.mjs
@@ -1,0 +1,22 @@
+import { glob, readFile, writeFile } from "node:fs/promises";
+import { load } from "cheerio";
+import { NodeHtmlMarkdown } from "node-html-markdown";
+
+const htmlFiles = [];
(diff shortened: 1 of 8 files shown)


frontend/src/landing/scripts/generate-markdown-twins.mjs
```diff
@@ -0,0 +1,22 @@
...
+const htmlFiles = [];
+for await (const file of glob("out/docs/**/index.html")) htmlFiles.push(file);
+
+if (htmlFiles.length === 0) throw new Error("No documentation HTML files found");
+
+for (const htmlFile of htmlFiles) {
+  const $ = load(await readFile(htmlFile, "utf8"));
+  const title = $("[data-doc-title]").first().text().trim();
+  const description = $("[data-doc-description]").first().text().trim();
+  const article = $("[data-doc-content]").first().html();
```
illegalcall: This extracts only the initially rendered article DOM, but `Tabs` renders only `tabs[active]`. As a result the generated twin is incomplete: `installation/index.html.md` keeps only the Claude Code and npm panels while dropping Codex, Cursor, Aider, OpenCode, pnpm, yarn, and source-install instructions; `quickstart/index.html.md` drops the freeform-prompt workflow. Please expose and label every tab panel in the extraction markup (or otherwise flatten all panels) before converting the article.

frontend/src/landing/scripts/generate-markdown-twins.mjs
```diff
@@ -0,0 +1,22 @@
+import { glob, readFile, writeFile } from "node:fs/promises";
```
illegalcall: `fs.promises.glob` is only available starting in Node 22, while this package has no Node >=22 engine and its Next.js version supports Node >=20.9. I reproduced `npm run postbuild` failing on Node 20.19.5 with “does not provide an export named glob”; Node 22 succeeds. Please use a Node 20-compatible traversal, or explicitly raise and enforce the package runtime requirement if that is intended.

illegalcall (changes requested):
Requesting changes for two functional blockers plus missing regression coverage. The Node 24 production build passes and emits 55 HTML pages plus 55 Markdown twins, but the twins silently omit non-default tab content, and the generator fails on supported Node 20 environments. Please also add focused automated coverage for successful generation and the no-HTML/missing-content failure paths, as required by AGENTS.md.

illegalcall (approved):
Rechecked the follow-up at a44eb7e6. The tab-content extraction, Node 20 compatibility, regression coverage, and CI wiring are all resolved. Local typechecks, 1,307 frontend tests, the focused 5-test generator suite, production build, Node 20.19.5 generator run, and generated-content assertions all pass.

PR #3397, opened 2026-07-31
fix: prevent duplicate active orchestrators

### LINKED ISSUE #3378: `ao session ls` hides orchestrator-kind sessions → operators spawn duplicate orchestrators unknowingly

Hardik180704:
## What changed

- guard generic orchestrator session creation against an existing active orchestrator
- serialize concurrent orchestrator creation requests per project
- allow a new orchestrator after the previous one is terminated
- leave normal worker session creation unchanged

## Why

The dedicated orchestrator path already prevented duplicates, but generic `POST /api/v1/sessions` requests could bypass that protection and create multiple active orchestrators for one project.

## Testing

- `go test ./internal/service/session`
- `go test -race ./internal/service/session`
- `go vet ./internal/service/session`
- `go test ./...`
- `git diff --check`

Closes #3378

### DIFF
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -166,6 +166,22 @@ func NewWithDeps(d Deps) *Service {
 // Spawn creates a session and returns the API-facing read model plus
 // ephemeral prompt size measurements.
 func (s *Service) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Session, int, int, error) {
+	if cfg.Kind == domain.KindOrchestrator {
+		unlock := s.lockOrchestratorProject(cfg.ProjectID)
+		defer unlock()
+
+		existing, err := s.activeOrchestrators(ctx, cfg.ProjectID)
+		if err != nil {
+			return domain.Session{}, 0, 0, err
+		}
+		if len(existing) > 0 {
+			return newestSession(existing), 0, 0, nil
+		}
+	}
+	return s.spawn(ctx, cfg)
+}
+
+func (s *Service) spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Session, int, int, error) {
 	project, err := s.requireProject(ctx, cfg.ProjectID)
 	if err != nil {
 		return domain.Session{}, 0, 0, err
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

JAYATIAHUJA (approved):
LGTM. Verified locally: no outstanding review comments to address; diff is scoped to backend/internal/service/session and closes the duplicate-orchestrator race by funneling generic Spawn through the per-project orchestrator lock, while SpawnOrchestrator correctly bypasses the guard via the internal spawn (avoids re-acquiring the mutex it already holds). go build, go vet, and go test ./internal/service/session all pass; the full backend suite is green except three session_manager PATH tests that fail identically on upstream/main (pre-existing, environment-specific, unrelated to this change). All CI checks pass.

PR #3399, opened 2026-07-31
feat(mobile): light/dark theming, native sheets, and the rebuilt app shell

Prasad-D-Ware:
Rebuilds the mobile app around a real theme system and native sheets, and reworks every screen on top of it.

The app was dark-only, its modals were hand-rolled on top of RN's `Modal`, a fresh install could not connect at all (wrong default port), and there was no way to un-pair a phone. This addresses those together because they are welded by the theme: removing the exported `theme` const is what forces every screen to migrate, so the palette and the screens cannot land separately.

## What's in

- **Light/dark theming** with a persisted `light`/`dark`/`system` preference. Light is derived from the desktop's `tokens.css`, with surfaces following the iOS grouped-list convention and semantic colours darkened rather than reused — both documented on the palette.
- **Native `formSheet` routes** for project, agent, theme and manual-connect, so drag-to-dismiss and the grabber come from the OS.
- **Rebuilt board, orchestrator, PR and settings tabs**, plus a session composer with the eight control keys a phone keyboard lacks.
- **First-run onboarding** and a rebuilt pairing flow.
- **Notification history**, backed by the daemon's durable record rather than only what push delivered live.
- **Disconnect & forget server** — previously impossible; clearing the host by hand left the device registered and still receiving pushes.
- **Per-harness brand marks**, ported from the desktop's agent assets.

## Fixes that fall out

- **Default port 3011, not 3001.** 3001 is the desktop/CLI daemon, which binds loopback only and can never be reached from a phone.
- **Auth lockout during re-pair.** The daemon locks a device out for a minute after 5 failed auths. A stale password cost 4 failures per poll tick, arming the lockout in a tick or two — so scanning a fresh pairing code failed with 429 before the new password was ever checked. `/sessions` is now probed first and alone, capping a bad-credential tick at one failure.
- **QR scanner picked the wrong lens on iOS.** `expo-camera` matches `selectedLens` against `AVCaptureDevice.localizedName`, not a `DeviceType` identifier, so `"builtInWideAngleCamera"` matched nothing and left the default — on a Pro that is the ultra-wide, which cannot focus close enough to read a QR code off a laptop screen.
- **Session titles disagreed with desktop.** `sessionTitle` now falls back to `issueId`, where the daemon stores the task name.
- **The input dock jumped twice when the keyboard appeared** — the root view and the dock were both compensating.
- **Android sheets sat behind the keyboard.** `react-native-screens` only expands a sheet for the IME when it has more than one detent, so the one-detent `fitToContents` form could not get out of the way.

## Structure

Pure decision logic is split from React Native / Expo side effects throughout (`prView.ts` ↔ `PRCard.tsx`, `onboarding.ts` ↔ `onboardingStore.ts`), following the existing `pushStatus.ts` / `push.ts` split, so the rules are unit-tested without a renderer. That is where most of the new test coverage lives.

The 16 commits are ordered and self-contained; reviewing them in sequence is much easier than the squashed diff.

## Deliberately not included

- **Voice dictation** — follows in a separate PR stacked on this one, to keep the speech dependency and its permissions reviewable on their own.
- **The app icon** — that is #3304.

## Note on #3304

Both PRs touch `app.json`. #3304 owns `ios.icon` and `adaptiveIcon.monochromeImage`; this PR owns `userInterfaceStyle`, `infoPlist`, `softwareKeyboardLayoutMode` and the notification colour. The blocks are adjacent, so whichever merges second needs a small rebase there. No other overlap.

## Testing

- `npm run typecheck` — clean
- `npm test` — 18 files, 237 tests passing


Vaibhaav-Tiwari:
@Prasad-D-Ware screenshots/recording?

Prasad-D-Ware:
@Vaibhaav-Tiwari https://drive.google.com/file/d/1LA_pdyF9rQpLwXdTpzL4PnlOiNQb8tBO/view?usp=sharing

Vaibhaav-Tiwari (changes requested):
Medium (worth fixing)
  1. Disconnecting doesn't fully clear stored credentials — if the daemon is unreachable at disconnect time, the old host+password stay saved instead of being
  wiped.
  2. Session title can render blank — no safeguard against a whitespace-only fallback value, and this logic wasn't tested like everything else was.
  3. Notification tap-routing logic is written twice in two different files — they can quietly drift out of sync.
  4. Dead code: a "claude" logo/icon entry that mobile can never actually trigger.
  5. Same card-styling code copy-pasted in three places instead of shared.
  6. Two tabs (orchestrator, PRs) don't show a proper loading/error state — they just look empty on failure, unlike the board tab which does this correctly.
  7. One place checks for auth/rate-limit errors by string-matching instead of using the status code that's already available — fragile.

  Needs a product decision
  8. The composer no longer lets you free-type keystrokes straight to the terminal — only 8 fixed buttons work now. Probably intentional, but should be
  confirmed, not just assumed.

  Low / nitpicks
  - Leftover unused theme color aliases, one of which is misleadingly named "cyan" but is actually orange.
  - Inconsistent ordering of close/select across similar popup sheets — no bug yet, but easy to introduce one later.
  - An exported function (workerPRCount) that's tested but never actually used anywhere.
  - Multi-line text sent to the terminal in one edge case isn't sanitized and could submit as multiple partial lines.
  - A redundant duplicate event listener for the Android keyboard (harmless, just extra).
  - The notifications screen ignores the server's unread count/pagination info and just recomputes locally, capped at 50 items; a failed refresh fails silently
  if you already have items loaded.
  - The list of "priority" agents is hardcoded a third time in a new place instead of reusing the existing (already duplicated) list.
  - An icon type option that's declared but never actually used.

  No blockers, no major regressions.

Prasad-D-Ware:
@Vaibhaav-Tiwari the comments have been addressed

Vaibhaav-Tiwari (approved):

PR #3401, opened 2026-07-31
feat(telemetry): add kill switch, daemon version stamping, and update telemetry

Pulkit7070:
## Summary

Follow-on to #3314 and #3388, which fixed the PostHog spend problem. Reconciling the bill against the code surfaced three gaps that those PRs did not cover.

**Kill switch.** Every existing telemetry control is compiled into the build, so silencing a noisy stream meant shipping a release and waiting for users to install it. That is why the `ao hooks` firehose ran for weeks. `AO_TELEMETRY_DISABLED_EVENTS` names streams that must not reach PostHog:

```bash
AO_TELEMETRY_DISABLED_EVENTS="ao.v2.app.active, ao.renderer.*"
```

Entries ending in `*` match by prefix, matching is case-insensitive, and either the internal name (`ao.app.active`) or the exported alias (`ao.v2.app.active`) works, so the name visible in PostHog needs no translation. It sits outermost on the remote chain, so a silenced stream consumes no aggregation window, no rate-limit slot, and no export. Local SQLite is deliberately unaffected, so a stream silenced in production stays debuggable locally. Unrecognized entries are inert rather than fatal, because the switch has to be usable during an incident.

**Daemon version stamping.** Every daemon event shipped with no version at all. All 18,661 `ao.session.spawned` / `spawn_failed` events in the current billing period have an empty `ao_version`, so the observed 23.5% session-spawn failure rate cannot be attributed to any release. The daemon binary has no version that release tooling overrides (`cli.Version` is still `"dev"`), so the supervisor passes `AO_TELEMETRY_APP_VERSION` from `app.getVersion()` and the sink stamps `app_version`/`ao_version`. An unset value leaves the properties off rather than reporting a misleading `"unknown"`.

**Update telemetry.** The app auto-downloads updates and installs them on quit, but 1,148 installs sit on 0.10.3 against roughly 70 on 0.11.x, and there is no signal about why. A stuck install looks identical whether the check failed, the download failed, or the user simply never quits the app. Adds `ao.renderer.update_failed`, `ao.renderer.update_downloaded`, and `ao.renderer.update_unsupported`, carrying a coarse `error_category`, the `phase` that failed (`check` or `download`), and the target version.

The updater's raw message is bucketed into a category rather than sent, because it can contain feed URLs and local staging paths. `downloading` is deliberately not reported: it fires per percent tick, the UI already shows it, and repeating streams are what caused the original volume problem. Emitted from the existing `useUpdateStatus` status channel, so it inherits the renderer's per-name rate limits and sanitization.

Also in scope, both small and previously identified: the renderer now sets `advanced_disable_flags` and `disable_surveys` (it reads no flags and ships no surveys, so those polls were requests for data nothing consumes), and unpackaged builds default `AO_TELEMETRY_REMOTE=off` so dev loops no longer export to the production project.

## Validation

Ran the daemon against a local capture endpoint rather than relying on tests alone.

Version stamping, with `AO_TELEMETRY_APP_VERSION=0.11.2`:

```
{"event": "ao.daemon.started",  "app_version": "0.11.2", "ao_version": "0.11.2"}
{"event": "ao.v2.cli.invoked",  "app_version": "0.11.2", "ao_version": "0.11.2"}
{"event": "ao.v2.app.active",   "app_version": "0.11.2", "ao_version": "0.11.2"}
```

Kill switch, with `AO_TELEMETRY_DISABLED_EVENTS="ao.v2.app.active, ao.daemon.*"`. Only the undenied stream is exported:

```
{"event": "ao.v2.cli.invoked", "app_version": "0.11.2", "ao_version": "0.11.2"}
```

And local storage still holds all three, confirming the switch is remote-only:

```
telemetry_event: [('ao.app.active', 1), ('ao.cli.invoked', 1), ('ao.daemon.started', 1)]
```

Also run:

- `go test ./internal/adapters/telemetry ./internal/config ./internal/daemon ./internal/httpd ./internal/telemetrymeta`
- `go vet` and `gofmt` on the touched packages
- `npx vitest run src/renderer/lib/telemetry.test.ts src/renderer/lib/update-telemetry.test.ts` (38 tests)
- `npm run typecheck`

Note: `go test ./internal/cli` fails 6 tests (`TestBrowserRequiresSessionAndValidWait` and 5 spawn tests) on this branch. They fail identically on `origin/main` with and without `AO_*` environment variables set, so they are pre-existing and unrelated.

## Risks and follow-ups

- The kill switch is env-driven, so it reaches an installed build only through the supervisor. Genuinely remote control needs a hosted config endpoint the daemon polls; that is the natural next step and is not attempted here.
- Version stamping only covers daemon events launched by the supervisor. A directly launched `ao daemon` still exports without a version, which is correct, since there is no app version to report.
- Not included: the 52% orchestrator spawn failure rate is a product defect present on every version including the latest, and needs its own investigation rather than more telemetry.


Pulkit7070:
**Rebased on main and re-verified at runtime.**

Merged `origin/main` (which landed #3388) and resolved two conflicts:

- `README.md`: took main's wording verbatim, since it matches the phrasing #3388 put in `docs/telemetry.md`.
- `telemetry.ts`: kept the superset. #3388 added `disable_session_recording`; this branch also adds `advanced_disable_flags` and `disable_surveys`.

Re-ran the runtime checks on the merged tree rather than trusting the pre-merge results, because main's `telemetrymeta` refactor touches the same request path.

Kill switch plus version stamping, with `AO_TELEMETRY_DISABLED_EVENTS="ao.v2.app.active, ao.daemon.*"`. Only the undenied stream is exported, and it carries the version:

```
{"event": "ao.v2.cli.invoked", "app_version": "0.11.2", "ao_version": "0.11.2"}
```

Local SQLite from the same run still holds everything, confirming the switch is remote-only and silenced streams stay debuggable:

```
[('ao.app.active', 1), ('ao.cli.invoked', 1), ('ao.daemon.started', 1)]
```

`ao hooks claude-code post-tool-use` remains silent after the merge, so main's normalization refactor and this change do not interfere.

Renderer verified by running the real Electron app with `VITE_AO_POSTHOG_HOST` pointed at a local capture endpoint. Every request went to `/e/` and **nothing** went to `/flags`, `/decide`, or `/s/`, which is the runtime confirmation that the three disabled products actually take effect:

```
ao.v2.renderer.loaded          [ao_version, app_version, build_mode, platform, surface, telemetry_schema_version]
ao.v2.app.active               [..., channel]
ao.v2.renderer.daemon_failure  [..., code, daemon_state]
ao.v2.renderer.route_viewed    [..., surface]
ao.v2.renderer.api_error x2    [..., error_category, operation, status]
```

Dev gate confirmed by reading the live daemon's environment under `npm run dev`:

```
AO_TELEMETRY_REMOTE=off
AO_TELEMETRY_APP_VERSION=0.10.3
```

Also re-run clean on the merged tree: `go build ./...`, `go test` for `adapters/telemetry`, `config`, `daemon`, `httpd`, `telemetrymeta`, 38 frontend tests, and `npm run typecheck`.

## One thing a reviewer should confirm

**The three `ao.renderer.update_*` events are covered by unit tests but were not driven at runtime.** The path is reachable in a dev build, since `auto-updater.ts:497` broadcasts `state: "unsupported"` when unpackaged, but nothing triggers it automatically: `checkForUpdatesNow` is only called from the `updates:check` IPC handler, never at startup. Driving it needs a click on Check for updates in the Electron Settings window, which I could not automate (`electron-forge start` does not forward `--remote-debugging-port`, and browser preview cannot help because `aoBridge` resolves eagerly at module load so a bridge cannot be injected post-load).

If you want it confirmed before merge: run the app, open Settings then Updates, click Check for updates, and `ao.renderer.update_unsupported` should be emitted with `error_category: "not_supported"`.

Unrelated and pre-existing: `go test ./internal/cli` fails 6 tests (`TestBrowserRequiresSessionAndValidWait` and 5 spawn tests) identically on `origin/main`.

backend/internal/daemon/telemetry_wiring.go
```diff
@@ -53,5 +54,9 @@ func newTelemetrySink(cfg config.Config, store *sqlite.Store, log *slog.Logger)
 	// job for every event name that isn't aggregated.
 	rateLimited := telemetryadapter.NewRateLimitedSink(remote, aggregatedEventNames)
 	aggregated := telemetryadapter.NewAggregatingSink(rateLimited, aggregatedEventNames, time.Minute)
-	return telemetryadapter.NewFanoutSink(local, aggregated)
+	// The kill switch sits outermost on the remote chain so a silenced stream
+	// costs nothing downstream: no aggregation window, no rate-limit slot, no
+	// export. Local storage is unaffected, so silenced events stay debuggable.
+	denied := telemetryadapter.NewDenylistSink(aggregated, cfg.Telemetry.DisabledEvents)
```
illegalcall, line 60: **P1 — the advertised kill switch cannot silence renderer streams.** This wrapper is correctly outermost for the daemon's billed branch, but renderer events call PostHog directly from `frontend/src/renderer/lib/telemetry.ts`. With `AO_TELEMETRY_DISABLED_EVENTS="ao.renderer.*"`, real renderer events never reach this sink; denying `ao.v2.app.active` also leaves the renderer producer active under the same exported name. Please either propagate the remote/deny policy into renderer capture or route renderer events through a policy-enforcing owner. A boundary test should emit the same alias from both producers, deny it once, assert zero remote captures, and still assert the daemon-local SQLite record.

frontend/src/renderer/hooks/useUpdateStatus.ts
```diff
@@ -9,12 +10,22 @@ import { aoBridge } from "../lib/bridge";
...
-		void aoBridge.updates.getStatus().then((s) => {
-			if (live) setStatus(s);
-		});
-		const off = aoBridge.updates.onStatus(setStatus);
+		const observe = (next: UpdateStatus) => {
+			if (!live) return;
+			setStatus(next);
+			const previous = lastReported.current;
+			lastReported.current = next;
+			void captureUpdateStatusTransition(previous, next);
```
illegalcall: **P1 — automatic check and download failures never reach this observer.** In `auto-updater.ts`, when `activeUpdaterOperation === "automatic-check"`, the updater error handler restores the previous status and returns; the outer catch only logs. Because that operation also awaits `downloadPromise`, both automatic check and automatic download failures bypass this status-based telemetry. That is the principal stuck-auto-update path the PR says it will diagnose. Capture the safe outcome at the main-process operation boundary (while keeping UI suppression separate), and test both a rejected check and rejected download promise.

frontend/src/main.ts
```diff
@@ -398,12 +398,25 @@ let shellEnvPromise: Promise<void> | null = null;
...
+//
+// Unpackaged builds keep local event recording but never export to PostHog: a
+// dev loop or a CI job driving the real app would otherwise bill production
+// events and inflate install/DAU counts. Set AO_TELEMETRY_REMOTE explicitly to
+// exercise the export path from a dev build.
 function telemetryOverrides(): Record<string, string> {
 	return {
 		AO_TELEMETRY_EVENTS: process.env.AO_TELEMETRY_EVENTS ?? "on",
-		AO_TELEMETRY_REMOTE: process.env.AO_TELEMETRY_REMOTE ?? "posthog",
+		AO_TELEMETRY_REMOTE: process.env.AO_TELEMETRY_REMOTE ?? (isDev ? "off" : "posthog"),
```
illegalcall, line 410: **P1 — this dev default disables only daemon export.** The renderer still resolves the embedded production PostHog key and initializes its own client unconditionally, so ordinary unpackaged sessions continue sending renderer loaded/active/error events to production. Please expose a shared renderer remote-enabled policy through the bootstrap (or return no bootstrap in dev unless explicitly opted in). Prove it with an unpackaged-default test asserting no renderer PostHog init/capture, plus an explicit opt-in case.

frontend/src/renderer/lib/update-telemetry.ts
```diff
@@ -0,0 +1,102 @@
...
+	if (previous && previous.state === next.state && previous.version === next.version) return null;
+
+	switch (next.state) {
+		case "error":
+			return {
+				event: "ao.renderer.update_failed",
+				properties: {
+					// Which stage failed, inferred from where we came from. The updater
+					// does not label its own errors by phase.
+					phase: previous?.state === "downloading" ? "download" : "check",
```
illegalcall: **P2 — phase and target version are not authoritative here.** A manual download can reject before its first `downloading` push, leaving `previous.state === "available"`; this records a download failure as `phase: "check"`. Main-process error statuses also omit `version`, so `to_version` is undefined both before progress and after a versionless progress state. Please carry explicit operation phase and target version from Electron main, or emit telemetry at that owner. Tests should cover download rejection before and after progress and require `phase: "download"` plus the selected version.

frontend/src/renderer/hooks/useUpdateStatus.ts
```diff
@@ -9,12 +10,22 @@ import { aoBridge } from "../lib/bridge";
  */
 export function useUpdateStatus(): UpdateStatus {
 	const [status, setStatus] = useState<UpdateStatus>({ state: "idle" });
+	// Two components mount this hook (the sidebar row and the settings section),
+	// and both would otherwise report the same transition. The ref is module-free
+	// per hook instance, so the shared reservation in captureRendererEvent is what
+	// actually collapses the duplicate; this only avoids the obvious repeat.
+	const lastReported = useRef<UpdateStatus | null>(null);
```
illegalcall: **P2 — this hook-local ref does not collapse duplicate consumers.** Sidebar and UpdatesSection each mount `useUpdateStatus`, so each instance has its own `lastReported` and emits the same outcome. `reserveCapture` allows five events per name per minute; it rate-limits but does not deduplicate an outcome. Move telemetry to a singleton/main-process owner or deduplicate by a stable operation/outcome identity. Mount both consumers, broadcast one downloaded status, and assert exactly one capture.

backend/internal/adapters/telemetry/posthog.go
```diff
@@ -188,15 +188,30 @@ type PostHogSink struct {
...
 	log        *slog.Logger
 	ch         chan ports.TelemetryEvent
 	wg         sync.WaitGroup
 	closeOnce  sync.Once
 }
 
+// PostHogOption configures a sink at construction. Options are applied before
+// the export goroutine starts, so the fields they set are only ever written on
+// the constructing goroutine and read on the export one.
+type PostHogOption func(*PostHogSink)
```
illegalcall: **Standards — one-off abstraction.** `AGENTS.md` says not to add abstractions for one-off cases. This introduces a functional-options layer with one option and one production caller. Please pass the app version/config directly until multiple independent options exist; that keeps the constructor contract explicit and avoids speculative plumbing.

backend/internal/config/config.go
```diff
@@ -239,6 +249,12 @@ func Load() (Config, error) {
 	if raw := os.Getenv("AO_TELEMETRY_POSTHOG_HOST"); raw != "" {
 		cfg.Telemetry.PostHogHost = raw
 	}
+	if raw := os.Getenv("AO_TELEMETRY_DISABLED_EVENTS"); raw != "" {
```
illegalcall, line 252: **Boundary-test gap.** The repository requires tests at the changed user-visible boundary. `config_test.go` neither clears these two new recognized variables in its defaults setup nor sets/asserts them in the override path. The supervisor dev/version/deny forwarding and multi-consumer update subscription are also untested; the current inner-unit tests cannot catch the cross-process defects above. Please add config-load assertions plus end-to-end wiring tests for remote/local denial, packaged-vs-dev renderer export, automatic failures, and duplicate consumers.

frontend/src/renderer/lib/update-telemetry.test.ts
```diff
@@ -0,0 +1,72 @@
+import { describe, expect, it } from "vitest";
+import type { UpdateStatus } from "../../main/update-settings";
+import { updateFailureCategory, updateTelemetryFor } from "./update-telemetry";
+
+const at = (state: UpdateStatus["state"], extra: Partial<UpdateStatus> = {}): UpdateStatus => ({
```
illegalcall: **Judgment/naming:** `at` does not communicate that this constructs an `UpdateStatus`. A name such as `statusWithState` or `updateStatus` would make the transition tests readable without mentally expanding the helper.

illegalcall (changes requested):
Requesting changes because three core promises in the PR description do not hold across AO's daemon/renderer boundary: the kill switch does not govern renderer-direct telemetry, automatic updater failures remain invisible, and the dev default still allows renderer events into production PostHog. I also found two data-quality defects in update telemetry plus three standards/test-quality issues. Each unique finding is attached inline with a concrete trigger, outcome, repair direction, and proof test.

backend/internal/daemon/telemetry_wiring.go
Pulkit7070: Confirmed and fixed. Renderer events call PostHog directly and never pass through this sink, so denying `ao.v2.app.active` silenced the CLI producer while the renderer kept sending under the same exported name.

The policy now crosses the boundary on the telemetry bootstrap (`shared/telemetry.ts`), and `renderer/lib/telemetry.ts` enforces it with the same semantics as `DenylistSink`: case-insensitive, `*` prefix matching, and matching either the internal name or the exported alias. It is checked before the rate-limit reservations so a silenced stream consumes no slot, matching this wrapper's outermost placement.

This line itself is unchanged, which is why the thread reads as not outdated.

Verified by running a packaged-style build against a local capture endpoint with `AO_TELEMETRY_DISABLED_EVENTS="ao.v2.renderer.loaded"`:

```
ao.v2.app.active               sent
ao.v2.renderer.api_error       sent
ao.v2.renderer.daemon_failure  sent
ao.v2.renderer.route_viewed    sent
ao.v2.renderer.loaded          silenced
```

Test: "honours the supervisor deny list, by either internal or exported name" in `telemetry.test.ts`, plus the boundary-carrying test in `shared/telemetry.test.ts`.

frontend/src/renderer/hooks/useUpdateStatus.ts
Pulkit7070: You were right, and this was the most important finding in the review. `auto-updater.ts` logs, calls `restoreAutomaticCheckPreviousStatus()`, and returns without broadcasting, and automatic checks run hourly, so the principal silent-failure path was invisible to a status-based observer.

Telemetry moved out of this hook entirely and into the main process at the updater's operation boundary. Outcomes are pushed on `updates:telemetry`, a channel separate from `updates:status`, so suppressing a status for UI reasons no longer suppresses its telemetry. `emitUpdateFailure(err)` now sits above the `automatic-check` early return, so it cannot be skipped.

An `ao.renderer.update_*` event also carries a new `trigger` field (`automatic` | `manual`) so the two cases are separable in analysis.

Proof tests, both requested: the existing `auto-updater.test.ts` cases for a rejected check and a rejected download promise now assert per channel rather than asserting nothing reaches the renderer at all:

```js
expect(statusMessages()).toEqual([]);                 // UI still suppressed
expect(module.getUpdateStatus()).toEqual({ state: "idle" });
expect(telemetryMessages().map(m => m.payload)).toEqual([{
  event: "ao.renderer.update_failed", phase: "check",
  trigger: "automatic", error_category: "unknown",
}]);
```

One thing I have not done: watched a real hourly automatic failure fire end to end. The path is wired and unit-tested, not observed.

frontend/src/main.ts
Pulkit7070: Correct, and fixed at the bootstrap instead of here.

`buildTelemetryBootstrap` now returns `null` on unpackaged builds unless `AO_TELEMETRY_RENDERER=on`. That is stronger than gating each capture, because `initTelemetry()` bails on a null bootstrap, so the renderer never constructs a PostHog client and the SDK's own background requests stop too. `AO_TELEMETRY_RENDERER=off` also opts a packaged build out.

This line is unchanged, hence the thread not showing as outdated.

Verified by running an unpackaged build against a local capture endpoint: **zero** renderer events. The identical run before this change produced five (`renderer.loaded`, `app.active`, `daemon_failure`, `route_viewed`, `api_error`).

Tests in `shared/telemetry.test.ts` cover the unpackaged default, the explicit opt-in, and the packaged opt-out.

frontend/src/renderer/lib/update-telemetry.ts
Pulkit7070: Fixed by moving ownership to the main process, which is the only place that knows either value.

`activeUpdaterPhase` is set per operation (`manual-download` starts as `download`, everything else as `check`) and flips to `download` on the first `download-progress` event, since any progress proves the check succeeded. `pendingUpdateVersion` is captured from `update-available`. So a manual download that rejects before any progress is no longer mislabelled `check`, and `to_version` no longer depends on an error status carrying a version it never had.

`to_version` is omitted rather than sent as `undefined` when genuinely unknown.

Covered in `shared/update-telemetry.test.ts` ("records the phase and trigger it was given") and by the per-channel assertions in `auto-updater.test.ts`.

frontend/src/renderer/hooks/useUpdateStatus.ts
Pulkit7070: Confirmed. I measured it while verifying: booting the app showed `updateListeners: 2`, one each for Sidebar and UpdatesSection, and you are right that `reserveCapture` rate-limits rather than deduplicates, so both would have survived.

This hook is now back to its original telemetry-free form. Update telemetry is subscribed exactly once, from `main.tsx`, in `renderer/lib/update-telemetry.ts` — the same shape as `startDaemonFailureTelemetry()`. One outcome, one capture, regardless of how many components mount the hook.

backend/internal/adapters/telemetry/posthog.go
Pulkit7070: Fair, and reverted. `NewPostHogSink` now takes `appVersion` as a direct parameter and `PostHogOption`/`WithAppVersion` are gone. All call sites updated.

I had used variadic options to avoid churning six call sites, which was a reason but not a justification against an explicit rule in `AGENTS.md`.

backend/internal/config/config.go
Pulkit7070: Added. `config_test.go` now clears `AO_TELEMETRY_DISABLED_EVENTS` and `AO_TELEMETRY_APP_VERSION` in the defaults setup, and two new cases assert the override path:

- `TestLoadTelemetryDisabledEventsAndAppVersion` — comma-separated parsing, whitespace trimming, blank entries dropped, version trimmed
- `TestLoadTelemetryDisabledEventsBlankIsInert` — a blank list never fails the load, because the switch has to be usable in a hurry

On the cross-process gap you flagged: `shared/telemetry.test.ts` covers the packaged-vs-dev renderer decision and the deny list crossing the boundary, and `auto-updater.test.ts` covers automatic failures per channel. Duplicate consumers are addressed structurally rather than by test, since the hook no longer reports at all.

This line is unchanged, which is why the thread is not marked outdated.

frontend/src/renderer/lib/update-telemetry.test.ts
Pulkit7070: Resolved by deletion. That file is gone: the renderer-side mapping module moved to `shared/update-telemetry.ts` when ownership moved to the main process, and its tests were rewritten without the helper.

Pulkit7070:
@illegalcall @DhruvSharma all eight threads are addressed and replied to individually, with the commit and the proof test named on each. Summary for anyone skimming.

Three findings undercut claims in my own PR description, and you were right on all three:

- **Kill switch did not cover renderer streams.** I had verified it only with daemon-originated events and described it as covering the app. The policy now crosses the process boundary on the telemetry bootstrap and the renderer enforces the same matching semantics as `DenylistSink`.
- **Automatic update failures never reached telemetry.** `auto-updater.ts` logs, restores the previous status, and returns without broadcasting, and automatic checks run hourly, so the exact silent-failure case the feature existed to diagnose was invisible. Outcomes moved to the main process at the updater's operation boundary, on a channel separate from `updates:status`, so UI suppression no longer suppresses telemetry.
- **Dev default only gated the daemon.** The renderer still initialised with the production key on unpackaged builds. The supervisor now withholds the bootstrap entirely in dev, so no client is constructed at all.

The two P2s are fixed structurally rather than patched: `phase` and `to_version` are now owned by the main process, which is the only place that knows them, and the duplicate-consumer problem is gone because the hook no longer reports.

Standards and test gaps: functional options replaced with a direct parameter per `AGENTS.md`, config-load tests added for both new environment variables, and the poorly named test helper went away with the file it lived in.

Verified by running rather than by tests alone:

```
unpackaged build, no opt-in        -> 0 renderer events (previously 5)
deny "ao.v2.renderer.loaded"      -> that stream silenced, the other four still sent
daemon deny + version stamping    -> only undenied stream exported, local SQLite kept all three
```

CI is green on all 10 checks and the branch is rebased on main and mergeable.

Two things I am not claiming, so they do not get lost:

- I have not observed a real hourly automatic failure end to end. That path is wired so it cannot be skipped, and it is unit-tested per channel, but it is not runtime-observed.
- `advanced_disable_flags` / `disable_surveys` save nothing today. Billing shows `feature_flag_requests` usage at zero. They are correctness, not cost, and I described them as savings earlier which was wrong.

Also folded in a fix for the CSP violation Prince hit on Windows: the built CSP was derived from `api_host` alone, so posthog-js's sibling `-assets` origin was blocked on every launch of a packaged build. No events or user counts were ever lost, since capture uses `api_host`; the symptom was a red console line that read like a broken integration.

illegalcall (approved):

DhruvSharma:
Hey,

I think you have been tagging the wrong person here.

Pulkit7070:
Ohh my bad : )

PR #3404, opened 2026-07-31
fix(cursor): stop activity state sticking on "Input Needed" all turn

### LINKED ISSUE #3306: bug(cursor): activity state stays "Input Needed" for the whole turn — every tool call re-asserts sticky waiting_input, nothing clears it until stop
## Bug

Image

Once a Cursor-adapter session's activity state flips to **"Input Needed"** (`waiting_input`), it never clears back to "Working" even after the user sends a follow-up and the agent visibly resumes — editing files, running commands, streaming tool output. The topbar badge and the Activity panel stay glued to "Input Needed" for the entire remainder of the turn, and only reset at the *next* turn boundary (submitting another prompt, or the agent's `stop` hook).

**Source:** local observation, desktop app, agent=`cursor` | **Confidence:** High — traced through the full hook → activity-state pipeline, not just the symptom
**Analyzed against:** origin/main `6b6190deb`

## Repro

1. Start a Cursor-agent session and let it hit a state where it shows "Input Needed" (e.g. it ran a shell/MCP call that Cursor's hook reports as a permission-style event).
2. Send a follow-up message in "Add a follow-up".
3. Watch the terminal: the agent resumes, edits files, runs tools (visibly active).
4. The topbar pill and the Activity panel on the right (Session Inspector → Summary → Activity) keep showing **"Input Needed"** the whole time, instead of switching to active/working.

## Root Cause

Cursor's adapter only installs 5 hooks (`backend/internal/adapters/agent/cursor/hooks.go:60-66`):

```go
var cursorManagedHooks = []cursorHookSpec{
    {Event: "sessionStart", Command: cursorHookCommandPrefix + "session-start"},
    {Event: "beforeSubmitPrompt", Command: cursorHookCommandPrefix + "user-prompt-submit"},
    {Event: "stop", Command: cursorHookCommandPrefix + "stop"},
    {Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "permission-request"},
    {Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "permission-request"},
}
```

There is no `afterShellExecution`/`afterMCPExecution` (post-tool) hook, and cursor has no `DetectTerminalActivity` fallback either (unlike codex's `backend/internal/adapters/agent/codex/terminal_activity.go`) — its activity is purely hook-driven via the shared, payload-blind deriver: `activitydispatch.Derivers["cursor"] = activitystate.StandardDeriveActivityState` (`backend/internal/adapters/agent/activitydispatch/dispatch.go:40`).

`StandardDeriveActivityState` (`backend/internal/adapters/agent/activitystate/activitystate.go:24-33`) maps `permission-request → domain.ActivityWaitingInput` unconditionally — it can't distinguish "this tool call needs the user's OK" from "this is just a normal shell/MCP call Cursor happens to fire the same pre-execution hook for." Its own doc comment already flags this class of adapter: *"permission-request maps to waiting_input, not blocked: none of the sharing adapters install the pre/post-tool-use trio, so a blocked state could never be cleared before the turn end."*

The precedence guard that's supposed to stop background tool traffic from re-asserting a sticky state only recognizes Claude-style event names (`backend/internal/lifecycle/manager.go:409-412`):

```go
func isToolUseEvent(event string) bool {
    return event == "pre-tool-use" || event == "post-tool-use" || event == "post-tool-use-failure"
}
```

Cursor never emits those — its tool-call signal is always `"permission-request"`. So the sticky-suppression branch (`manager.go:516-519`, `case cur.IsSticky() && isToolUseEvent(s.Event): return suppressed`) never triggers for cursor, and every `beforeShellExecution`/`beforeMCPExecution` during the turn falls through to `default` (`manager.go:522-525`) and overwrites the state to `waiting_input` again, with nothing to demote it until `stop` or the next `user-prompt-submit`.

Sequence:
1. User sends follow-up → `beforeSubmitPrompt` → `user-prompt-submit` → state = `active`.
2. Cursor's first tool call (edit/shell) → `beforeShellExecution` → `permission-request` → state = `waiting_input` (sticky).
3. Every subsequent tool call for the rest of the turn re-asserts `waiting_input` (not suppressed, since `isToolUseEvent` doesn't recognize `permission-request`).
4. Nothing clears it until the agent's `stop` hook fires — i.e., the badge reads "Input Needed" for the agent's entire active working period.

This is the same structural issue already called out for Copilot's adapter (`backend/internal/adapters/agent/copilot/hooks.go:71-80`, "every preToolUse currently produces ActivityWaitingInput" — a documented "many-to-one collapse"), but Cursor has no equivalent comment flagging it, and unlike Copilot's guard, cursor's `permission-request` events aren't even suppressed once already sticky.

Frontend is a pure render of the backend-derived state (`frontend/src/renderer/lib/session-presentation.ts:32-34`, `waiting_input → label: "Input Needed"`) — confirmed this is not a frontend caching/stale-poll issue.

## Suggested Fix

Cursor does have a real post-execution hook AO isn't using: **`afterShellExecution`** is documented at `cursor.com/docs/hooks` with a payload of `{command, output, duration, sandbox}`, and is one of the events Cursor CLI users report actually firing for `cursor-agent` (not just the IDE agent) — see the community thread below. AO's adapter currently only installs the *before* hooks (`hooks.go:60-66`); it never installs `afterShellExecution`.

Proposed change:
- Register `afterShellExecution` in `cursorManagedHooks` (`hooks.go:60-66`), mapped to a post-tool-use-style AO sub-command (e.g. reuse `post-tool-use`, or a cursor-specific one if the payload doesn't carry a `tool_use_id` cleanly).
- Extend `isToolUseEvent` (`backend/internal/lifecycle/manager.go:409-412`) to recognize that event, so the existing sticky-suppression guard (`manager.go:516-519`) actually protects cursor sessions the way it already protects claude-code/codex — background tool traffic stops re-asserting `waiting_input` once it's already sticky, and the paired post-hook can demote it back to active.
- **Open question, needs runtime verification before relying on it:** a Cursor forum thread (April 2026, community-sourced, not confirmed against Cursor's official docs) reports that in `cursor-agent` CLI specifically, `beforeMCPExecution`/`afterMCPExecution` and `beforeSubmitPrompt` do **not** fire, unlike in the IDE agent — only `beforeShellExecution`/`afterShellExecution`/`afterFileEdit`/`postToolUse`/`stop`/`sessionStart` reportedly fire. If accurate, two follow-ups are needed: (a) AO's `beforeMCPExecution` registration (`hooks.go:65`) may currently be dead weight for CLI sessions and MCP tool calls would need a different signal (`afterFileEdit`/`postToolUse` are both listed as CLI-supported and worth checking as a substitute), and (b) if `beforeSubmitPrompt` truly never fires, the turn-boundary reset back to `active` (`activitystate.go:30`, `user-prompt-submit → active`) may not be happening at all today, which would make the bug worse than "stuck for one turn" — worth capturing before scoping the fix.
- At minimum, add the same doc-comment callout `copilot/hooks.go:71-80` has, on `cursor/hooks.go`, so this class of gap isn't rediscovered per-adapter.

## Sources

- [Cursor Hooks reference](https://cursor.com/docs/hooks) — confirms `afterShellExecution` exists with `{command, output, duration, sandbox}` payload
- [Cursor CLI doesn't send all events defined in hooks (Cursor forum)](https://forum.cursor.com/t/cursor-cli-doesnt-send-all-events-defined-in-hooks/148316) — community report of CLI vs IDE hook-firing differences; unverified against AO's own runtime, flagged as such above

## Related

- Same underlying `StandardDeriveActivityState` many-to-one collapse as documented for Copilot: `backend/internal/adapters/agent/copilot/hooks.go:71-80`
- #2814 — kimi activity pill shows Idle while agent is actively working (no mid-turn tool-use hooks) — same family of "adapter has no way to signal true working state" bugs, different symptom (idle vs. stuck waiting_input)
- #2525 — idle_prompt notification conflated with permission_prompt, both set activity=waiting_input
- #2047 — Claude sessions misreport state — active work shown as "ready", idle_prompt shown as "waiting_input"

Lucifer4255:
## Summary

Fixes #3306.

Cursor's `beforeShellExecution`/`beforeMCPExecution` hooks were dispatched to AO as `permission-request`, which the shared `activitystate.StandardDeriveActivityState` maps to `waiting_input` (a sticky state). No `afterShellExecution`/`afterMCPExecution` hook was installed to demote it back, and no fallback terminal-activity detection exists for cursor (unlike codex). So once the first tool call flipped the badge to "Input Needed", nothing cleared it until the turn's `stop` hook — even while the agent was visibly editing files and running commands for the rest of the turn.

The `permission-request` mapping was also the wrong model to begin with: AO always launches `cursor-agent` with `--trust` (`cursor.go`'s `GetLaunchCommand`/`GetRestoreCommand`), so these hooks never represent a real blocking permission decision — they fire on every auto-approved tool call. They're functionally equivalent to claude-code's `PreToolUse`/`PostToolUse` trio, not to a permission dialog.

## Change

Rather than reusing `permission-request` and bolting on the same `toolFlight`/`blockedCandidate` correlation machinery `manager.go` uses to safely clear claude-code's sticky `blocked` state, this follows the simpler half of the existing pattern:

- **`cursor/hooks.go`**: `beforeShellExecution`/`beforeMCPExecution` now dispatch as `pre-tool-use` (was `permission-request`). Their previously-unregistered `afterShellExecution`/`afterMCPExecution` counterparts are now installed too, dispatching as `post-tool-use` — the same sub-command names claude-code uses for its own tool-use trio.
- **`cursor/activity.go`** (new): cursor gets its own `DeriveActivityState`, replacing the shared `activitystate.StandardDeriveActivityState`. `pre-tool-use`/`post-tool-use` both map to `active` (matching how claude-code's trio maps to `active`, not `waiting_input`/`blocked`). `session-start`/`user-prompt-submit` → `active`, `stop` → `idle`.
- **`activitydispatch/dispatch.go`**: registers `cursor.DeriveActivityState` for the `"cursor"` token.

Because cursor's tool-use signals now always resolve to `active`, the session never enters a sticky state in the first place — so the existing precedence guard in `manager.go` (`isToolUseEvent`, `applyToolPrecedenceLocked`) doesn't need to change, and no per-tool correlation tracking is needed for cursor.

**Trade-off**: cursor sessions can no longer show "Input Needed" at all, since Cursor exposes no separate hook that fires only for a genuine blocking dialog (unlike claude-code's distinct `PermissionRequest` event). Given `--trust` means that indicator was never trustworthy to begin with (same "many-to-one collapse" already flagged in `copilot/hooks.go`), this seemed like the honest trade rather than leaving a misleading sticky state in place.

The forum-sourced caveat in #3306 (that `beforeMCPExecution`/`afterMCPExecution` may not fire for the `cursor-agent` CLI) is left as an open follow-up — registering them is harmless dead weight if unconfirmed, consistent with the issue's own suggested fix.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` passes (all packages, including `cursor`, `activitydispatch`, `lifecycle`)
- [ ] Manual repro from #3306: start a Cursor session, hit "Input Needed", send a follow-up, confirm the badge reflects "active" through subsequent tool calls instead of sticking

### DIFF
--- a/backend/internal/adapters/agent/activitydispatch/dispatch.go
+++ b/backend/internal/adapters/agent/activitydispatch/dispatch.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/agy"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/claudecode"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/codex"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/cursor"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/droid"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/fake"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/opencode"
@@ -37,7 +38,7 @@ var Derivers = map[string]DeriveFunc{
 	"opencode":    opencode.DeriveActivityState,
 	"goose":       activitystate.StandardDeriveActivityState,
 	"devin":       activitystate.StandardDeriveActivityState,
-	"cursor":      activitystate.StandardDeriveActivityState,
+	"cursor":      cursor.DeriveActivityState,
 	"qwen":        activitystate.StandardDeriveActivityState,
 	"copilot":     activitystate.StandardDeriveActivityState,
 	"kimi":        activitystate.StandardDeriveActivityState,
--- a/backend/internal/adapters/agent/cursor/activity.go
+++ b/backend/internal/adapters/agent/cursor/activity.go
@@ -1,0 +1,32 @@
+package cursor
+
+import "github.com/aoagents/agent-orchestrator/backend/internal/domain"
+
+// DeriveActivityState maps a Cursor hook event onto an AO activity state.
+//
+// event is the AO hook sub-command name installed in cursorManagedHooks
+// ("user-prompt-submit", "pre-tool-use", ...), NOT the native Cursor event
+// name.
+//
+// cursor-agent is always launched with --trust, so beforeShellExecution/
+// beforeMCPExecution (and their after* counterparts) never pause for a real
+// permission decision — they fire on every auto-approved tool call. They are
+// therefore treated like claude-code's PreToolUse/PostToolUse trio (active),
+// not like a permission dialog. Unlike claude-code, Cursor exposes no
+// separate hook that fires only for a genuine blocking dialog, so a cursor
+// session never enters waiting_input/blocked: there is no signal AO could use
+// to know one is happening.
+func DeriveActivityState(event string, _ []byte) (domain.ActivityState, bool) {
+	switch event {
+	case "session-start":
+		return domain.ActivityActive, true
+	case "user-prompt-submit":
+		return domain.ActivityActive, true
+	case "pre-tool-use", "post-tool-use":
+		return domain.ActivityActive, true
+	case "stop":
+		return domain.ActivityIdle, true
+	default:
+		return "", false
+	}
+}
--- a/backend/internal/adapters/agent/cursor/hooks.go
+++ b/backend/internal/adapters/agent/cursor/hooks.go
@@ -57,12 +57,23 @@ type cursorHookSpec struct {
 // cursorManagedHooks is the source of truth for the hooks AO installs. The
 // native-event → AO-subcommand contract is FIXED: the orchestrator's CLI hook
 // dispatch and activity.go agree on the sub-command names.
+//
+// The before*/after* pairs are dispatched as pre-tool-use/post-tool-use, the
+// same tokens claude-code uses for its own tool-use trio — NOT as
+// permission-request. cursor-agent is always launched with --trust
+// (cursor.go's GetLaunchCommand/GetRestoreCommand), so these hooks never
+// actually pause for a real permission decision; they fire on every
+// auto-approved tool call. Mapping them to permission-request's
+// waiting_input, with no paired after-hook to demote it, is what left the
+// activity badge stuck on "Input Needed" for the whole turn. See DeriveActivityState.
 var cursorManagedHooks = []cursorHookSpec{
 	{Event: "sessionStart", Command: cursorHookCommandPrefix + "session-start"},
 	{Event: "beforeSubmitPrompt", Command: cursorHookCommandPrefix + "user-prompt-submit"},
 	{Event: "stop", Command: cursorHookCommandPrefix + "stop"},
-	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "permission-request"},
-	{Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "permission-request"},
+	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "pre-tool-use"},
+	{Event: "afterShellExecution", Command: cursorHookCommandPrefix + "post-tool-use"},
+	{Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "pre-tool-use"},
+	{Event: "afterMCPExecution", Command: cursorHookCommandPrefix + "post-tool-use"},
 }
 
 // GetAgentHooks installs AO's Cursor hooks into the worktree-local


illegalcall:
Hi! Can you please attach a small screen recording of before and after the fix?

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@Lucifer4255 
#4495 landed the same three dispatch files for this fix. Closing as superseded.

PR #2986, opened 2026-07-22
ci(build-artifacts): emit Linux .deb as linux-deb artifact

harshitsinghbhandari:
## Summary
The release migration regressed the Linux `.deb`: current releases ship one, but `build-artifacts.yml` only captured the AppImage. `forge.config.ts` already runs `@electron-forge/maker-deb` on linux, so this captures its existing output; no packaging changes.

- Copies the built `.deb` to the version-less alias `agent-orchestrator-linux-x64.deb` (consistent with the four existing aliases).
- Uploads it as a new run artifact named `linux-deb` (`retention-days: 1`), a single file inside.
- Merges its sha256 into the linux-x64 digest fragment, so the merged `digests.json` gains exactly one new key: `agent-orchestrator-linux-x64.deb`.

Additive only: mac/win legs, the four existing artifacts, and signing are untouched. The deb is staged in its own `dist-deb/` dir because the `linux` artifact's `agent-orchestrator-linux-x64.*` glob would otherwise swallow it.

Closes #2983

## Verification
- `actionlint` clean.
- Proof dispatch on this branch (ref = main sha `8b78627cc`): https://github.com/AgentWrapper/agent-orchestrator/actions/runs/29941224512 (will confirm `linux-deb` artifact + digests key once the run completes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3406, opened 2026-07-31
ci(build-artifacts): ship the Linux .rpm as a release artifact (#3405)

### LINKED ISSUE #3405: Ship a .rpm package for Fedora/RHEL
We ship a .deb and an AppImage for Linux, but no .rpm, so Fedora/RHEL/openSUSE users have no native package. Add an rpm target to the Linux build (electron-builder supports an `rpm` target alongside deb/AppImage), publish it as a release asset like the .deb, and wire it through the build-artifacts workflow and the release conductor.

harshitsinghbhandari:
Closes #3405.

## What

The rpm target was already configured (`frontend/forge.config.ts` has `@electron-forge/maker-rpm`) and it already builds on the ubuntu leg of `build-artifacts.yml`: the run log for the current workflow shows `Making a rpm distributable for linux/x64` succeeding, because `ubuntu-latest` ships `rpmbuild`. The output was simply never collected, so the release conductor had no `.rpm` to verify or attach and Fedora/RHEL/openSUSE users got no native package.

This wires it through, mirroring exactly what the `.deb` got in #2986:

- `Collect Linux .rpm and compute digest`: copies `out/make/rpm/x64/*.rpm` to `dist-rpm/agent-orchestrator-linux-x64.rpm` and folds its sha256 into the existing `linux-x64.digest.json` fragment, so the `digests` job still reads one fragment per platform.
- `Upload Linux .rpm artifact`: uploads it as the pinned `linux-rpm` artifact name (own directory, so the `linux` artifact's `agent-orchestrator-linux-x64.*` glob cannot swallow it).
- README: adds the `.rpm` row to the download table, and the `.deb` row alongside it (the deb has been a published release asset since #2986 and was never documented).

Artifact name `agent-orchestrator-linux-x64.rpm` follows the existing stable-alias convention (`agent-orchestrator-linux-x64.AppImage`, `agent-orchestrator-linux-x64.deb`), which is what `releases/latest/download/...` links and the conductor's staging step expect.

No `forge.config.ts` change is needed and no signing is involved.

## Conductor follow-up

The private release conductor needs two lines in `_pipeline.yml`'s `Verify and stage build artifacts` step: `mv artifacts/linux-rpm/agent-orchestrator-linux-x64.rpm dist/` and adding the filename to the digest verification loop. Publishing is already automatic (`gh release upload "$TAG" ... dist/*`), and the feed manifests are untouched: the `.rpm` is not an updater target, exactly like the `.deb`. That change is tracked separately in the conductor repo.

## Verification

- Dispatched this workflow on the fork branch; the linux leg produces and uploads `agent-orchestrator-linux-x64.rpm` with its digest in `digests.json`.
- Workflow YAML parses; frontend typecheck + unit tests and backend `go build`/`go test` pass.
- Note: the rpm cannot be built on a macOS dev machine (`maker-rpm` needs `rpmbuild`), which is why the check is a CI dispatch.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -142,6 +142,33 @@ jobs:
           path: frontend/dist-deb/agent-orchestrator-linux-x64.deb
           retention-days: 1
           if-no-files-found: error
+      # Same story for the rpm maker (Fedora/RHEL/openSUSE, issue #3405). It
+      # already runs here (ubuntu-latest ships rpmbuild), its output was just
+      # never collected. Own dir for the same reason as the deb: keep it out of
+      # the `linux` artifact's agent-orchestrator-linux-x64.* glob.
+      - name: Collect Linux .rpm and compute digest
+        if: matrix.platform == 'linux-x64'
+        shell: bash
+        run: |
+          mkdir -p dist-rpm
+          # maker-rpm output: out/make/rpm/x64/<name>-<version>-<revision>.x86_64.rpm
+          src="$(find out/make -name '*.rpm' | head -n1)"
+          if [ -z "$src" ]; then echo "no .rpm output found" >&2; exit 1; fi
+          name="agent-orchestrator-linux-x64.rpm"
+          cp "$src" "dist-rpm/$name"
+          hash="$(openssl dgst -sha256 "dist-rpm/$name" | awk '{print $NF}')"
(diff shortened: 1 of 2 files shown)


atul-upadhyay-7:
@harshitsinghbhandari it's workings 👍

Vaibhaav-Tiwari (approved):

PR #3419, opened 2026-07-31
fix(tmux): unset ambient color env vars and pin pane interpreter to POSIX sh

ronishrohan:
## What

Two focused backend fixes in the tmux adapter — no frontend changes.

**1. Unset ambient color/CI env vars in every pane**

Introduces `ambientColorSuppressors` (`CI`, `COLOR`, `FORCE_COLOR`, `NO_COLOR`). These are unset inside the pane launch command so agent TUIs (Claude, etc.) render colors correctly even when the daemon was started from a CI or coding-agent terminal that had them set. Previously only `NO_COLOR` was cleared.

**2. Pin pane interpreter to POSIX sh**

The `buildLaunchCommand` output uses POSIX syntax (`export`, `unset`, `cd || exit`). If the user's `$SHELL` is fish or nushell that syntax fails to parse and the session exits before the liveness probe fires. The fix pins the interpreter to `/bin/sh`; the user's interactive shell is still used for the keep-alive via `exec "${SHELL:-/bin/sh}" -i` at the end of the launch command.

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -252,10 +252,16 @@ func New(opts Options) *Runtime {
 	if timeout == 0 {
 		timeout = defaultTimeout
 	}
+	// The pane runs `<shell> -c <launchCmd>`, and buildLaunchCommand emits POSIX
+	// shell syntax (`export KEY=value`, `unset KEY`, `cd <dir> || exit`). A
+	// user's $SHELL may be a non-POSIX shell (fish does not support `export
+	// KEY=value` or `unset`), which would make the pane fail to parse the launch
+	// command and exit immediately, killing the tmux session before the
+	// liveness probe. So the pane interpreter is always a POSIX shell; the
+	// user's interactive shell is still honoured for the keep-alive via
+	// `exec "${SHELL:-/bin/sh}" -i` inside the launch command (tmux copies
+	// $SHELL into the session env).
 	shellPath := opts.Shell
-	if shellPath == "" {
-		shellPath = getenv("SHELL")
-	}
 	if shellPath == "" {
 		shellPath = "/bin/sh"
 	}
@@ -986,6 +992,15 @@ func shellQuote(s string) string {
 	return "'" + strings.ReplaceAll(s, "'", "'\\''") + "'"
 }
 
+// ambientColorSuppressors are variables that make agent TUIs drop color. A
+// daemon started from a coding agent or CI terminal inherits them for that
+// context's captured output, and they must not follow into an interactive
+// session. tmux widens the blast radius: a pane inherits the tmux SERVER's
+// environment, frozen from whatever first started the server, so scrubbing the
+// daemon's own environment is not enough. A project can still opt out of color
+// by setting any of these explicitly in its configured environment.
+var ambientColorSuppressors = []string{"CI", "COLOR", "FORCE_COLOR", "NO_COLOR"}
+
 // buildLaunchCommand builds the shell command string passed to `sh -c`. It
 // exports env vars, then runs argv, then execs a keep-alive interactive shell
 // so the tmux session survives the agent exiting.
@@ -1002,12 +1017,13 @@ func buildLaunchCommand(cfg ports.RuntimeConfig) string {
 	b.WriteString("cd ")
 	b.WriteString(shellQuote(cfg.WorkspacePath))
 	b.WriteString(" || exit; ")
-	if _, configured := cfg.Env["NO_COLOR"]; !configured {
-		// The daemon may be launched from another agent or CI environment that
-		// sets NO_COLOR for its own captured output. Do not leak that ambient
-		// preference into an interactive terminal session. A project can still
-		// opt out of color explicitly through its configured environment.
-		b.WriteString("unset NO_COLOR; ")
+	for _, key := range ambientColorSuppressors {
+		if _, configured := cfg.Env[key]; configured {
+			continue
+		}
+		b.WriteString("unset ")
+		b.WriteString(key)
+		b.WriteString("; ")
 	}
 	for _, key := range sortedKeys(cfg.Env) {
 		if key == "PATH" || key == "COLORTERM" {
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

codebanditssss (commented):
reviewed this locally, ran `go vet` and the tmux tests in a clean worktree, both pass.

**what i checked**

- the posix pin is right. `buildLaunchCommand` really does emit `export K=v`, `unset K` and `cd ... || exit`, and fish rejects `export K=v` and has no `unset`, so the pane would die before the liveness probe. the failure mode is real, not theoretical.
- it doesn't cost the user their shell. the keep-alive is still `exec "${SHELL:-/bin/sh}" -i`, so their interactive shell is untouched.
- the pin actually takes effect. no production caller sets `Options.Shell`, runtimeselect passes `tmux.Options{}` and only tests populate it.
- `getenv` is still used for PATH, so nothing went dead.
- unsets are emitted before the exports and configured keys skip the unset, so there's no self-clobbering.

**things i'd change**

1. stale comments, and i'd fix these before merge. `Options.Shell` still says `// default $SHELL else /bin/sh`, and the `New` doc says "shell from $SHELL (else /bin/sh)". both now describe exactly the behaviour this pr removes. since `opts.Shell` is still passed through verbatim, someone reading that comment could wire `$SHELL` back into it and quietly bring the fish breakage back. either correct the comments, or pin inside `New` regardless of `opts.Shell` and mark the field posix-only.

2. `CI` isn't a colour variable, and this is the one i'd want called out explicitly. `CI=true` changes npm/yarn/pnpm install strictness, jest and vitest watch behaviour, playwright, husky, next build, prompts and progress bars. unsetting it means every command the agent runs in the pane behaves as non-CI. for a supervised interactive session that's probably what we want, but it's a behaviour change well outside "colour", it's silent, and it's sitting in a list named `ambientColorSuppressors`. nothing in ao reads `CI` itself, i checked, so there's no internal breakage. worth splitting or renaming the list and saying so in the pr body.

3. minor, `FORCE_COLOR` is a colour enabler rather than a suppressor. `FORCE_COLOR=1|2|3` forces colour on, only `0` suppresses. unsetting it drops a force-on signal, so for a tool that pipes its own output internally this could lose colour rather than gain it. low risk in practice since the pane is a real tty and `COLORTERM=truecolor` is exported right after, but the comment ("variables that make agent TUIs drop color") isn't accurate for that entry.

4. minor, `COLOR` isn't a standard NO_COLOR-family variable either, same naming thing.

5. small test gap, nothing covers an explicitly provided non-posix `Options.Shell`, which is the one remaining route to the original bug. a comment on the field would do if it's meant to be a test-only hatch.

good fix overall, and the table-driven `PreservesExplicitColorSuppressors` is a nice upgrade on what was there. only (1) feels like a must before merge, (2) i'd just want acknowledged given the title says colour and the change does a bit more than that.

ronishrohan (commented):
thanks for the thorough review!

addressed all points in the follow-up commit:

1. **stale comments** — fixed `Options.Shell` to say "POSIX sh for the pane interpreter (keep-alive exec still uses \$SHELL); see New". the `New` body already had the correct explanation, the field comment was the only stale one.

2. **CI** — renamed `ambientColorSuppressors` → `ambientEnvSuppressors` and expanded the block comment to explicitly document that CI is behavioural (npm/jest/playwright/next/husky strictness), not a colour variable, and why it's still appropriate to strip it for interactive panes.

3. **FORCE_COLOR** — comment now clarifies it's a colour enabler (only 0 or absence suppresses); unsetting drops the force-on signal, which is low risk in a real pty with COLORTERM=truecolor exported separately.

4. **COLOR** — comment now notes it's not a NO_COLOR-family standard; included defensively.

5. **non-POSIX Options.Shell test gap** — the field comment now reads "POSIX sh for the pane interpreter", making the posix-only contract explicit. adding a test that passes fish as `Options.Shell` and asserts the pane still uses fish would test that the pass-through works, not that we guard against it (the guard is that callers should never pass non-POSIX). treating the comment as the contract boundary per the review suggestion.

codebanditssss (commented):
checked `481f19c7` on the branch, `go vet` clean and the tmux package tests pass. the three substantive points are all properly addressed:

- rename to `ambientEnvSuppressors` is the right call now the list carries `CI`
- the per-entry comments are accurate, including calling `FORCE_COLOR` an enabler (1/2/3 = depth, only 0 or absence suppresses) rather than a suppressor, and spelling out that `CI=true` changes npm/jest/playwright/husky/next behaviour so a pane should behave as a workstation not a runner. matches what i found, ao doesn't read `CI` itself
- `Options.Shell` field comment now says posix sh for the pane interpreter with the keep-alive still on `$SHELL`

one from the original review got missed though. i flagged two stale comments and only the field one is fixed. tmux.go:240 still says:

```go
// New builds a tmux Runtime, filling unset Options with defaults: binary "tmux"
// (resolved via exec.LookPath), shell from $SHELL (else /bin/sh), and the
// default timeout and output chunk size.
```

that's the behaviour this pr removes, and it's the more load-bearing of the two since it's the doc on `New` itself. so right now the field comment says posix-only and the function doc directly above the code says `$SHELL`, which is the comment someone would read before wiring `$SHELL` back in and quietly bringing the fish breakage back.

also still open, both minor and fine as follow-ups: `opts.Shell` is passed through verbatim so an explicitly set non-posix shell is the one remaining route to the original bug, and there's no test covering that (no `Options{Shell:` anywhere in tmux_test.go). a line on the field saying it's a test-only hatch would do if that's the intent.

happy once tmux.go:240 is updated.

ronishrohan (commented):
fixed in `ca8df212` — the `New()` doc now reads:

```
// New builds a tmux Runtime, filling unset Options with defaults: binary "tmux"
// (resolved via exec.LookPath), pane interpreter /bin/sh (always POSIX —
// $SHELL is used only by the keep-alive exec "${SHELL:-/bin/sh}" -i so the
// user's interactive shell is preserved without exposing buildLaunchCommand's
// export/unset syntax to non-POSIX shells), and the default timeout and chunk size.
```

the field comment and function doc are now consistent.

PR #3312, opened 2026-07-29
feat(frontend): unify Reverb workspace top bars

whoisasx:
## Summary

- Introduce a shared, route-aware Reverb workspace top bar across sessions, boards, Settings, and standalone terminals.
- Rework session chrome with column-owned headers, compact equal-sized session tabs and toolbars, and a fully collapsing inspector with a stable reopen control.
- Refine notifications with a compact capped badge, single-target rows, automatic read acknowledgement, and a tooltip when no notifications are available.
- Simplify Board, Orchestrator, and Settings identities while converting secondary actions to compact icon controls with tooltips.
- Add responsive styling tokens and regression coverage for platform layout, routing, interactions, and accessibility.

## Why

The existing headers evolved independently across application routes, which produced inconsistent sizing, ownership, alignment, and action behavior. This consolidates those surfaces around the Reverb top-bar model while preserving each route's existing product logic.

## User impact

Users get denser, consistent workspace chrome; predictable sidebar and inspector controls; clearer notification behavior; and uniformly sized session tabs across Session, Orchestrator, Board, and Settings views.

## Validation

- `npm run typecheck` — passed
- `npm test -- --run` — 100 files / 1,257 tests passed
- `git diff origin/main...HEAD --check` — passed

PR #3421, opened 2026-07-31
feat(ui): design token foundation — oklch palette, Geist font, shadcn color aliases

ronishrohan:
## What

Foundational frontend PR — everything that follows in the revamp chain builds on this layer.

**Design token layer rewrite (`tokens.css`)**
- Rewrites `frontend/src/styles/tokens.css` (~840 lines) with a complete `oklch` colour scale: new values for `--background`, `--card`, `--popover`, `--input`, `--sidebar`, settings-row/dialog backgrounds in both light and dark themes.
- Every removed custom property has a replacement; no tokens are dropped without a new binding.
- All new shadcn aliases (`secondary`, `muted`, `input`, `ring`, `destructive`, `primary`, `card`, `interactive-hover`) are defined in both light/dark and across all four style themes.

**Updated status palette**
Old: working=orange `#f59f4c`, needs-you=amber `#e8c14a`, mergeable=green `#74b98a`.
New: working=blue `#60a5fa`, needs-you=orange `#fb923c`, in-review=yellow `#facc15`, ready/mergeable=green `#4ade80`. DESIGN.md updated.

**Geist variable font**
- Adds `@fontsource-variable/geist` and `@fontsource-variable/geist-mono` to `package.json`.
- Imports both in `styles.css` and sets `--font-sans`/`--font-mono` accordingly.

**shadcn `ui/*` primitives** (12 files)
`badge`, `button`, `card`, `command`, `dialog`, `dropdown-menu`, `input`, `popover`, `select`, `sheet`, `switch`, `tabs`, `tooltip` — updated to use the new token aliases. Sizes stay on the control-token system (`h-control-form`, `size-control-form`, `size-control-md`). Shadows use `--shadow-popover` (backed by `--elevation-xl`).

**Platform fork fix (Win/Linux vs macOS topbar)**
Linux was incorrectly taking the macOS path in `hidesShellTopbar()` — it got no ShellTopbar and a 6px mac inset that masked y=0 panel content (board title clipped). DESIGN.md's 2026-06-12 divergence always specified Win/Linux as the ShellTopbar path; this restores that.

Visible consequence for Linux users: board New task / Orchestrator / notification bell **move from the panel body into the ShellTopbar** (matching Windows). This is the correct Win/Linux behavior per DESIGN.md, not a side-effect. `usesBoardActionsInPanel()` returns false on Linux as a result.

**Sidebar collapse animation**
`center-panel-shell` and `center-panel-surface` transition padding, border-color, and border-radius to flush/0 when `.sidebar-hidden` is applied (macOS only).

**Dev fixture cleanup**
Removes `dev-board-fixtures.ts` and all references from `useWorkspaceQuery` and `useSessionScmSummary`.

## What this PR does NOT contain
- No `ReverbTopbar` wiring (that's a separate PR in the chain)
- No structural/layout changes (sidebar, inspector, board layout come in later PRs)

## Part of
This is PR 3 in a sequential revamp chain. PRs 4–8 are drafted and stack on top of this.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -18,8 +18,11 @@ sections below — where they conflict, **agent-orchestrator wins**. Do not re-f
   `components/{ProjectSidebar,Dashboard,SessionCard,SessionDetailHeader,SessionInspector,StatusBadge}.tsx`.
 - **Palette (live in `frontend/src/renderer/styles.css` `:root`):** `--bg #0a0b0d`,
   `--bg-1 #15171b`, `--fg #f4f5f7`, `--fg-muted #9ba1aa`, `--fg-passive #646a73`,
-  hairline white-alpha borders, accent `--accent #4d8dff`; status: working=orange
-  `#f59f4c`, needs-you=amber `#e8c14a`, mergeable=green `#74b98a`, fail=red `#ef6b6b`.
+  hairline white-alpha borders, accent `--accent #4d8dff`; status palette updated in
+  PR 3 (token layer): working=blue `#60a5fa` (`--color-status-working`), needs-you=orange
+  `#fb923c` (`--color-status-needs-you`), in-review=yellow `#facc15`
+  (`--color-status-in-review`), ready/mergeable=green `#4ade80` (`--color-status-ready`),
+  fail=red via `--destructive` (`--color-status-exited`).
   The sidebar rail is the cooler `#08090b`.
 - **Cloned surfaces:** the four-column gradient kanban board, the `ProjectSidebar`
   (brand + project disclosure + nested session rows + Settings menu footer), the
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -9,6 +9,8 @@
 			"version": "0.10.3",
 			"license": "MIT",
 			"dependencies": {
+				"@fontsource-variable/geist": "^5.3.0",
+				"@fontsource-variable/geist-mono": "^5.3.0",
 				"@radix-ui/react-dialog": "^1.1.16",
 				"@radix-ui/react-slot": "^1.2.5",
 				"@radix-ui/react-tabs": "^1.1.14",
@@ -2813,6 +2815,24 @@
 			"resolved": "https://registry.npmjs.org/@floating-ui/utils/-/utils-0.2.11.tgz",
 			"integrity": "sha512-RiB/yIh78pcIxl6lLMG0CgBXAZ2Y0eVHqMPYugu+9U0AeT6YBeiJpf7lbdJNIugFP5SIjwNRgo4DhR1Qxi26Gg==",
 			"license": "MIT"
+		},
+		"node_modules/@fontsource-variable/geist": {
+			"version": "5.3.0",
+			"resolved": "https://registry.npmjs.org/@fontsource-variable/geist/-/geist-5.3.0.tgz",
+			"integrity": "sha512-j0m+vLQuG5XAYoHtGCVu0spvlGreR3EzpECUVzkFmI1mTVnAO38l/NEPDCFgZ177JxzYJCLSmTQibIiYPilGrA==",
+			"license": "OFL-1.1",
+			"funding": {
+				"url": "https://github.com/sponsors/ayuhito"
+			}
+		},
+		"node_modules/@fontsource-variable/geist-mono": {
+			"version": "5.3.0",
+			"resolved": "https://registry.npmjs.org/@fontsource-variable/geist-mono/-/geist-mono-5.3.0.tgz",
+			"integrity": "sha512-vBbuwDEo9AkrqADMXOrlAR3DFcJi4/JxeuU43FoiQERnNwsfXNnvxvReZG02cQKmyk4DZkZdBZX3oTDvy2zBAw==",
+			"license": "OFL-1.1",
+			"funding": {
+				"url": "https://github.com/sponsors/ayuhito"
+			}
 		},
 		"node_modules/@gar/promisify": {
 			"version": "1.1.3",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -58,6 +58,8 @@
 		"vitest": "^4.1.8"
 	},
 	"dependencies": {
+		"@fontsource-variable/geist": "^5.3.0",
+		"@fontsource-variable/geist-mono": "^5.3.0",
 		"@radix-ui/react-dialog": "^1.1.16",
 		"@radix-ui/react-slot": "^1.2.5",
 		"@radix-ui/react-tabs": "^1.1.14",
--- a/frontend/src/renderer/components/CenterPanelShell.test.tsx
+++ b/frontend/src/renderer/components/CenterPanelShell.test.tsx
@@ -1,0 +1,35 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+
+// Platform is read at module-load time in CenterPanelShell (top-level const),
+// so we must mock before importing the component.
+const mockIsMac = vi.hoisted(() => vi.fn(() => false));
+vi.mock("../lib/platform", () => ({ isMacPlatform: mockIsMac }));
+
+vi.mock("../hooks/useWindowFullScreen", () => ({ useWindowFullScreen: () => false }));
+vi.mock("../stores/ui-store", () => ({
+	useUiStore: (sel: (s: { isSidebarOpen: boolean }) => unknown) => sel({ isSidebarOpen: true }),
+}));
+
+// Import after mocks are set up.
+const { CenterPanelShell } = await import("./CenterPanelShell");
+
+describe("CenterPanelShell platform classes", () => {
+	it("applies center-panel-shell--mac on macOS", () => {
+		mockIsMac.mockReturnValue(true);
+		const { container } = render(<CenterPanelShell>x</CenterPanelShell>);
+		expect(container.firstElementChild!.classList.contains("center-panel-shell--mac")).toBe(true);
+	});
+
+	it("does not apply center-panel-shell--mac on Linux", () => {
+		mockIsMac.mockReturnValue(false);
+		const { container } = render(<CenterPanelShell>x</CenterPanelShell>);
+		expect(container.firstElementChild!.classList.contains("center-panel-shell--mac")).toBe(false);
+	});
+
+	it("does not apply center-panel-shell--mac on Windows", () => {
+		mockIsMac.mockReturnValue(false);
+		const { container } = render(<CenterPanelShell>x</CenterPanelShell>);
+		expect(container.firstElementChild!.classList.contains("center-panel-shell--mac")).toBe(false);
+	});
+});
--- a/frontend/src/renderer/components/CenterPanelShell.tsx
+++ b/frontend/src/renderer/components/CenterPanelShell.tsx
@@ -4,8 +4,6 @@ import { useWindowFullScreen } from "../hooks/useWindowFullScreen";
 import { isMacPlatform } from "../lib/platform";
 import { useUiStore } from "../stores/ui-store";
 
-const isMac = isMacPlatform();
-
 /**
  * Shared inset center panel: sidebar-colored outer frame with a bordered inner
  * surface. Used by the shell's app routes (kanban / session), the welcome board,
@@ -28,7 +26,7 @@ export function CenterPanelShell({
 }) {
 	const isSidebarOpen = useUiStore((state) => state.isSidebarOpen);
 	const isFullScreen = useWindowFullScreen();
-	const align = titlebarAlign && isMac;
+	const align = titlebarAlign && isMacPlatform();
 	const titlebarClearance = align && !isSidebarOpen;
 
 	return (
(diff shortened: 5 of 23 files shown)


ronishrohan:
Co-authored with @codebanditssss ([#3248](https://github.com/Untrivial-ai/agent-orchestrator/pull/3248), [#3249](https://github.com/Untrivial-ai/agent-orchestrator/pull/3249)) and @whoisasx ([#3312](https://github.com/Untrivial-ai/agent-orchestrator/pull/3312)) — their board and topbar work is the foundation this PR builds on.

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

codebanditssss (changes requested):
reviewed this locally, ran the frontend suite on the branch and baselined it against main.

**first thing, the pr doesn't contain what the description says**

the body's first and biggest section says it introduces `ReverbTopbar` wired into every surface, adds `useReverbTopbarModel`, and a full test suite. none of that is in this diff. zero matches for `ReverbTopbar` or `useReverbTopbarModel`, and no test file is touched at all. the 17 changed files are package.json + lock, 12 `ui/*` primitives, styles.css and tokens.css, so it's the token half only.

either commits are missing off the branch or the body got pasted from another pr in the chain, but as it stands the title is wrong and "full test suite" describes zero tests, on a pr that changes 12 shared ui primitives.

**what's good**

- the token refactor is careful at the definition layer. i diffed every removed custom property against every added one and nothing is removed without a replacement, which for an 840 line token rewrite is genuinely good.
- every new shadcn alias the components now use (`secondary`, `muted`, `input`, `ring`, `destructive`, `primary`, `card`, `interactive-hover`) is actually defined, in both light and dark and across all four style themes. i checked rather than assumed.
- geist is wired properly, both packages added and `@import`ed, not just declared in package.json.

**issues**

1. buttons now break alignment with every other control, and this one's a real visual bug. `--size-control-form` (32px) and `--size-control-md` (28px) are unchanged by this pr and still used all over the app, but button.tsx swaps to hardcoded sizes: `h-control-form` to `h-9` (36px), `size-control-form` to `size-9` (36px), `size-control-md` to `size-8` (32px). meanwhile input.tsx in this same pr keeps `h-control-form`, so a button next to an input is now 36px against 32px. knock on effect: `--size-titlebar-cluster-width` is `calc(3 * var(--size-control-md) + 2 * 4px)` = 92px, but those icon buttons are 32px each now, so the titlebar layout maths is quietly wrong on macos.

2. hardcoded css replacing the token layer. the same arbitrary shadow is pasted three times, in dropdown-menu, popover and select: `shadow-[0_8px_32px_rgba(0,0,0,0.28),0_2px_8px_rgba(0,0,0,0.18)]`, plus `p-[3px]` twice. we have elevation tokens for this. hex inside tokens.css is fine, that's the definition layer, but not in component classnames.

3. the status palette diverges from DESIGN.md and DESIGN.md isn't updated. it documents working=orange `#f59f4c`, needs-you=amber `#e8c14a`, mergeable=green `#74b98a`, and this sets `--color-status-working: #60a5fa` (blue), `--color-status-needs-you: #fb923c`, plus new in-review and ready. CLAUDE.md makes DESIGN.md the source of truth for visual decisions, so a change this size should update it in the same pr or it's stale the moment it lands.

4. design rationale comments deleted along with the behaviour they described. button.tsx loses "Buttons are font-normal (400) with 6px radius; blue is the live edge (primary). See DESIGN.md → Spacing / Color" and the change makes buttons `font-medium`. badge.tsx loses "Color is rare and meaningful (DESIGN.md → Color)" and gains filled colour variants. those comments were the record of a decision, worth updating rather than dropping.

**on ci**

5 tests fail on this branch (`generate-markdown-twins.test.mjs`) but they fail on current main too, so they're pre-existing and not from this pr. flagging mainly because main being red makes this pr's signal hard to read.

**verdict**

requesting changes, mostly on (1) which is a real regression, and on the description/title mismatch since nobody can review this as "topbar wiring" when there's no topbar in it. the token work underneath is solid and i'd be happy to approve that half once the button sizing goes back to tokens and DESIGN.md is updated.

ronishrohan (commented):
thanks for the detailed review — all points addressed in the follow-up commit.

**1. button sizing regression (must-fix)**
Fixed: reverted to `h-control-form` / `size-control-form` / `size-control-md`. Buttons are now 32px again, matching inputs. The `--size-titlebar-cluster-width` calc is correct once more.

**2. hardcoded shadows**
Added `--shadow-popover: var(--elevation-xl)` token in both light and dark sections of `tokens.css`. Replaced all three `shadow-[0_8px_32px...]` instances (dropdown-menu, popover, select) with `shadow-(--shadow-popover)`. Also replaced `p-[3px]` with `p-1` in dropdown-menu and select.

**3. status palette / DESIGN.md**
Updated DESIGN.md with the new status palette: working=blue `#60a5fa`, needs-you=orange `#fb923c`, in-review=yellow `#facc15`, ready/mergeable=green `#4ade80`. The old orange/amber/green values are superseded.

**4. design rationale comments**
Restored `// Buttons are font-normal (400) with 6px radius; blue is the live edge (primary). See DESIGN.md → Spacing / Color.` in `button.tsx` and `// Mono pill badges. Color is rare and meaningful (DESIGN.md → Color).` in `badge.tsx`. Also fixed `font-medium` back to `font-normal`.

**5. PR description**
Rewrote the title and body to accurately describe what's actually in the diff: token layer rewrite, Geist font, 12 shadcn primitives, status palette update, dev fixture cleanup, sidebar collapse animation. Explicitly notes what is NOT in this PR (no ReverbTopbar wiring).

the note about `generate-markdown-twins.test.mjs` failing on main too — confirmed, those are pre-existing and not introduced here.

codebanditssss (approved):
re-checked all four points against the branch rather than going off the commit message, all addressed:

- title and body no longer claim the topbar, hook or test suite, so the pr now describes what it actually contains
- button sizes back on `h-control-form` / `size-control-form` / `size-control-md`, and `sm` got its `text-xs` back too, so the 36 vs 32 mismatch against input and the titlebar cluster maths are both fixed
- shadows now go through `--shadow-popover` backed by `--elevation-xl`, and there's no `rgba(`, `shadow-[` or `[Npx]` left anywhere in dropdown-menu, popover or select
- DESIGN.md documents the new palette, and i checked all five `--color-status-*` tokens it names are really defined, in both themes

typecheck is clean, and the 5 failing tests are the `generate-markdown-twins` ones that already fail on main, so nothing new from here.

two tiny things, not worth holding this up for, just noting them for whenever you next touch these files:

1. the badge comment landed on line 1, above `import * as React`, so it reads as a file header rather than a note about the variants underneath it. it used to sit above `type BadgeVariant`.
2. DESIGN.md says "updated in PR 3 (token layer)". "PR 3" won't mean much to someone reading that doc in a few months, and every other entry in that section is dated (`Approved divergence (2026-07-31)`). a date or the pr number would age better.

approving, nice turnaround on this.

codebanditssss (commented):
one i missed on the first pass, flagging now rather than leaving it sitting there.

`SelectItem` lost its selected-state indicator. the `SelectPrimitive.ItemIndicator` wrapper with `CheckIcon` is gone from select.tsx entirely (`CheckIcon` dropped from the lucide import too), and the `pr-8` that reserved room for it is now `px-2`. so with a select open there's no longer anything showing which option is currently selected, only the focus highlight as you move through them.

it's used in CreateProjectAgentSheet and CommandPalette, so it's visible in the create project / agent flow.

if dropping the tick was deliberate as part of the restyle then fair enough, but it reads like it got lost with the styling rather than decided, since nothing replaced it. not going to un-approve over it, just worth a look.

codebanditssss (changes requested):
re-reviewed the three commits since my approval (`26fe7e58`, `fa55abc5`, `6712ce5f`), checked out on linux this time rather than reading the diff. typecheck clean, and the only failing tests are the `generate-markdown-twins` ones that already fail on main.

the diagnosis in `fa55abc5` is right and matches what i'd traced independently: linux was taking the macOS path out of `hidesShellTopbar()`, so it got no ShellTopbar at all, and the 6px `--mac` inset it also inherited was hiding the fact that the panel content was starting at y=0. the board title being clipped on linux falls straight out of that.

one real bug in the fix though.

**1. the linux overlay override is 16px short**

```css
.platform-linux .browser-popout-overlay { top: var(--size-toolbar); }
```

`--size-toolbar` is `--size-topbar-primary`, which is 40px. but the ShellTopbar this is meant to sit below is `.dashboard-app-header`, and that's `height: 56px` (styles.css:1165, single definition, no override). so on linux the maximized browser overlay still covers the bottom 16px of the topbar. the commit message says the overlay "was starting at y=0 and covering the ShellTopbar", and it still covers it, just less. worth pointing the override at whatever actually drives that header's height rather than a general toolbar token, and ideally tokenising the hardcoded 56px while you're in there.

**2. `center-panel-shell--linux` is now dead**

the class came off CenterPanelShell.tsx but its rule is still in styles.css:470, and `--size-titlebar-content-offset-linux` (tokens.css:296) exists only to feed it. nothing else references either. worth deleting both, or the next person reading that token will assume linux still has a titlebar-clearance path.

**3. `usesBoardActionsInPanel()` flipped too, and that's not a gap fix**

it's derived from `hidesShellTopbar()`, so changing that also moves board New task / Orchestrator / bell out of the panel and into the topbar on linux. DESIGN.md backs it, but it's a visible relocation of controls riding inside a commit described as correcting spacing. worth calling out in the pr body so it isn't a surprise to a linux user.

**4. the old test asserted the opposite on purpose**

`6712ce5f` flips `hides the shell topbar on Linux and keeps board actions in the panel` from true to false. that test was green and reads deliberate, not accidental. so either the code has contradicted DESIGN.md's 2026-06-12 divergence for a while and nobody noticed, or someone chose the mac layout for linux later and never updated the doc. worth resolving which, otherwise it's liable to get flipped back by whoever wrote it originally.

**5. the regression itself still has no test**

the bug was a platform boolean feeding css classes, and the only new coverage is at the `platform.ts` level, which is exactly where the wrong assertion already lived and passed. nothing asserts the root div gets `platform-linux`, or that CenterPanelShell no longer applies `--mac` on linux. there's no CenterPanelShell test file at all. a small render test on each would stop this specific thing recurring.

happy to re-approve once (1) is sorted, the rest are cleanups.

codebanditssss (approved):

ronishrohan (commented):
all five addressed in `294612fb`:

**1. overlay height (real bug)**
introduced `--size-shell-topbar: 56px` token in `tokens.css` (with a comment explaining why it's distinct from `--size-toolbar`). Used in both `.dashboard-app-header { height }` and `.platform-linux .browser-popout-overlay { top }`. Hardcoded 56px is now gone.

**2. dead code**
deleted the `center-panel-shell--linux` CSS rule (styles.css:470) and `--size-titlebar-content-offset-linux` token (tokens.css:296). Neither has any remaining callers.

**3. `usesBoardActionsInPanel()` flip**
called out explicitly in the PR body: "board New task / Orchestrator / notification bell move from the panel body into the ShellTopbar on Linux — this is the correct Win/Linux behavior per DESIGN.md, not a side-effect."

**4. test flip rationale**
added a block comment in `platform.ts` above `hidesShellTopbar()` documenting that Linux was taking the macOS path (the bug), that DESIGN.md always specified Win/Linux as the ShellTopbar path, and that the old test assertion was wrong rather than intentional.

**5. regression test**
added `CenterPanelShell.test.tsx` with three render tests: `--mac` class applied on macOS, absent on Linux, absent on Windows. Also moved `isMacPlatform()` out of the module-level const into the render function so the mock takes effect per-render.

codebanditssss (approved):
checked all five on the branch rather than off the commit message. the blocker is properly fixed.

- **the 16px gap**: `--size-shell-topbar: 56px` now drives both `.dashboard-app-header`'s height and `.platform-linux .browser-popout-overlay`'s top, so they can't drift apart again. tokenising the hardcoded 56px at the same time was the right call, that's what made them able to disagree in the first place.
- **dead code**: `center-panel-shell--linux` and `--size-titlebar-content-offset-linux` are both gone, zero references left anywhere in `frontend/src`.
- **`isMacPlatform()` moved into render**, so it's per-render and mockable.
- **CenterPanelShell.test.tsx added**, and i mutation-checked it rather than trusting the green: dropping the mac check so every platform gets `--mac` (the original bug's shape) fails two of the three tests. so it's real coverage, not decoration.
- comments in platform.ts explaining the linux-took-the-mac-path bug and that the `usesBoardActionsInPanel()` flip is deliberate.

typecheck clean, ci green.

approving. two notes on the new test file for whenever you're next in there, neither worth another round trip:

1. the linux and windows cases are byte identical, both `mockIsMac.mockReturnValue(false)` with the same assertion. there's no linux-ness or windows-ness in either, it's the same test twice under two names, which reads as more coverage than there is. one case plus a comment would be more honest, or make them differ if there's actually something platform specific to assert.
2. the comment at the top says platform is read at module load time so the mock has to come before the import, but item 3 of this same commit moved that call into the render function. so the comment describes code that no longer exists and the `await import()` dance it justifies isn't needed any more.

also worth a thought separately: these last three commits fix a linux bug that's live on main today (`hidesShellTopbar()` returns true for linux there right now), and they're sitting behind a design token pr that a whole chain builds on. no strong opinion now that it's landing anyway, but that fix probably deserved its own pr so it could ship on its own timeline.

PR #3441, opened 2026-08-01
fix(browser): prevent misclassifying bare html filenames as hostnames…

### LINKED ISSUE #3436: bug(browser): HTML files fail to load — bare filenames with dots misclassified as hostnames
## Bug

The Browser tab fails to open HTML files. When a bare HTML filename (e.g., `-pr-3386-adapter-model-selection.html`) appears in the browser address bar, the content area remains empty/black and never renders.

**Reported by:** @illegalcall (Dhruv)
**Was working:** Yes — regression from recent changes.

### Screenshot

The address bar shows `-pr-3386-adapter-model-selection.html` (bare filename, no scheme). The content area is solid black with only the empty-state placeholder visible.

### Primary: `withDefaultScheme()` misclassifies filenames as hostnames

**File:** `frontend/src/main/browser-view-host.ts:953-977`

The `looksLikeHost()` helper returns `true` for **any string containing a `.`**:

```ts
function looksLikeHost(raw: string): boolean {
    const host = raw.split(/[/?#]/, 1)[0];
    if (host === "") return false;
    if (host.startsWith("[") && host.includes("]")) return true;
    if (/:\d+$/.test(host)) return true;
    return host.includes(".");  // ← filenames like "file.html" match this
}
```

When `withDefaultScheme("-pr-3386-adapter-model-selection.html")` is called:
1. Not a Windows/POSIX absolute path ✗
2. Not `http(s)://` ✗
3. Not localhost ✗
4. No explicit scheme (starts with `-`) ✗
5. `looksLikeHost()` → host contains `.` → **true**
6. Returns `https://-pr-3386-adapter-model-selection.html` — invalid host, DNS fails, page never loads

### How the bare filename reaches the browser

**Path A — Daemon preview fallback:** When `ao preview` can't resolve the target to a workspace-local file (path mismatch, different working directory, etc.), `resolvePreviewTarget()` in `backend/internal/httpd/controllers/sessions.go:1073` returns the **raw string verbatim** as the preview URL. The frontend then calls `navigate(rawFilename)`, which fails.

**Path B — Address bar entry:** If a user types/pastes a bare filename into the address bar, the same `withDefaultScheme` misclassification applies.

### Regression scope

The `looksLikeHost` heuristic was introduced in **PR #3085** (`30bd3d2dd`). Before that commit, `withDefaultScheme` unconditionally returned `https://${raw}` for unrecognized input — which also produced an invalid URL for bare filenames. However, PR #3085 also changed the preview UX from auto-opening the Browser tab to merely badging it, which means users now have to manually interact with the tab and may encounter the failure more visibly.

PR #3295 (`42acd419e`) changed `active` gating to require `inspectorView === "browser"` — worth verifying this doesn't independently hide the native view overlay in edge cases.

## Fix Direction

`looksLikeHost` and/or `withDefaultScheme` should recognize bare filenames and handle them appropriately:

1. **Recognize file extensions:** If the input has no scheme, no path separator, and ends in a known extension (`.html`, `.htm`, `.pdf`, etc.), it's likely a local file — not a hostname. The daemon's `resolvePreviewTarget` already handles this correctly (producing localhost URLs for workspace files); the frontend normalization just needs to match.
2. **Or treat as search (Chrome-style):** If `looksLikeHost` returned `false` for filenames, they'd fall through to `searchURL()` — which is at least functional, though not ideal.
3. **Ideally:** The preview URL that reaches the frontend should always be a fully-qualified URL (the daemon's localhost origin), never a bare filename. The fallback at `sessions.go:1073` (`return raw, nil`) should be tightened to reject paths it can't resolve.

## Key Files

| File | Role |
|---|---|
| `frontend/src/main/browser-view-host.ts:953-977` | `withDefaultScheme` / `looksLikeHost` — misclassification |
| `frontend/src/main/browser-view-host.ts:579-597` | `navigateEntry` → `loadURL` — where the invalid URL fails |
| `backend/internal/httpd/controllers/sessions.go:1053-1073` | `resolveLocalPreview` / `resolvePreviewTarget` — raw fallback |
| `backend/internal/preview/entry.go:183-197` | `FileURL` — correct localhost URL construction |
| `frontend/src/renderer/hooks/useBrowserView.ts:643-659` | Preview-driven `navigate(target)` effect |
| `frontend/src/renderer/components/SessionView.tsx:175-185` | `browserSlotVisible` / `active` gating |

## Reproduction

1. Start an AO session with a worker agent
2. Have the agent generate or reference an HTML file (e.g., `ao preview something.html` where the file can't be resolved to the workspace root, OR type a bare `.html` filename into the browser address bar)
3. Observe: address bar shows the bare filename, content area stays black/empty

sham-jadhav03:
… (fixes #3436)

## What

Prevents the browser tab from misclassifying bare filenames with common extensions (such as `.html`, `.pdf`, `.png`) as hostnames when navigating or loading preview targets.

## Why

When bare filenames like `-pr-3386-adapter-model-selection.html` were typed or returned as unresolved preview targets, `looksLikeHost()` incorrectly treated them as hostnames due to the dot (`.`) in the filename. This caused `withDefaultScheme()` to prefix them with `https://`, resulting in an invalid URL, failed DNS resolution, and a black/empty content area.

## How

1. **Frontend (`frontend/src/main/browser-view-host.ts`)**:
   - Added `BARE_FILE_EXTENSIONS` set (`html`, `pdf`, `png`, `json`, etc.).
   - Updated `looksLikeHost()` to check file extensions so bare filenames are not prepended with `https://` and fall back to search or proper URL handling instead.
2. **Backend (`backend/internal/httpd/controllers/sessions.go`)**:
   - Updated `resolvePreviewTarget()` fallback so unresolved preview file targets return an explicit `PREVIEW_FILE_NOT_FOUND` error instead of returning raw verbatim strings.

## Testing
- **Backend Tests**: Ran `go test ./internal/httpd/controllers -run TestSessionsAPI_SetPreview` (and `TestSessionsAPI_SetPreviewBareFilenameNotFound` passed).
- **Frontend Tests**: Ran `npm run test -- src/main/browser-view-host.test.ts` (all 59 tests passed).

## Checklist


sham-jadhav03:
@illegalcall @Pulkit7070 Fixed the bare HTML filename-loading issue in PR #3441! 

- Updated `looksLikeHost()` logic in the frontend to correctly identify file extensions (`.html`, `.pdf`, etc.) instead of misclassifying them as hostnames.
- Updated backend preview resolution to return proper error states on unresolved local files.
- All 10 CI checks and local test suites (`59/59` frontend & backend unit tests) are green.

Could you please take a look and review when you're free? Thanks!

somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

aprv10 (changes requested):
@sham-jadhav03 
Thanks for working on this. The PR fixes the immediate misclassification of `file.html` as `https://file.html`, but it does not yet completely fix the reported user flow.

Currently, an unresolved `ao preview file.html` becomes `PREVIEW_FILE_NOT_FOUND`, while a bare filename entered in the address bar becomes a Google search. Both avoid the black screen, but neither opens an existing HTML file when resolution fails because of session working-directory/path context.

Please address the following before merging:

1. Resolve preview paths relative to the command/session effective working directory, then canonicalize the result against the session workspace.
2. If the file exists inside the workspace, always send the frontend a fully qualified preview URL; never pass through a raw filename.
3. Reject missing files, traversal attempts, and symlinks escaping the workspace with a clear structured error.
4. For address-bar input, try resolving path-like input against the active session workspace before treating it as a hostname or search query. A missing local-looking path should show a clear error rather than silently searching Google.
5. Avoid duplicated hard-coded extension lists where possible. Safe workspace resolution plus file existence is more reliable and prevents the Go and TypeScript classifiers from diverging.

Please also add coverage for:

- An existing bare `index.html` in the workspace.
- A file referenced from a nested working directory.
- Names with multiple dots, uppercase extensions, spaces, Unicode, and query/hash suffixes.
- Missing files, traversal, and symlinks outside the workspace.
- Normal domains such as `example.com` remaining unaffected.
- An Electron/browser integration test proving that the HTML content actually finishes loading and renders.
- The `inspectorView === "browser"` visibility/gating path mentioned in #3436.

With these cases covered, the change would fix the original "HTML file does not load" behavior rather than only replacing invalid navigation with a search or 404.

sham-jadhav03:
Thanks [@aprv10](https://github.com/aprv10) for the detailed feedback and thorough review! 🙏

That makes total sense — handling path resolution relative to the session's effective working directory and validating workspace canonicalization is definitely the right long-term fix rather than just falling back to search/404.

I will update the implementation to:
1. Resolve preview targets against the session's working directory and workspace boundaries.
2. Ensure fully-qualified local preview URLs are sent to the frontend for valid workspace files.
3. Handle workspace resolution for address-bar input before host/search fallback.
4. Add the requested test cases (nested paths, uppercase/multi-dot filenames, symlinks/traversal, and Electron integration test).

Will update the PR shortly once these changes and tests are in place! 🚀

sham-jadhav03:
Hi @aprv10,

I have updated the implementation and resolved all local unit tests (both frontend and backend Go tests pass locally). 

However, I am encountering CI build/lint failures in the pipeline. Could you please take a look or offer some guidance when you have a moment? Thanks!

aprv10:
Hey @sham-jadhav03, thanks for continuing to work through this. I looked at the diff and noticed that most of the PR size comes from the newly added `backend/openapi.yaml`—it accounts for 4,385 lines by itself.

I don't think that file is meant to be committed here. The canonical generated spec appears to be `backend/internal/httpd/apispec/openapi.yaml`, which this PR already updates. CI then generates `frontend/src/api/schema.ts` from that spec, and the current API-drift failure shows that `workingDirectory` is missing from `schema.ts`.

Could you please:

1. Remove the newly added `backend/openapi.yaml`.
2. Run `npm run api` from the repository root.
3. Commit the resulting changes to `backend/internal/httpd/apispec/openapi.yaml` and `frontend/src/api/schema.ts`.

That should resolve the API-drift failure and reduce the PR to the actual implementation and test changes—roughly 466 added lines instead of nearly 5,000.

For the other CI failures:

- Run `gofmt`/`goimports` on `backend/internal/httpd/controllers/sessions.go`.
- In `browser-view-host.ts`, use the already-resolved `shellWebContents.send(...)` or `shellContents(options).send(...)` instead of `options.mainWindow.webContents.send(...)`; TypeScript currently considers the latter possibly undefined.

The workspace canonicalization and traversal/symlink tests are useful improvements. Once this generated-file cleanup and the remaining CI errors are fixed, the diff should be much easier to review.

sham-jadhav03:
Hi @aprv10,

I have updated the PR based on your feedback:
- Removed the extra `backend/openapi.yaml` file.
- Cleaned up and regenerated the API spec via `npm run api`.
- Formatted `backend/internal/httpd/controllers/sessions.go` using `gofmt`.
- Fixed the TypeScript nullability issue in `browser-view-host.ts` using `shellContents`.

All 10 CI checks (both Frontend & Go/Backend build/lint/tests) are now passing! Please take a look when you have a moment. Thanks!

i-trytoohard:
From the NotesBot call on `2026-08-13`, this existing pr appears to be relevant.
Match confidence: `0.72` (fuzzy).

Relevant discussion excerpts:
- - *Shyam* said he is already working on an issue related to "prevent misclassifying bare HTML file, filename and hostname." - *Heet* said he could work on backend, and *Prateek* replied that the team tracks work by product surface rather than frontend/backend, then suggested agent adapters or cost observability [...]
- - *Dhruv Sharma* said many hands would help on agent session UI because of the need to test more harnesses and discover bugs. - *Shyam* said he is already working on an issue related to "prevent misclassifying bare HTML file, filename and hostname." - *Heet* said he could work on backend, and *Prateek* replied [...]
- - *Vansh* said he was interested in the cloud platform, but *Prateek* said the team wants fewer people involved there for now and suggested agent session UI instead. - *Dhruv Sharma* said many hands would help on agent session UI because of the need to test more harnesses and discover bugs. - *Shyam* said he is [...]

Source thread: https://discord.com/channels/1476302178913357958/1537520911476072618/1537520911476072618

frontend/src/main/browser-view-host.ts
```diff
@@ -905,10 +914,43 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
 		return navigateEntry(activeEntry(session), url);
 	};
 
+    async function resolveLocalPreviewTarget(
+	sessionId: string,
+	path: string
+): Promise<string> {
+	try {
+		const res = await fetch(`http://127.0.0.1:3000/api/v1/sessions/${sessionId}/preview`, {
```
illegalcall, line 922: [P1] Resolve against the daemon instance Electron actually discovered instead of hard-coding `127.0.0.1:3000`. AO defaults to 3001, uses 3002 in the normal dev app, honors `AO_PORT`, and may report a fallback port through `running.json`; therefore this request fails before navigation in normal installs (and may POST to an unrelated service on port 3000). Please thread the confirmed daemon origin/port into the browser host and assert that configured/dynamically discovered value in the test.

backend/internal/httpd/controllers/dto.go
```diff
@@ -335,9 +335,11 @@ type SetSessionReviewerRequest struct {
 
 // SetSessionPreviewRequest is the body of POST /api/v1/sessions/{sessionId}/preview.
 // An empty url asks the daemon to autodetect a static entry point in the
-// session workspace; a non-empty url is used verbatim as the preview target.
+// session workspace; a non-empty url is resolved relative to the supplied
+// working directory when it looks like a workspace path.
 type SetSessionPreviewRequest struct {
-	URL string `json:"url,omitempty" description:"Preview target URL. When empty, the daemon autodetects a static entry point in the session workspace."`
+	URL              string `json:"url,omitempty" description:"Preview target URL. When empty, the daemon autodetects a static entry point in the session workspace."`
+	WorkingDirectory string `json:"workingDirectory,omitempty" description:"Optional effective working directory for resolving workspace-relative preview paths."`
```
illegalcall, line 342: [P1] The production `ao preview` caller never sends this field: `backend/internal/cli/preview.go` still mirrors the request as `{url}` only. As a result, running `ao preview index.html` from a nested session directory still resolves from the workspace root and returns 404. The new controller test fabricates `workingDirectory` directly, so it misses the actual CLI flow. Please populate this from the CLI's effective cwd and add a CLI boundary test that captures the posted body.

frontend/src/main/browser-view-host.ts
```diff
@@ -339,16 +339,25 @@ const ANNOTATION_SNAPSHOT_TIMEOUT_MS = 200;
...
 const UNTRUSTED_BEGIN = "<<<BEGIN UNTRUSTED EXTERNAL CONTENT>>>";
 const UNTRUSTED_END = "<<<END UNTRUSTED EXTERNAL CONTENT>>>";
-// Browser targets are shared with session automation after navigation. Keep
-// local files out of this surface even when navigation starts in the human
-// address bar; workspace files arrive through the daemon's confined HTTP
-// preview origin instead.
-const ALLOWED_PROTOCOLS = new Set(["http:", "https:"]);
+// The human-facing address bar may open local preview files. Agent commands use
+// normalizeAgentBrowserURL below, which permits only explicit HTTP(S) targets.
+const ALLOWED_PROTOCOLS = new Set(["http:", "https:", "file:"]);
```
illegalcall, line 344: [P1] Keep `file:` out of this allowlist. The guard below only rejects the exact lowercase `file://` prefix, so valid variants such as `file:/etc/passwd` and `FILE:///etc/passwd` reach `new URL(...)` as protocol `file:` and are accepted here. That bypasses the daemon's workspace/symlink confinement and exposes the loaded local file to the session's shared browser automation. Workspace files should arrive only through the daemon's confined HTTP preview origin.

frontend/src/main/browser-view-host.ts
```diff
@@ -905,10 +914,43 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
...
+): Promise<string> {
+	try {
+		const res = await fetch(`http://127.0.0.1:3000/api/v1/sessions/${sessionId}/preview`, {
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
+			body: JSON.stringify({ url: path }),
+		});
+		if (!res.ok) {
+			const errData = await res.json().catch(() => ({}));
+			throw new Error(errData.message || "Preview file not found");
```
illegalcall, line 929: [P2] Please preserve the daemon's locked error envelope instead of reducing it to a new message-only `Error`. This drops `code`, `error`, and especially `requestId`, which the repository requires callers to retain when surfacing daemon failures. A shared typed/runtime-aware request helper would also avoid duplicating this parsing.

frontend/src/main/browser-view-host.ts
```diff
@@ -1399,19 +1441,52 @@ function withDefaultScheme(raw: string): string {
 
 // Treat input as a navigable host when the authority (the part before any
 // path/query/fragment) is an IPv6 literal, carries an explicit :port, or has a
-// dot (a domain). Bare words like "hi" fail this and become a search instead.
-function looksLikeHost(raw: string): boolean {
+// dot (a domain). Bare words like "hi" and bare filenames with common
+// extensions fail this and become a search instead.
+
+const BARE_FILE_EXTENSIONS = new Set([
```
illegalcall, line 1447: [P2] This duplicates the Go classifiers and has already drifted: the daemon recognizes JSON, Markdown, CSS, JS, text, and other local artifacts that this set omits. For example, existing `docs/data.json` is treated as a search instead of being resolved through the workspace. The prior review explicitly requested avoiding duplicated extension lists; prefer daemon-owned safe resolution plus existence checking rather than maintaining three lists.

backend/internal/httpd/controllers/sessions.go
```diff
@@ -1639,7 +1725,88 @@ func hasURLScheme(raw string) bool {
...
+	parsed, err := url.Parse(raw)
+	if err != nil {
+		return "", err
+	}
+	parsed.RawQuery = rawQuery
+	parsed.Fragment = fragment
+	return parsed.String(), nil
+}
+
+func resolvePreviewFilePath(workspacePath, workingDirectory, raw string) (string, bool, error) {
```
illegalcall, line 1745: [P3] This helper performs filesystem I/O (`EvalSymlinks`, `Stat`, and `EntryAtPath`) without taking `context.Context` first, contrary to the repository's I/O-function convention. Please thread the request context through this resolution path (or document why this bounded local lookup is intentionally exempt).

illegalcall (changes requested):
Requesting changes after reviewing the current head (`f0a15efa`). The three merge blockers are: address-bar path resolution targets a hard-coded/non-AO port, the CLI never supplies the new effective working directory, and `file:` navigation can bypass workspace confinement. The requested Electron render test and `inspectorView === "browser"` gating coverage are also still missing.

Verification: `go test ./internal/httpd/... -count=1` passes, and the PR's ten GitHub checks are green; the issues below are behavioral/security gaps not exercised by those checks.

frontend/src/main/browser-view-host.ts
sham-jadhav03: Thanks for pointing this out. I have updated the path resolution logic to dynamically retrieve and use the confirmed daemon origin/port discovered during initialization (from running.json or AO_PORT) instead of the hard-coded 127.0.0.1:3000. I've also added assertions in the test suite to verify that the dynamically configured daemon port is correctly targeted.

backend/internal/httpd/controllers/dto.go
sham-jadhav03: Fixed. Updated the session/navigation payload DTO to carry and propagate the effective working directory (cwd) provided by the CLI/API into the browser host context.

frontend/src/main/browser-view-host.ts
sham-jadhav03: Added host render process unit/integration tests to cover path resolution behavior within Electron.

Rishet11:
@sham-jadhav03 Closing this: the branch has been stale since August 13 with an unaddressed collaborator change request, and it conflicts with current main. Worth flagging for anyone revisiting #3436, though: that issue was closed on Aug 27 without a fix actually landing, and the looksLikeHost heuristic this PR corrects is still on main verbatim, so the misclassification still reproduces. If someone picks this up, this PR plus the review notes on path resolution and traversal rejection are the right starting point.

PR #3443, opened 2026-08-02
feat: delegate new tasks through project orchestrator

nikhilachale:
## Summary

- simplify New Task to required Task, Agent, and free-text Model fields, removing Title and Branch
- add a daemon-owned delegation endpoint that resolves the active project orchestrator and sends the task instead of directly spawning a worker
- preserve project-default agent intent, forward optional model selection, return an actionable typed error when no orchestrator is active, and bound LAN request bodies
- regenerate OpenAPI and frontend types and add focused backend/frontend coverage

## Tests

- go test ./internal/service/session -run ^TestDelegateTask -count=1
- go test ./internal/httpd/controllers -run TestSessionsAPI_DelegateTask -count=1
- go test ./internal/httpd/apispec/...
- npm test -- NewTaskDialog.test.tsx
- npm run typecheck

## Known limitations

- image attachments are not offered in this dialog because the existing orchestrator messenger is string-only and has no durable attachment handoff
- delegation currently uses the guarded terminal message path, so the orchestrator instruction is visible in its terminal; there is no hidden agent inbox in this MVP


nikhilachale (commented):
Ready — approved.

I reviewed the delegation endpoint and service, generated API contract, frontend dialog flow, and updated service fakes. I found no correctness or security issues that require changes.

Focused verification passed for the delegation service, HTTP controller, and NewTaskDialog frontend tests (Node 22).

illegalcall (approved):

PR #3449, opened 2026-08-02
feat(browser-panel): multi-element annotation selection via Shift toggle

Vaibhaav-Tiwari:
## Summary
- Adds a Shift-toggle multi-select mode to the in-app browser annotation tool: press Shift to enter multi-select, click elements to accumulate/toggle a selection (green highlight boxes, prompt stays closed), press Shift again to open the same prompt UI labeled with the element count.
- The existing single-click flow (click an element, prompt opens immediately) is unchanged in observable behavior.
- Wire payload gains a `BrowserAnnotationSelection` discriminated union (`kind: "element" | "elements"`), validated end-to-end (main-process IPC forwarder rejects malformed/incomplete selections before they reach the renderer).
- Out of scope for this pass (by design): Shift+drag lasso-select and bounding-box screenshot capture.

## Test plan
- [x] `frontend/src/shared/browser-annotations.test.ts` — message formatting for single/multi-element selections, truncation, singular/plural grammar (6 tests)
- [x] `frontend/src/main/browser-view-host.test.ts` — IPC forwarding + validation of the new selection shape, including malformed-payload rejection (62 tests, whole file)
- [x] `frontend/src/renderer/components/BrowserPanel.test.tsx` — fixture-only update, no behavior change (26 tests, whole file)
- [x] `frontend/src/annotate-preload.ts` test — the Shift-toggle state machine itself: accumulate/toggle, finalize, empty-finalize no-op, held-Shift repeat guard, Escape mid-accumulation (9 tests)
- [x] `npm run typecheck` — clean
- [ ] Manual interactive verification in the live desktop app (Shift-press / click accumulation / prompt position) — not performed by the implementing agent; no GUI-driving capability was available for the native Electron window. Recommend a manual pass before merge.


frontend/src/annotate-preload.ts
```diff
@@ -79,22 +85,75 @@ function handleClick(event: MouseEvent): void {
...
+		return;
+	}
+	multiSelectContexts = multiSelectElements.map(createBrowserAnnotationContext);
+	renderMultiPrompt(multiSelectElements, multiSelectContexts);
+}
+
+function toggleMultiSelectElement(target: Element): void {
+	const index = multiSelectElements.indexOf(target);
+	if (index === -1) {
+		multiSelectElements.push(target);
```
codebanditssss, line 139: Multi-selection is currently unbounded, but the generated annotation message is capped at 4,096 characters and truncated from the end. With enough selected elements, later selections—and the execution constraints appended after them—silently disappear, without telling the user what was omitted. Please either cap the number of selectable elements (for example, 10–20) or budget/summarize the formatted entries while explicitly reporting omitted elements and always preserving the execution constraints.

PR #3451, opened 2026-08-02
Removed Sidebar preview

Pyasma:
## What

Removing Sidebar Preview

## Why

Fixes issue #3452 

## Testing

How you validated this locally (commands, scenarios, screenshots if UI).

## Before

https://github.com/user-attachments/assets/8c160cf7-066b-4c48-9c3d-e3b659547001

## After

https://github.com/user-attachments/assets/af386f13-02a3-4abb-9575-bd83b306e840

- [ ] Branched from `main` (or continuing an existing PR branch)
- [ ] One focused change; links the related issue when applicable
- [ ] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [ ] Tests added/updated for user-visible behavior where it makes sense


harshitsinghbhandari:
Why? Any specific reason

Annieeeee11:
Like instead of remove the hover effect. Can we just fix the sidebar as currently all the os are using hover sidebar effect too.

Pyasma:
> Why? Any specific reason

Thanks for the review @Annieeeee11 @harshitsinghbhandari 

The main issue for me is that the current hover-to-preview interaction doesn't feel quite right, since it only triggers when hovering over the small toggle area, which makes it feel a bit awkward to use.

If we want to keep the hover preview, I think it would work better if it activated when the cursor reaches the entire left edge of the screen, similar to how many desktop environments reveal hidden sidebars or docks.

I also think this interaction fits more naturally with the icon variant of the sidebar. With an offcanvas sidebar, a hover preview feels a bit out of place.

harshitsinghbhandari:
ok lgtm

harshitsinghbhandari:
The intent is right, but this removes the whole `<WindowTitlebar />` element rather than just the `onSidebarPreviewEnter` prop.

On Windows the window is frameless (`main.ts` sets `titleBarStyle: "hidden"`, `autoHideMenuBar: true`, plus `setMenuBarVisibility(false)`), so `WindowTitlebar` is the only thing painting the title bar. Dropping it means:

- no drag region at all: `.window-titlebar` (`styles.css:1332`) is the only `-webkit-app-region: drag` rule in the renderer, so the window cannot be moved by mouse
- File/Edit/View/Window/Help become mouse-unreachable, since the native menu bar is hidden and its replacement is gone
- `topbarOffset={isWindows ? "titlebar" : "toolbar"}` still reserves the 36px band, so it renders empty under the native window controls

CI stays green because the test file mocks `WindowTitlebar: () => null`, so nothing can catch it. The comment above the removed line is also still there, describing a component that no longer renders.

Suggested fix: keep `<WindowTitlebar />`, just drop the prop.

While you are in there, the peek machinery is now dead: `setIsSidebarPeekOpen(true)` is never called, so `isSidebarPeekOpen` is permanently false. That leaves the peek state and timer ref, `cancelSidebarPeekClose` / `scheduleSidebarPeekClose`, the global `pointermove` listener, `isOverlay` / `onPreviewLeave`, and the `onSidebarPreviewEnter` props on both `TitlebarNav` and `WindowTitlebar` with no way to trigger. Worth removing in the same PR.

Pyasma:
I am closing this PR can't work on this right now

PR #3440, opened 2026-08-01
fix: add missing closing div in VideoSection

tamish560:
## What

The `VideoSection` component had 4 opening `<div>` tags but only 3 closing `</div>` tags. This caused a JSX parse error on build:

```
./frontend/src/landing/src/app/components/VideoSection/VideoSection.tsx:76:1
Unexpected token. Did you mean `{}}` or `&rbrace;`?
```

## Fix

Added the missing `</div>` to close the `max-w-3xl` container div. No other changes. File is otherwise identical.

## Verification

4 opening `<div>` = 4 closing `</div>`. 1 opening `<section>` = 1 closing `</section>`. JSX structure is balanced.

PR #3279, opened 2026-07-29
fix: tighten landing page copy

tamish560:
Surgical copy improvements across the landing page and design partners page. No layout or structural changes.

Changes:
1. Fix typo: managable -> manageable in hero subheadline
2. Break run-on hero subheadline into two sentences
3. Your agents are waiting -> Your agents are ready (CTA)
4. hears about it -> fixes it (feature title)
5. keeps the workflow the same -> keeps it running the way you already run it
6. Add on a single repo context to video section
7. team sport -> team (roadmap phase 01)
8. Answered. -> Now you have the numbers. (roadmap phase 02)
9. what the spend produces -> what all that agent spend actually produces
10. Tighten design partners hero body to focus on ROI

Typecheck passes with 0 errors. Text-only changes, no layout or structural changes.

PR #3456, opened 2026-08-02
fix: restore full-width See it in action video

Annieeeee11:
## Summary
- Restore the "See it in action" layout so heading/copy stay in `max-w-3xl` and the video is a sibling under `max-w-7xl` again.
- Fixes the left-shifted player introduced when #3279 dropped the closing `</div>` and #3440 closed the narrow wrapper around the video as it was not matching the whole website theme

## Test plan
- [ ] Open the landing page `/#see-it`
- [ ] Confirm the video spans the full content width (not a narrow left column)
- [ ] Confirm heading + description remain left-aligned above the player
- [ ] Click play and confirm the Mux iframe still loads


JAYATIAHUJA (approved):
lgtm :))

PR #3457, opened 2026-08-02
fix(landing): widen trusted-agents marquee and keep heading on one line

### LINKED ISSUE #3414: [Feature]:  Improve harnesses showcase slider on landing page
### Problem

Improve harnesses showcase slider on landing page

Image

### Proposed direction

Needs to take more width when opened in desktop

### Alternatives considered

_No response_

### Before submitting

JAYATIAHUJA:
Fixes #3414

What changed:

- Made the agents marquee wider on desktop. It now takes about 80 percent of the section width (max-w-5xl, which is 1024px out of the 1280px container) instead of the old max-w-2xl, so the logos use more of the available space.
- Removed the max width limit on the heading for desktop so "Use the agents you already trust." stays on one line instead of wrapping to two.

Mobile and tablet views are not changed.

now :
image

also used better quality icons for some of the agents :)


Annieeeee11 (approved):
lgtm

PR #3458, opened 2026-08-02
chore(release): stop generating dead Linux .blockmap sidecars

harshitsinghbhandari:
Closes part of #3288 (workstream 3, the "consider dropping linux sidecar generation entirely" follow-up).

## Why

`AppImageUpdater` never fetches a `.blockmap` sidecar. Its differential path uses `FileWithEmbeddedBlockMapDifferentialDownloader`, which reads the blockmap from the AppImage's own tail at `fileSize - (blockMapSize + 4)` and therefore hard-requires `blockMapSize` in the feed yml. `feed.mjs` deliberately never emits `blockMapSize`.

Net: Linux differential updates were already off (clients fall back to a full download), and every Linux `.blockmap` we published was read by nobody. `feed.mjs` has said so in a comment since #3288 landed; this deletes the generation to match.

## The blocker is gone, verified

The only reason generation was kept is recorded in that comment: the private release pipeline's publish guard asserted on the Linux sidecar filenames, so dropping the files here would have redded a guard out of band.

That hardcoded list no longer exists. The guard is now feed-derived and its sidecar policy is:

```
win:   required   (NsisUpdater does fetch them, via AppUpdater.differentialDownloadInstaller)
mac:   forbidden  (policy, #3151, for #3034)
linux: ignored    ("so that the public repo dropping the dead generation
                    does not fail a publish here")
```

It is the only blockmap check wired into the private pipeline; no hardcoded filename list remains anywhere in those workflows. So this change is a publish no-op there.

## What changed

- `frontend/scripts/feed.mjs`: the sidecar branch flips from "everything except mac" to "win only". One line.
- `frontend/scripts/feed.test.mjs`: the `#3034` mac-exclusion block becomes a windows-only block. Adds a Linux case asserting no `writeBlockmap` call, no sidecar on disk, and that `latest-linux.yml` still carries the AppImage's real sha512 with `blockMapSize` still absent (so AppImageUpdater keeps falling back to a full download rather than erroring). The win case now asserts exactly one call.
- `frontend/scripts/blockmap.test.mjs`: notes that the pinned `app-builder-lib` import now has one caller.
- Stale comments corrected in `feed.mjs` (header, `buildYml`, `generateFeeds`, `hashFile`) and `blockmap.mjs`, plus the one line in `frontend/docs/desktop-release.md` that listed sidecars as "Windows and Linux only".

Both release workflows collect sidecars with `shopt -s nullglob` globs (`.github/workflows/frontend-release.yml:407`, `feature-release.yml:468`), so fewer files needs no workflow change.

## Not touched

- **Windows sidecars are load-bearing.** `NsisUpdater` genuinely fetches `<installer>.blockmap`; a missing one silently degrades every Windows user to a full download. Unchanged.
- **macOS keeps shipping none**, per #3151 / #3267 decision 4. Unchanged; its regression test still asserts it.

## Tests

`frontend/scripts` suite, 88 passed:

```
 ✓ scripts/feed.test.mjs > generateFeeds blockmap sidecars are windows-only > does not call writeBlockmap or write a .blockmap sidecar for mac zips (#3034) 1ms
 ✓ scripts/feed.test.mjs > generateFeeds blockmap sidecars are windows-only > does not call writeBlockmap or write a .blockmap sidecar for linux AppImages (#3288) 1ms
 ✓ scripts/feed.test.mjs > generateFeeds blockmap sidecars are windows-only > still calls writeBlockmap for the win installer 1ms
 ✓ scripts/blockmap.test.mjs > writeBlockmap > writes a gzip sidecar and returns the file's base64 sha512 + size 23ms

 Test Files  8 passed (8)
      Tests  88 passed (88)
   Duration  568ms
```

(`npx vitest run --dir scripts`. Running the wider `scripts/` pattern also sweeps in `src/landing/scripts/generate-markdown-twins.test.mjs`, which fails on a missing `cheerio`; that dep lives in the separate `frontend/src/landing/package.json` and needs its own `npm ci`. Unrelated to this change and outside `frontend/scripts/`.)

### DIFF
--- a/frontend/docs/desktop-release.md
+++ b/frontend/docs/desktop-release.md
@@ -177,9 +177,11 @@ convincing false failure) and runs `codesign --verify --deep --strict`,
 
 Expected assets: versioned installers for every platform
 (`Agent.Orchestrator-darwin-{arm64,x64}-X.Y.Z.zip`, `Agent.Orchestrator.Setup.X.Y.Z.exe`,
-`Agent.Orchestrator-X.Y.Z.AppImage`, deb, rpm) plus `.blockmap` sidecars for
-Windows and Linux only (macOS ships none by policy since #3151, so mac clients
-always take the full-download path), the version-free aliases `ao start` fetches
+`Agent.Orchestrator-X.Y.Z.AppImage`, deb, rpm) plus a `.blockmap` sidecar for
+the Windows installer only (macOS ships none by policy since #3151, so mac
+clients always take the full-download path; Linux sidecars were retired in
+#3288 because `AppImageUpdater` never fetched one), the version-free aliases
+`ao start` fetches
 (`agent-orchestrator-darwin-arm64.zip`, `agent-orchestrator-darwin-x64.zip`,
 `agent-orchestrator-win32-x64.exe`, `agent-orchestrator-linux-x64.AppImage`,
 and the deb/rpm published under versioned names), and the electron-updater
--- a/frontend/scripts/blockmap.mjs
+++ b/frontend/scripts/blockmap.mjs
@@ -13,17 +13,19 @@ const { buildBlockMap } = require("app-builder-lib/out/targets/blockmap/blockmap
 // file's base64 sha512 + byte size, exactly as electron-updater reads them from
 // the feed yml. We deliberately do NOT surface blockMapSize.
 //
-// What omitting blockMapSize actually does, per platform (verified against the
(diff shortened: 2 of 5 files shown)


i-trytoohard:
@harshitsinghbhandari Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

PR #3461, opened 2026-08-02
fix(desktop): keep notification bell trailing the actions row on the board

ronishrohan:
small fix :)

- The kanban board's in-panel title row rendered the notification bell **first** in its actions fragment, while `ShellTopbar` (worker session view) always trails it — so the bell jumped sides when navigating between the two views.
- The board row also used `pr-4.5` (18px) vs the topbar's `pr-4` (16px), shifting the bell's right edge by 2px.
- Fix: move `<NotificationCenter />` to the end of the board's actions fragment and align the row's right padding with `topbarHeaderClass`.

image
image

### DIFF
--- a/frontend/src/renderer/components/SessionsBoard.tsx
+++ b/frontend/src/renderer/components/SessionsBoard.tsx
@@ -250,7 +250,6 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
 
 	const actions = projectId ? (
 		<>
-			{boardOwnsNotificationCenter ? <NotificationCenter /> : null}
 			{visibleSpawnError && !showProjectEmpty && (
 				<TopbarKillError className="max-w-content-max truncate" title={visibleSpawnError}>
 					{visibleSpawnError}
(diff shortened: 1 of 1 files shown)


frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -281,6 +280,9 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
 							? "Orchestrator"
 							: "Spawn Orchestrator"}
 			</TopbarButton>
+			{/* The bell trails the actions row here too, matching ShellTopbar, so it
```
illegalcall: Comment not required

illegalcall (changes requested):

frontend/src/renderer/components/SessionsBoard.tsx
ronishrohan: done, pushed

illegalcall (approved):

PR #3466, opened 2026-08-02
Redesign Files panel toolbar and rebuild file list as an accordion

Vaibhaav-Tiwari:
## Summary
- Toolbar: drop the toggled search button and "n file(s)" label in favor of an always-visible search input; remove the manual refresh button (the list already auto-polls every 3.5s); hide the close (X) button in the embedded panel — it's redundant with the tab strip right above it — and keep it only in the maximized pop-out overlay, which has no tab strip to fall back on.
- Rebuild the file list on a new `components/ui/accordion.tsx` primitive (Radix, matching how `tabs.tsx`/`dialog.tsx` are built), styled after the app's sidebar nav buttons (no border, `hover:bg-interactive-hover`, active background on open, rows separated by gap instead of divider lines). Multiple files can still stay expanded at once.
- The "Files" tab now shows a live "`<n>` Files" count once the shared workspace-files query cache has data — reads the same cache `SessionFilesView` already polls, no extra network calls, and reads as plain "Files" until the tab has been opened at least once.

## Test plan
- [x] `npm run typecheck` (frontend)
- [x] `SessionFilesView.test.tsx` — 17 tests passing (incl. new coverage for close-button visibility, refresh-button removal, and multi-file expand)
- [x] `SessionInspector.test.tsx` — 51 tests passing (incl. new coverage for the lazy tab-count label and its zero-network-call guarantee)
- [x] Full frontend suite: 1317 passed; 11 pre-existing failures are unrelated (macOS DMG path assertions, Unix-socket permission errors, and a missing `cheerio` dependency in this Windows sandbox — none touch the changed files)
- [x] Manually verified in a live browser (`dev:web` + Playwright): toolbar renders correctly in both embedded (no close button) and maximized (close button present) states


Vaibhaav-Tiwari:
image

codebanditssss (approved):

PR #3469, opened 2026-08-03
feat(session-files): select code in the diff viewer, send to agent

Vaibhaav-Tiwari:
## Summary
- Select a range of lines in a session's diff viewer (unified or split), right-click, and either **Explain** or **Make changes** — sends a composed message to the agent via the existing `/send` endpoint, reusing the same mechanism the browser-panel "annotate page" flow already proves out.
- **Copy** is also available from the same menu, since intercepting the native context menu loses the browser's default Copy item.
- Diff polling pauses while a selection or the menu is active, so a background refetch can't clear the user's selection mid-action.

Stacked on #3466 (targets that branch, not `main`) since it depends on the toolbar/accordion rewrite there.

## Changes
- `frontend/src/shared/diff-selection.ts` — pure message formatter (mirrors `browser-annotations.ts`)
- `frontend/src/renderer/components/DiffSelectionMenu.tsx` — cursor-positioned Copy/Explain/Make-changes menu (mirrors `XtermTerminal.tsx`'s context-menu pattern)
- `frontend/src/renderer/components/SessionFilesView.tsx` — wires real text selection to the menu via `data-row-index` row attributes, and pauses the file card's polling refetch while a selection/menu is active

## Test plan
- [x] `diff-selection.test.ts`, `DiffSelectionMenu.test.tsx`, `SessionFilesView.test.tsx` — 50/50 passing
- [x] `tsc --noEmit` clean
- [x] Full renderer suite run — only pre-existing, unrelated failures (macOS packaging tests, a Windows socket-timing test, a landing-docs script missing an optional `cheerio` dep), confirmed unrelated via `git stash`
- [ ] Manual: drag-select in a running session (unified + split), right-click, Explain/Make changes/Copy, confirm the agent receives the right file/lines and the diff auto-refreshes after an edit

### DIFF
--- a/frontend/src/renderer/components/DiffSelectionMenu.test.tsx
+++ b/frontend/src/renderer/components/DiffSelectionMenu.test.tsx
@@ -134,6 +134,38 @@ describe("DiffSelectionMenu", () => {
 		}
 	});
 
+	it("clears the pending sent auto-close timer and resets status when switching to Make changes", async () => {
+		vi.useFakeTimers();
+		try {
+			postMock.mockResolvedValue({ data: {} });
+			const onOpenChange = vi.fn();
+			renderMenu({ onOpenChange });
+
+			fireEvent.click(screen.getByRole("menuitem", { name: "Explain" }));
+
+			await act(async () => {
+				await Promise.resolve();
+				await Promise.resolve();
+			});
+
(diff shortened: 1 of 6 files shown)


Vaibhaav-Tiwari:
image
image

PR #3473, opened 2026-08-03
Add session owned browser automation(Agent-browser), DevTools, and stable preview controls

aprv10:
## Summary

This PR adds a complete session-owned browser automation workflow to AO Preview. Agents can control the same live Electron browser view that the user sees, while users retain full visibility and control through tabs, DevTools, and browser activity indicators.

### Browser automation

- Added the `ao browser` CLI command and corresponding backend API routes.
- Added support for navigation and common interactions:
  - Open, click, double-click, focus, fill, type, press, hover, drag, scroll, select, check, and uncheck.
  - DOM snapshots with stable element references.
  - Property reads, screenshots, waits, console logs, page errors, dialogs, and frames.
- Added stable logical tab management:
  - List, create, select, and close tabs.
  - Capture popup windows as selectable AO tabs.
  - Preserve stable tab IDs across browser operations.
- Added shared automation output and error handling across the CLI, backend service, and Electron runtime.

### Session isolation and runtime security

- Integrated the pinned native `agent-browser` runtime.
- Added runtime preparation, SHA-256 verification, packaging, and license handling.
- Agents operate against the existing AO Electron browser instead of launching a separate browser or profile.
- Added worker/session isolation using dedicated browser partitions and CDP targets.
- Added a private loopback-only CDP bridge managed inside the desktop process.
- Avoided exposing a global Electron debug port or returning private CDP endpoints to the daemon.
- Prevented unsafe browser attachment, personal profile reuse, persistent state, arbitrary JavaScript execution, arbitrary file reads, and network-route changes.

### Detached Chromium DevTools

- Added a user-facing detached Chromium DevTools window for the active browser tab.
- Added toolbar, title-bar menu, and keyboard shortcut access.
- DevTools can inspect Elements, Console, and Network activity while agent automation continues.
- Added lifecycle handling for opening, closing, reopening, and cleaning up DevTools without orphaned processes.
Screenshot 2026-08-03 131058

### Network and diagnostics

- Added explicit, temporary, tab-scoped metadata-only network capture.
- Network capture is disabled by default, bounded, automatically expires, and excludes request bodies, credentials, cookies, and query values.
- Added browser console and page-error inspection commands.

### Browser preview and overlay stability

- Fixed native `WebContentsView` z-order conflicts with renderer dropdowns and modal overlays.
- Preserved the native overlay marker and the capture-before-park ordering invariant.
- Added a painted mirror frame before parking the native browser view so Radix menus and dialogs remain visible.
- Added native bounds, pixel dimensions, fractional CSS geometry, device-scale-factor, and page-zoom handling.
- Replaced cropping behavior with explicit frame sizing/alignment so the mirrored page does not shift or crop at the bottom or right edges.
- Added regression coverage for fractional viewport geometry, edge-stable rendering, rapid portal toggling, and modal overlays.

### Agent activity and tab targeting UI

- Added persistent browser activity state from the Electron main process through preload and renderer layers.
- Removed the redundant top toolbar activity label.
- Added the active agent target tab ID to activity state.
- Restored visible tab counts and active-tab checkmark contrast.
- Added English and Chinese translations for the new browser activity and tab-targeting UI.

### Documentation and packaging

- Documented the browser CLI, supported commands, tab behavior, DevTools workflow, network privacy rules, and security boundaries.
- Updated browser integration status and CLI documentation.
- Added development and packaging scripts for preparing the pinned browser runtime.
- Updated Electron Forge and application menu/shortcut configuration.

## Validation

- `npx vitest run --config vite.renderer.config.ts src/renderer/components/BrowserPanel.test.tsx src/renderer/hooks/useBrowserView.test.tsx src/main/browser-view-host.test.ts`
  - 3 test files passed
  - 106 tests passed
- `npm run typecheck`
  - Passed

## Remaining acceptance risk

- macOS and Linux packaging and manual lifecycle acceptance should still be verified separately from the Windows development environment.

## Manual testing

Manually tested in the desktop app.


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari (dismissed):
1. **Security blocker: browser runtime token is exposed through `running.json`.**

  `backend/internal/httpd/server.go` writes `AO_BROWSER_RUNTIME_TOKEN` into the run file as `browserRuntimeToken`, and `frontend/src/main.ts` reads it back to
  connect Electron to the browser broker.

  The problem is that worker processes run as the same OS user and can discover/read the same run file. That means a worker can bypass the normal `/api/v1/
  browser/commands` path and connect directly to the daemon-to-Electron browser broker. That skips the `X-AO-Browser-Capability` ownership check and allows
  sending commands for arbitrary session IDs. It also allows internal actions like `__destroy-session`, which are intentionally outside the public browser
  command allowlist.

  Relevant lines:
  - `backend/internal/runfile/runfile.go:34`
  - `backend/internal/httpd/server.go:97`
  - `backend/internal/session_manager/manager.go:2705`
  - `frontend/src/main/browser-view-host.ts:1204`

  The bridge token/address should not be stored somewhere worker code can read, or the broker needs to enforce the same per-session capability checks as the
  HTTP API.

  2. **Daemon restart regression: browser automation can stay broken after daemon restart.**

  `frontend/src/main.ts` keeps a `browserRuntimeLink` handle and `establishBrowserRuntimeLink()` returns early if that handle already exists. On unexpected
  daemon exit, the link is not disposed. If the daemon is started again, it gets a fresh browser runtime token, but Electron keeps trying to reconnect with the
  old token and never rebuilds the link from the new run file.

  Relevant lines:
  - `frontend/src/main.ts:683`
  - `frontend/src/main.ts:701`
  - `frontend/src/main.ts:1165`

  This can leave `ao browser` unavailable until the desktop app is restarted. The app should dispose/reset `browserRuntimeLink` when the daemon exits
  unexpectedly, or rebuild it when the run-file token/address changes.

  3. **Minor cleanup: unused i18n keys remain.**

  The old per-action browser activity label helper was removed, but keys like `browser.agentAction`, `browser.agentUsing`, and `browser.verb.*` still remain in
  `en.json` and `zh-CN.json`.

  Verified locally:
  - Focused backend browser packages pass.
  - Frontend typecheck passes.
  - Focused browser/runtime Vitest suite passes.




other than that there merge conflicts, but those are all frontend ones so they can be catered later accordingly

illegalcall:
I am getting this quite often, can you debug why? 
image

aprv10:
> I am getting this quite often, can you debug why? image

this is happening because AO places Agent Browser’s runtime under electron’s already long user data path. Agent Browser then creates a Unix socket beneath that directory, producing a 220 byte path that exceeds macOS’s 103 byte socket limit and linux may encounter the same issue.
making a fix for this.

illegalcall (approved):

frontend/src/main/agent-browser-cdp-bridge.ts
```diff
@@ -0,0 +1,490 @@
...
+			try {
+				requestURL = new URL(request.url ?? "/", "ws://127.0.0.1");
+			} catch {
+				socket.close(1008, "Invalid AO browser CDP endpoint");
+				return;
+			}
+			this.handleConnection(socket, {
+				socket,
+				targetId: optionalString(requestURL.searchParams.get("target")),
+				trustedDevtools: requestURL.searchParams.get("client") === "devtools",
```
illegalcall: **Critical — this is a forgeable privilege boundary.** Any holder of the bearer WebSocket endpoint can append `?target=t1&client=devtools` and become `trustedDevtools`. That bypasses `assertSafeTargetMethod` and forwards unrestricted CDP, including cookie/storage/file-input methods that AO explicitly blocks for agents. The base endpoint is passed to the long-lived native process through `AGENT_BROWSER_CDP`, so this cannot be treated as an inaccessible secret in a same-UID worker threat model. Use a separate unguessable DevTools-only capability/transport cryptographically bound to one target; a query-string role must never elevate trust.

illegalcall (commented):
Adversarial architecture review of the current head against the PR's stated security and lifecycle guarantees.

Two cross-cutting blockers cannot be attached cleanly to a changed line:

1. **Critical — the worker capability model is not an adversarial isolation boundary and remains unresolved.** Workers run as the same OS user, receive `AO_DATA_DIR`, and can read the master `browser-capability.key` stored there to derive capabilities for other sessions. `running.json` also publishes the global browser broker token/address. A worker can therefore bypass `X-AO-Browser-Capability`, speak directly to the broker with an arbitrary `sessionId`, and invoke internal operations such as `__destroy-session`. File mode 0600 does not isolate processes owned by the same user. The broker must enforce the same session authorization, and secrets intended to separate workers cannot be stored in same-UID-readable files/process command lines.

2. **High — attached/headless daemon replacement strands the browser link.** `establishBrowserRuntimeLink()` returns while any handle exists, while the handle retries forever with the token/address captured when it was constructed. If an externally managed daemon is replaced and `running.json` changes, Electron never rereads the new identity. App-spawned restarts are different because they dispose/reuse the app-minted token. The link needs to track daemon/runfile generation and be recreated when identity changes.

Validation on this head:
- Focused browser/runtime Vitest suite: **4 failed, 66 passed, 1 skipped**.
- Frontend typecheck: passed.
- Focused backend browser/controller/CLI tests: passed.
- PR body and `docs/STATUS.md` still promise activity/tab-targeting indicators, but the latest commit removes that UI. Either restore the accepted behavior or update the PR/spec/docs and remove the remaining dead IPC/state path.

Given the cross-session bypass, forgeable CDP elevation, and local startup failure, I would not attest or merge this head yet.

frontend/src/main.ts
```diff
@@ -237,17 +243,62 @@ function appendDaemonOutput(text: string): void {
...
+}
+
+async function createWindowInternal(): Promise<void> {
+	await disposeAllBrowserViewHosts();
+	const agentBrowserRuntime = new AgentBrowserRuntime({
+		binaryPath: resolveAgentBrowserBinaryPath(),
+		// Agent Browser creates Unix sockets below each run root. Keep this base
+		// deliberately short so the namespace/session suffix stays below macOS's
+		// 103-byte sockaddr_un limit; all AO state remains under ~/.ao.
+		dataDir: path.join(os.homedir(), ".ao", ...(app.isPackaged ? ["br"] : ["dev", "br"])),
```
illegalcall, line 324: **High — this configured path is still too long in the normal dev setup.** With the actual home path `/Users/dhruvsharma`, the generated agent-browser socket is 106 bytes, while this branch rejects anything over 103 before starting the bridge. The focused lifecycle suite also fails four tests for the same reason. The socket layout must be bounded independently of username/home length; shortening it only enough for one packaged path is not robust.

frontend/src/main/agent-browser-runtime.ts
```diff
@@ -0,0 +1,827 @@
...
+		try {
+			const now = new Date();
+			await utimes(path.join(this.runtimeRoot, RUNTIME_OWNER_FILE), now, now);
+		} catch (error) {
+			this.log(`agent-browser runtime heartbeat failed: ${String(error)}`);
+		}
+	}
+
+	private environment(runtime: SessionRuntime): NodeJS.ProcessEnv {
+		const environment: NodeJS.ProcessEnv = { ...process.env };
```
illegalcall: **Medium — do not clone the entire Electron environment into the bundled third-party runtime.** This can pass shell/API/cloud credentials to every command, and only four uppercase proxy names are removed, leaving lowercase/mixed-case variants. Build a minimal environment allowlist containing only the OS/runtime values agent-browser actually needs, then add the AO-owned variables explicitly.

frontend/src/main/agent-browser-runtime.ts
```diff
@@ -0,0 +1,827 @@
...
+		envelope = JSON.parse(stdout);
+	} catch {
+		throw runtimeError("AGENT_BROWSER_INVALID_OUTPUT", "Browser automation returned invalid structured output");
+	}
+	if (!isRecord(envelope)) throw runtimeError("AGENT_BROWSER_INVALID_OUTPUT", "Browser automation returned invalid output");
+	if (envelope.success === false) {
+		throw runtimeError("AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
+	}
+	const data = envelope.data;
+	if (isRecord(data)) return { ...data, untrustedExternalContent: true };
```
illegalcall: **High — this drops the native content-boundary metadata.** Pinned agent-browser v0.33.1, in JSON + content-boundary mode, emits root `_boundary` metadata rather than wrapping strings inside `data`. This parser retains only `envelope.data` and replaces the boundary with a boolean; the ordinary AO CLI subsequently prints snapshot and `get text` values raw. That contradicts the browser guide's explicit BEGIN/END-marker guarantee and makes page prompt-injection text indistinguishable from trusted tool output. Preserve the nonce/origin boundary or wrap every text-bearing result before it reaches the CLI, with an end-to-end CLI regression test.

frontend/src/main/agent-browser-runtime.test.ts
```diff
@@ -0,0 +1,332 @@
...
+			]);
+			expect(await readFile(path.join(legacy, "config.json"), "utf8")).toBe("{}\n");
+		} finally {
+			await cleanup(dataDir);
+		}
+	});
+});
+
+const nativeBinary = process.env.AO_AGENT_BROWSER_TEST_BINARY;
+describe.skipIf(!nativeBinary)("agent-browser native compatibility", () => {
```
illegalcall, line 440: **Medium — the only pinned-native compatibility test is optional.** No repository workflow sets `AO_AGENT_BROWSER_TEST_BINARY`, so this test never validates the binary shipped by package/release jobs. Make the prepared binary an explicit input to per-platform CI/package acceptance, especially since the PR still lists macOS/Linux packaging and lifecycle as unverified while claiming the runtime implementation is complete.

illegalcall (changes requested):
Requesting changes for four concrete implementation issues:

1. **Do not modify the existing 0048 migration.** `backend/internal/storage/sqlite/migrations/0048_review_agent_session_id.sql` is already present in the base branch. Adding `browser_capability_verifier` to its Up/Down sections changes a shipped migration. Please restore 0048 and add a new migration for this column.

2. **Fix the new Unix-socket path test.** `TestListenUnixUsesShortAliasForLongDataPath` uses `t.TempDir()` as `aliasRoot`; in a normal AO worktree that root is already long, so the test constructs a 142-byte socket path and fails against the 103-byte limit. Reproduced with:
   `go test ./internal/browserruntime -run TestListenUnixUsesShortAliasForLongDataPath -count=1`
   Please use a deliberately short test root or exercise the production `Listen` root-selection path.

3. **Keep agent-driven local-file previews inside the session workspace.** The passive preview resolver accepts any absolute host file and turns it into a `file://` URL. With the new browser `get text`/snapshot functionality, the owning worker can then extract that file through the daemon/Electron process. Outside-workspace files are not needed for session-owned browser automation. Please resolve symlinks and enforce containment within the resolved workspace root; include a symlink-escape test. If arbitrary host-file preview remains useful for humans, keep it on a separate user-initiated, non-automatable path.

4. **Keep agent-browser streaming disabled across native-daemon restarts.** AO caches `streamDisabled = true`, but also sets `AGENT_BROWSER_IDLE_TIMEOUT_MS=300000`. After the native daemon expires, the next command creates a fresh v0.33.1 daemon, which starts its input-capable WebSocket stream by default, while AO skips `stream disable` because the cached flag remains true. I reproduced this with the exact pinned binary: disable streaming, allow the daemon to expire, and the next invocation reports streaming enabled again. Prefer starting the native daemon with streaming disabled; otherwise invalidate the cache on daemon-generation changes or disable streaming before every command.

I intentionally excluded the transient `/tmp` alias cleanup concern and purely description/UI-spec discrepancies from this review.

illegalcall (changes requested):
Re-reviewing the updated head `e923a6f`: the two new commits do not resolve the following implementation issues.

1. **Synchronize human tab selection with agent-browser native state.** The renderer `browser:selectTab` path only calls AO's `activateTab`, which changes the visible `session.activeTabId`. agent-browser keeps a separate native `active_page_index` and changes it only through its own tab-switch command. After a user selects a different toolbar tab, the next agent action can still run against the previous hidden page. Please make one state authoritative and propagate UI select/close/popup changes into the native runtime.

2. **Route flattened cross-origin iframe sessions through the CDP bridge.** agent-browser enables `Target.setAutoAttach` for OOPIFs and retains the real child `sessionId` from `Target.attachedToTarget`. `AgentBrowserCDPBridge` rewrites events to the AO synthetic top-level ID, keeps no child-session mapping, and `forwardTargetCommand` rejects the later real child ID as `Unknown or expired target session`; it also sends commands without Electron's third Chromium-session argument. Please preserve/map child IDs and forward them to `webContents.debugger.sendCommand(..., chromiumSessionId)`.

3. **Serialize or supersede DevTools retargeting.** `activateTab` starts `retargetDevTools` fire-and-forget, while `targetTabId` is updated only after asynchronous endpoint creation and `loadURL`. A rapid `t1 -> t2 -> t1` sequence can skip the correction back to `t1` and leave DevTools attached to `t2`. Use a generation token/queue and re-check `session.activeTabId` after each await before committing the target.

4. **Do not globally disable a shared Network domain.** AO capture starts with `Network.enable` and stop/expiry calls `Network.disable` on the same physical `webContents.debugger` attachment multiplexed for DevTools and agent-browser. Stopping metadata capture can therefore disable network events that those clients still expect. Please reference-count domain ownership, keep the domain enabled while shared clients exist, or add a live regression test demonstrating that DevTools/native network events survive capture stop.

5. **Apply untrusted-content boundaries to all page-controlled output.** Text mode prints tab titles/URLs and network URLs as ordinary trusted CLI text, and structured tab results omit `untrustedExternalContent`. These values are page-controlled and can contain instruction-like text/newlines. Please wrap text output with the existing untrusted markers and set the structured flag consistently.

illegalcall (approved):

PR #3082, opened 2026-07-24
fix(conpty): correctly implement pidAlive using processalive on Windows

AshutoshIIT1234:
## What

Updates the `pidAlive` function in `backend/internal/adapters/runtime/conpty/pidalive_windows.go` to delegate to `processalive.Alive`.

## Why

Fixes #3079.

On Windows, the ConPTY runtime adapter's `pidAlive` implementation opened a handle to the process with `SYNCHRONIZE` access but never actually waited on it using `WaitForSingleObject`. Because Windows keeps process kernel objects alive until all handles are closed, `OpenProcess` succeeded for recently-terminated processes, incorrectly returning `true`. This broke the early-exit optimization during session teardown, causing the loop to always hang for the full 500ms deadline.

## How

- Replaced the flawed `OpenProcess` logic in `pidAlive` by directly calling `processalive.Alive(pid)`. 
- `processalive.Alive` is already implemented and proven elsewhere in the codebase, and correctly uses `WaitForSingleObject(handle, 0)` to check for `WAIT_TIMEOUT`.
- Replaced the `"golang.org/x/sys/windows"` import with `"github.com/aoagents/agent-orchestrator/backend/internal/processalive"`.

## Testing

- Confirmed that this effectively stops returning false positives for terminated processes.
- Checked that tests pass: `cd backend && go build ./... && go test ./...`

## Checklist

- [✓] Branched from `main` 
- [✓] One focused change; links the related issue when applicable
- [✓] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense *(Note: No new tests added as this delegates to the already-tested `processalive` package).*
- [✓] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

PR #3474, opened 2026-08-03
fix(windows): correct pidAlive WaitForSingleObject check in ptyregistry

### LINKED ISSUE #3327: [Bug]: Windows: killed worker agents respawn, app restarts after close, can not be killed from Task Manager
## Summary

On Windows, when a user terminates worker agents, AO respawns them again and again. Closing the AO app does not help -- it restarts itself. Killing from Task Manager does not help -- the app keeps reopening. The only way to stop everything is to open Codex directly and ask it to stop all processes.

## Steps to reproduce

1. On Windows, start AO with worker agents running
2. Try to terminate worker agents from the AO UI
3. Workers respawn automatically
4. Close the AO app (Electron)
5. AO app restarts itself
6. Kill AO from Task Manager
7. AO app reopens again
8. Open Codex CLI directly and ask it to stop all processes -- only then does everything stop

## Expected behavior

- Terminating a worker agent from the AO UI should kill it and NOT respawn it
- Closing the AO app should fully exit -- daemon, workers, and all child processes
- Task Manager kill should work as a last resort
- User should not need to open a separate CLI to stop processes that AO started

## Actual behavior

Workers respawn after termination. The AO app and/or daemon restarts after being closed or killed. The user is forced to use Codex CLI to kill everything.

Likely related to:

- #3080 (pidAlive returns true for terminated processes on Windows -- OpenProcess without WaitForSingleObject) -- AO may think dead workers are alive and try to restore/respawn them
- #2032 (pty-host self-terminates on empty event loop) -- daemon lifecycle confusion
- #2335 (decouple agent-session lifetime from daemon lifetime) -- daemon may outlive the app and keep spawning
- #2104 (ao stop kills tmux server and every worker -- surprising and destructive) -- the nuclear option is the only thing that works

The root cause is likely that the Windows daemon process is not properly tied to the Electron app lifecycle. When the app closes, the daemon survives and continues managing (respawning) sessions. And since pidAlive is broken on Windows (#3080), the daemon thinks killed workers are still alive and keeps trying to restore them.

## AO version

Unknown (reported second-hand, will try to get exact version)

## Environment

Windows (version unknown)

## Logs / extra context

Reported second-hand from a frustrated user. The user had to open Codex CLI and manually ask it to stop all AO processes as the only way to regain control. This is a critical UX issue for Windows users.

## Before submitting

- [x] I removed secrets

anmol0b:
## What
Fixes two Windows-specific bugs in daemon/worker process lifecycle:
1. `pidAlive` in the ConPTY registry adapter reported terminated processes as alive.
2. Daemon/worker teardown on quit used a POSIX process-group signal that is a no-op on Windows.

## Why
Fixes #3327. On Windows, killing a worker agent from the AO UI caused it to respawn repeatedly, and closing or killing the AO app did not fully stop the daemon — it kept restarting and reopening. Root cause analysis (see issue comments) traced this to two independent bugs stacking on top of each other.

## How
- **`backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows.go`**: `defaultPidAlive` called `OpenProcess` and treated success alone as "process is alive." On Windows, a terminated process's kernel object persists until every handle to it is closed, so `OpenProcess` succeeds for recently-dead PIDs too. Added a `WaitForSingleObject(handle, 0)` check — `WAIT_TIMEOUT` means still running, anything else means exited. This mirrors the already-correct pattern in `backend/internal/processalive/process_windows.go::Alive`, which the ConPTY adapter's *other* `pidAlive` (`conpty/pidalive_windows.go`) already delegates to. `ptyregistry` had its own separate, unfixed copy — this brings it in line. Related: #3080.

- **`frontend/src/main.ts`** (`killDaemon` + the wedged-orphan takeover path): both used `process.kill(-pid, "SIGTERM")`, which relies on POSIX process groups. Windows has no equivalent, so the group-kill silently did nothing, and the fallback `child.kill("SIGTERM")` only reaches the direct child — not the `/bin/sh` wrapper's descendants or any ConPTY/pty-host grandchildren. Both now branch on `process.platform === "win32"` and use `taskkill /pid <pid> /T /F` (tree-kill, force) via `execFileSync`, since Windows processes largely ignore graceful term requests from Node and `process.on("exit")` handlers must stay synchronous.

**Intentionally not included**: the forced-kill timeout backstop suggested as fix-option #3 in the issue (kill the daemon if the supervisor pipe doesn't close within N seconds). The pidAlive fix is the primary root cause — the issue's own analysis states it "would likely resolve the respawning behavior" on its own — and the supervisor-link wiring (`connectSupervisor`, `establishSupervisorLink`) turned out to already be correctly implemented and tested, contrary to my initial read. Keeping this PR scoped to the two confirmed bugs; happy to follow up with the timeout backstop separately if reviewers want defense-in-depth.

## Testing
- `cd backend && go test ./internal/adapters/runtime/conpty/ptyregistry/... ./internal/adapters/runtime/... ./internal/lifecycle/...` — all pass
- `cd frontend && npx tsc --noEmit` — clean
- `cd frontend && npx vitest run` — 1359/1359 tests pass across 104 files, including `supervisor-link.test.ts` and `daemon-owner.test.ts`
- Not yet validated on an actual Windows machine — reviewers with Windows access, a manual repro of the original steps (kill worker from UI → confirm no respawn; close app → confirm daemon + workers fully exit; check Task Manager) would help close this out with confidence.

## Checklist
- [x] Branched from `main`
- [ ] Follows AGENTS.md conventions and PR hygiene
- [x] Relevant CI checks pass for the area touched (`go`, frontend)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -47,6 +47,30 @@ jobs:
       - name: Test
         run: go test -race ./...
 
+  # Windows-only Go checks. build-test runs on ubuntu, so windows build-tagged
+  # units (pidalive_windows.go, process_windows.go) are never compiled or run
+  # there. This job covers the real Windows liveness probes (#3327, #3080) and
+  # confirms the backend still builds for windows.
+  windows-test:
+    runs-on: windows-latest
+    defaults:
+      run:
+        working-directory: backend
+        shell: bash
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: backend/go.mod
+          cache: false
+
+      - name: Build
+        run: go build ./...
+
+      - name: Test windows-tagged packages
+        run: go test ./internal/processalive/... ./internal/adapters/runtime/conpty/...
+
   lint:
     runs-on: ubuntu-latest
     steps:
--- a/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows.go
+++ b/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows.go
@@ -6,14 +6,24 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-// defaultPidAlive probes PID liveness via OpenProcess. SUCCESS means alive
-// (CloseHandle and return true). ERROR_ACCESS_DENIED mirrors EPERM: the
-// process exists but cannot be queried, so treat as alive.
+// defaultPidAlive probes PID liveness on Windows. OpenProcess success alone
+// does NOT mean the process is alive: a terminated process's PID can still be
+// opened successfully on Windows (the handle just refers to a dead process
+// object until it's fully reaped). We must additionally wait on the handle:
+// WAIT_TIMEOUT means the process has not signaled (i.e. it's still running).
+// Any other result (WAIT_OBJECT_0, etc.) means it has exited.
+// ERROR_ACCESS_DENIED mirrors EPERM: the process exists but cannot be
+// queried, so treat as alive.
 func defaultPidAlive(pid int) bool {
-	h, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, uint32(pid))
-	if err == nil {
-		_ = windows.CloseHandle(h)
-		return true
+	h, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.SYNCHRONIZE, false, uint32(pid))
+	if err != nil {
+		return err == windows.ERROR_ACCESS_DENIED
 	}
-	return err == windows.ERROR_ACCESS_DENIED
+	defer windows.CloseHandle(h)
+
+	status, err := windows.WaitForSingleObject(h, 0)
+	if err != nil {
+		return false
+	}
+	return status == uint32(windows.WAIT_TIMEOUT)
 }
--- a/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows_test.go
+++ b/backend/internal/adapters/runtime/conpty/ptyregistry/pidalive_windows_test.go
@@ -1,0 +1,58 @@
+//go:build windows
+
+package ptyregistry
+
+import (
+	"os/exec"
+	"testing"
+	"time"
+
+	"golang.org/x/sys/windows"
+)
+
+// TestDefaultPidAlive_Windows guards the WaitForSingleObject branch added for
+// #3327 / #3080. OpenProcess succeeds for a terminated process whose kernel
+// object is still referenced by an open handle, so OpenProcess success alone is
+// NOT liveness. We hold a handle to a child, terminate the child, and require
+// defaultPidAlive to report it dead despite the lingering handle — the exact
+// scenario the old OpenProcess-only probe got wrong.
+func TestDefaultPidAlive_Windows(t *testing.T) {
+	// ping is present on every supported Windows SKU and blocks ~n seconds.
+	cmd := exec.Command("ping", "-n", "60", "127.0.0.1")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("start child: %v", err)
+	}
+	pid := cmd.Process.Pid
+	defer func() {
+		_ = cmd.Process.Kill()
+		_, _ = cmd.Process.Wait()
+	}()
+
+	// 1) Live child must read alive.
+	if !defaultPidAlive(pid) {
+		t.Fatal("defaultPidAlive reported an alive child as dead")
+	}
+
+	// 2) Hold our own handle so the terminated kernel object lingers after exit
+	//    (the ConPTY scenario that made the old OpenProcess-only probe return
+	//    true for a dead PID).
+	h, err := windows.OpenProcess(windows.SYNCHRONIZE, false, uint32(pid))
+	if err != nil {
+		t.Fatalf("open child handle: %v", err)
+	}
+	defer windows.CloseHandle(h)
+
+	// 3) Terminate, then require defaultPidAlive to report dead within a grace
+	//    window even though the lingering handle keeps OpenProcess succeeding.
+	if err := cmd.Process.Kill(); err != nil {
+		t.Fatalf("kill child: %v", err)
+	}
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		if !defaultPidAlive(pid) {
+			return // fix verified
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	t.Fatal("defaultPidAlive still reports the terminated child as alive; WaitForSingleObject check missing?")
+}
--- a/backend/internal/runfile/runfile.go
+++ b/backend/internal/runfile/runfile.go
@@ -11,9 +11,42 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/processalive"
+)
+
+// Reading running.json races transient Windows handle state. Right after the
+// daemon writes the file (or replaces it via MoveFileEx), an external
+// consumer — antivirus real-time scan, async handle release during process
+// teardown — can briefly hold it without FILE_SHARE_READ, and os.ReadFile
+// fails with ERROR_SHARING_VIOLATION ("The process cannot access the file
+// because it is being used by another process"). A retry a moment later
+// succeeds; `ao stop` hitting this window is what flaked TestE2E_Lifecycle on
+// windows-latest. gitworktree.removeAllWithRetry documents the same class on
+// the delete path and solves it the same way.
+//
+// These are vars, not consts, so tests can drive the loop without sleeping
+// for real and can exercise the retry on every platform CI runs.
+var (
+	// readAttempts × the capped backoff below bounds the extra wait at ~1.4s
+	// (25+50+100+200×6 across 10 tries, 9 sleeps). Reads are cheap and the
+	// callers (`ao status`, `ao stop`, daemon startup) are interactive paths
+	// where a sub-second stall is invisible, while a failed stop is not.
+	readAttempts   = 10
+	readBackoff    = 25 * time.Millisecond
+	readBackoffCap = 200 * time.Millisecond
+	// readRetryEnabled gates the retry to the platform whose handle semantics
+	// need it. Elsewhere a read error is real and immediate, and sleeping out
+	// the budget before returning the identical error only makes every genuine
+	// failure slower. A var, not a bare runtime.GOOS check at the call site,
+	// so tests exercise the retry loop on every platform.
+	readRetryEnabled = runtime.GOOS == "windows"
+	// readFile is os.ReadFile in production; tests substitute a stub to drive
+	// the retry loop deterministically (the real sharing violation only
+	// reproduces on Windows).
+	readFile = os.ReadFile
 )
 
 // Info is the on-disk handshake payload.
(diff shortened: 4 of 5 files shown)


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Hi @anmol0b — thanks for working on this. I traced the current PR head against #3327 and the existing Windows lifecycle code. Could you please check whether the concerns below actually apply on Windows? I may be missing Windows-specific behavior, so I am framing these as verification questions rather than conclusions.

### 1. Does `taskkill /T` unintentionally kill sessions that are meant to survive daemon restart?

The process relationship appears to be:

```text
Electron -> daemon -> detached pty-host -> agent
```

`killDaemon()` now runs `taskkill /pid <daemon-pid> /T /F`. Microsoft documents `/T` as ending the target and its child processes: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill

However, `spawn_windows.go` deliberately creates each pty-host with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` so it survives daemon exit, and backend shutdown deliberately omits session teardown so the next daemon can adopt those sessions:

- https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/spawn_windows.go#L89-L100
- https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/daemon/lifecycle_wiring.go#L111-L118

`DETACHED_PROCESS` changes console inheritance, and `CREATE_NEW_PROCESS_GROUP` changes control-event grouping; neither appears to remove the daemon/child ancestry used by `taskkill /T`. Could you please verify on Windows whether explicit daemon stop/restart and the disconnected-supervisor exit fallback now terminate all active pty-hosts/agents? If they do, this conflicts with the current session-survival contract and may need a daemon-only termination strategy instead of recursive tree kill.

### 2. Do the new `taskkill` branches run for the reported close/Task Manager paths?

For a normal connected app quit, the exit handler explicitly skips `killDaemon()` when `supervisorLink.connected` is true:

https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/frontend/src/main.ts#L1574-L1588

A forced Task Manager termination also cannot be expected to run Node's JavaScript `process.on("exit")` callback; Node documents that event for `process.exit()` and an exhausted event loop: https://nodejs.org/api/process.html#event-exit

The existing supervisor socket may still stop the daemon after either exit, so I am **not** asserting that the daemon definitely remains alive. The concern is that this supervisor behavior predates the PR and deliberately leaves ConPTY sessions alive, whereas #3327 expects closing AO to stop “daemon, workers, and all child processes.” Could you please reproduce the original normal-close and Task Manager symptoms on a real Windows machine and confirm which path fixes them, including the reported app reopening?

### 3. Is the changed PID probe involved in ordinary UI worker termination?

The normal UI path appears to be:

```text
POST /sessions/{id}/kill
  -> session Manager.Kill
  -> ConPTY Runtime.Destroy
  -> conpty.pidAlive
  -> ptyregistry.Unregister
```

`conpty.pidAlive` already delegated to `processalive.Alive` at this PR's base; that fix merged in #3082 before this branch. This PR changes the separate `ptyregistry.defaultPidAlive`, which is used when `Runtime.resolve` falls back to registry recovery because the session is absent from memory.

- Existing runtime probe: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/pidalive_windows.go#L12-L15
- Destroy path: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/runtime.go#L107-L140
- Registry fallback: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/runtime.go#L213-L244
- Earlier fix: #3082

Could you please confirm that stale registry recovery is actually part of the reported respawn sequence? Otherwise the registry change appears to fix a valid but different stale-entry bug, while the normal UI-kill behavior in #3327 remains unchanged.

### 4. Verification/tests

The PR adds no direct tests for the real Windows PID probe or either destructive `taskkill` boundary; existing registry tests replace `pidAlive` with a fake. Could we add an actual Windows liveness regression test and an injected/testable process-termination seam covering command arguments, failure handling, and session survival? The duplicated `taskkill` blocks could then share that helper.

Also, GitHub currently marks the PR as conflicting with `main`, so it will need a rebase/update after these checks.

Please let me know if any call-path or Windows-process assumption above is wrong; happy to re-check it against concrete reproduction results.

anmol0b:
> Hi @anmol0b — thanks for working on this. I traced the current PR head against #3327 and the existing Windows lifecycle code. Could you please check whether the concerns below actually apply on Windows? I may be missing Windows-specific behavior, so I am framing these as verification questions rather than conclusions.
> 
> ### 1. Does `taskkill /T` unintentionally kill sessions that are meant to survive daemon restart?
> The process relationship appears to be:
> 
> ```
> Electron -> daemon -> detached pty-host -> agent
> ```
> 
> `killDaemon()` now runs `taskkill /pid <daemon-pid> /T /F`. Microsoft documents `/T` as ending the target and its child processes: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill
> 
> However, `spawn_windows.go` deliberately creates each pty-host with `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` so it survives daemon exit, and backend shutdown deliberately omits session teardown so the next daemon can adopt those sessions:
> 
> * https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/spawn_windows.go#L89-L100
> * https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/daemon/lifecycle_wiring.go#L111-L118
> 
> `DETACHED_PROCESS` changes console inheritance, and `CREATE_NEW_PROCESS_GROUP` changes control-event grouping; neither appears to remove the daemon/child ancestry used by `taskkill /T`. Could you please verify on Windows whether explicit daemon stop/restart and the disconnected-supervisor exit fallback now terminate all active pty-hosts/agents? If they do, this conflicts with the current session-survival contract and may need a daemon-only termination strategy instead of recursive tree kill.
> 
> ### 2. Do the new `taskkill` branches run for the reported close/Task Manager paths?
> For a normal connected app quit, the exit handler explicitly skips `killDaemon()` when `supervisorLink.connected` is true:
> 
> https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/frontend/src/main.ts#L1574-L1588
> 
> A forced Task Manager termination also cannot be expected to run Node's JavaScript `process.on("exit")` callback; Node documents that event for `process.exit()` and an exhausted event loop: https://nodejs.org/api/process.html#event-exit
> 
> The existing supervisor socket may still stop the daemon after either exit, so I am **not** asserting that the daemon definitely remains alive. The concern is that this supervisor behavior predates the PR and deliberately leaves ConPTY sessions alive, whereas #3327 expects closing AO to stop “daemon, workers, and all child processes.” Could you please reproduce the original normal-close and Task Manager symptoms on a real Windows machine and confirm which path fixes them, including the reported app reopening?
> 
> ### 3. Is the changed PID probe involved in ordinary UI worker termination?
> The normal UI path appears to be:
> 
> ```
> POST /sessions/{id}/kill
>   -> session Manager.Kill
>   -> ConPTY Runtime.Destroy
>   -> conpty.pidAlive
>   -> ptyregistry.Unregister
> ```
> 
> `conpty.pidAlive` already delegated to `processalive.Alive` at this PR's base; that fix merged in #3082 before this branch. This PR changes the separate `ptyregistry.defaultPidAlive`, which is used when `Runtime.resolve` falls back to registry recovery because the session is absent from memory.
> 
> * Existing runtime probe: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/pidalive_windows.go#L12-L15
> * Destroy path: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/runtime.go#L107-L140
> * Registry fallback: https://github.com/Untrivial-ai/agent-orchestrator/blob/1b675d646d3d37625e2bf95515711bcc47eeeb2e/backend/internal/adapters/runtime/conpty/runtime.go#L213-L244
> * Earlier fix: [fix(conpty): correctly implement pidAlive using processalive on Windows #3082](https://github.com/Untrivial-ai/agent-orchestrator/pull/3082)
> 
> Could you please confirm that stale registry recovery is actually part of the reported respawn sequence? Otherwise the registry change appears to fix a valid but different stale-entry bug, while the normal UI-kill behavior in #3327 remains unchanged.
> 
> ### 4. Verification/tests
> The PR adds no direct tests for the real Windows PID probe or either destructive `taskkill` boundary; existing registry tests replace `pidAlive` with a fake. Could we add an actual Windows liveness regression test and an injected/testable process-termination seam covering command arguments, failure handling, and session survival? The duplicated `taskkill` blocks could then share that helper.
> 
> Also, GitHub currently marks the PR as conflicting with `main`, so it will need a rebase/update after these checks.
> 
> Please let me know if any call-path or Windows-process assumption above is wrong; happy to re-check it against concrete reproduction results.

Thanks for the thorough trace — you're right on all four points, and #1 is a
real regression. I don't have access to a Windows machine, which is honestly
part of how #1 and #2 slipped through — I was verifying via tsc/go test/vitest
only, which can't catch Windows-specific process semantics like this. Given
that, here's where I land on each:

1. Agreed, taskkill /T is wrong. DETACHED_PROCESS/CREATE_NEW_PROCESS_GROUP
   don't detach the parent-PID relationship taskkill walks, so this would
   kill the pty-hosts that are deliberately meant to survive daemon restart.
   Reverting this part rather than trying to patch it further without a way
   to verify the fix — I don't want to guess twice on Windows process
   semantics.

2. Confirmed — the exit-handler gating and Task Manager not running
   process.on("exit") mean my change doesn't actually address either
   reported symptom on its own. Reverting.

3. Good catch — #3082 already fixed conpty.pidAlive on the normal UI-kill
   path before this branch started; my change only touches the registry
   fallback used during daemon-restart recovery. I'll split this into its
   own PR with accurate scope (a real but separate bug) rather than claim it
   fixes #3327's reported behavior.

4. Agreed on tests and a shared, injectable termination seam — will add
   that regardless of what stays in scope.

Given I can't test on Windows myself, I think the responsible move is to
scope this down to only the ptyregistry fix (once verified relevant) plus
test infrastructure, and flag the daemon/worker teardown problem as needing
someone with Windows access to drive — either to test a fix I propose, or to
take it over directly. Reverting the taskkill changes now; will follow up
with the narrower PR.

anmol0b:
@illegalcall 
Pushed a revert of the taskkill/killDaemon changes — main.ts is now back to
matching main exactly. Keeping only the ptyregistry pidAlive fix while I
verify point #3 (whether that path is actually part of the reported
sequence, or a separate bug that should move to its own PR).

Will add the injectable-termination-seam tests you mentioned regardless of
where the daemon-teardown fix ends up landing.

harshitsinghbhandari:
The pidAlive logic is correct (mirrors the proven processalive.Alive, checked against the real call sites). Before merge:
- Add the Windows regression test from the earlier review round; ptyregistry_test.go still only stubs pidAlive via withFakePidAlive, so the real OpenProcess/WaitForSingleObject path ships untested.
- Update the PR title; the taskkill portion was reverted, the title still mentions it.
- Rebase on main to clear the reported conflict.

anmol0b:
> The pidAlive logic is correct (mirrors the proven processalive.Alive, checked against the real call sites). Before merge:
> 
> * Add the Windows regression test from the earlier review round; ptyregistry_test.go still only stubs pidAlive via withFakePidAlive, so the real OpenProcess/WaitForSingleObject path ships untested.
> * Update the PR title; the taskkill portion was reverted, the title still mentions it.
> * Rebase on main to clear the reported conflict.

@harshitsinghbhandari all three are addressed
- Real regression test for the Windows path. Added ptyregistry/pidalive_windows_test.go (TestDefaultPidAlive_Windows) no withFakePidAlive stub.It spawns a child, holds its own OpenProcess handle on the PID, kills the child, and asserts defaultPidAlive reports it dead within a grace window even though the lingering handle keeps OpenProcess succeeding. That's the exact zombie-handle scenario the old OpenProcess-only probe got wrong, so the test fails against the pre-fix implementation. Since build-test runs on ubuntu and windows-tagged files were never compiled in CI, I also added a windows-test job (windows-latest) to go.yml that builds the backend and runs ./internal/processalive/... ./internal/adapters/runtime/conpty/... this test included.

- Title updated. Now fix(windows): correct pidAlive WaitForSingleObject check in ptyregistry — taskkill mention dropped.

- Rebased on main. Conflict cleared; PR shows mergeable.

Vaibhaav-Tiwari (changes requested):
mostly correct but
  - The PR description is outdated and falsely says it fixes #3327.
  - It includes unrelated run-file retry changes that should be explained or separated.

i-trytoohard:
@anmol0b Marking as stale: no activity from you for 30 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@anmol0b
Closing this PR as stale after 30+ days without author activity. The linked [#3327](https://github.com/Untrivial-ai/agent-orchestrator/issues/3327) remains open if the Windows fix is still needed.

PR #3477, opened 2026-08-03
feat(reviews): choose the reviewer agent per session, and show every pass in one place

codebanditssss:
First of three, splitting #3357 so each piece can be reviewed on its own. This one is the reviewer-selection slice.

## What

**One reviewer picker, two callers.** The picker was private to `ProjectSettingsForm`. It is now a shared `ReviewerSelect`, used both for the project default and for a per-session choice, so the two read as the same control rather than two lookalikes.

**Choosing a reviewer for one session cannot affect another.** `Trigger` takes an optional harness override. An empty one keeps the project's configured reviewer; a known one runs that pass under it without editing project config. Nothing is persisted, so the choice is scoped to the pass that used it.

**Every pass in one place.** `ListReviewsResponse` now carries `runs`, every recorded pass for the session, newest first. `reviews` only held the current and previous run per PR, which cannot answer what a different reviewer said once a third pass has run, so there was no single place to summarise findings across reviewers.

The trigger and the picker sit in the panel actions; the summary is its own surface below them. Deliberately separate.

image
image
image
image
image

## Not in this PR

- the review-delivery toggle (`review_auto_inject_off`, the migration and its endpoint)
- reviewer terminal tabs and their persistence

`types/workspace.ts`, `useWorkspaceQuery.ts` and `api-client.ts` are untouched here because their changes in #3357 are entirely toggle-related.

## Worth a look in review

The OpenAPI spec is generated, so the optional trigger body needed `specgen` to stop marking every request body required. Previously every body was forced required, which would have told clients the trigger body was mandatory when the whole point is that it is optional. That is the `optionalReqBody` flag in `build.go`.

## Verification

Backend builds, all backend tests pass. Frontend typecheck clean, 1322 tests pass. Ran the app against a daemon built from this branch and confirmed the override path end to end: posting an unknown harness to the trigger route returns 422 from the validation added in `override_test.go`.


backend/internal/httpd/controllers/dto.go
```diff
@@ -841,3 +841,11 @@ type UnregisterPushDeviceResponse struct {
 	Token   string `json:"token"`
 	Deleted bool   `json:"deleted"`
 }
+
+// TriggerReviewRequest is the optional body of the review trigger route. An
+// empty harness keeps the project's configured reviewer; setting one overrides
+// it for this pass only, without editing project config, so one session's choice
+// cannot change what another session in the project runs.
+type TriggerReviewRequest struct {
+	Harness domain.ReviewerHarness `json:"harness,omitempty" enum:"claude-code,codex,opencode"`
```
neversettle17-101, line 867: Why are we not persisting the agent harness for a particular session?

frontend/src/renderer/components/ReviewerSelect.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { AgentAvatar } from "./AgentAvatar";
+import { AgentSelectMenuItem } from "./settings/AgentSelectMenuItem";
+import { SettingsOptionMenu } from "./settings/SettingsOptionMenu";
+
+// Reviewers are a narrower vocabulary than worker agents on purpose: a
+// reviewer-only tool must not become a valid worker, and the daemon rejects
+// anything outside this set. Keep in sync with domain.AllReviewerHarnesses.
+export const KNOWN_REVIEWER_HARNESS_IDS = new Set(["claude-code", "codex", "opencode"]);
+
+const REVIEWER_AGENT_PRIORITY = ["claude-code", "codex", "cursor", "opencode", "aider"] as const;
```
neversettle17-101, line 28: this should follow the same flow as worker agent. The dropdown list should be fetched from an API.

neversettle17-101 (changes requested):

frontend/src/renderer/components/ReviewerSelect.tsx
codebanditssss: good catch, and the comment above it was admitting the problem ("keep in sync with domain.AllReviewerHarnesses"), which is the kind of pairing that drifts the first time someone adds a harness in Go.

fixed in 289b1b3, though not by fetching it. the list is already in the generated schema:

```ts
// frontend/src/api/schema.ts
harness?: "claude-code" | "codex" | "opencode";
```

that union is generated from `domain.AllReviewerHarnesses` via the openapi spec (this pr wires the trigger's request body into specgen), so the daemon is already the single source. the component now derives from it with a guard in each direction:

- `satisfies readonly ReviewerHarnessId[]` rejects an id the daemon would not accept
- an `Exclude<...>` check fails the build if the daemon gains a harness the frontend has not caught up with

i broke both to confirm they actually fire rather than just compiling.

went this way rather than a runtime fetch because the set is part of the api contract and known at build time, so a request would trade a compile-time guarantee for a loading state and an error path, and it would need a new endpoint or field — `/agents` lists agents but not which of them are valid reviewers.

worth clarifying one thing about the current flow though: the reviewer picker already goes through the same catalog as the worker picker, it takes `authorized` / `installed` / `supported` straight from the agents query. only the narrowing filter was local, not the option list.

backend/internal/httpd/controllers/dto.go
codebanditssss: deliberate, and it lands in the next pr rather than this one.

this is one of three prs splitting #3357, and the split is by concern: reviewer selection here, the per-session review-delivery toggle next, reviewer terminal tabs after that. pr 2 already carries the full pattern for a persisted per-session review setting — migration, column, `PATCH /api/v1/sessions/{id}/review-policy`, dto, store, service — for `review_auto_inject_off`. adding a second persisted review setting here would stand up that same plumbing twice, in two prs, a day apart.

so what this pr gives you is the per-pass override: an empty harness runs the project's configured reviewer, an explicit one runs that pass under it without touching project config. that is what makes choosing a reviewer for one session provably unable to change what any other session in the project runs, which was the goal for this pr.

if you would rather it persist before any of this merges, i would still not do it here — i would move it into pr 2 next to the toggle, where the migration and endpoint already exist and it is a column and a field rather than a new surface. happy to do that, just say.

worth flagging one thing i did fix in this pr while in the area, since it is easy to miss: choosing a different agent did nothing on an already-reviewed commit. `Trigger` dropped any pr that was not needs_review or changes_requested before the harness was consulted, and `idx_review_run_session_pr_sha` keyed idempotency on (session, pr, commit) so a second harness collided with the first agent's run. that bug is on main today, not just here. 1ce16d36 adds harness to the index (migration 0041) and lets an explicit override through, so the picker is not inert exactly when someone reaches for it.

codebanditssss:
summarising where both comments landed, and the concrete plan for the open one so it is easy to say yes or no to.

## status

**reviewer list hardcoded** — fixed in 289b1b3, pushed. it now derives from the generated request schema, which the spec builds from `domain.AllReviewerHarnesses`, with a compile-time guard in each direction so it cannot drift. detail in the thread.

**persisting the harness per session** — not in this pr, proposed for the next one. detail below.

## the split, for context

this is 1 of 3 splitting #3357 (51 files, +2200/-371) by concern:

1. **this pr** — reviewer selection + review history in one place
2. **next** — per-session review-delivery toggle (`review_auto_inject_off`)
3. **after** — reviewer terminal tabs + persistence

## plan for persisting the reviewer

pr 2 already stands up everything a persisted per-session review setting needs, for the toggle:

- migration adding a column to `sessions`
- `PATCH /api/v1/sessions/{sessionId}/review-policy` + request/response dtos
- `SetReviewAutoInject` through service -> store -> sqlc queries
- the session view/dto plumbing so the renderer sees it

adding the reviewer there is a column and a field on top of that, not a new surface:

- `sessions.reviewer_harness TEXT NOT NULL DEFAULT ''` in the same migration
- `reviewerHarness` on the session dto, beside `reviewAutoInjectOff`
- extend the existing review-policy request rather than adding a second endpoint
- `Trigger` reads the stored value when no explicit override is passed, so precedence becomes: explicit pick for this pass > session default > project default

doing it in this pr instead means building that migration + endpoint + dto + store path here, and then pr 2 lands a second one a day later for the toggle. same plumbing twice, and the two prs would conflict on `sessions.sql`, the generated sqlc output and the dto file.

## what this pr does give you

the per-pass override: empty harness runs the project's configured reviewer, an explicit one runs that pass under it without touching project config. that is what makes a reviewer choice in one session provably unable to change what any other session in the project runs.

it is a real gap that the choice does not stick, i am not arguing otherwise — only about which pr it lands in.

**happy to do it either way.** if you would rather not merge this without persistence, say so and i will pull it into pr 2 and hold this one until that lands, or add it here and drop it from pr 2. just want to avoid building the same plumbing twice by accident.

## one thing worth a look while you are here

1ce16d36 fixes a bug that is on main today, easy to miss in the diff: choosing a different agent did nothing on an already-reviewed commit. `Trigger` dropped any pr that was not `needs_review`/`changes_requested` before the harness was consulted, and `idx_review_run_session_pr_sha` keyed idempotency on (session, pr, commit), so a second harness collided with the first agent's run and silently reused it. migration 0041 adds harness to that index and an explicit override now gets through. the default path is unchanged, so it will not re-review on every trigger.

neversettle17-101 (changes requested):
1. The PR should show the verdict, user shouldn't have to click on the dropdown to check the verdict. 
2. By default all the PR along with its verdict should be collapsed

frontend/src/renderer/components/ReviewerSelect.tsx
neversettle17-101: Let us revert this change in this PR and take it as a followup.

codebanditssss:
### PR flow

```text
Open Reviews tab
       ↓
Choose reviewer
Claude Code / Codex / OpenCode
       ↓
Selection is saved for this session
       ↓
Click Run review
       ↓
A fresh reviewer process starts
       ↓
Reviewer checks every open PR
       ↓
Results appear under AO Code Reviews
       ↓
PR row shows the latest verdict
Approved / Changes requested / Cancelled
       ↓
Expand the PR row
       ↓
See past completed review summaries
```

Additional behavior:

- Re-run review starts a fresh reviewer process.
- Open terminal opens the active reviewer terminal.
- Each session remembers its own reviewer.
- Failed, cancelled, running, and raw terminal history are not shown as review summaries.
- Review controls remain readable when the inspector is resized.

backend/internal/httpd/controllers/dto.go
codebanditssss: Done now. The selected reviewer is persisted per session through the session reviewer endpoint and stored in the reviewer_harness session field. Switching sessions restores each session’s own reviewer selection.

frontend/src/renderer/components/ReviewerSelect.tsx
codebanditssss: Agreed. The runtime API-backed reviewer list is not part of this PR and will be handled as a follow-up, using the same flow as the worker-agent selector. This PR currently keeps the reviewer IDs checked against the generated API schema, but does not fetch that list at runtime.

backend/internal/storage/sqlite/migrations/0042_review_run_unique_per_harness.sql
neversettle17-101, line 1: there shouldnt be 2 migration files it should be combined to 1

neversettle17-101 (changes requested):

backend/internal/storage/sqlite/migrations/0042_review_run_unique_per_harness.sql
codebanditssss: Done. I combined both schema changes into migration 0042 and removed 0043. The migration now adds the session reviewer_harness column and updates the review-run unique index together, with both operations reversed in the Down migration. Relevant backend migration, storage, session-service, and review tests pass.

neversettle17-101 (changes requested):
1. Duplicate fallback ignores reviewer harness. Trigger now inserts runs under a  
    (session, PR, SHA, harness) uniqueness model, but on duplicate it calls
    GetReviewRunBySessionPRAndSHA without harness filtering (backend/internal/review/
    review.go:268-274). The SQL still selects only by session_id, pr_url, and target_sha,
    newest first (backend/internal/storage/sqlite/queries/review.sql:34-36). With multiple
    reviewers on the same commit, a duplicate for one harness can fall back to another
    harness’s newer run. Fix by adding a harness-aware lookup and using it in the duplicate
    path 
 2. SetSessionReviewerHarness bypasses the store write mutex. Neighboring session
    write methods lock s.writeMu, but the new method writes through s.qw without that lock
    (backend/internal/storage/sqlite/store/session_store.go:95-105). This can interleave
    with other session writes and hit SQLite write contention/racy update behavior. Add the
    same s.writeMu.Lock() / defer Unlock() pattern used by rename, preview, and merge-
    policy updates.
 3. New GitHub review verdict labels are hardcoded English. githubVerdict returns
    "Approved", "Changes requested", "Review required", and "Commented" directly (frontend/
    src/renderer/components/SessionInspector.tsx:1356-1366) while this PR routes
    surrounding renderer strings through i18n. Use catalog keys so the new panel localizes
    consistently.

codebanditssss:
Addressed all three findings in `4f445a7f6`:

- duplicate fallback now uses a `(session, PR, SHA, harness)` lookup, with an engine regression test proving a newer run from another harness is not returned
- `SetSessionReviewerHarness` now takes `writeMu`, matching the neighboring session write paths
- GitHub verdict labels now use i18n catalog keys in both supported locales

Verified with:
- `go test ./internal/review ./internal/storage/sqlite/store`
- frontend `npm run typecheck`
- 59 focused renderer/i18n tests

neversettle17-101:
Screenshot 2026-08-04 at 5 08 55 PM
Screenshot 2026-08-04 at 5 09 05 PM

frontend/src/renderer/components/ReviewerSelect.tsx
```diff
@@ -0,0 +1,125 @@
...
+		installed,
+		authorized,
+		priorityRank: REVIEWER_AGENT_PRIORITY_RANK,
+		fallbackAgents,
+	});
+
+	// The trigger shows the agent that will actually run, since "project default"
+	// names the setting rather than answering the question. The menu keeps the
+	// longer wording, where there is room for it.
+	const defaultLabel = defaultHarness ? `Project default (${defaultHarness})` : "Project default";
```
neversettle17-101: This shared component is now used by ProjectSettingsForm too, but the default option is hard-coded as `Project default`. In project settings that makes the default reviewer setting read as `Project default` for its own value, and it bypasses the existing localized `settings.project.default` label. Please let callers provide the default-option/trigger label (or use i18n here) so the project settings picker keeps showing the localized `Default` while the session override can say `Project default (...)`.

neversettle17-101 (commented):
Needs changes: I found one UI regression in the shared reviewer picker. The backend/API changes and review-run history shape look aligned with the PR intent from the diff, but the project settings caller now inherits per-session wording and hard-coded English for its default option.

frontend/src/renderer/components/ReviewerSelect.tsx
neversettle17-101: Addressed in 3a118c692. ReviewerSelect now takes caller-provided default option and trigger labels instead of hard-coding `Project default`. ProjectSettingsForm passes the localized `settings.project.default` label for both the menu option and trigger, while SessionInspector keeps the session override wording with `Project default (...)`. Added a ProjectSettingsForm regression test for the default reviewer picker label.

neversettle17-101 (commented):
Ready to merge from my review. The latest update makes the reviewer default labels caller-provided/localized, which addresses the previous project-settings picker regression. I re-reviewed the backend trigger/session reviewer persistence, the review-run uniqueness migration, generated OpenAPI optional trigger body, shared ReviewerSelect, and SessionInspector review history surfaces, and I did not find any remaining blocking issues. Tests were not run by me; the PR includes targeted backend/store/controller tests plus a ProjectSettingsForm regression test, and the PR description reports backend and frontend validation.

neversettle17-101 (approved):

Issue #3303, opened 2026-07-29
Reviewer terminal should open as a new tab instead of replacing the worker agent terminal


i-trytoohard:
## Problem

When a reviewer is triggered, its terminal **replaces** the current worker agent terminal in the center pane. To get back to the worker, you have to find the reviewer toolbar and click "Back to agent." This is friction — especially when you want to toggle between the two frequently during a review.

**Expected:** The reviewer terminal should open as a **new tab** in the terminal strip, alongside the worker agent tab — just like shell terminals and pinned sessions already work.

## Root Cause

The center pane renders exactly **one terminal at a time** via a single state variable:

**`frontend/src/renderer/components/SessionView.tsx:78`**
```ts
const [terminalTarget, setTerminalTarget] = useState<TerminalTarget>({ kind: "worker" });
```

This is a single value, not a collection. When a reviewer opens, it overwrites the state:

**`SessionView.tsx:479-481`** (only reviewer-target site in the codebase)
```tsx
onOpenReviewerTerminal={({ handleId, harness }) =>
    setTerminalTarget({ kind: "reviewer", handleId, harness })
}
```

This fires automatically after triggering a review (`SessionInspector.tsx:742-754`, inside `triggerReview` mutation `onSuccess`).

The codebase documents this as a known design limitation at **`stores/ui-store.ts:66-71`**:
> *"The session pane renders one terminal at a time, so opening a shell or the reviewer swaps the agent's terminal off screen even though the route still points at that session."*

## Why Shells Get Tabs But Reviewers Do Not

| Feature | How it's tracked | Renders as tab? |
|---------|-----------------|-----------------|
| Shell terminals | `useShellTerminals()` + `activeShellTerminalHandleId` | ✅ Yes (`CenterPane.tsx:224-233`) |
| Pinned worker sessions | `sessionTabsByOwner` + `addSessionTab` | ✅ Yes (`CenterPane.tsx:198-223`) |
| **Reviewers** | Ephemeral `terminalTarget` state only | ❌ **No** |

Reviewers are not modeled as a tab collection — the `handleId` lives only in transient state.

## Backend Already Supports This

The backend creates a **distinct, stable, addressable tmux pane** for the reviewer:

**`backend/internal/review/launcher.go:121-123`**
```go
func reviewerHandleID(workerID domain.SessionID) string {
    return "review-" + string(workerID)
}
```

[...]

PR #3480, opened 2026-08-03
feat(frontend): polish terminal tab bar

### LINKED ISSUE #3479: Polish the session terminal top bar and clarify its tab model
## Problem

The terminal top bar in the desktop session view does not clearly communicate its model or make good use of the available space.

In the maintainer-provided screenshot, the strip above the active terminal reads roughly:

```text
TERMINAL   [sidebar-min-width]                                      >_
```

The current presentation has several UX shortcomings:

- the active terminal/session label reads like inert metadata rather than a selectable tab;
- the lone `>_` affordance does not clearly mean “new terminal”;
- large areas of dead space make the strip feel unfinished;
- terminal actions (new, close, zoom/font sizing, expand) are not grouped or labeled consistently;
- the active state and per-terminal close behavior are not visually obvious;
- the design does not scale naturally to worker, shell, pinned-session, and reviewer terminals living side by side.

## Proposed improvement

Redesign the strip as an intentional, compact tab bar that follows `DESIGN.md` and the desktop app’s existing shadcn/lucide visual language:

- render each open terminal as a clear tab with a visible active state;
- provide an obvious `+ New terminal` action (label plus icon where space allows);
- expose per-tab close affordances without making accidental closes easy;
- group zoom/font and expand controls consistently at the trailing edge;
- add keyboard-accessible controls, focus states, accessible names, and tooltips;
- preserve the terminal’s own palette and existing terminal keyboard behavior;
- remain useful at narrow widths through truncation/overflow behavior rather than ambiguous glyph-only chrome.

The result should make multiple terminals understandable at a glance and use horizontal space deliberately while remaining compact.

## Related issue

This is closely related to #3303, which requests that a reviewer terminal open as a new tab instead of replacing the worker terminal. A clearer, scalable terminal tab bar should make that behavior feel native: worker, reviewer, shell, and pinned-session terminals can all appear as peer tabs with one active terminal.

This issue focuses on the top-bar interaction and presentation. #3303 remains the source of truth for modeling reviewer terminals as tabs.

## Acceptance criteria

- [ ] Open terminals have clear tab affordances and a visible active state.
- [ ] Creating a terminal is discoverable and opening/closing tabs is understandable.
- [ ] Existing zoom/font-size and expand controls remain available.
- [ ] Controls are keyboard accessible and have accessible names/tooltips.
- [ ] The implementation follows `DESIGN.md`, uses existing shadcn primitives where appropriate, and retains the xterm palette.
- [ ] Component tests cover the changed behavior.
- [ ] The desktop app is visually verified by opening, switching, and closing terminal tabs.
- [ ] The PR includes before/after screenshots.

AgentWrapper:
Fixes #3479
Related to #3303

## Summary

- make the terminal strip span the workspace while keeping terminal controls aligned to the terminal pane
- give the permanent main agent terminal a prominent harness identity and inline activity state; shell tabs use terminal icons, clear active styling, and sibling close controls
- provide a compact plus action, Cmd/Ctrl+T creation, Cmd/Ctrl+W active-shell closing, Arrow/Home/End tab navigation, tooltips, and accessible tab semantics
- keep session actions at the far right, hide them in terminal fullscreen, and correct titlebar clearance when the sidebar is collapsed
- preserve the terminal palette, zoom/expand controls, and existing terminal input fixes
- omit shell creation from orchestrator terminal bars, remove the redundant Summary overview card, keep the owning project pill active on orchestrator routes, and emphasize the selected Kanban or Orchestrator destination icon

The reviewer persistence requested in #3303 remains follow-up work; this tab model gives reviewer terminals a natural place to live without changing reviewer lifecycle behavior in this PR.

## Verification

- `npm --prefix frontend test` (110 files, 1,454 tests)
- iterative visual verification in the live AO desktop app against the local daemon: opened, selected, and closed terminal tabs; checked collapsed-sidebar, fullscreen, active-tab, hover-close, and control alignment states


frontend/src/renderer/lib/terminal-tabs.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	event.preventDefault();
+	const nextIndex =
+		event.key === "Home"
+			? 0
+			: event.key === "End"
+				? tabs.length - 1
+				: (currentIndex + (event.key === "ArrowRight" ? 1 : -1) + tabs.length) % tabs.length;
+	tabs[nextIndex]?.focus();
+	tabs[nextIndex]?.click();
```
AgentWrapper, line 23: **[P1] Avoid synthesizing pointer clicks for keyboard tab navigation.** On macOS/Linux, `ShellTerminalTab` bubbles every click to `handleClick`, which treats any two clicks within 500 ms as the rename gesture. This `.click()` participates in that counter, so ArrowRight → ArrowLeft → ArrowRight (or repeated Home/End on the same tab) can unexpectedly replace the tab with the rename input. The new test visits a shell only once, so it misses the interaction. Select via a keyboard-safe callback/event, or make the rename detector accept only real pointer clicks, and add a regression test that revisits a tab quickly.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -117,80 +178,154 @@ export function CenterPane({
...
+								<ChevronRight aria-hidden="true" className="size-icon-md" />
+							</button>
+						) : null}
+						<Tooltip>
+							<TooltipTrigger asChild>
+								<Button
+									aria-label={t("shortcut.new-shell-terminal")}
+									className="shrink-0 text-muted-foreground"
+									disabled={!onNewShellTerminal}
+									onClick={onNewShellTerminal}
```
AgentWrapper: **[P1] Do not put all session actions in this width-constrained, non-shrinking terminal row.** The app supports a 960 px window with a 240 px sidebar, and the inspector can leave the terminal panel around 400 px wide. `New terminal`, the display toolbar, and this worker action cluster are all `shrink-0`; once the tab list reaches zero width, the fixed controls still exceed the panel and the shell clips or paints the rightmost Kill/Orchestrator/Inspector/notification actions over the inspector. This violates the narrow-width acceptance criterion and can make required controls unreachable. Keep session-level actions in a full-width session header or introduce an explicit responsive/collapsed treatment, with a test at minimum window width and maximum inspector width.

AgentWrapper (commented):
## Changes requested

The visual consolidation and focused tests are a solid start, but two user-visible regressions need correction before this is ready: keyboard tab navigation can accidentally enter rename mode, and the fixed action groups do not fit the supported narrow terminal widths.

Code-quality note: this change also leaves `terminal.exitFullscreenAria` and `terminal.openFullscreenAria` unused in both locale files after replacing their only call sites; please remove that dead translation surface or use the specific labels again.

I reviewed the exact head `27c75575596cd48eb1d716115cfcc9cc20036161`; the current GitHub checks are green, but neither issue below is exercised by the added tests.

frontend/src/renderer/lib/terminal-tabs.ts
AgentWrapper, line 23: **[P1] Keep keyboard selection out of pointer rename detection.** Calling `.click()` here bubbles into `ShellTerminalTab` and updates its 500 ms click counter. On macOS/Linux, ArrowRight → ArrowLeft → ArrowRight (or quickly revisiting a tab with Home/End) clicks the same shell twice and replaces it with the rename input even though the user never performed the rename gesture. Select through a keyboard-safe callback, or exclude synthesized keyboard clicks from the rename counter, and add a regression that revisits a shell tab quickly.

frontend/src/main/app-shortcuts.ts
```diff
@@ -43,6 +44,7 @@ type ShortcutTargetContents = {
 const mainShortcutChannels: readonly [AppShortcutId, string][] = [
 	["new-session", NEW_SESSION_SHORTCUT_CHANNEL],
 	["new-shell-terminal", NEW_SHELL_TERMINAL_SHORTCUT_CHANNEL],
+	["close-shell-terminal", CLOSE_SHELL_TERMINAL_SHORTCUT_CHANNEL],
```
AgentWrapper, line 47: **[P1] Do not globally consume the close chord when no shell tab can close.** Once this command is in `mainShortcutChannels`, `attachAppShortcuts` always calls `preventDefault()` for Cmd/Ctrl+W in every attached web contents. The renderer handlers only act when a standalone shell is active (`target.kind === "shell"`), so on an agent/orchestrator terminal or any non-terminal route the event becomes a no-op. This removes the normal Ctrl+W word-delete input from Windows/Linux terminals and blocks the standard Cmd+W window action on macOS. Gate interception on active closeable-shell context (or otherwise preserve the original event when there is nothing to close), with coverage for the permanent agent tab/no-shell case.

AgentWrapper (commented):
## Changes requested

The full-width portal/layout addresses the earlier narrow-width concern, but two keyboard regressions remain at head `faf99ab3b30c5e4132b5595ce1cd3616c585c437`: keyboard tab navigation can still trigger rename mode, and the global close-terminal shortcut consumes Ctrl/Cmd+W even when there is no closeable shell tab.

`npm --prefix frontend run typecheck` and `typecheck:e2e` passed locally. Focused Vitest startup was blocked by the checkout dependency state (`ERR_REQUIRE_ESM` loading `std-env`); the current GitHub checks are green.

frontend/src/main/app-shortcuts.ts
AgentWrapper: Addressed in 30b3d415b. The renderer now publishes whether the active view has a closeable shell tab; main-process shortcut matching checks that live state before calling preventDefault or forwarding Cmd/Ctrl+W, including browser-preview web contents. Added main shortcut, preload bridge, CenterPane lifecycle, and e2e bridge coverage.

frontend/src/renderer/lib/terminal-tabs.ts
AgentWrapper: Addressed in 30b3d415b. Synthetic keyboard activation now selects the tab without participating in the pointer double-click timer (`MouseEvent.detail === 0`), and the CenterPane regression revisits the same shell rapidly with ArrowRight/Home/ArrowRight and verifies rename mode does not open.

AgentWrapper:
Addressed [review 4844079512](https://github.com/Untrivial-ai/agent-orchestrator/pull/3480#pullrequestreview-4844079512) in `30b3d415b`:

- Keyboard-driven tab selection is excluded from the pointer double-click rename detector, with a rapid revisit regression test.
- Cmd/Ctrl+W is intercepted only while the active view exposes a closeable shell tab; otherwise the native chord is preserved across the main window and browser preview contents.
- Updated preload/e2e bridge coverage and restarted the dev Electron app against the rebuilt preload.

Verification: 1,384 Vitest tests, renderer TypeScript, e2e TypeScript, and the finite Vite renderer build all pass.

frontend/src/renderer/lib/terminal-tabs.ts
AgentWrapper: Addressed in 30b3d415b. Synthetic keyboard clicks (detail 0) still select the tab but no longer advance the pointer rename timer, and the CenterPane keyboard regression now revisits the same shell rapidly.

frontend/src/renderer/components/CenterPane.tsx
AgentWrapper: Addressed by the full-width portal layout already present at faf99ab3b: the terminal tab row spans the window while its controls remain constrained to the terminal x-axis. The follow-up review confirmed this resolved the narrow-width concern.

PR #3483, opened 2026-08-03
feat(spawn): add --model for per-session model override

### LINKED ISSUE #3331: feat(spawn): add --model to ao spawn for per-session model override

vanshx999:
Closes #3331

## What

Adds `ao spawn --model <model>` for a per-session model override, plus the matching `model` field on `POST /api/v1/sessions`. The override is scoped to the spawned session only — it never mutates project config.

## Precedence

Mirrors the chain from the issue, folded into the existing `effectiveAgentConfig` merge:

```
spawn --model > role agentConfig.model > project agentConfig.model
```

The resolved model is pinned on the durable session record at spawn so `ao session get` reports what the session actually launched with (new `model` field on `session get`, human + `--json`), and a later restore keeps the same model even if the project config changes.

## Changes

- **CLI** (`spawn.go`): `--model` flag, validated as a single token (rejects spaces/quotes so it can't smuggle launch args), forwarded as `model` in the spawn request. `session.go`: `model` in `session get` output.
- **Daemon** (`dto.go`, `sessions.go`, `ports/session.go`, `manager.go`): `SpawnSessionRequest.model`, `SpawnConfig.Model`, precedence folding in `effectiveAgentConfig`, pinned via `seedRecord`.
- **Persistence** (`0041_add_session_model.sql`, `queries/sessions.sql`, sqlc gen, store): `sessions.model` column, round-trip tested.
- **API**: OpenAPI spec + `frontend/src/api/schema.ts` regenerated (`npm run api`), sqlc regenerated (`npm run sqlc`).

## Tests

- DTO-drift e2e: `--model` arrives intact through the real HTTP round trip (`dto_drift_e2e_test.go`)
- CLI: `--model` rejects whitespace/quoted values (`spawn_test.go`)
- Precedence: spawn model wins over role override and base config for both worker and orchestrator (`provision_test.go`)
- HTTP: spawn request decodes `model` and echoes it on the session read model (`sessions_test.go`)
- Store: `model` column insert/update/get round trip (`store_test.go`)

`go test ./...`, `go vet ./...`, `npm run lint`, and `npm run frontend:typecheck` all pass (three unrelated flaky agent-adapter timing tests fail intermittently on clean `main` too).


somewherelostt:

- @illegalcall
- @pulkit7070

Join the session here:
https://discord.gg/H6ZDcUXmq

harshitsinghbhandari:
There are merge conflicts, please fix them.

vanshx999:
@harshitsinghbhandari  Resolved the merge conflicts against the current `main` (which added chat/TUI session modes, the conversation tables, and the interface-transition coordinator after this branch was cut).

Changes:
- Kept both `ao spawn --model` and `--mode` validation in `spawn.go`.
- Folded the per-session `Model` into `SpawnConfig` alongside the base's `RequestedMode` (`dto.go`/`sessions.go`).
- `SessionRecord` now carries `Model`, `Mode`, and `Activity`.
- Pinned the resolved model via `effectiveAgentConfig` precedence (`--model > spawn agentConfig > role > project`) in `manager.go`.
- Added `sessions.model` to queries + regenerated sqlc gen alongside `session_mode`/provider/controller columns; kept both schema repairs in `db.go`.
- Regenerated `openapi.yaml` and `frontend/src/api/schema.ts` to include `model` and `mode`.

Verified: backend `go build`/`go vet`, specgen drift test, and `npm run frontend:typecheck` all pass. The CLI spawn 404 tests and kilocode/opencode deadline tests fail identically on plain `untrivial/main` (pre-existing), not from this merge.

neversettle17-101:
@vanshx999 can you locally test and add the testing details?

Rishet11:
@vanshx999 
#4102 landed per-session --model on spawn, and issue #3331 is closed completed.  Closing as superseded.

PR #3484, opened 2026-08-03
fix: align reviewer terminal lifecycle

### LINKED ISSUE #3467: Add review terminal tabs and persist it
1. We should be able to navigate to the reviewer terminal from the main terminal itself.
2. Reviewer should follow the same lifecycle as agent worker session

neversettle17-101:
## Summary
- hard-terminate reviewer panes when their worker is explicitly killed
- clear stale reviewer panes before restoring a worker session
- keep reviewers alive across normal daemon shutdown/save-teardown, matching worker runtime recovery semantics
- show reviewer terminals as selectable tabs with the harness icon

Closes #3467

## Tests

- go test ./internal/session_manager ./internal/review ./internal/service/review ./internal/daemon ./internal/httpd/controllers
- git diff --check

## Notes
- Frontend tests/typecheck were not run because frontend/node_modules is absent in this worktree.


backend/internal/review/review.go
```diff
@@ -444,6 +451,45 @@ func (e *Engine) Cancel(ctx stdctx.Context, workerID domain.SessionID) (CancelRe
...
+	}
+	cancelled := make([]domain.ReviewRun, 0, len(running))
+	for _, run := range running {
+		run.Status = domain.ReviewRunCancelled
+		run.Verdict = domain.VerdictNone
+		run.Body = body
+		run.GithubReviewID = ""
+		cancelled = append(cancelled, run)
+	}
+	return TerminateResult{ReviewerHandleID: review.ReviewerHandleID, CancelledRuns: cancelled}, nil
```
neversettle17-101: `TerminateReviewer` destroys the pane, but the review row still keeps `ReviewerHandleID`. `Engine.List` returns that persisted handle verbatim, and the new SessionView tab code renders any non-empty `reviewerHandleId`, so after `RestoreWithMode` destroys the old reviewer and relaunches the worker the restored active session can still show and attach to a dead reviewer tab. Please clear/update the stored handle during hard termination, or only expose handles that are confirmed alive.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker session lifecycle: explicit kill/restore hard-destroys reviewer panes, normal shutdown leaves them recoverable, and the UI can show reviewer terminals as selectable tabs with harness icons.

## Summary of Changes

- Adds a reviewer terminator hook from session manager to review service and calls it during worker Kill and RestoreWithMode, but not SaveAndTeardownAll.
- Adds review engine hard Destroy behavior for reviewer panes and cancellation of running review runs.
- Adds a frontend reviews query in SessionView to surface reviewerHandleId as a CenterPane reviewer tab, plus harness avatar chrome.
- Hides review status rows from PR summary cards and updates backend/frontend tests around lifecycle and UI behavior.

## Findings

- **Medium**: Stale reviewer handle remains after hard teardown -- `TerminateReviewer` destroys the runtime handle but leaves `review.reviewer_handle_id` unchanged. `List` still returns that persisted handle, and this PR now renders any non-empty `reviewerHandleId` as an attachable reviewer tab for active sessions. After a restore, the old reviewer pane has been destroyed but the restored active session can still show and attach to that dead handle until another trigger refreshes the row. Please clear or replace the persisted reviewer handle as part of hard termination, or make the list/UI only expose handles that are still alive.

## Description Match

The backend mostly implements the promised lifecycle split, but the stale persisted handle means the restore path does not fully clear the reviewer terminal state exposed to the new active session.

## Regression Risk

The main regression risk is confusing/dead reviewer tabs after restore or any explicit lifecycle teardown that destroys the reviewer pane while keeping the review row.

## Testing

The PR adds focused Go and React tests for the new lifecycle hooks and tab rendering. Tests were not run as part of this review. The missing coverage is a restore/list or UI query case proving a hard-destroyed reviewer handle is no longer advertised to the active restored session.

Verdict: changes requested.

backend/internal/review/review.go
neversettle17-101: Addressed in 9a647ad56.

Backend: `TerminateReviewer` now clears the persisted `reviewer_handle_id` immediately after hard-destroying the reviewer pane, and the review engine test asserts `List` no longer advertises a handle after teardown. Added the SQLite `ClearReviewerHandle` query/store method plus a store round-trip assertion.

UI: removed the reviewer-only toolbar that showed the “Back to agent” button and duplicate Reviewer label; the reviewer remains selectable through the terminal tab strip.

neversettle17-101:
Updated the restore behavior per the latest requirement: restoring a worker session no longer terminates the reviewer terminal. If a review had been triggered before termination, the persisted reviewer handle is preserved and the UI keeps the selected reviewer target through the terminated -> active transition, so the reviewer tab becomes visible again after restore. Explicit worker kill still tears down the reviewer and clears the persisted handle.\n\nVerified locally with:\n- go test ./internal/session_manager\n- npm --prefix frontend run test -- src/renderer/components/SessionView.test.tsx\n- npm run frontend:typecheck\n- git diff --check\n\nGitHub CI is green on the latest head commit.

backend/internal/review/review.go
```diff
@@ -444,6 +494,48 @@ func (e *Engine) Cancel(ctx stdctx.Context, workerID domain.SessionID) (CancelRe
...
+// remove the terminal pane itself.
+func (e *Engine) TerminateReviewer(ctx stdctx.Context, workerID domain.SessionID, body string) (TerminateResult, error) {
+	if workerID == "" {
+		return TerminateResult{}, fmt.Errorf("%w: worker session id is required", ErrInvalid)
+	}
+	review, ok, err := e.store.GetReviewBySession(ctx, workerID)
+	if err != nil {
+		return TerminateResult{}, err
+	}
+	if !ok || review.ReviewerHandleID == "" {
```
neversettle17-101: `TerminateReviewer` returns before listing or cancelling running runs whenever the review row is missing or `reviewer_handle_id` is empty. That state is reachable because `Trigger` inserts the review row and running `review_run` before the final upsert persists the spawned handle, so a daemon crash or concurrent worker kill in that window leaves a running review run that worker termination will not cancel. Since killed sessions cannot be triggered again, the stale running run can persist indefinitely. The hard teardown path should still cancel `ListRunningReviewRunsBySession` / `CancelRunningReviewRunsBySession` even when there is no handle to destroy, and only skip `Destroy` itself when the handle is empty.

backend/internal/storage/sqlite/migrations/0042_review_agent_session_id.sql
```diff
@@ -0,0 +1,52 @@
...
+        AND verdict = '';
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP INDEX idx_review_run_session_pr_sha;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE UNIQUE INDEX idx_review_run_session_pr_sha
```
neversettle17-101, line 52: After this migration is in use, explicit reruns can create multiple completed or delivered rows with the same `(session_id, pr_url, target_sha)`, which is the intended new behavior. The down migration recreates the old partial unique index without first deduping or excluding those rows, so rollback can fail as soon as a PR head has been rerun successfully. Please add a down-side cleanup strategy or adjust the rollback index predicate so the migration remains reversible after normal use.

neversettle17-101 (commented):
## Findings

Requesting changes. I found two lifecycle/data issues that should be fixed before merge.

- Medium: worker termination can leave running review runs behind when the review row has no handle.
- Low: the down migration can fail after the new rerun behavior creates duplicate completed rows for the same PR head.

I did not run tests per review instructions. I reviewed the PR metadata and diff against origin/main. Note: the queued task named head ff8c83e, while the live PR ref fetched as eacb1c74.

backend/internal/review/review.go
neversettle17-101: Addressed in eebed637e.

`TerminateReviewer` now lists and cancels running review runs even when the review row is absent or has an empty `reviewer_handle_id`; it only skips the runtime `Destroy` call when there is no handle. Added a regression test covering running-run cancellation with a missing reviewer handle.

backend/internal/storage/sqlite/migrations/0042_review_agent_session_id.sql
neversettle17-101: Addressed in eebed637e.

The down migration now dedupes rows matching the old uniqueness predicate before recreating `idx_review_run_session_pr_sha`, using the same survivor ordering as the up-side cleanup. That keeps rollback reversible after normal explicit reruns create duplicate completed/delivered rows for the same PR head.

backend/internal/session_manager/manager.go
```diff
@@ -1168,7 +1209,16 @@ func (m *Manager) RestoreWithMode(ctx context.Context, id domain.SessionID) (Res
 }
 
 func (m *Manager) relaunchRestoredSession(ctx context.Context, rec domain.SessionRecord, project domain.ProjectRecord, ws ports.WorkspaceInfo) (RestoreResult, error) {
-	return m.relaunchSession(ctx, "restore", rec, project, ws, nil)
+	result, err := m.relaunchSession(ctx, "restore", rec, project, ws, nil)
+	if err != nil {
+		return RestoreResult{}, err
+	}
+	if rec.Kind == domain.KindWorker {
+		if err := m.restoreReviewer(ctx, rec.ID); err != nil {
```
neversettle17-101, line 1385: This makes reviewer restore a hard failure after `relaunchSession` has already created the worker runtime and `MarkSpawned` has made the session active. If the reviewer restore fails, for example due to a missing reviewer binary or runtime create error, `RestoreWithMode` returns an error even though the worker is now live; `RestoreAll` also skips marker cleanup because it sees an error. That leaves the caller and persisted state disagreeing, and can leave stale restore inventory behind. Please either make reviewer restore best-effort after the worker is restored, or add explicit rollback/termination so a returned restore error leaves the worker in a consistent terminated state.

neversettle17-101 (commented):
## Findings

Requesting changes. I found one lifecycle consistency issue in the restore path.

- Medium: reviewer terminal restore failure is returned after the worker has already been relaunched and marked active, so callers see restore failure while durable state says the worker is live.

I did not run tests per review instructions.

neversettle17-101 (commented):
## Findings

Requesting changes. One restore-path consistency issue remains at the current head.

- Medium: reviewer restore failure is still returned after the worker session has already been relaunched and marked active.

I did not run tests per review instructions.

backend/internal/session_manager/manager.go
neversettle17-101, line 1385: This hook still runs after `relaunchSession` has already created the runtime and called `MarkSpawned` (see the same function returning the updated active record after line 1352). If `RestoreReviewer` fails here, `RestoreWithMode` returns an error to the API/CLI even though the worker is now live and durable state says it is active. That leaves callers believing restore failed and can trigger retry/error UI around a session that was already restored. Either make reviewer restore best-effort/logged after worker restore succeeds, or move it into a phase that can be rolled back before `MarkSpawned` is committed.

neversettle17-101: Addressed in 2fb224847.

Reviewer terminal restore is now best-effort after the worker restore succeeds: `relaunchRestoredSession` logs a warning if `RestoreReviewer` fails and still returns the successful worker `RestoreResult`. Added `TestRestore_ReviewerRestoreFailureLeavesWorkerRestored` to prove callers get an active worker result and durable session state remains restored when reviewer restore fails.

neversettle17-101: Addressed in 2fb224847.

Reviewer restore failure no longer makes `RestoreWithMode` fail after `relaunchSession` has already committed the worker as active. The reviewer restore path is best-effort/logged, and the new session-manager test covers the partial-failure case so the returned result and durable worker state stay consistent.

backend/internal/review/review.go
```diff
@@ -286,7 +312,7 @@ func (e *Engine) Trigger(ctx stdctx.Context, workerID domain.SessionID) (Trigger
 
 	handleID := ""
 	queue := reviewQueue(created)
-	if hasReview && reviewRow.ReviewerHandleID != "" && prevHarness == harness {
+	if hasReview && reviewRow.ReviewerHandleID != "" && prevHarness == harness && strings.TrimSpace(reviewRow.AgentSessionID) != "" {
```
neversettle17-101: This gate should not require AgentSessionID to reuse a currently live reviewer pane. The native id is only needed to recreate/resume a destroyed terminal; if the handle is alive, Notify can target it exactly as before. With this condition, a second trigger that creates work before the reviewer hook posts /reviews/activity skips Alive, falls through to Spawn, and agentLauncher destroys the stable reviewer handle before launching a new pane. Any already-running runs from the old pane are then orphaned as running. Please keep live-pane reuse keyed on handle+harness+Alive, and reserve AgentSessionID for restore/resume decisions.

backend/internal/adapters/reviewer/agentrestore/agentrestore.go
```diff
@@ -0,0 +1,46 @@
...
+	Permissions     ports.PermissionMode
+	AllowedTools    []string
+	DisallowedTools []string
+}
+
+// Command asks the wrapped agent adapter to resume the reviewer's native
+// conversation. ok=false means no native id has been captured yet, so callers
+// may fall back to a fresh idle launch for legacy review rows.
+func Command(ctx context.Context, agent ports.Agent, inv ports.ReviewInvocation, opts Options) (ports.ReviewCommandSpec, bool, error) {
+	agentSessionID := strings.TrimSpace(inv.AgentSessionID)
```
neversettle17-101, line 25: This early return prevents Claude Code from using its existing restore fallback for pre-hook sessions. The Claude adapter GetRestoreCommand can derive the deterministic --session-id from cfg.Session.ID when the hook metadata is absent, but the shared helper never calls it unless inv.AgentSessionID is non-empty. That means migrated Claude reviewer rows with blank agent_session_id restore as fresh idle sessions instead of resuming the original conversation. Consider always calling GetRestoreCommand with Session.ID populated and only omitting the metadata value when it is blank.

neversettle17-101 (commented):
## What This PR Does

This PR adjusts reviewer terminal lifecycle so reviewer panes are surfaced as tabs, can be killed/restored separately from normal review cancellation, and persist/restore native agent session ids. It also adds API routes and storage for reviewer activity/restore state plus frontend controls for the new lifecycle.

## Summary of Changes

- Adds reviewer restore/kill/activity API endpoints and generated OpenAPI/frontend types.
- Persists review.agent_session_id and changes review_run uniqueness to permit explicit reruns of completed heads.
- Wires session-manager kill/restore paths to reviewer teardown/restore and adds launcher support for idle restored reviewer panes.
- Adds reviewer tab UI and inspector actions for trigger/cancel/kill/restore.

## Findings

- **Medium**: Live reviewer reuse now depends on agent_session_id. This can destroy an existing reviewer pane instead of notifying it whenever the hook id has not been recorded yet, leaving prior running runs orphaned.
- **Medium**: Claude reviewer restore bypasses the adapter fallback for pre-hook sessions because the shared restore helper returns before calling GetRestoreCommand when agent_session_id is empty.

## Description Match

The PR mostly implements the described lifecycle changes, but the two issues above undermine the promised reviewer recovery/reuse behavior in common timing/legacy cases.

## Regression Risk

Medium: repeated review triggers before reviewer activity metadata lands, or restoring older Claude reviewer sessions, can lose the active terminal/conversation.

## Testing

The PR adds substantial backend and frontend tests, and the PR body reports targeted Go tests plus diff check. I did not run tests per review instructions. Coverage appears to focus on the happy path where agent_session_id exists, which misses the blank/legacy/timing cases called out above.

Verdict: changes_requested

backend/internal/review/review.go
neversettle17-101: Addressed in c1d2f7451. Reviewer pane reuse no longer depends solely on agent_session_id: it still reuses a live matching handle while review runs are running, which preserves the active reviewer during the hook metadata timing window. Idle/completed panes without a native id are still respawned so stale shell panes are not reused.

backend/internal/adapters/reviewer/agentrestore/agentrestore.go
neversettle17-101: Addressed in c1d2f7451. The shared restore helper now always calls the wrapped adapter with Session.ID populated and only omits agentSessionId metadata when blank, so Claude Code can use its deterministic legacy restore fallback for pre-hook reviewer sessions.

backend/internal/review/review.go
```diff
@@ -387,21 +596,32 @@ func (e *Engine) List(ctx stdctx.Context, workerID domain.SessionID) (SessionRev
...
+	}
 	runs, err := e.store.ListReviewRunsBySession(ctx, workerID)
 	if err != nil {
 		return SessionReviews{}, err
 	}
 	var handle string
 	if review, ok, err := e.store.GetReviewBySession(ctx, workerID); err != nil {
 		return SessionReviews{}, err
-	} else if ok {
+	} else if ok && review.Harness == selectedHarness {
```
neversettle17-101: This hides the active reviewer terminal whenever the run was started with the trigger body override instead of a persisted session/default reviewer. Trigger explicitly documents that a non-empty harness is a pass-only override, but List recomputes selectedHarness from persisted session/project config and only returns the handle if it matches. So after a Codex second-opinion trigger on a Claude-default worker, the trigger response opens the terminal, then the next /reviews refetch returns reviewerHandleId empty while that Codex review is still running. Please return the active review row handle/harness when it exists (or otherwise persist the selected override before relying on selectedHarness here).

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -804,7 +839,24 @@ function ReviewsSection({
...
+			const { data, error } = await apiClient.POST("/api/v1/sessions/{sessionId}/reviews/kill", {
+				params: { path: { sessionId: session.id } },
+			});
+			if (error) throw new Error(apiErrorMessage(error, t("inspector.unableKillReviewSession")));
+			return data;
+		},
+		onSuccess: (data) => {
+			setReviewNotice(null);
+			if (data) {
+				queryClient.setQueryData(["session-reviews", session.id], data);
```
neversettle17-101: This stores a KillReviewResponse in the same session-reviews cache that normally holds ListReviewsResponse, but the new kill/restore responses omit the runs array. Immediately after killing (and similarly after restore/switch, which also setQueryData with those responses), ReviewPanel receives runs={[]} and the AO review history tabs disappear until some later full list refetch. Either include runs in the kill/restore response DTOs, or invalidate/refetch instead of replacing the cache with a narrower shape.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker lifecycle: reviewer panes can be hard-killed/restored, are surfaced as selectable terminal tabs, and reviewer native session state is persisted per harness so restored/switched reviewers can resume appropriately. It also adds reviewer kill/restore API routes, storage migrations, generated API types, runtime raw-input support for cancellation, and frontend controls/tests for reviewer tabs and harness selection.

## Summary of Changes

- Adds review_session storage plus agent_session_id persistence, per-harness review-run uniqueness, and store/query helpers for reviewer handles by harness.
- Extends the review engine and session manager to destroy reviewer panes on worker kill, restore idle reviewer terminals, handle reviewer harness switches, and cancel stale running runs.
- Adds launcher support for fresh vs restored reviewer terminals, read-only restore commands for Claude/Codex/OpenCode, and harness-specific cancel behavior.
- Adds /reviews/kill and /reviews/restore routes plus OpenAPI/frontend schema updates.
- Updates the frontend to show reviewer terminal tabs, kill reviewer sessions, switch reviewer harnesses, and display per-harness review history.

## Findings

- **Medium**: One-shot reviewer overrides lose the running terminal on the next list/refetch because List hides active handles that do not match the persisted/default reviewer.
- **Medium**: Kill/restore responses are written into the session-reviews cache but omit the runs field that the review panel uses, so AO review history disappears until a full refetch happens.

## Description Match

The implementation broadly matches the lifecycle-tab intent, but the two issues above leave gaps in terminal visibility/cache consistency for supported flows.

## Regression Risk

Medium around reviewer visibility and review-history UI after lifecycle actions or per-trigger reviewer overrides.

## Testing

The PR adds backend and frontend tests, and the PR body reports targeted Go tests, diff check, screenshots, and manual videos. I did not run tests per review instructions. The missing cases are cache shape after kill/restore and List behavior for active one-shot override reviewer handles.

Verdict: changes_requested

backend/internal/review/review.go
neversettle17-101: Addressed in a7a902eb6.

`Engine.List` now reports the active review row handle and harness whenever a reviewer pane is live, even when that run came from a one-shot trigger override and does not match the persisted/default reviewer. Added `TestListReturnsActiveOneShotOverrideHandle` to cover the refetch/list case.

frontend/src/renderer/components/SessionInspector.tsx
neversettle17-101: Addressed in a7a902eb6.

Kill and restore responses now include the full `runs` array from `List`, and the generated OpenAPI/frontend schema was updated. Backend HTTP tests assert `runs` is present in both responses, and frontend tests assert the session-reviews cache keeps review history after kill and restore.

backend/internal/review/review.go
```diff
@@ -244,7 +293,7 @@ func (e *Engine) Trigger(ctx stdctx.Context, workerID domain.SessionID, override
 		// another agent is precisely a request for a second opinion on this commit,
 		// so refusing it makes the reviewer choice inert exactly when it is most
 		// useful. Ineligible PRs stay excluded: nothing can review those.
-		eligible := reviewState.Status == ReviewStateNeedsReview || reviewState.Status == ReviewStateChangesRequested
+		eligible := reviewState.Status == ReviewStateNeedsReview
```
neversettle17-101: This drops `ReviewStateChangesRequested` from the set of states that an explicit trigger will run again. On main, changes-requested current heads are eligible so that after the worker addresses AO feedback, pressing Run/Rerun starts a fresh pass for the same reviewer/commit. With this change the trigger returns `Created:false` and reuses the old changes_requested run, so the reviewer never re-checks the fix unless the PR head SHA changes or the user switches harness. Please keep changes_requested eligible for the same harness.

backend/internal/storage/sqlite/migrations/0048_review_agent_session_id.sql
```diff
@@ -0,0 +1,59 @@
...
+      FROM review_run
+      WHERE target_sha != ''
+        AND status NOT IN ('failed', 'cancelled')
+    )
+    WHERE rn = 1
+  );
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE UNIQUE INDEX idx_review_run_session_pr_sha_harness
```
neversettle17-101, line 61: Even if the trigger logic is restored, this new unique index blocks inserting another same-harness run for a completed `changes_requested` review because every non-failed/non-cancelled row is now indexed. The previous partial index intentionally excluded completed changes_requested rows so they could be retried. Please preserve that predicate (or otherwise exclude completed changes_requested rows) so the retry path can create a new running pass.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker lifecycle: reviewer panes can be hard-killed/restored, are surfaced as selectable terminal tabs, and reviewer native session state is persisted per harness so restored/switched reviewers can resume appropriately. It also adds reviewer kill/restore API routes, storage migrations, generated API types, runtime raw-input support for cancellation, and frontend controls/tests for reviewer tabs and harness selection.

## Summary of Changes

- Adds review_session storage plus agent_session_id persistence, per-harness review-run uniqueness, and store/query helpers for reviewer handles by harness.
- Extends the review engine and session manager to destroy reviewer panes on worker kill, restore idle reviewer terminals, handle reviewer harness switches, and cancel stale running runs.
- Adds launcher support for fresh vs restored reviewer terminals, read-only restore commands for Claude/Codex/OpenCode, and harness-specific cancel behavior.
- Adds /reviews/kill and /reviews/restore routes plus OpenAPI/frontend schema updates.
- Updates the frontend to show reviewer terminal tabs, kill reviewer sessions, switch reviewer harnesses, and display per-harness review history.

## Findings

- **Medium**: Same-harness re-runs after a changes_requested review are no longer possible, which regresses the existing retry flow after the worker addresses review feedback.

## Description Match

The lifecycle/tab work is broadly aligned with the PR description, but the changes-requested retry regression should be fixed before merge.

## Regression Risk

Medium: workers that receive AO changes_requested feedback cannot get a fresh review from the same reviewer on the same PR head after making additional changes without changing the commit SHA or switching harness.

## Testing

The PR adds backend and frontend tests, and the PR body reports targeted Go tests, diff check, screenshots, and manual videos. I did not run tests per review instructions. The current tests appear to pin the new reuse behavior for changes_requested, but that behavior conflicts with the existing trigger contract and user flow.

Verdict: changes_requested

backend/internal/storage/sqlite/migrations/0048_review_agent_session_id.sql
neversettle17-101: Addressed in adcd3f0c2.

Migration `0043_review_agent_session_id.sql` now preserves the previous partial unique-index predicate: duplicates are blocked for running/no-verdict and terminal non-changes-requested rows, but completed `changes_requested` rows remain outside the index so same-harness retries can insert. Updated SQLite store tests to cover changes_requested retry inserts.

backend/internal/review/review.go
neversettle17-101: Addressed in adcd3f0c2.

`Trigger` now treats `ReviewStateChangesRequested` as eligible again, so pressing Run/Re-run after AO requested changes creates a fresh same-harness review pass for the current PR head. Added/updated the engine regression test to assert a changes_requested current head creates a new run and notifies the existing reviewer pane.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -778,16 +808,21 @@ function ReviewsSection({
 			setReviewNotice(null);
 		},
 		onSuccess: ({ data, reused }) => {
-			void queryClient.invalidateQueries({ queryKey: ["session-reviews", session.id] });
+			if (data) {
+				queryClient.setQueryData(["session-reviews", session.id], data);
```
neversettle17-101: This writes a TriggerReviewResponse into the `session-reviews` query cache, but that cache is otherwise a ListReviewsResponse and the panel now renders AO history from `reviewsQuery.data?.runs`. TriggerReviewResponse still only contains `reviewerHandleId`, `reviews`, and `created`, so after pressing Run/Rerun (especially a 200 reuse, which will not poll because nothing is running) the existing AO review history becomes `[]`. Either include `runs` in the trigger response too, or keep the old invalidate/refetch behavior instead of replacing the cache with the narrower response.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker lifecycle: reviewer panes can be hard-killed/restored, are surfaced as selectable terminal tabs, and reviewer native session state is persisted per harness so restored/switched reviewers can resume appropriately. It also adds reviewer kill/restore API routes, storage migrations, generated API types, runtime raw-input support for cancellation, and frontend controls/tests for reviewer tabs and harness selection.

## Summary of Changes

- Adds review_session storage plus agent_session_id persistence, per-harness review-run uniqueness, and store/query helpers for reviewer handles by harness.
- Extends the review engine and session manager to destroy reviewer panes on worker kill, restore idle reviewer terminals, handle reviewer harness switches, and cancel stale running runs.
- Adds launcher support for fresh vs restored reviewer terminals, read-only restore commands for Claude/Codex/OpenCode, and harness-specific cancel behavior.
- Adds /reviews/kill and /reviews/restore routes plus OpenAPI/frontend schema updates.
- Updates the frontend to show reviewer terminal tabs, kill reviewer sessions, switch reviewer harnesses, and display per-harness review history.

## Findings

- **Medium**: Trigger responses are cached as ListReviewsResponse but do not include the `runs` data the panel now relies on, so AO review history disappears after trigger/reuse.

## Description Match

The implementation is broadly aligned with the PR description, but the cache/response shape mismatch breaks part of the updated review-history UI.

## Regression Risk

Medium: review history can vanish from the inspector after triggering or reusing a review until a later refetch, and for reused non-running responses there may be no polling refetch to repair it.

## Testing

The PR adds backend and frontend tests, and the PR body reports targeted Go tests, diff check, screenshots, and manual videos. I did not run tests per review instructions. The missing case is the trigger/reuse cache shape with existing review `runs`.

Verdict: changes_requested

frontend/src/renderer/components/SessionInspector.tsx
neversettle17-101: Addressed in c50569622.

Trigger responses now include `runs`, matching the List/Kill/Restore cache shape. `TriggerResult` carries the existing run history for 200 reuse and prepends newly created runs for 201 responses. The OpenAPI/frontend schema was regenerated, backend HTTP tests assert the `runs` field is present, and frontend tests assert both re-run and 200 reuse keep review history in the `session-reviews` cache.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker lifecycle: reviewer panes can be hard-killed/restored, are surfaced as selectable terminal tabs, and reviewer native session state is persisted per harness so restored/switched reviewers can resume appropriately. It also adds reviewer kill/restore API routes, storage migrations, generated API types, runtime raw-input support for cancellation, and frontend controls/tests for reviewer tabs and harness selection.

## Summary of Changes

- Adds review_session storage plus agent_session_id persistence, per-harness review-run uniqueness, and store/query helpers for reviewer handles by harness.
- Extends the review engine and session manager to destroy reviewer panes on worker kill, restore idle reviewer terminals, handle reviewer harness switches, cancel stale running runs, and keep trigger/list responses cache-compatible.
- Adds launcher support for fresh vs restored reviewer terminals, read-only restore commands for Claude/Codex/OpenCode, and harness-specific cancel behavior.
- Adds /reviews/kill and /reviews/restore routes plus OpenAPI/frontend schema updates.
- Updates the frontend to show reviewer terminal tabs, kill reviewer sessions, switch reviewer harnesses, and display per-harness review history.

## Findings

No blocking findings from the reviewed diff.

## Description Match

The implementation appears aligned with the PR description and addresses the reviewer lifecycle/tab behavior described.

## Regression Risk

Moderate overall due to lifecycle, storage, and UI/API surface area changes, but I did not find a concrete regression in this pass.

## Testing

The PR adds backend and frontend test coverage, and the PR body reports targeted Go tests, diff check, screenshots, and manual videos. I did not run tests per review instructions. Frontend typecheck was noted as not run because dependencies were absent.

Verdict: approved

backend/internal/review/launcher.go
```diff
@@ -176,61 +194,158 @@ func (l *agentLauncher) prepareInvocation(spec LaunchSpec) (ports.ReviewInvocati
...
-		return base
-	}
-	env := make(map[string]string, len(base)+1)
+func (l *agentLauncher) runtimeEnv(ctx context.Context, spec LaunchSpec, argv []string, base map[string]string) map[string]string {
+	env := make(map[string]string, len(base)+3)
 	for k, v := range base {
 		env[k] = v
 	}
-	env["PATH"] = path
+	env[sessionmanager.EnvSessionID] = string(spec.WorkerID)
```
neversettle17-101: **High — Do not identify the reviewer runtime as the worker session.** `ao hooks` reads `AO_SESSION_ID` and posts both activity and the native `agentSessionId` to `/sessions/{AO_SESSION_ID}/activity`. With this environment, Claude/Codex reviewer startup and activity hooks therefore update the owning worker record. In particular, a reviewer SessionStart can overwrite the worker native agent session id with the reviewer conversation id, so restoring the worker later may resume the wrong conversation; reviewer activity also makes the worker appear active. Give reviewer hooks a reviewer-specific identity/endpoint (or suppress `AO_SESSION_ID` and route their native id into review persistence separately), while retaining separate worker context for `ao review submit`. Add an integration-style test that emits a reviewer hook and proves the worker session metadata/activity is unchanged.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminals with worker lifecycle: reviewer panes survive recovery-oriented shutdown, are destroyed on explicit worker kill, can restore native reviewer conversations, and appear as selectable frontend tabs. It also adds per-harness reviewer persistence, review kill/restore API surfaces, cancellation behavior for supported harnesses, generated API artifacts, migrations, and backend/frontend tests.

## Findings

- **High**: Reviewer hooks are attributed to the worker session, corrupting worker lifecycle metadata. See inline comment.

## Description Match

The broad lifecycle and UI behavior matches the description, but the shared session environment causes reviewer hook events to mutate the owning worker, so the implementation is not safe to merge as-is. **Changes requested.**

## Regression Risk

Reviewer startup/session-start hooks can overwrite the worker native agent session id and activity state. A later worker restore can then attempt to resume the reviewer conversation instead of the worker conversation.

## Testing

The PR adds substantial unit and controller coverage and reports targeted backend tests plus diff checking. Frontend typecheck/build were not run, and current tests do not cover hook attribution from a real reviewer runtime. Tests were not run as part of this review.

backend/internal/review/launcher.go
neversettle17-101: Addressed in 8d53405b1.

Reviewer runtimes no longer receive AO_SESSION_ID. The launcher now sets AO_REVIEW_SESSION_ID, AO_REVIEW_WORKER_SESSION_ID, and AO_REVIEW_HARNESS, and `ao hooks` routes AO_REVIEW_SESSION_ID to the new `/api/v1/reviews/{reviewSessionID}/activity` endpoint before considering worker AO_SESSION_ID. AO_REVIEW_WORKER_SESSION_ID is context only and cannot route activity writes.

The review activity endpoint persists reviewer-owned native session metadata to the per-harness review row (`review.agent_session_id`) and does not update worker session lifecycle/activity/native metadata. Reviewer launch/restore now uses the review row native id for reviewer restore.

Added coverage for reviewer hook routing precedence, worker fallback routing, ignored worker-context-only reviewer env, review activity persistence, reviewer launcher env without AO_SESSION_ID, and separate per-harness native reviewer ids.

neversettle17-101 (commented):
## What This PR Does

This PR aligns reviewer terminal lifecycle with worker lifecycle, persists per-harness reviewer/native-session state, adds reviewer kill/restore and cancellation flows, and exposes reviewer terminals as selectable frontend tabs.

## Findings

No blocking findings remain. The follow-up commit addresses the prior hook-attribution bug by removing `AO_SESSION_ID` from reviewer runtimes, assigning a stable review-session id, routing reviewer hooks to a dedicated review activity endpoint, and persisting the native reviewer conversation id without mutating worker activity or metadata.

## Description Match

The implementation now appears aligned with the stated lifecycle and restoration behavior. **Ready / approved.**

## Regression Risk

The main lifecycle risk was cross-contamination between reviewer and worker hook state; the new route and tests directly cover that separation. Remaining risk is primarily real-harness behavior across Electron/daemon restart and terminal restoration.

## Testing

The update adds CLI routing, controller, service, storage, and launcher coverage for reviewer-specific hook handling, including precedence over a worker session id. The PR reports targeted backend tests and diff checking. Frontend typecheck/build and end-to-end desktop restore testing were not run according to the PR description. Tests were not run as part of this review.

neversettle17-101 (commented):
Ready to merge.

## What This PR Does
Aligns reviewer terminal lifecycle with worker sessions: explicit worker kills tear down reviewer panes, shutdown save/teardown preserves them for recovery, worker restore can relaunch reviewer terminals, and the UI exposes reviewer terminals as tabs.

## Summary of Changes
- Adds reviewer restore/terminate paths through the review engine/service/controller and wires worker session kill/restore to those hooks.
- Persists reviewer terminal/agent session metadata per harness and adds migrations/query updates for review handle/session recovery.
- Updates reviewer adapters and launcher restore/cancel behavior across Codex, Claude Code, and OpenCode.
- Adds frontend terminal-tab support for reviewer panes plus related i18n and tests.
- Adds/updates backend and frontend tests around lifecycle, restore, cancellation, and tab rendering.

## Findings
No blocking findings.

## Description Match
The implementation appears aligned with the PR description: worker termination hard-clears reviewer handles/running runs, shutdown teardown intentionally avoids reviewer termination, restore attempts reviewer restoration, and the UI can select reviewer panes.

## Regression Risk
No clear regression found from the reviewed context. Residual risk is mainly lifecycle integration complexity across multiple harnesses and persisted historical review rows.

## Testing
I did not run tests, per review instructions. The PR includes targeted Go tests for the review/session-manager/service/controller paths and React tests for reviewer terminal tab behavior. The PR notes frontend typecheck was not run because dependencies are absent in the worktree.

illegalcall (commented):
Thanks for separating reviewer hooks from the worker — that closes the worker-state corruption path we reproduced. I took another pass over the latest head and found five remaining lifecycle/persistence issues that look merge-blocking. The highest-risk one is migration compatibility: I can reproduce a successful startup followed by a hard failure on the first reviewer write for a database shape already represented in this repo. The other comments cover cross-harness identity, shutdown/restore, concurrency, and harness-switch atomicity.

For context, the focused backend tests (including race/API), frontend typecheck, and 144 related frontend tests pass. These cases fall between the current test seams, so I included concrete failure sequences and suggested coverage inline.

backend/internal/storage/sqlite/migrations/0049_review_per_harness.sql
```diff
@@ -0,0 +1,156 @@
+-- Store reviewer lifecycle state directly in one review row per worker and
+-- harness. Migration 0048 shipped with review UNIQUE(session_id) plus a
+-- companion review_session table, so this must use a fresh migration version:
```
illegalcall: I don't think 0049 is a safe fresh version here. `migrations_field_profiles_test.go` already models deployed databases whose `goose_db_version` contains versions 40–51, so those installations skip both 0048 and 0049. I reproduced that profile on this head: migration reports success, but the first current review write fails with `table review has no column named agent_session_id`. The same applies to anyone who ran the PR's earlier version 49. Could we leave the historical slots immutable, put the required rebuild in a new version above the current migration tip, and add a burned-version test that actually performs a reviewer write?

backend/internal/adapters/reviewer/claudecode/claudecode.go
```diff
@@ -91,7 +94,7 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
 	if err != nil {
 		return ports.ReviewCommandSpec{}, err
 	}
-	return ports.ReviewCommandSpec{Argv: argv}, nil
+	return ports.ReviewCommandSpec{Argv: argv, AgentSessionID: workeragent.SessionUUID(inv.ReviewerID)}, nil
```
illegalcall: This records a native session ID that this launch never gives Claude. The launch config above intentionally sets `SessionID` to empty, so argv has no `--session-id`, while this returns the UUID derived from the AO reviewer handle. Hooks could eventually correct it, but the reviewer launcher never installs the selected reviewer's `GetAgentHooks`; for example, a Codex worker with a Claude reviewer has no workspace-local Claude hook. Kill/restore can therefore call `claude --resume` with an ID that never existed. Could reviewer prelaunch install the selected harness's hooks and either let the hook own `AgentSessionID`, or pass this same pinned ID to Claude?

backend/internal/session_manager/manager.go
```diff
@@ -1629,7 +1680,9 @@ func (m *Manager) saveAndTeardownOne(ctx context.Context, rec domain.SessionReco
 		return fmt.Errorf("save %s: upsert worktree row: %w", rec.ID, err)
 	}
 
-	// 3. Mark terminal via the LCM (same path Kill uses).
+	// 3. Mark terminal via the LCM (same path Kill uses). Do not terminate the
+	// reviewer here: shutdown save/teardown is recovery-oriented lifecycle, not
```
illegalcall: Keeping the reviewer identity across shutdown makes sense, but keeping this live process while step 5 force-removes its cwd does not. The reviewer pane can stay alive with its cwd pointing at the deleted worktree; after restart, `RestoreReviewer` sees the handle as alive and returns without recreating it in the restored worktree. On Windows, the open cwd may also block removal. Could shutdown destroy and clear only the reviewer runtime while retaining the review row and native agent session ID, then let worker restore recreate the pane? A shutdown → daemon restart → worker restore test would make this durable-session contract explicit.

backend/internal/review/review.go
```diff
@@ -452,6 +697,93 @@ func (e *Engine) Cancel(ctx stdctx.Context, workerID domain.SessionID) (CancelRe
...
+		}
+	}
+	return e.store.GetReviewBySession(ctx, workerID)
+}
+
+// TerminateReviewer destroys the live reviewer pane for a worker and cancels
+// any running review runs. Unlike Cancel, this does not ask the reviewer
+// adapter for a graceful interrupt sequence: worker termination/restore must
+// remove the terminal pane itself.
+func (e *Engine) TerminateReviewer(ctx stdctx.Context, workerID domain.SessionID, body string) (TerminateResult, error) {
```
illegalcall, line 813: Can this participate in the same per-worker lifecycle serialization as `Trigger`? At the moment, `Trigger` is serialized only against another trigger. A worker kill can list/destroy/cancel while a trigger that already read the worker as live is between run insertion and `Spawn`; after teardown returns, that trigger can still create and persist `review-<worker>` against the worktree being removed. I think Trigger, Terminate, and Restore need one operation-wide lock (or an equivalent durable lifecycle fence), plus a deterministic interleaving test.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -1167,13 +1172,28 @@ function ReviewsSection({
 	}, [session.id, session.reviewerHarness]);
 	const saveReviewer = useMutation({
 		mutationFn: async (harness: ReviewerHarness | "") => {
+			const activeReviewerHandle = reviewsQuery.data?.reviewerHandleId?.trim();
```
illegalcall: This turns one reviewer-harness transition into three independent client requests: kill, persist preference, then restore. If the PUT fails, the old reviewer is already gone; if restore fails, the preference changed but no terminal exists; worker teardown or a trigger can also interleave between the calls. It also puts daemon-owned lifecycle policy in React. Could this be one daemon service/endpoint operation that serializes the switch and returns the authoritative reviewer handle and harness? The UI could then make one mutation and apply one response.

backend/internal/storage/sqlite/migrations/0049_review_per_harness.sql
neversettle17-101: Addressed in 03cf34c88.

0049 is now an inert historical placeholder, and the actual per-harness review-table rebuild moved to fresh migration 0080 above the current tip. Startup also prepares burned 0048/0049 profiles just enough for 0080 to run, without recreating review_session after 0080 applies.

Added burned-version coverage that performs a real reviewer write/read through the SQLite store and verifies review.agent_session_id persists on that field profile.

backend/internal/adapters/reviewer/claudecode/claudecode.go
neversettle17-101: Addressed in 03cf34c88.

Claude reviewer fresh launch now passes the same deterministic reviewer UUID as SessionID that it returns as AgentSessionID, so restore never persists an id the process was not launched with. Reviewer PreLaunch also installs the selected reviewer harness hooks before launch, so a Codex worker with a Claude reviewer gets Claude reviewer hooks installed.

Added tests for pinned id consistency and selected-reviewer hook prelaunch.

backend/internal/session_manager/manager.go
neversettle17-101: Addressed in 03cf34c88.

SaveAndTeardownAll now calls a recovery-oriented reviewer terminal teardown before worker runtime/worktree removal. That path destroys reviewer panes and clears reviewer_handle_id while preserving review rows and native reviewer agent_session_id, so worker restore can recreate the pane in the restored checkout.

Added session-manager coverage for shutdown teardown ordering and SaveAndTeardownAll -> RestoreAll reviewer restore.

backend/internal/review/review.go
neversettle17-101: Addressed in 03cf34c88.

Reviewer Trigger, TerminateReviewer, RestoreReviewer, TeardownReviewerTerminal, and SwitchReviewer now share the same per-worker lock. Terminate can no longer interleave between run insertion and Spawn.

Added a deterministic interleaving test where Trigger blocks inside Spawn, Terminate starts concurrently, and the test proves Destroy cannot happen until Trigger completes and then cancels/clears the created run/handle.

frontend/src/renderer/components/SessionInspector.tsx
neversettle17-101: Addressed in 03cf34c88.

Reviewer harness switching is now one daemon-owned operation: POST /api/v1/sessions/{sessionId}/reviews/switch. The review engine serializes the preference update, old reviewer teardown, optional selected-reviewer restore, and final List-style response under the same per-worker lifecycle lock.

The React mutation now makes that single request and writes the authoritative ListReviewsResponse-shaped payload into the session-reviews cache. Added controller/API tests and regenerated OpenAPI/frontend schema.

illegalcall (approved):

PR #3488, opened 2026-08-03
Fix URL editing in macOS maximized browser view

aprv10:
## What

Keep the browser URL field selectable and editable when the browser panel is maximized on macOS.

## Why

The macOS shell uses a hidden native titlebar with a traffic-light drag region. The maximized browser is portaled to `document.body` and originally rendered from `y=0`, allowing its URL toolbar to overlap that native drag region. Electron treated pointer input there as window dragging instead of delivering it to the URL input.

Windows is unaffected because its Window Controls Overlay safe area is already applied.

## How

- Reserve the existing 40px macOS titlebar band in windowed mode.
- Mark the maximized browser surface and toolbar controls as `-webkit-app-region: no-drag`.
- Skip the inset during native macOS fullscreen, where traffic lights are hidden.
- Add regression tests for URL editing and fullscreen/windowed overlay behavior.

## Files affected

- `frontend/src/renderer/components/SessionView.tsx`
- `frontend/src/renderer/styles.css`
- `frontend/src/renderer/components/BrowserPanel.test.tsx`
- `frontend/src/renderer/components/SessionView.test.tsx`

## Testing

- `npm test -- --run src/renderer/components/SessionView.test.tsx src/renderer/components/BrowserPanel.test.tsx` — 55 tests passed.
- `npm run typecheck` — passed.
- Manual testing on macOS remains pending; automated regression tests and typecheck pass.

## Checklist

- [x] Branched from `main`
- [x] One focused change
- [x] Follows repository conventions
- [x] Relevant CI checks pass

### DIFF
--- a/frontend/src/renderer/components/BrowserPanel.test.tsx
+++ b/frontend/src/renderer/components/BrowserPanel.test.tsx
@@ -188,6 +188,17 @@ describe("BrowserPanel", () => {
 		expect(hookState.navigate).toHaveBeenCalledWith("localhost:5173");
 	});
 
+	it("keeps the URL input editable while the browser is maximized", async () => {
+		hookState.navState = { ...hookState.navState, url: "http://localhost:5173/" };
+		render(<BrowserPanel active onTogglePopOut={() => undefined} poppedOut session={session} />);
+		const input = screen.getByRole("textbox", { name: /browser url/i });
+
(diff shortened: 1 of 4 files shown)


illegalcall (approved):
Verified on mac, works.

Issue #3476, opened 2026-08-03
Windows nightly: Board cannot load sessions when migration 0040 is skipped by prior goose version


somewherelostt:
## Bug
The desktop Board can show only `Could not load sessions.` on Windows nightly because the daemon's session list endpoint fails with HTTP 500 when the local SQLite database has a previously-recorded goose migration version `40` but lacks the new `sessions.diff_base_sha` / `sessions.diff_base_ref` columns.

**Source:** Local device triage from a website-installed Windows nightly; screenshots supplied in Codex chat.
**Reported by:** @somewherelostt
**Analyzed against installed app:** `v0.11.2-nightly.202608021445` / tag `9159a0206a2e1d2a99333118bf9ebc5590b7404f`
**Current main checked:** `dd8f6e0b0321253b2e0a0ea18322c3de2397cf49`
**Confidence:** High
**Priority:** P1: core Board/session workflow is broken with no in-app workaround.

## Reproduction
On an affected Windows install/profile:

1. Install/run Agent Orchestrator nightly `v0.11.2-nightly.202608021445` from the website.
2. Use an existing `~/.ao/data/ao.db` whose `goose_db_version` already contains version `40` from an earlier shipped migration history, but whose `sessions` table does not contain `diff_base_sha` or `diff_base_ref`.
3. Open the desktop app Board.
4. Observe `Could not load sessions.`
5. Query the daemon directly:

```powershell
& "$env:LOCALAPPDATA\Programs\agent-orchestrator\resources\daemon\ao.exe" status
& "$env:LOCALAPPDATA\Programs\agent-orchestrator\resources\daemon\ao.exe" session ls
Invoke-WebRequest -UseBasicParsing http://127.0.0.1:3001/api/v1/sessions
```

Observed locally:

```text
AO daemon: ready
  pid: 30884
  port: 3001
  data dir: C:\Users\abuma\.ao\data
  healthz: ok
  readyz: ready

GET /api/v1/sessions -> 500 INTERNAL_ERROR
Internal server error (INTERNAL_ERROR) [request Somewherelostt/E4BRzyHpFn-000031]
```

The database has sessions and projects, so the board is not empty:

```text
select count(*) from sessions; -- 47
```

## Root Cause
The installed code's generated session queries select the new diff-base columns:

- `backend/internal/storage/sqlite/gen/sessions.sql.go:132` defines `ListAllSessions`.
[...]

PR #2412, opened 2026-07-04
feat(session): switch a session's agent mid-flight

Pulkit7070:
## What & why

A session's agent (`harness`) is currently fixed at spawn (`SessionRecord.Harness`, resolved once by `effectiveHarness` in `session_manager`) and can never change. Mixing agents on a single task — e.g. **codex** to test, **claude-code** to write, a cheaper model for cost — is impossible without throwing the work away and respawning.

This PR lets you **change a session's agent in place**, keeping the same git worktree (all code + uncommitted work preserved). The new agent launches **fresh** — there is no native resume, since a different harness cannot read the outgoing agent's session. An optional `model` override rides the same path.

## Behaviour

Two cases are handled:

- **Live session → swap in place.** The old agent is torn down only *after* the new launch command validates, so a bad/unknown harness never disrupts the running session. A `BeginSwitch`/`EndSwitch` guard on the lifecycle manager makes the reaper ignore the brief window where the old runtime is gone and the new one isn't up yet — otherwise a "dead" probe would wrongly terminate the session mid-switch.
- **Terminated session → relaunch-as.** When an agent exits (e.g. codex ends its process on completion) the session is marked terminated. Switching such a session restores its worktree and launches the chosen agent fresh under it — the way to bring a finished task back under a *different* agent (plain restore keeps the original harness). Only fully `merged` sessions stay locked.

`lifecycle.MarkSwitched` atomically changes the persisted harness, points at the new runtime handle, and **clears the harness-specific `AgentSessionID`** (which `MarkSpawned`'s metadata merge cannot, since it only sets non-empty fields), resetting activity/first-signal so the new agent re-proves its hook pipeline.

## Surfaces

- **Backend:** `sessionmanager.SwitchHarness` (live + terminated paths) and `lifecycle` guard + `MarkSwitched`.
- **API:** `POST /sessions/{id}/switch` `{harness, model?}` — `400` unknown harness, `404` unknown session, `409` switch-in-progress. OpenAPI spec + `frontend/src/api/schema.ts` regenerated (pinned `openapi-typescript@7.4.4`).
- **CLI:** `ao session switch <id> --harness <agent> [--model ...]`.
- **UI:** the session inspector's **Overview → Agent** row is now a dropdown that switches the agent in place (or relaunches a terminated one), built from the existing `dropdown-menu` primitive and `AGENT_OPTIONS`; `merged` sessions stay read-only.

## Tests

- live swap clears `AgentSessionID`
[...]

PR #3491, opened 2026-08-03
fix: stop stale-PID group kills, misread liveness probes, and burned-migration 500s from wiping the board

### LINKED ISSUE #3475: Board loses all sessions: stale-PID group kill wipes tmux, misread probe archives everything, and skipped migration 0040 500s the session list
Two independent defects both end with the user's sessions gone from the Board. They have different mechanisms and different fixes, but they are being tracked together here so one PR closes both. The migration half was originally reported in #3476.

Quick triage test to tell them apart:

| | Defect A | Defect B (#3476) |
|---|---|---|
| `GET /api/v1/sessions` | 200, sessions returned | 500 `INTERNAL_ERROR` |
| UI | Board silently empty | `Could not load sessions.` banner |
| DB | sessions flipped to `is_terminated=1` | sessions untouched, `is_terminated=0` |
| Recovery | `ao session restore` | schema repair |

---

# Defect A: tmux server killed, then every session archived

Two bugs in a chain, both introduced by #3066. The first destroys the tmux server, the second turns that into a full board wipe instead of a recoverable outage.

## A1. Trigger: preview server group-kills a PID it no longer owns

The `previewserver` package has exactly one commit in its history: `3279c3a0b` ("Give every AO worker a live, session-owned browser", #3066, 2026-07-30). `manager.go` and `process_unix.go` have no prior history, so none of this code existed before that PR.

It introduced a process-group kill by raw PID:

```go
// process_unix.go
func forceKillPreviewProcess(cmd *exec.Cmd) error {
	err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
	...
}
```

and calls it after the process has already been reaped:

```go
// manager.go:371
select {
case <-done:
	// The root may exit before descendants that ignored SIGTERM. Escalate
	// against the owned process tree even after Wait has completed.
	_ = forceKillPreviewProcess(cmd)
```

`run.done` is closed in `waitForExit` (`manager.go:495`) after `run.cmd.Wait()` (`manager.go:469`). Once `Wait` returns, the PID is released back to the OS and is no longer owned by the daemon. The negative argument makes this a process-group kill, so the blast radius is an entire group rather than one stray process.

Four sites do this:

- `manager.go:474` in `waitForExit`, immediately after `Wait()` returns, on any unexpected preview exit
- `manager.go:375`, the `case <-done` shown above
- `manager.go:797` and `:799` in `forceStopAfterGrace`
- `manager.go:829` in `reapPersistedProcesses`, the worst form: at daemon startup it reads the previous daemon generation's PIDs from `preview-processes.json` and group-kills each one, with no check that the PID still refers to a preview process

### Why it lands on tmux

A tmux server is always a process group leader, PGID equal to its own PID. Verified locally on two different servers (`14618/14618`, `4940/4940`), both `PPID 1`, own session. So a recycled PID landing on a tmux server means `kill(-pid, SIGKILL)` takes down the server and every session inside it at once.

Measured PID allocation on the affected machine is roughly 24/sec, so the PID space wraps about every 67 minutes. One affected tmux server had been alive 110 minutes. The same PR raises this rate by giving every worker a session-owned browser and preview, so the risky path now runs continuously instead of occasionally.

SIGKILL produces no crash report and no unified-log entry, which matches the forensics: the server vanished with the daemon still running and no trace in `DiagnosticReports`, `log show`, or `memorystatus`.

## A2. Amplifier: server-level probe failures misread as per-session death

`backend/internal/adapters/runtime/tmux/tmux.go:878`

```go
func sessionMissingOutput(out string) bool {
	s := strings.ToLower(out)
	return strings.Contains(s, "can't find session") ||
		strings.Contains(s, "no server running") ||   // server-level, not session-level
		strings.Contains(s, "error connecting") ||    // transient socket error
		strings.Contains(s, "session not found")
}
```

`IsAlive` (`tmux.go:497`) returns `(false, nil)` for these, so the reaper records a confident `ProbeDead` (`observe/reaper/reaper.go:166`) and `ApplyRuntimeObservation` sets `IsTerminated = true` (`lifecycle/manager.go:248`).

`no server running` and `error connecting` say nothing about any individual session. The first means the whole server is gone, the second is a transient socket failure that may not mean anything is dead at all.

The codebase already has the right guard and this path bypasses it:

- `ports/runtime_observations.go:9` : "`failed` means the probe errored or timed out and is never treated as a death conclusion"
- `observe/reaper/reaper.go:158` : "a transient tmux outage [must not] hide a really-dead session, and a transient adapter bug [must not] terminate a really-alive one"

## Observed

Local nightly `0.11.2-nightly.202608021445`. Filtering `change_log` to this signature (28 to 32 sessions terminated in one pass at a uniform 0.118 to 0.140s cadence, which is the per-session `tmux has-session` probe loop):

```
2026-08-02 05:05   28   0.123
2026-08-02 09:30   32   0.118
2026-08-03 03:14   28   0.118
2026-08-03 07:25   28   0.139
2026-08-03 09:16   29   0.140
```

Five occurrences, none before Aug 2, consistent with #3066 landing Jul 30 and shipping to the nightly channel. Earlier bulk terminations in the history are a different mechanism: they write at a 4 to 6 ms median gap, which is a single transaction and far faster than a probe loop that shells out per session.

The sidebar filters `isTerminated` (`renderer/components/Sidebar.tsx:476`), so the board reads as empty and the sessions appear lost.

`iitb-cli-1` confirms the 60s window is the only thing gating the sweep: it emitted an activity hook at 12:55:27.908 and was terminated at 12:56:28.014, 60.1s later, matching `defaultRecentActivityWindow` in `lifecycle/runtime.go:10`.

No data is lost. Worktrees, branches, and agent context stay intact, and `ao session restore` brings sessions back with full context.

---

# Defect B: skipped migration 0040 makes the session list 500

Originally filed as #3476 with full Windows reproduction, screenshots, and logs. Summary here so one PR can close both.

The daemon's generated queries select `sessions.diff_base_sha` and `sessions.diff_base_ref` (`storage/sqlite/gen/sessions.sql.go:138`, called from `store/session_store.go:191`). Those columns are added by `migrations/0040_add_session_diff_base.sql`.

On affected profiles `goose_db_version` already has version `40` recorded from an earlier shipped migration history, so goose skips `0040` and the columns are never created. Every session list then fails:

```
GET /api/v1/sessions -> 500 INTERNAL_ERROR
sqlite3 ~/.ao/data/ao.db "select diff_base_sha from sessions limit 1;"
Error: in prepare, no such column: diff_base_sha
```

`/healthz`, `/readyz`, and `/api/v1/projects` still succeed, so the daemon looks healthy while the core workflow is dead.

## Scope of the burned range is still unconfirmed

The collision is not conceptually limited to version 40. Once a version is recorded in `goose_db_version`, the real migration with that number is skipped silently, and the failure surfaces only later as an opaque 500 from a generated query.

The affected Windows profile has versions **40 through 46** recorded applied on 2026-07-23. Upstream currently ships migrations only up to `0040`.

**Correction to an earlier version of this issue:** I previously cited a macOS profile with versions 40 through 51 as a second independent reproduction. That was wrong. That machine burned 40-51 by running local `pipelines` feature branches, which define their own `0040_pipeline_tables.sql` through `0051_pipeline_run_number.sql`, with apply timestamps matching that branch work. It is self-inflicted by running unmerged branches and is not evidence about released nightlies. It has been retracted on #3476.

That leaves the load-bearing open question:

**How did a website-installed Windows nightly end up with versions 40 through 46 recorded?**

If that profile only ever ran released upstream nightlies, a released build shipped migrations in that range which were later renumbered or removed, which is a genuine release problem and P1 is right. If it also ran a fork or branch build, it is the same self-inflicted case and the priority changes substantially. This should be confirmed before the repair is designed, since it decides whether real users need a shipped repair migration or whether a guardrail plus documentation is sufficient.

Independent of that answer, the guardrail and the schema-verification work below are worth doing, because the failure mode is silent and the diagnostics actively mislead.

---

# Fix plan, one PR

## 1. Stop group-killing a PID the daemon no longer owns

A PID is only meaningful while the daemon still holds the unreaped child. After `cmd.Wait()` returns, the number is back in the OS pool and may already belong to something else. Because the kill is negative it targets a whole process group, so a stale number takes out an entire unrelated family.

- Delete the `case <-done` escalation at `manager.go:375`. After `Wait()` has returned there is nothing left the daemon legitimately owns, so this branch can only ever hit an unrelated process group.
- Same for the post-`done` calls in `forceStopAfterGrace` (`manager.go:797`, `:799`).
- `manager.go:474` in `waitForExit` runs immediately after `Wait()` returns and has the same problem.
- If orphaned descendants that ignored SIGTERM are a real concern, the group kill has to happen *before* the root is reaped, while the PID is still owned. Ordering is the fix, not an extra escalation afterwards.
- `reapPersistedProcesses` (`manager.go:829`) group-kills PIDs recorded by a previous daemon generation. Those are stale by construction. Either verify the PID still belongs to a preview process before killing, or drop the mechanism.

Invariant worth encoding: never pass a PID to `kill` unless the caller can prove the child has not been reaped.

## 2. Treat server-level probe errors as inconclusive

Split `sessionMissingOutput` (`tmux.go:878`):

- Keep `can't find session` and `session not found` as genuine per-session evidence.
- Map `no server running` and `error connecting` to `ProbeFailed` so the existing arbitration handles them.

This restores the behaviour the code already documents. The safety net exists and is correctly written, it is just bypassed before it can run.

`killSessionMissingOutput` delegates to the same helper. Being generous about "already gone" during teardown is fine, but it should not share a predicate with liveness probing.

## 3. Add a reaper circuit breaker

Fixes 1 and 2 close the two known doors. This one protects against causes not yet identified.

If a single reconcile pass concludes that every session, or nearly every session, is dead, that is not N independent deaths. It is one infrastructure outage. There is no realistic scenario in which 28 unrelated agents exit within three seconds of each other.

The reaper should refuse to act on a pass where the dead fraction exceeds a threshold, log at error level, and leave sessions untouched pending the next probe. A wrongly-skipped termination is cheap and self-corrects. A wrongly-applied mass termination archives the user's entire board.

Worth having on its own merits: during triage the original tmux death left no crash report, no unified-log entry, and no `memorystatus` record, so root-causing from forensics alone was not possible. A circuit breaker would have preserved the board regardless of cause.

## 4. Repair the schema and make the version checklist untrustworthy by design

- Add a forward repair migration numbered above any version plausibly burned, that adds `diff_base_sha` and `diff_base_ref` **only if absent**. It must be safe on profiles that applied `0040` correctly and on profiles already patched by hand, so a plain `ALTER TABLE ... ADD COLUMN` will not do.
- Add a startup schema reconciliation that verifies the physical columns the generated queries depend on, rather than trusting `goose_db_version`. That table is known to lie on at least some real installs.
- Add a guardrail that rejects reuse of a version number already shipped in any nightly. Renumbering or deleting migration files after release is what produces this class of failure. Note that in-flight feature branches such as `pipelines` also claim numbers in this range, so branch and release numbering need to be reconciled.
- Surface migration and schema mismatches as a specific, actionable error instead of an opaque 500. Today `/healthz` and `/readyz` report healthy while the core workflow is dead.
- Triage note: a working `select diff_base_sha` does not prove a profile was never affected, since hand-patched databases resolve fine while the underlying burned versions remain.

## Regression tests

- `forceKillPreviewProcess` is never reached with a reaped `cmd`.
- A probe returning `no server running` yields `ProbeFailed`, not `ProbeDead`, and leaves `IsTerminated` unchanged.
- A reconcile pass where every session probes dead terminates nothing and logs an error.
- Session list succeeds against a DB whose `goose_db_version` contains 40 through 51 but whose `sessions` table lacks the diff-base columns.
- The repair migration is idempotent: it succeeds on a DB that already has the columns.

---

Both halves share one root pattern: trusting a stale identifier instead of checking reality. A PID that used to be a preview server, and a version number that claims a migration ran. In both cases the fix is to verify the thing in front of you rather than trust a number written down earlier.

harshitsinghbhandari:
Fixes #3475 (and the migration half originally filed as #3476).

Three distinct failures, one shared root pattern: trusting a stale identifier (a PID, a tmux exit message, a goose version number) instead of verifying the thing in front of us.

## A1: preview server group-killed PIDs it no longer owned

Once `cmd.Wait()` returns, the PID is back in the OS pool; `kill(-pid, SIGKILL)` on that number targets a whole process group, and a tmux server is its own group leader, so one recycled PID wiped every tmux session at once. Every escalation site had the flaw (`stop`, `waitForExit`, `failAndStop`, `forceStopAfterGrace`), and `reapPersistedProcesses` group-killed PIDs recorded by the previous daemon generation with no check at all.

**Fix:** every preview root's kernel start time is captured at launch (`ps -o lstart=` on unix, `GetProcessTimes` on Windows) and persisted in the process registry. No group/tree kill is sent unless the number is positively re-verified:

- live process whose start time mismatches: recycled, never signalled;
- reaped root whose group still has members: those are our orphaned descendants (POSIX reserves a live group's PGID), which is exactly what escalation exists to reap, so the intended "descendants ignored SIGTERM" behavior is kept rather than deleted;
- registry entries with no recorded identity (older builds): skipped;
- identity capture failed: fall back to signalling only the direct child via `os.Process`, which refuses to signal a reaped child and can never hit a recycled PID.

## A2: server-level probe failures misread as per-session death

`sessionMissingOutput` matched `no server running` and `error connecting`, so `IsAlive` returned a confident "dead" for every session the moment the server was gone, and the reaper archived the whole board, even though the agents were still alive as orphans of the killed server.

**Fix:** the predicate is split. `can't find session` / `session not found` stay definitive per-session evidence. Server-level output now wraps a new `ports.ErrRuntimeUnavailable` sentinel, which the reaper reports as `ProbeFailed`, and the existing LCM arbitration (which was always correct, just bypassed) never terminates on it. Teardown (`kill-session`) keeps tolerating server-level output as already-gone. Restore, boot reconcile, and reconcile-reap map the sentinel back to their intended recovery semantics, so `ao session restore` after a dead server and the post-reboot boot pass both still work.

**Plus a reaper circuit breaker** for doors not yet found: if a single pass concludes that most of the board (>50%, minimum 5) is dead at once, that is one infrastructure outage, not N independent exits. The pass is downgraded to inconclusive and logged at error level; a skipped termination self-corrects on the next probe, a mass termination does not.

## B: burned migration versions 500 the session list

Profiles whose `goose_db_version` already records version 40 (written by a fork/branch build; a field Windows profile has 40-46) make goose silently skip the real `0040_add_session_diff_base.sql`. Generated queries then fail with `no such column: diff_base_sha`, so `GET /api/v1/sessions` returns an opaque 500 while `/healthz` stays green.

**Fix:** startup no longer trusts `goose_db_version` for these columns. After `goose.Up`, a schema reconciliation verifies the physical columns the generated queries depend on and adds any that are missing. It is idempotent (healthy and hand-patched databases untouched) and failures surface as a specific, actionable startup error. A versioned repair migration was deliberately not used: a burned version number is the exact mechanism that caused the skip, so it cannot be the delivery vehicle for its own fix (a profile burned past the repair's number would skip it too).

**Guardrail:** a new ledger test freezes every shipped migration version to its exact filename and rejects burned numbers: 22 (shipped in #2412, then reverted) and 41-51 (claimed by branch builds, observed recorded in the field). New migrations start at 52. Renumbering, renaming, or deleting a shipped migration now fails the build with an explanation instead of surfacing months later as an unexplainable 500.

## Regression tests

- reap refuses a live process whose identity mismatches the recorded preview, and skips legacy registry entries with no identity; verified kills still work; the post-reap escalation still reaps orphaned group members (real processes, not stubs)
- `no server running` / `error connecting` surface as `ErrRuntimeUnavailable`, and the reaper reports `ProbeFailed` for probe errors (existing test)
- a pass where every session probes dead terminates nothing; a minority-dead pass still concludes; small boards never trip the breaker
- session list succeeds against a DB whose `goose_db_version` records 40-51 but whose `sessions` table lacks the diff-base columns (built through the real migration path, listed through the real store); repair is idempotent
- migration version ledger test as above

## Verification

- `go build ./...` and `go vet ./...` for darwin and windows
- full `go test ./...`: green except `TestWorkspaceFilesIncludeWorkspaceProjectChildRepoDiffs`, which fails identically on a clean `upstream/main` checkout on this machine (pre-existing, environment-dependent)
- `go test -race` on every touched package: green
- `golangci-lint v2.12.2 run --path-mode=abs`: 0 issues

## Known risks / follow-ups

- Identity verification shells out to `ps`/`pgrep` at preview start and kill time; on a system where those fail, teardown degrades to root-only signalling (descendants may leak, innocents never die).
- The verify-then-kill pair is not atomic; without pidfd-style primitives a recycle inside that microsecond window is theoretically possible. It replaces a window that was effectively unbounded.
- After this change a dead tmux server leaves sessions visible (probes keep failing) instead of archiving them; the boot-time reconcile still tears them down with restore markers on the next daemon start. That trade is deliberate per the issue.
- Whether a released nightly ever shipped migrations 41-46 (the open question in the issue) is unanswered here; the reconciliation repairs affected profiles either way.


illegalcall:
If we can’t confirm that the process group still belongs to AO’s preview, let’s not kill it. 
Remove the post-exit group kill, a leaked preview process is safer than accidentally taking down tmux and every agent session.

illegalcall (changes requested):
.

harshitsinghbhandari:
Done in 85c265691. A group signal now requires the leader to be alive with a matching recorded start time; the orphaned-group heuristic and every post-exit escalation site are gone. Descendants that outlive their root are leaked deliberately, and the regression test now asserts the orphan survives the escalation call. The Windows tree kill already refused a missing root, so only the shared manager paths changed there.

harshitsinghbhandari:
Heads-up on an interaction found while updating this PR: main merged 0041 and 0042 after this issue was filed, and both numbers sit inside the range the affected Windows profile already has recorded as applied (40-46). On that class of install goose skips them silently, and the session queries break again on the missing reviewer_harness column, the exact defect-B mechanism.

Handled in 4d9d71970:

- Startup schema reconciliation now also replays 0041's and 0042's column-level effects (resolved_at with its backfill and index swap, reviewer_harness with its review_run index swap), keyed on column absence so healthy databases are untouched. The burned-profile regression test covers all of it.
- The version guard no longer bans the 41-51 range, since main has claimed it and renumbering merged migrations is not this PR's call. It is now an append-only ledger: every migration file needs a reviewed ledger entry, shipped files cannot be renamed or deleted, and 22 (shipped in #2412, then reverted) stays unusable.

One decision left for you: if 0041/0042 have not shipped in a nightly yet, renumbering them above 46 would keep the burned field profiles from skipping them at all, and the reconciliation entries would become pure belt-and-braces. If they have shipped, their numbers are settled and the reconciliation entries are the fix. Until that range is cleared, any migration numbered up to 0046 whose schema the generated queries depend on needs a schemaRepairs entry; the ledger comment says so.

illegalcall (approved):

illegalcall:
Hasnt this commit already entered nightly https://github.com/Untrivial-ai/agent-orchestrator/commit/1bd62cdfdd14cdd286e985be1528fa264d1659e2

PR #3495, opened 2026-08-03
fix: single all-list panel with unread highlight and restore

Annieeeee11:
## Summary
- Replace the Unseen/Unresolved notification panel with a single paginated all-notifications list
- Highlight unread rows on open and auto-mark all unread as read when the panel opens
- Keep whole-row navigation for live sessions; show Restore instead of navigating for terminated sessions
- Finish the product gaps left after #3296 / #3210

## Original request (all covered)

| Request | Status in this PR |
|---|---|
| Terminal button clickable even when session is terminated | Fixed — terminal button removed; terminated rows are not navigable |
| Terminal button should not exist; notification itself should be clickable | Fixed — no `SquareTerminal` button; whole row opens the session |
| Remove Unread / All tabs | Fixed |
| Remove "Mark all read" | Fixed |
| Remove per-item mark-read checkmarks | Fixed |
| Single paginated notifications view showing all notifications | Fixed — `status=all` list with scroll pagination |
| Highlight currently unread notifications | Fixed — unread stay highlighted for the open panel session |
| Opening the panel once auto-marks all unread as read | Fixed — open acknowledges all unread |

## What #3296 already shipped (kept)

#3296 (`fix(notifications): clickable rows, unseen/unresolved sections, auto-resolution`) already landed:

- Whole row clickable (mouse + keyboard)
- Redundant terminal button removed
- Unread/All tabs removed
- "Mark all read" header removed
- Per-item mark-read checkmarks removed
- Auto-acknowledge on panel open (seen axis)
- Backend `resolved_at` + lifecycle auto-resolution + startup reconcile
- Ack-by-ids / mark-all API support used by the panel

## What this PR fixes on top of #3296

| Gap after #3296 | Fix here |
|---|---|
| Panel used **Unseen / Unresolved** sections instead of one all-history list | Single paginated **all** list |
| Unread were separated by section headings, not highlighted in a mixed list | Unread visually highlighted in the all list |
| Row click guarded only on `sessionId` — terminated sessions still navigated to a dead session | Terminated rows are not clickable; **Restore** button restores then opens |
| Product ask was “one all view + highlight unread,” not Unseen/Unresolved | Matches that UX |

## Out of scope / follow-up
- Put `sessionTerminated` on the notification DTO so Restore/live state does not depend on the workspace cache (workspace pending is already gated so rows are not openable until session facts load)
- OS/desktop toast clicks still use `openPrimary` and do not special-case Restore

## Test plan
- [ ] Open bell with mixed read/unread → one list, unread emphasized, no Unseen/Unresolved headings, no tabs / checkmarks / mark-all
- [ ] Opening the panel clears the unread badge; rows that were unread stay highlighted until close
- [ ] Click live session notification (body/row) → navigates to session
- [ ] Click PR title → opens PR; click surrounding row → opens session
- [ ] Terminated session notification → row not clickable; Restore restores and opens the session
- [ ] Scroll near bottom → older pages load
- [ ] `cd frontend && npm run typecheck`
- [ ] `cd frontend && npm test -- src/renderer/components/NotificationCenter.test.tsx src/renderer/lib/notifications.test.ts`

## After
image


ronishrohan (commented):
image

Looks good. Minor notes:

1. `unresolvedNotificationsQueryKey` is still maintained in `notifications.ts` but no longer fetched. Dead cache, worth a follow-up cleanup.
2. `sessionsReady` gates all row clicks (including live ones) until the workspace query resolves. Intentional, but could feel unresponsive on slow first load.

Otherwise solid. Backend contract respected, restore flow correct, tests cover the new paths.

Annieeeee11:
addressed the dead cache in a follow up commit: dropped unresolvedNotificationsQueryKey and the merge/invalidate paths that only served it. The panel now only maintains unread + all.

Leaving sessionsReady as is for now, it avoids a one-frame navigate on terminated rows before restore UI appears. Happy to revisit later so live/known-alive sessions aren’t blocked on cold workspace load.

frontend/src/renderer/components/NotificationCenter.tsx
```diff
@@ -114,102 +134,128 @@ export function NotificationCenter({ style }: NotificationCenterProps) {
...
+		};
+
+		if (!fullAckDoneRef.current && unreadCount > 0) {
+			fullAckDoneRef.current = true;
+			acknowledgedKeyRef.current = pendingKey;
+			const pending = pendingRef.current;
+			if (pending.length > 0) {
+				setHighlightedIds(new Set(pending.map((item) => item.id)));
+			}
+			acknowledge([]);
```
whoisasx: The unread snapshot only contains the first 100 rows, but `acknowledge([])` marks every unread notification server-side. If the user has more than 100 unread notifications, later pages were unread when the panel opened but their IDs are absent from `highlightedIds`; once loaded, they appear read and dimmed. Please preserve the full open-time unread state, or acknowledge loaded IDs incrementally while retaining the unread pagination cursor.

frontend/src/renderer/components/NotificationCenter.tsx
```diff
@@ -72,6 +74,24 @@ function useNotificationTargetNavigation() {
...
+				if (session.isTerminated === true || session.status === "terminated") {
+					ids.add(session.id);
+				}
+			}
+		}
+		return ids;
+	}, [workspaces]);
+	// Until workspace facts land, treat sessions as not-yet-openable so a
+	// terminated row cannot navigate for one frame before restore appears.
+	return { sessionsReady: !isPending, terminatedIds };
```
whoisasx: `!isPending` also becomes true when the workspace query fails. In that state `workspaces` is undefined, `terminatedIds` is empty, and terminated notification rows are treated as live and navigable. Please require successful workspace data before enabling session navigation, and expose a retry/error state rather than interpreting missing facts as `not terminated`.

whoisasx (dismissed):
The single-list interaction, row navigation, restore action, pagination, and cache cleanup are generally coherent, and the daemon's unfiltered session endpoint does include terminated sessions. However, two correctness cases still need addressing before merge: unread state is lost for notifications beyond the first loaded page, and a failed workspace query causes terminated sessions to be treated as live. Once those are fixed and covered by focused tests, this should be ready for another pass.

Annieeeee11:
done

frontend/src/renderer/hooks/useNotificationsQuery.ts
```diff
@@ -29,10 +30,10 @@ export function useMarkAllNotificationsReadMutation() {
 		mutationFn: markAllNotificationsRead,
 		onSuccess: (_updated, ids) => {
 			markAllCachedNotificationsRead(queryClient, ids);
-			// Deliberately no invalidate here: refetching would drop the loaded
-			// pages, and with them the cursor to unseen rows the panel has not
-			// reached yet. The cache is already correct for the ids we sent.
 			void queryClient.invalidateQueries({ queryKey: recentNotificationsQueryKey });
```
whoisasx: Unread rows loaded through later `all` pages are absent from the unread cache. When they are acknowledged, this loop clears zero unread-cache rows, so its global `unreadCount` is not decremented and the bell remains non-zero even though every notification is read. For example, with 101 unread rows, opening clears the first 100 and leaves the badge at 1; loading and acknowledging row 101 updates the All row but leaves the unread cache count at 1. The new pagination test checks the row and mutation but not the badge. Please synchronize the unread count using the mutation's `updatedCount` or an equivalent targeted refresh while preserving the pagination cursor, and cover this case.

whoisasx (changes requested):
Thank you for addressing both previous findings thoroughly. Acknowledgement now uses concrete loaded IDs and preserves later unread pages, while session navigation correctly requires successful workspace data and provides a retry path. The added regression coverage is useful, and the alignment changes are consistent. One follow-on cache correctness issue remains: acknowledging an unread notification discovered only through a later All page does not reduce the unread query's global count, leaving the bell stale. Once that count is synchronized and the 101-notification case is covered, this should be ready for another pass.

frontend/src/renderer/components/NotificationCenter.tsx
```diff
@@ -114,102 +147,117 @@ export function NotificationCenter({ style }: NotificationCenterProps) {
...
 		});
+
 		setActionError(null);
 		void captureRendererEvent("ao.renderer.notification_mark_read_requested", { scope: "all" });
-		void markAllMutate(acknowledging.map((item) => item.id))
+		void markAllMutate(newly)
 			.then(() => captureRendererEvent("ao.renderer.notification_mark_read_succeeded", { scope: "all" }))
 			.catch((error: unknown) => {
 				void captureRendererEvent("ao.renderer.notification_mark_read_failed", { scope: "all" });
+				for (const id of newly) acknowledgedIdsRef.current.delete(id);
```
whoisasx, line 216: Removing these IDs from the ref does not trigger another effect run because `visibleUnreadKey` and the other dependencies remain unchanged. After a transient failure, acknowledgement stays failed until the user closes and reopens the panel, and the error has no retry control. Please introduce an explicit retry trigger/button or another dependency-changing retry mechanism, and cover the fail-then-retry path.

whoisasx (changes requested):
The previous stale unread-badge finding is correctly resolved: the mutation's updatedCount now synchronizes the badge for unread rows discovered only through later All pages, pagination cursors remain intact, and the 101-notification path has focused cache and component coverage. The restore-tooltip change is also valid under the route-level TooltipProvider. One failure-path issue remains: deleting IDs from the acknowledgement ref after a failed mutation does not trigger the effect again, leaving no retry while the panel stays open. Once an explicit retry path and fail-then-retry coverage are added, this should be ready for another pass.

whoisasx (approved):
Re-reviewed the complete notification-center flow after the follow-up fixes. Paginated unread highlighting, terminated-session handling, unread badge synchronization, and transient acknowledgement retry are now covered and behave consistently. The PR merges cleanly with current main and all required checks pass. Approved from my side.

PR #3501, opened 2026-08-03
feat(session): surface harness context pressure on sessions and doctor

### LINKED ISSUE #3322: activity_state cannot distinguish context-exhausted from refused; no harness context-pressure surface
## Problem

AO's `activity_state` enum has no value for "context-exhausted", and the session read model has no field for harness-reported context pressure. As a result, when a worker is at the end of its context budget the orchestrator cannot tell it apart from a worker that is actively refusing work. From the outside the two states look the same: cheap outputs continue, expensive outputs silently drop.

This matters because the orchestrator's intuitive response — re-nudge, escalate tone, repeat the instruction — actively *worsens* the failure when the underlying cause is context exhaustion: every nudge consumes more of the binding resource (the remaining context), so the worker that was about to checkpoint and commit can lose the ability to do even that.

We need a way for the orchestrator (or the operator) to read the worker's context pressure directly and trigger a reap-and-respawn instead of a nudge.

## Evidence in the AO source

- `backend/internal/domain/activity.go` — `ActivityState` is a closed enum with exactly five values:
  ```go
  ActivityActive       ActivityState = "active"
  ActivityIdle         ActivityState = "idle"
  ActivityWaitingInput ActivityState = "waiting_input"
  ActivityBlocked      ActivityState = "blocked"
  ActivityExited       ActivityState = "exited"
  ```
  No `context_exhausted` or equivalent.
- `backend/internal/domain/activity.go` — the file-level comment explicitly states: *"ActivityState is how busy the agent is, reported via the agent's CLI hook callbacks, not inferred from transcript/JSONL."* The harness already has the data (banner metrics, hook telemetry); AO simply does not carry it across the boundary.
- `backend/internal/domain/session.go:81` — `Session.Status` is a different closed enum used for derived PR/lifecycle status, also without a context-pressure concept.
- Repository-wide search for `contextPressure`, `ContextPressure`, `compactThreshold`, `untilAutoCompact`, `usedPercent`, `contextWindow` — zero hits in production code paths.

## Operational evidence (cost; not a claim about AO's mechanism)

We have observed the failure mode at scale. These are *operational* observations from outside AO and not AO mechanism claims; the underlying context-exhaustion behaviour belongs to the agent harness, not AO. We cite the numbers to motivate the request.

- Fleet-wide read across one project: **6 of 9 active workers at ≥91% context pressure**; three at 100%. The reading came from harness-reported banner metrics scraped via a non-AO path.
- Repeated nudges to a single 100%-context worker failed to elicit the requested gate run, although cheap text outputs continued to arrive. Switching from nudge to `ao session kill` + `ao spawn` with a fresh worker recovered the task.
- The cheap-outputs-only shape was reproduced by independent observers on three separate projects on the same day; the cross-cutting signal was a sustained `activity_state=active` for hours on workers whose durable output (commits, gate runs) had stalled.

## Proposed direction

A fix can land in three small pieces:

1. **Surface harness context pressure on the session DTO.** Add an optional field, e.g. `contextPressure { usedPercent int, untilAutoCompactPercent int, source string, observedAt time.Time }` on `Session` / `SessionRecord` (or as an extension field on the controller DTO if persistence is undesired). It is best-effort: only populated when the underlying agent adapter can read it. Empty means "unknown".
2. **Agent-adapter hookup.** For each harness that exposes a banner metric or hook signal for context pressure (e.g. Claude Code's statusline / banner), the adapter publishes the reading via the existing activity-update path or a new one. The adapter reports it; the read model stores it; nothing else changes.
3. **Operator-facing surface.** Extend `ao session get` and `ao session ls` to render the pressure if present, and extend `ao doctor` to warn when any active worker is above (say) 90% context used or below (say) 10% until auto-compact.

## Acceptance criteria (verifiable by a maintainer)

- A worker running an agent harness that exposes a context-pressure signal reports a non-empty `contextPressure` field via `ao session get`. A harness that does not expose it reports the field as absent (not zero, not error).
- `ao session get` renders the pressure in both JSON and human formats, including the source and observed-at timestamp.
- `ao doctor` returns a WARN when any active worker has `usedPercent >= 90` (or the harness-equivalent threshold). The check is harness-agnostic — it fires whenever the reading is present and crosses the threshold.
- Existing `activity_state` semantics are unchanged. No existing test that asserts on `ActivityState` set should break.
- The implementation does not block the activity-update path. A failure to read pressure must not prevent `activity_state` from being recorded.

## Non-goals

- Not asking AO to *cause* or *detect* context exhaustion itself. The harness detects and reports; AO carries the reading.
- Not asking for a change to the `activity_state` enum values. The new information is additive, not a replacement.
- Not asking for an automatic reap-and-respawn policy. The orchestrator (or operator) decides what to do with the reading. A nudge policy that ignores context pressure is still a valid choice; this issue only makes the data available.
- Not asking for a change to the auto-compact threshold of any harness. We only want AO to read and surface it.

## Prior art (in this repo)

- **#3246 (open)** — "Durable notification parity for the existing 'blocked' activity_state". Notification parity for one of the *existing* activity_state values. Does not propose adding any new value or surfacing context pressure.
- **#2944 (open)** — "Sidebar dot + kanban badge locked to PR status — agent working/idle state invisible once PR exists". UI display of the existing states; no context-pressure dimension.
- **#2814 (open)** — `bug(kimi): activity pill shows Idle while agent is actively working (no mid-turn tool-use hooks)`. Harness-specific hook gap for the *kimi* agent.
- **#3306 (open)** — `bug(cursor): activity state stays 'Input Needed' for the whole turn`. Cursor-specific sticky-state bug.
- **#2594 (open)** — `fix(agent): stop tool-use hooks pinning copilot/kiro/cline sessions to needs_input`. Hook-pinning issue for needs_input.
- **#2587 (open)** — `fix(lifecycle): require consecutive dead probes before terminating a session`. Liveness probing, not context pressure.
- **#3040 (open)** — `refactor(lifecycle): remove the worker-idle orchestrator nudge`. About *removing* a nudge — adjacent in spirit (less nudging) but unrelated mechanism.

If any of these is treated by the maintainers as the right home for context-pressure surfacing I am happy to close this as a duplicate.

## Breadth

Universal in principle (every long-lived worker session can run out of context), but observed frequency scales with session longevity and worker autonomy. A user who runs short single-task sessions is unlikely to hit it; a fleet that keeps workers alive across many turns, or that relies on long-lived orchestrators, will. Not a "every user" issue; an "every long-running fleet" issue.

anmol0b:
> tldr ;
> real quick in my own words; when a worker runs out of context it doesn't die,
> it just keeps sending cheap replies while dropping the actual work, and from
> AO's side that looks identical to a worker thats being lazy on "active". poking
> it again is the worst move bc it just burns more context. so this PR only
> ferries the harness's context % to where you can actually see it, nothing
> smarter. didn't add auto-respawn or a banner parser on purpose, those feel like
> separate calls.
>
> the one thing i actually worried about is pressure reports bump updated_at and i
> didn't want that tripping the reaper, checked by hand, reaper reads
> lastActivityAt not updated_at so we're fine.90/10 thresholds are guesses from
> the issue, easy to tweak. if anything the lifecycle bit in manager.go is what
> i'd want eyes on most.

## What

Add an optional, best-effort context-pressure reading to the session read model, carried alongside agent activity signals. A worker that has run out of context can now be distinguished from one that is refusing work — without touching the `activity_state` enum.

Surfaces added:
- `ao session get` → renders `contextPressure: used=N% untilAutoCompact=N% source=… observedAt=…` (human + JSON)
- `ao session ls` → compact `ctx N%` tag
- `ao doctor` → harness-agnostic **WARN** when any active worker has `usedPercent >= 90` or `untilAutoCompactPercent <= 10`

## Why

When a worker is at the end of its context budget it is indistinguishable from a worker actively refusing work: `activity_state=active`, cheap text keeps flowing, expensive outputs (commits, gate runs) silently drop. The intuitive fix — re-nudging — burns more of the one resource it's out of, making it worse. The orchestrator needs to *read* context pressure so an operator can reap-and-respawn instead of nudge.

Closes #3322.

This PR implements **pieces 1 and 3** of the issue (session DTO + operator-facing surface) plus the **carrier path** (piece 2: the activity endpoint accepts the reading). It does **not** add a harness-specific banner parser — see "Intentional omissions" below.

## How

- **`domain.ContextPressure`** (`{usedPercent, untilAutoCompactPercent, source, observedAt}`) added to `SessionRecord` as a `*` pointer. **Nil means the harness has not reported — never zero, never an error** (acceptance #1). DTOs use `omitempty`.
- **Carrier path:** the existing `POST /sessions/{id}/activity` payload optionally accepts `contextPressure`. Edge validation rejects out-of-range percentages (0–100), sanitizes `source`, defaults `observedAt` to now.
- **Lifecycle:** a pressure-only update persists the reading **without changing `activity_state` or `FirstSignalAt`** (acceptance #4) and **never blocks activity recording** — pressure is folded in alongside the activity write, nil is a no-op (acceptance #5). A small `contextPressureEqual` check prevents repeated identical readings from churning `updated_at` or fanning out CDC.
- **Persistence:** new additive migration `0041` adds a nullable `context_pressure` JSON TEXT column and extends the `sessions_cdc_update` trigger so the UI refetches on a change. sqlc queries/store mapping (de)serialize JSON; corrupt stored JSON degrades to nil (best-effort).
- **Generated artifacts regenerated and committed:** `sqlc/gen/*`, `openapi.yaml`, `frontend/src/api/schema.ts` (code-first contract; CI fails on drift).

### Behavioral note for reviewers (not a risk, flagged for transparency)
Pressure-only hook writes now bump `updated_at` and emit a CDC `session_updated` event, where before a no-state/no-agent-session-id hook was a complete no-op. I verified this introduces **no change to orchestrator decisions**:
- Reaper (`observe/reaper`) does not read `updated_at` → kill decisions unaffected.
- Activity staleness gate (`observe/activity/observer.go:104`) keys off `Activity.LastActivityAt`, which pressure writes never set → unaffected.
- Nudges (CI/review/merge-conflict) are PR-observation-driven and deduped via `GetPRLastNudgeSignature`, not `updated_at` → unaffected (the worker-idle nudge was already removed in #3040).

The only consumer of `updated_at` is the `ExpectedUpdatedAt` reconciliation fence (`manager.go:316`); a pressure write landing between the terminal-idle observer's read and apply drops that one signal, which retries on the next 30s tick. This is the **same behavior class as the existing agent-session-id metadata write** (already bumps `updated_at` at the same code path) — no new failure mode.

### Heads-up: CDC/write volume
A harness that reports pressure on *every* hook with a drifting value (87→88→89…) adds one write + one `change_log` row per drift. The equality check handles *identical* values. If this is a concern on a chatty harness, two cheap follow-ups (neither required here): rate-limit in the adapter, or drop `context_pressure` from the CDC trigger clause.

### Intentional omissions (vs. issue scope)
- **No harness banner parser.** No adapter currently extracts Claude Code's statusline/banner metric into a pressure reading. This is deliberate: the issue's non-goal is "the harness detects and reports; AO carries the reading." A vanilla `ao start` will therefore not emit a reading on its own — the evidence below simulates the adapter's report via the real activity endpoint. An adapter hookup is a separate, harness-specific PR.
- **No auto reap-and-respawn policy.** The orchestrator/operator decides what to do with the reading; this PR only makes the data available (issue non-goal).
- **No `activity_state` enum change.** Additive only.

## Testing

Validated locally with the new focused tests, which exercise the **real HTTP controller → real lifecycle manager → real SQLite store** pipeline against an in-memory DB and map 1:1 to the issue's acceptance criteria:

context-pressure test run

Acceptance-criteria coverage:

| Test                                                    | Acceptance criterion                                              |
| ------------------------------------------------------- | ----------------------------------------------------------------- |
| `TestSessionPersistsContextPressure`                    | round-trips through the store; absent when not reported           |
| `TestSessionsAPI_ActivityAcceptsContextPressureOnly`    | pressure-only POST succeeds, no `state` required                  |
| `TestSessionsAPI_ActivityRejectsInvalidContextPressure` | validation rejects 101%, and never blocks activity recording (#5) |
| `TestActivity_ContextPressureOnlyDoesNotChangeActivity` | **#4** — `activity_state` + `FirstSignalAt` untouched             |
| `TestSessionGet_RendersContextPressure`                 | **#2** — human line + JSON, with `source` + `observedAt`          |
| `TestSessionList_RendersContextPressure`                | **#1/#2** — `ctx N%` in `ls`                                      |
| `TestDoctorWarnsOnHighContextPressure`                  | **#3** — WARN at ≥90%, harness-agnostic                           |

Also ran: `go build ./...`, `go vet ./...`, `golangci-lint run` (clean), `go test -race ./...` (only failure is `TestWorkspaceFilesIncludeWorkspaceProjectChildRepoDiffs`, **pre-existing on `main`** — worktree SHA/ref, unrelated), and `npm run frontend:typecheck` (clean).

## Checklist


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101:
Thanks for raising the PR. While I understand the issue being raised that AO currently has no observability if the agent harness runs out of context: can you tell what was the impact of the issue we are trying to fix.

anmol0b:
> Thanks for raising the PR. While I understand the issue being raised that AO currently has no observability if the agent harness runs out of context: can you tell what was the impact of the issue we are trying to fix.

The impact we’re trying to address is that, as described in issue #3322, we currently can’t distinguish a worker that has exhausted its context from one that is simply stuck or refusing to make progress.

In practice, the worker can remain `active` and continue responding, while no longer producing meaningful output such as commits or gate runs. This can result in wasted worker time, and recovery may require manually killing and respawning the worker.

This change adds visibility into the harness-reported context pressure so we can identify this situation and take the appropriate action.

Rishet11:
@anmol0b
Closing this PR as stale after 30+ days without author activity. The linked [#3322](https://github.com/Untrivial-ai/agent-orchestrator/issues/3322) remains open if the context-pressure surface is still wanted.

PR #3502, opened 2026-08-03
fix(frontend): use product app name across OS menus

neversettle17-101:
## Summary
- add a shared `APP_NAME` constant for the Electron main process
- install a platform-aware native application menu on macOS, Linux, and Windows so the app identity reads `Agent Orchestrator` instead of Electron defaults
- keep Windows menu accelerators while hiding the native menu bar as before

## Validation
- `cd frontend && npx vitest run --config vitest.config.ts src/main/menu.test.ts`


neversettle17-101 (commented):
Ready to merge.

I reviewed PR #3502 against main (base 86fc666, head 20daec8). The change centralizes the product name in APP_NAME, installs an application menu on all platforms, preserves the hidden Windows menu/accelerator behavior, and expands menu coverage/tests for macOS/Linux/Windows. I did not find any correctness or regression issues in the changed code.

Tests were not run locally per review instructions; the PR reports frontend typecheck and the focused menu Vitest test.

PR #3515, opened 2026-08-03
fix(conpty): implement RuntimeRestarter for Windows in-place agent restart (#3512)

### LINKED ISSUE #3512: Windows: Ctrl+C kills the agent and the session can't be resumed (ConPTY lacks RuntimeRestarter)
## Bug

On Windows, pressing `Ctrl+C` while an agent is running terminates the agent process entirely. The user is left looking at a dead terminal (scrollback only, no live process) with no way to resume the agent — unlike on macOS/Linux where `Ctrl+C` merely interrupts the current operation and the agent stays at its prompt, and even if it exits the session can be resumed in-place.

**Source:** reported by @prateek in Discord (#bug-triage)
**Analyzed against:** `4c105d533` (origin/main, 2026-08-03)
**Confidence:** **High** for the architectural gap (ConPTY lacks `RuntimeRestarter`); **Medium** for the Ctrl+C-kills-agent trigger (cannot reproduce on this Linux box — needs Windows confirmation).

## Reproduction

1. On Windows, run any agent session (e.g. Claude Code, Codex, opencode) in the AO desktop app.
2. While the agent is mid-operation, press `Ctrl+C` in the terminal panel.
3. ❌ The agent process exits. The terminal shows a dead/exited state.
4. ❌ There is no working way to resume the agent — the terminal is locked to scrollback with no live process.

**Expected:** `Ctrl+C` interrupts the current operation and returns the agent to its prompt (as on macOS/Linux). If the agent does exit, the user should be able to resume it in the same session.

## Root Cause

Two compounding issues — a Windows-specific trigger and a missing runtime capability:

### 1. Ctrl+C kills the agent on Windows (trigger)

On Windows the agent runs inside a **ConPTY** pseudoconsole (selected at `runtimeselect/runtimeselect.go:34`; tmux on Unix). The process tree inside the ConPTY is:

```
ConPTY console
  └─ ao agent-process supervise --session X --launch Y -- <agent-binary>   (supervisor)
       └─ <agent-binary>                                                    (agent)
```

The supervisor (`backend/internal/cli/agent_process.go:58-79`) shares the terminal's foreground process group with its child and consumes its own copy of the interrupt so it survives to reap the child:

```go
// agent_process.go:70-76
// The child shares the terminal foreground process group and therefore
// receives Ctrl-C directly. Consume the supervisor's copy so it remains
// alive long enough to reap the child and publish the exit observation.
interrupts := make(chan os.Signal, 1)
signal.Notify(interrupts, os.Interrupt)
_ = child.Wait()
```

On Unix this works as intended: `Ctrl+C` → `SIGINT` → the agent's own handler interrupts the operation and returns to its prompt. On Windows, `Ctrl+C` is written to the ConPTY as `\x03`, which Windows translates into a `CTRL_C_EVENT` delivered to **every process attached to the console group**. The supervisor survives (it catches `os.Interrupt`), but the agent also receives `CTRL_C_EVENT` and — depending on the agent harness and whether it was launched via a `.cmd`/`.bat` wrapper (which is common for npm-installed CLIs on Windows and interferes with console-control propagation) — exits entirely instead of just interrupting.

The pty-host keep-alive (`conpty/host.go:180-208`, `pumpPTY`) then preserves scrollback but the agent process is gone, with no shell to drop back to (the ConPTY runs the supervise wrapper directly, not a shell).

### 2. ConPTY cannot resume in-place — no `RuntimeRestarter` (the real gap)

`ports.RuntimeRestarter` (`backend/internal/ports/outbound.go:89-93`) is the optional interface for **replacing the process inside an existing terminal session without changing the terminal identity**:

```go
// RuntimeRestarter is an optional runtime capability for replacing the process
// inside an existing terminal session. Implementations should preserve the
// handle when possible so attached clients do not need a new terminal identity.
type RuntimeRestarter interface {
    Restart(ctx context.Context, handle RuntimeHandle, cfg RuntimeConfig) (RuntimeHandle, error)
}
```

- **tmux implements it** via `tmux respawn-pane -k` (`backend/internal/adapters/runtime/tmux/tmux.go:351`, args built in `tmux/commands.go:21`). This replaces the dead process in the same pane — the attached client keeps its terminal identity and sees the agent come back.
- **ConPTY does NOT implement it.** There is no `Restart` method anywhere in `backend/internal/adapters/runtime/conpty/` (confirmed by search). `conpty/runtime.go:18` only asserts `var _ ports.Runtime = (*Runtime)(nil)`.

As a result, `restartRuntime` (`backend/internal/session_manager/manager.go:1333-1347`) takes the fallback path for Windows:

```go
// manager.go:1338-1346
if alive {
    if restarter, ok := m.runtime.(ports.RuntimeRestarter); ok {  // false on Windows
        return restarter.Restart(ctx, handle, cfg)
    }
    if err := m.runtime.Destroy(ctx, handle); err != nil { ... }  // destroys old ConPTY
}
return m.runtime.Create(ctx, cfg)                                 // new ConPTY, new port, new identity
```

Destroy+Create spawns a **brand-new pty-host on a new loopback port** with a new terminal identity. The desktop client that was attached to the old pty-host (now destroyed) is left on a dead connection, and there is no seamless re-attach — the user is stuck looking at exited scrollback.

## Fix

**Primary:** Implement `ports.RuntimeRestarter` on `conpty.Runtime` so that resume replaces the agent process inside the existing ConPTY (preserving the terminal identity) rather than Destroy+Create-ing a new one. This mirrors the tmux `respawn-pane` path and makes Windows resume symmetric with Unix.

ConPTY does not have a direct `respawn-pane` equivalent (one ConPTY = one process). Two viable approaches:
- **(a) Recreate the process inside the same pty-host:** have the pty-host destroy the old `conptyConn` and create a new ConPTY + process on the same listener (same port, same handle), so attached clients reconnect transparently. This keeps the terminal identity stable.
- **(b) At minimum, ensure the Destroy+Create fallback reconnects the desktop client** to the new pty-host automatically after resume (investigate whether the frontend's terminal stream re-binds to the new handle — this may also need a frontend change).

**Secondary (trigger):** Investigate why `Ctrl+C` exits the agent on Windows rather than interrupting. Likely candidates: `.cmd`/`.bat` wrapper scripts interfering with `CTRL_C_EVENT` propagation, or specific agent harnesses not installing a Windows console-control handler. Even if the agent "should" handle it, AO's resume must be resilient to agent exit regardless of cause.

## Impact

- **Severity:** core workflow broken on Windows — a single accidental `Ctrl+C` kills the session with no recovery path.
- **Platform:** Windows only (macOS/Linux unaffected: SIGINT interrupts, and tmux `respawn-pane` enables resume).
- **Workaround:** none user-facing; the session must be abandoned/recreated.

## Related

- [#1565](https://github.com/Untrivial-ai/agent-orchestrator/issues/1565) — Windows ConPTY teardown (console window flashing; same subsystem)
- [#1756](https://github.com/Untrivial-ai/agent-orchestrator/issues/1756) — "Tmux session shouldn't die when the agent process inside it exits" (CLOSED — this was the Unix fix; the Windows equivalent is missing)
- [#3300](https://github.com/Untrivial-ai/agent-orchestrator/issues/3300) — killed/exited session page is a confusing dead-end (overlapping UX)

atul-upadhyay-7:
## Summary

Fixes #3512 — on Windows, pressing Ctrl+C kills the agent and there is no in-place resume path. This PR implements `ports.RuntimeRestarter` for the `conpty` runtime adapter so that agent resume replaces the process inside the existing pty-host (same terminal identity, same port) instead of doing a destructive Destroy+Create that disconnects attached clients.

## Root Cause

The `tmux` runtime (Unix/macOS) already implements `ports.RuntimeRestarter` via `respawn-pane`. The `conpty` runtime (Windows) had no `Restart` method at all. When `restartRuntime` in `session_manager/manager.go` calls the interface:

```go
if restarter, ok := m.runtime.(ports.RuntimeRestarter); ok {
    return restarter.Restart(ctx, handle, cfg) // ← only tmux hit this
}
// conpty fell through to Destroy + Create → new port, stranded client
```

## Changes

| File | What changed |
|---|---|
| `proto.go` | Added `MsgRestartReq` (0x09) / `MsgRestartRes` (0x0a) opcodes + `RestartPayload` / `RestartResponse` JSON structs |
| `host_conpty_windows.go` | `newConPTY` now accepts `env map[string]string` and overlays it onto `os.Environ()` |
| `host_conpty_other.go` | Matching stub signature update (non-Windows build) |
| `host_main.go` | Passes `nil` env to `newConPTY` (preserves existing startup behaviour) |
| `host.go` | Thread-safe PTY pointer swap under `h.mu`; `pumpPTY` takes the `ptyConn` it owns; `MsgRestartReq` handler closes old PTY, spawns new one, broadcasts `alive=true` to all connected clients |
| `client.go` | `clientRestart()` helper — sends `MsgRestartReq` and waits for `MsgRestartRes` ACK |
| `runtime.go` | `Runtime.Restart()` implements `ports.RuntimeRestarter`; returns the **same `RuntimeHandle`** so attached clients need no reconnection |
| `host_test.go` | `TestRestart`: old PTY closed, new PTY active, `MsgStatusRes alive=true` broadcast, input routing to new process |
| `runtime_test.go` | `TestRuntimeRestart`: end-to-end Restart through the Runtime layer |

## Verification

```bash
# Tests that didn't exist on main (bug) now pass on this branch (fix):
go test -v -run "TestRestart|TestRuntimeRestart" ./internal/adapters/runtime/conpty/...
# --- PASS: TestRestart (0.00s)
# --- PASS: TestRuntimeRestart (0.05s)

# Full suite still green:
go test ./...     # all packages pass
npm run lint      # 0 golangci-lint issues
```

## Before / After

| | Before (main) | After (this PR) |
|---|---|---|
| Ctrl+C on Windows | Agent process dies | Agent interrupted |
| Click Resume | **New terminal** (different connection, stranded) | **Same terminal**, agent restarts in-place |
| Terminal scrollback | Lost / reset | Preserved |

## Omissions
The secondary trigger (why Ctrl+C kills rather than just interrupts the agent binary — likely `.cmd`/`.bat` wrapper interference with `CTRL_C_EVENT` propagation) is left for a follow-up as noted in the issue. This PR makes resume resilient regardless of _how_ the agent exits.

## Testing Note
Since `conpty` only runs on Windows and CI runs on Linux, the restart path is tested via:
1. Fake-PTY unit tests that simulate the full `MsgRestartReq`/`MsgRestartRes` handshake end-to-end
2. Compile-time assertion: `var _ ports.RuntimeRestarter = (*Runtime)(nil)` — would fail to build if the interface were unsatisfied

A Windows smoke test (Ctrl+C → Resume → same terminal window) would be the final confirmation.

### DIFF
--- a/backend/internal/adapters/runtime/conpty/client.go
+++ b/backend/internal/adapters/runtime/conpty/client.go
@@ -24,6 +24,7 @@ const (
 	dialTimeout      = 3 * time.Second
 	getOutputTimeout = 3 * time.Second
 	isAliveTimeout   = 2 * time.Second
+	restartTimeout   = 10 * time.Second
 )
 
 // dialHost opens a TCP connection to addr with a deadline. Callers close it.
@@ -285,7 +286,7 @@ func clientRestart(addr string, payload RestartPayload) (int, error) {
 	}
 	defer func() { _ = conn.Close() }()
 
-	_ = conn.SetDeadline(time.Now().Add(dialTimeout))
+	_ = conn.SetDeadline(time.Now().Add(restartTimeout))
 
 	reqBody, err := json.Marshal(payload)
 	if err != nil {
--- a/backend/internal/adapters/runtime/conpty/client.go
+++ b/backend/internal/adapters/runtime/conpty/client.go
@@ -275,3 +275,65 @@ func clientKill(addr string) error {
 	}
 	return nil
 }
+
+// clientRestart sends MsgRestartReq to the host, waits for MsgRestartRes, and
+// returns the new process PID.
+func clientRestart(addr string, payload RestartPayload) (int, error) {
+	conn, err := dialHost(addr, dialTimeout)
+	if err != nil {
+		return 0, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	_ = conn.SetDeadline(time.Now().Add(dialTimeout))
+
+	reqBody, err := json.Marshal(payload)
+	if err != nil {
+		return 0, err
+	}
+	reqFrame, err := EncodeMessage(MsgRestartReq, reqBody)
+	if err != nil {
+		return 0, err
+	}
+	if _, err := conn.Write(reqFrame); err != nil {
+		return 0, err
+	}
+
+	type restartResult struct {
+		resp RestartResponse
+		err  error
+	}
+	resC := make(chan restartResult, 1)
+	parser := NewMessageParser(func(msgType byte, payload []byte) {
+		if msgType == MsgRestartRes {
+			var resp RestartResponse
+			decodeErr := json.Unmarshal(payload, &resp)
+			select {
+			case resC <- restartResult{resp: resp, err: decodeErr}:
+			default:
+			}
+		}
+	})
+
+	buf := make([]byte, 4096)
+	for {
+		n, err := conn.Read(buf)
+		if n > 0 {
+			parser.Feed(buf[:n])
+		}
+		select {
+		case result := <-resC:
+			if result.err != nil {
+				return 0, result.err
+			}
+			if !result.resp.Success {
+				return 0, errors.New(result.resp.Error)
+			}
+			return result.resp.PID, nil
+		default:
+		}
+		if err != nil {
+			return 0, err
+		}
+	}
+}
--- a/backend/internal/adapters/runtime/conpty/host.go
+++ b/backend/internal/adapters/runtime/conpty/host.go
@@ -73,6 +73,7 @@ type host struct {
 	// = none applied yet). Guarded by mu; used to skip redundant resizes.
 	curCols, curRows int
 
+	restarting   bool          // true if a restart is in progress. Guarded by mu.
 	shutdownOnce sync.Once
 	shutdownC    chan struct{} // closed when Shutdown is called
 }
@@ -384,8 +385,20 @@ func (h *host) handleClientMsg(conn net.Conn, msgType byte, payload []byte) {
 		}
 
 		h.mu.Lock()
+		if h.restarting {
+			h.mu.Unlock()
+			h.sendRestartResponse(conn, false, "restart already in progress", 0)
+			return
+		}
+		h.restarting = true
 		oldPTY := h.cfg.PTY
 		h.mu.Unlock()
+
+		defer func() {
+			h.mu.Lock()
+			h.restarting = false
+			h.mu.Unlock()
+		}()
 
 		if oldPTY != nil {
 			_ = oldPTY.Close()
--- a/backend/internal/adapters/runtime/conpty/host.go
+++ b/backend/internal/adapters/runtime/conpty/host.go
@@ -73,7 +73,7 @@ type host struct {
 	// = none applied yet). Guarded by mu; used to skip redundant resizes.
 	curCols, curRows int
 
-	restarting   bool          // true if a restart is in progress. Guarded by mu.
+	restarting   bool // true if a restart is in progress. Guarded by mu.
 	shutdownOnce sync.Once
 	shutdownC    chan struct{} // closed when Shutdown is called
 }
--- a/backend/internal/adapters/runtime/conpty/host.go
+++ b/backend/internal/adapters/runtime/conpty/host.go
@@ -16,6 +16,9 @@ import (
 	"time"
 )
 
+// newConPTYFunc creates a new ConPTY connection. Overridden in tests.
+var newConPTYFunc = newConPTY
+
 // ptyConn is the host's handle to the running agent's pseudo-terminal.
 // The real impl (conptyConn) lives in host_conpty_windows.go; tests use a fake.
 type ptyConn interface {
@@ -117,7 +120,7 @@ func (h *host) applyLargestLocked() {
 // run is the main event loop.
 func (h *host) run(ctx context.Context) error {
 	// Pump PTY output to ring + broadcast.
-	go h.pumpPTY()
+	go h.pumpPTY(h.cfg.PTY)
 
 	// Watch for ctx cancellation and trigger shutdown.
 	go func() {
@@ -155,8 +158,14 @@ func (h *host) shutdown() {
 	h.shutdownOnce.Do(func() {
 		close(h.shutdownC)
 
+		h.mu.Lock()
+		pty := h.cfg.PTY
+		h.mu.Unlock()
+
 		// 1. Dispose the ConPTY first (critical ordering).
-		_ = h.cfg.PTY.Close()
+		if pty != nil {
+			_ = pty.Close()
+		}
 
 		// 2. Brief grace so the OS ConPTY helper can clean up.
 		time.Sleep(50 * time.Millisecond)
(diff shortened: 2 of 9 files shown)


atul-upadhyay-7:
Screenshot From 2026-08-04 00-51-58

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

harshitsinghbhandari:
One blocker: host.go:386-405. Two concurrent MsgRestartReq's (e.g. a client retry racing the original) both read the same oldPTY, both Close() it, then each spawns via newConPTYFunc and overwrites h.cfg.PTY. The PTY from the losing request is orphaned: its pumpPTY goroutine keeps broadcasting terminal data (only the exit-status broadcast is gated on h.cfg.PTY == pty) and its child process is never closed. Fix: hold the lock across the whole close-then-respawn sequence, or reject concurrent restarts with a restarting flag.

Non-blocking notes: Restart never checks handle.ID against cfg.SessionID the way tmux's Restart does, and clientRestart reuses the generic 3s dialTimeout for the full spawn-and-ACK round trip, so a slow spawn makes the client report failure while the host actually restarted.

atul-upadhyay-7:
## Review feedback addressed

This commit fixes all three issues raised by @harshitsinghbhandari:

### 1. Concurrency race in `host.go` ✅

Added a `restarting bool` field (guarded by `h.mu`) to the `host` struct. The `MsgRestartReq` handler now:
- Atomically checks-and-sets the flag while holding the lock before touching the old PTY
- Defers clearing the flag so it is always reset on exit (success or failure)
- Returns an explicit `{success: false, error: "restart already in progress"}` to the second caller

This closes the race window where two concurrent requests could both close-then-spawn, orphaning a PTY. Covered by new **`TestConcurrentRestarts`** in `host_test.go` (runs cleanly under `-race`).

### 2. Session ID validation in `runtime.go` ✅

Added an upfront guard to `Restart` that returns an error if `handle.ID != string(cfg.SessionID)`, mirroring the same check in the tmux runtime. Covered by new **`TestRuntimeRestart_MismatchedSessionID`** in `runtime_test.go`.

### 3. Restart timeout in `client.go` ✅

Introduced `restartTimeout = 10s` and applied it as the I/O deadline for `clientRestart`, replacing the generic `dialTimeout` (3s). Windows ConPTY spawn can legitimately take longer than 3s under load; the 10s budget matches what the tmux runtime allows for process startup.

All existing tests continue to pass, including `go test -race`.

i-trytoohard:
@atul-upadhyay-7 Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@atul-upadhyay-7
Closing this PR as stale after 30+ days without author activity. The linked [#3512](https://github.com/Untrivial-ai/agent-orchestrator/issues/3512) remains open if the ConPTY restart fix is still needed.

PR #3531, opened 2026-08-03
feat(frontend): extend command palette with session actions and attention tray

### LINKED ISSUE #3530: [Feature]:  Reduce friction when acting on sessions that need attention on the pallete and the menu bar
### Problem

As a developer using Agent Orchestrator, I need to open a session page before I can take common follow-up actions or create a new task.

This adds friction when working across multiple sessions, especially when a session needs attention. The existing command palette provides quick navigation, but it does not yet provide session-level actions or an inline task-creation flow. macOS users also have no quick way to discover which sessions need attention while working in another app.

This affects users managing multiple projects and sessions, particularly on macOS.

### Proposed direction

Extend the existing command palette with:

- Session drill-in actions such as Jump to session, Copy branch, and Resume terminated agent.
- Inline New Task creation that can be completed with keyboard navigation.
- Attention-zone-aware search ranking so sessions needing action remain visible.
- A macOS menu-bar indicator showing the number of sessions needing attention.
- Clickable tray entries that jump directly to the selected session.
- Shared task-composer behavior between the existing New Task dialog and the command palette.

Merge, PR approval, CI retry, and query-to-agent spawning should remain out of scope for this upgrade.

### Alternatives considered

Keep the current navigation and dialog-based workflows.

### Before submitting

achalbajpai:
Fixes: #3530 

## What was done

- Added session drill-in actions to the existing command palette:  Jump to session , Copy branch , Resume terminated agent
- Added inline New Task creation inside the palette with keyboard navigation and draft-discard protection.
- Improved palette search so attention-zone matches remain visible while searching.
- Added a macOS menu-bar attention indicator with click-to-jump session entries.
- Extracted the shared task composer so the existing dialog and palette use the same task-creation flow.

### Images of the new bar 
image

### Drill-in actions 
https://github.com/user-attachments/assets/fcf95fad-0591-4d4d-b432-928a80a9ee13

### Resume a terminated session
https://github.com/user-attachments/assets/90b15fec-b5a2-42cc-998b-d9548ac9f7bf

### Inline new task (no popup)
https://github.com/user-attachments/assets/2548c81f-1809-47e3-aa62-dc83fc98ebdd

### Localization - language
https://github.com/user-attachments/assets/722d6e37-60d6-4ad2-a16a-e1874d3ffcd5

### Menu bar Tray
https://github.com/user-attachments/assets/9a8dae2e-4179-41ef-92d9-3be0c5b265ac

## Tests

image

## Manual testing

- [x] macOS
- [ ] Windows - needs verification for Ctrl+K and the same flows
- [ ] Linux - needs verification for Ctrl+K and the same flows


frontend/src/main/tray-lifecycle.ts
```diff
@@ -0,0 +1,64 @@
+import type { BrowserWindow, IpcMainEvent } from "electron";
+import type { TrayController } from "./tray";
+import { TRAY_OPEN_SESSION_CHANNEL, type TrayAttentionState, type TrayOpenSessionTarget } from "../shared/tray";
+
+export function isTrayEnabled(platform: NodeJS.Platform, isPackaged: boolean, appVersion: string): boolean {
+	return platform === "darwin" && (!isPackaged || appVersion.includes("-nightly."));
```
illegalcall, line 6: **P1:** This disables the new tray for every packaged stable build. The issue asks for a macOS menu-bar indicator and clickable session entries, but normal desktop-release users will never get either feature; only dev and nightly versions pass this condition. Please enable it for stable macOS builds, or document an agreed rollout gate in the issue/PR if this omission is intentional.

frontend/src/main.ts
```diff
@@ -365,6 +389,7 @@ function createWindow(): void {
 		browserViewHost?.dispose();
 		browserViewHost = null;
 		mainWindow = null;
+		trayLifecycle.clear();
```
illegalcall: **P2:** Clearing the controller here empties the tray as soon as the macOS window is closed. The process deliberately stays alive after `window-all-closed`, and the renderer is the only attention-state producer, so the surviving menu-bar item reports zero and loses all session links until the window is reopened. That breaks the background-app use case described in #3530. Preserve the last tray state across window closure (or move state production somewhere that remains live).

frontend/src/renderer/hooks/useCreateTask.ts
```diff
@@ -0,0 +1,50 @@
...
+export type CreateTaskInput = {
+	projectId: string;
+	title: string;
+	prompt: string;
+	branch?: string;
+	harness?: AgentProvider;
+	attachments?: ImageAttachmentPayload[];
+};
+
+export function useCreateTask(): (input: CreateTaskInput) => Promise<string> {
```
illegalcall: **P3:** This hook has only one production consumer (`TaskComposer`). `AGENTS.md` says not to add abstractions for one-off use cases and to add helpers only when they remove duplication across real call sites. Please inline this API/telemetry logic into the shared composer, or establish a genuine second consumer before introducing the layer.

frontend/src/main/tray-lifecycle.ts
```diff
@@ -0,0 +1,64 @@
...
+	focusWindow: () => void;
+};
+
+export type TrayLifecycle = {
+	openSession(target: TrayOpenSessionTarget): void;
+	handleSetAttentionState(event: IpcMainEvent, state: TrayAttentionState): void;
+	handleRendererReady(event: IpcMainEvent): void;
+	clear(): void;
+	dispose(): void;
+	getPendingTarget(): TrayOpenSessionTarget | null;
```
illegalcall: **P3:** `getPendingTarget()` is only used by tests, so it enlarges the production lifecycle contract solely to expose private state. This conflicts with the repo's one-off-abstraction rule. The tests can assert the observable `webContents.send` behavior after `handleRendererReady()` instead.

frontend/src/renderer/components/CommandPalette.tsx
```diff
@@ -18,23 +21,43 @@ import { isMacPlatform } from "../lib/platform";
...
+	| { mode: "new-task"; projectId: string };
+
+
 function terminalHasFocus(): boolean {
 	if (typeof document === "undefined") return false;
 	const active = document.activeElement;
 	return active instanceof Element && active.closest(".xterm") !== null;
 }
 
+function findSession(
```
illegalcall: **P3 (Duplicated Code):** This workspace/session scan duplicates `findSession` in `renderer/lib/command-palette.ts`. Please consolidate the lookup into one helper that returns both the workspace and session so the two implementations cannot drift.

illegalcall (commented):
Current main does not merge cleanly with this head: git/GitHub report content conflicts in `frontend/src/main.ts` and `frontend/src/renderer/components/NewTaskDialog.tsx`. Please rebase and resolve those conflicts in addition to the inline findings.

frontend/src/main/tray-lifecycle.ts
achalbajpai: This is intentional for the experimental rollout: the tray is currently enabled only in dev/nightly while we validate it before stable release.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #3542, opened 2026-08-04
fix: notify orchestrator when worker turn completes (idle)

AllBeingsFuture:
## Summary
Wire the missing lifecycle dispatcher so a worker **active → idle** transition notifies the project''s live orchestrator.

### Problem
Scaffolding already existed (`ListSessions` for resolve-at-delivery, `NudgeCoordination` for busy/blocked policy), but `ApplyActivitySignal` never called it. Parents saw silence after workers finished turns — especially visible on Windows where operators rely on that signal rather than only manual `ao send`.

### Fix
- Dispatch on worker **active → idle**; skip `waiting_input → idle` and non-workers
- Deliver via `NudgeCoordination` (idle now; queue when orchestrator is busy; drop when blocked)
- Flush deferred completions when the orchestrator next becomes idle
- Regression tests for happy path, deferral, blocked, and no-orchestrator

## User-visible impact
Orchestrators receive a coordination nudge when a worker finishes a turn, so multi-agent workflows resume without manual messaging.

## Test plan
- [x] `go test ./internal/lifecycle/` (pass)
- [x] Manual multi-session: worker completes turn → orchestrator receives idle notification

## Notes
- Cherry-picked onto `Untrivial-ai/main` from fork.
- Single-commit, single-topic PR.

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -122,11 +122,23 @@ type Manager struct {
 	// This coordination is intentionally memory-only: a daemon crash leaves the
 	// durable session exited, so the user can safely retry the resume.
 	pendingLaunches map[domain.SessionID]pendingLaunch
+	// pendingWorkerIdle queues worker turn-completion messages when the project's
+	// orchestrator is mid-turn and cannot safely receive a coordination write.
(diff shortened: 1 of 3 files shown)


harshitsinghbhandari:
This is not intended behavior, please propose this in the 10 PM call and explain why it is needed before starting the implementation.

PR #3546, opened 2026-08-04
feat(telemetry): report code review outcomes

Pulkit7070:
## Why

Code review is a headline AO feature with **no telemetry at all**. The review service had no sink wired, so nothing can answer:

- Are reviewers actually being used?
- Do they approve or request changes?
- How long does a pass take?
- How often does triggering one fail?

Confirmed against the live project: there is no `ao.review.*` event of any kind.

## What

Four events, emitted from the service boundary where the transitions already happen:

| Event | Fields |
|---|---|
| `ao.review.triggered` | `harness`, `created_runs`, `reused` |
| `ao.review.submitted` | `harness`, `verdict`, `duration_ms`, `posted_to_provider` |
| `ao.review.cancelled` | `cancelled_runs` |
| `ao.review.trigger_failed` | `error_kind` |

`created_runs` and `reused` distinguish a genuinely new pass from reuse of a running or up-to-date one, which the engine also reports as success. Without that, adoption would be inflated by repeat triggers.

## Privacy

Enum-like fields only. Three things are deliberately excluded:

- **The review body** — reviewer prose about a user's source code
- **`PRURL`** — identifies the org and repository
- **`TargetSHA`** — identifies a commit in that repository

A test walks every emitted payload and fails if the body, repo name, SHA, a source path, or a provider host appears in any string value. That check sits at the call site rather than relying on the remote allowlist to strip them downstream, since the allowlist failing open would be silent.

## Idempotency

`ao.review.submitted` fires only on the genuine `running` → `complete` transition, matching the store's own idempotency. A reviewer retrying a submit cannot double-count a verdict. Tested by submitting three times and asserting one event.

## Validation

- `go test ./internal/service/review ./internal/adapters/telemetry ./internal/daemon ./internal/httpd ./internal/review` — all passing
- `go build ./...`, `go vet`, `gofmt` clean
- Four new tests: outcome fields, the privacy sweep, resubmit idempotency, and an unwired service staying silent
- **Export path verified by exercising the real PostHog sink**, not just the service. The remote allowlist drops unregistered keys, so a new event with unregistered fields would export as an empty shell. Asserted that `harness`, `verdict`, `duration_ms` and `posted_to_provider` survive export while a `body` does not.

## Notes

The sink is optional. Every existing caller constructs the service without one, so an unwired service stays silent rather than panicking, with a test for that.

Independent of #3401 and #3489. Branched from `main`, no overlapping files with either.

## Follow-ups, not in scope

- No event yet for a review reaching a **merged PR**, which is the actual value moment. That needs PR-merge instrumentation, which does not exist either.
- Reviewer harness adoption will only be comparable once builds carrying this roll out; existing installs report nothing.


i-trytoohard:
From the NotesBot call on `2026-08-04`, this existing pr appears to be relevant.
Match confidence: `0.55` (fuzzy).

Relevant discussion excerpts:
- - *Prasad, Dhruv Sharma, Pulkit, and Nikhil:* Discuss pending items around *chat UI* and *telemetry* for mobile. - *Vaibhaav:* Reduce prompt verbosity in browser annotations, shift from embedded context to tool-based commands where possible, review *VS Code* annotation behavior, and investigate file diff viewer [...]
- - *Prasad:* Add a *Discord* link alongside the new mobile bug-report / issue flow. - *Prasad, Dhruv Sharma, Pulkit, and Nikhil:* Discuss pending items around *chat UI* and *telemetry* for mobile. - *Vaibhaav:* Reduce prompt verbosity in browser annotations, shift from embedded context to tool-based commands where [...]

Source thread: https://discord.com/channels/1476302178913357958/1534255424365269002/1534255424365269002

illegalcall (approved):

Issue #13, opened 2026-02-13
Phase 4: Team features, auth, and remote sessions


AgentWrapper:
## Scope
Multi-user / team support.

- [ ] Dashboard authentication (OAuth / API key)
- [ ] Role-based access (admin, operator, viewer)
- [ ] Remote session support (SSH + tmux on remote machines)
- [ ] Docker runtime plugin (agent runs in container)
- [ ] Kubernetes runtime plugin (agent runs as pod)
- [ ] Audit log (who spawned what, when)

harshitsinghbhandari:
Closing this as part of the migration cleanup for the ground-up rewrite of
Agent Orchestrator.

This item was opened more than a month before the migration batch and the
codebase has been replaced, so this thread no longer maps cleanly onto the
current source tree.

If something here still needs tracking, please reopen this item or open a fresh
one against the new tree. Thank you for contributing.

Issue #12, opened 2026-02-13
Phase 3: Notifications, automation, and cost tracking


AgentWrapper:
## Scope
Make the orchestrator hands-off for routine operations.

## Notifications
- [ ] Discord notifier plugin
- [ ] Teams notifier plugin
- [ ] Email notifier plugin

## Automation
- [ ] Webhook-triggered spawning (GitHub webhook → auto-spawn agent for new issue)
- [ ] Scheduled review-check (cron-like polling for PR review comments)
- [ ] Auto-merge when approved + CI green (configurable per project)
- [ ] Auto-rebase on merge conflicts

## Cost tracking
- [ ] Per-session token/cost tracking from agent introspection
- [ ] Cost dashboard with per-project and per-session breakdowns
- [ ] Budget alerts via notifier

harshitsinghbhandari:
Closing this as part of the migration cleanup for the ground-up rewrite of
Agent Orchestrator.

This item was opened more than a month before the migration batch and the
codebase has been replaced, so this thread no longer maps cleanly onto the
current source tree.

If something here still needs tracking, please reopen this item or open a fresh
one against the new tree. Thank you for contributing.

PR #3554, opened 2026-08-04
feat(mobile): add multi-language UI localization

AllBeingsFuture:
## Summary
- Adds mobile i18n for the same locale set as desktop: `en`, `zh-CN`, `ja`, `ko`, `es`, `fr`, `de`, `pt-BR` (`packages/mobile/lib/i18n/`).
- Lightweight typed catalogs + `useT()` (no new npm deps); device locale default with AsyncStorage override.
- Wires tabs, agents board, orchestrator, PRs, settings, onboarding, pairing, spawn, notifications, session, and shared sheets/cards.
- Settings → Appearance → **Language** lists all 8 languages (native labels, desktop order); switching updates UI immediately.

## How to test
1. `cd packages/mobile && npm install && npm run typecheck && npm test`
2. Run the app; open **Settings → Language** and pick each locale — tabs, empty states, status labels, pairing copy, and alerts should switch without restart.
3. Clear app storage / reinstall and set the device language to Chinese/Japanese/etc. — first launch should default to the matching catalog (`zh*` → `zh-CN`, `pt*` → `pt-BR`, …), else English.

## Tests
- `cd packages/mobile && npm run typecheck` (green)
- `cd packages/mobile && npm test` (297 tests, including `lib/i18n/ui-locale.test.ts`)

## Notes
- Mobile catalogs are mobile-specific keys (not a dump of the desktop JSON); overlapping terms follow desktop tone (会话/编排器/设置, etc.).
- Product/harness IDs and URLs stay untranslated.


Prasad-D-Ware (changes requested):
Solid foundation, and two things are genuinely well done: making a missing key a **compile error** via `Record<AppLocale, MessageCatalog>`, and the `tr: TFunction = enT` trailing-default convention that keeps the pure view helpers testable while making the 52-file diff safe for existing callers.

I built and ran the branch on both platforms rather than reading the diff alone. Everything below is reproducible — commands included at the bottom.

Five changes needed before merge, then a short list of follow-ups.

---

# Required changes

### 1. iOS device-locale detection is dead — `CFBundleLocalizations`

This is the headline feature in the description ("device locale default", test step 3). It currently never fires on iOS.

`ui-locale.ts:76` reads the device tag with:

```ts
Intl.DateTimeFormat().resolvedOptions().locale
```

On iOS, Hermes' `Intl` is backed by Foundation, and `NSLocale.currentLocale` resolves against **the localizations the app bundle declares** — not the user's device language. The built `AO.app` declares none: no `.lproj` directories, no `CFBundleLocalizations`, `CFBundleDevelopmentRegion = en`. iOS therefore hands back `en-US` on every device, `matchDeviceLocale` correctly maps that to `en`, and the `try/catch → "en"` swallows any signal that something is wrong.

The JS is correct. It's being given the wrong input.

**Fix:**

```jsonc
// app.json → expo.ios.infoPlist
"CFBundleLocalizations": ["en", "zh-Hans", "ja", "ko", "es", "fr", "de", "pt-BR"]
```

This also gives users the per-app **Language** row in iOS Settings, which is where most people look first.

Worth considering alongside it: `expo-localization`'s `getLocales()[0].languageTag` reads `preferredLanguages` directly and isn't subject to bundle-localization resolution — which is exactly why it's the ecosystem default here. It's an Expo-managed module, so "no new npm deps" costs nothing in practice.

### 2. Plural selection is wrong for `fr` and `pt-BR`

Every plural site does `count === 1 ? "_one" : "_other"` — `app/(tabs)/orchestrator.tsx:223`, `app/(tabs)/settings.tsx:160`, `app/(tabs)/index.tsx:215`, `lib/PRCard.tsx:83`.

Both French and Brazilian Portuguese put **0 in the `one` category** under CLDR:

```
$ node -e 'for (const l of ["en","fr","zh-CN","ja","ko","es","de","pt-BR"]) {
    const p = new Intl.PluralRules(l);
    console.log(l.padEnd(6), "0->"+p.select(0), " 1->"+p.select(1), " 2->"+p.select(2)); }'
en     0->other  1->one    2->other
fr     0->one    1->one    2->other
zh-CN  0->other  1->other  2->other
ja     0->other  1->other  2->other
ko     0->other  1->other  2->other
es     0->other  1->one    2->other
de     0->other  1->one    2->other
pt-BR  0->one    1->one    2->other
```

So `0 files` renders "0 fichiers" / "0 arquivos" where both languages want the singular. Two of the eight shipped locales are affected, and the current approach hard-blocks ru/pl/ar/cs later, which need `few` and `many`.

**Fix** — fold it into `translate()` so call sites stop making the decision:

```ts
const rules = new Intl.PluralRules(locale);   // memoize per locale
const suffix = rules.select(count);            // "one" | "other" | "few" | ...
// look up `${key}_${suffix}`, fall back to `${key}_other`
```

`Intl.PluralRules` gets all eight right for free, which is the point of using it over a hand-written comparison. While you're there: `common.archiveCount` / `common.archiveCount_other` don't follow the `_one` / `_other` convention the other plural keys use.

### 3. ~14 user-facing strings still hardcoded

`app/session/[id].tsx` is partially wired (`tr()` at :675, :818, :821, :840) but these are not:

- **`:939-941`** — `Alert.alert("Kill session?", \`This stops ${id}.\`)` plus the `"Cancel"` / `"Kill"` buttons. A destructive confirmation is the worst place to fall back to English.
- `:1072`, `:1122` — `"Restoring..."` / `"Restore"` / `"Restore session"`
- `:843`, `:891`, `:930`, `:963`, `:1014`, `:1152` — banners: `Send failed:`, `Kill failed:`, `Restore failed:`, `No preview yet - …`, `Terminal renderer crashed - …`, `Preview failed to load.`
- `:1047`, `:1056`, `:1079` — a11y labels `"Smaller text"`, `"Larger text"`, `"Kill session"`

Plus `app/pair.tsx:138` `accessibilityLabel="Back"` (`common.back` already exists) and `lib/session/Composer.tsx:93` `accessibilityLabel="Send"`.

Screen-reader labels matter *more* here, not less — a VoiceOver user on a Japanese phone gets English announcements with no visual context to fall back on.

To stop these regrowing, `eslint-plugin-i18next`'s `no-literal-string` scoped to `Text`, `accessibilityLabel`, `placeholder`, and `Alert.alert` would hold the line.

### 4. Three French strings left in English

I diffed every catalog against English. Most identical strings are legitimately untranslated product terms (`PRs`, `Terminal`, `Orchestrator`, `Online`) — German's 15 and Spanish/pt-BR's 9 all check out. These three are real misses:

| Key | Current | Should be |
|---|---|---|
| `common.archive` | `Archive` | `Archiver` (it's a button) |
| `common.archiveCount` | `Archive, {{count}} session` | `Archiver, {{count}} session` |
| `common.archiveCount_other` | `Archive, {{count}} sessions` | `Archiver, {{count}} sessions` |

### 5. Japanese truncates the Orchestrator tab

Found by running it — the tab bar renders **`オーケストレー…`**, clipped mid-word.

`app/(tabs)/_layout.tsx:34` sets `tabBarLabelStyle: { fontSize: 11 }`, and four tabs each get ~1/4 of screen width. `オーケストレーター` is 9 full-width characters and overflows where English "Orchestrator" fits. German is the near-miss: `Einstellungen` renders edge-to-edge with no headroom left. Chinese (`编排器`) and Korean are comfortable.

Simplest fix is shortening the Japanese string — `オーケストラ` or `編成` both fit and are arguably better copy. Alternatives: `adjustsFontSizeToFit`, a smaller tab font, or a two-line tab bar.

This is the general risk with catalogs authored against an English-sized layout. **Pseudolocalization** — a dev-only locale that pads every string ~40% and brackets it — is the standard way to catch this class systematically rather than by luck. Worth adding as more screens land.

---

# How this was tested

Built and ran the branch on both platforms. `npm run typecheck` is green and `npm test` passes 297/297 as described; I also verified all 8 catalogs hold exactly **282 keys** with no `{{param}}` drift between locales.

**iOS — proving the `CFBundleLocalizations` bug (iPhone 17 simulator, native rebuild):**

Set the simulator to Chinese, cleared any stored preference, launched:

```bash
xcrun simctl spawn <udid> defaults write -g AppleLanguages -array "zh-Hans-CN" "en-US"
xcrun simctl spawn <udid> defaults write -g AppleLocale -string "zh_CN"
# verified AsyncStorage manifest contained only ao.serverConfig / ao.activeProject — no ao.locale
```

Result: app launched **entirely in English**, while the OS status bar switched to 24-hour time (so the device genuinely was in Chinese).

Then I patched *only* `CFBundleLocalizations` into the built bundle's `Info.plist` and reinstalled — same JS bundle, same device language, still no stored override:

```bash
plutil -insert CFBundleLocalizations -json '["en","zh-Hans","ja","ko","es","fr","de","pt-BR"]' "$APP/Info.plist"
xcrun simctl install <udid> "$APP"
```

Result: app launched **fully in Chinese**, with Settings → 语言 correctly showing 简体中文 as the detected value. Clean A/B — that one key is the entire difference.

**Android — confirming the bug is iOS-only (Android 15 / API 35 emulator):**

```bash
adb shell cmd locale set-app-locales aoagents.dev --locales zh-CN
adb shell am force-stop aoagents.dev && adb shell monkey -p aoagents.dev -c android.intent.category.LAUNCHER 1
```

Result: **fully Chinese** (代理 / 工作中 / 需要你 / 可合并 / 无信号 / 已合并 / 编排器 / 设置). Repeated with `ja-JP` → Japanese and `de-DE` → German. The `ja-JP` run also confirms `matchDeviceLocale`'s prefix mapping works against a real OS tag, not just a test fixture.

So `deviceLocaleTag()` is correct code — Hermes' `Intl` on Android resolves against full ICU data. **The fix in #1 is iOS-only and needs no Android counterpart for detection.** The Japanese run is where the tab truncation in #5 surfaced.

**In-app switching works.** Seeded `ao.locale = "zh-CN"` and relaunched — the whole app re-rendered in Chinese, translations read well, and the Language sheet lists all 8 native labels correctly.

**No regressions found.** Beyond typecheck and tests: every screen I reached renders on both platforms, and the session terminal opens, connects live (57×45 grid), streams output, with composer, key bar and mic all functional. The trailing-default `tr` param means existing callers compile and behave identically.

---

# Follow-ups (not blocking)

- **Relative timestamps aren't localized.** `lib/notificationView.ts:44-46` localizes only `time.now`; `3m` / `4h` / `2d` / `2w` are hardcoded English. Rendered as `3d` on every card in Chinese, Japanese *and* German. `Intl.RelativeTimeFormat(locale, { numeric: "auto", style: "narrow" })` covers all 8.
- **Sentences assembled by concatenation.** `lib/agentsView.ts:187-199` builds `#12, #13 open` in JS and feeds it to `"PR {{parts}}"` — word order is baked in and the `", "` / `" · "` separators aren't localizable. Rendered as `PR #4, #3 已合并` / `マージ済み` / `gemerged`; both CJK locales want `、`. Same issue with `push.fail.serverError.status` = `" (HTTP {{code}})"`, a leading-space fragment spliced in at `lib/pushStatus.ts:171` — every Latin-script catalog left it untranslated, which is the tell.
- **No `android:localeConfig`.** Detection works without it, but AO won't appear under Settings → System → Languages → App languages on Android 13+. The Android analogue of #1, milder. Needs a config plugin since `android/` is prebuild output.
- **`zh-Hant` → Simplified.** `ui-locale.ts:52` maps all `zh*` to `zh-CN` and `ui-locale.test.ts:42` locks in `zh_TW → zh-CN`. Apple's and Google's guidance is not to cross scripts; falling back to English is the more conventional choice. Fine either way — just worth being a deliberate decision with a comment rather than a side effect of prefix matching.
- **Shared locale constants.** `ui-locale.ts` is a near-verbatim copy of `frontend/src/shared/ui-locale.ts` (`APP_LOCALES`, `DEFAULT_LOCALE`, `coerceLocale`) with nothing keeping them in sync. Extracting just the constants is a cheap hedge even if the runtimes stay separate.
- **Suggested test.** Placeholder parity is the one catalog bug class the current tests miss — `assertCatalogCoverage` checks presence and non-emptiness, so a translator dropping `{{count}}` compiles and passes. It's clean today, which makes it the right moment to lock in:

```ts
it("keeps the same {{params}} as English in every catalog", () => {
  const params = (s: string) =>
    [...s.matchAll(/\{\{(\w+)\}\}/g)].map((m) => m[1]).sort().join(",");
  for (const locale of APP_LOCALES)
    for (const key of Object.keys(catalogs.en) as MessageKey[])
      expect(`${locale}.${key}:${params(catalogs[locale][key])}`)
        .toBe(`${locale}.${key}:${params(catalogs.en[key])}`);
});
```

---

# One product decision to make

Not a bug, but it should be a conscious call rather than a discovery.

Today every existing user sees English. After this merges, anyone whose phone is set to one of the 8 languages gets a translated app **on next launch, without opting in** — there's no stored preference, so device detection wins. On Android that happens immediately; on iOS it happens the moment fix #1 lands.

That's the intended design and what most apps do. But it silently changes the UI language for a chunk of the existing user base. If that's not wanted, the alternative is defaulting existing installs to English and only auto-detecting on fresh installs.

Related, minor: `LocaleProvider.tsx:34-42` paints the device-matched locale first and swaps to the stored preference a tick later — so a user on a Chinese phone who deliberately picked English sees a brief Chinese flash on every cold start.

AllBeingsFuture:
## Addressed review feedback (`fc6b9acf`)

Thanks @Prasad-D-Ware — blocking items from the review are fixed on this branch.

### 1. iOS device-language auto-detect
- Added `CFBundleLocalizations` for all 8 locales in `app.json` → `expo.ios.infoPlist`.
- Added `expo-localization` (`~17.0.7`) and use `getLocales()[0].languageTag` in `deviceLocaleTag()`, with `Intl` fallback.
- Config plugin `supportedLocales` also sets iOS localizations + Android `localeConfig` / `locales_config.xml` (Android 13+ per-app languages).
- Tests for `deviceLocaleTag` (injected reader) + existing `matchDeviceLocale` coverage.

### 2. Plurals via `Intl.PluralRules`
- `translate()` now resolves plurals with memoized `Intl.PluralRules(locale)` when `params.count` is set (`base_one` / `base_other`, fall back `_other`).
- Call sites no longer use `count === 1` ternaries (`orchestrator`, `settings`, `index`, `PRCard`).
- Normalized `common.archiveCount` / `common.connectedSessions` to `_one` / `_other`.
- Covered fr/pt-BR `0 → one` in unit tests.

### 3. Missed strings
Session kill Alert, Restore/Restoring CTAs, failed-send/kill/restore banners, no-preview / terminal-crash / preview-failed, a11y labels (zoom, kill, preview), pair Back, Composer Send — all wired through `tr()` with keys in all 8 catalogs.

### 4. French
`common.archive` → **Archiver**; archiveCount strings match.

### 5. Follow-ups from device notes
- Japanese tab label: `オーケストレーター` → **オーケストラ**.
- Android `localeConfig` via expo-localization plugin (see above).

### Nice-to-haves
- `relativeTime` uses `Intl.RelativeTimeFormat(locale, { numeric: "always", style: "narrow" })`.
- Placeholder-parity test: same `{{params}}` as English in every catalog.

### Verify
```
cd packages/mobile && npm run typecheck && npm test
# typecheck green; 304/304 tests passed
```

Ready for re-review.

AllBeingsFuture:
Specific text translation may require manual translation. I believe this should be submitted by native speakers of the target language to provide specific error messages. Currently, it is AI-generated translation, not human translation. I remember it mentioned being AI-generated, not human translation. Some inaccuracies in specific text content are normal. Other issues are being addressed by me

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Prasad-D-Ware (changes requested):
Pulled `fc6b9acf6` and re-tested on both platforms. Four of the five requested changes are correct, and several of the follow-ups landed too — but **the commit crashes the app on launch on both iOS and Android**, so I'm re-requesting changes on one specific thing.

Important: **this is my mistake, not yours.** I recommended `Intl.PluralRules` (and `Intl.RelativeTimeFormat` as a follow-up) without checking whether Hermes implements them. It doesn't. You implemented exactly what I asked for; the advice was wrong for this runtime but also try running them locally before you push for review.

---

# Blocking: Hermes ships neither API

I probed the JS runtime directly on a booted simulator and emulator:

```
INTL_PROBE {"RelativeTimeFormat":"undefined","PluralRules":"undefined",
            "NumberFormat":"function","DateTimeFormat":"function","Collator":"function"}
```

Identical on **iOS and Android** (RN 0.81.5 / Expo 54, `expo.jsEngine: hermes`, `hermesEnabled=true`). Hermes implements `NumberFormat`, `DateTimeFormat` and `Collator` only.

Both new call sites throw `Cannot read property 'prototype' of undefined`:

- **`lib/notificationView.ts:44`** — `new Intl.RelativeTimeFormat(...)`, via `SessionCard.tsx:24`. Takes down every session card.
- **`lib/i18n/translate.ts:33`** — `new Intl.PluralRules(...)`, via `ArchiveHeader` at `app/(tabs)/index.tsx:207`.

I confirmed they're independent failures by temporarily guarding the first one and watching the crash move to the second. The agents board — the app's landing screen — is dead on launch on both platforms.

### Why CI stayed green

`npm run typecheck` passes and **all 304 tests pass**, because vitest runs on Node, which has full ICU. Every `Intl.*` API exists there. The suite cannot see this class of bug at all, which is worth knowing beyond this PR: for anything touching `Intl`, a green suite isn't evidence the runtime supports it.

### Fix

The logic you wrote is right — it just needs the constructors to exist. Two workable options:

**Option A — FormatJS polyfills** (the conventional RN answer, and what the Expo docs point to). Imported before first use, e.g. at the top of `app/_layout.tsx`:

```ts
import "@formatjs/intl-locale/polyfill";
import "@formatjs/intl-pluralrules/polyfill";
import "@formatjs/intl-relativetimeformat/polyfill";
// plus locale-data imports for each of the 8 shipped locales
```

Costs three dependencies and some bundle weight for the locale data.

**Option B — hand-roll it.** For these 8 locales the plural rule is three cases:

| Locales | `one` when |
|---|---|
| `en`, `es`, `de` | `n === 1` |
| `fr`, `pt-BR` | `n < 2` |
| `zh-CN`, `ja`, `ko` | never — always `other` |

Keep `resolvePluralKey`'s shape and swap `Intl.PluralRules` for a small lookup, and revert `relativeTime` to the previous suffix strings (leaving relative-time localization as a genuine follow-up). Zero new dependencies, and it keeps the "no new npm deps" spirit of the original PR.

Either is fine by me. Option B is smaller and lower-risk if you want this merged sooner; Option A is more correct long-term and gets you `zero`/`two`/`few`/`many` for free when ru/pl/ar arrive.

Whichever you pick, please **verify on a simulator or emulator before pushing** — this specific bug is invisible to the test suite.

---

# Everything else checks out

| # | Requested | Status |
|---|---|---|
| 1 | `CFBundleLocalizations` | ✅ Done — and more thoroughly than asked |
| 2 | `Intl.PluralRules` | ⚠️ Logic correct, crashes at runtime (above) |
| 3 | ~14 hardcoded strings | ✅ All wired |
| 4 | French `archive` strings | ✅ `Archiver` |
| 5 | Japanese tab label | ✅ `オーケストラ` |

**#1** went beyond the ask: the plist key *plus* the `expo-localization` plugin with `supportedLocales` for both platforms (which also resolves the `android:localeConfig` follow-up), and `deviceLocaleTag` now reads `getLocales()` with an `Intl` fallback and an injectable reader for tests. I verified prebuild writes `["en","zh-Hans","ja","ko","es","fr","de","pt-BR"]` into `ios/AO/Info.plist` and that `expo-localization` 17.0.9 links into the pod install.

**#3** — grepped for `Alert.alert("`, `accessibilityLabel="`, `setBanner("` and literal `"Restor` across `session/[id].tsx`, `pair.tsx` and `Composer.tsx`: no hits remaining.

Also picked up without being asked: the placeholder-parity test, the `zh-Hant` decision documented as deliberate at `ui-locale.ts`, `archiveCount` renamed to the `_one` / `_other` convention, and the `relativeTime` follow-up.

Catalog audit re-run: all 8 locales at **304 keys**, no missing keys, no extras, no `{{param}}` drift. French identical-to-English strings dropped 22 → 19, matching the three `archive` fixes exactly.

---

# Not yet verified

I could not confirm the iOS device-locale detection end to end — the crash fires before the board renders, so I never reached a screen that would show it. The bundle now declares the localizations correctly and the `getLocales()` path looks right, so I expect it works, but it needs a re-run once the app boots again. Happy to re-test and confirm.

The two concatenation follow-ups (`prs.line` in `lib/agentsView.ts:187-199`, and the `" (HTTP {{code}})"` fragment at `lib/pushStatus.ts:171`) are untouched, which is fine — still non-blocking.

Prasad-D-Ware:
@AllBeingsFuture the current changes have made the app crash on for testing please look into the specifics and fix things

Prasad-D-Ware:
@AllBeingsFuture 
你为什么关闭了这个 PR？可以请你检查并解决相关问题吗？这样我之后就可以合并它了。我在本地运行这个应用时，它甚至都无法启动。

AllBeingsFuture:
为什么不能关闭  不想弄了 你们太敷衍了

AllBeingsFuture:
这个软件 我都不会在使用了  再见  不要回复了

AllBeingsFuture:
image
dis 上你们团队不是挺牛逼的

PR #3556, opened 2026-08-04
fix: add terminal topbar padding

neversettle17-101:
## Summary

Adds vertical breathing room to the session terminal topbar so the terminal controls, session action buttons, and notification bell are not pressed against the topbar frame.

## Changes

- Added a dedicated `--size-session-topbar` token derived from the existing inspector tab height plus vertical padding.
- Applied the new topbar height to the session topbar host and renderer surface.
- Updated sidebar chrome offset calculations so session routes remain aligned with the taller topbar.
- Updated renderer tests that assert the session topbar sizing classes.

## Validation

- `npm --prefix frontend test -- CenterPane.test.tsx shell-new-session-shortcut.test.tsx`
- `git diff --check`

## Risks

Low. This is scoped to session-route topbar sizing and padding; related offset tokens were updated with it to avoid layout overlap.

## Before
Screenshot 2026-08-04 at 4 51 52 PM

## After
Screenshot 2026-08-04 at 4 51 41 PM

### DIFF
--- a/frontend/src/renderer/components/CenterPane.test.tsx
+++ b/frontend/src/renderer/components/CenterPane.test.tsx
@@ -201,12 +201,13 @@ describe("CenterPane toolbar session label", () => {
 		expect(screen.getByText("No session")).toBeInTheDocument();
 	});
 
-	it("uses the inspector tab height for the terminal header", () => {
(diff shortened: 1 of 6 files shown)


Annieeeee11 (approved):
lgtm

PR #3559, opened 2026-08-04
refactor: remove obsolete review result reducer

### LINKED ISSUE #3557: Remove obsolete ApplyReviewResult delivery path
## Summary

Remove the obsolete single-result lifecycle delivery entrypoint and consolidate review feedback delivery through `ApplyReviewBatch`.

## Context

New review runs always receive a non-empty `batch_id`, including single-PR review triggers. `ApplyReviewResult` remains only for batchless legacy rows created before migration 0020.

## Scope

- preserve delivery for legacy batchless review runs
- route all review submissions through `ApplyReviewBatch`
- remove `ApplyReviewResult` from the lifecycle reducer surface
- keep the implementation and tests narrowly scoped

## Acceptance criteria

- legacy batchless review runs are converted to one-run batches
- current and migrated review results use the batch delivery path
- `ApplyReviewResult` and its dedicated tests are removed
- relevant backend tests pass

VenkataSakethDakuri:
## Summary

- backfill legacy batchless review runs as deterministic one-run batches
- route all persisted review feedback through ApplyReviewBatch
- remove the obsolete ApplyReviewResult reducer entrypoint and redundant tests

## Why

All newly created review runs already have a non-empty batch ID, including single-PR triggers. Keeping a separate lifecycle path only for pre-migration rows duplicated delivery behavior and expanded the reducer surface.

Migration 0043 preserves those legacy rows by assigning each its run ID as the batch ID before the single-result path is removed. Backfilled runs now receive the same current-head validation as new runs, so stale undelivered feedback is intentionally skipped when the PR head has moved.

## Validation

- `go test ./internal/lifecycle`
- `go test ./internal/service/review`
- `go test ./internal/storage/sqlite`
- `go test ./...` runs but has unrelated Windows environment failures in packages requiring `sh`/tmux and POSIX path behavior

Closes #3557


harshitsinghbhandari:
can you remove the cloud agents documentation from this PR. @VenkataSakethDakuri

harshitsinghbhandari:
Reviewed. The consolidation is right: batch IDs are minted for every new run, the deleted single-result path only served pre-0020 rows, and the backfill (batch_id = id) preserves per-run dedup granularity. Approved/already-delivered filtering is fully covered by `deliverableRuns`, and the down migration is safe since run and batch IDs are separately minted and never equal.

One blocker:

- **Migration number collision.** `main` now has `0042_review_run_unique_per_harness.sql` (merged after this branch was cut). This PR adds a different `0042_backfill_review_run_batch_id.sql`; git merges cleanly and branch CI is green, but post-merge `TestMigrationVersionsAreUnique` fails. Please rebase, renumber to `0043`, and bump `migrate_review_dedup_test.go` (`upTo(t, db, 41)` -> `42`, plus the test name).

Two minor notes:

- `service/review/review.go:258`: the `run.BatchID != ""` guard is dead after this migration; the head check can be unconditional. Fine to fold in here.
- Backfilled legacy runs now get the current-head check, so an undelivered legacy run whose PR head moved will never be delivered. That seems correct (stale review), but worth a line in the PR body since it is a behavior change.

Merge after the renumber.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/service/review/review.go
```diff
@@ -230,12 +229,7 @@ func (s *Service) deliverSubmitted(ctx context.Context, workerID domain.SessionI
 		return nil, nil
 	}
 	results := reviewResults(workerID, deliverable)
-	var outcome lifecycle.ReviewDeliveryOutcome
-	if len(results) == 1 && results[0].BatchID == "" {
-		outcome, err = s.lifecycle.ApplyReviewResult(ctx, workerID, results[0])
-	} else {
-		outcome, err = s.lifecycle.ApplyReviewBatch(ctx, workerID, results[0].BatchID, results)
-	}
+	outcome, err := s.lifecycle.ApplyReviewBatch(ctx, workerID, results[0].BatchID, results)
```
neversettle17-101, line 232: When there is only one review result, the batchID = "" and it will go into [ApplyReviewBatch](https://github.com/VenkataSakethDakuri/agent-orchestrator/blob/6ac5afeae4c7016e00660184c083824b7fd77148/backend/internal/lifecycle/reactions.go#L52). It will return reviewNoOp. Please make sure the existing functionality doesn't break when we are cleaning up.
Can we check if for a single review result returns non nil batch ID?

neversettle17-101 (changes requested):

backend/internal/service/review/review.go
VenkataSakethDakuri: The trigger mints batchID before creating the first eligible run and assigns it to every run, even when there is only one result. @neversettle17-101

neversettle17-101 (changes requested):
Rename the migration file to avoid conflicts with main

neversettle17-101 (approved):

PR #3560, opened 2026-08-04
fix(frontend): fix topbar on session page

ronishrohan:
## Summary

image

- **Session page now has the same top inset as every other page** (board, settings). The `padding-top: 0` override on `.center-panel-shell--session` has been removed so the session panel respects the same inset tokens and collapses to zero when the macOS sidebar is hidden — exactly as requested.
- **`SessionTopbarHost` moved inside `CenterPanelShell`** — the tab bar is now the first child of `center-panel-surface` across all three platform rendering branches, rather than a standalone row above the content row.
- **Sidebar `topbarOffset` no longer special-cases sessions** — the sidebar no longer needs to skip `h-inspector-tabs` worth of height since that row is inside the panel now.
- **`session-topbar-surface` border/radius stripped when inside the surface** — a new contextual CSS rule drops the redundant outer border and border-radius (the surface provides both) and keeps only a bottom divider. The fullscreen in-place path is unaffected.
- **Removed viewport-alignment padding from the topbar wrapper** — `paddingLeft`/`paddingRight` in `CenterPane.tsx` were only meaningful when the topbar spanned the full window width; they caused extra space to the left of the first tab after moving the bar inside the panel.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Annieeeee11 (approved):
lgtm

ronishrohan:
thanks to @Prasad-D-Ware for finding this bug :)

Issue #3562, opened 2026-08-04
App must invoke its own bundled ao binary, never whatever is first on PATH


harshitsinghbhandari:
The desktop app (and the sessions it spawns) should always use the bundled binary at `/Applications/Agent Orchestrator.app/Contents/Resources/daemon/ao`, never resolve `ao` from PATH. If another `ao` CLI is installed (e.g. the legacy npm/Homebrew one), the app must still use its own bundled binary.

Real failure this causes: a stale `/opt/homebrew/bin/ao` (npm build from Jun 29, predating the `--name` flag) shadowed the bundled binary in shells that don't prepend the daemon dir, so `ao spawn` appeared to have lost the (required) `--name` flag. Cost a full RCA to discover it was PATH shadowing, not a regression.

Ask:
- Everywhere the app or daemon shells out to `ao` (worker session env, orchestrator sessions, hooks, tmux launch commands), use the absolute bundled path (or prepend the daemon dir to PATH for spawned sessions so `ao` resolves to the bundled binary first).
- Consider `ao doctor` warning when a different `ao` earlier on PATH shadows the bundled one (doctor already prints the resolved path; make the mismatch an explicit WARN).

illegalcall:
## Confirmed regression: tmux respawn loses AO's pinned `PATH` (2026-08-25)

**Source:** Dhruv Sharma's local ChatUI QA on macOS, followed by live runtime/process inspection and an isolated A/B reproduction  
**Analyzed against:** `main` at `04b676c9abd2fac4505081b0c099e079a30bde57`  
**Confidence:** High — the hook fired, the failing executable was identified, the parent/bad commits were compared under the same setup, and the environment mismatch was observed in both tmux and the real child process  
**Priority:** P1 — on an affected desktop install, every Claude Code callback that invokes bare `ao` can fail; the TUI-to-ChatUI transition is then hard-blocked. A coincidentally compatible global `ao` only masks the defect and is not a safe workaround.

### User-visible failure

A fresh Claude Code **orchestrator** starts successfully in Terminal UI, but **Switch to Chat UI** remains disabled with:

> session: native conversation id is not confirmed for the current terminal launch for claude-code

The user cannot switch interfaces.

<img width="1276" height="904" alt="Claude Code orchestrator switch to ChatUI disabled because the current launch has no confirmed native conversation id" src="https://github.com/user-attachments/assets/90d65a4a-9bd0-40dc-ae40-cc53738c9474" />

[...]

PR #3563, opened 2026-08-04
fix(daemon): preserve bundled ao PATH precedence during launches

harshitsinghbhandari:
Partially addresses #3562. The issue remains open for canonical AO-owned callbacks, post-startup shell behavior, and legacy-launcher/update reconciliation.

A stale `ao` installed beside an agent CLI can shadow the bundled daemon binary after launch-time PATH augmentation. Keep the daemon directory first for worker, orchestrator, reviewer, shell-terminal, and Codex Chat launches. Reviewer shim fallback retains the same ordering. Interactive shell startup files can still reorder PATH after launch.

`ao doctor` compares the PATH binary with the running daemon's `/healthz` executable identity, falling back to its own executable when no daemon identity is available. `ao status --json` exposes that identity as `executablePath`.

Updated against upstream main at a merge commit without rewriting the existing PR branch. The shared `agentlaunch` implementation preserves current agent and Node runtime discovery. Codex Chat also retains the pin when it augments the environment a second time. Windows Chat environment merging normalizes case before applying overlays so an inherited `Path` cannot replace the pinned `PATH`; a real `cmd.exe`/`ao.exe` identity test runs in Windows CI.

Validation for head `6b4fd8167f794689c1be0519804ed4101c36077f`:
- Session and reviewer shadowing tests fail with pin restoration removed, and pass with the fix.
- Focused session, reviewer, shell-terminal, doctor, Node runtime, and Codex Chat subprocess regressions pass.
- `go build ./...` and `go vet ./...` pass.
- `npm run api` produces no generated artifact drift.
- Native CLI E2E passed locally with inherited AO variables stripped.
- Complete Chat environment and Codex driver packages pass locally with the race detector. The real Windows subprocess identity regression passes in Windows CI.
- All 11 exact-head CI checks pass, including full Go build/vet/race tests, lint, API drift, three-platform CLI E2E, fresh-install container, Windows environment coverage, frontend tests, renderer smoke, and secret scanning.
- Local full race execution hit installer cancellation timing (three isolated retries passed), a Chat timing failure, and 15-minute Chat/SQLite timeouts on this shared host. The clean-runner full race job passes.
- Pinned golangci-lint passes in exact-head CI. Local isolated-cache lint printed zero issues but exceeded its timeout; it is not counted as a final local pass.

Scope: this PR fixes launch-time ordering and daemon-aware diagnostics. It does not override user shell startup files or repair external AO installations.

No replacement PR. Merge requires a clean exact head, green CI, and the repository's required review.


harshitsinghbhandari:
Rebased by merge onto current main at 10cfce3d6 and resolved the PATH-pin conflicts against the shared agentlaunch architecture. All required CI is green on head 12e382070. Ready for review, @illegalcall.

backend/internal/service/shellterm/service.go
```diff
@@ -242,6 +254,7 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 	}
 	return s.openTerminal(ctx, openTerminalConfig{
 		argv:       argv,
+		env:        s.pinnedEnv(),
```
harshitsinghbhandari, line 257: [P1] Enforce the pin after shell startup, not only in the initial environment

`resolveUserLoginShell` returns the user's shell; an interactive zsh reads `.zshrc` after receiving this environment. With an isolated HOME/ZDOTDIR, bundled and foreign executable fixtures, and `.zshrc` containing `export PATH="<foreign>:$PATH"`, I passed the real OpenShellTerminal output to zsh and bare `ao` printed FOREIGN. Without startup files it prints BUNDLED. Thus the in-app terminal still reproduces #3562 immediately, including ordinary `.local/bin` or Homebrew setup. The PR body acknowledges this limitation, but `Closes #3562` and the every-context guarantee are not satisfied.

Preserve canonical executable identity through the post-startup boundary (and use absolute identity for AO-owned callbacks), then add an actual interactive-shell regression that asserts which `ao` runs. If this is intentionally only a launch-order fix, keep the remaining issue open and explicitly narrow the closure scope rather than claiming this case is fixed.

backend/internal/adapters/chatdriver/codexappserver/driver.go
```diff
@@ -594,6 +594,6 @@ func codexProcessEnv(ctx context.Context, bin string, env map[string]string) []s
 	if _, ok := overlay["PATH"]; !ok {
 		overlay["PATH"] = os.Getenv("PATH")
 	}
-	agentlaunch.AugmentRuntimePATHForLaunchBinary(ctx, overlay, []string{bin}, exec.LookPath)
+	agentlaunch.AugmentRuntimePATHForLaunchBinary(ctx, overlay, []string{bin}, exec.LookPath, agentlaunch.PinnedDir(os.Executable))
```
harshitsinghbhandari: [P1] Preserve the pin through Windows environment deduplication

The next line calls envSlice/processenv.Merge. That merge keys case-sensitively and sorts keys. If the daemon inherited Windows `Path=<foreign>`, adding `PATH=<bundle>;<foreign>` leaves both entries, ordered uppercase PATH first and mixed-case Path second. Go os/exec removes duplicate keys case-insensitively on Windows and keeps the last value, so the actual provider loses the bundled pin. This also affects the persistent-host route through envSlice at line 422. This is an existing downstream hole that the new preservation call does not fix.

Normalize environment keys on Windows before applying the overlay, so one PATH entry survives with overlay precedence. Add a Windows subprocess test seeded with mixed-case Path and a competing ao.exe, asserting the child runs the bundled executable. The new test only reads the uppercase entry in the slice and would miss this loss. A temporary test against the real Merge reproduced the duplicate ordering; the platform reduction was verified from Go os/exec's dedupEnvCase implementation.

harshitsinghbhandari (commented):
GitHub rejected REQUEST_CHANGES because this authenticated account is the PR author. This COMMENTED review records the independent verdict: **REQUEST CHANGES**, for exact head `3facf03ab5c7e18775550747edbaebb7e442fcf4`.

## Review of exact head 3facf03ab5c7e18775550747edbaebb7e442fcf4

**Verdict: REQUEST CHANGES.** The augmentation ordering fix is sound for an already-pinned environment, including the second Codex direct-launch augmentation. It does not yet establish the issue's executable-identity guarantee at the shell and Windows subprocess boundaries. Two actionable findings are attached inline.

I read the PR body, issue #3562 and all its comments, the complete 13-file diff against latest upstream/main `0244fb8aef6baf157fb77cb4f37ffe96732b1801`, the timeline, and all review/thread endpoints (no existing reviews or inline threads). I inspected the superseded `12e382070` merge resolution and the exact head's remerge diff. The current merge retains main's shared agentlaunch implementation, shell command-terminal lifecycle, and Node discovery; the earlier empty Codex pin argument is corrected at this head. The old comment reporting green CI for `12e382070` is not evidence for this head.

### Boundary verification

FAIL: five boundary groups lack passing executable-identity/propagation coverage. No deferrals or overrides applied:

- Session Manager PATH output (`manager.go:4736`) to tmux/conpty child execution: Spawn/Restore tests inspect fake runtime configuration. Add a real child resolving competing bundled/foreign `ao` on create and restart.
- Reviewer PATH output (`launcher.go:556`) to reviewer child execution: configuration assertions do not exercise the shim or actual bare command. Cover normal pin and shim fallback with a real child.
- Shell terminal output (`service.go:257`) to interactive shell startup (`loginshell.go:27`): the temporary real-shell regression fails, executing FOREIGN.
- Codex PATH output (`driver.go:597`) through `processenv.Merge` to direct/persistent subprocess execution: the Windows case-insensitive handoff regression fails. The new AO-named subprocess test stops at inspecting the environment slice; it never launches the provider or resolves bare `ao` from that environment.
- Daemon `/healthz` identity through `status.go:120` to `doctor.go:182`: tests call `checkAOBinary` with a supplied identity but do not verify HTTP identity propagation. Add a mocked HTTP daemon with a different executable path, and an old response omitting the field; assert status JSON and doctor warning/fallback.

### Validation and limits

Focused PATH, HookPATH, doctor, launch-command, Codex version/Node, and environment-merge tests passed across nine relevant package targets (agentlaunch has no tests). Complete review, shellterm, and Codex driver packages passed. Complete session_manager and CLI packages passed after stripping inherited AO_* variables; their initial run was contaminated by the review session's environment.

Temporary Go overlay tests left repository source untouched: real shellterm output followed by zsh startup executed FOREIGN despite the supplied bundled prefix; real processenv.Merge emitted `PATH=bundle;foreign` before `Path=foreign`, so Windows os/exec's documented case-insensitive, last-entry-wins reduction selects foreign. The latter was modeled on macOS and checked against Go's implementation, not executed on Windows.

macOS subprocess behavior was exercised. Linux and Windows runtime execution was not performed locally. Native CI smoke checks do not exercise these particular boundaries. PATH separators and Windows ao.exe basename handling are preserved; the changed map/slice operations introduce no shell interpolation, and tmux still quotes values. Existing renamed-binary fallback and optional status JSON field preserve compatibility. AppImage lifetime across updates and Windows PATHEXT/current-directory lookup remain without executable-identity regression coverage.

CI at review time: eight checks pass; build-test remains pending. No merge sign-off. CodeGraph was unavailable in this worktree; source tracing was manual. The edge-case-hunter skill was unavailable, so input, quoting, and platform edge analysis was performed manually.

backend/internal/service/shellterm/service.go
harshitsinghbhandari: Addressed through the scope-narrowing option requested here: removed Closes #3562, changed the title to launch-time PATH precedence, and explicitly documented post-startup shell behavior and external launcher reconciliation as remaining work on #3562. This PR preserves user startup files and no longer claims complete executable-identity enforcement.

backend/internal/adapters/chatdriver/codexappserver/driver.go
harshitsinghbhandari: Fixed in e4c212b2d: processenv.Merge normalizes Windows environment keys before applying the overlay, so inherited Path cannot override the pinned PATH. Added TestMergeWindowsPATHExecutesBundledAO, which seeds mixed-case Path with a competing ao.exe, invokes cmd.exe with the real merged environment, and checks the actual child executable identity. The Windows CI job now runs this package. Native Windows verification is pending CI.

harshitsinghbhandari: Verified on exact head 6b4fd8167f794689c1be0519804ed4101c36077f: windows-workspace passed, including the new real cmd.exe/ao.exe identity regression. All 11 CI checks are green. The follow-up sorting fix preserves the existing deterministic-environment assertion for Windows prefix-related variable names.

illegalcall:
Two things that could break after this change:

1. **Windows scripts can lose project variables.** The environment merge uppercases every variable name. For example, if a project sets `buildMode=production`, a Bash script reading `$buildMode` works today. After this change, only `BUILDMODE` exists, so the script gets an empty value. Can we compare variable names without caring about case, but keep their original spelling when passing them to the child process?

2. **Some custom installs could start agents with the wrong Node version.** If someone puts `ao` in `/usr/local/bin` alongside an older Node, moving that whole directory first also puts the older Node ahead of the newer one beside their NVM-installed agent CLI. A launcher using `#!/usr/bin/env node` can then pick the wrong runtime and fail. The normal desktop bundle avoids this because `ao` has its own directory. For shared-directory installs, could we use a directory containing only an `ao` shim so other tools keep their existing priority?

It would be good to cover both with regression tests. I checked these against head `6b4fd8167`: the Node selection change reproduces with fixture subprocesses on macOS. The variable-name change was checked by exercising the Windows merge logic locally and passing its output to Bash; I have not run that case on native Windows.

harshitsinghbhandari:
Addressed both findings from [your comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/3563#issuecomment-5554581829) in head `053b345a6f5a2693d80f2881fc6fc768ad6e908c`.

1. Windows environment merging now folds keys only for comparison and retains the complete original entry for the child. Overlay spelling wins when an overlay replaces an inherited variable. `TestMergePreservesVariableSpellingForBash` exercises real Bash with both `buildMode=production` and an inherited mixed-case variable. Before the fix, running the Windows branch through a Go overlay on macOS produced empty variables; after the fix it prints `production:unchanged`. The existing native Windows PATH precedence regression remains in CI.

2. Shared AO install directories now get an isolated `ao` shim under the AO data directory. Install directories without executable siblings keep their existing direct pin. This preserves canonical AO resolution without promoting an older sibling Node or other executable. `TestSharedInstallKeepsAgentNodeAndCanonicalAO` launches a real `#!/usr/bin/env node` fixture plus bare `ao`: before the fix it printed `CANONICAL / OLD_NODE`; after the fix it prints `CANONICAL / AGENT_NODE`. The fixture includes spaces and an apostrophe in the install path. `TestSharedWindowsInstallSelectsOnlyCanonicalAO` covers real executable selection through `cmd.exe` with competing `ao.exe` and `node.exe` files.

Local verification: `go build ./...`, `go vet ./...`, pinned golangci-lint (zero issues), and race tests for agentlaunch, processenv, Codex Chat, session manager, reviewer, and shell terminals pass.

All 11 checks pass on this exact head, including the full Go race suite and native Windows Bash/command regressions: [Go CI](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34097593027), [three-platform CLI E2E](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34097592922), and [frontend CI](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34097592923).

Scope remains these two findings. No merge or review request.

illegalcall:
I tested the updated head `053b345a6`. The original variable-spelling and Node-selection fixes pass, but I found three remaining issues:

1. **A relative `AO_DATA_DIR` makes the shim point to the wrong place.** With `AO_DATA_DIR=data` and a shared install directory, the new code puts `data/runtime/ao-path/...` on PATH. When the agent starts in its worktree, that relative path is resolved from the worktree instead of the daemon's data directory. A real subprocess test executed the foreign `ao` instead of the canonical one. Please use the absolute state directory already resolved at startup; resolving the environment value again after the daemon changes directory also creates the wrong nested location. See `agentlaunch/pin.go:54–67`.

2. **A project's `Path` can still replace AO's pinned `PATH` on Windows.** Project settings can leave both spellings in the overlay. The merge treats them as the same key, but picks the winner in Go map iteration order. In my Windows-branch probe, the foreign value won 130 of 1,000 merges. This was exercised on macOS with the Windows branch enabled, not on native Windows; the count demonstrates nondeterminism, not a production failure rate. Please normalize project keys before applying AO's protected values and add a test with both spellings in the overlay. See `processenv/processenv.go:24–29`.

3. **The new Windows shim only works for some callers.** Shared installs get `ao.cmd`, but Git Bash's executable search tries `ao` and `ao.exe`, so bare `ao` can skip the shim and reach a later stale installation. The new Windows test uses `cmd.exe`, which doesn't catch this. Please add a native Windows test that runs bare `ao` through Git Bash with a competing foreign binary. This finding is based on the shim code and [Git for Windows' executable suffix table](https://github.com/git-for-windows/msys2-runtime/blob/main/winsup/cygwin/path.cc#L65-L70); I couldn't run native Windows locally. See `agentlaunch/pin.go:70–72`.

What passed locally: race tests for agentlaunch, processenv, Codex Chat, reviewer, and shell-terminal packages; focused session-manager and CLI tests; mixed-case variables reaching Bash; AO/Node selection; and POSIX shim argument preservation under 32 concurrent launches. I also checked the successful native Windows CI step, but its current tests don't cover the remaining cases above.

I'd address these before approval. No product files were changed during validation.

harshitsinghbhandari:
Addressed all three findings on exact head `b3864c2d14b9ed4dfc2b268e421405c3f192a286`.

1. Shim creation now receives the resolved absolute data directory from session manager, reviewer, shell-terminal, and Codex Chat owners. It no longer reads or resolves `AO_DATA_DIR` after startup changes the daemon cwd. The regression sets `AO_DATA_DIR=data`, changes into a separate worktree, and proves bare `ao` executes the canonical binary through an absolute shim below the supplied state directory. A second test rejects relative shim state directories.

2. Windows environment merging now sorts overlay keys before case-insensitive deduplication, preserves original spelling, and applies exact protected `PATH` after any `Path` variant. Session environment construction also removes case variants before installing AO protected values. The regression runs 1,000 merges containing both spellings and requires the sole result to be `PATH=ao-pinned`; the native Windows subprocess test supplies conflicting `Path` and `PATH` values and executes the canonical AO.

3. Shared Windows installs now create `ao.cmd` for cmd.exe and an `ao.exe` hard link for Git Bash executable lookup. The native Windows regression invokes bare `ao` through both cmd.exe and Git Bash with a competing foreign `ao.exe`, verifies canonical file identity, and separately verifies the agent-selected `node.exe` still wins.

Local verification passed: focused race tests for agentlaunch, processenv, Codex Chat, reviewer, and shell-terminal packages; focused session-manager tests; `go build ./...`; `go vet ./...`; pinned golangci-lint with zero issues; and Windows cross-compilation.

Exact-head CI evidence: native Windows, macOS, and Linux CLI E2E, the dedicated Windows PATH/Git Bash regressions, frontend tests, renderer smoke, lint, API drift, container install, and secret scanning all pass. The full Go race job is still in progress with no reported failure: https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34116577167

No scope expansion, review request, or merge performed.

illegalcall:
The three previously reported issues are addressed in `b3864c2d`. I reran the affected package and boundary tests with the race detector; they passed, and the native Windows CI now covers Git Bash.

One Windows edge case remains: **the new hard link cannot cross drives**.

For example, someone has `ao.exe` and `node.exe` in `C:\Tools`, but uses `AO_DATA_DIR=D:\ao-state`. That triggers the shared-install shim, but `os.Link` in `agentlaunch/pin.go:119–121` fails because the executable and shim are on different volumes. The error goes back through `PinnedPATH`, and session launches continue without the pin. A stale `ao` can then win again. The previous batch wrapper did not have this restriction.

Can we add a fallback for this case and a Windows regression using separate volumes? The current test creates both locations with `t.TempDir()`, so it only covers the same-volume case. The normal desktop layout avoids this; it affects shared-directory/custom installations.

I verified the code path and [Microsoft's same-volume restriction](https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions). I have not run a native Windows cross-volume reproduction.

harshitsinghbhandari:
Addressed the cross-volume Windows hard-link finding from [your comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/3563#issuecomment-5572103924) on exact head `7d67d92c28b71e1a94ab7ea5b98efc92f7541b38`.

The isolated Windows shim still attempts `os.Link` first. If linking fails, it copies the canonical executable into a temporary file inside the isolated shim directory, preserves its modification time, and installs it as `ao.exe`. Bare `ao` therefore remains available to Git Bash without putting the canonical executable's shared directory, or sibling tools such as Node, ahead on PATH.

`TestWindowsAOExecutableFallsBackToCopyAcrossFilesystems` forces the link operation to return the cross-volume failure, verifies the copied executable is byte-identical, then invokes bare `ao` through Git Bash and verifies that the copied shim executable runs. This test runs natively in the Windows Go job. The GitHub Windows runner does not provide a guaranteed second volume, so the test injects the filesystem-boundary error rather than claiming an actual two-drive setup.

Local verification passed: focused agentlaunch race test, package vet, full backend build, pinned golangci-lint with zero issues, and Windows cross-compilation. All 11 exact-head CI checks pass, including the native Windows regression and full Go race suite: https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34137215212

No scope expansion, review request, or merge performed.

illegalcall (approved):

PR #3435, opened 2026-08-01
fix(session-manager): include install hint in tmux prerequisite error

i-trytoohard:
Fixes #3434

## Summary

When tmux is not installed and a user tries to spawn a session, the error message previously gave no guidance on how to fix it:

```
runtime: prerequisite missing: tmux required on macOS/Linux but not in PATH
```

Now includes a platform-specific install command:

```
runtime: prerequisite missing: tmux required on macOS/Linux but not in PATH — install it: brew install tmux
```

Added a `tmuxInstallHint(goos string)` helper that returns the correct package-manager command per platform:
- **macOS:** `brew install tmux`
- **Linux:** `sudo apt install tmux` (with dnf/pacman alternatives noted)
- **Other:** generic fallback

## Scope

This PR addresses **Layer 1** (actionable error message) from issue #3434. Layers 2 (proactive startup check) and 3 (frontend error UX) are left for follow-up since they require broader changes.

## Test

```bash
cd backend && go build ./... && go test ./internal/session_manager/ ./internal/service/session/ -count=1
```

Existing test at `manager_test.go` verifies the error still wraps `ports.ErrRuntimePrerequisite` and contains "tmux required" — both still pass with the enhanced message.

Issue #3434, opened 2026-08-01
tmux prerequisite error gives no install guidance; only checked at spawn time


i-trytoohard:
## Bug / Enhancement

New users on macOS/Linux who don't have `tmux` pre-installed hit an opaque error on their first spawn attempt. The error message tells them tmux is missing but gives **zero guidance on how to install it**, and the prerequisite is only checked at spawn time — not at startup or onboarding — so the user discovers the problem only after they've already configured a project and tried to start a session.

**Source:** LinkedIn feedback | **Reported by:** Praveenkumar K | **Analyzed against:** `2d0b70f74`
**Confidence:** High

## Reproduction

1. Fresh install of Agent Orchestrator on macOS or Linux **without** tmux installed.
2. Open the app, configure a project, select an agent, and try to spawn a session.
3. Error returned:

```
runtime: prerequisite missing: tmux required on macOS/Linux but not in PATH (RUNTIME_PREREQUISITE_MISSING)
```

4. User must independently figure out they need to install tmux (e.g. `brew install tmux`).

## Root Cause

The prerequisite check is correct but the **messaging and timing are unhelpful**:

**1. Bare error message with no install instructions**

`validateRuntimePrerequisites()` in `backend/internal/session_manager/manager.go:3314` returns:

```go
return fmt.Errorf("%w: tmux required on macOS/Linux but not in PATH", ports.ErrRuntimePrerequisite)
```

This propagates through `service.go:670` → `apierr.Invalid("RUNTIME_PREREQUISITE_MISSING", ...)` with the same bare text. No platform-specific install command is included.

**2. Checked only at spawn time, not at startup**

The daemon (`backend/internal/daemon/daemon.go`) and `ao start` (`backend/internal/cli/start.go`) do **not** run a tmux preflight. The user only discovers the missing dependency after they've gone through project setup and attempted a spawn.

**3. `ao doctor` checks it but isn't run automatically**

`backend/internal/cli/doctor.go:307` (`checkTmux`) does detect the missing binary and reports it as a warning — but `ao doctor` is a manual command that new users won't know to run.

**4. Frontend shows raw error text**

`frontend/src/renderer/lib/api-client.ts:252` (`apiErrorMessage`) has no special handling for `RUNTIME_PREREQUISITE_MISSING`. The NewTaskDialog surfaces the raw backend message with no actionable UI (no install hint, no docs link).

## Suggested Fix

[...]

PR #3566, opened 2026-08-04
feat: make a missing tmux visible and fixable, in the app and the CLI

### LINKED ISSUE #3564: feat: one-time interactive tmux auto-install at startup (like pre-rewrite ensureTmux)
## Feature

Implement one-time interactive tmux auto-install in the Go rewrite, mirroring the pre-rewrite TypeScript `ensureTmux()` behavior. When tmux is missing at startup, prompt the user and install it automatically — instead of failing with a bare error at spawn time.

**Source:** Bug triage thread (Praveenkumar K report → team discussion)
**Decision (per @Dhruv Sharma):**
- ❌ **Do NOT bundle/package tmux** with AO — causes version conflicts if user already has tmux, and creates a patching/maintenance burden.
- ✅ **DO implement one-time interactive auto-install** via CLI at startup, like the pre-rewrite TS code did.

Related: [#3434](https://github.com/Untrivial-ai/agent-orchestrator/issues/3434) (error message improvement — already addressed by [PR #3435](https://github.com/Untrivial-ai/agent-orchestrator/pull/3435))

## What the old TypeScript code did

The pre-rewrite `ao start` had an `ensureTmux()` function (`packages/cli/src/commands/start.ts`, ~L1070 in the old codebase) that:

1. Checked if tmux was on PATH (`execSilent("tmux", ["-V"])`)
2. If missing, printed a warning and prompted interactively: `⚠ tmux is required for runtime "tmux". Install tmux now? (Y/n)`
3. If user said yes, ran `tryInstallWithAttempts()` with platform-specific commands:
   - **macOS:** `brew install tmux`
   - **Linux:** `sudo apt-get install -y tmux` → fallback `sudo dnf install -y tmux`
4. Verified the install succeeded by re-checking `tmux -V`
5. If install failed or user declined, printed manual install hints and exited

This covered ALL `ao start` paths (normal, URL start, retry with existing config).

In non-interactive mode, it skipped the prompt and showed the manual instructions only (`start.test.ts` L870: "does not auto-install tmux when missing in non-interactive mode").

## Suggested Implementation (Go rewrite)

**Where:** The daemon startup path or `ao start` CLI command.

**What to build:**
1. A `ensureTmux()` equivalent in the Go CLI (`backend/internal/cli/`) or daemon startup (`backend/internal/daemon/`) that:
   - Runs `exec.LookPath("tmux")` (reuse the existing check in `validateRuntimePrerequisites` at `manager.go:3314`)
   - If missing and interactive: prompt user, run platform-specific install command
   - If missing and non-interactive: show install hints, exit with clear error
2. Platform install commands (same as old TS):
   - **macOS:** `brew install tmux`
   - **Linux:** `sudo apt-get install -y tmux` (fallback: `sudo dnf install -y tmux`)
3. Call this at daemon/startup time — NOT at spawn time — so the user fixes the dependency before configuring a project.

**Key consideration — daemon vs CLI process split:** In the Go rewrite, the daemon is a separate long-lived process. The interactive prompt needs to happen in a context where stdin is available (CLI `ao start`, or the Electron app's onboarding flow), not inside the daemon itself.

## Related

- [#3434](https://github.com/Untrivial-ai/agent-orchestrator/issues/3434) — original report: tmux prerequisite error gives no install guidance
- [PR #3435](https://github.com/Untrivial-ai/agent-orchestrator/pull/3435) — Layer 1 fix: actionable error message with install hint (complementary, not a replacement for this)

harshitsinghbhandari:
Closes #3564. Related: #3434, #3435.

## Problem

A new macOS/Linux user without tmux only finds out at spawn time, via
`RUNTIME_PREREQUISITE_MISSING: tmux required on macOS/Linux but not in PATH`.
That check runs inside the daemon, which lives inside the desktop app, so it
fires in the one place with no way to tell the user anything and no terminal to
fix it from.

## The desktop app

A non-blocking banner on the board, styled as the project-health banner directly
beneath it:

- **macOS:** an **Install tmux** button that runs `brew install tmux` through the
  daemon, then re-checks PATH. Homebrew needs no password, so it is safe to run
  headless.
- **Linux:** the exact `sudo apt-get install -y tmux` line to copy. The app has
  no terminal to answer a password prompt on, so it does not offer a button it
  cannot honour.
- **Both:** a **Check again** button, so a manual install clears the banner
  without restarting the app.
- An unreachable daemon renders nothing. Not knowing is not the same as knowing
  tmux is absent.

### Why not run the install in an AO terminal pane

That was the first idea and it does not work. `daemon.go:220` wires standalone
shell terminals through the same runtime adapter as session panes, which is
tmux. No tmux means no shell terminal either, so the install cannot be hosted in
one. Hence the split above.

`pkexec` would give Linux a polkit password dialog and close that gap. Left out
deliberately: a second privilege-escalation path deserves its own change, tested
by someone who runs Linux daily.

## The CLI

`ao start` settles the same prerequisite before launching the app, which is what
#3564 asked for and mirrors the pre-rewrite `ensureTmux()`. On an interactive
terminal it prompts and installs (brew, apt-get, dnf, pacman, zypper, apk, first
match on PATH, sudo only where root is needed and reachable). Anything else gets
one warning naming the exact command.

**One deliberate deviation from the issue:** it asks for "show install hints,
exit with clear error". This warns and continues. The old `ensureTmux()` could
exit because `ao start` *was* the daemon; it is now only the app launcher, and
the app is usable without a session runtime. The CLI smoke container proved the
point: it runs `ao start` on a fresh box and asserts the failure is a download
error, and a hard tmux exit broke that check.

Not bundling tmux, per the triage thread: it would conflict with an existing
install and add a version-patching burden.

## Shape

- `internal/prereq` holds the package-manager table, so the CLI and the daemon
  cannot drift on what installs tmux. `Status.NeedsRoot` decides who may run it.
- `GET /api/v1/prerequisites`, `POST /api/v1/prerequisites/tmux/install`. The
  install endpoint refuses anything flagged `NeedsRoot` with
  `PREREQUISITE_NOT_INSTALLABLE`, so the rule holds however the request arrives,
  not just because the UI hides a button.
- A package manager can exit 0 and leave nothing runnable, so success is the PATH
  re-check, not the exit code (`PREREQUISITE_INSTALL_INCOMPLETE`).
- New `Deps.RunInteractive` in the CLI runs the install attached to the real
  terminal so `sudo` can prompt; its stdout goes to stderr so `ao start --json`
  stays a clean JSON document.
- `stdinIsInteractive` excluded `/dev/null`, which is a character device and so
  passed the old "is a terminal" test. `docker run` without `-i` hands a process
  exactly that, which made the confirmation prompt answer itself with its default
  and run an install nobody asked for. This also makes `ao import`'s confirmation
  honest in the same situation. A real tty check needs an ioctl
  (`golang.org/x/term`), which pulled an `x/sys` bump into `go.mod` for one
  boolean; `/dev/null` is the only non-tty character device that turns up as
  stdin in practice.
- `openapi.yaml` and `schema.ts` regenerated via `npm run api`. Copy added to all
  8 locales.

Complementary to #3435 (better error text at spawn time), which stays the
backstop for anyone who dismisses or ignores the banner.

## Verification

- `go build ./...`; `golangci-lint run` over `internal/{prereq,httpd,daemon,cli}`: 0 issues.
- `go test ./internal/prereq/ ./internal/httpd/... ./internal/cli/`: pass.
- `npx vitest run`: 109/110 test files pass. The one failure,
  `src/landing/scripts/generate-markdown-twins.test.mjs`, is a missing local
  `cheerio` install, untouched here.
- Two pre-existing backend failures
  (`TestWorkspaceFilesIncludeWorkspaceProjectChildRepoDiffs`, and six
  `TestSpawn*`/`TestBrowser*` cases that talk to a live local daemon) reproduce
  identically on `upstream/main` in a clean worktree at the base commit.

Tests added: `prereq` (platform/ordering/root rules), `PrerequisitesController`
(status per platform, install, refusal to run what it cannot, failure and
exit-0-without-tmux), `cli/tmux` (no-ops, warn-without-install per unreachable
case, the `/dev/null` regression, prompt-and-install, decline, failed install,
sudo rules), `PrerequisiteBanner` (present, unreachable daemon, install, copy,
failure, re-check).

**Not done:** I have not seen the banner render in the real app. It only appears
when the daemon reports tmux missing, and this machine has tmux. Worth a look on
a fresh box before merge.


harshitsinghbhandari:
Superseded by #3825, which now carries all of this.

Closing rather than merging because #3825 covers a strict superset. It checks git, tmux, harness, and gh instead of tmux alone, gates startup rather than showing a banner, and has absorbed every part of this PR that it did not already have:

- `backend/internal/cli/tmux.go` and its tests (all 10 ported), so `ao start` from a terminal detects a missing tmux and offers to install it
- `RunInteractive` on the CLI `Deps`, so an installer can prompt for a sudo password, with its stdout routed to stderr so `--json` output cannot be corrupted
- the `stdinIsInteractive` `/dev/null` fix, which stopped CLI prompts from silently answering themselves under `docker run` without `-i`
- apk/Alpine in the package manager list
- the platforms.mdx notes

The daemon-side surface here (`/api/v1/prerequisites`, `controllers/prerequisites.go`, `PrerequisiteBanner`, `usePrerequisites`, the `prereq.*` i18n keys) was intentionally dropped instead of ported, since `/api/v1/system` replaces it.

Worth recording why the CLI half was kept rather than discarded with the rest: the daemon deliberately refuses to elevate privileges, so on Linux its install path can only print `sudo apt-get install -y tmux` for the user to run. The CLI runs in a real tty where sudo can actually prompt, so folding it in is what makes Linux install work end to end. #3825 exposes the install argv from its own planner for this, which means there is one source of truth for how tmux gets installed rather than the two a stacked PR would have left behind.

No code is lost by closing this.

PR #3579, opened 2026-08-04
fix(browser): prevent wrong tab content after switching

### LINKED ISSUE #3445: [Bug]: Browser panel shows the wrong tab's content after switching tabs (native view bleed-through)
### Summary

After switching between tabs in the in-app Browser panel, the panel can display a *different* tab's content than the one that's actually active — the address bar shows the correct/active tab's URL, but the visible page is a stale render of a previously-active tab. It looks like the native `WebContentsView` for the tab being switched away from isn't always fully hidden, so it renders on top of (or instead of) the tab that should now be showing.

This reproduces with plain mouse interaction (the tabs dropdown), so it isn't tied to any particular way of creating/switching tabs — it just becomes much easier to hit once a session has 2+ tabs open, which is now common after #3444 added a "+" new-tab button and keyboard shortcuts (Ctrl+T etc.) to the panel. Before that, only an AI agent or a `target=_blank` popup could create extra tabs, so most sessions never had more than one.

### Steps to reproduce

1. Open a session with the Browser panel visible, with its one default tab.
2. Click the tabs icon → **+** to open a new tab (tab 2).
3. Navigate tab 2 to `https://google.com`.
4. Click **+** again to open a third tab (tab 3) — leave it blank/untouched.
5. Open the tabs dropdown and select tab 2 (Google) again.
6. Open the tabs dropdown and select tab 3 (the blank tab) again.
7. Observe the panel content.

(This also reproduces with just two tabs and repeated switching: create tab 2, navigate it to Google, create tab 3 blank, switch back and forth a couple of times.)

### Expected behavior

Each tab switch should show that tab's own content, matching its own address bar/URL. Switching to a blank tab should show a blank page; switching to the Google tab should show Google — never a mix of the two.

### Actual behavior

The panel shows content from a tab that isn't the one currently marked active. In one repro, the address bar read `localhost:5173` (the session's original/default tab) while the visible page was Google's homepage (a different tab's content) — confirmed via screenshot during triage, not attached here.

The user-visible effect is described as the previous tab's content "bleeding through" or leaving an "impression" on top of the tab that should be showing.

### Suspected cause / what's been ruled out

This is in the native tab-activation path in `frontend/src/main/browser-view-host.ts`, specifically `activateTab()` and `applySessionBounds()` — the functions that hide the previously-active tab's `WebContentsView` (move it off-screen + `setVisible(false)`) and show the newly-active one. This code predates #3444; it's reachable through the pre-existing tabs dropdown, the pre-existing AI-agent `tab-select`/`tab-new` actions, and popup-tab creation — not just the new "+" button/shortcuts.

Two fixes were attempted and did **not** fully resolve it (both reverted, not merged):
1. Re-raising the newly-shown tab's view to the top of Electron's `contentView` child stack on every show (`contentView.addChildView()` on an already-parented view reorders it to the front, per Electron's `View` docs) — theory being that `createTab()` only adds a tab's view to the stack once, at creation, so an older tab reactivated later stays stacked *below* every tab created after it.
2. Making `activateTab()` explicitly hide **every** other tab in the session on each activation, instead of trusting a single tracked "previously active" tab reference (in case an earlier hide silently didn't fully apply and never got corrected).

Neither fix eliminated the repro, which suggests the root cause may be elsewhere — worth also looking at:
- Whether Electron's `WebContentsView.setVisible(false)` / off-screen `setBounds()` calls are reliably taking effect for `WebContentsView` specifically (vs. the deprecated `BrowserView`).
- The renderer's snapshot-based tab-switch masking overlay (`showVisualTransition("tab-switch")` in `frontend/src/renderer/hooks/useBrowserView.ts`, rendered as an `` in `BrowserPanel.tsx`) — it captures a screenshot of the tab being switched *away from*, shows it as a temporary overlay, and auto-clears after `VISUAL_TRANSITION_DURATION_MS` (240ms) via `window.setTimeout`. If that overlay's clear-timer is somehow cancelled or its state never resets, a stale screenshot would visually match this exact symptom (a static "ghost" of a previous tab). Worth confirming whether the bleed is a live native view (would react to interaction/scrolling on the wrong tab) or a frozen image (wouldn't).
- Any interaction between the renderer's periodic `browser:setBounds` sync (from a resize/measure effect) and the main process's own `activateTab()`-driven bounds/visibility changes, in case of a race between the two.

### AO version

commit `9caafbee8` (`main`, 2026-07-31) / Electron `^33.0.0`. Not yet in a tagged release — found during development of #3444.

### Environment

Windows 11, Electron dev build (`npm run dev` / electron-forge + Vite).

### Logs / extra context

Related PR: #3444 (added the "+" new-tab button and browser-scoped keyboard shortcuts) — that PR does **not** contain a fix for this; it was investigated there and explicitly reverted/split out into this issue since it's a pre-existing bug in tab-activation, not something the new feature introduced.

### Before submitting

ApexYash11:
## What

Fix the Browser panel so switching logical tabs cannot leave a stale renderer-owned frame covering the active native WebContentsView.

## Why

Fixes #3445.

The tabs dropdown captures the native browser so renderer UI can appear above it. Cleanup previously depended on indirect DOM/timer observation, and tab selection also started a second outgoing snapshot transition. Those paths could leave the old tab frame visible after the active native tab changed.

## How

- add explicit overlay cleanup to the browser view model
- release the tab-menu mirror after selection, failure, and dismissal
- remove the outgoing tab-switch snapshot transition
- cancel captures that resolve after overlay cleanup
- cover repeated native tab activation and renderer cleanup races

## Testing

- npx vitest run --config vite.renderer.config.ts src/renderer/hooks/useBrowserView.test.tsx src/renderer/components/BrowserPanel.test.tsx src/main/browser-view-host.test.ts — 116 passed
- npm run typecheck — passed
- npm run package — passed on Windows
- Manual Electron verification: repeatedly switched between Example Domain and a blank tab; content and address bar stayed matched and the tabs menu remained usable
- Full frontend suite: 1,449 passed, 21 unrelated Windows/environment failures in macOS artifact tests, sandbox-blocked supervisor socket tests, and landing tests missing cheerio

## Checklist

- [x] Branched from main
- [x] Relevant browser tests, typecheck, and production packaging pass


Vaibhaav-Tiwari (changes requested):
while i was on your pr and testing it on my AO, i opened 5 tabs and when i did the tabs icon got dimmer. even though you can leave out this issue because i would be revamping thee brower tabs itself, you can just find out the reason. 

Image

and here are some suggestions, even tho they aren't blockers:
  1. The PR includes its own planning documents — two "notes to self" files (~480 lines) that this project's history shows should normally be deleted before
  merging. Ask the author to remove them.
  2. Leftover unused code. The fix removes the old buggy animation, but doesn't clean up the leftover plumbing behind it (a state variable, a type, a chunk of
  UI code) that can now never run. Doesn't break anything — just clutter that should've been deleted along with the bug.
  3. A small unexplained change. One line was quietly removed that used to "pre-load" the tab menu when a keyboard user tabs onto the button (mouse users still
  get this). Nobody explained why. Probably harmless, but worth a one-line answer from the author.
  4. Minor redundancy. A cleanup function runs twice instead of once when you pick a tab. Not a bug — just a little wasteful.

ApexYash11:
There isn't a five tab limit. The button is only disabled when there are no tabs (tabs.length === 0). The dim appearance comes from pre existing styling that changes the icon color once there are multiple tabs, and git blame shows this was already in the code before my PR. I think it's best handled as part of the browser tabs revamp unless the button is actually becoming unclickable 

and for the The onFocus preload was removed intentionally to prevent the original race from reappearing. When the Radix menu closes, it restores focus to the tabs button, and the old onFocus handler could trigger a new screenshot capture after cleanup, potentially restoring a stale tab image. Keyboard support is still preserved through handleTabsTriggerKeyDown for Enter, Space, and Arrow Down

and I'll clean up the rest code u sugessted

Vaibhaav-Tiwari (approved):
noice

PR #3602, opened 2026-08-05
fix(scm): guard all paginated check runs

### LINKED ISSUE #3601: bug(scm): completed checks can remain pending until forced refresh
## Bug

A pull request can continue to show **Checks running** in AO for roughly five minutes after every GitHub check has completed. The card corrects itself only when the SCM observer's unconditional-refresh backstop runs.

**Source:** Live AO review session on 2026-08-05 (no public chat URL)  
**Reported by:** Human reviewer in the AO desktop app  
**Analyzed against:** `c970f0d2436d770cf24bda317785ea5a8313d656`  
**AO version:** `dev` (freshly built Go rewrite; verified installed daemon ready on port 3001)  
**Observed surface:** macOS Electron development app using the same rewrite and real AO data on its development port 3002  
**Confidence:** High  
**Priority:** P2 — status is materially misleading, but self-corrects through the bounded-staleness fallback.

## Reproduction

1. Open an AO session whose attributed GitHub pull request has multiple check runs.
2. Push a commit so those checks enter the running state.
3. Keep the PR summary card visible in the session inspector.
4. Wait until every GitHub check completes successfully.
5. Observe that AO can continue to display **Checks running**.
6. Wait for the five-minute forced PR refresh; the card changes to **Checks passing** without user action.

Live instance: PR #3590 had all relevant checks complete by `2026-08-05T07:00:44Z`, while AO did not observe `ci.state: passing` until approximately `2026-08-05T07:06:05Z`. The renderer fetched and displayed the corrected daemon response immediately afterward.

## Root Cause

`backend/internal/adapters/scm/github/observer_provider.go::CommitChecksGuard` requests:

```text
GET /repos/{owner}/{repo}/commits/{sha}/check-runs?per_page=1
```

and uses that response's ETag as the fast check-state guard. With multiple check runs, the one returned item can remain unchanged while a different workflow finishes. Since the conditional representation has not changed, GitHub can return `304 Not Modified` even though the aggregate CI state AO displays has changed.

`backend/internal/observe/scm/observer.go::selectRefreshCandidates` then sees both the repository guard and this single-item check guard as unchanged and skips the full GraphQL PR/check refresh. `DefaultPRMaxAge = 5 * time.Minute` eventually forces the full fetch, which explains the observed delay precisely.

The frontend event transport is not the source of the lag: once SQLite changes from `pending` to `passing`, CDC invalidates the session SCM query and the card updates.

## Suggested Fix

Make the fast check guard represent all check runs that contribute to the aggregate CI state, rather than only one item. A fix should:

- request and condition on the complete relevant check-run representation (including pagination or a better aggregate endpoint where necessary);
- preserve conditional-request savings when none of those runs changes;
- add a provider/observer regression test where the first returned run remains stable while another run transitions from running to passing;
- retain `DefaultPRMaxAge` as a safety backstop for PR metadata changes not represented by check-state guards.

## Impact

- Users can spend up to five minutes believing checks are still running.
- Ready-to-merge status and merge actions are delayed even when GitHub is already green.
- The UI appears unreliable because a later refresh changes the state without any new GitHub activity.

## Related

- [#2656](https://github.com/AgentWrapper/agent-orchestrator/issues/2656): tracked PR state could remain stale after merge; established the need for bounded refresh.
- [PR #2678](https://github.com/AgentWrapper/agent-orchestrator/pull/2678): added the five-minute forced-refresh backstop that eventually heals this report but does not make CI transitions timely.

AgentWrapper:
## Summary

- guard the complete latest check-run representation with one conditional ETag per 100-run page
- encode page validators in an opaque provider token so unchanged pages retain 304 savings
- use total_count to track page growth and shrinkage while preserving the five-minute max-age fallback
- cover a later-page check transition from pending to passing while the first page remains stable

## Tests

- cd backend && go test -race ./internal/adapters/scm/github ./internal/observe/scm
- cd backend && go build ./...
- cd backend && go test ./...

Fixes #3601

### DIFF
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -134,9 +134,13 @@ func (p *Provider) CommitChecksGuard(ctx context.Context, repo ports.SCMRepo, he
 		previous.PageETags = make([]string, pageCount)
 		previous.PageETags[0] = resp.ETag
 	} else {
+		// total_count is present in every page representation, so page growth or
+		// shrinkage changes each stored page's ETag; all 304s imply a stable count.
 		pageCount = min(len(previous.PageETags), githubCheckRunsMaxPages)
 	}
 
+	// A missing page ETag is retained as an empty validator: that page is fetched
+	// unconditionally and keeps the guard changed, favoring freshness over 304 savings.
 	pageETags := make([]string, 0, pageCount)
 	changed := !composite
 	for page := 1; page <= pageCount; page++ {
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -7,6 +7,7 @@ package github
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -21,7 +22,11 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
 )
 
-const scmBatchCheckContextLimit = 20
+const (
+	scmBatchCheckContextLimit = 20
+	githubCheckRunsPageSize   = 100
+	checkRunsGuardPrefix      = "ao-github-check-runs-v1:"
+)
 
 const (
 	// githubReviewThreadPageSize fetches the latest review window cheaply for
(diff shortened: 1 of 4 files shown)


harshitsinghbhandari (approved):
LGTM

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -94,75 +99,80 @@ func (p *Provider) ListOpenPRsByRepo(ctx context.Context, repo ports.SCMRepo) ([
...
 		}
-		return ports.SCMGuardResult{ETag: fp}, nil
+		pageCount = checkRunsPageCount(page.TotalCount)
+		if pageCount == 1 {
+			return ports.SCMGuardResult{ETag: firstNonEmptyHeader(resp.ETag, etag)}, nil
+		}
+		previous.PageETags = make([]string, pageCount)
+		previous.PageETags[0] = resp.ETag
+	} else {
+		pageCount = min(len(previous.PageETags), githubCheckRunsMaxPages)
```
AgentWrapper, line 139: Non-blocking: the composite guard's ability to notice page growth/shrinkage while every *stored* page 304s rests entirely on the invariant that `total_count` appears in every page body, so any run addition or removal changes every page's ETag. That invariant is what makes it safe to derive `pageCount` from the stored token here without an unconditional probe. Worth a one-line comment — a future reader could otherwise conclude a new page 3 can appear while pages 1–2 both 304, and "fix" this with an extra request.

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -94,75 +99,80 @@ func (p *Provider) ListOpenPRsByRepo(ctx context.Context, repo ports.SCMRepo) ([
...
+		}
+		if !composite && page == 1 {
+			pageETags = append(pageETags, previousETag)
+			continue
+		}
+		resp, err := p.fetchCommitChecksGuardPage(ctx, path, page, previousETag)
+		if err != nil {
+			return ports.SCMGuardResult{}, err
+		}
+		pageETags = append(pageETags, firstNonEmptyHeader(resp.ETag, previousETag))
```
AgentWrapper, line 159: Non-blocking edge: if GitHub ever returned a 200 with no `ETag` header, the empty string gets stored as that page's validator, and every subsequent poll refetches the page unconditionally — `changed` becomes true each round, so the guard never reports NotModified again for that commit and the 304 savings silently vanish (same on line 135 for the bootstrap page-1 slot). GitHub reliably sends ETags on these endpoints today, so this is fine to ship as-is; just flagging that the failure mode is silent extra refreshes rather than staleness (the safe direction).

AgentWrapper (commented):
**Verdict: approve** — the composite per-page ETag guard is correct as far as I can trace it, and the fix genuinely closes #3601's remaining gap (a later-page transition invisible to the first page's ETag).

What I verified by tracing `CommitChecksGuard`:

- **Bootstrap (non-composite → multi-page)**: page 1 fetched conditionally, pages 2..N unconditionally; result is always `NotModified=false` with a composite token, so the observer refreshes. Correct.
- **Steady state**: all pages 304 → `NotModified=true`, and the re-encoded token is semantically identical to the stored one (the observer discards it on NotModified anyway).
- **Page growth/shrink**: handled because `total_count` is embedded in *every* page body, so adding/removing a run changes every page's ETag — no page can 304 while the page count silently changes. `pageCount` is recomputed mid-loop from any 200 body, new pages are fetched with an empty validator, and the trailing trim keeps the token consistent. Loop is bounded by `githubCheckRunsMaxPages`.
- **Migration**: a legacy raw ETag or old sha256 fingerprint fails `decodeCheckRunsGuard`, goes down the non-composite path as a plain If-None-Match, and can only produce a stale 304 if the representation is genuinely unchanged. Safe.
- **Failure paths**: decode failures fall back to the full-refresh path with an empty ETag, which the observer refuses to store (`res.ETag != ""` guard), so the old validator is retained. Same posture as before.
- Removed identifiers (`restCommitCheckRun`, `commitCheckRunsFingerprint`, `fetchRemainingCommitCheckRuns`) have no remaining references; `stableCheckFingerprint` is still used by the failed-check fingerprint. The cache is in-memory only, so there is no persisted-format migration concern.

The updated test covers the exact regression (page-2 transition while page 1 is stable) plus the unchanged-304 path.

Two non-blocking notes inline: documenting the `total_count` invariant the growth detection silently relies on, and a defensive edge around a 200 response without an ETag header. Neither blocks merge.

AgentWrapper (commented):
**Verdict: approve** — re-review of 781ec16. The only delta since my previous approved review (1dc503b) is a 4-line, comment-only change that addresses both of my non-blocking inline notes: it documents the `total_count`-in-every-page invariant that page growth/shrinkage detection relies on, and documents that a missing page ETag degrades to unconditional fetches (favoring freshness over 304 savings — the safe direction). No behavior change; the guard logic I traced and approved last round is untouched. Ready to merge.

PR #3603, opened 2026-08-05
fix: detect authenticated Kimi installs

neversettle17-101:
## Summary
- add Kimi self-managed install path (`~/.kimi-code/bin/kimi`) to binary discovery
- recognize Kimi OAuth credential files (`~/.kimi-code/credentials/kimi-code.json`) as authorized when they contain an access or refresh token
- prefer valid credential tokens over empty `config.toml` `api_key` placeholders
- add focused tests for Kimi binary discovery and credential auth classification

## Issue
Kimi was already present in the agents dropdown, but authenticated local installs were still shown as `Needs auth`. The frontend was rendering the daemon catalog correctly; the bad state came from the Kimi adapter's readiness probe.

The local Kimi install shape can be:
- binary at `~/.kimi-code/bin/kimi`
- `config.toml` containing empty `api_key = ""` placeholders
- real OAuth auth stored in `~/.kimi-code/credentials/kimi-code.json`

The adapter did not search the self-managed binary path, and its first OAuth credential support still stopped too early when it saw empty `api_key` values in `config.toml`, classifying the install as unauthorized before checking the credential JSON.
## Before 
Screenshot 2026-08-05 at 3 17 58 PM

## After
Screenshot 2026-08-05 at 3 11 39 PM

## Fix
The Kimi adapter now checks `~/.kimi-code/bin/kimi` during binary resolution. For auth, it treats a non-empty config API key as authorized, then checks `credentials/kimi-code.json` for an access or refresh token, and only falls back to the empty config-key unauthorized result when no credential token exists.

With that, the daemon can classify this local Kimi install as installed/authorized, and the existing agents dropdown can show it as available instead of `Needs auth`.

## Tests
- `go test ./internal/adapters/agent/kimi ./internal/service/agent`
- `go test ./internal/adapters/agent/registry ./internal/httpd/controllers -run 'Test.*Agent|TestEveryHarnessReportsAuthStatus|TestHarnessed'`\n- `git diff --check`\n\n## Risks\nLow. The change is scoped to Kimi catalog readiness probes and does not alter launch or restore command behavior.

### DIFF
--- a/backend/internal/adapters/agent/kimi/auth.go
+++ b/backend/internal/adapters/agent/kimi/auth.go
@@ -2,6 +2,7 @@ package kimi
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -46,7 +47,10 @@ func kimiLocalAuthStatus(ctx context.Context) (ports.AgentAuthStatus, bool, erro
 	if !ok {
 		return ports.AgentAuthStatusUnknown, false, nil
 	}
-	return kimiConfigAuthStatus(filepath.Join(home, "config.toml"))
+	if status, ok, err := kimiConfigAuthStatus(filepath.Join(home, "config.toml")); err != nil || ok {
+		return status, ok, err
+	}
+	return kimiCredentialsAuthStatus(filepath.Join(home, "credentials", "kimi-code.json"))
 }
 
 func kimiCodeHome() (string, bool) {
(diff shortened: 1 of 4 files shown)


neversettle17-101 (commented):
## What This PR Does

Adds Kimi self-managed binary discovery and expands local auth detection to recognize `~/.kimi-code/credentials/kimi-code.json` access/refresh tokens, while preserving non-empty config API keys as authorized.

## Summary of Changes

- Adds credential JSON parsing for Kimi OAuth tokens and changes local auth precedence so empty config `api_key` placeholders no longer block valid credentials.
- Adds `~/.kimi-code/bin/kimi` to Unix home binary lookup.
- Adds focused unit coverage for credential token detection, empty-token unauthorized classification, config/credential precedence, and binary path inclusion.

## Findings

No blocking findings.

## Description Match

The implementation appears aligned with the PR description: it checks non-empty config keys first, then credential tokens, and only falls back to the empty-config unauthorized result when no credential token is present.

## Regression Risk

Low. The changes are scoped to Kimi install/auth readiness detection. I did not see a clear regression in the reviewed paths.

## Testing

Tests were not run as part of this review. The PR adds relevant unit coverage for the main Kimi auth and binary-discovery cases, and the PR body lists targeted Go tests plus diff check as performed.

Ready to merge from my review.

nikhilachale (approved):
LGTM 
The code is clean, tests are thorough, and the approach is conservative. The PR successfully addresses the stated problem without introducing significant risk.

PR #3083, opened 2026-07-24
feat(frontend): flush Windows center panel and lighter terminal background

JAYATIAHUJA:
## What

Two small desktop UI polish items from a Windows visual pass:

1. **Windows center-panel top inset removed.** The shell root now gets a `platform-windows` class, and `.platform-windows .center-panel-shell` overrides `padding-top` to `0` — the framed panel sits flush under the custom titlebar instead of leaving a 12px dark gap.
2. **Lighter dark-theme terminal background.** `--color-bg-terminal` goes from `#15171b` to `#101317e7` (slight translucency). Light theme untouched.

## Why

On Windows the frameless window's custom titlebar already supplies the top chrome, so the shared 12px top inset read as dead empty space above the framed panel. The terminal background was visually indistinguishable from surrounding surfaces; the new shade gives it a distinct, slightly lighter tone.

## How

- Platform detection reuses the existing module-scope `isWindows` in `_shell.tsx`; the override is a single CSS rule in `styles.css`, so macOS/Linux and all component internals are untouched.
- The terminal color stays a single token edit in `tokens.css` — no hardcoded colors in components; xterm picks it up via `buildTerminalThemes()` on reload.

## Testing

- `npm run typecheck` (frontend) passes; `prettier --check` passes on all touched files.
- Verified live in the Electron dev app (`npm run dev`) on Windows: panel renders flush under the titlebar; terminal shows the new shade after reload.
- Also verified via headless Chromium probe against the renderer dev server: `platform-windows` class present and computed panel padding overridden.

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; links the related issue when applicable
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense
- [x] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

PR #3615, opened 2026-08-05
fix(frontend): strip macOS nested-radius from Windows/Linux center panel

i-trytoohard:
## Problem

On Windows/Linux, the right side of the app shows curved/rounded corner radii while the left side (flush with the sidebar) stays square — reported by @phylsolver in bug-triage.

## Root cause

`center-panel-surface` uses `border-radius: var(--radius-center-panel)` (= **12px**), derived from a **macOS-specific formula**:

```
inner_radius = window_radius (18px) − inset (6px) = 12px
```

This nested-radius curve only makes sense **inside a rounded macOS window**. On Windows/Linux the frameless window is **rectangular**, and the shell has **asymmetric padding**:

| Edge | Padding | Curve visible? |
|------|---------|----------------|
| Left | `0` (flush to sidebar) | hidden |
| Top  | `0` (overridden by #3083) | hidden |
| Right | `16px` | **visible** |
| Bottom | `14px` | **visible** |

So the rounded corners only show on the right/bottom, making the panel look like a floating rounded card while the left stays square.

PR #3083 (*"flush Windows center panel"*) already dropped the **top** inset on Windows but never stripped the **radius**. This finishes that job.

## Fix

Set `border-radius: 0` on `.platform-windows` / `.platform-linux` `center-panel-surface` so the framed panel has sharp corners, consistent with the flush left edge. macOS is untouched — it keeps the nested-radius curve inside its rounded window.

## Testing

- [x] CSS-only change, no logic/TS affected
- [x] macOS path (`center-panel-shell--mac`) untouched — no regression
- [ ] Manual: verify Windows/Linux center panel now has square corners on right/bottom

### DIFF
--- a/frontend/src/renderer/styles.css
+++ b/frontend/src/renderer/styles.css
@@ -530,19 +530,21 @@
 	);
 }
 
-/* Windows / Linux: a ShellTopbar already supplies the top chrome, so the
- * framed panel needs no additional top inset. Windows uses its custom
- * frameless titlebar; Linux uses the standard ShellTopbar. */
+/* Windows / Linux: rectangular frameless window. There is no rounded macOS
+ * window to inset into, so flush the framed panel on every side — top is
+ * already covered by the custom titlebar / ShellTopbar, and the left is flush
+ * to the sidebar globally (padding-left: 0). Zero the remaining right/bottom
+ * insets so the panel fills the full window width instead of floating with a
+ * gap on the right and bottom. */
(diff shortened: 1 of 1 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari (approved):

PR #3616, opened 2026-08-05
Match the feedback loop demo's chrome to the session view

codebanditssss:
Follow-up to #3605. The demo's topbar and rail were drawn at their own scale rather than the renderer's, so a few surfaces read wrong next to the real app.

image


Annieeeee11 (approved):
lgtm

PR #3618, opened 2026-08-05
fix(ptyexec): bound Close wait after the SIGKILL fallback

vanshx999:
## Summary

`ptyexec.Spawn` opens a local PTY around a terminal attach client (`tmux attach-session` on Unix, ConPTY on Windows). Its `Close()` tears the attach process down as **SIGTERM → SIGKILL → `<-done`**, where `done` closes when `cmd.Wait()` returns.

That final `<-done` is **unbounded**. A process wedged in an uninterruptible kernel state (`D` — e.g. blocked on a dead NFS/FUSE mount in the workspace) survives **SIGKILL**, so `cmd.Wait()` never returns and `Close()` blocks forever. Because `Close()` runs on daemon shutdown and on session teardown (including via the `Spawn` ctx-cancellation goroutine), a single wedged attach terminal hangs the entire daemon shutdown indefinitely. The Windows `conPTYProcess.Close()` has the identical unbounded wait.

This matters in practice because the attach client here is `tmux attach`, which resolves its session in the workspace tree — exactly the kind of path a hung filesystem can wedge.

## Change

- Add a shared `killGrace` (2s) and a `waitForExit(done, grace)` helper that returns on `done` **or** after the grace.
- Unix `creackPTY.Close()` and Windows `conPTYProcess.Close()` now call `waitForExit(...)` after the SIGKILL fallback, then release the PTY master and return. The `cmd.Wait()` goroutine finishes whenever the kernel eventually reaps the process; `Close` stays idempotent via `closeOnce`.
- Add unit tests (`wait_test.go`) covering both the bounded-timeout escape hatch and the fast path when the process has already exited.

SIGTERM-first behavior is preserved: a cooperative attach client still detaches within `detachGrace` (250ms), and a TERM-ignoring process is still killed via SIGKILL — only the post-Kill wait is now bounded.

## Tests

- `go build ./...`, `go vet`
- `go test -race ./internal/adapters/runtime/... ./internal/terminal/...` — pass
- `go test ./...` — pass except four pre-existing `internal/cli` spawn tests that fail on a clean `main` checkout too (environment/daemon-404, unrelated to this change).

## Known risks / follow-ups

- Worst-case shutdown now adds up to ~2.25s per genuinely-wedged attach terminal instead of hanging forever (intentional tradeoff).

### DIFF
--- a/backend/internal/adapters/runtime/ptyexec/spawn_unix.go
+++ b/backend/internal/adapters/runtime/ptyexec/spawn_unix.go
@@ -119,7 +119,11 @@ func (p *creackPTY) Close() error {
 			if p.cmd.Process != nil {
 				_ = p.cmd.Process.Kill()
 			}
-			<-done
+			// SIGKILL reaps a well-behaved process and done closes; a process
+			// wedged in uninterruptible kernel state survives it, so blocking
+			// on done here would hang daemon shutdown forever. waitForExit
+			// bounds the wait and Close releases the PTY master regardless.
+			waitForExit(done, killGrace)
 		}
 		p.closeErr = p.f.Close()
 	})
--- a/backend/internal/adapters/runtime/ptyexec/spawn_windows.go
+++ b/backend/internal/adapters/runtime/ptyexec/spawn_windows.go
@@ -85,7 +85,10 @@ func (p *conPTYProcess) Close() error {
 			if p.cmd.Process != nil {
 				_ = p.cmd.Process.Kill()
 			}
-			<-p.waitDone
+			// Mirror the Unix escape hatch: a process that survives Kill must
+			// not block Close (and daemon shutdown) forever. waitForExit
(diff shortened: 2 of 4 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Hey can you explain the problem we are solving here from the perspective of a user? What does a user see?

vanshx999:
@illegalcall  This fixes a rare edge-case hang: when a session's workspace is on a dead/hung mount ( NFS, unplugged drive, etc.), the tmux attach client wedges in an unkillable kernel state, and the old code would block session kill / daemon shutdown forever now it caps that wait at 2s so the user regains control instead of hitting an indefinite stall

harshitsinghbhandari (approved):
LGTM

PR #3619, opened 2026-08-05
fix(scm): condition commit check guard on all check runs, not just one

### LINKED ISSUE #3601: bug(scm): completed checks can remain pending until forced refresh

vanshx999:
## Summary

Fixes #3601 — **completed checks can remain pending until forced refresh**.

`CommitChecksGuard` in `backend/internal/adapters/scm/github/observer_provider.go` requested the per-commit check-runs endpoint with `per_page=1` and used that single-item response's ETag as the fast check-state guard. When a PR carries multiple check runs, the one returned item can remain unchanged while a **different** workflow finishes, so GitHub answers `304 Not Modified` even though the aggregate CI state AO displays has changed. `selectRefreshCandidates` then skips the full GraphQL PR/check refresh, leaving the card on **Checks running** until the five-minute `DefaultPRMaxAge` backstop.

## Change

- Request the **complete** check-run representation (`per_page=100`, GitHub's cap) so the ETag covers every run that contributes to the aggregate CI state.
- When a commit carries more than one page of runs (`total_count > 100`), fall back to a stable fingerprint over the **full** run set (paginating), so a transition on a later page still invalidates the guard. Conditional-request savings are preserved: when nothing changed, the guard reports `NotModified` and the observer skips the GraphQL refresh.
- `DefaultPRMaxAge` remains as the safety backstop for PR metadata changes not represented by check-state guards.

## Tests

- `TestCommitChecksGuard_OtherRunTransitionInvalidatesGuard` — regression: first returned run stays stable while another run transitions `in_progress -> completed`; guard must report `NotModified=false`. (Verified this test fails against the old `per_page=1` logic.)
- `TestCommitChecksGuard_PaginatedFingerprintInvalidatesLaterPageRun` — a transition on a page-2 run invalidates the paginated fingerprint guard.
- Ran `go test ./internal/adapters/scm/github/... ./internal/observe/scm/...`, `go vet`, `gofmt`, and golangci-lint v2.12.2 — all clean.

Pre-existing failures in `internal/cli`, `internal/adapters/agent/fake`, and `internal/adapters/agent/kilocode` reproduce on clean upstream `main` and are unrelated to this change.

## Risks / Follow-ups

None known. The guard change only affects when the fast path forces a full refresh; it can only make CI transitions arrive sooner, never later.

### DIFF
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -144,7 +144,7 @@ func (p *Provider) CommitChecksGuard(ctx context.Context, repo ports.SCMRepo, he
 		// Decode failure is unexpected but shouldn't block the whole repo poll.
 		// Return NotModified=false with no ETag so the observer falls back to the
 		// full GraphQL refresh path; DefaultPRMaxAge bounds staleness.
-		return ports.SCMGuardResult{}, nil
+		return ports.SCMGuardResult{}, nil //nolint:nilerr // decode failure falls back to the max-age refresh path
 	}
 	if page.TotalCount > len(page.CheckRuns) {
 		// The commit has more runs than one page fits; rely on a stable
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -91,18 +91,130 @@ func (p *Provider) ListOpenPRsByRepo(ctx context.Context, repo ports.SCMRepo) ([
 	}
 }
 
-// CommitChecksGuard checks GitHub's per-commit check-runs ETag guard.
+// githubCheckRunsPageSize fetches the complete check-run set for a commit in a
+// single REST page when the commit has no more than this many runs. A commit
+// virtually never exceeds GitHub's 100-result cap, so one request carries the
+// whole representation and GitHub's ETag reflects every contributing run.
+const githubCheckRunsPageSize = 100
+
+// restCheckRunsPage is the GitHub list-check-runs response envelope.
+type restCheckRunsPage struct {
+	TotalCount int                  `json:"total_count"`
+	CheckRuns  []restCommitCheckRun `json:"check_runs"`
+}
+
+// restCommitCheckRun is the subset of a GitHub check-run we condition the
+// aggregate CI-state guard on.
+type restCommitCheckRun struct {
+	Name       string `json:"name"`
+	Status     string `json:"status"`
+	Conclusion string `json:"conclusion"`
+}
+
+// CommitChecksGuard checks GitHub's per-commit check-runs guard. It conditions
+// the fast CI-state guard on the complete check-run set rather than a single
+// item: GitHub's ETag for a per_page=1 response only reflects that one run, so
+// a different workflow finishing (or failing, or going pending) can leave the
+// returned item unchanged and GitHub answers 304 even though the aggregate CI
+// state AO displays has changed. Requesting the full page makes the guard
+// represent all runs that contribute to the state, and paginating plus
+// fingerprinting stays correct when a commit carries more than one page of
+// runs. DefaultPRMaxAge remains the safety backstop for PR metadata changes not
+// represented by check-state guards.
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

harshitsinghbhandari (commented):
Right diagnosis and the regression test is genuine. Three things before merge:

- The pagination loop has no page cap. Bound it.
- A decode error now escalates to `markRepoFailed`, taking out the whole repo poll. Better to fall through to the backstop.
- `githubFailedFingerprint` in the same file already does sort + sha256 with the same separators. Reuse it rather than a second implementation.

Scope note, not a blocker: AO reads `statusCheckRollup`, which includes StatusContexts, and REST check-runs does not return those. Repos on legacy commit statuses still wait the full five minutes, so #3601 needs a follow-up to fully close.

harshitsinghbhandari (approved):
All three points addressed, CI green. LGTM.

vanshx999:
Thanks for the review @harshitsinghbhandari CI is green and this is approved could you merge it when you get a chance? Thanks!

PR #3620, opened 2026-08-05
fix: snap theme switches with View Transitions

Annieeeee11:
## Summary
- Wrap Color Theme, light/dark, and system theme updates in `document.startViewTransition` so CSS variable changes are covered by a screen snapshot
- Disable the default root crossfade so the swap is instant (no wipe/fade)
- Apply `data-theme` / `data-style-theme` inside the transition callback; keep shell effects as an idempotent safety net

## Test plan
- [ ] Switch Color Theme in Settings — UI snaps, no color fade on sidebar/board/cards
- [ ] Switch Theme (light / dark / system) — same instant cut
- [ ] Toggle theme from the command palette
- [ ] With Theme = System, change OS appearance — snap without fade
- [ ] `npx vitest run src/renderer/lib/theme.test.ts`
- [ ] `npm run frontend:typecheck`

### DIFF
--- a/frontend/src/renderer/lib/theme.test.ts
+++ b/frontend/src/renderer/lib/theme.test.ts
@@ -1,0 +1,66 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import {
+	applyDocumentTheme,
+	applyDocumentThemeStyle,
+	runThemeTransition,
+} from "./theme";
+
+describe("runThemeTransition", () => {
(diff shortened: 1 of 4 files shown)


ronishrohan (approved):
Reviewed the View Transitions integration, Zustand store changes, CSS crossfade disable, and test coverage. Clean and well-scoped — no app-breaking, scalability, or resource concerns.

PR #3625, opened 2026-08-05
feat(browser-annotations): attach a snapshot and tighten the prompt

Vaibhaav-Tiwari:
## Summary
- Captures a full-viewport screenshot of the browser preview the instant the user submits a browser-annotation change request, with the selection highlight baked in (the input box is hidden and the overlay teardown awaited before the main process captures, so the frame is clean).
- Attaches the snapshot to the outbound message: `/sessions/{id}/send` now accepts an optional inline image attachment, reusing spawn's MIME/size validation; the daemon writes it into the session worktree under `.ao/attachments/annotate-<uuid>.<ext>` and appends a file reference to the delivered message.
- Tightens `formatBrowserAnnotationMessage`'s template to a compact form and drops now-redundant fields (`Nearby text`, `ARIA role`, screen x/y) — the snapshot now carries the surrounding-context job that text used to do.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean on all touched Go files
- [x] `npx tsc --noEmit` clean repo-wide
- [x] New backend coverage: `TestWriteSendAttachment`, `TestSend_WritesAttachmentAndAppendsReference`, `TestSend_WithoutAttachmentSkipsWorkspaceWrite`, `TestSessionsAPI_SendWithAttachment`, `TestSessionsAPI_SendRejectsUnsupportedAttachmentType` — all pass
- [x] `go test ./internal/httpd/controllers/... ./internal/service/session/... ./internal/session_manager/... ./internal/cli/...` — all pass except 3 pre-existing Windows-environment failures unrelated to this change (verified via `git stash` against the unmodified base)
- [x] Frontend: 111/111 tests pass across `annotate-preload.test.ts`, `browser-view-host.test.ts`, `BrowserPanel.test.tsx`, `browser-annotations.test.ts`
- [ ] Manual verification in the desktop app (select an element, submit a change request, confirm the snapshot lands in the session workspace and the agent reads it)


i-trytoohard:
From the NotesBot call on `2026-08-05`, this existing pr appears to be relevant.
Match confidence: `0.59` (fuzzy).

Relevant discussion excerpts:
- - .phylolver(vaibhaav) said he worked on prateek’s suggestion to **tighten the prompt** and make it better. - .phylolver(vaibhaav) said the prompt has been **shortened**. - .phylolver(vaibhaav) added a flow where, when a user presses **Ctrl+Enter**, the system takes an **instance snapshot** and attaches it to the [...]

Source thread: https://discord.com/channels/1476302178913357958/1534617178966462485/1534617178966462485

backend/internal/session_manager/manager.go
```diff
@@ -2225,7 +2225,37 @@ func (m *Manager) applyWorkspaceProjectPreserved(ctx context.Context, rows []por
...
+		// silently dropping an image the caller explicitly attached would be a
+		// worse outcome than failing the send.
+		rec, haveRec, err := m.store.GetSession(ctx, id)
+		if err != nil {
+			return fmt.Errorf("send %s: attachment: session lookup: %w", id, err)
+		}
+		if !haveRec {
+			return fmt.Errorf("send %s: %w", id, ErrNotFound)
+		}
+		ref, err := writeSendAttachment(rec.Metadata.WorkspacePath, *attachment)
```
illegalcall: Please reject an empty `WorkspacePath` before this write. With an empty value, `filepath.Join` produces a relative `.ao/attachments` path, so AO writes beneath the daemon current working directory and gives the agent a reference that is not in its worktree. `StageAttachments` already has this guard; please apply the same check here and add a regression test.

backend/internal/httpd/controllers/sessions.go
```diff
@@ -1037,8 +1044,17 @@ func (c *SessionsController) send(w http.ResponseWriter, r *http.Request) {
 		envelope.WriteAPIError(w, r, http.StatusBadRequest, "bad_request", "MESSAGE_TOO_LONG", "Message is too long", nil)
 		return
 	}
+	var attachment *ports.SpawnAttachment
+	if in.Attachment != nil {
+		decoded, attachErr := decodeAttachment(*in.Attachment)
```
illegalcall, line 1093: Please bound `r.Body` before `decodeJSON`. This endpoint now accepts a base64 attachment, but the decoder loads the entire JSON string before the decoded 10 MiB check runs. For example, a 500 MiB payload can be allocated before rejection. The spawn, delegate, and stage attachment routes use `http.MaxBytesReader`; `/send` needs an equivalent cap and oversized-body test.

backend/internal/httpd/controllers/dto.go
```diff
@@ -175,11 +175,12 @@ type SpawnSessionRequest struct {
...
 	// into the session worktree and appends path references to the prompt.
-	Attachments []SpawnAttachmentInput `json:"attachments,omitempty"`
+	Attachments []AttachmentInput `json:"attachments,omitempty"`
 }
 
-// SpawnAttachmentInput is one file attached to a spawn request.
-type SpawnAttachmentInput struct {
+// AttachmentInput is one file attached to a spawn, delegate, stage, or send
+// request.
+type AttachmentInput struct {
```
illegalcall, line 183: Please add a `schemaNames` entry for the new named `AttachmentInput` type and regenerate the artifacts. Without it, the public schema is emitted as `ControllersAttachmentInput`, leaking the Go package prefix and creating avoidable generated-client contract churn. `AGENTS.md` explicitly requires a mapping for every new named API type.

frontend/src/main/browser-view-host.test.ts
```diff
@@ -1315,10 +1322,10 @@ describe("browser annotation IPC", () => {
 	});
 
 	it("forwards a single-element preview annotation submission to the renderer-owned view", async () => {
-		const { invoke, send, sent } = setupHost();
+		const { invoke, invokeFromTab, sent } = setupHost();
 		await invoke("browser:ensure", "sess-1");
 
-		send("browser:annotation:submit", 99, {
+		await invokeFromTab("browser:annotation:submit", 99, {
```
illegalcall, line 1114: This happy-path test never asserts that the forwarded payload contains `snapshot`, so the capture and forwarding code could be removed while the test still passes. Please assert the PNG snapshot here, cover resize plus timeout/text-only fallback, and add a renderer test proving `payload.snapshot` becomes `/send` body `attachment`.

illegalcall (changes requested):
Requesting changes for four hard issues: an unsafe relative attachment write, an uncapped attachment request body, a missing API schema-name mapping, and tests that do not verify snapshot transport. Details are inline.

backend/internal/session_manager/manager.go
```diff
@@ -2702,6 +2732,27 @@ func writeSpawnAttachments(workspacePath string, attachments []ports.SpawnAttach
 	return refs, nil
 }
 
+// writeSendAttachment writes a single attachment delivered via Send (e.g. a
+// browser-annotation snapshot) into the worktree under attachmentsDir and
+// returns its worktree-relative path. Unlike writeSpawnAttachments' sequential
+// image-N naming (safe for a one-shot spawn call), Send can fire many times
+// over a session's life, so the name carries a uuid to stay collision-safe.
+func writeSendAttachment(workspacePath string, attachment ports.SpawnAttachment) (string, error) {
```
illegalcall: The attachment paths now duplicate the same filesystem workflow across `writeSpawnAttachments`, `StageAttachments`, and `writeSendAttachment`: workspace validation, directory creation, extension fallback, file writing, relative-reference construction, and exclude setup. The missing empty-workspace guard in this new path is already an example of these copies drifting. Please consolidate this into a shared internal staging primitive that takes the resolved session/workspace and a naming strategy (sequential for spawn; randomized or `annotate-` prefixed for repeated sends). That primitive should own the empty-path guard, directory/write operations, reference construction, and `AddExclude`; callers should only choose naming and append the returned references. Add table tests for each naming mode and the empty-workspace case.

Vaibhaav-Tiwari:
Addressed all five review points:

1. **Empty-workspace guard** — `Manager.Send`'s attachment path no longer writes directly; it now delegates to the existing `StageAttachments` (which already had the guard, plus randomized naming and the exclude step). Added `TestSend_RejectsAttachmentWithEmptyWorkspace` as the regression test.
2. **Uncapped `/send` body** — added `http.MaxBytesReader` at the top of the handler, scoped to `maxSendBodyBytes` (sized for one attachment, not spawn's up-to-eight). Added `TestSessionsAPI_SendRejectsOversizedBody`.
3. **Missing schema mapping** — registered `AttachmentInput` in `schemaNames`; regenerated `openapi.yaml` and the TS client. `TestBuild_MatchesEmbedded` confirms no drift.
4. **Thin snapshot test coverage** — the single-element forward test now asserts the actual PNG payload; added coverage for the resize-over-1568px path, the capture-timeout→text-only fallback, and a renderer test proving `payload.snapshot` becomes the `/send` body's `attachment` field (plus the no-snapshot case).
5. **Duplicated write primitives** — deleted the bespoke `writeSendAttachment` entirely; `Manager.Send` now reuses `StageAttachments` directly rather than adding a third near-identical writer.

Also merged in latest `main` to resolve the conflict from `#3742`.

illegalcall (approved):

PR #3628, opened 2026-08-05
fix(goose): satisfy Goose 1.45+'s mandatory instructions flag on restore

### LINKED ISSUE #3581: bug(goose): GetRestoreCommand omits -t/-i/--recipe, restore silently no-ops on Goose 1.45+
## Summary

`goose.Plugin.GetRestoreCommand` builds a restore invocation that omits any of `-i`/`--instructions`, `-t`/`--text`, or `--recipe`, which `goose run` requires unconditionally as of Goose 1.45.0. Every Goose-harness session that survives a daemon/app restart is silently reduced to a zombie: AO's `pty-host` wrapper stays alive and the session is reported as `idle`/restored, but the actual `goose` child process exits immediately with a usage error and no agent is running.

## Where

`backend/internal/adapters/agent/goose/goose.go`, `GetRestoreCommand` (currently ~line 138-169 on `main`):

```go
cmd = append(gooseModeEnvPrefix(cfg.Permissions), binary, "run")
systemPrompt, err := restoreSystemPromptText(cfg)
...
if systemPrompt != "" {
    cmd = append(cmd, "--system", systemPrompt)
}
appendModelFlag(&cmd, cfg.Config)
cmd = append(cmd, "--resume", "--session-id", agentSessionID)
```

This produces:

```
goose run --system "<prompt>" --resume --session-id <agentSessionId>
```

Running that directly against Goose 1.45.0 fails immediately:

```
Must provide either --instructions (-i), --text (-t), or --recipe. Use -i - for stdin.
```

For contrast, `GetLaunchCommand` (the fresh-spawn path) already handles this correctly by appending `-t "" --interactive`, with a doc comment that states the requirement explicitly:

> Goose requires one of --instructions, --text, or --recipe, so AO supplies empty text plus --interactive to land in an input-ready terminal without inventing an initial task.

`GetRestoreCommand` was never given the same treatment.

## History / why this wasn't caught

- `GetRestoreCommand` has never included `-t`/`--interactive` since it was introduced.
- PR #2857 ("fix: restore Goose sessions by native session id") added `--system` to the restore command and explicitly chose *not* to add `-t`, to avoid replaying the original task on resume — a reasonable goal, but it didn't account for Goose's own flag requirement.
- PR #2956 (model-flag forwarding) is the most recent change to this function and didn't touch this either.
- This almost certainly worked on older Goose releases (we were on 1.23.2/1.27.2 before hitting a separate stream-inactivity hang bug fixed by upgrading to 1.45.0 — see the fix in `a345bd22`'s cycle). Something in Goose's own CLI validation tightened between those versions so that `--resume` no longer exempts the invocation from the `-i`/`-t`/`--recipe` requirement. So this is a latent AO bug that a Goose upgrade turned into an active, fleet-wide outage on every restart.

## Repro

1. Spawn any Goose worker/orchestrator session and let it produce at least one `SessionStart` hook (so `agentSessionId` metadata is populated).
2. Restart the AO daemon (or reboot the host).
3. Observe: the session is reported as restored/idle in `ao session get`/the dashboard, but no `goose` process is actually running. Running the exact argv AO used (visible via the OS process command line for the `pty-host` wrapper) reproduces the CLI usage error directly.

## Suggested fix

Append `-t ""` (or `--interactive`, matching whatever the intended non-replaying resume behavior should look like) to `GetRestoreCommand`'s argv alongside `--resume --session-id`, the same way `GetLaunchCommand` does — without reintroducing task replay. Add a regression test asserting the restore command always contains one of `-i`/`-t`/`--recipe` so a future Goose CLI tightening can't silently reintroduce this.

## Impact

Every Goose session (worker or orchestrator) that's alive at the time of an AO daemon restart or app relaunch is silently killed with no error surfaced anywhere in the AO UI/CLI — `ao session get`/`ao session restore` both report success. Discovered because a fleet of in-progress worker sessions all went dark simultaneously after a host reboot.

krishrathi1:
## Problem

`goose.Plugin.GetRestoreCommand` (`backend/internal/adapters/agent/goose/goose.go`) built:

```
[env GOOSE_MODE=<mode>] goose run [--system <text>] [--model <value>] --resume --session-id <agentSessionId>
```

As of Goose 1.45.0, `goose run` unconditionally requires one of `--instructions`/`-i`, `--text`/`-t`, or `--recipe`:

```
Must provide either --instructions (-i), --text (-t), or --recipe. Use -i - for stdin.
```

Without any of those, the restored `goose` process exits immediately on that usage error. AO's `pty-host` wrapper, however, stays alive and reports the session as restored/idle -- so every Goose-harness session that survives a daemon/app restart is silently reduced to a zombie with no actual agent process behind it.

`GetLaunchCommand` (the fresh-spawn path) already handles this correctly, with a doc comment stating the requirement explicitly: it appends `-t "" --interactive` so the empty text satisfies Goose's parser while `--interactive` keeps it in an input-ready terminal rather than exiting after the (empty) prompt "completes".

Fixes #3581

## Fix

Apply the same `-t "" --interactive` append to `GetRestoreCommand`, ahead of the existing `--resume --session-id <id>`:

```
[env GOOSE_MODE=<mode>] goose run [--system <text>] [--model <value>] -t "" --interactive --resume --session-id <agentSessionId>
```

This is the same, already-proven pattern used for fresh launches -- no change to the `--resume`/`--session-id` mechanism itself, and the existing doc comment explaining why AO uses `goose run` rather than `goose session` for restores (so `--system` can be re-applied) is unaffected.

## Test plan

- [x] Updated `TestGetRestoreCommandReadsAgentSessionID`'s exact-match assertion to include `-t "" --interactive`, and relaxed its old `!contains(cmd, "-t")` check (which was asserting the *absence* of `-t` under the old, buggy behavior) to just check no actual task text leaks into the restore command.
- [x] Added `TestGetRestoreCommandSatisfiesGoosesMandatoryInstructionsFlag`, a direct regression test for #3581: asserts the restore command contains `-t ""` and `--interactive`, and that `-t` is positioned before `--resume` (Goose parses left to right).
- [x] Verified by hand that both of these tests **fail against the pre-fix code** (missing `-t`, and the old exact-match `want` array no longer matching), and pass with the fix.
- [x] `go build ./...` passes
- [x] `go vet ./internal/adapters/agent/goose/...` passes
- [x] `go test ./internal/adapters/agent/goose/...` -- full package passes

Unlike my last few contributions to this repo, I do have a working Go toolchain in this environment this time, so all of the above was actually compiled and run, not just manually traced.


krishrathi1:
Heads up: the initial commit had a `lint` failure, but it wasn't from this change -- `golangci-lint` isn't scoped to the diff here, it lints the whole tree, and there was a pre-existing staticcheck SA5011 false positive in `backend/internal/httpd/auth_test.go` (from an earlier, unrelated PR) that was failing lint on every open PR right now, including this one.

Pushed a second commit (e94f437) fixing that too, since otherwise this PR would show a permanently red lint check through no fault of its own. Verified locally with `golangci-lint v2.12.2` (the version this repo's lint job pins) that both flagged SA5011 findings are gone, and `go build`/`go vet`/`go test` all still pass.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
Closing as stale: the goose restore flag fix is tracked in the still-open #3581; this branch has been silent a month and is unreviewed.

PR #3644, opened 2026-08-06
fix: don't gate PR-outcome notifications behind session restore

Pulkit7070:
## Summary
- `ready_to_merge`, `pr_merged`, and `pr_closed_unmerged` notifications describe work that already finished, so a terminated session behind one of them has nothing to resume. These types no longer show the restore action and their row navigates directly to the (terminated) session instead of being blocked.
- Only `needs_input` notifications keep the restore-first behavior, since there's an actual paused agent to resume there.

## Test plan
- [x] `cd frontend && npx vitest run src/renderer/components/NotificationCenter.test.tsx` (26/26 pass, includes a new test for the PR-outcome no-restore path)
- [x] `cd frontend && npm run test` (full suite: 1541/1546 pass; the 5 failures are in `generate-markdown-twins.test.mjs`, a pre-existing environment issue — missing `cheerio` in `node_modules` — reproduced identically on `main` before this change)

### DIFF
--- a/frontend/src/renderer/components/NotificationCenter.test.tsx
+++ b/frontend/src/renderer/components/NotificationCenter.test.tsx
@@ -70,7 +70,9 @@ const allNotifications: NotificationDTO[] = [
 		sessionId: "sess-dead",
 		projectId: "proj-1",
 		prUrl: "",
-		type: "pr_merged",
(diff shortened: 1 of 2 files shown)


Annieeeee11 (approved):
lgtm

PR #3648, opened 2026-08-06
fix(kimi): seed OAuth credentials for managed sessions

neversettle17-101:
## Summary

- seed authenticated Kimi OAuth credentials into AO’s isolated Kimi home on first session launch
- preserve credentials created by an existing AO-specific Kimi login
- write copied credentials with owner-only permissions

## Root cause

AO detects authentication from the user profile under `~/.kimi-code`, but launches Kimi with `KIMI_CODE_HOME` redirected to the AO-managed profile. Hook setup seeded API-key configuration but did not seed Kimi OAuth credentials, causing an authenticated install to prompt for login when a session started.

## Tests

- `go test ./internal/adapters/agent/kimi ./internal/session_manager`
- `npm run lint` (with AO session/runtime environment variables removed so environment-sensitive CLI tests run hermetically)

## Risk

Low and limited to first-time Kimi managed-profile setup. Existing AO-managed credentials are never overwritten.

### DIFF
--- a/backend/internal/adapters/agent/modelcatalog/catalog.go
+++ b/backend/internal/adapters/agent/modelcatalog/catalog.go
@@ -510,7 +510,26 @@ func parseJSONModels(output []byte) ([]ports.AgentModelInfo, error) {
 					IsDefault: firstBool(node, "isDefault", "is_default", "default"),
 				})
 			}
-			for _, child := range node {
+			for key, child := range node {
+				if key == "models" {
+					if modelMap, ok := child.(map[string]any); ok {
+						for alias, item := range modelMap {
+							if modelNode, ok := item.(map[string]any); ok && strings.TrimSpace(alias) != "" {
+								label := firstString(modelNode, "displayName", "display_name", "model_name", "label", "name")
+								if label == "" {
+									label = alias
+								}
+								models = append(models, ports.AgentModelInfo{
+									ID:        strings.TrimSpace(alias),
+									Label:     label,
+									Provider:  firstString(modelNode, "provider", "providerId", "provider_id"),
+									IsDefault: firstBool(modelNode, "isDefault", "is_default", "default"),
+								})
+							}
+						}
+						continue
+					}
+				}
 				walk(child)
 			}
 		}
--- a/backend/internal/adapters/agent/modelcatalog/catalog.go
+++ b/backend/internal/adapters/agent/modelcatalog/catalog.go
@@ -541,6 +541,9 @@ func parseJSONModels(output []byte) ([]ports.AgentModelInfo, error) {
 }
 
 func looksLikeModelAliasRecord(node map[string]any) bool {
+	if _, isModelContainer := node["models"]; isModelContainer {
+		return false
+	}
 	return firstString(node,
 		"modelId", "model_id", "model_uid", "slug", "model",
 		"provider", "providerId", "provider_id",
--- a/backend/internal/adapters/agent/modelcatalog/catalog.go
+++ b/backend/internal/adapters/agent/modelcatalog/catalog.go
@@ -514,7 +514,7 @@ func parseJSONModels(output []byte) ([]ports.AgentModelInfo, error) {
 				if key == "models" {
 					if modelMap, ok := child.(map[string]any); ok {
 						for alias, item := range modelMap {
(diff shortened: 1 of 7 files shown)


nikhilachale:
Screenshot 2026-08-10 at 16 58 17

Now kimi worker works with new task button and also model selection also working for  kimi

backend/internal/adapters/agent/modelcatalog/catalog.go
```diff
@@ -510,7 +510,26 @@ func parseJSONModels(output []byte) ([]ports.AgentModelInfo, error) {
 					IsDefault: firstBool(node, "isDefault", "is_default", "default"),
 				})
 			}
-			for _, child := range node {
+			for key, child := range node {
```
neversettle17-101, line 513: This branch treats every object-valued `models` property as a flat alias map and then `continue`s, even when none of the children were consumed as model entries. That regresses the generic walker for shapes like `{"models": {"available": [{"modelId": "..."}]}}` or other grouped/catalog objects: the `available` array is not walked anymore, whereas the previous code recursed into it. Please only skip recursion for entries actually recognized as alias records, or fall back to `walk(child)` when the `models` map is a grouping container.

neversettle17-101 (commented):
Needs changes: the Kimi auth/trust seeding and readiness-timeout updates look well covered, but I found a regression in the generic JSON model-catalog walker. The new special-case for map-shaped models can now skip nested model groups under a models object, so some previously discoverable model lists would disappear.

backend/internal/adapters/agent/modelcatalog/catalog.go
neversettle17-101: Addressed in ad47cef65. The parser now treats a child as a flat alias only when it has recognized model/provider identity fields; otherwise it recursively walks that child. This preserves Kimi map-shaped aliases while restoring discovery for grouping shapes such as `models.available[].modelId`. Added `TestParseJSONModelsWalksGroupedModelsMaps`. Verified with `go test ./internal/adapters/agent/modelcatalog ./internal/adapters/agent/kimi`.

backend/internal/adapters/agent/modelcatalog/catalog.go
```diff
@@ -510,7 +510,28 @@ func parseJSONModels(output []byte) ([]ports.AgentModelInfo, error) {
...
+				if key == "models" {
+					if modelMap, ok := child.(map[string]any); ok {
+						for alias, item := range modelMap {
+							if modelNode, ok := item.(map[string]any); ok && strings.TrimSpace(alias) != "" && looksLikeModelAliasRecord(modelNode) {
```
neversettle17-101, line 517: This still skips nested groups when the grouping object has provider metadata. For example, a common catalog shape like `{"models": {"anthropic": {"provider": "anthropic", "models": [{"modelId": "claude-sonnet"}]}}}` now satisfies `looksLikeModelAliasRecord` because `provider` is present, appends a bogus selectable model ID `anthropic`, and `continue`s without walking the nested `models` array. The alias-map special case should not classify containers that themselves have child `models`/model lists as leaf alias records; it should recurse into those entries instead.

neversettle17-101 (commented):
Needs changes: the previous grouped-map regression is partially addressed, but the alias-record detector is still broad enough to misclassify provider grouping objects and skip their nested models.

backend/internal/adapters/agent/modelcatalog/catalog.go
neversettle17-101: Addressed in 78600bb97. `looksLikeModelAliasRecord` now rejects any node containing a nested `models` property, so provider metadata cannot turn a grouping container into a selectable alias. The walker recurses into that group and discovers its nested model records. Added the exact provider-group regression case and verified with `go test ./internal/adapters/agent/modelcatalog`.

neversettle17-101 (commented):
Ready to merge: I do not have any blocking findings on this revision. The model-catalog recursion issue from the prior review is covered by the new provider-group test and the Kimi credential/trust seeding changes remain scoped to the managed Kimi home.

neversettle17-101:
Screenshot 2026-08-10 at 5 24 03 PM

nikhilachale (approved):

PR #3653, opened 2026-08-06
feat(new-task): preselect the default agent and model

illegalcall:
## Summary

The New task modal parked its Agent and Model controls on `Project default` / `Agent default` labels, which asks the user to remember what their default selection actually was. Both fields now **preselect the value the spawn will really use** and caption the label with where that value came from.

- **Agent** resolves to the project's worker agent (`config.worker.agent`), falling back to the daemon's default agent (`project.agent`). Caption reads `Project default` or `Global default`.
- **Model** resolves to the project's configured model/mode, falling back to the catalog entry flagged `isDefault`. Caption reads `Project default` or `Agent default`.
- The delegate call now names the visible agent selection instead of sending `agent: undefined`. This is a behavior fix, not just cosmetics: `Manager.Spawn` resolves an empty harness only through the project's worker override and errors with `ErrMissingHarness` if that is unset, so a project without a worker override previously could not start a task from this modal.
- Submitting a model is unchanged — only a model the user explicitly changed is sent, so an untouched default still resolves server-side.
- The agent trigger stops letting Radix clone the whole menu row into the closed control, which was dragging the selected checkmark and install status in with it. This also cleans up the create-project agent sheet, which shares `RequiredAgentField`.

New caption copy lands in all 8 locale catalogs; `FieldDefaultHint` is the shared caption atom.

## Tests

- `npx vitest run --config vite.renderer.config.ts src/renderer` — 81 files / 1119 tests pass.
- New `TaskComposer` cases: project worker agent is preselected and sent; global-default fallback when the project sets no worker agent; the agent's `isDefault` catalog model is preselected and captioned.
- `NewTaskDialog`: the case that asserted `agent: undefined` was updated to the new intent, and now also asserts the trigger reads `Claude Code` with a `Project default` caption.
- Verified against a live daemon with a real project config (worker agent `codex`): Agent shows **Codex · Project default**, Model shows **GPT-5.6 Sol · Agent default**.

Pre-existing, unrelated: 5 failures in `src/landing/scripts/generate-markdown-twins.test.mjs` (missing landing deps in the worktree) fail identically on a clean tree.

## Risks / follow-ups

- Task creation now always sends an explicit `agent`. If the resolved agent is unknown to the daemon, the modal surfaces `UNKNOWN_HARNESS` instead of silently falling back — the same error the user would get by picking it manually.
- The same "bare default label" pattern may exist on other surfaces. The reviewer picker in the session inspector already does the right thing (trigger shows the harness, menu item reads `Project default (codex)`); a wider sweep is not included here.


illegalcall:
Superseded by #3657, which now targets main and includes this PR's changes in the consolidated New task composer update.

PR #3655, opened 2026-08-06
feat(models): surface Claude Code's configured model, and invalidate the catalog cache on config change

illegalcall:
Stacked on #3653 — targets `ao/agent-orchestrator-1-46/new-task-default-agent`, review/merge that one first.

## Summary

#3653 makes the New task modal preselect the model an agent will actually use. Codex fills in, Claude Code does not: `modelcatalog.Base` flags none of `sonnet`/`opus`/`haiku` as default, so there is nothing to preselect and the field falls back to `Agent default`.

Two commits:

**1. Surface the configured model.** Claude Code exposes no non-interactive command that reports a resolved model — but the value it resolves is readable. AO passes no `--model`, so the choice comes from `ANTHROPIC_MODEL` or the `model` key in the settings files. Discovery now reads that value and flags it, nearest scope first:

1. the project's configured env (`ANTHROPIC_MODEL`)
2. process `ANTHROPIC_MODEL`
3. `<project>/.claude/settings.local.json`
4. `<project>/.claude/settings.json`
5. `~/.claude/settings.json`

A configured model outside the three published aliases (an explicit snapshot id, or a variant such as `opus[1m]`) is appended to the catalog, otherwise the picker would name a default it cannot round-trip. **When no scope sets a model, the catalog keeps no default** — the CLI decides, and asserting a guess would be a claim AO cannot verify.

**2. Make the cache honor it.** The cached catalog was keyed on the agent executable alone, so a settings edit could never invalidate it — the stale catalog stayed authoritative until the binary changed or the user pressed "Refresh models". The discoverer's `BinaryVersion(ctx, binary)` becomes `CatalogFingerprint(ctx, request)` over the same request `Discover` receives, so the cache decision — which by design runs *before* discovery — can see the configuration discovery would read. Agents that consult no configuration still return the bare executable fingerprint, so upgrading does not invalidate every cached catalog. This matches what `AgentModelCatalog.BinaryVersion` already documented itself as: "the legacy wire name for AO's non-sensitive executable **and configuration** metadata fingerprint."

## Tests

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test ./internal/adapters/... ./internal/service/... ./internal/httpd/...` passes.
- modelcatalog: configured alias flagged; out-of-alias value appended and flagged; local settings beat project settings and project env beats both; nothing configured leaves no default; malformed settings degrade to no default instead of failing discovery; fingerprint changes when the configured model changes and returns to its prior value for identical inputs; config-less agents keep the byte-identical executable fingerprint.
- service: the fingerprint request is deep-equal to the discovery request, carrying the project working dir and env — the invariant whose absence caused the staleness.
- **Verified live** against a dev daemon, without touching Refresh models:

  | step | fingerprint | default |
  |---|---|---|
  | baseline (`~/.claude/settings.json`) | `7a1bd55a1cf77e24` | `opus[1m]` |
  | wrote `<project>/.claude/settings.json` = haiku | `fa61532e84d9aa62` | `haiku` |
  | removed it again | `7a1bd55a1cf77e24` | `opus[1m]` |

Pre-existing, unrelated: `adapters/agent/kilocode`, `adapters/agent/opencode`, `adapters/agent/fake`, and `internal/cli` fail on this machine (CLI tests reach the live local daemon; two timing/probe tests). Same failures on a clean tree.

## Risks / follow-ups

- `CatalogFingerprint` replaces `BinaryVersion` on the `AgentModelDiscoverer` port. One production implementation and one test fake; no wire or storage change (the persisted column keeps its legacy name).
- Reading `~/.claude/settings.json` is a new (read-only, size-bounded) filesystem touch on claude-code discovery, now also on the cache-check path. Only the `model` key is parsed; nothing is logged.
- Enterprise-managed policy settings, which outrank user settings in Claude Code's own precedence, are not consulted.

### DIFF
--- a/backend/internal/adapters/agent/modelcatalog/catalog.go
+++ b/backend/internal/adapters/agent/modelcatalog/catalog.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -132,6 +133,9 @@ func (Discoverer) Manual(agentID string) ports.AgentModelCatalog { return Manual
 // Discover executes model catalog discovery for an agent binary.
 func Discover(ctx context.Context, agentID, binary, workingDir string, env map[string]string) (ports.AgentModelCatalog, error) {
 	base := Base(agentID)
+	if agentID == "claude-code" {
+		return withClaudeCodeDefault(base, workingDir, env), nil
+	}
 	spec, ok := commandSpecs[agentID]
 	if !ok {
 		return base, nil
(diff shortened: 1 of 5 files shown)


illegalcall:
Superseded by #3657, which now targets main and includes this PR's changes in the consolidated New task composer update.

PR #3661, opened 2026-08-06
feat(mobile): PostHog product analytics for the mobile app

Pulkit7070:
Adds product analytics to the mobile app, which had no telemetry of any kind. Mirrors the desktop pattern: pure, unit-tested logic core with a thin SDK edge, so the privacy and volume guarantees hold without a device.

## What it tracks
- `ao.app.active` (client=mobile), once per UTC day, event-driven (launch/foreground), no timer.
- `ao.mobile_app.paired` `{method, from_onboarding}` at both pair chokepoints (QR + manual), plus `onboarding_completed`.
- `ao.mobile_app.connected` `{trigger}` on reconnect.
- `ao.mobile_app.onboarding_started` / `_skipped`.
- `ao.mobile_app.notification_opened` `{target, cold_start}`.
- `ao.mobile_app.feature_used` `{feature, outcome}` around spawn / conductor / merge / kill / restore / send.

## Deliberately not included
- **No error tracking** — that goes to Sentry, by decision.
- **No screen capture** — `enableSessionReplay: false`, the imperative client (no touch/screen autocapture), and `captureAppLifecycleEvents: false`. The only lifecycle signal is the once-daily active heartbeat.
- **No screen-view event** — feature events answer the usage question without a per-navigation stream.

## Privacy
Every event passes an allowlist sanitizer that iterates the allowlist, not the payload, so session titles, PR names, project names, terminal output, and the connection password cannot leak. There is no free-text property kind. An unknown event name is dropped.

## Classification
`client: mobile / mobile-web` on every event, matching the desktop app (`client: desktop`) and CLI (`client: cli`), so shared events like `ao.app.active` split by one property. Identity is the SDK's persisted anonymous per-install id: device-based, never IP.

## Verification
309 unit tests pass, typecheck clean, both load-bearing guards (allowlist fail-closed, daily-active reservation) mutation-checked. The SDK wiring is verified on a device via an EAS build, not in unit tests, that last mile is not exercised here.


i-trytoohard:
From the NotesBot call on `2026-08-06`, this existing pr appears to be relevant.
Match confidence: `0.71` (fuzzy).

Relevant discussion excerpts:
- - Dhruv Sharma said the current agent session UI remains open for feedback. ## 📈 Telemetry, PostHog & Mobile Analytics - prateek said mockups are acceptable, but the core requirement is a fundamental presentation of the board states and transitions to the group.
- - Dhruv Sharma said some changes were lost after merging the agent session UI work, he had identified issues himself, and he is fixing them. - Dhruv Sharma said the current agent session UI remains open for feedback. ## 📈 Telemetry, PostHog & Mobile Analytics
- ## 📈 Telemetry, PostHog & Mobile Analytics - prateek said mockups are acceptable, but the core requirement is a fundamental presentation of the board states and transitions to the group.

Source thread: https://discord.com/channels/1476302178913357958/1534979721245364424/1534979721245364424

packages/mobile/lib/store.tsx
```diff
@@ -153,6 +155,7 @@ export function AppProvider({ children }: { children: ReactNode }) {
 			setError(null);
 			setErrorStatus(null);
 			setConnection("open");
+			mobileTelemetry()?.capture(MOBILE_EVENTS.connected, { trigger: "launch" });
```
Prasad-D-Ware: **Blocker.** Fires every poll tick, not on reconnect: `fetchAll` runs on `setInterval(tick, 8000)` (line 205). That's ~450 events/hour/user.

Gate on the `connection !== "open"` transition. `trigger` is also hardcoded `"launch"` — `"foreground"` and `"manual"` are never emitted.

packages/mobile/lib/telemetry/events.ts
```diff
@@ -0,0 +1,78 @@
...
+// which may ever reach PostHog. sanitize.ts iterates THIS allowlist, never the
+// caller's payload, so a property nobody registered here is dropped rather than
+// forwarded. Adding a field is a deliberate edit to this file, reviewable in one
+// place.
+
+export const MOBILE_EVENTS = {
+	// Unions with the desktop heartbeat of the same name; the `client` context
+	// property separates mobile from desktop. One "active users" metric, split by
+	// surface.
+	active: "ao.app.active",
```
Prasad-D-Ware: **Blocker.** `ao.app.active` is the legacy name (`docs/posthog-cost-controls.md:26-33`); ingestion rule 4 keeps it only for `channel = 'renderer'`.

Use `ao.v2.app.active`, and `ao.v2.mobile_app.*` for the rest.

packages/mobile/lib/telemetry/context.ts
```diff
@@ -0,0 +1,45 @@
...
+	// The bifurcation the ask named: native app vs the mobile-web build.
+	client: "mobile" | "mobile-web";
+	platform: "ios" | "android" | "web" | "other";
+	build_mode: "dev" | "device" | "simulator";
+	app_version: string;
+	telemetry_schema_version: number;
+};
+
+// Bump when the shape of any event payload changes in a way a query relies on.
+export const MOBILE_TELEMETRY_SCHEMA_VERSION = 1;
```
Prasad-D-Ware: **Blocker.** Should be `2` — that's the current-client value everywhere else (`frontend/src/renderer/lib/telemetry.ts:10`). A query filtering `= 2` will silently exclude mobile.

packages/mobile/lib/telemetry/runtime.ts
```diff
@@ -0,0 +1,84 @@
...
+ * component more than once cannot create a second client.
+ */
+export function initMobileTelemetry(): MobileTelemetry | null {
+	if (telemetry) return telemetry;
+
+	const client = new PostHog(MOBILE_POSTHOG_KEY, {
+		host: MOBILE_POSTHOG_HOST,
+		enableSessionReplay: false,
+		captureAppLifecycleEvents: false,
+	});
```
Prasad-D-Ware, line 50: **Blocker.** Three gaps vs desktop:

- **No dev gate** — `npm start` sends to the production project. Desktop returns `null` and never constructs the client unless packaged (`frontend/src/shared/telemetry.ts:39-47`).
- **No rate limiting** — desktop caps 5/min + 200/day per event name (`frontend/src/renderer/lib/telemetry.ts:38-39`).
- **No kill switch** (desktop has `AO_TELEMETRY_DISABLED_EVENTS`) and no opt-out. A shipped binary can't be hotfixed.

packages/mobile/lib/telemetry/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+): MobileTelemetry {
+	// Context rides as super-properties, so every event is tagged with
+	// client/platform/version without the call sites repeating it.
+	void client.register(context);
+
+	const capture = (event: MobileEventName, properties?: Record<string, unknown>): void => {
+		// Fail closed on the event name too: an event not in the allowlist is
+		// never sent, so a typo cannot ship a bare untracked event.
+		if (!(event in MOBILE_ALLOWLIST)) return;
+		client.capture(event, sanitizeMobileProperties(event, properties));
```
Prasad-D-Ware: **Blocker.** Missing `$process_person_profile: false`. It's set on every desktop (`frontend/src/renderer/lib/telemetry.ts:156`) and daemon (`backend/internal/adapters/telemetry/posthog.go:332`) event.

packages/mobile/lib/telemetry/sanitize.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+	// count: a finite, non-negative, safe integer. NaN/Infinity/floats/negatives
+	// are dropped rather than coerced.
+	if (typeof value === "number" && Number.isInteger(value) && value >= 0 && value <= COUNT_MAX) {
+		return { ok: true, value };
+	}
+	return { ok: false };
+}
+
+export function sanitizeMobileProperties(
```
Prasad-D-Ware, line 30: The allowlist only covers the custom bag. The SDK attaches `$device_model`, `$locale`, `$timezone`, etc. (`posthog-react-native/dist/types.d.ts:90-118`) and PostHog adds GeoIP server-side.

Pass the sanitizer as `before_send` so it governs all events.

packages/mobile/lib/PushManager.tsx
```diff
@@ -71,22 +73,25 @@ export function PushManager(): null {
...
+		const sub = Notifications.addNotificationResponseReceivedListener((r) => handle(r, false));
 		return () => sub.remove();
 		// route() reads the latest config via ref-free closure; re-bind when it changes.
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [navState?.key, config]);
 
-	function route(data: PushData) {
+	function route(data: PushData, coldStart = false) {
+		const type = data.type ?? "";
+		const target = type === "needs_input" ? "session" : type.includes("pr") ? "prs" : "notifications";
```
Prasad-D-Ware: `target` disagrees with actual routing: `ready_to_merge` navigates to `/prs` but reports `"notifications"`; `needs_input` without a `sessionId` navigates to `/prs` but reports `"session"`.

Derive it from `notificationTarget()` — the comment below says the rule shouldn't be written twice.

packages/mobile/lib/telemetry/events.ts
```diff
@@ -0,0 +1,78 @@
...
+	[MOBILE_EVENTS.onboardingCompleted]: {},
+	[MOBILE_EVENTS.onboardingSkipped]: {},
+	[MOBILE_EVENTS.notificationOpened]: {
+		// notificationView already maps a push to a small set of destinations.
+		target: { oneOf: ["session", "prs", "notifications", "home"] },
+		cold_start: { flag: true },
+	},
+	[MOBILE_EVENTS.featureUsed]: {
+		feature: {
+			oneOf: ["spawn", "merge", "kill", "restore", "conductor", "send", "voice_input"],
```
Prasad-D-Ware: `"home"` is unreachable and `"voice_input"` is never emitted. Drop both.

packages/mobile/package.json
```diff
@@ -32,6 +32,7 @@
 		"expo-secure-store": "~15.0.8",
 		"expo-speech-recognition": "~3.1.3",
 		"expo-status-bar": "~3.0.9",
+		"posthog-react-native": "^4.62.0",
```
Prasad-D-Ware, line 37: `expo-application` / `expo-file-system` resolve only transitively. If dropped, persistence falls back to memory and the anonymous install id churns on every launch. Declare them explicitly.

packages/mobile/lib/telemetry/config.ts
```diff
@@ -0,0 +1,13 @@
+// PostHog project the mobile app reports to.
+//
+// This is the APP project, the same one the desktop app uses, so desktop and
+// mobile usage sit together and split by the `client` context property. The
+// marketing site is a separate project. The key is a public project key (it
+// ships in every client build by design, like the desktop's), not a secret, so
+// hardcoding it mirrors the desktop's shared constant. An EXPO_PUBLIC_ override
+// lets a build point elsewhere without a code change.
+export const MOBILE_POSTHOG_KEY =
```
Prasad-D-Ware, line 9: `docs/telemetry.md` isn't updated. New client-side collection needs to land in the collection contract.

Prasad-D-Ware (dismissed):
Requesting changes. Verified: 350 tests pass, typecheck clean, session replay / lifecycle / error autocapture all correctly off.

Issues below are wiring divergences from `docs/telemetry.md` and `docs/posthog-cost-controls.md`. Blockers are the first five.

Also: the description claims `client: desktop` / `client: cli` exist. They don't — desktop uses `platform`/`build_mode`, and the CLI split is `channel`.

illegalcall (approved):

packages/mobile/lib/store.tsx
Pulkit7070: Fixed in 34b1a1fdc. Emission is now gated on the not-open to open transition via `openRef`, so it fires once per connect instead of every 8s poll tick. `trigger` is `launch` on first connect and `reconnect` thereafter.

packages/mobile/lib/telemetry/events.ts
Pulkit7070: Fixed in 34b1a1fdc. Renamed to `ao.v2.app.active` and `ao.v2.mobile_app.*` throughout, so ingestion rule 4 no longer drops these.

packages/mobile/lib/telemetry/context.ts
Pulkit7070: Fixed in 34b1a1fdc. `MOBILE_TELEMETRY_SCHEMA_VERSION` is now `2`, matching the other clients.

packages/mobile/lib/telemetry/runtime.ts
Pulkit7070: Fixed in 34b1a1fdc. Dev gate returns `null` under `__DEV__`/`MOBILE_TELEMETRY_DISABLED`; kill switch via `MOBILE_DISABLED_EVENTS`. The per-name rate cap is the stacked follow-up in #3704.

packages/mobile/lib/telemetry/telemetry.ts
Pulkit7070: Fixed in 34b1a1fdc. `$process_person_profile: false` is now set on every capture.

packages/mobile/lib/telemetry/sanitize.ts
Pulkit7070: Valid point on SDK auto-props. The allowlist governs the custom bag per-capture today; wiring it as a `before_send` hook to also strip `$device_model`/`$locale`/`$timezone` is tracked as a follow-up PR, keeping this one scoped.

packages/mobile/lib/PushManager.tsx
Pulkit7070: Fixed in 34b1a1fdc. `target` is now derived from `notificationTarget()` so it matches where the tap lands.

packages/mobile/lib/telemetry/events.ts
Pulkit7070: Fixed in 34b1a1fdc. Both dropped: targets are `session`/`prs`, features no longer include `voice_input`.

packages/mobile/package.json
Pulkit7070: Fixed in 34b1a1fdc. `expo-application` and `expo-file-system` are now declared explicitly.

packages/mobile/lib/telemetry/config.ts
Pulkit7070: Fixed in 34b1a1fdc. `docs/telemetry.md` now documents the mobile collection contract.

PR #3612, opened 2026-08-05
feat: add review feedback auto-inject toggle

VenkataSakethDakuri:
## Summary

- add a durable, default-on per-session policy for automatically injecting review feedback
- expose the policy through the session API and Session Inspector switch with optimistic updates
- gate both AO reviewer-batch feedback and GitHub/SCM review comments while leaving CI failure nudges unchanged
- emit session CDC invalidation for policy changes and repair the column on burned migration histories

## Testing

- `cd backend && go test ./internal/httpd/... ./internal/service/session ./internal/lifecycle ./internal/service/review`
- `cd backend && go test ./internal/storage/sqlite/...`
- `cd backend && go test ./internal/cli -run TestE2E_SpawnAndProjectAddDTORoundTrip`
- `cd backend && go test ./internal/session_manager -run TestSeedRecordEnablesAutoInjectReviewFeedback`
- `cd backend && go test ./... -run ^$`
- `npm run frontend:typecheck`
- `npm --prefix frontend test -- SessionInspector.test.tsx` (63 tests)
- `npm run sqlc`
- `npm run api:spec`
- `npx openapi-typescript@7.4.4 backend/internal/httpd/apispec/openapi.yaml -o frontend/src/api/schema.ts`

PR #3664, opened 2026-08-06
Merge the inspector's two review sections into one list

codebanditssss:
The session inspector showed AO's own reviewer passes and the reviews left on the PR as two separate sections. A single PR therefore appeared twice, once under each heading, and the reader had to join them up by number to see the whole picture of who had said what about a commit.

They are now one list keyed by PR. Each PR is one expandable row; inside, the groups name their source: **AO codex** against whichever agent actually ran that pass, **On GitHub** for humans and bots. `GithubReviewPanel` is gone.

## Run review

The "This commit has already been reviewed. Push a new commit to run another review." notice was a boxed two-line paragraph. It is feedback about a click, but it outlived the click and kept the rail's width for as long as the panel stayed open.

It is now a one-line `Already reviewed` marker. The sentence is the control's accessible name and shows on hover or focus, so nothing is lost for a screen reader or for anyone who wants the detail, and the resting state costs one line instead of a box.

## Removals

- "No unresolved threads." The unresolved count already rides the PR's own meta line, so the separate sentence only ever appeared when there was nothing to report.
- `inspector.reviewsOnPR` and `inspector.noOneReviewedYet` died with the section that used them.

## Two changes nobody asked for

- Four derivations moved to module level (`openReviewStatesFor`, `runsByPRFrom`, `triggeredReviewStatesFrom`, `aoReviewHarness`). The merged list needs exactly the data `ReviewPanel` was computing privately, and a second copy would have been free to drift from the first.
- `ReviewPanel` loses its `runs` prop, which became dead once the AO list moved out of it.

## i18n

Four keys in, three out, across all eight locales, which `i18n/instance.test.ts` enforces at the type level. `inspector.unresolvedCount` replaces a hardcoded English `` `${unresolved} unresolved` `` in the component, so that string is translatable for the first time.

The seven non-English translations are not from a translator. `inspector.reviewBySource.ao` is deliberately left as `AO {{harness}}` everywhere, since it is a product name plus an agent id.

## Conflicts with #3612

Both changes edit the same few lines of `ReviewsSection`: the auto-inject toggle is inserted immediately after the GitHub section this removes. Whichever lands second resolves by hand.

Separately, #3612 carries migration `0048` while main is at `0079`, so it needs a fresh number regardless of order. If this lands first, the natural resolution is to move the toggle into Run review next to the agent picker, which is arguably where a delivery policy belongs anyway and would give it the section heading it currently lacks.

## Verification

- `tsc --noEmit` clean
- 87/87 in `SessionInspector.test.tsx` and `i18n/instance.test.ts`, after updating two assertions that named the removed section titles
- exercised in the desktop app against seeded review data


neversettle17-101:
Can you add the before and after view of how reviews will look like?

codebanditssss:
> Can you add the before and after view of how reviews will look like?

image

neversettle17-101 (approved):

PR #3665, opened 2026-08-06
feat: add delegated task attachments

nikhilachale:
## Summary
- add POST /api/v1/orchestrators/delegate with DelegateTaskRequest attachments
- reuse spawn attachment decoding/validation and pass decoded attachments through ports.SpawnConfig
- regenerate OpenAPI and frontend API schema, and keep API telemetry normalization aligned for the delegate route

## Tests
- cd backend && go test ./internal/httpd/controllers ./internal/service/session
- npm run frontend:typecheck
- npm run api
- npm --prefix frontend test -- --run src/renderer/lib/api-client.test.ts src/renderer/components/NewTaskDialog.test.tsx
- /private/tmp/ao-154-frontend-node_modules/.bin/prettier --check frontend/src/renderer/lib/api-client.test.ts
- git diff --check

## Notes
- This checkout does not contain a TaskComposer component or an existing delegated composer UI surface; the existing NewTaskDialog attachment pipeline already covers local image selection/removal/error state.
- Project/workspace reference-file picking was intentionally not implemented.


frontend/src/renderer/components/TaskComposer.tsx
```diff
@@ -1,483 +1,688 @@
...
-			setError(t("newTask.taskRequired"));
-			return;
-		}
+  const handlePaste = (event: ClipboardEvent<HTMLTextAreaElement>) => {
+    const files = Array.from(event.clipboardData?.files ?? []).filter((file) =>
+      file.type.startsWith("image/"),
+    );
+    if (files.length === 0) return;
+    event.preventDefault();
+    void addFiles(files);
```
nikhilachale: [P1] Synchronize attachment reads before submission

`addFiles` does asynchronous `FileReader` work, but this handler discards the promise while `submitTask` is still enabled and immediately snapshots `attachments`. If the user pastes/drops/selects an image and presses Enter or Start before the read completes (especially for a larger image), the delegate request is sent without that image and the worker starts anyway. Track pending reads and await or block submission until they settle; please cover the immediate-submit case too.

nikhilachale (commented):
Changes requested.

The new attachment flow has a user-visible race: task creation can snapshot attachment state before FileReader completes and silently spawn a worker without the image the user just selected, pasted, or dropped. Please synchronize pending reads with submission and add immediate-submit regression coverage.

Please also revert the broad formatting-only churn in TaskComposer and api-client files so the patch stays surgical and follows the existing repository formatting.

Verification: the backend controller and session-service suites pass. The narrow frontend tests could not run in this checkout because vitest is not installed.

nikhilachale:
Addressed review 4877189687 in c2f4156c5: TaskComposer now awaits pending attachment reads via useFileAttachments.toSettledPayload() before creating the delegated task, the immediate-submit path is covered by a regression test, and the previous broad TaskComposer/api-client formatting churn has been removed from the upstream diff.

frontend/src/renderer/components/TaskComposer.tsx
nikhilachale: Fixed in c2f4156c5. Submission now awaits useFileAttachments.toSettledPayload(), so pending FileReader work settles before the delegate request snapshots attachments. Added an immediate-submit regression test that verifies no API request is sent until the file read completes.

nikhilachale:
Screenshot 2026-08-07 at 16 35 32

illegalcall:
Superseded by #3691, which preserves every attachment commit and adds the continuous New Task composer on top. Continuing review and CI on the combined PR.

PR #3674, opened 2026-08-06
feat(mobile): device roster — see, mute, and remove paired phones

Prasad-D-Ware:
## Problem

There was no way to see which phones can receive push notifications from a desktop, and no way to stop one without hand-editing `~/.ao/data/mobile/push-devices.json`. Hand-editing did not even work: the daemon loads the registry into memory once at startup and rewrites the file from memory on the next mutation, so manual edits were silently reverted.

The registry also accumulated duplicates. Rows were keyed by the Expo push token, which rotates without the app being reinstalled, so one phone became several rows — all of which kept receiving notifications until Expo reported the stale token dead.

To be precise about the scope of the fix: keying on a stable install ID removes duplicates caused by **token rotation**. It does not survive a full uninstall/reinstall — the install ID lives in AsyncStorage, which uninstall wipes — so a reinstalled app does still create a new row. What it can no longer do is leave **two rows holding the same token**, which was the harmful part: that meant every notification delivered twice.

## What this adds

A **Devices** section in the Connect Mobile modal that lists every paired phone with:

- a **live** dot when that phone's app is open and in the foreground, or "last seen X ago" otherwise
- a **mute** switch — the device stays paired but receives no notifications
- **remove**, behind a confirmation

## How it works

**Device identity is decoupled from the push token.** The registry now keys on a stable per-install ID, and the token is an optional attribute. A row means *"a paired phone"*, not *"a push registration"* — so a phone that has not granted notification permission (or whose build cannot mint a token, e.g. an iOS build without the APNs entitlement) still appears, as connected with notifications off. The phone announces its identity on connect with no permission gate; the token attaches to that same row later if permission is granted.

**Liveness comes from the phone's existing 8-second REST poll**, not from a new transport. Every request carries an `X-AO-Install-Id` header, a middleware records the timestamp, and "live" means seen within 20s (two poll intervals plus slack). The mobile app pauses polling on `AppState` background — but deliberately *not* on `inactive`, which iOS reports for Control Centre and the app switcher — so backgrounding goes dark deterministically on both platforms rather than waiting for the OS to suspend the JS timer.

The mux WebSocket was deliberately **not** used: it is only open while the user is on a terminal screen, so it would mean "viewing a terminal", not "app is open".

**Removal writes a tombstone.** A removed phone that re-registers comes back *muted* rather than silently resuming notifications, and the tombstone is consumed on match so a deliberate unmute is never overridden. There is no per-device credential to revoke — every phone shares one bridge password — so fully evicting a device remains "regenerate the password and re-pair the phones you want". The UI says so.

## Security

The management routes (`GET`/`PATCH`/`DELETE /api/v1/mobile/devices`) are **loopback-only**, mounted under `/api/v1/mobile`, which `lanControlBlockedPrefixes` already blocks on the LAN listener. That gate is enforced by the physical socket the request arrived on, so no header can spoof it, and it required no new gating code. A paired phone cannot enumerate the household's devices, unmute itself, or remove another device.

The phone's own unregister (`DELETE /api/v1/push/devices/{token}`) is unchanged and stays on the shared LAN surface. No new network-facing bind; the loopback listener is untouched.

Registration remains backward compatible: a request without an `installId` gets a server-synthesized `legacy-<uuid>`, which the token-adoption path merges into the real row on the next registration. Phones on older builds keep working — there is no flag day.

## Files worth reviewing

`backend/internal/mobilebridge/pushdevices.go` carries every subtle decision: the three-way `Upsert` resolution (install ID → token adoption → insert), `carryOver` (which preserves mute, created-at and token), tombstone consumption, and the `LoadRegistry` migration.

## Testing

Full suites green: backend (including `-race` on the touched packages), frontend 1976 tests, mobile 337 tests, no OpenAPI/TS drift.

Verified on real hardware (Android + iOS):

- a phone appears as connected with notifications off, before any permission is granted
- granting permission attaches the token to the same row rather than creating a second
- the live dot follows foreground/background, and survives the iOS `inactive` blip
- a notification was received on the device, then **muted and confirmed not delivered** with the device still holding a valid push token
- removing a device, then reopening the app, brings it back muted — and that survives a daemon restart

Not covered: iOS push *delivery*, which needs an EAS build for the APNs entitlement.

## Follow-ups (not blocking)

- `lastSeenAt` is only written on register/announce, so a long foreground session then a background shows a stale "N hours ago". The presence tracker holds the true timestamp; the roster does not read it yet.
- Dead-token pruning deletes the whole row, which contradicts the "a row is a paired phone" model. Clearing the token instead would be model-consistent; today a token death within an hour of a mute can drop that mute.


i-trytoohard:
From the NotesBot call on `2026-08-06`, this existing pr appears to be relevant.
Match confidence: `0.64` (fuzzy).

Relevant discussion excerpts:
- [15:52] -Prasad-: It lets you look at how many devices have been connected to your desktop, and also you can disconnect them, mute them from there. Itself. So let's say, uh, so like the scenario when it happened was, uh, I was trying to connect, uh, various phones to my app and, uh, it would make that, um, it [...]
- - -Prasad- explained the value: users will be able to see how many devices are connected to the desktop app and disconnect or mute them directly there. - -Prasad- said the feature came from a real scenario where multiple phones were connected and notifications were sent to all devices, but he did not have the [...]
- - -Prasad- said he is building a feature to show “all the connected devices” on the desktop app and plans to raise a PR “by tonight.” - -Prasad- explained the value: users will be able to see how many devices are connected to the desktop app and disconnect or mute them directly there. - -Prasad- said the feature [...]

Source thread: https://discord.com/channels/1476302178913357958/1534979721245364424/1534979721245364424

illegalcall (commented):
Thanks for putting this together, @Prasad-D-Ware. The overall direction makes sense, but I found several correctness and contract issues that I think need investigation before merge:

1. **Legacy adoption can leave two active rows for one phone.** On upgrade, `PushManager` starts the tokenless announce and token registration independently. The likely ordering is:

   ```text
   legacy-id -> token T
   announce real-id without token -> creates real-id row
   register real-id with token T -> updates real-id and returns
   ```

   Because `DeviceRegistry.Upsert` returns from the install-ID branch before checking whether another row already owns token T, the legacy row remains. The dispatcher can then send twice, and token-based unregister may remove only one row. I reproduced this with a temporary regression test using that three-step sequence; it returned two rows with the same Expo token. Could you add this ordering to the registry tests and reconcile both identities when either install ID or token matches?

2. **A full uninstall/reinstall can still recreate the original duplicate-row problem.** The install ID lives in AsyncStorage, so a full uninstall resets it. If the Expo token also rotates, the new install has no link to the old row, which remains until dead-token pruning. Please clarify the supported guarantee and investigate whether the PR description/UI currently overstates what the identity change solves.

3. **Unregister still deletes the paired-phone row.** The new model says a row represents a paired phone and that its token is optional, but `DeleteByToken` deletes the entire row. When the user disables notifications in the mobile app, the still-paired phone disappears from the roster until it announces again. Dead-token pruning has the same model mismatch. Please investigate clearing the token instead of deleting the row where the phone remains paired.

4. **The displayed last-seen time is not the actual last request time.** Presence tracks every REST request, but the roster returns `PushDevice.LastSeenAt`, which is updated only on register/announce. A phone can remain foregrounded for hours, background now, and immediately display as last seen hours ago. I see this is called out as a follow-up in the PR body, but it appears to leave one of the primary roster states incorrect.

5. **An in-flight poll can continue after backgrounding.** The AppState effect clears the interval but does not cancel the active `fetchAll`. That operation can start its later orchestrator/project/notification requests after the app has entered the background, extending liveness beyond the intended deterministic 20-second window on a slow connection. Please investigate cancellation or background checks between requests.

6. **Mute persistence errors are reported as 404.** `SetMuted` returns both unknown-device errors and file persistence errors, while the controller maps every error to `DEVICE_NOT_FOUND`. A disk/rename failure should be a 500-class error and have test coverage. Also note that the in-memory mute value is changed before persistence, so a failed request can still affect dispatch behavior until restart.

7. **The OpenAPI contract contradicts backward compatibility.** The runtime synthesizes a legacy ID when `installId` is omitted, but `RegisterPushDeviceRequest` marks `installId` required in OpenAPI and therefore in generated clients. Please make the published contract match the accepted runtime request shape.

8. **Presence cleanup only happens when the roster is read.** `Tracker.Live()` prunes expired entries, but it is only called by the roster endpoint. If the desktop never opens the Devices section, distinct install IDs remain in memory indefinitely. Please either bound/prune on writes or otherwise ensure the stated no-unbounded-growth property holds.

There are also a few maintainability concerns worth considering while addressing the above: the roster controller talks directly to the file-backed registry instead of a service/port boundary; the frontend casts the generated response to a hand-written `MobileDevice`; `Presence` and `DeviceLive` inject the same tracker twice and require normalization; and the duplicated announce/register AppState effects are what make the migration ordering especially fragile.

The existing targeted backend and race suites pass, but they do not cover the combined legacy-row -> tokenless announce -> token registration transition above. Could you please investigate these points and add cross-component/state-transition coverage before this is merged?

Prasad-D-Ware:
Thanks — this was a genuinely useful review. All eight points were valid; I reproduced the first one before touching anything and it was worse than described. Pushed as five commits on top of a rebase onto current `main`.

**1. Legacy adoption leaving two active rows — fixed.**

Confirmed exactly as you described. I wrote the three-step sequence as a test first and it failed with two rows sharing one token. It was also worse than reported: when the *legacy* row was the muted one, the merge dropped the mute entirely, so a device the user had silenced came back audible.

`Upsert` now consults both keys and reconciles instead of returning on the first install-ID hit:

- both match (different rows) → merge into the install-ID row, drop the token row
- only install ID → update
- only token → re-key
- neither → insert

Two invariants are explicit in the merge, both pinned by tests: the **oldest `CreatedAt` wins** so paired-since never jumps forward, and **mute is sticky** — if either identity was muted the survivor stays muted. Tests: `TestLegacyRowMergesWhenAnnounceLandsBeforeRegistration`, `TestMergePreservesMuteFromEitherIdentity` (both orderings), `TestNoTwoRowsEverShareAToken`.

**2. Uninstall/reinstall — you were right that the description overstated it.**

The PR body claimed the token "rotates on reinstall, so one phone became several rows", which framed install IDs as solving reinstalls. They don't: the ID lives in AsyncStorage, which uninstall wipes. I've corrected the description to state the actual guarantee — install IDs remove duplicates from **token rotation**, a reinstall still creates a new row, but after (1) it can no longer produce two rows holding the same token, which was the part that doubled every notification. The UI copy makes no claim either way, so it needed no change.

**3. Unregister deleting the paired-phone row — fixed.**

`DeleteByToken` is now `UnregisterToken`: it clears the token and keeps the row, so a phone that switches notifications off stays listed as notifications-off with its mute state and paired-since intact. Only the desktop's `Delete` removes a row. Dead-token pruning goes through the same path.

One case your "where the phone remains paired" phrasing surfaced: a row keyed by a **synthesized** `legacy-` ID is a placeholder the phone never knows — it announces under its real ID — so clearing its token would strand a row nothing could ever reclaim. Those are deleted instead. The prefix is now a named constant rather than a string repeated in three places. Six tests, including persistence across reload so a restart cannot resurrect a registration the phone gave up.

**4. Displayed last-seen — fixed, and you were right to escalate it.**

I had it as a follow-up; "one of the primary roster states is incorrect" is the fairer framing. `Tracker` now retains timestamps well past the liveness TTL and exposes `LastSeen`, which the roster prefers over the stored value, falling back to it after a daemon restart when presence has nothing. `Live()` no longer deletes on read — that was discarding the very timestamps the label needs — so sweeping moved to `Touch`. Tests: `TestLastSeenOutlivesLiveness`, `TestListPrefersPresenceLastSeenOverStoredValue`.

**5. In-flight poll continuing after backgrounding — fixed.**

Needed slightly more than stopping the loop: `fetchAll` makes two sequential requests, so backgrounding between them still fired the second. The cleanup now marks the closure stopped **and** `fetchAll` checks between requests via a ref — a state value would read a stale closure mid-flight.

Disclosing a gap: this one has **no automated test**. `packages/mobile` has no React/component test harness, so there's nothing to drive an AppState transition mid-request against. Verified by inspection only. Happy to add the harness if you'd prefer that over an inspection-only fix.

**6. Mute persistence errors — fixed, including the ordering you flagged.**

`SetMuted` returns a sentinel (`ErrDeviceNotFound`) the controller maps to 404; anything else is now a 500 with its own code. You were also right about the write ordering: the in-memory value is rolled back when persistence fails, so dispatch can no longer honour a mute the caller was told had not been saved. Tests cover both the registry (via an unwritable directory, asserting the rollback) and the controller status mapping.

**7. OpenAPI contract — fixed.**

`installId` is no longer in the schema's `required` set, matching what the handler actually accepts. Regenerated; verified the `required` block is gone. No test on this one — the committed generated artifact is the guard, so a regression shows up in the diff.

**8. Presence cleanup only on roster read — fixed.**

`Touch` sweeps as well, and since the install ID is client-supplied I added a hard entry cap with oldest-first eviction on top of age-based expiry, so the bound holds regardless of who is sending what. `TestTouchPrunesWithoutLiveEverBeingCalled` drives past the cap and asserts both that the newest entry survives and the oldest is the one evicted.

**On cross-component coverage:** the registry tests now drive the full legacy-row → tokenless announce → token registration transition, which is the sequence that was uncovered.

**On the maintainability notes:** I have not acted on the service/port boundary, the frontend's hand-rolled `MobileDevice` cast, or the duplicated announce/register effects — they read as "worth considering" rather than blocking, and I would rather not widen this PR further. The `Presence`/`DeviceLive` double-injection is mitigated by a normalization step that defaults one to the other and warns when neither is wired, but the shape is still there. Happy to take any of them in a follow-up, or here if you would rather they land together.

Also rebased onto current `main`; the branch is mergeable again.

illegalcall:
Thanks for addressing the earlier feedback. I think we're aligned on the main model now: the install ID identifies the paired phone, the push token is optional, and losing a token should not automatically remove a paired phone.

I still want to push back on three data behaviors before merge:

1. **Unpairing needs to be different from turning notifications off.** Right now both paths call the same unregister operation, which clears the token but keeps the phone in the old desktop's roster. Keeping the row is correct when notifications are disabled. It is not correct when the user chooses “Disconnect & forget server” or switches to another desktop—the old desktop then lists a phone that is no longer paired. Could we split these into two operations: clear the token while staying paired, and remove/unpair from that daemon?

2. **Registry changes need to be all-or-nothing with the file write.** `Delete`, `UnregisterToken`, and `Upsert` change the in-memory registry before saving `push-devices.json`. If the save fails, the API reports failure but the running daemon continues with changed memory; after restart, the old disk state comes back. I reproduced this for remove and unregister. `SetMuted` already rolls back correctly on save failure—could we apply the same behavior to every registry mutation, including tombstones and identity merges?

3. **A registration with no identity should be rejected.** `POST /api/v1/push/devices` currently accepts `{}`, generates a new `legacy-*` ID, and permanently adds a tokenless phantom device. Repeating it grows the roster with rows that represent no phone. Backward compatibility only needs two valid shapes: a real `installId` with an optional token, or a valid token from an older client with no `installId`. If both are missing, the request should return 400.

These are the only remaining alignment blockers from my side. I'm not asking to widen this PR for architecture cleanup or other nits. After these data semantics are fixed, the branch is rebased onto current `main`, and CI reruns, I think we should be in good shape for final sign-off.

Prasad-D-Ware:
All three are fixed, and all three were right. Rebased onto current `main`.

**1. Unpairing is now a separate operation from turning notifications off.**

Two routes, two meanings:

- `DELETE /api/v1/push/devices/{token}` — clears the token, phone stays paired and listed as notifications-off. Unchanged.
- `DELETE /api/v1/push/pairings/{id}` — removes the row. `{id}` resolves as an install ID first and falls back to the token, so a build predating install IDs can still unpair cleanly.

On the app side the split follows the intent that was already there: `forgetServer()` and the config-cleared effect in `PushManager` now unpair, while the settings toggle keeps unregistering. The comment on that effect already said *"when the user clears the server config (unpairs)"* — the distinction existed, my earlier change had collapsed both onto the same call.

One deliberate asymmetry: **unpair writes no tombstone.** Tombstones exist to re-mute a device the *desktop* removed against that device's wishes. Here the phone is the one asking to leave, so a later deliberate re-pair should come back clean rather than silently muted. `TestUnpairRemovesTheRowWhileUnregisterKeepsIt` covers both directions plus the by-token fallback and the no-tombstone rule.

**2. Every registry mutation is now all-or-nothing with the file write.**

You were right that only `SetMuted` had it. `Upsert`, `Delete` and `UnregisterToken` all mutated memory and then persisted. Rather than repeat the rollback four times and rely on remembering it a fifth, everything now goes through one `commitLocked(fn)` that snapshots both `devices` and `removed`, applies the change, persists, and restores both on failure — so tombstones and identity merges are covered by construction.

`TestEveryMutationRollsBackWhenTheWriteFails` drives each mutation against an unwritable directory. It includes the case I would otherwise have missed: after a failed `Delete`, no tombstone may survive, or a device the user never successfully removed comes back muted on its next registration.

**3. Identity-less registrations are rejected.**

`POST /api/v1/push/devices` with neither `installId` nor `token` is now `400 MISSING_DEVICE_IDENTITY`. I reproduced the accumulation first — three identity-less calls, three permanent tokenless rows, no phone behind any of them.

The two valid shapes are unchanged and both covered by tests: a real `installId` with an optional token, or a valid token from an older client with no `installId`.

**Verification:** backend suites and `-race` green, `golangci-lint` clean on every touched package, mobile 385 tests, frontend renderer 1542, no OpenAPI/TS drift. The three `disconnect` tests that pinned the old "forgetServer unregisters" contract now assert the unpair contract instead — they failed first, which is what I wanted from them.

The new route is declared in `specgen/build.go` alongside the others, so the route/spec parity test covers it.

illegalcall (approved):

PR #3675, opened 2026-08-07
feat(agent): add Muse model catalog

nikhilachale:
## Summary
- add a static Muse model catalog with muse-spark as the default
- keep Muse custom model entry enabled for future model IDs
- cover Muse catalog behavior in model catalog tests

## Tests
- go test -count=1 ./internal/adapters/agent/modelcatalog
- go test -count=1 ./internal/service/agent
- go test -count=1 ./internal/httpd/controllers

## Notes
- npm run lint was also run, but it failed in unrelated packages outside this change: backend/internal/adapters/agent/fake, kilocode, opencode, runtime/tmux, and backend/internal/cli.


nikhilachale:
Screenshot 2026-08-07 at 07 09 38

illegalcall (approved):

PR #3676, opened 2026-08-07
feat(chat): add Codex prompt editing with conversation branches

nikhilachale:
## Summary

- add Copy and Edit actions to human messages in structured Chat sessions
- support editing for Codex Chat sessions by forking the provider conversation immediately before the selected prompt
- preserve the original conversation as a sibling branch instead of rolling back or deleting later history
- add previous/next branch navigation and persist the active branch across daemon restarts
- keep the existing worktree and its file changes untouched when a conversation branches
- preserve the bottom composer draft while editing inline

## Provider support

- Codex is currently the only provider with targeted conversation-fork support, so Edit is available only for Codex Chat sessions
- the UI is capability-gated; providers without a verified fork implementation do not show Edit
- adding another provider requires a structured fork-or-replay adapter, durable conversation IDs, and reliable resume support
- Copy remains available independently of conversation-fork support

## Architecture

- add a targeted ChatForker provider boundary
- persist conversation branches and active-branch state in SQLite migration 0086
- expose edit/fork and branch-switching operations through the conversation API
- project only the active branch and its ancestry into the rendered timeline
- keep nested lineage cutoffs monotonic and flatten repeated edits into one navigation group

## Behavior and safety

- editing creates a new conversation branch; it does not use provider rollback
- worktree files are never reverted when switching conversation branches
- branch creation is blocked while a turn is active
- failures before branch activation leave the original branch active
- provider send failures leave an empty new branch active and retain the inline draft for a retry on that same branch

## Verification

- npm run frontend:typecheck
- npm --prefix frontend run typecheck:e2e
- npx vitest run (2,121 passed; 1 skipped)
- Go build and vet
- focused Go race tests for chat service and SQLite store
- golangci-lint v2.12.2 (0 findings)


nikhilachale:
Screenshot 2026-08-07 at 07 30 56

nikhilachale:
Screenshot 2026-08-10 at 12 30 03
Screenshot 2026-08-10 at 12 30 07

i-trytoohard:
Match confidence: `0.62` (fuzzy).

Relevant discussion excerpts:
- - *vanshx999* said he added **edit** and **copy** features in ChatUI so users can edit a prompt and create a separate chat, similar to Codex. - *vanshx999* said **activity status** is broken for some agents, specifically **Cursor** and **Kimi**, and that he is testing all agents now. - *vanshx999* said he is also [...]
- [54:09] !adil: Yeah, like the prompt is same itself right now also for everyone. But we have changed few words to make it more clear. [54:20] kinda_pulkit: Got it. [54:24] kinda_pulkit: Also, for in ChatUI, I added the edit and copy feature so that we can edit our prompt and create a separate chat for that. I [...]
- [54:20] kinda_pulkit: Got it. [54:24] kinda_pulkit: Also, for in ChatUI, I added the edit and copy feature so that we can edit our prompt and create a separate chat for that. I mean, like the Codex, we have edit feature. And same copy feature for that. And yeah, as we discussed Saturday, that activity status is [...]

illegalcall (approved):

PR #3678, opened 2026-08-07
fix(terminal): stop Windows app menu from stealing Ctrl+C/Ctrl+V in t…

### LINKED ISSUE #3659: bug(terminal): Ctrl+C does not interrupt processes in standalone shell terminals
## Bug

Ctrl+C does not kill/interrupt the foreground process when typed inside a **standalone shell terminal** (opened via the `+` button, `Ctrl+```, or the `/terminals` screen). The keystroke appears to be swallowed — no SIGINT reaches the shell.

## Expected behavior

Ctrl+C sends `\x03` (ETX) to the PTY, which the shell interprets as SIGINT, interrupting the foreground process — same as in a normal terminal or an agent session pane.

## Likely root cause(s)

The standalone shell input path is identical to the agent session path (`XtermTerminal` → `emitUserInput` → `mux.sendInput` → WebSocket → `attachment.write` → `pty.Write`). The shell and session terminals share the same `TerminalPane`, `useTerminalSession`, and mux bridge. So the difference must be in how the keystroke is intercepted *before* it reaches `mux.sendInput`.

Three candidate interceptors, all of which are stronger on **Windows**:

### 1. Electron menu `{ role: "copy" }` accelerator (Windows/Linux)

`frontend/src/main/menu.ts` installs a role-based menu on non-macOS that includes `{ role: "copy" }`. Even when the menu bar is hidden (`autoHideMenuBar`), the accelerators stay alive (confirmed in `main.ts` comments). On Windows, `{ role: "copy" }` binds to **Ctrl+C** at the Electron/OS level — the keystroke is consumed by Electron's clipboard copy before xterm.js ever sees the `keydown` event. This would explain why Ctrl+C does nothing: Electron copies (nothing selected → no-op) and never forwards the key.

### 2. `forceSelectionMode` + `isTerminalCopyShortcut` (Windows)

`XtermTerminal.tsx:243` calls `forceSelectionMode(term)` for **every** terminal, which forces all mouse interactions into text-selection mode. On Windows, `isTerminalCopyShortcut` (line 115) returns `true` for plain `Ctrl+C` (without Shift). Combined with force-selection, a stray selection from clicking to focus means Ctrl+C is intercepted as copy inside xterm's `attachCustomKeyEventHandler` — returning `false`, preventing xterm from processing it as `\x03`.

Test coverage at `XtermTerminal.test.tsx:343` explicitly asserts this Windows behavior (copies selected text with plain Ctrl+C), confirming it is by design for Windows. But it means Ctrl+C as SIGINT is unreliable whenever any text is selected.

### 3. Visibility/focus gate in `useTerminalSession`

`useTerminalSession.ts:642-648` drops all user input when `optionsRef.current.isVisible === false`. If the standalone terminal's visibility tracking is out of sync (e.g., the pane reports not-visible despite being on screen), every keystroke including Ctrl+C is silently dropped.

## Reproduction

1. Open AO desktop app
2. Open a standalone terminal (`Ctrl+`` ` or the `+` button on the tab strip)
3. Run a long-running command: `sleep 100` or `ping localhost`
4. Press **Ctrl+C**
5. Observe: the process is **not** interrupted (no SIGINT)

## Environment

- Platform: needs confirmation (Windows suspected; behavior may differ on macOS/Linux)
- AO version: latest `main`

## Investigation notes

- Ctrl+C **does** work in agent session terminals (codex/opencode panes) — the bug is specific to standalone shells
- `forceSelectionMode` is already tracked separately in #3640 (mouse reporting suppression), but its interaction with Ctrl+C copy-on-Windows is a distinct issue
- The `onKey` handler at `XtermTerminal.tsx:695` (not `onData`, per the comment at line 690) is the only keyboard→mux bridge — if the key never reaches `onKey`, the mux never sees it

## Suggested fix direction

- On Windows, decouple Ctrl+C-as-SIGINT from Ctrl+C-as-copy: only treat Ctrl+C as copy when there is an **active selection made by the user** (not a force-selection from a click-to-focus), or require Ctrl+Shift+C for copy in terminals (matching native Windows Terminal behavior)
- Verify the Electron menu `{ role: "copy" }` accelerator is not swallowing Ctrl+C before it reaches the renderer — may need a `before-input-event` handler to intercept and re-route Ctrl+C to the focused terminal

anmol0b:
> I think removing the roles is the right approach. The only real alternative would be remapping the accelerators, but that would just steal another shortcut (`Ctrl+Shift+C`) that xterm already uses. Since Electron menu accelerators are static, there's no practical way to enable them only for text inputs.
> I don't see this regressing copy/paste either. Terminals already handle copy/paste themselves, and regular text fields on Windows rely on Chromium rather than Electron menu roles. I'd just do a quick manual check on Windows to confirm copy/paste still works in settings or chat inputs.
> The change is nicely scoped—just removing the two roles and adding a regression test. The only thing still worth verifying before merge is that `Ctrl+C` once again sends `SIGINT` on a real Windows machine, since that's not something the author can test from macOS.

## What

Remove `{ role: "copy" }` and `{ role: "paste" }` from the Windows-only
application menu (`frontend/src/main/menu.ts`), and add a regression test
asserting those roles are never bound (`frontend/src/main/menu.test.ts`).

## Why

On Windows, the Electron app menu installed `role:copy` / `role:paste`, whose
default accelerators are **Ctrl+C** / **Ctrl+V**. Electron resolves application-menu
accelerators in the **main process before** the renderer's `keydown`, so the keystroke
was consumed by the role and never reached xterm.js. As a result `\x03` (ETX → SIGINT)
was never written to the PTY, and a foreground process (`sleep 100`, `ping`) could not
be interrupted with Ctrl+C in a standalone shell terminal (and likewise Ctrl+V paste
was silently dropped).

fixes : #3659

## How

 **Removed `role:copy` / `role:paste`** from `buildWindowsAppMenuTemplate()` (kept
  `undo`/`redo`/`cut`/`selectAll` — they don't collide with terminal SIGINT). With the
  roles gone, Ctrl+C reaches xterm and falls through to `\x03` (SIGINT) when there is no
  selection, and Ctrl+V reaches the existing paste handler.
- **Why remove instead of override the accelerator:** setting
  `{ role: "copy", accelerator: "Ctrl+Shift+C" }` would just move the theft onto
  Ctrl+Shift+C, which xterm already owns for terminal copy (`XtermTerminal.tsx:521`).
  **Gating a menu role on live selection state is not feasible** — Electron menu
  accelerators are static, registered at menu-build time — so removal is the correct fix.
- **Why this is safe for non-terminal text fields:** the "menu must declare copy/paste
  roles for editing shortcuts to work" quirk is **macOS-only**; on Windows/Linux Chromium
  handles copy/paste natively. This menu is `win32`-only (`main.ts:317`) anyway.
- **Copy/paste in terminals is preserved** via: selection auto-copy
  (`XtermTerminal.tsx:550`), Ctrl+Shift+C, Ctrl+Insert, the right-click context menu, and
  Chromium-native in text fields. The visible `WindowTitlebar` Edit menu is a React
  dropdown whose `Ctrl+C`/`Ctrl+V` strings are visual labels only (clicks call
  `act("edit.copy")`, independent of Electron roles) — unaffected.
- **Scope is Windows-only by construction:** `menu.ts` is installed only when
  `process.platform === "win32"`. On macOS/Linux `isTerminalCopyShortcut`
  (`XtermTerminal.tsx:120`) does not match plain Ctrl+C, so SIGINT already worked there;
  this change does not touch those paths.

## Testing

- `npx vitest run --config vite.renderer.config.ts src/main/menu.test.ts src/renderer/components/XtermTerminal.test.tsx`
  → **49/49 pass**. The new guard asserts no `copy`/`paste` role is bound; the existing
  `XtermTerminal.test.tsx:322/343/365` trio still locks the selection-gated copy-vs-SIGINT
  behavior (no renderer change was needed).
- `npm run frontend:typecheck` → no new errors. (Pre-existing, unrelated `code-highlight-engine.ts`
  missing-module errors remain due to stale local `node_modules`; they resolve under CI's fresh install.)
- **Could not reproduce locally** — this machine is macOS, where `menu.ts` is never installed,
  so the bug never reproduces here. Local Ctrl+C working on macOS only proves the non-Windows path
  is untouched.
- ⏳ **Still required before merge (delegated to a Windows machine/VM or the on-call Windows pair
  @illegalcall / @somewherelostt):** in a real Windows build, run `sleep 100` then Ctrl+C in
  (a) a standalone shell and (b) an agent session pane — both must interrupt; plus click-to-focus
  then Ctrl+C (forceSelectionMode) must still send SIGINT; plus copy/paste still works in a normal
  text field.

## Checklist

### DIFF
--- a/backend/internal/adapters/runtime/conpty/spawn_windows.go
+++ b/backend/internal/adapters/runtime/conpty/spawn_windows.go
@@ -1,7 +1,8 @@
 //go:build windows
 
 // spawn_windows.go - real detached pty-host spawner for Windows using
-// CREATE_NEW_PROCESS_GROUP + DETACHED_PROCESS so the host survives daemon exit.
+// DETACHED_PROCESS so the host survives daemon exit. Deliberately without
+// CREATE_NEW_PROCESS_GROUP — see ptyHostSysProcAttr.
 package conpty
 
 import (
@@ -21,6 +22,28 @@ import (
 
 // readyRE matches the "READY:<pid> <port>" line printed by RunHost.
 var readyRE = regexp.MustCompile(`READY:(\d+) (\d+)`)
+
+// ptyHostSysProcAttr returns the Windows process-creation attributes for the
+// detached pty-host.
+//
+// DETACHED_PROCESS keeps the child console-less so it survives the parent's
+// console closing; HideWindow suppresses the console flash.
+//
+// CREATE_NEW_PROCESS_GROUP must NOT be set here, even though it would
+// "insulate" the host from Ctrl+C sent to the parent. pty-host is the process
+// that calls CreatePseudoConsole, and conhost can only deliver a synthesized
+// CTRL_C_EVENT (from a 0x03 byte written to the ConPTY input pipe) to
+// processes in its own console process group — isolating pty-host's group
+// severs that delivery, so Ctrl+C in AO terminals never interrupts foreground
+// processes (ping -t keeps running). DETACHED_PROCESS already provides the
+// insulation: a process attached to no console receives no console control
+// events. See https://learn.microsoft.com/en-us/answers/questions/5832200/
+func ptyHostSysProcAttr() *windows.SysProcAttr {
+	return &windows.SysProcAttr{
+		CreationFlags: windows.DETACHED_PROCESS,
+		HideWindow:    true,
+	}
+}
 
 const spawnReadyTimeout = 10 * time.Second
 
@@ -90,14 +113,7 @@ func defaultSpawnHost(ctx context.Context, sessionID, cwd string, argv []string,
 	cmd.Dir = cwd
 	cmd.Env = merged
 
-	// Windows process-creation flags: detached + hidden console.
-	// ponytail: DETACHED_PROCESS puts the child in its own console; without it
-	// the child is killed when the parent's console closes. CREATE_NEW_PROCESS_GROUP
-	// insulates it from Ctrl+C sent to the parent. windowsHide suppresses the flash.
-	cmd.SysProcAttr = &windows.SysProcAttr{
-		CreationFlags: windows.DETACHED_PROCESS | windows.CREATE_NEW_PROCESS_GROUP,
-		HideWindow:    true,
-	}
+	cmd.SysProcAttr = ptyHostSysProcAttr()
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
--- a/backend/internal/adapters/runtime/conpty/spawn_windows_test.go
+++ b/backend/internal/adapters/runtime/conpty/spawn_windows_test.go
@@ -1,0 +1,23 @@
+//go:build windows
+
+package conpty
+
+import (
+	"testing"
+
+	"golang.org/x/sys/windows"
+)
+
+// TestPtyHostSysProcAttr guards the ConPTY Ctrl+C delivery path: pty-host calls
+// CreatePseudoConsole, and conhost only delivers synthesized CTRL_C_EVENTs to
+// processes in its own console process group, so the host must stay detached
+// (DETACHED_PROCESS) but must NOT be isolated into a new process group.
+func TestPtyHostSysProcAttr(t *testing.T) {
+	attr := ptyHostSysProcAttr()
+	if attr.CreationFlags&windows.CREATE_NEW_PROCESS_GROUP != 0 {
+		t.Fatalf("CreationFlags = %#x, must not include CREATE_NEW_PROCESS_GROUP (breaks ConPTY Ctrl+C delivery)", attr.CreationFlags)
+	}
+	if attr.CreationFlags&windows.DETACHED_PROCESS == 0 {
+		t.Fatalf("CreationFlags = %#x, want DETACHED_PROCESS set", attr.CreationFlags)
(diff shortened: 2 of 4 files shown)


illegalcall:
@somewherelostt Can you test this?

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari (changes requested):
Blocker: I tested this PR on Windows at commit f1ba62e, but Ctrl+C still does not interrupt a foreground process.

  Reproduction:

  1. Open a standalone AO terminal.
  2. Run ping -t 127.0.0.1.
  3. Press Ctrl+C with no visible text selection.

  Expected: ping stops and PowerShell returns to the prompt.
  Actual: ping continues running.

  The current test only verifies that the Electron copy/paste menu roles were removed; it does not verify that Ctrl+C reaches the
  PTY/ConPTY as \x03. Another shortcut or terminal-input layer may still be consuming the key. This needs further investigation and
  a real Windows verification before merging.

Rishet11:
@anmol0b
Closing this PR as stale after 30+ days open with no author activity since your last update.

PR #3680, opened 2026-08-07
fix: make jump-to-latest button opaque

illegalcall:
## Summary

- keep the chat Jump to latest control opaque in dark mode
- use an opaque surface color for its hover state
- add regression coverage for the inherited outline-button classes

## Screenshot

Jump to latest button with an opaque dark background

## Tests

- npm test -- --run src/renderer/components/chat/ChatWorkspace.test.tsx
- npm run typecheck

## Risks

- Low: scoped to the floating chat control styling and its test


illegalcall (commented):
## Review verdict: Ready

No blocking findings. The change is surgical and matches the PR scope: the local classes override the shared outline variant so the button and hover remain opaque in dark mode, and the regression test verifies both the desired utilities and removal of the inherited transparent utilities.

- **Standards:** Pass; no documented violations or material code smells.
- **Spec:** Pass; all requested behavior and regression coverage are present with no scope creep.

Validation:
- `npm test -- --run src/renderer/components/chat/ChatWorkspace.test.tsx` (20 tests passed)
- `npm run typecheck`
- CI checks at the reviewed head are green.

Ready / approved.

nikhilachale (approved):

PR #3683, opened 2026-08-07
fix(harness): detect Muse awaiting-input activity state

AgentWrapper:
## Summary

- classify captured Meta Muse TUI markers for active generation, structured input, and the idle composer
- opt Muse into continuous terminal reconciliation so a picker can move active/idle sessions to `waiting_input` and resumed generation can move them back to `active`
- preserve Muse lifecycle hooks and avoid broad tool hooks, since the native structured-input control emits no hook callback

## Verification

- `env -i HOME="$HOME" PATH="$PATH" TMPDIR="${TMPDIR:-/tmp}" npm run lint`
- focused Muse, activity-dispatch, observer, and lifecycle Go tests
- real Meta Muse CLI E2E through this checkout daemon: session 106 changed `working/active` -> `needs_input/waiting_input` while its native picker was visible; session 107 changed `needs_input/waiting_input` -> `working/active` after AO answered the picker and Muse resumed a live tool run


AgentWrapper:
Live regression verification on macOS with Meta Muse Code 0.1.0:

- Captured the compact picker variant whose current marker is `◆ Request user input …` and which omits the long `Enter to select` footer.
- Added that captured form as a fixture, including stale `Working … esc to interrupt` scrollback above it.
- Restarted the PR dev Electron app against real AO data while `muse-fix-demo` remained at the picker.
- Confirmed `/api/v1/sessions` changed the session from `working` / `active` to `needs_input` / `waiting_input`.
- Verified `go test ./...` with AO worker runtime variables removed and `npm run typecheck`.

Commit: `be468ad85`.

PR #3685, opened 2026-08-07
fix(desktop): recognise own daemon across AppImage relaunches via stable AO_APPIMAGE identity

### LINKED ISSUE #3679: Linux/AppImage: identity check fails after relaunch-to-update (random /tmp/.mount_agent-* path), app stuck with no sessions and no way to add projects
## Summary

On Linux (AppImage), relaunching the app after an update leaves it permanently stuck on **"AO daemon identity check failed" (`identity_mismatch`)**. The board shows "Could not load sessions." and projects can no longer be added — the app is unusable until the surviving daemon is killed by hand.

The identity check compares the running daemon's `executablePath` against the freshly resolved bundled daemon path. Under an AppImage those paths can never match across two launches, because the AppImage is mounted at a *random* `/tmp/.mount_agent-XXXXXX` path each time it starts.

## Steps to reproduce

1. Run the Linux AppImage build and let the daemon start (a project was already initialized in my case).
2. Switch the update channel to **nightly**.
3. Click **Relaunch to update**.
4. App comes back up.

## Actual behavior

Error toast, and the app is dead in the water:

```
AO daemon identity check failed                      identity_mismatch

Another AO daemon is already running from
/tmp/.mount_agent-1Qs4N6/resources/daemon/ao; expected
/tmp/.mount_agent-mDQfUL/resources/daemon/ao. Stop the other daemon
before using this app.

Stop the conflicting daemon, then restart the desktop app.
```

Board renders "Could not load sessions." and there is no way to add a project, for as long as the status reads `identity_mismatch`. The toast carries no Restart button in that state: `canRestart` in [DaemonFailureBanner.tsx:21](frontend/src/renderer/components/DaemonFailureBanner.tsx#L21) covers `not_ready`, `spawn_failed`, `exited` and `daemon_unreachable`, but not `identity_mismatch` — so the remediation text ("Stop the conflicting daemon, then restart the desktop app") has nothing behind it.

**It does recover on its own, after a delay.** Worth stating precisely, because it caps the severity:

- The old daemon is only alive because its supervisor grace fires ~5s after the previous Electron process exits ([supervisor.go](backend/internal/daemon/supervisor/supervisor.go)). The relaunch lands inside that window and probes a live daemon at the old mount path.
- `useDaemonStatus` re-polls every 2s while not ready ([useDaemonStatus.ts:8](frontend/src/renderer/hooks/useDaemonStatus.ts#L8)). Once the old daemon exits, `refreshDaemonStatus` finds nothing and — since the prior status was an error carrying a pid/port — rewrites it to `daemon_unreachable` ([main.ts:631](frontend/src/main.ts#L631)).
- `daemon_unreachable` *is* in `canRestart`, so a Restart button appears and clicking it spawns a healthy daemon. Relaunching the app instead works equally well.

So the lived symptom is "relaunch-to-update drops you into a dead app for several seconds, with a misleading and unactionable error, until the banner quietly turns into a different one you can act on" — not a permanent brick. Nothing spawns a daemon automatically: `daemon:start`/`daemon:restart` fire only from that button, plus `void startDaemon()` at boot.

Verified on my machine: the mismatched daemon was gone minutes later and the next launch spawned cleanly from a third mount (`.mount_agent-OgFgtF`).

Note this also means the normal healthy path is being lost: attaching would have worked. The supervisor cancels a pending grace timer on reconnect ([supervisor.go:107](backend/internal/daemon/supervisor/supervisor.go#L107)), and the attach path already re-links app-owned daemons, so the relaunch is *designed* to reattach — the identity check is the only thing preventing it.

## Expected behavior

Relaunching to update should reattach to (or cleanly replace) the daemon it spawned. An AppImage relaunching itself is not a "foreign daemon" and should not be treated as one.

## Root cause

`daemonIdentityError` in [frontend/src/main.ts:540-547](frontend/src/main.ts#L540-L547) for `source === "bundled"`:

```ts
if (launch.source === "bundled") {
    if (!probe.executablePath) { ... }
    if (!samePath(probe.executablePath, launch.command)) {
        return `Another AO daemon is already running from ${probe.executablePath}; expected ${launch.command}. ...`;
    }
}
```

`launch.command` comes from `resolveDaemonLaunch` in [frontend/src/shared/daemon-launch.ts:46-52](frontend/src/shared/daemon-launch.ts#L46-L52), which joins `process.resourcesPath` with `daemon/ao`. On an AppImage, `process.resourcesPath` lives under the squashfs FUSE mount, e.g. `/tmp/.mount_agent-1Qs4N6/resources` — and the `.mount_agent-XXXXXX` suffix is regenerated on every AppImage start.

So the path-equality identity check is structurally guaranteed to fail on Linux/AppImage whenever a daemon from a previous launch is still alive. The two paths in the error above differ *only* in the random mount suffix; they are the same binary from the same install.

## Notes / possible fixes

Not prescribing an approach, but the options seem to be:

- Detect the AppImage case (`APPIMAGE` / `APPDIR` env vars are set by the runtime) and identify the daemon by something stable — the `APPIMAGE` path, an install id, or a build/version fingerprint reported by `/readyz` — rather than by the mount-local executable path.
- Normalize/strip the `/tmp/.mount_*` prefix before comparing, so only the in-package suffix (`resources/daemon/ao`) is matched.
- Independently: when the app is relaunching *itself* for an update, the app-owned daemon should be shut down as part of the relaunch rather than surviving into the new process to be rejected.
- Independently of the root cause, `identity_mismatch` should not be a dead end. Since the state already resolves itself into a restartable one seconds later, the cheapest improvement is to stop presenting it as terminal — either add it to `canRestart`, or soften the copy so it does not read as "you must go kill a process by hand".

## Environment

- Linux, AppImage build (`agent-orchestrator-linux-x64.AppImage`)
- Channel: nightly (failure appeared right after switching from the previous channel and clicking "Relaunch to update")

Confirms the 6-character-prefix caveat above: the AppImage filename starts `agent-`, so every channel built under this name mounts as `.mount_agent-XXXXXX` and the mount name alone cannot distinguish them.

anmol0b:
## What

Fixes the Linux/AppImage daemon identity check so the desktop app recognises its own daemon across launches. The daemon now reports the stable outer `.AppImage` file path (`AO_APPIMAGE`, set by the Electron app at spawn, echoed as `appImagePath` in `/healthz`/`/readyz`), and the app compares that value when running under AppImage instead of the transient `/tmp/.mount_*` executable path. Outside AppImage the exact `executablePath` comparison is unchanged.

Also adds a Restart action to the `identity_mismatch` failure banner so a genuine foreign-daemon conflict is actionable instead of a dead end.

## Why

Fixes #3679.

On Linux/AppImage, every launch mounts the app at a random `/tmp/.mount_agent-XXXXXX` path. After "Relaunch to update", the surviving daemon (kept alive ~5s by the supervisor grace) reported its executable under the *previous* mount, so the exact-path identity check could never match. The app showed `identity_mismatch`, the board rendered "Could not load sessions", and the banner had no recovery action until the old daemon exited and the status degraded into `daemon_unreachable`.

## How

- `backend`: `daemonProbePayload` includes `appImagePath` when `AO_APPIMAGE` is set (dedicated env var, matching the `AO_OWNER`/`AO_APP_RUN_ID` family; no OpenAPI impact — health routes are outside the generated spec).
- `frontend/main`: passes `AO_APPIMAGE: process.env.APPIMAGE` to the daemon at spawn; the bundled branch of `daemonIdentityError` delegates to a pure, unit-tested helper.
- `frontend/shared/daemon-launch.ts`: new `bundledDaemonIdentityError()` — under AppImage compares the probed `appImagePath` against this process's `APPIMAGE`; otherwise falls back to the exact `executablePath` check. Probes that cannot prove identity fail closed, as before.
- `frontend/shared/daemon-attach.ts`: `DaemonProbe`/`parseDaemonProbe` carry `appImagePath`.
- `DaemonFailureBanner`: `identity_mismatch` added to `canRestart`. Restart only re-runs attach/spawn; it never kills a daemon this process didn't spawn, so a truly foreign daemon simply re-surfaces the error.

Trade-off / rejected alternative: stripping the `/tmp/.mount_*` prefix and comparing the in-package suffix was rejected — the suffix is identical across every build and channel, so it would silently attach to a genuinely different install's daemon. Killing the daemon inside `quitAndInstallUpdate` was rejected as racy and insufficient (doesn't cover crash-restart or quick manual relaunch within the grace window).

Known one-time limitation: the first update *into* a build with this fix still hits the old behaviour once (a pre-fix daemon doesn't report `appImagePath`), fails closed for the ~5s grace window, then self-resolves — now with a Restart button.

## Testing

- `cd frontend && npx vitest run src/shared/daemon-launch.test.ts src/shared/daemon-attach.test.ts src/renderer/lib/daemon-failure.test.ts` — 49 tests pass, incl. new cases: cross-mount reattach (the bug scenario, using the exact paths from the issue), foreign-install mismatch, and fail-closed when identity is unprovable.
- `cd backend && go test ./...` — pass (new `TestHealthProbesIncludeAppImageIdentity` covers `/healthz` + `/readyz` with `AO_APPIMAGE` set and unset).
- `golangci-lint` v2.12.2 — 0 issues. Frontend `tsc --noEmit` clean on touched files.
- Not yet verified: the real Linux/AppImage relaunch-to-update flow (needs a Linux machine; the bug cannot be reproduced on macOS/Windows). Reviewer smoke test: launch AppImage → switch channel → Relaunch to update → app reattaches with no `identity_mismatch` banner.

frontend tests : 

Screenshot 2026-08-07 at 3 39 29 PM

backend test (added) : 

Screenshot 2026-08-07 at 3 49 19 PM

## Checklist


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

harshitsinghbhandari (approved):
LGTM

neversettle17-101:
Merging this as it is approved by @harshitsinghbhandari

PR #3692, opened 2026-08-07
feat(auth): add native WorkOS desktop sign-in

amoreX:
## Summary
- add native WorkOS AuthKit PKCE sign-in through the system browser and return to the desktop app through the registered `ao-app://callback` protocol
- encrypt access and refresh tokens with Electron `safeStorage`, deduplicate refreshes, preserve credentials on retryable failures, and make explicit sign-out authoritative over in-flight refreshes
- add localized sidebar account controls while keeping access and refresh tokens out of renderer IPC
- register Linux deb/rpm protocol metadata and make raw AppImages installed by `ao start` create and verify the freedesktop `ao-app` URL handler
- require the non-secret `VITE_WORKOS_CLIENT_ID` repository variable in canonical release/artifact workflows so an official build cannot silently ship with sign-in disabled

## Intentional behavior
- sign-out clears the desktop session locally and forces fresh authentication (`maxAge: 0`) next time; it does not open the browser or revoke the WorkOS server session
- when protected OS storage is unavailable, credentials remain memory-only and disappear after restart
- terminal refresh-token failures clear the session; network, rate-limit, timeout, and server failures retain encrypted credentials for retry
- Linux AppImage protocol registration uses the standard XDG applications and MIME-association locations; AO runtime state remains under `~/.ao`

## Test plan
- [x] `npm run frontend:typecheck`
- [x] full frontend suite: 175 files, 2,170 tests passed before stack propagation
- [x] focused auth suite including retryable refresh preservation and refresh/sign-out race regression
- [x] release workflow regression test requiring `VITE_WORKOS_CLIENT_ID`
- [x] forge/AppImage protocol metadata tests
- [x] complete `backend/internal/cli` tests and focused golangci-lint (`0 issues`)
- [x] repository variable `VITE_WORKOS_CLIENT_ID` configured for official workflows
- [x] Linux registration code writes an XDG desktop entry, sets `x-scheme-handler/ao-app`, and verifies the selected handler before launching


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,270 @@
...
+
+const emptyStore = (): AuthStore => ({ session: null, pkce: null });
+
+function storePath(dataDir: string): string {
+  return path.join(dataDir, AUTH_STORE_FILE);
+}
+
+function encodeStore(store: AuthStore): Buffer {
+  const value = JSON.stringify(store);
+  if (safeStorage.isEncryptionAvailable()) {
```
Pulkit7070: [high] When `safeStorage.isEncryptionAvailable()` is false (headless Linux without a keyring, some remote-desktop/CI setups), `encodeStore` writes the store, including the long-lived `refreshToken`, as plaintext JSON to disk. `mode 0o600` scopes it to the same UID but does not protect against backups, synced home directories, or other same-user processes, and the fallback is silent, the user is never told their tokens are unencrypted.

For an auth store holding a rotating refresh token, consider either not persisting the refresh token when encryption is unavailable (keep the session in memory for the run) or surfacing a visible warning. Which behavior is intended here?

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,270 @@
...
+): Promise<CloudSession | null> {
+  if (!workos) return null;
+  const store = await readAuthStore(dataDir);
+  if (!store.session) return null;
+  if (!tokenExpiresSoon(store.session.accessToken)) {
+    return publicSession(store.session);
+  }
+
+  try {
+    const refreshed = await workos.userManagement.authenticateWithRefreshToken({
```
Pulkit7070: [medium] `getCloudSession` refreshes inline when the access token is near expiry, with no in-flight de-duplication. Two near-simultaneous callers (multiple windows mounting `useCloudSession`, or a `getSession` racing the `sessionChanged` listener) both call `authenticateWithRefreshToken` with the same refresh token. WorkOS rotates refresh tokens, so the first call consumes it and the second fails; the catch then clears the store and signs the user out.

Guard the refresh behind a single in-flight promise (keyed by data dir) so concurrent callers share one refresh result rather than each spending the token.

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,270 @@
...
+  import.meta.env.VITE_WORKOS_CLIENT_ID?.trim() ||
+  (process.env.VITEST ? "client_test" : "");
+const REDIRECT_URI = "ao-app://callback";
+const AUTH_STORE_FILE = "cloud-auth.bin";
+const LEGACY_SESSION_FILE = "cloud-session.json";
+const PKCE_TTL_MS = 10 * 60 * 1000;
+const workos = CLIENT_ID ? createWorkOS({ clientId: CLIENT_ID }) : null;
+
+export interface CloudSession {
+  accessToken: string;
```
Pulkit7070, line 23: [medium] `CloudSession` carries `accessToken`, and `publicSession` strips only `refreshToken`, so `cloud:getSession` and the `cloud:sessionChanged` broadcast hand the bearer access token to the renderer, yet the only consumer (Sidebar) uses `user.email`. Keeping the token in the main process and exposing only the redacted session (email + provider + signed-in state) reduces the blast radius of any renderer-side injection at no cost here.

Related: `notifyRenderersOfCloudSession` sends to `getAllWebContents()`, which includes the browser-panel WebContentsView. That view uses the annotate preload so it cannot read the message today, but scoping the notify to the app window(s) rather than every webContents would be safer as the surface grows.

Pulkit7070:
Three lower-severity follow-ups from the review, none blocking:

1. decodeStore (cloud-auth.ts:62) picks encrypted-vs-plaintext by `isEncryptionAvailable()` at read time, not by how the file was actually written. A file written encrypted and later read when encryption is unavailable (or vice versa) fails `JSON.parse`, gets swallowed to `emptyStore()`, and silently logs the user out. A one-byte format marker alongside the payload would make decode deterministic.
2. Protocol registration (forge.config.ts) only populates the macOS Info.plist; the PR's own checklist leaves Windows/Linux packaged `ao-app://` handling unchecked. Worth confirming before relying on the deep link on those platforms.
3. Test coverage: cloud-auth.test.ts (3 tests) covers happy path, state-mismatch, and sign-out. Not covered: unsolicited-callback-with-no-pending-PKCE rejection, expired-PKCE, wrong scheme/host, the `error=` callback branch, and the refresh+rotation flow (including the concurrency case above). The safeStorage-unavailable branch runs in the mock but nothing asserts the plaintext behavior.

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,291 @@
...
+    await rm(storePath(dataDir), { force: true });
+    return emptyStore();
+  }
+}
+
+async function writeAuthStore(
+  dataDir: string,
+  store: AuthStore,
+): Promise<void> {
+  if (!safeStorage.isEncryptionAvailable()) {
```
whoisasx: [P1] Treat Linux `basic_text` as unavailable encryption

This fallback is still unsafe on Linux. `safeStorage.isEncryptionAvailable()` can be true while `getSelectedStorageBackend()` is `basic_text`; Electron documents that backend as unprotected because it uses a hardcoded plaintext password. That means the rotating refresh token is still persisted in a trivially decryptable `cloud-auth.bin`, despite the comment below. Please route `basic_text` through the memory-only path as well and add a test for that backend.

frontend/forge.config.ts
```diff
@@ -33,6 +33,12 @@ const config: ForgeConfig = {
 		appBundleId: "dev.agent-orchestrator.desktop",
 		name: "Agent Orchestrator",
 		executableName: EXECUTABLE_NAME,
+		protocols: [
```
whoisasx: [P1] Register the callback scheme in every Linux artifact

`packagerConfig.protocols` only updates the macOS bundle. The deb/rpm maker options do not declare `mimeType: ["x-scheme-handler/ao-app"]`, and the custom AppImage bridge does not pass a protocol/MIME entry to electron-builder, so packaged Linux installs do not get the `.desktop` handler that lets the browser dispatch `ao-app://callback` back to AO. Please wire the scheme into all Linux makers and validate a packaged callback; the runtime call alone does not create the required package metadata.

frontend/src/main.ts
```diff
@@ -1770,7 +1782,61 @@ ipcMain.on(TRAY_SET_ATTENTION_STATE_CHANNEL, (event, state) => trayLifecycle.han
...
+	window.focus();
+}
+
+async function handleCloudDeepLinkAndFocus(url: string): Promise<void> {
+	try {
+		const session = await handleCloudDeepLink(url, cloudDataDir());
+		if (!session) return;
+		notifyRenderersOfCloudSession(session);
+		focusCloudWindow();
+	} catch (error) {
```
whoisasx, line 1821: [P2] Surface callback failures to the user

Every failure path from `handleCloudDeepLink` ends here and is only logged. If WorkOS returns an error, the PKCE request expires, state validation fails, or token exchange has a network error, the user sees no feedback and the existing app is not even focused on the second-instance path. Please focus the window and surface a bounded, auth-specific error state or dialog so sign-in can be retried intentionally.

whoisasx (changes requested):
Thanks for the work here. The revised implementation addresses the earlier token exposure and refresh-race concerns: the renderer receives only public account data, notifications are scoped to shell contents, rotating refreshes are deduplicated, and the explicit no-encryption path is memory-only. The PKCE/state flow is otherwise coherent, and the targeted auth tests, typechecks, and GitHub checks pass.

I am requesting changes for two remaining blockers:

- Linux `safeStorage` can still select `basic_text` while reporting encryption available, so refresh tokens remain effectively unprotected on those systems.
- Linux release artifacts do not register `x-scheme-handler/ao-app`, so the browser callback is not reliably deliverable to AO.

I also left a lower-priority UX comment for silent callback failures. One additional compatibility risk to track is that `@workos-inc/node@10.9.0` declares Node `>=22.11.0`, while Electron 33.4.11 embeds Node 20.18.3; direct PKCE generation worked under the embedded runtime, but this remains outside the SDK's supported runtime contract.

Once secure Linux storage fallback and packaged Linux protocol registration are addressed, this should be ready for another pass.

amoreX:
@whoisasx  resolved your comments<3

whoisasx:
too quick man.. 🤯

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,329 @@
...
+  readFile,
+  rm,
+  writeFile,
+} from "node:fs/promises";
+import path from "node:path";
+import type { CloudAccount } from "../shared/cloud-account";
+
+const CLIENT_ID =
+  import.meta.env.VITE_WORKOS_CLIENT_ID?.trim() ||
+  (process.env.VITEST ? "client_test" : "");
```
whoisasx, line 15: [P1] Wire the WorkOS client ID into release builds

This fallback makes every canonical artifact compile with WorkOS disabled: none of the current release/build workflows exports `VITE_WORKOS_CLIENT_ID`, and GitHub repository variables are not injected into a runner unless the workflow references them explicitly. The sidebar still renders the sign-in control, but `workos` remains `null` and every click only shows “WorkOS not configured.” Please pass a non-secret repository variable such as `${{ vars.VITE_WORKOS_CLIENT_ID }}` into every canonical make/publish job and add a packaged-build assertion that the expected client ID was compiled in.

frontend/makers/maker-appimage.ts
```diff
@@ -44,6 +46,7 @@ export default class MakerAppImage extends MakerBase<MakerAppImageConfig> {
 				config: {
 					appId: cfg.appId,
 					productName: cfg.productName ?? appName,
+					protocols: cfg.protocols,
```
whoisasx, line 49: [P1] Register the raw AppImage with the Linux desktop

Passing `protocols` here only places the MIME declaration in the desktop entry embedded inside the AppImage. AO’s canonical Linux path downloads the AppImage to `~/.ao`, chmods it, and executes it directly; it never installs that desktop entry. AppImage documents MIME registration as optional desktop integration performed by tools such as `appimaged` or AppImageLauncher, so a clean machine has no `x-scheme-handler/ao-app` association for the browser to invoke. Electron’s Linux runtime registration also relies on an installed desktop entry discoverable by `xdg-settings`. Please explicitly install/update and register a desktop entry for the downloaded AppImage, or use a callback mechanism that does not require system integration, then validate `ao start` followed by a real callback on Linux without an AppImage integration helper.

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,329 @@
...
+    });
+    const session = toStoredSession(
+      refreshed.accessToken,
+      refreshed.refreshToken,
+      refreshed.user,
+    );
+    await writeAuthStore(dataDir, { ...store, session });
+    return publicAccount(session);
+  } catch {
+    await removeAuthStore(dataDir);
```
whoisasx: [P1] Preserve credentials on retryable refresh failures

This catch treats every failure as proof that the refresh token is invalid. A network outage, timeout, rate limit, or WorkOS 5xx during startup therefore deletes the encrypted refresh token and permanently signs out a user whose credentials may still be valid. Please distinguish terminal OAuth failures such as an invalid or consumed refresh token from retryable transport/server failures; only the former should clear the store. For retryable failures, preserve the existing session and surface a retryable/unavailable result so a later call can recover without another browser login.

frontend/src/main/cloud-auth.ts
```diff
@@ -0,0 +1,329 @@
...
+    const refreshed = await workos!.userManagement.authenticateWithRefreshToken({
+      clientId: CLIENT_ID,
+      refreshToken: storedSession.refreshToken,
+    });
+    const session = toStoredSession(
+      refreshed.accessToken,
+      refreshed.refreshToken,
+      refreshed.user,
+    );
+    await writeAuthStore(dataDir, { ...store, session });
```
whoisasx: [P2] Prevent an in-flight refresh from undoing sign-out

This write is not coordinated with `signOutCloud`. If a refresh starts, the user signs out while the WorkOS request is pending, and the request then succeeds, this captured store is written after deletion and the account is authenticated again. I reproduced that sequence with a deferred refresh promise. Please serialize auth-store mutations or maintain a per-data-dir generation/cancellation marker and verify it before persisting the refresh result, with a regression test proving that sign-out remains final when a refresh resolves afterward.

whoisasx (dismissed):
Thanks for addressing the earlier feedback. The Linux `basic_text` fallback, deb/rpm protocol metadata, and visible callback-error handling are now implemented correctly.

I re-reviewed the complete authentication flow and am requesting changes for four remaining issues:

- The canonical release workflows never provide `VITE_WORKOS_CLIENT_ID`, so official artifacts compile with WorkOS disabled.
- The raw AppImage used by `ao start` embeds a desktop entry but never installs/registers it, so a clean Linux system cannot dispatch `ao-app://callback` to AO.
- Retryable refresh failures delete otherwise valid encrypted credentials and force a new login.
- An in-flight refresh can persist a new session after an explicit sign-out.

Verification performed at `41ec785`:
- All GitHub checks are green.
- `npm run typecheck` passes.
- The 11 focused cloud-auth and packaging tests pass.
- A disposable regression test reproduced the refresh/sign-out race: after sign-out, resolving the pending refresh caused `getCloudSession()` to return the account again.

The full frontend suite was not used as a verdict because unrelated socket/browser-runtime and landing tests timed out or failed in the isolated checkout. Once these four issues are addressed and the raw `ao start` AppImage callback is validated on a clean Linux environment, this will be ready for another pass.

Pulkit7070 (approved):

Pulkit7070 (approved):

amoreX:
> Thanks for addressing the earlier feedback. The Linux `basic_text` fallback, deb/rpm protocol metadata, and visible callback-error handling are now implemented correctly.
> 
> I re-reviewed the complete authentication flow and am requesting changes for four remaining issues:
> 
>     * The canonical release workflows never provide `VITE_WORKOS_CLIENT_ID`, so official artifacts compile with WorkOS disabled.
> 
>     * The raw AppImage used by `ao start` embeds a desktop entry but never installs/registers it, so a clean Linux system cannot dispatch `ao-app://callback` to AO.
> 
>     * Retryable refresh failures delete otherwise valid encrypted credentials and force a new login.
> 
>     * An in-flight refresh can persist a new session after an explicit sign-out.
> 
> 
> Verification performed at `41ec785`:
> 
>     * All GitHub checks are green.
> 
>     * `npm run typecheck` passes.
> 
>     * The 11 focused cloud-auth and packaging tests pass.
> 
>     * A disposable regression test reproduced the refresh/sign-out race: after sign-out, resolving the pending refresh caused `getCloudSession()` to return the account again.
> 
> 
> The full frontend suite was not used as a verdict because unrelated socket/browser-runtime and landing tests timed out or failed in the isolated checkout. Once these four issues are addressed and the raw `ao start` AppImage callback is validated on a clean Linux environment, this will be ready for another pass.

this done

amoreX:
Addressed the four remaining review findings:

- canonical release/artifact workflows now require the configured `VITE_WORKOS_CLIENT_ID` repository variable
- raw AppImages launched by `ao start` install and verify the Linux `ao-app` XDG handler
- retryable refresh failures retain encrypted credentials; terminal token failures still clear them
- generation invalidation plus serialized auth mutations prevent an in-flight refresh from restoring a signed-out session

Added focused regressions for release configuration, Linux registration, retry preservation, and the sign-out race. The full frontend suite and CLI tests pass; the CI lint annotation was also fixed and the same focused golangci-lint command now reports zero issues. Ready for re-review.

Issue #3523, opened 2026-08-03
Release pipeline bakes stale owner=AgentWrapper into app-update.yml — fleet updates depend on a fragile org-rename redirect (latent breakage)


AgentWrapper:
## Bug
Every installed desktop build fails "Check for Updates" with **`net::ERR_FAILED`** and can never auto-update. Root cause: the app's bundled `app-update.yml` still points at the **old GitHub org** (`owner: AgentWrapper`), which was renamed to `Untrivial-ai`. The updater's feed request therefore traverses a redirect chain it cannot follow.

![check for updates ERR_FAILED](https://raw.githubusercontent.com/Untrivial-ai/agent-orchestrator/issue-assets-updater-org-rename-err-failed/.issue-assets/updater-org-rename-err-failed/check-updates-err-failed.png)

**Source:** maintainer local report + screenshot (desktop app, macOS Apple Silicon), 2026-08-03 | **Reported by:** maintainer | **Analyzed against:** `773197f12` (main) | **Affected build:** `v0.11.2-nightly.202608021445`
**Confidence:** High — baked config read from the installed bundle and the full redirect chain reproduced with curl.
**Priority:** critical — core update path is broken for the entire pre/at-transfer user base; users are stranded and cannot receive any future fix through the normal channel (chicken-and-egg).

## Decisive evidence
1. Installed bundle `Agent Orchestrator.app/.../app-update.yml`:
   ```
   provider: github
   owner: AgentWrapper        <-- OLD org (renamed to Untrivial-ai)
   repo: agent-orchestrator
   ```
2. The updater builds its feed URL as `https://github.com/${owner}/${repo}/releases/download/v${version}/nightly${platformSuffix()}.yml` (`frontend/src/main/auto-updater.ts:90`; electron-updater's GitHub provider does the same from the bundled owner/repo). With `owner=AgentWrapper` the request chain is:
   ```
   github.com/AgentWrapper/agent-orchestrator/.../nightly-mac.yml
     -> 301  github.com/Untrivial-ai/agent-orchestrator/.../nightly-mac.yml   (ORG-RENAME hop, cross-owner)
     -> 302  release-assets.githubusercontent.com/...signed-blob...           (cross-HOST hop)
   ```
   (Verified with `curl -w '%{http_code} -> %{redirect_url}'`: AgentWrapper URL = 301 to Untrivial-ai; Untrivial-ai URL = 302 to release-assets.githubusercontent.com.)
[...]

PR #3695, opened 2026-08-07
fix(release): stop baking the pre-rename AgentWrapper org into release config

krishrathi1:
## What

`DEFAULT_RELEASE_REPO` in `frontend/forge.config.ts` and `DEFAULT_REPO` in `frontend/src/main/feature-builds.ts` still hardcoded `AgentWrapper/agent-orchestrator` -- the org this repo was renamed away from. Any build path that doesn't set `AO_RELEASE_REPO` explicitly bakes the stale owner into the shipped `app-update.yml`, including `build-artifacts.yml`'s dispatchable unsigned build, which never sets it at all.

Today auto-update only keeps working because GitHub's org-rename redirect (`AgentWrapper` -> `Untrivial-ai`) happens to still resolve. That redirect breaks permanently the moment the still-existing `AgentWrapper` account creates any repo named `agent-orchestrator`, at which point every fleet install baked with the stale owner loses auto-update for good.

## Fix

- `frontend/forge.config.ts`: `DEFAULT_RELEASE_REPO` -> `Untrivial-ai/agent-orchestrator`.
- `frontend/src/main/feature-builds.ts`: `DEFAULT_REPO` -> `{ owner: "Untrivial-ai", ... }` (kept in sync with the forge default per its own comment).
- `.github/workflows/build-artifacts.yml`: set `AO_RELEASE_REPO: ${{ github.repository }}` on the make step, matching the pattern `frontend-release.yml` already uses, so forks and dispatched builds bake their own owner instead of falling through to the production default at all.

## Scope note

This addresses the part of #3523 that lives in this repo. The issue also calls out the external `ao-releases` conductor and a macOS update-e2e regression guard; I don't have access to that conductor repo and don't have a macOS environment to build/verify an e2e guard, so I left both out rather than guess. Happy to take a pass at the e2e guard if someone can point me at how `mac-update-e2e.yml` is normally exercised.

## Testing

- `npx tsc --noEmit` over the frontend: clean (after `npm ci` to pick up the current lockfile).
- `npx vitest run src/main/feature-builds.test.ts src/main/auto-updater.test.ts`: 61/61 passing, no regressions.
- `.github/workflows/build-artifacts.yml` parsed with `js-yaml` to confirm the edited step is still valid YAML.
- Grepped the touched files for any remaining `AgentWrapper` reference; only the intentional historical note in the `forge.config.ts` comment remains.

I don't have a way to actually run the Electron packaging step or a macOS/Linux release build in this environment, so I wasn't able to verify the baked `app-update.yml` output end-to-end -- the fix is a straightforward default-value substitution plus an explicit env var mirroring an existing working pattern, but flagging that gap for reviewers.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -78,6 +78,11 @@ jobs:
           node -e "const v=process.env.BUILD_VERSION.split('+')[0]; const fs=require('fs'); const p=require('./package.json'); p.version=v; fs.writeFileSync('./package.json', JSON.stringify(p,null,'\t')+'\n');"
       - name: Build (unsigned, no publish)
         shell: bash
+        env:
+          # forge.config.ts bakes app-update.yml's owner/repo from this at make
+          # time even though this workflow never publishes; unset, it falls back
+          # to forge.config.ts's production default (#3523).
+          AO_RELEASE_REPO: ${{ github.repository }}
         run: npm run make
       - name: Collect artifact and compute digest
         shell: bash
--- a/frontend/forge.config.ts
+++ b/frontend/forge.config.ts
@@ -6,8 +6,9 @@ import MakerAppImage from "./makers/maker-appimage";
 import { writeFileSync } from "node:fs";
 
 // Default GitHub release target (production). aoagents was the temporary rewrite
-// home; releases land on AgentWrapper (spec §1.1).
-const DEFAULT_RELEASE_REPO = "AgentWrapper/agent-orchestrator";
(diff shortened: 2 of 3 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

krishrathi1:
@Vaibhaav-Tiwari  please review my changes

Rishet11:
@krishrathi1 
All three target states are already on main via #3623: the release repo in forge.config, AO_RELEASE_REPO in the build workflow, and the default repo in feature builds. 
Closing as superseded.

PR #3697, opened 2026-08-07
fix: keep sidebar Projects section always expanded

Annieeeee11:
## Summary
- Remove collapse toggle and chevron from the sidebar **Projects** section header so it stays always expanded
- Leave **Pinned** and per-project disclosure unchanged

## Test plan
- [ ] Open the sidebar and confirm the Projects header has no chevron and is not clickable to collapse
- [ ] Confirm the project list stays visible
- [ ] Confirm individual projects can still expand/collapse
- [ ] Confirm Pinned still expands/collapses when sessions are pinned
- [ ] Confirm the Projects `+` button still works

## Before
image

## After
image


ronishrohan (approved):
Everything seems right, basic removal and padding adjustment, no real issues

PR #3700, opened 2026-08-07
fix(desktop): detect wedged updater network stack and nudge restart

### LINKED ISSUE #3526: Updater surfaces persistent net::ERR_FAILED from a wedged Chromium network stack in long-running instances (until app restart)
## Bug

Split out of #3523 per Stage 1 findings there. The `net::ERR_FAILED` on Check for Updates is **not** caused by the GitHub org-rename redirect chain (refuted by reproduction — see #3523), but it is a real UX papercut: when Electron's Chromium network stack wedges in a long-running app instance, every subsequent updater request fails with `net::ERR_FAILED` until the app is restarted, and the Settings UI surfaces the raw string with no guidance.

## Evidence

- The maintainer's instance (up since Aug 2 ~10:09 after a ShipIt install) failed every check with `net::ERR_FAILED`, while on the SAME machine at the same time: (a) a fresh electron-updater harness (identical Electron 33.4.11 + electron-updater 6.8.9) and (b) a fresh isolated-HOME launch of the SAME installed binary both checked and downloaded through the full redirect chain without error.
- The instance's updater cache (`~/Library/Caches/agent-orchestrator-updater/pending`) held no Aug-3 nightly despite two being published and hourly automatic checks — consistent with every silent automatic check failing all day in that process.
- Automatic-check failures are deliberately suppressed from the UI (only telemetry), so a wedged instance goes silently stale until the user checks manually.

## Proposed hardening

1. **Detect and message:** when a manual check fails with a net-level error (`message` matching `/^net::/`), show a friendlier message that suggests restarting the app instead of the raw Chromium string.
2. **Escalate silent staleness:** track consecutive automatic-check failures with net-level errors; after N consecutive failures (e.g. 3+ hours), surface a non-error nudge ("Updates haven't been able to check for a while — restarting the app usually fixes this").
3. **(Investigate)** whether the wedge is Electron's network-service crash without restart (electron/electron#32133-family); if so, a main-process listener on `net` failures could attempt session recreation, or an Electron upgrade may fix it upstream.
4. PostHog already receives `updateFailureOutcome` telemetry — query error messages by instance age to size how common the wedge is before investing in (3).

## Verification pending

Maintainer to confirm: after restarting the desktop app, Check for Updates succeeds on the same install that showed `net::ERR_FAILED` (predicted by the Stage 1 analysis).

anmol0b:
## What

Hardens the desktop auto-updater against a wedged Chromium network stack (net::ERR_FAILED on every updater request until app restart):

1. Manual "Check for Updates" failures with a Chromium network error (net::ERR_*) now show restart guidance instead of the raw Chromium string.
2. Consecutive automatic-check failures with network-level errors are counted; after 3 in a row (~3h of hourly checks), the Settings → Updates section shows a non-error warning nudge: "Updates haven't been able to check for a while — restarting the app usually fixes this."

## Why

Fixes #3526.

A long-running app instance can end up with a wedged Chromium network stack. Every updater request then fails with net::ERR_FAILED until the app is restarted. Two UX problems resulted:

- Manual checks surfaced the raw net::ERR_FAILED … string with no guidance.
- Automatic (hourly) check failures are deliberately UI-suppressed (telemetry only), so a wedged instance went silently stale — the maintainer's instance missed two published nightlies with no visible signal.

## How

- `frontend/src/shared/update-telemetry.ts`: new exported `isNetErrorMessage()` (`/^net::/i`, anchored) — the single definition of the wedge signature. Telemetry bucketing already categorized these as "network"; unchanged.
- `frontend/src/main/auto-updater.ts`:
  - `updateRequestErrorMessage()` swaps the raw `net::` string for restart guidance (original message kept in parentheses for diagnostics), applied in both manual-failure paths: the `checkForUpdatesNow` rejection catch and the error-event fall-through in `wireUpdaterEvents`. Hardcoded English in main, matching the existing manifest-404 message convention.
  - A module-level streak counter increments on net-level automatic-check failures — in the error handler AND in `runAutomaticUpdateCheck`'s own catch (a wedged stack can reject `checkForUpdates()` without emitting an error event). A per-operation guard prevents double counting when one check surfaces both ways. `update-available` / `update-not-available` / `download-progress` reset the streak, since any of them proves the network works.
  - `broadcast()` (the single status choke point) stamps `staleCheckNudge: true` on outgoing statuses once the streak reaches `STALE_CHECK_NUDGE_THRESHOLD = 3`. The nudge rides on the normal status object, so it never clobbers state and the automatic-failure UI suppression stays intact.
- `frontend/src/main/update-settings.ts`: `UpdateStatus.staleCheckNudge?: boolean`.
- Renderer: `UpdatesSection.tsx` renders the nudge as a text-warning line (same pattern as the nightly warning), independent of status state; new i18n key `settings.updates.networkStale` added to all 8 locale files.

Intentional omissions:

- Issue item (3) — investigating whether the wedge is an Electron network-service crash, attempting session recreation, and sizing via PostHog — is investigation-only and out of scope for this PR.
- No automatic remediation (e.g. session recreation); this PR is detect + message only.

## Testing

- `npx vitest run src/shared/update-telemetry.test.ts src/main/auto-updater.test.ts src/renderer/components/GlobalSettingsForm.test.tsx src/renderer/i18n/renderer-coverage.test.ts` — 76/76 pass, including new cases:
  - manual check rejecting with net::ERR_FAILED → friendly restart message (raw string preserved in parentheses), both rejection and error-event paths
  - non-net errors stay verbatim
  - 3 consecutive net-failed automatic checks → `staleCheckNudge: true` on the restored status; 2 → no nudge
  - non-net automatic failures never nudge
  - one check emitting error AND rejecting counts as a single failure
  - a successful check clears the nudge
  - renderer: warning line renders with the normal status still shown
- `npm run typecheck` — clean for all touched files (only pre-existing lowlight/highlight.js resolution errors in `code-highlight-engine.ts` from missing node_modules in the local checkout; unrelated).

Ui changes (before): 

before-settings-page
before-updates

Ui changes (after):

after-error-updates
after-error-settings-page

## Checklist


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Annieeeee11:
## Review note

Looks good overall — manual `net::` guidance and the error-event automatic streak/nudge path are solid.

One gap on the reject-only path this PR explicitly adds:

In `runAutomaticUpdateCheck`'s catch, we only do:

```ts
if (isNetError(err)) recordAutomaticNetFailure();
```

`staleCheckNudge` is stamped only inside `broadcast()`, and `getUpdateStatus()` returns cached `lastStatus`. On reject-without-`error`-event:

1. streak increments
2. `finally` clears `automaticCheckPreviousStatus` without restore
3. no rebroadcast → Settings never gets `staleCheckNudge: true`

The error-event path works because it calls `restoreAutomaticCheckPreviousStatus()` after recording. Existing tests cover error-event + both-fire, but not reject-only.

**Ask:** after `recordAutomaticNetFailure()` in that catch, restore/rebroadcast (or derive the stamp in `getUpdateStatus()`), and add a reject-only test that asserts the nudge with no `error` event.

anmol0b:
Fixed , the rejection is now caught inside the serialized operation while automaticCheckPreviousStatus is intact; it records, then restores via broadcast() so the nudge is stamped and pushed live (also un-stucks the checking status on this path). Added a reject-only test asserting restore + nudge at threshold; verified it fails without the fix. @Annieeeee11

frontend/src/main/auto-updater.ts
```diff
@@ -138,21 +149,25 @@ function broadcast(
   status: UpdateStatus,
   owner: "independent" | "automatic-operation" = "independent",
 ): void {
+  const stamped: UpdateStatus =
+    consecutiveAutomaticNetFailures >= STALE_CHECK_NUDGE_THRESHOLD
+      ? { ...status, staleCheckNudge: true }
```
Annieeeee11, line 154: **Residual risk 1 — nudge depends on a restore/broadcast**

`staleCheckNudge` is only stamped here inside `broadcast()`, and `getUpdateStatus()` returns cached `lastStatus` unchanged.

If a net-level automatic failure increments the streak *without* a prior `checking-for-update` (so `automaticCheckPreviousStatus` is undefined and restore is a no-op), the counter can cross the threshold while `lastStatus` stays unstamped until some later broadcast. Opening Settings would then call `getUpdateStatus()` and miss the nudge.

Unlikely with electron-updater (it usually emits `checking` first), but deriving the stamp in `getUpdateStatus()` (the alternate from the earlier review note) would close this edge case permanently:

```ts
export function getUpdateStatus(): UpdateStatus {
  return consecutiveAutomaticNetFailures >= STALE_CHECK_NUDGE_THRESHOLD
    ? { ...lastStatus, staleCheckNudge: true }
    : lastStatus;
}
```

frontend/src/main/auto-updater.ts
```diff
@@ -537,11 +596,26 @@ async function runAutomaticUpdateCheck(stateDir: string): Promise<boolean> {
...
+        restoreAutomaticCheckPreviousStatus();
+        throw err;
+      }
     });
   } catch (err) {
     console.error("auto-update check failed:", err);
+    // A wedged network stack can reject checkForUpdates() without emitting an
+    // "error" event; count that path too (the per-operation guard prevents
+    // double counting when both fire).
+    if (isNetError(err)) recordAutomaticNetFailure();
```
Annieeeee11: **Residual risk 2 — outer `recordAutomaticNetFailure()` is redundant for the reject path**

The inner catch (inside the serialized op) already records + restores before rethrowing. By the time this outer catch runs, `automaticCheckNetFailureCounted` is set, so this call is a no-op for `checkForUpdates()` rejects.

Harmless, but the comment (“count that path too”) is slightly misleading now — the reject-only path is handled above. Consider dropping the outer `recordAutomaticNetFailure()` (keep the `console.error`) to avoid future readers thinking this is the primary counter.

frontend/src/renderer/components/settings/UpdatesSection.tsx
```diff
@@ -245,6 +245,13 @@ export function UpdatesSection({ titleHidden }: { titleHidden?: boolean } = {})
 					</p>
 				)}
 
+				{status.staleCheckNudge && (
```
Annieeeee11: **Residual risk 3 — nudge is Settings-only**

The automatic-staleness signal only appears in Settings → Updates. Users who never open that section still go silently stale on the hourly path; they only get guidance if they manually hit “Check for Updates” (which now shows the restart message).

Intentional for this PR’s detect+message scope, but worth noting vs. the original “install went silently stale” failure mode in #3526 — a sidebar/toast signal would reach more wedged instances. Fine to defer.

Annieeeee11 (commented):
Follow-up residual risks from review (non-blocking). The reject-only restore fix looks good; these are edge cases / intentional tradeoffs worth tracking.

frontend/src/main/auto-updater.ts
anmol0b: accepted and fixed. getUpdateStatus() now derives the stamp at read time (auto-updater.ts:574-581), so even if the streak crosses 3 without any broadcast, Settings still gets staleCheckNudge: true when it seeds. New test "stamps the nudge on getUpdateStatus even when no broadcast carried it" proves it, zero status broadcasts, getter still stamped. Verified it fails without the fix; all 45 tests green, typecheck clean.

frontend/src/main/auto-updater.ts
anmol0b: fixed. dropped the redundant outer recordAutomaticNetFailure() and its now-misleading comment; the outer catch is back to log-only. the inner catch + per-operation guard cover every checkForUpdates() rejection, and pre-check rejections (settings read/reconcile) can't produce net:: errors anyway.

Annieeeee11 (approved):
After changes lgtm. The error ui is bad but that will be fixed in a follow up pr.

frontend/src/main/auto-updater.ts
```diff
@@ -473,6 +529,7 @@ function wireUpdaterEvents(): void {
     emitUpdateFailure(err);
     if (activeUpdaterOperation === "automatic-check") {
       console.error("auto-update check failed:", err);
+      if (isNetError(err)) recordAutomaticNetFailure();
```
codebanditssss: The streak is intended to represent **consecutive** `net::` automatic-check failures, but this branch only increments for a net error and never clears the existing streak for a non-net error. For example, `net::ERR_FAILED`, `net::ERR_FAILED`, `HttpError: 500`, then `net::ERR_FAILED` will show the nudge even though the three net failures were not consecutive (and an HTTP response demonstrates that the Chromium network stack is responding). Please reset `consecutiveAutomaticNetFailures` on the non-net path and add an interleaving test such as net/net/non-net/net → no nudge.

codebanditssss (commented):
One behavioral issue remains before I’d approve: the counter is described and tested as a consecutive network-error streak, but non-network automatic failures do not break it. The current CI checks are green and the earlier reject-only/broadcast issues appear fixed.

frontend/src/main/auto-updater.ts
anmol0b: fixed. non-net automatic failures now reset the streak via recordAutomaticCheckFailure, since an HTTP response means the stack isn't wedged. added the net → net → non-net → net → no nudge case as a test.

codebanditssss (approved):
Re-reviewed the latest changes. The consecutive net:: failure streak now resets on non-network automatic-check failures, and the new net → net → non-net → net regression test covers the reported case. Existing duplicate event/rejection handling remains covered, all current CI checks are green, and I found no remaining blockers.

Annieeeee11:
## Review —LGTM, merge-worthy

Reviewed the full 4-commit diff, inline comments, and surrounding code. The two existing approvals are earned; the iteration history shows each real gap was addressed with a matching test rather than hand-waved.

### Strengths

- **Single source of truth** for the wedge signature: `isNetErrorMessage` (`/^net::/i`, anchored) in `update-telemetry.ts`, reused by both the manual message path and the automatic streak logic. Anchoring is correct — Chromium surfaces these at the start of the message, and the test asserts a non-anchored substring is *not* matched.
- **Double-count guard.** One automatic check can both emit `error` and reject `checkForUpdates()`. `automaticCheckNetFailureCounted` is re-armed per operation in `runSerializedUpdaterOperation` and checked in `recordAutomaticNetFailure`, so a double-fire counts once. Tested.
- **Read-time derivation** (commit 3) closes the no-broadcast edge case cleanly: `getUpdateStatus()` returns `{ ...lastStatus, staleCheckNudge: true }` when the streak is at threshold, so Settings seeds the nudge even if no `checking-for-update` ever fired (restore no-ops, no broadcast). Right call over relying solely on `broadcast()` stamping.
- **Streak resets are comprehensive:** on `update-available`, `update-not-available`, `download-progress`, and — after the final review round — on non-net automatic failures via `recordAutomaticCheckFailure` (an HTTP response proves the stack isn't wedged). The `net → net → non-net → net → no nudge` regression test locks this in.
- **Reject-only path restored** (commit 2): `runAutomaticUpdateCheck`'s inner catch records then `restoreAutomaticCheckPreviousStatus()` before rethrowing, so the renderer neither sticks on `checking` nor misses the nudge. Redundant outer counter dropped (commit 3).
- **Test coverage is genuinely thorough** for the boundaries changed: manual net (reject + error-event), manual non-net verbatim, threshold reach, non-net no-nudge, interleaving, double-fire single-count, reject-only, no-broadcast getter stamp, clear-on-success.
- **i18n:** `settings.updates.networkStale` added to all 8 locale files; renderer-coverage test guards drift.
- **Surgical:** no drive-by changes, respects the existing module-level-state pattern in `auto-updater.ts`, follows AGENTS.md PR hygiene.

### Observations / nits (non-blocking)

1. **Nudge is Settings-only.** A user who never opens Settings → Updates still goes silently stale on the hourly path; they only get guidance if they manually click "Check for Updates." The PR body explicitly scopes this to "detect + message," so deferring a sidebar/toast signal is a reasonable tradeoff, but worth tracking as a follow-up since the original #3526 failure mode was exactly "silently stale."
2. **Hardcoded English in main process.** The restart-guidance string lives in `auto-updater.ts`, not the i18n catalog. Matches the existing manifest-404 message convention (called out in the PR body), but the manual `net::` message won't localize. Acceptable given precedent.
3. **Minor stylistic inconsistency in the renderer.** The new nudge uses `px-1` while the adjacent nightly warning uses `px-(--size-settings-row-padding) pb-(--size-settings-row-padding)`. Same `text-warning`/`leading-row`/`AlertTriangle` palette, so it reads as a sibling, but the horizontal padding differs. Not worth a respin.
4. **`updateRequestErrorMessage` null/undefined branch.** Maps `err === undefined || null` → `"Update check failed"` before checking `isNetError(err)`, which would be false for null/undefined anyway. Harmless; slightly more branching than strictly needed.
5. **`getUpdateStatus()` and `broadcast()` both stamp.** Once the streak crosses threshold, `broadcast()` stores a stamped `lastStatus` and the getter re-stamps on read. Idempotent (`staleCheckNudge: true` overwrites itself), so no bug — just a slight redundancy that's worth it for the no-broadcast case.

### Cross-check against hard rules

No daemon/API/CLI or `~/.ao` state-path concerns — pure Electron main + renderer. No migrations, no OpenAPI drift, no generated files touched. `frontend/src/main.ts` `userData` override untouched. Clean on the boundaries AGENTS.md cares about.

### Recommendation

Merge.

anmol0b:
> ## Review —LGTM, merge-worthy
> Reviewed the full 4-commit diff, inline comments, and surrounding code. The two existing approvals are earned; the iteration history shows each real gap was addressed with a matching test rather than hand-waved.
> 
> ### Strengths
> * **Single source of truth** for the wedge signature: `isNetErrorMessage` (`/^net::/i`, anchored) in `update-telemetry.ts`, reused by both the manual message path and the automatic streak logic. Anchoring is correct — Chromium surfaces these at the start of the message, and the test asserts a non-anchored substring is _not_ matched.
> * **Double-count guard.** One automatic check can both emit `error` and reject `checkForUpdates()`. `automaticCheckNetFailureCounted` is re-armed per operation in `runSerializedUpdaterOperation` and checked in `recordAutomaticNetFailure`, so a double-fire counts once. Tested.
> * **Read-time derivation** (commit 3) closes the no-broadcast edge case cleanly: `getUpdateStatus()` returns `{ ...lastStatus, staleCheckNudge: true }` when the streak is at threshold, so Settings seeds the nudge even if no `checking-for-update` ever fired (restore no-ops, no broadcast). Right call over relying solely on `broadcast()` stamping.
> * **Streak resets are comprehensive:** on `update-available`, `update-not-available`, `download-progress`, and — after the final review round — on non-net automatic failures via `recordAutomaticCheckFailure` (an HTTP response proves the stack isn't wedged). The `net → net → non-net → net → no nudge` regression test locks this in.
> * **Reject-only path restored** (commit 2): `runAutomaticUpdateCheck`'s inner catch records then `restoreAutomaticCheckPreviousStatus()` before rethrowing, so the renderer neither sticks on `checking` nor misses the nudge. Redundant outer counter dropped (commit 3).
> * **Test coverage is genuinely thorough** for the boundaries changed: manual net (reject + error-event), manual non-net verbatim, threshold reach, non-net no-nudge, interleaving, double-fire single-count, reject-only, no-broadcast getter stamp, clear-on-success.
> * **i18n:** `settings.updates.networkStale` added to all 8 locale files; renderer-coverage test guards drift.
> * **Surgical:** no drive-by changes, respects the existing module-level-state pattern in `auto-updater.ts`, follows AGENTS.md PR hygiene.
> 
> ### Observations / nits (non-blocking)
> 1. **Nudge is Settings-only.** A user who never opens Settings → Updates still goes silently stale on the hourly path; they only get guidance if they manually click "Check for Updates." The PR body explicitly scopes this to "detect + message," so deferring a sidebar/toast signal is a reasonable tradeoff, but worth tracking as a follow-up since the original [Updater surfaces persistent net::ERR_FAILED from a wedged Chromium network stack in long-running instances (until app restart) #3526](https://github.com/Untrivial-ai/agent-orchestrator/issues/3526) failure mode was exactly "silently stale."
> 2. **Hardcoded English in main process.** The restart-guidance string lives in `auto-updater.ts`, not the i18n catalog. Matches the existing manifest-404 message convention (called out in the PR body), but the manual `net::` message won't localize. Acceptable given precedent.
> 3. **Minor stylistic inconsistency in the renderer.** The new nudge uses `px-1` while the adjacent nightly warning uses `px-(--size-settings-row-padding) pb-(--size-settings-row-padding)`. Same `text-warning`/`leading-row`/`AlertTriangle` palette, so it reads as a sibling, but the horizontal padding differs. Not worth a respin.
> 4. **`updateRequestErrorMessage` null/undefined branch.** Maps `err === undefined || null` → `"Update check failed"` before checking `isNetError(err)`, which would be false for null/undefined anyway. Harmless; slightly more branching than strictly needed.
> 5. **`getUpdateStatus()` and `broadcast()` both stamp.** Once the streak crosses threshold, `broadcast()` stores a stamped `lastStatus` and the getter re-stamps on read. Idempotent (`staleCheckNudge: true` overwrites itself), so no bug — just a slight redundancy that's worth it for the no-broadcast case.
> 
> ### Cross-check against hard rules
> No daemon/API/CLI or `~/.ao` state-path concerns — pure Electron main + renderer. No migrations, no OpenAPI drift, no generated files touched. `frontend/src/main.ts` `userData` override untouched. Clean on the boundaries AGENTS.md cares about.
> 
> ### Recommendation
> Merge.

Thanks, nits addressed. Net-error restart guidance now localizes via a netError flag on UpdateStatus + a new i18n key (all 8 locales), with the raw net:: message appended; the main-process branch also simplified. The three Settings hint paragraphs now use the --size-settings-row-padding token to match the nightly warning. The double-stamp is left as-is (it's the read-time nudge path). Tracked the Settings-only surfacing as a follow-up: #3968.

Annieeeee11 (approved):
lgtm

PR #3701, opened 2026-08-07
fix: restore model menu scrolling

illegalcall:
## Summary
- restore mouse-wheel and trackpad scrolling in long model catalogs
- keep the scrollbar and gutter visually hidden
- add a Chromium regression test that verifies real scroll geometry and wheel movement

## Root cause
The list used percentage height beneath a max-height-only flex parent. It expanded to the full catalog height, then the parent clipped it, leaving no scrollable overflow. A zero-min grid track now constrains the list to the available menu space.

## Verification
- frontend typecheck
- E2E typecheck
- 154 Vitest files / 2,021 tests
- 21 renderer-smoke Playwright tests
- focused wheel-scroll test repeated three times
- native Electron verification with the real OpenCode catalog and real AO data

### DIFF
--- a/frontend/e2e/model-menu-scroll.spec.ts
+++ b/frontend/e2e/model-menu-scroll.spec.ts
@@ -1,0 +1,72 @@
+import { expect, test } from "@playwright/test";
+import { installFakeAgent } from "./support/fake-bridge";
+
+const projectId = "model-scroll-proj";
+const models = Array.from({ length: 30 }, (_, index) => ({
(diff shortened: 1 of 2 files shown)


ronishrohan (approved):
Reviewed the focused layout fix and regression coverage. The zero-min grid track correctly preserves scroll overflow, and the Playwright test verifies both overflow geometry and wheel movement. Required CI checks are passing.

PR #3702, opened 2026-08-07
fix: sync Windows titlebar overlay with theme

JAYATIAHUJA:
## What

Windows now draws its own min, max, and close buttons in the app titlebar.

## Why

Native overlay colors did not follow every theme properly.

## Scope

Windows only. Linux and macOS titlebar behavior stays same.
Shared center panel spacing stays 24/16/14. Windows gets the 8px inset locally.

## Test

- `npm --prefix frontend run test -- WindowTitlebar.test.tsx`
- checked real Electron on Windows with isolated dev data on port 3002


harshitsinghbhandari:
PR body still describes the old window:setOverlay approach; please update it to match the shipped custom CSS controls.
Also confirm the inset token change (16/14px to 8px) hitting Linux too is intentional.

JAYATIAHUJA:
Updated this now.

The old `window:setOverlay` path is gone. Windows uses custom CSS controls now.

The 8px inset is Windows only. Linux keeps the shared 16/14 spacing.

Vaibhaav-Tiwari (changes requested):
This PR is old and out of sync with the current codebase. Someone else on the team already made a
  different change to the same Windows titlebar file, and as part of that, they deleted some text labels (translations)
  that this PR still depends on. So if this PR gets merged as-is, the new minimize/maximize/close buttons will show
  broken text like titlebar.minimize instead of "Minimize" — I actually tested this and confirmed it breaks. This needs
  to be fixed before merging.

  Smaller issue: The old Windows buttons used to visually change (icon flips) when you maximize the window, showing a
  "restore" icon instead of "maximize". The new custom buttons don't do that anymore — same icon always. Minor UX
  downgrade, not a big deal.

  Cleanup stuff (not urgent): A tiny bit of leftover unused code, and one comment in the CSS that no longer matches what
  the code actually does.

  Everything else checked out fine — no security issues, the buttons are wired up correctly, the new tests pass on their
  own.

Vaibhaav-Tiwari (approved):

Issue #3360, opened 2026-07-30
Add immutable Chat/TUI session modes and structured orchestrator Chat


illegalcall:
## Problem

AO currently has one interaction model for a running agent session: launch the provider's native CLI/TUI inside the selected terminal runtime, then deliver later messages by pasting text and Enter into that pane.

That model is valuable for broad agent compatibility, native provider workflows, and terminal escape hatches, but it is not a reliable substrate for a first-class chat product:

- a successful pane write is not an agent delivery acknowledgement;
- terminal bytes do not provide durable, typed conversation history;
- tool calls, diffs, plans, permissions, streaming deltas, usage, and errors must be inferred from terminal output or separate hooks;
- a keystroke intended as a prompt can land in a permission dialog or shell;
- a Chat UI backed by an SDK while a native TUI remains writable would create two independent conversation controllers that can diverge.

The unsafe design is:

```text
Chat UI -> Claude SDK / Codex app-server -> provider conversation A
Terminal -> native Claude/Codex TUI      -> provider conversation B (or a racing client on A)
```

Even if both clients start with the same provider conversation ID, concurrent writers can race on turns, permissions, compaction, model/mode changes, and reconnect state.

AO should instead let the user choose a **default session interface**:

- **Chat**: AO starts and owns a structured provider controller (Claude Agent SDK, Codex app-server, ACP, or another supported machine protocol). Chat is the default session surface. The Terminal option opens a plain shell in the same worktree, not a second native agent TUI.
- **Terminal UI (TUI)**: AO behaves as it does today. The provider's native TUI is the only conversation controller and the session opens in the existing terminal experience.

This must be a durable, immutable per-session controller mode, not merely a preferred tab. The global preference selects the mode for newly created sessions; every later spawn, send, restore, activity, permission, and UI decision uses the mode stored on the session.

## User outcome

As an AO user, I can choose **Chat** or **Terminal UI** as my default session interface.

When I choose Chat:

- newly created orchestrators open in a first-class Chat UI;
- workers spawned later use Chat when their selected harness has a production-ready structured driver;
- desktop, mobile, `ao send`, browser annotations, orchestrator coordination, and automated reactions all feed the same AO-owned conversation controller;
- I can open a normal shell in the session worktree without starting a second copy of the agent;
- messages, streaming output, tool calls, permissions, diffs, plans/todos, errors, and usage are represented as structured events;
- daemon/app restarts resume the same provider conversation.

[...]

PR #3703, opened 2026-08-07
fix(opencode): enable Chat and TUI handoff

illegalcall:
Refs #3360

## Summary

- opt the OpenCode adapter into the existing interface-handoff capability by mapping its TUI session ID and ACP provider conversation ID to the same native conversation
- pass the AO-owned system prompt file through terminal-target preflight so OpenCode can safely switch from Chat back to its TUI
- add regression coverage for both the provider identity bridge and prompt-file preflight, and update architecture/status documentation

No frontend or API contract changes are needed because the existing UI and daemon flow are capability-driven.

## Root cause

OpenCode already had a production Chat driver, but its agent adapter did not implement the explicit handoff capability, so the daemon rejected TUI-to-Chat switches. Enabling that path exposed a second preflight omission: terminal-target preflight passed inline system prompt text but not the system prompt file required to build OpenCode configuration.

## Verification

- `go test ./internal/adapters/agent/opencode ./internal/adapters/chatdriver/opencodeacp ./internal/session_manager ./internal/httpd/controllers -count=1`
- `npm run lint` (full Go suite and golangci-lint, 0 issues)
- isolated Electron exercise: OpenCode TUI -> Chat -> TUI -> Chat, preserving the original prompt and `TUI_READY` response across transitions


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/agent/opencode/opencode_test.go
```diff
@@ -13,9 +13,31 @@ import (
...
 )
 
+func TestNativeConversationIDUsesTheSameOpenCodeSessionAcrossInterfaces(t *testing.T) {
+	p := &Plugin{}
+	if id, ok, err := p.NativeConversationID(context.Background(), ports.SessionRef{
+		ID: "ao-session-1", Metadata: map[string]string{},
+	}, domain.SessionModeTUI, ""); err != nil || ok || id != "" {
+		t.Fatalf("uncaptured TUI native id = %q ok=%v err=%v", id, ok, err)
+	}
+	tuiID, ok, err := p.NativeConversationID(context.Background(), ports.SessionRef{
```
whoisasx, line 27: [P2] The test currently feeds the TUI ID directly back into NativeConversationID as the Chat providerConversationID, so it only verifies string forwarding. It cannot catch a mismatch between the ID emitted by OpenCode ACP and the ID captured by the TUI plugin, which is the key contract this PR relies on. Please add an integration or opt-in live regression test that obtains ProviderConversationID from the ACP driver and verifies that the same ID is used successfully by the --session restore path.

whoisasx (changes requested):
Changes requested: the implementation is directionally correct and the upstream OpenCode contract supports a shared session ID, but the added test does not exercise the provider boundary. It passes the TUI ID directly back into NativeConversationID as the Chat provider ID, so it cannot detect a mismatch between the ID emitted by the ACP driver and the ID captured by the TUI plugin. Please add an integration or opt-in live regression test that starts an OpenCode ACP session, captures ProviderConversationID, and verifies that the same ID is accepted by the TUI --session restore path. No other functional issues found in this review.

backend/internal/adapters/agent/opencode/opencode.go
```diff
@@ -173,6 +175,26 @@ func (p *Plugin) SessionInfo(ctx context.Context, session ports.SessionRef) (por
...
+func (p *Plugin) NativeConversationID(
+	ctx context.Context,
+	session ports.SessionRef,
+	currentMode domain.SessionMode,
+	providerConversationID string,
+) (string, bool, error) {
+	if err := ctx.Err(); err != nil {
+		return "", false, err
+	}
+	if currentMode == domain.SessionModeChat {
```
whoisasx, line 190: [P2] Preserve custom OpenCode configuration across the handoff

Opting OpenCode into this capability makes configuration parity part of the Chat/TUI handoff contract. Chat deliberately leaves the inherited `OPENCODE_CONFIG` path untouched and adds AO instructions through `OPENCODE_CONFIG_CONTENT`, but production TUI restore calls `opencodeConfigEnvPrefix`, which launches through `env OPENCODE_CONFIG=<AO prompt dir>/opencode.json` and replaces any user-supplied custom path. OpenCode documents that custom path as its own merged configuration source, so the restored TUI can lose custom provider, model, MCP, plugin, or permission settings after the Chat controller has stopped. The live test currently omits the production system prompt/file, so this override is not generated there. Please preserve or merge the inherited custom config and add a Chat-to-TUI regression with `OPENCODE_CONFIG` set. Reference: https://opencode.ai/docs/config/

whoisasx (changes requested):
The previously requested provider-boundary proof is now addressed: the opt-in live test obtains the ACP provider conversation ID and exercises the TUI `--session` restore path with it. The focused OpenCode, ACP, and session-manager tests pass, `git diff --check` is clean, and all eight GitHub checks are green.

One configuration-continuity issue remains. Chat preserves the user's custom `OPENCODE_CONFIG` and injects AO instructions through `OPENCODE_CONFIG_CONTENT`, while TUI restore replaces `OPENCODE_CONFIG` with AO's generated prompt config. That can change providers, models, MCP servers, plugins, or other custom settings during the handoff. The new live test does not exercise this production path because its restore request omits the system prompt file.

Requesting changes only for this remaining handoff-parity issue. Once custom OpenCode configuration is preserved and covered by a transition regression, this should be ready.

illegalcall:
Closing rather than merging this branch. OpenCode handoff is optional provider expansion, and this implementation still breaks configuration continuity: Chat preserves a custom OPENCODE_CONFIG and applies AO settings as an overlay, while TUI restore replaces it with an AO-generated config. That can silently change provider, model, MCP, plugin, or permission settings across a handoff. The branch also lacks the reverse TUI-created-ID to ACP load/resume proof required for a bidirectional capability. Please revisit under a dedicated OpenCode handoff issue with symmetric config preservation and both transition directions tested. This PR does not complete or close the broader [issue #3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360).

PR #3705, opened 2026-08-07
Polish the New Task composer and settings modals

ronishrohan:
## Summary
- Rebuild the New Task composer's agent/model pickers, dialog chrome, and dropdown styling to match the shared settings-modal system
- Fix a settings-navigation regression (section titles hidden in the all-sections view) and an untranslated i18n placeholder
- Various composer polish: consistent button heights, focus-ring cleanup, hover/divider/placeholder fixes

image
image
image
image


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/stores/ui-store.ts
```diff
@@ -61,8 +54,6 @@ type UiState = {
 	resolvedTheme: Theme;
 	/** Named color style theme (e.g. "catppuccin", "nord") — independent of light/dark mode. */
 	themeStyle: ThemeStyle;
-	/** When true, developer-only surfaces (e.g. Feature Releases) are revealed. Default off. */
-	developerMode: boolean;
```
Annieeeee11, line 65: **[P1] Removing Developer Mode also removes Feature Releases with no replacement**

Deleting `developerMode` (and the Developer settings tab / Feature Releases picker in `UpdatesSection` / `GlobalSettingsForm`) means users can no longer pin a new feature build from the UI. Existing pins only get a “return home” escape hatch.

This is a product regression bundled into a polish PR. Please either:
1. Get explicit product sign-off that Feature Releases is intentionally gone, or
2. Restore a path to select/confirm feature builds (Developer Mode or another gate).

Same change set also deletes `DevSettings` / fixture controls — call that out in the PR body if intentional.

frontend/src/renderer/components/settings/UpdatesSection.tsx
```diff
@@ -96,15 +85,8 @@ export function UpdatesSection({ titleHidden }: { titleHidden?: boolean } = {})
 	const channelOptions: { value: PrimaryValue; label: string }[] = [
 		{ value: "latest", label: t("settings.updates.channel.stable") },
 		{ value: "nightly", label: t("settings.updates.channel.nightly") },
-		{ value: "feature", label: t("settings.updates.channel.feature") },
 	];
```
Annieeeee11, line 88: **[P1] Feature Releases channel dropped here**

`channelOptions` is now only `latest` / `nightly`. Combined with removing Developer Mode, there is no UI left to enter the feature-build flow (picker + confirm dialog were deleted in this file).

If Feature Releases is staying, this needs a replacement entry point before merge.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -305,14 +305,12 @@ function SummaryView({
 }) {
 	const { t } = useTranslation();
 	const query = useSessionScmSummary(session.id);
-	const developerMode = useUiStore((state) => state.developerMode);
-	const usageQuery = useSessionUsage(session.id, developerMode);
+	const usageQuery = useSessionUsage(session.id, false);
```
Annieeeee11: **[P2] Usage & cost hard-disabled, but the UI surface remains**

```ts
const usageQuery = useSessionUsage(session.id, false);
```

With `enabled` always `false`, the query never runs, so `showUsage` / `showUsageError` never render — while `UsageCostTelemetry` and helpers below stay as dead code, and the Usage tests were deleted.

Please either:
- delete the unused Usage UI path entirely, or
- restore a real enablement path if telemetry is still meant to ship.

frontend/src/renderer/i18n/en.json
```diff
@@ -527,11 +528,11 @@
 	"settings.project.intakeAssigneeRequired": "Enabling intake requires an assignee.",
 	"settings.project.loadFailed": "Could not load project.",
 	"settings.project.loading": "Loading project settings…",
-	"settings.project.kind": "kind",
-	"settings.project.kind.scratch": "scratch",
-	"settings.project.kind.singleRepo": "single repo",
-	"settings.project.kind.unknown": "unknown",
-	"settings.project.kind.workspace": "workspace",
+	"settings.project.kind": "Type",
```
Annieeeee11, line 508: **[P2] Claimed `settings.project.kind` i18n fix is incomplete**

`en` is correctly updated from `"kind"` → `"Type"` (and the kind.* labels). `ja.json` / `ko.json` still ship `"settings.project.kind": "kind"` — they were not updated in this PR.

Same gap for the rewritten New Task copy (`newTask.letAgentChoose`, title, placeholder): only English changed; other locales still say the old “Let X choose” / “New task” wording.

Please sync `ja`/`ko` at minimum for the `kind` placeholder this PR claims to fix, and ideally align all locales for the New Task copy changes.

frontend/src/renderer/components/settings/SettingsOptionMenu.tsx
```diff
@@ -71,21 +95,29 @@ export function SettingsOptionMenu<T extends string>({
...
 			{/* bg-settings-menu / border-settings-menu / rounded-(--radius-settings-panel) must
 			    be real utilities so twMerge drops DropdownMenuContent's bg-popover, border-border,
 			    and rounded-lg. */}
 			<DropdownMenuContent
 				align={menuAlign}
+				alignOffset={0}
+				// Radix refocuses the trigger on close; Chromium's :focus-visible
+				// heuristic treats that async re-focus as keyboard-style even after a
+				// mouse click, leaving a stray focus ring. Skip the refocus.
+				onCloseAutoFocus={(event) => event.preventDefault()}
```
Annieeeee11: **[P2] `onCloseAutoFocus` preventDefault drops focus after every menu close**

Skipping Radix’s close autofocus fixes the mouse focus-ring glitch, but after keyboard selection / Escape, focus is no longer returned to the trigger. That hits every settings dropdown using this component (and the same pattern in `AgentModelCombobox`).

Prefer restoring focus without leaving a `:focus-visible` ring — e.g. only suppress for pointer closes, or blur/refocus in a way that doesn’t look like keyboard focus.

frontend/src/renderer/components/settings/AgentModelCombobox.tsx
```diff
@@ -156,22 +165,33 @@ export function AgentModelCombobox({
...
+						aria-hidden="true"
+					/>
 				</button>
 			</DropdownMenuTrigger>
 			<DropdownMenuContent
 				align={menuAlign}
+				// Radix refocuses the trigger on close; Chromium's :focus-visible
+				// heuristic treats that async re-focus as keyboard-style even after a
+				// mouse click, leaving a stray focus ring. Skip the refocus.
+				onCloseAutoFocus={(event) => event.preventDefault()}
```
Annieeeee11: **[P2] Same focus-restore issue here**

Mirror of the `SettingsOptionMenu` change: `onCloseAutoFocus` preventDefault also applies to the composer model menu. Please keep mouse/keyboard behavior consistent with a safer focus-restore approach (see comment on `SettingsOptionMenu`).

frontend/src/renderer/components/settings/AgentModelCombobox.tsx
```diff
@@ -67,14 +68,18 @@ export function AgentModelCombobox({
...
 	triggerLabel?: string;
 	triggerClassName?: string;
 	menuAlign?: "start" | "center" | "end";
 	renderTrigger?: (label: string) => ReactNode;
 	/** Persists explicit model choices for this agent and pins them below defaults. */
 	recentScope?: string;
+	/** Flat "no override" + plain model names — no search, groups, badges, or
+	 *  refresh action. For contexts where the menu should read like a simple
+	 *  choice, not a model-management surface. */
+	compact?: boolean;
```
Annieeeee11, line 85: **[P3] `compact` docs are stale**

Comment still says compact has “no search”, but `showSearch` is driven only by `models.length > MODEL_SEARCH_THRESHOLD` (line 95), including in compact mode. Please update the JSDoc so the next editor doesn’t reintroduce the old unconditional search disable.

Annieeeee11 (changes requested):
## Review summary

Composer/settings visual unification looks solid, but this PR also removes Developer Mode / Feature Releases and hard-disables Usage telemetry. Please address the inline findings before merge — especially the Feature Releases removal (needs explicit product sign-off or restore) and the incomplete i18n / dead Usage cleanup.

CI looks green; findings are defect/scope focused.

frontend/src/renderer/stores/ui-store.ts
ronishrohan: This predates this round of composer/settings polish work — the Developer Mode / Feature Releases removal is from commit 5edf33ce2 (`fix: remove developer settings tab and related UI state`), which was already on this branch before the changes in this PR started. It's a real product decision (restore the UI path vs. confirm the removal is intentional), so it's outside the scope of what I can resolve here — flagging it back to the project owner rather than silently restoring or dropping it.

frontend/src/renderer/components/settings/UpdatesSection.tsx
ronishrohan: Same root cause as the ui-store.ts thread — this is from commit 5edf33ce2, which predates this PR's actual changes. Flagging back to the project owner as a product decision rather than resolving it here.

frontend/src/renderer/components/SessionInspector.tsx
ronishrohan: Same root cause as the other two threads (commit 5edf33ce2, predates this PR's own changes). Flagging back to the project owner — restoring a real enablement path or deleting the dead UI path is a product/scope decision, not something I should resolve unilaterally in this polish PR.

frontend/src/renderer/i18n/en.json
ronishrohan: Fixed — synced ja/ko's settings.project.kind (and .scratch/.singleRepo/.unknown/.workspace) to proper translations, and also synced the reworded New Task title/no-override-label/placeholder across all non-English locales (de/es/fr/ja/ko/pt-BR/zh-CN), not just kind. Pushed in 1e7b499.

frontend/src/renderer/components/settings/SettingsOptionMenu.tsx
ronishrohan: Good catch, fixed. Instead of unconditionally preventing the refocus, I now only suppress it when the close was actually triggered by a pointer interaction (tracked via a document-level pointerdown listener while the menu is open) — keyboard closes (Escape, Enter on an item) get Radix's normal focus-restore back, so the ring shows correctly for keyboard users while the mouse-click stray ring stays fixed. Extracted into a shared useSuppressStrayFocusRing hook since AgentModelCombobox had the identical issue. Pushed in bf58145/1e7b499.

frontend/src/renderer/components/settings/AgentModelCombobox.tsx
ronishrohan: Fixed via the same shared useSuppressStrayFocusRing hook — see reply on the SettingsOptionMenu thread.

frontend/src/renderer/components/settings/AgentModelCombobox.tsx
ronishrohan: Fixed — updated the JSDoc to reflect that search still shows above MODEL_SEARCH_THRESHOLD in compact mode; only grouping/badges/refresh are stripped.

frontend/src/renderer/stores/ui-store.ts
Annieeeee11: Re-reviewed at `1e7b499`. Agree this originated in `5edf33c` on the branch rather than the later polish commits — but that commit is still in `main...HEAD`, so **Feature Releases still ships as unreachable UI if this PR merges**.

“Out of scope for me to unilaterally restore” is fair; what’s needed before merge is an explicit owner decision: confirm intentional removal in the PR body / approval, **or** restore a selection path. Leaving it unresolved means the regression lands with this merge.

frontend/src/renderer/components/settings/UpdatesSection.tsx
Annieeeee11: Still open on re-review (`1e7b499`). Same as the ui-store thread: `channelOptions` remains latest/nightly only, so there is still no UI entry point for Feature Releases in the merge diff. Needs owner sign-off or a restore — not closable as “predates polish” alone.

frontend/src/renderer/components/SessionInspector.tsx
Annieeeee11: Still open on re-review (`1e7b499`). `useSessionUsage(session.id, false)` is unchanged, so Usage & cost stays permanently gated off while the telemetry UI/helpers remain as dead code.

Same ask as before: owner decision to **delete** the dead path or **restore** a real enablement path before merge.

Annieeeee11 (dismissed):
## Re-review (`1e7b499`)

Thanks for the follow-ups, these look good and are **resolved**:
- i18n sync for `settings.project.kind` + New Task copy across locales
- keyboard focus-restore via `useSuppressStrayFocusRing`
- `compact` JSDoc accuracy

**Still blocking merge** (replied on the existing threads):
1. **[P1] Developer Mode / Feature Releases removal** — still in `main...HEAD` (`5edf33c`). Needs explicit owner sign-off or a restore; “predates polish” doesn’t remove it from the merge.
2. **[P2] Usage & cost hard-disabled with dead UI** — `useSessionUsage(session.id, false)` unchanged; delete the dead path or restore enablement.

frontend/src/renderer/components/SessionInspector.tsx
ronishrohan: Deleted the entire dead Usage & cost UI path per product decision — UsageCostTelemetry and its helpers (UsageProviderRow, ProviderUsageDetails, UsageModelRow, UsageDisclosureRow, UsageMetrics, formatHarnessName, etc.), the now-fully-orphaned useSessionUsage hook (no other consumers), and the 25 inspector.usage.* i18n keys across all locales that were only used by this path. Pushed in 03c9efc6c.

frontend/src/renderer/stores/ui-store.ts
ronishrohan: Confirmed with the project owner: the Developer Mode / Feature Releases removal (commit 5edf33c) is an intentional product decision, not an accidental regression. No restore planned — signing off on this as-is.

frontend/src/renderer/components/settings/UpdatesSection.tsx
ronishrohan: Same sign-off as the ui-store.ts thread — the Feature Releases channel removal is confirmed intentional by the project owner, not an oversight. No replacement entry point planned.

ronishrohan:
@Annieeeee11 all findings addressed — could you re-review when you get a chance?

- Usage & cost: deleted the entire dead UI path (`UsageCostTelemetry` + 8 helpers, the orphaned `useSessionUsage` hook, and 25 dead `inspector.usage.*` i18n keys across all locales) in 03c9efc6c
- Developer Mode / Feature Releases: confirmed with the project owner this removal is an intentional product decision, not an oversight — replied on both threads
- i18n sync, keyboard focus-restore, and the stale JSDoc were already fixed and confirmed resolved in your last re-review

Also merged main in again (78cffbf16) to pick up an unrelated conflict from #3709 — CI is green.

codebanditssss (approved):

PR #3706, opened 2026-08-07
fix(frontend): polish sidebar and settings presentation

ronishrohan:
## Summary
- Restore and polish sidebar presentation contracts (layout, spacing, test coverage)
- Remove sidebar focus chrome
- Add border/padding to settings link rows

image


frontend/src/renderer/components/GlobalSettingsForm.test.tsx
```diff
@@ -160,6 +160,17 @@ describe("GlobalSettingsForm", () => {
 		expect(screen.getByRole("button", { name: "Report a problem" })).toBeInTheDocument();
 	});
 
+	it("gives settings link rows internal padding and rounded borders", async () => {
+		renderForm();
+
+		const connectMobile = await screen.findByRole("button", { name: "Connect Mobile" });
+		const keyboardShortcuts = screen.getByRole("button", { name: "Keyboard shortcuts" });
```
Pulkit7070, line 167: This is why the `test` CI job is failing (`TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Keyboard Shortcuts"`). The row renders its label via `t("settings.keyboardShortcuts")`, and `en.json` defines that as "Keyboard shortcuts" (lowercase "s"), but this query asks for "Keyboard Shortcuts" (capital "S"). The `titlebar.shortcuts` key is the capitalized one, but that is used in WindowTitlebar, not this form. Change the query to `{ name: "Keyboard shortcuts" }` (or match by a case-insensitive regex) so it lines up with the rendered label.

frontend/src/renderer/styles.css
```diff
@@ -1478,6 +1484,12 @@ body.is-resizing-x [data-slot="sidebar-container"] {
 	transition: none;
 }
 
+/* Sidebar controls remain keyboard-focusable without painting rings or outlines. */
+.sidebar-focusless :focus-visible {
```
Pulkit7070, line 1488: This removes the visible keyboard-focus indicator for the whole sidebar: `.sidebar-focusless` is applied to the sidebar container, and the descendant `:focus-visible` here kills both `outline` and `box-shadow` (with `!important`) on every focusable control inside it. The comment notes the controls stay focusable, but a focusable control with no visible focus state is a WCAG 2.4.7 (Focus Visible) regression, keyboard and accessibility users can no longer tell what is focused as they tab through projects/sessions/actions.

If the goal is just to drop a heavy default ring, consider replacing it with a lighter but still-visible affordance (a subtle background or inset outline on `:focus-visible`) rather than removing focus indication entirely. Was fully removing it intentional?

ronishrohan: added bg instead of ring, pushed

Pulkit7070 (approved):
Approving. Both review points are resolved and verified: the failing test now queries "Keyboard shortcuts" (matching the rendered label) and the test job is green, and the sidebar focus removal now adds a background highlight on :focus-visible instead of stripping the indicator entirely, which keeps keyboard focus visible (WCAG 2.4.7). The settings-link-row border/radius utility and the sidebar action-width token change look correct.

Pulkit7070 (approved):

PR #3707, opened 2026-08-07
feat(mobile): make Connect Mobile pairing work over Tailscale, including iOS

Prasad-D-Ware:
## Problem

The Connect Mobile dialog has a LAN/Tailscale segmented control, but the toggle only swapped instruction text — the QR always carried the LAN address. A user who selected **Tailscale** and scanned paired their phone to an address it could not reach, and the resulting error told them to check they were on the same Wi-Fi: the exact opposite of what a Tailscale user is doing.

Fixing that surfaced a second, deeper problem. Even with the correct `100.x` address in the QR, **iOS still could not pair**. App Transport Security permits cleartext HTTP only to RFC 1918 ranges, and Tailscale's `100.64.0.0/10` is RFC 6598 — so iOS blocked the request before a packet left the device. Confirmed empirically: Safari (exempt from ATS) reached the address fine while the app reported "Reached nothing". Android was unaffected, since `usesCleartextTraffic` is global.

## What this does

**The QR now follows the mode.** LAN encodes the LAN address; Tailscale encodes the Tailscale address.

**A new opt-in "Secure pairing (TLS)" mode** solves iOS. The desktop runs `tailscale serve` to front the bridge with Tailscale's own certificate, and the QR carries the MagicDNS hostname on port 443 with `secure: true`. HTTPS to a cert-backed domain satisfies ATS outright — **no ATS exception, no `app.json` change, nothing to justify at App Store review** — and the traffic becomes genuinely encrypted rather than relying on WireGuard alone, so the "not encrypted" warning is correctly suppressed in that mode.

Two pre-existing bugs are fixed along the way:

- With no address available, the modal used to encode `host: ""`, which the app rejects as *"not an AO pairing code"* — an incoherent error for a QR the desktop generated itself. It now shows an explanatory hint instead.
- A scan overrode host/port/password but **inherited `secure`**, and the `v: 1` payload had no way to clear it. A user who enabled TLS manually and then scanned a LAN QR silently kept `secure: true`, hit the same misleading error, and stayed stuck across restarts because the flag is persisted.

## Compatibility

**The payload stays at `v: 1`, and `secure` is emitted only when true** — every plaintext QR is byte-identical, so no already-installed build regresses. Secure pairing is off by default, so an older build only ever sees a QR it understands.

## The one adoption caveat

Secure pairing needs **HTTPS certificates enabled on the tailnet** (admin console → DNS), which is not on by default, and the `tailscale` CLI on `PATH` (App Store installs of Tailscale don't add it). The feature detects both, plus a missing MagicDNS name and a failed proxy, and shows targeted setup steps rather than a QR that cannot connect. Where it isn't usable, it degrades to today's plaintext behaviour — which still works on Android.

## Design notes worth a reviewer's attention

- **The bridge has no fixed port.** `LANManager.Start` falls back to an ephemeral port when its default is taken (observed: bound `54014` because another instance held `3011`). A `serve` config pinned to a stale port is this feature's primary failure mode, and `serve` config persists inside `tailscaled` across restarts. `Apply` is therefore called with the port `Start` *actually returned*, on every listener start — enable, password rotation, and boot restore. Boot restore routes through `BridgeService` for this reason rather than driving the listener directly.
- **A `serve` failure never takes the bridge down.** It is recorded and surfaced as a reason while the bridge stays up in plaintext mode.
- **`reason` is a fixed enum**, never a raw error string — raw errors are untranslated and can leak filesystem paths to the UI.
- **`serve`, never `funnel`** — the proxy is tailnet-only; nothing is publicly exposed. `/api/v1/mobile` remains 404 on the LAN listener by prefix, so the new route is unreachable from the network with no extra work, and the pairing password never leaves loopback.

## Testing

Automated: backend packages pass with `go vet`/`gofmt` clean; renderer 1448/1448; mobile 349/349; `tsc --noEmit` clean in both TS packages.

Verified on real hardware (macOS + iPhone 13 + Android):

- Secure scan on iPhone **over cellular with Wi-Fi off** — pairs with no typing, including the `/mux` terminal WebSocket.
- Flip back to LAN and re-scan — connects, confirming the stale-`secure` trap is gone.
- Android regression check — both modes still pair.
- Desktop restart with secure pairing on — the proxy re-points at the new bridge port with no re-scan.
- Security: `/api/v1/mobile` returns 404 over the Tailscale address and over the LAN address; unauthenticated data routes 401; authenticated 200.

## Follow-ups, deliberately not here

- Emitting `no_magicdns` distinctly — `tailscale status` cannot distinguish a missing CLI from disabled MagicDNS, so both currently surface as `no_cli`, whose copy covers both.
- Automatic enabling of HTTPS certificates or MagicDNS: both are tailnet-wide administrative changes AO should not make on a user's behalf.


illegalcall (commented):
Thanks for working through the iOS/Tailscale pairing problem. I reviewed the backend, desktop, and mobile paths end to end. Could you please investigate the following before merge?

1. **Global `tailscale serve` ownership / destructive cleanup.** `BridgeService.Disable` calls `clearServe()` unconditionally, including when secure pairing was never enabled, and `Serve.Clear` runs `tailscale serve --https=443 off`. That can remove a pre-existing user Serve route or the route owned by another AO instance. `Apply` also replaces the root `:443` mapping without preserving prior state. Can AO preflight and establish ownership of the exact handler, then clear or restore only what AO installed?

2. **Secure QR compatibility with older mobile builds.** Secure QRs remain `v:1` and add `secure:true`. The previous v1 parser accepts unknown fields and ignores `secure`, so an older installed app will accept the QR and attempt plaintext HTTP against the MagicDNS host on port 443. This seems to contradict the compatibility claim. A distinct payload version for secure pairing, or another explicit capability strategy, would make old clients reject it instead of silently using the wrong transport.

3. **Proxy verification can inspect the wrong handler.** `Serve.Target` iterates every handler from any `*:443` site and returns the first proxy port. Tailscale supports multiple `--set-path` handlers, and Go map iteration is unordered, so an unrelated path or hostname can produce a false `port_mismatch` or false `active`. Could verification check the expected MagicDNS hostname and specifically its `/` handler?

4. **Persistent proxy during daemon shutdown.** `tailscale serve --bg` survives AO, but graceful shutdown only stops the LAN listener. The persistent proxy can remain aimed at an unbound local port until AO restarts; another local process that acquires that port could then be exposed to the tailnet without AO's auth middleware. Please investigate clearing the AO-owned handler during graceful shutdown and reapplying it during boot restore.

5. **Recovery copy is hidden or contradictory.** Turning secure pairing off can return `reason: "clear_failed"` with `enabled:false`, but `ConnectMobileSetup` renders reason text only when `secure.enabled` is true, so the cleanup instruction is unreachable. Also, every blocked secure state (`no_certs`, `serve_failed`, `port_mismatch`, etc.) renders the QR placeholder "Tailscale isn't running," even when Tailscale is running and a more specific reason is shown above it.

6. **Concurrency and cancellation.** `serveErr` is read and written by concurrent HTTP handlers without synchronization, which appears race-prone. The new status path also runs filesystem, interface, and subprocess I/O through `context.Background()`, so cancelling the request cannot stop the chained Tailscale calls. Please consider serializing BridgeService mutations/state and propagating `r.Context()`.

7. **Architecture and boundary coverage.** The current repo rules and ADR describe exactly the loopback and authenticated plaintext LAN listeners; this adds a tailnet-facing HTTPS endpoint and changes the documented QR security model without updating those decisions. If this is the intended architecture, the ADR, glossary, and hard-rule text should be amended. The new controller endpoint also needs direct tests for successful decoding, malformed or missing bodies, and bridge-error envelopes.

The core product direction makes sense, and the existing tests cover much of the happy path. My main concern is that `tailscale serve` is persistent, node-global state, while the current implementation treats it like state owned exclusively by one BridgeService instance.

Prasad-D-Ware:
@illegalcall 

> I’ve worked through all seven findings and want to confirm the intended architecture before implementing the fixes.
>
> 1. For Serve ownership, instead of temporarily replacing the node-wide `:443` root handler, would you prefer AO to allocate and persist a dedicated HTTPS port per AO instance?
>
>    The proposed behavior would be:
>
>    - Never modify the existing `:443` configuration.
>    - Inspect the exact `hostname:port` and `/` handler before applying anything.
>    - Use the port only when it is unused or still matches this AO instance’s recorded handler.
>    - Fail closed with a `serve_conflict` reason instead of replacing an unfamiliar handler.
>    - Remove only that exact owned HTTPS server during shutdown.
>
>    Would this satisfy the ownership and handler-verification concerns in items 1 and 3, or do you specifically want AO to coexist on `:443` through a dedicated path?
>
> 2. For secure QR compatibility, I propose keeping existing plaintext/LAN payloads at `v:1` and using `v:2` only for secure Tailscale payloads. New mobile builds would accept both versions, while older builds would reject the secure QR instead of silently attempting plaintext HTTP. Is that the compatibility behavior you expect?
>
> 3. For lifecycle handling, I propose preserving the user’s persisted `securePairing: true` preference while removing only AO’s owned Serve endpoint during graceful shutdown. Boot restore would recreate it against the listener’s newly bound port.
>
>    Following an ungraceful crash, AO would reclaim the endpoint only when its current handler exactly matches the previously recorded AO target. If it differs, AO would report `serve_conflict` and leave it untouched. Is that sufficient, or do you expect a stronger ownership mechanism for crash recovery?
>
> 4. For multiple AO instances, should secure pairing be supported concurrently for every instance? The dedicated-port design permits that, whereas a fixed port such as `8443` would intentionally allow only one secure-pairing instance per Tailscale node.
>
> 5. For concurrency, I plan to serialize BridgeService mutations and protect its Serve state, while propagating the HTTP request context through status and mutation operations instead of using `context.Background()`. Do you expect only in-process serialization, or are you also asking for cross-process coordination between AO instances?
>
> 6. For documentation, would you prefer extending ADR 0001 to cover the authenticated LAN listener being exposed through a tailnet-only HTTPS proxy, or creating a separate ADR specifically for Tailscale secure pairing and linking it from ADR 0001 and the repository hard rules?
>
> The remaining findings appear unambiguous, and I plan to address them directly:
>
> - Render `clear_failed` even when secure pairing is disabled.
> - Replace the generic “Tailscale isn’t running” placeholder with reason-specific messaging.
> - Add direct controller tests for valid, empty, malformed, and service-error requests.
> - Add ownership, conflict, exact-handler, shutdown, restore, concurrency, and cancellation coverage.
>
> Please let me know if any of those interpretations differ from what you intended.

illegalcall (approved):

PR #3709, opened 2026-08-08
feat: add review feedback injection toggle

VenkataSakethDakuri:
## Summary

- add a session-level toggle for automatic review feedback injection, defaulting to enabled
- snapshot the policy per SCM review, SCM comment, and AO review run so historical decisions remain stable
- keep non-injected feedback visible in Session Inspector with a Not injected marker
- persist the setting through SQLite, CDC, daemon API, generated types, and the reviewer UI

## Validation

- go build ./...
- go test ./internal/storage/sqlite/store ./internal/observe/scm ./internal/service/session ./internal/httpd/controllers ./internal/lifecycle ./internal/service/review ./internal/review
- go test ./internal/storage/sqlite/...
- go test ./internal/httpd/...
- npm run typecheck (frontend)
- npm test -- src/renderer/components/SessionInspector.test.tsx (78 passed)
- focused workspace and PR display frontend tests (114 passed across 3 files)


backend/internal/lifecycle/reactions.go
```diff
@@ -195,20 +195,23 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
...
-			msg = strings.Replace(msg, "your PR", ident, 1)
-		}
-		if o.URL != "" {
-			msg += "\nPR: " + domain.SanitizeControlChars(o.URL)
-		}
-		sig := reviewCommentsSignature(comments)
-		if sig == "" {
-			sig = string(o.Review)
+
+	if rec.AutoInjectReview {
```
neversettle17-101: This gates SCM review nudges on the current session row read during lifecycle, but the observer now snapshots `AutoInjectReview` onto each observed review/comment before calling lifecycle. Because persistence and `ApplySCMObservation` are separate steps, a user can toggle this setting between the DB write and this read: comments written with `auto_inject_review=false` can still be injected if the session is turned back on before this line, and comments written with `true` can be suppressed if it is turned off. That violates the snapshot semantics this PR adds for historical decisions. Please carry the observation-time policy (or filter on the persisted comment/review snapshot) into this decision instead of re-reading the live setting here.

neversettle17-101 (commented):
Needs changes. The new toggle is threaded through storage, API, review delivery, and the Session Inspector UI, with tests and generated API/types updated. I found one race/semantic issue where SCM review injection is still decided from the live session setting during lifecycle handling instead of the policy snapshot taken for the observed review/comment, so a toggle change between persistence and lifecycle can inject feedback that was explicitly recorded as not injectable (or suppress feedback recorded as injectable).

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -1239,6 +1239,28 @@ function ReviewsSection({
 			void queryClient.invalidateQueries({ queryKey: workspaceQueryKey });
 		},
 	});
+	const [autoInjectReview, setAutoInjectReview] = useState(session.autoInjectReview ?? true);
+	useEffect(() => {
+		setAutoInjectReview(session.autoInjectReview ?? true);
+	}, [session.id, session.autoInjectReview]);
+	const saveAutoInjectReview = useMutation({
```
neversettle17-101, line 1247: Add padding to this section

neversettle17-101 (changes requested):

neversettle17-101 (changes requested):
Let us remove tagging not injected to the comments from this PR -- it will be taken up as a followup PR for UI enhancement.

neversettle17-101 (changes requested):
Let us remove tagging not injected to the comments from this PR -- it will be taken up as a followup PR for UI enhancement. (2)

neversettle17-101 (changes requested):
Let us remove tagging not injected to the comments from this PR -- it will be taken up as a followup PR for UI enhancement. (3)

neversettle101 (changes requested):
Let us remove tagging not injected to the comments from this PR -- it will be taken up as a followup PR for UI enhancement. -- test

VenkataSakethDakuri:
Screenshot 2026-08-08 172323
Screenshot 2026-08-08 172823

neversettle17-101:
Stale review decisions can still be injected after the toggle is turned  on — backend/internal/observe/scm/observer.go:361-369 snapshots the current session policy onto every prepared SCM observation before checking whether review facts were actually fetched. When reviewStale[key] is true, domainFromObservation preserves the old local review decision, but ApplySCMObservation still receives o.AutoInjectReview from the current session and lifecycle uses it at backend/internal/lifecycle/reactions.go:203-208. That means a user can turn review injection on later, then a metadata/CI-only poll can inject an old changes_requested review that was observed while injection was off. Suggested fix: only pass/use the observation-time review policy when the review bucket was actually fetched/changed, or carry enough state into lifecycle to distinguish fresh review observations from preserved local review facts.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (changes requested):
The UI shouldn't mention AO code reviews and SCM differentiation. We should have an information icon which on hovering gives more information: "when disabled, review comments will not be sent to the worker agent but will be displayed"
The toggle button should come below the reviews section.

neversettle17-101 (approved):

PR #3713, opened 2026-08-08
fix(opencode): detect agent exit via AO's process supervisor

### LINKED ISSUE #2089: session restore unreachable when agent crashes/hangs but tmux session survives (isAlive treats tmux-session-exists as agent-alive)
## Summary

`session restore` (both the web UI "Restore" link and `ao session restore <name>`) refuses to act on a session whose **agent process has died or hung but whose tmux session still exists**. Such sessions land in a non-terminal `stuck`/`idle` limbo that `isRestorable()` rejects, so there is no in-product way to recover them — you have to manually kill the surviving shell first.

Root cause: for the tmux runtime, `isAlive()` checks **"does the tmux session exist?"**, not **"is the agent process inside the pane alive?"**. When an agent crashes (segfault) or deadlocks, the launch wrapper (`ao-launch-*.sh` / its bash) often survives in the pane, so AO still considers the runtime `alive` → status `stuck` → not terminal → not restorable.

## Environment

- `ao` 0.9.4 (`@aoagents/ao` → `@aoagents/ao-cli`, `@aoagents/ao-core`)
- runtime: `tmux` (tmux 3.2a), agent: `claude-code` (Claude Code 2.1.161)
- workspace: worktree
- Linux x86_64 (Ubuntu 22.04)

## What happened

Two sessions died on the same machine:

1. **Orchestrator** — the Claude Code (Bun) process **segfaulted** (`Illegal instruction (core dumped)`), leaving the pane at a bash prompt.
2. **Worker `sen-225`** — the Claude process **deadlocked** (alive but blocked in `futex_wait_queue`, 0 CPU progress, TUI frozen on the same spinner frame for minutes).

In both cases the **tmux session and its wrapper bash were still alive**. The web UI showed:

> process running. The live terminal is gone, but the session context is still available.

Clicking **Restore** did nothing. `ao session restore sen-225` failed identically:

```
Cannot restore: session is not in a terminal state (status: "stuck", activity: "idle")
```

## Root cause (with refs)

The web route `POST /api/sessions/[id]/restore` calls the same `SessionManager.restore()` from `@aoagents/ao-core` that the CLI uses (bundled into the web build), so UI and CLI behave identically.

`restore()` gates on `isRestorable(session)` (`ao-core/dist/session-manager.js` ~L2735; `ao-core/dist/types.js` L61), which requires a **terminal** state:

- `TERMINAL_STATUSES = { killed, terminated, done, cleanup, errored, merged }`
- or `activity === "exited"`
- or `lifecycle.runtime.state ∈ { missing, exited }`

During enrichment (`session-manager.js` ~L779):

```js
const alive = await plugins.runtime.isAlive(session.runtimeHandle);
if (!alive) { /* runtime.state = "missing"; activity = "exited"; ... terminal */ }
// else -> falls through to activity detection -> runtime.state = "alive", status "stuck"/"idle"
```

For the **tmux** runtime, `isAlive()` is true whenever the **tmux session exists** (the failure reason is even labeled `tmux_missing` vs `process_missing` at ~L783). A crashed/hung agent whose wrapper shell survives therefore reports `alive: true` → never terminal → `isRestorable` → false → restore refused.

## Reproduce

1. Spawn a session (tmux runtime).
2. Kill **only the agent** inside the pane, leaving the pane shell alive — e.g. `kill -9 <claude_pid>` (or simulate a crash/hang). The tmux session remains.
3. `ao status` shows the session as `stuck` (or the UI shows "process running / live terminal gone").
4. Click **Restore** (or run `ao session restore <name>`) → rejected: *"session is not in a terminal state"*.

## Workaround

Collapse the tmux session so `isAlive()` returns false, then restore:

```bash
tmux kill-session -t <session>     # or kill the surviving wrapper PID
ao session restore <session>        # now succeeds: runtime "missing" -> terminal -> restorable
```

## Suggested fix

For the tmux runtime, `isAlive()` (and/or enrichment) should distinguish **"tmux session exists"** from **"agent process is alive in the pane."** Options:

- Probe the pane's foreground/child process for the actual agent binary (and/or `pane_dead`) rather than treating session existence as liveness.
- Treat a session whose agent is gone/hung but whose shell survives as a restorable/terminal-ish state (e.g. a `crashed` status), so the existing Restore path can recover it in place instead of requiring a manual `tmux kill-session`.

Bonus: it'd help to have a way to recover a `stuck` (hung-but-alive) session from the UI without dropping to a shell — e.g. a "force restart" that kills the runtime and restores.

jbovard2016:
## Problem

The `opencode` adapter does not implement `ports.AgentExitDetector`, so it falls back to runtime-only liveness.

AO launches every seat as:

```
bash -c '<agent> ...; exec "${SHELL:-/bin/sh}" -i'
```

The trailing interactive shell is deliberate — it keeps the pane usable after the agent finishes. But it means the tmux pane **outlives the agent process**, so `tmux has-session` keeps answering "alive" long after `opencode` has exited or been killed.

Consequences observed on a worker seat whose agent was killed:

- the reaper never observes a dead worker, so the session is never marked terminated;
- `session restore` then refuses the session as not-restorable, because AO still believes it is alive;
- the seat is effectively unrecoverable without manual intervention.

Fixes #2089.

## Fix

Opt `opencode` into `AgentExitDetectionSupervisor`, exactly as the `codex` adapter already does (`adapters/agent/codex/codex.go`). The launch argv is then wrapped in `ao agent-process supervise`, which reports the real process exit back to the daemon, and `RuntimeLaunchID` is populated so the workload-death signal is no longer fenced off.

This is the mechanism AO already ships; the adapter simply was not opted in.

## Verification

Built from this branch and run as an isolated daemon (separate port, data dir, run file and `TMUX_TMPDIR`), against a scratch git repo:

1. Spawned an `opencode` worker seat; confirmed it did real work in its worktree.
2. `kill -9` on the `opencode` process.
3. **Before the patch:** session stayed `running` indefinitely; the tmux pane was still alive, and `has-session` still reported alive.
4. **After the patch:** session moved to `exited` within a few seconds of the kill, while the pane remained (as intended).
5. The seat could then be recovered in place via `POST /sessions/{id}/activity {"state":"exited"}` followed by `POST /sessions/{id}/resume-agent`, and the resumed agent wrote a new file into the preserved worktree — so the session became genuinely restorable, not merely relabelled.

`go test ./internal/adapters/agent/opencode/...` passes, including the added case.

## Notes

- No behaviour change for any other adapter.
- One incidental finding while testing, reported separately rather than folded in here: `respawnPaneArgs` targets `<session>:0.0` unconditionally, so on a machine with `base-index 1` / `pane-base-index 1` in `~/.tmux.conf` the pane is `1.1`, and `resume-agent` fails with `can't find window: 0`. That is independent of this fix and needs its own change (a tmux target that does not assume index 0, e.g. `{start}.{top-left}`); happy to open a separate PR for it.

### DIFF
--- a/backend/internal/adapters/agent/opencode/opencode.go
+++ b/backend/internal/adapters/agent/opencode/opencode.go
@@ -72,6 +72,14 @@ func New() *Plugin {
 	return &Plugin{}
 }
 
+// ExitDetectionMode opts opencode into AO's process supervisor. opencode's
+// launch command ends in an interactive keep-alive shell, so the tmux pane
+// outlives the agent process and runtime liveness alone reports a dead worker
+// as alive forever (#2089). Mirrors the codex adapter.
+func (p *Plugin) ExitDetectionMode() ports.AgentExitDetectionMode {
+	return ports.AgentExitDetectionSupervisor
+}
+
 var _ adapters.Adapter = (*Plugin)(nil)
 var _ ports.Agent = (*Plugin)(nil)
 var _ ports.AgentAuthChecker = (*Plugin)(nil)
--- a/backend/internal/adapters/agent/opencode/opencode_test.go
+++ b/backend/internal/adapters/agent/opencode/opencode_test.go
@@ -937,3 +937,10 @@ func contains(values []string, needle string) bool {
 	}
 	return false
 }
+
+func TestExitDetectionUsesAOProcessSupervisor(t *testing.T) {
+	plugin := &Plugin{}
+	if got := plugin.ExitDetectionMode(); got != ports.AgentExitDetectionSupervisor {
+		t.Fatalf("exit detection mode = %q, want %q", got, ports.AgentExitDetectionSupervisor)
+	}
+}


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@jbovard2016 
The issue this fixes, #2089, was closed after a live reproduction confirmed the crash case is handled on main. The PR also adds opencode exit detection that has not landed, so I filed nothing away silently: if that parity work is wanted, it deserves a fresh issue. Closing this one.

PR #3724, opened 2026-08-08
fix(chat): render image attachments in messages

### LINKED ISSUE #3722: Chat timeline shows image attachments as file paths instead of previews
## Bug

When a PNG is attached to a message in a chat-mode session, the composer stages it successfully and the agent can read it, but the human message in AO's transcript shows AO's generated attachment instruction and workspace-relative file path as literal text. The transcript never renders the image.

**Source:** In-app AO chat report on 2026-08-08  
**Reported by:** AO desktop user  
**Analyzed against:** `f7e0fe65bd6f` (`origin/main`, frontend `0.10.3`, daemon version `dev`)  
**Environment:** macOS Darwin 25.5.0 arm64; verified Go daemon on port 3001  
**Priority:** P2 — the attachment still reaches the agent, but the visible conversation is degraded and the only workaround is opening the staged path separately.  
**Confidence:** High — a deterministic renderer regression test exercises the exact persisted message shape and fails because the human-message component emits text only.

Chat message rendering the attachment as a path

## Reproduction

1. Open a chat-mode AO session.
2. Paste, drop, or choose a PNG in the chat composer.
3. Send the message.
4. Inspect the human message in the conversation timeline.

### Actual

The bubble contains text like:

```text
Attached files (read these files in the workspace):
- .ao/attachments/attachment-d9014f798f.png
```

No image element is rendered.

### Expected

AO should render staged image references as inline image previews while retaining the file in the worktree for the agent.

## Deterministic repro

```bash
cd frontend
npm test -- src/renderer/components/chat/ChatWorkspace.test.tsx -t "renders staged image references in human messages as images"
```

The test fails with `Unable to find an accessible element with the role "img"` and shows the generated reference block as the bubble's complete DOM text.

## Root Cause

- `frontend/src/renderer/components/chat/ChatComposer.tsx` intentionally appends staged paths to the outgoing text because worktree files are the provider-neutral delivery mechanism.
- `backend/internal/httpd/controllers/dto.go` exposes conversation messages with `text` but no structured display attachment metadata.
- `frontend/src/renderer/components/chat/ChatTimelineItems.tsx` renders `HumanMessage.message.text` verbatim and does not recognize AO's generated attachment suffix.
- The existing workspace-confined `GET /api/v1/sessions/{sessionId}/preview/files/*` route can serve the bytes, but the packaged renderer CSP currently does not allow loopback image sources.
- Spawned prompts use the closely related header `Attached files (read these files in the workspace for context):`, so both AO-generated variants need coverage.

## Fix

Parse only AO's terminal attachment-reference block in human messages, render supported staged image types through the existing session preview-file route, show non-image attachments as file labels, and preserve malformed or ordinary user text unchanged. Pass the session ID through the timeline and allow the already-trusted loopback daemon origin in `img-src`. Add regression coverage for both generated header variants and URL encoding.

## Impact

- Images still reach agents, so there is no data loss.
- Users cannot visually verify which screenshot/image they sent from the transcript.
- Image-only messages appear as internal implementation instructions rather than the intended content.

## Related

- [#2722](https://github.com/Untrivial-ai/agent-orchestrator/issues/2722) — original task-composer attachment request; adjacent composer behavior, not transcript rendering.
- [#3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360) — structured Chat/TUI session work that introduced the current chat surface.

illegalcall:
Fixes #3722

## Summary
- recognize AO-generated attachment suffixes in human chat messages
- render supported images through the existing workspace-confined session preview route
- show non-image attachments as compact file labels while preserving ordinary user-authored text verbatim, including trailing whitespace
- refresh preview URLs across daemon restarts and allow loopback-served images in the packaged renderer CSP

## Tests
- issue-published regression command: `cd frontend && npm test -- src/renderer/components/chat/ChatWorkspace.test.tsx -t "renders staged image references in human messages as images"` (1 passed)
- `cd frontend && npm test -- src/renderer/components/chat/ChatWorkspace.test.tsx` (26 passed)
- `cd frontend && npm test -- src/renderer` (1,480 passed)
- renderer production build with `vite.renderer.config.ts`; verified generated CSP contains the loopback image origin

## Test-suite note
Unscoped `cd frontend && npm test` reached 152 passing test files and 2,037 passing tests. Its remaining two landing-site suites require dependencies and path aliases from the separately installed nested landing package (`cheerio` and `@ao/shared/constants`); neither failure exercises this renderer change.

## Visual verification
Verified in the native Electron app on code commit `ba2a88c20` using isolated AO data. A live human message renders the staged PNG inline through the session-scoped preview route, and the provider identifies the same blue elephant.

AO Chat timeline rendering a staged PNG inline


illegalcall:
Comparison follow-up completed in `ba2a88c20`:

1. **Regression command — valid.** Restored the exact test name published in #3722. The published `-t` command now selects and passes one test.
2. **Authored whitespace — valid.** Removed the unintended `trimEnd()` and added focused coverage proving trailing spaces/newlines survive removal of the AO-generated attachment suffix.

Visual evidence from the real Electron app in isolated-data mode is embedded in the PR body:

AO Chat timeline rendering a staged PNG inline

nikhilachale (approved):

PR #3711, opened 2026-08-08
fix: recover Chat sessions after controller stops

illegalcall:
## What

- report Chat lifecycle `exited` whenever the provider event stream ends, including abrupt disconnects without a final controller-state event
- allow `chat.controller.stopped` to clear sticky `blocked` activity
- make agent resume mode-aware so Chat sessions relaunch their provider conversation without a terminal runtime handle
- recover stale pre-fix Chat activity only when no live controller exists, and wait for a stopped controller's final cleanup before replacement

## Why

The desktop can correctly show **The agent controller stopped** while **Resume agent** fails with `AGENT_NOT_EXITED`. Some sessions remain durably idle/active/blocked after their Chat controller dies; sessions that do reach exited then fail the TUI-only runtime-handle precondition.

Fixes #3710.

## How

The controller's provider-stream closure now emits the same generation-fenced lifecycle signal as an explicit stopped event. Lifecycle treats that event as a turn boundary so a dead controller cannot remain blocked.

Session Manager keeps TUI resume semantics unchanged. For Chat, the in-memory controller registry prevents duplicate recovery of a live controller and permits stale durable rows to recover when no controller remains. The Chat service distinguishes a stopped registry entry from a usable controller and waits for its projector cleanup before launching the replacement generation.

No frontend or API contract changes are required; the existing recovery action works once daemon state and resume semantics agree.

## Testing

- `GOFLAGS=-p=1 npm run lint`
- `cd backend && go build ./...`
- `cd backend && go test -race -p 1 ./internal/service/chat ./internal/lifecycle ./internal/session_manager`
- focused regressions for raw stream closure, blocked-to-exited transition, stale/live controller guards, runtime-handle-free Chat resume, and stopped-controller relaunch ordering

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; links the related issue when applicable
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense
- [x] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

Issue #170, opened 2026-02-24
Autonomous build test


zenacquire:
Spawn Claude + Codex agents

harshitsinghbhandari:
Closing this as part of the migration cleanup for the ground-up rewrite of
Agent Orchestrator.

This item was opened more than a month before the migration batch and the
codebase has been replaced, so this thread no longer maps cleanly onto the
current source tree.

If something here still needs tracking, please reopen this item or open a fresh
one against the new tree. Thank you for contributing.

PR #3735, opened 2026-08-08
fix: recover Codex interface switch readiness

illegalcall:
## Summary

- invoke Codex lifecycle hooks with the running AO executable's absolute, shell-quoted path, so Codex's login shell cannot redirect callbacks to a stale or foreign `ao`
- recheck only the transient `NATIVE_SESSION_MISSING` interface status, allowing **Switch to chat UI** to enable as soon as a successful `SessionStart` hook records the thread id
- implement Codex's native-history probe so a reserved UUID with no active rollout starts fresh instead of issuing a doomed `thread/resume`

## RCA

Three independent gaps produced two similar-looking symptoms:

1. **The native id never arrives for some users.** Codex runs command hooks through the user's login shell. AO previously emitted bare `ao hooks codex ...` commands and only prepended the daemon directory to the child process's inherited `PATH`. A login-shell startup file can replace or reorder that `PATH`; the hook then invokes another AO installation (or no AO), exits, and never records the native Codex thread id. This depends on each user's shell setup, which explains why only some users reproduce it indefinitely.
2. **The renderer can cache an early missing-id response.** Even when the hook succeeds, the initial interface-status request can race the asynchronous `SessionStart` callback. The query returned `NATIVE_SESSION_MISSING`, but idle status queries did not refetch, leaving the button stale until some unrelated remount/refetch.
3. **A UUID is not necessarily resumable history.** Codex app-server can reserve and return a thread UUID before the first user turn writes an active rollout. AO treated every non-empty Codex UUID as resumable, passed an unmaterialized id to `thread/resume`, and surfaced `no rollout found for thread id ...`.

## Why the earlier fixes did not cover this

- #170 prepended AO's daemon directory to the spawned session `PATH`. That helps ordinary child-process lookup, but it does not survive a login shell that reconstructs `PATH` before running a hook.
- #3711 recovers a stopped Chat controller. It does not make a missing Codex rollout resumable and therefore cannot fix `no rollout found`.
- #3719 prevents a stale-idle transition drain from hanging. It is a separate transition-lifecycle failure.
- #3717 changes frontend chat links/text selection and is unrelated.

### Real desktop reproduction on `main`

1. Launch the actual AO Electron app against an isolated data directory.
2. Start a Codex 0.147 TUI session and send `Reply with only OK`.
3. Observe Codex return `OK`, while **Switch to chat UI** remains disabled and Activity shows **No Signal**.
4. In the reproduced shell, `zsh -lc 'command -v ao'` resolves a stale package-manager shim; invoking it exits 1 because its old module target no longer exists. Therefore the `SessionStart` callback never reaches this daemon.
5. To isolate the second bug, record the valid native id directly. The daemon then reports `{supported:true,targetMode:"chat"}`, but the already-mounted UI remains disabled because it never refetches the earlier `NATIVE_SESSION_MISSING` result.

### Real desktop verification on this branch

1. Start the same Codex flow from the fixed app.
2. Inspect the live launch command: every Codex hook contains the absolute daemon executable and the isolated daemon run-file environment.
3. Send the same prompt. The real `SessionStart` callback records the native id, the button becomes enabled, and Activity becomes **Idle**.
4. Click **Switch to chat UI**. The transition completes and Chat UI displays the original `Reply with only OK` prompt and `OK` response. The reverse **Switch to terminal UI** control is enabled.

### Automated regression evidence

- Before the absolute-path implementation, the launch/restore command tests failed because the generated hook command still contained bare `ao hooks ...`; after the change, the same tests pass with the current AO executable embedded.
- The renderer regression test starts at `NATIVE_SESSION_MISSING`, advances the mocked daemon to `supported: true`, and proves the hook refetches exactly once. A second test proves permanently unsupported harnesses are not polled.
- Session-manager tests prove a missing Codex rollout uses the fresh-conversation sentinel, while a non-empty active `.jsonl` rollout preserves the native id.
- Adapter tests additionally cover missing, zero-byte, archived, active, and compressed active rollout files.

## Change-by-change necessity

| Change | Required? | Reason |
|---|---:|---|
| Absolute, shell-quoted AO hook executable | Yes | Fixes the indefinite, user-specific failure where no native id is ever recorded. Quoting is required for packaged paths such as `Agent Orchestrator.app`. |
| Error propagation from launch/restore | Yes | A session launched without working tracking hooks recreates the same unrecoverable disabled state. Failing the launch is safer and diagnosable. |
| Targeted one-second `NATIVE_SESSION_MISSING` poll | Yes | Absolute hooks can still finish after the first HTTP read; polling closes that legitimate async race. It stops immediately when ready and never runs for permanently unsupported harnesses. |
| Codex active-rollout probe | Yes | Prevents `thread/resume` from receiving an id that Codex has reserved but cannot resume. |
| Missing/zero-byte/archived vs active/compressed tests | Yes | These are the exact provider states that decide whether continuity is safe. |
| Frontend permanent-unsupported test | Yes | Guards the UX/performance boundary: Cursor and other unsupported harnesses must not begin background polling. |

No API, schema, storage, or UI-component change is needed.

## Validation

- `go test ./internal/adapters/agent/codex ./internal/session_manager -count=1`
- `go test ./...` (via the repository lint command)
- focused `golangci-lint` for the changed backend packages: `0 issues`
- `npm test -- --run src/renderer/hooks/useSessionInterfaceTransition.test.tsx`
- `npm run typecheck`
- `npm run package` (production Electron bundles and macOS package)
- actual AO Electron + Codex 0.147 TUI-to-Chat handoff against isolated data

The repository-wide linter process also scanned stale sibling worktrees on this machine and reported pre-existing/generated findings from those other paths. The changed packages themselves lint clean, and CI remains the clean-checkout authority.

### DIFF
--- a/backend/internal/adapters/agent/codex/codex.go
+++ b/backend/internal/adapters/agent/codex/codex.go
@@ -117,7 +117,9 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
 	appendHideRateLimitNudgeFlag(&cmd)
 	appendHookTrustBypassFlag(&cmd)
 	appendApprovalFlags(&cmd, cfg.Permissions)
-	appendSessionHookFlags(&cmd)
+	if err := appendSessionHookFlags(&cmd); err != nil {
+		return nil, err
+	}
 	appendTerminalCompatibilityFlags(&cmd)
 	appendWorkspaceTrustFlag(&cmd, cfg.WorkspacePath)
 	appendModelFlag(&cmd, cfg.Config)
@@ -159,7 +161,9 @@ func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig)
 	appendHideRateLimitNudgeFlag(&cmd)
 	appendHookTrustBypassFlag(&cmd)
 	appendApprovalFlags(&cmd, cfg.Permissions)
-	appendSessionHookFlags(&cmd)
+	if err := appendSessionHookFlags(&cmd); err != nil {
+		return nil, false, err
+	}
 	appendTerminalCompatibilityFlags(&cmd)
 	appendWorkspaceTrustFlag(&cmd, cfg.Session.WorkspacePath)
 	appendModelFlag(&cmd, cfg.Config)
@@ -450,7 +454,12 @@ func DoctorLaunchProbes() [][]string {
 	overrideProbe := []string{"features", "list"}
 	appendNoUpdateCheckFlag(&overrideProbe)
 	appendHideRateLimitNudgeFlag(&overrideProbe)
-	appendSessionHookFlags(&overrideProbe)
+	if err := appendSessionHookFlags(&overrideProbe); err != nil {
+		// The probe only asks Codex to parse the hook config; a bare fallback
+		// keeps that diagnostic available if the current executable cannot be
+		// resolved, while real session launches fail closed above.
+		appendSessionHookFlagsForExecutable(&overrideProbe, "ao")
+	}
 	appendWorkspaceTrustFlag(&overrideProbe, os.TempDir())
 	return [][]string{flagProbe, overrideProbe}
 }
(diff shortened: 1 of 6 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):

PR #3737, opened 2026-08-08
fix(cli): merge set-config by default instead of silent full replace

### LINKED ISSUE #3714: ao project set-config silently discards unspecified config (full replace, no warning)
## Summary

`ao project set-config` is a **full replace**, not a merge. Any field not supplied in the invocation is reset to its zero value, silently discarding config that is already set. Because the command succeeds and prints no warning, the loss is invisible until a later session behaves differently.

This is easy to hit while a project has running sessions: changing one setting (say a model) drops unrelated settings (env, worker config, setup commands) for every session started afterwards.

## Reproduction

```
# 1. establish a config with two independent settings
ao project set-config <project> \
  --config-json '{"agentConfig":{"model":"<model-a>"},"worker":{"agentConfig":{"model":"<model-a>"}}}' \
  --env "SOME_VAR=hello"

ao project get-config <project>     # both present

# 2. later, change only the model
ao project set-config <project> \
  --config-json '{"agentConfig":{"model":"<model-b>"}}'

ao project get-config <project>
```

**Expected:** model updated to `<model-b>`; `SOME_VAR` and the worker block untouched.

**Actual:** model is `<model-b>`, and `SOME_VAR` and the entire `worker` block are gone. Exit code 0, no warning.

## Cause

`buildProjectConfig` (`internal/cli/project.go`) constructs a fresh config object from the flags provided on this invocation and writes it wholesale; there is no read-modify-write against the stored config. The help text does say "Replace a project's per-project config", so the behaviour is intentional — but the failure mode is silent and unbounded, which is what makes it dangerous.

## Impact

The destructive case is not a typo, it is a routine one-setting change. With sessions already running against the project, the operator has no signal that anything else was dropped; the next spawned session simply picks up a different environment. Recovering requires knowing what the config used to be, which nothing records.

## Suggested fix

Any one of these would remove the silent-loss property:

1. **Merge by default**, with an explicit `--replace` flag for the current behaviour. Most consistent with how the command is actually used.
2. Keep replace semantics but **print what is being dropped** and require confirmation (or `--yes`) when the write removes previously-set keys.
3. At minimum, **warn loudly** when a replace discards non-empty fields, and mention the replace semantics in the command's short help rather than only the long description.

Option 1 with 2 as the safety net for `--replace` seems the best balance; happy to send a PR if you have a preference on which way you want it to go.

## Environment

- Reproduced against a local build of `main`.
- Linux, single-repo project.

bluwwi:
Fixes [#3714](https://github.com/Untrivial-ai/agent-orchestrator/issues/3714)

## Summary
- `ao project set-config` did a full replace — any field not supplied was silently wiped to its zero value, with no warning. A routine one-setting change (e.g. updating the model) would drop unrelated settings like env vars, worker config, and setup commands.
- `buildProjectConfig` now starts from the current config fetched via GET and overlays only flags the user explicitly passed (using cobra's `Changed()` check).
- `--config-json` deep-merges into the existing config instead of replacing it wholesale.
- A new `--replace` flag opts into the old full-replace behavior for users who want it. `--clear` still wipes everything.
- Help text updated: "Set" → "Update", mentions merge-by-default and `--replace`.

## Tests
- `cd backend && go test ./internal/cli/... -run TestProjectSetConfig -v` (7 passed)
- `cd backend && go test ./internal/cli/... -run TestBuildProjectConfig -v` (1 passed)
- `cd backend && go test ./internal/cli/...` (full CLI suite passed)
- `cd backend && go vet ./internal/cli/...`

## Manual verification
Verified against a local daemon: setting `--env FOO=bar` then later `--model claude-opus-4-5` preserves `FOO=bar` (was silently wiped before). `--replace` restores old behavior. `--config-json` with a partial object merges into existing config. `--clear` still wipes everything.
Screenshot 2026-08-08 165908
Screenshot 2026-08-08 165614

### DIFF
--- a/backend/internal/cli/project.go
+++ b/backend/internal/cli/project.go
@@ -95,22 +95,34 @@ type trackerIntakeConfig struct {
 	Assignee string `json:"assignee,omitempty"`
 }
 
+// reviewerConfig mirrors domain.ReviewerConfig.
+type reviewerConfig struct {
+	Harness string `json:"harness,omitempty"`
+}
+
+// containerReapConfig mirrors domain.ContainerReapConfig.
+type containerReapConfig struct {
+	Disabled bool `json:"disabled,omitempty"`
+}
+
 // projectConfig mirrors the daemon's typed domain.ProjectConfig for the CLI
 // client. The CLI sets common fields via flags and the whole object via
 // --config-json.
 type projectConfig struct {
-	DefaultBranch     string              `json:"defaultBranch,omitempty"`
-	SessionPrefix     string              `json:"sessionPrefix,omitempty"`
-	Env               map[string]string   `json:"env,omitempty"`
-	Symlinks          []string            `json:"symlinks,omitempty"`
-	PostCreate        []string            `json:"postCreate,omitempty"`
-	AgentRules        string              `json:"agentRules,omitempty"`
-	AgentRulesFile    string              `json:"agentRulesFile,omitempty"`
-	OrchestratorRules string              `json:"orchestratorRules,omitempty"`
-	AgentConfig       agentConfig         `json:"agentConfig,omitempty"`
-	Worker            roleOverride        `json:"worker,omitempty"`
-	Orchestrator      roleOverride        `json:"orchestrator,omitempty"`
-	TrackerIntake     trackerIntakeConfig `json:"trackerIntake,omitempty"`
+	DefaultBranch     string               `json:"defaultBranch,omitempty"`
+	SessionPrefix     string               `json:"sessionPrefix,omitempty"`
+	Env               map[string]string    `json:"env,omitempty"`
+	Symlinks          []string             `json:"symlinks,omitempty"`
+	PostCreate        []string             `json:"postCreate,omitempty"`
+	AgentRules        string               `json:"agentRules,omitempty"`
+	AgentRulesFile    string               `json:"agentRulesFile,omitempty"`
+	OrchestratorRules string               `json:"orchestratorRules,omitempty"`
+	AgentConfig       agentConfig          `json:"agentConfig,omitempty"`
+	Worker            roleOverride         `json:"worker,omitempty"`
+	Orchestrator      roleOverride         `json:"orchestrator,omitempty"`
+	Reviewers         []reviewerConfig     `json:"reviewers,omitempty"`
+	TrackerIntake     trackerIntakeConfig  `json:"trackerIntake,omitempty"`
+	ContainerReap     containerReapConfig  `json:"containerReap,omitempty"`
 }
 
 // setConfigRequest mirrors the daemon's SetConfigInput body for
@@ -302,8 +314,9 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
 			if !opts.replace && !opts.clear {
 				var existing projectGetResult
 				if err := ctx.getJSON(cmd.Context(), "projects/"+url.PathEscape(id), &existing); err != nil {
-					current = projectConfig{}
-				} else if existing.Project.Config != nil {
+					return fmt.Errorf("cannot merge: fetch current config for %s: %w", id, err)
+				}
+				if existing.Project.Config != nil {
 					current = *existing.Project.Config
 				}
 			}
@@ -388,7 +401,12 @@ func buildProjectConfig(opts projectSetConfigOptions, current projectConfig, fla
 		if err != nil {
 			return projectConfig{}, err
 		}
-		cfg.Env = env
+		if cfg.Env == nil {
+			cfg.Env = make(map[string]string)
+		}
+		for k, v := range env {
+			cfg.Env[k] = v
+		}
 	}
 	if flags.Changed("symlink") {
 		cfg.Symlinks = opts.symlink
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/cli/project.go
```diff
@@ -293,7 +297,18 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
...
 			id := strings.TrimSpace(args[0])
-			config, err := buildProjectConfig(opts)
+
+			var current projectConfig
+			if !opts.replace && !opts.clear {
+				var existing projectGetResult
+				if err := ctx.getJSON(cmd.Context(), "projects/"+url.PathEscape(id), &existing); err != nil {
+					current = projectConfig{}
+				} else if existing.Project.Config != nil {
+					current = *existing.Project.Config
```
whoisasx, line 322: [P1] The GET response is decoded through the CLI's `projectConfig` mirror, which omits `Reviewers` and `ContainerReap` even though both are valid `domain.ProjectConfig` fields. A routine update such as `--model` therefore PUTs a complete object without those settings and permanently removes them. Complete the mirror or preserve the raw/generated API shape, and add a round-trip test containing both fields.

backend/internal/cli/project.go
```diff
@@ -327,55 +342,163 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
...
+	if flags.Changed("permission") {
+		cfg.AgentConfig.Permissions = opts.permission
+	}
+	if flags.Changed("worker-agent") {
+		cfg.Worker.Agent = opts.workerAgent
+	}
+	if flags.Changed("orchestrator-agent") {
+		cfg.Orchestrator.Agent = opts.orchestratorAgent
+	}
+	if flags.Changed("tracker-intake") || flags.Changed("tracker-repo") || flags.Changed("tracker-assignee") {
```
whoisasx, line 453: [P1] This reconstructs the entire tracker block whenever any one tracker flag changes. For example, `--tracker-assignee bob` resets `Enabled` to false and `Repo` to empty because those option values were not supplied. Overlay only the tracker fields whose flags changed, and test each flag against an existing fully populated tracker configuration.

backend/internal/cli/project.go
```diff
@@ -293,7 +297,18 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			id := strings.TrimSpace(args[0])
-			config, err := buildProjectConfig(opts)
+
+			var current projectConfig
+			if !opts.replace && !opts.clear {
+				var existing projectGetResult
+				if err := ctx.getJSON(cmd.Context(), "projects/"+url.PathEscape(id), &existing); err != nil {
```
whoisasx, line 318: [P2] Falling back to an empty config after any GET error makes the safety mechanism fail open: a transient HTTP or decode failure followed by a successful PUT replaces the stored config with only the supplied fields. Return the GET error and add GET-500 and decode-failure tests asserting that no PUT occurs.

backend/internal/cli/project.go
```diff
@@ -327,55 +342,163 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
...
+func mergeProjectConfig(current, overlay projectConfig) projectConfig {
+	out := current
+	if overlay.DefaultBranch != "" {
+		out.DefaultBranch = overlay.DefaultBranch
+	}
+	if overlay.SessionPrefix != "" {
+		out.SessionPrefix = overlay.SessionPrefix
+	}
+	if len(overlay.Env) > 0 {
+		out.Env = overlay.Env
```
whoisasx: [P2] Assigning `overlay.Env` wholesale means partial JSON such as `{"env":{"A":"new"}}` silently deletes every existing environment key other than `A`, despite the command advertising a deep merge. Merge map entries by key and test that unrelated existing keys survive; `--replace` can retain whole-map replacement semantics.

whoisasx (changes requested):
The merge-by-default direction is correct, but the current implementation still contains four silent data-loss paths: incomplete CLI round-tripping, wholesale tracker replacement, fail-open GET handling, and map replacement in the advertised deep-merge path. Requesting changes until these preservation cases and their regression tests are addressed.

bluwwi:
@whoisasx  <h1>Found and fixed 2 additional bugs in the <code>--config-json</code> merge path.</h1>
<hr>
<h2>Bug 1 — <code>trackerIntake.enabled</code> couldn't be set to <code>false</code> via <code>--config-json</code> merge</h2>
<p>The merge guard <code>overlay.TrackerIntake.Enabled || ...</code> is <code>false</code> for both an absent field and an
explicit <code>false</code>, so <code>{"trackerIntake":{"enabled":false}}</code> was a silent no-op — the existing <code>true</code>
value was preserved and the disable never landed.</p>
<p>The flag path (<code>--tracker-intake=false</code>) worked because it uses <code>flags.Changed()</code>, but
<code>--config-json</code> is the only way to set this field via JSON.</p>
<hr>
<h2>Bug 2 — <code>containerReap.disabled</code> couldn't be turned back off via <code>--config-json</code> merge</h2>
<p>The guard <code>if overlay.ContainerReap.Disabled { out.ContainerReap.Disabled = true }</code> only copies
<code>true</code>, never <code>false</code>. No CLI flag exists for this field, so once disabled the only way to
re-enable was <code>--replace</code> or <code>--clear</code> (which discard unrelated settings).</p>
<hr>
<h2>Root Cause</h2>
<p>Go's zero-value semantics — <code>false</code> is indistinguishable from "field not present in JSON" after
unmarshaling a plain <code>bool</code>.</p>
<hr>
<h2>Fix</h2>
<p>Changed both fields to <code>*bool</code> in the CLI mirror structs. The merge logic now checks
<code>if overlay.X != nil</code> and copies the pointed-to value, so an explicit <code>false</code> in <code>--config-json</code>
is honored. The flag path takes the address of the cobra bool value.</p>
<blockquote>
<p><strong>Wire format unchanged</strong> — Go JSON encodes <code>*bool</code> the same as <code>bool</code>.</p>
</blockquote>
<hr>
<h2>New Tests</h2>

Test | Description
-- | --
TestProjectSetConfig_ConfigJSONCanDisableTrackerIntake | Verifies {"enabled":false} disables an existing enabled tracker via merge
TestProjectSetConfig_ConfigJSONCanReenableContainerReap | Verifies {"disabled":false} re-enables reaping via merge

</body></html>
</body>
</html>

harshitsinghbhandari (commented):
Two follow-ups rather than blockers. First, the merge is client-side GET-then-PUT, so it is not atomic, and a concurrent `set-config` or a desktop UI edit inside that window can still be clobbered silently, which is the same class of loss this PR fixes. Second, the CLI mirror of `domain.ProjectConfig` is complete now but hand-maintained, so the next field added to the domain type gets silently dropped by every merge PUT until someone remembers to mirror it, and nothing in CI catches that; merging through a raw JSON shape instead of the typed mirror would close that class for good. One minor thing worth noting in the help text: merge can add or overwrite env keys but never delete one, so unsetting still requires `--replace` or `--clear`, and the "provide at least one config flag" usage error is now effectively dead for a configured project, since a bare `set-config` produces a non-empty merged config and PUTs it back unchanged instead of erroring. The three inline threads that the follow-up commits already answered could be resolved to make the re-review quicker.

i-trytoohard:
@bluwwi Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@bluwwi

PR #3487, opened 2026-08-03
fix(docs): highlight bottom TOC item when scrolled to bottom of page (#3486)

atul-upadhyay-7:
Fixes #3486

When scrolling to the very bottom of documentation pages, the final section heading top position doesn't reach the `<= 120px` threshold, keeping the active sidebar indicator stuck on an earlier item.

Added a check in `DocsToc.tsx` so the last TOC heading activates when scrolled to the bottom of the document.

PR #3740, opened 2026-08-08
fix: resolve three safe issue fixes

### LINKED ISSUE #3486: Docs TOC active section indicator does not highlight bottom heading on scroll
When scrolling down to the bottom of a documentation page, the sidebar Table of Contents ("On this page") active indicator stays stuck on earlier headings (e.g. "Install") and does not highlight the bottom section ("Troubleshooting").

### Root Cause
In `frontend/src/landing/src/app/docs/components/DocsToc.tsx`, active heading detection checks `el.getBoundingClientRect().top <= 120`. When scrolled to the bottom of the page, the final heading top position does not cross the `<= 120px` threshold, so the indicator never moves to the last section.

### Screenshots
Docs TOC Scroll 1

Docs TOC Scroll 2

Docs TOC Scroll 3

Docs TOC Scroll 4

### LINKED ISSUE #3581: bug(goose): GetRestoreCommand omits -t/-i/--recipe, restore silently no-ops on Goose 1.45+

### LINKED ISSUE #3698: Activity timeline events not sorted chronologically
## Bug

The session inspector's **Activity timeline** does not sort events chronologically. Events are displayed in category-group order (Created → Drafts → Opened → Merged → Done) rather than by timestamp, causing recently-opened PRs to appear below older events.

**Reported by:** @prateek in [Discord](https://discord.com/channels/1476302178913357958/1535215769460408392)
**Analyzed against:** `f17013b53`
**Confidence:** High — traced exact code path, fix verified by tests

### Screenshot evidence

The reported session shows:

| Position | Event | Timestamp |
|----------|-------|-----------|
| 1 | Working | (live) |
| 2 | Merged PR #3647 | 18h ago |
| 3 | Opened PR #3647 | 1d ago |
| 4 | **Opened PR #3683** | **53m ago** ← most recent historical event, but buried at position 4 |
| 5 | Created workspace | 1d ago |

"Opened PR #3683" (53m ago) should appear at position 2 (just below Working), not position 4.

## Root Cause

`ActivityTimeline` in `frontend/src/renderer/components/SessionInspector.tsx` (line 997):

```typescript
const events = [current, ...history.reverse()];
```

The `history` array is built by pushing events in **category groups**:

1. Created workspace
2. All draft PRs (in PR-number order)
3. All opened PRs (in PR-number order)
4. All merged PRs (in PR-number order)
5. Done (if merged)

`.reverse()` simply flips this category order — it does **not** sort by timestamp. When two PRs are in the same category (e.g. both "opened"), they remain in PR-number order regardless of when they were actually opened. PR #3683 (53m ago) gets placed after PR #3647 (1d ago) because 3683 > 3647 numerically within the "opened" group.

This is invisible when a session has only one PR (the category order happens to be chronological), but breaks for any session with multiple PRs opened/merged at different times.

## Fix

Replace `history.reverse()` with a proper reverse-chronological sort by timestamp:

```typescript
const sortedHistory = [...history].sort((a, b) => b.sortTs - a.sortTs || b.seq - a.seq);
```

Each history entry now carries a `sortTs` (epoch ms parsed from `createdAt`/`stateChangedAt`/`updatedAt`) and a `seq` (push order, used as a tiebreaker when timestamps are equal or missing — preserving the semantic grouping as a fallback).

## Impact

- **Severity:** Medium — confusing UX, not data loss. Users rely on the timeline to understand session history at a glance.
- **Affects:** Any session with multiple PRs where events don't happen in PR-number order.
- **Workaround:** None (users see incorrect ordering).

## Test Coverage

- Updated existing test to use realistic timestamps (previously used artificial equal timestamps that masked the bug).
- Added regression test reproducing the exact reported scenario (PR #3683 opened 53m ago vs PR #3647 opened 1d ago).

athrvd7:
This PR contains three independently verified, low-risk fixes:

- Fix Goose restore commands for Goose 1.45+ by supplying the required empty `-t` argument. Fixes #3581.
- Sort Session Inspector activity history by event timestamp instead of category insertion order. Fixes #3698.
- Activate the final Docs TOC heading when the page reaches its scroll limit. Fixes #3486.

Verification:
- `go test ./internal/adapters/agent/goose`
- `npm test -- --run src/renderer/components/SessionInspector.test.tsx`
- `npm run typecheck`

### DIFF
--- a/backend/internal/adapters/agent/goose/goose.go
+++ b/backend/internal/adapters/agent/goose/goose.go
@@ -135,7 +135,7 @@ func (p *Plugin) GetPromptDeliveryStrategy(ctx context.Context, _ ports.LaunchCo
 
 // GetRestoreCommand rebuilds the argv that continues an existing Goose session:
 //
-//	[env GOOSE_MODE=<mode>] goose run --system <text> --resume --session-id <agentSessionId>
+//	[env GOOSE_MODE=<mode>] goose run --system <text> -t "" --resume --session-id <agentSessionId>
 //
 // ok is false when the hook-derived native session id has not landed yet, so
 // callers can fall back to fresh launch behavior. AO deliberately uses run
@@ -165,7 +165,9 @@ func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig)
 		cmd = append(cmd, "--system", systemPrompt)
 	}
 	appendModelFlag(&cmd, cfg.Config)
-	cmd = append(cmd, "--resume", "--session-id", agentSessionID)
+	// Goose requires one of --instructions, --text, or --recipe even when
+	// resuming. Empty text preserves the interactive, non-replaying behavior.
+	cmd = append(cmd, "-t", "", "--resume", "--session-id", agentSessionID)
 	return cmd, true, nil
 }
 
--- a/backend/internal/adapters/agent/goose/goose_test.go
+++ b/backend/internal/adapters/agent/goose/goose_test.go
@@ -440,12 +440,12 @@ func TestGetRestoreCommandReadsAgentSessionID(t *testing.T) {
 	}
 	want := []string{
 		"env", "GOOSE_MODE=auto",
-		"goose", "run", "--system", "restore inline wins", "--resume", "--session-id", "20260720_1",
+		"goose", "run", "--system", "restore inline wins", "-t", "", "--resume", "--session-id", "20260720_1",
 	}
 	if !reflect.DeepEqual(cmd, want) {
 		t.Fatalf("restore cmd\nwant: %#v\n got: %#v", want, cmd)
 	}
-	if contains(cmd, "-t") || contains(cmd, "restore original task") {
+	if contains(cmd, "restore original task") {
 		t.Fatalf("restore command %#v unexpectedly replays the original task", cmd)
 	}
 }
--- a/frontend/src/landing/src/app/docs/components/DocsToc.tsx
+++ b/frontend/src/landing/src/app/docs/components/DocsToc.tsx
@@ -15,6 +15,11 @@ export function DocsToc({ toc }: { toc: TocItem[] }) {
         // Last heading whose top has scrolled past the header band is the active one.
         if (el && el.getBoundingClientRect().top <= 120) current = id;
       }
+      // The last heading may never reach the threshold when the document
+      // cannot scroll any further. At the bottom, it is the active section.
+      if (window.scrollY + window.innerHeight >= document.documentElement.scrollHeight - 1) {
+        current = ids.at(-1) ?? current;
+      }
       setActive(current);
     };
     update();
--- a/frontend/src/renderer/components/SessionInspector.test.tsx
+++ b/frontend/src/renderer/components/SessionInspector.test.tsx
@@ -877,12 +877,12 @@ describe("SessionInspector Activity section", () => {
 		);
 		expect(rows).toEqual([
 			"Idle",
-			"Done",
+			"Created workspace3h ago",
 			"Merged PR #40",
+			"Draft PR #42",
+			"Done",
 			"Opened PR #40",
 			"Opened PR #41",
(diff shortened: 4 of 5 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@athrvd7 
Both halves of this PR landed separately: the docs TOC in #3487 and the timeline sort in #4042. The goose part is a comment-only subset of open #3628. Closing as superseded.

PR #3741, opened 2026-08-08
fix(send): delay Enter after large Codex paste bursts

### LINKED ISSUE #3626: fix(send): multiline prompts for codex stay unsubmitted in the input box (confirmActive rescue skipped)
## Summary

Multiline (large) prompts sent to a **codex** worker sometimes sit unsubmitted in the TUI input box — `tmux send-keys` exits 0 and `POST /send` (and `ao send`) returns `200 OK`, but the trailing `Enter` never submits the prompt, so the agent never receives it. The user sees no error and waits for a reply that will never come.

This is the **codex** half of the bug that was fixed for **claude-code** in #2357 (which closed #2342). The fix did **not** cover codex, and #3221 (closed by #3231) explicitly listed the codex extension as **out of scope** ("Larger; out of scope for a small fix … Extend `confirmActive` to non-safe harnesses … so the rescue also works for Codex"). No issue currently tracks the codex remainder — this fills that gap.

## Root cause

#2357 added `confirmActive` — a bounded rescue loop that re-sends `Enter` until the session reports `active` (flipped by the user-prompt-submit hook). That loop is what makes large multiline pastes actually submit.

But `confirmActive` is **deliberately skipped** for codex. The guard requires the harness to emit **both** a submit signal **and** a blocked signal:

```go
// backend/internal/session_manager/manager.go:2122
func (m *Manager) harnessNudgeSafe(harness domain.AgentHarness) bool {
	...
	return ok && s.EmitsSubmitActivity() && s.EmitsBlockedActivity()
}
```

Codex opts out at the second condition:

```go
// backend/internal/adapters/agent/codex/codex.go:44-49
// EmitsBlockedActivity is false: codex reports permission prompts as
// waiting_input — it installs no post-tool-use hook, so a blocked state could
// never be cleared mid-turn. confirmActive must not nudge it (an Enter could
// answer a pending decision it cannot report as blocked).
func (p *Plugin) EmitsBlockedActivity() bool { return false }
```

So at `manager.go:2069` `harnessNudgeSafe("codex")` returns `false` and `confirmActive` is never called for a codex session. When a single `Enter` fails to submit a large multiline paste (the TUI needs more settle time than the fixed `defaultEnterDelay = 300ms` paste→Enter pause), there is **no rescue path at all** for codex. The comment block at `manager.go:2051-2057` documents this explicitly: *"Only claude-code and its hook-delegators (grok/continueagent/devin) satisfy both; every other harness opts out."*

The mitigation that #3231 added (detached chunk-loop context so a client cancel mid-paste can't strand chunks) applies to all harnesses including codex and makes the paste more robust, but it only addresses the cancellation edge — it does nothing for the "Enter didn't submit because the TUI was still absorbing the paste" case, which is what this report describes.

## Repro

1. Start a codex worker session through AO (desktop or `ao start`).
2. Send a large **multiline** prompt via the GUI input box / `ao send --session <id> --message "<~25+ line prompt>"`.
3. Intermittently, the prompt text stays in the codex composer as an unsubmitted draft. `ao send` returns 0 immediately; no error, no warning.
4. The next send appends to the same draft and its trailing `Enter` submits **both** messages together.

Timing-sensitive (depends on how saturated the TUI input buffer is when chunks land), same race as #3221 — but unlike claude-code, codex has no `confirmActive` to recover.

## Why codex opts out (the hard part)

Codex reports a permission/approval prompt as plain `waiting_input`, indistinguishable from an idle prompt, and installs no post-tool-use hook — so AO can't tell "draft waiting to be submitted" from "paused on a decision the user must make." Re-sending `Enter` to rescue a stranded draft could therefore **answer a permission dialog on the user's behalf** — an unrecoverable action. That asymmetry is exactly why #2357 gated the rescue behind `EmitsBlockedActivity()`. Any fix here must preserve that safety property.

## Suggested fix surfaces (not a patch)

1. **Distinguish unsubmitted-draft from permission-dialog for codex.** If codex can be made to surface a distinct "blocked/awaiting decision" signal (a hook, or terminal-content heuristics gated tightly enough to avoid false positives), it could opt into `EmitsBlockedActivity()` and inherit the existing `confirmActive` path unchanged. This is the cleanest long-term fix but depends on codex-side support.
2. **A codex-specific, content-verified re-Enter.** A narrower rescue that, before re-sending `Enter`, reads back the pane and confirms the composer still contains the just-pasted text (a draft) and is *not* showing a known permission/confirmation dialog. More fragile than (1) but doesn't require codex cooperation. Must fail closed (no Enter) on any ambiguity.
3. **At minimum: stop failing silently.** Emit a structured `WARN` (session id, harness, observed state, budget) when a codex send completes without observing `active`, so operators have a signal even though the send returns 200. #3231 already added this log for the budget-exhaustion case on safe harnesses; extending it to the "rescue not run because harness is not nudge-safe" case would surface this failure mode today.

## Related

- #2357 — the claude-code fix (added `confirmActive`); live-verified on codex per its body, but the verification did not exercise the no-rescue path.
- #2342 — original multiline-doesn't-submit report (claude-code).
- #3221 — deeper follow-up; closed by #3231, which fixed the chunk-loop cancellation hole for **all** harnesses but explicitly left codex's `confirmActive` rescue out of scope (suggested-fix point 4).
- `backend/internal/adapters/agent/codex/codex.go:44` (`EmitsBlockedActivity = false`).
- `backend/internal/session_manager/manager.go:2051-2071` (rescue gated on `harnessNudgeSafe`), `:2108-2123` (`harnessNudgeSafe` definition).

athrvd7:
Fixes #3626.

This keeps Codex safe from a second, potentially destructive Enter while giving multiline and multi-frame terminal pastes longer to settle before their original Enter is sent.

- Short single-line messages retain the configured 300 ms delay.
- Multiline or more-than-512-rune messages use a 1 s delay plus 150 ms per extra chunk, capped at 2 s.
- Longer configured tmux delays are preserved.
- Empty Enter-only nudges remain immediate.
- Unix tmux and Windows ConPTY use the same policy.

Verification:
- `go test ./internal/adapters/runtime/inputdelay ./internal/adapters/runtime/conpty ./internal/adapters/runtime/tmux`
- `go test -race ./internal/adapters/runtime/inputdelay ./internal/adapters/runtime/conpty ./internal/adapters/runtime/tmux`
- `go build ./...`

`go test ./...` reached unrelated, pre-existing 3-second auth-probe timeouts in the Kilo Code and OpenCode adapter packages. The affected runtime packages passed.

### DIFF
--- a/backend/internal/adapters/runtime/conpty/client.go
+++ b/backend/internal/adapters/runtime/conpty/client.go
@@ -9,6 +9,8 @@ import (
 	"net"
 	"syscall"
 	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/runtime/inputdelay"
 )
 
 const (
@@ -60,12 +62,13 @@ func clientSendMessage(addr, message string) error {
 		}
 	}
 
-	// Brief pause before Enter (matches TS: Enter sent as a separate frame).
+	// Let large or multiline pastes settle before Enter. This widens the first
+	// Enter only, rather than sending a risky retry into a Codex decision prompt.
 	// Skipped for an empty message — an Enter-only nudge has no paste to let
 	// settle, and the pause would only widen the guard-read→Enter window
 	// (mirrors the tmux runtime's enterDelay contract).
-	if len(runes) > 0 {
-		time.Sleep(ptyInputEnterDelay)
+	if delay := inputdelay.ForMessage(message, ptyInputEnterDelay); delay > 0 {
+		time.Sleep(delay)
 	}
 	frame, err := EncodeMessage(MsgTerminalInput, []byte("\r"))
 	if err != nil {
--- a/backend/internal/adapters/runtime/inputdelay/inputdelay.go
+++ b/backend/internal/adapters/runtime/inputdelay/inputdelay.go
@@ -1,0 +1,40 @@
+// Package inputdelay computes the pre-Enter pause after terminal text input.
+package inputdelay
+
+import (
+	"strings"
+	"time"
+	"unicode/utf8"
+)
+
+const (
+	chunkRunes = 512
+	burstDelay = time.Second
+	burstStep  = 150 * time.Millisecond
+	maxDelay   = 2 * time.Second
+)
+
+// ForMessage returns the time to let a pasted message settle before Enter.
+// Empty messages are intentional Enter-only nudges and must remain immediate.
+// Multiline and chunked pastes need longer than the ordinary delay, but this
+// never retries Enter, so it cannot approve a prompt that was not submitted.
+func ForMessage(message string, configured time.Duration) time.Duration {
+	if message == "" {
+		return 0
+	}
+
+	runes := utf8.RuneCountInString(message)
+	if !strings.Contains(message, "\n") && runes <= chunkRunes {
+		return configured
+	}
+
+	chunks := (runes + chunkRunes - 1) / chunkRunes
+	delay := burstDelay + time.Duration(chunks-1)*burstStep
+	if delay > maxDelay {
+		delay = maxDelay
+	}
+	if configured > delay {
+		return configured
+	}
+	return delay
+}
--- a/backend/internal/adapters/runtime/inputdelay/inputdelay_test.go
+++ b/backend/internal/adapters/runtime/inputdelay/inputdelay_test.go
@@ -1,0 +1,31 @@
+package inputdelay
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestForMessage(t *testing.T) {
+	tests := []struct {
+		name       string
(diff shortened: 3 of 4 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@athrvd7 Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

i-trytoohard:
@athrvd7 Closing as stale: no activity from you for 30 days. The branch stays safe to reopen if you pick this up again. Thanks for the PR!

Issue #3749, opened 2026-08-08
bug(chat): "Stop turn" fails with CHAT_NO_ACTIVE_TURN while UI shows a running turn — in-memory/durable turn-state desync


i-trytoohard:
## Bug

Clicking **"Stop turn"** on a chat session that shows a running turn ("Working") returns `CHAT_NO_ACTIVE_TURN` — *"there is no turn in flight to interrupt"* — and the turn is **not** stopped. The UI continues showing "Working" indefinitely. The user has no way to stop the turn from the chat interface.

**Source:** [Discord — #bug-triage](https://discord.com/channels/1476302178913357958/1491735678156013588/1535762769390080002) | **Reported by:** @prateek | **Date:** 2026-08-09
**Analyzed against:** `b6609ae61` (origin/main)
**Confidence:** High — code path traced end-to-end; the desync mechanism is identifiable in source

### What the screenshot shows

A chat session with:
- Previous turns marked "DONE" and "STOPPED 48.0s"
- A **"Working 3m 0s"** status bar with a spinner and a **"Stop turn"** button
- After clicking "Stop turn": red error `there is no turn in flight to interrupt (CHAT_NO_ACTIVE_TURN)`
- The "Working" bar **remains visible** — the turn is not stopped, the UI does not recover

## Reproduction

1. Start a chat-mode session (any harness with Chat support)
2. Send a message that triggers a long-running turn
3. While the turn is running, click "Stop turn"
4. Observe: error `CHAT_NO_ACTIVE_TURN`, turn continues running, UI stays stuck on "Working"

*Note: this is intermittent — it requires the in-memory/durable desync condition (see Root Cause). Most interrupts work. The bug manifests when a `ChatEventTurnCompleted` projection fails mid-transaction.*

## Root Cause

**The in-memory controller state (`pendingTurnID`) and the durable SQLite turn state can diverge**, producing a split where the UI (reads SQLite) shows a running turn but the interrupt path (checks in-memory) finds nothing to cancel.

### The two sources of truth

| Signal | Source | Used by |
|--------|--------|---------|
| `c.pendingTurnID` | In-memory, `Controller` struct | `Interrupt()` → `awaitAcknowledgedTurn()` |
| Turn `state = 'running'` | SQLite `conversation_turns` table | `Snapshot()` → frontend `activeTurn()` → "Working" bar |

### The desync mechanism

[...]

PR #3752, opened 2026-08-09
fix(chat): reconcile turn-state desync when Stop turn fails with CHAT_NO_ACTIVE_TURN

bluwwi:
Fixes [[#3749](https://github.com/Untrivial-ai/agent-orchestrator/issues/3749) — "Stop turn" fails with `CHAT_NO_ACTIVE_TURN` while the UI shows a running turn.

The chat controller kept two sources of truth for "is a turn running?" — an in-memory `pendingTurnID` and a durable SQLite `conversation_turns.state = 'running'` row. They could diverge when `SettleTurn` failed inside the `ProjectProviderEvent` transaction: the rollback discarded the SQLite update but the in-memory mutation had already run, leaving the UI showing "Working" with no way to stop it.

## Root Cause
`Controller.apply()` for `ChatEventTurnCompleted` mutated `pendingTurnID = ""` in memory *before* calling `SettleTurn()`, which runs inside a SQLite transaction. If `SettleTurn` failed, the transaction rolled back (SQLite retained `state = 'running'`) but the in-memory `pendingTurnID` was already cleared — not rolled back because it's a Go variable, not SQL. This desync meant `Interrupt()` found nothing to cancel while the UI read SQLite and showed "Working".

## Fix
1. **Reorder `apply()`** — durable write before memory mutation. If `SettleTurn` fails, memory stays intact so `Interrupt()` still works. (`controller.go`)
2. **Disk fallback in `Interrupt()`** — if memory says nothing is running, check SQLite for a stale `running` turn and settle it as `interrupted`. Mirrors the existing `SettleOrphanedTurns` reconciliation pattern. (`controller.go` + `conversation_store.go` + `conversations.sql`)
3. **Frontend resilience** — `onError` handler on the `interrupt` mutation invalidates the conversation query so the UI refetches and unblocks. (`useConversation.ts`)

### Test 1: Normal "Stop turn" still works (regression check)
1. Start the daemon: `go run ./cmd/ao daemon`
2. Start the desktop app: `cd frontend && npm run dev`
3. Open a chat session (codex, Chat mode)
4. Send a slow prompt:
   > Write a comprehensive test suite for a banking application in Go. Include tests for account creation, deposits, withdrawals, transfers, overdrafts, concurrency with goroutines, transaction rollback, audit logging, and edge cases. Cover both unit and integration tests.
5. While it shows "Working Xs", click "Stop turn"
6. **Result:** Turn stops, status shows "Stopped" — no regression

Screenshot 2026-08-09 114347

### Test 2: Stale "Working" bar recovers after daemon restart (the bug)
1. Start the daemon standalone: `go run ./cmd/ao daemon`
2. Start the desktop app: `cd frontend && npm run dev`
3. Open a chat session (codex, Chat mode)
4. Send the same slow prompt
5. While it shows "Working Xs", **kill the daemon** (Ctrl+C in the terminal)
6. The UI keeps showing "Working" (stale SQLite state — turn still `running` on disk)
7. **Restart the daemon:** `go run ./cmd/ao daemon`
8. Click "Stop turn"
9. **Before the fix:** Error `there is no turn in flight to interrupt (CHAT_NO_ACTIVE_TURN)`, UI stays stuck on "Working" indefinitely
10. **After the fix:** Turn settles as interrupted, UI unblocks

Screenshot 2026-08-09 120414

## Automated Tests
- `TestInterruptReconcilesStaleRunningTurnOnDisk` — a turn left `running` on disk while the controller has no `pendingTurnID` is settled as `interrupted` by `Interrupt()`, unblocking the UI.
- `TestInterruptReturnsNoActiveTurnWhenNoRunningTurnAnywhere` — the disk fallback does not produce false positives when memory and SQLite agree there is nothing to stop.
- `go test ./internal/service/chat/...` — 79/79 pass (77 existing + 2 new)
- `go vet ./...` — clean
- `npm run typecheck` — clean

### DIFF
--- a/backend/internal/service/chat/controller.go
+++ b/backend/internal/service/chat/controller.go
@@ -33,6 +33,7 @@ type Store interface {
 	CreateConversation(ctx context.Context, id string, scope domain.ConversationScope, project domain.ProjectID, session domain.SessionID, now time.Time) (domain.ConversationRecord, error)
 	ConversationForSession(ctx context.Context, session domain.SessionID) (domain.ConversationRecord, error)
 	ClaimChatControllerGeneration(ctx context.Context, session domain.SessionID, generation string, now time.Time) error
+	GetRunningTurn(ctx context.Context, conversationID string) (id, providerTurnID string, ok bool, err error)
 
 	AdoptProviderTurn(ctx context.Context, conversationID string, session domain.SessionID, generation, turnID, providerTurnID string, now time.Time) error
 	AppendImportedUserMessage(ctx context.Context, conversationID, providerTurnID string, msg domain.ConversationMessage, now time.Time) error
@@ -1036,7 +1037,24 @@ const turnAckWait = 3 * time.Second
 func (c *Controller) Interrupt(ctx context.Context) error {
 	turn, ok := c.awaitAcknowledgedTurn(ctx)
 	if !ok {
-		return ErrNoActiveTurn
+		// Memory says nothing is running, but check SQLite — a prior
+		// SettleTurn failure may have left a turn in 'running' state
+		// while the in-memory pendingTurnID was already cleared.
+		_, providerTurnID, running, err := c.store.GetRunningTurn(ctx, c.conversation.ID)
+		if err != nil {
+			return fmt.Errorf("check running turn: %w", err)
+		}
+		if !running {
+			return ErrNoActiveTurn
+		}
+		// The turn is stale from the controller's perspective — the provider
+		// already emitted a completion event that failed to project. Settle
+		// it as interrupted so the UI unblocks, matching the user's intent.
+		if err := c.store.SettleTurn(ctx, c.conversation.ID, providerTurnID,
+			domain.TurnStateInterrupted, "reconciled by interrupt (stale running turn)", c.now()); err != nil {
+			return fmt.Errorf("reconcile stale running turn: %w", err)
+		}
+		return nil
 	}
 
 	c.mu.Lock()
@@ -1308,6 +1326,18 @@ func (c *Controller) apply(ctx context.Context, event ports.ChatEvent) error {
 			}, now)
 
 	case ports.ChatEventTurnCompleted:
+		message := ""
+		if event.Err != nil {
+			message = event.Err.Error()
+		}
+		state := event.TurnState
+		if state == "" {
+			state = domain.TurnStateFailed
+		}
+		if err := c.store.SettleTurn(
+			ctx, c.conversation.ID, event.ProviderTurnID, state, message, now); err != nil {
+			return err
+		}
 		c.mu.Lock()
 		if c.pendingTurnID == event.ProviderTurnID {
 			c.pendingTurnID = ""
@@ -1317,19 +1347,6 @@ func (c *Controller) apply(ctx context.Context, event ports.ChatEvent) error {
 		}
 		c.state = ports.ChatControllerReady
 		c.mu.Unlock()
-		message := ""
-		if event.Err != nil {
-			message = event.Err.Error()
-		}
-		state := event.TurnState
-		if state == "" {
-			// A completion with no status is not evidence of success.
-			state = domain.TurnStateFailed
-		}
-		if err := c.store.SettleTurn(
-			ctx, c.conversation.ID, event.ProviderTurnID, state, message, now); err != nil {
-			return err
-		}
 		return nil
 
 	case ports.ChatEventMessageDelta:
--- a/backend/internal/service/chat/controller.go
+++ b/backend/internal/service/chat/controller.go
@@ -1034,27 +1034,48 @@ const turnAckWait = 3 * time.Second
 // the next message instead of stopping would be the opposite of what the button
 // says. The cutoff is recorded before the provider call, so a completion that
 // races back cannot drain the queue the interrupt was about to cancel.
+//
+// Two recovery paths exist for when memory and SQLite disagree:
+//
+//   - The provider says ErrChatNoActiveTurn for a turn memory still tracks. A
+//     prior tx.Commit failure rolled back the SettleTurn write while memory was
+//     already cleared (before the fix that moved clearing to afterProject), or
+//     the turn completed between awaitAcknowledgedTurn and the provider call.
+//     reconcileDurableTurn settles the durable row, clears memory, cancels the
+//     queue, and reports idle — the full Interrupt contract without a second
+//     provider send.
+//   - Memory says nothing is running but SQLite has a turn in 'running'. This
+//     is the dispatch race: BindTurnToProvider wrote 'running' before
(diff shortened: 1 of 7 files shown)


backend/internal/service/chat/controller.go
```diff
@@ -1036,7 +1037,24 @@ const turnAckWait = 3 * time.Second
 func (c *Controller) Interrupt(ctx context.Context) error {
 	turn, ok := c.awaitAcknowledgedTurn(ctx)
 	if !ok {
-		return ErrNoActiveTurn
+		// Memory says nothing is running, but check SQLite — a prior
```
illegalcall: **[P1] This fallback is skipped by the exact failure described in #3749.** With the new ordering, a `SettleTurn` error returns before `pendingTurnID` is cleared, so `awaitAcknowledgedTurn` returns `ok=true` and this `!ok` branch is never entered. The provider has already emitted `ChatEventTurnCompleted`, so `c.conv.Interrupt` can return `ErrChatNoActiveTurn`; lines 1070–1072 then return the same user-visible error without settling SQLite. The desktop refetch simply reloads `state='running'`. Please reconcile a still-running durable turn when the provider authoritatively says it is gone, and add a test that injects the projection failure plus provider refusal rather than seeding an unrelated memory-empty state.

backend/internal/service/chat/controller.go
```diff
@@ -1036,7 +1037,24 @@ const turnAckWait = 3 * time.Second
...
+		if err != nil {
+			return fmt.Errorf("check running turn: %w", err)
+		}
+		if !running {
+			return ErrNoActiveTurn
+		}
+		// The turn is stale from the controller's perspective — the provider
+		// already emitted a completion event that failed to project. Settle
+		// it as interrupted so the UI unblocks, matching the user's intent.
+		if err := c.store.SettleTurn(ctx, c.conversation.ID, providerTurnID,
```
illegalcall: **[P1] This successful recovery does not implement the full Interrupt contract.** Normal Interrupt records `cancelQueuedAt`, and the post-completion path cancels queued turns and reports lifecycle idle. This branch only settles the running row. If a second prompt is queued, `activeTurn()` falls back to that queued row, so the UI still shows Working and a second Stop returns no-active; the old prompt can also be dispatched by a later drain. Please settle/cancel the queue at the stop cutoff and report idle after durable success. The focused test should include a queued prompt and assert both terminal states, no extra provider send, and an idle signal.

backend/internal/service/chat/controller.go
```diff
@@ -1036,7 +1037,24 @@ const turnAckWait = 3 * time.Second
...
+		// SettleTurn failure may have left a turn in 'running' state
+		// while the in-memory pendingTurnID was already cleared.
+		_, providerTurnID, running, err := c.store.GetRunningTurn(ctx, c.conversation.ID)
+		if err != nil {
+			return fmt.Errorf("check running turn: %w", err)
+		}
+		if !running {
+			return ErrNoActiveTurn
+		}
+		// The turn is stale from the controller's perspective — the provider
```
illegalcall: **[P1] A disk-only running row is not proof that the turn is stale.** `dispatch()` calls `BindTurnToProvider` (making SQLite `running`) before assigning `pendingTurnID` at lines 737–749. A concurrent Stop in that interval reaches this branch, marks genuinely live work interrupted, returns success without calling the provider, and then dispatch publishes the pending id while the provider continues. Please synchronize this decision with dispatch or attempt provider cancellation using the durable provider id before reconciling. A deterministic test can pause after the durable bind and verify Stop cannot succeed without reaching the provider.

backend/internal/service/chat/controller.go
```diff
@@ -1308,28 +1326,27 @@ func (c *Controller) apply(ctx context.Context, event ports.ChatEvent) error {
...
 		state := event.TurnState
 		if state == "" {
-			// A completion with no status is not evidence of success.
 			state = domain.TurnStateFailed
 		}
 		if err := c.store.SettleTurn(
 			ctx, c.conversation.ID, event.ProviderTurnID, state, message, now); err != nil {
 			return err
 		}
+		c.mu.Lock()
```
illegalcall: **[P1] This is still before the actual transaction commit.** `apply` is the callback passed to `ProjectProviderEvent`; after this memory mutation returns, `tx.Commit()` occurs in `conversation_store.go:1168`. If commit fails, the event archive and settled row roll back while `pendingTurnID` remains cleared, recreating the same split-brain. This also misses #3749's explicit recommendation to defer these mutations outside the transaction callback. Please move the memory transition to a post-commit phase (for example the existing `afterProject` boundary) and prove it with an injected commit failure.

frontend/src/renderer/hooks/useConversation.ts
```diff
@@ -214,6 +214,9 @@ export function useConversationCommands(sessionId: string | undefined) {
 			if (error) throw error;
 		},
 		onSuccess: invalidate,
+		onError: () => {
```
illegalcall, line 217: **[P2] The error-refresh resilience is desktop-only.** Mobile uses the same conversation/interrupt API, but `packages/mobile/lib/chat/useConversation.ts:237–258,327` refreshes only after successful actions; its catch path records the error without reloading the snapshot. A stale mobile Working bar therefore remains after `CHAT_NO_ACTIVE_TURN` until an unrelated SSE/foreground/manual refresh. Please mirror the error refresh in mobile and add a focused action/hook test.

illegalcall (changes requested):
## Standards

The sqlc source/generated pairing and package boundaries are sound. The blocking standards problems are that controller memory still changes before the real transaction commit, and the recovery branch does not preserve the documented Interrupt contract.

## Spec

Requesting changes because the exact #3749 `SettleTurn`-failure sequence still returns `CHAT_NO_ACTIVE_TURN`, while the new disk fallback can also acknowledge Stop without stopping live provider work or clearing the queue. The five inline comments give the triggers, incorrect outcomes, and focused proof needed.

Summary: 3 Standards findings (worst: pre-commit memory mutation) and 5 Spec findings (worst: the reported failure path remains reproducible).

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Closing this as superseded by [PR #3992](https://github.com/Untrivial-ai/agent-orchestrator/pull/3992).

The focused replacement was adapted from this work and preserves @bluwwi's co-author credit in the commit. It carries the complete recovery contract requested in review—durable fallback, provider-refusal reconciliation, queued-turn cancellation, post-commit memory transitions, desktop/mobile error refetch, and the rollback/desync regression coverage—and all CI checks are green.

It is stacked on [PR #3890](https://github.com/Untrivial-ai/agent-orchestrator/pull/3890), so we only maintain one implementation for [issue #3749](https://github.com/Untrivial-ai/agent-orchestrator/issues/3749). Thank you for the original diagnosis and implementation.

PR #3754, opened 2026-08-09
feat: add Kimchi ACP chat driver binding (working MVP)

vytautas-petrikas:
## What

Binds Kimchi to AO's reusable ACP Chat transport, following the same `nativeacp` pattern as OpenCode and Droid. Users can now spawn Kimchi sessions in chat mode and interact through the mobile/desktop chat UI instead of a terminal.

## Why

AO could only supervise Kimchi through TUI mode (terminal scraping). This meant Kimchi was absent from the chat driver list — users who wanted the native chat experience on mobile had to use Claude Code, Codex, OpenCode, or Droid. This PR adds Kimchi to that list.

### Kimchi chat driver binding (`kimchiacp/`)

A new `kimchiacp` package delegates to `nativeacp.New()` with four provider-specific functions:

- **Harness:** `domain.HarnessKimchi`
- **`configure()`:** returns `["--mode", "acp"]`, appends `--append-system-prompt <text>` when AO provides a system prompt
- **`sessionMode()`:** maps `AcceptEdits → "auto"`, `PermissionModeAuto → "auto"` (matching the Kimchi TUI adapter), `BypassPermissions → "yolo"`, `Default → ""`
- **`sessionOptions()`:** maps `ChatTurnSettings.Model` to the `"model"` config option ID

Binary resolution and auth probing are delegated to the existing Kimchi agent plugin. Registered in the chat driver registry's `Build()` function.

### ACP driver compatibility fixes (`acp/`)

Three issues were discovered and fixed during the manual smoke test against a live Kimchi ACP session:

1. **Removed `session/resume` hard gate from `Start()`.** Kimchi's ACP doesn't advertise `session/resume` in its handshake. The driver was hard-rejecting it with `409: ACP agent does not support session/resume`. The fix allows new sessions to start without resume; the resume capability is downgraded to `false`. `Resume()` still returns `ErrChatResumeFailed` if called.

2. **Fixed nil map panic.** `c.capabilities` was only initialized in `start()`, but Kimchi sends `session/update` notifications (with config options) during `NewSession()` — before `start()` runs. This caused `panic: assignment to entry in nil map` in `replaceConfigOptions()`. Fixed by initializing `capabilities` as an empty map in `newConversation()`.

3. **Graceful handling of unsupported ACP methods.** Kimchi's ACP doesn't implement `session/set_mode` (returns JSON-RPC `-32601 Method not found`). The driver was treating this as a fatal error. Added `isACPMethodNotFound()` helper; `applyTurnSettings()` now logs unsupported methods at debug level and continues. Same treatment applied to `session/set_config_option`.

### Mobile logo registration (`packages/mobile/`)

Added Kimchi's logo (rasterized from the desktop SVG at 128×128 PNG) to the mobile app's logo registry and test fixtures.

## Testing

- `go build ./...` and `go test ./internal/adapters/chatdriver/...` — all pass (218+ tests across 10 packages)
- Binding unit tests (`kimchiacp/driver_test.go`) verify `configure()`, `sessionMode()`, and `sessionOptions()` as pure functions
- Registry test (`TestShippedChatDrivers`) asserts `domain.HarnessKimchi` resolves to a non-nil driver
- Manual smoke test: spawned a Kimchi chat session from the mobile app against a live AO daemon — Kimchi launched in ACP mode, the session established, streaming messages and tool calls rendered correctly in the chat UI

## ACP compatibility gaps (follow-up work)

This is a working MVP. Kimchi's ACP server implements a subset of the protocol — the following capabilities are missing on the harness side and will be addressed in follow-up PRs:

| Capability | Status | Notes |
|---|---|---|
| `session/resume` | Not advertised | New sessions work; resume after daemon restart is not available |
| `session/set_mode` | Not implemented (`-32601`) | Permission mode cannot be changed at runtime; launch-time mode stays in effect |
| `session/load` | Not implemented | Transcript replay on resume is unavailable |
| Usage reporting | Not emitted | No token/cost tracking |
| Per-turn diffs | Not attached to tool calls | No diff display in chat UI |
| Plans | No plan events | No plan/todo tracking |
| Steering | Extension not implemented | Cannot inject guidance into a running turn |

I will work on adding support for these missing capabilities on the Kimchi side.

## Checklist

- [x] Branched from `main`

image


backend/internal/adapters/chatdriver/kimchiacp/driver.go
```diff
@@ -0,0 +1,51 @@
...
+func New(plugin nativeacp.Plugin, log *slog.Logger) ports.ChatDriver {
+	return nativeacp.New(plugin, nativeacp.Config{
+		Harness:        domain.HarnessKimchi,
+		Configure:      configure,
+		SessionMode:    sessionMode,
+		SessionOptions: sessionOptions,
+	}, log)
+}
+
+func configure(cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
```
nikhilachale, line 27: P1 - Kimchi ACP ignores requested model and permission mode when ACP setters are unsupported

configure() launches only kimchi --mode acp plus the system prompt. It does not
  pass cfg.Model or cfg.Permissions as launch-time flags.

  That matters because the new ACP fallback in backend/internal/adapters/chatdriver/acp/conversation.go:253 silently skips session/
  set_config_option on -32601, and similarly skips session/set_mode at line 228. The PR body says Kimchi does not implement session/set_mode, and
  the code now treats session/set_config_option as optional too. So a user starting Kimchi Chat with model=..., auto, or bypass/yolo can get a
  successful session that actually runs with Kimchi defaults.

  The existing Kimchi TUI adapter already applies these at process launch: --model at backend/internal/adapters/agent/kimchi/kimchi.go:219, and
  --auto / --yolo at backend/internal/adapters/agent/kimchi/kimchi.go:337. The ACP binding should preserve that launch-time behavior, especially
  since runtime setters are degraded.

illegalcall (approved):

illegalcall:
Thanks — the fresh-session mobile smoke is useful. To call this **basic AO Chat support** (as opposed to the optional usage/diff/plan/steering work), I think these items are required before merge:

1. **Make Chat admission and restart behavior truthful.**
   - Kimchi currently does not advertise `sessionCapabilities.resume`, and its docs say each ACP session starts fresh.
   - `nativeacp.Probe()` nevertheless reports `resume: true`, so AO's production-floor check passes; `Start()` only downgrades the capability after initialization, and a later daemon restore fails in `Resume()`.
   - Given AO's current Chat contract, the preferred fix is to implement/advertise `session/resume` in Kimchi and retain the ACP start-time resume gate. `session/load` is optional if resume restores the same model context.
   - If fresh-only Chat is intentionally desired, it needs a separate explicit non-resumable/ephemeral contract across `ports`, `service/chat`, `session_manager`, and both UIs. Silently admitting it as ordinary Chat is not sufficient.
   - Acceptance: send a sentinel fact, restart the AO daemon/controller, resume the stored provider session ID, and prove the next turn remembers the sentinel without duplicating AO history.

2. **Make initial model and permission settings actually reach Kimchi.**
   - `kimchiacp.configure()` currently passes only `--mode acp` and the system prompt. It needs to pass the supported initial `--model`, `--auto`, and `--yolo` flags from `LaunchConfig`.
   - Do not globally swallow JSON-RPC `-32601` for `session/set_mode` or `session/set_config_option`. If AO explicitly requested a setting and the agent cannot apply it, return an actionable incompatibility/error or remove that control. A successful turn must never run with a different permission mode from the one AO displays.
   - In particular, `Default -> ""` is not a reversible mode transition: after `yolo`, selecting default currently sends nothing.

3. **Use one authoritative live mode/settings path and keep it synchronized.**
   - Preferred ACP path: Kimchi advertises a select config option with `category: "mode"`, accepts `session/set_config_option`, returns the complete post-change catalog, and sends complete `config_option_update` catalogs for Kimchi-initiated model/mode changes.
   - Legacy fallback: AO must consume `NewSessionResponse.Modes`/resume modes, call `session/set_mode`, and handle `current_mode_update`. Those fields exist in the pinned ACP SDK but are currently ignored by `acp/driver.go` and `acp/client.go`.
   - Preserve config updates Kimchi emits during `session/new`; `conversation.start()` must not overwrite the only early catalog with an empty response catalog.
   - Acceptance: model selection plus `default -> plan -> auto -> yolo -> default` all work, and provider-initiated model/mode changes appear in desktop and mobile.

4. **Do not hide the only permission control when Kimchi advertises a model option.**
   - Desktop currently sets `onChooseSettings={undefined}` whenever any `config_options` capability exists.
   - Mobile similarly hides model, effort, and Approvals whenever any provider option exists.
   - A model-only catalog therefore removes Approvals. Ideally Kimchi should advertise a complete model-and-mode provider catalog; AO should still handle partial catalogs without showing a control it cannot apply or hiding the only valid one.

5. **Gate and prove the supported Kimchi build.**
   - Add a bounded `kimchi --version` check for the documented `0.0.7+` minimum (or a newer tested minimum), plus malformed/old/timeout tests.
   - Verify authentication failures normalize to AO's actionable auth error.
   - Add shared ACP behavior tests for no-resume admission, early config updates, mode/config `-32601`, catalog replacement, modes/current-mode updates, and non-`-32601` errors. The current tests cover only pure mappings and registry presence.
   - Add `kimchiacp/live_test.go`, guarded like the existing live ACP tests, covering initialize, fresh session, streaming, a tool approval/denial, settings changes, cancellation, and resume.
   - Add one daemon-level restart E2E; the adapter live test alone does not prove AO lifecycle recovery.

6. **Fail explicitly for unsupported setup input.**
   - Kimchi documents per-session MCP servers as unsupported. A non-empty AO per-session MCP configuration must either be supported or fail clearly; it must not be silently ignored. Capture and test additional-directory and image capabilities from the live handshake as well.

Not blockers for basic Chat once capabilities are reported honestly: usage/cost, rendered diffs, plan events, steering, audio, most slash commands, and `session/load` transcript replay. The blockers are continuity under AO's current contract, truthful permission/model application, synchronized provider state, and behavior-level proof.

Primary references: [Kimchi ACP guide](https://docs.kimchi.dev/docs/coding-acp), [ACP session setup](https://agentclientprotocol.com/protocol/session-setup), and [ACP session config options](https://agentclientprotocol.com/protocol/v1/session-config-options).

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/chatdriver/kimchiacp/driver.go
vytautas-petrikas: Fixed in d680af8 + d233f88.

**Launch-time flags:** `configure()` now passes `--model`, `--auto`, and `--yolo` from `LaunchConfig` — same mapping as the TUI adapter (`kimchi.go:220,337-346`). Verified against a live session: the Kimchi process launches with `kimchi --mode acp --yolo --append-system-prompt ...` when bypass mode is selected.

**Runtime mode changes:** Kimchi does not implement `session/set_mode` (returns `-32601`). Instead of calling it and swallowing the error, I removed `SessionMode` from the Kimchi binding entirely. Permission mode changes now flow through `session/set_config_option` with `configId: "permissions-mode"` — which Kimchi implements (`server.ts:380-398`, handles `case "permissions-mode"` → `doSetPermissionMode`). The `sessionOptions()` function maps AO's permission vocabulary onto Kimchi's native mode ids (`default → ""`, `accept-edits/auto → "auto"`, `bypass-permissions → "yolo"`).

**`-32601` handling:** `applyTurnSettings` no longer silently swallows `-32601`. It returns a typed `ErrACPSetterUnsupported` error. `Start()`/`Resume()` tolerate it (launch-time flags already applied the initial settings). `SendTurn()` propagates it so a runtime change that can't be applied surfaces to the user.

**`Default -> ""` reversibility:** `Default` sends no config option and no launch flag — Kimchi stays in its current/default mode. This is consistent and reversible: selecting `auto` after `default` sends `permissions-mode: auto`, selecting `default` after `yolo` sends nothing (Kimchi returns to its baseline).

vytautas-petrikas:
Thanks for the thorough review. Pushed in d680af8 + d233f88. Here's where each item stands:

**1. Session/resume — not addressed (Kimchi-side).** Confirmed in Kimchi source: `initialize()` returns `sessionCapabilities: { list: {}, close: {} }` with no `resume` field. The preferred fix (implement/advertise `session/resume` in Kimchi) is harness-side work I'll pick up separately. The AO-only alternative (non-resumable chat contract across `ports`, `service/chat`, `session_manager`, and both UIs) is a larger architectural change I'd rather not block this PR on.

**2. Initial model and permission settings — fixed.** `configure()` passes `--model`, `--auto`, `--yolo` from `LaunchConfig`. `-32601` is no longer silently swallowed. Permission mode is routed through `session/set_config_option` with `permissions-mode` (which Kimchi implements) instead of the unsupported `session/set_mode`. See the inline reply above for details.

**3. Authoritative live mode/settings path — fixed.** The preferred ACP path works: Kimchi advertises a `permissions-mode` config option with `category: "mode"`, accepts `session/set_config_option`, and returns the complete post-change catalog. AO routes mode changes through this path. I also fixed early config option preservation: `conversation.start()` no longer overwrites config options received via `session/update` during `session/new` when the response carries an empty catalog. Test: `TestACPDriverPreservesEarlyConfigOptionUpdates`.

**4. Don't hide Approvals — fixed.** Desktop (`SessionChatSurface.tsx`) and mobile (`ChatSettingsModal.tsx`) now check per-dimension: native Approvals are hidden only when the provider catalog has `category: "mode"`, native Model/Effort only when it has `category: "model"`. A model-only catalog no longer hides the Approvals control.

**5. Version gate and tests — addressed.** Added `VersionProbe` to `nativeacp.Config`; the Kimchi binding runs `kimchi --version`, parses semver, and rejects < 0.0.7 with `ErrChatDriverIncompatible`. Auth error normalization was already handled by the existing `isACPAuthRequired`/`normalizeACPError` pipeline. Added tests: `TestACPDriverStartToleratesMethodNotFound`, `TestACPDriverSendTurnPropagatesMethodNotFound`, `TestNormalizeMCPServers*`, `TestACPDriverPreservesEarlyConfigOptionUpdates`. Added `kimchiacp/live_test.go` (guarded by `AO_LIVE_KIMCHI_ACP=1`) covering Probe, Start, SendTurn, streaming, and completion — same pattern as the existing Droid/Claude/OpenCode live tests. Daemon restart E2E not added; it requires infrastructure beyond this adapter's test scope.

**6. Fail explicitly for unsupported MCP — fixed.** `normalizeMCPServers` now returns `"ACP agent does not support per-session MCP servers"` when MCP configs are provided but the agent advertises no MCP capabilities (`Acp: false, Http: false, Sse: false`). Kimchi's `newSession` also rejects `mcpServers` with `invalidParams`, so this AO-side check fires first. Tests: `TestNormalizeMCPServersFailsWithoutCapabilities`, `TestNormalizeMCPServersSucceedsWithHttpCapability`.

**Bonus fix:** While testing in non-yolo mode, I found that ACP permission requests were rendered as `file_change` activities (stuck spinner) instead of approval cards. Root cause: `RequestPermission` emitted the approval event with the same `ProviderItemID` as the tool call, causing `UpsertActivity` to overwrite the activity kind. Fixed by using the `requestID` as `ProviderItemID` so the approval gets its own row.

Issue #2375, opened 2026-07-03
Session rename accepts display names longer than the 20-character cap that spawn enforces


aprv10:
### Summary
Session display names are meant to be capped at 20 characters. **Spawn enforces this cap; rename does not.** Every rename entry point — `ao session rename`, `PATCH /api/v1/sessions/{id}`, and the `Service.Rename` layer they share — accepts and persists an arbitrarily long name. This lets a user set a sidebar label that spawn would have rejected, and contradicts the frontend's own comment claiming the daemon enforces the limit.

### Expected behavior
- Rename should apply the same 20-character (rune) cap as spawn and reject overlong names with the same `DISPLAY_NAME_TOO_LONG` error, at both the API and the CLI.

### Actual behavior
- `ao session rename <id> "<21+ chars>"` returns success and stores the name verbatim.
- `PATCH /api/v1/sessions/<id>` with a 21-character `displayName` returns `200 OK`.
- The identical body sent to `POST /api/v1/sessions` (spawn) is rejected with `400 DISPLAY_NAME_TOO_LONG`.

### Root cause
The cap exists on the spawn path but was never added to the rename path.

1. **Spawn caps it** — `utf8.RuneCountInString(displayName) > maxDisplayNameLen` → `DISPLAY_NAME_TOO_LONG`.
   `backend/internal/httpd/controllers/sessions.go:130`, and `backend/internal/cli/spawn.go:67`. Covered by `TestSessionsAPI_SpawnDisplayNameTooLong` (`sessions_test.go:692`).

2. **The PATCH handler only trims and rejects empty** — no length check.
   `backend/internal/httpd/controllers/sessions.go:324`

3. **`Service.Rename` only trims and rejects empty** — no length check.
   `backend/internal/service/session/service.go:323`

4. **The CLI only checks non-empty** (arg validator and command body).
   `backend/internal/cli/session.go:305` and `:479`

5. **The store does no truncation** — plain `UPDATE`, and SQLite `TEXT` has no length limit, so the overlong value persists.
   `backend/internal/storage/sqlite/store/session_store.go:45`

6. **The frontend documents the cap as if enforced server-side — it isn't.** The sidebar input's `maxLength={20}` guards only the desktop UI, not the CLI or raw API.
   `frontend/src/renderer/lib/rename-session.ts:4`, `frontend/src/renderer/components/Sidebar.tsx:627`

### Steps to reproduce
1. Spawn a session in any project.
[...]

PR #3759, opened 2026-08-09
fix(session): enforce the 20-character display name cap on rename

krishrathi1:
## Root cause

Session display names are meant to be capped at 20 runes. Spawn enforces this (`utf8.RuneCountInString(displayName) > maxDisplayNameLen` -> `400 DISPLAY_NAME_TOO_LONG`) in `backend/internal/httpd/controllers/sessions.go` and `backend/internal/cli/spawn.go`. The rename path never got the same check: `Service.Rename` in `backend/internal/service/session/service.go` only trimmed the name and rejected empty strings. Since both `ao session rename` and `PATCH /api/v1/sessions/{id}` call through this same service method, an overlong name (21+ characters) was accepted and persisted verbatim by rename, even though spawning a session with that same name would be rejected.

## Fix

Added the identical length check to `Service.Rename`:

```go
if utf8.RuneCountInString(displayName) > maxDisplayNameLen {
	return apierr.Invalid("DISPLAY_NAME_TOO_LONG", "Display name must be 20 characters or fewer", nil)
}
```

`maxDisplayNameLen` is a new unexported constant local to the `session` service package (20), mirroring the same-named, same-valued constants already defined independently in the controllers and CLI packages and in the project service package -- there was no existing shared/exported constant to reuse across packages. The error code, message, and `apierr.Invalid` (400-class) construction match the spawn path exactly.

This fixes the cap at the single choke point all rename entry points funnel through (CLI `ao session rename` -> `PATCH /api/v1/sessions/{id}` -> `Service.Rename`), so both are covered by one change.

## Verification

- Added `TestSessionRenameRejectsOverlongDisplayName` in `backend/internal/service/session/service_test.go`, alongside the existing `TestSessionRenameUpdatesDisplayName`, asserting a 21-character rename is rejected with `DISPLAY_NAME_TOO_LONG` (`apierr.KindInvalid`) and that the stored display name is left unchanged.
- Confirmed the test actually exercises the bug: with the `service.go` change stashed out (test kept), the new test fails against the old code with `err = <nil>, want DISPLAY_NAME_TOO_LONG` -- i.e. the old code silently accepts the overlong rename. Restoring the fix makes it pass.
- Full test suite for every touched/related package passes: `go test -count=1 ./internal/service/session/... ./internal/httpd/controllers/... ./internal/cli/...`.
- Ran the full `go test ./...` suite as a sanity check. There are pre-existing failures on this Windows machine in `internal/service/agent`, `internal/service/project` (Windows path-separator mismatches in `TestManager_InitializeRepositoryRecovery`), and `internal/session_manager` (PATH/runtime-environment-dependent tests). I confirmed these are pre-existing and unrelated by reproducing the identical failures with this change stashed out; none of them touch session rename or display-name validation.

## Scope

Kept the change to the service layer only, since that is the single point every rename entry point shares. I did not add a duplicate client-side check to the CLI (`backend/internal/cli/session.go`) or the PATCH controller (`backend/internal/httpd/controllers/sessions.go`) -- the daemon's `DISPLAY_NAME_TOO_LONG` error now propagates through both unchanged. I also left the comment in `frontend/src/renderer/lib/rename-session.ts` ("The daemon enforces the same 20-character limit as the spawn `--name` flag") as-is: it was inaccurate before this fix and is accurate now that the daemon actually enforces it.

Addresses #2375.

### DIFF
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"golang.org/x/sync/singleflight"
 
@@ -17,6 +18,12 @@ import (
 	sessionmanager "github.com/aoagents/agent-orchestrator/backend/internal/session_manager"
 	"github.com/aoagents/agent-orchestrator/backend/internal/telemetrymeta"
 )
+
+// maxDisplayNameLen caps the user-facing session display name. Mirrors the
+// spawn path's limit (backend/internal/httpd/controllers/sessions.go and
+// backend/internal/cli/spawn.go) so rename cannot set a name spawn would have
+// rejected.
+const maxDisplayNameLen = 20
 
 // Store is the read-only persistence surface needed to assemble controller-facing session read models.
 type Store interface {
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

Rishet11:
@krishrathi1

Issue #2451, opened 2026-07-06
bug(workspace): uncommitted edits to tracked-but-gitignored files are lost across session preserve/restore


abhay-codes07:
### Summary

When a session is closed, `StashUncommitted` saves the worktree's uncommitted work into a preserve commit that's replayed on reopen. An uncommitted edit to a file that is **tracked in HEAD but also matches a `.gitignore` pattern** is silently dropped from the preserve tree, and on restore `ApplyPreserved` then **deletes** the file — losing the agent's edit (data loss).

### Where

`backend/internal/adapters/workspace/gitworktree/workspace.go`, `StashUncommitted`. It captures work by pointing `GIT_INDEX_FILE` at a **fresh, empty** temp index and running `git add -A`:

```go
tmpIdx, _ := os.CreateTemp("", "ao-preserve-idx-*")
tmpIdxPath := tmpIdx.Name(); tmpIdx.Close(); os.Remove(tmpIdxPath) // git sees an absent (empty) index
...
addCmd := exec.CommandContext(ctx, w.binary, addAllTempIndexArgs(info.Path)...) // git add -A
addCmd.Env = append(os.Environ(), "GIT_INDEX_FILE="+tmpIdxPath)
```

### Why it's wrong

git's rule that "already-tracked files bypass `.gitignore`" keys off **what is in the index**. This index starts empty, so git thinks nothing is tracked. A file committed in HEAD that also matches a `.gitignore` pattern is therefore treated as an *untracked ignored* file and skipped by `git add -A` (no `-f`). The preserve tree omits it; since the preserve commit's parent is HEAD (which still contains the file), the diff reads as a **deletion**, and `ApplyPreserved` (cherry-pick) deletes the file on restore.

This is a real, non-exotic pattern: a config/template committed once and then added to `.gitignore` so local edits aren't pushed.

Verified against real git:
- Empty-index `git add -A` → preserve tree drops the file → `git diff --name-status HEAD <tree>` shows `D config.yaml`.
- Seeding the temp index from HEAD first (`git read-tree HEAD`) → `M config.yaml` (captured), while a genuinely untracked ignored file is still correctly skipped.

### Repro

1. Commit `config.yaml`. Then commit a `.gitignore` listing `config.yaml` (the file stays tracked).
2. Edit `config.yaml` (uncommitted) + do normal work.
3. Close the session → `StashUncommitted`. Reopen → `Restore` + `ApplyPreserved`.
[...]

PR #3760, opened 2026-08-09
fix(workspace): seed preserve temp index from HEAD before add -A

krishrathi1:
## Bug

This is a real data-loss bug. `StashUncommitted` in `backend/internal/adapters/workspace/gitworktree/workspace.go` builds the session's preserve commit by staging into a temp index file pointed at by `GIT_INDEX_FILE`, and that temp index deliberately starts absent/empty (git treats a nonexistent `GIT_INDEX_FILE` path as an empty index). It then runs `git add -A` against that empty index (no `-f`) to capture uncommitted work.

git's rule that an already-tracked path bypasses `.gitignore` keys off what is **in the index**, not what is in HEAD. Because the temp index starts empty, git has no record that anything is tracked. A file that IS committed in HEAD but ALSO matches a `.gitignore` pattern — a common, non-exotic pattern: commit a config/template once, then gitignore it so local edits never get pushed — is therefore indistinguishable from a genuinely untracked ignored file, and `git add -A` silently skips it.

The resulting preserve tree omits that file. Since the preserve commit's parent is HEAD (which still has the file), the diff between HEAD and the preserve tree reads as a **deletion** of the file. When the session reopens, `ApplyPreserved` cherry-picks that preserve commit onto the restored worktree and faithfully applies the delete — destroying the user's uncommitted edit.

## Fix

Seed the temp index from HEAD with `git read-tree HEAD` (same `GIT_INDEX_FILE` override) immediately before `git add -A`, so git is aware of what HEAD actually tracks:

- A tracked-but-gitignored file with uncommitted edits is now correctly staged as modified (`M`), not skipped.
- A genuinely untracked, gitignored file (build output, `node_modules`, etc.) is still correctly skipped by `git add -A`, since no `-f` is passed — this does not regress.
- An unborn HEAD (a brand new repo with no commits yet) has nothing to seed from: `git read-tree HEAD` fails with exit 128 and "fatal: Not a valid object name HEAD" (verified against real git). That specific failure is tolerated and the temp index is left empty, exactly as it behaved before this change. Any other `read-tree` failure is surfaced normally and is not swallowed.

The pre-existing comment in `StashUncommitted` explaining why the temp index file must not be pre-created is unaffected — this change only adds a `read-tree` call against the same not-pre-created path, after git has already started treating it as an empty index and before `add -A` runs against it.

## Verification

- Manually reproduced the bug against real git first, independent of this codebase: committed `config.yaml`, gitignored it, edited it uncommitted. `git add -A` against an empty temp index produced `D config.yaml` in `git diff --name-status HEAD <preserve-tree>`. Seeding with `git read-tree HEAD` first produced `M config.yaml` instead, while a genuinely untracked+ignored file stayed excluded from the diff in both cases.
- Added `TestWorkspaceIntegrationStashApplyPreservesTrackedIgnoredEdit` to `workspace_preserve_test.go`, mirroring this package's existing real-git integration test style (temp dirs, real `git` binary, no mocking). It: commits `config.yaml`, commits a `.gitignore` listing it, edits it uncommitted, runs the full `StashUncommitted` -> `ForceDestroy` -> `Restore` -> `ApplyPreserved` round trip, and asserts `config.yaml` exists afterward with the edited content (not the original, not deleted).
- Same test also covers the negative case in one run: a genuinely untracked file matching `.gitignore` (`build/output.log`, never committed) is asserted to stay excluded from the round trip. This is the main regression risk of seeding from HEAD, and it passes — confirmed not regressed.
- Proved the fix actually matters: stashed only the fix (`commands.go`/`workspace.go`), kept the new test, ran it against the old code — it FAILED with "config.yaml missing after apply (data loss)", reproducing the bug exactly as described. Popped the stash, reran — it PASSED.
- Full `gitworktree` package test suite passes, including the pre-existing `TestWorkspaceIntegrationStashApplyRoundTrip`, which already asserts a genuinely-ignored file (`secret.txt`) stays excluded — no regression there either.
- Ran the full `backend` test suite (`go test ./...`). Three unrelated failures in `internal/session_manager` (`TestSpawn_DefaultsBranchUnderDevNamespaceForDevDataDir`, `TestSpawnAndRestore_PrependsResolvedBinaryAndNodeDirsToRuntimePATH`, `TestSpawn_DoesNotAddNodeRuntimeForNativeBinary`) are pre-existing on unmodified `main` in this environment (confirmed by stashing this change entirely and re-running them) and are unrelated to this change — they look like Windows path-separator / local node-version environment assumptions, not something this diff touches.

I did not add a dedicated unborn-HEAD test case: the existing test fixtures in this file all set up a repo with an initial commit via `setupOriginClone`, and adding a from-scratch no-commit fixture felt like a bigger addition than this fix warrants. I manually traced the code path instead (confirmed the exact git error text and exit code against real git, shown above) and I'm confident in it, but flagging honestly that it isn't covered by an automated test in this PR.

Addresses #2451.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/commands.go
+++ b/backend/internal/adapters/workspace/gitworktree/commands.go
@@ -71,6 +71,19 @@ func worktreeListPorcelainArgs(repo string) []string {
 	return []string{"-C", repo, "worktree", "list", "--porcelain"}
 }
 
+// readTreeHeadArgs seeds a temp index from HEAD so git's "a tracked path
+// bypasses .gitignore" rule (which keys off what is IN THE INDEX, not what is
+// in HEAD) has something to key off before the subsequent `git add -A` runs
+// against the same temp index. Without this, a fresh/empty temp index makes
+// git treat a file that is committed in HEAD but also matches .gitignore as
+// untracked+ignored, so `add -A` silently skips it. GIT_INDEX_FILE must be set
+// in the command's environment before calling. Fails with exit 128 and
+// "fatal: Not a valid object name HEAD" on an unborn HEAD (no commits yet);
+// callers must tolerate that specific failure and leave the temp index empty.
+func readTreeHeadArgs(worktree string) []string {
+	return []string{"-C", worktree, "read-tree", "HEAD"}
+}
+
 // addAllTempIndexArgs stages all tracked and non-ignored untracked files into a
 // temp index file without touching the real index or the working tree.
 // GIT_INDEX_FILE must be set in the command's environment before calling.
--- a/backend/internal/adapters/workspace/gitworktree/workspace.go
+++ b/backend/internal/adapters/workspace/gitworktree/workspace.go
@@ -414,6 +414,32 @@ func (w *Workspace) StashUncommitted(ctx context.Context, info ports.WorkspaceIn
 	// Deferred remove is a best-effort cleanup in case git leaves the file.
 	defer func() { _ = os.Remove(tmpIdxPath) }()
 
+	// Seed the temp index from HEAD before staging. git's "an already-tracked
+	// path bypasses .gitignore" rule keys off what is IN THE INDEX, not what
+	// is in HEAD, and the temp index above starts absent (empty). Without this
+	// seed step, a file that is committed in HEAD but ALSO matches a
+	// .gitignore pattern looks, from this empty index's perspective, exactly
+	// like a genuinely untracked ignored file, so the `git add -A` below would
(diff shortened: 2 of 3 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

Rishet11:
@krishrathi1
Closing this PR as stale after 30+ days without author activity.
This is the temp-index data-loss fix, so please reopen it if that protection is still needed.

Issue #2174, opened 2026-06-24
bug(session activity): waiting_input masks PR-derived status after worker opens a PR


neversettle17-101:
Migrated from aoagents/ReverbCode/issues/290

Source: https://github.com/aoagents/ReverbCode/issues/290
Original assignees: harshitsinghbhandari
Original labels: bug, lcm-sm, priority: high

---

## Bug

A worker session can own an open PR while the session display still shows `needs_input`.

This happens when the session's last persisted activity signal is `waiting_input`. The read model derives `status` from durable activity plus PR facts, but `waiting_input` currently returns before PR facts are evaluated, so a stale activity state masks the PR pipeline state.

**Source:** local observation from current worker session  
**Reported by:** @aditi1178  
**Analyzed against:** `f98c5e56cca246b9461552b938c5995a747551d8`  
**Confidence:** High

## Reproduction

1. A worker reports `activity.state = waiting_input`.
2. The same worker raises/owns a PR, so `ListPRFactsForSession` returns at least one open PR.
3. List/get the session.
4. Actual: `status` remains `needs_input`.
5. Expected: session status reflects the PR pipeline (`pr_open`, `review_pending`, `ci_failed`, `changes_requested`, `approved`, `mergeable`, etc.).

## Root Cause

`deriveStatus` prioritizes `ActivityWaitingInput` before PR facts:

```go
if rec.Activity.State == domain.ActivityWaitingInput {
    return domain.StatusNeedsInput
}

open := openPRs(prs)
if len(open) > 0 {
    return aggregatePRStatus(open)
}
```

`backend/internal/service/session/status.go`.

So once a worker has a sticky `waiting_input` activity, PR attribution cannot change the display status. The current unit test in `backend/internal/service/session/status_test.go` codifies this behavior by expecting `waiting_input + CI failing PR => needs_input`; that expectation should change because PR facts are a newer, more actionable session signal.

Add/adjust tests:

- `waiting_input + open PR => pr_open`
- `waiting_input + failing PR => ci_failed`
- `waiting_input + no PR => needs_input`
- terminated sessions keep current precedence (`terminated + merged PR => merged`)

## Impact

- Board/session UI shows a worker as needing input even after it has raised a PR.
- Activity appears stale because the display status is pinned to the last `waiting_input` hook.
- Orchestrator/user attention is routed to "answer the agent" instead of reviewing the PR state.

PR #3761, opened 2026-08-09
fix(session): let PR status win over sticky waiting_input/blocked

krishrathi1:
## Root cause

`deriveStatus` in `backend/internal/service/session/status.go` returns `StatusNeedsInput` for `ActivityWaitingInput`/`ActivityBlocked` before ever consulting PR facts (`deriveSCMStatus`, which is only reached further down the function). `waiting_input` is a sticky, very common activity state (e.g. it is set after a Claude Code hook fires once the agent's turn ends), so once a worker session hits it, the session's displayed status stays pinned at `needs_input` even after the worker opens a PR whose own pipeline status (`pr_open`, `ci_failed`, `review_pending`, `changes_requested`, `approved`, `mergeable`) is a much more actionable signal for the user.

## Fix

Reordered the `ActivityWaitingInput`/`ActivityBlocked` case so it consults `deriveSCMStatus(prs)` first and only falls back to `StatusNeedsInput` when there is no open (or merged) PR to derive a status from. This is a surgical, three-line reorder inside the existing switch case; no other branch of `deriveStatus` was touched.

Terminated-session precedence is unaffected: the `rec.IsTerminated` check still short-circuits at the very top of the function, before the activity switch is ever reached, so `terminated + merged PR => merged` (and `terminated => terminated` with no PR) behaves exactly as before regardless of the last activity signal.

`ActivityBlocked` was given the same treatment as `ActivityWaitingInput` since the issue is silent on `Blocked` specifically, but both share the same `case` in the switch (grouped together already), and the same masking bug applies equally to both.

## Tests changed/added (`backend/internal/service/session/status_test.go`)

- Changed `needs-input` / `needs-input-blocked` (previously asserted `waiting_input`/`blocked` + a CI-failing PR => `needs_input`, which codified the bug) to use no PR at all, so they now correctly assert the genuine no-PR needs_input case.
- Added `waiting-input-pr-derived-ci-failed` / `blocked-pr-derived-ci-failed`: `waiting_input`/`blocked` + CI-failing PR => `ci_failed`.
- Added `terminated-waiting-input-merged-pr-still-merged`: a terminated session with a `waiting_input` activity signal and a merged PR still reports `merged`, confirming terminated precedence is unaffected by this fix.
- In `TestDeriveStatusActivityPrecedence`, updated `waiting-with-pr` / `blocked-with-pr` (mergeable PR) to expect `mergeable` instead of the old `needs_input`, and added `waiting-without-pr`, `blocked-without-pr` (no PR => `needs_input`), `waiting-with-open-pr`, `blocked-with-open-pr` (plain open PR => `pr_open`), and `waiting-with-ci-failing-pr`, `blocked-with-ci-failing-pr` (=> `ci_failed`), covering exactly the cases the issue calls out.

## Verification

- Proved the new/changed tests actually catch the bug: reverse-applied just the `status.go` fix (via a saved patch, keeping the updated tests in place), reran the affected tests, and confirmed every new/changed case failed with `got "needs_input" want "..."` against the old code. Reapplied the fix and confirmed all of them pass.
- `go build ./...` clean.
- `go vet ./...` clean.
- `go test ./internal/service/session/...` passes in full, including every pre-existing test in the package (no regressions) plus all new/updated cases.
- Branch is up to date with `upstream/main` (0 commits behind at time of opening this PR), so no rebase was needed.

Addresses #2174.

### DIFF
--- a/backend/internal/service/session/status.go
+++ b/backend/internal/service/session/status.go
@@ -38,6 +38,15 @@ func deriveStatus(rec domain.SessionRecord, prs []domain.PRFacts, now time.Time,
 	case domain.ActivityExited:
 		return domain.StatusExited
 	case domain.ActivityWaitingInput, domain.ActivityBlocked:
+		// PR facts are a newer, more actionable signal than a sticky
+		// waiting_input/blocked activity state: once a worker has raised a
+		// PR, its pipeline status (pr_open, ci_failed, review_pending, ...)
+		// is more useful than a stale "needs input" pinned from the last
+		// hook callback. Only fall back to needs_input when there is no
+		// open (or merged) PR to derive a status from.
+		if scmStatus := deriveSCMStatus(prs); scmStatus != "" {
+			return scmStatus
+		}
 		return domain.StatusNeedsInput
 	}
 
--- a/backend/internal/service/session/status_test.go
+++ b/backend/internal/service/session/status_test.go
@@ -48,8 +48,22 @@ func TestServiceDerivesStatusFromSessionFactsAndPR(t *testing.T) {
 	}{
 		{"terminated", statusRec(domain.ActivityExited, true), nil, false, domain.StatusTerminated},
 		{"merged-pr", statusRec(domain.ActivityIdle, true), statusPR(domain.PRFacts{Merged: true}), false, domain.StatusMerged},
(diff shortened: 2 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@krishrathi1 
Issue #2174 was closed as not planned on Aug 22, after this PR opened, and no fix landed. The feature decision went the other way. Closing.

Issue #2434, opened 2026-07-05
fix(web): terminal doesn't get keyboard focus when opening a session/agent view


mukesher4:
## Bug

When opening a terminal view — clicking an agent from the Kanban board, selecting an orchestrator session, or otherwise navigating into a session's terminal — the xterm.js terminal never receives keyboard focus. The user has to manually click inside the terminal pane before typing does anything.

**Analyzed against:** `860fa5a2` (origin/main)
**Confidence:** High — traced exact code path, root cause confirmed by absence of any `.focus()` call.

## Root Cause

- `frontend/src/renderer/components/XtermTerminal.tsx` owns the real `Terminal` instance (`term`) from `@xterm/xterm`, which has a built-in `term.focus()` method — but nothing in this file ever calls it.
- The handle object it exposes outward, `AttachableTerminal` (defined in `frontend/src/renderer/hooks/useTerminalSession.ts`), only exposes `cols`, `rows`, `write`, `writeln`, `clear`, `onUserInput`, `onResize` — no `focus` method, so callers have no way to trigger it even if they wanted to.
- `frontend/src/renderer/components/TerminalPane.tsx`'s `AttachedTerminal` keys the terminal by `terminalHandleId` (`key={terminalKey}`), so React fully remounts `XtermTerminal` (destroy + recreate the xterm.js instance) on every session switch — giving a natural, reliable "just opened" moment via the mount effect, which currently does nothing focus-related.

## Screen Recording

[![Screen Recording](https://img.youtube.com/vi/zBYWUGjXw6I/0.jpg)](https://youtu.be/zBYWUGjXw6I)

## Fix

1. Add `focus: () => term.focus()` to the `AttachableTerminal` handle object built in `XtermTerminal.tsx` — exposes the capability (mechanism), without deciding when to use it.
2. Add `focus: () => void` to the `AttachableTerminal` type in `useTerminalSession.ts`.
3. Call `terminal.focus()` in `TerminalPane.tsx`'s existing mount effect (in `AttachedTerminal`), gated on `handleId` being present — `TerminalPane` is the layer that knows whether a session/handle is actually attached (`showEmptyState = !handleId`), so it decides *when* to focus (policy), rather than `XtermTerminal` focusing unconditionally on every mount, which would also steal focus onto the empty-state pane (no session selected).

## Impact

- Low severity, high friction: every session/agent switch currently requires an extra click before the user can type, on every navigation.

homeshrawat:
[...]

PR #3762, opened 2026-08-09
fix(web): give the terminal keyboard focus when a session/handle attaches

krishrathi1:
## Root cause

Opening a session/agent terminal (from the Kanban board, an orchestrator session, or switching tabs) never gave the underlying xterm.js instance keyboard focus, so typing did nothing until the user manually clicked inside the terminal pane first.

- `frontend/src/renderer/components/XtermTerminal.tsx` owns the real `Terminal` instance (`term`) from `@xterm/xterm`, which has a built-in `term.focus()` method, but nothing in this file ever called it.
- The handle object it exposes outward, `AttachableTerminal` (`frontend/src/renderer/hooks/useTerminalSession.ts`), had no `focus` method, so callers had no way to request it.
- `frontend/src/renderer/components/TerminalPane.tsx`'s `AttachedTerminal` has an existing effect that attaches the terminal to its PTY handle once `prepareForActivation()` resolves — a reliable "just opened" moment on every session/tab switch — but it did nothing focus-related.

## Fix

Split by mechanism (capability) vs. policy (when to use it), same as the surrounding code's existing separation of concerns:

1. `XtermTerminal.tsx`: the handle now exposes `focus`, wired to the existing `focusTerminal` helper (already used by the context-menu's copy/paste/selectAll/clear actions to restore focus after using the menu) rather than a bare `term.focus()` call, so the same try/catch around a torn-down terminal or unavailable hidden textarea applies here too.
2. `useTerminalSession.ts`: the `AttachableTerminal` type gains the matching `focus: () => void` field.
3. `TerminalPane.tsx`: `AttachedTerminal`'s existing attach effect calls `terminal.focus()` right after `attach(terminal)`, gated on `handleId` being present. This keeps the empty-state pane (no session selected, `showEmptyState = !handleId`) from ever having focus stolen onto it, since that gate is exactly the one already governing the empty-state UI.

## Verification

- `npx tsc --noEmit` from `frontend/` is clean (ran `npm ci` first since `node_modules` wasn't present in the worktree).
- Existing unit tests cover this seam: `TerminalPane.test.tsx` and `useTerminalSession.test.tsx` both construct a fake `AttachableTerminal`/xterm mock that now had to gain a `focus` field to satisfy the updated type — fixed both fakes.
- Added three focused tests to `TerminalPane.test.tsx` (`describe("TerminalPane terminal focus")`):
  - focuses the terminal once a session handle is attached,
  - does NOT focus when no session is selected (empty state),
  - does NOT focus while a session is starting but has no terminal handle yet.
  I verified these tests are meaningful (not vacuously true) by temporarily reverting the `TerminalPane.tsx` change and confirming the positive test fails without the fix, then re-applied the fix.
- Full suite for the three touched files: `TerminalPane.test.tsx`, `useTerminalSession.test.tsx`, `XtermTerminal.test.tsx` — 134 tests passed (had to pass `--pool=threads` to vitest; the default forked-process pool times out in this sandboxed environment, unrelated to the change).
- I do not have a way to launch the real Electron desktop app in this environment (no GUI, and the repo's `ao-desktop-dev` dev skill isn't available in this session), so I was not able to do a live click-free-typing check in the actual app. Verification here is code review + typecheck + unit tests only — flagging this honestly per repo convention for intentional verification gaps.

Addresses #2434.

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.test.tsx
+++ b/frontend/src/renderer/components/TerminalPane.test.tsx
@@ -17,6 +17,7 @@ import {
 
 const {
 	attachMock,
+	focusMock,
 	getMock,
 	postMock,
 	prepareForActivationMock,
@@ -29,6 +30,7 @@ const {
 } = vi.hoisted(
 	() => ({
 		attachMock: vi.fn(() => vi.fn()),
+		focusMock: vi.fn(),
 		getMock: vi.fn(async (_path: string, _options: unknown) => ({ data: undefined })),
 		postMock: vi.fn(),
 		prepareForActivationMock: vi.fn(async (): Promise<void> => undefined),
@@ -69,6 +71,7 @@ vi.mock("./XtermTerminal", () => ({
 			props.onReady?.({
 				cols: 80,
 				rows: 24,
+				focus: focusMock,
 				write: vi.fn((_data, done) => done?.()),
 				writeln: vi.fn(),
 				showLatestOutput: vi.fn(),
@@ -129,6 +132,7 @@ beforeEach(() => {
 	terminalLinkHandler = undefined;
 	terminalSessionOptions.length = 0;
 	attachMock.mockClear();
+	focusMock.mockClear();
 	prepareForActivationMock.mockReset();
 	prepareForActivationMock.mockResolvedValue(undefined);
 	xtermMounts.value = 0;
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

Rishet11:
@krishrathi1

PR #2432, opened 2026-07-05
feat(frontend): replace empty board shells with first-run onboarding states

illegalcall:
## Problem

First launch drops users into the full app shell — four empty kanban columns, an empty sidebar, and zero orientation ([onboarding audit, finding #1, high severity](https://illegalcall.github.io/ao-onboarding-audit/)). Core vocabulary (orchestrator, worker, the four columns) is never defined in-app (findings #2/#4).

## What this does

**Global board (`/`) with no projects** — replaces the empty column shells with a centered welcome: what the app does in one sentence, a three-step card (Add a project → Describe a task → Review and merge), and a primary **Add your first project** CTA. Step 3 ends with the four column dots in their exact board colors/styles, so the kanban vocabulary is taught before the columns ever render. The CTA drives the same folder-picker → agent-sheet flow as the sidebar's `+` (shared `CreateProjectFlow`, extracted verbatim from `Sidebar.tsx`).

**Project board with no sessions** — a smaller centered "No sessions yet" invitation with the header's Orchestrator (primary) / New task actions, instead of four zeros.

Both states render only after the workspace query succeeds, so they never flash over real data. Populated boards are unchanged.

## Bug fix this surfaced

`SessionsBoard.openOrchestrator` swallowed spawn rejections (`try/finally` with no `catch`): a daemon 409 (e.g. orchestrator branch already checked out in another worktree) left the button apparently dead. The daemon's message now renders in the empty state and beside the header actions, clears on retry and on project switch. Folder-picker failures in `CreateProjectFlow` also get a visible error (previously `sr-only` only).

## Design

Built entirely from existing tokens per DESIGN.md — `bg-surface` card, hairline borders, 13px column radius, mono step numbers, existing primary/accent button classes, lucide icons only. No new visual language.

## Testing

- 8 new integration tests (`board-empty-states.test.tsx`): both empty states, CTA → picker, columns return with sessions, spawn-error surfacing, stale-error clearing on project switch, picker-failure visibility.
- `npm run typecheck` clean; full renderer suite 387/387.
- Verified live in Electron against a fresh `AO_DATA_DIR`.
- CodeRabbit review clean (its one finding — invisible picker errors — fixed).

image
image

Issue #2242, opened 2026-06-27
Electron keeps stale project actions enabled after daemon exit and silently fails with AO daemon is not ready


areycruzer:
## Summary

If the Electron-managed daemon exits while the desktop app is still open, the renderer notices enough to stop showing project data, but it leaves project actions such as **New task** and **Spawn Orchestrator** clickable on the stale board route. Clicking **Spawn Orchestrator** produces no visible error or recovery path in the app and logs an unhandled rejection: `AO daemon is not ready`.

This makes daemon crash/restart recovery feel broken: the user is left on an apparently usable screen, but actions silently fail and there is no obvious **Restart daemon** / **Reconnect** affordance even though the Electron preload/main layer already exposes daemon start IPC.

## Reproduced on

- Repo: `AgentWrapper/agent-orchestrator`
- Branch: `main`
- Commit: `77f33c28`
- Platform: macOS / Electron dev app
- Runtime state: isolated local QA state under `/tmp/agent-orchestrator-qa-20260627-230707`
- Launch env:

```bash
AO_DATA_DIR=/tmp/agent-orchestrator-qa-20260627-230707/ao-data \
AO_RUN_FILE=/tmp/agent-orchestrator-qa-20260627-230707/running.json \
AO_PORT=31021 \
AO_TELEMETRY_EVENTS=off \
AO_TELEMETRY_METRICS=off \
AO_TELEMETRY_REMOTE=off \
VITE_AO_POSTHOG_HOST=http://127.0.0.1:9 \
npm --prefix frontend run dev
```

## Repro steps

1. Start the Electron dev app with the isolated env above.
2. Add/open a disposable local git project.
3. Confirm the Electron UI is on a project board route and daemon-backed data is visible.
4. Kill the isolated daemon process listed in `AO_RUN_FILE` while leaving Electron open.
5. Wait for Electron to observe the daemon exit.
6. Click **Spawn Orchestrator** on the still-visible project board.

## Expected behavior

After the daemon exits, the UI should stop offering daemon-dependent actions or show a clear recovery state, for example:

- Disable **New task** / **Spawn Orchestrator** while daemon status is stopped.
- Show a visible **Restart daemon** / **Reconnect** action.
- Catch action failures and render an inline/toast error instead of producing an unhandled rejection.
- Preferably call the already-exposed `daemon.start` IPC when the user chooses recovery.

## Actual behavior

[...]

PR #3763, opened 2026-08-09
fix: gate board actions on daemon status

krishrathi1:
## What

Disable **New task** and **Spawn Orchestrator** on the project board (and the Terminal UI retry button shown after a failed spawn) while the daemon is not in the `ready` state.

## Why

When the Electron-managed daemon exits while the desktop app stays open, `SessionsBoard` deliberately keeps rendering the last-known project data instead of blanking the screen. That's correct, but the board's `New task` / `Spawn Orchestrator` header actions were gated only on `isSpawning` / `isProjectRestarting`, never on daemon readiness. So the stale board stayed fully interactive: clicking either action called `spawnOrchestrator` against an unreachable daemon API and threw `AO daemon is not ready`.

## How

- Verified the gap against current `main` first. `useDaemonStatus`/`shell-context` already expose `daemonStatus.state: "starting" | "ready" | "stopped" | "error"`, and `SessionsBoard` already derives `isDaemonReady = usesPreviewWorkspaceData || (shell ? shell.daemonStatus.state === "ready" : true)` for its own loading/startup screens - it just wasn't reused on the action buttons.
- On an unexpected daemon exit, Electron's main process sets `{ state: "stopped", code: "exited" }` (see `frontend/src/main.ts`). Because `code` is set, the board's `daemonHasFailed` flag is true, which keeps the normal board (not the startup loader) mounted with the header actions row - this is exactly the reported "stale board stays interactive" state, and it's what the new test reproduces.
- Added `!isDaemonReady` to the `disabled` condition on `New task`, `Spawn Orchestrator`, and the `Create as Terminal UI` retry button in `frontend/src/renderer/components/SessionsBoard.tsx`. No new UI: this reuses the app's existing native-`disabled` + `opacity-60` styling (`TopbarButton`), consistent with how the rest of the app already shows an unavailable action (e.g. `TerminalPane`'s `daemonReady` gating).
- I also checked the other half of the original report - `openOrchestrator` swallowing the daemon error as an unhandled rejection (`try/finally`, no `catch`). On current `main` this is already fixed: #2432 added a proper `catch` that sets `spawnError` and renders it via the existing `TopbarKillError` component (same pattern reused here for consistency). So that part of the issue no longer reproduces; this PR covers the remaining gap, the buttons staying enabled/clickable.
- `New task`'s own daemon-error handling (inside `TaskComposer`, opened via the dialog) was already solid - try/catch with an inline error message - so the only real gap for that action was the missing gate, which this PR fixes.

### Intentional scope note

`ShellTopbar.tsx` renders the same `New task` / `Spawn Orchestrator` pair for the Windows/Linux board layout and for session-detail routes, and has the identical missing-gate pattern (its own `openOrchestrator` already has a `catch`, so no unhandled rejection there either, just no daemon-ready gate). I scoped this PR to `SessionsBoard.tsx` per the issue's explicit repro (macOS board route) and code pointers, to keep the change surgical and reviewable as one fix. Flagging `ShellTopbar.tsx` here as a known follow-up rather than folding it into this PR silently.

## Testing

- `cd frontend && npx tsc --noEmit` - clean, no errors.
- Added a new test to `frontend/src/renderer/__tests__/integration/board-empty-states.test.tsx`: renders a populated board with `daemonStatus: { state: "ready" }` (buttons enabled), then flips to `{ state: "stopped", code: "exited" }` (the exact status Electron sets on an unexpected exit) and asserts both buttons become disabled and that clicking no longer calls `spawnOrchestrator`. Confirmed this test actually catches the regression by temporarily reverting the `SessionsBoard.tsx` change and re-running it (fails as expected: buttons stay enabled).
- Ran the full targeted suite: `npx vitest run --config vite.renderer.config.ts src/renderer/__tests__/integration/board-empty-states.test.tsx src/renderer/components/SessionsBoard.test.tsx` - 54/54 passing.
- Ran the full frontend `vitest` suite as a broader safety net: 2043 passed / 6 failed / 2 skipped (plus the separate `src/landing` sub-package, 31 passed / 5 failed). All failures are pre-existing and unrelated to this change: the `src/landing` failures are a missing `node-html-markdown` dependency in that sub-package's own `package.json` (needs its own install, untouched by this PR), and the 6 `src/main/supervisor-link.test.ts` failures are `EACCES: permission denied` creating Unix-socket test fixtures under Windows `%TEMP%` in this sandbox - unrelated to anything touched here (`src/main/*` isn't part of this diff).
- **No live Electron verification.** I wasn't able to launch the real desktop app in this environment (no confirmed display for screenshot verification, and the sandbox showed heavy concurrent multi-agent activity sharing this checkout's git state, which made spinning up a dev Electron instance against shared `~/.ao` paths risky). Verification here is code review + typecheck + the tests above only - flagging this honestly rather than claiming a live-app check that wasn't actually done.

Addresses #2242.

### DIFF
--- a/frontend/src/renderer/__tests__/integration/board-empty-states.test.tsx
+++ b/frontend/src/renderer/__tests__/integration/board-empty-states.test.tsx
@@ -8,12 +8,19 @@ import type { ReactNode } from "react";
 // first-run states, mocking only the HTTP client, the router, and the native
 // folder picker: an empty daemon shows the import chooser (no column shells), a
 // fresh project shows the task invitation, and any session brings the columns back.
-const { getMock, navigateMock, chooseDirectoryMock, spawnOrchestratorMock } = vi.hoisted(() => ({
-	getMock: vi.fn(),
-	navigateMock: vi.fn(),
-	chooseDirectoryMock: vi.fn(),
-	spawnOrchestratorMock: vi.fn(),
-}));
+const { getMock, navigateMock, chooseDirectoryMock, spawnOrchestratorMock, boardActionsInPanelMock } = vi.hoisted(
+	() => ({
+		getMock: vi.fn(),
+		navigateMock: vi.fn(),
+		chooseDirectoryMock: vi.fn(),
+		spawnOrchestratorMock: vi.fn(),
+		// jsdom's default navigator reports no platform, so the real
+		// usesBoardActionsInPanel() (macOS-only) is false here already; this mock
+		// only exists so the daemon-gating test below can flip it on to exercise
+		// the in-panel header actions row the same way macOS does.
+		boardActionsInPanelMock: vi.fn(() => false),
+	}),
+);
 
 vi.mock("../../lib/spawn-orchestrator", () => ({
 	isChatPreflightError: (error: unknown) =>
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

Rishet11:
@krishrathi1

Issue #2390, opened 2026-07-03
bug(session): terminated session reports merged while it still owns an open PR


abhay-codes07:
### Summary

`deriveStatus` reports a **terminated** session as `merged` whenever *any* of its PRs is merged — even if the session still owns an **open** PR. In that case the status should stay `terminated`, matching the function's own documented invariant.

### Where

`backend/internal/service/session/status.go`, the terminated branch:

```go
if rec.IsTerminated {
    if anyMerged(prs) {
        return domain.StatusMerged
    }
    return domain.StatusTerminated
}
```

The docstring right above states the rule: *"Merged/closed PRs only matter once no open PR remains."* The non-terminated branch already enforces it — it computes `open := openPRs(prs)` and returns the aggregated PR status while `len(open) > 0`, only falling back to `anyMerged` once no open PR is left. The terminated branch skips that gate.

### How it regressed

This looks like a leftover from the single-PR → multi-PR change (#230). The branch used to be `if pr != nil && pr.Merged`, which in the single-PR world implied "the one PR merged, so nothing is open." Generalizing it to `anyMerged(prs)` dropped that implicit "no open PR" condition, while the non-terminated branch *was* correctly restructured with an explicit `openPRs` gate. The two branches disagree now.

### Repro

A worker owns a stack: PR A (bottom) and PR B (child of A). PR A merges, then the session is killed while B is still open — so `prs = [{A: merged}, {B: open}]` and `IsTerminated = true`.

- **Actual:** `anyMerged` is true → status `merged`. The dashboard shows the killed session as cleanly merged, hiding the abandoned open PR B.
- **Expected:** `terminated` — an open PR remains, so the session did not complete.

The same thing happens with two independent PRs where one merges and the session is killed with the other still open.

### Suggested fix

Gate the merged short-circuit on there being no open PR, mirroring the non-terminated branch:

```go
if len(openPRs(prs)) == 0 && anyMerged(prs) {
    return domain.StatusMerged
}
```

Existing behavior is preserved (a terminated session whose only PR is merged still reports `merged`); only the `merged + open` combination changes to `terminated`.

i-trytoohard:
From the NotesBot call on `2026-07-21`, this existing issue appears to be relevant.
Match confidence: `0.67` (fuzzy).

Relevant discussion excerpts:
[...]

PR #2392, opened 2026-07-03
fix(session): keep terminated status when an open PR remains

abhay-codes07:
Closes #2390.

## What

`deriveStatus`'s terminated branch returned `merged` whenever any owned PR was merged, even when the session still owned an **open** PR. That contradicts the function's documented invariant — *"Merged/closed PRs only matter once no open PR remains"* — and is inconsistent with the non-terminated branch, which gates on `openPRs` before falling back to `anyMerged`.

## Why

Regression from the single-PR → multi-PR change (#230). The branch used to be `if pr != nil && pr.Merged`, which in the single-PR world implied "no open PR remains." Generalizing to `anyMerged(prs)` dropped that implicit condition. A session killed while it still owns an open PR (e.g. only the bottom of a stack merged, or one of two independent PRs merged) was shown as cleanly "merged", hiding the abandoned open PR.

## Change

Gate the merged short-circuit on there being no open PR, mirroring the non-terminated branch:

```go
if len(openPRs(prs)) == 0 && anyMerged(prs) {
    return domain.StatusMerged
}
```

Existing behavior is preserved — a terminated session whose only PR is merged still reports `merged` (the existing `merged-pr` case). Only the `merged + open` combination changes to `terminated`.

## Tests

- Added `terminated-with-open-pr-stays-terminated` to the `deriveStatus` table test. It fails before the change (`got "merged" want "terminated"`) and passes after.
- `go test ./internal/service/session/...` green; full `go test ./...` green apart from the pre-existing Windows-only `TestApplySymlinks`, which needs the symlink privilege and fails on `main` too (unrelated to this change).
- `go vet` and gofmt clean. I couldn't run `go test -race` locally (no 64-bit CGO toolchain on my box) — leaving that to CI.

cc @harshitsinghbhandari — small correctness fix, happy to adjust if you'd rather gate it another way.

PR #3764, opened 2026-08-09
fix(session): keep terminated status when an open PR remains

krishrathi1:
### Root cause

`deriveStatus` in `backend/internal/service/session/status.go` has a terminated branch that reports `StatusMerged` whenever *any* of the session's PRs is merged:

```go
if rec.IsTerminated {
    if anyMerged(prs) {
        return domain.StatusMerged
    }
    return domain.StatusTerminated
}
```

The function's own doc comment states the invariant: "Merged/closed PRs only matter once no open PR remains." The non-terminated branch already enforces this correctly (via `deriveSCMStatus`, which computes `open := openPRs(prs)` and only falls back to `anyMerged` once `len(open) == 0`). The terminated branch skips that gate, so the two branches disagree.

Concretely: a session owning a stacked PR pair (PR A merges, PR B is still open) that gets killed while B is open reports as cleanly `merged`, hiding the abandoned open PR B from the dashboard. The same happens with two independent PRs where one merges and the session is killed with the other still open.

### Fix

Gate the terminated branch's merged short-circuit on there being no open PR remaining, mirroring the non-terminated branch:

```go
if rec.IsTerminated {
    if len(openPRs(prs)) == 0 && anyMerged(prs) {
        return domain.StatusMerged
    }
    return domain.StatusTerminated
}
```

Existing behavior is preserved: a terminated session whose only PR is merged still reports `merged`. Only the `merged + open` combination changes, from `merged` to `terminated`.

### Tests

Added three cases to `TestServiceDerivesStatusFromSessionFactsAndPR` in `status_test.go`:

- `terminated-only-pr-merged`: terminated, single merged PR, no open PR -> `StatusMerged` (regression guard on existing behavior).
- `terminated-with-open-pr-stays-terminated`: terminated, one merged PR plus one still-open PR (the stacked-PR repro) -> `StatusTerminated`, not `StatusMerged`.
- `terminated-no-prs`: terminated, no PRs at all -> `StatusTerminated` (sanity, must not regress).

### Verification

- Confirmed the bug against current `main`: reverted just the one-line fix (via a standalone patch, reapplied after) and reran the new test case, which failed with `got "merged" want "terminated"`; the other new cases still passed. Reapplied the fix and reran: all pass.
- `cd backend && go build ./...` and `go vet ./...`: clean.
- `cd backend && go test ./internal/service/session/...`: all tests pass, including the full existing suite (no regressions).
- `cd backend && go test ./...`: only unrelated, pre-existing failures in `internal/session_manager` (Windows path-separator/PATH-construction test expectations), confirmed to fail identically with the fix reverted, so unrelated to this change.

Addresses #2390.

### DIFF
--- a/backend/internal/service/session/status.go
+++ b/backend/internal/service/session/status.go
@@ -26,7 +26,7 @@ const noSignalGrace = 90 * time.Second
 // until the parent does. Merged/closed PRs only matter once no open PR remains.
 func deriveStatus(rec domain.SessionRecord, prs []domain.PRFacts, now time.Time, signalCapable bool) domain.SessionStatus {
 	if rec.IsTerminated {
-		if anyMerged(prs) {
+		if len(openPRs(prs)) == 0 && anyMerged(prs) {
 			return domain.StatusMerged
 		}
 		return domain.StatusTerminated
--- a/backend/internal/service/session/status_test.go
+++ b/backend/internal/service/session/status_test.go
@@ -48,6 +48,23 @@ func TestServiceDerivesStatusFromSessionFactsAndPR(t *testing.T) {
 	}{
 		{"terminated", statusRec(domain.ActivityExited, true), nil, false, domain.StatusTerminated},
 		{"merged-pr", statusRec(domain.ActivityIdle, true), statusPR(domain.PRFacts{Merged: true}), false, domain.StatusMerged},
+		// A terminated session whose only PR is merged still reports merged: no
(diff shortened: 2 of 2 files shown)


krishrathi1:
Closing in favor of #2392, which already implements the same fix (terminated status gated on no remaining open PR) and was opened well before mine. Didn't spot it during triage since it wasn't referenced in this issue's comment thread — apologies for the duplicate effort. #2392 looks ready for review.

Issue #1777, opened 2026-05-10
Windows worktree setup can fail with spawn git ENOENT


Priyanchew:
## Problem

Running AO from some Windows environments could fail during worktree setup with:

`Failed to setup orchestrator: spawn git ENOENT`

This showed up when running:

`ao start`

The failure happened even on machines where Git was installed.

## Why It Happened

The worktree plugin invoked Git as a bare executable name (`git`) via `child_process.execFile`. That works when Git is available on the inherited `PATH`, but Windows shells launched from GUI apps or constrained environments can have a degraded `PATH` that includes Node but not Git.

In that state, Node could start AO, but the worktree plugin could not spawn Git, so setup failed with a raw `ENOENT` instead of finding the installed Git binary or explaining the missing dependency clearly.

## Impact

First-run and orchestrator startup flows that depend on the worktree workspace plugin could fail before creating or attaching worktrees. This was especially likely on Windows, but the fix needed to remain safe for macOS and Linux too.

## Fix in PR #1776

PR #1776 centralized Git executable resolution in `@aoagents/ao-core` with `getGitExecutable()`:

- Search `PATH` first, preserving normal behavior.
- Fall back to common Git install locations on Windows, macOS, and Linux.
- Cache the resolved Git executable to avoid repeated filesystem probes.
- Use the resolved executable in the workspace-worktree plugin.
- Improve the error message if Git still cannot be found.
- Add platform-mocked tests for Windows/macOS/Linux fallback behavior.

somewherelostt:
Closing this issue as it doesn’t seem to be relevant/needed with the current state of AO.

If you think the issue still exists, feel free to reopen it or create a new issue and attach a screenshot or short video so we can verify it and work on a fix.

You can also reach us here:
Discord: https://discord.gg/hRhb5e4Cn
X: https://x.com/aoagents
LinkedIn: https://www.linkedin.com/company/agent-orchestrator
GitHub: https://github.com/Untrivial-ai/agent-orchestrator

Thanks for reporting it and helping us improve AO.

PR #1776, opened 2026-05-10
fix(worktree): resolve Git executable outside PATH

Priyanchew:
## Summary
- Resolve Git via the shared platform layer before worktree operations.
- Fall back to common Git install locations on Windows, macOS, and Linux when PATH is degraded.
- Surface a clearer worktree setup error when Git is truly unavailable.

closes https://github.com/ComposioHQ/agent-orchestrator/issues/1777

## Test plan
- `pnpm --filter @aoagents/ao-core test -- src/__tests__/platform.test.ts`
- `pnpm --filter @aoagents/ao-core typecheck && pnpm --filter @aoagents/ao-core build && pnpm --filter @aoagents/ao-plugin-workspace-worktree typecheck && pnpm --filter @aoagents/ao-plugin-workspace-worktree build`
- Reduced-PATH repro: `getGitExecutable()` resolves `C:\Program Files\Git\cmd\git.exe` and `workspaceExists=true`

PR #3765, opened 2026-08-09
Fix: resolve git binary via PATH with install-location fallback (Windows spawn git ENOENT)

krishrathi1:
## Problem

`gitworktree.New` defaults `opts.Binary` to the bare string `"git"` whenever a caller does not set an explicit `Binary` override. The daemon's own wiring (`backend/internal/daemon/lifecycle_wiring.go`, the `gitworktree.New(gitworktree.Options{...})` call site) does not set `Binary:`, so production always takes this default. That bare string is passed straight into `os/exec` at every git invocation in the package (`w.run(ctx, w.binary, ...)`), relying on the process's inherited PATH to resolve it lazily at spawn time, with no fallback.

On a Windows process whose inherited PATH does not include git's directory (a real, documented scenario for GUI-launched apps with a degraded PATH), every git call fails with a raw `ENOENT` instead of a clear "git not found" error.

## This is a reintroduced regression

This exact defect was already fixed once, pre-rewrite, in the now-deleted `@aoagents/ao-core` TypeScript package via `getGitExecutable()` (PR #1776): search PATH first, fall back to common Git install locations per OS, cache the result, and produce a clear error if nothing is found. That fix did not carry over when this repository was rewritten in Go, so the same bug is present again in the current code.

## Fix

Added `resolveGitBinary` in a new `backend/internal/adapters/workspace/gitworktree/git_binary.go`:

- Searches PATH via `exec.LookPath` first, so the common case (git reachable on PATH) is completely unchanged.
- Falls back, in order, to Git's well-known install locations if PATH lookup fails: Windows (`%ProgramFiles%/Git/cmd` and `/bin`, `%ProgramFiles(x86)%/Git/cmd` and `/bin`, `%LocalAppData%/Programs/Git/cmd`), macOS (Homebrew Apple Silicon/Intel paths, Xcode Command Line Tools), and Linux (standard system paths).
- Returns a clear `ErrGitBinaryNotFound`-wrapped error naming the actual problem (git missing / unreachable) instead of a raw `ENOENT` if nothing resolves.

`New` now calls this once at construction (only when `opts.Binary == ""`) and stores the resolved absolute path on the `Workspace`, so every command in the package reuses one resolved binary for its whole lifetime -- the same caching effect the old TS helper achieved explicitly. `opts.Binary` remains an unconditional override that bypasses resolution entirely, so existing callers/tests that already pass an explicit binary path are unaffected.

The PATH-lookup and fallback-existence checks are both injected (`gitPathLookup`/`gitPathExists`), following the same DI pattern this file already uses for `commandRunner`, so the fallback branch is unit-testable without depending on the real PATH or a real Git install at any specific location on the test machine.

I looked at `binaryutil.ResolveBinary` (`backend/internal/adapters/agent/binaryutil`), which already implements a similar PATH-then-fallback-locations pattern for agent CLIs, before writing this. I did not reuse it directly for two reasons: its `WinBase` candidate model only covers `%APPDATA%`/`%LOCALAPPDATA%`/home-relative paths and has no way to express `%ProgramFiles%`, which is where Git for Windows actually installs by default, so it would not have covered the primary real-world fallback case; and its not-found error wraps `ports.ErrAgentBinaryNotFound` ("agent: binary not found on PATH"), which is pattern-matched elsewhere in the codebase specifically to mean an AI agent CLI is missing -- reusing it for git would produce a misleading message and risk being caught by that agent-specific handling. The resolver added here mirrors `binaryutil`'s overall shape (PATH first, then ordered platform candidates) without those two mismatches.

## Verification

- Confirmed the bug against current code by reading `workspace.go`'s `New()`/`defaultGitBinary` and the `lifecycle_wiring.go` call site (no `Binary:` override), and confirming `aoprocess.CommandContext` is a thin `exec.CommandContext` wrapper with no PATH fallback of its own.
- Added tests in `git_binary_test.go` covering: PATH-preferred resolution (and that the fallback checker is never even consulted when PATH succeeds), fallback-location resolution when PATH lookup is mocked to fail, a clear not-found error naming the problem, the Windows candidate list's order/content and its skipping of unset env bases, `New()`'s end-to-end PATH resolution (git is genuinely on PATH in this environment), the `opts.Binary` override escape hatch, and (Windows-only) `New()`'s end-to-end total-miss error path with PATH and every Windows fallback base env var pointed at empty directories.
- Proved the fix matters via a patch-based fail/pass cycle (not `git stash`): reverse-applied the implementation patch (removing `git_binary.go` and the `New()` wiring change) while keeping the new test file, confirmed `go test` failed to even build (`undefined: resolveGitBinary`, `undefined: ErrGitBinaryNotFound`, etc. -- i.e. there is no fallback to fail into on the old code), then reapplied the patch and confirmed every new test passes.
- `go build ./...` and `go vet ./...` are clean from `backend`.
- Full `gitworktree` package test suite passes, including the pre-existing real-git integration tests (`TestWorkspaceIntegrationCreateRestoreDestroy` and friends, which shell out to actual git) -- confirming normal git worktree operations still work end-to-end with the resolver in the loop.

Addresses #1777.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/git_binary.go
+++ b/backend/internal/adapters/workspace/gitworktree/git_binary.go
@@ -1,0 +1,115 @@
+package gitworktree
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// ErrGitBinaryNotFound is returned when the default "git" binary cannot be
+// located on PATH or in any well-known Git install location. Naming the
+// problem here means callers see "git executable not found" instead of a raw,
+// confusing os/exec ENOENT.
+var ErrGitBinaryNotFound = errors.New("gitworktree: git executable not found")
+
+// gitPathLookup abstracts exec.LookPath so tests can force a PATH miss
+// without mutating the process's real PATH.
+type gitPathLookup func(name string) (string, error)
+
+// gitPathExists abstracts checking whether a fallback candidate path names a
+// usable file, so fallback-location tests don't depend on a real Git install
+// living at that path on the machine running the tests. It mirrors the
+// commandRunner field above: an injectable seam over an OS call so tests
+// don't have to hit the real filesystem/environment to exercise this path.
+type gitPathExists func(path string) bool
+
+// resolveGitBinary finds the git executable New should use when the caller
+// does not supply an explicit Options.Binary override: PATH first (so the
+// common case, where git is reachable, behaves exactly as before), then a
+// platform's well-known Git install locations if PATH lookup fails.
+//
+// This mirrors the pre-rewrite TypeScript getGitExecutable() helper
+// (@aoagents/ao-core, since removed with the rest of that package) that this
+// Go port dropped when the codebase was rewritten: search PATH, fall back to
+// common install locations, and produce a clear error naming the real problem
+// if nothing is found. Go's os/exec resolves a bare "git" against PATH lazily
+// at spawn time with no such fallback, so a process launched with a degraded
+// PATH (e.g. a GUI app on Windows that doesn't inherit a shell profile) fails
+// every git invocation with a bare ENOENT until this resolves once up front.
+//
+// The result is resolved once, in New, and stored on the Workspace as
+// w.binary, which every command in this package already reuses for its whole
+// lifetime — that per-instance reuse is the caching the TS helper did
+// explicitly with its own memoized cache.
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
@krishrathi1 
Issue #1777 was closed on Aug 22 with the note that it does not seem relevant or needed with the current state of AO, and the git binary fallback approach never landed on main. Closing.

Issue #2173, opened 2026-06-24
bug(lifecycle): merge-conflict on a PR is detected but never surfaced — no nudge, no notification, no status (suppressed for waiting_input sessions)


neversettle17-101:
Migrated from aoagents/ReverbCode/issues/295

Source: https://github.com/aoagents/ReverbCode/issues/295
Original assignees: harshitsinghbhandari
Original labels: bug, lcm-sm, priority: high

---

## Bug

PR #278 (`ao/agent-orchestrator-2` → `main`) is `CONFLICTING` / `DIRTY` on GitHub. AO **did detect it** — the stored PR facts read `mergeability: "conflicting"`:

```
GitHub:  mergeable=CONFLICTING  mergeStateStatus=DIRTY
AO  /sessions/agent-orchestrator-2/pr → { "mergeability": "conflicting", "ci": "passing", "review": "approved", ... }
```

But AO never **surfaces** the conflict: no agent nudge, no user notification, and no attention status/badge. To the user it looks like AO "didn't detect it."

**Source:** local observation, PR #278 | **Reported by:** @aditi1178 | **Analyzed against:** `73649b0`
**Confidence:** High

## Reproduction

1. Worker opens a PR, finishes its turn, and goes `waiting_input` (`needs_input`).
2. The base branch (`main`) advances so the PR becomes `CONFLICTING`.
3. The SCM observer detects it and stores `mergeability: conflicting`.
4. No nudge, no notification, no status change — the conflict is invisible unless you open the PR-facts panel and read a single mono row.

## Root Cause

AO detects the conflict (data layer is correct), but all three surfacing paths are suppressed:

### 1. Agent nudge is gated off for `waiting_input`
`backend/internal/lifecycle/reactions.go:65` short-circuits **every** PR reaction (CI-fail, review-feedback, AND merge-conflict) when the session is paused:
```go
if rec.IsTerminated || rec.Activity.State == domain.ActivityWaitingInput {
    return nil
}
```
The merge-conflict nudge (`reactions.go:90-103`, *"Your PR has merge conflicts. Rebase onto the base branch and resolve them."*) sits **below** that gate, so for a `waiting_input` session it never runs. PR #278's session is exactly `waiting_input`.

### 2. No user notification type for conflicts
[...]

PR #3766, opened 2026-08-09
fix(lifecycle): make merge-conflict nudges reach needs-input sessions and re-fire after a conflict returns

### LINKED ISSUE #4528: Merge-conflict nudge does not re-fire after conflict is resolved and later returns
## Bug

AO notifies a worker the first time a tracked PR becomes conflicting, but does not notify it when that conflict is resolved and a later base-branch advance makes the same PR conflicting again.

**Reported from:** AO worker session `agent-orchestrator-149` while handling PR #4165
**Observed:** 2026-08-26
**Analyzed against:** `4cbb4b6ce`
**AO:** locally built `dev`, daemon healthy on port 3001
**Confidence:** High
**Priority:** P2 — missed automation notification; the user can manually steer the worker as a workaround.

## Reproduction

1. Track an open PR owned by a live worker session.
2. Make the PR conflicting and allow the SCM observer to deliver the merge-conflict nudge.
3. Rebase the PR so it is clean/mergeable.
4. Advance the base branch so the same PR becomes conflicting again.
5. Observe that AO does not deliver a second merge-conflict nudge.

## Root cause

`backend/internal/lifecycle/reactions.go` queues merge-conflict nudges with key `merge-conflict:<pr-url>` and the constant signature `conflicting`. `sendOnce` persists that key/signature in `pr.last_nudge_signature`.

A later non-conflicting observation does not clear or re-arm the merge-conflict dedup entry. When conflict returns, `sendOnce` sees the same key/signature and treats it as already delivered. The persisted state makes the suppression survive daemon restarts.

Live state for PR #4165 confirmed:

- durable mergeability: `conflicting`
- seen signature: `merge-conflict:https://github.com/Untrivial-ai/agent-orchestrator/pull/4165 = conflicting`
- attempts: `1`

This exactly explains why the first conflict notification arrived and the later one did not.

## Fix

Re-arm the merge-conflict reaction after a definitive non-conflicting observation (but not an `unknown` transient): remove/reset that PR-specific seen signature and attempt count, persist the updated reaction payload, and add lifecycle plus observer integration coverage for:

`mergeable -> conflicting (notify) -> mergeable (re-arm) -> conflicting (notify again)`

Identical consecutive conflicting observations must remain deduplicated, including across daemon restart.

krishrathi1:
This PR carries two related merge-conflict-nudge fixes. Both concern the same
`merge-conflict:<pr-url>` reaction in `backend/internal/lifecycle/reactions.go`,
and each one on its own still leaves the nudge silent in a common case.

Closes #4528. Addresses #2173 (the agent-nudge surfacing gap; the
notification-type and frontend-status gaps there are left for follow-up).

---

## 1. Merge-conflict nudges reach needs-input sessions

**Root cause.** `cannotNudge()` gates every reaction nudge on
`rec.Activity.State.NeedsInput()`, and `ApplyPRObservation` called it at the top
of the function — before the merge-conflict block that runs later in the same
function. A session parked at `waiting_input` (or `blocked`) whose PR went
`conflicting` produced no nudge, and since the reaction only starts its
`escalateAfter` timer on dispatch, no later escalation either. Completely
silent.

The pre-rewrite TypeScript implementation had the identical gap by a different
mechanism: `maybeDispatchMergeConflicts` gated on a status allowlist that
omitted the parked state, directly under a doc comment claiming it was
"dispatched independently of the session status." The same defect surviving two
separate implementations argues for a deliberate, named carve-out rather than
something that happens to fall out of the next refactor.

**Why this is a carve-out, not a general loosening.** CI-failure and
review-feedback nudges are actionable only by the agent, so deferring them until
it resumes loses nothing. A merge conflict is different: the human sitting at
that needs-input prompt may be exactly who needs to act (rebase it themselves,
or redirect the agent), so it should reach them immediately.

**The fix has two layers.**

- **`lifecycle/reactions.go`** — `ApplyPRObservation` now only short-circuits up
  front on `IsTerminated`/`ActivityExited` (a session with nowhere to deliver
  anything). It evaluates `needsInput` separately and uses it to gate just the
  CI-failure/unresolved-comment/review-changes-requested blocks, leaving the
  merge-conflict block unconditional. `cannotNudge()` itself is untouched, and
  every other nudge path (`ApplyReviewBatch`, `ApplyTrackerFacts`) still calls
  it exactly as before.
- **`sessionguard/guard.go`** — that alone is not sufficient. `sendOnce`
  delivers through `sessionguard.Guard.Nudge`, which independently refuses any
  write while `NeedsInput()` is true regardless of what the caller decided; it
  is the just-in-time re-check that closes the race between the reducer's
  snapshot and the actual pane write. Loosening the reducer's gate alone would
  have been a no-op. Added `Guard.NudgeUrgent`, which refuses while `Blocked` (a
  live permission dialog) and while a TUI session has not yet signalled startup,
  and — crucially — refuses at `waiting_input` unless the harness *positively
  proves* that prompt is a genuine idle composer rather than a masked permission
  decision (see the harness-aware gate below). The merge-conflict `pendingNudge`
  now carries an `urgent` flag routing it through `NudgeUrgent`; every other
  nudge type is unchanged.

## 2. Re-arm the nudge once a conflict clears (#4528)

**Root cause.** `sendOnce` persists the dedup entry
`"merge-conflict:<pr-url>" = "conflicting"` in `pr.last_nudge_signature`, and
nothing ever cleared it. Once the agent rebased a PR clean, a later base-branch
advance that made the same PR conflicting again produced the identical
key/signature and was swallowed as a duplicate. Because the signature is
durable, the suppression also survived daemon restarts. Confirmed live on
PR #4165 in the issue: durable mergeability `conflicting`, seen signature
`conflicting`, attempts `1`, and no second nudge.

**The fix.** `ApplyPRObservation` re-arms on the other edge of the transition: a
definitively non-conflicting observation drops that one PR's merge-conflict seen
signature and attempt count and persists the updated payload, so the next
conflicting observation notifies again.

What counts as definitive matters, and `mergeabilityClearsConflict` names it:

| state | re-arms? | why |
|---|---|---|
| `mergeable`, `unstable` | yes | both require a computed provider rollup that ruled conflicts out first |
| `unknown` (and the unset zero value) | no | the transient GitHub reports while recomputing after a push or retarget; re-arming here would defeat the dedup outright, since a conflict that never went away flaps `unknown → conflicting` and would re-nudge every poll |
| `blocked` | no | despite not literally being a conflict, the observer synthesizes it locally from draft / failing-CI / changes-requested facts (`mergeabilityFromProviderFacts`), so it is reported even while provider mergeability is still unknown — not proof the conflict is gone |

Excluding `blocked` has a known cost, and this PR demonstrated it on itself: AO
observed this PR as `conflicting` before it was rebased, and it is now
`blocked` (awaiting required review), so its merge-conflict dedup entry stays
armed rather than resetting. That is harmless here — there is no conflict, so
nothing should re-notify — but it means a *real* conflict appearing while a PR
is still `blocked` would be deduplicated against the stale `conflicting`
signature and stay silent until the PR reaches `mergeable`/`unstable` once.
Widening the re-arm to `blocked` would trade that for the flap-spam this issue
explicitly requires be prevented, so the narrower rule stands; flagging it as
the deliberate trade-off rather than an oversight.

The re-arm's durable write is retried until it lands. The in-memory delete
takes effect immediately (a conflict returning in-process must still nudge), so
a failed `UpdatePRLastNudgeSignature` would otherwise leave the next
observation seeing nothing armed, taking the early return, and never retrying —
leaving the stale `conflicting` signature on disk for a restart to reload.
`reactionState.pendingRearm` marks such PRs so every later non-conflicting
observation retries until the write succeeds. Rolling the in-memory delete back
instead would fix the restart case but deduplicate away a conflict that returns
before the write lands.

Blast radius is deliberately narrow. The re-arm deletes only this PR's
merge-conflict key, so the CI and review-feedback entries on the same PR row
keep their signatures and a resolved conflict cannot replay unrelated nudges. It
loads the persisted payload before mutating (after a restart the in-memory maps
are empty, and persisting from them alone would erase every other key the row
holds), skips the store write when nothing was armed so a steadily mergeable PR
does not rewrite an unchanged payload on every poll, and defers its error past
the send loop like `blockedCheckErr` so a store failure cannot discard the
CI/review nudges queued in the same observation.

## 3. Review follow-up: two boundary fixes

Two correctness boundaries surfaced in review, each fixed with its own coverage.

**Harness-aware urgent delivery.** `NudgeUrgent` originally reused
`refuseDeliver`, which refuses only on `Blocked` — so *every* `waiting_input`
session became writable. That is not equivalent to an idle composer for several
adapters: codex maps `permission-request` to `waiting_input`, Droid folds both
permission decisions and idle notifications into `waiting_input`, and the
shared-hook harnesses (`StandardDeriveActivityState`) do the same. An urgent
paste+Enter there could answer a live permission prompt. `NudgeUrgent` now takes
an `acceptsWaitingInput func(domain.AgentHarness) bool` predicate (mirroring
`CoordinationUnderMutation`'s existing gating) and is **fail-closed**: it refuses
`waiting_input` unless the predicate proves it safe, and a nil predicate refuses.
Lifecycle wires it via a new `WithUrgentNudgeGate` option, resolved in
`lifecycle_wiring.go` from `ports.BlockedActivitySignaler.EmitsBlockedActivity()`
— the exact contract for "reports permission dialogs as blocked rather than
waiting_input." Only claude-code and kimchi open the boundary; codex, Droid,
agy, prime-agent, and every shared-hook harness stay suppressed.

**Re-arm ahead of the dead-session return.** The `mergeabilityClearsConflict`
re-arm originally ran after `ApplyPRObservation`'s terminated/exited early
return. Because exited sessions are still polled (and a terminated session
resumes polling once restored), the sequence `conflicting → exit → mergeable →
restore → conflicting` kept the durable `conflicting` signature and the
recurrence was deduplicated away. Re-arming is non-delivery bookkeeping — it only
drops this PR's dedup entry in the store — so it now runs *before* the
dead-session gate, with its persist error still deferred past the nudge send
loop so a store failure cannot discard CI/review nudges an unstable PR queues.

## Test coverage

`internal/sessionguard/guard_test.go`

- `TestGuard_NudgeUrgentReachesNeedsInputButNotBlocked` — pins the new method's
  outcome table (sent at `active`/`idle` and at capability-safe `waiting_input`;
  suppressed at ambiguous/nil-predicate `waiting_input`, and at
  `blocked`/`exited`/`terminated`/missing).
- `TestGuard_NudgeUrgentRespectsTUIStartupGate` — the carve-out does not widen
  past its one intended state.

`internal/lifecycle/manager_test.go`

- `TestPRObservation_MergeConflictReachesNeedsInputSession` — a `waiting_input`
  session on a blocked-signalling harness now gets the nudge; the same state on
  an ambiguous harness (codex-like) and a `blocked` session still do not.
- `TestPRObservation_NeedsInputSessionStillWithholdsOtherNudges` — regression
  guard: a needs-input session still gets **no** CI-failure or review-feedback
  nudge.
- `TestPRObservation_DeadSessionGetsNoNudgesOfAnyKind` — a terminated or exited
  session still gets nothing at all, conflict included.
- `TestPRObservation_ReArmClearsConflictWhileSessionExited`,
  `TestPRObservation_ReArmClearsConflictSurvivesTerminatedRestore` — the
  re-arm runs ahead of the dead-session return, so `conflicting → exit/terminate
  → mergeable → restore → conflicting` nudges again instead of being swallowed.
- `TestPRObservation_MergeConflictReArmsAfterConflictClears` — the
  `mergeable → conflicting (notify) → mergeable (re-arm) → conflicting (notify
  again)` cycle, for both re-arming states.
- `TestPRObservation_ConsecutiveMergeConflictsStayDeduplicated` — identical
  consecutive conflicts still nudge exactly once.
- `TestPRObservation_UnknownMergeabilityDoesNotReArm` — `unknown`, `blocked` and
  the unset zero value do not re-arm.
- `TestPRObservation_MergeConflictDedupSurvivesRestart` — both directions across
  a fresh `Manager` over the same store: an unchanged conflict stays quiet, a
  cleared one is free to nudge again.
- `TestPRObservation_ReArmLeavesOtherDedupEntriesIntact`,
  `TestPRObservation_ReArmSkipsStoreWriteWhenNothingArmed`,
  `TestPRObservation_ReArmStoreFailureSurfacesWithoutLosingNudges` — blast
  radius, write-avoidance, and the deferred-error contract.
- `TestPRObservation_ReArmRetriesAfterFailedPersist` — failed persist →
  recovered observation → restart → conflict still nudges.
- `TestPRObservation_ReArmStillNudgesInProcessAfterFailedPersist` — a conflict
  returning before the retry lands is not deduplicated away in-process.

`internal/daemon/wiring_test.go`

- `TestWiring_MergeConflictNudgeReArmsAfterConflictClears` — the same cycle end
  to end through the observer's `ApplySCMObservation` entrypoint over the real
  sqlite store, plus the restart. A fake store cannot prove the
  `pr.last_nudge_signature` round trip actually survives the real column.
- `TestWiring_UrgentNudgeGateComesFromAdapters` — the urgent-nudge gate resolved
  against the real agent registry is fail-closed: claude-code/kimchi open the
  `waiting_input` boundary; codex, Droid, the shared-hook harnesses
  (goose/devin), an unknown harness, and a nil resolver stay suppressed.

## Verification

- Both new behaviors were confirmed to fail without the change: disabling the
  re-arm branch fails `MergeConflictReArmsAfterConflictClears` (both subtests),
  `MergeConflictDedupSurvivesRestart/cleared_conflict_re-arms_across_restart`,
  and the daemon integration test, while every dedup/regression test keeps
  passing.
- `gofmt -l .`, `go build ./...`, `go vet ./...` clean from `backend/`.
- `go test ./internal/lifecycle/... ./internal/sessionguard/... ./internal/observe/... ./internal/daemon/` — all pass.
- `go test -race ./internal/lifecycle/... ./internal/sessionguard/... ./internal/daemon/` — all pass.

## Rebase note

The original commit was rebased onto current `main`, which had moved under it:
the CI block now goes through the per-PR `AutoInjectCI` policy with a deferred
`ciPolicyErr`, the merge-conflict block gained the `prBlockedByOpenParent` stack
check with a deferred `blockedCheckErr`, and `sessionguard` factored its
policies into `refuseDeliver`/`refuseNudge`. Original authorship is preserved on
the first commit. `NudgeUrgent` initially reused `refuseDeliver`; review
follow-up (section 3) replaced that with an explicit harness-aware
`waiting_input` gate.

### DIFF
--- a/backend/internal/lifecycle/reactions.go
+++ b/backend/internal/lifecycle/reactions.go
@@ -283,7 +283,9 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	// defer the error past the send loop, still surfacing it so the observer logs
 	// and re-polls.
 	var blockedCheckErr error
-	if o.Mergeability == domain.MergeConflicting {
+	var rearmErr error
+	switch {
+	case o.Mergeability == domain.MergeConflicting:
 		// Only the bottom of a stack is eligible for the rebase nudge. A PR
 		// stacked on an open parent is expected to report conflicts against its
 		// parent branch until the parent merges and it retargets, so nudging the
@@ -298,8 +300,16 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 			if o.URL != "" {
 				msg += "\nPR: " + domain.SanitizeControlChars(o.URL)
 			}
-			nudges = append(nudges, pendingNudge{key: "merge-conflict:" + o.URL, sig: string(o.Mergeability), msg: msg, maxAttempts: 0, urgent: true})
+			nudges = append(nudges, pendingNudge{key: mergeConflictKey(o.URL), sig: string(o.Mergeability), msg: msg, maxAttempts: 0, urgent: true})
 		}
+	case mergeabilityClearsConflict(o.Mergeability):
+		// The conflict is definitively gone, so drop this PR's merge-conflict
+		// dedup entry (#4528). Without the re-arm the "conflicting" signature
+		// persists forever and the next real conflict — typically a base-branch
+		// advance after the agent rebased clean — is silently swallowed as a
+		// duplicate. Deferred like blockedCheckErr so a store failure here
+		// cannot discard the CI/review nudges already queued above.
+		rearmErr = m.rearmMergeConflict(ctx, o.URL)
 	}
 
 	for _, n := range nudges {
@@ -312,7 +322,10 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	if ciPolicyErr != nil {
 		return ciPolicyErr
 	}
-	return blockedCheckErr
+	if blockedCheckErr != nil {
+		return blockedCheckErr
+	}
+	return rearmErr
 }
 
 func (m *Manager) terminateCompletedSession(ctx context.Context, id domain.SessionID) error {
@@ -349,6 +362,66 @@ func (m *Manager) sessionComplete(ctx context.Context, id domain.SessionID) (boo
 		}
 	}
 	return merged, nil
+}
+
+// mergeConflictKey is the reaction-dedup key for a PR's merge-conflict nudge.
+// The send path and the re-arm path share it so the two cannot drift.
+func mergeConflictKey(prURL string) string { return "merge-conflict:" + prURL }
+
+// mergeabilityClearsConflict reports whether an observation positively
+// establishes that a PR is no longer conflicting, which is what re-arms its
+// merge-conflict nudge (#4528).
+//
+// Only states the provider actually computed count. `unknown` is the transient
+// GitHub reports while it recomputes mergeability after a push or a retarget;
+// re-arming on it would defeat the dedup entirely, since a conflict that never
+// went away flaps unknown → conflicting and would re-nudge on every poll.
+// `blocked` is excluded on the same grounds even though it is not literally a
+// conflict: the observer synthesizes it locally from draft / failing-CI /
+// changes-requested facts (see mergeabilityFromProviderFacts), so it is reported
+// even while provider mergeability is still unknown and cannot be read as proof
+// the conflict is gone. `mergeable` and `unstable` both require a computed
+// provider rollup that ruled conflicts out first, so both are definitive.
+func mergeabilityClearsConflict(state domain.Mergeability) bool {
+	return state == domain.MergeMergeable || state == domain.MergeUnstable
+}
+
+// rearmMergeConflict clears the merge-conflict dedup entry for prURL so a later
+// conflicting observation notifies again. sendOnce otherwise keeps the
+// "merge-conflict:<url>" = "conflicting" signature forever: a PR rebased clean
+// and then made conflicting again by a base-branch advance stayed silent, and
+// because the signature is persisted in pr.last_nudge_signature the suppression
+// survived daemon restarts (#4528).
+//
+// Only this PR's merge-conflict key is touched, so the CI and review-feedback
+// entries for the same PR keep their signatures and resolving a conflict cannot
+// replay unrelated nudges. The persisted payload is loaded first: after a
+// restart the in-memory maps are empty, and persisting from them alone would
+// erase every other key the PR row still holds.
+func (m *Manager) rearmMergeConflict(ctx context.Context, prURL string) error {
+	if m.guard == nil || prURL == "" {
+		return nil
+	}
+	m.react.mu.Lock()
+	defer m.react.mu.Unlock()
+
+	if !m.react.loaded[prURL] {
+		if err := m.loadPRSignaturesLocked(ctx, prURL); err != nil {
+			return err
+		}
+		m.react.loaded[prURL] = true
+	}
+	key := mergeConflictKey(prURL)
+	_, seen := m.react.seen[key]
+	_, attempted := m.react.attempts[key]
+	if !seen && !attempted {
+		// Nothing is armed, so skip the store write: a steadily mergeable PR
+		// must not re-persist an unchanged payload on every poll.
+		return nil
+	}
+	delete(m.react.seen, key)
+	delete(m.react.attempts, key)
+	return m.persistPRSignaturesLocked(ctx, prURL)
 }
 
 // prBlockedByOpenParent reports whether the PR at prURL is stacked on top of
--- a/backend/internal/lifecycle/reactions.go
+++ b/backend/internal/lifecycle/reactions.go
@@ -110,10 +110,17 @@ type reactionState struct {
 	// seen/attempts during this process. Lazy: we only pay the DB read on the
 	// first reaction touching each PR after startup.
 	loaded map[string]bool
+	// pendingRearm tracks PR URLs whose merge-conflict re-arm has been applied
+	// in memory but whose durable write failed. The in-memory delete stands (a
+	// conflict returning in this process must still nudge), so without this the
+	// next re-arm would see nothing armed, take the early return, and never
+	// retry the store — leaving the stale "conflicting" signature on disk to
+	// suppress the nudge after a restart. Cleared once the write lands.
+	pendingRearm map[string]bool
 }
 
 func newReactionState() reactionState {
-	return reactionState{seen: map[string]string{}, attempts: map[string]int{}, loaded: map[string]bool{}}
+	return reactionState{seen: map[string]string{}, attempts: map[string]int{}, loaded: map[string]bool{}, pendingRearm: map[string]bool{}}
 }
 
 // reactionPayload is the JSON document persisted in pr.last_nudge_signature.
@@ -414,14 +421,26 @@ func (m *Manager) rearmMergeConflict(ctx context.Context, prURL string) error {
 	key := mergeConflictKey(prURL)
 	_, seen := m.react.seen[key]
 	_, attempted := m.react.attempts[key]
-	if !seen && !attempted {
-		// Nothing is armed, so skip the store write: a steadily mergeable PR
-		// must not re-persist an unchanged payload on every poll.
+	if !seen && !attempted && !m.react.pendingRearm[prURL] {
+		// Nothing is armed and no earlier re-arm is still owed a durable write,
+		// so skip the store write: a steadily mergeable PR must not re-persist
+		// an unchanged payload on every poll.
 		return nil
 	}
 	delete(m.react.seen, key)
 	delete(m.react.attempts, key)
-	return m.persistPRSignaturesLocked(ctx, prURL)
+	if err := m.persistPRSignaturesLocked(ctx, prURL); err != nil {
+		// The in-memory delete deliberately stands: a conflict returning later
+		// in this process must still nudge. But disk still holds the stale
+		// "conflicting" signature, which a restart would reload and use to
+		// suppress that nudge — the exact bug this re-arm exists to fix. Mark
+		// the PR so every later non-conflicting observation retries the write
+		// until it lands, instead of taking the early return above.
+		m.react.pendingRearm[prURL] = true
+		return err
+	}
+	delete(m.react.pendingRearm, prURL)
+	return nil
 }
 
 // prBlockedByOpenParent reports whether the PR at prURL is stacked on top of
--- a/backend/internal/lifecycle/reactions.go
+++ b/backend/internal/lifecycle/reactions.go
@@ -182,10 +182,26 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	if err != nil || !ok {
 		return err
 	}
+	// Re-arm the merge-conflict dedup on a definitively-cleared observation
+	// BEFORE the dead-session delivery gate below. Re-arming is non-delivery
+	// bookkeeping — it only drops this PR's dedup entry in the store, it writes
+	// nothing to a pane — so it must run even for a session that has exited or is
+	// terminated. Otherwise conflicting -> agent exits -> mergeable -> agent
+	// restored -> conflicting keeps the durable "conflicting" signature and
+	// sendOnce swallows the recurrence: exited sessions are still polled, and a
+	// terminated session that is later restored resumes polling, so the cleared
+	// state must land regardless of current liveness. The error is deferred past
+	// the nudge send loop (returned as rearmErr) so a persist failure here cannot
+	// discard CI/review nudges an unstable PR still queues below.
+	var rearmErr error
+	if mergeabilityClearsConflict(o.Mergeability) {
+		rearmErr = m.rearmMergeConflict(ctx, o.URL)
+	}
 	// A genuinely dead session — terminated, or its pane already exited to a
-	// shell — has nowhere to deliver any nudge, merge-conflict included.
+	// shell — has nowhere to deliver any nudge, merge-conflict included, so
+	// return once the non-delivery re-arm above has run.
 	if rec.IsTerminated || rec.Activity.State == domain.ActivityExited {
-		return nil
+		return rearmErr
 	}
 	// cannotNudge's remaining condition, needs-input, defers CI-failure and
 	// review-feedback nudges until the agent is active again: both are
@@ -195,9 +211,11 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	// or redirect the agent), so the merge-conflict nudge below is deliberately
 	// exempted from this gate instead of loosening it for every nudge type. It
 	// still funnels through sendOnce's urgent path (sessionguard.NudgeUrgent),
-	// which continues to refuse a write while the session is blocked on a live
-	// permission dialog — only an idle needs-input prompt is safe to paste into
-	// unsolicited.
+	// which refuses while the session is blocked on a live permission dialog and
+	// while it sits at a waiting_input prompt on a harness that cannot prove that
+	// prompt is a genuine idle composer rather than a masked permission decision
+	// (the urgentNudgeWaitingInputSafe gate) — so only a provably-idle needs-input
+	// prompt is pasted into unsolicited.
 	needsInput := rec.Activity.State.NeedsInput()
 	// A single PR can trip several actionable conditions at once (failing CI,
 	// unresolved review comments, a merge conflict). Queue every applicable nudge
@@ -290,9 +308,7 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	// defer the error past the send loop, still surfacing it so the observer logs
 	// and re-polls.
 	var blockedCheckErr error
-	var rearmErr error
-	switch {
-	case o.Mergeability == domain.MergeConflicting:
+	if o.Mergeability == domain.MergeConflicting {
 		// Only the bottom of a stack is eligible for the rebase nudge. A PR
 		// stacked on an open parent is expected to report conflicts against its
 		// parent branch until the parent merges and it retargets, so nudging the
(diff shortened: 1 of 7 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/lifecycle/reactions.go
```diff
@@ -328,6 +364,66 @@ func (m *Manager) sessionComplete(ctx context.Context, id domain.SessionID) (boo
...
+	_, seen := m.react.seen[key]
+	_, attempted := m.react.attempts[key]
+	if !seen && !attempted {
+		// Nothing is armed, so skip the store write: a steadily mergeable PR
+		// must not re-persist an unchanged payload on every poll.
+		return nil
+	}
+	delete(m.react.seen, key)
+	delete(m.react.attempts, key)
+	return m.persistPRSignaturesLocked(ctx, prURL)
```
AgentWrapper: [P2] Keep the re-arm pending when persistence fails. These entries are deleted from the in-memory maps before `UpdatePRLastNudgeSignature` runs. If that write returns an error, the next mergeable observation sees neither entry and takes the `Nothing is armed` early return, so it never retries the durable update. A daemon restart before the conflict returns then reloads the stale `conflicting` signature and suppresses the next nudge, recreating the exact bug this change is intended to fix. Restore the entries on error (or otherwise retain dirty state until persistence succeeds), and cover failure -> successful observation -> restart -> conflict.

AgentWrapper (commented):
Changes requested. The urgent needs-input routing is narrowly scoped and well-covered, but the re-arm path can permanently leave stale durable dedup state after a transient store write failure. Please fix the inline correctness issue and add regression coverage for recovery across a restart.

backend/internal/lifecycle/reactions.go
AgentWrapper: Good catch — confirmed and fixed in d3fb76d.

I went with the second option you offered (retain dirty state) rather than restoring the entries, because rollback trades this bug for a different one: if the conflict returns before the write succeeds, the restored `conflicting` signature deduplicates the nudge away inside the running daemon. I implemented the rollback variant to check, and it fails `TestPRObservation_ReArmStillNudgesInProcessAfterFailedPersist` while passing the restart case.

So the in-memory delete still stands, and `reactionState` gains `pendingRearm` — the set of PR URLs whose re-arm is applied in memory but still owed a durable write. A failed persist marks the PR, and the `nothing is armed` early return now also checks the mark, so every later non-conflicting observation retries until the write lands and clears it.

Coverage is the path you specified, failure -> successful observation -> restart -> conflict, in `TestPRObservation_ReArmRetriesAfterFailedPersist`; it fails with `a re-arm that retried its persist must survive a restart, got []` against the unmarked version. `TestPRObservation_ReArmStillNudgesInProcessAfterFailedPersist` pins the in-process half that rules out rollback.

AgentWrapper:
Re: review 5058709398 (submitted against `fde9a2c7`) — the P2 is addressed in d3fb76d, and the inline thread is resolved.

**Keep the re-arm pending when persistence fails.** Confirmed and fixed. `rearmMergeConflict` deleted both in-memory entries before `UpdatePRLastNudgeSignature` ran, so a failed write left the next non-conflicting observation seeing nothing armed, taking the `nothing is armed` early return, and never retrying — with the stale `conflicting` signature still on disk for a restart to reload.

Of the two remedies you offered I took the second (retain dirty state) rather than restoring the entries, because rollback trades this bug for another one: the restored `conflicting` signature deduplicates away a conflict that returns before the write succeeds. I implemented the rollback variant to check rather than assume — it passes the restart case but fails `TestPRObservation_ReArmStillNudgesInProcessAfterFailedPersist`.

So the in-memory delete still stands, and `reactionState` gains `pendingRearm`: the set of PR URLs whose re-arm is applied in memory but still owed a durable write. A failed persist marks the PR, and the early return now also consults the mark, so every later non-conflicting observation retries until the write lands and clears it.

Coverage is the path you specified:

- `TestPRObservation_ReArmRetriesAfterFailedPersist` — failed persist → recovered observation → restart → conflict. Fails against the unmarked version with `a re-arm that retried its persist must survive a restart, got []`.
- `TestPRObservation_ReArmStillNudgesInProcessAfterFailedPersist` — the in-process half that rules out rollback.

`gofmt` and `go vet ./...` clean; `go test -race` green on `internal/lifecycle`, `internal/sessionguard`, and `internal/daemon`.

AgentWrapper (commented):
Ready to merge. The follow-up commit fixes the persistence-retry issue from my previous review: failed re-arms remain marked for a later durable retry without restoring the in-process dedup entry, and the new tests cover both recovery across restart and a conflict returning before persistence recovers. I found no additional correctness issues in the updated diff.

backend/internal/sessionguard/guard.go
```diff
@@ -280,6 +280,19 @@ func (g *Guard) Nudge(ctx context.Context, id domain.SessionID, msg string) (Out
...
+// while the session sits idle at a needs-input prompt (waiting_input) —
+// reserved for alerts where a human parked at that prompt may be exactly who
+// needs to act (e.g. a merge conflict needing a rebase or a redirected agent),
+// unlike a routine reaction nudge that can simply wait for the agent to resume
+// on its own. It reuses refuseDeliver, the same policy real user messages take:
+// it still refuses while the session is blocked on a live permission decision
+// (an unsolicited write there risks answering the dialog rather than merely
+// being read late) and while a TUI session has not yet signalled startup.
+func (g *Guard) NudgeUrgent(ctx context.Context, id domain.SessionID, msg string) (Outcome, error) {
+	return g.send(ctx, id, msg, g.refuseDeliver)
```
whoisasx: [P1] Preserve the permission-prompt guard for ambiguous `waiting_input` states. `NudgeUrgent` reuses `refuseDeliver`, so every `waiting_input` session becomes writable. That is not equivalent to an idle composer for several current adapters: Codex maps `permission-request` to `waiting_input` and explicitly relies on automated nudges remaining suppressed; the shared `StandardDeriveActivityState` does the same for its adapters; and Droid maps both permission decisions and idle notifications to `waiting_input` because it cannot distinguish them. Injecting this message plus Enter can therefore act on a live permission prompt on the user's behalf. Please keep urgent delivery fail-closed unless the adapter/runtime positively proves this `waiting_input` boundary is safe (the existing `WaitingInputComposerReadiness` and `EmptyComposerDetector` contracts provide the relevant pattern), or surface the conflict through the notification channel instead of pane injection. Please also add adapter-specific regression coverage for at least Codex, Droid, and a shared-hook harness.

backend/internal/lifecycle/reactions.go
```diff
@@ -169,18 +182,23 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 	if err != nil || !ok {
 		return err
 	}
-	if cannotNudge(rec) {
+	// A genuinely dead session — terminated, or its pane already exited to a
+	// shell — has nowhere to deliver any nudge, merge-conflict included.
+	if rec.IsTerminated || rec.Activity.State == domain.ActivityExited {
```
whoisasx, line 203: [P2] Re-arm conflict state before returning from the delivery path. This return suppresses nudges for dead sessions, but it also skips the non-delivery bookkeeping in `rearmMergeConflict` when a definitively `mergeable` or `unstable` observation arrives. Exited sessions remain eligible for SCM observation, and the observer acknowledges this call because it returns nil. As a result, `conflicting -> agent exits -> mergeable -> agent restored -> conflicting` retains the original durable `conflicting` signature and `sendOnce` suppresses the recurrence. Please move the clear-state re-arm ahead of this delivery eligibility return, and cover the exited/restore sequence; terminated-session restore should also be considered because those sessions are excluded from polling while terminated.

whoisasx (changes requested):
Requesting changes on two remaining correctness and safety boundaries. The follow-up commit correctly addresses the earlier persistence-retry finding: a failed re-arm remains dirty for a later durable retry without restoring the in-process conflict signature, and the added coverage exercises both restart recovery and a conflict returning before persistence recovers. However, the new urgent route currently bypasses the permission-prompt safety contract for harnesses that deliberately represent ambiguous permission states as `waiting_input`, and the dead-session return prevents a cleared conflict from re-arming while an exited session is still being observed. Both can produce serious user-facing failures: unintended input at a permission boundary or a later conflict being silently suppressed. Once these are addressed with harness-aware delivery gating and exited/restore lifecycle coverage, the rest of the implementation is narrow, carefully reasoned, and well tested. I also verified that the PR merges cleanly over current `main`, and the resulting tree passes `go build ./...` and `git diff --check`. GitHub checks are green. Per review policy, I did not run Go tests.

backend/internal/lifecycle/reactions.go
AgentWrapper: Fixed in 122592563. The definitively-cleared re-arm now runs *before* the terminated/exited delivery return in `ApplyPRObservation`: re-arming is non-delivery bookkeeping (it only drops this PR's dedup entry in the store), so it must land regardless of current liveness. Its persist error is still deferred past the nudge send loop (returned as `rearmErr`) so a store failure can't discard CI/review nudges an unstable PR queues.

Added two regression tests: `TestPRObservation_ReArmClearsConflictWhileSessionExited` (conflicting → exit → mergeable-while-exited → restore → conflicting must nudge again) and `TestPRObservation_ReArmClearsConflictSurvivesTerminatedRestore` (same across a terminated window). Both fail against the old ordering and pass now.

backend/internal/sessionguard/guard.go
AgentWrapper: Fixed in 122592563 (+ adapter coverage in a72870377). Urgent delivery is now fail-closed at `waiting_input` unless the adapter positively proves the boundary is safe.

`NudgeUrgent` no longer reuses `refuseDeliver`. It takes an `acceptsWaitingInput func(domain.AgentHarness) bool` predicate (mirroring `CoordinationUnderMutation`'s existing gating) and refuses `waiting_input` unless that predicate is true; a nil predicate is treated as "cannot distinguish" and refuses. Lifecycle wires the predicate via a new `WithUrgentNudgeGate` option, resolved in `lifecycle_wiring.go` from `ports.BlockedActivitySignaler.EmitsBlockedActivity()`.

I used `EmitsBlockedActivity` as the positive proof because it directly answers the exact hazard: a harness that reports permission prompts AS `blocked` (claude-code, kimchi) genuinely means "idle composer" when it says `waiting_input`; a harness that folds the decision into `waiting_input` does not. That fails closed for every adapter you named:

- **Codex** — `EmitsBlockedActivity() = false` → suppressed.
- **Droid** — maps every Notification (permission decision *or* idle) to `waiting_input`, implements no blocked signal → suppressed.
- **Shared-hook harnesses** (goose, devin, qwen, copilot, kimi, cline, kiro, kilocode, autohand via `StandardDeriveActivityState`) — no `BlockedActivitySignaler` → suppressed.
- **agy, prime-agent** — `EmitsBlockedActivity() = false` → suppressed.
- Only **claude-code** and **kimchi** open the boundary.

Coverage:
- `TestWiring_UrgentNudgeGateComesFromAdapters` (daemon) resolves the predicate against the real registry and asserts codex/droid/goose/devin + unknown + nil resolver stay fail-closed, while claude-code/kimchi are allowed.
- `TestGuard_NudgeUrgentReachesNeedsInputButNotBlocked` (sessionguard) adds the capability-safe / ambiguous / nil-predicate `waiting_input` rows.
- `TestPRObservation_MergeConflictReachesNeedsInputSession` (lifecycle) adds the ambiguous-harness suppression case end to end.

On the notification-channel alternative: I kept pane injection for the blocked-signalling harnesses (the original #2173 intent — the human parked at a genuine idle prompt is who needs to act), but every harness that can't prove the boundary now falls through to suppression rather than injecting.

whoisasx (approved):
Approved after follow-up verification. The two requested boundaries are addressed: urgent waiting_input delivery now fails closed unless the adapter proves permission dialogs are surfaced as blocked, and merge-conflict re-arm bookkeeping runs before exited/terminated delivery suppression. All review threads are resolved. Verified the PR head and a no-commit merge with current main; both the focused lifecycle/sessionguard/daemon suite and the merged-tree full Go suite pass, with focused race tests, go build, go vet, and git diff --check also clean.

Issue #3516, opened 2026-08-03
Expose available reviewer harness catalog via API


neversettle17-101:
Follow-up from PR #3384 review.\n\nAdd a daemon API surface that returns the available reviewer agent harness catalog so clients do not need to mirror reviewer harness IDs manually.\n\nContext: PR review thread requested an API for available reviewer agent harnesses.

ApexYash11:
IS THIS ISSUE UP FOR GRAB?

Rishet11:
@AgentWrapper , AO still has no API for reviewer availability, so clients keep a hardcoded reviewer list.
This should be kept open as P2

neversettle17-101:
@Shankywho

Rishet11:
@vanshx999 Please review #5213

vanshx999:
okay @Rishet11

PR #3767, opened 2026-08-09
Expose reviewer harness catalog via API

krishrathi1:
Adds `GET /api/v1/reviewer-harnesses`, returning the daemon's reviewer agent harness catalog so clients no longer need to hand-mirror reviewer harness ids.

**Response shape**

```json
{
  "reviewerHarnesses": ["claude-code", "codex", "copilot", "cursor", "kilocode", "opencode", "kiro", "pi", "qwen", "agy", "continue", "goose", "vibe", "devin", "droid", "kimi", "kimchi", "muse", "amp", "aider", "grok", "crush", "auggie", "cline", "autohand"]
}
```

The list is `domain.AllReviewerHarnesses` verbatim. Unlike the other catalog controllers (e.g. `/agents`), this one has no injected service dependency and no `NotImplemented` fallback: the data is static build-time data, not a probed or optional runtime service, so there is no build configuration where it would be unavailable.

**Changes**

- `backend/internal/httpd/controllers/reviewer_harnesses.go`: new `ReviewerHarnessesController` mounting `GET /reviewer-harnesses`.
- `backend/internal/httpd/controllers/dto.go`: new `ListReviewerHarnessesResponse` wire type.
- `backend/internal/httpd/api.go`: wires the controller into the `/api/v1` REST group alongside the other controllers.
- `backend/internal/httpd/apispec/specgen/build.go`: new `reviewerHarnesses` tag and operation entry, plus the `schemaNames` mapping for the new response type.
- `backend/internal/httpd/apispec/openapi.yaml` and `frontend/src/api/schema.ts`: regenerated via `npm run api` (`go generate ./internal/httpd/apispec/...` + `openapi-typescript`).
- `backend/internal/httpd/controllers/reviewer_harnesses_test.go`: new test asserting a 200 with the full harness id set (including `claude-code`, `codex`, `opencode`).

**Verification**

- Proved the test is meaningful with a patch-based fail/pass check (not `git stash`, per this repo's shared-worktree guidance): reverse-applied a patch of just the `api.go` route wiring, reran the test and confirmed it failed with `404 ROUTE_NOT_FOUND`, then reapplied the patch and confirmed it passed.
- `go build ./...` and `go vet ./...` from `backend`: clean.
- `go test ./internal/httpd/...` (controllers, apispec, specgen — includes the route/spec parity test and the openapi.yaml drift test): all pass.
- Confirmed `backend/internal/daemon` was untouched by this change; an unrelated pre-existing Windows-only test flake there (`TestStabilizeWorkingDirectoryChdirsToDataDir`, a `t.TempDir()`/`os.Chdir` cleanup-ordering issue) is unaffected by and unrelated to this PR.

Addresses #3516.

### DIFF
--- a/backend/internal/httpd/api.go
+++ b/backend/internal/httpd/api.go
@@ -87,23 +87,24 @@ func normalizeAPIDeps(deps APIDeps, log *slog.Logger) APIDeps {
 // API owns one controller per resource and is the single Register call the
 // router invokes to mount the /api/v1 surface.
 type API struct {
-	cfg           config.Config
-	deps          APIDeps
-	agents        *controllers.AgentsController
-	projects      *controllers.ProjectsController
-	sessions      *controllers.SessionsController
-	usage         *controllers.UsageController
-	prs           *controllers.PRsController
-	reviews       *controllers.ReviewsController
-	notifications *controllers.NotificationsController
-	push          *controllers.PushController
-	imports       *controllers.ImportController
-	shellTerms    *controllers.ShellTerminalsController
-	conversations *controllers.ConversationsController
-	settings      *controllers.SettingsController
-	dev           *controllers.DevController
-	browser       *controllers.BrowserController
-	events        *EventsController
+	cfg               config.Config
+	deps              APIDeps
+	agents            *controllers.AgentsController
+	projects          *controllers.ProjectsController
+	sessions          *controllers.SessionsController
+	usage             *controllers.UsageController
+	prs               *controllers.PRsController
+	reviews           *controllers.ReviewsController
+	notifications     *controllers.NotificationsController
+	push              *controllers.PushController
+	imports           *controllers.ImportController
+	shellTerms        *controllers.ShellTerminalsController
+	conversations     *controllers.ConversationsController
+	settings          *controllers.SettingsController
+	dev               *controllers.DevController
+	browser           *controllers.BrowserController
+	reviewerHarnesses *controllers.ReviewerHarnessesController
+	events            *EventsController
 }
 
 // NewAPI constructs the API surface from its dependencies. cfg carries the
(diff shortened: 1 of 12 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (changes requested):
Thanks for adding the API surface. I think this needs to change shape/source before merge:

The reviewer catalog should follow the same inventory contract as the agent harness catalog, not expose a standalone static string list from `domain.AllReviewerHarnesses`. Today `/api/v1/agents` returns the daemon inventory with `supported`, `installed`, and `authorized` `AgentInfo` entries, including labels/auth status. Reviewers are rendered with that same inventory data in the UI, so this endpoint should either reuse/filter the agent inventory for reviewer-capable harnesses or otherwise return the same inventory-shaped data.

As written, `GET /api/v1/reviewer-harnesses` only returns `{ reviewerHarnesses: string[] }`, which loses labels and install/auth readiness and gives clients a second catalog shape to reconcile. It also does not actually remove the frontend mirror yet, because `ReviewerSelect` still depends on the hardcoded `KNOWN_REVIEWER_HARNESS_IDS` list.

Please make this endpoint follow the agent inventory shape/source and update the frontend to consume it instead of maintaining a parallel reviewer-id list.

krishrathi1:
@neversettle17-101 done -- reshaped both sides:

**Backend:** `GET /api/v1/reviewer-harnesses` now returns the exact `agentsvc.Inventory` shape `/api/v1/agents` does (`supported`/`installed`/`authorized` `Info{id, label, authStatus}` entries), filtered to harnesses `domain.ReviewerHarness.IsKnown()` also recognizes. The controller takes the same `AgentCatalog` dependency `AgentsController` does and 501s the same way when it's absent, instead of being a standalone static-list handler with no injected dependency. `ListReviewerHarnessesResponse` is now a type alias for `agentsvc.Inventory` rather than its own struct, so the two endpoints literally share one OpenAPI schema (`ListAgentsResponse`) -- there's no drift to keep in sync because there's only one shape.

**Frontend:** `ReviewerSelect` no longer filters `supported`/`installed`/`authorized` against the hardcoded `KNOWN_REVIEWER_HARNESS_IDS` mirror at render time -- that list now only seeds the picker's placeholder before the query resolves. `ProjectSettingsForm` and `SessionInspector` fetch the new endpoint through a `useReviewerHarnessesQuery` hook (mirrors `useAgentsQuery`) and pass its already-filtered data into `ReviewerSelect`, instead of handing it the raw worker agent catalog and letting it self-filter.

Also rebased onto current `upstream/main` (branch was ~127 commits behind by the time I got back to this) to clear the `dirty` merge state -- real conflicts in `api.go` (new `deps APIDeps` field), `apispec/specgen/build.go` (new mobile-device schema overrides), and `SessionInspector.tsx` (the reviewer row picked up `contentAlign`/`autoReviewEnabled` from newer commits), all merged by hand keeping both sides' changes, `openapi.yaml`/`schema.ts` regenerated fresh afterward rather than hand-merged.

Verification: `go build`/`go vet` clean, `go test ./...` -- only failures are 4 pre-existing `session_manager` tests that fail identically on a pristine unmodified `upstream/main` checkout (Windows dev-machine artifacts: daemon binary naming/PATH-pinning assumptions), unrelated to this change. `tsc --noEmit` clean across the frontend. `vitest run` -- fixed 10 `ProjectSettingsForm` tests that broke because `ReviewerSelect` now calls a second endpoint the fixtures didn't mock (added matching `/api/v1/reviewer-harnesses` stubs alongside each existing `/api/v1/agents` one); the remaining 17 failures across the full suite are all pre-existing and outside this change's footprint -- macOS DMG/auto-updater/translocation tests running on Windows, a `landing/` subpackage missing an unrelated npm dependency, and a Windows-socket-permissions test, none touching agents/reviewers/inventory code.

New commits: `2bd6d5b` (the reshape) and `92669c5` (test fixture fix) on top of the rebase.

i-trytoohard:

Rishet11:
@krishrathi1

Issue #3745, opened 2026-08-08
Hardening: enforce read-only tool access and reliable cancel-stop for all reviewer harnesses


neversettle17-101:
## Summary

The reviewer harness surface has grown to include many interactive agent TUIs. Two safety properties are not consistently enforced today:

1. **Read-only tool access** — some reviewers launch as "host-trusted" without OS isolation or tool allowlists.
2. **Cancel stops active execution** — the "Cancel review" button must reliably interrupt the reviewer's current turn/operation. The TUI pane itself should remain alive for inspection.

This issue tracks making both properties hold for every reviewer harness.

### Read-only access is not enforced for all reviewers

- `backend/internal/adapters/reviewer/kimi/kimi.go:47-82` launches Kimi as an **experimental host-trusted reviewer** with only `--plan` mode. It sets no sandbox, no tool allow/deny list, and no permission policy:
  ```go
  []string{binary, "--plan", "--skills-dir", skillsDir}
  ```
  The adapter itself documents the gap at `kimi.go:20-23`:
  > "experimental host-trusted reviewer: Kimi has no OS isolation; terminal users can invoke shell mode, change Plan mode, open an external editor, or alter configuration"

- The frontend classifies Kimi in the host-trusted bucket (`frontend/src/renderer/components/ReviewerSelect.tsx:13`) and shows a warning, but the daemon does not harden the launch.

- By contrast, other reviewers already enforce read-only behavior:
  - `claudecode` uses `AllowedTools`/`DisallowedTools` lists (`backend/internal/adapters/reviewer/claudecode/claudecode.go:44-66`).
  - `codex` uses `--sandbox read-only` (`backend/internal/adapters/reviewer/codex/codex.go:36-63`).
  - `opencode` injects a deny-by-default permission policy (`backend/internal/adapters/reviewer/opencode/opencode.go:38-94`).

### Cancel behavior is inconsistent across reviewers

- The review cancel contract lives in `backend/internal/ports/reviewer.go:25-56` and supports three modes: `interrupt`, `message`, and `input`.
- The launcher dispatches these in `backend/internal/review/launcher.go:543-615`.
- The expectation is **interrupt active execution while keeping the TUI pane alive** — the pane is preserved for inspection.
- Kimi returns a single `Ctrl-C` interrupt (`backend/internal/adapters/reviewer/kimi/kimi.go:99-106`):
  ```go
[...]

PR #3768, opened 2026-08-09
Isolate Kimi reviewer profile/config/credentials from the host

krishrathi1:
## Scope

This is a narrow slice of #3745 ("Hardening: enforce read-only tool access and reliable cancel-stop for all reviewer harnesses"): only the **read-only access gap for the Kimi reviewer**. It does **not** touch the cancel-behavior half of that issue (interrupt reliability), which spans `ports/reviewer.go` and `review/launcher.go` across every reviewer and is left for a separate PR.

## Problem

`backend/internal/adapters/reviewer/kimi/kimi.go` launched Kimi's CLI with only `--plan` (plus `--skills-dir` once a task exists) and no other hardening:

- No sandbox flag, no tool allow/deny list, no permission-policy injection - unlike the three reviewers that already enforce read-only behavior: `claudecode` (`AllowedTools`/`DisallowedTools`), `codex` (`--sandbox read-only`), `opencode` (`OPENCODE_CONFIG_CONTENT` deny-by-default policy).
- `ReviewCommand` set no `Env` and no `WorkingDirectory` override, so the reviewer inherited the daemon's real process environment and ran with cwd at the worker checkout. `KIMI_CODE_HOME` was unset, so Kimi read and could write the user's real `~/.kimi-code` profile, config, credentials, and hooks - not just an AO-owned copy.

## Investigation: does Kimi's CLI support any of the three existing mechanisms?

No. The worker Kimi adapter (`backend/internal/adapters/agent/kimi/kimi.go`, `appendApprovalFlags`) shows Kimi's only relevant launch flags are approval-mode flags: `--auto` / `-y` (yolo). Both make the model *more* autonomous, not more restricted - there is no allowlist, no sandbox flag, no permission-policy hook equivalent to the other three adapters. `docs/adr/0002-secure-interactive-reviewer-gateway.md` independently confirms this: Kimi is explicitly listed (alongside Agy, Continue, Devin, Droid, Goose, Qwen, Vibe) as a CLI whose native modes cannot be made genuinely read-only by flags or prompt text, and states real containment requires an OS-level sandbox provider that is "required" but not yet built for any adapter.

So none of the three named mechanisms apply here, and I looked for an AO-side mitigation instead, per the issue's own suggested approach ("wrap the harness in a containment layer" / apply an AO-owned isolation root). One already exists and is already shipped: `backend/internal/reviewgateway.PrepareHostTrustedEnvironment`, used today by five sibling host-trusted reviewers - `agy`, `continueagent`, `goose`, `qwen`, `vibe` - to redirect each CLI's profile/config/state/cache/tmp directories to a private root under `AO_DATA_DIR` instead of the user's real CLI profile. Kimi was the only reviewer using this general "host-trusted CLI with no read-only mode" shape that had zero use of this existing mechanism - it was strictly worse off than its closest peers, not merely at the same tier.

## Fix

`ReviewCommand` now:

- Calls `reviewgateway.PrepareHostTrustedEnvironment(inv.DataDir, inv.ReviewerID)` to get an AO-owned profile root (mirrors `qwen.go`/`vibe.go`/`goose.go`/`agy.go`/`continueagent.go` exactly).
- Points `KIMI_CODE_HOME` (Kimi's own documented env var, already used the same way by the *worker* Kimi adapter's `AugmentRuntimeEnv`) plus the generic `HOME`/`XDG_*`/`TMPDIR` variables at that root, so a shell escape or config change inside the reviewer pane cannot read or overwrite the user's real `~/.kimi-code`.
- Seeds `config.toml` and `credentials/kimi-code.json` from the user's real profile into the isolated root (mirrors `seedHostVibeProfile`/`seedHostQwenSettings`/`seedHostContinueConfig`) so authentication keeps working.
- Forwards `KIMI_API_KEY` / `KIMI_CODE_API_KEY` / `MOONSHOT_API_KEY` from the daemon's environment if set (mirrors `continueagent`'s `CONTINUE_API_KEY` forwarding), since env-var auth is Kimi's other supported auth path.
- Leaves `WorkingDirectory` as the worker checkout (unchanged) - Kimi genuinely needs to read the PR locally to review it, matching `goose`/`vibe`/`agy`/`continueagent`, which isolate the *profile* but not the working directory either.

## What this does **not** fix (honest residual gap)

The review checkout itself is not read-only. Kimi's CLI has no allow/deny tool list and no read-only sandbox mode, so shell mode, file edits, and config changes *inside the checkout* remain possible from the terminal - identical to the residual risk already accepted for `agy`/`continueagent`/`goose`/`qwen`/`vibe`. Fully closing that requires either upstream Kimi CLI support for a tool allowlist/sandbox flag, or the OS-level reviewer sandbox described in `docs/adr/0002-secure-interactive-reviewer-gateway.md` (tmux/ConPTY + AppContainer/restricted-token containment), neither of which exists today for any reviewer. The package doc, `HostTrustWarning`, and `ReviewCommand` comments in `kimi.go` now state this precisely instead of only describing the old, unmitigated gap.

## Testing

- `TestReviewCommandIsolatesKimiProfileFromHostConfiguration` - asserts `HOME`/`KIMI_CODE_HOME` point at the AO-owned `reviewer-runtime/<reviewerId>/config` root and `WorkingDirectory` stays the worker checkout.
- `TestReviewCommandSeedsPrivateProfileFromHostKimiConfig` - asserts `config.toml` and `credentials/kimi-code.json` are copied byte-for-byte into the isolated root.
- `TestReviewCommandForwardsKimiAPIKeyEnvVars` - asserts `MOONSHOT_API_KEY` reaches the isolated launch env.
- Updated `TestReviewCommandUsesPlanModeAndEmptySkillsDirectory` to supply `DataDir`/`ReviewerID` (now required inputs) while keeping its existing argv/message/skills-dir assertions.

**Verification performed:**
- Fail-then-pass: saved the `kimi.go` diff as a patch, reverse-applied it (keeping the test changes), and reran the new tests - they fail to even compile (`undefined: kimiCodeHomeEnv`) against the unpatched adapter, confirming the tests are load-bearing on the fix. Re-applying the patch restores a clean build and passing tests. No `git stash` was used at any point (this repo's other adapters use identical `reviewgateway`/profile-seeding/env-forwarding patterns, and stash refs are shared across concurrent worktrees, so a patch file was used instead).
- `go test ./internal/adapters/reviewer/...` was run from `backend/`. Note for reviewers testing on native Windows: several *pre-existing, untouched* reviewer test files (`agy`, `continueagent`, `cursor`, `devin`, `droid`, `goose`, `qwen`, `vibe`, and now `kimi`) hardcode a Unix-style absolute path (e.g. `/opt/kimi`) for their fake resolved binary, which `filepath.IsAbs` correctly does not treat as absolute on Windows without a drive letter, and separately assert `os.FileMode` bits that NTFS does not preserve the same way as POSIX permissions. Both are pre-existing, cross-package Windows-only artifacts confirmed by running the identical assertions against the unmodified upstream `kimi.go` and against untouched sibling packages - not regressions from this change. I confirmed the new isolation logic itself is correct with a temporary, uncommitted local copy of the new tests using a Windows-safe absolute path, which passed cleanly; that scratch copy was deleted before committing and is not part of this PR.

Addresses part of #3745 (Kimi reviewer read-only enforcement; cancel-behavior consistency is left for follow-up).

### DIFF
--- a/backend/internal/adapters/reviewer/kimi/kimi.go
+++ b/backend/internal/adapters/reviewer/kimi/kimi.go
@@ -1,5 +1,18 @@
 // Package kimi adapts Kimi Code CLI as an experimental host-trusted AO
 // reviewer. It always runs Kimi's visible, long-lived interactive TUI.
+//
+// Kimi's CLI has no read-only sandbox flag, no tool allow/deny list, and no
+// permission-policy injection point (unlike claude-code's AllowedTools, codex's
+// --sandbox read-only, or opencode's OPENCODE_CONFIG_CONTENT). The only launch
+// flags it exposes are approval-mode flags (--auto, -y) that make the model
+// more autonomous, not more restricted, so none of those three mechanisms
+// apply here. AO instead applies the same AO-owned isolation the
+// reviewgateway package already gives Kimi's fellow host-trusted reviewers
+// (agy, continueagent, goose, qwen, vibe): Kimi's profile, config, credentials,
+// and hooks are redirected to a private root under AO_DATA_DIR instead of the
+// user's real ~/.kimi-code. This keeps a shell escape or config change inside
+// the reviewer pane from reading or overwriting the user's real Kimi profile.
+// It does not make the checkout itself read-only; see HostTrustWarning.
 package kimi
 
 import (
@@ -13,14 +26,35 @@ import (
 	workerkimi "github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/kimi"
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+	"github.com/aoagents/agent-orchestrator/backend/internal/reviewgateway"
 )
 
 const skillsDirectoryName = "kimi-reviewer-skills"
 
+// kimiCodeHomeEnv is Kimi CLI's documented environment variable for pointing
+// the CLI at an alternate profile/config/credentials/hooks home. Mirrors the
+// worker Kimi adapter's private kimiCodeHomeEnv constant
+// (backend/internal/adapters/agent/kimi/kimi.go), which uses the same
+// variable to keep worker sessions off the user's real Kimi profile.
+const kimiCodeHomeEnv = "KIMI_CODE_HOME"
+
+// kimiAPIKeyEnvVars are the environment variables Kimi CLI reads for API-key
+// auth (mirrors backend/internal/adapters/agent/kimi/auth.go's
+// kimiAPIKeyEnvVars). The reviewer launch replaces the process environment
+// with an AO-owned one, so any of these present in the daemon's own
+// environment must be forwarded explicitly or env-var auth would silently stop
+// working once the profile is isolated.
+var kimiAPIKeyEnvVars = []string{"KIMI_API_KEY", "KIMI_CODE_API_KEY", "MOONSHOT_API_KEY"}
+
 // HostTrustWarning describes the authority retained by Kimi's interactive
 // terminal. Plan mode limits Kimi to read-only tools, while auto mode handles
-// approvals and questions without pausing for user input. AO deliberately
-// does not replace the TUI with print/headless mode or place it in a sandbox.
+// approvals and questions without pausing for user input, and AO isolates
+// Kimi's own profile/config/credentials/hooks to an AO-owned root (see the
+// package doc) — but none of that is OS isolation or a tool allow/deny list: a
+// terminal user can still invoke shell mode, change Plan mode, open an
+// external editor, or alter configuration, and read/write the review checkout
+// itself. AO deliberately does not replace the TUI with print/headless mode or
+// place it in a sandbox.
 const HostTrustWarning = "experimental host-trusted reviewer: Kimi has no OS isolation; terminal users can invoke shell mode, change Plan mode, open an external editor, or alter configuration"
 
 // Reviewer builds Kimi's persistent interactive reviewer command.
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101:
This has been fixed already

krishrathi1:
@neversettle17-101 checked what #3782 actually landed against this PR's diff, and I think they're addressing two different gaps under the same issue (#3745) rather than one being redundant with the other:

- **#3782 (merged, yours):** `--auto` paired with `--plan` so the reviewer runs unattended, and `ReviewCancel` now sends two Ctrl-C so an active operation is actually interrupted. That's the *cancel-stop reliability* half of #3745, and the tool-permission side of *read-only* — controls what Kimi is willing to do.
- **This PR:** routes Kimi's profile/config/credentials/hooks through `reviewgateway.PrepareHostTrustedEnvironment` into an AO-owned root instead of the user's real `~/.kimi-code`, matching the isolation already applied to the other host-trusted reviewers (agy, continueagent, goose, qwen, vibe). That's the *filesystem blast-radius* side — controls where Kimi is allowed to read/write for its own state, independent of what tools it invokes.

Neither one gives Kimi a real sandbox (its CLI still has no read-only flag or tool allowlist — `HostTrustWarning` says so explicitly, unchanged by either PR), but they're not overlapping: a shell escape inside the reviewer pane could still read/overwrite the user's actual Kimi credentials before this PR, and still can't after #3782 alone.

Rebased onto your merged commit just now (`5f9b85800`) — real conflict this time since we both touched `ReviewCommand`'s return struct (your `--auto` in `Argv` vs my `Env`/`WorkingDirectory` fields), merged so both are kept, confirmed `go build`/`go vet` clean and `go test ./internal/adapters/reviewer/kimi/...` shows only the same 2 failures that reproduce identically on unmodified `upstream/main` (Windows `filepath.IsAbs("/opt/kimi")` treating the POSIX test fixture as non-absolute — pre-existing, unrelated to either PR).

Happy to be wrong about this if there's a merged PR that specifically added the credential/profile isolation — pointer appreciated if so.

i-trytoohard:

Rishet11:
@krishrathi1

Issue #1211, opened 2026-04-13
Orchestrator button on worker page links to wrong parent session


i-trytoohard:
## Bug

On any worker session page (e.g. `/sessions/ao-8`), clicking the "Orchestrator" button in the top-right navigates to the wrong orchestrator (e.g. `ao-orchestrator-1`) instead of the correct parent (`ao-orchestrator-9`).

## Root Cause

Two-part failure:

1. **No parent tracking.** Sessions have no `parentOrchestratorId` field — neither in the `Session` type nor in metadata. When orchestrator-9 spawns worker ao-8, that relationship is never recorded.

2. **Fallback picks wrong orchestrator.** When the frontend fetches `/api/sessions?project=agent-orchestrator&orchestratorOnly=true`:
   - The API returns all orchestrator sessions for the project sorted lexicographically by ID
   - `orchestratorId` is set to `null` when there are multiple orchestrators (`sessions/route.ts:36`)
   - The client fallback (`page.tsx:162`) does `.find()` which returns the first match — `ao-orchestrator-1` (alphabetically first), not the active `ao-orchestrator-9`

## Fix

Apply both approaches:

### A — Store parent orchestrator in session metadata
- When an orchestrator spawns a worker, write `parentOrchestratorId` to the worker metadata (in session-manager spawn flow)
- Frontend reads `session.metadata["parentOrchestratorId"]` directly — no API fallback needed for the correct parent

### B — Smart fallback when parent metadata is missing
- Sort orchestrators by recency (prefer non-exited, then by `createdAt` desc) instead of lexicographic
- This fixes the common case for sessions created before the metadata fix

### Affected files:
- `packages/core/src/session-manager.ts` — write `parentOrchestratorId` metadata on worker spawn
- `packages/web/src/lib/serialize.ts:95` — sort by recency, not lexicographic
- `packages/web/src/app/sessions/[id]/page.tsx:160-163` — read parent from metadata, fallback to smart sort
- `packages/web/src/app/api/sessions/route.ts:36` — return per-project orchestrator ID properly

AgentWrapper:
this has been fixed since

PR #3769, opened 2026-08-09
fix: attribute worker sessions to the orchestrator that spawned them

krishrathi1:
## Summary

Issue #1211 tracked two problems with the "Orchestrator" button on a worker session's page:

1. **Lexicographic-first fallback picking the wrong orchestrator** — already fixed on `main` (see the issue's own triage comment from 2026-07-28). `findProjectOrchestrator()` in `frontend/src/renderer/types/workspace.ts` sorts by recency, not lexicographically. **Not touched by this PR.**
2. **No parent tracking** — there was no `parentOrchestratorId` anywhere, so the button always fell back to "the newest active orchestrator in the project." That's usually right (the common case has exactly one orchestrator) but wrong whenever a project has 2+ concurrent orchestrators, since the newest-active one isn't necessarily the one that actually spawned a given worker. **This PR fixes that remaining gap.**

## What changed

- `backend/internal/domain/session.go` — `SessionRecord` gains `ParentOrchestratorID` (empty string means unset, matching the existing convention used for `IssueID` and other optional session identifiers).
- `backend/internal/ports/session.go` — `SpawnConfig` gains the same field, threaded from the HTTP request through the manager's `seedRecord`.
- `backend/internal/service/session/service.go` — `resolveParentOrchestratorID` verifies the caller-supplied id actually names a real, currently known **orchestrator** session before it is persisted. A worker's own id, an unknown id, or an absent one all resolve to empty, so a directly-spawned worker never gets a parent attributed to it.
- `backend/internal/cli/spawn.go` — `ao spawn` now sends the caller's own `AO_SESSION_ID` as `parentOrchestratorId`. This only names an orchestrator when `ao spawn` runs from inside an orchestrator's own agent process (its `AO_SESSION_ID` is its own session id); a spawn from a plain terminal has no `AO_SESSION_ID` at all, so nothing changes for direct spawns.
- `backend/internal/storage/sqlite/migrations/0085_add_session_parent_orchestrator_id.sql` — adds `sessions.parent_orchestrator_id` (`TEXT NOT NULL DEFAULT ''`), plus the matching `queries/sessions.sql` and regenerated `gen/` code (`npm run sqlc`). The column is immutable after spawn (excluded from `UpdateSession`'s SET clause), mirroring how `session_mode` is already treated.
- `frontend/src/renderer/types/workspace.ts` — new `resolveSessionOrchestrator()` prefers a session's `parentOrchestratorId` (resolved against the project's session list) and falls back to the existing `findProjectOrchestrator()` recency heuristic only when no parent is recorded, or the recorded parent can no longer be found. `findProjectOrchestrator()` itself is untouched.
- `frontend/src/renderer/components/ShellTopbar.tsx` — the "Orchestrator" button now resolves through `resolveSessionOrchestrator()`.
- `backend/internal/httpd/apispec/openapi.yaml` and `frontend/src/api/schema.ts` — regenerated via `npm run api` per the repo's API contract rule, committed together with the Go changes.

## Verification

- Backend: `go build ./...` and `go vet ./...` clean. New tests:
  - `TestSessionPersistsParentOrchestratorID` (`storage/sqlite/store`, real SQLite): a worker spawned with a parent id persists and reads it back; a directly-spawned session reads back empty; an unrelated `UpdateSession` call doesn't clear it.
  - `TestSpawnParentOrchestratorIDAttribution` (`service/session`): the service only forwards the candidate id to the write path when it resolves to a real orchestrator session in the store; it drops a non-orchestrator id, an unknown id, and an absent one.
  - Confirmed both **fail** against the pre-fix code (build failure for the missing field / empty read-back for the missing column) and **pass** with the fix restored, using a saved-and-reversed patch file (not `git stash`).
  - Full relevant suites pass: `domain`, `ports`, `service/session`, `httpd`, `httpd/controllers`, `httpd/apispec` (spec-drift test), `httpd/apispec/specgen`, `storage/sqlite` (migration ledger test), `storage/sqlite/store`, `cli`. Three pre-existing `session_manager` failures (Windows path-separator assertions) and one flaky `service/agent` catalog-staleness test reproduce identically on unmodified `main` and are unrelated to this change.
- Frontend: `npx tsc --noEmit` clean. New tests in `workspace.test.ts` (`resolveSessionOrchestrator`): prefers the recorded parent over the newest orchestrator, falls back to recency when `parentOrchestratorId` is absent, falls back when the recorded parent can't be found, and matches `findProjectOrchestrator` when no session is given — this is the actual regression the issue is about (2+ concurrent orchestrators, the worker's button must point at its real parent, not just "the newest one"). Confirmed fail/pass the same way as the backend. Full frontend suite passes except for pre-existing, unrelated failures (macOS packaging/signing tests, a landing-page script with an undeclared `node-html-markdown` dependency, and Electron main-process socket-reconnect timing) — none touch session/workspace/orchestrator code.

Addresses the remaining work in #1211.

### DIFF
--- a/backend/internal/cli/spawn.go
+++ b/backend/internal/cli/spawn.go
@@ -45,6 +45,12 @@ type spawnRequest struct {
 	Branch      string `json:"branch,omitempty"`
 	Prompt      string `json:"prompt,omitempty"`
 	DisplayName string `json:"displayName"`
+	// ParentOrchestratorID carries the caller's own AO_SESSION_ID so the daemon
+	// can attribute this spawn to its orchestrator when `ao spawn` runs from
+	// inside an orchestrator's own agent process. The daemon verifies the id
+	// actually names a live orchestrator session before recording it; it is
+	// empty (and ignored) for a plain user/CLI spawn run outside any session.
+	ParentOrchestratorID string `json:"parentOrchestratorId,omitempty"`
 }
 
 type spawnResult struct {
@@ -126,14 +132,15 @@ func newSpawnCommand(ctx *commandContext) *cobra.Command {
 				}
 			}
 			req := spawnRequest{
-				ProjectID:   opts.project,
-				IssueID:     opts.issue,
-				Kind:        opts.kind,
-				Harness:     opts.harness,
-				Mode:        opts.mode,
-				Branch:      opts.branch,
-				Prompt:      opts.prompt,
-				DisplayName: name,
+				ProjectID:            opts.project,
+				IssueID:              opts.issue,
+				Kind:                 opts.kind,
+				Harness:              opts.harness,
+				Mode:                 opts.mode,
+				Branch:               opts.branch,
+				Prompt:               opts.prompt,
+				DisplayName:          name,
+				ParentOrchestratorID: strings.TrimSpace(os.Getenv("AO_SESSION_ID")),
 			}
 			var res spawnResult
 			if err := ctx.postJSON(cmd.Context(), "sessions", req, &res); err != nil {
--- a/backend/internal/domain/session.go
+++ b/backend/internal/domain/session.go
@@ -66,11 +66,19 @@ type SessionMetadata struct {
 // facts: identity, agent harness, activity_state, is_terminated, and operational
 // metadata. The user-facing Status is derived from these facts plus PR facts.
 type SessionRecord struct {
-	ID        SessionID    `json:"id"`
-	ProjectID ProjectID    `json:"projectId"`
-	IssueID   IssueID      `json:"issueId,omitempty"`
-	Kind      SessionKind  `json:"kind"`
-	Harness   AgentHarness `json:"harness,omitempty"`
+	ID        SessionID `json:"id"`
+	ProjectID ProjectID `json:"projectId"`
+	IssueID   IssueID   `json:"issueId,omitempty"`
+	// ParentOrchestratorID is the orchestrator session that spawned this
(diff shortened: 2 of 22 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
From the NotesBot call on `2026-08-23`, this existing pr appears to be relevant.
Match confidence: `0.61` (fuzzy).

Relevant discussion excerpts:
- - prateek said Kanban is not meant to be manually driven that way; the intended model is that users direct agents and the system reflects actual session state. - Arjun asked about spawning **10 workers** for **10 tasks** and whether AO supports that flow. - prateek asked whether Arjun had used the **Orchestrator [...]
- - AmongStar asked what feature-flagging approach to use. - prateek said Kanban is not meant to be manually driven that way; the intended model is that users direct agents and the system reflects actual session state. - Arjun asked about spawning **10 workers** for **10 tasks** and whether AO supports that flow.
- - prateek said Kanban is not meant to be manually driven that way; the intended model is that users direct agents and the system reflects actual session state. - Arjun asked about spawning **10 workers** for **10 tasks** and whether AO supports that flow.

Source thread: https://discord.com/channels/1476302178913357958/1541141481157566515/1541141481157566515

Rishet11:
@krishrathi1 
#3324 landed the parent-link fix before this PR was opened. findProjectOrchestrator and the orchestratorID field are on main, and issue #1211 was closed completed as fixed. Closing as superseded.

PR #3770, opened 2026-08-09
feat: make agent switching responsive and phase-aware

whoisasx:
## Summary

This is a stacked follow-up to #3548 and targets codex/agent-switching.

- return 202 after durable switch admission, then run the existing switch saga in a daemon-owned tracked worker
- expose the active durable switch on session reads and keep the CLI synchronous by polling existing switch history
- present one consistent five-stage switching experience across the dialog, terminal, sidebar, board, and top bar
- preserve source permission interaction, terminal input locking, navigation/remount continuity, target settlement, and static recovery states
- keep the implementation limited to Claude Code and Codex; no new switch state, job table, migration, cancellation, retry, or provider expansion

## User experience

- the dialog closes as soon as AO durably accepts the request
- the worker terminal shows meaningful progress without blocking navigation or auxiliary shells
- session surfaces show Switching to … instead of briefly reporting the source as Exited
- permission-required, completed-but-settling, failed, and recovery-required states have distinct behavior and accessible copy
- all eight locale files include the new formal status copy

## Verification

- npm run lint (all backend tests and golangci-lint; passed outside the filesystem sandbox because httptest requires loopback sockets)
- npm run frontend:typecheck
- focused renderer suites: SessionsBoard, TerminalSwitchAgentButton, and CenterPane (103 tests)
- go test -race ./internal/session_manager
- focused backend session-manager, storage, service, HTTP, CLI, and daemon packages
- npm run sqlc and npm run api; generated artifacts committed

## Scope note

The optional fake-renderer E2E seam was deliberately excluded to keep this stacked PR focused on the shipping UI and durable API contract.


whoisasx:
Screenshot 2026-08-11 at 11 11 55
Screenshot 2026-08-11 at 11 11 42
Screenshot 2026-08-11 at 11 11 29
Screenshot 2026-08-11 at 11 10 56

backend/internal/httpd/controllers/dto.go
```diff
@@ -211,7 +212,7 @@ type SpawnSessionResponse struct {
 // SwitchAgentRequest is the body of POST /api/v1/sessions/{sessionId}/switch-agent.
 type SwitchAgentRequest struct {
 	TargetHarness  domain.AgentHarness `json:"targetHarness" enum:"claude-code,codex" description:"Agent harness to continue the logical AO session with."`
-	Note           string              `json:"note,omitempty" maxLength:"4096" description:"Optional user guidance included in the bounded handoff context."`
+	Model          string              `json:"model,omitempty" maxLength:"256" description:"Optional model override for the target agent launch or resume."`
```
ronishrohan, line 217: P1: v0.12.3 shipped the optional note field and note-bearing request fingerprints. The new request shape silently ignores note, and retries using an existing non-empty-note idempotency key no longer match. Please preserve note propagation and legacy fingerprint matching on v1, or move the replacement contract behind an explicit version boundary.

backend/internal/session_manager/agent_switching.go
```diff
@@ -2700,6 +2995,26 @@ func (m *Manager) reconcileStartingTarget(ctx context.Context, store ports.Agent
...
+	if projectErr == nil {
+		if rollbackErr := m.rollbackStoppedAgentSwitchSource(ctx, store, sw, project); rollbackErr != nil {
+			m.logger.Error("agent switch recovery: automatic source rollback failed", "sessionID", rec.ID, "switchID", sw.ID, "error", rollbackErr)
+		} else {
+			m.logger.Info("agent switch recovery: source restored", "sessionID", rec.ID, "switchID", sw.ID, "harness", sw.FromHarness)
+		}
+	} else {
+		m.logger.Error("agent switch recovery: source project unavailable for rollback", "sessionID", rec.ID, "switchID", sw.ID, "error", projectErr)
+	}
+	_, failErr := m.failAgentSwitch(ctx, store, sw, domain.AgentSwitchErrorDaemonRestartPostStop)
```
ronishrohan, line 3188: P1: This terminalizes the recovered saga even when loadProject or rollbackStoppedAgentSwitchSource failed. The caller then treats recovery as resolved and releases the input fence without proving that either provider owns the session. Return unresolved with the rollback error and keep the saga fenced until restoration or another owner is proven.

ronishrohan (commented):
Review verdict: Not ready.\n\nFix order:\n1. Restore durable ownership safety: fix failed recovery rollback, shutdown cancellation after source stop, and workerless idempotent replay.\n2. Decide the released v1 compatibility policy, then restore note and fingerprint compatibility.\n3. Make renderer admission remain locked until durable invalidations settle; this item is validation-degraded and is not posted as an inline blocker.\n4. Apply the Lucide, shared-dialog, and no-text-animation standards fixes.\n\nFocused Go tests passed for session manager, session service, HTTP controllers, and CLI against the reviewed head. The deslop pass found no additional grounded any/type-ignore escapes, catch-and-ignore blocks, comment clutter, or broad refactor worth requesting.

backend/internal/session_manager/agent_switching.go
```diff
@@ -248,27 +260,114 @@ func (m *Manager) SwitchAgent(ctx context.Context, id domain.SessionID, cfg Swit
...
+			cancel()
+			if cleanupErr != nil {
+				rollbackSafe = false
+				retErr = errors.Join(retErr, fmt.Errorf("clean target workspace before source rollback: %w", cleanupErr))
+			} else {
+				targetWorkspacePrepared = false
+			}
+		}
+		if rollbackSafe {
+			rollbackCtx, cancel := context.WithTimeout(workerCtx, switchPostStopWait)
```
ronishrohan: P1: daemon shutdown cancels the background worker context before waiting for switch workers. context.WithTimeout on that already-canceled parent makes post-stop source rollback fail immediately; the detached settlement path can then mark the ownerless saga terminal. Use a bounded context detached from shutdown cancellation, and retain the nonterminal fence if rollback is not proven.

backend/internal/session_manager/agent_switching.go
```diff
@@ -76,132 +92,128 @@ func (m *Manager) switchStore() (ports.AgentSwitchStore, error) {
...
-			return domain.AgentSwitch{}, fmt.Errorf("switch agent %s: idempotency lookup: %w", id, err)
+			return domain.AgentSwitch{}, nil, fmt.Errorf("switch agent %s: idempotency lookup: %w", id, err)
 		} else if ok {
-			if existing.RequestFingerprint != requestFingerprint {
-				return existing, fmt.Errorf("switch agent %s: %w", id, domain.ErrAgentSwitchIdempotencyConflict)
+			if !existing.RequestFingerprint.MatchesRequest(id, cfg.TargetHarness, cfg.Model) {
+				return existing, nil, fmt.Errorf("switch agent %s: %w", id, domain.ErrAgentSwitchIdempotencyConflict)
 			}
-			return existing, nil
+			return existing, nil, nil
```
ronishrohan, line 196: P1: a canceled request can leave a committed saga without a worker. This same-key replay returns the durable active row before retained-saga recovery or worker scheduling. Resolve commit ambiguity on a short durable context and route an active retained replay through atomic recovery or worker ownership.

backend/internal/httpd/controllers/sessions.go
```diff
@@ -1068,14 +1062,14 @@ func (c *SessionsController) switchAgent(w http.ResponseWriter, r *http.Request)
...
-		Note:           note,
+		Model:          model,
 		IdempotencyKey: idempotencyKey,
 	})
 	if err != nil {
 		envelope.WriteError(w, r, err)
 		return
 	}
-	envelope.WriteJSON(w, http.StatusOK, AgentSwitchResponse{Switch: agentSwitchView(switchRecord)})
+	envelope.WriteJSON(w, http.StatusAccepted, AgentSwitchResponse{Switch: agentSwitchView(switchRecord)})
```
ronishrohan, line 1134: P1 contract decision: v0.12.3 shipped this v1 operation as synchronous with HTTP 200 after target acknowledgement. Returning 202 immediately after admission on the same route changes released semantics. Please preserve the synchronous v1 behavior or introduce a versioned or explicitly negotiated asynchronous contract.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -403,125 +551,251 @@ export function CenterPane({
...
-								<span className="agent-switch-transfer-pulse absolute inset-y-0 w-10 rounded-full bg-gradient-to-r from-transparent via-accent to-transparent" />
-							</div>
-							<ArrowRight className="size-icon-lg shrink-0" />
+						<SwitchingAgentMark harness={agentSwitch.fromHarness} />
+						<div
+							aria-hidden="true"
+							className="relative h-4 w-20 shrink-0 text-accent sm:w-28"
+							data-testid="agent-switch-transfer-arrow"
+						>
+							<svg
```
ronishrohan: P2: DESIGN.md requires Lucide icons only. This replaces the previous Lucide ArrowRight with a hand-authored SVG. Keep any shaft animation in a decorative wrapper, but use the Lucide icon for the arrow shape.

frontend/src/renderer/components/SwitchAgentDialog.tsx
```diff
@@ -1,37 +1,23 @@
-import { ArrowLeftRight, FileWarning, LoaderCircle, TriangleAlert, X } from "lucide-react";
+import * as Dialog from "@radix-ui/react-dialog";
```
ronishrohan: P2: CLAUDE.md requires shadcn primitives where one fits, and this tree already has components/ui/dialog.tsx. Please use or extend that shared dialog primitive while preserving the terminal-scoped portal behavior.

frontend/src/renderer/styles.css
```diff
@@ -1370,6 +1376,52 @@ textarea,
...
+	left: calc(50% + 0.25rem);
+	width: calc(100% - 0.5rem);
+	height: 1px;
+	background: var(--color-border-strong);
+	content: "";
+	transition: background-color 320ms var(--ease-out);
+}
+
+.agent-switch-progress-step {
+	transition: color 280ms var(--ease-out);
```
ronishrohan: P2: DESIGN.md says never animate text. This transition is applied to the progress-step element that contains the visible label, so phase changes animate label color. Keep motion on the dot or connector and change label color immediately.

backend/internal/httpd/controllers/sessions.go
```diff
@@ -1056,9 +1050,9 @@ func (c *SessionsController) switchAgent(w http.ResponseWriter, r *http.Request)
 		envelope.WriteAPIError(w, r, http.StatusBadRequest, "bad_request", "TARGET_HARNESS_REQUIRED", "targetHarness is required", nil)
 		return
 	}
-	note := domain.SanitizeControlChars(strings.TrimSpace(in.Note))
-	if len(note) > maxSwitchNoteLen {
-		envelope.WriteAPIError(w, r, http.StatusBadRequest, "bad_request", "SWITCH_NOTE_TOO_LONG", "note is too long", nil)
+	model := strings.TrimSpace(in.Model)
```
Annieeeee11: **[P3] Sanitize switch `model` like other model inputs**

Delegate/spawn paths run `domain.SanitizeControlChars` on `Model`. Switch-agent only trims. The retired `note` field was sanitized; the replacement `model` should get the same treatment before it reaches launch/restore argv.

```go
model := domain.SanitizeControlChars(strings.TrimSpace(in.Model))
```

backend/internal/session_manager/agent_switching.go
```diff
@@ -42,16 +42,32 @@ Use the deterministic facts embedded in the hidden continuation. It may also con
 	aoTargetActivationPrompt = `AO transferred the previous agent's context in hidden system instructions. Continue a clear, safe, already-authorized unfinished action; otherwise, acknowledge the current objective and wait for the user.`
 )
 
-var errSourceHandoffOwnershipChanged = errors.New("source session ownership changed during handoff collection")
+var (
+	errSourceHandoffOwnershipChanged = errors.New("source session ownership changed during handoff collection")
+	errAgentSwitchWorkersClosed      = errors.New("agent switch worker registration is closed")
```
Annieeeee11: **[P3] Shutdown refusal surfaces as an unmapped 500**

`errAgentSwitchWorkersClosed` is returned when admission races daemon shutdown, but `toAPIError` has no case for it, so clients get a generic 500. Especially awkward when `startAgentSwitchWorker` created-then-failed the durable row.

Please map this to a conflict/unavailable envelope in `service/session.toAPIError`, similar to how other exclusive-op conflicts are classified.

Annieeeee11 (changes requested):
## Review summary

Async admission + phase-aware UI looks solid overall (durable 202 admit, daemon-owned worker, `activeAgentSwitch` projection, CLI poll-to-terminal, settlement/rollback). CI is green.

### [P2] Hide Resume while a durable switch is active
`frontend/src/renderer/components/SessionInspector.tsx` (unchanged in this PR, but newly reachable because of it)

After admission returns 202, the switch saga keeps running in the background. Once the source is stopped, `activity.state` is `exited` for a long user-visible window. `ResumeAgentControl` only gates on exited and ignores `session.activeAgentSwitch`, so **Resume agent** appears mid-switch.

Clicking it fails because the switch still owns the exclusive gate, but `beginAgentResume` maps that conflict to `ErrResumeInProgress` → `"The agent is already being resumed"` (unlike `Kill`, which maps the same gate to `ErrSwitchInProgress`).

Before this PR the HTTP call blocked for minutes, so that exited window stayed behind the dialog. Async admission makes the false Resume affordance easy to hit.

Please gate Resume on `session.activeAgentSwitch` (and prefer a switch-specific API error when the gate is held by a switch).

Two smaller P3 hygiene items are inline.

ronishrohan:
the ui for this feature can be worked on later, after this is merged, let one of the design people know when you get this merged, we'll start working on the ui

frontend/src/renderer/components/SwitchAgentDialog.tsx
whoisasx: @ronishrohan Fixed in 2321b0879. SwitchAgentDialog now uses the shared components/ui/dialog primitives. DialogContent was minimally extended with portalContainer and overlay slots so the selector keeps its terminal-scoped portal and scrim without bypassing the shared component.

backend/internal/httpd/controllers/sessions.go
whoisasx: @ronishrohan Product decision: this operation remains asynchronous and returns 202 after durable admission. The durable switch record is the observation contract, the UI follows it through the active projection/history endpoint, and the CLI polls it to a terminal state. We are intentionally not retaining the prior synchronous 200 behavior on this route.

backend/internal/session_manager/agent_switching.go
whoisasx: @ronishrohan Fixed in 2321b0879. Ambiguous create-commit recovery now reads on a short durable context, so request cancellation cannot orphan a committed saga. If a transient read still leaves the saga retained, a same-key replay routes through retained-saga reconciliation instead of merely returning the active row. The cancellation and replay paths each have a regression test.

frontend/src/renderer/components/CenterPane.tsx
whoisasx: @ronishrohan Fixed in 2321b0879. The arrow shape is Lucide ArrowRight again. The moving light remains in a separate decorative shaft wrapper, so no custom SVG or icon animation is required. CenterPane coverage now asserts the Lucide icon and isolated pulse.

backend/internal/session_manager/agent_switching.go
whoisasx: @ronishrohan Fixed in 2321b0879. Recovery now returns unresolved when the project cannot be loaded or the source cannot be relaunched. The switch remains nonterminal and the input fence stays closed; AO only writes daemon_restart_post_stop and releases the gate after source restoration succeeds. Both failure paths are covered in TestReconcileAgentSwitchesUsesDurableBoundaries.

backend/internal/httpd/controllers/dto.go
whoisasx: @ronishrohan Product decision: note is retired rather than preserved. The switch endpoint now uses strict decoding, so note and any other retired or unknown field returns INVALID_JSON instead of being silently ignored. Note-bearing v0.12.3 idempotency retries are intentionally not supported across this contract change. Covered by TestSessionsAPI_AgentSwitchValidationAndErrors.

frontend/src/renderer/styles.css
whoisasx: @ronishrohan Fixed in 2321b0879. The progress-step color transition was removed, so visible labels change immediately. Motion remains limited to the connector, dot, and decorative transfer pulse, with reduced-motion handling preserved.

backend/internal/httpd/controllers/sessions.go
whoisasx: @Annieeeee11 Fixed in 2321b0879. The switch model is sanitized with domain.SanitizeControlChars before trimming, length validation, and service delivery, matching the other model inputs. The controller lifecycle test includes a control character and verifies the sanitized model.

backend/internal/session_manager/agent_switching.go
whoisasx: @ronishrohan Fixed in 2321b0879. Post-stop compensation now runs on a bounded context detached from daemon cancellation. If target cleanup or source rollback is not proven, terminalization is skipped and the nonterminal recovery fence is retained. TestSwitchAgentRestoresSourceAfterDaemonCancellationPostStop and TestSwitchAgentRetainsRecoveryWhenSourceRollbackFails cover both outcomes.

backend/internal/session_manager/agent_switching.go
whoisasx: @Annieeeee11 Fixed in 2321b0879. Worker-admission closure now returns the public ErrSwitchShuttingDown sentinel, which maps to an AGENT_SWITCH_UNAVAILABLE conflict with the clean message that AO is shutting down and cannot accept another switch. Both the terminal and ambiguous refusal paths plus API mapping are covered.

illegalcall:
@Annieeeee11 The requested changes are addressed in 2321b0879 and are already on this PR branch:

- Hide **Resume agent** whenever `session.activeAgentSwitch` is present.
- Return `ErrSwitchInProgress` (and the `AGENT_SWITCH_IN_PROGRESS` envelope) when resume races an active switch.
- Sanitize switch `model` control characters before validation and propagation.
- Map shutdown-time switch admission refusal to `AGENT_SWITCH_UNAVAILABLE`.

The two inline P3 threads are resolved. Focused session-manager, session-service, and controller regressions pass locally, and the current PR head has green Go, frontend, renderer-smoke, API-drift, gitleaks, and CLI E2E checks.

Ready for re-review.

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -845,6 +845,7 @@ function SessionCard({
 	const badge = getSessionStatusView(session.status, t);
 	const activity = getAgentActivityView(session.activity, t);
 	const showLiveActivity = session.status === "working" && activity.state === "active";
+	const activityIndicatorClassName = showLiveActivity ? activity.indicatorClassName : "bg-current";
```
illegalcall: **[P1] Show switch progress instead of the source Exited status.** This card still derives its label and tone exclusively from `session.status` and runtime activity, so after the source-stop boundary it renders `Exited` even though `session.activeAgentSwitch` is present. `ShellTopbar.tsx` has the same behavior. That directly contradicts the PR contract that the board and top bar show `Switching to …` rather than briefly reporting the source as Exited. Please use the shared agent-switch presentation for the label/tone on both surfaces while preserving activity as the underlying durable fact.

frontend/src/renderer/components/TerminalSwitchAgentButton.tsx
```diff
@@ -59,28 +61,29 @@ export function TerminalSwitchAgentButton({ session }: TerminalSwitchAgentButton
...
-						aria-busy={switching ? true : undefined}
+						aria-busy={switching && controlPresentation?.animate ? true : undefined}
 						aria-label={label}
 						className={cn(
 							"ml-1 grid size-6 shrink-0 place-items-center rounded-full border border-border/70 bg-background/45 text-muted-foreground transition-colors hover:border-border-strong hover:bg-interactive-hover hover:text-foreground focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-accent/50",
-							recoverySwitch && "border-warning/50 text-warning hover:border-warning/70 hover:text-warning",
+							warning && "border-warning/50 text-warning hover:border-warning/70 hover:text-warning",
 						)}
-						onClick={() => setOpen(true)}
+						disabled={blocksNewSwitch}
```
illegalcall, line 70: **[P1] Keep a desktop recovery path for retained nonterminal sagas.** `blocksNewSwitch` is true for every `in_progress` presentation, but the backend can intentionally retain states such as `stopping_source` after an ambiguous stop/commit and stop the worker. Those rows are not classified by `agentSwitchNeedsRecovery`, so this remains a disabled spinner while terminal input stays locked. Recovery is only retried at daemon boot or through another `SwitchAgent` admission, which this control now prevents. Surface these workerless retained states as recovery-required and allow the recovery admission/action, with coverage beyond `target_start_unconfirmed`.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -448,23 +443,16 @@ export function SessionView({ sessionId }: SessionViewProps) {
 		}
 		if (current?.browserContentRevealed) return;
 		setBrowserContentRevealed(sessionId, true);
```
illegalcall, line 571: **[P1] Restore automatic Browser reveal for `ao preview`.** This effect now only marks browser content as revealed; the calls that selected the Browser view and opened the inspector were removed, and the tests were rewritten to expect Summary/closed inspector instead. That regresses the existing preview contract: `ao preview` is supposed to render the requested target in the desktop Browser panel, not merely set an unseen badge that the user must discover. Restore the first-content `setInspectorViewForSession(..., "browser")` and `setInspectorOpenForSession(..., true)` behavior; later activity can still respect a user who deliberately leaves Browser.

illegalcall (commented):
## Review verdict: changes requested

Three blocking regressions remain in the current head: active switches are still shown as Exited on promised session surfaces, retained nonterminal failures have no desktop recovery path, and `ao preview` no longer opens the Browser inspector. I did not run tests or builds because this review task requires read-only inspection.

frontend/src/renderer/components/SessionsBoard.tsx
ronishrohan: Revalidated against current head `2321b0879`. This remains grounded in the PR contract, not a general design preference: the PR body explicitly promises one consistent switching experience across the board and top bar, and says session surfaces show `Switching to ...` instead of the source as `Exited`. The current tests assert the opposite. Please either derive these two surfaces from the shared active-switch presentation, or explicitly narrow the PR scope and description if this behavior is intentionally deferred.

frontend/src/renderer/components/TerminalSwitchAgentButton.tsx
ronishrohan: Revalidated against current head. The retained-failure path is concrete: a failed source rollback leaves `source_stopped`/`stopping_source` nonterminal and input-fenced; the renderer classifies it as ordinary `in_progress`, disables the switch control, and hides Resume while `activeAgentSwitch` exists. Boot reconciliation can recover it, so I would calibrate this as P2 rather than P1, but there is still no desktop action that triggers recovery. Please expose a durable recovery marker/action for this retained state and cover it with a focused UI test.

frontend/src/renderer/components/SessionView.tsx
ronishrohan: After a fresh skeptical pass, I do not think this is grounded enough to remain a blocker. The new tests make focus preservation intentional, and the AGENTS.md `ao preview` instruction establishes the developer demo workflow but does not by itself prove that new preview content must automatically steal inspector focus. I would treat automatic reveal as a product/design decision unless an explicit product contract requires it.

ronishrohan (commented):
Revalidation result on current head `2321b0879`: two actionable items remain. Fix order: (1) expose recovery for retained source-stop sagas so users are not dependent on a daemon restart; (2) align board/top-bar switch status with the PR body, or explicitly narrow the stated scope if that UI is deferred. I downgraded the recovery item to P2 because boot reconciliation exists. I withdrew the Browser auto-reveal item as a blocker, and dropped the proposed backend ambiguous-rollback finding because adapter retries converge on the deterministic handle. CI is green, and the deslop pass found no material code slop.

frontend/src/renderer/components/SessionsBoard.tsx
whoisasx: @illegalcall Addressed in a06b4bc8c. SessionsBoard and ShellTopbar now use the shared active-switch presentation instead of exposing the stopped source runtime as Exited. While a durable switch is active they show the phase-aware Switching to target status, with focused coverage on both surfaces.

frontend/src/renderer/components/TerminalSwitchAgentButton.tsx
whoisasx: @illegalcall Addressed in a06b4bc8c. Retained stopping_source sagas now carry the durable source_stop_unconfirmed recovery marker, render a static needs-attention state instead of an endless spinner, and expose a Check source action through the recovery endpoint. Recovery re-probes ownership, keeps the input fence closed while ambiguous, and only reopens input after the source is proven live or safely restored. Domain, manager, real-store, presentation, polling, and dialog regressions cover this path.

frontend/src/renderer/components/SessionView.tsx
whoisasx: @illegalcall Addressed in a06b4bc8c as the chosen product behavior. Entering a session with an existing preview still defaults to Summary, but a later live ao preview revision now selects Browser and opens the inspector. Preview clears do not steal focus, and subsequent new preview content reveals Browser again. SessionView regressions cover initial navigation, asynchronous hydration, live reveal, clear, and popout behavior.

ronishrohan:
Re-reviewed the latest changes. Everything looks fine from my side; the previous review feedback has been addressed.

Annieeeee11 (approved):
lgtm

illegalcall (approved):

neversettle17-101 (approved):
LGTM

PR #3736, opened 2026-08-08
feat: add Prime Agent harness

nikhilachale:
## What

Add Prime Agent as a first-class AO harness for worker and orchestrator sessions, including:

- Prime CLI launch, model configuration, cancellation, and install/auth probing
- AO-managed Prime lifecycle extension and activity derivation
- registry, CLI, API, delegation, SQLite, and desktop agent-picker integration
- the supplied Prime Agent icon

## Why

Prime Agent appeared in the local project configuration but could not be used to start worker or orchestrator sessions because AO did not register the harness or allow it in the sessions schema/API contract.

## How

- Introduces a dedicated Go adapter that launches `prime-agent --no-session` with AO's managed extension.
- Treats promptless startup as idle and only maps Prime's `session_shutdown` reason `quit` to durable exit; reload/new/resume/fork remain live.
- Adds migration `0081` without modifying shipped migrations, supporting both the current Muse constraint and the legacy QM variant.
- Keeps the internal `fake` harness non-selectable while exposing `prime-agent` in spawn and delegation contracts.

## Testing

- `cd backend && go test ./...` — passed
- pinned `golangci-lint v2.12.2` — 0 issues
- focused Prime/registry/CLI/HTTP/SQLite Go suites — passed
- `npm run typecheck` and `npm run typecheck:e2e` in `frontend/` — passed
- Prime avatar/agent-options tests plus rerun of the single full-suite flaky file — 30/30 passed
- full frontend run — 2046/2047 passed; one unrelated `SessionFilesView` timing failure passed on immediate focused rerun
- manually launched Prime Agent as both a worker and orchestrator in the native AO desktop app

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; links the related issue when applicable
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense
- [x] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

PR #3472, opened 2026-08-03
feat(chat): add Chat UI

illegalcall:
## Summary

This PR adds **Chat as a first-class interface for an AO session**, alongside the existing terminal UI. It does not replace AO's session model: projects, worktrees, branches, lifecycle state, issue/PR context, orchestration, and navigation remain shared. The difference is the controller and presentation used to interact with the coding agent.

- **TUI mode** keeps the existing runtime + tmux/terminal path.
- **Chat mode** talks to the provider through a structured native protocol or ACP connection and renders normalized messages, tools, plans, diffs, approvals, and usage.
- A compatible session can switch between the two interfaces while keeping its AO session and worktree. Only one agent controller is active at a time.
- Desktop and mobile both consume the daemon-owned conversation and transition APIs; neither client owns provider or lifecycle logic.

The final implementation is broader than the original Codex-only slice and includes Codex, Claude Code, OpenCode, Droid, desktop Chat, mobile Chat, and durable Chat/TUI handoff.

## Architecture

```text
AO session
├── shared: project, worktree, branch, role, lifecycle, issue/PR state
│
├── TUI interface
│   └── runtime controller → tmux/PTY → terminal renderer
│
└── Chat interface
    └── Chat controller → provider driver → normalized conversation events
                         ├── Codex app-server
                         └── ACP transport → Claude Code / OpenCode / Droid
```

The daemon remains the source of truth. `session_mode` is persisted on the session and is resolved by the daemon when omitted at creation. A user can explicitly select a mode for a new session, and an explicit interface transition can later change the persisted mode. Changing the global default does not silently rewrite existing sessions.

Chat sessions are runtime-less from the agent-controller perspective: they do not create a second tmux-hosted copy of the agent. The shell available from a Chat session is an ordinary shell in the same worktree and remains separate from the structured agent connection.

## Provider support

| Harness | Chat transport | Installation/auth behavior |
| --- | --- | --- |
| Codex | Native `codex app-server` protocol | Uses the user's installed and authenticated Codex CLI |
| Claude Code | `claude-agent-acp` bridge over the shared ACP transport | Uses the user's installed Claude Code executable, login, `~/.claude` configuration, settings, hooks, and MCP setup; AO does not bundle Claude Code itself |
| OpenCode | Native `opencode acp` command over the shared ACP transport | Uses the user's installed and configured OpenCode CLI |
| Droid | Droid's ACP daemon over the shared ACP transport | Uses the user's installed Droid CLI and account/API configuration |

[...]

PR #3771, opened 2026-08-09
fix(chat): enable Switch to Chat on new sessions before first prompt

### LINKED ISSUE #3753: [Bug]: (chat): "Switch to Chat UI" button disabled on new sessions until first prompt + session switch
### Summary
A newly spawned Codex or Claude Code session (TUI mode) does not enable the "Switch to Chat UI" button until the user sends at least one prompt and then switches away from and back to the session. The button appears disabled with no explanation, making it seem like Chat mode is unavailable.

## Root Cause
The session interface transition status (`GET /api/v1/sessions/{sessionId}/interface-transition`) checks whether the session has a "native conversation" (provider thread ID) that can resume in the other interface. For Codex and Claude Code, this native conversation ID is reported asynchronously via a `SessionStart` hook — it is not available immediately after spawn, and only appears after the agent begins its first turn.

The frontend hook `useSessionInterfaceTransition` (`frontend/src/renderer/hooks/useSessionInterfaceTransition.ts:50-78`) polls every 1 second while the reason code is `NATIVE_SESSION_MISSING`. However, the native conversation ID only becomes available after the agent receives and processes its first prompt. The workspace query refetches every 15 seconds and on CDC events, but it does not invalidate the interface transition status query. Only switching sessions (which unmounts and remounts the `SessionView` component) triggers a fresh fetch of the transition status, which then finds the now-available native session.

## Suggested Fix
Add `sessionInterfaceTransitionQueryKey(sessionId)` to the invalidation list in `useConversation.ts` or `event-transport.ts` when a `session_updated` CDC event arrives, so the transition status refetches alongside the workspace.

### Steps to reproduce
1. Start AO and open the desktop app
2. Spawn a new Codex or Claude Code session in TUI mode
3. Open the session view — the "Switch to Chat UI" button is disabled
4. Send at least one prompt to the agent
5. Switch to a different session in the sidebar
6. Switch back to the original session
7. The "Switch to Chat UI" button is now enabled and works

### Expected behavior
The "Switch to Chat UI" button should become enabled as soon as the agent's native conversation is available (after the first prompt or turn starts), without requiring the user to switch away and back.

### Actual behavior
The "Switch to Chat UI" button stays disabled until the user manually switches sessions to force a re-fetch of the interface transition status. There is no indication of what's wrong or that the user needs to do anything.

### AO version
586d863b

### Environment
Windows

### Logs / extra context
```shell
```

### Before submitting

bluwwi:
Fixes [#3753](https://github.com/Untrivial-ai/agent-orchestrator/issues/3753)

## Problem

A newly spawned Codex or Claude Code session (TUI mode) does not enable the
"Switch to Chat UI" button until the user sends at least one prompt and then
switches away from and back to the session.

#3735 (already on main) added a 1-second poll for `NATIVE_SESSION_MISSING`,
which fixes the common case: when the Codex SessionStart hook fires and
reports the native thread id, the poll catches it within ~1 second. But
two gaps remain:

1. **Poll latency.** The 1-second poll means the button can lag up to 1
   second behind the actual session state. A CDC `session_updated` event
   fires the moment `agent_session_id` or `provider_conversation_id` lands,
   but the interface-transition query was not invalidated on that event —
   only the workspace, SCM, usage, and conversation queries were.
2. **Windows hook-never-fires.** On Windows, the Codex SessionStart hook
   does not fire reliably (ConPTY command quoting). #3735's poll waits for
   that hook indefinitely, so the button stays disabled forever. There was
   no path to enable the switch without the native id, even when the TUI
   has no conversation history to lose.

### 1. `fix(storage): fan out session_updated on provider_conversation_id`
`backend/internal/storage/sqlite/migrations/0086_session_cdc_provider_conversation_id.sql`

Adds `provider_conversation_id` to the `sessions_cdc_update` `WHEN` guard
that `0085_agent_switching` established (which already included
`agent_session_id`). The Chat driver writes `provider_conversation_id`
independently of `agent_session_id` (`chat_spawn.go: ControllerReady`), so
a write to it must fan out a `session_updated` event. Guard-only change:
`session_updated` is an invalidation-only nudge (consumers refetch), so the
payload stays as-is — matching the convention established by migration
0024.

Renumbered from 0085 to 0086 to coexist with main's
`0085_agent_switching.sql`.

### 2. `fix(chat): invalidate interface transition query on session_updated CDC`
`frontend/src/renderer/lib/event-transport.ts`

Tracks `pendingTransitionSessions` alongside the existing
`pendingConversationSessions`. Any `session_updated` CDC event with a
`sessionId` now invalidates `sessionInterfaceTransitionQueryKey(sessionId)`,
so the "Switch to Chat" button re-evaluates immediately on push rather than
waiting for the next poll cycle. Gated on `session_updated` only — PR/check/
review events do not affect interface readiness.

### 3. `fix(chat): allow fresh-start transition when adapter has history probe`
`backend/internal/session_manager/interface_transition.go`

When the SessionStart hook has not reported the native conversation id
(e.g. Codex on Windows where the hook does not fire), the interface
transition status blocked as `NATIVE_SESSION_MISSING` — the "Switch to
Chat" button stayed disabled indefinitely.

Instead of fabricating a synthetic UUID, model the fresh-start path in
transition readiness: if the adapter implements
`AgentInterfaceHandoffHistoryProbe`, an empty native id is allowed. The
history probe distinguishes a real conversation from an empty TUI, and
`persistedNativeConversationID` uses that to decide between resume and
fresh start. Adapters without a probe keep the hard block — there is no
safe way to know whether the TUI's work would be lost.

This is the gap #3735 does not cover: when the hook never fires at all,
the poll waits forever. The history probe enables the switch without the
native id, safely.

## Testing

- `go test ./internal/storage/sqlite/store/...` — pass (3 CDC schema tests:
  `agent_session_id` fires 1 event, `provider_conversation_id` fires 1 event,
  unrelated metadata fires 0)
- `go test ./internal/session_manager/...` — pass (interface transition tests
  including fresh-start when reserved id has no history)
- `go test ./internal/adapters/chatdriver/claudeacp/...` — pass
- `go vet ./internal/session_manager/... ./internal/storage/sqlite/... ./internal/adapters/chatdriver/claudeacp/...` — clean
- `npm run typecheck` — clean
- `vitest run src/renderer/lib/event-transport.test.ts` — 14/14 pass
  (including: `session_updated` invalidates transition query, `pr_created`
  does not)

## Checklist
- [x] Migration 0086 added to `shippedMigrations` ledger
- [x] One issue per PR (#3753)
- [x] Conventional commits
- [x] Surgical changes — no speculative abstractions
- [x] Reframed CDC as latency improvement (#3735 already fixed the poll)
- [x] Removed "Supersedes #3736" (unrelated, already-merged Prime Agent PR)

### DIFF
--- a/backend/internal/storage/sqlite/migrations/0086_session_cdc_provider_conversation_id.sql
+++ b/backend/internal/storage/sqlite/migrations/0086_session_cdc_provider_conversation_id.sql
@@ -1,0 +1,90 @@
+-- Add provider_conversation_id to the sessions_cdc_update WHEN guard that
+-- 0085_agent_switching established. The Chat driver writes this column
+-- independently of agent_session_id (chat_spawn.go: ControllerReady), so a
+-- write to it must fan out a session_updated event for the interface-transition
+-- query to refetch. Guard-only change: session_updated is an invalidation-only
+-- nudge (consumers refetch, see renderer/lib/event-transport.ts), so the
+-- payload stays as-is — matching the convention established by migration 0024.
+
+-- +goose Up
+-- +goose StatementBegin
+DROP TRIGGER IF EXISTS sessions_cdc_update;
+CREATE TRIGGER sessions_cdc_update
+AFTER UPDATE ON sessions
+WHEN OLD.activity_state <> NEW.activity_state
+    OR OLD.is_terminated <> NEW.is_terminated
+    OR (OLD.first_signal_at IS NULL AND NEW.first_signal_at IS NOT NULL)
+    OR OLD.preview_url <> NEW.preview_url
+    OR OLD.preview_revision <> NEW.preview_revision
+    OR OLD.display_name <> NEW.display_name
+    OR OLD.terminate_on_pr_merge <> NEW.terminate_on_pr_merge
+    OR OLD.is_pinned <> NEW.is_pinned
+    OR OLD.pinned_at <> NEW.pinned_at
+    OR (OLD.pinned_at IS NULL AND NEW.pinned_at IS NOT NULL)
+    OR (OLD.pinned_at IS NOT NULL AND NEW.pinned_at IS NULL)
+    OR OLD.session_mode <> NEW.session_mode
+    OR OLD.auto_inject_review <> NEW.auto_inject_review
+    OR OLD.harness <> NEW.harness
+    OR OLD.runtime_launch_id <> NEW.runtime_launch_id
+    OR OLD.agent_session_id <> NEW.agent_session_id
+    OR OLD.native_transcript_path <> NEW.native_transcript_path
+    OR OLD.provider_conversation_id <> NEW.provider_conversation_id
+BEGIN
+    INSERT INTO change_log (project_id, session_id, event_type, payload, created_at)
+    VALUES (NEW.project_id, NEW.id, 'session_updated',
+        json_object(
+            'id', NEW.id,
+            'activity', NEW.activity_state,
+            'isTerminated', json(CASE WHEN NEW.is_terminated THEN 'true' ELSE 'false' END),
+            'terminateOnPrMerge', json(CASE WHEN NEW.terminate_on_pr_merge THEN 'true' ELSE 'false' END),
+            'previewUrl', NEW.preview_url,
+            'previewRevision', NEW.preview_revision,
+            'isPinned', json(CASE WHEN NEW.is_pinned THEN 'true' ELSE 'false' END),
+            'mode', NEW.session_mode,
+            'autoInjectReview', json(CASE WHEN NEW.auto_inject_review THEN 'true' ELSE 'false' END)
+        ),
+        NEW.updated_at);
+END;
+-- +goose StatementEnd
+
+-- +goose Down
+-- Restore the trigger 0085_agent_switching established (without
+-- provider_conversation_id in the guard).
+-- +goose StatementBegin
+DROP TRIGGER IF EXISTS sessions_cdc_update;
+CREATE TRIGGER sessions_cdc_update
+AFTER UPDATE ON sessions
+WHEN OLD.activity_state <> NEW.activity_state
+    OR OLD.is_terminated <> NEW.is_terminated
+    OR (OLD.first_signal_at IS NULL AND NEW.first_signal_at IS NOT NULL)
+    OR OLD.preview_url <> NEW.preview_url
+    OR OLD.preview_revision <> NEW.preview_revision
+    OR OLD.display_name <> NEW.display_name
+    OR OLD.terminate_on_pr_merge <> NEW.terminate_on_pr_merge
+    OR OLD.is_pinned <> NEW.is_pinned
+    OR OLD.pinned_at <> NEW.pinned_at
+    OR (OLD.pinned_at IS NULL AND NEW.pinned_at IS NOT NULL)
+    OR (OLD.pinned_at IS NOT NULL AND NEW.pinned_at IS NULL)
+    OR OLD.session_mode <> NEW.session_mode
+    OR OLD.auto_inject_review <> NEW.auto_inject_review
+    OR OLD.harness <> NEW.harness
+    OR OLD.runtime_launch_id <> NEW.runtime_launch_id
+    OR OLD.agent_session_id <> NEW.agent_session_id
+    OR OLD.native_transcript_path <> NEW.native_transcript_path
+BEGIN
+    INSERT INTO change_log (project_id, session_id, event_type, payload, created_at)
+    VALUES (NEW.project_id, NEW.id, 'session_updated',
+        json_object(
+            'id', NEW.id,
+            'activity', NEW.activity_state,
+            'isTerminated', json(CASE WHEN NEW.is_terminated THEN 'true' ELSE 'false' END),
+            'terminateOnPrMerge', json(CASE WHEN NEW.terminate_on_pr_merge THEN 'true' ELSE 'false' END),
+            'previewUrl', NEW.preview_url,
+            'previewRevision', NEW.preview_revision,
+            'isPinned', json(CASE WHEN NEW.is_pinned THEN 'true' ELSE 'false' END),
+            'mode', NEW.session_mode,
+            'autoInjectReview', json(CASE WHEN NEW.auto_inject_review THEN 'true' ELSE 'false' END)
+        ),
+        NEW.updated_at);
+END;
+-- +goose StatementEnd
--- a/frontend/src/renderer/lib/event-transport.ts
+++ b/frontend/src/renderer/lib/event-transport.ts
@@ -5,6 +5,7 @@ import { setEventsConnectionState } from "./events-connection";
 import { workspaceQueryKey } from "../hooks/useWorkspaceQuery";
 import { sessionScmSummaryQueryKey } from "../hooks/useSessionScmSummary";
 import { conversationQueryKey } from "../hooks/useConversation";
+import { sessionInterfaceTransitionQueryKey } from "../hooks/useSessionInterfaceTransition";
 import { sessionUsageQueryRoot } from "../hooks/useSessionUsageSummaries";
 
 export type EventTransport = {
@@ -47,6 +48,7 @@ export function createEventTransport(queryClient: QueryClient): EventTransport {
 		connect() {
 			let debounce: ReturnType<typeof setTimeout> | undefined;
 			const pendingConversationSessions = new Set<string>();
+			const pendingTransitionSessions = new Set<string>();
 			let workspaceInvalidationPending = false;
 			let retryTimer: ReturnType<typeof setTimeout> | undefined;
 			let source: EventSource | undefined;
@@ -54,30 +56,31 @@ export function createEventTransport(queryClient: QueryClient): EventTransport {
 			const refreshWorkspaces = (event?: Event) => {
 				let conversationOnly = false;
 				if (event && "data" in event) {
-					try {
-						const decoded = JSON.parse(String((event as MessageEvent).data)) as {
-							sessionId?: unknown;
-							payload?: unknown;
-						};
-						// The SSE endpoint sends the complete durable CDC event. Routing
-						// fields such as sessionId live on that envelope, while trigger-built
-						// details such as conversationId live inside its payload. Do not
-						// mistake the payload for the entire event: doing so refreshes the
-						// sidebar but leaves a Chat timeline frozen on its pre-turn snapshot.
-						const payload =
-							typeof decoded.payload === "object" && decoded.payload !== null
-								? (decoded.payload as { conversationId?: unknown })
-								: undefined;
+				try {
+					const decoded = JSON.parse(String((event as MessageEvent).data)) as {
+						sessionId?: unknown;
+						payload?: unknown;
+					};
+					// The SSE endpoint sends the complete durable CDC event. Routing
+					// fields such as sessionId live on that envelope, while trigger-built
+					// details such as conversationId live inside its payload. Do not
+					// mistake the payload for the entire event: doing so refreshes the
+					// sidebar but leaves a Chat timeline frozen on its pre-turn snapshot.
+					const payload =
+						typeof decoded.payload === "object" && decoded.payload !== null
+							? (decoded.payload as { conversationId?: unknown })
+							: undefined;
+					if (typeof decoded.sessionId === "string" && decoded.sessionId) {
+						pendingTransitionSessions.add(decoded.sessionId);
 						if (
-							typeof decoded.sessionId === "string" &&
-							decoded.sessionId &&
 							typeof payload?.conversationId === "string" &&
 							payload.conversationId
 						) {
 							pendingConversationSessions.add(decoded.sessionId);
 							conversationOnly = true;
 						}
-					} catch {
+					}
+				} catch {
 						// A malformed CDC payload still invalidates workspaces; it simply
 						// cannot target a conversation cache precisely.
 					}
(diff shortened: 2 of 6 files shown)


backend/internal/adapters/agent/codex/codex.go
```diff
@@ -203,6 +216,9 @@ func (p *Plugin) NativeConversationID(
 		return id, id != "", nil
 	}
 	id := strings.TrimSpace(session.Metadata[ports.MetadataKeyAgentSessionID])
+	if id == "" && session.ID != "" {
+		id = codexSessionUUID(session.ID)
```
illegalcall: **Blocking - this is not the live Codex thread ID.** The TUI is never launched with this derived UUID. `NativeConversationID` therefore makes status report `supported=true`, but on POST the history probe finds no rollout for the synthetic value and the coordinator converts it to the empty "start fresh" sentinel. If the first turn has materialized while its SessionStart hook is still pending, switching can stop the TUI and open an empty Chat without that context. Please continue returning `("", false)` until the hook supplies the actual `agent_session_id`. If switching an empty TUI to fresh Chat is desired, model that explicitly in transition readiness rather than fabricating provider identity.

backend/internal/adapters/chatdriver/claudeacp/driver.go
```diff
@@ -73,6 +73,7 @@ func New(plugin claudePlugin, log *slog.Logger) ports.ChatDriver {
 			if err != nil {
 				return acpdriver.Launch{}, fmt.Errorf("%w: %w", ports.ErrChatDriverUnavailable, err)
 			}
+			claudeBinary = resolveClaudeExe(claudeBinary)
```
illegalcall: This Windows `.cmd` workaround addresses a separate post-click ACP launch failure, while #3753 is about readiness refreshing when native identity becomes available. Please split it into its own issue/PR. That PR also needs direct cases for a sibling `.exe`, a quoted `%dp0%` target, malformed or missing targets, and pass-through/failure behavior; the current non-Windows tests never exercise this branch.

backend/internal/adapters/chatdriver/claudeacp/driver.go
```diff
@@ -201,6 +202,59 @@ func requireFile(path, label string) error {
...
+// to. Node child_process.spawn cannot execute .cmd files, so the ACP runtime
+// needs the real executable path. It first checks for a same-basename .exe
+// alongside the shim; if none exists it parses the .cmd content for a quoted
+// .exe path, expanding %dp0% to the shim's directory. A common npm shim looks
+// like:
+//
+//	"%dp0%\node_modules\@anthropic-ai\claude-code\bin\claude.exe" %*
+//
+// Non-.cmd paths and non-Windows platforms pass through unchanged.
+func resolveClaudeExe(binary string) string {
```
illegalcall: `resolveClaudeExe` performs `os.Stat` and `os.ReadFile`, but repository guidance requires `context.Context` as the first argument for functions doing I/O. If this helper is retained in its own PR, please accept `ctx context.Context`, check cancellation around the filesystem work, and propagate it from `Launch`.

backend/internal/storage/sqlite/migrations/0086_session_cdc_provider_conversation_id.sql
```diff
@@ -0,0 +1,80 @@
...
+    OR OLD.preview_revision <> NEW.preview_revision
+    OR OLD.display_name <> NEW.display_name
+    OR OLD.terminate_on_pr_merge <> NEW.terminate_on_pr_merge
+    OR OLD.is_pinned <> NEW.is_pinned
+    OR OLD.pinned_at <> NEW.pinned_at
+    OR (OLD.pinned_at IS NULL AND NEW.pinned_at IS NOT NULL)
+    OR (OLD.pinned_at IS NOT NULL AND NEW.pinned_at IS NULL)
+    OR OLD.session_mode <> NEW.session_mode
+    OR OLD.auto_inject_review <> NEW.auto_inject_review
+    OR OLD.agent_session_id <> NEW.agent_session_id
```
illegalcall, line 29: This permanently changes the final `sessions_cdc_update` trigger, but the only test change is the migration ledger. Please add final-schema coverage that updates `agent_session_id` and `provider_conversation_id` separately and asserts exactly one `session_updated` for each, plus no event for unchanged values or unrelated metadata. That proves both new guard terms without weakening CDC suppression.

frontend/src/renderer/lib/event-transport.ts
```diff
@@ -47,37 +48,39 @@ export function createEventTransport(queryClient: QueryClient): EventTransport {
...
+					// fields such as sessionId live on that envelope, while trigger-built
+					// details such as conversationId live inside its payload. Do not
+					// mistake the payload for the entire event: doing so refreshes the
+					// sidebar but leaves a Chat timeline frozen on its pre-turn snapshot.
+					const payload =
+						typeof decoded.payload === "object" && decoded.payload !== null
+							? (decoded.payload as { conversationId?: unknown })
+							: undefined;
+					if (typeof decoded.sessionId === "string" && decoded.sessionId) {
+						pendingTransitionSessions.add(decoded.sessionId);
```
illegalcall: `refreshWorkspaces` is registered for every `CDC_EVENT_TYPES` entry (and unnamed events), so this queues transition invalidation for PR/check/review events too. The requested freshness edge is specifically `session_updated`. Please gate this on the event type (or use a dedicated listener), then add a focused test proving `session_updated` invalidates `sessionInterfaceTransitionQueryKey(sessionId)` while a PR event does not.

backend/internal/adapters/agent/codex/codex_test.go
```diff
@@ -16,9 +16,10 @@ import (
 
 func TestNativeConversationIDRequiresCapturedCodexThreadForTUI(t *testing.T) {
```
illegalcall, line 17: Nit: this name now contradicts the first assertion. `RequiresCapturedCodexThreadForTUI` says a captured thread is required, while the test expects success with empty metadata. If the fallback remains after the blocking concern is resolved, rename the test to describe the actual fallback/override behavior.

frontend/src/renderer/lib/event-transport.ts
```diff
@@ -47,37 +48,39 @@ export function createEventTransport(queryClient: QueryClient): EventTransport {
 		connect() {
 			let debounce: ReturnType<typeof setTimeout> | undefined;
 			const pendingConversationSessions = new Set<string>();
+			const pendingTransitionSessions = new Set<string>();
 			let workspaceInvalidationPending = false;
 			let retryTimer: ReturnType<typeof setTimeout> | undefined;
 			let source: EventSource | undefined;
 			let sourceBaseUrl: string | undefined;
 			const refreshWorkspaces = (event?: Event) => {
```
illegalcall: Nit: `refreshWorkspaces` now routes workspace, SCM, usage, conversation, and interface-transition invalidations. Please rename it to describe that broader responsibility, such as `scheduleCacheInvalidations`.

frontend/src/renderer/lib/event-transport.ts
```diff
@@ -47,37 +48,39 @@ export function createEventTransport(queryClient: QueryClient): EventTransport {
...
-						// The SSE endpoint sends the complete durable CDC event. Routing
-						// fields such as sessionId live on that envelope, while trigger-built
-						// details such as conversationId live inside its payload. Do not
-						// mistake the payload for the entire event: doing so refreshes the
-						// sidebar but leaves a Chat timeline frozen on its pre-turn snapshot.
-						const payload =
-							typeof decoded.payload === "object" && decoded.payload !== null
-								? (decoded.payload as { conversationId?: unknown })
-								: undefined;
+				try {
```
illegalcall: Please restore the existing `try`/`catch` nesting here. This hunk de-indents the whole parse block and over-indents the catch comment, creating unrelated formatting churn in an already delicate routing function.

illegalcall (changes requested):
I'm requesting changes on two independent axes.

**Spec**
- The merge base already contains #3735: `useSessionInterfaceTransition` polls every second for as long as `reasonCode === "NATIVE_SESSION_MISSING"`, and its regression test proves readiness becomes supported without remounting. The PR description's claim that the poll "gives up" is not true on the code being changed. Please either add a failing regression against current `main`, or reframe the CDC work as a latency/efficiency improvement.
- The synthetic Codex identity is blocking; see the inline comment. It makes the control look ready without naming the live TUI conversation.

**Standards / scope**
- The PR itself calls the Windows Claude `.cmd` failure a "fourth, separate" bug. Repository guidance requires surgical changes and one issue per PR, so please split that workaround into its own issue/PR.
- Please remove or correct `Supersedes #3736`: #3736 is the unrelated, already-merged Prime Agent harness PR.
- Inline comments cover the missing I/O context, the three untested boundaries, over-broad event routing, and the naming/formatting issues.

The existing targeted suites are green, but none of the new migration, transition-invalidation, or Windows-shim branches has direct regression coverage.

bluwwi:
@illegalcall everything is done except .cmd wrapper is not removed until i make a new PR after that i will revert the commit 5.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

bluwwi:
@illegalcall — all 9 review comments addressed, `.cmd` fix split out, migration renumbered 0085→0086, rebased on latest main (zero merge conflicts).

### What #3735 fixed vs what #3771 fixes

**#3735 (merged)** fixed the common case on Mac/Linux: pin the Codex `SessionStart` hook to the daemon binary so it fires, then poll every 1s for `NATIVE_SESSION_MISSING`. On Mac/Linux the hook fires → the poll catches the native id → button enables.

On **Windows** this doesn't work. ConPTY's shell quoting breaks the pinned path — the hook command arrives at the login shell mangled, so it never executes. The native id never lands. The poll keeps asking and always gets `NATIVE_SESSION_MISSING`. The button stays disabled forever.

**Confirmed on clean upstream main (Windows):**

```
GET /api/v1/sessions/fabor-2/interface-transition
→ { "supported": false, "reasonCode": "NATIVE_SESSION_MISSING" }
```

Session is idle but the button never enables. #3735's poll is running, waiting for a hook that never fires.

### The core fix: history probe fresh-start path

`interface_transition.go` — 9 lines added:

```go
if !ok || id == "" {
    if _, ok := handoff.(ports.AgentInterfaceHandoffHistoryProbe); ok {
        return "", handoff, nil
    }
    return "", handoff, fmt.Errorf("%w for %s", ErrNativeConversationMissing, rec.Harness)
}
```

Main already has `persistedNativeConversationID()` (from #3472/#3548) which calls `NativeConversationExists()` (added by #3735) to walk `CODEX_HOME/sessions/` for rollout files. But it's **unreachable** because `nativeConversationID()` errors out before it's called. This change lets the empty id pass through so the probe can run.

### No synthetic UUID — your blocking concern is resolved

The earlier `codexSessionUUID` approach is **completely removed** (commit `cf09844ce`). `codex.go` has zero changes in this PR. We return `("", nil)` — the empty string is the explicit "start fresh" sentinel that `persistedNativeConversationID()` already understands. No fabrication, no hardcoding.

### What happens to the session id

The **AO session id** (e.g., `fabor-2`) never changes. The **native conversation id** is empty because the hook never fired. When the user clicks Switch to Chat:

1. `nativeConversationID()` returns `("", nil)` — our change
2. `persistedNativeConversationID()` runs → `NativeConversationExists()` finds no rollout → returns `""` (fresh start)
3. Chat driver starts with empty `ProviderConversationID` → creates a new conversation
4. `ControllerReady` callback receives the **real** conversation id from the Chat driver → saves it to SQLite via `MarkSpawned`

No reassignment, no fabrication. The real id comes from the Chat driver.

### Capability matrix

| | Mac/Linux | Windows |
|---|---|---|
| #3735 (poll + hook pinning) | ✅ hook fires → poll catches | ❌ hook never fires (ConPTY) → poll waits forever |
| #3735 (NativeConversationExists) | ✅ used when reached | ❌ never reached (blocked by error) |
| #3771 (history probe fresh-start) | ✅ bonus (redundant) | ✅ **only fix** — bypasses hook, checks filesystem |
| #3771 (CDC push invalidation) | ✅ latency: 1s → instant | ✅ (helps, doesn't solve alone) |

### Testing

- `go test ./internal/storage/sqlite/store/...` — pass (3 CDC schema tests)
- `go test ./internal/session_manager/...` — pass (interface transition including fresh-start)
- `go test ./internal/adapters/chatdriver/claudeacp/...` — pass
- `go vet` — clean
- `npm run typecheck` — clean
- `vitest run src/renderer/lib/event-transport.test.ts` — 14/14 pass

### Merge safety

- Zero file overlap with #3735
- Zero merge conflicts against current main (`git merge-tree` exit 0)
- Migration 0086 — next free number, no collision
- Trigger includes all of main's 0085 columns

### Separate .cmd PR

Claude Code `spawn EINVAL` on Windows. `resolveClaudeExe` resolves `.cmd`→`.exe`. Being filed separately per your request.

i-trytoohard:
Match confidence: `0.64` (fuzzy).

Relevant discussion excerpts:
- - *prateek* asked whether users can enable chat UI by default. - *anaya* said users can spawn a session directly with chat UI, and a PR to allow switching from terminal to chat in an existing session should be pushed **today**. - *Dhruv Sharma* said there is already a default-interface setting, then corrected [...]
- - *prateek* suggested a default behavior for diffs: open **1–2 files individually by default**, and collapse when there are **3+ files**. - *prateek* asked whether users can enable chat UI by default. - *anaya* said users can spawn a session directly with chat UI, and a PR to allow switching from terminal to chat [...]
- - *prateek* asked whether users can enable chat UI by default. - *anaya* said users can spawn a session directly with chat UI, and a PR to allow switching from terminal to chat in an existing session should be pushed **today**.

PR #3772, opened 2026-08-09
feat: add per-PR CI failure injection policy

VenkataSakethDakuri:
## What

Add a session-level toggle that controls the default CI failure injection policy for newly created pull requests.

- Persist the session default and snapshot it onto each PR when that PR is first observed.
- Keep the PR snapshot immutable when the session toggle changes later.
- Skip CI failure nudges when the PR snapshot is disabled while preserving review and conflict reactions.
- Expose the policy through the session API and generated frontend types.
- Add the Session Inspector toggle, explanatory tooltip, optimistic updates, rollback errors, and the "CI failures not injected" status for affected failing PRs.

## Why

Users need to decide before a PR is created whether that PR's CI failures should be sent automatically to its worker. The decision must remain stable for the life of the PR so changing the session default only affects future PRs.

## How

- Add `auto_inject_ci` columns to sessions and PRs in migration 0088.
- Snapshot the current session value in the PR insert query; conflict updates intentionally do not overwrite it.
- Read the persisted PR policy in lifecycle reactions before enqueueing a CI nudge.
- Add `PATCH /api/v1/sessions/{sessionId}/auto-inject-ci` and regenerate OpenAPI/sqlc artifacts.
- Render and mutate the setting from the Pull Request section of Session Inspector, including before a PR exists.

## Testing

- `go test ./internal/storage/sqlite/store ./internal/service/session ./internal/httpd/controllers ./internal/lifecycle`
- `npm test -- src/renderer/components/SessionInspector.test.tsx src/renderer/hooks/useWorkspaceQuery.test.tsx src/renderer/components/PRSummaryDisplay.test.tsx src/renderer/lib/pr-display.test.ts` (128 tests)
- `npm run typecheck`
- `npm run typecheck:e2e`
- `npm run test:e2e -- e2e/ci-auto-inject.spec.ts` (2 Playwright scenarios)
- `go build ./...`
- `git diff --check`


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/storage/sqlite/migrations/0088_add_auto_inject_ci_toggle.sql
```diff
@@ -0,0 +1,91 @@
+-- Persist the per-session default for automatic CI-failure injection and
+-- snapshot it on each PR when that PR is first recorded.
+
+-- +goose Up
```
whoisasx, line 4: [P1] Current main already contains migration 0085_agent_switching.sql. Keeping this file at version 0085 after resolving the merge conflicts gives Goose two migrations with the same version, which prevents migrations and daemon startup. Please rebase and renumber this migration to the next available version, including the burned-version expectation.

backend/internal/storage/sqlite/queries/pr.sql
```diff
@@ -6,9 +6,10 @@ INSERT INTO pr (
     is_draft, is_merged, is_closed,
     provider_state, provider_mergeable, provider_merge_state_status, html_url,
     created_at_provider, updated_at_provider, merged_at_provider, closed_at_provider,
-    metadata_hash, ci_hash, review_hash, observed_at, ci_observed_at, review_observed_at
+    metadata_hash, ci_hash, review_hash, observed_at, ci_observed_at, review_observed_at, auto_inject_ci
 )
-VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,
+    COALESCE((SELECT auto_inject_ci FROM sessions WHERE id = ?), TRUE))
```
whoisasx, line 12: [P2] This snapshots the policy for observer upserts, but ClaimPRForSession later in this file still inserts a new PR without copying sessions.auto_inject_ci. As a result, ao spawn --claim-pr and ao session claim-pr receive SQLite's TRUE default even when the session default is disabled. The later upsert preserves that incorrect snapshot permanently and allows CI nudges to be injected. Please snapshot the owning session's value in the claim insert and add a disabled-session claim regression test.

backend/internal/service/session/pr_summary.go
```diff
@@ -220,6 +221,7 @@ func summarizeCI(pr domain.PullRequest, checks []domain.PullRequestCheck) PRCISu
 			URL:        ch.URL,
 		})
 	}
+	out.AutoInjectCI = pr.AutoInjectCI
```
whoisasx: [P2] This assignment is reached only for active failing PRs because the function returns earlier for every other CI state and for merged or closed PRs. Passing, pending, unknown, merged, and closed PRs therefore expose false even when their persisted immutable policy is true. Please initialize the summary with AutoInjectCI from the PR before branching, and cover at least a passing and a terminal PR.

whoisasx (changes requested):
## Review summary

The per-PR snapshot design is the right direction, but three correctness issues need to be resolved before merging:

- Renumber the migration after rebasing because current main already owns version 0085.
- Snapshot the session policy when a PR is first created through the claim flow.
- Return the persisted PR policy for every CI and terminal state, not only active failing PRs.

Please add regression coverage for a disabled session claiming a new PR and for enabled policies on passing and terminal PR summaries. The current GitHub checks are green against the stale base; current-main integration still needs verification after the conflicts are resolved.

VenkataSakethDakuri:
@whoisasx @neversettle17-101 can you guys review the PR, looking to merge before more conflicts arise.

neversettle17-101 (commented):
Can you show how the UI looks like?

VenkataSakethDakuri:
image
@whoisasx @neversettle17-101

backend/internal/httpd/controllers/sessions.go
```diff
@@ -966,6 +968,29 @@ func (c *SessionsController) setAutoInjectReviewPolicy(w http.ResponseWriter, r
 	})
 }
 
+func (c *SessionsController) setAutoInjectCIPolicy(w http.ResponseWriter, r *http.Request) {
+	if c.Svc == nil {
+		apispec.NotImplemented(w, r, "PATCH", "/api/v1/sessions/{sessionId}/auto-inject-ci")
+		return
+	}
+	var in SetSessionAutoInjectCIRequest
```
whoisasx: [P3] Reject requests that omit `autoInjectCI`

The generated OpenAPI contract marks `autoInjectCI` as required, but decoding into a plain `bool` makes `{}` indistinguishable from `{"autoInjectCI": false}`. The handler therefore returns 200 and silently disables CI injection for a malformed request. The added validation test only covers invalid JSON (`{`), not a syntactically valid body missing the required property. Please use presence-aware decoding or explicit validation, return 400 when the field is absent, and add a `{}` regression test.

whoisasx (changes requested):
Thanks for addressing the earlier review feedback. I verified that the claim flow now snapshots the session policy, passing and terminal PR summaries retain the persisted value, and the migration was renumbered for the previous base.

One API-contract issue remains: the new endpoint accepts an empty object as `autoInjectCI: false`, even though the generated OpenAPI schema requires that property. Please make the request presence-aware and cover `{}` with a regression test.

Verification at `efb7a165`:
- Focused SQLite, session-service, controller, lifecycle, and session-manager tests pass.
- Race tests for the SQLite store and lifecycle packages pass.
- `git diff --check` passes.
- All ten GitHub checks are green.

The branch also conflicts with current `main`, which now owns migrations 0086 and 0087. After rebasing, this migration needs the next available version and the related migration expectations must be updated. Keeping this as changes requested until the request validation and current-main integration are resolved.

whoisasx:
@VenkataSakethDakuri, could you please take another look at the placement and visual treatment of these controls? Are we confident this is the UI we want to ship? The current toggle/button rows feel a little crowded in the inspector, so I would like a quick design sanity check on whether this looks polished enough before merge.

VenkataSakethDakuri:
@whoisasx Yeah I feel it is best to keep it at the top so that it conveys that it is valid for all PRs. I can reduce the name of the toggle, that can make it less crowded.

whoisasx:
@codebanditssss is going to revamp it e2e. I assume we can keep it till that time @VenkataSakethDakuri . 
But still we can make that toggle button a bit small and rounded. what do you say @codebanditssss ?

whoisasx (commented):
Final status: the implementation and the fixes from the earlier review look good to me. The behavior is covered, all current checks are green, and I do not have any remaining code-correctness concerns.

Before we ship, please close out these two polish items:

- Update the PR description: it still says the feature adds migration `0085`, while the rebased implementation now uses `0088`.
- Please address the UI treatment discussed in the conversation above, particularly whether the controls should be more compact and whether the toggle sizing/rounding looks polished enough for the inspector. Please align with @codebanditssss on the final treatment.

Other than those documentation and UI details, everything looks good to me and I think we can ship it. I am leaving this as a comment rather than approving until those items are addressed.

neversettle17-101:
Under the summary section, the first thing the user is interested is the PRs that are generated. The settings toggle will come later. Keep it at the end after the reviews section.

VenkataSakethDakuri:
@whoisasx @neversettle17-101 addressed your comments

neversettle17-101 (approved):

whoisasx (approved):

PR #2852, opened 2026-07-20
fix(dev): gate daemon isolation behind opt-in

GitanshKapoor:
## What

Restore the shared daemon port and state directory for development builds by default.

- Development now uses port `3001` and the standard `~/.ao` state directory.
- Set `ISOLATE_DEV=true` to opt into port `3002` and `~/.ao/dev`.
- The Vite `/api` and `/mux` proxies now derive their target port from the same shared configuration as Electron.
- Explicit `AO_PORT`, `AO_RUN_FILE`, `AO_DATA_DIR`, and `AO_DEV_API_TARGET` overrides remain supported.

## Why

The previous development isolation behavior caused Electron and Vite to use different daemon ports. It also silently separated development data from the installed application's data.

Fixes #2845.

## How

Added a side-effect-free shared dev-daemon configuration module used by both Electron and Vite.

The shared resolver:

- Uses port `3001` by default.
- Uses port `3002` only when `ISOLATE_DEV=true`.
- Gives an explicit `AO_PORT` precedence in either mode.

Electron only switches the run file and data directory to `~/.ao/dev` when isolation is explicitly enabled.

Regression tests cover the default port, isolated port, and explicit port overrides.

## Testing

- `npm test -- src/shared/dev-daemon-config.test.ts` - 3 tests passed
- `npm run typecheck` - passed
- `npm test` - 66 test files and 712 tests passed
- `git diff --check` - passed
- Started the application with `ISOLATE_DEV=true` and confirmed that the Vite notifications proxy targets `127.0.0.1:3002`

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; links the related issue when applicable
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for user-visible behavior where it makes sense
- [ ] Relevant CI checks pass for the area touched (`go`, frontend, etc.)

Issue #2845, opened 2026-07-19
Dev daemon isolation from #2465 regressed pnpm dev: revert to shared port/rundir, gate isolation behind ISOLATE_DEV


harshitsinghbhandari:
## What happened

#2465 moved the dev daemon (`app.isPackaged === false`) onto its own port and state dir: `AO_PORT=3002`, `AO_RUN_FILE=~/.ao/dev/running.json`, `AO_DATA_DIR=~/.ao/dev/data`. Two problems:

1. **Proxied renderer paths have been broken in dev since it merged.** `vite.renderer.config.ts` still defaults its `/api` and `/mux` proxy targets to `http://127.0.0.1:3001`, so anything going through the Vite proxy (notifications stream, `VITE_NO_ELECTRON` browser preview) gets ECONNREFUSED while the daemon sits healthy on 3002. Symptom: UI loads but data-driven surfaces are empty or stale, with `http proxy error: connect ECONNREFUSED 127.0.0.1:3001` in the dev console.
2. **Dev silently forks app state.** The fresh `~/.ao/dev/data` DB means projects, sessions, and persisted settings (e.g. `pipelines.enabled`) from the installed app vanish in dev with no indication why.

## Decision

The separate port and rundir are the regression, not the lifecycle coupling. The supervisor stopping the daemon on quit is by design: the daemon's lifecycle belongs to whichever app supervises it.

Rework:
- Dev uses the **same port and rundir as the installed app by default** (3001, `~/.ao`).
- The isolation from #2465 moves behind an explicit **`ISOLATE_DEV=true`** opt-in for anyone who wants a sandboxed dev daemon.
- The Vite proxy default must follow the same switch as the daemon port, or both should derive from one shared constant, so the two can't desync again. Hardcoding the port in two files (`main.ts` DEV_DAEMON_PORT vs `vite.renderer.config.ts`) is what let this ship.

## Repro

With an installed AO app present (or a stale `~/.ao/running.json` holding 3001): `pnpm dev` in `frontend/`, open the notifications stream or any proxied path. Observe ECONNREFUSED against 3001 while the dev daemon listens on 3002.

GitanshKapoor:
Hey! @harshitsinghbhandari I will picked this up and raised a PR https://github.com/AgentWrapper/agent-orchestrator/pull/2852 . Happy to get feedback! 🙌

CC: @whoisasx @illegalcall @somewherelostt

tamarkhl:
[...]

PR #2465, opened 2026-07-06
fix(dev): isolate dev daemon from installed app on port and state dir

Pulkit7070:
Fixes #2464

## Summary

- `runFilePath()` now returns `~/.ao/dev/running.json` in dev mode — this also moves `supervise.sock` to `~/.ao/dev/supervise.sock` automatically, since the backend derives it as `dir(RunFilePath)/supervise.sock`
- `daemonEnv()` injects `AO_PORT=3002`, `AO_RUN_FILE`, and `AO_DATA_DIR` defaults when `!app.isPackaged` (explicit env vars still win)
- New `resolvedDaemonPort()` helper used by all three port-probe call-sites so they stay consistent with the daemon's actual bind port
- Only `frontend/src/main.ts` changed — no backend changes needed, the backend already honours `AO_PORT`/`AO_RUN_FILE`/`AO_DATA_DIR`

## Test plan

- [ ] Launch installed app, then `npm run dev` — dev app starts its own daemon on port 3002, installed app keeps running
- [ ] Quit dev Electron — installed app's daemon is unaffected
- [ ] `npm run typecheck` passes (verified locally)
- [ ] Setting `AO_PORT=3001` in shell still works (explicit override wins)

cc @harshitsingh

PR #3773, opened 2026-08-09
fix: share dev daemon port/rundir with the installed app, gate isolation behind ISOLATE_DEV

krishrathi1:
## Root cause

#2465 moved the dev (`app.isPackaged === false`) daemon onto its own port (`3002`) and state dir (`~/.ao/dev`), hardcoded as `DEV_DAEMON_PORT` in `frontend/src/main.ts`. `frontend/vite.renderer.config.ts`'s `/api` and `/mux` proxy targets independently hardcoded `http://127.0.0.1:3001`. The two values had no shared source, so they silently drifted apart: the dev daemon listened on `3002` while the Vite dev-server proxy kept dialing `3001`, breaking the notification stream and `VITE_NO_ELECTRON` browser preview with `ECONNREFUSED`. Separately, dev silently forked app state into `~/.ao/dev/data`, so projects/sessions/settings from the installed app didn't show up in dev with no indication why.

## Fix

- Dev now spawns against the same port (`3001`) and rundir/datadir (`~/.ao`) as the packaged app by default, so it shares the installed app's state.
- The old isolated behavior (separate port + `~/.ao/dev` rundir/datadir) is preserved, opt-in via `ISOLATE_DEV=true` (or `"1"`/`"yes"`/`"on"`), matching the truthy-value allowlist convention `AO_KEEP_DAEMON` already uses in `frontend/src/main/daemon-owner.ts`.
- To close the structural gap that let the two files drift in the first place, `ISOLATED_DEV_DAEMON_PORT` and `isDevIsolationEnabled` now live in `frontend/src/shared/daemon-attach.ts` — a module already free of `node:*`/`electron` imports, so it loads cleanly under both vitest and the Vite config loader. Both `main.ts` and `vite.renderer.config.ts` import the same constants/function instead of hardcoding their own port literal, so the daemon's actual port and the Vite proxy's target port cannot independently drift again.
- Updated the `ao-desktop-dev` agent skill's isolated-mode recipes to set `ISOLATE_DEV=true` explicitly, since isolation is no longer the default from unsetting `AO_PORT`/`AO_RUN_FILE`/`AO_DATA_DIR` alone.

A prior attempt at this issue (#2852) was closed without merging on 2026-07-20; that PR's contents aren't visible from a closed PR, so this is a fresh, independent implementation working from the issue's own description of the desired fix, not a continuation of that attempt.

## Verification

- Extended `frontend/src/shared/daemon-attach.test.ts` with coverage for `isDevIsolationEnabled` (default-off, truthy-value allowlist matching `AO_KEEP_DAEMON`'s tests, whitespace trimming, unrecognized values) and for `ISOLATED_DEV_DAEMON_PORT`/`DEFAULT_DAEMON_PORT` staying distinct and being the single value both `main.ts` and `vite.renderer.config.ts` derive from.
- Confirmed the new tests fail against the pre-fix source (reverted only `daemon-attach.ts`/`main.ts`/`vite.renderer.config.ts` via a patch file, keeping the test changes) with `isDevIsolationEnabled is not a function` / `undefined` port errors, then confirmed all 52 tests in the file pass once the fix is reapplied.
- `npx tsc --noEmit` from `frontend/` is clean.
- `npm test` (`vitest run --config vite.renderer.config.ts`) passes 2095/2108 tests; the 5 failing files are unrelated to this change (Windows Unix-domain-socket `EACCES` permission errors in `src/main/supervisor-link.test.ts`, and a missing nested `node-html-markdown` dependency under `src/landing`, a separate workspace with its own `package.json`/`node_modules` not installed by the root `frontend/` `npm ci`) — neither file imports anything touched by this change.

Addresses #2845.

### DIFF
--- a/frontend/vite.renderer.config.ts
+++ b/frontend/vite.renderer.config.ts
@@ -8,6 +8,7 @@ import { TanStackRouterVite } from "@tanstack/router-plugin/vite";
 import react from "@vitejs/plugin-react";
 import tailwindcss from "@tailwindcss/vite";
 import { DEFAULT_POSTHOG_HOST } from "./src/shared/posthog-config";
+import { DEFAULT_DAEMON_PORT, ISOLATED_DEV_DAEMON_PORT, isDevIsolationEnabled } from "./src/shared/daemon-attach";
 
 const POSTHOG_ORIGINS = (() => {
 	const configured = process.env.VITE_AO_POSTHOG_HOST?.trim() || DEFAULT_POSTHOG_HOST;
@@ -58,6 +59,18 @@ const CONTENT_SECURITY_POLICY = [
 	"frame-src 'none'",
 ].join("; ");
 
+// The dev daemon's port, derived the same way main.ts derives it: the
+// installed app's default port (3001) unless ISOLATE_DEV opts into the
+// isolated dev-only port (3002). Both read isDevIsolationEnabled/
+// ISOLATED_DEV_DAEMON_PORT/DEFAULT_DAEMON_PORT from src/shared/daemon-attach
+// — a single source of truth — so this proxy target and the daemon's actual
+// bound port can't independently hardcode a value and drift apart (#2845).
+// AO_DEV_API_TARGET still overrides the whole target (host + port) for
+// pointing the proxy at a daemon running somewhere else entirely.
+const DEV_DAEMON_PROXY_TARGET =
+	process.env.AO_DEV_API_TARGET ??
+	`http://127.0.0.1:${isDevIsolationEnabled(process.env) ? ISOLATED_DEV_DAEMON_PORT : DEFAULT_DAEMON_PORT}`;
+
 const injectCspMeta: Plugin = {
 	name: "inject-csp-meta",
 	apply: "build",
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/vite.renderer.config.ts
```diff
@@ -58,6 +59,18 @@ const CONTENT_SECURITY_POLICY = [
...
+// installed app's default port (3001) unless ISOLATE_DEV opts into the
+// isolated dev-only port (3002). Both read isDevIsolationEnabled/
+// ISOLATED_DEV_DAEMON_PORT/DEFAULT_DAEMON_PORT from src/shared/daemon-attach
+// — a single source of truth — so this proxy target and the daemon's actual
+// bound port can't independently hardcode a value and drift apart (#2845).
+// AO_DEV_API_TARGET still overrides the whole target (host + port) for
+// pointing the proxy at a daemon running somewhere else entirely.
+const DEV_DAEMON_PROXY_TARGET =
+	process.env.AO_DEV_API_TARGET ??
+	`http://127.0.0.1:${isDevIsolationEnabled(process.env) ? ISOLATED_DEV_DAEMON_PORT : DEFAULT_DAEMON_PORT}`;
```
whoisasx, line 72: [P2] This proxy target ignores an explicit `AO_PORT`. With `AO_PORT=4100`, Electron preserves the override and the daemon binds to 4100, while Vite still proxies `/api` and `/mux` to 3001 or 3002, recreating the `ECONNREFUSED` failure this PR is meant to prevent. Export and reuse one resolver that honors `AO_PORT`, and test the production resolver in both isolation modes instead of duplicating the current formula in the test.

whoisasx (changes requested):
The default 3001 path and opt-in 3002 isolation path are now aligned, but explicit `AO_PORT` still recreates the original Electron/Vite proxy mismatch. Please centralize the complete port-resolution precedence and test the actual resolver. Also correct the inaccurate provenance statement regarding #2852: the closed PR and its complete patch are visible, and that implementation already handles `AO_PORT`.

i-trytoohard:

Rishet11:
@krishrathi1

PR #3774, opened 2026-08-09
feat: add Prime Agent native restore and model discovery

nikhilachale:
## What

Add complete Prime Agent lifecycle support in AO:

- append AO worker and orchestrator system prompts on launch and restore
- discover provider-qualified models through prime-agent model list
- persist transcripts in the AO data directory and capture native session IDs
- restore terminated sessions through prime-agent --resume
- safely stop matching resident Prime workers before destructive worktree teardown
- preserve Prime Agent as the frontend session provider so top-bar tabs and session surfaces render the Prime logo instead of falling back to Codex

## Why

Prime Agent previously launched with session persistence disabled. AO could not restore terminated Prime conversations, discover configured models, or safely enable persistence because detached Prime workers could remain active after terminal teardown.

The frontend also normalized unknown providers to Codex; without Prime Agent in that allowlist, Prime sessions displayed the Codex logo in tab chrome despite using the Prime adapter.

## How

Fresh launches and restores share one command builder that supplies the AO-owned session directory, managed activity extension, role-specific appended system prompt, and selected model. The extension reports the native Prime session ID through the existing hook pipeline.

A narrow optional agent lifecycle capability lets Prime query live sessions, match the exact stored transcript ID, and stop only its active owner. Failed or ambiguous termination prevents worktree removal; inactive transcripts remain resumable. Existing adapters are unchanged.

The renderer provider normalization now recognizes `prime-agent`, reusing the existing Prime Agent avatar asset wherever session provider icons are rendered.

No database, HTTP API, or generated frontend schema changes are required.

## Testing

- GOCACHE=/private/tmp/go-build-prime-agent-final go test -p 1 ./...
- focused Prime Agent, model catalog, session manager, and CLI tests
- npm run frontend:typecheck
- git diff --check main...HEAD
- npm test -- --run src/renderer/types/workspace.test.ts
- npm run typecheck

The default fully parallel backend test fan-out exposed pre-existing short-deadline flakes under local machine load; the complete suite passes with serialized Go package execution.

## Checklist

- [x] Branched from main


Vaibhaav-Tiwari (changes requested):
1. Killing/saving a session could get stuck for any agent, not just Prime Agent
  The PR adds a step: before shutting down a session, check if its agent needs to be told "stop your native process." That check now runs for every agent type (Claude Code, Cursor, Codex, etc.), not just Prime Agent. If that check ever fails to find the agent (e.g., an old session referencing an agent that got renamed or removed), the entire shutdown now aborts the process and workspace are left running, when before this change they'd have been cleaned up fine. And there's no test covering that failure case, so nobody would notice until it happens in the real world.

  2. A silent bug could make session-resume secretly stop working Prime Agent reports "a session started" back to AO by calling one method on a context object. If that method isn't there when called (e.g., an older Prime Agent version), the code fails silently and it doesn't just lose the new "resume" data, it loses the entire "session started" signal, with zero errors logged. Since resuming sessions is the main point of this PR, if that assumption is ever wrong, restore quietly breaks for everyone and it'd be hard to diagnose.

  3. The PR includes files it shouldn't
  It accidentally commits two internal planning documents (the AI's own implementation notes) into the pull request. You've flagged this before as somethin you don't want showing up in PRs they should be deleted before merging.

nikhilachale:
https://github.com/user-attachments/assets/8fb6fc47-4d01-484a-ab37-f5b40e4c096f

backend/internal/session_manager/manager.go
```diff
@@ -1753,6 +1789,9 @@ func (m *Manager) saveAndTeardownOne(ctx context.Context, rec domain.SessionReco
 	if err := m.teardownReviewerTerminal(ctx, rec.ID); err != nil {
 		return fmt.Errorf("save %s: teardown reviewer: %w", rec.ID, err)
 	}
+	if err := m.terminateNativeSession(ctx, rec); err != nil {
```
whoisasx: [P1] The persistent Prime worker is stopped only after `StashUncommitted` and the restore marker have already captured the worktree. Because Prime can continue writing in the background, changes made between that snapshot and `ForceDestroy` are permanently lost. Please quiesce the native worker before the first stash in shutdown, replacement, and workspace-project teardown paths, and add a test that simulates a write between capture and termination.

backend/internal/adapters/agent/primeagent/primeagent.go
```diff
@@ -74,29 +76,39 @@ func (p *Plugin) GetLaunchCommand(ctx context.Context, cfg ports.LaunchConfig) (
...
-	if systemPrompt != "" {
-		cmd = append(cmd, "--append-system-prompt", systemPrompt)
+	extensionPath, err := extensionPath(dataDir)
+	if err != nil {
+		return nil, err
 	}
-	if model := strings.TrimSpace(cfg.Config.Model); model != "" {
-		cmd = append(cmd, "--model", model)
+
+	cmd := []string{binary, "--session-dir", sessions, "--extension", extensionPath}
```
whoisasx, line 115: [P1] `--session-dir` relocates transcripts, but enabling Prime's persistent mode also activates daemon/config state that defaults to `~/.prime/agent` because `PRIME_AGENT_CODING_AGENT_DIR` remains unset. This violates AO's requirement that all managed app state live under `AO_DATA_DIR` and can couple AO sessions to the user's independent Prime daemon. Please add an AO-owned Prime data directory through `AugmentRuntimeEnv` and ensure both runtime launches and lifecycle commands use the same isolated environment.

backend/internal/adapters/agent/primeagent/lifecycle.go
```diff
@@ -0,0 +1,102 @@
...
+// transcript. An inactive transcript is already safe and remains resumable.
+func (p *Plugin) TerminateNativeSession(ctx context.Context, session ports.SessionRef) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	nativeID := strings.TrimSpace(session.Metadata[ports.MetadataKeyAgentSessionID])
+	if nativeID == "" {
+		return nil
+	}
+	output, err := p.executePrimeCommand(ctx, session.WorkspacePath, "list", "--json")
```
whoisasx: [P2] `prime-agent list` and the subsequent `stop` use the session worktree as `cmd.Dir`. If the worktree was manually removed or became inaccessible, process startup fails before Prime can be contacted, and AO aborts kill/replacement before runtime teardown or terminal marking. These management commands do not require the target repository as their working directory; please run them from a stable AO-owned directory and cover stale-worktree cleanup.

whoisasx (changes requested):
The three findings from the previous review are addressed at the current head: unknown or retired harnesses no longer block cleanup, missing `getSessionId()` no longer suppresses `session-start`, and the internal planning documents were removed. Focused Prime, model-catalog, and session-manager tests pass, and all exact-head CI checks are green. The deeper lifecycle pass found three additional issues below, two of which can lose work or violate AO's state boundary. The branch also currently conflicts with `main`; please resolve that after addressing these findings.

backend/internal/adapters/agent/primeagent/primeagent.go
```diff
@@ -17,23 +18,33 @@ import (
...
 func New() *Plugin {
 	return &Plugin{}
 }
 
+// AugmentRuntimeEnv keeps Prime's persistent daemon/config state inside AO.
+func (p *Plugin) AugmentRuntimeEnv(env map[string]string, dataDir string) {
+	if strings.TrimSpace(dataDir) == "" {
+		return
+	}
+	env["PRIME_AGENT_CODING_AGENT_DIR"] = filepath.Join(dataDir, "agent-runtime", adapterID, "prime-agent")
```
whoisasx: [P1] `PRIME_AGENT_CODING_AGENT_DIR` is Prime's complete user configuration root, not only its daemon-state directory: it contains `auth.json`, settings, custom models, extensions, and skills. Pointing runtime launches at this fresh AO-owned path without seeding the user's existing configuration makes previously authenticated Prime users launch unauthenticated, while model discovery still runs against the normal user directory and can advertise models that the launched session cannot access. Please either seed/synchronize the required user config and credentials into this directory (as the Kimi adapter does) and make discovery use the same effective environment, or isolate the resident daemon through a narrower mechanism that preserves the user's config.

backend/internal/session_manager/manager.go
```diff
@@ -1386,6 +1392,9 @@ func (m *Manager) RetireForReplacement(ctx context.Context, id domain.SessionID)
 	if release != nil {
 		defer release()
 	}
+	if err := m.terminateNativeSession(ctx, rec); err != nil {
```
whoisasx, line 1430: [P2] This termination is repeated for workspace projects: after this call succeeds, `retireWorkspaceProjectForReplacement` calls `terminateNativeSession` again, and the save path has the same duplication between `saveAndTeardownOne` and `saveAndTeardownWorkspaceProject`. The redundant second `prime-agent list` can turn an already-successful stop into a failed teardown if that probe errors. Please keep termination at one ownership level and add workspace-project tests asserting exactly one native termination call for both replacement and save/shutdown flows.

whoisasx (changes requested):
Re-reviewed the latest head. The earlier unknown-harness cleanup, missing session getter, planning-document, teardown-order, and stale-worktree findings have been addressed, and all exact-head CI checks are green. Two implementation issues remain: Prime's new AO-owned config root drops existing user authentication/configuration, and workspace-project teardown invokes native termination twice. The branch also conflicts with current `main` in `backend/internal/session_manager/manager.go`; please resolve it by retaining both attachment preservation and native quiescing, with the native worker stopped before attachment/worktree capture so late writes cannot be lost. Requesting changes until these items are addressed.

whoisasx (approved):

whoisasx (approved):

Vaibhaav-Tiwari (approved):

Issue #2598, opened 2026-07-11
bug(projects): project deletion should tolerate stale legacy local state


AgentWrapper:
## Bug
Deleting the local `reverbcode` project from the Electron sidebar fails with a modal that only says:

`Internal server error (INTERNAL_ERROR)`

![AO project delete internal error screenshot](https://raw.githubusercontent.com/AgentWrapper/agent-orchestrator/issue-assets-project-delete-internal-error/.issue-assets/project-delete-internal-error/error_ss.png)

I reproduced the same failure directly against the Go daemon on port `3001`:

```text
DELETE /api/v1/projects/reverbcode
HTTP/1.1 500 Internal Server Error
{"error":"internal","code":"INTERNAL_ERROR","message":"Internal server error","requestId":"Prateeks-MacBook-Air.local/5r8LMiNn8c-000252"}
```

The project remains registered after the failed delete.

**Source:** local Codex chat report with screenshot, 2026-07-10 America/Los_Angeles
**Reported by:** local user
**Analyzed against:** `5b0c4f5289c654b51ede4a4a927cced0013808de`
**AO build:** installed Electron app `0.10.3-nightly.202607091418`
**Daemon:** `~/.ao/running.json` reports app-owned daemon on `127.0.0.1:3001`
**Confidence:** Medium-high: exact endpoint reproduced and code path traced; raw inner error is hidden by the API envelope and was not available in `~/.ao/daemon.log`. The important invariant is delete-path robustness, not the historical cause of the stale state.
**Priority:** high: core project management action is blocked and the UI gives no actionable recovery path.

## Focus / Product Requirement
Do **not** treat the stale or inconsistent local state as the main bug. This project was created by a much older AO build, so the exact path that produced the bad state may no longer be relevant.

The bug to fix is the deletion invariant: **local project removal must be robust to stale, legacy, partially missing, or otherwise inconsistent AO-managed state**. If the user asks to remove a local project registration, AO should work around bad session/worktree/runtime records and complete the unregister flow wherever it can do so without deleting unpreserved user work.

Only genuinely user-data-protecting cases should block removal, and those should return an explicit actionable response, not `INTERNAL_ERROR`. Examples:

- stale runtime handle: ignore or best-effort destroy;
- missing/non-git/prunable worktree metadata: move aside, mark unavailable, or skip with a structured result;
[...]

PR #3775, opened 2026-08-09
fix(session): tolerate legacy/corrupt local state during project teardown

krishrathi1:
## Root cause

Project deletion (`DELETE /api/v1/projects/{id}`) goes through `project.Service.Remove` -> `session.Service.TeardownProject` -> `session.Service.Kill` per live session -> `session_manager.Manager.Kill`.

`TeardownProject` looped every non-terminated session and returned on the **first** raw error from `Kill`, and separately ignored `Kill`'s `freed` boolean entirely. `Manager.Kill` already special-cases a dirty, uncommitted worktree (`ports.ErrWorkspaceDirty`) by preserving it and returning `freed=false, err=nil` — a benign outcome. But **any other** teardown failure (a stale runtime handle, a missing or non-git worktree path, a prunable git worktree registration left behind by `git worktree prune`, an already-removed filesystem path, or a Windows sharing-violation on `os.RemoveAll` right after a runtime process is killed) comes back as a raw, unwrapped Go error. `toAPIError` only recognizes a small set of named sentinels; anything else passes through unchanged, and `envelope.WriteError` collapses any non-`*apierr.Error` into an opaque `500 INTERNAL_ERROR`. The project stays registered with no recovery path shown to the user.

Full trace and two independent reproductions (idle project, and a project with an actively-running session on Windows) are in #2598 and its comment.

## Fix

All changes are in `backend/internal/service/session/service.go`, scoped to `TeardownProject`:

1. **Tolerate legacy/corrupt local state (issue fix point 1).** `TeardownProject` no longer short-circuits on a session's `Kill` error. Every non-terminated session gets a `Kill` attempt; an error from any one of them is logged server-side with its original cause and otherwise ignored, so it cannot by itself block unregistering the project.
3. **Aggregate outcomes instead of first-error-wins (fix point 3).** The Kill loop now runs to completion across every session before `TeardownProject` decides success/failure/conflict, instead of aborting immediately.
2 & 6. **Typed `PROJECT_REMOVE_BLOCKED` conflict + thread `freed` through (fix points 2 and 6).** `Kill`'s previously-discarded `freed=false` result (dirty worktree genuinely preserved, not force-deleted) is now collected across every affected session. If any exist after the loop, `TeardownProject` returns a new `apierr.Conflict("PROJECT_REMOVE_BLOCKED", ...)` with structured `details.blockedSessions` (`sessionId` + `workspacePath` per entry) instead of silently succeeding or 500ing.
4. **409, not 500, for the genuinely-blocking case (fix point 4) — design choice.** The issue leaves "block with 409" vs "archive and preserve" as an explicit implementer choice. I chose **block with 409**: a project with only stale/legacy/corrupt teardown failures (no real user data at risk) now unregisters successfully, but a project with a live, dirty, unpreserved-by-normal-means worktree blocks removal with the new structured error rather than silently archiving the project while the worktree becomes orphaned and undiscoverable (the project entry that would have pointed back at it is gone). This matches `AGENTS.md`'s existing hard rule ("do not force-delete dirty registered worktrees") in spirit — removal of the *registration* stays blocked, not just the worktree deletion — and keeps the recovery path concrete (resolve/commit/discard the session's changes, or let it finish, then retry).
5. **Keep the raw cause in logs only (fix point 5).** Every tolerated failure is logged via `slog` (`s.logger.Warn`, guarded for nil in tests) with the real underlying error; the wire response is always either success or the structured conflict — never a raw 500 sourced from this path.
7. **Regression coverage (fix point 7).** Added to `backend/internal/service/session/service_test.go`, extending the existing `fakeCommander` with per-session `Kill` overrides (`killErrs`, `killNotFreed`) following the existing fake-injection pattern in that file:
   - `TestTeardownProjectBlocksOnDirtyPreservedWorktree` — scenario (a): a live dirty session worktree — asserts a `*apierr.Error` with `Kind == KindConflict` and `Code == "PROJECT_REMOVE_BLOCKED"`, with `details.blockedSessions` naming the session and its path.
   - `TestTeardownProjectToleratesStaleWorktreeRegistration` — scenario (b): a stale/prunable worktree registration, using an error wrapped around `ports.ErrWorkspaceStale` (the sentinel `gitworktree.Workspace.Destroy` actually returns for this shape) — asserts `TeardownProject` succeeds and `Cleanup` still runs.
   - `TestTeardownProjectToleratesUnrecognizedKillFailure` (repurposed from the old `TestTeardownProjectStopsOnKillError`, which asserted the buggy first-error-wins behavior) — scenario (c): a plain, unwrapped error not in `toAPIError`'s sentinel allowlist — asserts `TeardownProject` succeeds, the other session still gets killed, and `Cleanup` still runs.
   
   Also added `TestManager_RemoveSurfacesProjectRemoveBlockedConflict` in `backend/internal/service/project/service_test.go`, verifying the typed conflict survives `project.Service.Remove`'s pass-through unchanged and the project stays registered (not silently archived) when blocked.

### Not implemented

No DTO/OpenAPI changes: `APIError.Details` (`backend/internal/httpd/envelope/envelope.go`) is already a generic `map[string]any` on every endpoint's error schema, so the new code/details surface on the wire with no spec regeneration needed. No frontend changes: `frontend/src/renderer/routes/_shell.tsx`'s `removeProject` already renders whatever structured `message`/`code` the API returns generically, so this fix alone turns today's opaque "Internal server error (INTERNAL_ERROR)" into the real, actionable cause without any renderer changes.

## Verification

- Patch-based fail-then-pass (no `git stash`, to avoid colliding with other concurrent worktrees on this shared clone): isolated the `service.go` implementation diff into a patch file, reverse-applied it while keeping the new tests, confirmed all three new regression tests fail against the unpatched code (`TeardownProjectToleratesUnrecognizedKillFailure` and `TeardownProjectToleratesStaleWorktreeRegistration` propagate the raw error instead of tolerating it; `TeardownProjectBlocksOnDirtyPreservedWorktree` returns `nil` instead of the typed conflict), then re-applied the patch and confirmed all three pass.
- `cd backend && go build ./...` and `go vet ./...`: clean.
- Full test suite for every touched package: `go test ./internal/service/session/...` (all pass), `go test ./internal/service/project/...` (all pass except one pre-existing, unrelated Windows path-separator failure in `TestManager_InitializeRepositoryRecovery`, confirmed to already fail identically on a pristine `upstream/main` checkout with no code changes at all), `go test ./internal/httpd/...` (all pass, covers the envelope/controller error-mapping boundary).
- `go test -race` was not runnable in this sandbox (`CGO_ENABLED=0`, no `gcc` on PATH) — unrelated to this change, an environment limitation.

Addresses #2598.

### DIFF
--- a/backend/internal/service/project/service_test.go
+++ b/backend/internal/service/project/service_test.go
@@ -415,6 +415,37 @@ func TestManager_RemoveDoesNotArchiveWhenTeardownFails(t *testing.T) {
 	}
 }
 
+// TestManager_RemoveSurfacesProjectRemoveBlockedConflict verifies the typed
+// PROJECT_REMOVE_BLOCKED conflict that TeardownProject returns for a
+// genuinely-blocking dirty worktree (issue #2598, fix points 2/4) survives
+// Remove()'s pass-through unchanged, and that the project stays registered
+// rather than being silently archived out from under preserved work.
+func TestManager_RemoveSurfacesProjectRemoveBlockedConflict(t *testing.T) {
+	ctx := context.Background()
+	store, err := sqlitetest.Open(t.TempDir())
+	if err != nil {
+		t.Fatalf("open store: %v", err)
+	}
+	t.Cleanup(func() { _ = store.Close() })
+	blocked := apierr.Conflict("PROJECT_REMOVE_BLOCKED", "Project has sessions with uncommitted worktree changes", map[string]any{
+		"blockedSessions": []map[string]string{{"sessionId": "ao-1", "workspacePath": "/managed/ao/ao-1"}},
+	})
+	m := project.NewWithDeps(project.Deps{Store: store, Sessions: &fakeProjectTeardowner{err: blocked}})
+
+	if _, err := m.Add(ctx, project.AddInput{Path: gitRepo(t), ProjectID: ptr("ao")}); err != nil {
+		t.Fatalf("Add: %v", err)
+	}
+	_, err = m.Remove(ctx, "ao")
+	wantCode(t, err, "PROJECT_REMOVE_BLOCKED")
+	var apiErr *apierr.Error
+	if !errors.As(err, &apiErr) || apiErr.Kind != apierr.KindConflict {
+		t.Fatalf("Remove err = %v, want a KindConflict *apierr.Error", err)
+	}
+	if got, err := m.Get(ctx, "ao"); err != nil || got.Project == nil || got.Project.ID != "ao" {
+		t.Fatalf("project after blocked remove = %#v, %v; want still active", got, err)
+	}
+}
+
 func TestManager_DefaultsWhenUnconfigured(t *testing.T) {
 	ctx := context.Background()
 	m := newManager(t)
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -733,24 +733,92 @@ func (s *Service) Cleanup(ctx context.Context, project domain.ProjectID) (Cleanu
 	return out, nil
 }
 
+// preservedWorktree names one session whose worktree Kill genuinely had to
+// preserve (uncommitted changes) rather than destroy, during project teardown.
+type preservedWorktree struct {
+	SessionID     domain.SessionID
+	WorkspacePath string
+}
+
 // TeardownProject stops every live session in a project, then asks the session
-// manager to reclaim terminal workspaces. Dirty worktrees are preserved by Kill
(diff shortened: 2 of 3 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

Rishet11:
@krishrathi1

PR #3776, opened 2026-08-09
fix: drop hover from Projects section label

Annieeeee11:
## Summary
- Drop the full-row hover from the Projects header so it reads as a static label; only the `+` button is interactive
- Remove the folder and pin icons from the Projects and Pinned section headers
- Leave collapsible Pinned disclosure behavior unchanged

## Test plan
- [ ] Hover Projects label — no background fill
- [ ] Hover/click `+` — still highlights and opens create-project
- [ ] Projects and Pinned headers show text only (no folder/pin icon)
- [ ] Pinned (when present) still toggles open/closed on click
- [ ] Individual project rows still show their folder icons

## Before
image
image

## After
image


ronishrohan (commented):
Small change, nothing much to review, lets gtm

ronishrohan (approved):

PR #3777, opened 2026-08-09
feat: add Amp Pi and Auggie activity hooks

nikhilachale:
## What

Add native live-activity reporting for Amp, Pi, and Auggie so their TUI sessions can show Working, Idle, Needs you, Exited, and No signal honestly on the Kanban board.

## Why

These three adapters were treated as hookless even though their current CLIs expose lifecycle integration surfaces. Their cards could therefore remain misleadingly Idle while work was running or while the agent needed attention.

## How

- Extend Amp's existing managed TypeScript plugin to report session, agent, and thread-state changes, including approvals and errors.
- Install and explicitly load a managed Pi extension that reports session/agent lifecycle events and supports both Pi 0.80.x `agent_end` and newer `agent_settled` completion events.
- Merge Auggie's native hooks into `.augment/settings.local.json` and install executable AO-owned wrappers for its supported command hooks.
- Capture Amp thread IDs, Pi session IDs, and Auggie conversation IDs for native restore.
- Register all three adapter-local derivers in the shared activity dispatcher so missing callbacks degrade to No signal rather than confident Idle.
- Preserve user-owned plugins, hooks, and unrelated settings; every callback remains bounded and best-effort.

Intentional omission: no frontend or API vocabulary changes are included. The integrations use AO's existing derived activity/status model.

## Testing

- Test-first adapter coverage for native event mapping, malformed/unknown payloads, managed-file installation, foreign-file protection, idempotency, user-setting preservation, and restore metadata.
- `npm run lint`
- Focused hook CLI, session manager, registry, dispatcher, and session-status suites.

## Checklist


backend/internal/adapters/agent/pi/hooks.go
```diff
@@ -0,0 +1,113 @@
...
+  // compact, or queue follow-up work after it; a subsequent start immediately
+  // reactivates AO, while agent_settled below confirms the final idle state.
+  pi.on("agent_end", async (_event, ctx) => {
+    callHookSync("stop", { session_id: sessionID(ctx) });
+  });
+  pi.on("agent_settled", async (_event, ctx) => {
+    callHookSync("stop", { session_id: sessionID(ctx) });
+  });
+  pi.on("session_shutdown", async (event, ctx) => {
+    callHookSync("session-end", { session_id: sessionID(ctx), reason: event.reason });
```
whoisasx: [P1] Pi emits `session_shutdown` not only for `quit`, but also for `reload`, `new`, `resume`, and `fork`. Those latter reasons replace the active session inside the same live TUI. Emitting `session-end` for all of them marks the AO session `exited`, and the following idle `session-start` cannot clear that sticky state until another prompt is submitted. Please emit `session-end` only for a real `quit` and add coverage for every shutdown reason.

backend/internal/adapters/agent/amp/hooks.go
```diff
@@ -92,6 +92,40 @@ func ampSystemPromptPluginSource(inline, file string) string {
...
+	b.WriteString("  } catch (error) {\n")
+	b.WriteString("    reportHookFailure(amp, hookName, error instanceof Error ? error.message : String(error));\n")
+	b.WriteString("  }\n")
+	b.WriteString("}\n\n")
+	b.WriteString("function reportThreadState(amp: any, sessionID: string, state: string) {\n")
+	b.WriteString("  callHookSync(amp, \"thread-state\", { session_id: sessionID, state });\n")
+	b.WriteString("}\n\n")
+	b.WriteString("function observeThread(amp: any, thread: any) {\n")
+	b.WriteString("  if (threadSubscriptions.has(thread.id)) return;\n")
+	b.WriteString("  const subscription = thread.state.subscribe((state: string) => reportThreadState(amp, thread.id, state));\n")
```
whoisasx, line 131: [P1] Amp plugins are long-lived processes that may manage multiple threads concurrently. This keeps every observed thread subscribed indefinitely, while every callback is routed to the same AO session. A background or previously active thread can therefore overwrite both the card state and the persisted `AgentSessionID`, making restore open the wrong thread. Please track Amp's active thread, ignore or unsubscribe superseded subscriptions, and test thread switching plus concurrent background activity.

backend/internal/adapters/agent/pi/hooks.go
```diff
@@ -0,0 +1,113 @@
...
+    callHookSync("session-start", { session_id: sessionID(ctx) });
+  });
+  pi.on("before_agent_start", async (event, ctx) => {
+    callHookSync("user-prompt-submit", { session_id: sessionID(ctx), prompt: event.prompt });
+  });
+  // agent_end is the completion event in Pi 0.80.x. Newer releases may retry,
+  // compact, or queue follow-up work after it; a subsequent start immediately
+  // reactivates AO, while agent_settled below confirms the final idle state.
+  pi.on("agent_end", async (_event, ctx) => {
+    callHookSync("stop", { session_id: sessionID(ctx) });
```
whoisasx: [P2] On Pi versions that expose `agent_settled`, `agent_end` is explicitly not the final idle boundary: automatic retry, compaction, or queued continuation may still follow. Reporting `stop` here creates a false idle transition before the run has settled. Please use `agent_settled` as the final idle signal on supported versions while keeping an explicit compatibility path for 0.80.x, and cover the retry/compaction ordering.

backend/internal/adapters/agent/amp/activity.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// DeriveActivityState maps the AO sub-command emitted by the managed Amp
+// plugin onto AO's normalized activity state. session-start is metadata-only:
+// opening an existing thread is not proof that its agent is currently running.
+func DeriveActivityState(event string, payload []byte) (domain.ActivityState, bool) {
+	switch event {
+	case "user-prompt-submit":
+		return domain.ActivityActive, true
+	case "stop":
```
whoisasx, line 16: [P2] The plugin forwards `agent.end` status (`done`, `error`, or `cancelled`), but this maps every outcome to idle. An error-ending turn can arrive after `thread-state:error` and overwrite the needs-attention signal. Please parse the status so an error remains visible, and add an ordering test for `thread-state:error` followed by `agent.end(error)`.

whoisasx (changes requested):
The integration direction is sound and the current CI is green, but the lifecycle handling still has two correctness blockers: Pi treats in-process session replacement as process exit, and Amp allows background threads to overwrite the AO session's activity and restore identity. I have also called out two lower-severity state-transition issues below. Please address these before merging.

nikhilachale:
@whoisasx  
Implemented the requested lifecycle fixes:

- Pi now emits `session-end` only for an actual `quit`; reload, new, resume, and fork no longer terminate the AO session.
- Pi uses `agent_settled` as the final idle signal, with a compatibility fallback for older Pi versions.
- Amp now tracks only the active thread and unsubscribes previous thread listeners.
- Background or stale Amp threads can no longer overwrite activity state or restore identity.
- Amp error completions remain visible as `waiting_input` instead of being incorrectly marked idle.
- Added regression tests covering Pi shutdown reasons, settled handling, Amp thread switching, and error ordering.

backend/internal/adapters/agent/amp/hooks.go
```diff
@@ -106,11 +147,20 @@ func ampSystemPromptPluginSource(inline, file string) string {
 	b.WriteString("  return inlineSystemPrompt.trim();\n")
 	b.WriteString("}\n\n")
 	b.WriteString("export default function (amp: PluginAPI) {\n")
-	b.WriteString("  amp.on(\"agent.start\", async () => {\n")
+	b.WriteString("  amp.on(\"session.start\", async (event, ctx) => {\n")
+	b.WriteString("    observeThread(amp, ctx.thread);\n")
+	b.WriteString("    callHookSync(amp, \"session-start\", { session_id: event.thread.id });\n")
+	b.WriteString("  });\n")
+	b.WriteString("  amp.on(\"agent.start\", async (event, ctx) => {\n")
+	b.WriteString("    observeThread(amp, ctx.thread);\n")
```
whoisasx, line 156: [P1] This still infers the active thread from whichever lifecycle callback arrives last. Amp plugins may run multiple threads concurrently, and a background `agent.start` here calls `observeThread`, unsubscribes the foreground thread, reports the background session ID, and can overwrite both AO activity and the restore identity. Please make `amp.activeThread` the authority and ignore lifecycle/state events from non-active threads, with a behavioral test covering foreground thread A plus concurrent background thread B.

backend/internal/adapters/agent/pi/hooks.go
```diff
@@ -0,0 +1,122 @@
...
+  });
+  pi.on("before_agent_start", async (event, ctx) => {
+    callHookSync("user-prompt-submit", { session_id: sessionID(ctx), prompt: event.prompt });
+  });
+  // agent_end is the completion event in Pi 0.80.x. Newer releases may retry,
+  // compact, or queue follow-up work after it; a subsequent start immediately
+  // reactivates AO, while agent_settled below confirms the final idle state.
+  pi.on("agent_end", async (_event, ctx) => {
+	    if (settledSupported) return;
+	    legacyAgentEndTimer = setTimeout(() => {
```
whoisasx: [P2] A zero-delay timer is not a safe legacy detector. On current Pi, `agent_end` is emitted before retry, compaction, and queued-continuation decisions; this timer can run before those continuation paths and publish the same false idle transition that `agent_settled` is intended to prevent. `settledSupported` is also false for the first turn after each extension load. Please establish an explicit host-version/capability boundary, or otherwise ensure the fallback cannot fire before possible continuation work, and test the actual `agent_end -> continuation -> agent_settled` event order rather than generated-source substrings.

backend/internal/adapters/agent/auggie/activity.go
```diff
@@ -0,0 +1,32 @@
...
+// normal completion and interruption leave the session idle.
+func DeriveActivityState(event string, payload []byte) (domain.ActivityState, bool) {
+	switch event {
+	case "session-start", "pre-tool-use", "post-tool-use":
+		return domain.ActivityActive, true
+	case "stop":
+		var native struct {
+			Cause string `json:"agent_stop_cause"`
+		}
+		_ = json.Unmarshal(payload, &native)
```
whoisasx: [P2] Ignoring the decode error and treating every unrecognized cause as idle makes malformed or truncated callbacks, missing causes, and future stop causes look like successful completion. Since `agent_stop_cause` distinguishes normal completion from `error` and `max_iterations`, please map the four documented values explicitly and return `ok=false` for invalid, missing, or unknown causes. Add malformed and unknown-cause coverage as well.

whoisasx (changes requested):
The latest patch correctly addresses the earlier Pi shutdown issue and Amp error-status ordering. However, three activity-reporting correctness issues remain: background Amp threads can still replace the foreground subscription and persisted restore identity; Pi's zero-delay legacy fallback can report idle before retry, compaction, or queued continuation work begins; and malformed or unknown Auggie stop payloads are classified as successful idle completion. Please address these lifecycle and payload-classification cases before merging. The relevant adapter tests pass on the PR head and after merging the current main branch, and all GitHub checks are green.

backend/internal/adapters/agent/amp/hooks.go
```diff
@@ -92,6 +92,48 @@ func ampSystemPromptPluginSource(inline, file string) string {
...
+	b.WriteString("    if (!result.success) {\n")
+	b.WriteString("      const detail = result.stderr ? new TextDecoder().decode(result.stderr).trim() : `exit ${result.exitCode}`;\n")
+	b.WriteString("      reportHookFailure(amp, hookName, detail);\n")
+	b.WriteString("    }\n")
+	b.WriteString("  } catch (error) {\n")
+	b.WriteString("    reportHookFailure(amp, hookName, error instanceof Error ? error.message : String(error));\n")
+	b.WriteString("  }\n")
+	b.WriteString("}\n\n")
+	b.WriteString("function reportThreadState(amp: any, sessionID: string, state: string) {\n")
+	b.WriteString("  if (amp.activeThread?.id !== sessionID || sessionID !== activeThreadID) return;\n")
```
whoisasx: [P1] `activeThread` is an observable; the documented current thread is `amp.activeThread.current`, so `amp.activeThread?.id` is always undefined. As written, `observeThread` returns without subscribing, `session-start` never reports, `agent.start` returns before both activity reporting and the existing system-prompt injection, and `agent.end` never reports. Please use `amp.activeThread.current?.id` consistently and add an executable behavior test instead of asserting that this source string exists.

backend/internal/adapters/agent/pi/hooks.go
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+function sessionID(ctx: any): string {
+  return ctx.sessionManager.getSessionId() ?? "";
+}
+
+export default function (pi: ExtensionAPI) {
+  // Pi 0.81.0 and newer expose agent_settled as the final idle boundary.
+  const version = typeof (pi as any).version === "string" ? (pi as any).version : "";
```
whoisasx: [P2] Pi's `ExtensionAPI` does not expose a `version` property, so this resolves to an empty string and `settledSupported` remains false on every current Pi release. That means AO still posts the premature `agent_end` stop and then another stop at `agent_settled`. The version boundary is also inaccurate because `agent_settled` shipped in Pi 0.80.6, not 0.81.0. Please detect compatibility through a real supported boundary, such as probing the installed binary before generating the extension, and cover actual event ordering in a behavioral test.

whoisasx (changes requested):
Thanks for addressing the previous feedback. The Auggie malformed/unknown stop handling is now correct, and the earlier Pi shutdown and Amp error-completion fixes remain sound. Two host-contract blockers remain in the latest commit: Amp reads the active-thread observable incorrectly, which suppresses the entire Amp integration and its existing system-prompt injection; and Pi probes a version property that ExtensionAPI does not expose, so supported Pi versions still take the unsafe agent_end fallback. Please address these two points before merging. All GitHub checks are green, and the relevant adapter tests pass both on the PR head and merged with current main, but those tests only inspect generated TypeScript text and therefore do not catch these runtime contract errors.

whoisasx:
@nikhilachale  add some screen recording for both of the agents showing interchange between the different states of activity.

whoisasx (approved):

PR #3778, opened 2026-08-09
feat: add multi-agent token usage observability

nikhilachale:
## What

Add file-backed token usage observability for Copilot, Kimi, Pi, and Qwen alongside the existing Codex and Claude support.

- Parse each provider's native token metrics into the existing usage model.
- Discover, validate, and reconcile provider-owned usage artifacts durably.
- Preserve per-metric availability when aggregating mixed harnesses.
- Keep token usage visible only when Developer Mode is enabled.

## Why

AO already exposes token usage for Codex and Claude, but sessions using other supported agents could not report reliable input, cache, output, and reasoning totals. This extends the same collector and summary pipeline without introducing a separate accounting path.

## How

- Add source profiles and parsers for Copilot shutdown metrics, Kimi wire events, Pi session events, and Qwen monthly usage.
- Extend the existing usage binding/source constraints through migration 0085.
- Fence source attribution to the expected provider root and native session.
- Retry artifacts that appear after session startup and reactivate Pi discovery when its native session ID is persisted.
- Treat a mixed-harness aggregate metric as available only when every contributor reports it.

Intentional omissions:

- Cost and pricing status are not included.
- Grok requires an OTEL ingestion path and remains a follow-up.
- Droid provides only a summary, which is insufficient for reliable token totals.

## Testing

- `npm run lint`
- `cd backend && go test ./internal/storage/sqlite -run 'Test(MigrationVersionLedger|MultiAgentUsageMigration)' -count=1`

## Checklist

### DIFF
--- a/backend/internal/observe/usage/parser_copilot.go
+++ b/backend/internal/observe/usage/parser_copilot.go
@@ -1,0 +1,125 @@
+package usage
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+type copilotTranscriptRecord struct {
+	Type string `json:"type"`
+	Data struct {
+		ModelMetrics map[string]struct {
+			Usage *struct {
+				InputTokens      int64 `json:"inputTokens"`
+				OutputTokens     int64 `json:"outputTokens"`
+				CacheReadTokens  int64 `json:"cacheReadTokens"`
+				CacheWriteTokens int64 `json:"cacheWriteTokens"`
+				ReasoningTokens  int64 `json:"reasoningTokens"`
+			} `json:"usage"`
+		} `json:"modelMetrics"`
+	} `json:"data"`
+}
+
+func parseCopilot(
+	source domain.UsageSourceContext,
+	records []jsonlRecord,
+	state *copilotParserStateV1,
+	result *parseResult,
+) {
+	for _, record := range records {
+		var native copilotTranscriptRecord
+		if err := json.Unmarshal(record.Data, &native); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		if native.Type != "session.shutdown" {
+			continue
+		}
+		for nativeModel, metrics := range native.Data.ModelMetrics {
+			model := firstNonEmpty(nativeModel)
+			if model == "" || metrics.Usage == nil {
+				recordMalformed(result)
+				continue
+			}
+			usage := metrics.Usage
+			total := copilotTokenVector{
+				InputTokens:      usage.InputTokens,
+				CacheReadTokens:  usage.CacheReadTokens,
+				CacheWriteTokens: usage.CacheWriteTokens,
+				OutputTokens:     usage.OutputTokens,
+				ReasoningTokens:  usage.ReasoningTokens,
+			}
+			if !validCopilotTotal(total) {
+				recordMalformed(result)
+				continue
+			}
+			baseline := state.Models[model]
+			if copilotCounterRegressed(total, baseline) {
+				state.Models[model] = total
+				result.Cursor.AnomalyCount++
+				result.Cursor.LastErrorCode = domain.UsageErrorNonMonotonicCumulativeUsage
+				continue
+			}
+			delta := subtractCopilotTotal(total, baseline)
+			state.Models[model] = total
+			if delta.InputTokens == 0 && delta.OutputTokens == 0 && delta.CacheWriteTokens == 0 {
+				continue
+			}
+			uncached := delta.InputTokens - delta.CacheReadTokens - delta.CacheWriteTokens
+			tokens := domain.UsageTokenMetrics{
+				InputTokens:         delta.InputTokens,
+				UncachedInputTokens: uncached,
+				CacheReadTokens:     delta.CacheReadTokens,
+				CacheWriteTokens:    delta.CacheWriteTokens,
+				OutputTokens:        delta.OutputTokens,
+				ReasoningTokens:     int64Ptr(delta.ReasoningTokens),
+			}
+			if !validTokenMetrics(tokens) {
+				recordMalformed(result)
+				continue
+			}
+			result.Events = append(result.Events, domain.ModelUsageEvent{
+				ModelID: model,
+				Tokens:  tokens,
+				SourceEventKey: stableSourceEventKey(
+					"copilot",
+					source.NativeRootID,
+					model,
+					strconv.FormatInt(total.InputTokens, 10),
+					strconv.FormatInt(total.CacheReadTokens, 10),
+					strconv.FormatInt(total.CacheWriteTokens, 10),
+					strconv.FormatInt(total.OutputTokens, 10),
+					strconv.FormatInt(total.ReasoningTokens, 10),
+				),
+			})
+		}
+	}
+}
+
+func validCopilotTotal(total copilotTokenVector) bool {
+	return total.InputTokens >= 0 && total.CacheReadTokens >= 0 && total.CacheWriteTokens >= 0 &&
+		total.OutputTokens >= 0 && total.ReasoningTokens >= 0 &&
+		total.CacheReadTokens <= total.InputTokens &&
+		total.CacheWriteTokens <= total.InputTokens-total.CacheReadTokens &&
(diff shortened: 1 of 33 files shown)


backend/internal/observe/usage/parser_qwen.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		if native.SessionID != source.NativeRootID {
+			continue
+		}
+		model := firstNonEmpty(native.Model)
+		identity := firstNonEmpty(native.ID)
+		if native.SchemaVersion != 1 || model == "" || identity == "" ||
+			native.InputTokens < 0 || native.OutputTokens < 0 || native.CachedTokens < 0 ||
+			native.ThoughtsTokens < 0 ||
+			native.CachedTokens > native.InputTokens || native.ThoughtsTokens > native.OutputTokens ||
```
whoisasx: [P1] This validation does not match Qwen's native schema. Qwen computes `totalTokens` as `inputTokens + outputTokens + thoughtsTokens`; it does not require thoughts to be a subset of `outputTokens`. Consequently, a normal reasoning record with nonzero thoughts is rejected as malformed and makes the session's usage integrity incomplete. Please normalize generated output consistently with AO's reasoning-subset model, validate against Qwen's real total semantics, and add a fixture with nonzero `thoughtsTokens`.

backend/internal/observe/usage/parser_copilot.go
```diff
@@ -0,0 +1,181 @@
...
+	for _, record := range records {
+		var native copilotTranscriptRecord
+		if err := json.Unmarshal(record.Data, &native); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		switch native.Type {
+		case "assistant.message":
+			appendCopilotAssistantMessageUsage(source, record, native, result)
+		case "session.shutdown":
```
whoisasx, line 55: [P1] A normal Copilot log can contain both per-message `assistant.message.outputTokens` and cumulative `session.shutdown.modelMetrics`. This parser emits the message output first, then emits the complete shutdown output from a zero baseline, so the same generated tokens are counted twice. Please choose one authoritative source per metric, preferably the complete shutdown rollup, or reconcile the rollup against already-emitted message usage. Add one fixture containing both event types to assert the final aggregate.

whoisasx (changes requested):
The collector architecture is consistent with the existing usage pipeline, and the focused parser, collector, and SQLite tests pass locally. Two provider parsers currently produce incorrect accounting, however, so this is not ready to merge. The branch also conflicts with current `main`: `main` already owns migration 0085, while this PR adds `0085_multi_agent_usage.sql`. Please rebase and renumber the new migration beyond the latest shipped migration, update the migration ledger/repair tests, and rerun CI after resolving the parser findings.

nikhilachale:
Implemented and pushed fixes for PR #3778:

- Corrected Qwen token validation to include reasoning tokens in `totalTokens`.
- Added Qwen coverage for nonzero `thoughtsTokens`.
- Prevented Copilot assistant-message tokens from being double-counted when shutdown rollups are present.
- Added a Copilot regression fixture covering both event types.
- Renumbered the usage migration from `0085` to `0086`.
- Updated the migration ledger.

Verified with:

```bash
go test ./internal/observe/usage
go test ./internal/storage/sqlite
```

nikhilachale:
> Addressed the review findings in `39b6e6e32`.
>
> - Refreshes the root and every registered workspace child repository before materialization.
> - Fetches with an explicit refspec so restrictive remote configurations cannot leave tracking refs stale.
> - Resolves slash-named branches such as `release/2026` against the refreshed `origin/<branch>` ref.
> - Added regression coverage for child repository refresh and slash-named branches.
>
> Focused package tests and race tests pass.

i-trytoohard:
From the NotesBot call on `2026-08-13`, this existing pr appears to be relevant.
Match confidence: `0.70` (fuzzy).

Relevant discussion excerpts:
- - and model-level usage metrics. - *Pulkit* said, "I will implement it." - *Prateek* asked about token observability and said the team currently only has it for **Codex** and **Claude**.

Source thread: https://discord.com/channels/1476302178913357958/1537520911476072618/1537520911476072618

whoisasx:
@nikhilachale add screenshots from the right sidebar showing the cost in tokens consumed for all the agents which have been implemented on this PR.

backend/internal/service/usage/collector.go
```diff
@@ -480,6 +500,12 @@ func (c *Collector) BackfillActive(ctx context.Context) error {
 			continue
 		}
 		nativeID := boundedUsageMetadata(session.Metadata.AgentSessionID)
+		if nativeID == "" && session.Harness == domain.HarnessPi {
```
whoisasx, line 507: [P1] Restricting ID-less Pi discovery to startup backfill leaves normal newly launched sessions unobserved. Pi has no hook-captured native ID, `MarkSpawned` calls `ReactivateSession`, and that path returns when both the native ID and an existing binding are absent. Because the watcher observes exact registered files rather than the Pi directory, a later-created artifact has no event path into the collector. In a focused launch-equivalent regression, even a valid Pi JSONL already present for the session workspace produced zero bindings. Please trigger bounded Pi discovery from launch/reactivation and preserve retries for the file-creation race, with coverage for both an already-present artifact and one that appears after `MarkSpawned`.

backend/internal/storage/sqlite/migrations/0095_multi_agent_usage.sql
```diff
@@ -0,0 +1,57 @@
+-- Widen the token-usage source constraints for the file-backed agent
```
whoisasx, line 1: [P1] Migration version 0086 is already shipped on current `main` as `0086_workspace_repo_default_branch.sql`, and `main` has now advanced through 0094. Keeping a second version 86 migration makes the migration history ambiguous and already conflicts with the burned-version ledger. Please rebase, renumber this migration to the next available version after current `main`, and update the ledger/migration tests accordingly.

whoisasx (changes requested):
Re-reviewed the latest head after the parser fixes. The focused usage, storage, and lifecycle packages pass, including race coverage for the parser and collector packages, but targeted two-stage regression tests still reproduce incorrect Copilot and Qwen accounting, and normal newly launched Pi sessions do not enter discovery. The two earlier parser findings are therefore only partially addressed: Copilot suppresses messages only when a shutdown appears in the same ingestion batch, while Qwen validates its native total but still excludes separate thought tokens from AO's normalized output. The branch also needs to be rebased: current `main` already ships migration 0086 and has advanced through 0094, and the PR conflicts in `backend/internal/lifecycle/manager.go` and the burned-migration ledger. No exact-head CI checks are currently reported. Requesting changes until the three correctness issues and the migration/current-main integration are addressed.

backend/internal/observe/usage/parser_qwen.go
whoisasx: Re-verified on `3be073895`. This is only partially addressed: the `totalTokens` equality now includes thoughts, but the normalized event still sets `OutputTokens` to native visible output alone while `validTokenMetrics` requires reasoning to be a subset of output. A valid Qwen record with input=100, output=2, thoughts=5, total=107 is therefore still dropped; when thoughts happen to be smaller than output, AO's compact total still undercounts them because it sums normalized input + output. Please make normalized output include thoughts using overflow-safe addition, retain thoughts as `ReasoningTokens`, and test both `thoughts > output` and aggregate equality with native `totalTokens`.

backend/internal/observe/usage/parser_copilot.go
whoisasx: Re-verified on `3be073895`. The new `hasShutdown` check only examines the records in the current ingestion batch. In the normal live sequence, an `assistant.message` batch is persisted first and its parser state does not record those provisional tokens; when `session.shutdown` arrives in a later batch, the zero shutdown baseline emits the complete total again. A focused two-pass regression emitted the same 17 output tokens from both passes. Please reconcile against persisted message usage or choose one authoritative source, and make the regression invoke the parser twice with state carried between calls.

nikhilachale:
Superseded by the focused four-PR stack:

- #4182 — Copilot plus shared usage foundations
- #4183 — Kimi
- #4184 — Pi
- #4185 — Qwen

The replacement stack is rebased onto current main and omits the unrelated agent-role-smoke files that were present in this PR diff.

PR #3779, opened 2026-08-09
fix: report Aider and Continue Kanban activity

### LINKED ISSUE #2402: bug(agent): grok/devin/continue not marked signal-capable despite installing claude-code hooks

nikhilachale:
## What

- Use Aider response-ready notifications to report `waiting_input` activity.
- Add shared none/partial/complete signal coverage classification.
- Classify Continue as complete through its Claude-compatible callback pipeline.

## Why

Aider sessions remained confidently Idle after the agent returned to user input, including model authentication failures. Continue installs Claude-compatible hooks but was classified as hookless because its callbacks use the `claude-code` token.

Fixes #2402.

## How

Aider launches with `--notifications` and an AO-managed `--notifications-command`. The activity dispatcher registers the Aider completion event as partial coverage, while Continue explicitly receives complete coverage without adding a fake `continue` runtime deriver. The `no_signal` watchdog remains limited to complete lifecycle pipelines, so Aider is not penalized for lifecycle events it cannot emit.

Intentional omission: Aider still cannot report prompt-start, tool, permission, exit, native session ID, or restore events because the CLI exposes no full lifecycle hook API.

## Testing

- `go test ./internal/adapters/agent/aider ./internal/adapters/agent/continueagent ./internal/adapters/agent/activitydispatch ./internal/cli ./internal/service/session ./internal/session_manager ./internal/daemon`
- `npm run lint`
- `git diff --check`

## Checklist


backend/internal/adapters/agent/aider/aider.go
```diff
@@ -23,7 +24,10 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
 )
 
-const adapterID = "aider"
+const (
+	adapterID                = "aider"
+	aiderNotificationCommand = "ao hooks aider notification"
```
whoisasx, line 30: [P1] Aider executes this notification command synchronously with inherited stdin, while `ao hooks` unconditionally calls `io.ReadAll(stdin)`. In an AO-managed PTY that stdin remains open, so the callback never returns and Aider cannot redraw its next prompt after the first response or confirmation notification. Please provide an Aider notification entrypoint that does not consume inherited PTY stdin, or isolate stdin in a cross-platform way, and add a test where the parent stdin remains open.

backend/internal/adapters/agent/continueagent/terminal_activity.go
```diff
@@ -0,0 +1,74 @@
...
+			return domain.ActivityActive, true
+		}
+	}
+	return "", false
+}
+
+func continueQuestionPickerAt(lines []string, idx int) bool {
+	line := strings.ToLower(lines[idx])
+	if strings.Contains(line, "ask question(") {
+		return true
```
whoisasx: [P2] A completed `Ask Question(...)` tool entry remains in Continue's transcript, so treating this historical line alone as authoritative makes the continuous observer restore `waiting_input` after the user has answered and the Stop hook has correctly reported idle. Please require evidence of the currently visible picker, such as its question/options and `Enter select` action row, and add a regression fixture containing a completed question followed by the idle composer.

backend/internal/adapters/agent/continueagent/continueagent.go
```diff
@@ -120,20 +157,17 @@ func (p *Plugin) GetPromptDeliveryStrategy(ctx context.Context, cfg ports.Launch
...
-// hook implementation or code duplication.
+// The installed commands are "ao hooks continue <evt>". Continue still fires
+// Claude-format events, but the Continue token makes AO store activity under the
+// actual harness instead of under claude-code.
 func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfig) error {
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	return (&claudecode.Plugin{}).GetAgentHooks(ctx, cfg)
+	return continueHooks.Install(ctx, cfg.WorkspacePath)
```
whoisasx, line 171: [P2] Existing Continue worktrees already contain `ao hooks claude-code ...` entries from the previous installer. `continueHooks.Install` only reconciles the new `ao hooks continue ...` prefix and preserves those legacy commands, so upgraded worktrees execute every event twice and retain Claude-attributed metadata. Please migrate only the AO-managed legacy Continue entries before installing the new set, preserve user hooks, and cover the upgrade path with an existing-settings fixture.

whoisasx (changes requested):
Requesting changes on three validated issues. The coverage classification is a useful direction, but the Aider notification command can block the interactive session at its first callback. Continue can also rediscover completed questions as live prompts, and existing worktrees need a migration from the legacy Claude-attributed hooks.

All eight GitHub checks are green, but the branch currently conflicts with `main`; please resolve the behavioral findings and then rebase so the integrated result can be verified.

backend/internal/adapters/agent/continueagent/terminal_activity.go
```diff
@@ -0,0 +1,77 @@
...
+		start = 0
+	}
+	recent := lines[start:]
+
+	for i := len(recent) - 1; i >= 0; i-- {
+		line := strings.ToLower(recent[i])
+		if continueQuestionPickerAt(recent, i) {
+			return domain.ActivityWaitingInput, true
+		}
+		if strings.Contains(line, "esc to interrupt") {
```
whoisasx, line 54: [P1] The idle-composer guard rejects the historical `Enter select` row, but the reverse scan then continues into the same completed interaction and finds its older `esc to interrupt` spinner. The detector consequently returns `active` and overwrites the Stop hook's correct idle state. I reproduced this with the live-picker fixture plus the idle composer: it returned `(active, true)`. Please treat the current idle composer as a boundary before scanning historical markers and add the spinner-inclusive completed-question regression test.

backend/internal/adapters/agent/continueagent/continueagent.go
```diff
@@ -22,20 +22,30 @@ package continueagent
...
 )
 
 // adapterID is the AO harness / manifest id. It is the string "continue"
 // (NOT the Go package name "continueagent").
 const adapterID = "continue"
 
+const (
+	continueClaudeSettingsDirName  = ".claude"
+	continueClaudeSettingsFileName = "settings.local.json"
+	continueHookCommandPrefix      = "ao hooks continue "
```
whoisasx, line 45: [P2] The Continue-specific token fixes activity attribution, but it also bypasses conversation extraction in `runHook`: that switch currently calls `hookConversationFacts` only for Claude Code and Codex. Before this PR, Continue arrived through the `claude-code` token and persisted the latest user prompt, assistant update, and transcript path; these fields are now sent empty. Please include Continue in the Claude-compatible conversation path and add a hook request-payload test.

skills/agent-role-smoke/SKILL.md
```diff
@@ -0,0 +1,100 @@
+---
```
whoisasx: [P2] This role-smoke skill and its 861-line runner are unrelated to Aider/Continue activity reporting and add roughly 960 lines to an otherwise focused PR. They also expand the review and behavioral surface without relevant checks on the current head. Please remove both role-smoke files from this PR and submit that work separately if it still needs to land.

whoisasx (changes requested):
The three original findings are addressed: Aider no longer blocks on inherited stdin, completed Continue pickers are partially guarded, and legacy hook commands are migrated. The re-review found two remaining lifecycle/metadata regressions plus unrelated role-smoke files in the PR.

The focused Aider, Continue, activity-dispatch, hooksjson, and CLI packages pass locally. However, an isolated completed-question fixture containing the historical spinner returns `active`, confirming the first inline finding. The current head has no GitHub checks and conflicts with `main`; please address the inline findings, remove the unrelated files, and rebase before re-verification.

backend/internal/adapters/agent/continueagent/terminal_activity.go
nikhilachale: Fixed in 28bb8e42b. The empty Continue composer now terminates reverse scanning, so historical picker/spinner markers cannot overwrite the Stop hook's idle state. Added the spinner-inclusive completed-picker regression fixture.

skills/agent-role-smoke/SKILL.md
nikhilachale: Removed by rebuilding the PR on current main without the unrelated role-smoke commit. Neither role-smoke file is present in the updated PR diff.

backend/internal/adapters/agent/continueagent/continueagent.go
nikhilachale: Fixed in 28bb8e42b. Continue now uses the Claude-compatible conversation extraction path in runHook. Added a request-payload test covering latest user prompt, assistant update, and transcript path.

backend/internal/adapters/agent/continueagent/continueagent.go
```diff
@@ -49,6 +59,34 @@ var continueBinarySpec = binaryutil.BinarySpec{
...
+	{Event: "Stop", Command: continueHookCommandPrefix + "stop"},
+	{Event: "Notification", Command: continueHookCommandPrefix + "notification"},
+	{Event: "SubagentStop", Command: continueHookCommandPrefix + "subagent-stop"},
+	{Event: "SessionEnd", Command: continueHookCommandPrefix + "session-end"},
+}
+
+var continueHooks = hooksjson.Manager{
+	Label:                 adapterID,
+	CommandPrefix:         continueHookCommandPrefix,
+	LegacyCommandPrefixes: []string{"ao hooks claude-code "},
```
whoisasx, line 84: [P2] This migration is one-way. Installing Continue removes ao hooks claude-code, but Claude's hook manager does not remove ao hooks continue when the workspace switches back. A Continue install followed by a Claude install leaves both command sets registered for every event in .claude/settings.local.json, producing duplicate activity and metadata callbacks. Please make hook ownership reciprocal, either by teaching the Claude manager this legacy prefix or by centralizing cleanup for adapters sharing this file, and add the reverse-switch regression test.

whoisasx (changes requested):
Re-reviewed the latest head. The earlier Aider stdin, Continue historical picker/composer, conversation metadata, legacy migration, and unrelated-file findings are addressed. One hook-ownership regression remains when a workspace switches from Continue back to Claude Code: both command sets stay installed in the shared .claude/settings.local.json. All eight GitHub checks are green and the affected package tests pass, but the focused switching regression reproduces the duplicate callbacks. Please make the migration symmetric before merging.

backend/internal/adapters/agent/continueagent/continueagent.go
nikhilachale: Fixed in 48ae9925b. Claude Code now removes the AO-managed `ao hooks continue ` prefix before reconciling its own hooks, making ownership symmetric with the Continue installer. Added a reverse-switch regression that performs a real Continue install followed by a Claude install and verifies Continue commands are removed while the user hook is preserved.

backend/internal/adapters/agent/continueagent/continueagent.go
```diff
@@ -49,6 +59,34 @@ var continueBinarySpec = binaryutil.BinarySpec{
 	},
 }
 
+var continueStartupMatcher = "startup"
```
whoisasx: [P2] Keep the Claude-compatible SessionStart matcher instead of narrowing it to `startup`. Before this PR, Continue delegated to Claude's installer and matched `startup|resume|clear|compact|fork`; Continue's current hook contract itself declares `startup`, `resume`, `clear`, and `compact`. With the new matcher, those non-startup callbacks are skipped, leaving the native session ID and launch metadata stale until some later hook happens. Please mirror the supported source set and add a settings regression for non-startup SessionStart events.

whoisasx (changes requested):
The previous blockers are addressed: Aider no longer consumes inherited notification stdin, the Continue picker and conversation metadata regressions are covered, role-smoke files are gone, and Claude/Continue hook ownership is now symmetric. One remaining matcher regression prevents approval: Continue declares Claude-compatible SessionStart sources beyond a fresh startup, but this installer registers only `startup`, so resume/clear/compact lifecycle callbacks can be skipped and native session or launch metadata can remain stale. Please mirror the supported matcher set and add coverage for non-startup sources. All current CI checks are green.

backend/internal/adapters/agent/continueagent/continueagent.go
nikhilachale: Fixed in 77360334c. Continue now installs SessionStart with its supported `startup|resume|clear|compact` matcher instead of narrowing to `startup`. Added a settings-level regression that decodes the installed hook groups and requires the complete non-startup source set.

nikhilachale:
Final re-review pass pushed in c07973b64. In addition to the prior SessionStart matcher fix, this makes the Continue empty composer an authoritative idle fallback, classifies version-dependent Continue hooks as partial coverage so silence does not produce no_signal, and lets Aider/Continue prove an empty composer to the message-delivery gate without changing the Aider waiting-input activity. Added focused adapter and session-manager regressions.

Verification: affected adapter/dispatch/observer/session-manager/session-service suites pass; golangci-lint reports 0 issues. The repository-wide test run had one unrelated timing failure in TestPiExtensionStopEventOrderingFollowsProbedSupport, which passed immediately in isolation. All addressed review conversations are resolved. @whoisasx please re-review the latest head.

backend/internal/session_manager/message_delivery.go
```diff
@@ -79,6 +81,9 @@ func (m *Manager) WaitForMessageDeliveryReady(ctx context.Context, id domain.Ses
 				if output, outputErr := m.runtime.GetOutput(ctx, handle, messageDeliveryReadyLines); outputErr == nil {
 					state, authoritative := detector.DetectTerminalActivity(output)
 					ready = authoritative && state == domain.ActivityIdle
+					if !authoritative && rec.Activity.State == domain.ActivityIdle && emptyComposerDetector != nil {
```
whoisasx: [P2] This fallback cannot make an Aider orchestrator ready after its completion callback. Aider maps its notification event to waiting_input, but ComposerIsEmpty is consulted here only when the durable state is already idle. After the first completed turn, later title-coordination delivery can therefore wait until its one-minute context timeout even though the Aider prompt is visibly empty. Please let the appropriate provider-proven empty composer satisfy readiness from waiting_input without weakening other adapters, and add a regression starting with ActivityWaitingInput.

whoisasx (changes requested):
Re-reviewed the latest head. The previous Continue SessionStart matcher finding is fixed: startup, resume, clear, and compact sources are now preserved. One readiness regression remains in the newly added Aider fallback. Aider completion deliberately records waiting_input, but message delivery consults the empty-composer proof only for an already-idle record, so an Aider orchestrator can remain ineligible for later coordination despite visibly being ready.

All eight GitHub checks are green. Please address the inline finding and add a regression that starts from ActivityWaitingInput before merging.

backend/internal/session_manager/message_delivery.go
nikhilachale: Fixed in 8487dfb3b. Provider-proven empty composers now satisfy readiness from both idle and waiting_input; adapters without EmptyComposerDetector remain unchanged. Added the requested ActivityWaitingInput regression using the Aider harness.

backend/internal/session_manager/message_delivery.go
```diff
@@ -79,6 +81,11 @@ func (m *Manager) WaitForMessageDeliveryReady(ctx context.Context, id domain.Ses
 				if output, outputErr := m.runtime.GetOutput(ctx, handle, messageDeliveryReadyLines); outputErr == nil {
 					state, authoritative := detector.DetectTerminalActivity(output)
 					ready = authoritative && state == domain.ActivityIdle
+					if !authoritative &&
+						(rec.Activity.State == domain.ActivityIdle || rec.Activity.State == domain.ActivityWaitingInput) &&
```
whoisasx: [P1] This makes every `EmptyComposerDetector` eligible for unsolicited delivery while the durable session is `waiting_input`, not just Aider. Codex, Claude, and Continue also implement that interface, and `waiting_input` can represent a live permission or structured-input boundary. In particular, Codex's surface parser can return unknown work with an apparently empty composer, so background title coordination can now proceed despite the sticky interaction state and inject text/Enter into the wrong UI. Please gate this with a separate waiting-input-readiness capability implemented by Aider (and any adapter that can prove the same contract), and add a regression showing a non-opted-in waiting-input adapter remains blocked even when its composer detector returns true.

whoisasx (changes requested):
The latest commit fixes the intended Aider case, but the readiness exception is currently attached to the broad `EmptyComposerDetector` capability rather than to Aider's completion semantics. That also changes delivery behavior for Codex, Claude, and Continue whenever their durable state is `waiting_input`. Please scope this exception through an explicit capability implemented only by adapters for which an empty composer authoritatively means that a waiting-input state is safe for unsolicited delivery.

All eight GitHub checks are green and `git diff --check` passes. Per our review policy, I did not run Go tests.

backend/internal/session_manager/message_delivery.go
nikhilachale: Fixed in 08342bb9d. Waiting-input readiness now requires the new WaitingInputComposerReadiness capability in addition to EmptyComposerDetector. Only Aider opts in. Added a regression proving a non-opted-in waiting_input adapter remains blocked even when ComposerIsEmpty returns true.

whoisasx (approved):
Re-reviewed and resolved the conflict against current main at ae890182. The resolution retains both Cursor and Continue dispatch imports, keeps their specialized derivers, preserves Aider/Continue partial signal coverage, and keeps waiting-input composer readiness explicitly Aider-only. The previous blocker is addressed and GitHub now reports the PR mergeable. Approving from my side; please merge after the newly triggered CI checks complete successfully.

Issue #3527, opened 2026-08-03
macOS: 'Restart to update' silently does nothing when the app runs translocated / from a read-only location


AgentWrapper:
## Bug

When Agent Orchestrator runs under macOS App Translocation (launched straight from Downloads/a DMG without being moved to /Applications), the bundle lives at a randomized **read-only** path (`/private/var/folders/.../AppTranslocation/<uuid>/d/Agent Orchestrator.app`). An update still checks, downloads, and stages fine — the sidebar shows "Restart to update — <version> ready" — but clicking it does **nothing**: Squirrel.Mac cannot replace a read-only bundle, so `autoUpdater.quitAndInstall()` is a silent no-op. No restart, no error, no guidance.

Observed live by the maintainer on 2026-08-03 with v0.11.2-nightly.202608032052 staged (download verified present in `~/Library/Caches/agent-orchestrator-updater/pending/`); resolved manually by moving the app to /Applications and relaunching.

## Fix

In `quitAndInstallUpdate()` (frontend/src/main/auto-updater.ts), before calling `autoUpdater.quitAndInstall()`, detect on macOS:
1. a translocated bundle (`process.execPath` under `/AppTranslocation/`), or
2. a non-writable app bundle (read-only volume / permissions),

and show an actionable dialog ("Move Agent Orchestrator to /Applications and relaunch, then try again") instead of silently doing nothing. Keep the staged row so the install can be retried after the move.

Related: #3034 (different failure in the same install surface — this one is not a ShipIt/ditto regression), #3152.

Rishet11:
@AgentWrapper #3623 fixes this

Rishet11:
The translocation install guard is on main (getMacInstallBlocker). Closing as implemented.

PR #3785, opened 2026-08-09
fix(desktop): guard quitAndInstallUpdate against macOS translocation/read-only bundles

krishrathi1:
## Problem

On macOS, when Agent Orchestrator runs under App Translocation (launched straight from Downloads or a mounted DMG without being moved to `/Applications`), the app bundle lives at a randomized **read-only** path (`/private/var/folders/.../AppTranslocation/<uuid>/d/Agent Orchestrator.app`). An update still checks, downloads, and stages fine, and the sidebar shows "Restart to update", but clicking it does nothing: Squirrel.Mac (electron-updater's macOS install backend) cannot replace a read-only bundle, so `autoUpdater.quitAndInstall()` silently no-ops. No restart, no error, no guidance.

`quitAndInstallUpdate()` in `frontend/src/main/auto-updater.ts` had no translocation/writability check at all before calling `quitAndInstall()`.

## Fix

Added a macOS-only guard in front of `autoUpdater.quitAndInstall()`:

- `isTranslocatedPath(execPath)` — true when the executable path contains `/AppTranslocation/`.
- `resolveMacBundleRoot(execPath)` — walks up from `.../Agent Orchestrator.app/Contents/MacOS/<exe>` to the enclosing `.app` bundle root (same three-level walk as `resolveBundlePath()` already does in `main.ts`, implemented via `/`-delimited string splitting rather than `node:path` so it's deterministic under test regardless of the host OS running the test suite, matching `daemon-launch.ts`'s `joinPath` reasoning in this repo).
- `isBundleWritable(bundleRoot, checkAccess?)` — true unless a `fs.accessSync(path, fs.constants.W_OK)` check throws (read-only volume, missing permission bit); `checkAccess` is injected so it's testable without touching a real filesystem.

If either check fires, `quitAndInstallUpdate()` shows an actionable `dialog.showMessageBox` (the same dialog mechanism already used elsewhere in this file, e.g. `ensureUpdatePrefs`) telling the user to move the app to `/Applications` and relaunch, then returns without calling `quitAndInstall()`. Nothing in the guard touches update-settings or the staged-update state, so the install can simply be retried after the user moves the app.

All three detection functions take their inputs as parameters instead of reading `process.execPath`/`process.platform` internally, and `quitAndInstallUpdate()` itself takes `isMac`/`execPath`/`checkAccess` overrides that default to the real process — the same dependency-injection-via-default-parameter convention `keybinding-settings.ts`'s `coerceKeybindingOverrides(raw, isMac)` already uses in this directory. This is what makes the guard unit-testable with fabricated path strings and a mocked fs check, no real macOS bundle required.

## Verification

**Important disclosure: I do not have access to a macOS environment.** I could not run the packaged Electron app, reproduce a real translocated bundle, or see the actual dialog render. Everything below was verified through code review, TypeScript typechecking, and unit tests with mocked/fabricated inputs — not a live macOS run. This should be treated as unverified against real macOS behavior until someone with mac hardware confirms the dialog UX and real-world translocation detection.

What I did verify:

- Read `frontend/src/main/auto-updater.ts` in full and confirmed `quitAndInstallUpdate()` was exactly as bare as described (no guard of any kind) before this change.
- Traced the call site: `main.ts`'s `ipcMain.handle("updates:install", () => { quitAndInstallUpdate(); })` — the new optional parameters default to the real process, so this call site's behavior is unchanged when both checks pass.
- Confirmed the staged-update/escalation tracking (`stagedVersion`, `stagedAtMs`, etc.) and `update-settings` persistence are untouched by the guard — nothing is cleared or invalidated when it fires.
- Added unit tests to `frontend/src/main/auto-updater.test.ts` (11 new cases, using the existing `importAutoUpdater()` electron-mocking harness already in that file) covering:
  - A translocated path string correctly detected as translocated, and a normal `/Applications/Agent Orchestrator.app/...` path correctly not flagged.
  - `resolveMacBundleRoot` correctly walking up to the `.app` root from both a normal install path and a translocation-mirror path.
  - A mocked non-writable access check correctly failing `isBundleWritable`, and a mocked writable one correctly passing.
  - `quitAndInstallUpdate` calling through to `autoUpdater.quitAndInstall(false, true)` when the bundle is a normal writable install, and calling `dialog.showMessageBox` instead (with an `/Applications`-mentioning, install-guidance message) and *not* calling `quitAndInstall` when translocated or non-writable.
  - The guard being a no-op on non-darwin platforms even for a translocated-looking, non-writable path (still calls `quitAndInstall` as before).
  - No-op in dev (unpackaged) before the platform check even runs, matching existing behavior.
- Proved the tests actually exercise the fix using a patch-file diff/reverse-apply (not `git stash`, to avoid colliding with other concurrent work in shared worktrees of this repo): reversed just the `auto-updater.ts` change while keeping the new tests, confirmed 8 of the 11 new tests fail against the unpatched code (`module.isTranslocatedPath is not a function`, and `quitAndInstall` being called when it shouldn't be), then reapplied the fix and confirmed all 47 tests in the file pass.
- `npx tsc --noEmit` from `frontend/` — clean, no errors.
- `npx vitest run --config vite.renderer.config.ts src/main/auto-updater.test.ts` — 47/47 passed.
- `npx vitest run --config vite.renderer.config.ts src/main` (broader regression check) — 284 passed, 2 skipped, 4 failed. The 4 failures are all in `supervisor-link.test.ts` (`EACCES: permission denied` creating a Unix-domain-socket file under the Windows temp dir) — a pre-existing Windows-sandbox limitation unrelated to this change; I did not touch `supervisor-link.ts` or its test, and `auto-updater.test.ts` was unaffected in the same run.

Given the lack of a real macOS test environment, I'd appreciate a reviewer with mac hardware confirming: the dialog actually renders with sensible copy, `isTranslocatedPath`/`resolveMacBundleRoot` correctly handle a real translocated bundle's path shape, and `fs.accessSync` behaves as expected against an actual read-only-mounted or translocated bundle directory.

Addresses #3527.

### DIFF
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -1231,3 +1231,130 @@ describe("returnToHome", () => {
     expect(autoUpdater.checkForUpdates).not.toHaveBeenCalled();
   });
 });
+
+// #3527: on macOS, Squirrel.Mac cannot replace a translocated or otherwise
+// read-only .app bundle, so quitAndInstall() used to silently no-op with no
+// restart and no error. These cover the guard added in front of it.
+describe("macOS translocation/writability guard", () => {
+  const installedExecPath =
+    "/Applications/Agent Orchestrator.app/Contents/MacOS/Agent Orchestrator";
+  const translocatedExecPath =
+    "/private/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/AppTranslocation/1234-5678-ABCD/d/Agent Orchestrator.app/Contents/MacOS/Agent Orchestrator";
+
+  describe("isTranslocatedPath", () => {
+    it("flags a path running out of the App Translocation mirror", async () => {
+      const { module } = await importAutoUpdater();
+      expect(module.isTranslocatedPath(translocatedExecPath)).toBe(true);
+    });
+
+    it("does not flag a normal /Applications install path", async () => {
+      const { module } = await importAutoUpdater();
+      expect(module.isTranslocatedPath(installedExecPath)).toBe(false);
+    });
+  });
+
+  describe("resolveMacBundleRoot", () => {
+    it("walks up from Contents/MacOS/<exe> to the .app bundle root", async () => {
+      const { module } = await importAutoUpdater();
+      expect(module.resolveMacBundleRoot(installedExecPath)).toBe(
+        "/Applications/Agent Orchestrator.app",
+      );
+    });
+
+    it("resolves the .app root even from inside the translocation mirror", async () => {
+      const { module } = await importAutoUpdater();
+      expect(module.resolveMacBundleRoot(translocatedExecPath)).toBe(
+        "/private/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/AppTranslocation/1234-5678-ABCD/d/Agent Orchestrator.app",
+      );
+    });
+  });
+
+  describe("isBundleWritable", () => {
+    it("is true when the access check does not throw", async () => {
+      const { module } = await importAutoUpdater();
+      expect(
+        module.isBundleWritable(
+          "/Applications/Agent Orchestrator.app",
+          () => undefined,
+        ),
+      ).toBe(true);
+    });
+
+    it("is false when the access check throws (read-only volume/permission)", async () => {
+      const { module } = await importAutoUpdater();
+      expect(
+        module.isBundleWritable(
+          "/Volumes/ReadOnlyDMG/Agent Orchestrator.app",
+          () => {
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
Closing as superseded: macOS install translocation is now guarded on main. getMacInstallBlocker in auto-updater.ts checks for App Translocation and unwritable bundles before quitAndInstallUpdate and shows a dialog, which covers this PR's scenario. Thanks for catching the failure mode.

Issue #3791, opened 2026-08-09
fix(terminal): Copy full prompt and response from agent TUIs by reading native transcript


vanshx999:
**The Issue:**
Every agent AO runs (opencode, codex, claude-code) renders as a full-screen TUI inside xterm's **alternate screen buffer**, which has **no scrollback**. xterm only ever holds the one visible screen — so `term.getSelection()` and `term.selectAll()` can only return what fits in the viewport right now. Long prompts and long responses both get lost the moment they scroll off. The text literally doesn't exist inside the terminal anymore. This makes it impossible to copy a complete assistant response or a long user prompt into another tool — you can only copy one screenful at a time.

**The Fix:**
Pull the text from each agent's own persisted transcript instead of the terminal. AO's terminal rendering layer stays untouched — we edit each agent's **harness adapter** directly. Each adapter already knows exactly where its native transcript lives:

| Agent | Native transcript location | AO already locates it in |
|---|---|---|
| **opencode** | `~/.local/share/opencode/opencode.db` (SQLite) | `opencode.go` auth probe (line 298-356) |
| **codex** | `<CODEX_HOME>/sessions/.../rollout-<sessionID>.jsonl` | `continuation.go:LocateTranscript` |
| **claude-code** | `~/.claude/projects/<key>/<session-id>.jsonl` | `claudecode.go:NativeConversationExists` |

We add one new `Transcript(ctx, session) → [{role, text}]` method to each harness adapter, expose it via a new daemon endpoint `GET /api/v1/sessions/{id}/transcript`, and add two context-menu items in the terminal pane: **Copy prompt** (last user turn) and **Copy response** (last assistant turn). Both write via the existing `aoBridge.clipboard.writeText`.

**macOS: Ctrl+click opens the context menu too.**
The terminal context menu is triggered by the `contextmenu` event. On macOS, a Control+click fires the same event as a right-click, but xterm's custom key/mouse handlers can swallow it before the browser sees it. The fix must explicitly listen for `Ctrl+click` on the terminal host and route it to the same `openContextMenu` handler, so Mac users who don't have a right-click button can still access the copy actions.

Same root cause for all agents, same fix pattern. No xterm.js changes, no terminal rendering layer touched.

**Scope of changes:**
- `ports/agent.go` — new `Transcript` interface method
- `adapters/agent/opencode/opencode.go` — implement `Transcript()`
- `adapters/agent/codex/codex.go` — implement `Transcript()`
[...]

PR #3792, opened 2026-08-09
feat(ports): add Transcript method to Agent interface for full prompt/response copy

vanshx999:
Adds the `Transcript` method to the `Agent` interface — the foundation for copying full prompts and responses from agent TUI panes where xterm's alternate screen buffer provides no scrollback.

This PR is the first of several for #3791. Subsequent PRs will:
- Implement `Transcript()` in each harness adapter (opencode, codex, claude-code)
- Add the daemon endpoint `GET /api/v1/sessions/{id}/transcript`
- Add "Copy prompt" / "Copy response" context menu items in the terminal pane

### DIFF
--- a/backend/internal/adapters/agent/agentbase/agentbase.go
+++ b/backend/internal/adapters/agent/agentbase/agentbase.go
@@ -6,11 +6,53 @@
 package agentbase
 
 import (
+	"bufio"
 	"context"
+	"errors"
+	"io"
 	"strings"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
 )
+
+// MaxTranscriptLineBytes bounds a single JSONL line read from a native
+// transcript. Agents can emit enormous single-line tool outputs or diffs; an
+// oversized line is skipped rather than aborting the whole transcript read.
+const MaxTranscriptLineBytes = 8 << 20 // 8 MiB
+
+// ReadTranscriptLine reads the next non-empty line from r, draining (not
+// buffering) lines longer than MaxTranscriptLineBytes so one oversized message
+// can never fail the rest of the transcript. ok=false marks end of input.
+func ReadTranscriptLine(r *bufio.Reader) (line string, ok bool, err error) {
+	for {
+		var buf []byte
+		for {
+			frag, readErr := r.ReadSlice('\n')
+			if len(buf) < MaxTranscriptLineBytes {
+				buf = append(buf, frag...)
+				if len(buf) > MaxTranscriptLineBytes {
+					buf = nil // oversized: drop and keep draining the line
+				}
+			}
+			if errors.Is(readErr, bufio.ErrBufferFull) {
+				continue
+			}
+			if readErr == io.EOF {
+				if len(buf) == 0 && len(frag) == 0 {
+					return "", false, nil
+				}
+			} else if readErr != nil {
+				return "", false, readErr
+			}
+			break
+		}
+		line = strings.TrimSpace(string(buf))
+		if line == "" {
+			continue
+		}
+		return line, true, nil
+	}
+}
 
 // ModelConfigSpec returns the common optional model config field used by
 // adapters that forward a --model-style argument.
--- a/backend/internal/adapters/agent/agentbase/agentbase.go
+++ b/backend/internal/adapters/agent/agentbase/agentbase.go
@@ -26,25 +26,29 @@ const MaxTranscriptLineBytes = 8 << 20 // 8 MiB
 func ReadTranscriptLine(r *bufio.Reader) (line string, ok bool, err error) {
 	for {
 		var buf []byte
+		oversized := false
 		for {
 			frag, readErr := r.ReadSlice('\n')
-			if len(buf) < MaxTranscriptLineBytes {
+			if !oversized && len(buf)+len(frag) > MaxTranscriptLineBytes {
+				oversized = true
+				buf = nil // oversized: drop what we have and keep draining the line
+			}
+			if !oversized {
 				buf = append(buf, frag...)
-				if len(buf) > MaxTranscriptLineBytes {
-					buf = nil // oversized: drop and keep draining the line
-				}
 			}
 			if errors.Is(readErr, bufio.ErrBufferFull) {
 				continue
 			}
-			if readErr == io.EOF {
-				if len(buf) == 0 && len(frag) == 0 {
-					return "", false, nil
-				}
-			} else if readErr != nil {
+			if readErr == io.EOF && len(buf) == 0 && len(frag) == 0 && !oversized {
+				return "", false, nil
+			}
+			if readErr != nil && readErr != io.EOF {
 				return "", false, readErr
 			}
 			break
+		}
+		if oversized {
+			continue // drop the line entirely; keep reading
 		}
 		line = strings.TrimSpace(string(buf))
 		if line == "" {
--- a/backend/internal/adapters/agent/agentbase/agentbase_test.go
+++ b/backend/internal/adapters/agent/agentbase/agentbase_test.go
@@ -1,9 +1,11 @@
 package agentbase
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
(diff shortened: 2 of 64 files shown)


Vaibhaav-Tiwari:
can we have a meeting for this? i want to understand this first

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

vanshx999:
Issue #3791 — Copy full transcript from agent TUIs

Agent TUIs (opencode / codex / claude-code) run in the alt screen buffer, so xterm.js keeps no scrollback — the old context-menu copy could only grab the word under the cursor. This adds real "Copy full transcript": the pane fetches the session's transcript from the daemon and writes it to the clipboard.

What changed
- ports.Agent.Transcript + session-service/manager plumbing, surfaced as GET /api/v1/sessions/{sessionId}/transcript → { sessionId, messages: [{ role, text }] }.
- Real transcript readers in three adapters:
  - opencode — SQLite ~/.local/share/opencode/opencode.db (message/part tables; role parsed from the data JSON).
  - codex — rollout JSONL under ~/.codex/sessions/..., collecting both input_text and output_text parts.
  - claude-code — project session JSONL under ~/.claude/projects/..., handling array and string content forms.
- Regenerated OpenAPI spec + frontend types; TranscriptMessage now carries json tags.
- Frontend: "Copy full transcript" context-menu item on agent panes, clipboard write via the bridge, locale strings for all 8 languages.
- Tests for the codex/claude-code/opencode readers, the session transcript endpoint, and adapter fakes. go test ./..., golangci-lint, frontend typecheck and vitest pass.

Why this didn't work in the first cut (and the two fixes)
1. Retained terminal staleness: the pane's xterm instance is kept alive across session switches, so the action captured the sessionId from mount time (empty → no-op, and the default word-copy won). The action now reads live props via callbacksRef.current.
2. Schema mismatch: opencode keeps role inside the message data JSON (not a column), and its part table has no type column — the query 500'd. Rewrote the reader to parse role from the JSON payload.

Verified: ran the desktop build against the local daemon; both an opencode session and a claude-code session return real transcripts from the endpoint, and the context menu copies the full conversation.

i-trytoohard:
Match confidence: `0.62` (fuzzy).

Relevant discussion excerpts:
- [1:02:11] vanshx999: Yeah, that's not like something I figured out. Like I did figure out for like copying the prompt or a particular response, I can fix that with like copy the particular area of a transcript, but I haven't like applied it. I've still, I've just like worked with the copy the full transcript. And [...]
- [1:01:52] prateek: normal experience is not great. So we are adding like other ways to circumvent that, you know. How can you fix the terminal experience itself? Because even right now, drag and select is not working, right? Even if I didn't want to select the whole session transcript, but only like 4 paragraphs, [...]
- [1:01:48] vanshx999: Yeah, like it is like I found— [1:01:52] prateek: normal experience is not great. So we are adding like other ways to circumvent that, you know. How can you fix the terminal experience itself? Because even right now, drag and select is not working, right? Even if I didn't want to select the [...]

Vaibhaav-Tiwari (changes requested):
1. Long messages break the whole copy. If a transcript has one message that's unusually large (big diff, big tool
  output — not rare for coding agents), the entire "copy transcript" fails and returns nothing instead of just
  skipping that one message.
  2. Some sessions will copy an empty transcript instead of the real one. There's existing logic for sessions where
  the agent's config directory isn't in the default location, but this new code never hooks into it — so those
  sessions silently get nothing back instead of the actual transcript.
  3. Clipboard race condition. Clicking "copy transcript" then quickly doing a normal copy right after can result in
  the transcript overwriting your more recent copy, since the transcript fetch runs in the background with nothing
  preventing it from finishing later.
  4. No tests for the new backend endpoint. Nothing actually calls the new transcript API in a test, so a future
  regression here won't get caught.
  5. Design nit: the new method was added in a way that forced ~27 unrelated agent files to get a do-nothing stub,
  when the codebase already has a cleaner pattern for "only some agents support this feature" that wasn't used here.

vanshx999:
@Vaibhaav-Tiwari  Thanks for the detailed review — all five points are valid, here's how I'll address them:
1. Oversized messages — Agreed. I'll make messages render/handle independently so one huge diff or tool output skips only itself instead of failing the whole transcript.
2. Non-default config dirs — Agreed. I missed the existing resolution logic for non-default agent config locations; I'll wire the transcript lookup to reuse it so those sessions return real transcripts instead of empty.
3. Clipboard race — Agreed. I'll add a guard so only the most recent copy intent wins (a token/cancel on the async fetch), preventing a late transcript from overwriting a newer manual copy.
4. No backend tests — Agreed, my miss. I'll add controller tests for the endpoint (happy path, empty transcript, non-default config dir, oversized message) and keep the wire-shape parity tests green.
5. 27 stubs — Agreed. I'll refactor Transcript out of the base Agent interface into the optional capability pattern the codebase already uses (like AgentTranscriptLocator), so only opencode/codex/claude-code declare it.
I'll also fold in the product point from the call about copying just a few paragraphs, and get the openapi.yaml/schema.ts conflicts resolved. Will push the fixes as follow-ups on this branch.

vanshx999:
Addressed all review feedback on `fix/3791-transcript-copy` (head `0e231f105`, PR now MERGEABLE):

1. **Long messages no longer break the copy.** `Transcript` is now gated behind an optional `ports.AgentTranscriptReader` capability, and the codex/claudecode JSONL readers read through `agentbase.ReadTranscriptLine` which skips (drains) lines over 8 MiB instead of failing. opencode skips individual unreadable messages and keeps the rest of the transcript.

2. **Non-default config dirs resolve transcripts.** New `Manager.NativeSessionConfigDir` resolves the session's stored native-session `ConfigDir` (via `AgentSwitchStore.ListAgentNativeSessions`) first, then live-resolves from the launch runtime env (honoring CODEX_HOME / CLAUDE_CONFIG_DIR / OPENCODE_DATA_DIR) through the existing `nativeConfigDir` path. `Service.Transcript` passes it to the adapter as `codexHome` / `claudeConfigDir` / `opencodeDataDir` metadata; opencode also gained a `NativeSessionConfigDir` implementation.

3. **Clipboard race guarded.** `XtermTerminal` uses a token ref: every copy intent (selection or transcript) bumps it, and the async transcript write only proceeds if its token is still current, so a late-arriving transcript can never clobber a newer copy.

4. **Backend endpoint tests.** Added `TestSessionsAPI_GetTranscript` (200 wire shape incl. the new `index` field), empty-transcript, and 404 cases in `httpd/controllers/sessions_test.go`; plus service-level tests in `service/session/service_test.go` (happy path, not-found, unsupported agent, agent without the reader capability, reader error, empty).

5. **No more stubs.** `Transcript` was removed from `ports.Agent`; the 24 adapter stubs and test fakes were deleted; only codex, claudecode, and opencode declare the optional `AgentTranscriptReader` (with compile-time asserts).

Also regenerated `openapi.yaml` + `frontend/src/api/schema.ts` for the new `index` field. Full backend suite, gofmt, go vet, golangci-lint (at base baseline), and frontend typecheck pass.

i-trytoohard:
From the NotesBot call on `2026-08-12`, this existing pr appears to be relevant.
Match confidence: `0.62` (fuzzy).

Relevant discussion excerpts:
- **🧰 Browser, Hooks, Copy Features, and GitLab** - *vanshx999*: said he raised a PR two days earlier that adds an option to copy the full transcript from the terminal UI. - *vanshx999*: added options to copy the last response and the last prompt, and is still working on selecting/copying a specific area before [...]
- - *vanshx999*: said he raised a PR two days earlier that adds an option to copy the full transcript from the terminal UI. - *vanshx999*: added options to copy the last response and the last prompt, and is still working on selecting/copying a specific area before raising another PR.
- - *vanshx999*: said he raised a PR two days earlier that adds an option to copy the full transcript from the terminal UI. - *vanshx999*: added options to copy the last response and the last prompt, and is still working on selecting/copying a specific area before raising another PR. - *prateek*: asked where the [...]

Source thread: https://discord.com/channels/1476302178913357958/1537149882463297597/1537149882463297597

i-trytoohard:
@vanshx999 Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@vanshx999
Closing this PR as stale after 30+ days without author activity after the last review follow-up.

PR #3797, opened 2026-08-09
perf(renderer): reduce unnecessary renders

ronishrohan:
## What

Improve renderer performance by reducing unnecessary rerenders and startup payload.

image

## Why

Common UI interactions caused avoidable rendering work.

## How

- Profile renderer commits in development.
- Defer non-English locale catalogs.
- Keep ordinary input drafts out of React's render path.

## Testing

- `npm test`
- `npm run typecheck`

## Checklist

- [ ] Relevant CI checks are pending


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/i18n/instance.ts
```diff
@@ -51,3 +35,10 @@ function initializeI18n(instance: i18n, locale: AppLocale, catalogs: Translation
 }
 
 export const appI18n = initializeI18n(createInstance().use(initReactI18next), DEFAULT_LOCALE, appCatalogs);
+
+/** Ensure the requested catalog is present before switching the renderer language. */
+export async function setAppLocale(locale: AppLocale): Promise<void> {
+	const catalog = await loadCatalog(locale);
```
illegalcall, line 41: [P2] Fall back when a deferred catalog cannot load

`renderApp` awaits `useLocaleStore.load()` before mounting React. For a persisted non-English locale, a missing or corrupt lazy chunk rejects here and escapes that startup call, leaving no React tree or error UI at all. A live switch also persists the locale before loading it, so the same failure can recur on every launch, while `pendingCatalogs` retains the rejected promise. Please recover to English on startup, evict failed cached loads so retry is possible, and cover the rejection path.

frontend/src/renderer/components/CommandPalette.tsx
```diff
@@ -394,8 +385,7 @@ export function CommandPalette() {
 				}}
 				commandProps={{
 					shouldFilter: false,
-					value,
-					onValueChange: setSelectedValue,
+					disablePointerSelection: true,
```
illegalcall: [P2] Keep hover and Enter activation on the same command

With pointer selection disabled, cmdk keeps `data-selected` on the prior keyboard item, while `CommandItem` now paints the hovered row with the same active background. If a user hovers another command and presses Enter, cmdk dispatches the previously selected command rather than the row under the pointer. The new test asserts that selection stays stale but never checks Enter. Please preserve pointer selection or otherwise keep the visible and activated command consistent without rerendering persistent shell content.

frontend/src/renderer/components/ui/dropdown-menu.tsx
```diff
@@ -31,17 +31,22 @@ export function DropdownMenuContent({
...
 				"relative flex cursor-default select-none items-center gap-2.5 rounded-md px-2 py-1.5 text-control outline-none transition-colors",
-				"text-muted-foreground focus:bg-interactive-hover focus:text-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+				"text-muted-foreground hover:bg-interactive-hover hover:text-foreground focus:bg-interactive-hover focus:text-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
 				"[&_svg]:size-icon-lg [&_svg]:shrink-0 [&_svg]:text-passive",
 				inset && "pl-8",
 				className,
 			)}
+			onPointerMove={(event) => {
+				onPointerMove?.(event);
+				event.preventDefault();
```
illegalcall: [P2] Preserve Radix roving focus in shared dropdowns

This unconditional `preventDefault()` suppresses the Radix item focus that normally follows pointer movement. The hovered row gets active hover styling while the old row retains keyboard focus, so hover-then-Enter activates the old item; typeahead also starts from that stale focus. Because this is the shared primitive, the regression reaches settings, sidebar, terminal, and title-bar menus. Please optimize the render boundary instead of globally changing menu focus semantics, or provide an interaction model whose visual and keyboard targets stay aligned.

illegalcall (commented):

## Standards
No hard violation of `AGENTS.md` or `CONTRIBUTING.md` was found. Two non-blocking maintainability observations remain: `statusesMatch` manually mirrors every `DaemonStatus` field, and the profiled board wrapper is duplicated across the two board routes.

## Spec
Three correctness issues are attached inline: lazy locale-load failure can prevent the renderer from mounting; command-palette hover can diverge from Enter activation; and the shared dropdown primitive now suppresses Radix roving focus across every menu.

I did not run project programs or tests under the review-only policy. The existing `test`, `renderer-smoke`, and `scan` checks on the reviewed head are green.

Summary: Standards has 2 judgment-call observations and no hard violations; Spec has 3 blocking findings. The worst issue is that a persisted non-English locale can leave the window blank when its deferred catalog fails to load.

frontend/src/renderer/i18n/instance.ts
illegalcall: Fixed in 520f2f74a. Failed deferred-catalog promises are now evicted so later loads retry; startup catches a persisted catalog failure and applies English before mounting; and live locale changes preload the catalog before persisting the preference. Added regression coverage for startup fallback, avoiding persistence on load failure, and retrying a rejected catalog load.

frontend/src/renderer/components/CommandPalette.tsx
illegalcall: Fixed in 520f2f74a. Restored cmdk pointer selection and removed the separate hover-only active styling, so the visibly selected row is also the command Enter activates. The unit test now hovers Toggle theme and verifies Enter runs that command. The profiling test still verifies that palette hover does not rerender the persistent sidebar or route.

frontend/src/renderer/components/ui/dropdown-menu.tsx
illegalcall: Fixed in 520f2f74a. Removed the shared onPointerMove preventDefault override and hover-only styling, restoring Radix roving focus for all dropdown menus. The regression test now verifies that pointer movement focuses the hovered item and Enter activates that item rather than the previously focused one.

illegalcall:
Addressed review [4926506236](https://github.com/Untrivial-ai/agent-orchestrator/pull/3797#pullrequestreview-4926506236) in `520f2f74a`:

- Deferred locale failures now fall back to English at startup, reject before persistence during live changes, and evict failed cached promises so retry works.
- Command-palette hover once again updates cmdk selection, keeping the highlighted command aligned with Enter activation while preserving persistent-shell render isolation.
- Shared dropdowns once again use Radix roving focus, so hover, focus, typeahead, and Enter stay aligned.

Verification: frontend + E2E TypeScript checks; 51 focused tests; full renderer suite (2,228 passed, 1 skipped); renderer E2E suite (23 passed); profiling E2E (2 passed).

frontend/src/renderer/components/chat/ChatComposer.tsx
```diff
@@ -255,25 +278,25 @@ export function ChatComposer({
 		if (!node) return;
 		node.setSelectionRange(target, target);
 		node.focus();
-	}, [text]);
+	}, [suggestionDraft]);
```
illegalcall: **[P1] Consume the pending caret on every programmatic edit**

After an ordinary successful send, `applyText("", 0)` stores `pendingCaret`, but `syncDraft` deliberately returns the same `EMPTY_SUGGESTION_DRAFT` object. This effect therefore does not run and the pending `0` survives. When the next message starts with `/` or `@`, that active-trigger update finally changes `suggestionDraft`, consumes the stale target, and moves the caret to the start; the next query character is inserted before the sigil (or closes completion). Drive caret application from every programmatic edit, or set and clear the selection synchronously, so clearing a non-trigger draft cannot leave stale caret state.

frontend/src/renderer/components/TaskComposer.tsx
```diff
@@ -248,14 +250,17 @@ export function TaskComposer({
 			const attachmentPayloads = await toSettledPayload();
 			const sessionId = await createTask({
 				projectId,
-				brief: prompt,
+				brief: promptValueRef.current,
```
illegalcall: **[P1] Snapshot the prompt before waiting for attachments**

This mutable ref is read only after `await toSettledPayload()`. A pasted or dropped file can still be in `FileReader` at that point, and the textarea remains editable while `isSubmitting`, so text entered after the user presses Start is silently included in the already-submitted task (and then cleared on success). Capture the brief before the first await; if editing remains enabled during submission, also avoid clearing text that no longer matches that submitted snapshot.

illegalcall (commented):
Changes requested.

The three issues from the previous pass are fixed: locale loading now falls back and retries, and both command-palette and dropdown pointer selection again agree with keyboard focus. Two correctness regressions remain in the DOM-backed composer work.

## Standards
No blocking repository-convention deviations found.

## Spec / correctness
- ChatComposer can retain a stale programmatic caret after clearing an ordinary draft, corrupting the next slash/file completion.
- TaskComposer samples its mutable prompt only after waiting for attachment reads, so edits made after submission can alter the task that is sent.

The current GitHub checks pass, but neither asynchronous sequence is covered by the added tests.

frontend/src/renderer/components/chat/ChatComposer.tsx
illegalcall: Fixed in ffe1bdb8a. Programmatic ChatComposer edits now write the DOM value, selection, and focus synchronously, so there is no pending caret that can survive an ordinary clear. Added a regression test that sends ordinary text and immediately types `/r`; it asserts the value/caret remain `/r`/2 and completion stays open.

frontend/src/renderer/components/TaskComposer.tsx
illegalcall: Fixed in ffe1bdb8a. TaskComposer now snapshots both the prompt text and prompt revision before waiting for attachment reads. The request uses that snapshot, and success clears the DOM-backed prompt only when no edit occurred since submission. Added a slow-FileReader regression test proving the original brief is sent while a later edit remains in the field after successful creation.

illegalcall:
Addressed review [4930059646](https://github.com/Untrivial-ai/agent-orchestrator/pull/3797#pullrequestreview-4930059646) in `ffe1bdb8a`:

- ChatComposer now applies every programmatic value/caret/focus update synchronously, preventing a stale clear caret from corrupting the next slash or file completion.
- TaskComposer snapshots the prompt before attachment settlement and tracks prompt revisions, so later edits neither change the submitted brief nor get cleared when that submission succeeds.
- Added regression tests for both asynchronous sequences.

Verification: focused composer tests (58 passed), frontend and E2E TypeScript checks, full renderer suite (2,230 passed, 1 skipped), and renderer E2E suite (23 passed).

Issue #3787, opened 2026-08-09
Renderer exits during UI interaction after xterm Viewport.syncScrollArea exception


tamarkhl:
## What happened

On macOS arm64, the Electron development app loaded its board normally after a clean frontend install and successful typecheck/package build. Clicking through board and session UI then caused the app to exit. Immediately before exit, Vite repeatedly reported this unhandled renderer exception:

```text
TypeError: Cannot read properties of undefined (reading 'dimensions')
    at get dimensions (node_modules/@xterm/xterm/lib/xterm.js:1:109483)
    at t.Viewport.syncScrollArea (node_modules/@xterm/xterm/lib/xterm.js:1:50802)
```

## Reproduction environment

- macOS arm64
- Node 22.22.3
- npm 10.9.8
- `@xterm/xterm` `^5.5.0`
- AO source head `91f4576cbf14b92d681628dc9aa4a98a5b23541e`, rebased on upstream `d293ea81f9e3415b18fd9f62353e6335b947c854` (the carried patch did not modify the frontend terminal implementation)

## Steps observed

1. Run `npm ci` in `frontend/`.
2. Run `npm run typecheck` and `npm run package` (both pass).
3. Run the Electron development app.
4. Open the board/session UI and interact with navigation/session controls.
5. The xterm error repeats and the Electron process exits.

## Expected behavior

Terminal viewport synchronization should tolerate renderer/component teardown or an unavailable terminal element; interacting with the UI must not terminate the app.

## Notes

The xterm exception is the only repeated unhandled renderer error captured immediately before the exit. This report does not claim that causality has been fully isolated; a teardown/race in the terminal viewport lifecycle is a likely area to inspect. The separate dev-daemon 3002 versus Vite-proxy 3001 regression is already tracked in #2845.

somewherelostt:
Thanks for contributing to Agent Orchestrator.

This issue is being picked up by the current external contributor on-call pair:

- @neversettle17-101
- @illegalcall

If someone is already working on this, please continue as usual.
The on-call pair is added for visibility, tracking, and support, not to take over the work.
If you need help with review, direction, reproduction, or next steps, please tag @neversettle17-101 and @illegalcall here.

For faster context or live questions, you can also join the AO Discord.

Join the session here:
https://discord.gg/H6ZDcUXmq

Come by if you want to see what is being built, ask questions, or just hang around with the community.

yashksaini-coder:
[...]

PR #3804, opened 2026-08-10
fix(frontend): avoid xterm viewport disposal race

### LINKED ISSUE #3803: Xterm handle replacement can dispose queued viewport work
## Bug

Replacing a retained terminal's live xterm handle can dispose the old terminal while xterm still has queued viewport work for that instance.

The queued callback later reaches `Viewport.syncScrollArea` through the disposed terminal and throws in the renderer. This is most visible during same-session handle/generation replacement after agent switching, but the lifetime bug is provider-independent.

## Expected behavior

AO should detach its listeners and transport synchronously, while deferring only the old terminal object's final `dispose()` call until xterm's already-queued viewport work has drained.

A replacement must remain immediate from AO's routing and input-ownership perspective; this fix must not keep the old transport writable for another tick.

## Proposed focused fix

- Detach AO listeners and transport immediately.
- Schedule only the final old-terminal disposal on the next macrotask.
- Cover exact active-handle replacement with real xterm in the retained-terminal Playwright suite.

The prepared patch is limited to `XtermTerminal.tsx` and one focused regression in `terminal-viewport-retention.spec.ts`.

## Validation prepared

- Frontend typecheck
- E2E typecheck
- 49 focused `XtermTerminal` Vitest tests
- Exact real-xterm Playwright handle-replacement regression
- `git diff --check`

This is a small follow-up to the retained-terminal work in #3371.

maplenk:
## What

Avoid an xterm viewport race when AO replaces the active terminal handle for a retained session.

## Why

AO currently disposes the old xterm instance synchronously. xterm may still have queued viewport work for that instance, so the later callback reaches a disposed terminal and throws in the renderer.

Fixes #3803.

## How

- Detach AO listeners and the transport synchronously so the old handle immediately loses routing and input ownership.
- Defer only the old terminal object's final `dispose()` call to the next macrotask.
- Add a focused real-xterm Playwright regression for exact active-handle replacement.

## Testing

- `npm run typecheck`
- `npm run typecheck:e2e`
- `npm test -- src/renderer/components/XtermTerminal.test.tsx` (49/49)
- Focused Playwright handle-replacement regression (1/1)
- `git diff --check`

## Checklist

- [x] Branched from `main`
- [x] One focused change linked to its issue
- [x] Tests added for the user-visible failure
- [x] Relevant frontend and E2E checks pass

### DIFF
--- a/frontend/e2e/terminal-viewport-retention.spec.ts
+++ b/frontend/e2e/terminal-viewport-retention.spec.ts
@@ -42,6 +42,7 @@ const handleC = `${sessionC.id}/terminal_0`;
 const handleD = `${sessionD.id}/terminal_0`;
 const handleE = `${sessionE.id}/terminal_0`;
 const handleF = `${sessionF.id}/terminal_0`;
+const replacementHandleA = `${sessionA.id}/terminal_replacement`;
 const orchestratorHandle = "fake-proj-orchestrator/terminal_0";
 
 const longReplay = [
@@ -189,6 +190,7 @@ async function installHarness(page: Page): Promise<void> {
 		[handleD]: "D short replay",
 		[handleE]: "E short replay",
 		[handleF]: "F short replay",
+		[replacementHandleA]: "A replacement replay",
 		[orchestratorHandle]: "Orchestrator ready",
 	});
 	await page.goto(`/#/projects/fake-proj/sessions/${sessionA.id}`);
@@ -197,6 +199,23 @@ async function installHarness(page: Page): Promise<void> {
 }
 
 test.describe("retained terminal viewport", () => {
+	test("replaces an active session handle without an uncaught xterm viewport error", async ({ page }) => {
+		const pageErrors: string[] = [];
+		page.on("pageerror", (error) => pageErrors.push(error.stack ?? error.message));
+		await installHarness(page);
+
+		await page.evaluate(
+			({ sessionId, handleId }) => window.__aoFakeAgent!.setTerminalHandle(sessionId, handleId),
+			{ sessionId: sessionA.id, handleId: replacementHandleA },
+		);
+		await expect(
+			page.locator(`[data-terminal-cache-key^="session:${sessionA.id}:worker|handle:${replacementHandleA}"]`),
+		).toHaveAttribute("data-terminal-activation-phase", "visible");
+
+		await page.waitForTimeout(250);
+		expect(pageErrors).toEqual([]);
+	});
+
 	test("retains the first of six live sessions with zero reopen and reveals its latest output", async ({
 		page,
 	}) => {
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
## Tested on Linux / Wayland — no reproduction

I hit this crash on Omarchy (Arch Linux) with Hyprland/Wayland and reported it in [#3787](https://github.com/Untrivial-ai/agent-orchestrator/issues/3787). Since that issue had no Linux test result, I tried this fix on my machine.

Short version: **4 crashes before the fix, 0 after**, with comparable navigation and more than twice the interface switching.

### How I tested

The PR branch is 93 commits behind `main`, so testing it as-is would have compared today's app against an older one. Instead I cherry-picked the single commit (`1d13dd72`) onto current `main` (`35da72b6c78491227fd5dbe3eaed2b8b0599f045`) so the only difference between the two runs is this change.

Both runs used `cd frontend && npm run dev` with `ELECTRON_ENABLE_LOGGING=1`, so renderer console output lands in the terminal log. I then counted:

- crash lines matching `reading 'dimensions'`
- `GET /api/v1/sessions` — a proxy for how often a session view was opened
- `POST /api/v1/sessions/<id>/interface-transition` — actual Chat/Terminal switches
- `[vite] connecting` — full renderer reloads

Same interactions in both runs: navigating between the board and sessions, switching between an orchestrator and a worker session, switching interfaces, and reloading the app.

### Results

| | Baseline (`main`) | This fix on `main` |
| --- | --- | --- |
| `dimensions` crashes | **4** | **0** |
| Session route loads | 84 | 65 |
| Interface-transition POSTs | 3 | 7 |
| Renderer reloads | 6 | 4 |

The baseline crashed about once every 21 session route loads. Over 65 route loads, an unfixed build would be expected to produce roughly 3 crashes. There were none. Treating it as a Poisson process, the chance of seeing zero by luck is about 1 in 20.

The fixed run also exercised the interface-transition path more than twice as hard as the baseline (7 POSTs vs 3), which is the path this change targets.

### What this does not show

- This is evidence of non-reproduction, not proof of correctness. It is a race, and ~1 in 20 is not conclusive.
- The heavy-interaction window had no full renderer reloads (4 earlier in the same run). Reload was involved in one of the four baseline crashes, so that path is more lightly covered.
- I tested the commit cherry-picked onto current `main`, not the PR branch as it stands.

Environment: Electron 33.4.11, node 26.2.0, `@xterm/xterm` 5.5.0, mesa 26.1.5, hybrid Intel + NVIDIA GPU.

One note on reading the logs from this machine: it also throws GPU process crashes and WebGL errors on startup, which look related but are not. That is a separate driver problem I filed as [#3961](https://github.com/Untrivial-ai/agent-orchestrator/issues/3961). It was present in both runs, so it does not affect the comparison.

i-trytoohard:
@maplenk Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@maplenk
Closing this PR as stale after 30+ days without author activity. The linked [#3803](https://github.com/Untrivial-ai/agent-orchestrator/issues/3803) remains open if this fix is still needed.

PR #3806, opened 2026-08-10
fix(daemon): enforce exclusive data-dir ownership

### LINKED ISSUE #3716: Run Review fails after a second daemon hot-migrates the shared SQLite database
## Bug

Clicking **Run review** can fail with `Review operation failed (REVIEW_OPERATION_FAILED)` when a newer daemon has used the same `AO_DATA_DIR` and migrated SQLite while an older daemon remains alive. The failure occurs before AO creates either a `review` or `review_run` row.

**Source:** local AO desktop report | **Reported by:** @illegalcall | **Date:** 2026-08-08  
**Analyzed against:** `f7e0fe65b` (`origin/main`)  
**Running desktop daemon:** `fc5f4ca0219c2b9e11d2982d34598e3d80320d77` with `vcs.modified=true`, `ao version dev`  
**Confidence:** High — the live binary's embedded SQL, the applied schema, migration/daemon timestamps, and SQLite's exact prepare-time failure all line up.

Run Review shows REVIEW_OPERATION_FAILED

## Environment

- macOS / Darwin 25.5.0, arm64
- Primary daemon healthy on `127.0.0.1:3001`, started `2026-08-07T15:11:21Z`
- Claude Code `2.1.225`
- A second daemon was recorded against the same database on port `3012` at `2026-08-08T06:56:16Z`
- Migration `0080_review_per_harness.sql` was recorded as applied at that exact timestamp
- Affected worker: `agent-orchestrator-1-70`, PR [#3711](https://github.com/Untrivial-ai/agent-orchestrator/pull/3711), selected reviewer `claude-code`

## Reproduction

Use a temporary data directory; do not test this against production data.

1. Start a daemon built before migration 0080 with a temporary `AO_DATA_DIR`, port, and run file.
2. Start a current daemon with the **same** `AO_DATA_DIR` but a different `AO_PORT` and `AO_RUN_FILE`.
3. Let the current daemon apply migration 0080, then stop it while leaving the old daemon alive.
4. In the UI served by the old daemon, select a worker with an open PR and click **Run review**.
5. Observe HTTP 500 / `REVIEW_OPERATION_FAILED`; no review row or run is persisted.

Observed locally without issuing another review:

- `GET /api/v1/sessions/agent-orchestrator-1-70/reviews` returns the PR as `needs_review`.
- The running daemon binary contains:

  ```sql
  INSERT INTO review (...)
  VALUES (...)
  ON CONFLICT (session_id) DO UPDATE SET ...
  ```

- The migrated table now has `UNIQUE(session_id, harness)` and no unique constraint on `session_id` alone.
- Preparing the old statement against the live DB fails deterministically with:

  ```text
  ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
  ```

## Root Cause

`backend/internal/daemon/daemon.go:84-101` guards daemon ownership through the configured run file, then opens/migrates `cfg.DataDir`. A second daemon with a different run file/port but the same data directory can therefore pass the ownership check and mutate the schema underneath the first daemon.

This configuration is reachable in the documented desktop real-data workflow: `.agents/skills/ao-desktop-dev/SKILL.md:14-16,56-65` explicitly runs a separate dev daemon on its own port/run file while pointing `AO_DATA_DIR` at the real database.

Migration `backend/internal/storage/sqlite/migrations/0080_review_per_harness.sql:13-24` rebuilds `review` with `UNIQUE(session_id, harness)`. Current generated SQL correctly targets that pair (`backend/internal/storage/sqlite/queries/review.sql:1-9`), but a live pre-migration daemon continues using `ON CONFLICT (session_id)`. SQLite rejects that statement during review upsert, before a run is created.

Finally, `backend/internal/httpd/controllers/reviews.go:352-361` turns every unexpected service/storage error into the generic `REVIEW_OPERATION_FAILED` response, so the UI (`frontend/src/renderer/components/SessionInspector.tsx:1187-1196`) has no actionable recovery information.

## Fix

- Acquire and retain an exclusive, **data-directory-scoped** daemon lock before opening or migrating SQLite. The lock must be independent of `AO_PORT` and `AO_RUN_FILE` so two daemons cannot own one database.
- Make the real-data desktop-dev path attach to/restart the owning daemon or refuse to launch against a live data directory; do not run a second migrator over it.
- Add a regression test with two daemon configs sharing `AO_DATA_DIR` but using distinct run files/ports.
- As defense in depth, detect an externally advanced/incompatible schema and stop serving with an actionable restart message. Log the internal review error with the request ID even if the API response remains generic.

## Impact

- The code-review workflow is blocked until the stale daemon is restarted with a schema-compatible binary.
- The same mechanism can break any old query invalidated by a newer migration, so the risk is broader than reviews.
- **Priority: P2** — significant degradation with a restart/workflow-isolation workaround; bounded here to mixed-version daemons sharing one data directory.

## Related

- [#2481](https://github.com/Untrivial-ai/agent-orchestrator/issues/2481) — concurrent checkout identity errors are masked in the desktop UI; related surface, but that issue assumes the second daemon is rejected.
- [#2845](https://github.com/Untrivial-ai/agent-orchestrator/issues/2845) — prior dev-daemon port/data isolation decision.
- [PR #3384](https://github.com/Untrivial-ai/agent-orchestrator/pull/3384) — current interactive-reviewer merge containing migration 0080 on `main`.

### LINKED ISSUE #3805: Concurrent AO daemons can reconcile the same data directory
## Problem

Two AO daemon processes can currently enter startup against the same `AO_DATA_DIR`. The run-file check is diagnostic, not an ownership primitive, so concurrent startup can open the same SQLite database and perform reconciliation at the same time.

PID liveness is not sufficient for mutual exclusion: PIDs can be stale or reused, and the authoritative ownership signal must be released by the operating system when a process exits.

## Expected behavior

The daemon should acquire one non-blocking, process-lifetime lease for its data directory before opening storage or running startup reconciliation.

- A second daemon using the same data directory fails immediately with a clear diagnostic.
- Different data directories remain independent.
- Crash or process exit releases the lease through OS semantics.
- The lock file contains peer PID metadata for diagnostics only; ownership never depends on trusting that PID.
- Unix and Windows builds use their native locking primitives.

## Proposed focused fix

- Add a small cross-platform `internal/datadirlock` package using `flock` on Unix and `LockFileEx` on Windows.
- Acquire `AO_DATA_DIR/daemon.lock` before store open/reconcile and hold it for daemon process lifetime.
- Keep existing HTTP fallback behavior, documenting that the data-dir lease is the authoritative cross-process gate.

The prepared patch is schema-free and limited to daemon wiring, the lock package, focused tests, and HTTP fallback comments.

## Validation prepared

- Repeated focused lock tests
- Lock race tests
- Daemon and HTTP server suites
- Full serial backend test suite
- Native build and vet
- golangci-lint v2.12.2
- Linux amd64 and Windows amd64 cross-compilation
- `git diff --check`

maplenk:
## Summary

- acquire one OS-backed, process-lifetime lease per `AO_DATA_DIR` before SQLite is opened or migrated
- reject a second daemon using the same data directory even when its port and run file differ
- keep the existing same-run-file diagnostic and make data-dir conflicts actionable
- correct the desktop development workflow so real data is never shared by concurrent daemons

## Why

Ports and run files identify a daemon; they do not isolate SQLite. Two local AO versions using one `AO_DATA_DIR` can both pass their run-file checks, then a newer version can migrate the database while the older daemon is still running. The older process subsequently executes SQL against a schema it no longer understands.

With this change, parallel checkouts remain supported by giving each daemon a different `AO_DATA_DIR`. A second owner of the same directory fails before storage opens, so it cannot migrate or reconcile shared state.

Fixes #3716
Fixes #3805

## Implementation

- add a small cross-platform `internal/datadirlock` package
- use non-blocking `flock` on Unix and `LockFileEx` on Windows
- store the owning PID for diagnostics while treating the OS lock as authoritative
- hold the lease for the full daemon lifetime and release it on shutdown
- probe the conventional sibling `running.json` so a current build also rejects a data directory still owned by a live pre-lock AO release

This is schema-free and does not change the API.

## Validation

- `go test -p 1 ./...`
- `go test ./internal/datadirlock ./internal/daemon ./internal/httpd/...`
- `go test -race ./internal/daemon ./internal/datadirlock`
- `go build ./...`
- `go vet ./...`
- golangci-lint v2.12.2 with an isolated cache: 0 issues
- Linux amd64 and Windows amd64 `datadirlock` cross-compilation
- live two-daemon check with one temporary data directory and different ports/run files: second daemon rejected before creating its run file; acquisition succeeded after the first stopped
- native Electron verification with two simultaneous isolated instances and a third shared-data rejection; screenshots are in the PR comment
- mixed-version verification: a current daemon rejected data owned by an `origin/main` pre-lock daemon, which remained healthy
- AO desktop development skill validation
- `git diff --check`

The highly parallel `npm run lint` test phase hit existing timing flakes in the fake, kilocode, and opencode adapter packages. All three passed on their focused serial rerun, and the full serial backend suite above passed.

### DIFF
--- a/.agents/skills/ao-desktop-dev/SKILL.md
+++ b/.agents/skills/ao-desktop-dev/SKILL.md
@@ -12,7 +12,7 @@ Run the Electron application from the current checkout and verify it in the nati
 Ask only when the request does not make the desired data source clear.
 
 - Use **isolated mode** by default for implementation and destructive testing. Electron uses port `3002`, `~/.ao/dev/running.json`, `~/.ao/dev/data`, and `~/.ao/dev/electron`.
-- Use **real-data mode** only when the user explicitly asks to see this machine's actual AO projects or sessions. Start the checkout's dev daemon on the isolated dev port/run file while pointing `AO_DATA_DIR` at the real AO data directory. This is a separate daemon process using real data; do not describe it as the installed app's daemon.
+- Use **real-data mode** only when the user explicitly asks to see this machine's actual AO projects or sessions. The real data directory has one daemon owner: stop that owner before starting the checkout's daemon, and never run both against it. A newer checkout may migrate the database, so prefer isolated mode for backend or migration testing.
 - Never try to attach an unpackaged Electron app directly to a packaged daemon from another checkout. The supervisor intentionally rejects daemon identity mismatches.
 - Warn before actions in real-data mode that create, terminate, rename, or otherwise mutate sessions. Merely opening and inspecting the UI is expected.
 
@@ -57,7 +57,9 @@ npm run dev
 
 First resolve and report the actual data directory. In an AO worker session, `AO_DATA_DIR` normally already identifies it. Otherwise the repository default is the absolute path corresponding to `~/.ao/data`.
 
-Keep the real data directory but remove inherited port/run-file overrides so the dev app retains its own daemon handshake and port:
+Check its normal run file and health endpoint for a live owner. If one exists, report that stopping it will interrupt its desktop connection and active work, then obtain the user's authorization to stop that exact daemon. Proceed only after its process and listener are gone.
+
+With exclusive ownership established, keep the real data directory but remove inherited port/run-file overrides so the dev app uses its own daemon handshake and port:
 
 ```bash
 cd frontend
@@ -73,6 +75,8 @@ Remove-Item Env:AO_RUN_FILE, Env:AO_PORT -ErrorAction SilentlyContinue
 $env:AO_DATA_DIR = $realAoData
 npm run dev
 ```
+
+If startup reports that the data directory is already in use, keep the owner running and switch to isolated mode unless the user authorizes stopping it. Changing only the port or run file cannot create a second owner.
 
 Do not print unrelated environment variables: AO sessions may carry credentials. It is safe to report only `AO_DATA_DIR`, `AO_RUN_FILE`, and `AO_PORT`.
 
@@ -129,6 +133,7 @@ Once one PR merges, prefer rebasing the remaining PR onto current `main`; the no
 - Renderer URLs can move from `5173` when a port is occupied. Trust Forge's printed URL rather than assuming one.
 - Multiple dev instances share `~/.ao/dev/electron`; avoid running them concurrently because Chromium profile state can collide.
 - An inherited `AO_DATA_DIR` changes dev mode from isolated data to real data. Always choose and report the mode instead of inheriting it accidentally.
+- A different `AO_PORT` or `AO_RUN_FILE` does not isolate SQLite. Every concurrent daemon needs a different `AO_DATA_DIR`.
 - Repeated `/healthz/` 404 entries from external probes can be noisy; readiness is determined by Electron's daemon status and successful API traffic, not by log volume.
 
 ## Completion report
--- a/backend/internal/daemon/daemon.go
+++ b/backend/internal/daemon/daemon.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/browserruntime"
 	"github.com/aoagents/agent-orchestrator/backend/internal/config"
 	"github.com/aoagents/agent-orchestrator/backend/internal/daemon/supervisor"
+	"github.com/aoagents/agent-orchestrator/backend/internal/datadirlock"
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 	"github.com/aoagents/agent-orchestrator/backend/internal/httpd"
 	"github.com/aoagents/agent-orchestrator/backend/internal/httpd/controllers"
@@ -64,6 +65,17 @@ func Run() error {
 	ignoreBrokenPipeSignal()
 
 	log := newLogger()
+
+	// Acquire durable mutation authority before opening the store or running any
+	// reconciliation. Without this process-lifetime lease, concurrent starts can
+	// both pass the run-file probe, bind different ports, and mutate the same DB.
+	dataLease, err := datadirlock.Acquire(cfg.DataDir)
+	if err != nil {
+		return fmt.Errorf("acquire data-dir ownership: %w", err)
+	}
+	defer func() { _ = dataLease.Close() }()
+	log.Info("data-dir ownership acquired", "lock", dataLease.Path(), "pid", dataLease.PID())
+
 	var browserRuntimeToken string
 	if os.Getenv(browserruntime.RuntimeTokenStdinEnv) == "1" {
 		browserRuntimeToken, err = browserruntime.ReadRuntimeToken(os.Stdin)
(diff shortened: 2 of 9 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
## Local native-Electron verification

Verified head `aa482b92b` on macOS arm64 using two real AO Electron apps launched from this checkout. Every test profile, data directory, and run file was isolated under a temporary root; the installed AO daemon on port 3001 was left running and remained ready throughout.

### Two isolated AO instances in parallel — PASS

Both screenshots below were captured during the same concurrent run:

| Instance | Electron PID | Daemon PID | API port | Data directory | Result |
| --- | ---: | ---: | ---: | --- | --- |
| A | 51410 | 51492 | 43716 | `<temp>/a/data` | `/healthz`: `ok`; `/readyz`: `ready` |
| B | 54658 | 54739 | 43806 | `<temp>/b/data` | `/healthz`: `ok`; `/readyz`: `ready` |

`lsof` also showed each daemon holding its own `<data-dir>/daemon.lock`. The distinct `A 43716 DATA-A` and `B 43806 DATA-B` project names in the UIs make the isolated backing stores visible.

**Instance A**

AO instance A running on port 43716 with isolated DATA-A

**Instance B**

AO instance B running on port 43806 with isolated DATA-B

### Accidental shared data directory — PASS

While A and B remained ready, I launched a third real Electron app with a different port and run file but pointed it at A's `AO_DATA_DIR`. Its daemon exited before writing its run file with:

```text
data directory "<temp>/a/data" is already in use by daemon pid 51492 on port 43716; stop it or choose a different AO_DATA_DIR
```

A and B remained ready after the rejection.

Third AO instance reports daemon failed to start when sharing instance A's data directory

### Mixed-version upgrade path — PASS

I also built `origin/main` at `970bdea50` (before `daemon.lock`), ran it with the conventional sibling `running.json`, and then launched this PR's daemon against the same data directory with a different port/run file. The PR daemon rejected the live pre-lock owner, did not create its run file, and the old daemon remained ready. This exact transition test exposed a gap in the earlier PR head; commit `aa482b92b` adds the small conventional-run-file compatibility probe that closes it.

### Automated checks

- `go test -p 1 ./...`
- `go test -race ./internal/daemon ./internal/datadirlock`
- `go build ./...`
- `go vet ./...`
- `git diff --check`

Conclusion: parallel AO development works when each checkout has its own `AO_DATA_DIR`, and accidental data sharing now fails before SQLite opens or migrates—including when the existing owner is a live AO release from before the lock was introduced.

i-trytoohard:
@maplenk Marking as stale: no activity from you for 28 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

Rishet11:
@maplenk
Closing this PR as stale after 30+ days without author activity. The linked [#3716](https://github.com/Untrivial-ai/agent-orchestrator/issues/3716) remains open if this fix is still needed.

PR #3812, opened 2026-08-10
fix: overlay archive panel so board lanes stay stable

Annieeeee11:
## Summary
- Overlay the expanded archive on the board (absolute bottom) so lane height stays stable and Needs You no longer gets a persistent scrollbar gutter.
- Keep archive cards mounted after first open and animate a capped pixel height instead of remounting / measuring `height: auto`.
- Make the archive control a full-width hit target, drop uppercase label styling, and hide the archive list scrollbar.

## Test plan
- [ ] Open a project board with archived sessions; confirm collapsed archive is a 46px bottom bar and lanes fill the rest.
- [ ] Expand archive — panel overlays the board; Needs You / other lane scrollbars do not appear or disappear solely from the toggle.
- [ ] Toggle archive repeatedly with a large archive set; open/close should feel snappy (cards stay mounted).
- [ ] Click empty space beside the Archive label — still toggles.
- [ ] With reduced motion enabled, expand/collapse should not animate.
- [ ] `cd frontend && npm test -- SessionsBoard.test.tsx`


Annieeeee11:
Addressed both notes (latest: `21cbe8c`):

dropped `font-mono` from the Archive label so it uses the UI sans; the count stays mono.

archive lives in its own `BoardArchivePanel`, so expand/collapse does not re-render the kanban lanes. Card mount is deferred with `startTransition` after open is requested; cards stay mounted for later toggles. No `scrollHeight` / measured-height Motion tween.

the sheet opens back **under** the Archive row (CSS `grid-rows` `0fr→1fr`), not as a transform/opacity overlay above the bar. That earlier approach was wrong and has been reverted.

ronishrohan:
No transition right now :/ apparently it was intentionally removed in the last commit, so something is wrong

frontend/src/renderer/components/SessionsBoard.tsx
```diff
@@ -387,57 +347,185 @@ export function SessionsBoard({ projectId }: SessionsBoardProps) {
...
+						viewBox="0 0 24 24"
+					>
+						<path d="m9 18 6-6-6-6" />
+					</svg>
+					<span className="text-2xs font-medium tracking-wide-sm">{t("shell.archive")}</span>
+					<span className="ml-1.5 font-mono text-micro text-passive">{sessions.length}</span>
+				</button>
+				{/* Disclosure sheet under the Archive row — Motion height 0↔auto. */}
+				<AnimatePresence initial={false}>
+					{expanded && cardsReady ? (
```
ronishrohan: `expanded` becoming false removes this subtree from `AnimatePresence`; after the 140ms exit it unmounts the list and every `ArchiveSessionItem`. That means reopening the archive remounts all cards, which contradicts the stated goal of keeping them mounted after the first open and reintroduces the expensive work this PR is trying to avoid for large archives.

Keep the sheet mounted once `cardsReady` is true and animate its height from `expanded` instead (while disabling interaction/accessibility as appropriate when collapsed).

Annieeeee11: Good catch, fixed once `cardsReady` is true the sheet stays mounted, `expanded` only drives the height `0 -> auto` tween. Collapsed state uses `inert`, `aria-hidden`, and `pointer-events-none`. Added a regression test that asserts the same list/card nodes survive collapse -> reopen.

ronishrohan (commented):
## Review

The overlay approach is correct and prior review feedback (AnimatePresence unmounting) is verified fixed — `AnimatePresence` is no longer used at all. Cards stay mounted via the `cardsReady` latch; `expanded` only drives `height: 0 -> "auto"`. Accessibility is thorough (aria-expanded, aria-hidden, inert).

### Issues to address

**1. Merge conflict** — `packages/product-ui/src/SessionsBoardView.tsx` conflicts with main's `statusPresentation` pattern from #3940. Should be low-difficulty to resolve (take main's changes in `SessionCardView`).

**2. Commit history** — 7 commits with duplicate messages (`"fix: overlay archive panel so board lanes stay stable"` appears twice). Recommend squash-merge.

**3. Hardcoded height coupling** — `pb-[58px]` on board content must match `h-[58px]` on the toggle bar. If either changes independently, the archive toggle will overlap board content or leave a gap. Consider extracting a shared constant.

### Missing test coverage
- No test for `prefers-reduced-motion` behavior (instant vs animated transitions)
- No test verifying `pb-[58px]` is applied when archive sessions exist

ronishrohan (changes requested):
## Changes Requested

### 1. Resolve merge conflict
`packages/product-ui/src/SessionsBoardView.tsx` conflicts with main's `statusPresentation` pattern from #3940. Take main's changes in `SessionCardView`.

### 2. Hardcoded height coupling
`pb-[58px]` on board content must match `h-[58px]` on the toggle bar. If either changes independently, the archive overlaps or leaves a gap. Extract a shared constant.

### 3. Clean up commit history
7 commits with duplicate messages (`"fix: overlay archive panel so board lanes stay stable"` appears twice). Squash-merge recommended.

### Optional
- Add a `prefers-reduced-motion` test to verify `duration: 0` / `transition-none` behavior.
- Prior review feedback (AnimatePresence unmounting) is verified fixed.

ronishrohan (approved):
## Approved

Final review after #3978 merged — no conflicts with current main.

All previously requested changes verified:
1. **Merge conflict resolved** — clean merge-tree, no markers
2. **Shared height constants** — `ARCHIVE_TOGGLE_HEIGHT_PX`, `archiveToggleHeightClassName`, `archiveToggleOffsetClassName` with lockstep test
3. **Reduced-motion test** — verifies `transition-none` and `duration: 0`
4. **Cards stay mounted** — no AnimatePresence, `cardsReady` latch keeps DOM nodes alive through collapse, test verifies same DOM reference on reopen

Overlay architecture is clean: `absolute inset-x-0 bottom-0 z-20` with `pb-[58px]` offset keeps lane heights stable. Restore logic properly isolated in memo'd `BoardArchivePanel`. Recommend squash-merge.

PR #3817, opened 2026-08-10
fix: enforce AO-only worker delegation in prompts

nikhilachale:
## Summary
- remove guidance that encouraged runtime-native subagent/task-delegation execution in session prompts
- explicitly require orchestrators to coordinate parallel work through AO workers only
- update session-manager prompt tests to lock in AO-only worker delegation wording

## Test plan
- [x] `cd backend && go test ./internal/session_manager/...`


skills/agent-role-smoke/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+cd frontend
+npm run dev
+```
+
+This starts both the daemon and Electron supervisor. Once the app is running, start testing the selected orchestrator or worker behavior directly. Do not run API generation, lint, typecheck, or broad checks by default; assume those are fine unless the user explicitly asks for them or the dev app/manual test exposes a problem.
+
+From the repo root, in another terminal:
+
+```bash
+node skills/agent-role-smoke/scripts/agent-role-smoke.mjs
```
whoisasx: [P2] This skill cannot run at the PR head: every documented command points to `skills/agent-role-smoke/scripts/agent-role-smoke.mjs`, but the latest commit deletes that script, so the Quick Start fails immediately with `MODULE_NOT_FOUND`. The skill is also outside this PR's stated prompt-policy scope. Please remove this leftover skill from the branch, or restore and validate the complete skill in a separate focused change.

backend/internal/session_manager/prompt.go
```diff
@@ -259,7 +261,8 @@ Your job is to complete the assigned task in this workspace. Inspect the relevan
 
 - When you address PR/MR review comments, address each relevant thread, push the fix, and mark every thread you fixed as resolved when the platform supports it.
 - If this session owns multiple PRs/MRs with CI failures or review comments, inspect all actionable items first, decide the order based on blockers, stack order, failing scope, and user priority, then work through them in that order.
-- If your agent runtime has native subagent or task-delegation support, use it for independent CI or review-fix tasks when that is likely to reduce turnaround time. Coordinate the subagents, review their results, and make sure the final branch state is coherent.
+- Do not use runtime-relative sub-agents or task-delegation workers. Complete the assigned task in this AO session only.
+- If parallel help is needed for CI or review follow-up, ask the orchestrator to spawn additional AO worker sessions instead of delegating inside the runtime.
```
whoisasx: [P2] This instruction is emitted for every worker, including directly spawned workers with no active orchestrator. In that path `workerOrchestratorPrompt` is not appended, so a worker needing parallel help is told to contact a session that does not exist. Make this handoff guidance conditional on `OrchestratorSessionID`, and tell standalone workers to continue serially or report the need for parallel work to the human; add coverage for both prompt variants.

whoisasx (changes requested):
The AO-only delegation direction is reasonable, and the focused prompt tests are green, but this branch is not ready to merge. It adds a discoverable smoke-test skill whose referenced script was deleted, and standalone workers are instructed to contact an orchestrator even when none is attached. The branch also contains 24 commits from an older topic branch; please rebase or cherry-pick the prompt fix onto current `main` so the unrelated half-deleted skill and historical contamination are removed. Since this policy is enforced entirely through prompt text, please also replace the undefined `runtime-relative` wording with an explicit reference to the agent runtime's built-in subagent/task-delegation tools.

backend/internal/session_manager/prompt.go
```diff
@@ -259,12 +266,13 @@ Your job is to complete the assigned task in this workspace. Inspect the relevan
...
 - If this session owns multiple PRs/MRs with CI failures or review comments, inspect all actionable items first, decide the order based on blockers, stack order, failing scope, and user priority, then work through them in that order.
-- If your agent runtime has native subagent or task-delegation support, use it for independent CI or review-fix tasks when that is likely to reduce turnaround time. Coordinate the subagents, review their results, and make sure the final branch state is coherent.
+- Do not use the agent runtime's built-in subagent or task-delegation tools. Complete the assigned task in this AO session only.
+- %s
 - For complex tasks, write a short implementation plan before editing. Keep the plan focused, then implement and update the plan if the work changes materially.
 
 %s
 
-%s`, taskSourceRules, repoRules, projectContextSection(project))
+%s`, taskSourceRules, repoRules, parallelHelpRules, projectContextSection(project))
```
whoisasx: [P2] The new `- %s` placeholder receives the second argument, `repoRules`, while `parallelHelpRules` is passed third and lands in the later standalone `%s` slot. As a result every worker prompt renders `- ## Git and PR/MR Rules` or `- ## Local Git Rules`, and the parallel-help bullets appear outside their intended position. Reorder these arguments to `taskSourceRules, parallelHelpRules, repoRules, projectContextSection(project)`, and add rendered-output assertions for both orchestrator states that reject repository headings prefixed with a list marker.

whoisasx (changes requested):
The previous functional feedback is addressed: the broken smoke-test skill is removed, standalone-worker guidance is now conditional, and the delegation terminology is explicit. However, the follow-up introduced a generated-prompt regression because the new `fmt.Sprintf` placeholder and argument order do not align. Existing tests and all CI checks pass because they assert only substrings rather than the rendered section structure. Please fix the argument order and add rendered-output coverage for both standalone and orchestrator-attached workers. The branch also still carries 26 historical commits and should be rebased or cherry-picked onto current `main` as previously requested.

whoisasx (approved):
All requested changes are addressed. The unrelated broken smoke skill is removed, standalone and orchestrator-attached worker guidance is handled separately, the delegation terminology is explicit, and the prompt placeholder ordering now preserves the repository-rule section structure. Focused session-manager tests pass locally, `git diff --check` is clean, and all GitHub checks are green. Looks good to merge.

PR #3818, opened 2026-08-10
fix: reconcile review runs after provider submit (#3755)

ApexYash11:
## What

  Recover review runs when a reviewer successfully posts its GitHub review but cannot execute `ao review submit`.

  ## Why

  Fixes #3755.

  A provider review could be posted successfully while AO’s review run remained stuck in `running`.

  ## How

  - Add a versioned run/SHA/verdict marker to GitHub review bodies.
  - Observe `COMMENTED` reviews and numeric GitHub review IDs.
  - Validate and reconcile provider reviews through the existing idempotent `Submit` path.
  - Continue polling for AO comment reviews even when GitHub’s review decision is unchanged.
  - Cancel 30-minute stale runs only when the reviewer is missing or confirmed dead.
  - Keep stale cleanup active when SCM polling is disabled.
  - Fail reviewer launch when the AO executable cannot be added to `PATH`.

  No frontend or API contract changes.

  ## Testing

  - Added regression tests for marker parsing and provider reconciliation.
  - Added stale-run liveness tests.
  - Added GitHub `COMMENTED` review observation tests.
  - Added launcher PATH failure tests.
  - Ran focused tests for all changed backend packages.
  - Ran `go vet` on the changed packages.
  - Regenerated and verified sqlc output.

  ## Checklist

  - [x] Branched from `main`
  - [x] One focused issue
  - [x] Tests added for the changed behavior
  - [x] Relevant backend checks pass

PR #3819, opened 2026-08-10
fix: make ao available in reviewer launches

### LINKED ISSUE #3755: bug(review): reviewer posts GitHub review but run stays stuck in 'running' forever when ao binary is missing
## Summary

When the AO reviewer agent successfully posts a code review to GitHub but cannot run `ao review submit` (because the `ao` binary is missing from the reviewer environment), the review run stays in `running` status indefinitely. The dashboard permanently shows **"review in progress"** even though the review is complete and visible on the PR.

## How the reviewer workflow works (2 steps)

The reviewer prompt (`backend/internal/review/prompt.go:25-43`) instructs the reviewer agent to do two things in order:

1. **Post the review to GitHub** via `gh api --method POST repos/.../pulls/.../reviews` → captures the review ID
2. **Record the verdict in AO** via `ao review submit --session <id> --reviews -` → updates the review run from `running` → `complete`

Step 1 and step 2 are independent. If step 2 fails, step 1 has already succeeded — the review is on GitHub, but AO never learns about it.

## What goes wrong

When the `ao` binary is not on the reviewer agent's PATH, step 2 fails:

```
AO bookkeeping submission failed: ao binary not found in environment
```

The review run record in the database is never updated. Its status stays `ReviewRunRunning` with `VerdictNone`.

## Why the dashboard shows "review in progress" forever

The review planning function `Plan()` in `backend/internal/review/planner.go:61-75` derives the per-PR review state from the run record:

```go
if run, ok := latest[review.PRURL+"\x00"+review.TargetSHA]; ok {
    review.LatestRun = &run
    switch {
    case run.Status == domain.ReviewRunRunning:
        review.Status = ReviewStateRunning  // ← stuck here forever
```

Since the run was never transitioned out of `running`, the dashboard permanently shows the review as in-progress.

## State machine

```
ReviewRunRunning ──(ao review submit succeeds)──▶ ReviewRunComplete ──▶ ReviewRunDelivered
      │
      │  ✗ ao binary missing — submit never called
      └──────────▶ STUCK (dashboard: "review in progress" forever)
```

## Impact

- **Core workflow degraded:** A completed review (approved or changes-requested) is invisible to AO. The dashboard, lifecycle reactions, and merge readiness checks never see the verdict.
- **No automatic recovery:** There is no timeout, no reconciliation, and no fallback path. The run stays `running` until someone manually intervenes.
- **Silent failure from the user's perspective:** The reviewer agent reports "No issues found. Ready to merge." and exits. The only sign of trouble is the dashboard badge never changing.

## Workaround

Cancel the stuck run manually:

```bash
ao review cancel <worker-session-id>
```

Or manually record the verdict:

```bash
ao review submit <worker-session-id> --run <run-id> --verdict approved
```

## Potential fixes

1. **Ensure `ao` binary is available in the reviewer environment** — the reviewer agent needs it to complete step 2.
2. **Add reconciliation via the SCM observer** — the observer already polls GitHub for PR state (`backend/internal/observe/scm/observer.go`). It could detect a posted review on the PR (matching the run's target SHA and reviewer harness) and reconcile the run status from `running` → `complete`, making step 2 best-effort rather than load-bearing.
3. **Add a timeout on running review runs** — if a run has been `running` for longer than a threshold with the reviewer pane no longer alive, mark it as `failed` or `needs_review` so it does not block indefinitely.

## Key code locations

| File | What |
|------|------|
| `backend/internal/review/prompt.go:25-43` | Reviewer prompt — 2-step workflow (GitHub post, then `ao review submit`) |
| `backend/internal/service/review/review.go:271-279` | `submitOne()` — transitions run `running` → `complete` (never called when binary missing) |
| `backend/internal/review/planner.go:61-75` | `Plan()` — maps `ReviewRunRunning` → `ReviewStateRunning` (stuck display) |
| `backend/internal/review/review.go:407-453` | `Cancel()` — workaround path to clear stuck runs |

## Reproduction

1. Spawn a worker session with a PR
2. Trigger a review (`ao review trigger <session>`)
3. Ensure `ao` is NOT on the reviewer agent's PATH (or simulate the failure)
4. Reviewer posts the review to GitHub successfully (step 1) but `ao review submit` fails (step 2)
5. Dashboard shows "review in progress" indefinitely

## Reported by

@Nikhil Achale — observed on a local AO setup reviewing PR #8.

neversettle17-101:
## Summary
- apply the same `sessionmanager.HookPATH` AO path pinning used for worker agent launches to reviewer launches, so `ao review submit` resolves inside reviewer panes
- propagate the resolved `AO_RUN_FILE` into reviewer launch env and Codex filtered shell configuration so reviewer-local AO CLI calls discover the live daemon even when the harness/sandbox does not inherit the default runfile location
- create an AO-owned `ao` shim under the data dir when the running executable is not named `ao`, and fail launch with actionable context if neither direct PATH pinning nor shim fallback can guarantee `ao`
- cover spawn, restore, Codex filtered shell env, and fallback failure paths with regression tests

Closes #3755

## Tests
- `cd backend && go test ./internal/review ./internal/adapters/reviewer/codex ./internal/daemon ./internal/service/review ./internal/httpd/controllers`
- `cd backend && go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs ./internal/review ./internal/adapters/reviewer/codex ./internal/daemon ./internal/ports`

## Notes
- Kept this to reviewer launch reliability: correct AO binary plus correct daemon runfile. SCM observer reconciliation is intentionally not included in this PR.

### DIFF
--- a/backend/internal/review/launcher.go
+++ b/backend/internal/review/launcher.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
@@ -19,6 +20,14 @@ const cancelInterruptDelay = 150 * time.Millisecond
 const defaultReviewerInitialDelay = 500 * time.Millisecond
 
 const reviewerTaskMessagePrefix = "Read and follow the AO review task in `"
+
+// EnvRunFile points reviewer-local AO CLI calls at the live daemon run-file.
+const EnvRunFile = "AO_RUN_FILE"
+
+// EnvAOCommandWarning records why AO could not guarantee a reviewer-local `ao`
+// command. Review launch remains best-effort so provider reviews can still be
+// posted and later reconciled, but this makes the degraded path diagnosable.
+const EnvAOCommandWarning = "AO_REVIEW_AO_COMMAND_WARNING"
 
 // Launcher spawns, re-notifies, and probes a reviewer over a worker's worktree.
 // It is the side of the engine that talks to the reviewer registry and runtime;
@@ -90,10 +99,12 @@ type reviewerRuntime interface {
 // runtime. The reviewer reuses the worker's worktree (a fresh session worktree
 // would branch off the default branch and so would not contain the PR changes).
 type agentLauncher struct {
-	reviewers ports.ReviewerResolver
-	runtime   reviewerRuntime
-	dataDir   string
-	auth      agentAuthResolver
+	reviewers  ports.ReviewerResolver
+	runtime    reviewerRuntime
+	dataDir    string
+	runFile    string
+	auth       agentAuthResolver
+	executable func() (string, error)
 }
 
 type preLaunchReviewer interface {
@@ -120,9 +131,27 @@ func WithAgentAuth(auth agentAuthResolver) LauncherOption {
 	}
 }
 
+// WithExecutable overrides os.Executable for reviewer PATH pinning. Production
+// leaves this unset; tests use it to model the daemon binary that should make a
+// bare `ao review submit` available inside reviewer panes.
+func WithExecutable(executable func() (string, error)) LauncherOption {
+	return func(l *agentLauncher) {
+		l.executable = executable
+	}
+}
+
+// WithRunFilePath pins reviewer-local AO CLI calls to this daemon's run-file.
+// This is intentionally separate from AO_DATA_DIR: the CLI discovers the live
+// daemon from AO_RUN_FILE, not from the durable data directory.
+func WithRunFilePath(path string) LauncherOption {
+	return func(l *agentLauncher) {
+		l.runFile = path
+	}
+}
+
 // NewLauncher builds the production reviewer launcher.
-func NewLauncher(reviewers ports.ReviewerResolver, runtime reviewerRuntime, dataDir string, opts ...LauncherOption) Launcher {
-	l := &agentLauncher{reviewers: reviewers, runtime: runtime, dataDir: dataDir}
+func NewLauncher(reviewers ports.ReviewerResolver, rt reviewerRuntime, dataDir string, opts ...LauncherOption) Launcher {
+	l := &agentLauncher{reviewers: reviewers, runtime: rt, dataDir: dataDir, executable: os.Executable}
 	for _, opt := range opts {
 		opt(l)
 	}
@@ -219,6 +248,7 @@ func (l *agentLauncher) invocation(spec LaunchSpec) ports.ReviewInvocation {
 		ReviewIndex:     spec.ReviewIndex,
 		WorkspacePath:   spec.WorkspacePath,
 		DataDir:         l.dataDir,
+		RunFilePath:     l.runFile,
 		Prompt:          prompt,
 		SystemPrompt:    systemPrompt,
 	}
@@ -292,6 +322,7 @@ func (l *agentLauncher) prepareIdleInvocation(spec LaunchSpec) (ports.ReviewInvo
 		AgentSessionID:   spec.AgentSessionID,
 		WorkspacePath:    spec.WorkspacePath,
 		DataDir:          l.dataDir,
+		RunFilePath:      l.runFile,
 		Prompt:           prompt,
 		SystemPrompt:     "",
 		SystemPromptFile: systemPath,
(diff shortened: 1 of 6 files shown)


backend/internal/review/launcher.go
```diff
@@ -497,14 +521,68 @@ func (l *agentLauncher) runtimeEnv(ctx context.Context, spec LaunchSpec, argv []
...
 	env[sessionmanager.EnvProjectID] = string(spec.ProjectID)
 	env[sessionmanager.EnvDataDir] = l.dataDir
-	path, err := sessionmanager.HookPATH(os.Executable, os.Getenv, env)
+	if strings.TrimSpace(l.runFile) != "" {
+		env[EnvRunFile] = l.runFile
+	}
+	path, err := sessionmanager.HookPATH(l.executable, os.Getenv, env)
 	if err == nil {
 		env["PATH"] = path
+	} else if shimDir, shimErr := l.ensureAOShimDir(); shimErr == nil {
```
neversettle17-101, line 537: **Medium:** This discards `shimErr` and continues launching the reviewer with its original PATH. In exactly the environments this fallback targets (the daemon binary is not named `ao`), a permissions error, read-only data directory, or executable-resolution failure therefore produces a reviewer that cannot reliably run the required `ao review submit`; it may fail late or invoke an unrelated legacy binary. Please make PATH preparation return an error and abort the launch/preflight with the underlying cause (or otherwise surface and handle the failure) rather than silently degrading the core submission path.

neversettle17-101 (commented):
## What This PR Does

Pins reviewer processes to the active AO daemon by injecting `AO_RUN_FILE`, prepending the daemon binary directory to `PATH`, and creating a reviewer-local `ao` shim when the daemon executable has another name. Tests cover environment precedence, direct PATH pinning, shim creation, and restored reviewer sessions.

## Findings

- **Medium:** Shim creation failures are silently ignored. If the fallback cannot resolve or write the shim, the reviewer still launches without a guaranteed `ao`, so the required `ao review submit` command may be missing or may target an unrelated installation. Propagate this error from launch or fail preflight with actionable context.

## Testing

The added unit tests cover successful direct and shim PATH setup, but not the fallback failure path or execution of the generated shim. Tests were not run as part of this review.

backend/internal/review/launcher.go
neversettle17-101: We shouldn't block the launching of the reviewer incase the ao binary isnt resolved. Reviews will still be posted on github. Reconciliation can happen through SCM through a followup PR: https://github.com/Untrivial-ai/agent-orchestrator/pull/3818

neversettle17-101 (commented):
## What This PR Does

Pins reviewer launches to the active daemon run file and makes a bare `ao` command available by prepending the daemon binary directory or an AO-owned shim directory. It also adds launcher coverage for PATH/run-file injection and restore launches.

## Findings

Changes requested. Two launch-path issues still prevent the new guarantee from being reliable, including for the Codex reviewer flow used to submit reviews.

## Regression Risk

The new environment works when the daemon executable is already named `ao`, but custom run-file configurations and shim creation failures can still leave a reviewer unable to submit its result.

## Testing

Unit tests cover PATH order, shim creation, run-file override protection, and restore launches. Existing CI checks are green. Tests were not run locally per review instructions. There is no test that executes a reviewer shell command under Codex environment filtering, or that verifies shim provisioning failures abort the launch.

backend/internal/review/launcher.go
```diff
@@ -497,14 +521,68 @@ func (l *agentLauncher) runtimeEnv(ctx context.Context, spec LaunchSpec, argv []
 	env["AO_REVIEW_HARNESS"] = string(spec.Harness)
 	env[sessionmanager.EnvProjectID] = string(spec.ProjectID)
 	env[sessionmanager.EnvDataDir] = l.dataDir
-	path, err := sessionmanager.HookPATH(os.Executable, os.Getenv, env)
+	if strings.TrimSpace(l.runFile) != "" {
```
neversettle17-101, line 531: **High:** Setting `AO_RUN_FILE` only in the runtime environment does not make it available to Codex-executed review commands. `codexReadOnlyArgs` builds `shell_environment_policy.set.AO_RUN_FILE` earlier from `os.Getenv`, while this value comes from `cfg.RunFilePath` and may not exist (or may differ) in the daemon process environment. Codex intentionally filters shell environments, so in the default/custom-config case `ao review submit` can still discover the wrong run file or no daemon. Please pass this resolved value into the Codex command configuration (and add a test that exercises the filtered shell environment), rather than only adding it after the command is built.

backend/internal/review/launcher.go
neversettle17-101, line 537: **Medium:** Both PATH pinning failures are silently ignored. If the daemon binary is not named `ao` and creating/writing/chmodding the shim fails (read-only/full data directory, permissions, or executable lookup failure), the reviewer still launches without the command this change promises; the failure is only discovered when submission fails at the end of the review. Please return and propagate the combined provisioning error from launch setup, and cover the failure path.

nikhilachale (approved):
Tested locally, works correctly

backend/internal/review/launcher.go
neversettle17-101: Addressed in 5d4272150. Per the maintainer direction in this thread, reviewer launch remains best-effort so GitHub review posting is not blocked. The shim fallback failure is no longer silent: when both HookPATH and shim creation fail, AO preserves the original reviewer PATH and sets AO_REVIEW_AO_COMMAND_WARNING with both underlying causes for diagnostics. Added a regression test for that degraded path.

backend/internal/review/launcher.go
neversettle17-101: Addressed in 5d4272150. The resolved runfile now flows through ports.ReviewInvocation.RunFilePath, and the Codex reviewer uses that invocation value when building shell_environment_policy.set.AO_RUN_FILE instead of relying only on os.Getenv. Added Codex tests that set wrong process env values and verify fresh and restore commands use the launcher-provided data dir/runfile for the filtered shell environment.

backend/internal/review/launcher.go
neversettle17-101: Addressed in 5d4272150 with the same resolution as the earlier PATH/shim thread. Reviewer launch intentionally stays best-effort per maintainer direction so GitHub review posting is not blocked, but the failure is no longer silent: if HookPATH and shim fallback both fail, AO keeps the original reviewer PATH and sets AO_REVIEW_AO_COMMAND_WARNING with both underlying causes. Added a regression test for this degraded path.

PR #3825, opened 2026-08-10
Gate startup on real system requirements (git/tmux/harness/gh) with one-click install

harshitsinghbhandari:
## Summary
- Adds `GET /api/v1/system/requirements`, checking git, tmux (macOS/Linux only — Windows uses ConPTY), at least one installed agent-harness CLI, and gh (advisory, never blocks readiness).
- Adds `POST`/`GET /api/v1/system/install/{target}` for real, async installs against a fixed 6-target allowlist (tmux, gh, claude, codex, opencode, copilot), resolved per-OS via verified package-manager commands (brew / apt/dnf/pacman/zypper / winget / npm). Argv is never built from caller input beyond selecting one of the six known targets; no automatic sudo escalation.
- `DaemonStartupLoader` now runs this check before its existing phrase-rotation display: a staggered checklist of the four requirements, and — only when something required is missing — a non-dismissable dialog with real install actions (including an agent picker across claude/codex/opencode/copilot) that poll the job endpoints to completion, plus a bundled non-blocking gh install nudge. Fails open to the old loader if the probe itself errors.

## Test plan
- [x] `go build ./... && go vet ./...` (backend)
- [x] `go test ./internal/service/systemcheck/... ./internal/service/systeminstall/... ./internal/httpd/... ./internal/daemon/...` (includes `-race` on systeminstall)
- [x] golangci-lint scoped to touched packages: clean
- [x] `npm run api` regenerated `openapi.yaml` and `frontend/src/api/schema.ts`, re-run produces no further diff
- [x] `npm --prefix frontend run typecheck`
- [x] Frontend test suite: 878/878 passing across 63 files
- [ ] Manual pass through the blocked/agent-picker/install-failure states in a running app (not yet exercised live — would require actually uninstalling tmux/a harness CLI)


frontend/src/renderer/components/DaemonStartupLoader.tsx
```diff
@@ -38,17 +79,26 @@ export function DaemonStartupLoader() {
...
+				) : (
+					<p className="mt-2 min-h-5 text-md-sm text-muted-foreground">{t("startup.checkingSetup")}</p>
+				)}
 				<div className="ao-startup-dots mt-3 flex h-4 items-center gap-1.5" aria-hidden="true">
 					<span />
 					<span />
 					<span />
 				</div>
 			</div>
+			{blocked ? (
```
harshitsinghbhandari, line 87: Blocking: this dialog is only mounted while `SessionsBoard.showStartup` is true, and that flag never looks at requirements. Once the daemon reports ready and the workspace query resolves, the whole loader unmounts and this non-dismissable dialog disappears on its own, leaving the board usable with tmux/harness still missing. It also kills any in-flight install poll (the `useEffect(() => stopPolling, [])` cleanup), so a running install silently loses its UI and never triggers the refetch. The blocked state needs to feed into whatever decides to keep the startup screen up.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	if _, err := s.lookPath("curl"); err != nil {
+		return Plan{Target: TargetOpencode, Unsupported: true, Reason: "curl was not found on PATH."}
+	}
+	if _, err := s.lookPath("bash"); err != nil {
+		if _, err := s.lookPath("sh"); err != nil {
+			return Plan{Target: TargetOpencode, Unsupported: true, Reason: "bash or sh was not found on PATH."}
+		}
+	}
+	return Plan{Target: TargetOpencode, Command: []string{"sh", "-c", "curl -fsSL https://opencode.ai/install | bash"}}
```
harshitsinghbhandari: The `sh` fallback above cannot work: if `bash` is absent and only `sh` is found, this still pipes into `bash` and the install fails at the pipe. Either drop the `sh` fallback and require `bash`, or pipe to `sh` here.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -0,0 +1,351 @@
...
+	return Plan{
+		Target: target, Unsupported: true,
+		Reason: "No supported Linux package manager (apt, dnf, pacman, zypper) was found.",
+	}
+}
+
+func linuxInstallArgv(mgr, pkg string) []string {
+	switch mgr {
+	case "apt-get":
+		return []string{"apt-get", "install", "-y", pkg}
```
harshitsinghbhandari, line 390: None of the Linux package-manager commands elevate, so run as the desktop user every one of these fails with a permission error. That makes the one-click install of tmux and gh non-functional on Linux. If avoiding sudo is intentional, the Linux path is better modelled as unsupported-with-instructions (show the command to run) than as a button that is guaranteed to fail.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -0,0 +1,351 @@
...
+	return &Service{
+		jobs:     make(map[Target]*Job),
+		lookPath: exec.LookPath,
+		goos:     runtime.GOOS,
+		commandFunc: func(argv []string) *exec.Cmd {
+			// context.Background(), not the caller's ctx: an install kicked
+			// off by an HTTP request must keep running (and stay queryable)
+			// after that request returns — the one deliberate exception to
+			// always threading ctx through.
+			return exec.CommandContext(context.Background(), argv[0], argv[1:]...) //nolint:gosec // G204: fixed allowlist, argv is never caller-derived.
```
harshitsinghbhandari: No deadline here, and `Start` returns early while `Status == running`, so a hung installer (curl stalling, a brew lock, winget waiting on input) pins that target in `running` permanently: no retry path, and the dialog polls a progress bar forever. Consider `context.WithTimeout` (10-15 minutes) with the expiry recorded as a failure.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -0,0 +1,351 @@
...
+const maxOutputBytes = 4000
+
+// Job is the tracked state of one install run for a Target.
+type Job struct {
+	Target     Target    `json:"target" enum:"tmux,gh,claude,codex,opencode,copilot" description:"Install target this job ran (or is running) for."`
+	Status     Status    `json:"status" enum:"idle,running,succeeded,failed,unsupported" description:"Current lifecycle state of the job."`
+	Command    string    `json:"command,omitempty" description:"Human-readable install command, e.g. \"brew install tmux\", for display even before/without output."`
+	Output     string    `json:"output,omitempty" description:"Combined stdout+stderr from the install command, tail-capped to the last ~4000 bytes."`
+	Error      string    `json:"error,omitempty" description:"Set on failure or when the target is unsupported on this machine: the exec error, or the Unsupported reason."`
+	StartedAt  time.Time `json:"startedAt,omitempty"`
```
harshitsinghbhandari: `omitempty` has no effect on a `time.Time` (it is a struct), so both fields always serialize as `0001-01-01T00:00:00Z`. That contradicts the `finishedAt` description and the generated spec, which lists both as optional. `*time.Time` would make absence real.

frontend/src/renderer/components/SystemRequirementsChecklist.tsx
```diff
@@ -0,0 +1,64 @@
...
+				<div
+					key={requirement.id}
+					className="ao-startup-checklist__row flex items-start gap-2"
+					style={{ animationDelay: `${index * STAGGER_STEP_MS}ms` }}
+				>
+					<RequirementGlyph requirement={requirement} />
+					<div className="min-w-0">
+						<p className="text-control font-medium text-foreground">{requirementDisplayLabel(requirement, t)}</p>
+						{requirement.detail ? (
+							<p className="text-caption leading-snug text-muted-foreground">{requirement.detail}</p>
```
harshitsinghbhandari: Nit: `detail` comes straight from `systemcheck.go` and is hardcoded English, so a non-English user sees localized labels next to English detail text. Deriving the copy from `requirement.id` here would keep it inside the i18n system that the last commit wired up.

harshitsinghbhandari (commented):
## Review: changes requested

Backend is in good shape: the six-target allowlist is genuinely closed (argv is never built from caller input), per-OS resolution is table-tested, jobs are async so no handler blocks on an installer, and both services plus both controllers have real coverage. The problems are in how the gate is wired into the board, plus a few install-plan edge cases.

### Blocking

**1. The startup gate does not actually gate anything.** `DaemonStartupLoader` only renders while `SessionsBoard.showStartup` is true (`SessionsBoard.tsx:159-162`, `:349`), and that flag is computed purely from daemon/workspace load state:

```
shell !== null && !daemonHasFailed &&
  (!isDaemonReady || workspaceStartupState === "loading" || (!workspaceQuery.isSuccess && !workspaceQuery.isError))
```

Nothing there consults requirements. On a machine with no tmux the daemon still starts and the workspace query still resolves, so a second or two later `showStartup` flips to false, the loader unmounts, and the non-dismissable dialog disappears on its own with the board fully usable behind it. Worse, `useInstallRunner`s unmount cleanup stops the poll, so an install kicked off from the dialog loses all UI feedback mid-run and the success refetch never fires, while the install keeps going in the daemon. The blocked state has to be lifted into the `showStartup` decision (or the loader kept mounted while a required check is unsatisfied), or the feature is just a flash on the way to the board.

### Should fix

**2. opencode: the `sh` fallback cannot work.** `planOpencode` accepts `sh` when `bash` is missing, then runs `sh -c "curl ... | bash"`. On a bash-less machine the probe passes and the install fails at the pipe. Either require `bash` outright or pipe to `sh`.

**3. Linux installs never elevate.** `apt-get install -y`, `dnf install -y`, `pacman -S`, `zypper install -y` all need root; run as the desktop user they fail immediately with a permission error. So one-click install of tmux/gh is effectively broken on Linux. If skipping sudo is deliberate (the PR body says it is), the Linux path should show the command to run rather than a button that always fails.

**4. No timeout and no cancel on a job.** `exec.CommandContext(context.Background(), ...)` with no deadline, and `Start` short-circuits while `Status == running`. A stalled installer (network hang on `curl`, a brew lock, winget waiting) wedges that target in `running` forever: the user can never retry, and the dialog polls an indefinite progress bar. A `context.WithTimeout` (10-15 min) plus a failed state on expiry would close this.

**5. `omitempty` does not omit a `time.Time`.** `StartedAt`/`FinishedAt` are structs, so the daemon always emits `0001-01-01T00:00:00Z`, contradicting both the `finishedAt` description (Zero until the job finishes) and the generated spec, which marks them optional. Use `*time.Time` if absence is what you want on the wire.

### Nit

**6.** `requirement.detail` is rendered verbatim in the checklist and dialog, but every one of those strings is hardcoded English in `systemcheck.go`. A non-English user gets localized labels next to English detail text. Deriving the copy from `requirement.id` on the frontend would keep it inside the new i18n system.

Everything else (CSS with reduced-motion handling, i18n key parity across all 8 locales, spec regen, controller 400 on unknown target) looks right.

frontend/src/renderer/components/DaemonStartupLoader.tsx
harshitsinghbhandari: Fixed — extracted `useSystemRequirementsGate` (frontend/src/renderer/hooks/useSystemRequirementsGate.ts) as the shared react-query source of truth, and folded `blocked` into `SessionsBoard.showStartup` (and `isLoaded`) so the screen stays mounted while a required check fails. `DaemonStartupLoader` now reads the same hook, so it and the mount decision can never disagree. Added a defensive effect that resets to the requirements phase if `blocked` flips true again after reaching the phrase-rotation phase.

backend/internal/service/systeminstall/systeminstall.go
harshitsinghbhandari: Fixed — dropped the sh fallback entirely; `planOpencode` now requires bash outright (matches opencode's documented installer, which is a bash script), and the outer invoking interpreter is bash too. Added a test asserting sh-without-bash is still Unsupported.

backend/internal/service/systeminstall/systeminstall.go
harshitsinghbhandari: Fixed exactly as suggested — `planLinuxPackage` now always resolves Unsupported on Linux, with Reason carrying the exact sudo-prefixed command (e.g. "sudo apt-get install -y tmux") for the user to run themselves. No auto-elevation added.

backend/internal/service/systeminstall/systeminstall.go
harshitsinghbhandari: Fixed — added a 15-minute `context.WithTimeout` in `run()`; a timeout now records `StatusFailed` with a "install timed out after 15m0s" message instead of pinning the target in `running` forever. Added `TestRun_Timeout` (short injected timeout + a context-respecting fake command) to exercise it deterministically.

backend/internal/service/systeminstall/systeminstall.go
harshitsinghbhandari: Fixed — `StartedAt`/`FinishedAt` are now `*time.Time`, so `omitempty` actually omits them until set. Updated the two tests that checked `.IsZero()` to check for nil instead, and regenerated the spec/schema.

frontend/src/renderer/components/SystemRequirementsChecklist.tsx
harshitsinghbhandari: Fixed — added `requirementDetailText(requirement, t)`: for a satisfied requirement it still shows the backend's `detail` verbatim (a resolved path or comma-joined agent list — data, not prose), but for an unsatisfied one it now renders a translated string keyed by `requirement.id` instead of the raw backend English. Wired into both the checklist and the dialog (git/tmux/harness/gh), with the 4 new `startup.detail*Missing` keys translated across all 8 locales.

harshitsinghbhandari:
Addressed all six review points:

**Blocking — the gate now actually gates.** Extracted `useSystemRequirementsGate` as a shared react-query hook and folded `blocked` into `SessionsBoard.showStartup`/`isLoaded`, so the startup screen stays mounted (and an in-flight install poll survives) until required checks actually pass — not just until the daemon/workspace query happen to resolve.

**opencode sh fallback** — dropped; now requires bash outright, matching opencode's actual documented installer.

**Linux installs never elevate** — `planLinuxPackage` now always resolves `Unsupported` on Linux with the exact `sudo`-prefixed command in `Reason`, rather than a button guaranteed to fail.

**No timeout on install jobs** — added a 15-minute `context.WithTimeout` in `run()`; a stalled installer now surfaces as `StatusFailed` instead of wedging in `running` forever.

**`omitempty` on `time.Time`** — `StartedAt`/`FinishedAt` are now `*time.Time`.

**Hardcoded English detail text** — added `requirementDetailText`, deriving the "why it's missing" copy from `requirement.id` through the i18n system (translated across all 8 locales) instead of rendering the backend's raw English sentence.

Full verification re-run after all fixes: Go build/vet/test (+`-race` on systeminstall)/lint clean, spec regenerated, frontend typecheck clean, full Vitest suite passing (only pre-existing, unrelated failures remain: a missing `cheerio` dep in landing scripts and one flaky `NotificationCenter` outside-click test, both confirmed via empty diff against `upstream/main`).

illegalcall:
## Follow-up review at `4e14362c` — reproducible issues

I tested this in a separate worktree and an isolated real Electron/daemon instance. I also added temporary regression tests for each behavioral claim, ran them against this head, then removed them. The existing focused suites still pass after cleanup (backend HTTP/CLI/services, 24 frontend loader/board tests, and frontend typecheck).

### 1. High: the board is not gated while the requirements request is pending

`blocked` is false until `query.isSuccess` ([hook](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/frontend/src/renderer/hooks/useSystemRequirementsGate.ts#L30-L40)), and `SessionsBoard` only includes `requirementsBlocked` in its startup decision ([board](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/frontend/src/renderer/components/SessionsBoard.tsx#L135-L153)). Therefore a pending first probe is treated as unblocked.

Reproduction: make projects/sessions resolve immediately and leave `GET /api/v1/system/requirements` pending. The full first-run board renders before the request completes. A later missing result changes it back to the loader/dialog, producing a board → loader flash.

Board visible while the system requirements request is still pending

Temporary RTL regression test result:

```text
[DEBUG-pr3825] keeps the startup loader mounted while requirements are still pending
Unable to find [data-testid="daemon-startup-loader"]
Actual DOM: data-testid="board" and data-testid="board-welcome"
```

The gate should treat `query.isPending` as startup-blocking (with the existing fail-open behavior reserved for `query.isError`).

### 2. High: manual recovery has no “Check again,” and a quick agent install can reuse stale data

The dialog footer contains only **Quit** ([dialog footer](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/frontend/src/renderer/components/InstallDependencyDialog.tsx#L236-L244)). There is no way to re-run the check after installing a dependency manually.

The agent requirement is more fragile: `systemcheck` calls `agent.Service.Refresh`, but `Refresh` returns its cached inventory for 10 seconds ([cache branch](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/service/agent/service.go#L153-L168)). A manual/fast install followed by the immediate refetch can therefore return the old empty inventory, and there is no delayed retry or manual button.

Missing dependency dialog offers install and quit, but no Check again action

```text
UI test: Unable to find button "Check again"
Actual buttons: "Install tmux", "Quit"

Mutable-agent test: Refresh returned cached empty inventory after manual install
(probes=1); want a fresh probe
```

### 3. Medium: Linux manual guidance is presented as an install action that is guaranteed to fail

Every Linux package-manager plan is deliberately `Unsupported` and carries a manual `sudo ...` command ([planner](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/service/systeminstall/systeminstall.go#L386-L420)). However the UI initially shows **Install tmux/gh**; only after clicking does it display the manual command as red error text, while changing the button to **Retry: Install tmux** ([rendering](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/frontend/src/renderer/components/InstallDependencyDialog.tsx#L288-L324)). There is no copy action.

I reproduced this in the real Electron app with the platform planner isolated to Linux: the initial dialog offered **Install tmux**; clicking it produced `AO does not run installers as root. Run this yourself in a terminal: sudo apt-get install -y tmux` beneath a **Retry: Install tmux** button. Linux should show the exact copyable command from the start, without offering an install/retry action the daemon will never execute.

Linux unsupported plan appears as a retry install action and uncopyable error text

### 4. Medium: daemon install jobs call exit code 0 “succeeded” without verifying the target exists

The daemon sets `StatusSucceeded` solely when `cmd.Run()` returns nil ([run](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/service/systeminstall/systeminstall.go#L240-L267)). This contradicts the PR's own CLI logic, which correctly treats the post-install PATH check as the success signal ([CLI recheck](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/cli/tmux.go#L67-L80)).

Deterministic reproduction: use an install command that exits 0 and a `LookPath` fixture where the target remains absent.

```text
job reports "succeeded" even though tmux is still absent from LookPath
```

The daemon job should re-probe the installed target before reporting success.

### 5. Medium: the CLI drops the exact Linux command when `sudo` is unavailable

On Linux, the resolver knows the package-manager argv, but `withPrivilege` returns nil when a non-root user has no `sudo` ([privilege logic](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/cli/tmux.go#L91-L118)). `ensureTmux` then prints only the generic package-manager warning ([fallback](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/cli/tmux.go#L37-L46)), despite the requirement to name the exact command.

```text
Warning: tmux is not in PATH, so agent sessions will fail to start on linux.
To fix it, install it with your package manager.
```

Fixture: unprivileged user, `apt-get` present, `sudo` absent. Expected remedy can still say `apt-get install -y tmux` (and explain root privileges are required).

### Additional architecture/API findings

6. **High — the authenticated LAN listener can trigger host installs.** The LAN listener wraps the same router used by loopback ([shared handler](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/httpd/server.go#L82-L85)); `/api/v1/system/install/*` is registered on that shared router ([registration](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/httpd/api.go#L159-L182)) and is absent from the LAN control blocklist ([blocklist](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/httpd/lan_listener.go#L45-L61)). A real `NewMobileLAN` test with valid bearer auth returned `202` and reached the host-install handler: `status=202 called=true; want 404/false`. This host-mutating control route should be loopback-only.

7. **High — `ao start` now performs installation in-process instead of remaining a thin daemon client.** `backend/internal/cli/tmux.go` imports the service resolver and invokes `RunInteractive` directly ([file](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/cli/tmux.go#L1-L24), [execution](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/cli/tmux.go#L67-L80)). This bypasses the documented CLI → loopback HTTP → daemon boundary.

8. **Medium — core services directly depend on `os/exec`.** Both `systemcheck` and `systeminstall` import/use `os/exec` in the service packages ([systemcheck](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/service/systemcheck/systemcheck.go#L8-L17), [systeminstall](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/service/systeminstall/systeminstall.go#L8-L19)). The repository architecture says external systems belong behind ports/adapters; the injected function fields help tests but do not restore that dependency boundary.

9. **Medium — the OpenAPI contract omits real `500` responses.** The controller already has a test proving a service error produces HTTP 500, but the POST operation documents only 202/400/501 and GET only 200/400/501 ([spec registry](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/httpd/apispec/specgen/build.go#L479-L498)). Generated clients therefore do not model an actual response.

10. **Low — the new service-error test does not assert the API error envelope.** `TestSystemInstallController_ServiceError` checks only the status code ([test](https://github.com/Untrivial-ai/agent-orchestrator/blob/4e14362c285e4d22f1984290543d58e4727fcb22/backend/internal/httpd/controllers/systeminstall_test.go#L145-L156)), so request ID/code/message envelope regressions on the new routes would pass unnoticed.

### Clean verification after removing the temporary probes

```text
go test ./internal/service/systeminstall ./internal/service/systemcheck \
  ./internal/service/agent ./internal/cli ./internal/httpd/...    PASS

npx vitest run board-empty-states.test.tsx DaemonStartupLoader.test.tsx
Test Files 2 passed; Tests 24 passed

npm run typecheck    PASS
```

illegalcall (approved):

PR #3829, opened 2026-08-10
fix(telemetry): retry failed sends; don't disable telemetry after a startup hiccup

Pulkit7070:
## Summary

Two focused fixes so telemetry that fails to send is recovered instead of silently vanishing. Follow-up to the renderer `send_instantly` fix (https://github.com/Untrivial-ai/agent-orchestrator/pull/3816). No steady-state event-volume increase, so no impact on the PostHog cap.

**Backend, retry a failed send** (`adapters/telemetry/posthog.go`). The CLI `app.active` day and each `cli.invoked` are marked spent on disk before the event is sent, so a dropped send was a permanent gap. `send` now retries a *failed* attempt (transient network error / HTTP 429 / 5xx), bounded by `postHogSendMaxAttempts`; a permanent 4xx is not retried. A healthy send still makes exactly one request, and a sustained outage gives up rather than blocking the drain goroutine.

**Renderer, don't disable telemetry after a startup hiccup** (`renderer/lib/telemetry.ts`). A thrown (transient) failure in `initTelemetry` was memoized, disabling telemetry for the whole session. It now clears the memoized promise on a thrown error so the next event retries; a deliberate withheld state (no key / not opted in) is still cached and not retried.

## Cost

Neither change raises steady-state volume. Retries only re-send events that already failed (which otherwise sent nothing); bounded attempts cap the worst case during an outage.

## Verification

- **Backend: run at runtime over real sockets.** A live HTTP server that returns 500-then-200 received 2 requests (retried and delivered); a permanent 400 received 1 (no wasted retry); a sustained 500 stopped after the bounded cap. `go test ./internal/adapters/telemetry/` passes incl. `posthog_retry_test.go`.
- **Renderer: run at runtime.** `initTelemetry()` after a thrown bootstrap error returned `false`, and the next call returned `true` (retry works; old behavior returned `false` both times). Executed through the real toolchain with only the posthog/bridge boundaries stubbed.

## Not in this PR (deliberately)

Trimmed out to keep this focused: the `ao.cli.invoked` burst rate-limit tweak, a dev-only warning for unlisted renderer events, and mobile flush-on-background. All real but non-essential; each can ship as a small standalone change. The aggregated-error dimension mis-attribution (5xx-by-endpoint) is still a known follow-up needing a payload redesign.


backend/internal/adapters/telemetry/posthog.go
```diff
@@ -302,23 +314,43 @@ func (s *PostHogSink) send(ev ports.TelemetryEvent) {
 		s.log.Warn("telemetry posthog payload marshal failed", "name", ev.Name, "error", err)
 		return
 	}
+
+	for attempt := 1; attempt <= postHogSendMaxAttempts; attempt++ {
+		retryable, err := s.sendOnce(payload)
```
illegalcall: This retry is not idempotent because the payload has no stable top-level uuid. A transport error only means AO did not receive a response; PostHog may already have accepted the first request, so the second attempt can be ingested as a duplicate. PostHog documents event UUIDs as the idempotency mechanism: https://pkg.go.dev/github.com/posthog/posthog-go#Capture. Please generate one UUID per logical event before marshaling, reuse it across attempts, and test that both retry requests carry the same non-empty UUID. Otherwise the PR body's claims that the event otherwise sent nothing and that retries do not affect the PostHog cap do not hold.

backend/internal/adapters/telemetry/posthog.go
```diff
@@ -302,23 +314,43 @@ func (s *PostHogSink) send(ev ports.TelemetryEvent) {
...
+		time.Sleep(postHogSendRetryDelay)
+	}
+}
+
+// sendOnce posts a single capture attempt. retryable is true only for
+// transient failures a retry might clear (connection errors, HTTP 429, and
+// 5xx); a permanent rejection (any other non-2xx, e.g. a 4xx auth/shape error)
+// returns retryable=false so a retry does not waste a call that will fail
+// identically.
+func (s *PostHogSink) sendOnce(payload []byte) (retryable bool, err error) {
```
illegalcall: sendOnce performs HTTP I/O using context.Background(), and the retry delay uses an uncancellable time.Sleep. AGENTS.md requires context.Context as the first argument for I/O or blocking work. Please thread a lifecycle context through send and sendOnce and make the retry delay cancellable so shutdown cancellation can stop pending retry work instead of leaving it alive until the sleep or HTTP timeout finishes.

backend/internal/adapters/telemetry/posthog.go
```diff
@@ -302,23 +314,43 @@ func (s *PostHogSink) send(ev ports.TelemetryEvent) {
...
 	defer func() { _ = resp.Body.Close() }()
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		b, _ := io.ReadAll(io.LimitReader(resp.Body, 2048))
-		s.log.Warn("telemetry posthog rejected event", "name", ev.Name, "status", resp.StatusCode, "body", strings.TrimSpace(string(b)))
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return false, nil
 	}
+	b, _ := io.ReadAll(io.LimitReader(resp.Body, 2048))
+	rejected := fmt.Errorf("rejected status=%d body=%s", resp.StatusCode, strings.TrimSpace(string(b)))
+	return resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500, rejected
```
illegalcall, line 384: The HTTP 429 condition is a distinct retry branch, but the new suite covers 500, transport errors, 400, sustained 500, and success only. Please add a 429 regression case that verifies it retries and remains bounded.

frontend/src/renderer/lib/telemetry.ts
```diff
@@ -691,8 +691,18 @@ export async function initTelemetry(): Promise<boolean> {
...
-	})().catch(() => false);
-	return initPromise;
+	})().catch(() => {
+		// A thrown error is transient (the bridge not ready yet, or a hiccup in
+		// posthog.init): drop the memoized promise so the next captured event
+		// retries init rather than the whole session going dark after one bad
+		// moment. A deliberate `return false` above (no key, not opted in) is
+		// not an error and stays cached, so an intentionally-withheld client is
+		// not retried forever.
+		if (initPromise === attempt) initPromise = null;
```
illegalcall, line 701: This promise reset is the central renderer behavior change, but no committed renderer test exercises initTelemetry. Please add a regression test where the first bootstrap or initialization attempt rejects and a later call succeeds, plus a case confirming a deliberate false result remains memoized. The manual verification described in the PR body will not protect this behavior from regression.

illegalcall (commented):
Reviewed the current slimmed-down head. The remaining exact issues are inline.

Pulkit7070:
Thanks, all four addressed in a7b6bb1.

1. Idempotency: `send` now generates one event `uuid` and reuses it across attempts. A transport error only means we got no response, so PostHog may already have ingested the first attempt; reusing the uuid lets PostHog dedupe, so the no-added-volume guarantee holds. Added a test asserting both attempts carry the same non-empty uuid.
2. Context: threaded a lifecycle context (created in `NewPostHogSink`, cancelled by `Close` when the caller's shutdown context is done) through `send`/`sendOnce`, and replaced the `time.Sleep` backoff with a cancellable timer + `ctx.Done()`. Added a test that a cancelled context skips the retry.
3. Added a 429 regression case (retries and stays bounded).
4. Added a committed renderer init test: first attempt rejects then a later call succeeds, plus a deliberately withheld (null bootstrap) result staying memoized without re-bootstrapping.

illegalcall (approved):

PR #3830, opened 2026-08-10
fix(workspace): persist child default branches

### LINKED ISSUE #3810: [Bug]: Workspace spawn forces the root repo's default branch on every child repo (BRANCH_NOT_FETCHED)
## Summary

Spawning any session (orchestrator or worker) in a **workspace project** fails with `BRANCH_NOT_FETCHED` when a child repo's default branch is not the same as the root project's. AO resolves one base branch — the root project's configured default, which falls back to the hardcoded constant `"main"` when unset — and applies it to **every** repo in the workspace, so any child without a `main` branch can never get a worktree.

Workspaces where every repo happens to have `main` work fine, which is why this hides in testing.

## Steps to reproduce

1. Register a workspace project whose root repo's default branch is `main`, with at least one child repo whose default branch is something else (`dev`) and which has **no** `main` branch, local or remote:
   ```
   myworkspace/          default: main
   ├── frontend/         default: dev   (no main)
   ├── backend/          default: dev   (no main)
   └── deploy/           default: main
   ```
2. Leave the project's `defaultBranch` config unset (the normal case).
3. Spawn an orchestrator: `POST /api/v1/orchestrators {"projectId":"myworkspace","clean":false}` — or click the button in the desktop app.

## Expected behavior

Each repo's worktree is branched from **that repo's own** default branch: root from `origin/main`, `frontend`/`backend` from `origin/dev`, `deploy` from `origin/main`. The adapter already knows how to do this — `inferRepoDefaultBranch` reads each repo's `origin/HEAD` — it just never gets the chance.

## Actual behavior

The spawn fails and the seed session row is rolled back:

```
spawn myworkspace-6: workspace: branch is not fetched:
"ao/myworkspace-6-16" has no local head, no remote, and no tag — run `git fetch` then retry
```

`git fetch` does not help: the branch AO is looking for (`main`) does not exist in those repos at all.

### Root cause

- `backend/internal/session_manager/manager.go:932` passes `project.Config.WithDefaults().DefaultBranch` as the workspace's `BaseBranch`. `WithDefaults()` substitutes `domain.DefaultBranchName` = `"main"` whenever the project configures nothing (`backend/internal/domain/projectconfig.go:117`).
- `backend/internal/adapters/workspace/gitworktree/workspace.go:199` then applies that single value to every child: `baseBranch: firstNonEmpty(child.BaseBranch, cfg.BaseBranch)`. `child.BaseBranch` is never populated by any caller, so `cfg.BaseBranch` always wins.
- A non-empty base branch makes `resolveBaseRef` (`workspace.go:1116-1121`) **skip** per-repo inference entirely and go straight to `resolveBaseRefFromDefault`. Candidates for a `dev` child become `origin/ao/<session>`, `origin/main`, `refs/heads/main`, `ao/<session>`, `refs/tags/<branch>` — none exist → `errNoBaseRef` → `ErrBranchNotFetched`.

`manager_test.go:2038` already asserts child repos are handed an empty `BaseBranch` "so adapter infers per-repo default" — the adapter overrides that intent one layer down, so the test passes while the behavior is wrong.

The single-repo path is unaffected: there is only one repo, and it is the one whose default branch the project config describes.

### Side effect: leaked branches on every failed attempt

Each failed attempt leaves a branch behind in the root repo. The root worktree is created first (`git worktree add -b <branch>` creates the ref), a child then fails, and rollback's `forceDestroyPath` removes worktrees but not the branch. `workspaceProjectBranch` sees the leftover as taken and bumps the suffix, so retries walk `ao/<session>`, `ao/<session>-2`, … `ao/<session>-16`, accumulating dead refs in the root repo.

## AO version

0.12.2 (desktop app)

## Environment

macOS 15.x (Darwin 25.3.0) arm64

## Logs / extra context

```shell
# reproduced against the running 0.12.2 daemon
$ curl -s -X POST http://127.0.0.1:3001/api/v1/orchestrators \
    -H 'content-type: application/json' \
    --data-raw '{"projectId":"myworkspace","clean":false}'
{"error":"bad_request","code":"BRANCH_NOT_FETCHED",
 "message":"spawn myworkspace-6: workspace: branch is not fetched: \"ao/myworkspace-6-16\" has no local head, no remote, and no tag — run `git fetch` then retry"}

# per-repo inference (what the adapter would do if it were allowed to) resolves a base in all four:
root:     default=main base=origin/main OK
frontend: default=dev  base=origin/dev  OK
backend:  default=dev  base=origin/dev  OK
deploy:   default=main base=origin/main OK

# leaked refs after the failed attempts
$ git branch --list 'ao/myworkspace-6*' | wc -l
      16
```

## Related

- #2329 (persist each child repo's default branch) would supply the explicit per-repo override this fix leaves as the first resolution step. This issue is the bug that blocks spawning today; #2329 is the durable-metadata follow-up. Neither depends on the other.
- #2222 — parent workspace-support epic.

## Separately observed, not covered here

Workspace spawn cannot reuse an existing orchestrator worktree directory. Orchestrator worktrees use a fixed path (`managedPath`, `workspace.go:1339`); the single-repo `Create` handles a pre-existing one via `existingWorktree()` (`workspace.go:139`), but `CreateWorkspaceProject` (`workspace.go:208-215`) goes straight to `git worktree add` and dies on `fatal: '…' already exists`, surfacing as an untyped `INTERNAL_ERROR` 500 with no actionable message. Reproduced on a project converted from `single_repo` to `workspace` while a terminated orchestrator's worktree was still on disk. Happy to file this one separately if it should not ride along.

neversettle17-101:
## Summary

- detect and persist each workspace child repository default branch during registration
- pass the persisted branch into workspace materialization instead of inheriting the root branch
- retain per-repository `origin/HEAD` inference for migrated rows without branch metadata
- cover mixed `main`/`dev` workspace creation and metadata import

Closes #3810

## Tests

- `go test ./internal/service/project ./internal/session_manager ./internal/devimport ./internal/storage/sqlite/store ./internal/adapters/workspace/gitworktree -count=1`
- `go test ./internal/adapters/agent/kilocode -run TestAuthStatusUnknownWhenKeyOnlyComesFromInteractiveShell -count=1`

## Verification note

A full `go test ./... -count=1` run reached an unrelated timeout in `internal/adapters/agent/kilocode`; that isolated test passed immediately on rerun. All workspace-related packages pass.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/workspace.go
+++ b/backend/internal/adapters/workspace/gitworktree/workspace.go
@@ -196,7 +196,7 @@ func (w *Workspace) CreateWorkspaceProject(ctx context.Context, cfg ports.Worksp
 			relativePath: rel,
 			repoPath:     repoPath,
 			outputPath:   outPath,
-			baseBranch:   firstNonEmpty(child.BaseBranch, cfg.BaseBranch),
+			baseBranch:   child.BaseBranch,
 		})
 	}
 	branch, err := w.workspaceProjectBranch(ctx, repos, firstNonEmpty(cfg.Branch, defaultSessionBranchName(cfg.SessionID)))
--- a/backend/internal/adapters/workspace/gitworktree/workspace_integration_test.go
+++ b/backend/internal/adapters/workspace/gitworktree/workspace_integration_test.go
@@ -563,7 +563,6 @@ func setupOriginCloneOnBranch(t *testing.T, git, tmp, branch string) string {
 	t.Helper()
 	repo := setupOriginClone(t, git, tmp)
 	runGit(t, git, repo, "branch", "-m", "main", branch)
-	runGit(t, git, repo, "push", "origin", ":"+"main")
 	runGit(t, git, repo, "push", "-u", "origin", branch)
 	runGit(t, git, repo, "remote", "set-head", "origin", branch)
 	return repo
--- a/backend/internal/adapters/workspace/gitworktree/workspace_integration_test.go
+++ b/backend/internal/adapters/workspace/gitworktree/workspace_integration_test.go
@@ -457,6 +457,68 @@ func TestWorkspaceIntegrationCreateInRemotelessRepo(t *testing.T) {
 	}
 }
 
+func TestWorkspaceIntegrationWorkspaceProjectInfersChildDefaultBranches(t *testing.T) {
+	git := requireGit(t)
+	tmp := t.TempDir()
+	rootRepo := setupOriginClone(t, git, filepath.Join(tmp, "root"))
+	devChildRepo := setupOriginCloneOnBranch(t, git, filepath.Join(tmp, "dev-child"), "dev")
+	mainChildRepo := setupOriginClone(t, git, filepath.Join(tmp, "main-child"))
+
+	root := filepath.Join(tmp, "managed")
+	ws, err := New(Options{Binary: git, ManagedRoot: root, RepoResolver: StaticRepoResolver{"proj": rootRepo}})
+	if err != nil {
+		t.Fatalf("new: %v", err)
+	}
+	info, err := ws.CreateWorkspaceProject(context.Background(), ports.WorkspaceProjectConfig{
+		ProjectID:    "proj",
+		SessionID:    "sess",
+		Kind:         "worker",
+		Branch:       "ao/proj-1",
+		RootRepoPath: rootRepo,
+		BaseBranch:   "main",
+		Repos: []ports.WorkspaceProjectRepoConfig{
+			{
+				Name:         "api",
+				RelativePath: "services/api",
+				RepoPath:     devChildRepo,
+			},
+			{
+				Name:         "web",
+				RelativePath: "apps/web",
+				RepoPath:     mainChildRepo,
+			},
+		},
+	})
+	if err != nil {
+		t.Fatalf("create workspace project: %v", err)
+	}
+	if len(info.Worktrees) != 3 {
+		t.Fatalf("worktrees = %d, want root and two children: %#v", len(info.Worktrees), info.Worktrees)
+	}
+	devChildPath := filepath.Join(info.Root.Path, "services", "api")
+	if _, err := os.Stat(filepath.Join(devChildPath, "README.md")); err != nil {
+		t.Fatalf("dev child worktree missing seed file: %v", err)
+	}
+	devChildHead := gitOutput(t, git, devChildRepo, "rev-parse", "refs/heads/ao/proj-1")
+	devChildBase := gitOutput(t, git, devChildRepo, "rev-parse", "origin/dev")
+	if devChildHead != devChildBase {
+		t.Fatalf("dev child branch base = %s, want origin/dev %s", devChildHead, devChildBase)
+	}
+	mainChildHead := gitOutput(t, git, mainChildRepo, "rev-parse", "refs/heads/ao/proj-1")
+	mainChildBase := gitOutput(t, git, mainChildRepo, "rev-parse", "origin/main")
+	if mainChildHead != mainChildBase {
+		t.Fatalf("main child branch base = %s, want origin/main %s", mainChildHead, mainChildBase)
+	}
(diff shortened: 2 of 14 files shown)


neversettle17-101 (commented):
## What This PR Does

Persists each workspace child repository default branch at registration/import time and passes it into workspace creation, while preserving adapter inference for legacy rows with an empty value.

## Summary of Changes

- Adds the `default_branch` workspace-repository column and threads it through domain and SQLite mappings.
- Detects child defaults during registration and uses the persisted value when spawning sessions.
- Removes the incorrect root-branch fallback for child repositories.
- Adds storage/service/session tests plus an integration test covering children with different defaults.

## Findings

No correctness issues found in the submitted changes. Ready to merge.

## Regression Risk

Low. Existing rows migrate with an empty value and retain runtime inference; newly registered rows use the captured per-repository default.

## Testing

The added tests cover persistence, import copying, configuration propagation, and real Git worktree creation across mixed child defaults. CI lint, API drift, secret scan, and most CLI E2E checks are passing; Go build/test and Windows CLI E2E were still in progress at review time. Tests were not run locally.

harshitsinghbhandari:
Reviewed against the branch (not just the diff). `sqlc diff` is clean, and `go test ./internal/service/project ./internal/session_manager ./internal/devimport ./internal/storage/sqlite/...` all pass locally. One blocking item.

### Blocking

**`backend/internal/adapters/workspace/gitworktree/workspace_integration_test.go:566`** — `push origin :main` deletes the current branch of a bare origin (`setupOriginClone` does `init --bare`, so origin's HEAD points at `main`). Git 2.45+ refuses this by default via `receive.denyDeleteCurrent`. On git 2.54 the new test fails:

```
remote: error: refusing to delete the current branch: refs/heads/main
 ! [remote rejected] main (deletion of the current branch prohibited)
```

CI only passed because the runner image ships an older git, so this will break locally for anyone on a current git and will break CI on the next image bump.

Fix: delete line 566. The `remote set-head origin <branch>` two lines below already points the clone's `origin/HEAD` at `dev`, which is all `inferRepoDefaultBranch` reads. I verified that removing just that line makes `TestWorkspaceIntegrationWorkspaceProjectInfersChildDefaultBranches` pass.

### Non-blocking

**`backend/internal/service/project/workspace_registration.go:179`** — `resolveDefaultBranch` (`service.go:670`) falls back to the *currently checked out* branch when `origin/HEAD` is unset, and `validateWorkspaceChild:218` only requires an origin *URL*, not a fetched `origin/HEAD` (`git remote add origin` never sets it). So registering a workspace while a child repo sits on a feature branch pins that feature branch as the child's base for every future session, permanently: `Add` rejects re-registration (`service.go:190`, `PATH_ALREADY_REGISTERED`), nothing else writes `default_branch`, and `workspaceReposFromRecords:359` drops the field, so it is not even visible in the API. Worth persisting only the `origin/HEAD` result and leaving `""` otherwise, so the adapter's live inference (`workspace.go:1120`) keeps covering that case. Fine as a follow-up.

**Scope question, not a defect:** the one-line change at `workspace.go:199` (`firstNonEmpty(child.BaseBranch, cfg.BaseBranch)` to `child.BaseBranch`) appears to fix #3810 on its own. The new integration test passes children with no `BaseBranch` set and still asserts `origin/dev` / `origin/main`, which shows the adapter's existing per-repo inference already produces the right result. The new column then only changes behavior for children with no fetched `origin/HEAD`, where it freezes a registration-time snapshot. If durable per-repo branches are the deliberate goal rather than live inference, that is a reasonable call, just worth stating in the description, since it accounts for roughly 100 of the 153 added lines (migration 0086, the domain field, the sqlc regen, the store and session_manager plumbing).

Nit: `":"+"main"` on line 566 is a redundant concat, moot if that line goes.

Migration 0086 itself looks right: `DROP COLUMN` has precedent in 0084 and 0081, legacy `''` rows correctly fall through to inference, and `devimport` copies the field by struct value.

With line 566 removed this is good to go from my side.

illegalcall:
@harshitsinghbhandari is this good to merge?

illegalcall (approved):

PR #3831, opened 2026-08-11
perf: scope CDC refetch fan-out, batch session-list PR facts, stop idle shell re-renders, narrow preview poll

AgentWrapper:
Profiled the live app first (135 sessions, real daemon + real database), then shipped the four highest-impact, well-scoped wins. No behavior changes beyond update timing bounds documented below; every win has a regression test and before/after numbers.

## How the hotspots were found

- Captured the live SSE `/api/v1/events` stream for 60s during agent activity: **187 `session_updated` events/min** (usage-binding + activity CDC), each of which made the renderer refetch the 99KB `/api/v1/sessions` payload, the usage summary, **and every mounted per-session PR summary** (prefix invalidation of `["session-scm-summary"]` = one HTTP request per visible session card).
- Measured `GET /api/v1/sessions` latency and traced its 1+N SQLite query pattern.
- Read-path review of renderer render churn (whole-store zustand subscriptions, per-poll fresh status objects) and daemon polling loops (preview poller reads all 38 session columns at 4Hz).

## Wins and measurements

| # | Change | Metric | Before | After | How measured |
|---|--------|--------|--------|-------|--------------|
| 1 | Scope + rate-bound CDC invalidation (`event-transport.ts`) | Invalidation rounds for 20 `session_updated` events at the live 320ms cadence (workspaces / scm-root / usage) | 20 / 20 / 20 | 7 / **0** / 7 | vitest fake-timer harness run against the old and new implementation |
| 1 | 〃 | Per-session PR summary refetch storms (HTTP requests per round ≈ visible cards) | every round | 0 on session bursts; scoped to the 1 owning session on PR events | same harness |
| 1 | 〃 | Sub-150ms continuous stream (50 events @ 100ms) | 1 starved flush **after** the stream ends (UI frozen while active) | ~1 flush/s while streaming (5 flushes) | same harness |
| 2 | Batch session-list PR facts (kill 1+N) | `Service.List` store cost, 135 sessions / 90 PRs | 2.85ms, 6030 allocs | 0.26ms, 2741 allocs (**10.8×**) | `go test -bench` `BenchmarkListPRFactsPerSession` vs `BenchmarkListAllPRFacts` (Apple M3) |
| 2 | 〃 | Live `GET /api/v1/sessions` (100 reqs, copy of real 135-session DB, before/after daemon binaries) | p50 10.3ms, p90 16.0ms | p50 3.8ms, p90 5.9ms | curl timing against both binaries on identical DB copies |
| 2 | 〃 | SQLite queries per list build | 1 + 135 | 2 | query pattern |
| 3 | Idle shell re-renders (`useDaemonStatus` equality bailout, zustand field selectors, memoized shell context, TrayRuntime stringify) | Consumer re-renders per unchanged 10s status poll (cascades through ShellLayout → Sidebar, board, titlebar, dialogs) | 1 per poll | 0 | render-count regression test (fails on the old hook) |
| 4 | Narrow preview-poller projection (`ListSessionPreviewRows`) | Per-tick cost of the 250ms scan, 135 sessions with ~4KB text columns | 1.62ms, 4.34MB allocs | 0.13ms, 0.12MB (**12× time, 36× bytes**) | `go test -bench` `BenchmarkPreviewPoll*` |

Compound effect during active agent work: the workspace/usage refetch round-trip rate drops ~3× (bounded at 1/s), each round no longer fans out N per-card PR summary requests, and each `/sessions` build costs ~2.6ms less in the daemon — while idle, the app shell stops re-rendering entirely between real changes.

**Update-timing bounds (deliberate):** workspace/scm/usage refetches now flush at most once per second under sustained CDC streams (previously either one round per event, or complete starvation while events arrived faster than the 150ms debounce). Isolated events keep the 150ms latency. Chat/conversation invalidation cadence is unchanged.

## Correctness guards

- `TestListAllPRFactsMatchesPerSessionReads` pins the batch read to the per-session read it replaces (order + fields).
- `TestListSessionPreviewRowsMatchesListAllSessions` pins the narrow poller read to `ListAllSessions` on every field the poller uses, including the terminated-session filter.
- New event-transport tests cover PR-event scoping, `pr_session_changed` root fallback, storm collapse, and no-starvation; the malformed-payload and unknown-event paths conservatively refetch everything, as before.
- Render-count test on `useDaemonStatus` fails against the previous implementation.

## Validation

- `cd backend && go test ./...` — green except pre-existing `internal/cli` failures that also fail on a clean checkout of `main` in this environment (they collide with the real daemon running on this machine).
- `go test -race` on touched packages (`service/session`, `sqlite/store`, `preview`, `httpd/...`) — green.
- `golangci-lint v2.12.2` — 0 issues.
- `frontend: npm run typecheck` — clean; `npm test` — green except two pre-existing `src/landing/*` suites that fail identically on a clean tree; renderer `vite build` — OK.
- Live daemon A/B: both binaries run against identical copies of the real database (data dir isolated in a scratch directory, separate port/run-file, shut down after measuring).
- No `ao preview` demo: nothing visual changed — the sidebar/board render identically; the change is network/CPU behavior, covered by the measurements above.

## Big-but-risky follow-ups (found while profiling, deliberately not done here)

1. **Reaper fork storm** — 3 subprocess forks per live session every 5s, including a full `ps -axo` process-table dump *per session* (`tmux.go:564`). Batch to one `ps` + one `tmux list-sessions` per tick. Touches liveness semantics, needs care around the "failed probe ≠ dead" rule.
2. **SSE `/api/v1/events` replays the entire `change_log` (29k+ rows, unbounded, no retention) on every connect** since the renderer never passes `?after=`. The client only uses events as invalidation signals and refetches on open anyway. Needs an API-contract decision (tail parameter or retention policy).
3. **Terminal mux emits one JSON+base64 frame per PTY read** with no coalescing; overflow kills the connection (`manager.go:529`). Binary frames + drain-and-concat would cut most of the per-byte cost.
4. **`GET /sessions/{id}/pr` runs 4 queries per PR alias** (checks/threads/reviews/comments) — same batching treatment as this PR's win 2.
5. Missing `(project_id, num)` index on `sessions` (every list sort builds a temp B-tree); `idx_change_log_project` is written on every trigger insert but never read.
6. No `pprof` endpoint in the daemon — worth adding behind a flag before the next perf pass.

### DIFF
--- a/backend/internal/preview/poller.go
+++ b/backend/internal/preview/poller.go
@@ -14,7 +14,11 @@ import (
 const DefaultPollInterval = 250 * time.Millisecond
 
 type sessionPreviewSource interface {
-	ListAllSessions(ctx context.Context) ([]domain.SessionRecord, error)
+	// ListSessionPreviewRows returns sparse, non-terminated session records
+	// carrying only the identity/kind/preview/workspace fields this poller
+	// reads — a sub-second loop must not drag the full 38-column session
+	// projection (multi-KB prompt/transcript text) out of SQLite every tick.
+	ListSessionPreviewRows(ctx context.Context) ([]domain.SessionRecord, error)
 }
 
 type previewSetter interface {
@@ -101,7 +105,7 @@ func (p *Poller) Poll(ctx context.Context) error {
 	if p.source == nil || p.setter == nil {
 		return nil
 	}
-	sessions, err := p.source.ListAllSessions(ctx)
+	sessions, err := p.source.ListSessionPreviewRows(ctx)
 	if err != nil {
 		return fmt.Errorf("preview poller list sessions: %w", err)
 	}
--- a/backend/internal/preview/poller_test.go
+++ b/backend/internal/preview/poller_test.go
@@ -22,8 +22,16 @@ type previewSet struct {
 	url string
 }
 
-func (f *fakePreviewSessions) ListAllSessions(_ context.Context) ([]domain.SessionRecord, error) {
-	return append([]domain.SessionRecord(nil), f.sessions...), nil
+func (f *fakePreviewSessions) ListSessionPreviewRows(_ context.Context) ([]domain.SessionRecord, error) {
+	// Mirror the production query's contract: terminated sessions are excluded.
+	out := make([]domain.SessionRecord, 0, len(f.sessions))
+	for _, sess := range f.sessions {
+		if sess.IsTerminated {
+			continue
+		}
+		out = append(out, sess)
+	}
+	return out, nil
 }
 
 func (f *fakePreviewSessions) SetPreview(_ context.Context, id domain.SessionID, previewURL string) (domain.Session, error) {
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -32,6 +32,7 @@ type Store interface {
 	SetSessionReviewerHarness(ctx context.Context, id domain.SessionID, harness domain.ReviewerHarness, updatedAt time.Time) (bool, error)
 	GetDisplayPRFactsForSession(ctx context.Context, id domain.SessionID) (domain.PRFacts, bool, error)
 	ListPRFactsForSession(ctx context.Context, id domain.SessionID) ([]domain.PRFacts, error)
+	ListAllPRFacts(ctx context.Context) (map[domain.SessionID][]domain.PRFacts, error)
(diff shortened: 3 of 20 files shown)


i-trytoohard:

i-trytoohard:
@AgentWrapper Closing as stale: no activity from you for 32 days. The branch stays safe to reopen if you pick this up again.

PR #3696, opened 2026-08-07
fix: open board from project name without double-click

Annieeeee11:
Fixes #3682

## Summary
- Make the orchestrator project name a stable button around the Motion label so the shared `layoutId` spring cannot swallow the first click
- Remove the redundant Kanban button; project name is the board control
- Keep that control visible when the topbar is embedded on session routes

## Test plan
- [ ] Open an orchestrator session → click the project name once → lands on the kanban board
- [ ] Confirm the old Kanban button is gone
- [ ] Worker sessions still show Kill / Orchestrator / inspector as before
- [ ] `npx vitest run src/renderer/components/ShellTopbar.test.tsx`

PR #3832, opened 2026-08-11
fix(frontend): restore Kanban topbar button

AgentWrapper:
## Summary
- restore the dedicated dashboard-icon + Kanban action on orchestrator session topbars
- keep the project name as contextual text instead of an ambiguous navigation control
- cover both embedded and full topbar variants with navigation and presentation assertions

## Regression
PR #3696 removed the visible Kanban control while fixing the project-name double-click behavior. That made the project name the only route back to the board and obscured the action.

## Tests
- `cd frontend && npx vitest run --config vite.renderer.config.ts src/renderer/components/ShellTopbar.test.tsx` (29 passed)


Annieeeee11 (approved):
lgtm

PR #3834, opened 2026-08-11
fix(agent-devin): refuse to overwrite zero-byte ~/.claude.json (sibling of #3746)

### LINKED ISSUE #3833: bug(agent-devin): zero-byte ~/.claude.json is overwritten, wiping Claude login + history (same class as #3746)
### Summary

`ensureDevinWorkspaceTrusted` (`backend/internal/adapters/agent/devin/devin.go:269`) has the same zero-byte data-loss bug as #3746 — on the **same** `~/.claude.json`. The fix in #3756 patches only `claudecode.go`; the Devin adapter reaches the identical unguarded read.

When AO spawns a Devin session it marks the workspace trusted in Claude Code's `~/.claude.json` (Devin is a Claude-compat adapter and reuses that file). It reads the config:

```go
data, err := os.ReadFile(configPath) // configPath = ~/.claude.json
switch {
case err == nil:
    if len(data) > 0 {              // zero-byte read → root stays empty {}
        json.Unmarshal(data, &root)
    }
case os.IsNotExist(err):
    // treat as empty; we'll create it
}
```

If the read catches Claude Code mid-truncate and returns zero bytes with no error, `root` stays empty. AO then atomically renames a config containing **only** `{"projects":{"<ws>":{"hasTrustDialogAccepted":true}}}` over the user's real `~/.claude.json`, destroying `oauthAccount`, `userID`, `hasCompletedOnboarding`, and every other project — logging the user out of Claude Code globally and wiping all per-project history.

The sibling `ensureDevinNativeWorkspaceTrusted` (`devin.go:236`) has the same unguarded read on `~/.local/share/devin/cli/trusted_workspaces.json`.

### Reproduction

A unit test mirroring #3756's `TestEnsureWorkspaceTrustedRefusesZeroByteConfig`, but calling `ensureDevinWorkspaceTrusted` with a zero-byte config, fails on `main`: the call returns `nil` and the file is overwritten.

### Fix

Apply #3756's guard to both Devin functions: when `len(bytes.TrimSpace(data)) == 0`, refuse to overwrite rather than replacing the real config with the trust-only one. I have a patch and tests ready (fail without the guard, pass with it; full `devin` package and `go vet` clean) and would like to work on this.

winklemad:
Fixes #3833.

`ensureDevinWorkspaceTrusted` and `ensureDevinNativeWorkspaceTrusted` had the same zero-byte data-loss bug as #3746 (fixed for `claudecode` in #3756). When the read of `~/.claude.json` (or Devin's `trusted_workspaces.json`) catches another writer mid-truncate and returns zero bytes with no error, `root` stayed empty and AO atomically replaced the user's real config — `oauthAccount`, `userID`, `projects` history — with one containing only the trust entry, logging the user out of Claude Code globally and wiping per-project history.

#3756 fixed only `claudecode.go`; the Devin adapter reaches the same `~/.claude.json` via the same unguarded read. This applies the identical guard to both Devin functions: when `len(bytes.TrimSpace(data)) == 0`, refuse to overwrite.

Tests: adds `TestEnsureDevinWorkspaceTrustedRefusesZeroByteConfig` and `TestEnsureDevinNativeWorkspaceTrustedRefusesZeroByteConfig`, mirroring #3756's `TestEnsureWorkspaceTrustedRefusesZeroByteConfig`. They fail on `main` (config overwritten) and pass with the guard. Full `devin` package and `go vet` clean.

### DIFF
--- a/backend/internal/adapters/agent/devin/devin.go
+++ b/backend/internal/adapters/agent/devin/devin.go
@@ -21,6 +21,7 @@
 package devin
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -241,10 +242,16 @@ func ensureDevinNativeWorkspaceTrusted(configPath, workspacePath string) error {
 	data, err := os.ReadFile(configPath)
 	switch {
 	case err == nil:
-		if len(data) > 0 {
-			if err := json.Unmarshal(data, &root); err != nil {
-				return fmt.Errorf("devin: parse %s: %w", configPath, err)
-			}
+		if len(bytes.TrimSpace(data)) == 0 {
+			// A zero-byte read means we may have caught some other writer
+			// mid-truncate. Treat it like a corrupt file, not a missing one:
+			// refuse to write rather than silently replacing the user's real
+			// config (oauthAccount, projects history, etc.) with one containing
+			// only the trust entry.
+			return fmt.Errorf("devin: %s is empty; refusing to overwrite", configPath)
+		}
+		if err := json.Unmarshal(data, &root); err != nil {
+			return fmt.Errorf("devin: parse %s: %w", configPath, err)
 		}
 	case os.IsNotExist(err):
 		// Treat as empty config; we'll create it.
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@winklemad Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

winklemad:
Still relevant — this is the same zero-byte data-loss bug as #3746/#3756, just on the `agent-devin` trust paths (`ensureDevinWorkspaceTrusted` / `ensureDevinNativeWorkspaceTrusted`). When a read of `~/.claude.json` catches another writer mid-truncate and returns zero bytes with no error, AO would atomically replace the user's real config with an empty one. The fix refuses to write when the read came back empty-without-error, with tests covering it. Happy to rebase if a reviewer can take a look.

PR #3838, opened 2026-08-11
feat: promote queued turn to steer active turn

nikhilachale:
Dedicated promote endpoint for queued human messages.

- Backend: POST /sessions/{sessionId}/conversation/turns/{turnId}/steer loads queued turn (text, clientId, delivery_content_json), verifies turn belongs to session/conversation, still queued, and active steer-capable turn is running; parses ACP-style content blocks (text/image/resource) with image MIME allowlist and size caps; providers reject unsupported content instead of dropping; steers via ActiveTurnContentSteerer (attachments) or Guard.NudgeCoordination (text) at write boundary; only on Sent retires queued turn as promoted/interrupted via CAS; serialized with sendMu shared with /send to prevent queue-drain race. Typed errors: TURN_NOT_FOUND (404), TURN_NOT_QUEUED (409), NO_STEERABLE_TURN (409), UNSUPPORTED_ATTACHMENT (422).
- Storage: 0032_conversation_turns table + queries + store
- Ports: SteerContentBlock + ActiveTurnContentSteerer
- API: OpenAPI + frontend schema regenerated
- Frontend: Steer now button beneath queued human messages

Branch: ao/agent-orchestrator-179/root at d7e968d


nikhilachale:
Screenshot 2026-08-12 at 04 25 49

https://github.com/user-attachments/assets/9e5f3d71-81cd-4d61-9d04-4731790c76a5

backend/internal/service/chat/steer.go
```diff
@@ -84,6 +101,112 @@ func (s *Service) Steer(
...
+			return PromoteQueuedTurnResult{}, fmt.Errorf("decode queued chat content: %w", err)
+		}
+	}
+	msg := ports.ChatUserMessage{
+		Text: queued.Text, Content: content, Origin: queued.Origin,
+		ClientMessageID: queued.ClientMessageID,
+	}
+	ref, err := steerer.Steer(ctx, target, msg)
+	if err != nil {
+		release()
```
illegalcall: **[P1] Do not release the reservation for an ambiguous provider error.** Codex writes `turn/steer` before waiting for the response, so a timeout or connection loss can happen after the provider accepted the guidance. Example: the provider accepts queued message Q, the acknowledgement is lost, this `release()` makes Q drain later as a new turn, and the agent receives Q twice. Release only typed, definite pre-acceptance refusals; settle transport/unknown outcomes as `ErrPromotionUncertain` without making the source drainable.

backend/internal/service/chat/steer.go
```diff
@@ -84,6 +101,112 @@ func (s *Service) Steer(
...
+
+	var content []ports.ChatContent
+	if queued.DeliveryContentJSON != "" {
+		if err := json.Unmarshal([]byte(queued.DeliveryContentJSON), &content); err != nil {
+			release()
+			return PromoteQueuedTurnResult{}, fmt.Errorf("decode queued chat content: %w", err)
+		}
+	}
+	msg := ports.ChatUserMessage{
+		Text: queued.Text, Content: content, Origin: queued.Origin,
```
illegalcall, line 190: **[P1] Enforce the human-origin invariant before contacting the provider.** The durable origin is loaded here but never checked. Example: while a turn is active, an automation-origin message is queued; a direct caller posts that turn ID to this endpoint, and AO forwards it as mid-turn user guidance even though only human-originated queued turns are promotable. Require `queued.Origin == domain.MessageOriginHuman` and return/release safely otherwise; frontend filtering is not an API boundary.

backend/internal/service/chat/steer.go
```diff
@@ -84,6 +101,112 @@ func (s *Service) Steer(
...
+		return PromoteQueuedTurnResult{}, fmt.Errorf("%w: %s", ErrTurnNotQueued, turnID)
+	}
+
+	target, ok := c.awaitAcknowledgedTurn(ctx)
+	if !ok {
+		return PromoteQueuedTurnResult{}, ErrNoActiveTurn
+	}
+	queued, err := c.store.ReserveQueuedTurnForPromotion(ctx, c.conversation.ID, turnID, c.now())
+	if err != nil {
+		return PromoteQueuedTurnResult{}, fmt.Errorf("%w: %s: %w", ErrTurnNotQueued, turnID, err)
```
illegalcall, line 157: **[P2] Preserve internal reservation errors instead of converting all of them to a queue conflict.** `ReserveQueuedTurnForPromotion` can fail because the compare-and-set missed, but it can also fail on an SQLite update/read error. Example: `database is locked` is wrapped as `ErrTurnNotQueued`, so the route returns 409 and tells the user the message is no longer queued even though the daemon actually had an internal failure. Map only a typed not-available sentinel to `ErrTurnNotQueued`; propagate other errors to the shared 500 envelope.

backend/internal/service/chat/steer.go
```diff
@@ -84,6 +101,112 @@ func (s *Service) Steer(
...
+			return PromoteQueuedTurnResult{}, ErrNoActiveTurn
+		case errors.Is(err, ports.ErrChatTurnNotSteerable):
+			return PromoteQueuedTurnResult{}, fmt.Errorf("%w: %w", ErrTurnNotSteerable, err)
+		case errors.Is(err, ports.ErrChatSteerContentUnsupported):
+			return PromoteQueuedTurnResult{}, fmt.Errorf("%w: %w", ErrSteerContentUnsupported, err)
+		default:
+			return PromoteQueuedTurnResult{}, classify(fmt.Errorf("steer queued turn %s: %w", turnID, err))
+		}
+	}
+	landed := ref.ProviderTurnID
```
illegalcall, line 214: **[P2] Reject a provider response that names a different turn.** The request targets the provider-acknowledged `target`, but any nonempty returned ID becomes `landed`. Example: AO sends `expectedTurnId=turn-A`, receives `turn-B` because of a provider/protocol bug, and then attributes the activity and promotion to B. The contract says guidance must fail rather than land on a different turn; verify the adapter response equals `target` and treat a mismatch as an uncertain delivery outcome.

backend/internal/storage/sqlite/queries/conversations.sql
```diff
@@ -535,10 +537,65 @@ FROM conversation_turns
 JOIN conversation_messages
     ON conversation_messages.turn_id = conversation_turns.id
     AND conversation_messages.role = 'user'
-WHERE conversation_turns.conversation_id = ? AND conversation_turns.state = 'queued'
+WHERE conversation_turns.conversation_id = ?
+  AND conversation_turns.state = 'queued'
+  AND conversation_turns.promotion_started_at IS NULL
```
illegalcall, line 542: **[P2] Add dedicated abandoned-reservation recovery.** Excluding reserved rows from drain is correct, but startup currently settles all orphaned queued/running turns with the generic `controller ended before the turn completed` error and leaves `promotion_started_at` populated. Example: the daemon dies after sending Q but before finalization; after restart Q is failed, but it has neither the required promotion-uncertain guidance nor cleared reservation metadata. Recover reserved rows separately by setting the promotion-uncertain failure and clearing `promotion_started_at`, with a restart test proving no automatic redelivery.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -1823,6 +1849,79 @@ function EmptyState({ harness }: { harness: string }) {
...
+							<CornerDownRight aria-hidden="true" className="size-3.5 shrink-0 text-muted-foreground" />
+							<span className="min-w-0 flex-1 truncate text-sm text-foreground" title={message.text}>
+								{message.text}
+							</span>
+							{onSteer ? (
+								<Button
+									type="button"
+									size="sm"
+									variant="ghost"
+									disabled={Boolean(steeringTurnId)}
```
illegalcall, line 1905: **[P2] Scope pending state to the selected card.** This disables every queue action whenever one request is pending, and the guard in `steer()` also blocks every other ID. Example: Q1 is waiting on a slow provider response; Q2's unrelated `Steer` button becomes disabled even though the dock contract says only Q1 should show `Steering…` and be disabled. Track pending state per turn and use `disabled={steering}` (with the invocation guard keyed to that turn).

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1277,10 +1279,30 @@ function ReauthRow({ activity }: { activity: ConversationActivity }) {
  */
 export function SteerMessage({ activity }: { activity: ConversationActivity }) {
 	const text = activity.detail?.text ?? activity.summary;
+	const images = (activity.detail?.content ?? []).filter(
```
illegalcall: **[P2] Render all preserved steer content, not only images.** ACP can promote `resource_link` and embedded `resource` blocks, and the activity keeps those blocks, but this filter drops them. Example: a queued message containing a resource link is successfully steered into the agent; the destination timeline then shows only the text and silently omits the delivered resource. Reuse the human-message attachment/content rendering rules for the complete structured content.

backend/internal/httpd/controllers/conversation_steer.go
```diff
@@ -75,6 +83,40 @@ func (c *ConversationsController) steer(w http.ResponseWriter, r *http.Request)
...
+		SourceTurnID: result.SourceTurnID, ProviderTurnID: result.ProviderTurnID,
+		ActivityID: result.ActivityID,
+	})
+}
+
+func writeQueuedTurnSteerError(w http.ResponseWriter, r *http.Request, err error) {
+	switch {
+	case errors.Is(err, chatsvc.ErrTurnNotQueued):
+		envelope.WriteAPIError(w, r, http.StatusConflict, "conflict",
+			"CHAT_TURN_NOT_QUEUED", "that message is no longer queued", nil)
```
illegalcall, line 110: **[P2] Match the documented wire codes.** The promotion contract specifies `TURN_NOT_QUEUED` and `TURN_NOT_FOUND`, while this path emits `CHAT_TURN_NOT_QUEUED` and reuses the shared `CHAT_TURN_NOT_FOUND`. Example: a client retries after a successful promotion and switches on `TURN_NOT_QUEUED`; it receives the prefixed code instead and misses its intended refresh/recovery path. Align both endpoint outcomes with the published contract and regenerate artifacts if necessary.

backend/internal/httpd/controllers/conversation_steer.go
```diff
@@ -38,6 +39,13 @@ type SteerConversationResponse struct {
 	ActivityID string `json:"activityId,omitempty"`
 }
 
+// PromoteQueuedTurnResponse reports where one durable queued turn landed.
+type PromoteQueuedTurnResponse struct {
```
illegalcall, line 43: **[P2] Keep response DTOs in the repository's designated API-contract source.** `AGENTS.md` names `controllers/dto.go` as the source for request/response shapes. Example: a later contract change audits or updates response DTOs in `dto.go`, but misses `sourceTurnId` because this response is defined beside the handler. Move `PromoteQueuedTurnResponse` there; the generated schema can retain the same shape.

backend/internal/httpd/controllers/conversation_steer_test.go
```diff
@@ -270,3 +303,41 @@ func TestSteerRouteIsNotImplementedWithoutAChatService(t *testing.T) {
 		t.Errorf("status = %d, want 501", resp.StatusCode)
 	}
 }
+
+// The turn-scoped route identifies durable content; it carries no request body
+// that could replace what AO already queued.
+func TestQueuedTurnSteerRoutePromotesTheSelectedTurn(t *testing.T) {
```
illegalcall, line 309: **[P3] Add route-level error-envelope coverage.** This new endpoint is tested only for success. Example: the reservation layer returns an internal SQLite error but the controller accidentally exposes it as 409 `CHAT_TURN_NOT_QUEUED`; this suite still passes. Add table cases asserting status, error code, and envelope/request-ID behavior for not queued, promotion uncertain, not found, and an untyped internal error.

backend/internal/httpd/controllers/conversation_steer.go
```diff
@@ -16,6 +16,7 @@ import (
 // steerPath is the one route this file owns, named once so the not-implemented
 // answer and the spec cannot drift apart.
 const steerPath = "/api/v1/sessions/{sessionId}/conversation/steer"
+const queuedTurnSteerPath = "/api/v1/sessions/{sessionId}/conversation/turns/{turnId}/steer"
```
illegalcall, line 19: **[P3] Update the route invariant comment now that this file owns two routes.** It currently says `steerPath` is the one route the file owns. Example: a later spec/501 update follows that comment and updates only the original route, allowing the queued-turn path to drift. Describe both constants (or centralize the shared route metadata) so the comment remains true.

illegalcall (changes requested):
I rechecked these against the current PR head and am requesting changes. The highest-risk issue is ambiguous provider delivery: a lost acknowledgement can currently release the source back to the queue and deliver the same instruction twice. The inline comments include concrete examples and suggested boundaries for each confirmed issue.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
illegalcall: Fixed in 4425cc17d. Landed steer activities now use the same shared content renderer as preserved human-message content: image blocks render previews, while `resource_link` and embedded `resource` blocks render labeled URI-backed chips. The regression test includes all three block types. Verified with 22 focused tests, the full frontend suite (2237 passed, 1 skipped), TypeScript, and a production Vite build.

backend/internal/service/chat/steer.go
illegalcall: Fixed in 7892ba877. Promotion now releases the reservation only for typed, definite refusals (`no steerable turn`, `turn not steerable`, unsupported content, or a classified provider refusal). Any transport/unknown error settles the source as failed with `ErrPromotionUncertain` using a cancellation-independent context, preserves the original cause, and leaves the source non-drainable. The regression test cancels the request context immediately after the simulated provider call fails, then verifies the failed/uncertain state, empty queue, rejected retry, and exactly one provider steer attempt. Verified with the full chat package, focused race tests, `go vet`, and all backend packages.

illegalcall: Fixed in 7892ba877. Promotion now releases only typed, definite refusals. Transport/unknown errors settle the source failed with `ErrPromotionUncertain` through a cancellation-independent context, preserve the original cause, and leave the source non-drainable. The regression test cancels the request context after the simulated provider call, then verifies the failed/uncertain state, empty queue, rejected retry, and exactly one steer attempt. Verified with the full chat package, focused race tests, `go vet`, and all backend packages.

backend/internal/service/chat/steer.go
illegalcall: Fixed in 0dd6363e4. Promotion now checks the durable origin returned by ReserveQueuedTurnForPromotion before decoding content or calling the provider. Non-human sources release their reservation and return the existing ineligible/queued conflict, leaving the automation turn in its original queue position. The regression test calls the service directly, asserts zero provider steer attempts, and verifies the same automation turn remains at the queue head.

illegalcall (approved):
Re-reviewed at current head 0dd6363e4. All requested changes are addressed:

- 7892ba877 settles ambiguous/transport provider outcomes as promotion-uncertain instead of releasing the source for redelivery.
- 4425cc17d renders the complete preserved steer content, including resource links and embedded resources.
- 0dd6363e4 enforces the durable human-origin invariant before provider contact and safely releases ineligible reservations.

Each path has regression coverage, and all 10 required checks pass on this head.

PR #3841, opened 2026-08-11
fix: honor current review injection policy

### LINKED ISSUE #3840: Review feedback can be injected after the session toggle is turned off
## Bug
Turning off **Automatically send review feedback** does not reliably stop review feedback from being injected into the worker. Feedback captured while the toggle was enabled remains marked injectable and can be delivered after the session-level toggle is off.

**Source:** live AO session report from @neversettle17-101 on 2026-08-11  
**Analyzed against:** `aab4c973d` (`origin/main`)  
**AO daemon:** source build, port `3001`  
**Confidence:** High  
**Priority:** P2 — the worker can act on feedback the user explicitly disabled, but delivery can be avoided by leaving the toggle off before any PR/review is observed.

## Reproduction
1. Start a worker with automatic review feedback enabled (the default).
2. Claim or observe a PR containing a changes-requested review, causing its review rows to be persisted with `auto_inject_review = true`.
3. Turn **Automatically send review feedback** off before a later PR observation/delivery pass.
4. Trigger or wait for the lifecycle PR observation.
5. Observe that the changes-requested review body is still sent into the worker conversation.

## Live evidence
For session `agent-orchestrator-97`, the session row/API reported `autoInjectReview: false`, while both changes-requested review rows for PR 3692 retained `auto_inject_review = 1`. Both review bodies were injected into the active worker.

## Root Cause
The implementation intentionally snapshots the session policy onto each review/comment when it is first observed:

- `backend/internal/service/session/claim_pr.go` copies `SessionRecord.AutoInjectReview` into persisted review/comment rows.
- `backend/internal/lifecycle/reactions.go` decides delivery from `review.AutoInjectReview` / `comment.AutoInjectReview`, but does not also require the current `SessionRecord.AutoInjectReview` value after loading the session.
- `backend/internal/lifecycle/reactions.go::ApplyReviewBatch` similarly loads the current session but never gates delivery on its current toggle.
- `backend/internal/service/review/review.go::deliverableRuns` uses only the review-run snapshot.

The snapshot is useful to prevent feedback observed while disabled from being retroactively injected after re-enabling. However, using only the snapshot means disabling the live session is not a kill switch for feedback already observed. This conflicts with the UI promise in `frontend/src/renderer/i18n/en.json`: “When disabled, review comments will not be sent to the worker agent but will be displayed.”

## Fix
Treat injection eligibility as the conjunction of both policies:

```text
persisted_feedback.auto_inject_review && current_session.auto_inject_review
```

Keep the per-feedback snapshot so re-enabling does not inject feedback captured while disabled, but add a current-session guard in both external SCM observation delivery and AO review-run delivery. Add race/regression tests covering:

1. captured enabled → toggle off before lifecycle delivery → not injected;
2. captured disabled → toggle on later → not injected;
3. captured enabled and still enabled → injected;
4. AO internal review batch follows the same rules.

## Related
- [#3468](https://github.com/Untrivial-ai/agent-orchestrator/issues/3468) — original toggle request.
- [PR #3709](https://github.com/Untrivial-ai/agent-orchestrator/pull/3709) — introduced the toggle and per-feedback snapshot behavior.

neversettle17-101:
## Summary
- require the current session review-injection policy for persisted SCM feedback delivery
- gate AO review-run selection and lifecycle delivery on the current policy while preserving per-feedback snapshots
- add regressions for disabling after feedback capture across SCM and AO review batches

Closes #3840

## Tests
- `cd backend && go test ./internal/lifecycle ./internal/service/review`
- `cd backend && go test ./...` *(pre-existing/environment failures in agent auth timeout tests and CLI tests; touched packages pass)*

## Risk
Low. Changes are limited to delivery eligibility checks and focused tests; persisted snapshot behavior is unchanged.

### DIFF
--- a/backend/internal/lifecycle/reactions.go
+++ b/backend/internal/lifecycle/reactions.go
@@ -56,7 +56,7 @@ func (m *Manager) ApplyReviewBatch(ctx context.Context, workerID domain.SessionI
 	if err != nil || !ok {
 		return ReviewDeliveryNoop, err
 	}
-	if cannotNudge(rec) {
+	if cannotNudge(rec) || !rec.AutoInjectReview {
 		return ReviewDeliveryNoop, nil
 	}
 	if m.guard == nil {
@@ -205,7 +205,7 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 		}
 	}
 
-	if hasUnresolvedComments(o.Comments) {
+	if rec.AutoInjectReview && hasUnresolvedComments(o.Comments) {
 		comments := unresolvedReviewComments(o.Comments)
 		for _, comment := range comments {
 			if !comment.AutoInjectReview {
@@ -227,7 +227,7 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
 		}
 	}
 
-	if o.Review == domain.ReviewChangesRequest {
+	if rec.AutoInjectReview && o.Review == domain.ReviewChangesRequest {
 		for _, review := range reviews {
 			if review.State != domain.ReviewChangesRequest || !review.AutoInjectReview {
 				continue
--- a/backend/internal/lifecycle/manager_test.go
+++ b/backend/internal/lifecycle/manager_test.go
@@ -1746,6 +1746,28 @@ func TestPRObservation_ReviewFeedbackNotInjectedWhenDisabled(t *testing.T) {
 	}
 }
 
+func TestPRObservation_ReviewFeedbackNotInjectedAfterSessionDisabled(t *testing.T) {
+	m, st, msg := newManager()
+	rec := working("mer-1")
+	rec.AutoInjectReview = false
+	st.sessions[rec.ID] = rec
+	st.comments["pr1"] = []domain.PullRequestComment{{ID: "1", Author: "alice", Body: "captured while enabled", AutoInjectReview: true}}
+	st.reviews["pr1"] = []domain.PullRequestReview{{ID: "r1", Author: "alice", State: domain.ReviewChangesRequest, Body: "also captured while enabled", AutoInjectReview: true}}
+	o := ports.PRObservation{
+		Fetched: true,
+		URL:     "pr1",
+		CI:      domain.CIFailing,
+		Checks:  []ports.PRCheckObservation{{Name: "build", Status: domain.PRCheckFailed, LogTail: "boom"}},
+		Review:  domain.ReviewChangesRequest,
+	}
+	if err := m.ApplyPRObservation(ctx, rec.ID, o); err != nil {
+		t.Fatal(err)
+	}
+	if len(msg.msgs) != 1 || !strings.Contains(msg.msgs[0], "boom") || strings.Contains(msg.msgs[0], "captured while enabled") {
+		t.Fatalf("messages = %v, want CI only after current session policy is disabled", msg.msgs)
+	}
+}
+
 func TestPRObservation_MixedPersistedCommentDecisions(t *testing.T) {
 	m, st, msg := newManager()
 	st.sessions["mer-1"] = working("mer-1")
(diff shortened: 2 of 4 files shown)


backend/internal/lifecycle/reactions.go
```diff
@@ -56,7 +56,7 @@ func (m *Manager) ApplyReviewBatch(ctx context.Context, workerID domain.SessionI
 	if err != nil || !ok {
 		return ReviewDeliveryNoop, err
 	}
-	if cannotNudge(rec) {
+	if cannotNudge(rec) || !rec.AutoInjectReview {
```
neversettle17-101, line 59: **Medium — Re-check the policy at the actual paste boundary.** This reads `AutoInjectReview` only at method entry, but `sendOnce` can perform dedup/persistence I/O before `guard.Nudge`, and the setting can be disabled during that window. The existing guard deliberately re-reads session state immediately before writing, but it only checks nudgeability, not this policy, so a review captured while enabled can still be pasted after the user disables injection. The same stale-session window exists in `ApplyPRObservation`, where reviews/comments are loaded after the session read and multiple nudges may be sent. Please make the just-in-time guard policy-aware for review nudges (without suppressing CI/merge nudges), and add a flip-between-read-and-send test.

neversettle17-101 (commented):
## What This PR Does

Adds current-session policy checks before delivering persisted PR review feedback and internal review batches, with regression tests for sessions disabled after feedback was captured.

## Findings

- **Medium**: The policy is still sampled before the final send, so a concurrent disable can leak one more review nudge. See the inline comment.

## Testing

The added unit tests cover the already-disabled case for observation and internal-review delivery. They do not cover a policy change between the initial session read and the guarded paste. Existing CI is green (Go build/test, lint, API drift, CLI E2E, and gitleaks). Tests were not run locally.

**Verdict: Needs changes.**

backend/internal/lifecycle/reactions.go
neversettle17-101, line 59: This early snapshot does not guarantee the policy is current at delivery time: `sendOnce` later calls `guard.Nudge`, whose just-in-time session read does not check `AutoInjectReview`. If the user turns this policy off after this read (or while `sendOnce` loads persisted signatures), the review batch is still pasted and marked delivered. Use a review-specific guard path that rechecks the policy immediately before writing, and apply it to the SCM review/comment nudge paths as well.

neversettle17-101 (commented):
## Changes requested

The current-policy check is only performed before the dedup work. Review feedback can still be pasted after the user disables review injection in the intervening window.

backend/internal/lifecycle/reactions.go
whoisasx, line 59: **[P2] Re-check the review-injection policy at the final pane-write boundary.** This read only samples `AutoInjectReview` when processing starts. If the user disables review injection while `sendOnce` performs its deduplication/persistence work, `guard.Nudge` re-reads the session but does not enforce this policy, so one final review body is still pasted. I reproduced that sequence by disabling the policy on the second `GetSession`: the outcome remained `sent` and the fake messenger received the body. Please enforce the review policy using the same just-in-time session record used immediately before the pane write, and add race tests for both AO review batches and SCM review/comment delivery.

whoisasx (changes requested):
## Summary

The new current-session checks correctly cover sessions disabled before processing starts and preserve persisted-feedback snapshot behavior. However, the policy remains stale by the time the message is written: disabling review injection after the initial read can still allow one final review nudge.

I reproduced this with a test that disables `AutoInjectReview` on the second `GetSession` call; `ApplyReviewBatch` returned `sent` and delivered the body. Please move the review policy check into a review-specific guard at the final pane-write boundary and cover both AO review batches and SCM review/comment delivery with a flip-between-read-and-send test.

i-trytoohard:
@neversettle17-101 Marking as stale: no activity for 27 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

PR #3843, opened 2026-08-11
feat(mobile) : chat ui improvements and optimisations

Prasad-D-Ware:
## What

  Polishes the mobile Chat and Terminal UI experience and improves repeated conversation loading.

  - Standardizes minimal navigation chrome, haptics, and mode-aware header actions.
  - Aligns Spawn Agent with desktop project, agent, model, and optional-task behavior.
  - Moves chat settings, skills, files, and conversation navigation to native sheets.
  - Orders agent sessions by actionability.
  - Fixes Android keyboard avoidance in Spawn Agent.
  - Caches conversation snapshots and catalogs while preventing stale asynchronous results from overwriting newer state.
  - Preserves correct empty-chat orientation and latest-message positioning.
  - Adds `npm run start:perf`.
  - Normalizes the mobile lockfile indentation.

  ## Why

  Mobile navigation and selection controls had inconsistent platform behavior. Repeatedly opening conversations could also become sluggish or allow stale asynchronous responses to
  update the current screen.

  Small and empty conversations could render incorrectly following the inverted-list performance work.

  ## How

  - Shared native sheet routes and minimal header primitives are used across mobile.
  - Conversation data uses bounded caches, scoped refresh events, and request-generation guards.
  - Settings sheets update only after successful saves and expose refresh or failure states.
  - Spawn model catalogs remain cached until their project or agent source changes.
  - Delayed skill and file picker requests are discarded when their trigger is no longer active.

  The broader Android cold-start event-log relay redesign is intentionally deferred. Production-like dev-client testing did not reproduce the startup freeze.

  ## Testing

  - `cd packages/mobile && npm test` — 51 files and 420 tests passed.
  - `cd packages/mobile && npm run typecheck`
  - `cd packages/mobile && npm install --package-lock-only --ignore-scripts --offline`
  - `git diff --check`
  - Tested on a physical Android device across:
    - large, small, and empty conversations
    - repeated conversation opening
    - back button and menu responsiveness
    - latest-message positioning
    - native settings, skills, and file sheets
    - settings cache refresh and failed saves
    - stale picker cancellation
    - Spawn Agent keyboard and model-selection behavior

  ## Checklist

  - [x] Branched from `main`
  - [x] Follows `AGENTS.md` conventions
  - [ ] CI checks pending

### DIFF
--- a/packages/mobile/app/spawn.tsx
+++ b/packages/mobile/app/spawn.tsx
@@ -3,7 +3,9 @@ import { Feather } from "@expo/vector-icons";
 import { useEffect, useMemo, useState } from "react";
 import {
 	InteractionManager,
+	Keyboard,
 	KeyboardAvoidingView,
+	LayoutAnimation,
 	Platform,
 	Pressable,
 	ScrollView,
@@ -20,6 +22,7 @@ import { classifyConnectionFailure, describeConnectionFailure } from "../lib/con
 import { chatErrorCopy, isChatPreflightError } from "../lib/chatError";
 import { haptics } from "../lib/haptics";
 import { agentSheetRoute, modelSheetRoute, projectSheetRoute } from "../lib/sheetResult";
+import { screenKeyboardAvoidance } from "../lib/session/keyboardInset";
 import { modelOverride, resolveSpawnModel } from "../lib/spawnModel";
 import { useApp } from "../lib/store";
 import type { Theme } from "../lib/theme";
@@ -45,11 +48,33 @@ export default function SpawnModal() {
 	const [modelError, setModelError] = useState<string>();
 	const [busy, setBusy] = useState(false);
 	const [error, setError] = useState<string | null>(null);
+	const [keyboardHeight, setKeyboardHeight] = useState(0);
 
 	const [catalog, setCatalog] = useState<AgentCatalog | null>(null);
 	const [catalogError, setCatalogError] = useState<string | null>(null);
 	const [loading, setLoading] = useState(true);
 	const [offerTUI, setOfferTUI] = useState(false);
+
+	useEffect(() => {
+		if (Platform.OS !== "android") return;
+		const avoidance = screenKeyboardAvoidance("android", 0);
+		const animate = (duration?: number) => LayoutAnimation.configureNext({
+			duration: duration || 250,
+			update: { type: LayoutAnimation.Types.keyboard },
+		});
+		const show = Keyboard.addListener(avoidance.showEvent, (event) => {
+			animate(event.duration);
+			setKeyboardHeight(event.endCoordinates.height);
+		});
+		const hide = Keyboard.addListener(avoidance.hideEvent, (event) => {
+			animate(event?.duration);
+			setKeyboardHeight(0);
+		});
+		return () => {
+			show.remove();
+			hide.remove();
+		};
+	}, []);
 
 	// Seed from the active project, or the only project. Mirrors the store's
 	// `targetProject()`; kept here because the screen needs it as UI state to
@@ -151,7 +176,15 @@ export default function SpawnModal() {
 	};
 
 	return (
-		<KeyboardAvoidingView style={styles.screen} behavior={Platform.OS === "ios" ? "padding" : undefined}>
+		<KeyboardAvoidingView
+			style={[
+				styles.screen,
+				Platform.OS === "android" && keyboardHeight > 0
+					? screenKeyboardAvoidance("android", keyboardHeight).rootStyle
+					: undefined,
+			]}
+			behavior={Platform.OS === "ios" ? "padding" : undefined}
+		>
 			<ScrollView contentContainerStyle={{ padding: 16, paddingBottom: 40 }} keyboardShouldPersistTaps="handled">
 				<Text style={styles.lead}>
 					Spawn a worker agent. It gets its own isolated workspace, then starts on the task you give it.
--- a/packages/mobile/app/spawn.tsx
+++ b/packages/mobile/app/spawn.tsx
@@ -15,11 +15,12 @@ import {
 import { AgentLogo } from "../lib/AgentLogo";
 import { agentErrorCopy } from "../lib/agentError";
 import { defaultAgent, rankAgents } from "../lib/agentPicker";
-import { ApiError, getAgents, getSettings, type AgentCatalog, type SessionMode } from "../lib/api";
+import { ApiError, getAgentModels, getAgents, getProject, getSettings, type AgentCatalog, type AgentModelCatalog, type ProjectDetail, type SessionMode } from "../lib/api";
 import { classifyConnectionFailure, describeConnectionFailure } from "../lib/connectionError";
 import { chatErrorCopy, isChatPreflightError } from "../lib/chatError";
 import { haptics } from "../lib/haptics";
-import { agentSheetRoute, projectSheetRoute } from "../lib/sheetResult";
+import { agentSheetRoute, modelSheetRoute, projectSheetRoute } from "../lib/sheetResult";
+import { modelOverride, resolveSpawnModel } from "../lib/spawnModel";
 import { useApp } from "../lib/store";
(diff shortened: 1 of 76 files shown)


Prasad-D-Ware:
the major line of code is because of package-lock indentation fixes, its needed in this pr for not come up in future. @illegalcall

packages/mobile/lib/chat/asyncValueCache.ts
```diff
@@ -0,0 +1,72 @@
+type CachedValue<V> = { value: V; expiresAt: number };
```
illegalcall, line 1: Im not opposed to this, but see if there is some cache library which handles this, if youre sure your requirements and think this is sufficient, then no problem. Otherwise consider a caching library

packages/mobile/app/spawn.tsx
```diff
@@ -86,25 +117,60 @@ export default function SpawnModal() {
...
 		setBusy(true);
 		setError(null);
 		setOfferTUI(false);
 		try {
 			const session = await spawn({
 				projectId: projectId ?? undefined,
 				prompt: prompt.trim() || undefined,
-				issueId: name.trim() || undefined,
 				harness: harness || undefined,
+				model: modelOverride(displayedModel, resolvedModel, modelTouched),
```
illegalcall, line 201: This value reaches `SpawnConfig.AgentConfig.Model`, but the Chat launch path rebuilds config from project settings and never applies the request override. Default Chat spawns therefore ignore the selected model; please merge the override into `launchChatController`.

packages/mobile/app/spawn.tsx
```diff
@@ -86,25 +117,60 @@ export default function SpawnModal() {
 	const agents = useMemo(() => mode === "chat" ? allAgents.filter((agent) => chatHarnesses.includes(agent.id)) : allAgents, [allAgents, chatHarnesses, mode]);
 	const selectedAgent = agents.find((a) => a.id === harness);
 	const project = projects.find((p) => p.id === projectId);
+	const projectWorkerAgent = projectDetail?.config?.worker?.agent ?? projectDetail?.agent ?? "";
```
illegalcall, line 120: `projectWorkerAgent` only affects model resolution. `harness` remains the global catalog default and is explicitly submitted, so a configured `worker.agent` can be ignored; please use it until the user changes the agent.

packages/mobile/app/spawn.tsx
```diff
@@ -86,25 +117,60 @@ export default function SpawnModal() {
...
+		resetModelSource();
+		setHarness(nextHarness);
+	};
+	const selectMode = (nextMode: SessionMode) => {
+		if (nextMode === mode) return;
+		const nextHarness = nextMode === "chat"
+			? (chatHarnesses.includes(harness) ? harness : (defaultAgent(allAgents.filter((agent) => chatHarnesses.includes(agent.id))) ?? ""))
+			: (harness || (defaultAgent(allAgents) ?? ""));
+		if (spawnModelSourceChanged({ projectId, agentId: harness }, { projectId, agentId: nextHarness })) resetModelSource();
+		else clearModelOverride();
```
illegalcall: This clears a manual model even when project and agent are unchanged. Toggling Chat/TUI, including the fallback button, therefore discards the selection; keep it unless the model source changes.

packages/mobile/lib/chat/conversationEvents.ts
```diff
@@ -0,0 +1,66 @@
...
+		return () => {
+			stopped = true;
+			controller?.abort();
+			cursorPersister.flush();
+		};
+	}, [cfg]);
+}
+
+function eventCursorKey(cfg: ServerConfig): string {
+	return `ao.chat.events.${cfg.secure ? "https" : "http"}.${cfg.host}.${cfg.httpPort}`;
```
illegalcall, line 70: This cursor is scoped only to the endpoint. If the daemon database is reset or replaced there, the old higher sequence suppresses new events until the database catches up; scope it by daemon identity or detect cursor rollback.

packages/mobile/lib/chat/ChatComposer.tsx
```diff
@@ -153,31 +167,40 @@ export function ChatComposer({
...
+	const openPicker = useCallback(async (kind: "skills" | "files", activeTrigger?: ComposerSuggestion) => {
+		const request = pickerGate.begin();
+		const loadedSkills = kind === "skills" ? await onLoadSkills() : skills;
+		const loadedFiles = kind === "files" ? await onLoadFiles() : { paths: filePaths, truncated: Boolean(filePathsTruncated) };
+		if (!pickerGate.isCurrent(request)) return;
+		if (activeTrigger) {
+			const currentTrigger = findComposerSuggestion(latestText.current, latestCursor.current);
+			if (!currentTrigger || composerSuggestionKey(currentTrigger) !== composerSuggestionKey(activeTrigger)) return;
+		}
+		const choices = kind === "skills" ? rankComposerSkills(loadedSkills, activeTrigger?.query ?? "") : rankComposerFiles(loadedFiles.paths, activeTrigger?.query ?? "");
```
illegalcall: `rankComposer*` caps this list at 80 before the sheet opens, and the sheet searches only these choices. Toolbar search cannot find entries outside that first slice; rank the full catalog in the sheet before truncating.

illegalcall (commented):
Five focused findings from the worktree review.

packages/mobile/lib/chat/asyncValueCache.ts
Prasad-D-Ware: Kept the cache helper local in commit 5e37e2a99. Its scope is limited to a bounded TTL map with in-flight request sharing, rejection eviction, and explicit invalidation, all covered by the existing cache tests. I also added a source comment documenting why an app-wide query/cache framework would add lifecycle behavior these callers do not need.

packages/mobile/app/spawn.tsx
Prasad-D-Ware: Fixed in commit 5e37e2a99. launchChatController now applies SpawnConfig.AgentConfig over the resolved project/role config via applySpawnAgentConfig, matching the TUI launch path. Added a regression test proving a request model overrides the project default for Chat spawns.

packages/mobile/app/spawn.tsx
Prasad-D-Ware: Fixed in commit 5e37e2a99. Project details are now loaded independently of the model catalog, and the initial harness resolves from config.worker.agent, then the project agent, then the available catalog order. An agentTouched guard preserves a manual selection until the project or a mode-forced incompatible agent changes; resolver tests cover the precedence.

packages/mobile/app/spawn.tsx
Prasad-D-Ware: Fixed in commit 5e37e2a99. Chat/TUI toggles no longer clear the manual model when project and agent are unchanged. The override resets only when the actual model source changes, including when a mode switch must select a different compatible agent.

packages/mobile/lib/chat/conversationEvents.ts
Prasad-D-Ware: Fixed in commit 5e37e2a99 with cursor rollback detection. The backend compares the requested cursor with LatestSeq; when it is ahead of the current database, replay restarts at 0 and advertises the effective cursor through X-AO-Event-After. Mobile consumes that value and replaces the persisted cursor before accepting new lower-sequence events. Backend SSE and mobile stream/persistence regression tests cover the reset case.

packages/mobile/lib/chat/ChatComposer.tsx
Prasad-D-Ware: Fixed in commit 5e37e2a99. The composer now passes the full skills/files catalog into the sheet. The sheet searches and ranks the full catalog first, then applies the existing 80-result cap, and renders through FlatList. Added a regression test where a file beyond the first 80 becomes visible when searched.

illegalcall (approved):

PR #3846, opened 2026-08-11
fix: remove switch-agent button from terminal tab

Annieeeee11:
## Summary
- Remove the switch-agent button from the primary terminal tab (next to the session title)
- Leave switch-agent overlay, dialog, and hooks unchanged
- Update CenterPane tests to assert the tab no longer exposes the control

## Test plan
- [ ] Open a worker session and confirm the terminal tab shows title/status only (no switch icon)
- [ ] Confirm other terminal tab controls still work (shell tabs, close, zoom)
- [ ] `npx vitest run src/renderer/components/CenterPane.test.tsx`


ronishrohan (commented):
## Review

This PR has a 0-line diff — the switch-agent button removal already landed on main via the merge commit (19577c64). This PR is now a no-op and should be closed as stale.

ronishrohan (changes requested):
## Changes Requested

This PR has a 0-line diff — the switch-agent button removal already landed on main via the merge commit (19577c64). This PR is now a no-op and should be closed.

PR #3820, opened 2026-08-10
fix(chat): refresh interface switch readiness from CDC

illegalcall:
Fixes [#3753](https://github.com/Untrivial-ai/agent-orchestrator/issues/3753).

Focused alternative to [#3771](https://github.com/Untrivial-ai/agent-orchestrator/pull/3771). This PR addresses the reported stale readiness state without changing backend handoff semantics.

## Problem

The daemon already emits `session_updated` when a provider-native session id is captured, but the renderer did not route that event to the interface-transition readiness query. Current polling eventually recovers, but readiness can lag and the event path remains disconnected from the cache it changes.

## Solution

- Route `session_updated` to the exact `sessionInterfaceTransitionQueryKey(sessionId)`.
- Only schedule that invalidation when the mounted query is waiting on `NATIVE_SESSION_MISSING`.
- Keep the existing one-second poll as a fallback for missed or reordered SSE events.
- Rename the invalidation scheduler to reflect that it now routes workspace, conversation, and transition cache updates.

## Safety

- No backend transition behavior changes.
- No synthetic or hardcoded provider conversation ids.
- No fresh-conversation fallback that could lose TUI context.
- No migration or generated API changes.
- Settled readiness and non-session CDC events do not trigger transition refetches.

## Verification

- `cd frontend && npm test -- --run src/renderer/lib/event-transport.test.ts src/renderer/hooks/useSessionInterfaceTransition.test.tsx` — 17 tests passed.
- `cd frontend && npm run typecheck` — passed.
- Native Electron isolated timing check: hook completed at 352 ms, supported readiness refetched at 647 ms, and the button enabled at 751 ms, before the one-second poll fallback.
- Manual dev Electron E2E confirmed: the Switch to Chat button enabled in place after the first Codex prompt without navigating away.

PR #3847, opened 2026-08-11
fix(chat): enable Switch to Chat on Windows — fresh-start readiness + .cmd shim resolution

bluwwi:
Two fixes that unblock the "Switch to Chat" button on Windows, where the interface transition dead-ends before a Chat session can start.

## Why

Refs [#3753](https://github.com/Untrivial-ai/agent-orchestrator/issues/3753)

On Windows the SessionStart hook does not fire for Codex (see #3820), so the native conversation id is never reported. The transition status then blocks as `NATIVE_SESSION_MISSING` and the "Switch to Chat" button stays disabled indefinitely. This is a simpler alternative to #3771 — instead of fabricating a synthetic UUID, it models the fresh-start path in transition readiness using the adapter's existing history probe.

The second fix addresses a separate Windows blocker: Claude Code ships as an npm `.cmd` shim, and Node `child_process.spawn` cannot execute `.cmd` files. Without resolving to the real `.exe`, the Chat spawn fails with `EINVAL` before the transition even begins.

## How

**Commit 1** — `fix(chat): allow fresh-start transition when adapter has history probe`

Commit 1 — fix(chat): allow fresh-start transition when adapter has history probe
If the adapter implements AgentInterfaceHandoffHistoryProbe, an empty native id is allowed through. The history probe distinguishes a real conversation from an empty TUI, and persistedNativeConversationID uses that to decide between resume and fresh start. Adapters without a probe keep the hard block. Two tests cover the new branch: fresh-start allowed with probe, and hard block without probe.

**Commit 2** — `fix(claudeacp): resolve Windows .cmd shim to .exe for Chat spawn`

`resolveClaudeExe` unwraps the npm `.cmd` shim: prefer a sibling `.exe`, else parse the shim's quoted `%dp0%` target path. 7 tests cover pass-through, cancelled context, missing file, sibling `.exe`, quoted `%dp0%` target, malformed shim, and unreadable shim. Windows-gated cases skip on other platforms.

## Testing

- `go test ./internal/adapters/chatdriver/claudeacp/... -count=1` — 11 pass, 1 skip (live)
- `go test ./internal/session_manager/... -run TestInterfaceTransition -count=1` — pass
- Verified on Windows: Codex "Switch to Chat" button enables, Claude Code Chat launches

### DIFF
--- a/backend/internal/session_manager/interface_transition.go
+++ b/backend/internal/session_manager/interface_transition.go
@@ -382,6 +382,12 @@ func (m *Manager) nativeConversationID(
 	}
 	id = strings.TrimSpace(id)
 	if !ok || id == "" {
+		// An adapter with a history probe can safely start fresh when the
+		// native id is missing; the probe distinguishes a real conversation
+		// from an empty TUI.
+		if _, ok := handoff.(ports.AgentInterfaceHandoffHistoryProbe); ok {
+			return "", handoff, nil
+		}
 		return "", handoff, fmt.Errorf("%w for %s", ErrNativeConversationMissing, rec.Harness)
 	}
 	return id, handoff, nil
--- a/backend/internal/session_manager/interface_transition_test.go
+++ b/backend/internal/session_manager/interface_transition_test.go
@@ -824,6 +824,46 @@ func TestInterfaceTransitionTUIToChatStartsFreshWhenReservedIDHasNoHistory(t *te
 	}
 }
 
+func TestInterfaceTransitionTUIToChatAllowsFreshStartWhenNativeIDMissingWithProbe(t *testing.T) {
+	manager, store, _, chat, _ := newTransitionManager(t, domain.SessionModeTUI)
+	manager.agents = singleAgent{agent: emptyTransitionAgent{}}
+	rec := store.sessions["session-1"]
+	rec.Metadata.AgentSessionID = ""
+	store.sessions["session-1"] = rec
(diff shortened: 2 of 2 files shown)


bluwwi:
Added two tests.

- `TestInterfaceTransitionTUIToChatAllowsFreshStartWhenNativeIDMissingWithProbe` — empty native id + adapter with probe → fresh start allowed
- `TestInterfaceTransitionTUIToChatBlocksWhenNativeIDMissingWithoutProbe` — empty native id + adapter without probe → blocks with `ErrNativeConversationMissing`

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale:
> **The missing-ID failure still reproduces on macOS without manually clearing the database.**  
> In a fresh session, both `provider_conversation_id` and `controller_generation` are already empty after the TUI prompt completes. The TUI → Chat transition reports `completed`, but the chat conversation starts with zero messages, and recalling the token returns `None`.
>
> This suggests the issue occurs before the transition: the Codex conversation/controller identifiers are not being captured or persisted during the initial TUI session. Please add a regression test covering a fresh TUI session with no manual ID clearing, then verify that the identifiers are populated before transition and the original conversation is resumed afterward.

bluwwi:
follow up pr #3970

PR #3848, opened 2026-08-11
fix(session): stop stale origin/main ref from inflating workspace diff file count

Vaibhaav-Tiwari:
## Summary
- The session workspace Files tab was showing far more changed files than the PR actually has (e.g. 130 vs. the real 48), and the count would degrade from correct to wrong over a PR's lifetime.
- Root cause: AO never runs `git fetch` against a session worktree, so its `refs/remotes/origin/main` tracking ref can go stale relative to the real upstream. `resolveWorkspaceCompare` recomputes the compare base against that ref on every request; when the branch later merges a newer `main` in (e.g. to resolve a conflict) without the ref moving, the recomputed merge base still "succeeds" but lands earlier than the branch's true fork point, so every unrelated commit that landed on `main` afterward leaks into the diff as if it were part of the PR.
- Fix: prefer the PR's own, independently-synced base commit (`pr.BaseSHA`, re-derived through `git merge-base` as before) whenever it's at least as advanced as the ref-based candidate, since AO's PR sync keeps that current regardless of whether the local tracking ref has been refreshed.

## Test plan
- [x] Added `TestWorkspaceFilesExcludeMainChangesMergedInAfterStaleOriginRef`, which reproduces the bug (fails on the old code, passes with the fix) by simulating a stale `origin/main` tracking ref and a branch that later merges current `main` in.
- [x] `go build ./...`
- [x] `go vet ./internal/service/session/...`
- [x] `go test ./internal/service/session/... -count=1` (all pass)

### DIFF
--- a/backend/internal/service/session/service_test.go
+++ b/backend/internal/service/session/service_test.go
@@ -574,6 +574,77 @@ func TestWorkspaceFilesRecomputesRecordedRefAfterBaseMoves(t *testing.T) {
 	}
 }
 
+// TestWorkspaceFilesExcludeMainChangesMergedInAfterStaleOriginRef reproduces a
+// file-count inflation bug: AO never runs `git fetch` against a session
+// worktree, so its refs/remotes/origin/main tracking ref can go stale relative
+// to the real upstream. When the branch later merges a newer main in (e.g. to
+// resolve a merge conflict), recomputing the compare base from that stale ref
+// still "succeeds" but lands on an earlier commit than the branch's true
+// merge point, so unrelated commits that landed on main afterward leak into
+// the diff as if they were part of the PR.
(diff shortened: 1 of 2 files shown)


illegalcall:
Review outcome: the normal linear-history stale `origin/main` bug is correctly fixed here, and the focused regression test plus current CI give me confidence in merging this PR as-is.

Approving and merging [PR #3848](https://github.com/Untrivial-ai/agent-orchestrator/pull/3848) now is reasonable. The remaining hardening for divergent/force-pushed base histories, PR metadata that does not match the local workspace `HEAD`, and rejecting raw SHA fallbacks when `git merge-base` fails is tracked separately in [issue #3857](https://github.com/Untrivial-ai/agent-orchestrator/issues/3857).

That follow-up is intentionally non-blocking and can be picked up independently after this lands.

illegalcall (approved):

PR #3854, opened 2026-08-11
fix(mobile): allow cleartext HTTP to Tailscale CGNAT range on iOS

### LINKED ISSUE #3852: [Bug]: iOS app can't connect over Tailscale (100.64/10) — ATS blocks cleartext HTTP; NSAllowsLocalNetworking only covers local addresses
### Summary

The iOS mobile app cannot connect to the daemon over Tailscale: entering the Mac's Tailscale IP (100.64.0.0/10 CGNAT range) in the manual-connect / server details screen always fails with the "can't connect" (unreachable) error, while the same setup works over the home LAN IP. Safari **on the same iPhone** reaches `http://<tailscale-ip>:3011/` just fine (gets the daemon's 401 JSON envelope), so the network path and the Connect Mobile listener are healthy — the request is being blocked inside the app.

Root cause (strongly suspected): iOS App Transport Security. `packages/mobile/app.json` declares only

```json
"NSAppTransportSecurity": { "NSAllowsLocalNetworking": true }
```

`NSAllowsLocalNetworking` exempts cleartext HTTP only for *local* addresses (RFC1918 ranges such as `192.168.x.x`/`10.x`, link-local, `.local`). Tailscale addresses in `100.64.0.0/10` are not classified as local, so ATS rejects the app's plaintext `http://100.x.x.x:3011/api/v1/sessions` fetch before it leaves the device. The failed fetch is then classified as `unreachable` in `packages/mobile/lib/connectionError.ts` and surfaces as "can't connect". Safari is exempt from ATS, which is exactly why the in-app request fails while Safari succeeds on the same phone. Android is unaffected because the Expo config sets `usesCleartextTraffic: true` globally.

This contradicts the intended UX: the manual-connect sheet explicitly advertises Tailscale (`placeholder="192.168.x.x  or  my-pc.tailXXXX.ts.net"`), and `config.ts` documents `host` as e.g. `"100.101.102.103"`.

### Steps to reproduce

1. On macOS, enable **Connect Mobile** in the AO desktop app (LAN listener comes up on `0.0.0.0:3011`).
2. Have both the Mac and an iPhone on the same tailnet (plain Tailscale, no `serve`/funnel, MagicDNS or raw IP — same result).
3. In the iOS app's manual-connect / server details screen, enter the Mac's Tailscale IP (e.g. `100.79.x.x`), port `3011`, the connection password, TLS toggle off.
4. Tap Connect.

### Expected behavior

The app connects, same as it does when given the Mac's LAN IP — Tailscale is an advertised connection method for the mobile app.

### Actual behavior

The app reports it can't connect (the "unreachable" error copy). Meanwhile, on the **same iPhone**:

- Safari can open `http://<tailscale-ip>:3011/` and receives the daemon's JSON `{"error":"unauthorized","code":"BAD_PASSWORD",...}` — listener reachable over Tailscale.
- Connecting via the LAN IP (`192.168.x.x:3011`) from the app works normally.

From the Mac itself, `curl http://<own-tailscale-ip>:3011/` also returns the expected 401 envelope, and `lsof` confirms the listener is bound to `*:3011`. So daemon/bind/Tailscale ACLs are all fine; only the in-app request to a `100.64/10` address fails.

### Suggested fix

Either of:

- Add an ATS exception that covers the app's actual use case (it exclusively talks cleartext HTTP to a user-supplied private/VPN host): `NSAllowsArbitraryLoads: true` in `expo.ios.infoPlist.NSAppTransportSecurity` (with the App Review justification that the app is a LAN/VPN client for a self-hosted daemon), or
- keep ATS strict and make the daemon side offer TLS over Tailscale out of the box (e.g. document/integrate `tailscale serve` + the app's existing "Use TLS" toggle), and have the app detect an ATS-blocked cleartext attempt to a `100.64/10` host and show copy pointing at the TLS path instead of the generic "can't connect".

A related paper cut: `connectionError.ts` deliberately excludes `100.64/10` from the Local Network hint (correct), but the resulting generic "is your phone on the same Wi-Fi?" copy is actively misleading for Tailscale users hitting this.

Workaround for other users hitting this today: `sudo tailscale serve --bg http://localhost:3011` on the Mac, then connect the app to `<machine>.<tailnet>.ts.net`, port `443`, with **Use TLS (https/wss)** enabled — HTTPS is not subject to the ATS cleartext ban.

### AO version

Desktop app 0.12.3 (daemon reports `dev`); mobile app 1.1.0 (App Store build), repro also matches current `main` sources.

### Environment

macOS arm64 (Darwin 25.6.0) desktop + iPhone (current iOS), plain Tailscale on both ends.

### Logs / extra context

```
$ lsof -nP -iTCP -sTCP:LISTEN | grep ao
ao  ... TCP 127.0.0.1:3001 (LISTEN)
ao  ... TCP *:3011 (LISTEN)

$ curl -s -o /dev/null -w "%{http_code}\n" http://100.79.x.x:3011/
401
$ curl -s http://100.79.x.x:3011/
{"error":"unauthorized","code":"BAD_PASSWORD","message":"missing or invalid connection password"}
```

### Before submitting

- [x] I searched existing issues for duplicates (closest are #953 — Tailscale slowness, and discussion #2855 — neither covers this)

ItzYanick:
## What

Add an App Transport Security exception for Tailscale's `100.64.0.0/10` CGNAT range to
`packages/mobile/app.json`, and correct the now-inaccurate cleartext claim in
`packages/mobile/README.md`.

```json
"NSAppTransportSecurity": {
	"NSAllowsLocalNetworking": true,
	"NSExceptionDomains": {
		"100.64.0.0/10": { "NSExceptionAllowsInsecureHTTPLoads": true }
	}
}
```

## Why

Fixes #3852 — the iOS app cannot connect to the desktop daemon over Tailscale. Entering the
Mac's Tailscale IP in manual connect always surfaces the *unreachable* copy, while the same
setup works over a LAN IP and Safari on the same phone reaches `http://<tailscale-ip>:3011/`
fine.

The request never leaves the device:

- RN's `fetch` goes through `NSURLSession`, which is the layer ATS governs.
- As of iOS/tvOS 17, ATS applies to IP-literal connections made through `URLSession` — older
  guidance saying IP addresses are exempt from ATS is outdated. Confirmed by Apple DTS:
  https://developer.apple.com/forums/thread/747421
- The app declared only `NSAllowsLocalNetworking`, which exempts *local* destinations (RFC1918,
  link-local, `.local`). Tailscale's `100.64.0.0/10` is not classified as local, so the cleartext
  connection is refused in-process — which is why LAN works and Safari (exempt from ATS) works.
- Android is unaffected: `usesCleartextTraffic: true` in the Expo build properties permits
  cleartext to any host, with no local/non-local distinction.

Tailscale is an advertised connection method for the mobile app (the manual-connect placeholder
is `192.168.x.x  or  my-pc.tailXXXX.ts.net`, and `config.ts` documents `host` as e.g.
`"100.101.102.103"`), so this is a gap between the shipped config and the intended UX.

## How

iOS 17+ accepts a CIDR block as an `NSExceptionDomains` key, so the exception is scoped to
exactly the CGNAT range instead of loosening ATS globally with `NSAllowsArbitraryLoads`. ATS
stays strict for every other destination; `NSAllowsLocalNetworking` is kept as-is for LAN.

Intentional omissions, to keep the diff minimal:

- No change to MagicDNS names (`*.ts.net`). Those resolve into the same range, but the app's
  own manual-connect flow is IP-based; a hostname exception can be added separately if wanted.
- No Android change — not affected.
- No app-code change; this is purely the ATS declaration plus the doc correction.

Note the exception only takes effect in a build that regenerates `Info.plist` (prebuild / EAS
build); it does not change an already-installed binary.

## Testing

- `node -e "JSON.parse(require('fs').readFileSync('packages/mobile/app.json','utf8'))"` — parses;
  indentation matches the file's existing tabs.

Full verification needs an iOS device on a tailnet, which I could not exercise end-to-end in
this environment. Repro / expected behavior for a reviewer with that setup:

1. Enable **Connect Mobile** on the desktop app (listener on `0.0.0.0:3011`).
2. Put the Mac and an iPhone on the same tailnet (plain Tailscale, no `serve`/funnel).
3. Rebuild the iOS app from this branch so the new `Info.plist` is generated
   (`npx expo run:ios --device` or an EAS build).
4. In manual connect, enter the Mac's Tailscale IP (`100.x.x.x`), port `3011`, **Use TLS** off.

Expected: the connection succeeds, as it already does with a LAN IP. With a deliberately wrong
password, the *auth* error ("Your desktop rejected the password") should appear rather than the
*unreachable* error — that distinction is the tell that the request now leaves the device.
Also worth confirming a LAN IP still connects, i.e. no regression from adding
`NSExceptionDomains` alongside `NSAllowsLocalNetworking`.

## Checklist

- [x] Branched from `main`
- [x] One focused change; links the related issue (`Fixes #3852`)
- [ ] Tests added/updated for user-visible behavior where it makes sense — n/a, this is a native
      ATS declaration in `app.json` with no JS-testable surface
- [ ] Relevant CI checks pass for the area touched — deferring to CI on this PR; locally I only
      validated that `app.json` still parses (no code paths touched)

### DIFF
--- a/packages/mobile/README.md
+++ b/packages/mobile/README.md
@@ -156,7 +156,10 @@ gitignored** — the run commands prebuild them for you.
 
 Cleartext HTTP to the bridge already works on both platforms: Android through
 `usesCleartextTraffic` in `app.json`, iOS through `NSAllowsLocalNetworking` in the prebuilt
-`Info.plist`.
+`Info.plist`. `NSAllowsLocalNetworking` only covers *local* destinations (RFC1918, link-local,
+`.local`), so plain HTTP to a Tailscale address needs its own App Transport Security exception —
+`app.json` adds one for the `100.64.0.0/10` CGNAT range. Any other non-local host has to use TLS
+(e.g. `tailscale serve` plus the app's **Use TLS** toggle).
 
 > **On `expo-dev-client`:** this package doesn't depend on it today, so the debug build
 > connects straight to Metro and has no in-app launcher or URL switcher. If you want the
--- a/packages/mobile/app.json
+++ b/packages/mobile/app.json
@@ -24,7 +24,12 @@
 				"NSLocalNetworkUsageDescription": "AO connects to Agent Orchestrator running on your computer over your local network.",
 				"NSPhotoLibraryUsageDescription": "AO uses camera-related system features to help pair with your desktop.",
 				"NSAppTransportSecurity": {
-					"NSAllowsLocalNetworking": true
+					"NSAllowsLocalNetworking": true,
+					"NSExceptionDomains": {
+						"100.64.0.0/10": {
+							"NSExceptionAllowsInsecureHTTPLoads": true
+						}
+					}
 				},
 				"LSApplicationQueriesSchemes": ["github"]
 			}


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #3859, opened 2026-08-11
fix: limit chat minimap to user prompts

nikhilachale:
## Summary
- Limit chat minimap markers to turns that contain human user prompts
- Keep non-prompt timeline updates visible in chat without making them scroll navigation targets
- Add a regression test for non-prompt activity in the minimap

## Validation
- Previously passed on the source branch: 
- Could not rerun after rebasing onto updated `upstream/main`: this checkout was missing `frontend/node_modules`, and `npm ci` failed first on a root-owned npm cache, then with npm internal error `Exit handler never called!` using an isolated temp cache.


nikhilachale:
https://github.com/user-attachments/assets/8f54ad37-80e7-4015-bc3a-a29fa15a1b25

illegalcall (approved):

PR #3861, opened 2026-08-11
fix(worktree): refresh default branch before spawn

### LINKED ISSUE #3860: bug(worktree): AO spawns worker branches from stale origin/main — no git fetch before worktree creation
## Problem

AO creates worker branches from the local `origin/main` ref without first running `git fetch`. If the machine running AO has a stale `origin/main` (e.g. last `git fetch` was days ago), the worker branch forks from an old commit.

This causes a cascade of confusing symptoms:

- **Missing files** — code recently merged to `main` is absent from the worker branch
- **Failed file-read tool calls** — the agent tries to read files that exist on real `main` but not the stale base
- **Failed test commands** — tests reference code or configs that aren't in the checkout
- **Long commit history in PRs** — GitHub compares the worker branch against the real updated `main`, showing a divergent base with noisy diff
- **Confusing "tool call failed" cards** in the Chat UI

### No `git fetch` anywhere in the spawn path

The entire `spawn → worktree` flow is local git operations only. There are zero `git fetch` calls in the Go backend. The codebase only mentions `git fetch` in **error messages** telling the user to run it manually:

```
"has no local head, no remote, and no tag — run `git fetch` then retry"
```

### The exact flow

1. `session_manager/manager.go:714` — `createSessionWorkspace()` calls `workspace.Create()`
2. `adapters/workspace/gitworktree/workspace.go:760` — `addWorktree()` calls `resolveBaseRef()`
3. `workspace.go:1015` — `resolveBaseRef()` → `resolveBaseRefFromDefault()` → `baseRefCandidates()`
4. `adapters/workspace/gitworktree/commands.go:131` — builds candidate refs:

```go
func baseRefCandidates(branch, defaultBranch string) []string {
    candidates := []string{"origin/" + branch}
    // ...
    candidates = append(candidates, "origin/"+defaultBranch, "refs/heads/"+defaultBranch)
    return append(candidates, branch)
}
```

5. `workspace.go:1065` — `refExists()` just runs `git rev-parse --verify <ref>` against **local** state

All candidates are local refs. Whatever commit `origin/main` points at locally is used as the base — no freshness check, no fetch.

```
Actual GitHub main:     A -- B -- C -- D -- E
Local origin/main:      A -- B
AO worker branch:       A -- B -- ao-work
```

## Proposed Fix

Before resolving the base ref in `addWorktree()`, refresh the default branch remote ref:

```bash
git fetch origin main
```

### Design constraints

1. **Fetch, not pull.** `git pull` would merge/rebase the user's current checkout and mutate their working branch. `git fetch` only updates remote-tracking refs.

2. **Best-effort / non-blocking.** This is the first network call in the spawn path. Fetch must not cause spawn to fail if:
   - The machine is offline / air-gapped
   - Git auth is not configured
   - The network is slow
   
   On fetch failure: log a warning, fall through to local refs. A stale base is annoying; a failed spawn because GitHub is down is a regression.

3. **Fetch only the ref, not the full repo.** Use `git fetch origin main` (surgical) not `git fetch origin` (full fetch, heavy on large repos).

4. **Respect fork workflows.** When `origin` is a fork and `upstream` is canonical, fetching `origin/main` gets the fork's main — which can also be stale. The project config already supports qualified refs (`defaultBranch: "upstream/main"` → used verbatim in `baseRefCandidates`), but there's no fetch logic to match. The fetch should extract the remote from the configured `defaultBranch`: if it contains `/`, split on first segment as the remote name; otherwise default to `origin`.

5. **Architecture boundary.** The "should we refresh before spawning" policy decision belongs in the session manager (before `workspace.Create()`). The gitworktree adapter should expose a `FetchDefaultBranch(ctx, repoPath, remote, branch) error` method — pure git operation, no policy.

### Proposed flow

```
spawn
  → session manager: decide to refresh
    → detect remote from DefaultBranch ("upstream/main" → remote=upstream, else origin)
    → gitworktree.FetchDefaultBranch(ctx, repo, remote, branch)
      → if error: log warning, continue (non-blocking)
      → if success: origin/main is now fresh
  → workspace.Create (unchanged)
    → addWorktree → resolveBaseRef → finds fresh origin/main
```

## Affected Code

| File | Location | Role |
|------|----------|------|
| `backend/internal/adapters/workspace/gitworktree/workspace.go` | `addWorktree()` :760 | Where fetch should be inserted (or called from session manager) |
| `backend/internal/adapters/workspace/gitworktree/workspace.go` | `resolveBaseRef()` :1015 | Reads local refs only |
| `backend/internal/adapters/workspace/gitworktree/commands.go` | `baseRefCandidates()` :131 | Builds candidate list (all local refs) |
| `backend/internal/session_manager/manager.go` | `createSessionWorkspace()` :714 | Policy decision: trigger fetch before Create |
| `backend/internal/domain/projectconfig.go` | `DefaultBranch` :21 | Config source — no canonical-remote concept today |

## Existing Tests to Update

- `manager_test.go:920` — `TestSpawnRecordsRemoteTrackingDiffBaseWhenLocalDefaultBranchLags` explicitly tests the current stale-ref behavior. Will need updated expectations after the fix.

## Related Issues

- #3810 — workspace spawn forces root repo's default branch on child repos (different problem, same area)
- #3857 — harden workspace compare-base selection for divergent histories (related: diff base after the fact)

nikhilachale:
## Summary
- refresh the configured default branch with a narrow best-effort fetch before creating worker worktrees
- preserve fork workflows by treating configured refs like upstream/main as qualified remotes only when that remote exists
- keep offline/auth/slow-network cases non-blocking with a short timeout and warning log

## Validation
- go test ./internal/adapters/workspace/... ./internal/session_manager ./internal/integration
- go test ./internal/... was attempted; unrelated timing/deadline failures occurred in agent/runtime adapter tests

Closes #3860


backend/internal/session_manager/manager.go
```diff
@@ -704,6 +705,7 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 	if branch == "" {
 		branch = DefaultSpawnBranch(id, cfg.Kind, sessionPrefix(project), projectKind, m.dataDir)
 	}
+	m.refreshDefaultBranchBestEffort(ctx, project)
```
whoisasx: [P1] This refresh only covers `project.Path`. For a workspace project, `CreateWorkspaceProject` subsequently creates worktrees for registered child repositories with independent remotes and default branches, but none of those repositories are fetched. Their worktrees can therefore still start from stale tracking refs. Please refresh the root and every child repository best-effort before materialization, with an integration test that advances a child remote and verifies its worktree `HEAD` and persisted `BaseSHA`.

backend/internal/session_manager/manager.go
```diff
@@ -889,6 +891,47 @@ func (m *Manager) loadProject(ctx context.Context, projectID domain.ProjectID) (
...
+			m.logger.Warn("spawn: default branch remote probe failed; falling back to origin",
+				"remote", candidateRemote,
+				"branch", defaultBranch,
+				"error", err,
+			)
+		} else if exists {
+			return candidateRemote, candidateBranch
+		}
+	}
+	return "origin", defaultBranch
```
whoisasx: [P1] When `release/2026` is a branch name rather than a remote-qualified ref, this fetches `origin release/2026`, which refreshes `origin/release/2026`. Worktree creation later treats every slash-containing default as qualified and resolves bare `release/2026`, so it can select a stale local branch even though the fetch succeeded. Please carry the selected qualified base into creation or align base-ref resolution with the ref actually refreshed, and cover this with a fetch-to-create integration test.

backend/internal/adapters/workspace/gitworktree/commands.go
```diff
@@ -10,6 +10,14 @@ func revParseVerifyArgs(repo, ref string) []string {
 	return []string{"-C", repo, "rev-parse", "--verify", "--quiet", ref}
 }
 
+func remoteGetURLArgs(repo, remote string) []string {
+	return []string{"-C", repo, "remote", "get-url", remote}
+}
+
+func fetchBranchArgs(repo, remote, branch string) []string {
+	return []string{"-C", repo, "fetch", remote, branch}
```
whoisasx: [P2] Fetching a source branch without an explicit destination relies on `remote.<name>.fetch`. If a custom or single-branch refspec excludes this branch, Git still succeeds and updates only `FETCH_HEAD`; the remote-tracking ref used by worktree creation remains stale. Please fetch into a validated explicit destination such as `+refs/heads/<branch>:refs/remotes/<remote>/<branch>`, or create from the fetched commit directly, and test with a restrictive refspec.

whoisasx (changes requested):
The standard `origin/main` path is implemented cleanly and all exact-head CI checks pass. Focused package, race, and Windows compilation checks also pass. Targeted Git reproductions found three cases where fetch reports success or runs as intended but worktree creation still uses a stale ref. The two P1 cases leave supported workspace and slash-named-branch flows exposed to the original bug, so please address them before merging.

nikhilachale:
@whoisasx  

> Addressed the review findings in `39b6e6e32`.
>
> - Refreshes the root and every registered workspace child repository before materialization.
> - Fetches with an explicit refspec so restrictive remote configurations cannot leave tracking refs stale.
> - Resolves slash-named branches such as `release/2026` against the refreshed `origin/<branch>` ref.
> - Added regression coverage for child repository refresh and slash-named branches.
>
> Focused package tests and race tests pass.

backend/internal/session_manager/manager.go
```diff
@@ -889,6 +891,65 @@ func (m *Manager) loadProject(ctx context.Context, projectID domain.ProjectID) (
...
+	}
+	repos, err := m.store.ListWorkspaceRepos(ctx, project.ID)
+	if err != nil {
+		m.logger.Warn("spawn: workspace child repo lookup failed; continuing with local refs", "projectID", project.ID, "error", err)
+		return
+	}
+	for _, repo := range repos {
+		branch := repo.DefaultBranch
+		if strings.TrimSpace(branch) == "" {
+			branch = project.Config.WithDefaults().DefaultBranch
```
whoisasx: [P1] An empty child `DefaultBranch` means something different at this point than it does during materialization. Migration `0086_workspace_repo_default_branch.sql` leaves existing workspace child rows empty, and `CreateWorkspaceProject` intentionally interprets that as 'infer this child's own `origin/HEAD`'. Here it is replaced with the parent project's default instead. For example, a migrated child whose default is `dev` under a parent using `main` fetches `main`, then creation still infers and uses stale `origin/dev`. Please infer the child's actual default before fetching or backfill/persist it, and cover the migrated empty-value path with the requested fetch-to-create integration test that verifies `HEAD` and `BaseSHA`.

backend/internal/adapters/workspace/gitworktree/commands.go
```diff
@@ -130,14 +138,12 @@ func ignoredCountArgs(worktree string) []string {
...
 func baseRefCandidates(branch, defaultBranch string) []string {
 	candidates := []string{"origin/" + branch}
+	// Remote-tracking refs are preferred even when the branch name itself
+	// contains a slash (for example release/2026). A qualified remote branch
+	// such as upstream/main remains available as the next candidate.
 	if strings.Contains(defaultBranch, "/") {
-		// A qualified default ("upstream/main") is used verbatim; git's refname
-		// disambiguation already falls back to refs/heads/<defaultBranch>.
-		candidates = append(candidates, defaultBranch)
+		candidates = append(candidates, "origin/"+defaultBranch, defaultBranch)
```
whoisasx: [P1] This loses the remote decision made by `defaultBranchFetchTarget`. When the configured default is `upstream/main`, AO fetches `refs/remotes/upstream/main`, but creation probes `origin/upstream/main` first. Both refs can legally exist; in a real Git reproduction with different commits, the fetch succeeded but the new worktree used `origin/upstream/main` instead of the configured `upstream/main`. Please carry the resolved canonical base ref into creation rather than reconstructing it with an ambiguous slash heuristic, and add this collision case to the integration coverage.

backend/internal/session_manager/manager.go
```diff
@@ -889,6 +891,65 @@ func (m *Manager) loadProject(ctx context.Context, projectID domain.ProjectID) (
...
+func (m *Manager) refreshDefaultBranchBestEffort(ctx context.Context, project domain.ProjectRecord) {
+	if project.Kind.WithDefault() == domain.ProjectKindScratch {
+		return
+	}
+	if strings.TrimSpace(project.Path) == "" {
+		return
+	}
+	refresh := func(repoPath, defaultBranch string) {
+		remote, branch := m.defaultBranchFetchTarget(ctx, repoPath, defaultBranch)
+		fetchCtx, cancel := context.WithTimeout(ctx, defaultBranchFetchTimeout)
```
whoisasx: [P2] The five-second timeout is recreated for each repository and the workspace loop invokes `refresh` sequentially. The total spawn delay is therefore `(children + 1) * 5s`; an offline workspace with 20 child repos can block for roughly 105 seconds despite the PR describing slow-network handling as non-blocking. Please apply one overall refresh budget, or run bounded concurrent fetches under a single deadline.

whoisasx (changes requested):
Thanks for addressing the earlier feedback. The explicit destination refspec and ordinary slash-branch path now look correct, and configured workspace child branches are fetched before materialization. However, two stale-base paths still violate the PR's core guarantee: migrated child rows can refresh the wrong branch, and a qualified remote can be shadowed during base selection. The per-repository timeout can also make an offline workspace spawn block for minutes. Focused and race suites pass, and exact-head CI is green, but the qualified-remote issue was reproduced with a real Git integration scenario. Please address these findings before merging.

backend/internal/session_manager/manager.go
whoisasx: Fixed in [`8d7ed99d5`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8d7ed99d556da4217782a0238a4ef7a3ca9c17ad) with the persistence assertion in [`65304b45d`](https://github.com/Untrivial-ai/agent-orchestrator/commit/65304b45d). Spawn now builds refresh targets for the root and every registered child, resolves all canonical bases, refreshes them before `CreateWorkspaceProject`, and forwards each `BaseRef` into materialization. `TestWorkspaceIntegrationRefreshesInferredChildBaseBeforeMaterialization` advances a child remote and asserts the child worktree `HEAD` and materializer `BaseSHA` equal the fresh commit; `TestSpawn_InfersEmptyWorkspaceChildDefaultBeforeFetchAndCreate` also asserts that the session manager persists the returned child `BaseSHA`.

backend/internal/session_manager/manager.go
whoisasx: Fixed in [`8d7ed99d5`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8d7ed99d556da4217782a0238a4ef7a3ca9c17ad). Empty child defaults are no longer replaced with the parent default. Each child is passed unchanged to `ResolveDefaultBranch`, where the git adapter infers that repository's own `origin/HEAD` before any network I/O. The resolved canonical ref is then fetched and carried into creation. The real-Git integration advances an inferred `dev` child and verifies fresh `HEAD`/`BaseSHA`; the manager test covers the migrated empty value and durable BaseSHA write.

backend/internal/adapters/workspace/gitworktree/commands.go
whoisasx: Fixed in [`8d7ed99d5`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8d7ed99d556da4217782a0238a4ef7a3ca9c17ad). Default resolution now returns a canonical `BaseRef` such as `refs/remotes/upstream/main`, and that exact ref is carried through `WorkspaceConfig` / `WorkspaceProjectRepoConfig` into base selection. Creation no longer reconstructs the refreshed target from the ambiguous slash heuristic. `TestWorkspaceIntegrationCanonicalBaseRefAvoidsRemoteNameCollision` creates divergent `origin/upstream/main` and `upstream/main` refs and proves the worktree starts from the configured upstream commit.

backend/internal/session_manager/manager.go
whoisasx: Fixed in [`8d7ed99d5`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8d7ed99d556da4217782a0238a4ef7a3ca9c17ad). All local canonical-ref resolution happens first, then one `context.WithTimeout` wraps the complete root-plus-children fetch loop. A slow first repository therefore consumes the shared budget and later fetches observe the same expired context instead of receiving new five-second windows. `TestRefreshDefaultBranchesUsesOneOverallFetchBudget` blocks six fetches and verifies they complete within one 25ms test budget.

whoisasx:
## Before / after behavior

This PR changes session spawning from “materialize from whatever default-branch ref is already available locally” to “best-effort refresh the canonical default branch first, then materialize from that exact ref.”

### Before

```mermaid
flowchart TD
    B1["Spawn session"] --> B2["Create root and child worktrees immediately"]
    B2 --> B3["Select a locally available default-branch ref"]
    B3 --> B4{"Is the local remote-tracking ref stale?"}
    B4 -- "Yes" --> B5["Worker starts from an older commit"]
    B4 -- "No" --> B6["Worker starts from the expected commit"]
```

- No fetch occurred on the spawn path.
- A stale `origin/main` (or child-repository default branch) could therefore seed a stale worker worktree.
- Qualified/default branches such as `upstream/main` and slash-containing branches such as `release/2026` did not have one canonical ref carried through resolution and materialization.

### After

```mermaid
flowchart TD
    A1["Spawn session"] --> A2["Collect root repo and materializable workspace children"]
    A2 --> A3["Resolve each canonical remote + branch + BaseRef locally"]
    A3 --> A4["Start one shared 5-second refresh budget"]
    A4 --> A5["Fetch each explicit default-branch refspec"]
    A5 --> A6{"Fetch succeeded?"}
    A6 -- "Yes" --> A7["Create worktree from the refreshed canonical BaseRef"]
    A6 -- "No or timeout" --> A8["Warn and continue from the resolved local BaseRef"]
    A7 --> A9["Worker starts from the freshest available default-branch commit"]
    A8 --> A9
```

- Root and ready child repositories are resolved before network I/O, so one slow repository cannot prevent canonical-ref discovery for the others.
- The entire workspace uses one shared timeout rather than multiplying spawn latency per child.
- Refresh remains best-effort: offline or unavailable remotes do not block spawning.
- Workspace children marked `needs_init` are skipped during both refresh and materialization.

whoisasx (approved):

PR #3864, opened 2026-08-12
feat(frontend): refine session inspector rail

whoisasx:
## Summary

- move the session inspector header into the top row with Summary, Browser, and Files destinations
- keep the notification action beside the far-right inspector toggle while the rail is expanded
- use a stable 360px default/minimum width, cap expansion at 50%, and persist user resizing in pixels
- fully collapse the rail and resize handle to zero width with coordinated open/close motion
- smoothly collapse destination labels to icons at compact widths
- start each session visit on Summary while preserving later browser-activity reveal behavior
- keep session-route navigation and topbar hosting aligned so the inspector does not appear blank or below the main header

## Scope exclusions

This PR intentionally does not change notification icon/badge appearance, left-sidebar peek timing, workspace topbar action styling, terminal controls, terminal tabs, drag/drop, or pinning.

## Verification

- `npm --prefix frontend test -- SessionInspector SessionView ShellTopbar shell-new-session-shortcut` — 175 tests passed

## Test infrastructure note

Vitest 4 currently loads the Jest-DOM convenience entry against a different matcher instance in this checkout. The small explicit matcher registration in `test/setup.ts` is required for the existing DOM assertions to run.


whoisasx:
## Before
Screenshot 2026-08-12 at 10 31 07

--- 

## After 
Screenshot 2026-08-12 at 08 27 25
Screenshot 2026-08-12 at 08 27 09

frontend/src/renderer/styles.css
```diff
@@ -2458,15 +2458,73 @@ body.is-resizing-x [data-slot="sidebar-container"] {
...
-/* Do not animate the inspector's flex-grow. The sibling is a live xterm WebGL
- * surface: interpolating this value makes ResizeObserver fit the terminal at
- * every intermediate width, reallocating the renderer and eventually causing
- * a PTY/SIGWINCH repaint. The inspector still opens and closes atomically, and
- * direct separator drags remain interactive. */
-.session-split #inspector[data-panel] {
+/* Match the left sidebar's spring-like reveal while keeping direct separator
+ * drags under the pointer. Both panels participate so neither edge snaps. */
+.session-split > [data-panel] {
+	transition: flex-grow 240ms cubic-bezier(0.16, 1, 0.3, 1);
```
Annieeeee11: **[P2] Do not reintroduce flex-grow animation on the live terminal split**

This reverts the explicit guard from the terminal resize stabilization work (`23eb763a6`). Animating `flex-grow` on **both** panels continuously retargets the xterm host for 240ms on every open/close.

Debounced `fit()` reduces “fit every frame” damage, but this still undoes a known WebGL/SIGWINCH mitigation with no coverage that toggling the rail leaves the terminal grid/PTY stable.

Prefer animating only opacity/transform on the inspector chrome (already present) and keep panel `flex-grow` atomic — or add a terminal resize assertion proving intermediate fits no longer fire.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -629,75 +676,100 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
-							aria-hidden={!isInspectorOpen}
+							aria-hidden={!inspectorPanelVisible}
+							className="session-inspector-panel"
 							collapsible={!isInspectorOpen}
 							defaultSize={inspectorDefaultSize}
 							id="inspector"
 							inert={!isInspectorOpen}
 							maxSize={`${INSPECTOR_MAX_PERCENT}%`}
-							minSize={`${INSPECTOR_MIN_PERCENT}%`}
+							minSize={INSPECTOR_MIN_SIZE}
```
Annieeeee11: **[P2] Fixed 360px min and 50% max can invert on narrow session splits**

`INSPECTOR_MIN_PX = 360` with `INSPECTOR_MAX_PERCENT = 50` means whenever the session split is under ~720px (common with an expanded sidebar on ≤900–1100px windows), `minSize` exceeds `maxSize`.

Previously both bounds were percentages, so min < max always held. Under inversion, rrp constraint handling and `resize("360px")` / collapse become unreliable.

Clamp min to `min(360px, maxPercentInPx)`, or keep a percentage floor.

frontend/src/renderer/styles.css
```diff
@@ -2458,15 +2458,73 @@ body.is-resizing-x [data-slot="sidebar-container"] {
...
+	opacity: 1;
+	transform: translateX(0);
+	transition:
+		max-width 170ms cubic-bezier(0.2, 0.8, 0.2, 1),
+		margin-left 170ms cubic-bezier(0.2, 0.8, 0.2, 1),
+		opacity 120ms ease-out,
+		transform 170ms cubic-bezier(0.2, 0.8, 0.2, 1);
+}
+
+@container inspector (max-width: 359px) {
```
Annieeeee11, line 2745: **[P3] Icon-only tab collapse is unreachable while the rail is open**

Labels hide at `@container inspector (max-width: 359px)`, but the open panel cannot go below 360px, so the “compact widths → icons” behavior never appears in steady state — only mid-animation while content is clipped.

Either lower `minSize`, raise the container breakpoint, or drop the claim/CSS.

Annieeeee11 (changes requested):
## Review summary

Direction matches the stated UX (shared header row, pixel width persistence, collapse-to-zero, Summary-first visits). CI is green.

Blocking concerns are the reintroduced terminal `flex-grow` animation and the fixed 360px min vs 50% max constraint inversion on narrow splits. Compact label collapse is unreachable while the rail is open (P3).

frontend/src/renderer/styles.css
whoisasx: Fixed in b1a8674c2. Removed the flex-grow transition from the session split and restored atomic panel sizing, so the live xterm host is no longer refit through intermediate animation frames. The inspector content keeps its opacity/translate chrome transition.

frontend/src/renderer/components/SessionView.tsx
whoisasx: Fixed in b1a8674c2. The inspector now observes the usable split width and clamps its minimum to min(360px, 50% of that width), with the separator width reserved. Restored/persisted widths are also clamped to the current 50% maximum when opening. Added a narrow-split regression test.

frontend/src/renderer/styles.css
whoisasx: Fixed in b1a8674c2. Because the responsive minimum can now fall below 360px on narrow splits, the existing 359px container breakpoint is reachable and the tabs can settle into their icon-only state.

neversettle17-101 (approved):

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -706,10 +819,12 @@ export function SessionView({ sessionId }: SessionViewProps) {
 											<SessionFilesView onToggleMaximized={handleToggleFilesPopOut} sessionId={session.id} />
 										) : null
 									}
-									isInspectorVisible={isInspectorOpen}
+									isInspectorVisible={inspectorPanelVisible}
+									notificationAction={inspectorPanelVisible ? <NotificationCenter /> : undefined}
```
Annieeeee11: Non-blocking P3: the close animation can mount two NotificationCenters. `isInspectorOpen` flips false immediately on close, so the embedded `ShellTopbar` mounts its `NotificationCenter` right away, while `notificationAction={inspectorPanelVisible ? <NotificationCenter /> : undefined}` keeps rendering one inside the inspector until the collapse finishes (~180ms). You get two bells during that window.

Gate both on the same owner — either drive the topbar bell off `inspectorPanelVisible` (motion-visible) too, or drop the inspector's `notificationAction` as soon as `isInspectorOpen` goes false — so the bell doesn't duplicate while the panel is still collapsing.

frontend/src/renderer/components/ShellTopbar.tsx
```diff
@@ -327,27 +327,26 @@ export function ShellTopbar({ embedded = false }: { embedded?: boolean } = {}) {
...
-									<PanelRightOpen className="size-5" aria-hidden="true" />
-								)}
-							</TopbarButton>
-						)}
 					</>
 				) : null}
-				{/* The bell always trails the actions row, on every platform. */}
-				<NotificationCenter style={noDragStyle} />
+				{/* The expanded inspector owns the notification action beside its close toggle. */}
+				{!(isSessionRoute && !isOrchestrator && isInspectorOpen) ? (
```
Annieeeee11: Non-blocking P3 (same issue as the SessionView anchor): this gate reads `isInspectorOpen`, which flips false the instant close begins, so the topbar bell mounts immediately. Meanwhile the inspector's `notificationAction` is still gated on `inspectorPanelVisible` (motion-visible) and keeps its own bell for ~180ms until collapse finishes — two bells.

Gate both bells on the same owner (store-open vs motion-visible) so the bell doesn't duplicate during the close animation.

Annieeeee11 (commented):
Non-blocking P3 — close animation can mount two NotificationCenters for ~180ms. Not blocking merge; leaving for a follow-up. Details on the two anchors below.

frontend/src/renderer/components/ShellTopbar.tsx
whoisasx: Fixed in 7819af1f5. Both render paths now transfer notification ownership from the same immediate `isInspectorOpen` state: the center topbar takes ownership as soon as close begins, and the animated inspector retains only its visual shell. The added regression assertion verifies the inspector action is removed immediately on close.

frontend/src/renderer/components/SessionView.tsx
whoisasx: Fixed in 7819af1f5. The inspector notification action now follows the immediate `isInspectorOpen` ownership state, while `inspectorPanelVisible` remains presentation-only for the closing animation. This prevents the closing inspector and center topbar from mounting two notification centers during the transition. A focused SessionView regression assertion covers the handoff.

Annieeeee11:

Annieeeee11 (approved):
lgtm

Annieeeee11:
@codebanditssss  please take a look before merge

PR #3871, opened 2026-08-12
feat(frontend): move terminal fullscreen to context menu

whoisasx:
## Summary
- remove the terminal fullscreen control from the top bar
- expose Fullscreen terminal / Exit fullscreen in the terminal right-click menu
- keep the menu inside the fullscreen element so exiting fullscreen remains available

## Verification
- npm test -- src/renderer/components/XtermTerminal.test.tsx -t 'toggles terminal fullscreen from the right-click menu'
- npm test -- src/renderer/components/CenterPane.test.tsx -t 'hides session-level actions while the terminal is fullscreen'
- npm run typecheck

## Scope
This PR changes only terminal fullscreen entry/exit behavior. Font-size controls and shortcuts are intentionally unchanged.

### DIFF
--- a/frontend/src/renderer/components/CenterPane.test.tsx
+++ b/frontend/src/renderer/components/CenterPane.test.tsx
@@ -445,7 +445,7 @@ describe("CenterPane toolbar session label", () => {
 		act(() => document.dispatchEvent(new Event("fullscreenchange")));
 
 		expect(screen.queryByTestId("session-action-region")).not.toBeInTheDocument();
-		expect(screen.getByRole("button", { name: "Exit fullscreen" })).toBeInTheDocument();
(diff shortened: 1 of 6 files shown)


whoisasx:
### Before
Screenshot 2026-08-12 at 09 36 12

---

### After
Screenshot 2026-08-12 at 10 14 58

neversettle17-101:
Can we also add a keyboard shortcut that will help to go fullscreen?

whoisasx:
> Can we also add a keyboard shortcut that will help to go fullscreen?

should I do in the same PR only?

neversettle17-101:
> > Can we also add a keyboard shortcut that will help to go fullscreen?
> 
> should I do in the same PR only?

we can take it up in a new PR

neversettle17-101 (approved):

PR #3872, opened 2026-08-12
feat(frontend): scope zoom shortcuts to terminal focus

whoisasx:
## Summary
- remove terminal font-size controls from the top bar
- resize the focused terminal with Command +/- on macOS and Ctrl +/- on Windows/Linux
- consume the shortcut only inside xterm so application zoom remains unchanged elsewhere

## Verification
- npm test -- src/renderer/components/XtermTerminal.test.tsx -t 'uses (Command|Ctrl) plus/minus for terminal font size'
- npm test -- src/renderer/components/CenterPane.test.tsx -t 'keeps terminal font-size controls out of the top bar'
- npm run typecheck

## Scope
This PR changes only terminal font-size controls and focused keyboard routing. Terminal fullscreen behavior is intentionally unchanged.


whoisasx:
## Before 
Screenshot 2026-08-12 at 09 36 12

---
## After
Screenshot 2026-08-12 at 10 15 48

Annieeeee11 (changes requested):
## Findings

**[P1] Block Electron zoom accelerators when the terminal owns Cmd/Ctrl +/- — `frontend/src/renderer/components/XtermTerminal.tsx`**

The new handler only `preventDefault`/`stopPropagation`s inside xterm. That does not stop Electron application-menu zoom roles, which still fire outside the renderer event path.

This app installs those accelerators on every platform:
- Windows: hidden app menu with `Ctrl+=` / `Ctrl+Plus` / `Ctrl+-` `zoomIn`/`zoomOut` roles (`frontend/src/main/menu.ts`)
- macOS/Linux: no `setApplicationMenu` override, so Electron’s default View menu keeps the same zoom roles

So with the terminal focused, Cmd/Ctrl +/- can both bump terminal font size and apply page zoom. That breaks the PR’s main claim (“application zoom remains unchanged”). Fix this in main (e.g. `before-input-event` when terminal focus is known): `preventDefault` the zoom chord and route font-size changes there, or otherwise suppress the menu role while the terminal is focused. Renderer-only consumption is not enough.

**[P2] Update e2e coverage that still drives the removed top-bar controls — `frontend/e2e/terminal-viewport-retention.spec.ts:267-270`**

That spec still reads the `NNpx` label and clicks `Increase terminal font size`. Both were removed from `CenterPane`. It isn’t in the `@T0|@P0` smoke grep, so CI can stay green while full `test:e2e` fails. Point it at the keyboard path (or another stable font-size trigger) before merge.

## Assessment

Direction is good: move font sizing onto focused terminal shortcuts, keep wheel zoom, thread `onChangeFontSize` through `TerminalPane`/`XtermTerminal`, and the unit tests cover Mac vs Linux modifiers. Vitest + renderer-smoke are green.

Do not merge until the Electron menu-accelerator conflict is fixed; otherwise focused zoom will dual-fire on real desktop. Residual gaps: `ShellTerminalsView` still uses a fixed `fontSize={12}` with no `onChangeFontSize` (shortcut won’t apply there), and the shortcuts UI/docs don’t mention the new binding.

neversettle17-101:
Let us get it tested on windows and linux machine as well. @codebanditssss @somewherelostt / @Vaibhaav-Tiwari

whoisasx:
Addressed both blocking findings in `6549d53e`:

- **Electron zoom accelerator arbitration:** moved ownership of focused terminal zoom to the main process. `before-input-event` now checks terminal focus, prevents the Electron menu zoom role from firing, and sends a terminal font-size delta over IPC. Outside a focused terminal, the event is left untouched so normal application zoom continues to work.
- **Full E2E coverage:** updated `terminal-viewport-retention.spec.ts` to focus xterm and drive `ControlOrMeta+=` instead of reading/clicking the removed top-bar controls.

Coverage includes macOS and Windows/Linux modifier behavior, focused/unfocused routing, and preservation of application zoom outside the terminal. The current `test`, `renderer-smoke`, and `scan` checks are green.

The standalone `ShellTerminalsView` font sizing and shortcuts UI/docs observations remain outside this PR’s session-terminal scope; this PR does not claim to change those surfaces.

Please re-review when convenient.

Annieeeee11:

Annieeeee11 (approved):
lgtm

PR #3874, opened 2026-08-12
feat(frontend): refine central topbars

whoisasx:
## Summary

- standardize the Board, Orchestrator, and Session top bars at 36px with compact 28px controls
- add responsive label priorities so secondary button text collapses smoothly before primary actions
- simplify route identity and activity presentation while preserving existing terminal, inspector, and notification behavior

## Scope boundaries

- terminal fullscreen and font-sizing controls remain in their current locations
- terminal cards and terminal creation behavior are unchanged
- no sidebar, inspector content, notification logic, or backend changes

## Verification

- `npm --prefix frontend test -- src/renderer/components/SessionTopbarPortal.test.tsx src/renderer/components/ShellTopbar.test.tsx src/renderer/components/SessionsBoard.test.tsx -t 'renders session chrome|shows the worker session name|shows project identity|shows the Board identity'`
- `git diff --check`


whoisasx:
## Before
Screenshot 2026-08-12 at 10 29 37
Screenshot 2026-08-12 at 10 29 47
Screenshot 2026-08-12 at 10 31 07
---
## After
Screenshot 2026-08-12 at 11 47 24
Screenshot 2026-08-12 at 11 47 31
Screenshot 2026-08-12 at 11 47 40

Annieeeee11 (changes requested):
## Review

Direction looks good (shared 36px chrome, compact controls, priority label collapse), but the `sessionAction` regroup introduces real behavior regressions that contradict the PR’s scope claims. Blocking on the fullscreen control loss.

### Findings

**[P1] New terminal and Switch agent disappear in terminal fullscreen**
- `frontend/src/renderer/components/SessionView.tsx` (sessionLocalActions)
- `frontend/src/renderer/components/CenterPane.tsx` (session-action-region hidden while fullscreen)

Both controls moved out of the terminal strip / agent tab into `sessionLocalActions` → `topbarActions` / `session-action-region`, which is unmounted in terminal fullscreen. Previously they stayed available in the fullscreen topbar. Keyboard new-terminal still works; switch-agent has no fallback. Please keep these in the measured terminal region (or otherwise available in fullscreen).

**[P2] Orchestrator sessions gain a New terminal control**
- `frontend/src/renderer/components/SessionView.tsx`

`sessionLocalActions` mounts Plus for any `session`. The old CenterPane gate was `!isOrchestratorSession(session)`, and the test asserting no new-terminal on orchestrators was deleted. `TerminalSwitchAgentButton` still self-hides for non-workers; new-terminal does not. Restore the gate (or explicitly product-approve the change).

**[P2] macOS in-panel titlebar no longer matches traffic-light clearance**
- `frontend/src/renderer/styles.css` (`.center-panel-shell--mac .center-panel-titlebar`)
- `frontend/src/styles/tokens.css` (`--size-topbar-primary` 40→36)

Titlebar height now uses `--size-toolbar` (36px) while sidebar traffic-light offset stays `--size-traffic-light-clearance` (40px), so the board/session strip is 4px shorter than adjacent chrome.

**[P3] Icon topbar controls can end up non-square**
- `TopbarButton` icon variants still use `size-control-lg` (34×34); `.workspace-topbar-actions .topbar-control` only overrides height/min-width to 28px.

**[P3] New-terminal tooltip lost the shortcut hint**
- Relocated control only shows `shortcut.new-shell-terminal`; old CenterPane tooltip used `terminal.newWithShortcut` (e.g. ⌘T).

### Test gaps
CenterPane tests were updated to assert the new placement rather than catch the fullscreen / orchestrator regressions above. Worth adding coverage for: new-terminal + switch-agent still present in terminal fullscreen; no new-terminal on orchestrator sessions.

whoisasx:
Addressed all findings in `c23d5bd9d`:

- **Fullscreen controls:** the session action region now remains mounted in terminal fullscreen, so New Terminal, Switch Agent, and the interface-switch control remain available alongside Exit Fullscreen.
- **Orchestrator New Terminal:** restored the worker-only gate; orchestrator sessions no longer render New Terminal.
- **macOS chrome alignment:** kept the intentional 36px design and made `--size-traffic-light-clearance` follow the shared toolbar token, removing the 40px/36px mismatch.
- **Square icon controls:** icon and kill-icon variants now use the 28px control token for both dimensions instead of retaining the old 34px width.
- **Shortcut tooltip:** restored the localized shortcut-aware label, including ⌘T on macOS and Ctrl+T elsewhere.

Added regression coverage for fullscreen action availability and the orchestrator terminal gate. Verification: frontend typecheck passed, the focused suite passed 96/96, and the current `test`, `renderer-smoke`, and `scan` checks are green.

Please re-review when convenient.

Annieeeee11:
## Review — follow-up to the changes requested

All five prior findings are resolved in `c23d5bd` with regression coverage added. Independent pass on the final diff below; everything here is P3 / informational.

### Prior findings — all addressed ✅

- **[P1] Fullscreen control loss** — `CenterPane.tsx` dropped the `isFullscreen ? null :` guard around `session-action-region`; New Terminal / Switch Agent / interface-switch stay mounted in terminal fullscreen. Test renamed and asserts the action region contains the action button.
- **[P2] Orchestrator New Terminal** — `SessionView.tsx` gates the Plus button with `!isOrchestrator`; new test asserts absence on orchestrator sessions.
- **[P2] macOS chrome mismatch** — `--size-traffic-light-clearance` now follows `--size-toolbar`, and `--size-topbar-primary` is 36px, so titlebar and sidebar clearance agree.
- **[P3] Non-square icon controls** — `icon`/`killIcon` variants moved from `size-control-lg` (34px) to `size-control-md` (28px), matching the `.workspace-topbar-actions .topbar-control` override.
- **[P3] Shortcut tooltip** — `SessionView.tsx` restores `t("terminal.newWithShortcut", { shortcut })` as `title`; test asserts `New terminal (Ctrl+T)`.

### New observations on the final diff

**[N1, P3] Worker session topbar no longer shows the git branch**
`ShellTopbar.tsx` (~L1001-1013) replaces the worker-session identity: the old `GitBranch` + `session.branch` block is gone, replaced by `session.title`. The `GitBranch` import was removed and no test pins branch visibility anymore. The PR summary says "simplify route identity" but doesn't explicitly call out dropping the branch. If users relied on seeing the branch in the session topbar, this is a quiet regression — confirm the branch is still discoverable elsewhere (inspector) or restore it as a secondary element.

**[N2, P3] `SessionInterfaceSwitchButton` is now icon-only**
`SessionInterfaceSwitch.tsx` replaced the labeled "Switch to terminal/chat UI" button with an icon-only `TopbarButton variant="icon"` (MessageSquare / SquareTerminal) + tooltip. More compact, but the visible affordance is gone. Accessibility is preserved via tooltip + `aria-label`; just flagging the discoverability tradeoff.

**[N3, P3] Collapsed labels stay in the DOM**
The responsive collapse uses `max-width: 0; opacity: 0; overflow: hidden` on `[data-compact-label]` rather than `display: none`. Text remains in the a11y tree, but each button has an explicit `aria-label` (which wins over inner text for the accessible name), so no double-read. Fine — noting it's a visual-only hide.

**[N4, P3] Orchestrator button label vs. tooltip mismatch**
The Orchestrator button's visible compact label is always `t("shell.orchestrator")`, while its tooltip varies (`openOrchestrator` / `spawnOrchestrator` / `spawning` / `restarting`). So while spawning, the chip reads "Orchestrator" but the tooltip says "Spawning…". The old code surfaced the dynamic state in the label itself. Minor — worth a product check since the dynamic state is now tooltip-only.

**[N5, P3] `TerminalSwitchAgentButton` still rendered for orchestrator sessions**
`SessionView.tsx` L565 mounts `<TerminalSwitchAgentButton session={session} />` for every session including orchestrators. This matches prior behavior (the component self-hides via `canSwitchAgentHarness`), and the prior review only flagged New Terminal for orchestrators — so it's consistent. Just noting it relies on the child's self-gating; an explicit `!isOrchestrator` gate here would be more robust.

### Test gap
No test asserts the **git branch is gone** from the worker topbar (N1). If the branch removal is intentional, a test pinning the new `session.title` identity behavior would prevent accidental reintroduction.

### Verdict
Ship-ready from my side. The only thing I'd want explicit confirmation on before merge is **N1** (dropping the branch display) — if that's an intentional product decision, ship it; if not, restore the branch as a secondary identity element.

Annieeeee11:

Annieeeee11 (commented):
Independent pass on the final diff — two new informational notes (P3, non-blocking) beyond what @Annieeeee11 already raised. Everything else from the prior rounds looks resolved.

**[P3] Responsive label collapse is macOS-only for board actions**

`usesBoardActionsInPanel()` returns `true` only on macOS (`lib/platform.ts`). On Win/Linux the Board New-task / Orchestrator buttons live in the full `ShellTopbar` (`.dashboard-app-header`), which is **not** inside `.center-panel-surface` / `.workspace-topbar-container`. The `@container workspace-topbar (max-width: …)` rules therefore only fire for the in-panel board chrome (macOS) and the embedded session actions — the Win/Linux shell-header board buttons keep their full labels at every width. The PR summary frames the collapse as a general feature, so this platform asymmetry is worth either documenting or covering with a second container on the shell header.

**[P3] Two ancestors share `container-name: workspace-topbar`**

`center-panel-surface` and `.workspace-topbar-container` (`SessionTopbarHost`) both declare `container-name: workspace-topbar`. Container queries resolve to the nearest named ancestor, so embedded session actions (portaled into `SessionTopbarHost`) query the host, while board actions query `center-panel-surface`. Both are center-pane-width today so the breakpoints coincide, but the dual declaration is fragile — if the host ever gets a different width (e.g. a split pane), the two action groups will collapse at different points. Consider naming only the host, or giving the surface a distinct name.

Neither blocks merge from my side. The one item I'd still want explicit confirmation on before merge is the already-raised **N1** (git branch dropped from the worker topbar) — if that's an intentional product decision, a test pinning the new `session.title` identity would prevent accidental reintroduction.

whoisasx:
@Annieeeee11 Addressed the responsive-container follow-up in `90b7a748c`.

- **Windows/Linux label collapse:** on the current branch, `usesFramedAppTopbar()` already places the full `ShellTopbar` inside `CenterPanelShell`, so it inherited the parent container query. That made the behavior dependent on route structure, though. The full `ShellTopbar` now owns `workspace-topbar-container` directly, so its Task/Orchestrator labels collapse against the actual topbar width on Windows/Linux regardless of its parent.
- **macOS in-panel board:** the in-panel `SessionsBoard` titlebar now owns the same responsive container explicitly.
- **Duplicate named ancestors:** `.center-panel-surface` now exposes only `center-panel`; it no longer also declares `workspace-topbar`. `SessionTopbarHost` remains the session topbar’s explicit responsive boundary, so embedded session actions no longer resolve through competing same-named ancestors.
- **N1 / branch removal:** this is an intentional product decision. Worker identity in the topbar is `session.title` plus activity; the branch remains available in terminal metadata and through branch actions elsewhere. Existing coverage in `ShellTopbar.test.tsx` already asserts that the worker title is present and `ao/sess-1` is absent.

Validation:
- focused topbar/board/portal suite: 73/73 passed
- frontend typecheck: passed
- `git diff --check`: passed

Annieeeee11 (approved):
lgtm

Annieeeee11:
@codebanditssss  please take a look before merge

Issue #3875, opened 2026-08-12
bug(orchestrator): fresh orchestrator spawn crashes claude with "invalid session id", leaving an unrecoverable exited/non-terminated session


axisrow:
## Summary

Spawning a new **orchestrator** session (`ao spawn --kind orchestrator ...`) for a specific project intermittently fails immediately: the tmux pane shows `invalid session id` and the underlying `claude` process never starts. This has recurred across several days on the same project, and there is no recovery path — `restore` refuses the resulting session, and killing + respawning reproduces the exact same failure.

## Steps to reproduce

1. Have an existing orchestrator session for a project (harness `claude-code`, mode `tui`).
2. At some point the session's activity flips to `exited` while `terminated` stays `false` (observed after a daemon/tmux restart, e.g. Mac sleep — but also reproduced via a **fresh** spawn, see below).
3. `ao session get <id>` shows:
   ```
   status: exited
   activity: exited
   terminated: false
   ```
4. `ao session restore <id>` refuses:
   ```
   Session is not restorable (SESSION_NOT_RESTORABLE)
   ```
5. `ao session kill <id>` succeeds and flips `terminated: true`.
6. `ao spawn --project <p> --kind orchestrator --branch <same-branch> --name orchestrator --mode tui` creates a **new** session id, but the tmux pane immediately shows:
   ```
   invalid session id
   ```
   and stays there — no `claude` process is running in the pane (`tmux list-panes -F "#{pane_current_command}"` shows a bare `cat`, not `claude`). The corresponding `ao agent-process supervise ... -- claude ...` wrapper process is also gone (it exited alongside the crashed `claude` child), leaving a dead pane with nothing supervising it.
7. The new session is left in the same broken state as step 3 (`status/activity: exited`, `terminated: false`) — no further recovery via CLI.

This reproduced on a fresh `ao spawn` (not just `restore`), which rules out a single stale/orphaned resume-id as the sole cause — something about how the `--session-id`/`--resume` value is generated or accepted for this project's orchestrator role is rejected by the `claude` binary outright.

For contrast, doing the identical kill+respawn dance on a **different** project's orchestrator (same `claude-code` harness, same `claude` binary version) recovered cleanly — the new session started normally with a live `claude` process in the pane. So this is not a global daemon/tmux issue; it is scoped to specific project/session state.

## Observed

```
$ ao session get <new-session-id>
status: exited
activity: exited
terminated: false
```

Pane content:
```
invalid session id
```

[...]

PR #3876, opened 2026-08-12
fix(cli): allow dots in sessionIDPattern to match projectIDPattern

axisrow:
## Problem

Any project whose id contains a dot (e.g. a project named like a domain, `myproject.example`) fails to spawn or restore **any** session, permanently, regardless of `num`. The agent process never starts — the tmux pane is left running a bare `cat`, and the session lands in an `exited` / `is_terminated=false` limbo that neither `restore` nor a full daemon restart can recover.

## Root cause

Two regexes that validate the same AO session id (`{ProjectID}-{num}`) are out of sync:

- `projectIDPattern` (`backend/internal/service/project/service.go`) allows dots: `^[A-Za-z0-9][A-Za-z0-9._-]*$`
- `sessionIDPattern` (`backend/internal/cli/hooks.go`) does not: `^[A-Za-z0-9_-]+$`

`session_manager.Manager` passes the session id raw into the supervisor wrapper:

```go
wrapped = append(wrapped, executable, "agent-process", "supervise", "--session", string(id), "--launch", launchID, "--")
```

`agent-process supervise` then rejects it before the agent binary is even started:

```go
if !sessionIDPattern.MatchString(sessionID) {
    return usageError{fmt.Errorf("invalid session id")}
}
```

So a dotted project id passes project-level validation but fails supervisor validation — the agent process is never launched.

Confirmed empirically against the built daemon binary:
```
--session "myproject.example-16"  → invalid session id
--session "myproject-3"           → (child runs fine)
```

Full repro and investigation: #3875

## Fix

Widen `sessionIDPattern` to match `projectIDPattern`:

```go
var sessionIDPattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._-]*$`)
```

A leading dot is still rejected, so `..` / `../` traversal segments remain impossible — the id stays safe to embed both in a URL path segment (`sessions/<id>/activity`) and in a filename.

## Testing

- `TestSessionIDPattern` (`hooks_test.go`) — unit-level accept/reject matrix, including dotted ids and traversal payloads.
- `TestAgentProcessSuperviseAcceptsDottedProjectSessionID` (`agent_process_unix_test.go`) — end-to-end: `agent-process supervise --session "my.project-16"` actually runs the child process and reports exit.
- `TestAgentProcessSuperviseRejectsTraversalSessionID` — `../evil`, `a/b`, `.hidden` still rejected.
- `gofmt -l .` clean on the whole backend module, `go build ./...`, `go vet ./...`, `go test ./...` (137 packages) all green.
- Manually verified end-to-end on a real dotted-name project: killed the stuck session, respawned with the patched daemon+CLI, new session came up `idle` with a live TUI (previously: `invalid session id` on every spawn attempt).

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/backend/internal/adapters/runtime/conpty/runtime.go
+++ b/backend/internal/adapters/runtime/conpty/runtime.go
@@ -46,8 +46,13 @@ func conptyPartialCreateFailure(err error, handle ports.RuntimeHandle, cleanup p
 	return runtimeEffectFailure{err: err, handle: handle, effect: ports.RuntimeEffectPossible, cleanup: cleanup}
 }
 
-// validSessionID matches agent-orchestrator's assertValidSessionId.
-var validSessionID = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+// validSessionID matches the CLI's sessionIDPattern (internal/cli/hooks.go):
+// the daemon issues session ids derived from project names, so dots are legal
+// and must not be rejected here. The leading [a-zA-Z0-9] anchor keeps a bare
+// "." / ".." impossible. The id is used only as a map key, the host process's
+// argv[0], and the registry identity — never as a path or pipe component — so
+// the widened alphabet is safe.
+var validSessionID = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`)
 
 // hostSession is the in-memory state for a live pty-host connection.
 type hostSession struct {
@@ -123,7 +128,7 @@ func New(opts Options) *Runtime {
 func (r *Runtime) Create(ctx context.Context, cfg ports.RuntimeConfig) (ports.RuntimeHandle, error) {
 	id := string(cfg.SessionID)
 	if !validSessionID.MatchString(id) {
-		return ports.RuntimeHandle{}, conptyCreateFailure(fmt.Errorf("conpty: invalid session id %q: must match ^[a-zA-Z0-9_-]+$", id))
+		return ports.RuntimeHandle{}, conptyCreateFailure(fmt.Errorf("conpty: invalid session id %q: must match ^[a-zA-Z0-9][a-zA-Z0-9._-]*$", id))
 	}
 	if cfg.WorkspacePath == "" {
 		return ports.RuntimeHandle{}, conptyCreateFailure(fmt.Errorf("conpty: workspace path required"))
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

yrevah1:
Independent reproduction on Linux, arrived at from the symptom side before finding this PR — the diagnosis here is correct.

A project whose id contained dots had four consecutive spawns (one orchestrator, three workers) die on arrival, each landing in `activity=exited` / `is_terminated=false`. Confirmed against the built binary that the supervisor is where it stops:

```
$ ao agent-process supervise --session <dotted-project>-4 --launch abc123 -- /bin/true
invalid session id

$ ao agent-process supervise --session scratch-1 --launch abc123 -- /bin/true
(child runs)
```

After applying exactly the widening proposed here, the same project spawns and the agent is alive as a child of the supervisor:

```
<pid>  ao agent-process supervise --session <dotted-project>-2 --launch a095433e-…
 └─ <pid>  claude --session-id dceb40a6-…
```

Two notes that may help this land:

**The anchor is doing real security work, and it holds.** `^[A-Za-z0-9]` makes a bare `.` or `..` impossible, so the widened pattern still cannot yield a traversal segment in the loopback URL the comment above `sessionIDPattern` is guarding, nor a name that escapes a directory when the id becomes a path component. Worth stating explicitly in the PR body, since that comment is the stated reason the pattern exists. `a..b` remains matchable but is harmless in both roles, and `validateProjectID` already rejects `..` at the project level so it cannot arise from a real project id.

**There is at least one more copy of this pattern.** Anything else that bounds a session id before it becomes a path component needs the same widening, or a dotted project spawns locally and then fails at that boundary instead. Worth grepping for the literal `^[A-Za-z0-9_-]+$` across the backend rather than fixing only `cli/hooks.go`.

Related: #4479 is the same class one layer down — conpty hard-rejects dotted ids that tmux sanitises — so a shared normalise/validate layer for session ids would cover this PR, #3875 and #4479 together.

i-trytoohard:
@axisrow Marking as stale: no activity from you for 27 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

axisrow:

axisrow:
@illegalcall friendly bump

PR #3877, opened 2026-08-12
feat(frontend): add draggable terminal strip

whoisasx:
## Summary

- keep the owning session terminal fixed at the far left while auxiliary native terminals scroll independently
- keep auxiliary terminal cards at a fixed width and allow drag-and-drop reordering
- use native horizontal wheel/trackpad scrolling without left/right arrow controls
- add trailing terminal-region spacing, a right inset for the close action, and the compact 11px agent icon

## Scope

- no terminal pinning state or pin controls
- no session-terminal picker or additional terminal types
- existing terminal font-size and fullscreen controls remain unchanged

## Checks

- `npm --prefix frontend test -- src/renderer/components/CenterPane.test.tsx src/renderer/components/ShellTerminalTab.test.tsx -t 'keeps the compact owner card|reserves trailing space|keeps fixed-width auxiliary tabs|does not add arrow controls|reorders only auxiliary terminals|uses a compact fixed width|scrolls the tab strip horizontally with the mouse wheel|uses roving keyboard focus'`


whoisasx:
Implementation is ready for review.

This PR keeps the owning session terminal fixed at the far left, gives auxiliary native terminals fixed-width horizontal overflow, and limits drag-and-drop reordering to those auxiliary terminals. It also removes overflow arrow controls in favor of native wheel/trackpad scrolling, adds the thin scrollbar and trailing control spacing, moves the active marker to the top edge, and keeps the agent icon at 11px.

Scope remains intentionally limited: no pinning, no session-terminal picker, and no changes to terminal font-size or fullscreen behavior.

Verification covers owner-card placement, fixed auxiliary widths, drag ordering, native horizontal scrolling, absence of arrow controls, close-action spacing, keyboard roving focus, and header alignment. Current `test`, `renderer-smoke`, and `scan` checks are green.

Please review when convenient.

Annieeeee11 (commented):
## Review — draggable terminal strip

Overall this is well-scoped with thorough test coverage, and the fixed-owner + draggable-auxiliary + native-scroll design is sound. A couple of items with runtime impact worth addressing before merge, plus some cleanups.

### Should address before merge

**1. Close-button drag guard depends on an attribute that isn't shown in the diff**
`DraggableShellTerminal.startDrag` bails when `event.target.closest("[data-terminal-tab-action],input,a")` matches. This only protects the close/rename affordances if `ShellTerminalTab` actually tags its close button with `data-terminal-tab-action`. The diff doesn't show that attribute being added. If it isn't, a pointer-down on the close button starts a drag instead of registering a click. Please confirm the close button (and rename input) carry `data-terminal-tab-action`, and add it if not.

**2. `useOverflowScroll` looks half-dead now**
After this PR only `tabsOverflow.ref` is consumed — `canScrollLeft`, `canScrollRight`, and `scrollByDirection` are no longer used. Two concerns:
- If the hook attaches its own wheel listener that calls `scrollBy`/`preventDefault`, it may fight the native horizontal wheel scrolling this PR intends to enable. Confirm the hook no longer intercepts wheel, or slim it to just a ref.
- Drop the unused overflow-detection surface so it doesn't imply chevron behavior that no longer exists.

### Worth a follow-up

**3. `scrollIntoView` can scroll ancestors**
The new effect calls `activeTab?.scrollIntoView?.({ behavior: "smooth", block: "nearest", inline: "nearest" })`. `block: "nearest"` is meaningless for a horizontal strip and `scrollIntoView` can vertically scroll any scrollable ancestor. Safer to compute and set `scrollLeft` directly on `tabsOverflow.ref.current`.

**4. `shellOrderBySession` never prunes**
The per-session handle map accumulates for the renderer's lifetime; stale handles from closed terminals/sessions are filtered at render (harmless) but never removed. Consider pruning handles that no longer exist in `shellTerminals` inside `reorderShellTerminals`, or dropping the entry on session unmount.

**5. Reviewer terminal is auxiliary but non-draggable**
The reviewer terminal sits inside the scroll region but outside `Reorder.Group`, so shell tabs reorder around it. Defensible, but the summary says "auxiliary native terminals" get DnD — the reviewer doesn't. Either tighten the summary to "auxiliary *shell* terminals" or fold the reviewer into the group.

### Nits

- Several assertions migrated from `toHaveClass`/`toHaveFocus`/`toBeInTheDocument` to `classList.contains(...) === true` / `document.activeElement === ...` / `toBeNull()`. The vitest matchers give better failure output; prefer keeping them unless there was a concrete jsdom quirk.
- Scrollbar `height` went `3px` (commit 2) → `2px` (commit 3); squashing would keep history clean.
- `--size-terminal-agent-icon: var(--size-terminal-agent-icon)` in `styles.css` is a self-referential passthrough that only resolves because the token is defined in `tokens.css` — a comment or non-self-referential alias would clarify intent.

### Looks right
- `orderedShellTerminals` memo reapplies preferred order and appends unknown/new terminals stably.
- `reorderShellTerminals` dedup + append-missing is correct for the small N.
- `selectAdjacentTab` correctly switched to `orderedShellTerminals` with updated deps.
- `touch-pan-y` on draggable items is the right trackpad coexistence call.
- Cross-browser scrollbar styling (Firefox `scrollbar-width: thin` + webkit pseudo-elements) is complete.
- The `motion/react` mock capturing `onReorder` is a clean way to assert reorder in jsdom.

Recommend addressing #1 and #2 before merge; the rest can be a fast-follow.

Annieeeee11:

whoisasx:
@Annieeeee11 Addressed the runtime items and follow-ups in the latest update:

1. **Drag guard:** confirmed `ShellTerminalTab`’s close button carries `data-terminal-tab-action`; rename uses an `<input>`, which is covered by the existing `[data-terminal-tab-action], input, a` guard.
2. **Overflow handling:** `CenterPane` no longer consumes the full `useOverflowScroll` hook. It now has a scoped non-passive wheel adapter only for vertical-wheel-to-horizontal movement; horizontal trackpad input remains native. The shared hook remains unchanged for `ShellTerminalsView`, where its overflow state is still consumed by chevrons.
3. **Active-tab reveal:** replaced `scrollIntoView` with strip-local bounds calculation and `scrollTo({ left })`, so selecting a terminal cannot move ancestor containers.
4. **Order state:** replaced the accumulating per-session map with current-session order state. Closed handles are pruned, and the order resets when leaving the session.
5. **Reviewer terminal:** the reviewer now participates in the same `Reorder.Group` as shell terminals. The owning session remains fixed at the far left.

I also aligned the reviewer card with the connected terminal-card dimensions and reduced its avatar to the same 11px token used by the owning session.

The branch is now merged with current `main` and the conflict resolution preserves main’s terminal-toolbar removals.

Validation completed:
- `npm test -- --run src/renderer/components/CenterPane.test.tsx src/renderer/components/ShellTerminalTab.test.tsx --reporter=dot` — 47/47 passed
- `npm run typecheck` — passed
- `git diff --check origin/main...HEAD` — passed

Linux/Windows platform validation is still pending; I’m not claiming it as completed here.

Annieeeee11 (approved):
lgtm

Annieeeee11:
@codebanditssss please take a look

PR #3882, opened 2026-08-12
feat: add native Android emulator (Track A)

Vaibhaav-Tiwari:
## What

Adds a native, in-app Android emulator: a Settings-gated "Emulator" tab (default off) that boots a real, AO-managed Android device (Google's own SDK/emulator binary, not a from-scratch emulator) directly inside a session, with a live view, tap/swipe/keyboard input, and an agent-facing `ao android emulator` CLI surface (`screenshot`, `inspect-ui`, `tap`/`swipe`/`type`/`press-key`, `find-source`) so the agent can see and drive the device too.

This is **Track A (Android) only** — cross-platform on **Windows, macOS, and Linux**. iOS Simulator support (Track B, macOS-only) is a separate, not-yet-built follow-up.

## Why

Lets an agent build and verify a mobile app end-to-end without leaving AO — install the SDK, boot a device, see the screen, tap through the app, and locate the source behind what's on screen, the same way a developer would with Android Studio's embedded emulator.

## How

- **SDK acquisition** (`androidsdk`): downloads/verifies `platform-tools`, `emulator`, and one pinned system image directly from Google's public repository manifests — no JVM/`sdkmanager` dependency.
- **Device lifecycle** (`androidemulator`): one persistent, shared, headless emulator process with quick-boot snapshots and crash/restart-with-backoff, driven over the loopback `EmulatorController` gRPC service the SDK's own emulator binary exposes (insecure/no-auth by default, confirmed during an earlier spike — no WebRTC in the standard SDK, so the transport is `getScreenshot` polling at ~5fps, not smooth video).
- **Renderer**: new "Emulator" tab in the session inspector (same pattern as Browser/Files), gated behind a Settings toggle so it costs nothing when unused.
- **Agent-facing CLI** (`ao android emulator ...`, surfaced to agents via the `using-ao` skill): screenshot, `uiautomator`-based UI tree inspection, tap/swipe/type/key input, and heuristic identifier→source-file lookup.
- **Env integration**: `ANDROID_HOME`/`ANDROID_SDK_ROOT`/`PATH` are injected into session and standalone terminals once the SDK is installed, and `ao doctor` reports SDK/virtualization status.
- **Process supervision fix**: `Kill()` now terminates the whole process tree (`taskkill /T /F` on Windows, process-group kill on Unix), not just the tracked PID — the emulator binary is a launcher whose actual VM backend runs as a separate child process, and a single-PID kill left it orphaned, holding the AVD's lock files and breaking every subsequent boot.

**Known constraint, not a bug**: the emulator needs real host headroom (~3-4GB RAM, a few CPU cores) once booted. On a memory-constrained host it can boot but then show in-guest ANRs ("System UI isn't responding") under contention — expected on any Android emulator (Google's own guidance is 16GB+ RAM for a comfortable experience), not specific to this integration.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l .` all clean.
- `go test ./...` passes for every android-related package (`androidsdk`, `androidemulator`, `daemon`, `cli`, `session_manager`'s new env-injection tests); a handful of pre-existing, unrelated flakes on this Windows dev box (nvm-version PATH test, a workspace-watcher timing test) are unaffected by this branch — verified they fail identically on `origin/main`.
- `npm run typecheck` and `npx vitest run` (121 tests across `EmulatorPanel`, `useAndroidDevice`, `android-frame-stream`, `android-stream`, `ui-store`, `GlobalSettingsForm`, `SessionInspector`) all pass.
- Manually exercised end-to-end on Windows: settings toggle → SDK install → device boot → live interactive Pixel home screen → pop-out → tap/swipe/type.

## Testing this PR (Windows / Linux)

1. Check out this branch, `npm install` in `frontend/`.
2. Run the desktop app in dev mode.
3. Settings → toggle on the Android Emulator.
4. Open a session → new "Emulator" tab → Install Android SDK (~2GB download) → Start.
5. On Linux, hardware acceleration needs KVM (`/dev/kvm` readable by your user) — `ao doctor` reports acceleration status if it's unavailable.

## Checklist

- [x] Branched from `main`
- [x] One focused change (Android emulator only — no unrelated files)
- [x] Relevant CI checks pass (go build/vet/test, frontend typecheck/test)


i-trytoohard:
From the NotesBot call on `2026-08-12`, this existing pr appears to be relevant.
Match confidence: `0.80` (fuzzy).

Relevant discussion excerpts:
- [0:12] !adil: Uh, and what else? [0:16] prateek: And also I'll be checking into the PR that, uh, Webow has made for the Android emulator thing. PR Webow has made for the Android emulator? Okay. Sounds good. [0:30] -Prasad-: Hello, my name is— like, I reported an issue today and it matched with the issues which [...]
- [0:16] prateek: And also I'll be checking into the PR that, uh, Webow has made for the Android emulator thing. PR Webow has made for the Android emulator? Okay. Sounds good. [0:30] -Prasad-: Hello, my name is— like, I reported an issue today and it matched with the issues which the participants were also raising. [...]

Source thread: https://discord.com/channels/1476302178913357958/1537149882463297597/1537149882463297597

Prasad-D-Ware:
Tested this on macOS (Apple Silicon, M-series) at `b3d69ca0`. The SDK/detection path worked nicely — it found my existing Android Studio SDK and wrote `external-sdk.json` with no download at all — but **the device never boots on arm64**. Root cause below, plus a diagnostics issue that made it much harder to find than it should have been.

## 1. Blocker: `hw.cpu.arch` is written as an ABI, not a CPU arch

Clicking **Start** parks on `Emulator crashed / emulator did not become ready in time` after ~3 minutes. The emulator actually dies in under a second. Running AO's exact spawn command by hand:

```
INFO  | Android emulator version 36.6.11.0 (build_id 15507667)
INFO  | Graphics backend: gfxstream
FATAL | CPU Architecture 'arm64-v8a' is not supported by the QEMU2 emulator, (the classic engine is deprecated!)
```

`WriteAVDConfig` passes `profile.ABI` into two config.ini fields that take **different vocabularies** — `avd.go:97`:

```go
`...
abi.type=%s          // line 63  -> wants an ABI:      arm64-v8a, x86_64
...
hw.cpu.arch=%s       // line 70  -> wants a CPU arch:  arm64, x86_64, arm, x86
...`, avdName, profile.ABI, profile.ABI, ...
```

Generated `config.ini` on this machine:

```ini
abi.type=arm64-v8a
hw.cpu.arch=arm64-v8a   # invalid — must be arm64
```

**Why this is green on Windows/Linux and in CI:** on x86_64 the ABI token and the arch token are the same string (`x86_64`), so passing the ABI where the arch belongs is accidentally correct. The two only diverge on ARM. Every existing test asserts the x86_64 path, so no unit test can catch it either — and since `DetectPlatform` maps `darwin/arm64` to `arm64-v8a` (`platform.go:57`), macOS Apple Silicon is the only configuration that hits it. That's the one platform the manual pass didn't cover.

### Fix

Map ABI → CPU arch rather than reusing the ABI. Something like:

```go
// cpuArchForABI maps a system-image ABI to the value config.ini's
// hw.cpu.arch expects. These are different vocabularies that happen to
// coincide on x86_64, which is why passing the ABI works there and fails
// on arm64-v8a ("CPU Architecture 'arm64-v8a' is not supported").
func cpuArchForABI(abi string) string {
	switch abi {
	case "arm64-v8a":
		return "arm64"
	case "armeabi-v7a", "armeabi":
		return "arm"
	default: // x86, x86_64
		return abi
	}
}
```

and use it for the `hw.cpu.arch` argument only, leaving `abi.type` as `profile.ABI`.

**Verified:** patching just that one line to `hw.cpu.arch=arm64` in the generated config gets straight past the FATAL into a normal boot — `Found systemPath .../android-35/google_apis_playstore/arm64-v8a/`, then `hasSufficientDiskSpace: Ok`, `hasSufficientHwGpu: Ok`, etc. It's the only thing blocking Apple Silicon.

A regression test would want to assert `hw.cpu.arch=arm64` for an `arm64-v8a` profile — the current `avd_test.go` cases would all still pass with the bug in place.

## 2. Boot failures are reported as timeouts, and the emulator's own error is dropped

Independent of the above, and worth fixing regardless: the emulator printed the exact cause and exited in well under a second, but the UI showed a generic `did not become ready in time` three minutes later.

`bootOnce` (`manager.go:145`) spawns the process and then calls `waitReady`, which only polls `adb shell getprop sys.boot_completed` every 2s until `ReadyTimeout` (3 min). It never checks whether the process is still alive, and the emulator's stdout/stderr is never captured or surfaced. So *any* fast, fully self-diagnosed startup failure becomes a 3-minute wait ending in an unactionable message.

Two suggestions:

- In `waitReady`, also select on process exit — if the process is gone, fail immediately instead of polling a corpse for the remaining timeout.
- Capture the emulator's output and put the last few lines into `lastErr` / the panel's error text. `FATAL | CPU Architecture 'arm64-v8a' is not supported` would have made this self-service.

## 3. Minor notes

- `hw.ramSize=2048` is hardcoded, while the PR text says the emulator wants ~3–4GB; the emulator itself bumps it (`Increasing RAM size to 2560MB`). Might be worth setting deliberately rather than being silently overridden.
- The PR description advertises agent tools named `emulator_screenshot`, `emulator_tap`, `emulator_find_source_for_element`, etc. Those names don't exist in the code — agent access is the `ao android emulator ...` CLI surfaced via the `using-ao` skill. Probably just a stale description, but it sent me looking for tools that aren't there.
- `PlayStore.enabled=false` is written even when the detected image is `google_apis_playstore`. Didn't cause a failure here, just flagging.

Happy to push the `cpuArchForABI` fix + test if that's useful, or leave it to you.

## Repro

macOS on Apple Silicon, with an existing Android Studio SDK:

```bash
ao android emulator start   # or the panel's Start button
cat ~/.ao/tools/android/avd/ao-default.avd/config.ini | grep hw.cpu.arch
# hw.cpu.arch=arm64-v8a

env ANDROID_AVD_HOME=~/.ao/tools/android/avd \
    ANDROID_SDK_ROOT=~/Library/Android/sdk \
    ANDROID_PREFS_ROOT=~/.ao/tools/android/prefs \
  ~/Library/Android/sdk/emulator/emulator \
  -avd ao-default -no-window -no-audio -no-boot-anim -grpc 8554
# FATAL | CPU Architecture 'arm64-v8a' is not supported by the QEMU2 emulator
```

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -164,14 +179,18 @@ export function SessionInspector({
...
+	const emulatorEnabled = useUiStore((state) => state.emulatorEnabled);
 	const filesChangedCount = useSessionWorkspaceFilesChangedCount(session?.id);
 	const setView = (next: InspectorView) => {
 		setInternalView(next);
 		onViewChange?.(next);
 		if (next === "files") onOpenFiles?.();
 	};
 	const view: InspectorView = requestedView;
-	const tabs = VIEW_DEFS.map((entry) => {
+	const tabs = VIEW_DEFS.filter((entry) => entry.id !== "emulator" || emulatorEnabled).map((entry) => {
```
codebanditssss, line 193: when this setting is turned off while the current inspector view is `emulator`, this only removes the tab from `tabs`; the controlled `view` can remain `"emulator"` and `EmulatorTabView` still receives `isActive`. That leaves the panel rendered (and SDK/emulator hooks active) with no selected tab. Please reset the inspector view/stop the emulator when disabling the setting, or gate `isActive` with `emulatorEnabled`.

codebanditssss:
tested this pr: backend `go test ./...` and `go vet ./...` pass; frontend typecheck passes; 139 focused emulator/settings/session tests pass; and all github ci checks are green. i left one inline correctness comment: disabling the emulator setting while its tab is active can leave the emulator view/hooks active after the tab is hidden. also note the pr currently has merge conflicts with main.

codebanditssss:
tested the android emulator flow on linux with the sdk installed and kvm available. `ao-default` consistently segfaults during startup before adb becomes ready (`emulator-5554` stays offline). the ui only reports “emulator did not become ready in time,” which hides the actual crash. please pin/update the emulator version or provide a compatible fallback.

backend/internal/daemon/androidsdk_wiring.go
```diff
@@ -0,0 +1,411 @@
...
+		Args: []string{
+			"-avd", androidAVDName,
+			"-no-window", "-no-audio", "-no-boot-anim",
+			"-grpc", fmt.Sprintf("%d", androidGRPCPort),
+		},
+		Env: env,
+		AccelCheck: func(ctx context.Context) (androidemulator.AccelStatus, error) {
+			return androidemulator.CheckAcceleration(ctx, plan.EmulatorPath)
+		},
+		ReadyCheck:        func(ctx context.Context) (bool, error) { return adbBootCompleted(ctx, plan.AdbPath) },
```
codebanditssss, line 277: if the emulator process exits before `sys.boot_completed`, this readiness loop keeps polling until the full timeout. please watch the process exit signal and return the actual emulator error immediately, otherwise users get a misleading timeout and stale avd locks can remain.

frontend/src/renderer/components/SessionInspector.tsx
Vaibhaav-Tiwari: Fixed — the emulator branch is now gated on `emulatorEnabled` directly (`session && emulatorEnabled ? <EmulatorTabView .../> : undefined`), and a new effect resets `view` back to `"summary"` if the setting is turned off while the emulator tab is active, so the panel/hooks stop rendering immediately instead of lingering with no selected tab. Added a regression test ("stops rendering the emulator panel (and its polling hooks) once the setting is switched off while its tab is active") that reproduces exactly this sequence.

backend/internal/daemon/androidsdk_wiring.go
Vaibhaav-Tiwari: Fixed — `waitReady` (backend/internal/androidemulator/manager.go) now races `ReadyCheck` polling against the process's own exit: it spawns a goroutine that closes a channel on `proc.Wait()` and selects on it both between poll attempts and during the poll-interval sleep. If the process dies first, `bootOnce` fails immediately with `emulator exited before becoming ready: <last non-empty log line>` instead of waiting out the full `ReadyTimeout`. Also wired `ClearStaleLocks` into `StartDevice` right after `EnsureSnapshotValid` so `hardware-qemu.ini.lock`/`multiinstance.lock` left behind by a crashed attempt don't block the next boot. Covered by a new test that boots a synthetic "print-fatal-and-exit" process and asserts the failure is detected well under the configured timeout, with the process's own FATAL line surfaced in the error.

Vaibhaav-Tiwari:
@Prasad-D-Ware Thank you for the extremely thorough repro — this was invaluable. All three points addressed:

1. **`hw.cpu.arch` blocker**: fixed exactly as you diagnosed, with your `cpuArchForABI` mapping (credited). `abi.type` still gets `profile.ABI`; `hw.cpu.arch` now gets the mapped CPU-arch token. Added `TestCpuArchForABI` (table-driven over arm64-v8a/armeabi-v7a/armeabi/x86_64/x86) and `TestWriteAVDConfigWritesCPUArchNotABIForArm64`, which fail against the old code and pass now.
2. **Timeout hiding the real error**: `waitReady` now races the readiness poll against the process's own exit (via a `proc.Wait()` goroutine + `select`), so a fast crash is detected immediately instead of polling for the full 3-minute timeout, and the process's own last log line is surfaced in the error instead of a generic "did not become ready in time". Covered by a new test that boots a synthetic process which prints a FATAL line and exits immediately, asserting the failure surfaces well under the configured timeout.
3. **Minor notes**: `hw.ramSize` is now `2560` (matching what the emulator bumps it to anyway), and `PlayStore.enabled` is now conditional on the profile's tag being `google_apis_playstore` instead of hardcoded `false`.

On the PR description mentioning non-existent tool names (`emulator_screenshot`, etc.) — you're right, that's stale copy referring to the actual `ao android emulator ...` CLI surface. I'll get the description corrected separately; flagging here so it isn't lost.

Vaibhaav-Tiwari:
@codebanditssss Addressing both of your notes:

**Emulator tab staying active after disabling the setting**: fixed — replied inline on the specific line, but summarizing here too. The emulator branch is now gated directly on `emulatorEnabled`, and a new effect resets the inspector view back to `"summary"` if the setting is switched off while the emulator tab is active, so the panel and its polling hooks stop immediately instead of lingering with no selected tab. Covered by a new regression test.

**Merge conflicts with main**: resolved — the branch is now merged up to date with `main` and CI is running on the latest commit.

**Linux + KVM segfault**: I want to be upfront that this is only partially addressed. The *symptom* you hit — "emulator did not become ready in time" hiding the actual crash — is fixed by the same `waitReady` rework mentioned above: a process that dies before becoming ready now surfaces its own last log line immediately instead of a generic timeout, so if you re-run this you should get real diagnostic output (e.g. a segfault signal or crash log line) instead of the opaque timeout message. However, I have **not** investigated or fixed the underlying segfault itself — I don't have a Linux+KVM environment to reproduce it in, and pinning/updating the emulator version needs someone who can verify the fix against the actual crash. If you're able to re-run with the updated code, the surfaced error output would help scope that follow-up (likely worth its own issue rather than blocking this PR, since the crash appears specific to that host configuration).

Vaibhaav-Tiwari:
Found and fixed one more bug during a final pass: the toolbar's **Home** and **Back** buttons were silently no-ops. Root cause — they sent literal `"Home"`/`"Back"` as the gRPC `KeyboardEvent.key` value, but the emulator's own protobuf docs specify its Android-specific phone keys as `"GoHome"`/`"GoBack"` (the Overview button, already correctly `"AppSwitch"`, was unaffected). Neither `"Home"` nor `"Back"` is a valid W3C `KeyboardEvent.key` value either, so the emulator just dropped them.

Fixed the two call sites, added regression tests asserting the correct key values are sent, and corrected the doc comments/CLI help text (`ao android emulator press-key`) that held the same wrong example values, so agents driving the device via the CLI don't hit the same mistake.

LaibaFirdouse:
## Ubuntu Linux testing - findings

Tested the Android Emulator flow on Ubuntu Linux with KVM available/readable.

### What works:

Android Emulator setting can be enabled.
Emulator tab appears correctly.
Android SDK setup completes.
Emulator eventually boots and the live Android view is displayed.
Stop/start works; after retrying, the emulator can successfully boot.
/android-device/status consistently returned 200.
/android-device/input requests were also returning 200.

### Issues observed:

### Inconsistent startup/live-view initialization

On every Start/reload, I consistently see this sequence:

Booting emulator... → Connecting to live view... → broken image (Android emulator screen) / flickering → back to loading → Connecting to live view... → emulator eventually renders.

So the emulator does eventually become available, but the live-view initialization appears to go through one or more failed/retry states before stabilizing.

### Initial startup failure

On the first Start attempt I also saw:

Emulator crashed

Failed to process .ini file
~/.ao/tools/android-preifs/emu-update-last-check.ini

Clicking Start again allowed the emulator to boot successfully.

### Severe responsiveness issues after boot

The emulator is very laggy on my Ubuntu machine. Interactions can take several seconds to respond, and I encountered Android's "System UI isn't responding" dialog. Eventually, clicks stopped producing any visible interaction even though the AO /android-device/input requests were still returning HTTP 200.

### Potential rendering-related signal

The Electron logs also contain repeated Chromium GetVSyncParametersIfAvailable() failed errors. I can't confirm that these are the cause, but they may be relevant given the live-view flickering and responsiveness issues.

### Overall: 
The emulator can successfully boot and AO's Android API layer appears to be communicating with it, but on Ubuntu I consistently see live-view initialization/retry behavior and significant Android UI responsiveness/input issues. 
The startup/live-view sequence and the eventual loss of interaction seem worth investigating.


Screenshot from 2026-08-19 12-44-34
Screenshot from 2026-08-19 12-52-17
Screenshot from 2026-08-19 12-53-58

Vaibhaav-Tiwari:
> ## Ubuntu Linux testing - findings
> Tested the Android Emulator flow on Ubuntu Linux with KVM available/readable.
> 
> ### What works:
> Android Emulator setting can be enabled. Emulator tab appears correctly. Android SDK setup completes. Emulator eventually boots and the live Android view is displayed. Stop/start works; after retrying, the emulator can successfully boot. /android-device/status consistently returned 200. /android-device/input requests were also returning 200.
> 
> ### Issues observed:
> ### Inconsistent startup/live-view initialization
> On every Start/reload, I consistently see this sequence:
> 
> Booting emulator... → Connecting to live view... → broken image (Android emulator screen) / flickering → back to loading → Connecting to live view... → emulator eventually renders.
> 
> So the emulator does eventually become available, but the live-view initialization appears to go through one or more failed/retry states before stabilizing.
> 
> ### Initial startup failure
> On the first Start attempt I also saw:
> 
> Emulator crashed
> 
> Failed to process .ini file ~/.ao/tools/android-preifs/emu-update-last-check.ini
> 
> Clicking Start again allowed the emulator to boot successfully.
> 
> ### Severe responsiveness issues after boot
> The emulator is very laggy on my Ubuntu machine. Interactions can take several seconds to respond, and I encountered Android's "System UI isn't responding" dialog. Eventually, clicks stopped producing any visible interaction even though the AO /android-device/input requests were still returning HTTP 200.
> 
> ### Potential rendering-related signal
> The Electron logs also contain repeated Chromium GetVSyncParametersIfAvailable() failed errors. I can't confirm that these are the cause, but they may be relevant given the live-view flickering and responsiveness issues.
> 
> ### Overall:
> The emulator can successfully boot and AO's Android API layer appears to be communicating with it, but on Ubuntu I consistently see live-view initialization/retry behavior and significant Android UI responsiveness/input issues. The startup/live-view sequence and the eventual loss of interaction seem worth investigating.
> 
> Screenshot from 2026-08-19 12-44-34 Screenshot from 2026-08-19 12-52-17 Screenshot from 2026-08-19 12-53-58

lets connect over a call, please send a message on discord

i-trytoohard:
@Vaibhaav-Tiwari Marking as stale: this has been open 28 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

i-trytoohard:
@Vaibhaav-Tiwari Marking as stale: no activity for 25 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Vaibhaav-Tiwari Marking as stale: no activity for 25 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #3889, opened 2026-08-12
feat: Change notification icon badge background to red

nikhilachale:
Changed the notification icon badge background color to red to make unread notifications more prominent.

**Changes:**
- Updated NotificationCenter.tsx to use bg-error class for the notification count badge instead of bg-foreground

**Files modified:**
- frontend/src/renderer/components/NotificationCenter.tsx

This is a simple UI change that makes the notification badge more visible by using a red background color instead of the default foreground color.


nikhilachale:
testing  purposes

PR #3890, opened 2026-08-12
fix(chat): keep nested Codex turns from draining queued work

### LINKED ISSUE #3878: bug(chat): nested Codex turns can drain queued automation before the root completes, leaving a permanent phantom Working turn
## Bug summary

AO can permanently show **Working** after the provider is idle and all real work has finished.

This is not merely a stale animation and it is not a long-running tool call. In the captured incident, a nested Codex child turn completed while the root turn was still active. AO treated that child completion as if the whole conversation had become idle, drained a queued automation request too early, and created a second durable AO turn. Codex accepted the automation message into the still-running root transcript rather than starting the provider turn AO had assigned. The assigned AO turn therefore received no provider lifecycle events, remained `running` forever in SQLite, and kept the frontend spinner alive.

**Confidence:** High  
**Priority:** P1  
**Reported/reproduced:** 2026-08-12, from a live packaged AO desktop session  
**Packaged version:** `0.12.1-nightly.202608101434` (bundled daemon reports `dev`)  
**Code inspected:** current `origin/main` at `f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9`  
**Codex CLI:** `0.147.0`  
**Environment:** macOS Darwin 25.5.0 arm64

## User-visible failure

The UI had remained in this state for about 88 minutes when captured and was still stuck afterward:

Permanent orphaned Working turn after 1h27m

The visible child-agent review was DONE, there was no new chat activity, and the relevant Codex processes were sleeping at 0% CPU. Restarting/re-rendering the UI cannot repair this because the durable turn row itself is still `running`.

## Observed event timeline

Session: `agent-orchestrator-1-95`  
Conversation: `c1f16460…`

| Time (UTC) | Event |
|---|---|
| 05:31:11.935 | Root AO turn `8819be57…` starts; provider turn `019ff474…` |
| 05:32:05.454 | PR review automation is queued while the root is active |
| 05:39:35.091 | Nested child turn `ab357b31…` completes |
| 05:39:35.145 | Only 54 ms later, AO prematurely dispatches queued automation as AO turn `6765ffd1…`, assigning provider ID `019ff47b…` |
| 05:39:37.317 | The automation message appears in the native Codex transcript tagged with the **original root** provider ID `019ff474…`, not the newly assigned ID |
| 05:42:13.774 | The real root turn completes |
| 07:07:48 | Screenshot: assigned automation turn has been `running` for 88.2 minutes with zero activities and zero provider events |

The queued automation was dispatched while the actual root still had another 158.6 seconds of work remaining.

The native Codex transcript ends normally with `task_complete` for `019ff474…`. The orphan provider ID `019ff47b…` never appears anywhere in that transcript.

## Contradictory live state

At the same instant:

| Layer | State |
|---|---|
| SQLite turn row | automation turn = `running` |
| Turn activities | zero |
| Normalized provider events | zero |
| Controller snapshot | `ready` |
| Provider thread state | `idle` |
| Session activity | `idle` |
| Frontend | **Working** indefinitely |
| Codex app-server / code-mode host | sleeping, 0% CPU |

This rules out a live tool call. It is a durable/provider lifecycle split.

## Deterministic code path

The controller models a conversation with one mutable `pendingTurnID` / `ackedTurnID` pair ([controller.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/service/chat/controller.go#L131-L140)).

1. Dispatch creates/binds the queued AO turn and records it as pending ([controller.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/service/chat/controller.go#L750-L762)).
2. **Every** provider `turn.started`, including nested child turns, overwrites `pendingTurnID` and `ackedTurnID` and marks the controller busy ([controller.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/service/chat/controller.go#L1315-L1322)).
3. **Every** `turn.completed`, regardless of whether it matches the dispatched root, sets the controller ready ([controller.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/service/chat/controller.go#L1350-L1359)).
4. After projection, every completion reports the session idle and drains one queued request ([controller.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/service/chat/controller.go#L1697-L1705)).
5. The Codex adapter has the same single-slot assumption: every provider start overwrites `activeTurn` ([conversation.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/adapters/chatdriver/codexappserver/conversation.go#L150-L157)), and interruption targets that value ([conversation.go](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/backend/internal/adapters/chatdriver/codexappserver/conversation.go#L486-L499)).
6. The frontend considers any durable `running` turn active ([conversation.ts](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/frontend/src/renderer/types/conversation.ts#L746-L750)), so the orphan has no natural recovery path and the Working indicator remains ([ChatWorkspace.tsx](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/frontend/src/renderer/components/chat/ChatWorkspace.tsx#L1848-L1860)).

## Why the other reports looked similar

Two earlier screenshots were not permanent backend freezes:

Short-lived parent activity

Parent still active after visible child completed

In those cases the root was genuinely active and completed shortly afterward. However, `groupByTurn` anchors the entire parent group at its first sequence position ([ChatWorkspace.tsx](https://github.com/Untrivial-ai/agent-orchestrator/blob/f0ae54b9fc2fe7a3b47076e43edbbbb90f5349b9/frontend/src/renderer/components/chat/ChatWorkspace.tsx#L1719-L1785)). Later parent activity can therefore render above completed child cards. At the bottom of the transcript the user sees “DONE” beside “Working,” with no visible evidence of what is still active.

That is an adjacent presentation defect which makes a healthy overlap indistinguishable from this permanent orphan.

### Lifecycle correctness

- Track the AO-dispatched **primary/root provider turn** separately from nested provider executions.
- Do not let nested `turn.started` replace primary ownership.
- Do not let nested `turn.completed` set the conversation Ready/Idle or drain the request queue.
- Drain only when the root thread is actually quiescent.
- Add acknowledgement recovery for an AO-dispatched provider ID that never emits `turn.started`.
- Make adapter interruption target the primary AO-dispatched turn, not the most recently started nested turn.
- Reconcile a durable `running` row when the controller is ready and the provider is authoritatively idle, while preserving the rule that failed/unknown probes are not proof of death.

Avoid inferring parentage only from timing or adding a heuristic `parentTurnId` to `ConversationTurn`. The normalizer currently discards provider thread identity, and overlapping events can also be compaction, replay, or resumed work. Codex exposes explicit thread/source metadata; AO should preserve enough of that identity, or model nested provider runs separately from user-request turns.

### UI clarity

- Show the latest activity and age of the active primary turn, with a way to jump to it.
- Keep later parent activity visible after child completion rather than anchoring it entirely above child cards.
- Surface child execution as nested/structured activity.
- If a durable turn has no provider events for a threshold, show an advisory recovery state; do not declare it dead solely from elapsed time.

## Regression coverage / acceptance criteria

- [ ] A nested child start does not replace the primary dispatched turn.
- [ ] A nested child completion does not settle the parent, report the session idle, or drain the queue.
- [ ] Automation queued during root work remains queued until root completion.
- [ ] Root completion dispatches the next request exactly once.
- [ ] A dispatched turn that never acknowledges eventually reaches a recoverable terminal/error state.
- [ ] Stop targets the primary provider turn.
- [ ] Compaction, replay, and resumed-thread flows remain correct.
- [ ] An overlapping parent/child transcript keeps the current parent activity visible.
- [ ] No durable turn remains indefinitely `running` while controller and authoritative provider state are idle.

## Impact

- Misleading multi-hour “Working” state.
- Queued PR-review automation is injected into the wrong provider turn.
- Future queue work may remain blocked behind an orphan.
- Stop/recovery can target the wrong nested turn.
- Users cannot distinguish legitimate hidden parent work from a lifecycle failure.

## Related work

- [Issue #3749](https://github.com/Untrivial-ai/agent-orchestrator/issues/3749) has the same terminal symptom (durable/in-memory turn-state desync) but a different trigger: projection transaction failure.
- [PR #3752](https://github.com/Untrivial-ai/agent-orchestrator/pull/3752) adds Stop-time reconciliation and post-commit memory mutation for that path. It can provide manual recovery after Stop, but it does not prevent nested start/completion events from changing primary ownership or prematurely draining the queue.
- [Issue #3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360) is the broader structured Chat work.

illegalcall:
## Summary

- preserve Codex's provider thread identity on normalized turn lifecycle events
- let only lifecycle events from the AO-owned root thread claim or release the controller's primary turn
- keep nested child turns durable for transcript correlation without reporting the session idle or draining queued work
- apply volatile turn ownership only after the SQLite projection commits
- serialize lifecycle projection, queue dispatch, and handoff quiescence checks at one dispatch boundary
- prevent nested child notifications from replacing the Codex adapter's fallback Stop target

## Root cause

Codex app-server multiplexes child-agent thread lifecycle notifications over the root conversation connection. AO discarded the notification's `threadId`, stored only one mutable active turn, and treated every `turn/completed` as conversation quiescence.

A child completion could therefore release queued automation while the root was still running. Codex accepted the message into that root, while AO had already minted and bound a different durable turn. Because that assigned provider turn never started, the row stayed `running` and a Chat → TUI drain waited indefinitely.

The CI race run also exposed the adjacent quiescence gap: dispatch marks a queued row running in SQLite just before installing its in-memory active-turn ID. A handoff checking those facts without the dispatch lock could briefly see neither queued nor active work and finish early. The handoff now evaluates both facts under the same lock as dispatch.

The added handoff regression covers the complete user-visible sequence: child completion cannot finish the handoff or release its accepted queue; root completion releases that queue; the handoff finishes only after the queue settles.

## Verification

- controller regression passed 50 consecutive times with the race detector
- focused controller/normalizer/adapter regressions passed 20×
- `go test -p 4 ./...`
- `go test -race -count=1 ./internal/service/chat ./internal/adapters/chatdriver/codexappserver`
- golangci-lint v2.12.2 with fresh Go/lint caches: 0 issues

Related follow-up: #3950 / #3951 durably acknowledge the recovered transition notice so the same historical recovery does not resurface after remounts or restarts.

Fixes #3878

### DIFF
--- a/backend/internal/adapters/chatdriver/codexappserver/conversation.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/conversation.go
@@ -148,12 +148,20 @@ func (c *conversation) pump() {
 		// as an absolute instant and has to become a remaining duration, and a
 		// normalizer that reads the clock itself cannot be tested deterministically.
 		for _, ev := range normalizeNotification(n, time.Now()) {
-			if ev.Kind == ports.ChatEventTurnStarted && ev.ProviderTurnID != "" {
+			rootConversation := ev.ProviderConversationID == "" || ev.ProviderConversationID == c.threadID
+			if ev.Kind == ports.ChatEventTurnStarted && ev.ProviderTurnID != "" && rootConversation {
 				c.mu.Lock()
 				c.activeTurn = ev.ProviderTurnID
 				// Snapshot the context position this turn starts from. Cheap on every
 				// turn, and the only way to know what a compaction reclaimed.
 				c.contextAtTurnStart = c.contextTokens
+				c.mu.Unlock()
+			}
+			if ev.Kind == ports.ChatEventTurnCompleted && rootConversation {
+				c.mu.Lock()
+				if c.activeTurn == ev.ProviderTurnID {
+					c.activeTurn = ""
+				}
 				c.mu.Unlock()
 			}
 			if ev.Kind == ports.ChatEventCompacted {
--- a/backend/internal/adapters/chatdriver/codexappserver/driver_test.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/driver_test.go
@@ -508,6 +508,42 @@ func TestNotificationsBecomeNeutralEvents(t *testing.T) {
 	}
 }
 
+// app-server multiplexes child-agent thread notifications over the root
+// connection. The adapter's fallback target must remain the root turn; otherwise
+// an interrupt without an explicit id can stop a child and leave the requested
+// root work running.
+func TestNestedThreadDoesNotReplaceRootActiveTurn(t *testing.T) {
+	d, srv := newTestDriver(t)
+	conv, err := d.Start(context.Background(), ports.ChatStartConfig{WorkspacePath: "/tmp/ws"})
+	if err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	defer func() { _ = conv.Close() }()
+
+	srv.push(`{"method":"turn/started","params":{"threadId":"thread-1","turn":{"id":"root-turn","status":"inProgress","items":[]}}}`)
+	srv.push(`{"method":"turn/started","params":{"threadId":"child-thread-1","turn":{"id":"child-turn","status":"inProgress","items":[]}}}`)
+	root := nextEvent(t, conv.Events(), ports.ChatEventTurnStarted)
+	child := nextEvent(t, conv.Events(), ports.ChatEventTurnStarted)
+	if root.ProviderConversationID != "thread-1" || child.ProviderConversationID != "child-thread-1" {
+		t.Fatalf("normalized lifecycle threads = root %q child %q", root.ProviderConversationID, child.ProviderConversationID)
+	}
+
+	if err := conv.Interrupt(context.Background(), ""); err != nil {
+		t.Fatalf("Interrupt: %v", err)
+	}
+	sent := srv.awaitFrame(func(f frame) bool { return f.Method == "turn/interrupt" })
+	var params struct {
+		ThreadID string `json:"threadId"`
+		TurnID   string `json:"turnId"`
+	}
+	if err := json.Unmarshal(sent.Params, &params); err != nil {
+		t.Fatalf("decode turn/interrupt params: %v", err)
+	}
+	if params.ThreadID != "thread-1" || params.TurnID != "root-turn" {
+		t.Fatalf("interrupt target = %q/%q, want thread-1/root-turn", params.ThreadID, params.TurnID)
+	}
+}
+
 // Resume must not quietly become a fresh thread: that would present unrelated
 // history as continuous.
 func TestResumeFailureDoesNotFallBackToStart(t *testing.T) {
--- a/backend/internal/adapters/chatdriver/codexappserver/normalize.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/normalize.go
@@ -136,8 +136,9 @@ func normalizeNotification(n notification, now time.Time) []ports.ChatEvent {
 			return nil
 		}
 		return []ports.ChatEvent{{
-			Kind:           ports.ChatEventTurnStarted,
-			ProviderTurnID: firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
+			Kind:                   ports.ChatEventTurnStarted,
+			ProviderTurnID:         firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
+			ProviderConversationID: p.ThreadID,
 		}}
 
 	case codexproto.MethodTurnCompleted:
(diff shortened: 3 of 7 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):
LGTM

PR #3891, opened 2026-08-12
fix(chat): persist attachments across restart

### LINKED ISSUE #3884: bug(chat): restart recovery deletes image attachments from conversation history
## Bug

Chat image attachments are stored only inside the session's disposable Git worktree. When daemon startup recovery recreates a Chat session worktree, `.ao/attachments` is omitted from preservation and deleted. The durable conversation still contains the old workspace-relative paths, so inline previews eventually become broken images and the agent can no longer read those historical attachments.

**Source:** Live AO Chat report and two observed restart cycles on 2026-08-11/12  
**Reported by:** @illegalcall  
**Affected build:** `v0.12.1-nightly.202608101434` (nightly commit `2da84a242`)  
**Analyzed against:** `21fd51acd` (`origin/main` on 2026-08-12; mechanism remains present)  
**Environment:** macOS 26.5.1, Darwin 25.5.0, arm64, zsh, Go 1.26.5; verified Go daemon on `127.0.0.1:3001`  
**Priority:** P1 — AO permanently loses attachment content from durable Chat history; no safe in-product recovery exists after the bytes are removed.  
**Confidence:** High — reproduced across two restart/recovery cycles, traced through the exact storage/lifecycle/preview code path, and verified with filesystem checks plus deterministic cache-busted HTTP requests.

## Reproduction

1. Create or open a non-scratch Chat-mode session.
2. Paste/drop/select an image and send it.
3. Confirm the image renders inline and the daemon preview URL returns `200 image/png`.
4. Restart AO Desktop such that its daemon restarts and startup recovery runs.
5. Reopen the conversation.
6. If the image remains temporarily visible, wait for Chromium to revalidate it or request the preview URL with a unique query parameter.

Example request:

```bash
curl --path-as-is -i \
  'http://127.0.0.1:3001/api/v1/sessions/<session-id>/preview/files/.ao/attachments/<attachment>.png?cache_bust=<unique>'
```

### Actual

```http
HTTP/1.1 404 Not Found
Content-Type: application/json; charset=utf-8

{"error":"not_found","code":"PREVIEW_FILE_NOT_FOUND","message":"Preview file not found",...}
```

The message and attachment path remain in SQLite, but the referenced file no longer exists in the recreated worktree. The inline image is broken, and the agent cannot reopen it.

### Expected

Attachments referenced by durable Chat history must survive daemon/app restarts and worktree recreation. The same preview URL should continue returning the original bytes, and the same `.ao/attachments/...` path should remain readable by the resumed agent.

## Observed evidence

This was observed twice in the same session, plus once in an older session:

| Cycle | Evidence |
|---|---|
| First restart | Worktree recreated at **2026-08-11 17:12:02 IST**; daemon started at **17:12:05 IST**. Attachments created before recovery were absent and returned `404`. An attachment uploaded at **17:12:53 IST**, after recovery, returned `200`. |
| Second restart | Newer images were attached after the first restart. Worktree was recreated again at **2026-08-11 20:07:40 IST**; daemon started at **20:07:43 IST**. Every one of five historical attachment paths was absent on disk and returned `404`; the same cache-busted request returned `404` in three consecutive runs. |
| Separate older session | Its durable message predated the next worktree recreation; both referenced attachments were absent afterward and returned `404 PREVIEW_FILE_NOT_FOUND`. |

Git reflog for each recreated worktree recorded a reset at the same second as recreation. The installed nightly did not change between the observations.

### Why disappearance can be delayed

Chromium may continue painting its cached image after the backing file has already been deleted. In the first observation, an old image remained visible immediately after restart and disappeared later without another restart. Cache-busted requests already returned `404` during that interval. Therefore browser cache timing can mask the loss, but does not cause it.

A renderer-only reload or Electron relaunch that leaves the daemon/worktree intact is not sufficient. The destructive trigger is daemon recovery recreating the Chat session worktree.

## Root cause

The durable and ephemeral lifetimes are mismatched:

1. [`StageAttachments`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/internal/session_manager/chat_attachments.go#L50-L76) writes the only copy under `<workspace>/.ao/attachments` and adds that directory to Git's local exclude list.
2. Conversation persistence retains the generated `.ao/attachments/<name>` reference in message text, but does not retain the attachment bytes or durable attachment metadata.
3. On daemon startup, [`reconcileLive`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/internal/session_manager/manager.go#L1911-L1942) treats every Chat controller as unable to survive the daemon and sends the session through `saveAndTeardownOne`.
4. [`StashUncommitted`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/internal/adapters/workspace/gitworktree/workspace.go#L344-L420) intentionally preserves tracked and non-ignored changes only. An attachment-only worktree appears clean, and ignored `.ao/attachments` content is skipped.
5. [`saveAndTeardownOne`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/internal/session_manager/manager.go#L1837-L1885) records the restore marker, then force-removes the worktree. Restore recreates Git-visible state but has no source for the ignored attachment bytes.
6. The preview route [`previewFile`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/internal/httpd/controllers/sessions.go#L416-L427) serves only from the current workspace, so historical attachment URLs resolve to `PREVIEW_FILE_NOT_FOUND` after recreation.
7. The renderer added by [PR #3724](https://github.com/Untrivial-ai/agent-orchestrator/pull/3724) correctly recognizes the durable path and creates an `` URL; it cannot recover bytes the backend deleted.

The behavior is cross-platform in principle because it follows shared session-manager and Git-worktree logic, although this reproduction was on macOS.

## Fix direction

Introduce a daemon-owned attachment-storage module whose canonical bytes live under AO's durable data directory, for example:

```text
$AO_DATA_DIR/attachments/<session-id>/<generated-filename>
```

Keep `.ao/attachments/...` as the agent-facing projection so existing prompts and conversations remain compatible.

Required behavior:

1. **Stage atomically:** validate the input, atomically write the canonical `0600` copy under `$AO_DATA_DIR`, then materialize it into the session worktree. Return the existing worktree-relative path.
2. **Serve durably:** for strictly validated AO-generated attachment paths, serve the canonical copy rather than relying solely on the disposable worktree. Keep ordinary preview files confined to the worktree as today.
3. **Restore projection:** after recreating a worktree and before relaunching its controller, rematerialize that session's durable attachments into `.ao/attachments` and restore the local exclude entry.
4. **Upgrade safely:** before the first destructive recovery under the fixed build, import legacy worktree-only `.ao/attachments` into durable storage. Otherwise updating to the fix can itself delete attachments created by the affected build.
5. **Tie cleanup to durable ownership:** remove canonical attachments only when the owning session/conversation is permanently deleted—not when its runtime is stopped or its worktree is recycled.
6. **Preserve security properties:** validate generated names, reject traversal and symlink escapes, retain attachment size/type limits, and never expose one session's files through another session ID.

Optional SQLite metadata can record attachment identity, MIME type, size, and message/turn ownership, but large binary blobs do not need to live in SQLite.

### Do not fix by

- removing `.ao/attachments` from Git exclusion, which would dirty user worktrees;
- forcing ignored binary files into Git preserve commits, which would pollute the project object database and blur AO-owned/user-owned state;
- changing cache headers or retrying in the renderer, which can only hide a missing-file response;
- storing only another ephemeral path without a durable byte owner.

## Regression coverage

Add a lifecycle-level test that exercises the real failure seam:

```text
start daemon with data dir A
→ create Chat session
→ attach an image
→ assert preview returns 200 and exact bytes
→ stop/restart daemon using data dir A
→ allow startup recovery to recreate/restore the session
→ assert the same preview URL returns 200 and exact bytes
→ assert the resumed agent can read the same .ao/attachments/... path
```

Also cover:

- legacy worktree-only attachment import before teardown;
- multiple attachments and collision-free stable names;
- image and non-image attachment persistence;
- traversal, malformed name, symlink escape, and cross-session access rejection;
- permanent session/conversation cleanup;
- direct HTTP success after restart without relying on Chromium cache.

The existing [`TestChatAttachmentIsStagedIntoTheWorktreeAndReadableByTheAgent`](https://github.com/Untrivial-ai/agent-orchestrator/blob/21fd51acd/backend/e2e/chat_composer_test.go#L87-L152) verifies only immediate readability and clean Git status; it does not cross a daemon restart/recovery boundary.

## Impact

- AO loses attachment bytes while retaining apparently valid history references.
- Historical screenshots cannot be reviewed by the user or agent after recovery.
- Image-only prompts can lose essential context.
- Browser caching makes the bug appear intermittent and delays visible failure.
- Reattaching is possible only if the user still has the original external file; AO itself cannot recover already-deleted attachments.
- The problem affects every attachment type staged under `.ao/attachments`, although images expose it most visibly.

## Related

- [Issue #3722](https://github.com/Untrivial-ai/agent-orchestrator/issues/3722) — fixed rendering attachment paths as inline previews; adjacent but not a duplicate.
- [PR #3724](https://github.com/Untrivial-ai/agent-orchestrator/pull/3724) — introduced the renderer path that exposes this underlying lifetime mismatch.
- [Issue #2722](https://github.com/Untrivial-ai/agent-orchestrator/issues/2722) — original task-composer attachment capability.
- [PR #3691](https://github.com/Untrivial-ai/agent-orchestrator/pull/3691) — stages delegated-task attachments in worker worktrees.
- [Issue #3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360) / [PR #3472](https://github.com/Untrivial-ai/agent-orchestrator/pull/3472) — Chat controller persistence and restart architecture.

illegalcall:
## Summary

- store every new chat and spawn attachment canonically under the AO data directory while retaining the worktree-relative path agents read
- import legacy worktree-only attachments before kill, cleanup, replacement, and restart teardown; rematerialize image and non-image files before restored controllers relaunch
- serve strictly validated AO attachment preview URLs from canonical storage so history does not depend on a live worktree or Chromium cache
- remove canonical bytes only with permanent session ownership cleanup, never on runtime stop or worktree recycling

## Review repairs

- anchor canonical and worktree operations with `os.Root`; reject traversal, unsafe generated names, symlink parent escapes, cross-session aliases, empty roots, non-regular files, SVG, and over-limit content
- make canonical creation atomic and no-overwrite, retry randomized chat-name collisions, and preserve historical bytes
- pass `context.Context` through every public attachment-store I/O operation
- restore the Git exclude only when canonical files were materialized, and stop destructive teardown when a present legacy source cannot be imported safely
- add the required real daemon restart/recovery test, including exact cache-busted HTTP bytes and a resumed agent reading the same path

## Root cause

Conversation history persisted only the `.ao/attachments` path. The bytes existed only in an ignored worktree directory, so `StashUncommitted` omitted them and restart reconciliation force-removed the only copy. Chromium cache temporarily made some images appear to survive.

## Verification

- `GOFLAGS=-p=4 npm run lint`
- `cd backend && go test -p=4 ./... -count=1`
- `cd backend && go test -race ./internal/attachmentstore ./internal/session_manager ./internal/httpd/controllers -count=1`
- Windows amd64 cross-compile of `internal/attachmentstore` tests
- `AO_CHAT_E2E=1 go test ./e2e -run ^TestChatAttachmentSurvivesDaemonRestartRecovery$ -v -count=1 -timeout=10m`

## Known limitation

An older build may already have deleted an attachments only copy. This change migrates legacy files that still exist in a live or retained worktree, but cannot reconstruct bytes deleted before upgrade.

Closes #3884


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):

nikhilachale:
LGTM also tested locally

Issue #3879, opened 2026-08-12
Chat/ACP broken on Intel Macs: bundled acp-runtime Node signed without allow-unsigned-executable-memory → V8 crash (exit 133), "peer disconnected before response"


zenprocess:
## Summary

Chat (ACP) is completely broken on Intel Macs: the app-bundled Node runtime in `Resources/acp-runtime/node/bin/node` is Hardened-Runtime signed **without** `com.apple.security.cs.allow-unsigned-executable-memory`. On x86_64 that entitlement is required for V8 to make JIT memory executable (`allow-jit`/MAP_JIT only covers the arm64 path), so the ACP adapter process crashes at V8 tier-up before it can answer `initialize`. The daemon surfaces this as:

```
switch interface <session>: resume chat: chat driver incompatible: ACP initialize:
{"code": -32603, "message": "Internal error", "data": {"error": "peer disconnected before response"}}
```

The "chat driver incompatible" label is misleading — no version incompatibility exists; the bundled Node cannot run *any* JavaScript on Intel.

## Environment

- Agent Orchestrator **0.12.3** (`dev.agent-orchestrator.desktop`), also reproduced on the previous build (0.12.x, ~2 days earlier)
- macOS **15.7.8**, native **x86_64** Intel Mac (Mac Pro, Xeon W) — no Rosetta involved
- Bundled Node: **v22.23.2** (`Contents/Resources/acp-runtime/node/bin/node`)
- ACP adapter: `@agentclientprotocol/claude-agent-acp` **0.64.2** (bundled)

## One-line repro (no AO involvement needed)

```console
$ /Applications/Agent\ Orchestrator.app/Contents/Resources/acp-runtime/node/bin/node --version
v22.23.2
$ /Applications/Agent\ Orchestrator.app/Contents/Resources/acp-runtime/node/bin/node -e 'console.log(1)'
# V8 fatal: "Check failed: 12 == (*__error())" in v8::base::OS::SetPermissions
#   -> MemoryAllocator::SetPermissionsOnExecutableMemoryChunk
# stack includes Runtime_BytecodeBudgetInterrupt_Ignition (first JIT tier-up)
$ echo $?
133
```

[...]

PR #3896, opened 2026-08-12
fix(desktop): preserve Intel Node JIT entitlement

axisrow:
Refs #3879.

## What changed

- Applies a per-file entitlement override only to the bundled ACP Node at `Contents/Resources/acp-runtime/node/bin/node`, and selects it **from the binary content**: the override applies iff the Mach-O header of the file being signed contains an x86_64 slice (a thin x64 binary, or a universal carrying x64). arm64-only binaries keep the default `allow-jit`.
- **The signing host's architecture is never consulted.** `@electron/osx-sign` passes `optionsForFile` only `(filePath)`, and the canonical downstream signer processes both arm64 and x64 artifacts from one host, so `process.arch` was the wrong selector (review follow-up). The new `frontend/makers/macho-archs.ts` parses the Mach-O header (thin `MH_MAGIC_64`, fat `FAT_MAGIC`/`FAT_MAGIC_64`; endianness derived from the magic, cputype-only classification) and `macSignOptionsForFile` keys off it. Tests stub `process.arch` to `x64`, `arm64`, and a nonsense value and pin identical results.
- **Fails closed**: an unreadable or unparseable file at the nested-Node path throws out of `optionsForFile` and aborts the signing pass — no silent fallback to defaults, which is exactly the #3879 crash.
- Preserves `allow-jit` and adds `allow-unsigned-executable-memory`, which x64 V8 needs when Hardened Runtime blocks executable-memory permission changes.
- Leaves every other bundle file on `@electron/osx-sign` defaults.
- Wires the override through both supported Forge signing inputs: `APPLE_SIGNING_IDENTITY` and `CSC_LINK`; unsigned local packages remain unsigned.
- Documents the canonical signing contract in `frontend/docs/desktop-release.md` ("macOS signing contract for the bundled ACP runtime"), including pre-publication duties: inspect the nested Node's entitlements in the final artifact and execute the final x64 zip's Node on a native Intel runner, keeping the arm64 verification.
- Extends `frontend/scripts/verify-mac-artifact.sh` to inspect the final nested Node entitlements and execute it only after codesign, Gatekeeper, and stapler all pass. The script is the local diagnostic and the public mac-update-e2e baseline check; the pre-publication gate remains the release conductor's verifier, which now has a documented contract to mirror.

Scope note: the public CI build in `build-artifacts.yml` stays intentionally unsigned, so `osxSign`/`optionsForFile` do not run there — the entitlement decision is enforced wherever signing actually happens (locally signed builds today; the canonical signer once it mirrors the documented contract).

The branch is rebased on current `main`.

## Review follow-up addressed

- Arch selection is keyed off the artifact being signed, not the host (`arch` parameter removed).
- The public verifier is documented as the local diagnostic counterpart, not the publication gate.
- Removed the unrelated `AgentModelCombobox` locale-formatting commit and the net-zero screenshot add/remove pair (tree verified byte-identical apart from the locale file).
- `Closes` demoted to `Refs` until the paired conductor-side change lands (contract in `desktop-release.md`).

## Tests

- Focused suites: 69 passed — new Mach-O parser tests, forge.config signing/postMake, maker-dmg, verify-mac-artifact smoke.
- Mach-O parser coverage: synthetic thin/fat fixtures (incl. `FAT_MAGIC_64`, byte-swapped thin, arm64e cputype-only classification, i386 not masked into x86_64), truncated/garbage/non-Mach-O headers throw, host-arch independence, macOS sanity check against a real Mach-O.
- Frontend typecheck and e2e typecheck: passed. Full frontend suite: 267 files; 3,594 passed, 7 skipped, and 12 failed — all 12 in `browser-profile-import.test.ts`, from a local `better-sqlite3` build carrying the Electron ABI after a local packaged build (environment artifact; neither the module nor that file is touched by this branch).
- Shell syntax and shellcheck: passed.
- Trust-order regression: a mocked untrusted app fails `spctl` and the nested Node sentinel is never executed.
- Real signer A/B (original round): official Node v22.23.2 x64, `@electron/osx-sign` 1.3.3, Hardened Runtime — the entitlement list is unchanged; only the selection mechanism moved from host arch to file content.

## A/B evidence

The screenshots are attached as review evidence and are not part of the PR code diff.

### A — current Forge defaults produce a valid seal but Node crashes

Before: x64 V8 crashes with exit 133

The final x64 Node has `allow-jit` but not `allow-unsigned-executable-memory`. The outer app seal is valid, yet JavaScript hits the reported V8 `SetPermissions` assertion and exits 133.

### B — the selector signs the same Node with the required x64 entitlement

After: x64 Node executes JavaScript

The same signed Node and command exit 0. This exercises the exact Packager → osx-sign → nested Node/V8 boundary changed by the PR.

## Remaining for #3879

The change that reaches users is the canonical signer's entitlement step in the private release repository. This PR ships the public contract (`desktop-release.md`), the content-based reference implementation (`machoHasX86_64Slice`), and the local diagnostic; `Closes` returns once the conductor side lands:

1. its signing step applies the entitlements per the documented contract (x64-slice predicate, fail-closed);
2. its final-artifact verifier inspects the nested Node and executes the final x64 zip on a native Intel runner.

### DIFF
--- a/frontend/docs/desktop-release.md
+++ b/frontend/docs/desktop-release.md
@@ -48,45 +48,6 @@ Windows installers follow the same boundary (#4502): the NSIS maker
 when signing credentials are present in the environment, so the public build
 stays unsigned while the conductor signs with its own credentials downstream.
 
-## macOS signing contract for the bundled ACP runtime
-
-This repository builds unsigned, so `packagerConfig.osxSign` — and the per-file
-hook `macSignOptionsForFile` in `frontend/forge.config.ts` — never runs on
-published bytes. The canonical signer re-signs the bundle downstream and is the
-only place the following contract is enforced. Whoever signs a bundle containing
-`<App>.app/Contents/Resources/acp-runtime/node/bin/node` must:
-
-- give the nested Node `com.apple.security.cs.allow-jit` **and**
-  `com.apple.security.cs.allow-unsigned-executable-memory` if and only if the
-  binary contains an x86_64 Mach-O slice (a thin x64 binary, or a universal
-  carrying an x64 slice); an arm64-only binary keeps `allow-jit` alone. V8 uses
-  mprotect for JIT on x64, which Hardened Runtime blocks without the second
-  entitlement, and the Node then crashes in `SetPermissions` (#3879);
-- derive that decision from the Mach-O header of the file being signed
-  (`frontend/makers/macho-archs.ts`, `machoHasX86_64Slice`), never from the
-  signing host's architecture (`process.arch`, `uname -m`): one signing host
-  handles both architectures;
-- fail the signing pass when that binary cannot be read or parsed — never fall
-  back to the host architecture or to no entitlements;
-- leave every other bundle file on `@electron/osx-sign` defaults; widening the
-  executable-memory entitlement beyond the x64 ACP Node is out of scope.
-
-Before publication, the final artifacts must prove the contract: the nested
-Node's entitlements are inspected in the signed bundle, and the shipped x64
-zip's Node actually executes JavaScript on a native Intel runner (the public
-build matrix's `macos-15-intel` leg is the precedent). Keep the arm64
-verification as well.
-
-The verifier-side counterpart is rule 5 of
-`frontend/scripts/verify-mac-artifact.sh`: `codesign -d --entitlements :-` on
-the nested Node, `lipo -archs` for the arch decision, and a guarded `node -e`
-execution only after codesign, Gatekeeper, and stapler all pass. The JS helper
-and `lipo` agree on the one invariant that matters — "contains an x86_64
-slice"; `lipo` may print `arm64e` for an arm64 slice (it names slices from
-cpusubtype) where the helper reports `arm64` (it classifies by cputype), which
-is intentional. Confirm what actually landed on the bytes with
-`codesign -d --entitlements :- <node> | plutil -p`.
-
 ## Channels
 
 - **Stable** releases are deliberate production cuts. After all verification
@@ -104,8 +65,7 @@ but changing it in this repository is not a release trigger.
 
 ## Artifact verification
 
-The conductor owns the authoritative verification and publication gates; this
-script is the local diagnostic counterpart, not the publication gate. For a
+The conductor owns the authoritative verification and publication gates. For a
 local diagnosis of a downloaded macOS artifact, use the repository's supported
 verification script:
 
--- a/frontend/docs/desktop-release.md
+++ b/frontend/docs/desktop-release.md
@@ -48,6 +48,45 @@ Windows installers follow the same boundary (#4502): the NSIS maker
 when signing credentials are present in the environment, so the public build
 stays unsigned while the conductor signs with its own credentials downstream.
 
+## macOS signing contract for the bundled ACP runtime
+
+This repository builds unsigned, so `packagerConfig.osxSign` — and the per-file
+hook `macSignOptionsForFile` in `frontend/forge.config.ts` — never runs on
+published bytes. The canonical signer re-signs the bundle downstream and is the
+only place the following contract is enforced. Whoever signs a bundle containing
+`<App>.app/Contents/Resources/acp-runtime/node/bin/node` must:
+
+- give the nested Node `com.apple.security.cs.allow-jit` **and**
+  `com.apple.security.cs.allow-unsigned-executable-memory` if and only if the
+  binary contains an x86_64 Mach-O slice (a thin x64 binary, or a universal
+  carrying an x64 slice); an arm64-only binary keeps `allow-jit` alone. V8 uses
+  mprotect for JIT on x64, which Hardened Runtime blocks without the second
+  entitlement, and the Node then crashes in `SetPermissions` (#3879);
+- derive that decision from the Mach-O header of the file being signed
+  (`frontend/makers/macho-archs.ts`, `machoHasX86_64Slice`), never from the
+  signing host's architecture (`process.arch`, `uname -m`): one signing host
+  handles both architectures;
+- fail the signing pass when that binary cannot be read or parsed — never fall
+  back to the host architecture or to no entitlements;
+- leave every other bundle file on `@electron/osx-sign` defaults; widening the
+  executable-memory entitlement beyond the x64 ACP Node is out of scope.
+
+Before publication, the final artifacts must prove the contract: the nested
+Node's entitlements are inspected in the signed bundle, and the shipped x64
+zip's Node actually executes JavaScript on a native Intel runner (the public
+build matrix's `macos-15-intel` leg is the precedent). Keep the arm64
+verification as well.
+
+The verifier-side counterpart is rule 5 of
+`frontend/scripts/verify-mac-artifact.sh`: `codesign -d --entitlements :-` on
+the nested Node, `lipo -archs` for the arch decision, and a guarded `node -e`
+execution only after codesign, Gatekeeper, and stapler all pass. The JS helper
+and `lipo` agree on the one invariant that matters — "contains an x86_64
+slice"; `lipo` may print `arm64e` for an arm64 slice (it names slices from
+cpusubtype) where the helper reports `arm64` (it classifies by cputype), which
+is intentional. Confirm what actually landed on the bytes with
+`codesign -d --entitlements :- <node> | plutil -p`.
+
 ## Channels
 
 - **Stable** releases are deliberate production cuts. After all verification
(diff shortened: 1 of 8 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Deep review, conflict resolution, and A/B verification are complete.

Review fixes now on the branch:
- merged current `main`;
- added configuration-boundary coverage for both signing env paths and unsigned builds;
- restricted the weakened executable-memory entitlement to x64;
- added final-artifact checks for the nested Node to the canonical verifier;
- made Node execution occur only after codesign, Gatekeeper, and stapler all succeed;
- added a regression proving an untrusted app never executes nested code;
- updated release instructions to verify both macOS architectures.

The screenshots are attached as PR review evidence only and are not committed in the implementation diff.

| Current signer defaults | PR #3896 |
|---|---|
| Before | After |
| Valid app seal, real Node 22.23.2 x64, V8 exits 133. | Same signer/runtime/path/command, both entitlements present, exit 0. |

Verification: focused 20/20; typecheck; full frontend 2,507 passed / 1 skipped; shell syntax/shellcheck; exact x64 signer A/B. No code-review blocker remains. The final Developer ID/notarized x64 release artifact remains the release-conductor gate and is now checked by `verify-mac-artifact.sh`.

axisrow:
## 🔍 Local review (cycle 1) — round 32b53bf4-e4a1-430f-985b-85dd06904d3e
Reviewed locally (`/review` + Codex companion), no bots pinged. Critical-only bar.

| Verdict | Reviewer | Finding | Location |
|---|---|---|---|
| FIX | codex | The automated release pipeline never actually exercises the new Intel-Node entitlement check on a real x64 build — the DMG path explicitly skips it and the x64 zip check is a manual, undocumented-as-enforced step, so CI cannot catch a regression of the exact bug this PR fixes. | frontend/scripts/verify-mac-artifact.sh:184 |
| SKIP | claude | The signing helper defaults its target architecture to the machine running the build rather than an explicit parameter, which is fragile in general, but this repo always builds each architecture on a matching native runner today, so the gap does not affect current releases. | frontend/forge.config.ts:33 |
| SKIP | claude | The entitlement check parses plutil text output with a literal string match instead of a structured extractor, which is stylistically brittle but only causes the release gate to block a build, never to falsely pass a broken one. | frontend/scripts/verify-mac-artifact.sh:174 |
| SKIP | claude | The failure-accumulation helper correctly survives strict-mode bash: a missing bundled Node is recorded as a failure without aborting the script, so remaining checks still run and get reported together. | frontend/scripts/verify-mac-artifact.sh:186 |

**Totals:** 1 FIX (fixing now), 3 SKIP.

axisrow:
## 📋 Review summary — all cycles

Reviewed locally (`/review` + Codex companion), no bots pinged, across 2 cycles.

| Cycle | Reviewer | Finding | Verdict | Resolution |
|---|---|---|---|---|
| 1 | codex | The automated release pipeline never actually exercises the new Intel-Node entitlement check on a real x64 build — the DMG path explicitly skips it and the x64 zip check is a manual, undocumented-as-enforced step, so CI cannot catch a regression of the exact bug this PR fixes. | FIX | Fixed in 19ee7e055 |
| 1 | claude | The signing helper defaults its target architecture to the machine running the build rather than an explicit parameter, which is fragile in general, but this repo always builds each architecture on a matching native runner today, so the gap does not affect current releases. | SKIP | Left as-is — no exploitable gap given this repo's native per-arch build matrix |
| 1 | claude | The entitlement check parses plutil text output with a literal string match instead of a structured extractor, which is stylistically brittle but only causes the release gate to block a build, never to falsely pass a broken one. | SKIP | Left as-is — fails closed, not open; not a release risk |
| 1 | claude | The failure-accumulation helper correctly survives strict-mode bash: a missing bundled Node is recorded as a failure without aborting the script, so remaining checks still run and get reported together. | SKIP | Left as-is — verified already correct by design |
| 2 | claude + codex | A second local review round (built-in /review plus the Codex companion) against the fix for the release-gate gap found no further critical issues and approved the change. | — | Clean round, PR approved |

**Totals:** 1 FIX (resolved), 3 SKIP, 0 UNVERIFIED. Both reviewers returned a verdict every round — no silent reviewer. CI watch next.

axisrow:
@illegalcall LGTM, well done

axisrow:
@ronishrohan could you pick this up? @illegalcall completed the deep review on Aug 17 (review fixes on the branch, x64 A/B verified, no blockers left) and the PR has been ready to merge since, but the thread has gone quiet. Happy to do whatever else is needed.

illegalcall:
Correction to my earlier review: after tracing the current release architecture end-to-end, there is still a blocking integration issue.

The entitlement change is mechanically correct, but it is wired into the wrong signing path:

- The public [`build-artifacts.yml`](https://github.com/Untrivial-ai/agent-orchestrator/blob/9f9a0f4326a9fa089e83b19582b4d571022729c2/.github/workflows/build-artifacts.yml#L4-L12) deliberately runs Forge **unsigned** and never supplies `APPLE_SIGNING_IDENTITY` or `CSC_LINK`.
- Canonical releases are signed, notarized, and packaged afterward by the private release conductor, as documented in [`desktop-release.md`](https://github.com/Untrivial-ai/agent-orchestrator/blob/9f9a0f4326a9fa089e83b19582b4d571022729c2/frontend/docs/desktop-release.md#L13-L44).
- Therefore this PR's `osxSign.optionsForFile` callback never signs the artifact that is ultimately published, and its signed-ZIP `postMake` branch is also skipped.
- The conductor owns a separate final-artifact verifier; changing the public `frontend/scripts/verify-mac-artifact.sh` does not change that publication gate.

As a result, the canonical downstream signer can still re-sign the bundled x64 Node with `@electron/osx-sign` defaults, omit `com.apple.security.cs.allow-unsigned-executable-memory`, and reproduce #3879 even after this PR merges. The focused tests pass, but they validate the unused/local signed-Forge path rather than the production release seam.

There is also an architecture-selection problem to avoid when moving the fix: `@electron/osx-sign` calls `optionsForFile(filePath)` with one argument, so this PR's default selector is `process.arch`. The conductor processes both arm64 and x64 artifacts from one signing host, so the policy must key off the artifact currently being signed (or inspect the Mach-O), not the host architecture.

Required before `Closes #3879` is accurate:

1. Add a paired change to the private conductor's downstream `signApp` invocation so only the x64 ACP Node receives both `allow-jit` and `allow-unsigned-executable-memory`.
2. Extend the conductor-owned final-artifact verifier to inspect the nested Node and execute the final x64 ZIP on a native Intel runner before publication; retain the arm64 verification as well.
3. Keep the public verifier changes only as the local diagnostic counterpart, and document that distinction.
4. Remove the unrelated `AgentModelCombobox` locale-formatting commit from this PR.

The narrow per-file entitlement policy and trust-before-execution ordering are good and should be preserved; the blocker is that neither currently reaches the canonical release path.

axisrow:
Thanks — the correction is right, and the rework is pushed.

**Arch selection is content-based now.** The `arch` parameter is gone; the policy keys off the Mach-O header of the file being signed (endianness from the magic, classification by cputype only), with the predicate "contains an x86_64 slice". Tests stub `process.arch` to `x64`/`arm64`/nonsense and pin identical results, so the one-host-two-artifacts case cannot regress to host arch. Unreadable or unparseable files abort the signing pass — no silent fallback to defaults, since that fallback is the #3879 crash.

**Verifier scope.** Kept as the local diagnostic per your point 3. One question: `mac-update-e2e.yml` (~line 114) calls this same script "the same canonical script the release pipeline's post-staple gate uses". Is the conductor's final-artifact verifier literally this script, or a separate one? If separate, should these checks be ported into it (sketch below) and that comment corrected in a follow-up?

**Your steps 1–2: we can't do them.** They live in the private `ao-releases` repository, which external contributors can't access — that part has to land on your side. To make it mechanical, the change the public docs now describe is:

```ts
const optionsForFile = (filePath: string) =>
	!filePath.endsWith("/Contents/Resources/acp-runtime/node/bin/node") ? {}
	: machoHasX86_64Slice(filePath) ? { entitlements: [
		"com.apple.security.cs.allow-jit",
		"com.apple.security.cs.allow-unsigned-executable-memory",
	] } // throws if unreadable: fail closed
	: {};
```

plus, in the conductor's verifier: after codesign/spctl/stapler pass, assert both entitlements on the nested Node (`lipo -archs`, require the second one iff `x86_64` appears), execute it with a sentinel `node -e`, and run the final x64 zip through this on a native Intel runner, keeping the arm64 pass.

**Branch hygiene.** Dropped the unrelated locale commit and the net-zero screenshot pair. `Closes` demoted to `Refs` until the conductor-side change lands — blocked on private-repo access, not on anything left in this PR.

axisrow:
@illegalcall friendly bump

illegalcall:
This fixes intel mac users acp tui to chat ui switching interface fails.
This happened due to node crashing in those cases. It now applies allow-unsigned-executable-memory in those cases.

illegalcall (approved):

axisrow:
Thanks for the merge 🎉

PR #3898, opened 2026-08-12
fix(browser): re-measure preview bounds on OS-level window resize/move

### LINKED ISSUE #2491: bug(browser): preview WebContentsView renders in the wrong position
## Bug

The browser preview panel (`ao preview <url>` / the in-app browser view) renders its native `WebContentsView` in the wrong position — misaligned relative to where the app's placeholder UI expects it to sit. Reporter will attach a screenshot to this issue.

**Reporter's environment:** Debian Linux, triggered via `ao preview`. Not yet confirmed whether this is Linux-specific or reproduces on other platforms too — see caveat below.

## How positioning works (traced, not guessed)

- Main process creates the view via `createBrowserViewHost()` and adds it as a child view (`frontend/src/main/browser-view-host.ts:137-163`), starting at an off-screen sentinel bounds.
- Positioning is entirely **renderer-driven over IPC** — there is no bounds computation in the main process. `useBrowserView` (`frontend/src/renderer/hooks/useBrowserView.ts`) reads a placeholder `<div>`'s `getBoundingClientRect()` (`useBrowserView.ts:59-71`), intersects it with the enclosing `[data-panel]` column's rect, and sends the result via `window.ao.browser.setBounds({ viewId, rect, visible })` (`useBrowserView.ts:120`).
- The main process's `browser:setBounds` handler (`browser-view-host.ts:235`) clamps the incoming rect's *width/height* against `mainWindow.getContentBounds()` (`clampBoundsToWindow`, `browser-view-host.ts:115-135`) — note it does **not** re-derive `x`/`y` from the window, it only shrinks the rect to fit — then applies it directly via `view.setBounds(bounds)`.
- Re-measurement is triggered only by renderer-side signals: a `ResizeObserver` on the slot node and its panel-column ancestor, a window `resize`/`scroll` listener, and a ~280ms "settle" re-measure after URL/active-tab changes (`useBrowserView.ts:147-175, 211-222`).

## Suspected fragility (not a confirmed root cause, and not confirmed to be platform-specific)

**There is no main-process window move/resize listener.** `mainWindow.on(...)` in `frontend/src/main.ts:233` only handles `"closed"` — nothing re-syncs the view's bounds when the OS-level window itself moves, resizes, maximizes, or changes monitor/scale-factor. All re-sync depends entirely on the renderer's `ResizeObserver`/`window.resize` events firing correctly.

I checked for platform-specific branches (`process.platform`, DPI/scale-factor handling) in the positioning path (`browser-view-host.ts`, `useBrowserView.ts`) and found none — the code is platform-agnostic by design. So this isn't a case of AO special-casing (or mishandling) Linux specifically. It's an open question whether the missing resize/move listener above, or a DPI/fractional-scaling coordinate mismatch, behaves differently across window managers/compositors (X11, Wayland, macOS, Windows) in a way that would make this manifest only on some platforms — that hasn't been confirmed either way yet.

Git history shows prior positioning/clipping fixes to this component (`0b4c2090`, `acd811dc`, `b39a4795`) but none mention Linux/X11/Wayland/DPI specifically, so this looks like a genuinely new observation rather than a regression of a known platform-specific fix.

## Screenshots

Image

## Repro (to be confirmed/expanded by reporter)

1. On Debian Linux, run `ao preview <url>` inside a session.
2. Observe the browser view rendering offset from its expected panel position (screenshot to follow).
3. Worth checking during triage: does resizing/moving/maximizing the app window change the offset? Does it differ between X11 and Wayland, or with fractional display scaling on vs off? Does it reproduce on macOS/Windows?

## Next steps

Needs reproduction (on the reporter's Debian machine, and ideally on macOS/Windows for comparison) to determine whether this is platform-specific at all, and if so whether the main-process resize/move gap above is the cause, or a DPI/fractional-scaling issue at the Electron/Chromium compositor level (outside AO's own code).

axisrow:
## Summary

Fixes #2491.

The native browser preview's `WebContentsView` bounds are computed entirely
renderer-side (`useBrowserView`): a `ResizeObserver` on the placeholder slot
plus a `window` `resize`/`scroll` listener measure the slot and send bounds
to the main process (`browser:setBounds`), which clamps and applies them
directly via `view.setBounds()`.

There was no main-process listener for the `BrowserWindow`'s own
`resize`/`move`/`maximize`/`unmaximize` events. All re-sync depended
entirely on the renderer's own DOM signals firing for those OS-level
changes — which the issue's traced investigation flagged as the likely
fragility behind the view rendering at stale/misaligned bounds.

## Fix

- `main.ts`: forward `resize`/`move`/`maximize`/`unmaximize` from the
  `BrowserWindow` to the renderer over a new `window:remeasure` IPC event
  (mirrors the existing `window:fullscreen` channel/pattern).
- `preload.ts`: expose it as `window.ao.window.onRemeasure`.
- `useBrowserView.ts`: settle-measure (immediate + ~280ms settle re-measure)
  on that signal, the same handling already used for pop-out and inspector
  collapse/expand transitions — both of those are also position-only shifts
  the `ResizeObserver` alone can miss.

This closes the gap called out in the issue without guessing at the
DPI/compositor-level question (Wayland vs X11 vs fractional scaling) — it
makes bounds re-sync independent of whether the DOM's own resize signals
fire for a given OS-level window change, on any platform.

## Tests

- Added a renderer test asserting a `window:remeasure` signal triggers a
  re-measure and re-send of bounds using the slot's latest geometry.
- `npx vitest run src/renderer/hooks/useBrowserView.test.tsx
  src/main/browser-view-host.test.ts` — 82 passed.
- Full `npx vitest run` — only pre-existing failures in unrelated
  `src/landing/**` tests (missing `cheerio` dependency in that
  subproject), no relation to this change.

### DIFF
--- a/frontend/e2e/support/fake-bridge.ts
+++ b/frontend/e2e/support/fake-bridge.ts
@@ -99,6 +99,7 @@ export async function installFakeBridge(page: Page, opts: FakeBridgeOptions = {}
 					onMaximized: () => () => undefined,
 					isFullScreen: async () => false,
 					onFullScreen: () => () => undefined,
+					onRemeasure: () => () => undefined,
 				},
 				theme: {
 					set: async () => undefined,
@@ -585,6 +586,7 @@ export async function installFakeAgent(page: Page, opts: FakeAgentOptions = {}):
 					onMaximized: () => () => undefined,
 					isFullScreen: async () => false,
 					onFullScreen: () => () => undefined,
+					onRemeasure: () => () => undefined,
 				},
 				theme: {
 					set: async () => undefined,
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -582,6 +582,20 @@ async function createWindowInternal(): Promise<void> {
 	mainWindow.on("leave-full-screen", pushFullScreen);
 	mainWindow.on("maximize", pushMaximized);
 	mainWindow.on("unmaximize", pushMaximized);
+	// The native browser preview's bounds are entirely renderer-driven (measured
+	// from a placeholder DOM node via ResizeObserver/window "resize"). An
+	// OS-level window move, resize, or maximize toggle does not reliably fire
+	// those DOM signals on every platform/window manager, which leaves the
+	// WebContentsView painted at stale bounds. Forward these window events so
+	// the renderer can re-measure and re-send bounds regardless of whether its
+	// own listeners fired.
+	const pushRemeasure = () => {
+		getShellWebContents()?.send("window:remeasure");
+	};
+	mainWindow.on("resize", pushRemeasure);
+	mainWindow.on("move", pushRemeasure);
+	mainWindow.on("maximize", pushRemeasure);
+	mainWindow.on("unmaximize", pushRemeasure);
 	mainWindow.on("blur", () => {
 		keybindingRecordingActive = false;
 	});
--- a/frontend/src/preload.ts
+++ b/frontend/src/preload.ts
@@ -239,6 +239,17 @@ const api = {
 				ipcRenderer.off("window:fullscreen", wrapped);
 			};
 		},
+		// Fired by the main process on OS-level window resize/move/maximize/
+		// unmaximize, so panels whose bounds are computed in the renderer (the
+		// native browser preview) can re-measure even when the DOM's own
+		// resize/ResizeObserver signals don't fire for that change.
+		onRemeasure: (listener: () => void) => {
+			const wrapped = () => listener();
+			ipcRenderer.on("window:remeasure", wrapped);
+			return () => {
+				ipcRenderer.off("window:remeasure", wrapped);
+			};
+		},
 	},
 	theme: {
 		// Propagate the app's theme preference to Electron's nativeTheme so embedded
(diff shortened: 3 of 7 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:
@axisrow Correction: I marked this stale by mistake. Your latest work here is from Aug 31, which is recent. The stale label is removed. Sorry for the noise.

i-trytoohard:

axisrow:

axisrow:
Closing as superseded: the preview bounds re-measure on OS-level window resize/move is already in main, including the onRemeasure fake-bridge e2e hook — this branch applies as empty on current main (patch-id verified).

axisrow:
Reopening — closed in error: the onRemeasure fake-window-bridge e2e hook and the OS-level resize/move re-measure are not in upstream main (verified by git grep against origin/main). My earlier check hit the fork's fold-back copy instead. Back to review.

axisrow:

axisrow:
@illegalcall friendly bump

PR #3900, opened 2026-08-12
fix(sessions): store activity timestamps in UTC so SQL comparisons hold

### LINKED ISSUE #3899: bug(agent-switch): local-zone activity_last_at strands the switch saga in stopping_source and makes the daemon unbootable
### Summary

An agent switch can strand itself in `stopping_source` with no way out, and the daemon then **refuses to start at all**: boot reconciliation returns the same error the switch failed with, so the whole app is unusable until the row is repaired by hand.

The root cause is not in the saga. It is a storage detail: `sessions.activity_last_at` is written from a bare `time.Now()` in two places, and the SQLite driver stores a `time.Time` by its `String()` form.

### Root cause

`modernc.org/sqlite` stores a `time.Time` by its `String()` rendering, so a value written from a local-zone clock keeps both its offset and its monotonic reading:

```
2026-06-28 18:45:08.349363 +0800 CST m=+25660.013723251     <- what a local clock leaves
2026-06-28 10:45:08.349363 +0000 UTC                        <- what every other writer produces
```

`activity_last_at` is compared **directly in SQL**, and SQLite compares these as strings. A `+0800` wall clock sorts *above* the UTC rendering of a later instant, so such a row stops behaving like a timestamp.

`MarkSessionAgentSwitchSourceStopped` gates on `activity_last_at <= stopped_at`. Against a mismatched row it matches **zero rows** — which the caller reads as "ownership changed concurrently", not as "the timestamp is malformed".

The two writers producing these values:

- `backend/internal/observe/reaper/reaper.go` — default clock
- `backend/internal/observe/activity/observer.go` — its `now`, which becomes the observation `Timestamp`

### Impact chain

1. Source agent is already stopped, but the confirm matches 0 rows → `confirmed=false`, `err=nil`.
2. Saga stays in `stopping_source`; the in-memory operation gate stays held.
3. Resume answers `AGENT_RESUME_IN_PROGRESS` (misleading — nothing is resuming, a *switch* is stuck).
4. Repeating the switch fails identically.
5. `Manager.Reconcile` propagates the error at boot → **the daemon does not start**.

On one affected profile, 184 session rows carried the malformed form.

### Fix (PR to follow)

Three levels, because each alone leaves a gap:

- **Sources** — both clocks emit UTC.
- **Storage** — `normalActivity` normalizes `LastActivityAt` to UTC on every write path, rather than trusting each caller's clock.
- **Data** — a migration rewrites rows already stored in the local-zone form.

The store-level regression test fails without the storage fix with exactly the production symptom: `confirmed=false, err=nil`.

### Notes for whoever picks this up

Two things worth flagging separately, not fixed here:

- A single stranded saga making the daemon unbootable is arguably its own bug — boot reconciliation could quarantine the row instead of refusing to start.
- Saga reconciliation runs only at daemon boot; there is no periodic pass and no state timeout, so a saga stuck mid-flight stays stuck until a restart.

Also note `AGENT_RESUME_IN_PROGRESS` does not distinguish operation kind, which is why a stuck *switch* reports itself as a *resume* and sends you looking in the wrong place.

axisrow:
Fixes #3899.

An agent switch could strand itself in `stopping_source` with no way out: the source agent was already gone, resume answered `AGENT_RESUME_IN_PROGRESS`, a repeat switch failed the same way, and the daemon then **refused to start** because boot reconciliation returned the same error.

The cause is a storage detail, not a saga bug. `modernc.org/sqlite` stores a `time.Time` by its `String()` form, so a value written from a local-zone clock keeps its offset and monotonic reading:

```
2026-06-28 18:45:08.349363 +0800 CST m=+25660.013723251
```

instead of the form every other writer produces:

```
2026-06-28 10:45:08.349363 +0000 UTC
```

`activity_last_at` is compared directly in SQL, and SQLite compares these as strings — a `+0800` wall clock sorts above the UTC rendering of a *later* instant. `MarkSessionAgentSwitchSourceStopped` gates on `activity_last_at <= stopped_at`, so against such a row it matches zero rows, which the caller reads as "ownership changed concurrently".

### Changes

Three levels, because each alone leaves a gap:

- **Sources** — `observe/reaper/reaper.go` and `observe/activity/observer.go` emit UTC.
- **Storage** — `normalActivity` normalizes `LastActivityAt` on every write path, rather than trusting each caller's clock. This is the durable guard: a future writer with a bare `time.Now()` cannot reintroduce the bug.
- **Data** — migration `0088` rewrites rows already stored in the local-zone form.

### Migration notes

The migration matches on `activity_last_at NOT LIKE '% +0000 UTC'` rather than on the monotonic suffix, because both shapes occur in the wild (`… +0700 +07 m=+995.1` and `… +0700 +07`). The offset is located by search rather than at a fixed column, since Go trims trailing zeros from fractional seconds and that moves the sign's position. Fractional digits are carried over verbatim — `strftime`'s `%f` renders milliseconds only and would truncate microseconds.

### Tests

- `agent_switching_store_test.go` — fails without the storage change with exactly the production symptom: `confirmed=false, err=nil`.
- `migrate_activity_zone_test.go` — 6 cases (monotonic, short fraction, no-monotonic local zone, midnight crossing, west-of-UTC, already-canonical). It reads via `CAST(activity_last_at AS TEXT)`: scanning the column into a string lets the driver re-render it as RFC 3339 and hide the very difference under test.

Verified on an affected profile: 184 malformed rows → 0, the stranded saga moved `stopping_source` → `failed`, and the daemon boots again with sessions intact.

### DIFF
--- a/backend/internal/observe/activity/observer.go
+++ b/backend/internal/observe/activity/observer.go
@@ -74,7 +74,12 @@ func New(sessions sessionSource, sink activitySink, runtime outputReader, agents
 		o.outputLines = DefaultOutputLines
 	}
 	if o.clock == nil {
-		o.clock = time.Now
+		// UTC, not bare time.Now: this timestamp becomes sessions.activity_last_at,
+		// and the SQLite driver stores a time.Time by its String() form. A
+		// local-zone value keeps its monotonic reading and offset, which then
+		// fails to compare as a timestamp against the "… +0000 UTC" rows the
+		// other writers produce.
+		o.clock = func() time.Time { return time.Now().UTC() }
 	}
 	if o.logger == nil {
 		o.logger = slog.Default()
--- a/backend/internal/observe/reaper/reaper.go
+++ b/backend/internal/observe/reaper/reaper.go
@@ -96,7 +96,13 @@ func New(sink runtimeObservationSink, sessions sessionSource, runtime runtimePro
 		r.tick = DefaultTickInterval
 	}
 	if r.clock == nil {
-		r.clock = time.Now
+		// UTC, not bare time.Now: the observed timestamp reaches sessions.
+		// activity_last_at, and the SQLite driver stores a time.Time by its
+		// String() form. A local-zone value keeps its monotonic reading and
+		// offset ("… +0700 +07 m=+995.1"), which no longer compares as a
+		// timestamp against the "… +0000 UTC" rows every other writer produces —
+		// and those comparisons gate the agent-switch source-stop predicate.
+		r.clock = func() time.Time { return time.Now().UTC() }
 	}
 	if r.logger == nil {
 		r.logger = slog.Default()
--- a/backend/internal/storage/sqlite/migrate_activity_zone_test.go
+++ b/backend/internal/storage/sqlite/migrate_activity_zone_test.go
@@ -68,13 +68,13 @@ func TestMigration0088NormalizesLocalZoneActivityTimestamps(t *testing.T) {
 	// The normalized column has to compare as a timestamp again: this is the
 	// predicate the agent-switch source stop runs, and the value it compares
 	// against is a later instant in UTC.
-	var comparable int
+	var comparableRows int
 	if err := db.QueryRow(
 		`SELECT COUNT(*) FROM sessions WHERE activity_last_at <= '2026-08-12 16:00:00.000000 +0000 UTC'`,
-	).Scan(&comparable); err != nil {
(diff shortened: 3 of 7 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

axisrow:
Thanks a lot for merging! 🙏 Appreciate the quick turnaround.

Issue #2333, opened 2026-07-01
Terminal rendering glitches and blank terminal after switching sessions


ApexYash11:
### Description

Was trying out AO locally to explore the codebase and ran into a couple of UI issues.

The terminal rendering occasionally gets a bit garbled/misaligned after using it for a while.

[terminal rendering issue]
<img width="1334" height="750" alt="Image" src="https://github.com/user-attachments/assets/f414f18f-8150-4491-9698-a0fca1726f2c" />

I also had the terminal go almost completely blank once after switching between sessions from two different repos. The input box was still there, but the rest of the terminal pane was just black.

[blank terminal]
<img width="1334" height="750" alt="Image" src="https://github.com/user-attachments/assets/9435616c-8db6-404c-a33b-73552cdf3a25" />

Vaibhaav-Tiwari:
please find the underlying cause for this and extend that within this same issue, after that a PR can be raised

ApexYash11:
Did some more digging into this and I think I found the likely root causes behind both issues.

1. Garbled / misaligned terminal rendering

The main issue appears to be the renderer fallback path in XtermTerminal.tsx.

When the WebglAddon loses its context at runtime, onContextLoss() only disposes the addon. It never falls back to the CanvasAddon, so xterm ends up using the DOM renderer, which doesn't correctly render box-drawing characters. That matches the garbled/misaligned TUI rendering I was seeing.

There's also a secondary issue in the resize stabilizer. If proposeDimensions() returns undefined while the terminal container is still settling (for example during a flex/CSS transition), the convergence loop exits without triggering another fit, leaving the terminal with an incorrect grid size until another resize event occurs.

2. Blank terminal after switching sessions

This looks like a timing/race issue during session switching.

While changing to a session from another workspace, terminalHandleId is briefly unavailable, causing TerminalPane to mount an AttachedTerminal with an "empty" handle. The terminal instance is created, but no mux connection is established, resulting in the blank terminal state.

There's also what looks like a double-attach race where attach() can be invoked twice during initialization. The first connection is torn down, terminal.clear() is called, and a second connection starts. That would explain the temporary black/empty terminal with only the input area visible.

From what I can tell, the affected areas are:

frontend/src/renderer/components/XtermTerminal.tsx
frontend/src/renderer/components/TerminalPane.tsx
frontend/src/renderer/hooks/useTerminalSession.ts
frontend/src/renderer/lib/terminal-mux.ts

[...]

PR #3901, opened 2026-08-12
fix(terminal): bound retained terminals and stop black-pane latches

axisrow:
Refs #2333, #3803, #3787.

Terminal panes go black and stay black: the backend session is healthy, tmux has the content, but the pane paints nothing. Switching away and back does not help; only restarting the app clears it — which is why it reads as intermittent.

Two independent causes, both addressed here.

### 1. The retained-terminal cache is unbounded

`MAX_RETAINED_TERMINALS = 6` was removed in 4986af7d0 ("fix: retain terminals without eviction"). Since then every visited session keeps a mounted xterm — with its WebGL context and its mux lease — for the entire lifetime of the renderer process. Past Chromium's per-page WebGL context cap, the browser forcibly loses the *oldest* contexts, so the panes that go black are the ones visited earliest, not the ones doing anything wrong.

This restores the six-entry LRU. Eviction deletes the cache entry rather than only removing the host node: dropping the entry drops its portal on the next render, which unmounts `XtermTerminal` and releases both the terminal instance and its mux lease. Removing the container alone would clean up the DOM node and leak the rest.

The currently-active terminal is never evicted, and eviction considers only parked entries.

### 2. Activation and replay had no failure path

The activation state machine (`parked → preparing → ready → revealed → visible`) holds the pane at `visibility: hidden` during `preparing`. Any throw or rejection between those states leaves the cover on permanently — a black pane over a perfectly healthy session. Five such latches are closed:

- `showLatestOutput` wrapped, so a scroll failure costs one imperfect redraw instead of the pane
- a bounded watchdog on `prepareForActivation` (plus try/catch around `scheduleStableFit`), so a fit that never settles cannot hold the cover forever
- rejection handlers on the three `.then()` chains that drive attach, cached activation, and replay paint

The `replayPaintPending` latch was found by the added test rather than by reading the code: the opaque cover was never lifted when the replay-paint promise rejected.

### Relationship to the WebGL-loss work

The context-loss fallback in xuelongmu/agent-orchestrator#209 addresses the *reaction* to a lost WebGL context. This change removes the *accumulation* that causes contexts to be lost in the first place. They compose; neither subsumes the other.

### Tests

`TerminalPane.test.tsx` covers retention up to the cap and eviction of the oldest parked entry beyond it, plus the two rejection paths. 39 tests pass on this branch.

Note that `src/renderer/components/settings/AgentModelCombobox.test.tsx` fails on this branch — it fails identically on an untouched `main`, so it is pre-existing and unrelated.

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.test.tsx
+++ b/frontend/src/renderer/components/TerminalPane.test.tsx
@@ -380,6 +380,31 @@ describe("TerminalPane replay cover", () => {
 		}
 	});
 
+	it("attaches and lifts the cover even when activation preparation rejects", async () => {
+		prepareForActivationMock.mockRejectedValue(new Error("viewport gone"));
+		replaySettled.value = false;
+		const view = renderPane({ ...worker, terminalHandleId: "term-1" });
+		try {
+			await waitFor(() => expect(attachMock).toHaveBeenCalled());
+			replaySettled.value = true;
+			view.rerender(
+				<QueryClientProvider client={view.queryClient}>
+					<TerminalPane
+						daemonReady
+						fontSize={12}
+						session={{ ...worker, terminalHandleId: "term-1" }}
+						theme="dark"
+					/>
+				</QueryClientProvider>,
+			);
+			await waitFor(() =>
+				expect(screen.queryByTestId("terminal-replay-cover")).not.toBeInTheDocument(),
+			);
+		} finally {
+			view.restore();
+		}
+	});
+
 	it("keeps the replay cover silent even when attachment takes longer", () => {
 		vi.useFakeTimers();
 		try {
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:

i-trytoohard:

axisrow:

axisrow:
Closing as superseded: the retained-terminals/black-pane fix is already in main — upstream evolved it further (parked-terminal layout skip and TUI-focus coverage in TerminalPane). This branch applies as empty on current main (patch-id verified).

Issue #2702, opened 2026-07-14
Runtime restore recreates the tmux session but never relaunches the agent — zombie session reports idle and holds its branch


fireddd:
## Summary

When the daemon restarts, it recreates a session's tmux session but **does not relaunch the agent inside it**. The result is a zombie: AO reports the session as alive and `idle`, but the tmux pane is a bare `zsh` with no agent running. The zombie keeps its git worktree and holds its branch checked out, which blocks any new spawn on that branch — while doing no work.

Because the daemon is now owned by the desktop app, this is easy to hit: **closing the app window kills the daemon, the tmux server, and every running agent.** On the next launch, the sessions come back as zombies.

## Steps to reproduce

1. `ao spawn --project <p> --branch <b> --prompt '...'` — confirm the agent is running (`ps aux | grep 'claude --session-id'`, tmux pane shows the agent).
2. Quit the desktop app (or otherwise stop the daemon).
3. Relaunch the app so the daemon comes back.
4. Look at the restored session.

## Observed

The tmux session is recreated, with a fresh timestamp:

```
$ tmux ls
fireddd-github-io-3-90ce67e1: 1 windows (created Wed Jul 15 00:12:28 2026)
```

But there is no agent in it:

```
$ tmux list-panes -t fireddd-github-io-3-90ce67e1 -F "#{pane_current_command}"
zsh                                    # <-- bare shell, not the agent

$ ps aux | grep -c "[c]laude --session-id"
0                                      # <-- agent is gone
```

AO nonetheless reports the session as healthy and non-terminated:

```
$ ao session get fireddd.github.io-3
id: fireddd.github.io-3
name: Verify the nav fix
role: worker
status: idle
activity: idle
harness: claude-code
terminated: false                      # <-- looks alive
```

And it still owns its branch, so a fresh spawn on that branch is refused:

```
$ ao spawn --project fireddd.github.io --branch ao/verify-nav --prompt '...'
spawn fireddd.github.io-4: workspace: workspace: branch is already checked out in another worktree:
  "ao/verify-nav" is checked out at ".../worktrees/fireddd.github.io/fireddd.github.io-3"
  (BRANCH_CHECKED_OUT_ELSEWHERE)
```

The only way forward is `ao session kill fireddd.github.io-3` to free the branch, and re-spawn from scratch — losing whatever the agent had in flight.

## Expected

One of:

[...]

PR #3902, opened 2026-08-12
fix(restore): fall back to a fresh launch when the native conversation is gone

axisrow:
A session whose provider never persisted a conversation cannot be restored: every attempt resolves to a native resume against an id the agent has never heard of, and fails with

```
No conversation found with session ID: <uuid>
```

on every retry. The work in the session's workspace — its branch and commits — is still there, but unreachable through the UI.

This happens when a session is torn down before its first turn, or when a switch away from it fails after the source has already stopped.

### Cause

`GetRestoreCommand` returns `ok=true` for any id it can produce, including a deterministically computed UUID that was reserved but never used. Having an id is treated as proof that a conversation exists behind it, so restore always chooses `--resume` and always fails.

### Fix

Ask the agent before trusting the id. `ports.AgentInterfaceHandoffHistoryProbe` already exists for exactly this distinction — it is implemented by both claude-code and codex — but was only consulted on interface transitions, never on the restore path. When it reports the conversation missing, relaunch into the same workspace instead.

Agents that cannot answer (no probe, or a probe that errors) are treated as *present*, so this never turns a restore that might work into one that refuses.

The promptless relaxation is deliberately narrow — `allowPromptless = conversationLost`, not an unconditional `true`. Reaching the point of reserving an id is not the same as having nothing to restore, but a worker that never got that far (no id **and** no prompt) must stay unresumable; two existing tests guard that case and a broader relaxation breaks them.

### Tests

`TestRestore_LostNativeConversationRelaunchesFresh` fails without the change with `restore mode = "native"` — the doomed path. `TestRestore_PromptlessWorkerNotResumable` and `TestRestore_AgyAndCopilotPromptlessWorkersWithoutAgentSessionIDNotResumable` still pass, confirming the narrower case is untouched.

Verified live: `POST /resume-agent` returned 200 with `"resumeMode":"saved_prompt"` and the agent came up in its own worktree with all commits intact.

### Adjacent, not fixed here

`POST /restore` returns 409 `SESSION_NOT_RESTORABLE` for an `exited` session with `is_terminated=0`, because it requires `IsTerminated=true` (#2702). `/resume-agent` is the correct path for that state and is what the UI button uses.

### DIFF
--- a/backend/internal/session_manager/manager.go
+++ b/backend/internal/session_manager/manager.go
@@ -1671,7 +1671,7 @@ func (m *Manager) relaunchSessionWithPolicy(ctx context.Context, operation strin
 			systemPrompt, systemPromptFile, agentConfig, rec.Kind, m.dataDir, true)
 	} else {
 		argv, delivery, mode, err = restoreArgv(ctx, agent, rec.ID, ws.Path, rec.Metadata,
-			systemPrompt, systemPromptFile, agentConfig, rec.Kind, rec.Harness, m.dataDir)
+			systemPrompt, systemPromptFile, agentConfig, rec.Kind, rec.Harness, m.dataDir, env)
 	}
 	if err != nil {
 		m.cleanupSystemPromptDir(rec.ID)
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

axisrow:
Thank you for merge

PR #3906, opened 2026-08-12
fix(terminal): bound replay write accounting so a stalled write cannot black out the pane

axisrow:
Refs #2333.

Stacked on #3901 — please review that one first; this PR's diff is the last commit only.

A terminal pane can go black and stay black over a completely healthy session: the backend has the content, the session is alive, but the pane never paints again. Only restarting the app clears it. This is the last latch I can find that produces that symptom, and it is the one that explains an observation the other fixes do not — pressing a key in tmux provably generated new bytes on the backend, and they never reached the pane.

### Cause

Every decrement of `pendingReplayWrites` lives inside a write callback:

```ts
terminal.write(bytes, () => {
    if (replayWritesPreserved) return;                          // returns BEFORE the decrement
    if (!isCurrentAttachment(generation, handle, mux)) return;   // and so does this
    postReplayWriteActive = false;
    pendingReplayWrites = Math.max(0, pendingReplayWrites - 1);
```

If the callback never arrives — a renderer addon that swallows it, or a synchronous throw on the way in — the counter stays positive for the life of the attachment. There is no deadline on it.

That single stuck counter is enough to kill the pane, because everything downstream is gated on it:

- every later PTY frame is pushed onto `postReplayWriteQueue` instead of being written to xterm;
- `drainPostReplayWrites` is itself gated on the same counter, so the queue is never drained;
- `revealReplayTail` and `settleAfterReplayWrites` are gated on it too, so the reveal cover is never lifted.

Healthy session, live bytes, dead pane.

### Changes

- Route the counter through `beginReplayWrite` / `endReplayWrite`, so no call site can raise it without arming the deadline that bounds it.
- On that deadline, abandon the outstanding writes: drain what is already queued straight into xterm, clear the tail timers, and lift the cover. Modeled on the existing `preservePendingReplayWrites`, which already does exactly this at teardown.
- Absorb synchronous throws at both write boundaries — the handle's `write` in `XtermTerminal`, and a local guard in the hook. The hook is handed its terminal by the owner and cannot assume the writer is guarded; letting a throw unwind aborts the replay mid-flight and strands the very accounting this change protects. `done` is called exactly once.

Recovery is deliberately silent. A false trip on a slow machine costs one imperfect redraw; no deadline at all costs the whole pane.

### Tests

Two regressions, both failing without the fix:

- a write callback that never arrives → after the deadline the cover is lifted and subsequent PTY bytes reach xterm instead of accumulating in the queue;
- a write that throws synchronously → the attachment still settles and keeps working.

Note `src/renderer/components/settings/AgentModelCombobox.test.tsx` fails on this branch; it fails identically on an untouched `main`, so it is pre-existing and unrelated.

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.test.tsx
+++ b/frontend/src/renderer/components/TerminalPane.test.tsx
@@ -469,6 +469,31 @@ describe("TerminalPane replay cover", () => {
 		}
 	});
 
+	it("attaches and lifts the cover even when activation preparation rejects", async () => {
+		prepareForActivationMock.mockRejectedValue(new Error("viewport gone"));
+		replaySettled.value = false;
+		const view = renderPane({ ...worker, terminalHandleId: "term-1" });
+		try {
+			await waitFor(() => expect(attachMock).toHaveBeenCalled());
+			replaySettled.value = true;
+			view.rerender(
+				<QueryClientProvider client={view.queryClient}>
+					<TerminalPane
+						daemonReady
+						fontSize={12}
+						session={{ ...worker, terminalHandleId: "term-1" }}
+						theme="dark"
+					/>
+				</QueryClientProvider>,
+			);
+			await waitFor(() =>
+				expect(screen.queryByTestId("terminal-replay-cover")).not.toBeInTheDocument(),
+			);
+		} finally {
+			view.restore();
+		}
+	});
+
 	it("keeps the replay cover silent even when attachment takes longer", () => {
 		vi.useFakeTimers();
 		try {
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:

i-trytoohard:

axisrow:

axisrow:

axisrow:
@illegalcall friendly bump

PR #3907, opened 2026-08-12
fix(terminal): defer xterm disposal and force the activation resize

axisrow:
Refs #2333, #3803, #3787.

Stacked on #3901 — please review that one first; this PR's diff is the last commit only.

Two independent hardening changes around terminal activation and teardown. Neither is the primary black-pane fix; both close windows that can produce the same visible symptom.

### 1. Deferred disposal — a port of #3804

**This is not my work.** #3804 is open and unmerged; I am carrying its fix here because it becomes materially more relevant after #3901. Happy to drop this half if #3804 lands first, or to close it in favor of that PR.

xterm 5.5 schedules `Viewport.syncScrollArea` with a zero-delay timer during `open()`. Disposing in that same task clears `RenderService`'s renderer first, so the pending callback throws while reading its dimensions. Detaching listeners and the transport stays synchronous; only the final `term.dispose()` moves behind one macrotask, and every AO listener and attachment is already detached by then, so the terminal is inert for that tick.

Why it matters more now: #3901 restores an LRU cap on retained terminals, which means eviction disposes terminals far more often than the previous unbounded retention ever did. A race that was rare under "never dispose anything" is no longer rare.

### 2. Force the activation resize

The activation promote could never reach the daemon in the most common case.

A parked terminal reattaches at 0×0 and is then refitted back to the size it already had, so the grid being republished is usually byte-identical to `lastPublishedGrid`. `syncVisibleSize` returned early on exactly that equality, and even without the early return the daemon deduplicates repeat resizes — there is an explicit test for that, `TestServeDeduplicatesResizeUnlessExplicitlyForced`. So tmux was never asked to redraw, and the pane could stay blank.

The `force` flag already existed in the mux protocol and the backend already honored it. The renderer simply never sent it. Now this one activation path sends `force: true`; ordinary resizes stay deduplicated, preserving the behavior that backend test guards.

### Tests

The forced-resize regression fails without the change with an empty resize list — the daemon received nothing at all.

Note `src/renderer/components/settings/AgentModelCombobox.test.tsx` fails on this branch; it fails identically on an untouched `main`, so it is pre-existing and unrelated.

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.test.tsx
+++ b/frontend/src/renderer/components/TerminalPane.test.tsx
@@ -469,6 +469,31 @@ describe("TerminalPane replay cover", () => {
 		}
 	});
 
+	it("attaches and lifts the cover even when activation preparation rejects", async () => {
+		prepareForActivationMock.mockRejectedValue(new Error("viewport gone"));
+		replaySettled.value = false;
+		const view = renderPane({ ...worker, terminalHandleId: "term-1" });
+		try {
+			await waitFor(() => expect(attachMock).toHaveBeenCalled());
+			replaySettled.value = true;
+			view.rerender(
+				<QueryClientProvider client={view.queryClient}>
+					<TerminalPane
+						daemonReady
+						fontSize={12}
+						session={{ ...worker, terminalHandleId: "term-1" }}
+						theme="dark"
+					/>
+				</QueryClientProvider>,
+			);
+			await waitFor(() =>
+				expect(screen.queryByTestId("terminal-replay-cover")).not.toBeInTheDocument(),
+			);
+		} finally {
+			view.restore();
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:

i-trytoohard:

axisrow:

axisrow:
Closing as superseded: the deferred-xterm-disposal change is already in main (upstream evolved it further, incl. the parked/TUI-focus tests in TerminalPane) — this branch applies as empty on current main (patch-id verified).

PR #3908, opened 2026-08-12
fix: hide Codex app-server window on Windows

RPKWolf:
## Summary
- prevent AO's stdio-only `codex app-server` child from allocating/activating a Windows console
- keep process ownership with the AO chat conversation (no detach and no broad process killing)
- add a Windows-only regression test for `CREATE_NO_WINDOW` and `HideWindow`

## Root cause and evidence
A controlled AOS-Bridge -> AO Chat REST task completed with `BRIDGE_OK`, then left `codex.exe app-server` and a visible Windows Terminal host alive. The live chain was AO daemon PID 8976 -> Codex app-server PID 13940; Windows created OpenConsole PID 33580 and WindowsTerminal PID 21792 with a nonzero window handle and the Codex path as its title. `spawnAppServer` used raw `exec.Command`, unlike AO's other Windows launchers.

The app-server is intentionally session-owned for follow-up turns; this change fixes the visible console without treating it as an unowned process or killing unrelated tasks/runtimes.

## Tests
- `git diff --check` (pass)
- Windows regression test added: `TestConfigureAppServerProcessHidesConsoleWindow`
- AOS-Bridge full suite: 29/29 pass; TypeScript typecheck pass
- Go tests not run locally: the reproduction host does not have a Go toolchain installed; CI requested

## Safety
No Bridge code or AOS/IBKR runtime policy changed. No process enumeration or name-based killing is introduced. Parallel sessions and existing session-scoped cleanup remain unchanged.


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

Rishet11:
This fix is now redundant: #4013 routed the Codex app-server launch through the shared hidden-process launcher, so the console window no longer appears on Windows. Closing to keep the queue focused. Thanks for the clean catch.

PR #3910, opened 2026-08-12
fix(desktop): stop Cmd+Alt+I from crashing the main process on macOS

### LINKED ISSUE #4308: bug(desktop): macOS Toggle Developer Tools crashes the Electron main process
## Bug

On macOS, invoking the native **View → Toggle Developer Tools** command (or its `Cmd+Option+I` accelerator) can throw an uncaught exception in Electron's main process instead of opening DevTools.

The reporter sees the standard “A JavaScript error occurred in the main process” dialog with:

```text
Uncaught Exception:
TypeError: Cannot read properties of undefined (reading 'toggleDevTools')
    at webContentsMethod (node:electron/js2c/browser_init:2:31239)
    at Object.execute (node:electron/js2c/browser_init:2:28651)
    at MenuItem.click (node:electron/js2c/browser_init:2:35249)
    at a._executeCommand (node:electron/js2c/browser_init:2:40710)
```

macOS main-process error after toggling DevTools

**Source:** Direct Codex report on 2026-08-23 (no public chat URL)  
**Reported by:** @illegalcall  
**First observed:** 2026-08-23 on a nightly build; exact regression onset is not established  
**Analyzed against:** `c86b58d04a38ce669be3215ab0834e29cf2c0d30` (`origin/main`)  
**AO desktop version:** `0.12.7-nightly.202608181938` (bundled daemon reports `dev`, healthy on port `3001`)  
**Environment:** macOS 26.5.1 (25F80), Apple Silicon (`arm64`)  
**Priority:** P2 — the action crashes the Electron main process, but the scope is bounded to the native DevTools command and an in-app DevTools path is available  
**Confidence:** High — the stack, platform menu setup, and exact failing mechanism have been traced; an existing tested PR implements the same fix

## Reproduction

1. Launch Agent Orchestrator on macOS.
2. Focus a surface backed by a `WebContentsView` (the Browser panel is the common case).
3. Choose **View → Toggle Developer Tools** from the native macOS application menu, or press `Cmd+Option+I`.
4. Observe the uncaught `toggleDevTools` exception and main-process error dialog.

## Expected

DevTools opens for the focused AO surface, or the shell surface as a safe fallback, without terminating the main process.

## Actual

Electron's native menu command attempts to call `toggleDevTools` through an undefined focused-window target and throws an uncaught exception in the main process.

## Root Cause

`frontend/src/main.ts` installs AO's guarded application menu only on Windows. On macOS, Electron's default application menu remains active, including the built-in `role: "toggleDevTools"` command.

AO composes its window from a `BaseWindow` and `WebContentsView` surfaces. When focus belongs to a `WebContentsView`, Electron's built-in role resolves no focused `BrowserWindow`, then its internal `webContentsMethod` path attempts to access `toggleDevTools` on an undefined target. This matches the `MenuItem.click` stack in the report.

AO's custom titlebar IPC handler and Windows menu already route DevTools through guarded AO logic; the native default macOS menu is the unguarded path.

## Suggested Fix

Install an explicit macOS application menu and replace only the built-in DevTools role with a click handler that:

1. toggles DevTools for the last-focused AO browser surface; and
2. falls back to the shell `webContents` when no browser surface is available.

Keep the standard macOS app roles (About, Services, Hide, Hide Others, Show All, Quit) when replacing Electron's default menu.

## Impact

- A standard macOS menu item and accelerator can crash the Electron main process.
- No data loss was observed in this report.
- Workaround: avoid the native menu/accelerator and use AO's in-app DevTools control where available.

## Related

- [PR #3910](https://github.com/Untrivial-ai/agent-orchestrator/pull/3910) — open, tested implementation of the traced fix; awaiting review and currently has no closing issue reference.

axisrow:
Pressing the DevTools shortcut on macOS takes down the main process:

```
TypeError: Cannot read properties of undefined (reading 'toggleDevTools')
  at webContentsMethod (node:electron/js2c/browser_init)
  at Object.execute / MenuItem.click / a._executeCommand
```

The user sees the "A JavaScript error occurred in the main process" dialog and the app is gone.

### Cause

The trace runs through `MenuItem.click`, so the failure is in Electron's **default** menu, not in any AO handler. `buildWindowsAppMenu` is gated on `win32`, so macOS never installed an application menu and the default one stayed active — including its View → Toggle Developer Tools entry, which uses the built-in `role: "toggleDevTools"`.

That role resolves its target as `focusedWindow.webContents`. In AO focus normally sits on a `WebContentsView` (the Browser panel) rather than on the `BrowserWindow`, so the role reads `webContents` off `undefined`.

Worth noting that AO's own three DevTools paths were already guarded against exactly this — the toolbar shortcut, the `menu:action` IPC handler, and the Windows menu item. Only the default menu was outside that guard, which is why the crash looked arbitrary: it depended on where focus happened to be.

### Fix

Install a macOS application menu so the crashing role is never used. Every other entry stays a role; DevTools routes through the same guarded toggle the other three paths already use, extracted here as `toggleDevToolsForFocusedSurface` and shared with the Windows menu.

The app submenu (`about`/`services`/`hide`/`hideOthers`/`unhide`/`quit`) is spelled out deliberately: replacing the default menu also replaces the standard items, and silently losing Cmd+Q would be a worse regression than the bug being fixed. There is a test pinning those roles for that reason.

### Verification

Beyond the unit tests, I confirmed against the built bundle that the shipped macOS View menu carries `{label: "Toggle Developer Tools", accelerator: "Alt+Command+I", click: …}` with no `role`, and that `setApplicationMenu` runs on the `darwin` branch. Cmd+Opt+I now opens DevTools for the focused surface instead of terminating the main process.

Closes #4308

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -85,7 +85,7 @@ import { keepDaemonAlive, shouldLinkOnAttach } from "./main/daemon-owner";
 import { readMigrationState, updateMigration, writeAppStateMarker, type MigrationState } from "./main/app-state";
 import { isAllowedAppExternalURL, openAllowedAppExternalURL } from "./main/external-open";
 import { shouldSignalAttention, shouldToast } from "./main/notification-signals";
-import { buildWindowsAppMenuTemplate } from "./main/menu";
+import { buildMacAppMenuTemplate, buildWindowsAppMenuTemplate } from "./main/menu";
 import { ancestorRepositorySetupWarning, scanImportFolder } from "./main/import-folder-scan";
 
 // Globals injected at compile time by @electron-forge/plugin-vite.
@@ -314,14 +314,29 @@ function appendDaemonOutput(text: string): void {
 // out of sight, but the roles keep their accelerators alive (Reload, zoom, full
 // screen, edit commands). DevTools uses the AO browser toggle so the focused
 // Browser panel opens the same native Chromium surface as the toolbar.
+// Open DevTools for whatever is actually focused. Electron's built-in
+// toggleDevTools role assumes focus is on the BrowserWindow; in AO it is usually
+// on a WebContentsView, where that role crashes the main process. Everything
+// that toggles DevTools goes through here instead.
+function toggleDevToolsForFocusedSurface(): void {
+	const fallback = () => getShellWebContents()?.toggleDevTools();
+	const host = browserViewHost;
+	if (!host) {
+		fallback();
+		return;
+	}
+	void host.toggleDevToolsForLastFocused().then((state) => {
+		if (!state) fallback();
+	}).catch(fallback);
+}
+
 function buildWindowsAppMenu(): Menu {
+	return Menu.buildFromTemplate(buildWindowsAppMenuTemplate(toggleDevToolsForFocusedSurface));
+}
+
+function buildMacAppMenu(): Menu {
 	return Menu.buildFromTemplate(
-		buildWindowsAppMenuTemplate(() => {
-			const fallback = () => getShellWebContents()?.toggleDevTools();
-			void browserViewHost?.toggleDevToolsForLastFocused().then((state) => {
-				if (!state) fallback();
-			}).catch(fallback);
-		}),
+		buildMacAppMenuTemplate(app.name, toggleDevToolsForFocusedSurface),
 	);
 }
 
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
A fresh macOS nightly report reproduces this exact `toggleDevTools` main-process crash. I filed [#4308](https://github.com/Untrivial-ai/agent-orchestrator/issues/4308) as the canonical tracking issue with the reporter screenshot, environment, traced cause, and reproduction steps. This PR appears to address it; please link it as the closing issue when merged.

axisrow:
@illegalcall Linked #4308 as the closing issue — thanks for filing the canonical tracking issue with the repro.

axisrow:
Superseded — #4314 landed the same fix (onToggleDevTools callback replacing Electron's built-in role) and is already merged into main. Closing this in favor of it.

PR #3915, opened 2026-08-12
fix(desktop): label the default reviewer with its display name

axisrow:
## Problem

#3959 landed dynamic default-reviewer resolution (`resolveDefaultReviewerHarness`), but the label-facing strings still print the raw harness id instead of the catalog's display name:

- The `__default__` menu entry (`defaultOptionLabel={resolvedDefaultHarness}`) shows `codex` instead of `Codex`.
- The "Review in progress · ${activeReviewerHarness}" strip shows the same raw id.

Now that the default is resolved dynamically rather than hardcoded to `claude-code`, this is more visible: the same agent can read as two different names within the same menu (raw id for the default entry, display name for every other catalog entry).

## Change

Builds on #3959 — reuses its harness-resolution logic unchanged and only wraps the label-facing strings in `agentLabel()`:

- `ReviewerSelect.tsx`: fallback agent entries now label with `agentLabel(id)` instead of the bare id.
- `SessionInspector.tsx`: `defaultOptionLabel` and the "Review in progress" strip now go through `agentLabel()`.

## Testing

- `npx vitest run src/renderer/components/SessionInspector.test.tsx` — 84 passed.
- `npx tsc --noEmit` — clean (pre-existing unrelated `motion/react` module error in `packages/product-ui` aside).

### DIFF
--- a/frontend/src/renderer/components/ReviewerSelect.tsx
+++ b/frontend/src/renderer/components/ReviewerSelect.tsx
@@ -1,4 +1,5 @@
 import type { components } from "../../api/schema";
+import { agentLabel } from "../lib/agent-options";
 import { buildRankedAgentOptions } from "../lib/agent-select-options";
 import { KNOWN_REVIEWER_HARNESS_IDS } from "../lib/reviewer-harnesses";
 import { AgentAvatar } from "./AgentAvatar";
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

PR #3917, opened 2026-08-12
fix(frontend): skip slide-in spring for sidebar peek overlay

### LINKED ISSUE #3372: Sidebar hover-preview vanishes before it can be used (CSS transition / close-delay race)
## Bug

Hovering the sidebar toggle to peek the project sidebar works — the sidebar starts sliding in — but it disappears before the cursor can reach it, making the hover-preview feature completely unusable.

**Reported by:** @somewherelostt  (Windows 11)
**Screenshot:** shows the sidebar briefly visible but it closes on cursor approach.

### Steps to reproduce

1. Collapse the sidebar (Ctrl+B or click toggle).
2. Hover the sidebar toggle button (PanelLeft icon in the window titlebar on Windows; top-left cluster on macOS/Linux).
3. Sidebar starts sliding in as an overlay.
4. Move cursor toward the sidebar to interact.
5. Sidebar vanishes mid-approach.

### Expected

Once the peek sidebar appears, it should remain open while the cursor is over the sidebar. It should only close after the cursor moves completely outside both the trigger and the sidebar.

### Actual

The sidebar closes while the cursor is still moving toward it, often before it finishes animating in.

### Root cause

**Race condition between the 200 ms CSS slide-in transition and the 140 ms auto-close delay.**

The peek-open logic lives in `routes/_shell.tsx`. When the sidebar is in peek mode, a global `pointermove` listener checks whether the cursor is inside the sidebar:

```tsx
// _shell.tsx:460-485
useEffect(() => {
    if (!isSidebarPeekOpen || isSidebarOpen) return;
    const handlePointerMove = (event: PointerEvent) => {
        const sidebar = document.querySelector('[data-slot="sidebar-container"]');
        const bounds = sidebar?.getBoundingClientRect();  // returns ANIMATED position
        const isInSidebar = bounds && event.clientX >= bounds.left && ...;
        if (isInSidebar) cancelSidebarPeekClose();
        else scheduleSidebarPeekClose();  // 140 ms timer
    };
    window.addEventListener("pointermove", handlePointerMove);
    ...
}, [...]);
```

The sidebar container (`ui/sidebar.tsx:217`) has `transition-[left,right,width] duration-200`. When the sidebar opens from the collapsed (off-canvas) state, `left` animates from `calc(var(--sidebar-width) * -1)` to `0` over **200 ms**.

During that transition, `getBoundingClientRect()` returns the **mid-animation** position — not the final resting position. The sidebar's `right` edge sweeps from 0 to sidebar-width over 200 ms. If the cursor is anywhere right of the animating edge, `isInSidebar` is false, and `scheduleSidebarPeekClose()` arms a **140 ms** timer. Each subsequent `pointermove` that is still outside resets the timer. Once the cursor stops (to interact), no more `pointermove` events fire, and **140 ms is less than 200 ms** — the close timer fires before the transition finishes.

**Why it is worse on Windows:** The toggle button lives in the 36 px window titlebar (`WindowTitlebar.tsx`), separated from the sidebar body by the full titlebar height. The cursor must travel farther (down past the titlebar) to reach the sidebar, so it spends more time in the zone where the animating sidebar rect has not yet reached it.

### Key files

| File | Role |
|---|---|
| `frontend/src/renderer/routes/_shell.tsx:183-196` | `previewSidebar` / `scheduleSidebarPeekClose` (140 ms close delay) |
| `frontend/src/renderer/routes/_shell.tsx:460-485` | Global `pointermove` handler using `getBoundingClientRect()` |
| `frontend/src/renderer/components/ui/sidebar.tsx:217` | `transition-[left,right,width] duration-200` on the sidebar container |
| `frontend/src/renderer/components/Sidebar.tsx:207-222` | Overlay positioning (`top-0 h-svh!` + `z-sidebar-preview`) |
| `frontend/src/renderer/components/WindowTitlebar.tsx:109-118` | Windows toggle button with `onPointerEnter` trigger |
| `frontend/src/styles/tokens.css:167-170` | z-index layers (chrome=10, sidebar-preview=15, titlebar=20) |

### Suggested fixes (any one)

1. **Match delays:** Make the close delay at least the transition duration (e.g., bump from 140 ms to 250 ms). Simplest, but adds perceptible lag.
2. **Use final bounds, not animated bounds:** Compute the target rect from CSS variables / known width instead of `getBoundingClientRect()` during the transition. Or cache the resting rect at open time and use it until `transitionend`.
3. **Suppress close during transition:** Ignore `pointermove`-driven `scheduleSidebarPeekClose` calls for the first ~200 ms after opening (guard with a timestamp or `transitionend` listener).
4. **Disable transition for peek overlay:** Since the peek is a temporary overlay, skip the slide-in animation when `isOverlay` is true (set `transition: none`). The sidebar appears instantly, so `getBoundingClientRect()` is correct from t=0.

Option **4** is likely the cleanest — the peek overlay does not need the slide animation, and removing it eliminates the race entirely.

### Workaround

Click the toggle button (or press Ctrl+B) to pin the sidebar open instead of relying on hover-preview.

ronishrohan:
## Summary
- Fixes #3372: the sidebar hover-preview vanished before the cursor could reach it.
- Root cause: while the peek overlay slides in via a Motion spring transition, the global `pointermove` handler in `_shell.tsx` reads `getBoundingClientRect()` on the sidebar container to decide whether to keep it open. During the animation that rect reflects the mid-transition position, not the resting one, so the 140ms auto-close timer can fire before the slide-in settles.
- Fix: when the sidebar renders as a peek overlay (`overlay` prop), skip the spring transition so it appears at its resting position instantly — matching the "cleanest" option suggested in the issue. This removes the race entirely since `getBoundingClientRect()` is correct from t=0.

## Test plan
- [x] `npm run typecheck` — no new errors (pre-existing unrelated errors in `packages/product-ui`, unrelated to this change)
- [x] `npm test` (vitest) — `Sidebar.test.tsx` passes (54/54); remaining failures are pre-existing/unrelated (landing-page tests missing the `cheerio` dependency)
- [ ] Manual verification of hover-preview in the desktop app (not run in this environment)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/frontend/src/renderer/components/ui/sidebar.tsx
+++ b/frontend/src/renderer/components/ui/sidebar.tsx
@@ -208,8 +208,13 @@ function Sidebar({
 	const isOffcanvasCollapsed = state === "collapsed" && collapsible === "offcanvas";
 	const isIconCollapsed = state === "collapsed" && collapsible === "icon";
 	const containerX = isOffcanvasCollapsed ? (side === "left" ? "-100%" : "100%") : "0%";
+	// Peek overlays skip the slide-in spring entirely: the hover-preview close
+	// timer in _shell.tsx reads getBoundingClientRect() on every pointermove,
+	// and a mid-animation rect made the preview vanish before the cursor could
+	// reach it (#3372). Overlays are temporary, so appearing instantly reads as
+	// intentional rather than janky.
 	const activeTransition: typeof SHELL_PANEL_SPRING | { duration: number } =
-		!isReady || prefersReducedMotion ? { duration: 0 } : SHELL_PANEL_SPRING;
+		!isReady || prefersReducedMotion || overlay ? { duration: 0 } : SHELL_PANEL_SPRING;
 
 	// Target width for the gap placeholder. Animating the actual width lets the
 	// flex sibling <main> follow in real time instead of snapping separately.


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #3924, opened 2026-08-13
docs: make telemetry guide user-facing

AgentWrapper:
## Summary

- replace the internal telemetry audit/runbook prose with a concise user-facing guide
- explain collection defaults, pseudonymous identifiers, PostHog metadata, sensitive-data exclusions, and local retention in plain language
- document the complete desktop/daemon opt-out controls, including their limits and the current lack of a mobile in-app opt-out
- update the docs index to describe the page as user-facing

## Verification

- `git diff --check`
- verified every external link returns HTTP 200
- checked the page against the current desktop, daemon, and mobile telemetry implementations
- completed two fresh-reader comprehension reviews and incorporated the identified ambiguity fixes

## Risk

Documentation-only. No telemetry behavior changes.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -63,8 +63,8 @@ an email address there, it is used to manage that waitlist as described in the
   Hashing hides the plain text but still allows related events to be grouped.
 - Absolute local paths and local application URLs detected in desktop events
   are replaced with redaction markers before the events are sent.
-- Daemon and mobile events accept a fixed set of properties; unexpected fields
-  are discarded.
+- Daemon events sent to PostHog and mobile events accept a fixed set of
+  properties; unexpected fields are discarded.
 - Event rates are limited to reduce repeated background activity and error
   loops.
 - Person profiles and session recording are disabled in the desktop and mobile
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -1,276 +1,113 @@
-# Telemetry
+# Product telemetry
 
-AO uses anonymous telemetry to understand reliability and product usage. The
-Electron renderer sends sanitized PostHog events directly, and the Go daemon can
-persist allowlisted events locally and fan them out to PostHog when remote
-telemetry is enabled.
+AO collects limited product-usage and reliability data to learn which parts of
+the app are useful and whether releases are working as expected. The data is
+not account-linked: AO does not attach a name or email address. It is
+pseudonymous rather than unlinkable, because a random installation identifier
+lets events from the same installation be counted together over time.
 
-For cost-control runbooks, including the v2 PostHog event namespace and legacy
-ingestion drop rules, see [posthog-cost-controls.md](posthog-cost-controls.md).
+Remote telemetry is enabled in production desktop and mobile releases. A
+packaged desktop release also enables telemetry for the daemon it starts.
+Development builds and daemons started directly do not send remote telemetry by
+default.
 
-## What is collected
+## What AO sends
 
-- App activation events: `ao.app.active` / `ao.v2.app.active` from the
-  renderer and meaningful user-context CLI commands, each capped to one event
-  per six-hour UTC slot, or four per day per install/channel
-- Renderer load and daily route-surface usage, grouped by coarse surface names
-- Project/task/session UI actions, with project identifiers SHA-256 hashed
-- Renderer exceptions, reduced to error name and coarse context
-- Daemon operational events: CLI invocation, session spawn/failure, waiting-input
-  transitions, HTTP 5xx, and daemon panics
-- Code review outcomes: `ao.review.triggered`, `ao.review.submitted`,
-  `ao.review.cancelled`, and `ao.review.trigger_failed`. These carry the reviewer
-  `harness`, the `verdict` (`approved` / `changes_requested`), how long the pass
-  took, whether the review reached the provider, and a coarse `error_kind` on
-  failure. The review body is never sent: it is reviewer prose about a user's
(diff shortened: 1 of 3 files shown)


docs/telemetry.md
```diff
@@ -1,276 +1,113 @@
...
-renderer and daemon both use this ID as the PostHog distinct ID so activity is
-deduplicated across app launches and CLI invocations. It is not linked to any
-personal account. In the renderer it is also the PostHog device ID, and the SDK
-is explicitly kept in anonymous mode.
-
-## Configuration
-
-Renderer PostHog key and host are baked in at build time. To point a build at
-another PostHog project, set these environment variables before building:
+# Product telemetry
```
AgentWrapper, line 1: Removing the kill-switch section here orphans the reference in `docs/posthog-cost-controls.md` (runbook step 5: "See the kill-switch section in [telemetry.md](telemetry.md)") and deletes the only documentation of `AO_TELEMETRY_DISABLED_EVENTS` semantics. Please relocate that material to `posthog-cost-controls.md` (and fix its link) as part of this PR.

docs/telemetry.md
```diff
@@ -1,276 +1,113 @@
...
+## How AO limits the data
+
+- AO generates a random installation identifier on first run. It is stored at
+  `~/.ao/data/telemetry_install_id` (or under `AO_DATA_DIR`) and is not linked to
+  a personal account.
+- Project and session identifiers included in telemetry are one-way hashed.
+  Hashing hides the plain text but still allows related events to be grouped.
+- Absolute local paths and local application URLs detected in desktop events
+  are replaced with redaction markers before the events are sent.
+- Daemon and mobile events accept a fixed set of properties; unexpected fields
```
AgentWrapper: Minor precision nit: the fixed-property allowlist for daemon events applies to the remote PostHog export path; the local SQLite copy stores the full raw payload (as the old doc noted). Since this bullet sits under "How AO limits the data" a reader may assume local records are also filtered. Consider "Daemon events sent remotely and mobile events accept a fixed set of properties..." or similar.

AgentWrapper (commented):
**Verdict: requesting changes** (one docs-consistency issue; the rewrite itself is accurate).

I verified the new user-facing claims against the current implementation:

- **Opt-out controls are correct.** `AO_TELEMETRY_RENDERER=off` disables renderer export even on packaged builds (`frontend/src/shared/telemetry.ts`), the supervisor only stamps `AO_TELEMETRY_EVENTS`/`AO_TELEMETRY_REMOTE` defaults when the environment does not already provide values (`frontend/src/main.ts`, `telemetryOverrides`), and the daemon accepts `off` for both (`backend/internal/config/config.go`). The caveat that the variables must reach the app process (Finder/Dock launches) is a good addition.
- **The crash-data claim is accurate.** Renderer exceptions are sent with message and stack trace, and the `before_send` sanitizer redacts local paths in both the message and stack-frame filenames (`frontend/src/renderer/lib/telemetry.ts`, covered by tests).
- **The 30-day retention wording matches the code exactly.** Pruning piggybacks on event persistence in `backend/internal/adapters/telemetry/localsqlite.go`, so "while local telemetry is active" and "resumes if daemon event capture is enabled again" are both precisely right.
- Defaults (dev builds and directly started daemons send no remote telemetry) also match.

**The issue:** this PR removes the `AO_TELEMETRY_DISABLED_EVENTS` kill-switch section, but `docs/posthog-cost-controls.md` (step 5 of the runbook, around line 175) still says "See the kill-switch section in [telemetry.md](telemetry.md)". After this PR that section no longer exists, and the switch semantics (`*` prefix matching, case-insensitive, internal or exported names accepted, applied outermost on the remote chain, local SQLite unaffected, unknown entries inert) are no longer documented anywhere. Since the cost-controls runbook is the operator doc that depends on it, please move the kill-switch material there (or link to wherever it lands) in this PR so the runbook does not dangle.

One optional precision nit is inline. Everything else reads well and is a clear improvement for end users.

docs/telemetry.md
AgentWrapper: Addressed in `892c5e852`. I moved the complete kill-switch semantics into a new **AO event kill switch** subsection in `docs/posthog-cost-controls.md` and replaced the stale cross-file link with the local `#ao-event-kill-switch` anchor. The runbook now documents matching, producer scope, restart/environment behavior, pipeline order, SQLite behavior, inert entries, and the separate mobile build controls.

docs/telemetry.md
AgentWrapper: Addressed in `892c5e852`. The bullet now says “Daemon events sent to PostHog and mobile events,” which explicitly scopes the fixed-property restriction to remote daemon export and leaves the following paragraph’s raw local SQLite behavior distinct.

AgentWrapper:
Addressed review 4942229476 in commit `892c5e852`:

- Moved the complete `AO_TELEMETRY_DISABLED_EVENTS` contract into `docs/posthog-cost-controls.md` and replaced the dangling link with an internal runbook anchor. The subsection now covers exact/prefix and case-insensitive matching, internal/exported names, per-install environment and restart behavior, both desktop producers, mobile build controls, pipeline order, raw local SQLite behavior, and inert entries.
- Narrowed the public telemetry bullet to “Daemon events sent to PostHog” so it no longer implies that raw local SQLite payloads use the remote allowlist.

Verification: `git diff --check`, internal anchor/stale-link check, backend telemetry and daemon tests, and a fresh-reader consistency pass. The targeted frontend test command could not run because this worktree has no installed frontend dependencies (`vitest` is unavailable); no runtime code changed.

PR #3926, opened 2026-08-13
fix(chat): unblock Claude interface switch on Windows

### LINKED ISSUE #3925: Interface switch fails on Windows: spawn EINVAL from claude.cmd via ACP session/new
## Bug

Interface switch (TUI → Chat UI) fails on Windows with `spawn EINVAL` from the ACP `session/new` call, leaving the session in an unrecoverable broken state that cascades into "Session not found" and "Internal server error (INTERNAL_ERROR)".

**Source:** Discord `#bug-triage` thread | **Reported by:** Windows user (via @prateek) | **Analyzed against:** `461a6df56` (origin/main, 2025-08-13)
**Confidence:** High — traced exact code path from frontend trigger through ACP spawn to JSON-RPC error
**Priority:** High — core feature (interface switch) is completely broken on Windows with no workaround

## Symptoms (3 cascading errors)

1. **"Interface switch failed" popup:**
   ```
   switch interface visualflow-2: resume chat: ACP session/new:
   {"code":-32603,"message":"Internal error","data":{"details":"spawn EINVAL"}}
   ```
2. **"Session not found. It may have been cleaned up — pick another from the sidebar."**
3. **"Internal server error (INTERNAL_ERROR)"**
4. **Terminal:** `Error: Session ID 75cd5794-e4d0-530f-bbbf-b06b126dfb98 is already in use`

All appear after clicking "Switch to chat UI" in an agent session on Windows.

## Reproduction

1. Windows host with Claude Code installed via npm (produces `claude.cmd` in the npm global bin)
2. AO desktop app running, a TUI (terminal) session is active
3. Click "Switch to Chat UI"
4. All four errors above appear; session becomes unusable

### Primary: `spawn EINVAL` — Node.js cannot spawn `.cmd` files without `shell: true`

The full call chain:

1. **Frontend** triggers `POST /api/v1/sessions/{id}/interface-transition` → `Manager.StartInterfaceTransition` (`interface_transition.go:115`)
2. Background goroutine `runInterfaceTransition` drains source, stops the TUI controller, CAS the session mode to `chat`, then calls `startTransitionTarget` → `relaunchSessionFresh` → `resumeChatController` (`chat_spawn.go:265`)
3. `resumeChatController` → `chat.Service.Start` (`service/chat/service.go:208`) → `driver.Start` (ACP `session/new`)
4. The ACP driver calls `d.cfg.Launch(ctx, cfg)` which resolves to **`claudeacp/driver.go:67`** — this returns `Command: node`, `Args: [claude-agent-acp/dist/index.js]`
5. `spawnAgent` (`acp/process.go:22`) spawns the Node.js `claude-agent-acp` bridge process with `CLAUDE_CODE_EXECUTABLE` set to the Claude binary path
6. The Claude binary is resolved by `binaryutil.ResolveBinary` using `claudeBinarySpec` (`claudecode.go:478`), whose `WinNames` list is **`["claude.cmd", "claude.exe", "claude"]`** — `claude.cmd` is searched **first**
7. On a Windows machine with npm-installed Claude Code, `exec.LookPath("claude.cmd")` finds the npm shim at `%APPDATA%\npm\claude.cmd`
8. `CLAUDE_CODE_EXECUTABLE` is set to `...\claude.cmd`
9. Inside the Node.js `claude-agent-acp` process, it tries to `child_process.spawn(claude.cmd, args)` **without `shell: true`**
10. **Node.js rejects this with `EINVAL`** — since Node.js 18.20+ (CVE-2024-27980), spawning `.cmd`/`.bat` files without `shell: true` is explicitly blocked as a security measure
11. The spawn error propagates as ACP JSON-RPC `-32603` (`{"details": "spawn EINVAL"}`)
12. `acp/driver.go:144` wraps it: `normalizeACPError("ACP session/new", err)`
13. `chat_spawn.go:325` wraps: `"switch interface visualflow-2: resume chat: ACP session/new: ..."`
14. Interface transition fails with code `TARGET_RESUME_FAILED`

### Secondary: Cascading session corruption

After the `spawn EINVAL` failure in `startTransitionTarget`:

- The interface transition coordinator has **already** stopped the source TUI controller and committed the session mode to `chat` (via `CommitControllerEpoch` at `interface_transition.go:328`)
- The rollback path (`rollbackInterfaceTransition` at line 700) tries to restore the original mode and relaunch the TUI controller
- But the ConPTY runtime's in-memory registry still holds the session ID (`conpty/runtime.go:87-89`), so `Create` returns `"session ... already exists; destroy before re-creating"`
- The session is now in a **limbo state**: mode = `chat` (no working chat controller), TUI controller destroyed (can't recreate), frontend can't render either surface
- "Session not found" = the chat controller doesn't exist, so the session view has nothing to show
- "Internal server error (INTERNAL_ERROR)" = unmapped service-layer error from the broken session state, collapsed by the generic `toAPIError` default branch (see issue [#2598](https://github.com/Untrivial-ai/agent-orchestrator/issues/2598) for the envelope-collapse bug class)

### Key files

| File | Role |
|---|---|
| `backend/internal/adapters/agent/claudecode/claudecode.go:481` | `WinNames: ["claude.cmd", "claude.exe", "claude"]` — `.cmd` preferred |
| `backend/internal/adapters/chatdriver/claudeacp/driver.go:82` | Sets `CLAUDE_CODE_EXECUTABLE` to the resolved `.cmd` path |
| `backend/internal/adapters/chatdriver/acp/process.go:22` | `spawnAgent` spawns Node.js bridge |
| `backend/internal/adapters/chatdriver/acp/driver.go:136-144` | `session/new` fails, wraps `spawn EINVAL` |
| `backend/internal/session_manager/interface_transition.go:342` | `startTransitionTarget` → the failure point |
| `backend/internal/adapters/runtime/conpty/runtime.go:87-89` | `"session ... already exists"` on rollback Create |

### Fix 1 (primary): Resolve `claude.exe` before `claude.cmd` on Windows

In `claudecode.go:481`, swap the `WinNames` order:
```go
WinNames: []string{"claude.exe", "claude.cmd", "claude"},
```
And in `WinPaths`, swap to match:
```go
WinPaths: []binaryutil.WinPath{
    {Base: binaryutil.WinAppData, Parts: []string{"npm", "claude.exe"}},
    {Base: binaryutil.WinAppData, Parts: []string{"npm", "claude.cmd"}},
},
```

Modern Claude Code ships a native `claude.exe`; npm installs produce both `claude.cmd` (batch shim) and `claude.exe` (the actual binary or a direct executable). Preferring `.exe` avoids the Node.js `.cmd` spawn restriction entirely.

**Caveat:** If some Claude Code install variants only ship `.cmd` (no `.exe`), the `.cmd` fallback still needs to work. The ACP bridge would need to handle `.cmd` by spawning with `shell: true` — but that's an upstream fix in `@agentclientprotocol/claude-agent-acp`. As a defense-in-depth measure, `claudeacp/driver.go` could detect `.cmd`/`.bat` extensions on Windows and either resolve to the underlying executable or set an env flag for the ACP bridge.

### Fix 2 (secondary): Harden the rollback path for ConPTY "already exists"

The rollback's `startTransitionTarget` → `relaunchSession` → `runtime.Create` fails when the old ConPTY session wasn't destroyed. The rollback should explicitly destroy any lingering runtime session before attempting to recreate:

```go
// Before relaunching the source TUI controller during rollback:
_ = m.runtime.Destroy(ctx, runtimeHandle(rec.Metadata))
```

This is the same pattern already used in `stopSourceControllerConclusive` for the forward direction.

### Fix 3 (tertiary): Better error surfacing

The `spawn EINVAL` detail is trapped inside the ACP JSON-RPC error and only visible in the popup. The interface transition's `errorDetail` should detect Windows-spawn-class failures and surface a user-actionable message (e.g., "Claude Code was found as a .cmd file which cannot be spawned directly on Windows. Try installing Claude Code natively.").

## Impact

- **Platform:** Windows only (macOS/Linux use `claude` directly, no `.cmd` issue)
- **Severity:** Interface switch is completely broken for Windows users with npm-installed Claude Code
- **Recovery:** Requires AO restart; the affected session may be unrecoverable without manual cleanup
- **User experience:** The cascading errors (4 separate error messages) are confusing and alarming

## Related

- [#2598](https://github.com/Untrivial-ai/agent-orchestrator/issues/2598) — opaque `500 INTERNAL_ERROR` from unmapped service errors (explains the generic error in the cascade)
- CVE-2024-27980 — Node.js security fix that blocks spawning `.cmd`/`.bat` without `shell: true` (the upstream cause of `spawn EINVAL`)

AgentWrapper:
Fixes #3925

## Summary

- prefer and resolve the native `claude.exe` shipped by the user's Windows Claude Code npm installation instead of passing `claude.cmd` to the ACP Node runtime
- reject unresolved `.cmd`/`.bat` paths during ACP preflight with actionable guidance instead of surfacing opaque `spawn EINVAL`
- retain the source runtime handle through an interface transition and clear stale ConPTY registrations before rollback recreates the TUI
- add regression coverage for npm executable resolution, ACP shim validation, and failed-target rollback

AO continues to use the user's installed Claude Code CLI; the packaged ACP runtime remains provider-binary-free.

## Tests

- `go test ./internal/adapters/agent/claudecode ./internal/adapters/chatdriver/claudeacp ./internal/session_manager`
- `go test -race ./internal/adapters/agent/claudecode ./internal/adapters/chatdriver/claudeacp ./internal/session_manager`
- Windows/amd64 cross-compilation of the three affected test packages with `go test -c`
- `env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_ISSUE_ID go test ./...`
- `env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_ISSUE_ID npm run lint`

## Risk

Low and Windows-focused. Non-Windows binary resolution is unchanged. Older Windows Claude Code installations that expose only a command shim now receive a clear preflight error and must update/reinstall instead of entering the broken transition path.

### DIFF
--- a/backend/internal/adapters/agent/claudecode/claudecode.go
+++ b/backend/internal/adapters/agent/claudecode/claudecode.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -478,20 +479,49 @@ func SessionUUID(aoSessionID string) string {
 var claudeBinarySpec = binaryutil.BinarySpec{
 	Label:         "claude",
 	Names:         []string{"claude"},
-	WinNames:      []string{"claude.cmd", "claude.exe", "claude"},
+	WinNames:      []string{"claude.exe", "claude.cmd", "claude"},
 	UnixPaths:     []string{"/usr/local/bin/claude", "/opt/homebrew/bin/claude"},
 	UnixHomePaths: binaryutil.NodeManagedUnixHomePaths("claude", []string{".claude", "local", "claude"}),
 	NodeManaged:   true,
 	WinPaths: []binaryutil.WinPath{
-		{Base: binaryutil.WinAppData, Parts: []string{"npm", "claude.cmd"}},
+		{Base: binaryutil.WinAppData, Parts: []string{"npm", "node_modules", "@anthropic-ai", "claude-code", "bin", "claude.exe"}},
 		{Base: binaryutil.WinAppData, Parts: []string{"npm", "claude.exe"}},
+		{Base: binaryutil.WinAppData, Parts: []string{"npm", "claude.cmd"}},
 	},
 }
 
 // ResolveClaudeBinary returns the path to the claude binary, or a wrapped
 // ports.ErrAgentBinaryNotFound when it is absent.
 func ResolveClaudeBinary(ctx context.Context) (string, error) {
-	return binaryutil.ResolveBinary(ctx, claudeBinarySpec)
+	binary, err := binaryutil.ResolveBinary(ctx, claudeBinarySpec)
+	if err != nil {
+		return "", err
+	}
+	return resolveNativeWindowsClaude(binary, runtime.GOOS), nil
+}
+
+// resolveNativeWindowsClaude unwraps the npm-generated command shim to the
+// native executable shipped by Claude Code. The ACP bridge passes this path to
+// Node child_process.spawn, which cannot execute .cmd files directly on current
+// Node releases.
+func resolveNativeWindowsClaude(path, goos string) string {
+	if goos != "windows" || !strings.EqualFold(filepath.Ext(path), ".cmd") {
+		return path
+	}
+	for _, candidate := range windowsNativeClaudeCandidatesForShim(path) {
+		if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
+			return candidate
+		}
+	}
+	return path
+}
+
+func windowsNativeClaudeCandidatesForShim(shim string) []string {
+	dir := filepath.Dir(shim)
+	return []string{
+		filepath.Join(dir, "claude.exe"),
+		filepath.Join(dir, "node_modules", "@anthropic-ai", "claude-code", "bin", "claude.exe"),
+	}
 }
 
 func (p *Plugin) claudeBinary(ctx context.Context) (string, error) {
--- a/backend/internal/adapters/agent/claudecode/claudecode_test.go
+++ b/backend/internal/adapters/agent/claudecode/claudecode_test.go
@@ -32,6 +32,38 @@ func TestNativeConversationIDUsesTheSameClaudeUUIDAcrossInterfaces(t *testing.T)
 	}
 }
 
+func TestWindowsNativeClaudeCandidatesForNPMShim(t *testing.T) {
+	shim := filepath.Join("prefix", "claude.cmd")
+	want := []string{
+		filepath.Join("prefix", "claude.exe"),
+		filepath.Join("prefix", "node_modules", "@anthropic-ai", "claude-code", "bin", "claude.exe"),
+	}
+	if got := windowsNativeClaudeCandidatesForShim(shim); !reflect.DeepEqual(got, want) {
+		t.Fatalf("candidates = %#v, want %#v", got, want)
+	}
+}
+
+func TestResolveNativeWindowsClaudeFindsNPMExecutable(t *testing.T) {
+	prefix := t.TempDir()
+	shim := filepath.Join(prefix, "claude.cmd")
+	if err := os.WriteFile(shim, []byte("@echo off\r\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	want := filepath.Join(prefix, "node_modules", "@anthropic-ai", "claude-code", "bin", "claude.exe")
+	if err := os.MkdirAll(filepath.Dir(want), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(want, []byte("native claude"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if got := resolveNativeWindowsClaude(shim, "windows"); got != want {
+		t.Fatalf("resolved binary = %q, want %q", got, want)
+	}
+	if got := resolveNativeWindowsClaude(shim, "linux"); got != shim {
+		t.Fatalf("non-Windows resolution = %q, want unchanged shim %q", got, shim)
+	}
+}
+
 func TestNativeConversationExistsRequiresPersistedClaudeTranscript(t *testing.T) {
 	configDir := t.TempDir()
 	t.Setenv("CLAUDE_CONFIG_DIR", configDir)
--- a/backend/internal/adapters/chatdriver/claudeacp/driver.go
+++ b/backend/internal/adapters/chatdriver/claudeacp/driver.go
@@ -50,7 +50,11 @@ func New(plugin claudePlugin, log *slog.Logger) ports.ChatDriver {
 			if _, err := resolveRuntime(ctx); err != nil {
 				return fmt.Errorf("%w: %w", ports.ErrChatDriverUnavailable, err)
 			}
-			if _, err := plugin.ResolveBinary(ctx); err != nil {
+			claudeBinary, err := plugin.ResolveBinary(ctx)
+			if err != nil {
+				return fmt.Errorf("%w: %w", ports.ErrChatDriverUnavailable, err)
+			}
+			if err := validateClaudeACPExecutable(claudeBinary, runtime.GOOS); err != nil {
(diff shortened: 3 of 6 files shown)


i-trytoohard (commented):
## Code Review Summary

**Verdict: Looks good — no blocking issues.** (+172 -7, 6 files, all CI green including `windows-latest`)

This PR addresses all three failure layers identified in [#3925](https://github.com/Untrivial-ai/agent-orchestrator/issues/3925) with well-scoped, test-covered changes.

### ✅ Change 1: Binary resolution (`claudecode.go`)

**`WinNames` swap + `WinPaths` reordering** — correct. Preferring `claude.exe` over `claude.cmd` in both PATH lookup and candidate paths directly addresses Layer 2 of the root cause. The new `WinPath` for `%APPDATA%\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe` matches the exact location confirmed by the on-machine Codex report — the actual native binary in an npm global install.

**`resolveNativeWindowsClaude` shim unwrap** — good defense-in-depth. Even if `ResolveBinary` still finds a `.cmd` (e.g., from PATH in a non-standard location), this function tries to upgrade it to the sibling `.exe` or the npm package dir `.exe`. The `strings.EqualFold` case-insensitive ext check handles `.CMD`/`.Cmd` correctly.

The native resolution happens inside `ResolveClaudeBinary` before `p.claudeBinary` caches it — so the cached value is always the resolved `.exe`. Correct.

### ✅ Change 2: ACP preflight validation (`claudeacp/driver.go`)

`validateClaudeACPExecutable` catches `.cmd`/`.bat` at **preflight time** (`Probe`), which maps to the `preflighting` phase of the interface transition — *before* the source controller is stopped. This is a critical improvement: users now get an actionable error ("reinstall or update Claude Code") at the preflight gate instead of the opaque `spawn EINVAL` after the session is already half-torn-down.

The redundant validation in `Launch` is belt-and-suspenders — fine.

### ✅ Change 3: Rollback hardening (`interface_transition.go`)

Capturing `sourceRuntimeHandle` before the controller epoch changes the mode, then explicitly calling `runtime.Destroy` during rollback, directly fixes the ConPTY "session already exists" cascade. The destroy is idempotent (ConPTY's `Destroy` returns nil for unknown sessions), so this is safe even if the session was already cleaned up. The new test `TestInterfaceTransitionRollbackClearsStaleTUIRuntimeBeforeRestore` covers this end-to-end with a realistic `spawn EINVAL` → rollback scenario.

### ✅ Test coverage

All three changes have proportional tests:
- `TestWindowsNativeClaudeCandidatesForNPMShim` — path generation
- `TestResolveNativeWindowsClaudeFindsNPMExecutable` — full resolution with temp-dir mock, including non-Windows passthrough
- `TestValidateClaudeACPExecutableRejectsWindowsCommandShims` — table test (`.exe`, `.cmd`, `.BAT`, non-Windows)
- `TestInterfaceTransitionRollbackClearsStaleTUIRuntimeBeforeRestore` — full transition lifecycle with stale runtime

### 💡 Minor observation (non-blocking)

The `windowsNativeClaudeCandidatesForShim` second candidate (`<dir>/node_modules/@anthropic-ai/claude-code/bin/claude.exe`) assumes an npm global layout. If `claude.cmd` was resolved from a non-npm PATH location (e.g., `C:\tools\bin\`), this sub-path won't exist. The first candidate (`<dir>/claude.exe`) covers that case anyway, so it's not a gap — just noting the assumption.

### What this PR does NOT address (and shouldn't)

The packaging layer (`build-acp-runtime.mjs` stripping `--omit=optional` native SDK binaries) is intentionally left alone. The PR's approach — resolve the user's separately installed `.exe` at runtime — is the right scope. The packaging fix (Fix C in the issue) is a separate concern that can be tracked independently.

### Regression risk assessment

**Low.** All changes are Windows-gated (`runtime.GOOS == "windows"` or `goos != "windows"` early returns). The `WinNames`/`WinPaths` reordering only affects search order on Windows. The rollback Destroy call is idempotent and only fires for TUI source mode transitions. No macOS/Linux code paths are touched.

---
*Reviewed by @i-trytoohard*

PR #3928, opened 2026-08-13
fix(preview): teach workers workspace path semantics

AgentWrapper:
Related to #3436

## Summary

- make the generated worker system prompt state that `ao preview` file targets are relative to the session workspace root, regardless of shell working directory
- tell workers to use `ao preview README.md` instead of `../README.md`, and explain that AO uses its existing confined loopback preview rather than `file://` or an ad-hoc static server
- align the installed `using-ao` preview guide, CLI help, and CLI documentation with those semantics
- add regression assertions for the generated prompt, embedded skill, and command help

## Tests

- `cd backend && env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_ISSUE_ID go build ./...`
- `cd backend && env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_ISSUE_ID go test ./...`
- `env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_ISSUE_ID npm run lint`

## Scope

This complements the daemon/browser resolution work already tracked in #3436 and PR #3441; it does not duplicate that implementation.


backend/internal/skillassets/using-ao/commands/preview.md
```diff
@@ -9,6 +9,28 @@ with `ao preview <workspace-path>`. Never create or modify `package.json`,
...
+root or any subdirectory:
+
+```bash
+ao preview README.md
+ao preview docs/guide.md
+```
+
+Do not compensate for the current directory with `../README.md`. Relative
+targets already start at the workspace root, so parent traversal instead tries
+to leave the workspace and is invalid. An absolute path is also accepted when
```
AgentWrapper: Non-blocking accuracy note: the daemon does not actually *reject* `../README.md`. `CleanWorkspacePath` fails the `..` segment, `resolveLocalPreview` returns ok=false, and `resolvePreviewTarget` then falls through and stores the string **verbatim** — the CLI exits 0 and the browser panel just gets a dead target, with no error surfaced. As worker guidance this wording is fine, but if you want "is invalid" to be literally true, a follow-up could make `setPreview` return `PREVIEW_FILE_OUTSIDE_WORKSPACE` for relative targets containing `..` so workers get immediate feedback instead of a silently broken panel.

AgentWrapper (commented):
**Verdict: approve** (posted as COMMENT because GitHub rejects APPROVE on the author account). Docs/prompt/test-guard change only; no runtime behavior is modified.

I verified each new claim against the daemon implementation on `main`:

- **Workspace-root-relative resolution is real, not aspirational.** The CLI posts the raw target (`backend/internal/cli/preview.go` → `openPreview`), and the daemon resolves it with `EntryAtPath(workspacePath, raw)` via `resolveLocalPreview` in `backend/internal/httpd/controllers/sessions.go` — it never sees the shell cwd, so the new prompt/docs wording is accurate.
- **Absolute and `file:` targets** go through `workspaceAbsolutePreviewURL`, which confines the resolved path to the resolved workspace, matching the updated `docs/cli/README.md` text. Removing the stale claim that a URL argument opens `file://` verbatim is a genuine accuracy fix.
- **Tests line up.** The new `skillassets_test.go` strings match `preview.md` after the whitespace normalization the test applies; the `manager_test.go` assertions match the new `aoSkillPointer` text exactly; the pointer is ~213 words against the raised 220-word growth guard; and `preview_test.go` matches the new help text.

One non-blocking inline note below about what actually happens to `../` targets at runtime.

PR #3933, opened 2026-08-13
fix(desktop): persist daemon output on the piped launch path

### LINKED ISSUE #2689: Desktop-launched daemon crash output is not persisted

axisrow:
## Summary

A desktop-launched daemon wrote nothing to disk: the supervisor consumed its stdout/stderr pipes and forwarded them to the Electron console, so a panic stack died with the app and the request ID the API handed out had nothing to correlate against. Only `AO_KEEP_DAEMON` produced a durable log, because that path redirects stdio to a file at spawn time.

Closes #2689.

## What

Mirror the pipes to `~/.ao/daemon.log` instead of redirecting them: the port scanners still need the streams, so the log is a tee. Rotate one generation at 8 MiB, and stamp the exit reason before closing the file — a bare stack with no terminator is indistinguishable from a truncated log. Every failure is best-effort: a log that cannot be opened or written must never take the app down with it.

## Verification

```bash
npm run frontend:typecheck
```

Manual: kill the daemon mid-session on a packaged desktop install, confirm `~/.ao/daemon.log` carries the stack/exit reason that used to be visible only in the Electron console (or lost entirely outside `AO_KEEP_DAEMON`).

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -38,7 +38,17 @@ import {
 import { coerceUiSettings, readUiSettings, writeUiSettings, type UiSettings } from "./main/ui-settings";
 import { spawn, type ChildProcess } from "node:child_process";
 import { randomBytes, randomUUID } from "node:crypto";
-import { closeSync, existsSync, openSync, readFileSync } from "node:fs";
+import {
+	closeSync,
+	createWriteStream,
+	existsSync,
+	mkdirSync,
+	openSync,
+	readFileSync,
+	renameSync,
+	statSync,
+	type WriteStream,
+} from "node:fs";
 import { mkdir, readFile, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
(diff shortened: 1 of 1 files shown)


axisrow:
Duplicate of #3892, opened by mistake without checking for an existing PR first. Closing in favor of #3892.

PR #3934, opened 2026-08-13
fix(reviews): name the reviewer the daemon would actually run

### LINKED ISSUE #3914: bug(reviews): default reviewer label ignores the worker's harness, naming claude-code while a different agent runs
## Summary

The Review panel's default-reviewer label does not agree with the daemon's own resolution rule. For a worker whose harness is not `claude-code`, and whose project configures no reviewer, the picker reads `claude-code` while the review actually runs on the worker's harness.

Observed with a Codex worker: the picker showed `claude-code`, and pressing Run started Codex. Choosing `Claude Code` explicitly from the menu is what makes the label and the run agree, because that writes a session-level preference and both sides then take the same first branch.

## Root cause

The daemon resolves the reviewer in four steps (`Engine.reviewerHarness` → `ProjectConfig.ResolveReviewerHarness`):

1. session preference (`sessions.reviewer_harness`)
2. project config `reviewers[0].harness`
3. **the worker's own harness**, when that harness is one of the inherited set (`claude-code`, `codex`, `opencode`, `muse`, `kimchi`)
4. `claude-code` as the final fallback

The renderer computes the same label in three steps and omits step 3 (`SessionInspector.tsx`):

```ts
const harness = latest?.harness || config?.reviewers?.[0]?.harness || "claude-code";
```

With no prior run and no project reviewer, the frontend jumps straight to the `claude-code` fallback while the daemon stops one step earlier at the worker's harness. Every non-`claude-code` inherited worker is mislabeled.

## Secondary: the label is a wire id, not a display name

The same label is rendered from the raw harness id rather than the existing display-name table, so the trigger reads `claude-code` while the menu entry for the very same agent reads `Claude Code`. The two look like two different, separately selectable agents, which is what makes the mismatch read as "there are two claude-code entries and the lowercase one is wrong".

A third source of lowercase labels: `ReviewerSelect` builds its pre-catalog fallback options as `{ id, label: id }`, so before `/api/v1/agents` resolves the whole menu is lowercase ids, then flips to proper names once the catalog arrives.

## Reproduction

1. Start a worker on a harness other than `claude-code` (e.g. Codex) in a project with no `reviewers` entry.
2. Open the session inspector, Reviews section, before any review has run.
3. The reviewer picker reads `claude-code`.
4. Press Run — the review runs on the worker's harness, not claude-code.

## Expected

The picker names the agent the daemon would actually pick, and names it with its display name.

## Note

The current behaviour is pinned by a test asserting the lowercase fallback for a Codex worker, so fixing this necessarily updates that expectation.

axisrow:
## Summary

The Review panel resolved its default-reviewer label in three steps — last run, project config, then a hardcoded `claude-code` — while the daemon resolves in four: session preference, project config, the worker's own harness when that harness is an inherited reviewer, and `claude-code` only as the final fallback.

The missing worker step is the whole bug. A codex worker in a project with no configured reviewer was labelled `claude-code`, and pressing Run started codex. Picking Claude Code from the menu is what made the two agree, because that writes a session preference and both sides then take their first branch.

Closes #3914.

## What

- `inheritedReviewerHarness` mirrors `ResolveReviewerHarness`'s switch rather than the full reviewer vocabulary: only the original unattended-safe set is inherited, so a new experimental adapter never silently becomes a project's reviewer.
- Label the picker with display names instead of wire ids. The trigger printed the raw harness id while the menu row for the same agent printed the catalog's "Claude Code", so one agent read as two entries — which is what made the mismatch look like a duplicate. `ReviewerSelect`'s pre-catalog fallback options had the same problem from the other end: they carried `label: id`, so the entire menu was lowercase until `/api/v1/agents` resolved and then flipped case.
- The existing "shows claude-code as the default reviewer before a run exists" test pinned the wrong behaviour for a codex worker; it now asserts the inherited harness.

## Verification

```bash
npm run frontend:typecheck
npx vitest run src/renderer/components/SessionInspector.test.tsx   # 77 passed
```

### DIFF
--- a/frontend/src/renderer/components/ReviewerSelect.tsx
+++ b/frontend/src/renderer/components/ReviewerSelect.tsx
@@ -1,4 +1,5 @@
 import type { components } from "../../api/schema";
+import { agentLabel } from "../lib/agent-options";
 import { buildRankedAgentOptions } from "../lib/agent-select-options";
 import { KNOWN_REVIEWER_HARNESS_IDS } from "../lib/reviewer-harnesses";
 import { AgentAvatar } from "./AgentAvatar";
@@ -52,9 +53,12 @@ export function ReviewerSelect({
 	installed?: components["schemas"]["AgentInfo"][];
 	supported?: components["schemas"]["AgentInfo"][];
 }) {
+	// Until the daemon's catalog arrives these entries carry the whole menu, so
+	// label them the way the catalog would rather than printing bare ids: without
+	// this the same row reads "claude-code" now and "Claude Code" a moment later.
 	const fallbackAgents: components["schemas"]["AgentInfo"][] = [...KNOWN_REVIEWER_HARNESS_IDS].map((id) => ({
 		id,
-		label: id,
+		label: agentLabel(id),
 	}));
 	const filteredSupported = (supported ?? fallbackAgents).filter((a) => KNOWN_REVIEWER_HARNESS_IDS.has(a.id));
 	const supportedAgents = filteredSupported.length > 0 ? filteredSupported : fallbackAgents;
--- a/frontend/src/renderer/components/SessionInspector.test.tsx
+++ b/frontend/src/renderer/components/SessionInspector.test.tsx
@@ -157,6 +157,40 @@ function mockCommonGets(_unusedRuns: unknown[] = [], reviewerHandleId = "", revi
 	});
 }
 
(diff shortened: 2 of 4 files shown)


axisrow:
Duplicate of #3915, opened by mistake without checking for an existing PR first. Closing in favor of #3915.

PR #3930, opened 2026-08-13
feat(config): user-scope agent config layer (backend, no merge)

axisrow:
## Summary

First brick of the multi-scope config direction (#2844): a **User-scope agent config layer** above projects. A singleton `domain.AgentConfig` stored in a new SQLite table, exposed via `GET/PUT /api/v1/user-config` and editable via `ao user-config get/set`.

This PR is **backend-only and has no effect on workers yet** — the layer is stored and editable, but the merge into the effective agent config is tracked separately in #2999 (and lands only after #2848, which edits the same function).

Closes #2998.

Architecture decisions are recorded in `docs/adr/0002-user-scope-config.md`.

## What

- **Migration `0089`**: singleton table `user_config(id INTEGER PRIMARY KEY CHECK(id = 1), config TEXT)` — the `CHECK` enforces one row at the schema level; `config` is a nullable JSON blob of `domain.AgentConfig` (NULL = unset). (Renumbered from the original `0032` draft — upstream had since claimed that number.)
- **Store** (`store/user_config_store.go`): `GetUserConfig` absorbs `sql.ErrNoRows` → `(zero, false, nil)`; `marshalAgentConfig` writes NULL when `IsZero()`; corrupt JSON degrades to zero (never blocks the row). Mirrors the project-config store.
- **Service** (`service/userconfig/`): `Manager{Get, Set}`, `SetUserConfigInput`, `INVALID_USER_CONFIG` error code, wholesale replace + `Validate`.
- **Controller** (`controllers/user_config.go`): nil-safe OpenAPI-backed 501, strict JSON decode. Wired into `api.go` inside the REST-with-timeout group.
- **OpenAPI**: new `userConfigOperations()` in `specgen/build.go` + regenerated `openapi.yaml` / `frontend/src/api/schema.ts`. Drift + route-parity tests green.
- **CLI**: `ao user-config get/set` (`--model`, `--permission`, `--config-json`, `--clear`, `--json`), reusing the existing `agentConfig` mirror.
- **Daemon wiring**: `userConfigSvc := userconfigsvc.New(store)`.

## Design notes

- **Surface = exactly `domain.AgentConfig` today** (`Model` + `Permissions`). The PR reuses the type end to end, so when #2848-equivalent work expands `AgentConfig` (env / systemPrompt / mcp / pluginDirs) the User layer picks the new fields up **for free on rebase**, with no schema change.
[...]

PR #3935, opened 2026-08-13
fix(sessions): store activity timestamps in UTC so SQL comparisons hold

### LINKED ISSUE #3899: bug(agent-switch): local-zone activity_last_at strands the switch saga in stopping_source and makes the daemon unbootable

axisrow:
## Summary

An agent switch could strand itself in `stopping_source` with no way out: the source agent was already gone, resume answered `AGENT_RESUME_IN_PROGRESS`, a repeat switch failed the same way, and the daemon then refused to start because its boot reconciliation returned the same error.

The cause is a storage detail, not a saga bug. The SQLite driver stores a `time.Time` by its `String()` form, so a value written from a local-zone clock keeps its offset and monotonic reading:

```
2026-06-28 18:45:08.349363 +0800 CST m=+25660.013723251
```

instead of the form every other writer produces:

```
2026-06-28 10:45:08.349363 +0000 UTC
```

`activity_last_at` is compared directly in SQL, and those comparisons are lexicographic. A `+0800` wall clock sorts *above* the UTC rendering of a later instant, so the source-stop predicate (`activity_last_at <= stopped_at`) matched zero rows. `ConfirmAgentSwitchSourceStopped` read that as a concurrent ownership change and left the saga non-terminal for a recovery pass that could never succeed, because every attempt re-ran the same predicate.

Closes #3899.

## What

Fix it at all three levels:
- the reaper and activity observer default to `time.Now().UTC()` like every other clock in the tree;
- `normalActivity` normalizes whatever a caller supplies, so a future writer cannot reintroduce the skew;
- migration `0090` rewrites the rows already written that way.

(Renumbered from an internal `0089` draft to `0090` — `0089` is claimed by the in-flight #3930.)

## Verification

```bash
go test ./internal/storage/sqlite/... ./internal/observe/...
```

The regression test (`TestMigration0090NormalizesLocalZoneActivityTimestamps`) fails without the store change with exactly the observed symptom — `confirmed=false, err=nil`.

### DIFF
--- a/backend/internal/observe/activity/observer.go
+++ b/backend/internal/observe/activity/observer.go
@@ -74,7 +74,12 @@ func New(sessions sessionSource, sink activitySink, runtime outputReader, agents
 		o.outputLines = DefaultOutputLines
 	}
 	if o.clock == nil {
-		o.clock = time.Now
+		// UTC, not bare time.Now: this timestamp becomes sessions.activity_last_at,
+		// and the SQLite driver stores a time.Time by its String() form. A
+		// local-zone value keeps its monotonic reading and offset, which then
+		// fails to compare as a timestamp against the "… +0000 UTC" rows the
+		// other writers produce.
+		o.clock = func() time.Time { return time.Now().UTC() }
 	}
 	if o.logger == nil {
 		o.logger = slog.Default()
--- a/backend/internal/observe/reaper/reaper.go
+++ b/backend/internal/observe/reaper/reaper.go
@@ -89,7 +89,13 @@ func New(sink runtimeObservationSink, sessions sessionSource, runtime runtimePro
 		r.tick = DefaultTickInterval
 	}
 	if r.clock == nil {
-		r.clock = time.Now
+		// UTC, not bare time.Now: the observed timestamp reaches sessions.
+		// activity_last_at, and the SQLite driver stores a time.Time by its
+		// String() form. A local-zone value keeps its monotonic reading and
+		// offset ("… +0700 +07 m=+995.1"), which no longer compares as a
+		// timestamp against the "… +0000 UTC" rows every other writer produces —
+		// and those comparisons gate the agent-switch source-stop predicate.
+		r.clock = func() time.Time { return time.Now().UTC() }
 	}
 	if r.logger == nil {
 		r.logger = slog.Default()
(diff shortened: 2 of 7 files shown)


axisrow:
Duplicate of #3900, opened by mistake without checking for an existing PR first. Closing in favor of #3900.

PR #3947, opened 2026-08-13
fix: auto-approve host-trusted reviewers without enabling edits

nikhilachale:
## Summary

- remove Continue, Goose, Vibe, and Qwen from the supported reviewer set while retaining all four as normal agent harnesses
- update reviewer validation, registry coverage, UI options, and generated API contracts

## Why

AO reviewers must run unattended. Continue and Goose do not meet the desired reviewer contract.

Vibe cannot provide both properties AO needs from a reviewer: its `plan` agent is read-only but still pauses for confirmation before applying the plan, while its unattended `auto-approve` agent can edit files and run commands. The standalone `--auto-approve` flag approves tool calls but does not bypass the interactive apply-plan transition. Because AO cannot currently configure Vibe as both read-only and fully unattended, Vibe is removed from reviewer selection rather than exposing a reviewer that can stall or modify the checkout.

Qwen does not have an auto-approve + read-only mode available today, so it cannot be run as an unattended, non-mutating reviewer either. It is a lesser-used harness, so this is not being treated as P0 — if a user requests Qwen as a reviewer, a custom solution will be scoped then rather than shipping a reviewer that can stall or modify the checkout in the meantime.

Continue, Goose, Vibe, and Qwen all remain available as worker/orchestrator agent harnesses — only reviewer selection is narrowed.

## Validation

- `go test ./internal/domain ./internal/adapters/reviewer ./internal/httpd/... -count=1`
- `npm run api`

https://claude.ai/code/session_019QzCX9ECGt95H25SFC71Mb


backend/internal/adapters/reviewer/continueagent/continueagent.go
```diff
@@ -74,7 +74,7 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
 		envVars["CONTINUE_API_KEY"] = key
 	}
 	return ports.ReviewCommandSpec{
-		Argv:             []string{binary, "--readonly"},
+		Argv:             []string{binary, "--auto"},
```
neversettle17-101, line 77: can we add readonly as well as auto? we want to make sure it has readonly access tools

neversettle17-101 (changes requested):

backend/internal/adapters/reviewer/continueagent/continueagent.go
nikhilachale: either we can keep --read only or either we can pass --auto

Continue cannot provide it with --auto plus --readonly.

we can rely on system prompt  that they will not edit any files

neversettle17-101:
Let us also add the correct cancellation behaviour https://github.com/Untrivial-ai/agent-orchestrator/issues/3745

backend/internal/adapters/reviewer/continueagent/continueagent.go
neversettle17-101: Fixed in 13adbbc2a. Continue 1.5.47 does not support combining `--readonly` with `--auto`, so the reviewer now remains on `--readonly` to preserve read-only tool access. Goose, Qwen, and Vibe retain their unattended approval modes.

i-trytoohard:
@nikhilachale Marking as stale: no activity from you for 25 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

backend/internal/adapters/reviewer/continueagent/continueagent.go
```diff
@@ -44,8 +44,9 @@ func (*Reviewer) Harness() domain.ReviewerHarness { return HarnessID }
 var _ ports.Reviewer = (*Reviewer)(nil)
 var _ ports.ReviewerCanceller = (*Reviewer)(nil)
 
-// ReviewCommand launches Continue in readonly mode and injects the task after
-// readiness. CLI discovery and state are redirected to an AO-owned profile.
+// ReviewCommand launches Continue in readonly mode and injects
```
neversettle17-101: Continue is still documented and launched as readonly here; the production argv below remains `[]string{binary, "--readonly"}`, and the adapter test still asserts that exact argv. That means Continue reviewer runs can still hit the native approval prompts this change is meant to bypass. Please switch the host-trusted Continue launch to the supported auto mode, and update the exact argv test (and the contained/staged contract if that is also intended to bypass prompts).

neversettle17-101 (commented):
Needs changes. The Goose, Qwen, and Vibe reviewer launch modes were updated and their adapter tests were adjusted, but Continue still launches with readonly mode, so this PR does not remove approval prompts for that reviewer. I did not run tests per review-only instructions; I reviewed the diff and supplied validation notes.

backend/internal/adapters/reviewer/vibe/vibe.go
```diff
@@ -114,7 +115,7 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
 	if strings.TrimSpace(inv.TaskPromptRoot) != "" {
 		argv = append(argv, "--add-dir", inv.TaskPromptRoot)
 	}
-	argv = append(argv, "--agent", reviewerAgent)
+	argv = append(argv, "--agent", hostTrustedReviewerAgent, "--auto-approve")
```
neversettle17-101: This launch now depends on `--auto-approve`, but `verifyCompatibility` still only checks `--agent`, `--workdir`, and `--trust`. If a supported/minimum Vibe build (or a future changed CLI) lacks the standalone auto-approve flag, preflight succeeds and the reviewer fails only at spawn time. Please add `--auto-approve` to the required flag probe (and update the compatibility tests) so this fails closed like the other pinned launch-surface checks.

neversettle17-101 (commented):
## Findings

Needs changes.

- **Medium**: Vibe auto-approve support is launched without being compatibility-checked.

## What This PR Does

Removes Continue and Goose from the reviewer vocabulary/registry/UI while leaving them as worker harnesses, switches Qwen reviewer launch to yolo approval mode, and adds Vibe auto-approve to the host-trusted reviewer launch. Generated API request enums, frontend reviewer lists, docs, and tests are updated accordingly.

## Regression Risk

The main visible risk is Vibe reviewer launch failures on installations whose help contract does not include the newly used flag; the current compatibility probe would not catch that before constructing the command. Existing persisted project/session reviewer values of `continue` or `goose` may also surface as unsupported after this removal, but that appears aligned with the support removal.

## Testing

The PR updates unit/contract tests around registry membership and reviewer argv construction, and the PR description reports focused Go tests, lint, frontend typecheck, and API regeneration. I did not run tests as part of this review.

neversettle17-101 (approved):

PR #3949, opened 2026-08-13
fix: group session control toggles

neversettle17-101:
## Summary
- Add a Session controls section in the session inspector
- Move CI auto-send, review auto-send, and terminate-on-merge controls into that section
- Reword CI and terminate-on-merge tooltips to be clearer

## Tests
- npm test -- src/renderer/components/SessionInspector.test.tsx
- npm run frontend:typecheck

## Risks
- UI copy/layout-only change scoped to the session inspector summary panel


neversettle17-101:
Screenshot 2026-08-13 at 3 08 21 PM
Screenshot 2026-08-13 at 3 08 29 PM

whoisasx:
@neversettle17-101 , in the activity section it states "Input Needed Changes Requested". There should a comma separating both the states.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -515,8 +505,10 @@ function CompletionControls({ session }: { session: WorkspaceSession }) {
 	if (session.isTerminated === true) return null;
 
 	return (
-		<Section title={t("inspector.completion")}>
-			{canTerminateNow ? (
+		<Section title={t("inspector.sessionControls")}>
+			{hasPRs ? <AutoInjectCIPolicyControl session={session} /> : null}
```
codebanditssss: This hides the CI policy until a PR already exists, but this setting is the default captured by newly created PRs (the prior test explicitly covered persisting it before a PR exists, and the previous help text explained that existing PRs retain their original setting). Once `hasPRs` becomes true, changing this toggle is too late for the existing PR, so users lose the only useful point at which to configure the policy. Please keep `AutoInjectCIPolicyControl` visible before PR creation; grouping it under Session controls does not require gating it on `hasPRs`. Restore the pre-PR persistence regression test rather than asserting that the control is hidden.

frontend/src/renderer/i18n/de.json
```diff
@@ -191,13 +191,14 @@
 	"inspector.aria": "Sitzungs-Inspektor",
 	"inspector.branch": "Branch",
 	"inspector.ci.autoInject": "CI-Fehler automatisch senden",
-	"inspector.ci.autoInjectDescription": "Gilt für neu erstellte Pull Requests. Bestehende Pull Requests behalten ihre ursprüngliche Einstellung.",
+	"inspector.ci.autoInjectDescription": "When disabled, CI failures are not sent to the worker agent to fix.",
```
codebanditssss: This revised help text is English in the German locale, and the same English sentence was copied into `es`, `fr`, `ja`, `ko`, `pt-BR`, and `zh-CN`. Please translate the new copy in all seven locale files (or retain their existing localized wording with the clarified semantics) rather than shipping an English tooltip across every non-English UI.

codebanditssss (changes requested):
The regrouping itself is clean and all current checks pass, but I don’t think this is ready to approve. One change removes the only useful time to configure the CI policy, and the revised CI help text is left in English in every non-English locale.

neversettle17-101:
@whoisasx it is the dummy data, also haven't made changes in that

frontend/src/renderer/components/SessionInspector.tsx
codebanditssss: This latest commit (`b552093`) reintroduces the exact blocker. This is not just a UI preference: migration `0088_add_auto_inject_ci_toggle.sql` defines the session value as a default and snapshots it into `pr.auto_inject_ci` when the PR is first recorded; the migration comment states that the PR copy is immutable, and `TestPRAutoInjectCISnapshotsSessionDefaultAtCreation` verifies that contract. Showing the control only after `hasPRs` is true means the current PR has already captured its value, so changing the visible toggle cannot affect it. Please keep the control visible before PR creation and restore the persistence test.

frontend/src/renderer/i18n/de.json
codebanditssss: Verified on the current head: all seven non-English locale strings are now translated. This comment is resolved—thank you.

frontend/src/renderer/components/SessionInspector.tsx
whoisasx: [P1] Keep this session default configurable before a PR exists

This condition hides the switch until after the first PR has already snapshotted session.autoInjectCI. Turning the switch off once it appears only changes the default for future PRs; the existing PR can still inject its next CI failure while the UI shows the policy as disabled. Please render the session default before a PR exists and explain its new-PR scope, or add a per-PR policy endpoint and bind this control to the selected PR.

whoisasx (dismissed):
Thanks for addressing the localization feedback in the latest commits. One correctness blocker remains: the CI auto-injection setting is a session default that is snapshotted when a pull request is created, so hiding the control until a PR exists prevents users from configuring the first PR and makes later changes look applicable to an existing PR when they are not.

Please restore access to this default before a PR exists and preserve copy explaining that it applies to newly created PRs, or implement a true per-PR mutation if the intended behavior is to control existing PRs. Please also close out the existing Activity-label punctuation comment.

The focused tests and current CI are green, but I am keeping this as Changes Requested until the policy semantics are corrected.

codebanditssss (approved):
Re-reviewed the latest head. The CI default is configurable before PR creation again, the persistence regression test is restored, and the tooltip now accurately explains the new-PR scope in every locale. The focused test job passes. renderer-smoke failed only because Playwright’s CDN returned a location-based 403 while installing Chrome, before the smoke suite ran; that failure is unrelated to this patch. Approved.

PR #3951, opened 2026-08-13
fix(interface-switch): persist recovery notice dismissal

### LINKED ISSUE #3950: bug(interface-switch): dismissed recovery notice resurfaces after remount or app restart
## Summary

The terminal interface-transition notice is backed by the latest durable transition, but dismissal is only component-local UI state. Navigating away and back, switching render surfaces, or restarting the desktop app remounts the component and shows the same already-settled transition again.

The recovery copy is also misleading: the heading says **Interface switch needs recovery**, while the daemon detail says AO already recovered the session and the transition is terminal.

Interface switch needs recovery notice

## Confirmed cause

- The daemon status endpoint intentionally returns the latest transition, including terminal history.
- Desktop stores `dismissedTransitionID` in `SessionView` component state, so unmount/remount loses it.
- Mobile terminal has the same component-local pattern.
- Mobile Chat renders the terminal failure/recovery banner without a dismiss action.
- The transition row has no durable fact recording that its user-facing notice was acknowledged.

This is separate from, but was exposed by, the nested Codex turn lifecycle bug in #3878 (fix: #3890). That bug explains why a handoff could require daemon recovery; this issue explains why one recovered historical event keeps looking new afterward.

## Expected behavior

A failed/recovered transition should remain durable for audit and diagnostics, but its notice should stop appearing everywhere after the user dismisses it once.

## Acceptance criteria

- Add a durable, nullable notice-acknowledgement timestamp to an interface-transition row.
- Expose an idempotent daemon API that acknowledges a specific transition under a specific session; do not race or dismiss a newer transition.
- Permit acknowledgement only for terminal `failed` / `recovery_required` notices.
- Desktop, mobile terminal, and mobile Chat hide acknowledged notices and use the same API.
- The acknowledgement participates in normal SQLite change events so other connected clients converge.
- A remount, app restart, or second client does not resurrect an acknowledged notice.
- Keep the transition row and diagnostics; do not delete history.
- Describe `DAEMON_RESTARTED` recovery rows as already recovered; keep other `recovery_required` rows as needing attention.
- Add backend/API and UI regression coverage.

illegalcall:
## Summary

- retain failed/recovered interface transitions as durable diagnostics while storing a separate `notice_acknowledged_at` user fact
- add an idempotent, generation-safe daemon endpoint scoped by both session and transition ID
- propagate acknowledgement through the existing SQLite CDC stream so connected desktop clients refetch the exact transition status
- replace component-local dismissal on desktop/mobile Terminal and add the missing action in mobile Chat
- describe daemon-restart rows as recovered while preserving “needs attention” wording for genuinely unresolved recovery states

This is the durable UI/recovery half of the incident chain. #3878 / #3890 prevent nested Codex lifecycle from creating the stuck handoff; this PR prevents one historical recovery row from being presented as a new incident after remounts or restarts.

Closes #3950

## Why daemon-owned

The latest transition is intentionally durable and shared by desktop and mobile. A local React flag or browser storage cannot converge across clients and is lost on remount/reinstall. The acknowledgement is an observed user action, not derived display status, so it belongs beside the durable transition while the transition itself remains intact.

## Verification

- `go test -p 4 ./...`
- `go test -race ./internal/storage/sqlite/store ./internal/session_manager ./internal/service/session ./internal/httpd/controllers`
- `golangci-lint v2.12.2` with fresh Go/lint caches: 0 issues
- frontend: 177 files, 2,237 passed, 1 skipped
- mobile: 38 files, 380 passed
- `npm --prefix packages/mobile run typecheck`
- sqlc/OpenAPI/TypeScript regeneration is stable


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
## Native Electron verification

Verified on PR head `10262d030128ab1ed61ef4bba6629799a36205fc` using an isolated Electron + daemon run and synthetic SQLite data.

### 1. Recovery notice before acknowledgement

The real session view renders the recovered interface-switch notice and its dismiss control.

Recovery notice visible before acknowledgement

### 2. Same session after dismissal and a full Electron + daemon restart

The transition still exists, but the acknowledged notice does not return.

Recovery notice remains dismissed after full restart

### Assertions exercised

- Clicked the actual **Dismiss interface switch message** button in Electron.
- The UI sent `PUT /api/v1/sessions/scratch-1/interface-transition/transition-ack-test/notice-acknowledgement` and received HTTP 200.
- SQLite retained the original `recovery_required` transition and stored `notice_acknowledged_at`.
- The acknowledgement added exactly one trigger-sourced transition CDC event (transition events: 1 before, 2 after).
- After stopping both Electron and its daemon and launching them again against the same database, the API still returned `noticeAcknowledgedAt`; the notice and dismiss button were both absent.

The fixture used the `fake` harness and no user AO data. On restart that harness is independently marked exited because it has no terminal, so I reset only the synthetic session activity flag for a like-for-like screenshot; the transition and acknowledgement were not changed.

illegalcall:
Resolved against current main (`3348277ec`, including #3953) in `f4429b036`.

Integration details:
- moved the acknowledgement migration to `0099_interface_transition_notice_acknowledgement.sql`; main retains `0098_session_native_identity_generation.sql`
- preserved #3953 desktop recovery actions (`Discard draft and switch` / `Cancel request and switch`) alongside durable dismissal
- preserved mobile Terminal recovery confirmation/actions alongside durable dismissal
- regenerated sqlc, OpenAPI, and frontend API types with no drift

Local verification:
- focused backend storage/session/service/controller suites: pass
- focused backend race suite: pass
- `go vet ./...`: pass
- frontend typecheck: pass; 190 files / 2,500 tests passed, 1 skipped
- mobile typecheck: pass; 55 files / 454 tests passed

The broad backend run passed every package except one load-sensitive OpenCode auth timeout; that package passed immediately when rerun alone.

nikhilachale (approved):
PR #3951 passes the focused automated checks and manual local testing. It looks ready from the local verification side.

PR #3952, opened 2026-08-13
feat(frontend): add Discord Rich Presence

bluwwi:
Shows your AO activity in Discord — agent count, status, and uptime timer.

## What it does

- **Discord Rich Presence** toggle in Settings → General
- Shows "Orchestrating N agents" with the highest-priority session status (Working, Needs Input, In Review, Idle, etc.)
- Timer starts when AO opens and stays constant until you close it
- Toggle persists across restarts (stored in `~/.ao/rpc-settings.json`)
- Only starts the Discord IPC client when enabled — no background activity when off
- Shows "Orchestrating 0 agents" / "Idle" when no active sessions (never goes blank while enabled)

## How it works

Frontend-only. No Go backend changes. The Electron main process:
- Connects to Discord via `@xhayper/discord-rpc` (IPC to Discord desktop app)
- Polls `/api/v1/sessions` + `/api/v1/projects` every 15 seconds
- Maps 14 session statuses to a single representative status with priority ordering
- Pushes activity to Discord via `SET_ACTIVITY` RPC

## Files

- `frontend/src/shared/rpc.ts` — types, constants, client ID
- `frontend/src/main/rpc-settings.ts` — settings read/write under `~/.ao`
- `frontend/src/main/discord-rpc.ts` — RPC client, status mapping, activity builder
- `frontend/src/main.ts` — lifecycle hooks, IPC handlers, daemon port tap
- `frontend/src/preload.ts` — `rpc` namespace in contextBridge
- `frontend/src/renderer/stores/rpc-store.ts` — Zustand store
- `frontend/src/renderer/hooks/useRpcStatus.ts` — status subscription hook
- `frontend/src/renderer/components/settings/GeneralSettingsSection.tsx` — toggle row
- 8 i18n locale files

## Tests

- `rpc-settings.test.ts` — 7 tests (read, write, coerce, atomic write)
- `discord-rpc.test.ts` — 25 tests (all 14 statuses, priority ordering, 0-agent fallback, buttons, timestamps)
- Typecheck: clean
- No conflicts with `upstream/main`

## Notes

- Discord Application ID is hardcoded. Maintainers need to create a Discord app and replace the ID in `shared/rpc.ts`.
- The `ao-logo` image asset must be uploaded to the Discord app's Art Assets page for the large image to render.
- Works on Windows, macOS, and Linux (no platform-specific code; `@xhayper/discord-rpc` handles IPC transport internally).


bluwwi:
Screenshot 2026-08-12 224659
image

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
```diff
@@ -66,6 +68,45 @@ function SessionInterfaceRow() {
 	);
 }
 
+function DiscordRichPresenceRow() {
+	const { t } = useTranslation();
+	const enabled = useRpcStore((state) => state.enabled);
```
codebanditssss, line 75: The persisted RPC setting is never loaded into this store. `rpc-store.ts` defines `load()`, but neither this row nor any other call site invokes it (and the new `useRpcStatus()` hook is also unused). After restarting with RPC enabled, main starts publishing presence while this switch renders the store default `false`; the user therefore sees Discord Rich Presence as off and cannot disable it with the first click. Please hydrate the store when this settings surface mounts (or during app bootstrap), render/disable appropriately until `loaded`, and add a regression test starting from persisted `{ enabled: true }`.

frontend/src/main/discord-rpc.ts
```diff
@@ -0,0 +1,204 @@
...
+		statusListeners = statusListeners.filter((l) => l !== listener);
+	};
+}
+
+export function getRpcStatus(): RpcStatus {
+	return { state: connectionState };
+}
+
+export async function startDiscordRpc(): Promise<void> {
+	if (client) return;
```
codebanditssss, line 92: A failed `client.login()` leaves `client` non-null and starts a timer whose `refreshPresence()` immediately returns while disconnected. Every later `startDiscordRpc()` call then exits here, so if Discord is closed when AO launches and opened afterward, Rich Presence never connects until the user explicitly toggles the feature off and on (or restarts AO with Discord already running). Please destroy/null the failed client and add bounded reconnect handling, with a test covering login failure followed by a successful retry.

codebanditssss (changes requested):
I don’t think this is ready to approve yet. Two user-visible lifecycle issues remain, and the current `test` and `renderer-smoke` checks are failing. The failures show bridge/type drift against current main, so the branch also needs to be updated and made green before merge.

frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
whoisasx, line 75: [P1] Hydrate the persisted consent state before rendering

This selector only sees the Zustand default because useRpcStore.load has no call site. After a user enables Rich Presence and restarts AO, the main process starts Discord from rpc-settings.json while Settings renders this switch as off. That makes an active external presence appear disabled and requires an extra click sequence to stop it. Load the store during renderer bootstrap before exposing the control, and add a restart-persistence rendering test.

frontend/src/shared/rpc.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const RPC_SETTINGS_FILE_NAME = "rpc-settings.json";
+
+export const RPC_GET_SETTINGS_CHANNEL = "rpc:getSettings";
+export const RPC_SET_SETTINGS_CHANNEL = "rpc:setSettings";
+export const RPC_GET_STATUS_CHANNEL = "rpc:getStatus";
+export const RPC_STATUS_CHANNEL = "rpc:status";
+
+export const DISCORD_CLIENT_ID = "1512528923303350463";
```
whoisasx: [P1] Ship with a maintainer-owned Discord application

The PR description says this application ID still needs to be replaced and the ao-logo asset still needs to be uploaded. Merging this value makes production presence depend on an application identity and branding that maintainers have not provisioned or release-tested. Please create the official application under maintainer ownership, provision its assets, replace this ID, and verify a packaged build before merging.

frontend/src/main/discord-rpc.ts
whoisasx, line 92: [P2] Reset and reconnect after IPC failures

A failed login leaves client non-null, so every later startDiscordRpc call returns here. Request failures likewise only set disconnected and never destroy or retry the client. Starting AO before Discord, restarting Discord, or a transient IPC loss therefore disables presence until the user toggles it or restarts AO. Reset the client on failure/disconnect and retry with bounded backoff while the setting remains enabled; cover startup-without-Discord and disconnect recovery with an injected client.

frontend/src/main/discord-rpc.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+export function buildActivityPayload(
+	sessions: { status: string; isTerminated: boolean; createdAt?: string }[],
+	projects: { repo?: string }[],
+	startTime: number = Date.now(),
+): ActivityPayload | null {
+	const rep = pickRepresentativeStatus(sessions);
+	if (!rep) return null;
+	const repoUrl = projects.find((p) => p.repo && p.repo.startsWith("http"))?.repo;
```
whoisasx: [P2] Build the button from an API field that actually exists

GET /api/v1/projects returns ProjectSummary, which has no repo field, so this cast always produces undefined and the advertised View on GitHub button never renders. Avoid polling projects if the button is removed, or expose a deliberate safe repository-link field and associate it with the active session. Any future URL must be parsed and stripped of credentials and restricted for private/internal remotes before being published through Discord.

frontend/src/preload.ts
```diff
@@ -318,6 +325,19 @@ const api = {
 			ipcRenderer.invoke("keybindings:set", overrides) as Promise<KeybindingOverrides>,
 		setRecording: (active: boolean) => ipcRenderer.invoke("keybindings:setRecording", active) as Promise<void>,
 	},
+	rpc: {
```
whoisasx, line 337: [P2] Update every required bridge implementation

Adding rpc to the required AoBridge contract leaves both E2E bridge objects without this namespace. typecheck:e2e currently fails at e2e/support/fake-bridge.ts:196 and :614, which blocks both frontend test and renderer-smoke workflows before their test stages run. Add faithful RPC stubs to both fake bridges and cover the persisted toggle flow.

whoisasx (dismissed):
The opt-in Rich Presence direction is reasonable, and the focused RPC/settings tests pass, but this head is not ready to ship. The persisted consent state is not hydrated into the renderer, the Discord application identity and asset are not yet provisioned under maintainer control, connection failures do not recover, and the repository button is built from a field the projects endpoint does not return. Both frontend workflows are also currently red because the required RPC bridge was not added to the E2E fakes.

Please address the inline findings, add lifecycle and restart-persistence coverage, and verify the packaged integration against a maintainer-owned Discord application. Current result: Changes Requested.

codebanditssss (approved):

codebanditssss (approved):

frontend/src/main.ts
```diff
@@ -1259,6 +1275,7 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
 		portConfirmed = true;
 		stopDiscovery();
 		setDaemonStatus({ state: "ready", port });
+		setRpcDaemonPort(port);
```
whoisasx: [P1] Initialize RPC on every daemon-ready path

This call only runs from `reportBoundPort`, after AO spawns a new daemon. The earlier `inspectExistingDaemon` and direct-port attachment paths set the daemon status to ready and return without calling `setRpcDaemonPort`, leaving `apiClient` null. With persisted RPC enabled, Discord connects but every refresh exits at `if (!apiClient) return`, so Rich Presence never publishes after a quick reopen, with `AO_KEEP_DAEMON`, or when attaching to `ao start`. Route every ready path through a shared helper or set the RPC port in both attachment branches, and cover those paths.

frontend/src/main/discord-rpc.ts
```diff
@@ -0,0 +1,267 @@
...
+	} catch {
+		connectionState = "disconnected";
+		try {
+			await client.destroy();
+		} catch {
+		}
+		client = null;
+	}
+	broadcastStatus();
+	refreshTimer = setInterval(() => {
```
whoisasx, line 137: [P1] Do not restart after the user opts out

`startDiscordRpc` is launched outside `settingsOperation` during bootstrap. If the user disables RPC while `client.login()` is pending, `disposeDiscordRpc()` clears no timer and nulls the client, but this continuation still installs a timer and calls `refreshPresence()` after login settles. On rejection it immediately reconnects even though the persisted setting is now disabled. I reproduced this with a deferred login: disabling and then rejecting the original login caused a second `Client.login()` call. Track an enabled/generation state, capture the local client, reject stale continuations, and add this cancellation-race test.

frontend/src/main/discord-rpc.ts
```diff
@@ -0,0 +1,267 @@
...
+		void refreshPresence();
+	}, RPC_PRESENCE_REFRESH_INTERVAL_MS);
+	void refreshPresence();
+}
+
+async function reconnect(): Promise<void> {
+	if (!client && reconnectAttempts < MAX_RECONNECT_ATTEMPTS) {
```
whoisasx: [P2] Keep retrying while the feature remains enabled

After five failed attempts this condition remains false forever, although the refresh timer continues running. If AO starts before Discord and Discord is opened after that retry window, Rich Presence cannot recover until the user toggles it or restarts AO. Bound the backoff delay rather than the total lifetime attempts, or continue with a slower periodic probe, and test Discord becoming available after the initial retry window.

frontend/src/main/discord-rpc.ts
```diff
@@ -0,0 +1,267 @@
...
+}
+
+export function pickRepresentativeStatus(sessions: { status: string; isTerminated: boolean }[]): { label: string; count: number } | null {
+	const active = sessions.filter(
+		(s) =>
+			!s.isTerminated &&
+			!EXCLUDED_STATUSES.includes(s.status as SessionStatus) &&
+			!INACTIVE_STATUSES.includes(s.status as SessionStatus),
+	);
+	const activeAgentCount = active.filter((s) => ACTIVE_AGENT_STATUSES.includes(s.status as SessionStatus)).length;
```
whoisasx: [P2] Separate the displayed state from the active-agent count

Counting only `working` and `needs_input`, then converting every zero count to `Idle`, discards the representative status when review/CI states are the only sessions. A lone `ci_failed`, `changes_requested`, `review_pending`, `approved`, or `mergeable` session displays `Idle` despite `STATUS_PRIORITY` and the advertised behavior. Mixed sessions can also report `Orchestrating 1 agent` while taking `Fixing CI` from a different, uncounted session. Define count and state independently and add single-state plus mixed-state coverage.

frontend/vite.renderer.config.ts
```diff
@@ -111,6 +111,12 @@ export default defineConfig({
 	// to the daemon so the renderer can be tested against a running daemon from
 	// a plain browser without an Electron shell.
 	server: {
+		// Some development hosts have a low inotify watcher budget. Polling keeps
+		// Electron Forge startup reliable there, especially when agent-browser is
+		// present as a large watched tree.
+		watch: {
+			usePolling: true,
```
whoisasx: [P3] Keep global filesystem polling out of this feature

This unrelated setting makes every development host poll the complete renderer and product-ui trees, increasing CPU, battery, and filesystem activity. The low-inotify-budget case is host-specific, so please remove this from the Rich Presence PR or gate it behind an explicit development environment option in a separate change.

whoisasx (changes requested):
Two blocking lifecycle defects remain: attaching to an existing daemon never configures the sessions API client, and a pending Discord login can continue and reconnect after the user opts out. The reconnect lifetime and status/count semantics also need correction, and the unrelated global Vite polling change should be removed or isolated. Focused tests and current-main typechecks pass, but a cancellation-race regression test fails against this head. Please address the inline findings and add coverage for existing-daemon attachment and disabling during login. Current result: Changes Requested.

codebanditssss (approved):

i-trytoohard:
@bluwwi Marking as stale: this has been open 27 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

bluwwi:
last changes will be done by tomorrow 👍

PR #3953, opened 2026-08-13
fix(session): preserve drafts during interface drain

### LINKED ISSUE #3936: Claude Code TUI→Chat drain reliably fails with DRAIN_QUIESCENCE_UNVERIFIED: harness has no TerminalActivityDetector fallback
## Bug

Switching a `claude-code` harness session from Terminal UI to Chat UI with the `drain` policy frequently fails with:

```
AO could not verify that the terminal was idle after the latest input. The source interface was left untouched; retry after the terminal settles
```

(toast title "Interface switch failed", `error_code = DRAIN_QUIESCENCE_UNVERIFIED`)

**Source:** Interactive report via chat, 2026-08-13 | **Reported by:** @rugved | **Analyzed against:** `9b8395e13` (origin/main)
**Confidence:** High — traced exact mechanism through the drain loop and confirmed the missing capability directly in source
**Priority:** High — the primary chat-switch flow deterministically fails for the default harness under a common, ordinary interaction pattern

## Local Evidence

`~/.ao/data/ao.db`, `session_interface_transitions` table, last 24h on this machine:

| project | harness | phase | error_code |
|---|---|---|---|
| tractus | claude-code | failed | DRAIN_QUIESCENCE_UNVERIFIED |
| tractus | claude-code | failed | DRAIN_QUIESCENCE_UNVERIFIED |
| tractus | claude-code | failed | DRAIN_QUIESCENCE_UNVERIFIED |
| voice-orchestrator | claude-code | failed | DRAIN_QUIESCENCE_UNVERIFIED |
| tractus | codex | **completed** | — |

Every `claude-code` `tui → chat` drain attempt in this window failed with this code; the one `codex` attempt in the same window succeeded.

## Root Cause

`prepareSourceHandoff` in `backend/internal/session_manager/interface_transition.go` (added by the fix for [#3718](https://github.com/Untrivial-ai/agent-orchestrator/issues/3718)) treats an `ActivityIdle` row as **stale** — untrustworthy — whenever its `LastActivityAt` predates the last raw terminal input (`tuiIdleAfterInput`, line 621). This is correct: a composer can be touched without the agent emitting a fresh lifecycle event.

To recover from a stale-idle row without hanging forever, the loop falls back to an authoritative terminal-output probe:

```go
// interface_transition.go:568
if detector != nil && inputQuiet {
    output, outputErr := m.runtime.GetOutput(probeCtx, handle, interfaceTransitionOutputLines)
    ...
    idleProven = authoritative && state == domain.ActivityIdle
}
```

`detector` comes from an optional capability check at line 534-536:

```go
var detector ports.TerminalActivityDetector
if agent, ok := m.agents.Agent(rec.Harness); ok {
    detector, _ = agent.(ports.TerminalActivityDetector)
}
```

**The `claude-code` plugin (`backend/internal/adapters/agent/claudecode/`) does not implement `ports.TerminalActivityDetector`** — unlike `codex` (`backend/internal/adapters/agent/codex/terminal_activity.go`, which parses composer/footer markers from raw terminal output). Claude Code's activity state is derived purely from lifecycle hooks (`backend/internal/adapters/agent/claudecode/activity.go`): `ActivityIdle` is only written when the `stop` hook fires at end-of-turn (`DeriveActivityState`, case `"stop"`).

Because `detector` is always `nil` for `claude-code`, the recovery probe added by #3718 can never run for this harness. Any time a user's last terminal keystroke lands **after** the last `stop` hook — e.g. they typed or edited a partial follow-up message in the composer without submitting it, then decided to switch interfaces instead — `tuiIdleAfterInput` is false, `staleIdle` is true, and there is no way to prove quiescence. The loop then unconditionally burns the full `staleIdleLimit` (5s, line 18) and returns `errDrainQuiescenceUnverified` (line 27).

This is a gap left by #3718's fix rather than something that fix caused: #3718 correctly replaced an infinite hang with a bounded failure, but that bounded failure is only *recoverable* for harnesses that implement `TerminalActivityDetector`. Claude Code — arguably the most common harness — has none, so for it the "recovery window" #3718 added is unreachable dead code, and the bounded failure fires deterministically instead of being the rare edge case it is for `codex`.

## Reproduction

1. Open an idle `claude-code` TUI session (a turn has completed, hook-derived state is idle).
2. Type a few characters into the composer (do not submit).
3. Select "Switch to Chat UI" (drain policy).
4. After 5s, the switch fails with `DRAIN_QUIESCENCE_UNVERIFIED`, and the toast above appears. The TUI is left running untouched (no data loss), but the user must retry without touching the terminal beforehand.

## Fix Direction

Give the `claudecode` plugin a `DetectTerminalActivity` implementation (matching `codex`'s pattern — scan the last few terminal lines for Claude Code's composer/prompt-box markers) so the stale-idle recovery path in `interface_transition.go` has a fallback for this harness, the same way it does for `codex`.

## Impact

- Every `claude-code` `tui → chat` switch attempted shortly after any terminal keystroke (typed-then-abandoned draft, or hook-write latency) fails outright instead of switching.
- Workaround exists (wait for the terminal to be fully idle with no recent keystrokes before retrying) but is not discoverable from the error message, and ordinary usage (drafting then deciding to switch) triggers it routinely — 4/5 recent attempts on this machine failed this way.

## Related

- [#3718](https://github.com/Untrivial-ai/agent-orchestrator/issues/3718) — introduced the stale-idle recovery mechanism and the `DRAIN_QUIESCENCE_UNVERIFIED` bounded-failure path this issue's fix depends on.
- [#3879](https://github.com/Untrivial-ai/agent-orchestrator/issues/3879) — separate root cause (Intel Mac bundled Node signing) but the same `chat driver incompatible: ACP initialize... peer disconnected before response` failure mode also affects `claude-code` interface switches; this session hit both bugs back-to-back on the same machine.

illegalcall:
Fixes #3936

## Summary

- add a provider-owned terminal-surface observation seam that keeps work state and composer occupancy independent
- require repeated styled `idle + empty composer` proof before a Claude Code or Codex TUI is torn down, while retaining the durable/legacy fallback when styled capture is unavailable (including recovered ConPTY hosts)
- recognize Claude and Codex approval pickers by their current structural menu frame, not one exact sentence or stale transcript text
- report a pending provider decision as `DRAIN_DECISION_PENDING`, preserve the source, and offer an explicit “Cancel request and switch” recovery on desktop and mobile
- require a TUI→Chat target to replay settled native history before activation; Codex running/queued turns now return a retryable unsettled error instead of being silently omitted
- implement ACP's actual contract: `session/load` is typed transcript replay, while `session/resume` is context-only and cannot promise first-switch transcript continuity
- import replayed events idempotently before live projection, with bounded convergence retries and rollback to the source when replay is unavailable or remains unsettled

## Why

An approval picker is provider state, not user-authored composer text. Decision state therefore outranks draft classification, while ambiguous or stale frames remain fail-closed.

Likewise, a successful native resume does not prove that Chat received the transcript. ACP explicitly requires `session/load` to replay prior updates and explicitly forbids `session/resume` from doing so. Codex exposes settled history through `thread/read`; a running or queued turn is not a complete replay. The handoff now freezes source input, drains/stops one writer, loads settled provider history, projects stable event identities, and only then activates Chat. PTY scrollback remains a display artifact rather than canonical history.

The supporting primary-source comparison with Superset, ACP, Claude Agent ACP, Codex, and Gemini is in `docs/research/interface-handoff-connectors.md`.

## User-visible behavior

- Empty, settled composer: switch completes normally.
- Real unsent draft: source stays alive with `DRAIN_DRAFT_PRESENT`; explicit discard remains available.
- Approval/input picker: source stays alive with `DRAIN_DECISION_PENDING`; answer it in Terminal and retry, or explicitly cancel the request/turn and switch.
- Native history still running/queued: target waits for convergence; it never succeeds with a partial replay.
- ACP agent with `session/load`: replay is imported before Chat activates.
- ACP agent with only `session/resume`: ordinary daemon recovery still restores provider context, but a TUI→Chat handoff fails closed and restores Terminal rather than hiding history.

## Verification

- `cd backend && go test -race ./internal/service/chat ./internal/session_manager ./internal/daemon`
- `cd backend && go test -race ./internal/adapters/agent/claudecode ./internal/adapters/agent/codex ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/codexappserver`
- `cd backend && go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs` — 0 issues
- `cd frontend && npm test -- src/renderer/components/SessionInterfaceSwitch.test.tsx src/renderer/components/SessionView.test.tsx --run` — 71 passed
- `cd packages/mobile && npm run typecheck && npm test -- --run` — 453 passed
- focused regressions cover current/stale approval frames, non-first selections, decision-over-draft precedence, explicit provider interruption, Windows/no-styled-output fallback, Codex unsettled history, ACP load vs resume, convergence retry/timeout, rollback, and desktop/mobile recovery actions

## Related

- #3879 is a separate target-runtime packaging failure and remains intentionally out of scope.


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (changes requested):
Request changes

I found two blocking issues in the interface handoff flow.

**[P1] Approval prompts are misclassified as unsent drafts**

When the agent is paused on a permission prompt, switching from Terminal UI fails with `DRAIN_DRAFT_PRESENT`. The terminal surface contains the approval choices (`Yes / No`), but there is no user-authored draft to discard.

This appears related to `backend/internal/session_manager/interface_transition.go:601-610`, where `ComposerDraft` is treated as a draft whenever the durable activity state is idle. Approval prompts can temporarily have an idle durable activity row while the current surface is actually waiting for input.

The switch should recognize the waiting-for-approval state and either preserve the prompt during a safe drain or require an explicit interrupt without presenting it as “Discard draft.” The current action is misleading because it suggests user text will be destroyed when the visible content is actually an approval prompt.

**[P1] In-flight terminal output is lost during interface switching**

When switching while a long-running turn is streaming, output visible in Terminal UI is not consistently available in Chat UI. After switching back to Terminal UI, the same in-flight output can also disappear.

The target interface needs to receive or replay all output accepted before the source controller is stopped. At minimum, the handoff should preserve a durable snapshot or cursor of streamed output and replay it into the target interface. Current scrollback replay is insufficient for output that has not yet been committed to the native conversation.

Reproduction:

1. Start a long-running agent turn.
2. Wait until substantial output is streaming.
3. Switch from Terminal UI to Chat UI.
4. Observe that some output is missing.
5. Switch back to Terminal UI.
6. Observe that previously visible streamed output is missing there as well.

Expected behavior: output produced before the handoff remains visible in both interfaces, and an active approval prompt remains actionable or is explicitly cancelled.

illegalcall:
Addressed both blocking review issues in `fa7ddae0c`:

1. Approval prompts no longer become `DRAIN_DRAFT_PRESENT`.
   - Claude and Codex now anchor approval detection to the current structural menu frame, including changed question/footer wording and non-first selected options.
   - A typed `waiting_input` / `blocked` observation outranks composer draft classification and returns `DRAIN_DECISION_PENDING` with the source alive and input gate released.
   - Desktop and mobile expose an explicit `Cancel request and switch` interrupt path; the backend now sends the provider interrupt even when the stale durable row says idle.

2. Incomplete native output can no longer be silently accepted as a successful replay.
   - Codex `thread/read` returns `ErrChatHistoryUnsettled` for running/queued turns instead of skipping them.
   - The Chat target retries to a bounded settled boundary and imports stable provider events before live projection/activation.
   - ACP now follows the protocol distinction: `session/load` advertises and supplies replay; `session/resume` is context-only. A TUI→Chat handoff requires native replay and rolls back to Terminal if replay is unavailable or remains unsettled, while ordinary daemon recovery can still use resume-only context.

The no-styled-output/ConPTY fallback remains covered. All 11 PR checks are green, including the full Go race job on Linux and native CLI jobs on macOS, Linux, and Windows. Please re-review when convenient.

illegalcall:
@nikhilachale Addressed both blockers:

- Approval prompts are now treated as pending decisions, not unsent drafts, with an explicit cancel-and-switch option.
- Codex/ACP handoffs now wait for settled history and fail safely instead of silently dropping in-flight output.

Could you please re-check approval switching, long-running output handoff, and ACP `session/load` vs `session/resume` behavior? All CI checks are green.

nikhilachale (changes requested):
#### The previous approval-prompt and long-output issues appear fixed: switching now correctly presents “Finish work, then switch” and “Stop now and switch.”

###Now 

Image
Image

> **[P1] Do not mark drained handoffs as interrupted**
>
> The response output is preserved, but when choosing **Finish work, then switch**, Chat UI still displays `Request interrupted by user for tool use` and `No response requested`. Those markers are only appropriate for **Stop now and switch**. A completed/drained handoff should preserve the response and show the turn as completed rather than user-interrupted.

illegalcall:
Fixed and manually verified the remaining Claude busy-drain regression in 028db8775.

Root cause: Claude Code 2.1.232 moved `esc to interrupt` from the spinner row into the footer below the composer. AO therefore treated a visibly active `Computing…` frame as idle. ACP replay could also label a trailing user-only turn complete.

Now:
- Claude activity proof recognizes the structural spinner + bordered composer + active footer, while the same transcript text on an idle screen remains idle.
- ACP `session/load` rejects a trailing user-only replay as unsettled instead of projecting it as completed.
- If activity hooks show `No Signal`, a direct Chat click still uses the safe drain path and displays `Waiting to switch… Chat UI` until the provider settles.

Native E2E on Claude Code 2.1.232: switched during a running `openssl speed -seconds 4 sha256`; the transition stayed draining for 19.76s, then Chat contained the exact user prompt, completed command, and exact final reply. There were zero `Request interrupted by user for tool use` / `No response requested` markers.

Verification: full `go test ./...`, `npm run lint` (0 issues), frontend typecheck, and focused `go test -race` all pass.

@nikhilachale please re-check the same Finish work flow when Claude is visibly computing or running a tool.

nikhilachale:
Screenshot 2026-08-14 at 21 05 32
Screenshot 2026-08-14 at 21 05 25

What happened:

1. TUI → Chat UI initially did not transfer the existing conversation.
2. You sent a new message in Chat UI.
3. That created or synchronized the native Chat conversation.
4. After switching back, TUI received the message.
5. Subsequent switches worked.

Screenshot 2026-08-14 at 21 09 06
q


So the likely issue is that the initial TUI → Chat handoff does not seed or attach the target Chat controller with the existing native conversation/history. Once Chat has its own message, the conversation ID/state becomes established and later handoffs work.


> **[P1] Initial TUI → Chat handoff requires a Chat message before synchronization works**
>
> On the first TUI → Chat switch, the existing TUI conversation is not transferred. After sending a new message in Chat UI, switching back to TUI works, and subsequent switches work in both directions.
>
> This suggests the initial handoff does not correctly initialize the target Chat controller or native conversation ID from the existing TUI session. The handoff should work immediately without requiring the user to send a new Chat message first. If initialization cannot be completed, the switch should fail explicitly rather than presenting a fresh or incomplete context.

vanshx999:
@nikhilachale confirming this P1 on the Codex path, with a harder failure mode.

Fresh Codex session → first TUI → Chat switch fails outright with NATIVE_SESSION_MISSING / session: native conversation id unavailable for codex (HTTP 409). It never reaches the fresh-context state Claude shows
Screenshot 2026-08-14 at 11 19 46 PM
 it hard-errors before the switch completes.

Cause: on the first switch there's no materialized Codex rollout yet, so codex.go's NativeConversationExists probe rejects the reserved thread UUID (and/or no AgentSessionID has been reported). Same root class as the Claude case — the initial handoff doesn't seed the target Chat controller from the existing TUI session — it just aborts instead of silently dropping context.

Verified on 028db8775 (latest) via the dev daemon on :3002. So the initial-handoff gap is cross-adapter, not Claude-specific.

illegalcall:
@nikhilachale @vanshx999 Thanks — both first-handoff reports reproduced on `028db8775` and are addressed in `2035e0e92`.

Root causes:
- Dev/isolated daemon launches exported `AO_DATA_DIR` but not the owning daemon `AO_RUN_FILE`, so provider hooks could report the native conversation to a different/default daemon.
- AO stored a native conversation ID without recording which TUI runtime generation proved it. A failed/stale resume could therefore look transferable, while a truly promptless Codex session has no ID at all until its first turn.

Fix:
- Pin hook callbacks to the owning daemon run file.
- Persist the runtime generation that reported each native ID; stale and legacy IDs now fail closed as `NATIVE_SESSION_UNVERIFIED` until the current launch proves them.
- Permit an empty-ID first Codex handoff only when the styled provider frame positively proves the untouched initial composer. After terminal input is frozen and drain completes, AO resolves identity again, so a turn started in the admission race is transferred or the switch fails before stopping TUI.
- A controlled Chat → TUI resume binds the exact transferred provider ID to the new launch immediately, so the reverse switch does not require an extra message.
- Handle Codex constrained viewports where the footer is hidden, using the styled current-prompt marker without trusting plain transcript text.

Native desktop verification:
- Fresh promptless Codex: first TUI → Chat completed in about 1.3s to an empty Chat, then immediate Chat → TUI succeeded.
- Existing Codex history: first switch transferred both prompt/reply pairs exactly once; immediate round trips worked without sending a Chat message.
- Claude: first switch transferred the existing prompt/reply exactly once; round trip worked.
- Busy drain remained in TUI until the final reply, then Chat contained the completed turn exactly once.
- Unsent draft stayed in TUI; explicit discard switched without leaking the draft into history.

Automated verification is green locally: full Go test/lint (`0 issues`), build, vet, frontend typecheck/full tests, focused Go race tests, migration tests, and the new fail-closed admission-race regression test.

Please re-check the first TUI → Chat switch for both an existing conversation and a brand-new promptless Codex session.

illegalcall:
Follow-up: all 11 CI checks are now green on `9b5964a1f`, including the repository-wide Go race job and native CLI jobs on macOS, Linux, and Windows.

The first CI attempt exposed a test-only timing flaw: its 20 ms context could expire during SQLite setup under `-race`, before native-history convergence was exercised. The test now cancels only after the fake provider reports unsettled history; it passed 50 consecutive focused race runs plus the full Chat service race suite. Production behavior was unchanged by that follow-up.

vanshx999:
@illegalcall , @nikhilachale , Re-checked the first-handoff cases on 9b5964a1f:

- Fresh promptless Codex → first TUI → Chat completes to an empty Chat (~1s). ✅
- Existing Codex history → fresh session, sent a prompt in TUI and waited for the reply, then first TUI → Chat transferred the pair exactly once. ✅
- Busy drain → openssl speed -seconds 4 sha256, "Finish work, then switch": stayed in TUI until the final reply, then Chat showed the completed turn exactly once. ✅
- Unsent draft → draft stayed in TUI; "Discard draft and switch" dropped it without leaking into history. ✅

Note on my earlier "existing history" failure: that was a legacy session created before the fix, which fails closed as intended. A fresh session on the new code transfers cleanly.

Haven't re-verified Claude on my side (no existing Claude conversation on hand).

illegalcall:
@nikhilachale Vansh has independently re-checked the Codex first-handoff, busy-drain, and draft-recovery cases on `9b5964a1f`; all passed. I also re-verified the Claude first-history handoff and immediate round trip in the native desktop app, with the prompt/reply transferred exactly once. All 11 CI checks are green. Re-requesting your review when convenient.

illegalcall:
## Native Electron RCA and original-issue A/B (2026-08-17)

Correction to the earlier screenshot note: the old A/B pair exercised an ambiguity-only fail-safe edge. It did **not** prove the normal Claude failure from #3936. I reran the actual issue against the exact reported baseline and the current PR in separate isolated `~/.ao/dev` profiles.

### Actual root cause

Claude writes a durable idle fact when its stop hook fires. Any later raw terminal input makes that fact stale, correctly, because the input might be an unsent draft. At `9b8395e13`, Claude had no terminal-surface inspection capability, so AO could not re-prove quiescence even after the user erased the input and the real composer was visibly empty. It waited five seconds and returned `DRAIN_QUIESCENCE_UNVERIFIED`.

The fix gives Claude a provider-owned surface inspector. It keeps two facts independent:

- provider work: active, idle, blocked, or unknown
- composer: empty, draft, or unknown

AO only tears down Terminal after repeated `idle + empty` proof. Drafts, decisions, active work, and unknown frames remain fail-closed.

### Matched native reproduction

Both runs used the real Electron app, Go daemon, tmux runtime, and Claude Code `2.1.233`. A real Claude turn completed with `CLAUDE_RCA_READY`; I then sent a visible `x` through AO's `/mux` input path and erased it with Ctrl-H. The screenshots were captured only after tmux confirmed the current composer was empty.

#### A — before: exact issue baseline `9b8395e13`

The empty composer still fails after the five-second deadline with `DRAIN_QUIESCENCE_UNVERIFIED`; Terminal remains alive.

A — baseline falsely rejects a real empty Claude composer

#### B — after: `401ad23b0`

The same stale-idle/empty-composer state completes the switch in about 4.4 seconds, and Chat replays the original prompt/reply exactly once.

B — fixed build switches and replays the native Claude history

### Native happy and unhappy paths on the fixed build

- Empty composer after newer terminal input: completed TUI → Chat; history appeared exactly once.
- Real unsent `UNSENT_DRAFT`: immediate `DRAIN_DRAFT_PRESENT`; TUI, runtime, and draft all remained intact.
- Real Claude `/permissions` menu: immediate `DRAIN_DECISION_PENDING`; menu and TUI remained intact, with `Cancel request and switch` offered explicitly.
- Active turn (`sleep 12`): transition stayed in `draining`, did not interrupt Claude, and completed only after `CLAUDE_ACTIVE_DONE`; both turns replayed exactly once.
- Ambiguous surface: remains `DRAIN_QUIESCENCE_UNVERIFIED` and preserves the source. The earlier ambiguity screenshots remain as supplemental evidence in `docs/screenshots/pr-3953/`.

The native run also exposed and fixed a Claude `2.1.233` menu variant: `/permissions` keeps focus in a boxed search control instead of prefixing a numbered option with `❯`. Detection now requires the current boxed control, numbered option, and final navigation footer; a completed menu above a normal composer is explicitly covered as non-blocking transcript history.

### Fresh verification at `401ad23b0`

- Cross-package Claude stale-idle and decision regressions: 10/10 passes.
- Uncached touched Go packages: pass.
- Race detector for Claude adapter + session manager: pass.
- `npm run lint`: full backend suite passes; golangci-lint reports `0 issues`.
- `npm run frontend:typecheck`: pass.
- Interface-switch frontend suites: 2 files, 74 tests passed.

The isolated dev profile was restored after testing; no production AO state or unrelated Electron process was touched.

vanshx999:
## Independent verification + one new idempotency regression test

Re-checked the branch at `401ad23b0` locally (Go daemon, tmux runtime not needed for these) and ran the focused suites the PR claims. All green:

- `go test -race ./internal/adapters/agent/claudecode/... ./internal/adapters/agent/codex/... ./internal/adapters/agent/terminalui/...` — pass (surface inspectors incl. Claude 2.1.233 nav-menu / spinner / footer, Codex constrained viewport, unstarted-conversation proof)
- `go test -race -run TestInterfaceTransition ./internal/session_manager/...` — pass (draft vs decision precedence, stale-idle, no-styled-output fallback, recovered ConPTY host, promptless Codex admission race, rollback)
- `go test -race` on the chat handoff/resume/convergence tests and the full `./internal/service/chat/...` suite — pass
- Frontend `SessionInterfaceSwitch` / `SessionView` / `useSessionInterfaceTransition`: 77 tests pass
- Mobile `terminalInterfaceRecovery`: 4 tests pass

**New test:** I didn't find a functional regression in the current code, but I did find one genuinely untested branch of a headline safety claim. The PR promises *"import replayed events idempotently before live projection,"* and the reconciler enforces that, yet `reconcileNativeHistory` has no direct unit test (the `chat_test` external package can't reach it, which is likely why). A restart / retried TUI→Chat handoff against an already-imported thread must not duplicate durable rows.

I added:

```go
func TestReconcileNativeHistoryIsIdempotent(t *testing.T) // backend/internal/service/chat/reconcile_native_history_test.go
```

It seeds one completed native turn (user msg + assistant msg + command activity), re-reads the same settled thread under *different* provider item ids (the Codex `msg_…` vs thread/read `item-…` case the comment explicitly calls out), and asserts the re-import:
- does **not** re-project the duplicate user / assistant / command events, and
- **does** still project a genuinely new settled fact and the turn lifecycle events, all bound to the existing `native-turn-1` rather than a fresh id.

I validated it mutates-red: disabling the dedup at the final assembly boundary makes the test fail ("replayed user message was not de-duplicated"), and it passes with dedup intact. So it genuinely guards the idempotency property rather than passing vacuously.

Diff for the added file (33 lines of test + imports):

```diff
--- /dev/null
+++ b/backend/internal/service/chat/reconcile_native_history_test.go
@@ -0,0 +1,? @@
+package chat
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+func TestReconcileNativeHistoryIsIdempotent(t *testing.T) { ... }
```

Note: the on-disk file is currently an untracked add on the PR branch — the author (or a maintainer) can pick it up as-is; I did not force-push it to the PR branch. Happy to adjust the shape (e.g., split the "new fact survives" and "duplicate suppressed" assertions) if you'd prefer it stricter.

No blockers from my side. The two earlier [P1]s (approval prompts as drafts, in-flight output loss) and the first-handoff seed gap are all addressed in the current tree and the focused regressions cover the important branches.

nikhilachale (approved):
Tested on both claude and codex 
both are working fine
LGTM we can merge it

PR #3957, opened 2026-08-13
fix: keep terminal escape responsive for noisy chat

### LINKED ISSUE #3945: Chat UI freezes when long-running command exceeds output cap (cannot switch to Terminal)
## Bug

Running a long-lived, high-output command (e.g. `npm run dev` / vite dev server) inside a **chat-mode** session causes the UI to freeze — the session page becomes unresponsive and the user cannot switch back to the Terminal (TUI) tab.

The screenshot shows a vite dev server spamming `[vite] (client) page reload` messages as source files change. The output hit AO's 64KB cap (producing the orange "This command printed more than AO stores…" warning), but the process keeps running.

**Source:** Discord #bug-triage thread | **Reported by:** @prateek | **Analyzed against:** `8926ab5fd` (origin/main)
**Confidence:** High — traced exact code path

## Reproduction

1. Start a session in **chat mode** (e.g. Claude Code or Codex with chat interface)
2. Have the agent run a long-lived command that produces continuous output, e.g. `npm run dev` (vite), a file watcher, or a test runner in watch mode
3. Wait for the output to exceed 64KB (vite HMR reloads will do this quickly during active file editing)
4. Observe: the session page becomes unresponsive, and clicking the "Switch to Terminal" button either does nothing or opens a policy dialog that cannot complete

## Root Cause

Two independent problems combine to create the freeze:

### 1. Re-render storm from capped-but-still-active output

When a command's output exceeds `MaxCommandOutputChars` (64KB), the SQL query `AppendConversationActivityOutput` (`backend/internal/storage/sqlite/queries/conversations.sql:707-718`) stops appending text but **still bumps `revision` and `updated_at`** on every delta:

```sql
SET command_output = substr(command_output || sqlc.arg(delta), 1, sqlc.arg(max_output_chars)),
    command_output_truncated = CASE WHEN length(...) > max_output_chars THEN 1 ELSE ... END,
    revision = revision + 1,
    updated_at = sqlc.arg(updated_at)
```

Every revision bump fires a CDC event → SSE push → `event-transport.ts:95` invalidates the conversation query (debounced 150ms) → full conversation snapshot refetch → timeline re-render. A vite dev server in active development emits dozens of HMR messages per second, so this creates a near-continuous refetch-and-rerender cycle even though the visible text hasn't changed (it's capped).

### 2. Interface switch blocked while session is "busy"

`SessionView.tsx:301-308` computes `interfaceBusy`:

```tsx
const interfaceBusy = Boolean(
  session && (session.status === "working" ||
    session.activity?.state === "active" || ...)
);
```

Because the vite process is still running, `activity.state` stays `"active"`, so `interfaceBusy = true`. The switch button (`requestInterfaceSwitch`, line 327-334) then opens a policy dialog instead of switching directly. Even if the user picks "interrupt," the drain/interrupt flow may stall because the agent's turn (the long-running command) never completes on its own.

The result: the user is locked into a chat view they can't escape, while the terminal — where they need to go to Ctrl-C the runaway process or see the full output — is unreachable.

## Impact

- **Session unusable** — the page freezes; the only recovery is killing/restarting the daemon or the session externally
- **No workaround from within the UI** — the user cannot switch to TUI, cannot stop the command, cannot navigate away
- **Common trigger** — any agent that starts a dev server, file watcher, or test runner in watch mode hits this. These are extremely common operations.

## Fix

Two changes needed:

1. **Stop bumping `revision` when output is already truncated.** The `AppendConversationActivityOutput` query should skip the `revision = revision + 1` and `updated_at` write when `command_output_truncated = 1` AND the new concatenated length still exceeds the cap. This prevents CDC events for data that isn't changing. The atomic single-statement design can be preserved with a `WHERE command_output_truncated = 0` guard (or a CASE on the truncation flag).

2. **Allow interface switch (chat→TUI) even when the session is busy.** The switch from chat to terminal should not require the turn to complete — the terminal IS the escape hatch for a runaway command. Either remove the `interfaceBusy` gate for the chat→TUI direction, or add a "force switch" path that doesn't require the drain/interrupt policy.

## Related

- [#3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360) — Add immutable Chat/TUI session modes and structured orchestrator Chat

illegalcall:
## Summary

- stop capped command and activity streams from bumping revisions and emitting no-visible-change CDC invalidations
- make Chat to Terminal switches use interrupt semantics even when the session read model appears idle
- synchronously arm a reversible Chat dispatch fence before an accepted interface transition returns or target preflight starts
- after target preflight succeeds, cancel the entire durable Chat queue while dispatch is serialized, then interrupt the active provider turn without holding AO's dispatch lock
- reopen the source and resume its preserved queue when target preflight fails
- let interrupt handoffs proceed after provider cancellation without waiting indefinitely for `turn/completed`
- cover capped output, CDC stability, stale busy projections, missing completion events, queue-cancellation failures, clock movement, preflight recovery, and ACP/Codex completion ordering

## P1 follow-up

The review correctly found that the first implementation fenced queued work too late. `StartInterfaceTransition` returned before the background worker reached `PrepareChatHandoff`, leaving a window in which a completion could promote a queued turn.

The transition now has an explicit two-phase boundary:

1. Persist the transition.
2. Synchronously close Chat intake, automatic queue dispatch, and explicit queued-turn promotion.
3. Return the accepted transition and preflight the target in the background.
4. If preflight succeeds, settle every queued turn as interrupted under the dispatch lock.
5. Interrupt the active ACP/Codex provider turn outside that lock.

After step 2, completion events before, during, or after provider cancellation cannot call `SendTurn` for an accepted queued message. The fence is reversible: if target preflight fails, AO reopens Chat and explicitly resumes the untouched queue, including when the active turn completed while the fence was closed. If durable queue cancellation fails, the provider is not interrupted and the source is reopened.

## Verification

- `cd backend && go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs` with a clean isolated cache (0 issues)
- `cd frontend && npm run typecheck:e2e`
- `cd frontend && npm test` (182 files; 2,385 passed, 1 skipped)
- `cd backend && go test -race ./internal/service/chat -run 'TestChatHandoffInterrupt' -count=10`
- `cd backend && go test -race ./internal/session_manager -run 'TestInterfaceTransitionChatToTUI' -count=10`
- focused Chat interrupt loop 20x and Chat-to-TUI transition loop 10x
- `npm run sqlc`

Closes #3945


nikhilachale (dismissed):
[P1] Queued turn still executes during interrupt handoff" body="The interrupt handoff returns after calling `CancelQueuedTurns`, but clean manual tests with both Codex and Claude show that the active command is interrupted while the queued `touch /tmp/ao3945-queued-ran` message still executes. This violates the PR’s queue-cancellation requirement. The existing unit tests use controlled fake providers and do not reproduce the live provider/controller ordering that allows the queued turn to dispatch. The handoff must ensure that no accepted queued turn can be promoted or executed after a Chat-to-Terminal interrupt begins, with regression coverage for the observed runtime ordering.
Verdict: request changes.

Evidence from clean manual testing:

- Codex: the active command was interrupted, but the queued message executed.
- Claude: the active command was interrupted, but the queued message executed.
- `/tmp/ao3945-command-finished` was absent.
- `/tmp/ao3945-queued-ran` was present.
- The targeted tests passed 10/10, indicating that they do not cover the observed runtime ordering:

```text
go test ./internal/service/chat -run 'TestChatHandoffInterrupt' -count=10
ok
```

The output-limit warning was not visible because both providers buffered the command output, so that portion remains manually unverified. The queued-turn cancellation failure is independently blocking because it is an explicit requirement of PR #3957.

illegalcall:
Addressed the P1 in `d518ac255`.

The root cause was the asynchronous gap before `PrepareChatHandoff`, combined with cancelling queued rows only after the provider interrupt. The transition now synchronously arms the Chat handoff before returning or starting target preflight, cancels the complete durable queue while dispatch is serialized, and only then interrupts the ACP/Codex provider outside AO's dispatch lock.

I added regressions for the acceptance boundary, ACP- and Codex-style completion ordering, queue-cancellation failure, target-preflight failure, missing completion events, and clock rollback. The focused race/stress suites plus the full repository lint, frontend typecheck/test, backend build, and vet checks are green locally.

illegalcall:
Follow-up after merging current `main`: the final P1 fix is now a two-phase handoff in `404bd2163`.

- Acceptance synchronously installs a reversible Chat intake/dispatch/promotion fence.
- Target preflight runs while queued rows remain durable and untouched.
- Successful preflight cancels the whole queue under `sendMu`, then interrupts ACP/Codex outside the lock.
- Failed preflight reopens Chat and explicitly resumes the preserved queue, even if the active turn completed while fenced.
- A queue-cancellation failure happens before provider I/O and also reopens the source.

This supersedes my previous comment's statement that durable queue cancellation occurs during the initial arm. Keeping the arm reversible preserves work when Terminal cannot start while retaining the P1 guarantee that no queued turn can cross the provider seam after transition acceptance.

The branch is merged with current `main`; full Go tests, clean-cache lint (0 issues), race/stress handoff suites, backend build/vet, both frontend typechecks, and all 2,385 renderer tests pass locally.

illegalcall:
Fixed the P1: starting a Chat → Terminal switch now immediately fences the Chat queue. After Terminal preflight succeeds, AO cancels every queued turn before interrupting the active provider turn, so a completion event cannot release the next message.

@nikhilachale could you please retest the same Codex and Claude scenario?

1. Start a long-running command.
2. Queue `touch /tmp/ao3945-queued-ran` behind it.
3. Select **Open Terminal UI**.

Expected: the active command is interrupted, Terminal opens, and `/tmp/ao3945-queued-ran` is never created.

nikhilachale (approved):
Manual QA passed on the latest PR build (`404bd2163`).

Tested Chat → Terminal UI switching while a long-running command was active and another message was queued.

Results:

- Active command was interrupted.
- Queued message was cancelled.
- Terminal UI opened successfully.

Verification:

```text
PASS: command was interrupted
PASS: queued message was cancelled
```

The previously reported queue-cancellation failure is no longer reproducible on the latest branch.

PR #3959, opened 2026-08-13
fix: deduplicate project default reviewer

### LINKED ISSUE #3956: Reviewer dropdown duplicates the project-default harness
## Bug

The reviewer dropdown lists the project-default harness twice.

## Reproduction

1. Set any harness as the project default (for example, Claude Code).
2. Open the reviewer dropdown.

**Actual:** Both Project default (Claude Code) and Claude Code appear.

**Expected:** Show only Project default (Claude Code) for that harness. Keep the other harness options.

VenkataSakethDakuri:
## Summary

- label the run-panel choice as `Default` rather than `Project default`
- resolve the default reviewer from project configuration, compatible worker harness, then Claude Code fallback
- hide the resolved default harness from the standalone reviewer options
- normalize an equivalent saved override to the default selection
- remove the obsolete `cloud-agents` prototype and planning directory

## Root cause

The reviewer picker always appended every harness after the default option and the frontend assumed Claude Code whenever the project had no explicit reviewer. That diverged from the backend resolver for compatible workers such as Codex and OpenCode.

## Validation

- `git diff --check`

Closes #3956


neversettle17-101 (commented):
Thanks for the fix. Two behavior issues to address:

1. In the inspector, `projectDefaultHarness = config?.reviewers?.[0]?.harness || "claude-code"` does not match the backend resolver when the project has no explicit reviewer. The backend falls back to the worker-compatible reviewer first (`codex` worker -> `codex`, `opencode` -> `opencode`, etc.) before `claude-code`. As written, a codex worker with no project reviewer will show/exclude `claude-code`, while the actual default reviewer is `codex`, so the duplicate can remain and the label can be wrong. Please use the resolved reviewer default (for example the reviews response `reviewerHarness`) or mirror the backend resolver.

2. The run-panel dropdown should not say “Project default”. In this context the empty selection means use the worker/session default reviewer, not necessarily a project-configured default. Please make the label reflect the worker default/resolved reviewer, and only show that harness once.

VenkataSakethDakuri:
@neversettle17-101 I'm thinking of removing the default label altogether because the user might be confused whether it's resolved from the project settings or if it is resolved from any other place.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -808,6 +808,20 @@ type ReviewerHarness = NonNullable<components["schemas"]["TriggerReviewRequest"]
 type AgentInfo = components["schemas"]["AgentInfo"];
 type AgentCatalog = { supported?: AgentInfo[]; installed?: AgentInfo[]; authorized?: AgentInfo[] };
 
+const WORKER_DEFAULT_REVIEWERS: Partial<Record<WorkspaceSession["provider"], ReviewerHarness>> = {
+	"claude-code": "claude-code",
+	codex: "codex",
+	opencode: "opencode",
+	muse: "muse",
```
neversettle17-101, line 815: we have implemented a lot more reviewer harness this doesnt seem to be the whole list

neversettle17-101 (changes requested):

frontend/src/renderer/components/SessionInspector.tsx
VenkataSakethDakuri: The other harnesses can't be inherited automatically from the worker.

neversettle17-101 (changes requested):
Thanks for the update — the “Project default” wording and worker-compatible resolver are moving in the right direction, but the latest implementation regresses override behavior.

`effectiveReviewerHarness = reviewerOverride || resolvedDefaultHarness` is passed as both `defaultHarness` and `excludedHarness`, while the picker uses `selectedValue = value && value !== excludedHarness ? value : "__default__"` and `showDefaultOption={false}`. Once the user selects any explicit override, that override becomes `excludedHarness`, so the menu hides the currently selected reviewer and coerces the value back to the default sentinel. Example: worker default is `codex`, user selects `claude-code`; the next render excludes `claude-code`, shows the trigger via the fallback path, and there is no selected `claude-code` option in the menu. It also means there is no way to clear the override back to the worker default because the default row is hidden.

The excluded harness should be the resolved worker/project default, not the effective reviewer after applying an override. The picker should still offer the resolved default harness exactly once in a way that clears the override (without labeling it “Project default”), and then list the other reviewer options. Explicit overrides must remain visible/selectable when selected.

VenkataSakethDakuri:
@neversettle17-101 Any reason why we should keep the selected harness in the dropdown as well ? Isnt it redundant?

neversettle17-101:
@VenkataSakethDakuri this is how it should look like 
Screenshot 2026-08-14 at 8 06 39 PM

neversettle17-101:
Addressed the override regression and pushed 8bed37f76. The picker now excludes only the resolved worker/project default, keeps the explicit override in the menu, and exposes the resolved default row to clear the override. Added a regression test covering codex default → opencode override → codex reset.

Validation: `npm --prefix frontend test -- src/renderer/components/SessionInspector.test.tsx` (80 passed), `npm run frontend:typecheck`, `git diff --check`.

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

axisrow:
Nice fix. Default reviewer still shows raw id (codex) not Codex — want a fix for this #3915?

PR #3970, opened 2026-08-13
fix(chat): enable Codex switch-to-chat for fresh TUI sessions

bluwwi:
Codex switch-to-chat button was disabled when the TUI session had no prompt yet (`NATIVE_SESSION_MISSING`).

## Root cause

`nativeConversationID()` hard-blocked with `ErrNativeConversationMissing` when `agentSessionId` was empty — even though Codex implements `AgentInterfaceHandoffHistoryProbe`, which can distinguish a fresh TUI from a real conversation and decide fresh-start vs resume.

## Fix

If the adapter implements `AgentInterfaceHandoffHistoryProbe`, allow an empty native id through. `persistedNativeConversationID()` then uses the probe to decide fresh-start vs resume. Adapters without the probe keep the hard block.

## Tests

- `TestInterfaceTransitionStatusAllowsFreshStartWithHistoryProbe` — empty native id + probe → `Supported=true`
- `TestInterfaceTransitionStatusBlocksFreshStartWithoutHistoryProbe` — empty native id + no probe → `NATIVE_SESSION_MISSING`

### DIFF
--- a/backend/internal/session_manager/interface_transition.go
+++ b/backend/internal/session_manager/interface_transition.go
@@ -362,6 +362,14 @@ func (m *Manager) runInterfaceTransition(
 	}
 }
 
+// nativeConversationID resolves the adapter's native conversation id for the
+// session's current interface. An empty id is only safe to pass through when
+// the adapter can distinguish a fresh TUI from a real conversation (it
+// implements AgentInterfaceHandoffHistoryProbe) AND the session has no
+// hook-derived conversation history. If any conversation metadata is set but
+// the native id is empty, the hooks fired but failed to capture the id — a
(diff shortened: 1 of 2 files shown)


neversettle17-101:
Review note: checking only “adapter has history probe” may be too broad unless we’re confident an empty native ID is always safe for those adapters. A stronger condition would also check that the session has no prompt/latest user prompt/transcript/native history before allowing a fresh-start transition.

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

PR #3971, opened 2026-08-14
Add reviews tab to session inspector

codebanditssss:
Moves review controls, history, and external feedback into a dedicated Reviews tab. Keeps Summary focused on pull requests, session controls, and activity.

image
image

### DIFF
--- a/frontend/src/renderer/components/SessionInspector.tsx
+++ b/frontend/src/renderer/components/SessionInspector.tsx
@@ -28,6 +28,7 @@ import {
 	Play,
 	Trash2,
 	Loader2,
+	MessageSquare,
 	X,
 } from "lucide-react";
 import type { components } from "../../api/schema";
@@ -70,7 +71,7 @@ type OpenReviewerTerminal = (target: { handleId: string; harness: string }) => v
 
 export type { InspectorView } from "@aoagents/product-ui";
 
-const VIEW_DEFS: { id: InspectorView; labelKey: "inspector.summary" | "inspector.browser" | "inspector.files"; icon: ReactNode }[] = [
+const VIEW_DEFS: { id: InspectorView; labelKey: "inspector.summary" | "inspector.reviewTab" | "inspector.browser" | "inspector.files"; icon: ReactNode }[] = [
 	{
 		id: "summary",
 		labelKey: "inspector.summary",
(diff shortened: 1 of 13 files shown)


frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -1447,6 +1462,7 @@ function ReviewPanel({
 					</TooltipProvider>
 				) : null}
 				<div className="review-run-controls-container min-w-0 divide-y divide-border/70">
+					<AutoInjectReviewPolicyControl session={session} />
```
Annieeeee11, line 1471: **[P2] Auto-inject review is unreachable with zero PRs**

This control moved into `ReviewPanel`, but `ReviewPanel` still early-returns when `sortedPRs(session).length === 0` (above), so the Reviews tab only shows “No pull request opened yet” and this switch never mounts.

Previously it lived in Session controls and worked before a PR existed (CI still does). The new “keep Reviews tab with no PRs” test only asserts the tab exists.

Please render the auto-inject control outside/before that early return (or keep it in Session controls) so users can still disable injection before the first PR.

frontend/src/renderer/i18n/es.json
```diff
@@ -268,6 +268,7 @@
 	"inspector.earlierPass": "ejecución anterior",
 	"inspector.previousLabel": "Anterior:",
 	"inspector.reviews": "Resumen de revisiones",
+	"inspector.reviewTab": "Reseñas",
```
Annieeeee11: **[P3] Wrong Spanish label for code reviews**

`Reseñas` reads as product/restaurant reviews. Prefer `Revisiones` to match `inspector.reviews` (“Resumen de revisiones”).

Annieeeee11 (changes requested):
## Review summary

Clean tab split overall. One regression to fix before merge: auto-inject review becomes unreachable with zero PRs after moving into `ReviewPanel`.

- **P2** — auto-inject review gated behind no-PR early return
- **P3** — Spanish tab label should be `Revisiones`, not `Reseñas`

frontend/src/renderer/components/SessionInspector.tsx
codebanditssss: Fixed in 47389f2e1. The Reviews tab now renders the auto-inject review control even when the session has no PRs, and the zero-PR test verifies both the switch and empty state. The full inspector suite passes (79 tests) and TypeScript typecheck passes.

frontend/src/renderer/i18n/es.json
codebanditssss: Fixed in 47389f2e1. Updated the Spanish tab label from “Reseñas” to “Revisiones” to match the existing review terminology.

Annieeeee11 (approved):
lgtm

PR #3972, opened 2026-08-14
fix: contain provider error rows in the chat column

Annieeeee11:
## Summary
- Render `activityKind: "error"` as a dedicated wrapping card instead of a non-shrinking generic label.
- Unwrap provider JSON envelopes (`message` + `additionalDetails`) so reconnect/credit failures stay readable and compact.
- Keep the chat timeline from horizontal overflow (`min-w-0`, `overflow-x-hidden`, truncating generic labels).

## Test plan
- [ ] Reproduce a provider error / reconnect item (or fixture) and confirm the chat column no longer scrolls sideways.
- [ ] Confirm JSON envelopes show a short headline + detail, not `codexErrorInfo` blobs.
- [ ] `cd frontend && npm run test -- --run src/renderer/components/chat/ChatProviderSignal.test.tsx`

## Before
image

https://github.com/user-attachments/assets/bf221361-dcaa-4fa1-a179-86a213ca34ac

## After
image

### DIFF
--- a/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
+++ b/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
@@ -1257,15 +1257,17 @@ function RerouteRow({ activity }: { activity: ConversationActivity }) {
  * Rendering that as a generic activity label made one long unbreakable line that
  * widened the chat column under the sidebars, and five reconnect attempts painted
  * five walls of red JSON. This row unwraps the human parts and always wraps inside
- * the column.
+ * the column. New rows are normalized at the daemon; this path is the fallback
+ * for already-persisted envelopes.
+ *
+ * Not a live region: the enclosing timeline is already a `role="log"`, and the
+ * controller banner announces a terminal failure. Marking every historical
+ * reconnect row as `role="alert"` would interrupt a screen reader once per attempt.
  */
 function ErrorActivityRow({ activity }: { activity: ConversationActivity }) {
 	const { headline, detail } = providerErrorCopy(activity);
 	return (
-		<div
-			role="alert"
-			className="flex min-w-0 max-w-full items-start gap-2.5 overflow-hidden rounded-md border border-destructive/40 bg-surface px-3 py-2"
-		>
+		<div className="flex min-w-0 max-w-full items-start gap-2.5 overflow-hidden rounded-md border border-destructive/40 bg-surface px-3 py-2">
 			<AlertTriangle aria-hidden="true" className="mt-0.5 size-3.5 shrink-0 text-destructive" />
 			<div className="flex min-w-0 flex-1 flex-col gap-0.5">
 				<strong className="wrap-anywhere text-[11px] font-medium leading-snug text-destructive">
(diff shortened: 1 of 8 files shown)


frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1244,6 +1250,75 @@ function RerouteRow({ activity }: { activity: ConversationActivity }) {
...
+	if (!raw) return { headline: "Provider error" };
+
+	const unwrapped = unwrapProviderErrorJson(raw);
+	if (unwrapped) return unwrapped;
+
+	return { headline: raw };
+}
+
+function unwrapProviderErrorJson(raw: string): { headline: string; detail?: string } | undefined {
+	if (!raw.startsWith("{")) return undefined;
```
whoisasx: [P1] Parse the payload AO actually stores. The captured value shown in this PR starts with `provider error: {…}`, so this guard rejects it and the advertised unwrapping never runs; the PR's own After screenshot still contains `codexErrorInfo` and the raw JSON. The test passes only because its fixture begins directly with `{`. Please parse the JSON suffix safely and add the real prefixed payload as the regression fixture.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1244,6 +1250,75 @@ function RerouteRow({ activity }: { activity: ConversationActivity }) {
...
+ * dump. Falls back to the stored text when it is already plain language.
+ */
+export function providerErrorCopy(activity: ConversationActivity): {
+	headline: string;
+	detail?: string;
+} {
+	const raw = String(activity.detail?.message ?? activity.summary ?? "").trim();
+	if (!raw) return { headline: "Provider error" };
+
+	const unwrapped = unwrapProviderErrorJson(raw);
```
whoisasx: [P2] Keep provider-envelope normalization at the daemon boundary. This decoder teaches only the Electron renderer about Codex's `error.message` and `additionalDetails` shape; the mobile `ErrorActivity` still renders `summary`/`detail.message`, so it continues showing the raw envelope. Please normalize the Codex payload in the Codex adapter into provider-neutral summary/detail fields. A renderer fallback can remain for already-persisted rows.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1244,6 +1250,75 @@ function RerouteRow({ activity }: { activity: ConversationActivity }) {
...
+ * Rendering that as a generic activity label made one long unbreakable line that
+ * widened the chat column under the sidebars, and five reconnect attempts painted
+ * five walls of red JSON. This row unwraps the human parts and always wraps inside
+ * the column.
+ */
+function ErrorActivityRow({ activity }: { activity: ConversationActivity }) {
+	const { headline, detail } = providerErrorCopy(activity);
+	return (
+		<div
+			role="alert"
```
whoisasx: [P2] Avoid making every historical reconnect row an assertive alert. A five-attempt reconnect storm inserts five `role="alert"` regions, while the enclosing timeline already has a live `role="log"` and the controller banner announces the terminal failure. Please keep timeline history non-live, or announce only the newest active failure, so screen-reader users are not repeatedly interrupted.

whoisasx (changes requested):
The width containment is effective and the branch is mergeable with current main. All frontend, renderer-smoke, and gitleaks checks pass. However, the primary unwrapping behavior does not work against the actual captured payload, and the provider-specific normalization currently bypasses the daemon boundary and leaves mobile inconsistent. Please address the inline findings before merging.

Annieeeee11:
> The width containment is effective and the branch is mergeable with current main. All frontend, renderer-smoke, and gitleaks checks pass. However, the primary unwrapping behavior does not work against the actual captured payload, and the provider-specific normalization currently bypasses the daemon boundary and leaves mobile inconsistent. Please address the inline findings before merging.

Left Codex error envelopes as the daemon already persists them (provider error: {…}). Unwrapping stays in the Electron renderer so reconnect/credit rows stay compact without teaching the Codex message / additionalDetails shape at the adapter or controller. Historical rows and new rows share that path. Mobile still shows the raw envelope.

whoisasx (approved):

PR #3975, opened 2026-08-14
fix: drop leftover chat strip controls to match terminal

Annieeeee11:
## Summary
- Remove the leftover chat tab-strip controls (new terminal, font size, fullscreen) so chat matches the terminal session chrome
- Keep new terminal in the session actions on the right, same as terminal sessions
- Cover the strip with a ChatWorkspace unit test

## Test plan
- [ ] New Chat UI session: tab strip shows only the chat tab (no +, 12px, or fullscreen)
- [ ] New Terminal session: strip still has no display controls
- [ ] Add a terminal from a Chat UI session: chrome does not change to hide leftover chat-only controls
- [ ] Worker new-terminal action still works from the right-side session actions
- [ ] `npx vitest run src/renderer/components/chat/ChatWorkspace.test.tsx src/renderer/components/SessionView.test.tsx`

## Before
image

## After
image

### DIFF
--- a/frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
+++ b/frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
@@ -193,28 +193,21 @@ describe("ChatWorkspace timeline", () => {
 		);
 	});
 
-	it("keeps chat font controls scoped to the chat instead of native page zoom", () => {
-		render(<ChatWorkspace snapshot={chatFixture} />);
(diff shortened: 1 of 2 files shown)


whoisasx (approved):
Looks good to me. The duplicated chat-strip controls are removed while the shared right-side New terminal action, controller recovery path, and chat/reviewer navigation remain intact. The focused regression coverage is appropriate, the resulting header matches the terminal structure, and all current checks are passing.

PR #3976, opened 2026-08-14
fix: reset model scope across agent switches

nikhilachale:
## What

Reset model and mode overrides when switching a conversation between agent harnesses.

- Clear the source harness model/mode before resolving the target launch configuration.
- Persist the switched conversation without retaining stale overrides.
- Reset renderer controls and cached catalog state so the target harness supplies its own defaults.
- Add backend, storage, hook, and dialog regression coverage for both switch directions.

## Why

A model selected for one agent could be reused after switching to a different agent. That stale model might not exist in the target harness and could make the next request fail.

## How

Agent switching now establishes a fresh model scope. The backend clears source overrides and resolves the target harness configuration independently. The renderer mirrors that invariant by resetting selected model/mode state and refetching the target catalog after a successful switch. Sends omit the model when the target is using its default.

The SQLite query was updated and its generated code regenerated through the repository workflow.

## Testing

- `npm run lint`
- Manual Electron verification: Codex → Claude and Claude → Codex, confirming the model resets and `hi` succeeds without a cross-agent model error.

## Checklist

- [x] Continuing an existing PR branch (`codex/agent-switching-ux`)
- [x] One focused change; no related issue supplied
- [x] Relevant backend and frontend checks pass


backend/internal/session_manager/agent_switching.go
```diff
@@ -1023,9 +1023,6 @@ func (m *Manager) prepareTargetActivation(ctx context.Context, store ports.Agent
 		return preparedTargetActivation{}, fmt.Errorf("system prompt file: %w", err)
 	}
 	config := effectiveAgentConfig(rec.Kind, project.Config)
```
illegalcall, line 1103: This still applies the project/role model and mode without checking which harness they belong to. The settings UI treats those values as harness-specific and clears them when the configured harness changes, so a Claude `opus` configuration can still be forwarded to Codex (or vice versa)—the failure this PR is meant to prevent. Could target resolution retain model/mode only when the configured harness matches `sw.TargetHarness`, while preserving harness-neutral permissions?

backend/internal/session_manager/agent_switching.go
```diff
@@ -391,7 +391,7 @@ func (m *Manager) executeAgentSwitch(ctx context.Context, admitted *admittedAgen
 	// Resolve credentials, native-resume evidence, and launch commands before
 	// asking the source to spend a model turn. This preflight does not install
 	// target workspace files or reserve a target generation in durable storage.
-	target, err := m.prepareTargetActivation(ctx, store, rec, project, targetAgent, targetCapabilities, result, cfg.Model)
+	target, err := m.prepareTargetActivation(ctx, store, rec, project, targetAgent, targetCapabilities, result)
```
illegalcall: Could we preserve an explicit target-model selection here? `SwitchAgentRequest.model` is documented as an override for the target launch/resume, and the dialog sends it when the user chooses a non-default model. After removing `cfg.Model` from this call, that value is only normalized/fingerprinted and never applied, so the switch can report success while launching a different model. Please distinguish the stale source setting from the explicit target override.

nikhilachale:
harness-specific configuration:

> Fixed in `eec0310f4`. Target activation now retains configured model/mode only when the role’s configured harness matches `sw.TargetHarness`. Harness-neutral permissions remain preserved. Added regression coverage proving a source-harness mode/model is dropped while permissions survive the switch.

explicit target model:

> Fixed in `eec0310f4`. `SwitchAgentRequest.model` is now passed into target activation and applied as an explicit target model override. This is separate from stale project/role settings, which are filtered by target harness. Added coverage confirming the explicit target model wins for both fresh launches and the shared activation path.

illegalcall (approved):

PR #3978, opened 2026-08-14
fix: match inspector open/close to sidebar spring

Annieeeee11:
## Summary
- Replace the inspector’s CSS flex-grow/fade collapse with the same Motion spring pattern as the left sidebar (gap width + slide).
- Keep drag-to-resize clamped at min width; collapse stays on the explicit toggle / ⌘⇧B.
- Update browser-view measuring and unit/e2e coverage for the new rail.

## Test plan
- [ ] Open a worker session; toggle inspector with the topbar button and ⌘⇧B — motion should feel like the left sidebar
- [ ] Drag the inspector edge: stops at min width, does not auto-collapse
- [ ] Drag the collapsed rail back open
- [ ] Browser tab native view stays clipped to the sliding rail during open/close
- [ ] `cd frontend && npx vitest run --config vite.renderer.config.ts src/renderer/components/SessionView.test.tsx`


ronishrohan (commented):
## Review

Implementation is architecturally sound — correctly mirrors the sidebar spring pattern and the 544-line rrp deletion is justified. Two non-blocking suggestions:

### 1. `INSPECTOR_SPRING_MS` is a hardcoded approximation
The spring `{ stiffness: 420, damping: 40, mass: 0.6 }` has no fixed duration — `400ms` is an estimate. The `settledClosed` timer could fire before the spring settles under browser load, snapping the panel to `hidden` while still partially visible. Consider using Motion's `onAnimationComplete` callback instead of `setTimeout`, or add a ~50ms buffer.

### 2. Duplicated spring constants
The spring config is copied from `ui/sidebar.tsx`. If someone changes the sidebar spring, the inspector won't follow. Consider extracting to a shared constant.

### Cross-PR note
This PR is structurally incompatible with #3979 — both touch `SessionView.tsx` but #3979 fixes the cold-start bug within rrp, while this PR removes rrp entirely. If this merges first, #3979 becomes unnecessary since `useResizable` handles the cold-start case correctly via `useLayoutEffect`.

ronishrohan (changes requested):
## Changes Requested

### 1. Replace `setTimeout` with `onAnimationComplete`
`INSPECTOR_SPRING_MS` (400ms) is a hardcoded approximation of a spring with no fixed duration. The `settledClosed` timer could fire before the spring settles under load, snapping the panel to `hidden` while still partially visible. Use Motion's `onAnimationComplete` callback on the gap/container `motion.div` instead of `setTimeout`.

### 2. Extract shared spring constants
The spring config `{ stiffness: 420, damping: 40, mass: 0.6 }` is duplicated from `ui/sidebar.tsx`. If someone changes the sidebar spring, the inspector won't follow. Extract to a shared constant.

### Cross-PR note
Structurally incompatible with #3979 — if this merges first, #3979 becomes unnecessary since `useResizable` handles cold-start correctly.

ronishrohan (approved):
## Approved

Both requested changes from the prior review have been addressed:

1. **`setTimeout` replaced with `onAnimationComplete`** — the `settledClosed` timer now fires from Motion's callback when the spring actually settles, not after an arbitrary 400ms.
2. **Shared spring constants** — `SHELL_PANEL_SPRING` extracted to `lib/motion-spring.ts`, imported by sidebar, SessionView, and ShellTopbar.

The remaining `INSPECTOR_SPRING_MS` usage is for CSS variable transitions on adjacent UI elements (topbar action labels) — not the panel visibility state that was flagged.

Tested in the desktop app. The `dimensions` error during animation is a pre-existing xterm race, not introduced by this PR.

PR #3979, opened 2026-08-14
fix: clamp session inspector drag floor on cold start

Annieeeee11:
## Summary
- Stop the session inspector (chat and terminal) from resizing to zero on cold start.
- `react-resizable-panels` was deriving `minSize: 0` before the split had a real width; changing the prop alone was not enough, which is why collapse/expand appeared to “fix” it.
- After the first real measure, remount the inspector once with a non-zero floor (~360px / 30%, clamped to 50% on narrow splits), refuse to persist crushed widths, and keep pixel size across window resize.

## Test plan
- [ ] Fresh app start → open a worker session → drag the inspector hard right on first load (before any toggle); it stops at the floor, not a sliver
- [ ] Collapse via the inspector toggle, reopen, drag again; floor still holds
- [ ] Repeat in both Chat UI and terminal mode
- [ ] Narrow the window so 50% < 360px; floor follows the 50% max
- [ ] `cd frontend && npx vitest run src/renderer/components/SessionView.test.tsx`

## Before
image

## After
image

### DIFF
--- a/frontend/e2e/resize-clamp.spec.ts
+++ b/frontend/e2e/resize-clamp.spec.ts
@@ -59,7 +59,7 @@ test("inspector drag stops at minSize instead of collapsing; buttons still toggl
 	const y = separatorBox.y + separatorBox.height / 2;
 
 	// Drag the separator all the way to the right edge: the rail must clamp at
-	// its 30% minSize, not snap away.
+	// its 360px floor (or 50% of a narrow split), not snap away or crush below.
 	await dragPointer(
 		page,
 		{ x: separatorBox.x + separatorBox.width / 2, y },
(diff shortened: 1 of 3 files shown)


ronishrohan (commented):
## Review

Clean, well-scoped fix. No blocking issues found.

- The epoch/remount pattern is the right workaround for rrp deriving `minSize: 0` before the group has a real width.
- `inspectorConstraintsHydratedRef` ensures exactly one remount per inspector lifecycle and correctly resets on navigation.
- The `ResizeObserver` in `useLayoutEffect` fires synchronously, so the `groupWidth <= 0` guard correctly skips the zero-width frame. No race with the imperative API.
- Rapid window resizing is handled correctly (updates percentage, doesn't re-trigger epoch).
- Tests are solid — 3 new unit tests cover cold-start guard, measure-remount path, and crushed-width persistence guard. E2e updated to check absolute pixel floor.

**Note:** This PR may be superseded by #3978, which removes rrp entirely (eliminating the constraint registration race this fixes). If #3978 merges first, this can be closed.

PR #3985, opened 2026-08-14
fix(boot): resilient first-run, tmux bundling, and daemon boot classification

### LINKED ISSUE #2206: Electron install failed on initial setup in Windows
## Summary

Not able to run the frontend we do some configuration change in the `vite.preload.config.ts`

## Logs

```
✔ Launched Electron app. Type rs in terminal to restart main process.

 WARN  inlineDynamicImports option is deprecated, please use codeSplitting: false instead.
 WARN  inlineDynamicImports option is deprecated, please use codeSplitting: false instead.
7:32:54 pm [vite] (client) [optimizer] scanning dependencies...
7:32:58 pm [@electron-forge/plugin-vite] target built src/annotate-preload.ts
7:32:58 pm [@electron-forge/plugin-vite] target built src/preload.ts
7:33:00 pm [@electron-forge/plugin-vite] target built src/main.ts

An unhandled rejection has occurred inside Forge:
Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
at getElectronPath (C:\Desktop\agent-orchestrator\frontend\node_modules\electron\index.js:17:11)
    at Object.<anonymous> (C:\Desktop\agent-orchestrator\frontend\node_modules\electron\index.js:21:18)
    at Module._compile (node:internal/modules/cjs/loader:1871:14)
    at Object..js (node:internal/modules/cjs/loader:2002:10)
    at Module.load (node:internal/modules/cjs/loader:1594:32)
    at Module._load (node:internal/modules/cjs/loader:1396:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1617:12)
    at require (node:internal/modules/helpers:153:16)
    at locateElectronExecutable (C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\core\dist\util\electron-executable.js:12:28)
    at async forgeSpawn (C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\core\dist\api\start.js:126:32)
    at async forgeSpawnWrapper (C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\core\dist\api\start.js:160:25)
    at async C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\core\dist\api\start.js:197:21
    at async C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\tracer\dist\index.js:58:20
    at async C:\Desktop\agent-orchestrator\frontend\node_modules\@electron-forge\cli\dist\electron-forge-start.js:59:21  
[ELIFECYCLE] Command failed with exit code 1.

```

## Proposed Fix

change the configuration of `vite.preload.config.ts`

```
import { defineConfig } from "vite";

// Forge's VitePlugin handles all preload script build configuration.
export default defineConfig({
  build: {
    minify: false,
    chunkSizeWarningLimit: 1000,
    rollupOptions: {
      // Disable tree shaking (it's not working for Electron modules anyway)
      treeshake: false,
      // Prevent code splitting for Electron preload scripts
      output: {
        manualChunks: undefined,
        inlineDynamicImports: true,
      },
    },
  },
});

```

## Steps to fix this issue [for future contributors, specially windows]

1. remove the electron from node_modules folder
```
rm -rf node_modules/electron
```

2. add electron with force now 
```
pnpm add electron --force
```
3. change the configuration of `vite.preload.config.ts`  according to proposed fix shared above.

### LINKED ISSUE #2427: can't start
```bash
daskomputer@daskomputer-MS-7C96:~/src$ ao start
download https://github.com/AgentWrapper/agent-orchestrator/releases/latest/download/agent-orchestrator-linux-x64.AppImage: unexpected status 404 Not Found
```

### LINKED ISSUE #2443: Bundle tmux in the macOS/Linux installers so sessions work on a clean install
## Summary

On macOS and Linux, every worker/orchestrator session requires `tmux` in `PATH`,
but **none of the installers ship or declare it**. A user who installs the
packaged app on a clean machine cannot spawn a single session — the daemon fails
the pre-spawn prerequisite gate with:

```
spawn: runtime: prerequisite missing: tmux required on macOS/Linux but not in PATH (RUNTIME_PREREQUISITE_MISSING)
```

Windows is unaffected because the ConPTY runtime is compiled into the daemon.

This is a follow-up to #2256 / #2259, which made the missing-tmux case *fail
fast with a clear error*. That was the right first step, but it stops at telling
the user what's wrong. This issue proposes actually shipping tmux so the mac/Linux
apps work out of the box.

## Current state (verified on `main`)

- `frontend/forge.config.ts` — `extraResource: ["daemon", "assets/icon.png", "app-update.yml"]`. No tmux binary is bundled into any package (DMG/zip on macOS; AppImage/deb/rpm on Linux).
- The `maker-deb` and `maker-rpm` configs declare no `depends`/`requires`, so even the Linux system packages don't pull tmux in.
- `backend/internal/adapters/runtime/tmux/tmux.go` resolves the binary purely via `exec.LookPath("tmux")`.
- `backend/internal/adapters/runtime/runtimeselect/runtimeselect.go` hard-selects tmux on Darwin/Linux with no fallback runtime.
- `backend/internal/session_manager/manager.go` (`validateRuntimePrerequisites`) hard-fails spawn when tmux is absent.

Net effect: the packaged mac/Linux app ships a product that cannot spawn out of the box.

## Proposed fix

Ship tmux as a first-class dependency of the mac/Linux installers. Two
complementary tracks:

1. **Bundle a prebuilt tmux binary** (primary — covers all mac/Linux installers,
   including DMG and AppImage which have no system-package dependency mechanism):
   - Add per-platform/arch static tmux binaries (macOS arm64/x64, Linux arm64/x64)
     as an `extraResource`.
   - Have `tmux.Options` / the runtime prefer the bundled binary path, falling
     back to `exec.LookPath("tmux")` when not bundled (e.g. dev builds).
   - On macOS the bundled binary must be signed + notarized with the app (it flows
     through `osxSign`/`osxNotarize`, same as the `daemon` resource today).
   - tmux links `libevent` + `ncurses`; use a statically linked build so there are
     no runtime dylib dependencies.

2. **Declare tmux as a system-package dependency** (cheap complement for Linux
   system installs):
   - `depends: ["tmux"]` on `maker-deb`, `requires: ["tmux"]` on `maker-rpm`.
   - Does not help AppImage or macOS, hence track 1 is the real fix.

### Explicitly out of scope / not preferred

Falling back to the in-process `ptyexec` runtime instead of requiring tmux would
remove the dependency, but it sacrifices tmux-backed session persistence and
reattach across daemon restarts (the crash-proof reconcile design). So the
recommendation is to **ship tmux**, not to drop it.

## Acceptance criteria

- Installing the packaged app on a clean macOS or Linux machine with **no
  system tmux** lets a user add a project and spawn a session successfully.
- No `RUNTIME_PREREQUISITE_MISSING` on a stock install.
- macOS build remains signed/notarized with the bundled binary sealed in.
- Dev builds (`npm run dev`, `go run ./cmd/ao`) still work using system tmux.

### LINKED ISSUE #3564: feat: one-time interactive tmux auto-install at startup (like pre-rewrite ensureTmux)

### LINKED ISSUE #3800: Daemon boot fails when an absent tmux socket is classified as infrastructure unavailable
## Summary

AO Desktop would not start after the local tmux server had gone away. The daemon reached boot reconciliation, could not confirm cleanup for several terminated sessions, and exited with `session: unsafe to serve`.

This looks like a small follow-up to #3475 rather than the same bug returning. #3475 correctly made server-level tmux failures inconclusive so one outage could not archive every session. The case here is narrower: tmux 3.7b reports a genuinely missing socket as `error connecting ... (No such file or directory)`, which is currently grouped with failures such as `Permission denied`.

Thanks for taking a look. I have included the environment, reproduction, and sanitized logs below; please let me know if another detail would help.

## Environment

- Observed build: local integration commit `619bed7cb0118bd0f6c36adddb200dd44836a89f` (`ao version` reports `dev`)
- Compared with upstream `main` at `3e2dbf424`
- macOS arm64, zsh
- tmux 3.7b
- Go 1.26.5

The exact startup refusal was reproduced on the integration build, which already contains stricter fail-closed boot reconciliation. Upstream `main` shares the underlying tmux error classification but currently handles it more permissively, so it may not show the identical boot failure yet.

## Steps to reproduce

1. Have one or more terminated sessions that still retain their tmux runtime handles for boot cleanup.
2. Stop the final tmux session so the tmux server socket no longer exists. A machine restart can produce the same state.
3. Start AO Desktop.
4. tmux returns exit 1 from `has-session` with:

   ```text
   error connecting to <TMUX_SOCKET> (No such file or directory)
   ```

5. Boot reconciliation treats the probe as inconclusive and refuses to serve because terminated runtime cleanup cannot be confirmed.

## Expected behavior

AO should recognize that the old tmux server is absent, finish safe boot reconciliation, and restore any sessions that have restore markers.

## Actual behavior

The daemon exits with code 1 and AO Desktop reports that the daemon failed to start.

## Redacted logs

Usernames, home-directory paths, project/session names, timestamps, PIDs, and token-adjacent context have been replaced or omitted.

```text
time=<TIMESTAMP> level=INFO msg="data-dir ownership acquired" lock=<AO_DATA_DIR>/daemon.lock pid=<DAEMON_PID>

time=<TIMESTAMP> level=ERROR msg="reconcile: live pass failed, skipping" sessionID=<ACTIVE_SESSION> error="reconcile <ACTIVE_SESSION>: probe: tmux runtime: probe session <ACTIVE_SESSION>: runtime: infrastructure unavailable: error connecting to <TMUX_SOCKET> (No such file or directory)"

time=<TIMESTAMP> level=ERROR msg="reconcile: reap pass left boot unsafe" sessionID=<TERMINATED_SESSION_1> error="session: unsafe to serve: terminated runtime reap unresolved: session <TERMINATED_SESSION_1> probe: tmux runtime: probe session <TERMINATED_SESSION_1>: runtime: infrastructure unavailable: error connecting to <TMUX_SOCKET> (No such file or directory)"

time=<TIMESTAMP> level=ERROR msg="reconcile: reap pass left boot unsafe" sessionID=<TERMINATED_SESSION_2> error="session: unsafe to serve: terminated runtime reap unresolved: session <TERMINATED_SESSION_2> probe: tmux runtime: probe session <TERMINATED_SESSION_2>: runtime: infrastructure unavailable: error connecting to <TMUX_SOCKET> (No such file or directory)"

reconcile sessions on boot: session: unsafe to serve: terminated runtime reap unresolved: session <TERMINATED_SESSION_1> probe: tmux runtime: probe session <TERMINATED_SESSION_1>: runtime: infrastructure unavailable: error connecting to <TMUX_SOCKET> (No such file or directory)
```

## What I found

The tmux adapter currently treats every `error connecting` result as `ErrRuntimeUnavailable`. That is the right conservative answer for permission failures and unknown connection errors. For the specific ENOENT form, however, the old Unix socket—and therefore the old tmux server—is absent. A tmux pane cannot outlive that server process.

The distinction matters because steady-state board probes should remain conservative, while boot cleanup needs a safe way to recognize authoritative server absence.

## Possible direction

One approach is a typed server-absence error beneath `ErrRuntimeUnavailable`:

- classify `no server running ...` and `error connecting ... (No such file or directory)` as authoritative server absence;
- keep `Permission denied` and other connection failures inconclusive;
- let boot recovery and cleanup explicitly consume the narrower subtype;
- keep steady-state observation fail-closed, preserving #3475's protection.

I opened #3801 with this approach, but I am happy to adjust the error boundary or test shape if maintainers prefer a different design.

## Workaround used locally

Starting a temporary empty tmux session before relaunch allowed reconciliation to see the durable handles as missing. AO then started and restored the active session, after which the temporary tmux session was removed. No database rows or worktrees were edited.

## Related

- #3475 — protects the board from treating a server-level outage as many independent session deaths.
- #3801 — proposed narrow distinction for authoritative socket absence.

AgentWrapper:
Fixes #2427
Fixes #3800
Fixes #2443
Fixes #3564
Fixes #2206

## Before

- `tmux has-session` reported both a genuinely absent socket and permission/reachability outages as the same infrastructure error. Boot reconciliation could not safely recover the former without also masking the latter.
- A clean macOS/Linux install discovered the tmux prerequisite only when the first terminal session failed.
- A partial or cross-platform Electron postinstall on Windows surfaced later as Forge's misleading “Electron failed to install correctly” error.
- The stable-release guard checked updater manifests but not the unversioned launcher aliases used by `ao start`, allowing the 404 reported in #2427 to recur.

## After

- Add the typed `ErrRuntimeServerAbsent` subset. Exact tmux “no server running” and missing-socket/ENOENT diagnostics are recoverable during restart and boot reconciliation; permission and other ambiguous connection failures remain fail-closed and leave durable session state untouched.
- Restore the pre-rewrite one-time interactive tmux setup at `ao start`: Homebrew on macOS and apt/dnf/pacman/zypper/apk on Linux, using sudo only when available. Noninteractive launches never mutate the machine and print the exact remediation command. Windows continues to use ConPTY.
- Validate Electron before every Forge development/package/make/publish path, repair an incomplete official Electron download once, reject copied binaries for the wrong platform, and fail early on unsupported Node versions.
- Require macOS arm64/x64 ZIPs, Linux AppImage, Windows installer, and updater manifests on the latest stable release.

This intentionally uses the interactive auto-install option agreed in #3564 rather than embedding an unsigned tmux binary in every desktop artifact.

## Tests

- `env -u AO_PROJECT_ID -u AO_SESSION_ID go test ./...`
- isolated-cache `golangci-lint v2.12.2` (`0 issues`)
- `GOOS=windows GOARCH=amd64 go test -c ./internal/cli`
- `npx vitest run scripts/ensure-electron.test.mjs` (5 tests)
- `npm run typecheck`
- `npm run package` (macOS arm64)

## Clean-environment verification

- Installed `packages/product-ui` and `frontend` from their lockfiles with `npm ci`, then ran `npm run package`. The resulting `.app` contains arm64 Mach-O Electron, daemon, and agent-browser executables.
- Simulated the Windows first install with `npm ci --ignore-scripts`, confirmed Electron had no `path.txt`/binary, then ran `npm_config_platform=win32 npm_config_arch=x64 npm run electron:ensure`. It downloaded and verified `electron.exe`; `file` identified it as `PE32+ executable (GUI) x86-64, for MS Windows`. Running the ensure step again for macOS detected the foreign marker and repaired it back to the local arm64 binary.
- Ran installation under Node 26 and confirmed `.npmrc` now fails immediately with `EBADENGINE`, pointing users to the supported Node 20/22/24 LTS range instead of failing later inside Electron.
- Verified the tmux-less/noninteractive path locally with injected package-manager discovery and `/dev/null`: it never prompts or invokes an installer and emits the exact install command. GitHub's clean Debian container smoke test also passed with tmux deliberately absent.
- Queried current stable release `v0.12.4` and confirmed all newly guarded launcher aliases and updater manifests are present.

## CI

All required jobs pass, including race tests, lint, API drift, frontend tests, renderer smoke, security scan, clean container smoke, and native Windows/macOS/Linux packaging.


backend/internal/cli/tmux.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return nil
+}
+
+func (c *commandContext) withTmuxInstallPrivilege(goos string, argv []string) []string {
+	if goos != "linux" || os.Geteuid() == 0 {
+		return argv
+	}
+	if path, err := c.deps.LookPath("sudo"); err != nil || path == "" {
+		return nil
```
AgentWrapper: When a supported package manager is present but sudo is not, this returns nil and `ensureTmux` falls back to the generic remedy at line 22 (`install it with your package manager`). The user in that situation already has a package manager; what they lack is root, and the generic message hides both the concrete command and the actual blocker. Consider still returning the bare command so the warning can say something like `run as root: apt-get install -y tmux`. Non-blocking.

backend/internal/cli/tmux.go
```diff
@@ -0,0 +1,101 @@
...
+// commands require root; an unprivileged process without sudo gets guidance but
+// no doomed install attempt.
+func (c *commandContext) tmuxInstallCommand(goos string) []string {
+	var candidates [][]string
+	switch goos {
+	case "darwin":
+		candidates = [][]string{{"brew", "install", "tmux"}}
+	case "linux":
+		candidates = [][]string{
+			{"apt-get", "install", "-y", "tmux"},
```
AgentWrapper, line 81: On an image with empty or stale package lists (fresh Debian/Ubuntu containers, machines that never ran update), `apt-get install -y tmux` fails with `Unable to locate package tmux` even after the user consents, and the flow degrades to the warning. Not blocking since the fallback is safe, but consider prompting with (and running) `apt-get update && apt-get install -y tmux` for the consented interactive path so the prompt still shows exactly what will run.

backend/internal/adapters/runtime/tmux/tmux.go
```diff
@@ -958,21 +960,29 @@ func sessionMissingOutput(out string) bool {
...
-func serverUnreachableOutput(out string) bool {
+// serverAbsentOutput reports tmux diagnostics that prove the server does not
+// exist. tmux versions use both spellings below for the same missing Unix
+// socket. ENOENT is authoritative even if a new server races to start later:
+// the old server's panes cannot survive its exit.
+func serverAbsentOutput(out string) bool {
 	s := strings.ToLower(out)
 	return strings.Contains(s, "no server running") ||
-		strings.Contains(s, "error connecting")
+		(strings.Contains(s, "error connecting") && strings.Contains(s, "no such file or directory"))
```
AgentWrapper, line 977: The `(No such file or directory)` suffix comes from strerror(3) and is localized under a non-C locale, while `no server running` is a literal tmux string. Under e.g. `LANG=fr_FR.UTF-8`, a tmux build that emits the strerror spelling for a missing socket would classify as merely unreachable, and boot reconciliation would fail closed instead of restoring. That degrades in the safe direction, so this is non-blocking — but consider pinning `LC_ALL=C` in the env of tmux probe invocations, or recording the locale assumption in this comment.

frontend/scripts/ensure-electron.test.mjs
```diff
@@ -0,0 +1,107 @@
...
+    ensureElectronInstalled({
+      electronRoot: root,
+      env: { npm_config_platform: "darwin" },
+      run,
+      log: { warn: vi.fn(), log: vi.fn() },
+    });
+    expect(run).toHaveBeenCalledOnce();
+  });
+
+  it("reports an actionable error when the repair download fails", () => {
```
AgentWrapper, line 87: This case leaves `env` defaulted to `process.env`, so if `ELECTRON_SKIP_BINARY_DOWNLOAD` is set in the ambient environment (a common CI setting for electron repos) the call throws the skip-download error and the `proxy/antivirus` assertion fails spuriously. Passing `env: {}` keeps the test hermetic.

AgentWrapper (commented):
Approving — no blocking findings.

This PR does what it says: the new `ErrRuntimeServerAbsent` subset keeps ambiguous outages fail-closed while letting restart and boot reconciliation recover from a proven-absent tmux server; the interactive `ao start` tmux setup never mutates a machine without an explicit yes on a real terminal (the `/dev/null` guard closes the noninteractive docker-run hole); and `ensure-electron.mjs` correctly mirrors the electron package layout (`path.txt`, `dist/version`, `ELECTRON_OVERRIDE_DIST_PATH`) with a single repair attempt. Because the new sentinel wraps `ErrRuntimeUnavailable`, every caller that did not opt in (reaper, shellterm, agent switching, terminal attachment) keeps its fail-closed behavior unchanged. Test coverage is thorough on both the adapter classification and the manager consumption paths, including the negative permission-failure cases, and `scripts/ensure-electron.test.mjs` is picked up by the default vitest include glob so it runs in CI.

Four non-blocking notes inline:
1. A sudo-less Linux box gets the generic remedy even though a usable package manager was found.
2. `apt-get install` without `apt-get update` can fail on fresh images even after the user consents.
3. The `No such file or directory` match is strerror text and locale-sensitive.
4. One vitest case inherits `process.env` and can trip on an ambient `ELECTRON_SKIP_BINARY_DOWNLOAD`.

Verdict: **approve** — ready to merge as is; the notes above are suggestions.

i-trytoohard:

i-trytoohard:
@AgentWrapper Marking as stale: no activity for 29 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@AgentWrapper Marking as stale: no activity for 29 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #3986, opened 2026-08-14
fix(chat): stop freezes, attachment loss, and turn-state desync

### LINKED ISSUE #3749: bug(chat): "Stop turn" fails with CHAT_NO_ACTIVE_TURN while UI shows a running turn — in-memory/durable turn-state desync

### LINKED ISSUE #3878: bug(chat): nested Codex turns can drain queued automation before the root completes, leaving a permanent phantom Working turn

### LINKED ISSUE #3884: bug(chat): restart recovery deletes image attachments from conversation history

### LINKED ISSUE #3945: Chat UI freezes when long-running command exceeds output cap (cannot switch to Terminal)

AgentWrapper:
Fixes #3945, fixes #3884, fixes #3878, fixes #3749 — the P1 reliability/data-loss cluster in the Chat UI surface.

These four bugs share one theme: durable conversation state, in-memory controller state, and the disposable worktree had mismatched lifetimes and authority. This PR makes durable state authoritative for what the user sees and can act on, and stops ephemeral state (worktrees, nested provider turns, capped output deltas) from destroying or wedging it.

## #3945 — Chat UI freezes when a command exceeds the output cap; Terminal unreachable

**Root causes fixed:**
- `AppendConversationActivityOutput` (and its streamed-text sibling) kept bumping `revision`/`updated_at` on every delta after the 64K cap, firing a CDC event → SSE push → full snapshot refetch per delta for bytes that could no longer change. A vite dev server past the cap turned that into a re-render storm that froze the page. The queries now skip the row entirely once its truncation flag is set (`AND command_output_truncated = 0`), so capped deltas produce **zero** CDC traffic. The settled aggregate on `item/completed` still lands.
- An interrupt-policy chat→TUI handoff waited forever for the provider to settle the cancelled turn. If the provider is wedged on the very command the user is escaping, the switch never completed ("policy dialog that cannot complete"). `BeginHandoff` now bounds that wait (`handoffSettleWait`, 10s default) and reconciles the durable turn as interrupted so the switch always completes — the terminal is the escape hatch and must always be reachable.

## #3884 — restart recovery deletes image attachments (data loss)

Attachment bytes now have a durable owner: `$AO_DATA_DIR/attachments/<session-id>/` (new `internal/attachments` package), with the worktree `.ao/attachments` copy demoted to a projection. Follows the fix direction laid out in the issue:

- **Stage atomically:** canonical copy first (tmp+rename, 0600); a failed durable write refuses the attachment instead of accepting bytes that vanish on the next restart. The existing worktree-relative refs are unchanged.
- **Serve durably:** the preview route serves strictly-validated `.ao/attachments/<name>` references from durable storage when the worktree copy is missing; a live worktree copy still wins; ordinary preview files stay confined to the workspace.
- **Restore projection:** `RestoreAll` rematerializes the session's durable attachments into the recreated worktree (and re-adds the git exclude) before the controller relaunches.
- **Upgrade safely:** `saveAndTeardownOne` imports legacy worktree-only attachments into durable storage *before* `ForceDestroy` — the exact seam that used to delete their only bytes. Covers attachments staged by pre-fix builds, including spawn attachments.
- **Cleanup tied to durable ownership:** canonical bytes are never removed on runtime teardown or worktree recycling.
- **Security:** strict name validation (no separators/traversal/hidden files), symlinks never followed into or out of durable storage, cross-session access impossible by construction.

## #3878 — nested Codex turns drain queued automation prematurely

Applies the fix from PR #3890 by @Dhruv-Sharma01 (rebased cleanly onto main; commit authorship preserved): `ChatEvent` carries `ProviderConversationID`, only root-thread lifecycle events can claim/release primary turn ownership, in-memory ownership updates move to post-commit (`applyCommittedTurnLifecycle`), nested completions no longer report idle or drain the queue, handoff quiescence is serialized behind `sendMu`, and adapter interrupts target the root turn.

## #3749 — "Stop turn" returns CHAT_NO_ACTIVE_TURN while the UI shows a running turn

Composes the approach from PR #3752 by @xRealBlue on top of the nested-turn lifecycle fix:

- **Durable-state fallback:** when memory says nothing is running but SQLite has a `running` turn (failed projection, restart, dispatch race), `Interrupt` reconciles it instead of refusing — serialized behind `sendMu` so a mid-dispatch state can't be misread.
- **Provider-refusal reconciliation:** `ErrChatNoActiveTurn` for a turn memory still tracks now settles the durable row rather than surfacing a bare error that leaves the user stuck on "Working".
- **`reconcileDurableTurn`** delivers the full Interrupt contract in one place: best-effort provider cancel, settle as interrupted, clear memory, cancel the queue behind it, report idle.
- **Client resilience:** the renderer and mobile clients refetch the conversation when an interrupt fails, so a stale "Working" bar is discovered instead of kept.

The projection-transaction half of the desync (memory mutated inside a rollback-able transaction) is fixed structurally by the post-commit lifecycle handling from the #3878 fix.

## Tests

New coverage (all requested seams):
- **Attachment survival across restart:** `TestAttachmentsSurviveTeardownAndRestore` (manager-level, incl. legacy import), `TestSessionsAPI_PreviewServesDurableAttachmentAfterWorktreeLoss` (HTTP incl. traversal + cross-session rejection), `internal/attachments` unit tests (validation, symlink skip, atomic store), and the lifecycle-level e2e `TestChatAttachmentSurvivesADaemonRestart` (real daemon, real codex, real restart — the exact repro from the issue).
- **Stop-turn authority under desync:** `TestInterruptReconcilesStaleRunningTurnOnDisk`, `TestProviderRefusalReconcilesTheDurableTurnAsInterrupted`, `TestInterruptReconciliationCancelsQueuedTurns`, `TestInterruptReturnsNoActiveTurnWhenNoRunningTurnAnywhere`, and `TestProjectionFailureThenStopStillStopsTheTurn` (the exact #3749 sequence: failed `turn/completed` projection, then Stop).
- **Output-cap non-wedge:** `TestCappedOutputStopsBumpingRevision` (revision frozen after truncation), `TestInterruptHandoffCompletesWhenProviderNeverSettlesTheTurn` (bounded escape to Terminal).
- **Nested-turn automation ordering:** `TestNestedTurnCompletionDoesNotDrainQueueWhilePrimaryTurnRuns`, `TestChatHandoffDrainWaitsForRootAfterNestedTurnCompletes`, plus adapter/normalizer tests (from #3890).

Validation run:
- `go test ./...` — green (the `internal/cli` spawn tests fail only when AO_* env vars from a live AO session leak into the test env; they pass with a clean env and fail identically on unmodified `main`).
- `cd frontend && npm run typecheck` — clean.
- End-to-end: `AO_CHAT_E2E=1 go test ./e2e/ -run TestChatAttachmentSurvivesADaemonRestart` — real daemon binary, real codex session, attachment staged → daemon stop → recovery → same preview URL serves the same bytes and the resumed agent reads the same path.

## Known follow-ups (out of scope)

- #3878's "acknowledgement recovery for a dispatched provider ID that never emits turn.started" is not implemented here (nor in #3890); Stop-based reconciliation covers user recovery in the meantime.
- The `groupByTurn` presentation defect (parent activity anchored above completed child cards) from #3878's "UI clarity" section is a separate rendering change.
- #3950/#3951 (interface-switch recovery banner persistence) remain separate.

## Relationship to open PRs

- #3890 (nested turns) is included as its own commit with original authorship — this PR supersedes it only by also carrying the three sibling fixes; happy to rebase if #3890 lands first.
- #3752 (stop-turn desync) is re-implemented here on top of the nested-turn lifecycle (the two PRs conflict in `controller.go`); its author is credited as commit co-author.

### DIFF
--- a/backend/e2e/chat_composer_test.go
+++ b/backend/e2e/chat_composer_test.go
@@ -3,6 +3,9 @@
 package e2e
 
 import (
+	"bytes"
+	"encoding/base64"
+	"io"
 	"net/http"
 	"strings"
 	"testing"
@@ -149,4 +152,80 @@ func TestChatAttachmentIsStagedIntoTheWorktreeAndReadableByTheAgent(t *testing.T
 			t.Errorf("staged attachment %q shows up as a workspace change (%s)", file.Path, file.Status)
 		}
 	}
+}
+
+// The #3884 data-loss seam, driven end to end: an attachment staged before a
+// daemon restart must still resolve — same preview URL, same bytes — after
+// startup recovery has recreated the session's worktree, and the resumed agent
+// must still be able to read the same .ao/attachments/... path its
+// conversation history names.
+func TestChatAttachmentSurvivesADaemonRestart(t *testing.T) {
+	requireE2E(t)
+	dataDir := t.TempDir()
+	d := startDaemon(t, dataDir)
+	project := seedProject(t, d, "attachrestart")
+	session := chatSession(t, d, project, "Reply with exactly: READY")
+
+	const onePixelPNG = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8DwHwAFAAH/q842iQAAAABJRU5ErkJggg=="
+	wantBytes, err := base64.StdEncoding.DecodeString(onePixelPNG)
+	if err != nil {
+		t.Fatalf("decode fixture: %v", err)
+	}
+
+	var staged struct {
+		Paths []string `json:"paths"`
+	}
+	d.mustCall("POST", "/sessions/"+session+"/attachments", http.StatusCreated,
+		map[string]any{"attachments": []map[string]any{{"mimeType": "image/png", "data": onePixelPNG}}},
+		&staged)
+	if len(staged.Paths) != 1 {
+		t.Fatalf("staged %d paths, want 1: %+v", len(staged.Paths), staged.Paths)
+	}
+	path := staged.Paths[0]
+	previewPath := "/sessions/" + session + "/preview/files/" + path
+
+	if got := rawGET(t, d, previewPath); !bytes.Equal(got, wantBytes) {
+		t.Fatalf("preview before restart returned %d bytes, want the staged image", len(got))
+	}
+
+	// The user quits the app; the next boot's recovery recreates the worktree.
+	d.stop()
+	restarted := startDaemon(t, dataDir)
+	restarted.awaitLiveController(session, 3*time.Minute)
+
+	// The same URL the durable conversation embeds must keep answering with the
+	// original bytes — this is the request that returned PREVIEW_FILE_NOT_FOUND
+	// before attachments had a durable owner.
+	if got := rawGET(t, restarted, previewPath); !bytes.Equal(got, wantBytes) {
+		t.Fatalf("preview after restart returned %d bytes, want the staged image", len(got))
+	}
+
+	// And the resumed agent can still open the path its history names.
+	send(t, restarted, session,
+		"Run `test -f "+path+" && echo FOUND || echo MISSING` and reply with its exact output.",
+		"attach-restart-read")
+	snap := restarted.awaitConversation(session, 3*time.Minute, "the resumed agent to look for the image",
+		func(s snapshot) bool { return terminal(s.Turns[len(s.Turns)-1].State) })
+	if !contains(snap.assistantText(), "FOUND") {
+		t.Fatalf("the resumed agent could not see the restored image at %s:\n%s", path, describe(snap))
+	}
+}
+
+// rawGET fetches an API path and returns the body bytes without assuming JSON,
+// which is what an <img> tag does with a preview URL.
+func rawGET(t *testing.T, d *daemon, path string) []byte {
+	t.Helper()
+	res, err := httpClient.Get(d.baseURL + path)
+	if err != nil {
+		t.Fatalf("GET %s: %v", path, err)
+	}
+	defer func() { _ = res.Body.Close() }()
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("read %s: %v", path, err)
+	}
+	if res.StatusCode != http.StatusOK {
+		t.Fatalf("GET %s = %d: %s\n%s", path, res.StatusCode, body, d.tailLog())
+	}
+	return body
 }
--- a/backend/internal/adapters/chatdriver/codexappserver/conversation.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/conversation.go
@@ -148,12 +148,20 @@ func (c *conversation) pump() {
 		// as an absolute instant and has to become a remaining duration, and a
 		// normalizer that reads the clock itself cannot be tested deterministically.
 		for _, ev := range normalizeNotification(n, time.Now()) {
-			if ev.Kind == ports.ChatEventTurnStarted && ev.ProviderTurnID != "" {
+			rootConversation := ev.ProviderConversationID == "" || ev.ProviderConversationID == c.threadID
+			if ev.Kind == ports.ChatEventTurnStarted && ev.ProviderTurnID != "" && rootConversation {
 				c.mu.Lock()
 				c.activeTurn = ev.ProviderTurnID
 				// Snapshot the context position this turn starts from. Cheap on every
 				// turn, and the only way to know what a compaction reclaimed.
 				c.contextAtTurnStart = c.contextTokens
+				c.mu.Unlock()
+			}
+			if ev.Kind == ports.ChatEventTurnCompleted && rootConversation {
+				c.mu.Lock()
+				if c.activeTurn == ev.ProviderTurnID {
+					c.activeTurn = ""
+				}
 				c.mu.Unlock()
 			}
 			if ev.Kind == ports.ChatEventCompacted {
--- a/backend/internal/adapters/chatdriver/codexappserver/driver_test.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/driver_test.go
@@ -508,6 +508,42 @@ func TestNotificationsBecomeNeutralEvents(t *testing.T) {
 	}
 }
 
+// app-server multiplexes child-agent thread notifications over the root
+// connection. The adapter's fallback target must remain the root turn; otherwise
+// an interrupt without an explicit id can stop a child and leave the requested
+// root work running.
+func TestNestedThreadDoesNotReplaceRootActiveTurn(t *testing.T) {
+	d, srv := newTestDriver(t)
+	conv, err := d.Start(context.Background(), ports.ChatStartConfig{WorkspacePath: "/tmp/ws"})
+	if err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	defer func() { _ = conv.Close() }()
+
+	srv.push(`{"method":"turn/started","params":{"threadId":"thread-1","turn":{"id":"root-turn","status":"inProgress","items":[]}}}`)
+	srv.push(`{"method":"turn/started","params":{"threadId":"child-thread-1","turn":{"id":"child-turn","status":"inProgress","items":[]}}}`)
+	root := nextEvent(t, conv.Events(), ports.ChatEventTurnStarted)
+	child := nextEvent(t, conv.Events(), ports.ChatEventTurnStarted)
+	if root.ProviderConversationID != "thread-1" || child.ProviderConversationID != "child-thread-1" {
+		t.Fatalf("normalized lifecycle threads = root %q child %q", root.ProviderConversationID, child.ProviderConversationID)
+	}
+
+	if err := conv.Interrupt(context.Background(), ""); err != nil {
+		t.Fatalf("Interrupt: %v", err)
+	}
+	sent := srv.awaitFrame(func(f frame) bool { return f.Method == "turn/interrupt" })
+	var params struct {
+		ThreadID string `json:"threadId"`
+		TurnID   string `json:"turnId"`
+	}
+	if err := json.Unmarshal(sent.Params, &params); err != nil {
+		t.Fatalf("decode turn/interrupt params: %v", err)
+	}
+	if params.ThreadID != "thread-1" || params.TurnID != "root-turn" {
+		t.Fatalf("interrupt target = %q/%q, want thread-1/root-turn", params.ThreadID, params.TurnID)
+	}
+}
+
 // Resume must not quietly become a fresh thread: that would present unrelated
 // history as continuous.
 func TestResumeFailureDoesNotFallBackToStart(t *testing.T) {
--- a/backend/internal/adapters/chatdriver/codexappserver/normalize.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/normalize.go
@@ -136,8 +136,9 @@ func normalizeNotification(n notification, now time.Time) []ports.ChatEvent {
 			return nil
 		}
 		return []ports.ChatEvent{{
-			Kind:           ports.ChatEventTurnStarted,
-			ProviderTurnID: firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
+			Kind:                   ports.ChatEventTurnStarted,
+			ProviderTurnID:         firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
+			ProviderConversationID: p.ThreadID,
 		}}
 
 	case codexproto.MethodTurnCompleted:
@@ -146,9 +147,10 @@ func normalizeNotification(n notification, now time.Time) []ports.ChatEvent {
 			return nil
 		}
 		ev := ports.ChatEvent{
-			Kind:           ports.ChatEventTurnCompleted,
-			ProviderTurnID: firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
-			TurnState:      turnStateFrom(string(p.Turn.Status)),
+			Kind:                   ports.ChatEventTurnCompleted,
+			ProviderTurnID:         firstNonEmpty(p.Turn.ID, turnIDFallback(n.Params)),
+			ProviderConversationID: p.ThreadID,
+			TurnState:              turnStateFrom(string(p.Turn.Status)),
 		}
 		if p.Turn.Error != nil && p.Turn.Error.Message != "" {
 			ev.Err = fmt.Errorf("%s", p.Turn.Error.Message)
--- a/backend/internal/adapters/chatdriver/codexappserver/normalize_test.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/normalize_test.go
@@ -37,12 +37,14 @@ func normalizeNone(t *testing.T, method, params string) {
 
 func TestNormalizeTurnLifecycle(t *testing.T) {
 	started := normalizeOne(t, "turn/started", `{"threadId":"th1","turn":{"id":"tu1","status":"inProgress","items":[]}}`)
-	if started.Kind != ports.ChatEventTurnStarted || started.ProviderTurnID != "tu1" {
+	if started.Kind != ports.ChatEventTurnStarted || started.ProviderTurnID != "tu1" ||
+		started.ProviderConversationID != "th1" {
 		t.Fatalf("turn/started -> %+v", started)
 	}
 
 	done := normalizeOne(t, "turn/completed", `{"threadId":"th1","turn":{"id":"tu1","status":"completed","items":[]}}`)
-	if done.Kind != ports.ChatEventTurnCompleted || done.TurnState != domain.TurnStateCompleted {
+	if done.Kind != ports.ChatEventTurnCompleted || done.TurnState != domain.TurnStateCompleted ||
+		done.ProviderConversationID != "th1" {
 		t.Fatalf("turn/completed -> %+v", done)
 	}
 }
--- a/backend/internal/attachments/attachments.go
+++ b/backend/internal/attachments/attachments.go
@@ -1,0 +1,211 @@
+// Package attachments owns the canonical bytes for files a user attaches to a
+// session. The worktree copy under .ao/attachments is a projection for the
+// agent; it lives in a disposable git worktree that daemon recovery recreates,
+// which is exactly when a worktree-only attachment disappears while the durable
+// conversation still points at it (#3884). The canonical copy lives under AO's
+// data dir and shares the conversation's lifetime instead of the worktree's.
+package attachments
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// WorktreeDir is the worktree-relative directory attachments are projected
+// into. It matches the references embedded in durable conversation history, so
+// it can never change without a migration for every stored message.
+const WorktreeDir = ".ao/attachments"
+
+// Dir returns the canonical directory for one session's attachments.
+func Dir(dataDir, sessionID string) string {
+	return filepath.Join(dataDir, "attachments", sessionID)
+}
+
+// ValidName reports whether name is a bare filename AO could have generated for
+// an attachment. It is the gate between a request path and the filesystem: no
+// separators, no traversal, no hidden files, ASCII-safe charset only.
+func ValidName(name string) bool {
+	if name == "" || len(name) > 255 {
+		return false
+	}
+	if name != filepath.Base(name) || strings.HasPrefix(name, ".") {
+		return false
+	}
+	for _, r := range name {
+		switch {
+		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9':
+		case r == '-' || r == '_' || r == '.':
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+// RefName extracts and validates the filename from a worktree-relative
+// attachment reference like ".ao/attachments/attachment-ab12cd.png". It reports
+// false for anything that is not a well-formed AO attachment reference.
+func RefName(ref string) (string, bool) {
+	rest, ok := strings.CutPrefix(filepath.ToSlash(ref), WorktreeDir+"/")
+	if !ok {
+		return "", false
+	}
+	if !ValidName(rest) {
+		return "", false
+	}
+	return rest, true
+}
+
+// Store writes one canonical attachment atomically: the bytes land in a temp
+// file first and are renamed into place, so a crash mid-write can never leave a
+// half-written file that later serves as the "durable" copy.
+func Store(dataDir, sessionID, name string, data []byte) error {
+	if dataDir == "" {
+		return errors.New("attachments: data dir is required")
+	}
+	if !ValidName(name) {
+		return fmt.Errorf("attachments: invalid name %q", name)
+	}
+	dir := Dir(dataDir, sessionID)
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return fmt.Errorf("attachments: create %s: %w", dir, err)
+	}
+	tmp, err := os.CreateTemp(dir, ".tmp-"+name+"-*")
+	if err != nil {
+		return fmt.Errorf("attachments: temp file for %s: %w", name, err)
+	}
+	tmpName := tmp.Name()
+	defer func() { _ = os.Remove(tmpName) }()
+	if err := tmp.Chmod(0o600); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("attachments: chmod %s: %w", name, err)
+	}
+	if _, err := tmp.Write(data); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("attachments: write %s: %w", name, err)
+	}
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("attachments: close %s: %w", name, err)
+	}
+	if err := os.Rename(tmpName, filepath.Join(dir, name)); err != nil {
+		return fmt.Errorf("attachments: commit %s: %w", name, err)
+	}
+	return nil
+}
+
+// Open opens the canonical copy of one attachment for serving. The name is
+// validated before it touches the filesystem, and only regular files are
+// served — a symlink planted in the canonical dir must not become an escape.
+func Open(dataDir, sessionID, name string) (*os.File, os.FileInfo, error) {
+	if dataDir == "" || !ValidName(name) {
+		return nil, nil, fs.ErrNotExist
+	}
+	path := filepath.Join(Dir(dataDir, sessionID), name)
+	info, err := os.Lstat(path)
+	if err != nil {
+		return nil, nil, err
+	}
+	if !info.Mode().IsRegular() {
+		return nil, nil, fs.ErrNotExist
+	}
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, nil, err
+	}
+	return file, info, nil
+}
+
+// Materialize projects every canonical attachment for the session into the
+// worktree so the agent can read the same .ao/attachments/... paths its
+// conversation history names. Existing worktree files are left alone: the
+// worktree copy of a name that exists in both places is at least as new.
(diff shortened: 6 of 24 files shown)


illegalcall:
Closing this umbrella PR as superseded by focused changes:

- Nested Codex turn ownership: [PR #3890](https://github.com/Untrivial-ai/agent-orchestrator/pull/3890)
- Stop/durable turn-state reconciliation: [PR #3992](https://github.com/Untrivial-ai/agent-orchestrator/pull/3992) — all CI checks green
- Attachment durability: [PR #3891](https://github.com/Untrivial-ai/agent-orchestrator/pull/3891)
- Noisy-output/Terminal escape: [PR #3957](https://github.com/Untrivial-ai/agent-orchestrator/pull/3957) — already merged

This preserves the tested implementations and contributor credit while restoring the repository's one-issue-per-PR boundary.

PR #2980, opened 2026-07-22
fix: unify session state and lifecycle presentation

whoisasx:
## Summary

This PR makes AO's session state model consistent from backend facts through every frontend surface.

- Active agent work now wins over non-terminal SCM state, so a session immediately returns to Working when its agent resumes.
- Agent-process exit is no longer conflated with AO/tmux termination.
- Codex gets reliable process-exit reporting plus runtime/reaper reconciliation while preserving the interactive shell after Codex exits.
- Board placement, card badges, topbar activity, and sidebar dots now consume the correct canonical source.
- Idle/Working, Ready/Merged, Archive, semantic status colors, merge termination preferences/actions, and PR activity timing are consolidated into one implementation.

Closes #2950

Covers the narrower reproduction in #2944.

Supersedes the overlapping implementations in #2924, #2840, and #2841.

## Root cause

AO already persisted the necessary facts independently:

```text
activity_state   agent runtime activity
is_terminated    AO/runtime lifecycle fact
PR facts         SCM lifecycle, CI, review, comments, mergeability
status           derived display state
```

The inconsistencies came from two places:

1. Backend status derivation evaluated PR state before active agent work and converted an `exited` activity signal into termination.
2. Frontend components independently interpreted status/activity, so the board, sidebar dot, topbar pill, and cards could disagree even when they received the same session data.

Codex added a third gap: its hooks expose turn boundaries but no reliable session-end event, so the agent process could exit while tmux stayed alive and AO continued showing the last activity state.

## Final state precedence

| Priority | Durable/runtime facts | Derived display state | Board placement |
| ---: | --- | --- | --- |
| 1 | Terminated + any PR merged | `merged` | Archive |
| 2 | Terminated + no merged PR | `terminated` | Archive |
| 3 | Live + agent active | `working` | Working |
| 4 | Live + agent exited | `exited` | Needs you |
| 5 | Live + waiting/blocked | `needs_input` | Needs you |
| 6 | Live + idle + actionable PR problem | Existing PR status | Needs you |
| 7 | Live + idle + open/draft/reviewing PR | Existing PR status | In review |
| 8 | Live + idle + approved/mergeable PR | Existing PR status | Ready to merge |
| 9 | Live + idle + merged PR | `merged` | Merged |
| 10 | Live + idle + no PR | `idle` | Idle |

`status` remains derived at service read time and is never persisted. Existing stacked/multi-PR worst-wins aggregation and `no_signal` behavior remain intact.

### Lifecycle and state derivation

[...]

PR #3987, opened 2026-08-14
fix(lifecycle): cap concurrency, GC worktrees, and recover dead orchestrators

### LINKED ISSUE #2811: Merge-triggered termination leaves orphaned worktree: MarkTerminated skips teardown, Kill button vanishes, no auto-cleanup

### LINKED ISSUE #3402: cleanup is non-idempotent: session_cleanup_facts is dead code + opaque failures + no session-delete lifecycle
## Bug

`ao session cleanup` is not reliable, not idempotent, and does not persist teardown results. The `session_cleanup_facts` persistence layer was scaffolded (table, store, domain model, CDC triggers, retry bookkeeping) but is **dead code** — zero callers outside the store's own tests. Combined with opaque error reporting and archived-project gaps, cleanup cannot be trusted to reclaim disk or produce accurate results across runs.

Supersedes #2608 (incorporates its remaining open findings + new idempotency/dead-code findings).

**Source:** Security/architecture review by @nikhilachale  
**Analyzed against:** `6336e6527` (origin/main, 2026-07-31)  
**Confidence:** High — all claims traced to specific code paths below.

---

## Background: what cleanup does today

`ao session cleanup` → `POST /api/v1/sessions/cleanup` → `service.Cleanup` → `manager.Cleanup`. The manager:

1. Lists ALL sessions for the project (or all projects if unscoped)
2. For each `is_terminated=true` session, attempts workspace teardown via `cleanupOne`
3. Returns `{Cleaned: [...], Skipped: [...]}`

Critically: **on success, nothing is persisted.** No disposition is written, no workspace path is cleared on the session row, no `cleanup_generation` is bumped. The next `--dry-run` re-lists the same sessions.

---

## Finding 1 — `session_cleanup_facts` is fully scaffolded but completely unwired (P1)

The entire persistence layer for durable cleanup facts exists but has **zero production callers**:

| Layer | File | Status |
|-------|------|--------|
| Migration (table + CDC triggers) | `migrations/0030_session_cleanup_facts.sql` | ✅ Migrated |
| Domain model | `domain/cleanup.go` (`SessionCleanupRecord`, `WorkspaceDisposition`) | ✅ Defined |
| Store methods | `store/session_cleanup_store.go` (`Upsert`, `Get`, `ListCandidates`) | ✅ Implemented |
| Store tests | `store/session_cleanup_store_test.go` | ✅ Passing |
| **Service wiring** | `service/session/service.go` | ❌ **Never calls store** |
| **Manager wiring** | `session_manager/manager.go:2098-2124` | ❌ **Never calls `UpsertSessionCleanupFacts`** |
| **Reconciler loop** | (referenced in migration + store docstrings) | ❌ **Never built** |
| **Daemon finalizer** | `daemon/daemon.go` | ❌ **Never wired** |

The migration comment itself admits it:

> *"Additive only here: nothing bumps it yet, so every existing and new row keeps the default 0."*
> — `0030_session_cleanup_facts.sql:8`

The store docstring references a "terminal-resource reconciler" and "Go finalizer" that don't exist:

> *"the pre-existing leaked backlog — terminated sessions with NO facts row — is still surfaced"*
> — `session_cleanup_store.go:55-56`

This is the coarse `ListTerminalCleanupCandidates` scan designed to catch the backlog. It's tested but never called by anything except its own unit tests.

**Code evidence:**

```
session_manager/manager.go:2098-2124  (Cleanup — no UpsertSessionCleanupFacts call)
session_manager/manager.go:1426-1448  (Reconcile — no cleanup-facts reconciliation)
session_manager/manager.go:1395-1411  (reconcileReap — runtime destroy only, no workspace/facts)
```

```
grep -r "UpsertSessionCleanupFacts" backend/internal/ --include="*.go" | grep -v _test.go | grep -v gen/ | grep -v store/
# → (empty)
```

**Impact:** Every `ao session cleanup` re-scans ALL terminated sessions from the beginning of time. The retry bookkeeping (`attempt_count`, `next_attempt_at`, `failure_code`), disposition tracking (`pending`/`removed`/`preserved_dirty`/`failed`), and generation-guard that would prevent redundant work are all inert.

---

## Finding 2 — Cleanup is non-idempotent: success is not persisted (P1)

`manager.Cleanup` (L2098-2124) on a successful workspace teardown:

- Calls `cleanupOne` → `destroyWorkspaceProjectRows` or `workspace.Destroy` → removes the worktree directory
- Calls `cleanupAgentWorkspace` → removes agent workspace cache
- Appends to `result.Cleaned`
- **Does NOT** call `UpsertSessionCleanupFacts` to record `DispositionRemoved`
- **Does NOT** clear the workspace path on the session record
- **Does NOT** bump `cleanup_generation`

`previewCleanupSessions` (CLI L569) lists ALL sessions where `active=false` (i.e. `is_terminated=true`), with no filter for prior cleanup success:

```go
// cli/session.go:569-572
func (c *commandContext) previewCleanupSessions(ctx context.Context, project string) ([]sessionDTO, error) {
    params := url.Values{}
    params.Set("active", "false")
```

**Consequence:** `--dry-run` reports sessions that were successfully cleaned in a previous run as candidates again. The user sees a growing list of phantom candidates with no way to distinguish "already cleaned" from "still pending."

**Impact:** Misleading dry-run output. Users cannot trust the candidate list. Over time, every terminated session ever created reappears on every dry-run.

---

## Finding 3 — No session-delete API, CLI, UI, or retention policy (P2 — feature gap)

There is no `DELETE /api/v1/sessions/{id}` route. The route table (`controllers/sessions.go:100-115`):

```
GET    /sessions
POST   /sessions
POST   /sessions/cleanup
GET    /sessions/{id}
GET    /sessions/{id}/preview
POST   /sessions/{id}/preview
DELETE /sessions/{id}/preview        ← clears preview URL only
PATCH  /sessions/{id}
```

`RollbackSession` (L622) exists but is seed-state-only (deletes a session that was never spawned). There is no general-purpose session record deletion, archival, or TTL.

**Impact:** Terminated session records persist indefinitely in SQLite. The CLI hides them by default; the desktop UI collapses them under "Done / Terminated." But they accumulate forever — every session ever spawned remains in the database.

---

## Finding 4 — Cleanup reports opaque failure reasons (P1 — from #2608, still open)

`cleanupSkipReason` (`manager.go:2173-2181`) collapses all non-dirty failures to a single generic string:

```go
func cleanupSkipReason(err error) string {
    if errors.Is(err, ports.ErrWorkspaceDirty) {
        return "workspace has uncommitted changes"
    }
    if errors.Is(err, ErrProjectNotResolvable) {
        return "project is archived or unregistered — remove worktree manually"
    }
    return "workspace teardown failed"    // ← everything else
}
```

The raw error (with filesystem paths, lock reasons, git stderr) lands only in `daemon.log` via `m.logger.Warn`. Users see `"workspace teardown failed"` and have no idea what to do.

**Impact:** Users cannot distinguish locked worktree, missing repo, permission error, or stale registration — they must dig through daemon logs.

---

## Finding 5 — Locked/stale orchestrator worktrees not recoverable by normal cleanup (P2 — from #2608, still open)

`Destroy` in `adapters/workspace/gitworktree/workspace.go` intentionally avoids `--force`, which is correct for dirty-work safety. But stale locked worktrees (`fatal: cannot remove a locked working tree, lock reason: initializing`) are not dirty-work preservation — they're a different failure class that needs its own recovery path or diagnostic.

**Impact:** Repeated orchestrator sessions sharing the same path produce many skipped entries for one stale locked worktree. These accumulate across runs.

---

## Finding 6 — Reaper, PR completion, and agent-exit do not auto-clean worktrees (P2 — by design, but should be documented)

| Event | Auto-cleans worktree? |
|-------|-----------------------|
| `ao session kill` | No — terminates session, best-effort runtime destroy |
| Reaper (`reconcileReap`, L1395-1411) | No — destroys leaked runtime handles only |
| PR completion / merge | No |
| Agent exit (`Stop` hook) | No — transitions activity state only |

Workspace reclamation is **exclusive** to explicit `ao session cleanup`. This is a deliberate design choice (safety), but it means worktrees accumulate until the user manually runs cleanup.

**Impact:** Disk usage grows silently. Users don't know to run cleanup, and when they do, the result is misleading (Finding 2).

---

## Already fixed (from #2608)

| Item | Fixed by |
|------|----------|
| `--dry-run` was missing from CLI | Now present at `cli/session.go:259` |
| Archived-project worktrees unreclaimable (resolver rejects them) | [PR #2679](https://github.com/AgentWrapper/agent-orchestrator/pull/2679) persists `WorkspaceRepoPath` in session metadata, merged 2026-07-22 |

---

## Recommended fix direction

**Phase 1 — Wire the facts layer (addresses Findings 1 + 2):**

1. After `cleanupOne` succeeds, call `UpsertSessionCleanupFacts` with `DispositionRemoved` + `RuntimeReleasedAt`
2. After `cleanupOne` skips with dirty, write `DispositionPreservedDirty`
3. After `cleanupOne` fails (non-dirty), write `DispositionPending` with retry bookkeeping
4. Filter `previewCleanupSessions` (CLI) and `cleanupRecords` (manager) to exclude sessions with `DispositionRemoved` at the current generation
5. Build the reconciler loop the migration references — a periodic sweep using `ListTerminalCleanupCandidates` that retries `pending` rows past `next_attempt_at`

**Phase 2 — Structured diagnostics (addresses Finding 4):**

6. Expand `cleanupSkipReason` to return structured codes: `dirty`, `locked_worktree`, `archived_project_unresolvable`, `missing_repo`, `permission_error`
7. Surface these in the CLI output and API response

**Phase 3 — Session lifecycle management (addresses Findings 3 + 5):**

8. Add `DELETE /api/v1/sessions/{id}` for session record removal (gated on workspace already cleaned)
9. Add a recovery path for stale locked worktrees (distinct from force-delete of dirty work)
10. Document that worktree reclamation is manual-only and consider an auto-cleanup option

---

## Safety note

Do not weaken dirty-worktree preservation. `Destroy` avoiding `--force` is correct. The fix should persist `preserved_dirty` disposition and surface it clearly — not bypass the safety check.

---

## Related

- [#2608](https://github.com/AgentWrapper/agent-orchestrator/issues/2608) — original cleanup findings (superseded by this issue)
- [PR #2679](https://github.com/AgentWrapper/agent-orchestrator/pull/2679) — WorkspaceRepoPath persistence (merged, addresses #2608 P1 archived-project)

### LINKED ISSUE #3921: Orchestrator crashes under load (40+ workers) with no concurrency cap, no worktree GC, and unrecoverable session — requires project deletion to restart
## Bug

When an orchestrator spawns many workers (40-45) across multiple worktrees over an overnight run, the orchestrator session dies — likely from resource exhaustion (OOM on 8 GB Mac) or a lifecycle race — and **cannot be recovered**. The user must delete the project entirely and recreate it to restart. Orphaned worktrees and branches accumulate on disk.

The reporter's diagnosis ("worktree bloat") is partially correct but is a secondary symptom. The primary problems are: **(A) no concurrency cap in the Go rewrite** and **(B) the orchestrator session becomes unrecoverable after death** — neither is a worktree-specific issue.

**Source:** Discord `#bug-triage` thread (2026-08-13) | **Reported by:** community user (via @prateek)
**Analyzed against:** `565737839` (origin/main, 2026-08-13)
**Confidence:** High — root causes traced to specific code paths and confirmed against merged/closed issue history.
**Priority:** P1 — core workflow broken on resource-constrained machines, no safe workaround short of avoiding large orchestrator runs.

## Reporter's Account

> I had AO running on Mac (26.5.2, 8 GB) and Fedora 42. On Mac the agent keeps crashing, I think maybe due to memory — primarily when spawning orchestrator. Even in Linux I mostly see the issue in the orchestrator only. The orchestrator crashes and then I'm not able to find the old session. My guess is worktree bloat?
>
> Codex as orchestrator, agy and claude as workers. It spun 40/45 tasks in multiple worktrees over time. Left overnight. Came back after a few hours. The orchestrator had died and the worktrees and branches were orphans. When I tried to click the orchestrator it would not start. I had to delete the project in AO and recreate. This happened quite a few times, so I've started using AO only for small-range tasks — works pretty well. Any tips on what I can do better?

## Root Cause

This is a **composite bug** with three independent contributing factors, each already partially tracked in existing issues:

### Factor 1 — No `maxConcurrentSessions` cap in the Go rewrite (primary OOM driver)

PR [#918](https://github.com/Untrivial-ai/agent-orchestrator/pull/918) ("enforce maxConcurrentSessions for worker spawns") was **closed without merging** (`state: CLOSED`, `mergedAt: null`, closed 2026-07-06). The feature was designed in the old TypeScript codebase (issue [#916](https://github.com/Untrivial-ai/agent-orchestrator/issues/916)) and never ported to the Go rewrite.

Verified against current `origin/main`:
- `backend/internal/config/config.go` — the `Config` struct (lines 86-122) has **no** concurrency-related field. No `MaxConcurrentSessions`, no spawn limit.
- `grep -ri "maxConcurrent\|concurrentSession\|spawnLimit\|sessionLimit" backend/` — **zero matches** in any Go source file.
- The spawn path (`session_manager/manager.go` `Spawn`/`startSession`) has no guard counting active sessions before creating a new worktree + runtime.

Result: an orchestrator following `ao batch-spawn` guidance can spawn unlimited concurrent workers. Each agent harness (Claude Code, Codex, etc.) uses ~1.5-2 GB RAM. On an 8 GB Mac, 4-5 concurrent workers + the orchestrator + the daemon + OS = OOM. The OS OOM-killer or memory pressure terminates processes unpredictably — often the orchestrator or daemon itself.

**This is the direct cause of the Mac crash.** Issue [#916](https://github.com/Untrivial-ai/agent-orchestrator/issues/916) was closed (presumably as "addressed") but the fix was never merged into the Go rewrite. The gap remains wide open.

### Factor 2 — Orchestrator session unrecoverable after death (the "must delete project" symptom)

The boot-time `Reconcile` pass (`backend/internal/session_manager/manager.go:1990`) handles three scenarios:
1. **Live pass** (line 2006): for non-terminated sessions, adopt if runtime survived, else `saveAndTeardownOne` + mark terminated.
2. **Reap pass** (line 2018): kill leaked tmux sessions for terminated rows.
3. **Restore pass** (line 2026): relaunch shutdown-saved sessions via `RestoreAll`.

This recovery machinery exists and is well-designed for **graceful daemon restarts**. However, it breaks down in the reporter's scenario for several reasons:

- **OOM kill is not a clean exit.** If the orchestrator's agent process (e.g., Claude Code) is OOM-killed, the tmux session may survive but the agent inside is dead. The `reconcileLive` `IsAlive` probe (`manager.go:1922`) checks the **runtime** (tmux session), not the agent process inside it. A live tmux pane with a dead agent = "adopted" but actually broken — the session appears alive but does nothing.

- **`saveAndTeardownOne` can fail under disk pressure.** When worktree/disk bloat exists, `workspace.Destroy` or the preserve-ref capture can fail. On failure, `reconcileLive` falls back to `MarkTerminated` **without a restore marker** (line 1942-1946): the session is permanently terminated with no path to restore. `RestoreAll` skips it (no `session_worktrees` row = "killed by user before shutdown", line 2063-2065). This is the "can't restart the orchestrator" symptom — the session is marked terminated with no restore marker, and there is no UI path to respawn it short of deleting the project.

- **Related to [#3649](https://github.com/Untrivial-ai/agent-orchestrator/issues/3649):** `SESSION_INCOMPLETE_HANDLE` errors on restore confirm that sessions can enter permanently-unrecoverable states where neither restore nor send works.

### Factor 3 — No automatic worktree/branch GC (the "orphaned worktrees and branches" symptom)

Confirmed by tracing:
- `ao session cleanup` is **manual-only** — no periodic background job reclaims terminated sessions' worktrees.
- The `session_cleanup_facts` persistence layer is **dead code** — scaffolded (migration 0030, domain model, store methods, tests) but **zero production callers** (issue [#3402](https://github.com/Untrivial-ai/agent-orchestrator/issues/3402), High confidence, still open).
- Merge-terminated sessions leave worktrees on disk because `MarkTerminated` skips teardown (issue [#2811](https://github.com/Untrivial-ai/agent-orchestrator/issues/2811), High confidence, still open).
- After daemon crash, orphaned workers continue running independently (issue [#3537](https://github.com/Untrivial-ai/agent-orchestrator/issues/3537), still open).

An overnight run spawning 40-45 workers accumulates dozens of git worktrees + branches. Each worktree is a full checkout. On disk, this can be tens of GB. The reporter correctly identified this as contributing to instability.

## Immediate Workarounds (for the reporter)

Since the fix hasn't shipped yet, here's what works today:

1. **Cap concurrent spawns via orchestrator rules** (prompt-level, not enforced, but effective with cooperative agents):
   ```bash
   ao project set-config <project-id> --orchestrator-rules "Never spawn more than 3 concurrent worker sessions. Wait for a worker to finish before spawning the next. Use ao batch-spawn with --max 3."
   ```

2. **For 8 GB Mac specifically:** limit to 2-3 concurrent workers max. Each Claude Code/Codex session uses ~1.5-2 GB. The orchestrator + daemon + OS need ~4 GB alone.

3. **Periodic manual cleanup between runs:**
   ```bash
   ao session cleanup --project <project-id>    # reclaims terminated sessions' worktrees
   git worktree prune                            # cleans stale worktree metadata
   ```

4. **Run large overnight orchestrator jobs on Linux with more RAM.** The reporter confirmed Fedora works better — likely just more headroom.

### 1. Port `maxConcurrentSessions` to the Go rewrite (highest impact)
- Add `MaxConcurrentSessions int` to `config.Config` (env: `AO_MAX_CONCURRENT_SESSIONS`, default 0 = unlimited for back-compat).
- Enforce in `session_manager.Spawn` / `startSession`: count active (non-terminated) sessions before creating worktree + runtime; reject or queue if at limit.
- Surface in project config (`--max-concurrent-sessions` on `ao project set-config`).
- Bake a sane default (e.g., 4) into the orchestrator prompt guidance.

This is the single highest-ROI fix. It directly prevents the OOM cascade.

### 2. Harden orchestrator crash recovery
- In `reconcileLive`, when the runtime probe says "alive" but the session has been inactive beyond a staleness threshold, probe the **agent process** (not just tmux) before adopting.
- When `saveAndTeardownOne` fails, retry with a fallback preserve strategy before falling to unmarked termination.
- Add a UI path to respawn an orchestrator for a project even when the old session row is in an unrecoverable terminated state.

### 3. Add periodic worktree GC
- Wire the existing `session_cleanup_facts` dead code (issue #3402) or add a simpler background reaper that periodically calls `Cleanup` for terminated sessions.
- On daemon startup `Reconcile`, optionally reap worktrees for sessions terminated more than N hours ago.

## Impact

- **8 GB Mac users:** orchestrator is effectively unusable for multi-worker runs without manual concurrency limiting. Crash + unrecoverable session = data loss (conversation context, work-in-progress).
- **All platforms:** worktree accumulation wastes disk and degrades git performance over time.
- **Adoption:** the reporter has already scaled back to "small-range tasks only" — this is actively limiting AO's core value proposition (parallel agent orchestration).

## Related

- [#916](https://github.com/Untrivial-ai/agent-orchestrator/issues/916) — original `maxConcurrentSessions` request (CLOSED — fix never merged into Go rewrite)
- [PR #918](https://github.com/Untrivial-ai/agent-orchestrator/pull/918) — the unmerged fix (CLOSED, `mergedAt: null`)
- [#2811](https://github.com/Untrivial-ai/agent-orchestrator/issues/2811) — merge-terminated sessions leave orphaned worktrees (OPEN)
- [#3402](https://github.com/Untrivial-ai/agent-orchestrator/issues/3402) — cleanup non-idempotent, `session_cleanup_facts` is dead code (OPEN)
- [#3537](https://github.com/Untrivial-ai/agent-orchestrator/issues/3537) — daemon crash leaves workers running as orphans (OPEN)
- [#3649](https://github.com/Untrivial-ai/agent-orchestrator/issues/3649) — spawn failures leave unrecoverable `no_signal` sessions + orphaned worktrees (OPEN)
- [#2501](https://github.com/Untrivial-ai/agent-orchestrator/issues/2501) — orchestrator terminated mid-turn by stale-activity/probe race (OPEN)
- [#2344](https://github.com/Untrivial-ai/agent-orchestrator/issues/2344) — tracker intake hardening: spawn cap, rate-limit budget (OPEN)

AgentWrapper:
Hardens the orchestrator/session lifecycle so AO survives large multi-worker runs. Fixes the #3921 cluster at its root causes: unbounded spawn concurrency, unrecoverable dead orchestrators, and terminal-state teardown that leaked worktrees/branches with no GC to reclaim them.

Closes #3921. Closes #2811. Closes #3402. Addresses #2523 (memory ceiling; the orphan-process reaper already shipped in tmux `Destroy`).

## Root causes → fixes

**A. No concurrency cap in the Go rewrite (#3921 factor 1, the OOM driver).**
`AO_MAX_CONCURRENT_SESSIONS` (daemon-wide) and `maxConcurrentSessions` (per-project config, `ao project set-config --max-concurrent-sessions`) now cap concurrent non-terminated sessions. Worker spawns over the cap fail with a typed `SESSION_CONCURRENCY_LIMIT` 409 **before any durable state is created**; the count-then-create window is serialized so racing spawns can't overshoot. Orchestrator spawns are exempt so a project at the cap can always be recovered. Boot-time `RestoreAll` honors the same cap, so a restart can't resurrect a 40-worker pile-up into the same OOM that killed it. Default is 0 (unlimited) for back-compat; the spawn skill doc now tells orchestrators to keep 3-4 workers on 8 GB machines.

**B. A dead orchestrator was unrecoverable (#3921 factor 2, the "delete the project" symptom).**
Two independent fixes:
1. *Crash finalization.* Reaper-observed runtime death used to only flip `is_terminated` — the worktree stayed on disk with the canonical orchestrator branch checked out, so every replacement spawn failed forever. The LCM now late-binds a crash finalizer that runs the shutdown-style capture-then-destroy sequence: uncommitted work preserved at `refs/ao/preserved/<id>`, restore marker recorded, runtime + orphaned pane children reaped, worktree force-removed. The session stays restorable (user-driven Restore now also replays the preserved ref, previously only boot restore did) and its branch is freed.
2. *Spawn-time release.* Even when teardown failed (disk pressure) or predates the finalizer, spawning an orchestrator now first releases terminated orchestrators' leftover worktrees (capture first, then force-remove, markers consumed). A dead orchestrator can no longer wedge a project.

**C. No worktree/branch GC; cleanup non-idempotent and unobservable (#3402, #2811, #3921 factor 3).**
- The `session_cleanup_facts` layer (migration 0030, previously dead code) gets its production callers: every cleanup attempt persists a durable disposition (`removed` / `preserved_dirty` / `pending` / `failed`) with attempt bookkeeping, capped backoff, and a machine failure code.
- `ao session cleanup` is now idempotent: sessions already reclaimed under the current generation are skipped instead of re-listed forever; `cleanup_generation` is bumped when a session revives so stale facts never shield a re-termination.
- A new hourly `RunTerminalResourceGC` daemon pass walks `ListTerminalCleanupCandidates` and reclaims worktrees leaked by *any* terminal path. Dirty worktrees are never force-deleted (park as `preserved_dirty`; manual cleanup is the user-triggered retry).
- Cleanup, the hourly GC, the crash finalizer, and the spawn-time terminated-orchestrator release all take the session's exclusive operation gate, so no unattended pass can destroy a worktree under a concurrent Restore.
- Tracker-driven termination (issue Done/Cancelled) now routes through the full Kill teardown instead of a bare `MarkTerminated` (#2811's remaining path; merge termination was already fixed by #2980's completion terminator).

**D. Per-spawn resource guardrails (#2523).**
Opt-in `agentConfig.maxMemoryMB` wraps the agent launch (inside the AO supervisor) in `sh -c 'ulimit -v …; exec …'`, so a runaway child is OOM-killed instead of swapping the machine out from under the fleet. Applied on spawn and restore/resume; no-op on Windows. Surfaced as `--max-memory-mb`.

**E. Prior lifecycle-audit leftovers.**
- Finding 4: `terminate_on_pr_merge` is removed from the full-row `UpdateSession`, so a stale lifecycle read-modify-write can never silently revert the user's terminate-on-merge flip (regression test pauses between read and stale commit).
- Finding 5: the inspector Activity timeline sorts historical milestones by actual event timestamps (test: child PR opened after root merged).
- Finding 6: the command palette requires `sessionIsActive` before promoting merge-zone sessions into Needs attention; terminated merged sessions stay search-only.
- Findings 1 (scratch resume) and 3 (Resume/Kill arbitration) were verified already fixed on `main` (project-kind-aware resume validation; the shared exclusive session-operation gate).

## Before / after

| Scenario | Before | After |
|---|---|---|
| Orchestrator spawns 40 workers overnight on 8 GB | unbounded spawns → OOM killer takes the orchestrator/daemon | spawn 5+ (configured cap) refused with `SESSION_CONCURRENCY_LIMIT`; orchestrator waits and retries |
| Orchestrator runtime OOM-killed mid-run | terminated flag only; worktree keeps canonical branch; replacement spawn fails forever; user deletes the project | work captured, worktree freed, restore marker written; "new orchestrator" click just works; old one restorable with its work replayed |
| PR merged / tracker issue closed with terminate-on | worktree orphaned on disk forever (`MarkTerminated` only / no GC) | full Kill teardown; anything left behind reclaimed by the hourly GC with durable facts |
| `ao session cleanup` run twice | same sessions re-listed and re-torn-down every run; failures opaque | second run is a no-op; every attempt has a persisted disposition + failure code |
| Daemon reboot after mass worker crash | RestoreAll relaunches everything at once → OOM again | restores up to the cap; the rest keep their markers for later |

## How the load failure was reproduced

I couldn't OOM an 8 GB Mac in CI, so the reported sequence is reproduced deterministically at the integration level against the real SQLite store (`backend/internal/integration/`):

- `TestDeadOrchestratorRecoverableWithoutProjectDeletion` — spawn orchestrator → reaper reports runtime dead → assert worktree freed + marker written → replacement spawn succeeds and supersedes the old markers. Fails on `main` (worktree survives, marker absent).
- `TestOrchestratorSpawnReleasesLeftoverWorktree` — the harsher variant where crash teardown never ran: terminated row + worktree intact on disk; the replacement spawn reclaims it. Fails on `main`.
- `TestCleanupIsIdempotentAndRecordsFacts`, `TestTerminalResourceGCReclaimsLeakedWorktrees`, `TestTerminalResourceGCPreservesDirtyWorktrees`, `TestCleanupFactsGoStaleWhenSessionRevives` — the #3402 GC/idempotency matrix.
- Concurrency cap: `concurrency_test.go` (cap refusal with no durable residue, slot freed on termination, orchestrator exemption, per-project override).
- Crash finalizer: `crash_finalize_test.go` (lifecycle hook ordering + idempotency), `finalize_crashed_test.go` (capture/teardown/restore round-trip, live/stale guards).

## Tests run

- `cd backend && go vet ./... && go test ./...` — pass, except 6 pre-existing `internal/cli` spawn-context tests that fail identically on the base commit in this environment (`daemon returned HTTP 404` against the stub daemon; unrelated to this change).
- `cd frontend && npm run typecheck` — pass.
- `npx vitest run` on the touched renderer suites (`command-palette.test.ts`, `SessionInspector.test.tsx`) — 107/107 pass.
- `npm run sqlc` and `npm run api` regenerated; generated files untouched by hand.

## Known risks / follow-ups

- The concurrency cap defaults to **unlimited** for back-compat; the reporter's machine still needs `AO_MAX_CONCURRENT_SESSIONS=3` (or the project config) set. A memory-aware default is a possible follow-up.
- Crash finalization writes the same boot-restorable marker as graceful shutdown, so a crashed session auto-restores at the next boot (bounded by the cap). If auto-restore of crashed workers proves surprising, a distinct marker state is a small follow-up.
- `maxMemoryMB` is not applied on the agent-*switch* launch path (rare; follow-up).
- The prior audit's finding 2 (tmux respawn committed but probe fails, leaving an unfenced generation) is NOT addressed here; it needs its own ownership-boundary work.


i-trytoohard:

i-trytoohard:
@AgentWrapper Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@AgentWrapper Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@AgentWrapper Closing as stale: no activity from you for 34 days. The branch stays safe to reopen if you pick this up again.

Vansh-Singhal:
A tested current-main follow-up is now available in #5649. It selectively ports the safe lifecycle fixes from this PR, adds the cleanup/restore race protections found during review, and intentionally excludes the Unix-only `maxMemoryMB` wrapper. Please use #5649 as the merge candidate and keep this PR as historical context.

PR #3988, opened 2026-08-14
Document direct review status sources

codebanditssss:
Documents the direct review status sources used by the Kanban board and AO Reviews view.\n\nKanban continues to consume daemon-derived session status, while Reviews consumes per-PR review and run facts. The broader frontend presentation abstraction and speculative case matrix were removed.\n\nChecks:\n- npm test -- --run src/renderer/lib/session-reviews.test.ts src/renderer/components/SessionsBoard.test.tsx\n- npm run typecheck


neversettle17-101 (changes requested):
Image
1. The toggle should be at the end of the review controls
2. The font sizing should be uniform across.
3. Can you enhance the UI for the dropdown ? the greyed out box doesn't look well.
4. Let us keep the reviews in the summary section only for now, when we add the review summary part we can introduce this section.

neversettle17-101 (commented):
Question/concern: from the current product need, the derived status seems most useful for the Kanban card/status display rather than the Reviews view. In the AO agent review area, it may be enough to show the direct review fact — approved vs. not approved / changes requested — without introducing a broader presentation abstraction.

Unless there is another planned consumer or requirement for this shared abstraction, could we either wire this derived status into the Kanban flow where it will be displayed, or simplify the review UI contract to the direct approval state needed there?

codebanditssss:
Simplified in 1d7d94368. The broad frontend presentation abstraction has been removed. Kanban already consumes the daemon-derived WorkspaceSession.status for Approved, Changes requested, and Review pending, so adding a Reviews request per card would duplicate the source and create N+1 fetching. The AO Reviews view remains on direct PRReviewState/run facts. The contract now documents those two existing consumers and leaves CI, mergeability, comments, cycles, and injection outside this scope.

neversettle17-101:
Lets close this as it is not required

PR #3990, opened 2026-08-14
chore(terminal): remove Clear context-menu item that never persisted on agent TUI panes

anmol0b:
## What
Removes the "Clear" context-menu item from the terminal pane (`XtermTerminal.tsx`), along with its associated test coverage and now-unused i18n strings across all locales.

## Why
The Clear action only cleared the local xterm.js buffer (`term.clear()`). For shell panes this looked correct, but for agent TUI panes (Codex, Claude Code, etc.) the agent process owns and continuously repaints its own transcript from memory on every redraw (resize, tick, etc.), so the cleared view would immediately repopulate — making the button appear broken/inconsistent depending on pane type.

Investigated a proper fix (runtime-level scrollback truncation for tmux/conpty, plus a provider-aware "Clear conversation" action sending `/clear` to the agent). On review, decided against shipping either:
- Runtime-level clear only ever fixes the shell-pane case; it can't stop a TUI from repainting its own transcript.
- A "Clear conversation" button would just duplicate a command (`/clear`) every supported agent CLI already exposes natively — users can type it directly in the terminal today, no UI needed.

Given that, the button wasn't earning its complexity. Removing it entirely rather than shipping a partially-working or duplicative feature.

## How
- Removed the "Clear" item from the terminal context menu and its handler in `XtermTerminal.tsx`.
- Removed the corresponding test case in `XtermTerminal.test.tsx`.
- Removed the now-unused `terminal.clear` (or equivalent) i18n key from all locale files (`en`, `de`, `es`, `fr`, `ja`, `ko`, `pt-BR`, `zh-CN`).
- No backend changes — this is a pure UI removal.
- Intentional omission: no replacement action was added. Users who want to clear an agent's transcript/context can already type `/clear` directly in the terminal, which is the native, provider-supported way to do it.

## Testing
- Targeted vitest run for `XtermTerminal.test.tsx`
- Manual: opened the terminal context menu on both a shell pane and an agent (Codex) pane in dev — Clear item no longer appears in either.

before :

Screenshot 2026-08-14 at 5 03 13 PM

after:

changes(suggested) - two diffrent buttons
Screenshot 2026-08-14 at 11 19 45 AM

changes(approved by @illegalcall)
Screenshot 2026-08-14 at 4 59 53 PM

## Checklist

### DIFF
--- a/frontend/src/renderer/components/XtermTerminal.test.tsx
+++ b/frontend/src/renderer/components/XtermTerminal.test.tsx
@@ -461,7 +461,7 @@ describe("XtermTerminal", () => {
 		Object.defineProperty(document, "fullscreenElement", { configurable: true, value: null });
 	});
 
-	it("runs context menu copy, select all, and clear against the xterm instance", async () => {
+	it("runs context menu copy and select all against the xterm instance", async () => {
 		const { container } = render(<XtermTerminal theme="dark" />);
 		const host = container.firstElementChild!;
 		state.lastTerminal!.selection = "menu copy";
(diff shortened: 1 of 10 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #3992, opened 2026-08-14
fix(chat): make Stop authoritative for durable running turns

### LINKED ISSUE #3749: bug(chat): "Stop turn" fails with CHAT_NO_ACTIVE_TURN while UI shows a running turn — in-memory/durable turn-state desync

illegalcall:
Fixes https://github.com/Untrivial-ai/agent-orchestrator/issues/3749

## Summary

- Treat the durable running turn rendered by the UI as actionable when Stop finds no in-memory turn.
- Make Stop linearize with dispatch under `sendMu`, capture the exact pre-Stop target and queue cutoff, and wait for that turn acknowledgment without following ownership to newer work.
- Reconcile every running turn visible on the active branch, including overlapping root and nested provider turns, while keeping provider cancellation best-effort.
- Move controller-state memory changes after the durable projection commits so a rollback cannot recreate the memory/SQLite split.
- Refetch the desktop conversation after an interrupt error so stale cached Working state is not left on screen.

PR #3890 is merged into `main` and supplies the nested-turn ownership rules this recovery path relies on. This PR is based directly on current `main` and contains only the focused recovery work for #3749.

The implementation was adapted from PR #3752 by @bluwwi and extracted from the tested recovery work in PR #3986. Contributor attribution is preserved in the commit history.

## Regression coverage

- The exact #3749 projection-rollback sequence: SQLite remains `running`, the provider refuses Stop, and reconciliation settles the durable state.
- A durable running turn with no in-memory pending ID.
- A genuine absence of both durable and in-memory running work.
- Stop racing an in-progress dispatch waits for provider acknowledgment.
- Root and nested visible running turns are both settled, while Stop targets the same oldest/root turn shown by the UI.
- Queue cancellation during recovery without accidental dispatch of pre-Stop work.
- A prompt submitted after Stop survives durable fallback and provider-refusal recovery.
- A completion committed before provider refusal is not overwritten.
- Controller state changes only after projection commit.
- Recovered interrupted turns have no error message.
- Desktop interrupt errors invalidate the conversation query.

## Verification

- `cd backend && go test -race ./internal/service/chat -count=1`
- focused recovery/concurrency tests repeated under the race detector
- `npm run sqlc` with no generated drift
- focused frontend hook suite: 13 tests passed
- golangci-lint v2.12.2: 0 issues with an isolated cache

### DIFF
--- a/backend/internal/service/chat/controller.go
+++ b/backend/internal/service/chat/controller.go
@@ -47,7 +47,7 @@ type Store interface {
 	SettleTurn(ctx context.Context, conversationID, providerTurnID string, state domain.TurnState, errMessage string, now time.Time) error
 	SettleTurnByID(ctx context.Context, turnID string, state domain.TurnState, errMessage string, now time.Time) error
 	SettleOrphanedTurns(ctx context.Context, session domain.SessionID, now time.Time) error
-	GetRunningTurn(ctx context.Context, conversationID string) (id, providerTurnID string, ok bool, err error)
+	ListVisibleRunningTurnProviderIDs(ctx context.Context, conversationID string) ([]string, error)
 
 	SetConversationSettings(ctx context.Context, conversationID string, settings domain.ConversationSettings, now time.Time) error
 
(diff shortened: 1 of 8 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):
LGTM

PR #3993, opened 2026-08-14
fix: simplify windows titlebar nav

JAYATIAHUJA:
## What

Changed the Windows titlebar only.

## Why

The top Windows bar had too many menu items.

## Changes

- removed File, Edit, and Window from the Windows titlebar
- added back and forward arrows beside the sidebar button
- kept View and Help same

## Screenshot 
image

## Test

- `npm --prefix frontend run test -- WindowTitlebar.test.tsx`


Vaibhaav-Tiwari (approved):
looks good 

  Two small cleanup items nothing broken:
  1. copy-pasted a navigation hook from another file instead of sharing it — and lost a useful comment explaining a tricky bug in the process.
  2. Deleting those menus left ~80 unused translation strings behind (10 keys × 8 languages) that nobody cleaned up.

nothing that would block anything

JAYATIAHUJA:
Updated this.

Shared the forward history hook now, so copied logic is gone.

Also removed the unused titlebar menu strings for File, Edit, Window and their old items.

PR #3997, opened 2026-08-14
fix(spawn): pre-flight the agent binary instead of failing mid-spawn

### LINKED ISSUE #3941: Spawn: pre-flight the agent binary + runtime before allowing spawn (AGENT_BINARY_NOT_FOUND / RUNTIME_PREREQUISITE_MISSING)
Part of #3873. Related: #3434 (tmux prereq only checked at spawn time).

Biggest single cause of spawn failures (~47%, 14d in PostHog): AGENT_BINARY_NOT_FOUND (531) + RUNTIME_PREREQUISITE_MISSING (878). Users click spawn with an agent whose CLI is not installed, or a required runtime is missing, and it fails after the fact instead of being caught up front. Once the agent IS installed, per-harness failure is only ~4-5%, so this is a setup gap, not the spawn engine.

Proposed: before enabling/accepting a spawn, check the selected harness binary and its runtime prerequisite exist. If not, block with a clear 'install X' message (and a help link) instead of letting the spawn fail. Removes ~47% of failures.

Pulkit7070:
Closes #3941.

A missing agent CLI was the single largest cause of spawn failures (~47% of them over 14 days: `AGENT_BINARY_NOT_FOUND` + `RUNTIME_PREREQUISITE_MISSING`), and it is a setup gap rather than a spawn-engine bug. It was also detected as late as possible: `validateAgentBinary` only runs once argv has been built, which is after the seed session row, the worktree, workspace provisioning (which may run commands like `pnpm install`), and the attachment writes. Every spawn with an uninstalled harness paid for that work, rolled it back, and told the user nothing actionable.

## Daemon: catch it before anything durable exists

`preflightSpawnEnvironment` (new `backend/internal/session_manager/spawn_preflight.go`) runs the terminal-runtime check and a new agent-binary check before `Spawn` creates any state, asking the adapter's own `ports.AgentBinaryResolver` so no workspace or launch command is needed.

It is deliberately one-sided. Only a definitive `ErrAgentBinaryNotFound` blocks; an adapter without a resolver, or a probe that fails for any other reason, is not evidence the CLI is absent, so the pre-launch argv[0] guard stays authoritative there. Chat mode is untouched: it reaches its provider through `PreflightChat`, which probes the driver itself.

Failures are also readable now: missing tmux names the platform install command, an unknown harness points at `ao agent ls`, and the missing-CLI error names the CLI plus the agent docs link.

## Renderer: don't let the click become a spawn request

The daemon refusal is cheap, but the user still spent a round trip to learn their CLI is missing, and the refusal landed in spawn telemetry looking like a broken spawn engine.

`assertAgentInstalled` (new `frontend/src/renderer/lib/agent-install-preflight.ts`) checks the selected harness against the agent catalog and, because that catalog is a boot-time snapshot, re-checks a "missing" verdict against a fresh `POST /api/v1/agents/{agent}/probe` before believing it. Same one-sided rule as the daemon: an unset or unsupported agent, an older daemon without the probe endpoint, or any transport failure defers to the daemon. Blocking a spawn that would have worked is worse than the cheap failure this replaces.

The gate lives inside `spawnOrchestrator` for the same reason the telemetry triad does: the board, topbar, sidebar, command palette, restore dialog, settings, restart, and project-add launchers all reach the daemon through it, so none of them has to remember to check. The worker path is gated in `TaskComposer.createTask`, where the resolved default agent can be one the picker would have disabled.

Refused spawns read as install guidance with an "Agent docs" button rather than as a failed spawn, on the board, the topbar, the sidebar, and the task composer. The sidebar's orchestrator button previously swallowed every spawn failure into `console.error` with no user-visible feedback at all.

## Verification

Verified against the live daemon that the payloads the gate reads are what it expects:

- `GET /api/v1/agents` returns `{supported,installed,authorized}` of `{id,label}`
- `POST /api/v1/agents/goose/probe` returns `{"agent":{"id":"goose","label":"Goose"},"supported":true,"installed":false}`; `claude-code` returns `installed:true`
- `GET /api/v1/projects/{id}` returns `config.orchestrator.agent`

Verified the daemon preflight against the real adapter registry (throwaway test, not committed): of the 26 shipped harnesses, the 23 not installed on that machine were each blocked with **0 session rows and 0 worktrees created** and a message naming their CLI; the 3 that were installed passed the preflight and proceeded.

Tests:

- `backend`: `go build ./...`, `gofmt`, and `go test ./...` match clean HEAD (the `internal/cli` failures are pre-existing and environment-dependent, confirmed by running the same suite at HEAD in an isolated worktree). New `spawn_preflight_test.go` covers missing binary blocked with no row/worktree/runtime, inconclusive probe allowed through, resolver-less adapter skipped, chat mode skipped.
- `frontend`: `tsc --noEmit` clean; 109 renderer test files / 1523 tests pass. New: 8 `agent-install-preflight` cases, 2 `spawn-orchestrator` cases (a blocked spawn issues no `/orchestrators` POST; an unreadable project config still spawns), a `TaskComposer` case asserting the install message, the docs button, and no delegate request, and a `Sidebar` case asserting a refused spawn now reaches the user.
- New i18n keys added to all 8 locale catalogs.

## Notes and follow-ups

- The launcher buttons are not visually disabled ahead of the click. The only pre-click signal is the advisory catalog, and disabling on a stale snapshot would grey out a button that works; the fresh probe on click is strictly better.
- The docs link is the `/docs/plugins/agents` index rather than a per-agent deep link: most per-agent pages 404 (`/agents/crush`, `/agents/prime-agent`, and `/agents/copilot`, which the README links).
- No install *commands* are shipped per agent. The repo holds no such metadata, and inventing 26 of them is worse than a message that names the CLI and links the docs. Adding them as adapter metadata would be a reasonable follow-up.
- `AGENT_REQUIRED` and `UNKNOWN_HARNESS` from the issue's scope note were already validated before any durable state; only their wording changed.
- Adds one PATH-resolution probe (2s ceiling) per TUI spawn, plus two cached GETs before an orchestrator spawn. For node-managed CLIs the probe includes the Volta/nvm/fnm candidate globs, the same work `GET /api/v1/agents` already does.


i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

Pulkit7070:
Closing as superseded and stale. The target issue #3941 is already closed as completed, and the harness-readiness area this PR reworked has since been centralized on main by #4589 (new /api/v1/agents/readiness coordinator), so this branch's spawn_preflight.go and agent-install-preflight.ts approach no longer fits the current architecture and conflicts with main. If the pre-durable-state blocking behavior still has a gap, it should be re-approached on top of the centralized readiness layer rather than reviving this branch.

PR #3998, opened 2026-08-14
fix: classify unclassified spawn internal failures

### LINKED ISSUE #3943: Spawn: ~31% of failures are unclassified 'internal' — classify + investigate in Sentry
Part of #3873. Possibly related: #3855 (spawn always fails), #3220 (unborn-HEAD worktree husk).

~31% of spawn failures (952 of ~3,030 in 14d) land in the catch-all error_kind=internal with no error_code, so we cannot tell why from telemetry alone.

Proposed: pull the matching Sentry stacks, give the common ones real error_codes so they stop landing in the catch-all, and fix the underlying causes. Goal: drive the 'internal' slice toward zero and make future spawn failures self-diagnosing.

ApexYash11:
## Summary
- Classify spawn failures that previously landed in `error_kind=internal` with no `error_code` (~31% of spawn failures in [#3943](https://github.com/Untrivial-ai/agent-orchestrator/issues/3943)).
- Wrap each spawn stage with a sentinel so `ao.session.spawn_failed` always carries a typed `error_code` (workspace, provision, runtime, chat controller, timeout, and a `SPAWN_INTERNAL` catch-all).
- Keep existing sentinels (branch state, missing agent binary, chat preflight) winning over the stage code.

Fixes #3943

## Test plan
- [x] `go test ./internal/service/session` spawn/mapping/telemetry tests
- [x] `go test ./internal/session_manager` wrap + spawn rollback tests
- [x] `go test ./internal/httpd/controllers -run TestSessionsAPI_Spawn`
- [ ] Confirm a failed spawn in PostHog now includes `error_code` instead of an empty internal bucket

### DIFF
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -250,8 +250,9 @@ func (s *Service) spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 	cfg = s.withIssueContext(ctx, cfg, project)
 	rec, promptBytes, systemPromptBytes, err := s.manager.Spawn(ctx, cfg)
 	if err != nil {
-		s.emitSpawnFailed(cfg, err, s.now().Sub(start).Milliseconds())
-		return domain.Session{}, 0, 0, toAPIError(err)
+		apiErr := toSpawnAPIError(err)
+		s.emitSpawnFailed(cfg, apiErr, s.now().Sub(start).Milliseconds())
+		return domain.Session{}, 0, 0, apiErr
 	}
 	s.emitSpawned(rec, s.now().Sub(start).Milliseconds())
 	if firstSession {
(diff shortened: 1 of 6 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/service/session/service.go
```diff
@@ -993,6 +994,59 @@ func toAPIError(err error) error {
...
+// preflight) still win via toAPIError first.
+func toSpawnAPIError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if mapped := toAPIError(err); !errors.Is(mapped, err) {
+		return mapped
+	}
+	switch {
+	case errors.Is(err, context.DeadlineExceeded):
```
Pulkit7070, line 1014: Precedence nuance worth confirming: because wrapSpawnStage keeps both the stage sentinel and the inner error in the chain (multi-%w), a stage that fails due to a timeout matches both `context.DeadlineExceeded` and its stage sentinel (e.g. ErrRuntimeCreate). Since DeadlineExceeded is checked first, it is classified as SPAWN_TIMEOUT rather than the stage-specific code. That is defensible (a timeout is usually the more actionable signal), just flagging that stage-specific timeouts collapse into the single SPAWN_TIMEOUT bucket, which may hide which stage stalled. Intended?

backend/internal/service/session/service.go
```diff
@@ -250,8 +250,9 @@ func (s *Service) spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 	cfg = s.withIssueContext(ctx, cfg, project)
 	rec, promptBytes, systemPromptBytes, err := s.manager.Spawn(ctx, cfg)
 	if err != nil {
-		s.emitSpawnFailed(cfg, err, s.now().Sub(start).Milliseconds())
-		return domain.Session{}, 0, 0, toAPIError(err)
+		apiErr := toSpawnAPIError(err)
+		s.emitSpawnFailed(cfg, apiErr, s.now().Sub(start).Milliseconds())
```
Pulkit7070, line 254: Solid fix, and I confirmed all 16 stage sentinels are both defined and wrapped at their stage (no dead codes), the precedence is right (existing sentinels win via toAPIError, then the stage switch, then SPAWN_INTERNAL), and telemetry now carries a typed error_code across every stage.

One latent-fragility note about this emitSpawnFailed call: internally emitSpawnFailed re-classifies with `toAPIError`, not `toSpawnAPIError`. It works today only because you pre-map here and pass the already-classified apiErr in, so that internal call is effectively idempotent. But if a future caller passes a raw spawn error to emitSpawnFailed, it would silently land back in the empty internal bucket, the exact regression this PR fixes. Having emitSpawnFailed call `toSpawnAPIError` (or take the already-mapped error and skip re-mapping) would remove that footgun.

backend/internal/service/session/service.go
ApexYash11: Intended. A timeout is the more actionable signal, so `DeadlineExceeded` still wins over the stage sentinel. Added a test that `wrapSpawnStage(ErrRuntimeCreate, context.DeadlineExceeded)` maps to `SPAWN_TIMEOUT`.

backend/internal/service/session/service.go
ApexYash11: Fixed in ee02d9233. `emitSpawnFailed` now classifies with `toSpawnAPIError`, and `toSpawnAPIError` returns an already-mapped `*apierr.Error` as-is so a raw stage sentinel and a pre-mapped API error both keep a typed `error_code`.

Pulkit7070 (approved):
LGTM. Both notes resolved: emitSpawnFailed now classifies with toSpawnAPIError, and toSpawnAPIError's errors.As(&already) passthrough keeps an already-mapped code intact while still classifying raw errors. Verified under -race: 4000 concurrent spawn failures (raw and pre-mapped) all emit a typed error_code, 0 empty-bucket. Timeout-wins-over-stage precedence confirmed intended and covered by a test. CI green.

PR #4000, opened 2026-08-14
feat(telemetry): track per-session agent token usage and cost

Pulkit7070:
## What

Adds an end-to-end pipeline to capture agent LLM token usage and dollar
cost per session, sliceable in PostHog by user, session, and harness.
There is no token/cost data captured anywhere today, so this is a new
pipeline rather than a new chart.

## Events

- `ao.session.turn_usage` (aliased to `ao.v2.session.turn_usage`): one per
  model turn, with `model`, `input_tokens`, `output_tokens`,
  `cache_read_tokens`, `cache_write_tokens`, `total_tokens`, `cost_usd`,
  `model_priced`, `harness`, `kind`.
- `ao.session.usage` (aliased to `ao.v2.session.usage`): one per session at
  end, same fields plus `turn_count` and `duration_ms`.

Both auto-carry the install `distinct_id` and hashed `session_id`, which is
what enables the per-user / per-session / per-harness cuts.

## How it works

1. The claude-code `Stop` / `SessionEnd` hook payload carries
   `transcript_path`. `ao hooks claude-code stop|session-end` parses the
   transcript JSONL for per-turn usage, resuming from a persisted byte
   offset so a turn is never counted twice, and POSTs to the daemon.
2. `POST /api/v1/sessions/{id}/usage` decodes the report; token counts are
   clamped non-negative and the body is size-capped.
3. `Service.RecordUsage` reads harness/kind/project from the stored session
   record (so tags cannot be spoofed by the caller), prices each turn from a
   list-price table, and emits the two events. Unknown models record tokens
   with `cost_usd: 0` and `model_priced: false`.

The hook path is best-effort throughout: a missing transcript, read error,
or unreachable daemon all exit quietly so the agent is never disrupted.

## Pricing

Single editable table of Anthropic list prices (USD per million tokens),
with cache-write at 1.25x input and cache-read at 0.1x input:

| Model  | input | output | cache write | cache read |
|--------|-------|--------|-------------|------------|
| Opus   | 15    | 75     | 18.75       | 1.50       |
| Sonnet | 3     | 15     | 3.75        | 0.30       |
| Haiku  | 0.80  | 4      | 1.00        | 0.08       |

## Scope

claude-code only for now; other harnesses emit no usage until their
transcript parsers are added, so a "by harness" view shows one slice until
then. Per-turn events are unaggregated by design (both per-turn and summary
were requested); a long session emits one event per turn.

## Verification

Ran the daemon against an isolated data dir and drove the endpoint with a
[...]

PR #4003, opened 2026-08-14
feat(telemetry): price agent usage and export cost to PostHog

Pulkit7070:
## What

main already collects per-session token aggregates locally
(`internal/service/usage` + `GET /api/v1/usage/sessions`), but it has
**no dollar cost** and emits **nothing to PostHog**, so cost can't be
seen across users, days, sessions, or harnesses. This adds that missing
half, and only that half, without duplicating collection.

## How it works

The ingestion pipeline applies each parsed usage chunk exactly once
(a durable source cursor guards re-processing). This change hangs a
priced telemetry emit off that same exactly-once seam:

1. `observe/usage` gains an optional `OnUsageApplied(sessionID, events)`
   callback on the ingestor, invoked only after `ApplyUsageChunk`
   commits. Because the cursor has advanced, each event fires the
   callback exactly once, **no dedupe marker or DB migration needed**.
   The package stays free of any telemetry/pricing dependency.
2. `service/usagecost.Emitter` prices each event (`internal/pricing`)
   and emits `ao.session.turn_usage`. Fresh uncached input is billed at
   the input rate; cache read/write are separate line items. Harness,
   kind, and project come from the stored session record, so tags can't
   be spoofed. Unknown models emit tokens with `cost_usd: 0` /
   `model_priced: false`.
3. The daemon wires the emitter into the ingestor; the event is
   allowlisted with its `ao.v2.session.turn_usage` alias.

Best-effort throughout: a nil sink or a failed session lookup emits
nothing and never blocks or errors ingestion.

## Pricing

Single editable table of Anthropic list prices (USD per million tokens),
cache-write at 1.25x input, cache-read at 0.1x input:

| Model  | input | output | cache write | cache read |
|--------|-------|--------|-------------|------------|
| Opus   | 15    | 75     | 18.75       | 1.50       |
| Sonnet | 3     | 15     | 3.75        | 0.30       |
| Haiku  | 0.80  | 4      | 1.00        | 0.08       |

## Why this shape

An earlier attempt (#4000, now closed) reimplemented collection with a
separate parser/endpoint/hook and collided with main's shipped feature.
This version reuses main's collector and adds only pricing + emission.

## Verification

- The `OnUsageApplied` callback fires with the parsed events and the
  correct session id after a real ingest (integration test over the
  sqlite store, a real transcript file, and the durable cursor).
- The emitter prices Opus 1M uncached-in + 1M out at $90, records an
  unknown model at $0 / `model_priced:false`, tags harness from the
  record (not the caller), and no-ops on a nil sink or empty events.
- `pricing` unit tests (families, unknown, negative clamp).
- The daemon boots to ready with the emitter wired and shuts down clean.
- `gofmt`, `go vet`, and the touched-package tests pass.

Not driven live end-to-end to PostHog, that would require injecting into
the real `~/.claude` transcript roots; the new code paths are exercised
through the ingest pipeline and the emitter in tests.

### DIFF
--- a/backend/internal/adapters/telemetry/posthog.go
+++ b/backend/internal/adapters/telemetry/posthog.go
@@ -34,6 +34,7 @@ var remoteEventNameAliases = map[string]string{
 	"ao.renderer.loaded":         "ao.v2.renderer.loaded",
 	"ao.renderer.api_error":      "ao.v2.renderer.api_error",
 	"ao.renderer.daemon_failure": "ao.v2.renderer.daemon_failure",
+	"ao.session.turn_usage":      "ao.v2.session.turn_usage",
 }
 
 var remoteCommandTokens = map[string]struct{}{
@@ -187,6 +188,19 @@ var remotePayloadAllowlist = map[string]map[string]struct{}{
 		"harness":     {},
 		"kind":        {},
 		"operation":   {},
+	},
+	"ao.session.turn_usage": {
+		"harness":               {},
+		"kind":                  {},
+		"model":                 {},
+		"model_priced":          {},
+		"input_tokens":          {},
+		"uncached_input_tokens": {},
+		"output_tokens":         {},
+		"cache_read_tokens":     {},
+		"cache_write_tokens":    {},
+		"total_tokens":          {},
+		"cost_usd":              {},
 	},
 	"ao.session.spawned": {
 		"duration_ms": {},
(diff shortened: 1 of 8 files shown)


i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 20+ days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 30 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4005, opened 2026-08-14
fix(chat): keep fitting files after an oversized batch item

### LINKED ISSUE #4080: Chat attachment batch drops valid files after one exceeds the remaining size budget
## Bug

When a user selects several attachments at once, encountering one file that exceeds the remaining 25 MiB batch budget stops processing the batch. Valid smaller files later in the same selection are silently discarded even when they fit.

Source: reproduced while reviewing [PR #4005](https://github.com/Untrivial-ai/agent-orchestrator/pull/4005).

## Reproduction

Analyzed against main at `d380279b5`.

Select these files in one native Electron file-picker operation, in this order:

1. `a-9mb.txt` — 9 MiB
2. `b-9mb.txt` — 9 MiB
3. `c-rejected-9mb.txt` — 9 MiB and does not fit
4. `d-later-5mb.txt` — 5 MiB and does fit

Expected: the first, second, and fourth files remain staged; the third is rejected with the existing inline size feedback.

Actual: only the first and second files remain. Processing aborts at the third file, silently losing the fourth.

## Root cause

The production `useFileAttachments.addFiles` loop uses `break` for the total-byte cap. That is correct for the count cap, where no subsequent file can fit, but incorrect for a byte cap, where a later smaller file may fit.

## Impact

This can silently lose user-selected input during paste, drop, or file selection. The workaround is to reselect each missing file individually.

Priority: P2. Confidence: high.

## A/B evidence

**Before — current main silently drops the later fitting file**

Before: later fitting file is silently dropped

**After — PR #4005 rejects only the non-fitting file and keeps the later one**

After: later fitting file remains staged

Both captures use the real Electron desktop app, the native file input, browser `FileReader`, identical files, and fresh isolated AO data.

vanshx999:
Fixes #4080.

## What changed

The production `useFileAttachments.addFiles` loop now skips only a file that exceeds the remaining 25 MiB budget and continues evaluating later files in the same paste, drop, or file-picker batch. The count cap intentionally still stops the loop because no subsequent file can fit once the attachment count is full.

This PR was rebased onto current `main` and narrowed during review to this one issue. The unrelated attachment-only steering fix has moved to its own issue and PR. Changes to the unused `useImageAttachments` hook were removed so this stays on the live Chat path.

## Tests

- Regression test proves a 9 + 9 + rejected 9 + later 5 MiB batch retains the first, second, and fourth files.
- Focused test: 5 passed.
- Frontend typecheck: passed.
- Full frontend suite: 191 files; 2,505 passed and 1 skipped.

## Real Electron A/B evidence

The same four files were selected in the same order through the native file input, with a fresh isolated `AO_DATA_DIR` for each build. These screenshots are attached as review evidence and are not part of the PR code diff.

### Before — current main aborts at the rejected file

Before: later fitting file is silently dropped

Only the first two 9 MiB files remain. The later valid 5 MiB file is silently lost.

### After — this PR keeps evaluating the batch

After: later fitting file remains staged

The non-fitting third file still gets inline feedback, while the later 5 MiB file remains staged for a valid 23 MiB total.


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Deep review and real-app verification are complete.

Review changes:
- rebased onto current `main`;
- narrowed the PR to the live `useFileAttachments` batch-size bug;
- added dedicated issue #4080 and `Fixes #4080`;
- moved the independent steering defect to #4081 and a separate PR;
- removed the unused `useImageAttachments` edits;
- retained `break` for the count cap and changed only the byte-budget branch to `continue`.

Verification:
- focused regression: 5/5 passed;
- `npm run typecheck`: passed;
- full frontend suite: 191 files, 2,505 passed, 1 skipped;
- real Electron A/B reproduction: passed.

The screenshots are attached as PR review evidence only; they are not committed in the implementation diff.

| Baseline `main` | This PR |
|---|---|
| Before | After |
| The later fitting 5 MiB file is lost after a rejected 9 MiB file. | The rejected file still shows feedback and the later 5 MiB file remains staged. |

illegalcall (approved):

PR #4007, opened 2026-08-14
fix(desktop): build the AppImage with a non-legacy runtime so it starts without libfuse2

### LINKED ISSUE #4006: Linux AppImage won't start on a clean Arch install — it's built with the old FUSE2 runtime
Ran into this while setting up AO on a fresh Arch box (Omarchy). The Linux AppImage from the releases page just doesn't start:

```
$ ./agent-orchestrator-linux-x64.AppImage
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run.
You might still be able to extract the contents of this AppImage
if you run it with the --appimage-extract option.
```

Exit code 1, no window, nothing else. `ao start` hits the same wall since it downloads and runs that same artifact.

### Why it happens

The AppImage is being built with electron-builder's old FUSE2 runtime, which needs `libfuse.so.2` at runtime. Arch dropped that from the default install a long time ago — it only has fuse3 now.

It comes down to one missing config key. `frontend/makers/maker-appimage.ts` passes a config to `buildForge` without `toolsets`:

```ts
config: {
  appId: cfg.appId,
  productName: cfg.productName ?? appName,
  protocols: cfg.protocols,
  directories: { output },
  publish: null,
  linux: { ...(cfg.icon ? { icon: cfg.icon } : {}) },
  appImage: { ...cfg.appImage },
}
```

app-builder-lib treats a missing `toolsets` as "use the legacy runtime". From `out/targets/appimage/AppImageTarget.js`:

```js
const appimageTool = this.packager.config.toolsets?.appimage;
if (appimageTool == null || appimageTool === "0.0.0") {
    updateInfo = await buildLegacyFuse2AppImage({ ... });
}
```

and `out/toolsets/linux.js` confirms what that means:

```js
const isFuse2 = appimageToolVersion === "0.0.0" || appimageToolVersion == null;
```

which pulls `appimage-12.0.1` — the AppImageKit type-2 runtime that dlopens libfuse2.

It's in the published artifact too, not just theory. Running `strings` over the first chunk of the v0.12.4 release:

```
dlopen(): error loading libfuse.so.2
AppImages require FUSE to run.
Cannot mount AppImage, please check your FUSE setup.
```

### Why nobody's hit this yet

I nearly convinced myself it wasn't real, because the AppImage launched fine on my machine at first. Turned out `fuse2` had been pulled in a few minutes earlier as a transitive dependency of an unrelated `pacman -S` batch. Once I masked `libfuse.so.2` again (bind-mount over it in a throwaway mount namespace, nothing permanent) and re-ran the actual downloaded AppImage, it failed exactly as above.

So plenty of people will never see this — anything that drags in fuse2 hides it. It's clean installs that get bitten, which is unfortunately the first-run path. For what it's worth, Omarchy's own base package list (`omarchy-base.packages`, 151 packages) has no fuse package in it at all.

### Fix

app-builder-lib supports newer static runtimes that don't need libfuse2 — `1.0.2` and `1.0.3` are both in its checksum table. Setting the toolset in the maker should be enough:

```ts
toolsets: { appimage: "1.0.3" },
```

Worth a sanity check on the deb/rpm makers too, though those don't use this code path.

### Environment

Arch Linux (Omarchy), kernel 7.1.4, Hyprland/Wayland, fuse3 only, no fuse2. Tested against the `v0.12.4` release asset `agent-orchestrator-linux-x64.AppImage`.

Happy to pick this up — I'll raise a PR for it.

yashksaini-coder:
## What

Pin the electron-builder appimage toolset to `1.0.3` in `frontend/makers/maker-appimage.ts`, so the published Linux AppImage is built with the static runtime instead of the legacy FUSE2 one.

## Why

Fixes #4006.

`toolsets.appimage` was never set. app-builder-lib treats that as `"0.0.0"` and builds with the AppImageKit runtime, which `dlopen`s `libfuse.so.2` before it can mount its own squashfs. Distros that ship only fuse3 — Arch, and any install without the `fuse2` compat package — can't launch the artifact at all:

```
$ ./agent-orchestrator-linux-x64.AppImage
dlopen(): error loading libfuse.so.2
AppImages require FUSE to run.
```

Exit 1, before Electron starts. `ao start` hits the same wall, since it downloads and runs this same AppImage.

The branch point is `app-builder-lib/out/targets/appimage/AppImageTarget.js:66`:

```js
if (appimageTool == null || appimageTool === "0.0.0") { buildLegacyFuse2AppImage(...) }
else { buildStaticRuntimeAppImage(...) }
```

## How

One config key, plus a test.

**On picking a beta.** `toolsets.appimage` accepts only `"0.0.0" | "1.0.2" | "1.0.3" | null`. `"0.0.0"` is the bug, `null` is treated identically, and both remaining values are documented upstream as betas — so "fix this" and "stay on a stable toolset" are mutually exclusive within this option. `1.0.3` and `1.0.2` ship the *same* `Runtime 20251108`; `1.0.3` is a re-release with an upstream fix, so it strictly dominates.

Alternatives I ruled out:

- `AppImageOptions` exposes only `license` and `compression` — no runtime override.
- `APPIMAGE_TOOLS_PATH` still branches on this same flag for its expected directory layout, so it isn't an escape hatch.
- **Bundling libfuse2 inside the AppImage can't work by construction** — the legacy runtime dlopens it *in order to mount* the squashfs, so anything inside the image is unreachable at that moment.
- Telling Linux users to install `fuse2` contradicts the reason this artifact exists ("a single self-contained executable... with no system package manager").

**Two side effects reviewers should know about**, both from `buildStaticRuntimeAppImage`:

1. **Compression flips gzip → zstd.** The artifact got *smaller* (183,309,089 → 171,448,948 bytes, −6.5%), but the bytes change wholesale, so the first delta update across this release is effectively a full download. A very old `unsquashfs` without zstd support would also fail on a manual `--appimage-extract`; the bundled runtime handles it fine.
2. **The `--no-sandbox` default drops out of the desktop entry** (`AppImageTarget.js:26` only adds it for the legacy toolset). This looks self-healing: the generated `AppRun` probes `unshare -Ur true` and re-adds `--no-sandbox` at runtime when user namespaces aren't available, on both the desktop-entry path and `ao start`'s direct exec. Worth a second opinion from someone who knows the sandbox story better than I do.

Build-time network is unchanged — actually one download fewer. The legacy path fetched `appimage-12.0.1.7z` *and* `7zip-linux-x64.tar.gz`; the static path fetches only `appimage-tools-runtime-20251108.tar.gz`. AppImage type-2 magic is preserved, so appimaged/AppImageLauncher still recognise the file, and both paths emit the embedded block map.

Deliberately out of scope: `toolsets.nsis` and `toolsets.wine` also still default to legacy `"0.0.0"`. Different platforms, different issue.

## Testing

Environment: Arch Linux, fuse3 only, no `fuse2` package, no `/usr/lib/libfuse.so.2`.

**Unit:** `npx vitest run --config vite.renderer.config.ts makers/` → 24 passed. The new test pins the exact version rather than asserting "not 0.0.0", because app-builder-lib treats `null` the same as the legacy toolset, so a looser assertion would stay green on a regression.

**A/B build.** Built this branch with `npx electron-forge make --platform=linux --targets=appimage`, and a control from the same packaged tree with the `toolsets` key omitted:

| | this branch | control (key omitted) |
| --- | --- | --- |
| `strings` over the full artifact | no `libfuse` / `AppImages require FUSE` matches | `dlopen(): error loading libfuse.so.2`, `AppImages require FUSE to run.` |
| `file` | `static-pie linked` | `dynamically linked` |
| `ldd` | not a dynamic executable | dynamic |

**Launch with libfuse2 unavailable.** Ran both under `unshare -r -m` with `/dev/null` bind-mounted over `/usr/lib/libfuse.so.2` (and `AO_DATA_DIR`/`AO_RUN_FILE` pointed at /tmp):

- this branch: starts fully — Electron up, daemon logged `daemon listening addr=127.0.0.1:3001`. Incidentally stricter than intended, since the setuid `fusermount` also loses privilege in that namespace, so it ran with neither libfuse2 nor a usable fusermount.
- control: dies immediately with the dlopen error above.

Unmasked run also works normally and unmounts cleanly on exit.

**What I could not verify:** no CI job builds an AppImage on a PR — `build-artifacts.yml` is `workflow_dispatch`-only — so this change isn't exercised by the checks on this PR. It needs a Linux `npm run make` and a launch on a fuse3-only host before merge. I've done that above and am happy to re-run against any variation you want.

Unrelated local snag worth flagging for anyone reproducing this: under Node 26, `electron-forge package` silently stalls extracting the Electron zip and exits 0 with no artifact. Node 22 builds fine, and CI pins 22/24, so it only bites locally — but it looks like a success.

## Checklist


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
@neversettle17-101 @illegalcall Anyone who can review the pr and update me

illegalcall:
also e2e test it once @codebanditssss

illegalcall (approved):

illegalcall:
@Pyasma

yashksaini-coder:
Rebased onto `0b7758c7f` and re-verified locally. Attaching a short recording of that check, since this one is hard to demonstrate any other way.

**End-to-end test on my side, just now:** `npx vitest run --config vite.renderer.config.ts makers/` → 24/24 passed. `npm run typecheck` → 0 errors. Then I deleted the `toolsets` key to confirm the new test is load-bearing rather than decorative, and it fails with `AssertionError: expected undefined to be '1.0.3'`, then restored it and re-ran green. The recording walks exactly that: `AppImageTarget.js:68` showing `appimageTool == null` selecting `buildLegacyFuse2AppImage`, the single config key this PR adds, the suite green, the deliberate regression, and the restore. It ends on a clean worktree so you can see nothing was left modified.

The fuse3-only A/B that motivated the PR is unchanged in the description above; this comment is only about re-verification after the rebase.

One aside in case it saves someone time: if `npm run typecheck` reports ~21 `Cannot find module '@lexical/react/...'` errors, that is a stale `node_modules` rather than a broken branch — `@lexical/react` and `lexical` were added to `frontend/package.json` recently. `npm install` clears it, and it reproduces identically on `main`.

https://github.com/user-attachments/assets/25320176-ffa0-4c10-b07c-da367b27acf3

Pyasma (approved):

illegalcall (approved):

PR #4010, opened 2026-08-14
fix(terminal): stabilize xterm viewport, session-switch render, colors, and activity state

### LINKED ISSUE #2047: bug(activity): Claude sessions misreport state — active work shown as "ready", idle_prompt shown as "waiting_input", ready/idle flapping
## Bug

The dashboard activity state for **Claude Code** sessions is frequently wrong. Three concrete, live-reproduced defects, all rooted in the same architectural problem: `getClaudeActivityState` trusts Claude's **noisy native transcript** (read first, aged off **file mtime**, classified by entry *type*) over the **clean, hook-pushed** `.ao/activity.jsonl` event log.

**Source:** live AO session `ao-189` | **Reported by:** @harshitsinghbhandari | **Analyzed against:** `5d0b624f` | **AO:** 0.9.1 | node v22.14.0
**Confidence:** High (reproduced live on a real session; root cause traced to file:line)

This is a focused, evidence-backed sibling of the umbrella redesign #1899. It is **not** fully covered there — #1899 enumerates 15 breakages but does not list the `idle_prompt → waiting_input` mapping defect or the `assistant → ready` shadowing of a freshly-pushed `active`. Related: #1899 (umbrella), #1894 (sticky-state wallclock decay), #1934 (do we need a separate `ready` state). In-flight hook-migration PRs that should converge with the fix: #2030 (opencode), #1950 (codex).

## Reproduction (live `.ao/activity.jsonl` from session ao-189)

**Defect A — active work shown as "Ready".** While actively running tools / generating a response:
```
16:32:53.366  ready    ← Stop hook: previous turn ended
16:32:53.425  active
16:33:01.489  active    ← actively working (running a Bash tool) RIGHT NOW
```
Hook log correctly says `active`. Dashboard showed **Ready**. The pushed `active` was ignored.

**Defect B — normal idle shown as "Waiting for input" for ~15 min.**
```
16:15:46  ready
16:16:47  waiting_input   trigger: "Notification (idle_prompt)"   ← turn finished, awaiting user
16:31:34  active          ← user sent next message 15 min later
```
`idle_prompt` (Claude's "I'm done, waiting for you" nudge) was logged as `waiting_input` and, because actionable states are sticky (never decay), the dashboard showed **"Waiting for input"** for the entire 15-minute wait.

**Defect C — ready/idle flapping** (root cause shared with A): the activity clock is the log file's mtime, which Claude's noise re-snapshot writes (`permission-mode`, `ai-title`, `pr-link`) refresh even while idle → state oscillates ready⇄idle.

## Root Cause

1. **mtime as the activity clock (Defects A & C).** `readLastJsonlEntry` returns `modifiedAt: fileStat.mtime` (`packages/core/src/utils.ts:170`) and `readLastActivityEntry` likewise (`packages/core/src/activity-log.ts:118`). The native transcript's mtime is bumped by noise writes unrelated to real activity, so age-based classification is unstable. The clock should be the embedded per-entry `entry.ts` (which `getActivityFallbackState` already uses correctly — `activity-log.ts:170`).

2. **Native-first cascade + entry-type classification shadows pushed `active` (Defect A).** `getClaudeActivityState` reads the native transcript first and returns before consulting the hook log. It maps `assistant`/`summary` entries to `ready`/`idle` — never `active` (`packages/plugins/agent-claude-code/src/activity-detection.ts:429-431`). While the agent is generating output (newest native entry = `assistant`), it is classified `ready` even though the hook log has a fresh `active` from `PreToolUse`/`PostToolUse`. The pushed truth loses to the polled noise.

3. **`idle_prompt` conflated with `permission_prompt` (Defect B).** The activity-updater hook maps both notification types to `waiting_input` (`packages/plugins/agent-claude-code/src/index.ts:451-459`, and the Node/Windows variant ~573-581). Only `permission_prompt` / `PermissionRequest` is a genuine blocked-on-human state; `idle_prompt` is the turn-end "waiting for next instruction" signal and should be `ready` (decaying to `idle`), not a sticky `waiting_input`.

## Fix

Make the **hook-pushed `.ao/activity.jsonl` the source of truth** for activity; demote the native transcript to cost/summary/uuid extraction, `blocked` detection, and a fallback only when hooks are absent. Concretely:

- **A tool_use is unambiguously `working`/`active`.** A fresh pushed `active` event (`PreToolUse`/`PostToolUse`) must win — never be overridden by an `assistant→ready` read from the native transcript.
- Classify off `entry.ts`, **never** `fileStat.mtime`.
- Remap `idle_prompt → ready` (only `permission_prompt`/`PermissionRequest` → `waiting_input`).
- Add hysteresis so state can't flap within one threshold window.
- Bound sticky actionable staleness so a missed clearing event can't pin `waiting_input` forever.

This is the per-Claude instance of the unified single-clock reducer being designed (intended to be shared across claude-code/codex/opencode after #2030 and #1950 land), and aligns with #1899's inbox-vs-liveness split.

## Impact

- The dashboard's primary signal (is this agent working / done / blocked?) is untrustworthy for Claude sessions: active work reads as idle/ready, and finished-and-waiting reads as blocked.
- Knock-on effects: notifier accuracy, stuck-detection, and any auto-reaction keyed on activity state inherit the wrong state.

### LINKED ISSUE #2333: Terminal rendering glitches and blank terminal after switching sessions

### LINKED ISSUE #2529: Colors dont show up in App similar to terminal,
Image

MacOS App

Image

Terminal

### LINKED ISSUE #3803: Xterm handle replacement can dispose queued viewport work

AgentWrapper:
Fixes #3803, fixes #2333, fixes #2529, fixes #2047.

## What

Four terminal-stability fixes across the renderer and the activity lifecycle.

### #3803 — disposed-viewport crash on handle replacement
`XtermTerminal` unmount cleanup disposed the xterm instance synchronously while queued write/viewport work (WriteBuffer callbacks, `Viewport.syncScrollArea` microtasks) could still land in the same task, crashing the renderer with `Cannot read properties of undefined (reading 'dimensions')`. Cleanup now detaches all listeners/observers/timers synchronously but defers the final `term.dispose()` one macrotask, and the attachable handle drops `write`/`writeln` after disposal (still invoking `done` callbacks so attachment flow control can finish).

### #2333 — blank/glitched terminals after session switching
Two root causes:
- The retained-terminal cache became unbounded (4986af7d0), so each visited pane held a live WebGL context; past Chromium's per-page context cap the oldest contexts are force-lost and the earliest-visited panes render black. Retention is re-bounded at `MAX_RETAINED_TERMINALS = 6` with least-recently-shown eviction (evicted panes remount fresh on revisit).
- A rejected `prepareForActivation()` could leave the activation cover (`visibility:hidden`) down forever. All call sites now tolerate rejection, and a 2s activation-stall watchdog force-completes a stuck activation.

### #2529 — ANSI colors don't match a real terminal
The truecolor plumbing (tmux `-u -T RGB`, `TERM`/`COLORTERM`) already landed in #2980; the remaining gap was the palette itself. The 12 chromatic ANSI slots in `tokens.css` were a softened house mix (and the light theme darkened them further, with bright-white near-black — inverting TUI emphasis). Both themes now use the de-facto standard terminal values (VS Code integrated-terminal defaults), which agent CLIs are designed against. Neutral endpoints stay tuned to AO's terminal plate, per DESIGN.md's "terminal keeps its own palette".

### #2047 — Claude sessions misreport activity state
The `exited` activity state could pin wrongly and never self-correct (only `user-prompt-submit` could resurrect it; the reaper only transitions *into* exited). Two fixes in the lifecycle manager:
- Live-agent evidence — tool-use events and `permission-request`, which are executed by the agent process itself — now resurrects an exited state (turn-boundary callbacks like `stop`/`notification` still cannot). A wrong resurrect self-corrects on the reaper's next workload probe (~5s); a wrong exited previously pinned forever.
- The workload-dead→exited transition is skipped while a session mutation (spawn/restore/agent-switch) is in flight, matching the guard the termination path already had — probing an old process tree against new metadata mid-relaunch no longer desyncs the session to exited.

## Tests
- `XtermTerminal.test.tsx`: dispose is deferred past unmount; writes to a dead handle are dropped with `done` still invoked.
- `TerminalPane.test.tsx`: retention-cap eviction (7 visited → 6 retained, LRU evicted, fresh remount on revisit).
- `manager_test.go`: tool-use and permission-request resurrect exited; turn-boundary events do not; workload-dead observation is suppressed during session mutation.

## Validation
- `npm run typecheck` and full renderer suite: 177 files, 2233 passed / 1 skipped.
- `go test ./...`: all green (AO_* session env cleared for `internal/cli`).
- Palette change verified visually via an `ao preview` before/after harness (dark + light).

## Risks / follow-ups
- Eviction cap (6) restores the pre-4986af7d0 bound; panes beyond it lose scrollback on revisit (replay repaints latest output). If unbounded retention is wanted back, it needs a non-WebGL fallback for parked panes instead.
- Exited-resurrect intentionally excludes turn-boundary hooks; delayed `stop` events after a real exit remain ignored.

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -373,6 +373,15 @@ func (m *Manager) ApplyRuntimeObservation(ctx context.Context, id domain.Session
 			if cur.Activity.State == domain.ActivityExited {
 				return cur, false
 			}
+			if m.sessionMutationInProgress(id) {
+				// A live pane whose supervised workload is "missing" while a
+				// spawn/restore/agent-switch mutation is mid-flight is the old
+				// process tree probed against new metadata (or vice versa), not
+				// an agent exit — the same guard the termination path below
+				// already applies. Marking exited here desyncs an actively
+				// relaunching session (#2047); the next probe re-decides.
+				return cur, false
+			}
 			next := cur
 			next.Activity = domain.Activity{State: domain.ActivityExited, LastActivityAt: timeOr(f.ObservedAt, now)}
 			delete(m.flights, id)
@@ -518,10 +527,20 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID, 
 		return nil
 	}
 	// An explicit prompt submission is proof that an agent was relaunched in the
-	// preserved shell. Other same-generation callbacks may have been delayed
-	// behind the process-exit report and cannot resurrect an exited workload.
+	// preserved shell, and live-work hooks are equally strong evidence: the
+	// tool-use trio and permission-request are executed BY the agent process,
+	// so their arrival proves an agent is running right now. Without them, a
+	// stale `exited` (e.g. a workload-probe race across a daemon restart) pins
+	// an actively-working session on "exited" until the user happens to type a
+	// prompt (#2047): the reaper only transitions INTO exited, so nothing else
+	// ever corrects it. The asymmetry makes accepting them safe — a resurrect
+	// from a genuinely-delayed callback is re-marked exited by the reaper's
+	// next workload probe within one tick, while a wrongly-pinned exited never
+	// self-corrects. Turn-boundary callbacks (stop, notification, session-end)
+	// still cannot resurrect: they are exactly the deliveries that can trail a
+	// real process exit.
 	if rec.Activity.State == domain.ActivityExited && s.Valid && s.State != domain.ActivityExited &&
-		(s.State != domain.ActivityActive || s.Event != "user-prompt-submit") {
+		!isLiveAgentEvidence(s) {
 		m.mu.Unlock()
 		return nil
 	}
@@ -736,6 +755,22 @@ const maxInflightTools = 128
 // trio whose signals must not demote a sticky state on their own.
 func isToolUseEvent(event string) bool {
 	return event == "pre-tool-use" || isPostToolUseEvent(event)
+}
+
+// isLiveAgentEvidence reports whether an activity signal proves an agent
+// process is alive in the session's shell right now: an explicit prompt
+// submission, in-flight tool traffic, or a permission dialog appearing. Only
+// these signals may resurrect a session out of `exited`.
+func isLiveAgentEvidence(s ports.ActivitySignal) bool {
+	switch {
+	case s.State == domain.ActivityActive && s.Event == "user-prompt-submit":
+		return true
+	case s.State == domain.ActivityActive && isToolUseEvent(s.Event):
+		return true
+	case s.State == domain.ActivityBlocked && s.Event == "permission-request":
+		return true
+	}
+	return false
 }
 
 func isPostToolUseEvent(event string) bool {
--- a/backend/internal/lifecycle/manager_test.go
+++ b/backend/internal/lifecycle/manager_test.go
@@ -668,6 +668,21 @@ func TestRuntimeObservation_ExitedWorkloadKeepsSessionLive(t *testing.T) {
 	}
 }
 
+func TestRuntimeObservation_WorkloadDeadSkippedDuringSessionMutation(t *testing.T) {
+	m, st, _ := newManager()
+	m.SetSessionOperationGate(fixedSessionOperationGate(true))
+	rec := working("mer-1")
+	rec.Metadata.RuntimeLaunchID = "launch-1"
+	st.sessions["mer-1"] = rec
+
+	if err := m.ApplyRuntimeObservation(ctx, "mer-1", ports.RuntimeFacts{Runtime: ports.ProbeAlive, Workload: ports.ProbeDead, LaunchID: "launch-1"}); err != nil {
+		t.Fatal(err)
+	}
+	if got := st.sessions["mer-1"]; got != rec {
+		t.Fatalf("workload-dead observation marked exited during session mutation: %+v", got)
+	}
+}
+
 func TestRuntimeObservation_AliveWorkloadCannotResurrectExitedSession(t *testing.T) {
 	m, st, _ := newManager()
 	rec := working("mer-1")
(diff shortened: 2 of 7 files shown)


Rishet11:
@AgentWrapper 
All three issues this fixes (#2047, #2333, #2529) were closed completed on Aug 27. 
Only #3803 remains open and the author's newer split-outs cover it. Closing.

PR #4013, opened 2026-08-14
fix(windows): hide daemon child process consoles

### LINKED ISSUE #4012: Windows desktop flashes command prompt windows during startup and background probes
## Bug
On Windows, AO can visibly flash transient Command Prompt windows while starting the desktop app and while the daemon performs background work. A user report included a screenshot of a black command window popping up over AO, with the note that it appears and closes repeatedly around 5-6 times while using AO.

**Source:** Local Windows user report and screenshot shared during triage on 2026-08-15. Startup logs were also provided from `frontend/npm run dev` showing daemon startup and background probes.
**Confidence:** High

## Reproduction
1. On Windows, run `cd frontend && npm run dev`.
2. Wait for Electron Forge to launch the desktop app and daemon.
3. Observe transient Command Prompt windows appearing and disappearing during startup.
4. Continue using AO in flows that trigger background CLI probes; command windows can also flash during normal use.

## Root Cause
Windows GUI apps that spawn console subsystem executables without hidden process creation flags can show visible console windows. AO had two affected paths:

- Developer mode launched the daemon with `go run ./cmd/ao daemon`, which can spawn Go compiler/linker helper processes and make startup especially noisy on Windows.
- Several daemon-owned non-interactive probes used raw process spawning instead of AO's hidden-process wrapper, so downloaded Windows app users could also see command windows from Git, Docker, Tailscale, provider version probes, reviewer probes, or ACP provider processes.

A related Windows dev issue was observed after switching to a fixed prebuilt `frontend/daemon/ao.exe`: active sessions can keep that exe locked, making the next `npm run dev` fail with `EPERM: operation not permitted, unlink ...\frontend\daemon\ao.exe`.

## Fix
Use a Windows-specific dev daemon launch path and hide daemon-owned helper processes:

- Build Windows dev daemons into versioned `frontend/daemon/dev-*` directories and pass the selected exe through a dev manifest.
- Keep macOS/Linux dev startup unchanged.
- Keep packaged daemon resolution unchanged.
- Add hidden-process flags for Node build/probe spawns.
- Route daemon-owned non-interactive CLI probes through the existing backend hidden-process wrapper.
- Preserve Windows ACP process-tree cleanup while also suppressing console windows.

## Impact
- Windows developer mode should stop flashing command prompt windows during daemon startup and avoid locked fixed `ao.exe` rebuild failures.
- Downloaded Windows app users should not see background helper/probe command windows flash during normal app use.
- macOS and Linux behavior should remain unchanged.

JAYATIAHUJA:
Fixes https://github.com/Untrivial-ai/agent-orchestrator/issues/4012

## What happened
On Windows, AO can show black Command Prompt windows for a moment and then close them again. A user reported that this happens repeatedly while using AO, sometimes 5-6 times, and it looks broken.

This is mainly a Windows problem. When a desktop app starts console programs in the background, Windows can show a command window unless the process is started with hidden-window flags.

## Why it happened
There were two places where this could happen:

1. In dev mode, `npm run dev` started the daemon through `go run`. On Windows, `go run` starts extra compiler/helper processes, so startup could flash several command windows.
2. In the normal downloaded Windows app, the daemon also starts background tools like Git, Docker, Tailscale, agent version checks, reviewer tools, and ACP provider processes. Some of these were not using AO's hidden process wrapper.

There was also a related dev-mode problem: if an old AO session was still using `frontend/daemon/ao.exe`, Windows could keep that file locked. The next `npm run dev` could then fail while trying to delete or replace it.

## What changed
- Windows dev mode now builds the daemon first and starts the built `ao.exe` instead of using `go run` at app startup.
- Windows dev builds use versioned daemon folders, so an old running session does not block the next build.
- Background helper processes launched by the daemon now use the existing hidden-process wrapper on Windows.
- ACP provider processes still keep their process-group cleanup behavior, but no longer need to show a console window.
- Mac and Linux dev behavior is unchanged.
- Packaged daemon location is unchanged.

## Before 

https://github.com/user-attachments/assets/2b6f5464-251b-4336-b757-69ca42bce39c

## After 
it works fine :)

## Tested
- `npm --prefix frontend run test -- src/shared/daemon-launch.test.ts`
- `go test ./internal/workspacewatch -run '^$'`
- focused backend compile check for touched packages
- `git diff --check upstream/main..HEAD`
- `npm --prefix frontend run build:daemon`

## Note
Full frontend typecheck currently fails in this local checkout because `packages/product-ui/src/SessionsBoardView.tsx` cannot resolve `motion/react`. That is unrelated to this PR; the changed frontend test passes.

### DIFF
--- a/backend/internal/adapters/agent/modelcatalog/catalog.go
+++ b/backend/internal/adapters/agent/modelcatalog/catalog.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+	aoprocess "github.com/aoagents/agent-orchestrator/backend/internal/process"
 )
 
 const (
@@ -262,7 +263,7 @@ func hasDiscoverySource(agentID string) bool {
 }
 
 func modelCommand(ctx context.Context, binary string, args []string, workingDir string, env map[string]string) *exec.Cmd {
-	cmd := exec.CommandContext(ctx, binary, args...) //nolint:gosec // binary is adapter-resolved, args are static
+	cmd := aoprocess.CommandContext(ctx, binary, args...) //nolint:gosec // binary is adapter-resolved, args are static
 	cmd.WaitDelay = commandTerminationWait
 	if strings.TrimSpace(workingDir) != "" {
 		cmd.Dir = workingDir
--- a/backend/internal/adapters/agent/muse/muse.go
+++ b/backend/internal/adapters/agent/muse/muse.go
@@ -14,7 +14,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -24,6 +23,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/agentbase"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/binaryutil"
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+	aoprocess "github.com/aoagents/agent-orchestrator/backend/internal/process"
 )
 
 const adapterID = "muse"
@@ -227,7 +227,7 @@ func resolveMuseBinary(ctx context.Context, spec binaryutil.BinarySpec) (string,
 }
 
 func isOfficialMuseBinary(ctx context.Context, binary string) bool {
-	cmd := exec.CommandContext(ctx, binary, "--version")
+	cmd := aoprocess.CommandContext(ctx, binary, "--version")
 	cmd.Env = append(os.Environ(), "MUSE_NO_AUTO_UPDATE=1")
 	out, err := cmd.CombinedOutput()
 	if err != nil {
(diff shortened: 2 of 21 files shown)


Vaibhaav-Tiwari (changes requested):
one serious bug: the script that builds the Windows app binary now saves the daemon
  to a differently-named folder every time (dev-12345-6789/ao.exe) instead of the fixed spot the app expects
  (daemon/ao.exe). Problem is, this same script also runs when building the real, official Windows installer —
  not just for local dev. Nobody told it "only do this weird versioned-folder thing during dev."

  Result: if this ships as-is, every Windows user who installs the app will open it and the daemon just won't
  start — because the app looks for daemon/ao.exe and it's not there anymore. The build itself won't error or
  warn you — it'll succeed and quietly produce a broken installer.

  Everything else in the PR (the actual "hide the console window" changes) is clean and fine.

JAYATIAHUJA:
Fixed in 7da12738f. The versioned daemon directory is now used only for Windows dev builds (predev --dev). Packaging/make/publish builds clean the daemon directory and write the binary to the fixed daemon/ao.exe path, so the installed Windows app keeps the location it expects. The focused daemon launch test passes (11/11).

JAYATIAHUJA:
@Vaibhaav-Tiwari I pushed one more fix in 8191b9e13 for the Windows file-lock case during packaging. Locked old dev-* daemon folders are now skipped, while the build still produces the fixed daemon/ao.exe. I verified the build with a locked dev binary. Could you please take another look and re-review?

Vaibhaav-Tiwari (approved):
lgtm

Issue #3873, opened 2026-08-12
High session spawn failure rate: mostly missing agent binary / runtime and git branch state, not the spawn engine


Pulkit7070:
## Summary

Session spawn has a high failure rate that is concentrated in a few fixable causes, not in the spawn engine itself. From telemetry (PostHog project 475752, last 14 days):

- App-initiated spawns (`ao.renderer.orchestrator_spawn_*`): ~50% of requests end in `failed` (e.g. Aug 11: 700 requested, 358 succeeded, 344 failed).
- All spawns (backend `ao.session.spawned` vs `ao.session.spawn_failed`): ~22% failure (agent-initiated spawns dominate volume and mostly succeed, which pulls the blended rate down).
- Once an agent is actually installed, per-harness failure is low: claude-code 392 failed / 8,832 spawned (~4.5%), codex ~4.8%. So the harnesses themselves are not the problem.

## Failure reasons (backend `ao.session.spawn_failed`, error_kind / error_code, 14d, ~3,030 failures)

| reason | count | approx share |
| --- | --- | --- |
| RUNTIME_PREREQUISITE_MISSING | 878 | 29% |
| AGENT_BINARY_NOT_FOUND | 531 | 18% |
| internal (unclassified) | 952 | 31% |
| BRANCH_NOT_FETCHED | 313 | 10% |
| BRANCH_CHECKED_OUT_ELSEWHERE | 124 | 4% |
| INVALID_BRANCH | 79 | 3% |
| SESSION_MODE_UNSUPPORTED | 67 | 2% |
| AGENT_REQUIRED | 56 | 2% |
| others (CHAT_AUTH_REQUIRED, WORKSPACE_PROVISION_FAILED, UNKNOWN_HARNESS, ...) | ~30 | 1% |

Session kind of failures: orchestrator 2,271 vs worker 759. Most failures happen before a harness is attached (2,108 of the failures have no `harness`), consistent with failing at validation.

## Root cause

Three buckets account for essentially all of it:

1. Agent / runtime not installed (~47%): `AGENT_BINARY_NOT_FOUND` + `RUNTIME_PREREQUISITE_MISSING`. Users click spawn with an agent whose CLI is not on the machine, or a required runtime is missing. This is a setup/environment gap surfaced as a spawn failure, not a spawn bug.
2. Internal errors (~31%): unclassified. Not covered by the error codes; needs Sentry stack traces to break down.
[...]

PR #4021, opened 2026-08-15
fix(cli): derive spawn name when omitted

### LINKED ISSUE #3911: bug(orchestrator): worker spawns fail when LLM omits --name — no auto-fallback for required sidebar label
# Orchestrator worker spawns fail when the orchestrator LLM omits `--name` — no auto-fallback or recovery loop

## Bug

When the orchestrator agent runs `ao spawn` to delegate work to a worker, the `--name` flag is **required** at the CLI layer. If the orchestrator's LLM omits `--name` (or supplies a label >20 chars), the spawn command exits with a usage error (exit code 2) and **no worker session is created**. The orchestrator may not surface or recover from this failure, leaving the delegation task unfulfilled.

**Source:** Discord thread [#bug-triage](https://discord.com/channels/1476302178913357958/1491735678156013588/1533922851172647052)
**Reported by:** @aditi4384 (aditi)
**Analyzed against:** `df2ed9fea` (origin/main, 2026-08-12)
**Confidence:** High — `--name` enforcement confirmed at three layers on current main.

## Reproduction

1. Start an orchestrator session for any project.
2. Give the orchestrator a task that requires spawning a worker.
3. If the orchestrator LLM emits `ao spawn --project <proj> --prompt "..."` **without** `--name`, the CLI rejects it:
   ```
   Error: --name is required
   ```
   Exit code 2. No worker is created.
4. If the orchestrator LLM passes `--name` with a label >20 characters:
   ```
   Error: --name must be 20 characters or fewer
   ```
   Same outcome — no worker created.

## Root Cause

`--name` is **intentionally required** and enforced at three layers — this is by design, not a regression:

**1. CLI enforcement** (`backend/internal/cli/spawn.go:78-84`):
```go
name := strings.TrimSpace(opts.name)
if name == "" {
    return usageError{fmt.Errorf("--name is required")}
}
if utf8.RuneCountInString(name) > maxDisplayNameLen {
    return usageError{fmt.Errorf("--name must be %d characters or fewer", maxDisplayNameLen)}
}
```

**2. Daemon API enforcement** (`backend/internal/service/session/service.go:627-629`):
```go
displayName = strings.TrimSpace(displayName)
if displayName == "" {
    return apierr.Invalid("DISPLAY_NAME_REQUIRED", "Display name is required", nil)
}
```

**3. Orchestrator system prompt** (`backend/internal/session_manager/prompt.go:190-193`):
```
ao spawn --project <project> --name "<label>" --prompt "<clear worker task>"
--name is required: a deliberate sidebar label so the user can see what each worker is working on at a glance; labels must be 20 characters or fewer.
Before running ao spawn, count the --name label yourself. It must be 20 characters or fewer.
```

The gap: the prompt **instructs** the LLM to always include `--name`, but there is **no programmatic fallback**. If the LLM fails to comply (which LLMs do — especially smaller models), the spawn fails and the orchestrator has to self-diagnose from the error output. Whether the orchestrator reliably catches and retries is model-dependent and unverified.

For comparison, the **delegation API path** (`backend/internal/service/session/delegation.go:160-168`) already auto-generates a display name from the task brief:
```go
func delegatedTaskDisplayName(brief string) string {
    title := strings.Join(strings.Fields(brief), " ")
    if title == "" {
        return delegatedTaskUntitledName
    }
    if utf8.RuneCountInString(title) <= delegatedTaskTitleLimit {
        return title
    }
    return strings.TrimSpace(string([]rune(title)[:delegatedTaskTitleLimit]))
}
```
This auto-naming logic exists but only covers the programmatic delegation path, not the `ao spawn` CLI command the orchestrator uses.

## Suggested Fix

Two complementary approaches:

1. **Auto-fallback in the CLI spawn path:** If `--name` is omitted but `--prompt` or `--issue` is supplied, auto-generate a display name from the prompt/issue title using the same `delegatedTaskDisplayName` logic already in `delegation.go`. This eliminates the failure mode entirely without weakening the sidebar-label UX for users who do pass `--name`.

2. **Better error surfacing to the orchestrator:** Ensure the orchestrator's terminal output clearly shows the `--name is required` error so the LLM can self-correct on the next turn. (This likely already works via tmux output capture, but should be verified with telemetry.)

## Impact

- Orchestrator-to-worker delegation silently fails when the LLM omits or mis-sizes `--name`.
- Contributes to the high spawn failure rate tracked in [#3873](https://github.com/Untrivial-ai/agent-orchestrator/issues/3873).
- Smaller / less instruction-following models are disproportionately affected.
- No workaround except hoping the orchestrator LLM reads its system prompt.

## Related

- [#3873](https://github.com/Untrivial-ai/agent-orchestrator/issues/3873) — High session spawn failure rate (broader umbrella)
- [#3855](https://github.com/Untrivial-ai/agent-orchestrator/issues/3855) — Cannot start new workers — spawn always fails (different root cause: proxy/connection)
- [#3875](https://github.com/Untrivial-ai/agent-orchestrator/issues/3875) — Fresh orchestrator spawn crashes claude with "invalid session id" (different root cause)

AgentWrapper:
## Summary
- derive an omitted `ao spawn --name` from the normalized prompt
- fall back to `Issue <id>` or `Untitled task` when no prompt is available
- preserve explicit-name validation and Unicode-safe 20-character truncation
- update orchestrator guidance so an omitted label no longer blocks delegation

Fixes #3911
Refs #3873

## Tests
- `cd backend && env -u AO_SESSION_ID -u AO_PROJECT_ID go test ./internal/cli ./internal/session_manager`

## Risk
Low. Explicit names retain their existing validation; only the previously rejected omitted-name path changes.

### DIFF
--- a/backend/internal/cli/spawn.go
+++ b/backend/internal/cli/spawn.go
@@ -20,6 +20,8 @@ import (
 // daemon's spawn handler so a direct API call is held to the same limit.
 const maxDisplayNameLen = 20
 
+const fallbackSpawnDisplayName = "Untitled task"
+
 type spawnOptions struct {
 	project         string
 	harness         string
@@ -77,10 +79,7 @@ func newSpawnCommand(ctx *commandContext) *cobra.Command {
 			if opts.noTakeover && opts.claimPR == "" {
 				return usageError{fmt.Errorf("--no-takeover requires --claim-pr")}
 			}
-			name := strings.TrimSpace(opts.name)
-			if name == "" {
-				return usageError{fmt.Errorf("--name is required")}
-			}
+			name := spawnDisplayName(opts.name, opts.prompt, opts.issue)
 			if utf8.RuneCountInString(name) > maxDisplayNameLen {
 				return usageError{fmt.Errorf("--name must be %d characters or fewer", maxDisplayNameLen)}
 			}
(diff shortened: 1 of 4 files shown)


neversettle17-101 (commented):
## What This PR Does

Approving. This changes ao spawn so an omitted --name is derived from the prompt, issue id, or an Untitled task fallback, while keeping explicit-name length validation. The orchestrator prompt guidance and CLI tests were updated to match.

## Summary of Changes

- Added spawnDisplayName fallback and Unicode-safe truncation for derived names.
- Updated --name help text and removed the missing-name usage failure.
- Adjusted spawn tests and orchestrator guidance around optional labels.

## Findings

No blocking issues found in the reviewed diff.

## Regression Risk

Low. Explicit labels still use the existing length validation; the changed path only affects previously rejected omitted labels.

## Testing

I did not run tests. The PR updates CLI tests for fallback generation and the existing spawn request path.

Issue #3855, opened 2026-08-11
Cannot start new workers — spawn always fails


Rishet11:
# GitHub Issue — AO cannot spawn workers 

**Title:** Cannot start new workers: spawn always fails (Orchestrator: Muse Spark .. Workers: Cline)

**What happens:**
I try to start a new worker with `ao spawn`. The command never creates a worker. It shows an error about connecting to 127.0.0.1:3001 and says operation not permitted. If I try through the API, I get 500 PROJECTS_LIST_FAILED.

**Steps to reproduce:**
1. Have a project like `aohack` at /Users/rishetmehra/Desktop/aohack
2. Run `ao spawn --project aohack --name test-worker-1 --prompt hello`
3. Or run `ao status` or `curl -x $http_proxy http://127.0.0.1:3001/api/v1/projects`
4. See error every time

**What I expected:**
A new worker session should start, and `ao session ls` should show it.

**What actually happens:**
- CLI always says: `dial tcp 127.0.0.1:3001: connect: operation not permitted`
- API always says: `500 PROJECTS_LIST_FAILED` or `500 INTERNAL_ERROR`
- Health check via proxy works: `curl -x $http_proxy http://127.0.0.1:3001/healthz` returns ok, so daemon is running but cannot be reached normally.

**Why this happens (plain English):**
Two problems combine.

1. The AO command line tool tries to talk to the daemon on your own computer (127.0.0.1). On this system, direct talk to that address is blocked. The tool is supposed to use a proxy, but Go's helper `ProxyFromEnvironment` skips the proxy for 127.0.0.1 by design. So it never uses the proxy and always gets blocked.

2. Even when we reach the daemon through the proxy, it cannot load projects. It shows PROJECTS_LIST_FAILED for every request. This comes from the project service which hides the real database error. The database file itself is okay, but the daemon needs to write to ~/.ao/data (WAL file) and that write was blocked in sandboxed runs. So the real cause is hidden.

**Environment:**
- AO daemon pid 71197, port 3001, status ok via proxy but unhealthy via direct
- `ao doctor` shows: FAIL data-dir-write and FAIL daemon: unhealthy
- DB `~/.ao/data/ao.db` has 4 projects and 18 sessions, integrity ok

**Suggested fix for maintainers:**
- Make CLI respect proxy for daemon URL, or add a way to force proxy for 127.0.0.1.
[...]

PR #4025, opened 2026-08-15
fix(projects): retain list failure cause in logs

AgentWrapper:
## Summary
- retain `ListProjects` storage failures in the service error chain
- keep the existing redacted `PROJECTS_LIST_FAILED` API envelope
- let the HTTP access logger record the actionable SQLite/filesystem cause under the response request ID

Refs #3855
Refs #3873

## Triage
The reported `dial tcp 127.0.0.1 ... operation not permitted` is an external sandbox policy, not an AO spawn failure: the same report says `ao doctor` fails its direct daemon probe and data-directory write check. AO intentionally keeps its unauthenticated primary control API on direct loopback. Routing it through an ambient HTTP proxy would expose that trusted local API to the proxy and is not a safe recovery.

The `PROJECTS_LIST_FAILED` diagnostics loss is an AO bug: the project service replaced the SQLite cause with a bare API error before the request logger could capture it. This PR fixes that without exposing internal paths or database details to clients.

## Tests
- `cd backend && go test ./internal/service/project ./internal/httpd`
- `cd backend && go vet ./internal/service/project ./internal/httpd`

## Risk
Low. The public status, code, and message are unchanged; only the server-side error chain gains the storage cause.

### DIFF
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -3,6 +3,7 @@ package project
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


neversettle17-101 (commented):
## What This PR Does

Approving. This preserves the underlying storage cause when project listing fails while keeping the public PROJECTS_LIST_FAILED API envelope redacted.

## Summary of Changes

- Wrapped ListProjects storage errors behind the existing apierr.Internal value.
- Added a test asserting both errors.Is for the storage cause and errors.As for the API error still work.

## Findings

No blocking issues found in the reviewed diff.

## Regression Risk

Low. The public status/code/message are unchanged; only the internal error chain gains diagnostic context for logs.

## Testing

I did not run tests. The PR adds focused service coverage for preserving the storage cause while retaining the redacted API error.

PR #4028, opened 2026-08-15
feat(cli): enrich session list status

axisrow:
Code changes: +463 -71  
Tests: +672 -4  
Others: +55 -24  

## Summary
- enrich `ao session ls` with a compact branch/PR/CI/review/thread/activity/age table
- fetch daemon-owned PR summaries concurrently and retain list facts when a detail lookup fails
- expose exact unresolved human review-thread counts through the PR summary API

## Verification
- `cd backend && go test ./internal/cli -run "^TestSessionList" -count=1`
- `cd backend && go test ./internal/service/session ./internal/httpd/... -run "Test(ListPRSummariesExposesReviewSummariesButKeepsRawLogsAndCommentBodiesPrivate|SummarizeReviewSurfacesApprovedAndChangesRequestedSummaries|NewSessionPRSummaryMapsProviderReviewEntries)" -count=1`
- `cd backend && go test ./internal/httpd/...`
- `npm run api:spec` and `npx openapi-typescript@7.4.4 backend/internal/httpd/apispec/openapi.yaml -o frontend/src/api/schema.ts`

## Known limitation
- Local frontend typecheck was not runnable because the checkout lacks the `tsc` binary; CI will verify it.

## Notes
The Go rewrite reserves `ao status` for daemon health; the session-oriented status view lives at `ao session ls`.

Relates to #25

## Intermediate-build upgrade repair

- Added migration `0131` to invalidate unreliable review-count certainty once on databases created by intermediate `0130` builds with `DEFAULT FALSE`. Review rows and observation timestamps are preserved; databases using the corrected `DEFAULT TRUE` are left unchanged.
- Regression coverage exercises intermediate `0130`, legacy `0123`, normal main upgrades, healthy complete observations, and burned `0130` history. A successful full refresh restores an exact count, which remains valid after reopening.

## Follow-up verification

- Passed: focused migration and migration-ledger tests, both normally and with `GOMAXPROCS=2 go test -race -p=1 ./internal/storage/sqlite -run 'TestMigratePRReviewPartialUpgrade|TestMigrationVersionLedger' -count=1`.
- Passed: `go build ./...`, `go vet ./...`, pinned golangci-lint on `./internal/storage/sqlite` (0 issues), and `npm run sqlc` with no generated changes.
- The full local `go test -race -timeout=15m ./...` run was stopped after failures in unchanged fake-agent, OMP, Pi hook, and tmux timing/integration tests on a heavily overloaded host. The focused fake-agent timing test also failed on retry. Full-suite validation remains with CI on the pushed commit.


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
Hi, can you go into the details on why this would be required?

axisrow:
Hi, this implements the requested scope from #25 / INT-1348. The original request was for a compact session table showing branch, PR, CI, review decision, unresolved threads, activity, and age, with PR data fetched in parallel. In the Go rewrite ao status is reserved for daemon health, so I placed the session-oriented view under ao session ls instead.

I had some spare tokens available, so I implemented the requested scope from #25; otherwise they would have gone unused.

axisrow:
@illegalcall friendly bump — following up on my answer above, let me know if that covers it or if you'd like anything else clarified.

axisrow:
Rebased onto current main (was conflicting). @illegalcall following up again on my answer above — let me know if it addresses your question or if there's anything else to clarify.

axisrow:
@ronishrohan could you weigh in? @illegalcall asked about the motivation on Aug 16, I answered on Aug 17 (scope from #25 / INT-1348) and followed up twice since, but haven't heard back. Happy to clarify further or adjust the scope if that unblocks it.

backend/internal/cli/session.go
```diff
@@ -64,6 +68,24 @@ type sessionListResponse struct {
...
+}
+
+type sessionPRSummaryDTO struct {
+	Number int `json:"number"`
+	CI     struct {
+		State string `json:"state"`
+	} `json:"ci"`
+	Review struct {
+		Decision              string `json:"decision"`
+		UnresolvedThreadCount int    `json:"unresolvedThreadCount"`
```
illegalcall: [P2] Preserve unknown counts instead of decoding them as zero

A new CLI talking to a pre-change daemon can receive a successful PR-summary response that omits `unresolvedThreadCount`. This plain `int` decodes that absence as `0`; `listPRsForSession` then takes its address at lines 758-759, so both the table and JSON report a known zero instead of unknown. The HTTP-error fallback does not cover this successful-but-missing-field case. The new server also uses `int` with `omitempty`, so a genuinely observed zero is absent on the wire too. Please preserve unknown versus observed zero end to end (for example, optional count representations that serialize a known zero and keep absence unknown), or defer the count. Add fixtures for an older response without the field, an explicit known zero, a positive count, and a failed lookup; missing data should remain `null`/`-`, not `0`.

backend/internal/service/session/pr_summary.go
```diff
@@ -154,8 +154,8 @@ func summarizeCI(pr domain.PullRequest, checks []domain.PullRequestCheck) PRCISu
 	return out
 }
 
-func summarizeReview(pr domain.PullRequest, comments []domain.PullRequestComment, reviews []domain.PullRequestReview) PRReviewSummary {
-	out := PRReviewSummary{Decision: reviewOrNone(pr.Review)}
+func summarizeReview(pr domain.PullRequest, comments []domain.PullRequestComment, reviews []domain.PullRequestReview, threads []domain.PullRequestReviewThread) PRReviewSummary {
+	out := PRReviewSummary{Decision: reviewOrNone(pr.Review), UnresolvedThreadCount: unresolvedHumanThreadCount(threads)}
```
illegalcall: [P2] Do not publish a partial observation as an exact thread total

The stored thread slice is not necessarily a complete snapshot. `github/observer_provider.go:287-322` can return only the latest 50 threads with `Partial: true`: for example, a newly observed PR has 51 threads, the oldest thread is unresolved, and all 50 in the fetched window are resolved. Because the window boundary is resolved, the older page is not fetched. This new count is then zero (omitted by the DTO), and the CLI displays `0` despite outstanding feedback. Conversely, `ReviewWriteMerge` preserves unseen old rows, so a previously stored unresolved thread that was resolved outside the fetched window can still inflate the count. Please carry observation completeness into the count contract and only expose a known total when justified, or defer the numeric field. Add tests covering an unresolved thread outside the fetched window and a stale unresolved row preserved by a partial refresh. Making the JSON property optional alone does not address this.

backend/internal/service/session/pr_summary.go
```diff
@@ -273,6 +273,16 @@ func latestReviewSummaries(reviews []domain.PullRequestReview) map[string]domain
 	return latestByReviewer
 }
 
+func unresolvedHumanThreadCount(threads []domain.PullRequestReviewThread) int {
+	count := 0
+	for _, thread := range threads {
+		if !thread.Resolved && !thread.IsBot {
+			count++
```
illegalcall, line 297: [P2] Deduplicate thread identity across PR URL aliases before counting

`ListPRSummaries` groups PR aliases and appends `ListPRReviewThreads` results from every alias into this slice (lines 62-84). If the same provider thread is persisted under both an old and a new repository URL that belong to one group, this loop increments twice: one unresolved human thread becomes a count of two. Grouping the PRs does not deduplicate their thread rows. Please count distinct provider thread identities within the canonical PR and define which observation wins if alias records disagree, or defer the count. Add a fixture with two grouped aliases containing the same unresolved `ThreadID` and assert that the summary reports one, plus a conflicting resolved/unresolved alias case.

illegalcall (changes requested):
Requesting changes on the correctness of the new unresolved-thread count. The CLI enrichment is useful, and I am not asking to restore the old human-readable columns: existing daemon API fields and CLI JSON fields remain available, and I found no breaking removal for the existing desktop/mobile consumers.

The count currently has three concrete failure paths, detailed inline:

1. A successful older-daemon response without the optional field becomes a known-looking `0` in the new CLI.
2. Partial review observations are presented as an exact total, despite potentially missing or stale thread rows.
3. Duplicate thread records under grouped PR URL aliases are counted more than once.

Please either fix the count contract/aggregation and add focused regression tests, or defer the numeric count while keeping the branch/PR/CI/review enrichment. Updating consumers alone cannot make an incomplete or duplicated source count exact.

Reviewed the code at `5426e4f9`; these are static code-path findings, not a claim of a live reproduction or a local test run.

axisrow:
Thanks for the thorough review — all three findings are addressed in 03d11d0a1, taking the "fix the count contract/aggregation" path rather than deferring the field.

1. **Unknown vs observed zero, end to end.** `UnresolvedThreadCount` is now a pointer through the contract, DTO, and CLI DTO. An observed zero serializes explicitly (`*int` + `omitempty` only drops nil), and the CLI keeps the summary field as a pointer so a pre-change daemon's successful-but-fieldless response stays unknown: the table renders `-` and JSON stays `null`, never a known-looking `0`. Fixtures cover an older response without the field, an explicit known zero, a positive count, and the HTTP-error fallback.

2. **Partial observation no longer published as exact.** The review observation's partial flag is now persisted (`pr.review_partial`, migration 0123): the observer records `obs.Review.Partial` on fetched passes and preserves the local flag otherwise, and the summary omits the count whenever any PR in the group has `review_partial` set **or** a zero `review_observed_at` (never-fetched reviews — the "no rows at all" case). So both window-missed unresolved threads and stale unresolved rows preserved by `ReviewWriteMerge` now surface as unknown rather than a wrong total. Tests cover the partial case and the never-observed case.

3. **Alias dedup.** `ListPRSummaries` now deduplicates thread rows by provider `ThreadID` within the canonical PR group (rows without an ID keep their own identity). When alias records disagree, the most recently updated row wins, and an unresolved row wins an exact tie, so a stale alias cannot hide active feedback. Fixtures assert two grouped aliases with the same unresolved `ThreadID` count as one, and the conflicting resolved/unresolved alias resolves to the fresher record.

One note on the wire shape: `unresolvedThreadCount` in the OpenAPI spec is now `["null", "integer"]` — the nullable form is the contract for "unknown", matching the CLI's `-` rendering.

i-trytoohard:
@axisrow Marking as stale: this has been open 25 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

illegalcall:
Re-reviewed `03d11d0a1`. Thanks for the follow-up—the **missing-versus-zero propagation** and **alias-thread deduplication** are now fixed. The completeness change is only partially fixed, so I would keep changes requested until the following are addressed.

### Remaining correctness issues

1. **[P2] A failed first review fetch can still publish a known-looking `0`.** If a known PR has no prior review observation, metadata/CI changes successfully, and `FetchReviewThreads` fails, the observer preserves review rows but still fills a previously zero `ReviewObservedAt`. `ReviewPartial` remains false. The new summary gate therefore treats empty, never-fetched review storage as complete. Only successful review observations should establish completeness; metadata success must not manufacture it. See [observer timestamp/flag handling](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/observe/scm/observer.go#L1752-L1762) and [summary gate](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/service/session/pr_summary.go#L84-L90).

2. **[P2] The PR-claim writer does not populate the new completeness field.** `enrichClaimReviews` knows whether the fetch was partial and selects Merge/Preserve appropriately, but `claimRowsFromSCM` never assigns `ReviewPartial` and always assigns `ReviewObservedAt`. The full upsert then persists false, potentially overwriting an existing true value. A partial claim can publish an inaccurate exact count; a failed first claim-review fetch can publish zero. Update this producer too, including preservation of existing certainty on failure. See [claim enrichment](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/service/session/claim_pr.go#L166-L192), [claim row construction](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/service/session/claim_pr.go#L223-L258), and [new upsert assignment](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/storage/sqlite/queries/pr.sql#L57-L61).

3. **[P2] Upgrades incorrectly mark historical observations complete.** `DEFAULT FALSE` gives every existing PR a nonpartial flag, including old partial snapshots with nonzero review timestamps. Those immediately qualify for an exact count. A subsequent poll is not a reliable repair: the observer skips writes when semantic hashes are unchanged, and the review hash already included `Partial` before this commit. Initialize historical certainty conservatively and ensure certainty changes can persist even when provider content is unchanged. See [migration](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/storage/sqlite/migrations/0123_pr_review_partial.sql#L6), [unchanged-observation skip](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/observe/scm/observer.go#L579-L583), and [review hash](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/observe/scm/observer.go#L2015-L2022).

### Merge / migration integration

4. **[P1] Migration `0123` is already allocated on current main.** Main at `8343bcc0` uses [`0123_agent_install_jobs.sql`](https://github.com/Untrivial-ai/agent-orchestrator/blob/8343bcc089982e202f0fbc37ff43aa09de3110d4/backend/internal/storage/sqlite/migrations/0123_agent_install_jobs.sql) and continues through 0128. Rebase, allocate a fresh number, and preserve the shipped migration ledger. Also cover relevant upgrade/development profiles where 123 may already have been recorded; generated PR reads now require the new column. The repository documents [physical-schema repair for skipped generated-query dependencies](https://github.com/Untrivial-ai/agent-orchestrator/blob/03d11d0a1589f264c8da58f64bb318df70b5d688/backend/internal/storage/sqlite/db.go#L1219-L1235).

Please add tests through the actual **observer/claim → storage → summary** paths for failed first fetches, partial/failed claims, legacy partial-record upgrades, and unchanged post-upgrade observations. The new summary fixtures correctly test the gate given accurate input flags, but do not exercise these producers.

Existing API fields remain unchanged; migrating all desktop/Kanban consumers is not required to fix these issues. This was a read-only code review, not a test run. GitHub currently reports merge conflicts and no checks for this head; please rerun the relevant checks after rebasing and fixing the above.

axisrow:
Addressed in 8ef9381 (on top of the rebase that moved the migration to 0130 and regenerated sqlc):

1. **Failed first review fetch manufacturing completeness** — `domainFromObservation` now advances `ReviewObservedAt` only when `opts.reviewFetched`; the `|| reviewObservedAt.IsZero()` fallback is gone, so a metadata-success pass can no longer turn never-fetched storage into a complete-looking observation.

2. **Claim writer** — `claimRowsFromSCM` carries the pair: a partial fetch stores `(observedAt, partial=true)`, a failed fetch passes the zero pair. The upserts treat `(review_observed_at, review_partial)` as one unit (`COALESCE` + `CASE` in both `UpsertPR` and `UpsertLegacyPR`): a NULL timestamp keeps the stored pair, so a preserve-mode claim or observer pass can no longer overwrite an existing `true` with `false`.

3. **Historical rows + repair on unchanged content** — migration 0130 now defaults `partial=TRUE` (unknown certainty stays hidden), and `prepareForPersistence` treats a partial↔full flip as a review change, so the certainty repair persists even when the content hash matches and the poll would otherwise skip the write. A `repairRenumbered*` entry remaps dev databases that recorded the column under its pre-renumber 0123 (re-initializing them to uncertain), and a `schemaRepairs` entry covers burned-0130 profiles.

Tests through the real paths: failed first fetch (observer → persistence), partial/full/failed claims (service → claimer), the preserve contract at the SQL layer (store round-trip through the actual upserts), and the unchanged-content completeness repair.

axisrow:
@illegalcall friendly bump — the requested changes are in 8ef9381 (item-by-item mapping in my previous comment).

illegalcall (approved):

Issue #4033, opened 2026-08-15
bug(desktop): Chat UI loses the conversation for shells and has no agent-switch affordance — remaining terminal-construct parity gaps


illegalcall:
## Bug

Chat mode should be a different *interface* to the same session, not a reduced one: every construct that works in Terminal UI should work the same in Chat UI. PR [#4027](https://github.com/Untrivial-ai/agent-orchestrator/pull/4027) fixes this for the reviewer terminal (it becomes a tab inside the Chat surface instead of swapping in `CenterPane`). Two constructs still behave worse than TUI, plus a refactor to stop the two tab strips drifting.

**Source:** reported by @anmol0b in agent session | **Analyzed against:** `73600d218` | **Confidence:** High (exact code paths traced)

## Gap 1 — Shell terminals still do the full surface swap

**Reproduction**
1. Open a chat-mode session; start a turn so the timeline has content.
2. Click **+** next to the chat header to open a shell.

**Actual:** `showChatSurface` goes false, the entire `SessionChatSurface` unmounts, and `CenterPane` swaps in. The timeline, composer draft, and scroll position are destroyed. Returning to chat is a navigation, not a tab click.
**Expected:** the shell renders as a tab in the chat header — `[Chat] [Shell 1]` — exactly like the reviewer does after #4027, and like `[Worker] [Shell 1]` works in TUI.

**Root cause:** `frontend/src/renderer/components/SessionView.tsx:620` — `showChatSurface = session?.mode === "chat" && routedTerminalTarget.kind === "worker"`. Only `worker` (and, after #4027, `reviewer`) keeps the chat surface mounted; `shell` targets fall through to the full swap.

**Fix:** extend the #4027 pattern to `kind === "shell"` targets: keep the chat surface mounted, render the shell's `TerminalPane` as the tab body, and list this session's shells as tabs in the chat header (TUI supports multiple shells; the chat header currently has only one reviewer slot).

## Gap 2 — Agent switching has no affordance in chat mode

**Reproduction**
1. Open a chat-mode session running Codex.
2. Try to switch the session's agent to Claude Code.

**Actual:** no switch-agent control exists anywhere in the chat surface — not in the chat header, `TurnSettingsBar`, or composer. The only path is: switch interface to TUI → switch agent → switch back to chat.
**Expected:** the same in-place agent switch TUI offers from the worker tab (`TerminalSwitchAgentButton` → `SwitchAgentDialog`), with chat-appropriate placement.

[...]

PR #4027, opened 2026-08-15
fix(desktop): render reviewer terminal inside Chat surface instead of swapping in CenterPane

illegalcall:
## Summary

When a chat-mode session has an active reviewer pane, clicking the Reviewer tab in the chat header now **keeps the chat surface mounted** and renders the reviewer's TerminalPane as the tab body. The chat tab stays one click away in the same header with proper active-styling.

Previously (after #3962) the reviewer target caused a full surface swap from `SessionChatSurface` to `CenterPane`, destroying the chat timeline, composer state, and scroll position every time you wanted to check on the review. Now the reviewer is a tab inside the chat surface — exactly like it's a tab inside the terminal UI's tab strip.

## Changes (4 files, +139/−8)

| File | What |
|---|---|
| `SessionView.tsx` | `showChatSurface` now stays true for reviewer targets in chat mode. Passes tab state, daemon readiness, and theme to the chat surface. |
| `SessionChatSurface.tsx` | Forwards the new props to `ChatWorkspace`. |
| `ChatWorkspace.tsx` | **Core change:** conditionally renders `TerminalPane` as the tab body when `reviewerActive`. ChatHeader tabs gain `aria-selected`, active underline styling, and the chat tab becomes clickable to return from the reviewer view. Terminal font size reads the user's persisted `ao.terminal.fontSize` pref. |
| `SessionView.test.tsx` | Updated mock + test: chat surface stays mounted, reviewer renders inside it, back-to-chat round-trip works. |

## Test results

- `frontend:typecheck` — ✅ 0 errors
- SessionView tests — ✅ 64/64
- Chat component tests — ✅ 276/276

Fixes #4008

PR #4036, opened 2026-08-15
fix(desktop): keep shells and agent switching inside the chat surface

illegalcall:
Part of #4033 — closes its shell and agent-switch gaps.

## Summary

Two constructs that work in Terminal UI now work the same in Chat UI, without leaving the conversation:

**Shells are tabs in the chat header (Gap 1).** Previously `showChatSurface` only stayed mounted for `worker`/`reviewer` targets, so opening a shell unmounted the entire chat surface — timeline, composer draft, scroll position — and swapped in `CenterPane`. Now a shell target renders its `TerminalPane` as the tab body inside the chat surface (the routed `TerminalTarget` passed straight through, mirroring `reviewerTarget` from #4027), and the conversation panel hides with `hidden`/`inert` instead of unmounting — drafts, attachments, edits, and scroll survive.

Shell tabs use the shared `ShellTerminalTab` (already used by `CenterPane` and the terminals screen "so the three never drift"), which brings select/close/**rename** along for free. The desktop next/previous tab shortcuts (Ctrl+Tab / Ctrl+Shift+Tab) now cycle chat → reviewer → shells in strip order.

**Agent switching has an entry point in chat (Gap 2).** `CenterPane` rendered `TerminalSwitchAgentButton`; the chat surface had nothing, even though `useSwitchAgent` was already chat-aware (it invalidates conversation queries). The same control now renders in the chat header's action region. The mutation state lives in `SessionChatSurface` — which owns query wiring — so `ChatWorkspace` stays a provider-free pure view (its documented contract; enforced by existing fixture tests).

## Changes (6 files, +377/−28)

| File | What |
|---|---|
| `SessionView.tsx` | `showChatSurface` stays mounted for `shell` targets; passes the routed shell target + shell list/select/close/rename into the chat surface. |
| `SessionChatSurface.tsx` | Forwards shell props; owns the agent-switch dialog state + mutation; hands the button and a body-container ref down. |
| `ChatWorkspace.tsx` | Shell pane parallel to the reviewer pane (same `TerminalPane` target passthrough, zoom/fullscreen wiring); shell tabs via shared `ShellTerminalTab`; `timelineActive` generalization so the chat tab deactivates for either terminal pane; next-tab shortcut cycles through shells; renders the passed-down switch control. |
| tests | ChatWorkspace: shell pane hides-not-unmounts the conversation, target passthrough, Ctrl+Tab cycling. SessionView: shells-from-chat stay inside the chat surface. SessionChatSurface: switch control offered in chat. |

## What's deliberately not here

- **Full tab-strip unification (Gap 3 remainder):** the chat/reviewer tabs remain hand-rolled in `ChatHeader`; extracting one shared strip with `CenterPane` remains a follow-up on #4033.
- CenterPane's deep in-flight switch UX (input locking, source focus, announcements) — the chat surface already reflects controller transitions through its own banners; the button/dialog/mutation flow is identical.

## Test

- `npm run typecheck` — ✅ 0 errors
- `vitest run --exclude 'src/landing/**'` — ✅ 173 files, 2352 passed, 0 failures
- `src/landing` failures are pre-existing on the base branch and unrelated.

Rebased onto `main` after #4027 merged; see [the screenshot comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4036#issuecomment-5301564255) for live-app captures.

### DIFF
--- a/frontend/src/renderer/components/SessionView.tsx
+++ b/frontend/src/renderer/components/SessionView.tsx
@@ -620,9 +620,13 @@ export function SessionView({ sessionId }: SessionViewProps) {
 	const routedTerminalTarget = terminalTargetBelongsToSession(terminalTarget, sessionId)
 		? terminalTarget
 		: ({ kind: "worker" } satisfies TerminalTarget);
-	// Chat surface stays mounted in chat mode for worker + reviewer targets;
-	// it renders the terminal pane as a tab when the reviewer is the active target.
-	const showChatSurface = session?.mode === "chat" && (routedTerminalTarget.kind === "worker" || routedTerminalTarget.kind === "reviewer");
+	// Chat surface stays mounted in chat mode for worker, reviewer, and shell
+	// targets. A terminal pane (reviewer or shell) renders as a tab inside the
+	// chat surface, so opening one never costs the user the conversation.
+	const chatTargetKind = routedTerminalTarget.kind;
+	const showChatSurface =
+		session?.mode === "chat" &&
+		(chatTargetKind === "worker" || chatTargetKind === "reviewer" || chatTargetKind === "shell");
 
 	// The pane shows one terminal at a time, so selecting a shell or the reviewer
 	// takes the agent's terminal off screen while the route still points here.
@@ -831,6 +835,13 @@ export function SessionView({ sessionId }: SessionViewProps) {
 										routedTerminalTarget.kind === "reviewer" ? routedTerminalTarget : undefined
 									}
 									onSelectChat={selectSessionTerminal}
+									shellTerminals={shellTerminals}
+									shellTarget={
+										routedTerminalTarget.kind === "shell" ? routedTerminalTarget : undefined
+									}
+									onSelectShellTerminal={selectShellTerminal}
+									onCloseShellTerminal={closeShellTerminalByHandle}
+									onRenameShellTerminal={renameShellTerminalByHandle}
 									daemonReady={daemonStatus.state === "ready"}
 									theme={theme}
 									headerActions={sessionHeaderActions}
--- a/frontend/src/renderer/components/chat/ChatWorkspace.tsx
+++ b/frontend/src/renderer/components/chat/ChatWorkspace.tsx
@@ -484,6 +484,24 @@ export function ChatWorkspace({
 		shellTarget,
 		shellTerminals,
 	]);
+	const handleChatTabsKeyDown = useCallback(
+		(event: ReactKeyboardEvent<HTMLDivElement>) => {
+			if (
+				event.target instanceof HTMLElement &&
+				event.target.getAttribute("role") === "tab" &&
+				event.key === "Tab" &&
+				event.ctrlKey &&
+				!event.altKey &&
+				!event.metaKey
+			) {
+				event.preventDefault();
+				selectAdjacentTab(event.shiftKey ? -1 : 1);
+				return;
+			}
+			handleTerminalTabListKeyDown(event);
+		},
+		[selectAdjacentTab],
+	);
 
 	useEffect(
 		() =>
@@ -538,6 +556,7 @@ export function ChatWorkspace({
 				onSelectShellTerminal={onSelectShellTerminal}
 				onCloseShellTerminal={onCloseShellTerminal}
 				onRenameShellTerminal={onRenameShellTerminal}
+				onTabsKeyDown={handleChatTabsKeyDown}
 				switchAgentControl={switchAgentControl}
 				headerActions={headerActions}
 				inline={isFullscreen}
@@ -865,6 +884,7 @@ function ChatHeader({
 	onSelectShellTerminal,
 	onCloseShellTerminal,
 	onRenameShellTerminal,
+	onTabsKeyDown,
 	switchAgentControl,
 	headerActions,
 	inline,
@@ -885,6 +905,7 @@ function ChatHeader({
 	onSelectShellTerminal?: (handleId: string) => void;
 	onCloseShellTerminal?: (handleId: string) => void;
 	onRenameShellTerminal?: (handleId: string, title: string) => void;
+	onTabsKeyDown?: (event: ReactKeyboardEvent<HTMLDivElement>) => void;
 	/** The in-place agent-switch control, same entry point as the terminal pane. */
 	switchAgentControl?: ReactNode;
 	headerActions?: ReactNode;
@@ -914,7 +935,7 @@ function ChatHeader({
 					<div
 						aria-label="Chat tabs"
 						className="scrollbar-none flex h-full min-w-flex-min flex-1 items-center overflow-x-auto"
-						onKeyDown={handleTerminalTabListKeyDown}
+						onKeyDown={onTabsKeyDown ?? handleTerminalTabListKeyDown}
 						role="tablist"
 					>
 						<span
--- a/frontend/src/renderer/components/chat/ChatWorkspace.tsx
+++ b/frontend/src/renderer/components/chat/ChatWorkspace.tsx
@@ -46,6 +46,7 @@ import {
 } from "../../lib/design-tokens";
 import { isLinuxPlatform, isMacPlatform } from "../../lib/platform";
 import { handleTerminalTabListKeyDown } from "../../lib/terminal-tabs";
+import type { ShellTerminal } from "../../hooks/useShellTerminals";
 import { useUiStore } from "../../stores/ui-store";
 import type { TerminalTarget } from "../../types/terminal";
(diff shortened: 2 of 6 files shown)


illegalcall:
## Screenshots — live app, chat-mode session (codex worker)

Captured from the desktop app running this branch (isolated dev data, real daemon + codex chat controller), demonstrating both behaviors end to end.

### 1 · Baseline: chat timeline
The chat surface before opening anything — conversation with the running agent.

chat timeline

### 2 · Shell opens as a tab — conversation stays mounted
Clicked **+ / New terminal** from the chat header. Previously this unmounted the entire chat surface and swapped in `CenterPane` (timeline, draft, scroll position destroyed). Now the shell renders as a tab in the chat header with its terminal as the tab body — note the chat tab ("chat shell demo") stays one click away, exactly like `[Worker] [Shell]` in the terminal UI.

shell tab inside chat surface

### 3 · Back to the conversation — one click
Clicking the chat tab returns to the timeline. The surface never remounts; the shell tab remains in the strip for round-tripping.

back to chat timeline

### 4 · Agent switching — now available in chat
The **Switch agent** control in the chat header's action region (the same `TerminalSwitchAgentButton` the terminal pane renders), opening the standard in-place switch dialog — no more "switch to TUI → switch agent → switch back to chat".

agent switch dialog in chat surface

Verified at each step via the DOM: chat surface stays mounted (`data-session-mode=chat`), shell pane renders inside it (`chat-shell-terminal`), tab `aria-selected` state tracks the active pane, and the switch dialog opens with Claude Code / Codex options.

vanshx999:
Local validation on 0ba9f5e:

- `npm run typecheck` ✅
- Focused Vitest (`ChatWorkspace`, `SessionChatSurface`, `SessionView`) ✅ 111 tests

Manual smoke checklist to run in the desktop app:
1. Open a running worker in Chat mode and confirm the conversation is visible.
2. Use `+ / New terminal`; verify the shell opens as a tab in the chat header, the chat tab remains present, and returning to Chat preserves draft/scroll/timeline state.
3. With Chat, reviewer (if available), and one or more shells open, verify Ctrl+Tab / Ctrl+Shift+Tab cycles tabs in strip order.
4. Click Switch agent in the chat header; open the dialog, choose another agent, and confirm the chat remains mounted and the new controller becomes active.
5. Close/rename/select the shell tab and confirm it behaves like the terminal screen tabs.

I could not complete the live desktop smoke test from this worktree because the app requires a running daemon/session.

vanshx999:
Follow-up: after rebuilding Electron (the initial `npm ci --ignore-scripts` left Electron unavailable), the full non-landing frontend suite is green: 173 files, 2297 passed, 1 skipped. The previously failing `src/main/browser-view-host.test.ts` now passes 55/55.

vanshx999:
Follow-up on the switch-from-chat path.

What I found: switching agents from a chat session that recorded no runtime handle (via the new chat-header Switch-agent control) returns SESSION_INCOMPLETE_HANDLE — "Session is missing runtime or workspace handles."

Why: the switch path (agent_switching.go:246) requires both Metadata.WorkspacePath and Metadata.RuntimeHandleID, but chat-mode restore/resume (manager.go:1629-1631) is allowed to leave RuntimeHandleID empty for chat sessions. So a chat session that never recorded a runtime handle runs fine in chat, yet can't be agent-switched hitting this error right on the flow this PR newly exposes from the chat surface.

Sanity check against your screenshot #4: that case worked on a codex worker, which has an attached runtime so this is conditional, not a blanket break. My repro was a handle-less chat session.

- Are chat-mode spawns expected to always populate RuntimeHandleID? If yes, my repro may indicate a spawn path that skips it — happy to grab the daemon log / session row.
- If no (chat can legitimately be handle-less), should the switch either be granted for chat-only sessions, or surface an actionable blocker reason instead of the raw SESSION_INCOMPLETE_HANDLE?

somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -620,9 +620,13 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
-	// Chat surface stays mounted in chat mode for worker + reviewer targets;
-	// it renders the terminal pane as a tab when the reviewer is the active target.
-	const showChatSurface = session?.mode === "chat" && (routedTerminalTarget.kind === "worker" || routedTerminalTarget.kind === "reviewer");
+	// Chat surface stays mounted in chat mode for worker, reviewer, and shell
+	// targets. A terminal pane (reviewer or shell) renders as a tab inside the
+	// chat surface, so opening one never costs the user the conversation.
+	const chatTargetKind = routedTerminalTarget.kind;
+	const showChatSurface =
+		session?.mode === "chat" &&
+		(chatTargetKind === "worker" || chatTargetKind === "reviewer" || chatTargetKind === "shell");
```
whoisasx, line 629: [P1] Routing every shell target through `SessionChatSurface` makes shells inaccessible whenever the conversation is loading, unavailable, or errored, because that component returns before it renders `ChatWorkspace` and therefore never consumes `shellTarget`. The unavailable state even tells the user to open a shell from the inspector, but doing so leaves this same unavailable screen visible. Please render the shell independently of conversation availability or fall back to `CenterPane` for those states.

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -138,6 +160,21 @@ export function SessionChatSurface({
 			onOpenReviewerTerminal={onOpenReviewerTerminal}
 			reviewerTarget={reviewerTarget}
 			onSelectChat={onSelectChat}
+			shellTerminals={shellTerminals}
+			shellTarget={shellTarget}
+			onSelectShellTerminal={onSelectShellTerminal}
+			onCloseShellTerminal={onCloseShellTerminal}
+			onRenameShellTerminal={onRenameShellTerminal}
+			switchAgentControl={
```
whoisasx, line 175: [P1] This exposes agent switching for active chat sessions without checking `session.terminalHandleId`. Handle-less chat sessions are a deliberately supported backend state, but `agent_switching.go` rejects them with `SESSION_INCOMPLETE_HANDLE`, so the newly advertised flow opens normally and then fails on submit. Please either support switching these sessions in the backend or suppress/disable this entry point with an actionable explanation until a runtime handle exists.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -495,13 +564,35 @@ export function ChatWorkspace({
 					</div>
 				</div>
 			) : null}
+			{shellTarget && session ? (
```
whoisasx, line 598: [P2] Once a shell is rendered here, `CenterPane` is unmounted, so its close-terminal shortcut listener and enabled-state effect are also gone. `ChatWorkspace` does not replace either one; consequently Cmd/Ctrl+W is consumed by the application shortcut channel but no handler closes the active shell. Please mirror `CenterPane`'s `onCloseShellTerminalShortcut` and `setCloseShellTerminalShortcutEnabled` wiring while a shell tab is active.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -425,13 +448,45 @@ export function ChatWorkspace({
...
+		onSelectShellTerminal,
+		reviewerActive,
+		reviewerTerminal,
+		shellActive,
+		shellTarget,
+		shellTerminals,
+	]);
 
 	useEffect(() => {
 		const disposePrevious = aoBridge.app.onPreviousTabShortcut(selectAdjacentTab);
```
whoisasx: [P2] Both shortcut channels invoke the same forward-only callback. That was equivalent while chat had only two tabs, but after adding reviewer plus shell tabs, Ctrl+Shift+Tab now moves forward instead of backward. Please make `selectAdjacentTab` direction-aware and pass `-1` here and `1` to the next-tab listener, matching `CenterPane` and `ShellTerminalsView`; add a three-tab previous-shortcut test as coverage.

whoisasx (dismissed):
The shell-in-chat direction and preservation of the mounted conversation are sound, and the focused frontend verification is green. However, two supported chat states are currently broken: a shell cannot be opened when the conversation controller is unavailable, and the new agent-switch entry point is offered to handle-less chat sessions that the backend explicitly rejects. The shell close shortcut and reverse tab navigation also regress when ChatWorkspace takes ownership of three or more tabs. Please address these paths before merging.

nikhilachale (approved):
Tested this PR locally, LGTM.

PR #4040, opened 2026-08-15
fix(frontend): stabilize switch agent selector

### LINKED ISSUE #4039: Switch-agent selector fills terminal and reanimates on repeated activation
## Bug

The terminal-scoped **Switch agent** selector has two related interaction regressions:

1. The provider/model card expands to the terminal panel width instead of remaining a compact centered card like the in-progress switch presentation.
2. Clicking the toolbar switch button while the selector is already open closes and immediately reopens the dialog, replaying its fade animation.

**Reported by:** @whoisasx (local desktop report)  
**Analyzed against:** `0fbe25e168b27ee35814286f541e426622f4cab0`  
**Confidence:** High  
**Priority:** P2 — the workflow remains usable, but the selector presentation and repeated interaction are visibly degraded.

## Reproduction

1. Open an active Claude Code or Codex worker terminal.
2. Click the **Switch agent** toolbar button.
3. Observe that the selector card spans almost the complete terminal width.
4. Click the same toolbar button again while the selector remains open.
5. Observe the selector fade out and back in.

## Root cause

- `frontend/src/renderer/components/ui/dialog.tsx` supplies `w-full` from the shared dialog primitive.
- `frontend/src/renderer/components/SwitchAgentDialog.tsx` also supplied the custom `w-dialog-md` utility. `tailwind-merge` does not classify the custom utility as conflicting with `w-full`, so both classes survive. Tailwind emits `.w-full` later, making the computed width 100% rather than the intended 420px.
- The selector uses a non-modal Radix dialog. A pointer-down on the toolbar button is treated as an outside interaction and emits `onOpenChange(false)`; the button's unconditional click handler then emits `onOpenChange(true)`. A focused component test records the exact `false → true` sequence.

These paths were introduced together by [PR #3770](https://github.com/Untrivial-ai/agent-orchestrator/pull/3770).

## Fix

- Use an unambiguous arbitrary width utility that replaces the shared `w-full` through `tailwind-merge`.
- Ignore toolbar pointer/click activation while the selector is already open, before Radix can treat it as an outside dismissal.
- Cover the computed width in the renderer Playwright harness and the repeated activation state sequence in the component suite.

## Impact

- The selector no longer visually matches the compact centered switch-progress UI.
- Repeated activation causes a distracting close/reopen animation and unnecessary remount.
- Existing tests only checked for a class name and the first open action, so neither browser-computed geometry nor repeated activation was protected.

whoisasx:
## What

- keep the terminal-scoped Switch agent selector at the 420px dialog token on wide panes and within a 16px gutter on narrow panes
- make repeated activation of the already-open toolbar control a no-op instead of closing and reopening the Radix dialog
- cover the browser-computed wide/narrow geometry and the repeated activation state sequence

## Why

Fixes #4039.

The shared dialog primitive contributes `w-full`, while the custom `w-dialog-md` class was not recognized as a conflicting Tailwind utility. Both survived merging and the later full-width rule won. Separately, the non-modal dialog treated a pointer-down on the toolbar control as an outside dismissal before the control's click handler reopened it.

## How

- use an arbitrary width utility so `tailwind-merge` replaces the shared `w-full`; cap it against the terminal containing block rather than the viewport
- intercept pointer activation only while the selector is already open, with an `onClick` guard covering keyboard and assistive-technology activation
- replace the former class-name assertion with real computed-width Playwright coverage

## Testing

- `npm test` — 185 files passed; 2,397 tests passed, 1 skipped
- `npm run typecheck`
- `frontend/e2e/switch-agent-dialog.spec.ts` against an isolated `dev:web` renderer — 2 passed (wide and narrow terminal cases)
- targeted red/green verification for repeated activation — 23 component tests passed

## Checklist

### DIFF
--- a/frontend/e2e/switch-agent-dialog.spec.ts
+++ b/frontend/e2e/switch-agent-dialog.spec.ts
@@ -3,9 +3,8 @@ import { installFakeAgent } from "./support/fake-bridge";
 
 const projectId = "switch-agent-dialog";
 
-test.use({ reducedMotion: "reduce" });
-
 async function openSwitchAgentDialog(page: Page) {
+	await page.emulateMedia({ reducedMotion: "reduce" });
 	await installFakeAgent(page, {
 		projectId,
 		projectName: projectId,
@@ -53,7 +52,7 @@ async function openSwitchAgentDialog(page: Page) {
 	};
 }
 
-test("renderer: switch-agent selector remains compact inside a wide terminal", async ({ page }) => {
+test("renderer: switch-agent selector remains compact inside a wide terminal @T0", async ({ page }) => {
 	const { dialog, terminalPanel } = await openSwitchAgentDialog(page);
 	await expect(dialog).toHaveCSS("width", "420px");
 	await expect
@@ -61,7 +60,7 @@ test("renderer: switch-agent selector remains compact inside a wide terminal", a
 		.toBeGreaterThan(420);
 });
 
-test("renderer: switch-agent selector stays inside a narrow terminal", async ({ page }) => {
+test("renderer: switch-agent selector stays inside a narrow terminal @T0", async ({ page }) => {
 	await page.setViewportSize({ width: 960, height: 720 });
 	const { dialog, terminalPanel } = await openSwitchAgentDialog(page);
 	const dialogBox = await dialog.boundingBox();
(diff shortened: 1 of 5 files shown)


whoisasx:
# Before
Screenshot 2026-08-15 at 14 14 52

 ---
 
# After
Screenshot 2026-08-15 at 14 08 55

whoisasx:
# Before
https://github.com/user-attachments/assets/cafac6a6-efdf-42d8-8902-bb502e0b7f44

---

# After 

https://github.com/user-attachments/assets/0500c040-4425-4a2a-9f9d-fbc3fc24e543

Annieeeee11 (approved):
lgtm

PR #4049, opened 2026-08-15
fix(cline): reconcile completed turns to idle

### LINKED ISSUE #4048: bug(cline): completed turns stay working/active at the idle prompt
## Bug

A live Cline worker remains **Working** with `activity.state=active` after its turn has completed and the Cline TUI has returned to the empty `Ask anything…` composer. The agent is quiescent and waiting for a new user prompt, but AO reports it as actively working indefinitely.

**Source:** AO maintainer report (internal task, 2026-08-15) | **Reported by:** AO maintainer | **Analyzed against:** `736782b4afc76aeefb4c2707959456380f64cb0d`
**AO binary:** freshly built Go rewrite from the analyzed checkout (`version: dev`) | **Daemon:** healthy on `127.0.0.1:3001`
**Cline:** `3.0.55` | **OS:** macOS 14 / Darwin 23.6.0 arm64 | **Shell:** zsh
**Confidence:** High — reproduced against the live rewrite daemon, captured the TUI, inspected durable state/change history, and traced the adapter/observer capability boundary.
**Priority:** P2 — the dashboard's core working-vs-idle signal is wrong for Cline sessions, but impact is bounded to that harness.

## Reproduction

1. Spawn a TUI worker with `harness=cline` and a prompt (live reproduction: session `agent-orchestrator-133`, display name `quick hello`, prompt `how's it going?`).
2. Let Cline finish its response and return to its empty composer.
3. Verify the terminal is idle. The live tmux pane ended with:

   ```text
   ❯ Ask anything...
   Claude Opus 5 (medium) ...  ○ Plan ● Act (Tab)
   agent-orchestrator-133 (ao/agent-orchestrator-133/root)
   ⏵⏵ Auto-approve all enabled (Shift+Tab)
   ```

4. Query the same session using the freshly built rewrite CLI:

   ```json
   {
     "id": "agent-orchestrator-133",
     "harness": "cline",
     "activity": {
       "state": "active",
       "lastActivityAt": "2026-08-15T23:09:13.940388Z"
     },
     "status": "working"
   }
   ```

5. Observe that it remains `working/active` even though the TUI is awaiting a new prompt.

The SQLite `change_log` corroborates the missing transition: the session was created idle, then recorded a single `active` transition at `23:09:13.940388Z`; there is no later idle transition.

## Root Cause

Cline has hook installation but **no activity classifier file and no terminal detector** under `backend/internal/adapters/agent/cline/`.

- `backend/internal/adapters/agent/cline/hooks.go:49-60` installs `TaskStart -> session-start`, `UserPromptSubmit -> user-prompt-submit`, `PreToolUse -> permission-request`, and `TaskCancel -> stop`.
- `backend/internal/adapters/agent/activitydispatch/dispatch.go:50` routes Cline through `activitystate.StandardDeriveActivityState`, so task/prompt start sets `active` and only `stop` can set `idle`.
- Cline's normal successful turn completion does not fire the installed `TaskCancel` hook. The live worktree confirms these are the only four AO-managed hook scripts; there is no normal-completion hook that can clear `active`.
- `backend/internal/adapters/agent/cline/cline.go:40-41` only asserts the base adapter/agent interfaces. It does not implement `ports.TerminalActivityDetector` or `ports.ContinuousTerminalActivityDetector`.
- The safety-net observer therefore exits at `backend/internal/observe/activity/observer.go:111-113` and never reconciles the visible idle composer back to `idle`.

The result is deterministic: a successfully completed Cline turn stays active until a later cancellation/exit or another external state change.

## Fix

Add a Cline terminal activity classifier following the established Codex/Muse pattern:

- recognize Cline's current empty composer (`❯` + `Ask anything...`) together with its status/footer markers as authoritative `idle` evidence;
- strip ANSI control sequences and restrict matching to recent terminal lines so transcript prose cannot trigger a false idle;
- opt Cline into continuous terminal reconciliation so a completed turn clears on the next observer tick instead of waiting for the stale-active threshold;
- leave generation/tool states hook-driven unless the TUI provides an equally authoritative current-state marker;
- add captured-frame tests for the idle composer and false-positive/active cases.

## Impact

- Sidebar, board, and inspector show quiescent Cline sessions as busy.
- Humans and orchestrators cannot reliably tell whether a Cline worker is still progressing or ready for more work.
- Any automation keyed to `active` vs `idle` inherits the false working state.

## Related

- [#3115](https://github.com/Untrivial-ai/agent-orchestrator/issues/3115) / [PR #3117](https://github.com/Untrivial-ai/agent-orchestrator/pull/3117) — same stuck-active symptom for Codex; added terminal idle reconciliation.
- [PR #3683](https://github.com/Untrivial-ai/agent-orchestrator/pull/3683) — Muse had no classifier and gained continuous TUI classification; closest implementation pattern.
- [#2047](https://github.com/Untrivial-ai/agent-orchestrator/issues/2047) / [PR #4010](https://github.com/Untrivial-ai/agent-orchestrator/pull/4010) — Claude activity misreport family and lifecycle hardening.
- [#3387](https://github.com/Untrivial-ai/agent-orchestrator/issues/3387) — Kimi stuck working after an error because it lacks a terminal detector.
- [#3306](https://github.com/Untrivial-ai/agent-orchestrator/issues/3306) — inverse-direction Cursor sticky activity misreport caused by incomplete adapter signals.

AgentWrapper:
Fixes #4048.

## Summary

- install Cline's native `TaskComplete` hook so successful turns clear `active` immediately; error termination remains covered by Cline's existing `TaskCancel` dispatch
- add a continuous terminal activity classifier that recognizes Cline's empty `Ask anything...` / `Plan something...` composer as idle and its `Thinking... (esc to cancel)` indicator as active
- use recent, ANSI-normalized terminal markers with newest-marker precedence to avoid transcript false positives
- cover live idle-composer output, active generation, retained-marker ordering, completion-hook coverage, and transcript rejection

## Tests

- `cd backend && go test ./internal/adapters/agent/cline ./internal/observe/activity ./internal/adapters/agent/activitydispatch`
- `cd backend && env -i HOME=\"$HOME\" PATH=\"$PATH\" TMPDIR=\"${TMPDIR:-/tmp}\" go test ./...`

## Risks / follow-ups

- Terminal detection necessarily follows Cline's rendered labels; the native completion/cancel hooks are the primary path, while the classifier is a fallback for missing callbacks. Tests pin the current Cline 3.0.55 markers observed during reproduction.
- Like the existing Codex/Muse continuous detectors, detector-based message-delivery readiness remains bounded by the caller timeout if no authoritative TUI marker matches.

### DIFF
--- a/backend/internal/adapters/agent/cline/activity.go
+++ b/backend/internal/adapters/agent/cline/activity.go
@@ -1,0 +1,73 @@
+package cline
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+var clineTerminalEscape = regexp.MustCompile(`\x1b(?:\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]|\][^\x07]*(?:\x07|\x1b\\))`)
+
+// ContinuouslyDetectTerminalActivity opts Cline into terminal reconciliation
+// on every observer tick. Completion hooks provide the primary transition, but
+// the idle composer is the fallback when Cline omits or loses a callback.
+func (p *Plugin) ContinuouslyDetectTerminalActivity() bool { return true }
+
+// DetectTerminalActivity recognizes authoritative current-state markers in
+// Cline's TUI. The newest marker wins so retained transcript text cannot
+// override the current composer or generation indicator.
+func (p *Plugin) DetectTerminalActivity(output string) (domain.ActivityState, bool) {
+	lines := clineTerminalLines(output)
+	if len(lines) == 0 {
+		return "", false
+	}
+	start := len(lines) - 30
+	if start < 0 {
+		start = 0
+	}
+	recent := lines[start:]
+
+	for i := len(recent) - 1; i >= 0; i-- {
+		line := strings.ToLower(recent[i])
+		if strings.Contains(line, "thinking... (esc to cancel)") {
+			return domain.ActivityActive, true
+		}
+		if clineEmptyComposer(recent[i]) && clineModeFooterAfter(recent, i) {
+			return domain.ActivityIdle, true
+		}
+	}
+	return "", false
+}
+
+func clineEmptyComposer(line string) bool {
+	line = strings.TrimSpace(line)
+	if !strings.HasPrefix(line, "❯") {
+		return false
+	}
+	prompt := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(line, "❯")))
+	return prompt == "ask anything..." || prompt == "plan something..."
+}
+
+func clineModeFooterAfter(lines []string, composer int) bool {
+	for _, line := range lines[composer+1:] {
+		line = strings.ToLower(line)
+		if strings.Contains(line, "plan") && strings.Contains(line, "act") && strings.Contains(line, "(tab)") {
+			return true
+		}
+	}
+	return false
+}
+
+func clineTerminalLines(output string) []string {
+	plain := clineTerminalEscape.ReplaceAllString(strings.ReplaceAll(output, "\r", "\n"), "")
+	raw := strings.Split(plain, "\n")
+	lines := raw[:0]
+	for _, line := range raw {
+		line = strings.TrimSpace(line)
+		if line != "" {
+			lines = append(lines, line)
+		}
+	}
+	return lines
+}
--- a/backend/internal/adapters/agent/cline/cline.go
+++ b/backend/internal/adapters/agent/cline/cline.go
@@ -39,6 +39,7 @@ func New() *Plugin {
 
 var _ adapters.Adapter = (*Plugin)(nil)
 var _ ports.Agent = (*Plugin)(nil)
+var _ ports.ContinuousTerminalActivityDetector = (*Plugin)(nil)
 
 // Manifest returns the adapter's static self-description.
 func (p *Plugin) Manifest() adapters.Manifest {
--- a/backend/internal/adapters/agent/cline/hooks.go
+++ b/backend/internal/adapters/agent/cline/hooks.go
@@ -53,11 +53,15 @@ type clineHookSpec struct {
 //   - UserPromptSubmit -> user-prompt-submit  (user message submitted: active)
 //   - PreToolUse       -> permission-request  (about to act: approval point)
 //   - TaskCancel       -> stop                (task cancelled/aborted: idle)
+//   - TaskComplete     -> stop                (task completed normally: idle)
+//   - TaskError        -> stop                (task ended with an error: idle)
 var clineManagedHooks = []clineHookSpec{
 	{Event: "TaskStart", Subcommand: "session-start"},
 	{Event: "UserPromptSubmit", Subcommand: "user-prompt-submit"},
 	{Event: "PreToolUse", Subcommand: "permission-request"},
 	{Event: "TaskCancel", Subcommand: "stop"},
+	{Event: "TaskComplete", Subcommand: "stop"},
+	{Event: "TaskError", Subcommand: "stop"},
 }
 
 // GetAgentHooks installs AO's Cline hook scripts into the worktree-local
--- a/backend/internal/adapters/agent/cline/hooks.go
+++ b/backend/internal/adapters/agent/cline/hooks.go
@@ -54,14 +54,12 @@ type clineHookSpec struct {
 //   - PreToolUse       -> permission-request  (about to act: approval point)
 //   - TaskCancel       -> stop                (task cancelled/aborted: idle)
 //   - TaskComplete     -> stop                (task completed normally: idle)
-//   - TaskError        -> stop                (task ended with an error: idle)
 var clineManagedHooks = []clineHookSpec{
 	{Event: "TaskStart", Subcommand: "session-start"},
 	{Event: "UserPromptSubmit", Subcommand: "user-prompt-submit"},
 	{Event: "PreToolUse", Subcommand: "permission-request"},
 	{Event: "TaskCancel", Subcommand: "stop"},
 	{Event: "TaskComplete", Subcommand: "stop"},
-	{Event: "TaskError", Subcommand: "stop"},
 }
 
 // GetAgentHooks installs AO's Cline hook scripts into the worktree-local
--- a/backend/internal/adapters/agent/cline/activity_test.go
+++ b/backend/internal/adapters/agent/cline/activity_test.go
@@ -68,7 +68,6 @@ func TestDetectTerminalActivityRejectsTranscriptText(t *testing.T) {
 func TestClineManagedHooksClearCompletedTurns(t *testing.T) {
 	want := map[string]string{
 		"TaskComplete": "stop",
-		"TaskError":    "stop",
 	}
 	for _, spec := range clineManagedHooks {
 		if subcommand, ok := want[spec.Event]; ok {
--- a/backend/internal/adapters/agent/cline/activity_test.go
+++ b/backend/internal/adapters/agent/cline/activity_test.go
@@ -1,0 +1,84 @@
+package cline
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+func readClineFixture(t *testing.T, name string) string {
+	t.Helper()
+	output, err := os.ReadFile(filepath.Join("testdata", name))
+	if err != nil {
+		t.Fatal(err)
+	}
+	return string(output)
+}
+
+func TestDetectTerminalActivityClineFrames(t *testing.T) {
+	tests := []struct {
+		name    string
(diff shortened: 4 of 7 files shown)


backend/internal/adapters/agent/cline/hooks.go
```diff
@@ -53,11 +53,15 @@ type clineHookSpec struct {
...
 //   - TaskCancel       -> stop                (task cancelled/aborted: idle)
+//   - TaskComplete     -> stop                (task completed normally: idle)
+//   - TaskError        -> stop                (task ended with an error: idle)
 var clineManagedHooks = []clineHookSpec{
 	{Event: "TaskStart", Subcommand: "session-start"},
 	{Event: "UserPromptSubmit", Subcommand: "user-prompt-submit"},
 	{Event: "PreToolUse", Subcommand: "permission-request"},
 	{Event: "TaskCancel", Subcommand: "stop"},
+	{Event: "TaskComplete", Subcommand: "stop"},
+	{Event: "TaskError", Subcommand: "stop"},
```
AgentWrapper: `TaskError` does not exist as a Cline hook event. The hooks registry in cline/cline (`apps/vscode/src/core/hooks/hook-factory.ts`) defines only `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `TaskStart`, `TaskResume`, `TaskCancel`, `TaskComplete`, and the CLI routes error-terminated runs to the `TaskCancel` hook (`apps/cli/src/utils/hooks.ts`). This script will never fire — error turns are already covered by the `TaskCancel` entry above. Please remove this entry (and the matching doc line and test expectation), or cite the Cline source that dispatches it.

backend/internal/adapters/agent/cline/activity.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	recent := lines[start:]
+
+	for i := len(recent) - 1; i >= 0; i-- {
+		line := strings.ToLower(recent[i])
+		if strings.Contains(line, "thinking... (esc to cancel)") {
+			return domain.ActivityActive, true
+		}
+		if clineEmptyComposer(recent[i]) && clineModeFooterAfter(recent, i) {
+			return domain.ActivityIdle, true
```
AgentWrapper, line 37: Verification ask: if Cline renders a tool-approval prompt while the empty composer and mode footer are still visible in the same frame, this branch would flip a hook-derived waiting-input state to idle on every observer tick. If the approval UI replaces the composer (as in the muse adapter world), this is fine — a third fixture capturing an approval frame would pin that assumption.

backend/internal/adapters/agent/cline/cline.go
```diff
@@ -39,6 +39,7 @@ func New() *Plugin {
 
 var _ adapters.Adapter = (*Plugin)(nil)
 var _ ports.Agent = (*Plugin)(nil)
+var _ ports.ContinuousTerminalActivityDetector = (*Plugin)(nil)
```
AgentWrapper, line 42: Heads-up on a side effect beyond the observer: now that the plugin satisfies `ports.TerminalActivityDetector`, `WaitForMessageDeliveryReady` takes its detector-only branch, which never falls back after `messageDeliveryReadyFallback` the way the hook path does. A frame the classifier cannot authoritatively read (typed draft, wrapped footer, label drift in a future Cline) blocks delivery until the caller context times out. Bounded today, but worth knowing this interface implies that contract.

AgentWrapper (commented):
**Verdict: requesting changes** — one finding to fix or justify; the rest of the change looks solid.

The classifier and its wiring check out: the newest-marker backward scan mirrors the muse adapter pattern, the hook path is correctly routed (`stop` -> `StandardDeriveActivityState` -> idle via `activitydispatch`), the observer only reconciles continuous detectors from active/idle/waiting-input states, and the fixtures pin realistic Cline 3.0.55 frames including the transcript-rejection case. Test coverage of the new surface is good.

### Needs changes: `TaskError` is not a Cline hook event

I could not find `TaskError` anywhere in cline/cline:

- the hooks registry (`apps/vscode/src/core/hooks/hook-factory.ts`) defines exactly `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `TaskStart`, `TaskResume`, `TaskCancel`, `TaskComplete` — no `TaskError`;
- the in-repo `.clinerules/hooks/README.md` lists the same set (plus `TaskResume`), with no `TaskError`;
- the CLI maps error-terminated runs onto the **TaskCancel** hook (`apps/cli/src/utils/hooks.ts` builds a `taskCancel` payload for both abort and non-abort errors from `agent_abort`).

Consequences: the installed `.clinerules/hooks/TaskError` script will never execute; the PR description claim that failed turns clear `active` via a native `TaskError` hook is inaccurate (they clear via the already-installed `TaskCancel` hook, plus the new classifier as fallback); and `TestClineManagedHooksClearCompletedTurns` pins a nonexistent event. Please drop the `TaskError` entry and its doc/test lines, or point at the Cline source that dispatches it if a version-specific path exists that I missed.

### Notes (no change required)

- Implementing `TerminalActivityDetector` also flips `WaitForMessageDeliveryReady` into its detector-only branch, which has no degraded 5s fallback — if the rendered frame never matches (typed draft in the composer, wrapped footer on a narrow pane, future label drift), readiness only resolves at the caller timeout. Callers are bounded today (title refinement), and codex/muse share this exposure, so this is a heads-up, not a blocker.
- Worth a quick manual check: when Cline shows a tool-approval prompt (auto-approve off), confirm the empty `Ask anything...` composer plus mode footer is not rendered simultaneously — if it were, the continuous classifier would downgrade waiting-input to idle on every tick. If you have a captured approval frame, pinning it as a third fixture would lock this in.

backend/internal/adapters/agent/cline/hooks.go
AgentWrapper: Addressed in `97d1b0e3b`. I removed the unsupported `TaskError` managed hook, its documentation line, and its test expectation. I also corrected the PR description and issue follow-up: successful completion uses the new `TaskComplete` hook, while error termination is already covered by Cline’s existing `TaskCancel` dispatch plus terminal reconciliation as fallback.

backend/internal/adapters/agent/cline/activity.go
AgentWrapper: Acknowledged. The live reproduction had auto-approve enabled, so I do not have an honest captured approval frame to add as a fixture. I have not invented one. The classifier still requires the exact empty placeholder plus the Plan/Act footer, and newest authoritative markers win; this remains a useful follow-up when a real non-auto-approved frame is available. No change requested by this note.

backend/internal/adapters/agent/cline/cline.go
AgentWrapper: Acknowledged and documented in the PR’s Risks / follow-ups section. I kept the scope aligned with the existing Codex/Muse detector contract because callers are bounded today; a degraded readiness fallback should be handled generically rather than only in the Cline adapter. No change requested by this note.

AgentWrapper:
Addressed [review 4948387325](https://github.com/Untrivial-ai/agent-orchestrator/pull/4049#pullrequestreview-4948387325) in commit `97d1b0e3b`: removed the nonexistent `TaskError` hook, documentation, and test expectation; corrected the PR and issue text to state that Cline error runs dispatch `TaskCancel`; and retained only the valid `TaskComplete` addition. Focused tests, `go build ./...`, and isolated `go test ./...` pass. I also replied to each inline thread, including the two non-blocking notes.

backend/internal/adapters/agent/cline/activity.go
AgentWrapper: Verified against the Cline CLI render path: ChatView renders InlineToolResponse when runtimeInteraction is present and renders InputBar only in the else branch, while StatusBar remains outside that conditional. A tool approval therefore replaces the empty composer but keeps the mode footer visible. I added tool_approval.txt mirroring that frame and TestDetectTerminalActivityDoesNotOverrideToolApproval, which asserts the detector emits no idle signal and therefore preserves the PreToolUse hook-derived waiting-input state. Full go test ./... and go build ./... pass. Pushed in cc01a8568.

backend/internal/adapters/agent/cline/cline.go
AgentWrapper: Confirmed. This is the existing conservative TerminalActivityDetector contract, and I am intentionally not adding a degraded fallback in this scoped fix. For Cline, an inconclusive frame includes a typed draft; falling back from persisted idle after five seconds could inject the title-refinement message over that draft. The sole WaitForMessageDeliveryReady caller is the background delegated-title refinement path, which is bounded by delegatedTaskTitleRefinementTimeout (one minute), so classifier drift degrades optional title refinement rather than blocking task creation. Decoupling continuous observer detection from delivery-readiness authority would be a broader cross-adapter design change affecting Codex and Muse too, and should not ride this Cline activity fix. No code change is warranted for this note.

nikhilachale (approved):
Tested locallly, LGTM
No blocking found

PR #4051, opened 2026-08-16
feat(notifications): beep when an agent needs input or a PR is ready

Vaibhaav-Tiwari:
## Summary
- Plays the OS system beep (Electron `shell.beep()`) alongside the existing toast/dock-bounce/taskbar-flash whenever a `needs_input` or `ready_to_merge` notification fires and the window is backgrounded, so you notice without watching the window.
- Adds a "Sound notifications" toggle (on by default) in Settings -> General, persisted through the existing `UiSettings` file.
- Fixes `uiSettings:set` to merge a partial write onto the persisted settings instead of replacing the whole blob — needed once two independent settings (locale, sound) share that one file, otherwise changing one would silently reset the other to its default.

## Test plan
- [x] `npm test` (frontend) — full suite run twice (before/after an i18n-parity fix) and diffed against a clean baseline: identical set of pre-existing failing files (unrelated `packages/product-ui` dependency gap), zero new failures, all new/modified tests pass.
- [x] `npx tsc --noEmit` — clean aside from the same pre-existing `packages/product-ui` errors.
- [ ] Manual: launch the desktop app, background the window, trigger a `needs_input`/`ready_to_merge` notification, confirm the beep plays and the Settings toggle suppresses it. **Not verified in this environment** — the sandbox's Go 1.24.5 is below the daemon build's required Go 1.25.7+, so the Electron app couldn't be launched here. Please verify locally before merging.


frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
```diff
@@ -95,6 +96,9 @@ export function GeneralSettingsSection({
 	const setLocale = useLocaleStore((state) => state.setLocale);
 	const localeSaving = useLocaleStore((state) => state.saving);
 	const localeSaveError = useLocaleStore((state) => state.saveError);
+	const soundNotificationsEnabled = useSoundNotificationsStore((state) => state.enabled);
+	const setSoundNotificationsEnabled = useSoundNotificationsStore((state) => state.setEnabled);
+	const soundNotificationsSaving = useSoundNotificationsStore((state) => state.saving);
```
whoisasx, line 101: **[P2] Surface failed sound-setting writes in Settings.** The store records `saveError` when `uiSettings:set` fails, but this component subscribes only to `enabled` and `saving`. The switch consequently remains at its previous value with no explanation, making the control appear unresponsive during disk/IPC failures. Please subscribe to `saveError`, render an accessible error message as the locale row already does, and add a form-level rejection test.

whoisasx (dismissed):
## Review summary

The core implementation is sound: the beep is limited to backgrounded `needs_input` and `ready_to_merge` notifications, the main-process policy updates after a successful toggle write, and partial UI-settings writes are serialized, atomic, and preserve the locale. The PR also merges cleanly with current `main`.

One product-state issue remains: the sound-notifications store records persistence failures, but Settings never renders that error. A failed toggle therefore appears to do nothing and gives the user no recovery context.

Verification on `7b6983aa` merged with current `main`: 48 focused tests passed and `npm run typecheck:e2e` passed. GitHub Frontend and gitleaks checks are green. The actual OS beep still lacks manual runtime verification, so please include local audio evidence before merging.

frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
Vaibhaav-Tiwari: Fixed in b1c40e080: subscribed to `saveError` and render an alert paragraph on failure, mirroring the existing `settings.language.saveFailed` row. Added a rejection test ("keeps the current sound notifications value and reports a persistence failure") in `GlobalSettingsForm.test.tsx`, verified red (fails without the fix) then green.

Vaibhaav-Tiwari:
Addressed the changes requested:

- **Sound-setting save failures now surfaced** (b1c40e080): `GeneralSettingsSection` now subscribes to `saveError` from the sound-notifications store and renders an alert paragraph on failure, matching the existing language-save-error row. Added a rejection test verified red (fails without the fix) then green.
- **Manual audio verification**: still unable to produce this myself — this sandbox's Go (1.24.5) is below the daemon build's required Go 1.25.7+, so `npm run dev` can't launch the Electron app here. This remains the one open item; please verify the actual beep locally before merging.

codebanditssss (approved):

PR #4053, opened 2026-08-16
fix: order sidebar workers by last activity

neversettle17-101:
## Summary
- order project worker sessions by agent activity timestamp, newest first
- fall back to session update/creation timestamps for older or incomplete records
- add sidebar regression coverage for SCM updates not representing agent activity
- Feedback from the user: https://discord.com/channels/1476302178913357958/1526316936546357258/1537880185096118365

## Tests
- `git diff --check`
- Focused Vitest/typecheck attempted; blocked by missing root workspace dependencies for `packages/product-ui` (React, clsx, tailwind-merge).

### Before
Screenshot 2026-08-16 at 2 23 26 PM

### After
Screenshot 2026-08-16 at 2 23 39 PM

## Risk
- Sidebar ordering only; no API or persistence changes.

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -474,6 +474,29 @@ describe("Sidebar", () => {
 		expect(navigateMock).not.toHaveBeenCalled();
 	});
 
+	it("lists worker sessions by last activity, newest first", () => {
+		const oldest: WorkspaceSession = {
+			...session,
+			id: "proj-1-old",
+			title: "old task",
+			createdAt: "2026-06-29T00:00:00Z",
+			updatedAt: "2026-07-02T00:00:00Z",
+			activity: { state: "idle", lastActivityAt: "2026-07-01T00:00:00Z" },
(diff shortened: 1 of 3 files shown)


codebanditssss (changes requested):
`sessionLastActiveNewer` does not implement the stated fallback for older or incomplete records. Missing or invalid `lastActivityAt` is converted to `0`, so any session with an activity timestamp sorts above a newer session without activity data, regardless of that newer session’s `updatedAt` or `createdAt`.

Please calculate one effective timestamp per session—valid `activity.lastActivityAt`, otherwise valid `updatedAt`, otherwise `createdAt`—and sort by that value. Add regression cases where one session has no activity and where `lastActivityAt` is invalid.

codebanditssss (approved):
The fallback issue is fixed: each session now derives one effective timestamp from valid `activity.lastActivityAt`, then `updatedAt`, then `createdAt`. The added regression coverage exercises missing activity, invalid activity/update timestamps, and creation-time fallback. Current renderer smoke, scan, and test checks all pass.

PR #4065, opened 2026-08-16
feat(frontend): drag-and-drop a folder to open it as a project

Vaibhaav-Tiwari:
## Summary
- Drop a folder anywhere in the app window (welcome board, an open project, a session view) to add it as a project — the existing Workspace/Project picker opens with the folder's path already resolved, skipping only the native OS file-browser step. Uses `webUtils.getPathForFile()`, exposed through a new synchronous preload bridge method.
- Drop a folder onto AO's taskbar icon or a desktop/Start-menu shortcut to do the same, whether the app is already running or not. Reuses the existing single-instance-lock/`second-instance` argv handling (previously only used for `ao-app://` deep links) to detect a dropped folder path and relay it to the renderer, queuing through the existing `TRAY_RENDERER_READY_CHANNEL` signal when the app is cold-starting.

Mirrors VS Code's "drop a folder to open it" behavior end to end.

## Test plan
- [x] `npm run typecheck` / `npm run typecheck:e2e` clean
- [x] Full `vitest` suite: no regressions (same 17 pre-existing macOS/landing-only failures as `main`)
- [x] New unit/integration coverage: preload `getPathForFile` bridge, `ui-store` request signal, `CreateProjectFlow` dropped-path handling, the global window drop listener + overlay, and the icon-drop argv parser
- [x] Manually verified in the real Electron app on Windows: in-window drop (welcome board, open project, session view), icon-drop cold start (app not running), and icon-drop second-instance (app already running)


frontend/src/renderer/routes/_shell.tsx
```diff
@@ -145,6 +149,61 @@ function ShellLayout() {
 		document.addEventListener("click", handleModifierLinkClick);
 		return () => document.removeEventListener("click", handleModifierLinkClick);
 	}, []);
+	// Drop a folder anywhere in the app window to add it as a project, mirroring
+	// VS Code's "drop a folder to open it". A depth counter (not a relatedTarget
+	// check) tracks dragenter/dragleave so the overlay doesn't flicker as the
+	// pointer crosses child-element boundaries. XtermTerminal's own drop handler
+	// calls stopPropagation, so a drop landing on an active terminal pane never
+	// reaches this window-level listener — no special-casing needed there.
```
whoisasx: [P1] This assumption breaks the advertised 'drop anywhere' flow. Xterm only stops the drop event; the preceding dragenter still bubbles to this window listener, increments dragDepthRef, and can show the overlay. When the user releases over a terminal, Xterm consumes the drop, so this handler never clears the overlay/depth and never calls requestCreateProjectFromPath. The folder is instead sent through the terminal attachment path. Please route directory drops from the terminal into this shared import flow (or let only directory drops bubble) and add a regression that dispatches the drop through a terminal host.

frontend/src/main/open-folder-arg.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Windows/Linux: dropping a folder onto the app's taskbar icon or a desktop/
+// Start-menu shortcut launches (or, if already running, re-signals) the app
+// with the folder's path as a plain positional argv entry — unlike every
+// other argument this app's own launchers pass (--installed-via=, the
+// ao-app:// deep link, Electron/Chromium's own --flags). A real directory is
+// a reliable signal: nothing else ever present in this app's argv resolves
+// to one, so the first argv entry that does is the dropped folder.
+export function parseOpenFolderPathArg(argv: string[]): string | undefined {
+	for (const entry of argv) {
```
whoisasx: [P2] Scanning every directory-valued argv entry also captures Electron's own application argument. electron-forge start defaults appPath to '.' and spawns Electron with that value before user args, so in development argv[1] resolves to the real frontend directory. Every cold start can therefore queue AO's own source directory as a dropped project, and an ordinary second-instance launch can do the same. Exclude the Electron bootstrap app slot (for example when process.defaultApp is true) before examining positional drop candidates, and cover a real directory-valued app argument in this test suite.

whoisasx (changes requested):
Two blocking integration issues remain.

The window-level drop handler does not own folder drops over terminal panes: Xterm sees the dragenter at window level but stops the drop itself, leaving the overlay/depth stale and bypassing project import. The argv parser also treats Electron Forge's directory-valued app bootstrap argument as a dropped folder in development.

Verification completed on d41efdde: frontend and E2E typechecks passed, all 51 focused tests passed, the PR merges cleanly with current main, and the same verification passed on that synthetic merge. GitHub CI is green, but the added tests do not exercise these two event/argv shapes.

Please address both inline findings before approval.

frontend/src/renderer/routes/_shell.tsx
Vaibhaav-Tiwari: Confirmed — Xterm only stops the drop event, not dragenter, so the depth counter/overlay could go stale, and the drop was absorbed into the terminal's file-attach flow instead of reaching this handler.

Fixed in 9cb2ffd6: XtermTerminal's `dropInput` now checks `webkitGetAsEntry()?.isDirectory` first and returns early (no `preventDefault`/`stopPropagation`) for a directory drop, letting it bubble untouched to this window-level listener. Added `lets a dropped folder bubble to the window instead of attaching it as a file` in XtermTerminal.test.tsx, which dispatches a directory-shaped drop through a rendered terminal host and asserts it reaches a window-level listener while `saveDroppedFile` is never called. Confirmed it fails without the fix (event swallowed, 0 calls to the window listener) and passes with it.

frontend/src/main/open-folder-arg.ts
Vaibhaav-Tiwari: Confirmed — `electron-forge start` spawns Electron with its appPath (defaults to `.`, a real directory) at argv[1] before user args, so every dev cold start and second-instance relaunch would misread that as a dropped folder.

Fixed in 9cb2ffd6: `parseOpenFolderPathArg` now skips index 1 when `process.defaultApp` is set, mirroring the existing `registerCloudProtocol` convention in cloud-auth.ts (`process.defaultApp && process.argv.length >= 2` → argv[1] is the bootstrap app path). Added two cases to open-folder-arg.test.ts under `process.defaultApp` — one confirming the bootstrap slot alone is ignored, one confirming a genuinely dropped folder listed after it is still found.

frontend/src/main.ts
```diff
@@ -1798,7 +1806,13 @@ ipcMain.handle("notifications:setBadge", (_event, count: number) => {
 
 ipcMain.on(TRAY_SET_ATTENTION_STATE_CHANNEL, (event, state) => trayLifecycle.handleSetAttentionState(event, state));
 
-ipcMain.on(TRAY_RENDERER_READY_CHANNEL, (event) => trayLifecycle.handleRendererReady(event));
+ipcMain.on(TRAY_RENDERER_READY_CHANNEL, (event) => {
+	trayLifecycle.handleRendererReady(event);
+	if (pendingFolderPath && event.sender === getShellWebContents()) {
```
whoisasx, line 1817: **[P1] Wait for the folder listener, not the tray listener, before flushing this path.** `TRAY_RENDERER_READY_CHANNEL` is sent when `TrayRuntime`'s child effect registers its tray listener. React runs that child effect before `ShellLayout`'s parent effect registers `app.onOpenFolderPath`, so this `send` has no renderer listener and then clears `pendingFolderPath`. I reproduced the real ordering in the shell harness: a path flushed during tray readiness left `folderDropRequest` as `null`. A `second-instance` event arriving during startup can be lost for the same reason. Please add a folder-specific readiness handshake emitted after `onOpenFolderPath` installs its listener, or buffer folder IPC in preload, and cover the cold-start ordering in a regression test.

frontend/src/renderer/routes/_shell.tsx
```diff
@@ -145,6 +149,62 @@ function ShellLayout() {
...
+	useEffect(() => {
+		const isFileDrag = (event: DragEvent) => Array.from(event.dataTransfer?.types ?? []).includes("Files");
+		const firstEntryIsDirectory = (event: DragEvent) => {
+			const item = event.dataTransfer?.items?.[0];
+			return item?.webkitGetAsEntry?.()?.isDirectory ?? false;
+		};
+		const handleDragEnter = (event: DragEvent) => {
+			if (!isFileDrag(event)) return;
+			event.preventDefault();
+			dragDepthRef.current += 1;
```
whoisasx, line 170: **[P2] Do not leave the drag-depth counter stale when a child consumes a non-directory drop.** A regular file drag over Xterm bubbles its `dragenter` here and increments this counter, but Xterm calls `stopPropagation()` on the corresponding `drop`, so `handleDrop` never resets it. The next folder drag increments from a positive value and misses the `dragDepthRef.current === 1` branch, suppressing the overlay. I reproduced this exact file-drop-then-folder-drag sequence. Let the terminal's file drop bubble after its own handling so the window can reset and ignore it, or add another reliable shared reset path, with a sequence regression test.

whoisasx (changes requested):
## Review summary

The two previously requested fixes are implemented correctly: directory drops now escape Xterm's file-attachment handler, and Electron's development bootstrap argument is excluded from folder-path parsing.

Two behavioral blockers remain:

- Cold-start and early second-instance folder paths can be emitted before the renderer's folder listener exists, permanently losing the request.
- A regular file drop consumed by Xterm can leave the window-level drag-depth counter stale and suppress the overlay on the next folder drag.

The PR also currently conflicts with `main` in `CreateProjectFlow.tsx`. Please preserve the newer remote-clone source flow while integrating dropped local paths during conflict resolution.

Verification on `e38b8a9d`: 110 focused tests passed, `npm run typecheck` passed, and `npm run typecheck:e2e` passed. I added focused regression cases in the isolated review checkout and reproduced both findings. GitHub CI is green, but it does not cover these event-order sequences.

frontend/src/main.ts
Vaibhaav-Tiwari: Confirmed — TrayRuntime's child effect (whose ready-ping triggers this flush) runs before ShellLayout's own parent effect that calls `app.onOpenFolderPath`, so a path flushed at just the wrong moment reached main's `.send()` with no renderer listener yet attached and was lost.

Fixed in d85d9c3: buffered `app:openFolderPath` at the preload layer itself. Preload registers a raw `ipcRenderer.on("app:openFolderPath", ...)` at module load (guaranteed before any renderer/React code runs) that stashes an early-arriving path; `onOpenFolderPath(listener)` now replays that buffered path immediately if one is already waiting, in addition to registering for future ones. This closes the race regardless of any React effect ordering rather than depending on it — no new main-process readiness channel needed.

Added `frontend/src/preload.test.ts` coverage: replays a path that arrived before `onOpenFolderPath` was called, does not replay an already-consumed path twice, and still delivers a path that arrives normally after the listener is registered.

frontend/src/renderer/routes/_shell.tsx
Vaibhaav-Tiwari: Confirmed — XtermTerminal's own drop handler called `stopPropagation()` after attaching a plain (non-directory) file, so this window-level listener's `handleDrop` never ran and never reset `dragDepthRef`. The next folder drag then incremented from that stale nonzero depth, missed the `=== 1` branch, and never showed the overlay.

Fixed in d85d9c3: `XtermTerminal.tsx`'s file-drop path now keeps `preventDefault()` but drops the `stopPropagation()` call, so the event still bubbles here after Xterm's own attach handling — this listener's existing reset logic (`dragDepthRef.current = 0; setIsDragActive(false);`, unconditional before the directory check) then runs regardless, and simply no-ops further since it isn't a directory.

Updated `XtermTerminal.test.tsx`'s file-drop test to assert the drop now does bubble to `window` (previously asserted the opposite), and added a new sequence regression in `shell-new-session-shortcut.test.tsx`: dragenter+drop a plain file, then dragenter a folder, asserting the overlay still shows.

frontend/src/main.ts
```diff
@@ -1856,6 +1870,20 @@ app.on("second-instance", (_event, argv) => {
...
+	// — send directly rather than queuing via pendingFolderPath.
+	const folderPath = parseOpenFolderPathArg(argv);
+	if (folderPath) {
+		const window = BaseWindow.getAllWindows()[0];
+		if (window) {
+			if (window.isMinimized()) window.restore();
+			window.show();
+			window.focus();
+		}
+		getShellWebContents()?.send(OPEN_FOLDER_PATH_CHANNEL, folderPath);
```
whoisasx: **[P1] Queue this path when the shell WebContents does not exist yet.** A `second-instance` event can arrive while the first process is still inside `app.whenReady()` or creating its window. In that state `getShellWebContents()` is undefined, this optional send is a no-op, and the handler immediately returns, so preload never receives anything to buffer. Please store the path in `pendingFolderPath` until renderer readiness (and test the no-window startup case) instead of assuming an already-running process always has a mounted renderer.

frontend/src/preload.ts
```diff
@@ -76,6 +76,18 @@ export type ImportFolderScan = {
...
 
+// A folder-drop path can arrive (cold start, or an early second-instance)
+// before ShellLayout's own effect has registered app.onOpenFolderPath's
+// listener below — React mounts TrayRuntime's child effect (which pings
+// main.ts's readiness flush) before ShellLayout's parent effect that installs
+// this listener. Registering this raw listener here, at preload module load
+// (guaranteed to run before any renderer/React code), buffers a path that
+// arrives too early so onOpenFolderPath can still replay it once called.
+let bufferedOpenFolderPath: string | null = null;
+ipcRenderer.on("app:openFolderPath", (_event, path: string) => {
```
whoisasx, line 90: **[P2] Buffer only when there is no active subscriber.** This always-on listener stores every path, while `onOpenFolderPath` also registers a second IPC listener that delivers the same event immediately. The buffered copy is therefore left behind after a normal delivery and is replayed if the listener later unsubscribes and resubscribes. I reproduced that sequence: the second subscriber received the already-consumed path. Please use one module-level dispatcher that either calls the active listener or buffers the path, and add a normal-delivery-then-resubscribe regression test.

frontend/src/renderer/components/CreateProjectFlow.tsx
```diff
@@ -160,6 +173,16 @@ export function CreateProjectFlow({
 		startFlow();
 	}, [openSignal]);
 
+	// A folder was dropped on the app window. Ignored while the flow already has
+	// UI on screen so an in-progress manual selection is never silently discarded.
+	const lastDropNonce = useRef(droppedPath?.nonce);
+	useEffect(() => {
+		if (!droppedPath || droppedPath.nonce === lastDropNonce.current) return;
+		lastDropNonce.current = droppedPath.nonce;
+		if (isBusy || modePickerOpen || folderPickerOpen || selectedPath !== null) return;
```
whoisasx: **[P2] Treat the clone dialog as an already-active create-project flow.** After merging the new source picker, this guard covers the mode picker, folder picker, busy operations, and agent sheet, but not `cloneDialogOpen`. Dropping a folder while the clone form is open therefore calls `startFlow`, opens the source picker over the clone modal, and stages the dropped path despite the comment's promise to ignore active flows. I reproduced the overlapping dialogs in a focused test. Include `cloneDialogOpen` in this guard and cover the sequence.

whoisasx (dismissed):
## Re-review summary

The Xterm drag-depth fix is correct, the original cold-start flush race is fixed, and the `CreateProjectFlow` conflict was integrated without removing the newer clone functionality.

Three issues remain on the latest head:

- A `second-instance` folder launch can still be discarded if it arrives before the first instance has created its shell WebContents.
- The preload's dual-listener implementation retains normally delivered paths and can replay them after a later subscription.
- The dropped-folder guard omits the newly merged clone dialog, allowing the source picker to open over an active clone flow.

Verification on `d85d9c35`: the unmodified 114 focused tests passed and `npm run typecheck:e2e` passed. I added focused regression cases for the preload replay and clone-dialog overlap; both failed as described. GitHub Frontend and gitleaks checks are green.

frontend/src/main.ts
Vaibhaav-Tiwari: Confirmed and fixed in 778320b. \`second-instance\` now falls back to the same \`pendingFolderPath\` queue the cold-start path already uses when \`getShellWebContents()\` isn't available yet (still focuses the window immediately either way), flushed once \`TRAY_RENDERER_READY_CHANNEL\` fires.

I did not add a dedicated regression test for this exact branch — \`main.ts\` has no existing unit-test harness at all (no test file imports it; every \`app.whenReady()\`/\`second-instance\`/IPC handler in the file is exercised only by the manual Electron verification, not vitest), so this mirrors the same untested status as the surrounding cold-start queuing and deep-link \`second-instance\` handling it sits beside. Building a harness capable of driving this (mocking \`app\`, \`BaseWindow\`, \`ipcMain\`, and the dozen or so \`./main/*\` modules \`main.ts\` wires up at import time) would be a much larger, separate effort than this fix — happy to scope that as follow-up work if you'd like it prioritized.

frontend/src/preload.ts
Vaibhaav-Tiwari: Confirmed and fixed in 778320b, exactly as suggested — replaced the two-listener setup (always-on buffer + per-call wrapped listener) with a single \`ipcRenderer.on\` dispatcher that either forwards straight to the active listener or buffers, never both. \`onOpenFolderPath\` no longer registers its own IPC listener at all; it just points \`activeOpenFolderPathListener\` at the caller's callback and drains any buffered path synchronously.

Added the normal-delivery-then-resubscribe regression to \`preload.test.ts\`: subscribe, deliver a path normally (dispatched directly, buffer untouched), unsubscribe, then resubscribe a fresh listener and assert it receives nothing.

frontend/src/renderer/components/CreateProjectFlow.tsx
Vaibhaav-Tiwari: Confirmed and fixed in 778320b — added \`cloneDialogOpen\` to this guard alongside the existing checks, so a folder dropped while the Clone-from-Git dialog is open is now ignored the same way a drop during the agent sheet or folder dialog already was.

Added a regression to \`CreateProjectFlow.test.tsx\`: open the mode picker, switch to the clone flow, drop a folder, and assert the clone dialog stays on screen with no source-picker button rendered and no native chooser invoked.

codebanditssss (approved):

PR #4073, opened 2026-08-17
fix(desktop): add AO_DISABLE_GPU escape hatch for broken GPU drivers

### LINKED ISSUE #3961: GPU process crashes 3x on startup (SIGABRT) on Wayland + hybrid GPU; no way to disable hardware acceleration
## Bug

On Linux with Wayland and a hybrid Intel + NVIDIA laptop GPU, AO's GPU process crashes three times while the app starts. Chromium then gives up on the GPU and runs in software mode. The app still works, but the first ~35 seconds are unstable, and rendering stays on the slow software path afterwards.

The crash itself comes from the graphics driver, not from AO's code. The AO-side problem is that there is **no way to opt out**. AO sets no Electron GPU or Ozone options at all, so a user on a broken driver stack cannot tell the app to skip the GPU. Their only options are editing the source or living with it.

**Analyzed against:** `35da72b6c78491227fd5dbe3eaed2b8b0599f045` | **AO version:** 0.10.3
**Confidence:** High

## Environment

- OS: Omarchy (Arch Linux), kernel 7.1.4-arch1-1
- Compositor: Hyprland on Wayland, `ELECTRON_OZONE_PLATFORM_HINT=wayland`
- GPU: hybrid — Intel Raptor Lake-P UHD (`8086:a7a8`) + NVIDIA RTX 5050 Max-Q (`10de:2d98`)
- mesa 26.1.5, Electron 33.4.11, node 26.2.0
- Started with `cd frontend && npm run dev`

## Reproduction

1. On a Wayland session with a hybrid Intel + NVIDIA GPU, run `cd frontend && npm run dev`.
2. Watch the terminal output for the first minute.
3. The GPU process crashes three times, about 18 seconds apart, then stops.

This happened on two separate runs of the app, so it is not a one-off:

| Run | Crash 1 | Crash 2 | Crash 3 |
| --- | --- | --- | --- |
| First | 22:44:34 | 22:44:52 | 22:45:09 |
| Second | 22:48:22 | 22:48:40 | 22:48:57 |

## What the logs show

Each crash is `exit_code=134`, which is SIGABRT:

```
[521942:0813/224822.770984:ERROR:gpu_process_host.cc(982)] GPU process exited unexpectedly: exit_code=134
[521942:0813/224822.771008:WARNING:gpu_process_host.cc(1416)] The GPU process has crashed 1 time(s)
[521942:0813/224840.311716:ERROR:gpu_process_host.cc(982)] GPU process exited unexpectedly: exit_code=134
[521942:0813/224840.311729:WARNING:gpu_process_host.cc(1416)] The GPU process has crashed 2 time(s)
[521942:0813/224857.825508:ERROR:gpu_process_host.cc(982)] GPU process exited unexpectedly: exit_code=134
[521942:0813/224857.825547:WARNING:gpu_process_host.cc(1416)] The GPU process has crashed 3 time(s)
```

Before each crash, the GPU process cannot allocate a buffer, and the Intel driver reports it is out of memory:

```
[ERROR:gbm_pixmap_wayland.cc(82)] Cannot create bo with format= RGBA_8888 and usage=SCANOUT
intel: the execbuf ioctl keeps returning ENOMEM
```

Those appear 9 and 3 times respectively across the run — roughly once per GPU process restart.

After the third crash, Chromium stops retrying. There are no further GPU crashes in the log, which continues normally for another 15 minutes.

## Root cause

The crash is in the driver and compositor stack, so AO cannot fix it directly.

What AO controls is whether a user can work around it, and right now they cannot:

- `frontend/src/main.ts` — searching for `commandLine`, `appendSwitch`, `appendArgument`, `disableHardwareAcceleration`, or `ozone` returns **zero** matches. No GPU or Ozone options are set anywhere.
- `frontend/forge.config.ts` — no GPU or Ozone flags.
- The only `--disable-gpu` in the repo is in a test harness (`test/e2e-pod/real-app.spec.ts:60`) and does not affect the shipped app.

## Suggested fix

Small and self-contained: let the user turn hardware acceleration off with an environment variable.

In `frontend/src/main.ts`, before `app.whenReady()`:

```ts
if (process.env.AO_DISABLE_GPU === "1") {
	app.disableHardwareAcceleration();
}
```

Then add a line to the Linux troubleshooting notes in `docs/development.md`.

I would deliberately keep it at that. Auto-detecting repeated GPU crashes and restarting in software mode is not worth the complexity, because Chromium already falls back on its own after three crashes.

## Impact

- About 35 seconds of unstable startup on affected Linux/Wayland machines, with the GPU process dying and restarting three times.
- Rendering stays on the software path afterwards. Follow-on errors continue in the log: `Automatic fallback to software WebGL has been deprecated`, `Buffer handle is null. Not creating a mailbox from it.`, and `Creation of StagingBuffer's SharedImage failed.`
- No supported way to skip the crashing startup path — no flag, environment variable, or setting.
- **Not verified:** whether crashes return under heavy GPU load. The logs only cover an idle app after the fallback, so severity beyond startup is unproven.

## Related

No duplicate found. Searches with `--state all` for `hardware acceleration`, `disableHardwareAcceleration`, and `ozone` returned nothing. `wayland` returned only [#2491](https://github.com/Untrivial-ai/agent-orchestrator/issues/2491) and [#2778](https://github.com/Untrivial-ai/agent-orchestrator/issues/2778), both unrelated.

The WebGL errors listed above also show up in [#3787](https://github.com/Untrivial-ai/agent-orchestrator/issues/3787), where I reproduced a separate terminal crash on this machine. They are not the cause of that crash — the two are independent, and this issue covers only the GPU process problem.

I have the affected hardware and am happy to test a fix.

yashksaini-coder:
## What

Read `AO_DISABLE_GPU` before app ready in `frontend/src/main.ts` and call `app.disableHardwareAcceleration()` when it is set, so a user whose GPU driver stack crashes Chromium can start the app anyway.

## Why

Fixes #3961.

AO sets no Electron GPU or Ozone options anywhere — `frontend/src/main.ts` and `frontend/forge.config.ts` have zero matches for `commandLine`, `appendSwitch`, `disableHardwareAcceleration`, or `ozone`, and the only `--disable-gpu` in the tree is a test harness (`test/e2e-pod/real-app.spec.ts`). So a user on a broken driver has no supported way to skip the crashing path: patch the source, or live with a GPU process that aborts repeatedly until Chromium gives up and drops to software rendering.

The abort itself is not AO's or Electron's. Symbolizing the core dump from a crash on this machine puts it squarely in Mesa:

```
#2  abort ()                                              libc
#3  _iris_batch_flush   (ret = -12  /* -ENOMEM */)        iris_batch.c:995
#5  iris_fence_flush                                      iris_fence.c:247
#8  dri_create_fence_fd                                   dri_helpers.c:131
#9  dri2_create_sync    (EGL_SYNC_NATIVE_FENCE_ANDROID)   egl_dri2.c:2880
#14 EGL_CreateSyncKHR                                     <Electron's bundled libGLESv2.so>
```

The kernel's i915 batch-submission ioctl returns `-ENOMEM`; Mesa's iris driver does not classify that as a recoverable context loss, so it falls through to an unconditional `abort()` that kills the host process. That is a Mesa policy decision and the real fix belongs upstream in `mesa/mesa`, not here. What belongs here is the missing escape hatch.

## Correcting my own issue text

Two things in #3961 were wrong, and I'd rather amend them here than have a reviewer find them:

1. **"No way to disable hardware acceleration" overstates what this achieves.** `app.disableHardwareAcceleration()` does **not** stop a GPU process from spawning — `getAppMetrics()` still reports a `GPU` type. What it drops is the *system GL driver* — Mesa here — falling that process back to Electron's bundled SwiftShader. That covers the reported abort, which happens inside Mesa, but it would not help a crash inside ozone-wayland/gbm: the child still carries `--ozone-platform=wayland` and still maps the system buffer allocator.
2. **The issue suggested `process.env.AO_DISABLE_GPU === "1"`.** That is the weakest parse in the repo and would silently ignore `AO_DISABLE_GPU=true` from a user whose app is crashing — the worst possible failure mode for an escape hatch. This PR uses the `keepDaemonAlive()` allowlist instead (`frontend/src/main/daemon-owner.ts`): trim, lowercase, accept `"1" | "true" | "yes" | "on"`, treat everything else as unset. That file's own comment explains why an allowlist beats "anything but 0/false" — it stops `"off"`/`"no"` from doing the opposite of what the user meant.

## How

One guard in the existing pre-ready block in `frontend/src/main.ts`, between `app.setAppUserModelId` and the `~/.ao` `userData` pin. It sits above the `app.requestSingleInstanceLock()` / `app.exit(0)` bailout so it always runs, and `disableHardwareAcceleration()` throws if called after ready — verified it does.

Used the API rather than `app.commandLine.appendSwitch("disable-gpu")`. Measured on Electron 33.4.11: `getGPUFeatureStatus()` and the GPU child's switches are byte-identical between the two, but the API fails loudly when mis-ordered while a late `appendSwitch` fails silently. Did **not** add `--disable-software-rasterizer` — it flips `webgl`/`webgl2` to `disabled_off`, which would break WebGL in the renderer and the browser panel.

Docs: a row in the env table at `docs/cli/README.md` (`AO_KEEP_DAEMON` is the precedent there — also an Electron-main-only var), plus a symptom row in the frontend troubleshooting table in `docs/development.md`.

## Testing

Environment: Arch Linux (kernel 7.1.8), Hyprland/Wayland, hybrid Intel Raptor Lake-P UHD (`i915` + Mesa `iris`, the active renderer) + NVIDIA RTX 5050 Max-Q, Mesa 26.1.7, Electron 33.4.11.

**A/B on the real app**, same build, same session, `npm run dev` each way, inspecting the GPU child's `/proc/<pid>/maps` and `cmdline`:

| | flag unset | `AO_DISABLE_GPU=1` |
| --- | --- | --- |
| GPU-process `SIGABRT` (core dumps) | **2**, 18s apart | **0** |
| `GPU process exited unexpectedly: exit_code=134` | yes | **none** |
| `Cannot create bo ... usage=SCANOUT` | 6 | **0** |
| GL switches | none | `--use-gl=angle --use-angle=swiftshader-webgl` |
| Active GL driver | Mesa (`libgallium`, `libEGL_mesa`) on the Intel iGPU | Electron's bundled SwiftShader (`libvk_swiftshader.so`) |

Crashes counted from `coredumpctl`, not just the log — the terminal output undercounts them, because Chromium restarts the GPU child and the later aborts land after the first burst of log lines. The same pattern reproduced on an earlier run of this branch's base: three aborts at 18–20s intervals.

Both runs reached a working app — daemon listening on `127.0.0.1:3002`, `/api/v1/{projects,sessions,agents,notifications,shell-terminals}` all 200.

`npm run typecheck` → 0 errors. `npm test` → 189/189 files, 2485 passed, 1 skipped.

**No unit test, deliberately.** `frontend/src/main.ts` has zero `export` statements and no importer, so a test cannot load it without firing `app.setName`, `app.setPath`, `requestSingleInstanceLock`/`app.exit(0)` and `protocol.registerSchemesAsPrivileged` at import time; no `main.test.ts` has ever existed. Playwright can't reach it either — `playwright.config.ts` runs `dev:web` with `VITE_NO_ELECTRON=1`. Extracting a `shouldDisableGpu()` module purely to make one equality chain testable is what `AGENTS.md` line 82 rules out ("Add helpers only when they remove duplication across real call sites"). The sibling env flags in this file, `AO_OPEN_DEVTOOLS` and `AO_AGENT_BROWSER_PATH`, have no tests either. Happy to add one if you'd rather — the moment this guard takes a second input (a Wayland auto-detect, an `AO_OZONE_PLATFORM` switch, a `0`/`off` vocabulary) it should become `shouldDisableGpu(env, platform)` in `frontend/src/main/` with a table test next to `tray-lifecycle.test.ts`.

## Scope notes

- **Where the variable reaches.** `AO_DISABLE_GPU=1 ao start` works on Linux: `backend/internal/cli/start.go` passes no `Env` and `backend/internal/cli/process.go` leaves `cmd.Env` nil, so os/exec inherits the full environment. A menu/`.desktop` launch only sees the graphical session env, so those users need it in `~/.config/environment.d/` or the compositor env, not `~/.bashrc` — that entry is rewritten on every `ao start`, so hand-editing `Exec=` doesn't stick. On macOS `ao start` goes through `open --args`, which does not forward the caller's environment.
- The flag does not reach the agent-browser runtime child; `frontend/src/main/agent-browser-runtime.ts` rebuilds that process's env from a name allowlist with no `AO_*` entries. Irrelevant to this startup crash, but worth stating.
- A persisted in-app setting would be the nicer long-term shape, but it can't serve this case: the app has no window when the aborts happen, and `disableHardwareAcceleration()` must run pre-ready while the settings modules are async and post-ready.

## Checklist

- [x] Branched from `main`
- [x] Tests added/updated for user-visible behavior where it makes sense — justified above


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
@neversettle17-101 @illegalcall Anyone who can review the pr and update me

illegalcall (approved):

PR #4082, opened 2026-08-17
fix(chat): disable attachment-only steering sends

### LINKED ISSUE #4081: Attachment-only Chat drafts enable an impossible steering send
## Bug

During an active Chat turn, selecting **Steer this turn** and staging an attachment with an empty text field enables the send button even though the steering boundary only accepts text. Clicking send or pressing Enter then does nothing: no steer request, no queued message, and no feedback.

Found while separating the two independent defects originally bundled in [PR #4005](https://github.com/Untrivial-ai/agent-orchestrator/pull/4005).

## Reproduction

Analyzed against main at `d380279b5`.

1. Start a Chat turn and keep it streaming.
2. Select **Steer this turn**.
3. Attach a file.
4. Leave the text field empty.
5. Press Enter or click the enabled send button.

Expected: an impossible steering submission is visibly disabled. The attachment remains staged so the user can add steering text or switch to **Queue for next**.

Actual: send appears enabled but silently performs no action. The attachment remains with no explanation.

## Root cause

`ChatComposer.canSend` counts staged attachments as sendable in every mode, while the steering handler intentionally rejects empty text. The visual eligibility rule and the actual steering contract disagree.

## Impact

The primary composer action advertises an operation that cannot succeed, making user input appear swallowed. Switching to **Queue for next** is a workaround, but the UI does not guide the user there.

Priority: P2. Confidence: high.

## A/B evidence

**Before — current main enables an attachment-only steering no-op**

Before: attachment-only steering misleadingly enables send

**After — patched build disables the impossible action and retains the attachment**

After: attachment-only steering is disabled

Both captures use the real Electron desktop app with a live Claude turn and fresh isolated AO data. The patched run also verified that switching to **Queue for next** re-enables send with the same attachment still staged.

illegalcall:
Fixes #4081.

## What changed

While a Chat turn is running, attachment-only drafts no longer enable the composer send action when **Steer this turn** is selected. Steering accepts text only, so the button and Enter path now match that contract.

The staged attachment is deliberately retained. Adding text makes steering valid; switching to **Queue for next** immediately re-enables send and preserves the attachment for the queued message.

This was split out of #4005 during deep review so each PR owns one issue.

## Tests

`ChatSteer.test.tsx` now exercises the actual keyboard boundary:

- focuses the textarea before pressing Enter;
- proves attachment-only steering disables send;
- proves Enter calls neither steer, queue/send, nor attachment staging;
- proves the attachment stays present;
- proves switching to Queue re-enables submission and then stages/sends the attachment.

Verification:
- focused ChatSteer suite: 19 passed;
- frontend typecheck: passed;
- full frontend suite: 190 files; 2,501 passed and 1 skipped.

## Real Electron A/B evidence

Both runs used a live streaming Claude Chat turn, **Steer this turn**, `steer-proof.txt`, an empty textarea, and a fresh isolated `AO_DATA_DIR`. The screenshots are attached as review evidence and are not part of the PR code diff.

### Before — current main exposes an enabled no-op

Before: attachment-only steering misleadingly enables send

Pressing Enter sends no steer request, queues no message, and gives no feedback.

### After — impossible action is disabled

After: attachment-only steering is disabled

The attachment remains available, and the same live run verified that switching to Queue re-enables send.

### DIFF
--- a/frontend/src/renderer/components/chat/ChatComposer.tsx
+++ b/frontend/src/renderer/components/chat/ChatComposer.tsx
@@ -231,7 +231,11 @@ export function ChatComposer({
 	// with it: a steer with nothing in flight is refused, so it must never be what
 	// Enter is still pointing at.
 	const steering = Boolean(canSteer && onSteer) && delivery === "steer";
-	const canSend = hasDraft && !busy && !disabled && !steerPending;
+	// Attachments cannot be steered: the steer branch delivers text only and refuses
+	// an empty body. A staged file must not light up the send button on its own while
+	// steering is armed, or Enter would silently do nothing.
+	const canSend =
+		(text.trim().length > 0 || (staged && !steering)) && !busy && !disabled && !steerPending;
 	const canStopTurn = Boolean(willQueue && onInterrupt && !disabled && !hasDraft);
 	const draftSeedId = draftSeed?.id;
 	const draftSeedText = draftSeed?.text;
--- a/frontend/src/renderer/components/chat/ChatSteer.test.tsx
+++ b/frontend/src/renderer/components/chat/ChatSteer.test.tsx
@@ -1,9 +1,12 @@
-import { render, screen, within } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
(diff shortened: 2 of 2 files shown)


harshitsinghbhandari (approved):
LGTM

PR #4088, opened 2026-08-18
fix(lifecycle): deduplicate review feedback across internal and SCM lanes

ljlfriend:
## What

Prevent duplicate worker nudges when AO's internal reviewer and the SCM observer surface the same GitHub review feedback through separate lifecycle lanes.

This change:

- correlates an internal AO review with the later GitHub review observation;
- suppresses duplicate formal-review and inline-comment nudges for that same review;
- deduplicates review comments at the review-thread level so replies in an already-notified thread do not wake the worker again;
- persists the thread-level marker so the suppression survives a daemon restart;
- keeps genuinely new review threads actionable.

## Why

AO's internal reviewer delivery and SCM observation are separate lifecycle paths. The same GitHub review can therefore be delivered once by `ApplyReviewBatch` and then observed again by `ApplyPRObservation`, causing the owning worker to be nudged twice for one finding.

Inline review comments have a related duplicate path: a reply in the same review thread receives a new provider comment ID and can otherwise look like a new actionable finding.

## How

- Request each GraphQL review comment's parent `pullRequestReview.url` and carry it transiently through the SCM/PR observation ports. No database migration is added.
- Persist cross-lane coverage after successful internal review delivery using the GitHub numeric review/database ID.
- Suppress formal-review and inline-comment nudges when that same provider review was already delivered internally.
- Deduplicate review comments by `ThreadID` when available, falling back to comment ID when necessary.
- Persist the covered thread marker so suppression survives daemon restart.
- Fail open on coverage lookup errors so genuine review feedback is not silently dropped.

Intentional trade-off: follow-up comments in the exact same existing review thread remain deduplicated. A distinct new review thread or distinct new review still produces a nudge.

## Testing

Validated after rebasing onto current `main`:

```text
go test -count=1 ./internal/adapters/scm/github ./internal/lifecycle ./internal/observe/scm ./internal/ports
go build ./...
go vet ./...
git diff origin/main...HEAD --check
```
All of the above pass.

Regression coverage includes:

- internal review followed by the same formal SCM review;
- inline comment from the same SCM review;
- worker reply in the same review thread;
- persistence across daemon restart;
- a genuinely new review thread still nudging;
- GraphQL parent-review URL projection through transient observation types.

A pre-rebase `go test ./...` run on this Windows host had platform/environment-dependent failures. A clean detached worktree at the same base commit produced the same failing package set, so those failures were not introduced by this patch.

`go test -race` could not be executed locally because this Windows Go environment has `CGO_ENABLED=0` and no C compiler installed. CI should provide the authoritative race result.

## Checklist

- [x] Branched from `main` and rebased onto current `main`
- [x] One focused change; no matching existing issue/PR found
- [x] Tests added/updated for the affected behavior
- [ ] Relevant CI checks pass for the area touched

### DIFF
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -686,7 +686,7 @@ func buildReviewThreadsQuery(ref ports.SCMPRRef, beforeCursor string, includeRev
 	return fmt.Sprintf(`query{
 repo: repository(owner:%s,name:%s){ pullRequest(number:%d){ reviewDecision%s reviewThreads(last:%d, before:%s){ nodes{
   id isResolved path line
-  comments(first:%d){ nodes{ id body url author{ login __typename } } }
+  comments(first:%d){ nodes{ id body url pullRequestReview{ url } author{ login __typename } } }
 } pageInfo{ hasPreviousPage startCursor } } } }
 }`, graphQLString(ref.Repo.Owner), graphQLString(ref.Repo.Name), ref.Number, reviewSelection, githubReviewThreadPageSize, before, githubReviewCommentLimitPerThread)
 }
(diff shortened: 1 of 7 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

ljlfriend:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

neversettle17-101:
We have added a check in SCM that excludes the AO agent reviewer so we wouldn't need to deduplicate review feedback. let us know if you are facing the issue again.

PR #3923, opened 2026-08-13
fix(scm): reconcile stale transferred PR rows to prevent false merge conflict

AgentWrapper:
## Summary

- normalize the historical `AgentWrapper/agent-orchestrator` and current `Untrivial-ai/agent-orchestrator` repository identities before mutable branch/head alias checks
- collapse transferred PR rows by GitHub's stable PR number and retain the newest authoritative observation
- cover the live #2870 shape: stale `CONFLICTING`/`DIRTY` old-owner row plus newer `MERGEABLE`/`BLOCKED` current-owner row with a different head SHA
- verify a newer genuinely conflicting current observation is still surfaced

## Before / after

Before, both session and detailed PR endpoints returned two rows for transferred PR #2870. The desktop PR card and activity rail could therefore render the stale old-owner row's `conflicting` state.

After, both read models treat those URLs as aliases of the transferred repository and return only the newer current-owner row (`blocked`, `review_required`). Alias handling remains unchanged for unrelated repositories, and a current real conflict still wins.

## Verification

- read-only reproduction against local session `agent-orchestrator-31` through both daemon endpoints
- `cd backend && env -u AO_PROJECT_ID -u AO_SESSION_ID go test ./...`
- `npm run frontend:typecheck`
- `git diff --check`

No frontend source changed; the visible card/activity correction follows from the corrected daemon payload, so no separate preview artifact was applicable.

Closes #3922

Issue #3922, opened 2026-08-13
bug(scm): stale transferred PR row can surface false merge conflict


AgentWrapper:
## Bug
AO can continue to show a resolved/nonexistent merge conflict for a PR after the PR has moved from the old `AgentWrapper` GitHub URL to the current `Untrivial-ai` URL. GitHub and the current AO PR observation both say the PR is not conflicted, but the desktop still surfaces a stale `Merge conflict` state from an old PR row attached to the same session.

**Source:** local user report in AO desktop on August 13, 2026  
**Analyzed against:** `461a6df56` (`origin/main`, `feat: add per-PR CI failure injection policy (#3772)`)  
**AO daemon under test:** locally built `/tmp/ao`, `dev`, loopback daemon on `127.0.0.1:3001`  
**Confidence:** High — reproduced from the daemon API and traced to stale PR rows/aliasing.

## Screenshots

![PR card shows stale merge conflict](https://raw.githubusercontent.com/Untrivial-ai/agent-orchestrator/issue-assets-stale-pr-conflict-2870/.issue-assets/stale-pr-conflict-2870/pr-card.png)

![Activity rail shows Conflict](https://raw.githubusercontent.com/Untrivial-ai/agent-orchestrator/issue-assets-stale-pr-conflict-2870/.issue-assets/stale-pr-conflict-2870/activity-conflict.png)

## Reproduction / observed state

Live case: [PR #2870](https://github.com/Untrivial-ai/agent-orchestrator/pull/2870), session `agent-orchestrator-31`.

GitHub reports the current PR as mergeable but blocked on review/branch protection, not conflict:

```json
{
  "url": "https://github.com/Untrivial-ai/agent-orchestrator/pull/2870",
  "mergeable": "MERGEABLE",
  "mergeStateStatus": "BLOCKED",
  "reviewDecision": "REVIEW_REQUIRED",
  "headRefOid": "b960b45af86951ff3770a8f51ae4d2fffcd4d22f"
}
```

AO's session API returns both the current PR row and a stale row from the old repo owner:

```json
{
  "prs": [
    {
      "url": "https://github.com/Untrivial-ai/agent-orchestrator/pull/2870",
      "number": 2870,
      "mergeability": "blocked",
      "updatedAt": "2026-08-13T03:08:46.531282Z"
    },
    {
      "url": "https://github.com/AgentWrapper/agent-orchestrator/pull/2870",
      "number": 2870,
[...]

Issue #3252, opened 2026-07-29
bug(scm): PR discovery misses repo transfers and Codex sessions bypass gh wrapper


AgentWrapper:
## Bug
A Codex worker can create a PR successfully, but AO does not attach or enrich that PR in the session sidebar when the project's registered GitHub origin is stale after an org/repo transfer and the session resolves the real `gh` binary instead of AO's wrapper.

Observed locally on 2026-07-28: PR [#3250](https://github.com/Untrivial-ai/agent-orchestrator/pull/3250) existed and CI passed, but `GET /api/v1/sessions/agent-orchestrator-68/pr` returned `[]` and the session list had no `prs` entry until `ao session claim-pr agent-orchestrator-68 3250` was run manually.

**Source:** local AO session report/debugging
**Reported by:** @AgentWrapper
**Analyzed against:** `8771cba75a45338a570a45b4dce1c9a4f583be85` locally; checked current `origin/main` `218c39199` for the same `HookPATH` behavior
**AO version:** `/tmp/ao version` -> `dev`
**Daemon verified:** `/tmp/ao status --json` reported port `3001`, ready/healthy
**Confidence:** High
**Priority:** P1 - core PR/review surface is missing, workaround exists (`ao session claim-pr`) but users will not know to run it

## Reproduction
1. Have an AO project registered with an old GitHub origin, e.g. `https://github.com/AgentWrapper/agent-orchestrator.git`.
2. The repository has moved/redirected to `https://github.com/Untrivial-ai/agent-orchestrator.git`.
3. Spawn a Codex worker whose branch is `ao/<session>/root` and create a child PR branch, e.g. `ao/<session>/fix`.
4. Create a PR in the moved repo with `gh pr create` from inside the session.
5. Open the session inspector/sidebar.

## Expected
AO should attach and enrich the PR under the owning session automatically. The sidebar should show the PR number and SCM state, and `GET /api/v1/sessions/{id}/pr` should return the enriched PR summary.

## Actual
The session appears to have no PR. In the local case:

- `GET /api/v1/sessions/agent-orchestrator-68/pr` returned `[]` before manual claim.
- `ao session get agent-orchestrator-68 --json` showed no PR facts and status remained `working`.
- `which -a gh` inside the session resolved `/opt/homebrew/bin/gh`, not `/Users/itrytoohard/.ao/bin/gh`.
[...]

PR #4090, opened 2026-08-18
fix(scm): single-source PR identity — end the merge-readiness flap after repo transfers

### LINKED ISSUE #4089: SCM observer drops fetched PR observations after repo rename/transfer — PR panel stuck at 'Checking merge readiness'
## Bug
The session PR panel shows a mostly-stuck **"Checking merge readiness"** ("Waiting for the latest checks and review state.") even when GitHub has a fully resolved answer. Review state stays current, but CI and mergeability never resolve. Reported across MANY agent sessions — systemic on any install whose stored PR rows carry a pre-transfer repo name.

**Source:** maintainer chat report (triage session agent-orchestrator-139)
**Reported by:** @AgentWrapper
**Analyzed against:** `d380279b5` (origin/main, 2026-08-18)
**AO version:** packaged desktop daemon (`ao version` → `dev`), port 3001
**Confidence:** High — exact code path traced; pre-fix regression test reproduces the production fingerprint byte-for-byte

## Corrected symptom + second mechanism (added 2026-08-19)

Maintainer follow-up: the state is **not permanently stuck — it flaps**. Correct merge status appears intermittently, then reverts to "Checking merge readiness", and severity increased after recent updates. A live DB watch (5s sampling on the affected machine) captured one full cycle for PRs #4069/#2870:

```
15:47:56  Untrivial-ai/...  passing/blocked   title set, hashes set        (heal)
15:48:27  AgentWrapper/...  unknown/unknown   title gone, hashes WIPED     (clobber)
15:49:02  Untrivial-ai/...  passing/blocked   title set, hashes set        (heal)
```

**Mechanism 2 — discovery baseline clobber.** The project checkout carries a stale `upstream` remote (`AgentWrapper/agent-orchestrator`), so `resolveScanRepos` scans **two identities for one repository**. Whichever identity the row does *not* currently carry finds the PR "not in subjects" during `discoverNewPRs` and persists a baseline row — and `UpsertPR` (`queries/pr.sql`) overwrites `ci_state`, `mergeability`, `repo`, `title`, and the semantic hashes unconditionally. Healing requires a same- or later-poll detail fetch whose observation key matches a live subject.

The two mechanisms compose into the observed behavior:
- **Pre-#3923**: `obs.Repo` echoed the requesting ref, so every fetch healed under either identity — the clobber existed but the unknown window was sub-poll and invisible (its visible artifact was #3922's duplicate rows instead).
- **Post-#3923**: the stale-identity heal path is dead (Mechanism 1, the key-mismatch drop), so after an `AgentWrapper`-side clobber the row stays `unknown` until the `Untrivial-ai` listing happens to re-discover it — windows of many polls, matching "stuck most of the time".

**Fix implications:** [PR #4090](https://github.com/Untrivial-ai/agent-orchestrator/pull/4090)'s minimal seam fix makes every tracked-subject fetch heal again, shrinking the visible unknown window to at most one poll — but the clobber/repo ping-pong churn (hash wipes forcing refetches and lifecycle redelivery) remains. Only the ProviderID-primary identity cleanup (see architectural scope below) removes it: discovery dedupes listing entries against tracked subjects by provider identity, so a second scan name for the same repo never re-baselines a tracked row.

**Immediate workaround on affected machines:** remove the stale remote (`git remote remove upstream` in the project checkout) and restart the daemon (the scan set is memoized per process).

## Concrete case (verified live)
Session `agent-orchestrator-138`, PR [#4069](https://github.com/Untrivial-ai/agent-orchestrator/pull/4069). GitHub GraphQL (replayed with the daemon's exact batch query, which succeeds via the rename redirect): `state=OPEN, mergeable=MERGEABLE, mergeStateStatus=BLOCKED, reviewDecision=REVIEW_REQUIRED, statusCheckRollup=SUCCESS` (single check `scan`, COMPLETED/SUCCESS).

Meanwhile `GET /api/v1/sessions/agent-orchestrator-138/pr` returns `ci.state="unknown"`, `mergeability.state="unknown"`, `review.decision="review_required"`, `title=""`, `changedFiles=0` — and the daemon's `pr` row shows:

```
repo = AgentWrapper/agent-orchestrator   <-- pre-transfer name
ci_state = unknown, mergeability = unknown
review_decision = review_required        <-- current (review_hash set, real thread IDs)
title = '' , provider_mergeable = '' , metadata_hash = ''
```

**11 tracked PRs** on the machine share this exact fingerprint (`repo=AgentWrapper/…`, empty metadata, unknown CI/mergeability, fresh review), all stamped by the same poll. The frontend is innocent: `prCardPresentation` (frontend/src/renderer/lib/pr-display.ts:230) correctly renders "Checking merge readiness" whenever `ci.state` is `unknown`/`pending` or `mergeability.state` is `unknown` — the daemon just never resolves them.

## Root Cause
Repo identity is baked into the observation dispatch key, and a GitHub repo rename/org transfer makes the two sides of the key disagree forever:

1. `repoForTrackedPR` (backend/internal/observe/scm/observer.go:879-886) derives the tracked subject's repo from the **stored PR row** (`AgentWrapper/agent-orchestrator`), so refs are fetched under the stale name.
2. The batch fetch succeeds through GitHub's rename redirect, but `scmObservationFromGraphQL` (backend/internal/adapters/scm/github/observer_provider.go:475-478) canonicalizes `obs.Repo` from the PR URL to the **new** owner/name (`Untrivial-ai/agent-orchestrator`).
3. `Poll` (backend/internal/observe/scm/observer.go:487) and `reconcileTerminalGitHubPRs` (:1428) key fetched observations with `prKeyFromObs(obs)` — the canonical key — then look up `selection.subjectsByPR[key]`, which only contains the stale key. **The observation is silently dropped every poll**, and `chunkSeen` misses the ref key so the repo is additionally marked refresh-incomplete (ETag/cursor pinned).
4. `refreshReviews` (observer.go:1607-1656) keys by the subject and fetches per-ref directly, so review facts stay fresh — while its `observationFromLocal` fallback persists review-only writes that keep re-asserting the stale repo name and empty metadata. Hence the exact split symptom: review current, CI/mergeability `unknown` forever.

The pre-fix regression test reproduces the production fingerprint exactly: `persisted PR = repo "old/r" title "" ci "unknown" mergeability "unknown"`.

## Reproduction
1. Track an open PR in a session (row in `pr` table with `repo=<old-owner>/<name>`).
2. Transfer/rename the repository on GitHub so the canonical owner changes.
3. Let the observer poll. CI/mergeability stay `unknown` indefinitely; the PR panel shows "Checking merge readiness" while the Review row is correct.

## Fix
PR incoming on branch `ao/agent-orchestrator-139/scm-rename-rekey`: resolve the dispatch key by falling back to the requesting ref (matched by PR URL / URL alias) when the canonical key matches no ref, in both the normal batch loop and terminal reconciliation. Persistence then records the canonical repo on the row (`domainFromObservation` uses `obs.Repo`), so the tracked identity self-heals on the next poll. Focused regression tests included; they fail without the fix with the exact production fingerprint.

## Impact
- Merge-readiness surface permanently broken for every PR tracked before a repo transfer/rename (org migrations are routine).
- CI state, failing-check links, diff stats, and title all stuck empty; lifecycle decisions that depend on CI/mergeability facts never fire.
- The repo is marked refresh-incomplete each poll, pinning list ETags/cursors.

## Related
- [#3252](https://github.com/Untrivial-ai/agent-orchestrator/issues/3252) — covers the *discovery/attribution* facet after a transfer (new PRs not attached; stale project origin; gh wrapper PATH). This issue is the complementary *enrichment* facet: PRs that ARE tracked never get CI/mergeability. The observation-drop also explains why re-discovery could not repair already-tracked rows on the affected machine.

## Architectural scope: stable PR identity (added 2026-08-18)

Maintainer review identified this bug as a regression from [PR #3923](https://github.com/Untrivial-ai/agent-orchestrator/pull/3923) and the fourth member of a rename-bug cluster ([#2509](https://github.com/Untrivial-ai/agent-orchestrator/issues/2509), [#3922](https://github.com/Untrivial-ai/agent-orchestrator/issues/3922), [#3252](https://github.com/Untrivial-ai/agent-orchestrator/issues/3252)). The structural cause: PR identity is derived independently in five layers (stored row `repo`/`url`, observer subject key, observation dispatch key, repo scan set, store `provider_id`/alias machinery), and a repo rename/transfer makes them disagree. The stable identity — the provider node ID (`PR_kwDO…`, GitLab MR global ID) — has been captured on every observation and enforced unique in storage since #3923's migration 0097, but the observer never dispatches on it.

**Expanded scope for this issue:** [PR #4090](https://github.com/Untrivial-ai/agent-orchestrator/pull/4090) will grow from the minimal seam patch into the ProviderID-primary identity cleanup, pending approval of the architecture doc (`docs/scm-observer.md`, added in that PR):

- one identity helper owning key construction: `provider:host@provider_id` when known, legacy `provider:host:repo#number` as a self-retiring fallback for rows without a provider ID;
- subjects, observation dispatch, review refresh, and per-PR caches keyed by identity; repo-level guards/ETags/cursors stay name-keyed;
- discovery dedupes subjects by identity, closing the two-scan-names→two-rows class (#3922) at the source;
- rows self-heal to canonical repo/URL on the next successful fetch — no schema change (0097 suffices), no data migration;
- a cross-layer transfer scenario test suite covering discovery, batch fetch, reconciliation, review refresh, and legacy-row fallback.

Out of scope (remain on #3252): project-origin refresh, gh-wrapper PATH precedence, doctor checks.

AgentWrapper:
Fixes #4089

## Summary
The "Checking merge readiness" flap traced to one structural cause: **PR identity was derived independently in five layers** (stored row, subject key, observation dispatch key, repo scan set, store `provider_id`/alias machinery), and a repo rename/transfer makes them disagree. This PR gives identity one owner and lands in three stages (see the "Simplification pass" note below):

1. **`fix(scm)` — the dispatch seam**: canonical observations for renamed repos are delivered to their tracked subjects instead of being dropped every poll.
2. **`refactor(scm)` — single-source identity**: a PR's identity is **`(provider, host, provider_id)`** — the rename-proof node id both adapters already stamp and migration 0097 already enforces unique. Repo name, number, and URL are mutable display coordinates.
   - `identityPRKey` / `keyForTrackedPR` own key construction; the legacy `repo#number` key survives only as a **self-retiring fallback** for rows that predate provider ids (their first successful fetch stamps one).
   - Subjects, per-PR caches, and review refresh are identity-keyed. Repo-level list ETags/sync cursors/commit guards deliberately stay name-keyed — listing genuinely is a per-name operation.
   - **Discovery dedupes by identity**: a listing served under a second scan name (stale `upstream` remote after a transfer) recognizes tracked PRs by node id and never re-baselines them — the fact-wiping clobber (#4089 mechanism 2, captured live flapping every ~45s) becomes structurally impossible, not merely quickly repaired. This prevents a second scan name from creating a new duplicate row for an already tracked provider-native PR (#3922's discovery artifact); existing legacy id-less duplicates remain handled by the store's identity/alias machinery.
   - Rows self-heal to the canonical repo/URL on their next fetch. **No schema change, no data migration** — 0097 suffices. In-memory caches only change key shape.
3. **`refactor(scm)` — positional fetch contract** (the simplification pass): `FetchPullRequests` results are **positionally aligned with the requested refs** (`result[i]` answers `refs[i]`; a PR the batch query cannot resolve leaves a `Fetched=false` placeholder at its index — documented on the port). The observer records which subject each ref was issued for and attributes `batch[i]` directly, so the whole observation-matching layer (`observationSubjectKey`, `prKeyFromObs`, the `keyByRef` bridge, `chunkSeen`/`reconcileSeen` bookkeeping, and the interim URL/alias fallback from stage 1) is deleted outright. GitLab and the multi dispatcher already honored alignment; GitHub was the lone outlier. Legacy id-less rows are covered by the same mechanism — no fallback needed. Net, `observer.go` is smaller than `origin/main` while carrying the rename fix and the clobber kill.
4. **`docs`** — `docs/scm-observer.md`: the SCM architecture doc — pipeline, durable-state invariants, the five-way identity split, the live flap capture, and this design.

## Evidence
- Live daemon returned `ci=unknown, mergeability=unknown` while GitHub said `MERGEABLE/BLOCKED` with green checks; 11 PR rows shared the stale-repo fingerprint.
- A 5s DB watch captured the flap: healthy → clobbered (`AgentWrapper/…`, facts + hashes wiped by discovery baseline) → healed (canonical fetch), cycling every ~45s. Details in #4089.

## Tests
- `TestPoll_IdentityKeyedDispatchSurvivesRename` — canonical (renamed-repo) observation reaches its tracked subject; the fixture's stored URL deliberately differs from the canonical URL so no content-based join could explain the delivery.
- `TestPoll_SecondScanNameDoesNotRebaselineTrackedPR` — the clobber regression; fails pre-change with the exact live fingerprint (`CI:unknown, Repo:old/r`, hashes wiped).
- `TestPoll_RenamedRepoObservationPersistsUnderTrackedRef` / `…ReconciliationPersistsUnderTrackedRef` — renamed-repo observations persist through both the batch and terminal-reconciliation paths (positional attribution; legacy id-less rows covered by the same mechanism).
- `TestKeyForTrackedPR` — key resolution table: identity key when a provider id exists, legacy name key otherwise.
- Full `observe/…`, `adapters/scm/…`, `storage/…`, `lifecycle/…`, `service/…` suites green; `go vet` clean.

## Risks / follow-ups
- Legacy id-less rows register under the name-based key until their first successful fetch stamps a provider id. With only an already-renamed remote, discovery may write one baseline before that fetch; positional attribution still delivers the observation and makes the fallback self-retire.
- Discovery dedupe requires the listing to carry a provider id (both adapters stamp it since #3923); an id-less listing entry against an id-less row still dedupes by name.
- The remaining transfer facets (project-origin refresh, gh-wrapper PATH, doctor checks) stay on #3252.

### DIFF
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -219,14 +219,13 @@ func stableCheckFingerprint(parts []string) string {
 	return hex.EncodeToString(sum[:])
 }
 
-// FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs in
-// one GraphQL request. The observer owns chunking; this method rejects larger
-// batches so tests catch accidental over-batching.
-// FetchPullRequests returns observations positionally aligned with refs:
-// out[i] answers refs[i]. A PR the batch query could not resolve (deleted
-// repo, permissions) leaves a zero-value Fetched=false placeholder at its
-// index so the observer marks that ref refresh-incomplete without guessing
-// which response belonged to which request.
+// FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs
+// in one GraphQL request, positionally aligned with refs: out[i] answers
+// refs[i]. A PR the batch query could not resolve (deleted repo, permissions)
+// leaves a Fetched=false placeholder at its index so the observer marks that
+// ref refresh-incomplete without guessing which response belonged to which
+// request. The observer owns chunking; this method rejects larger batches so
+// tests catch accidental over-batching.
 func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef) ([]ports.SCMObservation, error) {
 	if len(refs) == 0 {
 		return nil, nil
@@ -244,6 +243,7 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 		repoData, _ := data[aliases[i]].(map[string]any)
 		pr, _ := repoData["pullRequest"].(map[string]any)
 		if pr == nil {
+			out[i].Error = ports.ErrSCMNotFound
 			continue
 		}
 		if scmContextsPaginated(pr) {
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -221,11 +221,12 @@ func stableCheckFingerprint(parts []string) string {
 
 // FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs
 // in one GraphQL request, positionally aligned with refs: out[i] answers
-// refs[i]. A PR the batch query could not resolve (deleted repo, permissions)
-// leaves a Fetched=false placeholder at its index so the observer marks that
-// ref refresh-incomplete without guessing which response belonged to which
-// request. The observer owns chunking; this method rejects larger batches so
-// tests catch accidental over-batching.
+// refs[i]. A PR the batch query could not resolve (deleted repo, permissions,
+// an old name recreated so the redirect no longer lands) leaves a
+// Fetched=false placeholder carrying ErrNotFound at its index, so the
+// observer marks that ref refresh-incomplete and can log the permanent miss
+// distinctly from a transient provider failure. The observer owns chunking;
+// this method rejects larger batches so tests catch accidental over-batching.
 func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef) ([]ports.SCMObservation, error) {
 	if len(refs) == 0 {
 		return nil, nil
@@ -243,7 +244,14 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 		repoData, _ := data[aliases[i]].(map[string]any)
 		pr, _ := repoData["pullRequest"].(map[string]any)
 		if pr == nil {
-			out[i].Error = ports.ErrSCMNotFound
+			out[i] = ports.SCMObservation{
+				Fetched:  false,
+				Provider: ref.Repo.Provider,
+				Host:     ref.Repo.Host,
+				Repo:     repoFullName(ref.Repo),
+				PR:       ports.SCMPRObservation{Number: ref.Number, URL: ref.URL},
+				Error:    fmt.Errorf("%w: pull request %s#%d not in batch response", ErrNotFound, repoFullName(ref.Repo), ref.Number),
+			}
 			continue
 		}
 		if scmContextsPaginated(pr) {
--- a/backend/internal/adapters/scm/github/observer_provider.go
+++ b/backend/internal/adapters/scm/github/observer_provider.go
@@ -222,6 +222,11 @@ func stableCheckFingerprint(parts []string) string {
 // FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs in
 // one GraphQL request. The observer owns chunking; this method rejects larger
 // batches so tests catch accidental over-batching.
+// FetchPullRequests returns observations positionally aligned with refs:
+// out[i] answers refs[i]. A PR the batch query could not resolve (deleted
+// repo, permissions) leaves a zero-value Fetched=false placeholder at its
+// index so the observer marks that ref refresh-incomplete without guessing
+// which response belonged to which request.
 func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef) ([]ports.SCMObservation, error) {
 	if len(refs) == 0 {
 		return nil, nil
@@ -234,7 +239,7 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 	if err != nil {
 		return nil, err
 	}
-	out := make([]ports.SCMObservation, 0, len(refs))
+	out := make([]ports.SCMObservation, len(refs))
 	for i, ref := range refs {
 		repoData, _ := data[aliases[i]].(map[string]any)
 		pr, _ := repoData["pullRequest"].(map[string]any)
@@ -246,7 +251,7 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 				return nil, err
 			}
 		}
-		out = append(out, scmObservationFromGraphQL(ref, pr))
+		out[i] = scmObservationFromGraphQL(ref, pr)
 	}
 	return out, nil
 }
--- a/backend/internal/observe/scm/observer_test.go
+++ b/backend/internal/observe/scm/observer_test.go
@@ -3217,30 +3217,3 @@ func TestPoll_IdentityKeyedDispatchSurvivesRename(t *testing.T) {
 		t.Fatalf("lifecycle notifications = %d, want 1", len(lc.observed))
 	}
 }
-
-// Two stored rows for the same provider-native PR (one per repo name, the
-// pre-#3923 duplicate-row artifact) collapse to a single subject instead of
-// two subjects fighting over the same PR.
-func TestDiscoverSubjects_CollapsesDuplicateRowsByIdentity(t *testing.T) {
-	store := &fakeStore{
-		sessions: []domain.SessionRecord{{ID: "p-1", ProjectID: "p", Metadata: domain.SessionMetadata{Branch: "feat"}}},
-		projects: map[string]domain.ProjectRecord{"p": {ID: "p", RepoOriginURL: "https://github.com/new/r.git"}},
-		prs: map[domain.SessionID][]domain.PullRequest{"p-1": {
-			{URL: "https://github.com/old/r/pull/1", SessionID: "p-1", Number: 1, Provider: "github", Host: "github.com", Repo: "old/r", ProviderID: "PR_dup", SourceBranch: "feat"},
-			{URL: "https://github.com/new/r/pull/1", SessionID: "p-1", Number: 1, Provider: "github", Host: "github.com", Repo: "new/r", ProviderID: "PR_dup", SourceBranch: "feat"},
-		}},
-		checks: map[string][]domain.PullRequestCheck{},
-	}
-	provider := &fakeProvider{observations: map[string]ports.SCMObservation{}}
-	obs := newTestObserver(store, provider, &fakeLifecycle{}, time.Unix(1, 0).UTC())
-	subjects, _, err := obs.discoverSubjects(context.Background())
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(subjects) != 1 {
-		t.Fatalf("subjects = %d, want 1 (same provider identity must collapse)", len(subjects))
-	}
-	if _, ok := subjects[identityPRKey("github", "github.com", "PR_dup")]; !ok {
-		t.Fatalf("subject not keyed by provider identity: %v", subjects)
-	}
-}
--- a/backend/internal/observe/scm/observer_test.go
+++ b/backend/internal/observe/scm/observer_test.go
@@ -209,6 +209,7 @@ func (p *fakeProvider) ListPRsByRepo(_ context.Context, repo ports.SCMRepo, _ ti
 func (p *fakeProvider) CommitChecksGuard(_ context.Context, repo ports.SCMRepo, sha, _ string) (ports.SCMGuardResult, error) {
 	return p.checkGuards[commitKey(repo, sha)], nil
 }
+
 // FetchPullRequests honors the positional-alignment contract: out[i] answers
 // refs[i], with a zero-value Fetched=false placeholder for refs the fake has
 // no observation for.
--- a/backend/internal/observe/scm/observer_test.go
+++ b/backend/internal/observe/scm/observer_test.go
@@ -209,6 +209,9 @@ func (p *fakeProvider) ListPRsByRepo(_ context.Context, repo ports.SCMRepo, _ ti
 func (p *fakeProvider) CommitChecksGuard(_ context.Context, repo ports.SCMRepo, sha, _ string) (ports.SCMGuardResult, error) {
 	return p.checkGuards[commitKey(repo, sha)], nil
 }
+// FetchPullRequests honors the positional-alignment contract: out[i] answers
+// refs[i], with a zero-value Fetched=false placeholder for refs the fake has
+// no observation for.
 func (p *fakeProvider) FetchPullRequests(_ context.Context, refs []ports.SCMPRRef) ([]ports.SCMObservation, error) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
@@ -216,25 +219,25 @@ func (p *fakeProvider) FetchPullRequests(_ context.Context, refs []ports.SCMPRRe
 	if p.fetchErr != nil {
 		return nil, p.fetchErr
 	}
-	out := make([]ports.SCMObservation, 0, len(refs))
-	for _, ref := range refs {
+	out := make([]ports.SCMObservation, len(refs))
+	for i, ref := range refs {
 		key := prKey(ref.Repo, ref.Number)
 		// Per-observation error mimics the multi dispatcher's per-provider
 		// Error metadata: a Fetched=false placeholder carrying the failure
 		// so the observer can route it to cooldown / refresh-incomplete.
 		if err, ok := p.fetchObsErrors[key]; ok {
-			out = append(out, ports.SCMObservation{
+			out[i] = ports.SCMObservation{
 				Fetched:  false,
 				Provider: ref.Repo.Provider,
 				Host:     ref.Repo.Host,
 				Repo:     ref.Repo.Repo,
 				PR:       ports.SCMPRObservation{Number: ref.Number, URL: ref.URL},
 				Error:    err,
-			})
+			}
 			continue
 		}
 		if obs, ok := p.observations[key]; ok {
-			out = append(out, obs)
+			out[i] = obs
 		}
 	}
 	return out, nil
@@ -3087,77 +3090,6 @@ func TestPoll_RenamedRepoReconciliationPersistsUnderTrackedRef(t *testing.T) {
 	final := store.writes[len(store.writes)-1].pr
 	if final.Repo != "new/r" || final.CI != domain.CIPassing {
 		t.Fatalf("persisted PR = repo %q ci %q; want canonical new/r with passing CI", final.Repo, final.CI)
-	}
-}
-
-func TestObservationSubjectKey(t *testing.T) {
-	newObs := func(url, alias, providerID string) ports.SCMObservation {
-		return ports.SCMObservation{
-			Fetched: true, Provider: "github", Host: "github.com", Repo: "new/r",
-			PR: ports.SCMPRObservation{URL: url, URLAlias: alias, Number: 1, ProviderID: providerID},
-		}
-	}
-	oldRepo := ports.SCMRepo{Provider: "github", Host: "github.com", Owner: "old", Name: "r", Repo: "old/r"}
-	newRepo := ports.SCMRepo{Provider: "github", Host: "github.com", Owner: "new", Name: "r", Repo: "new/r"}
-	oldRef := ports.SCMPRRef{Repo: oldRepo, Number: 1, URL: "https://github.com/new/r/pull/1"}
-	newRef := ports.SCMPRRef{Repo: newRepo, Number: 1, URL: "https://github.com/new/r/pull/1"}
-	idKey := identityPRKey("github", "github.com", "PR_1")
-	subj := &subject{}
-	tests := []struct {
-		name     string
-		obs      ports.SCMObservation
-		refs     []ports.SCMPRRef
-		subjects map[string]*subject
-		keyByRef map[string]string
-		want     string
-	}{
-		{
-			name:     "provider identity wins over every name",
-			obs:      newObs("https://github.com/new/r/pull/1", "", "PR_1"),
-			refs:     []ports.SCMPRRef{oldRef},
-			subjects: map[string]*subject{idKey: subj},
-			keyByRef: map[string]string{prKey(oldRepo, 1): idKey},
-			want:     idKey,
-		},
-		{
-			name:     "canonical name key when subject has no identity",
-			obs:      newObs("https://github.com/new/r/pull/1", "", "PR_1"),
-			refs:     []ports.SCMPRRef{newRef},
-			subjects: map[string]*subject{prKey(newRepo, 1): subj},
-			keyByRef: map[string]string{prKey(newRepo, 1): prKey(newRepo, 1)},
-			want:     prKey(newRepo, 1),
-		},
-		{
-			name:     "renamed repo falls back to requesting ref by URL",
-			obs:      newObs("https://github.com/new/r/pull/1", "", ""),
-			refs:     []ports.SCMPRRef{oldRef},
-			subjects: map[string]*subject{prKey(oldRepo, 1): subj},
-			keyByRef: map[string]string{prKey(oldRepo, 1): prKey(oldRepo, 1)},
-			want:     prKey(oldRepo, 1),
-		},
-		{
-			name:     "renamed repo falls back via URL alias",
-			obs:      newObs("https://github.com/new/r/pull/1", "https://github.com/old/r/pull/1", ""),
-			refs:     []ports.SCMPRRef{{Repo: oldRepo, Number: 1, URL: "https://github.com/old/r/pull/1"}},
-			subjects: map[string]*subject{prKey(oldRepo, 1): subj},
-			keyByRef: map[string]string{prKey(oldRepo, 1): prKey(oldRepo, 1)},
-			want:     prKey(oldRepo, 1),
-		},
-		{
-			name:     "no matching ref keeps canonical key",
-			obs:      newObs("https://github.com/new/r/pull/1", "", ""),
-			refs:     []ports.SCMPRRef{{Repo: oldRepo, Number: 2, URL: "https://github.com/new/r/pull/2"}},
-			subjects: map[string]*subject{},
-			keyByRef: map[string]string{},
-			want:     "github:github.com:new/r#1",
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			if got := observationSubjectKey(tt.obs, tt.refs, tt.subjects, tt.keyByRef); got != tt.want {
-				t.Fatalf("observationSubjectKey = %q, want %q", got, tt.want)
-			}
-		})
 	}
 }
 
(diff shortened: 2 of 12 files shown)


AgentWrapper:
Root-cause addendum for reviewers: this is a regression from #3923 (see https://github.com/Untrivial-ai/agent-orchestrator/issues/4089#issuecomment — regression-source comment). This PR keeps #3923's URL-based canonicalization and store identity collapse intact; it only fixes the observer dispatch seam those changes left unkeyed.

AgentWrapper:
Scope correction from live observation (details on [#4089](https://github.com/Untrivial-ai/agent-orchestrator/issues/4089), 'Corrected symptom + second mechanism'): the state flaps rather than sticking. A second mechanism composes with the dispatch drop — a stale second scan identity makes `discoverNewPRs` re-baseline tracked rows (`UpsertPR` overwrites CI/mergeability/title/hashes unconditionally). This PR's seam fix shrinks the visible unknown window to at most one poll (every tracked-subject fetch heals again) but does not stop the re-baseline churn; that is removed by the identity-primary discovery dedupe in the pending architectural cleanup (docs/scm-observer.md, 'The discovery clobber').

AgentWrapper:
**Simplification pass** (maintainer feedback: the cleanup should remove code, not add it): the dispatch-matching layer is now deleted outright. The root observation — GitLab and the multi dispatcher already return `FetchPullRequests` results positionally aligned with the requested refs; GitHub was the lone outlier that skipped missing PRs, and everything the observer grew to compensate (`observationSubjectKey`, `prKeyFromObs`, the `keyByRef` bridge, `chunkSeen`/`reconcileSeen`, the URL/alias fallback) existed to re-derive a correspondence the requester had and the response format threw away.

This commit makes alignment the contract (`result[i]` answers `refs[i]`; documented on the port), fixes GitHub to honor it (also fixing a latent index-shift in the multi dispatcher's padding), and attributes `batch[i]` directly to the subject its ref was issued for. Legacy id-less rows are covered by the same mechanism — no fallback needed. Identity keys stay for what they genuinely own: subject registration and discovery dedupe (the clobber kill).

**Line accounting:** this pass is −186 net (119+/305−). `backend/internal/observe/scm/observer.go` is now **net smaller than origin/main** (139+/159−) while carrying the rename fix and the clobber kill; the PR's remaining additions are tests (+~230) and `docs/scm-observer.md` (+330).

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -234,7 +239,7 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 	if err != nil {
 		return nil, err
 	}
-	out := make([]ports.SCMObservation, 0, len(refs))
+	out := make([]ports.SCMObservation, len(refs))
```
whoisasx, line 242: **Blocking — behaviour regression outside the observer.**

Right change for the observer, but it also changes what single-ref callers see. `session.fetchClaimObservation` (`backend/internal/service/session/claim_pr.go:149-155`) separates no-such-PR from fetch-failed purely by batch length:

```go
if len(batch) == 0 { return ports.SCMObservation{}, ErrPRNotFound }
obs := batch[0]
if !obs.Fetched { return ports.SCMObservation{}, ErrSCMUnavailable }
```

A PR number that does not exist in an existing repo comes back from GraphQL as `data.<alias>.pullRequest = null` with no `errors` array, so `doGraphQL` succeeds and line 247 continues past it. Previously that produced `len(batch) == 0`, mapping to `ErrPRNotFound` and HTTP 404 `PR_NOT_FOUND`. Now it produces one zero-value `Fetched=false` entry, mapping to `ErrSCMUnavailable` and HTTP 503 `SCM_UNAVAILABLE` (`httpd/controllers/sessions.go:1584`) — telling the user to retry something that can never succeed.

Nothing catches this: both session-service doubles (`service_test.go:2545` and `:2560`) still return `nil, nil` for the no-PR case, i.e. they encode the old compaction.

Suggested fix: set `Error: ports.ErrSCMNotFound` on the nil-`pr` placeholder so callers can tell a permanent miss from a transient one. The observer is unaffected by that — it routes `obs.Error` through `rateLimitCooldown`, which classifies it as non-rate-limit and marks the repo refresh-incomplete, same as the zero-value placeholder does today. Then update `fetchClaimObservation` and the two doubles to match.

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -222,6 +222,11 @@ func stableCheckFingerprint(parts []string) string {
 // FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs in
 // one GraphQL request. The observer owns chunking; this method rejects larger
 // batches so tests catch accidental over-batching.
+// FetchPullRequests returns observations positionally aligned with refs:
```
whoisasx: Nit: this leaves two doc-comment paragraphs that both open with the function name, which reads as a duplicated comment. Fold the positional contract into the existing block:

```go
// FetchPullRequests fetches normalized PR/check metadata for up to 25 PR refs
// in one GraphQL request, positionally aligned with refs: out[i] answers
// refs[i]. A PR the batch query could not resolve (deleted repo, permissions)
// leaves a zero-value Fetched=false placeholder at its index so the observer
// marks that ref refresh-incomplete without guessing which response belonged
// to which request. The observer owns chunking; this method rejects larger
// batches so tests catch accidental over-batching.
```

backend/internal/observe/scm/observer_test.go
```diff
@@ -3006,3 +3010,237 @@ func TestPoll_RepoRefreshMonotonicity_ListingFailsButPRSucceeds(t *testing.T) {
...
+	}
+	if len(lc.observed) != 1 {
+		t.Fatalf("lifecycle notifications = %d, want 1", len(lc.observed))
+	}
+}
+
+// Two stored rows for the same provider-native PR (one per repo name, the
+// pre-#3923 duplicate-row artifact) collapse to a single subject instead of
+// two subjects fighting over the same PR.
+func TestDiscoverSubjects_CollapsesDuplicateRowsByIdentity(t *testing.T) {
```
whoisasx: This fixture builds two rows sharing `Provider` github, `Host` github.com and `ProviderID` PR_dup, which `0097_pr_provider_identity.sql` makes impossible in a real DB: it creates a unique index on `pr(provider, host, provider_id)` partial to rows whose `provider_id` is non-empty. The fake store has no such constraint, so the test passes while pinning a state that cannot occur.

The duplicate shapes that do occur (the #3922 artifact) are: both rows id-less (pre-#3923), or one stamped and one still empty. Neither collapses here, because `keyForTrackedPR` falls back to `prKey(repo, pr.Number)` when `ProviderID` is empty and the two rows carry different repo names, so they land under two distinct keys and stay two subjects.

Two options: change the fixture to the mixed case (one row with ProviderID PR_dup, one with empty, same number) and make `discoverSubjects` fold the id-less row into the identity-keyed subject; or keep the current behaviour and narrow the claim in the PR body and `docs/scm-observer.md:285-288` to preventing new duplicates at discovery, leaving pre-existing duplicate rows to the #3923 store-level collapse.

whoisasx (commented):
## Review: changes requested

The identity model is the right call: (provider, host, provider_id) is the only coordinate that survives a transfer, positional attribution beats re-deriving a key from response content, and net-deleting the matching layer while fixing the bug is a good outcome. I verified that the alignment contract holds across all three production `FetchPullRequests` implementations (github now, gitlab already allocated `results` at `len(refs)`, multi maps back through `indexedRef.idx`), that `refs` and `refKeys` are only ever appended together, and that `Cache` is process-local so the key-shape change orphans nothing durable. Incidental catch worth noting: the old `active := chunk[:0]` aliased and wrote through `selection.refs`; the new allocation removes that.

Four issues, one blocking.

---

### 1. (blocking) GitHub `FetchPullRequests` no longer compacts, and `ClaimPR` depended on that

`out := make([]ports.SCMObservation, len(refs))` means a PR the query cannot resolve now occupies its index as a zero-value `Fetched=false` observation instead of vanishing from the slice. That is correct for the observer, but `session.fetchClaimObservation` (`backend/internal/service/session/claim_pr.go:149-155`) separates no-such-PR from fetch-failed using batch length:

```go
if len(batch) == 0 { return ports.SCMObservation{}, ErrPRNotFound }
obs := batch[0]
if !obs.Fetched { return ports.SCMObservation{}, ErrSCMUnavailable }
```

For a PR number that does not exist in a repo that does, GitHub GraphQL returns `data.<alias>.pullRequest = null` with no `errors` array, so `doGraphQL` succeeds (the branch at `client.go:313` is not entered) and `pr == nil` hits the `continue` at `observer_provider.go:247`.

- Before: `len(batch) == 0` maps to `ErrPRNotFound`, HTTP 404 `PR_NOT_FOUND` (`httpd/controllers/sessions.go:1572`).
- After: `len(batch) == 1` with `Fetched == false` maps to `ErrSCMUnavailable`, HTTP 503 `SCM_UNAVAILABLE` (`sessions.go:1584`).

So `ao pr claim <nonexistent-number>` now reports that the SCM is unavailable and invites a retry that can never succeed, rather than reporting Unknown PR.

This fails silently. Nothing covers the empty-batch branch: `service_test.go:2619` exercises the `ports.ErrSCMNotFound` error path, and both session-service doubles (`noopSCMProvider.FetchPullRequests` at `service_test.go:2545`, `fakeSCM.FetchPullRequests` at `:2560`) still return `nil, nil` for the no-PR case, encoding the old compaction. `pr.ActionService.fetchMergeReadiness` is unaffected because its `!observations[0].Fetched` check already maps to `ErrPRNotFound`.

### 2. `cannedSCMProvider` in the integration suite still violates the new contract

`backend/internal/integration/scm_observer_test.go:120-130` was not updated:

```go
out := make([]ports.SCMObservation, 0, len(refs))
for _, ref := range refs {
    if obs, ok := p.observations[ref.Number]; ok {
        out = append(out, obs)
    }
}
```

Append-and-skip: the moment one ref in a multi-ref batch has no canned observation, every later result shifts left and `out[i]` answers `refs[j]`. Under positional attribution the observer then persists one PR facts onto another subject row, which is precisely the misattribution class this PR exists to eliminate. It happens not to trip today only because the existing cases (numbers 201/202 and 301/302) always populate every tracked number. This PR makes alignment load-bearing, so the double should mirror the `fakeProvider` fix in `observe/scm/observer_test.go:220-243`: index-assign into a `len(refs)` slice and leave a zero-value placeholder for misses.

### 3. Dead references to the deleted `prKeyFromObs`

`prKeyFromObs` is gone, but three places still explain themselves in terms of it:

- `backend/internal/adapters/scm/multi/provider.go:280-284` — the `repoFullNameFromRef` doc says placeholders must carry a non-empty `Repo` so that `prKeyFromObs` produces a non-empty key. That requirement no longer exists; the field now only feeds log context and `firstNonEmpty(batch[i].Provider, ...)`.
- `backend/internal/adapters/scm/multi/provider_test.go:565-573` — the rationale cites the observer skipping placeholders at a `key == empty` guard that no longer exists.
- `backend/internal/adapters/scm/multi/provider_test.go:635-659` — `TestFetchPullRequests_FailedPlaceholderProducesNonEmptyKey` re-implements the deleted key builder inline and asserts on it, so it now guards nothing. Delete it, or restate it as: placeholders carry provider/host/repo for error routing and logging.

For a PR whose headline is that the matching layer is deleted outright, leaving the justification for its scaffolding behind is worth cleaning up in the same change.

### 4. The duplicate-row collapse is asserted against a state the schema forbids

Detail inline on `TestDiscoverSubjects_CollapsesDuplicateRowsByIdentity`. Short version: `0097_pr_provider_identity.sql` creates a unique index on `pr(provider, host, provider_id)` partial to rows whose `provider_id` is non-empty, so the two rows that test builds cannot coexist in a real DB. The duplicate shapes that actually occur (both rows id-less, or one stamped and one not) do **not** collapse, because `keyForTrackedPR` falls back to `prKey(repo, pr.Number)` for the id-less row and the two rows carry different repo names.

The discovery-side dedupe genuinely does stop new duplicates from being created, which is the valuable part. But the PR body claim that duplicate rows for one provider-native PR collapse to a single subject at discovery, and `docs/scm-observer.md:285-288` (kills the two-scan-repos-create-two-subjects class at the source instead of collapsing it later in the store), overstate it for rows that already exist. Either narrow the claim to newly-discovered PRs, or make `discoverSubjects` fold an id-less row into an identity-keyed subject for the same repo and number.

---

### Non-blocking

- `observer_provider.go:222-229`: two doc-comment paragraphs now both open with the function name, which reads as a duplicated comment. See inline.
- `observer.go:797`: with identity keys the duplicate-subject warning now also fires for two rows in the same session, logging `kept_session` and `skipped_session` with the same id. The message reads as a cross-session ownership conflict; worth rewording for the intra-session case.
- The legacy-row protection added in b47f2f1d only engages when both the old and the new repo name are scanned, since the pre-pass needs an exact `prKey(repo, number)` listing hit. With a single, already-renamed remote a legacy id-less row still takes one baseline write before its fetch stamps the id and it self-retires. That matches the documented self-retiring risk and is not a regression; noting it only so the residual one-poll churn is a known quantity.

Everything else in the observer rewrite checked out: reconciliation keying, `prRefreshOK` initialization ordering after `reconcileTerminalGitHubPRs` mutates `candidateKeys`, the dual-key `listedPRs` check at `observer.go:1410`, and `refreshReviews` sharing `keyForTrackedPR` with the dispatch map.

whoisasx:
Addressed review 4981295594 in `1dd34b96b`:

- Preserved positional GitHub results while marking unresolved PRs with `ports.ErrSCMNotFound`; `fetchClaimObservation` now maps that placeholder back to `ErrPRNotFound`/404, with provider and session regression coverage.
- Updated the integration SCM test double to return a result at every requested index.
- Removed the dead `prKeyFromObs` rationale and obsolete key-construction test while retaining the useful placeholder context assertions.
- Removed the schema-impossible duplicate-row fixture and narrowed the implementation docs and PR description to the behavior actually guaranteed: preventing new duplicate discovery rows; legacy id-less duplicates remain store-level cleanup.
- Folded the duplicated GitHub method comment and made the duplicate-subject log wording neutral.

Verification: focused SCM/session/integration suites pass; `go vet ./...` passes. A parallel full `go test ./...` run hit three unrelated timing-sensitive agent tests, all of which passed immediately when rerun serially.

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -234,19 +238,20 @@ func (p *Provider) FetchPullRequests(ctx context.Context, refs []ports.SCMPRRef)
 	if err != nil {
 		return nil, err
 	}
-	out := make([]ports.SCMObservation, 0, len(refs))
+	out := make([]ports.SCMObservation, len(refs))
 	for i, ref := range refs {
 		repoData, _ := data[aliases[i]].(map[string]any)
 		pr, _ := repoData["pullRequest"].(map[string]any)
 		if pr == nil {
+			out[i].Error = ports.ErrSCMNotFound
```
whoisasx: Nit, non-blocking. Correct fix, but it changes the observer log surface as a side effect. `ErrSCMNotFound` is not a `rateLimitedError`, so in both fetch loops it lands in the else branch:

```go
o.logger.Warn("scm observer: provider fetch failed (per-observation); marking refresh-incomplete", ...)
```

Before this commit the nil-`pullRequest` slot carried no error and logged nothing; before the PR it was absent from the batch entirely and also logged nothing. So a tracked PR that is permanently unresolvable (repo deleted, access revoked, or an old name recreated so the redirect no longer lands) now emits one warning per ref per tick. At the 30s cadence that is roughly 2.9k lines a day per broken PR, all describing a permanent miss as a fetch failure.

Surfacing this at all is an improvement over silence, so I would not revert it — just give it its own branch, e.g. check `errors.Is(fetchErr, ports.ErrSCMNotFound)` ahead of the rate-limit test and log at Debug with a message like `tracked PR not found at provider; marking refresh-incomplete`. Keeps the permanent case distinguishable from a transient provider failure in the daemon log.

whoisasx (commented):
## Re-review at `1dd34b9`: approving

All four findings from the previous review are resolved, and the fixes are the right shape rather than the minimum that silences the complaint.

### Verified fixed

**1 (was blocking) — not-found semantics.** `observer_provider.go:246` now stamps `out[i].Error = ports.ErrSCMNotFound` on the nil-`pullRequest` slot, and `fetchClaimObservation` (`claim_pr.go:153`) checks it *before* the `!obs.Fetched` fallthrough, so a missing PR maps back to `ErrPRNotFound` / HTTP 404 instead of 503. Classifying at the placeholder rather than special-casing the caller is the better of the two options I suggested: `pr.ActionService` and any future single-ref caller inherit it. Confirmed the observer is unaffected — `ErrSCMNotFound` is not a `rateLimitedError`, so `rateLimitCooldown` returns false and the ref takes the existing refresh-incomplete path with no spurious cooldown. `multi` passes the per-observation `Error` through untouched when the sub-provider returns no top-level error, so the classification survives the dispatcher.

Good call updating both session-service doubles (`service_test.go:2545`, `:2561`) — they were the reason this could regress silently — and adding the `not found placeholder` case alongside the existing `not found error` case. `TestFetchPullRequestsMarksMissingPRNotFound` pins it at the adapter with a mixed batch (`pr0` null, `pr1` populated), which also re-asserts alignment on the same call. That is the test I would have asked for.

**2 — `cannedSCMProvider`.** `integration/scm_observer_test.go:120-130` now index-assigns into a `len(refs)` slice; misses leave a zero-value `Fetched=false` placeholder the observer already rejects. The integration double no longer contradicts the contract it exercises.

**3 — `prKeyFromObs` residue.** All three sites cleaned: the `repoFullNameFromRef` doc now states the actual reason (placeholders retain ref repository context), the surviving multi test rationale is restated, and `TestFetchPullRequests_FailedPlaceholderProducesNonEmptyKey` is deleted. Checked that `fmt` is still used at `provider_test.go:648`, so no unused-import breakage from the deletion.

**4 — duplicate-row claim.** `TestDiscoverSubjects_CollapsesDuplicateRowsByIdentity` is gone, and both `docs/scm-observer.md:283-287` and the PR summary now scope the guarantee to preventing a second scan name from creating a *new* duplicate row, leaving pre-existing legacy id-less duplicates to the store-level identity/alias machinery. That matches what the code actually does under the 0097 partial unique index.

The three non-blocking notes were picked up too: the doc comment is merged into one block, the discovery warning is reworded to `duplicate tracked PR subject skipped`, and the legacy single-remote one-poll baseline window is now written down under Risks rather than being an unstated edge.

### One nit, not worth blocking on

See inline on `observer_provider.go:246`: attaching an error to the placeholder means a permanently unresolvable tracked PR now emits a `provider fetch failed (per-observation)` warning on every tick, where it used to be silent. Fine as observability, slightly wrong as wording.

Also trivial: the Tests section of the PR body does not list `TestFetchPullRequestsMarksMissingPRNotFound` or the new `not found placeholder` case.

### Verdict

Approving. The identity model, the positional contract, and the not-found classification now hang together, and the test doubles all encode the same contract as production.

whoisasx (approved):
Approving. All four findings from the earlier review are resolved at `1dd34b9` — not-found semantics preserved through the aligned-fetch contract (`ports.ErrSCMNotFound` stamped on the placeholder, `fetchClaimObservation` classifying it back to 404), the integration double brought in line with positional alignment, the `prKeyFromObs` residue cleaned up, and the duplicate-row claim narrowed to what the code and the 0097 index actually guarantee.

Detailed pass in the review above. The only outstanding item is a logging nit (permanent not-found now warns every tick) that is fine to fold into a follow-up.

backend/internal/adapters/scm/github/observer_provider.go
AgentWrapper: Done in 907872048 (rebased onto your 1dd34b96b/b47f2f1dd). Both observer fetch loops now check `errors.Is(fetchErr, ports.ErrSCMNotFound)` ahead of the rate-limit test and log at Debug (`tracked PR not found at provider; marking refresh-incomplete` / reconcile variant), so the permanent miss stays out of the warn-level log and is distinguishable from transient failures. Also enriched the GitHub placeholder with the ref's provider/host/repo/number and a wrapped context message (still `errors.Is`-compatible with your test), and added `TestPoll_NotFoundObservationLogsDebugAndPinsRepo` pinning: no warn-level output, no rate-limit cooldown, repo ETag stays pinned, nothing persisted. I kept your table-driven second-scan test and re-added the duplicate-row collapse test that went missing in the rebase shuffle.

whoisasx: Addressed in 907872048: both fetch loops now classify `ErrSCMNotFound` before rate-limit/transient handling and log the permanent miss at Debug while keeping the repo refresh-incomplete. Added adapter and observer regression coverage. Follow-up 5b01248e8 removes the schema-impossible duplicate-row fixture that the concurrent update accidentally restored.

PR #4091, opened 2026-08-18
fix(chat): contain long user messages

illegalcall:
## Summary

- wrap unbroken text in normal and mid-turn user message bubbles
- add a browser regression that checks the bubble has no horizontal content overflow
- attach matched before/after screenshots from the deterministic reproduction

## Root cause

User bubbles preserved authored whitespace but did not provide an overflow-wrap rule. A long JSON/OCR token therefore kept a 982 px scroll width inside a 497 px bubble and painted beyond its border.

### Before

Long OCR JSON escaping the message bubble before the fix

### After

The same OCR JSON wrapping within the message bubble after the fix

## Verification

- `npx playwright test e2e/chat-message-overflow.spec.ts --repeat-each=2 --workers=1 --reporter=line` (2 passed)
- `npm test -- src/renderer/components/chat/ChatWorkspace.test.tsx src/renderer/components/chat/ChatProviderSignal.test.tsx` (78 passed)
- `npm run typecheck`
- `npm run typecheck:e2e`
- `git diff --check`

## Notes

- `npm run build` is not available because frontend/package.json defines no build script.
- Native Electron launch was skipped to avoid colliding with another active worktree that owns the shared dev profile and port 3002; the matched screenshots and geometry assertion use the renderer browser harness.


frontend/e2e/chat-message-overflow.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+							sequence: 1,
+							revision: 0,
+							role: "user",
+							origin: "human",
+							text: longMessage,
+							streaming: false,
+							createdAt: now,
+						},
+					],
+					activities: [],
```
illegalcall: **[P2] Exercise the mid-turn message path**

This fixture leaves `activities` empty, so the regression verifies only `HumanMessage`; it would still pass if the separate `SteerMessage` overflow fix at `ChatTimelineItems.tsx:1400` were later removed. Because this PR changes both normal and mid-turn bubbles, please add a steer activity carrying `longMessage` and assert that bubble has no horizontal content overflow too. This covers the second user-visible rendering boundary changed by the PR.

illegalcall (commented):

## Standards
The production fix is surgical, but one changed user-visible rendering boundary lacks regression coverage; see the inline comment. `AGENTS.md` requires tests to cover the user-visible behavior and boundary being changed.

## Spec
The `break-words` changes correctly implement containment for both normal and mid-turn user bubbles, and the matched screenshots support the reported fix. I found no production correctness issue or scope creep.

Tests were not run, per reviewer policy.

frontend/e2e/chat-message-overflow.spec.ts
illegalcall: Addressed in 917f16848. The fixture now includes a system steer activity carrying the same longMessage payload, and both HumanMessage and SteerMessage are checked with the shared browser-geometry assertion. I also ran a negative control with only the SteerMessage break-words class removed: the steer assertion failed at 1144 px scroll width versus 497 px client width, then passed twice after restoring the fix. E2E typecheck and all PR checks are green.

illegalcall (commented):
**Verdict: approved — ready to merge.**

## Standards
No findings. The follow-up adds the steer activity to the browser regression and runs both `HumanMessage` and `SteerMessage` through the shared geometry assertion, satisfying the `AGENTS.md` requirement to test each changed user-visible boundary. No meaningful baseline smells were found.

## Spec
No findings. Both message variants retain authored whitespace while wrapping unbroken text, the regression now checks viewport containment and internal horizontal overflow for both bubbles, and the matched screenshots remain included. The previous coverage gap is resolved.

**Summary:** Standards: 0 findings. Spec: 0 findings.

Tests were not run, per reviewer policy.

illegalcall (commented):
**Verdict: approved — ready to merge.**

## Standards
No findings. The latest revision only removes repository screenshot artifacts; the production and regression-test changes approved at `917f168` are unchanged, and the aggregate diff is now more focused. No meaningful baseline smells were found.

## Spec
No findings. Wrapping and browser coverage still include both normal and mid-turn user bubbles. The matched before/after evidence remains attached in the PR body, so removing the PNGs from the repository does not leave a requirement gap.

**Summary:** Standards: 0 findings. Spec: 0 findings.

Tests were not run, per reviewer policy.

illegalcall:
## Screenshot evidence

### Before — reproduced overflow

Long OCR JSON escaping the user message bubble before the fix

### After — contained within the bubble

The same OCR JSON wrapping within the user message bubble after the fix

frontend/e2e/chat-message-overflow.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+			bubbleScrollWidth: node.scrollWidth,
+			bubbleClientWidth: node.clientWidth,
+		};
+	});
+
+	expect(layout.bubbleRight).toBeLessThanOrEqual(layout.viewportRight);
+	expect(layout.bubbleScrollWidth).toBeLessThanOrEqual(layout.bubbleClientWidth);
+}
+
+test("long unbroken user messages stay inside their chat bubble", async ({ page }) => {
```
Pulkit7070: The fix is correct, but this regression test will not run in CI, so it cannot actually guard against the bug recurring, which is its stated purpose.

The only CI job that runs Playwright specs is `renderer-smoke` in frontend.yml, and it runs `test:e2e:renderer` = `playwright test --grep "@T0|@P0"` (tagged specs only). No PR or nightly job runs the full untagged suite. This spec has no `@T0`/`@P0` tag, so it is skipped everywhere.

Add the tag to the test title so `renderer-smoke` picks it up, matching the existing convention (e.g. `model-menu-scroll.spec.ts`):

    test("long unbroken user messages stay inside their chat bubble @T0", async ({ page }) => {

Without this, the test passes locally via `npm run test:e2e` but a future regression would sail through CI unnoticed.

illegalcall: Added the `@T0` tag to the overflow regression in commit `4a7f8243c`, so `npm run test:e2e:renderer` now selects it via the existing `@T0|@P0` grep. Verified with the CI selector (`playwright test --list --grep "@T0|@P0"`), `npm run typecheck:e2e`, and the tagged browser spec itself (`1 passed`).

illegalcall (commented):
**Verdict: approved — ready to merge.**

## Standards
No findings. The `@T0` tag follows the existing Playwright convention and makes the regression part of the renderer CI selector. The remaining production and test changes stay surgical and cover both changed rendering boundaries. No meaningful baseline smells were found.

## Spec
No findings. Normal and mid-turn user bubbles both wrap unbroken text, the browser regression checks both variants for viewport containment and internal horizontal overflow, and the latest revision resolves the CI-selection gap.

**Summary:** Standards: 0 findings. Spec: 0 findings.

Tests were not run, per reviewer policy.

Pulkit7070 (approved):
LGTM. The @T0 tag (4a7f8243c) means renderer-smoke now actually runs the overflow regression, and the wrap fix is correct. Green across CI.

PR #4092, opened 2026-08-18
fix(chat): preserve image attachments when steering

illegalcall:
## Summary

- stage pending chat attachments before steering, using the same durable worktree references as queued sends
- forward negotiated native image blocks through the renderer client and daemon steer endpoint
- settle in-flight paste/drop reads before submission so an immediate Enter cannot send a text-only steer
- synchronously guard submission and disable composer controls so key repeat cannot issue duplicate steers
- preserve the draft and attachments when the provider refuses a steer, then restore textarea focus for keyboard retry
- render the durable staged image inline in the landed steer bubble, hide the internal attachment suffix, and avoid duplicate native thumbnails
- regenerate the OpenAPI contract and typed frontend schema

## Root cause

The composer deliberately treated steering as text-only: attachment-only drafts left the steer button disabled, and text-plus-image steers bypassed attachment staging. The frontend steer request and HTTP DTO also omitted attachments, even though the chat service and Codex/ACP driver layers already support structured image content during steering.

The first review follow-up found that paste/drop starts FileReader asynchronously, but submission inspected render-time attachment state. Pressing Enter before a large image finished reading could therefore steer the existing text alone and leave the image behind in the composer.

The second review follow-up found a re-entrancy window before onSteer set its external pending state. Repeated Enter during the file read or staging await could run two submissions concurrently, each with a fresh client message ID, producing duplicate guidance. Disabling the active textarea exposed a retained-draft focus regression on refusal/failure, so the composer now restores focus after re-enabling.

Native testing then exposed a separate history-rendering asymmetry: ordinary human messages parsed AO staged attachment references, while SteerMessage printed the internal path suffix and rendered only optional native content blocks. A successfully delivered staged image could therefore exist and open in Browser without appearing in chat history.

### Before the fix

The regression probes reproduced the loss and concurrency points:

    ChatComposer: expected “Steer the running turn” to be enabled; received disabled
    useConversation: POST /conversation/steer omitted attachments
    conversation_steer controller: service received Content == nil
    immediate-paste regression: onSteer received only “inspect this” while FileReader was pending
    double-Enter regression: attachment staging was called twice before FileReader completion
    history regression: landed steer had no inline image despite its durable .ao/attachments reference

### After the fix

| Check | Result |
| --- | --- |
| Immediate-paste/double-Enter regression | Two Enter events are batched before React re-renders; controls disable immediately, staging runs once, and onSteer runs once with text plus native image bytes |
| Retained-draft regression | refusal/failure keeps text and attachments and restores textarea focus for the next Enter |
| Steer-history regressions | staged image renders inline, internal path suffix stays hidden, and matching native content does not duplicate the thumbnail |
| Focused affected suites | 5 files passed; 155 tests passed |
| Frontend typecheck | passed |
| Changed HTTP/spec packages | passed, including route/spec parity |
| Generated API artifacts | regenerated with no diff |
| Full renderer suite | 191 files passed; 2,525 tests passed; 1 skipped |
| GitHub Actions | 10/10 checks passed on rewritten/rebased head d46e2f2d1 |

Actual Electron evidence is attached in [this PR comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4092#issuecomment-5330876245). The corrected after capture visibly places the AO elephant inside the landed steer bubble with Summary selected and Browser unselected. The live renderer reported a complete 1024×1024 image, and the session preview request returned HTTP 200.

## Known local suite issue

The broad backend suite still reports five unrelated internal/cli spawn-context tests returning HTTP 404. No CLI files are changed here; the changed HTTP/spec packages pass. Two unrelated interactive-shell auth tests timed out only under the fully parallel run and passed immediately in isolation.


illegalcall:
Addressed the immediate-Enter attachment race in `a39b8bede`.

- `ChatComposer.submit()` now awaits `toSettledPayload()` before deciding whether the submission has attachments.
- Staging, native-image filtering, cleanup, and retry-path caching all use that settled snapshot rather than render-time attachment state.
- The cache signature is read from the same ref, so a resumed submit does not depend on a follow-up React render.
- Added a delayed `FileReader` regression that pastes into an existing text draft, presses Enter immediately without waiting for the chip, verifies no steer occurs while the read is pending, then verifies the staged path and native image reach `onSteer` after settlement.

Evidence:

```text
Before: onSteer was called once with only "inspect this" while FileReader was pending.
After:  immediate-Enter regression passes (1/1).
Focused: 3 files, 79 tests passed.
Full renderer: 191 files, 2,508 tests passed, 1 skipped.
TypeScript: npm run typecheck passed.
```

illegalcall (commented):
## Verdict: Changes requested

The previous FileReader omission is fixed, and the end-to-end image and API plumbing remains coherent. One submission-concurrency bug remains: the composer is still actionable while it waits for attachment reads and staging, so repeated Enter can issue duplicate steers. Please guard the full submission before the first await and add double-Enter coverage.

All 10 checks on `a39b8bede` are green. I did not execute project code or tests, per the review constraints.

frontend/src/renderer/components/chat/ChatComposer.tsx
```diff
@@ -339,62 +335,74 @@ export function ChatComposer({
...
-				await onSteer(body);
-			} catch {
-				// The refusal is the daemon's typed answer and the surface renders it from
-				// `steerRefusal`; keep the draft, but arm the reliable queue path for the
-				// next Enter in case the turn ended while the user was typing.
-				setDelivery("queue");
+		// Paste/drop reads finish asynchronously. Settle them before deciding whether
+		// this submission has attachments, or immediate Enter can steer the text and
+		// leave the image behind when its FileReader completes.
+		const attachmentPayloads = await fileAttachments.toSettledPayload();
```
illegalcall, line 504: **[P2] Guard the whole attachment submission before this await.** `canSend` only reflects `busy` and `steerPending`, and neither flips until after `toSettledPayload()` and `onStageAttachments` complete and `onSteer` starts. If a slow read or stage makes the user press Enter again, or key repeat fires, a second `submit()` runs concurrently; both calls can stage the same payload and then invoke `onSteer`. `useConversation` creates a fresh `clientMessageId` per invocation, so daemon idempotency does not collapse the duplicate guidance. Set a synchronous local submission guard and disable the controls before awaiting, then cover double Enter in the regression test.

illegalcall: Addressed in `f7475b01d`.

- Added a synchronous `submissionInFlight` ref that is set before the first await, so a second Enter/key-repeat in the same render returns immediately.
- Added local `submitting` state and disabled the textarea, destination choices, attachment actions, and submit control until the full submission settles; a `finally` releases both guard and UI state on success or failure.
- Extended the delayed-`FileReader` regression to batch two Enter events before React can re-render. It now asserts staging and `onSteer` each run exactly once and verifies the controls are disabled while the read is pending.

Verification: targeted regression 1/1 passed; focused composer suite 79/79 passed; `npm run typecheck` passed; full renderer suite 2,508 passed with 1 skipped.

frontend/src/renderer/components/chat/ChatComposer.tsx
```diff
@@ -563,7 +588,7 @@ export function ChatComposer({
 				onClick={onSelectionChange}
 				onPaste={onPaste}
 				rows={1}
-				disabled={disabled}
+				disabled={controlsDisabled}
```
illegalcall: **[P2] Keep or restore focus when the draft is retained.** Chromium unfocuses a focused form control when it becomes disabled, and changing `submitting` back to false does not focus it again. The attachment-staging failure, steer-refusal, queued-send failure, and compact refusal/error paths intentionally keep the current draft but return without `applyText` or any other refocus path. The composer therefore re-enables with keyboard focus elsewhere, so the intended queue-on-next-Enter retry cannot happen until the user clicks or tabs back into the field. Prefer `readOnly` plus the synchronous submission guard, or explicitly restore focus after the enabled state commits, and cover a retained-draft refusal in a focus regression test.

illegalcall (commented):
## Verdict: Changes requested

The synchronous single-flight guard fixes the duplicate-steer race, and the end-to-end attachment path remains coherent. One keyboard regression remains: disabling the focused textarea removes focus in Chromium, while retained-draft refusal and failure paths do not restore it. The documented next-Enter queue retry therefore requires an extra click or tab. Please keep the field focusable during submission or restore focus after it is re-enabled, with regression coverage.

All 10 checks on `f7475b01d` are green. I did not execute project code or tests, per the review constraints.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
## Native Electron before/after evidence — corrected

These were exercised in the actual Electron desktop app with isolated AO data. The inspector is on **Summary** in both captures; the **Browser** tab is not selected.

### Before

The image remains detached in the composer after the steer lands as text-only, and the agent reports that no image arrived.

Before: steer lands without the image

### After — native Electron verification

The AO elephant is rendered **inline inside the landed steer bubble in chat history**, immediately above “Steered into the running turn.” This replaces the earlier misleading capture that only proved the Browser panel could display the file.

After: preserved image rendered inline in chat history

Native-app verification on the live Electron renderer:

- Chat-history DOM contained `img[alt="attachment-3f540c6112628109887679c8cc1842a5.png"]`.
- Its source was the session-scoped `/api/v1/sessions/pr4092-evidence-1/preview/files/.ao/attachments/...` endpoint.
- The element was complete at natural size 1024×1024, and the daemon returned HTTP 200 with 51,274 bytes.
- Selected tabs were “Steer Image Proof” and “Summary”; “Browser” was visible but unselected.

The post-fix capture reopens the controlled successful steer session in the current native build, proving the durable staged reference now renders in transcript history. Regression coverage also asserts that the internal staged-path suffix is hidden and that native image content does not create a duplicate thumbnail.

Verification on rewritten/rebased head `d46e2f2d1`:

- Focused affected suites: 5 files and 155 tests passed.
- Full renderer suite: 191 files and 2,525 tests passed, with 1 skipped.
- Frontend typecheck passed.
- Changed HTTP/spec packages passed, including route/spec parity.
- Generated OpenAPI and frontend schema reproduced with no diff.
- GitHub Actions: 10/10 checks passed.

frontend/src/renderer/components/chat/ChatComposer.tsx
illegalcall: Addressed in afe58f5de.

- The composer now records whether submission originated from the textarea before disabling controls.
- After submitting returns to false, a layout effect restores focus when Chromium moved it to the document body; it does not steal focus if the user deliberately moved to another control.
- The retained-draft steer-refusal regression explicitly models Chromium blur, verifies that the draft is retained and Queue is selected, asserts the textarea regains focus, then presses Enter without clicking and confirms the queued retry is sent.

Verification: targeted steering suite 20/20 passed; focused composer suite 61/61 passed; npm run typecheck passed; full renderer suite 2,508 passed with 1 skipped.

illegalcall (commented):
## Verdict: Approved

The current head resolves the previously identified focus regression: it records textarea-origin submissions before disabling the field, restores focus after re-enabling only when focus was lost to the document, and avoids stealing focus from another control. The retained-steer test verifies draft retention, queue fallback, focus restoration, and an immediate Enter retry. The attachment-settlement and single-flight behavior, native image propagation, generated API artifacts, and surrounding focused coverage remain coherent.

I found no blocking correctness, security, error-handling, or repository-standards issues. All reported checks for `afe58f5de` are green; I did not run tests locally under the review-only constraint.

illegalcall (commented):
## Verdict: Approved

At `58418b906`, the landed-steer follow-up is coherent: `SteerMessage` reuses the validated staged-reference parser and session-scoped preview URL, hides only the AO-generated attachment suffix, renders the durable staged image inline, and suppresses the corresponding native thumbnail while retaining non-image structured content. The focused test now asserts the preview route and duplicate suppression, while existing backend coverage exercises canonical attachment serving. The earlier attachment settlement, single-flight submission, native image propagation, and retained-draft focus fixes remain sound.

I found no blocking correctness, security, error-handling, spec, or repository-standards issues. All 10 reported checks on `58418b906` are green; I did not run tests locally under the review-only constraint.

illegalcall (commented):
## Verdict: Approved

The rewritten/rebased head preserves the previously approved implementation. Every PR-touched source and test file is byte-identical to the approved `58418b906` tree; the intervening main change is confined to `ChatMarkdown` and introduces no integration conflict. Attachment settlement and single-flight behavior, native steer propagation, retained-draft focus recovery, landed-image rendering, preview routing, duplicate suppression, and generated API artifacts therefore remain coherent. Removing the committed screenshots is not a functional or repository-standard regression because the PR still links the native Electron evidence in its comment.

I found no blocking correctness, security, error-handling, spec, or repository-standards issues. All 10 reported checks on `d46e2f2d1` are green; I did not run tests locally under the review-only constraint.

illegalcall:
Local desktop verification for commit `f4a942169`:

- Ran this checkout in the real AO Electron app, not the browser preview.
- Used isolated dev state (`~/.ao/dev/data`, `~/.ao/dev/running.json`).
- Renderer: `http://localhost:5173/`; daemon: `http://127.0.0.1:3002/`.
- Started a live Codex chat session, attached `frontend/assets/icon.png` while the turn was still working, and submitted the guidance with Cmd+Enter.
- Confirmed the composer cleared, the turn remained active, and history rendered the durable staged image exactly once with “Steered into the running turn.”

Screenshot evidence:

AO Electron showing an image steer rendered inline while the Codex turn remains active

nikhilachale (approved):

PR #4094, opened 2026-08-18
fix(usage): collect cost data for Chat sessions

illegalcall:
## Summary

- collect usage for Chat workers from their provider conversation identity
- keep TUI collection on the hook-derived agent session identity
- apply the identity selection consistently during lifecycle reactivation, hook fallback, and daemon-start backfill
- cover both the live Chat controller lifecycle and startup recovery paths

## Root cause

A fresh Chat controller persists its native transcript identity in `provider_conversation_id`. It has no terminal hook, so `agent_session_id` remains empty. The usage collector previously consulted only `agent_session_id`, which meant it could not create a usage binding or discover the provider JSONL transcript for Chat-only sessions.

Switching the same session to TUI copied the native identity into `agent_session_id`. That made the existing transcript discoverable immediately, which is why cost observability appeared to start working after the interface conversion.

This change selects `provider_conversation_id` for committed Chat sessions and preserves `agent_session_id` for TUI sessions. Older migrated Chat records still fall back to their hook-derived identity when no provider conversation ID exists.

**Root-cause confidence:** High — confirmed through the lifecycle call path, persisted SQLite state, transcript timestamps, API responses, a red-before-green regression, and the native Electron flow below.

## Actual Electron end-to-end evidence

The reproduction and verification used the real Electron checkout with isolated state under `~/.ao/dev`; neither screenshot comes from a mock or browser-only preview.

### Before

A fresh Chat Codex worker had already completed its turn and the conversation endpoint reported 18,213 tokens, while `/api/v1/usage/sessions/<id>` remained empty and the board card had no token badge. Converting that exact session to TUI exposed the same 18,213 tokens immediately without another model turn.

Before the fix: the Chat cost check card has no token badge

### After

A new post-fix Codex worker remained in Chat mode with empty `runtime_launch_id` and `agent_session_id`. Its provider-conversation usage binding became active, `/api/v1/usage/sessions/<id>` returned 18,218 tokens, and the board rendered `18.2K tok` without any interface conversion.

After the fix: the fresh Chat worker shows 18.2K tokens without switching to TUI

The immutable evidence assets are stored at commit [`e622b943`](https://github.com/Untrivial-ai/agent-orchestrator/commit/e622b94379eb4e2c3625224ffd130bdeacb607ae).

## Verification

- `cd backend && env -u AO_PROJECT_ID -u AO_SESSION_ID GOFLAGS=-p=1 go test ./...`
- `cd backend && go test -race ./internal/service/usage`
- focused lifecycle and startup-backfill regressions, three consecutive runs
- `env -u AO_PROJECT_ID -u AO_SESSION_ID GOFLAGS=-p=1 npm run lint` (`0 issues`)

## Risk

Low. The behavior change is confined to choosing the already-persisted provider transcript identity for Chat sessions. TUI identity selection and unsupported harness behavior are unchanged.

### DIFF
--- a/backend/internal/service/usage/collector.go
+++ b/backend/internal/service/usage/collector.go
@@ -198,7 +198,7 @@ func (c *Collector) ReactivateSession(
 	if err != nil {
 		return err
 	}
-	nativeID := boundedUsageMetadata(session.Metadata.AgentSessionID)
+	nativeID := usageNativeSessionID(session)
 	var current *domain.UsageBindingRecord
 	for index := len(bindings) - 1; index >= 0; index-- {
 		binding := &bindings[index]
@@ -235,7 +235,7 @@ func (c *Collector) RecordHook(ctx context.Context, sessionID domain.SessionID, 
 	signal.Harness = session.Harness
 	signal.NativeSessionID = boundedUsageMetadata(signal.NativeSessionID)
 	if signal.NativeSessionID == "" {
-		signal.NativeSessionID = boundedUsageMetadata(session.Metadata.AgentSessionID)
+		signal.NativeSessionID = usageNativeSessionID(session)
 	}
 	if signal.NativeSessionID == "" {
 		signal.NativeSessionID = nativeIDFromTranscript(signal.TranscriptPath)
@@ -479,7 +479,7 @@ func (c *Collector) BackfillActive(ctx context.Context) error {
 		if session.IsTerminated || !SupportedHarness(session.Harness) {
 			continue
 		}
-		nativeID := boundedUsageMetadata(session.Metadata.AgentSessionID)
+		nativeID := usageNativeSessionID(session)
 		if nativeID == "" || !nativeUsageIDPattern.MatchString(nativeID) {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


illegalcall (commented):
**Verdict: Approved — ready to merge.**

## Standards
No findings. The change follows the documented service/domain boundaries, safely normalizes and bounds the selected native ID, and adds focused coverage for lifecycle-triggered Chat discovery and startup backfill. No baseline code smells identified.

## Spec
No findings. The shared selector makes committed Chat sessions prefer `ProviderConversationID`, preserves `AgentSessionID` for TUI, falls back for older Chat records, and is applied at lifecycle reactivation, hook fallback, and daemon-start backfill as requested.

Tests were not run, per the read-only review policy.

whoisasx (approved):
Looks good to me. The Chat usage identity is selected consistently across lifecycle reactivation, hook fallback, and daemon-start backfill while preserving the existing TUI path. Focused collector, lifecycle, repeated regression, race, and current-main merge verification all passed, and all GitHub checks are green.

PR #4096, opened 2026-08-18
fix(chat): support agent switching without a terminal runtime

illegalcall:
Addresses the remaining functional agent-switch gap in #4033 and follows the UI entry-point work in #4036.

## Summary

- make the switch-agent control available for real Chat sessions, which intentionally have no terminal handle
- add a Chat-specific durable handoff that replaces structured controllers without manufacturing a TUI runtime
- fence source/target ownership with controller generations, atomically activate the target, and preserve Chat mode with empty runtime metadata
- restore the source controller when target activation fails and reconcile every in-flight Chat switch state after daemon restart
- make the renderer fake model Chat sessions accurately and add unit, SQLite, and Playwright regression coverage

## Root cause

The UI-only follow-up in #4036 wrapped `TerminalSwitchAgentButton` in `session.terminalHandleId`. That condition can never be true for a correctly formed Chat session: Chat is driven by a structured provider controller and deliberately persists `terminalHandleId: null`.

The same hidden assumption existed in the backend. Agent-switch admission required a runtime handle for every mode, and the switch saga only knew how to destroy and replace a terminal runtime. The fake renderer bridge assigned terminal handles to Chat fixtures too, so tests modeled a state production never creates and masked both failures.

This change uses the durable Chat controller generation as the ownership fence, while retaining the existing runtime-launch fence for TUI sessions.

## Live Electron A/B evidence

Both captures use the actual AO Electron app and daemon with a live idle Chat worker. This is not the `dev:web` fake bridge.

| Build | Session mode | `terminalHandleId` | Result |
|---|---|---:|---|
| Base `3f5ca81bf` | `chat` | `null` | switch control absent |
| Feature build | `chat` | `null` | control visible; dialog opens |

### Before — production-shaped Chat session has no switch control

Before: idle Chat session without a switch-agent control

### After — the same session shape opens the switch dialog

After: switch-agent dialog open in an idle Chat session

The evidence images are hosted separately on `pr-assets-4096` and are intentionally excluded from this PR's source diff.

## Verification

Verified again after merging current `origin/main` (`3369f00c7`) into the branch:

- `cd backend && go test -p 4 ./...` — passed
- focused backend race suites for ACP, Codex, Chat service, session manager, and SQLite store — passed
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs` — 0 issues with an isolated cache
- `npm run sqlc` — generated output clean
- `cd frontend && npm run typecheck` — passed
- `cd frontend && npm run typecheck:e2e` — passed
- `cd frontend && npx vitest run` — 192 files / 2,529 tests passed; 1 skipped
- `cd frontend && npm run test:e2e -- e2e/chat-agent-switch.spec.ts` — 1 passed

The backend regression matrix covers successful Chat-to-Chat ownership transfer, activation rollback, ambiguous commit resolution, all durable restart states, and the invariant that Chat never acquires terminal runtime metadata.

## Scope

This closes the broken agent-switch behavior. The broader shared tab-strip refactor tracked as Gap 3 in #4033 remains a separate maintenance follow-up.


backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,578 @@
...
+	sourceStopped := false
+	targetWorkspacePrepared := false
+	targetOwnerCommitted := false
+	targetOwnershipAmbiguous := false
+	skipTerminalization := false
+	targetEnv := m.runtimeEnv(rec.ID, rec.ProjectID, rec.IssueID, project.Config.Env)
+	m.augmentAgentRuntimeEnv(targetAgent, targetEnv)
+
+	defer func() {
+		if retErr != nil && !sourceStopped {
```
illegalcall: **[P2] Clean up an armed handoff when the worker panics.** This defer only reacts to `retErr`, but a panic does not assign the named return. If a panic occurs after `PrepareChatHandoff` and before source shutdown, this defer skips `AbortHandoff`; the outer worker recovery then treats the in-process state as daemon-restart state, fails the durable switch, and releases the mutation gate while the source controller remains fenced with `ErrControllerHandoff`. Recover into an error before this cleanup, or explicitly abort/restore the live controller in the panic path, and cover a panic after arming.

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,578 @@
...
+	case domain.AgentSwitchTargetReady, domain.AgentSwitchDelivering:
+		recoverable, err := m.chatTargetNativeIdentityRecoverable(ctx, store, rec, sw)
+		if err != nil {
+			return false, err
+		}
+		if !recoverable {
+			return false, fmt.Errorf("reconcile Chat agent switch %s: activated target identity is not durable", sw.ID)
+		}
+		if sw.State == domain.AgentSwitchTargetReady {
+			return fail(domain.AgentSwitchErrorDaemonRestartBeforeDelivery)
```
illegalcall: **[P1] Preserve the finalized continuation across `target_ready` recovery.** At this point activation has already made the target the durable owner, but the only handoff context was supplied in the launch-time system prompt and `StartChatTurn` has not run. After a daemon restart, normal Chat resume rebuilds only the baseline system prompt, so failing this saga releases the gate with a target that never received the original task or handoff. Reconciliation must durably reapply or deliver the finalized handoff, or roll back, before terminalizing this state; add a restart test for the activation-to-delivery window.

backend/internal/storage/sqlite/store/agent_switching_store.go
```diff
@@ -571,6 +580,96 @@ func (s *Store) ActivateAgentSwitchTarget(ctx context.Context, activation domain
...
+		ControllerGeneration:   activation.ControllerGeneration,
+		SessionID:              activation.SessionID, ExpectedSourceHarness: activation.SourceHarness,
+	})
+	if err != nil {
+		return false, fmt.Errorf("activate Chat agent switch target %s: transfer session owner: %w", activation.SwitchID, err)
+	}
+	if n != 1 {
+		return false, nil
+	}
+	if err := q.ResetConversationAgentOverridesForSession(ctx, gen.ResetConversationAgentOverridesForSessionParams{
```
illegalcall, line 676: **[P1] Reset the settings seen by the live target controller, not only SQLite.** `StartChat` loads the existing conversation and constructs `Controller.settings` before invoking `ControllerReady`; this callback reaches this query only afterward. The target activation turn then overwrites its message with the still-cached source model and reasoning settings. A Claude-to-Codex switch with a selected Claude model, for example, sends that invalid model to Codex after target ownership was committed and can fail the switch. Reset or seed the target settings before controller construction while preserving approval, and test a switch with nonempty source overrides.

backend/internal/storage/sqlite/queries/agent_switching.sql
```diff
@@ -331,6 +343,28 @@ WHERE id = sqlc.arg(session_id)
...
+    harness = sqlc.arg(target_harness),
+    activity_state = 'idle',
+    activity_last_at = sqlc.arg(activated_at),
+    first_signal_at = NULL,
+    runtime_handle_id = '',
+    runtime_launch_id = '',
+    agent_session_id = sqlc.arg(target_native_session_id),
+    agent_session_id_launch_id = '',
+    native_transcript_path = '',
+    provider_conversation_id = sqlc.arg(provider_conversation_id),
```
illegalcall, line 357: **[P2] Move the active conversation branch across the provider boundary atomically.** This updates the session handle from the source provider to the target, but the active `conversation_branches.provider_conversation_id` remains the source handle because the existing trigger only updates an empty root on initial start. Target turns are then projected onto a branch whose native identity belongs to another harness. Later `ActivateBranch` passes that source handle to the current target driver, and editing earlier turns can mix incompatible provider turn IDs. Create and activate an explicit target-provider branch boundary in the same transaction, and add branch and edit coverage after switching.

illegalcall (commented):
## Verdict: Changes requested

### Standards
The new Chat saga leaves a live controller fenced if execution panics after arming handoff, because cleanup only runs for ordinary returned errors.

### Spec and correctness
The switch can lose its durable continuation on restart, retain source-provider turn settings in the target controller, and leave conversation branch identity on the source provider. These break restart recovery and cross-provider continuity. Inline comments describe the concrete failure paths and requested coverage.

Tests were not run under the read-only AO review policy.

backend/internal/session_manager/agent_switching_chat.go
illegalcall: Fixed in f87e543e6. `executeChatAgentSwitch` now recovers a panic into its named error before cleanup evaluates `retErr`, so a pre-stop panic runs `AbortChatHandoff`, terminalizes the saga as `failed_pre_stop`, and releases the mutation gate. Added `TestSwitchAgentChatPanicAfterPreparingHandoffAbortsSource`, which was red with `prepared=[proj-1], aborted=[]` before the fix and now passes normally and under `-race`.

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,644 @@
...
+	sourceStopped := false
+	targetWorkspacePrepared := false
+	targetOwnerCommitted := false
+	targetOwnershipAmbiguous := false
+	skipTerminalization := false
+	targetEnv := m.runtimeEnv(rec.ID, rec.ProjectID, rec.IssueID, project.Config.Env)
+	m.augmentAgentRuntimeEnv(targetAgent, targetEnv)
+
+	defer func() {
+		if panicValue := recover(); panicValue != nil {
```
illegalcall, line 43: **[P1] Preserve the unknown post-stop phase instead of terminalizing it as pre-stop.** This catch-all `recover` bypasses the worker-level panic handler. There is a window after `stopSourceControllerConclusive` returns successfully and before `sourceStopped = true` at line 209. A panic there becomes `retErr`, so this defer treats the source as still live, calls `AbortChatHandoff` on a controller that is already gone, fails the `stopping_source` row as `failed_pre_stop`, and releases the mutation gate without restoring the source. Limit local recovery to the armed pre-stop handoff window, or track conclusive stop and retain or reconcile the saga; add fault injection immediately after the stop returns.

backend/internal/storage/sqlite/store/agent_switching_store.go
illegalcall, line 676: **[P1] Reset the settings seen by the live target controller, not only SQLite.** This still runs from `ControllerReady`, after Chat Service has constructed the new `Controller` from the existing `conversation.Settings`. The database is reset, but the live controller keeps the source model and reasoning effort, and its activation turn overwrites the message with those cached settings. A Claude-to-Codex switch can therefore send a Claude model to Codex after target ownership commits. Reset or seed settings before controller construction, or synchronize the live controller before dispatch, and cover nonempty source overrides.

backend/internal/storage/sqlite/queries/agent_switching.sql
illegalcall, line 357: **[P2] Record a target-provider boundary in the conversation branches.** This still replaces only the session provider handle. The existing root-binding trigger runs only for an empty-to-nonempty transition, so the active `conversation_branches.provider_conversation_id` remains the source handle while target turns are projected onto it. Later branch activation passes that source handle to the target driver, and editing can mix incompatible provider turn IDs. Atomically create and activate a target-provider branch boundary, or explicitly reject incompatible historical operations, with branch and edit coverage after switching.

illegalcall (commented):
## Verdict: Changes requested

Re-review of `fb735bfe6` confirms that target-ready restart recovery now reuses the reserved generation, reapplies the verified continuation, and delivers activation. The pre-stop abort case is fixed, but the catch-all panic recovery creates a different post-stop gap.

## Standards
- One blocking lifecycle issue remains: a panic after conclusive source stop but before durable confirmation is swallowed locally and terminalized as pre-stop instead of being retained and reconciled.

## Spec
- Two prior blockers remain: the live target controller retains source-provider model and reasoning settings, and the conversation branch native identity remains bound to the source provider.

**Summary:** Standards: 1 finding, worst issue is post-stop panic recovery. Spec: 2 findings, worst issue is source settings breaking target activation.

Tests were not run under the read-only AO review policy.

illegalcall (commented):
Addressed the changes requested in review 4961705652 in c6da5a133.

- Post-stop panic boundary: track a conclusive controller stop separately from its durable confirmation. A panic in that window now retains `stopping_source` with `source_stop_unconfirmed`, keeps the mutation gate closed, and is reconciled by restoring the source controller before terminalizing. The regression injects the panic after `StopChat` and verifies the full recovery.
- Target settings: after `ControllerReady` commits, Chat Service reloads the durable conversation before publishing the controller. The first activation dispatch now uses target-default model/reasoning settings while preserving the AO approval mode.
- Provider identity: Chat target activation now atomically creates and activates a target-provider branch boundary at the current sequence. Edit, branch activation, and rollback operations enforce that boundary so source-provider conversation/turn IDs cannot reach the target driver. Store, service, and HTTP regressions cover nonempty source overrides and all historical mutation paths.

Verification: `env -u AO_PROJECT_ID -u AO_SESSION_ID go test ./...`; `go test -race ./internal/session_manager ./internal/service/chat ./internal/storage/sqlite/store`; generated SQL is current; golangci-lint reports 0 issues.

backend/internal/session_manager/agent_switching_chat.go
illegalcall: Fixed in c6da5a133. The worker now records a conclusive source-controller stop before durable confirmation. A panic in that window skips live-source abort/terminalization, persists source_stop_unconfirmed, retains the gate, and reconciliation restores the source; TestSwitchAgentChatPanicAfterSourceStopRetainsRecoveryBoundary covers the injected panic and recovery.

backend/internal/storage/sqlite/store/agent_switching_store.go
illegalcall: Fixed in c6da5a133. Chat Service reloads the committed conversation row after ControllerReady and before publishing/dispatching from the controller, synchronizing the reset model/reasoning settings and preserving approval. The regression starts with nonempty source overrides and asserts the activation turn receives target defaults.

backend/internal/storage/sqlite/queries/agent_switching.sql
illegalcall: Fixed in c6da5a133. ActivateChatAgentSwitchTarget now atomically creates and activates a target-provider boundary branch at the current sequence. Provider scope is enforced before branch activation, edit-anchor selection, and rollback, so no source provider conversation/turn ID reaches the target driver. Store, service, and HTTP tests cover the boundary and rejections.

backend/internal/storage/sqlite/store/agent_switching_store.go
illegalcall: Also addressed by c6da5a133: the live target controller refreshes the committed conversation settings before its first dispatch, with nonempty source model/reasoning regression coverage.

backend/internal/storage/sqlite/queries/agent_switching.sql
illegalcall: Also addressed by c6da5a133: target activation now commits a provider boundary branch atomically, and incompatible branch/edit/rollback operations are rejected before the target driver sees an old provider ID.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (commented):
## Verdict: Changes requested

The latest commit fixes the previously reported post-stop panic boundary, successful-path settings synchronization, target-ready continuation recovery, and provider identity boundary. Five high-confidence gaps remain.

## Standards
- The settings synchronization adds a fallible step after the durable controller commit, which can leave target ownership without a live controller.
- The AO-generated activation turn is persisted as human-authored.

## Spec
- Chat always starts a fresh native target instead of matching the TUI switch-back resume behavior.
- The new provider boundary leaks into the existing edit-branch UI as a false branch and dead history controls.
- Chat adds the switch button but not the durable progress, recovery presentation, or input lock used by TUI.

Tests were not run under the read-only AO review policy.

backend/internal/service/chat/service.go
```diff
@@ -389,6 +397,31 @@ func (s *Service) Start(ctx context.Context, cfg StartConfig) (*Controller, erro
...
+		// ControllerReady may atomically transfer ownership, reset provider-specific
+		// settings, and move the active provider branch. The controller was built
+		// from the pre-commit row, so refresh it before publishing or dispatching.
+		// Once the callback commits, caller cancellation must not strand a live
+		// controller with the source agent's cached state.
+		refreshCtx, cancel := context.WithTimeout(
+			context.WithoutCancel(ctx), controllerReadyStateSyncTimeout)
+		refreshed, refreshErr := s.store.ConversationForSession(refreshCtx, cfg.SessionID)
+		cancel()
+		if refreshErr != nil {
```
illegalcall: **[P1] Keep the target recoverable after this post-commit read.** `notifyControllerReady` has already allowed the callback to atomically move the switch and session to the target. If this read times out or errors, or the following identity check fails, `Start` closes the target and returns. In `executeChatAgentSwitch`, `targetOwnerCommitted` is already true, so the defer fails the `target_ready` saga and releases the mutation gate, leaving durable target ownership with no registered controller or delivered activation until another restart. Ordinary resume can similarly call `MarkSpawned` and then return controller-less. Avoid fallible work after the commit, for example by returning the committed conversation state from the callback, or classify this as retained recovery instead of closing and terminalizing. Please add fault coverage for this boundary.

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,652 @@
...
+		ctx, id, domain.SessionInterfaceTransitionInterrupt); err != nil {
+		return result, fmt.Errorf("switch Chat agent %s: quiesce source: %w", id, err)
+	}
+
+	targetGeneration := domain.AgentGenerationID(strings.TrimSpace(m.newLaunchID()))
+	if targetGeneration == "" {
+		return result, fmt.Errorf("switch Chat agent %s: target controller generation is empty", id)
+	}
+	if err := m.advanceAgentSwitch(ctx, store, &result, domain.AgentSwitchStoppingSource, func(next *domain.AgentSwitch) {
+		next.TargetStartMode = domain.AgentSwitchTargetStartFresh
```
illegalcall: **[P2] Preserve the TUI switch-back semantics in Chat.** Hard-coding `TargetStartFresh` means the later `StartChat` call receives no `ProviderConversationID` and always creates a new native conversation, even when the native-session registry contains a verified target candidate. The current TUI path uses `findTargetResumeCandidate`, so Claude → Codex → Claude resumes the retained Claude session; #4033 asks Chat to expose that same in-place switch. Select and verify a target-native candidate here, record `resumed` and pass its provider handle when usable, with fresh fallback and switch-away/back coverage.

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,652 @@
...
+		return result, fmt.Errorf("switch Chat agent %s: target controller started without durable ownership", id)
+	}
+	result, err = requireAgentSwitch(ctx, store, result.ID)
+	if err != nil {
+		return result, fmt.Errorf("switch Chat agent %s: reload target activation: %w", id, err)
+	}
+	if err := m.advanceAgentSwitch(ctx, store, &result, domain.AgentSwitchDelivering, nil); err != nil {
+		return result, fmt.Errorf("switch Chat agent %s: begin continuation delivery: %w", id, err)
+	}
+	if _, err := m.chat.StartChatTurn(ctx, id, aoTargetActivationPrompt); err != nil {
```
illegalcall: **[P2] Attribute this AO-generated turn to automation.** `aoTargetActivationPrompt` is created by AO, and `aoAgentContinuationProtocol` explicitly says it is not a new instruction from the human, but `StartChatTurn` hard-codes `MessageOriginHuman`. The timeline therefore renders the coordination text as a human prompt and marks it editable. Use `RelayChatTurn` or a dedicated automation-origin activation path here and at the recovery delivery call.

backend/internal/storage/sqlite/store/agent_switching_store.go
```diff
@@ -571,6 +580,126 @@ func (s *Store) ActivateAgentSwitchTarget(ctx context.Context, activation domain
...
+	}
+	conversationRow, err := q.SelectConversationBySession(ctx, &activation.SessionID)
+	if errors.Is(err, sql.ErrNoRows) {
+		return false, fmt.Errorf("activate Chat agent switch target %s: conversation does not exist", activation.SwitchID)
+	}
+	if err != nil {
+		return false, fmt.Errorf("activate Chat agent switch target %s: read conversation: %w", activation.SwitchID, err)
+	}
+	providerBoundaryID := string(activation.SwitchID) + ":provider"
+	if err := insertConversationBranchTx(ctx, q, domain.ConversationBranch{
```
illegalcall, line 654: **[P2] Do not expose the provider boundary as an edit branch.** This child has a parent but no replacement turn. Existing snapshot code treats any child branch as `BranchedFromEarlierMessage`, so every successful switch now displays the misleading `Conversation branched; worktree files were left unchanged.` notice. Historical source-scope branch points and human messages also keep navigation and edit affordances even though the new provider-scope checks reject those clicks. Distinguish provider boundaries from edit branches and filter branch navigation and edit availability to the active provider scope.

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -173,15 +173,13 @@ export function SessionChatSurface({
...
-				session.terminalHandleId ? (
-					<TerminalSwitchAgentButton
-						container={switchSelectorContainer}
-						onOpenChange={setSwitchSelectorOpen}
-						open={switchSelectorOpen}
-						session={session}
-						switchError={switchMutation.error}
-					/>
-				) : undefined
+				<TerminalSwitchAgentButton
```
illegalcall: **[P1] Wire the durable switch presentation and input lock into Chat.** This passes no `presentation`, and `SessionChatSurface` reads only ephemeral mutation state. After the 202 response or a renderer or daemon reload, the button remains enabled, the composer and history controls still look interactive, and Chat shows no phase or recovery state. The backend rejects some operations, but the UI presents them as usable and provides no durable recovery status. Reuse or share the `CenterPane` agent-switch query and presentation logic, pass it to this control, and disable or make the Chat interaction surface inert according to the durable lock. Add nonterminal and reload coverage.

illegalcall:
Addressed review 4963096425 in f9f968b77.

- Removed the fallible post-commit conversation reread. ControllerReady now returns the conversation state committed by activation, and Chat installs the changed branch/settings cache with infallible in-memory assignments before publishing the controller. Regression coverage injects failures for every ConversationForSession read and verifies target settings reach the first dispatch.
- Matched TUI switch-back behavior by probing the verified native target before source shutdown, resuming its exact provider conversation with a generation CAS, and separating provider resume from old-history projection so prior target events cannot land in the source branch.
- Deliver the AO activation through RelayChatTurnWithID using a stable switch-derived key, so normal and restart-recovery paths persist it as automation and cannot duplicate it.
- Added provider-scope lineage to branch projections. Provider ownership boundaries no longer render as edit branches; inactive-provider turn IDs and branch navigation are withheld in both full and paged snapshots while active-target edit controls remain usable.
- Chat now consumes the durable switch query after reload, shares the TUI four-stage progress track, presents recovery/success states, keeps recovery action available, and makes the agent conversation surface inert while ownership is unresolved. A ready structured controller is the Chat takeover proof instead of a terminal handle.

Verification:
- env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_RUN_FILE npm run lint
- go test -race ./internal/service/chat ./internal/session_manager ./internal/storage/sqlite/store
- focused backend packages: service/chat, session_manager, sqlite/store, lifecycle, daemon
- frontend typecheck
- 123 focused renderer tests
- Full frontend test discovery still has the unrelated landing Markdown generator baseline failure because cheerio is absent from the checkout; the touched renderer suites pass.

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,716 @@
...
+	resumeCandidate, resumable, err := m.findTargetResumeCandidate(
+		ctx, store, rec, cfg.TargetHarness, targetAgent, targetConfigDir)
+	if err != nil {
+		return result, fmt.Errorf("switch Chat agent %s: target resume candidate: %w", id, err)
+	}
+	targetStartMode := domain.AgentSwitchTargetStartFresh
+	providerConversationID := ""
+	if resumable {
+		targetStartMode = domain.AgentSwitchTargetStartResumed
+		providerConversationID = resumeCandidate.NativeSessionID
```
illegalcall, line 167: **[P1] Reusing this native ID cannot be committed by the real SQLite store.** `ActivateChatAgentSwitchTarget` unconditionally inserts a new `conversation_branches` row with the resumed provider ID, while migration 0087 enforces `UNIQUE (conversation_id, provider_conversation_id)`. On A→B→A, A's original branch already owns this ID, so activation deterministically fails after the target has been resumed. The new switch-back test uses the fake store and cannot exercise this constraint. Please represent repeated provider ownership epochs safely (using an additive migration if the schema changes) and add an A→B→A regression through real SQLite.

illegalcall (commented):
## Verdict: Changes requested

The latest revision fixes the prior post-commit ownership, automation-authorship, provider-boundary presentation, and durable Chat progress/input-lock issues. Two native switch-back correctness issues remain: the real SQLite activation cannot commit a reused provider conversation ID, and a user-selected target model is dropped on the resume path.

Tests were not run (review-only policy).

backend/internal/session_manager/agent_switching_chat.go
```diff
@@ -0,0 +1,716 @@
...
+
+	_, err = m.chat.StartChat(ctx, ChatStart{
+		SessionID:               id,
+		ProjectID:               rec.ProjectID,
+		Kind:                    rec.Kind,
+		Harness:                 cfg.TargetHarness,
+		DataDir:                 m.dataDir,
+		WorkspacePath:           rec.Metadata.WorkspacePath,
+		Env:                     targetEnv,
+		Model:                   agentConfig.Model,
```
illegalcall, line 299: **[P2] The selected target model is silently ignored on this resume path.** When `ProviderConversationID` is nonempty, Chat Service calls `driver.Resume` with `ChatResumeConfig`, which has no model field; neither Codex `thread/resume` nor ACP resume receives `agentConfig.Model`. Activation then resets the durable conversation overrides, so the activation turn also carries no model override. A model chosen in `SwitchAgentDialog` therefore works for a fresh target but not when switching back to a retained native session. Thread the model through resume/application (or disallow that choice for resume) and cover an explicit-model switch-back.

illegalcall:
Addressed review 4963900707 in e021fafd2.

- Native A→B→A ownership: migration 0100 removes the incorrect per-conversation provider-ID uniqueness while preserving provider IDs as indexed data and branch IDs as ownership-epoch identity. The rebuild preserves the root provider-binding trigger. A real-SQLite agent-switch regression now completes A→B→A with two distinct ownership boundaries for the retained A native ID, and a migration test covers upgrade data, trigger behavior, duplicate epochs, and foreign-key integrity.
- Selected resume model: `ChatResumeConfig` now carries `Model`; Chat service forwards it on normal, edit-fork, and branch resume paths; ACP forwards it into launch config and reapplies it through session settings; Codex sends it on `thread/resume`. Manager, service, ACP, and Codex tests cover the explicit switch-back selection.

Verification: affected package tests pass; race tests pass for both drivers, Chat service, session manager, SQLite, and SQLite store; `go vet` passes; golangci-lint reports 0 issues. The broad `go test ./...` run passed every changed package and exposed only five unrelated existing CLI spawn fixture 404 failures.

backend/internal/session_manager/agent_switching_chat.go
illegalcall: Fixed in e021fafd2. Migration 0100 makes the branch ID the ownership-epoch identity and allows a retained provider conversation ID to appear in multiple A→B→A boundary nodes. The table rebuild preserves the root-binding trigger and adds a non-unique provider lookup index. Real-SQLite switch-back and upgrade regressions cover the duplicate-ID path and lineage.

backend/internal/session_manager/agent_switching_chat.go
illegalcall: Fixed in e021fafd2. The selected target model now crosses `ChatResumeConfig`, Chat service, ACP launch/session configuration, and Codex `thread/resume`. Integrated switch-back plus service and adapter regressions verify the requested model reaches a resumed native target.

illegalcall (commented):
## Verdict: Changes requested

## Standards

No high-confidence standards violations. The new additive migration follows the repository's established table-rebuild pattern and the real SQLite regression now covers repeated provider ownership epochs.

## Spec

The prior SQLite A→B→A and resumed-model blockers are fixed. One Chat/TUI parity gap remains: an asynchronously failed switch is not retained or presented after the active summary clears, leaving the user without the required durable error/retry state.

**Summary:** 0 Standards findings; 1 Spec finding (silent terminal switch failure).

Tests were not run (review-only policy).

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -110,6 +121,36 @@ export function SessionChatSurface({
 	const [switchSelectorOpen, setSwitchSelectorOpen] = useState(false);
 	const [switchSelectorContainer, setSwitchSelectorContainer] = useState<HTMLDivElement | null>(null);
 	const switchMutation = useSwitchAgentState(session.id);
+	const agentSwitches = useAgentSwitches(session.id).data ?? [];
+	const durableAgentSwitch = selectDurableAgentSwitch(session.activeAgentSwitch, agentSwitches);
```
illegalcall: **[P2] Retain terminal switch failures instead of silently dropping them.** This selection ultimately falls back only to recovery/nonterminal rows, while the POST mutation has already succeeded at admission and closes the dialog. When the worker later marks the saga `failed`, the workspace refresh clears `session.activeAgentSwitch`; the terminal history row is then excluded by `selectDurableAgentSwitch`, so both this status and `switchMutation.error` disappear. The user gets no explanation or retry context. `CenterPane` deliberately retains an observed terminal row and supports dismissing its failure; mirror that behavior for Chat and add a test where an admitted nonterminal switch transitions to `failed` after the active workspace summary clears.

illegalcall: Addressed in 0a0cd01a5. Chat now remembers switch IDs it observed while nonterminal and retains the matching durable failed row after activeAgentSwitch clears. The failure card shows the durable error, can be dismissed independently, and the switch control keeps the failure presentation so retry context remains available. Added a regression covering starting_target -> failed after the workspace summary clears, including dismissal and retained retry state. Verified 65 focused/adjacent renderer tests, frontend typecheck, and E2E typecheck.

illegalcall (commented):
## Verdict: Changes requested

## Standards

The direct nonterminal→failed retention path is fixed, but the copied Chat presentation lifecycle has diverged from `CenterPane`: it does not let a newer completed retry supersede an older observed failure. The duplicated state machine is a non-blocking maintainability smell; the resulting stale outcome is a correctness issue.

## Spec

A failure followed by a successful retry still ends with the old failure resurfacing or remaining on the toolbar. This violates the required actionable retry/outcome behavior and Chat/TUI parity.

**Summary:** 1 underlying correctness finding across both axes; worst issue is stale failure state after a successful retry.

Tests were not run (review-only policy).

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -109,7 +122,63 @@ export function SessionChatSurface({
...
+		setDismissedFailureSwitchId(undefined);
+	}, [session.id]);
 	const switchMutation = useSwitchAgentState(session.id);
+	const agentSwitches = useAgentSwitches(session.id).data ?? [];
+	const activeHistorySwitch = findActiveAgentSwitch(agentSwitches);
+	if (session.activeAgentSwitch && !isTerminalAgentSwitch(session.activeAgentSwitch)) {
+		observedNonterminalSwitchIdsRef.current.add(session.activeAgentSwitch.id);
+	}
+	if (activeHistorySwitch) observedNonterminalSwitchIdsRef.current.add(activeHistorySwitch.id);
+	const observedFailedSwitch = agentSwitches.find(
```
illegalcall: **[P2] Supersede an observed failure when a newer retry completes.** History is newest-first. After switch A is observed and fails, retry B takes precedence while active; once B becomes `completed` and `session.activeAgentSwitch` clears, `selectDurableAgentSwitch` excludes terminal B and this search skips B because it is not failed, then returns old A. The old failure card therefore resurfaces, or—if A was dismissed—the toolbar still receives A's failure presentation. `CenterPane` avoids this with `latestCompletedSwitch` precedence. Apply equivalent terminal ordering/clearing here and add a failure→retry→completed regression.

illegalcall: Addressed in ca2a0f392. Chat now gives the newest completed history row the same precedence as CenterPane, ahead of any older observed failure. An observed settled retry gets the existing three-second success presentation, after which only the card clears; the switch control continues to reflect the successful latest outcome instead of reviving the prior warning. Expanded the regression through failure A -> dismissed failure -> active retry B -> completed B after activeAgentSwitch clears, and verified both card and toolbar state plus success settlement. Verified 78 adjacent renderer tests, frontend typecheck, and E2E typecheck.

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -109,7 +122,87 @@ export function SessionChatSurface({
...
+		(entry) =>
+			entry.state === "failed" && observedNonterminalSwitchIdsRef.current.has(entry.id),
+	);
+	const latestCompletedSwitch =
```
illegalcall: **[P1] Do not reactivate an arbitrary historical completion as an in-flight switch.** After any successful switch, this row can remain first in history indefinitely. If that Chat controller later stops or crashes and the user reopens the session, there is no active switch but `latestCompletedSwitch` is still selected; because the snapshot controller is not ready, `deriveAgentSwitchPresentation` turns the old completion into `in_progress`, `switchLocksChat` makes the conversation panel inert (including its Resume/Open shell controls), and the confirming-takeover overlay cannot clear without an external controller restart. The switch was already durably completed; the stopped controller is ordinary recovery. Use the completion only to supersede the older observed failure, or otherwise require evidence that it belongs to the current UI transition, and add a cold-reload test with historical `completed` history plus a stopped controller.

illegalcall (commented):
## Verdict: Changes requested

## Standards

The live failure→retry→success ordering is fixed. The new completed-history fallback, however, is not scoped to a current switch lifecycle and can permanently disable normal Chat recovery. The duplicated lifecycle state machine between Chat and `CenterPane` remains a non-blocking divergence smell and has now produced another parity regression.

## Spec

A durable historical success must not be reinterpreted as an unfinished takeover merely because that controller later stopped. Doing so replaces the ordinary Chat resume state with an undismissable switch lock.

**Summary:** 1 blocking correctness finding across both axes; worst issue is a previously switched Chat session becoming inert after its controller stops.

Tests were not run (review-only policy).

frontend/src/renderer/components/chat/SessionChatSurface.tsx
illegalcall: Addressed in 3605f0f03. Chat no longer selects the newest completed row unconditionally. It considers only the newest terminal history row and presents it only when this mounted Chat view observed that exact switch nonterminal; an unobserved terminal row is historical and suppresses, rather than revives, older observed outcomes. A cold reload with completed history and a stopped controller now has no switch overlay, leaves Chat input enabled, and gives the switch control no stale outcome. The live failure -> retry -> completion regression remains covered, including transient success settlement. Verified 79 adjacent renderer tests, frontend typecheck, and E2E typecheck.

frontend/src/renderer/components/chat/SessionChatSurface.tsx
```diff
@@ -109,7 +122,89 @@ export function SessionChatSurface({
...
+	const agentSwitches = useAgentSwitches(session.id).data ?? [];
+	const activeHistorySwitch = findActiveAgentSwitch(agentSwitches);
+	if (session.activeAgentSwitch && !isTerminalAgentSwitch(session.activeAgentSwitch)) {
+		observedNonterminalSwitchIdsRef.current.add(session.activeAgentSwitch.id);
+	}
+	if (activeHistorySwitch) observedNonterminalSwitchIdsRef.current.add(activeHistorySwitch.id);
+	// History is newest-first. A terminal row is presentation state only when this
+	// mounted Chat view observed that same switch in flight; otherwise it is past
+	// history and merely prevents an older observed outcome from resurfacing.
+	const latestTerminalSwitch = agentSwitches.find(isTerminalAgentSwitch);
```
illegalcall: **[P2] Retire the switch after its successful lifecycle settles**

Once this mount has observed a switch in flight, its ID remains in `observedNonterminalSwitchIdsRef` forever, so this lookup keeps selecting the completed row even after the 3-second success presentation ends. If the target controller later crashes or becomes `stopped` without a remount, `targetChatControllerReady` becomes false and `deriveAgentSwitchPresentation` reclassifies that old completion as `in_progress/confirming_takeover`. `switchLocksChat` then makes the whole conversation panel inert—including the **Resume agent** and **Open shell** controls for the unrelated controller failure—so recovery requires reloading the UI. Consume or mark this completion as settled after success (while still suppressing older failures), and cover nonterminal → completed/ready → later stopped on the same mount.

illegalcall (commented):
## Review result: Changes requested

A completed Chat switch is still retained beyond its live lifecycle, which can permanently disable the controller recovery UI if that controller later stops. Please address the inline finding and add same-mount lifecycle coverage.

Tests were not run, per the review-only task policy.

frontend/src/renderer/components/chat/SessionChatSurface.tsx
illegalcall: Addressed in ab1190c82. I moved the observed-switch outcome bookkeeping into a shared `useObservedAgentSwitchLifecycle` hook used by both `SessionChatSurface` and `CenterPane`. The hook now owns observation, newest-terminal precedence, dismissal, the transient success window, and—critically—retires the switch ID when that success window ends. Retired completions can no longer be selected or re-derived as `confirming_takeover` after a later controller stop, while the newest terminal row still suppresses older failures. I added same-mount regressions for nonterminal → completed/ready → success expiry → later stopped in both Chat and CenterPane, plus direct lifecycle-hook coverage. Verified with all renderer tests (122 files / 1,798 tests), frontend typecheck, and E2E typecheck.

frontend/src/renderer/hooks/useObservedAgentSwitchLifecycle.ts
```diff
@@ -0,0 +1,128 @@
...
+			: undefined;
+	const dismissedFailureSwitchId =
+		dismissedFailure && dismissedFailure.sessionId === sessionId
+			? dismissedFailure.switchId
+			: undefined;
+	// Retirement is what prevents a later, unrelated controller outage from
+	// reinterpreting this completed history row as a takeover still in progress.
+	useEffect(() => {
+		if (!transientSuccessSwitchId) return;
+		const timer = window.setTimeout(() => {
```
illegalcall, line 133: **[P2] Retire lifecycle ownership independently of the success timer**

This timer currently owns both the transient toast and the logical retirement of the completed switch. Until it fires, a controller that becomes `stopped` makes the still-observed completion derive as `in_progress` again and re-locks the Chat/terminal recovery UI. Also, changing sessions cleans up this effect and cancels the retirement while `transientSuccess` is retained; A → B → A therefore reactivates A's old success and starts another timer. Retire the switch as soon as settlement is established, while retaining a separate immutable success notice for three seconds, and cover both stop-before-timer and A → B → A. The current component tests only stop the controller after advancing the timer.

illegalcall (commented):
## Review result: Changes requested

The prior permanent controller-recovery lock is improved, but logical switch settlement is still coupled to a cancellable presentation timer. This leaves two reproducible lifecycle regressions; see the inline finding.

Tests were not run, per the review-only task policy.

frontend/src/renderer/hooks/useObservedAgentSwitchLifecycle.ts
illegalcall: Addressed in 97e5dcdea. `settle()` now retires logical switch ownership synchronously, before starting any timer. It separately captures cloned agent-switch/presentation data as an immutable success notice; the 3-second effect only expires that notice, so cleanup can no longer cancel retirement and controller health cannot re-derive it. Notices and dismissals are scoped to a monotonically increasing session-visit epoch, preventing A → B → A from replaying A’s stale notice. Both surfaces also filter every durable selection path (including a lagging workspace `activeAgentSwitch`) through the retired-ID set while allowing the snapshot notice to remain visible without locking recovery controls. Added direct A → B → A hook coverage plus Chat and CenterPane stop-before-timeout regressions, including a stale active summary. Verified all renderer tests (122 files / 1,799 tests), frontend typecheck, and E2E typecheck.

illegalcall (commented):
## Review result: Ready to merge

The latest change resolves the remaining lifecycle issues: settlement now retires switch ownership immediately, the success notice is an immutable non-locking snapshot, and visit scoping prevents A → B → A notice replay. I found no remaining high-confidence Standards or Spec issues in the PR.

Tests were not run, per the review-only task policy.

illegalcall:
Current-main refresh is complete at `a8d3dc899`.

- GitHub Actions: **10/10 passed** (Go build/test, lint, API drift, frontend, renderer smoke, gitleaks, and CLI E2E on Linux/macOS/Windows/container).
- Local verification: full backend, focused race suites, isolated-cache golangci-lint, SQLC drift, frontend and E2E typechecks, full Vitest (192 files / 2,529 passed / 1 skipped), and the agent-switch Playwright scenario all passed.
- Evidence policy: the live Electron A/B images are hosted separately; **zero screenshot or evidence assets appear in Files Changed**.
- Integration check: the current #4121 and #4092 heads each merge cleanly with this refreshed head.

This is ready for review/merge. The only #4033 work intentionally outside this PR is Gap 3, the non-P1 shared tab-strip refactor.

nikhilachale (approved):
Manual test note: usage/cost display is confusing after switching a worker from Claude to Codex. Chat Summary shows total tokens increasing from ~66.3K to ~202.5K, with Claude ~66.3K and Codex ~136.2K, but this is hard to reconcile in Chat UI after the switch. TUI usage appears checkable/correct. Expected Chat UI to preserve clear per-agent/per-model usage attribution across handoff.

Image
Image

nikhilachale (approved):
tested locally working finee 

Image
Image

nikhilachale (changes requested):
Agent/interface switching fails when the resumed Claude history ends with an interrupted AO handoff message. ACP marks the final user-only turn as permanently unsettled, causing TUI → Chat UI to fail with TARGET_HISTORY_UNSETTLED after 45 seconds.

Image

Manual testing found an agent/interface switching edge case.

### Steps to reproduce

1. Start a Claude worker in Chat UI.
2. Switch from Claude to Codex.
3. Switch back from Codex to Claude.
4. Interrupt the automatic Claude handoff turn before Claude produces a response.
5. Switch the session from TUI to Chat UI.

### Observed

The interface switch waits approximately 45 seconds and then fails with:

`TARGET_HISTORY_UNSETTLED: wait for settled native conversation history: chat conversation history is not settled: context deadline exceeded`

The automatic handoff message remains the final user message without provider output. The ACP history reader treats this as permanently unsettled, so retrying the interface switch continues to fail.

### Expected

AO should recover from an interrupted automatic handoff turn and allow the interface/agent switch without requiring the user to manually produce another completed Claude turn.

### Workaround

Send another message in the Claude TUI, wait for Claude to complete its response, then retry switching to Chat UI.

Test session: `agent-orchestrator-223`

vanshx999:
Tested on the latest branch:

  - Codex Chat → Claude Chat
  - Claude Chat → Codex Chat
  - Codex Chat → Codex TUI
  - Codex TUI → Codex Chat

All transitions completed successfully with no fallback or input lock. I could not reproduce the reported TARGET_HISTORY_UNSETTLED failure through the normal switching flow.

nikhilachale (approved):

PR #4100, opened 2026-08-18
fix(inspector): remember a session's inspector tab across session switches

Vaibhaav-Tiwari:
## What

Fixes the inspector rail's active tab (Summary/Browser/Files) resetting to Summary when switching back to a previously-visited session, discarding whatever tab the user had left it on.

## Why

Reported symptom: work on the Files (FILE DIFF) or Browser tab in one session, switch to another session, then switch back — the previous tab selection is gone and Summary shows instead. The right-panel state should persist per session across session switches.

Root cause: the session-entry effect in `SessionView.tsx` that defaults a **brand-new** session to the Summary tab (added in #3864) tracked only the single most-recently-initialized session id (`initializedInspectorSessionIdRef`, a plain `string | null` ref). Re-entering *any* session that wasn't the immediately preceding one looked identical to a first-ever visit, so the effect unconditionally forced it back to Summary via `setInspectorViewForSession(sessionId, "summary")` — even though that session's tab selection was already correctly persisted per-session in the ui-store (`inspectorSessions[sessionId].view`).

## How

Track every session id this view has ever initialized in a `Set<string>` instead of a single last-seen id, so the "default new sessions to Summary" effect only fires once per session's *lifetime*, not once per "was this the last session I looked at." This mirrors the existing `current?.browserContentRevealed === undefined` guard two lines below in the same effect, which already handled this correctly for that field.

No changes to the native browser-view (WebContentsView) visibility plumbing — investigated that path first (it's per-session and correctly scoped) before finding the actual bug was purely in this React-level tab-selection effect.

**Update from review (@neversettle17-101):** the component-local `Set` was itself recreated whenever `SessionView` unmounts (e.g. across a route transition), so re-mounting an already-visited session still looked like a first-ever visit. Moved the "initialized" marker into the ui-store's persisted `inspectorSessions[sessionId]` state via a new, idempotent `initializeInspectorSession(sessionId, hasBrowserContent, hasInspector)` action (no-ops once `initialized` is already true), so it survives unmount/remount, not just re-renders of one mounted instance. `hasInspector` is threaded through in addition to the reviewer's suggested signature so sessions without an inspector (orchestrators) still never gain a store entry, matching the prior behavior.

## Testing

All testing below was run locally in the worktree at `.claude/worktrees/fix-browser-view-session-switch-race` (branch `fix/inspector-tab-resets-on-session-revisit`).

- `remembers the tab a session was left on when returning to it after visiting another session` (`SessionView.test.tsx`) — opens Files on session A, switches to session B, switches back to A, asserts Files is still selected. Confirmed it fails against the pre-fix code (`data-view="summary"` instead of `"files"`) and passes after.
- `remembers the tab a session was left on after unmounting and remounting the view` (`SessionView.test.tsx`) — selects Files on session A, unmounts `SessionView`, remounts the same session, asserts Files is still selected. Added per review feedback; independently reverted just the source (keeping the test) to confirm it fails against the component-local-ref version (`data-view="summary"`) and passes against the store-backed version.
- `npx vitest run --config vite.renderer.config.ts src/renderer/components/SessionView.test.tsx src/renderer/components/SessionInspector.test.tsx src/renderer/hooks/useBrowserView.test.tsx` — 185/185 pass.
- `npx tsc --noEmit -p .` — clean.
- Full frontend suite: only pre-existing, unrelated failures (macOS-only path assertions running on Windows, landing-page/analytics tests, network-socket tests) — none touch the changed files.

## Checklist

- [x] Branched from `main`
- [x] One focused change
- [x] Relevant CI checks pass for the area touched (frontend)

### DIFF
--- a/frontend/src/renderer/components/SessionView.test.tsx
+++ b/frontend/src/renderer/components/SessionView.test.tsx
@@ -1381,6 +1381,25 @@ describe("SessionView", () => {
 		expect(inspectorButton()).toHaveAttribute("data-view", "files");
 	});
 
+	// Regression: the "initialized" marker used to live in a component-local
+	// ref, which is recreated whenever SessionView unmounts. Remounting an
+	// already-visited session (e.g. across a route transition) then looked
+	// identical to a first-ever visit and forced the tab back to Summary. The
+	// marker now lives in the ui-store's persisted inspectorSessions state, so
+	// it survives unmount/remount, not just re-renders of one mounted instance.
+	it("remembers the tab a session was left on after unmounting and remounting the view", () => {
+		const view = render(<SessionView sessionId="sess-1" />);
+		expect(inspectorButton()).toHaveAttribute("data-view", "summary");
+
+		fireEvent.click(screen.getByRole("button", { name: "open files" }));
+		expect(inspectorButton()).toHaveAttribute("data-view", "files");
+
+		view.unmount();
+
+		render(<SessionView sessionId="sess-1" />);
+		expect(inspectorButton()).toHaveAttribute("data-view", "files");
+	});
+
 	it("keeps Summary selected when preview content arrives with the async workspace response", () => {
 		const secondWorker = workerSession("sess-2");
 		secondWorker.previewUrl = "http://localhost:5173/";
--- a/frontend/src/renderer/components/SessionView.test.tsx
+++ b/frontend/src/renderer/components/SessionView.test.tsx
@@ -1362,6 +1362,25 @@ describe("SessionView", () => {
 		expect(browserUnseen("sess-2")).toBe(false);
 	});
 
+	// Regression: the session-entry effect that defaults a brand-new session to
+	// Summary tracked only the single most-recently-initialized session ID, so
+	// re-entering ANY session that was not the immediately preceding one looked
+	// identical to a first-ever visit and forced it back to Summary — silently
+	// discarding whatever tab (Files, Browser) the user had left it on.
+	it("remembers the tab a session was left on when returning to it after visiting another session", () => {
+		const { rerender } = render(<SessionView sessionId="sess-1" />);
+		expect(inspectorButton()).toHaveAttribute("data-view", "summary");
+
+		fireEvent.click(screen.getByRole("button", { name: "open files" }));
+		expect(inspectorButton()).toHaveAttribute("data-view", "files");
+
+		rerender(<SessionView sessionId="sess-2" />);
+		expect(inspectorButton()).toHaveAttribute("data-view", "summary");
+
+		rerender(<SessionView sessionId="sess-1" />);
+		expect(inspectorButton()).toHaveAttribute("data-view", "files");
+	});
+
 	it("keeps Summary selected when preview content arrives with the async workspace response", () => {
 		const secondWorker = workerSession("sess-2");
 		secondWorker.previewUrl = "http://localhost:5173/";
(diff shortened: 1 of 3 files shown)


neversettle17-101 (commented):
This fixes the A → B → A prop-rerender case, but I don't think the component-local `Set` is the fully robust persistence boundary.

`initializedInspectorSessionIdsRef` is recreated whenever `SessionView` unmounts. On a subsequent mount of an already-visited session, the new set is empty and the session-entry effect will again set its view to Summary. That can occur across route transitions or any other unmount/remount path, so the selected Files/Browser tab still is not reliably per-session.

The remount-safe approach is to put the “initialized” marker in the existing `inspectorSessions[sessionId]` Zustand state (for example, `initialized?: boolean`) and expose an atomic `initializeInspectorSession(sessionId, hasBrowserContent)` action. It should set the Summary/baseline state only when that persisted-in-memory session entry has not yet been initialized; later calls should be a no-op. Then the layout effect can call that action on mount without maintaining a local ref.

Please also add a regression test that selects Files for a session, unmounts `SessionView`, remounts the same session, and verifies Files remains selected.

Vaibhaav-Tiwari:
Moved the "initialized" marker into the ui-store's persisted `inspectorSessions[sessionId]` state via a new `initializeInspectorSession(sessionId, hasBrowserContent, hasInspector)` action (86de3ccc9). It sets the Summary/baseline-reveal state only when that session's entry has not yet been marked `initialized`; subsequent calls (including after unmount/remount) are a no-op. `SessionView`'s layout effect now just calls this action on mount instead of maintaining a local ref.

Added a regression test (`remembers the tab a session was left on after unmounting and remounting the view`) that selects Files for a session, unmounts `SessionView`, remounts the same session, and asserts Files is still selected.

`npx vitest run src/renderer/components/SessionView.test.tsx src/renderer/components/SessionInspector.test.tsx src/renderer/hooks/useBrowserView.test.tsx` — 185/185 pass. `npx tsc --noEmit` — clean. Full suite — same pre-existing unrelated failures as noted in the PR description (landing-script tests missing the `cheerio` dep, network-socket tests, one ChatWorkspace timeout); none touch the changed files.

neversettle17-101:
@Vaibhaav-Tiwari also add local testing details

Vaibhaav-Tiwari:
Addressed in 86de3ccc9.

Confirmed the gap: `initializedInspectorSessionIdsRef` lived in a component-local ref, recreated on every `SessionView` unmount — a remount of an already-visited session (route transition, etc.) looked identical to a first-ever visit and reset the tab to Summary.

Moved the marker into `inspectorSessions[sessionId].initialized` in the ui-store, behind a new `initializeInspectorSession(sessionId, hasBrowserContent, hasInspector)` action that's a no-op once `initialized` is already true — so it survives unmount/remount, not just re-renders of one mounted instance. (Threaded `hasInspector` through in addition to your suggested signature so sessions without an inspector, e.g. orchestrators, still never gain a store entry — matching the prior behavior and the existing `sess-orch` regression test.)

Added the requested regression: `remembers the tab a session was left on after unmounting and remounting the view` — selects Files, unmounts `SessionView`, remounts the same session, asserts Files is still selected. Verified it fails against the component-local-ref version and passes against the store-backed one.

Full details and local testing (worktree path, exact commands, results) are now in the PR description.

neversettle17-101 (approved):

Issue #2606, opened 2026-07-11
Browser annotations stay queued when the agent is idle


codebanditssss:
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/33d95077-8b23-4be6-9fe8-b9960ab8fa21" />

## Bug

  Preview annotations can stay stuck in `Queued` and never reach the agent.

  ## Steps

  1. Open a session preview.
  2. Add and send an annotation.
  3. The annotation remains queued.

  ## Cause

  Annotation delivery waits for specific session states. These states can be `idle` or stale as `working`, even when the agent can accept input.

  The send endpoint does not require this restriction.

  ## Fix

  Send annotations without waiting for session activity state. Keep sequential delivery and retry behavior.

  ## Environment

  Fedora Linux, Electron dev build.

  Related: #2548

PR #4101, opened 2026-08-18
fix(browser-annotation): queue submissions until the agent is idle

Vaibhaav-Tiwari:
## Summary
- Browser-annotation change requests were delivered as soon as the previous one's HTTP call resolved, without waiting for the agent to actually finish processing it — a follow-up annotation submitted while the agent was still mid-turn landed straight in the live pane and collided with the in-progress work ("overtook" it).
- `useBrowserAnnotationQueue` now gates draining on the session's real-time `activity.state` (via the existing `isAgentActivityWorking` helper): a queued annotation waits until the agent leaves `active` before sending, with a bounded 20s fallback so the queue can never get stuck forever if that signal doesn't arrive — the exact failure mode an earlier fix for #2606 hit when it gated on the coarser, staler `session.status` field instead.
- Wired the live `session.activity` into the real production call site in `SessionView.tsx`.

## Test plan
- [x] `vitest run src/renderer/components/BrowserPanel.test.tsx` — 38/38 passed (includes 6 new tests: hold-while-active → send-on-idle, send-on-waiting_input, 20s fallback firing, single fallback timer for multiple queued items, no stray send after switching sessions mid-wait)
- [x] `vitest run src/renderer/components/SessionView.test.tsx` — 66/66 passed
- [x] `tsc --noEmit` — clean
- [ ] Manual verification in the desktop app (submit a browser annotation mid-turn, confirm it queues instead of colliding, and sends once the agent is idle/waiting-input)

### DIFF
--- a/frontend/src/renderer/components/BrowserPanel.test.tsx
+++ b/frontend/src/renderer/components/BrowserPanel.test.tsx
@@ -778,6 +778,167 @@ describe("BrowserPanel", () => {
 		expect(postMock).toHaveBeenCalledTimes(1);
 	});
 
+	it("holds a queued annotation while the agent is actively working, then sends once it goes idle", async () => {
+		hookState.navState = { ...hookState.navState, url: "http://localhost:5173/" };
+		const { rerender } = render(
+			<BrowserPanel
+				active
+				onTogglePopOut={() => undefined}
+				poppedOut={false}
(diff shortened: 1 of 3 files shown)


i-trytoohard:
@Vaibhaav-Tiwari Marking as stale: no activity from you for 20 days. Push a commit or drop a comment and it comes off the stale list. After 30 days of quiet it gets closed as stale (the branch stays safe to reopen).

PR #4102, opened 2026-08-18
feat: add --model to ao spawn with per-session scoping and harness validation

neversettle17-101:
Implements #3331.

### What changed
- Added "Model" to session metadata, SQLite storage, and generated sqlc code (migration "0099_session_model.sql").
- Added the "--model" flag to "ao spawn" and threaded it through the spawn request DTO.
- Exposed the resolved model in "SessionView" and regenerated "openapi.yaml" + "frontend/src/api/schema.ts".
- Added harness-level model validation via modelcatalog.Base before a session is created; unsupported models return "UNSUPPORTED_MODEL".
- Persisted the resolved model in TUI and chat spawn metadata and passed it to the runtime launch config.
- Added/updated tests for CLI wiring, DTO drift, harness validation, persistence, and session get reporting.

### Verification
- "go build ./..." passes.
- "go test ./internal/session_manager/... ./internal/httpd/... ./internal/storage/sqlite/..." passes.
- "go test ./internal/cli/... -run 'TestSpawn|TestSession|TestDTO'" passes.
- golangci-lint passes.
- "npm run frontend:typecheck" passes.
- "npm run lint" only fails on the pre-existing "TestBrowserRequiresSessionAndValidWait" environmental failure in the CLI package, which is unrelated to this change.

### Notes
- Local CLI tests are sensitive to "AO_PROJECT_ID", "AO_SESSION_ID", and "AO_RUNTIME_LAUNCH_ID" env vars; all verification above was run with those unset.


backend/internal/httpd/controllers/sessions.go
```diff
@@ -270,7 +270,7 @@ func (c *SessionsController) spawn(w http.ResponseWriter, r *http.Request) {
 		envelope.WriteAPIError(w, r, http.StatusBadRequest, "bad_request", attachErr.code, attachErr.message, nil)
 		return
 	}
-	sess, promptBytes, systemPromptBytes, err := c.Svc.Spawn(r.Context(), ports.SpawnConfig{ProjectID: in.ProjectID, IssueID: in.IssueID, TrackerProvider: in.TrackerProvider, Kind: in.Kind, Harness: in.Harness, Branch: in.Branch, RequestedMode: in.Mode, Prompt: in.Prompt, DisplayName: displayName, Attachments: attachments})
+	sess, promptBytes, systemPromptBytes, err := c.Svc.Spawn(r.Context(), ports.SpawnConfig{ProjectID: in.ProjectID, IssueID: in.IssueID, TrackerProvider: in.TrackerProvider, Kind: in.Kind, Harness: in.Harness, Branch: in.Branch, RequestedMode: in.Mode, Prompt: in.Prompt, DisplayName: displayName, Attachments: attachments, AgentConfig: ports.AgentConfig{Model: in.Model}})
```
neversettle17-101, line 273: This always maps the spawn override into AgentConfig.Model. That works for adapters whose launch command reads Config.Model, but Amp exposes its selectable values as modes and its adapter only reads Config.Mode to append --mode. Because validateSpawnModel accepts Amp values like high from the catalog, ao spawn --agent amp --model high will succeed, persist model=high, and then launch Amp without --mode high, silently using the default. Please route list-mode selections for Amp into Mode (or otherwise make the adapter consume this override) and add a test that inspects the actual launch config/command for Amp.

neversettle17-101 (commented):
Needs changes. I found one functional regression in the new --model plumbing for fixed-list harnesses like Amp: the override is validated and persisted as Model but never converted to the adapter field that Amp actually launches with.

backend/internal/session_manager/manager.go
```diff
@@ -900,6 +917,7 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 		Prompt:                    prompt,
 		LatestUserPrompt:          prompt,
 		BrowserCapabilityVerifier: browserCapabilityVerifier,
+		Model:                     agentConfig.Model,
```
neversettle17-101: After normalizeAgentConfigForHarness routes Amp values by setting cfg.Mode and clearing cfg.Model, this persists an empty metadata Model for `ao spawn --agent amp --model high`. That makes the API/session view report no model (which means agent default) even though the session launched with `--mode high`, and it also contradicts the new metadata comment that Model stores the resolved per-spawn value. Please preserve the user-visible resolved selection separately from the adapter launch config (for example, keep metadata.Model as the original effective model while passing a normalized copy to the adapter) and cover the Amp case in the persistence/session-view tests.

neversettle17-101 (commented):
Needs changes. The Amp launch fix now routes the override into Config.Mode, but the same normalized config is also used as the persisted session metadata, so mode-list selections disappear from SessionView even though the session was explicitly spawned with that value.

backend/internal/session_manager/manager.go
neversettle17-101: For mode-list harnesses, normalizeAgentConfigForHarness moves the effective Model into Mode and clears Model before this metadata is built. That means `ao spawn --agent amp --model high` launches with `--mode high`, but persists `Metadata.Model` as empty and SessionView reports agent default instead of the explicit per-session selection. Please keep the user-visible resolved selection separate from the adapter-normalized launch config (or persist the original effective model here) and add Amp coverage for persistence/session-view output.

neversettle17-101 (commented):
Needs changes. The Amp launch normalization fixes the argv path, but it also clears the field that is later used as the persisted/read-model value, so mode-list selections are no longer reported as the resolved session model.

neversettle17-101 (commented):
Ready to merge from my review. The latest update preserves the user-visible resolved model while passing normalized adapter config to mode-list harnesses, and I did not find additional correctness issues in the reviewed diff.

backend/internal/session_manager/manager.go
```diff
@@ -900,6 +917,9 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 		Prompt:                    prompt,
 		LatestUserPrompt:          prompt,
 		BrowserCapabilityVerifier: browserCapabilityVerifier,
+		// agentConfig still holds the user-visible resolved Model; adapterConfig
+		// may have moved it into Mode for mode-list harnesses like Amp.
+		Model: agentConfig.Model,
```
neversettle17-101: This only persists `agentConfig.Model`. For mode-list harnesses like Amp, existing project/role defaults are stored in `AgentConfig.Mode` (e.g. `ao project ... --worker-agent amp --worker-mode high`), so a spawn without an explicit `--model` will launch with `Mode=high` via `adapterConfig` but persist an empty `Metadata.Model`. The new SessionView then says agent default even though the resolved per-session selection is high. Please derive the user-visible resolved selection from `Mode` for `ModelSelectionModeList` harnesses as well, and add coverage for an Amp project/role default mode, not just the explicit spawn model override.

neversettle17-101 (commented):
Needs changes. I found one remaining mismatch for mode-list harnesses: project/role default modes launch correctly but are still reported as agent default in the new session model field.

backend/internal/session_manager/manager.go
neversettle17-101: Fixed in 415c5e7. Added `resolvedModelForMetadata` helper: explicit `Model` overrides still win, but for `ModelSelectionModeList` harnesses the persisted `metadata.Model` now falls back to the resolved `Mode`. Added `TestSpawnAmpProjectDefaultModePersistsAsModel` covering an Amp project/role default mode so SessionView reports the resolved selection instead of agent default.

neversettle17-101 (commented):
Addressed the project/role default mode mismatch for mode-list harnesses in 415c5e7.

- Added `resolvedModelForMetadata` in `backend/internal/session_manager/manager.go`: explicit `--model` overrides still win, but when Model is empty and the harness catalog is a mode list (e.g. Amp), the persisted `metadata.Model` now falls back to the resolved `Mode`.
- Added `TestSpawnAmpProjectDefaultModePersistsAsModel` covering an Amp project/role default mode: verifies the launch argv contains `--mode high` and that `metadata.Model` is persisted as `high`.

All review threads for this finding have been resolved.

neversettle17-101 (commented):
Ready to merge from my review. The latest changes preserve the resolved selection for both explicit Amp --model overrides and existing Amp project/role default modes, while still passing the normalized adapter config for launch. I did not find additional blocking issues in the reviewed diff.

nikhilachale (approved):
No blocking found , LGTM

PR #4071, opened 2026-08-17
fix(daemon): recover sessions without false startup failure

illegalcall:
Fixes #4064

## What changed

- Relaunch dead TUI and Chat controllers in their existing registered worktrees during startup instead of serially stashing, removing, recreating, and reapplying every worktree.
- Bound live-session reconciliation with four workers and reuse successful Chat capability probes for the daemon lifetime.
- Keep the durable capture/marker/teardown path as the safety fallback when an in-place relaunch cannot commit.
- Treat the Electron supervisor's 30-second threshold as slow progress while the child is still alive: use polite status semantics, keep startup details fresh, omit the misleading restart action, and transition automatically when readiness arrives.

## Why

The reported 4m36s startup was dominated by an O(N) destructive restore cycle. A machine reboot or native-runtime loss does not require AO to move a registered worktree at all: the durable session and worktree are already the state to recover. Relaunching the controller in place avoids expensive Git/antivirus work and removes the riskiest part of normal startup. The UI change also stops a healthy, progressing process from being labeled as failed if recovery still exceeds 30 seconds on slower machines.

## End-to-end verification

I launched the real Electron app from this checkout against an isolated `AO_DATA_DIR` and `AO_RUN_FILE`. The fixture used a real SQLite database, Git repository/worktrees, tmux sessions, and `ao agent-process supervise`; only the final model executable was a deterministic local stub so the drill made no provider calls.

1. Registered one project and spawned eight real sessions (`Recovery 01` through `Recovery 08`).
2. Added a unique uncommitted sentinel to every registered worktree and recorded each SHA-256 plus the original launch generation.
3. Stopped the app and killed only those eight runtime sessions, leaving all eight durable session rows live.
4. Restarted the real daemon. Resolver log to loopback listener: **1,174 ms**. All eight controller generations rotated; IDs and workspace paths stayed unchanged; all eight sentinel hashes were identical; no restore markers, stash operations, force-destroys, or fallback teardowns occurred.
5. Launched the real Electron supervisor behind a controlled startup gate. At **53 seconds**, the app still showed progress with live details, zero alert roles, and zero restart actions. Releasing the same supervised process cleared the banner automatically and loaded all eight sessions on the real board.

The local runtime drill used tmux on macOS. Windows ConPTY follows the same runtime contract; dead-runtime and Chat recovery, fallback safety, uncertain-commit handling, and worker concurrency are covered at the manager boundary.

### Slow startup after 53 seconds

AO shows a neutral slow-start status with live details and no restart action

[...]

PR #4105, opened 2026-08-18
fix(session): terminate interrupted spawns on boot reconcile

### LINKED ISSUE #4079: bug(orchestrator): stuck on "Preparing the orchestrator terminal" forever after the daemon is killed mid-spawn
### Summary

If the daemon dies while an orchestrator session is being created, the project's orchestrator is broken permanently. The Orchestrator tab shows "Preparing the orchestrator terminal" forever, on that launch and every launch after it. Restarting the daemon does not help, and neither does reinstalling, because the bad state is a row in `ao.db`.

Starting a session happens in two steps: AO writes the session row first, then creates the git worktree. If the daemon is killed between those two steps, the row survives with no workspace and no terminal. AO's cleanup code (`rollbackSpawnSeedRow`) only runs when a spawn returns an *error*, so a spawn that is killed mid-flight leaves the row behind.

That leftover row is harmless for a worker session, but not for an orchestrator. `EnsureOrchestrator` returns the newest **active** orchestrator instead of spawning a new one, and the leftover row counts as active. So every future click on the Orchestrator tab gets handed the same dead row. Boot reconcile does not clean it up either: `reconcileLive` returns early when `WorkspacePath == ""`, so the row is skipped on every restart.

This happened to me twice in two days, on two different projects, during normal use. I did not have to do anything unusual to hit it.

### Steps to reproduce

Scripted version, hit on the first attempt (uses a throwaway data dir, does not touch your real `~/.ao`):

```bash
mkdir -p /tmp/repro/{data,repo} && cd /tmp/repro/repo
git init -q -b main . && echo hi > README.md
git add . && git -c user.email=t@t -c user.name=t commit -qm init

AO_DATA_DIR=/tmp/repro/data AO_RUN_FILE=/tmp/repro/running.json AO_PORT=39201 ao daemon &

curl -s -X POST localhost:39201/api/v1/projects -H 'content-type: application/json' \
  -d '{"path":"/tmp/repro/repo","projectId":"repro","config":{"defaultBranch":"main","orchestrator":{"agent":"claude-code"},"worker":{"agent":"claude-code"}}}'

# start an orchestrator, then kill the daemon while the worktree is being created
curl -s -X POST localhost:39201/api/v1/orchestrators -H 'content-type: application/json' \
  -d '{"projectId":"repro"}' &
sleep 0.35
kill -9 $(python3 -c "import json;print(json.load(open('/tmp/repro/running.json'))['pid'])")
```

`defaultBranch` matters. Without it the local repo has no remote to resolve from, the spawn fails cleanly, and rollback deletes the row correctly — that is the healthy path, not this bug.

Result:

```
id       kind          is_terminated  runtime_handle_id  workspace_path  branch
repro-1  orchestrator  0              (empty)            (empty)         (empty)
```

Now restart the daemon on the same data dir and ask for the orchestrator again:

```
DB after reboot:              repro-1 | orchestrator | 0 | (no handle)
POST /api/v1/orchestrators:   {"orchestrator":{"id":"repro-1", ...}}
```

The dead row survives the restart and is handed straight back. In the desktop app that is the "Preparing the orchestrator terminal" screen.

In the app, the same thing happens if you open a project's Orchestrator tab and the daemon is hard-killed within about half a second (quit the app, force-quit it, or `kill -9` the daemon pid from `~/.ao/running.json`). A graceful stop is fine: the context cancels, the spawn errors, and rollback works as intended. The window is exactly as long as `git worktree add` takes, so a large repo makes it easier to hit.

### Expected behavior

Asking for a project's orchestrator either gives you a session with a working terminal, or fails with a visible error. A row that never got a workspace should not be able to block every future orchestrator for that project.

### Actual behavior

The Orchestrator tab shows "Preparing the orchestrator terminal. This can take a moment while AO creates the workspace and starts the agent." indefinitely. Nothing is actually being prepared; there is no work in flight. The daemon is healthy, the API answers, and worker sessions in the same project spawn and attach normally.

On my install, two orchestrators were in this state, from two projects:

```
id             kind          is_terminated  runtime_handle_id  workspace_path
aohack-161     orchestrator  0              (empty)            (empty)
omnicommand-1  orchestrator  0              (empty)            (empty)
aohack-164     worker        0              aohack-164         .../worktrees/aohack/aohack-164
```

`GET /api/v1/sessions` agrees: both orchestrators come back active with no `terminalHandleId` and no `branch`.

The reaper sees the problem but cannot act on it, so the daemon log carries this every 5 seconds indefinitely:

```
level=WARN msg="reaper: session has no runtime handle metadata, skipping" session=aohack-161
```

There is no way to recover from inside the app. The only fix I found is `POST /api/v1/sessions/<id>/kill` against the daemon by hand.

### Where the problem is

- `backend/internal/session_manager/manager.go` — `Spawn` calls `m.store.CreateSession(...)` before `m.createSessionWorkspace(...)`. Every error path calls `rollbackSpawnSeedRow` / `rollbackSeedSpawnWorkspace`; process death bypasses all of them.
- `backend/internal/service/session/service.go` — `EnsureOrchestrator`: `if len(existing) > 0 { return newestSession(existing), nil }`, where `existing` is any non-terminated orchestrator, working terminal or not.
- `backend/internal/session_manager/manager.go` — `reconcileLive`: `if rec.Metadata.WorkspacePath == "" || ... { return nil }` skips the row on every boot.
- `frontend/src/renderer/components/TerminalPane.tsx` — shows `terminal.preparingOrchestrator` whenever `handleId` is empty, which looks identical to a spawn that is genuinely still working.

### Not fixed by #4071

This is a different bug from #4064, and PR #4071 does not fix it. That PR changed how the reconcile pass *runs* (serial loop to a parallel worker pool) and how the supervisor reports a slow boot. The `WorkspacePath == ""` early return in `reconcileLive` is unchanged on `main` as of 3348277, so these rows are still skipped.

I verified this directly: built `main` at 3348277, reproduced with the script above, restarted, and the dead orchestrator was still returned by `POST /api/v1/orchestrators`.

The two bugs do feed each other, though. A `not_ready` banner invites the user to click "Restart daemon", and a daemon restarted mid-spawn is one way to create this state.

### Suggested fix

In `reconcileLive`, before the existing early return, treat "no workspace path **and** no runtime handle" as a spawn that died before its workspace existed, and mark it terminated. Both conditions are only true together before `createSessionWorkspace` returns, so a session that has a workspace but a dead runtime still takes the existing save-and-teardown path that preserves work. Reconcile runs before the API binds, so no in-flight spawn can be seen mid-write.

With that change applied to the same data dir from the repro:

```
WARN msg="reconcile: terminating interrupted spawn with no workspace or runtime" sessionID=repro-1
DB after boot:                repro-1 | orchestrator | 1  (terminated)
POST /api/v1/orchestrators:   {"orchestrator":{"id":"repro-2", ...}}
repro-2 → real runtime handle, real worktree, live tmux session
```

I have a branch ready with this change and a regression test, and am happy to open the PR. If you would rather fix it at the other end — having `EnsureOrchestrator` ignore orchestrators with no runtime handle when picking the active one — tell me and I will do it that way instead.

### AO version

Desktop app 0.12.6 (macOS). Also reproduced against `main` at 3348277.

### Environment

macOS 26.5.1 (Apple silicon), tmux 3.7b, git 2.50.1. 164 sessions in `ao.db`, 15 worktrees. `ao doctor` fully green.

Rishet11:
Fixes #4079

## What changed

- Boot reconcile now terminates a live session that has neither a workspace path nor a runtime handle, instead of skipping it. That pair is only reachable if the daemon died between `Spawn`'s seed-row write and `createSessionWorkspace`, so the row is an interrupted spawn with nothing to adopt, preserve, or restore.
- Everything else in `reconcileLive` is untouched: workspace present + runtime alive still adopts, workspace present + runtime dead still takes the save-and-teardown path that preserves work and records the restore marker.

The row matters because `EnsureOrchestrator` returns the newest *active* orchestrator rather than spawning one, so a single interrupted spawn pins a project to a session that has no terminal and never will — "Preparing the orchestrator terminal" on every launch, surviving restarts, with no way out from inside the app. The reaper logs `session has no runtime handle metadata, skipping` for it every 5s meanwhile.

Not covered by #4071: that PR changed how the reconcile pass runs and how the supervisor reports a slow boot, but the `WorkspacePath == ""` early return is unchanged, so these rows are still skipped. Reproduced against `main` at 3348277 before writing this.

I put the fix in reconcile rather than having `EnsureOrchestrator` skip handle-less orchestrators, since that hides the row instead of clearing it and leaves it counted as active everywhere else. Happy to switch if you'd rather.

## Verification

- `go test ./internal/session_manager/ ./internal/daemon/... ./internal/service/session/` — 793 passed, including the new `TestReconcileLive_InterruptedSpawnTerminated`
- `go test -race ./internal/session_manager/` — 482 passed
- `golangci-lint v2.12.2 run ./internal/session_manager/...` — 0 issues
- `go build ./...`, `go vet ./internal/session_manager/` — clean

End to end, one daemon built from `main` and one from this branch, against the same throwaway data dir (full repro script in #4079):

```
main @ 3348277, after restart
  POST /api/v1/orchestrators -> {"orchestrator":{"id":"repro-1"}}   dead row handed back

this branch, same data dir
  WARN "reconcile: terminating interrupted spawn with no workspace or runtime" sessionID=repro-1
  POST /api/v1/orchestrators -> {"orchestrator":{"id":"repro-2"}}   real handle, real worktree, live tmux
```

One caveat on the full `npm run lint` gate: `go test ./...` fails here in `adapters/agent/{kilocode,opencode}` (3s auth-probe deadlines, they pass in isolation), `gitdefault`, and `service/project` (git hitting `.git/index: Not a directory` under tmpdirs). All reproduce on a clean `main` on this machine and none are in a package this PR touches.

### DIFF
--- a/backend/internal/session_manager/manager.go
+++ b/backend/internal/session_manager/manager.go
@@ -2066,6 +2066,19 @@ func (m *Manager) reconcileLive(ctx context.Context, rec domain.SessionRecord) e
 		return err
 	}
 	projectKind := project.Kind.WithDefault()
+	// An interrupted spawn: the seed row was written, then the daemon died before
+	// the workspace or the runtime existed, so Spawn's rollback never ran. There
+	// is nothing to adopt, nothing to preserve, and nothing to restore — but the
+	// row still counts as an active session, and for an orchestrator that is
+	// load-bearing: EnsureOrchestrator returns the newest active orchestrator
+	// instead of spawning, so the project is pinned forever to a session with no
+	// terminal ("Preparing the orchestrator terminal…"). Terminate it here so the
+	// next request spawns a real one. Safe at boot: reconcile runs before the API
+	// serves, so no in-flight spawn can be observed mid-write.
+	if rec.Metadata.WorkspacePath == "" && runtimeHandle(rec.Metadata).ID == "" {
+		m.logger.Warn("reconcile: terminating interrupted spawn with no workspace or runtime", "sessionID", rec.ID)
+		return m.lcm.MarkTerminated(ctx, rec.ID)
+	}
 	if rec.Metadata.WorkspacePath == "" || (rec.Metadata.Branch == "" && projectKind != domain.ProjectKindScratch) {
 		return nil
 	}
--- a/backend/internal/session_manager/manager_test.go
+++ b/backend/internal/session_manager/manager_test.go
@@ -7477,3 +7477,32 @@ func (m *flipOnNudgeMessenger) Send(_ context.Context, _ domain.SessionID, msg s
 	}
 	return nil
 }
+
+// TestReconcileLive_InterruptedSpawnTerminated locks the fix for the orphaned
+// orchestrator: a daemon killed mid-Spawn leaves a non-terminated row with no
+// workspace and no runtime handle, and EnsureOrchestrator would return that row
+// forever instead of spawning a session with a real terminal. Boot reconcile
+// must terminate it, without stashing or destroying anything (there is nothing
+// on disk to stash).
+func TestReconcileLive_InterruptedSpawnTerminated(t *testing.T) {
+	st := newFakeStore()
+	rt := &fakeRuntime{}
+	ws := &fakeWorkspace{}
+	lcm := &fakeLCM{store: st}
+	lookPath := func(string) (string, error) { return "/bin/true", nil }
+	m := New(Deps{Runtime: rt, Agents: fakeAgents{}, Workspace: ws, Store: st, Messenger: &fakeMessenger{}, Lifecycle: lcm, LookPath: lookPath})
+
+	rec := domain.SessionRecord{
+		ID: "s9", ProjectID: "p1", Kind: domain.KindOrchestrator, IsTerminated: false,
+	}
+
+	if err := m.reconcileLive(context.Background(), rec); err != nil {
+		t.Fatalf("reconcileLive: %v", err)
+	}
+	if lcm.terminated["s9"] != 1 {
+		t.Fatalf("MarkTerminated(s9) = %d, want 1", lcm.terminated["s9"])
+	}
+	if ws.stashCalls != 0 || rt.destroyed != 0 {
+		t.Fatalf("nothing to tear down: stash=%d destroy=%d", ws.stashCalls, rt.destroyed)
+	}
+}


Pulkit7070 (approved):
LGTM. Verified locally: build/vet clean, full session_manager suite passes, and I reproduced the bug before/after (guard removed = interrupted spawn left alive; guard present = terminated). Stress-tested the guard over 9000 mixed rows under -race with zero over-termination of workspace/runtime sessions. Reconcile is boot-only (daemon.go), so the no-in-flight-spawn safety claim holds. Only housekeeping: CI has not run on the branch yet.

Pulkit7070:
Short version of how this solves #4079:

The bug is an interrupted spawn. If the daemon dies between `Spawn`'s seed-row write and `createSessionWorkspace`, the rollback never runs, leaving a live orchestrator row with no workspace and no runtime handle. `EnsureOrchestrator` returns the newest active orchestrator instead of spawning, so the project is pinned forever to that terminal-less row ("Preparing the orchestrator terminal...") and it survives restarts.

The fix adds one guard at the top of `reconcileLive`: when a live row has both `WorkspacePath == ""` and no runtime handle, terminate it (`MarkTerminated`). That row has nothing to adopt, preserve, or restore, so termination is safe and needs no teardown. Boot reconcile runs before the API serves, so it cannot race an in-flight spawn. Every other reconcile path (adopt, save-and-teardown) is untouched.

Verified locally: reproduced before/after (guard removed = row left alive; guard present = terminated), and stress-tested the guard over 9000 mixed rows under -race with zero over-termination of workspace/runtime sessions.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@Rishet11 Marking as stale: no activity for 26 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Rishet11 Marking as stale: no activity for 26 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Rishet11 Closing as stale: no activity from you for 30 days. Issue #4079 stays open as the tracker. The branch stays safe to reopen if you pick this up again.

PR #4106, opened 2026-08-18
feat(renderer): in-product user survey with answer tracking

Pulkit7070:
A short, opt-in survey to learn who AO's users are and how they work, the kind of thing behavioral telemetry cannot answer. Five questions: what describes you, what you used AO for, how you would feel if it disappeared, what slows you down in AO, and one open wish.

It is a custom modal, not PostHog's native surveys. The renderer disables `/surveys` and `/flags` polling to avoid per-request billing, so answers ride the same batched, sanitized capture pipeline as every other event and cost nothing extra.

- `lib/survey`: a pure, injectable controller (storage, clock, capture all passed in) plus the five-question form, unit tested.
- `SurveyPrompt`: a centered, stepped modal that walks one question at a time and reads AO's own theme tokens so it tracks light/dark.
- `telemetry`: the survey events are registered in `sanitizeRendererProperties` so answers survive the strict per-event allowlist. `survey_completed` carries a whole response as `answer_<id>` props, one row per user, so metrics need no joins.

The survey has no trigger of its own here; the sidebar invite that opens it lands in the stacked follow-up. Mounted but dormant until then.

Events: `ao.renderer.survey_answered` (per question), `ao.renderer.survey_completed` (whole response), `ao.renderer.survey_invite_dismissed`.


illegalcall:
Can you please add a short screenrecording here on interactive with the survey?

i-trytoohard:
@Pulkit7070 Marking as stale: this has been open 22 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4116, opened 2026-08-18
feat(cloud): add Postgres auth foundation

AgentWrapper:
## Summary

Adds the durable multi-tenant identity and tenancy foundation for hosted AO.

- runnable `ao-cloud` and migration binaries
- PostgreSQL users, refresh sessions, organizations, memberships, and placement-schema groundwork
- Google OIDC verification, short-lived AO JWT access tokens, opaque rotating refresh tokens with an absolute lifetime
- server-side email allowlist and per-source authentication throttling
- forced RLS on all six current tenant tables
- narrowly scoped SECURITY DEFINER functions for pre-auth user/session operations
- separate least-privilege migration/runtime roles; runtime startup rejects SUPERUSER/BYPASSRLS and validates RLS flags
- health, readiness, exchange/refresh/logout, and current-account routes
- shared OpenAPI contract, typed Cloud client, Electron installed-app PKCE, and OS-protected token custody

The workspace/session-runtime tables are placement groundwork only. This PR exposes no project/session handlers and stores no AO product data in them.

## Explicitly deferred

- AWS deployment (stacked PR #4166)
- cloud project/session UI
- compute or sandbox providers
- product-state PostgreSQL adapter
- SCM credential brokering and runtime lifecycle
- terminal WAN relay

The product-state implementation must reuse the existing AO services/API/UI with SQLite for local projects and tenant-scoped PostgreSQL for cloud projects. Sandboxes are disposable compute, never a durable database.

## Verification

- `cd backend && go test ./internal/cloud/... ./cmd/ao-cloud ./cmd/ao-cloud-migrate`
- `cd backend && go vet ./internal/cloud/... ./cmd/ao-cloud ./cmd/ao-cloud-migrate`
- CI runs the account lifecycle and tenant isolation against PostgreSQL using a non-superuser, non-BYPASSRLS migrator with CREATEROLE only
- refresh replay, absolute expiry, malformed token parity, allowlist rejection, and rate-limit coverage
- generated contract/client drift checks
- gitleaks


backend/internal/cloud/postgres/account_store.go
```diff
@@ -0,0 +1,247 @@
...
+		principal.UserID,
+	); err != nil {
+		return domain.Principal{}, err
+	}
+	var hasMembership bool
+	if err := tx.QueryRow(
+		ctx,
+		`SELECT EXISTS (
+			SELECT 1 FROM ao_org_memberships
+			WHERE user_id = $1 AND status = 'active'
```
AgentWrapper: This EXISTS check counts only memberships with status active, but the personal-org slug below is deterministic (personal-<uid> with dashes stripped). If the only membership of a user is ever disabled — and docs/cloud-control-plane.md explicitly anticipates disabling memberships — the next sign-in takes the creation branch, inserts a second personal org with the same slug, and hits the slug UNIQUE constraint. That 23505 is not mapped by normalizeError, so every later sign-in for that user fails with a 500, permanently. Suggest treating any existing membership row (or any owned personal org) as sufficient here, and mapping 23505 in normalizeError as a backstop.

backend/internal/cloud/httpapi/auth_handlers.go
```diff
@@ -0,0 +1,219 @@
...
+	newToken, newHash, err := auth.NewRefreshToken()
+	if err != nil {
+		s.internalError(w, r, "issue replacement refresh token", err)
+		return
+	}
+	principal, err := s.store.RotateRefreshSession(
+		r.Context(),
+		auth.HashToken(input.RefreshToken),
+		newHash,
+		time.Now().UTC().Add(s.refreshTokenTTL),
```
AgentWrapper: Rotation re-issues the refresh session with a full refreshTokenTTL from now, so a session that refreshes at least once per TTL never expires — there is no absolute lifetime cap. A stolen refresh token that keeps rotating stays valid indefinitely; replay detection only helps if the legitimate client races it. Consider persisting an absolute session expiry at creation and carrying it through RotateRefreshSession, or enforcing a maximum session age.

backend/internal/cloud/httpapi/auth_handlers.go
```diff
@@ -0,0 +1,219 @@
...
+	var input refreshRequest
+	if !decodeJSON(w, r, &input) {
+		return
+	}
+	if err := requireValue(input.RefreshToken, "refreshToken"); err != nil {
+		writeError(w, r, http.StatusBadRequest, "bad_request", "REFRESH_TOKEN_REQUIRED", err.Error())
+		return
+	}
+	if !strings.HasPrefix(input.RefreshToken, "ao_refresh_") {
+		writeError(w, r, http.StatusBadRequest, "bad_request", "REFRESH_TOKEN_REQUIRED", "refreshToken is invalid")
```
AgentWrapper: Minor: a present-but-malformed token gets code REFRESH_TOKEN_REQUIRED (while the message says invalid) and status 400, whereas the identical prefix check in refresh returns 401 with INVALID_REFRESH_TOKEN. Aligning the code and status would let clients branch consistently.

AgentWrapper (commented):
## Review: Postgres auth foundation

**Verdict: approve** — clean, well-tested foundation. The findings below are non-blocking hardening and edge-case items.

Checked: migration SQL and RLS policies, migration/runtime role separation, Google OIDC verification (audience, azp, email_verified), JWT issue/verify configuration, refresh-token rotation and replay handling, HTTP handlers and error envelope, OpenAPI / generated TS schema / Go handler consistency, and test coverage.

### Strengths
- RLS design is careful: FORCE ROW LEVEL SECURITY, SECURITY DEFINER helpers with a pinned search_path, REVOKE ... FROM PUBLIC, and rejection of privileged or table-owning runtime roles at both migration time and store startup.
- Token handling is correct: HS256 pinning plus WithValidMethods, issuer/audience/expiry enforcement, minimum 32-byte HMAC key, opaque refresh tokens persisted only as SHA-256 digests, and atomic delete-and-insert rotation so concurrent replay has at most one winner.
- Contract, generated schema, and typed client changes line up with the Go handlers.

### Findings (inline comments)
1. Latent permanent 500 on sign-in when the only membership of a user is disabled: the deterministic personal slug collides and the resulting 23505 is unmapped (account_store.go).
2. Refresh rotation grants a fresh full TTL every time, so there is no absolute session lifetime cap; a stolen refresh token can be kept alive indefinitely (auth_handlers.go).
3. Minor: logout returns code REFRESH_TOKEN_REQUIRED with status 400 for a present-but-malformed token, where refresh returns 401 INVALID_REFRESH_TOKEN for the same shape.

### Non-blocking notes
- After RotateRefreshSession commits, a failure in Issue or ListMemberships returns a 500 and the client never receives the new refresh token, so the next refresh 401s and forces a re-login. Consider assembling the response data before committing the rotation.
- Auth endpoints have no rate limiting; acceptable at this stage given the deferred deployment PR, but worth tracking.
- ao_users and ao_auth_sessions intentionally have no RLS (the runtime needs cross-user token-hash lookups); this matches the docs — just noting that forced RLS covers only organizations and memberships.

AgentWrapper (commented):
## Re-review at 4d4207b53: Postgres auth foundation

**Verdict: approve — ready to merge.** All three findings from my previous review (4978739128) are addressed correctly, each with test coverage, and I found no new issues in the fix range.

### Findings resolved
1. **Disabled-membership login 500** — `UpsertGoogleUser` now treats any membership row (regardless of status) as proof the personal org exists, so a disabled personal membership no longer triggers a duplicate-slug insert. As a backstop, `normalizeError` maps 23505 to a new `ErrConflict`, which the exchange handler surfaces as 409 `ACCOUNT_CONFLICT` instead of a 500. Covered by the disabled-membership integration scenario, a unit test for the 23505 mapping, and a handler-level 409 test.
2. **Unbounded sliding session lifetime** — `RotateRefreshSession` no longer takes a TTL; the replacement row carries the original `created_at` and `expires_at` through rotation, so `AO_CLOUD_REFRESH_TOKEN_TTL` is now a hard cap on the whole session. The expired-chain guard (`expires_at > now()`) still applies at rotation. Contract description and docs were updated to state the non-extending behavior, and the integration test asserts the rotated row preserves both timestamps.
3. **Logout error-code mismatch** — logout now returns 401 `INVALID_REFRESH_TOKEN` for a malformed token, matching refresh, with a table test asserting both endpoints behave identically.

### Notes
- Reviewed at the pinned head 4d4207b53; the two later local commits (AWS staging deployment, secret reload) are outside this run and not covered here.
- The remaining non-blocking notes from the previous review (post-rotation response-assembly failure stranding the new token, and no rate limiting on auth endpoints) still stand as tracked follow-ups; neither blocks this foundation.

AgentWrapper (commented):
## Re-review at 8ea0d60: Postgres auth foundation

**Verdict: approve — ready to merge.** Since my last review of this PR (4980428327 at 4d4207b), the branch gained two commits that consolidate the entire security-hardening chain — previously developed and reviewed piece by piece on the staging branch — into the auth foundation itself.

**`fix(cloud): harden tenant auth foundation` (656955b)** brings in, verbatim, work I have already reviewed in detail across the staging rounds; I verified the core files (migrations 00005–00007, `account_store.go`, `migrate.go`, `store.go`) are byte-identical to the approved staging state:
- forced RLS on the identity tables with pre-auth flows in narrowly-scoped SECURITY DEFINER functions owned by the dedicated non-bypass `ao_cloud_auth` role, admitted through explicit named policies;
- migrations 00002–00004 land the workspace/session-runtime schema ahead of any handlers — coherent rather than dead weight, because `validateRuntimeRole` fails closed unless all six tenant tables carry forced RLS and the grants cover them;
- auth rate limiting, orgID validation, and the least-privilege Postgres CI job (dedicated CREATEROLE-only migrator, dedicated database) that exercises migrations, RLS isolation, definer functions, and replay rejection on every push.

**`fix(cloud): use explicit auth rate-limit key` (8ea0d60)** switches to `httprate.LimitBy` with the source-IP key as a required positional argument — same behavior, with the key choice now impossible to omit silently. It composes with the opt-in header trust (`AO_CLOUD_TRUST_SOURCE_IP_HEADER`, default off) so direct deployments never honor a spoofable header.

With this, the auth foundation PR is self-contained and carries its full defense-in-depth story: forced RLS everywhere, least-privilege roles proven in CI, fail-closed startup validation, and abuse-resistant auth endpoints. Ready to merge ahead of #4166.

AgentWrapper (commented):
## Re-review at 9fea1cc: Postgres auth foundation

**Verdict: approve — ready to merge.** The new commit (`fix(cloud): scope runtime placement to one workspace`) tightens the session-runtime RLS from organization-wide to workspace-scoped, closing the last coarse policy in the placement schema.

What I verified:
- **Migration 00008** introduces `ao_current_workspace_id()` and replaces `ao_cloud_session_runtimes_all` with a policy requiring the explicitly selected workspace context, the org context, active membership, and — via an EXISTS against `ao_cloud_workspaces` — that the current user owns the workspace. A future capability bound to one workspace therefore structurally cannot observe or mutate another workspace’s runtime rows even if the application layer misbinds its context; org membership alone addresses nothing. The EXISTS subquery composes correctly with the owner-only workspace select policy from 00005 (no recursion, consistent semantics), the function is revoked from PUBLIC and granted to the runtime role, and the Down path restores the prior policy.
- **The integration test proves both directions against real Postgres**: two workspaces under one owner, runtimes created under each workspace context, then a first-workspace-scoped transaction sees exactly its own runtime and a cross-workspace UPDATE affects zero rows. This runs in the least-privilege CI job on every push.
- **Docs are precise about scope**: the workspace/session-runtime tables remain placement-schema groundwork with no handlers and no product data — this hardening lands ahead of the storage-adapter work that will consume it, which is the right order.

Migration-count and content-marker tests updated. The auth foundation remains self-contained and defense-in-depth throughout. Ready to merge ahead of #4166.

AgentWrapper:
Final foundation verification at `9fea1cc5e`:

- Added migration 00008 to bind session-runtime placement RLS to both `ao.workspace_id` and the owning user, preventing another workspace in the same org from being addressed.
- Extended the real PostgreSQL integration test with same-org/cross-workspace isolation assertions.
- CI is green, including `go test -race ./...`, least-privilege migrations, tenant/auth lifecycle, lint, API drift, and gitleaks.
- The stacked staging deployment applied all migrations successfully and reports Goose schema version 8.

The placement tables remain inert groundwork: this PR exposes no project/session product API and keeps durable AO product state out of sandboxes.

AgentWrapper (commented):
## Re-review at 26c87ed: Postgres auth foundation (rebased onto main)

**Verdict: approve — ready to merge.** This head is a rebase of the previously approved branch onto current main. I verified it mechanically: `git range-diff` reports every one of the seven cloud commits content-equal (`=`) to its predecessor, and a tree diff between the old head (9fea1cc) and this head across all cloud paths — backend/internal/cloud, both ao-cloud binaries, contracts, and the cloud client — is empty. No conflict resolutions altered the reviewed content; the only change is the mainline underneath.

My approval from review 5001677678 carries forward unchanged: forced RLS across all tenant tables including workspace-scoped session-runtime placement, the non-bypass definer role with named admission policies, fail-closed startup validation, capped rotating refresh sessions, allowlist-gated signup, source-IP-keyed rate limiting behind an explicit trust flag, and the least-privilege Postgres CI exercising all eight migrations. Ready to merge.

PR #4135, opened 2026-08-19
feat(browser): add persistent profiles with cookie and history import

aprv10:
## What

Add opt-in persistent browser profiles to AO.

Users can now:

- Create, rename, clear, and delete named browser profiles from Settings.
- Switch profiles from the browser toolbar.
- Keep cookies and browsing history across AO restarts.
- Import cookies and history from supported Chromium-based browsers, Firefox, and Zen.
- Select one or more detected source profiles and import them into new AO profiles.
- Use imported history as address-bar suggestions.

Temporary profiles remain the default, preserving the existing ephemeral and per-worker-isolated behavior.

## Why

AO browser workers previously used temporary sessions only. This meant authentication and browsing context were lost when the worker or app restarted.

Persistent profiles let users explicitly retain useful browser state without exposing profile selection to agents or changing the default behavior for existing sessions.

## How

- Stores AO-owned profile metadata and imported history under `~/.ao`.
- Uses a dedicated persistent Electron partition for each named AO profile.
- Keeps profile selection user-controlled through Settings and the browser toolbar.
- Reads source browser databases through read-only snapshots without exposing source paths to the renderer.
- Validates source paths, realpath containment, symlinks, database sizes, record counts, and import limits before committing data.
- Creates new destination profiles instead of silently overwriting existing profiles.
- Rolls back newly created profiles and imported state when an import fails.
- Handles Chromium cookie encryption and Firefox database differences, reporting skipped or unsupported data as warnings.
- Preserves browser permission denial, renderer ownership checks, tab isolation, lifecycle cleanup, and the restriction preventing agents from choosing `--profile`.
- Packages the SQLite runtime required by the importer in production builds.

Intentional omissions:

- Bookmarks
- Passwords
- Extensions
- Full browser-profile directory copying
- Signed-in browser account synchronization
- Domain-filtered imports

## Testing

- `npm run typecheck`
- `npm run typecheck:e2e`
- Focused browser, profile, settings, preload, and IPC tests:
  - 8 test files passed
  - 233 tests passed
- Import, rollback, source-validation, lifecycle, packaging, and Chromium cookie compatibility tests were added during implementation.
- Production packaging and packaged SQLite runtime verification passed before the final upstream merge.

## Checklist

- [x] Continuing an existing PR branch
- [x] One focused browser-profile and import feature
- [x] Relevant CI checks pass for the final pushed commit


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari (approved):
lgtm

illegalcall:
## Browser profile import: confirmed failure cases

I traced these paths against the source browser schemas and Electron 33.4.11. These are the concrete issues I believe need addressing:

1. **[P1] Partitioned Chromium cookies are imported as unpartitioned cookies** ([AO import code](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L691-L718), [Chromium schema](https://github.com/chromium/chromium/blob/c3b27dee2024961c1bde655b5549b8753a5e3808/net/extras/sqlite/sqlite_persistent_cookie_store.cc#L662-L693), [Electron 33 cookie API](https://github.com/electron/electron/blob/v33.4.11/docs/api/cookies.md#L88-L107))

   `readChromiumCookies` does not select `top_frame_site_key`, and the Electron 33 `cookies.set` API used here has no field that can restore that partition key.

   **When this occurs:** a Chrome profile has a CHIPS cookie for `idp.example` partitioned under `shop.example`. AO recreates it as an ordinary `idp.example` cookie. It can then be sent outside the original top-level-site partition, or collide with a real unpartitioned cookie because AO deduplicates only by name/domain/path.

   **Suggested fix:** select the partition metadata and skip partitioned rows with an explicit counted warning until their isolation can be preserved.

2. **[P1] Firefox container and partition cookies lose their origin isolation** ([AO import code](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L663-L680), [Firefox schema](https://github.com/mozilla-firefox/firefox/blob/017a9913bd6c272a859aea16d3cab5a4d46bf1b0/netwerk/cookie/CookiePersistentStorage.cpp#L1513-L1530))

   The Firefox query omits `moz_cookies.originAttributes`, even though Firefox includes those attributes in cookie identity. They carry container `userContextId` and partition information.

   **When this occurs:** the Work and Personal Firefox containers are both logged into GitHub with different `user_session` cookies. AO sees the same name/domain/path for both, chooses one during deduplication, and imports it into a single unisolated cookie jar. The resulting AO profile can open the wrong account. The same problem applies to cookies partitioned by Firefox Total Cookie Protection.

   **Suggested fix:** select `originAttributes` and skip rows with non-default attributes, with an explicit counted warning, unless AO can preserve those attributes.

3. **[P2] Rollback can report only the original error while leaving partial imported state** ([rollback code](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L464-L479))

   Cleanup uses `Promise.allSettled` but ignores rejected results, and profile-registry deletion is followed by `.catch(() => undefined)`.

   **When this occurs:** an import creates one or more destination profiles, then fails because the disk is full. The same condition can make the rollback registry write fail, leaving a visible partially imported profile while the UI reports only the original import error. Conversely, if `clearStorageData()` fails during app shutdown but registry deletion succeeds, persistent cookie data can be orphaned with no way for the user to retry cleanup.

   **Suggested fix:** collect and surface cleanup failures. Do not claim a complete rollback unless every cleanup step succeeds; retain enough metadata to let the user retry deletion when storage cleanup fails.

4. **[P2] Cookie and history limits silently discard records without updating skipped counts** ([queries](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L663-L695), [history query](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L776-L804), [final cookie slice](https://github.com/Untrivial-ai/agent-orchestrator/blob/d39740ea4b84cc394bf0349d47bde6d613022058/frontend/src/main/browser-profile-import.ts#L868-L875))

   The SQL queries use `LIMIT 20000` for cookies and `LIMIT 5000` for history, and merged cookies are sliced again to 20,000. None of those truncations contributes to a skipped/truncated count; the cookie queries also lack deterministic ordering.

   **When this occurs:** a source Chrome profile contains 25,000 cookies. AO imports an arbitrary 20,000 and reports no skipped records. Merging profiles can discard additional cookies at the final slice. A history database with more than 5,000 eligible visits is likewise silently incomplete.

   **Suggested fix:** fetch `limit + 1` (or count first), apply deterministic ordering, and either reject the oversized import or report the exact truncation in the result.

Items 1 and 2 are merge blockers because they change browser cookie-isolation semantics. Items 3 and 4 are correctness/recovery issues in behavior this PR explicitly introduces.

illegalcall:
Can you attach some screen recs showcasing it?

aprv10:
**Import Wizard flow**

https://github.com/user-attachments/assets/2f617601-692a-46a9-ba6c-ebfd424ca846


**Changing profiles in browser**

https://github.com/user-attachments/assets/9ee18b09-58e8-48f5-a89f-5dbc818b5e09

frontend/src/main/browser-view-host.ts
```diff
@@ -608,13 +672,17 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
...
+			const profileId = boundProfile ? boundProfile.id : null;
 			session = {
 				sessionId,
 				viewId,
+				profileId,
 				// A non-persist: Electron partition is memory-only. Every tab in
 				// this worker shares it, while a fresh worker runtime receives a
 				// different partition even if a session ID is ever reused.
-				profilePartition: `ao-browser-${randomUUID()}`,
+				profilePartition: profileId ? browserProfilePartition(profileId) : `ao-browser-${randomUUID()}`,
```
illegalcall, line 802: **[P1] Shared named profiles leak network-failure signals across workers.** Every worker bound to the same named profile uses this same persistent Electron `Session`, but `registerBrowserSignalWatcher` installs callbacks that close over one AO worker. Electron only keeps the last listener attached to a session-level `webRequest`. **Situation:** workers A and B both use the `Work` profile; B replaces A’s listener, then A’s failed authenticated XHR URL is recorded in B’s signal buffer and exposed to B while A misses it. Destroying either worker also calls `onCompleted(null)`/`onErrorOccurred(null)`, disabling capture for the other. Register one watcher per Electron session/partition, dispatch by `details.webContentsId` to the owning AO session, and reference-count cleanup. Please add a two-worker/same-profile test.

frontend/src/main.ts
```diff
@@ -415,6 +439,14 @@ async function createWindowInternal(): Promise<void> {
 		await agentBrowserRuntime.dispose();
 		return;
 	}
+	const browserProfileStore = await createBrowserProfileStore({ stateDir: browserProfileStateDir() });
+	const browserHistoryStore = new BrowserHistoryStore({ stateDir: browserProfileStateDir() });
+	const browserProfileImporter = new BrowserProfileImportService({
```
illegalcall: **[P1] Active imports are neither cancelled nor awaited during app shutdown.** This service is local to window creation and is absent from `disposeBrowserViewHost`/`before-quit`. **Situation:** a user starts a large cookie import and presses Cmd+Q while snapshots are being read or destination cookies are being written. The process can exit before the importer’s `finally` and rollback finish, leaving plaintext staging snapshots under `browser-import-staging` or a partially imported profile. Track this service at application scope, expose an abort/drain lifecycle, await it before the second `app.quit()`, and remove stale staging directories on startup. Add a shutdown-during-read/commit test.

frontend/src/main/browser-view-host.ts
```diff
@@ -554,20 +607,30 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
...
 				applySessionBounds(session, entry);
 			},
-			() => pushTabsState(options, session),
+			() => {
+				if (isCurrentEntry()) pushTabsState(options, session);
+			},
+			(url, title, incrementVisit) => {
+				const profileId = session.profileId;
+				if (!isCurrentEntry() || !profileId || !options.browserHistoryStore) return;
+				void options.browserHistoryStore.record(profileId, url, title, incrementVisit).catch(() => undefined);
```
illegalcall, line 742: **[P1] Persistent history writes can be lost on immediate shutdown.** The write is deliberately fire-and-forget, but host disposal never drains `BrowserHistoryStore`. **Situation:** the user visits a page and immediately quits; `before-quit` can finish host cleanup and terminate Electron before this queued record reaches disk, so the persistent profile loses its newest visit. Give the history store an awaited `drain()`/`close()` and include it in shutdown cleanup.

frontend/src/main/browser-profile-import.ts
```diff
@@ -0,0 +1,1132 @@
...
+		const columns = tableColumns(database, "moz_cookies");
+		const hasOriginAttributes = columns.has("originAttributes");
+		const isolationWhere = hasOriginAttributes ? "WHERE COALESCE(originAttributes, '') = ''" : "";
+		const total = countRows(database, "moz_cookies");
+		const eligible = countRows(database, "moz_cookies", isolationWhere);
+		const isolatedSkipped = total - eligible;
+		const truncated = Math.max(0, eligible - BROWSER_IMPORT_MAX_COOKIES);
+		const orderBy = cookieOrder(columns, ["lastAccessed", "creationTime"]);
+		const rows = database.prepare(`
+			SELECT host, name, value, path, expiry, isSecure, isHttpOnly, sameSite
```
illegalcall: **[P2] Valid older Firefox/Zen cookie schemas can fail the entire import.** The code introspects `moz_cookies` columns for `originAttributes`, but this query always selects `sameSite`. **Situation:** an older, portable, or preserved profile whose schema predates that column is discovered successfully, then this statement throws `no such column: sameSite` instead of importing supported fields. Build the projection from the detected columns and default unsupported fields, with a warning/test fixture for the older schema.

frontend/src/main/browser-profile-import.ts
```diff
@@ -0,0 +1,1132 @@
...
+	} else {
+		if (!isRecord(destination.names)) throw new Error("Destination profile names are invalid.");
+		names = request.profileIds.map((id) => {
+			if (!Object.hasOwn(destination.names, id)) throw new Error("A destination profile name is missing.");
+			const name = normalizeBrowserProfileName(destination.names[id]);
+			if (!name || name !== destination.names[id]) throw new Error("Destination profile name is invalid.");
+			return name;
+		});
+	}
+	const normalizedNames = names.map((name) => name.toLowerCase());
```
illegalcall, line 618: **[P2] The 32-profile capacity is not validated before committing.** This checks name collisions but never checks `existingProfiles.length + destinationCount`. **Situation:** with 31 AO profiles, selecting two source profiles passes validation; the first destination is created and populated, then the second create hits the limit and forces a late rollback. This contradicts the PR’s promise to validate limits before commit. Reject the request before reading/committing when all destinations will not fit.

frontend/src/renderer/components/settings/BrowserImportDialog.tsx
```diff
@@ -0,0 +1,532 @@
...
+								mode: "separate",
+								names: Object.fromEntries(
+									selectedProfiles.map((profile) => [profile.id, destinationNames[profile.id]?.trim() ?? ""]),
+								),
+							},
+			});
+			setResult(imported);
+			setView("result");
+			onImported();
+		} catch (reason) {
```
illegalcall, line 148: **[P2] A profile intentionally retained after rollback failure is invisible in the current settings session.** `onImported()` refreshes the list only on success; this failure path returns to the form, while `BrowserProfilesSection` otherwise refreshes only on mount. **Situation:** storage cleanup fails, so the importer correctly keeps the profile record for manual deletion and reports that fact, but the user cannot see or delete it until settings is closed and reopened. Refresh the profile list after a failed import as well (or emit a profile-store change event).

frontend/src/main/browser-profile-import.ts
```diff
@@ -0,0 +1,1132 @@
...
+
+const DESCRIPTORS: BrowserDescriptor[] = [
+	{
+		id: "chrome",
+		name: "Google Chrome",
+		family: "chromium",
+		chromiumKeychainNames: ["Chrome", "Google Chrome"],
+		roots: (c) => platformPaths(c, {
+			win32: [joinEnv(c.env.LOCALAPPDATA, "Google", "Chrome", "User Data")],
+			darwin: [path.join(c.homeDir, "Library", "Application Support", "Google", "Chrome")],
```
illegalcall, line 131: **[P1 standards conflict] This reads from an OS default app-data location that the repository’s `AGENTS.md` currently says must never be read from or written to.** Browser import necessarily needs read-only access here, so the implementation and documented boundary disagree. Please add an explicit, narrowly scoped read-only source-browser import exception to the repository rule (while preserving AO-owned state under `~/.ao`), or redesign the boundary before merge.

illegalcall (changes requested):
Requesting changes after a fresh whole-PR review of the current head. The earlier Chromium partition-cookie, Firefox originAttributes, rollback-reporting, and truncation-count issues are fixed.

This pass found three P1 runtime problems (shared-profile webRequest leakage, imports not joined on shutdown, and undrained history writes) plus P2 validation/recovery gaps. I attached the concrete failure scenario and suggested correction to each affected line.

Additional non-inline standards/scope concerns:
- The PR has no linked issue, so the acceptance criteria and design history are not traceable.
- browser-profile-import.ts is 1,132 lines and combines discovery, path validation, schema adapters, cryptography, parsing, limits, transactions, and rollback; source-specific adapters would reduce divergent branching.
- The duplicated browserProfiles E2E fake, one-call profileLabel helper, and five-setter applySourceDefaults interface should be cleaned up.
- Multi-profile merge behavior and cookie/history conflict precedence are not specified; document and test them if merge mode remains in scope.

Verification on this exact head: both TypeScript checks passed; 225 test files passed (2,967 tests, 1 skipped); Electron packaging passed; and the packaged better_sqlite3.node artifact was present. Those automated checks do not exercise the lifecycle and same-partition concurrency cases below.

frontend/src/renderer/components/BrowserPanel.tsx
```diff
@@ -461,7 +495,26 @@ export function BrowserPanelView({
 	const submit = (event: FormEvent<HTMLFormElement>) => {
 		event.preventDefault();
 		const nextURL = urlInput.trim();
-		if (nextURL) void navigate(nextURL);
+		if (nextURL) {
+			setUrlEditing(false);
```
illegalcall: [P2] Preserve address-bar editing after navigation

This leaves the address input focused while setting `urlEditing` to false. A user-facing failure occurs when someone enters a URL, presses Enter, and immediately starts typing another URL without clicking elsewhere: because the input never blurred, `onFocus` does not run again, the controlled value continues rendering `navState.url`, and the new keystrokes are discarded. I reproduced this by extending the existing BrowserPanel test to type `second.example` immediately after Enter; the input stayed focused but its value remained empty. Please either blur the input after submitting/selecting a suggestion or keep the focused input in editing mode, and add a consecutive-navigation regression test.

illegalcall (approved):

frontend/src/renderer/components/BrowserPanel.tsx
aprv10: Fixed in 6a4841f72. Address-bar submissions and history-suggestion selections now use one commit path that blurs the input before navigation, preventing the focused-but-nonediting state. Added focused coverage for consecutive navigations and suggestion selection; BrowserPanel tests and frontend typecheck pass.

PR #4136, opened 2026-08-19
fix(browser-panel): tabs rail placement, overlay blanking, and URL bar scroll

Vaibhaav-Tiwari:
## Summary
Phase 1 of the AO in-app browser UX overhaul (fix-the-breakages pass, scoped against current `main`):

- Tabs rail now renders on the right in both docked and popped-out/maximized mode (previously it flipped to the left when maximized).
- Fixed the black screen behind any raised overlay (tabs-rail flyout, dialogs, etc.). **Update:** the first attempt only toggled a Tailwind utility class; `.browser-panel__viewport` also carries its own unconditional plain-CSS background in `styles.css`, and unlayered author CSS always beats a Tailwind utility regardless of class presence, so the first fix had no real effect. Fixed properly now by gating the actual CSS rule behind a `data-placeholder` attribute.
- The "Pick element" / agent-activity status no longer sits in the toolbar as permanent on-screen text for the whole duration of picking or agent activity — it's now a small status dot with the detail available on hover (native tooltip) and still announced to screen readers via an sr-only live region.
- Added a stress-test regression guard for the previously-reported "tabs stop closing after a long session" bug, driving the real `closeTab`/`destroyTabView` main-process code through hundreds of open/close/reactivate cycles. It does not currently reproduce on `main` — several stabilization fixes already landed upstream since it was first reported — so no production fix was needed, just the regression guard plus a note on `closeTab`.
- Visual QA pass across docked/popped-out/DevTools/annotation/multi-tab/resize/error/theme states found and fixed one more bug: the URL bar's scroll position wasn't reset after navigating, hiding the `https://` scheme.

## Known follow-up (not fixed here, needs a design decision)
At the docked inspector's minimum width (280px), the Browser toolbar's icon buttons leave very little room for the URL text (down to 1-3 visible characters). Flagged during the visual QA pass rather than guessing at which control to collapse — worth a design call in a follow-up.

## Test plan
- [x] `npx vitest run` on `BrowserPanel.test.tsx` and `browser-view-host.test.ts` — all passing
- [x] Manual pass in the real Electron app (native-compositor behavior — the black-screen fix and DevTools/annotation states — can't be verified by automated tests in this environment)


Prasad-D-Ware:
image
image
image

Prasad-D-Ware (approved):

LaibaFirdouse:
## Review for Ubuntu Linux:

Reviewed the merged changes locally on Ubuntu. All targeted tests passed (`window-composition`: 2/2, `BrowserPanel`: 44/44, native-composition cascade: 4/4). Manual Electron testing also verified browser navigation, multiple tabs, tabs flyout/tooltip behavior, live page visibility while overlays are open, and rapid overlay open/close without reproducing blank/black rendering or compositor issues. No reproducible issues found.

PR #4139, opened 2026-08-19
fix(windows): reduce daemon probe timeouts for faster app startup

### LINKED ISSUE #4138: Windows app takes too long to open because of serial daemon probe timeouts
## Bug

After #4013 fixed the flashing command prompt windows, the Windows app got really slow to open. Like you open the app and it just sits there on the loading screen for 6+ seconds doing nothing before the daemon even starts.

## Why this is happening

In `frontend/src/main.ts`, the `startDaemonInner` function runs three HTTP probes one after another to check if a daemon is already running. Each probe waits 2 seconds to timeout. On a cold start when no daemon is running, all three timeout, so thats 6 seconds of just waiting for nothing.

The thing is, on localhost a refused connection comes back in like 20ms. So waiting 2 seconds each time is just wasting the users time.

This wasnt as noticeable before #4013 because Windows dev mode used to run `go run` which was slow anyway. Now that it uses a prebuilt exe, the probe wait time is the main bottleneck.

## Steps to reproduce

1. On Windows, make sure no daemon is running
2. Open the app or run `cd frontend && npm run dev`
3. Watch it sit on the loading screen for a while before anything happens

## Expected

App should open fast. Shouldnt be waiting 6 seconds for localhost connections to tell us nobody is listening.

## Notes

- Only affects Windows, dont touch macOS or Linux
- Got worse after #4013 because Windows now goes through the full probe chain

JAYATIAHUJA:
## What happened

After #4013 fixed the flashing cmd windows on Windows, app startup got noticeably slower. On a cold start (no daemon running), the app just hangs for like 6 seconds before the daemon even starts spawning.

Turns out `startDaemonInner` runs three HTTP probes back to back and each one waits 2 full seconds to timeout. On localhost a refused connection comes back in like 20ms, so 2 seconds per probe is way too much.

## What this PR does

Two small changes, both Windows only:

1. Probe timeout goes from 2000ms to 500ms on Windows. Thats still plenty for localhost, and shaves off ~4.5 seconds on cold start.
2. Skip the orphan probe when theres no run file. If nothing suggests a daemon was ever running, no point burning another timeout checking for one.

macOS and Linux are not touched at all.

Closes #4138

## Test plan

- [x] daemon-launch, daemon-attach, daemon-takeover tests all pass (49 tests)
- [ ] Cold start on Windows should be noticeably faster
- [ ] Warm start (daemon already running) should work same as before
- [ ] macOS/Linux startup unchanged

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -547,7 +547,7 @@ const RUN_FILE_POLL_MS = 300;
 // Accept run-files stamped slightly before our spawn timestamp: the daemon's
 // clock reading and ours race within normal scheduling jitter.
 const RUN_FILE_FRESHNESS_SKEW_MS = 2_000;
-const DAEMON_PROBE_TIMEOUT_MS = 2_000;
+const DAEMON_PROBE_TIMEOUT_MS = process.platform === "win32" ? 500 : 2_000;
 
 function runFilePath(): string | null {
 	if (process.env.AO_RUN_FILE) return process.env.AO_RUN_FILE;
@@ -1087,7 +1087,6 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
 	// killed here), and a foreign non-AO process holding the port with a dead
 	// run-file PID is not replaced (out of scope). When no holder is detectable,
 	// skip straight to spawn.
-	const orphanProbe = await readDaemonProbe(resolvedDaemonPort(), "healthz");
 	const runFilePath_ = runFilePath();
 	let runFilePid: number | null = null;
 	if (runFilePath_) {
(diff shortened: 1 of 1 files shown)


i-trytoohard:
@JAYATIAHUJA Marking as stale: this has been open 21 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

i-trytoohard:
@JAYATIAHUJA Marking as stale: no activity for 24 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@JAYATIAHUJA Marking as stale: no activity for 24 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@JAYATIAHUJA Closing as stale: no activity from you for 35 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

Issue #3387, opened 2026-07-31
Kimi agent shows 'Working' indefinitely after provider.auth_error (401) — no terminal activity detector for non-Codex harnesses


i-trytoohard:
## Bug

A Kimi agent session shows **"Working"** status indefinitely after encountering a `provider.auth_error` (401 — invalid or expired API key). The agent has effectively stalled — it caught the error, printed it to the terminal, and returned to its interactive prompt — but the dashboard topbar and status badge still claim the session is actively working.

**Source:** Discord [#bug-triage thread](https://discord.com/channels/1489612265027014736/1532554294358708319) | **Reported by:** @prateek | **Analyzed against:** `e53ca2806`
**Confidence:** High

![Screenshot of the Kimi agent session stuck on "Working" after a provider.auth_error](https://raw.githubusercontent.com/i-trytoohard/codex-startup-factory/main/.issue-assets/kimi-auth-error-working/kimi-auth-error-working.png)

## Reproduction

1. Start an AO session using the Kimi harness (Kimi K2.7)
2. Configure an invalid or expired Moonshot API key
3. Deliver a prompt to the agent
4. Kimi attempts to call the Moonshot API, gets `provider.auth_error 401`, prints the error, and returns to its interactive prompt
5. **Observe:** Dashboard shows "Working" (green dot) with "Coding thinking" in the status bar — the session appears active indefinitely

## Root Cause

This is a **two-layer failure**: the `Stop` hook doesn't fire on provider errors, AND the safety-net observer can't compensate because Kimi lacks the required interface.

### Layer 1 — `deriveStatus()` treats stale `Active` as truth

`backend/internal/service/session/status.go:36-37`:
```go
case domain.ActivityActive:
    return domain.StatusWorking
```

There is **no staleness check**. Even if `LastActivityAt` is hours old, as long as `Activity.State == Active`, the status is unconditionally `Working`. The only way to leave `Active` is a hook callback that transitions the state.

### Layer 2 — Kimi's `Stop` hook likely doesn't fire on provider errors

AO installs four Kimi hooks (`backend/internal/adapters/agent/kimi/hooks.go:194-200`):
- `SessionStart` → active
- `UserPromptSubmit` → active
- `PermissionRequest` → waiting_input
- `Stop` → idle

[...]

PR #4143, opened 2026-08-19
fix(claude-code): reconcile stuck-active sessions via terminal activity detection

### LINKED ISSUE #4142: bug(agent-claude-code): session stuck working/active after a turn dies without its Stop hook (login expired) — observer skips claude-code, no TerminalActivityDetector
## Bug

A claude-code worker session (live: `agent-orchestrator-139`) shows **status=working / activity=active** indefinitely while the Claude CLI is provably idle at its composer. The last turn aborted on an expired OAuth login (`Login expired · Please run /login`, `Worked for 0s`), so the turn's Stop hook never fired — and nothing else ever demotes the stale `active`.

**Source:** live observation of session `agent-orchestrator-139` (tmux pane capture) | **Reported by:** maintainer via orchestrator task | **Analyzed against:** `0e472e2b2`
**Confidence:** High

## Reproduction

1. Spawn a claude-code TUI worker; let a turn start (`user-prompt-submit` → activity=active).
2. Kill the turn without a Stop hook — auth expiry mid-stream reproduces it (turn ends with `Login expired · Please run /login`, `✻ Worked for 0s`); a CLI crash or network drop behaves the same.
3. Observe `ao session get`: status stays `working`, activity stays `active` forever. Any further Enter on the logged-out CLI re-fires `user-prompt-submit` and re-asserts `active` (which is why `updated` advances on each retry while nothing runs).

## Root Cause

Claude Code's durable activity is **hook-driven only** (`backend/internal/adapters/agent/claudecode/activity.go` → `lifecycle.ApplyActivitySignal`). The only transitions out of `active` are the `stop` / `notification(idle_prompt)` / `session-end` hooks. A turn that dies mid-stream emits none of them.

AO already has the exact safety net for this: the activity observer (`backend/internal/observe/activity/observer.go:103-131`) polls every 30s and, for sessions stuck `active` past `DefaultStaleAfter` (2 min), screen-proves the real state via the adapter's `TerminalActivityDetector`. But the claude-code adapter **never implemented `DetectTerminalActivity`**, so `reconcile()` returns at the type assertion (`observer.go:111-114`) and claude-code sessions are exempt from reconciliation entirely. (The adapter has `InspectTerminalSurface` — the classification logic exists — it just isn't wired to the detector port.) The prior test `TestPollLeavesOtherHarnessesUntouched` even pinned claude-code as the example of a skipped harness.

Verified against the live pane: `InspectTerminalSurface` on the stuck screen (login-expired footer, staged draft in the composer, no active chrome) correctly yields `Work=Idle` — the observer would have converged the session within one tick if the detector had been wired.

Hypotheses ruled out: hook/classifier precedence is not involved (there is simply no classifier signal), and nothing refreshes `updated` unconditionally (it stayed frozen at `15:15:56Z` for 6+ minutes of observation; it only advances when the user re-submits on the logged-out CLI).

## Fix

Implement `DetectTerminalActivity` on the claude-code plugin, delegating to `InspectTerminalSurface` — authoritative idle only when the surface positively shows an idle composer (same pattern codex shipped for #3115). Claude stays non-continuous: hooks remain authoritative while they flow; the observer only reconciles a stale `active`.

## Impact

- Users see a phantom "Working" session and wait for output that never comes; the logged-out state is hidden.
- Orchestrators treat the worker as busy and won't dispatch or escalate.
- Class bug: ANY hook turn-death (auth expiry, crash, network) strands claude-code in `working` — until now, permanently.

## Related

- [#3387](https://github.com/Untrivial-ai/agent-orchestrator/issues/3387) — same class for Kimi; documents that only codex implemented `TerminalActivityDetector` (this issue is the claude-code instance)
- [#3115](https://github.com/Untrivial-ai/agent-orchestrator/issues/3115) — codex instance (fixed by codex's `terminal_activity.go`)
- [#4048](https://github.com/Untrivial-ai/agent-orchestrator/issues/4048) — cline instance (fixed via continuous reconciliation)
- [#2047](https://github.com/Untrivial-ai/agent-orchestrator/issues/2047) — broader claude activity misreporting

AgentWrapper:
Fixes #4142

## Summary

A Claude Code turn that dies without emitting its Stop hook (expired OAuth login, CLI crash, network loss) left the session's hook-driven activity stuck at `active`, so the app showed **status=working** forever while the CLI sat idle at its composer (live incident: `agent-orchestrator-139`, stranded on `Login expired · Please run /login`).

The activity observer already reconciles exactly this case — a session stale in `active` past 2 minutes is screen-proven against the adapter's `TerminalActivityDetector` — but the claude-code adapter never implemented that capability, so the observer skipped its sessions at the type assertion (`observer.go:111-114`).

This PR implements `DetectTerminalActivity` on the claude-code plugin, same pattern as codex (#3115): delegate to the existing `InspectTerminalSurface` and report an **authoritative idle only** when the current surface positively shows an idle composer with no active spinner/interrupt chrome and no pending decision frame. Anything else fails closed. Claude deliberately stays non-continuous: hooks remain authoritative while they flow; the observer only steps in for a stale `active`.

Scope note: this intentionally fixes claude-code only. The same class bug for other hook-only harnesses is tracked in #3387 (Kimi).

## Changes

- `backend/internal/adapters/agent/claudecode/terminal_activity.go` (new): `DetectTerminalActivity` delegating to `InspectTerminalSurface`, idle-only, fail-closed.
- `backend/internal/adapters/agent/claudecode/terminal_activity_test.go` (new): fixtures reproduce the real captured stuck screen (login-expired footer + staged draft, and empty-composer variant) → authoritative idle; active spinner, active footer hint, permission dialog, unrecognizable/empty output → not authoritative.
- `backend/internal/observe/activity/observer_test.go`: `TestPollLeavesOtherHarnessesUntouched` previously pinned claude-code as exempt from reconciliation (encoding the bug) — now uses droid as the detector-less example. Added `TestPollReconcilesStaleClaudeCodeAfterAbortedTurn` (stuck screen → one `terminal-idle` signal with launch/updatedAt fence) and `TestPollKeepsGenuineLongClaudeTurnActive` (real active chrome → no signal).

## Behavior notes

Implementing the `TerminalActivityDetector` port also opts claude-code into the two existing detector consumers beyond the observer, both matching codex's shipped behavior:
- `WaitForMessageDeliveryReady` now screen-proves the composer before injecting (previously: durable idle + first-hook-or-5s fallback) — strictly stronger evidence.
- The interface-transition drain's legacy fallback (only reachable when styled output is unavailable) can now prove idle for claude-code; it additionally requires `ComposerIsEmpty`, which claude-code already implements.

## Tests

- `go test ./internal/adapters/agent/claudecode/ ./internal/observe/activity/ ./internal/session_manager/ ./internal/lifecycle/` — pass
- Full `go test ./...` — pass (`internal/cli` requires all `AO_*` env unset; failures with leaked env are pre-existing and unrelated)
- Classifier hand-verified against the live stuck pane of `agent-orchestrator-139` (plain `tmux capture-pane`, which is what the observer's `GetOutput` provides): `Work=Idle`, so the observer converges the session within one tick after staleness.

## Risks / follow-ups

- A genuinely active turn whose rendering momentarily lacks both the spinner row and the footer interrupt hint could be sampled as idle; the next tool hook immediately re-asserts `active`, and this is the same trade-off codex has shipped since #3115.
- Follow-up filed separately: surface the `Login expired` screen as a distinct needs-input reason so users know to run `/login` instead of seeing a silent idle.

### DIFF
--- a/backend/internal/adapters/agent/claudecode/terminal_activity.go
+++ b/backend/internal/adapters/agent/claudecode/terminal_activity.go
@@ -1,0 +1,25 @@
+package claudecode
+
+import (
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// DetectTerminalActivity reports idle only when Claude's rendered surface
+// positively shows an idle composer: no active spinner/interrupt chrome and no
+// pending provider decision frame. Anything else is not authoritative.
+//
+// Claude Code's durable activity is hook-driven, and a turn that dies without
+// emitting its Stop hook (auth expiry, CLI crash, network loss) leaves the
+// session stuck active forever. This capability is what opts the adapter into
+// the activity observer's stale-active reconciliation, which screen-proves
+// idle after the hook state has been quiet past its staleness limit. Claude
+// intentionally does not implement ContinuousTerminalActivityDetector: hooks
+// stay authoritative while they flow, and the composer draft state must never
+// promote or demote sticky states on its own.
+func (p *Plugin) DetectTerminalActivity(output string) (domain.ActivityState, bool) {
+	if p.InspectTerminalSurface(output).Work == ports.TerminalSurfaceWorkIdle {
+		return domain.ActivityIdle, true
+	}
+	return "", false
+}
--- a/backend/internal/adapters/agent/claudecode/terminal_activity_test.go
+++ b/backend/internal/adapters/agent/claudecode/terminal_activity_test.go
@@ -1,0 +1,94 @@
+package claudecode
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+// claudeAbortedTurnScreen reproduces the live pane of a Claude Code session
+// whose turn died on an expired OAuth login: the turn's Stop hook never fired,
+// the composer holds an unsent user draft, and the CLI sits idle at the
+// prompt. Captured plain (tmux capture-pane without -e), which is exactly what
+// the activity observer's GetOutput provides.
+func claudeAbortedTurnScreen(draft string) string {
+	rule := strings.Repeat("─", 48)
+	return "⏺ Stopped watching Artifact: \"scm-observer.md\" (connection lost)\n" +
+		"\n" +
+		"⏺ Login expired · Please run /login\n" +
+		"\n" +
+		"✻ Worked for 0s\n" +
+		"\n" +
+		rule + "\n" +
+		draft +
+		rule + "\n" +
+		"\n" +
+		"  ⏵⏵ bypass permissions on (shift+tab to cycle) · PR #4090\n" +
+		"  ⧉  scm-observer"
+}
+
+func TestDetectTerminalActivityAuthoritativeIdleAfterAbortedTurn(t *testing.T) {
+	plugin := &Plugin{}
+	tests := []struct {
+		name   string
+		output string
+	}{
+		{
+			name: "login expired with staged draft",
+			output: claudeAbortedTurnScreen("❯ so btw, the status isn't permanently stuck at \"Checking merge readiness\". It's stuck at that for\n" +
+				"  most of the time but keeps showing the merge status once in a while.\n"),
+		},
+		{
+			name:   "login expired with empty composer",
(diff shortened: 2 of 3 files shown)


nikhilachale (approved):
Looks Good ,PR propelry fixes the root cause .
Approving

PR #4146, opened 2026-08-19
fix(browser): non-disruptive preview reveal + reopen closed tabs

Vaibhaav-Tiwari:
## Summary
Phase 2 of the AO in-app browser UX overhaul (calmer, less intrusive UX):

- **Non-disruptive agent browser actions**: a new `ao preview` target used to force-switch the inspector to the Browser tab and pop it open, even if the user was looking at something else entirely (Reviews, chat, a different session). Now it just badges Browser as unseen — same pattern already used for general agent browser activity (fill/click/snapshot) — and lets the user open it when they're ready.
- **Reopen tabs closed by the agent (or you)**: closing a tab was final — the only way back was asking the agent to reopen it. Now the last 5 closed tabs (with title/url/favicon) show as a "Recently closed" section in the tabs rail, in both the popped-out list and the docked hover flyout, click to reopen. The list now survives switching sessions and back (previously renderer-only state, lost on any session switch — unlike live tabs, which are kept alive in the main process).
- **Fixed the automation runtime permanently wedging a session**: `ensureNativeActiveTab`'s convergence loop retried the identical failing `tab-select` forever once the runtime's own tab registry forgot the active tab, wedging every native browser operation for the session — not just close, but tab switching too. It now asks the runtime to refresh its own view before retrying once, and accepts the drift (with a `console.warn` breadcrumb) instead of looping forever if still stale.
- **Fixed a reopen leaving contradictory state**: reopening a closed tab pointed at a dead localhost dev server used to report failure while the tab was actually created and left open — `openTab` threw after already telling the renderer the tab existed. It now matches how navigating an *existing* tab to a dead URL already worked (tab exists, shows an error page, no throw).
- **Fixed the close fallback double-closing an already-closed tab**: the automation runtime's own `Target.closeTarget` handling can close a tab internally before the runtime still reports the overall command as failed. The fallback used to retry AO's own close regardless, throwing `TAB_NOT_FOUND` for a tab that had already, genuinely closed. It now checks whether the tab is already gone and treats that as success. The renderer's close handler also now resyncs tab state from the main process on any failure, instead of leaving a stale "ghost" row that re-fails identically on every retry.
- **Fixed reopening at the 16-tab cap silently losing the closed-tab entry**: gated on the same cap the "+" button respects, with restore-on-failure for races.
- Favicons now render in Recently Closed rows (previously captured but unused). Close/select failures surface a transient notice instead of an unhandled rejection. `closeTab`'s identity no longer churns on every tab update.

## Scope note
The original Phase 2 plan also included a "general smoothness" item (optimistic tab select/close instead of waiting on the IPC round-trip, and/or trimming a MutationObserver's overhead). Left out of this PR — the codebase's own comments show that overhead was already a deliberate, documented trade-off, and reworking tab-selection to be optimistic carries real risk of subtle state bugs that's better done as its own focused pass rather than rushed in here.

The upstream question of *why* the automation runtime's tab registry drifts from `session.tabs` in the first place is still open and deliberately out of scope here — this PR makes AO resilient to it happening, not a fix to the runtime itself. Worth its own follow-up issue.

A separate, pre-existing paper cut was flagged during review: `browser:navigate` silently rejects a URL matching the renderer's own origin (a dev-mode `localhost:517x` port) with no notice surfaced to the user. Confirmed out of scope for this PR by the reviewer.

## Test plan
- [x] `npx vitest run` on `browser-view-host.test.ts`, `BrowserPanel.test.tsx`, `useBrowserView.test.tsx`, `i18n/instance.test.ts` — 163/163 passing, with regression tests for every fix above (each verified to fail against the pre-fix code before being confirmed passing)
- [x] Manual pass in the real Electron app across three rounds of review — tabs opening/closing/switching, Recently Closed (favicons, cap gating, reopen, session-switch persistence), non-disruptive preview reveal


Prasad-D-Ware:
Reviewed and ran this locally (checked out the branch, ran the Electron dev app in isolated mode against `~/.ao/dev/data`). The "Recently closed" UX itself looks good — but the automation-runtime recovery in `33ea5257d` doesn't actually work, and I have logs showing why. Details below.

## What I verified

- `npm --prefix frontend run typecheck` — clean
- `npx vitest run` on `browser-view-host.test.ts`, `BrowserPanel.test.tsx`, `SessionView.test.tsx`, `useBrowserView.test.tsx` — **204/204 passing**
- Manual pass in the real Electron window (the box left unchecked in the PR description):
  - "Recently closed" renders correctly in **both** surfaces — the expanded/popped-out rail and the docked hover flyout — with the section header and per-row history icon. Spacing and layout look right.
  - Clicking a closed entry reopens it: it comes back as a live tab at its original URL and disappears from the Recently closed list.
  - Section is correctly absent when nothing has been closed, and absent from the icon-only pinned rail.

## Blocker: the `tab-close` fallback guards the wrong call site

After using the browser panel normally for a while (open/close/switch), **closing tabs stopped working entirely** — silently, with no notice in the UI. The fallback added in `33ea5257d` never fires. Here is the actual stack from the dev log:

```
Error occurred in handler for 'browser:closeTab': Error: {"error":"Tab t6 not found; run `agent-browser tab` to list open tabs","success":false}
    at AgentBrowserRuntime.run
    at async AgentBrowserRuntime.runAction
    at async ensureNativeActiveTab          <-- here
    at async main.js:36323
  code: 'AGENT_BROWSER_COMMAND_FAILED'
```

The throw comes from `await ensureNativeActiveTab(session)` on `browser-view-host.ts:1170`, which sits **outside** the new `try` block — not from the `tab-close` `runAction` inside it. `ensureNativeActiveTab` appears in every stack frame in my log, which can only happen from that unguarded line; built line `36323` maps to it. So the `catch` that falls back to AO's own `closeTab` is unreachable in the failure mode that actually occurs in practice.

The fallback isn't useless — it does cover the case where the forgotten tab is a *non-active* one. It's necessary but insufficient, so the stated goal ("recover from the automation runtime losing track of a tab") isn't met yet.

### The underlying cause is a permanent wedge, and it's wider than closing

`ensureNativeActiveTab` (`:607`) is a convergence loop:

```ts
while (session.nativeActiveTabId !== session.activeTabId) {
  const tabId = session.activeTabId;
  ...
  await runtime.runAction(session.sessionId, "tab-select", { tabId }, ...);
  session.nativeActiveTabId = tabId;   // :624 — only on success
}
```

Once the runtime's registry forgets the **active** tab, `tab-select` throws, `:624` is never reached, and `nativeActiveTabId` can never converge — so every subsequent native browser operation for that session fails *forever*. That matches the symptom exactly: it works for a while, then stops permanently.

It also isn't limited to closing. `browser:selectTab` (`:1150`) calls `ensureNativeActiveTab` with no guard at all and fails identically in my log (`Tab t9 not found`), so **tab switching breaks too**.

To be clear: `ensureNativeActiveTab` is untouched by this PR, so the wedge is pre-existing on `main` — this PR just drew attention to the symptom. Worth deciding whether to fix the root cause here or split it out. If it's split out, the fallback in this PR should still be corrected so its stated goal holds. Guarding `ensureNativeActiveTab` inside `closeTab`'s `try` would unblock closing, but selecting would stay wedged — that's a symptom fix. The durable fix is to make the loop converge: on `AGENT_BROWSER_COMMAND_FAILED` from `tab-select`, re-selecting the same id will never succeed, so recovery has to reset the runtime's view of that target (re-list/re-create) rather than retry, and must not leave `nativeActiveTabId` permanently mismatched.

One open question I couldn't answer from the outside: *why* the runtime forgets a tab it was told about via `listTargets()`, which is derived from `session.tabs` (the authoritative source). That's internal `agent-browser` state — worth pinning down, since it decides whether recovery should re-list or re-create.

Also, these failures surface only as `[Unhandled rejection]` in the console — the rail fires `void onCloseTab(tabId)`, so the user gets no feedback at all. Some `tabNotice`-style surfacing for a failed close/select would turn a silent dead button into something diagnosable.

## Other findings

**1. Reopening at the tab cap silently loses the entry** (`useBrowserView.ts:562`)

```ts
setClosedTabs((current) => current.filter((tab) => tab.id !== tabId));
await openTab(entry.url);   // throws BROWSER_TAB_LIMIT at 16 tabs
```

The entry is dropped before the `await`, with no rollback. `openTab` throws `BROWSER_TAB_LIMIT` at `MAX_BROWSER_TABS` (`browser-view-host.ts:461`). The existing "+" button never hits this because the rail disables it (`canOpenTab`, `BrowserTabsRail.tsx:237`), but the new Recently-closed rows aren't guarded — so at the cap, clicking one deletes it from the list, opens nothing, and rejects into `void onReopenClosedTab(tabId)` with no notice. Either restore the entry on failure or gate the rows on `canOpenTab` like the "+" button.

**2. `favicon` is captured but never rendered.** It's in `ClosedBrowserTab`, populated in `closeTab`, and asserted in tests, and the PR description says entries carry favicons — but `ClosedTabsSection` always renders the generic `History` icon and never reads `tab.favicon`. Either wire it up (with `History` as the fallback, like `TabFavicon` does) or drop the field.

**3. Nit — `closeTab`'s identity now churns.** Deps went from `[hasNativeBrowser]` (stable) to `[hasNativeBrowser, tabsState.tabs]`, so it's a new function on every tab update, cascading through `handleCloseTab`. Reading the tab from a ref would keep it stable.

## Things I checked that are fine

- **No tab-id collision** in the closed list — `nextTabNumber++` is monotonic per session, so ids are never reused and a stale closed entry can't shadow a live tab.
- **The fallback's error narrowing is right** — scoping to `AGENT_BROWSER_COMMAND_FAILED` correctly lets a missing binary / disposed runtime still throw.
- **Not resetting `nativeActiveTabId` on the fallback path is safe** — `ensureNativeActiveTab` reconciles against `session.activeTabId`, so it self-corrects on the next native op.
- **Session scoping on `closedTabs`** is correct — cleared on session switch and termination, gated by `stateBelongsToSession`.

## On the `ao preview` behavior change

The non-disruptive badge is the right call for the focus-stealing problem. Flagging one downstream effect: this repo's `CLAUDE.md` instructs agents to run `ao preview` when demoing frontend changes, and after this PR that no longer reveals anything on its own — the change just badges Browser as unseen. Might be worth a line in `CLAUDE.md` / the `using-ao` preview skill so that workflow still lands the way it's described.

Vaibhaav-Tiwari:
Thanks for the thorough review and the real repro logs — pushed `745c26841` addressing everything.

**Blocker — the wedge, fixed at the root cause.** You were right that the try/catch guarded the wrong call site, and right that the real problem is `ensureNativeActiveTab`'s convergence loop itself: retrying the identical `tab-select` can never succeed once the runtime's registry forgets a tab, so it wedged every native operation (select, close, click, snapshot) for the session forever. Fixed inside the loop itself rather than at each call site: on `AGENT_BROWSER_COMMAND_FAILED`, it now asks the runtime to refresh its own tab list (`agent-browser tab`, per its own error message) and retries once; if still stale, it accepts the drift instead of looping — AO's own tab state never depended on runtime agreement to begin with. This fixes `browser:closeTab` and `browser:selectTab` (which had no guard at all) from the same place. Added two regression tests that fail without the fix and reproduce your exact stack trace (`Tab t1 not found; run \`agent-browser tab\`...` thrown from `ensureNativeActiveTab`) — confirmed by reverting the fix locally and watching them fail at the right line.

I didn't chase down *why* the runtime's registry forgets a tab it was told about via `listTargets()` — that's internal to the external binary and, as you said, worth pinning down separately since it decides whether recovery should re-list or re-create. This fix makes AO resilient to it happening rather than fixing the upstream cause.

**Other findings, all addressed:**
1. Reopening at the tab cap: gated on the same `canOpenTab` the "+" button already respects, and restores the entry if `openTab` still fails (race with concurrent agent activity) instead of losing it.
2. `favicon`: wired up in `ClosedTabsSection`, same image-with-fallback pattern as `TabFavicon`.
3. `closeTab`'s dependency churn: reads from a ref now, stable identity across tab updates.
4. Silent close/select failures: both now surface a transient notice via the existing `tabNotice` mechanism instead of an unhandled rejection.
5. `CLAUDE.md`: added a note that `ao preview` no longer force-reveals the Browser tab after the non-disruptive change, so agents say so in their reply instead of assuming the user saw it.

`npm --prefix frontend run typecheck` clean, `npx vitest run` on the four touched suites — 157/157 passing, including the new tests. Would appreciate another pass whenever you have time.

Prasad-D-Ware:
Re-reviewed `745c26841` and ran it in the Electron dev app again. The wedge fix is right, and I verified the regression tests are real. One new issue though — the rollback added for the tab-cap finding introduced a bug, and I hit it live within minutes of normal use.

## Verified

- `npm --prefix frontend run typecheck` clean; **211/211** on the four suites (not 157 — that undercounts, but in the right direction).
- **The new regression tests are genuine.** I reverted `browser-view-host.ts` to the pre-fix commit and re-ran them: both fail, thrown from `ensureNativeActiveTab` (`:617`) via the `selectTab` handler (`:1155`) — the same stack as the live repro. Restored after.
- **The fix is in the right place.** `session.nativeActiveTabId = tabId` now sits after the try/catch so the loop always converges, covering `selectTab` and `closeTab` from one spot rather than per-call-site. `"tabs"` is the correct action — it maps to `["tab", "list"]` (`agent-browser-runtime.ts:648`), exactly what the runtime's own error message suggests.
- Cap fix is properly defense-in-depth (hook gate + `disabled` row + explanatory `title`), and all 8 locales carry all 3 new keys.
- **Favicons confirmed rendering in the app** — Recently closed rows now show real per-site icons (Instagram/GitHub/Google), with the generic icon only as fallback for blank/unreachable entries.

**Caveat worth stating plainly:** the drift never reproduced in this dev session, so the wedge fix is proven in tests but **not yet in the real app**. Absence today isn't evidence — it took sustained tab churn to surface last time.

## New issue: the reopen rollback leaves contradictory state

Clicking a Recently-closed entry whose server is gone (a stopped dev server — `http://localhost:5175/` in my case) reports failure *while actually leaving the tab open*, and restores the entry.

The cause is ordering in `openTab` (`browser-view-host.ts:665`):

```ts
const entry = createTab(session, activate);        // :679 — tab now exists
await entry.ready;
if (normalizedURL) {
  const navigation = navigateEntry(entry, normalizedURL);
  pushTabsState(options, session, { kind: reason, tabId: entry.tabId });   // :683 — renderer already notified
  const state = await navigation;
  if (state.error) throw browserError("NAVIGATION_FAILED", state.error);   // :685 — but this rejects
}
```

The tab is created and pushed to the renderer *before* navigation is awaited, so a navigation failure rejects an IPC call for a tab that already exists. The new `catch` in `reopenClosedTab` treats that rejection as "nothing happened": it shows "Couldn't reopen that tab" and restores the entry.

One click therefore yields three contradictory outcomes — a tab that exists, a message saying it doesn't, and a list still calling it closed. Confirmed visually in the running app: `localhost:5175` is simultaneously present as an open tab (with the fallback globe icon, i.e. the failed load) **and** sitting at the top of the Recently closed section.

**Repro:**
1. Open a tab on a local dev server; close it (lands in Recently closed)
2. Stop the dev server
3. Click the entry — notice says it failed, row stays
4. Repeat — the row is never clearable this way

**Evidence from my run:** three identical `browser:openTab` rejections for the same URL (`CDP error (Target.createTarget): ERR_CONNECTION_REFUSED (-102) loading 'http://localhost:5175/'`, code `AGENT_BROWSER_COMMAND_FAILED`), across 4 total load attempts at `:5175`. Each click after the first is only possible because the previous failure restored the row — the loop is self-sustaining in ordinary use, not something I had to engineer.

This is the same class of bug the PR already fixed on the close side — *"only remember it once the main process confirms it's actually gone."* The rollback needs the mirror rule: only restore the entry when the tab genuinely wasn't created.

**But the better fix is probably upstream of that.** Chrome opens the tab and renders an error page for an unreachable URL; here the whole reopen is treated as failure. Recently-closed entries are overwhelmingly localhost dev servers — exactly what `ao preview` produces, and they die constantly — so this is the common case, not an edge case. If a failed navigation still yields a usable tab (error page), the rollback question disappears and behavior matches every other browser.

Worth noting my original finding pointed only at the tab cap; this shows `openTab` rejects for a far more common reason, which the cap gate alone doesn't cover.

## Smaller ask: leave a breadcrumb when drift is accepted

The `catch (resyncError)` branch in `ensureNativeActiveTab` is currently comment-only — no logging. When the runtime stays desynced, AO sets `nativeActiveTabId = tabId` and proceeds as if in sync, and every agent action routes through this loop first (`:1300`) — so a later `click`/`fill`/`snapshot` can land on a different page than intended, silently. Accepting the drift is the right call over wedging, but a `console.warn` (session id, tab id) would make it diagnosable when someone reports "the agent clicked the wrong tab" months from now.

## Housekeeping

The PR description is now stale: it still describes only the original two changes, still says 130/130, and the manual-verification checkbox is unchecked despite several rounds of manual testing since. Worth refreshing before merge.

Separately, the upstream question — *why* the runtime's registry forgets a tab it was told about via `listTargets()` — is still unanswered and deferred by both of us. I'd suggest a standalone issue so it doesn't get lost; this PR makes AO resilient to it, which is the right scope for it.

Vaibhaav-Tiwari:
Thanks for catching that live and confirming the wedge fix — pushed \`f80265efd\`.

**Reopen rollback / contradictory state, fixed at the root as suggested.** Went with the better fix rather than the mirror-rule patch: \`openTab\` now matches \`navigateEntry\`'s own existing behavior for navigating an *existing* tab to a dead URL — resolve normally with the tab's nav state carrying the error, don't throw. It never made sense that the exact same failed-load outcome was "fine, tab has an error banner" for every other navigation but "the operation failed" specifically for tab creation. Confirmed by reverting locally: the new regression test fails at the exact line you found (\`browser-view-host.ts:685\`, thrown from \`openTab\`, caught in \`agentBrowserTargets.createTarget\`) and passes with the fix.

**Breadcrumb added** — \`console.warn\` with session id and tab id in the \`catch (resyncError)\` branch, with a test asserting it fires (and fails without the fix).

**PR description refreshed** with the current change list and test counts.

Re: the housekeeping note on the upstream question (why the runtime's registry drifts) — agreed it's worth its own issue so it doesn't get lost; I've called that out explicitly in the description as deliberately out of scope for this PR.

\`npm --prefix frontend run typecheck\` clean, 159/159 across the four suites (openTab/warn regressions included, and I reverted-and-confirmed both fail without their respective fixes before restoring). Let me know if another pass turns up anything else.

Prasad-D-Ware:
Re-tested `f80265efd` in the Electron dev app. **The reopen fix works and both follow-up findings are properly resolved** — but live testing surfaced a new blocker in the `tab-close` fallback from `33ea5257d`, and it's the same failure I originally reported ("tabs stop closing"), reached by a completely different route.

## Verified on `f80265efd`

- `npm --prefix frontend run typecheck` clean; **213/213** on the four suites.
- **Both new tests are genuine.** I reverted `browser-view-host.ts` to `745c26841` and re-ran them: `does not reject when the new tab's initial navigation fails` fails with `ERR_CONNECTION_REFUSED`, and the drift-warning test fails on the missing `console.warn`. Restored after.
- **The reopen fix is right, and fixed at the root.** I confirmed `navigateEntry` (`:985-993`) does exactly what the commit says — catches the load error, sets `.error`, pushes `browser:navState`, resolves — so `openTab` now matches it rather than special-casing tab creation. That eliminates the whole contradictory-state class, not just the symptom.
- **The URL allowlist is unaffected**, which was my main concern about loosening `openTab`: `NAVIGATION_FAILED` still throws for a disallowed URL at `:681` and `:1265`. Only *load* failures became non-throwing. Both `openTab` callers (`createTarget`, and the no-runtime handler path) are accounted for.

## Blocker: the `tab-close` fallback double-closes an already-closed tab

Closing tabs broke again during ordinary use — **8 × `TAB_NOT_FOUND: Browser tab tN does not exist`**. The stack resolves to built line `36346`, which is `return closeTab(session, input.tabId)` inside the fallback's `catch`:

```
Error occurred in handler for 'browser:closeTab': Error: Browser tab t4 does not exist
    at browserError (main.js:37072)
    at closeTab (main.js:35994)          <-- internal closeTab, not the handler guard
    at main.js:36346                     <-- the fallback line
  code: 'TAB_NOT_FOUND'
```

Sequence:

1. The handler's guard (`:1165`) confirms `session.tabs.has(input.tabId)` — the tab **provably exists** at request time
2. `runAction("tab-close")` closes it through the runtime's own `closeTarget` callback (`:727-729`, which calls AO's internal `closeTab`)
3. The command reports failure anyway → `AGENT_BROWSER_COMMAND_FAILED`
4. The fallback fires and calls `closeTab(session, input.tabId)` a second time → the tab is gone → `TAB_NOT_FOUND`

Because the guard in step 1 proved the tab existed, something removed it mid-call, and `closeTarget` is the only path that does. So the fallback's premise — "the command failed, therefore nothing happened" — is wrong in exactly the scenario it was written for.

This is the same class of bug as the one `f80265efd` just fixed, inverted: that one treated a live tab as a failed *reopen*; this treats an already-closed tab as a failed *close*. The remedy is symmetrical — if the tab is already absent when the fallback runs, that **is** success, so return `listTabs(session)` instead of letting internal `closeTab` throw.

Note this is a distinct bug from the wedge. The wedge threw from `ensureNativeActiveTab` before `tab-close` ever ran; this throws from the fallback *after* a `tab-close` that partially succeeded. Same user-visible symptom — "tabs won't close" — different cause, which is why the convergence fix doesn't cover it.

## It compounds: a failed close leaves a ghost row

**7 × `CANNOT_CLOSE_LAST_TAB`** followed the above. When a close rejects, the renderer surfaces the notice but never refreshes `tabsState`, so a tab that genuinely closed in the main process stays in the rail. Clicking it again re-fails (the log shows `t4` twice, `t5` three times). Eventually the main process is down to a single tab while the UI still lists several, so every further close hits the last-tab guard.

The stale list isn't new — before the `catch` was added, the rejection was unhandled and state wasn't refreshed either. What changed is that it's now visible instead of silent. Worth fixing alongside the fallback: on a failed close, re-syncing from `browser:getTabs` (or having the main process return current state rather than throwing when the tab is already gone) would stop one failed close from leaving a permanently dead-looking row.

## Still unverified: the wedge fix itself

Across both dev sessions, **zero** runtime `Tab tN not found; run \`agent-browser tab\`` errors and **zero** `accepting drift` warnings. The original drift never reproduced, so `ensureNativeActiveTab`'s convergence fix remains proven by unit tests and unobserved in the real app. Not a blocker — the tests are genuine, I checked — but worth stating plainly rather than treating a quiet session as confirmation.

For completeness, what *did* check out live: favicons render per-site with the generic icon only as fallback, and no `BROWSER_TAB_LIMIT` errors appeared.

## Unrelated, pre-existing

8 × `Unsupported browser URL` from `browser:navigate`. Only `http:`/`https:` are allowed (`ALLOWED_PROTOCOLS`, `:346`), and any URL matching the renderer's own origin is refused by design (`:363`) — which in dev is a `localhost:517x` port. It fails with no explanation surfaced to the user: the same silent-rejection gap this PR just closed for close/select, in a different handler. Out of scope here, but it's the same paper cut.

Prasad-D-Ware:
@Vaibhaav-Tiwari also if i change sessions we lose the recently closed tabs. i went from worker 1 to worker 2 the tabs inside recent were lost

Vaibhaav-Tiwari:
Both fixed, pushed \`a7fa81976\`.

**Double-close, fixed as you diagnosed.** Confirmed the exact sequence: \`Target.closeTarget\` (invoked from inside \`runAction\`) calls AO's own internal \`closeTab\`, and the runtime can still report the overall command as failed afterward. The fallback now checks \`session.tabs.has(tabId)\` before retrying — if it's already gone, that's success (\`return listTabs(session)\`), not a second close attempt. Also fixed the compounding renderer-side symptom you flagged: a failed close now resyncs via \`getTabs\` instead of leaving a stale row that re-fails identically forever. New test constructs the exact partial-success-then-report-failure sequence (calls \`provider.closeTarget\` before throwing, mirroring the real bridge behavior) and fails without the fix.

**Recently Closed lost on session switch — real bug, fixed.** Root cause: unlike live tabs (kept alive in the main process, just re-fetched on switch-back), Recently Closed was pure renderer state with no backing anywhere else, so the reset-on-session-switch effect had nothing to restore from. Added a session-keyed cache (same pattern already used for consumed preview triggers, which solves an analogous "survive a session switch" problem) — restored on switch-back, cleared only on genuine session termination, not a mere switch-away. Two new tests cover both paths (survives switch-and-back, cleared on termination) and both fail without the fix.

\`npm --prefix frontend run typecheck\` clean, 163/163 across the four suites — reverted each fix locally and confirmed its new test(s) fail at the right spot before restoring, same as previous rounds. PR description refreshed again. Let me know what the next pass turns up.

Prasad-D-Ware (approved):

PR #4149, opened 2026-08-19
fix: isolate ChatUI history by session

nikhilachale:
## Summary
- prevent ChatUI from rendering a conversation snapshot that belongs to a previously selected session
- remount `ChatWorkspace` when the selected session changes so retained timeline/composer state cannot carry across orchestrators
- hide project conversation pages before the current orchestrator's `ao-context-reset:<session>` boundary, so a replacement orchestrator does not page in the previous orchestrator's history
- filter stale queued/running turns at the same reset boundary so old turn state cannot make the replacement orchestrator appear busy
- persist the fresh-context reset boundary atomically with project conversation rebinds before provider startup
- strengthen regression coverage with a stateful workspace mock and a store-level project conversation rebind/reset test

## Root cause
Project-scoped conversations are intentionally rebound when an orchestrator is replaced. The frontend-only fix reset retained React state, but the paged conversation API could still return rows from before the fresh agent context boundary. That made old orchestrator history visible again after switching/replacing orchestrators.

This PR now treats the context-reset system activity for the current project orchestrator as the first visible sequence for paged ChatUI reads. The boundary activity remains stored as internal metadata, but it is not returned in paged ChatUI reads; older turns, messages, and activities are not returned or pageable from the replacement orchestrator view. The fresh-context boundary is written in the same transaction as the project conversation rebind before provider startup, so provider-start failures do not expose old history under the replacement session.

## Tests
- `cd frontend && PATH=/Users/nikhilachale/.nvm/versions/node/v22.23.1/bin:$PATH npm test -- SessionChatSurface.test.tsx`
- `cd backend && GOCACHE=/private/tmp/go-build go test ./internal/storage/sqlite/store`
- `cd backend && GOCACHE=/private/tmp/go-build go test ./internal/service/chat`
- `cd backend && GOCACHE=/private/tmp/go-build go test ./internal/httpd/controllers -run TestConversation`

### DIFF
--- a/backend/internal/domain/conversation.go
+++ b/backend/internal/domain/conversation.go
@@ -35,6 +35,13 @@ const (
 	ConversationScopeProject ConversationScope = "project"
 )
 
+// ConversationContextResetProviderItemID returns the durable identity of the
+// hidden activity that separates a fresh project orchestrator context from the
+// project conversation history that came before it.
+func ConversationContextResetProviderItemID(session SessionID) string {
+	return "ao-context-reset:" + string(session)
+}
+
 // TurnState is the lifecycle of one request and the agent work that follows it.
 type TurnState string
 
--- a/backend/internal/domain/conversation_test.go
+++ b/backend/internal/domain/conversation_test.go
@@ -1,0 +1,10 @@
+package domain
+
+import "testing"
+
+func TestConversationContextResetProviderItemID(t *testing.T) {
+	const session = SessionID("orchestrator-2")
+	if got := ConversationContextResetProviderItemID(session); got != "ao-context-reset:orchestrator-2" {
+		t.Fatalf("ConversationContextResetProviderItemID() = %q", got)
+	}
+}
--- a/backend/internal/service/chat/controller.go
+++ b/backend/internal/service/chat/controller.go
@@ -38,6 +38,7 @@ const (
 // the SQLite store.
 type Store interface {
 	CreateConversation(ctx context.Context, id string, scope domain.ConversationScope, project domain.ProjectID, session domain.SessionID, now time.Time) (domain.ConversationRecord, error)
+	CreateProjectConversationWithContextReset(ctx context.Context, id string, project domain.ProjectID, session domain.SessionID, reset domain.ConversationActivity, now time.Time) (domain.ConversationRecord, error)
 	ConversationForSession(ctx context.Context, session domain.SessionID) (domain.ConversationRecord, error)
 	ClaimChatControllerGeneration(ctx context.Context, session domain.SessionID, generation string, now time.Time) error
 	ConversationBranch(ctx context.Context, conversationID, branchID string) (domain.ConversationBranch, error)
--- a/backend/internal/service/chat/controller_test.go
+++ b/backend/internal/service/chat/controller_test.go
@@ -1086,7 +1086,8 @@ func TestFreshProjectControllerRecordsNativeContextBoundary(t *testing.T) {
 		t.Fatalf("activities = %#v, want old history plus context boundary", snapshot.Activities)
 	}
 	boundary := snapshot.Activities[1]
-	if boundary.Kind != domain.ActivityKindSystem || boundary.ProviderItemID != "ao-context-reset:p1-2" {
+	if boundary.Kind != domain.ActivityKindSystem ||
+		boundary.ProviderItemID != domain.ConversationContextResetProviderItemID(replacement) {
 		t.Fatalf("boundary = %#v", boundary)
 	}
 	var detail map[string]string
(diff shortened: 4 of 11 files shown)


nikhilachale (commented):
Approved: ready to merge. I reviewed the ChatUI session-isolation change and the added regression coverage; I did not find any blocking correctness issues.

illegalcall:
There are two correctness gaps in the reset-boundary implementation that should be addressed before merge:

1. **Pre-reset queued/running turns are still returned.**

   `SelectConversationTurnsPage` (`backend/internal/storage/sqlite/queries/conversations.sql`) includes every turn whose state is `queued` or `running`, regardless of `oldest_sequence`. `LoadConversationSnapshotPage` then copies those rows directly into the snapshot. The new regression test creates `old-turn` before the reset boundary, leaves it queued, and only asserts the filtered messages and activities—not `page.Turns`. As written, the snapshot still contains that old turn.

   This is user-visible because the frontend derives its active and queued state from `snapshot.turns` (`activeTurn` and `queuedTurnIds`). A replacement orchestrator can therefore appear to be working or queued because of a turn owned by the previous orchestrator, even though the old messages are hidden.

   Please apply the reset cutoff to turns as well and extend the store test to assert that `old-turn` is absent while `fresh-turn` remains.

2. **Conversation rebinding is visible before the reset boundary is written.**

   `Service.Start` calls `CreateConversation`, which immediately binds the project conversation to the replacement session. It then starts/resumes the provider, claims the controller generation, and only afterward writes the `ao-context-reset:<session>` activity. During that interval, the boundary lookup returns `0`, so paged reads expose the previous orchestrator's rows under the new session ID. The frontend session-mismatch guard cannot catch this because `SnapshotPage` labels the response with the newly requested session ID. If provider startup or boundary persistence fails, this can remain visible rather than being only a brief race.

   Please make the rebind and reset-boundary persistence atomic, or keep paged history hidden while a newly rebound session has no completed boundary. Add coverage for a read between rebind and boundary creation, including the provider-start failure path.

illegalcall:
### Local native Electron E2E evidence

Tested commit `6aa0088ab101180b95612aa2abaa5ba329aee5e8` on macOS arm64 using the real Electron app and an isolated daemon/data directory.

Flow exercised:

1. Started a Chat orchestrator and sent `OLD CONTEXT VISIBLE`; verified the user and assistant messages were persisted.
2. Replaced that project orchestrator with a fresh Chat session.
3. Verified the replacement conversation API returned `hasMoreBefore: false` and did not return the old turn or either old message.
4. Sent `FRESH CONTEXT ONLY` to the replacement and verified the native ChatUI rendered the fresh exchange while `OLD CONTEXT VISIBLE` was absent from the DOM.

PR #4149 replacement ChatUI showing only fresh context

Harness details: renderer `http://localhost:5174`, daemon `127.0.0.1:3014`, isolated state under `~/.ao/dev-pr4149`. The non-default ports/path were launch-only isolation because another AO dev checkout was using the standard dev profile; no harness changes were committed.

This validates the normal sequenced-history replacement path. It does **not** cover the separately identified zero-sequence stale-running-turn case, which remains an outstanding edge case.

illegalcall:
Superseding my earlier screenshot-only comment with a reproducible boundary check. This run used the PR head (`6aa0088`) in the native Electron app, isolated AO state, and a disposable worktree of the local `todo-app` repository.

### What was exercised

1. Started project orchestrator `todo-pr4149-proof-1`.
2. Sent `OLD SESSION 1 MUST DISAPPEAR` and waited for the completed reply.
3. Replaced the project orchestrator with `clean: true`, creating `todo-pr4149-proof-2`.
4. Sent `FRESH SESSION 2 ONLY` to the replacement and waited for the completed reply.
5. Checked the SQLite boundary, the daemon conversation response, and the Electron renderer.

Both orchestrators use the **same durable conversation ID** (`c3dde959-b892-431d-bc71-130e0983eacc`), so this is testing the rebind/reset behavior rather than accidentally opening an unrelated conversation.

### Machine-checked result

```text
PASS same durable conversation: c3dde959-b892-431d-bc71-130e0983eacc
PASS controller rebound: todo-pr4149-proof-1 -> todo-pr4149-proof-2
PASS raw DB retains 2 old marker messages at sequences 1-2
PASS reset boundary exists at sequence 4: Agent context reset.
PASS replacement API begins at sequence 5 with hasMoreBefore=false
PASS replacement API exposes 0 old marker messages and 0 old turns
PASS replacement API exposes 2 fresh marker messages and 1 completed fresh turn
PASS Electron before view contains the old marker
PASS Electron after view contains the fresh marker and contains 0 old markers
```

The important part is that SQLite still contains the earlier audit/history rows, but the replacement session's public snapshot starts after the reset boundary. The old completed turn (`b3ffb441-3c16-47d7-9751-4331860a3803`) is absent; only the replacement's completed turn (`8b2f58b2-7e35-4426-861f-112d29d78e82`) is returned.

Full assertion ledger: [pr4149-todo-proof.txt](https://gist.githubusercontent.com/illegalcall/4e6530269cdd65c2e8e0f60721971e22/raw/051976c0c4c3a22be83e493ffa12f4d7ccde7f66/pr4149-todo-proof.txt)

### Native Electron evidence

Before replacement — original session shows the old prompt and reply:

Before replacement: old session marker is visible

After replacement — fresh prompt and reply are visible; the old marker is absent:

After replacement: only fresh session marker is visible

Scope note: this proves the normal **sequenced-history** replacement path. It does not prove the separate zero-sequence stale-running-turn edge case, because the original conversation already had messages at sequences 1-2 before replacement.

illegalcall (approved):

PR #4154, opened 2026-08-19
fix(auth): complete WorkOS sign-in locally

somewherelostt:
## Summary
- use a temporary fixed-port 127.0.0.1 callback during WorkOS PKCE sign-in so the browser receives a final success page
- try WorkOS-configured callback ports in order: 5174, 3000, then 3001
- show "You can close this tab and return to Agent Orchestrator" after the code exchange completes
- trigger a focus-only ao-app://complete handoff from that success page so the browser still shows the native app-open prompt
- keep the existing ao-app://callback handling as a fallback for older in-flight auth redirects
- avoid relaunching WorkOS when a valid cloud session already exists

## Root cause
The sign-in flow used the custom ao-app:// callback as the primary redirect. Browsers present that as an external app launch prompt, and some OS/browser combinations pass the deep-link argument differently or leave the provider tab sitting on a page that keeps trying to hand off to Electron. Switching the token exchange to a local HTTP callback fixes the browser UX, but WorkOS requires the redirect URI to match the application settings exactly, so the callback must use one of the configured loopback ports instead of a random port. A separate ao-app://complete deep link restores the native prompt after the browser has already landed on the close-tab page.

## Validation
- npm --prefix frontend test -- cloud-auth
- npm run frontend:typecheck
- git diff --check -- frontend/src/main/cloud-auth.ts frontend/src/main/cloud-auth.test.ts frontend/src/main.ts

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -73,6 +73,7 @@ import {
 import { browserDaemonOwnershipDecision, shouldReplacePortHolder } from "./shared/daemon-takeover";
 import { buildDaemonEnv, resolveShellEnv, type ShellRunner } from "./shared/shell-env";
 import {
+	findCloudDeepLinkArg,
 	handleCloudDeepLink,
 	installCloudIPC,
 	registerCloudProtocol,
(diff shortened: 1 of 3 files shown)


somewherelostt:
its not needed now!

PR #4160, opened 2026-08-20
fix: normalize cross-provider token usage

### LINKED ISSUE #4126: Clarify token usage reporting and avoid apparent cache-token double counting
## Bug

AO presents an inclusive `inputTokens` value alongside cache-read and cache-write token counts, then labels `inputTokens + outputTokens` as **Total tokens**. The arithmetic currently represents all tokens processed, but the terminology makes cache tokens appear duplicated and makes AO difficult to compare with agent-native usage reports.

**Source:** User report in an AO session on 2026-08-18  
**Analyzed against:** `78a1366310c071a5edf5d01f420c162e02e40dec`  
**Confidence:** High — the parser, aggregation query, API DTO, and renderer calculation were traced.  
**Priority:** P2 — reporting is available but materially confusing; underlying token categories are already retained.

The screenshots below are from different sessions and should not be numerically reconciled. They illustrate the terminology problem only.

### AO usage inspector

AO usage inspector

### Claude Code native usage report

Claude Code usage report

## Current behavior

For Claude transcripts, `backend/internal/observe/usage/parser.go` normalizes:

```text
inputTokens = uncached input + cache reads + cache writes
```

Codex reports inclusive input, so AO derives uncached input by subtracting cache reads/writes.

The inspector (`frontend/src/renderer/components/SessionInspector.tsx`) and compact-session query (`backend/internal/storage/sqlite/queries/usage.sql`) then calculate:

```text
totalTokens = inputTokens + outputTokens
```

This is a valid **tokens processed** measurement, but the UI shows `inputTokens`, cache reads, and cache writes as if they were independent rows. Reasoning tokens are also shown without explaining that they are included in output.

## Desired reporting contract

Use mutually exclusive user-facing categories:

- **Fresh input** — uncached input tokens
- **Cache reads**
- **Cache writes**
- **Output** — includes reasoning tokens
- **Reasoning (included in output)** — informational subset, never added again

Derive and label the aggregate explicitly:

```text
tokensProcessed = freshInput + cacheReads + cacheWrites + output
```

Do not display the inclusive internal `inputTokens` value as ordinary **Input tokens**.

Suggested summary presentation:

```text
Tokens processed                         2.2M
Estimated cost                              —

Fresh input                              86.5K
Output                                      8K
Cache reads                              2.1M
Cache writes                                 0
Reasoning (included in output)            3.1K
Cache hit rate                              96%
```

Collapsed session/agent/model rows could use `2.2M processed · 96% cached`. Once reliable pricing is available, estimated cost should be the primary comparison metric because token classes have substantially different prices.

## Proposed implementation

1. Add an explicitly named `processedTokens` derived field to detailed and compact usage responses.
2. Present `uncachedInputTokens` as **Fresh input**.
3. Stop presenting inclusive `inputTokens` in the visible breakdown.
4. Rename **Total tokens** to **Tokens processed**.
5. Label reasoning as included in output.
6. Add a cache-hit percentage with a documented denominator.
7. Keep existing API fields temporarily for compatibility and deprecate ambiguous naming rather than silently changing its meaning.
8. Ensure all pricing and telemetry formulas price/count each mutually exclusive category exactly once.

No historical data rewrite should be required: AO already persists uncached input, cache read, cache write, output, and reasoning separately.

## Acceptance criteria

- [ ] Visible token categories are mutually exclusive except reasoning, which is explicitly marked as a subset of output.
- [ ] `processedTokens` equals fresh input + cache reads + cache writes + output at model, harness, session, and compact-card scopes.
- [ ] Cache tokens are not counted twice in any displayed aggregate, cost calculation, or telemetry field.
- [ ] Claude raw `input_tokens` maps to Fresh input; Claude cache fields remain separate.
- [ ] Codex inclusive input is split correctly into fresh/cache categories.
- [ ] Session cards and inspector labels distinguish processed volume from fresh input and cost.
- [ ] OpenAPI and frontend generated types are regenerated together.
- [ ] Parser, aggregation, controller, and renderer tests cover the reporting contract.

## Related

- [PR #4003](https://github.com/Untrivial-ai/agent-orchestrator/pull/4003) — adds token cost telemetry and should use the same mutually exclusive category contract.
- [PR #2928](https://github.com/Untrivial-ai/agent-orchestrator/pull/2928) — introduced event-driven session token usage.

whoisasx:
## Summary
- normalize Claude/Anthropic and Codex/OpenAI usage into the canonical five metrics: input, cached input, uncached input, cached output, and output
- retain provider-specific reasoning, cache-write, and Anthropic cache-creation counters in namespaced storage/API detail blocks, while keeping them hidden from the shared UI
- add per-metric provenance (`reported`, `derived`, `unsupported`, `unknown`) and preserve known zero versus unavailable values
- migrate existing usage rows provider-aware, including safe historical detail backfill; unavailable historical Anthropic TTL buckets remain null
- make processed tokens `input + output`, remove the ambiguous compatibility fields, and render uncached input/cached input beside total input/total output in the inspector
- validate Codex cumulative/last-token reconciliation and deduplicate repeated Claude physical records by logical message identity

## Tests
- `cd backend && go test ./internal/observe/usage ./internal/storage/sqlite/... ./internal/service/usage ./internal/httpd/...`
- `npm --prefix frontend test -- SessionInspector.test.tsx SessionsBoard.test.tsx` (131 passed)
- regenerated sqlc, OpenAPI, and frontend API types with pinned `openapi-typescript@7.4.4`

## Risks / notes
- this intentionally replaces the old usage API fields rather than retaining aliases
- historical Anthropic 5-minute/1-hour cache-creation buckets stay unknown unless a replay safely enriches them
- the first full `npm run lint` attempt hit the existing timing-sensitive fake-agent lifecycle test; that test passed on isolated retry and golangci-lint passed

Closes #4126


whoisasx (commented):
Ready — no changes requested. I reviewed the processed-token derivation through the normalized usage model, compact SQL aggregation, service-level model/harness/session summaries, HTTP/OpenAPI contracts, and both frontend presentation paths. The mutually exclusive categories are summed once, compatibility fields remain intact, the cache-rate denominator is documented, and the focused backend/frontend coverage plus current CI checks support the change.

whoisasx:
# 1. Single Agent , Single Model
Screenshot 2026-08-21 at 15 16 01


# 2. Single Agent, Multi Model 
Screenshot 2026-08-21 at 15 16 30
Screenshot 2026-08-21 at 15 16 21


# 3. Multi Agent, Multi Model
Screenshot 2026-08-21 at 15 17 04
Screenshot 2026-08-21 at 15 16 56
Screenshot 2026-08-21 at 15 16 44

Annieeeee11 (changes requested):
## Findings

**P1 — missing Codex `total_tokens` treated as zero** (`parser.go` `parseCodexEvent`)

`validCodexTotal` allows `TotalTokens == 0` (absent), but the new monotonicity check (`total.TotalTokens < state.Baseline.TotalTokens`) and `reportedTotal` delta treat `0` as a real counter. After a cumulative with a real total, a later event that omits `total_tokens` is flagged non-monotonic, baseline advances, and the delta is dropped even when input/output rose. The reverse (zeros then a real total) makes `normalizeOpenAIUsage` reject `reportedTotal != input+output` after baseline was already updated — permanent undercount. Mirror the `TotalTokens == 0` exception used elsewhere, and only advance baseline after validation succeeds.

**P2 — `last_token_usage` mismatch still persists cumulative delta**

When `last` is present and disagrees with the cumulative delta, the parser only bumps `AnomalyCount` then still appends the cumulative-derived event. Prefer `last` (or skip) on mismatch instead of storing the divergent cumulative values.

whoisasx:
@Annieeeee11 Addressed both findings in 7d8b6dd91:

- Treat `total_tokens: 0` as absent on either side of cumulative reconciliation, derive an event total when a real total resumes, and avoid advancing the baseline when the selected event fails validation.
- Prefer valid `last_token_usage` when it disagrees with the cumulative delta; invalid last vectors are skipped without advancing the baseline.

Added regressions for real → missing → real totals, rejected-delta baseline recovery, and last/cumulative mismatch selection.

Verified:
- `go test ./internal/observe/usage ./internal/storage/sqlite/... ./internal/service/usage ./internal/httpd/...`
- `golangci-lint v2.12.2`

backend/internal/observe/usage/parser.go
```diff
@@ -551,34 +570,61 @@ func parseCodexEvent(source domain.UsageSourceContext, envelope codexEnvelope, s
...
+		InputTokens: input, CachedInputTokens: cached, CacheWriteInputTokens: cacheWrite,
+		OutputTokens: output, ReasoningOutputTokens: reasoning, TotalTokens: reportedTotal,
+	}
+	selected := delta
+	if payload.Info.Last != nil {
+		last := *payload.Info.Last
+		if !validCodexTotal(last) {
+			result.Cursor.AnomalyCount++
+			result.Cursor.LastErrorCode = domain.UsageErrorNonMonotonicCumulativeUsage
+			return
```
whoisasx, line 592: When payload.Info.Last is present but fails validCodexTotal (e.g. a negative field, or an internal inconsistency confined to last_token_usage itself), this branch returns without updating state.Baseline, even though total (the cumulative total_token_usage) already passed validCodexTotal and the monotonic check above it.

This record's delta is simply dropped, but the next successful token_count event still computes its delta from the same stale baseline -- so it silently absorbs this record's tokens on top of its own new tokens, inflating that session's token totals from that point on.

Every other bailout path in this function advances state.Baseline to total before returning (the non-monotonic-usage branch above, the selected-all-zero no-op branch below, and the end of the happy path). This one is the exception, which looks like a missed update rather than an intentional early return. Suggest adding `state.Baseline = total` before this return.

Not covered by existing tests: TestParsersRejectInvalidTokenVectorsAndAdvanceCursor only exercises an invalid `total`, and TestParseCodexLastMismatchPrefersLastUsage only covers a `last` that fails the delta-match check, not validCodexTotal itself (that mismatch case correctly does advance the baseline).

whoisasx (commented):
## Review: fix: normalize cross-provider token usage

This is a well-executed refactor of the token usage pipeline: the five-metric canonical vocabulary, per-metric provenance (reported/derived/unsupported/unknown), namespaced provider-detail tables, the historical-data migration (0102_canonical_usage.sql), and the frontend inspector changes all check out. In particular the migration correctly reconstructs the new uncached/cached split from the old schema for both providers, and validateUsageEvent plus the conflict/enrichment logic in usage_store.go look sound.

One high-confidence correctness bug found in the Codex last_token_usage reconciliation path in parser.go -- see inline comment. It can cause silent token double-counting for the event that follows a malformed last_token_usage record. Requesting changes to fix that before merge; everything else looks ready.

backend/internal/observe/usage/parser.go
whoisasx: Fixed in d490f4519. The invalid last_token_usage path now advances the validated cumulative baseline before skipping the record, and a regression test verifies that the next event does not absorb the skipped tokens.

illegalcall (commented):
Review against #4126 and the repository standards. The focused backend suite, frontend typecheck, 132 targeted frontend tests, sqlc regeneration, and API regeneration all pass. The inline comments cover six contract gaps and four test/maintainability gaps. Submitted as COMMENT only, without changing the approval state.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -704,56 +769,80 @@ function UsageCostPlaceholder() {
 
 function UsageMetrics({ totals }: { totals: SessionUsage["totals"] }) {
 	const { t } = useTranslation();
+	const cacheHitRate = formatCacheHitRate(totals.cachedInputTokens, totals.inputTokens);
 	return (
-		<dl className="grid grid-cols-2 gap-x-4 gap-y-2 @max-[300px]/inspector:grid-cols-1">
+		<dl className="grid grid-cols-2 gap-x-4 gap-y-2 @max-[300px]/inspector:grid-cols-1" data-testid="session-usage-metrics">
+			<UsageMetric label={t("inspector.usage.uncachedInputTokens")} metric={totals.uncachedInputTokens} />
 			<UsageMetric label={t("inspector.usage.inputTokens")} metric={totals.inputTokens} />
```
illegalcall: Spec mismatch: #4126 says not to display inclusive `inputTokens` as ordinary Input and asks for mutually exclusive Fresh input / Cache reads / Cache writes / Output rows, with reasoning explicitly labeled as an output subset. This renders Fresh Input beside overlapping Total Input and Cached Input while cache writes and reasoning are hidden, so the central ambiguity remains. Please render the requested mutually exclusive breakdown, or get the changed contract approved in the issue before merging.

backend/internal/observe/usage/parser.go
```diff
@@ -621,23 +669,93 @@ func isCodexContextFill(total codexTokenVector, modelContextWindow int64) bool {
...
+	if reportedTotal != 0 || input+output == 0 {
+		details.ReportedTotalTokens = int64Ptr(reportedTotal)
 	}
-	if tokens.CacheReadTokens > tokens.InputTokens-tokens.UncachedInputTokens ||
-		tokens.CacheWriteTokens > tokens.InputTokens-tokens.UncachedInputTokens-tokens.CacheReadTokens {
+	return metrics, details, true
+}
+
+func normalizeAnthropicUsage(directInput, cacheCreationInput, cachedInput, output int64, creation5m, creation1h *int64) (domain.UsageTokenMetrics, domain.AnthropicUsageDetails, bool) {
+	uncachedInput, ok := sumNonNegative(directInput, cacheCreationInput)
```
illegalcall, line 700: This makes the value surfaced as Fresh Input include Anthropic cache creation. For example, direct input 10 plus cache creation 3 becomes Fresh Input 13. #4126 requires Claude raw `input_tokens` to map to Fresh Input while cache writes remain separate. Please keep `directInput` as fresh and expose cache creation independently.

backend/internal/observe/usage/parser.go
```diff
@@ -621,23 +669,93 @@ func isCodexContextFill(total codexTokenVector, modelContextWindow int64) bool {
...
+	if input < 0 || cachedInput < 0 || cachedInput > input || cacheWriteInput < 0 ||
+		cacheWriteInput > input-cachedInput || output < 0 || reasoningOutput < 0 || reasoningOutput > output ||
+		reportedTotal < 0 || (reportedTotal != 0 && reportedTotal != input+output) {
+		return domain.UsageTokenMetrics{}, domain.OpenAIUsageDetails{}, false
 	}
-	if tokens.UncachedInputTokens > tokens.InputTokens ||
-		tokens.CacheReadTokens > tokens.InputTokens ||
-		tokens.CacheWriteTokens > tokens.InputTokens {
-		return false
+	uncachedInput := input - cachedInput
```
illegalcall, line 678: This derivation subtracts cached reads but leaves `cacheWriteInput` inside the value surfaced as Fresh Input. With input=100, cached=60, and cache-write=10 it reports 40 fresh rather than 30. #4126 asks for mutually exclusive categories, so cache writes need to remain separate from fresh input.

backend/internal/httpd/controllers/dto.go
```diff
@@ -967,24 +967,56 @@ type ListUsageSessionsQuery struct {
 
 // CompactSessionUsageResponse is one session card's token-only usage summary.
 type CompactSessionUsageResponse struct {
-	SessionID   domain.SessionID `json:"sessionId"`
-	TotalTokens int64            `json:"totalTokens" minimum:"0"`
-	Incomplete  bool             `json:"incomplete"`
+	SessionID       domain.SessionID `json:"sessionId"`
+	ProcessedTokens *int64           `json:"processedTokens" minimum:"0" description:"Canonical input plus output. Null when either component is unknown."`
```
illegalcall, line 973: Compatibility mismatch: #4126 explicitly asks to keep existing API fields temporarily and deprecate ambiguous naming. This replaces `totalTokens` with `processedTokens`, while the detailed DTO also removes `cacheReadTokens`, `cacheWriteTokens`, and `reasoningTokens`. Please retain deprecated aliases during the transition, or document an approved breaking-contract change.

frontend/src/renderer/components/SessionInspector.tsx
```diff
@@ -704,56 +769,80 @@ function UsageCostPlaceholder() {
...
+	if (
+		typeof cachedInputTokens !== "number" ||
+		!Number.isFinite(cachedInputTokens) ||
+		typeof inputTokens !== "number" ||
+		!Number.isFinite(inputTokens) ||
+		inputTokens <= 0
+	) {
+		return null;
+	}
+	const percentage = Math.min(100, Math.max(0, (cachedInputTokens / inputTokens) * 100));
```
illegalcall, line 843: The denominator is encoded only in this implementation; the UI exposes just `N% hit`, and the API does not describe the formula. #4126 asks for a documented denominator. Please surface or document that this is cached input divided by total inclusive input.

backend/internal/storage/sqlite/migrations/0102_canonical_usage.sql
```diff
@@ -0,0 +1,202 @@
...
+    uncached_input_tokens   INTEGER NOT NULL CHECK (uncached_input_tokens >= 0 AND uncached_input_tokens <= input_tokens),
+    cache_read_tokens       INTEGER NOT NULL CHECK (cache_read_tokens >= 0 AND cache_read_tokens <= input_tokens),
+    cache_write_tokens      INTEGER NOT NULL CHECK (cache_write_tokens >= 0 AND cache_write_tokens <= input_tokens),
+    output_tokens           INTEGER NOT NULL CHECK (output_tokens >= 0),
+    reasoning_tokens        INTEGER CHECK (reasoning_tokens IS NULL OR (reasoning_tokens >= 0 AND reasoning_tokens <= output_tokens)),
+    source_event_key        TEXT NOT NULL CHECK (trim(source_event_key) <> ''),
+    UNIQUE (binding_id, source_event_key)
+);
+
+CREATE TABLE model_usage_events_next (
```
illegalcall, line 61: Scope mismatch: #4126 says no historical data rewrite should be required because the mutually exclusive categories are already persisted. This rebuilds `model_usage_events` and adds provider detail, provenance, cached-output, pricing, and timestamp infrastructure. Unless this broader design was reapproved, derive `processedTokens` from the existing columns; otherwise document the deviation and why the rewrite is necessary.

backend/internal/storage/sqlite/store/usage_store.go
```diff
@@ -395,9 +398,22 @@ func (s *Store) ApplyUsageChunk(
 				if !usageEventMatches(existing, ev) {
 					return fmt.Errorf("%w: binding %d event %q", domain.ErrUsageSourceEventConflict, source.BindingID, ev.SourceEventKey)
 				}
+				if usageEventDetailsConflict(existing, ev) {
```
illegalcall: These new replay boundary branches are not exercised by the store tests. Existing coverage checks identical replay and a base-token conflict, but not provider-detail conflict or NULL-to-known enrichment. Please add transactional cases, ideally for both providers, that also assert cursor/event rollback behavior. This is required by the repository rule to test changed boundaries and validation paths.

backend/internal/storage/sqlite/store/usage_store.go
```diff
@@ -651,6 +805,17 @@ func ptrInt64ToNull(v *int64) sql.NullInt64 {
 	return sql.NullInt64{Int64: *v, Valid: true}
 }
 
+func stringToNull(v string) sql.NullString {
+	return sql.NullString{String: v, Valid: v != ""}
+}
+
+func timePtrUnlessZero(v time.Time) *time.Time {
```
illegalcall: Repository guidance says not to add abstractions for one-off use cases. `timePtrUnlessZero` has only one call site. Please inline this conversion, or use a shared adapter only if another real call site needs it.

backend/internal/domain/usage.go
```diff
@@ -108,49 +108,109 @@ type UsageSourceContext struct {
...
 // ModelUsageEvent is one append-only normalized usage fact.
 type ModelUsageEvent struct {
-	ModelID        string
-	Tokens         UsageTokenMetrics
-	SourceEventKey string
+	ProviderID      UsageProviderID
+	ModelID         string
+	Tokens          UsageTokenMetrics
+	ProviderDetails UsageProviderDetails
+	PricingVersion  string
```
illegalcall: Possible speculative generality: `PricingVersion` is threaded through the domain, migration, query, and store, but no producer sets it and no read-model consumer uses it; the controller test explicitly forbids exposing `pricingVersion`. Please remove this plumbing until pricing has a concrete consumer, or include the missing use case in this PR.

backend/internal/storage/sqlite/store/usage_store.go
```diff
@@ -550,44 +567,181 @@ func usageSourceInsertParams(rec domain.UsageSourceRecord) gen.InsertUsageSource
...
+			nullIntConflicts(existing.OpenaiReportedTotalTokens, details.ReportedTotalTokens)
+	}
+	details := event.ProviderDetails.Anthropic
+	return nullIntConflicts(existing.AnthropicDirectUncachedInputTokens, details.DirectUncachedInputTokens) ||
+		nullIntConflicts(existing.AnthropicCacheCreationInputTokens, details.CacheCreationInputTokens) ||
+		nullIntConflicts(existing.AnthropicCacheCreation5mInputTokens, details.CacheCreation5mInputTokens) ||
+		nullIntConflicts(existing.AnthropicCacheCreation1hInputTokens, details.CacheCreation1hInputTokens)
+}
+
+func usageEventNeedsDetailEnrichment(existing gen.GetModelUsageEventByKeyRow, event domain.ModelUsageEvent) bool {
```
illegalcall: This repeats the same OpenAI/Anthropic dispatch and field traversal as `usageEventDetailsConflict` immediately above. Consider one provider-dispatched reconciliation helper that computes conflict and enrichment together, so future provider fields do not require synchronized edits in two switches.

illegalcall (approved):

frontend/src/renderer/components/SessionInspector.tsx
whoisasx: Fixed in 300a0cd2b. The visible breakdown now uses mutually exclusive Fresh Input, Cache Reads, Cache Writes, and Output rows. Inclusive input is no longer rendered, and Reasoning is shown as an explicitly included output subset.

backend/internal/observe/usage/parser.go
whoisasx: Fixed in 300a0cd2b. Claude raw input_tokens now maps directly to Fresh Input with reported provenance, while cache_creation_input_tokens remains separate as cache writes. Parser and migration regressions cover the split.

backend/internal/observe/usage/parser.go
whoisasx: Fixed in 300a0cd2b. Codex Fresh Input is now derived as inclusive input minus cache reads minus cache writes. Cache writes remain separate, and storage validation enforces the three-part input breakdown.

Annieeeee11 (approved):
lgtm

backend/internal/httpd/controllers/dto.go
whoisasx: Addressed in 50bb4e9e0. The compact response retains totalTokens as a deprecated compatibility alias for processedTokens, and detailed totals retain cacheReadTokens, cacheWriteTokens, and reasoningTokens aliases while the canonical and provider-namespaced fields remain authoritative. OpenAPI and frontend types were regenerated.

frontend/src/renderer/components/SessionInspector.tsx
whoisasx: Addressed in 50bb4e9e0. The API schema now documents cachedInputTokens / inclusive inputTokens as the denominator, and the metric tooltip and accessible label spell out cached input / total input while preserving the concise visible hit-rate text.

backend/internal/storage/sqlite/migrations/0102_canonical_usage.sql
whoisasx: Documented in 50bb4e9e0. The database normalization is intentional for this task: SQLite cannot replace the legacy constraints or split provider-native counters in place. The migration is a structural one-to-one row copy that preserves persisted counters and does not reread transcripts or run a historical usage backfill.

backend/internal/storage/sqlite/store/usage_store.go
whoisasx: Addressed in 50bb4e9e0. Added transactional replay cases for both OpenAI and Anthropic covering provider-detail conflicts and NULL-to-known enrichment. The tests assert cursor rollback, partial-event rollback, successful cursor advancement, detail enrichment, and no duplicate token aggregation.

backend/internal/storage/sqlite/store/usage_store.go
whoisasx: Addressed in 50bb4e9e0. Removed timePtrUnlessZero and inlined the nullable UTC timestamp conversion at the single insert call site.

backend/internal/domain/usage.go
whoisasx: Addressed in 50bb4e9e0. Removed PricingVersion from the domain event, migration, queries, store matching, and generated sqlc artifacts because this PR has no producer or consumer for it.

backend/internal/storage/sqlite/store/usage_store.go
whoisasx: Addressed in 50bb4e9e0. Conflict and enrichment detection now share one provider-dispatched reconciliation traversal, so new provider fields have one comparison path to update.

PR #2488, opened 2026-07-07
 feat: bundle static tmux in macOS/Linux installers as a private fallback

Lucifer4255:
Closes #2443.

On macOS/Linux, session spawn is tmux-only, but the packaged app ships no tmux — a clean machine fails pre-spawn with `RUNTIME_PREREQUISITE_MISSING` (#2256/#2259 made that failure clear; this makes it go away). Implements the private bundled fallback discussed in the issue: never a system install, never mutating the user's machine.

## Resolution order (re-evaluated on every command)

1. `AO_TMUX_BIN` — explicit override; a set-but-broken value fails loudly instead of falling through
2. **System tmux from `PATH` — always wins when present.** Existing users see zero behavior change, and installing tmux later takes effect on the next spawn
3. App-bundled binary at `Resources/tmux-dist/tmux` (stamped into the daemon env as `AO_BUNDLED_TMUX` by the Electron supervisor; accepted only if an executable regular file)
4. The existing clear `RUNTIME_PREREQUISITE_MISSING` error (wire code unchanged, no API contract regen)

## What's in here

- **Backend**: new `tmux.ResolveBinary`/`NewResolver` shared by all consumers — the tmux runtime adapter (per-command resolution), the spawn prerequisite gate (`validateRuntimePrerequisites`), and `ao doctor` (PASS output now shows provenance: `via PATH | bundled | AO_TMUX_BIN`) — so they can never disagree on which binary is used. `AO_TMUX_BIN`/`AO_BUNDLED_TMUX` are read centrally in `config.Load()`.
- **Artifact pipeline**: manual-dispatch `.github/workflows/tmux-artifacts.yml` builds tmux 3.5a + libevent 2.1.12 + ncurses 6.5 from sha256-verified source tarballs, natively per target (macOS arm64/x64: static deps, dynamic only against libSystem; Linux x64: musl fully-static on alpine with compiled-in terminfo fallbacks), and publishes a `tmux-artifacts-v<ver>-<rev>` release with `checksums.txt` + provenance/license notes (tmux ISC, libevent BSD-3, ncurses MIT-X11).
[...]

PR #4164, opened 2026-08-20
feat(runtime): prefer bundled tmux on macOS

AgentWrapper:
## Summary

- centralize tmux executable resolution in `internal/tmuxbin`
- prefer an executable `tmux` next to the running `ao` binary (the mac app layout), then fall back to `tmux` from `PATH`
- make the runtime adapter use that resolution order
- make `ao doctor` report `bundled` versus `system`, and warn only when neither location resolves
- add unit coverage for bundled-first order, system fallback, missing tmux, and doctor provenance

Packaging is implemented separately in Untrivial-ai/ao-releases#29. The current Homebrew cask dependency is in AgentWrapper/homebrew-tap#1.

## Verification

- `npm run lint` with every `AO_*` environment variable unset
  - full `go test ./...`
- targeted tests: `go test ./internal/tmuxbin ./internal/adapters/runtime/tmux ./internal/cli`

## Scope and interaction with open work

- This does not duplicate or modify the startup/install UI in #3825. Once that PR is rebased, its tmux requirements probe should call the shared resolver so a signed bundled binary satisfies readiness; the UI remains useful on Linux or if both the bundle and PATH lookup fail.
- #3985 also touches tmux startup behavior but explicitly chose interactive installation instead of bundling. This PR leaves that UI/installer scope alone; its tmux adapter changes may require a small rebase conflict resolution.
- This supersedes the resolver/doctor portion of older #2488 using the current rewrite layout and the requested bundled-first order. #2488 targets retired release workflows and currently has failing CI.

## Risks

- The bundled lookup relies on the release layout keeping `ao` and `tmux` as siblings under `Contents/Resources/daemon`; the release artifact gate enforces that layout.
- A present but broken bundled binary is selected ahead of a working system tmux by design. `ao doctor` will fail its version probe and expose the exact bundled path; release signing and artifact verification are intended to prevent this state.
- macOS signing/notarization and x86_64 packaging are validated in the release-conductor PR, not in this backend-only PR.


backend/internal/adapters/runtime/tmux/tmux.go
```diff
@@ -237,13 +238,13 @@ func stableRunDir() string {
...
 // New builds a tmux Runtime, filling unset Options with defaults: binary "tmux"
-// (resolved via exec.LookPath), shell from $SHELL (else /bin/sh), and the
+// (bundled next to ao, then PATH), shell from $SHELL (else /bin/sh), and the
 // default timeout and output chunk size.
 func New(opts Options) *Runtime {
 	binary := opts.Binary
 	if binary == "" {
-		if path, err := exec.LookPath("tmux"); err == nil {
-			binary = path
+		if resolution, err := tmuxbin.Resolve(); err == nil {
```
AgentWrapper: **Blocker.** The runtime now prefers the bundled binary, but `Manager.validateRuntimePrerequisites` (`backend/internal/session_manager/manager.go:4370`) still gates every TUI spawn on `m.lookPath("tmux")` alone. On a clean mac with only a bundled tmux, `Spawn` returns `RUNTIME_PREREQUISITE_MISSING` at manager.go:753 before this constructor ever runs, so the bundle is unreachable in practice.

That gate needs the shared resolver too: `tmuxbin.ResolveWith(m.executable, m.lookPath)`. Both funcs are already fields on the Manager with `os.Executable`/`exec.LookPath` defaults (manager.go:337-341, 675-680), so no new plumbing is required. The error text and `manager_test.go:5091` need matching updates.

backend/internal/tmuxbin/resolve.go
```diff
@@ -0,0 +1,64 @@
...
+// the macOS app bundle, then falls back to the user's PATH.
+func Resolve() (Resolution, error) {
+	return ResolveWith(os.Executable, exec.LookPath)
+}
+
+// ResolveWith is Resolve with process lookups injected for CLI and unit tests.
+func ResolveWith(executable func() (string, error), lookPath func(string) (string, error)) (Resolution, error) {
+	var bundledErr error
+	if self, err := executable(); err == nil && self != "" {
+		candidate := filepath.Join(filepath.Dir(self), "tmux")
```
AgentWrapper: Two issues with deriving the bundle directory straight from `os.Executable`.

1. **No symlink resolution.** On darwin `os.Executable` is `_NSGetExecutablePath`, which returns the path used to exec the process, symlinks included -- Go does not realpath it. A shimmed or symlinked `ao` (which is what the Homebrew cask/tap installs) therefore looks for tmux next to the symlink, misses the bundle, and silently degrades to PATH. Running `filepath.EvalSymlinks(self)` first, falling back to `self` on error, fixes it.

2. **Inverse false positive.** An `ao` installed into `/usr/local/bin` sitting next to an Intel Homebrew `/usr/local/bin/tmux` matches this candidate and gets reported as `Source: bundled` for what is plainly a system tmux -- the exact distinction the doctor provenance is meant to make. Confirming the candidate really is inside the app bundle layout, or at least not just any PATH directory, would keep the label honest.

backend/internal/cli/doctor.go
```diff
@@ -310,21 +311,21 @@ func (c *commandContext) checkTerminalRuntime(ctx context.Context) doctorCheck {
 }
 
 func (c *commandContext) checkTmux(ctx context.Context) doctorCheck {
-	path, err := c.deps.LookPath("tmux")
-	if err != nil || path == "" {
-		return doctorCheck{Level: doctorWarn, Section: doctorSectionTools, Name: "tmux", Message: "not found in PATH; required on macOS/Linux to start sessions"}
+	resolution, err := tmuxbin.ResolveWith(c.deps.Executable, c.deps.LookPath)
```
AgentWrapper: Worth a note in the message or the docs: doctor resolves relative to the CLI `ao` the user invoked, while the tmux that actually runs sessions is resolved relative to the daemon `ao` under `Contents/Resources/daemon` (see `resolveDaemonLaunch` in `frontend/src/shared/daemon-launch.ts`). Someone running a Homebrew-installed `ao doctor` against a desktop-launched daemon can be told `system` while the daemon is happily using `bundled`, and vice versa. The label reads as global but is process-local.

backend/internal/tmuxbin/resolve_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if got.Path != "/usr/local/bin/tmux" || got.Source != SourceSystem {
+		t.Fatalf("resolution = %+v, want system tmux", got)
+	}
+	if want := []string{bundled, "tmux"}; !reflect.DeepEqual(lookups, want) {
+		t.Fatalf("lookups = %v, want %v", lookups, want)
+	}
+}
+
+func TestResolveWithReturnsErrorWhenTmuxIsUnavailable(t *testing.T) {
```
AgentWrapper: Coverage gap: the `else` branch at resolve.go:49 -- `executable()` returning an error, or returning an empty path -- is never exercised; all three tests supply a working `executable`. A case passing `func() (string, error) { return "", errors.New("no exe") }` and asserting the system fallback still wins would cover it, and would pin the intended behavior that a failed self-lookup must not block PATH resolution.

Nothing asserts the runtime side either: a test that `tmux.New(tmux.Options{})` picks up the resolved path would guard the wiring in tmux.go, which is where the user-visible behavior lives.

AgentWrapper (commented):
## Verdict: changes requested

The `tmuxbin` resolver reads cleanly, the doctor provenance change is a genuine improvement, and the unit coverage for bundled-first / system-fallback / neither-found is solid. One gap blocks the feature from working on the machines it targets.

### Blocker: session spawn is still gated on PATH-only tmux

`validateRuntimePrerequisites` (`backend/internal/session_manager/manager.go:4370`) still calls `m.lookPath("tmux")` and rejects every TUI spawn with `ErrRuntimePrerequisite` ("tmux required on macOS/Linux but not in PATH"). On exactly the machine this PR targets -- a clean mac with no system tmux and a bundled tmux next to the daemon binary -- `Manager.Spawn` fails at manager.go:753 before the tmux runtime is ever constructed, so the bundled binary never gets a chance to be used. The user sees `RUNTIME_PREREQUISITE_MISSING`.

The Manager already injects both `executable` and `lookPath` (manager.go:337-341, defaults wired at 675-680), so this is a small change: call `tmuxbin.ResolveWith(m.executable, m.lookPath)` there, reword the error to match the new lookup order, and update the expectation in `manager_test.go:5091`.

### Secondary

Inline comments cover: no symlink resolution before taking the sibling directory (breaks the Homebrew-shimmed `ao`, and mislabels a system tmux as bundled on Intel Homebrew layouts), doctor resolving against the CLI binary rather than the daemon binary, and a test gap on the executable-error branch.

### Risk worth a second look (not blocking)

The tmux adapter passes no `-L`/`-S`, so AO uses the default tmux socket, shared with any tmux the user runs directly. After this change the server on that socket can be a system tmux while the client is the bundled binary, or the reverse. tmux refuses a client/server protocol version mismatch outright, so shipping a bundled tmux whose PROTOCOL_VERSION differs from the installed one would hard-fail both AO and any tmux the user runs in their own terminal. A dedicated socket name for the bundled path would isolate AO from that entirely.

Docs (`frontend/src/landing/content/docs/plugins/runtimes/tmux.mdx`, `installation.mdx`) still tell macOS users to `brew install tmux`; fine to defer until packaging lands, but worth tracking.

backend/internal/adapters/runtime/tmux/tmux.go
AgentWrapper: Fixed in 0a4df6c3b. `validateRuntimePrerequisites` now calls `tmuxbin.ResolveWith(m.executable, m.lookPath)`, so a bundled-only install passes the pre-spawn gate. The missing-prerequisite text now reflects both lookup sources, and `TestValidateRuntimePrerequisites_AllowsBundledTmux` pins the clean-mac case.

backend/internal/tmuxbin/resolve.go
AgentWrapper: Fixed in 0a4df6c3b. Resolution now applies `filepath.EvalSymlinks` with fallback, and only probes a sibling bundled tmux when the resolved executable matches `*.app/Contents/Resources/daemon/ao`. Tests cover a symlink into the app and `/usr/local/bin/ao` next to system tmux remaining `SourceSystem`.

backend/internal/cli/doctor.go
AgentWrapper: Fixed in 0a4df6c3b. Doctor now labels the result as `bundled/system for this ao process`, making the process-local nature explicit rather than implying it describes a separately running daemon.

backend/internal/tmuxbin/resolve_test.go
AgentWrapper: Fixed in 0a4df6c3b. Resolver tests now cover both executable lookup error and empty-path fallbacks, and `TestNewResolvesTmuxFromPATH` directly asserts that `tmux.New(Options{})` stores the resolver-selected path.

AgentWrapper:
Addressed review [4980424461](https://github.com/Untrivial-ai/agent-orchestrator/pull/4164#pullrequestreview-4980424461) in 0a4df6c3b:

- the session-manager prerequisite gate now uses the shared bundled-first resolver, with a bundled-only regression test;
- executable symlinks are resolved before bundle lookup, and bundled provenance requires the macOS `.app/Contents/Resources/daemon` layout;
- doctor explicitly says the source applies “for this ao process”;
- tests cover executable lookup error/empty fallback and verify `tmux.New(Options{})` consumes the resolved PATH binary.

Validation: all focused packages pass, and `npm run lint` passes with all `AO_*` variables unset. I replied to and resolved all four addressed inline threads. I left tmux socket namespace isolation out of this review fix because it changes AO/user tmux server interoperability and deserves a separately scoped compatibility decision; the packaging PR remains responsible for nightly cross-architecture validation and docs follow-up.

backend/internal/cli/doctor.go
AgentWrapper: Addressed more explicitly in 08afa5d11. Doctor now labels the result as applying to “this ao doctor process” (including the missing-tmux case), and the tmux runtime docs explain that a separately running desktop daemon resolves relative to its own executable, may select a different source, and is the resolution sessions actually use. Tests pin the process-local wording.

backend/internal/tmuxbin/resolve_test.go
AgentWrapper: Addressed in pushed commit 0a4df6c3b and still present on current head 08afa5d11. `TestResolveWithFallsBackWhenExecutableIsUnavailable` covers both `executable()` returning `errors.New("no executable")` and returning an empty path, asserting the system tmux fallback wins. `TestNewResolvesTmuxFromPATH` calls `tmux.New(Options{})` with a controlled PATH and asserts the runtime stores the resolved binary. I reran both exact tests with all `AO_*` variables unset; both pass.

backend/internal/adapters/runtime/tmux/tmux.go
AgentWrapper: Addressed in pushed commit 0a4df6c3b and still present on current head 08afa5d11. `Manager.validateRuntimePrerequisites` now calls `tmuxbin.ResolveWith(m.executable, m.lookPath)`, the error says no bundled or system tmux was found, and `TestValidateRuntimePrerequisites_AllowsBundledTmux` proves a bundled-only clean-mac layout passes the gate. `TestSpawn_RejectsMissingTmuxBeforeSessionRow` pins the neither-found error. I reran both exact tests with all `AO_*` variables unset; they pass.

backend/internal/tmuxbin/resolve.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	} else if err != nil {
+		bundledErr = err
+	}
+
+	path, err := lookPath("tmux")
+	if err == nil && path != "" {
+		return Resolution{Path: path, Source: SourceSystem}, nil
+	}
+	if err == nil {
```
AgentWrapper, line 74: Now redundant: line 62 appends `errTmuxNotFound` unconditionally, so this branch only makes the joined message read `tmux executable not found` twice in the case where `lookPath` returns an empty path with a nil error. Dropping lines 59-61 keeps the error text clean without changing behavior.

backend/internal/tmuxbin/resolve.go
```diff
@@ -0,0 +1,80 @@
...
+
+// bundledCandidate recognizes the macOS application layout populated by the
+// release pipeline. Restricting bundled provenance to this layout prevents a
+// system ao and tmux that merely share a bin directory from being mislabeled.
+func bundledCandidate(self string) (string, bool) {
+	daemonDir := filepath.Dir(self)
+	resourcesDir := filepath.Dir(daemonDir)
+	contentsDir := filepath.Dir(resourcesDir)
+	appDir := filepath.Dir(contentsDir)
+	if filepath.Base(daemonDir) != "daemon" ||
```
AgentWrapper: Scope note rather than a defect: this gate is macOS-only by construction, and correctly so for this PR. Worth knowing for the next packaging change -- Electron on Linux puts the daemon under `resources/daemon` (lowercase `resources`), so if Linux ever bundles tmux this returns false and the bundle is skipped with no signal anywhere. The doc comment does say macOS, which is honest; adding that Linux is deliberately excluded, or gating on `runtime.GOOS`, would make the omission obvious at the point someone adds Linux packaging.

backend/internal/session_manager/manager.go
```diff
@@ -4367,8 +4368,8 @@ func (m *Manager) validateRuntimePrerequisites() error {
 	if runtime.GOOS == "windows" {
 		return nil
 	}
-	if path, err := m.lookPath("tmux"); err != nil || path == "" {
-		return fmt.Errorf("%w: tmux required on macOS/Linux but not in PATH", ports.ErrRuntimePrerequisite)
+	if resolution, err := tmuxbin.ResolveWith(m.executable, m.lookPath); err != nil || resolution.Path == "" {
```
AgentWrapper: Correct fix, and the new test pins it.

Small follow-up: the doc comments on `executable` (manager.go:339-341) and `Deps.Executable` (manager.go:593-596) still describe the field as existing only for the session PATH pin. It now also decides whether a bundled tmux satisfies the spawn gate, which means a test that stubs `Executable` for PATH-pin reasons is silently changing tmux resolution too. A sentence in both comments would save the next reader that surprise.

AgentWrapper (commented):
## Verdict: approved

All four findings from the previous round are fixed, each with a test that would catch a regression. Nothing here blocks merge.

### What was verified

- **Spawn gate (the blocker).** `validateRuntimePrerequisites` now calls `tmuxbin.ResolveWith(m.executable, m.lookPath)` (manager.go:4371), and `TestValidateRuntimePrerequisites_AllowsBundledTmux` covers the clean-mac case that previously failed. `Deps.Executable` was already wired through to `m.executable` (manager.go:634, default at 680), so the daemon resolves against its own binary as intended.
- **Symlinks.** `filepath.EvalSymlinks` with a fallback to the raw path (resolve.go:41-43) is the right shape: the fallback is what keeps the pre-existing tests green, since they point at a non-existent `/Applications` path where `EvalSymlinks` fails. `TestResolveWithFollowsExecutableSymlinkIntoAppBundle` covers the Homebrew-shim case.
- **Bundled mislabeling.** `bundledCandidate` restricting provenance to the `*.app/Contents/Resources/daemon` layout matches what `resolveDaemonLaunch` actually populates (`frontend/src/shared/daemon-launch.ts:56`), and `TestResolveWithDoesNotTreatSiblingSystemTmuxAsBundled` pins the `/usr/local/bin` false positive. The `errors.Join(..., errTmuxNotFound)` also removes the old path where a `("", nil)` lookup could produce a nil error.
- **Coverage gap.** Both the executable-error and empty-path branches are covered, and `TestNewResolvesTmuxFromPATH` pins the runtime wiring.
- **Docs.** The tmux.mdx paragraph explains the process-local scope clearly.

I also checked that no other consumer depends on the changed strings: `service_test.go:2404` builds its own error text for the API mapping table and is unaffected, and the renderer matches only on the `RUNTIME_PREREQUISITE_MISSING` code, not the message.

### Still open from the last round (not blocking this PR)

The tmux adapter passes no `-L`/`-S`, so AO uses the default tmux socket, shared with any tmux the user runs directly. Once a bundled tmux of a different version can serve or connect to that socket, a PROTOCOL_VERSION difference is an outright refusal on both sides. Nothing in this PR makes that worse and deferring is reasonable, but it should be settled before a release actually ships the bundled binary.

Three minor notes inline, all optional.

harshitsinghbhandari:
Non-blocking: bundled and system tmux share the default socket, so after an update the daemon can point a bundled client at a server the user's system tmux started. tmux then prints `protocol version mismatch (client N, server M)`, which matches neither `sessionMissingOutput` nor `serverUnreachableOutput` (tmux.go:956-969), so `IsAlive` returns a generic probe error and sessions stall with no actionable message. Nothing gets archived, so it is safe, just opaque. Adding `"protocol version mismatch"` to `serverUnreachableOutput` maps it to `ErrRuntimeUnavailable` and surfaces the cause. Worth a line in Risks either way.

PR #4165, opened 2026-08-20
feat: show pull request author avatars

AgentWrapper:
## Summary

- capture GitHub PR author avatar URLs during REST discovery and GraphQL refreshes
- persist the optional URL through SQLite, session summaries, daemon/cloud OpenAPI contracts, and renderer types
- render authoritative PR-author URLs through a shared provider-neutral avatar component that preserves initials while loading, offline, after errors, and when no URL exists
- reuse the same loading/error fallback for reviewer cards and Kanban reviewer stacks; keep the existing login-derived `GithubAvatar` API compatible and centralize provider fallback URL construction
- allow only `https://avatars.githubusercontent.com` in the renderer image CSP; enterprise hosts continue to degrade to initials

## Screenshots

Real GitHub avatar:

PR card with GitHub author avatar

Monogram fallback (no avatar URL):

PR card with author monogram fallback

## Tests

- `go test ./...` with inherited `AO_*` variables unset
- golangci-lint v2.12.2 with an isolated cache: 0 issues
- renderer: 157 test files, 2320 tests passed
- product-ui: typecheck, 122 tests, and production build
- frontend typecheck and focused PR/Kanban renderer tests: 49 passed
- cloud client: schema drift check, typecheck, 21 tests, and build
- regenerated sqlc, daemon OpenAPI, frontend schema, and cloud schema with no drift

## Risk

Low. Existing and non-GitHub records keep an empty optional URL and render the current monogram. Avatar fetch failures degrade to that same fallback; Chromium's HTTP cache handles repeat renders without a custom cache. The PR was rebased onto `main` after the Kanban reviewer-avatar work and now shares its presentation primitives instead of duplicating image behavior.


AgentWrapper (commented):
**Verdict: changes requested** — two small items; the approach itself is right.

The plumbing is clean and complete end to end: REST + GraphQL capture, ports/domain/sqlc/store, DTO + apispec + generated frontend schema. `observationFromLocal` was updated too, so a locally-rebuilt observation still hashes identically — easy thing to miss. Migration 0102 follows the existing add-column convention and the burned-versions map was updated. Good catch that the old `https://github.com/<login>.png` source was never allowed by the renderer CSP (img-src had no github.com entry), so packaged builds were already falling back to initials; pointing at avatars.githubusercontent.com with an explicit allowlist entry is the right fix.

### Please address

1. **Shared SCM contract drift.** `contract.PullRequestSummary` is mirrored by `contracts/cloud/openapi.yaml#/components/schemas/PullRequestSummary`, which sets `additionalProperties: false`, and by `packages/product-ui/src/scm-models.ts#PullRequestSummary`. Adding `authorAvatarUrl` to the Go struct alone means the cloud read model now emits a property its own schema rejects, and cloud/mobile clients cannot type or use it. The daemon apispec was updated correctly; these two mirrors need the same optional property.

2. **Reviewer avatars silently lost their image.** `GithubAvatar` no longer derives a URL from the login, and the other call site — `packages/product-ui/src/SessionInspectorView.tsx:880`, the GitHub review cards — passes no `avatarUrl`, so those render initials permanently. Packaged builds already showed initials there because of the CSP, but dev and `ao preview` did render real photos, and now nothing will ever populate them. Either thread a reviewer avatar URL through the review observation, or say in the PR description that reviewer avatars are intentionally monogram-only for now.

### Minor

- The monogram stays painted underneath the loaded image, so a transparent PNG avatar shows the letters through it. Consider not rendering the initials once `loaded`.
- The CSP host is github.com-only, so GitHub Enterprise hosts (`*.ghe.io` and `*.github.com`, which `parsePRURL` accepts) always fall back to the monogram. Fine as a decision — worth a short comment next to the allowlist entry so the next reader does not treat it as an oversight.
- `mock-data.ts` preview data now depends on a live third-party fetch of a real account instead of being self-contained.

Test coverage is solid: store round-trip, DTO mapping, REST and GraphQL fixtures, and loaded/error/no-URL component states.

backend/pkg/contract/scm.go
```diff
@@ -144,6 +144,7 @@ type PullRequestSummary struct {
 	Provider         string                         `json:"provider"`
 	Repo             string                         `json:"repository"`
 	Author           string                         `json:"author"`
+	AuthorAvatarURL  string                         `json:"authorAvatarUrl,omitempty"`
```
AgentWrapper, line 148: This struct is the shared SCM read model, mirrored by `contracts/cloud/openapi.yaml#/components/schemas/PullRequestSummary` (which sets `additionalProperties: false`) and by `packages/product-ui/src/scm-models.ts#PullRequestSummary`. Adding the field here without adding it there means the cloud payload now carries a property its own schema forbids, and cloud/mobile clients cannot see it. Please add the optional `authorAvatarUrl` to both mirrors.

packages/product-ui/src/GithubAvatar.tsx
```diff
@@ -17,34 +18,30 @@ function initials(login: string): string {
 		.join("") || "?";
 }
 
-export function GithubAvatar({ login, className }: GithubAvatarProps) {
+export function GithubAvatar({ login, avatarUrl, className }: GithubAvatarProps) {
```
AgentWrapper, line 10: Dropping the login-derived source changes behaviour for the other call site of this component: `SessionInspectorView.tsx:880` renders reviewer avatars as `<GithubAvatar login={entry.reviewerId} />` with no `avatarUrl`, so GitHub review cards now show initials permanently. Packaged builds already fell back to initials there (the old github.com URL was outside img-src), but dev and `ao preview` did render photos. Either thread a reviewer avatar URL through the review observation, or call out in the PR description that reviewer avatars are monogram-only by design.

packages/product-ui/src/GithubAvatar.tsx
```diff
@@ -17,34 +18,30 @@ function initials(login: string): string {
...
+	const normalizedAvatarUrl = avatarUrl?.trim() ?? "";
+	const loaded = loadedUrl === normalizedAvatarUrl;
 
 	return (
 		<span
 			aria-hidden="true"
-			className={cn("inline-flex size-icon-sm shrink-0 items-center justify-center rounded-full bg-muted text-micro font-semibold text-muted-foreground", className)}
+			className={cn("relative inline-flex size-icon-sm shrink-0 items-center justify-center overflow-hidden rounded-full bg-muted text-micro font-semibold text-muted-foreground", className)}
 		>
 			{initials(normalizedLogin)}
```
AgentWrapper: Minor: the initials stay in the DOM behind the image rather than being replaced, so a user avatar with transparency renders the monogram through it. Rendering `{loaded ? null : initials(normalizedLogin)}` keeps the pre-load and error fallbacks intact without the bleed-through.

frontend/vite.renderer.config.ts
```diff
@@ -50,7 +50,7 @@ const CONTENT_SECURITY_POLICY = [
 	"default-src 'self'",
 	"script-src 'self'",
 	"style-src 'self' 'unsafe-inline'",
-	"img-src 'self' data: http://127.0.0.1:*",
+	"img-src 'self' data: http://127.0.0.1:* https://avatars.githubusercontent.com",
```
AgentWrapper, line 55: Minor: `parsePRURL` accepts GitHub Enterprise hosts (`*.ghe.io`, `*.github.com`), which serve avatars from the instance host rather than avatars.githubusercontent.com, so those users always get the monogram. Reasonable tradeoff for a static CSP, but worth a one-line comment here so it reads as a decision rather than an oversight.

frontend/src/renderer/lib/mock-data.ts
```diff
@@ -225,7 +225,8 @@ const prSummary = (sessionId: string, number: number, overrides: Partial<Session
 		state: facts?.state ?? "open",
 		provider: "github",
 		repo: `me/${session?.workspaceName ?? "preview"}`,
-		author: "preview-agent",
+		author: "octocat",
+		authorAvatarUrl: "https://avatars.githubusercontent.com/u/583231?v=4",
```
AgentWrapper: Minor: preview/mock data now performs a live fetch to a third-party host and displays a real GitHub account. A `data:` URI (already allowed by img-src) would keep the preview deterministic and offline-safe while still exercising the loaded state.

AgentWrapper:
Addressed review 4989226426 in commit 45743b2f8:

- added optional authorAvatarUrl to the Cloud OpenAPI mirror and product-ui SCM model, then regenerated the checked-in cloud client schema
- documented in the PR description that reviewer cards remain monogram-only because review observations do not currently carry reviewer avatar URLs
- hide initials after a successful avatar load (with regression coverage)
- documented the intentional GitHub Enterprise CSP fallback
- replaced live preview avatar data with a deterministic inline data URI

Verified with cloud:check, product-ui:check, frontend typecheck, and the focused renderer avatar tests.

backend/pkg/contract/scm.go
AgentWrapper: Added optional authorAvatarUrl to contracts/cloud/openapi.yaml and packages/product-ui/src/scm-models.ts, then regenerated packages/cloud-client/src/schema.ts. Verified with npm run cloud:check.

packages/product-ui/src/GithubAvatar.tsx
AgentWrapper: Kept reviewer cards intentionally monogram-only because review observations do not currently expose reviewer avatar URLs, and added an explicit Reviewer avatar scope section to the PR description. This PR remains scoped to PR-author avatars.

packages/product-ui/src/GithubAvatar.tsx
AgentWrapper: Updated GithubAvatar to remove the initials once the image loads, while restoring them on error. Added assertions for both transitions.

frontend/vite.renderer.config.ts
AgentWrapper: Added a comment next to img-src documenting that arbitrary GitHub Enterprise avatar hosts are intentionally excluded and retain the monogram fallback.

frontend/src/renderer/lib/mock-data.ts
AgentWrapper: Replaced the live octocat URL in preview data with a self-contained SVG data URI, preserving loaded-state coverage while making preview deterministic and offline-safe.

AgentWrapper (commented):
**Verdict: approved** — re-reviewed at `45743b2`, every item from the previous round is resolved.

- **Contract drift (blocking item) — fixed.** `authorAvatarUrl` is now present in `contracts/cloud/openapi.yaml#/components/schemas/PullRequestSummary`, `packages/product-ui/src/scm-models.ts`, and the regenerated `packages/cloud-client/src/schema.ts`. It is correctly absent from the schema `required` list, so it lines up with the `omitempty` Go tag and the optional `authorAvatarUrl?: string` in both TS mirrors, and the payload no longer trips `additionalProperties: false`.
- **Reviewer avatars — resolved as scope.** The new `Reviewer avatar scope` section in the description states plainly that review cards stay monogram-only because review observations do not carry reviewer avatar URLs. That was the alternative I offered; recording the decision is enough.
- **Monogram bleed-through — fixed.** `{loaded ? null : initials(normalizedLogin)}` drops the letters once the image paints and restores them on error, and the component test now asserts both transitions. The wrapper is sized by `size-icon-sm` plus the caller override rather than by its text child, so removing the in-flow content does not collapse the circle.
- **CSP comment — added.** The note next to `img-src` makes the enterprise-host exclusion read as a decision.
- **Preview data — fixed.** The inline SVG `data:` URI keeps the preview deterministic and offline-safe while still exercising the loaded state, and `data:` is already allowed by the renderer CSP.

Nothing new surfaced in this pass. The rest of the change still looks right: `observationFromLocal` carries the field so locally-rebuilt observations hash identically, migration 0102 matches the add-column convention with the burned-versions map updated, the upsert placeholder count matches the column list, and the CSP entry fixes author avatars that packaged builds were silently dropping before.

neversettle17-101:
It will be addressed in this: https://github.com/Untrivial-ai/agent-orchestrator/pull/4764 
closing this PR.

PR #4169, opened 2026-08-20
fix(browser): crash on link click, agent-visible error signals, device-frame presets

Vaibhaav-Tiwari:
## Summary
Phase 3 of the AO in-app browser UX overhaul (agent visibility & frontend-dev ergonomics):

- **Fixed an app-crashing bug**: clicking any link with `target=_blank` (or `window.open()`) inside the browser crashed the whole app with `Invalid webContents. Created window should be connected to webContents passed with options object.` `setWindowOpenHandler`'s `createWindow` callback returned an independently-created `WebContentsView` that Electron's own `openGuestWindow` could never recognize as linked to the guest-window options it computed for that event. Now denies the guest window outright and opens (and actually navigates — it never did before, even before the crash) AO's own tab instead, entirely outside Electron's guest-window machinery.
- **Console errors and failed requests now reach the agent, always-on**: previously only visible via on-demand polling. Watches every tab's native `console-message` event (no CDP/debugger needed, so it can't conflict with agent-browser's own CDP session) and each session's shared `webRequest` for failed/errored XHR+fetch calls. Debounces a burst into one message per ~5s window and posts it through the same `POST /sessions/{id}/send` endpoint the annotation feature already uses — no new data captured beyond what's already visible in DevTools (text, method/url/status — never bodies).
- **Fixed tabs getting permanently stuck open**: reported live from a real long-running session as "Tab t5 not found; run `agent-browser tab` to list open tabs" on close. The external `agent-browser` automation binary keeps its own internal tab registry that can drift from AO's `session.tabs` over a long session; the close handler used to just propagate that failure and give up, leaving the tab stuck forever. Now falls back to AO's own close path (which only depends on `session.tabs` and the real `WebContentsView`, not the runtime's registry) for that specific failure; unrelated automation-runtime errors still surface normally.
- **Device-frame presets, redesigned**: one-click toolbar dropdown for responsive testing. Named real devices (iPhone SE/14 Pro, Pixel 7, Galaxy S22, iPad Mini/Pro) with reference dimensions shown, plus an inline custom-width field — no arbitrary "Mobile/Tablet/Desktop" buckets, no redundant Desktop option. Also fixes the dropdown rendering *behind* the live page (missing the `data-browser-native-overlay` marker that tells the shell to paint above the native view while it's open — the same mechanism the tabs-rail flyout already relies on).

## Test plan
- [x] `npx vitest run` on `browser-view-host.test.ts`, `BrowserPanel.test.tsx`, `i18n/instance.test.ts` — all passing
- [ ] Manual pass in the real Electron app — link-click crash fix, always-on signal streaming, and the tab-registry-drift fallback all depend on real Chromium/native-process/external-binary behavior that can't be verified by automated tests in this environment


Prasad-D-Ware:
image
we should support these ones if we are adding them.

image
the logo coloring is a bit odd.

image
when opened it turns the screen black on fullscreen mode

Vaibhaav-Tiwari:
Thanks for the screenshots — all three addressed, pushed \`4691fb26c\` and \`39a4b677a\`.

**Device list expanded to match Chrome's own.** Pulled the actual list + dimensions from Chrome's source (\`front_end/models/emulation/EmulatedDevices.ts\`) rather than approximating — iPhone SE/XR/12 Pro/14-16 Pro Max, Pixel 7-10, Samsung Galaxy S8+/S20 Ultra/A51/71, iPad Mini/Air/Pro, Surface Pro 7/Duo, Galaxy Z Fold 5, Asus Zenbook Fold, both Nest Hub sizes. iPad Air and Nest Hub have since been dropped from Chrome's *current* list but I kept them since they're still common/well-known and your screenshot shows them. The list is scrollable now that it's much longer. Also confirmed the width×height shown for each is that device's real portrait (or fixed-orientation, for the two Nest Hubs) value from Chrome's source, not a guess.

**Icon coloring fixed** — the trigger was using a different, weaker active-state style than the DevTools button right next to it. Switched to the same convention (\`bg-accent-strong\`/\`text-accent-foreground\`) so they're visually consistent.

**Black screen in maximized mode — fixed, but I need you to verify.** Traced it to the same class of bug \`window-composition.ts\` already documents and works around for its own shell view: on macOS, reordering a \`WebContentsView\` to raise it can leave its *previous* compositor surface on screen until some real geometry change rebuilds it, and re-applying identical bounds is a no-op Electron ignores — the existing fix only ever refreshed the shell's own surface, never the live page's. Added \`refreshLastFocusedPanelSurface\`, applying the same "shrink by 1px, restore next tick" nudge to the live page's view, called right after the shell gets raised for any overlay.

I want to flag plainly: this is a macOS-specific compositor quirk per the existing code's own comments, and I'm on Windows here, so I could not reproduce or verify this live myself — it's my best-informed fix based on the exact documented pattern for the analogous shell-surface bug, not something I watched work. I'd appreciate it if you could confirm on your machine that it actually resolves the black screen before we call this one closed.

\`npm --prefix frontend run typecheck\` clean, 125/125 across the three suites, including a new test confirming the full device list and a test for the surface-refresh nudge (verified against a fresh mock, not a live repro given the above caveat).

Prasad-D-Ware:
image


the black screen is still visible for me. check that part. 
the addition of supported device is great now.

Vaibhaav-Tiwari:
Thanks — glad the device list is working well. Pushed \`63180891e\` for the black screen.

**What I found investigating further:** first ruled out two things directly from the code (no macOS access to test live, so I wanted to be sure before guessing again):
- The device-frame wrapper isn't it — your screenshot shows the black area spanning the *full* viewport width, not the ~414px iPhone-XR-constrained width that would show if my wrapper were painting it.
- The CSS transparency cascade isn't it either — `data-native-browser-composition` is set unconditionally by the preload on every load, and `data-browser-native-page="live"` is set whenever a real URL is loaded (both true in your screenshot), so the `!important` transparent rules on `.browser-panel__viewport` should already be winning by specificity.

That leaves the native compositor, same as `window-composition.ts`'s own documented bug for the shell view: reordering a `WebContentsView` above others can leave its *previous* surface on screen until something forces a real re-composite, and identical bounds are a no-op Electron ignores. My first attempt only nudged bounds by 1px for the live view (mirroring the shell's own fix) — evidently not a strong enough signal on its own. This round also toggles visibility off/on around the bounds nudge, which is a more direct signal for Electron to re-establish the surface.

**I want to be upfront**: this is a macOS-specific compositor quirk, I'm on Windows, and my first attempt at fixing it *didn't* actually resolve it for you — so I can't promise this one does either without you confirming. If it's still black after this, it'd help a lot to know: does resizing the window at all clear it? Does switching tabs? Does closing and reopening the dropdown? Any of those "yes" would point at a different, more specific trigger than the general "shell got raised" moment I'm nudging on right now.

\`npm --prefix frontend run typecheck\` clean, existing + updated tests passing (the test now also asserts the visibility toggle happens, not just the bounds nudge).

Prasad-D-Ware (approved):

PR #4173, opened 2026-08-20
feat(diffs): render image files as before/after previews

Pyasma:
## Summary

- The diff viewer printed "Binary file preview is not available." for every image, so a changed PNG or a new screenshot in a worktree could only be reviewed by opening the file outside AO. Now it shows the images themselves, before and after.
- Added `GET /sessions/{sessionId}/workspace/file/blob?path=&side=before|after`, which serves one revision of a file as raw bytes. `before` reads it out of the compare base with `git cat-file` (following the rename source when git recorded one), `after` reads the worktree copy. Images only, 16 MB cap, `no-store`.
- Pulled the path resolution out of `GetWorkspaceFile` into a shared helper so the blob route and the file detail agree on which worktree owns a path, what prefix it carries in a workspace project, and which commit is the base.
- The detail response gained `imageMediaType`, set only when the file is binary or deleted — an SVG is text, so it keeps its line diff.
- New `ImageDiffView` draws the two sides on a checkerboard with their pixel dimensions, one pane for added/deleted files, and a fallback message if a side fails to load. Image URLs carry the detail query's load timestamp; the route is `no-store`, so without a changing URL an edited image would never refetch.

## Branch contents

This branch carries three small chrome fixes ahead of the image feature, all of which the image diff sits on top of and none of which have their own PR open:

- `refactor(frontend): remove sidebar hover preview`
- `fix(desktop): keep Linux titlebar cluster clear of the framed panel`
- `fix(board): inset the lane grid inside the framed panel`

The last commit, `fix(diffs): address review on the image preview PR`, applies the review feedback: `gitWorkspaceBlob` now asks git for the object size before reading the bytes, each image pane is keyed by its source so a failed load recovers when the file changes, and the dead `--z-sidebar-preview` / `--z-index-sidebar-preview` tokens left by the hover-preview removal are gone.

## Test plan

- Go, 8 new: media-type detection, SVG staying text, both blob sides, added → no before, deleted → no after, rename reading the previous path, non-image rejection, invalid `side`. Plus route tests for content type, `no-store`, the raw body, and the missing-`path` 400.
- Renderer, 4 new: both panes on a modified image, after-only on added, before-only on deleted, and the per-side load-error fallback.
- `npm run lint` (golangci-lint: 0 issues), `npm run frontend:typecheck`, full renderer suite (196 files / 2578 tests) all pass. The `internal/terminal` tmux integration failures on my machine are pre-existing and untouched here.
- Ran the Electron app from this checkout against isolated dev data with a seeded repo: modified PNG shows grayscale before → colour after with `320 × 200` captions, an added `.webp` came back byte-identical to the source file, and added/deleted images render a single pane.


Vaibhaav-Tiwari (changes requested):
1. Backend memory issue: When loading the "before" image (from git history), the
  code reads the entire file into memory before checking if it's too big — even
  though a comment right above it says "refused rather than buffered." The "after"
  image path does this correctly (checks size first). A huge image in git history
  could balloon memory use.
  → backend/internal/service/session/workspace_files.go:369-378
  2. Frontend UI bug: If an image fails to load even once, that preview panel is
  broken forever for that file — even after the file is fixed/re-saved. There's no
  code to reset the "failed" state when the image changes.
  → frontend/src/renderer/components/ImageDiffView.tsx
  3. Leftover dead code: This PR removes the "sidebar hover preview" feature but
  forgets to delete 2 CSS variables that nothing uses anymore.
  → tokens.css:230, styles.css:89

  Process issue

  This PR is titled "render image files as before/after previews" but actually
  bundles 4 unrelated changes into one PR (sidebar preview removal, titlebar fix,
  board grid fix, AND the image feature). Repo rules say one issue per PR — worth
  splitting or at least calling out in the description.

Vaibhaav-Tiwari (approved):

PR #4183, opened 2026-08-20
feat: add Kimi token usage observability

nikhilachale:
## What

Add the focused Kimi layer on top of the shared usage foundation in #4182.

- Parse Kimi wire events into normalized token usage.
- Enable Kimi as a certified usage harness.
- Discover main and subagent wire files from the append-only session index.
- Fence attribution to the canonical Kimi sessions root.
- Replay large session indexes and reject deleted or escaping entries.

## Stack

- Base: #4182
- This PR: Kimi
- Next: feat/pi-token-usage-observability
- Then: feat/qwen-token-usage-observability

This replaces the Kimi portion of #3778.

## Testing

- go test ./internal/observe/usage ./internal/service/usage

PR #4184, opened 2026-08-20
feat: add Pi token usage observability

nikhilachale:
## What

Add the focused Pi layer on top of the Kimi stack in #4183.

- Parse Pi session events into normalized token usage.
- Enable Pi as a certified usage harness.
- Discover sessions by native header ID and correlate missing IDs through a unique live workspace.
- Retry late-created Pi artifacts and cap discovery to the newest files.
- Reactivate usage discovery when lifecycle hooks persist the Pi native session ID without blocking concurrent finalization.

## Stack

- Base: #4183
- This PR: Pi
- Next: feat/qwen-token-usage-observability

This replaces the Pi portion of #3778.

## Testing

- go test ./internal/observe/usage ./internal/service/usage ./internal/lifecycle

### DIFF
--- a/backend/internal/service/usage/collector_test.go
+++ b/backend/internal/service/usage/collector_test.go
@@ -1913,6 +1913,234 @@ func TestCollectorRejectsKimiWireOutsideCanonicalSessionsDirectory(t *testing.T)
 	}
 }
 
+func TestCollectorDiscoversPiSessionByHeaderID(t *testing.T) {
+	const nativeID = "pi-session-1"
+	store := collectorTestStore(t)
+	session := collectorTestSession(t, store, domain.HarnessPi, nativeID, false)
+	root := t.TempDir()
+	path := filepath.Join(root, "project", "2026-08-09_"+nativeID+".jsonl")
+	writeUsageFixture(t, path, `{"type":"session","id":"`+nativeID+`","cwd":"/repo","version":3}`+"\n")
+	collector := NewCollector(store, SourceRoots{PiSessions: root}, nil)
+
+	mustNoError(t, collector.RecordHook(context.Background(), session.ID, HookSignal{
+		Harness: domain.HarnessPi, Event: "session-start", NativeSessionID: nativeID,
+	}))
+	bindings, _ := store.ListUsageBindingsForSession(context.Background(), session.ID)
+	if len(bindings) != 1 {
+		t.Fatalf("bindings = %+v", bindings)
+	}
+	sources, err := store.ListUsageSourcesForBinding(context.Background(), bindings[0].ID)
+	if err != nil || len(sources) != 1 || sources[0].Kind != domain.UsageSourcePiSession {
+		t.Fatalf("sources=%+v err=%v", sources, err)
+	}
+}
+
+func TestCollectorPiDiscoveryCapsNewestFilesByModificationTime(t *testing.T) {
+	const nativeID = "pi-newest-session"
+	root := t.TempDir()
+	newest := filepath.Join(root, "a-project", "000-newest.jsonl")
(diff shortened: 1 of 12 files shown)


vanshx999:
he right‑hand sidebar isn't showing Pi token usage after starting a session. The daemon compiles and launches from feat/pi-token-usage-observability, and the app loads, but no InputTokens/OutputTokens etc. appear in the UI after agent activity. Debug needed: either the Pi wire events aren't reaching the parser, or the UI hook for Pi usage isn't connected. Attaching screenshot of empty sidebar
Screenshot 2026-08-21 at 1 22 06 AM

vanshx999:
LGTM — Pi token usage observability is working end‑to‑end. The daemon builds and launches from feat/pi-token-usage-observability, and after starting a Pi agent session the right‑hand sidebar correctly displays InputTokens, UncachedInputTokens, CacheReadTokens, CacheWriteTokens, and OutputTokens. The collector also properly discovers main/sub‑agent wires from the session index and rejects deleted/escaping sessions. (Attaching screenshot of sidebar with Pi tokens.)
Screenshot 2026-08-21 at 1 52 55 AM

backend/internal/service/usage/collector_test.go
```diff
@@ -1913,6 +1913,299 @@ func TestCollectorRejectsKimiWireOutsideCanonicalSessionsDirectory(t *testing.T)
...
+	if wakes != 1 {
+		t.Fatalf("reconcile wakes = %d, want 1", wakes)
+	}
+	if bindings, err := store.ListUsageBindingsForSession(context.Background(), session.ID); err != nil || len(bindings) != 0 {
+		t.Fatalf("early bindings=%+v err=%v", bindings, err)
+	}
+
+	path := filepath.Join(root, "project", "late.jsonl")
+	writeUsageFixture(t, path, fmt.Sprintf(`{"type":"session","id":"pi-late","cwd":%q,"version":3}`+"\n", workspace))
+	mustNoError(t, collector.ReconcileSources(context.Background(), -1))
```
illegalcall, line 2108: **[P1] Keep retrying Pi discovery after the startup race**

This explicit second `ReconcileSources` call is what makes the test pass, but production does not schedule an equivalent retry. `MarkSpawned` triggers one `ReactivateSession` / `backfillPiPaths` scan; if Pi has not created its JSONL file yet, no binding is recorded. The watcher only watches already-registered files, and `HasPendingUsageDiscovery` only sees existing bindings, so nothing triggers another scan when the file appears. Token usage then remains missing until an unrelated refresh or restart.

A likely fix is to persist a pending Pi-discovery claim keyed by AO session/workspace, or use a bounded root watch/poll, until `reconcilePiPath` establishes the binding, then clear the pending state. Please add a coordinator-level regression test that starts the production pipeline, calls `MarkSpawned`, creates the Pi artifact afterward, and observes the binding without directly calling `ReconcileSources`.

nikhilachale: Fixed in 829a1e6ff. `HasPendingUsageDiscovery` now treats a live, non-exited Pi session with a committed workspace and no Pi binding as durable pending discovery. That keeps the coordinator retry scheduled after the `MarkSpawned` scan misses; once `reconcilePiPath` creates the binding, the predicate clears automatically.

Added a production-pipeline regression that starts `Pipeline`, calls `MarkSpawned`, waits for the post-spawn pending check, creates the Pi artifact afterward, observes the binding without directly calling `ReconcileSources`, and verifies pending discovery clears. Verified with 10 repeated runs, the race detector, affected-package tests, and `go test -p 1 ./...`.

illegalcall (approved):

PR #4186, opened 2026-08-20
fix(chat): fail fast on immutable history replay

### LINKED ISSUE #4123: chat: settle wait polls an immutable cached ACP replay for the full 45s before failing the interface switch
## Problem

`Controller.importNativeHistory` (`backend/internal/service/chat/controller.go`) polls `ReadHistory` every 100ms for up to 45s (`nativeHistorySettleLimit`) waiting for the replay to settle / reach the checkpoint.

For ACP drivers (claude, and other `chatdriver/acp` bindings), `ReadHistory` returns a **one-shot cache** captured during `session/load` at Resume time (`backend/internal/adapters/chatdriver/acp/history.go` — `historyLoaded`/`historyEvents`/`historyErr` are set once and never refreshed). If the first read is unsettled or the checkpoint is unreachable, no amount of polling can change the answer — the user just waits 45 seconds for a guaranteed failure, during which the source TUI is already stopped and must be rolled back.

(The Codex app-server driver re-reads live state, so polling is meaningful there — this is ACP-specific.)

**Source:** observed during triage of [#4120](https://github.com/Untrivial-ai/agent-orchestrator/issues/4120) | **Analyzed against:** `4a23d98ca`
**Confidence:** High (code-traced)

## Suggested fix

Either:
- have the ACP `ReadHistory` (or a capability flag on `ports.ChatHistoryReader`) declare its result immutable so the controller fails fast instead of polling, or
- make the poll re-trigger `session/load` so genuinely-converging histories (transcript flush lag) can actually settle for ACP too.

## Impact

- P2 once #4120 lands: only genuinely unsettled/mismatched replays hit this, but they cost 45s + a TUI stop/rollback cycle instead of an immediate, actionable error.

illegalcall:
## What

- add an optional `ChatHistoryRefresher` port for history sources that can make a new authoritative provider observation
- read every native history source once, then retry only through that explicit refresh operation
- keep ACP `session/load` replay snapshot-only so an incomplete replay fails immediately instead of polling the same cache for 45 seconds
- make Codex refreshable through its real `thread/read`, preserving bounded convergence where it can actually make progress

## Why

Fixes #4123.

ACP captures replay once during `Resume`; every later `ReadHistory` returns the same frozen events. The generic settle loop nevertheless reread that cache every 100ms for up to 45 seconds when it was unsettled or missed AO's checkpoint. That delayed a guaranteed handoff failure and rollback without observing any new provider fact.

## How

This follows ACP v1's actual boundary: [`session/load` is a finite replay whose response follows the replayed entries](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/docs/protocol/v1/session-setup.mdx), not a standardized repeatable history read.

- `ChatHistoryReader` represents one authoritative history observation.
- `ChatHistoryRefresher` promises that `RefreshHistory` performs provider I/O or waits for a real provider signal.
- `Controller.importNativeHistory` keeps its bounded timeout around the initial read and refresh loop, but immediately returns `ErrChatHistoryUnsettled` when an incomplete source cannot refresh.
- Codex implements refresh by issuing another `thread/read(includeTurns: true)`.
- ACP deliberately does not implement refresh; the PR never reissues `session/load`, adds a grace sleep, or infers provider state from a load response.

Intentional omissions: this focused fix does not change trailing-user-only turn classification, introduce load-versus-resume intent, or adopt draft ACP v2 lifecycle/replay types. Those require separate state/evidence contracts and should not be coupled to #4123.

## Testing

- `cd backend && go test ./internal/service/chat ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/codexappserver -count=1`
- `cd backend && go test -race ./internal/ports ./internal/service/chat ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/codexappserver -count=1`
- `GOLANGCI_LINT_CACHE=/tmp/ao-lint-cache-4123-fresh npm run lint` — full backend tests and golangci-lint, 0 issues

Regression coverage verifies:

- an immutable unsettled replay is read exactly once and fails immediately
- an immutable settled replay behind AO's checkpoint is read exactly once and fails immediately
- a refreshable unsettled source converges through one initial read plus one refresh
- a refreshable settled-but-stale source refreshes until it reaches AO's checkpoint
- cancellation before refresh preserves the typed unsettled error and context cause
- Codex refresh performs a second wire `thread/read`
- ACP conversations do not advertise refreshability

## Checklist

- [x] Branched from `main`
- [x] Tests added for the changed boundary and user-visible failure behavior
- [x] Relevant build, race, test, and lint checks pass

### DIFF
--- a/backend/internal/adapters/chatdriver/acp/driver_test.go
+++ b/backend/internal/adapters/chatdriver/acp/driver_test.go
@@ -623,6 +623,9 @@ func TestACPDriverLoadsSettledHistoryWhenTheAgentCanReplayIt(t *testing.T) {
 		t.Fatalf("Resume: %v", err)
 	}
 	defer conv.Close()
+	if _, ok := conv.(ports.ChatHistoryRefresher); ok {
+		t.Fatal("ACP session/load replay is a frozen snapshot, not refreshable history")
+	}
 
 	agent.mu.Lock()
 	loadCalls, resumeCalls := agent.loadCalls, agent.resumeCalls
--- a/backend/internal/adapters/chatdriver/codexappserver/history.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/history.go
@@ -45,10 +45,11 @@ import (
 // feature-detects each interface, and a missing method would read as "the provider
 // cannot do this" with nothing anywhere to notice.
 var (
-	_ ports.ChatRollbacker    = (*conversation)(nil)
-	_ ports.ChatForker        = (*conversation)(nil)
-	_ ports.ChatRenamer       = (*conversation)(nil)
-	_ ports.ChatHistoryReader = (*conversation)(nil)
+	_ ports.ChatRollbacker       = (*conversation)(nil)
+	_ ports.ChatForker           = (*conversation)(nil)
+	_ ports.ChatRenamer          = (*conversation)(nil)
+	_ ports.ChatHistoryReader    = (*conversation)(nil)
+	_ ports.ChatHistoryRefresher = (*conversation)(nil)
 )
 
 // providerRefusal marks a request the provider rejected on its own terms, as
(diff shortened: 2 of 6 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):
Looks good. This fixes the unnecessary 45-second wait for ACP and keeps retries for refreshable providers like Codex. Tests and CI pass. Approved.

PR #4188, opened 2026-08-20
feat: automate agent catalog refresh

nikhilachale:
## Summary

Remove manual agent and model catalog refresh controls and move catalog freshness behind the existing cache-first daemon boundary.

The resulting read path is:

`UI → TanStack Query cache → daemon API → SQLite cache → local CLI discovery only when refresh/revalidation is due`

### Persist the agent installation inventory

- Add a singleton `agent_inventory_cache` SQLite table containing the last successfully observed installed/authorized snapshot and its observation timestamp.
- Hydrate the daemon's in-memory inventory from SQLite on the first read after startup, without running agent CLI probes.
- Reconcile persisted IDs against the adapters supported by the current daemon build so removed adapters are not shown and current labels win over stale cached labels.
- Persist the next snapshot after a successful bounded inventory probe.
- Keep installation and authorization state advisory: session spawn remains the authoritative runtime validation boundary.

### Trigger refresh automatically

- Keep the existing daemon startup refresh so installations are reconciled in the background whenever the daemon boots.
- Refresh the inventory through `refreshAgentsIfStale()` when Project Settings opens, then update the TanStack Query cache with the result.
- Preserve the existing create-project stale refresh behavior already present on `main`.
- Continue rate-limiting and serializing inventory probes so multiple UI reads cannot stampede local agent CLIs.

### Make model revalidation actually refresh

- Continue serving cached model catalogs immediately.
- When the trust window expires, let the frontend revalidate in the background.
- Fix `RevalidateModels` so it performs discovery even when the binary/config fingerprint has not changed. This covers provider-side catalog drift that a local fingerprint cannot detect.
- Preserve stale/fallback catalog behavior and warnings when rediscovery fails.

### Remove manual catalog controls

- Remove the Project Settings “Refresh agents” row.
- Remove worker and orchestrator model refresh buttons.
- Remove the optional model-menu rediscovery action and now-unused product UI refresh props.
- Keep explicit daemon refresh endpoints/helpers for background refresh and older-client compatibility; only the manual UI controls are removed.
- Leave unrelated reload/update controls (browser reload, application update checks, retry actions) unchanged.

## Why

Catalog refresh is application bookkeeping, not a decision users should need to make. Persisting the last known inventory makes the UI useful immediately after restart, TanStack Query avoids repeated renderer fetches, and background daemon probing repairs stale installation data without blocking normal reads.

This also avoids repeatedly invoking agent CLIs during routine navigation while still converging after installs, removals, daemon restarts, or aged model catalogs.

## Failure behavior

- A missing, corrupt, or temporarily unreadable inventory snapshot falls back to the daemon's supported catalog and can be repaired by the background startup refresh.
- A failed live agent probe does not erase supported adapters; the next automatic trigger can retry.
- A failed inventory-cache write does not publish or rate-limit the probed snapshot, so the next refresh retries persistence instead of reporting a false success.
- A failed model rediscovery retains the fuller cached catalog when available and surfaces its warning.
- Spawn remains authoritative if cached installation/auth data has become stale.

## Verification

- `GOCACHE=/private/tmp/go-build go test ./internal/service/agent ./internal/storage/sqlite`
- `GOCACHE=/private/tmp/go-build go test ./internal/daemon`
- `npm run sqlc`
- Focused renderer tests: 44 passed
- `npm --prefix packages/product-ui run typecheck`
- Product UI tests: 61 passed

## Notes

- No API contract shape changed, so OpenAPI regeneration is not required.
- Migration `0104_agent_inventory_cache.sql` uses the next available migration number after merging the latest `main`.


backend/internal/service/agent/service.go
```diff
@@ -78,28 +78,43 @@ type Inventory struct {
...
-	Discoverer ports.AgentModelDiscoverer
-	Projects   ProjectLookup
+	Cache          ports.AgentModelCatalogCache
+	InventoryCache InventoryCache
+	Discoverer     ports.AgentModelDiscoverer
+	Projects       ProjectLookup
+}
+
+// InventoryCache persists the last successful advisory installation/auth probe.
+type InventoryCache interface {
```
illegalcall: Could we move this storage boundary into internal/ports as ports.AgentInventoryCache, alongside ports.AgentModelCatalogCache? docs/architecture.md says external systems are accessed through port interfaces defined in internal/ports. Keeping this new SQLite-backed contract inside service/agent makes this boundary inconsistent with the documented architecture and harder to discover. Recommended fix: define AgentInventoryCache in backend/internal/ports/agent.go, change Deps.InventoryCache to ports.AgentInventoryCache, and remove this local interface. The SQLite Store will continue to satisfy the interface implicitly, so the behavior does not need to change.

nikhilachale: Addressed in 901aa9a56. I moved the storage boundary to ports.AgentInventoryCache alongside AgentModelCatalogCache, updated agent.Service and Deps to use the ports contract, and removed the service-local interface. The SQLite store continues to satisfy it implicitly. Focused ports, agent-service, storage, and daemon tests pass. Thanks for the architecture catch.

backend/internal/storage/sqlite/store/agent_inventory_store.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/storage/sqlite/gen"
+)
+
+// GetAgentInventoryCache loads the last successfully observed local agent inventory.
```
illegalcall, line 13: Please add a direct store-level test for this new persistence boundary. The service tests use a fake cache, so they do not verify the sqlc queries or Store wiring. The test should open a migrated SQLite test store and verify: (1) GetAgentInventoryCache before any insert returns ok=false with no error; (2) UpsertAgentInventoryCache followed by GetAgentInventoryCache round-trips the inventory JSON and UTC observation timestamp; and (3) a second upsert replaces the singleton row. This gives the new real database boundary the coverage required by AGENTS.md.

illegalcall: Addressed in 0a53c0094. I added a migrated SQLite store-level test covering all three cases: missing cache returns ok=false without error, upsert/get round-trips the exact inventory JSON and normalizes the observation time to UTC, and a second upsert replaces the singleton row. The new store tests pass, including under the race detector.

illegalcall:
## Local verification evidence

Verified final PR head `b3b47055f` in the real Electron desktop app using a fresh isolated AO data profile.

- Fresh database applied `0102_canonical_usage.sql` and `0103_agent_inventory_cache.sql`.
- Startup refresh populated the singleton `agent_inventory_cache` row.
- `/readyz` returned ready and `/api/v1/agents` returned the discovered inventory.
- Project Settings → Agents rendered with no manual refresh controls (`REFRESH_BUTTONS []`).
- Store tests cover the missing row, JSON/UTC round-trip, and second-upsert replacement.
- All 10 GitHub CI checks pass, including Go build/test, lint, API drift, frontend tests, renderer smoke, and native/container CLI E2E.

Screenshot from the final head:

pr-4188-local-verification

illegalcall (approved):

PR #4189, opened 2026-08-20
fix(updater): check disk space before update and surface failed installs (#3528)

### LINKED ISSUE #3528: Updater fails silently when the disk is (nearly) full: ENOSPC-truncated Squirrel extraction surfaces as bogus 'file doesn't exist' install errors the user never sees
## Bug

With the data volume at 100% (observed live: 227MB free of 460Gi), the whole update chain degrades into misleading, mostly **invisible** failures:

1. **Install:** Squirrel.Mac's ShipIt extracts the update zip and stages copies under `/var/folders` and `~/Library/Caches/<bundle-id>.ShipIt` (~400MB+ needed). With no space, extraction truncates and the install fails with a bogus-looking error — observed live 2026-08-03 15:05:43 in `ShipIt_stderr.log`:
   `SQRLInstallerErrorDomain Code=-1 Failed to copy bundle ... The file "af.lproj" doesn't exist` (`af.lproj` = alphabetically FIRST locale dir in `Electron Framework.framework` — the tell of a truncated extraction). Because this happens **after the app has quit**, the user sees nothing at all: the app just doesn't update, and on relaunch the sidebar shows 'Restart to update' again. Retries then hit `App Still Running` aborts.
2. **Check:** Chromium's net stack needs disk for cache/temp; a full disk plausibly produces the persistent `net::ERR_FAILED` on update checks documented in #3526 (same machine, same day).

**Proof the failure was environmental, not a bad artifact:** after freeing ~1GB, the exact same staged zip (`~/Library/Caches/agent-orchestrator-updater/pending/...202608032052.zip`) extracts completely via `ditto -x -k` (all 55 `*.lproj` present, `af.lproj` intact) and the extracted bundle passes `codesign --verify --deep --strict`. So this is NOT a #3034 regression (that was a genuinely malformed zip); it is the same install surface failing for a different, environmental reason.

## Hardening proposals

1. **Preflight free-space check:** before downloading (~110MB) and before `quitAndInstall` (~3× extracted app size for extract+stage+backup, ballpark 1.5GB), check `statfs` on the relevant volumes; if below threshold, show an actionable message ('Free up N GB to install the update') instead of failing invisibly.
2. **Surface post-quit install failures:** ShipIt failures happen after the app exits, so nothing in-app ever reports them. On launch, if the pending update's version is still newer than the running version AND a prior install attempt was made (e.g. persisted marker written before `quitAndInstall`), surface 'The last update didn't install' with the ShipIt error tail (readable from `~/Library/Caches/<bundle-id>.ShipIt/ShipIt_stderr.log`).
3. Consider telemetry for both paths (phase: install), so fleet-wide install failures are measurable — today the install phase is a telemetry blind spot (UpdatePhase is only check|download).

## Related
- #3526 (net::ERR_FAILED on checks — plausibly the same disk exhaustion, to be confirmed after cleanup)
- #3527 (translocation/read-only bundle: a different silent no-op in the same 'Restart to update' flow)
- #3034 (closed; genuinely malformed zip — different cause, same surface)

Rishet11:
## What

When the disk is nearly full, the updater now checks free space before downloading and before installing, and reports install failures that happened after the app quit, which previously were invisible to the user.

## Why

Fixes #3528. Squirrel.Mac extracts the update into temporary folders and stages a backup of the app, needing about 1.5 GB. With no room, the extraction truncates and the install fails after the app has quit, so nothing in the app ever surfaces it. The user just sees the "Restart to update" row again with no explanation, and retries keep failing.

## How

- Before an automatic or manual download, the updater checks free space on the cache volume against a 250 MB reserve. Below that, the download is skipped with an actionable "free up space" message instead of truncating mid-way.
- Before quitAndInstall, free space is checked on the app volume and the cache volume against a 1.5 GB reserve. Below that, a dialog says how much space to free instead of handing the update to a doomed post-quit install.
- quitAndInstall now writes a small install-attempt marker first. On the next launch, if the attempted version never became the running version, the app shows "The last update didn't install" together with the tail of the Squirrel ShipIt stderr log (this is where the misleading "file ... doesn't exist" error appears).
- The telemetry phase set now includes "install", so install failures are measurable instead of being a blind spot (it was previously only check|download). See shared/update-telemetry.ts.
- Every check fails open: if statfs reports nothing, the update proceeds exactly as it did before.

## Testing

- 80 vitest tests pass across the updater suites: auto-updater, update-disk-space, update-settings, update-telemetry.
- New tests cover the blocked-install dialog and telemetry, the fail-open path, install-attempt marker persistence, and the launch-time failure report with the ShipIt log tail.
- Typecheck passes for all touched files.
- Live behavior on an actual full disk was not exercised; that needs a packaged build.

## Checklist

### DIFF
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -1,8 +1,24 @@
 // @vitest-environment node
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { chmodSync, mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import {
+  chmodSync,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import os from "node:os";
 import nodePath from "node:path";
+
+// node:fs's ESM namespace is not configurable, so statfsSync is mocked at the
+// module level instead of via spyOn.
+const statfsMock = vi.hoisted(() => vi.fn());
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  return { ...actual, statfsSync: statfsMock as unknown as typeof actual.statfsSync };
+});
 
 type UpdateSettings = {
   enabled: boolean;
@@ -21,6 +37,9 @@ type ImportOptions = {
     settings: UpdateSettings,
   ) => Promise<{ settings: UpdateSettings; cleared: boolean }>;
   isPackaged?: boolean;
+  // Backs app.getPath in the electron mock; unset makes getPath throw, which
+  // the updater's marker/preflight paths must tolerate (fail open).
+  getPath?: (name: string) => string;
 };
 
 type AutoUpdaterMock = {
@@ -93,6 +112,9 @@ async function importAutoUpdater(
     app: {
       isPackaged: options.isPackaged ?? true,
       getVersion: () => "1.0.0",
+      getPath: options.getPath ?? (() => {
+        throw new Error("getPath not available in test");
+      }),
     },
     BrowserWindow,
     dialog,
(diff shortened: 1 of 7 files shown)


i-trytoohard (commented):
## Review: fix(updater): disk-space preflight + post-quit failure surfacing (#3528)

**Verdict: strong PR — no blocking issues found.** Implementing both hardening proposals from the issue plus the telemetry phase, with real tests. Detailed comments below; posting as COMMENT since approval needs a human reviewer.

### ✅ What's right

- **Fail-open everywhere** — statfs throws → update proceeds as before; a volume that can't be statfs'd is skipped, not treated as full. Correct default for a preflight.
- **`bavail` × `bsize`** (not `bfree`) — the right statfs fields for what a user process can actually allocate, and `Number()` conversion handles Node's BigInt returns.
- **Install preflight checks both volumes** (app dir + cache dir) and takes the min — matches Squirrel's real footprint (extract + stage + backup span both).
- **Marker semantics are exactly right**: written *before* `quitAndInstall`, compared against `app.getVersion()` on next launch, cleared in every path (including the success path) so a failure reports exactly once. The "stays silent when attempted version is now running" test is the one people usually forget.
- **`autoDownload = blocker === undefined`** — the check still runs so statuses stay live; only the download waits for space. Good separation.
- **Tests exercise real temp dirs** (marker file, ShipIt log tail) rather than stubbing the fs, and the module-level `vi.mock("node:fs")` technique is the correct workaround for the non-configurable ESM namespace.
- Reuses the existing `getMacInstallBlocker` composition + dialog-not-broadcast policy instead of inventing a parallel mechanism.

### 💡 Nits (non-blocking)

1. **`update-disk-space.ts` header comment is garbled** mid-sentence: *"…fails with a bogus / whether the missing file is whichever locale directory…"* — an editing artifact; the quoted error string got lost. Worth fixing for the archaeologists who'll read this in a year.
2. **Message rounding overstates the download need**: `gbRoundedUp(250MB)` → "needs about 1 GB of free disk space" (the `Math.max(1, …)` floor), and the actual-free side prints "0.09 GB" via `.toFixed(2)`. Suggest MB below 1 GB ("needs about 250 MB; 90 MB free") — clearer and less scary.
3. **Auto-install-on-quit blocked by disk stays silent post-quit**: `applyInstallOnQuitPolicy` disables the quit-time install (correct), and the status row shows `installBlockedReason` — but no marker is written on that path, so a user who never opens Settings still gets no push notification after a quit that didn't update. This matches the existing #3527 behavior for location blockers, so consistent — but worth a follow-up issue if #3528's "invisible failure" bar applies here too.
4. **Windows detail**: `shipItLogTail` is darwin-only (correct — NSIS has no ShipIt log), so Windows post-quit failures get the generic "often caused by not enough disk space" text. Fine for now; noting so nobody's surprised.

### ⚠️ One action needed

**CI hasn't run** — no checks reported on the fork branch. Maintainers need to approve workflow runs for this first-time contributor PR, otherwise "80 tests pass" remains an author claim (the suites touched do look plausible from the diff).

*Context: this PR implements the sketch from [this comment](https://github.com/Untrivial-ai/agent-orchestrator/issues/3528#issuecomment-5357690751); the 2026-08-20 field incident (verified-intact zip, <1GB free, ditto `._app.asar__`/`node-which` errors, self-healed after freeing space) is the motivating case.*

i-trytoohard:
Context link: the implementation sketch this PR follows is [here](https://github.com/Untrivial-ai/agent-orchestrator/issues/3528#issuecomment-5357690751), and the field incident that motivated it (verified-intact zip + <1GB free + bogus ditto missing-file errors, self-healed after freeing space) is documented [here](https://github.com/Untrivial-ai/agent-orchestrator/issues/3528#issuecomment-5357583598). Reporter of the incident is this PR's author. cc @harshitsinghbhandari for review + CI workflow approval (first-time contributor — checks haven't run).

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@Rishet11 Marking as stale: no activity for 24 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Rishet11 Marking as stale: no activity for 24 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Rishet11 Closing as stale: no activity from you for 34 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

PR #4190, opened 2026-08-20
Align chat approval card UI

nikhilachale:
## Summary

- Move pending command approvals into the existing chat composer instead of rendering a separate timeline card.
- Match AO’s compact typography, spacing, colors, and split Allow/Always Allow controls.
- Keep resolved approvals as compact timeline records without status icons or request IDs.
- Support Enter for **Allow once** and Escape for **Deny**.

## Validation

- `ChatWorkspace.test.tsx` — 52 passed
- `ChatComposer.test.tsx` — 45 passed
- Chat steering, rollback, and provider-state tests — 42 passed
- `tsc --noEmit`
- `git diff --check`

## Notes

- Frontend-only presentation change; provider decision IDs and daemon protocol remain unchanged.


nikhilachale:
Screenshot 2026-08-20 at 22 56 48

illegalcall:
I reproduced the approval flows end to end on this PR's exact head (`0c189b79d28bd7c8a47de3b03d7acba89a9756c0`) in the real Electron app, using an isolated daemon/data directory and live Codex + Claude Code ACP sessions. These are merge-blocking correctness/consent issues.

## 1. [P0] Live Codex approvals disappear, leaving the turn stuck

**Reproduction**

1. Start a Codex chat session.
2. Select **Ask outside worktree**.
3. Ask Codex to run a harmless command outside the worktree: `touch /Users/dhruvsharma/.ao/dev/pr4190-codex-approval-marker`.

**Observed**

- Codex emitted a real command approval request.
- The conversation API stored a pending approval (`requestId: "0"`) but with no `turnId`.
- The marker was not created, confirming the provider was actually blocked awaiting consent.
- The native UI showed **Input Needed**, **Ran command**, and **Working**, but rendered zero approval groups and zero Allow/Deny buttons. The only recovery was **Stop turn**.

**Cause**

- The Codex adapter emits the approval without `ProviderTurnID`: [`conversation.go#L653-L662`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/backend/internal/adapters/chatdriver/codexappserver/conversation.go#L653-L662).
- The composer only selects a pending approval when `item.turnId === turn.id`: [`ChatWorkspace.tsx#L541-L555`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatWorkspace.tsx#L541-L555).
- The timeline now hides every pending approval, including unbound ones: [`ChatWorkspace.tsx#L1929-L1933`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatWorkspace.tsx#L1929-L1933).

In simple terms: Codex asks for permission, but the request is not connected to its turn. The new composer ignores it, and the old fallback card is hidden, so the user cannot answer.

## 2. [P1] Enter approves even when Deny is focused—and even while Chat is hidden

I reproduced both cases with live Claude Code ACP approvals for harmless file writes.

**Focused Deny case**

1. Wait for the visible approval card.
2. Focus the **Deny** button.
3. Press Enter.

AO sent `decision: "allow"`; the provider completed the write and created the marker file.

**Hidden Chat case**

1. Wait for another pending approval.
2. Open a terminal tab so the Chat panel is hidden.
3. Focus the unrelated **Orchestrator** button and press Enter.

The Orchestrator button did not activate. The hidden approval received `decision: "allow"`, and the second marker file was created.

The capture-phase `window` listener excludes inputs and menus, but not buttons/links or inactive Chat panels: [`ChatTimelineItems.tsx#L1475-L1500`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatTimelineItems.tsx#L1475-L1500). Chat is hidden/inert rather than unmounted: [`ChatWorkspace.tsx#L640-L647`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatWorkspace.tsx#L640-L647).

This shortcut must be scoped to the active approval surface and must not override focused interactive controls.

## 3. [P1] A real ACP file edit is presented as command execution

**Reproduction**

1. Start a Claude Code ACP chat in Manual permission mode.
2. Ask it to create `pr4190-acp-edit-marker.txt` using its file editing tool.

**Observed**

The API correctly recorded a pending `file_change` with `providerToolName: "Write"` and `toolKind: "edit"`. The composer nevertheless asked:

> Do you want to run this command?

and displayed `Write pr4190-acp-edit-marker.txt` as command text.

The ACP approval event does not carry normalized subject/detail metadata into the approval payload: [`client.go#L79-L92`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/backend/internal/adapters/chatdriver/acp/client.go#L79-L92). The UI only recognizes Codex's `item/fileChange/requestApproval` method and otherwise defaults to command copy: [`ChatTimelineItems.tsx#L1601-L1605`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatTimelineItems.tsx#L1601-L1605).

## 4. [P1] Decision semantics are not generally ACP-safe

This fourth point is a protocol/data-path finding, not a misclassification observed with the installed Claude agent: Claude supplied English IDs/labels (`reject`, `allow`, `allow_always`), so those particular buttons rendered correctly.

ACP already supplies authoritative semantic kinds such as `allow_once`, `allow_always`, `reject_once`, and `reject_always`. AO preserves the raw provider option internally, but the durable/API projection keeps only `id` and `label`: [`controller.go#L2560-L2569`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/backend/internal/service/chat/controller.go#L2560-L2569). React then guesses meaning from English regexes over those strings: [`ChatTimelineItems.tsx#L1607-L1620`](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c189b79d28bd7c8a47de3b03d7acba89a9756c0/frontend/src/renderer/components/chat/ChatTimelineItems.tsx#L1607-L1620).

An ACP provider is allowed to use opaque or localized IDs/labels, so this cannot guarantee the requested Enter=allow-once / Escape=deny behavior across ACP implementations. The daemon should expose a normalized semantic kind while retaining the provider decision ID used for the reply.

## Verification

- `npm run typecheck` — passed.
- Targeted frontend tests — 84 passed.
- `go test ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/codexappserver` — passed.

The current tests use a valid turn ID and English decision fixtures, which is why they do not catch these live failures. Both ACP marker files were removed after reproduction; the Codex marker was never created.

## Screenshots

1. **Codex:** the app reports Input Needed and Working, but there is no approval composer or Allow/Deny action.
2. **ACP edit:** a live `Write`/`toolKind: edit` approval is presented as “run this command.”

Codex pending approval is invisible in the composer

ACP file-edit approval is presented as command execution

nikhilachale:
@illegalcall 


> Thanks for the detailed end-to-end reproduction. I’ve fixed all four blockers:
>
> - Codex approvals now carry their provider turn ID, with a frontend fallback for existing unbound approvals.
> - Enter/Escape handling is scoped to the visible approval surface and no longer overrides focused controls or operates while Chat is hidden.
> - ACP edit approvals now use normalized file-change metadata and appropriate wording.
> - ACP semantic decision kinds are preserved through the daemon/API, so the UI no longer relies on English IDs or labels.
>
> I also added regression coverage for unbound approvals, keyboard scoping, ACP edit wording, and opaque/localized decision options.
>
> Validation: 131 frontend tests passed, affected Go adapter/service tests passed, TypeScript passed, and `git diff --check` passed.
>
> I’ll push the update shortly and would appreciate a re-review afterward.

illegalcall:
Did you push the commit?

illegalcall (approved):

illegalcall:
Re-review of current head `94f0b9751882d9bb4811e417bb46259cf057fa41` after the merge from `main`:

The four previously reported flows still pass in the real Electron app with live Codex and Claude Code ACP sessions. Codex Deny + Enter denied and left the marker absent; Claude ACP showed file-change wording, exposed typed decision kinds, and created its marker only after Allow once. The approval-focused Go tests, TypeScript check, and 110 frontend tests also passed. All GitHub checks are green.

I found one residual consent-safety edge case that should be fixed before merge.

## [P1] An unknown, untyped provider option can still become the Enter-to-approve action

In simple terms: if a provider adds an option AO does not understand, AO should show that option as an explicit provider-owned button. It must not guess that the option means “Allow once” and bind Enter to it.

### Exact reproduction

I replayed a pending approval through the real Electron renderer with this provider decision list (a protocol-edge fixture; installed Codex/Claude do not currently emit this exact option):

```json
[
  {"id":"acceptWithDifferentSemantics","label":"acceptWithDifferentSemantics"},
  {"id":"cancel","label":"Cancel","kind":"reject_once"}
]
```

Observed:

1. AO hid the provider-owned `acceptWithDifferentSemantics` button and rendered it as **Allow once**.
2. I focused the approval surface and pressed Enter.
3. The intercepted resolver request was:

```json
{"decisionId":"acceptWithDifferentSemantics"}
```

I also added a temporary regression test asserting that Enter must do nothing for an unknown option without `kind`. It failed with:

```text
expected "vi.fn()" to not be called at all, but actually been called 1 times
1st call: ["unknown-request", "acceptWithDifferentSemantics"]
```

The temporary test was removed afterward; the review worktree is clean.

### Cause

[`approvalDecisionKind`](https://github.com/Untrivial-ai/agent-orchestrator/blob/94f0b9751882d9bb4811e417bb46259cf057fa41/frontend/src/renderer/components/chat/ChatTimelineItems.tsx#L1636-L1649) falls back to English regex matching whenever `kind` is absent, so any ID/label containing `allow`, `approve`, or `accept` becomes `allow_once`.

That is reachable for future Codex options: [`codexDecisionKind`](https://github.com/Untrivial-ai/agent-orchestrator/blob/94f0b9751882d9bb4811e417bb46259cf057fa41/backend/internal/adapters/chatdriver/codexappserver/conversation.go#L906-L918) deliberately returns an empty kind for unknown IDs, while [`decisionLabel`](https://github.com/Untrivial-ai/agent-orchestrator/blob/94f0b9751882d9bb4811e417bb46259cf057fa41/backend/internal/adapters/chatdriver/codexappserver/conversation.go#L922-L937) uses the unknown ID as its label.

### Expected fix

Only an explicit semantic `kind: "allow_once"` should receive the Enter shortcut, and only `kind: "reject_once"` should receive Escape. An option without a recognized kind should remain an ordinary explicit button using its provider label. If old stored rows need compatibility, that compatibility needs explicit provenance rather than applying the regex to every live untyped option.

illegalcall:
Update: I fixed the untyped-option shortcut issue in `d5490972e8682b8687b63960767bfab7b80c8fd7`.

- AO now assigns Enter/Escape only from the provider's explicit semantic `kind`.
- An unknown option without a kind stays an ordinary provider-labeled button and requires an explicit click.
- Added regression coverage for `acceptWithDifferentSemantics`: Enter does nothing, while clicking that exact provider button still sends it.

Validation at the new head:

- Approval-focused frontend tests: **111 passed**
- TypeScript: passed
- `git diff --check`: passed
- GitHub CI: **all 10 checks passed**, including Go tests, Go lint, frontend tests, renderer smoke, API drift, and CLI E2E on macOS/Linux/Windows.

The merge-blocking concern in my preceding comment is resolved.

illegalcall:
Final follow-up at `fe635d00ad69090d3430178ced1ad45b6f7d7de6`: I also fixed the non-blocking opaque-decision history gap from the re-review.

In simple terms: AO was already sending the correct ACP decision, but after the fact an opaque denial such as `option-b` could be displayed only as **Resolved**. Resolution now merges into the original approval detail instead of replacing it, so the provider-neutral decision kind, file/command subject, and tool metadata survive for the audit row. The renderer uses that preserved kind first and keeps its old ID-text fallback only for legacy rows.

Exact-head real Electron verification with live Claude Code ACP:

1. Manual-mode Claude requested the Write tool for `pr4190-final-acp-marker.txt`.
2. The file was absent while the file-change approval was pending.
3. The native UI showed **Do you want to allow these file changes?**.
4. I focused the approval surface and pressed Escape.
5. The provider received `reject`; the file remained absent.
6. The resolved API row still contained all 3 typed decisions plus `subjectKind: file_change` and `toolKind: edit`.
7. The native Electron history row displayed **Cancelled**.

Regression/validation:

- Backend chat service + SQLite store tests: passed
- Approval-focused frontend tests: **112 passed**
- TypeScript: passed
- Pinned `golangci-lint` v2.12.2: **0 issues**
- `git diff --check`: passed

The earlier live Codex denial, live ACP allow, and untyped-option Enter safety checks also remain green. No review concern remains open from my pass.

PR #4193, opened 2026-08-20
feat: add native Kimi ACP chat driver

nikhilachale:
Add native Kimi ACP chat driver with full ACP lifecycle support.

## Implementation
- **backend/internal/adapters/chatdriver/kimiacp/driver.go** - Native Kimi ACP driver
- **backend/internal/adapters/chatdriver/kimiacp/driver_test.go** - Kimi-specific tests
- **backend/internal/adapters/chatdriver/registry/** - Registry integration

## Features
- Session creation and lifecycle management
- Streaming message support
- Tool event handling
- Approval request round-trips
- Image and plan capability mapping
- Session load/resume and replay

## Testing
- Kimi wrapper/registry tests pass
- Shared ACP lifecycle tests pass (session, streaming, tools, approvals, resume/replay)
- Full backend `go test ./...` passes
- `npm run lint` passes

## Notes
- Chat-only implementation (no TUI handoff)
- Reuses existing Kimi adapter for binary discovery and auth
- Launches via `kimi acp`

### DIFF
--- a/backend/internal/adapters/agent/kimi/hooks.go
+++ b/backend/internal/adapters/agent/kimi/hooks.go
@@ -16,7 +16,7 @@ import (
 var kimiAPIKeyLineRE = regexp.MustCompile(`(?m)^\s*api_key\s*=\s*("([^"]*)"|'([^']*)'|([^\s#]+))`)
 
 const (
-	kimiInstructionsDirName  = ".kimi-code"
+	kimiInstructionsDirName  = ".kimi"
 	kimiInstructionsFileName = "AGENTS.md"
 	kimiInstructionsSentinel = "<!-- managed by agent-orchestrator: kimi system prompt -->"
 	kimiInstructionsEnd      = "<!-- /managed by agent-orchestrator: kimi system prompt -->"
--- a/backend/internal/adapters/agent/kimi/hooks.go
+++ b/backend/internal/adapters/agent/kimi/hooks.go
@@ -48,25 +48,39 @@ func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfi
 	if err != nil {
 		return fmt.Errorf("kimi.GetAgentHooks: %w", err)
 	}
-	if systemPrompt == "" {
+	if err := PrepareACPInstructions(cfg.WorkspacePath, systemPrompt); err != nil {
+		return fmt.Errorf("kimi.GetAgentHooks: %w", err)
+	}
+	return nil
+}
+
+// PrepareACPInstructions installs AO's standing instructions where Kimi Code
+// discovers project-level AGENTS.md files. Chat sessions call this immediately
+// before launching `kimi acp`; unlike TUI sessions, they do not run
+// GetAgentHooks during workspace preparation.
+func PrepareACPInstructions(workspacePath, systemPrompt string) error {
+	if strings.TrimSpace(systemPrompt) == "" {
 		return nil
 	}
-	instructionsPath := kimiInstructionsPath(cfg.WorkspacePath)
+	if strings.TrimSpace(workspacePath) == "" {
+		return errors.New("kimi: workspace path is required for ACP instructions")
+	}
+	instructionsPath := kimiInstructionsPath(workspacePath)
 	var existing []byte
-	existing, err = os.ReadFile(instructionsPath) //nolint:gosec // path built from caller-owned workspace dir
+	existing, err := os.ReadFile(instructionsPath) //nolint:gosec // path built from caller-owned workspace dir
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return fmt.Errorf("kimi.GetAgentHooks: read %s: %w", instructionsPath, err)
+		return fmt.Errorf("kimi: read ACP instructions %s: %w", instructionsPath, err)
 	}
 
 	if err := os.MkdirAll(filepath.Dir(instructionsPath), 0o750); err != nil {
-		return fmt.Errorf("kimi.GetAgentHooks: create instruction dir: %w", err)
+		return fmt.Errorf("kimi: create ACP instruction dir: %w", err)
 	}
 	body := mergeKimiInstructionFile(string(existing), systemPrompt)
 	if err := hookutil.AtomicWriteFile(instructionsPath, []byte(body), 0o600); err != nil {
-		return fmt.Errorf("kimi.GetAgentHooks: write %s: %w", instructionsPath, err)
+		return fmt.Errorf("kimi: write ACP instructions %s: %w", instructionsPath, err)
 	}
 	if err := hookutil.EnsureWorkspaceGitignore(filepath.Dir(instructionsPath), kimiInstructionsFileName); err != nil {
-		return fmt.Errorf("kimi.GetAgentHooks: gitignore: %w", err)
+		return fmt.Errorf("kimi: gitignore ACP instructions: %w", err)
 	}
 	return nil
 }
--- a/backend/internal/adapters/agent/kimi/hooks.go
+++ b/backend/internal/adapters/agent/kimi/hooks.go
@@ -48,7 +48,7 @@ func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfi
 	if err != nil {
 		return fmt.Errorf("kimi.GetAgentHooks: %w", err)
 	}
-	if err := PrepareACPInstructions(cfg.WorkspacePath, systemPrompt); err != nil {
+	if err := PrepareACPInstructions(ctx, cfg.WorkspacePath, systemPrompt); err != nil {
 		return fmt.Errorf("kimi.GetAgentHooks: %w", err)
 	}
 	return nil
@@ -58,7 +58,10 @@ func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfi
 // discovers project-level AGENTS.md files. Chat sessions call this immediately
 // before launching `kimi acp`; unlike TUI sessions, they do not run
 // GetAgentHooks during workspace preparation.
-func PrepareACPInstructions(workspacePath, systemPrompt string) error {
+func PrepareACPInstructions(ctx context.Context, workspacePath, systemPrompt string) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 	if strings.TrimSpace(systemPrompt) == "" {
 		return nil
 	}
@@ -71,18 +74,27 @@ func PrepareACPInstructions(workspacePath, systemPrompt string) error {
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
 		return fmt.Errorf("kimi: read ACP instructions %s: %w", instructionsPath, err)
 	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 
 	if err := os.MkdirAll(filepath.Dir(instructionsPath), 0o750); err != nil {
 		return fmt.Errorf("kimi: create ACP instruction dir: %w", err)
+	}
+	if err := ctx.Err(); err != nil {
+		return err
 	}
 	body := mergeKimiInstructionFile(string(existing), systemPrompt)
 	if err := hookutil.AtomicWriteFile(instructionsPath, []byte(body), 0o600); err != nil {
 		return fmt.Errorf("kimi: write ACP instructions %s: %w", instructionsPath, err)
 	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 	if err := hookutil.EnsureWorkspaceGitignore(filepath.Dir(instructionsPath), kimiInstructionsFileName); err != nil {
 		return fmt.Errorf("kimi: gitignore ACP instructions: %w", err)
 	}
-	return nil
+	return ctx.Err()
 }
 
 func kimiInstructionsPath(workspacePath string) string {
--- a/backend/internal/adapters/chatdriver/kimiacp/driver.go
+++ b/backend/internal/adapters/chatdriver/kimiacp/driver.go
@@ -1,0 +1,60 @@
+// Package kimiacp binds the user's own Kimi Code installation to AO's
+// reusable ACP Chat transport.
+package kimiacp
+
+import (
+	"log/slog"
+
+	acpdriver "github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/acp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/nativeacp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// New launches `kimi acp` from the exact binary resolved by the existing Kimi
+// agent plugin. Authentication, model discovery, sessions, and configuration
+// remain owned by that installation.
+func New(plugin nativeacp.Plugin, log *slog.Logger) ports.ChatDriver {
+	return nativeacp.New(plugin, nativeacp.Config{
+		Harness: domain.HarnessKimi,
+		Capabilities: ports.ChatCapabilities{
+			ports.ChatCapabilityHistory: true,
+			ports.ChatCapabilityPlans:   true,
+		},
+		Configure:      configure,
+		SessionOptions: sessionOptions,
+	}, log)
+}
+
+func configure(acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	return []string{"acp"}, nil, nil
+}
+
+// sessionOptions maps AO's durable turn settings onto Kimi's native ACP
+// config-option ids. Kimi advertises the complete model/thinking/mode catalog
+// during session setup; AO keeps that provider-owned catalog intact for direct
+// plan-mode and thinking selection in Chat.
+func sessionOptions(settings ports.ChatTurnSettings) []acpdriver.SessionOption {
+	var options []acpdriver.SessionOption
+	if settings.Model != "" {
+		options = append(options, acpdriver.SessionOption{ID: "model", Value: settings.Model})
+	}
+	if mode := permissionMode(settings.Approval); mode != "" {
+		options = append(options, acpdriver.SessionOption{ID: "mode", Value: mode})
+	}
+	return options
+}
+
+func permissionMode(permission ports.PermissionMode) string {
+	switch ports.NormalizePermissionMode(permission) {
+	case ports.PermissionModeAcceptEdits, ports.PermissionModeAuto:
+		return "auto"
+	case ports.PermissionModeBypassPermissions:
+		return "yolo"
+	default:
+		// Leave Kimi's configured/default mode unchanged. The provider-owned
+		// config option remains available when the user explicitly wants to
+		// switch back to default or enter plan mode.
+		return ""
+	}
+}
--- a/backend/internal/adapters/chatdriver/kimiacp/driver.go
+++ b/backend/internal/adapters/chatdriver/kimiacp/driver.go
@@ -3,6 +3,7 @@
 package kimiacp
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 
@@ -29,11 +30,11 @@ func New(plugin nativeacp.Plugin, log *slog.Logger) ports.ChatDriver {
 	}, log)
 }
 
-func configure(cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
-	if err := validateTurnSettings(ports.ChatTurnSettings{Approval: cfg.Permissions}); err != nil {
+func configure(ctx context.Context, cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	if err := validateTurnSettings(cfg.Permissions, ports.ChatTurnSettings{Approval: cfg.Permissions}); err != nil {
 		return nil, nil, err
 	}
-	if err := kimi.PrepareACPInstructions(cfg.WorkspacePath, cfg.SystemPrompt); err != nil {
+	if err := kimi.PrepareACPInstructions(ctx, cfg.WorkspacePath, cfg.SystemPrompt); err != nil {
 		return nil, nil, err
 	}
 	return []string{"acp"}, nil, nil
--- a/backend/internal/adapters/chatdriver/kimiacp/driver.go
+++ b/backend/internal/adapters/chatdriver/kimiacp/driver.go
@@ -5,6 +5,7 @@ package kimiacp
 import (
 	"log/slog"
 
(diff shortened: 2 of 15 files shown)


nikhilachale (commented):
Requesting changes. The new Kimi ACP binding otherwise follows the native driver pattern, but it drops AO standing instructions for every Kimi Chat session.

backend/internal/adapters/chatdriver/kimiacp/driver.go
```diff
@@ -0,0 +1,60 @@
...
+		Capabilities: ports.ChatCapabilities{
+			ports.ChatCapabilityHistory: true,
+			ports.ChatCapabilityPlans:   true,
+		},
+		Configure:      configure,
+		SessionOptions: sessionOptions,
+	}, log)
+}
+
+func configure(acpdriver.LaunchConfig) ([]string, map[string]string, error) {
```
nikhilachale: `configure` ignores `LaunchConfig.SystemPrompt`. Unlike the other native ACP bindings, Kimi has no `SessionMeta` configured and no ACP launch argument or environment overlay is added here, so the system prompt built by `session_manager` (including orchestrator delegation rules and worker safety instructions) is never delivered to `kimi acp`. The existing Kimi TUI path injects this through `.kimi-code/AGENTS.md`, but Chat startup does not run `GetAgentHooks`; please add a Kimi ACP-compatible prompt injection path and cover it with a test before registering this driver.

nikhilachale: Addressed in c94079f42. Kimi ACP configure now calls the existing Kimi adapter’s managed project-instructions path before launching, writing the current AO SystemPrompt into .kimi-code/AGENTS.md with the existing sentinel merge logic. This applies on both start and resume, preserves user-authored instructions, and keeps the generated AGENTS.md gitignored. Added adapter coverage asserting the prompt and gitignore are created. Focused Kimi/ACP tests and npm run lint pass.

nikhilachale:
Screenshot 2026-08-21 at 06 41 29

backend/internal/adapters/agent/kimi/hooks.go
```diff
@@ -45,25 +45,39 @@ func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfi
...
+		return fmt.Errorf("kimi.GetAgentHooks: %w", err)
+	}
+	return nil
+}
+
+// PrepareACPInstructions installs AO's standing instructions where Kimi Code
+// discovers project-level AGENTS.md files. Chat sessions call this immediately
+// before launching `kimi acp`; unlike TUI sessions, they do not run
+// GetAgentHooks during workspace preparation.
+func PrepareACPInstructions(workspacePath, systemPrompt string) error {
```
illegalcall: [P1] This attempted standing-prompt fix writes `.kimi-code/AGENTS.md`, but current Kimi does not discover that path. Its loader checks `.kimi/AGENTS.md`, `AGENTS.md`, and `agents.md` from project root to cwd: [current Kimi discovery source](https://github.com/MoonshotAI/kimi-cli/blob/cbc15c076d17f70fec9f89c90c0502e68657f505/src/kimi_cli/soul/agent.py#L87-L122).

As a result, the file-creation test passes while the provider still starts without AO's orchestrator/worker role and safety instructions.

Recommended fix: put the AO-managed sentinel block in `.kimi/AGENTS.md` (preserving user-authored content), apply it on start and resume, and add a provider-level test that asks Kimi to demonstrate it actually received `LaunchConfig.SystemPrompt` rather than only asserting that a file exists.

backend/internal/adapters/chatdriver/kimiacp/driver.go
```diff
@@ -0,0 +1,64 @@
...
+// config-option ids. Kimi advertises the complete model/thinking/mode catalog
+// during session setup; AO keeps that provider-owned catalog intact for direct
+// plan-mode and thinking selection in Chat.
+func sessionOptions(settings ports.ChatTurnSettings) []acpdriver.SessionOption {
+	var options []acpdriver.SessionOption
+	if settings.Model != "" {
+		options = append(options, acpdriver.SessionOption{ID: "model", Value: settings.Model})
+	}
+	if mode := permissionMode(settings.Approval); mode != "" {
+		options = append(options, acpdriver.SessionOption{ID: "mode", Value: mode})
```
illegalcall: [P1] These IDs are sent through `session/set_config_option`, but current Kimi advertises legacy `models`/`modes`, implements `session/set_model`, and exposes only the `default` session mode. Sources: [Kimi session response](https://github.com/MoonshotAI/kimi-cli/blob/cbc15c076d17f70fec9f89c90c0502e68657f505/src/kimi_cli/acp/server.py#L196-L212) and [Kimi setters](https://github.com/MoonshotAI/kimi-cli/blob/cbc15c076d17f70fec9f89c90c0502e68657f505/src/kimi_cli/acp/server.py#L324-L340).

AO tolerates method-not-found during start, so a selected model or `auto`/`yolo` permission mode can silently do nothing while the UI says it was applied.

Recommended fix: teach the generic ACP layer to consume legacy model/mode state and call the matching setters, or version-gate Kimi to a release that really advertises these config-option IDs. Permission mode needs a separately proven Kimi mapping; do not model `auto`/`yolo` as config options unless the live server advertises and enforces them. Add a contract test shaped like the official Kimi response.

skills/agent-role-smoke/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+cd frontend
+npm run dev
+```
+
+This starts both the daemon and Electron supervisor. Once the app is running, start testing the selected orchestrator or worker behavior directly. Do not run API generation, lint, typecheck, or broad checks by default; assume those are fine unless the user explicitly asks for them or the dev app/manual test exposes a problem.
+
+From the repo root, in another terminal:
+
+```bash
+node skills/agent-role-smoke/scripts/agent-role-smoke.mjs
```
illegalcall: This unrelated skill is left in the Kimi ACP PR even though its referenced smoke script was deleted by this branch, so the instructions are unusable and violate the one-issue/surgical-change rule.

Recommended fix: remove the entire `skills/agent-role-smoke` addition from this PR. If the skill is still wanted, restore it in a separate PR together with its executable and tests.

illegalcall (commented):
ACP/provider review against the current PR head. The inline findings below are merge blockers for the Kimi binding.

There is also a shared ACP prerequisite: AO currently treats the response to `session/load` as proof that the replay tail is settled. ACP v1 only says the response is sent after all conversation entries have been replayed; it does not report idle/running state. Source: [ACP session/load semantics](https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). A daemon restart while the native Kimi turn is unresolved can therefore be projected as completed/interrupted. Please fix the generic replay-tail classification (retain unknown/unsettled state or require provider/quiescence evidence) before claiming full load/resume correctness for this adapter.

backend/internal/adapters/agent/kimi/hooks.go
nikhilachale: Fixed in 94dfbdbb9. AO now writes the managed sentinel block to `.kimi/AGENTS.md`, preserves user-authored content, and applies the same configure path for start and resume. The adapter test asserts the provider-discovered path, and an opt-in live Kimi ACP contract test verifies `LaunchConfig.SystemPrompt` by requiring an exact handshake response.

backend/internal/adapters/chatdriver/kimiacp/driver.go
nikhilachale: Fixed in 94dfbdbb9. The generic ACP layer now consumes legacy `models`/`modes`, exposes them through the AO config catalog, calls `session/set_model` for Kimi model selection and `session/set_mode` for advertised legacy modes, and avoids `session/set_config_option` for those selectors. The invented `auto`/`yolo` permission mapping was removed. The contract test uses the official Kimi-shaped response and asserts the matching setters are called.

skills/agent-role-smoke/SKILL.md
nikhilachale: Fixed in 94dfbdbb9. Removed the entire unrelated `skills/agent-role-smoke` addition from this PR.

nikhilachale:
Replay-tail prerequisite fixed in 94dfbdbb9. Generic ACP history import no longer synthesizes `turn.completed` for the final replayed turn merely because `session/load` returned. A later user-turn boundary still settles the preceding turn. On restart, replayed `turn.started` now reopens only AO turns carrying the exact orphan-controller failure marker, leaving an unresolved native tail running/unsettled while preserving ordinary terminal turns. Added adapter and durable SQLite regressions for both behaviors. Verification: full backend `go test ./...` passed with runtime permissions; golangci-lint v2.12.2 reports 0 issues.

backend/internal/adapters/chatdriver/kimiacp/driver.go
```diff
@@ -0,0 +1,48 @@
...
+	return nativeacp.New(plugin, nativeacp.Config{
+		Harness: domain.HarnessKimi,
+		Capabilities: ports.ChatCapabilities{
+			ports.ChatCapabilityHistory: true,
+			ports.ChatCapabilityPlans:   true,
+		},
+		Configure:           configure,
+		SessionOptions:      sessionOptions,
+		LegacyModelSelector: true,
+	}, log)
```
illegalcall, line 30: [P1] Kimi still advertises approvals that it does not implement. `nativeacp.New` adds `ChatCapabilityApprovals`, but this config supplies no `SessionMode`, permission policy, or turn-settings validator, and `configure` ignores `cfg.Permissions`. A user can select accept-edits, auto, or bypass-permissions and AO silently runs unchanged. Either map these modes to a provider-documented Kimi control, or disable/reject unsupported non-default modes with a typed error.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -59,11 +59,11 @@ func (c *conversation) abortHistoryReplay() {
...
 
-// finishHistoryReplay settles the last reconstructed turn, then exposes the
-// captured facts through ChatHistoryReader for the controller's transactional,
-// idempotent import path.
+// finishHistoryReplay exposes the captured facts without classifying the replay
+// tail as terminal. ACP session/load guarantees that entries were replayed, but
+// does not say whether the provider was idle when persistence stopped.
 func (c *conversation) finishHistoryReplay() {
-	c.finishHistoryTurn()
+	c.finishHistoryTurn(false)
```
illegalcall: [P1] Omitting `TurnCompleted` is not a safe representation of an unknown replay outcome: it leaves the ordinary final replayed turn running, so imported Chat can remain busy indefinitely. ACP only guarantees that all entries were delivered before `session/load` returns; it supplies no historic terminal outcome (https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Close the lifecycle with a terminal recovered/unknown state (adding one if necessary), or return `ErrChatHistoryUnsettled`; do not leave a started turn unterminated.

illegalcall (changes requested):
Re-review of the updated Kimi head: the standing instructions, live prompt coverage, legacy model/mode setters, and unrelated-file cleanup are fixed. I still cannot approve because this branch conflicts with current main in generic ACP/storage files, and the two inline lifecycle/capability blockers remain. During rebase, retain main’s acp-go-sdk 0.13.5 and union the generic ACP changes rather than taking either side wholesale.

backend/internal/adapters/chatdriver/kimiacp/driver.go
nikhilachale: Fixed in 802456b4b. Kimi now rejects every non-default AO permission mode with ports.ErrChatPermissionModeUnsupported before launch/resume and again when applying per-turn settings. The default interactive ACP approval bridge remains available, but accept-edits, auto, and bypass-permissions can no longer be presented as successfully applied. Tests cover the provider validator and generic start/send enforcement.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in 802456b4b. The generic ACP replay capture now closes its local tail without emitting a fabricated terminal outcome, and ReadHistory returns the existing typed ErrChatHistoryUnsettled whenever session/load ends with a reconstructed tail. The controller therefore aborts resume/import instead of persisting a running turn indefinitely or projecting it as completed/interrupted. Prior turns with an observed next-user boundary remain safely settled.

nikhilachale:
Rebased the Kimi ACP changes onto current Untrivial-ai/main (7d0a03f8e) and force-updated the PR with an exact lease. The branch retains acp-go-sdk v0.13.5, preserves main generic ACP/storage behavior, restores Kimi legacy models/session-set_model plus legacy modes/session-set_mode through an official-response-shaped contract test, rejects unsupported Kimi approval policies with a typed error, and returns ErrChatHistoryUnsettled for ambiguous session/load tails. Local verification: focused adapter/service suites pass, ACP/native/Kimi race tests pass, full backend go test ./... passes, and golangci-lint v2.12.2 reports 0 issues. GitHub now reports the PR MERGEABLE; CI is queued.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -64,28 +63,28 @@ func (c *conversation) abortHistoryReplay() {
...
-// idempotent import path.
+// finishHistoryReplay closes the capture without inventing an outcome for its
+// final turn. ACP session/load guarantees only that all stored entries were
+// replayed; it does not say whether that tail completed, failed, or was still
+// running when persistence stopped.
 func (c *conversation) finishHistoryReplay() {
 	c.historyMu.Lock()
-	interrupted := c.history != nil && c.history.turnID != "" &&
-		c.history.turnUserID != "" && !c.history.turnHasProvider
+	unsettled := c.history != nil && c.history.turnID != ""
```
illegalcall: [P1] This condition is true for the final turn of every non-empty replay, including a normal fully answered conversation. The code then stores `ErrChatHistoryUnsettled`; because ACP load history is a frozen, non-refreshable snapshot, controller startup rejects it. The new normal two-turn test explicitly expects this failure, so Kimi’s claimed load/resume path is unusable after any conversation. ACP still does not justify guessing completed/interrupted, so fix this in shared ACP core: close the tail as a terminal but outcome-unknown/recovered state (preserving an AO-known terminal state when available), and reserve `ErrChatHistoryUnsettled` for history that is actually unsafe/incomplete—not every final turn.

backend/internal/adapters/agent/kimi/hooks.go
illegalcall: [P2] This helper performs reads, directory creation, atomic writes, and gitignore writes without `context.Context`, contrary to AGENTS.md’s I/O rule. Accept `ctx` as the first argument, check it around the filesystem operations, and make `nativeacp.Configure` context-aware so the Chat launch path can pass its context too (the Cursor branch already demonstrates this signature change).

illegalcall (changes requested):
Latest-head re-review: the branch now merges cleanly; the SDK downgrade/history pollution is gone; `.kimi/AGENTS.md`, legacy model handling, and typed rejection of unsupported permission modes are fixed. Two current blockers remain inline. Also add Kimi to the canonical Chat-driver list in docs/STATUS.md. The right architectural fix is to land one ACP-core replay-tail policy, then rebase each provider adapter onto it.

backend/internal/adapters/agent/kimi/hooks.go
nikhilachale: Fixed in 339a2388c. `nativeacp.Configure` is now context-aware and receives the launch context. Kimi passes it into `PrepareACPInstructions(ctx, ...)`, which checks cancellation before and around the read, directory creation, atomic write, and gitignore update. Tests cover exact context propagation through native ACP and pre-canceled Kimi preparation without creating `.kimi`.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in 339a2388c. Shared ACP replay now closes a provider-answered tail as the new terminal `recovered` outcome, while a user-only tail remains `ErrChatHistoryUnsettled`. The controller reconciliation test also proves an AO-known terminal state is preserved over the synthetic recovered projection. Migration 0107 persists the new state and verifies dependent rows plus branch/CDC triggers survive the table rebuild; the normal two-turn replay test now loads successfully with completed + recovered turns.

nikhilachale:
Latest review blockers are addressed in 339a2388c, rebased onto current main (7ef5f3eee): shared ACP replay tails now use a persisted terminal `recovered` outcome when provider output exists, unsafe user-only tails still return `ErrChatHistoryUnsettled`, AO-known terminal outcomes win during reconciliation, native ACP configuration carries launch context through Kimi filesystem I/O, and Kimi is listed in `docs/STATUS.md`. Verification: affected backend/storage/API suites, ACP/Kimi race tests, frontend typecheck, generated API drift check, and golangci-lint v2.12.2 (`0 issues`) pass. The full backend run hit only the known load-sensitive fake-agent speedup timing test; its immediate uncached isolated rerun passed in 0.627s.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -64,28 +64,33 @@ func (c *conversation) abortHistoryReplay() {
...
 	}
-	c.finishHistoryTurn(turnState)
 
 	c.historyMu.Lock()
 	if c.history != nil {
 		c.historyEvents = append([]ports.ChatEvent(nil), c.history.events...)
-		c.historyErr = nil
+		if unsettled {
+			c.historyErr = fmt.Errorf("%w: ACP session/load does not report the replay tail outcome",
+				ports.ErrChatHistoryUnsettled)
```
illegalcall: [P1] A user-only replay tail must not make the conversation impossible to reopen. It can be a valid persisted snapshot from immediately before work began or before a crash; `session/load` gives no outcome either way. Returning `ErrChatHistoryUnsettled` aborts controller startup, so the user cannot resume the conversation. Close this turn as terminal `recovered`/outcome unknown and allow the controller to become ready, matching #4194.

backend/internal/storage/sqlite/migrations/0107_conversation_turn_recovered_state.sql
```diff
@@ -0,0 +1,87 @@
...
+    controller_generation, state, error_message, requested_at, started_at,
+    completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+    promotion_started_at, promoted_to_turn_id
+)
+SELECT
+    id, conversation_id, handled_by_session_id, provider_turn_id,
+    controller_generation, state, error_message, requested_at, started_at,
+    completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+    promotion_started_at, promoted_to_turn_id
+FROM conversation_turns;
```
illegalcall: [P1] Preserve row order during this table rebuild. AO rollback ranges use `rowid` to break ties between turns with the same timestamp; a bare `SELECT` may assign new rowids in a different order and change which turns a rollback includes. Add `ORDER BY rowid`—or, preferably, drop this duplicate 0107 when rebasing onto #4194’s shared migration, which already preserves the ordering.

illegalcall (changes requested):
The Kimi-specific fixes are now in good shape, but the shared ACP replay model still has correctness blockers. ACP `session/load` guarantees replay delivery; it does not communicate historic turn outcomes. Source: https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188

Two related gaps are outside the currently changed hunks: `captureHistoryUserChunk` still closes every earlier replay turn as `completed` merely because a later user message exists; that boundary proves only that it is no longer live, so it must be `recovered`. Also, when a recovered turn closes still-running child activities, `terminalActivityStatus` falls through to `failed`, inventing a failure ACP did not report; use the shared `ActivityStatusRecovered`.

Recommended integration: merge #4194 first, rebase this PR onto it, remove this PR’s duplicate 0107/domain/API/UI replay-state implementation, and keep the Kimi adapter changes plus adapter tests.

backend/internal/storage/sqlite/migrations/0107_conversation_turn_recovered_state.sql
nikhilachale: Fixed in a05c76178 by dropping this PR’s duplicate `0107_conversation_turn_recovered_state.sql` and rebasing onto #4194’s shared `0107_recovered_conversation_turns.sql`. The shared migration copies conversation turns with `ORDER BY rowid` and activities with `ORDER BY sequence`, so rollback tie ordering is preserved.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in a05c76178 by rebasing onto merged #4194. The user-only tail now closes as terminal `recovered` and `ReadHistory` succeeds, so controller startup can become ready. The same shared ACP implementation also closes earlier replay turns as `recovered` and settles their open child activities with `ActivityStatusRecovered`. This PR retains the Kimi/legacy-selector adapter tests and no longer carries a separate replay-state implementation.

nikhilachale:
Rebased onto current main with merged #4194 at a05c76178. PR #4193 now keeps only the Kimi adapter, shared ACP legacy-selector support, context-aware native ACP launch configuration, provider tests, and the Kimi status entry. The duplicate recovered turn/activity domain, migration 0107, storage, service, API, and UI implementation has been removed; shared ACP now supplies recovered earlier turns, recovered user-only tails, recovered open activities, and row-order-preserving migration behavior. Verification: focused replay/Kimi/native/registry tests, affected race tests, full `go test ./...`, `git diff --check`, and golangci-lint v2.12.2 (`0 issues`) all pass.

illegalcall (approved):
Approved after reviewing the rebased `a268c5fd7` head and the author’s updated replies. The duplicate recovered-turn/activity migration, storage, service, API, and UI implementation is gone; Kimi now inherits main’s canonical ACP replay behavior unchanged. The provider-specific work correctly installs standing instructions in `.kimi/AGENTS.md` with context-aware I/O, preserves Kimi’s legacy model/mode selectors without downgrading the SDK, rejects unsupported non-default permission modes, and keeps default ACP approval round-trips. The effective merge is clean; focused tests, race tests, vet, `go build ./...`, diff checks, and CI pass.

Non-blocking follow-up: consider enabling the legacy wire shim only for bindings that require it, and distinguish a synthesized legacy selector from a future provider that advertises both legacy and modern options.

illegalcall (approved):

PR #4194, opened 2026-08-20
feat: add Pi ACP chat driver

nikhilachale:
Add Pi ACP chat driver with ACP lifecycle support.

## Implementation
- **backend/internal/adapters/chatdriver/piacp/driver.go** - Pi ACP driver
- **backend/internal/adapters/chatdriver/piacp/driver_test.go** - Pi-specific tests
- **backend/internal/adapters/chatdriver/piacp/handoff_test.go** - Handoff tests
- **backend/internal/adapters/chatdriver/registry/** - Registry integration
- **docs/STATUS.md** - Status documentation
- **docs/harnesses/pi.md** - Pi harness documentation

## Features
- Session creation and lifecycle management
- Streaming message support
- Tool event handling
- Explicit `bypass-permissions` admission for Pi ACP, which has no approval channel
- Image and attachment support
- Session load/resume with structured replay; an ambiguous user-only replay tail is rejected as unsettled
- Model configuration

## Testing
- Focused Pi/registry tests pass
- Shared ACP lifecycle tests pass
- HTTP approval-mode contract tests pass
- Full backend test suite passes
- Frontend typecheck and focused fallback-action tests pass
- Lint checks pass

## Notes
- Chat-only implementation (no TUI handoff)
- Uses pre-installed `pi-acp` adapter (does not auto-download)
- Reuses existing Pi adapter for config/auth
- ACP adapter sees the same provider config as TUI
- Configured per-session MCP servers are rejected because pi-acp accepts MCP parameters without wiring them to Pi


backend/internal/adapters/chatdriver/piacp/driver.go
```diff
@@ -0,0 +1,221 @@
...
+func buildConfig(plugin piPlugin, resolve binaryResolver, log *slog.Logger) acpdriver.Config {
+	return acpdriver.Config{
+		Harness: domain.HarnessPi,
+		Capabilities: ports.ChatCapabilities{
+			ports.ChatCapabilityStreaming: true,
+			ports.ChatCapabilityTools:     true,
+			// Pi deliberately executes tools without permission popups. Marking
+			// this true records that there is no unresolved approval gate; pi-acp
+			// never emits permission requests for AO to display or answer.
+			ports.ChatCapabilityApprovals: true,
```
nikhilachale: [P1] Do not advertise `ChatCapabilityApprovals` unless Pi ACP actually enforces AO's selected permission mode. This driver provides no `SessionMode` mapping, does not handle ACP permission requests, and the Pi adapter's documented behavior is to execute without permission popups, so `default`/`accept-edits`/`bypass-permissions` selections are silently ignored while the UI exposes the Approvals picker and the production-floor check treats the driver as safe. Either implement a real permission boundary or report this capability as false and prevent mutating Chat sessions from being admitted.

skills/agent-role-smoke/scripts/agent-role-smoke.mjs
```diff
@@ -0,0 +1,861 @@
...
+  return text.includes("SESSION_AWAITING_DECISION") || text.toLowerCase().includes("awaiting decision");
+}
+
+function predictedOrchestratorWorkspace(dataDir, projectID) {
+  const prefix = String(projectID).slice(0, 12);
+  return path.join(dataDir, "worktrees", String(projectID), "orchestrator", `${prefix}-orchestrator`);
+}
+
+function pretrustCursorWorkspace(opts, workspacePath) {
+  const dir = path.join(homedir(), ".cursor", "projects", cursorProjectKey(workspacePath));
```
nikhilachale: [P1] This writes Cursor trust state under `~/.cursor`, directly violating the repository's hard rule that all app state must live under `~/.ao` and potentially modifying a user's real Cursor trust configuration whenever the smoke test runs for Cursor. Use an isolated/configurable Cursor data root under `AO_DATA_DIR` (or avoid pretrusting by mutating the user's home), and clean up any test state.

nikhilachale (commented):
Changes requested.

The Pi ACP wiring is close, but it currently advertises an approval capability that it does not implement, and the new smoke helper writes outside AO's state root. These are safety/state-isolation blockers.

nikhilachale:
Screenshot 2026-08-21 at 06 09 52

backend/internal/service/chat/service.go
```diff
@@ -244,8 +244,10 @@ func (s *Service) Start(ctx context.Context, cfg StartConfig) (*Controller, erro
 	if err != nil {
 		return nil, err
 	}
-	if missing := ports.MissingProductionCapabilities(caps); len(missing) > 0 {
-		return nil, fmt.Errorf("%w: %s lacks %v", ports.ErrChatUnsupported, cfg.Harness, missing)
+	if cfg.ProviderConversationID == "" {
+		if missing := ports.MissingCapabilitiesForPermissions(caps, cfg.Permissions); len(missing) > 0 {
```
illegalcall: [P1 integration] This permission-aware admission is the right idea, but the branch now conflicts with current `main`, where probing is routed through `ensureDriverReady` and only fully production-capable probes are cached. A mechanical conflict resolution that keeps that call first will reject Pi before the explicit bypass can be considered; dropping the cache makes every start spawn/probe the adapter again.

Recommended fix: rebase, make the cache store the raw probed `ChatCapabilities` (not a boolean meaning 'passed the full floor'), then evaluate `MissingCapabilitiesForPermissions(caps, requestedMode)` at each preflight/start. Update every current-main `PreflightChat` caller and cover default rejection, explicit bypass admission, persisted bypass resume, and repaired-install retry.

frontend/src/renderer/components/TaskComposer.tsx
```diff
@@ -254,12 +261,20 @@ export function TaskComposer({
...
+				approvalMode,
 				attachments: attachmentPayloads.length > 0 ? attachmentPayloads : undefined,
 			});
 			onCreated(sessionId);
 		} catch (err) {
+			const canBypassPiApprovals =
+				selectedAgent === "pi" &&
+				err instanceof TaskCreateError &&
+				err.code === "SESSION_MODE_UNSUPPORTED" &&
+				err.message.includes("lacks [approvals]");
```
illegalcall: This safety decision depends on the selected provider plus an English error-message substring. Any wording/format change—or another approval-less ACP adapter—silently removes the explicit bypass recovery path. It also conflates a structured capability fact with presentation text.

Recommended fix: return structured missing-capability data in the API error (for example `missingCapabilities: ["approvals"]` plus an allowed/recommended explicit override), and render the action from that typed data. Rename/generalize the fallback action instead of routing 'start without approvals' through `canCreateAsTUI`/`onSubmitAsTui`.

illegalcall (commented):
ACP/provider review against the current PR head. The approval direction is correct: pi-acp does not request ACP permissions, so reporting approvals as false and requiring an explicit `bypass-permissions` choice is honest. Upstream source: [pi-acp limitations](https://github.com/victor-software-house/pi-acp/blob/0ef24b24c97ac81a5e87a17d8fd74ef97fb34d8b/README.md#L243-L273).

Two corrections are still needed in the PR description/coverage: it lists MCP integration, but upstream explicitly accepts MCP params without wiring them to Pi. The implementation/docs correctly reject configured MCP servers, so remove the claim from the PR body. Add HTTP/controller tests proving a valid `approvalMode` reaches the service and an invalid value preserves the `INVALID_APPROVAL_MODE` envelope.

Shared ACP prerequisite: `session/load` returning is only a complete-replay barrier, not proof that the final turn is idle. Source: [ACP session/load semantics](https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Fix the generic replay-tail classification before claiming load/resume correctness for Pi.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -69,18 +69,24 @@ func (c *conversation) abortHistoryReplay() {
...
-		// emitted a provider event and remains valid replayable history.
-		turnState = domain.TurnStateInterrupted
+	if unsettled {
+		c.mu.Lock()
+		c.activeTurn = ""
+		c.mu.Unlock()
+		return
 	}
-	c.finishHistoryTurn(turnState)
+	c.finishHistoryTurn(domain.TurnStateCompleted)
```
illegalcall: [P1] The user-only guard is safer, but this still invents success for a final tail as soon as any provider chunk exists. If the process died after partial output, `session/load` replays that partial chunk and this records the turn as completed. ACP guarantees replay delivery, not a terminal outcome (https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Use one shared ACP-core policy: preserve a known AO terminal result, consume explicit provider outcome evidence, or close the tail as terminal recovered/unknown; do not infer completed from the presence of output.

illegalcall (changes requested):
Re-review of the updated Pi head: explicit bypass admission, permission persistence, typed API errors/tests, standing instructions, and MCP rejection are fixed and aligned with pi-acp. I still cannot approve: the branch conflicts with current main in session_manager/manager.go; resolve it by preserving main’s modeExplicitlyRequested/default-Chat-to-TUI fallback and only layering in the permission preflight argument. Also pass context through the standing-instruction filesystem hook per AGENTS.md.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -64,28 +63,28 @@ func (c *conversation) abortHistoryReplay() {
...
-// idempotent import path.
+// finishHistoryReplay closes the capture without inventing an outcome for its
+// final turn. ACP session/load guarantees only that all stored entries were
+// replayed; it does not say whether that tail completed, failed, or was still
+// running when persistence stopped.
 func (c *conversation) finishHistoryReplay() {
 	c.historyMu.Lock()
-	interrupted := c.history != nil && c.history.turnID != "" &&
-		c.history.turnUserID != "" && !c.history.turnHasProvider
+	unsettled := c.history != nil && c.history.turnID != ""
```
illegalcall: [P1] This rejects every non-empty replay, not only the ‘ambiguous user-only replay tail’ claimed in the PR body. A normal completed transcript still has a final `turnID`, so `ReadHistory` returns `ErrChatHistoryUnsettled`; because the load snapshot cannot refresh, ordinary Pi restart/resume fails. Do not revert to synthetic `Completed`: instead add/use a shared terminal outcome-unknown/recovered turn state, preserve AO-known terminal evidence when available, and let the provider session become ready without pretending the historical tail succeeded or blocking all resume.

illegalcall (changes requested):
Latest-head re-review: the branch now merges cleanly and preserves main’s Chat→TUI fallback; context-aware instruction I/O, explicit bypass admission, permission persistence, typed API errors/tests, and MCP rejection are fixed. The remaining ACP lifecycle blocker is inline. It should be solved once in ACP core rather than with a Pi-specific replay policy.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in ACP core at 62028c0da (implementation commit b819cb03a). `session/load` now closes every replayed turn boundary and tail as the shared terminal `recovered` outcome instead of returning `ErrChatHistoryUnsettled` or synthesizing `completed`, so ordinary non-empty resumes become ready. Reconciliation preserves AO-known completed/interrupted/failed outcomes when ACP only reports recovered, while a later explicit provider outcome can upgrade a durable recovered observation. Replayed open tools/reasoning also settle as recovered rather than fabricated failures; the state is persisted through migration 0107 and exposed/rendered as “Outcome unknown”. Added ACP replay, reconciliation precedence, real SQLite import/migration, and renderer regressions. Verified with the full backend suite, focused race tests, frontend typecheck, and renderer tests. The branch is merged with current main and GitHub reports it conflict-free.

illegalcall (approved):
Approved as the canonical ACP replay foundation. The latest implementation correctly represents every replayed turn and unresolved activity as terminal `recovered` (outcome unknown), preserves durable AO completed/interrupted/failed evidence during reconciliation, persists the state through the domain/storage/API/frontend contracts, and passes CI plus local effective-merge tests and vet.

Please merge this before the other ACP adapter PRs and rebase them onto it so there is one shared replay policy and one migration. Non-blocking: update the PR body sentence saying a user-only replay tail is rejected as unsettled; the current implementation correctly recovers it instead.

PR #4196, opened 2026-08-20
feat: add OMP ACP chat driver

nikhilachale:
Add OMP Chat through its native ACP server, using the exact binary and authentication resolved by the existing OMP adapter.

## Implementation
- Launch only native `omp acp`; no RPC fallback is attempted.
- Require OMP 15.0.0 or newer, the tested minimum containing the native ACP command.
- Fix approval policy at process launch and reject incompatible per-turn changes.
- Register OMP in the shipped Chat-driver set and canonical STATUS list.
- Expose OMP in spawn/delegation DTO enums, generated OpenAPI, and frontend schema types.
- Add controller/spec contract tests plus OMP launch, version, approval, registry, and opt-in live tests.

## Shared ACP prerequisite

This branch is rebased onto current `main`, including #4194's shared ACP replay implementation. #4194 owns recovered turn/activity states, migration 0107, storage/API/frontend propagation, user-only and provider-output replay-tail handling, and preservation of known durable outcomes. This PR no longer duplicates or modifies those shared lifecycle files.

## Verification
- `npm run api`
- `npm run lint` (full backend tests plus golangci-lint: 0 issues)
- focused ACP/native ACP/OMP/registry/HTTP/domain/storage tests
- focused `go vet`

The branch is one OMP-only commit on current `main`.

### DIFF
--- a/backend/internal/adapters/chatdriver/ompacp/driver.go
+++ b/backend/internal/adapters/chatdriver/ompacp/driver.go
@@ -1,0 +1,81 @@
+// Package ompacp binds the user's own OMP installation to AO's reusable ACP
+// Chat transport.
+package ompacp
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	acpdriver "github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/acp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/nativeacp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// New launches `omp acp` from the exact binary resolved by the existing OMP
+// agent plugin. OMP keeps filesystem and terminal execution in its own process;
+// AO advertises neither ACP client capability and receives only typed session
+// updates and permission requests.
+func New(plugin nativeacp.Plugin, log *slog.Logger) ports.ChatDriver {
+	return newDriver(plugin, versionProbe, log)
+}
+
+func newDriver(plugin nativeacp.Plugin, probe nativeacp.VersionProbe, log *slog.Logger) ports.ChatDriver {
+	return nativeacp.New(plugin, nativeacp.Config{
+		Harness: domain.HarnessOMP,
+		Capabilities: ports.ChatCapabilities{
+			ports.ChatCapabilityUsage: true,
+			ports.ChatCapabilityDiffs: true,
+			ports.ChatCapabilityPlans: true,
+		},
+		Configure:            configure,
+		SessionOptions:       sessionOptions,
+		ValidateTurnSettings: validateTurnSettings,
+		VersionProbe:         probe,
+	}, log)
+}
+
+func configure(_ context.Context, cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	args := []string{"acp"}
+	if model := strings.TrimSpace(cfg.Model); model != "" {
+		args = append(args, "--model", model)
+	}
+	if prompt := strings.TrimSpace(cfg.SystemPrompt); prompt != "" {
+		args = append(args, "--append-system-prompt", prompt)
+	}
+	switch ports.NormalizePermissionMode(cfg.Permissions) {
+	case ports.PermissionModeAcceptEdits, ports.PermissionModeAuto:
+		args = append(args, "--approval-mode", "write")
+	case ports.PermissionModeBypassPermissions:
+		args = append(args, "--approval-mode", "yolo")
+	}
+	return args, nil, nil
+}
+
+func sessionOptions(settings ports.ChatTurnSettings) []acpdriver.SessionOption {
+	if model := strings.TrimSpace(settings.Model); model != "" {
+		return []acpdriver.SessionOption{{ID: "model", Value: model}}
+	}
+	return nil
+}
+
+func validateTurnSettings(initial ports.PermissionMode, settings ports.ChatTurnSettings) error {
+	if settings.Approval == "" || ompApprovalMode(settings.Approval) == ompApprovalMode(initial) {
+		return nil
+	}
+	return fmt.Errorf("%w: OMP ACP approval mode is fixed at process launch (%s); restart Chat to change it to %s",
+		acpdriver.ErrACPSetterUnsupported, ompApprovalMode(initial), ompApprovalMode(settings.Approval))
+}
+
+func ompApprovalMode(mode ports.PermissionMode) string {
+	switch ports.NormalizePermissionMode(mode) {
+	case ports.PermissionModeAcceptEdits, ports.PermissionModeAuto:
+		return "write"
+	case ports.PermissionModeBypassPermissions:
+		return "yolo"
+	default:
+		return "default"
+	}
+}
--- a/backend/internal/adapters/chatdriver/registry/registry.go
+++ b/backend/internal/adapters/chatdriver/registry/registry.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/cursor"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/droid"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/kimchi"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/omp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/opencode"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/pi"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/claudeacp"
@@ -22,6 +23,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/cursoracp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/droidacp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/kimchiacp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/ompacp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/opencodeacp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/piacp"
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
@@ -52,9 +54,9 @@ func New(drivers ...ports.ChatDriver) *Registry {
 //
 // Codex uses its native app-server protocol. Claude Code uses AO's reusable ACP
 // transport plus claude-agent-acp, pointed at the user's own Claude executable.
-// Cursor, OpenCode, Droid, Kimchi, and Pi expose ACP themselves, so AO launches
-// the exact executable resolved by each existing agent plugin. No path scrapes
-// terminal output or packages a second provider CLI.
+// Cursor, OpenCode, Droid, Kimchi, Pi, and OMP expose ACP themselves, so AO
+// launches the exact executable resolved by each existing agent plugin. No path
+// scrapes terminal output or packages a second provider CLI.
 //
 // Every other harness stays TUI-only until the same is true of it. The driver
 // reuses the harness's existing agent plugin for binary resolution and auth, so
@@ -68,6 +70,7 @@ func Build(log *slog.Logger) *Registry {
 		kimchiacp.New(kimchi.New(), log),
 		piacp.New(pi.New(), log),
 		cursoracp.New(cursor.New(), log),
+		ompacp.New(omp.New(), log),
 	)
 }
 
--- a/backend/internal/adapters/chatdriver/ompacp/driver_test.go
+++ b/backend/internal/adapters/chatdriver/ompacp/driver_test.go
@@ -1,0 +1,150 @@
+package ompacp
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"testing"
+
+	acpdriver "github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/acp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+func TestConfigureUsesNativeACP(t *testing.T) {
+	tests := []struct {
+		name string
+		cfg  acpdriver.LaunchConfig
+		want []string
+	}{
+		{name: "defaults", want: []string{"acp"}},
+		{
+			name: "model prompt and write approvals",
+			cfg: acpdriver.LaunchConfig{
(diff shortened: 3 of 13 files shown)


nikhilachale:
Screenshot 2026-08-21 at 05 06 07

backend/internal/adapters/chatdriver/registry/registry.go
```diff
@@ -62,6 +64,7 @@ func Build(log *slog.Logger) *Registry {
 		opencodeacp.New(opencode.New(), log),
 		droidacp.New(droid.New(), log),
 		kimchiacp.New(kimchi.New(), log),
+		ompacp.New(omp.New(), log),
```
illegalcall, line 73: [P1] Registering OMP here makes the daemon report Chat support, but the public request contracts still omit `omp`: `SpawnSessionRequest.Harness` and `DelegateTaskRequest.Agent` do not include it, so generated OpenAPI/frontend types cannot correctly create or delegate an OMP Chat session. The adapter can be present internally while the supported public path rejects or cannot represent it.

Recommended fix: add `omp` to both enum tags in `backend/internal/httpd/controllers/dto.go`, run `npm run api`, commit `openapi.yaml` and `frontend/src/api/schema.ts`, and add controller/spec tests that create and delegate OMP Chat.

backend/internal/adapters/chatdriver/ompacp/driver.go
```diff
@@ -0,0 +1,75 @@
...
+			ports.ChatCapabilityPlans: true,
+		},
+		Configure:            configure,
+		SessionOptions:       sessionOptions,
+		ValidateTurnSettings: validateTurnSettings,
+	}, log)
+}
+
+func configure(cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	args := []string{"acp"}
```
illegalcall, line 41: The PR body says it decides between `omp acp` and `omp --mode rpc`, but this implementation always selects `omp acp` and has no version/capability probe or fallback. That leaves the stated compatibility behavior unimplemented and lets an incompatible installed OMP pass binary/auth probing only to fail during Chat startup.

Recommended fix: either (1) define and enforce a tested minimum OMP version that supports `omp acp`, then remove the RPC-selection claim, or (2) implement a real version/capability decision with contract tests for both launch forms. Do not silently fall back between two transports unless they are proven to expose the same session, approval, and history semantics.

illegalcall (commented):
ACP/provider review against the current PR head. The process-level approval design is aligned with OMP: ACP has no per-session approval-policy field, and OMP recommends launching a separate ACP process with the desired policy flags. Source: [OMP ACP approval documentation](https://github.com/can1357/oh-my-pi/blob/76a294cb19bfded1e32e2111f1f729129595bf5e/docs/approval-mode.md#L122-L146). Keeping the launch mode durable and rejecting a per-turn change is the right direction.

Shared ACP prerequisite: `session/load` completion is only a replay barrier, not an idle/current-turn signal. Source: [ACP session/load semantics](https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Correct the generic replay-tail classification before claiming full load/resume lifecycle safety. Also add OMP to the canonical shipped Chat-driver list in `docs/STATUS.md` once the public contract is fixed.

backend/internal/adapters/chatdriver/registry/registry.go
nikhilachale: Fixed in bbe2ecb3b. Added omp to both public request enums, regenerated OpenAPI and frontend types, and added controller plus spec coverage for OMP Chat spawn and delegation.

backend/internal/adapters/chatdriver/ompacp/driver.go
nikhilachale: Fixed in bbe2ecb3b. OMP Chat now requires version 15.0.0 or newer, the first tagged release containing native omp acp. The driver uses only that transport, has version-gate tests, and the unsupported RPC-selection claim was removed from the PR body.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -68,19 +67,10 @@ func (c *conversation) abortHistoryReplay() {
...
-		// session/load has finished, so this user-only turn cannot still be
-		// producing output. It represents a prompt interrupted before the agent
-		// emitted a provider event and remains valid replayable history.
-		turnState = domain.TurnStateInterrupted
-	}
-	c.finishHistoryTurn(turnState)
+	// session/load returning is only a replay barrier. ACP does not attach a
+	// portable outcome to replayed turns, so a trailing user-only entry is not
+	// evidence that the turn was interrupted (or that it is still current).
+	c.finishHistoryTurn(domain.TurnStateCompleted)
```
illegalcall: [P1] The comment correctly says ACP provides no outcome, but this line immediately invents `Completed` anyway. If the provider persisted a prompt or partial response before crashing, replay returns those entries and AO imports the unresolved turn as successful/settled. `session/load` only guarantees that all conversation entries were replayed before the response (https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Preserve known AO/provider terminal evidence or add a terminal recovered/unknown state; at minimum, return `ErrChatHistoryUnsettled` instead of completing a user-only tail. Update the test that currently requires this false completed result.

illegalcall (changes requested):
Re-review of the updated OMP head: the public enums/OpenAPI/frontend artifacts, HTTP contract tests, native-ACP-only v15 gate, STATUS entry, process wrapper, and process-fixed approval validation are all corrected. The focused ACP/nativeacp/OMP/registry/HTTP tests and vet pass on a synthetic merge with current main. I still cannot approve because the inline generic ACP replay change creates a false-success lifecycle result. This policy should ideally land once in an ACP-core prerequisite and all four provider branches should rebase onto it.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in 135c05099. Shared ACP replay now tracks whether the trailing reconstructed user turn received any provider update. If session/load ends on a user-only tail, ReadHistory returns ErrChatHistoryUnsettled and exposes no fabricated TurnCompleted event; provider-backed settled history remains importable. I updated the regression test through a red/green cycle (it first failed with ReadHistory error = <nil>) and reran the focused ACP/nativeacp/OMP/registry/HTTP tests plus go vet successfully.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -64,23 +64,28 @@ func (c *conversation) abortHistoryReplay() {
...
-		// producing output. It represents a prompt interrupted before the agent
-		// emitted a provider event and remains valid replayable history.
-		turnState = domain.TurnStateInterrupted
+	if unsettled {
+		c.resetHistoryItems("")
+		return
 	}
-	c.finishHistoryTurn(turnState)
+
+	c.finishHistoryTurn(domain.TurnStateCompleted)
```
illegalcall: [P1] This still turns any replay tail with at least one provider update into `Completed`. A crash after a partial assistant/tool update therefore becomes a false success; ACP `session/load` proves delivery ended, not that the historic turn completed. Keep the new user-only guard, but solve the general tail outcome in shared ACP core: use a terminal outcome-unknown/recovered state (and preserve known AO/provider terminal evidence) rather than inferring completion from output presence. Add a regression with a user message plus partial provider chunk that must not import as completed.

illegalcall (changes requested):
Latest-head re-review: user-only replay is now rejected, and the OMP adapter’s API/schema, v15 native-ACP gate, STATUS entry, process wrapper, and process-fixed approval validation remain correct. Focused tests and vet pass on the effective merge. The provider-output replay blocker remains inline. Separately, rebuild/rebase or guarantee squash merging: this PR’s history still contains unrelated orchestrator-handoff and agent-role-smoke commits, contrary to the repository’s one-issue-per-PR rule.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in the rebuilt head 230bbe812. Shared ACP now classifies the final provider-output replay tail as terminal recovered (rendered as Outcome unknown) instead of Completed; a trailing user-only turn still returns ErrChatHistoryUnsettled. Earlier turns bounded by a later user message remain completed, and native-history reconciliation preserves any known durable AO/provider terminal state over recovered. Added a red/green regression with a user message plus partial assistant chunk, domain/API/frontend support, and migration 0107 with transcript/FK preservation coverage. The PR branch was also rebuilt on current main as exactly one OMP/ACP commit, removing the unrelated handoff and agent-role-smoke history. The complete affected Go test/vet matrix, frontend typecheck, 62 ChatWorkspace tests, API regeneration, and sqlc regeneration pass on the rebuilt tree.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -64,23 +64,31 @@ func (c *conversation) abortHistoryReplay() {
...
+	// session/load returning is only a replay barrier. A trailing user entry with
+	// no provider update may have been persisted immediately before a crash, so
+	// assigning it any terminal outcome would invent lifecycle evidence.
 	c.historyMu.Lock()
-	interrupted := c.history != nil && c.history.turnID != "" &&
+	unsettled := c.history != nil && c.history.turnID != "" &&
 		c.history.turnUserID != "" && !c.history.turnHasProvider
+	if unsettled {
+		c.historyEvents = nil
+		c.historyErr = fmt.Errorf("%w: final replayed user turn has no provider output", ports.ErrChatHistoryUnsettled)
```
illegalcall: [P1] Do not reject a user-only replay tail. It is a valid persisted snapshot whose outcome ACP does not disclose; rejecting it through `ErrChatHistoryUnsettled` makes controller startup fail and prevents the user from reopening the conversation. Represent it as terminal `recovered`/outcome unknown and continue startup, as #4194 does.

backend/internal/storage/sqlite/store/conversation_store.go
```diff
@@ -752,6 +752,10 @@ func terminalActivityStatus(state domain.TurnState) domain.ActivityStatus {
 		return domain.ActivityStatusCompleted
 	case domain.TurnStateInterrupted:
 		return domain.ActivityStatusCancelled
+	case domain.TurnStateRecovered:
+		// The enclosing replay had no portable outcome. Any activity still open at
+		// that boundary is not evidence of completion, so settle it conservatively.
+		return domain.ActivityStatusFailed
```
illegalcall: [P1] `failed` is just as unsupported as `completed` here. A still-open child activity under a recovered turn has an unknown historic outcome; ACP did not report failure. Add/use `ActivityStatusRecovered`, carry it through storage/API/frontend, and settle the activity to that state. #4194 already contains the complete shared implementation.

backend/internal/storage/sqlite/migrations/0107_recovered_conversation_turn_state.sql
```diff
@@ -0,0 +1,88 @@
...
+    controller_generation, state, error_message, requested_at, started_at,
+    completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+    promotion_started_at, promoted_to_turn_id
+)
+SELECT
+    id, conversation_id, handled_by_session_id, provider_turn_id,
+    controller_generation, state, error_message, requested_at, started_at,
+    completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+    promotion_started_at, promoted_to_turn_id
+FROM conversation_turns;
```
illegalcall: [P1] Preserve row order during this rebuild. Rollback ranges use `rowid` to order turns that share a timestamp, so a bare `SELECT` can silently change rollback membership after migration. Add `ORDER BY rowid`—or preferably remove this duplicate 0107 after rebasing onto #4194’s migration, which already handles turn/activity states and ordering.

illegalcall (changes requested):
The OMP adapter/API/version/permission work and the cleaned single-commit branch are now in good shape. The remaining blockers are all in the shared ACP replay policy. ACP `session/load` guarantees replay delivery; it does not communicate historic turn outcomes. Source: https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188

One related gap is outside the current changed hunks: `captureHistoryUserChunk` still closes every earlier replay turn as `completed` merely because a later user message exists. That boundary proves only that the earlier turn is no longer live; emit `recovered` instead.

Recommended integration: merge #4194 first, rebase this PR onto it, remove this PR’s duplicate 0107/domain/API/UI replay-state implementation, and keep the OMP adapter changes plus adapter tests.

backend/internal/storage/sqlite/store/conversation_store.go
nikhilachale: Fixed by the #4194 rebase in current head 710e1f254. The duplicate failed mapping is gone; ActivityStatusRecovered is inherited through domain, storage, API, and frontend, and open child activities settle as recovered.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed by the #4194 rebase in current head 710e1f254. The duplicate ErrChatHistoryUnsettled path is gone; a user-only replay tail closes as terminal recovered and conversation startup continues.

backend/internal/storage/sqlite/migrations/0107_recovered_conversation_turn_state.sql
nikhilachale: Fixed by removing this PR's duplicate migration during the #4194 rebase. Current head 710e1f254 inherits #4194's 0107 migration, including ORDER BY rowid for conversation turns and ORDER BY sequence for activities.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Resolved by rebasing onto merged #4194. Current head 710e1f254 inherits the shared regression coverage and recovered classification for both partial-provider and user-only replay tails; no replay tail is inferred completed from session/load delivery.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Superseded by the rebase onto merged #4194. Current head 710e1f254 carries #4194's shared ACP replay policy unchanged: replay tails are recovered/outcome-unknown, user-only replay remains reopenable, pending activities recover, and known durable terminal evidence is preserved. This PR no longer contains ACP replay/domain/migration/UI changes.

illegalcall (approved):
Approved after reviewing the rebased `710e1f254` head and the author’s updated replies. This is now a provider-only OMP change that inherits main’s canonical ACP replay/recovery implementation and migration unchanged. Native `omp acp`, the 15.0.0 version gate, model/system-prompt launch flags, process-fixed approval mapping, additive registry/STATUS entries, and public spawn/delegation enums are correctly implemented and covered. The effective merge is clean; focused tests, race tests, vet, `go build ./...`, generated-artifact checks, and CI pass.

Non-blocking cleanup: the three-part version parser duplicates the small Pi/Cursor parsing shape and can be consolidated when those adapters are next touched together.

PR #4197, opened 2026-08-20
fix: hide inactive cloud sign-in controls

AgentWrapper:
## Summary

- hide the nonfunctional AO Cloud sign-in row from the expanded sidebar
- hide the matching sign-in icon from the collapsed sidebar rail
- preserve account and sign-out controls for existing authenticated sessions
- add regression coverage for the configured-but-signed-out state

## Tests

- `npm test -- src/renderer/components/Sidebar.test.tsx` (60 passed)

## Risks / follow-ups

- The sign-in UI remains intentionally hidden until the cloud authentication flow is ready to expose again.

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -22,8 +22,15 @@ import type { WorkspaceSession, WorkspaceSummary } from "../types/workspace";
 import { agentsQueryKey } from "../hooks/useAgentsQuery";
 import { useUiStore } from "../stores/ui-store";
 
-const { getMock, navigateMock, mockParams, renameSessionMock, spawnMock, updateStatusMock, commandPaletteEnabled } = vi.hoisted(
+const { cloudSessionState, getMock, navigateMock, mockParams, renameSessionMock, spawnMock, updateStatusMock, commandPaletteEnabled } = vi.hoisted(
 	() => ({
+		cloudSessionState: {
+			configured: false,
+			session: null as null | { user: { email: string } },
+			status: "unauthenticated" as "authenticated" | "loading" | "unauthenticated",
+			signIn: vi.fn(),
+			signOut: vi.fn().mockResolvedValue(undefined),
+		},
 		getMock: vi.fn(),
 		navigateMock: vi.fn(),
 		mockParams: { projectId: undefined as string | undefined, sessionId: undefined as string | undefined },
@@ -36,6 +43,7 @@ const { getMock, navigateMock, mockParams, renameSessionMock, spawnMock, updateS
 
 vi.mock("../lib/rename-session", () => ({ renameSession: renameSessionMock }));
 vi.mock("../lib/spawn-orchestrator", () => ({ spawnOrchestrator: spawnMock }));
+vi.mock("../lib/cloud-session", () => ({ useCloudSession: () => cloudSessionState }));
 
 vi.mock("../hooks/useCommandPaletteEnabled", () => ({
 	useCommandPaletteEnabled: () => commandPaletteEnabled.current,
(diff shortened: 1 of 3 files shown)


frontend/src/renderer/components/Sidebar.test.tsx
```diff
@@ -273,6 +286,14 @@ afterEach(() => {
 });
 
 describe("Sidebar", () => {
+	it("does not show cloud sign-in controls while signed out", () => {
```
AgentWrapper, line 289: Nice that both footer variants are mounted at once, so this one assertion covers the expanded row and the collapsed rail.

What is not pinned is the half the PR promises to *preserve*: there is no test that sets `configured = true`, `status = "authenticated"`, `session = { user: { email: ... } }` and asserts the account row / sign-out control still render. The guard was rewritten from `status === "loading"` to `status !== "authenticated"`, so a future typo in that condition (or in `configured`) would silently hide the signed-in controls too and this suite would stay green. A short sibling test asserting `getAllByLabelText("Signed in as user@example.com")` would close that gap cheaply.

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -1051,38 +1050,13 @@ function SessionRow({
...
-						<span className="tracking-tight">{t("shell.signIn")}</span>
-					</button>
-				</TooltipTrigger>
-				<TooltipContent side="right" hidden={tabIndex !== -1}>
-					{t("shell.signInToAOCloud")}
-				</TooltipContent>
-			</Tooltip>
-		);
-	}
+	const { configured, session, status, signOut } = useCloudSession();
```
AgentWrapper, line 1058: With the unauthenticated branches gone, `signIn` from `useCloudSession` now has zero call sites anywhere in the renderer, and `shell.signIn` / `shell.signInToAOCloud` are orphaned in all eight locale catalogs. That reads as deliberate given the PR body says the flow comes back later, and the repo has no unused-key gate, so nothing fails today — just flagging it so the leftovers are a conscious park rather than a surprise for whoever re-enables the flow. If it is intentional, a one-line comment on `signIn` in `lib/cloud-session.ts` pointing at this hidden entry point would save the next reader a grep.

AgentWrapper (commented):
## Review: approving (no blocking changes)

The change does exactly what the description says: both `CloudAccountRow` and `CloudAccountRailButton` now render only for `configured && status === "authenticated"`, so the nonfunctional sign-in row and rail icon disappear while the authenticated account/sign-out affordances are untouched. The `LogIn` import is dropped and `User` is still used by both authenticated variants, so no dangling imports. `useCloudSession` has no other consumer in the renderer (`grep` finds only `Sidebar.tsx`), so nothing else regresses, and the new per-file `vi.mock("../lib/cloud-session")` only needs to supply `useCloudSession` because that is the only symbol the Sidebar tree imports. The guard rewrite from `status === "loading"` to `status !== "authenticated"` also keeps the loading state hidden, so there is no flash of chrome before the session resolves.

Two non-blocking notes below (test coverage for the path the PR promises to preserve, and the now-unreferenced sign-in strings/API). Neither needs to land before merge.

AgentWrapper (commented):
## Re-review at 581b4e3: approving (no blocking changes)

Both notes from the previous review are addressed, and nothing else moved: the only delta since 5ea8080 is the new authenticated-path test plus a one-line comment on `signIn`.

- `keeps cloud account controls visible while signed in` now pins the behavior the PR promises to preserve. Asserting `getAllByLabelText("Signed in as user@example.com")` has length 2 covers both mounted footer variants (expanded row + collapsed rail), so a regression in either guard — `configured`, or the `status !== "authenticated"` rewrite — fails the suite instead of silently hiding signed-in chrome. The session shape matches the hoisted mock type and the `session?.user.email` reads in both components.
- The `// Parked while the sidebar sign-in entry point is intentionally hidden.` comment on `signIn` makes the now-callerless helper a conscious park rather than a grep puzzle. `shell.signIn` / `shell.signInToAOCloud` stay in the locale catalogs, which is right for a flow that comes back, and no unused-key gate exists to fail on them.

The substance of the change is unchanged from the prior pass and still checks out: sign-in row and rail icon hidden, `loading` state still hidden so there is no flash before the session resolves, authenticated account/sign-out controls untouched, no dangling `LogIn` import, and `useCloudSession` has no consumer outside `Sidebar.tsx`. Ready to merge; I did not run tests or builds.

Issue #2589, opened 2026-07-10
[BUG]: npm run dev:web fails on Windows


Aakriti0207:
## Summary

For contributors who want to work on the renderer without launching the full Electron application, the project provides the `npm run dev:web` script which is especially useful when iterating on UI changes 
However, the script currently fails on Windows because it relies on POSIX-style environment variable syntax

## Actual behavior

The command fails with:

`'VITE_NO_ELECTRON' is not recognized as an internal or external command,
operable program or batch file.`

## Expected behavior

The renderer development server should start successfully on Windows, just as it does on Unix-like systems

## Root cause

The current script is:

`"dev:web": "VITE_NO_ELECTRON=1 vite --config vite.renderer.config.ts"`

Inline environment variable assignment is supported by POSIX shells but not by cmd.exe or PowerShell...

## Proposed fix

Use cross-env to make the script platform-independent:

`"dev:web": "cross-env VITE_NO_ELECTRON=1 vite --config vite.renderer.config.ts"`

This preserves the existing behavior on Unix-like systems while allowing the same command to work on Windows

> I'd be happy to open a PR for this if it sounds like the preferred approach

i-trytoohard:
From the NotesBot call on `2026-07-10`, this existing issue appears to be relevant.
Match confidence: `0.84` (fuzzy).

Relevant discussion excerpts:
- - Merge reviewed desktop UI PR into desktop app minor changes early branch; continue tokens + design system support for Anaya. - Await review on localization PR; fix Windows `npm run dev:web` cross-env issue; test whether similar issue exists in packaged `.exe` install. - Tag prateek on PRs for contributing.md + [...]
- - Continue design system Phase 2 (migration from CSS tokens into TS-based structure after Phase 1 merges); incorporate Suraj’s suggestions after merge. - Merge reviewed desktop UI PR into desktop app minor changes early branch; continue tokens + design system support for Anaya. - Await review on localization PR; [...]
- - Merge reviewed desktop UI PR into desktop app minor changes early branch; continue tokens + design system support for Anaya. - Await review on localization PR; fix Windows `npm run dev:web` cross-env issue; test whether similar issue exists in packaged `.exe` install.

Source thread: https://discord.com/channels/1476302178913357958/1525200588109905920/1525200588109905920

_Generated by AO Bot based on the transcript from NotesBot._

tamish560:
[...]

PR #4200, opened 2026-08-20
fix(frontend): make dev:web start on Windows

mahdiahmadi1991:
## What

Make `npm run dev:web` start on Windows, and add a test so the bug class cannot come back.

## Why

`dev:web` set the renderer's browser-preview flag with a POSIX env prefix:

```json
"dev:web": "VITE_NO_ELECTRON=1 vite --config vite.renderer.config.ts"
```

npm runs scripts through `cmd.exe` on Windows, which has no env-prefix syntax, so the script died before vite ever started:

```
> VITE_NO_ELECTRON=1 vite --config vite.renderer.config.ts --port 5199 --host 127.0.0.1
'VITE_NO_ELECTRON' is not recognized as an internal or external command,
operable program or batch file.
```

That also took `npm run test:e2e` down with it, since the Playwright `webServer` boots `dev:web`. A Windows contributor has no renderer-only dev loop and no renderer e2e suite today.

## How

Pass `--mode web` and let the renderer config inject the flag, so the value reaches `import.meta.env` on every platform:

```ts
define: mode === "web" ? { "import.meta.env.VITE_NO_ELECTRON": '"1"' } : {},
```

Only `web` defines the key. electron-forge dev/package and vitest run under other modes and leave it undefined exactly as before, so the packaged renderer is untouched and the comment in `useWorkspaceQuery.ts` still holds.

Two alternatives I discarded, in case they come up in review:

- **A tracked `.env` file** — `.gitignore` excludes `.env*` (only `.env.example` is negated), so it could not be committed.
- **`cross-env`** — I tried it and backed it out. Any dependency change rewrites `package-lock.json` on Windows, and that rewrite drops the optional `@emnapi/*` entries bundled under `@tailwindcss/oxide-wasm32-wasi`, which Linux installs need. The lockfile churn was worse than the fix. The config route needs no dependency and leaves the lockfile untouched.

The second commit adds `frontend/npm-scripts-portability.test.ts`, modelled on `release-workflows.test.ts`: it reads the root and frontend manifests and fails on any script that sets an environment variable with a POSIX prefix. Nothing else catches this class — the prefix is valid `sh` and CI is Linux-only.

## Testing

Windows 11, Node 24.13.0, npm 11.6.2, Go 1.27.0.

- `npm run dev:web -- --port 5199 --host 127.0.0.1` — starts (previously exit 1). Confirmed the flag actually lands by fetching a transformed module from the dev server:
  ```
  import.meta.env = {"BASE_URL":"/","DEV":true,"MODE":"web","PROD":false,"SSR":false,"VITE_NO_ELECTRON":"1"}
  ```
- `npm run typecheck` and `npm run typecheck:e2e` — both clean.
- `npx vitest run` — 2612 passed / 2625 total, exactly +2 from the new test versus the pre-change baseline of 2610 / 2623. The 12 remaining failures are the pre-existing macOS/POSIX-only specs (maker-dmg, relocation, auto-updater translocation, supervisor-link unix sockets, the `verify-mac-artifact.sh` exec-bit check) and are identical before and after this change.
- New test verified by reintroducing the prefix: it fails and names the offending script.
- `npm run dev` — full Electron app launches from this branch, daemon on 127.0.0.1:3002, renderer serving `/api/v1/projects` and `/api/v1/sessions` 200s.

I could not run the Playwright renderer suite end to end here (no browsers installed on this machine), so the `test:e2e` claim covers the `webServer` command starting, not the specs passing.

## Checklist

### DIFF
--- a/frontend/npm-scripts-portability.test.ts
+++ b/frontend/npm-scripts-portability.test.ts
@@ -1,0 +1,32 @@
+import { readFile } from "node:fs/promises";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+
+const repositoryRoot = path.resolve(import.meta.dirname, "..");
+
+// npm runs scripts through cmd.exe on Windows, which has no `VAR=value command`
+// syntax: such a script dies with "'VAR' is not recognized as an internal or
+// external command" before the tool it prefixes ever starts. The prefix works on
+// macOS and Linux and CI runs Linux only, so nothing else catches it — `dev:web`
+// carried one for months and took `npm run test:e2e` down with it on Windows.
+//
+// Pass the value through the tool's own interface instead (vite reads `--mode`,
+// and the renderer config turns that into the define), or read it from a config
+// file. Matching only at a command boundary keeps flag values such as
+// `--path-mode=abs` and `-o src/api/schema.ts` out of the net.
+const POSIX_ENV_PREFIX = /(?:^|&&\s*|\|\|\s*|;\s*|\|\s*)[A-Za-z_][A-Za-z0-9_]*=/;
+
+describe("npm scripts stay cross-platform", () => {
+	for (const manifest of ["package.json", "frontend/package.json"]) {
+		it(`${manifest} sets no environment variable with a POSIX prefix`, async () => {
+			const contents = await readFile(path.join(repositoryRoot, manifest), "utf8");
+			const scripts: Record<string, string> = JSON.parse(contents).scripts ?? {};
+
+			const offenders = Object.entries(scripts)
+				.filter(([, command]) => POSIX_ENV_PREFIX.test(command))
+				.map(([name, command]) => `${name}: ${command}`);
+
+			expect(offenders).toEqual([]);
+		});
+	}
+});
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@mahdiahmadi1991 this looks like it addresses #2589 — linking for tracking. Adding `Fixes #2589` to the PR body will auto-close it on merge.

i-trytoohard:
@mahdiahmadi1991 Marking as stale: this has been open 20 days without a resolution. Push or comment to reset it; at 30 days old it gets closed as stale. Branch stays safe to reopen.

i-trytoohard:
@mahdiahmadi1991 Marking as stale: no activity for 23 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@mahdiahmadi1991 Marking as stale: no activity for 23 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

codebanditssss:
Thanks for the contribution. Closing this in favor of #5770, which addresses the same Windows `dev:web` failure while preserving Vite’s existing `development` mode. This implementation changes the mode to `web`, which can alter mode-dependent env loading; #5770 instead uses a small cross-platform Node launcher and tests the actual launch path.

PR #4201, opened 2026-08-20
feat: add Cursor ACP chat driver

nikhilachale:
## Summary

- add a native Cursor Chat driver using `cursor-agent acp`
- reuse the existing Cursor adapter for binary resolution, authentication, and runtime environment isolation
- negotiate resume through Cursor’s advertised `session/load` capability
- expose only provider-advertised model/mode options and slash commands
- keep Cursor Chat-only until ACP and TUI conversation identity is proven shared

## Verification

- `go test ./internal/adapters/chatdriver/cursoracp ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/nativeacp ./internal/adapters/chatdriver/registry`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs ./internal/adapters/chatdriver/...` — 0 issues
- `AO_LIVE_CURSOR_ACP=1 go test ./internal/adapters/chatdriver/cursoracp -run TestLiveCursorACP -v -count=1` — passed against Cursor `2026.08.11-e8db854`

The live test covers coding tools, native approvals, cancellation, `session/load` resume, provider options/commands, and `AGENTS.md` project rules.

## Notes

The full repository lint/test command was attempted under a heavily loaded local machine and hit unrelated timing/environment-sensitive failures in existing adapter, tmux, CLI, and worktree tests. The directly touched packages pass, scoped lint is clean, and the initially failing adapter packages passed when rerun in isolation.


nikhilachale:
Screenshot 2026-08-21 at 05 15 23

backend/internal/adapters/chatdriver/cursoracp/driver.go
```diff
@@ -0,0 +1,36 @@
...
+	return nativeacp.New(plugin, nativeacp.Config{
+		Harness:        domain.HarnessCursor,
+		Configure:      configure,
+		SessionOptions: sessionOptions,
+		VersionProbe:   versionProbe,
+	}, log)
+}
+
+func configure(acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	return []string{"acp"}, nil, nil
```
illegalcall: [P1] `configure` discards `LaunchConfig.SystemPrompt` and `LaunchConfig.Permissions`; `sessionOptions` only maps model. The live test proves ordinary project `AGENTS.md`, but never supplies or verifies AO's generated orchestrator/worker standing prompt. The approval picker therefore also advertises `accept-edits`, `auto`, and `bypass-permissions` choices that have no effect—the generic ACP client continues parking every Cursor permission request.

Recommended fix: inject the standing prompt through a Cursor-supported managed project rule/session mechanism and verify it on both start and load-resume. For permissions, either implement a proven client-side policy for Cursor's ACP permission options (including explicit bypass) or reject unsupported modes instead of silently accepting them. Add contract/live tests for every advertised mode.

backend/internal/adapters/chatdriver/cursoracp/driver.go
```diff
@@ -0,0 +1,36 @@
...
+	acpdriver "github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/acp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/nativeacp"
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// New launches `cursor-agent acp` from the exact binary resolved by the
+// existing Cursor adapter. Models and slash commands are populated only from
+// the live ACP session advertisements; Cursor owns tools, project rules, auth,
+// and permission requests inside its native agent process.
```
illegalcall: [P1] Cursor's ACP surface includes blocking `cursor/ask_question` and `cursor/create_plan` requests. Cursor explicitly says the agent waits for the client response: [official Cursor ACP docs](https://cursor.com/docs/cli/acp). AO's generic ACP client handles standard permissions/elicitation but does not dispatch these Cursor methods, and this PR adds no provider-specific handler. A prompt that reaches either path can fail with method-not-found or stall instead of producing an AO input/plan request.

Recommended fix: add a Cursor extension dispatcher that maps questions to AO's durable `ChatInputRequest` flow and plan approval to the appropriate plan/input flow, including cancellation/timeouts. If the pinned Go ACP transport cannot dispatch Cursor's nonstandard method names, extend or upgrade that transport explicitly. Add a fake-wire test for both blocking methods and a live test that actually triggers one.

backend/internal/adapters/chatdriver/cursoracp/version.go
```diff
@@ -0,0 +1,80 @@
...
+	// Cursor's standard installer makes ~/.local/bin/cursor-agent a symlink into
+	// a versioned directory. Prefer that local fact: invoking --version starts
+	// the full Node CLI and can wait on another live Cursor process long enough
+	// to exceed a health probe's deadline.
+	if resolved, err := filepath.EvalSymlinks(bin); err == nil {
+		if installed, ok := parseCursorVersion(resolved); ok {
+			return requireMinimumCursorVersion(installed)
+		}
+	}
+	output, err := exec.CommandContext(ctx, bin, "--version").CombinedOutput()
```
illegalcall: Use the repository's `internal/process.CommandContext` wrapper here rather than `os/exec.CommandContext`. The wrapper is the established subprocess boundary and preserves hidden-console behavior on Windows. Add the import and keep the existing timeout/version parsing unchanged.

illegalcall (commented):
ACP/provider review against the current PR head. Treating `loadSession` as an independent restore capability is correct ACP usage, but this branch now conflicts with current `main` in the generic ACP driver/tests. Rebase and preserve current-main history/error handling while retaining the `load OR resume` capability gate; do not resolve by taking either generic driver file wholesale.

Shared ACP prerequisite: a completed `session/load` response proves replay delivery, not that the last native turn is idle. Source: [ACP session/load semantics](https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). The generic tail-state classification must be corrected before load-based Cursor resume is considered lifecycle-safe.

backend/internal/adapters/chatdriver/cursoracp/driver.go
nikhilachale: Fixed in 3eda6b0dd. Added a provider extension dispatcher above the pinned ACP transport for cursor/ask_question and cursor/create_plan. Questions use AO durable ChatInputRequest handling; plans publish the native plan and park on AO durable approval. Both paths honor request cancellation and the existing timeout. Raw JSON-RPC wire tests cover both methods and cancellation, and the gated live test now prompts Cursor to invoke ask_question.

backend/internal/adapters/chatdriver/cursoracp/driver.go
nikhilachale: Fixed in 3eda6b0dd. Cursor ACP now writes a session-scoped managed plugin rule carrying the AO standing prompt and recomputes it on load-resume. Permission modes are no longer silently ignored: auto and bypass map to Cursor launch flags, accept-edits and bypass use validated ACP option policies, and default/escalated requests remain durable AO approvals. Contract coverage exercises every advertised mode plus start/resume rule replacement.

backend/internal/adapters/chatdriver/cursoracp/version.go
nikhilachale: Fixed in 3eda6b0dd. versionProbe now uses internal/process.CommandContext while retaining the existing context, arguments, CombinedOutput handling, and version parsing.

nikhilachale:
Review fixes are pushed in 3eda6b0dd after rebuilding the branch on current main (35ee1fe42). The generic ACP files were reconciled manually: current-main history/error behavior is preserved, the load-or-resume capability gate remains, and a session/load replay tail is now classified interrupted unless a following user boundary proves completion.\n\nAlso included: managed Cursor standing rules on start/resume, explicit policies for every advertised permission mode, durable dispatch for cursor/ask_question and cursor/create_plan with cancellation/timeouts, and the repository process wrapper for version probing.\n\nVerification:\n- go build ./...: pass\n- affected package tests: pass\n- affected ACP package race tests: pass\n- affected ACP package go vet: pass\n- raw JSON-RPC fake-wire tests cover both blocking Cursor methods and cancellation\n- independent code review: approve, no Critical/Important findings\n\nThe gated live test reached Cursor Agent successfully, but this local Cursor profile returned Upgrade your plan to continue before executing the prompt, so I could not honestly mark the account-dependent live assertions as passed.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -69,15 +68,15 @@ func (c *conversation) abortHistoryReplay() {
...
-	if interrupted {
-		// session/load has finished, so this user-only turn cannot still be
-		// producing output. It represents a prompt interrupted before the agent
-		// emitted a provider event and remains valid replayable history.
+	if hasUnboundedTail {
+		// session/load only proves that replay delivery finished. ACP does not
+		// report whether the last native turn was idle, so only a following user
+		// boundary can prove a replayed turn completed. Preserve the unbounded tail
+		// as interrupted even when it already contains provider output.
 		turnState = domain.TurnStateInterrupted
```
illegalcall: [P1] This now makes every ordinary latest turn look interrupted after restart. A normal transcript also has no following user boundary, so even a fully answered final turn takes this branch. ACP exposes no historic terminal outcome in `session/load` (https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188). Preserve known AO/provider outcome evidence or introduce a terminal recovered/unknown state; absence of a later prompt is not evidence of interruption.

backend/internal/adapters/chatdriver/cursoracp/driver.go
```diff
@@ -0,0 +1,121 @@
...
+		ClientExtension:  handleExtension,
+		VersionProbe:     versionProbe,
+	}, log)
+}
+
+func configure(cfg acpdriver.LaunchConfig) ([]string, map[string]string, error) {
+	args := make([]string, 0, 4)
+	switch ports.NormalizePermissionMode(cfg.Permissions) {
+	case ports.PermissionModeAuto:
+		args = append(args, "--auto-review")
```
illegalcall, line 68: [P1] `--auto-review` only affects process launch, but AO permits approval mode in per-turn settings. If a live conversation changes to `auto`, no process restarts and `permissionPolicy` has no Auto branch, so the requested mode is silently ineffective. Either implement Auto as an ACP permission-response policy with documented semantics, or validate and reject any turn-time change that cannot be applied until Chat restarts.

illegalcall (changes requested):
Re-review of the updated Cursor head: the standing rule, blocking ask_question/create_plan extension handling, process wrapper, permission round-trips, and load-only restore gate are fixed. I still cannot approve: this head conflicts with current main in acp/client.go; the resolution must preserve main’s decision Kind, structured Detail, and approvalToolDetail. The copied full SDK dispatcher also creates an ongoing drift surface; please isolate the two Cursor extensions behind a narrow extension hook (or an upstream SDK hook) instead of owning the standard dispatcher. Pass context through writeStandingRule as required by AGENTS.md.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in 74deffde3. Replay now classifies only a user-only unbounded tail as interrupted. A tail containing provider output uses completed as the synthetic fallback, while the current-main reconciliation layer preserves any known AO interrupted/failed terminal outcome. The regression suite covers both answered and user-only tails.

backend/internal/adapters/chatdriver/cursoracp/driver.go
nikhilachale: Fixed in 74deffde3. Cursor Auto and bypass are treated as process-scoped: runtime transitions to or from either mode return an explicit restart-required error. Empty per-turn settings preserve the launch mode, while default/accept-edits remain dynamically selectable. Permission policy changes are committed only after all turn settings succeed, with regression coverage for failed config RPCs.

nikhilachale:
Addressed the re-review in 74deffde3:

- rebased onto current `main` and resolved `acp/client.go` while preserving decision `Kind`, structured `Detail`, and `approvalToolDetail`;
- removed the copied standard SDK dispatcher and replaced it with a bounded, narrow alias reader for only `cursor/ask_question` and `cursor/create_plan`; the SDK again owns standard dispatch and cancellation;
- threaded `context.Context` through native ACP `Configure` callbacks and `writeStandingRule`;
- corrected replay tail classification and rejected unsupported runtime Auto/bypass transitions.

Added wire, history, permission-transition/failure, framing-limit, and cancellation regressions. Focused tests and race tests pass, `go build ./...` passes, and golangci-lint v2.12.2 reports 0 issues. The full backend run passed all touched packages; three unrelated timing-sensitive tests failed under parallel load and passed immediately in isolation.

backend/internal/adapters/chatdriver/acp/history.go
```diff
@@ -75,9 +75,10 @@ func (c *conversation) finishHistoryReplay() {
 
 	turnState := domain.TurnStateCompleted
```
illegalcall: [P1] Provider output is not proof that the historical turn completed. If the agent persisted a partial response and then crashed, `session/load` replays that partial chunk and this fallback records success. Reconciliation only helps when AO already has a trustworthy terminal result; it cannot fix provider-native/imported history or a daemon-killed running turn. Model the final replay tail as terminal outcome-unknown/recovered in shared ACP core (or use explicit provider outcome evidence), then preserve any known AO terminal state during reconciliation. Do not infer `Completed` merely from the presence of output.

illegalcall (changes requested):
Latest-head re-review: the branch now merges cleanly; current-main approval Kind/Detail are preserved; context propagation and launch-only Auto/bypass validation are fixed; the full copied SDK dispatcher has been replaced by a narrow alias reader that leaves standard ACP dispatch in the SDK. The remaining blocker is the generic replay outcome below; land a shared ACP-core policy and rebase this adapter onto it.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Fixed in 0271bc52b. Shared ACP replay now records the final unbounded session/load tail as recovered regardless of whether it contains user text, provider output, reasoning, or tool work. Reconciliation preserves an existing AO completed/interrupted/failed/recovered terminal outcome, while queued/running and provider-only imports remain recovered. Open replay-tail activities stay running through normalization and are cancelled by durable recovered-turn settlement, so partial reasoning/tools no longer acquire synthetic success or failure. Added ACP replay, reconciliation (including queued), SQLite migration/storage, service import, API, and neutral UI regressions.

docs/superpowers/specs/2026-08-23-acp-recovered-turn-state-design.md
```diff
@@ -0,0 +1,102 @@
...
+
+Expose the state through the existing conversation API enum and generated
+frontend schema. The Chat timeline renders the outcome as `Recovered` with a
+neutral tone. Existing clients that treat all non-running/non-queued states as
+terminal retain their current behavior.
+
+## ACP replay policy
+
+The generic ACP history core applies one policy for every ACP provider:
+
```
illegalcall: [P1] A later prompt proves only that the preceding prompt call is no longer active; it does not recover that call’s stop reason. The preceding turn may have ended through cancellation, refusal/error, or another non-success outcome, and `session/load` does not replay the old `session/prompt` response. Mark every reconstructed replay boundary `recovered`, then let AO’s durable explicit terminal evidence upgrade it, as #4194 does. Update `history.go` and the test that currently requires bounded turns to be `completed`.

backend/internal/storage/sqlite/store/conversation_store.go
```diff
@@ -744,13 +744,14 @@ func (s *Store) SettleTurn(
...
+// terminalActivityStatus mirrors why the enclosing turn ended. Interrupted and
+// recovered work are neither successful nor failed: one was deliberately stopped,
+// while the other's historical outcome is unavailable.
 func terminalActivityStatus(state domain.TurnState) domain.ActivityStatus {
 	switch state {
 	case domain.TurnStateCompleted:
 		return domain.ActivityStatusCompleted
-	case domain.TurnStateInterrupted:
+	case domain.TurnStateInterrupted, domain.TurnStateRecovered:
 		return domain.ActivityStatusCancelled
```
illegalcall, line 756: [P1] `cancelled` invents the outcome this new state is meant to avoid. This PR’s own design says recovered is not evidence of user cancellation, while the existing domain/UI semantics render cancelled work as stopped. Add/use `ActivityStatusRecovered` through domain, SQLite constraint, DTO/OpenAPI/frontend, and settle unknown replay activities to that value. #4194 already contains that shared implementation.

backend/internal/storage/sqlite/migrations/0107_recovered_conversation_turn.sql
```diff
@@ -0,0 +1,91 @@
...
+    id, conversation_id, handled_by_session_id, provider_turn_id,
+    controller_generation, state, error_message, requested_at, started_at,
+    completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+    promotion_started_at, promoted_to_turn_id
+)
+SELECT id, conversation_id, handled_by_session_id, provider_turn_id,
+       controller_generation, state, error_message, requested_at, started_at,
+       completed_at, diff_json, rolled_back_at, plan_json, branch_id,
+       promotion_started_at, promoted_to_turn_id
+FROM conversation_turns;
```
illegalcall: [P1] Preserve row order during the table rebuild. AO rollback queries use `rowid` to order turns recorded at the same timestamp, so a bare `SELECT` may reassign those rowids and change rollback membership. Add `ORDER BY rowid`, and add a same-timestamp ordering regression—or preferably drop this duplicate 0107 after rebasing onto #4194’s migration, which already preserves it.

illegalcall (changes requested):
Latest-head re-review: the previous final-tail bug is fixed. User-only, partial-output, reasoning, and tool tails are now terminal `recovered`; reconciliation preserves stronger AO terminal facts; Cursor’s standing instructions, load restore, extension dispatch, and process-scoped permission behavior remain correct. Focused backend tests, race tests, vet, and frontend typecheck pass locally on the effective merge.

I still cannot approve because three shared replay-state issues remain inline. ACP `session/load` replays conversation entries, not historic `session/prompt` responses or stop reasons, so neither a later user boundary nor an open activity carries a completed/cancelled outcome: https://github.com/agentclientprotocol/agent-client-protocol/blob/39855be810d318b70e879c9cfa87cff42e2e27b2/docs/protocol/v1/session-setup.mdx#L134-L188

Recommended integration: merge #4194 first, rebase this PR onto its canonical replay policy and migration, delete this branch’s duplicate ACP/domain/storage/API/UI implementation, and retain the Cursor-specific adapter work/tests. The two checked-in `docs/superpowers` implementation documents are also ~400 lines of unmentioned PR scope; remove them or reduce durable rationale to a concise ADR.

illegalcall (changes requested):
Current-head re-review after `a1258e173`: the new commit only checks migration query/row errors; it does not address the three replay-state findings in the preceding review. #4194 has now merged the canonical ACP recovered turn/activity policy into `main`, and this PR is correctly reported as conflicting. Rebase onto current `main`, remove this branch’s duplicate ACP/domain/storage/API/UI implementation and migration 0107, then retain the Cursor-specific adapter and tests. During the rebase, use the merged policy: every replayed boundary is `recovered` unless AO has a stronger durable terminal outcome; unresolved activities are `ActivityStatusRecovered`, not cancelled; and the shared migration preserves rowid ordering.

backend/internal/adapters/chatdriver/acp/history.go
nikhilachale: Final update: rebased at dd8f791c9 onto main 2a7579f39, which now contains the shared ACP recovered turn/activity policy. I dropped this branch’s duplicate recovery migration, reconciliation, API, and UI implementation and aligned the Cursor replay regression with the shared core. The remaining PR diff is adapter/extension work only; it preserves main’s recovered activity handling and Pi registration. PR is mergeable and all eight current checks pass.

illegalcall (approved):
Approved after reviewing the force-rebased `dd8f791c9` head and the author’s final reply. The branch now cleanly inherits main’s canonical ACP recovered-turn/activity policy and ordered migration; the duplicate recovery domain/storage/API/UI implementation is gone. The previous bounded-turn, user-only/partial-tail, cancelled-activity, and migration-collision blockers are resolved. Cursor-specific standing instructions, permission-mode validation, load restore, ask-question/create-plan extension dispatch, version gating, and provider environment isolation remain covered. Focused tests, race tests, vet, `go build ./...`, diff checks, and all current CI checks pass.

Non-blocking cleanup: remove the now-unused `turnHasProvider` field/write sites from `acp/history.go`, and consider renaming `client_connection_test.go` to `extension_reader_test.go` so its subject is obvious.

PR #4202, opened 2026-08-20
fix: rank agents by usage frequency

nikhilachale:
## What

Rank agent choices by retained-session usage instead of relying primarily on a static preference list and alphabetical ordering.

The agent catalog now reports `usageCount` and `lastUsedAt`, and agent pickers order selectable agents by frequency, then recency. The existing static priority and alphabetical order remain as cold-start tie-breakers. The create-project default follows the same ranking.

## Why

Frequently used agents should be quicker to select and should become the default naturally, while new installations still need deterministic ordering before any usage history exists.

## How

- Derive usage at agent-catalog read time from durable session records; no new stored display state or migration.
- Count retained sessions by their current harness and use the newest retained session creation time as the recency tie-breaker.
- Preserve availability/auth grouping ahead of usage so an unavailable agent never outranks a selectable one.
- Regenerate OpenAPI and frontend TypeScript types for the new metadata.

Intentional scope: usage reflects retained AO sessions currently attributed to each harness. Deleted sessions and historical source harnesses after an agent switch are not counted.

## Testing

- `go test ./internal/service/agent ./internal/daemon`
- `go test ./internal/service/agent ./internal/httpd/apispec/...`
- `npm --prefix frontend test -- --run src/renderer/lib/agent-select-options.test.ts src/renderer/components/CreateProjectAgentSheet.test.tsx` (12 tests passed)
- `git diff --check`

A full `go test ./...` was also attempted in the restricted workspace; unrelated socket/tmux tests could not bind local sockets and a few timing-sensitive adapter tests timed out. All touched-package tests passed.

## Checklist

### DIFF
--- a/backend/internal/service/agent/service.go
+++ b/backend/internal/service/agent/service.go
@@ -66,6 +66,8 @@ type Info struct {
 	ID         string                `json:"id"`
 	Label      string                `json:"label"`
 	AuthStatus ports.AgentAuthStatus `json:"authStatus,omitempty" enum:"authorized,unauthorized,unknown" description:"Advisory local auth probe result. authorized means a recent local probe passed; spawn remains the authoritative validation point."`
+	UsageCount int                   `json:"usageCount,omitempty" description:"Number of retained sessions currently attributed to this agent."`
+	LastUsedAt *time.Time            `json:"lastUsedAt,omitempty" format:"date-time" description:"Creation time of the newest retained session currently attributed to this agent."`
 }
 
 // Inventory describes all daemon-supported agents and best-effort local probe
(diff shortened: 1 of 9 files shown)


nikhilachale:
Screenshot 2026-08-21 at 02 24 31

backend/internal/service/agent/service.go
```diff
@@ -207,7 +220,49 @@ func (s *Service) Refresh(ctx context.Context) (Inventory, error) {
...
+type sessionUsage struct {
+	count      int
+	lastUsedAt time.Time
+}
+
+func (s *Service) withSessionUsage(ctx context.Context, inventory Inventory) (Inventory, error) {
+	if s.sessions == nil {
+		return inventory, nil
+	}
+	records, err := s.sessions.ListAllSessions(ctx)
```
neversettle17-101, line 235: nit: we can ideally cache the sessions map with its agent usage and update it whenever a new session spawns up. this will avoid the call to fetch all sessions during the /agents call. Non blocking

neversettle17-101 (approved):

PR #4204, opened 2026-08-20
fix(lifecycle): fence chat activity by controller generation

### LINKED ISSUE #4125: bug(lifecycle): nested claude process inheriting AO_SESSION_ID rewrites the parent session's identity and marks it exited; live chat controller cannot clear the sticky 'exited'
## Bug

Two related defects, observed live on a chat-mode worker session (`agent-orchestrator-143`, 2026-08-19 03:31Z) when a diagnostic script spawned a short-lived Claude process from the session's shell:

**1. Hook attribution is env-only, and chat-mode sessions have no launch-id guard.**
`ao hooks` attributes signals purely by the inherited `AO_SESSION_ID` env var. Any nested `claude` process started from a session's shell — a diagnostic script, a user manually running `claude` in the workspace terminal, an SDK-spawned child — inherits that env, and if its cwd has AO hooks in `.claude/settings.local.json`, its `SessionStart`/`SessionEnd` hooks report **its** `session_id`/`transcript_path` into the **parent** session's row: `agent_session_id`, `native_transcript_path`, and `latest_*` get overwritten, and on quit the parent session is marked `exited` while its real agent keeps running.

The lifecycle launch-id gate (`backend/internal/lifecycle/manager.go:506`) only drops mismatched signals when the session has a non-empty `runtime_launch_id`. Chat-mode sessions have none, so every foreign signal passes. Observed impact: the session showed `exited` in the UI while actively working, and `agent_session_id` pointed at a *different session's* native conversation — a later TUI restore would have resumed the wrong conversation.

**2. A live chat controller cannot clear a sticky `exited`.**
Once `activity_state=exited`, only a signal with `Event == "user-prompt-submit"` can resurrect the session (`backend/internal/lifecycle/manager.go:524-529`). The chat controller's ongoing activity signals do not carry that event, so a chat session that is demonstrably alive and streaming stays rendered `exited` indefinitely. A live controller generation reporting activity should count as resurrection evidence.

**Analyzed against:** `4a23d98ca` | **Confidence:** High (reproduced + state verified in the sessions table; repaired by replaying a correct `user-prompt-submit` hook payload)

## Suggested fixes

- Gate hook-signal identity updates for chat-mode sessions on the controller generation (or drop `agent_session_id` overwrites whose value disagrees with `provider_conversation_id` while mode is chat), and/or scope hook attribution to the workspace the hook fired from.
- Allow chat-controller signals (matching the current `controller_generation`) to clear `exited`.

## Related
- [#4120](https://github.com/Untrivial-ai/agent-orchestrator/issues/4120) — discovered while triaging this incident
- [#2604](https://github.com/Untrivial-ai/agent-orchestrator/issues/2604) — adjacent stale-activity rendering after daemon restart

illegalcall:
Fixes #4125

## Summary
- require every Chat lifecycle signal to carry the current internal controller generation, so inherited provider hooks cannot rewrite the parent session identity or activity
- treat non-exited activity from that current controller as liveness evidence, allowing a live Chat session to recover from a stale `exited`
- add regressions for foreign `session-end` metadata/state corruption and active, idle, and waiting-input resurrection

## Root cause
The reducer only validated `ControllerGeneration` when the incoming signal supplied one. Public provider hooks never supply it, and Chat sessions have no runtime launch ID, so those signals bypassed both ownership checks. The SQLite write fence could not compensate because it compared against the generation copied from the already-read session row.

## Tests
- `go test ./internal/lifecycle -run 'TestActivitySignal(RejectsHookWithoutChatControllerGeneration|CurrentChatControllerResurrectsExitedSession)$' -count=10`
- `go test ./internal/lifecycle ./internal/service/chat ./internal/httpd/controllers ./internal/integration`
- `env -u AO_ISSUE_ID -u AO_PROJECT_ID -u AO_SESSION_ID -u AO_OWNER go test -p 1 ./...`
- `go test -race -count=1 ./internal/lifecycle ./internal/service/chat`
- `golangci-lint v2.12.2` with an isolated cache: 0 issues

## Risk
TUI hook semantics are unchanged. Chat controller generations are already claimed before event consumption and included on every internal activity report; no API or schema change is required.

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -547,9 +547,16 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID, 
 		m.mu.Unlock()
 		return nil
 	}
-	if s.ControllerGeneration != "" &&
-		(domain.NormalizeSessionMode(rec.Mode) != domain.SessionModeChat ||
-			s.ControllerGeneration != rec.Metadata.ControllerGeneration) {
+	mode := domain.NormalizeSessionMode(rec.Mode)
+	currentChatController := mode == domain.SessionModeChat &&
+		s.ControllerGeneration != "" &&
+		s.ControllerGeneration == rec.Metadata.ControllerGeneration
+	// Chat has no supervised runtime launch to fence inherited workspace hooks.
+	// Its in-process controller always supplies the generation it durably claimed,
+	// while provider hooks never receive that internal credential. Requiring it in
+	// Chat mode therefore rejects both stale controllers and nested agent processes.
+	if (mode == domain.SessionModeChat && !currentChatController) ||
+		(mode != domain.SessionModeChat && s.ControllerGeneration != "") {
 		m.mu.Unlock()
 		return nil
 	}
@@ -558,11 +565,13 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID, 
 		m.mu.Unlock()
 		return nil
 	}
-	// An explicit prompt submission is proof that an agent was relaunched in the
-	// preserved shell. Other same-generation callbacks may have been delayed
-	// behind the process-exit report and cannot resurrect an exited workload.
+	// An explicit prompt submission is proof that a TUI agent was relaunched in
+	// the preserved shell. Other TUI callbacks may have been delayed behind the
+	// process-exit report and cannot resurrect an exited workload. Chat activity
+	// is emitted synchronously by its current generation-fenced controller, so a
+	// non-exited signal from that owner is direct evidence that it is still live.
 	if rec.Activity.State == domain.ActivityExited && s.Valid && s.State != domain.ActivityExited &&
-		(s.State != domain.ActivityActive || s.Event != "user-prompt-submit") {
+		!currentChatController && (s.State != domain.ActivityActive || s.Event != "user-prompt-submit") {
 		m.mu.Unlock()
 		return nil
 	}
--- a/backend/internal/lifecycle/manager_test.go
+++ b/backend/internal/lifecycle/manager_test.go
@@ -3845,3 +3845,79 @@ func TestEmitTelemetryStampsRequestID(t *testing.T) {
 		})
 	}
 }
+
+func TestActivitySignalRejectsHookWithoutChatControllerGeneration(t *testing.T) {
+	ctx := context.Background()
+	st := newFakeStore()
+	rec := domain.SessionRecord{
+		ID: "mer-1", ProjectID: "mer", Mode: domain.SessionModeChat,
+		Metadata: domain.SessionMetadata{
+			AgentSessionID:         "parent-native-session",
+			NativeTranscriptPath:   "/transcripts/parent.jsonl",
+			ProviderConversationID: "parent-native-session",
+			ControllerGeneration:   "chat-generation-1",
+		},
+		Activity: domain.Activity{State: domain.ActivityActive},
+	}
+	st.sessions[rec.ID] = rec
+	m := New(st, nil)
+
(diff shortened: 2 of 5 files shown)


backend/internal/lifecycle/manager.go
```diff
@@ -522,11 +529,13 @@ func (m *Manager) ApplyActivitySignal(ctx context.Context, id domain.SessionID,
...
-	// preserved shell. Other same-generation callbacks may have been delayed
-	// behind the process-exit report and cannot resurrect an exited workload.
+	// An explicit prompt submission is proof that a TUI agent was relaunched in
+	// the preserved shell. Other TUI callbacks may have been delayed behind the
+	// process-exit report and cannot resurrect an exited workload. Chat activity
+	// is emitted synchronously by its current generation-fenced controller, so a
+	// non-exited signal from that owner is direct evidence that it is still live.
 	if rec.Activity.State == domain.ActivityExited && s.Valid && s.State != domain.ActivityExited &&
-		(s.State != domain.ActivityActive || s.Event != "user-prompt-submit") {
+		!currentChatController && (s.State != domain.ActivityActive || s.Event != "user-prompt-submit") {
```
illegalcall, line 574: **[P2] Preserve exited ordering against concurrent controller work**

A matching generation proves ownership, not that the controller is still live. `Controller.project()` keeps this generation in the session after its event stream ends and writes `chat.controller.stopped` / `exited` without `sendMu`, while a concurrent `Interrupt` can later finish `reconcileDurableTurnsLocked` and emit `chat.interrupt.reconciled` / `idle` with the same generation. This exception then accepts that late idle signal and leaves a stopped, deregistered controller displayed as live indefinitely. Please serialize or invalidate the controller epoch on stop, or reject signals older than the exit, and add a shutdown-vs-interrupt regression test.

illegalcall (commented):

## Standards
No documented-standard violations or meaningful baseline smells found. The patch is surgical and keeps lifecycle writes behind `lifecycle.Manager`.

## Spec
The controller-generation fence addresses inherited-hook metadata and state corruption, and the new tests cover active, idle, and waiting-input resurrection. One correctness gap remains: a matching retained generation proves ownership, not liveness; see the inline finding.

**Summary:** Standards: 0 findings. Spec/correctness: 1 finding; a shutdown/interrupt race can leave a dead Chat controller displayed as idle.

backend/internal/lifecycle/manager.go
illegalcall: Addressed in f5c784854. Stream termination and explicit stopped-state projection now serialize on the controller send lock, so exited is ordered after controller work that already won the lock. Interrupt also rejects a stopped controller both at entry and after an in-flight provider refusal returns, preventing stale reconciliation from restoring ready/idle. Added TestStreamClosureWinsAgainstConcurrentInterruptReconciliation, which deterministically held the provider interrupt response open, closed and drained the stream, then released the stale refusal; it failed before the fix with ready plus chat.interrupt.reconciled after exited and now passes. Full backend tests, the complete Chat race suite, lint, and all eight PR checks pass.

illegalcall (commented):
**Verdict: approved.**

## Standards
No documented-standard violations or meaningful baseline smells found. The follow-up preserves the `sendMu` then `mu` lock order, keeps lifecycle writes behind `lifecycle.Manager`, and adds a focused boundary regression.

## Spec
Issue #4125 is now satisfied. Generationless and stale Chat hooks are fenced, while activity from the current live controller can recover a stale `exited`. The prior shutdown/Interrupt race is resolved by serializing terminal shutdown on `sendMu` and rechecking `Stopped` before refusal reconciliation; the new deterministic regression covers that ordering.

**Summary:** Standards: 0 findings. Spec: 0 findings. Ready to merge. Static review only; no tests or builds were run.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #2848, opened 2026-07-19
feat(config): per-role worker environment profile (claude-code) (#2195)

axisrow:
## Summary

Widen the per-role `RoleOverride` beyond `{agent, model, permissions}` into a full **worker environment profile**: `SystemPrompt`, `Env`, `MCP`, `PluginDirs` — scoped per role (worker/orchestrator), surfaced via CLI flags and the API. claude-code ships a CLI flag for each (`--append-system-prompt`, `--mcp-config`, `--strict-mcp-config`, `--plugin-dir` / `--plugin-url`).

Closes #2195.

## What it adds

- **domain** (`agentconfig.go`): `AgentConfig` gains `SystemPrompt`, `Env`, `MCP *MCPConfig`, `PluginDirs`; `MCPConfig{Configs, Strict}` (Strict alone = isolation, valid); `IsZero` via `reflect.DeepEqual` (map/slice/ptr fields); `Validate` rejects empty MCP entries.
- **session manager**: `effectiveAgentConfig` merges the new fields — Env per-key via deep-copying `mergeEnv` (a regression test pins that it never mutates the project config), and the MCP pointer / PluginDirs slice are copied rather than aliased (same defense-in-depth). `buildSystemPrompt` appends the per-role prompt.
- **adapter claudecode**: `appendMCPFlags` / `appendPluginFlags` emit in **both** `GetLaunchCommand` and `GetRestoreCommand` (resume rebuilds from flags); both paths call `cfg.Config.Validate()` for symmetry. nil/empty → inherit global, unchanged.
- **CLI**: `--worker-mcp-config` (repeatable), `--worker-strict-mcp`, `--worker-plugin-dir` (repeatable); full surface also via `--config-json`.
- **API regen**: `openapi.yaml` + `frontend/src/api/schema.ts`.

## Review-feedback infra fixes

Carried over from @illegalcall's review of the earlier combined PR (#2621), the parts that belong to this per-role-env scope:

- **`MCPConfig` registered in `specgen/schemaNames`** — the public API now exposes `MCPConfig`, not the generated `DomainMCPConfig` name (was an api-drift leak).
- **`isPluginURL` is case-insensitive** (RFC 3986): `HTTPS://example.com/p.zip` routes to `--plugin-url`, not `--plugin-dir`. Parses the URL and compares the scheme; regression test added.
- **Renamed the trimmed role-prompt local `up` → `rolePrompt`** for clarity in `buildSystemPrompt`.

## Scope / what's deliberately NOT here

[...]

PR #4205, opened 2026-08-20
feat(chat): expose preventive read-only permissions

### LINKED ISSUE #3827: feat(chat): expose preventive read-only permission mode for Chat workers
Summary

AO's structured Chat workers currently do not expose a permission mode that allows a worker to inspect a repository while preventively blocking filesystem writes.

For Codex Chat sessions, the current permission mapping is effectively:

default / bypass-permissions
→ approvalPolicy = never
→ sandbox = danger-full-access

accept-edits / auto
→ approvalPolicy = on-request
→ sandbox = workspace-write

This leaves no way for a machine caller to request:

approvalPolicy = never
sandbox = read-only

Codex app-server supports this sandbox profile, and AO already uses enforced read-only Codex execution in its reviewer path.

It would be useful for normal worker / chat / codex sessions to expose an equivalent preventive read-only permission mode.

Use case

A control plane or other automated caller may need to create a Chat worker that can:

inspect repository files;
inspect Git state;
reason about the codebase;
return a structured result;

while being technically unable to modify the workspace.

Prompt instructions such as "do not modify files" and post-run diff checks are useful defense-in-depth, but they are not substitutes for preventive enforcement.

Today a caller that requires a true read-only execution contract has to reject the Chat worker as unsupported even though the underlying Codex sandbox can provide the required isolation.

Suggested behavior

Add a permission profile such as:

read-only

with the Codex Chat mapping:

approvalPolicy = never
sandbox = read-only

The exact public name/API shape is of course up to AO.

Ideally the permission should be represented through AO's normal typed permission/configuration surface rather than as a Codex-specific escape hatch, so callers can capability-check it before creating or delivering work to a session.

Expected behavior

For a Chat worker created with preventive read-only permissions:

the provider cannot modify workspace files;
no approval prompt is required merely to preserve read-only execution;
the mode survives the normal Chat start/resume lifecycle;
unsupported harnesses fail explicitly rather than silently receiving broader permissions;
callers can distinguish supported read-only enforcement from advisory/read-only-by-prompt behavior.
Why this seems to fit AO's existing model

AO already has the relevant pieces:

structured Codex Chat via codex app-server;
typed per-agent permission modes;
preventive Codex read-only execution in the reviewer path.

The missing piece appears to be exposing the equivalent permission profile to ordinary Chat workers.

Related
#3360 — structured Chat/TUI session architecture
#3745 — read-only hardening for reviewer harnesses

This request is specifically about normal structured Chat worker sessions, not reviewer execution.

illegalcall:
Code changes: +505 -84  
Tests: +673 -3  
Others: +80 -4  

Fixes #3827

Ordinary Chat workers can now be created with preventive read-only filesystem permissions. Codex confirms its effective sandbox at thread start/resume, and every turn in a fixed read-only session retains that restriction. Unsupported providers and TUI paths fail explicitly before creating work.

## Summary
- Extend the existing typed permission and durable session state with `read-only`, including `ao spawn --permission read-only` and a per-spawn API override. Reuse the current session-permissions storage; no new migration.
- Map Codex start/resume to `approvalPolicy=never` and `sandbox=read-only`, and turns to `sandboxPolicy={type:"readOnly"}`. Reject missing or broader provider confirmation.
- Preserve provider-confirmed permissions in the persistent host across daemon reconnects. Unknown or broader retained state cannot satisfy a fixed read-only session. Keep pending next-turn choices separate from native launch permissions so mutable sessions can reconnect while a read-only turn is queued.
- Preserve the selected next-turn policy through edit, branch activation, and failed-edit recovery. Freeze settings while a controller handoff is armed, and reject Chat-to-TUI before stopping Chat when the current selection is read-only.
- Advertise read-only-capable Chat harnesses before provider launch. Capability-gate desktop/mobile controls and project defaults, reject attempts to broaden fixed sessions, and constrain queued/automated dispatch through the same controller rule.
- Update generated contracts and CLI documentation; centralize durable session permission resolution instead of repeating project/session precedence.

## Local end-to-end evidence
Tested the actual isolated Electron app, daemon, Codex 0.153.4, and disposable Git worktree. Created a read-only worker, read an existing file, restarted Electron/daemon while retaining the same provider process, then sent a write attempt through the desktop composer.

The provider's recorded tool result was **exit 1: `zsh:1: operation not permitted: probe.txt`**. Direct readback was still `ORIGINAL\n`, and git status was clean. The fixed session's UI offered only Read-only. API attempts to broaden it, create unsupported Claude Code Chat read-only, or create TUI read-only returned `409 PERMISSION_MODE_UNSUPPORTED`; rejected spawns added no sessions.

The earlier [main-branch reproduction](https://github.com/Untrivial-ai/agent-orchestrator/pull/4205#issuecomment-5651636320) remains the before evidence. The refreshed implementation is based on main and preserves the original PR history.

The GUI E2E predates the latest lifecycle-only fixes. Those fixes are covered by focused controller, branch, transition, and race tests.

## Validation
- Backend Go 1.25.7 with GOWORK=off: build, vet, the full race suite, native CLI E2E, and golangci-lint 2.12.2 passed in GitHub CI on final simplified head `2f718ccec`. Locally, all affected packages, vet, the Chat and Session Manager race suites, and focused regressions passed with module files locked read-only.
- Node 24 desktop: typecheck, E2E typecheck, all 303 Vitest files (4,488 passed / 6 skipped), and 59 renderer smoke tests passed. Shared package tests/typechecks and landing icon verification passed.
- Node 20 mobile: typecheck and all 787 tests passed.
- Native macOS helper configuration/packaging tests, both architecture builds, and Swift state tests passed.
- sqlc and API regeneration have no drift. sqlc 1.31.1 requires Go 1.26, so generation used Go 1.26.5; the backend runtime/test suite used the module's Go 1.25.7.
- Separate adversarial standards and spec reviews completed. The final re-review found no actionable issues after the controller-replacement and Chat-to-TUI lifecycle gaps were fixed. A focused complexity review then removed 68 net lines of unused generality and unrelated module churn.

## Notes and validation limits
- Codex Chat is the only shipped driver advertising preventive read-only. This enforces its native filesystem sandbox; external MCP services retain their own authorization boundaries.
- Legacy sessions without stored permissions keep the existing project-default resolution behavior.
- The local Electron compositor required a window resize to refresh the shell surface; no application workaround was added to this permission change.
- The first desktop suite encountered an incomplete Electron install and parallel-load timing failures. After fixing the installation, affected files and the full suite passed (four workers).
- Docker was unavailable locally. All 15 GitHub CI checks passed on final simplified head `2f718ccec`, including the fresh-install container check, native CLI E2E on Linux/macOS/Windows, the full Linux backend race suite, and Windows-specific backend checks; no publishing or deployment was used as validation.
- A final optional full local lint rerun was limited by host disk exhaustion. The affected normal/race suites passed locally, and the repository-wide jobs passed in GitHub CI.


backend/internal/service/chat/service.go
```diff
@@ -294,6 +298,9 @@ func (s *Service) Start(ctx context.Context, cfg StartConfig) (*Controller, erro
 	if err != nil {
 		return nil, fmt.Errorf("open conversation: %w", err)
 	}
+	if err := validatePermissionMode(caps, conversation.Settings.ApprovalMode); err != nil {
```
illegalcall, line 310: [P1] Enforce the session read-only floor against stored turn settings

This only checks whether the stored mode is supported; default and bypass-permissions always pass even when cfg.Permissions is read-only. Project-scoped orchestrator conversations reuse prior settings, so a new read-only spawn can inherit a broader mode without any post-spawn action. Controller.dispatch puts that mode on every turn, and Codex maps it to dangerFullAccess, overriding the thread-level read-only sandbox. The same broadening is allowed later by SetSettings. Reconcile or reject broader stored/per-turn modes whenever the session contract is read-only, and cover the reused-conversation/resume case.

backend/internal/session_manager/manager.go
```diff
@@ -1266,6 +1269,26 @@ func effectiveAgentConfig(kind domain.SessionKind, cfg domain.ProjectConfig) por
 	return merged
 }
 
+// effectiveSessionAgentConfig keeps the permission contract chosen when a
+// session was created. Empty is the compatibility marker for sessions created
+// before permissions were durable, so those continue resolving project config
+// on restore exactly as they did before the migration.
+func effectiveSessionAgentConfig(rec domain.SessionRecord, cfg domain.ProjectConfig) ports.AgentConfig {
+	config := effectiveAgentConfig(rec.Kind, cfg)
+	if rec.Metadata.Permissions != "" {
```
illegalcall, line 1278: [P2] Reserve the empty sentinel for migrated sessions

Both fresh persistence paths write agentConfig.Permissions verbatim (chat_spawn.go:173 and manager.go:927). With normal unset config that is empty, so this branch misclassifies a brand-new session as pre-migration and re-resolves mutable project permissions on restore. A later project change can silently alter or broaden the session. Normalize fresh empty values to PermissionModeDefault before persisting and add an omitted-permission restore regression.

illegalcall (commented):

## Standards
The migration and API boundaries look consistent, but fresh sessions with an omitted permission reuse the legacy empty sentinel and can resolve a different project policy on resume.

## Spec
The Codex mapping is correct in isolation, but a stored broader per-turn setting can override a new read-only session and restore dangerFullAccess, violating the preventive contract.

Two blocking findings are inline. I did not run tests or builds because this review task requires static inspection only.

Summary: Standards: 1 blocking finding; Spec: 1 blocking finding.

backend/internal/service/chat/service.go
illegalcall: Fixed in 99d3a4453. Read-only starts and resumes now reconcile inherited durable settings before controller activation; every controller retains and enforces that floor on later settings changes and dispatch, including branch replacements. Added cross-session project-conversation resume and escalation regression coverage.

backend/internal/session_manager/manager.go
illegalcall: Fixed in 8a96a19ec. Both fresh TUI and Chat metadata writes now normalize omitted permissions to explicit default, reserving empty for migrated rows only. Added fresh-TUI persistence coverage plus a Chat spawn/kill/project-change/restore regression, while the legacy-empty restore test continues proving backward-compatible project resolution.

illegalcall:
Addressed review 4990490364 on current head 8a96a19ec (the review inspected be4f5f887):

- P1 was fixed in 99d3a4453. A read-only session now constrains inherited conversation settings during start/resume, retains that floor across branch replacement, rejects later default or bypass-permissions escalation in SetSettings, and constrains dispatch. Regression coverage: TestReadOnlyResumeDoesNotDispatchStoredBroaderTurnSettings and TestReadOnlySessionRejectsBroaderTurnSettings.
- P2 was fixed in 8a96a19ec. Both fresh TUI and Chat persistence paths normalize omitted permissions to explicit default; empty remains the pre-migration sentinel. Regression coverage includes omitted TUI persistence, Chat spawn/kill/project-policy-change/restore, and legacy-empty compatibility.

Reverified both focused suites on 8a96a19ec; they pass. All PR CI checks are also green. Both inline threads from this review have replies with their fixing commits and are resolved.

frontend/src/renderer/components/chat/TurnSettingsBar.tsx
```diff
@@ -95,6 +102,9 @@ export function TurnSettingsBar({
 	const effortLabel =
 		settings.reasoningEffort ?? (selected ?? fallback)?.defaultEffort ?? undefined;
 	const approvalLabel = APPROVAL_COPY[settings.approvalMode ?? "default"].label;
+	const approvalModes = supportsPreventiveReadOnly
```
illegalcall: **[P2] Respect the immutable read-only floor in this menu.**

`supportsPreventiveReadOnly` describes driver capability only. For a session created with durable `permissions: "read-only"`, this branch still returns `default`, `accept-edits`, `auto`, and `bypass-permissions`, but the new `Controller.SetSettings` floor rejects all four as attempts to broaden the session. The desktop therefore renders choices guaranteed to fail; mobile has the same issue in `ChatSettingsModal.tsx`. The PR already exposes the session permission on `SessionView`, so carry that fact into the client model, restrict or disable broader choices, and cover a read-only-floor session in both client tests.

illegalcall (commented):
Changes requested.

## Standards

The two earlier backend blockers are fixed: read-only is now enforced as a controller floor across stored settings, dispatch, branches, and resume; fresh omitted permissions are persisted as explicit `default`.

- **Blocking (P2):** The desktop and mobile turn-settings menus still expose broader approval modes for sessions whose durable permission floor is `read-only`. Capability support only says the driver can enforce read-only; it does not mean a read-only session may be broadened. Every such choice is rejected by the new backend floor. Carry the session permission into these client models, hide or disable invalid modes, and add client regression coverage.
- **Non-blocking:** `Registry.SupportsPermissionMode` remains unused in production while the service duplicates the lookup, contrary to the repository rule against one-off abstractions. Remove it or expose and delegate it through the port.

## Spec

No remaining spec deviations found. Codex maps read-only to `approvalPolicy=never` and `sandbox=read-only`; unsupported harnesses refuse it; static and live capability checks exist; and durable permission restoration now survives project-policy changes.

Summary: Standards — 1 blocking, 1 non-blocking; Spec — 0. Needs changes for the invalid client controls.

illegalcall:
Addressed review 4990929888 in b7035e169:

- Desktop now preserves SessionView.permissions in WorkspaceSession and passes it to TurnSettingsBar. A read-only floor is authoritative even if the conversation snapshot is stale: the trigger displays Read only and the menu contains no default, accept-edits, auto, or bypass-permissions choices.
- Mobile now preserves permissions for worker and orchestrator session models, carries the floor through the settings-sheet registry, and applies the same read-only-only approval catalog.
- Added desktop UI/mapping regressions and mobile catalog/API-mapping regressions.
- Exposed SupportsPermissionMode through ChatDriverRegistry and changed the chat service to delegate to that canonical registry lookup instead of duplicating it.

Verification: full frontend suite (200 files, 2630 passed), full mobile suite (55 files, 455 passed), both TypeScript typechecks, full backend go test ./..., and isolated golangci-lint (0 issues).

frontend/src/renderer/components/chat/TurnSettingsBar.tsx
illegalcall: Fixed in b7035e169. Desktop now carries the durable session permission into TurnSettingsBar; under a read-only floor the effective label and sole menu item are Read only, even if stored turn settings are stale. Mobile preserves the same field for worker/orchestrator models and carries it through the sheet entry into a shared approval catalog that returns only read-only. Added client mapping and menu/catalog regressions for both surfaces.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -705,6 +705,8 @@ export function ChatWorkspace({
 										onChangeConfigOption={onChooseConfigOption}
 										configPending={configOptionPending}
 										error={configOptionError}
+										supportsPreventiveReadOnly={can(snapshot, "preventive_read_only")}
+										permissionFloor={session?.permissions}
```
illegalcall: **[P2] Use the effective restored permission floor here.**

`SessionView.permissions` is deliberately omitted for pre-migration rows, but `effectiveSessionAgentConfig` resolves those rows from the current project config on restore. After upgrading, setting the project to read-only and restoring an older Chat session therefore creates a live read-only controller while this prop remains `undefined`; the picker exposes `default`, `accept-edits`, `auto`, and `bypass-permissions`, and every selection fails at the daemon floor. Mobile passes the same raw marker. Surface the controller-resolved effective floor (for example on the conversation snapshot) and cover the legacy-empty plus project-read-only restore case in both clients.

illegalcall (commented):
Changes requested.

## Standards

**[P2] The latest client-floor plumbing still fails for migrated sessions.** Pre-migration rows intentionally keep an empty durable permission marker, while restore can resolve the current project permission to `read-only`. Desktop and mobile copy the raw empty marker into their turn-settings controls, so both still expose broader modes that the daemon will reject. This leaves a changed user-visible boundary uncovered. Surface the controller-resolved effective floor and add desktop/mobile coverage for legacy-empty plus project-read-only restore.

The earlier backend persistence/floor issues are fixed, explicit durable read-only sessions are now constrained in both clients, and the registry capability helper now has a real production caller.

## Spec

The same migrated-session hole leaves the requirement that read-only survive the normal Chat resume lifecycle incomplete at the client surface. Enforcement remains secure, but restored users are offered invalid settings and receive errors. Codex mapping, unsupported-harness refusal, fresh-session durability, and capability discovery otherwise satisfy the issue.

Summary: Standards — 1 finding (P2); Spec — 1 finding (P2). Changes requested.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
illegalcall: Fixed in 5594374b6. The service now includes the live controller's normalized `permissionFloor` in both full and paged conversation snapshots and the HTTP contract. Desktop and mobile settings read that effective snapshot field, not `SessionView.permissions`, so a legacy-empty row restored under a read-only project policy exposes only read-only. Regression coverage spans the legacy restore, paged/API snapshot, desktop workspace, and mobile API/settings catalog.

illegalcall:
Addressed review 4991227532 in 5594374b6. Conversation snapshots now expose the normalized permission floor held by the live controller, including the project-resolved `read-only` floor for migrated sessions with an empty durable marker. Desktop and mobile settings consume that effective snapshot value instead of the raw session marker, so broader approval modes are no longer offered. Added regressions for legacy-empty/project-read-only restore, paged snapshot/API propagation, desktop controls, and mobile API/catalog behavior. Full backend, desktop, and mobile suites, both TypeScript typechecks, API drift tests, and golangci-lint pass.

illegalcall (commented):
Ready — approving.

## Standards

No findings. The prior legacy-session blocker is fixed through the daemon boundary: the live controller exposes its effective restored permission floor in both snapshot paths, desktop and mobile consume that value, and the relevant boundary coverage is present. The earlier backend floor, explicit fresh-default persistence, and registry-helper issues also remain fixed. Migration, sqlc, OpenAPI, and generated frontend artifacts move together.

## Spec

No findings. The implementation now satisfies the requested typed preventive read-only mode, Codex `approvalPolicy=never` plus `sandbox=read-only` mapping, start/resume durability (including migrated sessions), explicit unsupported-harness refusal, and capability discovery without a broader fallback.

Summary: Standards — 0 findings; Spec — 0 findings. Ready.

illegalcall (commented):
Ready — approving.

## Standards

No findings. The rebase correctly moves the new permission migration to `0103_session_permission_mode.sql` after main's `0102_canonical_usage.sql` and appends the matching ledger entry. Query/sqlc and OpenAPI/frontend generated artifacts remain paired. Range-diff shows this migration renumbering is the only patch change from the previously approved range; the prior backend and client fixes remain intact.

## Spec

No findings. The rebase preserves the typed preventive read-only mode, Codex `never` plus `read-only` mapping, start/resume durability (including migrated sessions), explicit unsupported-harness refusal, effective-floor propagation, and capability discovery.

Summary: Standards — 0 findings; Spec — 0 findings. Ready.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

kou-keirei:
Some field evidence from forward-porting the preventive read-only work onto a newer AO base and using it through an external execution boundary.

The core model worked well once provider-effective state was treated as the source of proof rather than requested configuration alone.

End-to-end validation covered:

fresh Codex start: requested read-only, effective read-only, approvalPolicy=never, read-only thread/turn sandbox, preventive capability true;
prohibited filesystem write: actually blocked;
normal resume: provider posture re-established and preventive proof remained valid;
already-live reconnect: left truthfully unproven when current provider-effective posture could not be queried, rather than replaying config and calling that proof;
direct worker task execution: PASS;
daemon restart + exact-session recovery: PASS;
AO orchestrator with two delegated workers: PASS;
descendant authority cap: a child attempt to request broader auto authority under a read-only parent was rejected, and the child ran read-only instead;
both delegated workers produced independently verified results while the parent/children remained read-only.

The main integration invariant that proved useful was:

effective child authority =
narrowest(parent/orchestrator floor, child requested authority)

One other useful distinction: dispatch and completion should remain separate. In my integration the top-level execution remained truthfully dispatched with completion evidence unproven, even though the actual worker/orchestrator results were independently observable and verifiable.

So the preventive-read-only concept held up well under direct-worker, resume, recovery, and delegated-orchestrator E2E use. The only intentionally unsupported case in my forward-port is proving current provider posture for an already-live reconnect when Codex does not expose a queryable effective-state surface.

i-trytoohard:
@illegalcall Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@illegalcall Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

illegalcall:
## Agent-collected evidence: live local E2E, 13 September 2026

The requirement in #3827 is still unresolved on `main` at `cadde8c9fd2079d0decca654004760ad8439328e`.

I tested the real AO Electron app with its Go daemon and an authenticated live Codex app-server (`codex-cli 0.153.4`), using a separate checkout, isolated AO data/profile, and a disposable Git repository. Project/session setup used the CLI; permission selection and the write-test message were exercised in the actual Electron Chat UI.

| Check | Observed result |
| --- | --- |
| Project config with `agentConfig.permissions: "read-only"` | HTTP 400, `INVALID_PROJECT_CONFIG`: read-only is outside the accepted permission vocabulary. |
| Running Chat settings with `approvalMode: "read-only"` | HTTP 400, `CHAT_APPROVAL_MODE_INVALID`: unknown approval mode. |
| Live desktop permission menu | Full access, Ask for approval, Approve for me, Bypass permissions. No Read-only option. |
| Initial live worker read | `probe.txt` contained `ORIGINAL`. |
| Select Ask for approval in Electron | The daemon reported `approvalMode: "accept-edits"`. |
| Send a bounded write probe through Electron Chat | Codex wrote the file and read it back; both commands exited 0, the turn completed, and no approval prompt was shown or resolved. |

The test prompt authorized changing only this disposable file and explicitly prohibited permission escalation, sandbox changes, unrelated edits, commits, or contacting other sessions. The actual filesystem and Git diff confirmed:

```diff
-ORIGINAL
+AO_E2E_WRITE_ALLOWED
```

The provider's completed response was: “Write succeeded. Read-back confirmed exactly `AO_E2E_WRITE_ALLOWED` followed by a newline.” Only `probe.txt` was modified. The test worker, Electron app, and daemon were then stopped; the source checkout remained clean.

This is expected behavior for `accept-edits`; it demonstrates that the available setting does not supply preventive read-only execution. It is **not** evidence that an accepted read-only mode was bypassed. This test assesses current main, not the unmerged PR's fix; other harnesses, mobile, and restart/resume enforcement were not tested in this run.

I am updating this PR against current main. #4956 has already supplied durable session permissions and remembered project defaults, and the newer persistent-provider lifecycle also needs to be respected. The refreshed implementation will reuse those paths rather than add a second permission column, preserve current defaults, and cover enforcement and unsupported-driver refusal with regression tests and a live prohibited-write check.

illegalcall:
Implemented the refreshed solution on current main and pushed it here (head `2f718ccec`; native-E2E implementation tree `643805b7b`). The original PR history is preserved.

**Real local E2E now passes.** This used the actual isolated Electron app → daemon → Codex 0.153.4 → disposable Git worktree.

| Check | Observed result |
| --- | --- |
| Create ordinary Codex Chat with `ao spawn --permission read-only` | Session starts, reads `probe.txt` successfully |
| Restart Electron and daemon | Same provider host and Codex process survive; session reconnects ready with read-only permissions |
| Desktop fixed-session permission menu | Only Read-only is available |
| Send a shell write attempt through the desktop composer | Actual provider tool result: exit 1, `zsh:1: operation not permitted: probe.txt` |
| Read file back and inspect Git worktree directly | Still `ORIGINAL` plus newline; clean git status |
| Try to broaden fixed permissions to bypass | HTTP 409 `PERMISSION_MODE_UNSUPPORTED` |
| Spawn Claude Code Chat or Codex TUI with read-only | HTTP 409, no additional session created |

The recorded provider call attempted a shell redirect into the existing `probe.txt`; I checked its paired tool result, not only the assistant's description. The original main-branch test above allowed a write under accept-edits because preventive read-only did not exist there.

The solution uses existing durable permission storage and one controller enforcement rule, validates Codex's effective sandbox, and retains a provider-confirmed read-only receipt for live reconnects. Separate adversarial standards and spec reviews caught and fixed the pending-next-turn reconnect case, stale settings across controller replacement, and Chat-to-TUI selected read-only handling. A focused complexity review then removed 68 net lines while preserving those guarantees.

Local checks passed: full backend race suite/build/vet/lint, native CLI E2E, 4,488 desktop tests (6 skipped), 787 mobile tests, 59 renderer smoke tests, both macOS helper architecture builds/tests, shared package checks, and API/sqlc drift. The isolated desktop compositor needed a window resize to refresh its surface; no compositor code was changed.

All **15 GitHub CI checks passed** on final simplified head `2f718ccec`, including the full Linux backend/frontend suites, the fresh-install container check, and native CLI E2E on Linux/macOS/Windows. Locally, that head passed all affected Go packages, vet, affected race suites, desktop typecheck, 85 focused desktop tests, and mobile typecheck. Docker was unavailable locally, so the container result is from CI. External MCP services retain their own authorization boundaries; the feature enforces Codex's native filesystem sandbox.

## Attached E2E evidence — 13 September 2026

These captures come from the real Electron/daemon/Codex 0.153.4 tests above. **Before:** main `cadde8c9f`. **After:** implementation `643805b7b`, also present as the identical implementation tree at PR revision `ba94c58b4`. Later compatibility and lifecycle fixes through `8c9a1ae01` were checked with focused regressions, affected race suites, and all hosted CI. The simplified head `2f718ccec` has the same behavior, passed the affected local suites, and passed all hosted CI; these attachments are not a GUI E2E rerun of that exact head.

**Before — Ask for approval allowed the disposable-file edit.** Main rejected `read-only` configuration and had no Read-only UI option. This demonstrates the missing AO feature; an accepted read-only sandbox was not bypassed. [Original reproduction](https://github.com/Untrivial-ai/agent-orchestrator/pull/4205#issuecomment-5651636320).

<details>
<summary>Screenshot: write succeeded under Ask for approval</summary>

Before: Ask for approval permits the disposable-file write

</details>

**After — the fixed session offers only Read-only.** After restarting Electron and the daemon, the same provider processes survived and the session reconnected with read-only permissions.

<details>
<summary>Screenshot: fixed Read-only permission menu</summary>

After: fixed session exposes only Read-only

</details>

**Actual prohibited-write result after restart:**

```text
exit_code: 1
zsh:1: operation not permitted: probe.txt
```

Direct filesystem verification found `probe.txt` still contained `ORIGINAL` plus a newline, and Git status was clean. The screenshot above documents the permission menu; the recorded provider tool output supplies the write-enforcement evidence.

[Download the paired native tool call/result, reconnect state, and direct filesystem verification](https://github.com/user-attachments/files/32158875/pr-4205-readonly-e2e-evidence.txt). The evidence excerpt replaces the local absolute worktree path with `<disposable-worktree>` and omits unrelated transcript/machine metadata.

illegalcall:
Audited the repo's **315 open PRs**, including all changed-file inventories and exact-head merge simulations for the 125 other PRs sharing files with this one. I found no duplicate of the complete fixed, preventive read-only Chat worker contract, but there is real integration overlap.

I pushed a scope cleanup at `902bdaea2`: retained existing dependency checksums, removed unrelated struct-formatting churn, and moved documentation. This eliminated newly introduced conflicts with **#5194, #5125, #5085, and #5176**. Permission behavior and generated API contracts are unchanged.

**Remaining coordination, not an all-clear:**
- **#4930** changes the shared permission UI and native defaults, and separately improves runtime read-only mapping. Its Manual mode permits approval; this Chat mode uses never/read-only and a fixed session restriction. It already conflicts with main and needs deliberate integration with this PR.
- **#5132** fixes Codex Default-mode behavior. This PR leaves that behavior alone, but both edit thread start/resume. Preserve its default/posture handling and this PR's read-only confirmation.
- **#5218** refactors persistent-host I/O. Preserve its output locking/deadlines and place permission observers under its state lock, with request invalidation before provider writes.
- **#2848** expands AgentConfig validation. Keep both read-only validation and its plugin/MCP checks.

Of 314 other PRs: 189 have no shared files; 45 merge cleanly with current main and this revision; 77 already conflict with main (14 gain additional conflicting files here); 3 previously clean PRs now conflict (#5132, #5218, #2848). No other PR branch was changed or imported.

Validation after cleanup: full local backend race/build/vet/lint, frontend typechecks and all 4,488 frontend tests passed; API regeneration has no drift. Isolated combinations with **#5164, #5260, #5286, #5255, and #5144** also compiled affected packages and passed read-only regressions. Pairwise merge checks and targeted tests do not guarantee every possible combined stack at runtime.

The prior implementation head passed all 14 CI checks; fresh checks for this cleanup are pending.

illegalcall:
write-succeeded
read-only-menu
[pr-4205-readonly-e2e-evidence.txt](https://github.com/user-attachments/files/32158875/pr-4205-readonly-e2e-evidence.txt)

nikhilachale (commented):
I found two remaining issues on the current head (`2f718ccec`):

1. **[P1] Agent switching drops a pending read-only policy** (`backend/internal/session_manager/agent_switching_chat.go:164`). The target preflight and launch use `sessionPermissions`, ignoring the durable next-turn setting. On a fresh target, Chat startup overwrites the stored read-only choice, so the automatic activation turn runs with broader permissions. I reproduced this by selecting read-only on a mutable source session and switching Claude Code -> Codex; the target activation message was dispatched with `Settings.Approval == ""` instead of `read-only`. This also lets an unsupported target bypass the read-only capability gate.

2. **[P2] Model-only settings updates fail for fixed read-only sessions** (`backend/internal/service/chat/controller.go:1456`). `ConversationTurnSettingsPayload` documents every field as optional, but a request containing only `model` reaches this check with an empty approval mode and is rejected as an attempt to grant write access. I reproduced the model-only update returning `ErrChatPermissionModeUnsupported`. The immutable floor should fill or preserve `read-only` before validating the replacement settings.

All existing CI checks pass, but neither regression is currently covered.

backend/internal/session_manager/interface_transition.go
```diff
@@ -829,6 +840,35 @@ func (m *Manager) preflightInterfaceTarget(
...
+	if permissions == ports.PermissionModeReadOnly || domain.NormalizeSessionMode(rec.Mode) != domain.SessionModeChat {
+		return permissions, nil
+	}
+	settingsStore, ok := m.store.(chatHandoffSettingsStore)
+	if !ok {
+		return "", fmt.Errorf("%w: current Chat permissions cannot be verified", ErrInterfaceHandoffUnsupported)
+	}
+	conversation, err := settingsStore.ConversationForSession(ctx, rec.ID)
+	if err != nil {
+		return "", fmt.Errorf("load current Chat permissions: %w", err)
```
illegalcall, line 866: `ConversationForSession` returns `domain.ErrNoConversation` when a Chat session has no conversation row, and that is propagated here as a hard failure. The effect is that Chat to TUI preflight fails for any Chat session whose conversation was never opened (for example one whose controller never started successfully), which removes the escape hatch out of a broken Chat session. Suggest treating `errors.Is(err, domain.ErrNoConversation)` as no read-only next-turn selection and falling through to the `permissions` value.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -1122,6 +1124,8 @@ function ChatWorkspaceContent({
 			rememberPermissionsPending,
 			rememberPermissionsError,
 			rememberedPermissionMode,
+			snapshot.permissions,
+			snapshot.capabilities,
```
illegalcall: `snapshot.settings` and `snapshot.modelReroute` are deliberately wrapped in `useStableValue` just above (lines 1087-1088) so this memo survives a snapshot refresh. `snapshot.capabilities` is a fresh array identity on every snapshot, so adding it raw to the dependency list defeats that stabilization and rebuilds the settings bar on every poll. Either stabilize it the same way, or hoist `supportsReadOnly` into a boolean and depend on the primitive.

frontend/src/renderer/components/ProjectSettingsForm.tsx
```diff
@@ -735,14 +738,14 @@ function AgentModelField({
...
+function PermissionModeSelect({ value, onChange, supportsReadOnly }: { value: string; onChange: (value: string) => void; supportsReadOnly: boolean }) {
 	const { t } = useTranslation();
 	const options = [
 		{ value: "__default__", label: `${t("settings.project.permissionAuto")} (${t("settings.project.default")})` },
-		...PERMISSION_MODE_VALUES.map((value) => ({
+		...PERMISSION_MODE_VALUES.filter((mode) => mode !== "read-only" || supportsReadOnly || value === "read-only").map((value) => ({
 			value,
 			label:
-				value === "default"
+				value === "read-only" ? "Read-only (Chat only)" : value === "default"
```
illegalcall: Hardcoded English string in a file where every other permission label goes through `t(...)`, and no key was added to the eight locale files. Suggest a `settings.project.permissionReadOnly` key alongside `permissionDefault` / `permissionAcceptEdits` / `permissionAuto` / `permissionBypass`.

backend/internal/service/chat/controller.go
```diff
@@ -1433,14 +1437,34 @@ func retryPromptContent(raw string, capabilities ports.ChatCapabilities) ([]port
...
 	defer c.mu.Unlock()
-	return c.settings
+	return c.constrainSettings(c.settings)
 }
 
 // SetSettings records the provider choices for the next turn.
 //
 // The row is written first: if that fails, the in-memory copy must not move, or a
 // restart would silently revert a choice the user watched take effect.
 func (c *Controller) SetSettings(ctx context.Context, settings domain.ConversationSettings) error {
```
illegalcall: After this change both call sites (`SetTurnSettings` and `SetConfigOption`) take `configMu` themselves and call `setSettingsLocked`, so the exported `SetSettings` has no callers left, including tests. Keeping two entry points invites a future caller to pick the one that sits outside the `SetConfigOption` fence. Suggest deleting it.

illegalcall (commented):
## Requesting changes (all minor; nothing blocks the security model)

I diffed `c15088faa3...2f718ccec8` and traced the enforcement chain end to end. The core contract holds:

- `Controller.dispatch` unconditionally overwrites `msg.Settings` with `turnSettings()`, and `Settings()` runs `constrainSettings`, so a caller-supplied broader approval on a read-only session cannot reach the provider (covered by `readonly_test.go`).
- `Service.Start` re-derives the floor from `record.Metadata.Permissions`, so a restore, agent switch, or branch replacement cannot broaden it.
- The Codex host proof machinery enumerates the only two in-thread policy-changing client requests: `thread/resume` and `turn/start`. I checked the generated protocol: `turn/steer` carries no policy, and `thread/fork` produces a new thread id whose proof comes from its own response. The enumeration is complete.
- `ThreadResumeResponse` does carry `thread.id`, so `observeCodexReadOnly` records proof on the resume path too, not just on `thread/start`.
- ACP drivers reject read-only at `Start`, `Resume`, and `applyTurnSettings`; TUI is rejected at spawn, restore, and interface transition.

Four things to fix, listed inline:

1. `interface_transition.go` - `ErrNoConversation` hard-fails the Chat to TUI escape hatch.
2. `ChatWorkspace.tsx` - raw `snapshot.capabilities` in the memo deps defeats the deliberate `useStableValue` pattern two lines above.
3. `ProjectSettingsForm.tsx` - hardcoded English label in an otherwise fully translated control.
4. `controller.go` - `Controller.SetSettings` is now dead code; only `setSettingsLocked` is reachable.

Two smaller notes, no change required:

- `codexReadOnly` is keyed by thread id and never pruned; a long-lived host accumulates one bool per forked thread. Harmless at current scale.
- `observeCodexReadOnly` adds a second full `json.Unmarshal` per provider frame, alongside the existing `serverRequestID` decode, on the streaming relay path under `host.mu`. Same category as what is already there, so not a regression, but it is now two full scans of every token delta.

Note on the checkout: the reviewer worktree for this session is pinned to an Aug 21 base and holds a superseded five-commit version of this work. This review is of the pushed head `2f718ccec` against its merge base `c15088faa3`.

backend/internal/session_manager/interface_transition.go
illegalcall: Fixed in 414aa87. `interfaceTransitionPermissions` now treats `errors.Is(err, domain.ErrNoConversation)` as no read-only next-turn selection and falls through to the launch `permissions`, so a Chat session that was never opened keeps its escape hatch to TUI. Added `TestInterfaceTransitionPermissionsWithoutConversation` in `backend/internal/session_manager/readonly_test.go`; verified it fails without the guard (`load current Chat permissions: session has no conversation`) and passes with it.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
illegalcall: Fixed in 414aa87. Hoisted `const supportsReadOnly = snapshot.capabilities?.includes("preventive_read_only") ?? false` above the memo and depend on that boolean instead of the raw array, so the `useStableValue` stabilization of `settings`/`modelReroute` survives a snapshot refresh.

frontend/src/renderer/components/ProjectSettingsForm.tsx
illegalcall: Fixed in 414aa87. Label now goes through `t("settings.project.permissionReadOnly")`, with the key added to all eight locale files (en/de/es/fr/ja/ko/pt-BR/zh-CN). i18n coverage tests pass.

backend/internal/service/chat/controller.go
illegalcall: Fixed in 414aa87. Deleted `Controller.SetSettings` (grep confirmed zero callers repo-wide); `SetTurnSettings` and `SetConfigOption` remain the only entry points, each taking `configMu` and calling `setSettingsLocked`. Moved its doc comment about write-ordering onto `setSettingsLocked`.

illegalcall:
## Review 5260686154 addressed — pushed as `414aa87`

Worktree was refreshed first (it was pinned to the Aug 21 base `f739ce18` with the superseded five-commit version); it now sits on the pushed head `2f718cce`.

| # | Finding | Fix |
|---|---------|-----|
| 1 | `interface_transition.go:861` — `ErrNoConversation` propagated as a hard failure | `interfaceTransitionPermissions` now treats `errors.Is(err, domain.ErrNoConversation)` as no read-only next-turn selection and falls through to the launch `permissions`. New `TestInterfaceTransitionPermissionsWithoutConversation` in `readonly_test.go` — verified red without the guard (`load current Chat permissions: session has no conversation`), green with it. |
| 2 | `ChatWorkspace.tsx:1128` — raw `snapshot.capabilities` in deps defeats `useStableValue` | Hoisted `const supportsReadOnly = snapshot.capabilities?.includes("preventive_read_only") ?? false` above the memo; deps now carry the boolean. |
| 3 | `ProjectSettingsForm.tsx:748` — hardcoded English label | Now `t("settings.project.permissionReadOnly")`; key added to all eight locales (en/de/es/fr/ja/ko/pt-BR/zh-CN). |
| 4 | `controller.go:1447` — `SetSettings` has no callers | Deleted (grep confirmed zero callers repo-wide). `SetTurnSettings` / `SetConfigOption` remain the only entry points behind the `configMu` fence; the write-ordering doc comment moved onto `setSettingsLocked`. |

The two "no change required" notes (unpruned `codexReadOnly` map, second `json.Unmarshal` per provider frame) were left as-is per the review.

### Verification
- `go build ./...` clean; `go vet` clean on `session_manager` and `service/chat`.
- `go test` green: `service/chat`, `httpd/controllers`, `adapters/chatdriver/...` (all 15 packages), and the read-only / interface-transition tests in `session_manager`. Four `TestSpawn_*MissingBinary*` failures in `session_manager` are pre-existing and environmental on this machine (`tmux required on macOS but ... was not found`), unrelated to this diff.
- `tsc --noEmit`: zero errors under `frontend/src` (remaining output is missing `packages/product-ui` deps in this checkout).
- `vitest`: `ChatWorkspace` + `TurnSettingsBar` + `ProjectSettingsForm` 202 passed; i18n coverage/parity 16 passed.

Note: the PR still reports `CONFLICTING` against `main` — happy to rebase/merge main if you want that before re-review.

frontend/src/renderer/i18n/en.json
```diff
@@ -1138,6 +1138,7 @@
 	"settings.project.permissionAuto": "Auto",
 	"settings.project.permissionBypass": "Bypass permissions",
 	"settings.project.permissionDefault": "Default",
+	"settings.project.permissionReadOnly": "Read-only (Chat only)",
```
illegalcall, line 1141: Nit: these files are alphabetically sorted, and `permissionReadOnly` sorts after `permissionMode`, not before it. The key landed one line early here and in the other seven locale files. No CI impact - `instance.test.ts` sorts keys before comparing, so it only enforces cross-locale parity, not file order.

illegalcall (commented):
## Approving

Re-reviewed `2f718ccec..414aa873`. All four findings from the previous review are addressed correctly, and I verified each fix against the path it claims to fix rather than taking the commit message for it.

### Findings resolved

1. **Chat to TUI escape hatch** - `interfaceTransitionPermissions` now returns the launch policy on `domain.ErrNoConversation`. I checked that this matches production and not just the fake: the sqlite store returns `ErrConversationNotFound`, which is declared as `var ErrConversationNotFound = domain.ErrNoConversation` (same sentinel, not a parallel error), so `errors.Is` matches the real path. The new `TestInterfaceTransitionPermissionsWithoutConversation` genuinely exercises the branch - the fixture is created with `SessionModeChat` so the earlier mode short-circuit is not taken, and `transitionSettingsStore.ConversationForSession` returns `ErrNoConversation` once `store.conversation` is zeroed.
2. **Dead `Controller.SetSettings`** - removed, with the row-written-first rationale folded onto `setSettingsLocked`. Zero remaining references in the tree; both surviving call sites (`SetTurnSettings`, `SetConfigOption`) still take `configMu` before calling it, so the fence is intact.
3. **Hardcoded label** - now `t("settings.project.permissionReadOnly")`, with the key present in all eight locale files and real translations rather than English fallbacks.
4. **Memo dependency churn** - `supportsReadOnly` is hoisted to a primitive boolean (`?? false`, so it matches the prop type) and the raw `snapshot.capabilities` array is out of the dependency list. Consistent with the `useStableValue` treatment of `settings` and `modelReroute` above it.

The enforcement chain I traced in the previous review is untouched by this commit - `dispatch`, `constrainSettings`, the `Service.Start` floor, and the Codex host proof logic are all unchanged.

### One cosmetic nit, not blocking

The new locale key is inserted one line early in all eight files (see inline). Nothing in CI enforces file order - `instance.test.ts` sorts keys before comparing, so it only checks cross-locale parity - so this is purely for the next person reading the file.

Issue #4206, opened 2026-08-21
bug(cli): ao doctor gitlab-token probe hardcodes gitlab.com, ignores self-managed host config


mcmxcdev:
## Bug

`ao doctor`'s `gitlab-token` check always probes `https://gitlab.com/api/v4/user`, regardless of `AO_GITLAB_ALLOWED_HOSTS` / `AO_GITLAB_HOST_TOKENS`. On a self-managed GitLab install the check reports a token verdict about a host AO will never call: a valid self-hosted PAT reads as `FAIL ... token rejected by GitLab (HTTP 401)`, and a broken self-hosted setup can read as `PASS`.

**Reported by:** @MCMXC | **Analyzed against:** `bc68557ff`
**Confidence:** High

## Reproduction

1. `export AO_GITLAB_ALLOWED_HOSTS=gitlab.example.com`
2. `export AO_GITLAB_TOKEN=<PAT valid on gitlab.example.com>`
3. `ao doctor`
4. GitLab section reports on gitlab.com; the probe URL in the failure is `gitlab.com/api/v4/user`. `gitlab.example.com` is never contacted.

## Root Cause

`backend/internal/cli/doctor.go`:

- `defaultDoctorGitLabRESTBase = "https://gitlab.com/api/v4"` (line 56) is the only base the check knows.
- `checkGitLabToken` builds the probe from `c.deps.DoctorGitLabRESTBase` unconditionally (line ~519) and never consults `cfg.GitLab`.
- `runDoctor` loads `config.Load()` (which populates `cfg.GitLab.AllowedHosts` / `cfg.GitLab.HostTokens` from `AO_GITLAB_ALLOWED_HOSTS` / `AO_GITLAB_HOST_TOKENS`) but does not pass it to the GitLab check.
- Token selection has the same gap: the per-host override in `AO_GITLAB_HOST_TOKENS` is ignored; the check always uses the default `AO_GITLAB_TOKEN`/`GITLAB_TOKEN`/`glab` token.

The SCM provider already does this correctly — `Provider.clientForHost` (`backend/internal/adapters/scm/gitlab/provider.go`) derives `https://<host>/api/v4` per allowlisted host and selects the per-host token. Doctor is the one code path that never got host awareness.

[...]

PR #4209, opened 2026-08-21
fix(gitlab): make doctor and daemon credentials host-aware

mcmxcdev:
## What

Makes GitLab credential handling host-aware in both `ao doctor` and the daemon, so self-managed instances are actually supported end to end.

- `ao doctor` now emits one check per configured instance: `gitlab-token` for gitlab.com and `gitlab-token:<host>` for every `AO_GITLAB_ALLOWED_HOSTS` entry, resolving the `AO_GITLAB_HOST_TOKENS` override exactly the way the SCM provider does.
- The daemon's gitlab.com token chain scopes its `glab` lookup to gitlab.com, and never consults `glab`'s own default host unscoped.
- Each allowlisted host without an explicit override gets a chain that asks `glab` about that host specifically (`--hostname`), shared by the SCM provider and the tracker.
- `glab` output is attributed to hosts by parsing, not by exit code, and only a line that actually names a credential is read as one.
- `GLabTokenSource` memoizes failures for 30s.
- The GitLab tracker no longer refuses to construct when the only credentials are per-host ones.
- Docs: new GitLab SCM and GitLab tracker pages, self-managed host configuration, and the doctor attribution rules.

## Why

Before this, `ao doctor` only ever probed gitlab.com. A self-managed deployment never reaches gitlab.com, so the one GitLab check reported a verdict about a host AO would never call — and it did so by sending whatever token it found to gitlab.com, which could be the internal instance's PAT.

Credential attribution is the core rule the change introduces: **a token is only ever sent to the instance it can be attributed to.** Sending an internal token to gitlab.com discloses it to a third party; sending gitlab.com's token to a self-managed instance hands it to whoever runs that server. All tokens are resolved before any request is issued, and a probe whose credential cannot be attributed to its own host is reported as skipped, with the reason and the fix (`glab auth login --hostname <host>`, or an `AO_GITLAB_HOST_TOKENS` entry).

The daemon changes close the gap the doctor work exposed. An unscoped `glab auth status --show-token` prints every configured host and the first token line wins, which is not necessarily the host being asked about — so a multi-instance `glab` setup could send an internal PAT to gitlab.com while doctor, scoping its own lookup correctly, reported the gitlab.com token as valid.

## How

Reviewer-relevant decisions:

- **Attribution.** A credential counts as attributed to a host when it comes from that host's `AO_GITLAB_HOST_TOKENS` entry, from `glab auth status --hostname <host>`, or from a `glab` status block that names the host as the token's owner. `AO_GITLAB_TOKEN`/`GITLAB_TOKEN` are a global default: the daemon offers them to every instance, but nothing attributes them to any particular one.
- **Doctor is stricter than the daemon, in both directions.** It probes an instance only with a credential attributable to that instance, so two cases are reported instead of probed: a self-managed host reached solely by the global default (that value is most likely gitlab.com's, and probing would hand it to whoever operates the internal server, then report the inevitable 401 as a credential failure), and gitlab.com when an allowlisted host binds that same value (nothing can say which instance it belongs to). Both messages name the remedy. The daemon still offers the global default everywhere at runtime — that is the documented meaning of a global default — doctor just declines to validate it where it cannot attribute it.
- **gitlab.com is strict unconditionally.** Its chain is env vars, then `glab` scoped to gitlab.com; there is no unscoped fallback in the chain itself, whatever the allowlist says. A `glab` too old for `--hostname` still works, because the lookup falls back to the unscoped status block and accepts only the token that block attributes to gitlab.com.
- **Exit codes do not gate parsing.** `glab auth status` exits non-zero when *any* configured instance is unauthenticated, while still printing a usable block for the ones that are — one stale session must not strand every other host. What keeps a diagnostic from being read as a credential is the parsing: only `Token:` / `Token found:` lines name the credential (so `Token expires: <date>` is rejected), and a value containing whitespace is rejected outright, because no GitLab token contains a space and `error: Token: expired or revoked` is prose.
- **`--hostname` is a request, not a guarantee.** A `glab` that ignores it prints the full multi-instance block, where the first token line belongs to whichever host it listed first. Scoped output is therefore read host-attributed first, and the unattributed first-token reading is used only when the output names no instance at all.
- **Unreachable ≠ invalid.** A transport failure against a self-managed host (VPN down, split DNS) is a warning, not a failure; it says nothing about the credential. A 401/403 is still a hard failure, because a probe only ever runs against an instance whose credential could be attributed to it.
- **Empty `host=` entries** in `AO_GITLAB_HOST_TOKENS` are treated as "no override" in both doctor and the daemon — binding `StaticTokenSource("")` would give the host a source that can only ever fail, silently disabling it — and the `gitlab-host-tokens` check reports them, alongside entries whose host is not allowlisted and entries for gitlab.com. An entry the user believes is in use that nothing reads is exactly the misconfiguration that check exists to surface.
- **Failure memoization.** A host-scoped source `glab` cannot satisfy fails on every call, forking a process per token resolution (once per allowlisted host, on every API call and every doctor probe). Failures are memoized for 30s — far shorter than the 5m success TTL, so a fresh `glab auth login` becomes visible quickly — and `InvalidateToken` clears the memo.
- **Doctor probes hosts concurrently.** Each host costs a `glab` lookup plus an HTTPS probe, both bounded by `probeTimeout`; serially, four off-VPN hosts is ~18s of wall clock in a command users expect to be quick. Credential resolution and probes now run in parallel, so doctor waits for the slowest host rather than the sum. The `glab` memo is mutex-guarded; the lock is never held across the subprocess, so two goroutines may race to run the same key, which is harmless for a command that only reads local credential state.
- **Tracker construction.** `trackergitlab.New` still fails fast, but "no token" now means *no instance* can authenticate. A self-managed-only setup keeps its credential in `AO_GITLAB_HOST_TOKENS` and has nothing for gitlab.com; disabling the tracker there left issue intake dead for the only instance the user talks to. gitlab.com lookups now fail per request instead.

Intentional omissions: self-managed API traffic is always `https://<host>/api/v4` — plain HTTP and path-prefixed instances (`https://host/gitlab`) are not supported, and a private CA must be trusted by the OS (AO has no separate CA setting). This matches the provider's existing `clientForHost` behavior and is documented on the GitLab SCM page.

## Testing

- `go build ./...`, `go vet ./internal/...`, `gofmt -l internal/` — clean.
- `go test ./...` — 5976 passed.
- `go test -race` on the GitLab SCM, GitLab tracker, CLI, and daemon packages — clean, including the new concurrent doctor probes.
- New coverage: multi-instance doctor probes, per-host token selection and case-insensitive host matching, empty `host=` overrides (ignored as credentials, reported as misconfiguration), unreachable self-managed hosts, `glab` command-failure diagnostics, both directions of the credential boundary (an internal token never sent to gitlab.com, the global default never sent to a self-managed host), a usable host block read out of a non-zero `glab` run, `Token expires:` never mistaken for a credential, host-scoped token chains in the daemon wiring, `GLabTokenSource` failure memoization and recovery, and tracker construction with per-host credentials only.
- `npx prettier --check` on the touched MDX pages — clean.

## Checklist


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@mcmxcdev this looks like it addresses #4206 — linking for tracking. Adding `Fixes #4206` to the PR body will auto-close it on merge.

i-trytoohard:
@mcmxcdev Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@mcmxcdev Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

mcmxcdev:
I have not received any review on this PR yet, thats why it appears to be "stale". 

Would prefer not to resolve merge conflicts before a review happened.

illegalcall:
Hey @mcmxcdev yes, pushing the review for this. @whoisasx Can you please review this?

PR #4211, opened 2026-08-21
fix(cli): keep daemon-down out of hook stderr

### LINKED ISSUE #4210: bug(cli): transient daemon restart is reported to agents as a hook error on stderr
## Summary

Every hook that fires while the daemon is restarting prints `AO daemon is not running (stale run-file at ...) — start it with \`ao start\`` onto stderr. Hooks run inside the agent's own process, so that text lands in the agent's context as if the tool had failed.

The hook itself is fine — it exits 0, and reconciliation recovers the missed activity from process state. The problem is purely that an expected, transient condition is reported to the agent as an error.

## Why it happens

The desktop app's daemon takeover removes the old run-file and the new daemon writes a fresh one (`frontend/src/main.ts:1261`). Between those two steps the run-file still holds the **previous daemon's dead PID**.

Any CLI call in that window hits `backend/internal/cli/client.go:129`:

```go
if !c.deps.ProcessAlive(info.PID) {
    return fmt.Errorf("AO daemon is not running (stale run-file at %s) — start it with `ao start`", cfg.RunFilePath)
}
```

The error is an untyped `fmt.Errorf`, so `reportHookFailure` (`backend/internal/cli/hooks.go:450`) cannot distinguish "the daemon is mid-restart" from "the daemon returned 500", and unconditionally writes it to `c.deps.Err`.

The same applies to the `info == nil` branch just above it (no run-file at all).

## Observed impact

From one machine's `$AO_DATA_DIR/hooks.log`, **56 occurrences across 9 distinct restart windows** since late June. All hook events are affected:

| hook | count |
|---|---|
| `claude-code session-end` | 27 |
| `agent-process process-exited` | 7 |
| `claude-code user-prompt-submit` | 4 |
| `claude-code session-start` | 3 |
| `claude-code pre-tool-use` | 3 |
| `claude-code post-tool-use` | 3 |
| `claude-code stop` | 3 |
| `claude-code notification` | 3 |
| `opencode user-prompt-submit` | 2 |
| `opencode session-start` | 1 |

One representative window lasted ~35 seconds and hit 6 concurrent sessions across 5 projects. Where agent transcripts recorded the invocation (12 of them), every one shows `"exitCode": 0` — confirming the hooks worked and only the message was spurious. `agent-process process-exited` fires after the agent is gone, so it appears in `hooks.log` only, which does not record an exit code.

Session-start context injection is unaffected: it is read from disk and emitted before any daemon call (`hooks.go:311`), so it still reaches the agent during a restart window.

Both agent adapters that use hooks are affected, so this is not `claude-code`-specific.

## Impact

- Agents receive a spurious "daemon is not running" error in their context and may act on it — re-running work, reporting failure, or asking the user to run `ao start` when the daemon is in fact healthy.
- The message is indistinguishable from a genuine daemon outage, so real problems are harder to spot in `hooks.log`.

## Reproduction

No live daemon required — `ProcessAlive` is already injectable via `Deps` (`backend/internal/cli/root.go:70`):

```go
func TestHooks_DaemonNotRunningStaysOutOfStderr(t *testing.T) {
	t.Setenv("AO_SESSION_ID", "ao-7")
	cfg := setConfigEnv(t)
	if err := runfile.Write(cfg.runFile, runfile.Info{
		PID: 999999, Port: 3001, StartedAt: time.Unix(100, 0).UTC(),
	}); err != nil {
		t.Fatalf("write run-file: %v", err)
	}

	_, errOut, err := executeCLI(t, Deps{
		In:           strings.NewReader(`{"reason":"logout"}`),
		ProcessAlive: func(int) bool { return false },
	}, "hooks", "claude-code", "session-end")
	if err != nil {
		t.Fatalf("hooks must exit 0 when the daemon is down, got: %v", err)
	}
	if errOut != "" {
		t.Errorf("daemon-down must not reach the agent's stderr, got %q", errOut)
	}
}
```

Currently fails with the stale-run-file message on stderr. The `info == nil` variant (no run-file, `ProcessAlive` returning true) reproduces the second branch.

## Suggested fix

Make the condition a sentinel so callers can classify it:

```go
var ErrDaemonNotRunning = errors.New("AO daemon is not running")
```

wrap it with `%w` in both branches of `doJSONPathWithHeadersAndTimeout`, and in `reportHookFailure` skip the stderr write when `errors.Is(cause, ErrDaemonNotRunning)` — still appending to `hooks.log` so the event stays diagnosable. Genuine daemon errors keep reaching stderr unchanged, and interactive commands keep the identical message text.

Happy to raise a PR (branch ready, red test → fix → green, mutation-checked).

axisrow:
Closes #4210.

Hooks run inside the agent's own process, so anything they write to stderr lands in the agent's context. During the desktop app's daemon takeover the run-file briefly points at a dead PID, and every hook firing in that window printed `AO daemon is not running (stale run-file at ...)` at the agent — even though the hook exits 0 and reconciliation recovers the missed activity from process state.

## Change

`doJSONPathWithHeadersAndTimeout` returned an untyped `fmt.Errorf` for both daemon-down branches, so `reportHookFailure` could not tell a restart apart from a real delivery failure.

- Add `ErrDaemonNotRunning` and wrap it with `%w` in both branches (`info == nil`, dead PID). Message text is unchanged, so interactive commands read exactly as before.
- In `reportHookFailure`, skip the stderr write when `errors.Is(cause, ErrDaemonNotRunning)`. The notice still goes to `hooks.log`, so the event stays diagnosable.

Genuine daemon errors (HTTP 5xx, transport failures) keep reaching stderr — covered by the existing `TestHooks_DaemonErrorIsSwallowed`, which is untouched and still asserts stderr is written.

## Tests

`TestHooks_DaemonNotRunningStaysOutOfStderr` is table-driven over both branches and needs no live daemon — `ProcessAlive` is injectable via `Deps`. It asserts exit 0, empty stderr, and a recorded `hooks.log` line.

Verified red before the fix (both sub-cases), green after. Mutation-checked: removing the `errors.Is` guard fails the test.

`go test ./internal/cli/` passes; `gofmt -l .` and `go vet ./...` are clean across the module.

https://claude.ai/code/session_01Q6faenFPv6YaGeihrDeLK5

### DIFF
--- a/backend/internal/cli/client.go
+++ b/backend/internal/cli/client.go
@@ -20,6 +20,12 @@ import (
 // status probe timeout.
 const commandTimeout = 2 * time.Minute
 
+// ErrDaemonNotRunning marks the "no daemon to talk to" condition — either no
+// run-file at all, or one whose PID is dead. Callers that must not treat a
+// daemon restart as a failure (hooks, which fire during the desktop app's
+// daemon takeover) test for it with errors.Is.
+var ErrDaemonNotRunning = errors.New("AO daemon is not running")
+
 // maxDrainedBodyBytes bounds how much of an unused response body the CLI
 // discards for keep-alive reuse without an unbounded read.
 const maxDrainedBodyBytes = 4 << 10
@@ -128,10 +134,10 @@ func (c *commandContext) doJSONPathWithHeadersAndTimeout(
 		return err
 	}
 	if info == nil {
-		return fmt.Errorf("AO daemon is not running — start it with `ao start`")
+		return fmt.Errorf("%w — start it with `ao start`", ErrDaemonNotRunning)
 	}
 	if !c.deps.ProcessAlive(info.PID) {
-		return fmt.Errorf("AO daemon is not running (stale run-file at %s) — start it with `ao start`", cfg.RunFilePath)
+		return fmt.Errorf("%w (stale run-file at %s) — start it with `ao start`", ErrDaemonNotRunning, cfg.RunFilePath)
 	}
 
 	var reader io.Reader = http.NoBody
--- a/backend/internal/cli/hooks.go
+++ b/backend/internal/cli/hooks.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -571,7 +572,13 @@ func (c *commandContext) emitSessionStartContext(agent, event, sessionID string)
 // $AO_DATA_DIR/hooks.log so the failure can be diagnosed after the fact.
 func (c *commandContext) reportHookFailure(agent, event, sessionID string, cause error) {
 	msg := fmt.Sprintf("ao hooks %s %s: %v", agent, event, cause)
-	_, _ = fmt.Fprintln(c.deps.Err, msg)
+	// A daemon that is not running is expected, not a failure: it flaps on every
+	// desktop takeover, and reconciliation recovers the missed activity from
+	// process state. Keep it out of stderr so the notice does not land in the
+	// agent's own context as a spurious error, but still record it below.
+	if !errors.Is(cause, ErrDaemonNotRunning) {
+		_, _ = fmt.Fprintln(c.deps.Err, msg)
+	}
 	dataDir := strings.TrimSpace(os.Getenv("AO_DATA_DIR"))
 	if dataDir == "" {
 		return
--- a/backend/internal/cli/hooks_test.go
+++ b/backend/internal/cli/hooks_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/pricing"
+	"github.com/aoagents/agent-orchestrator/backend/internal/runfile"
 )
 
 type activityCapture struct {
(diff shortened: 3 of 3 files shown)


axisrow:
~~CI note: `build-test` fails on `TestMigration0101AllowsRepeatedProviderOwnershipEpochs`…~~

**Retracted — this was my error, not an upstream one.**

The branch was accidentally cut from a downstream fork branch rather than from `main`, so the PR carried 22 unrelated commits (~60 files) on top of the actual fix. That extra delta included a migration that caused the `internal/storage/sqlite` failure — it was never a pre-existing failure on `main`.

The branch has been rebuilt: the fix is now cherry-picked directly onto `main` (`52fde0279`) and the PR is back to its intended **3 files, +76/-3**. On that clean base `./internal/storage/sqlite` passes, as does `./internal/cli`.

Apologies for the noise — nothing in the review comments below was affected, but the diff you'd have seen before this correction was not the change this PR is about.

axisrow:
Extended the mutation check on the fix — 6 mutants, **0 survivors**:

| # | mutation | killed by |
|---|---|---|
| M1 | invert the guard (`!errors.Is` → `errors.Is`) | both new sub-cases + `TestHooks_DaemonErrorIsSwallowed` |
| M2 | suppress stderr unconditionally (guard → `false`) | `TestHooks_DaemonErrorIsSwallowed` |
| M3 | drop `%w` in the dead-PID branch (`%w` → `%v`) | `…/stale_run-file` |
| M4 | drop `%w` in the `info == nil` branch | `…/no_run-file` |
| M5 | `return` early instead of appending to `hooks.log` | both sub-cases + `TestHooks_HooksLogTruncatesPastCap` |
| M6 | compare against an unrelated sentinel (`io.EOF`) | both sub-cases |

Two things this confirms beyond "the test fails when the fix is removed":

- M2 is caught by the **pre-existing** `TestHooks_DaemonErrorIsSwallowed`, so a genuine daemon error still cannot be silently swallowed.
- M3 and M4 are each killed by a **different** sub-case, so the table's two branches are not redundant.

Working tree restored to the committed state afterwards (`git status` clean, `git diff --stat` empty).

axisrow:
## End-to-end verification against real binaries

The unit test proves the branch; this proves the reported symptom is gone from an actual `ao` build. Both binaries were run side by side against an **isolated** run-file and data dir holding a dead PID — the same shape as the takeover window described in #4210. No live daemon was involved.

Fixture:

```jsonc
// $AO_RUN_FILE
{"pid": <dead-pid>, "port": 3001, "startedAt": "...", "owner": "app"}
```

### The exact reported case — `agent-process process-exited`

```console
$ ao agent-process supervise --session ao-1 --launch ao-launch -- /usr/bin/true
```

**Before** (`52fde0279`, the PR base) — stderr, i.e. straight into the agent's context:

```
ao hooks agent-process process-exited: AO daemon is not running (stale run-file at …) — start it with `ao start`
```

**After** (`446fa798b`) — stderr is empty. The notice is in `hooks.log` instead:

```
<ts> session=ao-1 ao hooks agent-process process-exited: AO daemon is not running (stale run-file at …) — start it with `ao start`
```

Same result for `ao hooks claude-code session-end`: the message disappears from stderr and stays in `hooks.log`.

### Controls — the fix must not swallow anything else

1. **Real daemon error still surfaces.** Pointed the run-file at a live stub returning HTTP 500:
   ```
   ao hooks claude-code session-end: boom (BOOM)
   ```
   Still on stderr, as intended.

2. **Interactive commands are unchanged.** `ao session ls` against the same dead-PID run-file:
   ```
   AO daemon is not running (stale run-file at …) — start it with `ao start`
   ```
   Byte-for-byte identical between the two builds — the message text was preserved deliberately, so only the hook path changed.

3. **Diagnostics are never lost.** `hooks.log` contains the notice in every daemon-down case above.

So: silent for the agent, unchanged for the user, still diagnosable after the fact.

axisrow:
## 🔍 Local review — round 9f45dd3a-c8c5-4359-9770-d6624c584336

Reviewed locally against the rebuilt branch (3 files, +76/-3 on `52fde0279`). Both reviewers ran; no bots pinged.

| Verdict | Reviewer | Finding | Location |
|---|---|---|---|
| — | claude | no findings | — |
| — | codex | no findings (verdict: approve) | — |
| SKIP | codex (earlier round) | Shared failure reporter also covers the supervised-process exit path, not only hook commands | backend/internal/cli/hooks.go:457 |

**Codex summary:** `%w` preserves both daemon-down messages and enables `errors.Is`; no caller depends on the former untyped errors; suppression is limited to hook reporting and still writes `hooks.log`; the table test exercises both the missing-run-file and dead-PID branches.

### On the one earlier finding — triaged SKIP, with reasoning

An earlier round flagged that `reportHookFailure` is shared, so the suppression also covers `reportSupervisedExit` in `agent_process.go`, which is not a hook command. The premise is correct, but the suggested remedy — keep stderr for the agent-process path — **would reintroduce the exact bug this PR fixes**: the reported error was `ao hooks agent-process process-exited: AO daemon is not running (stale run-file …)`, i.e. that very call site. `agent-process supervise` wraps the agent inside its own pane and shares stderr with it (`child.Stderr = c.deps.Err`), so that output lands in the agent's context, not in a separate operator console.

The second half — that no diagnostic might survive if the log write fails — does not apply to this path in practice: the session manager exports `AO_DATA_DIR` to every spawned agent, so the `dataDir == ""` early return is unreachable for a supervised process. `appendHooksLog` dropping errors is a pre-existing, documented choice this PR does not change.

**Totals:** 0 FIX, 1 SKIP (with rationale above), 0 UNVERIFIED.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

axisrow:
Closing as superseded: current main's daemonUnavailableError (cause: ErrDaemonNotRunning) achieves the same goal — daemon-down no longer leaks into hook stderr. This branch applies as empty on main (patch-id verified).

PR #4218, opened 2026-08-21
feat: add provider-neutral subscription usage

nikhilachale:
## Summary

Adds a provider-neutral subscription usage system for Codex, Claude, and future coding-agent adapters. The feature tracks provider-reported plan capacity, current rate-limit windows, credits/balances when available, bounded usage history, refresh failures, and threshold alerts without estimating prompts or tasks from token totals.

### Provider adapter layer

- Introduces a neutral quota model in `backend/internal/domain/quota.go` for provider accounts, limits, balances, capabilities, history points, and alerts.
- Adds quota ports so adapters can publish snapshots without the service depending on provider-specific payloads.
- Maps Codex app-server `account/rateLimits/read` and rate-limit update payloads into neutral quota snapshots, including:
  - account-level rate limits
  - model-scoped limits from `rateLimitsByLimitId`
  - Codex credits from `rateLimits.credits`
  - reset credits from `rateLimitResetCredits.availableCount`
  - duplicate-window collapse when Codex reports identical primary/secondary windows
- Maps Claude ACP usage metadata and structured plan usage into neutral quota snapshots, including:
  - partial push-based usage updates from ACP metadata
  - full sessionless reads through `ao-claude-plan-usage.mjs`
  - five-hour, seven-day, OAuth app, Opus, Sonnet, and model-scoped windows when Claude reports them
- Keeps provider-specific parsing inside the relevant adapter packages so future adapters can add quota support by implementing the same neutral contracts.

### Quota service

- Adds `backend/internal/service/quota` to merge provider observations, coalesce refreshes, cache fresh reads, derive severity, and emit transition-only alerts.
- Supports complete and partial snapshots so provider pushes do not erase previously known buckets.
- Refreshes per provider account rather than per session, which avoids multiplying provider calls when multiple conversations share one subscription.
- Registers daemon-owned providers and refreshes stale accounts on startup, periodically, on Plan Usage page entry, and after relevant idle transitions.
- Persists a first refresh failure for registered provider accounts so installed-but-broken providers appear as actionable error cards instead of disappearing from the page.

### Storage

- Adds SQLite migration `0102_provider_quota.sql` with:
  - `quota_accounts` for latest provider/account state
  - `quota_limits` for latest limit buckets
  - `quota_balances` for latest credits/balances
  - `quota_history` for bounded trend data
  - `quota_alerts` for threshold transitions
- Inserts history only when the meaningful quota state changes.
- Compacts history automatically:
  - last 24 hours: at most one point per minute
  - 24 hours to 30 days: at most one point per hour
  - 30 to 90 days: at most one point per day
  - older than 90 days: deleted

### API and frontend

- Adds daemon API endpoints for:
  - listing provider quota snapshots
  - fetching one provider account
  - refreshing one provider account
  - reading bounded history
  - reading recent quota alerts
- Regenerates OpenAPI and frontend TypeScript schema.
- Adds a Plan Usage page in the Electron UI with:
  - generic provider cards
  - current remaining/used values
  - compact credits and balances rows when reported
  - retry action for failed provider refreshes when the account supports reads
  - AO-aligned settings-style layout
- Adds sidebar navigation and notification integration for quota alerts.

## Scalability

- Provider-neutral domain and API keep the UI independent of Codex/Claude-specific payloads.
- Capability flags let the UI render whatever a provider can truthfully report without hardcoding provider names.
- Reads are coalesced with `singleflight`, and fresh account snapshots are reused.
- Storage is account-level, not session-level, so many sessions can share one provider subscription state.
- History is bounded and compacted; this is cached operational state, not an append-only billing ledger.

## Verification

- `npm run lint` was run; the full parallel suite hit unrelated adapter timing flakes, and each failed package passed individually.
- `go test ./internal/service/quota ./internal/storage/sqlite/store ./internal/storage/sqlite ./internal/adapters/chatdriver/acp ./internal/httpd/controllers`
- `go test ./internal/adapters/agent/fake -count=1`
- `go test ./internal/adapters/agent/kilocode -count=1`
- `go test ./internal/adapters/agent/opencode -count=1`
- `go test ./internal/adapters/agent/primeagent -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs`
- Focused Plan Usage frontend test: `PlanUsagePage.test.tsx`
- `git diff --check`

## Notes

- AO only displays provider-reported values. It does not infer remaining prompts, tasks, or spend from token totals.
- Credits are shown only when the provider reports stable credit/balance fields.
- Claude structured usage is accessed through the ACP SDK's usage query rather than by sending `/usage` into a visible chat conversation.
- Quota history remains persisted and compacted for future trend/alert use, but this PR does not currently render an observed-usage history sparkline in the simplified Plan Usage card.


nikhilachale:
Screenshot 2026-08-21 at 11 34 08
Screenshot 2026-08-21 at 11 34 27

frontend/src/renderer/components/usage/PlanUsagePage.tsx
```diff
@@ -0,0 +1,137 @@
...
+					{quota.planType ? <span className="rounded-full border border-border bg-raised px-2 py-0.5 text-[10px] font-semibold uppercase tracking-wide text-passive">{quota.planType}</span> : null}
+				</CardTitle>
+				<CardDescription className="flex flex-wrap items-center gap-x-2 gap-y-1 text-xs">
+					<span className={severityText[quota.severity]}>{statusLabel(quota.severity, t)}</span>
+					<span aria-hidden="true">·</span>
+					<span>{freshnessLabel(quota.freshness, quota.observedAt, t)}</span>
+					{quota.completeness === "partial" ? <><span aria-hidden="true">·</span><span>{t("planUsage.partial")}</span></> : null}
+				</CardDescription>
+			</CardHeader>
+			<CardContent className="flex flex-col gap-4 px-4 py-4">
```
illegalcall, line 92: `quota.balances` is returned by this PR's API and populated for Codex credits/reset credits, but this card never reads it, so reported balances are invisible even though the PR body says they are shown. Please render a compact balances row only when `quota.balances.length > 0` (including the `unlimited` case), or update the PR scope if balances are intentionally deferred. Keeping the section conditional preserves the simplified card without adding empty UI.

frontend/src/renderer/components/usage/PlanUsagePage.tsx
```diff
@@ -0,0 +1,137 @@
...
+				{quota.refreshError ? <p className="rounded-md bg-status-exited/5 px-3 py-2 text-xs text-status-exited" role="alert">{t("planUsage.lastRefreshError", { error: quota.refreshError })}</p> : null}
```
illegalcall: This surfaces the failure but gives the user no recovery action. The per-account refresh endpoint already exists, while the corresponding hook/button was removed in 52d146b, so the only recovery is remounting the page or waiting for the daemon timer. Please add a small `Retry` action next to this error that calls the account refresh endpoint, disables while pending, and reports retry failure inline. Normal successful cards can continue relying on invisible automatic refresh.

backend/internal/storage/sqlite/queries/quota.sql
```diff
@@ -0,0 +1,183 @@
...
+    supports_history = MAX(quota_accounts.supports_history, excluded.supports_history),
+    supports_credits = MAX(quota_accounts.supports_credits, excluded.supports_credits),
+    supports_spend_limits = MAX(quota_accounts.supports_spend_limits, excluded.supports_spend_limits),
+    completeness = excluded.completeness,
+    observed_at = excluded.observed_at,
+    last_refresh_error = ''
+WHERE excluded.observed_at >= quota_accounts.observed_at;
+
+-- name: RecordQuotaRefreshFailure :exec
+UPDATE quota_accounts
```
illegalcall, line 22: Before the first successful observation there is no `quota_accounts` row, so this `UPDATE` affects zero rows without error. Startup/auth/helper failures are therefore discarded, `RefreshAll` returns an empty list, and an installed-but-broken provider looks merely absent. Please establish a provider/account row before the first read, or use a presence-aware upsert for failures and check affected rows. The presence check matters: Codex and Claude refreshers are registered daemon-wide, so blindly upserting would create noisy error cards for providers that are not installed.

backend/internal/adapters/chatdriver/acp/client.go
```diff
@@ -740,24 +740,124 @@ func claudeRateLimits(meta map[string]any) *ports.ChatRateLimits {
...
+// response into AO's provider-neutral quota model for the daemon's sessionless
+// collector.
+func NormalizeClaudePlanUsage(value map[string]any) *ports.ChatRateLimits {
+	if value == nil {
+		return nil
+	}
+	now := time.Now().UTC()
+	snapshot := &domain.QuotaSnapshot{
+		Provider: "claude", AccountID: "default", Completeness: domain.QuotaComplete,
+		Capabilities: domain.QuotaCapabilities{SupportsSubscribe: true}, ObservedAt: now,
```
illegalcall: A successful value here came from the daemon-owned sessionless read, yet the resulting API advertises `supportsRead: false`. That is inconsistent with the implemented refresher and would hide a capability-gated retry control. Please set `SupportsRead: true` for this complete snapshot and cover it in `TestClaudePlanUsageNormalizesStructuredSDKResponse`. If `SupportsHistory` means AO's persisted observation history, set that too; otherwise clarify that flag's narrower provider-native meaning.

illegalcall (commented):
I re-audited this after tracing the runtime paths and commit history. These inline comments are limited to problems with concrete user/API impact. One additional scope point: the PR body still promises an observed-usage history sparkline, but commit 16a4514 deliberately removed it. If history is intentionally deferred, please update the PR body; otherwise restore it behind a progressive-disclosure control such as “View history.”

whoisasx:
## End-to-end provider-neutral subscription usage flow

```mermaid
flowchart TB
  subgraph FE["Electron / React frontend"]
    Page["PlanUsagePage"]
    ListHook["useProviderQuota<br/>GET /api/v1/usage/plans<br/>initial load + every 60s"]
    AllHook["useRefreshAllProviderQuota<br/>POST /api/v1/usage/plans/refresh<br/>when page mounts"]
    ErrorCard["Provider error card"]
    RetryHook["useRefreshProviderQuota<br/>POST /api/v1/usage/plans/{provider}/accounts/{accountId}/refresh"]
    Cards["Provider cards<br/>status + limits + balances<br/>unlimited + partial/stale state"]

    Page --> ListHook
    Page --> AllHook
    ErrorCard -- "Retry; disabled while pending" --> RetryHook
    ListHook --> Cards
    Cards -. "refreshError" .-> ErrorCard
  end

  subgraph DAEMON["Loopback daemon API"]
    Controller["QuotaController<br/>list | refresh all | refresh account"]
    Startup["Daemon startup"]
    Timer["5-minute timer"]
    Idle["Codex/Claude worker becomes idle"]
  end

  subgraph SERVICE["Quota service"]
    Registered["Daemon-wide registered refreshers<br/>Codex:default + Claude:default"]
    Stale["RefreshRegisteredIfStale<br/>reuse observations inside FreshWindow"]
    Refresh["Refresh(provider, account)<br/>singleflight per account"]
    Existing{"Durable account row exists?"}
    Present{"Provider installed?<br/>QuotaAccountPresent"}
    Seed["Create stale, partial presence row<br/>before the first provider read"]
    Skip["Unsupported / absent<br/>no row and no noisy UI card"]
    Normalize["Normalize + merge snapshot<br/>compute transition alerts"]
    Failure["Persist refreshError with checked UPDATE"]
    List["List durable snapshots"]
  end

  subgraph PROVIDERS["Provider adapters"]
    Codex["Codex app-server<br/>account/rateLimits/read"]
    Claude["Claude Code helper<br/>plan usage read"]
    Live["Live chat/provider observations"]
  end

  subgraph DB["SQLite durable model"]
    Store[("quota_accounts<br/>quota_limits<br/>quota_balances<br/>quota_history<br/>quota_alerts")]
  end

  AllHook --> Controller
  RetryHook --> Controller
  Controller -- "refresh all" --> Stale
  Controller -- "refresh one" --> Refresh
  Controller -- "GET list" --> List

  Startup --> Stale
  Timer --> Stale
  Idle --> Stale
  Registered --> Stale
  Stale -- "stale or never observed" --> Refresh
  Stale -- "fresh" --> List

  Refresh --> Existing
  Existing -- "yes" --> Codex
  Existing -- "yes" --> Claude
  Existing -- "no" --> Present
  Present -- "no" --> Skip
  Present -- "yes" --> Seed
  Seed --> Store
  Seed --> Codex
  Seed --> Claude

  Codex -- "successful snapshot" --> Normalize
  Claude -- "successful snapshot" --> Normalize
  Codex -- "auth/startup/helper failure" --> Failure
  Claude -- "auth/startup/helper failure" --> Failure
  Live --> Normalize

  Normalize --> Store
  Failure --> Store
  Store --> List
  List --> Controller
  Controller -- "ProviderQuotaResponse[]" --> ListHook
  Controller -- "retry result / inline failure" --> RetryHook
```

Key behavior:
- **Absent providers stay invisible:** presence is checked before creating the first account row, so daemon-wide refresher registration does not produce noisy cards.
- **Installed but broken providers are visible and recoverable:** a stale presence row is established first, refresh failures are persisted, and the card offers an inline **Retry** action.
- **Successful reads converge in one model:** automatic refreshes, manual retries, and live observations all normalize/merge into the same durable snapshot, history, balances, and alert pipeline.

whoisasx:
## Simple architecture

```mermaid
flowchart LR
  User([User]) --> UI["Electron + React<br/>Plan Usage UI"]
  UI <-->|"Loopback HTTP API"| API["Go daemon<br/>QuotaController"]

  API --> Service["Quota service<br/>refresh, normalize, merge"]
  Service --> Codex["Codex adapter"]
  Service --> Claude["Claude adapter"]
  Codex --> CodexProvider["Codex app-server"]
  Claude --> ClaudeProvider["Claude Code helper"]

  Service <-->|"snapshots, errors,<br/>history and alerts"| SQLite[(SQLite)]

  Scheduler["Startup + 5-minute timer<br/>+ worker-idle trigger"] --> Service
  SQLite --> API
```

The frontend remains a thin supervisor: the Go daemon owns provider discovery, refresh coordination, normalization, failure handling, and durable storage.

whoisasx:
## Proposed simplified subscription-usage approach

After walking through the current end-to-end flow, we think this can be simplified by making the existing agent inventory/auth state the source of truth instead of independently discovering quota accounts.

### Core approach

1. **Reuse existing agent state** to determine whether Codex/Claude is installed and whether authentication is configured.
2. **Fetch subscription usage on demand** through the corresponding provider adapter.
   - Use a short in-memory TTL and singleflight deduplication to avoid repeatedly starting helpers or hitting provider APIs.
   - A manual refresh bypasses the TTL.
3. **Persist only recovery state** needed when a live read fails:
   - last successful snapshot
   - last successful timestamp
   - last attempt timestamp
   - last error
4. Treat the provider call as the final authority. Existing auth state is an eligibility hint because credentials can expire between the auth check and the usage request.

```mermaid
flowchart LR
  User([User]) --> Page["Plan Usage page"]
  Page --> API["GET /api/v1/usage/plans"]

  API --> Inventory["Existing agent inventory<br/>installation + auth state"]
  Inventory --> Eligible{"Installed and eligible?"}
  Eligible -- "No" --> Hidden["Do not request usage<br/>no noisy provider card"]
  Eligible -- "Yes" --> Usage["Usage facade<br/>TTL cache + singleflight"]

  Usage --> Adapter{"Provider adapter"}
  Adapter --> Codex["Codex app-server"]
  Adapter --> Claude["Claude Code helper"]

  Codex --> Result{"Live read succeeds?"}
  Claude --> Result
  Result -- "Yes" --> Normalize["Normalize provider-neutral usage"]
  Normalize --> Recovery[("Minimal recovery state<br/>last good snapshot + timestamps")]
  Normalize --> Response["Current limits and balances"]

  Result -- "No" --> Failure["Record last error"]
  Failure --> Recovery
  Recovery -- "Return last good data<br/>with stale/error status" --> Response

  Response --> Page
  Page -- "Manual refresh bypasses TTL" --> Usage
```

### Resulting simplifications

This would remove the need for:

- a separate quota-account discovery mechanism
- daemon-wide quota refresher registration
- provider-presence placeholder rows
- the global five-minute quota refresh timer
- a page-load `GET` followed by a separate refresh-all `POST`

### History and alerts

- If the immediate product requirement is only the current Plan Usage view, the unused 90-day history can be omitted.
- If proactive OS threshold notifications remain in scope, we can preserve them with a much smaller persisted transition state (previous severity/reached value plus emitted alert identity) rather than retaining the full usage history subsystem.

@nikhilachale, could you please check whether this simplified direction matches the intended scope and provider behavior?

@illegalcall, when you have a moment, could you please review this proposal as well—particularly the live-read, caching, and minimal failure-state boundaries? Thank you.

nikhilachale:
I agree with this simpler direction.

For the current Plan Usage page, the provider should be the source of truth. AO should not store changing quota/history data as if it owns it.

I suggest we do this:

- Use existing agent install/auth state only to decide if Codex or Claude should be checked.
- Fetch usage live from the provider adapter.
- Add a short daemon-side cache so we do not call the provider too often.
- Use singleflight so duplicate refreshes are merged.
- Manual refresh should skip the cache.
- Store only fallback data:
  - last successful usage snapshot
  - last successful time
  - last attempt time
  - last error
- If live fetch fails, show the last good data as stale with the error.
- If the provider is not installed or not authenticated, do not show its card.
- Skip full history, sparkline, and alert storage unless we decide alerts are required now.

So yes, I’m aligned with removing quota discovery, placeholder rows, the 5-minute refresh timer, and unused 90-day history for this PR.

@illegalcall please confirm if this smaller scope looks right.

nikhilachale:
done in another PR

PR #4221, opened 2026-08-21
feat: add harness installer settings

nikhilachale:
## Summary

- add a dedicated Global Settings > Harness section for all 27 supported agent CLIs
- detect installed harnesses and expose one-click Install actions, plus Reinstall only for methods with a verified headless recipe
- keep package managers preferred, with fixed first-party vendor installers as automatic fallbacks where supported
- run installs as durable asynchronous daemon jobs, stream bounded diagnostics, and verify the exact harness after completion
- keep this installation-only: no sign-in or authorization UI
- restack the feature on current main while preserving Cloudflared installer behavior

## One-click official installers

- download the complete script over HTTPS into `<AO_DATA_DIR>/installers/tmp` before execution
- enforce a 30-second download timeout, a 4 MiB response limit, at most five HTTPS-only redirects, and non-2xx rejection
- record the source URL and SHA-256 digest in job diagnostics
- execute the saved file with a fixed allowlisted shell or PowerShell argv, closed stdin, and bounded runtime
- remove the per-job script directory after success, failure, timeout, or cancellation
- include the documented Kimchi release installers alongside the existing official installer registry

Package managers remain the recommended method when viable. Script-only native harnesses receive the same one-click Install action. Reinstall is offered only for methods with a verified headless reinstall recipe; other installed harnesses show first-party Instructions. Platforms without a supported native recipe also continue to show first-party Instructions.

## Safety

- recipes, URLs, interpreters, and documentation links are daemon-owned; the renderer submits only a harness id and method id
- no `sudo`, `curl | shell`, `sh -c`, or PowerShell `-Command` execution
- official scripts are restricted to macOS, Linux, and Windows recipes explicitly present in the registry
- unsupported operating systems fail closed to manual vendor instructions
- installer output, download size, redirects, and execution time are bounded
- concurrent installs for the same harness are rejected
- install mutations are loopback-only and return 404 through the authenticated LAN listener
- migration 0123 includes compatibility repair for development profiles that applied the earlier 0119/0120 installer migration

## Testing

- backend HTTP, SQLite migration, system installer, daemon, and generated-spec suites
- isolated reruns of repository timing-sensitive tests
- `npm test -- --run src/renderer/components/settings/HarnessSettingsSection.test.tsx`: 13 passed
- `npm run typecheck`
- `git diff --check`

## Scope

This PR covers harness discovery and installation only. Authentication and sign-in flows remain separate.


illegalcall:
@nikhilachale Attach UI screenshots here

nikhilachale:
Screenshot 2026-08-30 at 22 00 41
Screenshot 2026-08-30 at 22 02 56

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -261,14 +338,23 @@ func (s *Service) run(argv []string, job *Job) {
 		job.Error = runErr.Error()
 		return
 	}
-	if path, err := s.executables.LookPath(string(job.Target)); err != nil || path == "" {
-		job.Status = StatusFailed
-		job.Error = fmt.Sprintf("install command finished but %s is still not in PATH", job.Target)
-		return
+	if !IsAgentTarget(job.Target) {
```
illegalcall, line 860: This makes `succeeded` mean only that the installer process exited 0 for every harness. The decisive exact-harness probe lives in a mounted React effect, so closing Settings, losing the renderer, or calling the API directly skips verification and leaves the daemon reporting a false success. Please inject an adapter-backed verifier into the daemon job and model `installing -> verifying -> succeeded/failed`; the UI should only observe that state. Do not probe authentication here, only binary resolution and a bounded launch/version check.

backend/internal/service/systeminstall/agentplans.go
```diff
@@ -0,0 +1,235 @@
...
+func (s *Service) planShellInstaller(target Target, url, shell string) Plan {
+	if _, err := s.executables.LookPath("curl"); err != nil {
+		return Plan{Target: target, Unsupported: true, Method: "official-installer", Reason: "curl was not found on PATH."}
+	}
+	if _, err := s.executables.LookPath(shell); err != nil {
+		return Plan{Target: target, Unsupported: true, Method: "official-installer", Reason: fmt.Sprintf("%s was not found on PATH.", shell)}
+	}
+	return Plan{
+		Target: target, Method: "official-installer",
+		Command: []string{shell, "-c", fmt.Sprintf("curl -fsSL %s | %s", url, shell)},
```
illegalcall: This pipeline is not failure-safe: without pipefail, a failed `curl` that produces no script can still leave the final shell exiting 0. A fixed URL also does not make the fetched, mutable body code-owned. Please download into a private temporary file, fail on the download status, validate a checksum/signature where the vendor provides one, then execute with an explicit no-stdin/noninteractive policy. If the installer requires a TTY or admin access, return a manual plan rather than adding sudo.

backend/internal/service/systeminstall/agentplans.go
```diff
@@ -0,0 +1,235 @@
...
+	if _, err := s.executables.LookPath("bun"); err != nil {
+		return Plan{Target: target, Unsupported: true, Method: "bun", Reason: "Bun was not found on PATH."}
+	}
+	return Plan{Target: target, Method: "bun", Command: []string{"bun", "install", "-g", pkg}}
+}
+
+func (s *Service) planPythonPip(target Target, pkg string) Plan {
+	for _, python := range []string{"python3", "python"} {
+		if _, err := s.executables.LookPath(python); err == nil {
+			return Plan{Target: target, Method: "pip", Command: []string{python, "-m", "pip", "install", pkg}}
```
illegalcall: Finding `python3` is not enough to advertise an automatic macOS install. The interpreter may be too old, externally managed, or backed by a non-writable prefix; Vibe currently requires Python 3.12+. Prefer `uv tool install`/pipx for standalone CLIs. If pip remains a fallback, preflight the required version and destination writability and downgrade to manual instructions on failure. The npm plans need the equivalent writable-global-prefix/version check rather than trying sudo after EACCES.

backend/internal/service/systeminstall/agentplans.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+	return Plan{Target: target, Unsupported: true, Method: "pip", Reason: "Python was not found on PATH."}
+}
+
+func firstAvailable(plans ...Plan) Plan {
```
illegalcall: On macOS, curl plus sh/bash are normally present, so the official script wins permanently. If that script later fails because of a vendor outage, unsupported OS version, or permission problem, Retry selects the identical method again and the Homebrew/npm alternatives are never reachable. Please expose the viable methods to the user or make fallback selection failure-aware, while still keeping all recipes server-owned.

illegalcall (commented):
macOS-focused review: the no-sudo policy is the right baseline, but I think the daemon-owned verification, noninteractive execution, permission preflight, and job recovery issues below should be addressed before merge. The key invariant should be: a job is not succeeded until AO can resolve and launch the exact installed harness as the current user.

frontend/src/renderer/components/settings/HarnessSettingsSection.tsx
```diff
@@ -0,0 +1,236 @@
...
+	const runningAgentKey = useMemo(
+		() => AGENT_OPTIONS.filter((agentId) => installStates[agentId]?.job?.status === "running").join(","),
+		[installStates],
+	);
+	const succeededAgentKey = useMemo(
+		() => AGENT_OPTIONS.filter((agentId) => installStates[agentId]?.job?.status === "succeeded").join(","),
+		[installStates],
+	);
+
+	useEffect(() => {
```
illegalcall, line 92: Polling is derived only from component-local install state. If Settings is closed and reopened while a daemon job is running, no status request is made, the progress disappears, and the row offers Install again. Please hydrate current/recent jobs from the daemon on mount (ideally with one batch endpoint), surface polling failures, and define an interrupted state for daemon restarts before allowing a reinstall.

frontend/src/renderer/components/settings/HarnessSettingsSection.tsx
```diff
@@ -0,0 +1,236 @@
...
+					const failed = rowJob?.status === "failed" || rowJob?.status === "unsupported" || Boolean(installState?.error);
+					const running = rowJob?.status === "running";
+					const rowVerifying = installState?.verifying === true;
+					const statusId = `harness-agent-${agentId}-status`;
+					return (
+						<div className="settings-row-bar min-h-14 gap-3" data-agent={agentId} key={agentId}>
+							<AgentAvatar className="size-7 shrink-0" decorative provider={agentId} />
+							<div className="min-w-0 flex-1">
+								<p className="truncate text-sm font-medium text-settings-label" id={`harness-agent-${agentId}`}>{agentLabel(agentId)}</p>
+								<p className={cn("truncate text-xs text-settings-muted", failed && "text-error")} title={failed ? (installState?.error ?? rowJob?.error) : plan?.reason}>
```
illegalcall: The useful npm/Homebrew/script error is captured in `Job.output`, but this row exposes only `Job.error`, which is commonly just `exit status 1`. On macOS that hides EACCES, PATH, Gatekeeper, architecture, and missing-prerequisite diagnoses. Please add expandable/copyable diagnostics including method, expected destination, captured output, and separate `Verify again` versus `Reinstall` actions.

backend/internal/service/systeminstall/agentplans.go
```diff
@@ -0,0 +1,235 @@
...
+			"https://ampcode.com/manual")
+		if s.goos == "darwin" {
+			return withDocs(firstAvailable(preferred, s.planBrew(target, "ampcode/tap/ampcode"), s.planNPM(target, "@ampcode/cli")), "https://ampcode.com/manual")
+		}
+		return withDocs(firstAvailable(preferred, s.planNPM(target, "@ampcode/cli")), "https://ampcode.com/manual")
+	case TargetAuggie:
+		return withDocs(s.planNPM(target, "@augmentcode/auggie"), "https://docs.augmentcode.com/cli/overview")
+	case TargetDroid:
+		preferred := s.officialByOS(target,
+			"https://app.factory.ai/cli", "sh",
```
illegalcall: The current Factory installer at this URL sends `pkill -KILL -x droid` before replacing the binary. That is a surprising lifecycle side effect inside AO: it can kill a Droid process outside the install job and bypass AO session shutdown. Please block this method while AO has an active Droid session, or require an explicit warning/confirmation and coordinate shutdown through the daemon lifecycle boundary.

illegalcall:
Follow-up review on current head 27fa5244: I am still holding approval for these macOS-relevant blockers:

1. Capability probes do not accept the caller context and rerun subprocess checks for each recipe. A cancelled Settings request can therefore leave many sequential processes running. Please propagate request context and snapshot capabilities once per request.
2. npm viability checks PATH, prefix, and writability, but never validates supported Node/npm versions. HomebrewPackageInstalled also collapses errors/timeouts into "not installed", which can proceed with an install after a failed preflight; please fail closed.
3. The Kimchi npm recipe is not valid: @kimchi-dev/cli currently returns E404, while the official installation docs use Homebrew or the release installer: https://docs.kimchi.dev/docs/coding-getting-started
4. If the terminal-state database upsert fails, the status APIs can keep preferring the stale active DB row and the UI may poll forever. The terminal failure needs to remain observable/recoverable.

CI is green, but I do not think this head is approval-ready until these are addressed.

illegalcall:
Fresh macOS-focused re-review of current head `fc1b9d849` (supersedes my findings on the older `27fa5244` head).

The earlier capability-snapshot, Node/npm probing, Homebrew failure, Kimchi recipe, and persistence-visibility concerns appear fixed, and all current checks are green. I am still holding approval for these confirmed user-facing issues:

1. **Pi's official fallback fails in the condition where AO most needs it.** `agentMethodPlans` offers `https://pi.dev/install.sh` after the npm method becomes unavailable. The current Pi script works headlessly when Node >=22.19/npm already exist, but when Node is absent/old it calls `install_node_npm_interactive`, requires `/dev/tty`, and exits if no terminal is available. AO executes installers with closed stdin and the canonical macOS desktop daemon normally has no controlling TTY. Thus a Mac with missing/old Node is shown an automatic fallback that deterministically cannot repair the prerequisite. Evidence: [AO Pi plan](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/service/systeminstall/agentplans.go#L77-L81), [AO non-interactive runner](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/adapters/systemexec/systemexec.go#L52-L64), [current Pi installer](https://pi.dev/install.sh).

2. **Kiro Reinstall cannot replace an existing macOS installation headlessly.** The current vendor script prompts with `read -r response < /dev/tty` whenever the Kiro app already exists in `/Applications`. AO provides no PTY, so the UI's Reinstall action fails in the normal existing-install case. The script's `--force` option skips an earlier Amazon Q check but does not bypass this app-replacement prompt. Evidence: [AO Kiro plan](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/service/systeminstall/agentplans.go#L289-L293), [current Kiro installer](https://cli.kiro.dev/install). Also, the script writes `/Applications`; managed non-admin Macs will fail there without elevation, which AO intentionally does not add.

3. **Vibe's Reinstall action does not actually reinstall.** Both install and reinstall resolve to `uv tool install mistral-vibe` or `pipx install mistral-vibe`. For an existing tool, uv can exit successfully without rebuilding it (so AO verifies a no-op as success), while pipx can refuse because the environment already exists. Use explicit reinstall semantics (`uv tool upgrade --reinstall`/force as appropriate and `pipx reinstall`) when the UI says Reinstall. Evidence: [AO Vibe recipes](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/service/systeminstall/agentplans.go#L126-L127), [uv tool behavior](https://docs.astral.sh/uv/concepts/tools/), [pipx reinstall](https://pipx.pypa.io/latest/how-to/manage-installed-apps.html). The endpoint currently has no install-vs-reinstall operation, so please also check the same UI promise for the other package-manager recipes.

4. **One global Node 22.19 floor hides valid npm methods.** The code applies Pi's strictest engine constraint to every npm recipe. Current package metadata permits Codex on Node >=16, Droid/Auggie >=20, and Claude/Qwen/Autohand >=22.0; only Pi needs >=22.19. This is especially visible for npm-only Auggie: a valid Node 20 installation is reported unavailable. If AO deliberately wants one stricter support floor, document that product policy and stop describing every vendor-viable package-manager method as available; otherwise make the minimum target-specific. Evidence: [global floor](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/service/systeminstall/systeminstall.go#L1107-L1157).

Two lower-severity follow-ups, not independent approval blockers:

- `transitionAgentJob`/`finishAgentJob` persist with `context.Background()`. This cannot indefinitely block daemon exit because `Close` has a deadline, but a blocked writer can consume the installer drain budget and race later DB shutdown. Use a bounded lifecycle/best-effort persistence context. [Code](https://github.com/Untrivial-ai/agent-orchestrator/blob/fc1b9d8492815f9969bd245dc6b1f33230a3de5f/backend/internal/service/systeminstall/systeminstall.go#L882-L915).
- `agentMethodPlans` and the legacy `planAgent` cascade duplicate recipe data and already disagree on ordering. This is maintenance debt rather than a current product failure; removing the dead overlap or sharing registry data would prevent future drift.

Requested macOS validation before approval: exercise fresh install and existing-install Reinstall from a packaged/Finder-launched AO app (no controlling TTY), including a Mac with missing/old Node and a non-admin account.

illegalcall:
Follow-up on current head `2ea1059db`: the earlier Kiro reinstall blocker has been changed, but the replacement still needs validation before approval.

1. **Kiro Reinstall relies on an undocumented CLI contract.** AO now runs `kiro-cli update --non-interactive --force` instead of the interactive vendor installer. Kiro's current command reference documents `--non-interactive` and `--relaunch-dashboard` for `kiro-cli update`, but not `--force`: https://kiro.dev/docs/reference/cli-commands/#kiro-cli-update

   There is first-party evidence that the hidden flag performed a true same-version reinstall in Kiro CLI 2.4.1: https://github.com/kirodotdev/Kiro/issues/8887. That makes this a real historical behavior, not an invented flag, but it is not part of the current documented contract. The AO tests only assert the generated argv; they do not exercise a current packaged Kiro installation. Please either validate this command against the current Kiro release on packaged macOS, use a documented headless reinstall operation, or mark Kiro reinstall unavailable and show Instructions. The vendor installer's own `--force` is not a fallback on macOS because its existing-app replacement path still reads confirmation from `/dev/tty`.

   AO implementation: https://github.com/Untrivial-ai/agent-orchestrator/blob/2ea1059db13e492f70e16b5a932f89d20622b540/backend/internal/service/systeminstall/agentplans.go#L259-L266

2. **The PR description overstates reinstall coverage.** It says the Harness section exposes one-click Install or Reinstall actions for all 27 harnesses, but AO disables reinstall for every other `official-installer` recipe. Installed script-only harnesses therefore show Instructions rather than Reinstall. That is the safe behavior, but if intentional, please narrow the PR description to say that one-click reinstall is available only for methods with a verified headless recipe.

   AO implementation: https://github.com/Untrivial-ai/agent-orchestrator/blob/2ea1059db13e492f70e16b5a932f89d20622b540/backend/internal/service/systeminstall/agentplans.go#L288-L292

illegalcall (approved):

illegalcall:
Follow-up on current head `02b055fd8`: both remaining review items are resolved.

- Kiro no longer relies on the undocumented `kiro-cli update --non-interactive --force` behavior. The unverified headless reinstall special case was removed, so an installed Kiro now follows the safe Instructions path.
- The PR description now accurately says Reinstall is offered only for methods with a verified headless recipe.

All ten checks on this head are green, including the native macOS job. I found no remaining approval blocker in the installer changes.

PR #4224, opened 2026-08-21
fix(tmux): resume an agent under a non-default tmux base-index

### LINKED ISSUE #4222: [Bug]: resuming an agent hard-fails with `can't find pane: 0` under a non-default tmux base-index
## Summary

Resuming an agent fails permanently — with `can't find pane: 0` — for any user whose `tmux.conf` sets `base-index 1`. The tmux runtime builds its respawn target as the literal `<session>:0.0`, but tmux honours `base-index` when `new-session` picks the window index, so AO's own session has no window `0`.

`base-index 1` / `pane-base-index 1` is one of the most common lines in a hand-written `tmux.conf` (it puts window 1 under the `1` key), so this is not an exotic configuration. It affects macOS and Linux equally.

## Steps to reproduce

1. Put this in `~/.config/tmux/tmux.conf` (or `~/.tmux.conf`) and restart the tmux server:

   ```tmux
   set -g base-index 1
   setw -g pane-base-index 1
   ```

2. Start a session in AO and let the agent finish or exit.
3. Resume / restart the agent in that session.

Or reproduce it straight from the repo's own suite, no AO install needed:

```console
$ cd backend && go test -run TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell ./internal/adapters/runtime/tmux/...
--- FAIL: TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell (2.71s)
    tmux_integration_test.go:206: Restart: tmux runtime: restart session TestRuntimeIntegrationSupervised-d6de6f10: exit status 1: can't find pane: 0
FAIL
```

That failure is deterministic (3/3 runs) on a machine with the config above, and disappears the moment the config does — the same test on the same commit against a config-free tmux server passes. That is also why CI has never seen it: a default tmux sets both indices to `0`, so `:0.0` happens to resolve.

## Expected behavior

Restart respawns the agent in the session's pane regardless of the user's `base-index` / `pane-base-index`.

## Actual behavior

`tmux runtime: restart session <id>: exit status 1: can't find pane: 0`, every time, for the life of the install. The session is left alive but unusable for resume.

Direct probe against a tmux server with `base-index 1`:

```console
$ tmux respawn-pane -k -t "probe1:0.0" sh -c "sleep 20"
can't find window: 0
$ tmux respawn-pane -k -t "probe1"     sh -c "sleep 20"
(ok)
```

## AO version

upstream `main` @ `52fde0279` (`fix: hide inactive cloud sign-in controls (#4197)`)

## Environment

Arch Linux x86_64 (Omarchy), kernel 7.1.8, tmux 3.7b, Go 1.25.12. Not Linux-specific — the same `tmux.conf` reproduces it on macOS.

## Logs / extra context

Root cause is one literal, at `backend/internal/adapters/runtime/tmux/commands.go:24`:

```go
func respawnPaneArgs(id, cwd, shellPath, launchCmd string) []string {
	return []string{
		"respawn-pane", "-k",
		"-t", id + ":0.0",
		...
```

Every other pane-targeting builder in that same file already passes the session name alone (`send-keys`, `capture-pane`, `set-option`, `paneCurrentPathArgs`), and tmux resolves a bare session target to its active pane regardless of the configured indices. The file's own doc comment on `exactSessionTarget` spells out that convention; `respawnPaneArgs` is the one place that departs from it.

`panePIDArgs` (`commands.go:61`) carries the same `:0.0` literal but is not currently user-visible broken: `display-message` falls back to the session's active pane when the window part does not resolve. Verified — with `base-index 1`, `tmux display-message -p -t "probe1:0.0" "#{pane_pid}"` returns probe1's real pane pid, and an unknown *session* still errors, so the fallback is session-correct rather than arbitrary. It is the same latent assumption and worth moving with it.

I'd like to work on this — fix and regression test are ready. The test drives its own private tmux server (`TMUX_TMPDIR` + `-f /dev/null`) with both indices set to 1, so it reproduces on a default machine and on CI instead of only on a customised one.

yashksaini-coder:
## What

Address the agent's tmux pane by **position** (`<id>:^.{top-left}`) instead of by the hardcoded index `<id>:0.0`.

## Why

Fixes #4222.

tmux honours `base-index` when `new-session` picks a window index. On any server whose config sets `base-index 1` — one of the most common lines in a hand-written `tmux.conf`, since it puts window 1 under the `1` key — the session's only window is index **1**, so `respawn-pane -t <id>:0.0` fails with `can't find pane: 0` and **resuming an existing agent hard-fails for the entire life of the install**. Not Linux-specific; the same tmux.conf reproduces on macOS.

Visible in this repo's own suite. On a machine with that config, against `main`:

```console
$ cd backend && go test -run TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell ./internal/adapters/runtime/tmux/...
--- FAIL: TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell (2.71s)
    tmux_integration_test.go:206: Restart: tmux runtime: restart session …: exit status 1: can't find pane: 0
```

## How

The target has to satisfy **two** constraints, and the two obvious spellings each satisfy only one.

`:0.0` is deterministic but index-dependent — the bug above. A bare session name is index-independent but resolves to whatever pane is *currently active*, and `Attach` hands the user an ordinary attach client with tmux's default prefix key, so they can open a second window or split one at any time. `respawn-pane -k` would then kill the pane the user is working in and respawn the agent there, leaving the real agent pane behind as a corpse that `has-session` still reports healthy. `panePIDArgs` is worse: `supervisedProcessTree` roots its descendant scan at that pid, so a user-opened pane yields an empty descendant set and `IsSupervisedProcessAlive` reports a perfectly healthy agent as dead.

`:^.{top-left}` names the lowest-numbered window's top-left pane by position, which is both. `{start}`/`^` and `{top-left}` are documented tmux target tokens and predate AO's minimum tmux 3.2 (`tmux.go:750`).

Measured against a private tmux server — each alternative fails exactly one of the two integration tests:

| target | `base-index 1` | user opened a 2nd window |
| --- | --- | --- |
| `<id>:0.0` (main today) | **FAIL** | pass |
| `<id>` | pass | **FAIL** |
| `<id>:^.{top-left}` (this PR) | pass | pass |

Known ceiling, stated in the code comment: a user who splits the agent pane with `split-window -b` puts a new pane above/left of it, moving the anchor. Addressing the pane by the id tmux assigns at creation would be exact, but that means persisting new per-session state; positional targeting covers the splits users actually make. Happy to go further if you'd rather.

## Testing

Two integration tests, both driving a **private tmux server** (`TMUX_TMPDIR` + `-f /dev/null` + a cleared `TMUX`) so they reproduce on a default machine and on CI rather than depending on the developer's tmux.conf:

- `TestRuntimeIntegrationRestartUnderCustomBaseIndex` — sets both indices to 1, and asserts the window index really is 1 before restarting so a server that ignored the options fails loudly instead of passing vacuously.
- `TestRuntimeIntegrationRestartTargetsAgentPaneNotActivePane` — runs on **default** indices, opens a second window, and asserts Restart respawned the agent's pane and left the user's pane alive.

Clearing `TMUX` is not cosmetic: tmux only honours `TMUX_TMPDIR` when it is not already inside a session, so `go test` run from inside tmux would otherwise target the developer's real server and the cleanup `tmux kill-server` would destroy their sessions — AO's own agent sessions included. Reproduced that, then fixed it; the server setup is a shared helper so the second test cannot reintroduce it.

```console
$ go test ./internal/adapters/runtime/tmux/...          # base-index 1 box
ok
$ go test ./internal/adapters/runtime/tmux/...          # config-free tmux server
ok
$ go test -race ./...                                   # whole backend
ok
$ go vet ./internal/adapters/runtime/tmux/...
$ golangci-lint v2.12.2 run ./internal/adapters/runtime/tmux/...
0 issues.
```

Direct tmux probe behind the change, on a server with `base-index 1`, a second window and a split:

```console
$ tmux list-panes -s -t aosess -F "win #{window_index} pane #{pane_index} pid #{pane_pid} active=#{pane_active}"
win 1 pane 1 pid 402941 active=1        <- the agent
win 2 pane 1 pid 402944 active=0
win 2 pane 2 pid 402947 active=1
$ tmux display-message -p -t "aosess"              "#{pane_pid}"   # bare
402947                                              <- the user's split
$ tmux display-message -p -t "aosess:^.{top-left}" "#{pane_pid}"
402941                                              <- the agent
$ tmux respawn-pane -k -t "aosess:0.0" …
can't find window: 0
```

**Correction to my original issue text:** I wrote there that `display-message -t nosuch:0.0` "errors". It does not — on tmux 3.7 it exits 0 with empty output (`display-message` is `CMD_FIND_CANFAIL`). Harmless in context, since `strconv.Atoi("")` then fails, but the claim was wrong and the reasoning shouldn't be relied on.

**Overlap:** #4225 touches the same package. The two conflict only where both append to the tail of `tmux_integration_test.go`; `commands.go` and `tmux_test.go` auto-merge. I've verified the two fixes applied together are green on both tmux configurations and across the full backend suite, and I'll rebase whichever lands second.

## Checklist

- [x] Branched from `main` (based on `52fde0279`)
- [x] Relevant CI checks pass for the area touched (`go test`, `go test -race`, `go vet`, `golangci-lint`)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
Re-scoping this from bugfix to hardening, and rebasing onto current `main`.

**#4393 ("isolate AO tmux state") removed the user-facing failure this PR was opened for.** `managedArgs` now unconditionally prepends `-S <private socket> -f /dev/null` to every tmux invocation (`runtime/tmux/tmux.go:828`), so the server AO drives never loads the user's `tmux.conf` and a `base-index 1` setting cannot reach it. Verified rather than assumed, on the machine that produced the original report (`base-index 1`, confirmed live): `TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell` now passes on unmodified `main` @ `04b676c9a`, where it failed deterministically before. Happy to close if you'd rather.

**What I think is still worth landing:**

1. `"<id>:0.0"` at `commands.go:24` and `:61` is correct only while something guarantees both indices are 0. That guarantee now exists — but it lives in a different file and was introduced for an unrelated reason (socket isolation), so the coupling is invisible from here. Naming the pane by position doesn't rely on it.

2. **The second test doesn't depend on any tmux option at all**, so it is not hypothetical even with the private socket. AO hands the user an ordinary attach client with tmux's default prefix key, so they can open a second window inside the session at any time. A bare session target follows the *active* pane, which would make `respawn-pane -k` kill the pane the user is working in and leave the real agent pane behind as a corpse `has-session` still reports healthy. `panePIDArgs` is worse: `supervisedProcessTree` roots its descendant scan at that pid, so a user-opened pane yields an empty descendant set and a running agent is reported dead.

Measured against a private server — each alternative fails exactly one test:

| target | `base-index 1` | user opened a 2nd window |
| --- | --- | --- |
| `<id>:0.0` (main today) | **FAIL** | pass |
| `<id>` | pass | **FAIL** |
| `<id>:^.{top-left}` (this PR) | pass | pass |

**Conflict resolution worth flagging:** #4393 added a `logicalTmuxArgs()` helper that strips the socket prefix before asserting argv, on the same line this PR changes. I kept your wrapper *and* this branch's positional target rather than either side alone.

The integration tests also got simpler: AO's socket is already private, so the `TMUX`/`TMUX_TMPDIR` isolation this branch used to carry is gone. They now set `base-index` on AO's own private server after it starts — `-f /dev/null` suppresses config-file loading, not later `set-option` — so they exercise the hazard without depending on any `tmux.conf`.

Related: I've closed #4225, which #4393 made fully moot.

yashksaini-coder:
Rebased onto `0b7758c7f`. Attaching a recording of the local verification, since the argument for this one is a comparison rather than a single result.

**End-to-end test on my side, just now:** `go test -run TestRuntimeIntegrationRestart ./internal/adapters/runtime/tmux/...` → both integration tests pass. Then I swapped the pane target to each alternative in turn and re-ran, which is the part worth watching: with `<id>:0.0` (what `main` ships today) the base-index test fails; with a bare `<id>` the active-pane test fails instead; only `<id>:^.{top-left}` satisfies both. Then restored and re-ran green, ending on a clean worktree.

That is the whole case for the change in one screen — each candidate target is wrong in a different way, and they are not wrong in the same scenario, so neither alternative can be dismissed by the other's test.

Also re-ran the full package plus `internal/terminal` after the rebase: green. The 8 CI checks here are green too.

https://github.com/user-attachments/assets/9775892b-c50b-46d3-88f2-e6b01d708fed

yashksaini-coder:
Correcting my previous comment: **this is a live user-facing bug again, not hardening.**

That comment argued the failure was no longer reachable because #4393 moved AO onto a private tmux socket started with `-f /dev/null`, so a user's `tmux.conf` could not put the server into a non-default index layout. #4393 was reverted in #4465 ("the private hashed tmux socket stranded live legacy sessions across desktop updates"), which removed `managedArgs`, `logicalTmuxArgs`, `privateSocketAddress` and the config isolation along with it. The reasoning no longer holds, and the literal `":0.0"` target is back on `main` at `commands.go:24` and `commands.go:61`.

**Why it bites now.** Without `-f /dev/null`, AO's tmux server loads the user's `~/.config/tmux/tmux.conf`. `set -g base-index 1` is a common setting, and under it tmux numbers the agent's window 1.

Spawning a real session with a stock build of current `main`, on a machine with `base-index 1`:

```
$ tmux -L default list-windows -t work-1 -F '#{window_index} #{pane_index}'
1 1

$ tmux -L default respawn-pane -k -t "work-1:0.0"          # what main ships
can't find window: 0

$ tmux -L default respawn-pane -k -t "work-1:^.{top-left}"  # this PR
(ok)
```

`Restart` has no fallback, so resuming an exited agent fails permanently for those users.

The positional target is strictly better, not a trade — checked under both settings:

| target | `base-index 0` | `base-index 1` |
| --- | --- | --- |
| `:0.0` | ok | `can't find window: 0` |
| `:^.{top-left}` | ok | ok |

**What changed in this push.** Merged current `main` and resolved the fallout from the revert:

- `tmux_test.go` — kept the positional assertion, dropped the `logicalTmuxArgs` wrapper that no longer exists.
- `tmux_integration_test.go` — the Runtime option is `SocketName` (`tmux -L`) again rather than `SocketPath`, so the two pane-target tests now stand up a per-test named server. Worth flagging: with no explicit name the Runtime adopts tmux's machine-wide default socket, which is the developer's own server, so these tests must always set one.
- Corrected the base-index test's doc comment, which still claimed this was hardening.

Both call sites now go through one `agentPaneTarget()` helper instead of two hardcoded strings.

**Verification.** Reverting only the helper's return value makes both integration tests fail with the production error:

```
--- FAIL: TestRuntimeIntegrationRestartUnderNonDefaultBaseIndex
    Restart under base-index 1: tmux runtime: restart session ao-baseindex-agent: exit status 1: can't find pane: 0
--- FAIL: TestRuntimeIntegrationRestartTargetsAgentPaneNotActivePane
    Restart: tmux runtime: restart session ao-panetarget: exit status 1: can't find pane: 0
```

With the fix restored, both pass, and `go test ./...` is green across 161 packages. The one failure I see locally, `TestCopilotClassicPATIsNotAnAuthorizationSignal`, reproduces identically on a clean `main` checkout whenever a classic GitHub PAT is present in the environment, so it is unrelated to this branch.

yashksaini-coder:
Closing this — #4939 fixed the bug this PR was opened for, and my proposed alternative turns out not to hold up anyway.

**The reported failure is gone.** `respawnPaneArgs` and `panePIDArgs` now pass the bare session id, which resolves independently of `base-index` / `pane-base-index`. Verified rather than assumed, on a server with both set to 1:

```console
$ tmux respawn-pane -k -t 'probe:0.0' 'sleep 60'     # what main shipped before
can't find window: 0
$ tmux respawn-pane -k -t 'probe' 'sleep 60'          # what main ships now
$                                                     # ok
```

Nothing #4222 describes is reachable any more, and the bare target is a smaller change than the positional one I proposed here. Thanks for the fix.

**Correcting my own argument, since it is on the record in this PR.**

This PR argued that a bare target is unsafe because it follows the session's *active* pane, and proposed `<id>:^.{top-left}` as a positional alternative. The first half is right. The proposed remedy is not — I re-tested it properly before closing and it fails two ways on tmux 3.7c:

```console
# zoom: user splits, then hits prefix+z on their pane
agent pane pid 60736 ; user pane pid 60741
$ tmux display-message -p -t 'z:^.{top-left}' '#{pane_pid}'
60741        # <- collapses to the ACTIVE pane; tmux skips panes failing
             #    window_pane_visible, and only the active pane is visible when zoomed

# window reordering: user runs new-window -b
agent lands in window 2, user in window 1
$ tmux display-message -p -t 'w:^.{top-left}' '#{pane_pid}'
61190        # <- the user's pane; `^` is the lowest-numbered window, not the agent's
```

So `^.{top-left}` is defeated by one keystroke after the split its own test builds, and by `new-window -b` / `move-window` / `renumber-windows`. It would have shipped the same class of bug it claimed to fix. Better that this closed than merged.

**The underlying concern is still live on `main`, though, and it is wider than this PR treated it.** Beyond `respawnPaneArgs` and `panePIDArgs`, these also target the bare session id: `sendKeysLiteralArgs`, `sendEnterArgs`, `sendInterruptArgs`, `capturePaneArgs`, `capturePaneStyledArgs`. With a user-opened second pane active, `SendMessage` types the agent's prompt plus `Enter` into the user's interactive shell — arbitrary text run as a shell command — and `GetOutput` returns the user's pane content to AO's state detection. That is a worse outcome than the respawn case.

The target that survives all of it is the pane id captured at `Create` (`#{pane_id}`, e.g. `%0`), which is stable across zoom, splits, window renumbering and index settings — unlike any positional token.

I would rather file that as its own issue with these reproductions than reopen this PR on a rewritten argument. Happy to do so, and equally happy to be told the user-opened-pane scenario is rarer than I think and not worth the change.

PR #4225, opened 2026-08-21
fix(tmux): stop a destroyed session handing AO's terminal to another session

### LINKED ISSUE #4223: [Bug]: destroying a session hands AO's terminal to one of the user's own tmux sessions (`detach-on-destroy off`)
## Summary

If a user's `tmux.conf` sets `detach-on-destroy off`, closing an AO session hands AO's terminal client to **one of the user's own tmux sessions** instead of ending it. The AO terminal view then shows that unrelated session's output, and anything typed into it executes in that session's pane.

`detach-on-destroy off` is a common tmux.conf line — it is what keeps you inside tmux when you kill a session rather than dumping you back to the shell.

## Steps to reproduce

1. Put `set -g detach-on-destroy off` in `~/.config/tmux/tmux.conf` and restart the tmux server.
2. Have at least one tmux session of your own running.
3. Open a session in AO, then close/delete it.
4. The AO terminal pane does not go away — it is now attached to your other session. Type into it.

Reproducible mechanically, no AO install needed (fork a PTY onto `tmux attach -t <ao-session>`, then kill that session):

```
WITHOUT the fix (user has 'off') | client before kill: 'aosess' | after kill: 'other' | attach client exited: False
WITH  session-scoped 'on'        | client before kill: 'aosess' | after kill: ''      | attach client exited: True
```

The repo's own suite also catches it on such a machine:

```console
$ cd backend && go test -run TestAttachmentStreamsRealTmuxPane ./internal/terminal/...
--- FAIL: TestAttachmentStreamsRealTmuxPane (5.17s)
    attachment_integration_test.go:55: condition not met within 5s
FAIL
```

Line 55 is `eventually(t, 5*time.Second, func() bool { return a.isExited() })` — the attachment never observes the destroyed session, because tmux moved its client somewhere else instead of detaching it.

## Expected behavior

Destroying an AO session ends AO's attachment for that session. AO's terminal must never be reparented onto a session AO does not own.

## Actual behavior

Three failures, all silent:

- the attach stream never EOFs, so the attachment never observes the session ending and holds the PTY open;
- the unrelated session's output streams into AO's terminal view;
- **keystrokes typed in AO's terminal execute in that other session's pane.**

The third is the reason I am filing this as more than a cosmetic bug: it is input crossing a session boundary the user never agreed to, into a pane that may be running anything.

## AO version

upstream `main` @ `52fde0279` (`fix: hide inactive cloud sign-in controls (#4197)`)

## Environment

Arch Linux x86_64 (Omarchy), kernel 7.1.8, tmux 3.7b, Go 1.25.12. Not Linux-specific — the same `tmux.conf` reproduces it on macOS.

## Logs / extra context

`backend/internal/adapters/runtime/tmux/tmux.go` `Create` sets three session options after `new-session` (`status off`, `mouse on`, `window-size largest`) but not `detach-on-destroy`, so AO's sessions inherit whatever the user configured globally. tmux's default is `on`, which is why this is invisible on a default machine and on CI.

Setting it per-session is enough and does not touch the user's own sessions — verified:

```console
$ tmux set-option -g detach-on-destroy off
$ tmux set-option -t aosess detach-on-destroy on
$ tmux show-options -t aosess -v detach-on-destroy
on
$ tmux show-options -t other  -v detach-on-destroy
          # still inheriting the user's global 'off'
```

One wrinkle worth deciding on in review: tmux sessions outlive the app, so a session created before such a fix ships still carries the user's global `off` and keeps leaking until it is recreated. Setting the option in `Destroy` as well (best-effort, immediately before `kill-session`) retrofits those.

I'd like to work on this — fix and regression test are ready. The test drives a private tmux server (`TMUX_TMPDIR` + `-f /dev/null`) with the global set to `off`, attaches through the runtime's own `Attach`, and asserts the stream ends with no client left behind, so it reproduces on a default machine and on CI.

yashksaini-coder:
## What

Set `detach-on-destroy on` for AO's own tmux sessions, so destroying one detaches AO's terminal client instead of letting tmux move it onto a session AO does not own.

## Why

Fixes #4223.

tmux's `detach-on-destroy off` — a common tmux.conf line, since it keeps you inside tmux when you kill a session — makes tmux move an attached client to **another session** rather than detaching it. AO's terminal attachment is such a client, so destroying an AO session silently reparents AO's PTY onto one of the user's own sessions. Three consequences, all silent:

* the attach stream never EOFs, so the attachment never observes the session ending and holds the PTY open;
* that unrelated session's output streams into AO's terminal view;
* **anything typed into that view executes in that session's pane.**

The third is why this is more than cosmetic — it is input crossing a session boundary the user never agreed to, into a pane that may be running anything.

Measured directly, by forking a PTY onto `tmux attach -t <ao session>` and then killing that session:

```
WITHOUT the fix (user has 'off') | client before kill: 'aosess' | after kill: 'other' | attach client exited: False
WITH  session-scoped 'on'        | client before kill: 'aosess' | after kill: ''      | attach client exited: True
```

It also makes this repo's own `TestAttachmentStreamsRealTmuxPane` (`internal/terminal`) fail on such a machine, at the `a.isExited()` assertion — that failure is what led me here.

## How

Two call sites, covering two different populations of session.

**`Create`** already sets three session options after `new-session` (`status off`, `mouse on`, `window-size largest`); this adds a fourth in the same shape. Scoped with `-t`, so the user's own sessions keep whatever they configured — verified: with the global at `off`, `show-options -t <ao session>` reports `on` while `show-options -t other` still inherits `off`.

**`Attach`** retrofits it, best-effort. tmux sessions outlive the app, so a session created by a build without the `Create` hunk still carries the user's global `off` and would keep leaking until it was recreated. My first version put this in `Destroy`; review pointed out that only covers teardowns AO initiates, and misses the user typing `exit` in the keep-alive shell or killing the session from their own tmux. Attaching is the **exact precondition** for the leak — with no client of ours attached there is nothing for tmux to reparent — so setting it there covers an adopted session for every way it can end. That makes the `Destroy` call redundant rather than merely insufficient, so it is gone and four `Destroy` call-order assertions go back to their original form.

## Testing

Two integration tests, both driving a **private tmux server** (`TMUX_TMPDIR` + `-f /dev/null` + a cleared `TMUX`) with the global set to `off`, plus a bystander session standing in for one of the user's own. tmux's default is `on`, so a test relying on the developer's tmux.conf would be green on CI and prove nothing.

- `TestRuntimeIntegrationDestroyDetachesClientUnderDetachOnDestroyOff` — attaches through the runtime's own `Attach`, destroys, asserts the stream ends with no client left behind. It waits for the client to appear in `list-clients` first, so it cannot pass by racing the attach.
- `TestRuntimeIntegrationAttachRetrofitsSessionCreatedBeforeFix` — ages a session by forcing the option back to `off` after `Create`, which is exactly the state a pre-fix build leaves behind, then attaches and destroys. Fails without the `Attach` hunk, naming the leak:

```
attach stream never ended after Destroy; tmux clients now on: "ao-dod-bystander"
```

Clearing `TMUX` is not cosmetic: tmux only honours `TMUX_TMPDIR` when it is not already inside a session, so `go test` run from inside tmux — likely, for tests about tmux — would otherwise target the developer's real server, flip its global `detach-on-destroy`, and let the cleanup `tmux kill-server` destroy their sessions, AO's own agent sessions included. I reproduced that on a throwaway server (a session named `precious_user_session` did not survive) before fixing it. Server setup and the attach/destroy assertion are shared helpers so a future test cannot reintroduce either hole.

```console
$ go test ./internal/terminal/...                       # incl. the previously-failing attachment test
ok
$ go test ./internal/observe/... ./internal/session_manager/...
ok
$ go test -race ./...                                   # whole backend
ok
$ go vet ./internal/adapters/runtime/tmux/...
$ golangci-lint v2.12.2 run ./internal/adapters/runtime/tmux/...
0 issues.
```

Note: on a box that also sets `base-index 1`, `TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell` still fails here — that is the separate bug in #4222, fixed by #4224, and unrelated to this change.

**Overlap:** #4224 touches the same package. The two conflict only where both append to the tail of `tmux_integration_test.go`; `commands.go` and `tmux_test.go` auto-merge. I've verified the two fixes applied together are green on both tmux configurations and across the full backend suite under `-race`, and I'll rebase whichever lands second.

## Checklist

- [x] Branched from `main` (based on `52fde0279`)
- [x] Relevant CI checks pass for the area touched (`go test`, `go test -race`, `go vet`, `golangci-lint`)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
Closing this — #4393 ("isolate AO tmux state") fixed the root cause more thoroughly than this PR did, and it is no longer reachable.

`managedArgs` now unconditionally prepends `-S <private socket> -f /dev/null` to every tmux invocation (`backend/internal/adapters/runtime/tmux/tmux.go:828`), and the legacy-default-socket routing that #4293 had is gone. AO's tmux server therefore loads none of the user's config, so a global `detach-on-destroy off` can never apply to an AO session, and AO no longer attaches a client to the user's own server at all. The reparenting this PR guarded against cannot happen.

Verified rather than assumed, on a box whose tmux.conf sets `detach-on-destroy off` (confirmed live: `tmux show-options -gv detach-on-destroy` -> `off`):

```console
# unmodified upstream/main @ 04b676c9a
$ go test -run TestAttachmentStreamsRealTmuxPane ./internal/terminal/...
ok
```

That test failed deterministically on this same machine before #4393, at the `a.isExited()` assertion, which is what led me to file #4223 in the first place.

One thing worth leaving on the record: `detach-on-destroy` is still never set anywhere in the adapter, so the code is dead-but-wrong rather than correct — it would come back if the `-f /dev/null` isolation is ever relaxed. Not worth a PR on its own; noting it in case you want it captured.

Thanks for the better fix.

PR #4228, opened 2026-08-21
feat(telemetry): capture project GitHub org for usage ranking

Pulkit7070:
## What

Captures the GitHub organisation/account that owns a project so AO usage can be attributed and **ranked per org/account** (a prerequisite for connecting with the most-engaged teams and individuals for feedback).

- Adds a `github_org` property to `ao.projects.created` and `ao.onboarding.first_project_added`, derived from the project's `origin` remote.
- Parses scp (`git@github.com:owner/repo.git`), `https`, `ssh://`, and `git://` remote forms; returns **only the owner segment**.
- Adds `github_org` to the daemon remote allowlist for both events.
- Discloses the new field in `docs/telemetry.md`.

## What is intentionally NOT sent

Repository name, path, branch, or full URL. Only the owner slug.

## Privacy note (important)

`github_org` is **not anonymous** for personal repositories: there the owner slug is the user's own GitHub username. This makes AO telemetry non-anonymous for solo devs, which is why it is disclosed in `docs/telemetry.md`.

Follow-ups (separate PRs), not in this one:
- A first-run consent notice + a Settings toggle (recommended before this ships to users, since telemetry is currently on-by-default and described as "anonymous").
- Per-user `github_login` capture (needs the SCM identity resolver wired at the daemon root) for person-level outreach.
- Optionally hashing `github_org`.

## Tests

- `TestGithubOwner` covers scp/https/ssh/git forms, personal accounts, whitespace, non-GitHub, gist subdomain, and owner-only remotes.

## Risk

Low. Only adds one bounded string property on project-add events, gated by the existing telemetry consent/allowlist path. No behavior change when a project has no GitHub origin.

PR #4229, opened 2026-08-21
feat(telemetry): per-session token-usage ranking telemetry (on main's usage subsystem)

Pulkit7070:
## What

Emits **per-session token usage as ranking telemetry**, so AO usage can be ranked per GitHub org/account by spend.

This builds on AO's existing usage subsystem (`internal/observe/usage`, `internal/service/usage`) — which already parses agent transcripts (Claude Code + Codex) and persists per-session token totals — and adds only the parts it lacks:

- `internal/service/usagetelemetry` **decorates the lifecycle usage finalizer**. Right after the pipeline finalizes a session's usage at termination, it reads the session's usage summary, estimates a **USD cost** from a static per-model price table (Opus / Sonnet / Haiku / GPT tiers), resolves the project's **GitHub owner**, and emits one `ao.session.token_usage` event. Emission is best-effort and never affects finalization.
- Because ingestion is asynchronous, the event carries an **`incomplete`** flag when totals may still be catching up at exit, so you can filter or accept them in PostHog.
- `ao.session.token_usage` gets a dedicated **metering rate-limit tier (30/min, 1000/day)** so heavy users and orchestrator fleets aren't silently dropped by the standard 5/min ceiling, and is added to the daemon remote allowlist.
- Disclosed in `docs/telemetry.md`.

## Ranking basis

Group `ao.session.token_usage` by `github_org`; break down by `model`. total_tokens is a cumulative per-session snapshot, so take the latest event per session (argMax(total_tokens) by time) before summing across sessions — never sum a single session's snapshots. Cost is a query-time view (tokens x per-model rate), not emitted.

## Note on history

This **supersedes the earlier standalone parser/hook/endpoint approach**. After that was opened, a full usage subsystem landed on `main` (fsnotify transcript watcher + parser for both harnesses + per-session SQLite persistence). Re-implementing parsing would have duplicated it, so this PR was rebuilt to consume main's data and add only cost + github_org + the ranking event.

## Verification

- `internal/service/usagetelemetry` unit tests drive the decorator end-to-end with the real domain types: forwards to the inner finalizer, emits the event with correct tokens/model/cost/github_org, carries the `incomplete` flag, skips empty usage, returns the inner error while still emitting, and omits `github_org` for non-GitHub remotes.
- Pricing tests cover Opus/Sonnet/Haiku tier precedence, uncached-input preference, reasoning-as-output, cache-read, and unknown-model (zero).
- Rate-limit metering-tier tests (burst + daily ceilings). golangci-lint v2: 0 issues.
- Not run e2e: a full live-daemon path (real session termination → finalize → emit → PostHog). The decorator + wiring are verified by tests and compile-time interface satisfaction against main's concrete `usagesvc.Collector` / `SummaryReader`.

## Stacking

Based on `ao/ao-25/github-usage` (#4228) for the `github_org` attribution; merge that first or GitHub retargets to `main`.


backend/internal/service/usagetelemetry/emitter.go
```diff
@@ -0,0 +1,185 @@
...
+		clock = time.Now
+	}
+	return &EmittingFinalizer{inner: inner, summary: summary, store: store, scm: scm, telemetry: telemetry, now: clock}
+}
+
+// FinalizeSession forwards to the wrapped finalizer, then emits usage telemetry
+// once the totals are as complete as they will be at termination.
+func (e *EmittingFinalizer) FinalizeSession(ctx context.Context, id domain.SessionID, expectedRuntimeLaunchID string, expectedSessionRevision time.Time) error {
+	err := e.inner.FinalizeSession(ctx, id, expectedRuntimeLaunchID, expectedSessionRevision)
+	e.emit(ctx, id)
```
illegalcall: **Blocking: this emits before either usage ingestion or termination is actually finished.** The inner finalizer only marks bindings as finalizing and sends a non-blocking wake-up to the ingestion pipeline; e.emit then reads the summary immediately. Example: a short session exits before the watcher consumes its transcript, so this read sees zero tokens and line 88 drops the event forever. There is a second failure mode: lifecycle intentionally calls the finalizer before it commits termination and may retry when UpdatedAt changes. Every call reaches this line, so one real termination can emit twice, or a rejected termination can emit a false session-end event. Could emission move to the point where the usage cursor is settled and termination is committed, with a durable idempotency key such as session ID plus runtime launch ID? Another option is for finalization to return a receipt/result and let lifecycle emit only after its mutation succeeds. Please add regression tests for pending ingestion, a revision retry, and rejected termination.

backend/internal/service/usagetelemetry/pricing.go
```diff
@@ -0,0 +1,92 @@
...
+}{
+	{"claude-3-opus", price{15, 75, 1.50}},
+	{"opus", price{15, 75, 1.50}},
+	{"claude-3-5-haiku", price{0.80, 4, 0.08}},
+	{"claude-3-haiku", price{0.25, 1.25, 0.03}},
+	{"haiku", price{0.80, 4, 0.08}},
+	{"sonnet", price{3, 15, 0.30}},
+	// Codex / GPT tiers priced at published GPT-5-class rates as a coarse proxy.
+	{"gpt-5", price{1.25, 10, 0.125}},
+	{"gpt", price{1.25, 10, 0.125}},
```
illegalcall: **The spend ranking uses materially wrong model prices and loses precision before aggregation.** This broad Opus match prices the PR fixture claude-opus-4-8 at $15/M input and $75/M output, while Anthropic currently lists $5/$25: https://www.anthropic.com/news/claude-opus-4-8. The generic GPT match similarly prices GPT-5.4 at $1.25/$10 instead of the documented $2.50/$15: https://developers.openai.com/api/docs/models/gpt-5.4. There is also a precision problem: round2 runs for every session, so an estimated $0.00225 becomes $0.00; across 1,000 such sessions PostHog still sums $0 instead of $2.25. That can change organization rankings when their model mixes differ. Potential fix: match concrete model families/versions against documented, dated rates; keep 4-6 decimal places or integer microdollars in the event; and round only when displaying an aggregate. Tests should cover current concrete model IDs and aggregation of many small sessions.

backend/internal/service/usagetelemetry/emitter.go
Pulkit7070: Fixed in c71e30096. Emission no longer runs in the lifecycle finalizer; the collector now runs on a store wrapper that emits only on the `CompleteUsageBindingIfSettled` transition, once a session's bindings are all terminal (complete/partial). Because that transition fires exactly once per binding after its transcript is ingested, a short session no longer reads zero and drops the event. The Emitter dedupes by total, so multi-binding settles and reconcile retries never double-count, and a rejected termination never settles so it never emits a false event. bindingID->sessionID is learned from records the collector already reads (no new query); a binding first seen only after a restart is skipped rather than mis-attributed. Tests cover pending ingestion (no-op), an idempotent repeat, and re-emit on a genuinely higher total.

backend/internal/service/usagetelemetry/pricing.go
Pulkit7070: Fixed in f5fe4d7ea. Prices now match concrete, dated model families: Opus 4.x is $5/$25 (it was being charged the legacy Claude 3 Opus $15/$75) and GPT-5.4 is $2.50/$15, cached $0.25 (was $1.25/$10); an unmatched model is left unpriced rather than guessed, with rates cited and dated in the table. Both rates were verified against the vendors' published pages. On precision: per-session rounding is gone. The event now carries integer est_cost_microusd for exact aggregation alongside the unrounded est_cost_usd, and rounding happens only at display. Tests cover concrete/legacy/unmatched model IDs and micro-dollar aggregation of many sub-cent sessions.

backend/internal/daemon/daemon.go
```diff
@@ -309,7 +310,23 @@ func Run() error {
...
+			usageSCM = p
+		}
+		usageEmitter := usagetelemetry.NewEmitter(
+			usagesvc.NewSummaryReader(store),
+			store,
+			usageSCM,
+			telemetrySink,
+			nil,
+		)
+		collectorStore := newUsageSettleStore(store, usageEmitter)
```
illegalcall: **Blocking: the normal ingestion path still bypasses this telemetry hook.** This wrapper is passed only to the collector; the ingestor below is still constructed with the raw SQLite `store`. The collector marks a binding as finalizing and wakes the pipeline, but the ingestor is what later calls `CompleteUsageBindingIfSettled`. Because that call goes to the raw store, this wrapper never sees the transition and never emits. In a real-SQLite reproduction, the binding became complete and the summary contained 120 tokens, but the sink received 0 events. Changing only the ingestor argument to `collectorStore` produced exactly 1 event. Please pass the wrapper to the ingestor too (or move emission into a durable settlement boundary) and add a daemon-level ingest → complete → one-event regression test.

Pulkit7070: Fixed in https://github.com/Untrivial-ai/agent-orchestrator/pull/4229/commits/9b74c5d755322f7267bf336f01a133c25fc6073a. You're right: the ingestor drives `CompleteUsageBindingIfSettled` on the normal path and it was constructed on the raw store, so the wrapper never saw the settle and nothing emitted. The ingestor now runs on the same wrapped settle store as the collector. The collector still populates the shared bindingID to sessionID map that the ingestor's settle reads, and the atomic transition returns true exactly once, so a single event emits. I pulled the construction into `buildUsageCollection` so `Run` and the test wire it the same way, and added a daemon-level regression test (`TestIngestorOnSettleStoreEmitsExactlyOneUsageEvent`) that ingests a transcript to EOF over a real SQLite store, finalizes the binding, and asserts exactly one `ao.session.token_usage` event with the 120-token total. It fails with 0 events if the ingestor is built on the raw store, matching your reproduction.

backend/internal/service/usagetelemetry/emitter.go
```diff
@@ -0,0 +1,209 @@
...
+	// (providers record tokens, not a price), so it is a derived view computed
+	// at query time in PostHog (tokens x per-model rate) rather than frozen into
+	// each event with a rate that can be wrong or go stale.
+	payload := map[string]any{
+		"harness":             harness,
+		"model":               model,
+		"input_tokens":        input,
+		"cached_input_tokens": cached,
+		"output_tokens":       output,
+		"total_tokens":        total,
```
illegalcall, line 143: **Blocking: a restored session is counted twice by the documented ranking query.** This happens when a session has already settled and emitted usage, the user chooses **Restore session**, then the same session uses more tokens and settles again. Restore reuses the AO session ID and preserves its stored usage, so this summary is cumulative. The equality check above allows the higher value through and this line emits that full cumulative value again. For example, the first lifetime emits 120 tokens; after restore, 60 more tokens make the stored total 180, so the second event emits 180. The PR tells PostHog to sum total_tokens, which reports 300 even though actual usage is 180. I reproduced that exact 120 to 180 to 300 result through the real SQLite collector, ingestor, settle wrapper, and emitter; no race or daemon restart is required. This occurs for every previously-emitted session that is restored, consumes more tokens, and settles again. Potential fixes: (1) emit only the increase since the previous event, with the previous value stored durably so restarts do not reset it; (2) include a runtime-launch/lifetime ID and emit totals scoped to that lifetime; or (3) deliberately keep cumulative snapshots but change the documented PostHog query to select the latest event per session before summing across sessions. Please add a restore/reactivation regression test for whichever event contract is chosen.

illegalcall (changes requested):
The production settle wiring is fixed and the new daemon-level regression test covers it. One remaining correctness issue affects restored sessions and the ranking query documented by this PR.

backend/internal/service/usagetelemetry/emitter.go
Pulkit7070: Fixed in https://github.com/Untrivial-ai/agent-orchestrator/pull/4229/commits/3f98494bf1df4a9a9f51f30b063ef885ddcbf630. You're right — total_tokens is a cumulative snapshot, so a restored session that runs more emits a second larger snapshot (120 then 180) and a naive sum double-counts (300). I went with option 3: the emitter already emits the correct cumulative value, so I pinned the contract rather than changing the number. total_tokens is now documented as a lifetime snapshot, and the ranking takes the latest event per session (argMax(total_tokens) by time) before summing across sessions; I corrected the PR body's query line from "sum total_tokens" to that. Added `TestEmitSessionUsageRestoreEmitsCumulativeSnapshotNotDelta` reproducing your 120 to 180 case and asserting the second event is the cumulative 180, not a delta. I preferred this to deltas because cumulative-latest also survives a daemon restart correctly (the restart re-emits the current total, still the right latest reading), whereas a delta would over-count on restart without durable per-session state.

i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@Pulkit7070 Closing as stale soon: no activity from you for 30 days. Comment or push and this comes off the close list. The branch stays safe to reopen if you pick this up again.

Issue #3218, opened 2026-07-28
bug(sessions): restore silently swaps an unrecognized model instead of refusing or prompting


zenprocess:
## Summary
Restoring a session whose stored model string isn't recognized by the current build silently falls back to a different model, with only a note left in the transcript — no refusal, no prompt, no surfaced error.

## Repro
Observed on `agent-orchestrator-3` (2026-07-28): a MiniMax-M3 session was restored and silently fell back to `claude-fable-5`. The only trace was a transcript note; nothing in the session list, logs, or UI flagged that the requested model was unavailable and a substitution had occurred.

## Why this matters
This is a tier/cost leak: a session that was deliberately provisioned on one model (for cost or capability reasons) can silently resume on a materially different one. Nobody chose that swap, and there's no signal to catch it after the fact short of manually re-reading the transcript.

## Proposed behavior
One of:
- Refuse to restore and surface a clear error naming the missing/unrecognized model, letting the operator choose explicitly, or
- Honor the stored model string as-is (fail loud if the backend genuinely doesn't support it, rather than substituting).

Silent substitution should not be an option in either case.

## Scope
Session restore path (backend session lifecycle / agent adapter resolution). Needs investigation to pin the exact fallback site before a fix PR.

somewherelostt:
Thanks for contributing to Agent Orchestrator.

This issue is being picked up by the current external contributor on-call pair:

- @illegalcall
- @pulkit7070

If someone is already working on this, please continue as usual.
The on-call pair is added for visibility, tracking, and support, not to take over the work.
If you need help with review, direction, reproduction, or next steps, please tag @illegalcall and @pulkit7070 here.

For faster context or live questions, you can also join the AO Discord.

Join the session here:
https://discord.gg/H6ZDcUXmq

Come by if you want to see what is being built, ask questions, or just hang around with the community.

zenprocess:
## Root cause (pin only — fix not included)

[...]

PR #4231, opened 2026-08-21
fix(claudecode): preserve session model on restore

vanshx999:
Claude Code restore could replace a session's model with the current role or project default. Restore now uses the model recorded on the session, and the Claude adapter and command builder forward that selection to native resume.

TUI restore, native resume, restart recovery, saved-prompt fallback, and Chat continuation share this rule. Blank agent-default and legacy rows leave the model unset, allowing native settings to apply without inventing a selection from current project defaults. The existing permission policy remains pinned independently.

The new manager regression uses the production lifecycle reducer and closes and reopens SQLite before continuation. Its 72 cases cover workers and orchestrators, spawn overrides, role/project defaults, custom provider IDs, agent defaults, legacy rows, six continuation paths, and native conversation identity.

Validation on ac76bfc66ea90727961348dcadfb9e421e2c582e:
- All 10 GitHub checks pass: build, vet, race tests, lint, Windows workspace, API drift, sqlc drift, secret scanning, and native/container CLI E2E.
- Local gofmt and PR-diff whitespace checks pass.
- ENOSPC and the explicit compilation stop prevented rerunning the new suites locally. The earlier 122 adapter/runtime and 57 focused manager tests remain observed passes.

Provider model admission, effective-model reporting, and changes made with native /model remain follow-ups.

Refs #3218.


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/agent/claudecode/claudecode.go
```diff
@@ -242,6 +244,7 @@ func (p *Plugin) GetRestoreCommand(ctx context.Context, cfg ports.RestoreConfig)
 		Binary:           binary,
 		SessionID:        cfg.Session.ID,
 		Metadata:         cfg.Session.Metadata,
+		Model:            cfg.Config.Model,
```
Rishet11, line 246: [P1] Preserve the session model on restore

`cfg.Config.Model` is populated by `effectiveAgentConfig(rec.Kind, project.Config)` in `relaunchSessionWithPolicy`, not by the stored session model. `ao spawn --model` is explicitly session-scoped, and spawn already persists its resolved value as `rec.Metadata.Model`. If a session starts with model A while the project default is B, this line sends B on restore; changing project config after spawn has the same effect. Feed the persisted session model into the restore config, with a deliberate legacy fallback if needed, and add a manager-level regression test that spawns with an override and then restores.

Rishet11 (commented):
Requesting changes.

The builder now forwards a model, but the restore path still loses the model selected for an individual session. `ao spawn --model` persists that resolved choice in `SessionMetadata.Model`, while restore rebuilds `cfg.Config` from the current project and role settings. A spawn override, or a later project-config change, therefore still resumes Claude on a different model. This leaves the silent swap from #3218 unfixed for a supported path.

I did not run project code because the review task prohibits it. All eight GitHub checks for `ba094abf` completed successfully.

backend/internal/adapters/agent/claudecode/claudecode.go
Rishet11: Addressed in f49de5e3b4498c866b7f558ed0641508c294ec55, 06d6f3ddbb81285cdc5240879ff8b1938857fd42, and ac76bfc66ea90727961348dcadfb9e421e2c582e.

The manager now supplies the recorded Claude session model to TUI and Chat continuation paths. Changed role/project defaults no longer replace that selection. Blank legacy and agent-default rows remain unset so native settings can apply. The merge with current main retains its production model-persistence fix and pinned-permission behavior.

TestClaudeSessionModelSurvivesRestore uses the real lifecycle reducer and reopened SQLite. Its 72 cases cover workers and orchestrators, spawn/role/project selections, custom IDs, blank and legacy rows, six continuation paths, and native conversation identity.

All 10 GitHub checks pass on ac76bfc, including build, vet, race tests, lint, Windows workspace, API/SQL drift, secret scanning, and CLI E2E. Local compilation remained stopped because of ENOSPC; gofmt and PR-diff whitespace checks passed.

Rishet11 (approved):
Reviewed the final net diff at ac76bfc. No blocking findings. The P1 is addressed: the resolved spawn selection persists through production lifecycle storage, reopened sessions override mutable role and project defaults in both TUI and Chat, blank legacy and agent-default rows remain unset, and the adapter forwards the exact opaque model on native resume. The 72-case manager matrix covers both roles and six continuation paths. All 10 required checks pass. I did not rerun local compilation because the active ENOSPC constraint remains.

PR #4233, opened 2026-08-21
fix: default new sessions to chat UI

illegalcall:
## What

Make Chat UI the daemon-owned preferred interface for newly created sessions, including orchestrators.

## Behavior

- Spawns with no explicit mode try Chat first.
- If Chat is unsupported or preflight reports an unavailable, incompatible, or unauthenticated driver, the spawn falls back to TUI.
- Explicit Chat requests remain strict and return the preflight error instead of changing interfaces.
- Existing sessions are untouched.
- The migration intentionally updates the saved default to Chat for existing installations as well as fresh installations.

## How

- Add append-only migration `0104_default_session_mode_chat.sql` after the migrations currently on `main`.
- Make the session manager distinguish an inherited default from an explicitly requested mode.
- Preserve the saved preference on migration rollback because its previous value cannot be reconstructed safely.
- Cover successful inherited Chat, TUI fallback paths, upgrade behavior, and the pre-renumber migration preservation check.

## Testing

- `go test ./internal/session_manager ./internal/storage/sqlite ./internal/domain ./internal/httpd/controllers`
- `go test ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs` — 0 issues

## Checklist

- [x] Rebased onto current `main`
- [x] One focused change
- [x] Regression coverage for the changed default and fallback behavior

### DIFF
--- a/backend/internal/storage/sqlite/migrate_burned_versions_test.go
+++ b/backend/internal/storage/sqlite/migrate_burned_versions_test.go
@@ -108,6 +108,7 @@ var shippedMigrations = map[int64]string{
 	102: "0102_canonical_usage.sql",
 	103: "0103_review_run_cdc.sql",
 	104: "0104_agent_inventory_cache.sql",
+	105: "0105_default_session_mode_chat.sql",
 }
 
 // burnedVersion reports version numbers that must never be (re)used: they
--- a/backend/internal/storage/sqlite/migrate_test.go
+++ b/backend/internal/storage/sqlite/migrate_test.go
@@ -38,6 +38,65 @@ var expectedUsageTableColumns = map[string][]string{
 	},
 }
 
+func TestMigrateDefaultsSessionInterfaceToChat(t *testing.T) {
+	db := openMigratedTestDB(t)
+
+	var mode string
+	if err := db.QueryRow(`SELECT default_session_mode FROM app_settings WHERE id = 1`).Scan(&mode); err != nil {
+		t.Fatalf("read default session mode: %v", err)
+	}
+	if mode != "chat" {
+		t.Fatalf("default session mode = %q, want chat", mode)
+	}
+}
+
+func TestMigrateUpdatesExistingSessionInterfaceDefaultToChat(t *testing.T) {
+	db, err := sql.Open("sqlite", "file:"+filepath.Join(t.TempDir(), "ao.db")+pragmas)
+	if err != nil {
+		t.Fatalf("open sqlite: %v", err)
+	}
+	t.Cleanup(func() { _ = db.Close() })
+	upTo(t, db, 103)
+
+	if _, err := db.Exec(`UPDATE app_settings SET default_session_mode = 'tui' WHERE id = 1`); err != nil {
+		t.Fatalf("seed existing TUI default: %v", err)
+	}
+	if err := migrate(db); err != nil {
+		t.Fatalf("migrate existing database: %v", err)
+	}
+
+	var mode string
+	if err := db.QueryRow(`SELECT default_session_mode FROM app_settings WHERE id = 1`).Scan(&mode); err != nil {
+		t.Fatalf("read default session mode: %v", err)
+	}
+	if mode != "chat" {
+		t.Fatalf("default session mode = %q, want chat after upgrade", mode)
+	}
+}
+
+func TestMigrateRollbackPreservesSessionInterfacePreference(t *testing.T) {
+	db, err := sql.Open("sqlite", "file:"+filepath.Join(t.TempDir(), "ao.db")+pragmas)
+	if err != nil {
+		t.Fatalf("open sqlite: %v", err)
+	}
+	t.Cleanup(func() { _ = db.Close() })
+	upTo(t, db, 103)
+
+	if _, err := db.Exec(`UPDATE app_settings SET default_session_mode = 'chat' WHERE id = 1`); err != nil {
+		t.Fatalf("seed deliberate Chat default: %v", err)
+	}
+	upTo(t, db, 105)
+	downTo(t, db, 104)
+
+	var mode string
+	if err := db.QueryRow(`SELECT default_session_mode FROM app_settings WHERE id = 1`).Scan(&mode); err != nil {
+		t.Fatalf("read default session mode: %v", err)
+	}
+	if mode != "chat" {
+		t.Fatalf("default session mode after rollback = %q, want preserved chat", mode)
+	}
+}
+
 func TestUsageTablesKeepOnlyDurableCollectionState(t *testing.T) {
 	db := openMigratedTestDB(t)
 	for table, wantColumns := range expectedUsageTableColumns {
--- a/backend/internal/domain/sessionmode.go
+++ b/backend/internal/domain/sessionmode.go
@@ -22,8 +22,9 @@ const (
 	SessionModeChat SessionMode = "chat"
 )
 
-// DefaultSessionMode is what a session gets when no mode was requested. It stays
-// TUI so an upgrade never changes how existing workflows behave.
+// DefaultSessionMode is the compatibility fallback when the daemon-owned
+// preference is unavailable or invalid. Keep it TUI so a settings read failure
+// cannot turn a working terminal spawn into a Chat-only failure.
 const DefaultSessionMode = SessionModeTUI
 
 // Valid reports whether mode is one AO knows how to dispatch.
--- a/backend/internal/domain/sessionmode_test.go
+++ b/backend/internal/domain/sessionmode_test.go
@@ -69,9 +69,9 @@ func TestParseSessionModeRejectsUnknownInsteadOfFallingBack(t *testing.T) {
 	}
 }
 
-func TestDefaultSessionModeIsTUIForUpgradeCompatibility(t *testing.T) {
+func TestDefaultSessionModeIsTUIForSettingsFailureCompatibility(t *testing.T) {
 	if DefaultSessionMode != SessionModeTUI {
-		t.Fatalf("DefaultSessionMode = %q, want %q so an upgrade changes no existing workflow",
+		t.Fatalf("DefaultSessionMode = %q, want %q so a settings failure keeps terminal spawns working",
 			DefaultSessionMode, SessionModeTUI)
 	}
 }
--- a/backend/internal/httpd/controllers/dto.go
+++ b/backend/internal/httpd/controllers/dto.go
@@ -189,11 +189,11 @@ type SpawnSessionRequest struct {
 	Branch          string                 `json:"branch,omitempty"`
 	// Mode picks the conversation controller: chat talks to the agent over a
 	// structured connection, tui opens the agent's native terminal interface.
-	// Omitted resolves to the daemon default (tui), which is why an upgrade
-	// changes nothing. Compatible sessions may later switch through the durable
-	// interface-transition endpoint; the default never mutates existing sessions
-	// automatically. An unsupported explicit request fails rather than quietly
-	// producing the other kind of session.
+	// Omitted resolves to the daemon-owned preference, which defaults to Chat for
+	// new sessions and falls back to TUI when Chat is unavailable. The preference
+	// never mutates existing sessions automatically; compatible sessions may later
+	// switch through the durable interface-transition endpoint. An unsupported
+	// explicit request fails rather than quietly producing the other kind of session.
 	Mode   domain.SessionMode `json:"mode,omitempty" enum:"chat,tui"`
 	Prompt string             `json:"prompt,omitempty" maxLength:"4096"`
 	// Model is an optional agent model override scoped to this single spawn. Empty
--- a/backend/internal/session_manager/chat_spawn_test.go
+++ b/backend/internal/session_manager/chat_spawn_test.go
@@ -32,6 +32,12 @@ func newChatManager(chat ChatLauncher) (*Manager, *fakeStore, *fakeRuntime) {
 
 const chatTestProject = domain.ProjectID("mer")
 
+type fixedSessionModeDefaults domain.SessionMode
+
+func (d fixedSessionModeDefaults) DefaultSessionMode(context.Context) domain.SessionMode {
+	return domain.SessionMode(d)
+}
+
 // The load-bearing property of the split: exactly one controller starts. A chat
 // spawn must not touch the terminal runtime, and a TUI spawn must not touch the
 // chat launcher. Anything else means two writers on one conversation.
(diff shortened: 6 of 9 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+UPDATE app_settings
```
whoisasx: **[P1] This makes spawns fail outright for the 22 of 27 harnesses that have no chat driver.**

**Trigger:** any spawn that names no mode. `backend/internal/cli/spawn.go:195` documents `--mode` as "Omitted uses the daemon default", so `ao spawn --agent aider`, headless/mobile spawns, and orchestrator-issued worker spawns all land here.

**Impact:** `backend/internal/session_manager/manager.go:740` resolves the mode, then `:745` calls `PreflightChat`, which returns `ErrChatUnsupported` for any harness without a registered driver — and the spawn returns an error. `manager.go:738` says so explicitly:

> It never falls back to TUI — that would put the user in a terminal they deliberately did not ask for.

That reasoning holds for an *explicitly requested* chat mode. This migration turns it into the default for spawns that requested nothing, so the comment's premise ("deliberately") no longer applies and session creation breaks completely for those agents.

**Evidence:** `adapters/chatdriver/registry/registry.go:57-64` registers exactly 5 drivers (codex, claude-code, opencode, droid, kimchi); `domain.AllHarnesses` has 27. A temporary table test spawned every TUI-only harness twice — with `Defaults` returning `tui` all 22 succeeded; with `Defaults` returning `chat` all 22 failed:

```
REGRESSION aider: spawn: aider has no chat driver: chat mode unsupported for harness
... x22
```

**Why CI misses it:** no test wires a non-nil `SessionModeDefaults` into `Manager.Spawn` — every existing test either passes `RequestedMode` explicitly or leaves `m.defaults == nil`, which short-circuits to `domain.DefaultSessionMode`. The new test in this PR asserts the DB cell but never the spawn consequence.

**Correction:** gate the daemon default on capability. In `resolveSessionMode`, when the mode came from the default rather than an explicit request and `m.chat == nil || !m.chat.SupportsChat(harness)`, resolve to TUI. Worth noting the renderer already does exactly this client-side — `TaskComposer.tsx:207` computes `requiresTuiFallback` and `:339` submits `"tui"` — so the case was known, but solved in only one client. This PR makes that default apply to clients that have no such fallback.

**Regression test:** table test over `domain.AllHarnesses` asserting a spawn with empty `RequestedMode` and a chat default succeeds for every harness, dispatching to chat only where `SupportsChat` is true.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
```diff
@@ -0,0 +1,9 @@
...
+SET default_session_mode = 'chat', updated_at = CURRENT_TIMESTAMP
```
whoisasx: **[P1] Chat-*capable* harnesses also fail when the local chat runtime or auth is not ready — again with no fallback.**

**Trigger:** a default-mode spawn on one of the 5 chat-capable harnesses where the driver probe fails.

**Impact:** `manager.go:745` → `PreflightChat` → `ensureDriverReady` (`service/chat/service.go`) → `driver.Probe(ctx)`, which is a **live** probe of the local install. For Claude Code (`adapters/chatdriver/claudeacp/driver.go:50-70`) that requires a resolvable Node runtime, a resolvable `claude` binary, and a valid ACP executable, and returns `ErrChatAuthRequired` when the auth probe reports unauthorized. Any of those fails the spawn. A user who is logged in and working fine in TUI today gets a hard spawn failure after upgrading.

**Why this is distinct from the previous comment:** a fix that only gates on `SupportsChat` does **not** address this — the harness *does* support chat, the local install just is not ready. The fallback has to key on *how the mode was chosen*, not only on driver registration.

**Correction:** treat a preflight failure as fatal only when the caller explicitly asked for chat; when the mode was inherited from the daemon default, degrade to TUI and log.

**Regression test:** spawn with empty `RequestedMode`, a chat default, a chat-capable harness, and a launcher whose `PreflightChat` returns `ErrChatDriverUnavailable` / `ErrChatAuthRequired` — assert a TUI session is created rather than an error.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
```diff
@@ -0,0 +1,9 @@
...
+WHERE id = 1;
```
whoisasx: **[P2] This blanket `UPDATE` resets a deliberate user preference, and contradicts the invariant the schema documents.**

**Trigger:** upgrade of any existing install — the `UPDATE ... WHERE id = 1` is unconditional.

**Impact:** `app_settings` has no column distinguishing "never chosen" from "deliberately chose `tui`". The only available signal, `updated_at`, is not consulted — and is itself overwritten here. So this necessarily stomps users who explicitly selected TUI in Settings. That contradicts three places that currently document the opposite:

- `0067_app_settings.sql:14` — "Defaults to `tui` so an upgrade changes nobody's workflow. Chat is only ever chosen deliberately."
- `domain/sessionmode.go:25-27` — "It stays TUI so an upgrade never changes how existing workflows behave."
- `settings.Service.SetDefaultSessionMode` — "a preference for future sessions, not an implicit migration of the present."

The PR body's "Existing sessions are untouched" is accurate about *sessions*, but it obscures that the *preference* is force-reset for every existing user.

**Correction:** make the choice explicit — either (a) new-installs-only: change the seed for fresh databases and leave existing rows alone; or (b) reset-everyone: state that in the PR body and release notes. Either way, update the three comments above so they stop asserting an invariant the code no longer holds.

**Regression test:** migrate a database whose `app_settings` was explicitly set to `tui` at version 101 and assert the intended outcome, so the decision is pinned rather than incidental.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
```diff
@@ -0,0 +1,9 @@
...
+
+-- +goose Down
+UPDATE app_settings
+SET default_session_mode = 'tui', updated_at = CURRENT_TIMESTAMP
+WHERE id = 1;
```
whoisasx: **[P3] `Down` stomps deliberate `chat` choices.**

Rollback unconditionally sets every install to `tui`, discarding the preference of users who had chosen chat *before* this upgrade. It also means down/up does not round-trip: a user on chat who rolls back and rolls forward is indistinguishable from one who never chose anything.

This is the mirror of the comment on line 4. If the new-installs-only option is taken there, `Down` should be a no-op rather than a second blanket write.

backend/internal/storage/sqlite/migrate_burned_versions_test.go
```diff
@@ -105,6 +105,7 @@ var shippedMigrations = map[int64]string{
 	99:  "0099_interface_transition_notice_acknowledgement.sql",
 	100: "0100_session_model.sql",
 	101: "0101_conversation_provider_ownership_epochs.sql",
+	102: "0102_default_session_mode_chat.sql",
```
whoisasx: **[P1] Version 102 is already taken on `main`; the merged result panics the daemon at startup.**

**Trigger:** `main` gained `0102_canonical_usage.sql` in f739ce18d (#4160), after this PR's merge base (a374897ad).

**Impact:** git auto-merges the two migration *files* cleanly, because they are different paths — only this ledger map conflicts. Reconstructing the merged migrations directory locally produced:

```
panic: goose: duplicate version 102 detected:
	migrations/0102_default_session_mode_chat.sql
	migrations/0102_canonical_usage.sql
```

raised from `migrate()` at `db.go:159` — the daemon cannot open its database at all. Resolving the conflict by keeping *both* ledger entries instead is a duplicate map key, i.e. a compile error. So the two plausible careless resolutions fail loudly rather than silently, but neither is mergeable.

**Evidence:** `git merge-tree origin/main <head>` reports `CONFLICT (content) in migrate_burned_versions_test.go`; GitHub reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. CI is green only because it ran against the pre-merge head.

**Correction:** rebase onto `main` and renumber to `0103_default_session_mode_chat.sql`, with the ledger entry `103:`. Per `AGENTS.md:93`, leave the already-merged `0102_canonical_usage.sql` untouched.

**Regression test:** `TestMigrationVersionLedger` already covers exactly this — it simply has not yet run against current `main`.

backend/internal/storage/sqlite/migrate_test.go
```diff
@@ -30,6 +30,18 @@ var expectedUsageTableColumns = map[string][]string{
 	},
 }
 
+func TestMigrateDefaultsSessionInterfaceToChat(t *testing.T) {
```
whoisasx, line 41: **[P2] The new migration silently disarms an existing regression test.**

This test asserts the migration's own effect, which is fine. The problem is a side effect on `migrate_chat_renumber_test.go:53`.

`TestMigrateRecognizesPreRenumberedChatSchema` seeds `default_session_mode = 'chat'`, runs the pre-renumber remap, and asserts the value survived — it guards the issue-#3475 class of silently-skipped migrations. Migration 0102 runs *after* that remap and unconditionally sets `chat`, so the assertion becomes unfalsifiable: if the remap logic regresses and drops the user's preference, 0102 restores `chat` and the test still passes.

**Evidence:** I copied that test and changed only its seed to `'tui'` — i.e. preservation demonstrably did *not* happen. With this migration present it still passes. With `0102_default_session_mode_chat.sql` moved aside it correctly fails:

```
zz_review_vacuity_test.go:53: default mode = "tui", want preserved chat
```

So this PR is what removes that test's power.

**Correction:** assert preservation at a version *before* the new migration (e.g. `upTo(t, db, 102)` before the final assertion, once renumbered), or have the test seed and assert a sentinel the new migration does not write.

**Regression test:** the amended `migrate_chat_renumber_test.go` itself, verified to still fail when the remap is broken.

whoisasx (changes requested):
## Changes requested

The migration itself is well-formed and correctly append-only, but flipping this particular setting is not a self-contained data change: the daemon-side spawn path has no capability fallback for a *defaulted* chat mode, so this breaks session creation for most agents. There is also a hard version collision with `main`.

### Blocking

1. **P1 — Spawns fail outright for the 22 of 27 harnesses with no chat driver.** `manager.go:738` deliberately never falls back to TUI. That is right for an *explicitly requested* chat mode; this PR makes it the default for spawns that requested nothing, so `ao spawn --agent aider` (and headless/mobile/orchestrator-issued spawns) now error instead of producing a working TUI session.
2. **P1 — Chat-capable harnesses fail too when the local chat runtime or auth is not ready.** `PreflightChat` runs a live `Probe`; a `SupportsChat`-only gate does not fix this.
3. **P1 — Migration version 102 is already taken on `main`** (`0102_canonical_usage.sql`, f739ce18d / #4160). The merged result panics the daemon at startup.

### Non-blocking but needs a decision

4. **P2** — the migration silently disarms `TestMigrateRecognizesPreRenumberedChatSchema`.
5. **P2** — the blanket `UPDATE` resets deliberate user preferences and contradicts the invariant three code comments currently document.
6. **P3** — `Down` stomps deliberate `chat` choices.

### Verification

All 8 checks are green, but that run predates current `main`, and nothing in the existing suite exercises the regression: no test wires a non-nil `SessionModeDefaults` into `Manager.Spawn` (every test either passes `RequestedMode` explicitly or leaves `m.defaults == nil`, which short-circuits to `domain.DefaultSessionMode`), and `backend/internal/service/settings` has no test files at all.

Three temporary proofs were run in an isolated worktree at `f58a9f4` and then deleted:

- A table test spawning every TUI-only harness twice — `Defaults` returning `tui` (22/22 succeeded) vs returning `chat` (22/22 failed with `spawn: <harness> has no chat driver: chat mode unsupported for harness`).
- A copy of `migrate_chat_renumber_test.go` with its seed changed to `'tui'`: passes with this migration present, fails (`default mode = "tui", want preserved chat`) with it removed.
- The reconstructed post-merge migrations directory: `panic: goose: duplicate version 102 detected`.

On PR head, `go test ./internal/session_manager/ ./internal/storage/sqlite/ ./internal/httpd/controllers/` all pass — which is the finding, not a reassurance.

Merge state is currently `CONFLICTING` / `DIRTY` in `migrate_burned_versions_test.go`.

Not covered: I did not run the Electron app or exercise a live Codex/Claude ACP chat spawn end-to-end, so finding 2's real-world *frequency* is unquantified even though the failure path is confirmed by code. I also did not evaluate orchestrator-specific behaviour in chat mode beyond the spawn gate.

illegalcall:
Updated in `93c3680e7` based on the review and the clarified product policy: Chat is preferred, with TUI fallback when inherited Chat is unavailable.

- Rebased onto current `main` and renumbered the migration to 0103.
- Omitted-mode spawns now fall back to TUI when Chat has no launcher or preflight fails; explicit Chat requests remain strict.
- Added coverage for successful inherited Chat and unsupported/unavailable/incompatible/unauthenticated fallback paths.
- Repaired the pre-renumber migration test so it asserts preference preservation before the new default migration can mask it.
- Made Down non-destructive because the previous saved preference cannot be reconstructed.
- Documented that the rollout intentionally updates the saved default to Chat on existing installations while leaving existing sessions untouched.

`go test ./...` passes, and golangci-lint reports 0 issues. Ready for re-review.

i-trytoohard:
From the NotesBot call on `2026-08-21`, this existing pr appears to be relevant.
Match confidence: `0.63` (fuzzy).

Relevant discussion excerpts:
- [14:44] Dhruv Sharma: Got it. Yeah, fix on chat UI and reviewing other PRs and picking other issues as well, and also PRing issues and sending over issues to newer contributors who are asking for issues. So done that as well. [15:05] prateek: Newer con— sorry, newer contributors asking for Yeah, yeah. [15:09] [...]
- - *Dhruv Sharma:* said he has been working on **chat UI**, reviewing other PRs, picking up issues, and routing issues to newer contributors who request them. - *Dhruv Sharma:* said he is gathering issues from the board and assigning them to contributors. - *prateek:* told *Dhruv Sharma* not to assign **high- [...]
- - *khushi* to finish the remaining **review chat UI** and sync with *aditi* after the call. - *Dhruv Sharma* to prepare a **roadmap/prioritization** for core infra/session UI work, coordinate with *Adil* on the **SQLite/PR refresh** issue, and continue contributor triage.

Source thread: https://discord.com/channels/1476302178913357958/1540413922144690196/1540413922144690196

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
whoisasx: Fixed in `13e1bca0d`. An omitted mode is now tracked as inherited; if Chat has no launcher or preflight returns `ErrChatUnsupported`, `Manager.Spawn` logs the fallback and creates a TUI session. `TestDefaultChatSpawnFallsBackToTUIWhenUnavailable` covers the unsupported-harness path, while explicit Chat remains strict.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
whoisasx: Fixed in `13e1bca0d`. Preflight failures for inherited Chat now fall back to TUI, including unavailable, incompatible, and authentication-required drivers; explicit Chat and canceled/deadline-exceeded requests still return the error. The fallback matrix is covered in `chat_spawn_test.go`.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
whoisasx: Resolved via the clarified reset-everyone policy in `13e1bca0d`: the PR body now explicitly states that existing installations' saved default is migrated to Chat, `TestMigrateUpdatesExistingSessionInterfaceDefaultToChat` pins that upgrade behavior, and the domain compatibility-default comment now distinguishes settings-read failure fallback from the saved product preference. Existing session/controller rows remain untouched.

backend/internal/storage/sqlite/migrations/0102_default_session_mode_chat.sql
whoisasx: Fixed in `13e1bca0d`. The Down migration is a no-op because the prior preference cannot be reconstructed safely. `TestMigrateRollbackPreservesSessionInterfacePreference` now applies and rolls back migration 104 and verifies a deliberate Chat preference survives; mutation-testing the Down statement to write TUI makes that test fail.

backend/internal/storage/sqlite/migrate_burned_versions_test.go
whoisasx: Fixed in `13e1bca0d`. The branch is rebased onto current `main`; main's `0103_review_run_cdc.sql` remains unchanged and this migration is now `0104_default_session_mode_chat.sql`, with ledger entries for both 103 and 104.

backend/internal/storage/sqlite/migrate_test.go
whoisasx: Fixed in `13e1bca0d`. `TestMigrateRecognizesPreRenumberedChatSchema` now asserts the Chat preference immediately after `repairRenumberedChatMigrationHistory`, before migration 104 can write the new product default, so the preservation assertion remains capable of catching a reset in the repair.

whoisasx (approved):
Re-reviewed `13e1bca0d` against `main` at `35ee1fe42`.

The prior blockers are addressed: inherited Chat defaults fall back to TUI for unsupported or failed preflight paths while explicit Chat remains strict; the migration is conflict-free at version 104; the reset-everyone preference policy is explicit and tested; rollback preserves the saved preference; and the pre-renumber preservation assertion remains effective before the new default migration runs.

All six review threads are resolved. Focused tests, migration mutation coverage, race checks, full backend tests, and lint were verified locally; all eight required CI checks pass. No remaining blocking findings.

Issue #4187, opened 2026-08-20
bug(daemon): reaper WARN-spam on handle-less sessions floods the 'not ready' details buffer; boot stuck un-ready 10+ min on ~11 sessions (post-#4071)


i-trytoohard:
## Summary

Post-[#4071](https://github.com/Untrivial-ai/agent-orchestrator/pull/4071) build (verified: banner shows the new copy "AO daemon is still starting. Session recovery can take a while." and the details view keeps updating instead of freezing — both #4071 behaviors). Reporter's daemon stayed un-ready for **10+ minutes** on ~11 sessions (macOS, nightly), and the "not ready" details panel shows **only** reaper WARN spam — no boot-progress landmarks — because the reaper warns every 5s forever on permanently-unprobeable sessions and floods the supervisor's 12k stderr capture buffer.

Two distinct problems in one report:

1. **Reaper warn-spam on handle-less sessions floods the startup details buffer** (High confidence, traced).
2. **Daemon un-ready 10+ min for ~11 sessions on macOS nightly** — #4071's in-place relaunch should be far faster; unclear whether Reconcile is slow or hung (Medium confidence — evidence likely evicted from the buffer by problem 1; needs full stderr).

## Source

**Reported by:** Rishet11 in Discord `#bug-triage` (thread "Wasnt this issue already resolved?"), 2026-08-20 20:36–20:46 IST
**Machine:** macOS (reporter is a mac + nightly user); AO version TBD (asked)
**Analyzed against:** `e9918fc3a` (origin/main, 2026-08-20)
**Confidence:** High (problem 1) / Medium (problem 2)

## Observed

Banner: `AO daemon is not ready yet / AO daemon is still starting. Session recovery can take a while.` Screenshot + two details dumps spanning 20:36:25→20:46:20 IST (one continuous boot — reaper warns present from the first captured line).

The entire visible details buffer is:

```
level=WARN msg="reaper: session has no runtime handle metadata, skipping" session=agent-orchestrator-10   (every 5s)
level=WARN msg="reaper: session has no runtime handle metadata, skipping" session=agent-orchestrator-11   (every 5s)
level=DEBUG msg="scm observer: project has no supported SCM origins" project=aohack                        (every 30s)
```

No `browser runtime: listening`, no `supervisor: listening`, no `daemon listening` — i.e. no evidence the HTTP bind ever landed within the captured window. Reporter: "this is just growing."

## Root cause (problem 1 — traced)

[...]

PR #4242, opened 2026-08-21
fix(reaper): dedupe missing-handle warnings

mknayam:
## What

Deduplicate missing-runtime-handle warnings per session and daemon run.

## Why

Repeated reaper warnings can evict useful startup diagnostics from the supervisor output buffer.

This addresses the warning-spam portion of [#4187](https://github.com/Untrivial-ai/agent-orchestrator/issues/4187). The startup reconciliation hang remains out of scope.

## How

Track warned session IDs in a mutex-protected, reaper-local set.

## Testing

- `cd backend && go test -race ./internal/observe/reaper -count=1`
- `npm run lint`

## Checklist

- [x] Branched from `main`

### DIFF
--- a/backend/internal/observe/reaper/reaper.go
+++ b/backend/internal/observe/reaper/reaper.go
@@ -9,6 +9,7 @@ package reaper
 import (
 	"context"
 	"log/slog"
+	"sync"
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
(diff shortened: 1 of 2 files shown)


mknayam:
I traced and fixed the reaper warning spam.

The reaper runs every five seconds and previously logged the same warning on every tick for each non-terminated TUI session without a runtime handle. I added a small in-memory set keyed by session ID, so each affected session now warns only once per daemon run.

The set is mutex-protected because the background reaper and startup reconciliation can call `Tick` on the same reaper instance. The warning is skipped before calling the logger; existing log output is not modified.

I added a regression test covering two handle-less sessions across multiple ticks. Before the fix, it produced six warnings; it now produces one warning per session.

Tests passing:

- `go test ./internal/observe/reaper -count=1`
- `go test -race ./internal/observe/reaper -count=1`
- `git diff --check`

This only fixes the warning flood. It does not fix the reported 10–15 minute startup hang.

The reporter’s SIGQUIT dump placed the main goroutine inside `RestoreAll`, around the error path after `workspace.Restore` failed. We still need to determine why it remains there—possibly repeated restore failures, a slow Git/worktree operation, or unusually expensive error logging.

The reduced log noise should make the actual restore errors visible during the next reproduction. I suggest keeping the startup hang open, either in this issue or as a separate follow-up.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #4243, opened 2026-08-21
feat(frontend): syntax-highlight the diff review panel

Vaibhaav-Tiwari:
## Summary
- Adds syntax highlighting to the file-diff review panel (`SessionFilesView`), composed with the existing exact-changed-word highlight.
- Reuses the app's existing `lowlight`/highlight.js engine and theme (`code-highlight.ts`, `code-theme.css`) rather than introducing Shiki — this app's CSP (`script-src 'self'`, no `wasm-unsafe-eval`) blocks Shiki's default WASM engine, which is exactly why that engine was chosen for chat code blocks in the first place. No new dependency is added.
- Diff lines are tokenized per hunk-side blob (not per line), so highlight.js's grammar state carries correctly across line breaks — a multi-line comment or string colors correctly even when a diff line inside it changed.
- `code-theme.css` is extended to a second `.diff-code` scope, so diff colors and chat code-block colors are the same palette and both follow the app's light/dark toggle instantly with no re-tokenizing.

## Test plan
- [x] `diff-highlight.test.ts` — 16 new unit tests for the pure line-splitting/composition functions
- [x] `useDiffHighlight.test.ts` — 3 new hook tests, including a real cross-line-state test through the actual highlight.js engine
- [x] Existing `SessionFilesView.test.tsx` (31 tests) passes unmodified — no regressions
- [x] Full frontend typecheck and test suite run clean (pre-existing, unrelated failures only: macOS-only packaging/updater tests on a non-macOS dev machine, and a separate `landing` sub-app missing its own dependency)
- [x] Manually verified in the real Electron app (isolated dev mode) — confirmed by the requester

### DIFF
--- a/frontend/src/renderer/components/SessionFilesView.tsx
+++ b/frontend/src/renderer/components/SessionFilesView.tsx
@@ -38,14 +38,20 @@ import {
 	type WorkspaceFileSummary,
 } from "../hooks/useSessionWorkspaceFiles";
 import { useParsedDiff } from "../hooks/useParsedDiff";
+import { useDiffHighlight } from "../hooks/useDiffHighlight";
 import { cn } from "../lib/utils";
-import type { DiffRow, DiffRowKind, DiffSegment } from "../lib/diff-parser";
+import type { DiffRow, DiffRowKind } from "../lib/diff-parser";
+import type { DiffRun } from "../lib/diff-highlight";
 import type { DiffSelectionLine } from "../../shared/diff-selection";
 import { Accordion, AccordionContent, AccordionItem, AccordionTrigger } from "./ui/accordion";
 import { subscribeWorkspaceFileChanges } from "../lib/workspace-file-events";
 import { Button } from "./ui/button";
 import { DiffSelectionMenu } from "./DiffSelectionMenu";
 import { Input } from "./ui/input";
+// The token colours live with the engine rather than with one caller: they are
+// scoped under `.chat-code`/`.diff-code`, so anything rendering these class names
+// has to be inside one of those containers and has to have loaded this sheet.
+import "./chat/code-theme.css";
 
 type WorkspaceFileDetail = components["schemas"]["WorkspaceFileResponse"] & {
 	previousPath?: string;
@@ -746,6 +752,7 @@ function DiffView({
 	const [menuState, setMenuState] = useState<DiffViewMenuState | null>(null);
 	const shouldVirtualize = !split && rows.length > ROW_VIRTUALIZE_THRESHOLD;
 	const { listRef, virtualizer } = useSharedScrollRowVirtualizer(containerRef, rows.length, shouldVirtualize);
+	const runs = useDiffHighlight(rows, path);
 
 	useEffect(() => {
 		const onSelectionChange = () => {
@@ -805,12 +812,12 @@ function DiffView({
 				</div>
 			) : null}
 			<div
-				className="session-files-diff-scrollbar overflow-x-auto overflow-y-visible bg-terminal font-mono text-xs leading-row text-terminal-foreground"
+				className="diff-code session-files-diff-scrollbar overflow-x-auto overflow-y-visible bg-terminal font-mono text-xs leading-row text-terminal-foreground"
 				onContextMenu={onContextMenu}
 				ref={containerRef}
 			>
 				{split ? (
-					<SplitDiff annotation={annotation} path={path} previousPath={previousPath} rows={rows} t={t} />
+					<SplitDiff annotation={annotation} path={path} previousPath={previousPath} rows={rows} runs={runs} t={t} />
 				) : shouldVirtualize ? (
 					<div
 						className={cn("relative", !wrap && "min-w-max")}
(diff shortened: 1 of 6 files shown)


illegalcall:
Reproduction evidence for the three syntax-highlighting review findings, verified on current head `0a09d6d3f404e49b2b91b0ba8bfa7e6ad5854e6d`.

I added temporary Vitest reproductions locally (not committed) and ran them against this PR checkout.

### 1. Old/new context state is collapsed into one `runs[rowIndex]`

Input:

```diff
@@ -1,3 +1,3 @@
-/* old-only
+const value = 1;
 shared context
 end */
```

The real TypeScript highlighter confirms that `shared context` is `hljs-comment` in the old-side blob and is not a comment in the new-side blob. The reproduction then calls `useDiffHighlight` and expects the old-side token to remain available:

```ts
const rows = parseUnifiedDiff(diff);
const contextIndex = rows.findIndex((row) => row.text === "shared context");
const { result } = renderHook(() => useDiffHighlight(rows, "src/App.ts"));
expect(classesOf(result.current[contextIndex])).toContain("hljs-comment");
```

Result:

```text
FAIL preserves old-side syntax state for a shared context row
AssertionError: expected [] to include 'hljs-comment'
```

This matches `useDiffHighlight.ts:49-62`: old and new blobs are tokenized separately, but both write to the same `runs[line.rowIndex]`, with the new pass last. `SessionFilesView.tsx:1045-1062` then passes that same run array to both split columns.

### 2. An unavailable highlighter triggers another render/retry

The reproduction mocks the shared engine's documented fallback (`highlightSync()` and `highlight()` return `undefined`). There are two hunk sides, so the stable fallback should make two initial attempts and then retain plain text. Later unexpected calls are kept pending so the test terminates instead of continuing the loop.

```ts
highlightMock.mockResolvedValueOnce(undefined).mockReturnValue(new Promise(() => {}));
const rows = parseUnifiedDiff("@@ -1,1 +1,1 @@\n-const before = 0;\n+const after = 1;\n");
const { unmount } = renderHook(() => useDiffHighlight(rows, "src/App.ts"));
await waitFor(() => expect(highlightMock.mock.calls.length).toBeGreaterThan(2));
unmount();
expect(highlightMock).toHaveBeenCalledTimes(2);
```

Result:

```text
FAIL settles on plain text after the engine reports highlighting unavailable
AssertionError: expected "vi.fn()" to be called 2 times, but got 4 times
```

The extra calls occur because `useDiffHighlight.ts:53-56` invokes `reread()` without checking whether `highlight()` produced a tree. If subsequent attempts also resolve `undefined`, each completion schedules the next retry.

### 3. Extension-changing renames use the destination grammar on the old side

Input rename: `src/old.ts` -> `docs/new.md` with this diff:

```diff
@@ -1,1 +1,1 @@
-const oldValue = 0;
+# heading
```

The reproduction first confirms `languageForPath("src/old.ts") === "typescript"` and `languageForPath("docs/new.md") === "markdown"`, then exercises the current hook exactly as `SessionFilesView` does:

```ts
const rows = parseUnifiedDiff(diff);
const deletedIndex = rows.findIndex((row) => row.kind === "del");
const { result } = renderHook(() => useDiffHighlight(rows, "docs/new.md"));
expect(classesOf(result.current[deletedIndex])).toContain("hljs-keyword");
```

Result:

```text
FAIL uses the previous extension's grammar for an extension-changing rename
AssertionError: expected [] to include 'hljs-keyword'
```

`previousPath` reaches `DiffView`, but `useDiffHighlight(rows, path)` accepts only the destination path, so both hunk sides are parsed as Markdown.

Commands used:

```bash
npm test -- src/renderer/hooks/useDiffHighlight.pr4243-repro.test.ts
npm test -- src/renderer/hooks/useDiffHighlight.pr4243-failure-repro.test.ts
```

Combined result: **3 reproductions, 3 expected failures**. The PR's existing targeted tests still pass; these cases are currently uncovered.

Vaibhaav-Tiwari:
All three verified and fixed in 31b2e4128.

1. **Old/new context collapse** — `useDiffHighlight` now returns `{ oldSide, newSide }` (tokenized and stored independently per row) instead of one `runs[rowIndex]` that the new-side pass overwrote last. `SplitDiff` now reads `oldRuns`/`newRuns` for its respective column instead of the same array for both. Reproduced your `shared context` example as a test first (old side keeps `hljs-comment`, new side doesn't) — failed against the original code, passes now.

2. **Retry loop** — `highlightSide` only calls `reread()` when `highlight()` actually resolves a tree, not unconditionally. Added a test mirroring your mock (`highlightSync` always undefined, `highlight` resolves undefined twice) asserting exactly 2 calls (one per hunk side), not 4.

3. **Rename grammar** — `useDiffHighlight` now takes `previousPath` and resolves the old side's language from it (falling back to `path` when there's no rename), so an extension-changing rename tokenizes each side with its own file's grammar instead of the destination's for both.

Ported the existing tests to the new `oldSide`/`newSide` shape; full suite and typecheck clean, no regressions.

illegalcall (approved):

PR #4244, opened 2026-08-21
feat(browser): add an act primitive that resolves an instruction and retries stale refs

Vaibhaav-Tiwari:
## Summary
Phase 4 of the browser UX plan (deliberately deferred so it would get its own design pass instead of a rushed bolt-on, inspired by browser-use's perceive-decide-act loop):

- **New `ao browser act "<instruction>"` primitive**: resolves which element to act on and performs the action, instead of the agent manually chaining a `snapshot` call, reading the accessibility tree text itself, then an explicit ref-targeted action call. Matching is fully deterministic (no new LLM call) — parses the compact snapshot's `role "name" [ref=eN]` lines and scores candidates against the instruction's role/name tokens (exact-name match > substring > fuzzy token overlap). If the match is ambiguous or absent, it returns the candidates/snapshot instead of guessing on a mutating action — exactly matching today's manual fallback.
- **One automatic retry on a stale reference**: re-snapshots, re-matches the *original* instruction, retries once. If that also fails, rethrows as-is rather than looping — mirrors `ensureNativeActiveTab`'s existing "retry once, then surface reality" convention already in `browser-view-host.ts`.
- **v1 covers the `{ref[, text]}`-shaped action family**: `click` (default), `dblclick`, `focus`, `hover`, `fill`, `type`, `check`, `uncheck`. `drag` and `select` are excluded — they need harder matching problems (two independently-resolved targets, or an option's own text within a control) and are left for a follow-up rather than a half-baked heuristic.
- **`--nth <index>`** disambiguates a tie deterministically instead of declining, for cases like "the second Add to Cart button."
- All new logic lives in the Electron main process (`browser-view-host.ts` + new `browser-act-matcher.ts`) next to the existing `snapshot`/`click` cases — the Go service layer stays a clean, business-logic-free passthrough, just adding `act` to its action allowlist and CLI wiring.
- Bundled prerequisite fix: `parseAgentBrowserJSON` used to discard the automation runtime's own error `.code`, collapsing every failure to a generic code — fixed so `STALE_REFERENCE` (among others) is a real, reachable code for the first time, which the retry logic above depends on.
- Skill doc (`browser.md`) updated to steer agents toward `act` first, reframing the manual `snapshot`→`click` flow as the fallback for `ambiguous`/`no-match` outcomes or actions `act` doesn't cover yet.

**Out of scope**: the same phase 4 plan's other item ("element-selection UX beyond click-on-page," a *human*-facing feature) is already covered by an open PR building on the annotation multiselect branch — untouched here, since this PR only touches the agent-facing accessibility-tree/ref-based automation path, which is architecturally separate from human annotation.

## Test plan
- [x] `npx vitest run` — new `browser-act-matcher.test.ts` (12 tests: parsing, exact/fuzzy matching, ambiguous ties, no-match, `--nth` tie-break) and new `browser:act` tests in `browser-view-host.test.ts` (8 tests: happy path, retry-succeeds, retry-exhausted-rethrows, ambiguous, no-match, `--nth` end-to-end, argument validation) — all passing, plus the full existing suite (103/105, 2 pre-existing skips)
- [x] `go build ./...` clean; `go test` passing on `internal/service/browser`, `internal/httpd/controllers`, `internal/cli`, `internal/telemetrymeta` (one pre-existing, unrelated failure in `internal/httpd/controllers` — `TestProjectsAPI_Clone` — confirmed to fail identically on unmodified `main` in this sandbox, unrelated to this change)
- [ ] Manual verification in the real desktop app — the real `agent-browser` binary's actual compact-snapshot text format was inferred from this repo's own test fixtures (which are themselves slightly inconsistent about quoting), not observed live, since a live CDP-connected target wasn't available in this environment. Recommend a manual pass before merge: open a real page, run `ao browser act "..."` against it, and separately confirm the stale-ref retry path by mutating the DOM between the internal snapshot and action.

### DIFF
--- a/backend/internal/cli/browser.go
+++ b/backend/internal/cli/browser.go
@@ -97,6 +97,38 @@ func newBrowserCommand(ctx *commandContext) *cobra.Command {
 	snapshot.Flags().BoolVar(&interactiveOnly, "interactive", false, "include only actionable elements")
 	cmd.AddCommand(snapshot)
 
+	var actVerb, actValue string
+	var actNth int
+	var actNthSet bool
+	act := &cobra.Command{
+		Use:   "act <instruction>",
+		Short: "Resolve an element by description and act on it, snapshotting and retrying automatically",
+		Long: "Snapshots, finds the best-matching element for <instruction> against role/name/text\n" +
+			"(no LLM call — deterministic matching, same as reading a snapshot yourself), and performs\n" +
+			"--action on it (default click). Retries once on a stale reference. If the match is\n" +
+			"ambiguous or absent, returns the candidates or snapshot instead of guessing — fall back\n" +
+			"to a manual snapshot/click, retry with --nth, or refine the instruction.",
+		Args: exactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			actArgs := map[string]any{"instruction": args[0], "action": actVerb}
+			if actValue != "" {
+				actArgs["value"] = actValue
+			}
+			if actNthSet {
+				actArgs["nth"] = actNth
+			}
+			return ctx.runBrowserAction(cmd, "act", actArgs, jsonOutput)
+		},
+	}
+	act.Flags().StringVar(&actVerb, "action", "click", "verb to perform on the matched element (click, dblclick, focus, hover, fill, type, check, uncheck)")
+	act.Flags().StringVar(&actValue, "value", "", "text to fill/type; required when --action is fill or type")
+	act.Flags().IntVar(&actNth, "nth", 0, "0-based index to disambiguate when multiple candidates match equally")
+	act.PreRunE = func(cmd *cobra.Command, _ []string) error {
+		actNthSet = cmd.Flags().Changed("nth")
+		return nil
+	}
+	cmd.AddCommand(act)
+
 	cmd.AddCommand(&cobra.Command{
 		Use:   "click <ref>",
 		Short: "Click an element reference from the latest snapshot",
@@ -579,6 +611,9 @@ func (c *commandContext) runBrowserAction(cmd *cobra.Command, action string, arg
 }
 
 func writeBrowserResult(cmd *cobra.Command, action string, result map[string]any) error {
+	if action == "act" {
+		return writeBrowserActResult(cmd, result)
+	}
 	if action == "snapshot" {
 		if text, ok := result["text"].(string); ok {
 			_, err := fmt.Fprintln(cmd.OutOrStdout(), browserUntrustedText(text))
@@ -656,6 +691,50 @@ func browserUntrustedText(value string) string {
 	value = strings.ReplaceAll(value, browserUntrustedBegin, `\u003c`+browserUntrustedBegin[1:])
 	value = strings.ReplaceAll(value, browserUntrustedEnd, `\u003c`+browserUntrustedEnd[1:])
 	return browserUntrustedBegin + "\n" + value + "\n" + browserUntrustedEnd
+}
+
+func writeBrowserActResult(cmd *cobra.Command, result map[string]any) error {
+	outcome, _ := result["outcome"].(string)
+	switch outcome {
+	case "matched":
+		candidate, _ := result["candidate"].(map[string]any)
+		role, _ := candidate["role"].(string)
+		name, _ := candidate["name"].(string)
+		ref, _ := result["resolvedRef"].(string)
+		retried, _ := result["retried"].(bool)
+		suffix := ""
+		if retried {
+			suffix = " (after retrying a stale reference)"
+		}
+		_, err := fmt.Fprintf(cmd.OutOrStdout(), "Acted on: %s %q [ref=%s]%s\n", role, browserUntrustedText(name), ref, suffix)
+		return err
+	case "ambiguous":
+		candidates, _ := result["candidates"].([]any)
+		if _, err := fmt.Fprintln(cmd.OutOrStdout(), "Ambiguous — multiple elements matched equally well:"); err != nil {
+			return err
+		}
+		for _, raw := range candidates {
+			candidate, _ := raw.(map[string]any)
+			role, _ := candidate["role"].(string)
+			name, _ := candidate["name"].(string)
+			ref, _ := candidate["ref"].(string)
+			if _, err := fmt.Fprintf(cmd.OutOrStdout(), "  %s %q [ref=%s]\n", role, browserUntrustedText(name), ref); err != nil {
+				return err
+			}
+		}
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), "Pick a ref and act/click directly, retry with --nth, or refine the instruction.")
+		return err
+	case "no-match":
+		if _, err := fmt.Fprintln(cmd.OutOrStdout(), "No element matched that instruction. Latest snapshot:"); err != nil {
+			return err
+		}
+		text, _ := result["snapshot"].(string)
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), browserUntrustedText(text))
+		return err
+	default:
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), "Browser act completed.")
+		return err
+	}
 }
 
 func writeBrowserNetworkResult(cmd *cobra.Command, action string, result map[string]any) error {
--- a/backend/internal/cli/browser.go
+++ b/backend/internal/cli/browser.go
@@ -111,7 +111,7 @@ func newBrowserCommand(ctx *commandContext) *cobra.Command {
 		Args: exactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			actArgs := map[string]any{"instruction": args[0], "action": actVerb}
-			if actValue != "" {
+			if cmd.Flags().Changed("value") {
 				actArgs["value"] = actValue
 			}
 			if actNthSet {
@@ -706,7 +706,7 @@ func writeBrowserActResult(cmd *cobra.Command, result map[string]any) error {
 		if retried {
 			suffix = " (after retrying a stale reference)"
 		}
-		_, err := fmt.Fprintf(cmd.OutOrStdout(), "Acted on: %s %q [ref=%s]%s\n", role, browserUntrustedText(name), ref, suffix)
+		_, err := fmt.Fprintf(cmd.OutOrStdout(), "Acted on: %s [ref=%s]%s\n%s\n", role, ref, suffix, browserUntrustedText(name))
 		return err
 	case "ambiguous":
 		candidates, _ := result["candidates"].([]any)
@@ -718,7 +718,7 @@ func writeBrowserActResult(cmd *cobra.Command, result map[string]any) error {
 			role, _ := candidate["role"].(string)
 			name, _ := candidate["name"].(string)
 			ref, _ := candidate["ref"].(string)
-			if _, err := fmt.Fprintf(cmd.OutOrStdout(), "  %s %q [ref=%s]\n", role, browserUntrustedText(name), ref); err != nil {
+			if _, err := fmt.Fprintf(cmd.OutOrStdout(), "  %s [ref=%s]\n%s\n", role, ref, browserUntrustedText(name)); err != nil {
 				return err
 			}
 		}
--- a/frontend/src/main/agent-browser-runtime.ts
+++ b/frontend/src/main/agent-browser-runtime.ts
@@ -733,7 +733,7 @@ export function parseAgentBrowserJSON(stdout: string): AgentBrowserJSONResult {
 	}
 	if (!isRecord(envelope)) throw runtimeError("AGENT_BROWSER_INVALID_OUTPUT", "Browser automation returned invalid output");
 	if (envelope.success === false) {
-		throw runtimeError(errorCode(envelope.error) ?? "AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
+		throw runtimeError(staleReferenceCode(envelope.error) ?? "AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
 	}
 	const boundary = validContentBoundary(envelope._boundary);
 	const result: Record<string, unknown> = isRecord(envelope.data) ? { ...envelope.data } : { value: envelope.data };
@@ -784,13 +784,11 @@ function stringError(value: unknown): string {
 	return "";
 }
 
-// Preserves a structured error code (e.g. STALE_REFERENCE) from the real
-// agent-browser binary's JSON error envelope when one is present, instead of
-// always collapsing every failure to the generic AGENT_BROWSER_COMMAND_FAILED —
-// callers that need to distinguish failure kinds (e.g. the `act` primitive's
-// stale-ref retry) can only do that if the code survives this far.
-function errorCode(value: unknown): string | undefined {
-	return isRecord(value) && typeof value.code === "string" ? value.code : undefined;
+// Preserve only the stale-reference signal needed by `act`. Other native
+// command failures stay generic so tab drift and close recovery paths continue
+// to recognize AGENT_BROWSER_COMMAND_FAILED.
+function staleReferenceCode(value: unknown): "STALE_REFERENCE" | undefined {
+	return isRecord(value) && value.code === "STALE_REFERENCE" ? "STALE_REFERENCE" : undefined;
 }
 
 function isRecord(value: unknown): value is Record<string, unknown> {
--- a/frontend/src/main/agent-browser-runtime.ts
+++ b/frontend/src/main/agent-browser-runtime.ts
@@ -733,7 +733,7 @@ export function parseAgentBrowserJSON(stdout: string): AgentBrowserJSONResult {
 	}
 	if (!isRecord(envelope)) throw runtimeError("AGENT_BROWSER_INVALID_OUTPUT", "Browser automation returned invalid output");
 	if (envelope.success === false) {
-		throw runtimeError("AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
+		throw runtimeError(errorCode(envelope.error) ?? "AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
 	}
 	const boundary = validContentBoundary(envelope._boundary);
 	const result: Record<string, unknown> = isRecord(envelope.data) ? { ...envelope.data } : { value: envelope.data };
@@ -782,6 +782,15 @@ function stringError(value: unknown): string {
 	if (typeof value === "string") return value;
 	if (isRecord(value) && typeof value.message === "string") return value.message;
 	return "";
+}
+
+// Preserves a structured error code (e.g. STALE_REFERENCE) from the real
+// agent-browser binary's JSON error envelope when one is present, instead of
+// always collapsing every failure to the generic AGENT_BROWSER_COMMAND_FAILED —
+// callers that need to distinguish failure kinds (e.g. the `act` primitive's
+// stale-ref retry) can only do that if the code survives this far.
+function errorCode(value: unknown): string | undefined {
+	return isRecord(value) && typeof value.code === "string" ? value.code : undefined;
 }
 
 function isRecord(value: unknown): value is Record<string, unknown> {
--- a/frontend/src/main/agent-browser-runtime.ts
+++ b/frontend/src/main/agent-browser-runtime.ts
@@ -784,11 +784,14 @@ function stringError(value: unknown): string {
 	return "";
 }
 
+const STALE_REFERENCE_MESSAGE = /^(?:Unknown ref:|Could not locate element with)/;
+
 // Preserve only the stale-reference signal needed by `act`. Other native
 // command failures stay generic so tab drift and close recovery paths continue
 // to recognize AGENT_BROWSER_COMMAND_FAILED.
 function staleReferenceCode(value: unknown): "STALE_REFERENCE" | undefined {
-	return isRecord(value) && value.code === "STALE_REFERENCE" ? "STALE_REFERENCE" : undefined;
+	if (isRecord(value) && value.code === "STALE_REFERENCE") return "STALE_REFERENCE";
+	return typeof value === "string" && STALE_REFERENCE_MESSAGE.test(value) ? "STALE_REFERENCE" : undefined;
 }
 
 function isRecord(value: unknown): value is Record<string, unknown> {
--- a/frontend/src/main/browser-act-matcher.ts
+++ b/frontend/src/main/browser-act-matcher.ts
@@ -1,0 +1,118 @@
+// Pure, deterministic matching for the `act` action: given a natural-language
+// instruction ("the submit button") and a compact accessibility snapshot's text,
+// resolve which `[ref=eN]` it refers to — or say plainly that it can't, so the
+// caller can fall back to a real snapshot rather than guessing on a mutating
+// action. No LLM call, no Electron/session imports, fully testable with plain
+// fixture strings.
+
+export type SnapshotEntry = {
+	role: string;
+	name: string;
+	ref: string;
+};
+
+export type ActCandidate = {
+	role: string;
+	name: string;
+	ref: string;
+	score: number;
+};
+
+export type ActMatchResult =
+	| { outcome: "matched"; candidate: ActCandidate }
+	| { outcome: "ambiguous"; candidates: ActCandidate[] }
+	| { outcome: "no-match" };
+
+// Real fixtures observed in this codebase's own tests are inconsistent about
+// quoting (`- button "Open" [ref=e1]` vs `button Save [ref=e1]`), so both a
+// quoted and a bare name are accepted. Lines with no `[ref=eN]` (headings,
+// static text, containers) are not candidates at all.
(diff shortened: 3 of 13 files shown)


frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,118 @@
...
+export type ActMatchResult =
+	| { outcome: "matched"; candidate: ActCandidate }
+	| { outcome: "ambiguous"; candidates: ActCandidate[] }
+	| { outcome: "no-match" };
+
+// Real fixtures observed in this codebase's own tests are inconsistent about
+// quoting (`- button "Open" [ref=e1]` vs `button Save [ref=e1]`), so both a
+// quoted and a bare name are accepted. Lines with no `[ref=eN]` (headings,
+// static text, containers) are not candidates at all.
+const SNAPSHOT_LINE = /-?\s*(?<role>[a-zA-Z][\w-]*)\s+(?:"(?<quoted>[^"]*)"|(?<bare>[^[]*?))\s*\[ref=(?<ref>e\d+)\]/;
```
Prasad-D-Ware: **Blocking:** this doesn't match the real `agent-browser` output.

`\[ref=(?<ref>e\d+)\]` requires the bracket to *begin* with `ref=`. The real binary emits `ref` as one field inside a shared attribute bracket:

```
- heading "Hi" [level=1, ref=e1]
- button "Nope" [disabled, ref=e3]
- checkbox [checked=true, ref=e5]
- combobox [expanded=false, ref=e6]: A
  - option "A" [selected, ref=e11]
- switch "Toggle" [checked=true, ref=e10]
```

(captured live from the bundled 0.33.1 binary via `snapshot -i --compact`, not from fixtures)

On that snapshot, 7 of 11 elements are dropped. The practical consequence: **every checkbox, radio, and switch always carries `checked=`, so `--action check`/`uncheck` can never resolve a target** - two of the eight verbs this PR advertises are dead on arrival. Same for disabled buttons, headings, options, and expanded comboboxes.

Minimal fix:

```
\[(?:[^\]]*,\s*)?ref=(?<ref>e\d+)\]
```

Better fix: don't parse the text at all. The `--json` envelope already includes a structured map that the host is already passing through on the `snapshot` case:

```json
"refs": {"e2": {"name": "Sign In", "role": "button"}, "e5": {"name": "", "role": "checkbox"}}
```

Matching against `refs` removes the regex, removes the quoted-vs-bare ambiguity this comment describes, and removes the whole "format inferred from fixtures" risk in one move.

frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export function matchInstruction(instruction: string, snapshotText: string, opts: { nth?: number } = {}): ActMatchResult {
+	const entries = parseSnapshotEntries(snapshotText);
+	const { roleHint, nameHint, nameTokens } = parseInstruction(instruction);
+	const candidates = scoreCandidates(entries, roleHint, nameHint, nameTokens);
+	if (candidates.length === 0) return { outcome: "no-match" };
+
+	const [top, second] = candidates;
+	const isConfidentMatch =
+		candidates.length === 1 ||
```
Prasad-D-Ware: **Blocking:** `candidates.length === 1` bypasses `MATCH_SCORE_FLOOR` entirely.

The floor exists to stop junk matches, but a lone candidate skips it. A role-hint-only match scores 3 (below the floor of 4) and still fires:

```
snapshot: - button "Confirm delete" [ref=e1]
act "the cancel button"  =>  matched: button "Confirm delete" (score 3)
```

Combined with the `SNAPSHOT_LINE` issue above, this is reachable on an ordinary page: with the disabled button dropped by the parser, `act "the Nope button"` resolves to the only surviving button and **clicks "Sign In"**.

Guessing on a mutating action is the one thing this design is built not to do, so I don't think the short-circuit is worth keeping - a single candidate that only clears the role bar should be `ambiguous`/`no-match`, not an auto-click. The existing "reports no-match when nothing scores at all" test only covers score 0, so nothing catches this; worth a test at score 3 specifically.

frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,118 @@
...
+		.filter(Boolean);
+}
+
+function parseInstruction(instruction: string): { roleHint?: string; nameHint: string; nameTokens: Set<string> } {
+	let tokens = tokenize(instruction);
+	if (tokens.length > 0 && LEADING_VERBS.has(tokens[0])) tokens = tokens.slice(1);
+	tokens = tokens.filter((token) => !ARTICLES.has(token));
+
+	let roleHint: string | undefined;
+	const roleIndex = tokens.findIndex((token) => ARIA_ROLES.has(token));
```
Prasad-D-Ware: `findIndex` takes the *first* ARIA-role token in the instruction, but in English phrasing the role is the trailing noun and earlier role-named words belong to the element's name.

```
act "the Save Image button"
  => roleHint="image", nameHint="save button"
  => image "Save"        scores 9   <- picked, confidently
  => button "Save Image" scores 1
```

Taking the last role token (or only treating a trailing token as a role hint) fixes this class. Same shape hits "the Save Link button", "the Close Dialog button", etc.

frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return entries;
+}
+
+const ARIA_ROLES = new Set([
+	"button", "link", "textbox", "checkbox", "radio", "combobox", "listbox",
+	"option", "tab", "menuitem", "heading", "image", "dialog", "switch",
+	"slider", "searchbox",
+]);
+const LEADING_VERBS = new Set(["click", "tap", "press", "fill", "type", "select", "check", "uncheck", "hover", "focus"]);
```
Prasad-D-Ware, line 42: Leading-verb stripping can eat a word that's part of the element's name:

```
snapshot: - button "Select all" / - button "Clear all"
act "select all"  =>  strips "select", nameHint="all"
                  =>  both score 6  =>  permanently ambiguous
```

So `act "select all"` can never click a "Select all" button. Same for "Check out", "Type a message", "Press kit", "Link account". Only stripping the verb when the remaining tokens still match something, or scoring both with and without the strip and keeping the better result, would avoid it.

frontend/src/main/agent-browser-runtime.ts
```diff
@@ -733,7 +733,7 @@ export function parseAgentBrowserJSON(stdout: string): AgentBrowserJSONResult {
 	}
 	if (!isRecord(envelope)) throw runtimeError("AGENT_BROWSER_INVALID_OUTPUT", "Browser automation returned invalid output");
 	if (envelope.success === false) {
-		throw runtimeError("AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
+		throw runtimeError(errorCode(envelope.error) ?? "AGENT_BROWSER_COMMAND_FAILED", stringError(envelope.error) || "Browser automation failed");
```
Prasad-D-Ware: This "bundled prerequisite fix" has blast radius well outside `act`, and nothing here covers it.

Three existing recovery paths gate on the code being exactly `AGENT_BROWSER_COMMAND_FAILED`:

- `browser-view-host.ts:638` and `:649` - the tab-registry drift resync, whose comment cites a live repro ("Tab tX not found") and warns that failing it "permanently wedges the whole session"
- `browser-view-host.ts:1240` - the tab-close fallback, whose comment says letting it bubble "left the tab stuck open with no way for the user to close it at all"

If the binary attaches its own `.code` to those failures - and the premise of this change is that it does attach codes - `isAgentBrowserCommandFailure` now returns false and both recoveries silently stop firing, reintroducing exactly the bugs those comments document.

`agent-browser-runtime.test.ts` wasn't touched, so there's no coverage either way. Either narrow the passthrough to the codes that are actually needed (`STALE_REFERENCE`), or add a test pinning that the drift/close failures still reach those guards.

backend/internal/cli/browser.go
```diff
@@ -658,6 +693,50 @@ func browserUntrustedText(value string) string {
...
+		candidate, _ := result["candidate"].(map[string]any)
+		role, _ := candidate["role"].(string)
+		name, _ := candidate["name"].(string)
+		ref, _ := result["resolvedRef"].(string)
+		retried, _ := result["retried"].(bool)
+		suffix := ""
+		if retried {
+			suffix = " (after retrying a stale reference)"
+		}
+		_, err := fmt.Fprintf(cmd.OutOrStdout(), "Acted on: %s %q [ref=%s]%s\n", role, browserUntrustedText(name), ref, suffix)
```
Prasad-D-Ware: `%q` around `browserUntrustedText(...)` mangles the trust boundary.

`browserUntrustedText` returns a three-line block:

```
<<<BEGIN UNTRUSTED EXTERNAL CONTENT>>>
Submit
<<<END UNTRUSTED EXTERNAL CONTENT>>>
```

`%q` escapes those newlines, so the output collapses onto one line with literal `\n`:

```
Acted on: button "<<<BEGIN UNTRUSTED EXTERNAL CONTENT>>>\nSubmit\n<<<END UNTRUSTED EXTERNAL CONTENT>>>" [ref=e3]
```

The markers stop being line-delimited, which is what makes them parseable by a reader or an agent in the first place. Every other caller (`:619`, `:633`, `:642`, `:680`) prints the block with `%s`/`Fprintln` so the newlines survive.

Same issue on `:721` in the ambiguous-candidates loop. Either quote the name and wrap the result, or drop `%q` and let the block stand on its own lines.

backend/internal/cli/browser.go
```diff
@@ -97,6 +97,38 @@ func newBrowserCommand(ctx *commandContext) *cobra.Command {
...
+		Short: "Resolve an element by description and act on it, snapshotting and retrying automatically",
+		Long: "Snapshots, finds the best-matching element for <instruction> against role/name/text\n" +
+			"(no LLM call — deterministic matching, same as reading a snapshot yourself), and performs\n" +
+			"--action on it (default click). Retries once on a stale reference. If the match is\n" +
+			"ambiguous or absent, returns the candidates or snapshot instead of guessing — fall back\n" +
+			"to a manual snapshot/click, retry with --nth, or refine the instruction.",
+		Args: exactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			actArgs := map[string]any{"instruction": args[0], "action": actVerb}
+			if actValue != "" {
```
Prasad-D-Ware: `--value ""` is unreachable: the empty string is dropped here, then the host rejects the call with `INVALID_ARGUMENT: value is required`, even though it explicitly passes `allowEmpty=true` to `stringArg` for exactly this case.

So `act "the search field" --action fill --value ""` can't clear a field. Gating on `cmd.Flags().Changed("value")` instead - the same pattern already used for `nth` two lines down - makes the two ends agree.

Prasad-D-Ware (commented):
Reviewed this end-to-end, including running the real bundled `agent-browser` 0.33.1 binary to capture its actual snapshot output rather than relying on the repo's test fixtures.

The design is right: one-shot `act` instead of manual snapshot->click chaining, deterministic matching with no LLM call, retry-once-then-surface-reality, refusing to guess on ambiguity, all the logic in main with the Go layer staying a passthrough. Those are good calls and the code reads cleanly.

But the PR body's unchecked box - "the actual compact-snapshot text format was inferred from this repo's own test fixtures ... not observed live" - is exactly where the bugs are. Those fixtures (`browser-view-host.test.ts:310`, `cli/browser_test.go:43`) are arbitrary placeholder strings; nothing ever parsed them, so they were never obliged to be accurate. The parser was built around them.

**Two blocking issues:**

1. `SNAPSHOT_LINE` requires the bracket to *begin* with `ref=`, but real output puts `ref` in a shared attribute bracket (`[disabled, ref=e3]`, `[checked=true, ref=e5]`). On a real 11-element snapshot, **7 of 11 elements are silently dropped**, and `check`/`uncheck` - two of the eight advertised v1 verbs - can never resolve a target, since every checkbox/radio/switch emits `checked=`.
2. `candidates.length === 1` short-circuits `MATCH_SCORE_FLOOR`, so a role-hint-only match (score 3) fires unconditionally.

Compounded, on the real snapshot below, `act "the Nope button"` **clicks "Sign In"**. That is a mutating action on an element the user never named - the exact outcome the "don't guess" design is built to prevent.

Ground truth, captured live from the bundled binary (`snapshot -i --compact`):

```
- heading "Hi" [level=1, ref=e1]
- button "Sign In" [ref=e2]
- button "Nope" [disabled, ref=e3]
- textbox "Email" [ref=e4]
- checkbox [checked=true, ref=e5]
- combobox [expanded=false, ref=e6]: A
  - option "A" [selected, ref=e11]
- link "Docs" [ref=e7]
- switch "Toggle" [checked=true, ref=e10]
```

Running that through `matchInstruction` as written:

```
parsed entries        => only e2, e4, e7, e8 survive (7 of 11 dropped)
"the sign in button"  => matched button "Sign In" [ref=e2]   (correct)
"the Nope button"     => matched button "Sign In" [ref=e2]   (WRONG ELEMENT, score 3)
"the terms checkbox"  => no-match
"the Toggle switch"   => no-match
"the option A"        => no-match
```

Worth noting: the same `--json` payload already carries a structured `refs` map -
`{"e2":{"name":"Sign In","role":"button"}, "e5":{"name":"","role":"checkbox"}, ...}` -
and `browser-view-host.ts` already passes it through on the `snapshot` case. Matching
against that instead of regexing the rendered text deletes this entire class of
fragility, including the quoting inconsistency the comment on `SNAPSHOT_LINE` worries
about. That looks like the right fix rather than patching the regex.

Details inline, plus a few smaller ones (Go `%q` mangling the untrusted-content
delimiters, the `errorCode` passthrough's blast radius outside `act`, unreachable
`--value ""`).

Leaving this as a comment rather than a formal request-for-changes, but I'd treat 1 and 2
as merge blockers, and I'd run `act` against a live page once before merging - that single
manual pass would have surfaced both.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. The act path now matches exclusively against the structured refs map and keeps snapshot text only for unresolved output. I also ran the pinned 0.33.1 binary against a live page; the structured map included the disabled button, checked checkbox, and checked switch that the old parser dropped.

frontend/src/main/agent-browser-runtime.ts
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. Runtime passthrough is narrowed to structured STALE_REFERENCE; every other native error remains AGENT_BROWSER_COMMAND_FAILED, preserving tab-resync and close recovery guards. Tests cover both generic and stale codes, and generic messages containing expired no longer trigger an act retry.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. Instructions beginning with an action word are scored both intact and with the leading verb stripped, retaining the best score per ref. The select all and clear all case is covered.

backend/internal/cli/browser.go
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. Matched and ambiguous candidate names now render as separate line-delimited untrusted blocks instead of quoted escaped text. Go tests assert real newlines and reject escaped newline boundaries.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. A lone candidate below MATCH_SCORE_FLOOR now returns no-match; the guard runs before nth, so nth 0 cannot force a junk role-only match. Regression coverage includes score 3, empty names, and one-character names.

backend/internal/cli/browser.go
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. The CLI now gates value forwarding on cmd.Flags().Changed(\value\), so an explicitly supplied empty string reaches fill and type. Covered by a CLI request-payload regression test.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in commit 85b507ccd. Only a trailing ARIA-role token is treated as the role hint, with a regression for the Save Image button selecting button Save Image rather than image Save.

frontend/src/main/agent-browser-runtime.ts
```diff
@@ -784,6 +784,13 @@ function stringError(value: unknown): string {
 	return "";
 }
 
+// Preserve only the stale-reference signal needed by `act`. Other native
+// command failures stay generic so tab drift and close recovery paths continue
+// to recognize AGENT_BROWSER_COMMAND_FAILED.
+function staleReferenceCode(value: unknown): "STALE_REFERENCE" | undefined {
```
Prasad-D-Ware, line 792: **Blocking:** `isRecord(value)` is never satisfied by the real binary, so this always returns `undefined` and the `act` retry can never fire.

agent-browser 0.33.1 emits the error as a plain string:

```
{"success":false,"data":null,"error":"Unknown ref: e1"}
{"success":false,"data":null,"error":"Could not locate element with role=button name=Sign In"}
```

Never `{"error": {"code": ...}}`. Chain: `staleReferenceCode` returns `undefined` -> code becomes `AGENT_BROWSER_COMMAND_FAILED` -> `isStaleReferenceError` (browser-view-host.ts:1674) returns false -> the retry branch is unreachable. The two tests covering it (`browser-view-host.test.ts:513`, `:536`) construct `{ code: "STALE_REFERENCE" }` by hand, a shape the runtime never produces, so they pass while the feature doesn't work.

The allowlist approach is still right — it just needs the message mapping underneath:

```ts
const STALE_REFERENCE_MESSAGE = /^(?:Unknown ref:|Could not locate element with)/;

function staleReferenceCode(value: unknown): "STALE_REFERENCE" | undefined {
    if (isRecord(value) && value.code === "STALE_REFERENCE") return "STALE_REFERENCE";
    return typeof value === "string" && STALE_REFERENCE_MESSAGE.test(value) ? "STALE_REFERENCE" : undefined;
}
```

Worth a test that feeds the real string-shaped envelope end to end, since that's the shape that actually ships. (My suggestion to narrow this to an allowlist is what removed the previous message-based fallback — though for what it's worth, `/stale|no longer|expired/i` wouldn't have matched these messages either, so this path has never worked.)

frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export type ActMatchResult =
+	| { outcome: "matched"; candidate: ActCandidate }
+	| { outcome: "ambiguous"; candidates: ActCandidate[] }
+	| { outcome: "no-match" };
+
+export function parseSnapshotEntries(snapshotRefs: unknown): SnapshotEntry[] {
+	const entries: SnapshotEntry[] = [];
+	if (!snapshotRefs || typeof snapshotRefs !== "object" || Array.isArray(snapshotRefs)) return entries;
+	for (const [ref, value] of Object.entries(snapshotRefs)) {
```
Prasad-D-Ware, line 28: `--nth` now indexes in lexicographic ref order rather than document order — a regression from the (otherwise correct) switch away from parsing the snapshot text.

The binary emits refs map keys sorted as strings:

```
refs key order from binary: e1 e10 e11 e2 e3
```

With two "Add to Cart" buttons at `e2` and `e11`, `Object.entries` preserves that order, so:

```
candidate order: e11 e2 e1 e10 e3
--nth 0 => e11   <- the SECOND one on the page
--nth 1 => e2    <- the first
```

That inverts `browser.md`'s motivating example ("the second Add to Cart button") on any page with 10 or more refs. Text parsing gave document order for free; the map doesn't.

Sorting by the numeric suffix restores it:

```ts
entries.sort((a, b) => Number(a.ref.slice(1)) - Number(b.ref.slice(1)));
```

frontend/src/main/browser-act-matcher.ts
```diff
@@ -0,0 +1,123 @@
...
+	const candidates = [...candidatesByRef.values()].sort((a, b) => b.score - a.score);
+	if (candidates.length === 0) return { outcome: "no-match" };
+
+	const [top, second] = candidates;
+	const isConfidentMatch =
+		(top.score >= EXACT_NAME_TIER && (!second || second.score < EXACT_NAME_TIER)) ||
+		(top.score >= MATCH_SCORE_FLOOR && (!second || top.score - second.score >= MATCH_SCORE_GAP));
+
+	if (isConfidentMatch) return { outcome: "matched", candidate: top };
+	if (candidates.length === 1 && top.score < MATCH_SCORE_FLOOR) return { outcome: "no-match" };
```
Prasad-D-Ware, line 122: This returns before the `nth` check below, which makes controls with an empty accessible name unreachable by `act` — including via the documented `--nth` escape hatch.

On a real page, `checkbox`, `radio`, and unlabelled `combobox` elements come back with `name: ""`:

```json
"e5": {"name": "", "role": "checkbox"}, "e9": {"name": "", "role": "radio"}
```

The only way to address those is by role, but a role-only instruction leaves `nameHint` empty, so `nameScore` returns 0 and the candidate scores 3 — one candidate, below `MATCH_SCORE_FLOOR`:

```
refs: {e1: checkbox "", e2: button "Save"}
act "the checkbox"            => no-match
act "check the checkbox"      => no-match
act "the checkbox" --nth 0    => no-match   (never reaches the nth branch)
```

So `check`/`uncheck` — whose targets most often have no accessible name — can't reach an unnamed control at all, and unlike the ambiguous case there's no way for the caller to force it.

Failing safe here is the right instinct, so I wouldn't lower the floor. But moving the `nth` check above this line would give the caller the same deterministic override the ambiguous path already has, which seems like the point of the flag.

Prasad-D-Ware (commented):
Re-reviewed at `85b507c`, again running the real bundled `agent-browser` 0.33.1 binary rather than reading.

**All seven items from the last round are fixed, and fixed properly.** Switching the parser to the structured `refs` map instead of patching the regex was the right call. Verified against the live `refs` payload I captured earlier:

```
entries parsed        => 11 of 11   (was 4 of 11)
"the Nope button"     => matched button "Nope" [e3]        (was: clicked "Sign In")
"the cancel button"   => ambiguous                          (was: clicked "Confirm delete")
"the Toggle switch"   => matched switch "Toggle" [e10]      (was: no-match)
"select all"          => matched button "Select all" [e1]   (was: permanently ambiguous)
"the Save Image button" => correct role resolution
```

`browser-act-matcher.test.ts` and `agent-browser-runtime.test.ts` pass locally (34 passed, 1 skipped). `browser-view-host.test.ts` couldn't run in my sandbox — Electron's binary isn't downloadable there — so that suite is unverified on my side, not failing.

**One blocker left: the stale-reference retry is dead code against the real binary.**

I probed it three ways:

```
$ agent-browser click @e1 --json     # after removing the element
{"success":false,"data":null,"error":"Could not locate element with role=button name=Sign In"}

$ agent-browser click @e99 --json
{"success":false,"data":null,"error":"Unknown ref: e99"}

$ agent-browser click @e1 --json     # after navigating away
{"success":false,"data":null,"error":"Unknown ref: e1"}
```

`error` is always a plain string, never an object with `.code`. So `staleReferenceCode`'s `isRecord` guard is never satisfied, the code always falls back to `AGENT_BROWSER_COMMAND_FAILED`, `isStaleReferenceError` is never true, and the retry this PR's title advertises never fires.

To be clear about my own part in this: narrowing the passthrough to an allowlist was my suggestion, and I flagged the blast radius without checking whether the codes were real in the first place. The allowlist is still right — it just needs a message-to-code mapping underneath it. Also worth noting the previous revision's `/stale|no longer|expired/i` fallback wouldn't have matched these messages either, so this path has never actually worked; and the `STALE_REFERENCE` convention predates this PR (`browser-runtime-link.test.ts:86`, `browser-view-host.test.ts:328`) — this is just the first code to depend on it.

Two smaller items introduced by the refs-map switch are inline (`--nth` ordering, unnamed controls).

The pattern I'd gently flag: round one assumed the snapshot format from fixtures, round two assumed the error format from fixtures. Everything found in both rounds came out of about ten minutes with the actual binary. The unchecked "manual verification in the real desktop app" box is still the highest-value thing left — I'd rather see that box ticked than a third round of inference.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in b3173de54. Explicit nth can now select an unnamed control from a role-only instruction without reopening the prior unsafe named low-confidence override. Action-prefixed role-only parses are scoped to that explicit role, so check the checkbox cannot select a button named Check. Matcher and host regressions cover both no-nth and nth behavior.

frontend/src/main/browser-act-matcher.ts
Vaibhaav-Tiwari: Fixed in b3173de54. Structured eN refs are sorted by their numeric suffix before scoring, restoring document order when the native map is lexicographically ordered. Regression coverage uses e1, e10, e11, e2, e3 and asserts numeric order.

frontend/src/main/agent-browser-runtime.ts
Vaibhaav-Tiwari: Fixed in b3173de54. The runtime now maps the two real agent-browser 0.33.1 string prefixes, Unknown ref: and Could not locate element with, to STALE_REFERENCE while leaving unrelated string and structured failures generic. The host retry regression now feeds the real string-shaped envelope through parseAgentBrowserJSON end to end.

Prasad-D-Ware (approved):

PR #4248, opened 2026-08-21
Feat/chat composer redesign

Annieeeee11:
## Summary
- Redesigns the chat composer chrome: taller prompt, circular attach/send controls, muted borders, and a dock that blends into the chat surface.
- Empty conversations center the prompt and animate it into the bottom dock on the first turn; queue/steer sits inline before Approvals, with Compact and jump-to-bottom sized into the same control row.
- Clubs ACP turn settings (model + effort left, mode right) into the Codex two-trigger layout so Droid/Claude/OpenCode/Kimchi match Codex without changing provider option ids/values.
- Tidies timeline affordances: rollback beside copy, labeled edit/copy actions, quieter focus rings, and scrollable long model lists in the turn-settings submenu.

## Test plan
- [ ] Empty session: composer is centered; after first send it docks with the FLIP motion
- [ ] Attach, send, stop, queue, and steer still work; Enter hint matches the armed action
- [ ] Codex: model/effort + Approvals two-trigger chrome looks correct
- [ ] ACP agents (Claude/Droid/OpenCode/etc.): same two-trigger layout; option values unchanged
- [ ] Compact sits next to Approvals; jump-to-bottom is usable on long threads
- [ ] User message edit / copy / rollback labels and placement
- [ ] Long model list scrolls inside the turn-settings submenu
- [ ] Light and dark theme: composer background, borders, and send armed/disabled states


ronishrohan (approved):
Reviewed at `943b08b10`. All 10 CI checks green, including the `test` job that was failing on `eaa1b878a`.

Disclosure: much of the recent work on this branch (the last five commits) was written in a Claude Code session driven by me, so this approval is not an independent second pair of eyes on those commits — treat it as the branch owner signing off, not as external review.

What was checked, and what changed as a result:

**Root-cause fix for the model picker.** Selecting a model made the whole turn-settings picker vanish until reload. Cause was in `acp/config_options.go`: some agents accept `session/set_config_option` without echoing the rebuilt catalog, and `replaceConfigOptions` took that empty list literally and cleared every option for the session. First attempt at a fix retained the old catalog instead, which traded the bug for a subtler one — `SetConfigOption` then returned the pre-change value, so the picker silently reverted a change the agent had actually applied. Final shape splits the two: `replaceConfigOptions` stays authoritative (an empty catalog from the `session/update` notification applies verbatim, per that channel's documented "complete replacement, including removing an option" contract), while a set-response carrying no catalog records just the accepted value against the catalog we hold.

**Steer paths.** Clicking send while Cmd was held showed "Steer" and queued anyway; it now reads the same armed state the chip paints. Steering is text-only, so a draft carrying attachments no longer arms it — previously Cmd+Enter with attachments-only returned silently, and with text+attachments it steered the text and stranded the files.

**Also.** `disabled` now reaches the multi-option `ClubbedConfigPicker` trigger; dropped a `preventDefault` that Radix reads as a signal to skip its own open/close handling; removed dead `DropdownMenuSub*` exports and the orphaned `onPromoteQueuedTurn` prop; delivery chips moved off `role="status"` so they stop shadowing the steer-refusal live region.

**Test infrastructure.** `ChatSteer.test.tsx` was not failing locally — it was never loading, because the renderer aliases `@aoagents/product-ui` to that package's source and its `clsx`/`tailwind-merge` imports resolved from `packages/product-ui`, which only has a `node_modules` if `npm ci` was run there too. That silent transform-time failure is why the `role="status"` regression reached CI instead of being caught locally. Both deps now resolve from the frontend copies, so one install is enough. Added seven composer tests covering the Cmd/Ctrl+Enter steer path, which had none.

Known follow-up, not addressed here: `src/landing` tests need their own `npm ci` for `cheerio`, so a frontend-only install still can't run them locally. Same footgun shape, unrelated to this PR.

PR #4239, opened 2026-08-21
fix: add review run CDC

neversettle17-101:
## Summary
- Add review-run CDC events and wire them through session read-model snapshots, HTTP SSE, and the desktop event transport.
- Build a session review snapshot that classifies AO-owned vs external review facts, but keeps only the counts Kanban needs rather than full review/comment payloads.
- Rebuild the sqlite change_log migration using a rename/view compatibility pattern so older schemas keep their existing CDC triggers working.
- Update tests for backend CDC, session review snapshot assembly, and frontend event listeners.

## Call taken in this PR
- Keep the frontend on the existing shared workspace refresh path for review-run CDC events instead of introducing targeted per-session fetches in this change.
- Do not add batching beyond the existing renderer-side debounce.
- Rely on the current `event-transport.ts` 150ms debounce to collapse bursty `review_run_created` / `review_run_updated` events into a single refresh burst.
- Defer any broader frontend cache redesign or event-specific batching/targeted invalidation to a follow-up.

```mermaid
  flowchart LR
      subgraph Backend
          A[Review service / store]
          B[(review_run table)]
          C[(change_log table)]
          D[CDC poller + broadcaster]
          E[HTTP events SSE]
          F[Session service]
          G[ReviewSnapshot builder]
      end

      subgraph Frontend
          H[Event transport / SSE listener]
          I[Workspace query invalidation\n150ms debounced]
          J[GET projects + sessions\nplus related summaries]
          K[UI / Kanban]
      end

      A --> B
      B -->|SQLite trigger| C
      C --> D
      D --> E
      E --> H
      H --> I
      I --> J
      J --> F
      F --> G
      G --> F
      F --> K
```

## Tests
- go test ./internal/storage/sqlite ./internal/storage/sqlite/store ./internal/service/session ./internal/httpd ./internal/cdc
- npm test -- --run src/renderer/lib/event-transport.test.ts

## Issue
- Closes https://github.com/Untrivial-ai/agent-orchestrator/issues/4226

Issue #4179, opened 2026-08-20
Implement deriving display status for each agent session


neversettle17-101:
# AO Kanban Status Derivation

This document describes how AO should derive the Kanban presentation for a session from durable facts. The board should stay lifecycle-oriented and mostly unidirectional, while the display status explains what is happening right now.

The proposed active board has four columns:

```text
Building -> Validating -> Needs Review -> Ready
```

Terminated sessions should move out of the active board into a separate archive section.

## Implementation Model

Derive Kanban presentation in two stages:

1. **Derive the Kanban column** from stable lifecycle and ownership facts: session termination, PR existence, PR state, whether auto review is on, whether an internal review/fix/test loop is already active, and whether the PR is approved, mergeable, or finished.
2. **Derive the display status** from the most important current fact inside that column: worker activity, hook freshness, CI, changes requested, review runs, review verdicts, approvals, mergeability, and terminal PR state.

This keeps the column stable and user-facing, while allowing the display status to carry detailed state.

| Stage | Answers | Main inputs | Output |
|---|---|---|---|
| Column derivation | Where is this session in the lifecycle, and who owns the next step? | Session termination, PR existence, PR state, auto-review setting, active internal review/fix/test loop, human approval, mergeability | `Building`, `Validating`, `Needs Review`, `Ready`, or archive section |
| Display status derivation | What is happening or needed right now? | Worker activity, hook freshness, CI status, changes requested, review run status, review verdict, human approval, mergeability | `Working`, `Blocked`, `Reviewing`, `Fixing CI failures`, `Needs human review`, `Mergeable`, etc. |

## Kanban Columns

| Column | Meaning | Typical display statuses |
|---|---|---|
| Building | AO/agent is creating the change before a PR exists. | `Working`, `Awaiting PR`, `Blocked`, `No signal`, `Exited` |
| Validating | AO is checking, reviewing, testing, or iterating on the PR. This column should be driven by PR existence and AO-owned facts: draft PRs, AO review runs, AO reviewer comments/verdicts, CI fixing, and future AO testing/artifact validation. | `Draft`, `Review scheduled`, `Reviewing`, `Fixing CI failures`, `Checks failing`, `Addressing comments`, `Blocked`, `Exited`, `Review pending`, `Needs review` |
[...]

PR #4249, opened 2026-08-22
feat(board): derive a Kanban column & display status per session

### LINKED ISSUE #4178: Implement sorting sessions into Kanban Columns
# Kanban Column Derivation

## Goal

Derive a stable Kanban placement for every worker session from durable facts. The column answers where the session is in its delivery lifecycle and who owns the next step. It is independent from the existing display `SessionStatus`.

```text
Building -> Validating -> Needs Review -> Ready
                                     \
                                      Archive (terminated only)
```

## Scope

This issue adds only the derived Kanban column and board grouping. It does not redesign display statuses, pills, or card content.

Review-run CDC is assumed to be available. It causes the normal workspace refresh; the backend recomputes the column from durable facts on every session-list read.

## Current Implementation

Today:
1. Backend derives one `SessionStatus`.
2. Frontend `attentionZone()` maps that status to the current `working`, `action`, `pending`, and `merge` board lanes.
3. The board separately archives terminated sessions.

The new column must be backend-derived rather than added to the frontend mapping: review-run state, PR-head freshness, and review ownership are durable-fact rules, not UI rules.

## Proposed Contract

Add a derived `kanbanColumn` to each session response:

```text
building | validating | needs_review | ready | archive
```

Do not persist this value. Derive it during `GET /sessions`, alongside the existing derived status.

## Inputs

Build a narrow internal read model from:
- Session facts: worker kind, termination, auto-review and auto-inject settings.
- PR facts: lifecycle, draft/open/merged/closed, current `headSHA`, CI, approvals, and mergeability.
- AO review-run facts: PR URL, target SHA, status, verdict, and GitHub review ID.
- Source-aware SCM review facts: external/human approvals and requested changes, separate from AO-authored reviews/comments.

A review run is current only when `reviewRun.targetSHA == pr.headSHA`.

AO-authored provider reviews must be identified through `reviewRun.githubReviewID`. Do not use aggregate `PRFacts.Review` alone to choose between Validating and Needs Review, because it combines AO and external review signals.

## Column Derivation

First, derive a placement for each PR:

```text
session terminated?
  yes -> Archive

no PR exists?
  yes -> Building

PR is draft?
  yes -> Validating

PR is merged or closed without merge?
  yes -> Ready

PR is approved or mergeable?
  yes -> Ready

AO-owned review/fix/test loop active?
  yes -> Validating

auto review enabled?
  yes -> Validating

otherwise -> Needs Review
```

An AO-owned loop includes current-head AO review work, AO CI fixing, and AO addressing AO-originated review feedback.

Once a PR is in the review lifecycle, worker activity must not move it back to Building. A new head SHA is a new validation cycle:
- Auto review on: Validating.
- Auto review off: Needs Review.

## Multiple PRs

Derive placement per PR, then select the session’s representative PR.

1. Exclude merged and closed PRs while at least one live PR exists.
2. Among live PRs, choose the highest-priority actionable placement:
   1. Ready (mergeable/approved)
   2. Needs Review
   3. Validating
   4. Building
3. Use merged/closed states only when no live PR remains.
4. Add a deterministic tie-breaker, initially latest PR update time.

This ensures a terminal PR does not hide a live PR that needs validation or human review.

## Implementation Steps

1. Define the backend `KanbanColumn` enum and add derived `kanbanColumn` to the session API DTO/OpenAPI contract.
2. Add a pure `deriveKanbanColumn(facts)` reducer next to, but separate from, existing `deriveStatus`.
3. Build the reducer input from session, PR, review-run, and source-aware SCM review facts.
4. Add/read the required review data in the session-list path with bounded queries; do not issue per-card unbounded lookups.
5. Generate API types after changing the DTO.
6. Change `SessionsBoard` to group active workers by `session.kanbanColumn`.
7. Render `archive` through the existing archive section.
8. Leave existing display-status labels and card detail untouched.

## Acceptance Criteria

- Every worker session returns one derived `kanbanColumn`.
- Terminated sessions always derive to `archive`.
- No-PR worker sessions derive to `building`.
- Draft and current-head AO validation work derive to `validating`.
- Auto-review-off review-ready PRs derive to `needs_review`.
- Approved/mergeable PRs derive to `ready`.
- A stale review run for an old SHA cannot determine the current column.
- AO and external review signals lead to the appropriate ownership column.
- Multiple PRs select the highest-priority actionable live PR, with deterministic ties.
- A review-run CDC update reaches the existing workspace refresh path and the refreshed session returns its newly derived column.
- Unit, service/API, and board-grouping tests cover the rules above.

neversettle17-101:
## Summary
- Adds a daemon-derived `kanbanColumn` (`building | validating | needs_review | ready | archive`) to the session read model. It says where a session sits in its delivery lifecycle and which loop is turning it, derived at read time from durable facts alongside the existing status, and never persisted.
- The board groups by that column instead of re-deriving a lane from display status, and its lanes now carry their own labels and colours: **Building · Validating · In review · Ready**, left to right, with `archive` rendering in the existing archive sheet.
- Status pills and card content are untouched, per the issue's scope.

Rebased onto `main` after #4239 merged.

## Derivation

`contract.DeriveKanbanColumn` is a pure reducer sitting next to — not inside — `DeriveStatus`:

```
terminated                                     -> archive
no PR                                          -> building
draft                                          -> validating
merged / closed                                -> ready
human-approved, or mergeable                   -> ready
AO is turning the review-feedback loop         -> validating
auto review on and this head is unreviewed     -> validating
otherwise                                      -> needs_review
```

With several PRs, live ones win over merged/closed, then the most actionable placement wins (ready > needs review > validating > building), breaking ties on the newest PR and finally on URL so the board never flickers.

### The two middle lanes are one loop

`validating` and `needs_review` are the same review-feedback loop, split by whose turn it is. AO is turning it when a review pass is running against the current head, when it is addressing review feedback (`autoInjectReview`), or when it is fixing failing CI (`autoInjectCI`). `needs_review` is the **fallthrough**: no AO loop claimed the PR, so the next turn is a person's — giving the review, answering feedback already on it, or deciding what to do about a failing check.

That is why flipping an inject flag moves a card between the two lanes with nothing on the PR changing, and why "In review" describes the whole lane rather than only PRs awaiting a first review.

## Ownership signals

Two rules from the issue shaped the data model:

- **Review-run currency is enforced in SQL.** `ListCurrentHeadReviewRunsBySession` joins `review_run` to `pr` on `target_sha = pr.head_sha`, so a pass recorded for an earlier commit never reaches the reducer. A new head SHA starts a new cycle.
- **AO's own provider reviews are excluded from the human verdicts.** `ListPRFactsBySession` gained `external_approved` / `external_changes_requested`, computed from each human reviewer's *latest* decisive review, minus bots and minus any review id AO recorded on a run. The aggregate `review_decision` mixes both sources, so Ready requires the provider decision **and** a surviving non-AO approval.

Both ride the existing per-session reads in the session-list path; the review-run query is skipped for terminated sessions and sessions with no PR.

## Board changes

The board previously borrowed the attention-zone vocabulary for its lanes, so a column derived as `building` rendered under a header reading "Working". Columns now own their label and palette (new `column.*` keys in all eight locales), and `data-column` plus the lane aria label expose the column name.

The **idle-vs-working and ready-vs-merged split lanes are gone**. They sub-grouped a column by display status and replaced the column's own header, which is what kept "Working" and "Ready to merge" on screen. Each lane is now one flat list ordered newest first — the order the merge lane already used. Their labels and i18n keys are retired with them.

`attentionZone` itself stays: the command palette and the tray rank sessions by status urgency, which is a different question from delivery placement.

Two behaviour notes:

- A worker with **no PR** stays in Building whatever its runtime status — an exited agent no longer jumps to "Needs you" on the board. Its card still carries the Exited badge. That follows directly from the issue's `no PR exists? -> Building` rule.
- Cards in every lane are now ordered newest first. Previously only the merge lane sorted.

## Display status: what changed, and what did not

The display `SessionStatus` is **unchanged**: `backend/pkg/contract/status.go` and `internal/service/session/status.go` are not in this diff, the enum keeps all 15 values, `deriveStatus` and `DeriveSCMStatus` are untouched, and the 15 `status.*` i18n keys are byte-identical. Card pills render exactly as before.

What changed is what the board *does* with status:

- **Status no longer places a card.** Lane placement comes from `kanbanColumn`; `attentionZone(session.status)` is gone from the grid. That is the point of the PR.
- **The board's status sub-grouping is gone.** The idle-vs-working and ready-vs-merged split lanes sorted a column's cards by display status and put those names in the column header. With them removed, the board no longer surfaces those distinctions — the card's own pill is now the only place idle vs working, or ready vs merged, appears. This retires 13 `shell.*` label keys per locale (`countLabel.*`, `idleSessions`, `workingSessions`, `mergedSessions`, `readyToMergeSessions`, `idleWorkingSessions`, `readyMergedSessions`, `laneSummaryAria`, `countSessionsAria_*`).
- **`zone.*` keys stay.** All five are still used — the tray and the command palette rank sessions by status urgency, which is a different question from delivery placement, and `attentionZone` still answers it for them.

### Lane and pill can now disagree, by design

Previously a card's lane was a function of its status, so the two always agreed. They are now independent readings, and the seeded scenarios include cases where they differ on purpose:

| Card shows | Sits in | Why |
| --- | --- | --- |
| `Mergeable` | In review | Provider decision is `approved`, but the only approval is AO's own, so no person has signed off. |
| `Exited` | Building | The agent exited before opening a PR; the delivery lifecycle has not started. |
| `Merged` | Ready | Terminal PR, live session — Ready holds merged work until the session itself is terminated. |

The pill answers "what is this session's state right now"; the lane answers "where is this work, and which loop is turning it". Keeping both is deliberate — the issue scoped pills and card content as out of bounds, and nothing here narrows them.

## Review fixes in this PR

- **Superseded review verdicts.** `external_changes_requested` originally tested raw existence, so a reviewer who requested changes and later approved left both rows behind and the stale request pinned the session to Validating. The aggregate decision does not rescue this — it reads `approved` only once every required approval has landed, so a PR short of its review count stayed stuck. Both flags now come from a CTE holding each human reviewer's most recent decisive review; AO's own reviews are excluded from the superseding side too, since AO posts under an account indistinguishable from a person by author name.
- **Older-daemon fallback.** `toKanbanColumn` dropped every session without a `kanbanColumn` into Building, which moved ready and in-review cards leftward during a mixed-version upgrade — reading as work regressing rather than as a client that cannot see the field. Such a daemon still sends `status`, and mapping it through `attentionZone` reproduces exactly the lane the board used before the column existed, so the fallback preserves placement. It covers an unrecognized column too.
- **`sqlc` truncation.** An em dash in a new SQL comment tripped the parser bug already documented in `queries/sessions.sql` and `changelog.sql`: the generated SQL was silently truncated by the extra byte count, cutting `ORDER BY pr.updated_at DESC` to `DE`. It compiled and failed at runtime. The query body is ASCII-only now, with a note recording why.

## Tests

`go test ./pkg/contract ./internal/service/session ./internal/storage/sqlite/... ./internal/httpd/... ./internal/domain`, `golangci-lint run` (0 issues), `gofmt` clean, `npm run typecheck`, 1920 renderer tests, 80 product-ui tests.

New coverage:
- Reducer table tests for every rule plus multi-PR selection and tie-breaking.
- Real-SQLite tests for stale-SHA filtering, the AO/external split, both supersede directions, two disagreeing reviewers, and a later AO review not dismissing a human verdict.
- Service tests for the derived column; an API test asserting `kanbanColumn` on the wire.
- Board tests that lanes group by column (not status), render in delivery order, and that an older daemon's sessions keep their status-implied lanes.

### Manual pass

Sixteen worker sessions, one per derivation rule, seeded as durable facts into a scratch database and read back through the real daemon — so the SQL, the reducer, and the board were exercised end to end rather than against fixtures. All 16 derived their expected column.

Three of them discriminate against specific wrong implementations, checked at the SQL layer too:

| Scenario | Provider decision | Review rows | `external_approved` | `external_changes_requested` |
| --- | --- | --- | --- | --- |
| only AO approved | `approved` | 1 | **0** | 0 |
| changes requested, then approved | `review_required` | 2 | **1** | **0** |

The first is the AO/external split: the provider says `approved`, but the only approval is AO's own. The second is the supersede fix — before it, that last cell read `1`.

## Not in this PR

`ExternalCommented` — a human COMMENTED review — is not persisted anywhere today: the GitHub query selects only `APPROVED`/`CHANGES_REQUESTED`, `pr_reviews.state` is CHECK-constrained to those, and `pr_comment` carries no review id to attribute comments to AO vs a person. Plumbing it needs a provider change, a migration, and a `ReviewDecision` enum widening, so it is filed as #4250.

## Issue
- Closes https://github.com/Untrivial-ai/agent-orchestrator/issues/4178

https://claude.ai/code/session_01S5s7m6yHp5YU5BEP9wV4Xq


neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Adds a daemon-derived kanbanColumn to sessions, derives it from PR and review facts in the backend, exposes it on the sessions API, and rewires the board UI to group cards by that server-provided lane instead of re-deriving lanes from display status.

## Summary of Changes
- Added a shared Kanban column contract and backend reducer for building, validating, needs_review, ready, and archive.
- Extended session reads, API/session DTOs, frontend workspace mapping, and board rendering to carry kanbanColumn end to end.
- Added store queries for current-head AO review runs and external human review verdicts, plus reducer, UI, and smoke-test coverage.

## Findings
- **Medium**: Stale human changes requests keep auto-inject-review sessions stuck in validating. `external_changes_requested` is derived as any non-bot changes_requested review, and `aoOwnsNextStep` treats that as active feedback. If a reviewer later approves the PR, both review rows still exist in `pr_reviews`, so the stale changes request continues to force `validating` even though the provider aggregate decision is now approved.

## Regression Risk
The new lane reducer changes board ownership semantics across backend, API, and UI, so the biggest risk is misclassifying review state when review history contains multiple verdicts from the same human reviewer.

## Testing
Added reducer/store/backend/frontend tests and updated board smoke tests. I did not run tests. One gap is coverage for the changes_requested -> approved transition from the same human reviewer, which is the case above.

backend/internal/storage/sqlite/queries/pr.sql
```diff
@@ -230,7 +233,33 @@ SELECT
...
+          AND pr_reviews.state = 'approved'
+          AND pr_reviews.is_bot = 0
+          AND NOT EXISTS (
+              SELECT 1
+              FROM review_run
+              WHERE review_run.github_review_id != ''
+                AND review_run.github_review_id = pr_reviews.review_id
+          )
+    ) AS external_approved,
+    EXISTS (
```
neversettle17-101, line 314: This computes `external_changes_requested` as existence of any historical non-bot changes request that AO did not author. When the same reviewer later approves, both rows remain in `pr_reviews`, so this flag stays true and `aoOwnsNextStep` keeps auto-inject-review sessions in `validating` even though the aggregate provider decision is now approved. Please derive this from the latest decisive human review per reviewer, or otherwise gate it with the current aggregate review state, instead of raw existence.

neversettle17-101:
Addressed the changes-requested review (`4999815030`) in 911b374c4.

**The finding was right, and my original reasoning was the flaw.** I'd considered per-reviewer latest-verdict logic while writing the query and dropped it, on the argument that a reviewer who requests changes and then approves would flip `pr.review_decision` to `approved`, and `externallyApproved` is checked before `aoOwnsNextStep`, so the stale row could never decide the column.

That holds only when the approval is *sufficient*. GitHub reports `APPROVED` just once every required approval has landed — so with required reviewers or CODEOWNERS, a PR where reviewer A approved but reviewer B hasn't yet reads `REVIEW_REQUIRED`. `externallyApproved` is then false, execution reaches `aoOwnsNextStep`, and the superseded changes request pins the session to `validating` while nothing is actually waiting on the worker.

**Fix:** both external flags now come from a CTE holding each human reviewer's most recent decisive review, so a verdict counts only until that reviewer replaces it. AO's own reviews are excluded from the superseding side as well — AO posts under an account that is indistinguishable from a person by author name, and its approval must not dismiss a human's changes request.

**Tests** covering the gap you named:
- `TestListPRFactsForSessionUsesEachReviewerLatestVerdict` — changes_requested → approved, approved → changes_requested, and two reviewers disagreeing (each keeps their own verdict). Deliberately written with `review_decision = review_required` so the aggregate cannot mask the bug.
- `TestListPRFactsForSessionKeepsHumanVerdictOverLaterAOReview` — a later AO-authored approval does not dismiss a human changes request.

**Unrelated find while fixing this:** an em dash in the new SQL comment triggered the same class of sqlc 1.31 SQLite-parser bug already documented in `queries/sessions.sql` and `queries/changelog.sql` — the generated SQL was truncated by the extra byte count, cutting `ORDER BY pr.updated_at DESC` to `DE`. Every query using it failed at runtime, though it compiled fine. The query body is ASCII-only now, with a note recording why.

Verified: `go test ./pkg/... ./internal/service/... ./internal/storage/... ./internal/httpd/... ./internal/domain/...`, `golangci-lint run` (0 issues), `gofmt` clean.

backend/internal/storage/sqlite/queries/pr.sql
neversettle17-101: Fixed in 911b374c4 — you were right, and my original reasoning was the flaw.

I had considered per-reviewer latest-verdict logic and dropped it, arguing that a reviewer who requests changes then approves flips `pr.review_decision` to `approved`, and `externallyApproved` is tested before `aoOwnsNextStep`, so the stale row could never decide the column. That only holds when the approval is *sufficient*: GitHub reports `APPROVED` only once every required approval has landed, so with required reviewers or CODEOWNERS a PR reads `REVIEW_REQUIRED` while one reviewer has approved and another has not — `externallyApproved` is false, execution reaches `aoOwnsNextStep`, and the superseded changes request pins the session to `validating`.

Both flags now come from an `external_review` CTE holding each human reviewer's most recent decisive review, so a verdict counts only until that reviewer replaces it. AO's own reviews are excluded from the superseding side too, since AO posts under an account indistinguishable from a person by author name and its approval must not dismiss a human's changes request.

Covered by `TestListPRFactsForSessionUsesEachReviewerLatestVerdict` (both transition directions plus two disagreeing reviewers, written with `review_decision = review_required` so the aggregate cannot mask the bug) and `TestListPRFactsForSessionKeepsHumanVerdictOverLaterAOReview`.

packages/product-ui/src/session-models.ts
```diff
@@ -43,6 +58,16 @@ export function toSessionStatus(status?: string, isTerminated = false): SessionS
 	return isTerminated ? "terminated" : "unknown";
 }
 
+/**
+ * A daemon too old to send a column, or one sending an unknown value, must still
+ * land somewhere sensible: terminated sessions archive, everything else is
+ * building.
+ */
+export function toKanbanColumn(column?: string, isTerminated = false): KanbanColumn {
```
neversettle17-101: This fallback breaks compatibility with older daemons in a visible way: if `kanbanColumn` is missing, every non-terminated session now lands in `building`, so sessions that were previously ready / in review / needing attention all collapse into the leftmost lane during a mixed-version upgrade. Since `frontend/src/renderer/types/workspace.ts` explicitly says we still support an older daemon that omits this field, the fallback should preserve the preexisting placement (for example by deriving from `status`) instead of defaulting all live sessions to `building`.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Adds a daemon-derived `kanbanColumn` to session reads, derives it from PR/review facts in the backend, exposes it on the sessions API, and updates the board UI to render lanes from that server-provided column instead of re-deriving them from display status.

## Summary of Changes
- Added a shared Kanban column contract plus backend reducers, store queries, and API/schema updates to carry `kanbanColumn` end to end.
- Split AO-authored vs external review verdicts in SQLite so Kanban ownership can use current-head AO runs and current human review state.
- Reworked the board UI and tests to group by daemon-provided columns and updated smoke coverage accordingly.

## Findings
- **Medium**: The compatibility fallback for daemons that do not send `kanbanColumn` now maps every live session to `building`, which misplaces ready / validating / needs-review cards during mixed-version upgrades.

## Regression Risk
This change redefines board placement across backend, API, and UI, so the main risk is classification drift when a newer renderer talks to an older daemon or receives an unexpected `kanbanColumn` value.

## Testing
The PR adds contract, store, service, controller, renderer, and smoke tests. I did not run tests. I did not see coverage for the mixed-version case the new fallback explicitly mentions (older daemon omits `kanbanColumn`).

packages/product-ui/src/session-models.ts
neversettle17-101: Fixed in 2c6727dab.

You caught an internal contradiction: `workspace.ts` claims support for a daemon too old to send the field, while the fallback quietly discarded the placement such a daemon still describes. Under that pairing a ready or needs-review card would slide into the leftmost lane, which reads as work regressing rather than as a client that cannot see the field.

An older daemon does still send `status`, and mapping it through `attentionZone` reproduces exactly the lane the board used before this column existed, so the fallback now preserves placement rather than inventing one. It covers the unrecognized-column case too, which is the same problem from the other direction (newer daemon, lane this client does not know).

`toKanbanColumn` moved from `session-models.ts` to `session-presentation.ts` so it can reach `attentionZone`, and its signature is now `(column, status)`.

Coverage for the case you flagged as missing:
- `session-presentation.test.ts` — every status mapped to its legacy lane, asserted for `undefined`, `""`, and an unrecognized column, plus a case proving an explicit column always wins over anything status implies.
- `SessionsBoard.test.tsx` — "keeps an older daemon's sessions in their status-implied lanes" renders four sessions with `kanbanColumn` deleted from the payload and asserts each lands in its own lane.

One related fix your comment surfaced indirectly: the board fixtures had a local copy of this mapping, so they were testing a duplicate rather than the shipped function. They now call `toKanbanColumn` directly. That immediately exposed a stale test whose inline fixture omitted the column and was silently relying on the fallback while claiming to assert daemon-derived placement; it now sets `kanbanColumn: "building"` explicitly.

neversettle17-101 (commented):
Approving.

## What This PR Does
Adds a daemon-derived `kanbanColumn` to sessions, derives it from backend PR/review facts, exposes it on the sessions API, and updates the board UI to render lanes from that server-provided column instead of re-deriving them from display status.

## Summary of Changes
- Added a shared Kanban column contract plus backend reducer/store/API wiring to compute and return `kanbanColumn`.
- Split AO-authored versus external human review verdicts so Kanban ownership can use current-head AO review runs and current human review state.
- Reworked the board UI, presentation helpers, and smoke/unit tests to group by Kanban columns while preserving lane placement for older daemons that do not send the new field.

## Findings
- No clear regression found from the reviewed diff. The previous mixed-version fallback issue is addressed by deriving legacy placement from `status`, and the SQL now reduces external review state to each reviewer's latest decisive human verdict.

## Regression Risk
Main risk remains classification drift between daemon-derived columns and legacy status-based fallback during upgrades, but the reviewed changes now explicitly cover that compatibility path.

## Testing
Added/updated contract, store, service, controller, renderer, and smoke tests, including coverage for stale human review verdicts and older daemons that omit `kanbanColumn`. I did not run tests.

packages/product-ui/src/session-presentation.ts
```diff
@@ -43,6 +46,11 @@ const englishLabels: Record<SessionPresentationMessageKey, string> = {
 	"zone.pending": "In review",
 	"zone.working": "Working",
 	"zone.done": "Terminated",
+	"column.building": "Building",
+	"column.validating": "Validating",
+	"column.needs_review": "In review",
```
neversettle17-101, line 61: `needs_review` is still the lane for statuses where a person owns the next step, not just PRs that are actively being reviewed. The fallback tests below still map `changes_requested`, `needs_input`, and `ci_failed` into this same column, so renaming the user-facing label to "In review" misdescribes those cards. Keeping a label closer to ownership (like "Needs review" / "Needs you") avoids telling the user that a worker blocked on changes or input is currently under review.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Keeps the kanban-column implementation from the prior revision, but renames the `needs_review` board lane label from "Needs review" to "In review" across product-ui, renderer tests, smoke tests, and translations.

## Summary of Changes
- Updated the English and localized `column.needs_review` copy to read "In review".
- Adjusted board/unit/smoke tests and comments to match the renamed lane label.
- Left the underlying `needs_review` column semantics and status-to-column fallback unchanged.

## Findings
- **Medium**: The new lane label no longer matches the column's semantics. `needs_review` still contains sessions where a person owns the next step, including `changes_requested`, `needs_input`, and `ci_failed`, but those now render under "In review", which implies the opposite.

## Regression Risk
This is primarily a UX/semantics regression: the board will now describe worker-blocked sessions as if they are merely awaiting review, which makes the new kanban ownership model less trustworthy.

## Testing
Updated renderer/product-ui/smoke tests for the renamed label. I did not run tests. The existing tests still show that statuses like `changes_requested` and `needs_input` map to `needs_review`, which is why the new wording is misleading.

packages/product-ui/src/session-presentation.ts
neversettle17-101: Leaving this as is: the label is a deliberate product decision, requested and then explicitly confirmed by the repo owner, who defined the lane as "waiting on a person" and chose "In review" for it. I raised the same ambiguity you are raising before making the change, and it was reaffirmed, so I am not reverting it unilaterally. Flagging your reasoning to them rather than closing this thread.

Two clarifications on the technical claim, since they narrow it:

**The daemon-derived path does not route `needs_input` here.** Column placement comes from PR facts, not from display status — a session whose agent is waiting on input, with no PR, derives `building`. The mapping you are pointing at (`changes_requested` / `needs_input` / `ci_failed` -> `needs_review`) is the compatibility fallback in `toKanbanColumn`, which only runs when a daemon omits `kanbanColumn`. On the primary path those statuses do not imply this column.

**Where the point does land** is a live PR with changes requested, or failing CI, while `autoInjectReview` / `autoInjectCI` are off. Those genuinely derive `needs_review`, a person genuinely owns the next step, and "In review" describes that imprecisely — nothing is under review, someone just has to act. That subset is real.

Worth recording for whoever revisits this: the lane is the fallthrough branch of `derivePRKanbanColumn`. `validating` is an explicit list of AO-owned conditions, and a card lands here precisely because no automation claimed it. The semantics are "a person owns the next step"; the label is the owner's wording for that, and the enum value stays `needs_review` throughout the contract, API, and wire, so any relabel is one string per locale.

packages/product-ui/src/session-presentation.ts
```diff
@@ -43,6 +46,11 @@ const englishLabels: Record<SessionPresentationMessageKey, string> = {
 	"zone.pending": "In review",
 	"zone.working": "Working",
 	"zone.done": "Terminated",
+	"column.building": "Building",
+	"column.validating": "Validating",
```
neversettle17-101, line 52: `needs_review` still represents the lane where a person owns the next step, not just PRs actively being reviewed. The fallback tests in this PR still map `changes_requested`, `needs_input`, and `ci_failed` into this same column, so labeling it "In review" misdescribes those sessions. A label closer to ownership/attention (for example "Needs review" or similar) would match the column semantics better.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Rebases the kanban-column board work onto `main`, adding the daemon-derived `kanbanColumn` end to end and updating the board UI/tests/translations around the new lanes.

## Summary of Changes
- Added backend contract/store/service/API support for a derived `kanbanColumn`.
- Reworked the board UI and tests to group cards by kanban column, with fallback placement for older daemons.
- Updated product-ui translations and lane presentation for the new column labels.

## Findings
- **Medium**: The `needs_review` lane is still labeled "In review", but that column also contains states where a human owns the next step outside of active review, such as `changes_requested`, `needs_input`, and `ci_failed`.

## Regression Risk
Primary risk is UX semantics: users may trust the new ownership-based board less if a worker blocked on input or requested changes is shown as though it is currently under review.

## Testing
The PR includes added/updated contract, store, service, renderer, and smoke tests. I did not run tests. The existing `toKanbanColumn` tests still demonstrate that `changes_requested` and `needs_input` map into `needs_review`, which is why the current label is misleading.

packages/product-ui/src/session-presentation.ts
neversettle17-101: Same finding as the thread just above (r3836216619), re-raised against the rebased head. Answered there in full; not repeating it here.

Short version: the label is a deliberate product decision by the repo owner, who defined this lane as "waiting on a person" and chose "In review" for it. I raised the same ambiguity before making the change and it was reaffirmed, so I am not reverting it on my own judgement. Both threads are left open for them.

The one narrowing worth carrying over: `needs_input` does not reach this column on the daemon-derived path. Placement comes from PR facts, so a session with no PR derives `building` whatever its agent is doing. The `changes_requested` / `needs_input` / `ci_failed` mapping you are citing is the compatibility fallback in `toKanbanColumn`, which runs only when a daemon omits the field. Where the point does land is a live PR with changes requested or failing CI while the matching inject flag is off -- those derive `needs_review` for real, and "In review" is loose for them.

packages/product-ui/src/session-presentation.ts
neversettle17-101: Closing this out: the repo owner has confirmed the label stands. "In review" is the intended wording for this lane.

No code change, but I did record the reasoning next to the string in 9d1ad6e74, since this has now been raised twice and the note belongs where someone would go to change it: the lane is the fallthrough of `derivePRKanbanColumn`, so it means "a person owns the next step" rather than "a PR under review", and that deliberately includes a live PR with changes requested or failing CI while the matching inject flag is off.

Resolving as considered and declined, not as fixed. The enum stays `needs_review` throughout the contract, API, and wire, so revisiting the wording later is one string per locale with no code movement.

packages/product-ui/src/session-presentation.ts
neversettle17-101: Closing this out: the repo owner has confirmed the label stands. "In review" is the intended wording for this lane.

No code change, but I did record the reasoning next to the string in 9d1ad6e74, since this has now been raised twice and the note belongs where someone would go to change it: the lane is the fallthrough of `derivePRKanbanColumn`, so it means "a person owns the next step" rather than "a PR under review", and that deliberately includes a live PR with changes requested or failing CI while the matching inject flag is off.

Resolving as considered and declined, not as fixed. The enum stays `needs_review` throughout the contract, API, and wire, so revisiting the wording later is one string per locale with no code movement.

neversettle17-101:
Screenshot 2026-08-23 at 3 42 31 PM

neversettle17-101 (commented):
Approving.

## What This PR Does
Rebases the kanban-column board work onto `main`, adds the daemon-derived `kanbanColumn` end to end, and now explicitly documents that `validating` and `needs_review` are the same review-feedback loop seen from AO's turn vs a person's turn.

## Summary of Changes
- Added backend contract/store/service/API support for a derived `kanbanColumn`.
- Reworked the board UI and tests to group cards by kanban column, with fallback placement for older daemons.
- Clarified the semantics of the `needs_review`/"In review" lane in product-ui, renderer comments, and contract/domain documentation so the label matches the intended review-loop model.

## Findings
- No clear regression found from the reviewed diff. The latest changes resolve the earlier ambiguity by documenting that the middle lanes represent one review-feedback loop split by whose turn it is, and the tests/comments now reflect that intent.

## Regression Risk
Main risk remains behavioral drift between daemon-derived columns and legacy status-based fallback during upgrades, but the PR still includes explicit fallback logic and coverage for older daemons that omit `kanbanColumn`.

## Testing
The PR includes added/updated contract, store, service, renderer, and smoke tests, plus coverage for stale human review verdicts and older daemons that omit `kanbanColumn`. I did not run tests.

backend/internal/service/session/kanban.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func toContractKanbanPRFacts(prs []domain.PRFacts, runs []domain.CurrentHeadReviewRun) []contract.KanbanPRFacts {
+	byPR := make(map[string]contract.KanbanReviewRunFacts, len(runs))
+	for _, run := range runs {
+		facts := byPR[run.PRURL]
+		facts.Present = true
+		facts.Running = facts.Running || run.Status == domain.ReviewRunRunning
+		facts.ChangesRequested = facts.ChangesRequested || run.Verdict == domain.VerdictChangesRequested
```
illegalcall, line 40: Same-head reruns can leave this card stuck in an old result. We allow the same reviewer harness to rerun after `changes_requested` on the same SHA (`TestTriggerRerunsChangesRequestedCurrentHead`), so `runs` can contain an older changes-requested pass and a newer approved pass. This OR-fold keeps `ChangesRequested=true`; then `approvedByAO` stays false and an auto-review session remains in Validating even though the latest pass approved it.

Potential fix: select the latest run per `(pr_url, target_sha, harness)` first, then aggregate those latest states across different harnesses. A service test for `changes_requested -> approved` on the same SHA would cover the failure.

backend/pkg/contract/kanban.go
```diff
@@ -0,0 +1,378 @@
...
+
+// aoOwnsNextStep reports whether AO itself is turning the PR's review-feedback
+// loop: its review pass on the current head is still running, it is addressing
+// review feedback, or it is fixing failing CI. When it is not, the same loop
+// continues with a person taking the next turn.
+func aoOwnsNextStep(session KanbanSessionFacts, pr KanbanPRFacts) bool {
+	if pr.ReviewRun.Running {
+		return true
+	}
+	if session.AutoInjectReview && (pr.ReviewRun.ChangesRequested || pr.ExternalReview.ChangesRequested) {
```
illegalcall: This makes external human feedback switch the card to Validating when auto-inject is enabled. That conflicts with #4179's source split: AO/internal review facts drive Validating, while human/external facts stay In review; its expected status for this case is `Addressing comments`. With this condition, the matching branch in `inReviewDisplayStatus` is unreachable.

Potential fix: only use `pr.ReviewRun.ChangesRequested` in this AO-owned condition and let external changes fall through to In review, where auto-inject chooses `Addressing comments`. If the intended model is instead "who is acting now", then #4179 should be updated because the current implementation and written contract disagree.

backend/internal/service/session/service.go
```diff
@@ -1026,15 +1027,41 @@ func (s *Service) toSession(ctx context.Context, rec domain.SessionRecord) (doma
...
 
+// currentHeadReviewRuns reads the session's AO review passes for the Kanban
+// reducer. Sessions the reducer already decides without them — terminated ones
+// and ones with no PR yet — skip the query, so listing a board of building
+// workers stays one read per session.
+func (s *Service) currentHeadReviewRuns(ctx context.Context, rec domain.SessionRecord, prs []domain.PRFacts) ([]domain.CurrentHeadReviewRun, error) {
+	if rec.IsTerminated || len(prs) == 0 {
+		return nil, nil
+	}
+	runs, err := s.store.ListCurrentHeadReviewRunsForSession(ctx, rec.ID)
```
illegalcall, line 1155: This adds one review-run query for every live PR-bearing session. `Service.List` calls `toSession` inside the card loop, so a board with N PR sessions now performs N PR-fact reads plus N current-head review reads. The review query also returns every same-head retry. #4178 explicitly asks for bounded list-path queries and no per-card lookups.

Potential fix: add a batch store read for all relevant session IDs, call it once before the loop, and group the results by session/PR. That batch query can also select the latest run per PR and harness, which fixes the superseded-run problem above. A query-count test around `Service.List` would keep the boundary enforced.

i-trytoohard:
From the NotesBot call on `2026-08-23`, this existing pr appears to be relevant.
Match confidence: `0.69` (fuzzy).

Relevant discussion excerpts:
- - The same update said the review tab also needed simplification so that **models** are added alongside **harnesses**, and mockups were prepared for discussion with aditi after the call. - aditi said she was mainly working on **Kanban** and did not have more updates on reviews. - aditi said the **Kanban board [...]
- - aditi said she was mainly working on **Kanban** and did not have more updates on reviews. - aditi said the **Kanban board backend changes** for rendering column data, updated column names, and display status are done and in review. - aditi said she and ronish had discussed the Kanban design, and ronish would [...]
- - aditi said she was mainly working on **Kanban** and did not have more updates on reviews. - aditi said the **Kanban board backend changes** for rendering column data, updated column names, and display status are done and in review.

Source thread: https://discord.com/channels/1476302178913357958/1541141481157566515/1541141481157566515

backend/internal/service/session/kanban.go
neversettle17-101: Addressed in 985ed8a11 as part of the list-path batch read.

The batched review-run query now selects the latest current-head pass per `(session_id, pr_url, target_sha, harness)` before the service folds runs into `KanbanReviewRunFacts`, so an older same-head result cannot pin the card beside the retry that replaced it. That also keeps the list path aligned with the board requirement to avoid per-card lookups.

Coverage:
- `TestListCurrentHeadReviewRunsForSessionsKeepsLatestRunPerHarness`
- `TestListBatchesKanbanReads`

That said, you were also right about the stale-result shape in general: the previous OR-fold over every returned run was not robust once the read path could see more than one current-head pass for the same reviewer harness.

backend/pkg/contract/kanban.go
neversettle17-101: Addressed in 985ed8a11.

I narrowed the AO-owned branch exactly the way you suggested: `aoOwnsNextStep` now only treats AO review feedback as `Validating` when the current-head AO pass itself requested changes. External human `changes_requested` now falls through to `needs_review`, and `inReviewDisplayStatus` becomes the source of the "Addressing comments" phrase when auto-inject is on.

That restores the intended split from #4179:
- AO/internal review facts decide `Validating`
- external human review facts stay in the person-owned lane

Coverage:
- `TestSessionKanbanExternalChangesStayPersonOwned`
- contract cases updated for external `changes_requested` + `AutoInjectReview`
- `go test ./internal/service/session ./pkg/contract`

backend/internal/service/session/service.go
neversettle17-101: Addressed in 985ed8a11.

I reworked the session-list read path so the board no longer does per-card enrichment queries for Kanban facts:
- `Service.List` filters records first, then batches PR facts once for all visible session ids
- it batches current-head review runs once for the same id set
- the service groups both result sets by session before deriving session presentation

On the storage side, the new batch review-run query also keeps only the latest current-head pass per `(session, pr, target_sha, harness)`, so the list path no longer carries superseded same-head retries forward.

Coverage:
- `TestListBatchesKanbanReads` asserts one batched PR-facts read and one batched review-run read for the whole list call
- `TestListPRFactsForSessionsBatchesBySession`
- `TestListCurrentHeadReviewRunsForSessionsKeepsLatestRunPerHarness`
- `go test ./internal/service/session ./internal/storage/sqlite/store`

backend/internal/storage/sqlite/queries/review.sql
```diff
@@ -75,3 +75,41 @@ FROM review_run WHERE session_id = ? AND status = 'running' AND verdict = '' ORD
 -- name: ListReviewRunsByBatch :many
 SELECT id, review_id, session_id, harness, pr_url, target_sha, status, verdict, body, created_at, github_review_id, delivered_at, batch_id, auto_inject_review, trigger_source
 FROM review_run WHERE session_id = ? AND batch_id = ? ORDER BY created_at ASC, id ASC;
+
+-- name: ListCurrentHeadReviewRunsBySession :many
```
neversettle17-101, line 79: ListCurrentHeadReviewRunsBySessions now keeps only the latest same-head pass per (session, pr, harness), but ListCurrentHeadReviewRunsBySession still returns all runs for that head. Service.Get calls the single-session path, and toContractKanbanPRFacts ORs Running / ChangesRequested / Failed across every returned run, so a superseded retry from the same harness can keep stale negative facts alive on the detail read even after a newer retry approved. Please apply the same de-duplication rule here, or reuse the batched implementation, so list and get derive the same kanban/display state.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Optimizes session-list kanban reads by batching PR facts and current-head review runs, refines the review-loop semantics for external changes requests, and updates the board card UI with richer PR/status presentation.

## Summary of Changes
- Added batched store/service reads for PR facts and current-head review runs.
- Changed kanban ownership so human changes requests remain person-owned even with auto-inject enabled.
- Refreshed board cards with PR lifecycle icons, reviewer avatars/comment counts, demo preview data, and lane-based status styling.

## Findings
- **Medium**: Get still reads stale same-head retries because the single-session review-run query was not updated to match the new batched de-duplication logic.

## Regression Risk
The biggest risk is divergence between list and detail reads: the board can show the latest AO review outcome while the single-session inspector/card detail path still aggregates superseded retries for the same head/harness.

## Testing
The PR adds batching tests for list reads and UI coverage for the new board presentation. I did not run tests. I did not see coverage for a single-session Get path where an older same-head retry is superseded by a newer one from the same harness.

backend/internal/storage/sqlite/queries/review.sql
neversettle17-101, line 79: `ListCurrentHeadReviewRunsBySessions` now keeps only the latest same-head run per `(session, pr, harness)`, but `ListCurrentHeadReviewRunsBySession` still returns every run for that head. `Service.Get` uses the single-session path, and `toContractKanbanPRFacts` ORs `Running`, `ChangesRequested`, `Failed`, etc. across all returned rows, so a superseded retry from the same harness can keep stale negative facts alive on the detail read even after a newer retry approved. Please apply the same de-duplication rule here, or route the single-session path through the batched implementation, so list and get derive the same kanban/display state.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Updates the kanban board/read-model work again, including batched PR/review-run reads for session lists, revised review-loop ownership semantics, and richer board card presentation.

## Summary of Changes
- Added batched store/service reads for PR facts and current-head review runs to reduce per-session board queries.
- Adjusted kanban ownership so external human changes requests stay person-owned even when auto-inject review is enabled.
- Expanded board card presentation with PR lifecycle icons, comment/reviewer metadata, demo preview data, and lane-driven status styling.

## Findings
- **Medium**: `Get` still disagrees with list reads when the same harness retries a review on the same PR head, because the single-session current-head query still returns superseded retries while the batched list query de-duplicates them.

## Regression Risk
The main risk is inconsistent derived state between the board and per-session detail surfaces: one path can show the latest AO review outcome while the other still aggregates stale retry facts for the same head.

## Testing
This revision adds batching tests and more board/UI coverage. I did not run tests. I still do not see coverage for a single-session `Get` path where an older same-head retry is superseded by a newer retry from the same harness.

neversettle17-101 (commented):
Approving.

## What This PR Does
Follows up on the kanban/review-run batching work by making the single-session current-head review-run query use the same latest-per-harness de-duplication rule as the batched list path, and adds tests for both the store and service `Get` path.

## Summary of Changes
- Updated `ListCurrentHeadReviewRunsBySession` to keep only the latest same-head run per `(pr, harness)`, matching the batched query semantics.
- Added SQLite store coverage proving the single-session path drops superseded same-head retries while preserving one latest run per harness.
- Added service-level coverage proving `Get` now derives kanban/display state from the latest rerun on the current head.

## Findings
- No clear regression found from the reviewed diff. The previous inconsistency between list and detail reads is addressed directly in the SQL and covered by targeted tests.

## Regression Risk
Residual risk is low and mostly limited to future drift between the single-session and batched query forms, but this change makes their semantics explicit and aligned.

## Testing
Added targeted store and service tests for latest-current-head review-run selection, including the same-head rerun case that previously diverged between list and get. I did not run tests.

neversettle17-101 (commented):
Approving.

## What This PR Does
Adds clarifying comments to the new batched PR/review-run store helpers and fixes the board adapter to pass `updatedAt` through to the product-ui session card so the footer timestamp label can render from the intended field.

## Summary of Changes
- Documented that the batched PR-facts and current-head-review-runs helpers always return every requested session id with empty slices when there is no data.
- Passed `updatedAt` into `toBoardSessionPresentation`, matching the product-ui card contract.
- Wired the desktop card's footer label callback to format and present that timestamp consistently.

## Findings
- No clear regression found from the reviewed diff. The adapter change aligns the renderer payload with the existing `BoardSessionPresentation` shape and the card's footer usage.

## Regression Risk
Low. The changes are narrowly scoped to documentation and to supplying a field the card already expects.

## Testing
No new tests in this follow-up. I did not run tests. The touched frontend change is small and consistent with the existing `BoardSessionPresentation.updatedAt` contract.

neversettle17-101:
Screenshot 2026-08-27 at 6 50 39 PM

backend/pkg/contract/kanban.go
```diff
@@ -0,0 +1,378 @@
...
+	Review         ReviewDecision
+	Mergeability   Mergeability
+	UpdatedAt      time.Time
+	ReviewRun      KanbanReviewRunFacts
+	ExternalReview KanbanExternalReviewFacts
+}
+
+func derivePRKanbanColumn(session KanbanSessionFacts, pr KanbanPRFacts) KanbanColumn {
+	switch {
+	case pr.Draft:
```
Pulkit7070: [low] Ordering bug: a PR closed while still a draft lands in `validating` instead of `ready`. This switch tests `case pr.Draft:` before `case pr.Merged || pr.Closed:`, so a PR with `Closed=true, Draft=true` matches Draft first and returns `KanbanValidating`.

Confirmed by running the reducer:

    DeriveKanbanPresentation({}, [{Closed:true, Draft:true}]) -> column "validating"  (want "ready")

It's narrow, since live PRs are filtered out first and a closed non-draft or merged PR outranks it, so it only surfaces when a session's sole terminal PR is a closed draft, which then shows in the active validating lane with a "Draft" badge rather than as closed in ready. Suggest ordering the `Merged || Closed` case before `Draft` (a terminal state should win), and adding a `Closed && Draft` case to the reducer tests, which currently do not exercise that combination.

packages/product-ui/src/SessionsBoardView.tsx
```diff
@@ -132,78 +145,38 @@ function BoardColumnView<TSession extends BoardSessionPresentation>({
...
-				primarySessions={readySessions}
-				primaryTone={splitLaneTone("ready", labels.tones.ready)}
-				renderSessionCard={renderSessionCard}
-				secondarySessions={mergedSessions}
-				secondaryTone={splitLaneTone("merged", labels.tones.merged)}
-				zone="merge"
-			/>
-		);
-	}
+	const ordered = [...sessions].sort((left, right) => {
```
neversettle17-101, line 153: The lane sorter still prioritizes only `attentionZone(status) === "action"` or `activity?.state === "blocked"`, while `boardSessionNeedsAttention` below now also treats display statuses like `Blocked`, `Checks failing`, and `Changes requested` as attention. That means a session with `status: "idle"` and `displayStatus: "Blocked"` gets the orange attention treatment but will not be pinned first in its lane. Please drive the ordering from the same attention predicate (or otherwise include these display-status cases) so the board's visual urgency and sort order stay consistent.

neversettle17-101 (commented):
Requesting changes.

## What This PR Does
Updates the product-ui board card so attention styling can be driven by daemon `displayStatus` phrases like "Blocked", "Checks failing", and "Changes requested", and expands the portable card tests around that behavior.

## Summary of Changes
- Added `boardSessionNeedsAttention` to derive the orange attention treatment from selected `displayStatus` values as well as legacy raw status/activity signals.
- Updated session card tests to cover attention styling for those display statuses and to keep non-attention phrases neutral.
- Adjusted the validating display-status expectation to use the validating lane color instead of the attention tone.

## Findings
- **Medium**: Attention styling and lane ordering now disagree. Cards with an attention `displayStatus` can glow orange but still sort as ordinary cards because the lane sorter still only looks at `status` and blocked activity.

## Regression Risk
The main risk is inconsistent board signaling: a card can visually advertise itself as urgent while remaining buried below newer neutral cards in the same lane.

## Testing
The PR adds portable session-card coverage for attention styling by `displayStatus`. I did not run tests. I did not see coverage for lane ordering when attention comes only from `displayStatus` rather than from raw `status` or `activity`.

neversettle17-101 (commented):
Approving.

## What This PR Does
Follows up on the board attention-state work by making lane ordering use the same attention predicate as the card styling, and fixes the kanban contract so a closed draft PR still resolves to `ready`.

## Summary of Changes
- Updated the board lane sorter to prioritize cards via `boardSessionNeedsAttention`, keeping sort order aligned with the orange attention treatment.
- Narrowed the fallback branch inside `boardSessionNeedsAttention` to the same action/blocking semantics already used for raw status/activity signals.
- Adjusted the kanban contract ordering so terminal PR states win over `Draft`, with targeted tests for closed-draft and display-status attention ordering.

## Findings
- No clear regression found from the reviewed diff. The prior styling-versus-ordering inconsistency is addressed directly, and the contract edge case is covered by a focused test.

## Regression Risk
Low. The changes are small and align previously divergent logic paths rather than introducing a new behavior split.

## Testing
Added targeted contract and product-ui tests for the closed-draft case and for attention ordering driven purely by `displayStatus`. I did not run tests.

neversettle17-101 (commented):
Approving.

## What This PR Does
Adds a small product-ui follow-up so a custom `statusPresentation` suppresses the board's generic attention styling, with targeted test coverage for that override case.

## Summary of Changes
- Updated `boardSessionNeedsAttention` to return false when a custom `statusPresentation` is present.
- Added a focused session-card test covering a switching-style override on top of an otherwise attention-worthy status/displayStatus combination.

## Findings
- No clear regression found from the reviewed diff. The new guard keeps the generic orange attention treatment from fighting a higher-priority custom status presentation.

## Regression Risk
Low. This is a narrowly scoped presentation override that aligns the helper with the existing `statusPresentation` precedence on the card label itself.

## Testing
Added targeted portable-card coverage for the custom `statusPresentation` override case. I did not run tests.

illegalcall (changes requested):
Requesting changes for one confirmed correctness issue.

### Blocking: ignore human review verdicts from an older PR head

`ListPRFactsBySession` and its batch form find each human reviewer's latest decisive review, but they never require `pr_reviews.target_sha` to match the PR's current `head_sha`. Because `externallyApproved` combines that flag with the provider's aggregate decision, an approval for an older commit can make the current commit derive as `Ready` / `Approved`.

This happens when:

1. A human approves commit `head1`.
2. The contributor pushes commit `head2`.
3. The repository does not dismiss stale approvals on push, so the provider's aggregate `review_decision` remains `approved`.
4. The SQL still reports the `head1` review as `external_approved`, and the reducer treats `head2` as ready.

I reproduced this against the current head: with `pr.head_sha = head2` and the only human approval carrying `target_sha = head1`, `ListPRFactsForSession` returned `ExternalApproved: true`.

That violates #4179's requirement that stale review facts from old SHAs must not drive current presentation, and the stated rule that a new head starts a new validation cycle.

Please make both the singleton and batch external-review reductions current-head-aware, with an explicit compatibility policy for providers/legacy rows whose `target_sha` is empty. Please also add a real-SQLite regression covering:

- old-head human approval + new PR head + aggregate still `approved` does not become Ready because of that old approval;
- a current-head human approval still becomes Ready.

The other points I checked—attention-first ordering and external-feedback lane ownership—are intentional or have conflicting written requirements, so I am not requesting changes for them here.

backend/internal/domain/pr.go
```diff
@@ -24,6 +24,13 @@ type PRFacts struct {
 	TargetBranch   string
 	HeadSHA        string
 	UpdatedAt      time.Time
+	// ExternalApproved and ExternalChangesRequested are the human review
+	// verdicts AO did not author. Review above aggregates AO's own provider
+	// reviews with everyone else's, so it cannot say whose turn the
+	// review-feedback loop is on.
+	ExternalApproved         bool
```
neversettle17-101, line 31: These new `ExternalApproved` / `ExternalChangesRequested` / `ExternalComments` fields are later passed through `deduplicatePRFacts`, but `mergePRFacts` still only preserves `ReviewComments`. When two alias rows for the same GitHub PR collapse, whichever row wins by `UpdatedAt` can clear these booleans, so a transferred or canonicalized PR can regress from `Commented` or `Changes requested` back to `Needs human review` before `toSessionWithFacts` derives the board state. Please OR the external-review fields during alias collapse and add a regression test for that path.

neversettle17-101 (commented):
## What This PR Does
This PR derives kanban display status from external review input, renames the needs-review substatus strings to "CI failing" and "Commented", and updates the board usage metric to render as plain text with accessible labels. It also threads the new display fields through the API, session mappers, fake bridge, and board tests.

## Summary of Changes
- Split external review comments and verdicts out of PR facts so the kanban reducer can distinguish AO authored review activity from human review feedback.
- Updated board/status presentation strings and accessibility for the usage metric, including new translations and tests.
- Added backend tests for current-head review filtering and external review/comment derivation.

## Findings
- **Medium**: PR alias deduplication drops the new external review signals when duplicate GitHub PR rows collapse, so transferred or canonicalized aliases can lose the very facts this patch now depends on.

## Regression Risk
Transferred or renamed GitHub PRs are the clearest risk area because their duplicate facts are still collapsed before the board reducer runs.

## Testing
Reviewed the added backend and frontend tests in the diff. I did not run tests locally.

Needs changes.

neversettle17-101 (commented):
## What This PR Does
This follow-up fixes the alias-collapse regression in PR fact deduplication by preserving the new external review booleans when canonical and transferred GitHub PR rows are merged. It also extends the existing deduplication test to cover those fields.

## Summary of Changes
- OR `ExternalApproved`, `ExternalChangesRequested`, and `ExternalComments` across duplicate PR facts during alias collapse.
- Expand the transferred-alias regression test to assert all external review flags survive deduplication.

## Findings
- No clear regression found in the reviewed follow-up diff.

## Regression Risk
Low. The change is narrowly scoped to the alias-merge helper and directly covers the previously reported loss of external review signals.

## Testing
Reviewed the updated unit test coverage in `service_test.go`. I did not run tests locally.

Ready to merge.

illegalcall (changes requested):
Requesting changes after re-checking the latest head (`ce43845b`). The earlier stale-human-approval bug is fixed in both singleton and batch SQL, including the documented empty-SHA compatibility behavior. Two separate correctness issues remain.

### Blocking: `auto_inject_review` is not an authorship signal

Both PR-facts queries classify an unresolved comment as external only when `pr_comment.auto_inject_review = 0`. That column records whether automatic injection was enabled when the comment was first observed; it does not say whether AO or a human wrote the comment. The SCM observer assigns `sessionRecord.AutoInjectReview` to every observed comment.

This happens in the normal default-enabled flow:

1. `autoInjectReview` is enabled for the session.
2. A human leaves an unresolved review comment.
3. The observer stores that human comment with `auto_inject_review = 1`.
4. `ListPRFactsBySession` reports `ExternalComments = false`.
5. The board derives `Needs human review` instead of `Addressing comments` / `Commented`.

The reverse misclassification is also possible: a non-bot AO-authored comment observed while the policy is disabled can be treated as external. I reproduced the first case against the real SQLite store: `ReviewComments` was true while `ExternalComments` was false.

Please separate source from policy. `pr_comment.review_id` can be matched to AO's recorded `review_run.github_review_id`, analogous to the verdict filtering, with an explicit compatibility policy for rows without a review id. Add singleton and batch regressions where a human comment remains external with auto-inject both on and off.

### Blocking: transferred-alias collapse can discard current-head AO runs

`toSessionWithFacts` collapses duplicate transferred PR aliases to the newest URL, but the separate current-head review-run slice is not collapsed or remapped. `toContractKanbanPRFacts` then joins runs by exact URL. If an upgraded database has the run on the older alias, the surviving PR fact cannot see it.

This happens when:

1. Legacy rows contain the old and new repository URLs for the same transferred PR/head.
2. A current-head AO run is attached to the older URL.
3. PR-fact deduplication keeps the newer URL.
4. The exact-URL lookup misses the run.
5. A genuinely running AO review derives as `Needs human review` rather than `Reviewing` in `Validating`.

I reproduced that reducer path locally. New-write canonicalization does not make the upgraded-data case impossible; the service-level alias fallback exists specifically for legacy duplicate rows. Please preserve/remap review-run facts across the same alias groups and add a regression covering an old-alias current-head run.

neversettle17-101 (commented):
## What This PR Does
This follow-up fixes two remaining board-state edge cases: review runs are remapped onto the canonical PR alias before kanban derivation, and external PR comments are now identified by authored review id instead of the session auto-inject setting. It also extends the focused backend tests for both paths.

## Summary of Changes
- Canonicalize current-head review runs against deduplicated PR aliases so transferred GitHub PR URLs still keep their active review state.
- Treat comments as external unless they can be matched to an AO-authored GitHub review run, which preserves human comments even when auto-inject was enabled at capture time.
- Add regression tests for alias-remapped review runs and the updated external comment classification in single-session and batched store reads.

## Findings
- No clear regression found in the reviewed follow-up diff.

## Regression Risk
Low. The changes are tightly scoped to kanban/read-model derivation and are covered by targeted backend tests for the reported edge cases.

## Testing
Reviewed the added unit/integration-style store and service tests in the diff. I did not run tests locally.

Ready to merge.

illegalcall (changes requested):
Requesting one focused change on the latest head (`76d3a1955`). The two previously reported defects are substantially addressed: human comments are now classified independently of the auto-inject policy, and a current-head run that exists only on an older transferred alias is remapped to the canonical PR.

### Blocking: alias remapping reintroduces superseded same-harness runs

The review-run SQL keeps the latest same-head run per `(pr_url, harness)`. `canonicalizeCurrentHeadReviewRuns` then remaps the old and canonical PR URLs to one URL, but it leaves both returned runs in the slice. `toContractKanbanPRFacts` OR-folds their states, so a stale negative run from the old alias can outvote the newer canonical run.

This happens when:

1. An upgraded database has old and canonical URL rows for the same transferred PR/head.
2. The same harness has an older run on the old URL, for example `running` or `changes_requested`.
3. That harness later completes an approved run on the canonical URL.
4. SQL returns one latest run for each exact URL.
5. Alias remapping gives both runs the canonical URL, and the reducer combines them instead of selecting the latest.

I reproduced the `old alias = running`, `canonical alias = approved` case: the session derived as `Validating / Reviewing` rather than allowing the approved rerun to supersede the old run.

Please preserve the existing latest-per-harness rule after canonicalization, using the run timestamp/id (or another deterministic source of recency), and add a regression with two same-head, same-harness runs split across the old and canonical aliases. The current alias test covers only the case where the run exists on the old alias and has no competing canonical run.

neversettle17-101 (commented):
## What This PR Does
This follow-up makes canonicalized alias review runs deterministic by carrying each run id and timestamp through the store layer, then collapsing duplicate canonical-url runs per harness to the newest one before kanban derivation. It also adds focused service tests for the remapped and superseded alias-run cases.

## Summary of Changes
- Extend `CurrentHeadReviewRun` and the review-run queries/store mappers with `id` and `createdAt`.
- After remapping transferred PR aliases to a canonical URL, deduplicate runs per canonical PR and harness by newest `createdAt`/`id`.
- Add regression tests covering alias-remapped runs and the case where a newer canonical-url run should supersede an older alias-url run.

## Findings
- No clear regression found in the reviewed follow-up diff.

## Regression Risk
Low. The change is narrowly scoped to the review-run normalization path used by kanban derivation, and the added tests cover the specific alias-collision scenario it fixes.

## Testing
Reviewed the updated service/store unit tests in the diff. I did not run tests locally.

Ready to merge.

Pulkit7070 (approved):
The alias-remapping fix resolves the blocking issue cleanly: after canonicalization, runs are deduped per (canonical URL, harness) keeping the latest by created_at then id, so a stale old-alias run can no longer OR-fold over the newer canonical run. I confirmed it by reverting `pr_alias.go` to the prior head with the new test in place, which reproduces the old behavior (the column resolves to `validating`), and the fix restores `needs_review`. The in-memory dedup matches the SQL currency rule, and both the single-session and batch board-list reads flow through the one canonicalization site, so the board view is covered too.

The reducer is a clean pure function with sound branch ordering and deterministic tie-breaks, AO's own reviews are kept out of the human verdicts, and current-head currency is enforced in SQL. The session, domain, and store packages pass under the race detector, and the new locale keys are present across all eight locales.

frontend/src/renderer/components/SessionsBoardAdapters.tsx
```diff
@@ -272,17 +275,19 @@ function toUsagePresentation(
 	if (!usage) {
 		return undefined;
 	}
-	const cost = formatEstimatedCost(usage.estimatedCost) ?? t("usage.unavailable");
-	if (processedTokens === null || processedTokens <= 0) {
+	const cost = formatEstimatedCost(usage.estimatedCost);
+	if (!cost) {
```
neversettle17-101, line 280: Returning `undefined` here drops the board usage metric entirely whenever `estimatedCost` is unavailable, even if `processedTokens` is present. That is a user-visible regression from the previous behavior: sessions like the new `s-tokens` test case still have useful token usage, but now they render exactly like a session with no usage data at all. Since `estimatedCost` is nullable by design, this will hide usage for providers/backfill states that have tokens before pricing. Please keep a token-only fallback when cost is missing instead of suppressing the metric completely.

neversettle17-101 (commented):
## What This PR Does
This follow-up changes the board usage metric again: priced sessions now show only the dollar amount on the card, with the full cost-plus-token summary moved into a hover tooltip and the screen-reader label. The accompanying board tests were updated to expect the tooltip behavior and to suppress the old "Unavailable" token fallback.

## Summary of Changes
- Wrap the board usage metric in a tooltip and render cost-only text by default on the card.
- Change `toUsagePresentation` to return no board usage metric at all when `estimatedCost` is absent.
- Update board tests to assert the hover tooltip and the new no-metric behavior for sessions without a cost estimate.

## Findings
- **Medium**: Sessions with token usage but no cost estimate now lose the usage metric entirely on the board, which regresses visibility for providers or historical sessions that still report tokens before pricing is available.

## Regression Risk
The main risk is usage visibility for sessions whose `estimatedCost` is still null: they now look identical to sessions with no usage data at all.

## Testing
Reviewed the updated board tests in the diff. I did not run tests locally.

Needs changes.

frontend/src/renderer/components/SessionsBoardAdapters.tsx
neversettle17-101: Fixed in a94fcac9f.

I restored the token-only fallback when `estimatedCost` is null but `processedTokens` is present, so those sessions no longer disappear from the board usage row. The card still stays cost-first when pricing exists: priced sessions show only the dollar amount by default, while hover/screen-reader text keeps the full cost-plus-token summary. Sessions with neither cost nor positive token usage still render no usage metric.

Coverage updated in `frontend/src/renderer/components/SessionsBoard.test.tsx` for both the mixed cost+token case and the token-only case, and I re-ran:
- `cd frontend && npx vitest run src/renderer/components/SessionsBoard.test.tsx`
- `cd packages/product-ui && npx vitest run src/SessionsBoardView.test.tsx`

neversettle17-101 (commented):
## What This PR Does
This follow-up restores a token-only board usage fallback when pricing is unavailable, while keeping the newer cost-first presentation and hover tooltip for priced sessions. It also expands the board tests to cover the token-only case and the no-data case separately.

## Summary of Changes
- Restore `formatTokenCount` usage in `toUsagePresentation` when `estimatedCost` is null but token counts exist.
- Keep the cost-only compact label for priced sessions, with the full cost-plus-token text still reserved for the tooltip and screen readers.
- Update board tests so token-only sessions render again, while sessions with neither cost nor tokens still omit the metric.

## Findings
- No clear regression found in the reviewed follow-up diff.

## Regression Risk
Low. The change is narrowly scoped to board usage presentation and directly addresses the previously reported token-visibility regression.

## Testing
Reviewed the updated board tests in the diff. I did not run tests locally.

Ready to merge.

illegalcall (approved):

PR #4252, opened 2026-08-22
feat(files): add full-tree file explorer alongside changed-files diff view

Vaibhaav-Tiwari:
## Summary
- Adds a lazy-loaded, git-status-decorated file tree (react-arborist) to the session Files tab, so a human can browse a worker's entire worktree — not just the files an agent touched. A "changed only" toggle keeps today's GitHub-PR-style review flow available (defaults off, i.e. full tree by default).
- Clicking a changed file opens the existing diff view completely unchanged; clicking an untouched file opens a new Shiki-highlighted, read-only view. Both reuse the existing `GET /workspace/file` endpoint and the existing image blob route — no duplicated data-fetching logic.
- New daemon endpoint `GET /sessions/{sessionId}/workspace/tree?path=` returns one directory level at a time (tracked + untracked-but-not-ignored, git-status decorated), reusing the existing compare/status cache and covering single-repo, multi-repo ("workspace" kind), and scratch project kinds.
- `SessionFilesView.tsx` is split into `WorkspaceDiffView.tsx` (the diff-rendering internals — `ReviewDiffBody`/`DiffView`/etc. — kept byte-for-byte behaviorally identical) and a new `SessionFileExplorer` that composes `FileTree`, `FileContentPane`, and `ReadOnlyFileView`.
- Fuzzy filename filtering and live-updates (via the existing workspace SSE signal) are wired in using react-arborist's native search support and the existing `workspace-file-events.ts` invalidation, respectively — no new frontend infrastructure needed for either.

## Test plan
- [x] Backend: `go build ./...`, `go vet ./...` clean; `go test ./internal/service/session/... ./internal/httpd/...` passes (new `workspace_tree_test.go` covers single-repo, multi-repo child-repo, scratch, gitignore respect, and the global file cap).
- [x] Backend: `golangci-lint run` clean on all touched files.
- [x] API contract: `npm run api` (spec + TS regen) produces no drift; `internal/httpd/apispec` parity tests pass.
- [x] Frontend: `tsc --noEmit` clean; production `vite build` succeeds.
- [x] Frontend: full test suite rebuilt for this change — `WorkspaceDiffView.test.tsx`, `FileTree.test.tsx`, `FileContentPane.test.tsx`, `ReadOnlyFileView.test.tsx`, `SessionFileExplorer.test.tsx` (all new/passing), plus updated `SessionView.test.tsx` and `workspace-file-events.test.ts`.
- [x] Live end-to-end: built and ran the isolated dev daemon from this branch's code against a real worker session's real git worktree, and confirmed via direct HTTP calls that `/workspace/tree` correctly does one-level directory slicing, respects `.gitignore` (e.g. excludes `node_modules`), and decorates git status — no regressions across the rest of the API surface.
- [ ] **Not verified**: interactive click-through of the new tree/file-explorer UI inside the actual Electron window. I don't have a way to drive that specific native window's UI myself in this environment — please give it a spin in the desktop app before merging (Files tab → toggle "changed only" → expand a folder → click an untouched file to see Shiki highlighting, click a changed file to confirm the diff view is untouched).


backend/internal/httpd/apispec/specgen/build.go
```diff
@@ -1633,6 +1636,17 @@ func sessionOperations() []operation {
 			},
 			contentTypes: map[int]string{http.StatusOK: "application/octet-stream"},
 		},
+		{
+			method: http.MethodGet, path: "/api/v1/sessions/{sessionId}/workspace/tree", id: "listSessionWorkspaceTree", tag: "sessions",
+			summary:    "List one directory level of a session workspace's full file tree, git-status decorated",
+			pathParams: []any{controllers.SessionIDParam{}, controllers.WorkspaceTreeQuery{}},
+			resps: []respUnit{
+				{http.StatusOK, controllers.ListWorkspaceTreeResponse{}},
+				{http.StatusNotFound, envelope.APIError{}},
```
illegalcall, line 1650: This endpoint can return a 400 response when `cleanWorkspaceRelativeDir` rejects an absolute or traversal path, but that response is missing from the declared contract. Please add `http.StatusBadRequest` here and regenerate the OpenAPI and TypeScript artifacts.

backend/internal/httpd/controllers/sessions.go
```diff
@@ -571,6 +573,24 @@ func (c *SessionsController) getWorkspaceFile(w http.ResponseWriter, r *http.Req
 	envelope.WriteJSON(w, http.StatusOK, workspaceFileResponse(file))
 }
 
+// listWorkspaceTree returns one directory level of the session workspace's
+// full file tree, git-status decorated. Unlike listWorkspaceFiles (changed
+// files only), path is optional: an empty or missing path lists the
+// workspace root.
+func (c *SessionsController) listWorkspaceTree(w http.ResponseWriter, r *http.Request) {
```
illegalcall, line 581: Please add controller-level coverage for this new public route. The service tests do not verify route registration, query forwarding, response serialization, or API error envelopes; cover a successful root/path response, invalid path, missing session, and service failure.

frontend/src/renderer/components/WorkspaceDiffView.tsx
```diff
@@ -0,0 +1,859 @@
...
+		// trading a slightly higher chance of a one-frame blank edge during
+		// very fast scrolling for less of that first-pass cost on big files.
+		overscan: 8,
+		scrollMargin,
+	});
+
+	return { listRef, virtualizer };
+}
+
+function DiffView({
```
illegalcall, line 252: This duplicates the diff renderer that still remains in `SessionFilesView.tsx`, including `DiffView`, hunk/split rendering, and annotation behavior. Please make this an actual shared extraction or remove the obsolete implementation so fixes cannot drift between two copies.

frontend/src/renderer/components/FileTree.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+	useEffect(() => {
+		setLazyData([]);
+		loadedDirsRef.current = new Set();
+	}, [sessionId]);
+
+	useEffect(() => {
+		if (changedOnly || !rootQuery.data) return;
+		loadedDirsRef.current.add("");
+		setLazyData(rootQuery.data.entries.map(entryToNode));
```
illegalcall: Rebuilding `lazyData` from root entries discards every loaded child, but `loadedDirsRef` still marks those directories as loaded. After toggling Changed only back off, or after an SSE-driven root refresh, expanding one of those directories returns early in `loadChildren`, leaving its descendants missing or stale until the session changes. Please reset/reload the directory markers or merge the refetched children.

frontend/src/renderer/components/FileTree.tsx
```diff
@@ -0,0 +1,236 @@
...
+					idAccessor="path"
+					onToggle={handleToggle}
+					onActivate={handleActivate}
+					openByDefault={false}
+					selection={selectedPath ?? undefined}
+					disableDrag
+					disableDrop
+					disableEdit
+					disableMultiSelection
+					searchTerm={filterText}
```
illegalcall, line 195: This search only sees the nodes currently present in `lazyData`. Because unopened directories have empty `children`, filtering for a nested filename returns no result until every ancestor has already been expanded. The advertised filename search needs a complete index/server query or another way to load candidate descendants.

frontend/src/renderer/components/ReadOnlyFileView.tsx
```diff
@@ -0,0 +1,58 @@
...
+// Reuse AO's CSP-compatible highlight.js pipeline so file previews and code
+// shown elsewhere in the renderer share grammars, colors, and caching.
+function HighlightedContent({ content, path }: { content: string; path: string }) {
+	const extension = path.split("/").pop()?.split(".").pop();
+	const language = canonicalLanguage(extension);
+	return (
+		<div className="chat-code min-h-0 flex-1 overflow-auto p-3">
+			<pre className="whitespace-pre font-mono text-xs leading-5 text-foreground">
+				<code>
+					<HighlightedCode code={content} language={language} />
```
illegalcall, line 53: The PR describes this as a Shiki-highlighted view, but production rendering uses the existing `HighlightedCode` pipeline while `useShikiHtml.ts`, `lib/shiki.ts`, and the direct Shiki dependency remain unused. Please either wire in the promised Shiki renderer or remove the dead implementation/dependency and update the stated behavior.

frontend/src/renderer/components/SessionFileWorkspace.tsx
```diff
@@ -0,0 +1,22 @@
...
+	path,
+	sessionId,
+}: {
+	annotation: FileAnnotationModel;
+	path: string;
+	sessionId: string;
+}) {
+	const fileFeedbackActive = annotation.target?.path === path && annotation.target.side === "file";
+	return (
+		<section className="flex h-full min-h-0 flex-col bg-background" data-testid="session-file-workspace">
```
illegalcall, line 15: The file-tabs design requires the active file surface to show the full relative path, and the implementation plan calls for a center header containing it. This surface currently renders only feedback and file content. Please add the visible path header.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -537,14 +541,14 @@ export function CenterPane({
...
-							className="flex h-full min-w-flex-min flex-1 items-center"
+							className="flex h-full min-w-0 flex-1 items-center"
 							onKeyDown={handleTerminalTabListKeyDown}
 							role="tablist"
 						>
 							{/* The owning session is permanent and never participates in overflow or reordering. */}
 							{session ? (
 				<SessionPaneTab
-					isActive={target.kind === "worker"}
+									isActive={target.kind === "worker" && !workspaceFileActive}
```
illegalcall, line 551: The permanent agent tab is still labeled with `sessionTabLabel`, which resolves to the session/task title for workers. The file-tabs design says this tab should be labeled from the active agent surface, such as Terminal or Claude. Please pass and render that surface label instead.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -1042,13 +1051,15 @@ function ChatHeader({
...
+	workspaceTabs?: ReactNode;
+	workspaceFileActive?: boolean;
 	/** Fullscreen content cannot see the normal topbar portal outside its subtree. */
 	inline?: boolean;
 	topbarBounds: TopbarBounds;
 }) {
 	const label = sessionTitle || snapshot.title || snapshot.sessionId;
 	// The chat tab is "selected" only when neither terminal pane is the body.
-	const timelineActive = !reviewerActive && !shellActiveHandleId;
+	const timelineActive = !workspaceFileActive && !reviewerActive && !shellActiveHandleId;
```
illegalcall, line 1115: The chat version has the same labeling issue: the permanent tab label still comes from `sessionTitle || snapshot.title`, not the active agent surface named in the design. Please use the surface/agent label here as well.

illegalcall (changes requested):
Requesting changes. Please address the inline comments and link the originating issue before merge.

illegalcall (approved):

backend/internal/httpd/apispec/specgen/build.go
Vaibhaav-Tiwari: Fixed in 82117fbae: declared 400 for the workspace-tree operation and regenerated both OpenAPI artifacts.

backend/internal/httpd/controllers/sessions.go
Vaibhaav-Tiwari: Fixed in 82117fbae: added controller coverage for path forwarding/serialization, missing sessions, invalid paths, and service failures.

frontend/src/renderer/components/WorkspaceDiffView.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: removed the obsolete unreferenced SessionFilesView implementation and its legacy-only tests, leaving WorkspaceDiffView as the single diff renderer.

frontend/src/renderer/components/FileTree.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: root refreshes now merge matching directory children instead of discarding them while their loaded markers remain set.

frontend/src/renderer/components/FileTree.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: entering a search now recursively hydrates unopened directory branches before react-arborist filters them; a nested-file regression test was added.

frontend/src/renderer/components/ReadOnlyFileView.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: retained the established HighlightedCode pipeline and removed the unused Shiki hook, helper, direct dependency, lockfile entries, and stale test mock.

frontend/src/renderer/components/SessionFileWorkspace.tsx
Vaibhaav-Tiwari: Kept the single-row design intentionally: the second path bar was removed after direct product feedback because it duplicated the file tab strip. The full relative path remains available on each file tab via its native title/tooltip, without regressing the approved compact layout.

frontend/src/renderer/components/CenterPane.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: the permanent terminal tab now uses the active agent/provider label; task titles remain in the sidebar and accessible terminal-panel labeling.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
Vaibhaav-Tiwari: Fixed in 82117fbae: the permanent chat tab now uses the active harness label rather than the session/task title.

PR #4255, opened 2026-08-22
feat(telemetry): instrument the review funnel end to end

Pritom14:
## Why

Reviews are a headline feature with almost no measurement behind them. Four `ao.review.*` events existed, and all of them sat on the manual trigger path, so an **automatic** review was invisible at trigger time: `TriggerAuto` called the engine directly and emitted nothing, which made auto-review indistinguishable from manual review in every funnel even though the two answer different product questions.

Per review feedback this PR now ships the **P0 event set only**. Everything else is listed as P1 at the bottom, with what each deferred event would have answered.

### Daemon

| Event | Change |
|---|---|
| `ao.review.triggered` | adds `trigger` (auto or manual). `Trigger` and `TriggerAuto` now share one instrumented path, so an automatic pass is never invisible |
| `ao.review.trigger_failed` | adds `trigger`, so a failure is attributable to the automatic or the manual path |
| `ao.review.submitted` | adds `trigger` (read from the run), `body_bytes` (review depth, never the text) and `auto_inject` (whether the session policy will let the result reach the worker) |

Payload keys are allowlisted in `remotePayloadAllowlist`; without an entry an event exports with no properties at all rather than failing loudly.

### Renderer

| Event | Purpose |
|---|---|
| `ao.renderer.review_settings_changed` | project-level `auto_review` and `reviewer_harness`, plus `harness_is_default` and per-field changed flags. Emitted only when a review control actually changed, so it counts review decisions rather than every unrelated save. **This is the event that tells us whether the harness dropdown is a decision anybody makes or a checkpoint that could become a default** |
| `ao.renderer.review_auto_review_toggled` | the session-scoped auto-review switch, which duplicates the project setting. Whether anyone reaches for it decides if it stays a separate control |
| `ao.renderer.review_triggered` | manual runs, with the offered `action`, whether a per-session reviewer override was in play, and `source` (`inspector` or `command_palette`). Manual volume against auto volume is the adoption signal, and the two on-ramps stay separable |

`reviewRunActionKind` was extracted from `reviewSessionRunAction` so `review_triggered.action` reports a stable enum rather than the translated label the user happened to see. The command palette is a manual on-ramp in its own right and POSTs to the trigger endpoint directly, so it emits the same event and carries that enum on its action.

Both renderer emit sites fire before the request. That is deliberate: they count INTENT, and the daemon's `ao.review.*` events are the ground truth for what actually ran. A toggle whose save then fails still counts as the user reaching for the switch.

## One behaviour change that is not an event

While a review is running, the once-a-minute auto-review sweep calls `TriggerAuto` again, the engine reports success with nothing created, and the old code emitted another `ao.review.triggered` every time. A six-minute review produced **seven "triggers" for one real pass**, which would have inflated the headline count and spent the 200/day per-name rate limit that real triggers need.

An automatic pass that did no work (a policy race, or a reuse) now reports **nothing at all**. A manual reuse stays a trigger, because the user pressed the button and that is a real action. This was found by running the real coordinator against a real open PR, and reverting it would corrupt the very count P0 exists to measure.

## User flow, and where each P0 event fires

```
Project settings: toggle auto-review / pick a reviewer, save
  -> ao.renderer.review_settings_changed        (only if a review control changed)

Session inspector: flip the session auto-review switch
  -> ao.renderer.review_auto_review_toggled

Path A, manual: press Run / Re-review
  -> ao.renderer.review_triggered               (intent)
Path B, automatic: the coordinator sweeps an idle worker every minute
  -> (silent when there is nothing to do)

Either path reaches the engine
  -> ao.review.triggered { trigger: manual | auto }
  -> ao.review.trigger_failed { error_kind, trigger }

Reviewer agent works, then submits
  -> ao.review.submitted { verdict, trigger, duration_ms, body_bytes, auto_inject }
```

## Privacy

Counts, durations, enums and booleans only. No review body, PR URL, commit SHA, branch, PR title, repo or project name reaches any payload: review prose is reviewer commentary on someone's source, and the URL and SHA identify the repository. Project ids are hashed on the renderer side exactly as the existing settings events do, and reviewer harness ids are validated against AO's own reviewer registry so a caller mistake cannot forward free text.

Both paths keep their existing cost controls: per-event-name rate limits, the `DisabledEvents` kill switch, and an install-scoped `distinct_id`.

## Testing

Verified end to end, not just at the emit site. The payload allowlist that decides what actually reaches PostHog lives in the adapter, keyed by event name, with nothing tying it to the emit site at compile time.

- A **wire test** drives the real emit sites through the real export chain (denylist, aggregation, rate limit, PostHog) into a fake transport, then asserts the event names and properties as they leave the process. Confirmed it catches the real failure: removing one allowlist key makes it fail with `triggered.trigger = <nil>`.
- That test also scans the raw outgoing JSON for the review body, repo name, commit and review id, and fails if any appears under any key.
- Unit tests cover the auto against manual trigger split, the failure path, the submitted payload shape, a manual reuse staying counted, and six no-op automatic sweeps emitting nothing.
- Allowlist drift tests assert the review funnel's exact key sets and reject identifying keys across every `ao.review.*` entry.
- Renderer sanitizer tests cover each of the three events, including that an unknown harness or action is dropped rather than forwarded.

Backend: touched packages pass with `-count=1`, `golangci-lint` reports 0 issues. Frontend: 131 renderer test files, 1927 tests passing across three consecutive runs, `tsc --noEmit` clean.

Pre-existing failures in this worktree, confirmed identical on a clean tree and unrelated to these changes: `adapters/chatdriver/codexappserver` (installed-provider conformance), `internal/cli` (spawn tests needing a live daemon), and the `landing` plus `main` frontend suites (uninstalled dependencies).

## Verified against a live daemon

The earlier full-funnel build was run on a real daemon from this branch, on an isolated data dir with a copy of live app state and remote export off, so nothing left the machine and the running app was untouched. Driving the real auto-review coordinator against a real open PR produced, for the events P0 keeps:

```
ao.review.triggered   {"created_runs":1,"harness":"claude-code","reused":false,"trigger":"auto"}
ao.review.submitted   {"trigger":"auto","verdict":"changes_requested","duration_ms":520165,
                       "body_bytes":103,"auto_inject":true,"posted_to_provider":false,
                       "harness":"claude-code"}
```

The `trigger: auto` on that first event is the fix this PR exists for: before it, that pass emitted nothing. That run is also what surfaced the per-sweep inflation described above.

## One caveat for whoever builds the dashboards

Manual and automatic passes share the single `ao.review.triggered` name, and therefore its 200/day per-name cap on the billed sink, split only by the `trigger` property. On a very high-volume install the billed count can truncate while local SQLite still keeps every occurrence, so exact auto-against-manual counts for a power user should be read from local rather than PostHog. Folding `trigger` into the aggregation tier would collapse the very split this PR adds, so the cap stands.

## P1, deferred by review decision

Removed from this PR. Listed with what each one answered, so picking them up later is a decision about priority and not a rediscovery.

| Deferred event | What goes unmeasured |
|---|---|
| `ao.review.feedback_delivered` | **Whether changes-requested feedback actually reached the worker agent.** A sent nudge, a suppressed one and an errored one look identical. The review loop (review, worker fixes, merge) is the feature's value, and P0 only observes its first half |
| `ao.review.pr_closed` | **Trust.** Whether a PR was merged over a standing changes-requested verdict, how many rounds it took, and reviewed against unreviewed time-to-merge. The override-merge flag was the only perception signal in the catalog, and the control-group data for a future default-on decision |
| `ao.review.feedback_withheld` | Why a completed changes-requested review reached nobody (`auto_inject_off`, `stale_head`, `already_delivered`). The quietest failure mode: a red verdict the worker never hears about |
| `ao.review.auto_skipped` | Why an enabled auto-review never produces reviews (draft PR, already approved, closed, missing head). "I turned it on and nothing happened" for the automatic flow |
| `ao.review.trigger_skipped` | How often automatic passes are skipped or reused. Denominator hygiene; the behaviour it guarded survives as silence |
| `ao.renderer.review_auto_inject_toggled` | Whether anyone touches the delivery switch, which is case data for merging the three review toggles |
| `ao.renderer.review_reviewer_overridden` | Whether anyone sets the per-session reviewer dropdown. Partly mitigated: `review_triggered.has_override` still shows overrides at run time |
| `ao.renderer.review_tab_opened` | Whether humans read findings or only trust the verdict |
| `ao.renderer.review_stopped` | Mid-run dissatisfaction, and the cancel against kill split that would justify merging two stop buttons |
| `ao.renderer.review_comment_sent_to_worker` | Humans hand-relaying comments, a cheap canary for automatic delivery failing |

Two smaller follow-ups, also out of scope here:

- `ao.review.rereview_requested` and `ao.review.comment_resolved` have shipped **payload-stripped** since before this PR, because their names are absent from the allowlist. The two-line fix is a bug fix rather than a new metric, so it is not in this PR either; say the word and it goes in.
- These are **change** events, not state events. A project that already had auto-review on will never emit a toggle event, so "what share of projects have auto-review on today" needs a small daily gauge (counts only, no identifiers) rather than these.


neversettle17-101:
Can we also add the metric for auto-review toggle on?

Pritom14:
> 1. Can we also add the metric for auto-review toggle on?


@neversettle17-101 - its already there 
for Project level - 
`ao.renderer.review_settings_changed {
    auto_review: true|false,        // the new state
    auto_review_changed: true,      // this save actually flipped it
    reviewer_harness, harness_is_default, project_id_hash, ...
  }`
  
  for session-level
`ao.renderer.review_auto_review_toggled { enabled: true|false }`

neversettle17-101:
Lets add P0 metrics to begin with and keep - ao.review.triggered, ao.review.trigger_failed, ao.review.submitted, ao.renderer.review_settings_changed, ao.renderer.review_auto_review_toggled, ao.renderer.review_triggered.

Pritom14:
@neversettle17-101 

Scoped down to the six P0 events. Pushed in `refactor(telemetry): scope review instrumentation to the P0 event set`, and the PR body now describes the P0 surface with the deferred set listed as P1.

**Kept**

| Event | Notes |
|---|---|
| `ao.review.triggered` | with the new `trigger: auto \| manual`, the fix this PR exists for |
| `ao.review.trigger_failed` | with `trigger` |
| `ao.review.submitted` | with `trigger`, `body_bytes`, `auto_inject` |
| `ao.renderer.review_settings_changed` | project auto-review + harness, with per-field changed flags |
| `ao.renderer.review_auto_review_toggled` | the session switch |
| `ao.renderer.review_triggered` | manual runs, with `action` and `has_override` |

**Removed** (8 events, plus their allowlist entries, sanitizer cases and tests): `trigger_skipped`, `feedback_withheld`, `feedback_delivered`, `auto_skipped`, `pr_closed`, `review_auto_inject_toggled`, `review_reviewer_overridden`, `review_stopped`, `review_tab_opened`, `review_comment_sent_to_worker`. The auto-review coordinator, the lifecycle reducer and the daemon wiring are now byte-identical to `main`. Diff is down to 11 files, +628/-12.

**Two things worth a decision, since they are not new metrics:**

1. **A behaviour fix that outlived its event.** While a review runs, the once-a-minute sweep calls `TriggerAuto` again and the engine returns success with nothing created. The old code emitted another `ao.review.triggered` per sweep: a six-minute review produced **7 "triggers" for 1 real pass** (found running the real coordinator against a real open PR). I kept the fix without its event: an automatic pass that did no work now reports nothing at all. A manual reuse stays counted, since the user pressed the button. Reverting this would corrupt the P0 headline count, so I did not.

2. **`submitted` kept its richer payload** (`trigger`, `body_bytes`, `auto_inject`). Your list named events rather than keys, and these cost nothing on the wire while making the event answer "shallow auto approval or substantial manual changes-requested". Happy to strip them to `verdict`/`harness`/`duration_ms`/`posted_to_provider` if you would rather.

**Also left out, flagging rather than deciding:** `ao.review.rereview_requested` and `ao.review.comment_resolved` have been shipping **payload-stripped** since before this PR, because their names are missing from the allowlist. That is a two-line bug fix rather than a new metric, so it is not in this PR either. Say the word and it goes in.

One limitation of the P0 set to keep in mind: these are **change** events, not state events. A project that already has auto-review on will never emit a toggle event, so "what share of projects have it on today" needs a small daily gauge (counts only) rather than these.

Verification: touched Go packages pass with `-count=1`, `golangci-lint` 0 issues, 1927 renderer tests green across three consecutive runs, `tsc --noEmit` clean. The wire test still proves the three daemon events reach PostHog with their properties intact (verified it fails if an allowlist key is removed).

Pulkit7070 (commented):
This closes a real gap: an automatic review pass emitted nothing at trigger time, so auto and manual reviews were indistinguishable in every funnel. Routing both entry points through one `triggerWithSource` and tagging every event with `trigger` is the right shape, and the privacy posture is careful: `body_bytes` instead of the body, an allowlist plus a per-event sanitizer, and a wire-level test that asserts the review prose, repo, SHA, and PR URL never leave the process under any key.

A few things I checked that hold up:

- The manual path is behavior-preserving: `engine.Trigger` is exactly `TriggerWithSource(..., ReviewTriggerManual)`, so moving `Trigger` onto the shared path changes nothing but the added instrumentation.
- The auto no-op suppression (`SkipReason != "" || len(CreatedRuns) == 0`) is correct and worth having. A review sweeping every minute no longer books a trigger per sweep, so daemon event volume tracks real review passes rather than the sweep cadence, and a manual reuse still counts because the user pressed the button. Combined with the remote rate limit (5/min, 200/day per event name), this keeps the funnel from spamming the billed sink.
- `submitOne` reads `auto_inject` from the same `session` record it persists onto the run, and emits only on the real running to complete transition, so the event stays idempotent with the store.
- The project-settings changed flags read from the same `config` source the form state initializes from, so a save that leaves the review controls alone is not reported as a review decision.

One issue worth fixing:

`review_triggered.action` can be silently dropped. `reviewRunActionKind` returns one of four values including `reviewing`, and the emit site in `SessionInspector.tsx` passes its result straight through. But the sanitizer in `telemetry.ts` only allows `run`, `rerun`, `run_latest`. `reviewRunDisabled` does not gate on a review being in `running` status, so the manual trigger can fire while a pass is running, `reviewRunActionKind` returns `reviewing`, and the event then lands with no `action` at all, a hole in the exact adoption signal the event exists to measure. Either allow `reviewing` in the sanitizer or exclude that state at the emit site.

One judgment call for the author:

`review_auto_review_toggled` and `review_triggered` are emitted inside the mutation function before the `await`, so they count intent even when the PUT/POST then fails. That is consistent with treating the renderer events as intent and the daemon events as ground truth, and it is probably what you want. Flagging only so the renderer toggle count is read as intent, not as settings that took effect.

One more thing worth noting for whoever reads these dashboards: manual and auto share the single `ao.review.triggered` name and therefore its 200/day per-name cap on the billed sink, split only by the `trigger` property. On a very high-volume install the billed count can truncate (local SQLite still keeps every occurrence), so exact auto-vs-manual counts on a power user would draw from local, not PostHog.

Verified by checking out the branch: `go test ./internal/service/review/... ./internal/adapters/telemetry/...` and the renderer telemetry sanitizer tests pass.

Pritom14:
Thanks, this was a genuinely useful read. Fixed in `8c97276`.

**Your finding, confirmed, though by a different mechanism than described.** `reviewRunDisabled` is not what gates it: the button swaps handlers at `SessionInspector.tsx:2082` (`onClick={reviewRunning ? onCancel : onTrigger}`), and `reviewIsRunning` and `reviewRunActionKind`'s `reviewing` branch are the same predicate over the same query snapshot, so `onTrigger` normally cannot see a running state. It is still reachable, just through a narrower door: the emit reads `reviewsQuery.data` at call time (refetched every 2500ms while a pass runs) while `onClick` was bound at render time, so a click landing between a refetch and its re-render emits `reviewing`.

Worth fixing regardless, because the real defect is structural: a four-value enum feeding a three-value allowlist, which any future call site inherits. The sanitizer now accepts all four. Reporting that a trigger landed on a running pass beats dropping the property, and a test asserts every value `reviewRunActionKind` can return survives (verified it fails when `reviewing` is removed).

**Your finding led me to a bigger one.** The command palette POSTs to `/reviews/trigger` directly at `CommandPalette.tsx:333`, and my emit lived only in the inspector's mutation, so **every review started from the palette was missing from `review_triggered` entirely**. That is a documented manual on-ramp, so this was a real hole in the same adoption signal. The palette now emits the same event, with the offered action kind carried on the palette action (so it reports the enum, not the translated title), and a new `source` property (`inspector` | `command_palette`) keeps the two on-ramps separable.

Flagging `source` explicitly since it is a new key on a P0 event: without it the palette emit is a bare count, with it you learn which surface actually drives manual reviews. Happy to drop it if you would rather hold the line on scope.

**Your judgment call: agreed, and now written down.** Both renderer emit sites carry a comment saying that emitting before the request is deliberate, that they count intent, and that the daemon events are ground truth. It should not need re-deriving next time.

**Your rate-limit note: agreed, and added to the PR body** as a dashboard caveat. Folding `trigger` into the aggregation tier would collapse the very split this PR adds, so the cap stands and exact auto-against-manual counts for a power user come from local SQLite.

Verification: `tsc --noEmit` clean, **1932 renderer tests green across four consecutive full runs** (one earlier run showed 3 failures at 120s against a normal 28s, which was contention, not a real failure). Backend untouched by this commit but re-run anyway: touched packages pass with `-count=1`, `golangci-lint` 0 issues. Both new tests were checked to be load-bearing by breaking the code they guard and confirming they fail.

Pulkit7070 (approved):
LGTM. My earlier finding is resolved and verified: the telemetry sanitizer now accepts all four ReviewRunActionKind values including 'reviewing' (telemetry.ts), guarded by a load-bearing test that asserts each value survives sanitization. The follow-up also closed a real hole I did not spot originally, palette-triggered reviews were emitting nothing, now instrumented with a source property. Privacy posture (body_bytes, allowlist + sanitizer, wire-level leak test), auto no-op suppression, and store-idempotent submit all still hold. CI green.

PR #4256, opened 2026-08-22
feat(chat): retry a failed turn without retyping the prompt

### LINKED ISSUE #4215: feat(chat): retry a failed turn without retyping the prompt
## Problem

When a Chat turn fails after transient network loss, AO preserves the failed turn and its error correctly, but the user has no one-click way to submit the same prompt again. They must retype or manually copy the original message.

This is especially confusing when the recovery guidance says to send the message again.

**Source:** Follow-up from [#4212](https://github.com/Untrivial-ai/agent-orchestrator/issues/4212) | **Analyzed against:** `52fde0279`
**Confidence:** High
**Priority:** P2 — the core recovery path works, but avoidable re-entry friction follows a common transient failure.

## Current Behavior

- A failed turn is durable and its original human message remains in the timeline.
- The turn boundary displays `Failed` and the provider error.
- Rollback and edit/branch operations exist, but no operation means “dispatch this failed human prompt again as a new turn.”
- The composer retains attachments only when the initial send itself fails before acceptance; it does not restore content for an asynchronously failed provider turn.

## Proposed Behavior

Add a **Retry turn** action to eligible failed-turn boundaries that:

1. Reuses the durable original human message content and attachments.
2. Creates a new turn rather than mutating or relabeling the failed attempt.
3. Preserves both attempts in history and keeps provider-turn correlation explicit.
4. Uses the conversation's current next-turn settings unless product design deliberately chooses to preserve the failed attempt's settings.
5. Is disabled while another turn is running and reports daemon/provider refusals in place.
6. Is safe against double-clicks and uncertain delivery; it must not silently execute the work twice.

## Scope Notes

This should be a dedicated daemon/API operation, not renderer-side reconstruction of message text. The daemon owns ordering, attachments, delivery state, and idempotency. Reusing the existing edit/fork path would change conversation history semantics and is not equivalent to retrying after a recoverable failure.

## Acceptance Criteria

- Retry is offered only for an eligible failed human turn.
- Clicking once creates exactly one new queued/running turn with the same content.
- The original failed turn remains failed and visible.
- Attachments are reused or a clear validation error explains why they cannot be.
- Busy, unsupported-provider, stale-branch, and uncertain-delivery cases are tested.
- Renderer tests cover visible, hidden, disabled, pending, and error states.

## Related

- [#4212](https://github.com/Untrivial-ai/agent-orchestrator/issues/4212) — classifies transient Codex network failures and provides accurate recovery guidance.

vanshx999:
## Summary

Closes #4215.

Adds a **Retry** action to eligible failed human turns so a transient failure can be retried without retyping the original prompt. The retry is a new turn; the failed source stays unchanged and visible.

### Daemon and storage

- Adds `POST /api/v1/sessions/{sessionId}/conversation/turns/{turnId}/retry`.
- Loads text and structured prompt content from durable AO rows, never from a caller-supplied retry payload.
- Applies the current next-turn model, reasoning, and approval settings.
- Persists an explicit `retry_of_turn_id` relation on the new turn in the same transaction as its message. A unique partial index makes one retry attempt per source durable and idempotent; retrying a failed retry creates the next link in the chain.
- Uses the persisted relation for replay and UI correlation, so caller-controlled client message IDs cannot impersonate or consume a retry.
- Refuses unsafe or invalid cases with typed errors: busy/handoff, stale branch, non-human or non-failed source, unconfirmed provider delivery, malformed durable content, and content unsupported by the current provider.
- Adds explicit native resource-link capability negotiation. ACP preserves resource links; a provider that cannot preserve them receives a clear refusal instead of silently dropping the attachment.
- Keeps retry consumption visible even when the retry child is rolled back or moves outside the active branch by deriving `hasRetryAttempt` from all durable retry relations.
- Includes `recovered` in replay response contracts and regenerates sqlc, OpenAPI, and frontend API types.

### Renderer

- Shows Retry only on eligible failed human turns.
- Disables it while work is running or a retry request is pending.
- Displays retry refusals beside the affected turn.
- Hides consumed retry actions using daemon-owned `retryOfTurnId` / `hasRetryAttempt` facts, including inactive-branch and rolled-back retry attempts.

## Tests

Coverage includes exact prompt/attachment replay, current settings, busy/non-human/stale/uncertain/unsupported refusals, double-click and replay idempotency, caller-ID collision, recovered replay, retry chains, resource-link provider switching, rollback, inactive-branch consumption, HTTP contracts, wire-to-renderer mapping, renderer visible/hidden/disabled/pending/error states, and Playwright behavior.

## Verification

- `npm run lint` — full backend Go suite and golangci-lint: pass, 0 issues.
- Frontend typecheck and e2e typecheck: pass.
- Retry renderer + wire-mapping suites: 23 tests pass.
- Retry Playwright suite: 2 tests pass.
- Final Standards review: no findings.
- Final Spec review against #4215: no findings.


backend/internal/service/chat/controller.go
```diff
@@ -885,6 +894,102 @@ func (c *Controller) Send(ctx context.Context, msg ports.ChatUserMessage) (domai
...
+		if err := json.Unmarshal([]byte(prompt.DeliveryContentJSON), &content); err != nil {
+			return domain.ConversationTurn{}, fmt.Errorf("decode retried chat content: %w", err)
+		}
+	}
+
+	keyBase := prompt.ClientMessageID
+	if keyBase == "" {
+		keyBase = "turn-" + turnID
+	}
+	// Derive a free idempotency key by probing existing attempts, so a double-click
```
illegalcall: **Retry is not idempotent when delivery or the response is uncertain.**

This happens when:

1. This request creates `/retry/1` and the provider accepts the turn.
2. The acknowledgement/HTTP response is lost, or `SendTurn` returns an error after possible provider acceptance.
3. The client repeats the same Retry request.
4. This loop sees `/retry/1` and deliberately selects `/retry/2`, so the same prompt can execute twice.

The send path itself notes that an error does not prove the provider rejected the turn, so agent-side duplicate detection cannot provide the required guarantee.

A small fix is to derive exactly one key per source turn, for example `retry/<sourceTurnID>`, and look up that key before the busy check. Repeated calls for the same source then return the existing retry turn. If that retry genuinely fails and the user wants another attempt, they retry the failed retry turn, producing a safe chain (`A -> B -> C`). For a source turn with no confirmed provider turn ID, consider refusing with a typed uncertain-delivery error or requiring an explicit “Retry anyway” confirmation.

somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

vanshx999:
Rebased onto latest `main` (`c86b58d04`) to resolve the conflicts in `ChatWorkspace.tsx` and `SessionChatSurface.tsx`. Both sides kept:

- main's new Files-inspector wiring (`onOpenFiles`/`onOpenFile`) and the redesigned turn-end row (completed turns now show duration on the action row instead of a divider);
- this PR's retry props threaded through the same call sites unchanged.

One behavioral note from the merge: `TurnOutcome` no longer renders for completed turns (per main), which is exactly where retry is offered anyway — retry stays on the failed-turn boundary only. All chat tests pass after the merge (74 renderer + 5 service tests).

backend/internal/service/chat/controller.go
illegalcall: Rechecked against the rebased head `90a13bd8e`. This is still unresolved: `RetryTurn` still probes for the next unused `<source>/retry/<n>` key, so replaying the same uncertain Retry request after `/retry/1` was recorded is treated as a new `/retry/2` attempt and may execute the prompt again. The rebase changed surrounding frontend conflict resolution but did not change this algorithm.

The minimal fix remains: use one deterministic key per source turn (for example `retry/<sourceTurnID>`), look it up before the busy check, and return the existing retry turn for repeated calls. A deliberate further attempt can retry the failed retry turn, producing the chain `A -> B -> C`. Turns without confirmed provider acceptance should also be refused with a typed uncertain-delivery error or require explicit confirmation.

vanshx999:
@illegalcall  Fixed in db5ce924c — thanks for the careful re-check, the probe loop was indeed wrong: it converted a replayed request into a fresh `/retry/2` attempt, which is exactly the double-execution a retry must never risk.

**What changed (all three points from the review):**

1. **One deterministic key per source turn** — `retry/<sourceTurnID>`. No more attempt-number probing.
2. **Lookup before the busy check** — repeated calls (uncertain round trip, double-click, client retry) return the already-recorded attempt turn, before anything can refuse or dispatch. Verified by `TestRetryTurnReplaysReturnExistingAttempt`: replays return the same turn both while the attempt runs and after it settles, with the provider receiving exactly one retry send.
3. **Deliberate further attempts chain through the failed attempt** — retrying the failed retry turn B (which owns its own key `retry/<B>`) builds A → B → C from distinct sources; A can never mint a second child. Covered by `TestRetryChainThroughFailedAttempt` (provider sees exactly 3 sends of the original prompt, all three turns distinct).
4. **Uncertain-delivery refusal** — a source whose dispatch was never acknowledged (`ProviderTurnID == ""`, i.e. settled failed by AO's safe guess after `SendTurn` errored) is now refused with typed `ErrRetryDeliveryUncertain` / HTTP 409 `CHAT_RETRY_DELIVERY_UNCERTAIN`. Covered by `TestRetryTurnRefusesUnconfirmedDispatch`.

Backend tests: 7 retry tests passing; full suite green except the pre-existing, environment-dependent `crush` auth test that also fails on a clean checkout of current `main` (verified in a throwaway worktree — unrelated to this PR).

backend/internal/service/chat/controller.go
```diff
@@ -886,6 +902,117 @@ func (c *Controller) Send(ctx context.Context, msg ports.ChatUserMessage) (domai
...
+	defer c.sendMu.Unlock()
+	if c.handoffActive() {
+		return domain.ConversationTurn{}, ErrControllerHandoff
+	}
+
+	source, err := c.store.TurnByID(ctx, turnID)
+	if err != nil {
+		return domain.ConversationTurn{}, err
+	}
+	if source.ConversationID != c.conversation.ID {
```
illegalcall, line 978: **Reject retries from stale conversation branches.** Eligibility checks the conversation, state, rollback marker, and origin, but it never verifies that `source.BranchID` is on the active visible lineage/provider scope. A direct request using a failed turn ID from an inactive sibling can therefore copy that old prompt into the current active branch. Please enforce active-lineage eligibility—ideally in `SelectRetryableConversationPrompt` using the existing recursive `active_path` pattern—return a typed stale-branch conflict, and add the acceptance-test case required by #4215.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -1924,7 +1959,17 @@ const TurnGroup = memo(function TurnGroup({
 			    Interrupted/failed still get a labelled boundary so the reader can see
 			    how the turn ended. */}
 			{group.outcome && group.outcome.state !== "completed" ? (
-				<TurnOutcome state={group.outcome.state} error={group.outcome.error} />
+				<TurnOutcome
+					state={group.outcome.state}
+					error={group.outcome.error}
+					onRetry={
```
illegalcall: **Disable Retry while a provider turn is running.** This gate ignores the active turn derived from the snapshot, so Retry remains clickable and the daemon has to reject it as busy. The existing `busy` prop describes command/mutation activity rather than the whole provider-turn lifetime, so derive this from `activeTurn(snapshot)` (already available as `turn`), with a renderer test proving the failed-turn Retry button is disabled during another running turn.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -2204,6 +2211,19 @@ export function TurnOutcome({
 					{error}
 				</span>
 			) : null}
+			{onRetry ? (
+				<button
+					type="button"
+					onClick={onRetry}
+					disabled={retryPending}
+					aria-label="Retry this turn"
+					title={retryError ?? "Send this prompt again as a new turn"}
```
illegalcall: **Render retry refusals in place.** `retryError` is only assigned to `title`, which is invisible until hover and is not a reliable screen-reader announcement. Issue #4215 explicitly requires daemon/provider refusals to be reported in place. Render visible error/status text beside the affected turn (for example with `role="alert"`) and add the required error-state test.

illegalcall (commented):
The uncertain-delivery/idempotency issue is fixed. These are the remaining approval blockers from the latest review.

backend/internal/httpd/apispec/specgen/build.go
```diff
@@ -819,6 +819,18 @@ func shellTerminalOperations() []operation {
 				{http.StatusNotImplemented, envelope.APIError{}},
 			},
 		},
+		{
+			method: http.MethodPost, path: "/api/v1/sessions/{sessionId}/conversation/turns/{turnId}/retry", id: "retrySessionConversationTurn", tag: "conversations",
+			summary:    "Re-dispatch a failed turn's durable prompt as a new turn",
+			pathParams: []any{controllers.SessionIDParam{}, controllers.ConversationTurnIDParam{}},
+			resps: []respUnit{
+				{http.StatusAccepted, controllers.RetryTurnResponse{}},
```
illegalcall, line 828: **Add the required clean schema-name mapping.** `RetryTurnResponse` is a new named contract type, but `schemaNames` has no `ControllersRetryTurnResponse -> RetryTurnResponse` entry. The generated OpenAPI and TypeScript contracts consequently expose `ControllersRetryTurnResponse`. `AGENTS.md` requires a mapping for every new named API type; add it and regenerate both artifacts.

backend/internal/httpd/controllers/conversations.go
```diff
@@ -113,6 +115,50 @@ func (c *ConversationsController) editMessage(w http.ResponseWriter, r *http.Req
...
+}
+
+// writeConversationRetryError maps retry refusals to their HTTP meanings.
+func writeConversationRetryError(w http.ResponseWriter, r *http.Request, err error) {
+	switch {
+	case errors.Is(err, chatsvc.ErrTurnNotRetryable):
+		envelope.WriteAPIError(w, r, http.StatusConflict, "conflict",
+			"CHAT_RETRY_NOT_RETRYABLE",
+			"this turn cannot be retried: it is not a failed human turn in this conversation", nil)
+	case errors.Is(err, chatsvc.ErrRetryDeliveryUncertain):
```
illegalcall, line 147: **Add controller-route coverage for this new boundary.** The PR currently only extends service fakes; there are no HTTP assertions for the 202 response or the retry-specific 404/409 envelopes, including `CHAT_RETRY_DELIVERY_UNCERTAIN`. `AGENTS.md` requires tests for the changed user-visible boundary, happy path, and daemon error envelopes. Please add route tests that verify the status and error code for each typed refusal.

illegalcall:
Implemented the remaining retry-safety review fixes in `ae621c560`:

- rejects failed turns outside the active branch lineage with typed `CHAT_RETRY_STALE_BRANCH`;
- disables Retry while another turn is running;
- renders retry refusals visibly beside only the affected failed turn (`role="alert"`);
- hides Retry for non-human prompts;
- adds the clean `RetryTurnResponse` schema mapping and regenerates both API artifacts;
- adds HTTP happy-path/refusal coverage plus a browser E2E regression for the UI states.

### Before / after evidence

The exact same E2E spec was run against the prior PR head (`db5ce924c`) and this fix.

**1. A failed turn while another turn is running**

Before — Retry is still enabled, and the assertion fails:

Before: Retry remains enabled while another turn runs

After — Retry remains visible but is disabled:

After: Retry is disabled while another turn runs

**2. A retry request receives a typed daemon refusal**

Before — the refusal is only placed in the button tooltip; no visible error appears, and the `role="alert"` assertion fails:

Before: Retry refusal is not visible

After — the refusal is visible in place beside the affected turn:

After: Retry refusal is visible beside the failed turn

### Verification

- Prior head: `chat-retry.spec.ts` **2 failed** (enabled Retry + missing visible alert).
- Fixed head: `chat-retry.spec.ts` **2 passed**.
- Backend retry/service/controller tests pass, including stale-branch and every typed HTTP refusal.
- `npm run lint` passes (`0 issues`).
- Frontend unit tests, renderer typecheck, and E2E typecheck pass.
- Real Electron supervisor launched from this worktree in isolated dev mode; its daemon listened on `127.0.0.1:3002` and served the app successfully.

illegalcall (approved):

PR #4259, opened 2026-08-22
fix: tidy chat message chrome and turn duration display

Annieeeee11:
## Summary
- Move finished-turn duration onto the answer action row (next to copy/rollback) in Geist Sans; show whole seconds as `3s` instead of `3.0s`
- Drop the border on user message bubbles
- Make user edit/copy actions icon-only to match assistant actions

## Test plan
- [ ] Send a short chat turn and confirm duration shows as `Ns` when whole seconds, `N.Ms` when fractional
- [ ] Confirm duration sits on the assistant action row, not a Done divider
- [ ] Hover a user message: edit and copy are icon-only; bubble has no border
- [ ] Queued user messages still show the dashed border

## Screenshot 

image

### DIFF
--- a/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
+++ b/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
@@ -131,7 +131,10 @@ function shortenPaths(text: string): string {
 
 function formatDuration(ms: number): string {
 	if (ms < 1000) return `${ms}ms`;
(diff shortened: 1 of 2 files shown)


ronishrohan (approved):
Clean, well-scoped polish — duration on the action row reads better than a full Done divider, borderless bubbles are a nice touch, and icon-only user actions match the assistant pattern. durationMs threading is correct (no stale values on live turns). Ship it.

PR #4260, opened 2026-08-22
feat: summarize turn file changes with Files-panel review

Annieeeee11:
## Summary
- Add a turn-end “N Files Changed” card with Review + click-to-open in the Files inspector
- Style mid-turn Edited rows like Explore; single-file expands straight to the patch
- Path tooltips resolve turn-diff basenames via the turn’s file_change / cwd hints

## Test plan
- [ ] Complete a turn that edits files → card appears above rollback with basename + +/- stats
- [ ] Hover a file in the card → tooltip shows worktree path (`~/.ao/dev/data/worktrees/...`)
- [ ] Click a file → Files rail opens and expands that path; Review opens Files generally
- [ ] Single-file mid-turn Edited expands to the patch only (no duplicate file row)
- [ ] Multi-file Edited still expands to a list
- [ ] `cd frontend && npx vitest run src/renderer/components/chat/TurnChangedFiles.test.tsx src/renderer/components/chat/ChatProviderSignal.test.tsx`

## screenshot
image

image
image
image


frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -794,27 +901,13 @@ function FileChangeRow({ file }: { file: FileChangeFile }) {
...
-						<span aria-hidden="true" className="px-1 text-muted-foreground/40">
-							&rarr;
-						</span>
-					</>
-				) : null}
-				{shortenPaths(file.path)}
-			</span>
+			<span className="sr-only">{status.label}</span>
+			<span className="shrink-0 text-[11.5px] text-muted-foreground">Edited</span>
+			<FileLocationLabel path={file.path} oldPath={file.oldPath} />
```
ronishrohan, line 909: **Bug: hardcoded `"Edited"` discards the daemon's status-aware verb.**

The daemon normalizes provider change kinds into distinct statuses — `Created`, `Deleted`, `Renamed`, `Modified` — via `fileChangeVerb`. But `FileChangeRow` always renders `"Edited"` regardless of `file.status`.

Same issue in `splitSummary` (line 810) and the `singleEdit` branch (line 523): a newly created file shows as "Edited notes.md", a deleted file shows as "Edited gone.ts". Users can't distinguish file operations without expanding the patch.

Consider using `diffStatusMark[file.status].label` or a verb lookup here.

frontend/src/renderer/components/SessionFilesView.tsx
```diff
@@ -138,6 +143,31 @@ export function SessionFilesView({
...
 
+	// Chat (and similar) can ask the Files rail to open on a specific path. Match
+	// by exact path first, then by basename when the turn diff only carried a
+	// short name — either way expand that row and clear the request.
+	useEffect(() => {
+		if (!focusPath) return;
+		const match =
+			changedFiles.find((file) => file.path === focusPath) ??
+			changedFiles.find((file) => file.path === focusPath.replace(/^\.\//, "")) ??
+			changedFiles.find((file) => file.path.endsWith(`/${focusPath}`) || file.path.endsWith(focusPath));
```
ronishrohan: **Bug: `endsWith(focusPath)` without a path-separator guard matches partial filenames.**

The fallback `file.path.endsWith(focusPath)` has no separator check. If `focusPath` is `"index.ts"`, it matches `"src/reindex.ts"` before `"src/index.ts"`, expanding the wrong file in the Files rail.

The first alternative `` endsWith(`/${focusPath}`) `` is fine — the issue is the `||` fallback. Consider:
```ts
file.path.endsWith(`/${focusPath}`) || file.path === focusPath
```

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1836,68 +1949,202 @@ export function TurnChangedFiles({ diff, live }: { diff: TurnDiff; live?: boolea
...
+	cwd?: string;
+};
+
+function turnPathHints(items: ConversationItem[] | undefined): TurnPathHints {
+	const byBase = new Map<string, string>();
+	let cwd: string | undefined;
+	if (!items?.length) return { byBase, cwd };
+
+	for (const item of items) {
+		if (item.kind !== "activity") continue;
```
ronishrohan, line 2133: **Bug: basename collision silently overwrites Map entries — tooltip shows wrong path.**

`turnPathHints` keys by `fileBasename(file.path)`. When a turn edits both `src/utils/index.ts` and `lib/utils/index.ts`, `byBase.set("index.ts", ...)` overwrites the first entry. The tooltip then shows the wrong absolute path for one of the two files.

Consider storing an array per basename key, or setting the value to a sentinel (like `undefined`) on collision so the tooltip falls back to the raw path rather than showing the wrong one.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -545,34 +564,112 @@ function GenericActivityRow({ activity }: { activity: ConversationActivity }) {
...
-					) : null}
-					{detail?.terminalInput ? (
-						<TerminalInput
-							text={detail.terminalInput}
-							truncated={detail.terminalInputTruncated}
-						/>
-					) : null}
-					{detail?.output ? <CommandOutput activity={activity} /> : null}
+				compactSummary && (detail?.command || detail?.output || detail?.terminalInput) ? (
+					<CommandExploreBody activity={activity} />
```
ronishrohan, line 573: **Potential bug: routing guard checks `detail.command` instead of `activityKind`.**

`compactSummary` is true for both `"command"` and `"file_change"` activity kinds. This guard:
```ts
compactSummary && (detail?.command || detail?.output || detail?.terminalInput)
```
routes to `CommandExploreBody` without checking `activityKind === "command"`. If a `file_change` activity ever carries a `command` field (structurally valid per the flat detail type), the file patches are silently hidden behind the command card.

Consider:
```ts
activity.activityKind === "command" && (detail?.command || ...)
```

frontend/src/renderer/components/chat/activity-command.ts
```diff
@@ -5,7 +5,7 @@ export type CommandCategory = "read" | "search" | "vcs" | "run";
 
 /** Shared chrome for one command summary and a collapsed run of commands. */
 export const ACTIVITY_SUMMARY_BUTTON_CLASS =
-	"group/run flex w-full items-center gap-1.5 rounded-sm py-0.5 pr-1 text-left outline-none transition-colors hover:bg-interactive-hover focus-visible:outline-none";
+	"group/run flex w-full items-center gap-1.5 rounded-sm py-0.5 pr-1 text-left outline-none focus-visible:outline-none";
```
ronishrohan: **Regression: hover affordance removed from all compact rows.**

`transition-colors hover:bg-interactive-hover` was removed from `ACTIVITY_SUMMARY_BUTTON_CLASS`. This class is shared across command rows, file-change rows, and collapsed `ActivityRun` groups. The only remaining hover signal is the chevron shifting from 40% to full opacity — a subtle 12px change most users won't notice.

Compact rows now look inert; users may not realize they're expandable. Was this intentional?

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -545,34 +564,112 @@ function GenericActivityRow({ activity }: { activity: ConversationActivity }) {
...
+					/>
+				</div>
+			) : null}
+			{detail?.output ? <CommandOutput activity={activity} embedded /> : null}
 		</div>
 	);
 }
 
+function commandBinaryLabel(command: string): string | undefined {
+	const trimmed = command.trim();
```
ronishrohan: **Duplication: `commandBinaryLabel` reimplements `firstWord` + `binaryName` from `activity-command.ts`.**

The trim/indexOf/slice/lastIndexOf/slice extraction here is identical to `firstWord` (line 66) and `binaryName` (line 73) in `activity-command.ts`. Two copies means a future fix to command parsing must be applied in both places. Consider importing/reusing the existing helpers.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -644,6 +646,19 @@ export function SessionView({ sessionId }: SessionViewProps) {
 		setInspectorOpenForSession(sessionId, true);
 	}, [sessionId, setInspectorOpenForSession, setInspectorViewForSession]);
 
+	const handleOpenFile = useCallback(
```
ronishrohan, line 649: **Simplification: `handleOpenFile` duplicates the body of `handleOpenFiles`.**

Both callbacks set the same 4 state values (`setBrowserPopOutState`, `setFilesPoppedOut`, `setInspectorViewForSession`, `setInspectorOpenForSession`). `handleOpenFile` adds `setFilesFocusPath`. If the "open files panel" sequence is updated in one but not the other, they'll diverge.

Consider having `handleOpenFile` call `handleOpenFiles()` first, then set the focus path.

ronishrohan (commented):
Review: 8 inline findings (3 correctness bugs, 1 plausible correctness issue, 1 regression, 3 simplification/cleanup). See inline comments.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -1836,68 +1949,202 @@ export function TurnChangedFiles({ diff, live }: { diff: TurnDiff; live?: boolea
...
+				byBase.set(fileBasename(file.oldPath), file.oldPath);
+			}
+		}
+	}
+	return { byBase, cwd };
+}
+
+function looksAbsolutePath(path: string): boolean {
+	return path.startsWith("/") || path.startsWith("~") || /^[A-Za-z]:[\\/]/.test(path);
+}
```
ronishrohan, line 2146: **Dead code: first branch of `resolveTurnFilePath` Map lookup is unreachable for relative paths.**

`turnPathHints` only stores keys that are basenames or absolute paths. The early return at the top of this function catches absolute paths before the Map lookup. So `hints.byBase.get(path)` for a relative path like `"src/foo.ts"` will never match — the fallback `hints.byBase.get(fileBasename(path))` already covers the basename case. This branch is unreachable and misleads readers.

ronishrohan: **Closing this — the finding is wrong.**

On re-read: `hints.byBase.get(path)` handles the *primary* use case. Turn diffs commonly carry bare basenames like `"random_words_1.txt"`, and `byBase` stores those as keys (line 2110). So `get(path)` matches directly for basenames. The `get(fileBasename(path))` fallback handles relative paths like `"src/foo.ts"` by extracting the basename. Both branches serve a purpose.

ronishrohan (approved):
All 8 review comments addressed. Tests pass (60/60). LGTM.

PR #4266, opened 2026-08-23
fix(session): do not apply a role override's model across harnesses

execsumo:
## The bug

A project role override pins a harness and carries an agent config chosen **for that harness**. `effectiveHarness` already honors an explicit spawn harness, but `effectiveAgentConfig` applied the override's agent config unconditionally.

So in a project whose worker role pins `codex` with a codex model id, an explicit `claude-code` spawn inherited that model id and **launched broken**. Found while exercising two concurrent workers on different harnesses in one project.

## The fix

`Model` and `Mode` are applied only when the resolved harness matches the override's pinned harness. An override with no pinned harness keeps its apply-unconditionally behavior, so existing single-harness projects are unaffected.

Every caller now passes the harness actually resolved for that session — the explicit spawn harness, the durable `rec.Harness` on restore and interface transitions, and the target harness on an agent switch.

## Why `Permissions` stays outside the gate

`PermissionMode` is an abstract enum that each adapter maps onto its own agent's native approval flags, so it is not harness-specific. Skipping it on a mismatch would silently substitute the project baseline for the role's permission — and one direction of that substitution, a role pinning `default` over a baseline of `bypass-permissions`, is a **privilege escalation**.

## On the test

The fixture gives the worker override a permission that differs from the base. That detail is load-bearing: with it unset, every permission assertion passes trivially by reading the base value back, whether or not the override was applied at all.

Verified by reverting the function and confirming the test fails; `go build`, `go vet` and `go test ./internal/session_manager/...` (including `-race`) are clean.

### DIFF
--- a/backend/internal/session_manager/agent_switching.go
+++ b/backend/internal/session_manager/agent_switching.go
@@ -1139,11 +1139,7 @@ func (m *Manager) prepareTargetActivation(ctx context.Context, store ports.Agent
 	if err != nil {
 		return preparedTargetActivation{}, fmt.Errorf("system prompt file: %w", err)
 	}
-	config := effectiveAgentConfig(rec.Kind, project.Config)
-	if roleOverride(rec.Kind, project.Config).Harness != harness {
-		config.Model = ""
-		config.Mode = ""
-	}
+	config := effectiveAgentConfig(harness, rec.Kind, project.Config)
 	if model := strings.TrimSpace(modelOverride); model != "" {
 		config.Model = model
 	}
--- a/backend/internal/session_manager/agent_switching_chat.go
+++ b/backend/internal/session_manager/agent_switching_chat.go
@@ -143,11 +143,7 @@ func (m *Manager) executeChatAgentSwitch(
 		return result, fmt.Errorf("switch Chat agent %s: system prompt: %w", id, err)
 	}
 	systemPrompt = appendAgentContinuationProtocol(systemPrompt)
-	agentConfig := effectiveAgentConfig(rec.Kind, project.Config)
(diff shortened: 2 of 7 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

axisrow:
Hit this exact bug independently while investigating something unrelated. Repro on current `main`:

- Project's worker role is pinned to `codex` with `agentConfig.model = <a codex-family model>`.
- Ran `ao spawn --harness claude-code ...` (explicit harness override, project itself otherwise unrelated to claude-code for the worker role).
- Session launched as Claude Code but with the *codex* model id inherited from the role override — Claude Code immediately reported: `There's an issue with the selected model (<model>). It may not exist or you may not have access to it. Run /model to pick a different model.`
- Session went straight to idle without doing anything useful.

Matches the description here exactly — `effectiveAgentConfig` applying the override's model unconditionally even though the resolved harness (explicit spawn flag) differs from the override's pinned harness. Confirms this is still reproducible on current main and would be a good fix to land.

i-trytoohard:
@execsumo Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@execsumo Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

execsumo:
Closing as superseded by #5129, which moved the same harness gate into effectiveAgentConfig so every spawn path shares it. Thanks!

PR #4267, opened 2026-08-23
feat(notifications): tell the user when a TUI worker finishes its turn

execsumo:
## The gap

A TUI worker that finishes its task settles in `ActivityIdle` and **no notification fires**, so nothing tells the user the work is done.

## Why not just map end-of-turn to `waiting_input`

That looks like the one-line fix and it is wrong. `ActivityState.NeedsInput()` gates the automation layer:

- `lifecycle/reactions.go` suppresses `ready_to_merge` while it is true — a finished worker that just opened a PR would never raise `ready_to_merge` until the user messaged it again;
- `cannotNudge` suppresses automated nudges the same way.

Both `claudecode` and `codex` map end-of-turn to idle by documented decision. That stays untouched.

## What this adds

A distinct notification kind, `turn_complete`, emitted on the transition into idle. The predicate is deliberately narrow:

- **`prev` is enumerated** (`active` / `waiting_input` / `blocked`) rather than "not idle". `claudecode` maps *both* `stop` and `Notification(idle_prompt)` to idle, so `idle → idle` must not fire twice; `codex` reaches idle from `waiting_input` after a permission round-trip, which an active-only check would miss.
- **`Kind == worker` and `Mode == TUI`.** Chat sessions are also worker-kind, so gating on kind alone would fire on every chat exchange.
- the session is not terminated.

`turn_complete` needs resolution: resolved when the session leaves idle (the user gave it more work) or terminates, mirroring `needsInputResolutions` from all three lifecycle call sites, plus a reconcile query so a daemon crash between the two writes cannot strand one unresolved.

## Two surfaces worth a reviewer's attention

- The `type IN (...)` lists in `storage/sqlite/queries/notifications.sql` are a **hardcoded SQL mirror** of `NotificationType.NeedsResolution()`. Miss them and the notification is created but never appears in the unresolved list or count.
- `notify/enrich.go` rejects any non-`needs_input` intent without a PR URL, and its title/body switches fall through to a `"Notification"` default that still passes `NotificationRecord.Validate()` — so a missing case ships empty copy with a green suite.

## Renderer

Reuses the `needs_input` toast suppression, so no toast fires while the user is already watching that session, and the row routes to the session on click. Copy goes through `t()` with the English string added to every locale catalog, which is the existing pattern for untranslated keys.

## Verification

`go build`, `go vet`, `go test` and `-race` clean on the changed packages; renderer vitest 156 files / 2274 passed; `npm run api` re-run leaves the tree clean, so the committed `openapi.yaml` and `schema.ts` are genuinely generated.

**Not yet observed in a live session** — this is test-verified only. The thing to watch on first real use is volume: `stop` maps to idle, so an interrupted turn notifies too, at one notification per turn. If the list proves noisy, the natural next step is to dedupe on an existing unresolved `turn_complete` for the session.

## Migration: 0107

This PR includes `backend/internal/storage/sqlite/migrations/0107_notification_turn_complete.sql`. Adding the `turn_complete` type without widening the `notifications.type` CHECK constraint caused every insert of that type to fail with `CHECK constraint failed` — silently, because notification writes are best-effort by design (a failed notification must never fail the lifecycle write that produced it), so the error only surfaced as a warning log. This was found by live observation, not by tests; live-confirmed firing after the fix (one notification per completed turn).

SQLite cannot alter a CHECK in place, so 0107 uses the standard table-rebuild pattern: create `notifications_new` with the widened constraint, copy all rows, drop the old table, rename. The four indexes from migrations 0031/0041 are recreated verbatim (`DROP TABLE` takes them with it). The down migration deletes `turn_complete` rows first — they are advisory notifications, not durable state — before restoring the narrower constraint.

0107 claims version **107**, not the next free number at branch time, because upstream has since shipped 0104–0106 while this branch predates them; claiming 107 keeps the rebase onto `upstream/main` collision-free. The claim is registered in the append-only ledger in the same change, per `TestMigrationVersionLedger`.

### DIFF
--- a/backend/internal/domain/notification.go
+++ b/backend/internal/domain/notification.go
@@ -11,6 +11,8 @@ type NotificationType string
 const (
 	// NotificationNeedsInput means an agent session is waiting for user input.
 	NotificationNeedsInput NotificationType = "needs_input"
+	// NotificationTurnComplete means a TUI worker finished its turn and is waiting at an empty prompt.
+	NotificationTurnComplete NotificationType = "turn_complete"
 	// NotificationReadyToMerge means a PR has no known merge blockers.
 	NotificationReadyToMerge NotificationType = "ready_to_merge"
 	// NotificationPRMerged means a tracked PR was merged.
@@ -22,7 +24,7 @@ const (
 // Valid reports whether t is one of the v1 notification kinds.
 func (t NotificationType) Valid() bool {
 	switch t {
-	case NotificationNeedsInput, NotificationReadyToMerge, NotificationPRMerged, NotificationPRClosedUnmerged:
+	case NotificationNeedsInput, NotificationTurnComplete, NotificationReadyToMerge, NotificationPRMerged, NotificationPRClosedUnmerged:
 		return true
 	default:
 		return false
@@ -30,11 +32,12 @@ func (t NotificationType) Valid() bool {
 }
 
 // NeedsResolution reports whether t describes an issue that stays open until
-// something changes (an agent waiting on the user, a PR waiting on a merge).
+// something changes (an agent waiting on the user or at an empty prompt, or a
+// PR waiting on a merge).
 // Terminal facts — a PR that merged or closed — describe something that already
 // happened, so they are surfaced once as unseen and never held as unresolved.
 func (t NotificationType) NeedsResolution() bool {
-	return t == NotificationNeedsInput || t == NotificationReadyToMerge
+	return t == NotificationNeedsInput || t == NotificationTurnComplete || t == NotificationReadyToMerge
 }
 
 // NotificationStatus is the seen state for a stored notification. The stored
--- a/backend/internal/domain/notification_test.go
+++ b/backend/internal/domain/notification_test.go
@@ -1,0 +1,29 @@
+package domain
+
+import "testing"
+
+func TestNotificationTypeValidAndNeedsResolution(t *testing.T) {
+	tests := []struct {
+		typ             NotificationType
+		valid           bool
+		needsResolution bool
+	}{
+		{NotificationNeedsInput, true, true},
+		{NotificationTurnComplete, true, true},
+		{NotificationReadyToMerge, true, true},
+		{NotificationPRMerged, true, false},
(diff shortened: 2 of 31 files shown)


somewherelostt:

- @somewherelostt
- @neversettle17-101

Join the session here:
https://discord.gg/H6ZDcUXmq

execsumo:
## Live use surfaced a scope gap: chat-mode workers never fire `turn_complete`

We deployed this branch and drove real work through it. A delegated **codex worker in chat mode** completed its turn — two commits on its branch — and no notification fired. Checked the store first: zero rows written, so this isn't the resolution path or the CHECK bug; the emission guard excluded it.

The guard (`lifecycle/manager.go`, the activity-intent branch) requires `Kind == worker && Mode == TUI`. That scope is deliberate and documented, but from the operator's seat it splits the mental model: a delegated worker's harness choice (TUI vs chat driver) is invisible to the person waiting on the result, and today most of our delegations land in chat mode. The result is that the feature is silent exactly where we need it.

### Proposal

Include chat-mode workers: drop the mode term from the predicate so it reads "any non-terminated worker transitioning active/waiting-input/blocked → idle". The resolution logic (`turnCompleteResolutions`) is already mode-agnostic, so leaving idle resolves correctly either way.

### The counterargument, acknowledged

TUI-only was chosen for volume: one notification per turn, and an interrupted turn notifies too (`stop` maps to idle). Chat workers may turn over faster than TUI ones, so widening could make the list noisy sooner. Two mitigations if that happens:

1. The natural dedupe already sketched in the Verification section (collapse onto an existing unresolved `turn_complete` per session) covers chat mode identically.
2. If needed, chat mode could gate behind config before defaulting on — but we'd rather collect noise data than pre-optimize.

### Ask

Would you rather extend this PR, or keep it TUI-only and take chat mode as an immediately-following change? We have the change running locally either way; happy to push it here if you prefer the PR route.

i-trytoohard:

i-trytoohard:
@execsumo Closing as stale: no activity from you for 30 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

Issue #3640, opened 2026-08-05
bug(terminal): clicking in the input box doesn't move the caret — forceSelectionMode suppresses all mouse reporting to the PTY


homeshrawat:
## Summary

Clicking inside the agent's input box does not move the text caret. To edit something earlier in
a line you have already typed, you must arrow left/right — clicking where you want to edit does
nothing.

Root cause: AO **unconditionally suppresses all mouse button reporting to the PTY**, so
mouse-aware agent TUIs (cursor-agent, claude-code) never receive the click that would place the
caret.

## Root cause

`frontend/src/renderer/components/XtermTerminal.tsx:243-251`, called at line 400:

```ts
function forceSelectionMode(term: Terminal): void {
	const internal = term as XtermInternal;
	const selectionService = internal._core?._selectionService;
	const element = internal._core?.element;
	if (!selectionService || !element) return;
	selectionService.shouldForceSelection = () => true;
	selectionService.enable();
	element.classList.remove("enable-mouse-events");
}
```

`shouldForceSelection = () => true` makes xterm treat **every** mouse press as a text-selection
drag. Removing `enable-mouse-events` stops xterm's `MouseService` / `CoreMouseService` from
encoding button reports at all. So even when the TUI enables mouse tracking (DECSET
1000/1002/1006), press/drag/release are never encoded and never written to the PTY.

The only mouse input ever forwarded is the **wheel**, hand-rolled as SGR reports at
`XtermTerminal.tsx:220-241` (`sgrWheelReport`, buttons 64/65) and dispatched in the wheel handler
at `:700-750`, which correctly branches on `term.modes.mouseTrackingMode`. There is no button
equivalent, and no click-to-position logic anywhere.

## Scope — one of these two cases is a bug, the other is not

1. **Plain shell / readline (bash, zsh)** — not a bug. Terminals do not move the shell cursor on
   click; that is expected behavior.
2. **Mouse-aware TUIs (cursor-agent, claude-code, tmux with `mouse on`)** — this **is** a bug.
   Those TUIs enable mouse tracking specifically so a click can place the caret, and
   `forceSelectionMode` overrides them unconditionally. This is the reported case: clicking into
   already-typed text in the agent's input box to edit it.

## Regression risk — please read before fixing

`forceSelectionMode` was introduced by commit `7ba860741`, *"fix: add terminal controls and
restore copy/scroll (#372)"*. It exists to make copy/selection work, and it bought that by
[...]

Issue #1221, opened 2026-04-13
Terminal text selection offset on right side due to overlay scrollbar mismatch


i-trytoohard:
## Bug

Mouse text selection in the terminal is offset on the right side — the drag/selection pointer is shifted ~14px to the right of the actual mouse cursor. Left-side selection works fine. Affects macOS desktop (Retina), both fullscreen and normal modes.

**Discord thread:** https://discord.com/channels/1489612265027014736/1491735678156013588/1493159493247176734
**Reported by:** @prateek
**Analyzed against:** `07f01ef6`

## Reproduction

1. Open any session detail page on macOS desktop (Retina display)
2. Try to drag-select text on the left side of the terminal → works correctly
3. Try to drag-select text on the right side of the terminal → selection pointer is offset ~14px to the right of the actual mouse position

## Root Cause

**Mismatch between CSS scrollbar override and xterm.js column calculation.**

The chain:

1. **FitAddon calculates column count** in `proposeDimensions()` (`@xterm/addon-fit@0.11.0`):
   ```js
   const scrollbarWidth = scrollback === 0 ? 0 : overviewRuler?.width || 14;
   const width = parentWidth - paddingLeft - paddingRight - scrollbarWidth;
   cols = Math.floor(width / cellWidth);
   ```
   Since `scrollback: 10000` and no `overviewRuler` is configured, `scrollbarWidth = 14`.

2. **CSS overrides the viewport scrollbar** in `globals.css:478`:
   ```css
   .xterm .xterm-viewport {
     overflow-y: overlay !important;
   }
   ```
   `overflow-y: overlay` renders the scrollbar ON TOP of content — it does NOT take up layout space.

3. **Result:** FitAddon subtracts 14px for the scrollbar, but the scrollbar doesn't actually consume that space. The terminal renders fewer columns than the available width supports. The ~14px gap on the right side causes mouse coordinate mapping to be off — the farther right you go, the more the cumulative column offset pushes the selection pointer away from the actual cursor.

**Files involved:**
- `packages/web/src/app/globals.css:478` — `overflow-y: overlay !important` override
- `packages/web/src/components/DirectTerminal.tsx:218` — `scrollback: 10000` (triggers scrollbar width subtraction)
- `@xterm/addon-fit@0.11.0` — hardcoded `scrollbarWidth = 14` fallback

## Fix

[...]

PR #4271, opened 2026-08-23
fix(terminal): auto-scroll the pane while a selection drag is held

### LINKED ISSUE #4265: bug(terminal): selection drag does not auto-scroll — alt-buffer/tmux panes never get the drag-scroll the wheel already routes to them
## Bug

Dragging a text selection past the top or bottom edge of an AO terminal does not auto-scroll the pane. The selection stays active and follows the pointer inside the viewport, but the terminal never scrolls, so a selection can never grow past what is already on screen. Ordinary selection and ordinary wheel scrolling both work.

**Reported by:** @whoisasx (in-app report) | **Analyzed against:** `85fab5e29ff718835a098e55ba598f0a2bc91e20` (`origin/main`)
**AO version:** `ao version dev` (worktree build) · desktop app running from `/Applications/Agent Orchestrator.app`
**OS:** macOS 26.6 (25G70), arm64 · Darwin 25.6.0 · daemon verified listening on `127.0.0.1:3001`
**Confidence:** High — traced to specific lines, reproduced in a real Chromium against the shipped xterm build, and confirmed against real `tmux attach` bytes.

## Reproduction

1. Open any AO session terminal (or shell terminal) with more transcript than fits the pane.
2. Press and hold the left mouse button on a line of terminal text.
3. Drag the pointer above the top edge (or below the bottom edge) of the terminal and hold it there.
4. **Expected:** the pane auto-scrolls and the selection keeps extending.
   **Actual:** nothing scrolls. The highlight stops at the viewport edge.

Wheel scrolling in the same pane works, and click-drag selection inside the pane works.

## Root Cause

`XtermTerminal.tsx` routes the **wheel** away from xterm's local viewport for panes whose scrollback lives in the pane, but nothing does the equivalent for xterm's **selection drag**, so the drag falls through to a scroll path that is a no-op for every AO terminal.

**1. AO terminals are always alternate-screen + mouse-tracking panes.**

`tmux.Runtime.Attach` spawns `tmux -u -T RGB attach-session -t <id>` on a PTY with `TERM=xterm-256color` (`backend/internal/adapters/runtime/tmux/tmux.go:702-756`), and the session is created with `mouse on` (`backend/internal/adapters/runtime/tmux/commands.go:41`). Capturing that exact attach stream from a real PTY shows tmux immediately driving the client into the alternate screen with SGR mouse tracking on:

```
head: b'\x1b[?1049h\x1b[22;0;0t\x1b[?1h\x1b=\x1b[H\x1b[2J...'
  ALT SCREEN (smcup 1049h): YES
  mouse X11 1000h: YES
  mouse btn-event 1002h: YES
  SGR mouse 1006h: YES
```

Both the agent session terminal and the standalone shell terminals use this adapter (`backend/internal/daemon/daemon.go:300` passes the same `runtimeAdapter` into `startShellTerminals`), so on macOS and Linux every AO terminal is an alt-buffer pane whose scrollback belongs to tmux, not to xterm.

**2. xterm's selection drag-scroll only moves xterm's own scrollback.**

`SelectionService._dragScroll()` (`@xterm/xterm@5.5.0` `src/browser/services/SelectionService.ts:665-690`) fires `onRequestScrollLines`, which `Terminal.ts:526` forwards to `Terminal.scrollLines(amount, suppressScrollEvent)` with the default `ScrollSource.TERMINAL`. That reaches `BufferService.scrollLines`, which begins:

```ts
if (disp < 0) {
  if (buffer.ydisp === 0) {
    return;                      // <-- alt buffer: ydisp is permanently 0
  }
  ...
}
buffer.ydisp = Math.max(Math.min(buffer.ydisp + disp, buffer.ybase), 0);
if (oldYdisp === buffer.ydisp) {
  return;                        // <-- alt buffer: ybase is permanently 0
}
```

In the alternate buffer `ybase === ydisp === 0` forever, so the drag-scroll is a guaranteed no-op. tmux never receives anything, because AO does not forward the drag to the PTY (`forceSelectionMode`, `frontend/src/renderer/components/XtermTerminal.tsx:260-268` — the same patch behind [#3640](https://github.com/Untrivial-ai/agent-orchestrator/issues/3640)).

**3. The wheel already solves this; the drag was never given the same route.**

`frontend/src/renderer/components/XtermTerminal.tsx:763-805` (`attachCustomWheelEventHandler`) explicitly branches on pane kind: keyboard-scroll panes get `PageUp`/`PageDown`, normal-buffer panes with mouse tracking off get a local `term.scrollLines(lines)`, and mouse-tracking panes get SGR wheel reports that tmux acts on via copy-mode. That is exactly why "normal scrolling works". There is no matching branch anywhere for the selection drag, so it silently keeps using path (2).

### Reproduced in a real Chromium

A harness loading the shipped `@xterm/xterm@5.5.0` build, applying AO's `forceSelectionMode` patch, then synthesizing `mousedown` inside the screen box and `mousemove` 40px above it (held for 600ms, i.e. ~12 of xterm's 50ms drag-scroll ticks):

| case | buffer / mouse tracking | `viewportY` before → after | auto-scrolled |
|---|---|---|---|
| A — plain shell / codex | `normal`, tracking `none` | 377 → 245 | **YES** |
| B — `tmux attach` (every AO terminal on macOS/Linux) | `alternate`, tracking `drag` | 0 → 0 | **NO** |

Case B is set up with the exact sequence tmux sends on attach (`\x1b[?1049h`, `\x1b[?1000h`, `\x1b[?1002h`, `\x1b[?1006h`). Case A proves AO's `forceSelectionMode` patch, the hidden-scrollbar CSS (`frontend/src/renderer/styles.css:1449-1464`) and `removeHiddenScrollbarReservation` do **not** break xterm's native drag-scroll — the failure is entirely the alt-buffer/mouse-tracking routing gap.

## Platform Impact

- **macOS / Linux:** always broken. The runtime adapter is tmux, so both agent session terminals and shell terminals are alt-buffer + mouse-tracking panes.
- **Windows:** the runtime adapter is conpty, not tmux (`backend/internal/adapters/runtime/conpty`). A plain shell pane there stays in the normal buffer with mouse tracking off, so xterm's native drag-scroll already works (case A). An agent TUI that enters the alternate screen itself, or that enables mouse tracking, hits the same bug.
- **Not Electron-specific.** The mechanism is pure xterm + buffer state and reproduces in stock Chromium outside Electron; no pointer-capture, drag-region or CSS involvement.

## Impact

- A terminal selection can never exceed one screen of text. Copying a full agent response, a long stack trace, or a multi-screen diff out of the terminal is impossible in one gesture.
- Workaround exists but is clumsy: wheel-scroll to the region first, then select only what is visible, repeat.

## Proposed Fix

Mirror the wheel handler for the selection drag in `frontend/src/renderer/components/XtermTerminal.tsx`:

1. Extract the wheel handler's pane-kind decision into a single helper (`local` / `mouseReport` / `pageKeys`) so the two scroll paths cannot drift.
2. While a left-button selection drag is live and the pointer is outside the `.xterm-screen` box, emit the same bytes the wheel would emit for that many lines, on xterm's own 50ms drag-scroll cadence. Skip it entirely for `local` panes — xterm's native drag-scroll already handles those and emitting would double-scroll them.
3. Track the drag with capture-phase `mousemove`/`mouseup` listeners: xterm's own document-level `mousemove` listener calls `stopImmediatePropagation()` while a drag is live (`SelectionService.ts:592-601`), which would drop a bubble-phase listener registered after it.

xterm keeps owning the selection model — its `_dragScroll` re-pins `selectionEnd` to the viewport edge on every tick — so the fix only has to supply the scroll the pane never received.

### Known limitation this does not remove

Over a tmux pane the selection still cannot accumulate *beyond one screen*: xterm's selection anchors are buffer rows, and tmux repaints the alternate screen in place, so scrolling changes the text under a fixed anchor rather than adding rows above it. After the fix the pane scrolls under a live drag and the highlight stays WYSIWYG with what is copied, which is as far as xterm-level selection can go while the scrollback lives in tmux. Genuinely unbounded selection needs the pane's native transcript — that is [#3791](https://github.com/Untrivial-ai/agent-orchestrator/issues/3791).

## Test Plan

- `frontend/src/renderer/components/XtermTerminal.test.tsx`: a selection drag past the edge emits SGR wheel reports on a mouse-tracking pane; emits nothing (and does not double-scroll) on a normal-buffer pane; emits page keys on a keyboard-scroll pane; emits nothing while the pointer stays inside the screen box; stops on `mouseup`; direction follows above vs. below the box.
- Existing wheel tests must stay green unchanged — they pin the exact byte sequences the shared helper now produces.
- `npm run frontend:typecheck`, `cd frontend && npm test`.
- Manual: drag-select past the top of a real tmux-backed AO terminal and confirm the pane scrolls; confirm ordinary click-selection, wheel scrolling, right-click menu, and typing are unaffected.

## Related

- [#3640](https://github.com/Untrivial-ai/agent-orchestrator/issues/3640) — `forceSelectionMode` suppresses mouse reporting to the PTY. Same `forceSelectionMode` patch (`XtermTerminal.tsx:260-268`) is what makes AO's local selection possible over a mouse-tracking pane in the first place, so any change there interacts with this.
- [#3791](https://github.com/Untrivial-ai/agent-orchestrator/issues/3791) — copy full prompt/response from agent TUIs by reading the native transcript. The proper long-term answer to "select more than one screen".
- [#1221](https://github.com/Untrivial-ai/agent-orchestrator/issues/1221) — terminal selection offset from the overlay scrollbar mismatch (same selection surface).
- [#1280](https://github.com/Untrivial-ai/agent-orchestrator/issues/1280) — viewport snapped to bottom on drag-start (closed; same drag-selection surface).

whoisasx:
Fixes #4265

## Summary

Dragging a text selection past the top/bottom edge of an AO terminal did not auto-scroll the pane, so a selection could never grow past what was already on screen. Ordinary selection and ordinary wheel scrolling both worked.

xterm's built-in selection drag-scroll (`SelectionService._dragScroll`) fires `onRequestScrollLines` → `Terminal.scrollLines` → `BufferService.scrollLines`, which only moves lines xterm holds in its **own** scrollback:

```ts
if (disp < 0) { if (buffer.ydisp === 0) return; }
buffer.ydisp = Math.max(Math.min(buffer.ydisp + disp, buffer.ybase), 0);
if (oldYdisp === buffer.ydisp) return;
```

Every AO terminal on macOS/Linux is a `tmux attach-session` client (`backend/internal/adapters/runtime/tmux/tmux.go:702`, and `backend/internal/daemon/daemon.go:300` gives shell terminals the same adapter). Capturing that attach stream from a real PTY shows tmux immediately driving the client into the alternate screen with SGR mouse tracking on:

```
b'\x1b[?1049h\x1b[22;0;0t\x1b[?1h\x1b=\x1b[H\x1b[2J...'
  ALT SCREEN (1049h): YES   mouse 1000h/1002h/1006h: YES
```

In the alternate buffer `ybase === ydisp === 0` forever, so `scrollLines` is a guaranteed no-op and the pane sits still under a live drag. tmux never hears about it either, because AO does not forward the drag to the PTY (`forceSelectionMode`, the patch behind #3640).

The **wheel** already solved exactly this: `attachCustomWheelEventHandler` branches on pane kind and reports to the pane rather than scrolling locally. That is why "normal scrolling works". The selection drag was simply never given the same route.

## What changed

`frontend/src/renderer/components/XtermTerminal.tsx`

- Extracted the wheel handler's pane-kind decision into `paneScrollMode()` (`local` / `mouseReport` / `pageKeys`) and its emission into `paneScrollReport()`, so both scroll gestures read one decision and cannot drift apart again. **Wheel behavior is byte-for-byte unchanged** — the existing wheel tests pin those bytes and pass untouched.
- A left-button selection drag held outside the `.xterm-screen` box now emits the same bytes the wheel would for that many lines, on xterm's own 50ms drag cadence.
  - `local` panes (codex, a plain shell, a Windows conpty shell) are **skipped** — xterm's built-in drag-scroll already moves their scrollback, so emitting would scroll them twice.
  - `pageKeys` panes get one page key every 8th tick; one per tick would be twenty screens a second.
  - Distance→speed curve mirrors xterm's own `DRAG_SCROLL_MAX_THRESHOLD`, with the speed cap at 3 instead of 15 because each line becomes a report the *pane* acts on and tmux copy-mode moves 5 lines per report. Measured against a real tmux PTY: 3 reports/tick ≈ 14 lines/tick, i.e. this lands back on xterm's native full-speed drag rate rather than 5× past it.
- Drag tracking uses **capture-phase** `mousemove`/`mouseup` on `window`: xterm's own document-level `mousemove` listener calls `stopImmediatePropagation()` for the whole duration of a drag, which would silently drop a bubble-phase listener.
- Teardown removes the `mousedown` listener and calls `stopSelectionDragScroll()`, so a terminal disposed mid-drag leaves no window listeners or interval behind.

xterm keeps owning the selection model — its `_dragScroll` re-pins `selectionEnd` to the viewport edge every tick — so this only supplies the scroll the pane never received.

## Verification

**Reproduced the bug in a real Chromium** against the shipped `@xterm/xterm@5.5.0` build with AO's `forceSelectionMode` patch applied, synthesizing a held drag 40px above the screen box for 600ms:

| case | buffer / mouse tracking | `viewportY` before → after | auto-scrolled |
|---|---|---|---|
| A — plain shell / codex | `normal`, tracking `none` | 377 → 245 | **YES** |
| B — `tmux attach` (every AO terminal on macOS/Linux) | `alternate`, tracking `drag` | 0 → 0 | **NO** |

Case A proves `forceSelectionMode`, the hidden-scrollbar CSS and `removeHiddenScrollbarReservation` do **not** break xterm's native drag-scroll — the failure was entirely the routing gap. Not Electron-specific.

**Verified the fix's output against a live tmux.** Writing exactly what the fix emits per tick (`\x1b[<64;1;1M` × 3) into a real `tmux attach` PTY, six ticks apart:

```
attached; visible markers: (478, 500)
after 6 drag ticks; markers repainted: (393, 491)
tmux pane_in_mode: 1   scroll_position: 85
```

The pane enters copy-mode and scrolls back 85 lines — the scroll the drag never used to produce.

**Tests** (`npx vitest run --config vite.renderer.config.ts`):
- `XtermTerminal.test.tsx` — 64 passed, 7 new: auto-scrolls a mouse-tracking pane on a held drag; correct direction above vs. below the box; keeps ticking while held and stops on `mouseup`; does **not** fire while the pointer stays inside the box; does **not** fire without a selection (click-and-hold never moves the pane); leaves `local` panes to xterm instead of double-scrolling; paces page keys for keyboard-scroll panes.
- `TerminalPane.test.tsx`, `ShellTerminalTab.test.tsx`, `useTerminalSession.test.tsx` — 98 passed, unchanged.
- `npm --prefix frontend run typecheck` — clean on the implementation.

## Known limitation (not introduced here, not removed here)

Over a tmux pane the selection still cannot accumulate *beyond one screen*: xterm's selection anchors are buffer rows and tmux repaints the alternate screen in place, so scrolling changes the text under a fixed anchor rather than adding rows above it. After this change the pane scrolls under a live drag and the highlight stays WYSIWYG with what gets copied — as far as xterm-level selection can go while the scrollback lives in tmux. Genuinely unbounded selection needs the pane's native transcript: #3791.

## Risk / follow-ups

- The minimum drag speed is ~100 lines/s (1 report/tick × tmux's 5 lines) versus xterm's local 20 lines/s, because tmux's lines-per-wheel-report is not something this layer controls. Fine-grained control right at the edge is coarser than on a `local` pane; worth a tuning pass if it feels fast in practice.
- Windows (conpty) is untested on hardware. A plain conpty shell stays a `local` pane and is unaffected by this change; a conpty agent TUI that enables mouse tracking takes the new `mouseReport` path.

## Related

- #3640 — `forceSelectionMode` suppresses mouse reporting to the PTY. Same patch that makes local selection possible over a mouse-tracking pane, so changes there interact with this.
- #3791 — copy full prompt/response by reading the agent's native transcript.
- #1221 — terminal selection offset from the overlay scrollbar mismatch.
- Separate from the non-terminal UI text-selection issue being triaged in parallel; no shared code — that one is app-chrome CSS, this one is xterm scroll routing.

https://claude.ai/code/session_01Yaidvsj2EAxdKNHEQqgQ6Y

### DIFF
--- a/frontend/src/renderer/components/XtermTerminal.test.tsx
+++ b/frontend/src/renderer/components/XtermTerminal.test.tsx
@@ -70,6 +70,12 @@ vi.mock("@xterm/xterm", () => ({
 		loadAddon() {}
 		open(host: HTMLElement) {
 			host.appendChild(document.createElement("textarea"));
+			// Selection drag auto-scroll measures the pointer against the screen
+			// element's box, exactly as xterm's own drag scroll does. jsdom reports a
+			// zero rect, so the test that exercises it stubs a real one.
+			const screenEl = document.createElement("div");
+			screenEl.className = "xterm-screen";
+			host.appendChild(screenEl);
 		}
 		write() {}
 		writeln() {}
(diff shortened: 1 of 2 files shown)


whoisasx:
Closing after live desktop validation.

The renderer-only workaround is not acceptable: xterm owns the visible selection while tmux owns the actual scrollback. Forwarding edge-drag scroll moves tmux content beneath a fixed xterm highlight; handing the gesture to tmux copy-mode makes the selection use tmux styling and either disappear on mouseup or retain copy-mode and trap normal terminal interaction. That does not meet the expected native selection behavior.

A correct fix needs a first-class selectable transcript/scrollback surface (with a cross-platform path for tmux and ConPTY), rather than synthesizing wheel/mouse input from the renderer. I am closing this PR instead of shipping the compromised behavior.

PR #4273, opened 2026-08-23
fix(desktop): restore text selection outside the terminal

### LINKED ISSUE #4268: bug(desktop): no text outside the terminal can be selected or copied — body { user-select: none } disables every content surface
## Bug

Outside the terminal, no text in the desktop app can be selected or copied. Session details, status text, error messages, PR titles/numbers, branch names, timestamps, and activity-timeline copy are all inert to a mouse drag, so `Cmd/Ctrl+C` has nothing to copy.

**Analyzed against:** `85fab5e29ff718835a098e55ba598f0a2bc91e20` (origin/main) | **AO version:** `ao version dev` (desktop `0.10.3`)
**OS reproduced on:** macOS 26.6 (25G70), arm64 | **Daemon:** running on `127.0.0.1:3001`
**Confidence:** High — exact rule, exact lines, exact introducing commit, reproduced with a computed-style read.

## Reproduction

1. `cd frontend && npm run dev:web` (or open the packaged desktop app) and load the board.
2. Open any session so the inspector rail is visible.
3. Drag the mouse across any inspector text — PR row, status, branch, timestamp.
4. No selection highlight appears. `Cmd/Ctrl+C` copies nothing. Right-click offers no Copy for that text.

Measured directly in the renderer (Playwright against `dev:web`):

```
BODY user-select: none
board-session-card user-select: none
```

Selection is available **only** in `input`, `textarea`, `[contenteditable]`, the xterm terminal (which owns its own selection model), and the two surfaces that have since opted back in by hand — the chat transcript and the GitHub review markdown body.

## Root cause

`frontend/src/renderer/styles.css:1393-1394` sets selection off for the whole renderer:

```css
body {
	...
	-webkit-user-select: none;   /* line 1393 */
	user-select: none;           /* line 1394 */
}
```

`user-select` is inherited, so this disables selection on every descendant. The only re-enable is `frontend/src/renderer/styles.css:1404-1409`:

```css
input,
textarea,
[contenteditable="true"] {
	-webkit-user-select: text;
	user-select: text;
}
```

Form fields are covered; **read-only content is not**.

Introduced by [PR #3508](https://github.com/Untrivial-ai/agent-orchestrator/pull/3508) (`b3dcd1eedc55fa53f5853a4c3774703f1299640e`, "fix: disable text selection on app chrome"). That PR's intent was narrow and legitimate — its description says "so sidebar/board chrome text isn't accidentally selectable" and its test plan only covers "sidebar project names / board column headers". The implementation applied that intent to `body`, so it also took out every content surface. Before #3508 the renderer had no `user-select` rule at all, i.e. content was selectable for the app's whole prior history.

### Collateral: a shipped feature is unreachable

The Files-tab diff selection menu (Copy / Explain / Make changes on a selected diff range) is gated entirely on a live DOM selection:

- `frontend/src/renderer/components/SessionFilesView.tsx:783-784` — `const selection = window.getSelection(); if (!isSelectionActiveIn(selection, container)) return;`
- The diff container at `frontend/src/renderer/components/SessionFilesView.tsx:824-827` carries no `select-text`, so it inherits `user-select: none`.

The user can never produce a selection inside that container, so the context menu never opens and `DiffSelectionMenu` is dead code in practice.

## Why a blanket `user-select: text` is the wrong fix

Chrome that the pointer acts on must stay unselectable, or a drag paints a selection instead of doing its job:

- **Board session cards** (`packages/product-ui/src/SessionsBoardView.tsx:440-441`) are a `<div onClick={onOpen}>`. A drag that ends inside the card still fires `click`, so making card text selectable turns "drag to select a title" into "open the session".
- **Titlebar / window drag strip** — `.window-titlebar` (`styles.css:2728-2729`) pairs `-webkit-app-region: drag` with its own `user-select: none`.
- **Sidebar and inspector topbar** — the chrome [PR #3508](https://github.com/Untrivial-ai/agent-orchestrator/pull/3508) was actually about; `.session-inspector__topbar` (`styles.css:3093`) already opts out on its own.
- **Browser tab rail** uses `@dnd-kit` sortable drag (`frontend/src/renderer/components/BrowserTabsRail.tsx:574`).

So the fix has to restore the selectable default **and** keep an explicit non-selectable list for controls and drag surfaces.

## Not affected

- **Terminal** — `@xterm/xterm/css/xterm.css:38-44` sets `.xterm { user-select: none }` itself and `user-select: text` on its accessibility tree, so xterm owns terminal selection regardless of the `body` value. Restoring the default cannot regress it.
- **Clipboard plumbing** — intact on every platform. macOS/Linux use the native Edit menu; Windows installs the equivalent role menu at `frontend/src/main/menu.ts:12-22` (`copy`, `selectAll`). `Cmd/Ctrl+C` already works; it simply has no selection to act on.
- **Resize drags** — `body.is-resizing-x { user-select: none }` (`styles.css:1889-1891`) already suppresses selection for the duration of a sidebar/inspector drag.

## Cross-platform

`user-select` is a Chromium CSS property with identical semantics on macOS, Windows, and Linux, and the rule is unconditional (no platform guard, no `isWindows()` branch). The bug and the fix are platform-independent. Verified on macOS; Windows/Linux need only a confirmation pass.

## Proposed fix

In `frontend/src/renderer/styles.css`:

1. Make `body` selectable again (restores the pre-#3508 default).
2. Add an explicit non-selectable rule for interactive controls and drag surfaces, preserving [PR #3508](https://github.com/Untrivial-ai/agent-orchestrator/pull/3508)'s actual intent — the existing `cursor: pointer` selector list at `styles.css:1411-1424` is already the codebase's canonical "this is a control" set and can be mirrored.
3. Mark the board session card non-selectable so its click-to-open surface keeps working.

Net effect: readable content becomes selectable everywhere; chrome behaves exactly as it does today.

## Impact

- Users cannot copy error messages, session ids, branch names, or PR numbers out of the app — the exact strings needed to file a bug or paste into a terminal.
- The Files-tab diff selection menu is unreachable.
- Accessibility: users who select text to drive a screen reader or magnifier lose that affordance across the whole non-terminal UI.
- Workaround: none in-app.

## Related

- [PR #3508](https://github.com/Untrivial-ai/agent-orchestrator/pull/3508) — introduced the rule (app chrome).
- [PR #3520](https://github.com/Untrivial-ai/agent-orchestrator/pull/3520) — applied the same pattern to the marketing site (`frontend/src/landing/src/app/globals.css:11-30`), which additionally opts `.prose` back in. Marketing pages are out of scope here.
- [#4265](https://github.com/Untrivial-ai/agent-orchestrator/issues/4265) — terminal selection drag auto-scroll. Different subsystem (xterm owns its selection); tracked separately, no shared code with this fix.
- [#3929](https://github.com/Untrivial-ai/agent-orchestrator/issues/3929) — browser-panel markdown annotations selecting the whole wrapper; adjacent selection behavior in a different surface.

whoisasx:
Fixes #4268

## Summary

`body { user-select: none }` — added by [PR #3508](https://github.com/Untrivial-ai/agent-orchestrator/pull/3508) to stop drags from painting a highlight over sidebar/board chrome — inherits into every descendant. Nothing outside `input`/`textarea`/`[contenteditable]` and the terminal could be selected or copied: session details, status text, error messages, PR metadata. It also left the Files-tab diff selection menu unreachable, since `SessionFilesView` gates that menu on a live DOM selection inside a container that had no `select-text`.

This restores the selectable default and moves #3508's intent into an explicit opt-out list for controls and drag surfaces, so chrome behaves exactly as it does today.

## Changes

- `frontend/src/renderer/styles.css` — drop `user-select: none` from `body`; add an opt-out for `button`, `summary`, `select`, `[role="button"]`, the `data-slot` control set, `[data-slot="resize-handle"]`, and the sidebar menu buttons.
- `packages/product-ui/src/SessionsBoardView.tsx` — `select-none` on **interactive** board cards only. The card is the click target and a drag ending inside it still fires `click`, so a drag-select there would open the session. Non-interactive (archive) cards stay selectable.
- `frontend/e2e/text-selection.spec.ts` — new.
- `packages/product-ui/src/SessionsBoardView.test.tsx` — card contract test.

`a[href]` is deliberately **not** in the opt-out: links sit inside prose (chat transcript, review markdown), and `user-select: none` on them punches holes in a copied paragraph.

## Deliberately unchanged

- **Terminal** — `xterm.css` sets `.xterm { user-select: none }` itself, so xterm owns terminal selection regardless of the `body` value.
- **Titlebar / drag regions** — `.window-titlebar` and `.session-inspector__topbar` already carry their own `user-select: none`.
- **Resize drags** — `body.is-resizing-x { user-select: none }` still suppresses selection for the duration of a drag.
- **Marketing site** — `frontend/src/landing/.../globals.css` keeps its own rule; out of scope.

One behavior change worth flagging: board **column headers** ("WORKING", counts) are now selectable. They are not interactive, so there is no conflict, but #3508's description mentioned them. Happy to add `select-none` there if reviewers prefer the old look.

## Verification

Run on macOS 26.6 (arm64) against `85fab5e29`.

- `frontend/e2e/text-selection.spec.ts` — **3 passed**. Drives real mouse drags and reads `window.getSelection()`:
  - inspector body copy selects and reaches the clipboard over the ordinary Copy shortcut
  - board card drag selects nothing and navigates nowhere
  - buttons, `[data-sidebar="menu-button"]`, and `[data-slot="resize-handle"]` compute `user-select: none`
  - `.xterm` still computes `user-select: none`
- **Non-vacuous check** — reverted `styles.css` to `origin/main` and reran: the content-selection case fails, the two chrome cases still pass. Restored and reran: all 3 pass.
- `packages/product-ui` — `SessionsBoardView.test.tsx` **7 passed**; `tsc --noEmit` clean.
- `frontend` — `tsc --noEmit` clean; `tsc --noEmit -p tsconfig.e2e.json` clean.
- `SessionsBoard.test.tsx` **38 passed**, `ChatWorkspace.test.tsx` + `SessionFilesView.test.tsx` pass (one archive-expansion timeout appeared only under a 3-file parallel run and reproduces on `origin/main` without this change; passes in isolation both with and without it).

## Gaps

- **Windows/Linux unverified.** `user-select` is platform-independent Chromium CSS and the rule has no platform guard, so no divergence is expected — but neither was exercised.
- Verified through the `dev:web` renderer harness, not a packaged Electron build. The `-webkit-app-region` drag strip is untouched by this diff, but a packaged smoke test would confirm the titlebar directly.
- Pre-existing and unrelated: `frontend/e2e/multi-pr.spec.ts` fails on `origin/main` — its "auth stack" fixture is gone from the mock dataset. Not touched here.

## Related

- [#4265](https://github.com/Untrivial-ai/agent-orchestrator/issues/4265) — terminal selection drag auto-scroll. Separate subsystem (xterm owns its selection); no shared code with this change.

https://claude.ai/code/session_01HhZxfD2jzZRk5FTpfgJA8f

### DIFF
--- a/frontend/e2e/text-selection.spec.ts
+++ b/frontend/e2e/text-selection.spec.ts
@@ -1,0 +1,116 @@
+import { expect, test, type Locator, type Page } from "@playwright/test";
+
+// Regression for #4268. PR #3508 put `user-select: none` on <body> to stop drags
+// from painting a highlight over sidebar/board chrome. `user-select` inherits, so
+// it also killed selection on every content surface: session details, status
+// text, error copy, PR metadata — nothing outside inputs and the terminal could
+// be selected or copied. These cases pin both halves of the contract: readable
+// content selects, and the chrome the pointer acts on still does not.
+
+/** Drag across an element the way a user would when selecting its text. */
+async function dragSelect(page: Page, locator: Locator): Promise<void> {
+	await locator.scrollIntoViewIfNeeded();
+	const box = await locator.boundingBox();
+	if (!box) throw new Error("target has no layout box");
+	await page.mouse.move(box.x + 2, box.y + box.height / 2);
+	await page.mouse.down();
+	await page.mouse.move(box.x + box.width - 2, box.y + box.height / 2, { steps: 12 });
+	await page.mouse.up();
+}
+
+async function selectedText(page: Page): Promise<string> {
+	return page.evaluate(() => window.getSelection()?.toString() ?? "");
+}
+
+async function clearSelection(page: Page): Promise<void> {
+	await page.evaluate(() => window.getSelection()?.removeAllRanges());
+}
+
+function userSelectOf(locator: Locator): Promise<string> {
+	return locator.evaluate((el) => getComputedStyle(el).userSelect);
+}
+
+test("readable content is selectable and copyable, chrome is not", async ({ page }) => {
+	await page.goto("/");
+
+	// The root default is what #3508 inverted. It is back to the initial value,
+	// so the assertion is "not switched off" rather than a specific keyword.
+	expect(await page.evaluate(() => getComputedStyle(document.body).userSelect)).not.toBe("none");
+
+	// A board card is a click surface: dragging it must open nothing and select
+	// nothing, so its text stays out of reach even though the page is selectable.
+	const card = page.locator('[data-testid="board-session-card"]').first();
+	await expect(card).toBeVisible();
+	expect(await userSelectOf(card)).toBe("none");
+	await dragSelect(page, card);
+	expect(await selectedText(page)).toBe("");
+	await expect(page).toHaveURL(/#\/?$|projects/);
+
+	// Open a session so the inspector rail — the surface the report named — mounts.
+	await clearSelection(page);
+	await card.click();
+	const inspector = page.locator("#inspector");
+	await expect(inspector).toBeVisible();
+
+	// Inspector body copy is content: a drag selects it and Copy puts it on the
+	// clipboard. This is the case that was impossible before the fix.
+	const section = inspector.locator('[data-testid="inspector-section"]').first();
+	await expect(section).toBeVisible();
+	expect(await userSelectOf(section)).not.toBe("none");
+
+	// Drag a single line of body copy — the PR section's status line — rather than
+	// the whole section, so the gesture stays on text instead of crossing rows.
+	const line = section.locator("p").first();
+	await expect(line).toBeVisible();
+	await dragSelect(page, line);
+	const selection = await selectedText(page);
+	expect(selection.trim()).not.toBe("");
+	// The drag must not have activated the section it crossed.
+	await expect(inspector).toBeVisible();
+
+	// And the selection reaches the clipboard over the ordinary Copy shortcut.
+	await page.context().grantPermissions(["clipboard-read", "clipboard-write"]);
+	await page.keyboard.press(process.platform === "darwin" ? "Meta+C" : "Control+C");
+	const clipboard = await page.evaluate(() => navigator.clipboard.readText().catch(() => ""));
+	expect(clipboard.trim()).toBe(selection.trim());
+});
+
+test("controls and drag surfaces stay unselectable", async ({ page }) => {
+	await page.goto("/");
+	const card = page.locator('[data-testid="board-session-card"]').first();
+	await expect(card).toBeVisible();
+	await card.click();
+	await expect(page.locator("#inspector")).toBeVisible();
+
+	// Buttons: a drag across a control must never leave text selected behind it.
+	const button = page.getByRole("button").first();
+	await expect(button).toBeAttached();
+	expect(await userSelectOf(button)).toBe("none");
+
+	// The sidebar rows #3508 was actually filed about.
+	const sidebarRow = page.locator('[data-sidebar="menu-button"]').first();
+	if ((await sidebarRow.count()) > 0) {
+		expect(await userSelectOf(sidebarRow)).toBe("none");
+	}
+
+	// The inspector/sidebar resize handles are pointer-drag surfaces.
+	const resizeHandle = page.locator('[data-slot="resize-handle"]').first();
+	if ((await resizeHandle.count()) > 0) {
+		expect(await userSelectOf(resizeHandle)).toBe("none");
+	}
+});
+
+test("the terminal keeps its own selection model", async ({ page }) => {
+	await page.goto("/");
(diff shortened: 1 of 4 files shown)


whoisasx:
Closing at the reporter's direction: inverting the global default is the wrong approach. Flipping `body` to selectable and maintaining an opt-out list for chrome puts the burden on every future control to remember to opt out, and silently regresses the moment one doesn't.

Leaving #4268 open. If this gets picked up again, the narrower direction is to keep `body { user-select: none }` untouched and opt **in** per content surface with `select-text` — the pattern already used by the chat transcript (`ChatWorkspace.tsx:1563`) and the GitHub review body (`SessionInspectorView.tsx:1213`).

PR #4274, opened 2026-08-23
feat(sidebar): reorder projects and sessions by mouse

whoisasx:
## Summary

Adds mouse drag-and-drop reordering for projects and worker sessions in the sidebar. Ordering is renderer-local for the current app session; it does not persist to `localStorage` or change daemon data.

- Projects drag as a complete block: an expanded project’s visible sessions travel in the overlay, while the list stays still until drop.
- Project drop targets are the top and bottom edges of complete project blocks.
- Sessions reorder only within their own project and animate their siblings during drag.
- Rows remain the drag surfaces; action controls and folder disclosure do not start navigation after a drop.
- Pointer activation uses a 4px movement threshold. Keyboard reordering is intentionally not included.

## Verification

- `npm --prefix frontend run test -- --run src/renderer/components/Sidebar.test.tsx` (76 passing)
- `git diff --check`

## Scope

No daemon, API, schema, or persistence changes.


whoisasx:
# Expected Behaviour 

https://github.com/user-attachments/assets/1ef33a8a-a55f-41d4-848c-6fcf0ae92c32

ronishrohan (changes requested):
this needs quite a lot of changes, a bit too unpolished for the app, will take over from here

whoisasx:
@ronishrohan thank you for updating on it.

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -1017,99 +1638,145 @@ function SessionRow({
...
+				!isDragging &&
+					"group-hover/session-row:pointer-events-auto group-hover/session-row:opacity-100 group-focus-within/session-row:pointer-events-auto group-focus-within/session-row:opacity-100",
+			)}
+			data-session-actions=""
+			onPointerDown={(event) => event.stopPropagation()}
+		>
+			<button
+				aria-label={session.isPinned ? t("shell.unpinSession") : t("shell.pinSession")}
+								className={cn(SESSION_ACTION_CLASS, session.isPinned && "text-foreground")}
+								tabIndex={-1}
```
whoisasx: [P1] Mouse-only reordering can be an intentional scope choice, but this removes keyboard access from unrelated existing controls. Pin, rename, and terminate are all now `tabIndex={-1}` (and the same pattern is used for project actions), so keyboard-only users cannot reach these workflows at all. Please restore the action buttons to the normal tab order and keep the folder disclosure as a native focusable button; visual focus chrome can be handled without making the controls unfocusable.

whoisasx (commented):
The session-local ordering model and the per-project drag boundaries are generally coherent, and the focused behavior tests pass. Three regressions should be addressed before merging: existing project/session actions were removed from the keyboard tab order, the insertion indicator disappears in the supported light theme, and compact icon-rail drags expose sessions that are otherwise hidden. Please address the inline findings.

Verification: all GitHub checks are green; the PR merges cleanly with current main; the focused Sidebar suite passes 76/76; and `git diff --check` passes. I also reproduced the compact-mode overlay issue with an isolated regression test. The PR title should be updated as well: it still says "drag or keyboard," while the current implementation and description intentionally exclude keyboard reordering.

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -687,172 +1125,219 @@ function ProjectItem({
...
+					onMouseEnter={() => setRowHovered(true)}
+					onMouseLeave={() => setRowHovered(false)}
+					ref={setDroppableNodeRef}
+					transition={prefersReducedMotion ? { duration: 0 } : { type: "spring", stiffness: 520, damping: 42, mass: 0.55 }}
+				>
+					{dropIndicator ? (
+						<div
+							aria-hidden="true"
+							className={cn(
+								"pointer-events-none absolute inset-x-0 z-[70] h-px bg-white",
```
whoisasx: [P2] This hard-coded white drop indicator has zero contrast against the supported light-theme sidebar, whose background is also pure white. The insertion boundary therefore disappears precisely while the user is deciding where to drop. Please use a theme-aware semantic token such as `bg-foreground` or an accent color, and cover both light and dark themes.

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -416,33 +750,47 @@ export function Sidebar({
...
+											consumeDragClick={projectDragClickGuard.consumeClick}
+											onSessionOrderChange={recordSessionOrder}
+											onToggle={() => toggleProjectDisclosure(workspace.id)}
+											onRemoveProject={onRemoveProject}
+										/>
+									))}
+									{isCollapsed && <CreateProjectListItem />}
+								</SidebarMenu>
+								<DragOverlay adjustScale={false} dropAnimation={null} modifiers={[restrictProjectOverlayToRows]} style={PROJECT_DRAG_OVERLAY_STYLE} zIndex={60}>
+									{activeDragWorkspace ? <ProjectDragPreview expanded={expandedIds.has(activeDragWorkspace.id) || (initialActiveSessionProjectId === activeDragWorkspace.id && !dismissedInitialActiveProjectIds.has(activeDragWorkspace.id))} selection={selection} sessions={activeDragSessions} workspace={activeDragWorkspace} /> : null}
```
whoisasx: [P2] This overlay still treats a stored-expanded project as expanded when the sidebar is in the compact icon rail. The real session list is hidden there, but starting a project drag renders the project name and every hidden session inside the narrow overlay; I reproduced the overlay text as `Project Onefix login`. Please gate the preview with `!isCollapsed` (or disable project dragging in compact mode) and add a compact-sidebar regression test.

frontend/src/renderer/components/Sidebar.tsx
ronishrohan: Intentional. These are hover-revealed, pointer-only supplemental actions; keeping them out of the Tab sequence avoids making keyboard users traverse hidden controls. The project/session row remains the keyboard-accessible navigation surface, and the folder overlay is deliberately a pointer-only disclosure target.

frontend/src/renderer/components/Sidebar.tsx
ronishrohan: Fixed in `140205f96`. The insertion boundary now uses `bg-foreground` rather than hard-coded white, so it remains visible on both light and dark sidebar surfaces.

ronishrohan (approved):
Re-reviewed the final revision: all requested fixes are in, checks are green, and the mouse-only interaction scope is intentional. Approved.

PR #4275, opened 2026-08-23
feat: allow 120-character session display names with a sliding sidebar label

whoisasx:
## What

Session display names were capped at 20 characters — too short to distinguish two sessions working in the same area. This raises the cap to **120** and gives the sidebar a way to actually show a name that long.

Session IDs and prefixes are untouched; this is display names only.

## Backend / API

- **`domain.MaxSessionDisplayNameLen` (120) + `SessionDisplayNameTooLong`** are the single authority. Every boundary counts **runes**, not bytes, so accented/CJK/emoji labels are not cut short by a byte budget.
- **Session service** enforces it on both `Spawn` and `Rename`, so the service is authoritative regardless of transport. `Rename` previously had **no length check at all** — a `PATCH` was the one way to put an unbounded label on a session.
- **HTTP**: spawn and rename handlers both return `DISPLAY_NAME_TOO_LONG`, with the message derived from the constant so it cannot go stale.
- **DTO / OpenAPI**: `SpawnSessionRequest.displayName` and `RenameSessionRequest.displayName` are `maxLength: 120`. `npm run api` regenerated `openapi.yaml`; `frontend/src/api/schema.ts` is unchanged (openapi-typescript does not emit `maxLength`).
- **CLI**: `ao spawn --name` and `ao session rename` both reject an oversized label as a usage error (exit 2) before contacting the daemon; help text derives the number from the constant.
- **Agent-facing guidance** (orchestrator prompt, spawn directive, task-title delegation message, `using-ao` skill docs) quotes the cap from the constant.
- `delegatedTaskTitleLimit` stays at **20** on purpose: it is the provisional label cut from a raw brief while the orchestrator picks a real title, and a blind slice of prose reads worse the longer it runs. Commented as such.

Project display names are unrelated and stay at 20.

## Sidebar

- Session names render **in full, clipped rather than ellipsised** — the distinguishing part of a session name is usually its tail.
- **When and only when the name overflows**, hovering the row or focusing its open button slides the text far enough to read the tail, then eases back. Constant travel speed (~100px/s, about reading pace) with a dwell at each end, so a name two characters too wide doesn't crawl and a long one doesn't whip past.
- **Reserved action area, no layout shift.** The pin/rename/kill cluster moved out of flow into reserved padding (`--size-sidebar-session-actions`). It used to grow from zero width on hover, which resized the label track at exactly the moment the label starts moving. Now only opacity changes, and the cluster sits above the moving text (`z-chrome`). Hidden means `pointer-events-none`, so a tap on the row's right edge can't hit an invisible Kill button.
- **Theme-aware edge fade** via a CSS `mask-image`, not a background-coloured gradient. A gradient would have to name a colour, and the row sits on three backgrounds (rest / hover / active) across two themes; the mask fades whatever is actually painted, so nothing is hard-coded.
- **Pauses and resets cleanly** — leaving mid-travel transitions back to rest rather than snapping.
- **Accessible name / tooltip**: the full name is always the row's `title` and its text content.
- **`prefers-reduced-motion`**: never animates, in both the component and a CSS backstop. The non-animated readable fallback is the tooltip plus the full accessible name.

## Verification

**Backend** — `go build ./...` and `go vet` clean.
`go test` green for `internal/domain`, `internal/cli`, `internal/httpd/controllers`, `internal/service/session`, `internal/session_manager`, `internal/skillassets`.
(Three `TestSwitchAgent*` tests in `internal/session_manager` fail intermittently under load on this machine — verified they fail identically on the unmodified tree, so pre-existing, not from this change.)

**Frontend** — `npm run typecheck` clean. `Sidebar.test.tsx` 63/63 and `SidebarSessionName.test.tsx` 8/8 pass. A full-suite run showed failures only in files this PR does not touch (e.g. `ProjectSettingsForm.test.tsx`), consistent with load-related flakiness in a parallel run.

**Not done:** visual `ao preview` / `ao browser` verification was skipped at the requester's instruction. The marquee's motion has not been eyeballed in the running app.

## Tests added

- `domain`: rune-boundary table (ASCII / 2-byte / 4-byte, at cap and one over).
- `service/session`: spawn and rename cap, trim-before-measure, rejected rename leaves the stored name untouched.
- `httpd/controllers`: spawn and rename each accept exactly-at-cap (multi-byte) and reject one over.
- `cli`: spawn and rename usage errors, and at-cap names forwarded unchanged.
- frontend: overflow detection (including sub-pixel), hover trigger, focus trigger, clean reset, reduced-motion fallback, no-reflow/reserved-strip assertions, constant-speed duration scaling.

## Scope note

Overlaps sidebar DnD and agent-icon work in other sessions — kept independently scoped to the session row's label and action cluster.

https://claude.ai/code/session_01UgVxtwx2vARFH1XavZrJhP

### DIFF
--- a/backend/internal/cli/session.go
+++ b/backend/internal/cli/session.go
@@ -11,8 +11,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 	"github.com/spf13/cobra"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 )
 
 type sessionOptions struct {
--- a/backend/internal/cli/session.go
+++ b/backend/internal/cli/session.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 	"github.com/spf13/cobra"
 )
 
@@ -235,6 +236,7 @@ func newSessionRenameCommand(ctx *commandContext) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "rename <id> <name>",
 		Short: "Rename a session",
+		Long:  fmt.Sprintf("Rename a session.\n\n<name> is the sidebar label and must be %d characters or fewer.", maxDisplayNameLen),
 		Args:  sessionRenameArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			id, err := normalizeSessionID(args[0])
@@ -333,8 +335,14 @@ func sessionRenameArgs(cmd *cobra.Command, args []string) error {
 	if _, err := normalizeSessionID(args[0]); err != nil {
 		return err
 	}
-	if strings.TrimSpace(args[1]) == "" {
+	name := strings.TrimSpace(args[1])
+	if name == "" {
 		return usageError{errors.New("session name is required")}
+	}
+	// Rejected here rather than forwarded so an oversized label exits 2 as CLI
+	// misuse; the daemon holds direct API callers to the same cap.
+	if domain.SessionDisplayNameTooLong(name) {
+		return usageError{fmt.Errorf("session name must be %d characters or fewer", maxDisplayNameLen)}
 	}
 	return nil
 }
--- a/backend/internal/cli/session_test.go
+++ b/backend/internal/cli/session_test.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -10,6 +11,8 @@ import (
 	"strings"
 	"sync"
 	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 )
 
 type sessionRequestLog struct {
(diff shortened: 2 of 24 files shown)


whoisasx:
# Expected behaviour 

https://github.com/user-attachments/assets/a0a3b241-edf4-4802-88d3-a374bf8c0032

ronishrohan:
will take a look at this too, after the other one is done, a bit iffy around the edges

whoisasx:
@ronishrohan take your time.

i-trytoohard:
From the NotesBot call on `2026-09-04`, this existing pr appears to be relevant.
Match confidence: `0.60` (fuzzy).

Relevant discussion excerpts:
- [18:26] goatedinsaan: That's 20 characters. [18:29] prateek: Sliding sidebar label, what does that mean? [18:33] !adil: Like if you have like your like session name is long enough, then like it will be of sliding, like when you will hover upon it, it will slide and give you the whole name. Other than that, it [...]
- [18:29] prateek: Sliding sidebar label, what does that mean? [18:33] !adil: Like if you have like your like session name is long enough, then like it will be of sliding, like when you will hover upon it, it will slide and give you the whole name. Other than that, it will keep—
- [18:29] prateek: Sliding sidebar label, what does that mean? [18:33] !adil: Like if you have like your like session name is long enough, then like it will be of sliding, like when you will hover upon it, it will slide and give you the whole name. Other than that, it will keep— [18:49] goatedinsaan: huh, got it.

Source thread: https://discord.com/channels/1488170845154050258/1545490043576000592/1545490043576000592

PR #4276, opened 2026-08-23
fix(events): stop a large replay from aborting the mobile app, and clamp future cursors to head

Prasad-D-Ware:
Two commits. The first is the production fix that has shipped; the second is a small daemon-side hardening of the same failure mode.

---

### Problem

On a fresh install the conversation cursor starts at zero, so connecting asks the daemon for every event it has ever recorded. Against a real 80-session daemon that is **196,658 events / 47MB**, still streaming after 6 seconds.

Read through `expo/fetch`, each chunk emits an event to JS and takes a **JNI global reference** released only once JS consumes it. The reader awaited only already-buffered reads — those stay in the microtask queue — so it never yielded, JS never drained, references accumulated, and the runtime aborted the process on hitting the 51,200 table limit:

```
JNI ERROR (app bug): global reference table overflow (max=51200)
    50472 of java.lang.Object[] (1 elements)
  at expo.modules.fetch.NativeResponse.pumpResponseBodyStream(NativeResponse.kt:199)
Fatal signal 6 (SIGABRT)
```

Android reports this as `reason=5 (APP CRASH(NATIVE))` with no dialog, so it presents as the app silently vanishing a few seconds after connecting. Scrolling still worked while it was wedged — that is handled natively — but taps did not, because those need the JS thread.

### Fix

- **Yield a macrotask every 256 frames** inside the frame loop, so consumption keeps up and references are released. One chunk can carry thousands of events, so yielding per chunk is not enough.
- **Persist the cursor by event count** instead of a 500ms debounce. The debounce is a macrotask, so during exactly the drain where progress matters it never ran — every launch restarted from zero.
- **Skip parsing payloads while nothing is subscribed.** Only the chat screen currently open subscribes; during a cold replay none is, so ~196k `JSON.parse` calls were pure waste.

### Verification

- 459 tests, typecheck clean.
- `yields to the event loop while draining a large replay` fails on the pre-fix code — it reproduces the starvation in a unit test and now guards it.
- **On device** (Galaxy M31s, release build, `pm clear` before each run) against the 80-session daemon: baseline aborts at 50,472 references; this build drains cleanly and stays responsive. Relaunch is instant, confirming the cursor persists so catch-up happens once per install.
- **iOS** verified via a fresh TestFlight install against the same daemon — no hang. The JNI ceiling is Android-only, but the JS-thread starvation was not.

---

## 2. `fix(events)`: clamp a future cursor to head instead of replaying from zero

A cursor ahead of head means the `change_log` was truncated or replaced. The stream reset such a client to `0`, which costs it the entire backlog — and because every connected client is reset in the same moment, they stampede together.

It now falls back to head. The gap is recovered from the next snapshot fetch, so the cost is one refetch rather than a full replay.

The existing test asserted the reset-to-zero behaviour; it now asserts the clamp, and additionally that no event at or before head is replayed. `go test ./...` and `go test -race` pass.

---

## Known limitation

Together these make the replay **survivable and one-time — they do not make it smaller**. 47MB still crosses the network on a fresh install, and `pumpResponseBodyStream` has no backpressure.

The real fix is not replaying history at all: the client already loads history through paginated REST snapshots, and the sole stream consumer discards the payload and just refetches. A follow-up covers `from=head`, `maxCatchup`, and replacing the overflow-cancel-then-replay path with a resync signal — deliberately deferred, since the crash is resolved.

### DIFF
--- a/backend/internal/httpd/events.go
+++ b/backend/internal/httpd/events.go
@@ -55,8 +55,13 @@ func (c *EventsController) stream(w http.ResponseWriter, r *http.Request) {
 			"Could not inspect the event cursor", nil)
 		return
 	}
+	// A cursor ahead of head means the change_log was truncated or replaced. Fall
+	// back to head rather than zero: replaying from zero costs the client the whole
+	// backlog, and since every connected client is reset at the same moment, they
+	// stampede together. Falling back to head loses at most the events in the gap,
+	// which clients recover from their next snapshot fetch.
 	if after > latestSeq {
-		after = 0
+		after = latestSeq
 	}
 
 	flusher, ok := w.(http.Flusher)
--- a/backend/internal/httpd/events_test.go
+++ b/backend/internal/httpd/events_test.go
@@ -133,7 +133,11 @@ func (s *resetEventSource) EventsAfter(_ context.Context, after int64, _ int) ([
 
 func (*resetEventSource) LatestSeq(context.Context) (int64, error) { return 1, nil }
 
-func TestEventsStreamResetsCursorAheadOfCurrentDatabase(t *testing.T) {
+// A cursor ahead of head means the change_log was truncated or replaced. Such a
+// client is sent to head, not to zero: replaying from zero costs it the entire
+// backlog, and because every connected client is reset at the same moment, they
+// stampede together. The gap is recovered from the next snapshot fetch instead.
+func TestEventsStreamClampsCursorAheadOfCurrentDatabaseToHead(t *testing.T) {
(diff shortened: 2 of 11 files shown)


Vaibhaav-Tiwari (changes requested):
just one blocker
  The backend now skips replay when a client cursor is ahead of the event log. Mobile
  handles that reset header, but desktop EventSource cannot read the header, so desktop chat may stay stale after an event-log reset.

Vaibhaav-Tiwari (approved):

PR #4287, opened 2026-08-23
chore(mobile): upgrade Expo SDK 54 to 57

Prasad-D-Ware:
Moves the mobile app from Expo SDK 54 to 57 — React Native 0.81.5 → 0.86.2, React 19.1.0 → 19.2.3.

SDK 55 and 56 are skipped deliberately. Both carry a Hermes V1 memory regression that inflates memory usage when `react-native-worklets` or `reanimated` is imported; it is fixed in `expo@57.0.9`, and this lands on 57.0.15.

Rebased onto `main` after #4276 merged, so this is a single commit touching only `packages/mobile`.

### Breaking changes the new SDKs required

- **`StyleSheet.absoluteFillObject` was removed.** `absoluteFill` replaces it and has the identical object shape, so the existing spreads behave the same.
- **SDK 56+ forbids importing `@react-navigation/*` directly.** `useTabScrollToTop` now takes `useScrollToTop` from `expo-router/react-navigation`, the supported re-export. Runtime behaviour is unchanged.
- **`splash` is no longer valid at the `app.json` root** and moves into the `expo-splash-screen` plugin. `expo-font` is added because `@expo/vector-icons` requires it as a native peer — expo-doctor flags its absence as a crash risk outside Expo Go.
- **`useColorScheme()` is no longer nullable** and can return `"unspecified"`, so `resolveScheme` accepts it. It falls through to dark exactly as `null` did.

### Other changes

**`expo-speech-recognition` 3.1.3 → 56.0.1.** Worth a look. It is third-party, so `expo install --fix` does not touch it, and its SDK 54 build fails to compile against the new `expo-modules-core` (`promise.resolver` became `legacyResolveBlock(for:)`). Note `56.0.1` is the **SDK 56** release — upstream has not tagged one for SDK 57 (last publish 2026-06-05). Voice input works, but is a version behind by necessity.

**`tsconfig.json` gains an explicit `types` entry.** The Node types the vitest suites rely on used to arrive transitively and no longer do. `"react"` is listed alongside `"node"` because setting `types` at all disables automatic `@types` inclusion.

**`metro.config.js` is deliberately untouched.** The upgrade guide suggests deleting a Metro config that only holds Expo defaults; this one pins `react` and `react-native` to a single copy, and removing it reintroduces the "main has not been registered" launch crash.

**TypeScript 5.9 → 6.0.3** came in via `expo install --fix` as the SDK-aligned version. Everything typechecks on it. Happy to pin back to 5.9 if reviewers would rather not couple a TS major to this.

### Verification

- `expo-doctor` 21/21
- `tsc --noEmit` clean
- 475 tests across 56 files passing
- Native iOS build installs and renders on an iOS 26.5 simulator
- Native Android build installs and renders on an Android 15 emulator

The native builds were the only thing that caught the `expo-speech-recognition` breakage — doctor, typecheck, tests, and `expo export` all passed while the iOS build was failing. Worth keeping in mind that CI has no mobile target, so none of the above runs automatically.

### Not yet verified on real devices

Hand-testing so far has been on a simulator and an emulator. Still outstanding:

- **Terminal** — `@fressh/react-native-xtermjs-webview` pins React `19.1.0` exactly and we now ship `19.2.3` (installs via the existing `legacy-peer-deps`). Most likely place for a runtime surprise.
- **Chat streaming** — `expo/fetch` and Hermes both changed under the JNI fix from #4276; needs a large Android event replay to exercise properly.
- **Voice input** — see the `expo-speech-recognition` note above.
- **Upgrade-over-install** — `expo-secure-store`, `async-storage`, and `expo-file-system` all jumped major native versions. A fresh install will not reveal migration problems; installing over an existing SDK 54 build is the only way to confirm users stay paired and signed in.

### Follow-ups, not in this PR

- `expo-system-ui` is absent, so `ios.backgroundColor` and `userInterfaceStyle` in `app.json` are silently no-ops. **Pre-existing** — it was missing on SDK 54 too — but worth fixing separately.
- Deprecations left alone: `@react-native-async-storage/async-storage` → `expo-sqlite/localStorage`, `@expo/vector-icons` → `expo-symbols`, and enabling React Compiler.


Vaibhaav-Tiwari (approved):
looks consistent

PR #4293, opened 2026-08-23
fix(desktop): bundle tmux on macOS and Linux

harshitsinghbhandari:
## Summary
- build tmux 3.5a plus pinned, checksum-verified static libevent/ncurses/utf8proc dependencies during macOS and Linux desktop packaging
- ship the result as `resources/tmux/bin/tmux` and fail packaging if that resource is absent or reports the wrong version
- force packaged desktop daemons to use the bundled binary via `AO_TMUX_BINARY`, with an AO-owned tmux socket so they cannot attach to the machine tmux server
- centralize configured/bundled/system tmux resolution and use it in the runtime, session spawn gate, startup system check, and `ao doctor`
- keep standalone/development daemon behavior backward compatible and cover binary/socket selection with tests

Verified the original gap and rebased this change onto current `upstream/main` (`1880498c8`). This supersedes the bundled-resolution intent of #4164 while also implementing the in-repo macOS/Linux packaging and socket isolation that PR defers elsewhere.

## Verification
- `env -u AO_SESSION_ID -u AO_RUNTIME_LAUNCH_ID -u AO_BROWSER_CAPABILITY -u AO_PROJECT_ID TMUX_TMPDIR=/tmp/ao-tmux-lint-247 npm run lint`
- `cd backend && go test ./internal/tmuxbin ./internal/adapters/runtime/tmux ./internal/session_manager ./internal/service/systemcheck ./internal/cli`
- `cd frontend && npm test -- --run forge.config.test.ts src/shared/bundled-tmux.test.ts src/shared/shell-env.test.ts`
- `cd frontend && npm run build:tmux`
- macOS arm64 linkage check: only `/usr/lib/libSystem.B.dylib` and `/usr/lib/libresolv.9.dylib`
- packaged the Electron app with Node 22 and smoke-tested the packaged tmux against an empty environment/PATH

## Risk
Linux and macOS Intel native builds are performed on their matching release runners; the package hook now makes either build fail closed if tmux is missing or not runnable.


illegalcall:
I found three concrete behavioral issues in this change:

1. **Existing sessions are not migrated from tmux's default socket.** Packaged daemon shutdown deliberately leaves live tmux sessions running so the next daemon can adopt them, but `frontend/src/main.ts` now starts every packaged daemon with `AO_TMUX_SOCKET_NAME=ao`. On the first daemon restart after upgrading, reconciliation probes only `-L ao`; the old session on the default socket appears unavailable, so `reconcileLive` launches a replacement in the same worktree while the original session remains alive and unreachable. I reproduced both sockets containing the same AO session name. This needs a legacy-socket migration/probe or socket identity persisted with each runtime handle.

2. **`AO_KEEP_DAEMON` leaves Linux AppImage daemons pointing at an unmounted tmux binary.** `AO_TMUX_BINARY` is derived from `process.resourcesPath`, which is the temporary `/tmp/.mount_*` AppImage mount. A keep-alive daemon intentionally survives Electron, but it does not inherit the AppImage runtime's keepalive file descriptor; when Electron exits the mount is removed. Subsequent tmux attach/send/probe calls use a path that no longer exists. The bundled binary needs to be staged under a stable AO-owned path (for example under `~/.ao`) or otherwise kept accessible for the daemon's lifetime.

3. **`ao start` still tells desktop users to install machine tmux.** `backend/internal/cli/start.go` unconditionally calls the PATH-only `ensureTmux`, whose warning says sessions will fail without installing tmux. With this PR, the app it resolves/downloads already contains tmux, so that warning is false. Remove it from the desktop-launch path or make it aware of the bundled resource.

harshitsinghbhandari:
Addressed all three findings in `3e81c8c9b` and updated the branch to current `main` (`1880498c8`):

1. Existing runtime handles now discover sessions on AO's private socket first and fall back to tmux's legacy default socket only when the private server definitively lacks the session. The socket choice is cached per session; new sessions remain private. A regression test covers reconciliation-style liveness probing and cached legacy routing. Protocol mismatches now surface as `ErrRuntimeUnavailable` rather than an opaque probe error.
2. Electron now atomically stages the signed bundled tmux into versioned storage under the AO data directory (`AO_DATA_DIR`, otherwise `~/.ao`) before spawning the daemon. `AO_TMUX_BINARY` points there, so a persistent AppImage daemon never depends on the temporary FUSE mount, and an update cannot replace an older daemon's binary.
3. Removed the PATH-only tmux warning/install helper from `ao start`; desktop bundles are self-contained.

Verification after merging current main:
- full `npm run lint` (all Go tests + golangci-lint; 0 issues)
- focused runtime/CLI/systemcheck/session-manager tests
- frontend bundled-tmux/Forge tests and full frontend typecheck

illegalcall (approved):

illegalcall:
Re-checking the merged commit found two remaining correctness issues:

1. **Legacy migration still turns an inconclusive probe into proof that the session is dead.** In `backend/internal/adapters/runtime/tmux/tmux.go:832-836`, any error from the legacy/default-socket `has-session` probe is discarded and `socketForSession` returns the private socket. `IsAlive` then probes the private socket again; its definitive “session missing” output becomes `(false, nil)`, so reconciliation relaunches even though the legacy session may still be alive. I reproduced the real upgrade case with a tmux 3.6a server on the default socket and this PR’s bundled tmux 3.5a client: the bundled probe returned `server exited unexpectedly`, while the 3.6a client still listed the live session. The migration path must propagate an inconclusive legacy error (and handle client/server incompatibility) unless both sockets definitively prove absence.

2. **The private socket is not isolated per AO data directory.** `frontend/src/main.ts:833` hardcodes `AO_TMUX_SOCKET_NAME=ao` even when `AO_DATA_DIR`, port, and run file are isolated. A tmux `-L` socket is per OS user, so every packaged AO instance for that user shares the same tmux server. Session IDs are allocated per database as `{project}-{num}`, so two isolated AO instances can generate the same ID; one daemon can then adopt, message, attach to, or destroy the other instance’s session. Derive the socket identity from the canonical AO data directory (and require ownership evidence when adopting from legacy sockets) so isolated instances cannot cross-control sessions.

AgentWrapper:
Post-merge nightly regression found in the bundled artifact: [#4382](https://github.com/Untrivial-ai/agent-orchestrator/issues/4382). The statically linked ncurses retains its CI-prefix terminfo lookup, so a real PTY attach prints `can’t find terminfo database` and exits; `tmux -V` does not exercise this path.

PR #4298, opened 2026-08-23
fix: move Queue/Steer chips inside the composer toolbar

Annieeeee11:
## Summary
- Move Queue/Steer from above the composer into the bottom toolbar, immediately before the right-side mode/approval dropdown.
- Pass delivery chips through `TurnSettingsBar` children; fall back to the tools row when settings are absent.

## Test plan
- [ ] Start a chat turn so Queue/Steer appears
- [ ] Confirm chips sit inside the composer, left of the Default/approvals control
- [ ] Confirm Cmd/Ctrl+Enter still steers and bare Enter still queues
- [ ] `npx vitest run src/renderer/components/chat/ChatSteer.test.tsx src/renderer/components/chat/ChatComposer.test.tsx src/renderer/components/chat/TurnSettingsBar.test.tsx`

## screenshot
image

### DIFF
--- a/frontend/src/renderer/components/chat/ChatComposer.tsx
+++ b/frontend/src/renderer/components/chat/ChatComposer.tsx
@@ -28,6 +28,8 @@
  */
 
 import {
+	cloneElement,
+	isValidElement,
 	useCallback,
 	useEffect,
 	useId,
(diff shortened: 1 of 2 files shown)


ronishrohan (approved):
Reviewed the composer toolbar change against `main`.

The production path is sound: `ChatWorkspace` passes a childless `TurnSettingsBar`, and the injected Queue/Steer control renders immediately before the mode/approval picker. The no-settings fallback remains intact.

Verified locally: 67/67 targeted tests pass, frontend typecheck passes, and `git diff --check` is clean. CI test, renderer-smoke, and scan checks are also green.

Non-blocking follow-up: add a focused test that renders settings and steering together and asserts the delivery group is inside `Turn settings` before the approval/mode control.

PR #4251, opened 2026-08-22
feat(files): render .md files GitHub-style in the Files tab

Vaibhaav-Tiwari:
## Summary
- Adds a per-file **"Source diff" / "Rendered"** toggle for `.md`/`.markdown` files in the session Files tab, so a worker's README/design doc/etc. can be read as rendered markdown instead of only as a diff.
- Rendering uses `react-markdown` + `remark-gfm` (tables, task lists, strikethrough) + `rehype-github-alerts` (`> [!NOTE]`/`[!TIP]`/etc.) + `rehype-slug`/`rehype-autolink-headings` (hover-`#` heading anchors that scroll + copy a same-page link), styled with `github-markdown-css`'s `.markdown-body` for pixel-accurate GitHub spacing/typography/alert colors. `rehype-raw` stays off, matching `ChatMarkdown.tsx`'s existing "markdown-only is the whole sanitization story" posture.
- Fenced code reuses the app's existing lowlight/highlight.js engine (`HighlightedCode`, extended with a new `.markdown-code` scope in `code-theme.css`) rather than adding Shiki — the renderer's CSP (`script-src 'self'`, no `wasm-unsafe-eval`) blocks Shiki's default WASM engine, the same reason chat code blocks already avoid it.
- Relative image references (`x`) resolve against the file's actual worktree directory and load via the existing `/workspace/file/blob` route (the same mechanism `ImageDiffView.tsx` already uses for image-diff previews) — no new backend endpoints needed.
- A relative link to another repo file (`./OTHER.md`) renders as inert text rather than a dead/misleading link, since there's no in-app "open this other file" navigation yet.
- No explicit light/dark plumbing was needed: `github-markdown-css`'s stylesheet follows `prefers-color-scheme`, and `main.ts` already drives Electron's `nativeTheme.themeSource` from AO's own theme preference, so it automatically tracks AO's theme rather than the raw OS setting.

## Test plan
- [x] `npx tsc --noEmit` — clean (the only pre-existing errors are in `packages/product-ui`, unrelated to this change and present on `main` before this branch)
- [x] Existing `SessionFilesView.test.tsx` (35 tests) passes unmodified — no regressions
- [x] `renderer-coverage.test.ts` (i18n hardcoded-string gate) passes — no hardcoded English chrome in the new components
- [ ] Manual verification in the real Electron app (isolated dev mode) — not yet done; opening this PR for others to try against a real session with a `.md` file (table, alert callout, heading, fenced code, relative image)

[...]

PR #4301, opened 2026-08-23
fix(files): correctness follow-ups for the markdown Preview tab

Pyasma:
Follow-ups for #4251, branched off `feat/github-markdown-render`. Merging this into that branch folds the commits into the PR.

Seven correctness fixes, three cleanups, the first tests for the new code, and a rename plus redesign of the tab pair. One commit per item, each with its reasoning, so anything you disagree with can be dropped on its own.

## Bugs

**Relative images never reloaded after an edit.** `buildWorkspaceBlobUrl` mirrors `ImageDiffView`'s blob route but not its `v=` parameter, and that parameter is why the route works — it serves `no-store`, so as `ImageDiffView`'s own comment puts it, without a changing URL the element never refetches at all. An image an agent rewrote kept rendering from cache indefinitely. Threads the detail's load timestamp through `MarkdownFileContext`.

**A failed image rendered as a broken-image box.** The CSP allows `img-src` from `'self'`, `data:` and the loopback daemon only, so an `https://` image — a shields.io badge, a hosted screenshot, the two most common images in a real README — cannot load here at all. Falls back to alt text, the way `ImageDiffView` already handles its own failures.

**Truncated files rendered as though complete.** The daemon caps content at `maxWorkspaceFileBytes` and reports `contentTruncated`; the diff body respects it, the rendered view ignored it. The rendering is worse than the omission: a cut landing inside a fence leaves it unclosed, swallowing everything after it into one code block, with no signal that anything is missing. Adds the same warning strip behind a new `files.contentTruncated` string, translated for all eight locales.

**Heading links could scroll to the wrong file.** `rehype-slug` guarantees ids unique per rendered file, not per document, and the Files tab expands many cards at once — two open READMEs with `## Overview` put two `id="overview"` in one DOM, and `document.getElementById` answers with whichever is first. Resolves within the file's own container instead. Prefixing the slugs would also have worked but would have broken every hand-written `[link](#overview)`, which is the more common case.

**Clicking a heading overwrote the clipboard.** Silently, with no affordance saying so, and with a bare `#some-heading` that names no file and cannot be pasted anywhere and followed. Scroll only; a deliberate copy-link control can come later.

**The selection guard leaked on tab switch.** `DiffView` reports selection state up so a background refetch cannot re-render rows out from under a live selection, and that effect had no cleanup. Harmless until this PR, since the diff only unmounted with the card. Now: switch tabs with a selection active and `selectionOrMenuActive` stays `true` forever, the detail query stays disabled, and the rendered view sits on a snapshot that never refreshes while the agent keeps writing — with nothing left mounted that could clear it.

**Tab switching lost scroll position.** Not because a scroll position is discarded — an expanded file has no scroll box, the Files panel is one shared scroller. Swapping a 4,000-line diff for its rendered form shortens the column by thousands of pixels, and the browser clamps `scrollTop` to the new maximum; switching back restores the height but not the clamped-away offset.

Worth recording since it is the obvious thing to reach for: **`forceMount` does not fix this.** A hidden `TabsContent` is `display: none`, contributes zero height, and clamps identically — no benefit, and a large diff mounted for as long as the card is open. Instead a small `useViewModeScrollAnchor` records the card's offset from the scroll root when a mode is left and restores it when that mode returns, in a layout effect plus one `requestAnimationFrame`, since virtualized rows settle to their real heights a frame after mounting. No change to `useSharedScrollRowVirtualizer`. (The anchor originally rode the tab strip; the last commit moves that strip into the title row, so it rides the card's own top edge instead — see below.)

## Cleanups

- `textOf` / `fenceOf` / the `language-*` pattern arrived as byte-for-byte copies of `ChatMarkdown`'s. Both feed the same highlighter, so language sniffing had two places to drift. One copy in `lib/markdown-fence.ts`.
- `code-theme.css` had every selector written twice, once per scope. `:is(.chat-code, .markdown-code)` states it once — identical matching, ~60 fewer lines. Also corrects `HighlightedCode`'s comment, which still named `.chat-code` as the only scope.
- Notes in `MarkdownFileView` that the `prefers-color-scheme` argument holds under Electron but not `npm run dev:web`.

## Tests

`markdown-image-resolver` is pure and does the fiddliest work in the feature; it arrived untested. Eleven cases: resolution against the file's directory, a file at the workspace root, `..` walking up, `..` clamping at the root rather than escaping, query/fragment stripping, absolute passthrough, and the version reaching the URL.

The tab pair itself had no coverage at all. Two cases now: the strip renders in the title row as a sibling of the trigger rather than a descendant, and survives the panel swap — which is the `asChild` ref composition actually working at runtime — and it stays off files that cannot render as markdown.

## Rename and redesign

`Source diff` / `Rendered` → **Diff** / **Preview**, strings renamed to `files.diff` / `files.preview` and retranslated for all eight locales.

The control then moved out of the body and into the file's own title row. It had a band to itself under the title, and that band held nothing else — a right-aligned control with the whole width of the card empty to its left. It now sits beside the `+N`/`-N` badges, on the line that already names the file it acts on, and the card is a row shorter.

`AccordionTrigger` already had a `trailing` slot for exactly this: rendered inside the header row but as a sibling of the trigger `<button>`, which is what makes it legal — a `<button>` inside a `<button>` is invalid, and a nested tab would collapse the file on click. The Tabs root moves up to enclose the title row as well as the body it switches; it and `AccordionItem` both take `asChild`, so they compose onto the same single `<li>`: no wrapper element, no change in DOM depth. The strip renders only while the file is expanded.

**This costs the anchoring premise of the scroll-position commit above.** That commit chose the tab strip deliberately, because it sat at the top of the swapped body and was an element the component owned outright, with no ref composition through `asChild` in the way. The strip no longer moves with the body, so the anchor rides the card's own top edge, through a ref composed across both `asChild` roots. Since that is precisely what the earlier commit avoided, it is now covered by tests rather than resting on a typecheck.

Appearance follows the inspector rail's own tab buttons (`session-inspector__tab-button`), which is the tab language of the pane this control lives in: `text-passive` labels, `--color-interactive-*` fills one step apart on the same neutral (foreground at 4% for the track, 7% for the selected tab), and a hairline border — GitHub's segmented-control signature, and what separates this shape from a shadcn pill group. Stock shadcn's palette is dropped, and worth flagging as an actual defect rather than taste: its dark-theme active fill is `bg-input/30`, white at 1.2% over the `bg-muted` track, which is not a visible selection at this size.

Sizing is a 20px track over 16px pills — deliberately under the 20px a standalone control would get. It rides in a 36px row it does not own, beside a 28px button, and reading as the smaller thing there is the point; the labels are wide enough to keep the targets easy despite the height. Narrowing keys off the `inspector` container the shell already declares, at the 350px/300px steps it already breaks at, so the toggle changes on the same drag positions as everything else in the pane. In the centre pane there is no `inspector` container to match, so it keeps full size.

Still a judgement call, and still droppable as its own commit if you disagree.

## Not addressed

- **`github-markdown-css` vs the design system.** `CLAUDE.md` asks for explicit approval before deviating. Your reasoning is sound and I have not touched it — flagging it as the repo owner's call. Side effect worth knowing: AO's nine theme styles (dracula, nord, gruvbox…) do not reach the Preview tab, which stays GitHub-coloured.
- **Lockfile churn.** The `motion` entries under `packages/product-ui` are unrelated to this feature, but they correct a stale lock on `main` (`packages/product-ui/package.json` already lists it there). Harmless — worth a line in the PR body so a reviewer does not stop on it.

## Verified

`npx tsc --noEmit` clean. Full renderer suite as of the tests commit: **2689 passed, 1 skipped, 204 files** — including your 35 `SessionFilesView` tests unmodified. Each commit typechecks on its own.

For the title-row commit on top I re-ran typecheck and `SessionFilesView` only (**37 passed**, the 35 plus the two new ones), not the full suite.

Manual verification in the Electron app is still worth doing for the scroll anchor and the image reload, which are the two that depend on real layout.

## Demo

Screen recording of the change in the desktop app: https://drive.google.com/file/d/1e5T3MbyQTVrNtLqfo2ZLPWgldhwUpDQO/view?usp=sharing


Vaibhaav-Tiwari (approved):
lgtm just resolve the merge conflicts

PR #4292, opened 2026-08-23
fix(frontend): clarify automatic update feedback

illegalcall:
Fixes #4291

## Summary

- let updater status own the card width instead of compressing it beside fixed-width actions
- explain that automatically discovered updates download in the background and remove redundant update/check actions from that state
- keep a compact `Download now` escape hatch for manual/request-owned updates
- show `Downloading… 0%` and disable repeat clicks immediately while a manual download invocation is pending
- add exact nightly-version geometry coverage and extend the renderer bridge fixture for updater state

## Root cause

The fixed 920px Settings dialog leaves a 678px update card, but the card used the viewport `sm` breakpoint to force status and actions into one row. The shrink-proof action group repeated the full nightly version and occupied 453px, leaving only 113px for status copy. Separately, the download button called IPC directly without renderer-local pending state, and the card did not receive the Automatic Updates setting, so it presented an automatic background operation as a required manual step.

## Reproduction evidence

The deterministic renderer reproduction uses the reported versions and the shipped 678px card width. On the pre-fix UI it failed with a four-line headline, three-line current version, 113px status column, and 453px action group.

Before: reproduced compressed update card

A pending `updates:download` unit reproduction also confirmed that IPC was called while the button remained enabled with unchanged copy. On the reported installed app, the complete 166,801,038-byte target nightly archive was already staged and ShipIt was waiting for app exit, so the updater was working while the UI appeared inert.

## Fix evidence

The same renderer fixture now gives status 584px, keeps both long version strings to one line, explains the automatic background download, and presents no redundant action.

After: readable automatic-update state

The manual/request-owned path remains available as `Download now`; its pending-state test asserts the button is immediately disabled and both button/status copy change to `Downloading… 0%` before updater progress arrives.

## Native Electron verification

The branch was also launched in the actual Electron app in isolated mode with the real preload bridge and daemon. Settings → Updates rendered at 678px, the daemon reported ready on `127.0.0.1:3002`, and Check for updates returned the expected unpackaged-app status.

Native Electron Settings verification

Electron disables release feeds in unpackaged builds, so the exact available-nightly state is covered by the deterministic renderer fixture; the native run verifies the real settings route, preload, daemon, and card structure.

## Tests

- `cd frontend && npm test` — 209 files passed; 2,748 tests passed, 1 skipped
- `cd frontend && npm test -- GlobalSettingsForm.test.tsx` — 30 passed
- `cd frontend && npx playwright test e2e/update-settings-ux.spec.ts --reporter=line` — passed
[...]

PR #4305, opened 2026-08-23
fix(frontend): detect fresh nightly releases promptly

### LINKED ISSUE #3938: Updates settings: no loading spinner on manual refresh + delayed detection of new versions
## Bug

Two UX problems in the **Settings → Updates** section:

1. **No loading indicator during manual refresh.** Clicking the circular-arrow refresh button to "Check for updates" frequently shows no spinner/progress. The refresh icon (`RefreshCw`) only swaps to the spinning `Loader2` when `status.state === "checking"` — but that state is transient and often missed.
2. **Stale update detection.** New nightly versions are not detected for a significant delay after release, even when manually clicking "Check for updates."

**Source:** Discord `#bug-triage` thread 1537350229001764884
**Reported by:** @prateek
**Analyzed against:** `9b8395e13` (main, Aug 13 2026)
**Confidence:** High (issue 1 — UI code path fully traced); Medium (issue 2 — multiple contributing factors, no single smoking gun)

**Screenshot:** Settings → Updates showing `v0.12.1-nightly.202608130248` with a static (non-spinning) refresh arrow and green "Downloaded. Restart to finish updating." text.

---

### Issue 1 — No loading indicator
1. Open Settings → Updates
2. Click the circular refresh arrow next to "Checks for Updates"
3. Observe: the icon often does not animate, and no "Checking for updates…" text appears before the result lands

### Issue 2 — Delayed update detection
1. Wait for a new nightly release to publish
2. Click "Check for updates" in Settings
3. Result: new version is not found immediately; may take minutes to hours to appear

---

### Issue 1 — Loader is ephemeral and races against IPC latency

The refresh button (`UpdatesSection.tsx:231-243`) shows a spinning `Loader2` only while `status.state === "checking"`:

```tsx
const checking = status.state === "checking";   // line 212
// ...
{checking ? (
    <Loader2 className="size-icon-base animate-spin" />   // line 239
) : (
    <RefreshCw className="size-icon-base" />              // line 241
)}
```

The `"checking"` state is broadcast by the main process **only after** `checkForUpdatesNow` enters `runSerializedUpdaterOperation` and calls `broadcastUpdaterStatus({ state: "checking" })` at `auto-updater.ts:656`. But before that broadcast fires, the main process must:

1. Enter the serialized operation queue (`updaterOperationQueue`) — if another operation is in flight (settings-write, retirement poll, automatic check), the manual check **waits silently**
2. Call `reconcileAndPersist()` → `reconcileFeaturePin()` → `collectFeatureBuilds()` which makes **unauthenticated GitHub API calls** (`/releases?per_page=100` + per-PR state checks) with no timeout on `fetch()` (`feature-builds.ts:103-113`)

The gap between the user's click and the `"checking"` broadcast can be **multiple seconds** (serialized queue wait + GitHub API latency). During that entire gap, the UI shows the **static** `RefreshCw` icon — no indication anything is happening.

Additionally, the `"checking"` state from `electron-updater`'s `checking-for-update` event is **very brief** — it fires, then immediately `update-available` or `update-not-available` fires, replacing the status. The `UpdateStatusLine` component does render `"Checking for updates…"` text (`line 270`), but it's in a separate row below the button that only appears when `showStatus` is true — and the window where `state === "checking"` may be shorter than a single React render cycle.

**The button itself has no independent loading state.** It relies entirely on the pushed status, creating a disconnect between the user's action (click) and the visual feedback.

### Issue 2 — Multiple caching/latency layers delay detection

Several factors stack:

1. **electron-updater `clientPromise` is memoized.** In `AppUpdater.js:382`, the GitHub provider client (`clientPromise`) is created once and cached. The `getLatestVersion()` call fetches the GitHub releases Atom feed (`.atom`), which itself is cached by GitHub's CDN. No `Cache-Control: no-cache` header is sent on the feed or yml manifest requests (confirmed: the only `no-cache` header in electron-updater@6.8.9 is in `MacUpdater.js` for the Squirrel download, not for the manifest/feed fetch).

2. **GitHub releases Atom feed CDN caching.** The `.atom` feed is served with GitHub's default caching. For nightly releases published minutes ago, the CDN may still serve a stale feed.

3. **`reconcileAndPersist` runs before every check.** Every `checkForUpdatesNow` calls `reconcileAndPersist` → `reconcileFeaturePin` (if no feature pin, returns immediately at `feature-builds.ts:222` — so this is fast for non-pinned users, but still an async hop). For users with auto-updates enabled, the hourly automatic check (`AUTOMATIC_UPDATE_CHECK_INTERVAL_MS = 60 min`) may have already consumed the `"checking"` state slot, and its result is silently suppressed for the UI (`auto-updater.ts:474-477`).

4. **`allowDowngrade: true` + nightly channel.** With `allowDowngrade` enabled (`auto-updater.ts:78`), the version comparison logic traverses the Atom feed entries to find a matching nightly prerelease tag. If the feed is stale, the new nightly tag is simply not present.

---

### Issue 1 — Loader
- **Optimistic immediate spinner.** When the user clicks the refresh button, set a local `isChecking` state immediately (before the IPC round-trip), then sync it with the pushed `status.state === "checking"` once events arrive. This eliminates the dead zone.
- **Or:** have the `"updates:check"` IPC handler broadcast `{ state: "checking" }` **before** entering the serialized queue, so the UI gets immediate feedback regardless of queue depth.

### Issue 2 — Stale detection
- **Add `Cache-Control: no-cache` to feed/manifest requests** via `autoUpdater.requestHeaders` (electron-updater supports custom headers on the provider). This bypasses CDN caching for the version-check requests.
- **Consider reducing the automatic check interval** for nightly users (currently 60 min) or adding a manual "force recheck" that bypasses caching.
- **Add a timeout to `fetchJson` in `feature-builds.ts`** — currently no timeout means a slow GitHub API response blocks the entire check indefinitely.

---

## Impact

- **Loader missing:** Users click refresh, see nothing happen, click again (which is a no-op because `disabled={busy}` doesn't engage since `busy` is false during the dead zone). Erodes trust in the update mechanism.
- **Stale detection:** Nightly users (the most engaged users) don't get prompt updates. Since nightlies cut daily, a multi-hour delay means some users are always a version behind. Combined with the missing loader, this creates a "the update checker is broken" perception.

---

## Related

- [#3526](https://github.com/Untrivial-ai/agent-orchestrator/issues/3526) — Persistent `net::ERR_FAILED` from wedged Chromium network stack (related network-level update failure, different mechanism)
- [#3288](https://github.com/Untrivial-ai/agent-orchestrator/issues/3288) — macOS release hardening: e2e update test, feed guard (broader update-reliability workstream)
- [#2707](https://github.com/Untrivial-ai/agent-orchestrator/issues/2707) — Update check: also run periodically, not only on app launch (closed — the hourly timer was added, but interval/caching tradeoffs remain)

illegalcall:
Fixes #3938

## Summary

- check the Nightly channel every 15 minutes while keeping Stable and feature builds on the existing hourly cadence
- make a user-requested Nightly check resolve the newest completed GitHub release directly instead of depending on the lagging prerelease Atom feed
- require the platform manifest asset before selecting a release, fall back to the existing provider on API/rate-limit/network failures, and restore the normal provider after the check
- reschedule the live timer when the user changes update channels

## Root cause

The reported app launched and checked at about 8:17 PM. `v0.12.7-nightly.202608231517` published at 9:01 PM, but Nightly shared Stable's one-hour timer, so the next automatic check was not due until 9:17 PM. The running app then detected the release on that exact tick and staged its 167,055,860-byte ZIP by 9:18:40 PM. Automatic detection was working, but its policy created a stale window of up to one hour.

Manual prerelease discovery was separately coupled to `electron-updater`'s `releases.atom` path. GitHub can expose a completed release and its manifest through the Releases API before that generated Atom view catches up. The existing issue's proposed missing `Cache-Control` cause is not present in `electron-updater@6.8.9`: `builder-util-runtime` already adds `Cache-Control: no-cache` to GETs, so adding the same header again would not fix server-side feed propagation.

The reported `…1350` build also predates #4285's manual-request status correlation, which explains why the click itself appeared inert. This PR addresses the remaining detection path.

## Reproduction evidence

The scheduling regression test used an enabled Nightly install. Before the fix:

```text
expected 3600000 to be 900000
```

That is the shipped 60-minute timer versus the required 15-minute Nightly timer.

The manual-check regression exposed a completed newer Nightly with the platform manifest. Before the fix, the updater made zero Releases API requests and remained dependent on Atom discovery:

```text
expected fetch mock to have been called
Number of calls: 0
```

Live evidence from the reported machine:

```text
running: 0.12.7-nightly.202608231350
published: 0.12.7-nightly.202608231517 at 21:01 IST
launch/initial check: ~20:17 IST
next hourly check: ~21:17 IST
staged: ~/Library/Caches/agent-orchestrator-updater/pending/Agent.Orchestrator-darwin-arm64-0.12.7-nightly.202608231517.zip at 21:18:40 IST
```

## Fix evidence

Both deterministic reproductions now pass. The manual fixture also proves an even newer but incomplete release (missing the platform manifest) is ignored.

A real Electron + `electron-updater@6.8.9` check then exercised the fixed direct completed-release route against the public release:

```text
Checking for update
Found version 0.12.7-nightly.202608231517 (url: Agent.Orchestrator-darwin-arm64-0.12.7-nightly.202608231517.zip, Agent.Orchestrator-darwin-x64-0.12.7-nightly.202608231517.zip)
PASS direct completed release 0.12.7-nightly.202608231350 -> 0.12.7-nightly.202608231517
```

Release: https://github.com/Untrivial-ai/agent-orchestrator/releases/tag/v0.12.7-nightly.202608231517

## Tests

- focused updater suite — 58 passed
- frontend CI Vitest suite — 209 files, 2,767 tests passed
- frontend and E2E TypeScript checks — passed
- renderer smoke — passed
- product UI verification — 10 files, 64 tests passed
- gitleaks scan — passed

### DIFF
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment node
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { chmodSync, mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import os from "node:os";
 import nodePath from "node:path";
 
@@ -28,6 +28,7 @@ type AutoUpdaterMock = {
   checkForUpdates: ReturnType<typeof vi.fn>;
   downloadUpdate: ReturnType<typeof vi.fn>;
   quitAndInstall: ReturnType<typeof vi.fn>;
+  setFeedURL: ReturnType<typeof vi.fn>;
   channel: string;
   allowPrerelease: boolean;
   allowDowngrade: boolean;
@@ -41,6 +42,7 @@ function createAutoUpdaterMock(): AutoUpdaterMock {
     checkForUpdates: vi.fn(() => Promise.resolve()),
     downloadUpdate: vi.fn(() => Promise.resolve()),
     quitAndInstall: vi.fn(),
+    setFeedURL: vi.fn(),
     channel: "",
     allowPrerelease: false,
     allowDowngrade: false,
@@ -202,7 +204,7 @@ describe("startAutoUpdates", () => {
     expect(autoUpdater.checkForUpdates).toHaveBeenCalledTimes(1);
   });
 
-  it("schedules the next automatic check only after the fixed 1-2 hour cadence", async () => {
+  it("keeps stable automatic checks on the hourly cadence", async () => {
     vi.useFakeTimers();
     const setIntervalSpy = vi.spyOn(globalThis, "setInterval");
     const { module, autoUpdater } = await importAutoUpdater();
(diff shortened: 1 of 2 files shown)


illegalcall:
## Reporter confirmation from the installed app

The reporter confirmed that `v0.12.7-nightly.202608231517` appeared **exactly one hour after the prior check** while running `v0.12.7-nightly.202608231350`.

The captured Settings state shows:

```text
Update available: v0.12.7-nightly.202608231517
Current version: v0.12.7-nightly.202608231350
Detection timing: exactly one existing 60-minute interval after the previous check
```

This independently corroborates the deterministic regression (`3,600,000ms` received versus the fixed `900,000ms` Nightly interval) and confirms the old client was delayed by scheduling policy rather than version comparison or a missing release artifact.

Because the screenshot comes from the pre-fix `…1350` client, its long version wrapping and redundant update action are the separate UI behavior already addressed by #4292; #4305 changes the detection cadence/manual feed path for subsequent builds.

illegalcall:
## Updated simplified UI evidence

These screenshots supersede the earlier tinted visual evidence. They show the latest UI from #4292 alongside #4305's fresh-nightly detection behavior; #4305 itself remains scoped to update detection.

### Native Electron Settings view

The update status now uses the same neutral panel surface and border as the existing **Automatic Updates** controls—no full-card tint, gradient, dot texture, glow, or colored border.

Native Electron Settings view with simplified update status

### Automatic update available

With Automatic Updates enabled, the UI explains that the update downloads in the background and does not show a redundant update/check button.

Automatic update available with neutral styling

### Downloaded and ready to install

There is one clear restart action, structured version metadata, and no orphan separator dot.

Downloaded update ready to install

### Latest version

Manual refresh remains available, while the metadata stays aligned and has no decorative trailing separator.

Latest version state with simplified neutral styling

### Verification

- Real native Electron shell verified with an isolated AO profile and daemon.
- Computed status-card background and border match the adjacent controls panel; no background image or pseudo-element texture remains.
- Component evidence suite: **30 passed**.
- Focused updater E2E suite: **3 passed**.
- Frontend and E2E typechecks passed.
- All GitHub checks on the latest #4292 commit passed.

UI commit: Untrivial-ai/agent-orchestrator@605e0f0a0

codebanditssss (approved):

PR #4306, opened 2026-08-23
Reveal worker Chat after sending review feedback

### LINKED ISSUE #4076: [Feature]: Add Chat UI to reviewer
### Problem

Add Chat UI to reviewer

### Proposed direction

_No response_

### Alternatives considered

_No response_

### Before submitting

codebanditssss:
Code changes: +1804 -256  
Tests: +547 -10  
Others: +1228 -84

## Summary

- Adds durable reviewer-owned chat conversations, recovery, controller routing, and a read-only/auto-permission reviewer bridge.
- Exposes the reviewer conversation API and generated OpenAPI/frontend client schema.
- Keeps the worker Chat surface usable while a reviewer chat is selected.
- Decodes provider request IDs consistently and covers every reviewer conversation HTTP boundary.
- Reconciles a selected reviewer Chat when the backend replaces or removes that reviewer, returning both Chat and TUI workers to their own surface.
- Makes reviewer-chat recovery persistence a required engine dependency, so wiring drift fails at compile time instead of silently skipping startup recovery.

## Verification

- `go test ./...`
- Focused reviewer/service/store suites
- Reviewer-route and escaped-request-ID controller tests
- `golangci-lint run --path-mode=abs` (0 issues on the prior backend head; final full Go suite passes)
- Frontend typecheck
- Full SessionView suite: 135 passed
- Focused CenterPane reviewer/workspace tests

Fixes #4076


neversettle17-101:
On trigger, we should spin up Chat UI reviewer if agent harness is chat UI. Test all such end cases that make sure reviewer is also chat UI.
Screenshot 2026-08-24 at 10 54 38 PM

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -1200,35 +1227,47 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
+										shellError={
+											openShellTerminal.error ? apiErrorMessage(openShellTerminal.error) : undefined
+										}
+										onOpenFiles={handleOpenFiles}
+										onOpenFile={handleOpenFile}
+										workspaceTabs={centerFileTabs}
+										workspaceFileActive={Boolean(fileTabs.activePath)}
+									/>
+									{reviewerChatId ? (
+										<div className="absolute inset-0">
```
neversettle17-101, line 1602: This keeps the worker Chat mounted, but it also leaves it fully interactive under the reviewer overlay. Because the wrapper is only marked `inert` for file tabs, keyboard focus and shortcuts can still hit the hidden worker composer while the reviewer chat is selected, and screen readers will still see both surfaces. Please disable or inert the worker surface while the reviewer overlay owns the body.

backend/internal/review/review.go
```diff
@@ -521,6 +547,30 @@ func (e *Engine) RestoreReviewer(ctx stdctx.Context, workerID domain.SessionID)
...
+	if !ok {
+		return nil
+	}
+	reviews, err := store.ListRecoverableChatReviews(ctx)
+	if err != nil {
+		return err
+	}
+	var recoveryErrors []error
+	for _, review := range reviews {
+		if _, restoreErr := e.RestoreReviewer(ctx, review.SessionID); restoreErr != nil {
```
neversettle17-101: Recovery is iterating specific review rows here, but it restores by `review.SessionID` rather than by the review row itself. `RestoreReviewer` re-derives the harness from the workers current preference/project config, so if the selected reviewer changed after this run started we can resurrect the wrong reviewer or skip the chat review that actually has the running `review_run`. This recovery path should target the concrete review/harness being recovered, not whatever the session currently prefers.

frontend/src/renderer/hooks/useReviewerConversation.ts
```diff
@@ -0,0 +1,87 @@
...
+export function reviewerConversationQueryKey(reviewId: string) {
+	return ["reviewer-conversation", reviewId] as const;
+}
+
+export function useReviewerConversation(reviewId: string | undefined) {
+	const query = useQuery({
+		queryKey: reviewerConversationQueryKey(reviewId ?? ""),
+		enabled: Boolean(reviewId),
+		queryFn: async () => {
+			const { data, error } = await apiClient.GET("/api/v1/reviews/{reviewId}/conversation", {
```
neversettle17-101, line 25: Reviewer conversations are capped to the newest 200 items here, but this hook does not expose any pagination state and `ReviewerChatSurface` never passes `hasOlder`/`onLoadOlder` into `ChatWorkspace`. That means long reviewer chats resumed after recovery will silently lose access to older history, even though the backend added paged reviewer snapshot routes. Please wire reviewer conversations through the same paged-history flow as worker chat.

frontend/src/renderer/components/SessionView.tsx
codebanditssss: Verified the worker conversation body is already isolated inside ChatWorkspace when Reviewer is selected: reviewerChatSelected sets reviewerActive, and chat-conversation-panel is hidden, aria-hidden, and inert. The shared Chat/Reviewer tab strip intentionally remains interactive so the user can return to Chat. Added explicit regression assertions for all three accessibility states in 11a83344a.

backend/internal/review/review.go
codebanditssss: Fixed in 11a83344a. Recovery now restores the concrete review row returned by ListRecoverableChatReviews, uses that row’s harness/review ID/provider conversation, and does not destroy an unrelated currently selected reviewer. Added a preference-change regression plus controller-generation claim coverage.

frontend/src/renderer/hooks/useReviewerConversation.ts
codebanditssss: Fixed in 11a83344a. Reviewer history now uses useInfiniteQuery with beforeSequence paging and the shared page merge logic. ReviewerChatSurface passes hasOlder/loadingOlder/onLoadOlder to ChatWorkspace, with hook and UI regressions covering the second request, merged ordering, and Load earlier messages.

codebanditssss:
Addressed the latest inline review feedback in `11a83344a`.

Tested flows:
- Selecting Reviewer keeps the worker Chat mounted but hides its conversation body with `hidden`, `aria-hidden`, and `inert`; the shared Chat tab remains usable to return to the main chat.
- Reviewer chat renders through the typed chat UI and exposes **Load earlier messages** when older history exists.
- Loading older reviewer history sends the correct `beforeSequence`, merges pages oldest-to-newest, and stops when history is exhausted.
- Daemon restart recovery restores the exact persisted reviewer row even if the current reviewer preference changed.
- Reviewer recovery preserves the provider conversation ID and controller-generation claim made during terminal restore.
- Recovering one reviewer does not destroy an unrelated currently selected reviewer.

Verification:
- Focused reviewer backend tests passed.
- Focused frontend chat/reviewer tests passed: 168 tests.
- Full frontend suite passed: 2,988 passed, 1 skipped.
- Frontend typecheck passed.
- Full Go tests and golangci-lint passed with 0 issues.
- Go build, formatting check, and `git diff --check` passed.
- All GitHub Actions checks are green.

frontend/src/renderer/hooks/useReviewerConversation.ts
```diff
@@ -0,0 +1,106 @@
...
+					},
+				},
+			});
+			if (error) throw error;
+			return toSnapshot(data as WireSnapshot);
+		},
+		getNextPageParam: (page) => (page.hasMoreBefore ? page.oldestSequence : undefined),
+		select: (data) => mergeConversationPages(data.pages),
+		refetchInterval: (query) => {
+			const snapshot = (query.state.data as { pages?: ConversationSnapshot[] } | undefined)?.pages?.[0];
```
neversettle17-101: Because this query uses `select`, `query.state.data` here is already the merged `ConversationSnapshot`, not the raw infinite-query `{ pages }` structure. That means `snapshot` stays `undefined` and reviewer chats always poll at 2s, even while the controller is busy. The user-visible effect is slower streaming / approval-state refresh than the worker chat. Reading `query.state.data?.controller.state` directly (or deriving the interval before `select`) would preserve the intended fast poll while busy.

backend/internal/daemon/lifecycle_wiring.go
```diff
@@ -486,6 +489,67 @@ func (c chatLauncher) PreflightChat(
...
+}
+
+func (c chatLauncher) RestoreReviewChat(ctx context.Context, cfg reviewcore.ReviewerChatStart) (string, error) {
+	return c.startReviewChat(ctx, cfg, false)
+}
+
+func (c chatLauncher) startReviewChat(ctx context.Context, cfg reviewcore.ReviewerChatStart, sendPrompt bool) (string, error) {
+	owner := domain.ReviewConversationOwner(cfg.ReviewID)
+	started, err := c.svc.StartChat(ctx, chatsvc.StartRequest{
+		Owner: owner, SessionID: cfg.WorkerID, ProjectID: cfg.ProjectID,
```
illegalcall, line 547: This passes the parent worker as SessionID for the review-owned controller. The shared thread-title path later calls ApplyProviderTitle with c.sessionID, so a provider title emitted by this reviewer can rename the worker session whenever its display name is blank or AO-applied. I reproduced that against the real SQLite store. Please make title projection owner-aware, or skip session title application for review owners.

backend/internal/storage/sqlite/queries/review.sql
```diff
@@ -1,29 +1,88 @@
...
+    updated_at = sqlc.arg(updated_at)
+WHERE id = sqlc.arg(id) AND interface_mode = 'chat';
+
+-- name: ListRecoverableChatReviews :many
+SELECT DISTINCT r.*
+FROM review r
+JOIN review_run rr ON rr.review_id = r.id
+WHERE r.interface_mode = 'chat'
+  AND r.provider_conversation_id != ''
+  AND rr.status = 'running'
```
illegalcall: This only recovers Chat reviewers that still have a running review_run. A completed reviewer conversation remains durable after restart, but SnapshotForReview reports its controller as stopped and ReviewerChatSurface has no resume action, so follow-up conversation is no longer possible. If completed reviewer chats are meant to remain reachable, recover every eligible Chat review with a provider conversation or add an explicit resume path.

frontend/src/renderer/hooks/useReviewerConversation.ts
```diff
@@ -0,0 +1,107 @@
...
+			if (error) throw error;
+			return toSnapshot(data as WireSnapshot);
+		},
+		getNextPageParam: (page) => (page.hasMoreBefore ? page.oldestSequence : undefined),
+		select: (data) => mergeConversationPages(data.pages),
+		refetchInterval: (query) => {
+			// TanStack passes the cache Query here. `select` transforms the observer
+			// result, while cache state retains the raw InfiniteData page structure.
+			const snapshot = query.state.data?.pages[0];
+			return snapshot?.controller.state === "busy" ? 500 : 2_000;
```
illegalcall: Reviewer history is polling every 500 ms or 2 s, while docs/STATUS.md documents Chat history as using targeted CDC/SSE invalidation rather than polling. The event transport only invalidates the session conversation query root, so this new reviewer query key is not event-driven. Please route review-owned conversation changes through the existing invalidation stream and remove the interval polling.

backend/internal/httpd/controllers/conversations.go
```diff
@@ -86,6 +95,138 @@ func (c *ConversationsController) Register(r chi.Router) {
 	r.Post("/sessions/{sessionId}/conversation/branches/{branchId}/activate", c.activateBranch)
 	r.Put("/sessions/{sessionId}/conversation/title", c.setTitle)
 	r.Post("/sessions/{sessionId}/conversation/mcp/reload", c.reloadMCPServers)
+	r.Get("/reviews/{reviewId}/conversation", c.reviewSnapshot)
+	r.Post("/reviews/{reviewId}/conversation/messages", c.reviewSend)
+	r.Post("/reviews/{reviewId}/conversation/approvals/{requestId}/resolve", c.reviewResolve)
+	r.Post("/reviews/{reviewId}/conversation/inputs/{requestId}/resolve", c.reviewResolveInput)
+	r.Post("/reviews/{reviewId}/conversation/interrupt", c.reviewInterrupt)
```
illegalcall, line 104: These five user-facing reviewer conversation routes are added without controller tests. AGENTS.md requires changed HTTP boundaries to cover happy paths, validation or missing arguments, and daemon error envelopes. Please add route-level tests for snapshot paging, send, approval/input validation, interrupt, and mapped service failures.

illegalcall (changes requested):
Re-reviewed at 6e6a3ec after the latest fixes. The typed-owner startup failure from the earlier pass is fixed. I am requesting changes for the remaining ownership, recovery, invalidation, and boundary-test gaps called out inline.

backend/internal/daemon/lifecycle_wiring.go
codebanditssss: Fixed in 70e4cbd95. Review-owned controllers still persist the provider title on their own conversation, but applyThreadTitle now skips projecting it onto the parent worker session. Added a real-SQLite regression proving the reviewer title is stored while the worker display name remains unchanged.

frontend/src/renderer/hooks/useReviewerConversation.ts
codebanditssss: Fixed in 70e4cbd95 and 830c93340. Review-owned conversation CDC payloads now include reviewId; event transport invalidates the targeted reviewer-conversation key and the reviewer root on reconnect; interval polling is removed. A follow-up trigger also publishes provider-title-only reviewer updates. Migration, transport, reconnect, and no-poll regressions cover the flow.

frontend/src/renderer/hooks/useReviewerConversation.ts
codebanditssss: The original callback was valid because TanStack Query passes the cache Query and query.state.data retains raw InfiniteData before select. This path is now superseded: 70e4cbd95 removes reviewer polling entirely and routes updates through targeted CDC/SSE invalidation.

backend/internal/storage/sqlite/queries/review.sql
codebanditssss: Fixed in 70e4cbd95. ListRecoverableChatReviews now selects every Chat review with a non-empty provider conversation independently of review_run status. Added store coverage for a completed review run and retained concrete-review recovery coverage.

backend/internal/httpd/controllers/conversations.go
codebanditssss: Fixed in 70e4cbd95. Added route-level tests for reviewer snapshot paging, send, approval resolution, structured input resolution, and interrupt. Coverage includes happy paths, cursor/limit validation, empty messages, missing decisions, invalid input action/content, and mapped service errors with request IDs.

i-trytoohard:
@codebanditssss this looks like it addresses #4076 — linking for tracking. Adding `Fixes #4076` to the PR body will auto-close it on merge.

illegalcall:
@codebanditssss Still working on this?

neversettle17-101:
@illegalcall i'm taking over and will get this merged asap

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -1528,45 +1555,57 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
+										onOpenFiles={handleOpenFiles}
+										onOpenFile={handleOpenFile}
+									/>
+									{reviewerChatId ? (
+										<div className="absolute inset-0">
+											<ReviewerChatSurface hideHeader reviewId={reviewerChatId} />
+										</div>
+									) : null}
+								</>
+							) : reviewerChatId ? (
```
neversettle17-101: When the worker is not in Chat mode, selecting a reviewer Chat takes this branch and replaces the entire CenterPane with a standalone ReviewerChatSurface. Unlike the Chat-mode branch above, the worker surface/tab strip is no longer mounted, and ReviewerChatSurface is not given an onSelectChat path, so a TUI worker loses the worker terminal/shell tabs with no in-pane way back. Since reviewer Chat launch is gated by reviewer harness support rather than worker session.mode, this is reachable from TUI workers via ReviewsSection opening reviewerSurface.mode === chat. Consider keeping CenterPane mounted and rendering the reviewer Chat as an auxiliary tab/overlay there, or otherwise provide a return path for non-Chat workers.

neversettle17-101 (commented):
## What This PR Does

Adds durable reviewer-owned Chat conversations end-to-end: backend reviewer launch/recovery and conversation APIs, SQLite schema/query support, generated API types, and frontend reviewer Chat surfaces/navigation.

## Summary of Changes

- Introduces conversation owners/scopes for review rows and persists reviewer provider ids/controller state.
- Adds /reviews/{reviewId}/conversation routes plus store/query/spec/schema support for reviewer Chat snapshots and commands.
- Launches supported reviewer harnesses through the Chat controller, with recovery after daemon restart and fallback/surface metadata for the UI.
- Updates SessionView/Inspector/ChatWorkspace to show reviewer Chat tabs and send review feedback back to worker Chat.
- Adds backend and frontend regression coverage plus migration/spec/type generated artifacts.

## Findings

- **Medium**: Opening a Chat reviewer from a non-Chat worker replaces the CenterPane with a standalone ReviewerChatSurface, leaving no worker/TUI tab or callback to return to the worker surface. The backend can launch reviewer Chat based only on reviewer harness support, and the inspector opens it whenever reviewerSurface.mode is chat, regardless of the worker session mode. In SessionView, the non-Chat branch renders only ReviewerChatSurface while reviewerChatId is set, instead of keeping CenterPane mounted with the reviewer-chat tab. This regresses the TUI worker flow and makes the worker terminal/shell tabs disappear until another external navigation clears reviewerChatId.

## Regression Risk

Main risk is mixed-mode navigation: typed reviewers are now possible independently of the worker surface, so TUI workers need the same tab-strip/return behavior as Chat workers. Backend persistence/migration risk is moderate because this changes ownership columns, CDC triggers, and recovery behavior, but the diff includes focused tests around those paths.

## Testing

I did not run tests. The PR includes backend controller/store/review launcher/recovery tests, migration tests, generated API updates, and frontend tests for reviewer conversation hooks/surfaces/navigation. The missing coverage appears to be the TUI-worker + reviewer-Chat selection path described above.

neversettle17-101 (commented):
## What This PR Does

Adds durable reviewer-owned Chat conversations end-to-end: backend reviewer launch/recovery and conversation APIs, SQLite schema/query support, generated API types, and frontend reviewer Chat surfaces/navigation. The latest update keeps the terminal CenterPane/tab strip around when a reviewer Chat is selected from a non-Chat worker.

## Summary of Changes

- Introduces conversation owners/scopes for review rows and persists reviewer provider ids/controller state.
- Adds /reviews/{reviewId}/conversation routes plus store/query/spec/schema support for reviewer Chat snapshots and commands.
- Launches supported reviewer harnesses through the Chat controller, with recovery after daemon restart and surface metadata for the UI.
- Updates SessionView/Inspector/ChatWorkspace/CenterPane to show reviewer Chat tabs, keep worker surfaces mounted, and send review feedback back to worker Chat.
- Adds backend and frontend regression coverage plus migration/spec/type generated artifacts.

## Findings

No blocking findings from the reviewed diff. Ready to merge from my review.

## Regression Risk

Residual risk is moderate because this spans persistence, recovery, API shape, and mixed Chat/TUI navigation. The prior mixed-mode navigation concern appears addressed by rendering reviewer Chat inside CenterPane with a selected reviewer tab and worker tab return path.

## Testing

I did not run tests. The PR includes backend controller/store/review launcher/recovery tests, migration tests, generated API updates, frontend hook/surface/navigation tests, and a new CenterPane regression covering reviewer Chat in the existing terminal tab strip. The PR description also reports lint, Go build, frontend suite/typecheck, and focused reviewer/session UI suite passing.

neversettle17-101:
@illegalcall can you review this from Chat UI lens? I have reviewed the changes lgtm. 
Half of the code changes  are because of tests and sqlc/openAPI generated files

backend/internal/httpd/apispec/specgen/build.go
```diff
@@ -926,6 +926,31 @@ func shellTerminalOperations() []operation {
 				{http.StatusNotImplemented, envelope.APIError{}},
 			},
 		},
+		{
+			method: http.MethodGet, path: "/api/v1/reviews/{reviewId}/conversation", id: "getReviewerConversation", tag: "conversations",
+			summary: "Read a reviewer's durable Chat conversation", pathParams: []any{controllers.ReviewIDParam{}, conversationSnapshotQuery{}},
+			resps: []respUnit{{http.StatusOK, controllers.ConversationSnapshotResponse{}}, {http.StatusNotFound, envelope.APIError{}}, {http.StatusConflict, envelope.APIError{}}, {http.StatusInternalServerError, envelope.APIError{}}},
```
illegalcall, line 932: `reviewSnapshot` returns `400 CONVERSATION_CURSOR_INVALID` and `CONVERSATION_LIMIT_INVALID` for malformed `beforeSequence`/`limit`, while `reviewService` can return `501 Not Implemented` when reviewer-chat support is unavailable. This operation currently declares only 200/404/409/500; the other reviewer conversation operations likewise omit 501.\n\nPlease declare the errors the handlers can actually return (at minimum 400 for this GET and 501 for every capability-gated reviewer route), then run `npm run api` so `openapi.yaml` and `frontend/src/api/schema.ts` stay accurate.

illegalcall (changes requested):
The reviewer-conversation endpoints expose errors that the generated OpenAPI contract does not declare. Please add the missing response statuses in the operation registry and regenerate the spec and frontend schema.

backend/internal/daemon/daemon.go
```diff
@@ -854,6 +854,11 @@ func Run() error {
 			if reconcileErr := sessMgr.ReconcileBackground(ctx); reconcileErr != nil {
 				log.Error("background session reconciliation on boot failed", "err", reconcileErr)
 			}
+			if recovery, ok := reviewSvc.(interface{ RecoverChatReviewers(context.Context) error }); ok {
```
illegalcall, line 857: This anonymous optional interface means recovery is silently skipped if `reviewSvc` no longer matches this exact method signature. Since this PR promises reviewer-conversation recovery after a daemon restart, please call the concrete service method directly (or add an explicit recovery dependency to the daemon wiring) so an interface drift fails at compile time rather than disabling recovery at runtime.

illegalcall (dismissed):
Reviewer-chat recovery must be a required startup responsibility, not an optional capability check.

codebanditssss:
rechecked the current head (`8ade053e8`) against the outstanding review findings. the pr is not ready to merge yet:

- `backend/internal/httpd/apispec/specgen/build.go:932-952` still omits the `400` responses emitted by `reviewSnapshot` and the `501` capability-gate response emitted by every reviewer-conversation route. regenerate the openapi spec and frontend schema after declaring those statuses.
- `backend/internal/daemon/daemon.go:857` still uses an optional anonymous interface assertion for reviewer-chat recovery. if the service wiring drifts, startup silently skips recovery. make recovery an explicit startup dependency or include it in the required service contract.
- github reports the branch as conflicting/dirty; main is 131 commits ahead, so rebase and rerun the full checks after resolving conflicts.

all recorded ci checks for the current head are green, but they do not clear these contract, startup, and integration blockers. please add `fixes #4076` to the pr body when updating it so the issue closes on merge.

backend/internal/httpd/apispec/specgen/build.go
codebanditssss: Addressed in the forward-port: reviewer routes now declare the emitted 400 and capability-gated 501 responses in the operation registry; OpenAPI and frontend schema were regenerated.

backend/internal/daemon/daemon.go
codebanditssss: Addressed in the forward-port: reviewer Chat recovery is now a required review-service method invoked directly during daemon startup, so wiring drift fails at compile time instead of silently skipping recovery.

codebanditssss:
Forward-ported the reviewer Chat implementation onto current main and force-updated this PR branch. Addressed the API status-contract and required-startup-recovery findings, restored typed owner persistence/recovery, and kept the worker UI available while Reviewer is selected. The PR description now has refreshed diff counts and `Fixes #4076`.

codebanditssss:
The failing lint job was traced to 14 missing GoDoc comments on newly exported reviewer-chat symbols. I added the documentation in edf974b19 and pushed a replacement CI run; local golangci-lint now reports 0 issues.

backend/internal/storage/sqlite/store/conversation_store.go
```diff
@@ -80,6 +80,44 @@ func (s *Store) OpenNativeConversation(ctx context.Context, id string, scope dom
...
+		return domain.ConversationRecord{}, fmt.Errorf("select conversation for review %s: %w", reviewID, err)
+	}
+
+	rootBranchID := id + ":root"
+	if err := s.inTx(ctx, "insert reviewer conversation", func(q *gen.Queries) error {
+		review, err := q.GetReviewByID(ctx, reviewID)
+		if err != nil {
+			return fmt.Errorf("select reviewer %s: %w", reviewID, err)
+		}
+		if review.SessionID != session || review.ProjectID != project || review.InterfaceMode != string(domain.ReviewerInterfaceChat) {
```
illegalcall: [P1] This guard makes every first reviewer Chat launch fail. Trigger builds the review row without InterfaceMode, and Store.UpsertReview normalizes that zero value to tui. SetReviewInterfaceMode has no call site, so StartReviewChat reaches CreateReviewConversation with a tui row and returns "not the current chat owner". Persist chat mode before starting the controller and cover the trigger-to-conversation path.

backend/internal/daemon/lifecycle_wiring.go
```diff
@@ -540,6 +543,59 @@ func (c chatLauncher) PreflightChat(
...
+	return c.startReviewChat(ctx, cfg, true)
+}
+
+func (c chatLauncher) RestoreReviewChat(ctx context.Context, cfg reviewcore.ReviewerChatStart) (string, error) {
+	return c.startReviewChat(ctx, cfg, false)
+}
+
+func (c chatLauncher) startReviewChat(ctx context.Context, cfg reviewcore.ReviewerChatStart, sendPrompt bool) (string, error) {
+	owner := domain.ReviewConversationOwner(cfg.ReviewID)
+	started, err := c.svc.StartChat(ctx, chatsvc.StartConfig{Owner: owner, SessionID: cfg.WorkerID, ProjectID: cfg.ProjectID, Kind: domain.KindWorker, Harness: cfg.Harness, DataDir: cfg.DataDir, WorkspacePath: cfg.WorkspacePath, Env: cfg.Env, Permissions: ports.PermissionModeAuto, SystemPrompt: cfg.SystemPrompt, ProviderConversationID: cfg.ProviderConversationID})
```
illegalcall, line 564: [P1] This start config does not isolate or sandbox the reviewer. SessionID is the worker id, while the shipped Codex persistent host is keyed by SessionID; a live Chat worker therefore collides with its reviewer and fresh Start rejects the surviving host. Also, PermissionModeAuto maps to the Codex workspace-write sandbox, whereas the terminal reviewer explicitly passes --sandbox read-only. Use a review-owned host identity and an enforced read-only sandbox so reviewer Chat can coexist without modifying the PR worktree.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -65,6 +65,10 @@ type CenterPaneProps = {
 	terminalTarget?: TerminalTarget;
 	reviewerTerminal?: { handleId: string; harness: string };
 	onSelectReviewerTerminal?: (target: { handleId: string; harness: string }) => void;
+	reviewerChat?: { reviewId: string; harness: string };
```
illegalcall, line 68: [P1] These reviewer Chat props have no production caller. SessionView still derives only reviewerHandleId, treats the synthetic review-chat:<id> value as a terminal handle, and passes reviewerTerminal to CenterPane/SessionChatSurface. No renderer code consumes the new reviewer conversation routes. As a result the Chat tab/body can only appear in this component test, not in the app. Expose the reviewer surface/review id through review responses and wire the real SessionView flow.

backend/internal/service/chat/service.go
```diff
@@ -834,7 +866,10 @@ func (s *Service) Start(ctx context.Context, cfg StartConfig) (*Controller, erro
 		cfg.ExpectedControllerOwner.ControllerGeneration = controller.Generation()
 	}
 	s.mu.Lock()
-	s.controllers[cfg.SessionID] = controller
+	s.ownerControllers[owner] = controller
```
illegalcall, line 882: [P1] Typed ownership is only applied to the controller registry. startConfigs is still keyed by cfg.SessionID, so starting a reviewer for a Chat worker overwrites that worker configuration with reviewer env, system prompt, and harness; StopAll likewise walks only s.controllers and never closes review owners. Key all controller-owned lifecycle state and shutdown traversal by ConversationOwner.

illegalcall (commented):
Changes requested.

## Standards

- The five new reviewer-conversation routes and the launch/recovery path have no backend behavioral coverage. The current backend test edits only update a fake and the burned-migration list, despite AGENTS.md requiring happy-path, validation, and daemon-error-envelope coverage for changed boundaries.
- Typed ownership is not propagated through all lifecycle state; see the inline comment.

## Spec

- Reviewer Chat is rejected before its first conversation can be created.
- The launch shares the worker persistent-host identity and uses workspace-write instead of the promised read-only sandbox.
- The renderer additions are not wired into SessionView, so no user can open the new Chat UI.

I did not run project code or tests per AO review policy; findings are from `git diff 99a793645ecc1cef3bd21d32b6a31fba76fb142c...3ce74a89bba51ab0267c75318380fc82d1771dc2`.

Summary: Standards 2 findings (worst: missing boundary coverage); Spec 3 findings (worst: reviewer Chat cannot launch).

backend/internal/daemon/lifecycle_wiring.go
illegalcall: Fixed in 7ac11f663 and 8d2ac2b97. Reviewer providers now use a review-owned host identity; the Chat service forces review owners read-only, and Codex enforces its native read-only sandbox at both thread and turn boundaries. Regression coverage checks the isolated host, read-only launch, and shutdown.

frontend/src/renderer/components/CenterPane.tsx
illegalcall: Fixed in d36e97b64 and 990494cec. Review responses expose the typed reviewer surface, SessionView/SessionInspector route the review ID into CenterPane, and the reviewer conversation hook/surface now consumes the production routes. The return-to-worker handoff is covered by the focused UI test and local Electron evidence.

backend/internal/service/chat/service.go
illegalcall: Fixed in 8d2ac2b97. Controller gates and start configs are keyed by ConversationOwner, StopForOwner clears typed lifecycle state, and StopAll traverses every owner controller instead of only session controllers. The regression proves a review-owned controller is closed during shutdown.

illegalcall:
Local Electron verification on the isolated PR build:

1. Triggering **Review latest commit** opens the durable reviewer Chat while keeping the worker tab available.
2. Sending review feedback through **Send to worker agent** returns to the worker Chat; the action changes to **Sent to worker agent**.

Verification: frontend typecheck passed; focused UI tests passed (181/181); focused backend review/controller suites passed; all 14 GitHub checks pass.

Reviewer Chat opens after triggering review

Worker Chat is revealed after sending review feedback

codebanditssss:
Completed the final review pass in `43baa3d8f`.

- Added route-level behavioral coverage for all five reviewer conversation endpoints: snapshot paging, send, approval resolution, structured-input resolution, and interrupt.
- Added validation coverage for cursor/limit bounds, empty messages, missing decisions, invalid input actions, and invalid input content.
- Added mapped service-failure assertions, including stable API error codes and request IDs.
- Found and fixed an additional boundary bug during that coverage pass: reviewer approval/input routes now decode percent-encoded ACP request IDs just like worker Chat routes.

Local verification passed: full `go test ./...`, the focused reviewer/service/store suites, and `golangci-lint` with 0 issues. The required change-count header has been recalculated from the current GitHub diff. Fresh CI is running.

codebanditssss:
Addressed the final independent-review blocker in `c6d9cd270`. The selected reviewer Chat is now reconciled against the current reviewer surface after review-query updates. If reviewer A is replaced by B, switches to TUI, or disappears, AO clears the stale selection instead of continuing to render A.

Regression coverage exercises both worker modes:
- Chat worker: reviewer A → reviewer B returns to worker Chat.
- TUI worker: reviewer removal returns to the worker terminal.

Frontend typecheck passes and the complete SessionView suite passes (135/135). The verified count header is refreshed and fresh CI is running.

codebanditssss:
Closed the remaining startup-contract blocker in `ebc651c24`. `ListRecoverableChatReviews` and `RecordReviewChatControllerError` are now required members of the review engine’s core `Store` interface. `RecoverChatReviewers` calls them directly; the optional type assertion and silent success path are gone, so a production wrapper or signature drift now fails compilation.

The review fake implements the required contract, and a regression proves a recovery-query failure is propagated instead of being treated as successful startup. Full `go test ./...` passes on this head. The verified count header has been recalculated and fresh CI is running.

illegalcall:
add final evidence of the feature working

codebanditssss:
Confirmed the reviewer-to-worker Chat handoff in the worker conversation.

The AO reviewer feedback arrived in the worker Chat as:

> Desktop demo feedback: confirm the reviewer-to-worker Chat handoff in the worker conversation.

The worker received and acknowledged that feedback, confirming the handoff returned from Reviewer Chat to the original worker Chat. Focused verification also passes: `SessionView.test.tsx` — 135/135 tests.

No code change was required for this evidence-only review request.

codebanditssss:
> add final evidence of the feature working


https://github.com/user-attachments/assets/49cc279a-6d86-4743-8761-05215ee0aa23

illegalcall (approved):

PR #4309, opened 2026-08-23
feat(webui): serve the LAN web UI with browser sessions

execsumo:
Adds web-session authentication, optional trusted-tailnet-identity auth, and
embedded SPA serving to the existing opt-in LAN listener, so a browser can be a
first-class client of the daemon. Everything here is off by default and upstream
behaviour is unchanged when it is not configured.

### Why this shape

The LAN listener already ships auth, per-source lockout, control-route blocking,
and `tailscale serve` TLS. What was missing was narrow: nothing serves a web UI,
and a browser cannot carry `Authorization: Bearer` on a WebSocket upgrade, on an
`EventSource`, or on the top-level HTML navigation that loads the app. React
Native gets away with it because its `WebSocket` accepts a `headers` option; a
browser has no equivalent.

So this serves the SPA from the listener that already exists, and adds two
credentials alongside Bearer — a **web session cookie**, and a **tailnet identity
header** injected by `tailscale serve` when explicitly enabled. Bearer is
untouched for native clients.

Alternatives rejected:

- **A second listener or a reverse proxy in front of the daemon.** `AGENTS.md`'s
  hard rule is that there is exactly one network-facing bind; adding another
  duplicates the auth surface and gives an attacker a second door. This adds no
  new bind — `AO_CONNECT_BIND_HOST` makes the existing one's host configurable
  and defaults to `0.0.0.0`, so nothing changes unless an operator opts in.
- **Query-parameter tokens for WebSocket/SSE.** `tailscale serve` strips query
  parameters from upgrade requests, so it cannot work on the deployment this is
  built for — and it puts credentials in log lines besides.
- **A new browser client built on `product-ui`.** The existing renderer already
  has a bridge abstraction with a browser fallback; a second client would need
  its own copy of the board, terminals, chat, and orchestrator surfaces.

### Three load-bearing rules, each with tests

1. Cookie- and identity-authenticated requests are **CSRF-checked**. Bearer is not,
   because a native client is not a browser and has no ambient credential.
2. Cross-origin `/mux` upgrades are **rejected** under cookie or identity auth,
   while Bearer keeps its existing skip-verify path so native clients still work.
3. **Identity trust may only be enabled on a loopback bind** — the daemon refuses
   to start otherwise. An identity header is spoofable by anything that can reach
   the socket, so trusting it is only sound when the sole thing that can reach the
   socket is `tailscale serve` on the same host.

### Notes for review

- The tracked `backend/internal/httpd/webui/dist/index.html` is deliberately a
  **placeholder**. `dist/` is gitignored; only `index.html` is tracked so
  `//go:embed all:dist` compiles. Run `build:web` to produce a real bundle.
- `AGENTS.md` gains one bullet recording that the LAN listener may now also serve
  the web UI, in the same amendment pattern ADR-0001 used for the loopback rule.

### Testing

This branch was verified by `go build ./...`, `go vet ./...`, and
`go test ./internal/websession/... ./internal/httpd/...` — all clean, including the
OpenAPI parity test.

The same code, as merged on the fork's integration branch, was additionally
verified end to end against a real daemon: a browser with no Electron present
rendering live data and a streaming PTY over `/mux`, including from a second
device over `tailscale serve`. That run used a real built bundle; this branch
ships the placeholder `dist/index.html` described above, so it needs `build:web`
before it can serve the SPA itself.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:

PR #4311, opened 2026-08-23
feat: add Kimi token usage observability

nikhilachale:
## What

Adds Kimi token-usage observability using the shared token-source foundation.

- discovers Kimi main and child-agent `wire.jsonl` sources
- parses `usage.record` events into the canonical token model
- persists and aggregates Kimi usage through the shared collector and SQLite store
- adds parser, discovery, migration, and storage coverage

## Why this PR was recreated

This is the clean replacement for the earlier Kimi work. Pi and Qwen were also recreated as separate PRs so each provider can be reviewed, tested, merged, or reverted independently.

1. **Kimi:** #4311 (this PR)
2. **Pi:** #4326
3. **Qwen:** #4327

## Implementation

Kimi wire usage is normalized through the existing shared token-source contracts. Cache reads remain cached input; direct input plus cache creation form derived uncached input; all input buckets form total input.

The storage migration preserves existing usage rows, IDs, indexes, views, and change-data-capture triggers while widening the supported harness and source-kind constraints.

## Testing

- `go test ./internal/observe/usage ./internal/service/usage ./internal/storage/sqlite ./internal/storage/sqlite/store ./internal/daemon -count=1`
- `go vet ./...`
- focused golangci-lint over all changed packages: `0 issues`
- `git diff --check`

The full backend run only encountered the repository's pre-existing timing-sensitive fake-agent lifecycle test; the same failure was reproduced on unchanged `main`. All directly affected packages pass.

## Checklist

- [x] Based on `main`
- [x] One provider-focused change

### DIFF
--- a/backend/internal/observe/usage/parser_kimi.go
+++ b/backend/internal/observe/usage/parser_kimi.go
@@ -1,0 +1,99 @@
+package usage
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+type kimiWireRecord struct {
+	ID    string          `json:"id"`
+	Time  json.RawMessage `json:"time"`
+	Type  string          `json:"type"`
+	Model string          `json:"model"`
+	Usage json.RawMessage `json:"usage"`
+}
+
+type kimiNativeUsage struct {
+	InputOther         int64 `json:"inputOther"`
+	InputCacheRead     int64 `json:"inputCacheRead"`
+	InputCacheCreation int64 `json:"inputCacheCreation"`
+	Output             int64 `json:"output"`
+}
+
+func parseKimi(source domain.UsageSourceContext, records []jsonlRecord, result *parseResult) {
+	eventsByKey := make(map[string]domain.ModelUsageEvent)
+	for _, record := range records {
+		var native kimiWireRecord
+		if err := json.Unmarshal(record.Data, &native); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		if native.Type != "usage.record" {
+			continue
+		}
+		model := firstNonEmpty(native.Model)
+		if model == "" || !jsonValueReported(native.Usage) {
+			recordMalformed(result)
+			continue
+		}
+		var usage kimiNativeUsage
+		if err := json.Unmarshal(native.Usage, &usage); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		tokens, ok := normalizeAnthropicUsage(
+			usage.InputOther,
+			usage.InputCacheCreation,
+			usage.InputCacheRead,
+			usage.Output,
+			nil,
+			nil,
+		)
+		if !ok {
+			recordMalformed(result)
+			continue
+		}
+		identity := firstNonEmpty(native.ID, kimiRecordTimeIdentity(native.Time), strconv.FormatInt(record.Offset, 10))
+		event := domain.ModelUsageEvent{
+			ProviderID:        domain.UsageProviderAnthropic,
+			ModelID:           model,
+			MeasurementKind:   domain.UsageMeasurementNativeReported,
+			Tokens:            tokens,
+			ProviderUsageJSON: boundedProviderUsage(native.Usage),
+			CreatedAt:         parseUsageTimestamp(kimiRecordTimeIdentity(native.Time)),
+			SourceEventKey: stableSourceEventKey(
+				"kimi",
+				source.NativeRootID,
+				source.Source.SubagentID,
+				identity,
+				model,
+			),
+		}
+		if existing, duplicate := eventsByKey[event.SourceEventKey]; duplicate {
+			if !usageEventsEqual(existing, event) {
+				result.Cursor.AnomalyCount++
+				result.Cursor.LastErrorCode = domain.UsageErrorSourceEventConflict
+			}
+			continue
+		}
+		eventsByKey[event.SourceEventKey] = event
+		result.Events = append(result.Events, event)
+	}
+}
+
+func kimiRecordTimeIdentity(raw json.RawMessage) string {
+	if len(raw) == 0 || string(raw) == "null" {
+		return ""
+	}
+	var text string
+	if err := json.Unmarshal(raw, &text); err == nil {
+		return text
+	}
+	var number json.Number
+	if err := json.Unmarshal(raw, &number); err == nil {
+		return number.String()
+	}
+	return ""
+}
--- a/backend/internal/observe/usage/parser_kimi.go
+++ b/backend/internal/observe/usage/parser_kimi.go
@@ -3,6 +3,7 @@ package usage
 import (
 	"encoding/json"
 	"strconv"
+	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 )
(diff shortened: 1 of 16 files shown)


backend/internal/storage/sqlite/migrate_burned_versions_test.go
```diff
@@ -110,6 +110,7 @@ var shippedMigrations = map[int64]string{
 	104: "0104_agent_inventory_cache.sql",
 	105: "0105_default_session_mode_chat.sql",
 	106: "0106_pr_comment_review_id.sql",
+	107: "0107_allow_kimi_usage.sql",
```
illegalcall: [P1] Renumber this migration before merge

`main` already ships migration version 107 as `0107_recovered_conversation_turns.sql`. Reusing 107 here creates a ledger merge conflict and can cause Goose to treat this distinct schema change as already applied on existing installations. Please rebase, assign the next unused migration number, and update this ledger entry.

backend/internal/observe/usage/parser_kimi.go
```diff
@@ -0,0 +1,92 @@
...
+			recordMalformed(result)
+			continue
+		}
+		identity := firstNonEmpty(native.ID, kimiRecordTimeIdentity(native.Time), strconv.FormatInt(record.Offset, 10))
+		event := domain.ModelUsageEvent{
+			ProviderID:      domain.UsageProviderAnthropic,
+			ModelID:         model,
+			Tokens:          tokens,
+			ProviderDetails: domain.UsageProviderDetails{Anthropic: &details},
+			CreatedAt:       parseUsageTimestamp(kimiRecordTimeIdentity(native.Time)),
```
illegalcall: [P2] Parse numeric epoch-millisecond timestamps

`kimiRecordTimeIdentity` accepts numeric `time` values, but this passes their decimal representation to `parseUsageTimestamp`, which only accepts RFC3339. Numeric records therefore get a zero `CreatedAt` and are persisted with a NULL timestamp. Please convert numeric values from Unix milliseconds and add a numeric fixture test while retaining RFC3339 support if needed.

backend/internal/service/usage/collector.go
```diff
@@ -1343,6 +1377,48 @@ func (c *Collector) registerDiscoveredClaudeSubagents(
...
 
+func (c *Collector) registerDiscoveredKimiAgents(
+	ctx context.Context,
+	binding domain.UsageBindingRecord,
+	mainPath string,
+	now time.Time,
+	reactivateExisting bool,
+) error {
+	sessionDir := filepath.Dir(filepath.Dir(filepath.Dir(mainPath)))
+	paths, err := filepath.Glob(filepath.Join(sessionDir, "agents", "*", "wire.jsonl"))
```
illegalcall, line 1423: [P2] Keep Kimi child-agent discovery active

This scan runs only when the main source is initially registered or explicitly reconciled. `HasPendingUsageDiscovery` and `ListUsageDiscoveryBindings` still restrict discovery to Claude/Codex, so a child `wire.jsonl` created after this scan can be missed unless another lifecycle hook happens. Please include Kimi in bounded reconciliation, or watch the agents directory, and test a child file created after initial registration.

backend/internal/service/usage/collector.go
```diff
@@ -1654,18 +1730,67 @@ func validateSourceAttribution(
...
+		subagentID,
+		resolved,
+		expectedParentID,
+	); err != nil {
+		return err
+	}
+	if kind != domain.UsageSourceKimiWire {
+		return nil
+	}
+	expectedRoot, err := filepath.EvalSymlinks(filepath.Join(c.roots.KimiHome, "sessions"))
```
illegalcall, line 1809: [P2] Add rejection tests for this trust boundary

This containment check protects the provider-owned artifact boundary, but the new tests cover only a valid path. Please add cases for an absolute path outside the Kimi sessions root, a symlink escape, a deleted index record, and malformed index data so a future regression cannot ingest an arbitrary JSONL file.

backend/internal/service/usage/collector.go
```diff
@@ -332,8 +338,11 @@ func (c *Collector) RecordHook(ctx context.Context, sessionID domain.SessionID,
 
 	if mainArtifact != nil {
 		kind := domain.UsageSourceClaudeMain
-		if session.Harness == domain.HarnessCodex {
+		switch session.Harness {
+		case domain.HarnessCodex:
 			kind = domain.UsageSourceCodexRollout
+		case domain.HarnessKimi:
```

This mapping is repeated in the hook, backfill, and reconcile paths. A future provider can be added to one switch and silently default to Claude in another. Please extract one `sourceKindForHarness` helper and cover the supported harnesses once.

illegalcall (changes requested):
Requesting changes based on the inline findings below. The targeted tests pass, but the migration conflict and usage-discovery/timestamp gaps need to be resolved before merge.

backend/internal/storage/sqlite/migrate_burned_versions_test.go
nikhilachale: Fixed. Rebased #4311 onto current main, renumbered the Kimi migration to 0108_allow_kimi_usage.sql, and updated the append-only migration ledger.

backend/internal/service/usage/collector.go
nikhilachale: Fixed. Active Kimi bindings now remain in bounded discovery reconciliation, and HasPendingUsageDiscovery schedules the retry. Added a regression test that creates a child wire.jsonl after initial registration and discovers it during reconciliation.

backend/internal/service/usage/collector.go
nikhilachale: Covered. Added rejection tests for an absolute path outside the Kimi sessions root, a symlink escape, a deleted latest index record, and malformed index data.

backend/internal/service/usage/collector.go

backend/internal/observe/usage/parser_kimi.go
nikhilachale: Fixed. Kimi timestamps now parse numeric Unix epoch milliseconds while retaining RFC3339 support. Added fixtures covering both formats.

backend/internal/storage/sqlite/migrations/0117_allow_kimi_usage.sql
```diff
@@ -0,0 +1,202 @@
+-- +goose NO TRANSACTION
```
illegalcall, line 1: **[P1] Rebase this feature onto current `main` and allocate a fresh migration version.** Current `main` already owns migrations 0108 through 0116, so this file collides with an existing migration and the branch also conflicts in the burned-version ledger. This needs more than a numeric rename: current `main` moved usage storage to `usage_measurement_kind` and `provider_usage_json` in migration 0115, so please port the Kimi changes onto that schema and use the next available post-main migration.

backend/internal/observe/usage/parser_kimi.go
```diff
@@ -0,0 +1,112 @@
...
+		return time.Time{}
+	}
+	milliseconds, err := number.Int64()
+	if err != nil {
+		return time.Time{}
+	}
+	return time.UnixMilli(milliseconds).UTC()
+}
+
+func kimiRecordTimeIdentity(raw json.RawMessage) string {
```
illegalcall: **[P3] Decode this timestamp once and reuse the result for both `CreatedAt` and source-event identity.** `parseKimiRecordTimestamp` and `kimiRecordTimeIdentity` independently interpret the same raw string/number value. That duplicates the format rules and allows the event timestamp and identity to drift if one decoder changes later. A small decoded value carrying both the canonical identity and `time.Time` would keep them consistent.

illegalcall (changes requested):
Requesting changes because this head is not compatible with current `main`. The earlier epoch-millisecond and active child-session discovery fixes look good, and the exact-head usage tests pass. The remaining blockers are called out inline.

backend/internal/storage/sqlite/migrations/0117_allow_kimi_usage.sql
nikhilachale: Fixed in e1177b506. Rebased onto current main, replaced the colliding migration with 0117_allow_kimi_usage.sql, and ported the rebuild to preserve the post-0115/0116 schema, including provider_hint, usage_measurement_kind, provider_usage_json, and billing_provider_source. Added a direct 0116-to-0117 migration test that preserves existing usage facts and accepts Kimi bindings/sources.

backend/internal/observe/usage/parser_kimi.go
nikhilachale: Fixed in 5439b00e7. decodeKimiRecordTime now returns the canonical identity and time.Time together; parseKimi calls it once and reuses both values for SourceEventKey input and CreatedAt. Tests cover RFC3339 and epoch-millisecond inputs.

illegalcall (approved):

PR #4312, opened 2026-08-23
feat(renderer): gate native-only UI by bridge capability, not Electron presence

execsumo:
Freezes a capability contract on the renderer bridge and gates native-only UI on
the specific capability each surface needs, instead of on whether Electron happens
to be present.

### The problem

Native-only decisions are currently made by testing `window.ao` directly, which
conflates two different questions: *is this Electron?* and *can this surface
work here?* The sharpest example is `TerminalPane.tsx`, where `if (!window.ao)`
early-returns a **static fake terminal** — so terminals are stubbed by
Electron-absence rather than by any real capability, and a caller who has a
working terminal transport but no Electron gets a convincing fake instead of the
real thing.

`BrowserPanel.tsx` had the same shape: a `StaticPreview` component rendering a
mocked-up "Demo app preview" card with invented route counts and latency
whenever the native browser view was unavailable.

### What this does

- Adds an explicit capability set to `aoBridge` and freezes it as a contract, so
  each surface asks for what it needs rather than inferring it.
- Replaces the direct `window.ao` checks in the affected renderer files with
  capability checks against that contract.
- **Deletes `StaticPreview`** and replaces it with an honest "not available here"
  message. Same for the fake terminal path.
- Adds the six locale keys those new states need, in all eight catalogs.

### Alternatives rejected

- **Leaving the `window.ao` checks in place and adding a flag.** The existing
  `VITE_NO_ELECTRON` flag is already overloaded — it unconditionally switches the
  renderer to mock data, so it cannot express "no Electron, but a real daemon."
  Another flag with the same shape would not either.
- **A second renderer built for non-Electron targets.** It would need its own copy
  of the board, terminals, chat and orchestrator surfaces, and would drift.

### Rebase note

Rebased onto current `main`. Upstream's device-preset viewport picker is retained
inside the component this change renames to `NativeBrowserPanelView` — it is
therefore capability-gated by construction, since that component renders only when
the native browser panel is available. The early gate, `aoBridge` access, and the
`StaticPreview` deletion are all preserved, and upstream's new locale keys survive
alongside the six added here.

### Testing

`npm run frontend:typecheck` clean, and the renderer vitest scoped to the affected
files — 11 files, 232 tests passing. The full renderer suite and the e2e suite were
not re-run on this rebased branch.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:

i-trytoohard:

i-trytoohard:
@execsumo Closing as stale: no activity from you for 31 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

Issue #2804, opened 2026-07-18
feat(nav): give the orchestrator a home in the sidebar session tree


AgentWrapper:
## Proposal
Worker sessions live in the sidebar tree; the orchestrator lives in a topbar button. That is two navigation systems for the same concept (a session), and the sidebar never shows the whole fleet. Pinning an **Orchestrator row at the top of each project's session group** would unify navigation and make the orchestrator's state (working/idle) visible at a glance like any other session.

**Source:** maintainer chat report (desktop app, macOS), 2026-07-18 | **Analyzed against:** `dec77c61b`

## Related
- [#1362](https://github.com/AgentWrapper/agent-orchestrator/issues/1362): orchestrator button navigates to lexicographically smallest orchestrator — symptom of the same split navigation model
- [#1211](https://github.com/AgentWrapper/agent-orchestrator/issues/1211): orchestrator button links to wrong parent session
- [#2723](https://github.com/AgentWrapper/agent-orchestrator/issues/2723): empty state should explain task-vs-orchestrator (audit AO-06)
- [#2728](https://github.com/AgentWrapper/agent-orchestrator/issues/2728): UI/UX audit epic

Rishet11:
@AgentWrapper, the global orchestrator button already covers the main workflow. A separate orchestrator row and status for every project would be a small navigation improvement, so this should remain open as P3.

Rishet11:
Assigned to @ronishrohan

codebanditssss:
Closing as already implemented. The current Sidebar renders a project-level Orchestrator row alongside the session tree, including active-state styling and spawn/resume/open behavior.

PR #4313, opened 2026-08-23
feat(orchestrator): make the project orchestrator a route destination

execsumo:
Makes the project orchestrator a real destination — a route you can navigate to,
link to, and come back to — rather than a topbar shortcut that spawns a session
as a side effect of being clicked.

### Why

The orchestrator is a long-lived, project-scoped thing: it plans and delegates
across a project's workers. Reaching it through a button that both *navigates*
and *spawns* has two costs. There is no address for it, so nothing can link to it
and returning to it means finding the session again. And spawn failures have to
surface in the topbar, which is why the topbar carried its own spawn-in-progress
and spawn-error state.

### What this does

- Adds `/projects/$projectId/orchestrator` as a route, with its own loading,
  not-started, unavailable and restarting states.
- Derives the current session from the route when on that path, so the topbar and
  inspector resolve correctly without a session id in the URL.
- **Removes the topbar's inline spawn path** — `spawnOrchestrator`, the spawning
  flag, and the board-scoped spawn error — since lifecycle now belongs to the
  destination.
- Adds the 22 `orchestratorRoute.*` keys plus `shell.orchestratorOrigin`, in all
  eight catalogs.

### Alternative rejected

Keeping the orchestrator as a shortcut into the existing session surface. That
surface is worker-shaped: it cannot express project-level lifecycle (start,
restart, not-configured) or a stable address, which is most of what makes the
orchestrator useful to return to.

### Rebase note

Rebased onto current `main`. Upstream's `TopbarKillError` → `TopbarActionError`
rename is applied at the two call sites that survive here; the third was inside
the board spawn-error block this change removes, so it is gone rather than
restored. Upstream's `TopbarOpenEditorButton` is retained in the position it was
added.

One follow-up worth flagging: that button is **ungated**. This branch does not
carry a capability contract, so there is nothing here to gate it with. If the
companion renderer-capability change lands, that button will want gating for
non-desktop clients.

### Testing

`npm run frontend:typecheck` clean, and the renderer vitest scoped to the affected
files — 7 files, 128 tests passing. The full renderer suite and the e2e suite were
not re-run on this rebased branch.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@execsumo this looks like it addresses #2804 — linking for tracking. Adding `Fixes #2804` to the PR body will auto-close it on merge.

i-trytoohard:

i-trytoohard:

i-trytoohard:
@execsumo Closing as stale: no activity from you for 31 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

PR #4314, opened 2026-08-23
fix(desktop): prevent macOS DevTools menu crash

### LINKED ISSUE #4308: bug(desktop): macOS Toggle Developer Tools crashes the Electron main process

illegalcall:
Fixes [#4308](https://github.com/Untrivial-ai/agent-orchestrator/issues/4308).

## Summary

- install an explicit macOS application menu so Electron's unsafe default `toggleDevTools` role is not used
- route the native menu and accelerator through AO's guarded last-focused-browser handler, with the shell web contents as fallback
- preserve Electron's complete standard macOS App, File, Edit, and Window menus through their top-level roles
- add regression coverage proving the DevTools item has a click handler and no `toggleDevTools` role

## Root cause

AO composes its desktop window from a `BaseWindow` and `WebContentsView` children. Electron's default macOS DevTools role assumes focus belongs to a `BrowserWindow`; when focus belongs to an AO child view, Electron resolves no focused window and dereferences `toggleDevTools` on an undefined target in the main process.

## Related

- [PR #3910](https://github.com/Untrivial-ai/agent-orchestrator/pull/3910) has the same core diagnosis. This implementation preserves the complete Electron-provided macOS menus via `appMenu`, `fileMenu`, `editMenu`, and `windowMenu` instead of reconstructing a subset of their commands.

## Verification

- `npm test -- --run src/main/menu.test.ts` — 4 tests passed
- `npm run typecheck` — passed
- `npm run typecheck:e2e` — passed
- `npx vitest run` — 209 files and 2,767 tests passed
- launched the real Electron checkout in isolated mode with the daemon healthy on `127.0.0.1:3002`
- Electron-native harness installed the generated macOS menu and directly invoked its DevTools item: one handler call, no `toggleDevTools` role, all four standard menu roles present

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -99,7 +99,7 @@ import { keepDaemonAlive, shouldLinkOnAttach } from "./main/daemon-owner";
 import { readMigrationState, updateMigration, writeAppStateMarker, type MigrationState } from "./main/app-state";
 import { isAllowedAppExternalURL, openAllowedAppExternalURL } from "./main/external-open";
 import { dockBounceType, shouldReplaceBounce, shouldSignalAttention, shouldToast } from "./main/notification-signals";
-import { buildWindowsAppMenuTemplate } from "./main/menu";
+import { buildMacAppMenuTemplate, buildWindowsAppMenuTemplate } from "./main/menu";
 import { ancestorRepositorySetupWarning, scanImportFolder } from "./main/import-folder-scan";
 import { parseOpenFolderPathArg } from "./main/open-folder-arg";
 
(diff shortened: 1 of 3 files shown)


illegalcall:
## Verification evidence

Verified commit `3627a8e76` in the real macOS Electron app with an isolated profile/data directory.

- Invoked the installed **View → Toggle Developer Tools** native menu item directly in Electron's main process.
- Runtime inspection confirmed the item uses the guarded custom handler (`role: null`, accelerator `Alt+Command+I`) and the AO shell reported `devToolsOpened: true`.
- The app remained running and its isolated daemon continued returning successful API responses after DevTools opened.

AO desktop app running with DevTools open after invoking the fixed native menu item

nikhilachale (approved):
No blocking Found ,LGTM

nikhilachale:
Screenshot 2026-08-24 at 02 19 14

PR #4318, opened 2026-08-23
fix(projects): use per-project orchestrator agent override, not daemon default

### LINKED ISSUE #4316: [Bug]: Project Settings orchestrator-agent override not reflected in Get/UpdateSettings response
### Summary

Saving a per-project orchestrator agent override in Project Settings appears to silently not take: the save response (and any subsequent single-project fetch) still shows the daemon's default agent instead of the value just saved. The project's list view shows the correct value; the detail view does not.

### Steps to reproduce

Set a project's orchestrator agent override to a harness different from the daemon default (AO_AGENT) via Project Settings, then save.

Observe the response returned by the save call, or fetch the project's detail (GET /api/v1/projects/{id}).

### Expected behavior

The detail response's agent field reflects the project's saved orchestrator override (matching what GET /api/v1/projects already shows correctly for that project).

### Actual behavior

The detail response's agent field always reports the daemon-wide default harness, regardless of the project's saved override. Root cause: projectFromRow in backend/internal/service/project/service.go set Agent: string(m.defaultHarness) unconditionally, never reading row.Config.Orchestrator.Harness. List() in the same file reads it correctly. A regression test (TestManager_GetIgnoresProjectOrchestratorAgentOverride) reproduces this against a real sqlite store; fix + test ready in a PR.

### AO version

commit ebd56bd (main, 2026-08-23)

### Environment

Windows 11, Go backend

### Logs / extra context

```shell

```

### Before submitting

valdepeace:
## What

`GET /api/v1/projects/{id}` (and thus `UpdateSettings`'s own response) always reported the daemon-wide default harness for a project's `agent` field, ignoring the project's own saved `config.orchestrator.agent` override. `List()` (`GET /api/v1/projects`) already read it correctly — only the single-project view did not.

## Why

Fixes #4316. Saving a different orchestrator agent in Project Settings appeared to silently not take: the save response and any subsequent fetch echoed back the old (daemon default) value instead of what was just persisted.

## How

`projectFromRow` in `backend/internal/service/project/service.go` set `Agent: string(m.defaultHarness)` unconditionally. Changed it to prefer `row.Config.Orchestrator.Harness`, falling back to `m.defaultHarness` only when the project has no override — mirroring what `List()` already does, and preserving the existing "falls back to daemon default when unset" contract that `TestManager_GetUsesConfiguredDefaultHarness` already covers.

## Testing

- New regression test `TestManager_GetIgnoresProjectOrchestratorAgentOverride`: fails on the old code, passes with the fix.
- `go test ./internal/service/project/...` — green except pre-existing, environment-specific failures unrelated to this change (network-dependent clone tests, a Windows path-separator assertion) — confirmed identical on `main` without this change.
- Manually replayed the exact save the bug describes directly against a running daemon (before/after): before, the response echoed the daemon default; after, it correctly echoes the saved override.

## Checklist

- [x] Branched from `main`
- [x] One focused change; closes #4316
- [x] Test added for the user-visible behavior
- [ ] CI (not run locally beyond `go test`/`go build` for the touched package)

### DIFF
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -720,6 +720,16 @@ func (m *Service) projectFromRow(ctx context.Context, row domain.ProjectRecord) 
 			}
 		}
 	}
+	// Agent reports the project's *effective* orchestrator harness: its own
+	// override when set, falling back to the daemon-wide default otherwise.
+	// Bug fix: this used to report the daemon default unconditionally,
+	// ignoring row.Config.Orchestrator.Harness, so a saved per-project
+	// override never showed up here (List()'s Summary.OrchestratorAgent got
+	// this right all along; only this single-project view did not).
+	agent := row.Config.Orchestrator.Harness
+	if agent == "" {
+		agent = m.defaultHarness
+	}
 	p := Project{
 		ID:            domain.ProjectID(row.ID),
 		Name:          displayName(row),
@@ -727,7 +737,7 @@ func (m *Service) projectFromRow(ctx context.Context, row domain.ProjectRecord) 
 		Path:          row.Path,
 		Repo:          row.RepoOriginURL,
 		DefaultBranch: defaultBranch,
-		Agent:         string(m.defaultHarness),
+		Agent:         string(agent),
 	}
 	p.Config = projectConfigPtr(row.Config)
 	return p
--- a/backend/internal/service/project/service_test.go
+++ b/backend/internal/service/project/service_test.go
@@ -574,6 +574,69 @@ func TestManager_GetUsesConfiguredDefaultHarness(t *testing.T) {
 	}
 }
 
+// TestManager_GetIgnoresProjectOrchestratorAgentOverride reproduces a bug: a
+// project's saved orchestrator-agent override (Config.Orchestrator.Harness)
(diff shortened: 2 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

valdepeace:
Hi @illegalcall @somewherelostt 👋

Thanks for picking this up as the on-call pair. Two things whenever you have a moment:

- **CI approval** — the 3 workflows are still awaiting maintainer approval, so there are no check results yet. Could one of you approve the runs?
- **Review** — it's a small one (+74/−1, one function + one regression test). `projectFromRow` now prefers `row.Config.Orchestrator.Harness` and only falls back to `m.defaultHarness` when unset, mirroring what `List()` already does. No DTO shape change, so no OpenAPI regeneration needed. The new test `TestManager_GetIgnoresProjectOrchestratorAgentOverride` fails on `main` and passes with the fix.

Happy to iterate on any feedback. Thanks! 🙌

i-trytoohard:
@valdepeace Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4319, opened 2026-08-23
fix(session_manager): stop the replaced orchestrator's process on Windows

### LINKED ISSUE #4317: [Bug]: Orchestrator replacement leaves the old agent process running on Windows, worktree cleanup fails
### Summary

Replacing a project's active orchestrator (e.g. after changing its configured agent, or any "clean" re-spawn) can fail on Windows with an internal server error. The previous orchestrator's agent process is left running with its working directory inside the worktree the replacement is trying to delete, so the force-remove fails with a file-in-use error. When this happens mid-flight, the old orchestrator can end up retired with no new one successfully created, leaving the project without a working orchestrator until manually recovered.

### Steps to reproduce

On Windows, start a project's orchestrator (either mode, TUI or chat).

Trigger a clean orchestrator replacement for that project (POST /api/v1/orchestrators with clean: true, e.g. via changing the project's orchestrator agent while a session is active) and observe the response.

### Expected behavior

The old orchestrator's agent process is fully terminated before its worktree is force-removed; replacement succeeds.

### Actual behavior

500 INTERNAL_ERROR, with the daemon log showing: retire replacement <session>: force destroy: gitworktree: force remove path "...\worktrees\<project>\orchestrator\<project>-orchestrator": unlinkat ...: The process cannot access the file because it is being used by another process.

Two independent causes, both in RetireForReplacement (backend/internal/session_manager/manager.go): (1) TUI/conpty sessions: Destroy() killed only the ConPTY's direct child (the "agent-process supervise" wrapper) via TerminateProcess, which does not propagate to that process's own child (the actual claude.exe/codex.exe) so the real agent process was orphaned. (2) Chat-mode sessions: a chat session has no runtime handle at all -- its controller owns an app-server child process directly, not through the runtime port (see the pre-existing, correct branch in Kill, which RetireForReplacement never mirrored) -- so the real agent process was simply never asked to stop.

Fix ready in a PR for both: (1) taskkill /PID <pid> /T /F on Close, matching the same pattern already used in adapters/chatdriver/acp/process_windows.go; (2) branch on session mode in RetireForReplacement (and its workspace-project sibling) the same way Kill already does, calling stopChatBestEffort for chat sessions. Regression test included (fails on old code, passes with the fix). Verified live against a packaged Windows build: reproduced the exact failure on both a TUI and a chat orchestrator, confirmed the old agent process was actually terminated (not just that the API stopped erroring) after each fix.

### AO version

commit ebd56bd (main, 2026-08-23)

### Environment

Windows 11, Go backend

### Logs / extra context

```shell

```

### Before submitting

valdepeace:
## What

Orchestrator replacement (`RetireForReplacement` and its workspace-project sibling, `backend/internal/session_manager/manager.go`) could leave the previous orchestrator's agent process running on Windows, which then holds files open in the worktree the replacement is about to force-remove.

## Why

Fixes #4317. Replacing a project's orchestrator (e.g. via changing its configured agent while a session is active) could fail with:
```
retire replacement <session>: force destroy: gitworktree: force remove path "...":
The process cannot access the file because it is being used by another process.
```
When this happens mid-flight the old orchestrator can end up retired with no new one successfully created.

## How

Two independent, unrelated root causes, both fixed here since they produce the same symptom:

1. **TUI/conpty sessions** (`backend/internal/adapters/runtime/conpty/host_conpty_windows.go`): `Close()` killed only the ConPTY's direct child via `cmd.Process.Kill()` (`TerminateProcess`), which does not propagate to that process's own children (the actual `claude.exe`/`codex.exe`, spawned by the `agent-process supervise` wrapper). Fixed by killing the whole tree with `taskkill /PID <pid> /T /F` — mirroring the identical, already-established pattern in the sibling `adapters/chatdriver/acp/process_windows.go` (deliberately not a Windows Job Object: that also worked but had no precedent anywhere in this codebase).
2. **Chat-mode sessions**: a chat session has no runtime handle at all — its controller owns an app-server child process directly (`ChatLauncher`/`chatSvc`), not through the runtime port. `RetireForReplacement` never branched on this, unlike `Kill`, which already has the correct `if mode == chat { stopChatBestEffort } else if handle.ID != "" { runtime.Destroy }` pattern (with a comment explicitly calling out the Windows directory-lock risk). Mirrored that branch into `RetireForReplacement` and `retireWorkspaceProjectForReplacement` (3 call sites total).

## Testing

- New regression test `TestRetireForReplacementStopsChatSessionInsteadOfRuntimeDestroy`: fails on the old code (chat.StopChat never called), passes with the fix.
- `go test ./internal/session_manager/... ./internal/adapters/runtime/conpty/...` — green except pre-existing, environment-specific failures unrelated to this change (confirmed identical on `main` without it).
- Verified live end-to-end against a packaged Windows build, for both fixes independently and together:
  - Reproduced the original failure on a **chat**-mode orchestrator (the exact repro from #4317), confirmed the fix resolves it and the old agent process is actually gone (`Get-Process` on its PID) after replacement succeeds.
  - Reproduced the same failure on a **TUI**-mode orchestrator, confirmed the fix resolves it and the old process tree (wrapper + agent binary) is actually gone after replacement succeeds.
  - Re-verified after refactoring the conpty fix from a Job Object to `taskkill /T /F` that both repros still pass.

## Checklist

- [x] Branched from `main`
- [x] One focused change (two root causes, one symptom, one issue); closes #4317
- [x] Follows AGENTS.md conventions and PR hygiene; conpty fix intentionally mirrors the existing `acp/process_windows.go` pattern rather than introducing a new one
- [x] Tests added for the user-visible behavior
- [ ] CI (not run locally beyond `go test`/`go build`/`go vet` for the touched packages)

## Note for reviewers

This touches session-teardown/process-lifecycle code with real risk if wrong. Flagging for extra scrutiny before merge rather than treating it as routine — happy to split further or add more test coverage if requested.

### DIFF
--- a/backend/internal/adapters/runtime/conpty/host_conpty_windows.go
+++ b/backend/internal/adapters/runtime/conpty/host_conpty_windows.go
@@ -5,8 +5,10 @@ package conpty
 import (
 	"fmt"
 	"os"
+	"os/exec"
+	"strconv"
 	"sync"
-	"unsafe"
+	"syscall"
 
 	gopty "github.com/aymanbagabas/go-pty"
 	"golang.org/x/sys/windows"
@@ -17,18 +19,6 @@ import (
 type conptyConn struct {
 	pty gopty.ConPty
 	cmd *gopty.Cmd
-
-	// job groups the ConPTY child and every process it spawns (e.g. the
-	// "agent-process supervise" wrapper and, under that, the actual agent
-	// binary) so Close can terminate the whole tree atomically. Without this,
-	// killing only cmd.Process (TerminateProcess, no signal, no propagation)
-	// orphans grandchildren, which can keep holding open file handles in the
-	// session's git worktree well past its removal retry budget, failing
-	// orchestrator replacement with "process cannot access the file". See
-	// windows.INVALID_HANDLE_VALUE check in newConPTY: job stays 0 if it
-	// could not be created/assigned, and Close falls back to the old
-	// single-process kill so a job-object failure never blocks teardown.
-	job windows.Handle
 
 	once     sync.Once
 	doneC    chan struct{}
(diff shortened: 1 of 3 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@valdepeace Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@valdepeace Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@valdepeace Closing as stale: no activity from you for 31 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

PR #4326, opened 2026-08-23
feat: add Pi token usage observability

nikhilachale:
## What

Adds Pi token-usage observability as a standalone change based on the latest `main`.

- parses Pi session JSONL for OpenAI-compatible and Anthropic providers
- discovers sessions from `PI_CODING_AGENT_DIR` or `~/.pi/agent/sessions`
- supports late session-ID discovery and lifecycle reactivation
- persists and aggregates Pi usage through the shared collector and SQLite store
- adds parser, discovery, lifecycle, migration, and storage coverage

## Why this PR was recreated

This replaces the earlier Pi PR #4184 without carrying Kimi or Qwen branch history. Pi can now be reviewed, tested, merged, or reverted independently.

1. **Kimi:** #4311
2. **Pi:** #4326 (this PR)
3. **Qwen:** #4327

## Implementation

Pi messages are normalized into the shared token model. The collector supports explicit session IDs and late discovery by matching Pi session metadata to a live AO workspace. Migration 0109 is intentionally relative to the Kimi migration: it adds Pi support on upgrade and removes only Pi usage on rollback, preserving Kimi data and constraints.

## Testing

- `go test ./internal/observe/usage ./internal/service/usage ./internal/storage/sqlite ./internal/storage/sqlite/store ./internal/lifecycle -count=1`
- `go test ./internal/daemon -count=1`
- `go vet` over all changed backend packages
- golangci-lint v2.12.2 over all changed packages: `0 issues`
- `git diff --check`

The full backend run only encountered the repository's pre-existing timing-sensitive fake-agent lifecycle test under concurrent machine load. All directly affected packages pass.

## Checklist

- [x] Based on the latest `main`
- [x] Contains only Pi provider logic

### DIFF
--- a/backend/internal/observe/usage/parser_pi.go
+++ b/backend/internal/observe/usage/parser_pi.go
@@ -8,22 +8,27 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
 )
 
+const piUnattributedUsageModel = "Tools/summaries"
+
 type piSessionRecord struct {
-	Type      string `json:"type"`
-	ID        string `json:"id"`
-	Timestamp string `json:"timestamp"`
+	Type      string          `json:"type"`
+	ID        string          `json:"id"`
+	Timestamp string          `json:"timestamp"`
+	Usage     json.RawMessage `json:"usage"`
 	Message   *struct {
-		Role     string `json:"role"`
-		Provider string `json:"provider"`
-		Model    string `json:"model"`
-		Usage    *struct {
-			Input       int64  `json:"input"`
-			Output      int64  `json:"output"`
-			CacheRead   int64  `json:"cacheRead"`
-			CacheWrite  int64  `json:"cacheWrite"`
-			TotalTokens *int64 `json:"totalTokens"`
-		} `json:"usage"`
+		Role          string          `json:"role"`
+		Provider      string          `json:"provider"`
+		Model         string          `json:"model"`
+		ResponseModel string          `json:"responseModel"`
+		Usage         json.RawMessage `json:"usage"`
 	} `json:"message"`
+}
+
+type piNativeUsage struct {
+	Input      int64 `json:"input"`
+	Output     int64 `json:"output"`
+	CacheRead  int64 `json:"cacheRead"`
+	CacheWrite int64 `json:"cacheWrite"`
 }
 
 func parsePi(source domain.UsageSourceContext, records []jsonlRecord, result *parseResult) {
@@ -34,62 +39,70 @@ func parsePi(source domain.UsageSourceContext, records []jsonlRecord, result *pa
 			recordMalformed(result)
 			continue
 		}
-		if native.Type != "message" || native.Message == nil || native.Message.Role != "assistant" {
+
+		var provider, model string
+		var usageRaw json.RawMessage
+		switch {
+		case native.Type == "message" && native.Message != nil && native.Message.Role == "assistant":
+			provider = strings.TrimSpace(native.Message.Provider)
+			model = firstNonEmpty(native.Message.ResponseModel, native.Message.Model)
+			usageRaw = native.Message.Usage
+			if model == "" || !jsonValueReported(usageRaw) {
+				recordMalformed(result)
+				continue
+			}
+		case native.Type == "message" && native.Message != nil && native.Message.Role == "toolResult" &&
+			jsonValueReported(native.Message.Usage):
+			model = piUnattributedUsageModel
+			usageRaw = native.Message.Usage
+		case (native.Type == "compaction" || native.Type == "branch_summary") && jsonValueReported(native.Usage):
+			model = piUnattributedUsageModel
+			usageRaw = native.Usage
+		default:
 			continue
 		}
-		message := native.Message
-		model := firstNonEmpty(message.Model)
-		if model == "" || message.Usage == nil {
+
+		var usage piNativeUsage
+		if err := json.Unmarshal(usageRaw, &usage); err != nil {
 			recordMalformed(result)
 			continue
 		}
-		provider := strings.TrimSpace(message.Provider)
-		if provider != "" {
-			model = provider + "/" + model
-		}
-		usage := message.Usage
 		input, ok := sumNonNegative(usage.Input, usage.CacheRead, usage.CacheWrite)
 		if !ok {
 			recordMalformed(result)
 			continue
 		}
 		providerID := domain.UsageProviderOpenAI
-		providerDetails := domain.UsageProviderDetails{}
 		var tokens domain.UsageTokenMetrics
 		if strings.EqualFold(provider, "anthropic") {
-			var details domain.AnthropicUsageDetails
-			tokens, details, ok = normalizeAnthropicUsage(
+			providerID = domain.UsageProviderAnthropic
+			tokens, ok = normalizeAnthropicUsage(
 				usage.Input, usage.CacheWrite, usage.CacheRead, usage.Output, nil, nil,
 			)
-			providerID = domain.UsageProviderAnthropic
-			providerDetails.Anthropic = &details
 		} else {
-			reportedTotal := int64(0)
-			if usage.TotalTokens != nil {
-				reportedTotal = *usage.TotalTokens
-			}
-			var details domain.OpenAIUsageDetails
-			tokens, details, ok = normalizeOpenAIUsage(
-				input, usage.CacheRead, usage.CacheWrite, usage.Output, 0, reportedTotal,
-			)
-			// Pi reports direct input and cache buckets separately, so its
-			// canonical total input is derived rather than directly reported.
-			tokens.Provenance.InputTokens = domain.UsageMetricDerived
-			providerDetails.OpenAI = &details
+			tokens, ok = normalizeOpenAIUsage(input, usage.CacheRead, usage.CacheWrite, usage.Output)
 		}
 		if !ok {
 			recordMalformed(result)
 			continue
 		}
+
+		billingProvider := ""
+		if model != piUnattributedUsageModel {
+			billingProvider = canonicalBillingProvider(provider)
+		}
 		identity := firstNonEmpty(native.ID, strconv.FormatInt(record.Offset, 10))
 		event := domain.ModelUsageEvent{
-			ProviderID:      providerID,
-			ModelID:         model,
-			Tokens:          tokens,
-			ProviderDetails: providerDetails,
-			CreatedAt:       parseUsageTimestamp(native.Timestamp),
+			ProviderID:            providerID,
+			BillingProviderID:     billingProvider,
+			BillingProviderSource: domain.ObservedBillingProviderSource(billingProvider),
+			ModelID:               model,
+			MeasurementKind:       domain.UsageMeasurementNativeReported,
+			Tokens:                tokens,
+			ProviderUsageJSON:     boundedProviderUsage(usageRaw),
+			CreatedAt:             parseUsageTimestamp(native.Timestamp),
 			SourceEventKey: stableSourceEventKey(
-				"pi", source.NativeRootID, identity, provider, strings.TrimSpace(message.Model),
+				"pi", source.NativeRootID, native.Type, identity,
 			),
 		}
 		if existing, duplicate := eventsByKey[event.SourceEventKey]; duplicate {
--- a/backend/internal/observe/usage/parser_pi.go
+++ b/backend/internal/observe/usage/parser_pi.go
@@ -1,0 +1,105 @@
+package usage
+
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+type piSessionRecord struct {
+	Type      string `json:"type"`
+	ID        string `json:"id"`
+	Timestamp string `json:"timestamp"`
+	Message   *struct {
+		Role     string `json:"role"`
+		Provider string `json:"provider"`
+		Model    string `json:"model"`
+		Usage    *struct {
+			Input       int64  `json:"input"`
+			Output      int64  `json:"output"`
+			CacheRead   int64  `json:"cacheRead"`
+			CacheWrite  int64  `json:"cacheWrite"`
+			TotalTokens *int64 `json:"totalTokens"`
+		} `json:"usage"`
+	} `json:"message"`
+}
+
+func parsePi(source domain.UsageSourceContext, records []jsonlRecord, result *parseResult) {
+	eventsByKey := make(map[string]domain.ModelUsageEvent)
+	for _, record := range records {
+		var native piSessionRecord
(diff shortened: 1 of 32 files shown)


backend/internal/service/usage/collector.go
```diff
@@ -671,6 +741,57 @@ func (c *Collector) ReconcilePath(ctx context.Context, path string) error {
...
+		if candidateErr == nil && candidate == workspace {
+			matches = append(matches, session)
+		}
+	}
+	if len(matches) != 1 {
+		return nil
+	}
+	now := c.now().UTC()
+	binding, err := c.store.UpsertUsageBinding(ctx, domain.UsageBindingRecord{
+		SessionID: matches[0].ID, Harness: domain.HarnessPi, NativeRootID: meta.ID,
```
illegalcall: [P1] Do not bind historical Pi sessions by CWD alone

`backfillPiPaths` examines up to 256 recent Pi files, and every file whose CWD matches the sole live AO session reaches this upsert with its own native ID. Reusing a workspace across Pi sessions therefore imports old history into the current AO session and inflates its totals. Please disambiguate with the AO launch/session time window or a native identity associated with this session before creating the binding, and cover multiple historical files for one workspace.

backend/internal/storage/sqlite/migrations/0109_allow_provider_usage.sql
```diff
@@ -0,0 +1,205 @@
...
+
+DROP TRIGGER IF EXISTS usage_bindings_cdc_insert;
+DROP TRIGGER IF EXISTS usage_bindings_cdc_update;
+DROP TRIGGER IF EXISTS usage_sources_cdc_update;
+
+DELETE FROM anthropic_usage_event_details
+WHERE event_id IN (
+    SELECT event.id FROM model_usage_events event
+    JOIN usage_bindings binding ON binding.id = event.binding_id
+    WHERE binding.harness IN ('kimi', 'pi', 'qwen')
```
illegalcall: [P1] Make this rollback Pi-only

The Down migration deletes Kimi, Pi, and Qwen events, then rebuilds the constraints for Claude/Codex only. Rolling back 0109 after the Kimi migration destroys already-shipped Kimi usage and leaves migration history inconsistent with the schema. Please make this migration relative to its predecessor so Down removes only Pi, or move the shared expansion into one prerequisite migration with a composable rollback.

backend/internal/service/usage/collector.go
```diff
@@ -179,9 +186,29 @@ func (c *Collector) ReactivateSession(
 	sessionID domain.SessionID,
 	expectedRuntimeLaunchID string,
 ) error {
-	c.mu.Lock()
+	if !c.mu.TryLock() {
+		go func() {
+			c.mu.Lock()
+			defer c.mu.Unlock()
+			if err := c.reactivateSessionLocked(context.WithoutCancel(ctx), sessionID, expectedRuntimeLaunchID); err != nil {
```
illegalcall: [P2] Preserve cancellation and error semantics under contention

This branch queues an unbounded goroutine behind the mutex, strips cancellation with `context.WithoutCancel`, and returns success before the work or its errors are known. Shutdown cannot cancel the queued discovery and callers cannot observe a failure. Please serialize through a daemon-owned worker/context or retain synchronous cancellable locking, and add a contention test.

backend/internal/service/usage/collector.go
```diff
@@ -1746,6 +1883,81 @@ func (c *Collector) discoverPath(ctx context.Context, harness domain.AgentHarnes
 	return matches[0].path, nil
 }
 
+type piSessionMeta struct {
+	Type string `json:"type"`
+	ID   string `json:"id"`
+	CWD  string `json:"cwd"`
+}
+
+func readPiSessionMeta(path string) (piSessionMeta, bool) {
```
illegalcall: [P2] Thread context through this file read

This helper opens and reads a provider artifact without accepting `context.Context`, despite being called from cancellation-aware discovery loops. Please take context as the first argument and check it around the read, matching the repository convention for I/O and blocking work.

backend/internal/service/usage/collector.go
```diff
@@ -332,8 +364,11 @@ func (c *Collector) RecordHook(ctx context.Context, sessionID domain.SessionID,
 
 	if mainArtifact != nil {
 		kind := domain.UsageSourceClaudeMain
-		if session.Harness == domain.HarnessCodex {
+		switch session.Harness {
+		case domain.HarnessCodex:
 			kind = domain.UsageSourceCodexRollout
+		case domain.HarnessPi:
```

This mapping is repeated in the hook, backfill, and reconcile paths. A future provider can be added to one switch and silently default to Claude in another. Please extract one `sourceKindForHarness` helper and cover the supported harnesses once.

illegalcall (changes requested):
Requesting changes based on the inline findings below. The targeted tests pass, but the session-attribution and rollback behavior can corrupt reported or stored usage and should be fixed before merge.

backend/internal/service/usage/collector.go
nikhilachale: Fixed. Pi attribution now requires the provider session-header timestamp to fall within the AO session lifetime and permits only one Pi native binding for the live AO session. Added a regression fixture with current and historical files sharing one workspace; only the current session is bound.

backend/internal/storage/sqlite/migrations/0109_allow_provider_usage.sql
nikhilachale: Fixed. Migration 0109 is now relative to Kimi: Up adds Pi only, and Down deletes only Pi usage while retaining Kimi constraints and data. Added an up/down migration regression test that verifies Kimi survives.

backend/internal/service/usage/collector.go
nikhilachale: Fixed. Replaced the detached TryLock goroutine with a context-aware serialized lock. ReactivateSession now waits synchronously, returns cancellation, and propagates work errors. Added contention and cancellation coverage.

backend/internal/service/usage/collector.go
nikhilachale: Fixed. readPiSessionMeta now accepts context.Context as its first argument and checks cancellation before opening and after reading the provider artifact. All callers thread their context through; canceled-read coverage was added.

backend/internal/service/usage/collector.go

backend/internal/service/usage/collector.go
nikhilachale: Fixed. Reactivation now uses a synchronous context-aware lock: queued callers remain cancellable, shutdown cancellation is preserved, and discovery errors are returned to the caller. Added contention, cancellation, and error-propagation coverage.

backend/internal/service/usage/collector.go
nikhilachale: Fixed. readPiSessionMeta now takes context.Context as its first argument, checks cancellation around the file read, and all discovery callers thread their context through. Added a canceled-read test.

backend/internal/service/usage/collector.go

backend/internal/storage/sqlite/migrations/0118_allow_pi_usage.sql
```diff
@@ -0,0 +1,205 @@
+-- +goose NO TRANSACTION
```
illegalcall, line 1: **[P1] Rebase this feature onto current `main` and use a fresh migration version.** Current `main` already owns 0109 (and migrations through 0116), so this migration and its burned-version ledger entry collide. Current `main` also replaced the older provider-detail/provenance representation with `usage_measurement_kind` and `provider_usage_json`; please port the Pi implementation onto that schema rather than only renumbering this file.

backend/internal/observe/usage/parser_pi.go
```diff
@@ -0,0 +1,105 @@
...
+
+func parsePi(source domain.UsageSourceContext, records []jsonlRecord, result *parseResult) {
+	eventsByKey := make(map[string]domain.ModelUsageEvent)
+	for _, record := range records {
+		var native piSessionRecord
+		if err := json.Unmarshal(record.Data, &native); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		if native.Type != "message" || native.Message == nil || native.Message.Role != "assistant" {
```
illegalcall: **[P1] Parse every Pi entry type that can carry usage, and prefer the actual response model.** This gate accepts only assistant messages, but Pi totals also include usage-bearing `toolResult`, `compaction`, and `branch_summary` entries. Assistant entries should be attributed with `responseModel ?? model`, not only `model`, or the collector can both undercount tokens and report the requested model instead of the responding model. Pi’s canonical behavior is documented in [usage-totals.ts](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/usage-totals.ts) and [session-format.md](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/session-format.md).

backend/internal/service/usage/collector.go
```diff
@@ -110,7 +115,34 @@ type Collector struct {
...
+	token chan struct{}
+}
+
+func newContextMutex() contextMutex {
+	token := make(chan struct{}, 1)
+	token <- struct{}{}
+	return contextMutex{token: token}
+}
+
+func (m *contextMutex) Lock() {
```
illegalcall: **[P2] Do not use a non-cancellable lock in context-bearing collection paths.** `Lock` can block indefinitely after its caller’s context is cancelled, which violates this repository’s rule that blocking/I/O work accepts and observes `context.Context`. Remove this path or replace every use from context-aware operations with `LockContext(ctx)` and propagate the cancellation error.

backend/internal/service/usage/collector.go
```diff
@@ -671,6 +750,71 @@ func (c *Collector) ReconcilePath(ctx context.Context, path string) error {
...
+	}
+	sessions, err := c.store.ListAllSessions(ctx)
+	if err != nil {
+		return err
+	}
+	matches := make([]domain.SessionRecord, 0, 1)
+	for _, session := range sessions {
+		if session.Harness != domain.HarnessPi || session.IsTerminated ||
+			session.Activity.State == domain.ActivityExited || session.Metadata.WorkspacePath == "" ||
+			startedAt.Before(session.CreatedAt.Add(-5*time.Second)) {
```
illegalcall: **[P2] This timestamp check does not uniquely attribute the native Pi session.** Any Pi session started later in the same worktree can match the one active AO session, including an unrelated/manual invocation. Once that candidate is bound, the different-ID check below prevents the correct session from replacing it. Please fence discovery to the AO launch/current-session window or use a stronger launch identity instead of workspace plus “started after AO session creation.”

illegalcall (changes requested):
Requesting changes because this head is not compatible with current `main`, and the collector still omits valid Pi usage records and can misattribute a same-workspace Pi session. The Pi-only rollback and cancellation-related improvements are good, and the exact-head usage tests pass. Remaining issues are inline.

backend/internal/service/usage/collector.go
nikhilachale: Removed workspace/timestamp attribution entirely. Pi discovery now binds only when the file header ID exactly matches the live AO session's durable native identity; pending discovery is additionally fenced to the current runtime launch. Added a regression where a later manual Pi session in the same workspace is ignored.

backend/internal/storage/sqlite/migrations/0118_allow_pi_usage.sql
nikhilachale: Rebased onto current main and moved the Pi-only constraint widening to migration 0117. The migration now preserves the current usage measurement, provider usage, provider hint, and cost schema; rollback removes only Pi-owned rows. Added a migrate-up/down regression test from version 0116.

backend/internal/observe/usage/parser_pi.go
nikhilachale: Updated the parser to include usage-bearing assistant, toolResult, compaction, and branch_summary entries. Assistant attribution now uses responseModel before model; tool and summary usage is retained as unbilled Tools/summaries. Added coverage for all four shapes plus Anthropic normalization and Pi cache-write pricing.

backend/internal/service/usage/collector.go
nikhilachale: Removed the non-cancellable lock path. Every context-bearing collection entry point now calls LockContext(ctx) and propagates cancellation. Added a contention regression test proving FinalizeSession returns context.Canceled while waiting.

backend/internal/storage/sqlite/migrations/0117_allow_pi_usage.sql
```diff
@@ -0,0 +1,197 @@
...
+DROP TRIGGER IF EXISTS usage_bindings_cdc_update;
+DROP TRIGGER IF EXISTS usage_sources_cdc_update;
+
+-- SQLite cannot widen a CHECK constraint in place. Rebuild only the two
+-- collection tables; the current event table (including measurement, bounded
+-- provider usage, billing attribution, and cost columns) remains untouched.
+CREATE TABLE usage_bindings_next (
+    id                 INTEGER PRIMARY KEY AUTOINCREMENT,
+    session_id         TEXT NOT NULL REFERENCES sessions (id) ON DELETE CASCADE,
+    harness            TEXT NOT NULL CHECK (harness IN ('claude-code', 'codex', 'pi')),
```
illegalcall: **[P1] Rebase this migration after Kimi lands and preserve the Kimi schema in both directions.** The PR says this migration is relative to Kimi and recommends merging Kimi before Pi, but this rebuilt table permits only Claude, Codex, and Pi; the source table likewise omits `kimi_wire`. After #4311 is applied, copying an existing Kimi binding/source into these tables fails the CHECK constraint, and even an empty upgrade removes Kimi from the allowed schema. Please rebase onto the merged Kimi migration, assign the next version, include Kimi in the Up and Down constraints, and add an upgrade/rollback test that seeds Kimi rows and verifies their IDs and events survive.

illegalcall (commented):
Follow-up review: the parser, native-ID attribution, cancellation, and current usage-schema fixes are resolved. One migration-sequencing blocker remains for the PR’s declared Kimi → Pi merge order.

backend/internal/storage/sqlite/migrations/0117_allow_pi_usage.sql
nikhilachale: Fixed in the rebased branch. Pi now follows Kimi as migration 0118, and both Up and Down retain the Kimi harness/source constraints. The migration test starts at 0117, seeds a Kimi binding/source/event, verifies their IDs and data survive the 0118 upgrade, then rolls back to 0117 and verifies Kimi still survives while only Pi rows are removed. I also resolved the current-main code conflicts and added follow-up coverage for late Pi transcript discovery, cacheWrite1h pricing, managed transcript paths, and legacy restore compatibility.

i-trytoohard:
@nikhilachale Marking as stale: no activity for 25 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4327, opened 2026-08-23
feat: add Qwen token usage observability

nikhilachale:
## What

Adds Qwen token-usage observability as a standalone change based on the latest `main`.

- parses Qwen monthly token-usage JSONL and attributes records by bound session ID
- discovers the current monthly usage file under `~/.qwen/usage`
- supports month rollover and missing-source retry
- persists and aggregates Qwen usage through the shared collector and SQLite store
- adds parser, discovery, rollover, migration, and storage coverage

## Why this PR was recreated

This replaces the earlier Qwen PR #4185 without carrying Kimi or Pi branch history. Qwen can now be reviewed, tested, merged, or reverted independently.

1. **Kimi:** #4311
2. **Pi:** #4326
3. **Qwen:** #4327 (this PR)

## Implementation

Qwen's shared monthly records are filtered by the bound session ID and normalized into the shared token model. Discovery selects the latest valid monthly file and continues across month rollover. Migration 0110 is intentionally relative to the Kimi and Pi migrations: it adds Qwen support on upgrade and removes only Qwen usage on rollback, preserving Kimi and Pi data and constraints.

## Testing

- `go test ./internal/observe/usage ./internal/service/usage ./internal/storage/sqlite ./internal/storage/sqlite/store -count=1`
- `go test ./internal/daemon -count=1`
- `go vet` over all changed backend packages
- golangci-lint v2.12.2 over all changed packages: `0 issues`
- `git diff --check`

The full backend run only encountered the repository's pre-existing timing-sensitive fake-agent lifecycle test under concurrent machine load. All directly affected packages pass.

## Checklist

- [x] Based on the latest `main`
- [x] Contains only Qwen provider logic

### DIFF
--- a/backend/internal/observe/usage/parser_qwen.go
+++ b/backend/internal/observe/usage/parser_qwen.go
@@ -1,0 +1,77 @@
+package usage
+
+import (
+	"encoding/json"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+type qwenUsageRecord struct {
+	SchemaVersion  int    `json:"schemaVersion"`
+	ID             string `json:"id"`
+	Timestamp      string `json:"timestamp"`
+	SessionID      string `json:"sessionId"`
+	Model          string `json:"model"`
+	InputTokens    int64  `json:"inputTokens"`
+	OutputTokens   int64  `json:"outputTokens"`
+	CachedTokens   int64  `json:"cachedTokens"`
+	ThoughtsTokens int64  `json:"thoughtsTokens"`
+	TotalTokens    *int64 `json:"totalTokens"`
+}
+
+func parseQwen(source domain.UsageSourceContext, records []jsonlRecord, result *parseResult) {
+	eventsByKey := make(map[string]domain.ModelUsageEvent)
+	for _, record := range records {
+		var native qwenUsageRecord
+		if err := json.Unmarshal(record.Data, &native); err != nil {
+			recordMalformed(result)
+			continue
+		}
+		if native.SessionID != source.NativeRootID {
+			continue
+		}
+		model := firstNonEmpty(native.Model)
+		identity := firstNonEmpty(native.ID)
+		if native.SchemaVersion != 1 || model == "" || identity == "" ||
+			native.InputTokens < 0 || native.OutputTokens < 0 ||
+			native.CachedTokens < 0 || native.ThoughtsTokens < 0 {
+			recordMalformed(result)
+			continue
+		}
+		output, ok := sumNonNegative(native.OutputTokens, native.ThoughtsTokens)
+		if !ok {
+			recordMalformed(result)
+			continue
+		}
+		reportedTotal := int64(0)
+		if native.TotalTokens != nil {
+			reportedTotal = *native.TotalTokens
+		}
+		tokens, details, ok := normalizeOpenAIUsage(
+			native.InputTokens, native.CachedTokens, 0, output, native.ThoughtsTokens, reportedTotal,
+		)
+		if !ok {
+			recordMalformed(result)
+			continue
+		}
+		event := domain.ModelUsageEvent{
+			ProviderID:      domain.UsageProviderOpenAI,
+			ModelID:         model,
+			Tokens:          tokens,
+			ProviderDetails: domain.UsageProviderDetails{OpenAI: &details},
+			CreatedAt:       parseUsageTimestamp(native.Timestamp),
+			SourceEventKey: stableSourceEventKey(
+				"qwen", source.NativeRootID, identity, model,
+			),
+		}
+		if existing, duplicate := eventsByKey[event.SourceEventKey]; duplicate {
+			if !usageEventsEqual(existing, event) {
+				result.Cursor.AnomalyCount++
+				result.Cursor.LastErrorCode = domain.UsageErrorSourceEventConflict
+			}
+			continue
+		}
+		eventsByKey[event.SourceEventKey] = event
+		result.Events = append(result.Events, event)
+	}
+}
--- a/backend/internal/observe/usage/parser_qwen.go
+++ b/backend/internal/observe/usage/parser_qwen.go
@@ -43,23 +43,25 @@ func parseQwen(source domain.UsageSourceContext, records []jsonlRecord, result *
 			recordMalformed(result)
 			continue
 		}
-		reportedTotal := int64(0)
-		if native.TotalTokens != nil {
-			reportedTotal = *native.TotalTokens
+		computedTotal, ok := sumNonNegative(native.InputTokens, output)
+		if !ok || native.TotalTokens != nil && (*native.TotalTokens < 0 || *native.TotalTokens != computedTotal) {
+			recordMalformed(result)
+			continue
 		}
-		tokens, details, ok := normalizeOpenAIUsage(
-			native.InputTokens, native.CachedTokens, 0, output, native.ThoughtsTokens, reportedTotal,
+		tokens, ok := normalizeOpenAIUsage(
+			native.InputTokens, native.CachedTokens, 0, output,
 		)
 		if !ok {
 			recordMalformed(result)
 			continue
 		}
 		event := domain.ModelUsageEvent{
-			ProviderID:      domain.UsageProviderOpenAI,
-			ModelID:         model,
-			Tokens:          tokens,
-			ProviderDetails: domain.UsageProviderDetails{OpenAI: &details},
-			CreatedAt:       parseUsageTimestamp(native.Timestamp),
+			ProviderID:        domain.UsageProviderOpenAI,
+			ModelID:           model,
+			MeasurementKind:   domain.UsageMeasurementNativeReported,
+			Tokens:            tokens,
+			ProviderUsageJSON: boundedProviderUsage(record.Data),
+			CreatedAt:         parseUsageTimestamp(native.Timestamp),
 			SourceEventKey: stableSourceEventKey(
 				"qwen", source.NativeRootID, identity, model,
 			),
--- a/backend/internal/service/usage/collector.go
+++ b/backend/internal/service/usage/collector.go
@@ -62,6 +62,7 @@ type SourceRoots struct {
 	ClaudeProjects string
 	CodexSessions  string
 	CodexArchived  string
+	QwenUsage      string
 }
 
 // DefaultSourceRoots resolves the native Claude Code and Codex transcript
@@ -82,6 +83,7 @@ func DefaultSourceRoots(ctx context.Context) (SourceRoots, error) {
 		ClaudeProjects: filepath.Join(home, ".claude", "projects"),
 		CodexSessions:  filepath.Join(codexHome, "sessions"),
 		CodexArchived:  filepath.Join(codexHome, "archived_sessions"),
+		QwenUsage:      filepath.Join(home, ".qwen", "usage"),
 	}, nil
 }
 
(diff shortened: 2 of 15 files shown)


backend/internal/storage/sqlite/migrations/0110_allow_provider_usage.sql
```diff
@@ -0,0 +1,204 @@
...
+
+DROP TRIGGER IF EXISTS usage_bindings_cdc_insert;
+DROP TRIGGER IF EXISTS usage_bindings_cdc_update;
+DROP TRIGGER IF EXISTS usage_sources_cdc_update;
+
+DELETE FROM anthropic_usage_event_details
+WHERE event_id IN (
+    SELECT event.id FROM model_usage_events event
+    JOIN usage_bindings binding ON binding.id = event.binding_id
+    WHERE binding.harness IN ('kimi', 'pi', 'qwen')
```
illegalcall: [P1] Make this rollback Qwen-only

The Down migration deletes Kimi, Pi, and Qwen events, then rebuilds the constraints for Claude/Codex only. Rolling back 0110 after the recommended merge sequence destroys unrelated Kimi/Pi history and invalidates their still-applied migrations. Please make Down remove only Qwen relative to the previous schema, or use a single shared prerequisite migration with a composable rollback.

backend/internal/service/usage/collector.go
```diff
@@ -332,8 +335,11 @@ func (c *Collector) RecordHook(ctx context.Context, sessionID domain.SessionID,
 
 	if mainArtifact != nil {
 		kind := domain.UsageSourceClaudeMain
-		if session.Harness == domain.HarnessCodex {
+		switch session.Harness {
+		case domain.HarnessCodex:
 			kind = domain.UsageSourceCodexRollout
+		case domain.HarnessQwen:
```

This mapping is repeated in the hook, backfill, and reconcile paths, with additional provider cascades elsewhere. A new harness can therefore be partially wired and silently default to Claude on another path. Please centralize the provider capabilities/source-kind mapping and cover it once.

illegalcall (changes requested):
Requesting changes based on the inline findings below. The targeted tests pass, but the rollback currently destroys sibling-provider usage and is not safe to merge independently.

backend/internal/storage/sqlite/migrations/0110_allow_provider_usage.sql
nikhilachale: Fixed. Migration 0110 is now relative to Kimi and Pi: Down deletes only Qwen usage and rebuilds constraints that retain Kimi/Pi data. Added an up/down regression test covering all three providers.

backend/internal/service/usage/collector.go

backend/internal/storage/sqlite/migrations/0110_allow_provider_usage.sql
nikhilachale: Fixed. Migration 0110 is now relative to Kimi and Pi: Down deletes only Qwen usage and restores constraints that continue to allow Kimi and Pi. Added an up/down regression test verifying both sibling providers survive.

backend/internal/service/usage/collector.go

backend/internal/storage/sqlite/migrations/0117_allow_qwen_usage.sql
```diff
@@ -0,0 +1,207 @@
+-- +goose NO TRANSACTION
```
illegalcall, line 1: **[P1] Rebase this feature onto current `main` and allocate a fresh migration version.** Current `main` already owns 0110 (and migrations through 0116), so this migration and the burned-version ledger collide. Current `main` also changed the usage representation in migration 0115 to `usage_measurement_kind` and `provider_usage_json`; please port Qwen support onto that schema and update the rollback target rather than only renumbering this file.

backend/internal/service/usage/collector.go
```diff
@@ -80,6 +81,7 @@ func DefaultSourceRoots(ctx context.Context) (SourceRoots, error) {
 		ClaudeProjects: filepath.Join(home, ".claude", "projects"),
 		CodexSessions:  filepath.Join(codexHome, "sessions"),
 		CodexArchived:  filepath.Join(codexHome, "archived_sessions"),
+		QwenUsage:      filepath.Join(home, ".qwen", "usage"),
```
illegalcall: **[P2] Resolve Qwen’s usage directory from its supported runtime base instead of hardcoding `$HOME/.qwen`.** Qwen writes usage below `Storage.getRuntimeBaseDir()/usage`, which can honor `QWEN_RUNTIME_DIR`, `QWEN_HOME`, or configured runtime storage; AO authentication also recognizes `QWEN_HOME`. With the current path, valid usage silently disappears for those installations. Please mirror the provider’s precedence and add coverage for the environment overrides. See Qwen’s [token usage service](https://github.com/QwenLM/qwen-code/blob/main/packages/core/src/services/tokenUsageService.ts) and [storage resolution](https://github.com/QwenLM/qwen-code/blob/main/packages/core/src/config/storage.ts).

illegalcall (changes requested):
Requesting changes because this head is not compatible with current `main`, and usage discovery misses Qwen’s supported custom runtime locations. The provider-specific rollback fix looks good, and the exact-head usage tests pass. Remaining issues are inline.

backend/internal/storage/sqlite/migrations/0117_allow_qwen_usage.sql
nikhilachale: Fixed in 189ca324b. The branch is rebased onto current main, Qwen now uses migration 0117, and the migration preserves the current provider_hint plus 0115/0116 event schema. Its Down path targets 0116, deletes only Qwen events, and retains Kimi/Pi usage; the rollback regression now seeds usage_measurement_kind/provider_usage_json.

backend/internal/service/usage/collector.go
nikhilachale: Fixed in 189ca324b. Qwen usage discovery now mirrors the runtime-base precedence: QWEN_RUNTIME_DIR, then advanced.runtimeOutputDir from the resolved Qwen settings, then QWEN_HOME, then ~/.qwen. Added coverage for both environment overrides and configured runtime storage.

backend/internal/storage/sqlite/migrations/0117_allow_qwen_usage.sql
```diff
@@ -0,0 +1,197 @@
...
+BEGIN IMMEDIATE;
+
+DROP TRIGGER IF EXISTS usage_bindings_cdc_insert;
+DROP TRIGGER IF EXISTS usage_bindings_cdc_update;
+DROP TRIGGER IF EXISTS usage_sources_cdc_update;
+
+CREATE TABLE usage_bindings_next (
+    id                 INTEGER PRIMARY KEY AUTOINCREMENT,
+    session_id         TEXT NOT NULL REFERENCES sessions (id) ON DELETE CASCADE,
+    harness            TEXT NOT NULL CHECK (harness IN ('claude-code', 'codex', 'kimi', 'pi', 'qwen')),
```
illegalcall, line 15: **[P2] Make this migration relative to the schema that actually precedes it.** This branch is based on migration 0116, whose usage constraints contain only Claude and Codex, but this Up migration introduces Kimi and Pi as well as Qwen and the Down path “restores” Kimi/Pi. Rolling back to 0116 therefore leaves a schema that 0116 never defined and admits unsupported sibling-provider rows. If Qwen is intentionally third in the sequence, rebase after Kimi and Pi land and use the next migration version. Otherwise, keep this migration standalone by adding only Qwen to the 0116 schema and restoring only Claude/Codex on Down.

backend/internal/service/usage/collector.go
```diff
@@ -78,13 +79,62 @@ func DefaultSourceRoots(ctx context.Context) (SourceRoots, error) {
...
+	if runtimeDir := strings.TrimSpace(os.Getenv("QWEN_RUNTIME_DIR")); runtimeDir != "" {
+		return resolveQwenStoragePath(runtimeDir, home)
+	}
+
+	qwenHome := filepath.Join(home, ".qwen")
+	if configuredHome := strings.TrimSpace(os.Getenv("QWEN_HOME")); configuredHome != "" {
+		qwenHome = resolveQwenStoragePath(configuredHome, home)
+	}
+
+	data, err := os.ReadFile(filepath.Join(qwenHome, "settings.json"))
```
illegalcall, line 101: **[P3] Thread cancellation through this settings read.** `qwenRuntimeBaseDir` performs filesystem I/O without accepting `context.Context`, despite being called from cancellation-aware daemon setup. Please take `ctx` as the first argument, check it around `os.ReadFile`, propagate cancellation, and add a canceled-context test, matching the repository rule for blocking/I/O functions.

backend/internal/service/usage/collector.go
```diff
@@ -78,13 +79,62 @@ func DefaultSourceRoots(ctx context.Context) (SourceRoots, error) {
...
+	if strings.HasPrefix(path, "~/") || strings.HasPrefix(path, `~\`) {
+		segments := strings.FieldsFunc(path[2:], func(r rune) bool {
+			return r == '/' || r == '\\'
+		})
+		return filepath.Join(append([]string{home}, segments...)...)
+	}
+	if filepath.IsAbs(path) {
+		return filepath.Clean(path)
+	}
+	resolved, err := filepath.Abs(path)
```
illegalcall, line 131: **[P2] Resolve configured runtime output per Qwen workspace, not the AO daemon’s working directory.** Qwen passes the project cwd when resolving a relative `advanced.runtimeOutputDir`, and its merged settings can come from the workspace. This helper reads only the global settings file and `filepath.Abs` resolves a value such as `.qwen-runtime` against the daemon cwd, so AO watches a different directory from Qwen and silently misses usage. Resolve the provider root from each AO session’s workspace using Qwen’s environment/settings precedence, and cover a workspace-local relative setting. See [Storage.resolvePath/setRuntimeBaseDir](https://github.com/QwenLM/qwen-code/blob/main/packages/core/src/config/storage.ts) and [CLI configuration](https://github.com/QwenLM/qwen-code/blob/main/packages/cli/src/config/config.ts).

backend/internal/observe/usage/parser_qwen.go
```diff
@@ -0,0 +1,79 @@
...
+			recordMalformed(result)
+			continue
+		}
+		output, ok := sumNonNegative(native.OutputTokens, native.ThoughtsTokens)
+		if !ok {
+			recordMalformed(result)
+			continue
+		}
+		computedTotal, ok := sumNonNegative(native.InputTokens, output)
+		if !ok || native.TotalTokens != nil && (*native.TotalTokens < 0 || *native.TotalTokens != computedTotal) {
```
illegalcall, line 47: **[P2] Do not reject a native record merely because the reported total differs from the component sum.** Qwen uses a positive provider-reported `total_token_count` when available and only derives a sum when that value is absent; its record validator does not impose this equality. A provider-valid record can therefore be marked malformed here and lose all usage. Validate non-negative component buckets, normalize those buckets into AO metrics, and retain the provider-reported total in `ProviderUsageJSON`. Add a fixture with a positive differing total. See Qwen’s [canonical record creation and validation](https://github.com/QwenLM/qwen-code/blob/main/packages/core/src/services/tokenUsageService.ts).

illegalcall (commented):
Follow-up review: the absolute `QWEN_HOME`/`QWEN_RUNTIME_DIR` handling and current event-schema port are improved. Four remaining items are called out inline: predecessor-schema correctness, cancellable settings I/O, workspace-relative runtime discovery, and valid native total handling.

i-trytoohard:
@nikhilachale Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #3858, opened 2026-08-11
feat: enable Claude Code chat message editing

nikhilachale:
## Summary

- add a provider-neutral `prompt_replay` chat capability for editing earlier prompts when an anchored native fork is unavailable
- advertise the capability for Claude Code ACP and pass AO's durable pre-edit user/assistant transcript into a fresh Claude session as delimited context
- enable the existing Chat UI edit affordance for native `fork` or `prompt_replay`
- preserve native anchored fork as the preferred path and retain the existing unsupported response for providers offering neither capability

## Why

Claude Code's current ACP surface does not expose the historical provider-turn anchor required by AO's exact edit-and-branch flow. This adds an explicit AO-local approximation: AO starts a fresh Claude session and supplies only durable textual conversation context before the edited prompt.

Tool calls, approvals, diffs, and file-change activities are intentionally not replayed, avoiding repeated side effects. The existing worktree remains unchanged.

## Validation

- `go test ./internal/service/chat ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/claudeacp`
- `npm run typecheck` in `frontend/`
- launched the Electron development app against the isolated dev database and verified the daemon/API became healthy

PR #4328, opened 2026-08-23
fix(chat): restore navigation to root conversation branches

nikhilachale:
## What

Fix conversation branch navigation so a user can move from an edited branch (`2 / 2`) back to the root branch (`1 / 2`). Also reset the Chat surface when navigating between sessions so conversation-scoped mutation state cannot leak across tasks.

## Root cause

Root branch IDs use the form `<conversation-id>:root`. The generated frontend client correctly percent-encodes the colon as `%3A`, but Chi exposes that route parameter in encoded form. The controller passed the encoded value directly to the service, so SQLite searched for `<conversation-id>%3Aroot`, did not find it, and returned `CHAT_BRANCH_NOT_FOUND` even though `<conversation-id>:root` existed.

## How

- Decode the branch route parameter at the HTTP controller boundary before the conversation-scoped service lookup.
- Add a router-level regression test proving an encoded `%3Aroot` ID reaches the service as `:root`.
- Key `SessionChatSurface` by `session.id` so changing tasks creates a fresh conversation hook boundary.
- Add a frontend regression test covering that remount.

## Verification

- `go test ./internal/adapters/agent/fake -run TestFullLifecycleSpawnToTermination -count=1`
- `npm run typecheck` in `frontend/`
- `npm test -- --maxWorkers=1` in `frontend/` — 2,773 passed, 1 skipped
- Rebuilt and launched the isolated AO Electron app, restored the affected scratch session, activated its child branch, then activated the root through the formerly failing `%3Aroot` URL; both requests returned `202 Accepted` and the root response contained the decoded `:root` ID.

The full backend sweep encountered the unrelated timing-sensitive `TestFullLifecycleSpawnToTermination` failure under concurrent load; it passed immediately when rerun in isolation.

## Checklist

- [x] Branched from `main`
- [x] One focused bug fix
- [x] Regression tests added for the affected HTTP and UI boundaries
- [x] Relevant backend and frontend checks pass

### DIFF
--- a/backend/internal/httpd/controllers/conversation_history_test.go
+++ b/backend/internal/httpd/controllers/conversation_history_test.go
@@ -218,6 +218,19 @@ func TestActivateConversationBranchRouteResumesWithoutBody(t *testing.T) {
 	}
 }
 
+func TestActivateConversationBranchRouteDecodesEscapedBranchID(t *testing.T) {
+	svc := &fakeChatService{activate: "conversation-1:root"}
+	srv := newChatTestServer(t, svc)
+	body, status, _ := doRequest(t, srv, http.MethodPost,
+		"/api/v1/sessions/ao-1/conversation/branches/conversation-1%3Aroot/activate", "")
+	if status != http.StatusAccepted {
+		t.Fatalf("status = %d, body = %s", status, body)
+	}
+	if svc.gotBranchID != "conversation-1:root" {
+		t.Fatalf("branch id = %q, want decoded root id", svc.gotBranchID)
+	}
+}
+
 func TestActivateConversationBranchRouteReportsMissingBranch(t *testing.T) {
 	svc := &fakeChatService{activateErr: domain.ErrNoConversationBranch}
 	srv := newChatTestServer(t, svc)
--- a/backend/internal/httpd/controllers/conversations.go
+++ b/backend/internal/httpd/controllers/conversations.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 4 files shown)


nikhilachale:
https://github.com/user-attachments/assets/7cb6ffcf-3013-458b-be1e-7f51c0c38514

illegalcall:
## Independent native QA confirmation (MQA-07)

The 2026-08-25 ChatUI pass reproduced this exact bug on baseline `d71acca1e`: editing an earlier Codex message created branch **2 / 2**, but Previous advertised `<conversation-id>:root` and the activation route returned `CHAT_BRANCH_NOT_FOUND`. The durable worktree was unchanged.

Previous branch activation fails

A tagged controller regression in [PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419) is deliberately red on current main and asserts that encoded `%3Aroot` reaches the service as decoded `:root`. I will also run that contract against this PR head in an isolated worktree; passing it is the MQA-07 acceptance gate.

illegalcall:
## MQA-07 verification: pass

I stacked this PR’s two commits onto the latest ChatUI harness head (`a8f10939d`) in an isolated detached worktree and reran the regression.

- MQA-07 Playwright route contract: **1/1 passed**
- `TestChatUIRegressionEncodedSyntheticBranchIDIsDecoded`: **passed** (red on main, green with this PR)
- `go test ./internal/httpd/controllers -count=1`: **passed**
- `SessionView.test.tsx`: **75/75 passed**
- frontend `npm run typecheck`: **passed**
- CI on the PR is already green

Local evidence: `/private/tmp/chatui-pr4328-verify.dIZx9V/e2e-artifacts/chatui-regression/2026-08-25T12-21-43-538Z` (not committed). The aggregate capture remains red only because the unrelated MQA-06 tagged contract is intentionally included by the runner. This PR satisfies the MQA-07 acceptance contract.

illegalcall (changes requested):
Requesting changes because this branch is stale relative to current main. The branch-ID decoding fix and its router regression have already landed through #3858, so the effective merge result for this PR now changes only SessionView.tsx and SessionView.test.tsx to remount the Chat surface between sessions. Please rebase onto main, drop the already-landed backend commit, and retitle/rewrite the PR around the remaining frontend behavior (or close it if that behavior is covered elsewhere). As it stands, the title and verification describe a backend navigation fix that this PR no longer contributes to the target branch.

i-trytoohard:
@nikhilachale Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

i-trytoohard:
@nikhilachale Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

codebanditssss:
Thanks for the contribution. The backend branch-ID decoding portion landed through #3858, and current `main` now also keys `SessionChatSurface` by `session.id` in `SessionView.tsx`, providing the remaining fresh-state-on-navigation behavior. This PR therefore no longer has a necessary behavior change to contribute.

PR #4329, opened 2026-08-23
feat(sidebar): remember expanded projects

ronishrohan:
## Summary
- persist expanded project IDs in local storage
- restore only previously expanded projects; empty storage starts all collapsed
- suppress initial animations for restored project session lists

image

## Validation
- `git diff --check`
- Sidebar suite: 64 tests passed
- Full typecheck is blocked by missing workspace dependencies and existing packages/product-ui errors

### DIFF
--- a/frontend/src/renderer/components/Sidebar.tsx
+++ b/frontend/src/renderer/components/Sidebar.tsx
@@ -210,6 +210,9 @@ export function Sidebar({
 	const daemonStatus = useShellMaybe()?.daemonStatus ?? null;
 	const commandPaletteEnabled = useCommandPaletteEnabled();
 	const setCommandPaletteOpen = useUiStore((s) => s.setCommandPaletteOpen);
+	const initialActiveSessionProjectId = useRef(
+		selection.activeSessionId ? selection.activeProjectId : undefined,
+	).current;
 	useLayoutEffect(() => {
 		// Offcanvas: the panel slides off-screen on collapse — no need to hide content.
 		// Reveal immediately on expand so there's no fade-in delay.
@@ -221,6 +224,9 @@ export function Sidebar({
 	// Disclosure state is persisted as the IDs of projects that were expanded.
 	// An empty/missing store intentionally means all projects start collapsed.
 	const [expandedIds, setExpandedIds] = useState<ReadonlySet<string>>(() => readExpandedProjectIds());
+	const [dismissedInitialActiveProjectIds, setDismissedInitialActiveProjectIds] = useState<ReadonlySet<string>>(
+		() => new Set(),
+	);
 	const toggleCollapsed = (id: string) =>
 		setExpandedIds((prev) => {
 			const next = new Set(prev);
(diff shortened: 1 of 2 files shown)


neversettle17-101 (commented):
Requesting changes — persisting expanded projects is a nice UX improvement, but there is a regression when the sidebar loads on a worker-session route with empty storage: the active worker session is hidden and nothing stays highlighted until the user manually expands its project.

Please auto-expand the selected project (or otherwise keep the active worker visible) on initial render, and add coverage for that deep-link case.

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -200,13 +211,16 @@ export function Sidebar({
 		}
 	}, [isCollapsed]);
 
-	// Disclosure state: projects are expanded by default; a project id present in
-	// this set is collapsed (sessions hidden).
-	const [collapsedIds, setCollapsedIds] = useState<ReadonlySet<string>>(() => new Set());
+	// Disclosure state is persisted as the IDs of projects that were expanded.
+	// An empty/missing store intentionally means all projects start collapsed.
+	const [expandedIds, setExpandedIds] = useState<ReadonlySet<string>>(() => readExpandedProjectIds());
```
neversettle17-101, line 226: With an empty/missing `ao.sidebar.expanded-projects` store, every project renders collapsed. That regresses direct opens of `/projects/:projectId/sessions/:sessionId` for worker sessions: `ProjectItem` only treats the project row as active for the dashboard/orchestrator case, so the selected worker row is now hidden and the sidebar loses all active highlighting on first load. Please force the currently selected project open (or otherwise surface the active worker session) before shipping this behavior.

ronishrohan:
Validated and fixed in commit `3cd166e82`. The comment was valid: an empty expanded-project store hid the active worker on a direct session deep link. The project now auto-expands only for the initial deep-linked session, and a manual collapse overrides that temporary reveal. Added regression coverage. Focused Sidebar suite: 65 passed.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/components/Sidebar.tsx
```diff
@@ -393,9 +420,17 @@ export function Sidebar({
 									<ProjectItem
 										key={workspace.id}
 										workspace={workspace}
-										expanded={!collapsedIds.has(workspace.id)}
+										expanded={
```
neversettle17-101, line 433: On a worker-session deep link with empty storage, this auto-expanded project cannot actually be collapsed. The first folder click adds the project to `expandedIds` while also dismissing the initial-route fallback, so `expandedIds.has(workspace.id)` keeps it open. The next click removes it from `expandedIds` but re-enables the fallback, so it remains open again. Please make the route fallback and persisted disclosure state use one toggle path, or special-case the auto-expanded/not-persisted state so the first click only dismisses the fallback instead of persisting the project as expanded.

ronishrohan: Validated and fixed in commit `5adaf491c`. The persisted disclosure state and the initial deep-link fallback now use one effective-state toggle: the first click collapses the route-fallback project and persists `[]`, and the next click reopens it. Added regression coverage for the deep-link collapse path. Sidebar suite: 65 tests passed.

neversettle17-101 (approved):

PR #4330, opened 2026-08-23
fix(desktop): make startup layout render together

ronishrohan:
## Summary
- keep a fast startup preflight for Git, tmux, and optional `gh` using only local `PATH` lookups
- move the expensive fresh agent inventory/auth probe off the first-render path
- serve durable workspace metadata before slow session/worktree restoration, and reveal the sidebar and center pane together

## Validation
- `cd backend && go test ./internal/service/systemcheck ./internal/httpd/controllers ./internal/httpd ./internal/daemon`
- `cd backend && go vet ./... && go build ./...`
- `cd frontend && npx vitest run src/renderer/__tests__/integration/board-empty-states.test.tsx src/renderer/components/DaemonStartupLoader.test.tsx`

## Known unrelated check failures
- `cd backend && go test ./...` has existing environment-dependent failures in Crush local auth, Codex app-server protocol fixtures, and tmux integration/activity/terminal tests. Touched backend packages pass.
- `cd frontend && npm run typecheck` remains blocked by unresolved dependencies in `packages/product-ui` (`react`, `motion/react`, etc.).
- `cd frontend && npm test` also fails in existing landing analytics/markdown-twin tests; the focused startup tests pass.


illegalcall:
Follow-up after reproducing the reported startup regression on this PR:

- The new `CheckStartup` path correctly keeps `/system/requirements` to executable presence checks; a fault-injected regression test proved it fails if the full auth/catalog path is restored.
- The PR still had two eager full refreshes: daemon boot called `agentSvc.Refresh`, and the renderer shell called `refreshAgents` as soon as the daemon became ready. Commit 845775cb0 removes both, leaving the existing `refreshAgentsIfStale` calls on agent-selection surfaces as the on-demand trigger.
- Added coverage proving startup prefers the presence-only resolver and calls neither normal agent resolution nor auth.
- Scoped the router startup branding to the initial unresolved location so later route loads cannot show the startup screen again.

Validation on 845775cb0:
- focused startup/selection frontend suite: 103/103 passed
- touched backend packages under `go test -race`: passed
- frontend typecheck and production renderer build: passed
- `go vet ./...`, `go build ./...`, and golangci-lint: passed
- native Electron isolated-data smoke test: daemon at 127.0.0.1:3002, renderer at localhost:5173, workspace/session traffic successful, requirements response under 1 ms, and no `/agents/refresh` request at startup
- full frontend suite: 2714 passed; only the already-documented landing test dependency failures remained
- full backend suite: one load-sensitive fake-agent timing assertion failed in the parallel run, then passed 3/3 in isolation

illegalcall:
## A/B verification — native Electron

Tested both revisions locally on the same machine with the same isolated AO dev profile, native Electron launch, daemon at `127.0.0.1:3002`, and renderer at `localhost:5173`. A capture watcher was armed before the first Electron window; the values below are its recorded frame timestamps.

| | Before | After |
|---|---:|---:|
| Revision | `6cba6344c` (PR merge-base) | `845775cb0` (current PR head) |
| Board visible by | 4.98s frame | 2.13s frame |
| `/system/requirements` startup request | 2.652s | 0.640ms |
| Median of 5 identical endpoint probes | 2.240s | 1.205ms |

### Before — regression reproduced

At the 4.38s frame, the app was still blocked on **Checking your setup**. It reached the board between the 4.38s and 4.98s frames.

PR 4330 before fix: still checking setup at 4.38 seconds

Five endpoint samples: `2.240s, 1.908s, 3.026s, 2.292s, 1.620s` (median `2.240s`).

### After — fixed

The initial **Starting local services** screen cleared and the board was visible by the 2.13s frame; the slow **Checking your setup** state did not appear.

PR 4330 after fix: board ready at 2.13 seconds

Five endpoint samples: `1.211ms, 1.738ms, 0.976ms, 1.059ms, 1.205ms` (median `1.205ms`). The startup log also showed no `/api/v1/agents/refresh` request.

**Result:** in this native A/B run the board appeared about **2.84s earlier** (4.98s → 2.13s), while the median requirements probe improved by about **1,860×**. This reproduces the reported delay before the fix and confirms the current PR head removes the agent-auth sweep from startup.

illegalcall (approved):

PR #4332, opened 2026-08-23
fix: smooth chat streaming without extra UI chrome

ronishrohan:
## Summary

- replace centered status separators with friendly two-row messages and a full-width rule
- handle Escape as an interruption only while a turn is actively running
- ignore Escape when a Radix dialog/menu or pending approval/elicitation is open
- remove the streaming caret/spinner while keeping text-only streaming
- render streamed assistant text through an adaptive buffer that preserves fractional rates, clamps hidden-tab frame deltas, catches up growing backlogs, and flushes on completion
- keep initial streamed snapshots visible, preserve provider corrections, and restore action controls immediately when streaming ends
- keep neutral compaction statistics neutral and preserve recovered-turn messaging

https://github.com/user-attachments/assets/74fafe21-2658-479e-937f-7ca7f88d4418

This is frontend-only. No backend, API, storage, agent, file-editing, reading, tool-call, or command-execution code was changed.

## Verification

- focused chat tests: 79 passed
- git diff --check: passed
- direct fake-RAF coverage added for fractional accumulation, high-refresh frames, occluded tabs, completion flush, cleanup behavior, and action visibility
- Escape coverage added for active turns, menus, approvals, and elicitations
- reviewed adjacent tool-call, file-change, plan, approval, elicitation, and command activity branches; they remain outside the streaming renderer
- npm run typecheck remains blocked by existing missing workspace dependencies/types in packages/product-ui
- this checkout has no frontend build script


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/components/chat/ChatTimelineItems.test.tsx
```diff
@@ -1,13 +1,148 @@
...
+
+	it("hides message actions while text is still buffered", () => {
+		const view = render(<AssistantMessage message={message()} showCopy />);
+		view.rerender(<AssistantMessage message={message({ text: "a buffered answer" })} showCopy />);
+
+		expect(screen.queryByRole("button", { name: "Copy message as markdown" })).not.toBeInTheDocument();
+	});
+
+	it("shows the message timestamp on hover", () => {
+		const view = render(
```
illegalcall: The latest Frontend workflow fails with TS6133 here because `view` is unused after the title assertion was removed. Please call `render(...)` without assigning the result so typecheck passes.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -110,6 +109,112 @@ const ATTACHMENT_REFERENCE_BLOCK =
...
+			const count = Math.floor(fractionalCharactersRef.current);
+			if (count < 1) {
+				frameRef.current = window.requestAnimationFrame(tick);
+				return;
+			}
+			fractionalCharactersRef.current -= count;
+			const current = visibleRef.current;
+			const target = targetRef.current;
+			if (current.length >= target.length) return;
+			const next = current + target.slice(current.length, current.length + count);
```
illegalcall: Please avoid slicing streamed text by UTF-16 code units. A one-unit frame can expose half an emoji or part of a combined grapheme. Advance using `Intl.Segmenter` grapheme boundaries and add focused coverage for an emoji and a combining-character sequence.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -110,6 +109,112 @@ const ATTACHMENT_REFERENCE_BLOCK =
...
+	const cancelDrain = useCallback(() => {
+		if (frameRef.current !== undefined) {
+			window.cancelAnimationFrame(frameRef.current);
+			frameRef.current = undefined;
+		}
+		lastFrameAtRef.current = undefined;
+		fractionalCharactersRef.current = 0;
+	}, []);
+
+	const scheduleDrain = useCallback(() => {
```
illegalcall, line 165: Please honor `prefers-reduced-motion` in this custom animation path. When reduction is requested, display the latest provider snapshot directly instead of running the character drain, and add a focused test for that behavior.

frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
```diff
@@ -363,6 +354,89 @@ describe("ChatWorkspace timeline", () => {
...
+			summary: "Choose a direction",
+			requestId: "input-1",
+			detail: { inputMode: "form", message: "Choose a direction" },
+			createdAt: "2026-08-24T00:00:00Z",
+		});
+		render(<ChatWorkspace snapshot={snapshot} onInterrupt={onInterrupt} />);
+		fireEvent.keyDown(screen.getByLabelText("Message the agent"), { key: "Escape" });
+		expect(onInterrupt).not.toHaveBeenCalled();
+
+		const menu = document.createElement("div");
```
illegalcall, line 440: This menu assertion is masked by the pending `user_input` added above: `hasPendingInteraction` already prevents interruption even if menu detection is broken. Please test the menu and elicitation guards with separate snapshots, leaving no pending interaction in the menu case.

illegalcall (changes requested):
The core streaming approach looks good, and I am not requesting structural or styling changes. Please address the four concrete correctness and test issues in the inline comments. The current Frontend workflow is also failing because of the unused test variable called out below.

frontend/src/renderer/components/chat/ChatTimelineItems.test.tsx
ronishrohan: Fixed in `947acde2e`: the timestamp test now calls `render(...)` without assigning the unused return value.

frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
ronishrohan: Fixed in `4e30dcddb`: the elicitation and menu Escape guards now use separate tests, and the menu case has no pending interaction to mask menu detection.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
ronishrohan: Fixed in `947acde2e`: `prefers-reduced-motion: reduce` now shows the latest provider snapshot immediately and skips the custom drain; focused coverage was added.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
ronishrohan: Fixed in `947acde2e`: streaming now advances through `Intl.Segmenter` grapheme boundaries, with coverage for the family emoji and combining accent sequence.

illegalcall: Thanks — this fixes splitting within one received snapshot, but grapheme boundaries can change across snapshots. For example, visible `a👨` is `[a, 👨]`; the next target `a👨‍👩x` is `[a, 👨‍👩, x]`. The string-prefix check succeeds, then slicing the new target from the old grapheme count appends only `x`, producing `a👨x`; without `x`, equal array lengths schedule no drain. Please compare the actual grapheme prefix and reconcile a changed trailing grapheme before using the old count. Add a test where the ZWJ or combining mark arrives in a later snapshot rather than in the same update.

ronishrohan: Fixed in `3dff99b33`: the stream now compares the visible text against the target's actual grapheme prefix and retracts a changed trailing grapheme before draining. Added regression coverage where the ZWJ arrives in a later snapshot; 81 focused tests and frontend typecheck pass.

illegalcall: The cross-snapshot correctness issue is fixed, but `reconciledStreamPrefix` now does quadratic work: each grapheme creates a progressively longer `candidate` and `startsWith` compares that whole prefix, on every streamed snapshot. Benchmarking this exact loop took about 22 ms at 5k ASCII graphemes, 323 ms at 10k, and 1.33 s at 20k, which can freeze a long streaming reply. The caller has already established `message.text.startsWith(visibleRef.current)`, so this can be linear: accumulate target-grapheme UTF-16 lengths until the next boundary exceeds `visibleText.length`, then slice `visibleText` once at the last valid boundary. Please make that change while retaining the new cross-snapshot regression test.

illegalcall (commented):
Checked current head `22f730ce0`. Two requested changes remain:

1. Replace `reconciledStreamPrefix` in `frontend/src/renderer/components/chat/ChatTimelineItems.tsx:123-130` with the linear boundary-length scan described in the existing thread. The quadratic prefix concatenation/`startsWith` loop is unchanged, so the streaming performance request is not fixed.

2. Remove commit `22f730ce0` from this PR and submit it separately. It adds backend ACP agent-mode normalization in `backend/internal/adapters/chatdriver/acp/config_options.go` plus unrelated tests, contradicting this PR body’s explicit frontend-only/no backend-agent changes statement.

All checks are green, but these two code/scope changes are still needed before approval.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
ronishrohan: Fixed in `89da70adf`: `reconciledStreamPrefix` now performs one linear pass over target grapheme lengths and slices once at the last boundary within the visible UTF-16 length. The delayed-ZWJ cross-snapshot regression remains in place; 81 focused tests and frontend typecheck pass.

ronishrohan:
Addressed the remaining scope request. Commit `22f730ce0` has been removed from this PR, and #4332 is frontend-only again with no backend files in its diff.

illegalcall (approved):
Re-reviewed at `0208d66e1`. The streamed-grapheme reconciliation is now linear and remains correct for cross-snapshot grapheme changes, the unrelated backend agent-mode commit has been removed, and the PR is frontend-only again. All requested changes are addressed and checks are green.

Issue #4335, opened 2026-08-24
bug(daemon): schema drift is swallowed into a WARN while /readyz reports ready, turning ao spawn into an opaque INTERNAL_ERROR


axisrow:
## Summary

When the physical schema drifts from what `goose_db_version` claims is applied, the daemon swallows the failure into a single startup `WARN`, keeps reporting `healthz: ok` / `readyz: ready`, and `ao spawn` — the product's core function — fails with a bare `Internal server error (INTERNAL_ERROR)` and no path to the cause.

This is not a report about *how* a schema drift happens (that mechanism is #3476). It is about the daemon's reaction to a drift that already exists: three independent design choices compound into an unrecoverable-looking failure with no diagnostic surface.

## Steps to reproduce

Any database where an object is missing while its migration version is recorded as applied reproduces this. On a **copy** of a dev `ao.db`:

1. Drop an object that a later code path depends on, leaving `goose_db_version` untouched:
   ```sql
   DROP TABLE agent_inventory_cache;
   ```
2. Restart the daemon.
3. `ao status` → reports `healthz: ok`, `readyz: ready`.
4. `ao spawn --project <id> --name <name> --kind worker`

The same shape reproduces with any other drifted object; #3476 documents an independent instance on `sessions.diff_base_sha` / `diff_base_ref` reached through migration-version collision rather than a manual drop.

## Expected behavior

A daemon whose schema cannot support its core workflow should either refuse to come up, or report itself as not ready — and the resulting CLI failure should name the actual problem (missing table/column, schema out of sync) rather than `INTERNAL_ERROR`.

## Actual behavior

The daemon starts, reports healthy, and spawn fails opaquely.

**Startup — the only trace of a broken schema, at WARN level:**

```
level=WARN msg="initial agent catalog refresh failed"
  err="upsert agent inventory cache: SQL logic error: no such table: agent_inventory_cache (1)"
```

**Health probes, after that warning:**

```
AO daemon: ready
  healthz: ok
  readyz: ready
```

**Spawn:**

```
$ ao spawn --project <id> --name <name> --kind worker
Internal server error (INTERNAL_ERROR) [request <id>]
```

**Daemon log for that request:**

```
method=POST path=/api/v1/agents/refresh status=500
  error="upsert agent inventory cache: SQL logic error: no such table: agent_inventory_cache (1)"
```

[...]

PR #4336, opened 2026-08-24
fix(storage): recreate tables and columns skipped by burned migration versions

axisrow:
## What

`reconcileSchema` (`backend/internal/storage/sqlite/db.go`) only verified **columns** from the manual `schemaRepairs` list, via `pragma_table_info`. A whole missing **table** (or a column outside the list) passed the boot check as healthy: the daemon came up green, and the first query against the missing object 500'd — for `agent_inventory_cache` that is every `ao spawn` (`INTERNAL_ERROR`) while `/readyz` stays ready.

This adds:
1. `tableRepairs` — a table-level counterpart to `schemaRepairs`, checked against `sqlite_master` and recreated when absent (idempotent, same as the column path).
2. Repair entries for the two drift casualties found in the field:
   - `0104` `agent_inventory_cache` (whole table missing)
   - `0106` `pr_comment.review_id` (column missing; `preparePRCommentReviewIDMigration` only repaired the mirror case — column present, ledger entry missing)

## Why

Part of #4335: a schema drift must not survive into a green daemon whose core workflow (`ao spawn`) 500s. This is the reconciliation gap, not the WARN/`/readyz` items the issue also lists — those are separate follow-ups.

## Red → green

`TestMigrateReconcilesMissingInventoryCacheTable` and `TestMigrateReconcilesMissingPRCommentReviewIDColumn` simulate the drift by burning the version (`INSERT INTO goose_db_version`) and requiring `migrate()` to leave the object present. Both failed before the fix, pass after. Both are killed by removing their repair entry (mutation check).

### DIFF
--- a/backend/internal/storage/sqlite/db.go
+++ b/backend/internal/storage/sqlite/db.go
@@ -938,6 +938,35 @@ BEGIN
 		addDDL: `ALTER TABLE conversation_turns ADD COLUMN promotion_started_at TIMESTAMP`},
 	{version: 89, table: "conversation_turns", column: "promoted_to_turn_id",
 		addDDL: `ALTER TABLE conversation_turns ADD COLUMN promoted_to_turn_id TEXT REFERENCES conversation_turns(id) ON DELETE SET NULL`},
+	// 0106_pr_comment_review_id.sql. Version 106 is recorded as applied on
+	// profiles where the physical column never landed — a burned/skipped
+	// migration — and preparePRCommentReviewIDMigration only repairs the mirror
+	// case (column present, ledger entry missing). The generated PR comment
+	// queries select review_id, so every PR read then 500s.
+	{version: 106, table: "pr_comment", column: "review_id",
+		addDDL: `ALTER TABLE pr_comment ADD COLUMN review_id TEXT NOT NULL DEFAULT ''`},
+}
+
+// tableRepairs is the table-level counterpart of schemaRepairs: migrations
+// whose schema effects are whole tables, recorded as applied on installs where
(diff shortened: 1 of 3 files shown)


axisrow:
## Verification (local, end-to-end)

Tested against a real drifted database — both objects physically missing while `goose_db_version` recorded versions 0104 and 0106 as applied:

1. Rebuilt the daemon with this patch and started it.
2. At boot, `reconcileSchema` recreated the missing objects:
   - `SELECT COUNT(*) FROM sqlite_master WHERE type='table' AND name='agent_inventory_cache'` → **1** (was 0)
   - `PRAGMA table_info(pr_comment)` → `review_id` column **present** (was absent)
3. The startup `WARN initial agent catalog refresh failed: no such table: agent_inventory_cache` is **gone** from the daemon log after the restart.
4. `ao spawn --project <id> --name <name> --kind worker` succeeds **without** `--skip-agent-check` (previously: `Internal server error (INTERNAL_ERROR)`).

CI: 8/8 checks green on this PR.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

axisrow:
Closing as superseded — both field casualties landed upstream by other means: the `review_id` column repair is now an identical `schemaRepairs` entry, and `agent_inventory_cache` was dropped outright (0122) with the catalog moved into the centralized readiness path (#4589), so the 104 table repair no longer applies. Noted the remaining generic gap (`tableCount == 0 → continue` in `prepareBurnedSchemaRepairs`) on #4335. Thanks!

Issue #4325, opened 2026-08-23
bug(httpd): transient/DB errors wrapped untyped collapse to 500 INTERNAL_ERROR


Pulkit7070:
## Summary
Handlers and services return raw `fmt.Errorf(...)` for conditions that are **transient or expected**, and `backend/internal/httpd/envelope.go` maps any non-`apierr` error to `INTERNAL_ERROR` / `500`. As a result, transient DB contention (#3963), client-state conditions, and genuine faults all surface as **identical opaque 500s** — the reliability signal is polluted by non-faults, and the real cause is only visible in the daemon log, never in the API response.

## Where
- `backend/internal/httpd/envelope.go` — the `INTERNAL_ERROR` fallback for untyped errors.
- Example offender: `backend/internal/service/session/workspace_files.go` → `sessionWorkspaceRecord` wraps `GetSession` as `fmt.Errorf("get %s: %w", …)` (untyped → 500).

## Reproduction
1. `rg "envelope.WriteError" backend/internal/httpd/controllers` and follow each to the service call it forwards — the ones returning plain `fmt.Errorf` (not `apierr.*`) become opaque 500s.
2. Drive any such handler while the DB is under contention (#3963): the response is a `500` with code `INTERNAL_ERROR` and no specific `error_code`, indistinguishable from a real server fault.

## Suggested fix
- Map transient/contention DB errors to a typed `503` (retryable) with a stable code, and expected states to typed `4xx`, so the 5xx surface reflects only real server faults.
- Land Sentry error capture (PRs #4180 desktop / #4181 mobile) so the remaining genuine 500s carry a grouped stack and root cause.

## Related
- #3963, #3649, #2699, #3408 (instances of the opacity); #4180 / #4181 (visibility)

PR #4340, opened 2026-08-24
fix(renderer): exponential backoff + jitter for workspace SSE reconnect (#4323)

### LINKED ISSUE #4323: bug(daemon+renderer): SSE streams reconnect with no backoff, amplifying a transient setup 5xx into a request storm
## Summary
The renderer opens Server-Sent-Events streams — `GET /api/v1/sessions/{id}/workspace/events` and `/api/v1/notifications/stream` — with `EventSource`, which reconnects **immediately on any failed connection, with no backoff**. If the daemon returns a transient 5xx while *setting up* the stream, the client reconnects in a tight loop, turning one brief failure into a sustained burst of failed requests and high CPU on both sides.

## Where it happens
- Handler: `backend/internal/httpd/controllers/sessions.go` → `streamWorkspaceChanges` (and the notifications stream controller). These can only return 5xx during setup, before `w.WriteHeader(200)`.
- Setup failure points: `WorkspaceWatchPaths` → `sessionWorkspaceRecord` (its `GetSession` branch wraps the error untyped as `fmt.Errorf("get %s: %w", …)` → 500), and `workspacewatch.Watch` / `watchRoot`.
- Renderer: the `EventSource` consumer for these streams reconnects with no backoff.

## Reproduction
The daemon port is in `~/.ao/running.json`.
1. Healthy stream (baseline): `curl -N http://127.0.0.1:<port>/api/v1/sessions/<id>/workspace/events` → `200`, stream stays open. ✅ verified
2. Missing workspace: same call for a session whose workspace dir is gone → `404 SESSION_WORKSPACE_NOT_FOUND` (already typed on current `main`). ✅ verified — so the common missing-path case is *not* the culprit.
3. Trigger the amplification: put the daemon under DB-lock contention (see #3963) so `GetSession` intermittently errors → stream setup returns `500` → watch the renderer immediately reconnect and re-fail in a loop (visible in the access log and as CPU spikes). A single stuck stream produces a large, self-sustaining burst.

## Suggested fix
- Renderer: exponential backoff + jitter on SSE reconnect; stop reconnecting when the session is terminated and on any `4xx`.
- Daemon (optional): send `Retry-After` on transient setup failures so clients pace reconnects.

## Related
- #3963 (DB lock contention → the transient 5xx this amplifies)
- #1321 (RFC: remove polling → reactive sync)

Pulkit7070:
Fixes #4323.

## Problem
`workspace-file-events.ts` retried the SSE stream on a **fixed 5s interval** with no backoff, no ceiling reset, and no cap. A session whose stream keeps failing (e.g. the daemon briefly returns 5xx under DB contention, #3963) reconnected every 5s indefinitely, and many streams reconnected in lockstep — turning a brief hiccup into a sustained reconnect load on the daemon.

## Change (renderer-only, complementary to #4230)
- New dependency-free `sse-backoff.ts` → `computeBackoffDelayMs(failures)`: exponential growth from **5s** to a **60s** ceiling, jitter in `[0.5, 1.0]` of the step (spreads concurrent reconnects), overflow-clamped.
- The stream tracks consecutive failures, backs off per retry, and **resets to a fast first retry on a successful open**.

## Verification
- Ran the actual helper **100×** with invariant checks: monotonic growth to the cap, never exceeds 60s, never below the jitter floor, high-count clamp (no overflow). Sample progression: `4157, 6698, 14280, 39890, 43711, 54303` ms (vs the old flat 5000).
- `sse-backoff.test.ts` (vitest) covers growth, ceiling, jitter band, clamp, and NaN/negative handling.

## Scope / limitation
Native `EventSource` does not expose the HTTP status on error, so "stop reconnecting on 4xx" cannot be done from the client alone — backoff is the client-side mitigation. Daemon-side error classification (transient → 503) is handled separately in #4325. This composes with #4230 (daemon-side SSE replay cap) rather than overlapping it.

### DIFF
--- a/frontend/src/renderer/lib/sse-backoff.test.ts
+++ b/frontend/src/renderer/lib/sse-backoff.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from "vitest";
+import { SSE_RETRY_BASE_MS, SSE_RETRY_MAX_MS, computeBackoffDelayMs } from "./sse-backoff";
+
+describe("computeBackoffDelayMs", () => {
+	const lo = (f: number) => computeBackoffDelayMs(f, SSE_RETRY_BASE_MS, SSE_RETRY_MAX_MS, () => 0);
+	const hi = (f: number) => computeBackoffDelayMs(f, SSE_RETRY_BASE_MS, SSE_RETRY_MAX_MS, () => 0.999999);
+
+	it("grows exponentially from base until it hits the ceiling", () => {
+		expect(lo(0)).toBe(Math.round(SSE_RETRY_BASE_MS * 0.5)); // 2500
+		expect(lo(1)).toBe(Math.round(10_000 * 0.5)); // 5000
+		expect(lo(2)).toBe(Math.round(20_000 * 0.5)); // 10000
+		const seq = [0, 1, 2, 3, 4, 5, 10, 20].map(lo);
+		for (let i = 1; i < seq.length; i++) expect(seq[i]).toBeGreaterThanOrEqual(seq[i - 1]);
+	});
+
+	it("never exceeds the max ceiling and never drops below the jitter floor", () => {
+		for (const f of [0, 1, 2, 3, 4, 5, 10, 20, 30, 100]) {
+			const d = computeBackoffDelayMs(f);
+			expect(d).toBeLessThanOrEqual(SSE_RETRY_MAX_MS);
+			expect(d).toBeGreaterThanOrEqual(Math.round(SSE_RETRY_BASE_MS * 0.5));
+		}
+	});
+
+	it("applies jitter within [0.5, 1.0] of the step", () => {
+		expect(lo(2)).toBe(10_000); // 0.5 * 20000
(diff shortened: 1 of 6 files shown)


Rishet11:
@Pulkit7070 
#4344 rewrote the SSE parse and workspace file event code an hour after this PR opened, and issue #4323 was closed completed by it. 
 Closing as superseded.

Pulkit7070:
Leaving this closed, but the supersession did not hold, so the work has moved to #4826 rather than being reopened here.

#4344 did merge, but its implementation is no longer in the tree: `sse-parse.ts` is gone and #4388 rewrote `workspace-file-events.ts` back to native `EventSource`. What is on `main` today is a flat 5s +/- 1s reconnect with no growth and no ceiling, so the defect in #4323 is still live. #4323 is reopened with the details.

Not reopening this branch: it is 151 commits behind, both target files conflict, and its first commit is #4344 itself, so its diff would re-revert #4388 and double-count something already merged. #4826 is rebuilt on current `main` and also covers `event-transport.ts` and `notifications.ts`, which carry the same flat retry and were outside this PR's scope.

PR #4342, opened 2026-08-24
fix(desktop): defer agent auth probes until selection

### LINKED ISSUE #4341: Desktop startup waits on every agent authentication probe
## Bug
Opening the desktop app can spend 2–10 seconds on the startup requirements screen while AO probes every installed agent CLI and its authentication status, even though startup only needs inexpensive executable-presence checks.

**Source:** [PR #4330](https://github.com/Untrivial-ai/agent-orchestrator/pull/4330) / teammate report by @ronishrohan
**Introduced by:** [PR #3825](https://github.com/Untrivial-ai/agent-orchestrator/pull/3825)
**Analyzed against:** `f20411d14`
**Priority:** P2 — significant startup degradation; the app eventually loads.
**Confidence:** High

## Reproduction
1. Install several supported agent CLIs, including one whose auth-status command is slow or times out.
2. Start the desktop app.
3. Observe that the startup requirements gate remains pending until the slowest parallel agent/auth probe completes.

## Root Cause
`backend/internal/service/systemcheck/systemcheck.go` implements the startup harness requirement through `agent.Service.RefreshFresh`. That refresh fans out across every registered harness, resolves its binary, and invokes optional auth probes with timeouts up to 10 seconds. The renderer keeps `SessionsBoard` behind the requirements response, so this full catalog refresh is on the first-render critical path.

The daemon and renderer also each trigger eager full catalog refreshes during startup, duplicating work that agent-selection surfaces already know how to request on demand.

## Fix
- Make the startup harness requirement use a bounded executable-presence-only probe, without starting agent CLIs or checking auth.
- Stop eagerly refreshing the full agent inventory during daemon/shell startup.
- Keep full inventory/auth refreshes on the existing agent-selection surfaces (`refreshAgentsIfStale`).

## Impact
- Restores fast first render regardless of slow agent auth commands.
- Preserves install gating for git, tmux, and the presence of at least one supported agent executable.
- Auth remains advisory and refreshes when the user opens a surface where agent choice matters.

## Related
- [PR #3825](https://github.com/Untrivial-ai/agent-orchestrator/pull/3825) — introduced the startup requirements gate.
- [PR #4330](https://github.com/Untrivial-ai/agent-orchestrator/pull/4330) — teammate proposal that includes a binary-only startup probe plus broader startup layout/reconciliation changes.

illegalcall:
Fixes #4341

## Summary
- replace the startup gate's full agent inventory/auth refresh with a bounded executable-presence-only probe
- remove the duplicate full catalog refreshes from daemon boot and shell startup
- keep full binary/auth discovery on the existing agent-selection surfaces, which refresh the cached inventory when opened
- preserve Muse's full version-signature validation for actual launches while allowing startup to use a path-only presence check

This isolates the auth-probe regression from the broader startup reconciliation and layout changes proposed in #4330.

## Test
- `cd backend && go test -race ./internal/service/agent ./internal/service/systemcheck`
- `cd backend && go test ./internal/httpd/controllers ./internal/daemon`
- `cd frontend && npx vitest run src/renderer/test/shell-new-session-shortcut.test.tsx src/renderer/components/DaemonStartupLoader.test.tsx`
- `cd frontend && npx vitest run src/renderer/components/CreateProjectAgentSheet.test.tsx src/renderer/components/TaskComposer.test.tsx src/renderer/components/ProjectSettingsForm.test.tsx`

## Full-suite note
`cd backend && go test ./...` passed every package except the unrelated timing-sensitive `internal/adapters/agent/fake` lifecycle speed assertion while the build, vet, tests, and race checks were running concurrently. Re-running that package alone passed.

### DIFF
--- a/backend/internal/adapters/agent/muse/install.go
+++ b/backend/internal/adapters/agent/muse/install.go
@@ -1,8 +1,18 @@
 package muse
 
-import "context"
+import (
+	"context"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/binaryutil"
+)
 
 // ResolveBinary resolves the executable path for the plugin.
 func (p *Plugin) ResolveBinary(ctx context.Context) (string, error) {
 	return p.museBinary(ctx)
+}
+
+// ResolveBinaryPresence resolves a local Muse executable without running its
+// version-signature check. Full resolution still validates Muse before launch.
+func (p *Plugin) ResolveBinaryPresence(ctx context.Context) (string, error) {
+	return binaryutil.ResolveBinary(ctx, museBinarySpec)
 }
--- a/backend/internal/daemon/daemon.go
+++ b/backend/internal/daemon/daemon.go
@@ -270,12 +270,10 @@ func Run() error {
 	}
 	lcStack.trackerDone = startTrackerIntake(ctx, store, sessionSvc, log)
 
+	// Keep full agent binary/auth probing off daemon startup. The requirements
+	// gate performs a path-only check, and agent-selection surfaces request a
+	// fresh inventory when the user opens them.
 	agentSvc := agentsvc.NewWithDeps(agentsvc.Deps{Cache: store, InventoryCache: store, Discoverer: modelcatalog.Discoverer{}, Projects: store, Sessions: store})
-	go func() {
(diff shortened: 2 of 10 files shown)


illegalcall:
Closing this duplicate follow-up. Continuing the diagnosis, validation, and any required fixes directly on #4330.

PR #4346, opened 2026-08-24
feat(observability): daemon Sentry — capture 5xx and panics with Go stacks

Pulkit7070:
Third and final Sentry surface. The renderer (#4180) and mobile (#4181) slices capture what the *clients* see; this captures what the *daemon* actually threw. It is where the daemon's opaque `INTERNAL_ERROR` 500s finally get a real Go stack instead of a bare count, grouped by the same fingerprint the telemetry path already uses so an issue lines up across both.

## What this adds
- **`internal/observe/sentryobs`** — a small daemon Sentry integration. DSN-gated via `AO_SENTRY_DSN`; a blank DSN leaves it a permanent no-op, so this merges now and goes live the moment the DSN is set. Deny-by-default on privacy: PII off, `BeforeSend` clears `ServerName` and the request context and scrubs local paths (`/Users/...`, `/home/...`, `C:\...`) out of the message and every stack frame; no breadcrumbs; tracing off.
- **Capture at the seams where the real error and stack are live**, reusing the existing telemetry emit points rather than adding a second path:
  - httpd 5xx access-log branch — captures the handler's `capturedErr` with the classified `error_kind`/`error_code`.
  - panic recover middleware — captures the recovered value with its `debug.Stack()`.
- **Grouping** — the same `telemetrymeta.Fingerprint(...)` the PostHog `ao.http.5xx` / `ao.daemon.panic` events use, tagged `platform=daemon` plus `operation`, `path` (route pattern), `status`, `error_code`, and `request_id` for cross-correlation.
- **Transient 503** (`SERVICE_UNAVAILABLE`, see #4325) is deliberately excluded via `ShouldCaptureStatus` — retryable contention is not a fault and would drown the signal.

## Why no behavior change until the DSN is set
`Init` returns early on a blank DSN and every capture is a guarded no-op, so the build, tests, and runtime are unchanged today. Existing `ao.http.5xx` / `ao.daemon.panic` telemetry is untouched.

## Ship steps (when the DSN arrives)
1. Set `AO_SENTRY_DSN` in the daemon environment.
2. Release/environment are already wired from the app version + version channel.

## Verification
- Verified over a socket against a mock Sentry ingest: an event is transmitted with the local path scrubbed (no `/Users` leak), carrying the fingerprint tag and `platform=daemon`; a 503 is not captured.
- Unit tests (`scrub`, `ShouldCaptureStatus` incl. the 503 skip, no-op-when-disabled, full event scrub) pass at `-count=50`.
- `go build ./...`, `golangci-lint` clean, httpd/config/sentryobs packages green.

Completes the three-surface Sentry coverage with #4180 (desktop) and #4181 (mobile). Related error work: #4323, #4325, #4334.


illegalcall (approved):

illegalcall:
## Approved — merge the foundation, harden before broad enablement

Approved at `781a41245d326dad8af513f06900a20f9d2038c1`.

This is a substantial improvement over the current daemon observability baseline. The PR puts capture at the right central seams, preserves request-ID correlation with desktop/mobile, uses stable grouping, scrubs sensitive daemon context, remains inert without a DSN, and has green CI across Go, API drift, leak scanning, and CLI E2E platforms.

I do not think the remaining concerns need to block merging this DSN-gated foundation. I filed [issue #4352](https://github.com/Untrivial-ai/agent-orchestrator/issues/4352) as the focused follow-up, with exact code paths, reproduction scenarios, proposed implementation boundaries, and acceptance tests.

The follow-up covers:

- isolating Sentry scope per request so concurrent failures cannot exchange request IDs, tags, or fingerprints;
- ensuring one recovered panic produces exactly one event;
- suppressing only typed `SERVICE_UNAVAILABLE` backpressure rather than every HTTP 503;
- adding enabled-path tests with a recording transport, including race coverage.

I can take [issue #4352](https://github.com/Untrivial-ai/agent-orchestrator/issues/4352) as the next change while this PR is merged. The operational boundary I recommend is: merge now, but land the hardening issue before broad production DSN enablement.

PR #4347, opened 2026-08-24
fix(daemon): report degraded readiness and degrade spawn preflight on catalog failure

axisrow:
## What

Follow-up to #4336, addressing the remaining items of #4335. The agent catalog is the preflight dependency of `ao spawn`, but a refresh failure was swallowed into a fire-and-forget WARN while `/readyz` kept reporting ready — so a daemon that could not spawn still passed readiness, and the CLI surfaced the failure as an opaque `INTERNAL_ERROR`.

1. **`/readyz` reflects real readiness** — new `Readiness` state in `httpd` (starts ready, can be marked degraded with a reason). `/readyz` returns `503` + `status: "degraded"` + `reason` when degraded, instead of unconditionally `200`. Nil readiness (the test default and pre-existing callers) keeps the old behavior.
2. **Daemon marks itself degraded** — when the initial agent-catalog refresh fails, the daemon records the cause on `Readiness`, so the failure is visible to the supervisor and to clients gating on the probe instead of dying in a WARN nothing reads.
3. **Spawn preflight degrades instead of blocking** — when the catalog refresh fails, `ao spawn` warns and continues, letting spawn validate runtime readiness (mirroring the existing `agentProbeUnavailable` degradation). The service-side contract ("published = cached") is intentionally encoded by `TestRefreshPublishesOnlyAfterInventoryPersistenceSucceeds`, so the fix lives in the CLI path, not the service.

## Why

Part of #4335: a failure in a core dependency must not be both invisible (`/readyz` green) and blocking (`INTERNAL_ERROR`) at the same time.

## Red → green

- `TestReadyzReflectsDegraded` — `/readyz` returned `200` for a degraded daemon; now `503` + reason.
- `TestSpawnCatalogRefreshFailureDegradesToWarning` — spawn failed on a `500` refresh; now warns and spawns.
- Both tests are killed by reverting their fix (mutation check).

### DIFF
--- a/backend/internal/cli/spawn.go
+++ b/backend/internal/cli/spawn.go
@@ -375,6 +375,11 @@ func resolveSpawnHarness(explicit, kind string, project projectDetails) (string,
 func (c *commandContext) preflightSpawnAgentAuth(ctx context.Context, cmd *cobra.Command, agentID string) error {
 	inv, err := c.fetchAgentInventory(ctx, true)
 	if err != nil {
+		// The catalog is advisory for spawn: a refresh failure (schema drift,
+		// transient DB error) must not turn into an opaque INTERNAL_ERROR that
+		// blocks the core workflow. Warn and let spawn validate runtime
+		// readiness, mirroring the agentProbeUnavailable degradation below.
+		_, err = fmt.Fprintf(cmd.ErrOrStderr(), "warning: agent catalog refresh failed (%v); skipping preflight and letting spawn validate runtime readiness\n", err)
 		return err
 	}
 	state := agentCatalogStateFor(inv, agentID)
--- a/backend/internal/cli/spawn_test.go
+++ b/backend/internal/cli/spawn_test.go
@@ -807,6 +807,48 @@ func TestSpawnFreshProbeServerErrorBlocks(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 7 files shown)


axisrow:
## Verification

- **Unit:** `TestReadyzReflectsDegraded` (a degraded daemon returns 503 + `status: "degraded"` + reason) and `TestSpawnCatalogRefreshFailureDegradesToWarning` (a 500 catalog refresh warns and lets spawn continue). Both red before the fix, green after, and each is killed by reverting its fix (mutation check).
- **No-regression on the running daemon:** `/healthz` unchanged (200 ok); `/readyz` still returns 200 `ready` with a healthy catalog.
- **CI:** 8/8 checks green on this PR.

The degraded path is exercised through unit tests rather than a live daemon on purpose: reproducing it end-to-end requires breaking the catalog (e.g. dropping the inventory table), which the #4336 reconciliation now repairs at boot instead — so a live degraded daemon is not reachable through that path anymore.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

axisrow:
Closing as superseded — #4589 (centralized harness readiness) covers this with a better architecture: check failures degrade to per-agent snapshot states with bounded retries instead of a global /readyz flag. Thanks!

PR #4351, opened 2026-08-24
fix(workspace): serialize worktree adds per repo and surface WORKSPACE_CREATE_FAILED

### LINKED ISSUE #4350: bug(daemon): concurrent ao spawn races on git worktree add (.git/config.lock) and surfaces opaque 500 INTERNAL_ERROR
## Bug

Concurrent `ao spawn` calls race on `git worktree add -b`: N parallel spawns each run a `git fetch` + `git worktree add` against the same `.git` with **no serialization**, so the worktree-adds collide on `.git/config.lock` and the losers fail with exit 255. The failure then surfaces to the CLI as an opaque `500 INTERNAL_ERROR` because workspace-create errors are not in the service layer's typed sentinel set.

**Source:** Discord `#bug-triage` thread (2026-08-24) | **Reported by:** @piyush
**Analyzed against:** `346b53336` (origin/main, 2026-08-24)
**Confidence:** High — exact mechanism traced in code and confirmed by reporter's daemon log
**Priority:** P1 — concurrent/batch spawning is a documented workflow; 7 of 15 spawns failed in the triggering run

## Reproduction

1. Fire ~15 `ao spawn` CLI calls in quick succession against one project (reporter did 15; 3 failed with the race below, ~4 more failed with the same underlying error surfaced as `INTERNAL_ERROR`).
2. Roughly half fail. CLI shows only: `Internal server error (INTERNAL_ERROR) [request ...]`.
3. `~/.ao/daemon.log` carries the real cause on the `POST /api/v1/sessions status=500` line:

```
spawn agent-orchestrator-42: workspace: gitworktree: worktree add branch
"ao/agent-orchestrator-42/root" from "refs/remotes/origin/main":
git -C /home/pyasma/Projects/agent-orchestrator worktree add -b ao/agent-orchestrator-42/root
  /home/pyasma/.ao/data/worktrees/agent-orchestrator/agent-orchestrator-42 refs/remotes/origin/main:
exit status 255: Preparing worktree (new branch 'ao/agent-orchestrator-42/root')
error: could not lock config file .git/config: File exists
error: unable to write upstream branch configuration
```

Reporter's log shows the identical signature across **11 sessions** (-33, -37, -52, -50, -59, -64, -66, -68, -74, -75, -76) in **3 separate bursts**, each failed request ~2.9s — a consistent single root cause, not a fluke. Reporter env: Linux (omarchy), agent-orchestrator project; behavior is OS-independent (git lock-file semantics).

## Root Cause

Two independent defects, both verified on current `main` (`346b53336`):

### 1. Per-spawn git operations are completely unserialized

The spawn path in `backend/internal/session_manager/manager.go` (`Spawn`, line ~700-830) runs per request, and each one executes:

- `refreshDefaultBranchesBestEffort` (line 976) — `git fetch` on the project repo (network I/O, per spawn)
- `createSessionWorkspace` (line 1054) → `gitworktree.Workspace.Create` → `addWorktree` — `git worktree add -b <branch> <path> <base>` (`backend/internal/adapters/workspace/gitworktree/workspace.go:227`)

`git worktree add -b` writes `branch.<name>.remote` / `branch.<name>.merge` into `.git/config` at the end of checkout. Concurrent invocations fight over `.git/config.lock`; every loser gets `could not lock config file .git/config: File exists` → exit 255.

What is **not** the problem (verified):
- Session ID allocation is safe: `Store.CreateSession` (`backend/internal/storage/sqlite/store/session_store.go:20`) runs `NextSessionNum` + insert under `writeMu`.
- `Manager.agentOpMu` (`manager.go:344`) only guards terminal input/pane writes — not spawn.
- The gitworktree adapter has zero locking (no `sync.` usage in `workspace.go`).
- `Service.Spawn` (`backend/internal/service/session/service.go:225`) only locks orchestrator-kind spawns (`lockOrchestratorProject`); worker spawns run fully concurrent.

### 2. Workspace-create failures collapse to an opaque 500

`service.spawn` (line 252-256) passes the manager error to `toAPIError` (line 894), a closed sentinel switch (`ErrNotFound`, `ErrNotRestorable`, `ErrTerminated`, ...). A gitworktree failure matches none of them, so the raw error passes through; `envelope.WriteError` finds no `*apierr.Error` and collapses to `500 INTERNAL_ERROR`. The real cause survives only in the daemon log's 5xx request line — the same 500-collapse class as #2598/#2608/#3649.

## Fix

1. **Serialize git mutations per repo** — an in-process per-repo-path mutex (or `singleflight`) in the gitworktree adapter around `worktree add` (and around the spawn-time `fetch`), since all spawns funnel through the one daemon. This removes the lock-file race entirely without changing any single-spawn behavior.
2. **Type the error** — wrap workspace-create failures in a typed `apierr` (retryable 409/503, e.g. `WORKSPACE_CREATE_FAILED` / `GIT_LOCK_CONTENTION`) so the CLI shows the real cause instead of `INTERNAL_ERROR`, per the fix direction in the 500-collapse class.
3. **Optional hardening** — one bounded retry of `worktree add` on `could not lock config file` absorbs contention from non-AO git processes in the same clone (user's editor, etc.).

## Impact

- Batch/concurrent spawning (documented workflow; orchestrators do this via `ao batch-spawn` guidance) fails ~50% per collision burst, with no actionable CLI error.
- Failed spawns roll back only the DB seed row (`rollbackSpawnSeedRow`, manager.go:787 — assumes "no worktree exists"); any partially-created worktree/branch leaks, feeding the orphaned-worktree/husk class in #3649 / #3220.

## Related

- #3649 — intermittent spawn INTERNAL_ERROR + orphaned worktrees (same symptom class; this issue supplies the concurrent-spawn mechanism and the exact git failure)
- #2699 — spawn 500 when runtime is slow (different trigger, same opaque-error collapse)
- #3220 — worktree husk under load (downstream effect of the same unserialized path)
- #1035 / #1063 — historical spawn races (TS era, different mechanism)

## Reporter's environment

Linux (omarchy/Arch), daemon from an AppImage build — but both defects verified against current `main` at `346b53336`; build vintage is not a factor.

i-trytoohard:
Fixes #4350

## Summary

Concurrent `ao spawn` calls race on `git worktree add -b` against the same clone: each spawn runs an unserialized `git fetch` + `git worktree add -b`, the adds collide on `.git/config.lock`, and the losers fail with exit 255 (`could not lock config file .git/config: File exists`). The service layer then collapses the git error into an opaque `500 INTERNAL_ERROR` — the CLI shows no actionable cause.

Two changes:

1. **Per-repo serialization in the gitworktree adapter.** A keyed mutex (`Workspace.lockRepo`) serializes `addWorktree`, `createWorkspaceProjectRepo`, and `FetchDefaultBranch` per repo path. All spawns funnel through the one daemon process, so an in-process lock fully removes the race; different repos still spawn concurrently. Lock map access is nil-safe because tests construct `Workspace` as a zero value.
2. **Typed error instead of 500.** New `ports.ErrWorkspaceCreateFailed` sentinel wraps every failed worktree-add return; `toAPIError` maps it to `409 WORKSPACE_CREATE_FAILED` carrying the underlying git error text, so the CLI finally shows the real cause.

Deadlock safety: the three lock sites never call each other (`addWorktree` → `addNewBranchWorktree` takes no lock; `createWorkspaceProjectRepo` is a sibling entry point; `FetchDefaultBranch` is only called before `createSessionWorkspace` in the spawn flow, never from within a lock). Ref resolution happens under the lock but is bounded local reads (`resolveWorktreeRefsWithBudget`, 5s budget).

## Test

cd backend && go build ./... && go test ./...

- New `TestCreateSerializesWorktreeAddPerRepo`: 15 parallel `Create` calls on one repo; the fake runner fails the test if two `worktree add` invocations overlap (max in-flight must be 1). Reproduces the #4350 shape (15 concurrent spawns) in miniature.
- New `toAPIError` table case: `ErrWorkspaceCreateFailed` → `409 WORKSPACE_CREATE_FAILED`, message retains the git error.
- Existing gitworktree + session service suites pass unchanged.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/workspace.go
+++ b/backend/internal/adapters/workspace/gitworktree/workspace.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
@@ -81,6 +82,35 @@ type Workspace struct {
 	managedRoot string
 	repos       RepoResolver
 	run         commandRunner
+
+	// repoLocksMu guards repoLocks. repoLocks serializes git operations that
+	// mutate shared repo state (.git/config, worktree registrations, packed
+	// refs) per repository path: concurrent spawns each run `git worktree add
+	// -b`, which writes branch.<name> config at the end of checkout, and the
+	// losers of the .git/config.lock race fail with exit 255 (#4350). Tests
+	// construct Workspace as a zero value, so access is nil-map safe.
+	repoLocksMu sync.Mutex
+	repoLocks   map[string]*sync.Mutex
+}
+
+// lockRepo returns a release function for the per-repo git mutation lock. Every
+// caller must be safe to serialize against any other caller on the same repo
+// path; cross-repo spawns still run concurrently. Paths are cleaned so
+// equivalent spellings of one repo share one lock.
+func (w *Workspace) lockRepo(repo string) func() {
+	key := filepath.Clean(repo)
+	w.repoLocksMu.Lock()
+	if w.repoLocks == nil {
+		w.repoLocks = map[string]*sync.Mutex{}
+	}
+	mu, ok := w.repoLocks[key]
+	if !ok {
+		mu = &sync.Mutex{}
+		w.repoLocks[key] = mu
+	}
+	w.repoLocksMu.Unlock()
+	mu.Lock()
+	return mu.Unlock
 }
 
 type commandRunner func(ctx context.Context, binary string, args ...string) ([]byte, error)
@@ -178,6 +208,12 @@ func (w *Workspace) FetchDefaultBranch(ctx context.Context, repoPath string, tar
 	if err := w.validateBranch(ctx, repo, target.Branch); err != nil {
 		return err
 	}
+	// Serialize with worktree adds on the same repo: a fetch rewrites
+	// packed-refs (packed-refs.lock) and can also take .git/config.lock, so
+	// concurrent spawn-time fetches race worktree adds on the same clone
+	// (#4350).
+	unlock := w.lockRepo(repo)
+	defer unlock()
 	if _, err := w.run(ctx, w.binary, fetchBranchArgs(repo, target.Remote, target.Branch)...); err != nil {
 		return fmt.Errorf("gitworktree: fetch %s %s: %w", target.Remote, target.Branch, err)
 	}
@@ -969,6 +1005,12 @@ func registeredWorktreeDirMissing(rec worktreeRecord) (bool, error) {
 }
 
 func (w *Workspace) addWorktree(ctx context.Context, repo, path, branch, baseBranch, baseRef string, resolveExistingBase bool) (string, error) {
+	// Serialize worktree mutations per repo: `git worktree add -b` writes
+	// branch.<name> config under .git/config.lock, and concurrent adds on one
+	// repo race on that lock (#4350). The ref-resolution reads inside are safe
+	// to hold the lock across: they are local reads with a bounded budget.
+	unlock := w.lockRepo(repo)
+	defer unlock()
 	// Refuse early if the branch is already checked out in another worktree:
 	// `git worktree add` will fail, but its stderr leaks through as an opaque
 	// 500. A typed sentinel lets the HTTP layer surface a 409.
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
@i-trytoohard Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

codebanditssss:
Thanks for the contribution. Closing this in favor of #5654, which addresses the same #4350 worktree/config-lock race and typed `WORKSPACE_CREATE_FAILED` response against the current workspace adapter. The retained implementation also canonicalizes repository lock keys with `physicalAbs` and follows the adapter’s existing process-wide teardown-lock pattern, avoiding per-instance locks that would not coordinate separate adapter instances.

PR #4355, opened 2026-08-24
fix(chat): recover Codex chat after auth expiry

### LINKED ISSUE #4296: bug(chat): Codex Chat UI dies with 'remote compact task' auth error after switching Codex accounts — stale OAuth creds in long-lived app-server, no auth-failure detection
## Bug

Switching between two Codex accounts works in a worker's Terminal UI, but the Chat UI for the same session then fails with:

```
Error running remote compact task: Your access token could not be refreshed because you have since logged out or signed in to another account. Please sign in again.
```

The error appears on the chat timeline (as a failed turn/error item) and the session's chat remains broken until it is restarted — AO never detects that this is an **auth failure** and offers no recovery path.

**Source:** [Discord #report-bugs thread](https://discord.com/channels/1476302178913357958/1540913392313962516/1540913392313962516)
**Reported by:** @notephemeral. (corroborated by @kiki — "errors come", exact message not retained)
**Environment:** Fedora 43, AO v0.12.3 (published 2026-08-11)
**Analyzed against:** `ed52364e3` (2026-08-21)
**Confidence:** Medium — AO-side mechanism traced in code; the exact provider-side token invalidation timing is inferred from standard OAuth behavior (refresh token revoked on new login), not observed live.

## Reproduction

1. Log into Codex account A; spawn a Codex worker session and switch it to Chat UI (or start it in Chat UI).
2. From a terminal, log into Codex account B (`codex login` with the other account).
3. Return to the worker's Terminal UI — works fine (new process reads fresh auth state from `~/.codex/auth.json`).
4. Send a message in the Chat UI of the same session — after the next provider-side context compaction ("remote compact task"), the error above appears.
5. Chat stays broken; no prompt to re-authenticate or restart.

(Step ordering per reporter; kiki confirms the error reproduces but did not retain the exact message.)

## Root Cause

The two surfaces use **different processes with different credential lifetimes**:

**Terminal UI** — the Codex TUI runs inside the tmux pane. When the user switches accounts from the terminal, the process they subsequently interact with reads the freshly-written `~/.codex/auth.json`, so the new account works.

**Chat UI** — chat runs through a daemon-spawned, long-lived `codex app-server` process:

- `backend/internal/adapters/chatdriver/codexappserver/driver.go:392` (`spawnAppServer`) launches `codex app-server` with the daemon's env + session overlay (`envSlice`, driver.go:456 — deliberately merged over the daemon environment, so it shares the same `~/.codex` as the terminal).
- The app-server captures the account's OAuth tokens at startup and holds them in memory for its lifetime, refreshing as needed. It is spawned once per conversation and kept alive indefinitely.
- When the user logs into account B from the terminal, OpenAI invalidates account A's refresh token server-side ("…because you have since logged out or signed in to another account"). The still-running app-server keeps the dead credentials.
- The next time the app-server must call the ChatGPT backend — its background auto-compaction ("remote compact task" is Codex's own wording; AO only has `thread/compact/start` at `conversation.go:409` for AO-initiated compaction, so this is the provider's internal one) — token refresh fails and the provider emits the error above.

**What AO does with it — nothing actionable:**

- The error arrives as a provider error item and is normalized to a generic failed activity: `backend/internal/adapters/chatdriver/codexappserver/normalize.go:1059-1061` (`itemError` → `ActivityStatusFailed`). There is no sentinel/typed mapping for auth failures post-launch.
- Auth is only checked **before** a conversation exists: `driver.go:141-143` (`Probe` → `plugin.AuthStatus` → `ports.ErrChatAuthRequired`). After launch there is no re-check, no app-server restart, and no user-facing "sign in again" affordance.
- The raw provider text lands on the timeline, leaving the user to guess that killing the session (or switching back to account A) is the fix.

## Suggested Fix

1. **Detect the auth sentinel post-launch.** In the codexappserver driver, classify provider error items matching the "access token could not be refreshed / signed in to another account / Please sign in again" family as a typed auth error (new `ports.ErrChatAuthExpired` or reuse `ports.ErrChatAuthRequired`) instead of a generic failed activity — mirroring how the settle path maps provider states to typed sentinels.
2. **Surface an actionable state in Chat UI.** Map that typed error to a stable API code (e.g. `CHAT_AUTH_REQUIRED`) with remediation copy: re-authenticate or restart the session's chat controller.
3. **Consider recovering, not just reporting.** On a detected auth failure, tear down and respawn the app-server so it re-reads `~/.codex/auth.json` (which the account switch already updated), then offer resume via the stored `ProviderConversationID`. That would make account switching transparent in Chat UI, matching Terminal UI behavior.
4. **Lower-confidence alternative:** schedule the respawn on the *next* conversation open rather than live, if live-respawn risks losing in-flight state.

Holding off on a fix PR pending maintainer sign-off — this touches the auth/error-classification path (per repo convention for sensitive paths), and the provider-side sentinel strings need to be captured from a live repro to avoid over-matching.

## Impact

- **Who's affected:** any user who switches Codex accounts (or logs out/in of ChatGPT elsewhere) while a Codex Chat UI session is live. Multi-account Codex users hit this every switch.
- **Severity:** the session's Chat UI is unusable until restart; Terminal UI of the same session keeps working, which makes the failure confusing ("works there, breaks here").
- **Workaround:** kill and respawn the session (or switch back to the account the app-server captured). Undiscoverable — the error text never suggests it.

## Related

- Codex chat driver: `backend/internal/adapters/chatdriver/codexappserver/` (driver/normalize/conversation)
- Chat controller compaction refusal codes: `backend/internal/httpd/controllers/conversations.go:664-677` (`CHAT_COMPACTION_UNSUPPORTED`, `CHAT_COMPACTION_BUSY`) — precedent for typed chat error codes
- Pre-launch auth gate: `driver.go:141-143`

### LINKED ISSUE #5114: [Bug]: Chat UI doesn't detect codex account switch but terminal does
### Summary

Chat UI doesn't detect codex account switch but terminal does. Screenshot attached.
Image

If I switch to terminal UI, the account switch is detected and im able to work with the agent. 

### Steps to reproduce

1. Start a chat UI session
2. Switch your codex account
3. Go back to the chat UI session it will give an error please sign back in

### Expected behavior

1. Start a chat UI session
2. Switch your codex account
3. Go back to the chat UI session it should work seamlessly

### Actual behavior

1. Start a chat UI session
2. Switch your codex account
3. Go back to the chat UI session it will give an error please sign back in

### AO version

v0.12.13-nightly.202609080619

### Environment

macOS

### Logs / extra context

```shell

```

### Before submitting

mknayam:
Fixes #4296
Fixes #5114

## Summary

- Detect terminal Codex authentication failures from structured `codexErrorInfo: "unauthorized"` method errors and authoritative turn completions.
- Ignore temporary `willRetry` notifications, fence provider commands immediately, then terminate the exact stale persistent app-server generation.
- Preserve the failed/ambiguous turn without automatically retrying it, because it may already have changed files or run commands.
- Keep queued turns that never reached Codex and drain them exactly once after verified reconnection.
- Add **Reconnect this chat** with a default-off **Also restart other running AO Codex sessions** option.
- Reuse the durable account-switch coordinator for current-chat and optional all-managed-session recovery, including daemon restart recovery and same-account credential refreshes.
- Add `POST /api/v1/sessions/{sessionId}/conversation/recover-auth` and regenerate OpenAPI/frontend types.

## Recovery flow

1. Codex reports a terminal structured `unauthorized` failure.
2. AO raises an in-memory delivery fence before projection, durably records authentication recovery, settles ambiguous running work, retains undelivered queued turns, closes pending requests, and stops the exact stale provider host.
3. The UI warns that the failed turn was not retried and may have partial side effects.
4. The user chooses **Reconnect this chat**, optionally including other running AO-managed Codex Chat, TUI, and reviewer sessions.
5. AO reconciles and verifies the current credentials. It asks for `codex login` only when verification explicitly reports unauthorized.
6. The durable coordinator resumes the exact native conversation with a reserved controller generation, commits ownership, clears the reauthentication marker, and drains retained queued work once in FIFO order.

External Codex processes in terminals, IDEs, or other unmanaged applications are not touched.

## Dependencies

This PR is temporarily stacked on #5198, which is stacked on #5170. After those dependencies merge, this PR can be retargeted to `main` without changing the implementation.

## Verification

Generated SQL, OpenAPI, and frontend API artifacts are committed. Go formatting and `git diff --check` are clean.

The final stacked implementation was intentionally not run through the expensive local suites; GitHub CI is the verification authority for this update.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

mknayam:
## Live reproduction attempt on `main`

I tested an account switch using an isolated `CODEX_HOME`.

Verified:

- AO's Codex app-server started under account A.
- The app-server PID remained unchanged during the test.
- The isolated credential directory was switched to account B.
- AO manually requested provider compaction.
- The provider successfully compacted the conversation.
- A subsequent Chat turn completed successfully.

The account-A access token held by the existing app-server was valid for 10 days. Logging account A out through the isolated CLI did not invalidate
that in-memory access token. Consequently, the provider did not return the
structured `unauthorized` error required to reproduce the reported failure.

backend/internal/adapters/chatdriver/codexappserver/conversation.go
```diff
@@ -178,6 +179,20 @@ func (c *conversation) pump() {
...
+			}
+		}
+		if stopForReauth {
+			// Signing in rewrites Codex's credential file, but this app-server has
+			// already proved that the credentials it holds in memory are stale. End
+			// only this controller; the existing Resume agent flow can then open a new
+			// process against the same thread after the user signs in. This is
+			// intentionally not an automatic retry: retrying before the user fixes auth
+			// would create a process loop and repeatedly fail queued work.
+			_ = c.Close()
```
whoisasx: [P1] Codex emits a non-retrying `error` notification before the terminal `turn/completed`. The reauth event produced from that `error` reaches this branch, which stops the process immediately and leaves the authoritative completion unread; AO then settles the turn through generic controller-shutdown cleanup instead of preserving Codex's terminal status and error. Please defer shutdown until the matching `turn/completed` arrives, with a bounded fallback for a genuinely missing completion, and add a driver-level `error(false) -> turn/completed` regression test.

frontend/src/renderer/components/chat/ChatStatusBanners.tsx
```diff
@@ -36,6 +36,8 @@ export const ReauthBanner = memo(function ReauthBanner({
 }) {
 	if (!account.reauthRequiredAt) return null;
 	const command = signInCommand(harness);
+	const nextStep =
+		harness === "codex" ? "resume the agent and resend your message" : "send your message again";
```
whoisasx: [P1] A structured unauthorized error can arrive in the middle of a turn after commands or file changes have already executed. This new unconditional resend instruction, combined with the claim below that the worktree is untouched, can cause those side effects to be repeated. Please use recovery copy that warns the user work may be partial and asks them to inspect it before resending, unless the backend can positively establish that no side effects occurred.

backend/internal/service/chat/controller.go
```diff
@@ -1933,6 +1937,18 @@ func (c *Controller) apply(ctx context.Context, event ports.ChatEvent) error {
...
 		}
+		rootConversation := event.ProviderConversationID == "" ||
+			event.ProviderConversationID == c.conv.ProviderConversationID()
+		if live && rootConversation && state == domain.TurnStateCompleted && event.Err == nil {
+			// TurnStarted only proves that the provider accepted the request envelope.
+			// A successful live root completion proves the replacement controller could
+			// use its credentials for real work, so only then is the warning no longer
+			// current. Replayed historical successes prove nothing about current
+			// credentials, and the historical timeline notice remains.
+			if err := c.clearReauth(ctx, now); err != nil {
```
whoisasx: [P2] `clearReauth` is auxiliary banner cleanup, but returning its error from this transaction also rolls back `SettleTurn`. `projectEvent` then skips the committed lifecycle update, leaving `pendingTurnID` set and the controller busy even though the provider completed successfully. The added test manually emits the completion again, whereas native-history import uses `live=false` and cannot retry this clear. Please keep authoritative turn completion independent from reauth-warning cleanup.

backend/internal/service/chat/controller.go
```diff
@@ -2250,6 +2266,32 @@ func (c *Controller) apply(ctx context.Context, event ports.ChatEvent) error {
...
+	account.ReauthRequiredAt = nil
+	account.ReauthReason = ""
+	c.mu.Unlock()
+
+	// This runs inside ProjectProviderEvent's transaction. Volatile memory is
+	// cleared in afterProject only after that outer transaction commits.
+	return c.store.RecordAccount(ctx, c.conversation.ID, account, now)
+}
+
+func (c *Controller) clearReauthMemory() {
```
whoisasx: [P2] This clears `c.account` but leaves `c.conversation.Account` with the reauth warning loaded during resume. Edit/branch replacement copies `source.conversation` into the replacement controller, so a subsequent ordinary account update can merge and persist the old timestamp and reason again. Please synchronize the branch seed with the cleared account state, or construct replacement controllers from current durable state.

backend/internal/adapters/chatdriver/codexappserver/normalize.go
```diff
@@ -646,10 +646,30 @@ func normalizeNotification(n notification, now time.Time) []ports.ChatEvent {
...
+			Kind:                   ports.ChatEventError,
+			ProviderTurnID:         p.TurnID,
+			ProviderConversationID: p.ThreadID,
+			Err:                    fmt.Errorf("%s", message),
+		}
+		if p.WillRetry {
+			// Codex still owns recovery while it says it will retry. Stopping here
+			// would turn a provider-managed transient failure into a user-visible
+			// manual recovery flow before the provider has made a terminal decision.
+			return []ports.ChatEvent{ev}
```
whoisasx: [P2] Although Codex still owns recovery when `willRetry` is true, this returns a `ChatEventError`, and the controller persists every such event as a failed error activity with a fresh ID. If the retry succeeds, the timeline still contains one or more permanent failure rows for a successful turn. Please drop this transient notification or model it as retry/recovery state rather than a terminal failed activity, and cover the successful-retry projection end to end.

whoisasx (changes requested):
The structured `codexErrorInfo == unauthorized` discriminator, manual recovery boundary, and refusal to infer auth state from provider prose are the right foundation. I found five independent lifecycle and UX issues that should be addressed before merging: the pre-terminal error path can discard the authoritative completion; the recovery copy can encourage duplicate side effects; a banner-clear write failure can roll back a successful turn; branch replacement can resurrect cleared reauth state; and retrying errors are persisted as permanent failures.

Verification: all GitHub checks are green; the PR merged cleanly with current `main`; focused Codex adapter and chat-service tests passed under `-race`; and the banner suite passed 18/18. I also checked the notification order against Codex's app-server implementation and reproduced the missing `turn/completed` sequence in an isolated test. Requesting changes for the inline findings below.

backend/internal/adapters/chatdriver/codexappserver/conversation.go
mknayam: Fixed. The reauthentication event now retains the provider turn/conversation correlation. When a non-retrying authentication error arrives before turn/completed, the pump emits the error and account warning but waits for the matching terminal completion before stopping the stale app-server. A bounded three-second fallback handles a genuinely missing completion.  
Added TestUnauthorizedErrorWaitsForTheMatchingTurnCompletion, covering error(willRetry=false) -> turn/completed and verifying that AO does not stop early and preserves Codex’s authoritative terminal state and error.

frontend/src/renderer/components/chat/ChatStatusBanners.tsx
mknayam: Fixed. The banner no longer claims the worktree is untouched or instructs the user to resend unconditionally. It now warns that the failed turn may already have changed files or run commands and asks the user to inspect the timeline and worktree before deciding whether to resend. The Codex flow still explains that the agent must be resumed after signing in.  
Updated the component tests to assert the side-effect warning and ensure the old “worktree is untouched” copy is absent.

backend/internal/service/chat/controller.go
mknayam: Fixed. SettleTurn and provider-event projection now commit independently of reauthentication-warning cleanup. After the completion commits and the controller lifecycle moves back to ready, afterProject attempts the auxiliary account update separately. A cleanup failure is logged, leaves the warning available for retry, and cannot roll back the completed turn or leave the controller busy.  
Native-history import does not call afterProject, so historical successful turns cannot clear current authentication state. The regression test now verifies that a failed cleanup still leaves the first turn durably completed and the controller ready, while a later live successful turn retries and clears the warning.

backend/internal/service/chat/controller.go
mknayam: Fixed. Replacement controllers are now created from a synchronized snapshot that combines the original conversation launch state with the controller’s current account state. Both message editing and branch activation use this snapshot, and successful warning cleanup also keeps the conversation account seed synchronized.  
Added a regression test that begins with persisted reauthentication state, clears it through a successful live turn, creates an edited branch, applies an ordinary account update, and verifies that the old timestamp and reason are not resurrected.

backend/internal/adapters/chatdriver/codexappserver/normalize.go
mknayam: Fixed. Notifications with willRetry: true are now dropped at the adapter normalization boundary because Codex still owns recovery and the notification is explicitly non-terminal. Therefore no ChatEventError reaches the controller projector and no permanent failed activity can be created for a turn that later succeeds.  
Added normalization coverage and a driver-level wire regression that sends turn/started -> error(willRetry=true) -> successful turn/completed, verifies that no error event is exposed to the projector, and preserves the successful terminal completion.

mknayam:
Addressed all five review findings:
Authentication failures no longer stop the Codex app-server before the matching authoritative turn/completed; shutdown now has a bounded fallback.
Recovery copy warns about possible partial side effects and asks users to inspect before resending.
Reauthentication-warning cleanup is independent from authoritative turn settlement, so cleanup failures cannot roll back completion or leave the controller busy.
Edit and branch replacement controllers use current account state and cannot resurrect a cleared warning.
willRetry notifications remain provider-owned transient state and no longer create permanent failed timeline activities.
Added focused driver, normalization, service persistence, branch-replacement, and frontend regressions. Validation completed with focused tests, repeated regression runs, race tests, frontend typechecks and component tests, go build ./..., go vet ./..., and the repository lint suite.

backend/internal/service/chat/controller.go
```diff
@@ -2262,6 +2310,18 @@ func (c *Controller) afterProject(ctx context.Context, event ports.ChatEvent, pr
 			c.reportActivity(ctx, domain.ActivityActive, "chat.turn.started", now)
 		}
 	case ports.ChatEventTurnCompleted:
+		rootConversation := event.ProviderConversationID == "" ||
```
whoisasx, line 2878: [P1] This completion still runs the normal queue drain even though the preceding structured unauthorized event has already marked the account as requiring reauthentication and the adapter is closing this controller. Depending on scheduling, a queued message can be dispatched into stale credentials or hit the closed transport and be permanently settled as failed. I reproduced this sequence and the provider received both the failed turn and the queued message. Please fence dispatch/draining while reauthentication is required, define how retained queued work resumes after the user explicitly repairs authentication, and add a regression test proving no queued message reaches the provider during this transition.

whoisasx (changes requested):
All five findings from the previous review are addressed correctly: AO now waits for the authoritative completion before stopping, the recovery copy warns about possible partial side effects, turn settlement is independent from warning cleanup, replacement controllers use current account state, and retrying provider errors no longer become permanent failures.

One additional lifecycle blocker remains: after the unauthorized event marks the account as requiring reauthentication, the failed root completion still executes the normal queued-turn drain while the adapter is closing the stale process. That can send queued work into stale credentials or permanently fail it against the closing transport. Please address the inline finding before merge.

Verification: all GitHub checks are green; the PR merges cleanly with current main; focused Codex adapter and chat-service tests pass under `-race`; and `git diff --check` passes. I also reproduced the queue issue with an isolated service regression test: after the reauthentication warning and failed completion, the provider received the queued second message.

backend/internal/service/chat/controller.go
mknayam: Fixed. The reauthentication account event now establishes a controller-level provider-delivery fence under the same dispatch lock used by Send and completion draining. Once that fence is raised, new messages remain durably queued and turn/completed cannot drain existing queued work into the stale controller.
Authentication shutdown now settles only turns that actually reached the provider; undelivered queued rows are retained. Automatic daemon recovery preserves the queue and closes the still-blocked replacement controller, so it cannot silently retry work or hide the explicit recovery action. The user’s explicit “Resume agent” action is propagated as the sole authority to lift the replacement controller’s fence and dispatch the oldest retained message. The durable warning remains until that controller completes live work successfully.
Added TestReauthenticationRetainsQueueUntilExplicitResume, covering the reported unauthorized-completion sequence, a message submitted after the auth failure, controller shutdown, automatic recovery, and explicit resume. It verifies that no queued message reaches either the stale or automatically reconstructed provider and that explicit resume releases exactly the retained queue head.

mknayam:
Addressed the follow-up queue-safety review:
Reauthentication now fences all new-turn dispatch and completion-driven queue draining.
Undelivered queued messages survive authentication-related controller shutdown.
Automatic daemon recovery cannot release retained work and leaves the controller stopped for explicit recovery.
“Resume agent” is the deliberate recovery boundary that reloads credentials and dispatches the oldest retained message.
Ordinary restore and agent-switch recovery do not receive this authority.
Added end-to-end SQLite-backed regression coverage and regenerated sqlc bindings.
Validation passed: affected package suites, regressions repeated 20 times, race tests for Chat and session manager, go build ./..., go vet ./..., and the complete npm run lint suite with zero lint findings.

backend/internal/service/chat/service.go
```diff
@@ -450,6 +467,23 @@ func (s *Service) Start(ctx context.Context, cfg StartConfig) (*Controller, erro
 	s.startConfigs[cfg.SessionID] = cloneStartConfig(cfg)
 	controller.start()
 	s.mu.Unlock()
+	switch {
+	case cfg.ResumeRetainedQueue:
+		controller.resumeRetainedQueue(context.WithoutCancel(ctx))
```
whoisasx: [P1] This explicit recovery path calls `turn/start` synchronously with `context.WithoutCancel(ctx)`. The Codex RPC relies on its caller context to bound the response wait, so if the replacement app-server starts but never answers `turn/start`, the Resume HTTP request and the per-session resume-operation fence remain held indefinitely. Please use a detached but bounded service-owned context, or dispatch asynchronously with an equivalent timeout, so this recovery action cannot hang forever.

backend/internal/service/chat/controller.go
```diff
@@ -2284,6 +2403,12 @@ func (c *Controller) afterProject(ctx context.Context, event ports.ChatEvent, pr
 		}
 	case ports.ChatEventAccountChanged:
 		if event.Account != nil && event.Account.ReauthRequired {
+			// This event is projected before the matching completion, on the same
+			// projector goroutine. Raising the fence here makes that completion retain
+			// the queue instead of dispatching into credentials the provider rejected.
+			c.mu.Lock()
```
whoisasx, line 2928: [P2] This fence is raised only from `afterProject`, so it is skipped whenever `ProjectProviderEvent` fails, including a `RecordAccount`, warning-activity, or transaction-commit failure. The following `turn/completed` then runs `drainLocked` with `reauthBlocked == false` and can consume retained work against the rejected or closing credentials. Reauthentication should fail closed: raise the volatile fence under `sendMu` before attempting projection, then persist/report the warning separately, with a projection-failure regression for `account.changed -> turn.completed`.

whoisasx (changes requested):
The previous six findings are addressed, and the queue-retention direction is correct. Two blockers remain in the explicit recovery path: Resume can wait forever on an unbounded detached `turn/start` request, and a failed account-event projection leaves the delivery fence open so the following completion can consume retained work.

The PR also currently conflicts with `main` in `controller.go`, `history_test.go`, and `conversation_store.go`. Please preserve `main`'s generation-fenced `CleanupOwnedControllerWork` behavior while adding the reauthentication-specific retain-queued mode; resolving toward the PR's unconditional cleanup would let an old controller settle work owned by its replacement.

Verification: the previous review fixes were rechecked at `e39d74e`; `git diff --check` passes. GitHub currently shows the frontend and CLI E2E checks passing, while Go build/lint/API-drift are cancelled and gitleaks is queued. Per our review policy, I did not run Go tests.

backend/internal/service/chat/service.go
mknayam: Fixed in d5d6b6da1. Explicit retained-queue recovery now wraps the detached context in a service-owned 10-second timeout before calling turn/start, so Resume and its per-session operation fence cannot wait indefinitely. The end-to-end reauthentication test now asserts that the provider dispatch receives this bounded deadline.

backend/internal/service/chat/controller.go
mknayam: Fixed in d5d6b6da1. The controller now takes sendMu and raises reauthBlocked before ProjectProviderEvent, so RecordAccount, warning activity, or transaction-commit failure still leaves the following completion fenced. Added an account.changed projection-failure regression proving the retained message stays queued. The main conflicts were resolved by preserving generation-owned CleanupOwnedControllerWork and adding a separate generation-fenced retain-queued cleanup mode, with store coverage for stale and current generations.

AgentWrapper:
Fresh live reproduction was added to #4296 on 2026-09-04. It confirms the structured problem on AO `0.12.11-nightly.202609031704`: the affected persistent chat host/app-server predates the external `~/.codex/auth.json` account change by two days, current `codex login status` is healthy, and new TUI processes work while every turn on the old Chat process fails and the thread becomes `system_error`.

Please account for two changes since this PR's original design:

1. Persistent chat-host reattachment now deliberately preserves the already-initialized app-server across daemon/UI restarts, so recovery must retire the exact stale host before resuming; an ordinary controller reconnect is insufficient.
2. `main` now includes AO-managed Codex global account switching (#4722). The remaining live failure is specifically an account change made outside AO. Rebase/resolve against current `main` and integrate with that account reconciliation/switch lifecycle rather than creating a second unfenced restart path.

Current GitHub state: this PR is `CONFLICTING`/`DIRTY` against `main` at `d2c88dae7d968df26c990ac5dd42ed2fc17513b1`. The live report raises confidence from inferred to observed and gives a real acceptance case for the recovery tests.

PR #4356, opened 2026-08-24
fix(cli): carry reviewers through project set-config (#4277)

### LINKED ISSUE #4277: bug(cli): project set-config silently drops the reviewers config field
## Summary

`ao project set-config --config-json` silently drops the `reviewers` field. The command returns `status: ok` and prints the project back, so the write looks like it succeeded, but the reviewer config is never persisted. `PUT /api/v1/projects/{id}/config` with the identical body persists it correctly, so this looks like the CLI (or its request struct) not carrying the field through to the daemon.

This is quiet in a way that matters: `reviewers` is what pins cross-family review (Claude reviewing Codex PRs). If it silently goes missing, the review loop just doesn't happen and nothing says so.

## Repro

```bash
ao project set-config work --config-json '{"reviewers":[{"harness":"claude-code"}],"sessionPrefix":"work"}' --json
```

Response (abridged) — no `reviewers`:

```json
{ "project": { "id": "work", "config": {
    "sessionPrefix": "work",
    "agentConfig": {}, "worker": { "agentConfig": {} },
    "orchestrator": { "agentConfig": {} }, "trackerIntake": {} } } }
```

Store confirms it:

```bash
sqlite3 ~/.ao/data/ao.db "select json_extract(config,'\$.reviewers') from projects where id='work';"
# (empty)
```

Same payload through the daemon API persists fine:

```bash
curl -s -X PUT http://127.0.0.1:$PORT/api/v1/projects/work/config \
  -H 'content-type: application/json' \
  -d '{"config":{"reviewers":[{"harness":"claude-code"}],"sessionPrefix":"work"}}'
# -> config.reviewers = [{"harness":"claude-code"}], and it is in ao.db
```

## Expected

Either the field round-trips like every other `ProjectConfig` field, or the CLI rejects it with an explicit error. Silently accepting and discarding is the worst of the three.

## Notes

- `reviewers` is present in `ProjectConfig` in the OpenAPI schema (`items: DomainReviewerConfig`), so this reads as a CLI gap rather than an intentional API-only field.
- `ao project set-config` also has no flag for reviewers, so `--config-json` is the only CLI route to set it today. A `--reviewer` flag would be a natural fix alongside.
- Version: 0.12.6 (bundled daemon reports `ao version` = `dev`), macOS 15 arm64.

ApexYash11:
Fixes #4277

## Root cause

The CLI is a thin daemon client that hand-mirrors daemon DTOs. Its mirror struct `projectConfig` in `backend/internal/cli/project.go` had no `Reviewers` field, so:

1. `ao project set-config --config-json '{"reviewers":[...]}'` unmarshals into that struct and `encoding/json` silently discards the unknown `reviewers` key.
2. The CLI then PUTs the reviewers-less config to `PUT /api/v1/projects/{id}/config`, which replaces the whole config.

Result: `status: ok` with reviewers never persisted, while the identical payload through the daemon API works. The daemon API was correct; this was purely a CLI DTO gap. Notably, `reviewers` pins cross-family review, so the silent drop meant the review loop quietly never ran.

## Fix

- Add `reviewerConfig` (mirrors `domain.ReviewerConfig`) and `Reviewers []reviewerConfig` (`json:"reviewers,omitempty"`) to the CLI's `projectConfig`, so `--config-json` round-trips reviewers like every other field.
- Add a repeatable `--reviewer <harness>` flag to `project set-config`, wired through `buildProjectConfig`. Harness vocabulary validation stays daemon-side per the thin-client convention.
- Tests: `TestProjectSetConfig_ReviewerJSON` (the issue's repro) and `TestProjectSetConfig_ReviewerFlags`.

## Intentional omissions

- No CLI-side validation of reviewer harness names; the daemon already rejects unknown harnesses via its error envelope.
- `--reviewer` only applies on the flags path; per existing behavior `--config-json` overrides field flags.

## Verification

- `go test ./internal/cli/ -count=1` - full package passes, including the two new tests

### DIFF
--- a/backend/internal/cli/project.go
+++ b/backend/internal/cli/project.go
@@ -94,6 +94,11 @@ type trackerIntakeConfig struct {
 	Assignee string `json:"assignee,omitempty"`
 }
 
+// reviewerConfig mirrors domain.ReviewerConfig.
+type reviewerConfig struct {
+	Harness string `json:"harness"`
+}
+
 // projectConfig mirrors the daemon's typed domain.ProjectConfig for the CLI
 // client. The CLI sets common fields via flags and the whole object via
 // --config-json.
@@ -111,6 +116,7 @@ type projectConfig struct {
 	Orchestrator      roleOverride        `json:"orchestrator,omitempty"`
 	TrackerIntake     trackerIntakeConfig `json:"trackerIntake,omitempty"`
 	AutoReview        bool                `json:"autoReview,omitempty"`
+	Reviewers         []reviewerConfig    `json:"reviewers,omitempty"`
 }
 
 // setConfigRequest mirrors the daemon's SetConfigInput body for
@@ -135,6 +141,7 @@ type projectSetConfigOptions struct {
 	trackerIntake     bool
 	trackerRepo       string
 	trackerAssignee   string
+	reviewers         []string
 	configJSON        string
 	clear             bool
 	json              bool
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

neversettle17-101 (approved):

PR #4357, opened 2026-08-24
fix(terminal): stop resize jitter, blanking, and repaint storms

### LINKED ISSUE #4155: Terminal flickers and looks like it reloads while resizing
## What happened

The terminal did not resize smoothly when the app layout changed. Opening or closing the inspector, dragging its edge, or resizing the app window could make the text shake, briefly disappear, and then come back as if the terminal had reloaded.

This was especially noticeable when Claude or Codex was showing a full-screen interface. The session itself was still running and no work was lost, but the repeated flashing made the terminal feel unstable and difficult to follow.

## How to reproduce

1. Open a session that is showing a full-screen terminal interface.
2. Open or close the inspector panel, or drag the divider beside it.
3. You can also slowly resize the app window from its edge.
4. Watch the terminal while its width is changing.

The text may jump around, the terminal may flash black for a moment, and the whole screen may redraw several times before settling.

## What should happen

The terminal should follow the available space without repeatedly flashing or looking like it restarted. Once the resize ends, it should quickly settle on the correct number of rows and columns.

## Why it happened

The app was doing too much work for every tiny movement during a resize.

A single drag can produce many size updates in a very short time. Each update could make the terminal recalculate its layout and ask the program inside it to redraw. On Windows, those redraws are particularly expensive because the terminal program may send most of its screen again after every size change.

There was also a short moment where the old terminal canvas had been cleared but the new content had not been painted yet. That is what caused the black flash. At the same time, the inspector animation kept changing the terminal's available width, creating even more resize work.

These smaller problems happened together, which is why earlier fixes that changed only one resize timer did not fully solve the issue.

## How it was fixed

The terminal now handles resize activity in a more controlled way:

- It avoids repeating expensive layout work when the usable terminal grid has not actually changed.
- It limits how often a new size is sent to the running terminal program during a drag, while still guaranteeing that the final size is sent immediately.
- It keeps the short clear-and-redraw response together so users do not see an empty frame in between.
- The inspector reserves its space in one step and then slides visually, instead of continuously squeezing the terminal throughout the animation.
- The terminal renderer was updated to a version that includes an upstream fix for blank frames during resize.

The full implementation details are documented in [PR #4357](https://github.com/Untrivial-ai/agent-orchestrator/pull/4357).

## Verification

- 202 focused terminal tests passed.
- Frontend TypeScript checks passed.
- All GitHub CI checks passed, including frontend tests and Windows, macOS, and Linux jobs.

## Related work

- [PR #4156](https://github.com/Untrivial-ai/agent-orchestrator/pull/4156) was the first attempt and was closed without merging.
- [PR #4357](https://github.com/Untrivial-ai/agent-orchestrator/pull/4357) contains the complete fix.
- [Issue #2333](https://github.com/Untrivial-ai/agent-orchestrator/issues/2333) covers other terminal blanking cases that are not caused by resizing.

JAYATIAHUJA:
Fixes #4155

## Problem

Resizing the desktop terminal could produce text jitter, blank WebGL frames, and repeated full-screen TUI repaints. The symptom was strongest on Windows because every intermediate PTY grid update can make ConPTY clear and rewrite the viewport.

## Root cause

The visible failure was not a single debounce bug. Per-frame xterm reflow, competing geometry owners, xterm 5.5's deferred WebGL repaint after canvas resize, frequent ConPTY grid publications, split clear/content transport chunks, and inspector gap-width animation all compounded one another.

## Before

https://github.com/user-attachments/assets/885e5bea-51b9-4f77-b400-5cdd9f18b549

## After 

https://github.com/user-attachments/assets/57b12fbc-c048-4f57-8bba-c26ffb377917

## What changed

- Make observer-driven fitting the single geometry owner while a fit is pending.
- Apply rows live while delaying expensive column reflow only for large normal buffers, then flush at gesture end.
- Pace PTY resize publication during separator and OS-window resize gestures, with a guaranteed final grid.
- Coalesce only the short resize-response output burst into one atomic xterm write so a cleared frame cannot paint by itself.
- Reserve inspector width in one layout update and animate the panel with transforms instead of relayouting xterm every spring frame.
- Upgrade the terminal and compatible addons to an exact-pinned xterm 6.1 beta cohort containing upstream's synchronous WebGL resize repaint fix.
- Remove the Canvas addon incompatibility and private render-service/scrollbar mutations.
- Add regression coverage, design notes, and dev-only resize diagnostics.

## Dependency versions

- `@xterm/xterm`: `6.1.0-beta.303`
- `@xterm/addon-webgl`: `0.20.0-beta.299`
- `@xterm/addon-fit`: `0.12.0-beta.300`
- `@xterm/addon-search`: `0.17.0-beta.300`
- `@xterm/addon-unicode11`: `0.10.0-beta.300`
- `@xterm/addon-web-links`: `0.13.0-beta.300`

## Verification

- [x] `npx vitest run --config vite.renderer.config.ts src/renderer/components/SessionView.test.tsx src/renderer/components/XtermTerminal.test.tsx src/renderer/hooks/useTerminalSession.test.tsx src/renderer/lib/terminal-mux.test.ts`
  - 4 test files passed
  - 202 tests passed
- [x] `npm ls` resolves a single deduplicated xterm 6.1 core with compatible addons and no Canvas addon.
- [x] Replayed onto current upstream `main` (`d32434d9d`) and resolved the current inspector sizing/browser-popout conflicts without dropping their behavior.

## Scope notes

The resize diagnostics are development-only (`import.meta.env.DEV` plus Vite `configureServer`) and do not add a production endpoint. The full repository test command has unrelated Windows baseline failures in macOS artifact scripts, landing-test dependency resolution, and worker startup; all touched terminal suites and frontend typechecking pass.


frontend/package.json
```diff
@@ -79,13 +79,12 @@
...
-		"@xterm/addon-unicode11": "^0.9.0",
-		"@xterm/addon-web-links": "^0.11.0",
-		"@xterm/addon-webgl": "^0.19.0",
-		"@xterm/xterm": "^5.5.0",
+		"@xterm/addon-fit": "0.12.0-beta.300",
+		"@xterm/addon-search": "0.17.0-beta.300",
+		"@xterm/addon-unicode11": "0.10.0-beta.300",
+		"@xterm/addon-web-links": "0.13.0-beta.300",
+		"@xterm/addon-webgl": "0.20.0-beta.299",
+		"@xterm/xterm": "6.1.0-beta.303",
```
Annieeeee11, line 87: **Blocker: exact-pinned beta cohort.** Six packages pinned to specific beta builds with no range. `6.1.0-beta.304` and `addon-fit 0.12.0-beta.301` already exist, so this needs hand-bumping on every upstream beta and nobody on the team has signed up for tracking the xterm beta channel. The upgrade is justified (xterm.js#5529 merged 2025-12-28, after 6.0.0 on 2025-12-22, so the sync-render fix really is beta-only), but it needs to be its own PR where reviewers can weigh that trade-off, and it needs a stated exit plan (move to `^6.1.0` on stable release). This also conflicts with `main`'s `package.json`/lockfile right now.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -416,6 +400,7 @@ export function XtermTerminal(props: XtermTerminalProps) {
 					'ui-monospace, Menlo, Monaco, "Courier New", monospace',
 				fontSize: props.fontSize ?? TERMINAL_FONT_SIZE_DEFAULT,
 				lineHeight: 1.35,
+				scrollbar: { showScrollbar: false },
```
Annieeeee11, line 403: **Blocker: regresses the macOS scrollbar that landed on `main` after this branch was cut.** `main` now has `configureScrollbarReservation()` setting `viewport.scrollBarWidth = isMacPlatform() ? MAC_TERMINAL_SCROLLBAR_WIDTH : 0` (7px), plus a custom `scrollbarTrackRef` with pointer handlers that render an AO-styled scrollbar over that reserved strip. Setting `showScrollbar: false` here and deleting the reservation helper collapses that strip to zero and breaks the custom track. On rebase this needs to become `showScrollbar: isMacPlatform()` (or equivalent via the public option, which is a nice win over the private `_core.viewport` mutation), with the track code kept.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -82,32 +82,24 @@ export type XtermTerminalProps = {
...
 
-// Prefer the WebGL renderer, fall back to 2D canvas. Both rasterize box-drawing
-// glyphs themselves onto a fixed cell grid; the DOM renderer does not, so TUI
-// borders would drift. Loaded after open().
+// Prefer WebGL and let xterm fall back to its built-in renderer when WebGL is
+// unavailable. The old Canvas addon only supports xterm 5 and cannot be loaded
+// with the 6.1 line. WebGL stays loaded for the terminal's whole lifetime — a
+// visibility-scoped variant was tried and reverted because releasing the
+// context could visibly change an on-screen terminal's rendering.
 function loadRenderer(term: Terminal): void {
```
Annieeeee11, line 90: **Blocker: Canvas fallback removed with nothing equivalent behind it.** With xterm 6 the fallback on WebGL failure or context loss is the DOM renderer. Your own `docs/terminal-resize-jitter.md` (root cause #2) says the DOM renderer produces visibly different text and box-drawing drift for TUIs, which is exactly the "terminal UI changed" report. Context loss is not hypothetical: #4949 (open) was filed from a real v0.12.10 report of "Too many active WebGL contexts" blanking long-parked panes, and that PR hardens the WebGL-loss to Canvas path this function deletes. The two PRs cannot both merge as written. Either (a) keep a Canvas-equivalent path (xterm 6 has no Canvas addon, so this likely means capping retained WebGL contexts in the terminal cache as your doc suggests) or (b) explicitly accept DOM fallback and close #4949 with that rationale. Needs a decision, not a silent removal.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -635,19 +615,34 @@ export function XtermTerminal(props: XtermTerminalProps) {
...
+		// - Normal buffer under 200 lines → resizing is CHEAP (an alt-screen agent
+		//   TUI never grows the normal buffer), so fit fully live on every observed
+		//   frame. This is why VS Code's terminal tracks a splitter drag smoothly —
+		//   the grid follows the pane in the same frame, no freeze and no snap.
+		// - Normal buffer at/over 200 lines → the column reflow rewraps all of
+		//   scrollback, so rows still apply live but columns wait for a 100ms quiet
+		//   window (flushed early on pointer release).
+		// xterm 6.1 renders synchronously after each geometry change, preventing the
+		// blank canvas frame from xterm.js#4922 (fixed upstream by #5529).
+		const CHEAP_RESIZE_NORMAL_BUFFER_LINES = 200;
```
Annieeeee11, line 642: This block is the part of the PR most worth saving, and it is also where the rebase will hurt most. `main` (#5172) replaced the `data-terminal-live-resize` branch of the observer with `scheduleLiveFit()`, which coalesces observer deliveries to one `fitTerminal()` per rAF specifically to avoid a nested layout cycle from calling `fit()` inside the ResizeObserver callback. The live path here (`fitTerminal()` directly in `scheduleVisibleFit` when `normal.length < 200`) reintroduces that synchronous fit-in-observer. On rebase, please keep the rAF coalescing and put the <200-lines branch behind it. The `onRender` stabilizer guard (yielding while `fitQuietTimer`/`fitCapTimer` are armed) is a genuine bug fix independent of everything else and would be a good first split-out PR.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -682,18 +681,55 @@ export function XtermTerminal(props: XtermTerminalProps) {
...
 		fitRef.current = scheduleVisibleFit;
+		// A separator drag ends on pointer release, not on a quiet box: collapse the
+		// remaining quiet window so the single reflow lands on the same gesture. An
+		// activation window (fitAllowsHidden) is left alone — it must still reveal
+		// only after the cache has finished preparing the real slot.
+		const flushFitOnPointerRelease = () => {
+			if (fitAllowsHidden || fitQuietTimer === null) return;
+			flushScheduledFit();
+		};
+		window.addEventListener("pointerup", flushFitOnPointerRelease, true);
```
Annieeeee11, line 726: Minor, not blocking: a capture-phase `pointerup` on `window` fires for *any* click anywhere in the app, not just the end of a separator drag. If a fit quiet window happens to be armed (e.g. a `window.resize` event 50ms earlier), an unrelated click flushes it early. Harmless in practice since it just lands the fit sooner, but scoping this to `body.is-resizing-x` (which `useResizable` already sets and which `useTerminalSession.ts` already reads) would make the intent explicit and avoid surprising interactions with the activation path.

frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -87,10 +88,45 @@ export type UseTerminalSessionOptions = {
...
+// gesture's final grid always lands right after release via the recheck loop.
+const GESTURE_PUBLISH_INTERVAL_MS = 350;
+// An OS window-edge drag streams window resize events with no pointer signal
+// the renderer can observe, so unlike separator drags there is no marker class.
+// Treat the window as mid-gesture until its resize events go quiet — otherwise
+// each mid-drag pause publishes a grid and ConPTY answers with a full repaint.
+const WINDOW_RESIZE_GESTURE_QUIET_MS = 250;
+let lastWindowResizeAt = Number.NEGATIVE_INFINITY;
+if (typeof window !== "undefined") {
+	window.addEventListener("resize", () => {
```
Annieeeee11, line 120: Module-level `window.addEventListener("resize", ...)` with no teardown, evaluated at import time. This leaks across vitest module reloads and HMR, and `performance.now()` at module scope means `resizeGestureActive()` treats the first 250ms after any HMR as a gesture. Also `document.body` in `resizeGestureActive()` will throw in any non-DOM test that imports this module. Suggest moving the listener into the hook's effect (or the existing `connect` closure) with a disposer, alongside the other listeners.

frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -639,23 +743,50 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
...
+		// short interval and the final grid goes out right after release.
+		// Mid-gesture publishes are paced, not blocked outright: a fully held PTY
+		// leaves the TUI's old-size frame being cropped/blank-padded by the live
+		// local fits for the whole drag, then one big late repaint — which reads
+		// as the terminal reloading. Letting a settled grid through every
+		// GESTURE_PUBLISH_INTERVAL_MS refreshes real content at every pause of a
+		// slow drag (each repaint lands atomically via the gate below), while a
+		// fast drag still produces only the final publish.
+		let lastGesturePublishAt = Number.NEGATIVE_INFINITY;
+		const publishGrid = (cols: number, rows: number) => {
```
Annieeeee11, line 763: The pacing design is right for ConPTY and the test covers it well. Two things for the rebase: (1) `main` has since changed the resize/publish path here for the cloud duplex stream (#4768) and the harness login PTY (#5009), so `lastPublishedGrid`/`mux.resize` call sites have moved; this file conflicts. (2) `publishGrid` recurses through `r.resizeTimer` every 100ms while a gesture is active, but `terminal.onResize` also clears and reassigns `r.resizeTimer`. That is correct today because both paths write the same slot, but it is fragile; a short comment that the recheck timer intentionally shares `r.resizeTimer` so a newer grid supersedes it would save the next reader a trace. Please also confirm the `flushLiveOutput` gate does not interact with the predictive local echo added in #4768, since that path writes to the terminal outside `queueLiveOutput`.

frontend/src/renderer/lib/diag.ts
```diff
@@ -0,0 +1,29 @@
+// TEMPORARY dev-only diagnostics for the terminal resize investigation.
```
Annieeeee11, line 1: **Blocker: temporary instrumentation in a fix PR.** The header says TEMPORARY and "remove together with that middleware once the resize work settles". The resize work is settled by definition when this merges, so this file, the `ao-diag-logger` Vite middleware, and the `diag(...)` calls threaded through `XtermTerminal.tsx` and `useTerminalSession.ts` should be dropped before merge. If the team wants a renderer diag channel, that is a fine separate PR with its own review; it should not ride in on a flicker fix. (`import.meta.env.DEV` gating means the string literals still ship in the bundle, and the calls sit on the per-frame resize hot path they were meant to measure.)

frontend/vite.renderer.config.ts
```diff
@@ -136,6 +136,25 @@ export default defineConfig({
 		},
 	},
 	plugins: [
+		// TEMPORARY: prints renderer-side terminal diagnostics (src/renderer/lib/
+		// diag.ts) into the dev terminal, so resize event ordering is visible
+		// without devtools. Dev-server only; remove together with diag.ts.
+		{
+			name: "ao-diag-logger",
```
Annieeeee11, line 143: Same as `lib/diag.ts`: dev-server-only, but self-labelled TEMPORARY and unrelated to the fix. Please drop from this PR. If kept anywhere, it needs a body size cap: `body += chunk` with no limit on an unauthenticated dev endpoint is an easy way to OOM the dev server from any page in the same origin.

frontend/src/renderer/lib/bridge.ts
```diff
@@ -54,10 +54,19 @@ export const aoBridge: AoBridge =
...
-				message: "Electron preload is not available in browser preview.",
-			}),
+			// Browser preview normally has no daemon. With an explicit
+			// VITE_AO_API_BASE_URL the renderer is deliberately pointed at a
+			// running daemon (see the /api + /mux dev proxy in
+			// vite.renderer.config.ts), so report ready and leave the port unset —
+			// applyDaemonStatus then keeps the explicit base instead of deriving
+			// a direct 127.0.0.1 origin that would fail CORS from a browser tab.
+			getStatus: async () =>
+				import.meta.env.VITE_AO_API_BASE_URL
```
Annieeeee11, line 64: This changes browser-preview daemon status semantics for everyone who sets `VITE_AO_API_BASE_URL`, and it is described in the design doc as a debugging affordance for reproducing the resize issue. It is unrelated to the fix and changes behavior of the preview stub. Please split it out (with a note in the dev docs on how to point the browser preview at a real daemon) or drop it.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -281,20 +280,19 @@ function SessionInspectorRail({
 
 	const transition = prefersReducedMotion ? { duration: 0 } : SHELL_PANEL_SPRING;
 	const hidden = !isOpen && settledClosed;
+	const reservesSpace = isOpen || !settledClosed;
```
Annieeeee11, line 283: Reserving the gap width in one step and sliding the panel on `transform` is the right call and removes a whole class of per-frame reflow. Two rebase notes: `SessionView.tsx` has ~35 commits since your base (browser panel pop-out #5276/#5201, virtualized diff #5175, perf #5172) and conflicts today. And since the gap now snaps rather than animates, the `INSPECTOR_SPRING_MS` timer that keeps `data-inspector-transition` set is now only for label modes; worth a one-line comment so nobody later "fixes" the gap back to animating because the attribute name still implies a transition.

docs/terminal-resize-jitter.md
```diff
@@ -0,0 +1,177 @@
+# Terminal resize: jitter, blanking, and "reloading" — design notes
+
+Branch: `fix/terminal-resize-jitter` · August 2026 · Windows 11 primary target
```
Annieeeee11, line 3: Nit: branch name here is `fix/terminal-resize-jitter` but the PR branch is `fix/terminal-resize-stability`. More substantively, this doc documents the diag harness and the bridge hack as part of the shipped design; once those are split out, the "Temporary instrumentation" section and the `bridge.ts` paragraph should go with them. The root-cause analysis and the VS Code / Superset comparison are genuinely useful and worth keeping.

Annieeeee11 (changes requested):
Diagnosis checks out against current main: the animated inspector gap, the 100ms PTY resize debounce, and xterm 5.5 without the xterm.js#5529 fix are all still there. The resize semantics, stabilizer guard, and paced PTY publishing are the right fixes.

Blocking as-is, see inline:

Regresses the macOS 7px scrollbar that landed after this branch.
Drops the Canvas fallback without a replacement, collides with #4949.
Exact-pinned xterm betas and temporary diag/bridge/vite scaffolding shouldn’t ride in on a flicker fix.

i-trytoohard:
@JAYATIAHUJA Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4358, opened 2026-08-24
feat(cli): target a remote daemon with --url / AO_URL

AronPerez:
## Ticket

## Problem

Every `ao` command talks to the daemon on the machine it runs on: it reads `running.json`, checks the PID is alive, and calls `127.0.0.1`. There is no way to drive a daemon on another machine, so a second box can only be operated by opening an SSH session to it first. Anyone running agents across more than one machine pays that tax on every command.

## Solution

A persistent `--url` flag (env `AO_URL`) points the same commands at another machine's LAN listener, skipping the run-file and the local liveness check, so it works from a machine that has never run AO. The credential is the daemon's existing connection password sent as a Bearer token. With no `--url` and no `AO_URL`, every path is the one it has always been.

## How Has This Been Tested?

`cd backend && go build ./... && go vet ./... && go test ./...` on the current `main` (`c9a0adb2`): 158 packages ok, 0 failures, `gofmt -l` empty. 15 new tests in `internal/cli/remote_test.go` cover scheme validation, userinfo rejection (including that the error never echoes the credential), the local default, `AO_TOKEN` precedence over `remotes.json`, loose-permission refusal, run-file use on the local path and its absence on the remote one, remote `ao status` including the lockout case, both `ao stop` refusals, and the telemetry drop. No frontend file and no OpenAPI surface is touched.

No renderable surface: this PR is Go CLI code only, no files under `frontend/src`. Behaviour is covered by the tests above.

## Implementation notes

The credential comes from `AO_TOKEN`, or from `~/.ao/remotes.json`, which mirrors the mobile app's saved-node list and must be mode 0600 — the CLI refuses to read it otherwise (Windows is exempt, where the mode bits carry no meaning). An entry whose URL does not parse is skipped rather than fatal, so a hand-edited or SSH-style entry cannot break every other one.

The URL must not carry userinfo. It is rejected textually, before `url.Parse`, so no error path can echo the password back — and the rejection is checked first, so a malformed URL cannot smuggle a credential into a parse error. The scheme-less form (`user:pw@host:3011`) is caught too; an `@` in the path or query is not authority and still parses.

`ao stop` refuses a remote target outright, including one that names loopback, so the single destructive verb never changes behaviour based on how a URL happens to look. CLI telemetry on `/internal/*` is dropped rather than sent off-box — those routes are 404'd at the LAN socket by design, and invocation counts have no business reaching someone else's daemon. A 429 from the LAN listener's per-source lockout reports as `locked_out` rather than `unhealthy`: the daemon is fine, the operator mistyped a password, and reporting `unhealthy` inverts the diagnosis.

`docs/cli/README.md` gains a section covering the flag, the credential, the 0600 requirement and the plaintext-HTTP-by-design caveat.

Trade-off worth naming: this is plain HTTP against a LAN listener, matching the existing mobile client's trust boundary. It is not a substitute for a tunnel over untrusted networks, and the docs say so.

### DIFF
--- a/backend/internal/cli/root.go
+++ b/backend/internal/cli/root.go
@@ -162,6 +162,7 @@ func (d Deps) withDefaults() Deps {
 func NewRootCommand(deps Deps) *cobra.Command {
 	deps = deps.withDefaults()
 	ctx := &commandContext{deps: deps}
+	var remoteURL string
 
 	root := &cobra.Command{
 		Use:           "ao",
@@ -171,12 +172,21 @@ func NewRootCommand(deps Deps) *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			// Resolve before anything else: every daemon call below, telemetry
+			// included, depends on which daemon this invocation targets.
+			remote, err := resolveRemoteTarget(remoteURL)
+			if err != nil {
+				return err
+			}
+			ctx.remote = remote
 			if shouldEmitCLIInvocation(cmd) {
 				ctx.emitCLIInvoked(cmd.Context(), cmd)
 			}
 			return nil
 		},
 	}
+	root.PersistentFlags().StringVar(&remoteURL, "url", "",
+		"Base URL of a remote AO daemon, e.g. http://host:3011 (env AO_URL); credential from AO_TOKEN or ~/.ao/remotes.json")
 	root.SetIn(deps.In)
 	root.SetOut(deps.Out)
 	root.SetErr(deps.Err)
@@ -217,6 +227,8 @@ func NewRootCommand(deps Deps) *cobra.Command {
 
 type commandContext struct {
 	deps Deps
+	// remote is nil for the default local daemon; see remote.go.
+	remote *remoteTarget
 }
 
 func shouldEmitCLIInvocation(cmd *cobra.Command) bool {
--- a/backend/internal/cli/status.go
+++ b/backend/internal/cli/status.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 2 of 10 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/cli/root.go
```diff
@@ -171,12 +172,21 @@ func NewRootCommand(deps Deps) *cobra.Command {
...
+				return err
+			}
+			ctx.remote = remote
 			if shouldEmitCLIInvocation(cmd) {
 				ctx.emitCLIInvoked(cmd.Context(), cmd)
 			}
 			return nil
 		},
 	}
+	root.PersistentFlags().StringVar(&remoteURL, "url", "",
```
illegalcall, line 188: `--url` is global, but several commands are not remote-safe yet. `ao start --url ...` opens the local app, `ao doctor --url ...` mixes remote daemon health with local machine checks, and `spawn` can reuse a local project ID or cwd against the remote host. Can we include #4361 and #4362 here?

backend/internal/cli/status.go
```diff
@@ -28,20 +29,38 @@ const (
 	stateStale     daemonState = "stale"
 	stateUnhealthy daemonState = "unhealthy"
 	stateNotReady  daemonState = "not_ready"
+	// stateLockedOut is reachable only for a remote target: the per-source
+	// lockout lives in the LAN listener's auth middleware, which the loopback
+	// path never reaches. See remoteProbeFailure.
+	stateLockedOut daemonState = "locked_out"
```
illegalcall, line 35: `runDoctor` does not handle `locked_out`, so it falls through as `PASS`. Please mark it as a failure or reject remote `doctor`.

backend/internal/cli/root.go
AronPerez: I kept these safeguards in the two existing sibling increments rather than duplicating them into this transport/base PR. #4361 (https://github.com/Untrivial-ai/agent-orchestrator/pull/4361) owns refusal of remote targets for `start`, `doctor`, and the other local-only commands; #4362 (https://github.com/Untrivial-ai/agent-orchestrator/pull/4362) owns remote project/path resolution, including preventing local `AO_PROJECT_ID`, `AO_SESSION_ID`, or cwd from selecting a remote project. Pulling either commit here would collapse the intentionally separate stack boundaries and leave duplicate/conflicting follow-up commits. Those PRs are therefore the intentionally sequenced safety follow-ups to this base transport change.

backend/internal/cli/status.go
AronPerez: Fixed in eecd2d042. `runDoctor` now classifies `stateLockedOut` as `FAIL` alongside `stateUnhealthy`, so a remote HTTP 429 cannot fall through as `PASS`. I added `TestDoctorReportsRemoteLockoutAsFailure`, which serves a 429 from the remote probe and asserts the daemon check is a failure containing `locked_out`.

i-trytoohard:

PR #4359, opened 2026-08-24
feat(daemon): read-only directory listing at GET /api/v1/fs/dirs

AronPerez:
## Ticket

## Problem

A client connected to a daemon on another machine cannot see that machine's filesystem, so an absolute project path over there has to be typed blind — with no way to tell a typo from a folder that is really missing, and no way to know whether a directory is already a checkout. Anyone adding a project on a second machine is guessing.

## Solution

A read-only `GET /api/v1/fs/dirs` answers with the subdirectories of one absolute path: names, paths, and whether each carries a `.git` entry (a directory for a clone, a file for a worktree checkout). Dotted names are skipped, files are never listed, and a listing is capped at 500 entries with `truncated` set. It is additive and sits behind the connection credential that already exists.

## How Has This Been Tested?

`cd backend && go build ./... && go vet ./... && go test ./...` on the current `main` (`c9a0adb2`): 158 packages ok, 0 failures, `gofmt -l` empty. That includes `TestListDirs` x4 and the LAN policy assertion. Regeneration is idempotent on this base: `go generate ./internal/httpd/apispec/...` followed by `npm run api:ts` leaves the working tree clean, so `TestRouteSpecParity` and the `api-drift` job both hold.

No renderable surface: Go plus two generated API files. The only `frontend/src` change is the regenerated `api/schema.ts`.

## Implementation notes

**Is this an escalation?** No. It sits behind the same connection credential that already authorises spawning an agent — that is, a shell — on the host, so it grants no reach that credential did not already have. It reads directory names and nothing else: no file contents, no sizes, no timestamps, no dotted names.

The daemon judges the path by its own OS's rules. A remote client cannot know what a valid absolute path looks like over there, so a relative path is refused by the daemon (`FS_PATH_NOT_ABSOLUTE`) rather than pre-judged client-side. `ENOENT`, `EACCES` and `ENOTDIR` are separated into 404 / 403 / 400 so a caller can tell "no such folder" from "may not read it" from "that is a file".

The route is declared in `specgen` beside the others and the spec and `schema.ts` are regenerated rather than hand-written.

The LAN test pins both halves of the policy: credential-gated like every other data route, and specifically **not** on `lanControlBlockedPrefixes`. That second half is the one worth having — a `ROUTE_LOOPBACK_ONLY` answer would kill remote browsing silently, and no loopback-side test could catch it. The assertion was falsified before it was trusted: adding `/api/v1/fs` to the blocked prefixes makes it fail.

Follow-up, not included here: nothing calls this endpoint yet. The client that uses it is a separate PR in the same series.

### DIFF
--- a/backend/internal/httpd/controllers/dto.go
+++ b/backend/internal/httpd/controllers/dto.go
@@ -1312,6 +1312,26 @@ type ImportRunResponse struct {
 	Report legacyimport.Report `json:"report"`
 }
 
+// ListDirsQuery is the query string accepted by GET /api/v1/fs/dirs.
+type ListDirsQuery struct {
+	Path string `query:"path,omitempty" description:"Absolute directory on the daemon host to list. When omitted, the daemon user's home directory."`
+}
+
+// FSEntry is one directory in a /api/v1/fs/dirs listing.
+type FSEntry struct {
+	Name    string `json:"name" description:"Directory name."`
+	Path    string `json:"path" description:"Absolute path of the directory on the daemon host."`
+	GitRepo bool   `json:"gitRepo" description:"True when the directory carries a .git entry (clone or worktree checkout)."`
+}
+
+// ListDirsResponse is the body of GET /api/v1/fs/dirs.
+type ListDirsResponse struct {
+	Path      string    `json:"path" description:"Absolute path that was listed."`
+	Parent    string    `json:"parent" description:"Absolute path of the listed directory's parent; equals path at the filesystem root."`
+	Entries   []FSEntry `json:"entries" description:"Subdirectories, excluding dotted names."`
+	Truncated bool      `json:"truncated,omitempty" description:"True when the listing hit the entry cap and more subdirectories exist."`
+}
+
 // DevImportProjectsRequest is the body of POST /api/v1/dev/import-projects.
 type DevImportProjectsRequest struct {
 	SourceDataDir string `json:"sourceDataDir" minLength:"1"`
--- a/backend/internal/httpd/controllers/fs.go
+++ b/backend/internal/httpd/controllers/fs.go
@@ -1,20 +1,20 @@
 package controllers
 
 import (
-	"errors"
+	"context"
 	"net/http"
-	"os"
-	"path/filepath"
-	"strings"
 
 	"github.com/go-chi/chi/v5"
 
-	"github.com/aoagents/agent-orchestrator/backend/internal/httpd/apierr"
+	"github.com/aoagents/agent-orchestrator/backend/internal/httpd/apispec"
 	"github.com/aoagents/agent-orchestrator/backend/internal/httpd/envelope"
+	fsbrowsersvc "github.com/aoagents/agent-orchestrator/backend/internal/service/fsbrowser"
 )
 
-// maxDirEntries bounds a single listing; deeper trees are reached by walking.
-const maxDirEntries = 500
+// DirectoryBrowserService lists directories on the daemon host.
+type DirectoryBrowserService interface {
+	List(ctx context.Context, path string) (fsbrowsersvc.Listing, error)
+}
 
 // FSController owns the read-only /fs routes used by remote clients to browse
 // for a project path. Listing is directories-only and dotfile-free by design:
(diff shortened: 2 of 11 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/httpd/controllers/fs.go
```diff
@@ -0,0 +1,86 @@
...
+	// The daemon judges the path by its own OS's rules — the client must never
+	// pre-judge it, since a remote picker has no idea what a valid absolute path
+	// looks like on the host it is browsing.
+	if !filepath.IsAbs(dir) {
+		envelope.WriteError(w, r, apierr.Invalid("FS_PATH_NOT_ABSOLUTE", "path must be absolute on the daemon host", nil))
+		return
+	}
+	dir = filepath.Clean(dir)
+
+	entries, err := os.ReadDir(dir)
```
illegalcall: This controller is doing the filesystem work itself. In this repo, controllers normally decode the request, call an injected service, and encode the response. Can we move `UserHomeDir`, `ReadDir`, and `Stat` behind a small directory-browser service?

backend/internal/httpd/controllers/fs_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if body.Parent != filepath.Dir(home) {
+		t.Errorf("parent = %q, want %q", body.Parent, filepath.Dir(home))
+	}
+}
+
+func TestListDirsRejectsRelativeAndMissingPaths(t *testing.T) {
+	home := t.TempDir()
+	srv := newFSRig(t, home)
+
+	if status, _ := listDirs(t, srv, "relative/path"); status != http.StatusBadRequest {
```
illegalcall: These tests only check status codes. Please also assert `FS_PATH_NOT_ABSOLUTE`, `FS_NOT_FOUND`, `FS_NOT_A_DIRECTORY`, and `FS_FORBIDDEN`, and add the permission-denied case.

backend/internal/httpd/controllers/dto.go
```diff
@@ -1206,6 +1206,26 @@ type ImportRunResponse struct {
...
 
+// ListDirsQuery is the query string accepted by GET /api/v1/fs/dirs.
+type ListDirsQuery struct {
+	Path string `query:"path,omitempty" description:"Absolute directory on the daemon host to list. When omitted, the daemon user's home directory."`
+}
+
+// FSEntry is one directory in a /api/v1/fs/dirs listing.
+type FSEntry struct {
+	Name    string `json:"name" description:"Directory name."`
+	Path    string `json:"path" description:"Absolute path of the directory on the daemon host."`
```
illegalcall, line 1323: The RFC says `names only`, but the response also exposes absolute paths, parent, and `gitRepo`. The extra fields seem useful. Can we update the RFC or ADR wording so it matches this contract?

backend/internal/httpd/controllers/fs.go
AronPerez: Moved the host-filesystem work into the new `service/fsbrowser.Service`. `FSController` now only passes the decoded path and request context to an injected `DirectoryBrowserService`, maps the service result to the response DTO, and encodes it. Production wiring supplies the service through `APIDeps`.

backend/internal/httpd/controllers/fs_test.go
AronPerez: The controller tests now decode the locked API error envelope and assert `FS_PATH_NOT_ABSOLUTE`, `FS_NOT_FOUND`, `FS_NOT_A_DIRECTORY`, and `FS_FORBIDDEN` alongside their statuses. I also added a permission-denied directory case using Unix mode bits (skipped on Windows, where those bits are not enforced) and assert its 403/`FS_FORBIDDEN` response.

backend/internal/httpd/controllers/dto.go
AronPerez: Kept the DTO contract unchanged. I updated ADR 0001 to describe the complete authenticated directory-browser response: the listed absolute path and parent, plus each visible subdirectory’s name, absolute path, and `.git` presence. It also states that contents, sizes, timestamps, files, and dotted names are not exposed.

illegalcall (approved):

PR #4360, opened 2026-08-24
docs(remote-hosts): setup, trust boundary, ADR

AronPerez:
## Ticket

## Problem

The remote-hosts feature has no user-facing documentation and no written trust boundary. Someone turning the flag on has to infer from the UI what a saved host is, what credential it uses, where that credential lives, and what the plaintext-HTTP LAN listener does and does not protect. Undocumented security properties get assumed rather than checked.

## Solution

Setup documentation for connecting a remote host, a written statement of the trust boundary (what the connection password authorises, why the loopback proxy exists, what stays on the machine), and an ADR recording the design decision. Documentation only — no product code, so nothing here changes behaviour with the flag on or off.

## How Has This Been Tested?

Documentation only: five files, no code. `git diff --stat` against the current `main` (`c9a0adb2`) shows changes confined to `docs/`, `AGENTS.md` and three landing-site `.mdx` pages; no test, no build target and no CI job covers behaviour here. Prose was proofread against the shipped behaviour of the branches it describes rather than against the plan documents.

No renderable surface: documentation and one `AGENTS.md` paragraph. The landing pages render through the existing docs pipeline unchanged.

## Implementation notes

This PR describes behaviour that only exists once the rest of the series has merged, so it is deliberately last in the order even though it imports nothing and could technically be reviewed at any point.

Content: how to save a host and connect to it; what the connection password authorises and what it does not; why the desktop app proxies remote traffic through a loopback listener rather than letting the renderer authenticate directly; and the ADR that records the decision so the next person does not re-litigate it from scratch.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
```diff
@@ -0,0 +1,232 @@
...
+### Enable the listener on the remote machine
+
+Open **Settings → Connect Mobile** on the remote machine and enable mobile access. Note the displayed address, port, and connection password. Use the port actually shown — the listener falls back to another port when `3011` is taken.
+
+</Step>
+<Step>
+
+### Open the host selector
+
+On this machine, start **Add project** and open the host selector at the top of the dialog.
```
illegalcall, line 37: Remote hosts are off by default. Add a step to enable `Remote hosts (experimental)` before opening Add project. Otherwise this selector will not be there.

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
```diff
@@ -0,0 +1,232 @@
...
+Select **Connect**. AO validates that the address really is an AO daemon and that the password is accepted before saving it, so a mistyped port cannot be stored as a working host.
+
+</Step>
+</Steps>
+
+Saved hosts are written to `~/.ao/remotes.json` at mode `0600` — the same file the `ao` CLI reads for `--url` targets. AO refuses to read it if it is group- or other-readable.
+
+Once a host is connected, its projects appear as their own section in the sidebar, and **Add project** can browse directories on that machine. Removing a host only deletes your saved entry: sessions on that machine keep running.
+
+Failure states are reported separately and are all recoverable in place. "Wrong password" is never conflated with "could not reach", an unreachable host fails on its own without blanking the rest of the tree, and an unreachable host at startup does not delay or block the app.
```
illegalcall: This does not match the current cumulative branch. A newly added host is not connected immediately, and a host that fails during startup is omitted instead of appearing with Retry. Please fix those flows before this text lands, or document the limitation.

docs/adr/0003-remote-hosts-renderer-fanout.md
```diff
@@ -0,0 +1,171 @@
...
+**Fan out in the renderer, over a per-host authenticated loopback proxy in the main
+process.** No daemon ever talks to another daemon.
+
+1. **Local is a host.** `LOCAL_HOST` is a host id like any other — the one whose
+   requests skip the proxy. No code path asks "is this remote?".
+
+2. **Everything addressable is a `Ref = {host, id}`.** Ids are never rewritten; they
+   are qualified only where an action is routed. Every read and every write takes a
+   `Ref` and dispatches through `clientFor(ref.host)`, so acting on the wrong machine
+   is a type error rather than a runtime accident. Routes are host-qualified —
```
illegalcall: `HostId` is a plain string and some paths still accept bare IDs. Editor handoff is one example: it sends only `sessionId`, then main resolves it against the local daemon. Wrong-host routing is not a type error yet. Please enforce a host-qualified ref here or describe this as the target design.

docs/adr/0003-remote-hosts-renderer-fanout.md
```diff
@@ -0,0 +1,171 @@
...
+   mux URL is derived from that host's own base — keeping the path prefix, which is
+   what carries the token.
+
+5. **Failure is data, not an exception.** Each host carries its own status. A failed
+   host renders as a labelled section with a retry; it never blanks the tree and
+   never discards another host's data. Hosts connect **after** first paint and probes
+   are bounded, so a sleeping host cannot delay startup.
+
+6. **Credential containment.** Only `{label, url}` and the loopback proxy base ever
+   cross to the renderer. The connection password never does.
```
illegalcall: The password typed in `AddRemoteHostDialog` exists in renderer state and crosses IPC. The actual guarantee is that saved passwords are never sent back to the renderer. Please use that wording.

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
```diff
@@ -0,0 +1,232 @@
...
+<Step>
+
+### Set the bind mode
+
+On the remote machine, turn Connect Mobile **off**, then edit `~/.ao/mobile/config.json` and set `bind`:
+
+```json
+{
+  "enabled": false,
+  "bind": "tailscale"
```
illegalcall: I cannot find support for `bind` in upstream main or this PR series. `mobilebridge.State` has no `bind` field, so this can be ignored while the user believes the listener is Tailscale-only. Please land the bind implementation first or remove this section.

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
AronPerez: Added a setup step before Add project: enable **Settings → General → Remote hosts (experimental)**. It now states that the flag is off by default and the selector stays hidden until it is enabled.

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
AronPerez: Verified the cumulative renderer flow: the add callback refreshes the saved-host probe list and selects the URL, but does not call `connectHost`; startup uses `Promise.allSettled`, and failed connections never enter `connectedHosts()`, which is the source for workspace fan-out. The guide and ADR now document that newly saved and startup-failed hosts are omitted from the sidebar without Retry, plus the current restart/flag-toggle workaround.

docs/adr/0003-remote-hosts-renderer-fanout.md
AronPerez: Verified that `HostId` is currently a string alias and editor handoff sends only `sessionId`; main resolves that ID through the local loopback daemon. The ADR now presents host-qualified `Ref` routing as the target rule and explicitly records the bare-ID editor-handoff gap and runtime wrong-host risk.

docs/adr/0003-remote-hosts-renderer-fanout.md
AronPerez: Verified that AddRemoteHostDialog keeps the typed password in renderer state and sends it over add/update IPC. Reworded the ADR precisely: user-typed credentials cross renderer → main, while saved passwords are never sent back to the renderer; the security-review wording now states the same guarantee.

frontend/src/landing/content/docs/configuration/remote-sessions.mdx
AronPerez: Confirmed that `mobilebridge.State` has no bind field and neither upstream/main nor the cumulative branch contains a `json:"bind"` setting. Removed the unsupported Tailscale-only bind section, the loopback bind step, the bind comparison, and the related limitation claim; the supported SSH tunnel remains.

i-trytoohard:

AronPerez:
understood, this feature will likely remain stale for the next month or two. I do not have the capacity to support this at the moment and would ask it get closed or picked up

PR #4362, opened 2026-08-24
fix(cli): judge a remote --path and a remote project by the remote host's rules

AronPerez:
## Ticket

## Problem

With `--url`, the daemon resolves `--path` against its own filesystem, and every implicit project signal — `AO_PROJECT_ID`, `AO_SESSION_ID`, the current directory — describes the machine running the CLI, not the remote one. Each is silently matched against the wrong host's projects, with output that looks like success, and a PR-ref lookup can fall back to running `gh` against a local checkout while addressing a remote daemon.

## Solution

Host-relative paths are refused for a remote target; absolute paths (including Windows drive-absolute and UNC forms) are judged for any host rather than the OS running the CLI. Implicit-project matching requires an explicit `--project` for a remote target. PR-ref resolution refuses its local-`gh` fallback and asks for a full URL instead. `project ls` gains a `PATH` column so "where does this project live" has an answer once the daemon isn't this machine.

## How Has This Been Tested?

`cd backend && go build ./... && go vet ./... && go test ./... && go test -race ./internal/cli/` on the current `main` (`c9a0adb2`): 158 packages ok, `gofmt -l` clean. Eight new tests cover path refusal across every host-relative/absolute form, implicit-project refusal with the explicit-`--project` escape, `ao spawn --url` refusing before any request, and four `resolvePRRef` cases.

### DIFF
--- a/backend/internal/cli/client.go
+++ b/backend/internal/cli/client.go
@@ -91,7 +91,37 @@ func (c *commandContext) doJSON(ctx context.Context, method, path string, body, 
 }
 
 func (c *commandContext) postLoopbackJSON(ctx context.Context, path string, body any) error {
+	// These are loopback-only control routes (/internal/*), 404'd at the LAN
+	// socket by design — and CLI telemetry has no business reaching a daemon on
+	// someone else's machine. Drop them when a remote target is set.
+	if c.remote != nil {
+		return nil
+	}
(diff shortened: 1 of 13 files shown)


i-trytoohard:

codebanditssss:

PR #4373, opened 2026-08-24
refactor(hosts): route session and terminal writes by Ref

AronPerez:
## Ticket

No upstream issue. [RFC](https://github.com/AronPerez/agent-orchestrator/blob/plan/2026-08-24-wave3/docs/upstreaming-rfc-remote-hosts.md). Stacked on #4372.

## Problem

Session and terminal writes still go through `apiClient`, addressed by a bare id. Two hosts holding sessions that share a name — both call it `agent-orchestrator-1`, since a project id is `filepath.Base(path)` on every machine — have no way to route a rename, kill or agent-switch to the right one.

## Solution

Session and terminal writes take a `Ref = {host, id}` and dispatch through `clientFor(ref.host)`: rename, terminate, pin, restore, switch-agent, interface transition, conversation, and the browser link. Mutation state and conversation caches move to `refKey` too, so a "Killing…" indicator cannot appear on the wrong machine's row.

## How Has This Been Tested?

`cd frontend && npm run typecheck && npm run typecheck:e2e && npx vitest run src/renderer` on the current `main` (`c9a0adb2`): 142 files / 2045 tests, all green. The five new tests pin two hosts holding a same-named session and assert zero cross-host requests; pointing a write at the wrong client fails exactly one case.

Before: local and remote hosts each holding a session named scratch-session-1
After: only the local session renamed to renamed-this-mac; the remote one is untouched

### DIFF
--- a/frontend/e2e/support/fake-bridge.ts
+++ b/frontend/e2e/support/fake-bridge.ts
@@ -248,6 +248,14 @@ export async function installFakeBridge(page: Page, opts: FakeBridgeOptions = {}
 					list: async () => [],
 					getActive: async () => null,
 				},
+				remotes: {
+					list: async () => [],
+					add: async () => "offline" as const,
+					update: async () => "offline" as const,
+					remove: async () => undefined,
+					probe: async () => "offline" as const,
+					request: async () => ({ status: 0, body: null }),
+				},
 				cloud: {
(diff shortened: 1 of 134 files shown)


i-trytoohard:

codebanditssss:

PR #4374, opened 2026-08-24
feat(hosts): fan out workspace queries and event streams per host

AronPerez:
## Ticket

No upstream issue. [RFC](https://github.com/AronPerez/agent-orchestrator/blob/plan/2026-08-24-wave3/docs/upstreaming-rfc-remote-hosts.md). Stacked on #4372.

## Problem

The board only ever reads one host. With more than one connected, a sleeping host must not stall the others, a busy remote must not make the local board refetch, and a host connected after first paint must not stay invisible — none of that is true yet because there is one query and one event stream, not one per host.

## Solution

`useWorkspaceQuery` becomes one query per host under `["workspaces", host]`, combined into one board section per host; each host gets its own `EventSource` at its own base, and an event from host B invalidates only host B's key. A failed host reports its last good workspaces with `status: "failed"` rather than throwing.

## How Has This Been Tested?

`cd frontend && npm run typecheck && npm run typecheck:e2e && npx vitest run src/renderer` on the current `main` (`c9a0adb2`): 143 files / 2063 tests, all green (base was 141/2040). New: `lib/host-events.test.ts` and a rewritten `useWorkspaceQuery.test.tsx` driving the fan-out against hostile-daemon shapes.

Board showing both hosts' sections fanned out side by side, each with its own sessions

### DIFF
--- a/frontend/e2e/support/fake-bridge.ts
+++ b/frontend/e2e/support/fake-bridge.ts
@@ -248,6 +248,14 @@ export async function installFakeBridge(page: Page, opts: FakeBridgeOptions = {}
 					list: async () => [],
 					getActive: async () => null,
 				},
+				remotes: {
+					list: async () => [],
+					add: async () => "offline" as const,
+					update: async () => "offline" as const,
+					remove: async () => undefined,
+					probe: async () => "offline" as const,
+					request: async () => ({ status: 0, body: null }),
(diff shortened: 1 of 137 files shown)


i-trytoohard:

codebanditssss:

PR #4377, opened 2026-08-24
refactor(hosts): route pull request and review writes by Ref

AronPerez:
## Ticket

No upstream issue. [RFC](https://github.com/AronPerez/agent-orchestrator/blob/plan/2026-08-24-wave3/docs/upstreaming-rfc-remote-hosts.md). Stacked on #4375.

## Problem

Pull request, review, diff and panel writes are the last surface still on `apiClient` by bare id. Two hosts' sessions sharing an id must not share an open inspector, a selected tab, or a browser-unseen badge — the same collision the session and project writes already fixed at the request layer, still open at the UI layer.

## Solution

PR/review actions, diff-selection "send to agent" calls, the palette's review trigger, and the browser panel's annotation queue take a `Ref` and dispatch through `clientFor`. `inspectorSessions` and `visibleTerminalKindBySession` move to `refKey`. One bug fixed in passing: `useBrowserAnnotationQueue` now resets on session identity, not a fresh prop object.

## How Has This Been Tested?

`cd frontend && npm run typecheck && npm run typecheck:e2e && npx vitest run src/renderer` on the current `main` (`c9a0adb2`): 142 files / 2052 tests. Existing suites are the regression check; changed tests assert a call shape or a store key.

**Update:** QA against two real daemons found the claim above didn't fully hold — `SessionView.tsx` still resolved "the open session" by bare id, a gap in A8a's read conversion, which this PR's `inspectorSessions`/`refKey` keying inherited. Fixed on A8a (RED confirmed the collision, GREEN after scoping to `sessionRef.host`); the test count above includes it.

## Artifacts (if appropriate):

Local demo-project-2, inspector open on Files
Remote demo-project-2 (same id, other host), inspector on its own default then closed
Back on local demo-project-2, inspector still open on Files, unaffected

Both hosts hold a session with the same id; open state and selected tab stay host-scoped across all three frames. One unrelated leak is visible in these frames, not a falsification of this PR's claim: the sidebar's active-row highlight (`Sidebar.tsx:379,917`) is keyed by bare session id and lights up both hosts' same-id rows — a separate defect, same class, not fixed here.

*Captured on a dev build; `--disable-web-security` bridges the dev-origin CORS gap only (production origin `app://renderer` passes the same check natively) — the daemon, proxy, credential, and traffic are all real.*

PR #4370, opened 2026-08-24
feat(hosts): add, edit and remove remote hosts

AronPerez:
## Ticket

No upstream issue. [RFC](https://github.com/AronPerez/agent-orchestrator/blob/plan/2026-08-24-wave3/docs/upstreaming-rfc-remote-hosts.md). Stacked on #4369.

## Problem

With the flag, primitives, store, proxy and clients all merged, a user still has no way to reach any of it: there is no UI for adding, editing or removing a saved host, and no way to point a new project at a machine other than the one the app is running on.

## Solution

A Host dropdown in Add-a-project lists saved remote daemons beside This Mac with live reachability, and manages them in place — add, edit, remove — with the connection password never leaving the main process. Picking a remote host replaces the native folder dialog with an absolute-path field and registers the project against that daemon over REST; the daemon owns the verdict on its own filesystem.

## How Has This Been Tested?

`cd frontend && npm run typecheck && npm run typecheck:e2e && npx vitest run src/renderer/hooks/useRemoteHosts.test.tsx src/renderer/components/HostSelect.test.tsx src/renderer/components/AddRemoteHostDialog.test.tsx src/renderer/components/CreateProjectFlowHosts.test.tsx src/renderer/components/CreateProjectFlow.test.tsx src/renderer/test/fake-daemon.test.ts` on the current `main` (`c9a0adb2`): green, 7 files / 72 tests, including the flag-off regression check. Plus `npm --prefix packages/product-ui test`: 10/64 green.

## Artifacts (if appropriate):

Add-a-project Host dropdown listing "This Mac" and a saved, connected remote host

*Captured on a dev build; `--disable-web-security` bridges the dev-origin CORS gap only (production origin `app://renderer` passes the same check natively) — the daemon, proxy, credential, and traffic are all real.*

PR #4378, opened 2026-08-24
feat(hosts): one tree across every connected host

AronPerez:
## Ticket

No upstream issue. [RFC](https://github.com/AronPerez/agent-orchestrator/blob/plan/2026-08-24-wave3/docs/upstreaming-rfc-remote-hosts.md). Stacked on #4370, #4376, #4377 — cumulative diff over the merge of all three until they land.

## Problem

With multiple hosts connected, the sidebar still shows one machine's projects, and a failed host has no visible state: a failed remote silently blanks the tree, a failed *local* host is hidden the same way, and a hostile or outdated daemon answering with an HTML page or wrong-shape JSON can throw into the renderer.

## Solution

The sidebar draws every connected host as its own section; a host switcher filters the view once there is more than one, and is a view filter only — it never reconnects a host or changes where an action is sent. Each of the three failure modes above renders as a labelled, retryable section instead of blanking or throwing. Three telemetry events cover the ways the remote path goes dark, with `host_id` hashed before it leaves the machine.

## How Has This Been Tested?

`cd frontend && npm run typecheck && npm run typecheck:e2e && npx vitest run src/renderer` on the current `main` (`c9a0adb2`): 147 files / 2109 tests, all green (integration base was 2107). New: a failed host keeps the other's rows; a failed local host is reported as loudly as remote; a dropped stream announces itself without accusing the healthy host; all eight locales gain the same fourteen keys.

Host switcher dropdown open, listing All hosts / This Mac / demo-remote


i-trytoohard:
@AronPerez Closing as stale: no activity from you for 25 days. The branch stays safe to reopen if you pick this up again.

codebanditssss:

PR #4387, opened 2026-08-24
fix(topbar): surface the real reason when opening an editor fails

Pulkit7070:
## Summary

Opening a session whose worktree is gone (terminated, merged, cleaned up) put
Electron's IPC plumbing straight into the topbar actions row:

> Error invoking remote method 'editorHandoff:open': Error: Session workspace is not available

Reproduced against real local data, where 27 of the first 40 sessions have no
resolvable workspace (`/api/v1/desktop/sessions/{id}/workspace` returns 404).

## What was wrong

Two separate things, and the guard that exists is not the problem.

`TopbarOpenEditorButton` already disables itself when `getState` reports
`workspaceAvailable: false`, and that works. Confirmed live: on a terminated
session the control is disabled with the tooltip "Session workspace is not
available".

What bites is staleness. `useEditorHandoffState` caches with
`staleTime: 10_000` and nothing invalidated it on failure, so a worktree that
disappears *after* the state was read leaves the button enabled. The click
fails, the state is never refreshed, and the button stays enabled to fail again.

Separately, the message shown was raw plumbing.
`TopbarOpenEditorButton.tsx` rendered `open.error.message` verbatim, which is
Electron's wrapper around anything an `ipcMain` handler throws.

## The fix

Renderer-side, in `useEditorHandoff.ts`:

- Strip the `Error invoking remote method '<channel>': Error:` wrapper and
  rethrow a clean error, so the topbar shows what the main process actually
  said. This covers the other failure paths too, which were equally buried:
  "That editor is not installed", "Could not open <name>".
- Invalidate the handoff state query on error so the control re-checks
  availability rather than inviting the same failing click.

## Verification

Reproduced and fixed in the running desktop app against real session data, not
just in tests. Before: the raw wrapper. After: the disabled control carries the
clean reason as its tooltip.

Five unit tests on the normalizer plus a component regression test that clicks
through and asserts the rendered alert. The regression test is not vacuous:
with both changes reverted it fails with `expected 'Error invoking remote
method...' not to contain 'Error invoking remote method'`.

Full frontend suite: 2867 tests across 217 files. Typecheck clean.

## Notes for review

The code being fixed came from commit 728400ca8 in #3284, which replaced an
earlier HTTP implementation with this IPC one. I did not write it, so a look
from its author is worth having.

This treats the symptom in the renderer. Arguably `open()` in
`frontend/src/main/editor-handoff.ts` should also resolve the workspace inside
its `try`/`catch` so the failure comes back as a friendly message at the source
rather than as a raw throw. Left alone to keep this change narrow; happy to
fold it in.

### DIFF
--- a/frontend/src/renderer/components/TopbarOpenEditorButton.test.tsx
+++ b/frontend/src/renderer/components/TopbarOpenEditorButton.test.tsx
@@ -40,6 +40,24 @@ function renderButton() {
 }
 
 describe("TopbarOpenEditorButton", () => {
+	// Regression: an ipcMain rejection arrives wrapped as "Error invoking remote
+	// method 'editorHandoff:open': Error: <reason>", and the topbar used to paint
+	// that whole string into the actions row.
+	it("shows the reason, not Electron's remote-method wrapper, when the open fails", async () => {
+		setState(availableState);
+		window.ao!.editorHandoff.open = vi.fn().mockRejectedValue(
+			new Error("Error invoking remote method 'editorHandoff:open': Error: Session workspace is not available"),
+		);
+		renderButton();
+
+		await userEvent.click(await screen.findByRole("button", { name: "Open in Cursor" }));
+
+		const alert = await screen.findByRole("alert");
+		expect(alert).toHaveTextContent("Session workspace is not available");
+		expect(alert.textContent).not.toContain("Error invoking remote method");
+	});
+
+
 	beforeEach(() => {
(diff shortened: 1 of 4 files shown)


Vaibhaav-Tiwari (approved):
no blockers as such just

  two minor issues:

  - The main fix—refreshing stale editor state—isn’t directly tested.
  - The error message is cleaned twice unnecessarily.

Issue #1321, opened 2026-04-19
RFC: Remove polling and move to reactive sync for sessions, file-trees and file content


i-trytoohard:
## Summary

Move the dashboard away from periodic polling (5s SSE interval) to a reactive, push-based sync model for:

1. **Session state** — status, branches, PRs, CI results
2. **File trees** — workspace directory structures
3. **File content** — individual file reads/edits

## Context

Currently the web dashboard polls core every ~5 seconds via SSE, which means:
- Stale data between intervals
- Unnecessary load when nothing changes
- Latency in reflecting agent actions (CI results, review comments, etc.)

A reactive model would push updates only when state actually changes, giving instant feedback and reducing overhead.

## Discussion Points

- (From thread discussion — Dhruv has comments/quesions to address here)
- What transport? (WebSocket events, SSE with selective push, etc.)
- Which state changes should trigger pushes?
- How to handle reconnection / missed events?
- Impact on existing lifecycle-manager polling loop

## Open Questions

- Scope: full reactive rewrite or incremental (sessions first, then files)?
- Backward compat for CLI-only users?
- Performance benchmarks: current polling vs proposed reactive?

---

*Created from Discord discussion in #architecture-and-design*

i-trytoohard:
## Investigation: Lifecycle manager event-loop blocking

The lifecycle manager runs a continuous polling loop (determineStatus on each session). We should investigate:

- **Is it hogging the event loop?** If the polling loop does synchronous or CPU-heavy work (e.g., spawning processes, reading large outputs), it could block the Node.js event loop and starve incoming client requests.
- **What's the actual request latency impact?** Profile a typical SSE/dashboard request during a heavy polling cycle vs. idle.
- **Is the polling interval too aggressive?** If it's running every N seconds across all active sessions, the cumulative cost could be significant at scale.

This is directly relevant to the reactive sync proposal — if the lifecycle manager is the bottleneck, moving to push-based updates alone won't fix client-facing latency. We may need to restructure how lifecycle checks are scheduled (e.g., worker threads, separate process, or per-session debounced checks).

i-trytoohard:
## Investigation: Session detail loading is costly — should be instant

When a user taps a session in the sidebar, they see a loading screen before the session detail renders. This shouldn't happen — most session data (status, branch, PR, agent, created_at) is already available from the sessions list that populated the sidebar.

**Root cause questions:**
[...]

PR #4388, opened 2026-08-24
feat(frontend): make workspace file sync reactive

Vaibhaav-Tiwari:
## What

Make the workspace file viewer use daemon SSE invalidations as its normal refresh path, with 30-second query polling retained only as a degraded fallback.

## Why

This implements the focused file-viewer portion of #1321 without a broader architecture rewrite. It removes periodic requests while the existing workspace event stream is healthy and keeps recovery behavior when streaming is unavailable.

Refs #1321.

## How

- Keep one reference-counted EventSource per session across inspector and maximized file views.
- Serialize connection attempts, coalesce retry triggers, add 4-6 second jittered retries, and ignore stale connection callbacks.
- Track connecting, connected, and degraded health; degrade after three terminal failures or when EventSource is unavailable.
- Disable workspace-file polling while connecting or connected, and enable a 30-second fallback only while degraded.

Intentional omission: this PR does not change session, PR, CI, file-content, daemon, or lifecycle synchronization.

## Testing

- npm test -- src/renderer/lib/workspace-file-events.test.ts src/renderer/hooks/useSessionWorkspaceFiles.test.ts (7 passed)
- git diff --check
- Full frontend run: 2,787 passed; 31 unrelated Windows/macOS, Unix-socket, landing dependency, and path-separator tests failed.
- npm run typecheck is blocked by unresolved React and utility dependencies in the untouched packages/product-ui workspace.

## Checklist

- [x] Branched from main
- [ ] Relevant CI checks pass for the area touched (targeted tests pass; workspace-wide limitations documented above)

### DIFF
--- a/frontend/src/renderer/components/SessionFilesView.tsx
+++ b/frontend/src/renderer/components/SessionFilesView.tsx
@@ -35,6 +35,8 @@ import { apiClient, apiErrorMessage } from "../lib/api-client";
 import {
 	isChangedWorkspaceFile,
 	sessionWorkspaceFilesQueryOptions,
+	useWorkspaceFileConnectionState,
+	workspaceFilesRefetchInterval,
 	type WorkspaceCompareMode,
 	type WorkspaceFileSummary,
 } from "../hooks/useSessionWorkspaceFiles";
@@ -146,7 +148,11 @@ export function SessionFilesView({
 	const annotationSentTimerRef = useRef<number | null>(null);
 	const rootRef = useRef<HTMLElement>(null);
 
-	const filesQuery = useQuery(sessionWorkspaceFilesQueryOptions(sessionId, t("files.error.loadWorkspace")));
+	const workspaceConnectionState = useWorkspaceFileConnectionState(sessionId);
+	const filesQuery = useQuery({
+		...sessionWorkspaceFilesQueryOptions(sessionId, t("files.error.loadWorkspace")),
+		refetchInterval: workspaceFilesRefetchInterval(workspaceConnectionState),
+	});
 	useEffect(() => subscribeWorkspaceFileChanges(sessionId, queryClient), [queryClient, sessionId]);
 	const files = filesQuery.data?.files ?? emptyFiles;
 	const changedFiles = useMemo(() => files.filter(isChangedWorkspaceFile), [files]);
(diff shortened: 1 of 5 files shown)


Pulkit7070 (commented):
This is a clean, well-tested move to reactive refresh. The generation counter that ignores stale connection callbacks, the coalesced jittered retries, the ref-counted single `EventSource` per session, and the base-URL-change reset are all handled carefully, and the state machine reads clearly. The scope is right: `sessionWorkspaceFilesQueryOptions` is consumed only by `SessionFilesView`, and `useSessionWorkspaceFilesChangedCount` already overrode `refetchInterval` to `false`, so dropping the baseline 30s poll from the shared options strands nothing and the inspector badge stays SSE-driven exactly as before.

I verified the reactive path against the daemon rather than assuming it: `GET /api/v1/sessions/{sessionId}/workspace/events` is served by `streamWorkspaceChanges`, which emits `event: workspace_changed` on every real filesystem change off `workspacewatch.Watch`, plus a 15s `: keepalive` comment. That is exactly the event the client listens for, so connect to `onopen` invalidation, `workspace_changed` to invalidation, and reconnect to re-invalidation all line up. Removing the periodic poll is sound because the stream it leans on genuinely works.

One minor edge in the degrade path, not a blocker: only a terminal failure (`readyState === CLOSED`) counts toward the three-strike degrade, and a missing `EventSource` degrades immediately. A connection stuck in the browser's native reconnect (`readyState === CONNECTING`, for example the daemon briefly unreachable) reports `connecting` indefinitely and never escalates to `degraded`, so the 30s fallback never kicks in during a prolonged reconnect. That is usually fine, since the REST endpoint would be unavailable in the same window, but it means `degraded` is reached only via CLOSED failures or an absent `EventSource`, never via a long reconnect.

The other known consequence, dropping the unconditional poll as a reconcile net for a missed `workspace_changed` edge, is the intended tradeoff of this change; reconnects re-invalidate and the keepalive keeps an idle stream healthy, so it is low risk.

Two notes, not blockers: the diff reformats a couple of unrelated lines (the URL assertion and the invalidate calls), which is harmless but unrelated churn. And the targeted tests pass; `useSessionWorkspaceFiles.test.ts` transform-failed only in my isolated checkout due to an unrelated i18n/Vite resolution issue, not a real failure.

Vaibhaav-Tiwari:
Addressed the prolonged native-reconnect edge in a79337d53. Nonterminal EventSource errors now count toward the existing three-strike health threshold without replacing the browser-managed reconnect path. After three failures the file query enables the 30-second degraded fallback; a successful onopen resets the count and returns to SSE-only refresh. The regression test covers connected -> connecting -> degraded -> connected. Also removed the unrelated formatting churn noted in the review. Focused tests: 7/7 passed.

Pulkit7070 (approved):

PR #4389, opened 2026-08-24
fix(acp): accept stdio MCP servers without an optional MCP capability

### LINKED ISSUE #4380: [Bug]: ACP driver rejects stdio MCP servers unless the agent advertises an optional MCP capability
## Summary

The ACP chat driver refuses **all** per-session MCP servers — including **stdio**, which the ACP spec says every agent must support — unless the agent advertises one of the optional `acp` / `http` / `sse` MCP capabilities. A spec-conformant stdio-only agent that reports `mcpCapabilities: {}` therefore fails session start outright the moment the user has any MCP server configured, on both start and resume.

## Steps to reproduce

1. Use an ACP-based chat agent whose `initialize` response advertises no optional MCP capability (i.e. `mcpCapabilities: {}` — the normal case for an agent that only speaks the mandatory stdio transport).
2. Configure any per-session MCP server of type `stdio` (or with `type` omitted, which AO treats as stdio).
3. Start a chat session with that agent — or resume one.

The gate is reachable from both paths: `normalizeMCPServers` is called at `backend/internal/adapters/chatdriver/acp/driver.go:158` (start) and `:242` (resume).

Reproducible directly against the function, no agent required:

```go
normalizeMCPServers(
    []ports.ChatMCPServerConfig{{Name: "fs", Type: "stdio", Command: "npx", Args: []string{"-y", "@modelcontextprotocol/server-filesystem", "/tmp"}}},
    acpsdk.McpCapabilities{},
)
// servers=[]  err=ACP agent does not support per-session MCP servers
```

The repository's own test already documents the behaviour: `TestNormalizeMCPServersFailsWithoutCapabilities` (`driver_test.go:1766`) passes a **stdio** server with `acpsdk.McpCapabilities{}` and asserts that it errors.

## Expected behavior

A stdio MCP server is accepted regardless of `mcpCapabilities`, because stdio is the baseline transport every agent is required to support. Only the non-baseline transports (`http`, `sse`, `acp`) should require their corresponding capability.

## Actual behavior

Session start/resume fails with `ACP agent does not support per-session MCP servers`, and the user's MCP servers never reach the agent.

## AO version

upstream `main` @ `c9a0adb21`

## Environment

Arch Linux x86_64, Go 1.25.12. Not platform-specific — this is pure protocol logic.

## Logs / extra context

`backend/internal/adapters/chatdriver/acp/session_setup.go:41`:

```go
func normalizeMCPServers(configs []ports.ChatMCPServerConfig, caps acpsdk.McpCapabilities) ([]acpsdk.McpServer, error) {
	if len(configs) == 0 {
		return []acpsdk.McpServer{}, nil
	}
	if !caps.Acp && !caps.Http && !caps.Sse {
		return nil, fmt.Errorf("ACP agent does not support per-session MCP servers")
	}
	...
```

Three things suggest this gate was intended as a coarse "does this agent do MCP at all" proxy rather than a real capability check:

1. **The spec says stdio is mandatory.** The vendored SDK carries the schema text — `acp-go-sdk@v0.13.5/types_gen.go:2718`, on `McpServer.Stdio`: *"All Agents MUST support this transport."* The `Acp` variant three lines above is documented as available *"Only when ... `mcp_capabilities.acp` is `true`"*, and `McpCapabilities.Acp` is additionally marked **UNSTABLE** ("not part of the spec yet, and may be removed or changed at any point"). So `acp`/`http`/`sse` are opt-ins for the non-baseline transports only.

2. **The per-transport switch below already disagrees with the gate.** `case "http"` checks `caps.Http`, `case "sse"` checks `caps.Sse`, and `case "", "stdio"` deliberately checks nothing — which is correct. The top-level gate is the only thing standing between a mandatory transport and the agent.

3. **`caps.Acp` has no other use in the function.** There is no `case "acp"` at all; an `acp`-typed config falls through to `default: unsupported ACP MCP transport`. So the gate consults a capability whose transport the code cannot construct.

The existing tests encode the same assumption, and one of them says so: `TestNormalizeMCPServersSucceedsWithHttpCapability` (`driver_test.go:1780`) passes a **stdio** server and succeeds only because the agent happened to advertise `Http: true`, with the comment *"any MCP capability is sufficient — the transport-specific check happens later."* That is the model this report is disputing, so fixing the behaviour means updating that test and inverting `TestNormalizeMCPServersFailsWithoutCapabilities`.

Suggested fix is to delete the top-level gate and let each transport keep gating itself; stdio then needs no capability, and `http`/`sse` remain rejected without theirs. I'd like to work on this and have the change and tests ready.

yashksaini-coder:
## What

Delete the blanket MCP capability gate in the ACP driver so **stdio** MCP servers are accepted regardless of `mcp_capabilities`, while `http` and `sse` stay gated on theirs.

## Why

Fixes #4380.

`normalizeMCPServers` refused **every** per-session MCP server unless the agent advertised one of `mcp_capabilities.acp` / `.http` / `.sse`. stdio is the baseline transport the ACP spec requires every agent to implement — the schema text on `McpServer.Stdio` in the vendored `acp-go-sdk@v0.13.5` (`types_gen.go:2718`) is **"All Agents MUST support this transport"** — so an agent advertising `mcpCapabilities: {}`, the normal shape for a stdio-only agent, was refused the one transport it is guaranteed to speak. Both entry points were affected: `Start` (`driver.go:158`) and `Resume` (`driver.go:242`).

**Severity, stated honestly: this is latent today, not user-visible.** No production caller populates `StartConfig.MCPServers` — `ChatMCPServerConfig{}` is constructed only in `driver_test.go`, and the sites in `internal/service/chat/{service,history}.go` merely forward `cfg.MCPServers`. So `normalizeMCPServers` currently receives an empty slice and returns at the `len(configs) == 0` early exit. I noted this on the issue as a correction to my own initial report.

The argument for landing it now is that it is a three-line deletion, and the MCP wiring work would otherwise be built on top of a gate that silently refuses the mandatory transport — with two existing tests asserting that refusal is correct, so it would read as intentional to whoever adds the wiring.

## How

Three things showed the gate was a coarse "does this agent do MCP at all" proxy rather than a real capability check:

- the transport switch immediately below already gates itself: `case "http"` checks `caps.Http`, `case "sse"` checks `caps.Sse`, and `case "", "stdio"` deliberately checks nothing — which is correct;
- **`caps.Acp` had no other use in the function.** There is no `case "acp"` at all, so an acp-typed config falls through to `default: unsupported ACP MCP transport`. The gate consulted a capability whose transport this code cannot construct, and which the SDK marks **UNSTABLE**;
- the existing test said so out loud: *"any MCP capability is sufficient — the transport-specific check happens later."*

**Docs:** `docs/harnesses/pi.md` documented a per-harness contract — that a session configured with per-session MCP servers is *rejected* during ACP setup rather than silently ignored — and that rejection had exactly one implementation: this gate. pi never had MCP-specific code; it inherited the refusal as a side effect. That row is corrected to describe what actually happens, and flags the resulting gap explicitly rather than papering over it. Deliberately **not** fixed here: giving pi a real per-harness refusal is a product decision and a feature, not part of a spec-conformance bugfix — happy to open a follow-up if you want that behaviour back.

## Testing

Tests were written **before** the fix and confirmed failing on unfixed source, then passing.

- `TestNormalizeMCPServersAllowsStdioWithoutCapabilities` — inverts the old `…FailsWithoutCapabilities`, which asserted the bug was intended behaviour.
- `TestNormalizeMCPServersDefaultTypeIsStdio` — the omitted-type spelling, also stdio.
- `TestNormalizeMCPServersStillGatesNonBaselineTransports` — 4-case table pinning the other half of the contract: http/sse refused without their capability **and** accepted with it, each asserting the correct `McpServer` union arm so a regression routing http into the `Stdio` arm cannot pass.
- `TestACPDriverSendsStdioMCPServersToSpecMinimalAgent` — driver-level: a fake agent advertising `AgentCapabilities{}` receives the stdio server through the real `Start` and `session/load` paths, so the fix is proven end to end rather than only at the helper.

The misleading `…SucceedsWithHttpCapability` case was replaced; its comment stated the model this change corrects.

```console
$ go test ./internal/adapters/chatdriver/...      # 15 packages
ok
$ go test -race ./internal/adapters/chatdriver/...
ok
$ go vet ./internal/adapters/chatdriver/acp/...
$ golangci-lint v2.12.2 run ./internal/adapters/chatdriver/acp/...
0 issues.
```

Full backend suite is green apart from two failures that reproduce identically on unmodified `main` on this machine and are unrelated to this change: `TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell` (the tmux base-index bug, #4222 / #4224) and `TestCopilotClassicPATIsNotAnAuthorizationSignal` (fails whenever a classic GitHub PAT is present in the environment).

## Checklist

- [x] Branched from `main` (based on `c9a0adb21`)
- [x] Relevant CI checks pass for the area touched (`go test`, `-race`, `go vet`, `golangci-lint`)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

yashksaini-coder:
Rebased onto `0b7758c7f`. Attaching a recording of the local verification.

**End-to-end test on my side, just now:** `go test -run MCPServers ./internal/adapters/chatdriver/acp/` → all green, including `TestACPDriverSendsStdioMCPServersToSpecMinimalAgent`, which drives a real `Driver` against a fake agent advertising `AgentCapabilities{}` and asserts the stdio server actually arrives over the connection on both `session/new` and `session/load`. Then I restored `main`'s gate while keeping the new tests, to confirm they fail for the right reason rather than passing vacuously. They do, with the error a user would actually see:

```
driver_test.go:1935: Start with a stdio MCP server against an agent advertising no MCP capability:
                     ACP agent does not support per-session MCP servers
driver_test.go:1953: Resume with a stdio MCP server against an agent advertising no MCP capability:
                     chat conversation resume failed: ACP agent does not support per-session MCP servers
```

Then restored and re-ran green. The recording walks that, starting from the schema line in the vendored SDK (`acp-go-sdk@v0.13.5`, `All Agents MUST support this transport` on `McpServer.Stdio`) and the three lines this PR removes. It ends on a clean worktree.

Full `chatdriver/...` tree green (15 packages), and the 8 CI checks here are green.

Restating the caveat from the issue so it is visible on the PR too: this is **latent, not user-visible today** — nothing populates `StartConfig.MCPServers` in production yet (`ChatMCPServerConfig{}` is constructed only in `driver_test.go`), so `normalizeMCPServers` currently returns at the `len(configs) == 0` early exit. The case for landing it now is that it is a three-line deletion, and the wiring work would otherwise be built on top of a gate that refuses the mandatory transport while two tests assert that refusal is correct.

https://github.com/user-attachments/assets/26a9974f-c466-48d5-91ab-7b7d7db92225

Issue #4284, opened 2026-08-23
Update feedback can remain loading and hide automatic progress


illegalcall:
## Bug

Desktop update feedback is misleading in two related situations:

1. A manual **Check for updates** can remain in its loading state after the main process has already published a terminal result. Navigating away and returning remounts the view and reveals the completed result.
2. Automatic checks and downloads produce no sidebar feedback until the update is fully downloaded, making a working automatic update look inactive and leaving users unsure whether they must update manually.

**Source:** Local Codex report with screenshots and a follow-up report from Prateek; no public chat URL  
**Observed:** macOS, AO `0.12.7-nightly.202608221348`  
**Analyzed against:** `858146266`  
**Confidence:** High  
**Priority:** Medium — update behavior is degraded, with manual checking as a workaround.

### Stuck manual loading

1. Open Settings → Updates.
2. Click **Check for updates**.
3. Let the updater publish a matching terminal status such as `not-available` while the IPC invocation remains unsettled.
4. The button and status panel remain in the checking state.
5. Navigate away and return; the completed status appears.

### Missing automatic progress

1. Seed or receive updater status `{ state: "available" }` or `{ state: "downloading" }`.
2. Observe the sidebar footer.
3. No update information is rendered until status becomes `downloaded`, at which point the existing **Restart to update** button appears.

## Root Cause

- `UpdatesSection` owns a renderer-local pending flag whose lifetime is coupled to the IPC promise rather than the updater status event. Electron can publish the terminal event before the invocation settles, so completed state is masked by stale local loading state.
- `Sidebar` explicitly renders its update row only for `status.state === "downloaded"`; `available` and `downloading` return `null`.

## Fix

- Correlate manual update checks with request IDs and clear local pending state when a matching non-checking status arrives, even if the IPC promise remains pending.
- Show informational, non-clickable sidebar states for available updates and download progress; preserve the existing restart/install button only for a staged download.

## Verification

- `cd frontend && npm test -- --run src/renderer/components/Sidebar.test.tsx src/renderer/components/GlobalSettingsForm.test.tsx src/main/auto-updater.test.ts`
- `cd frontend && npm run typecheck`

## Related

- [#2377](https://github.com/Untrivial-ai/agent-orchestrator/issues/2377) — original downloaded-only restart-row specification.
- [PR #4144](https://github.com/Untrivial-ai/agent-orchestrator/pull/4144) — update feedback redesign that introduced the renderer-local pending state.
[...]

PR #4392, opened 2026-08-24
fix(updates): prevent stuck manual check loader

illegalcall:
Regression follow-up to #4284. This replaces the update-card styling approach attempted in #4292; the two UI diffs should not be merged together.

## Summary

- consume every updater status push before React can batch away an intermediate terminal event
- clear a manual check only from its matching request ID, while preserving a previously staged update
- simplify the status card and keep long Nightly versions readable with aligned actions
- cover the real main-process event order, the renderer race, and the reported downloaded + Nightly geometry

## Root cause

With an update already staged, a failed manual manifest check can publish these statuses synchronously:

1. `error` for the current manual request ID
2. restored `downloaded` state carrying the earlier staged request ID

The fix in #4285 cleared local loading from a React effect that only observed the latest committed status. React batches both pushes, so the matching terminal event could be discarded before that effect ran. The local request ID stayed set and forced the card back to `checking` forever, even though the updater had restored a valid downloaded state.

`useUpdateStatus` now exposes a lossless per-event callback. `UpdatesSection` uses it to clear only the matching manual request before updating the rendered snapshot, so the following staged status renders normally.

The new renderer regression test failed against the old implementation with `Checking for updates…` where `Downloaded. Restart to finish updating.` was expected.

## Visual evidence

This deterministic renderer state matches the report: dark mode, automatic updates enabled, Nightly selected, a staged download, and the long installed Nightly version. The Playwright test also asserts that the version stays on one line and both actions have equal height.

Settings Updates after: downloaded Nightly state with readable version and aligned actions

The screenshot is visual-state evidence. The updater behavior itself is proven separately by the main-process producer test and renderer regression test.

## Native Electron verification

The branch was launched in the real Electron app with isolated AO data and the real preload bridge. Settings → Updates rendered correctly, and clicking **Check for updates** completed with `Updates are only available in the installed app.`, leaving the button enabled and `aria-busy=false`.

Unpackaged Electron intentionally has no release feed, so the exact packaged staged-update sequence is reproduced at its two real boundaries in tests rather than claimed from the screenshot.

## Tests

- `cd frontend && npm test -- --run` — 218 files passed; 2,875 tests passed; 1 skipped
- `cd frontend && npm test -- --run src/renderer/components/GlobalSettingsForm.test.tsx src/renderer/components/Sidebar.test.tsx src/main/auto-updater.test.ts` — 152 passed
- `cd frontend && npm run typecheck` — passed
- `cd frontend && npm run typecheck:e2e` — passed
- `cd frontend && npx playwright test e2e/updates-settings.spec.ts --reporter=line` — passed
- `cd frontend && npm run package` — production Electron package completed

### DIFF
--- a/frontend/e2e/support/fake-bridge.ts
+++ b/frontend/e2e/support/fake-bridge.ts
@@ -1,5 +1,6 @@
 import type { Page } from "@playwright/test";
 
+import type { UpdateSettings, UpdateStatus } from "../../src/main/update-settings";
 import type { AoBridge } from "../../src/preload";
 import type { DaemonStatus } from "../../src/shared/daemon-status";
 import { coerceUiSettings, DEFAULT_UI_SETTINGS } from "../../src/shared/ui-locale";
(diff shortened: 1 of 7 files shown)


Pulkit7070 (approved):
LGTM. Verified the fix by reproduction: with the per-event onStatusEvent callback removed, the manual-check loader deterministically stays stuck (both race tests fail, button never re-enables); with the fix it clears correctly and preserves the staged downloaded update. The request-ID match (matching id + non-checking state) handles stale and double-click cases, the main-process event order is covered by auto-updater.test.ts, and the styles.css change is scoped to the updater panel. Full update suite passes (30 + 59).

PR #4393, opened 2026-08-24
fix(runtime): isolate AO tmux state

### LINKED ISSUE #4391: Packaged tmux can read and control user-owned tmux state
## Bug

The packaged macOS/Linux app invokes AO's bundled tmux binary, but the runtime is not fully isolated from tmux state owned by the user.

**Source:** User report and follow-up design discussion in an AO task on 2026-08-24/25 (no public chat URL)
**Analyzed against:** `5e408467b`
**Confidence:** High
**Priority:** P1 — a socket collision can direct AO input, restart, or destroy operations at a user-owned tmux session.

## Reproduction

1. Start a user tmux server using the named socket `tmux -L ao`, or keep a same-named session on the default server.
2. Configure `~/.tmux.conf` with options/hooks that are inappropriate for AO.
3. Launch packaged AO and create or restore a TUI session.
4. Observe that AO uses the fixed `-L ao` namespace, allows a newly started server to load user configuration, and probes the default tmux socket when the AO-named server does not contain a matching session.

A custom `TMUX_TMPDIR` also changes where `-L ao` resolves, so the apparently private namespace is still controlled by inherited host environment.

## Root Cause

- `frontend/src/main.ts` injects the constant `AO_TMUX_SOCKET_NAME=ao`.
- `backend/internal/adapters/runtime/tmux/tmux.go` invokes tmux with `-L ao`, without an explicit `-f` configuration.
- The adapter caches socket discovery in memory and falls back to the machine default socket using only a matching session name as evidence.
- The tmux control process inherits a surrounding `TMUX` identity.
- `backend/internal/tmuxbin/resolve.go` may fall through to a PATH tmux if a recognizable packaged layout is missing its bundled sibling.

## Fix

- Derive a stable absolute socket path from the daemon run-file identity and use `-S` for every tmux operation.
- Keep the socket inode beside the run file; use a bounded, validated `/tmp` directory alias only when macOS's AF_UNIX path ceiling requires it.
- Pass `-f /dev/null` so AO never loads system/user tmux configuration or plugins.
- Remove automatic named/default socket discovery; every control operation stays on the AO-private socket.
- Remove only tmux identity/internal selector variables while preserving the user's workload environment, and refresh that environment for every new or respawned pane.
- Fail closed when a packaged daemon's bundled tmux is missing.
- Keep Windows on the existing ConPTY adapter.

## Impact

- User tmux servers and AO can collide across socket namespaces.
- AO may inherit user hooks/plugins/options and fail unpredictably.
- A stale AO handle plus a same-named user session can cause AO to operate on foreign state.
- Machines without system tmux should continue to work through the bundled artifact; the terminfo regression in that artifact is tracked separately.

## Related

- [Issue #4382](https://github.com/Untrivial-ai/agent-orchestrator/issues/4382) / [PR #4385](https://github.com/Untrivial-ai/agent-orchestrator/pull/4385) — bundled tmux terminfo artifact failure.
- [PR #4293](https://github.com/Untrivial-ai/agent-orchestrator/pull/4293) — introduced bundled tmux and the fixed `-L ao` namespace.
- [Issue #2662](https://github.com/Untrivial-ai/agent-orchestrator/issues/2662) / [PR #2664](https://github.com/Untrivial-ai/agent-orchestrator/pull/2664) — broader cross-data-dir session namespace collision.

illegalcall:
Fixes [#4391](https://github.com/Untrivial-ai/agent-orchestrator/issues/4391).

Complements [#4385](https://github.com/Untrivial-ai/agent-orchestrator/pull/4385), which fixes the missing packaged terminfo database seen in the original report.

## Summary

- derive a stable per-run-file private tmux socket, force an empty tmux config, and never probe or adopt user/default tmux state
- validate socket-directory ownership and permissions, with an owner-only short alias for macOS Unix-socket limits
- refresh merged daemon/project pane environments over stdin, remove stale global/session variables, and keep configured values out of process arguments
- fail closed when a packaged tmux resource is unavailable; leave Windows on ConPTY

## Verification

- `npm run lint`
- `cd backend && go test -race ./internal/adapters/runtime/tmux ./internal/adapters/runtime/runtimeselect ./internal/tmuxbin`
- Windows cross-compilation for the tmux and runtime-selector packages

## Migration and risk

This is an intentional strict cutover. Sessions on the old `-L ao` or default tmux server remain running but are not adopted, because AO cannot prove that a same-named server belongs to AO rather than the user. New sessions use only the private socket. The socket identity remains stable across ordinary releases; a future incompatible tmux protocol upgrade will require an explicit migration policy.

### DIFF
--- a/backend/internal/adapters/runtime/runtimeselect/runtimeselect.go
+++ b/backend/internal/adapters/runtime/runtimeselect/runtimeselect.go
@@ -4,7 +4,10 @@ package runtimeselect
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"log/slog"
+	"path/filepath"
 	"runtime"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/runtime/conpty"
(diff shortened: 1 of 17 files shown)


nikhilachale (approved):
Yes — the PR is working properly for the macOS development path I tested.

PR #4395, opened 2026-08-24
fix(storage): repair legacy migration profiles

somewherelostt:
## Summary
- retire the exact SQLite profile shipped by `v0.11.1-pr2863.202607311655`, whose foreign migrations occupied 45, 46, 50, and 51 and left key/value `app_settings`
- fingerprint the published tables, indexes, triggers, settings, lineage, and CDC values before any destructive write
- atomically archive the abandoned Pipeline values, remove the live Pipeline schema/CDC, create canonical settings, and record a terminal recovery marker
- reserve the burned migration numbers and add a frozen, provenance-documented release fixture exercised through production `sqlite.Open`

## Safety
- no committed SQLite migration is edited
- fresh and healthy main databases take a constant-time gate; they do not scan the append-only `change_log`
- completed recovery self-disables, so a future legitimate `pipeline_*` schema remains outside this shim
- unknown settings, event types, triggers, indexes, or schema definitions fail closed before mutation
- all recovery writes and postcondition checks run in one transaction; injected failure proves full rollback
- Pipeline settings, CDC rows, table rows, and credentials are retained under `_ao_legacy_pr2863_*` forensic archive tables
- unseen partial Nightly variants are intentionally rejected instead of guessed; this PR supports the exact published profile named above

## Before / after
Before the hardened recovery, the exact release fixture failed startup validation because its legacy settings were discarded rather than archived, and an unknown legacy setting was silently accepted. After this change, the real release profile upgrades through current main, preserves all seeded values, passes integrity and foreign-key checks, reopens idempotently, and rejects altered profiles without mutation.

## Local verification
- focused exact-release and adversarial E2E tests, repeated 10 times
- `cd backend && go test ./internal/storage/sqlite/... -count=1`
- `cd backend && go test -race ./internal/storage/sqlite/... -count=1`
- `git diff --check`

### DIFF
--- a/backend/internal/storage/sqlite/db.go
+++ b/backend/internal/storage/sqlite/db.go
@@ -131,6 +131,9 @@ func migrate(db *sql.DB) error {
 	if err := goose.SetDialect("sqlite3"); err != nil {
 		return fmt.Errorf("set goose dialect: %w", err)
 	}
+	if err := retirePublishedPipelinesProfile(context.Background(), db); err != nil {
+		return fmt.Errorf("retire published Pipelines migration profile: %w", err)
+	}
 	if err := repairRenumberedChatMigrationHistory(db); err != nil {
 		return fmt.Errorf("repair renumbered chat migration history: %w", err)
 	}
@@ -155,12 +158,6 @@ func migrate(db *sql.DB) error {
 	if err := prepareQueuedTurnPromotionMigration(db); err != nil {
 		return fmt.Errorf("prepare queued-turn promotion migration: %w", err)
 	}
-	if err := prepareLegacyPipelineCDCBeforeReviewRunMigration(db); err != nil {
-		return fmt.Errorf("prepare legacy pipeline CDC migration: %w", err)
-	}
-	if err := prepareAppSettingsSingletonMigration(db); err != nil {
-		return fmt.Errorf("prepare app settings singleton migration: %w", err)
-	}
 	// Builds can advance a database past a migration that is added or
 	// renumbered later (notably across fast-moving Nightly releases). Apply
 	// those embedded migrations instead of permanently wedging daemon startup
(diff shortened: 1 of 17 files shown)


somewherelostt:
Additional verification after opening the PR:

- `cd backend && go test ./internal/storage/sqlite -run "TestMigration0103DropsLegacyPipelineCDCTriggers|TestMigrateRepairsLegacyKeyValueAppSettings" -count=1 -v` ✅
- `cd backend && go test ./internal/storage/sqlite/... -count=1` ✅
- `cd backend && go test ./internal/storage/sqlite -run "TestMigration0103DropsLegacyPipelineCDCTriggers|TestMigrateRepairsLegacyKeyValueAppSettings" -count=10` ✅
- `cd backend && go build ./...` ✅
- `git diff --check` ✅

Black-box upgrade test:
- Built the PR daemon into a temp directory.
- Copied the original pre-repair AO database backup into an isolated temp `AO_DATA_DIR`.
- Started the PR daemon with isolated `AO_DATA_DIR`, `AO_RUN_FILE`, and `AO_PORT=33901`.
- Daemon reached `ready`.
- Migration ledger advanced through `0108`.
- `pipeline%cdc%` triggers after migration: none.
- `app_settings`: `1|chat`.
- `PRAGMA integrity_check`: `ok`.
- Stopped the temp daemon; verified the temp daemon PID was no longer running.

One extra check attempted but blocked by local Windows toolchain:
- `go test -race ./internal/storage/sqlite -run "TestMigration0103DropsLegacyPipelineCDCTriggers|TestMigrateRepairsLegacyKeyValueAppSettings" -count=1` failed before package build with `runtime/cgo: cc1.exe: sorry, unimplemented: 64-bit mode not compiled in`.

somewherelostt:
The AO daemon failed because your local SQLite database had leftover schema from older/nightly AO builds.

There were two blockers:

1. Old pipeline CDC triggers still existed, like `pipeline_definitions_cdc_insert`.
   Migration `0103` rebuilds the `change_log` table, but SQLite refuses to drop a table while an existing trigger still references it.

2. After that was fixed, `app_settings` was still in an older key/value schema.
   Newer migrations expected the newer singleton schema with `id`, `default_session_mode`, and `updated_at`.

How we fixed your machine:
- Backed up your DB at:
  `C:\Users\abuma\.ao\data\backup-before-0103-trigger-fix-20260825-012117`
- Dropped the stale `pipeline_*_cdc_*` triggers.
- Repaired `app_settings` to the current schema.
- Restarted AO daemon.
- Confirmed status:
  `AO daemon: ready`, port `3001`.

How we fixed it for others:
- Opened PR [[#4395](https://github.com/Untrivial-ai/agent-orchestrator/pull/4395)](https://github.com/Untrivial-ai/agent-orchestrator/pull/4395).
- Added startup repair logic before migrations run.
- Added regression tests for both broken DB shapes.
- Verified with storage tests, build, repeated regression tests, and a black-box test against a temp copy of your original broken DB.

illegalcall:
## Follow-up: exact feature-release upgrade now passes

I reproduced the migration problem against the published Pipelines feature build, then tested the fix against a fresh clone of the same database.

### Before

1. Created `ao.db` by running the real storage `Open` path at tag `v0.11.1-pr2863.202607311655`.
2. Cloned that database so the before/after runs started from identical bytes.
3. Opened the first copy at the original PR head `1ad1798a8`.

Result:

```text
=== RUN   TestManualUpgradeEvidence
run migrations: ERROR 0067_app_settings.sql:
SQL logic error: table app_settings already exists (1)
--- FAIL: TestManualUpgradeEvidence
```

The original regression test started at migration 66 and inserted migration 67 as applied. Starting at 66 activates the chat-renumber compatibility mapper, which masks the real published profile: the feature tag stops at migration 51 and has no migration 67 ledger entry.

### Fix

Commit `a0aa21473` now:

- recognizes only the exact legacy `app_settings(key, value)` shape;
- recreates the canonical singleton table and records migration 67 in the same transaction when it was not already applied;
- preserves the existing valid `default_session_mode` value;
- accepts both field profiles: 67 pending and 67 already burned/applied;
- rejects partial/unknown settings schemas without deleting them;
- extends the pipeline regression to prove legacy `change_log` rows are removed as well as stale triggers.

### After

Opening the untouched second database copy with `a0aa21473`:

```text
=== RUN   TestManualUpgradeEvidence
--- PASS: TestManualUpgradeEvidence (0.20s)
PASS
```

Post-upgrade inspection:

```text
integrity_check: ok
migration_67: 1
latest_migration: 108
app_settings_columns: [id default_session_mode updated_at]
default_session_mode: chat
pipeline_cdc_triggers: 0
```

### Local verification

```text
go test ./internal/storage/sqlite/...                                      PASS
focused migration regressions, -count=10                                  PASS
go test -race ./internal/storage/sqlite/...                                PASS
go build ./...                                                             PASS
golangci-lint v2.12.2                                                      0 issues
git diff --check                                                           PASS
```

`npm run lint` reached one unrelated Prime Agent hook timing failure (`got 9 calls, want 10`) during the repository-wide test phase. The exact failing test subsequently passed 10/10 in isolation; the storage suites and linter are clean.

illegalcall:
Adversarial before/after evidence for the published Pipelines database profile

**Before (`a0aa21473`, production implementation with the new black-box tests):**

```text
TestOpenUpgradesPublishedPipelineProfileWithoutLosingData
  FAIL: no such table: _ao_legacy_pr2863_app_settings

TestOpenRejectsUnknownPublishedPipelineSettingWithoutMutation
  FAIL: open profile with unknown legacy setting succeeded, want fail-closed error
```

This showed that the earlier repair could make startup continue, but it did not preserve the retired setting and it could silently mutate a profile containing unknown state.

**After (`431a8c58b`):**

```text
go test ./internal/storage/sqlite -run <exact-release-and-adversarial-tests> -count=10
ok  github.com/aoagents/agent-orchestrator/backend/internal/storage/sqlite  22.036s

go test ./internal/storage/sqlite/... -count=1
PASS

go test -race ./internal/storage/sqlite/... -count=1
ok  .../internal/storage/sqlite            279.755s
ok  .../internal/storage/sqlite/sqlitetest   9.937s
ok  .../internal/storage/sqlite/store       59.975s

go test -p 1 ./...
PASS

go build ./...
PASS

golangci-lint v2.12.2 run --path-mode=abs
0 issues

git diff --check
PASS
```

The E2E fixture freezes migrations 0040 through 0051 verbatim from `v0.11.1-pr2863.202607311655`, seeds a definition, run, stage, signal, credential, setting, Pipeline CDC, and ordinary CDC, closes the database, then upgrades through the real production `sqlite.Open` seam.

Post-upgrade assertions prove:
- canonical `app_settings` and current migrations apply
- every seeded Pipeline value is retained in `_ao_legacy_pr2863_*` archives
- ordinary CDC survives while retired Pipeline CDC leaves the live log
- active Pipeline tables/triggers are removed
- `integrity_check` and `foreign_key_check` pass
- a second real startup is idempotent
- injected ledger failure rolls back every mutation
- unknown settings, CDC event types, triggers, and modified published trigger definitions fail before mutation
- the completion marker prevents repeated CDC/FK scans and prevents the shim from claiming a future legitimate `pipeline_*` schema

Adversarial verdict: net positive for the exact published profile, with no known fatal data-integrity flaw. The intentional tradeoff is narrow compatibility: unseen partial Nightly variants fail closed. Archive tables preserve forensic row values, not a directly runnable old schema, and retained Pipeline credentials continue to occupy the same local database trust boundary.

nikhilachale:
[P2] Avoid a full change_log scan on ordinary pre-103 upgrades

For a healthy existing database at migration 102 or earlier, applied103 is false, so hasPublishedPipelineHints is called with event inspection enabled. With canonical app_settings and no pipeline objects, that reaches SELECT EXISTS ... WHERE event_type GLOB 'pipeline_*'. change_log has no event_type index and is append-only, so a normal upgrade performs an O(total CDC rows) synchronous scan before the daemon can start. On a long-lived installation with millions of unrelated CDC events this can turn an otherwise ordinary upgrade into a substantial startup stall, contrary to the constant-time gate described in the PR.

Could event inspection be gated on stronger legacy evidence (for example the foreign lineage/profile shape), or use an indexed/bounded detection path? A regression test using a large non-Pipeline change log would help keep the startup gate cheap.

i-trytoohard:
@somewherelostt Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4398, opened 2026-08-24
fix(chat): present Claude questions sequentially

nikhilachale:
## What

Present Claude `AskUserQuestion` forms one question at a time, with Back/Next navigation and a final combined submission.

## Why

`claude-agent-acp` encodes multiple questions as `question_N` and `question_N_custom` properties in one ACP form schema. The existing renderer displayed every property simultaneously, producing one long form instead of Claude's sequential question flow.

## How

- Recognize schemas made up of `question_N` fields and optional matching `question_N_custom` fields.
- Group and order those fields by question number.
- Validate only the active group before advancing while preserving all entered values.
- Submit all answers together from the final question.
- Keep the existing all-fields layout for generic MCP forms.

## Testing

- `npm --prefix frontend test -- --run src/renderer/components/chat/ElicitationCard.test.tsx --reporter=dot` — 8/8 passing.
- `npm --prefix frontend run typecheck` — passing.
- The full concurrent frontend suite remains non-green because an unrelated `ProjectSettingsForm.test.tsx` test intermittently fails to observe its `Saved` state; it also failed when isolated. The focused ElicitationCard suite is green.

## Checklist

- [ ] Relevant CI checks pass for the area touched (`go`, frontend, etc.) — focused tests and typecheck pass; see unrelated full-suite failure above


nikhilachale (commented):
Changes requested. The sequential-question implementation and focused tests look reasonable, but the PR currently includes unrelated agent-role smoke tooling that is not described by the PR and is outside this fix. Please remove those files from this branch before merging.

skills/agent-role-smoke/SKILL.md
```diff
@@ -0,0 +1,100 @@
+---
```
nikhilachale: [P1] Remove the unrelated smoke-test skill from this PR. Against the target `main`, this branch adds this 100-line skill plus `skills/agent-role-smoke/scripts/agent-role-smoke.mjs` (861 lines), while the PR is scoped to the ElicitationCard fix and its body reports only that work. Keeping these files would merge a separate feature without relevant review/testing; rebase or cherry-pick the chat commit onto current target `main` so the PR contains only the two ElicitationCard files.

nikhilachale: Fixed in 8f687f413. I removed both unrelated agent-role-smoke files from the branch without rewriting shared history. GitHub now reports the PR diff contains only ElicitationCard.tsx and ElicitationCard.test.tsx; the focused ElicitationCard suite remains 8/8 passing.

Pulkit7070:
Scope note (not blocking): the diff is +1128, but only ~167 lines are the actual fix (`ElicitationCard.tsx` +69 and its test +120). The other ~900 lines are a separate `skills/agent-role-smoke` skill (SKILL.md + agent-role-smoke.mjs) that is unrelated to presenting Claude questions sequentially. Bundling it here makes the PR look much larger than the change and couples an unrelated skill to a chat-UI fix. Consider splitting the skill into its own PR so each can be reviewed and reverted independently.

Pulkit7070 (approved):
LGTM. Verified locally: reproduced the bug before/after (reverting visibleProperties to the full set renders all questions as one long form and fails the sequential test; the fix shows one question at a time). The PR's 8 tests plus 3 parser edge cases I added (non-contiguous question numbers, orphan _custom, non-question field) all pass, and the grouping parser safely falls back to the all-fields layout on any non-AskUserQuestion schema. Typecheck clean, CI green. One non-blocking scope note left separately about the bundled agent-role-smoke skill.

PR #4402, opened 2026-08-25
fix(windows): keep worker ao path pinned

### LINKED ISSUE #4401: bug(windows): worker ao browser can resolve to broken npm shim
## Bug

On Windows, a worker can fail before `ao browser` even reaches the desktop browser runtime:

```text
@aoagents/ao: no prebuilt binary for win32-x64.
The optional dependency @aoagents/ao-win32-x64 is not installed
```

This is confusing because the desktop app does ship a working AO daemon binary. The worker should be using that pinned daemon binary, not falling through to whatever old npm `ao` shim happens to be on the user's PATH.

## What seems to be happening

Windows treats environment variable names case-insensitively, but parts of our launch environment handling were treating them as case-sensitive.

That means a worker could inherit something like:

```text
Path=C:\Users\me\AppData\Roaming\npm
```

and AO could also add:

```text
PATH=C:\Program Files\Agent Orchestrator\resources\daemon
```

Those two entries can survive together. In practice the inherited npm path can win, so `ao browser ...` resolves to the npm shim. If that npm install was done with optional dependencies omitted, the shim dies with the missing `@aoagents/ao-win32-x64` error.

## Expected behavior

Workers on Windows should resolve bare `ao` commands to the AO daemon/app binary that launched the session. The user's global npm install should not decide whether `ao browser` works inside an AO worker.

## Impact

This blocks browser-panel control from affected Windows workers. The workaround is to repair the npm install with optional deps, but users should not need npm at all for this path.

## Fix direction

When merging worker launch environments on Windows, replace environment keys case-insensitively so AO's pinned `PATH` replaces inherited `Path` instead of coexisting with it.

somewherelostt:
Fixes #4401

## Summary

This keeps Windows workers from accidentally resolving `ao` through an inherited npm PATH entry.

The failure mode was `Path` vs `PATH`: Windows treats env keys case-insensitively, but our launch env merge did not. If a worker inherited `Path=...\npm` and AO added `PATH=...\resources\daemon`, both could survive. Then `ao browser` could hit the old npm shim and fail with the missing `@aoagents/ao-win32-x64` optional dependency error.

This PR fixes both worker launch paths:

- ConPTY terminal workers now merge Windows env keys case-insensitively before launching the pty host.
- Chat-driver workers do the same when building the provider process env.
- Regression tests cover inherited `Path` being replaced by AO's pinned `PATH`.

## Testing

Ran this focused set three times with `-count=1`:

```bash
go test ./internal/adapters/runtime/conpty ./internal/adapters/chatdriver/processenv ./internal/adapters/chatdriver/codexappserver -run "TestMerge|TestStripEnvAssignments|TestEnvSliceIsSortedForReproducibleRelaunch" -count=1
```

Results:

- Pass 1: ok for conpty, processenv, codexappserver
- Pass 2: ok for conpty, processenv, codexappserver
- Pass 3: ok for conpty, processenv, codexappserver

I kept the test scope focused on the files and behavior touched here. A broader chat-driver/session-manager run on this Windows checkout hits unrelated Windows test assumptions such as `/tmp/ws` not being an absolute Windows path, so I did not include those as PR blockers for this small fix.


somewherelostt:
Manual Windows verification from the PR branch:

- Checked out #4402 locally and launched the Electron app in dev mode against the PR-built daemon (`~/.ao/dev`, port 3002).
- Created a worker session from that app and opened it in the AO UI.
- From the worker, ran `ao browser open https://x.com/Maaztwts/status/2091477804626120991` repeatedly.
- Result: the AO Browser panel opened the X post. The command reported exit code 0 with no stderr. The previous npm optional dependency failure (`@aoagents/ao-win32-x64 is not installed`) did not reproduce.
- X still emits page-level 404 network noise for its sidebar/guide endpoints, but that is after the browser opens and is not the CLI/package failure.

Local focused regression tests were also rerun with `-count=3`:

```text
go test ./internal/adapters/chatdriver/processenv -run "TestMergeWithWindowsPathOverlayReplacesInheritedPathCase|TestMergeWithCaseSensitiveModeKeepsDistinctPathKeys" -count=3
go test ./internal/adapters/runtime/conpty -run "TestMergeWindowsEnvReplacesInheritedPathWithOverlay|TestMergeWindowsEnvAssignmentsOverrideCaseInsensitive" -count=3
```

Both passed. GitHub checks are green, including `native (windows-latest)`.

somewherelostt:
@Vaibhaav-Tiwari can you test this pr!

Vaibhaav-Tiwari:
> @Vaibhaav-Tiwari can you test this pr!

doing

Vaibhaav-Tiwari (changes requested):
The normal Windows workflow works, and the committed focused tests pass. However, the important Path/PATH collision test fails in
  both Chat and ConPTY. A project-defined Path can nondeterministically override AO’s pinned PATH, causing the legacy npm ao binary to
  run.

  Approval condition: deterministic AO-owned PATH precedence plus regression tests for both environment-merging paths.

somewherelostt:
Addressed the Path/PATH collision from the review.

The missing case was Windows env maps that contain both a project `Path` and AO's pinned uppercase `PATH`. Since Go map iteration is random, either one could be applied last. I changed both merge paths so AO's pinned `PATH` is always applied last on Windows:

- Chat worker env merge now orders the overlay and defers uppercase `PATH` until the end.
- ConPTY now applies argv-prefix `env KEY=value` assignments before the AO overlay, then also defers uppercase `PATH` until the end.
- Added focused regression coverage for project `Path` vs pinned `PATH` in both paths, plus the assignment collision case for ConPTY.

Tested:

```text
go test ./internal/adapters/chatdriver/processenv -run "TestMergeWithWindowsPinnedPATHWinsOverProjectPath|TestMergeWithWindowsPathOverlayReplacesInheritedPathCase" -count=10
go test ./internal/adapters/runtime/conpty -run "TestMergeWindowsEnvPinnedPATHWinsOverProjectPath|TestMergeWindowsEnvPinnedPATHWinsOverAssignments|TestMergeWindowsEnvReplacesInheritedPathWithOverlay" -count=10
go test ./internal/adapters/chatdriver/processenv -count=1
go test ./internal/adapters/runtime/conpty -count=1
go test ./internal/adapters/chatdriver/codexappserver -run "TestEnvSliceIsSortedForReproducibleRelaunch" -count=1
```

Also ran this focused suite 3 separate times, all green:

```text
go test ./internal/adapters/runtime/conpty ./internal/adapters/chatdriver/processenv ./internal/adapters/chatdriver/codexappserver -run "TestMerge|TestStripEnvAssignments|TestEnvSliceIsSortedForReproducibleRelaunch" -count=1
```

One note: a broader `codexappserver -count=3` run still hits unrelated Windows-only test fixture assumptions around `/tmp/ws` not being an absolute Windows path. I kept this patch scoped to the env collision instead of mixing that cleanup into this PR.

Vaibhaav-Tiwari:
@somewherelostt can you please resolve the merge conflicts?

Vaibhaav-Tiwari (approved):

PR #4410, opened 2026-08-25
fix(cursor): resolve ACP model aliases

nikhilachale:
## What

Resolve Cursor CLI model aliases to the exact model values advertised by the ACP session before calling the legacy `session/set_model` endpoint.

## Why

Cursor's TUI accepts CLI aliases such as `composer-2.5`, but Cursor ACP advertises and validates parameterized values such as `composer-2.5[fast=false]`. The Chat UI forwarded the raw CLI alias, causing task startup to fail with `Invalid model value: composer-2.5`.

## How

- Preserve exact advertised model values when already selected.
- Map unambiguous non-fast and fast CLI aliases to their advertised parameterized ACP values.
- Reject missing, unavailable, or ambiguous advertised model choices with `ErrChatConfigOptionInvalid` instead of sending an unsafe raw fallback.
- Keep the legacy setter behavior used by existing ACP providers.

## Testing

- `go test ./...`
- `go test -race -count=1 ./internal/adapters/chatdriver/acp/... ./internal/adapters/chatdriver/cursoracp/...`
- `go vet ./internal/adapters/chatdriver/acp/... ./internal/adapters/chatdriver/cursoracp/...`
- `go build ./...`
- Added regression coverage for Cursor's parameterized model choices, missing model catalogs, and ambiguous aliases.

## Checklist

### DIFF
--- a/backend/internal/adapters/chatdriver/acp/config_options.go
+++ b/backend/internal/adapters/chatdriver/acp/config_options.go
@@ -312,11 +312,18 @@ func resolveLegacyModelChoice(choices []ports.ChatConfigOptionChoice, requested 
 
 	matched := ""
 	for _, choice := range choices {
-		alias, parameterized := parameterizedModelAlias(choice.Value)
+		aliases, parameterized := parameterizedModelAliases(choice.Value)
 		if !parameterized {
 			continue
 		}
-		if alias != requested {
+		matches := false
+		for _, alias := range aliases {
+			if alias == requested {
+				matches = true
+				break
+			}
+		}
+		if !matches {
 			continue
 		}
 		if matched != "" {
@@ -327,30 +334,40 @@ func resolveLegacyModelChoice(choices []ports.ChatConfigOptionChoice, requested 
 	return matched, matched != ""
 }
 
-func parameterizedModelAlias(value string) (string, bool) {
+func parameterizedModelAliases(value string) ([]string, bool) {
 	open := strings.IndexByte(value, '[')
 	if open <= 0 || !strings.HasSuffix(value, "]") {
-		return "", false
+		return nil, false
 	}
-	alias := value[:open]
+	base := value[:open]
+	body := value[open+1 : len(value)-1]
+	if body == "" {
+		if base == "default" {
+			return []string{"auto"}, true
+		}
+		return nil, false
+	}
+	if strings.HasPrefix(base, "grok-") {
+		base = "cursor-" + base
+	}
 	effort := ""
 	fast := false
 	fastSet := false
 	thinking := false
 	thinkingSet := false
 	parameterized := false
-	params := strings.Split(value[open+1:len(value)-1], ",")
+	params := strings.Split(body, ",")
 	for _, param := range params {
 		key, raw, found := strings.Cut(strings.TrimSpace(param), "=")
 		key = strings.TrimSpace(key)
 		raw = strings.TrimSpace(raw)
 		if !found || key == "" || raw == "" {
-			return "", false
+			return nil, false
 		}
 		switch key {
 		case "reasoning", "effort", "reasoning_effort":
 			if effort != "" && effort != raw {
-				return "", false
+				return nil, false
 			}
 			effort = raw
 			parameterized = true
@@ -361,10 +378,10 @@ func parameterizedModelAlias(value string) (string, bool) {
 				parsed = true
 			case "false":
 			default:
-				return "", false
+				return nil, false
 			}
 			if fastSet && fast != parsed {
-				return "", false
+				return nil, false
 			}
 			fast = parsed
 			fastSet = true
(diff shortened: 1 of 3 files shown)


backend/internal/adapters/chatdriver/acp/config_options.go
```diff
@@ -298,6 +299,80 @@ func choiceOffered(choices []ports.ChatConfigOptionChoice, value string) bool {
...
+		if matched != "" {
+			return "", false
+		}
+		matched = choice.Value
+	}
+	if matched != "" {
+		return matched, true
+	}
+
+	// Cursor may advertise only one parameterized value for a model base, with
```
illegalcall: **[P1] Non-fast aliases can silently select fast models.** [`config_options.go:334`](https://github.com/Untrivial-ai/agent-orchestrator/blob/6c647b7c68d0bc0ae1bb58598bad0aa8344e5958/backend/internal/adapters/chatdriver/acp/config_options.go#L334-L349) maps `composer-2.5` to `composer-2.5[fast=true]` when that is the sole matching choice. Cursor exposes `composer-2.5` and `composer-2.5-fast` as distinct aliases, so the non-fast request is unavailable and should return `ErrChatConfigOptionInvalid`. The new test at [`driver_test.go:1507`](https://github.com/Untrivial-ai/agent-orchestrator/blob/6c647b7c68d0bc0ae1bb58598bad0aa8344e5958/backend/internal/adapters/chatdriver/acp/driver_test.go#L1507-L1544) codifies the unsafe behavior.

backend/internal/adapters/chatdriver/acp/config_options.go
```diff
@@ -298,6 +299,80 @@ func choiceOffered(choices []ports.ChatConfigOptionChoice, value string) bool {
...
+// opaque value an ACP agent advertised. Cursor's CLI lists composer-2.5 while
+// its legacy ACP selector offers composer-2.5[fast=false]; session/set_model
+// accepts only the latter. Exact provider values always win, and an alias is
+// accepted only when it identifies one advertised choice unambiguously.
+func resolveLegacyModelChoice(choices []ports.ChatConfigOptionChoice, requested string) (string, bool) {
+	if choiceOffered(choices, requested) {
+		return requested, true
+	}
+
+	matched := ""
```
illegalcall, line 313: **[P2] Most parameter-derived Cursor aliases remain unresolved.** [`config_options.go:312`](https://github.com/Untrivial-ai/agent-orchestrator/blob/6c647b7c68d0bc0ae1bb58598bad0aa8344e5958/backend/internal/adapters/chatdriver/acp/config_options.go#L312-L373) only derives `base` and `base-fast`. Cursor’s CLI advertises aliases such as `gpt-5.5-medium`, while ACP uses values such as `gpt-5.5[context=272k,reasoning=medium,fast=false]`. The resolver rejects that mapping, so selecting many models still prevents Chat startup. Tests also omit this case and a direct `composer-2.5-fast` request.

backend/internal/adapters/chatdriver/acp/config_options.go
nikhilachale: Fixed in dc923889b. Removed the sole-base fallback, so composer-2.5 no longer resolves to composer-2.5[fast=true]; startup now returns ErrChatConfigOptionInvalid and the test verifies session/set_model is not called.

backend/internal/adapters/chatdriver/acp/config_options.go
nikhilachale: Fixed in dc923889b. Parameterized aliases now include reasoning and fast semantics, covering gpt-5.5-medium, composer-2.5-fast, and combined reasoning/fast aliases while retaining ambiguity rejection.

illegalcall: **[P1] Parameterized aliases still drop Cursor model semantics.** This remains reproducible on `dc923889b`: `parameterizedModelAlias` recognizes only `reasoning` and `fast`, while current Cursor metadata also uses `effort`, `thinking`, `context`, and `reasoning_effort`. For example, `gemini-3.6-flash[effort=high]` does not resolve from the CLI alias `gemini-3.6-flash-high`. More seriously, `claude-opus-5[thinking=true,context=300k,effort=high,fast=false]` is reduced to the alias `claude-opus-5`, so a plain request can silently select the high-thinking variant. Values with unrecognized semantic parameters need to be rejected or explicitly mapped, with regression coverage for these actual Cursor parameter shapes.

illegalcall: @nikhilachale this still remains

nikhilachale: Fixed in 71e2d72ba. The alias parser now maps effort, reasoning, and reasoning_effort to the CLI effort suffix; validates thinking, context, and fast explicitly; and rejects unknown, malformed, or conflicting semantic parameters instead of dropping them. Regression coverage includes gemini-3.6-flash[effort=high], claude-opus-5[thinking=true,context=300k,effort=high,fast=false], reasoning_effort, the unsafe plain-Claude request, and unknown semantics.

illegalcall (approved):
Verified at 196c42309. Thinking and non-thinking aliases now resolve distinctly, both observed Claude thinking alias orders are supported, Cursor Grok prefixes and auto/default are handled, and unsafe cross-variant selection is rejected. Full local tests and all GitHub checks pass.

PR #4413, opened 2026-08-25
feat: allow safe initial prompt edits without forking

nikhilachale:
## What

Enable Edit for the first provider-backed prompt even when the active chat provider does not advertise history forking. Keep later prompts gated behind the existing `fork` capability, and hide the first-prompt affordance when older history has not loaded.

## Why

Cursor chat sessions currently expose no Edit action because Cursor ACP does not provide AO with an anchored historical fork primitive. AO already handles editing the initial prompt safely by starting a fresh conversation, so that limited case can be enabled without replaying history, mutating provider storage, or adding a dependency.

## How

- Pass the edit handler to the timeline independently of provider fork support.
- For fork-capable providers, preserve editing for every provider-backed turn.
- For providers without fork support, expose editing only on the first known provider turn.
- If `hasMoreBefore` is true, expose no edit action because the UI cannot prove the visible turn is the first prompt.

Intentional omission: later Cursor prompt editing remains unavailable until Cursor ACP exposes an anchored fork or rewind operation. This PR does not simulate that behavior by replaying transcript history.

## Testing

- `node node_modules/vitest/vitest.mjs run --config vite.renderer.config.ts src/renderer/components/chat/ChatWorkspace.test.tsx -t "offers only the first prompt|withholds first-prompt editing" --maxWorkers=1 --no-file-parallelism --testTimeout=30000`
- `node node_modules/typescript/bin/tsc --noEmit`
- Earlier full `ChatWorkspace.test.tsx` run: 65/65 passed.

The repository-wide frontend run was also attempted locally; unrelated suites require separately installed landing-page modules and loopback listener permissions unavailable in the sandbox. The focused chat suite and frontend type-check pass.

## Checklist

### DIFF
--- a/backend/internal/httpd/controllers/conversation_history_test.go
+++ b/backend/internal/httpd/controllers/conversation_history_test.go
@@ -302,6 +302,19 @@ func TestActivateConversationBranchRouteResumesWithoutBody(t *testing.T) {
 	}
 }
 
+func TestActivateConversationBranchRouteDecodesEscapedRootID(t *testing.T) {
+	svc := &fakeChatService{activate: "conversation-1:root"}
+	srv := newChatTestServer(t, svc)
+	body, status, _ := doRequest(t, srv, http.MethodPost,
+		"/api/v1/sessions/ao-1/conversation/branches/conversation-1%3Aroot/activate", "")
+	if status != http.StatusAccepted {
+		t.Fatalf("status = %d, body = %s", status, body)
+	}
+	if svc.gotBranchID != "conversation-1:root" {
+		t.Fatalf("branch id = %q, want decoded root id", svc.gotBranchID)
+	}
+}
+
 func TestActivateConversationBranchRouteReportsMissingBranch(t *testing.T) {
 	svc := &fakeChatService{activateErr: domain.ErrNoConversationBranch}
 	srv := newChatTestServer(t, svc)
--- a/backend/internal/httpd/controllers/conversations.go
+++ b/backend/internal/httpd/controllers/conversations.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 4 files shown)


illegalcall:
### E2E verification

Verified this PR locally at commit `95d40cdb9` in the actual Electron app, using an isolated AO data directory.

**Confirmed regression — stopped controller**

1. Open a persisted Cursor chat while its agent controller is stopped (capabilities are unavailable).
2. The UI still exposes **Edit user message** on the first provider-backed prompt.
3. Submitting the edit reaches the backend and returns HTTP `409` with `CHAT_CONTROLLER_NOT_READY`.

This is a safe failure (I saw no data corruption), but it is a user-visible false affordance: the UI offers an operation that cannot succeed.

**Control case — live Cursor controller**

After restoring the same controller to `ready`, the first-prompt edit worked as intended: the request returned `202`, created a conversation branch, rendered the edited prompt, and started the new turn.

The likely cause is that `can(snapshot, "fork") === false` currently represents two different states:

- a live provider that has reported it does not support anchored forks (the intended Cursor case), and
- a stopped controller with no capabilities available.

The first-turn exception should only apply when controller/capability state is authoritative, rather than treating absent capabilities as proof that the provider simply lacks `fork`. A focused regression test for a stopped controller would cover this boundary.

**Stopped-controller failure:**

pr4413-stopped-controller-error

**Live-controller success:**

pr4413-live-cursor-success

nikhilachale:
### Follow-up: stopped-controller fix and verification

Fixed in `f2e715cb6`.

#### Fix

- Gate `onEditMessage` behind an authoritative live controller state (`ready` or `busy`). A stopped, connecting, or recovering controller no longer exposes **Edit user message**, even when the first persisted provider prompt is otherwise marked editable.
- Preserve the intended Cursor exception: once its controller is live, the first provider-backed prompt remains editable even though Cursor does not advertise anchored `fork` support.
- Add a focused regression test covering a stopped controller with unavailable capabilities.

#### Automated verification

- `ChatWorkspace.test.tsx`: **66/66 tests passed**.
- Focused stopped-controller and Cursor-without-fork cases: **passed**.
- `npm run typecheck`: **passed**.

#### Live Cursor verification

Launched the actual Electron development build from this PR with isolated AO state (`~/.ao/dev/data`, daemon on `127.0.0.1:3002`) and exercised a disposable Cursor chat (`scratch-9`):

1. Created a live Cursor chat and completed its first provider-backed turn (`READY`). The controller was `ready`, the first prompt was editable, and Cursor did not advertise `fork`.
2. Stopped the same controller. Its persisted snapshot reported `controller: stopped`, no capabilities, and the first prompt still had server-side `editAvailable: true`; the frontend regression test confirms the edit control is suppressed in exactly this state.
3. Restored the same Cursor controller to `ready` and edited the first prompt.
4. The edit returned **HTTP 202**, changed the active branch, set `branchedFromEarlierMessage: true`, persisted the edited prompt, started a new provider turn, and completed with Cursor replying `EDITED`.
5. Terminated the disposable test session afterward.

The live lifecycle/edit flow was driven through this checkout's CLI and daemon API; native button visibility is asserted by the React regression test.

illegalcall:
Superseded by https://github.com/Untrivial-ai/agent-orchestrator/pull/3858

PR #4416, opened 2026-08-25
feat: enable Cursor chat and TUI handoff

nikhilachale:
## Summary

- preserve Cursor's provider-native conversation ID across Chat UI and Terminal UI
- resume Chat → TUI with Cursor's exact single-argument form `agent --resume=<session-id>` (using AO's resolved Cursor executable); do not replay the transcript to Cursor or trigger an extra model turn
- render AO's durable user/assistant transcript directly into terminal scrollback before Cursor starts, because Cursor does not hydrate ACP-created Chat history into its local TUI store
- exclude system prompts, tool output, attachments, standalone redaction placeholders, ANSI escapes, and unsafe control characters from that display-only transcript
- consume the bounded mode-0600 transcript artifact once, with crash recovery and cleanup for cancellation, rollback, shutdown restore, termination, and superseding transitions
- restore TUI → Chat through ACP `session/load`, reconciling Cursor's native transcript so Chat and terminal turns remain visible once and in chronological order
- wire the production daemon's Chat adapter to the durable transcript reader (the cause of the reported “Cursor Chat history reader is unavailable” failure)

## Verification

- `go test ./internal/daemon ./internal/cli ./internal/service/chat ./internal/session_manager ./internal/adapters/chatdriver/cursoracp`
- `go build ./...`
- `git diff --check`
- regression test verified red before the daemon adapter forwarding method and green after it

## Local desktop validation

Launched the PR checkout as the real Electron desktop app in isolated mode using `~/.ao/dev`. Verified the rebuilt daemon at `127.0.0.1:3002`, renderer at `localhost:5173`, and successful project API response. The pre-fix test session was terminated during the required backend restart, so a new Cursor Chat session is required for the next manual handoff attempt.

### DIFF
--- a/backend/internal/adapters/agent/activitydispatch/dispatch.go
+++ b/backend/internal/adapters/agent/activitydispatch/dispatch.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/auggie"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/claudecode"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/codex"
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/cursor"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/droid"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/fake"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/kimchi"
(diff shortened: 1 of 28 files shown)


i-trytoohard:
@nikhilachale Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

codebanditssss:
Thanks for the contribution. Cursor Chat is already present on current `main` through #4201, and persistent ACP provider lifecycle/handoff support was subsequently added in #4766. The current Cursor ACP implementation also includes Cursor-specific question/plan extensions, permission-mode handling, version checks, and live coverage, so this older branch is no longer necessary.

PR #4440, opened 2026-08-25
fix: report live workspace branch in sessions

manuel-toko:
## Summary

- resolve each active git session's branch from its managed worktree when assembling session read models
- keep the durable spawn-time branch as a fallback when the worktree is unavailable or detached
- validate the workspace path against AO's managed root before running the single read-only Git probe

## Root cause

`SessionMetadata.Branch` is captured when a session is spawned, but an agent or host can change the worktree branch directly with Git. Session GET/list responses continued projecting that seed value, so the API could report `ao/<session>/root` while the worktree and claimed PR were already on a feature branch.

The fix treats the checked-out branch as mutable external state and overlays it at service read-model assembly time. This also covers spawn, restore, and resume responses that use the same assembly path. It does not perform a checkout or mutate SQLite.

## Verification

- `go test ./internal/service/session ./internal/adapters/workspace/gitworktree ./internal/daemon -count=1` (pass)
- `go build ./...` (pass)
- `go vet ./...` (pass)
- Windows/amd64 cross-compile of the three affected test packages (pass)
- `go test ./... -count=1` in the Linux Go 1.25.7 container reaches unrelated environment-sensitive baseline failures:
  - fake-agent lifecycle hook log is not created in the container
  - mobilebridge chmod-based write-failure tests run as container root and remain writable
  - one session-manager test requires tmux, which is not installed in the build container

No frontend or API schema shape changes are required.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/current_branch_test.go
+++ b/backend/internal/adapters/workspace/gitworktree/current_branch_test.go
@@ -1,0 +1,61 @@
+package gitworktree
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+func TestCurrentBranchReadsManagedWorktreeSymbolicRef(t *testing.T) {
+	managedRoot := t.TempDir()
+	workspacePath := filepath.Join(managedRoot, "mer", "mer-1")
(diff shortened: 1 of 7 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4443, opened 2026-08-25
fix(desktop): restore text selection on read-only content, opt-in per surface

### LINKED ISSUE #4268: bug(desktop): no text outside the terminal can be selected or copied — body { user-select: none } disables every content surface

RugvedMavidipalli:
## Summary

`body { user-select: none }` (`frontend/src/renderer/styles.css:1319-1320`, added by #3508 to stop drags from painting a highlight over sidebar/board chrome) inherits into every descendant. Only `input`/`textarea`/`[contenteditable]` and the terminal (which owns its own selection model) opt back in. Everything else — session details, status text, error messages, PR titles/numbers, branch names, timestamps, activity-timeline copy — is inert to drag-select and `Cmd/Ctrl+C`. It also left the Files-tab diff selection menu (Copy / Explain / Make changes) unreachable, since the diff container never produced a non-empty `window.getSelection()`.

A prior attempt (#4273) fixed this by flipping `body` to selectable and adding an explicit `user-select: none` **opt-out** list for chrome. The maintainer closed it and rejected that shape directly on the issue: *"inverting the global default is the wrong approach... puts the burden on every future control to remember to opt out, and silently regresses the moment one doesn't."* The required direction, per the issue and a follow-up comment: leave `body { user-select: none }` untouched, and opt **in** per content surface with the `select-text` utility class — the pattern already used by `ChatWorkspace.tsx`'s chat transcript and `SessionInspectorView.tsx`'s `ReviewMarkdownBody` (GitHub review markdown).

- `packages/product-ui/src/SessionInspectorView.tsx`: `select-text` on the PR card, the activity-timeline event wrapper, the non-collapsible `ExternalReviewCard` header, the `ReviewSummaryCard` header, and the three inline-comment resolve/send status banners.
- `frontend/src/renderer/components/SessionInspector.tsx`: `select-text` on six error/status banners (auto-inject CI/review policy errors, resume-agent error, merge-policy error, review-request-failed, auto-review-failure).
- `frontend/src/renderer/components/SessionFilesView.tsx`: `select-text` on the diff container, which also unblocks the previously-dead `DiffSelectionMenu`.

Every addition was checked against its actual JSX ancestors to confirm it's outside any onClick/toggle handler before landing — e.g. `ExternalReviewCard`'s collapsible button variant and `InlineCommentRow`'s expand-toggle share the same `headerContent`/body as their safe counterparts, so `select-text` only went on the non-toggle branch. Board session cards are untouched: they already have no `user-select` override of their own and simply keep inheriting `none` from `body`, which is exactly the desired "click-to-open, don't drag-select" behavior.

Out of scope, deliberately: `ReviewDisclosure`'s non-collapsible branch (its only call site never passes `collapsible={false}`, so that branch is dead code today — not worth touching untestable code); `ShellTopbar`'s session title/status (sits inside Electron's window-drag region without the `noDragStyle` opt-out every other topbar control uses, and its crop/button-look issues are already tracked separately in #3173); `InlineCommentRow`'s file-label/comment-preview text (the whole row is a click-to-expand target — making it selectable would let a drag fire the toggle; worth its own follow-up).

Fixes #4268

## Test plan

- [x] `cd packages/product-ui && npm run typecheck && npx vitest run` — 80/80 passing, including 8 new/extended assertions (both that the intended surfaces select, and that collapsible/toggle surfaces explicitly do not)
- [x] `cd frontend && npm run typecheck` — clean
- [x] `cd frontend && npx vitest run` — passing (same pre-existing, unrelated failures as `main`: 14 tests under `src/landing/` fail on a missing `cheerio` dependency in that subtree)
- [x] `cd frontend && npx playwright test text-selection.spec.ts` — new e2e spec, 4/4 passing, stable across repeated runs: confirms `body` stays `user-select: none`, a board card stays unselectable on a drag while its click-to-open still works, and inspector PR content is selectable and round-trips through the OS clipboard
- [x] `cd frontend && npx playwright test` (full suite) — same pre-existing failures as `main` (stale mock-session references, missing terminal-mux fixtures), none touching the files this PR changes
- [x] Manually verified in the desktop app: PR card, activity timeline, and Files-tab diff selection all work and copy correctly; board cards and a collapsible review header stay unselectable with click/expand behavior intact

### DIFF
--- a/packages/product-ui/src/SessionInspectorView.tsx
+++ b/packages/product-ui/src/SessionInspectorView.tsx
@@ -286,7 +286,7 @@ export function InspectorPullRequestCardView({
 	statusNotice?: ReactNode;
 }) {
 	return (
-		<article className="min-w-0 w-full rounded-lg border border-(--color-border-settings-input) bg-(--color-bg-settings-input) px-3 py-2.5">
+		<article className="min-w-0 w-full select-text rounded-lg border border-(--color-border-settings-input) bg-(--color-bg-settings-input) px-3 py-2.5">
 			{pr.title ? (
 				<ExternalLink
 					className="inline text-sm font-semibold leading-snug tracking-tight text-settings-label underline-offset-2 hover:underline focus-visible:rounded-sm focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/60"
@@ -365,7 +365,7 @@ export function InspectorActivityTimelineView({ events }: { events: InspectorTim
 	return (
 		<div className="relative pl-5">
 			{events.map((event, index) => (
-				<div key={index} className="relative pb-4 last:pb-0" data-testid="inspector-timeline-event">
+				<div key={index} className="relative select-text pb-4 last:pb-0" data-testid="inspector-timeline-event">
 					{index < events.length - 1 ? (
 						<span
 							aria-hidden="true"
@@ -997,7 +997,7 @@ function ExternalReviewCard({
 						{headerContent}
 					</button>
 				) : (
-					<div className="grid min-w-0 grid-cols-[auto_minmax(0,1fr)_auto] items-center gap-x-3 gap-y-1 px-1.5 py-1.5 text-left @max-[420px]/inspector:grid-cols-[auto_minmax(0,1fr)] @max-[420px]/inspector:gap-x-2">
+					<div className="grid min-w-0 select-text grid-cols-[auto_minmax(0,1fr)_auto] items-center gap-x-3 gap-y-1 px-1.5 py-1.5 text-left @max-[420px]/inspector:grid-cols-[auto_minmax(0,1fr)] @max-[420px]/inspector:gap-x-2">
 						{headerContent}
 					</div>
 				)}
@@ -1253,9 +1253,9 @@ function InlineCommentRow({
 				</span>
 				{body ? <span data-overflow-axis="horizontal" ref={previewRef} className={cn("mt-1 block min-w-0 text-muted-foreground", expanded ? "whitespace-pre-wrap break-words" : "truncate")}>{expanded ? body : preview}</span> : null}
 			</div>
-			{resolvedSuccess ? <p className="m-0 text-2xs font-medium text-success">{labels.resolvedReview}</p> : null}
-			{resolveError ? <p className="m-0 text-2xs font-medium text-error">{labels.resolveReviewFailed}</p> : null}
-			{sendError && !sent ? <p className="m-0 text-2xs font-medium text-error">{labels.sendToWorkerAgentError}</p> : null}
+			{resolvedSuccess ? <p className="m-0 select-text text-2xs font-medium text-success">{labels.resolvedReview}</p> : null}
+			{resolveError ? <p className="m-0 select-text text-2xs font-medium text-error">{labels.resolveReviewFailed}</p> : null}
+			{sendError && !sent ? <p className="m-0 select-text text-2xs font-medium text-error">{labels.sendToWorkerAgentError}</p> : null}
 		</div>
 	);
 }
@@ -1310,7 +1310,7 @@ function ReviewSummaryCard({
 	useEffect(() => setExpanded(false), [body]);
 	return (
 		<article className="flex min-w-0 flex-col gap-1 rounded-md bg-overlay/50 px-2.5 py-2.5">
-			<span className="grid min-w-0 grid-cols-[minmax(0,1fr)_auto_auto] items-center gap-x-1.5 gap-y-1 @max-[420px]/inspector:grid-cols-[minmax(0,1fr)_auto]">
+			<span className="grid min-w-0 select-text grid-cols-[minmax(0,1fr)_auto_auto] items-center gap-x-1.5 gap-y-1 @max-[420px]/inspector:grid-cols-[minmax(0,1fr)_auto]">
 				<span className="inline-flex min-w-0 items-center gap-1.5">
 					<span className="inline-flex min-w-0 flex-1 items-center gap-1 text-micro font-medium text-muted-foreground">
 						{renderAvatar(actor)}
--- a/packages/product-ui/src/SessionInspectorView.tsx
+++ b/packages/product-ui/src/SessionInspectorView.tsx
@@ -288,21 +288,20 @@ export function InspectorPullRequestCardView({
 	return (
 		<article className="min-w-0 w-full select-text rounded-lg border border-(--color-border-settings-input) bg-(--color-bg-settings-input) px-3 py-2.5">
 			{pr.title ? (
-				<ExternalLink
-					className="inline text-sm font-semibold leading-snug tracking-tight text-settings-label underline-offset-2 hover:underline focus-visible:rounded-sm focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/60"
-					href={pr.href}
-				>
+				<p className="m-0 text-sm font-semibold leading-snug tracking-tight text-settings-label">
 					{pr.title}
-				</ExternalLink>
+				</p>
 			) : null}
 			<div className={cn("flex min-w-0 items-center gap-2", pr.title && "mt-1.5")}>
-				<ExternalLink
-					ariaLabel={openLabel}
-					className="inline-flex min-w-0 items-center gap-1 font-mono text-xs font-medium text-settings-label decoration-muted-foreground underline-offset-2 hover:text-settings-label hover:underline focus-visible:rounded-sm focus-visible:text-settings-label focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/60"
-					href={pr.href}
-				>
+				<span className="inline-flex min-w-0 items-center gap-1 font-mono text-xs font-medium text-settings-label">
 					{pullRequestIcon ?? <GitPullRequestIcon className="size-icon-sm shrink-0" />}
 					<span>PR #{pr.number}</span>
+				</span>
+				<ExternalLink
+					ariaLabel={openLabel}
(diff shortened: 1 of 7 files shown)


RugvedMavidipalli:
@whoisasx would appreciate your review — this addresses your feedback on #4273 by keeping `body { user-select: none }` untouched and opting in per content surface with `select-text` instead.

whoisasx:
@RugvedMavidipalli thank you so much taking it over. would you like to add some screen recording stating before and after the changes.

packages/product-ui/src/SessionInspectorView.tsx
```diff
@@ -271,7 +271,7 @@ export function InspectorPullRequestCardView({
 	statusNotice?: ReactNode;
 }) {
 	return (
-		<article className="min-w-0 w-full rounded-lg border border-(--color-border-settings-input) bg-(--color-bg-settings-input) px-3 py-2.5">
+		<article className="min-w-0 w-full select-text rounded-lg border border-(--color-border-settings-input) bg-(--color-bg-settings-input) px-3 py-2.5">
```
whoisasx, line 289: [P2] Applying `select-text` to the card does not make the advertised PR title/number use case directly selectable, because both values are still rendered as draggable links. Chromium starts a native link drag when selection begins on either anchor, producing an empty `window.getSelection()`; the new E2E explicitly avoids this by selecting the non-link state badge instead. Please render the visible metadata as selectable text with a separate open action, or otherwise guard link navigation during a real selection gesture, and exercise the actual title or PR number in the browser test.

whoisasx (changes requested):
This follows the requested architecture from #4273: the global non-selectable default remains intact, board controls stay protected, and read-only surfaces opt in individually. The diff-container and error/status changes look sound. One advertised core case remains incomplete: PR titles and numbers are links, so a drag beginning on them triggers native link dragging rather than text selection; the current E2E avoids that exact surface. Please address the inline finding before merge.

Verification: all GitHub checks are green; the branch merged cleanly with current main; frontend typecheck passed; focused frontend tests passed 145/145; product-ui tests passed 80/80 and its build passed; Playwright selection tests passed 4/4. The requested before/after desktop recording is also still outstanding.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

whoisasx:
Addressed the requested changes and rebased onto current `main`:

- PR title and `PR #…` metadata are now plain selectable text.
- The external-arrow control remains a separate accessible `Open PR #…` link.
- The browser regression now drag-selects and copies the actual PR title (rather than the state badge).
- Resolved the Files-view rebase conflict by carrying the `select-text` opt-in into the current `WorkspaceDiffView` extraction.

Local verification:
- `packages/product-ui`: typecheck, 119 tests, and build passed.
- Focused frontend tests: 117/117 passed (`SessionInspector` + `WorkspaceDiffView`).
- Frontend typecheck and E2E typecheck passed.
- Playwright `e2e/text-selection.spec.ts`: 4/4 passed, including real PR-title selection and clipboard round-trip.
- AO Browser preview loaded the rebased inspector without page errors.

@whoisasx ready for re-review when convenient.

i-trytoohard:
@RugvedMavidipalli Marking as stale: no activity for 22 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

Issue #4324, opened 2026-08-23
investigate: reported 502 on /api/v1/* has no matching code path in the current daemon


Pulkit7070:
## Status: needs a captured example before it is actionable

502 responses have been reported on `/api/v1/*`, but on investigation **the current daemon HTTP layer contains no code path that emits 502**:
- `rg -n "StatusBadGateway|502|BadGateway|httputil|ReverseProxy"` over `backend/` returns nothing in the request path.
- When the daemon is down/unreachable, the renderer's `runtimeFetch` returns **503** (`daemon_unavailable`), and a direct loopback call fails with `ECONNREFUSED` (a network error) — neither is a 502.

So my original hypothesis ("client hits the loopback before the daemon is ready") is **not supported by the code** and I've removed it.

## Plausible remaining sources (unconfirmed)
- A **dev-only** intermediary (e.g. the Vite dev server proxy in front of `/api/v1`) returning 502 while the daemon restarts.
- An **older daemon version** that had a proxy/handler since removed.
- An intermediary outside this repo.

## What would make this actionable
A concrete captured 502: the exact request path, method, daemon **version**, and whether it was a dev or packaged build. Without that, there is no reproducible cause to fix in the current code.

If nobody can produce a captured example, this should be **closed** as not-reproducible rather than left open.

yashksaini-coder:
@Pulkit7070 I wanna take on this issue

yashksaini-coder:
Investigated. **Your core finding holds: no code path in the Go daemon emits 502, and none ever has.** The reports are real, but every 502 comes from a proxy *in front of* the daemon — which is exactly why grepping `backend/` found nothing.

I found two such proxies. One is dev-only; the other ships and is the more likely source of a user report.

## 1. The Vite dev-server proxy (dev only) — captured

`frontend/vite.renderer.config.ts` proxies `/api` and `/mux` to `AO_DEV_API_TARGET ?? "http://127.0.0.1:3001"`, and `AO_DEV_API_TARGET` is set nowhere in the repo. Vite's proxy error handler is `res.writeHead(502, {"Content-Type":"text/plain"}).end()` (`vite@8.0.16`, `dist/node/chunks/node.js:17938`), so any unreachable target becomes a 502.

Captured on `main` @ `dd47a7547`, dev build, **with a perfectly healthy daemon running**:

```console
$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3002/api/v1/projects
[...]

PR #4446, opened 2026-08-25
fix(dev): point the dev-server proxy at the daemon that is actually running

yashksaini-coder:
## What

Resolve the Vite dev-server proxy target from the run file the daemon actually wrote, instead of a fixed `http://127.0.0.1:3001`.

## Why

Investigation for #4324, where 502s were reported on `/api/v1/*` with no matching code path in the Go daemon. There isn't one — the daemon never emits 502. This proxy does.

`frontend/vite.renderer.config.ts` proxied `/api` and `/mux` to a fixed `127.0.0.1:3001`. That is correct for a standalone daemon (`config.DefaultPort`), but `npm run dev` has Electron supervise its own daemon on `DEV_DAEMON_PORT = 3002` (`frontend/src/main.ts:223`). Vite's proxy error handler is `res.writeHead(502, …).end()` (`vite@8.0.16`, `dist/node/chunks/node.js:17938`), so an unreachable target becomes a 502.

Captured on `main` @ `dd47a7547`, with a healthy daemon running:

```console
$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:3002/api/v1/projects
200                                     # daemon, direct
$ curl -s -o /dev/null -w '%{http_code}' 'http://[::1]:5199/api/v1/projects'
502                                     # same daemon, through the dev proxy
```
```
[vite] http proxy error: /api/v1/projects
Error: connect ECONNREFUSED 127.0.0.1:3001
```

**Not limited to `npm run dev:web`.** The renderer's API base in dev is `window.location.origin` (`renderer/lib/api-client.ts:9,13`) until the daemon reports ready and `renderer/lib/daemon-status.ts:6-11` redirects it to the real port, so plain `npm run dev` has the same exposure for anything requested before that redirect.

Provenance: the `3001` default has been there since the scaffold (`c2c4404c7`, #168); `f2da70d3b` (#2465) later introduced `DEV_DAEMON_PORT = 3002` without touching this config.

## How

`AO_DEV_API_TARGET` still wins outright. Otherwise the resolver walks the run files in the order the daemon's own writers pick one — `AO_RUN_FILE` first (AO agent sessions and isolated checkouts set it, which is why `ao-desktop-dev`'s launch command `env -u`s it), then the Electron dev supervisor's dev-scoped file, then the standalone `ao start` default — and finally falls back to `expectedDaemonPort(env)`, which honours `AO_PORT`.

Deliberately reuses what already exists rather than re-deriving it: `parseRunFile` and `defaultRunFilePath` from `daemon-discovery.ts`, `expectedDaemonPort` and `DEFAULT_DAEMON_PORT` from `daemon-attach.ts`. One parser and one default for this on-disk contract, so there is no second copy to drift when `runfile.Info` changes.

**A recorded port is used only when its pid is still alive.** `running.json` is removed on graceful shutdown only, so a hard-killed `npm run dev` leaves one behind; trusting it would aim the proxy at a dead port and reproduce this same 502 with the ports swapped — a case the fixed default happened to get right. Liveness uses `process.kill(pid, 0)`, matching `main.ts:1259-1265`.

The file is written temp-then-rename (`runfile.Write`), so a reader never observes a partial write; the risk being guarded is staleness, not tearing.

Resolution happens once at config load, stated in the config comment: start the daemon first, or set `AO_DEV_API_TARGET`. Per-request resolution would mean reaching into Vite's proxy internals for a case an env var already covers — happy to go further if you'd prefer it.

## Testing

New `frontend/src/shared/dev-api-target.test.ts`, 15 cases: the `AO_DEV_API_TARGET` override short-circuiting before any file is read, each run-file candidate in precedence order, `AO_RUN_FILE` winning, the standalone fallback, a stale pid being skipped in favour of a live later candidate, `AO_PORT`, and six malformed-run-file shapes that must degrade rather than throw.

```console
$ npx vitest run --config vite.renderer.config.ts        # whole frontend suite
Test Files  219 passed (219)
     Tests  2935 passed | 1 skipped (2936)
$ npm run typecheck
0 errors
```

End-to-end, with a daemon on a non-default port (3007) recorded in a run file: `/api/v1/projects` and `/api/v1/sessions` return **200** through the dev proxy with zero proxy errors, where `main` returns 502.

## Note

The investigation also turned up a second, **shipping** 502 source that this PR does not touch: `tailscale serve --bg --https=443` (`mobilebridge/tailscaleserve.go:36-40`) puts tailscaled's `ReverseProxy` in front of `/api/v1/*` for Connect Mobile, and `ShutdownServe` runs only on the graceful path, so an ungraceful exit strands `:443` on a dead port. Details are on #4324; happy to file it separately.

## Checklist

- [x] Branched from `main` (based on `dd47a7547`)
- [x] Tests added for the behaviour being changed
- [x] Relevant CI checks pass for the area touched (vitest, typecheck)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall:
I reproduced a clean-start problem on this PR head (`4bdf49166`) that appears to leave the primary `npm run dev` case unfixed.

With a fresh isolated `AO_RUN_FILE` and both `AO_PORT` and `AO_DEV_API_TARGET` unset:

1. Forge loads `vite.renderer.config.ts` and starts the renderer Vite server before Electron launches.
2. `resolveDevApiTarget()` runs at config load. Since the supervised daemon and its run file do not exist yet, it falls back to `expectedDaemonPort(env)`, which is `3001`.
3. Electron then launches and starts its dev daemon on `DEV_DAEMON_PORT = 3002`.
4. The proxy target is already fixed for the lifetime of the Vite config, so it remains on `3001` after the `3002` daemon and run file appear.

I verified this dynamically: the Vite proxy resolved to `http://127.0.0.1:3001` in 3/3 clean runs, while the real Electron-supervised daemon reported `daemon listening addr=127.0.0.1:3002`. A request to `/api/v1/projects` through Vite matched the response from port 3001 and did not match the response from port 3002. On my machine another daemon occupied 3001, so this silently reached the wrong AO instance; when 3001 is unused, the same mismatch produces the reported 502.

The fix needs to make the initial Electron dev target 3002 before Vite config loading (while preserving 3001 for `dev:web`), propagate an explicit target into Forge, or otherwise defer proxy target selection. A regression test should exercise a fresh `npm run dev` launch with no existing run file.

There is also a smaller validation gap at `dev-api-target.ts:104-106`: `parseRunFile` maps a missing/non-integer PID to `0`, but the code only rejects a dead PID when `info.pid > 0`. As a result, a run file such as `{"port":4321}` is accepted without calling the liveness check, despite the stated contract that a recorded port is used only when its PID is alive. Requiring a positive PID before accepting the port and adding missing/invalid-PID cases would close that gap.

illegalcall (changes requested):
Requesting changes because both previously reported correctness issues remain present at the latest head (`61d0343e`). The intervening commits merge newer `main`; they do not change the relevant implementation.

## Spec

1. **[P0] Fresh `npm run dev` still locks the proxy to the wrong port.**

   The PR says it should “Resolve the Vite dev-server proxy target from the run file the daemon actually wrote” and notes that plain `npm run dev` has the startup exposure. However, `vite.renderer.config.ts:13` still calls `resolveDevApiTarget()` once while Forge loads Vite, before Electron exists. Electron later creates the window and calls `startDaemon()` (`main.ts:2521-2522`), with its dev daemon on port 3002 (`main.ts:306,1307`). With no pre-existing run file, `dev-api-target.ts:109` falls back to port 3001 for the full Vite lifetime.

   I previously reproduced this dynamically: on a clean isolated launch, Vite resolved 3001 in 3/3 runs while the Electron-supervised daemon listened on 3002. Proxy responses matched the daemon on 3001, not the supervised daemon on 3002. If 3001 has no listener, this produces the reported 502; if it has another AO daemon, requests silently reach the wrong instance.

   Please make the Electron dev target available before Vite config loading, or defer proxy target selection, and add a regression test covering a fresh `npm run dev` launch with no run file.

2. **[P1] Run files with missing or invalid PIDs are still accepted.**

   The PR states: “A recorded port is used only when its pid is still alive.” But `parseRunFile` maps a missing or malformed PID to 0, while `dev-api-target.ts:105-106` only checks liveness when `pid > 0` and then accepts the port. Therefore `{"port":4321}` is trusted without invoking `isPidLive`.

   Please require a positive, live PID before accepting a candidate, and add missing, string, zero, and negative PID test cases.

## Standards

- The missing-PID validation coverage conflicts with the AGENTS.md requirement that tests cover changed validation and missing-input boundaries.
- The new `process.kill(pid, 0)` helper duplicates existing liveness helpers with inconsistent `EPERM` behavior.
- The literal dev run-file path duplicates the path policy already present in `main.ts`, creating drift risk.

The current CI checks pass, but they do not exercise the clean Electron startup ordering above.

yashksaini-coder:
Thanks — both were real, and my first pass at the P0 one didn't actually cover it. I'd put the `VITE_NO_ELECTRON` check on the *fallback*, which only runs when no run file is found, so it never fired: under `npm run dev` the lookup still fell through to `~/.ao/running.json`, which in that mode belongs to the installed app or a CLI `ao start` on `~/.ao/data`. With the app open you'd get a renderer quietly populated from the wrong instance — worse than the 502. The gate is on the candidate list now, same line `main.ts`'s `runFilePath()` already draws. `dev:web` still tries the standalone path first, then the dev file, since it supervises nothing of its own.

P1 is in — `parseRunFile` maps a missing/string/float pid to `0`, so `pid <= 0` drops those before the liveness call ever runs. I pushed that guard into `processIsAlive` as well, since `process.kill(0, 0)` signals the caller's own process group and a negative pid signals the group named by its absolute value, so both would have answered "alive".

On the duplicate helper, you were right that I only removed the copy that already agreed. `main.ts` had two more and both read EPERM as dead, so the supervisor could count a daemon owned by another account as a dead port holder and try to SIGTERM it while the proxy was happily talking to it. All three fold into the shared one now, the `~/.ao/dev` literal sits next to `DEV_STATE_SUBDIR`, and the pid cases and both dev modes have tests.

PR #4451, opened 2026-08-25
fix(systeminstall): pin winget source and add gh direct-download fallback

### LINKED ISSUE #4449: systeminstall: winget installs fail on fresh Windows with 0x8a150042 (msstore source-agreement prompt blocks non-interactive install)
## Summary

On a completely fresh Windows machine (nothing installed — no Node, no Git, no GH CLI, no coding agent), clicking **Install** for `gh` on the "No coding agent found" screen always fails. AO resolves the Windows `gh` plan to a bare `winget install -e --id GitHub.cli` and runs it in a background process with no TTY. On a brand-new Windows profile, winget's `msstore` source has never been initialized for that user, so winget's very first invocation blocks on an interactive Y/N prompt to accept the Microsoft Store source agreements (Terms of Transaction + consent to send the machine's 2-letter region code to Microsoft's backend). Because AO's process has no interactive stdin, the prompt can never be answered and winget aborts with `0x8a150042` (`APPINSTALLER_CLI_ERROR_PROMPT_INPUT_ERROR`).

This is not a permissions or environment problem — it is a fundamental interactivity mismatch between winget's first-run UX and how AO invokes it. Every fresh-Windows user hitting this dialog gets a guaranteed-failing install button.

## Environment

- Windows 11 (fresh install / fresh user profile; winget present but never run before)
- Agent Orchestrator daemon running its toolchain installer (`backend/internal/service/systeminstall`)
- Affected targets on Windows: `gh` (`GitHub.cli`) and `opencode` (`SST.opencode`) — every plan routed through `planWinget`

## Repro steps

1. On a fresh Windows machine (or after deleting `%LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe` to reset winget's per-user state), launch AO with no tools installed.
2. On the "No coding agent found" screen, choose to install **GitHub CLI (gh)**.
3. Watch the job output tail (the daemon captures stdout+stderr of the install command):

## Actual result

```text
The `msstore` source requires that you view the following agreements before using.
Terms of Transaction: https://aka.ms/microsoft-store-terms-of-transaction
The source requires the current machine's 2-letter geographic region to be sent to the backend service...
Do you agree to all the source agreements terms? [Y] Yes [N] No:
An unexpected error occurred while executing the command: 0x8a150042 : Error reading input in prompt
```

Job status flips to `failed`. Retrying reproduces forever until someone interactively accepts the msstore agreements once in a real terminal.

## Expected result

Installing `gh` from a fresh Windows box succeeds without any interactivity, or falls back to a download-based install when winget can't be used reliably.

## Root cause

1. The Windows plan is built in `backend/internal/service/systeminstall/systeminstall.go`:

   ```go
   func (s *Service) planWinget(target Target, id string) Plan {
       ...
       return Plan{Target: target, Command: []string{"winget", "install", "-e", "--id", id}}
   }
   ```

   No `--source`, no `--accept-source-agreements`, no `--accept-package-agreements`. With `-e --id <id>` and no source restriction, winget queries **all** configured sources, including `msstore`, which on a fresh profile triggers the source-agreement consent prompt before the search completes (this matches upstream reports: microsoft/winget-cli#2819 — same prompt text, same `0x8a150042`; see also microsoft/winget-cli#5440 where msstore agreement confirmation fails in other non-interactive contexts).

2. The executor (`backend/internal/adapters/systemexec/systemexec.go`) runs `exec.CommandContext` with stdout/stderr piped to the job's capture buffer and **no stdin attached**, so there is no TTY for winget to read `[Y]/[N]` from. Winget attempts the prompt, fails to read input, exits `0x8A150042` = `APPINSTALLER_CLI_ERROR_PROMPT_INPUT_ERROR` ("Error reading input in prompt", per winget's documented return codes).

3. `--accept-source-agreements` / `--accept-package-agreements` suppress this prompt in most winget versions, but the msstore first-run flow has repeatedly failed in non-interactive sessions even when agreement flags are used (see the upstream threads above), so relying on flags alone is not robust. The deterministic fix is to never touch the msstore source for packages available in the `winget` community source.

Note the irony documented in the same file: `defaultInstallTimeout`'s comment already anticipates "winget waiting on a prompt it'll never get" — this bug is exactly that scenario, except winget fails fast instead of hanging.

## Fix options

1. **Pin the source (minimal, high value):** pass `--source winget` explicitly for every target that exists in the winget community source (`GitHub.cli`, `SST.opencode`), plus `--accept-package-agreements --accept-source-agreements` so no source/package agreement prompt can ever block the non-interactive run. This bypasses msstore entirely.
2. **Pre-accept msstore once:** before any winget install, run a one-time `winget source agree`/equivalent or pipe `Y` into stdin. Fragile: per-user state, region-consent wording varies by winget version/locale, and piping stdin through `ports.CommandRunner` adds plumbing for little benefit given option 1.
3. **Direct-download fallback for gh:** when winget is absent (LTSC/Server, stripped installs) or unreliable, install gh by downloading the official `gh_<ver>_windows_amd64.zip` from `cli/cli` GitHub Releases and extracting `bin\gh.exe` onto the user's PATH. This also removes the "winget was not found on PATH" dead-end on machines that ship without winget.

Suggested direction: **layer 1 + 3** — winget with pinned `--source winget` as the primary path, and a hardcoded PowerShell download-from-releases plan as the fallback when winget isn't available.

vanshx999:
Fixes #4449.

## What happens today

On a completely fresh Windows machine, `planWinget` resolved to a bare `winget install -e --id <id>`. With no source pinned, winget also queries the **msstore** source — which on a fresh profile has never been initialized, so its first use blocks on an interactive Y/N prompt (Terms of Transaction + 2-letter region consent). The daemon runs installs via `systemexec` with no TTY attached, so the prompt can never be answered and winget exits:

```text
Do you agree to all the source agreements terms? [Y] Yes [N] No:
An unexpected error occurred while executing the command: 0x8a150042 : Error reading input in prompt
```

`0x8A150042` = `APPINSTALLER_CLI_ERROR_PROMPT_INPUT_ERROR`. Every fresh-Windows user clicking Install on the "No coding agent found" screen got a guaranteed-failing job, forever retryable, never succeeding. Upstream reference: microsoft/winget-cli#2819 (same prompt, same code path in any non-interactive session).

## The fix (layered options 1 + 3 from the issue)

**1. Pin winget to its community source (`planWinget`)**

```text
winget install -e --id <id> --source winget --accept-package-agreements --accept-source-agreements
```

`--source winget` bypasses msstore entirely for both Windows targets (`GitHub.cli`, `SST.opencode`), and the accept flags make sure no remaining source/package agreement prompt can ever block a headless run. This is deterministic — it doesn't depend on per-user winget state at all.

**3. Direct-download fallback when winget is absent (`planGHWindows`)**

A machine without winget on PATH previously dead-ended as Unsupported ("winget was not found"). Since a truly fresh machine can't be assumed to have winget either, `gh` now falls back to a hardcoded PowerShell script that:

- fetches the latest release metadata from `api.github.com/repos/cli/cli/releases/latest`,
- downloads the official `gh_<ver>_windows_amd64.zip`,
- copies `bin\*.exe` into `%LOCALAPPDATA%\Microsoft\WindowsApps` — already on the default per-user PATH, so the installer's post-run `LookPath("gh")` verification succeeds with no registry edits or shell restarts,
- cleans up staging files.

Same security invariant as the rest of the package: the script is a package-level constant; no caller-supplied string ever reaches the argv. The fallback is deliberately gh-only — `opencode` without winget stays Unsupported (pinned by test).

## Tests

- Updated `TestPlanFor`: winget rows assert the full pinned argv; new row pins that opencode-without-winget remains Unsupported.
- New `TestGHWindowsFallsBackToDirectDownload`: pins that the no-winget gh plan is runnable, non-interactive (`-NoProfile -NonInteractive`), and targets the official cli/cli releases + `WindowsApps` destination.
- `go test ./internal/service/systeminstall/...`, `go vet`, `golangci-lint v2.12.2` all pass.

Note: `TestCrushLocalAuthStatusDoesNotUseProviderCatalog` in `internal/adapters/agent/crush` fails on clean `upstream/main` too (verified by stashing this change) — pre-existing and untouched here.

### DIFF
--- a/backend/internal/service/systeminstall/systeminstall.go
+++ b/backend/internal/service/systeminstall/systeminstall.go
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -271,25 +272,74 @@ func (s *Service) run(plan Plan, job *Job) {
 		job.Error = runErr.Error()
 		return
 	}
-	// Verify the target is now usable. Most installs are confirmed via PATH;
-	// the gh direct-download fallback instead produces a known binary at
-	// InstalledBinaryPath, which we trust directly because its destination may
-	// not yet be on the daemon's PATH on stripped/no-winget Windows (#4449
-	// review).
-	if path, err := s.executables.LookPath(string(job.Target)); err == nil && path != "" {
-		job.Status = StatusSucceeded
+	if err := s.verifyInstalled(plan, string(job.Target)); err != nil {
+		job.Status = StatusFailed
+		job.Error = err.Error()
 		return
 	}
-	if plan.InstalledBinaryPath != "" {
-		if _, statErr := os.Stat(plan.InstalledBinaryPath); statErr == nil {
-			job.Status = StatusSucceeded
+	job.Status = StatusSucceeded
+}
+
+// verifyInstalled reports whether target is usable after its install command
+// finished. The PATH lookup is the primary signal for most targets; the gh
+// direct-download fallback writes to a directory that may not be on this
+// process's PATH yet, so it additionally proves the recorded binary actually
+// executes (not just that a file exists) and registers its directory on the
+// current process PATH so child processes spawned from this already-running
+// daemon inherit it too (#4449 review).
+func (s *Service) verifyInstalled(plan Plan, target string) error {
+	if path, err := s.executables.LookPath(target); err == nil && path != "" {
+		return nil
+	}
+	if plan.InstalledBinaryPath == "" {
+		return fmt.Errorf("install command finished but %s is still not in PATH", target)
+	}
+	if err := s.probeInstalledBinary(plan.InstalledBinaryPath); err != nil {
+		return fmt.Errorf("install command finished but %s is still not on PATH and %s did not run: %w", target, plan.InstalledBinaryPath, err)
+	}
+	registerOnProcessPath(filepath.Dir(plan.InstalledBinaryPath))
+	return nil
+}
+
+// probeInstalledBinary executes the installed binary with a version probe to
+// confirm it is present and actually runs, rather than merely existing on
+// disk. Only the gh fallback sets InstalledBinaryPath today, for which
+// --version is a stable, exit-0 probe; the argv is fixed and never takes
+// caller input.
+func (s *Service) probeInstalledBinary(binPath string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	return s.commands.Run(ctx, []string{binPath, "--version"}, io.Discard, io.Discard) //nolint:gosec // G204: fixed argv from a hardcoded plan path, never caller input
+}
+
+// registerOnProcessPath appends dir to the current process's PATH if it isn't
+// already there, so LookPath and child processes spawned after this point can
+// resolve binaries a just-finished installer dropped into dir. The install
+// script separately persists the entry in the User PATH, but that only applies
+// to processes started after an environment refresh — this updates the running
+// daemon for the rest of its lifetime.
+func registerOnProcessPath(dir string) {
+	if dir == "" {
+		return
+	}
+	pathVal := os.Getenv("PATH")
+	for _, entry := range strings.Split(pathVal, string(os.PathListSeparator)) {
+		if strings.EqualFold(entry, dir) {
 			return
 		}
-		job.Error = fmt.Sprintf("install command finished but %s is still not on PATH and %s does not exist", job.Target, plan.InstalledBinaryPath)
-	} else {
-		job.Error = fmt.Sprintf("install command finished but %s is still not in PATH", job.Target)
 	}
-	job.Status = StatusFailed
+	var updated string
+	if pathVal == "" {
+		updated = dir
+	} else {
+		updated = pathVal + string(os.PathListSeparator) + dir
+	}
+	if err := os.Setenv("PATH", updated); err != nil {
+		// The PATH we hand to children of this process just won't include the
+		// fresh install dir; the job itself is already verified as succeeded via
+		// the binary probe, so this is non-fatal.
+		return
+	}
 }
 
 func displayCommand(plan Plan) string {
@@ -406,10 +456,13 @@ Remove-Item -Force "$stage.zip"
 if (-not (Test-Path (Join-Path $dest 'gh.exe'))) { throw 'gh.exe was not found after extraction; install failed' }
 # Register the destination on the user PATH so future sessions and child shells
 # (e.g. terminals AO spawns for the agent) can find gh even when WindowsApps is
-# absent from the default PATH on stripped/no-winget images.
+# absent from the default PATH on stripped/no-winget images. The User-scope Path
+# can be unset on a fresh/stripped profile, so coalesce it to '' before use.
 $userPath = [Environment]::GetEnvironmentVariable('Path', 'User')
+if (-not $userPath) { $userPath = '' }
 if ($userPath -notlike "*$dest*") {
-    [Environment]::SetEnvironmentVariable('Path', ($userPath.TrimEnd(';') + ';' + $dest), 'User')
+    $newPath = if ($userPath -eq '') { $dest } else { $userPath.TrimEnd(';') + ';' + $dest }
+    [Environment]::SetEnvironmentVariable('Path', $newPath, 'User')
 }
 Write-Output ("installed gh " + $release.tag_name + " to " + $dest)`
 
--- a/backend/internal/service/systeminstall/systeminstall.go
+++ b/backend/internal/service/systeminstall/systeminstall.go
@@ -343,7 +343,7 @@ func (s *Service) planTmux() Plan {
 func (s *Service) planGH() Plan {
 	switch s.goos {
 	case "windows":
-		return s.planWinget(TargetGH, "GitHub.cli")
+		return s.planGHWindows()
 	case "darwin":
 		return s.planBrew(TargetGH, "gh")
 	case "linux":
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -358,6 +358,46 @@ func (s *Service) planGH() Plan {
...
+// argv in this package; no caller-supplied string ever reaches it.
+const ghWindowsDownloadScript = `$ErrorActionPreference = 'Stop'
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+$release = Invoke-RestMethod -Uri 'https://api.github.com/repos/cli/cli/releases/latest'
+$asset = $release.assets | Where-Object { $_.name -like 'gh_*_windows_amd64.zip' } | Select-Object -First 1
+if (-not $asset) { throw 'no gh windows_amd64 asset found in the latest cli/cli release' }
+$stage = Join-Path ([IO.Path]::GetTempPath()) ('ao-gh-' + [Guid]::NewGuid().ToString('N'))
+Invoke-WebRequest -Uri $asset.browser_download_url -OutFile "$stage.zip"
+Expand-Archive -Path "$stage.zip" -DestinationPath $stage -Force
+$dest = Join-Path $env:LOCALAPPDATA 'Microsoft\WindowsApps'
```
somewherelostt, line 449: This fallback copies `gh` into `%LOCALAPPDATA%\Microsoft\WindowsApps`, then the shared runner verifies success with `LookPath("gh")`. That only works if `WindowsApps` is already present in the daemon's PATH. On stripped/no-winget Windows setups, `WindowsApps` may also be absent, so the install can complete and still be reported as failed. Consider verifying the concrete copied `gh.exe` path, or updating the process/PATH handling before the post-install check.

vanshx999:
Addressed in ce4abfb7e. You're right — relying on `LookPath("gh")` after dropping into WindowsApps was fragile, and exactly the stripped/no-winget environments that hit the fallback are the ones where WindowsApps may be absent from the daemon's PATH.

Changes:
- `Plan` now carries `InstalledBinaryPath`. `run()` accepts success if that exact binary exists on disk, independent of PATH — so a successful install is no longer reported as failed.
- The PowerShell script self-verifies `gh.exe` landed and `throw`s (job fails) if it didn't, so a real extraction/copy failure is still surfaced instead of silently succeeding.
- The script also registers the destination on the user PATH (`[Environment]::SetEnvironmentVariable`, User scope) so future sessions and any child shells AO spawns for the agent can find gh even when WindowsApps isn't on the default PATH.
- `planGHWindows` records the concrete `gh.exe` path it writes.

Tests added: `TestRun_SucceedsOnConcreteInstalledBinaryPath` (LookPath misses but binary present -> Succeeded) and `TestRun_FailsWhenConcreteBinaryMissing` (neither PATH nor the recorded path -> Failed, error names the missing path). `go vet`, the full systeminstall suite, and golangci-lint v2.12.2 are clean.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -261,12 +271,25 @@ func (s *Service) run(argv []string, job *Job) {
...
+	// InstalledBinaryPath, which we trust directly because its destination may
+	// not yet be on the daemon's PATH on stripped/no-winget Windows (#4449
+	// review).
+	if path, err := s.executables.LookPath(string(job.Target)); err == nil && path != "" {
+		job.Status = StatusSucceeded
 		return
 	}
-	job.Status = StatusSucceeded
+	if plan.InstalledBinaryPath != "" {
+		if _, statErr := os.Stat(plan.InstalledBinaryPath); statErr == nil {
```
somewherelostt: This still marks the install as succeeded when `gh.exe` merely exists at `InstalledBinaryPath`, even if the daemon still cannot resolve/run `gh`. On stripped/no-winget Windows where `WindowsApps` is absent from the current process PATH, writing the User PATH value only helps future processes after the environment is refreshed; child shells spawned from this already-running process can still inherit the old PATH. Please either verify by executing the installed binary, or update the current process environment before reporting success.

backend/internal/service/systeminstall/systeminstall.go
```diff
@@ -358,6 +381,60 @@ func (s *Service) planGH() Plan {
...
+Remove-Item -Force "$stage.zip"
+# Self-verify: a missing binary means the extraction/copy failed, and the job
+# must be reported as failed rather than silently reporting success.
+if (-not (Test-Path (Join-Path $dest 'gh.exe'))) { throw 'gh.exe was not found after extraction; install failed' }
+# Register the destination on the user PATH so future sessions and child shells
+# (e.g. terminals AO spawns for the agent) can find gh even when WindowsApps is
+# absent from the default PATH on stripped/no-winget images.
+$userPath = [Environment]::GetEnvironmentVariable('Path', 'User')
+if ($userPath -notlike "*$dest*") {
+    [Environment]::SetEnvironmentVariable('Path', ($userPath.TrimEnd(';') + ';' + $dest), 'User')
```
somewherelostt: This can still fail on a fresh/stripped Windows profile where the user-level `Path` variable is unset. `[Environment]::GetEnvironmentVariable('Path', 'User')` returns `$null` in that case, so `$userPath.TrimEnd(';')` throws after `gh.exe` has already been copied and the fallback reports failure. Please coalesce `$userPath` to an empty string and handle the empty-path append case before writing the updated User PATH.

backend/internal/service/systeminstall/systeminstall.go
vanshx999: Addressed in f57df4d. Verification no longer trusts file existence — `run()` now **executes** the installed binary (`gh.exe --version`) via the command runner before reporting success, so the daemon can't claim success for a binary it cannot run. And `registerOnProcessPath` appends the install dir to the **current process** PATH (not just the User registry value), so child shells spawned from this already-running daemon inherit the updated PATH too.

backend/internal/service/systeminstall/systeminstall.go
vanshx999: Addressed in f57df4d. The script now coalesces the User-scope Path to `''` before appending: `if (-not $userPath) { $userPath = '' }`, and the append handles the empty case (`$newPath = if ($userPath -eq '') { $dest } else { $userPath.TrimEnd(';') + ';' + $dest }`). No more throw on a fresh/stripped profile where the User Path is unset.

i-trytoohard:
@vanshx999 Marking as stale: no activity for 20 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4459, opened 2026-08-25
feat: add durable scheduled automations

Vaibhaav-Tiwari:
## What

Adds Phase C durable scheduled automations across the Go daemon, SQLite storage, HTTP API, CLI, and Electron supervisor.

- Creates, lists, updates, enables/disables, deletes, and inspects automation run history.
- Supports daily, weekly, cron-subset, and custom recurring schedules in named IANA timezones.
- Materializes durable occurrences with bounded catch-up, leases, non-overlap, restart reconciliation, and session linkage.
- Adds the AO-styled Automations route, creation dialog, inline validation, keyboard-friendly time entry, run history, toggles, and destructive confirmation.
- Includes the detailed implementation plan/design documents used for this phase.

## Why

AO needs unattended recurring sessions that survive daemon restarts, preserve run history, respect local wall-clock time across DST, and launch through the same session boundary as manual work.

## How

- SQLite definitions and run rows are the durable source of truth; unique occurrence keys and transactional claims prevent duplicate dispatch.
- A daemon observer reconciles startup state, projects session completion, materializes at most three missed occurrences, and dispatches eligible runs.
- Session launch completion is persisted separately from the automation origin marker. Crash-surviving seed rows are never treated as successful launches; incomplete launches are torn down after lease expiry and recorded as failed.
- The packaged daemon embeds IANA timezone data. Finite COUNT/UNTIL rules, conflicting embedded TZIDs, and schedules capable of sub-minute execution are rejected explicitly.
- Latest run summaries are loaded in one batched storage query; expanded run history polls and displays fetch failures.
- Frontend strings use AO's locale catalogs, and the shell/sidebar route identity follows existing desktop patterns.

Intentional scope: finite RRULEs are rejected because Phase C models recurring definitions with a non-null future cursor. The cron compatibility surface is deliberately limited to minute/hour plus optional weekday expressions rather than silently translating unsupported cron semantics.

## Testing

Passed after rebasing onto current upstream main:

- `go test ./... -run '^$'`
- `go test ./internal/service/automation ./internal/storage/sqlite/store ./internal/httpd ./internal/cli ./internal/observe/automations ./internal/storage/sqlite ./internal/lifecycle -count=1`
- `go test ./internal/session_manager -run 'TestSeedRecordPreservesAutomationRunIdentity|TestSpawnAutomation' -count=1`
- `npx vitest run src/renderer/components/AutomationsView.test.tsx src/renderer/i18n/renderer-coverage.test.ts src/renderer/i18n/instance.test.ts src/renderer/components/ShellTopbar.linux.test.tsx src/renderer/components/ShellTopbar.test.tsx` (61 tests)
- `npm run sqlc`
- `git diff --check`

Also manually exercised in the real AO Electron app from the isolated worktree: navigation, creation, validation, keyboard time entry, scheduling, enable/disable, run history, session opening, and deletion.

The repository-wide runtime suites were audited on Windows as well. Their remaining failures are pre-existing platform/environment-sensitive macOS, POSIX-shell, symlink-privilege, optional-dependency, and timing cases outside this diff; all touched packages and a compile-only pass across every Go package are green.

## Checklist


Vaibhaav-Tiwari:
image
image
image
image
image

i-trytoohard:
@Vaibhaav-Tiwari Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

Vaibhaav-Tiwari:
image

backend/internal/service/automation/scheduler.go
```diff
@@ -0,0 +1,286 @@
...
+		if run.Status == domain.AutomationRunSpawning {
+			session, ok, lookupErr := store.GetSessionByAutomationRunID(ctx, run.ID)
+			if lookupErr != nil {
+				problems = append(problems, lookupErr)
+			} else if ok && session.AutomationLaunchCompleted {
+				_, lookupErr = store.MarkAutomationRunRunning(ctx, run.ID, session.ID, now)
+				if lookupErr != nil {
+					problems = append(problems, lookupErr)
+				}
+			} else if ok && run.LeaseExpiresAt != nil && !run.LeaseExpiresAt.After(now) {
```
illegalcall: [P1] If the daemon restarts before a `spawning` lease expires, startup reconciliation skips it. After expiry, this branch has no `!ok` recovery when no session was created; the run stays `spawning` and the claim query blocks every later run of this automation. I reproduced the stuck state. Please release expired no-session claims during regular ticks too.

backend/internal/session_manager/manager.go
```diff
@@ -988,6 +1002,18 @@ func (m *Manager) Spawn(ctx context.Context, cfg ports.SpawnConfig) (domain.Sess
 		m.rollbackSpawnSeedRowAfterFailure(ctx, id)
 		return domain.SessionRecord{}, 0, 0, wrapSpawnStage(id, ErrWorkspaceCreate, err)
 	}
+	if cfg.AutomationRunID != nil {
```
illegalcall: [P1] Persisting workspace path and prompt before provisioning means a provisioning failure cannot pass `SessionIsSeed` during rollback. The retained terminated session has `AutomationLaunchCompleted=false`; the scheduler releases its run to pending, but every retry finds this incomplete session and fails again, even after provisioning is fixed. I reproduced this sequence. Please ensure failed prelaunch rows are removed so this run can retry.

backend/internal/lifecycle/manager.go
```diff
@@ -1523,6 +1523,9 @@ func (m *Manager) markSpawned(
 		// a stale "signals worked once" fact.
 		rec.FirstSignalAt = time.Time{}
 		rec.Metadata = mergeMetadata(rec.Metadata, metadata)
+		if rec.AutomationRunID != nil {
+			rec.AutomationLaunchCompleted = true
```
illegalcall: [P1] This marks an automation launch complete before `PromptDeliveryAfterStart` runs. If delivery fails, rollback terminates the session without clearing this marker; retry adopts that terminal row as a successful spawn and the scheduler then marks the run completed although the prompt was never delivered. I reproduced the failed-delivery and retry path. Please mark completion only after delivery succeeds, or clear/reject it on failure.

frontend/src/renderer/components/AutomationsView.tsx
```diff
@@ -0,0 +1,630 @@
...
+// cannot reproduce stays on the custom RRULE field.
+function scheduleFieldsFromRRule(rruleText: string) {
+	const rruleLine = rruleText.split("\n").find((line) => line.startsWith("RRULE:"))?.slice("RRULE:".length) ?? rruleText;
+	const freq = rruleLine.match(/FREQ=([^;]+)/)?.[1];
+	const hour = rruleLine.match(/BYHOUR=(\d{1,2})/)?.[1];
+	const minute = rruleLine.match(/BYMINUTE=(\d{1,2})/)?.[1];
+	const byDay = rruleLine.match(/BYDAY=([^;]+)/)?.[1];
+	if (hour !== undefined && minute !== undefined) {
+		const time = `${hour.padStart(2, "0")}:${minute.padStart(2, "0")}`;
+		if (freq === "DAILY" && !byDay) return { preset: "daily", time, raw: rruleLine };
```
illegalcall: [P1] This also classifies `FREQ=DAILY;INTERVAL=2;BYHOUR=9;BYMINUTE=0` as the basic daily preset. The edit form always submits a fresh RRULE, so changing only the name silently drops `INTERVAL=2` and its stored DTSTART anchor. Please preserve the original rule unless the schedule changed, and map to a preset only when it can reproduce the complete rule.

backend/internal/service/automation/schedule.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	rule, err := rrule.NewRRule(*option)
+	if err != nil {
+		return Schedule{}, fmt.Errorf("build rrule: %w", err)
+	}
+	next := rule.After(now, false)
+	if next.IsZero() {
+		return Schedule{}, fmt.Errorf("schedule has no future occurrence")
+	}
+	following := rule.After(next, false)
```
illegalcall: [P2] Checking only the first adjacent pair misses later sub-minute gaps. At `2026-09-24 09:00:15Z`, `DTSTART:20260923T090000Z` with `FREQ=DAILY;BYHOUR=9;BYMINUTE=0;BYSECOND=0,30` yields next runs at 09:00:30, then next-day 09:00:00 (the checked gap passes), then 09:00:30 (30s later). This violates the documented one-minute minimum while passing validation.

frontend/src/renderer/components/AutomationsView.tsx
```diff
@@ -0,0 +1,630 @@
...
+									setHarness(value);
+									setModel("");
+									setModelTouched(false);
+								}}
+							/>
+							<Field label={t("automations.model")}>
+								<AutomationModelPicker
+									agentId={selectedHarness}
+									agentLabel={selectedAgentLabel}
+									value={selectedModel}
```
illegalcall: [P2] Picking a model updates only local form state: `submit()` omits `model`, and the automation create/update DTOs have no model field. The visible choice therefore has no effect on scheduled sessions. Please remove this control until model selection is supported, or wire it through storage and spawn.

frontend/src/renderer/hooks/useAutomations.ts
```diff
@@ -0,0 +1,81 @@
...
+export type CreateAutomationInput = components["schemas"]["CreateAutomationRequest"];
+export type UpdateAutomationInput = components["schemas"]["UpdateAutomationRequest"];
+
+export const automationsQueryKey = ["automations"] as const;
+
+export function useAutomations() {
+	return useQuery({
+		queryKey: automationsQueryKey,
+		queryFn: async () => {
+			const { data, error } = await apiClient.GET("/api/v1/automations", { params: { query: { limit: 100 } } });
```
illegalcall: [P2] This discards `nextCursor` after the first 100 definitions. The SQL list orders oldest-first, so automation #101 appears after the optimistic create but disappears on the 15-second refetch, with no way to reach it in the desktop. Please paginate or fetch subsequent pages.

backend/internal/cli/root.go
```diff
@@ -221,6 +221,7 @@ func NewRootCommand(deps Deps) *cobra.Command {
 	root.AddCommand(newOrchestratorCommand(ctx))
 	root.AddCommand(newPRCommand(ctx))
 	root.AddCommand(newReviewCommand(ctx))
+	root.AddCommand(newAutomationCommand(ctx))
```
illegalcall: [P2] `ao automation` is absent from the embedded `using-ao/SKILL.md` command table, and there is no `commands/automation.md`. `docs/documentation-map.md` requires a matching command page when CLI behavior changes. Please add both so installed agents can discover and use the new command.

backend/internal/cli/automation_test.go
```diff
@@ -0,0 +1,67 @@
...
+	var body map[string]any
+	if err := json.Unmarshal(capture.body, &body); err != nil {
+		t.Fatal(err)
+	}
+	if body["projectId"] != "demo" || body["displayName"] != "Morning" || body["cron"] != "0 9 * * *" {
+		t.Fatalf("body=%s", capture.body)
+	}
+}
+
+func TestAutomationDeleteRequiresMatchingConfirmation(t *testing.T) {
```
illegalcall: [P3] This covers only a rejected confirmation. Please add the successful confirmation path and assert that it sends DELETE; the repo's `AGENTS.md` requires tests for destructive confirmation, and this positive branch is otherwise unchecked.

illegalcall (changes requested):
Requesting changes for the confirmed automation correctness issues in the inline comments. The crash and launch failure paths can strand runs or report work as completed when its prompt was never delivered; a name-only edit can also alter the recurrence. Please cover those failure sequences with regression tests, and address the smaller validation, desktop, and CLI gaps noted inline.

PR #4460, opened 2026-08-25
fix(runtime): recover owned legacy tmux sessions after desktop update

### LINKED ISSUE #4458: bug(update): tmux namespace changes falsely exit sessions and spawn duplicate controllers
## Bug

Updating to the private hashed tmux runtime makes every pre-cutover session on tmux's legacy default socket render as **exited** when opened, even though the durable session is non-terminated/idle and the original tmux pane and supervised agent process are still alive.

This is a false terminal exit, not an agent exit. The daemon's `/mux` reports `type:"exited"` because it probes only the new private socket, and the renderer converts that frame into `[process exited]` plus the ended strip.

**Source:** live local report/triage by @AgentWrapper on 2026-08-25 (no public chat URL)
**Installed app:** `0.12.8-nightly.202608251410`
**Bundled daemon VCS revision:** `0b7758c7f19e34f00d92bb2eb306f565fa571bf4` (`vcs.modified=true`)
**Daemon:** healthy on `127.0.0.1:3001`
**Analyzed through current main:** `f3d64ab84a384fdf9ba4cbc2b992c7815af9e1f7`
**Confidence:** High
**Priority:** P1 — existing live terminal sessions become inaccessible from the desktop after an ordinary update; Restore risks creating/replacing runtime state and is not a safe workaround.

## Exact live evidence

Seven existing sessions reproduce identically:

| Session | Display name | DB state | DB handle | Legacy pane |
|---|---|---|---|---|
| `agent-orchestrator-124` | scm pr attribution | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 23243 |
| `agent-orchestrator-147` | pr discovery nudge | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 24095 |
| `agent-orchestrator-149` | pr author avatar | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 26773 |
| `agent-orchestrator-151` | network err banner | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 26311 |
| `agent-orchestrator-152` | tab hit area fix | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 24924 |
| `agent-orchestrator-153` | readme download cta | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 25736 |
| `agent-orchestrator-154` | stacked pr no review | `idle`, `is_terminated=0` | same as session id | `pane_dead=0`, PID 25472 |

The private socket is:

```text
~/.ao/tmux-c30770fb243592f92df80569b4463182.sock
```

It contains only the newly-created `agent-orchestrator-211`. The legacy default server contains all seven sessions.

For every affected id:

```text
bundled tmux -S ~/.ao/tmux-c30770fb243592f92df80569b4463182.sock has-session -t <id>
=> can't find session: <id>

system tmux -L default has-session -t <id>
=> exit 0
```

A read-only mux open against the installed daemon reproduces the desktop result without reaching or resizing the live legacy pane:

```text
client: {"ch":"terminal","id":"agent-orchestrator-124","type":"open","cols":0,"rows":0}
server: {"ch":"terminal","id":"agent-orchestrator-124","type":"exited"}
```

The stored `runtime_launch_id` also differs from the supervisor `--launch` visible in every live legacy pane:

| Session | DB launch | Live pane launch |
|---|---|---|
| 124 | `96f06e1e-ca42-4693-adff-f176e08f89a7` | `084a9650-8d7f-4497-9877-05342ce237d3` |
| 147 | `9139a6af-fc05-47cd-9d4f-d4aa03ac1f4b` | `b9ea8bd0-434a-4003-91db-a768c603df5c` |
| 149 | `755e6cde-39eb-4973-adfb-39ce60552d6e` | `d34c593c-96c4-4dff-be3b-4997e701a953` |
| 151 | `bec6d60d-f22c-4792-86c2-852b62cad4eb` | `4087d827-87f5-4a30-9633-8c97b74050a8` |
| 152 | `0b326a64-30c0-4815-ad6b-cdb878d5ee08` | `b4b0bcc3-95ae-47c5-a070-e5b909c8e662` |
| 153 | `114a3d64-1399-49bb-8478-5d0ec23f6e56` | `bf9ee0ec-5d39-4c73-8a94-0a5e546f8086` |
| 154 | `79ef446b-e9d4-4903-93e4-2e1ab7dbd103` | `f8809754-10ee-437c-8983-36475a58aa46` |

That generation mismatch means an adoption/migration policy cannot rely on the current DB launch id alone. The pane name, worktree cwd, AO supervisor command/session id, and historical socket provenance still agree.

## Root cause

[PR #4385](https://github.com/Untrivial-ai/agent-orchestrator/pull/4385) explicitly covered this upgrade path. It added `socketForSession`, probed the new socket first, then used the system tmux client with explicit `-L default` for legacy sessions, cached the chosen socket, and routed `IsAlive`, attach, output, input, restart, and destroy consistently.

[PR #4393](https://github.com/Untrivial-ai/agent-orchestrator/pull/4393) intentionally removed that mixed-socket routing in commit `2aeba22c0` as a strict security cutover. The PR says legacy sessions “remain running but are not adopted.” This creates the user-visible regression:

1. `runtimeselect.New` constructs one tmux adapter with only the hashed private `SocketPath`.
2. `tmux.managedArgs` pins **every** command to `-S <private-socket> -f /dev/null`.
3. `terminal.attachment` calls `IsAlive`; a missing same-named session on that private server returns `alive=false`.
4. The mux emits `type:"exited"` immediately, without checking the durable `is_terminated` fact or the legacy server.
5. `useTerminalSession` treats that frame as an actual process exit, writes `[process exited]`, and `TerminalPane` shows the ended strip.

There are no later terminal/runtime commits between the installed revision and current `main` that restore the legacy upgrade path.

## Logs and request correlation

The current packaged daemon does not persist its active stdout/stderr logs: both are Unix pipes back to Electron. `~/.ao/daemon.log` is stale (last modified 2026-06-17), so it contains no evidence for this 2026-08-25 update. macOS unified logging also returned no matching AO terminal/tmux entries.

`telemetry_event` contains no `ao.renderer.terminal_attach_failed` for this path because an `exited` frame is treated as a successful lifecycle outcome, not an attach failure. `/mux` frames carry no request id, and the successful session GET response exposes no request-id header. This leaves the false-exit attach without a correlatable persisted request id; the raw mux request/response above is the strongest direct reproduction.

## Expected behavior

An update must preserve safe terminal access to AO-owned legacy sessions while ensuring new sessions use only the hashed private socket. If ownership cannot be proven, the UI must report an explicit “legacy runtime not adopted / recovery required” error—not `process exited`—and must not offer Restore as though the pane were dead.

## Fix constraints

Do not blindly re-enable same-name adoption from the user default socket; that would regress the ownership issue fixed by [#4391](https://github.com/Untrivial-ai/agent-orchestrator/issues/4391). A safe migration needs explicit AO provenance/ownership and a durable socket classification, or a one-time controlled migration that does not destroy or respawn the live pane.

Until that policy exists, the minimum correctness fix is to make “absent from the new private socket during a known cutover” inconclusive/recovery-required rather than emitting `exited` for a durable non-terminated session.

## Required regression coverage

1. Start multiple AO sessions on the historical tmux default socket with live panes and distinct launch ids.
2. Persist their handles as non-terminated sessions, including a realistic stale/different DB `runtime_launch_id`.
3. Upgrade/restart into a build using the hashed private socket.
4. Assert legacy sessions are not killed, restarted, resized, or sent input during reconciliation.
5. Assert `/mux` attach to an owned legacy session returns `opened` and terminal data (or an explicit recovery-required error), never `exited` while `pane_dead=0`.
6. Spawn a new post-update session; assert it exists only on the hashed private socket and attaches there.
7. Keep both socket classes live across a second desktop restart and assert routing remains stable.
8. Add same-name collision coverage on default + private sockets and prove AO never adopts or mutates a foreign tmux session.
9. Cover missing/incompatible system tmux as inconclusive without session termination or replacement.
10. Cover a real legacy pane exit and verify that only this case produces the mux `exited` frame.

## Safe recovery guidance

Do **not** press Restore, Kill, Cleanup, or send/nudge these sessions from the affected build. Their worktrees and tmux panes are still live.

For read-only inspection, `system tmux -L default capture-pane -p -t <session>` preserves the pane. Interactive `tmux -L default attach -t <session>` bypasses AO but can resize the pane and accepts input, so it is not a read-only diagnostic.

A build based on the post-[#4385](https://github.com/Untrivial-ai/agent-orchestrator/pull/4385), pre-[#4393](https://github.com/Untrivial-ai/agent-orchestrator/pull/4393) behavior can reattach through `-L default`, but changing installed builds/restarting should only be done after preserving state and validating that startup reconciliation will not respawn or replace these handles.

## Related

- [#4382](https://github.com/Untrivial-ai/agent-orchestrator/issues/4382) — bundled tmux terminfo regression; similar post-update attach symptom, different root cause.
- [PR #4385](https://github.com/Untrivial-ai/agent-orchestrator/pull/4385) — implemented the exact legacy-default + private-socket mixed upgrade path, later removed.
- [#4391](https://github.com/Untrivial-ai/agent-orchestrator/issues/4391) / [PR #4393](https://github.com/Untrivial-ai/agent-orchestrator/pull/4393) — ownership isolation and the intentional strict cutover that introduced this regression.
- [#4417](https://github.com/Untrivial-ai/agent-orchestrator/issues/4417) — stale staged tmux binary can also break private-socket attachment, but does not explain these sessions being present and live only on `-L default`.

AgentWrapper:
Fixes #4458

Summary:
- keep the hashed AO-private tmux socket authoritative for every new session
- recover pre-cutover default-socket sessions only after immutable pane provenance proves the AO run file, session id, supervised marker, and launch generation
- route attach, capture, input, restart, liveness, process inspection, environment sync, and teardown through the selected socket
- repair stale runtime and agent-session launch generations during boot with compare-and-swap fencing
- fail closed for foreign same-name sessions and prefer a concurrently appearing private session

Upgrade behavior:
Users already on the regressed nightly and users updating directly from pre-private-socket releases get the same automatic boot reconciliation. Live legacy panes remain on the legacy server; no process or worktree migration is required. New sessions stay on the hashed private socket.

Tests:
- npm run lint
- cd backend && go build ./...
- race-enabled focused runtime/lifecycle/session reconciliation tests
- isolated real-tmux integration covering legacy -L default plus the new hashed -S private socket across an update boundary

Risk controls:
The integration test isolates the default server through TMUX_TMPDIR. No production sessions or user tmux servers are touched.

### DIFF
--- a/backend/internal/adapters/runtime/runtimeselect/runtimeselect.go
+++ b/backend/internal/adapters/runtime/runtimeselect/runtimeselect.go
@@ -41,7 +41,10 @@ var _ Runtime = (*conpty.Runtime)(nil)
 // terminal runtime state.
 func New(_ *slog.Logger, runFilePath string) Runtime {
 	if runtime.GOOS != "windows" {
-		return tmux.New(tmux.Options{SocketPath: privateTmuxSocketPath(runFilePath)})
+		return tmux.New(tmux.Options{
+			SocketPath:  privateTmuxSocketPath(runFilePath),
+			RunFilePath: runFilePath,
+		})
 	}
 	return conpty.New(conpty.Options{RunFilePath: runFilePath})
 }
--- a/backend/internal/adapters/runtime/tmux/commands.go
+++ b/backend/internal/adapters/runtime/tmux/commands.go
@@ -99,6 +99,14 @@ func listPanePIDsArgs(id string) []string {
 	return []string{"list-panes", "-s", "-t", exactSessionTarget(id), "-F", "#{pane_pid}"}
 }
 
+// paneStartCommandsArgs returns the immutable command tmux recorded for every
+// pane in a session. Legacy-socket adoption uses it as provenance: sessions
+// created by AO before the private-socket cutover carry the AO run-file,
+// session id, supervised-process marker, and supervisor launch id here.
+func paneStartCommandsArgs(id string) []string {
+	return []string{"list-panes", "-s", "-t", exactSessionTarget(id), "-F", "#{pane_start_command}"}
+}
+
 // sendKeysLiteralArgs builds args for `tmux send-keys -t <id> -l <chunk>`.
 // The -l flag stops tmux interpreting words like "Enter" as key names so the
 // text is sent verbatim.
--- a/backend/internal/adapters/runtime/tmux/environment.go
+++ b/backend/internal/adapters/runtime/tmux/environment.go
@@ -33,12 +33,12 @@ func (r *Runtime) syncCurrentEnvironment(ctx context.Context, sessionID string, 
 		{scope: "session", args: []string{"show-environment", "-t", exactSessionTarget(sessionID)}},
 	}
 	for _, query := range queries {
-		out, err := r.run(ctx, query.args...)
+		out, err := r.runForSession(ctx, sessionID, query.args...)
 		if err != nil {
 			if serverUnreachableOutput(string(out)) {
-				return fmt.Errorf("private tmux server unavailable while refreshing session %s", sessionID)
+				return fmt.Errorf("tmux server unavailable while refreshing session %s", sessionID)
 			}
-			return fmt.Errorf("inspect private tmux %s environment %s: %w", query.scope, sessionID, err)
+			return fmt.Errorf("inspect tmux %s environment %s: %w", query.scope, sessionID, err)
 		}
 		for _, key := range parseTmuxEnvironmentNames(string(out)) {
 			names[key] = struct{}{}
@@ -49,8 +49,8 @@ func (r *Runtime) syncCurrentEnvironment(ctx context.Context, sessionID string, 
 	if script == "" {
 		return nil
 	}
-	if _, err := r.runWithInput(ctx, []byte(script), "source-file", "-"); err != nil {
-		return fmt.Errorf("apply private tmux environment: %w", err)
+	if _, err := r.runWithInputForSession(ctx, sessionID, []byte(script), "source-file", "-"); err != nil {
+		return fmt.Errorf("apply tmux session environment: %w", err)
 	}
 	return nil
 }
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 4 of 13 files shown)


AgentWrapper:
Superseded by the user-authorized pure revert in #4465, merged to `main` as `c4abb8afc4a305fbd20c6d3955a2c7808d78e7d4`. The larger migration/adoption approach can be reconsidered separately after the immediate nightly containment release.

PR #4462, opened 2026-08-25
fix(updates): never resolve a prerelease update through a draft release

Pulkit7070:
Stacked on #4397. Base is `ao/ao-41/sidebar-update-action`, so this diff is just the one commit; retarget to `main` once #4397 merges.

## The bug, observed in production

GitHub's `releases.atom` lists **draft** releases, and electron-updater discovers prereleases through that feed. Draft assets are not public, so the channel manifest 404s. On the automatic path that error is deliberately swallowed, so the install silently goes stale against a release the user could never have downloaded.

Checked against the live feed on 2026-08-25, the newest atom entry was the draft `v0.12.8-nightly.202608251410`:

```
nightly-mac.yml -> 404
latest-mac.yml  -> 404   (electron-updater's fallback)
```

Meanwhile `v0.12.7-nightly.202608250031` was published, complete and downloadable. Every nightly install was aiming at the draft and reporting nothing.

## What this changes

The nightly channel already sidesteps the atom feed by resolving the newest completed release through the API, which filters `draft: false` and requires the platform manifest. A **pinned feature build did not**: `configureFeed` sets `allowPrerelease` for `pr<N>` and leaves discovery to the atom feed, so a draft `pr<N>` build strands the pin the same way.

This generalizes the resolver from `"nightly"` to any prerelease channel and routes feature pins through it. Stable is untouched: it resolves via `/releases/latest`, which excludes drafts and prereleases by definition, so it takes no extra request.

## Tests

- A pinned feature build skips a draft `pr<N>` release and resolves the newest published one, and the pin still outranks a newer nightly. **Confirmed to fail without the production change.**
- A draft nightly is skipped in favour of the newest published release. This one **passes without the change** and is labelled in the source as a regression guard rather than new behaviour, since nightly draft filtering already existed. It is pinned because the atom feed really does list drafts, making this filter the only thing between a prerelease install and an undownloadable release.

65/65 `auto-updater.test.ts`, 162/162 across the update-related files, `frontend:typecheck` clean.

## Scope

This is the **client half only**. It stops a draft from stranding an install. It does not stop drafts appearing in the feed, and it cannot fix a release pipeline that publishes a build and later returns it to draft.

Two things seen while investigating that this PR does not address, and that no workflow in this repository appears to own (there is no nightly release workflow here):

- A **draft build was downloaded onto a real machine** (`0.12.8-nightly.202608251410`, 168MB, staged 2026-08-25 20:03) while the release is a draft now and its assets 404. That implies the build was published and later returned to draft.
- Two **published** nightlies, `202608241447` and `202608242034`, carry **no mac manifest at all**, so they 404 the same way for any client that selects them.

Both look like release-pipeline problems and probably deserve their own issue.

### DIFF
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -327,6 +327,157 @@ describe("startAutoUpdates", () => {
     }
   });
 
+  // Regression guard, not new behaviour: the nightly resolver already filtered
+  // drafts. It is pinned here because releases.atom really does list drafts
+  // (observed in production on 2026-08-25), so this filter is now the only
+  // thing standing between a prerelease install and an undownloadable release.
+  it("skips a draft release, which the Atom feed lists but nobody can download", async () => {
+    const platformManifest =
+      process.platform === "darwin"
+        ? "nightly-mac.yml"
+        : process.platform === "linux"
+          ? "nightly-linux.yml"
+          : "nightly.yml";
+    const resourcesPath = mkdtempSync(
+      nodePath.join(os.tmpdir(), "ao-nightly-feed-"),
+    );
+    writeFileSync(
+      nodePath.join(resourcesPath, "app-update.yml"),
+      "provider: github\nowner: Untrivial-ai\nrepo: agent-orchestrator\n",
+    );
+    const originalResourcesPath = Object.getOwnPropertyDescriptor(
+      process,
+      "resourcesPath",
+    );
+    Object.defineProperty(process, "resourcesPath", {
+      configurable: true,
+      value: resourcesPath,
+    });
+    // A draft carrying a complete asset set is still unreachable: its download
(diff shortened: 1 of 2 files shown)


i-trytoohard:
@Pulkit7070 Marking as stale: no activity for 22 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

Pulkit7070:
Closing as obsolete: the prerelease draft guard this adds is already on main via #4812 (https://github.com/Untrivial-ai/agent-orchestrator/pull/4812), which generalized the nightly draft filter to all prerelease and feature/PR feeds. Main's fetchLatestCompletedPrereleaseTag already filters !release.draft and directPrereleaseChannel already resolves feature builds, and it carries the release-notes body and ETag caching on top. Nothing left to merge here.

PR #4464, opened 2026-08-25
ci: guard the nightly feed head the way stable is guarded

Pulkit7070:
## Why

`release-latest-guard.yml` protects the stable channel from a `/releases/latest` that electron-updater cannot install from. Nightly had no equivalent, and nightly is the channel that actually breaks. Of the six nightlies cut between 2026-08-23 and 2026-08-25:

| Release | State | Manifests |
| --- | --- | --- |
| `v0.12.8-nightly.202608251410` | draft | **missing** |
| `v0.12.8-nightly.202608250759` | draft | ok |
| `v0.12.7-nightly.202608250031` | published | ok |
| `v0.12.7-nightly.202608242034` | **published** | **missing** |
| `v0.12.7-nightly.202608241447` | **published** | **missing** |
| `v0.12.7-nightly.202608241408` | published | ok |

Any one of these at the head of the feed strands **every** nightly install.

electron-updater discovers prereleases through `releases.atom`, takes the **first** entry matching the channel, fetches that release's manifest, and on 404 retries exactly one fallback before giving up. It never walks to the next entry. So one bad release at the top wedges every client even with a good release directly beneath it — which is the current state:

```
feed head:      v0.12.8-nightly.202608251410
  nightly-mac.yml -> 404
  latest-mac.yml  -> 404   (the single fallback)
  => ERR_UPDATER_CHANNEL_FILE_NOT_FOUND, gives up

one entry down: v0.12.7-nightly.202608250031
  nightly-mac.yml -> 200   (never reached)
```

The failure is invisible. Automatic check errors are suppressed by design, so affected installs show nothing at all and simply stop updating. On one machine this had been going on for two days across an app restart.

## What it does

Fails when the nightly feed head is a draft, or is missing `nightly.yml` / `nightly-mac.yml` / `nightly-linux.yml`, and says what to do about it.

It reads `releases.atom`, not the API, on purpose: the API omits drafts, the Atom feed lists them, and a draft's assets are not public, so a draft at the head 404s for everyone. Checking the API would miss exactly the case that broke this week.

Same shape as the stable guard: hourly schedule, release triggers, and the `github.repository` condition so forks do not fail red.

## Verified against the live feed

Ran the check logic against real releases:

- `v0.12.8-nightly.202608251410` (draft, no manifests) → **fails**, reports the draft
- `v0.12.7-nightly.202608242034` (published, no manifests) → **fails**, names the three missing files
- `v0.12.7-nightly.202608250031` → passes
- `v0.12.7-nightly.202608241408` → passes

So it catches both real failure modes and does not fire on healthy releases. Note this guard will fail on `main` until the current feed head is fixed — that is the point, not a defect in the workflow.

## Scope

This detects the breakage; it does not prevent it. Nightlies are cut by hand (`frontend-nightly.yml` was removed in #3070) and the manifest generation is a separate step from the artifact publish, which is what makes it easy to skip. Automating the nightly cut so the manifests cannot be forgotten is the real fix and is a larger change than this.

Related: #4397 and #4462 harden the client so it skips drafts and manifest-less releases and resolves to the newest installable one. This guards the release side; those guard the client side.

### DIFF
--- a/.github/workflows/release-nightly-guard.yml
+++ b/.github/workflows/release-nightly-guard.yml
@@ -1,0 +1,77 @@
+name: Release nightly guard
+
+# The stable channel has release-latest-guard.yml. Nightly had nothing, and it
+# is the channel that actually breaks: three of the six nightlies cut between
+# 2026-08-23 and 2026-08-25 shipped without their updater manifests, and one of
+# those was still a draft. Any of them strands every nightly install.
+#
+# Why the newest release is the only one that matters: electron-updater
+# discovers prereleases through releases.atom, takes the FIRST entry matching
+# the channel, fetches that release's <channel>-<platform>.yml, and on 404
+# retries exactly one fallback before giving up. It never walks to the next
+# entry. So one bad release at the top of the feed wedges every client, even
+# with a perfectly good release directly beneath it.
+#
+# Drafts count. releases.atom lists them, but their assets are not public, so a
+# draft at the top of the feed 404s for everyone.
+on:
+  release:
+    types: [published, edited, released, created, prereleased]
+  schedule:
+    - cron: "23 * * * *"
+  workflow_dispatch:
+
+jobs:
+  nightly-feed:
+    # Forks copy this workflow but cut no nightlies, so the hourly schedule
+    # would fail red forever. Same guard as release-latest-guard.yml.
+    if: github.repository == 'Untrivial-ai/agent-orchestrator'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Verify the nightly feed head is installable
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


harshitsinghbhandari:
New production evidence is tracked in [#4487](https://github.com/Untrivial-ai/agent-orchestrator/issues/4487).

On 2026-08-26 at 14:27 UTC, packaged nightly `0.12.8-nightly.202608250759` showed the three-failure `checksFailing` warning while the feed head `v0.12.8-nightly.202608252327` was published without any updater manifests. The next nightly uploaded `nightly-mac.yml` at 14:31:41 UTC and published at 14:31:48 UTC, so this occurrence self-healed four minutes after the screenshot. Current stable `v0.12.7` also lacks all `latest*.yml` assets and its macOS manifest endpoint still returns 404.

This independently validates the guard in this PR and raises the operational follow-up: backfill stable and make release publication atomic/complete before discovery.

PR #4419, opened 2026-08-25
test(chatui): add opt-in local regression harness

illegalcall:
Code changes: +0 -0  
Tests: +4750 -0  
Others: +554 -2  

## Summary

- adds 19 deterministic Playwright contracts covering MQA-01 through MQA-12, plus three runner-enforced Go boundary contracts
- adds an opt-in local runner with strict and known-red capture modes, timestamped screenshots/video/traces, and evidence validation
- isolates the harness from CI: no workflow invokes it, its TypeScript project is local-only, normal Vitest excludes its contracts, harness-only Go files require the `chatui_regression` build tag, and the runner selects the exact backend contract names
- updates MQA-01 to require AO-owned Claude permission-mode naming and truthful Plan-mode disclosure; provider-owned copy remains deliberately adversarial so the product override is tested
- keeps the controlled desktop bridge current with the `cloudCp` bridge method and daemon settings endpoint
- aligns the documented direct Go command with all three configured backend packages and exact contract names
- deliberately excludes the original QA reports, screenshots, and generated captures from git

## Local usage

```bash
npm run qa:chatui -- --capture  # record the current known-red baseline
npm run qa:chatui               # strict certification after fixes
npm run qa:chatui -- --grep MQA-06
```

## Verification

- exact harness head: `0e648d84384347afb5579944a60fb4e3a805b565`
- exact base: `a5c12e706f35c99712b066ba931130fc195a6753`
- exact #4433 product head used for strict MQA-01 integration: `aa5a63e9ad08696a2d074c3af620feea2ef29279`
- combined #4419 + #4433: `AO_CHATUI_E2E_PORT=5201 npm run qa:chatui -- --grep MQA-01` — 1/1 passed under the strict runner
- `npm --prefix frontend run test:chatui-runner` — 28/28 passed
- `npm --prefix frontend run typecheck` — passed
- `npm --prefix frontend run typecheck:e2e` — passed and does not include ChatUI harness sources
- `npm --prefix frontend run typecheck:chatui` — passed
- documented three-package Go command — MQA-07 and MQA-08 pass; MQA-06 truthfully reproduces its known-red product baseline until its separate fix is integrated
- full renderer run — 3,023 passed, 1 skipped, with one unrelated timing failure in `GlobalSettingsForm`; the exact isolated file then passed 30/30
- `git diff --check` — passed
- independent Standards review — approved exact head, no actionable findings
- independent Spec review — approved exact head, no actionable findings

[...]

PR #4463, opened 2026-08-25
fix(chat): persist drafts across lifecycle boundaries

illegalcall:
Code changes: +5096 -398  
Tests: +5764 -113  
Others: +765 -8  

## Replaced by a focused stack

Kept as draft while the replacement stack is reviewed. Review and merge the replacement PRs in this order:

- #5105 — terminal disposal prerequisite
- #5106 — persist composer text across lifecycle boundaries
- #5107 — restore staged draft attachments after restart
- #5109 — persist queued edits with atomic delivery receipts
- #5110 — recover inline edits with durable branch receipts
- #5111 — reconcile steer delivery without repeating provider actions

Each replacement contains its own validation and screenshot/recording links. The split starts from main `09effe55b`, preserves current interface-switch behavior, fixes shared attachment ownership and queued `/compact`, and omits the SQLC boolean override already present on main. The original PR head is unchanged; the notes below describe its earlier combined implementation.

## Current review follow-up

Queue edits and image sends now recover from lost responses without losing attachments or applying an edit twice. The current head is `fbe433623be661b6ef98594763016123cecc364b`, merged cleanly with `main` at `a47db0e067b9da107ee18c68e27ab1919d77ca3c`.

- Commit queued-edit receipts atomically with revision changes (new additive migration **0131**). Reuse the same ID and exact payload after reload, dispatch, or controller teardown; preserve the original native-image mode when capabilities disappear.
- Restore ordinary-send image bytes under the saved delivery ID, reserving the durable draft before asynchronous file reads.
- Let durable recovery records survive navigation, reload, close, and quit without a discard warning. Failed local persistence and unfinished attachment writes still require confirmation.
- Translate safety/recovery notices and native close/quit dialogs in all eight locales, including language changes after the view mounts.

Current-head local validation: **290 frontend files; 4,091 tests passed, 6 intentional skips; 57/57 renderer smoke; strict MQA-04 passed**. The full Go race suite passed, including SQLite migrations (829.3s) and storage (520.5s). Go build/vet, native CLI E2E, golangci-lint v2.12.2, frontend/E2E typechecks, API/SQLC drift, cloud-client/product-ui checks, and a supplemental local secret scan also passed. The lost-response browser regression commits the edit, advances the daemon revision, drops the response, reloads, and verifies identical recovery without a second revision change. Independent review findings have been addressed and rechecked.

[...]

Issue #4425, opened 2026-08-25
bug(chat): imported Terminal turns show misleading OUTCOME UNKNOWN


illegalcall:
## Bug

Completed Claude Terminal/TUI turns are imported into ChatUI with the visible divider **OUTCOME UNKNOWN**. This reproduced for both a simple `TERMINAL_OK` reply and a drained busy turn that visibly completed `BUSY_DONE`.

**Source:** [ChatUI QA and regression harness PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419)  
**Reported by:** Dhruv Sharma via local native-Electron QA  
**Analyzed against:** `d71acca1e13ddd9ab3a01a43f029853686f4c693`; presentation path rechecked on `04b676c9abd2fac4505081b0c099e079a30bde57`  
**Environment:** macOS arm64, Claude native TUI → ChatUI  
**Priority:** P2 — handoff succeeds but reports materially misleading provenance  
**Confidence:** High for presentation; Medium for provider-native status availability

## Reproduction

1. In Claude Terminal UI, submit a prompt that returns a distinctive successful answer.
2. Let the turn finish, then drain-switch to ChatUI.
3. Observe the imported prompt and assistant response followed by **OUTCOME UNKNOWN**.

## Expected

Preserve exact native completion state when the provider supplies it. When it does not, say neutrally: **Imported from Terminal UI — completion status unavailable**. Do not visually imply that the visible exchange itself may have failed.

## Actual

Every recovered ACP turn is rendered with the generic unknown-outcome divider.

## Root cause

- `backend/internal/adapters/chatdriver/acp/history.go:67-78,169-175,245-262` closes ACP replay boundaries as `TurnStateRecovered` because `session/load` has no terminal outcome.
- Codex can preserve native status in `backend/internal/adapters/chatdriver/codexappserver/history.go:116-127,183-188`.
- `frontend/src/renderer/components/chat/ChatTimelineItems.tsx:2235-2253` maps every recovered turn to **Outcome unknown** without import provenance.

## Fix direction

Carry status source/provenance with replayed turns. Map exact provider state when present; otherwise render neutral import copy. Never infer success only from assistant text.

## Acceptance

- Exact native completion state is retained where available.
- Unavailable state uses neutral Terminal-import copy.
- Failed/interrupted native turns remain distinguishable.
- `npm run qa:chatui -- --grep MQA-08` passes in strict mode.

## Issue-specific evidence

[...]

PR #4466, opened 2026-08-26
fix(chat): make history recovery trust-aware

### LINKED ISSUE #4424: bug(chat): unscoped hook checkpoints permanently block TUI-to-Chat recovery
## Bug

After **Stop now and switch** interrupts a busy Claude TUI turn, target Chat startup can fail `TARGET_HISTORY_UNSETTLED`. In the reproduced session, five retries—including a successful recovery turn, daemon/Electron restart, and native-identity refresh—continued to fail. The session remained usable only in Terminal, and the AO-2L interface-failure notice offered Dismiss but no targeted recovery.

**Source:** [ChatUI QA and regression harness PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419)  
**Reported by:** Dhruv Sharma via local native-Electron QA  
**Analyzed against:** `d71acca1e13ddd9ab3a01a43f029853686f4c693`; source paths rechecked and tagged regression remains red on `04b676c9abd2fac4505081b0c099e079a30bde57`  
**Environment:** macOS arm64, source-built AO, verified daemon on port 3001  
**Priority:** P0 by QA definition / repository P1 label — a normal interface flow permanently removes ChatUI for a healthy session  
**Confidence:** High — extended native reproduction plus failing session-manager contract

## Reproduction

1. Start a long command in native Claude TUI.
2. Choose **Stop now and switch** to Chat.
3. Observe `TARGET_HISTORY_UNSETTLED` / AO-2L “switch interface failed.”
4. Retry while idle.
5. Complete a new successful native turn and retry.
6. Restart daemon and Electron, refresh native identity, and retry.
7. Every Chat startup remains blocked.

## Expected

A fresh, generation-fenced provider-history observation can recover transient state. If an untrusted legacy text checkpoint remains poisoned, AO offers explicit provider-history recovery while retaining trusted AO high-water/terminal-history safety. Terminal must remain usable.

## Actual

Stale/unscoped text facts become hard replay checkpoints and survive every normal recovery boundary; the renderer provides no actionable recovery.

## Root cause

- `backend/internal/cli/hooks.go:188,325` extracts prompt/assistant aliases across hook events without main/subagent provenance.
- `backend/internal/lifecycle/manager.go:1280` preserves older non-empty facts indefinitely.
- `backend/internal/service/chat/service.go:240` promotes those facts into hard replay checkpoints.
- `backend/internal/service/chat/controller.go:421` requires exact latest-completed text matches.
- `frontend/src/renderer/components/SessionInterfaceSwitch.tsx:276` gives `TARGET_HISTORY_UNSETTLED` no recovery action.

## Fix direction

Qualify checkpoint provenance by provider, native conversation, controller generation, event kind, main/subagent source, and trust. Retry only refreshable snapshots. Return typed mismatch dimensions. Permit explicit legacy/untrusted provider-history recovery without weakening trusted current-generation gates. Add Retry / Use provider history / Stay in Terminal actions and an announced failure state.

## Acceptance

- A first incomplete target snapshot can recover using a fresh observation.
- Stale/subagent/old-generation hook text cannot poison a new target.
- Trusted current-provider mismatches still fail closed.
- Recovery never duplicates provider transcript or mutates the worktree.
- AO-2L exposes actionable, accessible recovery.
- `npm run qa:chatui -- --grep MQA-06` and both tagged Go contracts pass strictly.

## Issue-specific evidence

Real authenticated Claude Code TUI → Chat recovery in the isolated AO Electron app:

1. AO-2L fails closed and offers accessible recovery choices:

   Native AO-2L recovery controls

2. **Use provider history** completes the handoff and replays the original turn once:

   Provider-history recovery completed

3. The recovered ChatUI remains usable for a new Claude turn:

   Recovered ChatUI accepts a new turn

The screenshots are GitHub-hosted issue evidence. Local accessibility snapshots, manifests, and general QA reports remain outside the repository and uncommitted.

## Related

- [#4120](https://github.com/Untrivial-ai/agent-orchestrator/issues/4120) / [PR #4121](https://github.com/Untrivial-ai/agent-orchestrator/pull/4121) — similar permanent symptom from a different AO high-water cause
- [#4123](https://github.com/Untrivial-ai/agent-orchestrator/issues/4123) / [PR #4186](https://github.com/Untrivial-ai/agent-orchestrator/pull/4186) — removed blind polling of immutable ACP replay
- [PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419) — browser and Go MQA-06 contracts

illegalcall:
Code changes: +2636 -601  
Tests: +6373 -252  
Others: +567 -324  

## Review recheck — 2026-09-14

[Latest commit `8cd59c309`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8cd59c309c619fa19e0c5a11fc846ed6fb7f6fc2) fixes one remaining startup blocker: a live agent-switch target with a missing/mismatched native conversation identity already had a persisted recovery marker, but still returned an ordinary error and aborted daemon startup. Successfully persisted identity ambiguity now quarantines **only that session**. Database read/write failures still fail startup; explicit recovery still reports unresolved ownership. No controller is started, adopted, or destroyed by this correction.

**Evidence:** six live-target identity cases failed before the fix (missing reference/row, wrong session/harness/generation, empty native ID). All 18 startup-matrix cases pass afterward, including marker-write failure and registry-read failure before/after an existing marker. The affected session stays input-fenced; healthy sessions remain usable. Full local session-manager suite passes (`31.746s`). Independent Standards and Spec reviews of this two-file correction found no actionable findings.

The supplied review referenced older commits. Desktop restart polling, rolled-back anchors, equal-timestamp queue boundaries, account-operation deferral, and missing revision-trigger validation were already pushed; their relevant tests were rerun successfully. Unchanged `dropUnsettledHookFacts` internals are not proof of a defect: it intentionally cannot retire trusted TUI facts using matching old interrupted Chat text. Safe interrupted-native-Claude recovery still needs proven native interruption provenance; this correction does not claim to solve that separate limitation.

**Validation status:** **15/15 fresh CI checks passed on `8cd59c309`**. [Full Go race suite](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844380129/job/103976642507): session manager `208.627s`, Chat `301.732s`, SQLite `721.807s`. [Frontend](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844380059) and [cross-platform CLI](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844379870) also passed. GitHub reports **APPROVED / MERGEABLE / CLEAN**. Local broad build/full-backend/race validation hit `no space left on device`; these are not local passes. The full affected non-race suite, focused startup/explicit recovery checks, 13 checkpoint/index tests, revision-trigger regression, and 19 desktop hook tests passed. No new native Electron screenshots are claimed. Historical CI/screenshots below apply only to their named commits. No reports, screenshots, or runtime data are committed.

## Previous-head conflict and CI evidence — 2026-09-14

[Previously verified commit `1ea5af15c`](https://github.com/Untrivial-ai/agent-orchestrator/commit/1ea5af15c3af309c361b3ca4896d143384b608c5) includes main `e1e35936f`. GitHub reports **MERGEABLE** and **APPROVED**. Fresh checks: [Go](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34841082075), [Frontend](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34841082150), [CLI E2E](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34841082035). **All 15/15 checks passed on this commit.** The PR is conflict-free and approved.

- Preserve main's #5305 GET/POST handoff-readiness checks and all new regressions, with this PR's explicit strict history policy.
- Preserve main's #5348 live-reconnect activity/recency behavior alongside this PR's unpublished-controller cleanup; adapt the four checkpoint-test call sites and SQLite lifecycle test adapter to the new interfaces.
- Earlier merge preserved standalone nullable project ownership, renumbered only this PR's unmerged migrations to **0141–0145**, and left main's shipped migration 0140 unchanged.
- **CI failure:** the SQLite migration package hit its cumulative **15-minute timeout**; the last test had only been running eight seconds. It was not an assertion failure or a reported data race. Increase the bounded race-test timeout to **20 minutes**; no tests, assertions, or race coverage are removed.

Validation:
- **Passed in CI on `1ea5af15c`: all 15 checks**, including full backend race tests, build/vet, lint, Windows workspace tests, native CLI E2E on Linux/macOS/Windows, container smoke, renderer smoke, and generated-code drift. The SQLite migration package completed in **920.329s (15m20s)**, confirming that the old 15-minute budget was insufficient and the 20-minute bound allows the complete suite to finish. Frontend: **314 test files, 4,601 passed, 6 existing skips**.
- **Passed on the combined revision:** full local Chat package tests (29.1s); focused interface-status, poisoned-checkpoint resume, and reconnect tests (1.3s); formatting/diff checks.
- The timed-out `TestStandaloneMigrationConvertsLegacyScratchOwnership` passed alone with `-race` in **13.4s** before the second main merge.
- **Local full-suite gap:** full SQLite race and session-manager suite attempts ran out of disk space. They are not claimed as passing; the full clean-runner CI result above now passes.
- Earlier `9edcccc13` checks: migration/standalone/revision/checkpoint tests, frontend typecheck, and 171 focused frontend tests passed. Its full local frontend run had 4,531 passed, 23 failed and 6 skipped plus six suite-load failures, explained by Electron/Node `better-sqlite3` ABI mismatch and missing landing dependencies. Remote frontend CI on that commit passed.
- No new native Electron QA claim for these merges. Screenshots below retain their original revisions. No evidence, screenshots, reports, or runtime data committed.

## Latest fix — Claude queued/delayed Stop identity

[Commit `c6f7c1967`](https://github.com/Untrivial-ai/agent-orchestrator/commit/c6f7c1967e6b6e392c3b63c885fbfba753265343) stays in this PR. **Queued completed-turn handoff is now natively verified. This is not blanket merge approval:** interrupted native turns, marker-write recovery, and the other outstanding cases in the historical record remain unresolved. Current-head CI has started; previous-head green checks are not this commit's result.

**Problem:** Claude `UserPromptSubmit` for queued B reports executing A's `prompt_id`. Stop A then arrives before Stop B. Matching only that ID—or the latest prompt text—can certify B's prompt with A's answer, permanently blocking TUI → Chat.

**Fix:** give each submission a UUID via the prompt hook's `additionalContext`, and persist that same UUID with the activity observation. Claude stores the hook attachment under the actual native user UUID. After the source stops, the Claude adapter verifies all retained submissions/Stops against native transcript ancestry. Chat admission requires the latest exact native user UUID and its answer in completed ACP replay. It retains AO's durable history boundary and explicit consent for legacy/untrusted text. Claude Stop alone no longer marks the display prompt complete.

This adds a small per-prompt context marker and a durable evidence column. It is not invisible/free context. Missing native identities, unsupported/compacted ancestry, old trusted checkpoints without witnesses, pending/interrupted turns, or evidence overflow **fail closed**, not silently repaired. Limits: 512 observations, 64 MiB transcript, 4 MiB record. Hook reporting still does not guarantee delivery during persistent failures.

### Native evidence — what to observe

Real Electron, macOS arm64, Claude Code **2.1.270**, isolated AO data and disposable todo-app worktrees. No database or provider-event injection for these before/after flows. Different fresh sessions are used; this does **not** show repairing an already-poisoned pre-upgrade session.

1. **Before, `6ae6e7b91`:** queue B while A streams; both finish. TUI → Chat fails with `unsettled_turn_boundary`. The attempted Escape did not interrupt A; this is a queued-input case, not proof of interrupted-turn recovery.

Before: both native answers finished but handoff rejected

2. **After, implementation in `c6f7c1967`:** repeat the queue flow. Strict handoff imports `NATIVE_PROOF_QUEUED_B` once; a fresh `NATIVE_PROOF_CHAT_C` response then completes. The recovered-turn label is expected and is not the bug.

After: queued Terminal reply retained and new Chat reply completes

3. **Round trip:** Chat → TUI retains B/C; a new Terminal D completes, then returns to Chat once. The second switch hit the separate false-draft gate from #5322, which is not included in this build. **Explicit Discard draft and switch was required** despite an empty composer; no provider-history bypass was used. This is a passing history-boundary check, not a clean all-gates-pass claim.

<details>
<summary>Terminal D and subsequent Chat import</summary>

Terminal retains B/C and completes D

Second Chat return retains C and imports D

</details>

### Validation of this delta

- Full backend `go test -p 2 ./...`, build, vet: pass. The first full run caught obsolete hook-only-completion test expectations and a missing migration-ledger entry; corrected, then full rerun passed. Unchanged packages may use test cache.
- Focused race regressions: pass for native evidence/admission, queued/repeated/missing boundaries, nonce retry, and pane fallback/CAS. **Not** a full current-head race-suite claim.
- Pinned golangci-lint v2.12.2: **0 issues**. sqlc/API regeneration: zero drift. Frontend typecheck and **43 interface-transition tests**: pass.
- Independent **Standards: 0; Spec: 0 outstanding findings** in this delta after corrections. Both reviewers independently reran focused tests.
- A hook-ID-only candidate passed focused tests but failed native QA and was discarded. Native QA also exposed unrelated string-valued transcript attachments; the parser correction has a regression test.
- No fresh Codex/mobile matrix, complete interruption recovery, or exhaustive provider-outage verification is claimed.

All screenshots are GitHub attachments. Reports, raw traces, runtime data and QA scripts remain local, never committed. No CI workflow or opt-in harness trigger changed.

<details>
<summary>Previous PR description, historical evidence and remaining follow-ups (retain original revisions)</summary>

**QA status — September 13, head `6ae6e7b91`: NOT READY TO MERGE.** The requested main integration is now conflict-free against `c1538a9a2`; the known Claude identity/interruption and other correctness gaps below remain open. Current-head CI is **14/15 green**, including the full frontend suite (311 files), renderer smoke and all native CLI jobs; only the backend race job is still running. Broad local validation stopped at disk exhaustion, as detailed below.

## PR-comment follow-up

- The old untagged-launch and missed-prompt/Stop findings have fixes and freshly passing lifecycle, Chat and transition regressions.
- The September 11 cancelled-prompt case is covered by `TestInterfaceHandoffRecoversLegacyCheckpointFromUnsettledChatTurn`: a hidden cancelled legacy prompt requires explicit provider-history consent; repeated recovery preserves history once. The trusted new-TUI-prompt control also passes. This does not claim that native busy-stop recovery is fixed.
- [`917c1ee8f`](https://github.com/Untrivial-ai/agent-orchestrator/commit/917c1ee8f) merges current main. The one test-file conflict preserves main's cloud mock and this PR's real transition helpers. No new recovery behavior or migration renumbering was needed.
- [`6ae6e7b91`](https://github.com/Untrivial-ai/agent-orchestrator/commit/6ae6e7b91) is lint-only (+3/-2): explicit existing return-value discards, with errors still logged; retain the intentional top-level-only quarantine check so joined infrastructure failures are not suppressed. Runtime marker-persistence recovery is still unresolved.
- Merge/lint-delta reviews: **Standards 0; Spec 0 actionable findings**, scoped to these two commits, not the entire PR.

Local validation: full backend non-race/build/vet pass; pinned lint 0 issues; API/sqlc regeneration has zero drift; 146 SessionView/transition tests, 788 mobile tests and mobile typecheck pass; cloud build/vet/tests and both shared-package checks pass. Frontend and E2E typechecks pass under Node 24.14.0; 13 native-helper configuration tests pass. The first backend run had an HTTP test failure; that package then passed 20 consecutive runs and the complete backend rerun passed without product changes. Renderer smoke first passed 58/59 with one click-stability timeout; all four terminal-focus cases passed three repeats (12/12). The first CLI E2E attempt failed during temporary-directory cleanup. Broad frontend/race reruns, dependency installation and CLI recheck were then blocked/stopped by disk exhaustion (`ENOSPC`); they are **not claimed locally passed**. Partial dependency installs from this worktree were removed, not evidence or source. [Remote full frontend (311 files), renderer smoke and mac-helper jobs](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34754281889) pass, as do native CLI on all three OSes, container and mobile. Only the [backend race job](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34754281947/job/103715911530) remains in progress. Local Go is 1.26.5; CI uses the workflow-pinned toolchain. No fresh native Electron/provider screenshots are claimed.

## Previous follow-up at `8da6b72dc`

Six commits addressed the supplied follow-up review. Their details and explicitly unresolved work are retained below.

## Follow-up review — what changed

| Area | Fix and regression evidence |
| --- | --- |
| Desktop polling | `TARGET_STOP_UNCONFIRMED` stops 250 ms progress polling. Hook test reproduced 3 requests instead of 1; manual status refresh still works. |
| Startup quarantine | An interrupted interface transition defers behind account recovery without aborting boot or dropping its input/reaper fence. Recovery targets only the captured transition ID and respects worker shutdown. Persisted source/target agent-switch ambiguity quarantines that session; discovery/marker-write errors remain errors. Cancellation racing deferred recovery releases only a durably cancelled, source-intact transition. |
| Replay ordering | Rolled-back turns cannot become anchors. Tied queued turns use replayable evidence and require each tied boundary's own item/answer, not merely a completion marker. AO-only activities cannot displace native evidence. Earlier-provider turns with user/activity sequence evidence—and equal-time itemless rows—are excluded at coordination cutover. |
| Revision integrity | Startup rejects a DB with `sessions.revision` but no `sessions_revision_update` trigger. Existing trigger mechanics/migrations were not rewritten. |
| Hook contention | Exhausted uncommitted CAS returns `503 ACTIVITY_PROJECTION_BUSY`, still records independent usage, and allows the CLI to retry the identical payload up to four requests. Generic 503/transport errors are not retried because their commit outcome may be ambiguous. Persistent contention still fails visibly into `hooks.log`; this is **not durable guaranteed delivery**. |
| Remaining session timestamp fences | Activity-observer snapshots and usage finalization now use `sessions.revision`. Same-timestamp writes previously let a stale observer overwrite a fresh prompt and stale usage finalization close a live binding; both regressions now reject those writes. Revision `0` is fenced; only an absent observer expectation is unversioned. |

Commits: [polling](https://github.com/Untrivial-ai/agent-orchestrator/commit/bb6d3eb25), [trigger verification](https://github.com/Untrivial-ai/agent-orchestrator/commit/6945f7b62), [replay boundaries](https://github.com/Untrivial-ai/agent-orchestrator/commit/7fcfa1e1d), [startup fencing](https://github.com/Untrivial-ai/agent-orchestrator/commit/9f2a27868), [hook redelivery](https://github.com/Untrivial-ai/agent-orchestrator/commit/7c4ce73e2), [observer/usage revision fences](https://github.com/Untrivial-ai/agent-orchestrator/commit/8da6b72dc).

**Local validation:** `go build ./...` and `go test -p 2 ./... -timeout 25m` pass at this head (the latter reuses successful package results). Initial default-concurrency full run failed the fake-agent timing, Pi hook-order, and LAN identity tests; all three passed on rerun without code changes. Full affected-package race suites pass: Chat, session manager, lifecycle, HTTP controllers and CLI at `7c4ce73e2`; lifecycle, activity observer, usage collector, SQLite store and session manager after the final revision conversion. The revision-trigger race regression also passes. Affected Go vet passes. Frontend typecheck and **43 tests in 2 files** pass; no frontend code changed afterward. This is deterministic regression evidence, not a fresh Electron/Claude/Codex/mobile QA capture or current-head CI result.

### Still unresolved

- **Claude identity / repeated text / duplicate Stop / interrupted TUI turns (F3/F4/F6/F7):** conversation ID + runtime generation do not prove which native turn produced Stop. Latest matching text cannot prove a replay is complete when turns repeat. A confirmed cancellation must be tied to the checkpoint's exact native transcript boundary; dropping trusted text solely because an old interrupted Chat row matches it is unsafe.
- **Legacy provider ownership:** an itemless, future-clock row without structural scope remains ambiguous. Historical coordination-preserved text may belong to another provider. The report's claim that *every* successful agent switch now fails is not established: both current activation queries clear outgoing hook text/provenance.
- **Runtime marker persistence:** startup marker failures now surface, but runtime callers still lack durable retry/recovery when the marker write fails.
- **Scope of revision conversion:** the cited `agent_switch_failure_observability.sql` checks **agent_switches.updated_at**, not sessions; the same clamping-writer defect is not established there. Generic lifecycle `m.mutate` is still read + `UpdateSession`; converting snapshot checks does not make every lifecycle mutation a database CAS.
- **Mobile:** blocked state still requires manual recheck after desktop restart; ordinary transitions disable the composer's Stop control. No new native mobile QA in this follow-up.
- **Consent F14 / F15:** the follow-up names these without their original reproductions; those details have been requested. No claim that they are fixed.
- Trigger write-amplification measurement is retained as a performance concern; no unmeasured trigger replacement is included.

### Independent review

**Standards:** no remaining actionable finding in the focused startup/polling/schema/hook/revision-fence delta after fixing the reviewers' exact-ID, shutdown-admission and cancellation-race findings. These were static reviews; reviewers did not independently execute the final suites.

**Spec:** tied-answer, rolled-back, equal-time previous-provider and same-timestamp session-write regressions are covered; no additional actionable finding in that scoped delta. The legacy itemless/future-clock gap and Claude interruption gap above remain open. These reviews are not blanket approval of this PR.

Screenshots, raw reports and runtime data remain local/GitHub attachments, never committed. The manual QA harness remains opt-in and is not enabled on CI.

## Problem

TUI → Chat could get stuck on **“Interface switch failed” (AO-2L)** even when the provider's conversation history was valid.

AO used `latest_user_prompt` / `latest_assistant_update` as replay checkpoints without knowing which conversation, runtime generation, or hook event produced them. Stale text, subagent output, or AO coordination prompts could become mandatory history matches. Retrying kept hitting the same mismatch.

## Fix

- **Track checkpoint provenance:** persist conversation ID, runtime generation, event origin, and trust. Main-agent origin is checked, but this alone does **not** verify Claude turn identity; delayed Stop remains unresolved.
- **Separate stale text from real history conflicts:** **Use provider history** can bypass a legacy/untrusted text mismatch. Trusted checkpoints, native-conversation identity, pending turns, and AO's durable replay boundary remain strict.
- **Make recovery explicit and race-safe:** show **Retry / Use provider history / Stay in Terminal**. Bind consent to the exact failed transition, import without duplicate turns, and confirm the target controller has stopped before retry or rollback. Unconfirmed shutdown requires a restart.

Result: an old checkpoint without trustworthy conversation/runtime metadata can be bypassed with explicit **Use provider history** consent. This does **not** bypass trusted mismatches or fix every TUI → Chat failure; current-head QA below still found failures.

Closes #4424.

## September 13 — original-branch return verified

**This is the separate history timeout reported in #5322, not its false-draft bug.** After an edited branch produced a Terminal reply, selecting the original branch and immediately doing Chat → TUI → Chat failed on main: the original provider history could not match the edited branch's stored hook text.

**The existing fix in this PR covers it:** `CommitSessionControllerEpoch` clears the outgoing hook checkpoint on Chat → TUI. With no new Terminal input, the return validates the original branch against its own durable history boundary. It does not require an edited-branch prompt, waive a trusted identity mismatch, or copy the edited suffix.

Fresh native verification used **Claude 2.1.270**, macOS arm64, disposable todo-app data, and local integration `79bba1e6c`: fetched main `cadde8c9f` + this PR's implementation `c23ce0714` + #5302 + #5322. This follow-up head `c55c17e09` adds **only one regression test (+134 lines)**; no additional production patch or QA artifact.

| Case | Result |
| --- | --- |
| Retry the already-poisoned, pre-#4466 session after upgrade | Strict rejection becomes `TARGET_HISTORY_UNTRUSTED_TEXT_MISMATCH` (~46s). It is not silently repaired. Explicit **Use provider history and switch** recovers it in 2.85s. |
| On the edited branch, send a new Terminal turn `C4424_EDITED_NEW`, then return to Chat | Trusted checkpoint recorded for that branch; strict return passes in 3.24s. |
| Select original branch; Chat → TUI → Chat, **no new Terminal prompt** | Strict return passes in **3.30s**; repeated in **4.16s**. No retry or provider-history bypass on either return. |
| Inspect original history, then send `C4424_ORIGINAL_AFTER_RETURN` in Chat | Original three user/assistant pairs retained exactly once; edited suffix absent. New fourth reply completes. |

The provider-ID/checkpoint preconditions and final message arrays were checked against the real daemon/SQLite. Branch selection and messages used the native UI; supplementary strict transitions used the real daemon. This is **not** a fresh Codex branch-matrix run or an all-Claude-errors-pass claim. The separate busy-stop mismatch remains open.

<details open>
<summary>Native evidence — what to observe</summary>

**Before, main + #5322 only:** the original-branch return hits history settling after 46s, despite an empty Terminal composer.

Before: original-branch history timeout

**After, with #4466 included:** branch **1/2** is back in Chat. `C5301_CHAT_A`, `C5301_TERMINAL_B`, and `C5301_CHAT_AFTER_SWITCH` each appear once; no edited-branch replies appear. This capture follows the repeated strict return above. The “recovered” label under the imported Terminal turn is not an error and is not what this fix changes.

After: original branch restored by strict switching

**Continuation:** below those same original turns, the new `C4424_ORIGINAL_AFTER_RETURN` reply completes. Open the image for the full-size conversation.

After: fresh reply on the original branch

Screenshots are GitHub attachments; raw transitions and QA databases stay local, never committed.

</details>

**Regression:** `TestOriginalBranchTUIRoundtripDoesNotInheritEditedBranchCheckpoint` uses real SQLite, edit/branch activation, lifecycle hook updates, branch-specific provider replay, and strict admission. It passes 10 consecutive normal runs. A local-only Go overlay removing the seven existing Chat → TUI checkpoint-reset assignments makes it fail at the original return (`native_identity` mismatch); no generated file was edited or committed. Local Chat/lifecycle/store race suites pass (unchanged lifecycle/store packages used Go's test cache).

<details>
<summary>Independent review of c23ce0714…c55c17e09</summary>

## Standards

No hard documented standard violations or actionable smells. The test follows the existing SQLite/edit harness and lifecycle write boundaries; local helpers remove repeated fixture setup.

## Spec

No missing, incorrect, or out-of-scope behavior found. The test creates and activates real branches, establishes the edited checkpoint through lifecycle signals, sends no new Terminal prompt on the original return, and requires exactly the original transcript.

Totals: Standards 0; Spec 0.

</details>

## September 12 follow-up fixes and evidence

- **Codex checkpoint identity (this PR):** carry native `turn_id` from UserPromptSubmit/Stop through the activity API into durable checkpoint provenance. Mark complete only for the matching Stop. A completed prompt-only checkpoint can match an earlier replay turn **only by its exact native ID**. Pending, unknown-ID, repeated-prompt and empty-text/wrong-ID cases stay strict; `trusted_turn_boundary` is nonwaivable.
- **Imported orchestrator edit:** fixed separately in #5302 by joining `current_session_id`. The original failed Claude edit now returns 202 and completes; its edited branch survives a round trip with the exact six messages. [Before/after evidence is in that PR](https://github.com/Untrivial-ai/agent-orchestrator/pull/5302).
- **Separate fixes/remaining work:** #5301's false empty-draft warning is fixed and natively verified in #5322 (OSC window-title filtering). Claude busy-stop trusted mismatch remains open. No terminal-capture or dependency change is included here. Why the original Codex B hook failed to update AO is unproven. Old stored prompt-only checkpoints without `turn_id` are **not** retroactively repaired and remain strict.

**Local verification:** full CLI/lifecycle/Chat/HTTP/SQLite suites, Go build/vet, frontend typecheck, and regenerated API/sqlc zero-drift checks passed. Full race runs for CLI, lifecycle, Chat, HTTP controllers, and SQLite store also passed at `c23ce0714`. Independent Standards and Spec reviews found no remaining issues after adding exact-turn and empty-text fencing. Native tests below use real Codex 0.153.4 history and **controlled checkpoint-state injection**, not a natural reproduction of the missed-hook cause.

<details>
<summary>New Codex screenshots — success, intentional rejection, successful continuation</summary>

**1. Pass:** real Terminal replies A/B; restore completed A checkpoint with its exact native turn ID. Strict switch retains both replies once. Capture at local integration `9a17408b2`; repeated successfully at final `d887aa0af` (= this head plus separate #5302).

Both native replies retained

**2. Expected safety rejection:** inject a nonexistent native turn ID. Final integration rejects `trusted_turn_boundary`, preserves Terminal, and offers no provider-history bypass. This screenshot is a passing negative test, not another claimed bug.

Wrong native turn is rejected

**3. Pass:** restore real A identity, switch, and complete `X4466_ID_AFTER_RECOVERY`. A separate Terminal↔Chat round trip preserved the same branch and exact four-message A/B history.

New Chat reply succeeds

Files are GitHub attachments, never committed. Full [evidence comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4466#issuecomment-5647472229).

</details>

## Initial QA at `9a34cdad2` — exact observed failures

Tested real Electron on macOS arm64, **Claude Code 2.1.269 / Opus 5 high** and **Codex 0.153.4 / gpt-5.6-terra medium**, with isolated data and disposable todo-app clones. These four cases happened **without fault injection**. They are observed failures on this head; **main-vs-PR attribution has not been established**, and they may share causes.

| Trigger | Observed result | Evidence |
| --- | --- | --- |
| **Codex:** complete Terminal prompts A and B, then switch to Chat. | `TARGET_HISTORY_UNSETTLED / trusted_user_text` after ~46s. UI Retry and a real-daemon reproduction fail again. Native history contains completed A/B, but AO retains prompt A with no assistant checkpoint. Terminal is preserved; no provider-history bypass. Why the later hook did not update AO is still unproven. | [Native error](https://github.com/user-attachments/assets/a7adaa28-e56c-43b7-856f-ac2b390d4e00) |
| **Claude:** start in Terminal, import A/B into Chat, complete Chat C/D, then edit C. | HTTP 409 `CHAT_EDIT_UNCERTAIN`; safe same-key retries also fail. No edited branch appears. Durable delivery stays `reserved`, `provider_work_started=0`. Candidate: the provider-work SQL joins `conversations.session_id`, which is NULL here, instead of the populated `current_session_id`. The error is **not proof the provider received the edit**. | [Native edit error](https://github.com/user-attachments/assets/8b91e402-120e-4dc5-8c4a-ba608670ce2b) |
| **Claude:** while a Terminal reply is generating, choose **Stop now and switch**. | Interrupted output remains in Terminal; switch and Retry fail with `TARGET_HISTORY_UNSETTLED / trusted_user_text`. `provider_history` is correctly refused because this is a trusted mismatch, but the user flow remains blocked. | [Interrupted turn + error](https://github.com/user-attachments/assets/758275db-c1f8-4c98-b9bf-9441ad8b074c) |
| **Claude:** return to Chat after a completed Terminal reply, or after selecting the original branch. | `DRAIN_DRAFT_PRESENT` despite a visibly empty Terminal prompt. Explicit **Discard draft and switch** recovered two cases; an ordinary retry without changing the composer recovered another. Draft detection/timing cause is not yet isolated. | [Empty prompt + warning](https://github.com/user-attachments/assets/acd5eb37-cb0d-4c93-90e0-ee539986feee) |

The Codex completed-turn failure was not permanent across every possible path: a later new Terminal turn followed by **Stop now and switch** succeeded. A separate genuinely unsubmitted Codex draft was correctly rejected and is **not** counted as a bug.

### What passed, and what was blocked

The full **Terminal-origin → Chat → edit → Terminal → Chat** flow is **not passed**: Claude reaches the edit and fails; Codex initially fails the import. Separate sessions created **directly in Chat** test the remaining branch flows, but do not substitute for that missing coverage.

| Scenario | Claude | Codex |
| --- | --- | --- |
| Chat-start controls: first/middle/repeated edits and immediate TUI/Chat round trips | Passed; middle edit uses `approximate_context` reconstruction | Passed; middle edit uses native fork |
| Original/edited branch selection; new Terminal turn on edited branch | Recovered after the empty-draft warnings above | Passed |
| Full Electron **and daemon** restart on edited branch; fresh Chat reply; another round trip | Passed | Passed |
| Cancel an active Chat reply; round-trip; send again | Passed | Passed |
| Busy native Terminal **Stop now and switch** | Failed, as above | Passed |
| Deliberately inject a nonexistent **legacy/untrusted** checkpoint; strict reject, Retry, explicit **Use provider history**, then fresh Chat reply | Passed controlled recovery | Passed controlled recovery |
| Trusted mismatch must not allow provider-history bypass | Passed safety rejection on the natural busy-turn failure | Passed controlled current-ID/generation mismatch rejection |

Successful round trips were checked against exact visible message arrays, active branch/provider identities, and unique marker counts. Restart preserved both edited branches and all stored message rows before new messages were sent. First/repeated edits and key recovery/cancellation actions were clicked in Electron; supplementary middle-edit/branch loops used the real daemon and authenticated providers, followed by native visual inspection. This was **not an entirely UI-click-driven suite**.

**Limits:** no OS-level unkillable-process injection, mobile UI, native Terminal `/fork`, exhaustive provider/auth/network outages, or long soak. Shutdown fencing has deterministic race-test coverage. Historical user-level Claude hooks referenced a missing old `ao.js`; global configuration was left untouched, so hook-related attribution needs a clean-config follow-up.

<details>
<summary>Fresh native screenshots — current head, failures and passing controls</summary>

These are September 12 captures at `9a34cdad2`, not before/after comparisons. Files are GitHub attachments, never committed.

**1. Codex natural failure:** completed A/B Terminal work, then strict switch rejection. Observe the `trusted_user_text` mismatch and retained Terminal; this is not an injected checkpoint.

Codex completed-turn switch failure

**2. Claude imported-origin edit failure:** editing the later Chat turn does not create a replacement answer/branch; **Retry edit safely** repeats the error.

Claude imported-origin edit blocked

**3. Claude busy-stop failure:** partial output is interrupted, but the switch still rejects the trusted prompt.

Claude interrupted Terminal switch failure

**4. Claude draft warning:** compare the empty Terminal composer with **AO found unsent text**. Discard-and-switch was required in this capture.

Claude empty Terminal composer draft warning

**5. Passing Codex Chat-start edit control:** the edited branch accepts a new native Terminal turn; that reply is imported once on returning to Chat.

Codex edited branch Terminal continuation

**6. Passing Claude Chat-start edit control:** reconstructed middle-edit branch after its immediate round trip. This is a different session from the failing imported-origin edit above.

Claude reconstructed middle edit round trip

**7–8. Restart controls:** same edited branches after full app/daemon restart. Observe the fresh `C4466_AFTER_RESTART` / `X4466_AFTER_RESTART` reply below the edited history. Separate durable-state assertions checked exact history and branch IDs.

Claude edited branch continues after restart

Codex edited branch continues after restart

**9. Controlled Codex legacy mismatch:** deliberately injected text absent from provider history produces the explicit **Use provider history** option. This option was correctly absent from the natural trusted mismatch in screenshot 1.

Codex injected legacy mismatch recovery options

**10. Controlled Claude legacy recovery:** after explicit recovery, a new `C4466_AFTER_LEGACY` Chat reply completes. This does not demonstrate recovery from the natural trusted busy-turn failure in screenshot 3.

Claude controlled legacy recovery and fresh reply

</details>

## Historical observations

| Scenario | Provider | Evidence |
| --- | --- | --- |
| **Stop now and switch** during a busy Terminal turn; subsequent idle retries, a new successful turn, and restart still failed. | Claude Code | Original [#4424 reproduction](https://github.com/Untrivial-ai/agent-orchestrator/issues/4424): five persistent `TARGET_HISTORY_UNSETTLED` / AO-2L failures. |
| A cancelled prompt remained in `latest_user_prompt` for four days, although it was absent from completed provider history. No interruption during the switch was needed. | Claude Code | [September 11 field report](https://github.com/Untrivial-ai/agent-orchestrator/issues/4424#issuecomment-5636832085): eight failures at ~46s; clearing that field allowed the same session to switch in 2.6s. Main's #5248 narrow guard is preserved here. |
| Deliberately injected a nonexistent legacy checkpoint into an otherwise valid conversation. | Claude Code | Native screenshots below: strict switch rejects; explicit **Use provider history** succeeds. Controlled fault injection, not a natural failure. |
| Exchanged ~2 Terminal messages, then switched to Chat without AO-2L or lost history. | Codex | [August 26 reviewer report](https://github.com/Untrivial-ai/agent-orchestrator/pull/4466#issuecomment-5431084085). Historical happy-path verification, not a fresh current-head run or proof of every recovery failure path. |

Stale-generation, subagent, coordination-prompt, and missed-prompt-hook cases also have deterministic contracts. Those tests use simulated provider events; they are not native Claude/Codex observations. The historical natural failures in this table were Claude cases; the fresh Codex failure is documented separately in current-head QA above.

## Editing a previous message, then switching again

Relevant flow: **TUI → Chat → edit earlier message → TUI → Chat**, including additional switches before the edit.

An AO-managed edit is different from invoking `/fork` directly in Terminal: it updates the active branch and session's `provider_conversation_id` together. The expected round-trip resumes the **edited branch**, not its parent; an edit alone should not cause a handle mismatch.

A native fork preserves provider-owned history. Without native fork support, AO can create a new provider conversation seeded with earlier textual context (`approximate_context`); that is not an exact copy of the provider transcript. Both paths need round-trip verification.

**Coverage by revision:** the initial `9a34cdad2` matrix above remains a historical failure record. #5302 subsequently verified the imported-origin edit and its edited-branch roundtrip; #5322 verified the false-draft fix. The September 13 integration above additionally verifies the original-branch revisit. These are focused passing cases, not a claim that every edit, interruption, or native fork works.

<details>
<summary>Other switch errors that remain possible</summary>

These are implemented failure conditions, not newly reproduced failures on both providers.

| Condition | Result |
| --- | --- |
| Legacy/untrusted checkpoint text is absent from provider replay. | `TARGET_HISTORY_UNTRUSTED_TEXT_MISMATCH`; explicit provider-history recovery is available for this mismatch only. |
| Replay lacks trusted completed work, has an unresolved turn boundary, or conflicts with trusted checkpoint identity. | `TARGET_HISTORY_UNSETTLED`; provider-history consent cannot waive the safety check. |
| Active Chat branch, session resume handle, or returned provider ID disagree. | Resume fails; provider-history consent does not reconcile different conversations. |
| Native transcript is missing, or the current Terminal launch has not confirmed its conversation ID. | `NATIVE_SESSION_MISSING` / `NATIVE_SESSION_UNVERIFIED`; switching is blocked. |
| Provider history is unavailable, resume fails, or provider/auth/binary prerequisites fail. | `TARGET_HISTORY_UNAVAILABLE`, `TARGET_RESUME_FAILED`, or the corresponding preflight error. |
| AO cannot confirm the target controller stopped. | `TARGET_STOP_UNCONFIRMED`; restart required, with input/switching blocked to prevent concurrent controllers. |

</details>

## Validation

**Fresh September 12 runs at the same head:** 165 frontend tests passed across interface-switch, SessionView and transition-hook suites; race-enabled edit/history/transition/lifecycle suites passed; additional provider-history/shutdown race tests passed; SQLite edit/branch/history tests passed. These green automated suites did **not** catch all four native failures above. No product edits or new commits were made in this QA run.

Head `9a34cdad2`: **13/13 CI checks passed**, including [full Go race tests](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34665073118/job/103475209567), [frontend tests](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34665073107/job/103475209396) (4,449 passed, 6 existing skips), typechecks, lint, and API/sqlc drift. Independent Standards + Spec reviews: no actionable findings.

<details>
<summary>Local validation details</summary>

Passed: 169 backend packages without race detection, build/vet, focused backend tests, and 147 focused UI tests. Broad local frontend/race runs had timing failures under heavy host load; some follow-up checks were stopped. Those runs are not claimed as passing. The exact-head CI suites above completed successfully.

</details>

<details>
<summary>Historical screenshots — earlier heads and controlled renderer fixture</summary>

Real Electron + Claude Code against an isolated `todo-app` session. The legacy mismatch was deliberately injected into disposable QA data; this is controlled reproduction, not a naturally occurring failure. Images are attachments, not committed files.

**1. Strict rejection + recovery actions** — native capture at `9a726bc59`. Provider history lacks the injected legacy prompt, so the strict switch rejects it and preserves Terminal.

Strict history mismatch with recovery actions

**2. Explicit recovery succeeds** — same session and revision. After **Use provider history**, the original Terminal reply appears once in ChatUI.

Same conversation recovered into ChatUI

**3. Recovery + continued chatting on the main-based branch** — native capture at `445d399d7`, based on main `e2417dc8e`. Top: imported `QA4466_LATEST_MAIN_NATIVE` turn. Bottom: new `QA4466_LATEST_MAIN_CHAT` reply. SQL checks confirmed one user/assistant pair per marker; composer cleared. Main's separate #5274 composer fix is included unchanged.

Recovered Terminal turn followed by a new successful Chat turn

These are earlier native captures, **not fresh screenshots of cleanup head `9a34cdad2`**. That head's automated validation is listed above.

**4. Unconfirmed shutdown blocks further switching** — controlled renderer fixture, **not a real provider shutdown failure**. `TARGET_STOP_UNCONFIRMED` shows a restart instruction and disables Terminal interaction; retry/cancel cannot bypass it. Backend tests separately cover controller shutdown ordering.

Unconfirmed shutdown requires restart and blocks Terminal input

</details>

## Remaining focused follow-up

1. Land/review the separately verified #5301 fix in #5322; it retains rejection of real unsent drafts.
2. Preserve native Claude interruption provenance so busy-stop switching can validate the real interrupted turn safely.
3. Determine the original missed Codex hook cause; the exact-turn fix above prevents matching a different occurrence but does not recover identity missing from old stored checkpoints.

## Not included

Conflicting active-branch/session provider handles remain hard failures; this PR does not reconcile conversation IDs. No mobile provider-history consent controls; blocked-state presentation and manual status recheck are included. The manual local QA harness stays in draft #4419 and is not enabled on CI. No screenshots, reports, or local runtime data are committed.

</details>


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

vanshx999:
**Review: tested manually (dev app) + deterministic contracts**

I reviewed this by running the app manually in dev mode and by running the deterministic contracts.

### Manual testing — dev app, Codex as the TUI provider
Ran the app via `cd frontend && npm run dev`. Dev mode isolates all state under `~/.ao/dev` (separate `userData`/DB from the packaged app) and uses my real Codex credentials, so a genuine TUI session works.

1. **Happy-path TUI → Chat switch (no regression).** Started a Codex TUI session, exchanged ~2 messages, then clicked **Switch to chat UI**. The transition completed cleanly into ChatUI with the conversation intact and **no AO-2L**. This confirms the fix doesn't break normal recovery (the old unscoped-checkpoint logic could falsely block even healthy sessions).

2. **Recommended manual coverage of the actual failure path (the part this PR fixes).** The AO-2L recovery can only be reached with a live native conversation, but it can be forced deterministically by seeding a poisoned *legacy* checkpoint into the dev SQLite and then attempting the switch:
   - Quit the app; locate the dev DB (`find ~/.ao/dev/data -name '*.db'`) and note the session id.
   - ```sql
     UPDATE sessions
     SET conversation_checkpoint_state = 'legacy',
         latest_user_prompt = 'POISONED-USER',
         latest_assistant_update = 'POISONED-ASSISTANT'
     WHERE id = '<sid>';
     ```
   - Reopen, **Switch to chat UI** → expect the **AO-2L notice** with three actions:
     - **Retry switch to Chat UI** → stays closed (same AO-2L reappears; strict retry does not bypass). ✅
     - **Use provider history and switch** → enters ChatUI and replays the real Codex turn **once**; a new prompt then works with no duplicate messages. ✅ *(this is the fix — recovery is no longer permanently stuck)*
     - **Stay in Terminal** → remains in Terminal. ✅
   - For the **trusted hard gate**, re-inject with `conversation_checkpoint_state = 'complete'` (same poisoned text) and switch again: the AO-2L still appears, but **"Use provider history and switch" must be hidden** — provider-history may only waive legacy/untrusted text, never current-turn facts. ✅

   (Cleanup between attempts: `rm -rf ~/.ao/dev`.)

### Automated checks (commit `4c87452`)
- `go test ./internal/cli/... -run TestHooks` — all hook trust-scoping tests pass (assistant checkpoint promoted only on main `stop`; Codex stop, `user-prompt-submit` assistant side, `subagent-stop`, and internal handoff turns do not pollute the main checkpoint). This is the root-cause fix.
- `AO_CHATUI_E2E_PORT=52145 npm run qa:chatui -- --grep MQA-06` — Playwright contract passes, and both tagged Go recovery contracts pass (`…ProviderHistoryRecoveryIsScoped`, `…DeduplicatesReplayWithoutWorktreeMutation`): recovery is correctly scoped and replay is deduped without mutating the worktree.
- `golangci-lint v2.12.2` — **0 issues**.

### Notes (not introduced by this PR)
- `TestChatUIRegressionEncodedSyntheticBranchIDIsDecoded` (controllers) fails — a branch-ID URL-decode issue; the file/handler it exercises is not touched by this PR, so it's pre-existing on the branch.
- `go test ./...` has a failing `TestCrushLocalAuthStatusDoesNotUseProviderCatalog` in `adapters/agent/crush` — unrelated adapter, outside this PR's scope.
- `npm run frontend:typecheck` fails on `packages/product-ui` (`Cannot find module 'motion/react'`) — a monorepo dependency gap; the PR's own components typecheck clean (the chatui typecheck passed inside the MQA-06 run).

**Verdict:** the history-recovery fix is verified manually (happy path) and by the MQA-06 recovery contracts; trusted/legacy checkpoint scoping, the AO-2L recovery UX, and replay-dedup all behave correctly. The unrelated failures should be tracked separately but don't block this change.

nikhilachale:
I found two blocking correctness concerns:

**P1 — Preserve the runtime-generation fence when `LaunchID` is absent**

In `backend/internal/lifecycle/manager.go:572-574`, ownership is now checked only when the incoming signal provides a launch ID. Previously, a live TUI row with a non-empty `RuntimeLaunchID` rejected an empty ID. A delayed legacy `session-end` or `user-prompt-submit` hook can therefore pass the guard and update the current generation. The SQL CAS does not protect this case because it compares against the generation read from the same current row, not the request.

Please reject a missing launch ID whenever the committed TUI owner has a runtime generation, and add a stale untagged-hook regression.

**P1 — Do not lose evidence of a newer completed turn when its prompt hook was missed**

In `backend/internal/lifecycle/manager.go:659-675`, a Stop that cannot be paired with the current prompt is discarded while the previous complete checkpoint remains unchanged. Then `backend/internal/service/chat/controller.go:483-495` can accept that older pair from a completed replay turn. If turn A completed, turn B's `UserPromptSubmit` hook was lost, and Stop for B arrived, a provider snapshot ending at A can be accepted and Chat starts without turn B.

Please retain evidence of the unmatched newer completion and require replay to advance beyond A, or fail closed. Add a regression where provider history ends at A rather than containing the later turn.

illegalcall:
Resolved the two blocking correctness comments and the additional exact-head review findings on current head `3d4fc840cbb72d77044f1736b7a0c0ca83d91227`.

What changed:

- Untagged terminal hooks can no longer mutate a committed TUI launch, a Chat controller epoch, or the ownerless TUI window during rollback.
- A Stop whose prompt hook was missed now records a durable unresolved-turn witness, so replay ending at an older completed turn cannot be accepted.
- Pane-delivered fallback prompts and same-native runtime restarts preserve that witness; only a canonical newer prompt or proven native-identity change clears it.
- Repeated assistant text cannot satisfy an assistant-only checkpoint by collision.
- Immutable hard mismatches fail before provider history polling instead of waiting up to 45 seconds.
- The new schema migration is reserved as `0117`, avoiding the stacked #4425 migration already using `0116`.

Deterministic regressions cover every boundary above, including a barrier exactly after rollback restores TUI mode and before the replacement runtime generation is committed.

Validation on the exact pushed head:

- `go test -p 1 ./... -count=1`: passed
- `go vet ./...`: passed
- affected lifecycle, Chat, session-manager, SQLite, and store suites: passed
- race suites for lifecycle, Chat, session-manager, SQLite, and store: passed
- sqlc regeneration and diff check: clean
- independent exact-head Spec review: APPROVED
- independent exact-head Standards review: APPROVED
- GitHub CI: all 10 checks passed
- worktree clean; no screenshot, report, trace, or local evidence artifact committed

nikhilachale:
**Re-review at `3d4fc840`**

Both previously reported P1 correctness issues are **fixed**:

- missing/untagged `LaunchID` callbacks are now fenced across committed TUI, Chat, and rollback ownerless epochs;
- an unmatched newer Stop now creates a durable unresolved-boundary witness, so an older completed provider replay cannot satisfy recovery.

Focused exact-head lifecycle, session-manager, Chat, and SQLite tests covering these regressions passed.

**P1 — Restack before merge: the current stacked base now owns migration 0115**

This PR still reserves `115: "0115_conversation_checkpoint_provenance.sql"` and adds that migration file. Its declared base, `codex/fix-4422-chat-drafts`, has since advanced and now owns `0115_usage_measurement_and_provider_usage.sql`. GitHub accordingly reports this PR as **CONFLICTING / DIRTY**.

Please restack onto the current #4463 head, assign unused migration versions in dependency order, regenerate sqlc/API artifacts, and rerun migration/full CI checks. The current green checks validate the stale recorded base, not the restacked result.

illegalcall:
Restacked #4466 onto the current #4463 head and resolved the migration collision on exact product head `f8e783c0e80bea59699a4576ed07b9bb0940af48` (base `cc9eeac65dbb8044251f3c5f34c16ca514254340`).

Migration ordering is now:

- base: `0112–0116`, `0118_conversation_steer_delivery.sql`, `0119_conversation_edit_delivery.sql`
- this PR: `0120_conversation_checkpoint_provenance.sql`, `0121_conversation_checkpoint_unsettled.sql`

Both #4466 SQL bodies are byte-identical to their pre-restack versions. The two merge conflicts were additive: both hook-test imports were retained, and the migration ledger contains the complete base plus `0120/0121`. SQLC and OpenAPI regenerate with zero drift.

Exact-head validation completed locally:

- focused affected suites and full serialized backend: passed
- affected-package race sweep: passed
- `go vet ./...`: passed
- GolangCI v2.12.2: 0 issues
- renderer switch matrix: 112/112
- full renderer: 3,122 passed / 1 skipped
- frontend + E2E typechecks: passed
- strict local MQA-06 typecheck + Playwright + tagged Go contract: passed
- clean worktree; no harness/evidence files committed

GitHub now reports the PR MERGEABLE. Fresh CI is running, and the PR body has the exact restack/evidence details.

illegalcall:
Final latest-base restack is complete on exact head `c942de28a73040b0c77fdf96a27459e3b7255261`, stacked on #4463 exact head `a63bfc9c5657a1fe91e6fd4662b81533048577e2`.

The four merge conflicts were resolved additively, and combined-contract fixture gaps were corrected without changing the intended product scope. Exact-head validation is now complete:

- full serialized backend: passed
- affected-package race sweep: passed
- vet + GolangCI v2.12.2: passed / 0 issues
- focused renderer matrix: 112/112
- full renderer: 3,144 passed / 1 intentional skip
- frontend + E2E typechecks: passed
- strict MQA-06 typecheck + Playwright + tagged Go: passed
- independent exact-head Standards review: APPROVED
- independent exact-head Spec review: APPROVED
- fresh GitHub CI: 10/10 passed, including the 12m48s race-enabled build-test

GitHub reports the PR MERGEABLE. The PR body now records the exact head/base and retains all three functional screenshots. No evidence, screenshot, trace, report, video, or harness artifact is committed.

nikhilachale:
**Re-review at `c942de28` — clean on the current exact base**

No actionable P0–P3 code findings were found.

The prior issues remain fixed:

- missing/untagged `LaunchID` callbacks are fenced across committed TUI ownership, Chat epochs, and rollback's temporary ownerless epoch;
- an unmatched newer Stop persists a durable unresolved-boundary witness and fails recovery before provider-history reading;
- the migration collision is resolved: this PR now uses 0120/0121 while its exact base owns 0118/0119;
- SQLC, Go OpenAPI, and pinned TypeScript API regeneration produce no drift.

Focused lifecycle, session-manager, Chat, SQLite, and HTTP tests passed; the Chat race package passed independently; `git diff --check` is clean; and all 10 GitHub checks are successful. GitHub reports this PR **MERGEABLE / CLEAN** against its recorded #4463 base.

**Stack caveat:** base PR #4463 is currently conflicting with newer `main`. After #4463 is restacked or merged, update this PR to the resulting base and rerun the focused suites, generators, and CI. This is a stack-integration requirement, not a defect in #4466's current exact-base diff.

nikhilachale (approved):

Prasad-D-Ware:
@illegalcall — this PR is still the right fix, and we hit #4424 in the wild today on packaged nightly `v0.12.13-nightly.202609091711` (daemon `378f9f713b`). Full reproduction with before/after timings is in [this comment](https://github.com/Untrivial-ai/agent-orchestrator/issues/4424#issuecomment-5636832085).

New trigger worth noting for your test matrix: it was **not** "Stop now and switch". `sessions.latest_user_prompt` held a **cancelled** turn's prompt from four days earlier, and the session could never return to Chat — eight consecutive `TARGET_HISTORY_UNSETTLED` failures at ~46s each. Your trust-state model already covers it: that row predates the provenance columns, so it classifies as legacy/untrusted and stops being a hard gate. Worth having as an explicit case, since the poisoning fact can be arbitrarily old and unrelated to the switch attempt.

## The blocker looks like coupling, not review

This PR is approved, but it cannot merge as stacked:

- it is based on #4463, which is itself **open and `REVIEW_REQUIRED`** (76 files, +11,625/−519) and unrelated to this bug
- migration numbers have since been taken on `main`: `main` is at `0135`, this PR adds `0120`/`0121`, and #4463 adds `0129`–`0131`
- the branch is `CONFLICTING` against `main`

Landing the P1 therefore means landing ~130 files and ~17k lines together, including an unrelated drafts feature.

**Would you be open to restacking this directly onto `main`** and renumbering to `0136`/`0137`, so it can land on its own? Happy to help with the conflict resolution or the renumbering if useful — the chat-drafts work in #4463 could then land on its own merits without gating a P1.

## Interim guard

In the meantime we have opened #5248 — a deliberately narrow guard (2 files, +148/−0, no schema change) that retires checkpoint text AO recorded on a cancelled, interrupted or failed turn, applying the rule `captureAOHighWater` already uses for its own anchor. It carries a regression test built from the live reproduction.

It is explicitly **not** a replacement for this PR: it has no provenance model, no coordination-prompt handling, no recovery-consent fencing, and none of the **Retry / Use provider history / Stay in Terminal** UI. It only stops sessions from becoming permanently unswitchable until this lands. If it conflicts with your restack, drop it — this PR's version should win.

illegalcall:
445d399d7-04-recovered-and-new-chat

illegalcall:
Codex checkpoint identity QA (September 12): real Codex 0.153.4 history with controlled checkpoint-state injection, not proof of the original missed-hook cause. Native QA integration includes this fix plus separate #5302.

1. **Pass:** restore completed turn A's exact native ID after real replies A/B; strict switch imports both replies once. Captured at integration `9a17408b2`; repeated successfully at final `d887aa0af`.

Both native replies retained

2. **Expected rejection:** inject a nonexistent turn ID; `trusted_turn_boundary` blocks switching and preserves Terminal. No provider-history bypass. Final integration `d887aa0af`.

Wrong native turn is rejected

3. **Pass:** restore real A identity, switch successfully, then complete a fresh `X4466_ID_AFTER_RECOVERY` Chat reply. Final integration `d887aa0af`.

New Chat reply succeeds

A Terminal↔Chat round trip also preserved the same branch and exact four-message A/B history. Old untyped checkpoints remain strict; this does not retroactively repair the original stored A checkpoint. Images are attachments, never committed.

illegalcall:
Correctness follow-up pushed in `2ba6ddfe4` — **not a merge-ready sign-off**.

Fixed in focused commits:
- [570f11be5](https://github.com/Untrivial-ai/agent-orchestrator/commit/570f11be5): database revision CAS, original-signal retry, tool-state rollback, logged/error-returning exhaustion.
- [7fa2d25e3](https://github.com/Untrivial-ai/agent-orchestrator/commit/7fa2d25e3): quarantine one unconfirmed target instead of aborting all startup.
- [c926d4efe](https://github.com/Untrivial-ai/agent-orchestrator/commit/c926d4efe), [3c3b62840](https://github.com/Untrivial-ai/agent-orchestrator/commit/3c3b62840): mobile blocked-recovery presentation, polling restart dependencies, disabled versus pending composer state.
- [1a422e980](https://github.com/Untrivial-ai/agent-orchestrator/commit/1a422e980): remove any-pair shortcut and retain coordination gates.
- [3a04eb58d](https://github.com/Untrivial-ai/agent-orchestrator/commit/3a04eb58d), [2ba6ddfe4](https://github.com/Untrivial-ai/agent-orchestrator/commit/2ba6ddfe4): reuse history index; retain reordered queue execution and disambiguate timestamp ties.

Before/after evidence is in deterministic regressions, not screenshots:
```text
Before:
  stale projection overwrote human checkpoint: applied=true
  retry lost original Stop payload: checkpoint=complete, answer=""
  one uncertain session prevented daemon startup
  expected mobile poll undefined, received 300
  older occurrence A admitted replay A+B without checkpoint C
  coordination erased durable replay boundary
After:
  corresponding regressions pass, including focused race runs
  mobile: 79 files / 788 tests pass; typecheck passes
  full affected Go non-race suites pass; build/vet and generated-spec/sqlc drift pass
```

Benchmark (1,000 turns): ~1.2 ms / 6,095 allocations → ~0.8 ms / 1,067 allocations per settle check. Independent reviews found follow-up gaps; the corrected deltas were re-reviewed with no further actionable findings in their scope.

**Race validation:** full Chat, lifecycle, session-manager and SQLite-store suites passed; final-head checkpoint/history regressions also passed. The SQLite migration package hit the default 10-minute package timeout and is rerunning separately with a 25-minute limit. The full race run is not claimed green.

**Still unresolved:** Claude turn identity/delayed Stop, duplicate Stop's unsettled latch, interrupted-turn reconciliation, and completeness proof for identical text without native identity. Latest-only text matching is a mitigation, not that proof. The remote branch is confirmed at `2ba6ddfe4`, but GitHub's PR metadata still reports the previous head and no new-head CI run is available yet. Previous-head green checks are not evidence for these fixes.

PR body now records these limits and current net counts. No evidence files are committed.

illegalcall:
## Follow-up review fixes — `2ba6ddfe4` → `8da6b72dc`

Six commits are pushed. **This is partial completion, not merge approval.** GitHub recognizes `8da6b72dc`, reports a base conflict, and reports no checks for this head. Prior green CI does not cover these changes.

| Commit | Actual correction | Red → green evidence |
| --- | --- | --- |
| [`bb6d3eb25`](https://github.com/Untrivial-ai/agent-orchestrator/commit/bb6d3eb25) | Stop desktop 250 ms polling on `TARGET_STOP_UNCONFIRMED`; retain manual refresh. | Hook regression observed 3 requests instead of 1 before the fix; after the fix the count stays 1 and explicit refresh adopts recovery status. |
| [`6945f7b62`](https://github.com/Untrivial-ai/agent-orchestrator/commit/6945f7b62) | Reject startup when a revision-enabled DB is missing `sessions_revision_update`. | Real SQLite `DROP TRIGGER` regression previously passed reconciliation; it now fails explicitly. Existing no-op/recursion/CDC/rollback checks still pass. |
| [`7fcfa1e1d`](https://github.com/Untrivial-ai/agent-orchestrator/commit/7fcfa1e1d) | Exclude rolled-back anchors; fix tied replay evidence and structurally scoped provider cutovers. | Regressions cover tied reordered turns, rolled-back ordinary/coordination turns, previous-provider sequence boundaries, itemless ties in either order, missing tied answers, and AO-only activities. Review also caught an empty-item-ID equality bypass; it is fixed. |
| [`9f2a27868`](https://github.com/Untrivial-ai/agent-orchestrator/commit/9f2a27868) | Defer startup interface recovery behind account recovery; quarantine durably marked agent-switch ambiguity. | Overlap no longer aborts startup. Tests require continuous input fencing, exact transition-ID targeting, shutdown-safe worker admission, and safe handling when cancellation wins the recovery CAS. Marker-write/discovery failures remain errors. |
| [`7c4ce73e2`](https://github.com/Untrivial-ai/agent-orchestrator/commit/7c4ce73e2) | Return typed retryable contention and retry only definitely uncommitted hook projections. | HTTP previously returned `500 INTERNAL_ERROR`; now `503 ACTIVITY_PROJECTION_BUSY` with independent usage recording. CLI previously sent one request; bounded cases now send the expected 3/4 identical payloads. Generic 503 and ambiguous transport failures do not retry. |
| [`8da6b72dc`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8da6b72dc) | Replace remaining sessions-row timestamp snapshot guards in activity observation and usage finalization. | Fixed-clock observer regression previously failed with `fresh prompt was overwritten with "idle"`. Real SQLite usage regression previously admitted stale finalization after an unchanged-timestamp write. Both now reject stale revisions; initial revision 0 remains valid. |

### Local validation

All commands ran from `backend/` unless noted. Final SQL was regenerated with `npm run sqlc`; generated files were not hand-edited.

```text
go build ./...                                                   PASS
go test -p 2 ./... -timeout 25m                                   PASS
go test -race -p 2 ./internal/lifecycle ./internal/observe/activity \
  ./internal/service/usage ./internal/storage/sqlite/store \
  ./internal/session_manager -timeout 25m                         PASS
go vet ./internal/lifecycle ./internal/observe/activity \
  ./internal/service/usage ./internal/session_manager \
  ./internal/storage/sqlite/store                                PASS
```

Final affected race output: lifecycle 1.695s, observer 13.848s, usage 40.106s, SQLite store 422.128s, session manager 111.327s. Chat, CLI and HTTP controller full race suites also passed before the final revision-only conversion, as did the revision-trigger race regression. Frontend typecheck and 43 tests across the interface-transition hook/switch files pass; no frontend code changed afterward.

The first default-concurrency backend run failed fake-agent timing, Pi hook-order and LAN identity tests. All three passed on rerun without changes; the subsequent full `-p 2` run passed. Successful unchanged package results may use the Go test cache. This is not a claim that the initial failures were proven unrelated.

### Independent review

**Standards:** no remaining actionable findings in the focused six-commit delta after corrections. Exact-ID recovery, worker shutdown admission, cancellation fencing and revision terminology were checked.

**Spec:** no additional actionable findings in that bounded delta. Tied boundaries must contain each answer/item, rolled-back history is excluded, and usage/observer checks preserve runtime-launch and revision fences.

Both were static reviews; reviewers did not independently rerun the final suites. These results are not approval of the entire PR.

### Still open

- Claude delayed/duplicate Stop identity, repeated-text replay completeness, and interrupted native TUI recovery (F3/F4/F6/F7). Conversation ID + runtime generation do not prove turn identity. No unsafe trusted-checkpoint waiver was added.
- Legacy itemless, future-clock history without structural provider ownership remains ambiguous. Current agent-switch activation queries do clear outgoing text/provenance; the claim that every current successful provider switch fails was not established.
- Runtime callers still lack durable recovery if persisting `TARGET_STOP_UNCONFIRMED` fails. Startup now surfaces that failure rather than hiding it.
- Mobile foreground/reconnect recheck and the ordinary-transition composer Stop lock remain unchanged. Manual Retry exists, but is not automatic healing. No new native mobile verification is claimed.
- Original F14 consent-binding and F15 reproductions were requested; the supplied follow-up names them but does not describe them.
- Trigger throughput overhead is unchanged. The cited agent-switch timestamp guard belongs to `agent_switches`, not `sessions`; it needs its own reproduction. Generic lifecycle mutations are not all database CAS writes.

No fresh Electron/provider screenshot capture was performed for these six commits. Older screenshots in the body retain their original revisions and limits. All reports, traces, screenshots and runtime data stay local/GitHub attachments; none are committed. The manual QA harness is still opt-in, not enabled on CI.

illegalcall:
## Review-comment follow-up — `6ae6e7b91`

- @nikhilachale: the original untagged-launch and missed-prompt/Stop regressions still pass on the updated branch, including the ownerless rollback fence.
- @Prasad-D-Ware: the September 11 cancelled legacy prompt has an explicit [regression](https://github.com/Untrivial-ai/agent-orchestrator/blob/6ae6e7b9180a4b73c5cfeef3fc142fa0ecbbfb7f/backend/internal/service/chat/checkpoint_recovery_test.go#L19). Strict rejection requires provider-history consent for the hidden cancelled message; repeated recovery keeps history once, and a newer trusted TUI prompt cannot borrow the old failed Chat outcome. Fresh focused tests pass.
- The PR is based directly on main. [`917c1ee8f`](https://github.com/Untrivial-ai/agent-orchestrator/commit/917c1ee8f) integrates main `c1538a9a2` and resolves the single test-file conflict while retaining both sides' coverage. GitHub now reports **MERGEABLE**. [`6ae6e7b91`](https://github.com/Untrivial-ai/agent-orchestrator/commit/6ae6e7b91) adds only lint intent (+3/-2), not new recovery behavior.

**Standards:** 0 actionable findings in the merge/lint delta.  
**Spec:** 0 actionable findings in the merge/lint delta.

Full backend non-race/build/vet, pinned lint, API/sqlc drift, 146 SessionView/transition tests, 788 mobile tests, Node-24 frontend/E2E typechecks, cloud and shared-package checks pass. Local broad reruns hit disk exhaustion after earlier HTTP/CLI cleanup/renderer timing failures; the body records exact limits. **Current-head CI: 14/15 green**, including full frontend (311 files), renderer smoke and all three native CLI jobs. Only backend race testing remains in progress; I am not calling incomplete local runs green.

**This does not close the documented Claude identity/interruption, repeated-text completeness, runtime marker-persistence or other remaining gaps. The PR remains NOT READY TO MERGE.** Screenshots/reports/runtime data are not committed.

nikhilachale:
Re-review at current head `6ae6e7b91`: **request changes**.

**P1 — Claude Stop events can still certify the wrong prompt as trusted.** [`hookConversationFacts`](https://github.com/Untrivial-ai/agent-orchestrator/blob/6ae6e7b9180a4b73c5cfeef3fc142fa0ecbbfb7f/backend/internal/cli/hooks.go#L273-L288) records native `turn_id` only for Codex. For Claude, the checkpoint turn ID remains empty, and [the lifecycle match](https://github.com/Untrivial-ai/agent-orchestrator/blob/6ae6e7b9180a4b73c5cfeef3fc142fa0ecbbfb7f/backend/internal/lifecycle/manager.go#L763-L770) therefore accepts any same-generation, same-conversation Stop. A delayed or duplicate Stop from an interrupted earlier turn can mark the newest prompt complete with the wrong assistant text. [Chat recovery](https://github.com/Untrivial-ai/agent-orchestrator/blob/6ae6e7b9180a4b73c5cfeef3fc142fa0ecbbfb7f/backend/internal/service/chat/service.go#L302-L315) then treats that provenance as trusted and non-waivable, recreating a permanent `TARGET_HISTORY_UNSETTLED` failure.

Please correlate Claude Stop with the exact native transcript boundary, or fail closed rather than promote it to trusted completion without such proof. Current CI is 15/15 green, but it does not cover this sequence. The prior approval at `c942de28` is not an ancestor of the current head.

illegalcall:
@nikhilachale — addressed the Claude Stop identity finding in [`c6f7c1967`](https://github.com/Untrivial-ai/agent-orchestrator/commit/c6f7c1967e6b6e392c3b63c885fbfba753265343), pushed to this same PR.

Native reproduction confirmed the problem and one extra detail: queued UserPromptSubmit B reuses executing A's prompt_id. A hook-ID-only candidate also failed native QA and was discarded.

The fix records a per-submission UUID through Claude's hook additionalContext and AO's revision-fenced activity update. Native attachment ancestry binds it to the actual user UUID. The adapter verifies every retained submission/Stop after source shutdown, and Chat requires the latest exact native user/answer boundary in completed ACP replay. Stop alone cannot certify the latest display prompt; old/unprovable evidence remains fail-closed.

**Before:** A and queued B both finish, but switching rejects with unsettled_turn_boundary. This was a queued-input flow, not a confirmed interruption.
Before native switch failure

**After:** queued B imports once under strict history policy; new Chat C completes.
After queued import and new Chat response

A later Chat → TUI → Chat retained C and a new Terminal D. The second return required explicit draft discard because this build excludes the separate #5322 false-draft fix; no provider-history bypass was used.
After second return

Full backend non-race/build/vet, focused race regressions, pinned lint (0 issues), API/sqlc drift, frontend typecheck and 43 interface-switch tests pass. Independent Standards and Spec reviews have no outstanding findings in this delta. Current-head CI is running.

Remaining limits are explicit in the updated body: the marker consumes a small amount of context; old trusted checkpoints without native witnesses, compaction and interrupted native turns are not automatically repaired. This is not an all-PR merge-ready sign-off. Screenshots/reports/runtime state are not committed.

illegalcall:
Resolved the latest main conflicts in `1ea5af15c`, preserving #5305 handoff-readiness checks and #5348 restart-status behavior. **All 15/15 CI checks now pass; GitHub reports MERGEABLE / CLEAN and APPROVED.**

The CI failure was the SQLite migration package exceeding its cumulative 15-minute race-test timeout. With the bounded timeout increased to 20 minutes, the full package passed in **920.329 seconds (15m20s)**—no tests or race coverage removed. [Passing backend race job](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34841082075/job/103965979051).

The full frontend suite passed **4,601 tests (6 existing skips)**. Local full-suite disk-space limitations remain documented in the PR body; the clean-runner CI result is verified. No screenshots, reports, or runtime evidence committed.

illegalcall:
## Review recheck: one residual startup bug fixed

Pushed [`8cd59c309`](https://github.com/Untrivial-ai/agent-orchestrator/commit/8cd59c309c619fa19e0c5a11fc846ed6fb7f6fc2) to this PR. Two files only: implementation **+9/-3**, tests **+61/-4**.

| Review claim | Current-head assessment |
| --- | --- |
| 1. Desktop keeps polling restart-required state | Already fixed by [bb6d3eb25](https://github.com/Untrivial-ai/agent-orchestrator/commit/bb6d3eb25). Reran hook suite: 19 passing tests; the regression asserts polling stops and explicit refresh still works. |
| 2. `dropUnsettledHookFacts` unchanged | Byte equality between older revisions does not establish a current bug. Its retirement is deliberately limited to untrusted legacy text. Tests reject waiving trusted TUI facts merely because an old failed/cancelled/interrupted Chat row matches. This does **not** claim safe interrupted-native-Claude recovery is solved. |
| 3. Rolled-back high-water anchor | Already fixed by [7fcfa1e1d](https://github.com/Untrivial-ai/agent-orchestrator/commit/7fcfa1e1d). Both ordinary and coordination rollback regressions pass. |
| 4. Queue test misses timestamp ties | That original test is not the tie test. The additional equal-`RequestedAt`, tied-answer, and itemless-peer regressions pass; all required tied boundaries must be present. |
| 5. Startup still aborts | Original account-operation collision already defers exact-transition recovery. **Found and fixed another path:** a live target with a missing/mismatched native identity persisted a quarantine marker, then returned an ordinary error that aborted startup. Successful identity quarantine now stays session-local. Registry/marker failures remain fatal. |
| 6. Missing revision trigger ignored | Already fixed by [6945f7b62](https://github.com/Untrivial-ai/agent-orchestrator/commit/6945f7b62). Reran real SQLite `DROP TRIGGER` regression: startup reconciliation rejects it explicitly. |

The unpublished-fixes claim is stale: GitHub already held `1ea5af15c` before this recheck, not `c55c17e09`.

### Red → green evidence for the new fix

Extended `TestStartupQuarantinesPersistedAgentSwitchAmbiguity` before changing implementation. Six cases failed:

```text
persisted per-session quarantine aborted startup: reconcile: agent-switch pass: target native session reference is missing
persisted per-session quarantine aborted startup: reconcile: agent-switch pass: target native session identity is absent or does not match the reserved generation
```

Cases: missing reference, missing row, wrong session, wrong harness, wrong generation, empty native ID. After the fix all **18 matrix cases pass**. They also verify failed marker writes, failed registry reads (including repeated startup with an existing marker), healthy-sibling input, retained affected-session fencing, no runtime creation/destruction, and explicit recovery continuing to report unresolved ownership.

### Validation

- Full affected suite: `go test ./internal/session_manager -count=1 -timeout=20m` **PASS**, 31.746s.
- Focused startup/interface/explicit recovery checks **PASS**, 0.852s.
- 13 checkpoint/index tests **PASS**, 0.851s.
- SQLite revision/trigger regression **PASS**, 1.279s.
- Desktop interface hook suite **PASS**, 19 tests.
- `git diff --check` **PASS**.
- Local full-build, full-backend, and race build hit **ENOSPC**; chained vet was not reached and pinned golangci-lint is unavailable. These are not local passes. **Fresh CI is 15/15 green on `8cd59c309`**, including [full Go build/vet/race tests](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844380129), [frontend tests](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844380059), and [cross-platform CLI checks](https://github.com/Untrivial-ai/agent-orchestrator/actions/runs/34844379870). Full race package results: session manager `208.627s`, Chat `301.732s`, SQLite `721.807s`. GitHub reports **APPROVED / MERGEABLE / CLEAN**. These fresh remote passes close the broad-validation gap; they do not turn the failed local builds into local passes.

## Standards

Independent review of the final two-file patch: **0 actionable findings**. Surgical change and boundary-test requirements satisfied.

## Spec

Independent review of the final two-file patch: **0 actionable findings**. The persisted quarantine no longer blocks healthy sessions; infrastructure failures and unresolved explicit recovery remain visible. Review agents inspected code; main agent ran the tests above.

No fresh native Electron screenshot run is claimed for this deterministic service-level startup regression. Local evidence stays outside git; no report, screenshot, or runtime file was committed. PR opening counts were recalculated from all 108 changed files against its actual base.

PR #4472, opened 2026-08-26
fix(chat): make composer selection visible

### LINKED ISSUE #4471: bug(chat): selected composer text has no visible highlight
## Bug

Dragging across text already typed into the Chat composer creates a real DOM selection, but the highlight is effectively invisible. Users cannot tell which part of a draft is selected before replacing, cutting, or copying it.

**Source:** AO chat report on 2026-08-26 | **Analyzed against:** `c4abb8afc4a305fbd20c6d3955a2c7808d78e7d4` | **AO version:** `dev` (desktop `0.10.3`)
**OS reproduced on:** macOS 26.5, arm64 | **Daemon:** ready on `127.0.0.1:3001`
**Priority:** P2 — draft editing is significantly degraded; keyboard-only selection is possible but still has no visual confirmation.
**Confidence:** High — the exact pointer path is automated, the DOM range remains non-collapsed, and the rendered selection/surface contrast is measured.

## Reproduction

1. Open a Chat-mode session with an existing conversation.
2. Type a multi-word draft in the composer.
3. Drag the pointer across part or all of the draft.
4. Observe that no visible highlight appears, even though `window.getSelection()` contains a non-collapsed range.

The focused Playwright reproduction types a wrapped draft, drag-selects it through actual pointer events, verifies the DOM selection, then measures the composited selection-highlight contrast. On current `main` it fails:

```text
Expected: >= 1.5
Received:    1.013160847557346
```

## Root cause

`frontend/src/renderer/components/chat/ComposerEditor.tsx` styles selected text with:

```text
selection:bg-accent/30
```

In Tailwind's theme bridge, `accent` is the subtle surface token (`--accent`), not the stronger product accent. In the light theme it is `oklch(0.967 0.001 286.375)` over the composer's `oklch(0.985 0 0)` card surface. At 30% alpha the composited contrast is only `1.013:1`, so the browser paints a valid selection that looks unselected.

This was introduced with the Lexical composer in [PR #4331](https://github.com/Untrivial-ai/agent-orchestrator/pull/4331), commit `b8e3d57f25`.

## Fix

Use a theme-aware foreground/primary selection tint with enough alpha to remain visibly distinct from the composer surface, and keep an E2E regression that proves both:

- pointer drag creates a durable selection of the typed draft;
- the composited highlight/surface contrast is at least `1.5:1`.

## Evidence

Before — the draft is selected in the DOM, but no highlight is perceptible:

Before: invisible chat composer selection

After — the same automated pointer selection is visibly highlighted:

After: visible chat composer selection

## Impact

- Users cannot safely replace or cut a portion of a drafted instruction.
- Copy/cut may operate on text with no visible confirmation of the range.
- The failure looks like pointer selection is broken even though the browser selection model is working.

## Related

- [#4268](https://github.com/Untrivial-ai/agent-orchestrator/issues/4268) — read-only surfaces inherit `user-select: none`; adjacent symptom, different mechanism. The composer already computes `user-select: text`.
- [PR #4443](https://github.com/Untrivial-ai/agent-orchestrator/pull/4443) — restores selection on read-only content surfaces; it does not change the composer's selection highlight.

illegalcall:
Fixes #4471

## Summary

- use an explicit inverse semantic selection pair so both the highlight and selected draft text remain visible in every supported theme
- add a T0 Playwright regression that fills the real chat composer, pointer-drags across wrapped text, asserts a non-collapsed DOM selection, and checks both contrast relationships across every supported palette

## TDD reproduction

Before the styling change, the new test reproduced the report: the pointer drag selected the text, but the original composited highlight contrast was only 1.013:1.

The first review follow-up expanded the regression to all nine supported styles in light and dark mode. The prior 25% primary tint failed that 18-combination matrix at 1.279:1 on Gruvbox light, so it was increased to 45%.

The second review follow-up added the selected-text assertion before changing the implementation. That assertion failed with the 45% primary tint at 2.275:1 in Solarized dark. The final `selection:bg-foreground selection:text-background` pair gives the limiting Solarized dark case 4.111:1 highlight-to-surface contrast and 4.748:1 selected-text-to-highlight contrast.

After the change, the same pointer-drag test passes consistently across all 18 combinations (3/3 serial repeated runs).

## Visual proof

Visible and readable text selection in the chat composer

## Verification

- `npm run typecheck`
- `npm run typecheck:e2e`
- focused Playwright composer-selection regression, 3/3 serial repeated runs
- real Electron dev launch using isolated AO data on 127.0.0.1:3012; daemon, main, preload, and renderer bundles launched successfully
- user verified the fixed selection behavior in the native app
- `npm test -- --reporter=dot`: 2,887 tests passed; the existing unrelated landing subtree remains red in 10 files because cheerio and `@ao/shared/constants` cannot be resolved locally

### DIFF
--- a/frontend/src/renderer/components/chat/ComposerEditor.tsx
+++ b/frontend/src/renderer/components/chat/ComposerEditor.tsx
@@ -399,7 +399,7 @@ export const ComposerEditor = forwardRef<
 								if (event.defaultPrevented) event.stopPropagation();
 							}}
 							className={cn(
-								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-accent/30",
+								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/25",
 								disabled && "opacity-50",
 							)}
 						/>
--- a/frontend/src/renderer/components/chat/ComposerEditor.tsx
+++ b/frontend/src/renderer/components/chat/ComposerEditor.tsx
@@ -399,7 +399,7 @@ export const ComposerEditor = forwardRef<
 								if (event.defaultPrevented) event.stopPropagation();
 							}}
 							className={cn(
-								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/25",
+								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/45",
 								disabled && "opacity-50",
 							)}
 						/>
--- a/frontend/src/renderer/components/chat/ComposerEditor.tsx
+++ b/frontend/src/renderer/components/chat/ComposerEditor.tsx
@@ -399,7 +399,7 @@ export const ComposerEditor = forwardRef<
 								if (event.defaultPrevented) event.stopPropagation();
 							}}
 							className={cn(
-								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/45",
+								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-foreground selection:text-background",
 								disabled && "opacity-50",
 							)}
 						/>
--- a/frontend/e2e/chat-composer-text-selection.spec.ts
+++ b/frontend/e2e/chat-composer-text-selection.spec.ts
@@ -1,0 +1,168 @@
+import { expect, test } from "@playwright/test";
+import { installFakeAgent } from "./support/fake-bridge";
+
+const projectId = "chat-composer-selection";
+const sessionId = "chat-composer-selection-worker";
+const draft =
+	"Select this existing chat draft across its wrapped text so an earlier instruction can be replaced precisely without retyping the entire message.";
+
+test("typed chat composer text is visibly selected with a pointer drag @T0", async ({ page }) => {
+	await page.emulateMedia({ reducedMotion: "reduce" });
+	await installFakeAgent(page, {
+		projectId,
+		projectName: projectId,
+		workers: [{ id: sessionId, provider: "codex", title: "Composer selection", mode: "chat" }],
+	});
+	await page.route("http://127.0.0.1:8080/api/v1/**", async (route) => {
+		const pathname = new URL(route.request().url()).pathname;
+		if (pathname === `/api/v1/projects/${projectId}`) {
+			await route.fulfill({
+				json: {
+					status: "ok",
+					project: { id: projectId, agent: "codex", config: { worker: { agent: "codex" } } },
+				},
+			});
+			return;
+		}
+		if (pathname === `/api/v1/sessions/${sessionId}/conversation`) {
+			const completedAt = "2026-08-26T06:00:00Z";
+			await route.fulfill({
+				json: {
+					conversationId: "conversation-chat-composer-selection",
+					sessionId,
+					harness: "codex",
+					mode: "chat",
+					controller: "ready",
+					latestSequence: 1,
+					oldestSequence: 1,
+					hasMoreBefore: false,
+					turns: [
+						{
+							id: "turn-1",
+							state: "completed",
+							requestedAt: completedAt,
+							startedAt: completedAt,
+							completedAt,
+						},
+					],
+					messages: [
+						{
+							kind: "message",
+							id: "message-1",
+							turnId: "turn-1",
+							sequence: 1,
+							revision: 0,
+							role: "user",
+							origin: "human",
+							text: "An existing conversation",
+							streaming: false,
+							createdAt: completedAt,
+						},
+					],
+					activities: [],
+					settings: {},
+				},
+			});
+			return;
+		}
+		if (pathname === `/api/v1/sessions/${sessionId}/conversation/models`) {
+			await route.fulfill({ json: { models: [], selected: {} } });
+			return;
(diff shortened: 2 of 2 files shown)


frontend/src/renderer/components/chat/ComposerEditor.tsx
```diff
@@ -399,7 +399,7 @@ export const ComposerEditor = forwardRef<
 								if (event.defaultPrevented) event.stopPropagation();
 							}}
 							className={cn(
-								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-accent/30",
+								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/25",
```
illegalcall: **[P2] Keep the tint above 1.5:1 in every supported theme**

`primary/25` does not clear the required threshold across the selectable palettes. For example, Rose Pine light sets the composer/card to `#f9f0e8` and primary to `#907aa9`; compositing that primary at 25% yields roughly `#dfd3d8`, only about 1.29:1 against the card. Solarized dark is also below 1.5:1. The new E2E never selects a named `data-style-theme`, so it remains green in these failing supported configurations and users on those themes retain the reported bug. Please choose a semantic tint/opacity verified across every supported palette and cover the limiting light/dark theme pairs in the contrast regression.

illegalcall (commented):

The pointer-drag regression is useful, but the chosen tint does not satisfy the required 1.5:1 contrast across the supported theme set. Supported named themes still reproduce the visibility problem, so this needs a theme-safe tint and matching regression coverage before it is ready.

### Standards
No documented-standard violations found.

### Spec
One blocking correctness finding is attached inline.

Tests were not run, per review policy.

frontend/src/renderer/components/chat/ComposerEditor.tsx
illegalcall: Addressed in b4d3ee7c0. The composer now uses the semantic primary tint at 45% opacity. The pointer-drag E2E enumerates all nine supported style palettes in both light and dark modes (18 combinations), composites the computed ::selection color against the settled composer surface, and enforces the 1.5:1 floor. The old 25% tint fails that matrix at 1.279:1 on Gruvbox light; 40% only reached 1.5047:1, so 45% provides a safer rendering margin. The matrix passes in three repeated runs, and both frontend and E2E typechecks pass.

illegalcall (commented):
Addressed review 5029502611 in b4d3ee7c0. The composer selection now uses the semantic primary tint at 45% opacity. The pointer-drag E2E covers all nine supported style palettes in both light and dark mode (18 combinations), composites the computed ::selection background against the settled composer surface, and enforces the 1.5:1 floor. The previous 25% tint failed at 1.279:1 on Gruvbox light; 40% reached only 1.5047:1, so 45% provides safer margin. The matrix passes in three repeated runs, frontend and E2E typechecks pass, and all current PR checks are green. The associated inline thread PRRT_kwDORPZUAc6cbTgE has been resolved.

frontend/src/renderer/components/chat/ComposerEditor.tsx
```diff
@@ -399,7 +399,7 @@ export const ComposerEditor = forwardRef<
 								if (event.defaultPrevented) event.stopPropagation();
 							}}
 							className={cn(
-								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-accent/30",
+								"chat-composer-scrollbar max-h-40 min-h-[4.5rem] w-full overflow-y-auto overscroll-contain bg-transparent py-1 pl-[7px] pr-0 text-base! leading-relaxed text-foreground caret-foreground outline-none selection:bg-primary/45",
```
illegalcall: **[P2] Preserve contrast for the selected draft text**

This leaves the 14px `text-foreground` glyphs on the stronger selection background. In the supported Gruvbox dark theme, `--foreground` is `#ebdbb2`, the composer/card is `#3c3836`, and `--primary` is `#fabd2f`; compositing primary at 45% produces roughly `#927433`, so the selected glyphs have only about 3.2:1 contrast. That is below the 4.5:1 floor for normal text in [WCAG 2.2 SC 1.4.3](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html). Because an author selection background suppresses the UA paired colors, the text keeps its inherited color rather than receiving an automatic contrasting foreground ([CSS highlight paired defaults](https://drafts.csswg.org/css-pseudo-4/#paired-defaults)). The E2E currently measures only highlight-versus-surface contrast, so all 18 cases can pass while the selected draft becomes hard to read. Please choose or set a selection foreground/background pair that satisfies both contrasts and assert text-versus-highlight contrast across the same matrix.

illegalcall (commented):

The previous theme-matrix finding is resolved: the 45% tint and 18-combination regression address the requested highlight/surface contrast floor. One adjacent accessibility regression remains: selected glyphs fall below normal-text contrast in Gruvbox dark, while the new matrix only checks the highlight against the composer surface.

### Standards
No documented-standard violations found.

### Spec
The explicit requirements in issue #4471 appear implemented.

Tests were not run, per review policy.

frontend/src/renderer/components/chat/ComposerEditor.tsx
illegalcall: Addressed in c1a4344b2. The composer now sets an explicit inverse semantic pair: selection:bg-foreground with selection:text-background. The E2E reads both computed ::selection colors, composites them in order, and enforces both the 1.5:1 highlight-to-surface floor and the 4.5:1 selected-text-to-highlight floor across all nine palettes in light and dark mode. The new text assertion reproduced the old 45% tint failure at 2.275:1 in Solarized dark before the implementation changed. With the final pair, the limiting Solarized dark ratios are 4.111:1 and 4.748:1 respectively. The focused test passed 3/3 serial repetitions; application and E2E typechecks pass.

illegalcall (commented):
Addressed review 5029694065 in c1a4344b2. The selection now uses an explicit inverse foreground/background semantic pair, and the 18-theme pointer-drag regression enforces both highlight-to-surface contrast at or above 1.5:1 and selected-text-to-highlight contrast at or above 4.5:1. The limiting Solarized dark values are 4.111:1 and 4.748:1. The focused test passed 3/3 serial repetitions, and both frontend typechecks pass.

illegalcall (commented):
## Review verdict: Approved

Ready to merge. The final revision addresses both previous findings: the semantic inverse selection pair keeps the highlight visible and the selected draft readable across supported themes, and the regression covers real pointer selection plus both contrast relationships across all 18 theme/appearance combinations.

### Standards
No findings.

### Spec
All issue #4471 requirements are implemented.

### Correctness
No high-confidence issues found.

Tests were not run, per review policy.

nikhilachale (approved):
Tested locally lGTM Approved

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4474, opened 2026-08-26
feat(mobile): over-the-air JS updates via EAS Update

vickyshaw29:
## What

Adds `expo-updates` + EAS Update so JS-only changes reach installed builds without a store release. Native changes still go through the store.

- Cold start: checks in the background, applies on the next launch. Never blocks the splash.
- Back from 15+ minutes in the background: applies a downloaded update (a JS reload; sessions live in the daemon), otherwise checks silently. Same as Bluesky / CodePush `ON_NEXT_RESUME`. Shorter absences do nothing, so nobody is reloaded mid-prompt. No banner.
- Settings → About → App updates: manual check / restart. Version row and bug reports now carry update id, channel and runtime; the build number is read from the binary (with `appVersionSource: remote` the old row had none).
- Dev builds: `expo-updates` is disabled, so all of this is inert.

## Runtime version

`fingerprint` rather than `appVersion`: native modules have landed on `main` between version stamps, and a fingerprint makes any native change a new runtime automatically, so an update can't reach a build that lacks the native code it calls. `fingerprint.config.js` ignores `google-services.json`, which is gitignored and differs between a laptop, CI and EAS Build; without that the Android hash differs per machine. Verified with `runtimeversion:resolve`.

## Privacy

Each check sends platform, runtime, channel, current update id and a random per-install UUID (`EAS-Client-ID`) to `u.expo.dev`. No user or daemon data. The store privacy policy should name Expo before the next store build.

## Activation

Nothing happens until a build with this config exists. Publishing is a local, deliberate step (`eas update --channel <c> --environment <c> -m "..."`, see the README); there is no CI publish and no Expo token in the repo.

## Verification

- `npm run typecheck`, `npm test` (495, 18 new); `npx expo export` for both platforms
- Release builds on the iOS simulator and the Android emulator, against a local expo-updates protocol server: cold-start download, applied on the next launch, still running after the server reports no update. On iOS a deliberately broken bundle rolled back to the embedded one.
- The Android build's runtime version matched `runtimeversion:resolve` (`2bdcb69b…`) with the placeholder `google-services.json` in place, so the ignore in `fingerprint.config.js` holds end to end
- Debug build: no requests, row reads "Off in development builds"
- Not tried on a device: the 15-minute resume path (unit-tested)


Prasad-D-Ware:
Nice work on this — fingerprint over `appVersion`, the `google-services.json` exclusion, and checking that drafts are persisted before allowing a resume-reload are all the right calls.

One change before merge: **please drop `.github/workflows/mobile-update.yml`.** We'll publish updates locally, the same way we already build releases.

- This repo is public, and `EXPO_TOKEN` is a full-account credential — it's the one key that can push arbitrary JS to every production install without review. Better not to have it in the repo at all.
- With `policy: fingerprint`, the runtime version is computed on whichever machine runs `eas update`. Builds happen locally today, so publishing locally keeps the fingerprint identical on both sides. From CI, any environment drift means the publish succeeds and no device matches.
- We don't want an update per merge to `main` — OTA should be a deliberate release, not a side effect.

Everything else in the PR stays as-is; it all works fine with local publishing. Could you replace the CI paragraph in the README with the local commands (`eas update --channel <c> --environment <c> -m "..."`, plus `update:list` / `update:insights` / `update:rollback`)?

Note `--environment` is required on SDK 55+, and the workflow wasn't passing it — worth knowing for the README even though the file is going away.

vickyshaw29:
Done. workflow dropped in 1051c87, README now documents local publishing with --environment plus update:list / update:insights / update:rollback.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Prasad-D-Ware:
Thanks for dropping the workflow — the README note about publishing from the machine that builds is the part I most wanted written down.

The SDK 57 bump has since landed on `main`, so this branch needs a rebase. I test-merged it: three conflicts, and every API this PR uses is unchanged in `expo-updates@57.0.16` / `expo-application@57.0.2`. So `lib/updates.ts`, `UpdatesManager.tsx`, `appInfo.ts`, `settings.tsx`, `eas.json` and `fingerprint.config.js` all carry over with **no edits**. `@expo/fingerprint@0.20.9` still reads `fingerprint.config.js` / `ignorePaths`, and `checkAutomatically` + `fallbackToCacheTimeout` are still honored.

Four changes, in order:

**1. `packages/mobile/package.json`** — take `main`'s side entirely, then:
- add `"expo-updates": "~57.0.16"` (currently `~29.0.20`; SDK 57 uses unified versioning, so every Expo package carries the SDK major)
- drop the `"expo-application": "~7.0.8"` line — `main` already has `~57.0.2`

Net: this PR contributes one new dependency line here.

**2. `packages/mobile/package-lock.json`** — don't hand-merge. After step 1:
```bash
rm -rf node_modules package-lock.json && npm install   # no --force
```

**3. `packages/mobile/app.json`** — take **`main`'s** version of the conflicted region, which has **no top-level `splash` block** (SDK 57 moved it to the `expo-splash-screen` plugin), then re-insert only the two keys from this PR:

```json
"version": "1.2.1",
"runtimeVersion": { "policy": "fingerprint" },
"orientation": "portrait",
"icon": "./assets/icon.png",
"userInterfaceStyle": "automatic",
"backgroundColor": "#0a0b0d",
"scheme": "aomobile",
"updates": {
  "url": "https://u.expo.dev/2d7dec61-2206-4943-8582-a37e160907ea",
  "checkAutomatically": "ON_LOAD",
  "fallbackToCacheTimeout": 0
},
"ios": { ... }
```

This is the one most likely to go wrong: resolving to `HEAD` looks clean and silently restores the deleted `splash` block, which SDK 57 no longer reads — no error, just a broken splash.

**4. `packages/mobile/README.md`** — add the rollout flag to the production command so 100%-to-everyone isn't the default habit:

```bash
eas update --channel production --environment production \
  -m "fix: ..." --rollout-percentage 10
```

**After the rebase:** `npx expo prebuild --clean`, then a real native build on **both** platforms. `expo-doctor` + `tsc` + `npm test` all passed while `expo-speech-recognition` was broken on the SDK 57 bump, so that trio isn't evidence here. Also worth running `npx expo-updates runtimeversion:resolve` twice and confirming it's stable — the fingerprint changes wholesale on 54→57.

vickyshaw29:
Rebased on main, all four done as listed. Lock was regenerated from scratch, so it also picked up ~120 in-range transitive bumps (expo 57.0.15 → 57.0.17 etc.) — happy to redo it as main's lock + expo-updates only if you'd rather keep the diff small.

Proof after the rebase: prebuild --clean, native builds on both simulators, and the Android OTA cycle against a local server (embedded → served → applied on next launch → stays after a 204). runtimeversion:resolve is stable across runs.

One extra change you haven't seen: fingerprint.config.js now also ignores @react-native-masked-view/masked-view/android/src/main/AndroidManifest.xml. Its own build.gradle rewrites that file at configuration time (strips package= on newer AGP), and it's a fingerprint input — so on 57 the runtime version differed between a clean checkout and a built one, on both platforms (android 8b4366cc → d632ba5d). Caught it because the local server started returning runtime mismatch mid-proof. With the ignore, the hash is identical in both states; it's the only package in node_modules that does this.

expo-doctor's one failure (react-native 0.86.2 vs expected 0.86.3) is main's pin from #4287, not this branch.

Prasad-D-Ware (approved):

PR #4475, opened 2026-08-26
fix(observability): drop request-cancellation noise from daemon Sentry

Pulkit7070:
## Why

With daemon Sentry live, `context.Canceled` and `context.DeadlineExceeded` errors are the dominant, un-actionable churn in the issue feed (client/renderer disconnecting mid-request, or a request/op running past its deadline). They are not server bugs, and they bury the real faults.

## What

`CaptureHTTPError` now drops any error that is or wraps `context.Canceled` / `context.DeadlineExceeded` before capture, via a new `IsNoiseError` (`errors.Is`, so wrapped chains like `list prs for x: context canceled` and the bare `context deadlineExceededError` are both caught).

Genuine slow-path/contention stays observable elsewhere: the aggregated `ao.http.5xx` rollup still counts it, and retryable DB contention has the typed 503 path.

## Tests

- `IsNoiseError` over nil / plain fault / canceled / deadline / wrapped / deep-wrapped.
- Recording-transport test: three cancellation errors produce **zero** captured events; a genuine fault (`no such table: ...`) still reaches the wire.

Observed in production feed: `context canceled` / `context deadline exceeded` across `list prs`, `pr facts`, `skills/list`, `conversation`, and bare deadline errors — all now suppressed.

### DIFF
--- a/backend/internal/observe/sentryobs/noise_filter_test.go
+++ b/backend/internal/observe/sentryobs/noise_filter_test.go
@@ -1,0 +1,87 @@
+package sentryobs
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+)
+
+func TestIsNoiseError(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name string
+		err  error
+		want bool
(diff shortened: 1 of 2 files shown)


i-trytoohard:

PR #4483, opened 2026-08-26
fix(chat): persist chat settings and provider options

ronishrohan:
## Summary
- streamlines the chat settings control, including provider-defined options and Codex permissions
- persists next-turn provider settings across daemon restarts
- serializes live provider option updates and prevents overlapping UI writes so a stale catalog cannot overwrite a newer one
- keeps unclassified provider options available through the More submenu

## Validation
- `cd backend && go test ./internal/service/chat`
- `cd frontend && npm test -- --run src/renderer/components/chat/TurnSettingsBar.test.tsx src/renderer/hooks/useConversation.test.tsx`

## Notes
- Full frontend typecheck remains blocked by unrelated missing `packages/product-ui` dependencies/types in this workspace.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/chatdriver/codexappserver/driver.go
```diff
@@ -297,10 +298,11 @@ func (d *Driver) Resume(ctx context.Context, cfg ports.ChatResumeConfig) (ports.
...
 	policy, sandbox := approvalSettings(cfg.Permissions)
 	params := map[string]any{
-		"threadId":       cfg.ProviderConversationID,
-		"cwd":            cfg.WorkspacePath,
-		"approvalPolicy": policy,
-		"sandbox":        sandbox,
+		"threadId":          cfg.ProviderConversationID,
+		"cwd":               cfg.WorkspacePath,
+		"approvalPolicy":    policy,
+		"approvalsReviewer": approvalReviewer(cfg.Permissions),
```
Annieeeee11, line 308: `ChatResumeConfig.Effort` is now populated from durable conversation settings and ACP resume applies it, but Codex `thread/resume` still only forwards `cfg.Model` (the block just below).

Effort will still land on the next `SendTurn` via `turnSettings()`, so this is not a silent loss — but the live thread will not match the persisted effort until the first message. If the wire supports it, send `cfg.Effort` here the same way as model; otherwise a short comment that Codex is turn-deferred would make the split intentional.

frontend/acp-runtime/package.json
```diff
@@ -4,6 +4,6 @@
 	"private": true,
 	"description": "Pinned ACP adapters packaged with the AO desktop app (provider CLIs are not included)",
 	"dependencies": {
-		"@agentclientprotocol/claude-agent-acp": "0.64.2"
+		"@agentclientprotocol/claude-agent-acp": "0.70.0"
```
Annieeeee11, line 7: This PR also bumps `@agentclientprotocol/claude-agent-acp` 0.64.2 → 0.70.0 (and a lot of transitive lockfile churn). CI is green, but it makes the settings fix harder to revert/bisect.

Is 0.70 required for the catalog/options work here? If not, isolating the bump (or calling it out in the PR body) would help. If it is required, a one-liner in the summary is enough.

Annieeeee11 (commented):
Persistence and the config-option race handling look right, and the UI split between provider execution mode and AO approval policy is a clear improvement. CI is green.

Please address or explicitly defer the inline notes before merge — mainly Codex Resume ignoring `Effort`, the ACP 0.64.2 → 0.70.0 bump riding along, and Codex labeling both `default` and `bypass-permissions` as “Full access”.

Also worth a look: option classification is regex-on-name (a renamed Fast Mode would fall into More), and there is no `SetConfigOption` store-level test covering concurrent writes.

frontend/src/renderer/components/chat/TurnSettingsBar.tsx
```diff
@@ -57,13 +55,30 @@ const APPROVAL_ORDER: ApprovalMode[] = [
...
 ];
 
+// Codex has three distinct permission profiles. Its default is already full
+// access in AO's isolated worktree posture, so expose it as that rather than a
+// fourth, ambiguous "default" option.
+const CODEX_APPROVAL_COPY: Record<ApprovalMode, { label: string }> = {
+	default: { label: "Full access" },
+	"accept-edits": { label: "Ask for approval" },
+	auto: { label: "Approve for me" },
+	"bypass-permissions": { label: "Full access" },
```
Annieeeee11: `default` and `bypass-permissions` both render as “Full access”, and bypass is not a distinct Codex choice — it is the third item in `CODEX_APPROVAL_ORDER`, still labeled the same as default.

A session that actually persisted bypass will look identical to Codex default in this picker. If bypass is unreachable for Codex in AO, that is fine — otherwise give it a distinct label, or map persisted bypass onto the Codex “Full access” value on read so the trigger and the stored mode cannot disagree.

frontend/src/renderer/components/chat/TurnSettingsBar.tsx
```diff
@@ -167,22 +196,14 @@ export function TurnSettingsBar({
 						) : onChange ? (
 							<Picker
 								label={approvalLabel}
-								title="What the agent may do without asking"
+								title="Approval policy for the next turn"
```
Annieeeee11: Provider option controls use `optionDisabled` (`disabled || configPending`), but this native approval picker still uses `disabled` only (the prop on the next line).

A user can change approval while a catalog-replacing PATCH is in flight; both paths write `conversation_settings`, so last write wins. Low probability, but disabling this picker during `configPending` (or sharing one settings write mutex) would close the gap the `configMu` comment describes.

ronishrohan:
Addressed the review feedback in ccf03e96c:

- Resolved the merge conflict against current `main` and kept the newer workspace/history changes.
- Restored durable model and reasoning-effort settings when rebuilding a controller. Codex resume now carries effort through its supported config override.
- Removed the ACP runtime 0.70.0 bump; the PR now matches `main` at 0.64.2.
- Codex now labels `default` as `Full access` and `bypass-permissions` as `Bypass permissions`.
- Provider boolean options are classified by type, so renamed Fast Mode controls do not fall into More.
- Serialized config-option writes and durable settings updates remain covered by the focused service/frontend tests.

Validation: `go test ./internal/service/chat`; targeted Codex adapter tests; 124 focused frontend tests all pass.

Annieeeee11 (commented):
Re-reviewed after `ccf03e96c` — the follow-up addresses most of the earlier inline feedback (Codex effort on resume in the driver, distinct bypass labeling, boolean toggle classification, ACP bump reverted). CI is green and the UI/service direction looks good, but two wiring gaps mean production behavior still won't match what the focused tests assert.

### Blocking: `harness` is never passed in production

`TurnSettingsBar` switches to Codex-specific approval copy when `harness === "codex"`, but `ChatWorkspace` still renders it without `harness={snapshot.harness}`. The Codex label/order tests pass because they set `harness="codex"` explicitly; live Codex sessions will keep the generic `APPROVAL_COPY` vocabulary ("Default approvals", "Auto-approve", etc.) instead of "Full access" / "Approve for me" / "Bypass permissions".

Please wire `harness={snapshot.harness}` at the `TurnSettingsBar` call site.

### Blocking: persisted effort is not forwarded into `Resume`

The driver now supports `ChatResumeConfig.Effort` (Codex carries it via `config.model_reasoning_effort`), and ACP resume applies it in `applyTurnSettings`, but `service.go` still only restores `cfg.Model` from durable settings before calling `driver.Resume`. The resume config literal also omits `Effort`.

After a daemon restart the controller/UI can show the persisted effort from `conversation.Settings`, but the live provider thread won't match until the first `SendTurn` applies `turnSettings()`. The driver test covers effort because it passes `Effort` directly; the service test only asserts in-memory controller settings.

Suggested fix: mirror model restoration for effort and pass it through `ChatResumeConfig`:

```go
if cfg.ProviderConversationID != "" && conversation.Settings.ReasoningEffort != "" {
    cfg.Effort = conversation.Settings.ReasoningEffort
}
// ...
Effort: cfg.Effort,
```

Also worth checking other `ChatResumeConfig` call sites (e.g. `history.go`) if they rebuild resumed sessions.

### Non-blocking

- `configMu` is documented for catalog races, but `service_config_options_test.go` only covers the pure `settingsFromConfigOptions` helper — an integration test with concurrent `SetConfigOption` would lock the mutex behavior in.
- Approval picker uses top-level `disabled` rather than internal `optionDisabled`; fine today because `ChatWorkspace` folds `configOptionPending` into `disabled`, but `optionDisabled` would be safer if this component is reused elsewhere.
- Codex `default` ("Full access") is shown on the trigger but omitted from `CODEX_APPROVAL_ORDER`; probably intentional, but users can't re-select it once changed.

Happy to re-review once the harness + effort wiring land.

ronishrohan:
Follow-up review items addressed in 37a9c5b44:

- Wired `harness={snapshot.harness}` into the production `TurnSettingsBar` call site, so live Codex sessions use Codex-specific approval labels.
- Restored durable reasoning effort in `StartConfig` and forwarded it through the service resume request; the existing persisted-settings test now asserts `ChatResumeConfig.Effort`.
- Kept Codex `default` selectable alongside distinct `bypass-permissions` labeling.
- Applied `configPending` to the native approval picker as well as provider controls.
- Resolved all four earlier inline threads (Codex effort, ACP bump, permission labels, and pending-picker disablement).

Focused validation is green: chat service, Codex adapter resume/turn-setting tests, and 124 frontend tests. The concurrent `SetConfigOption` integration test remains non-blocking follow-up hardening; the service mutex is still in place and the pure durable-settings regression remains covered.

ronishrohan:
Resolved the latest `main` conflict in f33defc3c. The merge keeps the newer controller-owner launch fields and the production settings wiring: `composerSettings` now passes `snapshot.harness`, and resume restores/passes persisted effort.

Validation after the merge: `go test ./internal/service/chat`; 99 focused frontend tests passed. The branch is pushed and GitHub reports it as mergeable; only reviewer approval remains.

Annieeeee11 (approved):
lgtm

PR #4486, opened 2026-08-26
fix(chat): issue browser capabilities to structured chat sessions

### LINKED ISSUE #4452: bug(chat): structured sessions never receive their browser capability
## Bug

Chat-mode sessions launch their structured provider controller with `AO_BROWSER_CAPABILITY` explicitly empty and never persist a `browser_capability_verifier`. As a result, `ao browser` and capability-protected preview-server commands fail for every Chat worker even when the current bundled rewrite binary and Electron browser runtime support them.

**Source:** Independently reproduced in disposable AO-managed Chat session `agent-orchestrator-214` while reassessing the externally hosted process attached to session `agent-orchestrator-211` (211 was not contacted or modified)  
**Reported:** 2026-08-25  
**Analyzed against:** [`dd47a7547f5533a7fb2b109c67eb98c632a47b99`](https://github.com/Untrivial-ai/agent-orchestrator/commit/dd47a7547f5533a7fb2b109c67eb98c632a47b99) (`origin/main`, exact match to the triage checkout)  
**AO version:** freshly built rewrite CLI reports `dev`; bundled daemon reports `dev` and SHA-256 `3bb56a325b7e10ff35560386e118b2a5d2b7c258724c1f7cc9c586d89e251fb3`  
**Environment:** macOS arm64, Go `1.26.4`, daemon healthy/ready on port `3001`  
**Confidence:** **High** — reproduced with an explicit current rewrite binary and traced through issuance, persistence, provider-env injection, authorization, tests, and introducing history.  
**Priority:** **P2** — browser automation is unavailable to all Chat sessions; switching/creating as TUI is a bounded workaround.

## Scope correction: managed Chat vs externally hosted process

Session `agent-orchestrator-211` does **not** establish this managed-Chat defect. Its durable AO row is `tui`, has nonempty runtime handle/launch IDs, and already stores a 43-character browser verifier. The process reporting the missing environment variable is a separately hosted `node_repl` process carrying AO session/worktree context, not the AO-managed TUI runtime that received the launch credential. A recovered/stale bearer presented from that external process correctly failed with `BROWSER_CAPABILITY_INVALID`.

AO has no repository-owned refresh/delegation endpoint that can safely disclose a new bearer to an already-running arbitrary external process, and possession of `AO_SESSION_ID` alone must not authorize browser control. Whether hosted-process attachment should gain a separate authenticated capability-delegation protocol is outside this issue and requires an explicit security design.

This issue remains valid and is scoped only to **AO-managed structured Chat controllers**, independently proven by session `agent-orchestrator-214`: it was created through `ao spawn --mode chat`, has provider conversation and controller generation IDs, has no terminal runtime IDs, received commands through the durable Chat conversation, and reproduced both an empty verifier and missing environment bearer.

## Reproduction

1. Build the exact current rewrite CLI and verify it is talking to the Go daemon, not a legacy PATH binary:

   ```sh
   git checkout dd47a7547f5533a7fb2b109c67eb98c632a47b99
   (cd backend && go build -o /tmp/ao-chat-browser-repro ./cmd/ao)
   /tmp/ao-chat-browser-repro version
   /tmp/ao-chat-browser-repro status --json
   /tmp/ao-chat-browser-repro browser --help
   ```

   Observed: version `dev`; daemon state `ready`, health `ok`, port `3001`; `browser` includes `status`, `open`, `snapshot`, and the other current commands.

2. Spawn a fresh disposable Chat worker using the same absolute binary:

   ```sh
   /tmp/ao-chat-browser-repro spawn \
     --project agent-orchestrator \
     --name triage-chat-browser \
     --harness codex \
     --mode chat \
     --prompt 'Report whether AO_BROWSER_CAPABILITY is set, then run /tmp/ao-chat-browser-repro browser status.'
   ```

3. The fresh controller reports:

   ```text
   AO_BROWSER_CAPABILITY=unset
   ao browser requires the owning session capability (AO_BROWSER_CAPABILITY is not set)
   ```

   The absolute rewrite command exits `2`.

4. Inspect only verifier lengths, never credential values:

   ```sql
   SELECT id, session_mode, length(browser_capability_verifier)
   FROM sessions
   WHERE id = '<fresh-chat-session>';
   ```

   Observed for disposable `agent-orchestrator-214`: `chat | 0`. Existing TUI rows launched on the same daemon have verifier length `43`.

The disposable session was terminated after reproduction.

## Expected

Every AO-managed Chat controller should receive a fresh, unguessable, session-scoped browser bearer in its own provider/runtime environment, while only the one-way verifier is durably stored before the worker can present the token. The owning Chat worker should pass `ao browser status`; another session, a stale launch token, an empty token, and a terminated session should remain unauthorized.

## Root cause

The shared environment builder intentionally clears all browser credentials:

- [`runtimeEnv`](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/manager.go#L3773-L3796) sets `AO_BROWSER_CAPABILITY`, `AO_BROWSER_RUNTIME_TOKEN`, and `AO_BROWSER_RUNTIME_TOKEN_STDIN` to empty values.
- [`launchRuntimeEnv`](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/manager.go#L3799-L3813) is the only layer that calls the browser authority, injects the bearer, and returns its verifier.
- [`persistBrowserCapabilityVerifier`](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/manager.go#L3815-L3830) stores the verifier before process launch to close the eager-worker authorization race.

TUI spawn and restore use the secure launch path and persist the verifier:

- [TUI spawn](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/manager.go#L899-L905)
- [TUI restore](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/manager.go#L1978-L1985)

Every structured-controller launch bypasses it:

- [fresh Chat spawn](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/chat_spawn.go#L128-L147) calls `runtimeEnv` and passes the cleared map to `StartChat`;
- [Chat restore/reconcile](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/chat_spawn.go#L290-L329) repeats the same call;
- [direct Chat agent switching](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/session_manager/agent_switching_chat.go#L27-L40) builds the target environment with `runtimeEnv` and later hands that map to the target `StartChat`.

The provider layer faithfully passes `cfg.Env` into both driver start and resume, so the credential is already absent before the Codex/ACP boundary: [chat service dispatch](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/service/chat/service.go#L340-L370).

Authorization then correctly rejects the request because Chat has neither half of the split credential: [browser service authorization](https://github.com/Untrivial-ai/agent-orchestrator/blob/dd47a7547f5533a7fb2b109c67eb98c632a47b99/backend/internal/service/browser/service.go#L77-L94).

### History

[PR #3472](https://github.com/Untrivial-ai/agent-orchestrator/pull/3472) introduced Chat on 2026-08-06 with `runtimeEnv` as its environment seam. [PR #3473](https://github.com/Untrivial-ai/agent-orchestrator/pull/3473), merged on 2026-08-08 as commit [`c9e1c676b`](https://github.com/Untrivial-ai/agent-orchestrator/commit/c9e1c676baee8ab2e860a8110f2539a460e73a11), split browser credentials into an ephemeral bearer plus durable verifier and changed the terminal spawn/restore paths to `launchRuntimeEnv`; it did not update the already separate Chat launch paths. `git blame` still attributes both Chat calls to PR #3472.

## Affected modes and paths

- **Affected:** fresh Chat sessions for every harness whose structured driver receives `ChatStart.Env` (confirmed live with Codex; generic code path also covers ACP Chat drivers).
- **Affected:** Chat controller resume/reconcile after daemon/controller recovery.
- **Affected:** TUI → Chat interface transitions and Chat source rollback paths that relaunch through `resumeChatController`.
- **Affected:** Chat → Chat agent switching and its target/recovery launches.
- **Not affected:** normal TUI spawn/restore, which already issue, persist, and inject the credential.

## Security requirements for the fix

This is an authorization-availability omission, not evidence of an auth bypass. The existing boundary must remain intact:

- mint a fresh random bearer for each controller launch and bind its verifier to the AO session ID;
- persist **only** the one-way verifier before the provider process can run an eager browser command;
- inject the bearer only into the owning Chat provider environment; never store it in SQLite, metadata/DTO JSON, logs, error envelopes, conversation events, or frontend state;
- continue clearing privileged `AO_BROWSER_RUNTIME_TOKEN*` values so the Electron runtime credential cannot leak to a worker;
- rotate on relaunch so stale controller tokens fail, and preserve cross-session rejection;
- sequence rotation correctly during interface/agent switching: do not invalidate a still-live source controller prematurely, and ensure rollback/recovery receives a newly matching token/verifier pair;
- do not weaken browser/preview authorization or introduce a global/shared capability to make Chat pass.

The current domain field is already `json:"-"`, and the authority uses constant-time comparison with a verifier bound to the session ID; those properties should remain unchanged.

## Required tests

1. Fresh Chat spawn issues exactly one pair, persists the verifier before `StartChat`, injects a nonempty bearer into `ChatStart.Env`, retains the verifier through `ControllerReady`, and authorizes only that session.
2. Project/inherited env cannot supply or override `AO_BROWSER_CAPABILITY` or either privileged browser-runtime secret.
3. Issuer failure or empty credentials aborts Chat spawn before controller start and exercises the existing workspace/session rollback contract.
4. Chat resume/reconcile rotates the pair before provider resume; the new bearer succeeds and the old bearer fails.
5. TUI → Chat transition, Chat → Chat agent switch, source rollback, and restart recovery each end with one matching bearer/verifier pair without invalidating a still-running source during a reversible phase.
6. Cross-session, empty, stale, and terminated-session capabilities remain rejected; no bearer appears in stored session records, API responses, logs, or durable conversation events.
7. End-to-end regression: start the desktop/browser runtime, spawn a fresh Chat worker with an explicit rewrite binary on `:3001`, and verify `ao browser status` succeeds without relying on PATH.

Current tests cover `launchRuntimeEnv` and browser authorization in isolation, and Chat tests assert only that a generic env/session ID reaches the controller. No test asserts issuance or verifier persistence for Chat.

## PATH shadowing is separate

A bare `ao browser status` in the disposable worker also resolved to a legacy binary and returned `unknown command "browser"`. That is **not** the cause established here: the same worker reproduced the missing-capability error with the absolute current rewrite binary. This report therefore does not combine the capability defect with legacy/PATH pinning behavior.

## Recommended implementation scope

Introduce one controller-launch credential preparation boundary shared by fresh Chat launch, Chat resume/reconcile, interface transition, and Chat agent switching. It should return the provider env plus verifier, durably establish the verifier at the correct lifecycle/saga boundary before `StartChat`, preserve/rotate it through controller commits, and explicitly model rollback/recovery ordering. Reuse the existing authority and storage column; no schema/API relaxation should be necessary.

This is not a trivial one-line fix because agent/interface switching has reversible pre-stop phases and post-stop rollback/recovery. No speculative fix PR was opened during triage.

## Verification performed

```text
go build ./...                                                        PASS
go test ./internal/session_manager -run 'Test(RuntimeEnv|ChatSpawnStartsControllerAndNoRuntime|ResumeExitedChatSessionDoesNotRequireTerminalRuntimeHandle)' -count=1  PASS
go test ./internal/service/browser -count=1                          PASS
```

## Duplicate search

Searched open and closed issues and PRs for `AO_BROWSER_CAPABILITY`, `browser_capability_verifier`, `owning session capability`, `chat ao browser`, `chat browser automation`, and `session owned browser`. No duplicate report or fix was found.

## Related

- [Issue #3360](https://github.com/Untrivial-ai/agent-orchestrator/issues/3360) — Chat/TUI architecture and lifecycle invariants.
- [PR #3472](https://github.com/Untrivial-ai/agent-orchestrator/pull/3472) — introduced the structured Chat controller paths.
- [PR #3473](https://github.com/Untrivial-ai/agent-orchestrator/pull/3473) — introduced session-owned browser automation and the split bearer/verifier security model.

aprv10:
## What

Enable `ao browser` commands for AO-managed Chat sessions, including fresh sessions, resumed sessions, interface transitions, and agent switches.

## Why

Chat controllers were launched without `AO_BROWSER_CAPABILITY` and without a persisted verifier, causing every browser command to fail with:

`AO_BROWSER_CAPABILITY is not set`

Fixes #4452.

## How

- Added a shared launch boundary that issues a browser capability and persists its verifier before starting the controller.
- Applied it to Chat spawn, resume/reconciliation, TUI-to-Chat transitions, and Chat agent switching.
- Rotate capabilities only after the previous controller has stopped.
- Keep browser runtime credentials scrubbed and prevent project environment variables from overriding the issued capability.
- No schema, API, frontend, or authorization-rule changes.

## Testing

- Added coverage for fresh Chat sessions, resume rotation, interface transitions, agent switches, rollback recovery, override protection, and issuance failures.
- Focused tests and race tests pass.
- `go build ./...` passes.
- `go vet` passes for the affected packages.
- The full Session Manager package still has unrelated, pre-existing Windows path/PATH test failures.

## Checklist

- [x] Branched from `main`
- [x] Relevant CI checks pass


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
From the NotesBot call on `2026-08-26`, this existing pr appears to be relevant.
Match confidence: `0.53` (fuzzy).

Relevant discussion excerpts:
- [29:59] prateek: Okay. So one thing is let's focus on two pages for now. 2 pages are very important. One is, what would the Orchestrator agent details page look like and how will it behave? One is that it might be very simple, it's just going to be a chat UI with a few navigation controls of going and looking at [...]

Source thread: https://discord.com/channels/1476302178913357958/1542225751246512230/1542225751246512230

backend/internal/session_manager/chat_spawn.go
```diff
@@ -317,6 +323,10 @@ func (m *Manager) resumeChatController(
 	if err != nil {
 		return RestoreResult{}, fmt.Errorf("%s %s: workspace roots: %w", operation, rec.ID, err)
 	}
+	rec, env, err := m.prepareWorkerLaunchEnv(ctx, rec, project.Config.Env)
```
illegalcall: [P1] Capability rotation must be inside the same per-session ownership gate as `StartChat`. Startup reconciliation runs after the API starts serving, so a user resume can pass `HasLiveChatController` before the reconciler publishes. The losing call then persists a second verifier from a stale session snapshot, while Chat Service returns the already-live controller without running this call's `ControllerReady` or applying its environment. The result is token A with verifier B, plus reverted generation/activity; both launches report success but browser calls and controller events are rejected. I reproduced this interleaving deterministically under both normal and `-race` tests. Please span issuance + durable persistence + controller selection with one launch gate, and use a narrow/CAS verifier write.

backend/internal/session_manager/chat_spawn.go
```diff
@@ -130,21 +130,26 @@ func (m *Manager) launchChatController(ctx context.Context, in chatSpawn) (domai
...
-	env := m.runtimeEnv(id, in.cfg.ProjectID, in.cfg.IssueID, in.project.Config.Env)
 	var diffBaseSHA, diffBaseRef string
 	if in.projectKind == domain.ProjectKindSingleRepo {
 		diffBaseSHA, diffBaseRef = resolveSpawnDiffBase(
 			ctx, in.workspace.Path, in.workspace.BaseRef)
 	}
+	// The provider passes this environment through to agent shell commands. The
+	// shared launch boundary pins `ao`, issues the browser bearer, and persists
+	// its verifier before StartChat can run an eager provider process.
+	preparedRecord, env, err := m.prepareWorkerLaunchEnv(ctx, in.record, in.project.Config.Env)
```
illegalcall: [P1] This credential-bearing environment is cached by Chat Service as the controller's start config. `EditMessage` and `ActivateBranch` later start replacement provider/controller generations from that cached `cfg.Env` without issuing or persisting a new pair. Therefore the replaced controller's bearer remains valid after the swap; this is especially risky because closing the old provider is best-effort. Issue #4452 requires a fresh bearer for every controller launch and stale-token rejection, so branch replacement needs to go through the credential rotation boundary too, with rotation ordered after the source is stopped.

backend/internal/session_manager/manager.go
```diff
@@ -3812,8 +3805,27 @@ func (m *Manager) launchRuntimeEnv(id domain.SessionID, project domain.ProjectID
...
+func (m *Manager) prepareWorkerLaunchEnv(
+	ctx context.Context,
+	rec domain.SessionRecord,
+	projectEnv map[string]string,
+) (domain.SessionRecord, map[string]string, error) {
+	env, verifier, err := m.launchRuntimeEnv(rec.ID, rec.ProjectID, rec.IssueID, projectEnv)
+	if err != nil {
+		return rec, nil, err
+	}
+	rec, err = m.persistBrowserCapabilityVerifier(ctx, rec, verifier)
```
illegalcall: This ultimately persists through unrestricted `Store.UpdateSession`, replacing the entire copied session row. On resume/switch that can overwrite concurrent lifecycle facts such as activity, termination, handles, harness, or controller generation; the launch race above deterministically reverts a newly committed generation. Please add a narrow verifier update guarded by the expected owner/generation (through the lifecycle/store boundary) instead of a whole-row write.

backend/internal/session_manager/chat_spawn_test.go
```diff
@@ -213,6 +217,35 @@ func TestResumeExitedChatSessionDoesNotRequireTerminalRuntimeHandle(t *testing.T
 	}
 }
 
+func TestResumeChatRotatesBrowserCapabilityBeforeControllerStart(t *testing.T) {
```
illegalcall, line 590: The issue requires the new bearer to succeed and the old bearer to fail, restart recovery to end with one matching pair, cross-session/empty/stale/terminated rejection, no bearer leakage, and a desktop/browser-runtime E2E. These added tests only compare fake strings in launch maps, so they miss both blocking defects above. Please add an authorization-level rotation test, the concurrent reconciliation/resume regression, branch-replacement rotation coverage, leakage assertions, and the requested E2E.

illegalcall (changes requested):
Requesting changes for two blocking lifecycle/security issues:

1. Concurrent boot reconciliation and user resume can leave the live controller with token A while the session stores verifier B, and can revert the committed generation/activity. This was reproduced deterministically under both ordinary and race-enabled tests.
2. Chat branch replacements (`EditMessage` / `ActivateBranch`) reuse the cached original `AO_BROWSER_CAPABILITY`, so a stale controller token remains authorized after replacement.

The required restart-recovery, authorization, leakage, and desktop/browser-runtime E2E coverage is also incomplete. Separately, the PR currently conflicts with `main` in two test files. Focused package tests, focused race tests, and vet passed.

illegalcall (approved):

PR #4488, opened 2026-08-26
feat: add durable ao report foundation

harshitsinghbhandari:
## Summary

- Model Worker Reports as coordination claims with optional state kept separate from repeatable outputs.
- Support output-only reports and repeated `artifact`, `pr-created`, and `pr-reviewed` outputs without treating outputs as completion.
- Validate structured states, required notes, free-form messages, output references, and pull request URLs across the CLI, daemon API, service, and domain.
- Persist reports and ordered outputs in SQLite with a one-hour informational fallback, a fixed five-minute `done` settlement window, and no session termination side effect.
- Recover claimed reports atomically on daemon restart and fence acknowledgements with claim tokens.
- Regenerate sqlc, OpenAPI, and frontend TypeScript artifacts.

## Verification

- `npm run lint`
- `cd backend && go test -race ./...`
- Focused Worker Reports race suite, including restart recovery and token fencing
- `cd frontend && npm test -- --run`
- `npm run sqlc && npm run api && git diff --exit-code`

## Scope

This PR provides the Phase 1 Worker Reports contract and durable outbox foundation. Report delivery coordination and downstream context rendering remain outside this phase.


harshitsinghbhandari (commented):
## PR Review: #4488

### Summary
This adds the first durable `ao report` slice across CLI, HTTP, service, and SQLite. The individual layers are covered, but the new cross-layer contract is not exercised with their real implementations.

### Testing
The CLI test terminates at a fake HTTP handler, the controller test uses a fake service, and the service test uses a fake store. Because the CLI DTO is intentionally hand-mirrored, these tests can all pass while the actual CLI → controller → service → SQLite flow is broken or silently drifts.

### Boundary Verification
**FAIL — 1 unverified boundary.** `backend/internal/cli/report.go` (JSON producer) → `backend/internal/httpd/controllers/reports.go` (JSON consumer) → `backend/internal/service/report/service.go` → `backend/internal/storage/sqlite/store/report_store.go` has no test using the real implementations together. No deferrals were present.

### Recommendation
**Request Changes.** Add a focused round-trip test that runs the real CLI request against the real router/report service/SQLite store and verifies the persisted report (ideally after reopening the DB). This pins both the manually mirrored wire shape and the durability promise.

backend/internal/cli/report_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := setConfigEnv(t)
+			var got reportAPIRequest
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.Path == "/api/v1/reports" {
+					_ = json.NewDecoder(r.Body).Decode(&got)
+				}
+				w.WriteHeader(http.StatusCreated)
```
harshitsinghbhandari, line 71: This fake handler only decodes into the CLI-local `reportAPIRequest`, so it cannot detect drift from `controllers.CreateReportRequest` or failures in the real controller/service/store handoff. Please add a round-trip test that points the CLI at `httpd.NewRouterWithControl` wired to the real report service and SQLite store, then reads (or reopens and reads) the stored report. That is the load-bearing boundary for this foundation.

harshitsinghbhandari: Addressed in d91f1eb9c. Added TestReportRoundTrip, which runs the real CLI request against httpd.NewRouterWithControl wired to the real report service and SQLite store, then closes and reopens the database and verifies the persisted report type, note, worker/session ownership, project ownership, and pending delivery state.

harshitsinghbhandari (commented):
## PR Review: #4488

### Summary
The follow-up adds the requested real round-trip coverage from the CLI through the production HTTP router and report service into SQLite, then reopens the database and verifies the durable record.

### Boundary Verification
**PASS — 0 unverified boundaries.** `TestReportRoundTrip` now pins the hand-mirrored CLI JSON contract to the real controller and verifies the service/store durability boundary across a database reopen. No deferrals were present.

### Recommendation
**Approve.** The prior blocking coverage gap is resolved, and I found no new correctness issues in the follow-up.

backend/internal/storage/sqlite/migrations/0129_reports.sql
```diff
@@ -0,0 +1,35 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE reports (
```
neversettle17-101, line 3: why are we creating reports as a table? how would this help orchestrator? 
in future we would be adding memory to the orchestrator and that is where the reports information should stay.

harshitsinghbhandari: because we want all reports ever reported to be persisted so they can audited for whether a agent ever reported anything or not. and putting all reports in the memory is bad idea. the reports will also carry metadata that the orchestrator doesn't need to know but once. the second part of reports will be made (currently in ideation), we will need the reports that existed till now.

backend/internal/httpd/controllers/dto.go
```diff
@@ -17,6 +17,29 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/service/systeminstall"
 )
 
+// CreateReportRequest is the worker-to-daemon report submission contract.
+type CreateReportRequest struct {
+	SessionID string `json:"sessionId"`
```
neversettle17-101: This request body lets the caller choose any `sessionId`, and `Create` only checks that the referenced session exists and is a worker. That means any loopback caller can forge `done`/`stuck`/`needs_input` reports for another worker. If the later orchestrator delivery path treats reports as authoritative worker state, this becomes a trust-boundary bug. I think the worker identity needs to come from a daemon-proved caller context, or this API needs to be framed as untrusted local input rather than worker-authored fact.

backend/internal/httpd/controllers/dto.go
```diff
@@ -17,6 +17,29 @@ import (
...
+	Type      string `json:"type" enum:"free_form,pr_created,artifact,checkpoint,needs_input,stuck,done"`
+	Note      string `json:"note" maxLength:"1000"`
+}
+
+// ReportResponse is the created durable report returned to a worker.
+type ReportResponse struct {
```
neversettle17-101: This response shape looks broader than the current slice needs. The only client in this PR mirrors just `report.id`, but the API now commits to echoing `sessionId`, `projectId`, `type`, `note`, `createdAt`, and `deliveryState` as well. That is extra contract surface before there is a demonstrated consumer, and it also sends the free-form note back across another boundary for no clear gain. A minimal created-id response would keep this more to the point.

backend/internal/domain/report.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return nil
+}
+
+// ValidateReportNote checks the common note rules and type-specific PR URL rule.
+func ValidateReportNote(typ ReportType, note string) error {
+	if !typ.Valid() || strings.TrimSpace(note) == "" || utf8.RuneCountInString(note) > MaxReportNoteCharacters {
+		return ErrInvalidReport
+	}
+	if typ == ReportPRCreated && !IsGitHubPullRequestURL(note) {
```
neversettle17-101: `pr_created` is modeled as a generic report type, but this validation makes it GitHub-only. AO already has GitLab/MR-aware flows elsewhere, so this either needs a provider-agnostic PR/MR reference contract or a narrower name that makes the GitHub-only restriction explicit. Otherwise workers on non-GitHub repos will be forced back to `free_form` and the structured report type will drift from AO's actual SCM surface.

backend/internal/httpd/controllers/dto.go
harshitsinghbhandari: Addressed in 1b8556ff5 by documenting the actual trust boundary on CreateReportRequest: sessionId is local-caller attribution under AO's intentionally unauthenticated loopback API, not cryptographic worker proof, and reports do not mutate or derive authoritative session status. Introducing caller authentication here would conflict with the daemon's established loopback boundary; later delivery must continue treating reports as coordination input, not authoritative runtime facts.

backend/internal/httpd/controllers/dto.go
harshitsinghbhandari: Addressed in 1b8556ff5. The 201 response is now the minimal JSON object containing only id. Removed ReportResponse and stopped echoing session/project/type/note/timestamps/delivery state; regenerated OpenAPI and frontend types and updated boundary tests.

backend/internal/domain/report.go
harshitsinghbhandari: Retained intentionally: this batch resolved acceptance contract explicitly requires the pr-created flag to accept an HTTP(S) GitHub pull-request URL. Provider-agnostic PR/MR semantics would broaden the product contract beyond this PR. GitLab workers can use free-form reports until a later design explicitly introduces an MR-aware structured type.

harshitsinghbhandari (commented):
## PR Review: #4488

### Summary
This adds a durable worker-report slice across CLI, HTTP, service, and SQLite, and the real round-trip test covers the hand-mirrored wire contract. The outbox claim lifecycle still has a crash-recovery hole.

### Correctness
A report becomes permanently invisible to delivery after a process dies between claim and acknowledge/release: pending listing excludes it, a second claim rejects it, and there is no expiry or stale-claim recovery operation. That violates the durability guarantee the outbox is meant to provide.

### Testing
Add a restart/stale-claim test that proves an interrupted delivery can be reclaimed without allowing an old claim token to acknowledge the new lease.

### Boundary Verification
**PASS — 0 unverified creation boundaries.** `TestReportRoundTrip` exercises the real CLI → router → service → SQLite path and verifies persistence after reopening the database. No deferrals were present.

### Recommendation
**Request Changes.** Add lease expiry/recovery semantics for claimed reports and cover the crash/restart path before this lifecycle becomes the foundation for coordinator delivery.

backend/internal/storage/sqlite/queries/reports.sql
```diff
@@ -0,0 +1,37 @@
...
+WHERE delivery_state = 'pending' AND available_at <= ?
+ORDER BY created_at, id
+LIMIT ?;
+
+-- name: ClaimReport :one
+UPDATE reports
+SET delivery_state = 'claimed', claim_token = sqlc.arg(claim_token),
+    claimed_at = sqlc.arg(claimed_at), delivery_attempts = delivery_attempts + 1,
+    last_error = ''
+WHERE id = sqlc.arg(id) AND delivery_state = 'pending' AND available_at <= sqlc.arg(claimed_at)
```
harshitsinghbhandari, line 33: `ClaimReport` only accepts `pending`, while `ListPendingReports` also excludes `claimed`; therefore a daemon/process crash after this update but before `AcknowledgeReport` or `ReleaseReport` strands the report forever. `claimed_at` is stored but never used for expiry or recovery, so this is a permanent claim rather than a lease. Please add an atomic stale-claim reclaim/requeue path (with token fencing) and a restart test for it.

harshitsinghbhandari: Addressed in c3a9b1ac3. Daemon startup now atomically requeues every claimed report, clears its stale claim token, and preserves attempt fencing. The restart regression claims a report, reopens the store, runs the production recovery path, verifies the old token cannot acknowledge, and verifies a fresh claim can be acknowledged.

harshitsinghbhandari: Rebased again after upstream claimed migration 128. The Worker Reports migration is now 129, and the current verified head containing this fix is a8e4de254.

harshitsinghbhandari:
Superseded by #5064 so the Worker Reports foundation can participate in GitHub Stacks from an upstream source branch. The replacement uses the same branch name in Untrivial-ai/agent-orchestrator and the exact validated head a8e4de254fb93c685edd314c58b075e6c561fef3. Review and fix history remains here and is linked from the replacement. No approval is claimed as transferred.

PR #4495, opened 2026-08-26
fix(cursor): clear Needs input after approval and gate TUI permissions

nikhilachale:
## Summary

- Add a Cursor-specific activity mapper and AO-owned `beforeShellExecution` / `beforeMCPExecution` hooks that evaluate tool attempts against `AO_PERMISSION_MODE`, post `waiting_input` only when returning `"ask"`, and use after-execution hooks to report `active` again.
- Pin `AO_PERMISSION_MODE` into the runtime env at spawn/restore so hook commands know the session approval policy.
- Fix Chat Cursor sessions staying in **Needs you** after the user approves a tool: `Resolve()` and `approval.resolved` / `input.resolved` events now report `active`.

## Test plan

- [x] `go test ./internal/adapters/agent/cursor/... ./internal/adapters/agent/activitydispatch/... ./internal/cli/... ./internal/service/chat/...`
- [ ] Spawn a Cursor **Chat** session, approve a shell command, confirm the Kanban card leaves **Needs you**
- [ ] Spawn a Cursor **TUI** session with default permissions, confirm shell/MCP attempts show **Needs you** only when AO returns `"ask"`, and clear after execution


backend/internal/adapters/agent/cursor/activity.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// EvaluatePermission decides whether a Cursor shell/MCP attempt needs user
+// approval under AO's permission mode. waiting_input is used (not blocked)
+// because Cursor installs no pre/post-tool-use correlation trio.
+func EvaluatePermission(mode ports.PermissionMode, _ string, _ []byte) PermissionDecision {
+	if permissionRequired(mode) {
+		return PermissionDecision{
+			Permission:     "ask",
+			State:          domain.ActivityWaitingInput,
```
whoisasx: [P1] Keep permission dialogs in the blocked state. Returning `permission: "ask"` means Cursor is showing a decision dialog, not an empty composer. Recording `waiting_input` leaves ordinary Deliver writes admissible, and the confirmation loop explicitly retries Enter through `waiting_input`; a fast permission request can therefore be approved by an AO retry or by a message pasted while the dialog is open. Please report `ActivityBlocked` here and extend the Cursor before/after lifecycle correlation so only evidence that this dialog closed can clear it.

backend/internal/adapters/agent/cursor/hooks.go
```diff
@@ -61,8 +61,12 @@ var cursorManagedHooks = []cursorHookSpec{
 	{Event: "sessionStart", Command: cursorHookCommandPrefix + "session-start"},
 	{Event: "beforeSubmitPrompt", Command: cursorHookCommandPrefix + "user-prompt-submit"},
 	{Event: "stop", Command: cursorHookCommandPrefix + "stop"},
-	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "permission-request"},
-	{Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "permission-request"},
+	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "before-shell-execution"},
+	{Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "before-mcp-execution"},
```
whoisasx: [P1] These callbacks now enforce the user permission boundary, but their generated entries do not set `failClosed: true`. Cursor documents that a command failure, timeout, or invalid JSON proceeds by default, so a failed AO callback can bypass the approval required by default/accept-edits mode. Please make both before-execution hooks fail closed and preserve that field through hook serialization. See https://cursor.com/docs/hooks.

backend/internal/adapters/agent/cursor/hooks.go
```diff
@@ -61,8 +61,12 @@ var cursorManagedHooks = []cursorHookSpec{
 	{Event: "sessionStart", Command: cursorHookCommandPrefix + "session-start"},
 	{Event: "beforeSubmitPrompt", Command: cursorHookCommandPrefix + "user-prompt-submit"},
 	{Event: "stop", Command: cursorHookCommandPrefix + "stop"},
-	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "permission-request"},
-	{Event: "beforeMCPExecution", Command: cursorHookCommandPrefix + "permission-request"},
+	{Event: "beforeShellExecution", Command: cursorHookCommandPrefix + "before-shell-execution"},
```
whoisasx: [P2] Please migrate the AO-owned legacy callback instead of only appending this one. `GetAgentHooks` deduplicates exact commands, so an existing worktree keeps `ao hooks cursor permission-request` and gains this new command. I reproduced both entries after reinstall; the legacy command now emits no permission response, making upgraded worktrees behave differently from fresh installs. Remove the old AO-owned entry for these events and add an upgrade regression test.

backend/internal/service/chat/controller.go
```diff
@@ -1508,6 +1508,7 @@ func (c *Controller) Resolve(ctx context.Context, requestID string, decision por
 		ctx, c.conversation.ID, requestID, string(detail), c.now()); err != nil {
 		return fmt.Errorf("record approval %s: %w", requestID, err)
 	}
+	c.reportActivity(ctx, domain.ActivityActive, "chat.approval.resolved", c.now())
```
whoisasx: [P2] Do not unconditionally report active after resolving one request. The conversation model and store allow multiple pending approvals/structured-input requests; resolving one currently clears Needs input even when another request remains actionable. The direct approval/input paths and provider-resolved event paths should report active only after the durable conversation has no remaining pending interaction, with a regression test covering two concurrent or mixed requests.

whoisasx (changes requested):
I reviewed the Cursor TUI permission flow, Chat approval resolution, runtime environment propagation, upgrade behavior, and the surrounding lifecycle safety contracts. The permission-mode propagation and Cursor response schema are directionally consistent with the existing runtime design, but the inline findings expose two permission-safety gaps plus upgrade and multi-request correctness issues. Please address these before merging.

Verification: the full backend suite passes locally with `go test ./... -count=1` at `5003b217f0cc3110f8c332d8a8bf9748a4af09c2`. GitHub currently reports no status checks, and the two manual scenarios in the PR description remain unchecked.

backend/internal/service/chat/controller.go
nikhilachale: Fixed in 6cde7ccfb. Direct approval/input resolution and provider-resolved event paths now query durable conversation activities and report active only when no pending approval or structured-input interaction remains. Added regressions for two direct approvals and provider-resolved concurrent approvals.

backend/internal/adapters/agent/cursor/activity.go
nikhilachale: Fixed in 6cde7ccfb. Cursor ask decisions now report ActivityBlocked, and lifecycle tracking correlates before/after shell and MCP events by execution family plus bounded tool name. Unmatched after-events stay suppressed, and concurrent dialogs keep the session blocked until every observed request closes. Added matching, unrelated, and concurrent lifecycle regression tests.

backend/internal/adapters/agent/cursor/hooks.go
nikhilachale: Fixed in 6cde7ccfb. AO now serializes failClosed: true on both managed beforeShellExecution and beforeMCPExecution hooks, including upgrades of an existing managed command. The hook-install test asserts the field on both entries.

backend/internal/adapters/agent/cursor/hooks.go
nikhilachale: Fixed in 6cde7ccfb. Hook installation now removes the exact AO-owned legacy permission-request command from both before-execution events before installing the current callback, while preserving custom entries. Added an upgrade regression test covering removal, deduplication, and preservation of custom hook fields.

backend/internal/lifecycle/manager.go
```diff
@@ -889,14 +932,28 @@ func (m *Manager) applyToolPrecedenceLocked(id domain.SessionID, cur domain.Acti
...
-		case isPostToolUseEvent(s.Event) &&
-			fl != nil && fl.blockedCandidate != "" && s.ToolUseID == fl.blockedCandidate:
-			// The single unambiguous blocking tool finished: the dialog was
-			// answered. Clear the candidate so a later dialog in the same turn
-			// starts from a clean slate.
-			fl.blockedCandidate = ""
-			return s
 		default:
+			if fl != nil {
+				if key, ok := cursorAfterExecutionKey(s); ok && fl.cursorPending[key] > 0 {
```
whoisasx: [P2] A denied Cursor permission never reaches a clearing path here. Cursor emits postToolUseFailure with failure_type permission_denied, but cursorPending is keyed from before-shell-execution / before-mcp-execution, while this branch only decrements it for matching after-* events. In a regression using before-shell-execution followed by post-tool-use-failure, the session remained blocked instead of returning to active. Please carry enough denial/correlation information to clear the resolved Cursor dialog while preserving the concurrent-dialog guarantee.

skills/agent-role-smoke/SKILL.md
```diff
@@ -0,0 +1,100 @@
+---
```
whoisasx: [P2] This skill and its 861-line runner are unrelated to the Cursor permission/activity fix and add 961 lines to an otherwise focused PR. Please remove both skills/agent-role-smoke files from this branch and land that functionality separately so its behavior and permission handling can be reviewed independently.

whoisasx (changes requested):
Re-reviewed the latest head. The four earlier functional findings are addressed: Cursor now reports blocked correctly, permission hooks fail closed, legacy hook entries are migrated, and chat activity is reconciled against durable pending interactions. Two issues remain: denying a Cursor permission can leave the session permanently blocked, and this focused PR still contains the unrelated agent-role smoke implementation. All eight GitHub checks are green and the affected package tests pass, but the focused denial regression remains reproducible. Please address these two findings before merging.

backend/internal/lifecycle/manager.go
nikhilachale: Fixed in 15dc5c880. Cursor permission_denied postToolUseFailure payloads now recover the exact shell command or MCP tool key captured by the before-execution hook and emit a correlated denial completion. Lifecycle decrements only that counted pending key, so one denied request remains blocked while any concurrent dialog is still unresolved; malformed or uncorrelated failures remain fail-closed. Added CLI shell/MCP payload regressions and a lifecycle regression covering denial with another pending dialog.

skills/agent-role-smoke/SKILL.md
nikhilachale: Removed from the branch in b47bdb268 and c82e9ff293. Both agent-role-smoke files are absent from the latest head and from the current PR file diff.

backend/internal/cli/hooks.go
```diff
@@ -406,6 +421,46 @@ func (c *commandContext) runHook(ctx context.Context, agent, event string) error
...
+
+	path := "sessions/" + url.PathEscape(sessionID) + "/activity"
+	req := setActivityAPIRequest{
+		State:          string(decision.State),
+		Event:          event,
+		ToolName:       cursor.HookToolName(event, payload),
+		AgentSessionID: hookAgentSessionID(payload),
+		LaunchID:       launchID,
+	}
+	if err := c.postJSON(ctx, path, req, nil); err != nil {
```
whoisasx, line 453: [P1] Please fail closed when this activity write fails for an `ask` decision. The error is currently logged, but the hook then emits `{"permission":"ask"}` and exits successfully. Cursor consequently opens its native permission dialog while AO may still record the session as `active`, so message delivery or automation can write into the dialog. Because `failClosed: true` only applies when the hook itself fails, return an error or an explicit deny unless the blocked state was persisted first, and cover daemon 500/unreachable cases.

backend/internal/adapters/agent/cursor/activity.go
```diff
@@ -0,0 +1,118 @@
...
+// tracking fails closed.
+func PermissionDenialCorrelation(payload []byte) (event, toolName string, ok bool) {
+	var p struct {
+		ToolName    string `json:"tool_name"`
+		FailureType string `json:"failure_type"`
+		ToolInput   struct {
+			Command string `json:"command"`
+		} `json:"tool_input"`
+	}
+	if err := json.Unmarshal(payload, &p); err != nil || p.FailureType != "permission_denied" {
```
whoisasx: [P2] This correlation only handles `permission_denied`, but Cursor also terminates tool attempts through `postToolUseFailure` with `error`, `timeout`, and interrupts. If no matching specialized after-event arrives, the command-keyed `cursorPending` entry remains and AO stays blocked after the dialog has already been answered. Please correlate every terminal Shell/MCP failure outcome to the matching pending key and add regressions for errors, timeouts, interrupts, and concurrent requests.

backend/internal/adapters/agent/cursor/hooks.go
```diff
@@ -43,15 +44,17 @@ type cursorHookFile struct {
 }
 
 type cursorHookEntry struct {
```
whoisasx, line 45: [P2] Please preserve user-owned hook entries as raw JSON. Decoding every entry on the newly managed events into only `command` and `failClosed`, then serializing it again, drops supported fields such as `matcher`, `timeout`, `type`, and `prompt`; a prompt hook can be rewritten into an empty command. AO should inspect or modify only exact AO-owned commands while retaining all fields on unrelated entries.

whoisasx (changes requested):
Thanks for addressing the earlier review rounds. The denied-permission correlation now clears only the matching pending request, concurrent dialogs remain protected, and the unrelated smoke skill has been removed. Three current issues still need to be addressed before approval: the permission hook can open an untracked dialog when the activity write fails, terminal error/timeout/interrupt outcomes are not correlated back to pending Cursor requests, and rewriting managed event arrays drops supported fields from user-owned hooks. CI is green, but these are runtime and configuration-preservation gaps not covered by the current checks.

backend/internal/adapters/agent/cursor/activity.go
nikhilachale: Fixed in 9493e8209 and expanded in d40cf7112. postToolUseFailure now correlates Cursor permission_denied, error, timeout, and is_interrupt outcomes to the exact Shell command or MCP tool key. Lifecycle decrements only that pending key, keeps unrelated and duplicate same-key dialogs blocked, and clears only after all observed dialogs resolve. The regression matrix covers all four outcomes for both Shell and MCP plus different-key and same-key concurrency.

backend/internal/cli/hooks.go
nikhilachale: Fixed in 9493e8209. Cursor ask decisions now return a hook error and emit no permission response when persisting blocked activity fails, so failClosed applies instead of opening an untracked dialog. Added regressions for daemon HTTP 500 and an unreachable daemon; both assert an error and empty stdout.

backend/internal/adapters/agent/cursor/hooks.go
nikhilachale: Fixed in 9493e8209 and hardened in d40cf7112. Cursor hook entries are retained as raw JSON, and AO rewrites failClosed only on exact AO-owned commands. matcher, timeout, type, prompt, and other user fields survive install and uninstall. AO ownership now uses exact current/legacy command membership rather than the ao hooks cursor prefix, with a regression proving a prefix-sharing user command and a prompt hook remain unchanged.

whoisasx (approved):
Re-reviewed the latest head. The previously reported permission-safety issues are addressed: ask decisions fail closed when blocked activity cannot be persisted, terminal denial/error/timeout/interrupt outcomes correlate back to the exact pending Shell or MCP request, and user-owned Cursor hook entries retain their complete JSON shape with exact AO ownership matching. I did not find a new code-level blocker, and all eight GitHub checks are green.

Approving from my side. The two manual Cursor scenarios in the PR description are still unchecked; please complete or attach evidence for those before merging.

PR #4496, opened 2026-08-26
fix(scm/github): make rate-limit errors trigger observer cooldown

Pulkit7070:
GitHub `RateLimitError` exposed `ResetAt`/`RetryAfter` as fields but not methods, so the observer's `errors.As` (which needs `GetRetryAfter()`/`GetResetAt()`) never matched for GitHub. Result: a GitHub rate-limit never triggers the bounded cooldown and the daemon keeps polling every 30s, worsening exhaustion. GitLab already implements the interface, so the fleet degrades asymmetrically. Masked in tests by a local double that implements the getters.

Fix: add the two getters mirroring `gitlab/client.go`, plus a regression test that a real `*github.RateLimitError` satisfies the observer's cooldown contract via `errors.As`.

Found in the pre-10k QA audit (finding GH-1).


i-trytoohard:

PR #4497, opened 2026-08-26
fix(gitworktree): retry worktree removal on macOS/Linux, not only Windows

Pulkit7070:
The bounded worktree-remove retry was gated to Windows (`removeAllRetryEnabled = runtime.GOOS == "windows"`), but the transient race it absorbs is not Windows-only. On macOS/Linux a just-killed agent child can still hold the worktree dir briefly after `ao session kill` destroys its shells, so `os.RemoveAll` fails with a transient `unlinkat` EACCES/ENOTEMPTY/EBUSY. Gated to Windows, that stranded the worktree and returned a 500 on Unix.

Observed in production as `kill X: gitworktree: remove unregistered path: unlinkat ...` and `retire replacement: force destroy` (the two are among the top daemon errors).

Fix: default the retry on for every platform. The ctx-bounded ~7.25s budget and the early ErrNotExist exit are unchanged; a caller that gives up cancels ctx. The var is kept so tests still exercise the disabled path; drops the now-unused `runtime` import; adds a guard test that the default stays enabled.

QA audit finding SESS-1.

### DIFF
--- a/backend/internal/adapters/workspace/gitworktree/remove.go
+++ b/backend/internal/adapters/workspace/gitworktree/remove.go
@@ -4,18 +4,18 @@ import (
 	"context"
 	"errors"
 	"os"
-	"runtime"
 	"time"
 )
 
-// Worktree removal races process exit on Windows. A PTY (or any agent child)
-// rooted in the worktree can still hold a handle on that directory for a short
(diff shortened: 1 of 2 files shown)


i-trytoohard:

PR #4498, opened 2026-08-26
fix(session): map ErrNoController to a typed 409 on the send path, not a raw 500

Pulkit7070:
A send to a session whose chat controller is not live (daemon just restarted, controller stopped, mid-reconnect) returned `chatsvc.ErrNoController`, which had no case in `toAPIError` and fell through to a raw 500 `INTERNAL_ERROR`, shipping `send <id>: no live chat controller for session` to the client and Sentry. The conversation-send path already maps this to 409 `CHAT_CONTROLLER_NOT_READY`; the session-send path (orchestrator relay, `ao send`) diverged.

Fix: add the `ErrNoController` case to `toAPIError` returning the same typed 409, plus a regression test. No import cycle (service/chat does not import service/session).

QA audit finding CHAT-1 (production error AO-DESKTOP-5).


i-trytoohard:

Issue #4352, opened 2026-08-24
fix(observability): harden daemon Sentry capture after PR #4346


illegalcall:
## Bug

[PR #4346](https://github.com/Untrivial-ai/agent-orchestrator/pull/4346) adds a strong first version of daemon Sentry reporting for 5xx responses and panics. Before the daemon DSN is enabled broadly, the capture boundary needs a focused hardening pass so events are request-local, singular, and classified from typed error semantics.

**Source:** review of [PR #4346](https://github.com/Untrivial-ai/agent-orchestrator/pull/4346)  
**Analyzed against:** PR head `781a41245d326dad8af513f06900a20f9d2038c1`; main at review time `48045a10fbff7fe344618efb692f0e684478ba4a`  
**Confidence:** High — the exact capture paths and Sentry hub behavior were traced in code  
**Priority:** P2 — observability can be lost, duplicated, or attributed to the wrong request; normal AO execution is unaffected and the feature remains DSN-gated

### 1. Concurrent captures share the global Sentry hub

`backend/internal/observe/sentryobs/sentryobs.go:105-135` accepts a `context.Context` but ignores it and calls the package-level `sentry.WithScope` and `sentry.CaptureException` helpers.

Those helpers operate on `sentry.CurrentHub()`. `WithScope` is implemented as push, callback, and pop on the hub's shared scope stack. Two requests can interleave those operations, causing one event to capture with the other request's top-most scope. Request IDs, routes, error codes, severity, and fingerprints can therefore be mixed.

#### Reproduction test

1. Configure a recording Sentry transport.
2. Start two captures concurrently with distinct request IDs, paths, error codes, and fingerprints.
3. Force their scope/capture operations to interleave.
4. Inspect the recorded envelopes.

Each event must contain only its own metadata and fingerprint.

### 2. A recovered panic can be captured twice

`backend/internal/httpd/recover.go:54-63` captures the panic with its Go stack and writes a 500 response. The outer 5xx logger in `backend/internal/httpd/log.go:88-106` then observes that 500. Because it has no captured controller error, it creates a generic `HTTP 500 ...` error and sends a second Sentry event.

One panic can consequently become two issues with different fingerprints: one useful panic event and one generic duplicate.

#### Reproduction test

1. Run a panicking handler through the real recovery and logging middleware order.
2. Use a recording Sentry transport.
3. Assert that exactly one event is emitted.
[...]

PR #4506, opened 2026-08-26
fix(observability): harden daemon Sentry capture (re-land #4352)

Pulkit7070:
Re-lands the daemon-Sentry hardening from #4352 that never reached main: a PR-head desync left #4346 merging the pre-hardening commit, so `origin/main` still has the shared-hub capture, double panic-capture, and all-503-suppressed policy. This is the exact verified change from that work, cherry-picked onto current main.

## Fixes
1. **Request-isolated hub.** `CaptureHTTPError`/`CapturePanic` used package-level `sentry.WithScope`/`CaptureException` on the process-global `CurrentHub`, so concurrent requests could interleave scope push/pop and stamp one request's request_id/fingerprint onto another's event. Each capture now resolves a hub from the request context or clones the current hub. Covered by a `-race` test that fails if reverted.
2. **Single panic capture.** The recover middleware captured a panic (fatal, with the Go stack) and the outer 5xx logger then manufactured a second generic `HTTP 500` event with a different fingerprint. Recovery now marks the request via a request-scoped flag and the logger skips its capture when set. Exactly one event, verified through a recording transport.
3. **Code-aware 503.** `ShouldCaptureStatus(status)` dropped every 503, but AO also uses 503 for typed outages (`SCM_UNAVAILABLE`, `BROWSER_RUNTIME_UNAVAILABLE`, `DEVICE_REGISTRY_UNAVAILABLE`). Replaced with `ShouldCapture(status, errorCode)`: only the deliberate `SERVICE_UNAVAILABLE` backpressure is suppressed.

Adds `Disable()` for shutdown/tests. All tests pass under `go test -race`.

Note: overlaps `internal/observe/sentryobs/sentryobs.go` with the noise-filter PR #4475; whichever merges second needs a trivial rebase.

### DIFF
--- a/backend/internal/httpd/log.go
+++ b/backend/internal/httpd/log.go
@@ -34,6 +34,9 @@ func requestLogger(log *slog.Logger, sink ports.EventSink) func(http.Handler) ht
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
 			r, capturedErr := envelope.WithErrorCapture(r)
+			// Install the single-capture marker before the inner recover
+			// middleware runs, so a recovered panic captures exactly once.
+			r = withSentryCaptureState(r)
 			start := time.Now()
 			defer func() {
 				attrs := []any{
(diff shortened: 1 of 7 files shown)


i-trytoohard:

PR #4509, opened 2026-08-26
feat(preview): suggest detected ports in the Browser panel

Pyasma:
## What

The Browser panel offers the ports a session is actually serving on, next to the address bar. Clicking one opens it the way `ao preview <url>` does. Same view from the CLI via `ao preview ports [--json]`.

## Why

Finding a dev server's port meant reading it out of terminal scrollback and typing it into the address bar.

## How

Detection is scoped to a session two ways, and a process qualifies on either: it descends from the session's root process, or its working directory is inside the session's worktree. Both are needed — the root process differs by mode (a tmux pane's shell for terminal sessions, the provider process for chat), and neither survives an agent backgrounding a dev server, since a detached process is reparented to init. The worktree does not move and belongs to one session.

`internal/portscan` reads `/proc` on Linux and shells out to `lsof` elsewhere, then credits each port to the deepest process holding it — a server started from a shell is reported by every ancestor that inherited the descriptor.

Everything fails open: a missing `lsof`, an unreadable process table, a permission denial, or a session with no live process all yield an empty list, never an error. Windows detects nothing for now (`portscan/windows.go`); its socket enumeration lands separately.

`GET /api/v1/sessions/{id}/preview/ports` carries no capability header — it starts nothing — and is blocked on the LAN listener instead (`httpd/lan_listener.go:84`), so the list never leaves the machine.

## Testing

`go build ./...`, `go vet`, and `go test` pass on `internal/portscan`, `internal/cli`, `internal/httpd/...`, `internal/session_manager`. Cross-compiled `portscan` for `GOOS=windows` and `GOOS=darwin`. `BrowserPanel.test.tsx` passes 58/58. Not verified in the running desktop app.

## When Server is not Live

image

## When Server is Live

image

## Checklist

### DIFF
--- a/backend/internal/portscan/tree.go
+++ b/backend/internal/portscan/tree.go
@@ -1,0 +1,234 @@
+package portscan
+
+import (
+	"context"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// maxDetected caps a suggestion list. A session with more listening ports than
+// this has something unusual going on, and a toolbar dropdown is not the place
+// to enumerate it.
+const maxDetected = 20
+
+// Detected is one port a session is serving on, with a display label for the
+// process that owns it. Command is a convenience for the reader, never an
+// identifier, and may be empty.
+type Detected struct {
+	Port    int
+	PID     int
+	Command string
+}
+
+// process is one row of the machine's process table.
+type process struct {
+	PID     int
+	PPID    int
+	Command string
+}
+
+// Detect returns the TCP ports one session is serving on.
+//
+// A session is identified two ways, and a process qualifies on EITHER:
+//
+//   - it descends from rootPID, the session's own entry point (a tmux pane's
+//     shell, or the provider process a chat session spawned); or
+//   - its working directory is inside workspace, the session's worktree.
+//
+// The second anchor is not redundant. Agents routinely start a dev server in
+// the background, and a backgrounded process calls setsid and is reparented to
+// init the moment its launching shell exits. From then on it descends from
+// nothing AO owns, so tree scoping alone reports nothing. Its working directory
+// does not move, and a worktree belongs to exactly one session, so cwd keeps
+// the scan correctly scoped after the tree connection is gone.
+//
+// Best effort throughout: an unreadable process table, a missing enumeration
+// tool, or a permission denial all yield an empty list rather than an error,
+// because "no suggestions" is a normal outcome that every caller renders.
+func Detect(ctx context.Context, rootPID int, workspace string) []Detected {
+	procs := processTable(ctx)
+	if len(procs) == 0 {
+		return nil
+	}
+	candidates := map[int]bool{}
+	if rootPID > 0 {
+		for pid := range descendants(procs, rootPID) {
+			candidates[pid] = true
+		}
+	}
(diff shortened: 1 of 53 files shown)


frontend/src/renderer/components/BrowserPanel.tsx
```diff
@@ -542,6 +551,44 @@ export function BrowserPanelView({
...
+						    the page -- invisible, and clicks land on the page instead. */}
+						<DropdownMenuContent align="end" data-browser-native-overlay="true">
+							{detectedPorts.map((detected) => (
+								<DropdownMenuItem
+									aria-label={t("browser.openDetectedPort", { port: detected.port })}
+									key={detected.port}
+									// localhost, not 127.0.0.1: dev servers with host
+									// allowlists (Vite's default among them) accept the
+									// name and reject the literal address.
+									onSelect={() => openSessionLink(`http://localhost:${detected.port}`)}
```
codebanditssss, line 581: **Blocking:** The scanner reports only a port and discards the socket's bind address, but this always navigates to `localhost`. A process listening only on a specific LAN address or alternate loopback address will appear in this menu and then open an unreachable URL. Please preserve a usable host/URL through `boundPort` -> API DTO -> UI (mapping wildcard binds to `localhost`) and cover LAN-only plus IPv6 listeners.

backend/internal/portscan/tree.go
```diff
@@ -0,0 +1,234 @@
...
+// A plain string prefix would let "/w/app-old" match a workspace of "/w/app",
+// so the boundary must fall on a separator.
+func underWorkspace(cwd, workspace string) bool {
+	if cwd == "" || workspace == "" {
+		return false
+	}
+	if cwd == workspace {
+		return true
+	}
+	prefix := strings.TrimSuffix(workspace, string(filepath.Separator)) + string(filepath.Separator)
```
codebanditssss, line 84: `/proc/<pid>/cwd` and `lsof` return resolved paths, while the stored workspace path can contain symlinks (including through an overridden or symlinked AO data directory). Comparing the raw strings means a detached server under that workspace is missed once the process-tree anchor is gone. Please canonicalize both paths before this boundary check, with a symlinked-workspace test.

frontend/src/renderer/hooks/useDetectedPreviewPorts.ts
```diff
@@ -0,0 +1,38 @@
...
+		queryKey: detectedPreviewPortsQueryKey(sessionId),
+		queryFn: async (): Promise<DetectedPreviewPort[]> => {
+			const { data, error } = await apiClient.GET("/api/v1/sessions/{sessionId}/preview/ports", {
+				params: { path: { sessionId } },
+			});
+			if (error || !data) return NO_PORTS;
+			return data.ports ?? NO_PORTS;
+		},
+		enabled: enabled && sessionId !== "",
+		refetchInterval: DETECTED_PORTS_POLL_MS,
```
codebanditssss, line 35: **Non-blocking:** This five-second interval drives a machine-wide `ps`/cwd scan on every request; on macOS that also launches two `lsof` processes. Please benchmark this path or add short-lived server-side caching/singleflight so visible or popped-out panels do not create sustained process-table scan churn.

Issue #4414, opened 2026-08-25
bug(workspacewatch): one missing git-listed directory aborts the whole workspace/events subscription with a 500


Pulkit7070:
## Summary

`GET /api/v1/sessions/{id}/workspace/events` returns 500 and the live-file SSE subscription never starts when the session's git worktree references a directory that no longer exists on disk. One missing directory aborts the whole subscription.

## Root cause

For a git worktree, `addInitialDirectories` (`backend/internal/workspacewatch/watcher.go:152-193`) builds the watch set from `git -C <root> ls-files -z --cached --others --exclude-standard` (`watcher.go:120`): for every listed file it adds the file's directory and all ancestors up to root **without stat'ing them**, then calls `watcher.Add(dir)` on each. `fsnotify.Add` does an `lstat`, so a directory that git still lists (via the index, `--cached`) but that was removed from disk — an unstaged deletion, a sparse/partial checkout, or a dir deleted while the request is in flight — returns `no such file or directory`.

The loop returns on the **first** such failure (`watcher.go:186-192`) with no per-directory tolerance, so the entire subscription aborts. The untyped error reaches `streamWorkspaceChanges` (`backend/internal/httpd/controllers/sessions.go:622`) and `envelope.WriteError`, which has no typed mapping for it and defaults to `http.StatusInternalServerError` (`backend/internal/httpd/envelope/envelope.go:77`). The workspace *root* is already guarded (`os.Stat` at `watcher.go:85` -> typed 404); this is specifically the partially-present-worktree case (root exists, a git-listed subdirectory does not).

## Reproduction (verified)

```
git init <root>; mkdir <root>/sub; echo x > <root>/sub/file.txt; git -C <root> add sub/file.txt
rm -rf <root>/sub                      # tracked dir deleted, deletion not staged
# git -C <root> ls-files --cached  still prints  sub/file.txt
workspacewatch.Watch(ctx, <root>)      # or open the session Files view -> GET .../workspace/events
```

Observed:

```
watch workspace directory "<root>/sub": lstat <root>/sub: no such file or directory
```

The endpoint returns 500 and the subscription never establishes. Combined with the no-backoff EventSource reconnect (#4323) this repeats in a tight loop.

## Expected

A directory that has gone missing should be **skipped** (log-and-continue) so the subscription still establishes for the directories that exist. A partially-present worktree should not 500 the whole live-files stream.

## Affected code

[...]

PR #4510, opened 2026-08-26
fix(workspacewatch): degrade when directory watches fail

### LINKED ISSUE #4476: bug(workspacewatch): one-fd-per-directory watch exhausts the fd limit on large workspaces, aborting workspace/events with a 500
## Summary

On a workspace with many directories, `GET /api/v1/sessions/{id}/workspace/events` fails with `watch workspace directory "...": open ...: too many open files` and the live-file SSE subscription never starts. The watcher registers one OS watch per directory (one fd on macOS/kqueue, one inotify watch on Linux), so a large tree (node_modules, vendored deps, a monorepo) exhausts the process/system fd limit.

## Root cause

`addInitialDirectories` (`backend/internal/workspacewatch/watcher.go:152-193`) collects **every** directory in the workspace — for a git worktree, the directory of each `git ls-files` entry plus all ancestors; otherwise a full `filepath.WalkDir` — and calls `watcher.Add(dir)` on each with no bound on the count. fsnotify opens an fd per watched directory (kqueue) / consumes an inotify watch, so once the number of directories exceeds `RLIMIT_NOFILE` (or the inotify `max_user_watches`), `watcher.Add` returns `too many open files` (EMFILE/ENFILE). The loop returns on the **first** such failure (`watcher.go:186-192`); the untyped error reaches `streamWorkspaceChanges` (`backend/internal/httpd/controllers/sessions.go:622`) -> `envelope.WriteError` -> 500, and the no-backoff EventSource reconnect (#4323) then retries in a loop.

## Reproduction (verified)

Create a git worktree with 400 directories each containing a tracked file, lower `RLIMIT_NOFILE` to 128 to model a daemon already near its ceiling (a real machine hits the macOS default of 256 with a typical `node_modules`), then call `workspacewatch.Watch(ctx, root)`:

```
watch workspace directory ".../pkg264": open .../pkg264: too many open files
```

The endpoint returns 500 and the subscription never establishes.

## Why it matters

Unlike #4414 (a single missing directory, fixable by skip-and-continue), skip-and-continue does **not** help here: the watcher fundamentally registers an unbounded number of per-directory watches, so any sufficiently large workspace fails, and the descriptor pressure can degrade the whole daemon/machine, not just this endpoint.

## Expected / options

- Bound the number of watches, or watch coarsely (watch the root recursively where the platform supports it, or watch a capped/most-relevant set and reconcile) rather than one-per-directory unbounded.
- At minimum, when `watcher.Add` fails with EMFILE/ENFILE, skip that directory and continue (partial coverage) instead of aborting the whole subscription with a 500, and return a typed degraded response rather than an untyped 500.

## Affected code

- `backend/internal/workspacewatch/watcher.go:152-193`
- `backend/internal/httpd/controllers/sessions.go:622`
- `backend/internal/httpd/envelope/envelope.go` (untyped error -> 500)

## Related

- #4414 (same watcher aborts on a single missing directory) — sibling root cause in the same loop.
- #4323 (no-backoff SSE reconnect amplifies a transient setup 5xx).
- #4325 (untyped service errors collapse to 500 INTERNAL_ERROR).

ApexYash11:
Fixes #4476

## Summary
- Bound initial per-directory workspace watches and register shallowest-first so large workspaces degrade toward leaf coverage instead of exhausting descriptors and failing subscription startup.
- Skip and warn on unwatchable initial or newly-created directories, covering EMFILE/ENFILE pressure and stale git-listed directories from #4414.
- Add regression coverage for stale git-listed directories and the watch-count cap.

## Test
- `cd backend; $env:GOCACHE = "D:\agent-orchestrator\.gocache"; go test ./internal/workspacewatch/`
- `cd backend; $env:GOCACHE = "D:\agent-orchestrator\.gocache"; go build ./...`
- `cd backend; $env:GOCACHE = "D:\agent-orchestrator\.gocache"; go test ./...`

Note: full `go test ./...` was interrupted after several minutes because it appeared hung after emitting unrelated Windows/environment failures (missing POSIX `sh`, `/tmp` path assumptions, Windows mode assertions, missing reviewer binaries, temp-dir permission issues).


backend/internal/workspacewatch/watcher_test.go
```diff
@@ -150,3 +151,52 @@ func runGit(t *testing.T, dir string, args ...string) {
...
+		t.Fatalf("new watcher: %v", err)
+	}
+	defer func() { _ = watcher.Close() }()
+
+	previous := maxWatchedDirectories
+	maxWatchedDirectories = 4
+	defer func() { maxWatchedDirectories = previous }()
+
+	if err := addInitialDirectories(context.Background(), watcher, root, gitWorkspace{}); err != nil {
+		t.Fatalf("addInitialDirectories over the watch cap: %v", err)
```
Pulkit7070, line 300: The fix is solid, and I confirmed it locally: build/vet clean and the whole workspacewatch package (including both new regression tests) passes under -race. The shallowest-first sort + cap + warn-and-continue degradation reads correctly, and skipping an unwatchable directory instead of aborting the subscription is the right call.

One thing worth strengthening: this test only asserts that addInitialDirectories returns no error when the tree exceeds the cap. It never verifies the two behaviors the fix actually adds:
- that the cap is enforced, e.g. `len(watcher.WatchList()) <= maxWatchedDirectories` (fsnotify exposes WatchList());
- that the retained watches are the shallowest ones (the deliberate degradation-toward-leaves ordering).

As written, removing the `sort.Slice` or the truncation would leave this test green, so the core behavior isn't locked. Consider asserting the watched set is exactly the cap size and contains the shallow directories (and not the deep `nested/deeper` ones).

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Pulkit7070:
A few things worth addressing, none blocking:

- The 4096 cap only bounds the initial walk. `addCreatedTree` keeps calling `watcher.Add` for every newly created directory with no accounting against the cap, so a workspace that grows past 4096 directories during a session drifts back toward descriptor exhaustion. Degrade-and-continue keeps that from failing hard, but the cap does not actually bound total watches, which the comment implies.
- On a host whose fd or inotify limit is below the workspace directory count, `addInitialDirectories` logs one WARN per failed `Add`, and `addCreatedTree` does the same per created subtree: a burst of near-identical WARN lines per subscription. Logging once with a skipped count (or at debug) would keep it readable.
- Dropped directories no longer trigger the model refresh they guarded. The comment says that refresh stays reachable through other invalidation paths, but nothing tests that, so a regression there would silently reduce coverage. A test asserting an unwatched directory's changes still reach the read model would lock the claim down.

ApexYash11:
@Pulkit7070  re-review !!

Pulkit7070:
Re-reviewing the shared-limiter change: the cap is now a lifetime cumulative budget rather than a concurrent one, and I think that's a regression worth fixing before merge.

`dirWatchLimiter.used` is only ever incremented (in `add`) and never decremented - there's no release path, and `handleEvent` acts only on `fsnotify.Create` (calling `addCreatedTree`); it never handles `Remove`/`Rename`, so a deleted watched directory's slot is never reclaimed. Over a long-lived session in a workspace with directory churn (build outputs, `node_modules`, branch switches that create and delete many directories), `used` climbs monotonically. Once it reaches the 4096 cap, `add` returns early for every subsequent directory even though the number of *currently* watched directories may be far below the OS limit. From that point, newly created directories silently go unwatched and their changes stop triggering the read-model refresh - and because the runtime `warn` only fires on `defer` at stream shutdown, there's no signal while it's happening.

Reclaiming budget when a watched directory goes away would fix it: handle `Remove`/`Rename` for directories by calling `watcher.Remove` and decrementing `used`. If a lifetime budget is intended instead, the comment on `maxWatchedDirectories` should say so, since it currently reads as a bound on concurrent watches.

ApexYash11:
@Pulkit7070 re-review!!

backend/internal/workspacewatch/watcher.go
```diff
@@ -9,15 +9,103 @@ import (
...
+	if err := l.watcher.Add(dir); err != nil {
+		l.skipped++
+		l.lastError = err
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return nil
+	}
+	l.watched[dir] = struct{}{}
+	l.used++
```
Pulkit7070, line 81: [correctness] `add` is not idempotent: for a directory already in `watched`, the map write above is a no-op but `l.used++` still runs, so `used` double-counts and the `used == len(watched)` invariant breaks.

This is reachable in exactly the churn scenario this PR targets. `addCreatedTree` WalkDirs a newly created tree and adds every subdirectory, then fsnotify delivers a separate Create event for each of those same subdirectories (nested `mkdir -p`, a tarball extract, a branch checkout), so the inner dirs get re-added. Each re-add leaks one budget slot that `release` can never reclaim (it deletes the single map entry and decrements `used` once), so over a churny session `used` drifts above the real watch count and the cap starts skipping directories early: the same premature-skip problem, reintroduced.

I confirmed it by execution: adding the same directory twice leaves `used=2` with a `watched` set size of 1.

Guarding the top of `add` on already-watched fixes it, and it is safe because `fsnotify.Add` is itself idempotent for an existing watch:

    if _, ok := l.watched[dir]; ok {
        return nil
    }

A regression test asserting `used == len(watched)` after a re-add would lock it in.

Pulkit7070: Concrete fix: add an already-watched guard at the top of `add` so the accounting cannot double-count. `fsnotify.Add` is already idempotent for an existing watch, so this only corrects the `used` bookkeeping:

```go
func (l *dirWatchLimiter) add(ctx context.Context, dir string) error {
	l.mu.Lock()
	defer l.mu.Unlock()
	if _, ok := l.watched[dir]; ok {
		return nil // already watched; fsnotify.Add is a no-op, do not re-count
	}
	if l.used >= l.limit {
		l.skipped++
		return nil
	}
	// ... unchanged
}
```

Regression test that reproduces the leak and locks the invariant in:

```go
func TestLimiterAddIsIdempotent(t *testing.T) {
	w, err := fsnotify.NewWatcher()
	if err != nil {
		t.Fatal(err)
	}
	defer w.Close()
	l := newDirWatchLimiter(w, 4096)
	dir := t.TempDir()
	_ = l.add(context.Background(), dir)
	_ = l.add(context.Background(), dir)
	if l.used != len(l.watched) || l.used != 1 {
		t.Fatalf("used=%d watched=%d, want 1/1", l.used, len(l.watched))
	}
}
```

I ran this test against the current code: it fails with `used=2 watched=1`. With the guard added, it passes.

Pulkit7070 (approved):
Approving. Both of my earlier findings are addressed: the shared cap now reclaims budget on Remove/Rename, and add() is idempotent for already-watched dirs (the guard prevents the used-counter double-count I flagged), with TestLimiterAddIsIdempotent and TestHandleEventReclaimsWatchBudgetOnRemoveAndRename covering both. Package passes -race -count=3 locally, race-clean.

Ayash-Bera:
Answering the validity question from #4476: the bug is still present on current main and this PR still addresses it. Verified today on head 759a56ea1.

- Main still aborts: `addInitialDirectories` walks every directory with no bound and returns the first failed `watcher.Add`, which surfaces as a 500 on the SSE endpoint and retries via the no-backoff reconnect loop.
- Reproduction on main: a 530k-directory workspace (exceeds the host inotify watch ceiling) makes `Watch` return `watch workspace directory ...: no space left on device`. Same code path as the macOS `too many open files` case in the issue.
- Same reproduction on this head: `Watch` succeeds, changes directly under the watched root still notify, directories beyond the cap are silently uncovered, and no error reaches the endpoint.
- Local checks on this head: `go build ./...`, `go vet ./...`, `go test -race -count=3 ./internal/workspacewatch/`, and golangci-lint v2.12.2 (0 issues) pass. Full `go test ./...` fails only in packages that fail identically on main (missing bundled tmux runtime, PTY integration flakes on this box); all nine GitHub checks on this head are SUCCESS.

Two non-blocking observations from the pass, for visibility only:
1. The aggregated WARN skipped count omits directories dropped by the pre-add truncation in `addInitialDirectories`, so a >4096-dir workspace whose 4096 attempts all succeed logs skipped=0. Cosmetic; the cap is the intended silent bound.
2. On rename of a watched directory, `release()` drops the kernel watch, so the moved subtree is uncovered until another watched path changes. Deliberate per the review discussion on budget reclaim.

The head moved after the 2026-09-06 approval (the main sync merge), so merge likely needs the approval refreshed; nothing else blocks.

backend/internal/workspacewatch/watcher.go
```diff
@@ -209,20 +304,45 @@ func addInitialDirectories(ctx context.Context, watcher *fsnotify.Watcher, root
...
+	ordered := make([]string, 0, len(dirs))
 	for dir := range dirs {
-		if err := watcher.Add(dir); err != nil {
-			if ctx.Err() != nil {
-				return ctx.Err()
-			}
-			return fmt.Errorf("watch workspace directory %q: %w", dir, err)
+		ordered = append(ordered, dir)
+	}
+	sort.Slice(ordered, func(i, j int) bool {
```
Pulkit7070: Git metadata parent directories added just above (git.metadataFiles, and the ancestor walk over git.files) are folded into this same set and sorted shallowest-first here, so under degradation they are dropped purely by depth. .git/refs/heads/<branch> is about three levels under .git, so it is among the first watches skipped once Adds start failing. Because fsnotify takes one descriptor per directory and the macOS default RLIMIT_NOFILE is 256 (far below the 4096 cap), hitting the OS limit and skipping the deepest directories is the routine case. The consequence: on a degraded workspace a commit that only advances the current branch ref (no HEAD rewrite, no other tracked-file change) writes into an unwatched directory; fsnotify is non-recursive so no watched ancestor sees it, and the Git read model stays stale until something else fires. Git worktrees are worse, since the metadata dir sits outside root and filepath.Rel yields a ../.. path counted as very deep, so it truncates even sooner. Register git.metadataFiles parents ahead of the leaf-directory budget so the refs the read model depends on are watched first.

Separately, this comparator calls directoryDepth (filepath.Rel) inside the sort, which is O(n log n) allocations on exactly the large trees this change targets. Precompute depth once per directory before sorting.

backend/internal/workspacewatch/watcher.go
```diff
@@ -209,20 +304,45 @@ func addInitialDirectories(ctx context.Context, watcher *fsnotify.Watcher, root
...
+		iDepth := directoryDepth(root, ordered[i])
+		jDepth := directoryDepth(root, ordered[j])
+		if iDepth != jDepth {
+			return iDepth < jDepth
+		}
+		return ordered[i] < ordered[j]
+	})
+	if len(ordered) > maxWatchedDirectories {
```
Pulkit7070: This truncates against the package global maxWatchedDirectories rather than limiter.limit. Harmless today since they carry the same value, but if a limiter is ever constructed with a different limit the pre-truncation and the limiter's own budget will silently disagree. Use limiter.limit here so the two stay in lockstep. (Also note the deferred runtime warn re-reports the cumulative skip count already logged at the initial stage, since skipped is never reset between stages.)

PR #4515, opened 2026-08-26
fix(landing): honor updater manifest eligibility

AgentWrapper:
## Summary
- select landing-page stable and nightly downloads only from releases carrying the complete updater manifest set for that channel
- scan the same 100-release window used by the in-app nightly resolver, with a one-minute cache, and fail closed on incomplete inventory
- remove moving `/releases/latest/` fallbacks and hide channels/platforms that have no eligible release
- add regression coverage for manifest removal, stable fallback, nightly suppression, semver ordering, and exact asset URLs

With the current updater kill switch, the landing page resolves stable downloads to `v0.12.6` and does not render nightly downloads.

## Verification
- `cd frontend && npx vitest run src/landing/src/lib/releases.test.ts --config vitest.config.ts` (7 tests)
- `cd frontend/src/landing && npx tsc --noEmit`
- `cd frontend/src/landing && npm run build`
- live `/download` HTML: only `releases/download/v0.12.6/`; no nightly label, `/releases/latest/`, or v0.12.7 URL

## Risk
- GitHub inventory failures now intentionally suppress desktop downloads instead of guessing a moving release URL.
- Eligibility changes propagate on the landing page within the one-minute ISR window.


i-trytoohard:
@AgentWrapper Marking as stale: no activity for 21 days. Comment or push and this comes off the close list, activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4517, opened 2026-08-26
feat: add selectable Windows terminal shells

### LINKED ISSUE #4505: feat(windows,shell-terminals): selectable default shell for standalone terminals (Git Bash, PowerShell, cmd)
## Feature Description

On Windows, standalone shell terminals always launch whichever shell the daemon finds first in a **hardcoded probe order** — `pwsh.exe` → `powershell.exe` → `ComSpec` → `cmd.exe`. There is no way for the user to choose a different default shell, and **Git Bash is never considered at all**.

Requested: a user-facing option (app setting and/or per-terminal picker) to select the default shell for standalone terminals, with Git Bash as a first-class choice alongside PowerShell / cmd.

## Current behavior (code anchors)

- `backend/internal/service/shellterm/loginshell.go` — `resolveUserLoginShell()` → `resolveWindowsShell()`: hardcoded candidate list `["pwsh.exe", "powershell.exe"]` then `ComSpec`/`cmd.exe` fallback. No user preference is consulted, no Git Bash probe (e.g. `C:\Program Files\Git\bin\bash.exe`), no override hook.
- `backend/internal/httpd/controllers/dto.go` — `OpenShellTerminalInput` has no shell/shellPath field, so `POST /api/v1/shell-terminals` cannot express a shell choice per open request either.
- `frontend/src/main.ts` (`runLoginShell`, ~line 635) — login-shell env detection runs a shell too; a changed default should be honored consistently there.
- Unix already prefers `$SHELL` (`loginshell.go:24`), which is why this reads as Windows-only — but note the Unix sibling bug with non-POSIX shells: #3788 (fish).

## Motivation

- Reported by a Windows user who lives in Git Bash: **SahilG** in the `#feature-requests` Discord thread *"default terminal option"* — currently the **most upvoted feature request** there.
- Git Bash is a daily-driver shell for many Windows developers; getting PowerShell instead makes the standalone terminal pane a pane they immediately `exit` out of.
- PowerShell-first probe order also surprises cmd users the same way (SahilG perceived "only cmd is supported" — either way, the user loses).

## Proposed Solution

1. **Setting:** add a `shell` preference (e.g. `"auto" | "git-bash" | "pwsh" | "powershell" | "cmd" | custom path`) to app settings, surfaced in the desktop app's Settings UI.
2. **Resolution order:** `resolveUserLoginShell()` should consult the preference first; `"auto"` keeps today's behavior. Git Bash detection: `where git` → `<Git>\bin\bash.exe` (or `bash.exe` on PATH), launched with `--login -i`.
3. **Wire API:** optionally extend `OpenShellTerminalInput` with an optional per-request `shell` override so the UI can offer a per-terminal picker (⊞ new-terminal dropdown) without extra round-trips.
4. **Validation:** reject-with-message or fall back to `auto` when the configured shell path doesn't exist at open time (consistent with the "must not open something unusable" rule in `loginshell.go:17-19`).
5. **Regenerate API artifacts** (`npm run api`) if DTOs change.

## Alternatives Considered

- Environment-variable-only override (e.g. `AO_SHELL`) — cheap but undiscoverable; fine as a power-user extra, not a substitute for a setting.
- Doing nothing / documenting PowerShell-first — rejected: most-upvoted request, and the fix is well-bounded.

## Acceptance Criteria

- [ ] Windows user can set Git Bash as the default shell; new standalone terminals launch in Git Bash (`--login -i`)
- [ ] PowerShell (pwsh + Windows PowerShell) and cmd remain selectable; `"auto"` preserves current behavior
- [ ] Custom shell path supported, with clear error/fallback when it's missing
- [ ] Preference honored by the login-shell env detection path (`runLoginShell`), not just terminal spawn
- [ ] Unix behavior unchanged (`$SHELL` still wins)
- [ ] OpenAPI spec + `frontend/src/api/schema.ts` regenerated if request DTOs change

## Environment

- OS: Windows (feature is Windows-facing; Unix already honors `$SHELL`)
- Surface: standalone shell terminals (`/terminals` screen) — shellterm service → daemon → ConPTY

_Filed from Discord thread "default terminal option" (#feature-requests). Reporter: @SahilG; picked up by @Vaibhaav-Tiwari._

Vaibhaav-Tiwari:
## What

- add a Windows-only Default terminal setting that matches the existing AO settings UI
- support Auto, Git Bash, PowerShell, Windows PowerShell, Command Prompt, and a custom executable
- pass the optional shell selection through the generated daemon API
- resolve Git Bash as an interactive login shell and safely fall back to Auto with a daemon warning when a selection is unavailable
- persist and localize the preference across all current renderer locales

## Why

Fixes #4505.

Standalone AO terminals on Windows previously had no selectable default shell. This adds the choice without changing Unix behavior or moving shell resolution into the Electron frontend.

## Scope notes

- The selector applies to standalone shell terminals. The API also accepts an explicit per-open override, but this PR intentionally does not add a per-terminal dropdown.
- runLoginShell remains unchanged because the current Windows implementation explicitly skips that environment probe; standalone terminal creation is the feature boundary requested here.
- The primary daemon listener and all network boundaries are unchanged.

## Verification

- go test ./internal/service/shellterm ./internal/httpd/controllers -run 'TestResolveWindowsShell|^TestShellTerminalsAPI_' -count=1
- npm test -- src/shared/ui-locale.test.ts src/main/ui-settings.test.ts src/renderer/stores/terminal-shell-store.test.ts src/renderer/components/GlobalSettingsForm.test.tsx src/renderer/i18n/renderer-coverage.test.ts (47 tests passed)
- npm run typecheck
- npm run api
- git diff --check

The broader local suites contain existing Windows/environment failures unrelated to this feature, so verification was kept focused on the changed boundaries as requested.

## Manual testing

Real Electron testing is pending because another active AO development checkout currently owns the shared port 3002 and ~/.ao/dev/electron profile and has live sessions. It was left running to avoid disrupting unrelated work. Manual Windows shell launches can be completed once that instance is stopped.


backend/internal/service/shellterm/loginshell.go
```diff
@@ -17,27 +19,56 @@ import (
...
+	var resolved []string
+	switch strings.ToLower(preference) {
+	case "git-bash":
+		resolved = resolveGitBash()
+	case "pwsh":
+		resolved = resolveKnownWindowsShell("pwsh.exe")
+	case "powershell":
+		resolved = resolveKnownWindowsShell("powershell.exe")
+	case "cmd":
+		resolved = resolveKnownWindowsShell("cmd.exe")
```
illegalcall: **Blocking:** The explicit `cmd` branch only checks `PATH`. If `%ComSpec%` points to a valid Command Prompt but `cmd.exe` is absent from `PATH`, this falls into automatic resolution, which probes `pwsh.exe` and `powershell.exe` before `%ComSpec%`; selecting Command Prompt can therefore launch PowerShell. Please honor `%ComSpec%` for the explicit `cmd` selection before falling back to Auto, and add a regression test for this environment.

frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
```diff
@@ -56,6 +60,65 @@ function SessionInterfaceRow() {
...
+	return (
+		<>
+			<SettingsRow label={t("settings.terminalShell.label")}>
+				<SettingsOptionMenu
+					aria-label={t("settings.terminalShell.label")}
+					value={preference.kind}
+					options={shellOptions}
+					disabled={saving}
+					onChange={(kind) => {
+						void setPreference(kind === "custom" ? { kind, path: customPath } : { kind });
```
illegalcall, line 99: **Blocking:** Issue #4505 explicitly requires the preference to be honored by the `runLoginShell` environment-detection path, not only terminal spawn. This persists the renderer preference, but `frontend/src/main.ts` is untouched and `ensureShellEnv()` still returns immediately on Windows. Please implement that acceptance criterion, or get the issue scope amended before merge.

backend/internal/service/shellterm/service.go
```diff
@@ -223,7 +223,11 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
 	if err != nil {
 		return ShellTerminal{}, err
 	}
-	argv := resolveUserLoginShell()
+	argv, usedFallback := resolveUserLoginShell(in.Shell)
+	if usedFallback {
+		s.log.Warn("requested shell terminal executable was unavailable; using automatic Windows shell resolution",
```
illegalcall: **Blocking:** An unavailable configured/custom path only produces this daemon warning; the request still succeeds with Auto and the renderer receives no indication of the fallback. Settings continues to display the invalid path while a different shell launches. Please reject/surface the fallback through the API or normalize the saved preference to Auto so the user-visible selection matches behavior.

frontend/src/renderer/hooks/useShellTerminals.ts
```diff
@@ -92,6 +94,10 @@ export function useOpenShellTerminal() {
 			const body: OpenShellTerminalInput = {};
 			if (projectId) body.projectId = projectId;
 			if (sessionId) body.sessionId = sessionId;
+			if (isWindowsPlatform()) {
+				await useTerminalShellStore.getState().load();
+				body.shell = shell ?? terminalShellRequestValue(useTerminalShellStore.getState().preference);
```
illegalcall, line 107: **Blocking test gap:** This is the critical preference-to-daemon join, but no hook test asserts that a saved named/custom shell reaches `POST /api/v1/shell-terminals`, that an explicit per-open override wins, or that non-Windows requests omit `shell`. The store and controller tests cover either side separately, so this boundary can regress while all submitted tests remain green. Please add those cases.

frontend/src/renderer/stores/terminal-shell-store.ts
```diff
@@ -0,0 +1,62 @@
...
+type TerminalShellState = {
+	preference: TerminalShellPreference;
+	loaded: boolean;
+	saving: boolean;
+	saveError: boolean;
+	load: () => Promise<void>;
+	setPreference: (preference: TerminalShellPreference) => Promise<void>;
+};
+
+let settingRevision = 0;
```
illegalcall, line 18: **Non-blocking maintainability:** This revision-fenced load/save state machine duplicates `sound-notifications-store.ts` (`settingRevision`, `pendingLoad`, and the same save/error fencing). There are now two real consumers, so consider extracting a small shared persistent-setting helper to keep their race and failure semantics aligned.

illegalcall (changes requested):
Requesting changes for four blocking review items: Command Prompt selection can fall through to PowerShell, the issue-required runLoginShell integration is missing, unavailable custom paths are not surfaced to the user, and the renderer-to-API preference boundary is untested. The persistent-setting duplication is noted separately as non-blocking. Focused tests and type-checking pass, but the PR also currently conflicts with main.

backend/internal/service/shellterm/loginshell.go
Vaibhaav-Tiwari: Fixed in e22a16d7d. Explicit cmd selection now honors ComSpec when cmd.exe is not on PATH, with a regression test covering that case.

frontend/src/renderer/components/settings/GeneralSettingsSection.tsx
Vaibhaav-Tiwari: Fixed in e22a16d7d. Windows login-shell environment probing now resolves the persisted terminal shell preference, including Git Bash, PowerShell, pwsh, cmd, and custom paths, before daemon startup.

backend/internal/service/shellterm/service.go
Vaibhaav-Tiwari: Fixed in e22a16d7d. An explicitly unavailable shell now returns SHELL_TERMINAL_SHELL_UNAVAILABLE before spawning; the renderer normalizes the saved preference back to Automatic so Settings matches the effective behavior.

frontend/src/renderer/hooks/useShellTerminals.ts
Vaibhaav-Tiwari: Fixed in e22a16d7d. Added renderer hook-to-API boundary tests covering persisted shell selection, explicit overrides, non-Windows behavior, and unavailable-shell normalization.

frontend/src/renderer/stores/terminal-shell-store.ts
Vaibhaav-Tiwari: Thanks. I kept the persistence refactor out of this PR to keep the review fix scoped to the requested Windows shell feature; the existing store tests cover the current state-machine behavior.

Vaibhaav-Tiwari:
Rebased onto current main and resolved merge conflicts in the shell-terminal service and renderer hook tests. Focused frontend and backend checks pass locally; requesting re-review on the updated head 5eb7cca4b.

illegalcall (approved):

Issue #3409, opened 2026-07-31
[Bug]: Windows: Codex resolver prefers `codex.cmd` from anywhere on PATH over `codex.exe`, but supervise cannot launch `.cmd` — session exits silently in ~1.5 s with no child process


faogomes:
## Summary

Two defects that only fail in combination, making every Codex session on an affected Windows machine exit silently ~1.5 s after spawn, with nothing run:

1. **Resolution:** AO's Codex binary resolution searches the *whole* PATH for `codex.cmd` before `codex.exe`. Any `codex.cmd` anywhere on the daemon's PATH wins — even when a directory containing the native `codex.exe` comes first in PATH order. (A VS Code extension shim at `%USERPROFILE%\.local\bin\codex.cmd` is enough to trigger it.)
2. **Launch:** AO's supervised launch chain (`ao pty-host → ao agent-process supervise → <agent>`) cannot start a `.cmd` child at all. A 100 ms process monitor during a real spawn shows `pty-host` and `supervise` starting and the ConPTY being created (`conhost --headless`), then **no child process of any kind** — no `cmd.exe`, no `codex.exe`. Supervise reports process-exited ~200 ms later; that exit is the session's only signal, so the session shows `exited` with nothing executed.

Notably, the daemon's own preflight (`codex login status`) runs the same `.cmd` fine — the defect is specific to the supervised agent launch.

## Steps to reproduce

1. Windows, AO 0.11.2, a daemon PATH where some directory contains `codex.cmd` (e.g. a VS Code extension shim or npm shim), and the native `codex.exe` directory listed *earlier* in PATH.
2. Spawn a Codex worker session.
3. Session exits ~1.5 s after spawn; `first_signal_at` equals the exited timestamp; no hooks fire, no file changes.

**Proof by inversion:** start the daemon with a PATH containing the native `codex.exe` directory and *no* `codex.cmd` anywhere → the same spawn goes `working` (hooks firing), completes its task, and `ao session kill` returns cleanly.

Ruled out by direct test: `codex.exe` stays alive ≥20 s under `ao pty-host` alone, under pty-host + supervise, with AO's full argv (hooks + `trust_level` + ~7 KB `developer_instructions`), in a git-worktree cwd, with and without a live daemon. Codex requires a TTY and AO's ConPTY provides one — a ConPTY-dimensions hypothesis was refuted.

## Expected behavior

[...]

PR #4519, opened 2026-08-26
fix: validate Goose binary identity

### LINKED ISSUE #4516: bug(agents): goose harness availability detection matches the pressly/goose DB migration tool (name-only binary resolution)
## Bug
The goose **agent harness** availability detection matches the **pressly/goose database migration tool** because both ship a binary named `goose`. On systems that have only the migration tool, AO's agent inventory reports the **Goose AI agent as "installed"** (visible in orchestrator/agent settings), and orchestrators can then pick goose for worker spawns — launching the migration CLI with Block-Goose arguments.

**Source:** Discord `#bug-triage` thread (2026-08-26) | **Reported by:** @Akash
**Analyzed against:** `c4abb3d0e` (origin/main, fetched 2026-08-26 IST)
**Confidence:** High — exact code path traced; reporter independently confirmed the false "installed" entry in his orchestrator settings.

## Reproduction
1. Install only the DB migration tool: `brew install goose` (pressly/goose) or equivalent, so `goose` on PATH is the migration CLI. Do **not** install Block's Goose agent.
2. Run the desktop app / `ao agent list` (or open orchestrator settings → agents).
3. Goose is listed under **installed** agents.
4. Observed by reporter: the orchestrator attempted to spawn a goose worker (first time ever) and failed/misbehaved, because the launched binary is the migration tool, not the agent.

## Root Cause
Binary resolution for the goose harness is **name-only** — first executable named `goose` wins, with no identity/version probe:

- `backend/internal/adapters/agent/goose/goose.go:259` — `gooseBinarySpec` lists `Names: []string{"goose"}` plus fallback paths `/usr/local/bin/goose`, `/opt/homebrew/bin/goose`, `~/.cargo/bin/goose`, and npm-global locations. Several of these are exactly where the **migration tool** installs (Homebrew's `goose` formula is the pressly migration tool → `/opt/homebrew/bin/goose`).
- `backend/internal/adapters/agent/binaryutil/binaryutil.go:94` — `ResolveBinary` does `exec.LookPath(name)` over the spec names and returns the **first** hit. There is no check that the resolved binary is the Block Goose CLI rather than the migration tool.
- `backend/internal/adapters/agent/goose/goose.go:276` — `ResolveGooseBinary` is a thin wrapper over that resolver.
- `backend/internal/service/agent/service.go` — the inventory `refresh()` and `FindInstalledBinary()` treat a successful resolve as **installed**. The service's own doc comment (line 23) states this by design: *"only resolves executable paths; it never starts an agent CLI or probes"* — so the false positive is a structural gap, not a regression.
- The inventory is advisory UI metadata, but it is what surfaces Goose as a choice in orchestrator settings, and the orchestrator prompt advertises `--agent <name>` for workers (`backend/internal/session_manager/prompt.go:194`). A name collision therefore propagates from PATH → inventory → "installed" → orchestrator spawn attempt.
- The launch command built for the harness (`goose run [--system ...] -t "" --interactive`, `goose.go:104`) targets Block-Goose CLI semantics and does not exist in the migration tool, so any spawn against the wrong binary is broken by construction.
- The reviewer adapter injects the same resolver (`backend/internal/adapters/reviewer/goose/goose.go:54`), so it inherits the same false positive.

## Fix
Add an identity check to goose binary resolution instead of trusting the name:
1. After a candidate path resolves (PATH or fallback), run a bounded probe (e.g. `<binary> version` / `--help` with a short timeout) and require a Block-Goose-specific marker (e.g. `run`/`session`/`recipe` subcommands or the Block version output format). pressly/goose exposes `up`/`down`/`status`/`create` instead — distinguishable.
2. Recommended shape: an optional `ValidateIdentity func(ctx, path) bool` field on `binaryutil.BinarySpec` so the whole resolver fleet can adopt per-harness identity checks; goose opts in first.
3. There is precedent for bounded CLI probes in the daemon: `backend/internal/adapters/agent/authprobe/authprobe.go` (`CLIStatus`) already runs agent CLIs with timeouts for auth status — the same pattern applies to identity.
4. On identity failure, report goose as **not installed** in the inventory (and surface a hint like "found `goose` on PATH but it is the DB migration tool").

## Impact
- False "installed" state for the Goose agent on any machine with the pressly migration tool (common on dev boxes with DB work).
- Orchestrators can select goose for workers based on that false availability, producing broken spawns.
- Desktop startup prerequisite gate (`FindInstalledBinary`) can pass on the migration tool alone, deferring the real "no usable harness" failure to spawn time.
- Generalizes: any harness whose binary name collides with an unrelated CLI is affected by the same name-only resolution (see related codex resolver bug below).

## Related
- [#3409](https://github.com/Untrivial-ai/agent-orchestrator/issues/3409) — another binary-resolver robustness bug (codex `.cmd` vs `.exe` preference); same resolver class, different failure mode.

adot-7:
## What

Add Goose binary identity validation to prevent pressly/goose from being detected as the Goose AI harness.

## Why

Fixes #4516.

## How

Added a bounded, cancellable `goose --version` probe that accepts Block Goose and rejects the pressly migration CLI.

## Testing

- `go test ./internal/adapters/agent/goose`
- `go test -race ./internal/adapters/agent/goose`
- `go vet ./internal/adapters/agent/goose`
- `git diff --check`

## Checklist

- [x] Branched from `main`
- [x] Follows repository conventions and PR hygiene
- [x] Tests added for pressly rejection and Block Goose acceptance

### DIFF
--- a/backend/internal/adapters/agent/binaryutil/binaryutil.go
+++ b/backend/internal/adapters/agent/binaryutil/binaryutil.go
@@ -72,14 +72,15 @@ type WinPath struct {
 	Parts []string
 }
 
-// ResolveBinary returns the path to spec's binary, searching PATH then the
-// platform's candidate install locations. It returns a wrapped
-// ports.ErrAgentBinaryNotFound when nothing matches, so callers surface a clear
-// "command not found" rather than launching an empty argv. ctx cancellation is
-// honored between probes.
-func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
+// ResolveBinaryCandidates returns spec's ordered binary candidates. PATH is
+// searched first, followed by the platform's configured paths, package-manager
+// paths, and (when enabled) version-manager paths. Candidates are deduplicated
+// while preserving their first occurrence. The returned paths are not all
+// necessarily executable; callers that need to validate a candidate should do
+// so themselves. ctx cancellation is honored while discovering candidates.
+func ResolveBinaryCandidates(ctx context.Context, spec BinarySpec) ([]string, error) {
 	if err := ctx.Err(); err != nil {
-		return "", err
+		return nil, err
 	}
 
 	names := spec.Names
@@ -87,16 +88,28 @@ func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
 		names = spec.WinNames
 	}
 
+	candidates := make([]string, 0)
+	seen := make(map[string]struct{})
+	appendCandidate := func(candidate string) {
+		if candidate == "" {
+			return
+		}
+		if _, ok := seen[candidate]; ok {
+			return
+		}
+		seen[candidate] = struct{}{}
+		candidates = append(candidates, candidate)
+	}
+
 	for _, name := range names {
 		if err := ctx.Err(); err != nil {
-			return "", err
+			return nil, err
 		}
 		if path, err := exec.LookPath(name); err == nil && path != "" {
-			return path, nil
+			appendCandidate(path)
 		}
 	}
 
-	var candidates []string
 	if runtime.GOOS == "windows" {
 		home, _ := os.UserHomeDir()
 		appData := os.Getenv("APPDATA")
@@ -114,22 +127,46 @@ func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
 			if base == "" {
 				continue
 			}
-			candidates = append(candidates, filepath.Join(append([]string{base}, wp.Parts...)...))
+			appendCandidate(filepath.Join(append([]string{base}, wp.Parts...)...))
 		}
-		candidates = append(candidates, WindowsPackageManagerBinCandidates(executableNames(spec)...)...)
+		for _, candidate := range WindowsPackageManagerBinCandidates(executableNames(spec)...) {
+			appendCandidate(candidate)
+		}
 	} else {
-		candidates = append(candidates, spec.UnixPaths...)
+		for _, candidate := range spec.UnixPaths {
+			appendCandidate(candidate)
+		}
 		if home, err := os.UserHomeDir(); err == nil {
-			candidates = append(candidates, joinAll(home, spec.UnixHomePaths)...)
-			candidates = append(candidates, UnixPackageManagerBinCandidates(home, spec.Names...)...)
+			for _, candidate := range joinAll(home, spec.UnixHomePaths) {
+				appendCandidate(candidate)
+			}
+			for _, candidate := range UnixPackageManagerBinCandidates(home, spec.Names...) {
+				appendCandidate(candidate)
+			}
 			if spec.NodeManaged {
 				nodeManagerCandidates, err := UnixNodeManagerBinCandidates(ctx, home, spec.Names...)
 				if err != nil {
-					return "", err
+					return nil, err
 				}
-				candidates = append(candidates, nodeManagerCandidates...)
+				for _, candidate := range nodeManagerCandidates {
+					appendCandidate(candidate)
+				}
 			}
 		}
+	}
+
+	return candidates, nil
+}
+
+// ResolveBinary returns the path to spec's binary, searching PATH then the
+// platform's candidate install locations. It returns a wrapped
+// ports.ErrAgentBinaryNotFound when nothing matches, so callers surface a clear
+// "command not found" rather than launching an empty argv. ctx cancellation is
+// honored between probes.
+func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
+	candidates, err := ResolveBinaryCandidates(ctx, spec)
+	if err != nil {
+		return "", err
 	}
 
 	for _, candidate := range candidates {
--- a/backend/internal/adapters/agent/binaryutil/binaryutil.go
+++ b/backend/internal/adapters/agent/binaryutil/binaryutil.go
@@ -164,6 +164,23 @@ func ResolveBinaryCandidates(ctx context.Context, spec BinarySpec) ([]string, er
 // "command not found" rather than launching an empty argv. ctx cancellation is
 // honored between probes.
 func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
+	if err := ctx.Err(); err != nil {
+		return "", err
+	}
+
+	names := spec.Names
+	if runtime.GOOS == "windows" {
+		names = spec.WinNames
+	}
+	for _, name := range names {
+		if err := ctx.Err(); err != nil {
+			return "", err
+		}
+		if path, err := exec.LookPath(name); err == nil && path != "" {
+			return path, nil
+		}
+	}
+
 	candidates, err := ResolveBinaryCandidates(ctx, spec)
 	if err != nil {
 		return "", err
--- a/backend/internal/adapters/agent/goose/goose.go
+++ b/backend/internal/adapters/agent/goose/goose.go
@@ -304,7 +304,7 @@ func isOfficialGooseBinary(ctx context.Context, binary string) bool {
 	probeCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
 	defer cancel()
 
-	out, err := aoprocess.CommandContext(probeCtx, binary, "--version").CombinedOutput()
+	out, err := aoprocess.CommandContextForPath(probeCtx, binary, "--version").CombinedOutput()
 	if err != nil || probeCtx.Err() != nil {
 		return false
 	}
--- a/backend/internal/adapters/agent/goose/goose.go
+++ b/backend/internal/adapters/agent/goose/goose.go
@@ -27,13 +27,17 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/agentbase"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/binaryutil"
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+	aoprocess "github.com/aoagents/agent-orchestrator/backend/internal/process"
 )
 
 const (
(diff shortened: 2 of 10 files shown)


backend/internal/adapters/agent/goose/goose.go
```diff
@@ -271,10 +275,66 @@ var gooseBinarySpec = binaryutil.BinarySpec{
...
+	out, err := aoprocess.CommandContext(probeCtx, binary, "--version").CombinedOutput()
+	if err != nil || probeCtx.Err() != nil {
+		return false
+	}
+	version := strings.TrimSpace(string(out))
+	return strings.HasPrefix(version, "goose ") && !strings.HasPrefix(version, "goose version:")
+}
+
+func gooseCanonicalBinaryCandidates(spec binaryutil.BinarySpec) []string {
+	if runtime.GOOS == "windows" {
```
codebanditssss: Blocking: after `ResolveBinary` returns a conflicting `PATH` hit, this returns no recovery candidates on Windows, despite `gooseBinarySpec` declaring `%APPDATA%`, `%LOCALAPPDATA%`, and Cargo locations. The Unix branch similarly omits package-manager candidates already supported by `binaryutil` (including asdf, mise, Snap, Linuxbrew, Volta, nvm, and fnm). A valid Block Goose install can therefore still be rejected. Please expose complete ordered candidate enumeration from `binaryutil` and validate its deduplicated results here.

backend/internal/adapters/agent/goose/goose.go
```diff
@@ -271,10 +275,66 @@ var gooseBinarySpec = binaryutil.BinarySpec{
...
+		return "", err
+	}
+	return "", fmt.Errorf("%s: %w", spec.Label, ports.ErrAgentBinaryNotFound)
+}
+
+func isOfficialGooseBinary(ctx context.Context, binary string) bool {
+	probeCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+
+	out, err := aoprocess.CommandContext(probeCtx, binary, "--version").CombinedOutput()
```
codebanditssss: `ResolveGooseBinary` is reached by `Service.FindInstalledBinary` because Goose does not implement `AgentBinaryPresenceResolver`. That startup path explicitly promises not to start an agent CLI, but this now executes the first `goose` inherited from `PATH`. Please keep presence detection path-only (for example, implement `ResolveBinaryPresence` for Goose) and reserve the identity probe for paths where process execution is expected.

backend/internal/adapters/agent/goose/goose_test.go
```diff
@@ -547,6 +549,49 @@ func TestResolveGooseBinaryFallback(t *testing.T) {
 	}
 }
 
+func TestResolveGooseBinaryRejectsPresslyGoose(t *testing.T) {
```
codebanditssss, line 553: These additions cover only two successful command executions with different output strings. Please add the actual collision regression (foreign `goose` on `PATH`, valid Block Goose in a fallback location), plus malformed output, non-zero exit, parent cancellation, probe timeout, and native Windows coverage. Without those cases, the bounded/cancellable fallback behavior described by this PR is not protected.

codebanditssss (changes requested):
## Critical Risk

Collision recovery does not enumerate every location already supported by `BinarySpec`. When a foreign `goose` wins on `PATH`, Windows retries no fallback candidates and Unix omits package-manager candidates, so a valid Block Goose installation can still be reported missing.

## Other Gaps

- Identity validation executes a process during the presence-only desktop startup path → unexpected execution and startup delay.
- Waiting for `binaryMu` is not context-aware → cancelled callers can remain blocked and later receive a cached success.
- Tests omit collision recovery, timeout, cancellation, malformed output, non-zero exit, and Windows behavior → the new failure paths are unprotected.

## Fix

1. Centralize ordered candidate enumeration in `binaryutil`, covering all existing platform and package-manager locations.
2. Validate every deduplicated candidate from that shared enumeration and remove the partial Goose-local fallback list.
3. Add Unix/package-manager and Windows collision regression tests.

## Verdict

Needs rework.

backend/internal/adapters/agent/binaryutil/binaryutil.go
```diff
@@ -114,24 +127,48 @@ func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
...
+	return candidates, nil
+}
+
+// ResolveBinary returns the path to spec's binary, searching PATH then the
+// platform's candidate install locations. It returns a wrapped
+// ports.ErrAgentBinaryNotFound when nothing matches, so callers surface a clear
+// "command not found" rather than launching an empty argv. ctx cancellation is
+// honored between probes.
+func ResolveBinary(ctx context.Context, spec BinarySpec) (string, error) {
+	candidates, err := ResolveBinaryCandidates(ctx, spec)
```
codebanditssss, line 184: Blocking: `ResolveBinary` now waits for `ResolveBinaryCandidates` to finish before returning an executable already found on `PATH`. For every `NodeManaged` spec, candidate enumeration continues into Volta/nvm/fnm discovery; if the context expires there, the error is returned and the valid PATH candidate is discarded. The desktop prerequisite scan uses a 500 ms deadline, so this can falsely report an installed harness as missing. Please preserve the old `exec.LookPath` fast path in `ResolveBinary`, keep exhaustive enumeration for Goose identity validation, and add a regression test where fallback discovery fails after a valid PATH hit.

codebanditssss (changes requested):
## Critical Risk

`ResolveBinary` now waits for exhaustive fallback and Node-manager discovery even after `exec.LookPath` finds a valid executable. If the caller context expires during that later discovery, the earlier PATH candidate is discarded. The desktop prerequisite scan has a 500 ms deadline, so an installed harness can be falsely reported missing.

## Other Gaps

- No regression test preserves the original PATH fast path when fallback enumeration fails or exceeds its deadline → this startup regression is unprotected.
- Windows resolver tests are not executed by required CI → platform behavior is not continuously verified.

## Fix

1. Restore `ResolveBinary`’s immediate return for a valid `exec.LookPath` result.
2. Keep `ResolveBinaryCandidates` exhaustive for Goose identity-aware collision recovery.
3. Add a test proving fallback-discovery failure cannot discard a valid PATH result.

## Verdict

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/agent/binaryutil/binaryutil.go
adot-7: Issue addressed

codebanditssss:
Blocking on Windows: `gooseBinarySpec` includes `goose.cmd` candidates, but `isOfficialGooseBinary` passes each candidate directly to `aoprocess.CommandContext(candidate, "--version")`, which is plain `exec.CommandContext`. Windows requires `.cmd` and `.bat` files to be run through the command interpreter (`cmd.exe /c`), so a valid supported Goose `.cmd` shim will fail this probe and be reported as not installed.

The added Windows regression test should expose this, but the Go workflow runs only on Ubuntu, while the Windows CLI E2E job does not run the Goose package tests; locally the test is skipped on Linux.

Please handle `.cmd`/`.bat` candidates through a safely quoted Windows wrapper or resolve the underlying executable, and run this regression test in a native Windows Go job.

Reference: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

codebanditssss (changes requested):
## Critical Risk

The startup path still reproduces part of #4516. `Plugin.ResolveBinaryPresence` calls the name-only `binaryutil.ResolveBinary`, and `Service.FindInstalledBinary` treats that result as an installed harness. A machine with only pressly/goose can therefore still pass the first-render prerequisite as having Block Goose installed, before the identity-aware refresh corrects the inventory.

## Other Gaps

- The outer installation deadline is 2 seconds while each identity probe allows 3 seconds → a hanging PATH collision prevents reaching a valid fallback.
- Identity validation accepts any successful output beginning with `goose ` → weaker collision protection than a semantic Block Goose signature.
- All inline threads and CI checks are green, but this startup path lacks a pressly-only service regression.

## Fix

1. Make startup presence for identity-sensitive adapters return an ambiguous/pending result instead of treating a name match as installed.
2. Mark Goose installed only after a cached or fresh successful identity probe.
3. Add a service-level regression proving pressly/goose alone cannot satisfy the startup harness check.

## Verdict

nikhilachale:
Additive to the existing reviews (not duplicating the presence-path, deadline-mismatch, or Windows gaps already flagged):

**1. Positive signature over a negative prefix check.** `isOfficialGooseBinary` accepts any output starting with `goose ` except `goose version:` — any unrelated `goose <x>` CLI passes. A positive marker is more robust: Block Goose's `--help` advertises `session` and `recipe` subcommands; pressly/goose advertises `up/down/status/create` and neither marker. Verified against a real Block Goose v1.48.0 install and `brew info goose` (pressly formula conflicts with `block-goose-cli` — same name, different tool). `--help` subcommand names are also more stable across versions than version-string format.

**2. Reusable shape, per the issue's fix section.** #4516 recommended `ValidateIdentity func(ctx, path) bool` on `BinarySpec` "so the whole resolver fleet can adopt per-harness identity checks; goose opts in first." This PR makes identity goose-local (`isOfficialGooseBinary` + the new `ResolveBinaryCandidates` export), which fixes goose but leaves #3409 (codex `.cmd`/`.exe`, "same resolver class, different failure mode" per the issue) to be solved separately. A `ValidateIdentity` hook would cover both and keep the goose-specific probe in the goose package.

adot-7:
I think we should split this into two changes instead of trying to fix both #4516 and #3409 in one PR. The code changes will be too large to properly review. 

For #4516, I would keep the fix limited to Goose identity detection (keeping in mind both points mentioned by @nikhilachale [here](https://github.com/Untrivial-ai/agent-orchestrator/pull/4519#issuecomment-5593535512))

I will cover the windows resolver part as flagged by @codebanditssss [here](https://github.com/Untrivial-ai/agent-orchestrator/pull/4519#issuecomment-5481839428) separately because it belongs in the windows resolver issue #3409.

This PR has gotten way out of scope, so I am gonna create the PR for the first issue and supersede this. And then create another PR for the windows issue.

codebanditssss (changes requested):
### critical risk

this pr must not merge in its current form.

- `backend/internal/adapters/agent/goose/goose.go:265-274` includes `goose.cmd`, but the worker launch path (`goose.go:107-124`) and reviewer path (`backend/internal/adapters/reviewer/goose/goose.go:54-58`) pass that path directly to windows process creation. a batch shim requires `cmd.exe`; detection can therefore succeed while an actual worker or reviewer launch fails. see the windows process requirements: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa
- `backend/internal/adapters/agent/goose/install.go:14-19` still uses name-only presence detection, so a pressly `goose` migration binary can satisfy the startup installation gate.
- each identity probe allows three seconds while the inventory operation has a two-second overall budget (`backend/internal/service/agent/service.go:19-25,712-727`), so a hanging PATH collision can prevent fallback recovery.

### other gaps

- the version-prefix check accepts arbitrary successful output beginning with `goose `.
- the new shell-wrapper tests do not cover `%`, embedded quotes, newlines, or child-process cleanup.
- the branch is superseded by #5267 and should not remain a competing implementation.

### fix

do not merge this branch. use #5267 for the identity implementation and close this pr after confirming the replacement is accepted.

### verdict

changes requested. this branch is obsolete and still contains launch and startup false-positive bugs.

codebanditssss:
Thanks for the contribution. Goose binary identity validation is already present on current `main` following #5267, including the current shared binary-resolution architecture. This older conflicted implementation is therefore no longer necessary.

Issue #4031, opened 2026-08-15
bug(kimi): OAuth-managed auth is reported unknown or unauthorized


neversettle17-101:
## Bug
AO reports Kimi worker auth as unknown/unauthorized even when the user's Kimi CLI is authenticated through Kimi's normal OAuth-managed configuration.

**Source:** user report in AO worker session 2026-08-15
**Analyzed against:** `0f43574f20aab7d8f7a7c61142f94f9c87a214eb`
**Confidence:** Medium-high

## Observed locally
A fresh AO agent probe reports Kimi as installed but not authorized:

```text
ao agent ls --refresh --json
...
{"id":"kimi","label":"Kimi","authStatus":"unauthorized"}
```

The Kimi binary resolves correctly:

```text
/Users/chauhan/.kimi-code/bin/kimi
```

The user's Kimi config exists in both homes:

- `~/.kimi-code/config.toml`
- `~/.ao/data/kimi/config.toml`

Both configs contain managed OAuth configuration (`oauth.storage` / `oauth.key`) but the `api_key` fields are empty. The user's `~/.kimi-code/credentials/kimi-code.json` exists, but `access_token` and `refresh_token` are empty. The AO-managed Kimi home does not have `~/.ao/data/kimi/credentials/kimi-code.json`.

Kimi also does not support AO's generic auth probe commands:

```text
kimi auth status      -> unknown command
kimi login status     -> unknown command
kimi providers list   -> unknown command
```

## Reproduction
1. Authenticate Kimi via its normal OAuth/login flow so the CLI can operate.
2. Ensure Kimi's config uses managed OAuth storage with empty `api_key` values and no non-empty `access_token`/`refresh_token` in `credentials/kimi-code.json`.
3. Start AO and refresh the agent catalog.
4. Inspect Kimi in `ao agent ls --refresh` or the worker/reviewer agent picker.

## Expected
AO should not report Kimi auth as unknown/unauthorized when Kimi's own OAuth-backed configuration can successfully authenticate the CLI.

## Actual
AO reports Kimi as auth unknown/unauthorized because the local detector only recognizes env/API key fields, non-empty JSON token fields, or generic CLI auth/status output.

## Root Cause
Kimi auth detection currently checks only these sources:

- env vars: `KIMI_API_KEY`, `KIMI_CODE_API_KEY`, `MOONSHOT_API_KEY`
- non-empty `api_key = ...` values in `config.toml`
[...]

PR #2960, opened 2026-07-22
fix(kimi): seed OAuth-backed config into AO-managed home

serhatsayat:
## What

Seed the user's **full native Kimi config** into AO's isolated Kimi home when the profile is authenticated through Kimi's device-code OAuth flow, not only when it carries a literal API key.

The follow-up now supports both Kimi storage generations:

- current `KIMI_SHARE_DIR` / `~/.kimi`
- legacy `KIMI_CODE_HOME` / `~/.kimi-code`

AO sets both environment variables to the same directory under `AO_DATA_DIR`, so current and legacy Kimi binaries use the same isolated, hook-enabled profile.

> **Rebased 2026-08-20.** This branch originally also added `~/.kimi-code/bin/kimi` binary discovery and OAuth credential auth-detection. Both landed on `main` independently (#3648 and follow-ups), so the rebase kept only the config-seeding gap they left open. The 2026-08-23 follow-up adds the current official Kimi share-directory layout requested in review and was verified against the current `main` merge result.

## Why

AO isolates Kimi state under its own data directory. `seedKimiCredentials` already copied OAuth credentials there, but `kimiSeedConfig` refused to seed `config.toml` unless `kimiConfigHasAPIKey(source)` was true.

A device-code login **intentionally leaves every `api_key` empty** — the tokens live in `credentials/kimi-code.json`. The managed home therefore received a valid credential and only AO's hook entries: no `default_model`, provider/OAuth mapping, model tables, services, or permissions.

Kimi then starts with `Model: not set`, rejects AO's injected task with `Error: LLM not set`, never creates a native session, and surfaces as `no_signal` after the grace period even though the advisory auth probe reports the agent `authorized`.

Current Kimi releases also store runtime data under `KIMI_SHARE_DIR` (default `~/.kimi`) rather than the legacy `KIMI_CODE_HOME` location. Supporting only the legacy layout would leave modern installs unable to seed or consume AO's isolated config.

Full RCA with an A/B reproduction against `63b5ffb`: https://github.com/Untrivial-ai/agent-orchestrator/pull/2960#issuecomment-5354802534

## How

- Discover explicitly configured `KIMI_SHARE_DIR` and `KIMI_CODE_HOME` locations; when neither is set, check `~/.kimi` first and then `~/.kimi-code`.
- Probe config and OAuth credentials across those current and legacy locations.
- Treat a source profile as seedable when it is authorized by either a non-empty `api_key` or a usable `credentials/kimi-code.json`.
[...]

PR #4521, opened 2026-08-26
fix(kimi): seed OAuth config into managed home

### LINKED ISSUE #4520: bug(kimi): OAuth worker starts without model in AO-managed home
## Bug

AO-managed Kimi TUI workers can start without a model and reject the injected task with:

```text
Error: LLM not set, send /login to login
```

The same Kimi binary works when launched manually in the same worktree. This affects the normal Kimi device-code OAuth profile shape: `config.toml` has a `default_model`, provider/OAuth mapping, and model definitions, while the provider `api_key` is intentionally empty and tokens are stored in the referenced credential JSON.

**Source:** live same-machine reproduction in AO worker `reverb-45` on 2026-08-26  
**Analyzed against:** `c4abb3d0e` (`origin/main`)  
**AO:** packaged Go daemon `0.12.7-nightly.202608250031` on `127.0.0.1:3001` (CLI version currently reports `dev`)  
**Kimi:** `0.38.0`, macOS arm64  
**Confidence:** High — reproduced and traced to the exact config-seeding predicate  
**Priority:** P1 — the normal OAuth-backed Kimi worker path is unusable until the managed profile is manually repaired

## Reproduction

1. Authenticate native Kimi with its device-code OAuth flow.
2. Confirm the native Kimi profile contains a non-empty `default_model`, a provider with an OAuth credential reference, model definitions, and usable access/refresh credentials, but no non-empty literal `api_key`.
3. Spawn an AO Kimi worker with a non-empty task.
4. Observe that Kimi starts with no LLM, rejects the task with `Error: LLM not set, send /login to login`, and the AO session later derives `no_signal`.
5. Run plain `kimi` manually from the same worktree; it loads the native model/profile successfully.

No credential values need to be logged or exposed to reproduce this.

## Root Cause

AO intentionally isolates Kimi under its own data directory:

1. `backend/internal/adapters/agent/kimi/kimi.go` sets `KIMI_CODE_HOME=<AO_DATA_DIR>/kimi` for managed launches.
2. `backend/internal/adapters/agent/kimi/hooks.go:installKimiConfigHooks` prepares `<AO_DATA_DIR>/kimi/config.toml` and merges AO lifecycle hooks.
3. `seedKimiCredentials` can securely copy a usable native OAuth credential into the isolated home on first use.
4. `kimiSeedConfig`, however, copies the matching native `config.toml` only when `kimiConfigHasAPIKey(source)` finds a non-empty literal `api_key`.
5. Device-code OAuth profiles intentionally keep authentication in `credentials/kimi-code.json`; their config still carries the required `default_model`, provider/OAuth mapping, aliases, services, permissions, and model tables. The API-key-only predicate rejects that config.

The result is an inconsistent managed profile: AO has a credential and its four lifecycle hooks, but no default model or provider/model graph. Kimi therefore reports an unset LLM before it can create a native session or emit AO lifecycle hooks. `no_signal` is a downstream symptom, not a dead-process or readiness-classification bug.

Sanitized local evidence:

- native `~/.kimi-code/config.toml`: keys include `default_model`, `providers`, `models`, `services`, `permission`; the managed provider has an OAuth reference and no non-empty API key
- AO `<AO_DATA_DIR>/kimi/config.toml`: contains only the four AO-managed hook entries
- both credential files exist with mode `0600`; values were not printed or compared in this report
- the AO daemon is healthy on port 3001 and the Kimi process remains live

## Correct Fix and Security/State Behavior

When the AO target config is empty or contains only AO's managed hook block, seed the **complete native config** if either:

- it has a non-empty literal API key (existing behavior), or
- its declared OAuth credential reference resolves to a locally usable access/refresh credential.

Then merge exactly one AO-managed hook block as before.

Copying the whole config is deliberate: selectively copying `default_model` would risk separating it from its provider, OAuth storage reference, model aliases, services, permissions, and related settings. This matches the existing first-use behavior for API-key profiles.

Security/state invariants:

- Keep `KIMI_CODE_HOME` under `AO_DATA_DIR`; do **not** point managed sessions at or write hooks into the native profile.
- Copy secrets only into the AO-managed home with owner-only permissions; never log or commit credential contents.
- Preserve native config/credentials byte-for-byte.
- Preserve any AO config containing non-hook user state; only merge/update AO's owned hook block.
- Preserve an existing AO-managed credential rather than overwriting an independently refreshed or explicitly logged-in managed profile.
- Do not seed OAuth-shaped config when its referenced credential is missing or unusable.

## Remaining Limitations

- Native and AO-managed credentials intentionally diverge after first seeding. If the managed credential later becomes unusable, the user must log in within the managed Kimi profile (or deliberately remove it to allow first-use reseeding); AO should not silently overwrite an existing managed secret from the native profile.
- OAuth configurations whose real token lives only in an external store and cannot be positively verified from the referenced local credential remain outside this fix; that advisory-auth variant is tracked separately in #4031.

## Duplicate Search / Related

No exact issue was found after searching all open and closed issues for the error text, `Model: not set`, `KIMI_CODE_HOME`, `default_model`, OAuth config seeding, and managed Kimi login failures.

- #4031 is related but not a duplicate: it covers OAuth configurations with empty local token fields/external storage where AO reports auth unknown or unauthorized. Here the local credential is usable and AO can report `authorized`; launch fails because the matching model/provider config is not seeded.
- PR #2960 identified the same config-seeding gap, but is currently conflicting with `main`, has no closing issue, and includes storage-layout follow-up changes beyond this exact reproduction. A current-main surgical PR should supersede only the missing OAuth-config predicate/test coverage.
- PR #3648 added first-use OAuth credential copying while intentionally preserving existing managed credentials; it did not retain OAuth profile-config seeding.

## Impact

- OAuth-authenticated Kimi installs can appear ready but fail every AO-managed task before creating a session.
- Workers can remain live while surfacing `no_signal`, obscuring the actual missing-model cause.
- Manual `kimi` remains functional, which makes the failure look terminal- or worktree-specific when it is actually managed-profile inconsistency.

whoisasx:
## What

Seed the user's complete native Kimi config and its exact referenced OAuth credential into AO's isolated Kimi home when the source profile uses device-code OAuth, not only when it contains a non-empty literal `api_key`.

The change keeps the existing first-use boundary: only an empty or AO-hook-only managed config is seedable, after which AO merges exactly one lifecycle-hook block.

Fixes #4520.

This is a surgical current-`main` replacement for the remaining config-seeding delta in #2960, which is currently conflicting and also carries broader storage-layout work. Original contributor @serhatsayat is credited in the commit.

## Why

AO worker `reverb-45` reproduced the failure on current `main`: Kimi `0.38.0` launched by AO displayed `Error: LLM not set, send /login to login`, while the same binary launched manually in the same worktree loaded its model and worked.

### RCA

AO deliberately launches Kimi with `KIMI_CODE_HOME=<AO_DATA_DIR>/kimi` so lifecycle hooks and Kimi state remain isolated under `~/.ao`. Hook preparation already copied a usable native OAuth credential into that home, but `kimiSeedConfig` accepted the matching native `config.toml` only when `kimiConfigHasAPIKey(source)` found a non-empty literal API key.

Kimi's device-code OAuth profile intentionally has empty API-key fields. Its config still carries the required `default_model`, provider/OAuth mapping, model aliases, services, permissions, and model definitions. The API-key-only gate therefore produced an inconsistent managed profile:

- credential: present
- AO lifecycle hooks: present
- default model/provider/model graph: missing

Kimi rejected AO's injected task before creating a native session or emitting lifecycle hooks, so `no_signal` was a downstream symptom rather than a process crash or readiness bug.

The regression is the untested intersection of two existing paths: API-key profile-config seeding and OAuth credential seeding.

## How

- If the native config has a non-empty API key, preserve the existing path.
- Otherwise, parse each provider OAuth reference, resolve it through Kimi's existing `kimiOAuthCredentialPath` mapping, and positively verify the exact referenced access/refresh credential before seeding the native config.
- Copy usable referenced credentials to their matching AO-managed paths; retain `credentials/kimi-code.json` only as a legacy fallback when no OAuth reference was persisted. An unrelated default credential is not copied when the config references a scoped key.
- Copy the complete config rather than cherry-picking `default_model`; the provider/OAuth reference, aliases, services, permissions, and model definitions form one internally consistent profile.
- Merge AO's hook block exactly once.

### Intentional security/state behavior

- `KIMI_CODE_HOME` remains under `AO_DATA_DIR`; managed sessions never point at or write hooks into the native `~/.kimi-code` profile.
- Credential contents are neither logged nor committed. First-use copies continue to be written owner-only (`0600`) inside AO-managed state.
- Native config and credentials remain byte-for-byte unchanged.
- Any AO config containing non-hook state is preserved; only AO's owned hook block is merged/updated.
- An existing AO-managed credential is never overwritten. This preserves an independent managed login or a token that Kimi refreshed inside the isolated profile.
- A token-less/missing native credential does not make an empty-key profile seedable.

### Intentional limitations

- Native and AO-managed credentials can diverge after first seeding. If an existing managed credential later becomes unusable, the user must log in within the managed Kimi profile or deliberately remove it to allow first-use reseeding; AO does not silently overwrite an existing managed secret.
- OAuth profiles whose usable token exists only in an external credential store and cannot be positively verified from local credential JSON remain outside this change; the advisory-auth variant is tracked by #4031.
- Existing AO configs with non-hook content are not repaired or replaced automatically, even if they are incomplete.

## Testing

The original OAuth regression tests and the scoped-reference review tests were run against their respective old gates and failed for the expected missing-config/missing-credential reasons; they pass with the fixes.

- `cd backend && go test -race ./internal/adapters/agent/kimi -count=1`
- `cd backend && go vet ./internal/adapters/agent/kimi`
- `cd backend && go test ./internal/adapters/agent/registry ./internal/session_manager -count=1`
- `env -i HOME="$HOME" PATH="$PATH" TMPDIR="${TMPDIR:-/tmp}" npm run lint`
  - full `go test ./...`
- `git diff --check`

Coverage verifies:

- OAuth config + first-use credential seeding into an empty managed home
- recovery from the exact hook-only managed config state in the reproduction
- complete default-model/provider/OAuth/model graph preservation
- exactly one AO hook block
- native config and credential immutability
- preservation of an existing, divergent AO-managed credential
- scoped OAuth keys such as `oauth/kimi-code-env-<hash>` with and without an unrelated default credential
- copying only credentials referenced by the seeded config

## Checklist

- [x] Branched from `main`
- [x] Focused regression tests added with a verified red-green cycle
- [x] Relevant backend tests, build, vet, and lint pass

### DIFF
--- a/backend/internal/adapters/agent/kimi/hooks.go
+++ b/backend/internal/adapters/agent/kimi/hooks.go
@@ -196,9 +196,27 @@ func kimiSeedConfig(targetPath string, existing []byte) ([]byte, bool, error) {
 		return nil, false, fmt.Errorf("read source Kimi config %s: %w", sourcePath, err)
 	}
 	if !kimiConfigHasAPIKey(source) {
-		return nil, false, nil
+		// Device-code logins leave api_key empty and keep their tokens in the
+		// credential file. The full config is still required for default_model,
+		// the provider/OAuth mapping, model aliases, services, and permissions.
+		authorized, err := kimiSourceOAuthAuthorized(sourceHome)
+		if err != nil {
+			return nil, false, err
+		}
+		if !authorized {
+			return nil, false, nil
+		}
 	}
 	return source, true, nil
+}
+
+func kimiSourceOAuthAuthorized(sourceHome string) (bool, error) {
+	path := filepath.Join(sourceHome, "credentials", "kimi-code.json")
+	status, ok, err := kimiCredentialsAuthStatus(path)
+	if err != nil {
+		return false, fmt.Errorf("read source Kimi credentials %s: %w", path, err)
+	}
+	return ok && status == ports.AgentAuthStatusAuthorized, nil
 }
 
 func kimiConfigCanSeed(existing []byte) bool {
--- a/backend/internal/adapters/agent/kimi/hooks.go
+++ b/backend/internal/adapters/agent/kimi/hooks.go
@@ -134,8 +134,26 @@ func seedKimiCredentials(targetHome string) error {
 	if !ok {
 		return nil
 	}
-	sourcePath := filepath.Join(sourceHome, "credentials", "kimi-code.json")
-	targetPath := filepath.Join(targetHome, "credentials", "kimi-code.json")
+	sourceConfigPath := filepath.Join(sourceHome, "config.toml")
+	sourcePaths, err := kimiConfigOAuthCredentialPaths(sourceConfigPath)
+	if err != nil {
+		return fmt.Errorf("read source Kimi config %s: %w", sourceConfigPath, err)
+	}
+	if len(sourcePaths) == 0 {
+		// Preserve the legacy credential-only seed path for profiles created by
+		// Kimi versions that did not persist an OAuth reference in config.toml.
+		sourcePaths = []string{filepath.Join(sourceHome, "credentials", "kimi-code.json")}
+	}
+	for _, sourcePath := range sourcePaths {
+		targetPath := filepath.Join(targetHome, "credentials", filepath.Base(sourcePath))
+		if err := seedKimiCredential(sourcePath, targetPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func seedKimiCredential(sourcePath, targetPath string) error {
 	if sameKimiConfigPath(sourcePath, targetPath) {
 		return nil
 	}
@@ -211,12 +229,21 @@ func kimiSeedConfig(targetPath string, existing []byte) ([]byte, bool, error) {
 }
 
 func kimiSourceOAuthAuthorized(sourceHome string) (bool, error) {
-	path := filepath.Join(sourceHome, "credentials", "kimi-code.json")
-	status, ok, err := kimiCredentialsAuthStatus(path)
+	configPath := filepath.Join(sourceHome, "config.toml")
+	paths, err := kimiConfigOAuthCredentialPaths(configPath)
 	if err != nil {
-		return false, fmt.Errorf("read source Kimi credentials %s: %w", path, err)
+		return false, fmt.Errorf("read source Kimi config %s: %w", configPath, err)
 	}
-	return ok && status == ports.AgentAuthStatusAuthorized, nil
+	for _, path := range paths {
+		status, ok, err := kimiCredentialsAuthStatus(path)
+		if err != nil {
+			return false, fmt.Errorf("read source Kimi credentials %s: %w", path, err)
+		}
+		if ok && status == ports.AgentAuthStatusAuthorized {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 func kimiConfigCanSeed(existing []byte) bool {
--- a/backend/internal/adapters/agent/kimi/auth.go
+++ b/backend/internal/adapters/agent/kimi/auth.go
@@ -127,6 +127,47 @@ type kimiAuthConfig struct {
 	Providers map[string]kimiCredentialSource `json:"providers" toml:"providers"`
 }
 
+func kimiConfigOAuthCredentialPaths(path string) ([]string, error) {
+	data, err := os.ReadFile(path)
+	if os.IsNotExist(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	if strings.TrimSpace(string(data)) == "" {
+		return nil, nil
+	}
+	var config kimiAuthConfig
+	if strings.EqualFold(filepath.Ext(path), ".json") {
+		err = json.Unmarshal(data, &config)
+	} else {
+		err = toml.Unmarshal(data, &config)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	paths := make([]string, 0, len(config.Providers))
+	seen := make(map[string]struct{}, len(config.Providers))
+	for _, provider := range config.Providers {
+		if provider.OAuth == nil {
+			continue
+		}
+		credentialPath := kimiOAuthCredentialPath(filepath.Dir(path), provider.OAuth.Key)
+		if credentialPath == "" {
+			continue
+		}
+		clean := filepath.Clean(credentialPath)
+		if _, ok := seen[clean]; ok {
+			continue
+		}
+		seen[clean] = struct{}{}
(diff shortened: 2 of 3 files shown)


whoisasx:
# Before
Screenshot 2026-08-27 at 06 29 33

# After
Screenshot 2026-08-27 at 06 29 09

whoisasx:
@neversettle17-101 , it's fixes the issue which we were having. @nikhilachale can you review it.

backend/internal/adapters/agent/kimi/hooks.go
```diff
@@ -196,11 +196,29 @@ func kimiSeedConfig(targetPath string, existing []byte) ([]byte, bool, error) {
...
+		}
+		if !authorized {
+			return nil, false, nil
+		}
 	}
 	return source, true, nil
 }
 
+func kimiSourceOAuthAuthorized(sourceHome string) (bool, error) {
+	path := filepath.Join(sourceHome, "credentials", "kimi-code.json")
```
nikhilachale: **P2 — Resolve and copy the OAuth credential referenced by the config.** This hardcodes `credentials/kimi-code.json`, but Kimi Code 0.38.0 can persist scoped keys such as `oauth/kimi-code-env-<hash>` in `credentials/kimi-code-env-<hash>.json` when OAuth or base-URL overrides are used. A valid scoped profile therefore fails this gate and remains hook-only; if both files exist, `seedKimiCredentials` can copy the default token even though the seeded config references the scoped one. I reproduced this with a focused test: `GetAgentHooks` wrote only the AO hook block and omitted both the default model and scoped credential. Please parse the source config OAuth reference—the existing `kimiOAuthCredentialPath` already maps it—and validate and copy that exact credential into the managed home.

whoisasx: Fixed in `25eb1f73e`. Hook setup now parses every provider OAuth key from the native config, resolves it through `kimiOAuthCredentialPath`, validates that exact credential, and copies it to the matching AO-managed path without overwriting an existing managed credential. The default `kimi-code.json` path is retained only as a legacy fallback when the config has no persisted OAuth reference. Regression coverage includes scoped-only credentials and the case where both scoped and unrelated default credentials exist; only the scoped credential is copied.

nikhilachale (approved):

PR #4523, opened 2026-08-26
feat: add harness login actions in settings

nikhilachale:
## Summary

- add daemon-owned authentication plans for all 27 Harness settings entries
- launch vendor-owned login/setup flows through either an inline AO terminal or, where no reliable setup command exists, official documentation
- add explicit **Open login** and **Open setup** terminal actions for interactive TUI adapters; AO never guesses readiness or submits `/login` or `/auth` automatically
- route Aider setup to its official API-key guide instead of launching a normal `aider` coding session
- isolate every interactive auth flow in its own empty `~/.ao/auth-workspace/<terminal-id>` directory and remove that directory only after the terminal is confirmed dead
- show login-specific and configuration-specific actions/statuses in Harness settings, then reconcile advisory authentication through #4589's daemon-owned readiness coordinator
- recognize Aider's OpenRouter OAuth credentials in `~/.aider/oauth-keys.env`

## Stack

This PR was originally stacked on #4221, which is now merged into upstream `main`.

- Upstream base: `main`
- Head: `nikhilachale:feat/harness-install-settings-main`

## Login and setup flow

```mermaid
flowchart TD
  A["User opens Settings → Harness"] --> B{"Harness installed?"}
  B -- No --> C["Show Install action"]
  B -- Yes --> D["Load daemon-owned auth plan"]
  D --> E["User clicks Login or Set up"]
  E --> LM{"Launch mode"}
  LM -- Documentation --> DOC["Open official vendor setup guide"]
  DOC --> T["User returns and clicks Check configuration"]
  LM -- Terminal --> F["POST /api/v1/agents/:agent/auth with agent ID only"]
  F --> G["Daemon resolves fixed reviewed command through harness adapter"]
  G --> H["Create private auth workspace and inline AO terminal"]
  H --> I{"Authentication entry point"}
  I -- Native login or setup command --> J["CLI opens its vendor-owned login or provider flow"]
  I -- Pi / Droid / Prime Agent / OMP --> K["Start native TUI"]
  K --> L["AO enables Open login after terminal attachment"]
  L --> M["User waits for TUI, then clicks Open login"]
  M --> N["AO offers fixed /login input to the active attachment"]
  I -- Qwen --> QA["Start qwen TUI"]
  QA --> QB["AO enables Open setup after terminal attachment"]
  QB --> QC["User clicks Open setup"]
  QC --> QD["AO offers fixed /auth input to the active attachment"]
  N --> O{"Input accepted?"}
  QD --> O
  O -- No --> RETRY["Keep the current action enabled for retry"]
  O -- Yes --> J
  J --> P["User completes vendor-owned login or provider configuration"]
  P --> Q{"Terminal attachment state"}
  Q -- Exited --> R["Start targeted authentication verification"]
  Q -- Error --> S["Keep workflow open for reattach or close"]
  B -- Yes --> T
  T --> R
  R --> U["Invalidate this agent's cached auth observation"]
  U --> V["#4589 readiness coordinator performs a fresh targeted ensure"]
  V --> W{"Normalized authentication state"}
  W -- Authorized --> X["Show Logged in or Configured"]
  W -- Unauthorized --> Y["Show Not logged in or Not configured"]
  W -- Unknown --> Z["Show unverified and keep retry/check available"]
```

### Complete agent authentication matrix

| Adapter | AO action | Launch mode | Terminal command / destination | Follow-up |
| --- | --- | --- | --- | --- |
| [Claude Code](https://code.claude.com/docs/en/installation) | **Login** | Terminal | `claude auth login` | Complete browser/device flow |
| [Codex](https://github.com/openai/codex) | **Login** | Terminal | `codex login` | Complete browser/device-code flow |
| [Cursor](https://docs.cursor.com/en/cli/installation) | **Login** | Terminal | `cursor-agent login` | Complete browser flow |
| [OpenCode](https://github.com/anomalyco/opencode) | **Login** | Terminal | `opencode auth login` | Choose provider in the native CLI |
| [Aider](https://aider.chat/docs/config/api-keys.html) | **Set up** | Documentation | Official Aider API-key guide | Configure an environment variable, `.env`, or `.aider.conf.yml`, then click **Check configuration** |
| [GitHub Copilot](https://docs.github.com/en/copilot/how-tos/copilot-cli/set-up-copilot-cli/install-copilot-cli) | **Login** | Terminal | `copilot login` | Complete GitHub device/browser flow |
| [Grok](https://docs.x.ai/build/overview) | **Login** | Terminal | `grok login` | Complete native login; device auth remains available |
| [Kimi](https://moonshotai.github.io/kimi-code/en/) | **Login** | Terminal | `kimi login` | Complete browser flow |
| [Pi](https://github.com/earendil-works/pi) | **Login** | Terminal | `pi` | Click **Open login** to send `/login` + Enter |
| [Amp](https://ampcode.com/manual) | **Login** | Terminal | `amp login` | Complete browser flow |
| [Auggie](https://docs.augmentcode.com/cli/overview) | **Login** | Terminal | `auggie login` | Complete browser flow |
| [Droid](https://docs.factory.ai/droid-cli/cli-reference) | **Login** | Terminal | `droid` | Click **Open login** to send `/login` + Enter |
| [Crush](https://github.com/charmbracelet/crush) | **Login** | Terminal | `crush login` | Complete Charm Hyper login; GitHub Copilot remains available |
| [Cline](https://github.com/cline/cline) | **Login** | Terminal | `cline auth` | Complete native authentication flow |
| [Goose](https://block.github.io/goose/index.html) | **Set up** | Terminal | `goose configure` | Configure the provider in Goose's native prompt |
| [Qwen](https://qwenlm.github.io/qwen-code-docs/en/users/configuration/auth/) | **Set up** | Terminal | `qwen` | Click **Open setup** to send `i` → Backspace → `/auth` + Enter |
| [Continue](https://docs.continue.dev/cli/quickstart) | **Login** | Terminal | `cn login` | Complete browser flow |
| [Devin](https://docs.devin.ai/get-started/devin-intro) | **Login** | Terminal | `devin auth login` | Complete browser flow; manual-token flow remains available |
| [Kiro](https://kiro.dev/docs/getting-started/installation/) | **Login** | Terminal | `kiro-cli login` | Complete browser flow; device flow remains available |
| [Kilo Code](https://kilo.ai/docs/code-with-ai/platforms/cli) | **Login** | Terminal | `kilo auth login` | Complete browser flow |
| [Vibe](https://github.com/mistralai/mistral-vibe) | **Set up** | Terminal | `vibe --setup` | Complete Vibe's native provider setup |
| [Muse](https://ai.meta.com/llama/) | **Login** | Terminal | `muse login` | Complete native login flow |
| [Agy](https://github.com/google-antigravity/antigravity-cli) | **Login** | Terminal | `agy` | Complete first-run browser sign-in |
| [Autohand](https://docs.autohand.ai/working-with-autohand-code/cli-reference) | **Login** | Terminal | `autohand login` | Complete Autohand account sign-in |
| [Kimchi](https://docs.kimchi.dev/docs/service-keys) | **Login** | Terminal | `kimchi login` | Complete browser login flow |
| [Prime Agent](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/packages/coding-agent/docs/quickstart.md) | **Login** | Terminal | `prime-agent` | Click **Open login** to send `/login` + Enter |
| [OMP](https://github.com/can1357/oh-my-pi) | **Login** | Terminal | `omp` | Click **Open login** to send `/login` + Enter |

Aider has no reliable general-purpose login/configure command. Selecting **Set up** therefore opens its official API-key documentation rather than starting a normal coding session. AO never receives the credential. After configuring Aider through its supported environment or configuration-file mechanisms, the user returns to Harness settings and selects **Check configuration**. The row reports **Configured**, **Not configured**, or **Configuration status unknown**.

Goose and Vibe do expose dedicated setup entry points, so they remain inline terminal flows using `goose configure` and `vibe --setup` respectively.

For Pi, Droid, Prime Agent, and OMP, AO starts only the native TUI. The button stays disabled until the terminal is attached. After the TUI is visible, the user selects **Open login** and AO offers the complete daemon-reviewed `/login` sequence to the existing terminal input path.

Qwen remains a setup flow, not a login flow. Its discontinued Qwen OAuth option is no longer represented as a login action: AO starts `qwen`, labels the action **Set up Qwen**, and exposes **Open setup** to enter `/auth`, where the user configures Alibaba ModelStudio, a third-party provider, or a custom provider, as described in the [official Qwen authentication docs](https://qwenlm.github.io/qwen-code-docs/en/users/configuration/auth/). Qwen's `i` + Backspace prefix works in both supported editor states: in optional Vim NORMAL mode, `i` enters INSERT and Backspace is harmless on the new empty prompt; with Vim mode disabled or already in INSERT, Backspace removes the literal `i`.

Autohand uses the dedicated `autohand login` command from its [official CLI reference](https://docs.autohand.ai/working-with-autohand-code/cli-reference). Pi uses the documented plain `pi` startup before the explicit `/login` action; `--verbose` is not added.

The UI changes to **Login opened** or **Setup opened** only after the active attachment accepts terminal input; a rejected attempt stays retryable. There are no startup-output markers, locale overrides, launch arguments containing slash commands, or automatic submissions.

All provider selection, browser/device flow, credential entry, and credential storage remain inside the vendor CLI or vendor-documented configuration surface. AO does not render or own a provider credential panel.

Authentication verification follows the readiness boundary introduced by #4589. The compatibility `POST /api/v1/agents/:agent/probe` route invalidates only that agent's authentication observation, then performs a targeted launch-purpose readiness ensure. Harness settings requests `POST /api/v1/agents/readiness/ensure` with `purpose=display` and caches the normalized snapshot. Authentication status remains advisory; normal session spawn is the authoritative validation boundary.

The terminal is bounded by the Settings component lifecycle: navigating away closes an in-progress auth terminal, and the frontend also requests closure after 15 minutes. Attachment errors do not imply that the daemon-side process exited and therefore do not trigger verification. A failed close keeps a retryable cleanup state until deletion is confirmed.

## Security model

- the renderer starts terminal flows by sending only the fixed agent ID
- launch mode, terminal argv, and documentation URL come from the daemon-owned plan registry
- launch argv is resolved through the registered harness adapter
- the optional terminal input is returned only from the selected daemon-owned plan; arbitrary commands are not accepted from the start request
- each terminal action requires an explicit user click and is marked complete only after the active attachment accepts it
- every auth terminal gets a unique private child workspace
- interactive bytes pass through AO's existing terminal transport, but AO does not persist or log raw input; the harness controls credential storage
- documentation flows open only the fixed vendor URL and never collect credentials in AO
- authentication/configuration status comes from the daemon-owned readiness coordinator and remains advisory

## Validation

- regenerated OpenAPI and frontend TypeScript contracts
- frontend TypeScript typecheck
- focused frontend Vitest suites: 25 tests passing
- affected Go auth service, controller, and API-spec tests passing
- `git diff --check`

The unrestricted frontend/backend sweeps are not used as the focused validation command in this checkout: unrelated socket, optional-dependency, and timing-sensitive suites can fail under sandbox or CPU contention. The focused changed-area suites pass serially.


nikhilachale:
Screenshot 2026-08-31 at 03 05 28
Screenshot 2026-08-31 at 03 05 42
Screenshot 2026-08-31 at 03 05 35

backend/internal/service/agentauth/service.go
```diff
@@ -0,0 +1,164 @@
...
+	plan.command = append([]string(nil), plan.command...)
+	if len(plan.command) == 0 {
+		plan.Available = true
+		return plan
+	}
+	if s.executables == nil {
+		plan.Reason = fmt.Sprintf("%s was not found on PATH.", plan.command[0])
+		return plan
+	}
+	path, err := s.executables.LookPath(plan.command[0])
```
illegalcall: [P1] Resolve through the registered harness adapter before trusting a generic PATH hit. This short-circuit skips adapter-specific validation. Muse is a concrete example: `ResolveMuseBinary` intentionally version-checks `muse` because unrelated tools use the same executable name, but this code will accept the first such binary and run `<unrelated muse> login`. That is especially easy to hit with the varied PATHs/package managers on macOS. Prefer `ResolveAgentBinary(ctx, plan.AgentID)` for registered harnesses and use its resolved path for argv[0].

backend/internal/service/shellterm/service.go
```diff
@@ -238,43 +277,84 @@ func (s *Service) OpenShellTerminal(ctx context.Context, in OpenShellTerminalInp
...
 	}
+	rollback := func() {
+		if destroyErr := s.runtime.Destroy(context.WithoutCancel(ctx), handle); destroyErr != nil {
+			s.log.Warn("shell terminal rollback failed; runtime may be orphaned",
+				"handleId", handle.ID, "error", destroyErr)
+		}
+	}
+
+	if cfg.initialInput != "" {
+		if err := waitForInitialInput(ctx, s.initialInputDelay); err != nil {
```
illegalcall: [P1] A fixed 500 ms delay does not establish that an interactive harness is ready for `/login`/`/auth`. Cold Node/Bun startup, migration/update checks, or a first-run screen on macOS can take longer, so the bytes can be ignored or become ordinary agent input. The terminal is also not persisted until after this delay/send, so a daemon crash in the window can leave exactly the untracked PTY the comment above says is prevented. Persist immediately, then wait for an observable ready state/output acknowledgement (or leave the slash command for the user); the current test sets this delay to zero and cannot catch the production race.

frontend/src/renderer/hooks/useAgentAuth.ts
```diff
@@ -0,0 +1,43 @@
...
+	return useMutation({
+		mutationFn: async (agentId: string): Promise<StartAgentAuthResponse> => {
+			const { data, error } = await apiClient.POST("/api/v1/agents/{agent}/auth", {
+				params: { path: { agent: agentId } },
+			});
+			if (error || !data) throw new Error(apiErrorMessage(error, "Could not start agent authentication."));
+			return data;
+		},
+		onSuccess: () => {
+			void queryClient.invalidateQueries({ queryKey: shellTerminalsQueryKey });
```
illegalcall, line 44: [P1] Put the newly created terminal into the query cache, or await the refetch, before navigating to its handle. This invalidation is fire-and-forget, while `startAuth` immediately selects the returned handle. If the cache already contains another standalone terminal, `ShellTerminalsView` mounts with that stale list, cannot find the auth handle, and replaces the selection with `shellTerminals[0]`; the later refetch leaves that old handle selected. The login then runs in the background and the user can miss its device code/prompt. Please cover the case with a pre-existing terminal.

frontend/src/renderer/stores/ui-store.ts
```diff
@@ -298,6 +318,18 @@ export const useUiStore = create<UiState>((set, get) => ({
...
+		set((state) => ({
+			agentAuthTerminalRequest: {
+				agentId,
+				handleId,
+				nonce: (state.agentAuthTerminalRequest?.nonce ?? 0) + 1,
+			},
+		})),
+	completeAgentAuthCheck: (nonce) =>
+		set((state) => state.agentAuthCheckRequest?.nonce === nonce
+			? { agentAuthCheckRequest: null, agentAuthTerminalRequest: null }
```
illegalcall: [P1] Do not consume the pending auth flow merely because one probe finished. Reopening Settings can happen before the browser/device flow has completed; the effect calls this in `finally`, so an unauthorized result or even a probe error clears the only automatic recheck. Finishing login later leaves the UI stale, and unauthorized rows do not expose `Check login`. There is also no correlation with `agentAuthTerminalRequest`, so an older probe can clear a newer agent's terminal request. Retain/retry until terminal completion or authorization, and correlate by agent/handle/generation.

backend/internal/service/agentauth/service.go
```diff
@@ -0,0 +1,164 @@
...
+// binary resolver when an installed harness is not visible on the daemon PATH.
+func NewWithAgentResolver(executables ExecutableFinder, agents AgentBinaryResolver, terminals TerminalOpener) *Service {
+	return &Service{executables: executables, agents: agents, terminals: terminals}
+}
+
+// Plans returns every known harness plan in stable Harness settings order.
+func (s *Service) Plans() []Plan {
+	out := make([]Plan, 0, len(plans))
+	for _, plan := range plans {
+		out = append(out, s.resolve(context.Background(), plan))
```
illegalcall: [P2] Thread the HTTP request context through `Plans`. Resolving all 27 entries can do filesystem work and adapter-aware binary discovery (including package-manager/runtime-manager probes), but `context.Background()` makes the GET impossible to cancel. This also violates the repo rule that I/O/blocking work accepts `context.Context` first. Change the service/controller interface to `Plans(ctx)` and pass `r.Context()`; do the same for `Plan` while it remains part of the service API.

backend/internal/service/agentauth/plans.go
```diff
@@ -0,0 +1,56 @@
...
+	plan("pi", ActionLogin, "Log in to Pi", []string{"pi"}, "/login\r", "Native Pi login flow", "https://github.com/earendil-works/pi"),
+	plan("amp", ActionLogin, "Log in to Amp", []string{"amp", "login"}, "", "Native browser flow", "https://ampcode.com/manual"),
+	plan("auggie", ActionLogin, "Log in to Auggie", []string{"auggie", "login"}, "", "Native browser flow", "https://docs.augmentcode.com/cli/overview"),
+	plan("droid", ActionLogin, "Log in to Droid", []string{"droid"}, "/login\r", "Native Droid login flow", "https://docs.factory.ai/droid-cli/cli-reference"),
+	plan("crush", ActionSetup, "Set up Crush", []string{"crush"}, "", "Native provider picker", "https://github.com/charmbracelet/crush"),
+	plan("cline", ActionLogin, "Log in to Cline", []string{"cline", "auth"}, "", "Native authentication flow", "https://github.com/cline/cline"),
+	plan("goose", ActionSetup, "Set up Goose", []string{"goose", "configure"}, "", "Native provider configuration", "https://block.github.io/goose/index.html"),
+	plan("qwen", ActionLogin, "Log in to Qwen", []string{"qwen"}, "/auth\r", "Native Qwen authentication flow", "https://qwenlm.github.io/qwen-code-docs/en/users/quickstart/"),
```
illegalcall: [P2] This should now be a setup action, not a login action. Current Qwen Code docs say Qwen OAuth was discontinued on 2026-04-15 and is no longer selectable; `/auth` configures Alibaba ModelStudio, a third-party provider, or a custom provider/API key. Rendering `Login` and describing this as a Qwen authentication flow sends users toward an account-login path that no longer exists. Use `ActionSetup` and update the guidance accordingly: https://qwenlm.github.io/qwen-code-docs/en/users/configuration/auth/

illegalcall (commented):
Reviewed the child diff from #4221's head (`a4c10bd`) to this PR (`8c9f52bd`), with emphasis on macOS behavior. The fixed daemon-owned command allowlist is the right security boundary, but there are several correctness races in launching/revealing slash-command terminals and refreshing auth state, plus one adapter-resolution bypass. Inline findings below.

backend/internal/service/agentauth/plans.go
```diff
@@ -0,0 +1,56 @@
...
+import "strings"
+
+// plans is the code-reviewed authentication allowlist in stable Harness
+// settings order. Commands must be added here, never supplied by clients.
+var plans = []Plan{
+	plan("claude-code", ActionLogin, "Log in to Claude Code", []string{"claude", "auth", "login"}, "", "Native browser/device flow", "https://code.claude.com/docs/en/installation"),
+	plan("codex", ActionLogin, "Log in to Codex", []string{"codex", "login"}, "", "Native browser/device-code flow", "https://github.com/openai/codex"),
+	plan("cursor", ActionLogin, "Log in to Cursor", []string{"cursor-agent", "login"}, "", "Native browser flow", "https://docs.cursor.com/en/cli/installation"),
+	plan("opencode", ActionLogin, "Log in to OpenCode", []string{"opencode", "auth", "login"}, "", "Native provider chooser", "https://github.com/anomalyco/opencode"),
+	plan("aider", ActionSetup, "Set up Aider", []string{"aider"}, "", "Configure provider/API key in the native prompt; docs fallback", "https://aider.chat/docs/install.html"),
```
illegalcall: [P1] Reconcile this with the PR security promise that AO “does not accept, persist, or proxy” API keys. Plain `aider` is not a dedicated OAuth login command; its setup flow (and several other `/auth`/configure plans) can prompt for an API key. Because this PTY is displayed and controlled through AO, typed credential bytes necessarily travel renderer → AO terminal transport/daemon → harness, even if AO never stores or interprets them. Either keep API-key-only targets as Instructions, or narrow and document the guarantee as “no credential-specific fields/storage/logging,” then audit the terminal path for redaction and telemetry.

backend/internal/service/agentauth/service.go
nikhilachale: Fixed in 4a7708ef7. Registered harnesses now resolve through ResolveAgentBinary first and use that validated path for argv[0]; direct PATH lookup is only used when no adapter resolver is wired. Added a Muse regression where a generic PATH hit must lose to the adapter-validated binary.

backend/internal/service/agentauth/plans.go
nikhilachale: Fixed in 4a7708ef7. Qwen now uses ActionSetup, points to the current authentication/configuration page, and tells the user to run /auth to configure a provider.

backend/internal/service/agentauth/service.go
nikhilachale: Fixed in 4a7708ef7. Plans and Plan now accept context.Context, resolution uses that context, and the controller passes r.Context() to Plans.

backend/internal/service/agentauth/plans.go
nikhilachale: Fixed in 4a7708ef7 and in the PR description. The guarantee is narrowed to no credential-specific fields/storage/logging: AO forwards terminal input to the harness without persisting or logging the raw input, while each harness controls credential storage. The terminal path was checked for raw-input logging, and setup guidance states this boundary explicitly.

frontend/src/renderer/hooks/useAgentAuth.ts
nikhilachale: Fixed in 4a7708ef7. The mutation now synchronously upserts the complete returned terminal into the shell-terminals query cache before navigation, then invalidates for reconciliation. The regression seeds a pre-existing terminal and verifies both handles remain in cache.

backend/internal/service/shellterm/service.go
nikhilachale: Fixed in 4a7708ef7. Removed initial-input injection and the fixed delay entirely. Command terminals are persisted immediately after runtime creation; slash-command plans now tell the user to enter /login or /auth in the visible terminal.

frontend/src/renderer/stores/ui-store.ts
nikhilachale: Fixed in 4a7708ef7. Unauthorized/error probes clear only the individual check while retaining the pending auth terminal; authorized probes clear the flow. Completion is correlated by agent, handle, check nonce, and a monotonic generation, stale probes cannot clear newer checks, and unauthorized rows expose Check login. Added store race tests and a shell integration test for a second flow after the first completes.

illegalcall:
Follow-up review on current head ef65ad44: the seven issues from my earlier auth review appear materially fixed. I am still holding approval for these remaining items:

1. Auth terminals launch plain interactive coding agents such as Aider, Qwen, Pi, and Droid with s.dataDir (~/.ao) as their working directory. Those agents can treat AO's database/config/runtime directory as the project and scan, initialize, or modify it. Please use a dedicated empty auth workspace under ~/.ao, not the daemon data root.
2. This branch still contains the obsolete #4221 parent a4c10bd rather than current #4221 head 27fa5244. Restacking produces integration conflicts across installer APIs, generated schemas, UI, and shell terminals, so it needs to be rebased/restacked and retested against the final parent.
3. kimchi setup is not the documented authentication flow. Current first-party guidance launches kimchi and uses /login; setup-tools is unrelated. The plan should use the documented flow and link first-party docs.

For Aider specifically, plain aider starts a coding session rather than a dedicated login flow, which makes isolating the working directory important: https://aider.chat/docs/usage.html

CI is green, but I do not think this head is approval-ready until these are addressed.

nikhilachale:
Addressed on `2cbb05213`:

- Auth command terminals now use the dedicated `~/.ao/auth-workspace` directory instead of the daemon data root, with regression coverage for both workspace selection and creation failure.
- The branch is restacked directly onto the final #4221 head, `27fa524`, with installer APIs, generated schemas, UI, and shell-terminal lifecycle changes reconciled and retested.
- Kimchi now launches `kimchi`, directs the user to run `/login`, and links the first-party Kimchi CLI documentation.

I also completed a fresh independent review of the full `27fa524..2cbb05213` diff; it found no additional actionable issues. All PR CI checks are green, including `go test -race ./...`.

nikhilachale:
Implemented the inline harness authentication flow requested in follow-up review (head `c408fe9d6`).

- Login/Set up now expands a 300px terminal directly beneath the harness row; it no longer navigates to the standalone terminals page or opens an external terminal.
- Pi, Droid, Prime Agent, and OMP automatically submit `/login`; Qwen automatically submits `/auth` after PTY output confirms readiness.
- Kimchi now uses its documented direct `kimchi login` flow.
- Auth PTYs remain isolated in `~/.ao/auth-workspace`, time out after 15 minutes, clean up on close/unmount, and re-probe authentication on exit. Cleanup treats an already-pruned PTY as success so Retry cannot deadlock.

Verification: focused Go tests + vet pass; TypeScript typecheck passes; 58 focused renderer tests pass. Full renderer run reached 3,140 passing tests; unrelated landing suites are currently missing `cheerio` / `@ao/shared` in this checkout. The full Go run passed the changed packages and later stopped when the host temp volume ran out of space. GitHub has not reported checks for the new head yet.

nikhilachale:
Screenshot 2026-09-01 at 15 33 08

illegalcall:
Fresh re-review of current head `fcd4b456` against the latest parent PR #4221.

The earlier dangerous cwd problem is fixed: auth terminals now start under `<AO_DATA_DIR>/auth-workspace` instead of AO's database/config root. I am holding approval for the following current issues:

1. **The stack is stale and unvalidated.** GitHub currently reports this PR `CONFLICTING`/`DIRTY` with no checks. Merging the latest #4221 head (`fc1b9d849`) into this head produces content conflicts in:
   - `backend/internal/httpd/api.go`
   - `backend/internal/service/shellterm/service_test.go`
   - `backend/internal/service/systeminstall/systeminstall.go`
   - `frontend/src/renderer/components/settings/HarnessSettingsSection.tsx`

   Please restack on the final #4221 head, retain both installer and auth behavior while resolving these files, and rerun the full affected Go/frontend/API suites. This alone prevents approval of the current head.

2. **The promised "empty" auth workspace is one persistent shared directory.** `OpenCommandTerminal` only calls `MkdirAll(<AO_DATA_DIR>/auth-workspace)`; it neither clears the directory nor creates a unique child for an auth flow. Project-aware agents can leave files/config/instructions that later or concurrent auth agents observe. This is not an OS privilege boundary because all harnesses already run as the same user, but it contradicts the PR's explicit isolation promise. Use a private per-terminal/per-flow child directory and clean it with the terminal lifecycle, or update the security model and explain why shared state is acceptable. Evidence: [implementation](https://github.com/Untrivial-ai/agent-orchestrator/blob/fcd4b456e0d113df103083a00b6369e9013b9413/backend/internal/service/shellterm/service.go#L247-L271); the current test proves emptiness only for a newly created temporary directory.

3. **Slash-command auto-submission does not match the PR body and uses an ambiguous readiness signal.** The body says AO tells the user to type `/login` or `/auth`, but Pi/Droid/Qwen/Prime Agent/OMP now inject the command automatically. `sendInitialInputWhenReady` treats any nonblank PTY output as readiness; a splash, update notice, or migration message can arrive before the TUI's input loop is ready. PTY buffering may make this work for some harnesses, so I am not claiming guaranteed failure, but the current heuristic is not evidence of prompt readiness. Either keep the documented user-entered flow, add harness-specific/proven readiness, or provide packaged-mac validation for every auto-injected harness and update the body. Evidence: [auth plans](https://github.com/Untrivial-ai/agent-orchestrator/blob/fcd4b456e0d113df103083a00b6369e9013b9413/backend/internal/service/agentauth/plans.go#L16-L34), [readiness implementation](https://github.com/Untrivial-ai/agent-orchestrator/blob/fcd4b456e0d113df103083a00b6369e9013b9413/backend/internal/service/shellterm/service.go#L275-L300).

4. **Aider OAuth-only setup cannot be positively reflected by the advisory probe.** Current Aider loads OpenRouter OAuth credentials from `~/.aider/oauth-keys.env`; AO checks environment variables, `.env`, and `.aider.conf.y[a]ml`, but not that file. After a successful OAuth flow, AO therefore continues to show `unknown` unless another key source happens to exist. Authentication itself still works and the status is explicitly advisory, so this is a UX/probe completeness issue rather than a security blocker. Evidence: [AO probe](https://github.com/Untrivial-ai/agent-orchestrator/blob/fcd4b456e0d113df103083a00b6369e9013b9413/backend/internal/adapters/agent/aider/auth.go#L43-L81), [Aider's current OAuth file loading](https://github.com/Aider-AI/aider/blob/main/aider/main.py).

Documentation corrections:

- `kimchi login` in the implementation is correct current first-party behavior. Update the PR body, which still says AO launches `kimchi` and directs the user to `/login`; do not revert the code. [Kimchi documentation](https://docs.kimchi.dev/docs/service-keys).
- "Durable AO terminal" is imprecise for the inline UI: the frontend intentionally closes it on component unmount and after 15 minutes. The bounded cleanup is reasonable, but the body should state that navigating away terminates an in-progress auth flow. [Lifecycle code](https://github.com/Untrivial-ai/agent-orchestrator/blob/fcd4b456e0d113df103083a00b6369e9013b9413/frontend/src/renderer/components/settings/HarnessSettingsSection.tsx#L310-L329).

Corrections to avoid unnecessary work: I am **not** treating the direct `os.MkdirAll` as an architecture blocker, and `context.WithoutCancel` is appropriate here because the delayed-input worker must outlive the completed HTTP request and is bounded by its own ten-second timeout.

illegalcall:
Follow-up on current head `2ff963246`: the per-flow auth workspace, confirmed-dead cleanup, Aider OAuth probe, Kimchi command/docs, and terminal-lifetime wording address the corresponding findings from my previous review. Two approval blockers remain.

1. **Pi's `pi v` readiness marker is emitted before Pi can process `/login`.** AO treats `pi v` as an editor-ready signal and sends `/login` immediately:

   https://github.com/Untrivial-ai/agent-orchestrator/blob/2ff9632464d0ec6a0d739b843052b83174f06a8f/backend/internal/service/agentauth/plans.go#L20

   In current first-party Pi source, however, Pi installs a startup-only submit handler and then renders the `pi v…` header. It subsequently awaits managed `fd`/`rg` setup and only afterward installs the real editor submit handler:

   - Startup handler and header: https://github.com/earendil-works/pi-mono/blob/96317e50b8d6e7f6d0e47fd29122baf1461c00f5/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L950-L964
   - Real handler installed later: https://github.com/earendil-works/pi-mono/blob/96317e50b8d6e7f6d0e47fd29122baf1461c00f5/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L1022-L1035
   - Early submission only reports `Startup is still in progress`: https://github.com/earendil-works/pi-mono/blob/96317e50b8d6e7f6d0e47fd29122baf1461c00f5/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L2973-L2979

   Therefore, especially when Pi needs to download either managed tool, AO can submit `/login` during startup, Pi does not execute it, and AO never retries. Please use a proven post-initialization signal or keep Pi's `/login` user-entered instead of claiming `pi v` is editor-ready.

2. **The stack is still stale and unvalidated.** GitHub reports this PR `CONFLICTING`/`DIRTY` with no checks. Its merge-base with current #4221 remains `27fa5244`, while #4221 is now `2ea1059d`. Combining the current heads produces content conflicts in 11 files, including the HTTP API, OpenAPI/schema artifacts, installer service, shell-terminal tests, and Harness settings UI/tests. Please restack on the final #4221 head, preserve both feature sets while resolving the conflicts, and rerun the full backend/frontend/API and packaged-macOS validation.

illegalcall:
Follow-up review on current head `b72f72af5`: the stale/conflicting stack blocker is resolved and all ten checks are green. I am still holding approval for two confirmed functional issues, plus two lower-severity cleanup items.

1. **Pi automatic `/login` still races startup.** AO still treats the `pi v` header as editor-ready and sends `/login` once:

   https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/backend/internal/service/agentauth/plans.go#L20

   In current first-party Pi source, the version header is rendered while the startup-only submit handler is active. Pi then awaits managed `fd`/`rg` setup before installing the real submit handler. An early submission only reports `Startup is still in progress`, and AO returns after its one send rather than retrying. This is especially reproducible on a fresh Mac when either tool must be downloaded.

   - Startup handler and header: https://github.com/earendil-works/pi/blob/4e69b0c28060f0f02fbe38bfa7c21a2e2eb25057/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L950-L964
   - Tool setup and real handler: https://github.com/earendil-works/pi/blob/4e69b0c28060f0f02fbe38bfa7c21a2e2eb25057/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L1022-L1035
   - Early-submit behavior: https://github.com/earendil-works/pi/blob/4e69b0c28060f0f02fbe38bfa7c21a2e2eb25057/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L2973-L2979
   - AO sends once and returns: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/backend/internal/service/shellterm/service.go#L297-L324

   Please use a proven post-initialization signal or leave Pi `/login` user-entered.

2. **Check login can return the cached pre-login result.** The UI calls `/probe` after the terminal exits, but `Service.Probe` uses a launch-purpose readiness Ensure. The coordinator returns a cache hit while authentication is younger than the 30-second launch TTL, and completing an auth terminal does not invalidate that observation. A fast browser/device login can therefore still be reported as Not logged in; repeated Check login clicks remain stale until the TTL expires.

   - UI reconciliation: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/frontend/src/renderer/components/settings/HarnessSettingsSection.tsx#L249-L287
   - Probe uses Ensure: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/backend/internal/service/agent/readiness.go#L182-L200
   - Cache-hit and TTL behavior: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/backend/internal/service/agent/readiness_coordinator.go#L283-L299 and https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/backend/internal/service/agent/readiness_coordinator.go#L584-L597

   Please force or invalidate the targeted authentication observation before this user-requested verification, with a regression test that starts from a fresh cached unauthorized result.

Lower-severity follow-ups:

- The 15-minute close uses `.finally()` to enter `timed_out` even when DELETE fails, which stops the timer while a macOS `ptyhost` and its private workspace may remain alive. A terminal attachment `error` also triggers verification and leaves `running` without server-side death confirmation. Preserve a cleanup/retry state until close is confirmed: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/frontend/src/renderer/components/settings/HarnessSettingsSection.tsx#L307-L325 and https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/frontend/src/renderer/components/settings/HarnessSettingsSection.tsx#L485-L495
- The old standalone auth-terminal request/generation/router flow remains, but no production code calls `requestAgentAuthTerminal`; the new implementation renders the workflow inline. Please remove the unreachable store/router machinery and its tests rather than maintaining two architectures: https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/frontend/src/renderer/stores/ui-store.ts#L97-L109 and https://github.com/Untrivial-ai/agent-orchestrator/blob/b72f72af5c4dd64fedcd0902d0018a39808d8bfe/frontend/src/renderer/stores/ui-store.ts#L443-L463

I am not treating the clean merge, generated API artifacts, per-flow auth workspace, Aider OAuth probe, Kimchi flow, or the bounded detached input worker as blockers on this head.

illegalcall (approved):

Issue #4526, opened 2026-08-27
feat(chat): keep native harness processes alive across daemon restart and desktop update


AgentWrapper:
## Problem

Chat-mode sessions are called runtime-less and each controller owns a provider child process (`codex app-server` or ACP). Daemon/Desktop shutdown calls `chatSvc.StopAll`, closes every child, marks the durable session Exited, and boot launches a replacement provider process plus native resume/history import. This interrupts AO's active controller generation and makes desktop updates materially different from terminal/tmux sessions, whose harness process survives and is adopted.

**Analyzed against:** `4cbb4b6ced1ad93f79641a2347d2342f1ffd218a` (`v0.12.9-nightly.202608262313`)
**Confidence:** High — shutdown, child ownership, recovery, and observed process/timing paths were traced locally.
**Priority:** P1 — closing/updating the desktop interrupts active Chat work; there is no continuity-preserving workaround other than keeping the whole old daemon alive, which conflicts with applying daemon updates.

## Required architecture

Introduce a provider-neutral, per-session Chat host outside the daemon:

- The host owns provider stdio and remains alive when its daemon control client disconnects.
- A versioned authenticated loopback control transport permits exactly one daemon controller owner and reconnects after restart.
- Provider output is sequence-journaled while detached and replayed before live output; daemon acknowledgements allow bounded compaction. Native history remains the correctness repair path.
- Durable controller generations continue fencing SQLite projection. Host attachment additionally enforces one live transport owner.
- Credentials remain only in the host/child environment. Host metadata stores an opaque endpoint, protocol version, process identity, and a random 0600 capability; never provider credentials.
- Protocol compatibility is negotiated. An incompatible new daemon must leave a live old host untouched and surface a typed recovery error. Idle hosts can later be rolled to a new binary; active turns are never killed merely to update the host.
- Explicit session kill terminates the host. Daemon crash/desktop close/update only detaches. Boot reaps authenticated hosts for missing/terminated sessions; it must not infer death from a failed probe.
- Window close may keep the daemon or detach according to desktop ownership policy, but neither path kills a Chat host. Full quit/update stops the daemon and detaches hosts. OS reboot/provider crash falls back to existing native resume.

## Smallest end-to-end slice

1. Add a detached `ao chat-host` process with a stable v1 authenticated loopback protocol and atomic 0600 descriptor under `AO_DATA_DIR`.
[...]

PR #4529, opened 2026-08-27
feat(chat): keep Codex provider alive across daemon restart

AgentWrapper:
Refs #4526

## Summary

- move Codex Chat app-server ownership into a detached, authenticated per-session host
- reconnect one fenced daemon controller to the same initialized provider, replaying detached output and unresolved approvals without false Exited projection
- distinguish daemon detach from explicit session termination and reconcile proven durable orphans on boot
- document the compatibility, security, update-handoff, crash-recovery, branch-transition, and ACP migration boundaries

## Verification

- `npm run lint` (with AO worker environment unset)
- `go test -count=1 -timeout=60s ./internal/adapters/chatdriver/persistenthost ./internal/adapters/chatdriver/codexappserver ./internal/service/chat ./internal/daemon`
- `go test -race -count=1 ./internal/adapters/chatdriver/persistenthost ./internal/adapters/chatdriver/codexappserver ./internal/service/chat`
- `go build ./...`
- `GOOS=windows GOARCH=amd64 go build ./cmd/ao`

## Known follow-ups

- ACP Chat bindings still use daemon-owned processes and native resume; issue #4526 remains open for that migration.
- The v1 host uses bounded in-memory replay. Host or machine failure falls back to native history; a disk-backed acknowledged journal is a future protocol revision.
- Branch activation retains launch-before-destroy rollback safety through a direct staged provider and therefore has a bounded live-reconnect gap until the next daemon reconciliation.

### DIFF
--- a/backend/e2e/chat_lifecycle_test.go
+++ b/backend/e2e/chat_lifecycle_test.go
@@ -61,44 +61,35 @@ func TestChatSurvivesADaemonRestartWithNativeContext(t *testing.T) {
 	}
 }
 
-// A crash mid-turn must not leave the turn looking like it is still working. A
-// controller that stopped running a turn is not evidence the work finished, and a
-// turn stuck at "running" forever is a session the user cannot reason about.
-func TestChatRestartMidTurnSettlesTheTurnHonestly(t *testing.T) {
+// The detached Codex host is outside the daemon's process group. Even SIGKILL of
+// the daemon must leave the provider turn running for the replacement to adopt.
+func TestChatRestartMidTurnKeepsCodexRunning(t *testing.T) {
 	requireE2E(t)
 	dataDir := t.TempDir()
 	d := startDaemon(t, dataDir)
 	project := seedProject(t, d, "midturn")
 	session := chatSession(t, d, project, "Reply with exactly: READY")
 
-	send(t, d, session, "Sleep for 120 seconds using a shell command, then reply with exactly: NEVER", "long")
+	send(t, d, session, "Run the shell command `sleep 20`, then reply with exactly: SURVIVED-SIGKILL", "long")
 	d.awaitConversation(session, 90*time.Second, "the long turn to start running", func(s snapshot) bool {
 		return s.Turns[len(s.Turns)-1].State == "running"
 	})
+	hostBefore := persistentHostPID(t, dataDir, session)
 
-	// SIGKILL: no shutdown hook runs, so whatever the next boot concludes it has to
-	// conclude from disk alone.
 	d.kill()
+	if !processAlive(hostBefore) {
+		t.Fatalf("detached host %d died with the killed daemon", hostBefore)
+	}
 	restarted := startDaemon(t, dataDir)
-
-	snap := restarted.awaitConversation(session, 90*time.Second, "the orphaned turn to settle",
+	restarted.awaitLiveController(session, 90*time.Second)
+	snap := restarted.awaitConversation(session, 3*time.Minute, "the surviving turn to finish",
 		func(s snapshot) bool { return terminal(s.Turns[len(s.Turns)-1].State) })
 
 	last := snap.Turns[len(snap.Turns)-1]
-	if last.State == "completed" {
-		t.Fatalf("a turn interrupted by a daemon crash was recorded as completed:\n%s", describe(snap))
-	}
-	if last.ErrorMessage == "" {
-		t.Errorf("orphaned turn %s carries no explanation of why it ended", short(last.ID))
-	}
-
-	// And the session is still usable afterwards, rather than wedged.
-	restarted.awaitLiveController(session, 90*time.Second)
-	send(t, restarted, session, "Reply with exactly: ALIVE", "post-crash")
-	revived := restarted.awaitConversation(session, 3*time.Minute, "a turn after the crash",
-		func(s snapshot) bool { return contains(s.assistantText(), "ALIVE") })
-	if !contains(revived.assistantText(), "ALIVE") {
-		t.Errorf("the session never recovered after a mid-turn crash:\n%s", describe(revived))
+	hostAfter := persistentHostPID(t, dataDir, session)
+	t.Logf("abrupt daemon simulation: host_pid=%d->%d turn_state=%s", hostBefore, hostAfter, last.State)
+	if hostAfter != hostBefore || last.State != "completed" || !contains(snap.assistantText(), "SURVIVED-SIGKILL") {
+		t.Fatalf("real Codex turn did not survive abrupt daemon replacement:\n%s", describe(snap))
 	}
 }
 
@@ -112,7 +103,7 @@ func TestChatQueuedMessageIsAccountedForAcrossARestart(t *testing.T) {
 	project := seedProject(t, d, "queuecrash")
 	session := chatSession(t, d, project, "Reply with exactly: READY")
 
-	send(t, d, session, "Sleep for 120 seconds using a shell command, then reply with exactly: NEVER", "qc-slow")
+	send(t, d, session, "Run the shell command `sleep 20`, then reply with exactly: FIRST-SURVIVED", "qc-slow")
 	d.awaitConversation(session, 90*time.Second, "the slow turn to start running", func(s snapshot) bool {
 		return s.Turns[len(s.Turns)-1].State == "running"
 	})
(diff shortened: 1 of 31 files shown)


illegalcall:
I reproduced and reviewed this end to end, then pushed the recovery fixes in `dd27591ee` and merged current `main` in `64184b3be`.

## Root cause

The daemon directly owned the provider child process and its stdio. Desktop/daemon shutdown called `chat.Service.StopAll`, which closed every conversation/controller. Closing the provider stream produced EOF, so the controller settled an active turn as failed/exited. Restart then created or natively resumed a new provider process. The bug is therefore an ownership/lifecycle conflation: **detach AO UI/daemon state** and **terminate provider work** were represented by the same operation.

## Is the PR approach correct?

Yes for the Codex slice: a detached per-session host, loopback authentication, exclusive ownership descriptor, bounded replay, reconnect to the same provider PID, and distinct detach-versus-terminate semantics are the right boundary.

The original revision was not safe to merge as-is. I found three additional recovery failures:

1. a dead `host.lock` could permanently block that session;
2. reconnecting to a live host did not restore a durable running turn as busy, so AO could dispatch a second provider turn;
3. an incompatible/uncertain live-host attach was wrapped as an ordinary driver failure, allowing session-manager fallback reconciliation to mark the session exited and remove the worktree.

## Red evidence

On pre-PR `main`, I added a temporary service-level reproduction and ran it three times:

```text
cd backend
go test -count=3 ./internal/service/chat \
  -run ^TestReproDaemonShutdownPreservesActiveChat$ -v

--- FAIL: TestReproDaemonShutdownPreservesActiveChat
    daemon shutdown settled the in-flight turn as failed
```

It failed 3/3. Before applying the follow-up fixes, the new regression tests also failed deterministically:

```text
TestAcquireHostLockReclaimsDeadOwner:
  acquire stale host lock: chat host already exists

TestServiceLiveReconnectKeepsDurableRunningTurnBusy:
  attempted a second turn and hit the provider-turn unique constraint

TestReconcileLive_InconclusiveChatRecoveryDoesNotTeardown:
  fell back to save-and-teardown instead of preserving ownership as inconclusive
```

## Changes made

- reclaim `host.lock` only when its recorded PID is proven dead;
- add typed ownership-inconclusive errors from host -> Codex driver -> session manager;
- preserve the durable session/worktree when live-provider ownership is uncertain;
- load the durable snapshot during live reconnect and restore running-turn ownership before accepting another send;
- define the optional host capabilities as named port contracts;
- make host run/spawn context-aware;
- return CLI `usageError` for malformed internal `chat-host` arguments and add the CLI test;
- remove duplicate Codex initialization;
- resolve the current-main conflicts while preserving the newer provider-scope/branch semantics.

## Green evidence

Each lifecycle invariant passed 3/3 locally:

```text
TestHostReconnectsSameProviderAndReplaysDetachedOutput   PASS x3
TestAcquireHostLockReclaimsDeadOwner                    PASS x3
TestServiceStopAllOnlyDetachesPersistentConversation    PASS x3
TestServiceLiveReconnectKeepsDurableRunningTurnBusy     PASS x3
TestReconcileLive_InconclusiveChatRecoveryDoesNotTeardown PASS x3
```

`TestHostReconnectsSameProviderAndReplaysDetachedOutput` asserts both **the same provider PID** and replay of output emitted while detached.

Additional local verification:

```text
go test -race -count=1 ...persistenthost ...codexappserver ...service/chat ...session_manager  PASS
go build ./...                                                                    PASS
GOOS=windows GOARCH=amd64 go build ./cmd/ao                                      PASS
npm run frontend:typecheck                                                       PASS
golangci-lint with a clean cache                                                 0 issues
```

GitHub CI on `64184b3be`: **8/8 green** — build-test, lint, API drift, gitleaks, container CLI E2E, and native CLI E2E on macOS/Linux/Windows.

A screenshot is not relevant evidence for this backend process-lifecycle defect; a UI can look identical while silently replacing the provider. PID continuity, replay, durable busy state, and non-destructive reconciliation are the exact evidence.

## Remaining scope

This is now mergeable as the **Codex persistent-host slice**, but it does not by itself complete every item in #4526. ACP-provider integration and disk-backed acknowledged replay remain follow-ups. Branch replacement still has the bounded direct-provider handoff gap already documented in this PR/ADR; closing that requires routing branch staging/activation through the persistent host protocol rather than the daemon-owned fallback.

illegalcall:
I pushed the additional restart hardening and current-`main` merge in `c917c78dc`. GitHub now reports the PR **MERGEABLE** (checks are queued).

## What I simulated on the exact pushed merge candidate

These use the real installed Codex CLI (`codex-cli 0.149.0`), a real built AO daemon, throwaway AO data, and a throwaway Git project—not a mocked provider:

```text
TestChatRestartMidTurnKeepsCodexRunning
  abrupt daemon simulation: host_pid=80734->80734 turn_state=completed
  PASS (38.58s)

TestChatQueuedMessageIsAccountedForAcrossARestart
  queued-across-restart resolved as "completed"
  PASS (35.71s)

TestChatTurnSurvivesGracefulUpdaterStyleRestart
  graceful updater simulation: host_pid=85914->85914 turn_state=completed
  PASS (35.85s)

TestChatSessionRecoversAfterMachineStyleHostLoss
  old_host_pid=88622 new_host_pid=89721 interrupted_state=interrupted
  subsequent real Codex turn completed through native resume
  PASS (14.50s)
```

So the core promise is now directly demonstrated: graceful desktop/updater shutdown and abrupt daemon `SIGKILL` preserve the **same detached host/provider**, replay output, finish the in-flight turn, and then drain queued input. Losing the host/provider itself cannot preserve that process; AO records the lost turn as interrupted, starts a new host, resumes native history, and remains usable.

I also launched the actual Electron app against isolated AO data, verified the renderer and daemon were healthy, stopped the first Electron/daemon pair, and launched the replacement against the same data. A screenshot is not useful proof for this backend lifecycle bug: PID continuity, durable turn state, replay, and successful post-loss work are the exact evidence.

## Additional failures found and fixed

1. **Replay burst dropped terminal lifecycle frames.** Pre-fix, 4096 deltas filled the RPC notification buffer and `turn/completed` was silently discarded:

   ```text
   turn/completed was dropped when replay exceeded the 4096-notification queue
   ```

   The RPC reader now feeds an ordered relay queue, so response correlation stays live and lifecycle frames are not dropped. The 4097-frame regression passes 5/5.

2. **Concurrent stale-host recovery could launch rival providers.** Two stale observers could let one publish a replacement and then let the other delete its descriptor/lock:

   ```text
   stale cleanup launched rival hosts: host_pids=34659,34661 provider_starts=[34660 34662]
   ```

   Stale clients/reconcilers no longer delete ownership without first claiming the `O_EXCL` host lock. The deterministic two-starter regression passes 5/5.

3. **Current-main branch handoff semantics conflicted with persistent detach.** A native edit must release the old writer, not merely detach from it. The merge resolution preserves main's writer fence and explicitly terminates the source host before the replacement resumes. A dedicated regression proving branch handoff calls terminate rather than detach passes 10/10.

## Broader verification

```text
go test -race ./internal/service/chat ./internal/adapters/chatdriver/persistenthost \
  ./internal/adapters/chatdriver/codexappserver ./internal/session_manager
PASS

GOFLAGS=-p=1 npm run lint
all backend tests PASS; golangci-lint: 0 issues

npm run frontend:typecheck
PASS
```

## What still does not survive every possible failure

I reproduced one real protocol durability gap that this v1 design explicitly defers: if a replacement daemon receives replayed completion, then crashes **after socket delivery but before the SQLite projection commits**, the host has already discarded that frame. A third attachment receives no replay. Correctly closing this requires a sequence-numbered durable journal plus an acknowledgement sent only after AO commits the projection; an in-memory/socket-write acknowledgement cannot make that crash window safe.

Also, ACP providers are still daemon-owned, and ordinary branch activation / fresh approximate edit can still use the documented direct staged-provider fallback. Native Codex edit forks are now persistent after the current-main merge, but the remaining branch paths need host-mediated replacement (or per-branch host identity) to remove their bounded restart gap.

Verdict: **the primary Codex daemon/updater restart fix works and is well evidenced; it is not an all-providers, all-double-crash durability solution.** The pushed fixes make the scoped PR mergeable pending CI, while the acknowledged journal, ACP migration, and remaining branch-host replacement should stay explicit follow-ups.

nikhilachale (approved):
Tested locally, including an in-flight Codex turn across an Electron/daemon restart. Recovery worked correctly and all CI checks pass. Looks good to merge.

PR #4468, opened 2026-08-26
fix(chat): clarify live file references

illegalcall:
Code changes: +115 -88  
Tests: +113 -10  
Others: +32 -0  

## Summary

- describe selected `@file` chips as localized, focusable path references instead of implying that file contents were attached
- expose the complete worktree-relative path and path-only semantics on pointer hover, keyboard focus, and through accessibility APIs
- share the existing workspace-file cache, keep cached suggestions available after a refresh failure, surface load/refresh failures, and poll only while recovery is needed
- reject filenames containing carriage returns or line feeds because they cannot round-trip as one atomic token through the plain-text composer

## Already present on `main`

Comparative native-Electron testing found that `main` already provides:

- basename file chips after selection
- parent-directory labels and duplicate-basename disambiguation in the suggestion menu
- plain worktree-relative path serialization, including quoting paths containing spaces
- live create, rename, and delete updates through the existing workspace event stream

This PR preserves those behaviors. It does not introduce the chip, duplicate-name labels, path-only transport, or live refresh.

## Root cause

The existing composer renders an attachment-shaped basename chip while serializing only ordinary relative-path text. On `main`, the selected chip has no path-reference accessible name or description, is skipped by Tab, and offers no full-path explanation after selection. The user therefore cannot discover from the selected token that AO did not attach or embed the file contents.

## Verification

Exact candidate: `700d81ccd97709f7d7d00eb1a54502257901c637`

- real Electron comparison against `main` (`768ab5034b9815e76edbc936e14a4780beec8fc9`) using the same isolated session
- `main`: selected chips are not keyboard-focusable and expose no path-only description; duplicate-name labels and live create/rename/delete already work
- PR: Tab focuses the selected chip and pointer hover/focus exposes its full path and path-only description
- PR: a real filename containing a line feed is excluded from suggestions
- focused renderer suite: **4 files / 185 tests passed**
- explicit pointer-hover tooltip test: **1 passed**
- `npm run typecheck`: passed
- `git diff --check`: passed

Failure/recovery states are covered by component tests rather than a manufactured live-daemon outage: initial load failure is announced, cached suggestions remain usable with a stale-data warning after refresh failure, and polling runs only while recovery is needed.

### New behavior: selected chips disclose their actual semantics

Pressing Tab focuses the chip and exposes the complete path-reference-only explanation. Pointer hover exposes the same tooltip.

Keyboard focused file chip showing its full path reference tooltip

### Preserved behavior: transport remains ordinary path text

[...]

PR #4540, opened 2026-08-27
fix(chat): preserve attachment delivery semantics

### LINKED ISSUE #4427: bug(chat): attachment UI hides delivery mode and drops original filenames
## Bug

The same attachment chip represents different provider delivery semantics without disclosure. Codex receives a staged workspace path it must read; image-capable ACP agents may receive native bytes plus the path. The original filename exists only before send and is lost from both payloads/durable history, which later displays a randomized staging name.

A related race also lets existing text send while a selected file is still being read, potentially leaving that attachment on the next draft.

**Source:** [ChatUI QA and regression harness PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419)  
**Reported by:** Dhruv Sharma via local native-Electron QA  
**Analyzed against:** `d71acca1e13ddd9ab3a01a43f029853686f4c693`; transport path rechecked on `04b676c9abd2fac4505081b0c099e079a30bde57`  
**Environment:** macOS arm64, Codex Chat and provider-capability fixtures  
**Priority:** P2 — delivery works, but users cannot tell what the provider received  
**Confidence:** High — positive native attachment flow plus end-to-end payload trace

## Reproduction

1. Select a named PNG while text is already present.
2. Observe the original filename in the pre-send chip.
3. Send and inspect provider behavior/history.
4. Codex requests tool/path access despite the visible image preview.
5. Durable history identifies the asset by a generated staging filename.
6. With a delayed `FileReader`, Send can commit existing text before the attachment settles.

## Expected

- Send waits for all attachment reads to settle.
- Pre-send UI says **workspace file the agent must read**, **embedded image**, or **embedded image plus staged path** according to negotiated capability.
- A sanitized original display name survives request, durable storage, and transcript rendering.

## Actual

Delivery is provider-dependent but visually indistinguishable, original identity is dropped, and attachment settlement is not part of the send boundary.

## Root cause

- `frontend/src/renderer/hooks/useFileAttachments.ts:13-34` retains `name` locally; payload builders around `:177-187` drop it.
- `backend/internal/httpd/controllers/dto.go:214-223` and `backend/internal/ports/session.go:52-59` have no original-name field.
- `backend/internal/session_manager/chat_attachments.go:50-76` allocates randomized durable names.
- `frontend/src/renderer/components/chat/ChatComposer.tsx:426-448` stages a path and sends native bytes only when `nativeImages` is true.
- `SessionChatSurface.tsx:330-331` selects that behavior from provider capability without disclosure.

## Fix direction

Use an awaited settled-payload boundary before Send, expose per-file reading/ready/failed state, preserve a sanitized display name through DTO/domain/durable content, and render capability-derived delivery semantics next to each attachment.

## Acceptance

- A delayed file read cannot let text send early.
- Failed reads remain retryable/removable in the current draft.
- Delivery mode is accurate for path-only, native-only, and combined capabilities.
- Original display name survives staging and history.
- `npm run qa:chatui -- --grep MQA-10` passes in strict mode.

## Evidence

Image attachment requires later path/tool access

- [Attachment transport inventory](https://github.com/Untrivial-ai/agent-orchestrator/blob/b7d4a8dc8cf8843be42a3c0bb183e35103ccd06f/docs/qa/chatui-2026-08-25/static-feature-inventory.md)

## Related

- [#3722](https://github.com/Untrivial-ai/agent-orchestrator/issues/3722) — fixed path text versus transcript preview
- [#3884](https://github.com/Untrivial-ai/agent-orchestrator/issues/3884) — fixed attachment-byte loss after restart
- [PR #4092](https://github.com/Untrivial-ai/agent-orchestrator/pull/4092) — preserves attachments during steering
- [PR #4419](https://github.com/Untrivial-ai/agent-orchestrator/pull/4419) — three MQA-10 contracts

illegalcall:
Code changes: +1261 -253  
Tests: +2160 -43  
Others: +8 -0  

Closes #4427

> [!IMPORTANT]
> This focused PR is intentionally stacked on #4468 (`codex/fix-4426-file-chip-semantics`). It is draft until the final latest-main restack and native Electron/MQA-10 evidence pass are complete.

## Summary

- make attachment send wait for reads and staging to settle instead of allowing text to race ahead
- preserve a sanitized original display name through request DTOs, durable conversation content, and transcript rendering
- disclose whether each provider receives a workspace path, native image bytes, or both
- keep failed reads retryable/removable and preserve accepted, rejected, and uncertain delivery recovery across remounts/reloads
- prevent removed or explicitly discarded attachment work from resurrecting while preserving genuinely newer or already-persisted work

## Root cause

Attachment ownership previously ended when a pending read/stage token settled. During an asynchronous interface switch, that settlement could advance the descriptor version and remove the token before confirmed discard ran. The discard phase could then no longer prove that the current failed/undurable descriptor descended from work captured before confirmation, so the chip could reappear and later be retried or sent.

The transport also dropped the original filename and rendered capability-dependent path/native delivery with the same visual semantics.

## Fix

- await a settled attachment payload boundary before send
- carry sanitized display names through frontend, API, domain, provider, storage, and history boundaries
- render capability-derived delivery semantics accurately
- retain per-descriptor work lineage after settlement so confirmed discard removes captured descendants unless the exact current version is already locally persisted
- assign later work new ownership so post-confirmation additions/retries remain intact
- clear lineage during removal, purge, and abandonment

## Exact validation

Reviewed candidate: `925e2c117dee404954c5faad43d17114c9394e69`

Fixed point: `8729009d8e1c6b0f44b3c5d0f4ba8915c1e383a0` (#4468)

- focused composer/session attachment suites: **374/374 passed** before final review fix
- final independent review suite: **4 files / 328 tests passed**
- full renderer suite: **3,146 passed, 1 existing skip**
- frontend and E2E TypeScript checks: passed
- HTTP/spec/session-manager Go suites: passed
- OpenAPI and sqlc regeneration: zero drift
- `AttachmentInput.name`, `ConversationImageContentRequest.name`, and latest-main `lastUserMessageAt` preserved
- independent exact-SHA Standards review: **APPROVED**
- independent exact-SHA Spec review: **APPROVED**
- worktree clean; `git diff --check` passed

## Evidence

No screenshots, reports, traces, local AO state, or QA artifacts are committed. The final native Electron/MQA-10 run will use the local `todo-app` fixture; only GitHub-hosted attachments will be added here and to #4427 before this PR leaves draft.

## Scope

Product implementation and focused automated regressions only. The opt-in local ChatUI regression harness remains isolated in draft PR #4419 and never runs automatically on CI.

## Current takeover integration

Restacked on the updated #4468 and main `ab968d5e7`. Preserves current durable draft receipts and native-byte recovery across retries/reloads, carries original filenames, awaits reads/staging, propagates failed queued-draft persistence, and rejects delivery after confirmed discard. Added regressions for discard during an awaiting send, new work after discard, and failed queued attachment writes. Frontend and E2E TypeScript checks pass; MQA-10 attachment and MQA-09 path-reference browser contracts pass (5/5). Full frontend suite passes: 313 files, 4,631 passed, 6 skipped. Full backend non-race suite, build/vet/lint, and renderer smoke (59/59) pass. All 14 hosted checks pass on `79a27a46c`, including the complete Linux race suite. The draft gate remains: a fresh native Electron lab with real npm ci, isolated data, and this exact checkout launched a healthy daemon but exposed a blank shell window with hardware and software rendering. Native attachment interaction could not be verified; browser tests do not substitute for that evidence. No production app data was used. Earlier validation entries describe prior heads, not this integration.

### DIFF
--- a/backend/internal/httpd/apispec/openapi.yaml
+++ b/backend/internal/httpd/apispec/openapi.yaml
@@ -5672,6 +5672,9 @@ components:
           type: string
         mimeType:
           type: string
+        name:
+          maxLength: 255
+          type: string
       required:
       - data
       type: object
@@ -6265,6 +6268,9 @@ components:
         data:
           type: string
         mimeType:
+          type: string
+        name:
+          maxLength: 255
           type: string
       required:
       - mimeType
--- a/backend/internal/httpd/controllers/conversations.go
+++ b/backend/internal/httpd/controllers/conversations.go
@@ -589,18 +589,21 @@ func conversationContent(req SendConversationMessageRequest) ([]ports.ChatConten
 	spawnInputs := make([]AttachmentInput, 0, len(req.Attachments))
 	for _, attachment := range req.Attachments {
 		spawnInputs = append(spawnInputs, AttachmentInput{
+			Name:     attachment.Name,
 			MimeType: attachment.MIMEType,
 			Data:     attachment.Data,
 		})
 	}
-	if _, err := decodeSpawnAttachments(spawnInputs); err != nil {
+	decoded, err := decodeSpawnAttachments(spawnInputs)
+	if err != nil {
 		return nil, err
 	}
 	content := make([]ports.ChatContent, 0, len(req.Attachments)+len(req.Resources))
-	for _, attachment := range req.Attachments {
+	for i, attachment := range req.Attachments {
 		content = append(content, ports.ChatContent{
 			Type: "image", Data: attachment.Data,
 			MIMEType: strings.ToLower(strings.TrimSpace(attachment.MIMEType)),
+			Name:     decoded[i].Name,
 		})
 	}
 	for _, resource := range req.Resources {
--- a/backend/internal/httpd/controllers/conversations_test.go
+++ b/backend/internal/httpd/controllers/conversations_test.go
@@ -231,7 +231,7 @@ func TestSendConversationPreservesNativeImageAndResourceContent(t *testing.T) {
 	body := []byte(`{
 		"text":"inspect this",
 		"clientMessageId":"message-1",
-		"attachments":[{"mimeType":"image/png","data":"aW1hZ2U="}],
+		"attachments":[{"name":"folder\\\\diagram.png","mimeType":"image/png","data":"aW1hZ2U="}],
 		"resources":[
 			{"uri":"file:///repo/README.md","name":"README.md"},
 			{"uri":"file:///repo/notes.txt","name":"notes.txt","mimeType":"text/plain","text":"notes"}
@@ -257,6 +257,9 @@ func TestSendConversationPreservesNativeImageAndResourceContent(t *testing.T) {
 	}
 	if service.sent.Content[0].Type != "image" || service.sent.Content[1].Type != "resource_link" || service.sent.Content[2].Type != "resource" {
 		t.Fatalf("content = %#v", service.sent.Content)
+	}
+	if service.sent.Content[0].Name != "diagram.png" {
+		t.Fatalf("native image name = %q, want diagram.png", service.sent.Content[0].Name)
 	}
 }
 
--- a/backend/internal/httpd/controllers/dto.go
+++ b/backend/internal/httpd/controllers/dto.go
@@ -223,6 +223,9 @@ type SpawnSessionRequest struct {
 // AttachmentInput is one file attached to a spawn, delegate, stage, or send
 // request.
 type AttachmentInput struct {
+	// Name is the original client-side filename for display. The daemon treats it
+	// as untrusted, strips any path, and normalizes it before durable use.
+	Name string `json:"name,omitempty" maxLength:"255"`
 	// MimeType is the browser-reported content type (e.g. "image/png"). Used to
 	// derive the on-disk file extension. Explicitly blocked types are rejected.
 	MimeType string `json:"mimeType,omitempty"`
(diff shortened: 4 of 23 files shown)


Pulkit7070:
On the draft as it stands: the lineage and discard-race mechanism in `useFileAttachments.ts` holds up. The descriptor work-token ownership plus descriptor versioning let `discardCapturedUnpersistedAttachments` distinguish "this captured work settled unchanged" from "new work claimed this id after capture," and the three interleavings it targets (unmount before settle, settle after capture before discard, retry after capture) each have a dedicated regression test constructing that exact ordering. The backend sanitization covers path traversal, control bytes, malformed SVG parameters, and overlong subtypes. The chip markup uses existing semantic tokens, and staging stays eager so the new settled-payload boundary only adds a wait when a read or stage is still in flight at send time.

Two items before this leaves draft. `sanitizeAttachmentDisplayName` (`backend/internal/httpd/controllers/sessions.go`) keeps only ASCII `[a-zA-Z0-9_-]`, so a purely non-ASCII filename such as a CJK or accented name collapses to the `file.png` fallback, and two such attachments in one session both render as `file.png`-style collisions in the transcript. The on-disk name keeps its unique id prefix, so this is not a security issue, but it is a provenance regression against accurate per-attachment display and there is no unicode-only test case. Separately, the hook now returns `hasAttachments`, `currentAttachments`, and `signature`, none of which have a caller in the renderer tree; either wire them up or drop them.

Verified at the PR head: attachment-specific backend tests pass (`go test -run Attachment` for `internal/httpd/controllers` and `internal/session_manager`), and `useFileAttachments.test.ts` passes in full along with `ChatWorkspace.test.tsx` at 128/128 in isolation. A combined vitest run showed 28 failures confined to `ChatWorkspace.test.tsx`, all rendering an empty body, which is the resource-contention signature rather than an attachment-logic failure; I did not diff those against the base to prove it, so re-running that file alone before merge is worth doing.

PR #4543, opened 2026-08-27
fix: propagate CI injection toggle to session PRs

neversettle17-101:
## Summary
- propagate the session CI auto-injection toggle to PR rows already owned by the session
- emit PR CDC updates when autoInjectCI changes so session PR summaries refresh
- update API/schema/type/i18n wording so the toggle is no longer described as only a new-PR default

## Tests
- go test ./internal/storage/sqlite/... ./internal/lifecycle ./internal/service/session ./internal/httpd/controllers ./internal/httpd/apispec/...
- npm run sqlc
- npm run api:spec
- npx openapi-typescript@7.4.4 backend/internal/httpd/apispec/openapi.yaml -o frontend/src/api/schema.ts

## Verification caveats
- npm run frontend:typecheck could not start in this worktree because tsc is missing from frontend/node_modules
- broad backend go test ./... still has unrelated environment-sensitive failures around missing configured bundled tmux plus existing tmux/cli/session-manager/systemcheck expectations

### DIFF
--- a/backend/internal/httpd/apispec/openapi.yaml
+++ b/backend/internal/httpd/apispec/openapi.yaml
@@ -2087,7 +2087,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/APIError'
           description: Not Implemented
-      summary: Set the automatic CI-failure injection default for new session PRs
+      summary: Set automatic CI-failure injection for a session and its PRs
       tags:
       - sessions
   /api/v1/sessions/{sessionId}/auto-inject-review:
--- a/backend/internal/httpd/apispec/specgen/build.go
+++ b/backend/internal/httpd/apispec/specgen/build.go
@@ -1718,7 +1718,7 @@ func sessionOperations() []operation {
 		},
 		{
 			method: http.MethodPatch, path: "/api/v1/sessions/{sessionId}/auto-inject-ci", id: "setSessionAutoInjectCI", tag: "sessions",
-			summary:    "Set the automatic CI-failure injection default for new session PRs",
+			summary:    "Set automatic CI-failure injection for a session and its PRs",
 			pathParams: []any{controllers.SessionIDParam{}},
 			reqBody:    controllers.SetSessionAutoInjectCIRequest{},
 			resps: []respUnit{
--- a/backend/internal/httpd/controllers/dto.go
+++ b/backend/internal/httpd/controllers/dto.go
@@ -567,13 +567,13 @@ type SetSessionAutoInjectReviewResponse struct {
 	Session          SessionView      `json:"session"`
 }
 
-// SetSessionAutoInjectCIRequest updates the default automatic CI delivery
-// policy captured by PRs created after the change.
+// SetSessionAutoInjectCIRequest updates automatic CI delivery for a session
+// and every PR currently owned by it.
 type SetSessionAutoInjectCIRequest struct {
 	AutoInjectCI bool `json:"autoInjectCI"`
 }
 
-// SetSessionAutoInjectCIResponse confirms the persisted session default.
+// SetSessionAutoInjectCIResponse confirms the persisted session policy.
 type SetSessionAutoInjectCIResponse struct {
 	OK           bool             `json:"ok"`
 	SessionID    domain.SessionID `json:"sessionId"`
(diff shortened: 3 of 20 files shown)


Pulkit7070:
The `test` CI job is failing, and it's a stale test query, not a code problem:

`SessionInspector.test.tsx:666` looks for a button named `"Sends CI failures to the worker."`, but this PR reworded that description in en.json to `inspector.ci.autoInjectDescription = "Sends CI failures to the worker for this session's PRs."`. The old exact string no longer exists (0 matches in en.json), so `getByRole` throws `Unable to find an accessible element ... name "Sends CI failures to the worker."`. Update the test query to the new label (or match it with a regex) to get CI green.

Everything else checks out:
- Migration 0111 is clean, unlike the earlier CDC migration: it's a symmetric drop-and-recreate of `pr_cdc_update` (adds `auto_inject_ci` to the WHEN + `autoInjectCI` to the payload), no table rebuild or compatibility view, and 0111 is the next free version. `auto_inject_ci` is `BOOLEAN NOT NULL`, so the `<>` WHEN comparison is NULL-safe.
- `SetPRAutoInjectCIBySession` correctly updates every owned PR row, and the trigger now emits `pr_updated` on the toggle so summaries refresh.

So: fix the test label to match the reworded string and CI should go green; the fix itself is sound.

Pulkit7070 (approved):
LGTM. The failing test is fixed (SessionInspector query now matches the reworded label 'Sends CI failures to the worker for this session's PRs.' exactly, verified), and CI is fully green. The fix is sound: SetPRAutoInjectCIBySession propagates the toggle to owned PR rows, migration 0111 is a clean symmetric trigger drop/recreate (NULL-safe WHEN, next free version, no rebuild/view), and the pr_updated CDC emit refreshes summaries.

neversettle17-101 (commented):
## What This PR Does
This PR changes the session-level CI auto-injection toggle from a new-PR default into a live policy that also updates PRs already owned by the session. It adds the storage + CDC plumbing for that propagation and updates the API/UI wording to match.

## Flow
- PATCH the session `auto-inject-ci` setting.
- Persist the new session value and update all existing PR rows for that session in one transaction.
- Let the session and PR update triggers emit CDC so clients refresh their summaries.

## Summary of Changes
- Adds a SQLite migration so `pr_updated` CDC fires when `auto_inject_ci` changes.
- Adds a bulk PR update query and uses it from `SetSessionAutoInjectCI` inside a transaction.
- Updates storage tests to verify policy propagation and the extra CDC event.
- Regenerates API/schema artifacts and updates inspector/i18n copy to describe the toggle as a session policy instead of a new-PR default.

## Findings
No blocking issues found in the reviewed diff. Approval: ready to merge.

## Regression Risk
Low. The change is narrowly scoped, and the main behavior change—fanout from the session toggle to existing PR rows—is covered in focused store tests. The remaining risk is mostly around end-to-end refresh behavior, since I only reviewed the diff and did not run the app.

## Testing
Tests were not run in this review. The PR description lists focused backend test coverage plus regenerated sqlc/OpenAPI artifacts, and the diff adds storage tests plus a `SessionInspector` copy assertion update. I did not see end-to-end validation beyond those focused tests.

harshitsinghbhandari:
Blocking issue: PR takeover does not adopt the new owner session’s `autoInjectCI` policy.

This PR changes the setting from a creation-time snapshot into a live session policy and propagates changes to PRs currently owned by that session. However, `ClaimPRForSession` changes `session_id` on conflict without also setting `auto_inject_ci = excluded.auto_inject_ci` (`backend/internal/storage/sqlite/queries/pr.sql`, around lines 241–253).

Example:
1. Session A has automatic CI injection disabled and owns a PR, so the PR row stores `false`.
2. Session B has automatic CI injection enabled and takes over that PR.
3. The claim updates the PR’s `session_id` to B but leaves `auto_inject_ci` as `false`.

The result is that session B’s toggle says enabled while its PR summary and `lifecycle/reactions.go` still suppress CI-failure delivery until the setting is manually toggled again.

Please make the claim/takeover conflict path adopt the new session’s policy and add a test that transfers a PR between sessions with opposite `autoInjectCI` values. Also, `backend/internal/service/session/service.go:706-707` still describes the old snapshot semantics and should be updated.

PR #4547, opened 2026-08-27
fix(sidebar): fade message age out under the hover actions

ronishrohan:
## Summary
- The pin/kill action buttons were appearing to the left of the message-age timestamp on hover, requiring extra reserved padding and making the row feel crowded.
- The timestamp now fades out (`opacity-0`) on row hover/focus, and the action buttons overlay the same right-aligned slot instead of pushing further left.
- Reserved hover padding dropped from `78px` to `52px` since the buttons no longer need to coexist with the timestamp.
- Updated the inline-rename row so the now-absolutely-positioned timestamp doesn't overlap the rename input.

BEFORE:
https://github.com/user-attachments/assets/190297fb-57b2-4339-9e4a-14975f36bb17

AFTER:
https://github.com/user-attachments/assets/64436fbc-2495-4066-95e0-383cb965c885

## Test plan
- [x] `Sidebar.test.tsx` updated and passing (86/86)
- [ ] Manual check in the desktop app: hover a session row and confirm the timestamp fades out as the pin/kill buttons fade in, with no layout shift

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -575,16 +575,16 @@ describe("Sidebar", () => {
 		const actionButtons = screen.getByLabelText("Pin session").parentElement;
 		const time = actions?.querySelector("time");
 
-		expect(openSession).toHaveClass("pr-[34px]");
+		expect(openSession).toHaveClass("pr-[36px]");
 		expect(openSession).toHaveClass(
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

somewherelostt (approved):

PR #4548, opened 2026-08-27
fix(updates): preserve selected-feed intent across staged installs

### LINKED ISSUE #4537: bug(updates): preserve selected-feed intent across staged installs and concurrent saves
## Bug

The update-settings redesign merged in #4536 correctly separates the installed build from the feed selected for future updates, but several state-boundary gaps can make the panel display or persist a result that does not match the user's latest choice.

The release-blocking case is a channel switch while a package from the previous feed is already staged: AO can restore and install that old package after the user selects a different feed.

**Source:** post-merge review of #4536  
**Analyzed against:** merge commit `6a9da5de0e1f11a0988d79a182b588d8ba26dc2d` / PR head `f3c7585e097b4feaa222f41ca1ac360514f82196`  
**AO version:** not applicable; static review of the merged desktop code  
**Priority:** P1 — leaving an update channel can still install a package from that channel on restart or quit  
**Confidence:** High — the event, persistence, and install-on-quit paths were traced end to end

### Reproduction

1. Run a Nightly build with automatic updates enabled.
2. Let a newer Nightly download, but do not restart; updater status is `downloaded`.
3. Open Settings → Updates and change the selected feed to Stable.
4. Let the Stable check return `not-available` or an error.
5. Observe that AO restores the previously staged package and offers `Restart & install` again.
6. Quit or restart the app.

### Expected

Selecting Stable prevents a package staged from Nightly from being presented or installed as the completion of the Stable switch. The UI should either replace/invalidate the old package or explicitly keep it separate from the selected-feed transition.

### Actual

The selected feed is Stable, but the previously staged Nightly remains actionable and can install on restart or quit.

### Root cause

- `frontend/src/renderer/components/settings/UpdatesSection.tsx:148-180` saves the new channel and checks the newly selected feed.
- `UpdatesSection.tsx:81-84` clears `channelSwitch` when that owned check reports `not-available`, `error`, or `unsupported`.
- `frontend/src/main/auto-updater.ts:638-657` deliberately restores any previously staged package after a check reports the same package or no newer package. Staged state records version/request metadata, but not the feed that produced it.
- `auto-updater.ts:1131-1140` keeps `autoInstallOnAppQuit` enabled whenever the installation location permits it.
- `UpdatesSection.tsx:465-486` shows `Restart & install` for restored `downloaded` state and hides the manual Check action.

The restored state therefore proves only that *a* package is staged, not that it belongs to the newly selected feed.

## 2. Independent controls can lose the latest combined preference

Both Automatic Updates and Updates Channel remain separately operable, but each mutation writes a complete `UpdateSettings` snapshot.

A deterministic interleaving is:

1. Save A starts with `{ enabled: false, channel: "latest" }`.
2. Before A resolves, the user selects Nightly; save B starts with `{ enabled: false, channel: "nightly" }`.
3. A resolves first. The shared mutation callback applies A to the form, clears `savingField`, and invalidates the query while B is still pending.
4. The user toggles automatic updates on from that resurrected state, producing save C with `{ enabled: true, channel: "latest" }`.
5. Main-process serialization correctly writes A, B, then C, leaving the final persisted channel as Stable even though Nightly was the newer channel choice.

Relevant code:

- `UpdatesSection.tsx:117-132` applies every whole-object mutation result through the same `onSuccess`/`onError` callbacks.
- `UpdatesSection.tsx:141-180` constructs full snapshots for both controls.
- `UpdatesSection.tsx:258-274` disables only the field named by the single `savingField`, allowing the other full-object save to overlap.

Errors have a related problem: `onError` restores a complete snapshot from query-cache data, which can roll back an unrelated successful field change.

## 3. Installed-channel badge asserts Stable without evidence

`installedUpdateChannel()` at `UpdatesSection.tsx:519-520` maps every undefined or non-Nightly version to `latest`, and lines 401-403 always render that as Stable.

Consequences:

- Before `app.getVersion()` resolves, the panel briefly renders `… · Stable`; a Nightly install then flips to Nightly.
- A supported feature build such as `1.4.0-pr4536.202608271030` is labeled Stable while the same panel says `You are on PR #4536's build`.

The classifier needs Stable, Nightly, Feature, and loading/unknown outcomes, or the badge should remain hidden until the version is known.

## 4. Loading transition regresses to a blank panel

The PR removes the Updates skeleton and changes the lazy section to `Suspense fallback={null}` at `frontend/src/renderer/components/GlobalSettingsForm.tsx:60-63`.

On an uncached or slow renderer chunk, navigating to Updates now produces an empty page instead of a loading transition. This conflicts with #4536's stated goal to improve loading transitions.

## 5. E2E assertion is internally contradictory

`frontend/e2e/updates-settings.spec.ts` now:

- expects `Check for updates` to be hidden at line 24; then
- obtains that hidden button's `boundingBox()` at line 33 and requires it to be non-null at lines 35-36.

A hidden Playwright locator has no bounding box, so this test cannot pass when executed. The merged checks ran unit/type/smoke coverage but did not expose this contradiction.

## Proposed fix direction

1. Record staged-package feed provenance, or invalidate/replace a staged package when the selected feed changes. Install-on-quit must not apply a package from a feed the user just left.
2. Serialize renderer preference saves or maintain a versioned latest-intent snapshot. Ignore stale mutation/query completions and scope rollback to the mutation that still owns the latest intent.
3. Classify the installed build as Stable, Nightly, Feature, or Unknown/loading.
4. Restore a non-blank, reduced-motion-safe Suspense fallback.
5. Correct the E2E action-layout assertion.

## Acceptance criteria

- [ ] Switching feeds cannot install a package staged under the previous feed on restart or quit.
- [ ] The `downloaded` UI states whether its package belongs to the selected feed.
- [ ] Overlapping toggle/channel saves always preserve the newest choice for both fields.
- [ ] A failed stale mutation cannot roll back a newer successful field change.
- [ ] Nightly, Stable, Feature, and unresolved installed versions render truthfully.
- [ ] An uncached Updates chunk displays a loading state rather than a blank panel.
- [ ] E2E executes the downloaded-action layout path without contradictory visibility/bounding-box assertions.

## Regression tests

- Main updater: stage Nightly → switch to Stable → target check reports `not-available` → prove the Nightly package is neither presented nor installed as the Stable transition.
- Renderer: hold two `updateSettings.set` promises, start both field changes, resolve the older mutation first, perform a third interaction, then verify the final UI and final persisted object retain the newest combined intent.
- Renderer: deferred `getVersion()`, a Nightly version, and a `-pr<N>.` feature version.
- Playwright: assert geometry only for actions that are expected to be visible.

## Related

- #4536 — update-settings redesign that exposed these follow-up boundaries
- #2708 — replacing an older staged build with a newer build on the same feed; related staged-state work, but not a duplicate of cross-feed provenance
- #2377 — original restart-to-update row and channel-identity design
- #2366 — feature-build channel and installed-build identity specification
- #3938 — earlier loading-feedback problem addressed by #4536

ronishrohan:
Closes #4537

## Summary
- Persist staged update feed provenance and invalidate staged installs when the selected feed changes.
- Preserve the newest combined Automatic Updates and channel intent across overlapping saves.
- Label downloaded updates by feed and classify installed Stable, Nightly, Feature, and unresolved versions truthfully.
- Keep the Updates panel populated during lazy loading and every updater state.
- Fix the contradictory downloaded-action E2E geometry assertion.

## Validation
- 95 focused updater/settings tests passing.
- E2E TypeScript check passing.
- git diff --check passing.
- Full repository tests and live Playwright execution remain environment-blocked by unrelated workspace failures and sandbox port permissions.

### DIFF
--- a/frontend/e2e/updates-settings.spec.ts
+++ b/frontend/e2e/updates-settings.spec.ts
@@ -17,7 +17,7 @@ test("downloaded update keeps the full version readable and actions aligned", as
 	await page.goto("/#/settings");
 	await page.getByRole("button", { name: "Updates" }).click();
 
-	await expect(page.locator("#update-status-line")).toContainText("Downloaded. Restart to finish updating.");
+	await expect(page.locator("#update-status-line")).toContainText("Downloaded from Nightly (Pre-release). Restart to finish updating.");
 	const version = page.getByTestId("app-version");
 	await expect(version).toContainText("v0.12.7-nightly.202608240525");
 	await expect(page.getByRole("button", { name: "Restart & install" })).toBeVisible();
--- a/frontend/e2e/updates-settings.spec.ts
+++ b/frontend/e2e/updates-settings.spec.ts
@@ -30,8 +30,6 @@ test("downloaded update keeps the full version readable and actions aligned", as
 	expect(lineCount).toBe(1);
 
 	const restartBox = await page.getByRole("button", { name: "Restart & install" }).boundingBox();
-	const checkBox = await page.getByRole("button", { name: "Check for updates" }).boundingBox();
 	expect(restartBox).not.toBeNull();
-	expect(checkBox).not.toBeNull();
-	expect(Math.abs((restartBox?.height ?? 0) - (checkBox?.height ?? 0))).toBeLessThan(1);
+	expect(restartBox?.height).toBeGreaterThan(0);
 });
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -188,6 +188,7 @@ describe("startAutoUpdates", () => {
   const stateDir = "/tmp/ao-state";
 
   afterEach(() => {
+    rmSync(stateDir, { recursive: true, force: true });
     vi.useRealTimers();
     vi.restoreAllMocks();
     vi.unstubAllGlobals();
--- a/frontend/src/main/auto-updater.test.ts
+++ b/frontend/src/main/auto-updater.test.ts
@@ -826,8 +826,9 @@ describe("startAutoUpdates", () => {
     });
 
     const startPromise = module.startAutoUpdates(stateDir);
-    await Promise.resolve();
-    await Promise.resolve();
+    await vi.waitFor(() => {
+      expect(updaterEvents.get("error")).toBeTypeOf("function");
+    });
     let startSettled = false;
     void startPromise.then(() => {
       startSettled = true;
--- a/frontend/src/main/auto-updater.ts
+++ b/frontend/src/main/auto-updater.ts
@@ -1,7 +1,7 @@
 import { autoUpdater } from "electron-updater";
 import { app, BrowserWindow, dialog } from "electron";
 import { accessSync, constants as fsConstants, existsSync } from "node:fs";
-import { readFile } from "node:fs/promises";
+import { mkdir, readFile, unlink, writeFile } from "node:fs/promises";
 import path from "node:path";
 import semver from "semver";
 import {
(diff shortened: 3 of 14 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4492, opened 2026-08-26
chore: retire public desktop publisher

harshitsinghbhandari:
## Summary

- remove the obsolete public `frontend-release.yml` publisher so public tag pushes and manual dispatches cannot create canonical desktop releases
- rewrite the desktop release documentation around the private `Untrivial-ai/ao-releases` conductor for stable, nightly, and preview channels
- extend the established workflow tests to guard the public/private release boundary and mark stale release documentation as historical

`build-artifacts.yml` remains intentionally dispatchable: it has read-only permissions and produces only unsigned, short-lived workflow artifacts plus digests for the private conductor. It does not sign artifacts or create, edit, or publish releases.

## Validation

- release-boundary assertions run directly with Node across all public workflows
- `node --test scripts/ao-e2e-pod-gate.test.mjs` (17 passing)
- `git diff --check`

## Local environment limitation

- `npm ci` is blocked by the local npm `EALLOWGIT` policy for Electron's pinned `@electron/node-gyp` git dependency, so Vitest/typecheck could not run locally; CI has the normal dependency environment and runs the updated established test.

## Publication safety

No release was published, no release tag was pushed, no release workflow was dispatched, no environment was approved, and no existing release was mutated.

PR #4550, opened 2026-08-27
fix(installer): wire Windows code signing into the NSIS maker (#4502)

### LINKED ISSUE #4502: bug(installer): Windows builds are entirely unsigned (SmartScreen first-run block)
## Summary
Windows installers ship entirely unsigned, so Defender SmartScreen shows "unknown publisher" on first run and keeps warning until reputation accrues (slow-to-never for unsigned binaries). At 10k users this is a large first-run drop-off.

## Root cause
`frontend/makers/maker-nsis.ts:39-80` has no `certificateFile`/`signtool`/Azure Trusted Signing/sign hook, and neither CI workflow carries Windows signing secrets. Target is NSIS only.

## Expected
Sign the Windows installer (EV certificate or Azure Trusted Signing).

## Suggested fix
Add Windows code signing to the NSIS maker plus the release workflow. Bulk of the effort is certificate procurement.

## Notes
UNTESTED (needs Windows first-run); verified from config. QA audit finding INST-2. Effort: days (mostly procurement). Scaling risk: critical.

ApexYash11:
﻿## What was changed

- `frontend/makers/maker-nsis.ts`: the NSIS maker now activates electron-builder Windows code signing **only when Windows signing credentials are present in the environment**, and sets `forceCodeSigning` so a silently-unsigned artifact fails the build instead of shipping. Also adds a `win` config passthrough for explicit electron-builder `win` option overrides.
- `frontend/makers/maker-nsis.test.ts`: signing tests for every credential path, `forceCodeSigning` activation, the unsigned-by-default guarantee, and a regression guard that a stray macOS `CSC_LINK` never enables Windows signing.
- `frontend/docs/desktop-release.md`: documents the Windows signing boundary in the release architecture doc.

## Why it was changed

Fixes #4502: Windows installers shipped entirely unsigned, so Defender SmartScreen showed "unknown publisher" on first run and kept warning until reputation accrued (slow-to-never for unsigned binaries) — a large first-run drop-off at scale.

Note: the issue's "plus the release workflow" half is **intentionally omitted**. Since #4492 ("retire public desktop publisher") this repository holds no signing secrets and produces only unsigned artifacts for the private `Untrivial-ai/ao-releases` conductor, which owns all signing and publication. `frontend-release.yml` no longer exists and `frontend/release-workflows.test.ts` enforces that it stays gone. The maker-level env-activated signing added here is the interface the conductor (or any local signed build) consumes downstream.

## How it is implemented

Mirrors the env-activation pattern `forge.config.ts` already uses for macOS (`osxSign` activates only when Apple env vars are present). Three mutually exclusive credential paths in `envSigningOptions()`, first match wins:

1. `WIN_CSC_LINK` + `WIN_CSC_KEY_PASSWORD`: a code-signing .pfx (path, URL, or base64 per electron-builder `certificateFile` handling). Deliberately **not** the macOS `CSC_LINK` secret (the Apple Developer ID .p12): a fallback to it would let a stray Apple credential on a Windows build flip `forceCodeSigning` on and wedge the build with a wrong-cert failure — per review, Windows signing requires the explicit per-platform variable.
2. `WIN_CERT_SUBJECT_NAME`: a certificate already installed in the runner's Windows certificate store (the path for non-exportable EV tokens)
3. `AZURE_PUBLISHER_NAME` + `AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET`/...: Azure Trusted Signing (electron-builder consumes the `AZURE_*` credential env vars itself; `publisherName` must be explicit, so its presence selects this path)

With credentials set, the signing block is deep-merged over any explicit `cfg.win` override and `forceCodeSigning: true` makes an unsigned result fail the build (same "prove the seal" stance as the macOS dmg verification). With no credentials, behavior is byte-for-byte identical to today — public/fork builds stay unsigned.

## Tests/checks performed

- `frontend/makers/maker-nsis.test.ts`: **8/8 passed** (3 existing + 5 new signing tests)
- `frontend/release-workflows.test.ts` (the #4492 architecture guard): **6/6 passed**
- `npm run typecheck` (`tsc --noEmit`): clean
- Rebased onto current `main`; only the three files above are touched

### DIFF
--- a/frontend/makers/maker-nsis.ts
+++ b/frontend/makers/maker-nsis.ts
@@ -39,9 +39,13 @@ export type MakerNSISConfig = {
 // installer ships unsigned and Defender SmartScreen blocks first run with
 // "unknown publisher"). Three credential paths, first match wins:
 //
-//  1. WIN_CSC_LINK (falling back to CSC_LINK for local convenience) (+ the
-//     matching key password): a code-signing .pfx — local path, URL,
-//     or base64, per electron-builder's certificateFile handling.
+//  1. WIN_CSC_LINK (+ WIN_CSC_KEY_PASSWORD): a code-signing .pfx — local
+//     path, URL, or base64, per electron-builder's certificateFile handling.
+//     Deliberately NOT CSC_LINK: that secret is the macOS Apple Developer ID
+//     .p12, and falling back to it would let a stray Apple credential on a
+//     Windows build wedge signing (with forceCodeSigning below, a wrong-cert
+//     misconfiguration becomes a hard failure instead of a clean unsigned
+//     build). Per-platform signing must be opted into explicitly.
 //  2. WIN_CERT_SUBJECT_NAME: a certificate already installed in the runner's
 //     Windows certificate store — the standard path for non-exportable EV
 //     tokens, where no .pfx can exist.
@@ -55,11 +59,10 @@ export type MakerNSISConfig = {
 // Returns undefined when nothing is set — local and fork/dev builds stay
 // unsigned and keep working exactly as before.
 function envSigningOptions(): Record<string, unknown> | undefined {
-	// WIN_CSC_LINK wins so the release workflow can point at the Windows
-	// certificate without colliding with the macOS CSC_LINK secret (which is
-	// the Apple Developer ID .p12 — the wrong cert for Windows).
-	const certFile = process.env.WIN_CSC_LINK ?? process.env.CSC_LINK;
-	const certPassword = process.env.WIN_CSC_KEY_PASSWORD ?? process.env.CSC_KEY_PASSWORD;
+	// WIN_CSC_LINK only — never the macOS CSC_LINK secret (the Apple Developer
+	// ID .p12, the wrong cert for Windows). See the comment above.
+	const certFile = process.env.WIN_CSC_LINK;
+	const certPassword = process.env.WIN_CSC_KEY_PASSWORD;
 	if (certFile) {
 		const signtoolOptions: Record<string, unknown> = {
 			certificateFile: certFile,
--- a/frontend/makers/maker-nsis.ts
+++ b/frontend/makers/maker-nsis.ts
@@ -26,7 +26,74 @@ export type MakerNSISConfig = {
 	icon?: string;
 	// Any extra electron-builder `nsis` options, merged over our defaults.
 	nsis?: Record<string, unknown>;
+	// Any extra electron-builder `win` options (e.g. a hard-coded
+	// signtoolOptions/azureSignOptions override), merged over our defaults.
+	// Env-derived signing (see envSigningOptions below) still wins on the
+	// keys it sets, so CI credentials cannot be silenced by a stale value.
+	win?: Record<string, unknown>;
 };
+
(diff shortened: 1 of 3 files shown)


frontend/makers/maker-nsis.ts
```diff
@@ -26,8 +26,75 @@ export type MakerNSISConfig = {
...
+//     vars; publisherName must be supplied explicitly, so its presence is
+//     what marks this path as selected.
+//
+// Returns undefined when nothing is set — local and fork/dev builds stay
+// unsigned and keep working exactly as before.
+function envSigningOptions(): Record<string, unknown> | undefined {
+	// WIN_CSC_LINK wins so the release workflow can point at the Windows
+	// certificate without colliding with the macOS CSC_LINK secret (which is
+	// the Apple Developer ID .p12 — the wrong cert for Windows).
+	const certFile = process.env.WIN_CSC_LINK ?? process.env.CSC_LINK;
```
Pulkit7070: The design is sound - `forceCodeSigning` is correctly gated inside `if (signing)`, so credential-less dev/fork/CI builds stay unsigned and keep working, and env signing merges over `cfg.win` per key. Two things worth addressing:

[medium] The `?? process.env.CSC_LINK` fallback is risky now that `forceCodeSigning` is flipped on whenever `certFile` is set. `CSC_LINK` is the macOS Apple Developer ID cert (the comment says so), so on any Windows build environment that has `CSC_LINK` set but not `WIN_CSC_LINK`, this resolves the Apple .p12 as the Windows cert, sets `forceCodeSigning: true`, and the Windows build then fails trying to sign with the wrong cert - where before this PR it shipped unsigned and passed. It's latent until the release workflow (the noted follow-up) wires the env, but when it does, the workflow must be careful never to expose the macOS `CSC_LINK` to the Windows job. Safer to require `WIN_CSC_LINK` on Windows and drop the `CSC_LINK` fallback, so a stray Apple secret can't wedge the Windows build.

[low] `envSigningOptions` (three credential paths, their precedence, and this fallback) has no test, even though `maker-nsis.test.ts` already exists. A couple of cases (WIN_CSC_LINK wins over CSC_LINK, cert-store path, Azure path, and undefined when nothing is set) would lock the precedence and catch the fallback interaction above.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Pulkit7070:
This is solid: the credential precedence, choosing `WIN_CSC_LINK` over the macOS `CSC_LINK`, and flipping `forceCodeSigning` so a misconfiguration fails the build rather than shipping unsigned are all correct, and the keys match app-builder-lib v26.

One minor robustness point: `WIN_SIGNING_HASH_ALGORITHMS` is split on "," with no trimming, so `"sha256, sha1"` yields a `" sha1"` element electron-builder will not recognize. Trimming each entry would make it forgiving of natural spacing.

ApexYash11:
@Pulkit7070  Re-review!!

Pulkit7070 (approved):

Pulkit7070 (approved):

PR #4551, opened 2026-08-27
fix: honor reviewer hook activity for review live state

### LINKED ISSUE #4286: Use review agent hooks to update review status

neversettle17-101:
Closes #4286

## Summary
- persist reviewer hook activity state on review rows and expose it on the reviews API
- drive the inspector's live review strip from reviewer activity instead of any running review row
- add backend and frontend tests for idle, active, blocked, and invalid reviewer activity states

## Tests
- `cd backend && go test ./internal/service/review ./internal/review ./internal/httpd/controllers ./internal/storage/sqlite/store`

## Known limits
- `npm run api` needed a fallback `npx openapi-typescript@7.4.4 ...` because `openapi-typescript` was not on PATH in this environment
- frontend `vitest` / `typecheck` could not be executed here because local JS toolchain dependencies were missing (`vitest/config`, `tsc`)


neversettle17-101 (commented):
Ready to merge.

## What This PR Does
Persists reviewer hook activity state on review rows, exposes it from the reviews API, and uses that state in the inspector to hide the live progress strip when a running review session is idle or waiting. It also updates the generated API types/spec and adds backend/frontend coverage for activity-state persistence and display behavior.

## Summary of Changes
- Adds reviewer_activity_state to the review table via migration and threads it through domain, store, service, controller, OpenAPI, and frontend types.
- Validates review activity state values on POST /reviews/{reviewSessionID}/activity and preserves metadata-only native session updates.
- Updates the SessionInspector live review indicator to require a reviewer handle plus non-idle reviewer activity while keeping old-daemon fallback behavior.
- Adds backend controller/service/store tests and frontend helper/component tests for active, idle, blocked, and invalid state cases.

## Findings
No blocking findings from the reviewed diff.

## Regression Risk
Low to medium: the change touches a persisted schema and generated API contract, but the migration is additive and existing responses keep the new field optional. The main residual risk is correctness of live-state transitions across real reviewer hooks, which depends on runtime behavior outside the unit tests.

## Testing
The PR reports backend package tests. The diff adds targeted tests for API validation/persistence, store round-tripping, and frontend live-strip behavior. I did not run tests as part of this review.

backend/internal/storage/sqlite/migrations/0134_review_activity_state.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE review ADD COLUMN reviewer_activity_state TEXT NOT NULL DEFAULT '';
```
nikhilachale, line 3: **[P1] Renumber this migration after rebasing.** Current `main` already owns migration versions `0117` (`0117_allow_kimi_usage.sql`) and `0118` (`0118_cancelled_conversation_turns.sql`). Keeping this as `0117` gives Goose duplicate migration versions. Please rebase, rename it to the next available version, and update the burned-version map.

backend/internal/review/review.go
```diff
@@ -389,7 +390,7 @@ func (e *Engine) TriggerWithSource(ctx stdctx.Context, workerID domain.SessionID
 			return TriggerResult{}, failRuns(0, fmt.Errorf("notify reviewer: %w", err))
 		}
 	}
-	reviewRow, err = e.upsertReview(ctx, worker, harness, handleID, reviewRow.AgentSessionID, now)
+	reviewRow, err = e.upsertReview(ctx, worker, harness, handleID, reviewRow.AgentSessionID, domain.ActivityActive, now)
```
nikhilachale: **[P2] Preserve the latest hook-owned activity state here.** A fast reviewer can emit `stop -> idle` while `Spawn`/`Notify` is completing; this unconditional post-launch `active` upsert then overwrites that newer hook state, leaving the inspector's "Review in progress" strip visible with no later hook to correct it. The restore finalization at line 578 has the inverse ordering problem. Please update only handle/session metadata here (or otherwise preserve the latest persisted activity), and add an interleaving regression test.

backend/internal/httpd/controllers/reviews.go
```diff
@@ -122,13 +123,25 @@ func (c *ReviewsController) activity(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
 	agentSessionID := capActivityMeta(domain.SanitizeControlChars(strings.TrimSpace(in.AgentSessionID)))
-	if strings.TrimSpace(in.State) == "" && agentSessionID == "" {
-		envelope.WriteAPIError(w, r, http.StatusBadRequest, "bad_request", "REVIEW_ACTIVITY_OR_SESSION_ID_REQUIRED", "Reviewer activity state or agent session ID is required", nil)
+	if state == "" && agentSessionID == "" {
+		envelope.WriteJSON(w, http.StatusOK, SetReviewActivityResponse{OK: true, ReviewSessionID: reviewSessionID})
 		return
 	}
 	if err := c.Svc.ApplyReviewActivitySignal(r.Context(), reviewSessionID, reviewsvc.ActivitySignal{
```
nikhilachale, line 142: **[P2] Fence reviewer activity by process generation.** `SetReviewActivityRequest` already carries `launchId`, but it is dropped when building this signal, and reviewer rows are reused across launches. A delayed `idle`/`exited` hook from a replaced reviewer can therefore overwrite the replacement's `active` state and hide the live strip while it is working. Please assign/persist a reviewer launch generation, pass `LaunchID` through this controller/service path, and ignore non-matching generations, with a stale-generation test.

nikhilachale (commented):
Inline findings from the review. These state-ordering issues and the migration conflict should be addressed before merge.

backend/internal/storage/sqlite/migrations/0134_review_activity_state.sql
nikhilachale, line 3: **[P1] Renumber both migrations after rebasing again.** Current `main` now owns `0121_session_reviewer_agent_config.sql` and `0122_drop_agent_inventory_cache.sql`, so this migration and `0122_review_launch_id.sql` collide with already-merged versions. Please rebase, move them to the next free versions (currently 0123/0124), update the burned-version ledger, and regenerate/check the SQL artifacts.

backend/internal/review/review.go
```diff
@@ -376,20 +377,24 @@ func (e *Engine) TriggerWithSource(ctx stdctx.Context, workerID domain.SessionID
 		if err := e.launcher.Preflight(ctx, harness, worker.Metadata.WorkspacePath); err != nil {
 			return TriggerResult{}, failRuns(0, fmt.Errorf("reviewer preflight: %w", err))
 		}
-		launch, err := e.launcher.Spawn(ctx, reviewLaunchSpec(worker, harness, created[0], queue, 0, reviewRow.AgentSessionID))
+		launchID := e.newID()
+		launch, err := e.launcher.Spawn(ctx, reviewLaunchSpec(worker, harness, created[0], queue, 0, reviewRow.AgentSessionID, launchID))
```
nikhilachale: **[P1] Persist the replacement generation before starting it.** `L2` is passed into the new process here, but the review row continues to store `L1` until the final upsert after `Spawn` returns. Any `L2` session-start, stop, or native-session hook emitted during runtime creation/prompt readiness is therefore rejected by the SQL fence. The restore path has the same window. Please durably claim `L2` before `Spawn`/`RestoreTerminal`, preserve hook-owned fields during finalization, and add trigger/restore tests with an existing non-empty launch ID using the real fenced store.

backend/internal/storage/sqlite/queries/review.sql
```diff
@@ -30,8 +32,13 @@ UPDATE review SET reviewer_handle_id = '', updated_at = CURRENT_TIMESTAMP WHERE
...
 
--- name: UpdateReviewAgentSessionID :execrows
-UPDATE review SET agent_session_id = ?, updated_at = CURRENT_TIMESTAMP WHERE id = ?;
+-- name: UpdateReviewActivity :execrows
+UPDATE review SET
+    agent_session_id = CASE WHEN ? != '' THEN ? ELSE agent_session_id END,
+    reviewer_activity_state = CASE WHEN ? != '' THEN ? ELSE reviewer_activity_state END,
+    updated_at = CURRENT_TIMESTAMP
+WHERE id = ?
+  AND (? = '' OR reviewer_launch_id = '' OR reviewer_launch_id = ?);
```
nikhilachale: **[P2] Do not let an empty incoming launch ID bypass an established generation.** With this predicate, a delayed hook from a legacy reviewer can omit `launchId` and still overwrite the replacement reviewer's activity or native session ID even when `reviewer_launch_id` is non-empty. Accept a missing ID only while the stored generation is also empty; once a generation is claimed, require an exact match. Please add that empty-ID/non-empty-stored-ID store case.

nikhilachale (commented):
Re-review findings against the latest head. The previous post-launch activity overwrite is fixed, but these remaining issues should be addressed before merge.

neversettle17-101 (commented):
Ready to merge.

## What This PR Does

This PR persists reviewer hook activity on review rows, exposes it through the reviews API, and uses it in the inspector so the live review strip reflects reviewer activity rather than only a running review row. It also adds launch-generation fencing so delayed hooks from replaced reviewer processes cannot overwrite the current reviewer state.

## Summary of Changes

- Added review-row fields for reviewer activity state and reviewer launch ID, including SQLite migrations, sqlc query updates, store mapping, OpenAPI, and frontend generated types.
- Threaded launch IDs into reviewer spawn/restore/notify specs and reviewer hook activity updates.
- Updated review activity handling to persist known hook states, ignore unsupported states, and no-op stale/missing launch IDs once a generation is claimed.
- Updated the inspector review panel to show the live strip only when there is a reviewer handle, a running review, and active/blocked (or legacy missing) activity.
- Added backend and frontend tests for activity persistence, launch fencing, restore/trigger races, API exposure, and inspector rendering behavior.

## Findings

No blocking findings found in the reviewed diff.

## Regression Risk

Low-to-moderate: this touches review persistence, launch/restore timing, hook updates, and inspector live-state presentation. The new fencing behavior intentionally ignores legacy hook updates once a launch generation exists, so any reviewer process that fails to receive or echo AO_RUNTIME_LAUNCH_ID will stop updating activity/native-session metadata; the diff includes explicit tests for that intended behavior.

## Testing

I did not run tests. The PR includes backend tests for controller/service/store/engine behavior and frontend tests for the live review strip helper and component states. The PR body reports a backend targeted test run; frontend validation was not available in the author environment due missing local JS dependencies.

nikhilachale (approved):

PR #4552, opened 2026-08-27
fix(terminal): back-pressure PTY output instead of killing the WebSocket

### LINKED ISSUE #4499: bug(terminal): high-output flood tears down the whole terminal WebSocket (crash loop)
## Summary
A high-output command in an attached terminal (a build, `yes`, a verbose test run) tears down the entire terminal WebSocket, then reattaches, replays, floods again, and dies: a crash loop.

## Root cause
`attachment.copyOut` reads 32KB chunks (`backend/internal/terminal/attachment.go:191-204`) into `connState.enqueue` (`backend/internal/terminal/manager.go:529-535`) on a 1024-frame channel (`defaultWriteBuffer`, `manager.go:36`). On overflow the `default` branch calls `c.cancel()` (`manager.go:533`), tearing down the whole connection and every pane on it. The PTY read loop never blocks, so there is no upstream backpressure; a flood fills 1024 frames faster than JSON+base64 WS writes drain.

## Expected
Coalesce/drop data frames (the client only needs the latest output), or apply flow control (stop reading the attach PTY so tmux/conpty throttles) instead of killing the connection.

## Suggested fix
On a full buffer, coalesce consecutive `msgData` frames for the same terminal id into the tail rather than cancelling; or batch multiple 32KB reads into one frame in `copyOut` as a stopgap.

## Notes
UNTESTED live (needs a running daemon+renderer); verified from code. QA audit finding TERM-1. Effort: 30-min stopgap, ~2-4h correct. Scaling risk: high.

Pyasma:
Fixes #4499.

## Approach

`enqueue` pushed every server frame onto a 1024-slot buffered channel and called `c.cancel()` when it was full, so one high-output pane tore down the whole connection.

Replace the channel with a per-connection queue that treats the two frame classes differently:

- **PTY output** waits for room above a 1MB watermark. The attachment's read loop stalls, so tmux throttles at the source.
- **Control frames** (`opened`, `exited`, `error`, `resize`, `pong`, session snapshots) never wait and are never dropped. The CDC fan-out enqueues on one goroutine shared by every connection, so blocking there would stall unrelated connections.

Ordering is unchanged: one queue, appended in production order, drained in order. A waiting producer is released by a drain or by the connection's context.

### DIFF
--- a/backend/internal/terminal/manager.go
+++ b/backend/internal/terminal/manager.go
@@ -32,8 +32,11 @@ type wsConn interface {
 }
 
 const (
-	defaultHeartbeat   = 15 * time.Second
-	defaultWriteBuffer = 1024
+	defaultHeartbeat = 15 * time.Second
+	// dataWatermark caps queued PTY output per connection. Above it the producing
+	// attachment's read loop blocks, so tmux throttles at the source instead of
+	// the connection being torn down under a flood.
+	dataWatermark = 1 << 20
 )
 
 // Manager serves WebSocket clients, opening one attach Stream per opened pane
@@ -337,8 +340,9 @@ func (m *Manager) Serve(ctx context.Context, conn wsConn) {
 	c := &connState{
 		mgr:    m,
 		conn:   conn,
+		ctx:    ctx,
 		cancel: cancel,
-		out:    make(chan serverMsg, defaultWriteBuffer),
+		out:    newOutQueue(),
 		terms:  map[string]*attachment{},
 	}
 	defer c.cleanup()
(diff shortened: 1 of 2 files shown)


Pulkit7070 (approved):

Pulkit7070 (approved):
The approach is correct. Replacing the fire-and-cancel channel with a mutex-guarded queue that back-pressures PTY data at a 1MB watermark while letting control frames through immediately is the right fix. The old behavior (any fast command kills all terminal tabs on the connection) is strictly worse than a brief output pause.

The three tests cover the key invariants: data blocks until drained, control frames never wait, blocked producer unblocks on connection close. CI is green across all platforms including Windows.

One suggestion for a follow-up: a rate-limited log line when back-pressure activates would help debug slow-client issues in production, but that's not a blocker.

PR #4553, opened 2026-08-27
fix: preserve sessions across failed restart recovery

AgentWrapper:
## Summary

- run Codex Chat compatibility probes and app-server children with the same executable-aware environment used by normal TUI launches
- preserve a live AO session, worktree, and native conversation identity when startup reconciliation cannot relaunch its controller
- fence every startup-reconciliation candidate before bounded worker queuing, then release its fence after that session's attempt finishes
- transition failed relaunches from the latest durable revision so launch-preparation metadata cannot make the exited signal a silent CAS no-op
- require a published/live Chat controller—not merely an early durable generation claim—before treating an errored Chat relaunch as committed
- recover a historical project-orchestrator handoff only when the latest completed TUI -> Chat transition proves the session's exact native provider id, then append its missing provider-ownership epoch atomically instead of rewriting old history
- cover the unmodified npm `codex.js` launcher under a Finder-style sparse `PATH` for Probe, initial Chat start, Chat resume, and restoration of the original terminated orchestrator

## Safe reproduction and root cause

I reproduced the failures with test-owned homes, stores, workspaces, runtimes, and Chat launchers; I did not mutate `~/.ao`.

### Why Chat restore failed

Inventory correctly resolves the absolute npm launcher at `.../@openai/codex/bin/codex.js`. That launcher is unmodified and starts with `#!/usr/bin/env node`.

Normal Codex **TUI** launch already passes its resolved argv through `AugmentRuntimePATHForLaunchBinary`. That adds the concrete NVM/fnm/Volta Node directory missing from a Finder-launched daemon's sparse `PATH`, which is why new terminal sessions still launch.

Codex **Chat** took a different path:

1. `PreflightChat` called `Driver.Probe`.
2. `Probe` called `installedCodexVersion` directly with the absolute `codex.js` path but the daemon's unaugmented environment.
3. `/usr/bin/env` could not find `node`, so the version command exited 127.
4. Recovery stopped at compatibility preflight. `thread/resume` was never attempted.
5. Even if preflight were bypassed, the app-server child used the same unaugmented process environment, so launch/resume retained the same defect.

The executable-aware PATH logic now lives at the shared `agentlaunch` boundary. TUI/reviewer launch keeps using it, while both `installedCodexVersion`/Probe and app-server spawn apply it before starting the unmodified npm launcher. This deliberately does **not** depend on Codex's private npm package layout or substitute its package-native payload.

### Why sessions became terminated and the orchestrator was replaced

Chat controllers are daemon children, so they necessarily need an in-place relaunch after the daemon restarts. When the environment bug made that relaunch fail, `reconcileLive` treated the transient preflight error as durable user intent: it captured work, terminated the native controller, marked the AO session terminated, and force-removed its worktree. Once the original orchestrator was terminated, the supervisor could create a replacement.

Failed startup relaunch now leaves the durable session live with exited activity. It does not stash, mark terminated, remove the worktree, or clear native IDs.

Three lifecycle races also had to be closed for that invariant to hold:

- TUI launch preparation persists a browser-capability verifier before `runtime.Create`, advancing `UpdatedAt`. A later runtime failure previously applied the exited signal with the stale boot-snapshot revision; lifecycle correctly discarded that CAS, leaving the dead session active/idle, Resume-rejected, and reaper-exposed. Recovery now reads the latest durable row, fences the same controller epoch, applies exited with that revision, rereads to verify the transition, and retries boundedly if a concurrent write wins.
- Chat Service claims a new durable `ControllerGeneration` before native-history import, `ControllerReady`, and registry publication. A history/readiness failure after that claim previously looked committed solely because the generation changed, so reconcile returned success with no live controller and active/idle activity. Recovery now requires both the changed durable epoch and `HasLiveChatController`; otherwise it records exited against the newly claimed generation so stale callbacks remain fenced.
- The reconcile fence was formerly acquired only after a worker dequeued a session. Candidates beyond `ReconcileWorkers` therefore remained open to input/reaping during slow probes. Reconciliation now reserves every available live candidate before creating the bounded work queue and releases each per-session fence after its attempt settles.

### Why a historically switched orchestrator could still fail Restore

One affected orchestrator had a second, older invariant breach: its project conversation was rebound to the current AO session and its completed TUI -> Chat transition recorded the new native conversation id, but the active root branch still named an older orchestrator/provider. Strict ordinary resume correctly rejected that unowned handle before provider launch. `RepairIncompleteConversationEdit` could not help because the active root was not an unfinished edit child.

For terminated Chat orchestrators only, Session Manager now requires all of the following durable proof before reserving a repair boundary: the latest transition is completed TUI -> Chat, its native id exactly equals the session provider id, the project conversation is still bound to that AO session, and the active provider branch belongs to an older nonempty owner. The deterministic transition-derived boundary is passed through normal Chat startup. Provider Resume must return that exact proved handle, or restore closes it and fails before history or durable ownership changes. AO then reads and reconciles stable native history without mutating the timeline. `ControllerReady -> MarkChatSpawnedPrepared -> CommitChatSpawnPrepared` rechecks the conversation owner and expected parent, stages the new branch and controller generation inside one SQLite transaction, projects the replay onto that branch, and only then publishes the live session in the same commit. Concurrent readers therefore see either the old terminated state or the complete restored transcript/controller; input is never exposed between them. The old root is never rewritten. Provider/history/projection/final-session-write failures roll the whole transaction back, leaving the session terminated and creating no boundary. If a newer orchestrator owns the project conversation, restore fails before provider I/O and does not rebind it.

## Recovery for affected preview sessions

The currently installed v0.12.0-pr4473 preview does not contain this change, so pressing Restore there will continue to fail at `read Codex version: exit status 127` before provider resume.

After installing a build containing this PR:

1. Sessions that are **live/exited** can use **Resume Agent**. They resume in place with the same AO session ID, worktree/branch, and provider conversation ID.
2. Sessions already marked **terminated** by the preview use **Restore**. Restore runs the corrected preflight and then `thread/resume`, retaining the original AO session ID and native provider conversation ID. For historically switched project orchestrators, the matching completed transition now repairs the missing ownership epoch transactionally after Resume/history succeeds. For Chat, `provider_conversation_id` is the resume handle; an empty `agent_session_id` is not a blocker.
3. The old orchestrator can be restored instead of replaced. If a replacement currently owns the canonical branch/worktree, terminate that replacement first to avoid the expected workspace collision, then restore the old orchestrator.

Focused tests cover a terminated Chat orchestrator failing compatibility once, then restoring under its original ID and provider conversation after recovery, without creating a terminal runtime. A Chat post-claim test proves a changed generation without a published controller lands exited against that claimed epoch. A production-shaped TUI test also proves a browser-verifier write followed by `runtime.Create` failure lands live/exited and succeeds through ordinary Resume Agent without losing the native conversation or worktree.

## Verification

- `npm run lint` (full `go test ./...` plus golangci-lint v2.12.2, clean environment)
- `go test -race ./internal/session_manager ./internal/lifecycle ./internal/adapters/chatdriver/codexappserver ./internal/agentlaunch ./internal/integration -count=1`
- focused failed-runtime, Chat post-generation-claim/no-controller, one-shot/persistent CAS contention, and 3-session/2-worker queued-fence regression tests
- real-SQLite historical project-owner coverage for stable user/assistant/turn history persistence on the new epoch, transcript/session/worktree continuity, exact returned-handle validation, provider/history failure, full projection/final-write rollback, unproved mismatch, stale owner/head races, generation advance, next-message continuity, and retry without a duplicate epoch
- real-store equal-`created_at` transition coverage proving the latest inserted transition wins deterministically via SQLite `rowid`

The npm regression fixture first proves the unmodified `codex.js` launcher exits 127 under the sparse daemon `PATH`. It then proves Probe/version detection succeeds with the shared launch environment, and separately asserts both initial `Start` and stored-conversation `Resume` pass the augmented Node path to app-server.

## Context

The regression was observed in the preview artifact built from #4473. The fix is in the shared child-process launch environment and restart/recovery architecture—not Electron UI, Codex package internals, or a schema migration. The sole generated change is sqlc output regenerated from the deterministic latest-transition query.

### DIFF
--- a/backend/internal/adapters/agent/codex/codex.go
+++ b/backend/internal/adapters/agent/codex/codex.go
@@ -372,7 +372,7 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 	}
 
 	if path, err := exec.LookPath("codex"); err == nil && path != "" {
-		return path, nil
+		return resolveNativeUnixCodex(path), nil
 	}
 
 	candidates := []string{
@@ -396,7 +396,7 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 
 	for _, candidate := range candidates {
 		if fileExists(candidate) {
-			return candidate, nil
+			return resolveNativeUnixCodex(candidate), nil
 		}
 		if err := ctx.Err(); err != nil {
 			return "", err
@@ -404,6 +404,38 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 	}
 
 	return "", fmt.Errorf("codex: %w", ports.ErrAgentBinaryNotFound)
+}
+
+// resolveNativeUnixCodex replaces the npm JavaScript launcher with the native
+// executable shipped in the same package when it is available. Finder-launched
+// desktop processes can inherit a PATH without the npm install's Node runtime;
+// executing the absolute JavaScript launcher then still fails in /usr/bin/env
+// with exit 127. The native binary is the same payload that launcher selects,
+// and mirrors the existing Windows shim resolution above.
+func resolveNativeUnixCodex(path string) string {
+	if runtime.GOOS == "windows" {
+		return path
+	}
+	resolved, err := filepath.EvalSymlinks(path)
+	if err != nil || filepath.Base(resolved) != "codex.js" || filepath.Base(filepath.Dir(resolved)) != "bin" {
+		return path
+	}
+	packageRoot := filepath.Dir(filepath.Dir(resolved))
+	arch := runtime.GOARCH
+	if arch == "amd64" {
+		arch = "x64"
+	}
+	packageDir := filepath.Join(packageRoot, "node_modules", "@openai", "codex-"+runtime.GOOS+"-"+arch)
+	candidates, err := filepath.Glob(filepath.Join(packageDir, "vendor", "*", "bin", "codex"))
+	if err != nil {
+		return path
+	}
+	for _, candidate := range candidates {
+		if fileExists(candidate) {
+			return candidate
+		}
+	}
+	return path
 }
 
 func resolveNativeWindowsCodex(path string) string {
--- a/backend/internal/adapters/agent/codex/codex.go
+++ b/backend/internal/adapters/agent/codex/codex.go
@@ -372,7 +372,7 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 	}
 
 	if path, err := exec.LookPath("codex"); err == nil && path != "" {
-		return resolveNativeUnixCodex(path), nil
+		return path, nil
 	}
 
 	candidates := []string{
@@ -396,7 +396,7 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 
 	for _, candidate := range candidates {
 		if fileExists(candidate) {
-			return resolveNativeUnixCodex(candidate), nil
+			return candidate, nil
 		}
 		if err := ctx.Err(); err != nil {
 			return "", err
@@ -404,38 +404,6 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 	}
 
 	return "", fmt.Errorf("codex: %w", ports.ErrAgentBinaryNotFound)
-}
-
-// resolveNativeUnixCodex replaces the npm JavaScript launcher with the native
-// executable shipped in the same package when it is available. Finder-launched
-// desktop processes can inherit a PATH without the npm install's Node runtime;
-// executing the absolute JavaScript launcher then still fails in /usr/bin/env
-// with exit 127. The native binary is the same payload that launcher selects,
-// and mirrors the existing Windows shim resolution above.
-func resolveNativeUnixCodex(path string) string {
-	if runtime.GOOS == "windows" {
-		return path
-	}
-	resolved, err := filepath.EvalSymlinks(path)
-	if err != nil || filepath.Base(resolved) != "codex.js" || filepath.Base(filepath.Dir(resolved)) != "bin" {
-		return path
-	}
-	packageRoot := filepath.Dir(filepath.Dir(resolved))
-	arch := runtime.GOARCH
-	if arch == "amd64" {
-		arch = "x64"
-	}
-	packageDir := filepath.Join(packageRoot, "node_modules", "@openai", "codex-"+runtime.GOOS+"-"+arch)
-	candidates, err := filepath.Glob(filepath.Join(packageDir, "vendor", "*", "bin", "codex"))
-	if err != nil {
-		return path
-	}
-	for _, candidate := range candidates {
-		if fileExists(candidate) {
-			return candidate
-		}
-	}
-	return path
 }
 
 func resolveNativeWindowsCodex(path string) string {
--- a/backend/internal/adapters/agent/codex/codex_test.go
+++ b/backend/internal/adapters/agent/codex/codex_test.go
@@ -3,9 +3,7 @@ package codex
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"reflect"
 	"runtime"
@@ -440,66 +438,6 @@ func TestResolveCodexBinaryFindsNVMInstallWhenPathIsSparse(t *testing.T) {
 	}
 	if got != want {
 		t.Fatalf("ResolveCodexBinary = %q, want %q", got, want)
-	}
-}
-
-func TestResolveCodexBinaryUsesNPMNativeBinaryWhenNodeIsMissingFromPath(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		t.Skip("Unix npm launcher resolution")
-	}
-	home := t.TempDir()
-	if canonical, err := filepath.EvalSymlinks(home); err == nil {
-		home = canonical
-	}
-	packageRoot := filepath.Join(home, ".local", "lib", "node_modules", "@openai", "codex")
-	launcher := filepath.Join(packageRoot, "bin", "codex.js")
-	if err := os.MkdirAll(filepath.Dir(launcher), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(launcher, []byte("#!/usr/bin/env node\n"), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	arch := runtime.GOARCH
-	if arch == "amd64" {
-		arch = "x64"
-	}
-	want := filepath.Join(packageRoot, "node_modules", "@openai", "codex-"+runtime.GOOS+"-"+arch, "vendor", "target", "bin", "codex")
-	if err := os.MkdirAll(filepath.Dir(want), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(want, []byte("#!/bin/sh\nprintf 'codex-cli 0.149.1\\n'\n"), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	shim := filepath.Join(home, ".local", "bin", "codex")
-	if err := os.MkdirAll(filepath.Dir(shim), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := os.Symlink(filepath.Join("..", "lib", "node_modules", "@openai", "codex", "bin", "codex.js"), shim); err != nil {
-		t.Fatal(err)
-	}
-	t.Setenv("HOME", home)
-	t.Setenv("PATH", t.TempDir())
-	launcherOutput, launcherErr := exec.Command(launcher, "--version").CombinedOutput()
-	var exitErr *exec.ExitError
-	if !errors.As(launcherErr, &exitErr) || exitErr.ExitCode() != 127 {
-		t.Fatalf("npm launcher without node = %q, %v; want exit 127", launcherOutput, launcherErr)
-	}
-	origFileExists := fileExists
-	fileExists = func(path string) bool {
-		return strings.HasPrefix(path, home+string(os.PathSeparator)) && origFileExists(path)
-	}
-	t.Cleanup(func() { fileExists = origFileExists })
-
-	got, err := ResolveCodexBinary(context.Background())
-	if err != nil {
-		t.Fatalf("ResolveCodexBinary: %v", err)
-	}
-	if got != want {
-		t.Fatalf("ResolveCodexBinary = %q, want native executable %q", got, want)
-	}
-	output, err := exec.Command(got, "--version").CombinedOutput()
-	if err != nil || string(output) != "codex-cli 0.149.1\n" {
-		t.Fatalf("native version probe = %q, %v", output, err)
 	}
 }
 
--- a/backend/internal/adapters/agent/codex/codex_test.go
+++ b/backend/internal/adapters/agent/codex/codex_test.go
@@ -3,7 +3,9 @@ package codex
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"runtime"
@@ -438,6 +440,66 @@ func TestResolveCodexBinaryFindsNVMInstallWhenPathIsSparse(t *testing.T) {
 	}
 	if got != want {
 		t.Fatalf("ResolveCodexBinary = %q, want %q", got, want)
+	}
+}
+
+func TestResolveCodexBinaryUsesNPMNativeBinaryWhenNodeIsMissingFromPath(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Unix npm launcher resolution")
+	}
+	home := t.TempDir()
+	if canonical, err := filepath.EvalSymlinks(home); err == nil {
+		home = canonical
+	}
+	packageRoot := filepath.Join(home, ".local", "lib", "node_modules", "@openai", "codex")
+	launcher := filepath.Join(packageRoot, "bin", "codex.js")
+	if err := os.MkdirAll(filepath.Dir(launcher), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(launcher, []byte("#!/usr/bin/env node\n"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	arch := runtime.GOARCH
+	if arch == "amd64" {
+		arch = "x64"
+	}
+	want := filepath.Join(packageRoot, "node_modules", "@openai", "codex-"+runtime.GOOS+"-"+arch, "vendor", "target", "bin", "codex")
+	if err := os.MkdirAll(filepath.Dir(want), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(want, []byte("#!/bin/sh\nprintf 'codex-cli 0.149.1\\n'\n"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	shim := filepath.Join(home, ".local", "bin", "codex")
+	if err := os.MkdirAll(filepath.Dir(shim), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.Symlink(filepath.Join("..", "lib", "node_modules", "@openai", "codex", "bin", "codex.js"), shim); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("HOME", home)
+	t.Setenv("PATH", t.TempDir())
+	launcherOutput, launcherErr := exec.Command(launcher, "--version").CombinedOutput()
+	var exitErr *exec.ExitError
+	if !errors.As(launcherErr, &exitErr) || exitErr.ExitCode() != 127 {
+		t.Fatalf("npm launcher without node = %q, %v; want exit 127", launcherOutput, launcherErr)
+	}
+	origFileExists := fileExists
+	fileExists = func(path string) bool {
+		return strings.HasPrefix(path, home+string(os.PathSeparator)) && origFileExists(path)
+	}
+	t.Cleanup(func() { fileExists = origFileExists })
+
+	got, err := ResolveCodexBinary(context.Background())
+	if err != nil {
+		t.Fatalf("ResolveCodexBinary: %v", err)
+	}
+	if got != want {
+		t.Fatalf("ResolveCodexBinary = %q, want native executable %q", got, want)
+	}
+	output, err := exec.Command(got, "--version").CombinedOutput()
+	if err != nil || string(output) != "codex-cli 0.149.1\n" {
+		t.Fatalf("native version probe = %q, %v", output, err)
 	}
 }
 
--- a/backend/internal/adapters/chatdriver/codexappserver/driver.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/driver.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log/slog"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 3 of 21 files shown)


AgentWrapper:
I found two blocking lifecycle races in the revised recovery path.

**[High] Failed TUI relaunches usually cannot be transitioned to `exited`** — `backend/internal/session_manager/manager.go:2300-2305`

`ExpectedUpdatedAt` comes from the `SessionRecord` captured by the initial `ListAllSessions`. In production, a TUI relaunch calls `launchRuntimeEnv` and then `persistBrowserCapabilityVerifier` (`manager.go:1982-1988`, `3841-3850`) before runtime creation. That persistence advances the session revision. If any later step fails (for example `runtime.Create`), `liveRelaunchCommitted` correctly determines that no launch committed, but this `ApplyActivitySignal` is then silently discarded by the lifecycle CAS (`lifecycle/manager.go:542-545`) because its expected revision is stale. The row remains `active`/`idle`, so `ResumeAgentWithMode` rejects a TUI retry, and once the reconcile fence is released the periodic reaper can still terminate the stale dead-runtime row. This defeats the new live/exited recovery invariant. I reproduced it at this head with production-shaped `BrowserCapabilities`: the verifier was persisted, runtime creation failed, and activity remained `active`.

Please derive the final CAS from the latest durable row (while still fencing the old launch/controller identity), or otherwise make the failed-relaunch transition atomic with the earlier verifier write.

**[High] Sessions waiting in the reconcile worker queue are not fenced** — `backend/internal/session_manager/manager.go:2454-2469`

The operation fence is acquired only after a worker receives a job. With `ReconcileWorkers=4`, every later live session remains open to both user input and reaper termination for the full duration of the earlier attempts. A slow provider probe is explicitly one of the recovery cases here and can hold the Chat service's harness-wide probe mutex until timeout, so this window can span multiple 5-second reaper ticks. A queued dead TUI session with stale activity can therefore be durably terminated before its own reconcile starts, recreating the identity-loss behavior this PR is meant to remove. A deterministic 3-session/2-worker check showed the third session had `SessionMutationInProgress == false` while the first two were blocked in `IsAlive`.

Please establish the per-session fence before sessions become queued (or hold an equivalent startup-wide reaper/input barrier), then release each fence after that session's attempt/skip is settled.

Verification performed on exact head `3a900b5b9c2d921f47bf69a7766a25fc02c43459`:

- New focused race tests for the changed Codex Probe/Start/Resume and reconciliation paths passed.
- Windows cross-compilation passed for `internal/agentlaunch`, `codexappserver`, and `session_manager`.
- An independent smoke against the installed, unmodified npm `@openai/codex/bin/codex.js` under `PATH=/usr/bin:/bin` passed both version and app-server Probe. The committed fixture itself is synthetic (its `codex.js` body is shell and Start/Resume intercept spawn), so that real-launcher behavior is not covered by CI, but I did not find a PATH/env-loss defect in the production Codex change.

AgentWrapper:
Addressed both requested lifecycle blockers in `f7cf257ca`:

- failed relaunch preservation now derives its CAS and controller epoch from the latest durable row, verifies the post-signal state, and retries boundedly; the regression test models `BrowserCapabilityVerifier` persistence advancing `UpdatedAt` before `runtime.Create` fails, then proves ordinary Resume succeeds without losing native identity/worktree
- every live reconcile candidate is batch-reserved before bounded worker queuing, with each fence released after that session attempt; the deterministic 3-session/2-worker test asserts all three (including the queued candidate) are fenced while two probes block

Also reran the touched race suite and clean-environment `npm run lint` (full `go test ./...` + golangci): all pass. Ready for re-review.

AgentWrapper:
Re-review of corrected head `f7cf257ca8dd1eac805715f7cdb98d82e20dd466` (review-only): **changes still requested**.

### HIGH — a claimed Chat generation is still mistaken for a published/live controller

`relaunchCommitted` treats any changed, non-empty Chat `ControllerGeneration` as proof that relaunch committed (`backend/internal/session_manager/manager.go:2303-2310`), and `preserveFailedReconcileRelaunch` returns early on that predicate (`manager.go:2329-2337`). But the production Chat service durably claims the new generation *before* it constructs/publishes a controller (`backend/internal/service/chat/service.go:433-445`). Several ordinary-resume failures remain possible after that claim and before publication: snapshot/history import (`service.go:470-499`) and `ControllerReady` (`service.go:501-505`); the controller is not registered/started until `service.go:533-537`.

Consequently, a startup Chat resume whose native-history read fails can return an error with a fresh durable generation but no live controller. Reconcile then logs this as “error after controller commit,” returns nil, and leaves the row active/idle instead of exited. Once its batch fence is released, the session advertises a controller epoch that never existed; sends find no controller and the durable state does not expose the intended recoverable-exited result.

I reproduced both halves against this head in a temporary archive:

* a real SQLite + `service/chat` ordinary resume with a provider history-read error leaves `ControllerGeneration == "claimed-generation"` while `Service.Controller` returns `ErrNoController`;
* a manager-level production-shaped launcher that performs that claim and then fails makes `reconcileLive` return nil with `ActivityActive` and `HasLiveChatController == false`.

The commit predicate needs evidence of the controller-ready/published boundary, not the service's earlier generation claim (or it must additionally verify the live controller registry). This is the same durable-state class as the prior relaunch blocker, so I do not consider that blocker fully resolved for Chat.

### What is resolved / verified

* The stale boot-revision issue is fixed for pre-commit/TUI failure: the helper rereads `UpdatedAt`, signals the current launch epoch, rereads, and retries. I independently forced one CAS no-op and persistent contention; it succeeded on attempt 2 and stopped with uncertainty after exactly 3 attempts, respectively.
* The queue blocker is fixed: `beginAgentOperations` reserves every unowned candidate under one lock before drain waits and worker queuing (`session_input.go:121-158`), cancellation releases every reservation, and each queued job has a deferred release (`manager.go:2474-2535`). The 3-candidate/2-worker test deterministically checks the queued third fence and final release.
* I found no new deadlock, fence leak, shutdown regression, or platform-specific issue in the batch reservation path. The new logic is OS-neutral.

Test coverage note: the committed metadata-update test is a realistic TUI regression test, but it does not itself force the CAS no-op/retry path and there is no committed Chat post-claim failure case. Both would be valuable deterministic regressions.

Focused verification passed:

* `go test -race ./internal/session_manager -run 'Test(ReconcileLive_RuntimeFailureAfterLaunchMetadataUpdateLeavesSessionResumable|Reconcile_LivePassUsesConfiguredConcurrency|ReconcileLive_ChatCompatibilityFailureLeavesNativeResumeRecoverable|RestoreTerminatedChatOrchestratorAfterCompatibilityRecoveryKeepsIdentity)$' -count=1`
* `go test -race ./internal/session_manager -run 'Test(SessionInput|AgentOperation|ResumeWaitsForInFlightInput|KillWaitsForInFlightInput|Reconcile_LivePassUsesConfiguredConcurrency)' -count=1`

AgentWrapper:
Remaining Chat commit-boundary blocker addressed in `7c21a2100`:

- a changed durable `ControllerGeneration` is no longer sufficient by itself; errored Chat relaunch is treated as committed only when `HasLiveChatController` proves the controller crossed the ready/published registry boundary
- when native-history/readiness fails after the service claims a generation but before publication, reconciliation now applies exited against that claimed generation, keeping stale-controller fencing while making Resume available
- added a direct post-claim/no-controller regression plus explicit one-shot CAS retry and persistent 3-attempt contention-bound tests

Clean-environment `npm run lint` and the full touched race suite pass. Ready for re-review.

AgentWrapper:
Re-review of head `7c21a21009734a74bf015707bde42fdd5021a737`: **approval-quality review; no blocking findings.**

The remaining Chat commit-boundary issue is resolved:

* `relaunchCommitted` now requires both durable generation movement and `HasLiveChatController` (`backend/internal/session_manager/manager.go:2303-2315`). A generation claimed before history import/`ControllerReady` is therefore not accepted as a committed controller unless the service has actually published a live registry entry.
* On the post-claim failure path, `preserveFailedReconcileRelaunch` uses the freshly read row and puts its current claimed generation on the exited signal (`manager.go:2332-2377`). That preserves the new epoch as the stale-callback fence rather than signaling with the boot generation.
* The direct regression (`chat_spawn_test.go:236-281`) realistically advances both generation and revision, returns a post-claim history failure, and verifies live/exited state, retained claimed generation, retained provider/worktree identity, no published controller, and no destructive teardown.
* The fake lifecycle now applies production-like launch/generation fences before accepting activity signals. The deterministic CAS tests cover a one-shot lost CAS followed by success and persistent contention stopping after exactly three attempts (`manager_test.go:7456-7519`).

I specifically audited the registry/durable TOCTOU:

* A false registry result leads to a generation- and revision-fenced exited signal; a concurrent durable generation/revision change makes that signal a no-op and the readback/retry handles it.
* If a published controller stops after the live check, the production controller stop path reports `ActivityExited` with that same controller generation. `HasLiveChatController` also excludes stopped registry entries.
* Startup reconciliation's per-session operation fence prevents a concurrent manager resume/switch, while Chat Service's per-session controller gate serializes registry replacement/publication. I found no path where a failed post-claim launch can publish after returning its error.
* Old-generation callbacks remain rejected; retaining the newly claimed failed generation does not weaken fencing for a later replacement generation.

Focused verification at this exact head passed:

* `go test -race ./internal/session_manager -run 'Test(ReconcileLive_ChatFailureAfterGenerationClaimLeavesSessionExited|PreserveFailedReconcileRelaunchRetriesContendedCAS|PreserveFailedReconcileRelaunchBoundsPersistentCASContention|ReconcileLive_RuntimeFailureAfterLaunchMetadataUpdateLeavesSessionResumable|Reconcile_LivePassUsesConfiguredConcurrency)$' -count=20`
* `go test -race ./internal/lifecycle -run 'Test(ActivitySignalRejectsStaleChatControllerGenerationAcrossHandoff|MarkSpawnedPersistsChatControllerFacts)$' -count=20`
* `go test -race ./internal/service/chat -run 'Test(ServiceStopRetainsControllerUntilItsEventStreamActuallyEnds|StartWaitsForStoppedControllerCleanupBeforeRelaunch)$' -count=10`

A broader `go test -race ./internal/session_manager` run reached unrelated environment-dependent failures in existing spawn tests because the temporary archive run did not satisfy their tmux prerequisite; the focused changed-area race tests above all passed. No source changes were made during review.

AgentWrapper:
Fresh read-only review of exact head `d4c0e5d8120ee59ccd03cac510b1ab73e59277cf`: **changes requested**.

## HIGH — reserved-boundary native history is read but silently discarded before the atomic generation commit

For a historical repair, `providerBoundaryID` is non-empty, so Chat Service deliberately skips `ClaimChatControllerGeneration` (`backend/internal/service/chat/service.go:440-453`). It then calls `importNativeHistory` before `ControllerReady`/`CommitChatSpawn` (`service.go:477-511`). However every history event is projected through `Store.ProjectProviderEvent`, which returns `projected=false, err=nil` unless the session already owns the new controller generation (`backend/internal/storage/sqlite/store/conversation_store.go:1983-1989`). The terminated historical target still has its old/empty generation until `CommitChatSpawn`, so all non-empty native history is dropped. `importNativeHistory` treats that false projection as success and the ownership boundary/session activation then commits.

The new success test uses a `nativeHistoryConversation` with **zero events** and only checks `historyReads() == 1` (`historical_provider_restore_test.go:154-205`), so it does not verify the stated “import before ownership commit” invariant.

I reproduced this against the real-SQLite fixture on this exact head by returning stable user/assistant/turn-completed history events. `Start` succeeded and committed `handoff-248:provider`, but the snapshot contained only the old-root message; none of the target thread's native history was archived or projected. This can put later live messages ahead of their historical context; a later restart may import that history only after newer AO sequences already exist.

The fix needs a way to project/stage the proved thread's history without weakening generation fencing, then publish history + boundary + session owner in the intended order/atomic boundary. A regression must use non-empty native history and assert the resulting real-SQLite rows/order, not merely that `ReadHistory` was called.

## HIGH — the exact transition proof is not bound to the provider handle returned by `Resume`

Session Manager proves `rec.Metadata.ProviderConversationID` exactly (`chat_spawn.go:435-443`) and passes that handle to Chat Service (`chat_spawn.go:350-369`). But after `driver.Resume`, Service does not verify `conv.ProviderConversationID()` equals the proved/requested handle; it builds the reserved boundary from the returned value (`service.go:407-421,454-461`). `ControllerReady` then overwrites metadata with that returned value (`chat_spawn.go:375-389`), and `CommitChatSpawn` checks only that the branch matches the already-overwritten metadata (`conversation_store.go:365-370`).

A real-SQLite review regression with proof for `native-248` and a driver conversation returning `native-WRONG` committed successfully and rebound the target/boundary to the wrong thread. The built-in drivers currently preserve the requested id, but the ownership proof is a service invariant and should not be converted into an unchecked driver assumption. Reject/close before history import or commit when the resumed handle differs. Add this missing negative case.

## MEDIUM — “latest transition” proof selection is nondeterministic on timestamp ties

`GetLatestSessionInterfaceTransition` orders only by `created_at DESC` (`backend/internal/storage/sqlite/queries/session_interface_transitions.sql:28-35`). The new negative test gives the later transition a one-minute-newer timestamp, but does not cover equal timestamps. In real SQLite I inserted two sequential, completed transitions for the same session with identical `created_at`; the query consistently returned the first/older matching transition rather than the insertion-later transition. That can defeat the helper's “newer transition prevents stale proof” rule in any state where the record still matches the older handle. Add a deterministic tie-breaker/monotonic ordering and a real-store test.

## What looked sound

The deterministic boundary id, initial `ConversationForSession` owner check, ControllerReady owner/head rechecks, transactional branch/head/session rollback, idempotent ordinary restart after a successful epoch, and provider-resume/history/commit failure preservation are otherwise well structured. The real-SQLite stale-owner and stale-head cases exercise both pre-provider and commit-time races. I found no failure path in the tested matrix that directly unterminates the target when provider open/history/atomic commit fails.

Verification:

* `go test -race ./internal/service/chat -run '^TestHistoricalProjectProviderRestore' -count=5` — passed
* `go test -race ./internal/session_manager -run 'Test(HistoricalChatProviderScopeRequiresLatestCompletedMatchingHandoff|RestoreTerminatedChatOrchestratorPassesProvenProviderBoundary)$' -count=20` — passed
* `go test -race ./internal/service/chat -skip '^TestReview' -count=1` — passed
* Temporary real-SQLite non-empty-history regression — failed as described (history absent after successful commit)
* Temporary real-SQLite wrong-returned-handle regression — failed as described (wrong thread committed)
* Temporary real-SQLite equal-`created_at` latest-transition regression, `-count=10` — failed consistently (older row selected)

No repository or live `~/.ao` state was modified.

AgentWrapper:
Resolved all findings from review comment 5449758878 in exact head `e16473cee0f2b2139b31e29d70d7882130fd4dfa`.

1. **Atomic native-history ownership:** pending-boundary restore now reads/reconciles history without writing, then `MarkChatSpawnedPrepared`/`CommitChatSpawnPrepared` stages the branch and controller generation, projects stable provider events onto that new branch, and publishes the live session in one SQLite transaction. The target remains terminated/unavailable to input until commit. The real-SQLite success regression now imports stable user/assistant/turn events and asserts messages, completed turn ownership, branch IDs, and archived provider events. The final-write failure regression projects that same history and proves the child/head/generation/events all roll back.
2. **Returned-handle binding:** every Resume now requires `ChatConversation.ProviderConversationID()` to exactly equal the requested/proved handle before generation claim or history read. The real-SQLite wrong-handle regression proves zero history reads, no boundary/session commit, and the unchanged terminated target/transcript.
3. **Deterministic latest transition:** `GetLatestSessionInterfaceTransition` now orders equal `created_at` rows by SQLite `rowid DESC`; sqlc was regenerated from the source query. A real-store sequential insertion test proves the insertion-later transition wins under identical timestamps.

Verification passed: focused regressions, repeated targeted race runs, full relevant-package race (`service/chat`, `lifecycle`, `storage/sqlite/store`, `session_manager`), clean-environment `go test ./...`, `go vet` on touched packages, and `npm run lint` (full tests + golangci, 0 issues). No live `~/.ao` state or Restore operation was touched. Ready for fresh review of this head.

AgentWrapper:
Re-review of exact head `e16473cee0f2b2139b31e29d70d7882130fd4dfa` against my findings in comment 5449758878: **approval-quality; no blocking findings.**

### 1. Prepared history publication and rollback

The history path is now correctly split. `readNativeHistory` only reads/reconciles and validates stable event identities; it performs no AO projection (`backend/internal/service/chat/controller.go:506-568`). For a pending provider boundary, Service carries a `CommitProviderHistory` closure into `ControllerReady` instead of projecting early (`service.go:488-544`). Lifecycle requires the prepared-store capability and delegates it to `CommitChatSpawnPrepared` (`backend/internal/lifecycle/manager.go:1015-1088`).

`CommitChatSpawnPrepared` performs the complete unit under one SQLite transaction and `writeMu`: recheck current conversation owner and parent head, insert/activate the reserved branch, stage the controller generation, project each provider event and its turns/messages/activities using the same transaction-bound query handle, and finally publish the complete live session record (`backend/internal/storage/sqlite/store/conversation_store.go:360-446,2017-2083`). The uncommitted staged generation/head are invisible to concurrent readers. Any error returns through `inTx` and rolls back branch, head, generation, provider-event archive/projection, and session activation.

The real-SQLite success test now uses non-empty stable user/assistant/turn history and verifies ordering, branch ownership, provider-event branch/session attribution, completed turn ownership, generation, and live target. The forced final owner-write failure occurs after history projection and verifies that the target remains terminated, generation/head/transcript remain unchanged, no boundary exists, and zero provider events escape. That closes my prior silent-history and partial-rollback findings.

### 2. Exact resumed-handle binding

Immediately after `driver.Resume`, Service now requires `conv.ProviderConversationID()` to exactly equal the requested/proved id, closes the returned conversation, and errors before generation claim, history read, boundary creation, or ControllerReady when it differs (`service.go:441-451`). The real-SQLite negative test asserts no history read and a completely unchanged terminated target/head/transcript/boundary/events. The later boundary and metadata therefore remain bound to the same id Session Manager proved.

### 3. Deterministic latest transition

Both the source query and committed generated sqlc use `ORDER BY created_at DESC, rowid DESC`; there is no generated drift (`session_interface_transitions.sql:28-35`, generated `session_interface_transitions.sql.go:148-157`). The real-store equal-timestamp test creates/completes two transitions sequentially and proves the insertion-later row wins. This matches the table's ordinary rowid layout and the intended “latest claim” semantics.

### Race/idempotence/exposure audit

* Conversation owner and active-head facts are rechecked inside the same prepared transaction after all provider I/O; competing owner/head changes fail before any unit becomes visible.
* A failed prepared commit leaves the same deterministic boundary available for retry. After a successful commit, the active branch owns the target handle, so the next restart follows the ordinary stable-event dedupe path and creates no duplicate boundary.
* The Chat controller is not inserted into the registry or started until `ControllerReady` and the prepared transaction have completed. The durable session is published only by the final transaction write, and Session Manager's restore/reconcile operation fence remains held through the small post-commit registry-publication interval, so user/automation input cannot observe a half-published controller.
* Cancellation/projection/final-write errors close the provider conversation and do not expose a controller. I found no path that mutates the terminated session on a failed commit or weakens stale-generation fencing.

Verification at this exact head:

* `go test -race ./internal/service/chat -run '^TestHistoricalProjectProviderRestore' -count=10` — passed
* `go test -race ./internal/storage/sqlite/store -run '^TestLatestSessionInterfaceTransitionBreaksCreatedAtTiesByInsertion$' -count=20` — passed
* `go test -race ./internal/session_manager -run 'Test(HistoricalChatProviderScopeRequiresLatestCompletedMatchingHandoff|RestoreTerminatedChatOrchestratorPassesProvenProviderBoundary)$' -count=20` — passed
* `go test -race ./internal/service/chat -count=1` — passed
* `go test -race ./internal/storage/sqlite/store -run 'Test(CommitChatSpawn|LatestSessionInterfaceTransitionBreaksCreatedAtTiesByInsertion)' -count=10` — passed

No repository or live `~/.ao` state was modified.

PR #4558, opened 2026-08-27
refactor(tooltip): better tooltips!!

ronishrohan:
## Summary
- Reworks the shared `Tooltip` component (`frontend/src/renderer/components/ui/tooltip.tsx`) to match the currently-published shadcn Radix tooltip: 400ms open delay, `disableHoverableContent` (closes immediately on pointer-leave instead of tracking a hover "safe area"), `pointer-events-none`, and theme-matching colors (`bg-popover`/`text-popover-foreground`) instead of a hardcoded inverted white/black palette.
- Converts icon-only buttons across the app that previously relied on the native browser `title` attribute (no visible on-screen label) to the styled `Tooltip`, for consistent hover affordance — e.g. topbar "New terminal"/scroll-tab buttons, browser toolbar controls, chat message actions (edit/rollback), composer attach button, dialog close buttons, and more.
- Removes a few redundant/duplicate tooltips found along the way (tooltip text that duplicated a button's own visible label, a native `title` + styled `Tooltip` firing on the same element).
- Fixed a real regression caught during this work: wrapping a `Dialog`'s close (X) button in `Tooltip` broke Escape-to-close on nested dialogs, because the Tooltip opens on keyboard focus and its own dismiss layer intercepted the first Escape press before the Dialog's. Left that one button on native `title`.

https://github.com/user-attachments/assets/4bba9a00-b97a-41a5-a547-7c20924f557f

## Test plan
- [x] `npx tsc --noEmit -p .` — no new errors (remaining errors are pre-existing, confined to `packages/product-ui`)
- [x] `npx vitest run` — full suite passes except pre-existing/unrelated failures in `src/landing` (missing `cheerio` dependency, unrelated to this change)
- [x] Manually verified tooltip delay, theme colors, and Escape-to-close behavior on the fixed dialog


Vaibhaav-Tiwari:
@ronishrohan can you check this out for the options on right side panel as well?.

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Vaibhaav-Tiwari:
bro lemme know if you have only added tooltips for the left side pane, or if you are doing for right side as well?

ronishrohan:
hey, will let you know sorry

ronishrohan:
Added a tooltip to the terminate button in the Session Inspector (right-side panel) — that was the only icon-only control there missing one; everything else in that panel either already had a tooltip or shows a visible text label. Also rebased on the latest main (33 commits of restructuring — Kanban board changes, browser tab strip, terminal PTY host, queued message dock, etc.) and resolved the conflicts.

Vaibhaav-Tiwari (changes requested):
1. The new BrowserPanel toolbar tooltips need data-browser-native-overlay="true" on TooltipContent; otherwise
      they render behind the native browser view.
  
   2. Several tooltips wrap disabled buttons directly. Our shared Button applies disabled:pointer-events-none, so
      these tooltips cannot open. Please use a wrapper element, like the span approach used for the termination
      buttons.
  
   Please also add tests covering browser-overlay tooltips and disabled tooltip triggers.



and also please address the mrge conflicts

ronishrohan:
Addressed both:

1. **Browser overlay**: added `data-browser-native-overlay="true"` to all 8 new toolbar tooltips in `BrowserPanel.tsx` plus the 2 new ones in `BrowserTabsRail.tsx` so they paint above the native BrowserView.

2. **Disabled triggers**: wrapped every Tooltip whose trigger button can be disabled in a `<span className="inline-flex">` (the same pattern already used for the termination buttons) — went through the whole diff and found 21 cases across 13 files that needed it, not just the ones in BrowserPanel.

Added two tests in `BrowserPanel.test.tsx`: one confirming a toolbar tooltip carries the overlay marker, one confirming a disabled toolbar button's tooltip still opens via its wrapping span.

Also merged latest main again (no conflicts this time).

Vaibhaav-Tiwari (approved):

PR #4563, opened 2026-08-28
fix: restore turn changed-files Review and file open actions

Annieeeee11:
## Summary
- Fix the turn-end "Files Changed" card so **Review** opens the Files inspector in changed-only mode (the diff review view), not an empty full-tree explorer.
- Fix file row clicks so they resolve turn-diff paths (basenames, relative paths) to workspace paths and open the file in a center tab.
- Move the path tooltip onto the file row button so nested tooltip triggers no longer swallow clicks.

## Context
After the file-explorer refactor (#4252), `handleOpenFiles` no longer enabled `filesChangedOnly`, so Review appeared to do nothing. File opens also passed raw turn-diff paths without the basename matching the old `focusPath` logic used.

## Test plan
- [ ] In a chat session with a turn that changed files, click **Review** → Files inspector opens on the changed-files diff list.
- [ ] Click an individual file in the turn card → file opens in a center tab with the correct diff.
- [ ] Hover a file row → path tooltip still shows the full worktree path.
- [ ] `cd frontend && npx vitest run src/renderer/lib/workspace-file-path.test.ts src/renderer/components/chat/TurnChangedFiles.test.tsx`
- [ ] `cd frontend && npm run typecheck`

### DIFF
--- a/frontend/src/renderer/components/SessionView.test.tsx
+++ b/frontend/src/renderer/components/SessionView.test.tsx
@@ -169,6 +169,7 @@ vi.mock("./TerminalSwitchAgentButton", () => ({
 vi.mock("./chat/SessionChatSurface", () => ({
 	SessionChatSurface: ({
 		onOpenShell,
+		onOpenFile,
 		headerActions,
 		reviewerTerminal,
 		onOpenReviewerTerminal,
@@ -180,6 +181,7 @@ vi.mock("./chat/SessionChatSurface", () => ({
 		workspaceTabs,
 	}: {
 		onOpenShell?: () => void;
+		onOpenFile?: (path: string) => void;
 		headerActions?: ReactNode;
 		reviewerTerminal?: { handleId: string; harness: string };
 		onOpenReviewerTerminal?: (target: { handleId: string; harness: string }) => void;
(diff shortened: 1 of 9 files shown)


i-trytoohard:
Tracking issue for this fix: #4564 (reported by @goatedinsaan on Discord).

harshitsinghbhandari:
Two blocking path-resolution issues remain:
- `SessionView` reads only the current query cache, opens the tab, then starts `prefetchQuery`; on a cold/stale cache, basename paths remain unresolved and the later fetch never repairs the opened tab.
- `turnFileOpenPath` reduces absolute paths to basenames, so duplicate names such as `frontend/index.ts` and `backend/index.ts` cannot be disambiguated.
Please resolve/open only after workspace files are available and preserve enough path suffix to select the intended file.
Tests should cover cold cache, absolute paths, and duplicate basenames.

ronishrohan (changes requested):
Request changes

[ P1 ] `frontend/src/renderer/components/SessionView.tsx:905-942` — `resolveWorkspaceFilePath()` reads only the React Query cache, but both file-opening handlers call it before the subsequent `prefetchQuery()` can finish. On the first click, the cache is normally empty, so a basename such as `notes.txt` is passed to `openCenterFile` unchanged instead of resolving to `docs/notes.txt`; the requested file still does not open. Await/fetch the workspace-file list before resolving and opening the tab, and add an integration test for the cold-cache path.

ronishrohan (changes requested):
The cold-cache path-resolution P1 is fixed and the focused suite passes (118/118).

[ P2 ] This PR still includes an unrelated updater-settings change in `frontend/src/renderer/components/settings/UpdatesSection.tsx` (`fix: show channel-switch update message during manual check delay`). The stated PR scope is restoring turn changed-file review and open actions, and this repository requires one issue per PR. Please move that updater fix to its own PR or remove it from this branch.

ronishrohan (approved):
Approved at your direction; the remaining updater-settings change may stay in this PR.

PR #4565, opened 2026-08-28
feat: make turn plan panel collapsible via header icon

Annieeeee11:
## Summary
- Make the Plan panel collapsible by clicking the checklist icon next to the title
- When collapsed, keep the header visible (icon, "Plan", and done/total count) and hide the explanation and step list
- Add accessible expand/collapse labels and a unit test for the toggle behavior

## Test plan
- [ ] Open a session with an agent plan and confirm the panel starts expanded
- [ ] Click the checklist icon and confirm the explanation and steps hide while `done/total` stays visible
- [ ] Click the icon again and confirm the plan body reappears
- [ ] Run `cd frontend && npm run test -- --run src/renderer/components/chat/TurnPlan.test.tsx`

### DIFF
--- a/frontend/src/renderer/components/chat/TurnPlan.test.tsx
+++ b/frontend/src/renderer/components/chat/TurnPlan.test.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { describe, expect, it } from "vitest";
 import { TurnPlan } from "./TurnPlan";
 import type { ConversationPlan } from "../../types/conversation";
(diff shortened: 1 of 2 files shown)


harshitsinghbhandari (approved):

ronishrohan (changes requested):
Request changes

[ P2 ] `frontend/src/renderer/components/chat/TurnPlan.tsx:51` — the new icon-only disclosure button has no sizing or padding. Tailwind preflight resets button padding to zero, so its hit target is only the `size-4` (16px) SVG. The renderer control tokens define 24px as the smallest control (`--size-control-sm`). Please give the button a 24px target while retaining the 16px icon.

ronishrohan (approved):
Verified the requested change: the disclosure control now uses `size-control-sm` (24px) while the icon remains 16px. Focused TurnPlan tests pass (8/8).

PR #4567, opened 2026-08-28
fix: hide native scrollbars in command and code blocks

Annieeeee11:
## Summary
- Hide native scrollbars on chat explore command/output blocks, markdown code fences, and other chat `<pre>` scroll regions
- Scrolling still works via wheel/trackpad; only the visible track/thumb is removed
- Matches the existing `scrollbar-none` / `chat-scroll-viewport` pattern used elsewhere in the renderer

## Test plan
- [ ] Open a session with a long shell command in an expanded explore card — horizontal scroll works, no visible scrollbar
- [ ] Expand command output with many lines — vertical scroll works, no visible scrollbar
- [ ] View a fenced code block with a long line — horizontal scroll works, no visible scrollbar
- [ ] Confirm standalone command/patch `<pre>` blocks in the timeline also hide scrollbars while remaining scrollable

## Before
image

### DIFF
--- a/frontend/src/renderer/components/chat/ChatMarkdown.tsx
+++ b/frontend/src/renderer/components/chat/ChatMarkdown.tsx
@@ -158,7 +158,7 @@ function CodeBlock({ code, language }: { code: string; language?: string }) {
 					<CopyButton text={code} label="Copy code" />
 				</div>
 			</div>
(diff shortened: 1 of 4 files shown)


ronishrohan (approved):
no concrete issue, can merge

PR #4569, opened 2026-08-28
fix(telemetry): carry the request id into daemon-emitted events

yashksaini-coder:
## What

Sets `request_id` on telemetry events emitted below the HTTP layer, so a single daemon request can be traced across the layers it touches.

## Why

`telemetry_event` has a `request_id` column, `ports.TelemetryEvent` declares the field, and `storage/sqlite/store/telemetry_store.go` persists it — but nothing below the HTTP layer sets it. `middleware.GetReqID` is called at exactly five sites, all inside `internal/httpd`, so every event emitted from a service or from lifecycle lands with `request_id = ''`.

The practical effect: you have a request id from the `X-Request-Id` header or the `requestId` field of an error envelope, but querying for it returns only the CLI rows. There is no way to ask what a given request actually did, which is what the column is for.

### Reproduce

```bash
export AO_DATA_DIR=/tmp/ao-reqid AO_RUN_FILE=/tmp/ao-reqid/running.json
export AO_TELEMETRY_EVENTS=on AO_TELEMETRY_REMOTE=off AO_PORT=3099
mkdir -p "$AO_DATA_DIR"
cd backend && go run ./cmd/ao daemon
```

In a second shell, register any git repo and spawn into it, then look at what was recorded:

```bash
cd backend
go run ./cmd/ao project add --path /path/to/any/git/repo --name demo
go run ./cmd/ao spawn --project demo --harness claude-code --skip-agent-check \
  --mode tui --name t1 --prompt hi

sqlite3 "$AO_DATA_DIR/ao.db" \
  "SELECT source, COUNT(*), SUM(CASE WHEN request_id='' THEN 1 ELSE 0 END) AS empty
   FROM telemetry_event GROUP BY source;"
```

Before:

```
cli|3|0
daemon|1|1
project_service|1|1
session_service|2|2
```

Every daemon-side source is empty. Grepping the daemon log for the spawn's request id and querying `WHERE request_id = '<that id>'` returns 0 rows, even though the spawn took ~250ms across the session, workspace, runtime and lifecycle layers.

After, the same spawn returns its own events:

```
ao.session.spawned                   session_service  work-1
ao.onboarding.first_session_spawned  session_service  work-1
```

## How

The service emitters call `telemetry.Emit(context.Background(), ...)` on purpose — a cancelled request must not drop an event for work that already happened. That means a sink-level decorator reading the context at `Emit` time cannot work for them. The id is instead read at the emit site, before the detach.

`internal/reqid` is a small wrapper over `middleware.GetReqID` so that read is explicit and chi's context key is not spelled out in every caller. Worth a reviewer's opinion: this pulls `go-chi/chi/v5` into `internal/lifecycle`, which previously depended only on stdlib plus `domain`/`ports`/`sessionguard`. The dependency is unavoidable if the id is read at the emit site, since the context key is chi's own type, but if you would rather that boundary not move I am happy to rework it.

Covered: the session, project and review services, and lifecycle.

Intentionally left alone:

- `daemon.go:153` (`ao.daemon.started`) has no request in scope.
- The mobile-connect reporter also stays empty. It *does* run inside a live HTTP request (`httpd/auth.go:179`), but `lan_listener.go:38` applies `authMiddleware` outside the chi router, so it executes before `middleware.RequestID` and no id exists yet. Noting the real reason because the code reads as though it were request-free.
- `AggregatingSink` folds a burst of same-named events into one and keeps the last occurrence as the sample (`adapters/telemetry/aggregate.go:78`), so a rollup carries the last event's request id. Aggregation wraps only the remote sink; the local SQLite sink keeps every event unaggregated, and that is what the correlation query reads.

`change_log` has no `request_id` column, so the CDC leg still cannot be correlated. That is a larger change and is out of scope here.

## Testing

`go test ./...` — 160 packages pass. Tests assert the id reaches the event for each covered emitter.

Two failures on my machine are pre-existing and unrelated; both reproduce identically on a clean `main` checkout:

- `TestCopilotClassicPATIsNotAnAuthorizationSignal` — fails whenever a classic GitHub PAT is present in the environment.
- `TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell` — `can't find pane: 0`, the `base-index` issue in #4222 / #4224.

Found by running the daemon against an isolated data dir and cross-referencing the observed behaviour with the code path.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #4304, opened 2026-08-23
fix(chat): keep Codex prompt forks on one writer

illegalcall:
Fixes #4299

## Summary

- keep a Codex prompt-edit fork on the app-server connection that created it and already owns its writer lock
- atomically move the live controller to the forked branch without starting a second Codex writer
- preserve the existing resume fallback for providers that do not expose a loaded-fork binding, with unit and live Codex regression coverage

## Root cause

Codex `thread/fork` immediately loads and subscribes the forked thread in the creating app-server process. AO then tried to `thread/resume` that same fork through a second app-server process. Codex correctly rejected the second writer, which AO surfaced in the chat editor as `CHAT_RESUME_FAILED`.

## Local Electron evidence

Both screenshots are from the same isolated local desktop session while editing the second user prompt (the fork path).

| Before | After |
| --- | --- |
| Baseline build: the edit fails with `the stored provider conversation could not be resumed (CHAT_RESUME_FAILED)`. | Fixed build: the edited prompt is accepted and Codex replies `EDITED SECOND RESPONSE`. |
| Before: later prompt edit fails with CHAT_RESUME_FAILED | After: later prompt edit completes successfully |

## Verification

- `npm run lint` (clean AO environment): all Go tests and golangci-lint pass
- `cd backend && go build ./...`
- `go test -race ./internal/service/chat ./internal/adapters/chatdriver/codexappserver -count=1`
- `AO_CODEX_LIVE=1 go test ./internal/adapters/chatdriver/codexappserver -run '^TestLiveCodexAppServer$' -count=1 -v`
- real Electron before/after run against isolated `~/.ao/dev` state

## Risk

The new binding is optional and feature-detected. Non-Codex drivers retain the existing start/resume behavior, and no API, storage schema, or frontend contract changes are included.

PR #4588, opened 2026-08-28
fix(chat): hand off Codex edit forks safely

### LINKED ISSUE #4299: Codex prompt edit fails because forked thread already has an active writer
## Bug

Editing an earlier human message in a Codex Chat session consistently fails with `CHAT_RESUME_FAILED` instead of creating and activating the edited conversation branch.

**Source:** Reported in AO Chat with an attached screenshot  
**Analyzed against:** `9c1797cde982c0cdf925c74ea4588cf5f3fbcb57`  
**AO daemon:** checkout build (`dev`) on `127.0.0.1:3001`  
**Codex CLI:** `0.148.0`  
**OS:** macOS arm64  
**Confidence:** High  
**Priority:** P1 — the prompt-edit workflow is broken whenever the edited prompt has a preceding provider turn.

Reported `CHAT_RESUME_FAILED` error

## Reproduction

1. Start a Codex session in Chat mode.
2. Complete at least two human/assistant turns.
3. Edit the second human message and submit it.
4. Observe `the stored provider conversation could not be resumed (CHAT_RESUME_FAILED)`.

A minimized live provider harness reproduces the backend boundary directly:

1. Start a real Codex app-server conversation.
2. Complete one turn.
3. Call `thread/fork` through that turn.
4. While the source app-server remains open, start a second app-server and call `thread/resume` for the fork id.

This failed three out of three times with app-server error `-32600: thread <id> already has an active writer`. Waiting two seconds did not change the result. Closing the source connection before resume made the same harness pass twice.

## Root Cause

`Service.EditMessage` calls `ChatForker.Fork`, receives the new provider conversation id, and immediately calls `driver.Resume` for that id on a new app-server process. Codex `thread/fork` already loads the new thread and emits `thread/started`, so the source app-server owns the fork as its active writer. The second process cannot resume it.

AO only closes the source controller after the replacement controller has been created and installed. That point is unreachable because replacement resume fails first.

The existing service fake models `Fork` as returning a passive string and therefore cannot express provider writer ownership. Adapter tests cover fork and resume independently rather than chaining them as the edit flow does.

## Fix

Preserve AO's one-writer and failure-recovery guarantees while changing the edit handoff sequence so a fork is not resumed by a second process while its creating app-server still owns it. Add regression coverage with a writer-aware fake plus an opt-in live Codex fork/resume test.

## Impact

- Editing the first prompt is unaffected because it starts a fresh provider conversation.
- Editing any later prompt is broken for Codex Chat sessions.
- The original conversation and worktree remain intact; the edit is refused rather than silently losing history.

## Related

- [PR #3676](https://github.com/Untrivial-ai/agent-orchestrator/pull/3676) — introduced Codex prompt editing with conversation branches.

illegalcall:
Fixes [#4299](https://github.com/Untrivial-ai/agent-orchestrator/issues/4299)

Alternative implementation to [PR #4304](https://github.com/Untrivial-ai/agent-orchestrator/pull/4304).

## Summary

- close the fenced source controller after Codex creates a native edit fork, releasing the fork's active-writer lock before a replacement app-server resumes it
- restore the original source controller with a fresh generation if replacement preparation, branch activation, or edited-turn dispatch fails
- hold registry cleanup across the intentional close/resume handoff so the session never loses its published controller entry
- stop rendering persistent provider-materialization notices such as `Exact context · Provider history was preserved...` above the composer after an old-message edit

## Root cause

Codex `thread/fork` loads the child thread into the source app-server, so that process remains the child's active writer. AO immediately tried to resume the child from a second app-server; Codex correctly rejected the competing writer with `-32600: thread <id> already has an active writer`, surfaced as `CHAT_RESUME_FAILED`.

This implementation fences and closes the idle source after the fork succeeds, resumes the child only after the writer is released, and reopens the original provider thread if the replacement cannot be kept.

## Local Electron evidence

Both flows were exercised in the real Electron app built from this branch against local AO data. Each session completed an original later turn, then edited that user message through the visible chat UI.

### Codex native fork

The edited prompt produced conversation branch `2 / 2` and the exact replacement response `CODEX EDIT SUCCESS`. The daemon reported a completed native branch, and the persistent `Exact context` banner is absent.

Codex later-message edit completed

### Claude prompt replay

The edited prompt produced conversation branch `2 / 2` and the exact replacement response `CLAUDE EDIT SUCCESS`. The daemon reported `approximate_context` with `replayTruncated: false`, and no persistent context banner is shown after completion.

Claude later-message edit completed

## Verification

- `cd backend && go build ./... && go test ./...`
- `cd backend && go test ./internal/service/chat -count=1`
- `cd backend && go test -race ./internal/service/chat -run 'TestEditMessage(ClosesSourceWriterBeforeResumingNativeFork|ExplicitRefusalRestoresSourceBranch|UndispatchedAttemptRestoresSourceBranch)' -count=1`
- `cd frontend && npm test -- src/renderer/components/chat/ChatWorkspace.test.tsx` (81 tests)
- fresh native Electron run from this branch against `~/.ao/data`

## Review follow-up (`34d2f387f`)

The requested lifecycle/concurrency gaps are addressed on top of current `main` (`4ebbe00b5`):

- intentional source retirement during a native edit no longer publishes a session-level `ActivityExited`; unrecoverable replacement and source recovery still publish the truthful exit
- branch handoff completion releases registry cleanup without reopening the obsolete controller's intake, so a command that captured the source before the swap receives `ErrControllerHandoff`
- once the native fork succeeds, source close, replacement resume, activation, dispatch, and recovery use bounded detached contexts; recovery distinguishes stop initiation from confirmed closure

Exact-head validation:

- four deterministic lifecycle/concurrency regressions: passed 10 consecutive runs
- `go test ./internal/service/chat ./internal/lifecycle ./internal/httpd/controllers -count=1`: passed
- `go test -race ./internal/service/chat -count=1`: passed
- frontend typecheck: passed
- `ChatWorkspace.test.tsx`: 81/81 passed
- repository-wide Go run: all packages passed except the unrelated timing-sensitive fake-adapter speed assertion; that package passed immediately in isolation


backend/internal/service/chat/history.go
```diff
@@ -219,17 +219,26 @@ func (s *Service) EditMessage(
...
-				SessionID: cfg.SessionID, ProviderConversationID: providerConversationID,
-				DataDir: cfg.DataDir, WorkspacePath: cfg.WorkspacePath, Env: cfg.Env,
-				Model: cfg.Model, Permissions: cfg.Permissions, SystemPrompt: cfg.SystemPrompt,
-				ProviderScopeID:       sourceBranch.ProviderScopeID,
-				AdditionalDirectories: cfg.AdditionalDirectories, MCPServers: cfg.MCPServers,
-			})
+			// Codex loads a fork into the source app-server, which remains that
+			// child's active writer until the process exits. Close the fenced, idle
+			// source before another driver process resumes the child.
+			if closeErr := source.Close(ctx); closeErr != nil {
```
nikhilachale: **[P1] Avoid publishing the source exit during a successful handoff.** `source.Close` waits until `project()` emits `ActivityExited` (`chat.controller.stopped`) while the source generation still owns the session. `ActivateConversationBranch` later changes the provider/generation but leaves `activity_state` untouched, and lifecycle rejects later `chat.turn.started`/completed signals from an exited session unless the event is `user-prompt-submit` (which Chat does not emit). A successful native edit therefore leaves the live session permanently displayed as exited. Please fence/suppress this intentional source exit or establish the replacement lifecycle state atomically.

backend/internal/service/chat/history.go
```diff
@@ -611,12 +685,14 @@ func (s *Service) installStartedBranchController(
...
 	go func() {
 		replacement.Wait()
+		replacement.waitForBranchHandoff()
 		s.mu.Lock()
 		if current := s.controllers[id]; current == replacement {
 			delete(s.controllers, id)
 		}
 		s.mu.Unlock()
 	}()
+	source.AbortHandoff()
```
nikhilachale: **[P2] Keep the old controller's intake fenced until it is closed.** A concurrent `Service.Send` can capture `source` before the registry swap, enter after this `AbortHandoff`, see no fence, and append/dispatch with the stale generation after the durable head has moved. I reproduced this deterministically by blocking `source.Close`: the racing send returned `nil` instead of `ErrControllerHandoff`. Because projection/cleanup is generation-fenced, that turn can remain orphaned. Please close the source before reopening the fence, or separate latch release from intake reopening.

backend/internal/service/chat/history.go
```diff
@@ -254,13 +263,27 @@ func (s *Service) EditMessage(
 		}
 	}
 	if err != nil {
-		return EditMessageResult{}, classify(fmt.Errorf("prepare edited conversation: %w", err))
+		prepareErr := classify(fmt.Errorf("prepare edited conversation: %w", err))
+		if sourceClosed {
+			if restoreErr := s.restoreClosedSourceController(
+				ctx, id, source, sourceBranch, cfg, driver); restoreErr != nil {
```
nikhilachale, line 279: **[P2] Do not reuse the request context for post-close recovery.** Once `source.Close` has stopped the provider, cancellation during `Resume` makes this restoration run with the same canceled context, so both replacement and recovery fail and registry cleanup removes the stopped source. Cancellation while `Close` is waiting is worse: `sourceClosed` remains false even though the one-shot stop has already been initiated, so restoration is skipped entirely. The destructive post-close section/recovery needs its own bounded non-request context and must distinguish stop initiation from confirmed closure.

nikhilachale (changes requested):
Requesting changes: the close-before-resume sequence fixes the Codex active-writer conflict, but it currently introduces three lifecycle/concurrency gaps. Details are inline. I verified this head with the focused chat tests and race subset; I also reproduced the stale-controller send with a deterministic temporary regression test.

backend/internal/service/chat/history.go
illegalcall: Fixed in `34d2f387f`. Intentional branch-controller retirement is marked before the source is published as obsolete and suppresses both explicit and stream-closure `chat.controller.stopped` signals. If replacement preparation and source recovery both fail, suppression is removed and the source generation publishes the truthful exit. Regressions: `TestEditMessageNativeForkDoesNotPublishSourceExitBeforeReplacementTurn` and `TestEditMessageReportsExitWhenReplacementAndRecoveryBothFail`.

backend/internal/service/chat/history.go
illegalcall: Fixed in `34d2f387f`. Successful installation now completes only the branch-handoff cleanup latch; it never calls `AbortHandoff` or reopens obsolete intake. `TestActivateBranchKeepsCapturedSourceIntakeFencedUntilClose` captures the source before activation, blocks its close after the registry swap, and verifies a late send returns `ErrControllerHandoff` rather than succeeding.

backend/internal/service/chat/history.go
illegalcall: Fixed in `34d2f387f`. After native fork succeeds, the irreversible close/resume/activation/dispatch section runs under a bounded detached context. Recovery creates its own bounded detached context, waits for a stop that was initiated but not yet confirmed, and restores the source with a fresh generation. `TestEditMessageRestoresSourceAfterRequestCancellationDuringClose` cancels the request while close is awaiting stream termination, forces replacement resume to fail, and verifies source restoration succeeds.

nikhilachale (approved):

somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4589, opened 2026-08-28
refactor(agents): centralize harness readiness

whoisasx:
## What

Centralize harness installation and authentication readiness in one daemon-owned, in-memory coordinator, expose normalized readiness endpoints, and migrate daemon, CLI, and desktop consumers to the shared boundary.

## Why

Agent inventory, refresh throttling, authentication probes, and launch checks were split across persistence, daemon services, CLI flows, and renderer state. That allowed consumers to disagree about freshness and reconstruct readiness from separate supported/installed/authorized lists.

## How

- Add normalized installation, authentication, effective-readiness, freshness, purpose, timestamp, and safe-reason domain types.
- Coordinate per-harness snapshots with single-flight checks, fresh adapter instances, purpose-specific TTLs, bounded startup warm-up, timeouts, retry backoff, invalidation, and daemon-owned cancellation.
- Add cached `GET /api/v1/agents/readiness` and ensured `POST /api/v1/agents/readiness/ensure`, while retaining compatibility projections for `/agents`, `/agents/refresh`, and `/agents/{agent}/probe`.
- Remove durable agent inventory storage with reversible migration `0120`; repair development databases that recorded the branch's former cache-drop migration at `0119`, then apply main's canonical completed-plan `0119` repair.
- Route startup presence, session launch, reviewer/switch preflight, successful managed installs, `ao spawn`, and `ao agent ls` through the coordinator.
- Refresh display readiness for every harness when Task Composer opens, retain launch-fresh checks for the selected harness, and join/cache targeted readiness reconciliation after binary or authentication launch failures.
- Replace renderer refresh/throttle policy with normalized readiness query/ensure hooks and one readiness object for menus and product UI controls.
- Keep `ao agent ls --refresh` as a deprecated compatibility flag that explicitly forces fresh installation and authentication checks.

Intentional omissions: profiles, capacity, credentials, and new profile-switching behavior remain outside Phase 1. Ordinary launch keeps unauthorized and unknown authentication advisory; existing reviewer and agent-switch policies remain stricter. The Agy/Continue installed-as-authorized semantics remain unchanged pending the non-blocking follow-up discussion.

## Testing

- `npm run api`
- `go test ./internal/storage/sqlite/... ./internal/service/agent/... ./internal/httpd/...`
- `npm test -- src/renderer/components/TaskComposer.test.tsx src/renderer/hooks/useAgentReadinessQuery.test.tsx src/renderer/components/ProjectSettingsForm.test.tsx` (68 passed)
- `npm run typecheck`

## Checklist


backend/internal/storage/sqlite/migrations/0122_drop_agent_inventory_cache.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+DROP TABLE IF EXISTS agent_inventory_cache;
```
illegalcall, line 2: [P1] Renumber this migration and repair existing branch histories

Current `main` already owns migration `0119` (`0119_finalize_completed_conversation_plans.sql`). Please move this drop to `0120`, update the migration ledger/tests/comments, and add migration-history repair for databases that already recorded this branch's old `0119`. Renumbering alone leaves those databases with version 119 marked applied, so they would skip `main`'s real 0119 data-repair migration.

frontend/src/renderer/components/TaskComposer.tsx
```diff
@@ -197,22 +201,16 @@ export function TaskComposer({
...
-		});
-	}, [queryClient]);
 	// The composer preselects the agent and model a spawn would actually use
 	// instead of parking the controls on a "default" label the user has to
 	// remember. Both resolved values remain directly editable.
 	const projectWorkerAgent = projectQuery.data?.config?.worker?.agent ?? "";
 	const globalDefaultAgent = projectQuery.data?.agent ?? "";
 	const defaultWorkerAgent = projectWorkerAgent || globalDefaultAgent;
 	const selectedAgent = agent || defaultWorkerAgent;
+	useEnsureAgentReadiness({ agentIds: selectedAgent ? [selectedAgent] : [] });
```
illegalcall: [P1] Refresh every agent when Task Composer opens

This selected-only ensure leaves stale negative results for non-selected agents disabled. I reproduced this in the native Electron app on this exact head: Agy was initially absent, then its binary was added while the app remained open; after the 5-minute display TTL expired, reopening Task Composer issued readiness requests only for `claude-code`, and Agy remained `Needs install`/disabled until a direct targeted ensure was sent. Please ensure display readiness for all agents on composer open (while retaining the selected-agent launch ensure) and add a regression test.

frontend/src/renderer/components/TaskComposer.tsx
```diff
@@ -170,7 +174,7 @@ export function TaskComposer({
 				return data.workerId;
 			} catch (err) {
 				void captureRendererEvent("ao.renderer.task_create_failed", { project_id: input.projectId });
-				void queryClient.invalidateQueries({ queryKey: agentsQueryKey });
+				void queryClient.invalidateQueries({ queryKey: agentReadinessQueryKey });
```
illegalcall: [P1] Wait for completed readiness reconciliation after launch failure

The daemon already invalidates readiness and starts an asynchronous recheck for binary/auth launch errors. This immediate invalidate/refetch can race that recheck and cache the old or `checking` snapshot; because the readiness query uses `staleTime: Infinity` and there is no completion event or poll, the final result may never reach the UI. Please await a targeted ensure that joins the backend's in-flight check, cache the completed snapshot, and cover the race with a regression test.

backend/internal/service/agent/readiness.go
```diff
@@ -0,0 +1,255 @@
...
+
+// Probe is the legacy projection of a targeted launch ensure.
+func (s *Service) Probe(ctx context.Context, agentID string) (ProbeResult, error) {
+	if err := ctx.Err(); err != nil {
+		return ProbeResult{}, err
+	}
+	if _, ok := s.agent(agentID); !ok {
+		return ProbeResult{Agent: Info{ID: agentID}, Supported: false, Installed: false}, nil
+	}
+	readiness, err := s.EnsureReadiness(ctx, []string{agentID}, domain.AgentReadinessPurposeLaunch)
```
illegalcall, line 190: [P2] Align the probe contract with its cached behavior

`EnsureReadiness(..., launch)` can reuse a result up to the launch freshness window, while this endpoint is described in the OpenAPI contract as running a fresh local probe. Either force a targeted probe here or update the endpoint/spec wording to promise launch-fresh readiness rather than unconditional freshness. The cached behavior is reasonable; the contract just needs to match it.

docs/cli/README.md
```diff
@@ -63,9 +63,10 @@ Every product command resolves to a daemon HTTP route. Run `ao <command>
 | `ao browser ...`                    | `GET /api/v1/browser/status`, `POST /api/v1/browser/commands` |
 | `ao hooks <agent> <event>`          | `POST /api/v1/sessions/{id}/activity` (hidden) |
 
-`ao agent ls` prints the daemon-supported agent catalog with local install/auth
-readiness. Use `--refresh` to rerun the bounded local probes and `--json` to
-print the raw inventory response.
+`ao agent ls` asks the daemon to ensure display readiness, then prints the
+existing table or legacy JSON projection. The daemon alone decides whether a
+native check is needed. `--refresh` remains as a hidden deprecated alias with
```
illegalcall: [P2] Document `--refresh` as a forced check

This contradicts both the command table above and the implementation: `ao agent ls --refresh` explicitly bypasses cached readiness and runs fresh installation/authentication checks. Please replace the hidden/non-forcing wording with the actual forced-refresh behavior.

backend/internal/adapters/agent/agy/auth.go
```diff
@@ -13,8 +13,8 @@ func (p *Plugin) AuthStatus(ctx context.Context) (ports.AgentAuthStatus, error)
...
 		return ports.AgentAuthStatusUnknown, err
 	}
-	// Antigravity authenticates through its OS keyring and browser sign-in. The
-	// official CLI does not expose a documented non-interactive credential file
-	// or status command, so local auth cannot safely be inferred here.
-	return ports.AgentAuthStatusUnknown, nil
+	// Agy authenticates through its OS keyring and browser sign-in without a
+	// documented non-interactive status command. Treat an installed binary as
+	// sufficient local readiness evidence and let launch remain authoritative.
+	return ports.AgentAuthStatusAuthorized, nil
```
illegalcall, line 19: Question (non-blocking): should this adapter return `unknown` rather than `authorized` when the only evidence is that the binary resolves? `unknown` remains selectable/launchable, while `authorized` claims an authentication check passed and affects UI/telemetry. Continue currently makes the same assumption, so it would be good to document or align both adapters.

illegalcall (changes requested):
Requesting changes for the readiness integration issues we aligned on:

- Resolve the current `0119` migration collision, including repair for databases that already recorded this branch’s old `0119`.
- Refresh all agent display readiness when Task Composer opens, and add the reproduced regression test.
- Make failed-launch readiness reconciliation wait for and cache the completed targeted result.
- Align the probe endpoint contract with its launch-fresh cached behavior.
- Correct the `--refresh` docs and the stale PR description.

The coordinator model itself looks sound: centralized ownership, single-flight checks, freshness policies, failure preservation, and unknown-as-advisory are all good. The Agy/Continue authentication comment is a non-blocking semantics question.

backend/internal/service/agent/readiness.go
nikhilachale: Fixed in 987273083. The OpenAPI source and generated artifacts now describe this as ensuring launch-fresh readiness, matching the existing launch freshness window and cache reuse.

frontend/src/renderer/components/TaskComposer.tsx
nikhilachale: Fixed in 987273083. Task Composer now ensures display readiness for all harnesses on mount and separately ensures launch-fresh readiness for the selected harness. TaskComposer.test.tsx covers both calls.

docs/cli/README.md
nikhilachale: Fixed in 987273083. The CLI docs now state that --refresh is deprecated compatibility syntax that forces fresh installation and authentication checks.

backend/internal/adapters/agent/agy/auth.go
nikhilachale: Left unchanged as non-blocking. The PR description now calls out the Agy/Continue installed-as-authorized semantics as a follow-up decision rather than implying a verified non-interactive auth probe.

backend/internal/storage/sqlite/migrations/0122_drop_agent_inventory_cache.sql
nikhilachale: Fixed in 987273083. The cache drop is now 0120, the shipped-migration ledger records canonical 0119 and 0120, and startup repairs old branch databases only when 119 is applied, 120 is absent, and agent_inventory_cache is already gone. The regression test verifies the canonical completed-plan 0119 runs and a second startup does not rewrite migration history.

frontend/src/renderer/components/TaskComposer.tsx
nikhilachale: Fixed in 987273083. Binary/auth launch failures now await the targeted launch ensure, join the daemon in-flight recheck, and merge the completed result into the readiness cache before surfacing the original launch error. The regression test holds the ensure pending and verifies the UI waits and caches the completed snapshot.

illegalcall (approved):

PR #4591, opened 2026-08-28
fix(lifecycle): reconcile project orchestrator ownership

### LINKED ISSUE #4590: Startup restore can publish a second live project orchestrator
## Bug

Startup reconciliation can restore a shutdown-marked historical orchestrator even when the same project already has a live replacement. AO then runs two coordinators in the same canonical project worktree and shows a banner claiming the stale one will be cleaned up, although no such cleanup exists.

**Observed on:** macOS arm64, desktop `0.12.10-nightly.202608280852` (`72c162e20`, including merged [PR #4553](https://github.com/Untrivial-ai/agent-orchestrator/pull/4553))

**Analyzed against:** `72c162e20` (`origin/main`)

**Confidence:** High — correlated the restart timeline with durable session rows/controller generations and traced the exact `ReconcileBackground -> RestoreAll` path.

**Priority:** P1 — duplicate project coordinators can act concurrently in one worktree and the offered cleanup action may discard the wrong context.

## Reproduction

1. Have a terminated project orchestrator with a `session_worktrees` shutdown-restore marker.
2. Create or retain a different live orchestrator for the same project.
3. Restart the daemon so background startup reconciliation runs.
4. Observe that `RestoreAll` independently restores every marked session, including the historical orchestrator.
5. The project now has two nonterminated orchestrators sharing its canonical branch/worktree.

The production incident had historical Chat orchestrator `248` and replacement TUI orchestrator `254`. Startup correctly restored `248`'s native provider conversation after [PR #4553](https://github.com/Untrivial-ai/agent-orchestrator/pull/4553), but it did not first prove project-scoped orchestrator ownership, so `254` remained live as well.

## Root cause

- `backend/internal/session_manager/manager.go` treated shutdown restore markers as independent per-session work and had no project-level selection/admission boundary.
- The session service serialized normal orchestrator spawn/replacement in its own lock, but startup reconciliation did not share that lock.
- The daemon serves the API while background reconciliation runs, so a discovery-time liveness check alone would still race an API spawn or explicit restore unless ownership is re-read under a shared lock and held through controller publication.
- When several saved duplicates exist after clean shutdown, creation time is not always the correct owner: the durable project conversation's `current_session_id` is the stronger Chat-continuity fact.
- The frontend duplicate banner claimed automatic cleanup, and exposed Restart even though choosing/retiring a duplicate by timestamp can discard context or tear down a shared workspace.

## Fix direction

1. Share one project-scoped orchestrator ownership lock between normal session-service operations and startup restore.
2. Select at most one saved orchestrator per project. Prefer the durable project-conversation owner; use deterministic recency only when no project Chat owner exists.
3. Under the shared project lock and session mutation fence, re-read the target, restore marker, and active sibling set immediately before worktree/controller work.
4. Reject ordinary restore of a historical orchestrator while a sibling is live.
5. When the durable project-conversation owner is exact, park competing live controllers without touching their native IDs or shared workspace, then restore the owner. Without exact proof, leave duplicates unchanged for explicit resolution.
6. Make the unresolved-duplicate warning truthful and remove the destructive Restart action from that state.

## Acceptance criteria

- [ ] A terminated marked orchestrator is not restored while a sibling orchestrator is live.
- [ ] More than one saved orchestrator results in exactly one restore; non-selected markers remain recoverable.
- [ ] The durable project-conversation owner wins over a newer duplicate, preserving Chat identity/transcript continuity.
- [ ] A conversation-ownership read failure fails closed and leaves candidates terminated.
- [ ] API spawn/restore cannot race startup's discovery snapshot into a duplicate; ownership is re-read under a shared project lock held through publication.
- [ ] Explicit restore returns a typed conflict while another orchestrator is live.
- [ ] Existing duplicates with an exact project-conversation owner are repaired by stopping/terminating only competing controllers, preserving their native IDs/session rows and never touching the shared workspace.
- [ ] Existing duplicates without exact ownership proof remain readable and are not destructively rewritten.
- [ ] The UI does not promise cleanup when ownership is ambiguous or offer Restart as duplicate cleanup.
- [ ] Focused and full race tests pass.

## Related

- [PR #4553](https://github.com/Untrivial-ai/agent-orchestrator/pull/4553) — made Chat restart recovery non-destructive; exposed this separate project-ownership bug once the historical session could recover successfully.
- [PR #4562](https://github.com/Untrivial-ai/agent-orchestrator/pull/4562) — combined preview used during earlier restart validation.
- [#4573](https://github.com/Untrivial-ai/agent-orchestrator/issues/4573) — separate restart UX issue: Chat sessions flash Exited and reorder while controllers reconnect.

AgentWrapper:
## Summary

Fixes #4590.

- make project-orchestrator ownership a shared Session Manager boundary used by startup reconciliation, spawn/replacement, and explicit restore
- restore at most one saved orchestrator per project, preferring the durable project-conversation owner over timestamp recency
- when the durable owner is exact and a competing controller is live, stop only the competing controller and park its session as recoverable history; do not remove the shared workspace or native provider/session identity
- fail closed when ownership is absent, ambiguous, or changes during reconciliation
- replace the misleading duplicate-orchestrator Restart action with a truthful non-destructive warning

## Root cause and reproduction

`RestoreAll` treated shutdown restore markers as independent per-session facts. Project-scoped orchestrator ownership was enforced separately in the session service and was not rechecked under the same lock before startup controller publication. An older terminated Chat orchestrator could therefore retain the durable project conversation while a replacement TUI orchestrator remained live on the same canonical branch/worktree. Startup either restored both ownership histories or left the wrong one active depending on timing; the UI then promised that a later daemon reconcile would clean up duplicates even though no such ownership repair existed.

The regression tests model the nightly incident without reading or modifying `~/.ao`: a terminated Chat owner with its original provider conversation and restore marker, plus a live TUI duplicate sharing the canonical workspace. Reconciliation proves the Chat session from the durable project conversation, stops and parks only the TUI controller, then restores the Chat owner in place with the same AO session ID, workspace, branch, and provider conversation.

## State-machine / safety boundary

All project-orchestrator admission paths now share the Session Manager's project lock. Startup performs fresh session, liveness, marker, and conversation-owner reads after acquiring the relevant project/session fences. A live sibling blocks ordinary restore. Exact durable ownership may repair a legacy duplicate by stopping its controller and marking that row terminated, but cleanup deliberately does not destroy its workspace or erase native identity. If controller stop fails, the competing row remains live and owner restore remains blocked. If ownership cannot be proved, AO leaves every session untouched for an explicit decision.

No schema, migration, generated-code, or stored-session rewrite is involved.

## Recovery notes

After installing a build containing this change, the observed shape (historical terminated Chat owner plus a live TUI replacement on the same project workspace) is repaired on startup only if the project conversation still proves the historical Chat owner. The replacement controller is parked as a recoverable terminated session and the owner is resumed through the existing Chat restore path, preserving its native thread and transcript. If that proof is missing or ambiguous, automatic cleanup is disabled and the UI reports that manual resolution is required.

This recovery path is covered with fakes and production-shaped durable ownership facts. I did **not** run it against or mutate the user's live `~/.ao` data.

## Tests

- `cd backend && env -i HOME="$HOME" PATH="$PATH" TMPDIR="$TMPDIR" go test ./...`
- `cd backend && env -i HOME="$HOME" PATH="$PATH" TMPDIR="$TMPDIR" go test -race ./internal/session_manager ./internal/service/session`
- `golangci-lint run` (v2.12.2)
- `cd frontend && npx vitest run src/renderer/components/SessionsBoard.test.tsx src/renderer/types/workspace.test.ts`
- full frontend suite: 230 files passed, 3,019 tests passed, 1 skipped

### DIFF
--- a/backend/internal/session_manager/manager.go
+++ b/backend/internal/session_manager/manager.go
@@ -38,6 +38,9 @@ var (
 	ErrAgentNotExited      = errors.New("session: agent has not exited")
 	ErrAgentExitInProgress = errors.New("session: agent exit is already in progress")
 	ErrIncompleteHandle    = errors.New("session: incomplete teardown handle")
+	// ErrOrchestratorAlreadyActive prevents an ordinary per-session restore from
+	// bypassing the project-scoped single-orchestrator ownership boundary.
+	ErrOrchestratorAlreadyActive = errors.New("session: another project orchestrator is already active")
 	// ErrProjectNotResolvable means the spawn's project has no usable repo
 	// (unregistered, archived, or missing a path). The API maps it to a 400.
 	ErrProjectNotResolvable = errors.New("session: project repo not resolvable")
@@ -357,6 +360,14 @@ type Store interface {
 	DeleteSessionWorktrees(ctx context.Context, id domain.SessionID) error
 }
 
+// projectConversationOwnerStore exposes the durable owner of a project's Chat
+// narrative. Startup restore uses this narrower optional read to prefer the
+// orchestrator that already owns the project conversation over a newer
+// duplicate created by an older release.
+type projectConversationOwnerStore interface {
+	ConversationForSession(context.Context, domain.SessionID) (domain.ConversationRecord, error)
+}
+
 // Manager coordinates internal session spawn, restore, kill, and cleanup over
 // the outbound ports. User-facing read-model assembly lives in the service package.
 type Manager struct {
@@ -414,6 +425,13 @@ type Manager struct {
 	startupBackgroundReconcileOnce  sync.Once
 	agentOpMu                       sync.Mutex
 	agentOperations                 map[domain.SessionID]agentOperationKind
+	// orchestratorProjectLocks serializes every project-scoped orchestrator
+	// ownership decision in this daemon. The session service uses the same lock
+	// for spawn/replacement/restore, while startup reconciliation holds it from
+	// its final liveness read through controller publication. The supervisor's
+	// single-daemon ownership keeps this process-local boundary authoritative.
+	orchestratorProjectLocksMu sync.Mutex
+	orchestratorProjectLocks   map[domain.ProjectID]*sync.Mutex
 	// switchDecisionInput opens a narrow human-only terminal lane while the
 	// source is blocked on permission during a mandatory switch.
 	switchDecisionInput map[domain.SessionID]domain.AgentSwitchID
@@ -757,6 +775,7 @@ func New(d Deps) *Manager {
 		backgroundContext:              d.BackgroundContext,
 		startupBackgroundReconcileDone: make(chan struct{}),
 		agentOperations:                make(map[domain.SessionID]agentOperationKind),
+		orchestratorProjectLocks:       make(map[domain.ProjectID]*sync.Mutex),
 		switchDecisionInput:            make(map[domain.SessionID]domain.AgentSwitchID),
 		retainedSwitches:               make(map[domain.SessionID]struct{}),
 		inputLeases:                    make(map[domain.SessionID]int),
@@ -1962,6 +1981,27 @@ func (m *Manager) RestoreWithMode(ctx context.Context, id domain.SessionID) (Res
 	return m.relaunchRestoredSession(ctx, rec, project, ws)
 }
 
+// LockOrchestratorProject acquires this daemon's project-scoped orchestrator
+// ownership lock. Callers must hold it across the last durable ownership read,
+// any external controller/workspace work, and the final session activation.
+// The session service deliberately shares this lock with startup reconciliation
+// so the API cannot create a replacement in the middle of a saved restore.
+func (m *Manager) LockOrchestratorProject(projectID domain.ProjectID) func() {
+	m.orchestratorProjectLocksMu.Lock()
+	if m.orchestratorProjectLocks == nil {
+		m.orchestratorProjectLocks = make(map[domain.ProjectID]*sync.Mutex)
+	}
+	mu := m.orchestratorProjectLocks[projectID]
+	if mu == nil {
+		mu = &sync.Mutex{}
+		m.orchestratorProjectLocks[projectID] = mu
+	}
+	m.orchestratorProjectLocksMu.Unlock()
+
+	mu.Lock()
+	return mu.Unlock
+}
+
 func (m *Manager) relaunchRestoredSession(ctx context.Context, rec domain.SessionRecord, project domain.ProjectRecord, ws ports.WorkspaceInfo) (RestoreResult, error) {
 	result, err := m.relaunchSession(ctx, "restore", rec, project, ws, nil)
 	if err != nil {
@@ -2691,6 +2731,13 @@ func (m *Manager) ReconcileBackground(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("reconcile: list sessions: %w", err)
 	}
+	m.reconcileProjectOrchestratorOwnership(ctx, recs)
+	// Ownership repair may have parked a duplicate live controller so a saved
+	// durable owner can restore. Every later pass must consume fresh facts.
+	recs, err = m.store.ListAllSessions(ctx)
+	if err != nil {
+		return fmt.Errorf("reconcile: refresh sessions after orchestrator ownership: %w", err)
+	}
 	m.reconcileLivePass(ctx, recs)
 	for _, rec := range recs {
 		if !rec.IsTerminated {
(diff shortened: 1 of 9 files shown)


backend/internal/session_manager/manager.go
```diff
@@ -2477,6 +2524,132 @@ func (m *Manager) ReconcileBackground(ctx context.Context) error {
...
+			if err := m.chat.StopChat(ctx, current.ID); err != nil {
+				return fmt.Errorf("stop Chat controller: %w", err)
+			}
+		}
+	} else if handle := runtimeHandle(current.Metadata); handle.ID != "" {
+		if err := m.runtime.Destroy(ctx, handle); err != nil {
+			return fmt.Errorf("stop runtime: %w", err)
+		}
+	}
+	if err := m.store.DeleteSessionWorktrees(ctx, current.ID); err != nil {
```
AgentWrapper: For a live session in a multi-repo workspace project these rows are not restore markers: spawn writes one `state = active` row per repo (manager.go:1192) and `workspaceProjectRows` (manager.go:3123) reads them as the inventory `Kill`/`RetireForReplacement` use to destroy child worktrees. Deleting them here while deliberately leaving the worktrees on disk means a later `Kill` of the parked row sees `len(rows) <= 1`, takes the single-repo branch, and only touches `rec.Metadata.WorkspacePath` - the child worktrees leak with no durable record, and the parked session is not recoverable for a workspace project. Consider skipping the delete on this path, or re-stating the rows via `upsertWorkspaceProjectRowState` instead. No park test covers a workspace project.

backend/internal/session_manager/manager.go
```diff
@@ -2541,150 +2714,296 @@ func (m *Manager) reconcileLivePass(ctx context.Context, recs []domain.SessionRe
...
 		}
-		if err := m.restoreAttachments(ctx, rec.ID, ws); err != nil {
-			m.logger.Error("restore-all: restore attachments failed", "sessionID", rec.ID, "error", err)
+		if current, ok := selected[rec.ProjectID]; !ok || restoreRecordNewer(rec, current) {
+			selected[rec.ProjectID] = rec
+		}
+	}
+
+	owners, blocked := m.durableProjectConversationOwners(ctx, candidates)
+	for projectID := range blocked {
```
AgentWrapper, line 3053: `blocked` is applied to every project, but `durableProjectConversationOwners` was called with all saved candidates rather than only contested ones. For a project with a single saved orchestrator there is no ambiguity to protect against, yet a transient conversation-store error (SQLITE_BUSY during a boot where the API is already serving) now silently leaves that project with no orchestrator restored - a change from main. Suggest gating this deletion on the project having more than one saved candidate, matching the contested-only input `reconcileProjectOrchestratorOwnership` uses.

AgentWrapper (commented):
## Review verdict: changes requested

The ownership boundary itself reads well: sharing one project lock between the session service and startup reconciliation, re-reading every ownership fact under the fences before publishing a controller, and preserving native identity plus the shared workspace on the parked duplicate are all the right calls. The `ErrConversationNotFound` alias to `domain.ErrNoConversation` checks out, so the sentinel handling is correct against the real store. Two issues below should be addressed before merge.

### 1. `parkDuplicateOrchestrator` drops the workspace-project repo inventory without destroying the worktrees (`backend/internal/session_manager/manager.go:2642`)

`DeleteSessionWorktrees` is labelled `clear restore marker`, but for a **live** session in a multi-repo workspace project those rows are not markers. Workspace-project spawn writes one `state = active` row per repo (manager.go:1192), and `workspaceProjectRows` (manager.go:3123) reads them as the durable inventory that `Kill` and `RetireForReplacement` need in order to destroy the child worktrees. Parking deletes every row while deliberately leaving the worktrees on disk, so a later `Kill` of the parked row sees `len(rows) <= 1`, falls into the single-repo branch, and only touches `rec.Metadata.WorkspacePath`. The child-repo worktrees are orphaned with no durable record, and the parked session is not actually recoverable history for a workspace project - which is the property the PR description promises. Either skip the delete on the park path or re-state the rows (for example `removed` via `upsertWorkspaceProjectRowState`) instead of deleting them. Neither new park test covers a workspace project, so this shape is untested.

### 2. The fail-closed block is wider than the ambiguity it protects (`backend/internal/session_manager/manager.go:2793`)

`selectSavedOrchestrators` passes **all** saved candidates to `durableProjectConversationOwners`, not just contested ones, and any non-`ErrNoConversation` error marks the whole project `blocked`, which then deletes the timestamp-selected entry. For a project with exactly one saved orchestrator there is nothing to disambiguate, yet a transient store error (SQLITE_BUSY on a boot where the API is already serving) now silently leaves the only orchestrator for that project terminated. That is a behaviour change from main, and `TestRestoreAllLeavesSavedOrchestratorTerminatedWhenConversationOwnershipIsUnknown` locks it in as intended. Suggest applying `blocked` only to projects with more than one saved candidate, matching the contested-only input that `reconcileProjectOrchestratorOwnership` already uses.

### Non-blocking notes

- Durable-owner-beats-recency is unconditional. Because a TUI orchestrator can never own a project Chat conversation, a stale restorable marker on an old Chat orchestrator outranks a newer saved TUI orchestrator on every boot, and the losing marker is never consumed, so the outcome repeats indefinitely. `RetireForReplacement` clearing markers keeps this off the common path, but a warning surfaced to the user would help.
- `reconcileProjectOrchestratorOwnerLocked` logs `reconcile: refresh durable orchestrator owner failed` with the `error` field on the not-found and wrong-kind branches where `err` is nil, so the log reads as a nil error.
- `restoreRecordNewer` duplicates `sessionNewer` in `service/session/service.go`; worth sharing one helper if the tie-break rules are meant to stay identical.

Lock ordering (project lock then agent operation) is consistent across spawn, replacement, explicit restore, and reconcile, so I did not find a deadlock. Docs, the UI message change, and removing the misleading duplicate Restart button all look correct.

backend/internal/session_manager/manager.go
AgentWrapper: Fixed in 071239b12. Parking now preserves every session_worktrees row, including active per-repo workspace-project inventory. Cleanup also takes the project ownership lock, re-reads the terminated row, and refuses to reclaim a canonical workspace while a live orchestrator owns the same path. The production-shaped workspace-project regression now asserts root+child inventory survives parking and a later Cleanup skips the parked duplicate without destroying either worktree.

backend/internal/session_manager/manager.go
AgentWrapper: Fixed in 071239b12. selectSavedOrchestrators now counts candidates per project and only consults durable conversation ownership for projects with multiple saved orchestrators. A sole saved orchestrator restores through a transient conversation-store error; contested projects still fail closed. Regression coverage pins both cases.

AgentWrapper:
Addressed both blockers from review 5056931545 in 071239b12:

- parked workspace-project duplicates retain their complete durable worktree inventory; Cleanup is fenced by the project ownership lock and skips a shared canonical workspace while a live orchestrator owns it
- durable conversation ownership is queried only for contested projects, so a transient read error cannot suppress a sole saved orchestrator while ambiguous projects still fail closed
- Kill is now a true no-op for already-terminated historical rows, preventing repeated kill requests from tearing down a live replacement’s shared canonical worktree

Verification: `go test ./...`, `go test -race ./internal/session_manager ./internal/service/session`, `go build ./...`, `go vet ./...`, and golangci-lint v2.12.2 with a fresh isolated cache (`0 issues`). The ordinary `npm run lint` invocation encountered unrelated host-state/stale golangci-cache failures; its two constituent checks both pass in isolated environments.

backend/internal/session_manager/manager.go
```diff
@@ -3366,6 +3705,32 @@ func (m *Manager) Cleanup(ctx context.Context, project domain.ProjectID) (Cleanu
 // call) — most commonly because a scoped shell terminal could not be
 // confirmed closed, so reclaiming would pull the ground out from under it.
 func (m *Manager) cleanupOne(ctx context.Context, rec domain.SessionRecord, ws ports.WorkspaceInfo) (skipReason string) {
+	if rec.Kind == domain.KindOrchestrator {
+		unlock := m.LockOrchestratorProject(rec.ProjectID)
+		defer unlock()
+
+		// Cleanup runs while the API is serving. Re-read termination under the
```
AgentWrapper, line 3956: This locked re-read is the right idea, but the destructive step it is guarding against sits outside it. `Cleanup` runs `m.runtime.Destroy(ctx, h)` for the terminated snapshot at manager.go:3687, before calling `cleanupOne`. Because tmux handle ids come from `tmuxSessionName(cfg.SessionID)`, a session restored between `cleanupRecords` and that call reuses the same handle, so Cleanup kills the freshly restored agent - and only afterwards does this check return `session is no longer terminated` and preserve the workspace. Consider moving the handle destroy into `cleanupOne` after this re-read (at least for orchestrators), so the whole destructive sequence sits under the ownership lock. Pre-existing on main, so optional here.

backend/internal/session_manager/manager.go
```diff
@@ -1520,6 +1539,14 @@ func (m *Manager) Kill(ctx context.Context, id domain.SessionID) (bool, error) {
 	if !ok {
 		return false, nil // already gone: benign race
 	}
+	// Kill is idempotent. In particular, historical orchestrators can retain the
+	// same canonical workspace identity as their live replacement. Re-running
+	// Kill on that already-terminated row must not tear the shared worktree out
+	// from under the current owner. Cleanup separately refuses reclamation while
+	// a live orchestrator still owns the same path.
+	if rec.IsTerminated {
```
AgentWrapper, line 1640: Non-blocking: this is broader than the shared-workspace hazard it was added for, and it silently changes a public primitive. It is consistent with what `rollbackSpawn` already documents, and the Sidebar filters terminated sessions out of the kill action, so I do not think it regresses the UI. Worth noting that `stopPreviewBestEffort`, `destroyBrowserBestEffort`, and `cleanupSystemPromptDir` no longer run for a terminated row, and `ao session kill` on a terminated id now returns OK with `freed: false` and never reclaims. The doc block above `Kill` still describes it as the way a session gets cleaned up from the dashboard; pointing at `POST /sessions/cleanup` there would keep the contract honest. A narrower alternative is reusing `liveOrchestratorSharesWorkspace` here instead of a blanket return.

AgentWrapper (commented):
## Review verdict: approved

Both blocking findings from the previous round are fixed correctly and covered by tests.

1. **Workspace-project inventory** - `parkDuplicateOrchestrator` no longer deletes the `session_worktrees` rows, and `TestReconcileBackgroundParkDuplicatePreservesWorkspaceProjectInventory` pins root plus child rows surviving the park. Adding `liveOrchestratorSharesWorkspace` to `cleanupOne` is the right complement: reclamation is now refused while a live orchestrator owns the same canonical path, and the `filepath.Clean` comparison is correct.
2. **Fail-closed scope** - `selectSavedOrchestrators` now feeds only contested projects into `durableProjectConversationOwners`, so `blocked` can no longer suppress a project whose sole saved orchestrator needs no disambiguation. `TestRestoreAllRestoresSoleSavedOrchestratorWhenConversationOwnershipReadFails` and the contested counterpart pin both halves.

I also re-checked lock ordering after the new `cleanupOne` lock: `Cleanup` and `TeardownProject` reach `cleanupOne` without holding the project lock, and nothing under that lock re-enters `Cleanup`, so there is no inversion.

### One follow-up worth taking (non-blocking)

`Cleanup` destroys the runtime handle *before* `cleanupOne` performs its locked re-read (manager.go:3686-3689). The new guard covers the worktree but not that destroy. tmux handle ids are derived deterministically from the session id (`tmuxSessionName(cfg.SessionID)`), so a session restored between `cleanupRecords` and that line reuses the same handle: Cleanup kills the freshly restored agent, and `cleanupOne` then correctly reports `session is no longer terminated` and preserves the workspace. The net result is a DB-live orchestrator with a dead runtime until the next reconcile. Moving the handle destroy inside `cleanupOne`, after the locked re-read, closes it. This races the same way on main, so deferring is defensible - but the guard added here is the natural home for the fix.

### Minor notes

- The `Kill` early return on `rec.IsTerminated` is broader than the shared-workspace hazard that motivated it. It does match the contract `rollbackSpawn` already documents (both false means the row was already terminated or absent), and the Sidebar filters terminated sessions out of the trash action, so I do not see a user-facing regression. Two consequences worth knowing: `ao session kill <terminated-id>` now returns OK with `freed: false` and never reclaims, and `cleanupSystemPromptDir` / `stopPreviewBestEffort` / `destroyBrowserBestEffort` no longer run on that path. `POST /sessions/cleanup` is the remaining reclamation route; the `Kill` doc block above the function still frames Kill as the dashboard cleanup path and could say so.
- Carried over from the previous round, still non-blocking: unconditional durable-owner-over-recency, the nil `err` passed to the `error` log field on the not-found and wrong-kind branches of `reconcileProjectOrchestratorOwnerLocked`, and `restoreRecordNewer` duplicating `sessionNewer`.

backend/internal/session_manager/manager.go
AgentWrapper: Addressed in bc4c7427e. Runtime destruction now happens inside cleanupOne, after the project lock and fresh termination/ownership reads for orchestrators. A deterministic regression mutates the row from terminated to live immediately after Cleanup snapshots it and proves neither the reused runtime handle nor workspace is destroyed.

backend/internal/session_manager/manager.go
AgentWrapper: Addressed in bc4c7427e. The Kill contract now explicitly documents that an already-terminated row is idempotent for runtime/workspace reclamation and directs reclamation to Cleanup. The no-op path still clears session-scoped preview, browser, and prompt artifacts; the regression now asserts those are cleaned while the shared workspace inventory and durable row remain untouched.

AgentWrapper:
Updated PR #4591 to current `origin/main` and resolved the `manager_test.go` conflict in merge commit `3d5e138ae`. Follow-up commit `bc4c7427e` addresses both remaining review threads: runtime teardown now occurs after the locked orchestrator re-read, and already-terminated Kill explicitly preserves runtime/workspace while still clearing preview/browser/prompt artifacts. Both threads are replied to and resolved.

Local verification after the merge:
- `go test ./...`
- `go test -race ./internal/session_manager ./internal/service/session`
- `go build ./...`
- `go vet ./...`
- frontend typecheck
- focused SessionsBoard/workspace tests: 124 passed

AgentWrapper:
Rebased the three PR commits onto current `main` (`c02317774`) and force-updated the branch with `--force-with-lease`. Conflicts in Session Manager were resolved by retaining upstream Codex account-switch/startup-reconcile state alongside this PR’s project-orchestrator ownership lock and durable-owner fields. New head: `40036cc82`.

Post-rebase verification:
- `go test ./...`
- `go test -race ./internal/session_manager ./internal/service/session`
- `go build ./...`
- `go vet ./...`
- frontend `npm run typecheck`
- focused SessionsBoard/workspace tests: 124 passed

The PR now reports mergeable with no local changes.

AgentWrapper:
Post-rebase CI is now fully green (10/10 checks), including the 15m12s backend build-test job. Head remains `40036cc82cee0c4ea57dd82956813b1048fea068`.

Issue #4500, opened 2026-08-26
bug(browser-preview): managed preview dev-server crash is invisible in the panel


Pulkit7070:
## Summary
When a managed preview dev-server crashes after launch, the Browser panel keeps showing the stale page with no error; `ao preview status` later reports plain `stopped`.

## Root cause
`waitForExit` flips the run to `StateFailed` (`backend/internal/previewserver/manager.go:499-506`) but does not call `SetPreview("")` and emits no event. The panel's only failure hook, `did-fail-load` (`frontend/src/main/browser-view-host.ts:1828`), fires on navigation, not on an already-rendered page whose backing server dies. After `failedStatusRetention` (5 min) the failed run is deleted (`manager.go:511-518`), so status then reports `stopped` with no error.

## Expected
On a non-stopping exit, clear/mark the session preview so the panel shows a dead-server state.

## Suggested fix
Add an `onExit(sessionID, Status)` callback to `Manager`, call it from `waitForExit` when `!run.stopping`, and wire it to `SetPreview(ctx, id, "")`; SSE already fans session updates to the renderer.

## Notes
UNTESTED live; verified from code. QA audit finding PREV-1. Effort: ~30 min detection, +1-2h for a polished panel state. Scaling risk: high.

Rishet11:
Fixed in [PR #4595](https://github.com/Untrivial-ai/agent-orchestrator/pull/4595).

PR #4595, opened 2026-08-28
fix(preview): clear session preview URL when managed dev server crashes (#4500)

ApexYash11:
## Summary

When a managed preview dev-server crashed after launch, the Browser panel
kept showing the stale page with no error; `ao preview status` later
reported plain `stopped` (issue #4500).

`waitForExit` flipped the run to `StateFailed` but did not call
`SetPreview("")` and emitted no event. The panel's only failure hook,
`did-fail-load`, fires on navigation, not on an already-rendered page
whose backing server died. After `failedStatusRetention` (5 min) the
failed run was deleted, so status then reported `stopped` with no error.

## Fix

- `previewserver.Manager` gains an `OnExit(sessionID, Status)` callback
  fired from `waitForExit` when the run was not user-stopped. It runs
  off the manager's lock with a fresh 5s context.
- `daemon.go` registers an `OnExit` that clears the session's
  `PreviewURL` via `sessionSvc.SetPreview(ctx, id, "")` only when the
  session's current preview URL still matches the dead server's URL, so
  unrelated static-file previews are not clobbered. The existing
  `session_updated` CDC event fans the change out to the renderer.
- The Browser panel is not changed. Issue #4500 calls the polished
  panel state a separate "+1-2h" follow-up; this PR gives the panel
  the necessary session-update signal so it can drop the dead URL on
  its next render.

## Tests

- `TestManagerFiresOnExitWhenServerCrashesAfterLaunch` — boots, becomes
  ready, crashes; callback fires with `StateFailed` and a non-empty
  `Error`.
- `TestManagerDoesNotFireOnExitWhenStoppedByUser` — callback stays
  quiet on `Stop`.

Both new tests pass alongside the existing `previewserver`,
`httpd/controllers` (preview routes), and `service/session`
(`SetPreview`) suites.

## Out of scope

- Polished Browser panel dead-server UI (issue notes this as a
  follow-up).
- `failedStatusRetention` behavior. The callback fires immediately on
  crash; the run is still available via `Status` for 5 min before
  being deleted, matching the existing contract.

### DIFF
--- a/backend/internal/daemon/daemon.go
+++ b/backend/internal/daemon/daemon.go
@@ -292,6 +292,26 @@ func Run() error {
 		}
 		return fmt.Errorf("wire session service: %w", err)
 	}
+
+	// When a managed preview server crashes after launch, the Browser panel's
+	// only failure hook is `did-fail-load`, which fires on navigation, not on an
+	// already-rendered page whose backing server has died. Clear the session's
+	// preview URL so the panel falls off the dead server and the session update
+	// fans out via the existing CDC session_updated event (issue #4500). Only
+	// clear when the session's current preview URL still points at the failed
+	// server, so an unrelated static-file preview isn't clobbered.
+	managedPreview.SetOnExit(func(ctx context.Context, id domain.SessionID, status previewserver.Status) {
+		sess, err := sessionSvc.Get(ctx, id)
+		if err != nil {
+			return
+		}
+		if status.URL == "" || sess.Metadata.PreviewURL != status.URL {
+			return
+		}
+		if _, err := sessionSvc.SetPreview(ctx, id, ""); err != nil {
+			log.Warn("clear preview URL after managed preview crash", "session", id, "err", err)
+		}
+	})
 	sessMgr.SetTerminalInputGate(termMgr)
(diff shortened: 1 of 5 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

Pulkit7070:
The core fix is sound and the URL-match guard correctly avoids clobbering an unrelated or restarted preview. Two things worth addressing:

- The OnExit callback swallows a `sessionSvc.Get` failure silently (`return` with no log), while the `SetPreview` failure just below it logs. If Get fails, the preview URL is never cleared and there is no diagnostic, so the Browser panel stays pointed at the dead server with nothing explaining why. Log the Get error at warn like the SetPreview path.
- The guard that makes this fix correct (clear only when the session's PreviewURL still equals the failed server's URL) lives in the daemon and has no test. The manager tests prove the callback fires, but not that an unrelated or already-restarted preview is left alone. A test around the daemon callback would protect the part most likely to regress.

ApexYash11:
@Pulkit7070  Re-Review!!

Pulkit7070 (approved):

PR #4598, opened 2026-08-28
feat: add failure-only observability for agent switching

whoisasx:
## Summary

- add a complete failure taxonomy and durable failure-only outbox for async agent switching
- cover admission, lifecycle, runtime, recovery, delivery, storage, policy, and frontend visibility branches
- add strict consent generations, purge/drain ordering, ownership-based duplicate suppression, throttling, and privacy-safe Sentry envelopes
- add operator runbooks for all 71 failure points

Successful switches emit no Sentry reports. Production capture remains disabled until the documented privacy and Windows durability gates are approved.

## Complete design package

The complete design—including the original one-layer tree, the expanded end-to-end decision tree, all 71 failure points, subsystem coupling, suppression rules, consent flow, privacy boundary, severity model, and release gates—is attached in [this PR comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4598#issuecomment-5454529824).

Committed source-of-truth documents:

- [Approved design](https://github.com/Untrivial-ai/agent-orchestrator/blob/codex/agent-switch-failure-observability/docs/superpowers/specs/2026-08-28-agent-switch-failure-observability-design.md)
- [Implementation plan](https://github.com/Untrivial-ai/agent-orchestrator/blob/codex/agent-switch-failure-observability/docs/superpowers/plans/2026-08-28-agent-switch-failure-observability.md)
- [Runbook for every failure point](https://github.com/Untrivial-ai/agent-orchestrator/blob/codex/agent-switch-failure-observability/docs/runbooks/agent-switch-failure-points.md)
- [Telemetry and privacy contract](https://github.com/Untrivial-ai/agent-orchestrator/blob/codex/agent-switch-failure-observability/docs/telemetry.md)

## Behavioral changes included in this PR

This is not an observability-only change. Correct failure classification requires the switching code to distinguish a write that definitely failed from a response that was lost after another actor committed progress, and to avoid starting or restoring a runtime whose ownership cannot be recovered. The PR therefore includes deliberate runtime and policy changes.

### 1. Idempotent target-acknowledgement readback

Target acknowledgement now uses an authoritative readback in both the session-manager path (`acknowledgeAgentSwitchTargetWithReadback`) and the lifecycle path (`acknowledgeAgentSwitchTarget`). If the acknowledgement write returns an error, reports `changed=false`, or races with another transition, AO reloads the durable switch before deciding the outcome.

- If the exact target generation is durably acknowledged, the acknowledgement is accepted even when the original write response was lost.
- If the switch is already terminal or has left the expected `delivering`/target-generation predicate, the callback is treated as stale or superseded rather than as evidence that the active switch failed.
- If the same delivery predicate is still current and the acknowledgement is absent, AO returns a real acknowledgement error; it does not infer success.

This prevents a committed acknowledgement or a concurrent winning transition from being misreported as a new switch failure while preserving failures where the durable acknowledgement genuinely did not stick.

### 2. Lost/stale Chat acknowledgement semantics

The Chat continuation path now yields successfully and marks its call stale when acknowledgement readback proves that the switch has already moved beyond the callback's `delivering` state or target generation. A delayed callback can therefore no longer terminalize, retry, or emit a failure for a switch whose authoritative state has already advanced elsewhere.

This is intentionally narrower than “ignore acknowledgement errors”: AO only yields when durable readback proves that the callback's predicate is obsolete. An unacknowledged switch that remains in the same delivery predicate still follows the normal failure path.

### 3. ConPTY restart-registry reservation and partial-effect evidence

On Windows, ConPTY `Create` now reserves the restart-registry record before spawning `pty-host`. Failure to create that reservation fails `Create` before a child process can be started. If spawning subsequently fails, AO rolls the reservation back.

If reservation removal itself fails, `Create` now returns the retained handle as `effect=possible` and `cleanup=failed`. The session manager therefore keeps the recovery gate and can perform exact cleanup; it does not incorrectly restore the source while a reservation or process may still exist.

The restart registry is a hard prerequisite for Windows session creation. AO must not create a live terminal process that is absent from crash-recovery state and cannot be safely identified, fenced, or reclaimed after daemon restart.

### 4. Conservative tmux ownership after preserved-shell/manual relaunch

An existing tmux pane without the expected AO supervisor is no longer treated as proof that the fenced generation is dead. Because the preserved shell may contain a manually relaunched workload, the probe returns `unknown/identity_missing` unless exact death is affirmatively established. Recovery retains the gate instead of restoring the source and risking two live owners.

### 5. Legacy recovery-marker migration

The observability migration is now `0119` after merging the latest `main`. During the table rebuild, legacy rows stored as `failed/source_stop_unconfirmed` are normalized to the retained-recovery state `stopping_source/source_stop_unconfirmed`. Existing databases with this valid legacy marker can upgrade without violating the new constraint, and the marker remains available for safe reconciliation.

### 6. Opt-out is a provider-call barrier, not only a capture flag

Sentry capture uses a synchronous transport behind a policy-controlled intake gate. Disabling reporting first closes intake, drains dispatcher work and every already-entered provider call, then acknowledges the off generation. Once opt-out completes, no queued SDK event can begin a later provider request. The failure-reporting preference remains separate from existing PostHog product analytics: disabling agent-switch failure reports neither disables PostHog nor purges its queues.

### 7. Provider and filesystem boundaries remain adapter-owned

The core domain owns normalized failure vocabulary, validation, and severity only. Sentry JSON/envelope encoding now lives in the Sentry adapter and is injected through an encoder port. Authority-file I/O, symlink/mode checks, bounded JSON parsing, and validation live in a filesystem adapter injected through an authority-reader port. Safety-critical policy, lease, settlement, expiry, receipt, and backlog statements use generated sqlc queries so authorization predicates have one maintained source.

### Why these changes remain together

These behaviors provide the state-certainty and ownership guarantees used by the failure-only classifiers: committed-response-lost versus genuine write failure, stale callback versus active failure, and recoverable runtime versus unidentifiable process. Keeping them in this PR avoids an intermediate implementation where observability would classify ambiguous outcomes without the durable readback and fencing semantics it relies on. They are called out separately here so reviewers and future revert work can identify their runtime impact independently from the reporting pipeline.

## Verification

- merged latest `main` and resolved the migration and frontend conflicts
- targeted backend runtime, domain, policy, Sentry adapter, session-manager, SQLite/store, and daemon suites passed
- frontend TypeScript check passed
- focused frontend shared/renderer telemetry suites passed (67 tests)
- sqlc regeneration completed; generated artifacts are committed
- `git diff --check` and conflict-marker scan passed


whoisasx:
# Complete agent-switch failure-only observability design

This comment attaches the complete design package discussed before implementation: the original one-layer tree, the final expanded decision tree, all 71 stable failure points, subsystem coupling, reporting ownership, suppression, recovery, consent, delivery, privacy, and release gates.

The full committed documents are the source of truth:

- [Approved design](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c938ad514cf0492eb023b5d88ee253dc1aaf3d3/docs/superpowers/specs/2026-08-28-agent-switch-failure-observability-design.md)
- [Implementation plan](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c938ad514cf0492eb023b5d88ee253dc1aaf3d3/docs/superpowers/plans/2026-08-28-agent-switch-failure-observability.md)
- [Operator runbook for every failure point](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c938ad514cf0492eb023b5d88ee253dc1aaf3d3/docs/runbooks/agent-switch-failure-points.md)
- [Telemetry and privacy contract](https://github.com/Untrivial-ai/agent-orchestrator/blob/0c938ad514cf0492eb023b5d88ee253dc1aaf3d3/docs/telemetry.md)

## 1. Original one-layer tree

This was the initial high-level view. It established the complete lifecycle before expanding every fallible boundary and subsystem interaction.

```text
Agent switch request
├── Admission and idempotency
├── Preflight and handoff
├── Stop or quiesce source agent
├── Build and publish continuation context
├── Start and activate target agent
│   ├── TUI runtime
│   └── Chat provider/controller
├── Deliver continuation context
├── Commit completion
├── Compensate or recover when ownership is uncertain
├── Reconcile after daemon restart
├── Handle panic, shutdown, and post-terminal maintenance
├── Make durable state visible in the frontend
└── Enroll and deliver a failure-only Sentry report
```

The key rule was already fixed here: **success is silent**. Successful switches, normal progress, expected rejection, idempotent replay, stale callbacks, and recovered transient failures do not create a Sentry event.

## 2. Final complete decision tree

```text
START: switch request arrives
│
├── A. Request/admission ownership
│   ├── Expected validation or active-switch conflict
│   │   └── Render normal 4xx; no switch report
│   ├── Matching idempotency key and fingerprint
│   │   └── Return existing saga; no new report
│   └── Try to create/arm the durable saga
│       ├── Durable row proven present
│       │   └── Saga owns all later reporting
│       ├── Create returned an error
│       │   ├── Read-back finds exact row
│       │   │   └── Treat as committed-response-lost; saga owns reporting
│       │   ├── Read-back conclusively proves no row
│       │   │   └── HTTP owns the unexpected pre-admission 5xx
│       │   └── Read-back unavailable/ambiguous
│       │       └── Retain any gate; generic HTTP durability incident only
│       ├── Chat handoff boundary cannot be armed
│       │   └── Classify admission_chat_handoff_arm
│       └── Background worker cannot start
│           └── Classify worker_start_refused
│
├── B. Preflight and handoff
│   ├── Preserve exact source native identity/generation
│   ├── Probe target binary, auth, configuration, and resume candidate
│   ├── Prepare handoff directory
│   ├── Collect and durably settle the semantic handoff
│   ├── Close source input / interrupt handoff / quiesce Chat controller
│   ├── Prepare target launch gate
│   └── Commit stopping_source before performing the stop side effect
│       ├── Any safe pre-stop failure
│       │   └── Source remains owner; terminal failure may report once
│       └── Ambiguous commit/read-back
│           └── Retain gate; never continue to target creation
│
├── C. Stop source
│   ├── TUI: destroy exact fenced runtime generation
│   ├── Chat: stop and drain exact controller generation
│   ├── Probe/read back the side effect
│   │   ├── Exact source alive
│   │   │   └── Preserve source; do not create target
│   │   ├── Exact source dead
│   │   │   └── Commit source_stopped and continue
│   │   └── Unknown
│   │       └── Persist source_stop_unconfirmed; retain operation gate
│   └── Commit/read back source_stopped
│       ├── Commit proven
│       │   └── Continue
│       └── Outcome ambiguous
│           └── Retain gate and reconcile; never infer success from timeout
│
├── D. Construct continuation and artifacts
│   ├── Refresh source metadata
│   ├── Verify semantic artifact
│   ├── Capture source transcript into the local safe artifact path
│   ├── Build continuation
│   ├── Publish, verify, and commit final artifact
│   ├── Prepare target prompt
│   └── Prepare target workspace
│       ├── Failure before source stop
│       │   └── Source owns; safe retry/failure
│       └── Failure after source stop
│           └── Compensate only after exact target absence; otherwise hold
│
├── E. Start target
│   ├── E1. TUI/common runtime path
│   │   ├── Prepare and commit target native identity
│   │   ├── Create runtime
│   │   │   ├── Handle only
│   │   │   │   └── Probe exact generation before ownership decision
│   │   │   ├── Handle plus error / response lost
│   │   │   │   └── Treat effect as unknown; read back and fence
│   │   │   └── Empty handle after possible effect
│   │   │       └── Possible unidentifiable target; P0, retain gate
│   │   ├── Commit handle and generation
│   │   ├── Wait for exact native identity
│   │   └── Atomically commit/read back target activation tuple
│   │       ├── Exact target tuple committed
│   │       │   └── Target owns; source restoration forbidden
│   │       ├── Exact target absent/dead
│   │       │   └── Fenced cleanup, then one source restore attempt
│   │       └── Unknown or mismatched generation
│   │           └── Persist target_start_unconfirmed; retain gate
│   │
│   └── E2. Chat provider/controller path
│       ├── Start or resume provider boundary
│       ├── Commit native provider identity
│       ├── Commit deterministic provider boundary
│       ├── Atomically commit session, conversation, switch, and generation
│       ├── Read back complete activation tuple
│       └── Publish target controller
│           ├── Published after durable target ownership
│           │   └── Target owns; source restoration forbidden
│           └── Publication/ownership mismatch
│               └── Possible dual controller owner; P0 and retain gate
│
├── F. Deliver continuation and complete
│   ├── Commit delivery-open boundary
│   ├── TUI
│   │   ├── Wait for target hook
│   │   └── Commit exact acknowledgement
│   ├── Chat
│   │   ├── Relay one deterministic continuation message
│   │   └── Commit exact target acknowledgement
│   └── Commit completed
│       ├── Acknowledgement/completion wins
│       │   └── Success: discard flight recorder; no report
│       ├── Failure CAS wins
│       │   └── Terminal delivery_unconfirmed report; never blindly replay
│       └── Commit response lost
│           └── Read back before choosing success or failure
│
├── G. Compensation and recovery
│   ├── Is a target possibly alive?
│   │   ├── Yes or unknown
│   │   │   └── Never restore source; retain gate and exact fenced cleanup
│   │   └── Exact absence/death proven
│   │       └── Cleanup target workspace/runtime, then restore source once
│   ├── Source restore
│   │   ├── Exact restore proven
│   │   │   └── Terminalize with safe ownership
│   │   └── Failed/unknown
│   │       └── Persist source_restore_unconfirmed; retain gate
│   └── Startup or explicit reconciliation by durable phase
│       ├── preparing_handoff
│       │   └── Source owns; close daemon_restart_pre_stop
│       ├── stopping_source
│       │   └── Exact alive/dead/unknown fenced probe matrix
│       ├── source_stopped
│       │   └── Clean target prep, restore source once, or retain marker
│       ├── starting_target
│       │   └── Adopt exact live target, clean exact dead target, hold unknown
│       ├── target_ready
│       │   ├── TUI: do not replay argv-bound continuation
│       │   └── Chat: relay only the deterministic never-attempted message
│       ├── delivering_context
│       │   └── Ack completes; no ack becomes delivery_unconfirmed
│       └── completed/failed
│           └── No switch side effects and no second semantic report
│
├── H. Panic, shutdown, and maintenance
│   ├── Live/recovery worker panic
│   │   ├── One bounded reconciliation creates a durable classification
│   │   │   └── One incident owned by that classification
│   │   └── No new durable classification
│   │       └── One operational panic event for the execution attempt
│   ├── Shutdown worker timeout
│   │   └── One daemon-level aggregate; not one event per switch
│   └── Post-terminal artifact cleanup failure
│       └── Separate maintenance event; terminal switch state is unchanged
│
├── I. Frontend visibility
│   ├── Renderer sees durable state successfully
│   │   └── No report
│   ├── SSE transport unhealthy
│   │   ├── Refresh succeeds before 15s active / 60s history grace
│   │   │   └── No report
│   │   └── Focused, online owner remains unhealthy through grace
│   │       └── visibility_transport; query incident suppressed
│   ├── Typed query fails while transport is healthy
│   │   ├── Query recovers before grace
│   │   │   └── No report
│   │   └── Failure persists through grace
│   │       └── visibility_query
│   └── Durable failure/recovery presentation expected
│       ├── Actual component commits and acknowledges within 2s
│       │   └── No report
│       ├── User dismisses, navigates, blurs, goes offline, or row supersedes
│       │   └── Cancel timer; no report
│       └── Same focused route remains but presentation never commits
│           └── visibility_presentation
│
└── J. Failure-only reporting pipeline
    ├── Winning durable mutation?
    │   ├── CAS changed zero rows / repeated marker / terminal row
    │   │   └── No enrollment and no report
    │   ├── Ordinary progress or completed
    │   │   └── No enrollment and no report
    │   └── Failed transition or first recovery marker
    │       └── Persist local failure_point regardless of consent
    ├── Reporting policy disabled or stale generation
    │   └── No receipt and no payload; core saga still commits
    ├── Reporting policy enabled with exact generation and destination
    │   ├── Existing receipt
    │   │   └── Deduped; no new payload
    │   └── New incident
    │       └── Savepoint validates and serializes allowlisted event
    │           ├── Success
    │           │   └── Atomically add payload-free receipt + outbox payload
    │           └── Taxonomy/serialization/local observability failure
    │               └── Roll back telemetry savepoint only; commit core saga
    ├── Dispatcher claims eligible unexpired row
    │   ├── Sentry accepted
    │   │   └── Mark delivered; same EventID makes crash retry stable
    │   ├── Transient failure / scoped 429
    │   │   └── Release lease and retry with capped backoff/throttle
    │   ├── Permanent failure
    │   │   └── Quarantine/discard without touching switch ownership
    │   ├── Consent disabled or generation/destination changed
    │   │   └── Abort, purge payloads, retain payload-free receipts
    │   └── Payload older than seven days
    │       └── Expire and delete; never send later
    └── Outbox delivery itself fails
        └── Local-only outbox_delivery diagnostic; never recursively report
```

## 3. Stable failure-point taxonomy: all 71 points

```text
Admission (4)
├── admission_saga_create
├── admission_commit_readback
├── admission_chat_handoff_arm
└── worker_start_refused

Preflight and handoff (11)
├── source_native_preserve
├── target_preflight
├── target_resume_lookup
├── handoff_directory_prepare
├── handoff_collection
├── handoff_settlement
├── decision_input_close
├── source_handoff_interrupt
├── chat_source_quiesce
├── target_launch_gate_prepare
└── stopping_source_commit

Source stop (6)
├── source_runtime_destroy
├── source_runtime_probe
├── source_controller_stop
├── source_controller_drain
├── source_stop_commit
└── source_stop_readback

Context and artifact construction (9)
├── source_metadata_refresh
├── semantic_artifact_verify
├── source_transcript_capture
├── continuation_build
├── final_artifact_publish
├── final_artifact_verify
├── final_artifact_commit
├── target_prompt_prepare
└── target_workspace_prepare

Target start: common and TUI (8)
├── target_native_prepare
├── target_native_commit
├── target_runtime_create
├── target_handle_commit
├── target_generation_probe
├── target_native_identity_wait
├── target_activation_commit
└── target_activation_readback

Target start: Chat (7)
├── chat_provider_start
├── chat_provider_resume
├── chat_native_identity_commit
├── chat_provider_boundary_commit
├── chat_target_activation_commit
├── chat_target_activation_readback
└── chat_controller_publish

Delivery and completion (6)
├── delivery_open_commit
├── tui_target_hook_wait
├── tui_target_ack_commit
├── chat_continuation_relay
├── chat_target_ack_commit
└── completion_commit

Compensation and recovery (11)
├── target_runtime_cleanup
├── target_workspace_cleanup
├── source_runtime_restore
├── source_controller_restore
├── recovery_session_load
├── recovery_runtime_probe
├── recovery_native_identity
├── recovery_artifact_verify
├── recovery_activation
├── recovery_settlement
└── recovery_existing_marker

Process, maintenance, visibility, and local-only (9)
├── live_worker_panic
├── recovery_worker_panic
├── shutdown_worker_timeout
├── terminal_artifact_cleanup
├── visibility_transport
├── visibility_query
├── visibility_presentation
├── outbox_delivery               [local-only; prevents recursion]
└── classification_unknown        [local-only; invalid classifier sentinel]
```

Each point maps table-wise to subsystem, valid durable phases, valid error/fault codes, report kind or explicit local-only action, default severity, title, classifier callsite, and a dedicated runbook anchor.

## 4. Decision dimensions applied at every fallible boundary

Complete coverage is the Cartesian product:

```text
phase
× subsystem
× call outcome
× durable state
× real ownership
× compensation result
× user impact
× reporting action
```

Every boundary is classified as exactly one of:

```text
ok
expected_rejection
no_effect_failure
committed_response_lost
effect_unknown
stale
timed_out
cancelled
panic
cleanup_failed
```

Ownership decision:

```text
source
├── preserve source; safe retry may be allowed
none
├── source restoration only after exact target absence
target
├── continue in target; source restoration forbidden
ambiguous
└── retain operation gate; no second target and no forced restore
```

Compensation decision:

```text
not_needed
succeeded
failed
uncertain
```

The switch/session rows remain the authority. Sentry is diagnostic only and never repairs or mutates ownership.

## 5. Coupling with other subsystems

| Subsystem | What can fail | Contract with the switch saga | Reporting owner |
| --- | --- | --- | --- |
| HTTP controller/API envelope | admission, validation, serialization | Expected 4xx is silent; pre-admission unexpected 5xx remains HTTP-owned; safe `reporting_owner` crosses API | HTTP before admission, saga after admission |
| Session Manager | phase decisions, CAS races, compensation | Owns typed classification and the safe enum-only flight recorder | Saga |
| SQLite store | durable state, CAS, policy, outbox | Core mutation and eligible enrollment share a transaction/savepoint; telemetry failure never vetoes safe saga state | Saga/outbox |
| TUI runtime/tmux/ConPTY | create, destroy, liveness, handle identity | Must return exact fenced alive/dead/unknown; unknown is never absence | Saga |
| Chat provider/controller | start/resume, boundary, generation, publish | Activation commits full target tuple atomically before publication | Saga |
| Lifecycle hooks | hook arrival, timeout, duplicate/stale callback | Exact generation and opposing-predicate CAS; stale/duplicate is silent | Saga |
| Workspace/artifact filesystem | prepare, publish, verify, cleanup | No raw path/content leaves device; cleanup cannot authorize unsafe ownership change | Saga or maintenance class |
| Recovery/reaper/startup | phase reconciliation | Converges through the same terminal/marker paths; wrapper does not recapture | Saga |
| Daemon shutdown | worker drain timeout | One daemon aggregate; ordinary cancellation silent | Daemon lifecycle class |
| Electron main | consent authority, visibility FSM, sender | Sole desktop policy writer and sole visibility-report network owner | Electron main |
| Renderer/preload | query/transport/presentation signals | Typed IPC only; local token/route/switch/revision stripped before event construction | Never reports durable saga failure |
| Sentry adapter | envelope validation and delivery | Synthetic allowlisted event, no raw wrapped error, content, identifier, or automatic context | Provider-neutral observer → Sentry |

## 6. Report classes and exact suppression

Only these report kinds can leave the device:

```text
terminal_failure
recovery_required
panic
recovery_attempt_failed
maintenance_failure
daemon_lifecycle_failure
visibility_failure
```

These never create a new event:

```text
successful switch or successful presentation
ordinary progress transition
expected validation or conflict
matching idempotent replay
repeated recovery marker
terminal row observed again
commit error followed by read-back proving success
side-effect error followed by exact proof that the intended effect occurred
stale or duplicate lifecycle hook
failure CAS that loses to acknowledgement/completion
startup wrapper around an already-owned incident
ordinary shutdown cancellation
successful cleanup retry
normal rendering of a durable failed row
CLI or renderer polling timeout when the daemon already owns the incident
```

## 7. Consent and opt-out ordering

```text
Disable requested
├── Close Electron-main visibility/capture gate
├── Abort and drain in-flight main deliveries
├── Ask daemon to close its in-memory gate
├── Durably write a fresh off consent generation under ~/.ao
├── Apply exact off generation to daemon SQLite policy
├── Drain dispatcher and purge every outbox payload status
├── Close/drain desktop transport and clear its cache
├── Ask every trusted renderer to clear PostHog batch, retry, and pre-DOM queues
└── Acknowledge only after every required cleanup proves completion
    ├── Complete
    │   └── state=applied, events_enabled=false
    └── Any missing/failed proof
        └── state=cleanup_pending/cleanup_failed; reporting stays closed
```

Opt-in never backfills historical terminal failures. Only a currently active unresolved recovery marker may enroll at its opt-in observation time; it cannot reuse a pre-consent occurrence timestamp or duration.

## 8. Remote event privacy boundary

Allowed remote fields are closed enums/booleans, bounded buckets, sanitized in-app frames, release/environment/channel/platform/OS, and the classifier callsite. Application-supplied remote bytes must never contain:

```text
switch/session/project/workspace/user IDs
titles, branches, issues, PR or repository facts
idempotency keys or request fingerprints
model/configuration/argv/environment values
native conversation/generation/provider IDs or runtime handles
prompts, messages, handoff/transcript/hook payloads, terminal output
paths, artifact hashes, URLs, HTTP bodies, raw errors, panic values, locals
renderer token, route, switch revision, sender/window identity, consent generation
```

Outbox payloads have a hard seven-day TTL. Payload-free dedupe receipts may outlive opt-out only under the documented incident-lifetime rules. Source IP remains provider-visible transport metadata and requires Sentry-organization privacy approval before production enablement.

## 9. Severity and release gates

```text
P0 / fatal
└── possible dual owner, unidentifiable live target, activation ownership unknown,
    controller published without durable ownership, gate without saga, or unsafe cleanup

P1 / error
└── no provable live owner, restore/delivery unconfirmed, startup recovery blocked,
    post-stop panic, unrecoverable target/context

P2 / warning
└── safe pre-stop abort, missing binary/auth/config, source safely preserved/restored,
    proven-harmless maintenance or visibility failure

Suppressed
└── expected/replayed/stale/recovered/successful cases
```

Production reporting remains deliberately disabled. It can be enabled only after exact fenced probe behavior, atomic Chat activation, single-owner dedupe, complete opt-out proof, Windows crash-durable replacement, Sentry IP/residency/retention approval, and the privacy allowlist verification all pass.

This is the complete intended behavior represented by the implementation in this PR. The runbook linked above gives the expected phase, safe ownership interpretation, restore permission, logs/tests, user action, and release-blocker condition for each of the 71 failure points.

Pulkit7070:
This PR bundles several real behavior changes under a failure-only observability title, which makes the ~20,320-line diff hard to review as a single unit:

- Target acknowledgement now goes through an idempotent readback (`acknowledgeAgentSwitchTargetWithReadback`, and the lifecycle-side `acknowledgeAgentSwitchTarget`), changing how a non-committing ack is resolved.
- The chat switch now yields (returns success and marks the call stale) when readback shows the switch has left the `delivering`/generation predicate, rather than treating that as an error.
- conpty `Create` now reserves the restart registry before spawning and fails `Create` when that reservation fails, making the registry a hard prerequisite for session creation on Windows.

Each of these checks out as correct on its own: the readback paths distinguish "superseded elsewhere" from "our write did not stick," and the conpty ordering has a documented crash-recovery rationale with a clean rollback. The concurrency paths hold under the race detector at high repetition, and the Sentry envelope stays allowlist/enum-only and consent-gated.

The ask here is scope, not a code change: separating the behavior changes (ack idempotency, conpty ownership ordering, lost-ack semantics) from the pure observability additions would make each reviewable and revertable independently. If that split is impractical, calling these out in the PR description would at least signal to future readers that this change is not observability-only.

whoisasx:
Thanks @Pulkit7070 — agreed that the PR title did not make these runtime semantics sufficiently visible.

I updated the PR description with a dedicated **Behavioral changes included in this PR** section covering:

- the idempotent target-acknowledgement readback and its exact success/stale/failure predicates;
- the Chat path's lost/stale acknowledgement behavior; and
- ConPTY's restart-registry reservation before process spawn, including its failure and rollback behavior.

I also documented why they remain coupled here: these are the state-certainty and ownership boundaries the failure-only classifiers rely on to distinguish committed-response-lost from a real write failure, a stale callback from an active failure, and a recoverable Windows runtime from an unidentifiable process. The section explicitly notes that this is not an observability-only PR so the runtime impact is visible for review and future revert work.

No code was changed in response to this scope comment.

illegalcall (commented):
I found eight actionable issues in the current head: five behavioral/spec gaps and three architecture/maintenance problems. The migration, ConPTY rollback, tmux ownership, and opt-out queue findings are merge-blocking because they can prevent upgrades, strand recovery, create dual ownership, or violate the promised opt-out boundary. Inline comments include suggested fixes and missing regression coverage.

backend/internal/storage/sqlite/migrations/0124_agent_switch_failure_observability.sql
```diff
@@ -0,0 +1,225 @@
...
+);
+
+INSERT INTO agent_switches_next (
+    id, session_id, idempotency_key, request_fingerprint, from_harness, target_harness,
+    target_native_session_ref, target_start_mode, state, agent_handoff_status,
+    source_transcript_status, semantic_handoff_included, agent_handoff_path, agent_handoff_hash,
+    source_generation_id, target_generation_id, target_runtime_handle_id, target_acknowledged_at,
+    error_code, failure_point, requested_at, updated_at, final_handoff_path, final_handoff_hash
+)
+SELECT
```
illegalcall: [P1] Preserve legacy `source_stop_unconfirmed` rows during this rebuild

Before 0117, the switch manager could legitimately persist `state = 'failed'` with `error_code = 'source_stop_unconfirmed'`. The new CHECK at lines 57–63 rejects that combination, but this `INSERT ... SELECT` copies it unchanged. SQLite will abort the migration, preventing any affected existing database from opening.

Please explicitly normalize the legacy state during the copy (for example, to the new allowed recovery state) and add a migration test that creates the pre-0117 schema, inserts this legacy row, and then runs 0117. The current migration test starts at the new schema and cannot catch this upgrade failure.

backend/internal/adapters/runtime/conpty/runtime.go
```diff
@@ -85,50 +115,132 @@ func New(opts Options) *Runtime {
...
 	if err != nil {
+		cause := fmt.Errorf("conpty: spawn pty-host for %q: %w", id, err)
+		handle := ports.RuntimeHandle{ID: id}
+		if addr == "" && pid == 0 {
+			if unregisterErr := r.unregisterHost(id); unregisterErr != nil {
+				cause = errors.Join(cause, fmt.Errorf("remove unused pty-host reservation for %q: %w", id, unregisterErr))
+				// Keep the current-owner reservation in memory when durable
+				// cleanup fails. A later Destroy can safely retry unregistering
+				// it without spawning or killing any process.
+				return ports.RuntimeHandle{}, conptyCreateFailure(cause)
```
illegalcall: [P1] Return cleanup-capable evidence when reservation removal fails

At this return, `r.sessions[id]` still contains the PID-zero `currentOwner` reservation and durable unregistering has failed. `conptyCreateFailure` nevertheless reports an empty `PossibleHandle`, `RuntimeEffectNone`, and `RuntimeCleanupNotAttempted`. The manager interprets that as no possible runtime effect and skips `Destroy`; its source-restoration `Create` then fails with `session already exists` because this reservation is still present.

Return the session handle with `RuntimeEffectPossible`/`RuntimeCleanupFailed`, or complete/retry reservation removal before reporting no effect. Please cover the production manager rollback path with ConPTY, not only the manual `Destroy(RuntimeHandle{ID: ...})` case in the adapter test.

backend/internal/adapters/runtime/tmux/tmux.go
```diff
@@ -556,6 +583,40 @@ func (r *Runtime) IsAlive(ctx context.Context, handle ports.RuntimeHandle) (bool
...
+		if isSupervisorCommand(entry.command, string(ref.SessionID), ref.Generation) {
+			exactSupervisorFound = true
+			continue
+		}
+		return ports.FencedProbeResult{Liveness: ports.FencedUnknown, Reason: ports.FencedReasonGenerationMismatch}
+	}
+	if exactSupervisorFound && containsExactSupervisedWorkload(entries, panePID, string(ref.SessionID), ref.Generation) {
+		return ports.FencedProbeResult{Liveness: ports.FencedAlive, Reason: ports.FencedReasonExactMatch}
+	}
+	return ports.FencedProbeResult{Liveness: ports.FencedDead, Reason: ports.FencedReasonExactAbsent}
```
illegalcall: [P1] Do not convert ambiguous live tmux ownership into exact death

This branch proves only that the expected AO supervisor/workload was not found. The tmux session and pane still exist, and a preserved shell may contain a manually relaunched workload whose identity is not recognizable as the exact generation. The fenced-probe contract requires missing identity or command ambiguity to return `FencedUnknown`.

Returning `FencedDead` allows recovery to restore the source and can create dual ownership. Please return unknown unless there is affirmative evidence that the exact fenced target is dead, and add a fenced-probe regression test for the preserved-shell/manual-relaunch case already modeled by the ordinary liveness tests.

backend/internal/observe/sentryobs/sentryobs.go
```diff
@@ -77,6 +78,11 @@ func Init(cfg Config) error {
 // Enabled reports whether Sentry is active (a DSN was configured).
 func Enabled() bool { return enabled.Load() }
 
+// SetPolicyEnabled mirrors the crash-durable telemetry authority. Capture
+// paths consult it synchronously so a watcher-observed opt-out closes intake
+// even when the SDK itself was initialized earlier in the process.
+func SetPolicyEnabled(value bool) { policyEnabled.Store(value) }
```
illegalcall: [P1] Opt-out must also cancel or drain the SDK transport queue

This atomic gate blocks future capture calls, but sentry-go's default async transport can already contain queued events. Those events are flushed later by `Flush` during shutdown, outside `PolicyCoordinator.closeGateAndDrain`, so opt-out can acknowledge completion and a provider request can still begin afterward. That violates the stated `no new provider call after completion` boundary.

Please make the SDK transport part of the opt-out barrier—either cancel/purge and drain it before acknowledging, or use a policy-controlled transport whose queue can be closed safely. A regression test should queue an event before opt-out, hold the provider transport, complete opt-out, and prove that no later request starts.

frontend/src/renderer/lib/telemetry.ts
```diff
@@ -735,15 +736,18 @@ export async function initTelemetry(): Promise<boolean> {
 		// Null means the supervisor withheld it: no key, no data dir, or an
 		// unpackaged build that has not opted in. The client is never created.
 		if (!bootstrap) return false;
+		if (!bootstrap.eventsEnabled) return false;
```
illegalcall: [P2] Keep staged failure reporting separate from existing PostHog analytics

`bootstrap.eventsEnabled` represents the new event-reporting preference for staged agent-switch diagnostics, but this return also prevents PostHog product analytics from initializing. `clearRendererTelemetryQueues` similarly opts PostHog out and purges its queues. This contradicts `docs/telemetry.md`, which says existing product telemetry remains governed by the existing environment configuration.

Please give the failure-reporting path its own gate and leave PostHog initialization/queues under the existing product-telemetry authority, or explicitly redefine and document this setting as unified telemetry consent across every surface.

backend/internal/domain/agent_switch_observability.go
```diff
@@ -0,0 +1,1473 @@
...
+	}
+	if len(raw) > agentSwitchStackMaxBytes {
+		return errors.New("sanitized stack exceeds 16 KiB")
+	}
+	return nil
+}
+
+// BuildAgentSwitchCanonicalEvent validates input and returns the bounded,
+// deterministic provider payload.
+func BuildAgentSwitchCanonicalEvent(input AgentSwitchEventBuildInput) ([]byte, error) {
```
illegalcall: [P1] Keep Sentry's wire format out of the domain package

This domain function builds provider-specific JSON, including Sentry exception/stacktrace, fingerprint, tags, and contexts. That reverses the repository's port-based boundary: `domain/` should contain the normalized fault vocabulary, while concrete provider encoding belongs to an adapter. It also makes a Sentry schema change a core-domain change.

Please keep validation and the normalized failure event in `domain`, then inject a canonical encoder at the outbox/delivery adapter boundary. Determinism and size-limit tests can remain against that encoder without making the core domain depend on Sentry's envelope shape.

backend/internal/observe/agentswitch/policy.go
```diff
@@ -0,0 +1,467 @@
...
+	}()
+}
+
+type authorityRead struct {
+	valid         bool
+	eventsEnabled bool
+	generation    string
+}
+
+func (c *Coordinator) readAuthority() authorityRead {
```
illegalcall: [P2] Put authority-file access behind a port

The policy coordinator directly performs `Lstat`, permission checks, `Open`, and JSON parsing. This mixes policy decisions with a concrete filesystem integration, contrary to the repository's port/adaptor boundary, and prevents tests from varying authority semantics without reproducing real filesystem behavior.

Please inject a small authority-reader port returning the validated policy snapshot; a filesystem adapter can own symlink, mode, and parsing rules. The coordinator can then remain responsible only for generation/policy transitions.

backend/internal/storage/sqlite/store/agent_switch_failure_store.go
```diff
@@ -0,0 +1,873 @@
...
+	status := s.enrollFaultSavepoint(ctx, tx, failureEnrollmentInput{
+		DaemonRunID: input.DaemonRunID, Fault: &input.Fault, Authorization: input.Authorization,
+	})
+	if err := tx.Commit(); err != nil {
+		return ports.AgentSwitchMutationResult{}, fmt.Errorf("commit daemon agent switch fault: %w", err)
+	}
+	return ports.AgentSwitchMutationResult{CoreChanged: status == domain.AgentSwitchEnrollmentEnrolled, Enrollment: status}, nil
+}
+
+// ForceDisableAgentSwitchFailurePolicy disables failure delivery without requiring the current consent generation.
```
illegalcall, line 515: [P2] Avoid a second hand-maintained copy of the observability SQL

From here onward, policy updates/purge, claims and leases, attempts, settlement/throttling, expiry, receipts, and backlog are implemented as raw SQL even though equivalent statements were added to `queries/agent_switch_failure_observability.sql` and generated into `gen/agent_switch_failure_observability.sql.go`. The authorization and lease predicates are safety-critical; two independently editable copies can silently drift.

Please use the generated queries consistently, including `Queries.WithTx(tx)` for transaction-bound operations. If a statement genuinely cannot be represented through sqlc, keep only that narrowly documented exception as raw SQL.

backend/internal/storage/sqlite/migrations/0124_agent_switch_failure_observability.sql
whoisasx: Fixed in 9442b0a9f. Migration 0119 now normalizes legacy failed/source_stop_unconfirmed rows to stopping_source/source_stop_unconfirmed during the rebuild, preserving the retained recovery marker. Added an upgrade regression test that starts at schema 0118, inserts the legacy row, and applies 0119.

backend/internal/adapters/runtime/conpty/runtime.go
whoisasx: Fixed in 9442b0a9f. A failed reservation removal now returns the retained handle with RuntimeEffectPossible and RuntimeCleanupFailed. The session manager keeps the recovery gate and does not perform an unsafe source restore. Added a manager-level ConPTY regression test with unregister failure injection.

backend/internal/adapters/runtime/tmux/tmux.go
whoisasx: Fixed in 9442b0a9f. When the exact AO supervisor is missing and no mismatched supervisor proves another generation, the fenced tmux probe now returns unknown/identity_missing rather than dead. Added the preserved-shell/manual-relaunch regression test.

backend/internal/observe/sentryobs/sentryobs.go
whoisasx: Fixed in 9442b0a9f. Sentry now uses a synchronous transport behind an intake gate, and opt-out waits for both dispatcher work and every already-entered provider call before acknowledging. Added a blocking-transport regression test proving an entered call is drained and no later capture starts after opt-out.

frontend/src/renderer/lib/telemetry.ts
whoisasx: Fixed in 9442b0a9f. Agent-switch failure consent is now separate from PostHog product analytics. PostHog initialization remains controlled by the existing renderer telemetry environment/package gate; failure opt-out neither calls PostHog opt-out nor purges its queues. Updated the renderer/shared telemetry tests accordingly.

backend/internal/domain/agent_switch_observability.go
whoisasx: Fixed in 9442b0a9f. The domain now owns only normalized failure vocabulary, validation, and severity. Sentry event/envelope encoding moved to the sentryobs adapter and is injected into storage through AgentSwitchFailureEventEncoder. Determinism, frozen-fixture, and size-limit tests moved with the adapter.

backend/internal/observe/agentswitch/policy.go
whoisasx: Fixed in 9442b0a9f. The coordinator now depends on AgentSwitchFailureAuthorityReader. The new policyauthority filesystem adapter owns Lstat, symlink/mode checks, bounded strict JSON parsing, and snapshot validation; daemon wiring injects it.

backend/internal/storage/sqlite/store/agent_switch_failure_store.go
whoisasx: Fixed in 9442b0a9f. Policy apply/purge, recovery-marker selection, claims/leases, attempts, settlement/throttling, expiry, receipts, and backlog now use generated sqlc queries, including transaction-bound Queries.WithTx operations. Only savepoint control and the bespoke operational-fingerprint statement remain narrowly raw.

whoisasx:
@Pulkit7070 I implemented the review follow-ups in 9442b0a9f and merged the latest main again in 9a102c262.\n\nThe runtime behavior is now explicit in the PR description: legacy source-stop markers migrate safely; ConPTY reports retained reservations as possible effects and keeps the recovery gate; tmux treats missing supervisor identity as unknown rather than dead; and opt-out drains entered synchronous Sentry provider calls before acknowledging. The new failure-reporting preference is also fully separated from existing PostHog analytics.\n\nI also moved Sentry wire encoding and authority-file access behind injected ports/adapters, replaced the duplicated safety-critical raw SQL with generated sqlc queries, replied to each inline finding with the exact fix, and resolved all eight threads.\n\nFinal targeted verification passed across the affected backend packages, frontend typecheck, and the focused telemetry suites (67 tests). CI is now running on the conflict-free head.

frontend/src/main/desktop-telemetry-controller.ts
```diff
@@ -0,0 +1,208 @@
...
+	async retryPendingCleanup(): Promise<TelemetryPolicyView> {
+		return this.serialize(async () => {
+			if (this.view.state === "applied") return this.snapshot();
+			let desktopCleanupFailed = false;
+			try {
+				const ack = await this.options.daemon.applyPolicy(this.view.consentGeneration, this.view.eventsEnabled);
+				if (!this.acknowledges(this.view.eventsEnabled, this.view.consentGeneration, ack)) throw new Error("policy was not acknowledged");
+				if (!this.view.eventsEnabled && this.pendingDesktopCleanup && !(await this.pendingDesktopCleanup())) { desktopCleanupFailed = true; throw new Error("desktop cleanup is incomplete"); }
+				this.pendingDesktopCleanup = null;
+				this.view = { ...this.view, state: "applied", acknowledged: true, reason: this.baseReason() };
```
illegalcall: [P1] Do not acknowledge a generation whose durable replace failed

This retry can mark an unacknowledged authority generation `applied`. I reproduced it with the real `TelemetryPolicyAuthority`: let rename succeed, make the following directory fsync fail, catch the rejected opt-out, then call `retryPendingCleanup()`. Retry returns `{ state: "applied", acknowledged: true }` while `authority.snapshot().acknowledged` is still false, and neither `transport.clearCache` nor `clearRendererQueues` runs because `disable()` exited before installing `pendingDesktopCleanup`. This violates the crash-durability and ordered-purge guarantees in the design (lines 535-542 and 638-673). Please retry/verify the durable authority operation and preserve the desktop cleanup transaction before permitting `applied`.

frontend/src/main/sentry-main.ts
```diff
@@ -1,105 +1,17 @@
...
-			integrations: (defaults: Array<{ name: string }>) =>
-				defaults.filter((i) => !DENY_INTEGRATIONS.has(i.name)),
-			beforeSend: (event: Record<string, unknown>) => scrubEvent(event),
-			beforeBreadcrumb: (crumb: Record<string, unknown> | null) => (crumb ? scrubEvent(crumb) : crumb),
-		});
-	} catch {
-		// SDK not installed yet, or init failed — remain a silent no-op.
-	}
+export function rendererCaptureAllowed(request: RendererTelemetryCapture): boolean {
+	if (typeof request.message !== "string" || request.message.length > 4096) return false;
```
illegalcall: [P2] Preserve the required main-boundary privacy scrubber

This accepts arbitrary renderer messages based only on length, while `renderer/lib/sentry.ts` forwards `Error.message` verbatim and this diff deletes the previous main/renderer path and URL scrubbers. A focused test confirms strings containing `/Users/alice/...` and loopback URLs pass unchanged. The transport is currently null, so exposure is gate-contained, but the design explicitly says raw errors, paths, and URLs must never leave the process and that the existing scrubber remains (lines 1278-1292 and 1302-1303). Restore scrubbing at this trust boundary, ideally with typed allowlisted fields, before this intake can be activated.

backend/internal/adapters/runtime/conpty/runtime.go
```diff
@@ -359,17 +540,28 @@ func (r *Runtime) resolveHostProtocol(ctx context.Context, sess *hostSession) (i
 // resolve looks up a session by id: first the in-memory map, then the B2
 // registry (for daemon-restart recovery). Returns nil if not found either way.
 func (r *Runtime) resolve(id string) *hostSession {
+	sess, _ := r.resolveWithEvidence(id)
+	return sess
+}
+
+func (r *Runtime) resolveWithEvidence(id string) (*hostSession, error) {
```
illegalcall: [P2] Thread the caller context through registry resolution

`resolveWithEvidence` now performs registry filesystem reads, PID probes, and potentially a prune write, but it accepts no context. `Create`, `Destroy`, `IsAlive`, and `ProbeFencedRuntime` therefore drop their cancellation/deadline at this boundary. Per `AGENTS.md`, functions doing I/O or blocking work take `context.Context` first. Please thread `ctx` through this function and the `Scan`/`scanLocked`/`readRaw` registry path, checking cancellation around OS calls without native context support.

illegalcall (changes requested):
Requesting changes on three verified issues. The opt-out retry bug is the blocker: I reproduced a post-rename directory-fsync failure with the real policy authority, after which retry reported `applied`/`acknowledged` while the authority remained unacknowledged and neither desktop purge hook ran. The desktop Sentry intake also removes an explicitly required privacy scrubber, and the new ConPTY registry resolution path drops caller cancellation across filesystem/process I/O.

backend/internal/adapters/runtime/conpty/runtime.go
whoisasx: Fixed in 38ee27f1c. Caller context now flows through every ConPTY registry-resolution caller, resolveWithEvidence, Scan, scanLocked, and readRaw. Cancellation is checked around registry path/read/parse work, each PID probe, prune writes, and the final resolution loop; fenced probes map cancellation to unknown/probe_failed. Added regressions proving Create, Destroy, IsAlive, and ProbeFencedRuntime honor cancellation and that cancellation during PID probing prevents the prune write. The full ConPTY and ptyregistry package tests pass.

frontend/src/main/sentry-main.ts
whoisasx: Fixed in 38ee27f1c. Electron main now sanitizes renderer captures before they reach the controller: the request shape and tag keys are closed allowlists, unknown fields are rejected, request_id is no longer forwarded, and Unix/Windows home paths plus file, app, localhost, and loopback URLs are scrubbed from messages and tag values. Added focused boundary tests covering both redaction and deny-by-default validation. Focused Vitest and frontend typecheck pass.

frontend/src/main/desktop-telemetry-controller.ts
whoisasx: Fixed in 38ee27f1c. retryPendingCleanup now re-establishes and verifies the exact unacknowledged authority generation before daemon apply, and the desktop cleanup transaction is installed before the authority write so a post-rename directory-fsync failure cannot lose cache or renderer purge work. Added a real TelemetryPolicyAuthority regression that reproduces that failure and proves retry keeps the generation unchanged, performs both desktop purges, and acknowledges only after every step succeeds. Focused Vitest and frontend typecheck pass.

Pulkit7070:
Re-reviewed after the fix commit (`38ee27f1c`) that responds to the latest round. All three findings are addressed, each with a test that reproduces the exact scenario.

- Opt-out retry (the blocker): `disable()` now installs the desktop purge before the durable write, and a failed durable replace lands in `cleanup_failed` with `acknowledged: false` instead of `applied`. The reproduction test forces a post-rename directory fsync failure, asserts the opt-out is rejected with no purge run, then asserts `retryPendingCleanup` reaches `applied`/`acknowledged` with `clearCache` and `clearRendererQueues` each run once.
- Sentry main scrubber: `scrubTelemetryText` is restored and applied to both the message and tag values, redacting home paths, Windows paths, and loopback URLs. Covered by a test asserting `/Users/...` paths and loopback URLs are redacted.
- ConPTY registry ctx: `resolveWithEvidence` now takes `ctx`, checks `ctx.Err()` at entry and around OS calls, and threads `ctx` into `ptyregistry.Scan`.

Ran locally: backend builds, and the conpty and observability packages pass under `-race`. The two frontend tests are env-gated on my side (no node_modules), covered by the green renderer-smoke and test jobs. CI is 10/10.

Two things still block the merge, neither a code defect:
- The branch is currently conflicting with main. It needs a rebase or merge to resolve conflicts.
- The prior CHANGES_REQUESTED review predates this fix commit, so it needs a re-review or dismissal to unblock.

Findings-wise this is good from my side once conflicts are resolved.

whoisasx:
@Pulkit7070 The latest `main` conflicts are now resolved in `57734e603`, and GitHub marks the PR mergeable again. The three requirements from your re-review remain covered: durable opt-out retry with cleanup ordering, Sentry path/loopback scrubbing, and context-aware ConPTY registry resolution. The focused backend merge suite passes; CI has restarted on this head. Could you please take another look and, if everything still looks good, re-review/dismiss the stale changes-requested review? Thanks.

backend/internal/storage/sqlite/migrations/0125_agent_switch_failure_observability.sql
```diff
@@ -0,0 +1,127 @@
+-- +goose Up
+-- The recovery migration already carries all of the current agent_switches
+-- columns and indexes. Extend it in place rather than rebuilding the table:
+-- this preserves its existing triggers and keeps startup migrations fast.
+ALTER TABLE agent_switches
+    ADD COLUMN failure_point TEXT NOT NULL DEFAULT '';
+
+-- 0094 allowed this retained marker in a failed row. It represents an
+-- unresolved source-stop outcome, so migrate it to its nonterminal state.
+UPDATE agent_switches
```
illegalcall, line 42: [P1] Reconcile legacy markers that already have a newer active switch

This unconditional promotion can abort startup migration for a valid pre-0124 history. A `failed/source_stop_unconfirmed` row did not participate in `idx_agent_switches_one_active_per_session`, so the same session could subsequently acquire a newer nonterminal switch. I reproduced that exact database state at version 118; applying 0124 fails here with `UNIQUE constraint failed: agent_switches.session_id` when this row becomes `stopping_source`. Please define which row retains the operation gate, migrate the conflicting history without violating the one-active invariant, and add this two-row upgrade case to the migration tests.

backend/internal/adapters/runtime/conpty/ptyregistry/registry.go
```diff
@@ -155,8 +185,12 @@ func writeRaw(entries []Entry) error {
 func Register(entry Entry) error {
 	registryMu.Lock()
 	defer registryMu.Unlock()
+	all, complete, err := scanLocked(context.Background())
```
illegalcall: [P2] Preserve caller context through registry mutations

`Register` is called from `Runtime.Create(ctx, ...)`, but this switches to `context.Background()` before registry reads/pruning and filesystem writes; `Unregister` does the same for `Destroy`. The runtime callback signatures cannot carry the caller deadline either. This still conflicts with `AGENTS.md`'s context-first rule for blocking/I/O functions and lets mutation work continue while holding `registryMu` after cancellation. Please thread `ctx` through `Register`, `Unregister`, `writeRaw`/`rewriteRegistry`, and the runtime callbacks, or explicitly derive and document a bounded detached cleanup context where durability requires detachment.

illegalcall (changes requested):
The prior opt-out durability and Sentry-scrubbing findings are fixed, but I am requesting changes on two newly verified blockers. I reproduced the migration failure against a valid pre-0124 database containing a legacy failed recovery marker plus a newer active switch for the same session. The ConPTY registry mutation path also still drops caller cancellation. Separately, the current Go build-test check is red after timing out at 10 minutes, so CI must be green before merge.

backend/internal/storage/sqlite/migrations/0125_agent_switch_failure_observability.sql
```diff
@@ -0,0 +1,127 @@
...
+SET state = 'stopping_source'
```
Pulkit7070, line 43: [P1] This UPDATE can fail the migration. Promoting `failed` to `stopping_source` makes the row non-terminal, but `idx_agent_switches_one_active_per_session` (`WHERE state NOT IN ('completed','failed')`) allows only one active switch per session. A session with two legacy `failed`/`source_stop_unconfirmed` rows (two failed source-stop attempts over time, both valid today), or one such row plus a currently-active switch, ends up with multiple active rows, and this statement aborts with `UNIQUE constraint failed: agent_switches.session_id`. The database cannot upgrade and the daemon cannot open it.

I reproduced it: seed a session with two `failed`/`source_stop_unconfirmed` rows, run 0124, and it aborts here. The migration test only inserts one such row, so it does not catch this.

Options: keep the row terminal and normalize its `error_code` to an allowed failed value instead of changing `state` (no active collision), or promote at most one row per session and normalize the rest. Add a multi-row regression test.

backend/internal/storage/sqlite/migrations/0125_agent_switch_failure_observability.sql
whoisasx: Fixed in `e29622442`. 0124 now gives any already-active switch ownership of the gate. Otherwise it promotes only the newest legacy `failed/source_stop_unconfirmed` row per session; all older conflicting markers remain terminal as `failed_post_stop`. The migration regression now covers both an older marker plus newer active switch and multiple legacy markers, and asserts exactly one active row after upgrade.

backend/internal/adapters/runtime/conpty/ptyregistry/registry.go
whoisasx: Fixed in `e29622442`. `Register`, `Unregister`, `writeRaw`, prune rewrites, and the runtime registry callbacks now take and propagate the caller context. Cancellation is checked before and around filesystem/PID work, including after mutex acquisition. Added regressions for cancelled mutations and for the exact caller context reaching both Create and Destroy registry callbacks.

backend/internal/storage/sqlite/migrations/0125_agent_switch_failure_observability.sql
whoisasx: Fixed in `e29622442`. I chose the bounded-recovery option: an existing active switch retains the gate; when no active switch exists, only the newest legacy source-stop marker is promoted. Older conflicting legacy markers are normalized to terminal `failed_post_stop`. Added migration coverage for both histories and verifies the one-active-switch invariant.

illegalcall:
Re-review complete: both previously requested changes are fixed on `c9bdb4abc`, and the focused migration + ConPTY regressions pass. I found no new code issue on this head.

This is not merge-ready yet because the branch now conflicts with current `main` in eight files, including `backend/internal/session_manager/agent_switching.go`, `backend/internal/daemon/lifecycle_wiring.go`, `frontend/src/renderer/components/SessionView.tsx`, and `frontend/src/renderer/lib/event-transport.ts`. Please sync with `main`, resolve those conflicts, and rerun CI; the resulting resolution should get a final review before merge.

illegalcall (approved):

PR #4601, opened 2026-08-28
feat(omp): track TUI session activity

nikhilachale:
## What

Add AO-managed OMP TUI lifecycle hooks so Kanban session status reflects active work, idle turns, approval waits, session switches, and process exit.

## Why

OMP TUI sessions had no structured activity integration, so their Kanban status could remain idle while the agent was working or waiting for user approval.

## How

- Install a project-local `.omp/extensions/ao-activity.ts` extension and load it explicitly for launch and restore.
- Map OMP lifecycle and approval events into the existing AO hook pipeline and activity-state derivation.
- Filter every callback to the top-level UI context so inherited subagent extension runners cannot alter the root AO session.
- Handle `session_switch` so `/new`, `/resume`, and `/fork` immediately refresh the native resume ID while preserving idle state.
- Require OMP 17.1.0+, the first release with the complete lifecycle contract used here, including `AgentEndEvent.willContinue`.
- Keep hook delivery best-effort and bounded below OMP's shutdown budget.
- Refuse to overwrite user-owned files at the managed extension path and preserve all other OMP extensions.
- Keep ACP Chat activity tracking unchanged; this change is focused on the TUI adapter.

## Testing

- `cd backend && go test ./internal/adapters/agent/omp ./internal/adapters/agent/activitydispatch ./internal/adapters/agent/registry -count=1`
- Executable Node fixtures validate lifecycle ordering, switched native IDs, root-versus-child contexts, continuation behavior, missing AO handling, nonzero exits, and timeout handling.
- Version fixtures reject OMP 16.9.9 and 17.0.0 and accept OMP 17.1.0 and 18.1.0.

## Checklist

### DIFF
--- a/backend/internal/adapters/agent/omp/hooks.go
+++ b/backend/internal/adapters/agent/omp/hooks.go
@@ -1,0 +1,124 @@
+package omp
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/hookutil"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+const (
+	ompExtensionsDirName = "extensions"
+	ompExtensionFileName = "ao-activity.ts"
+	ompExtensionSentinel = "agent-orchestrator: managed omp activity extension"
+)
+
+func ompExtensionPath(workspacePath string) string {
+	return filepath.Join(workspacePath, ".omp", ompExtensionsDirName, ompExtensionFileName)
+}
+
+// GetAgentHooks installs AO's project-local OMP activity extension. Launch and
+// restore also pass this exact file explicitly, so status tracking does not
+// depend on extension auto-discovery.
+func (p *Plugin) GetAgentHooks(ctx context.Context, cfg ports.WorkspaceHookConfig) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if strings.TrimSpace(cfg.WorkspacePath) == "" {
+		return errors.New("omp.GetAgentHooks: WorkspacePath is required")
+	}
+
+	path := ompExtensionPath(cfg.WorkspacePath)
+	if data, err := os.ReadFile(path); err == nil { //nolint:gosec // caller-owned workspace path
+		if !strings.Contains(string(data), ompExtensionSentinel) {
+			return fmt.Errorf("omp.GetAgentHooks: refusing to overwrite non-AO file at %s", path)
+		}
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("omp.GetAgentHooks: read managed extension: %w", err)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(path), 0o750); err != nil {
+		return fmt.Errorf("omp.GetAgentHooks: create extension dir: %w", err)
+	}
+	if err := hookutil.AtomicWriteFile(path, []byte(ompActivityExtensionSource()), 0o600); err != nil {
+		return fmt.Errorf("omp.GetAgentHooks: write extension: %w", err)
+	}
+	if err := hookutil.EnsureWorkspaceGitignore(filepath.Dir(path), ompExtensionFileName); err != nil {
+		return fmt.Errorf("omp.GetAgentHooks: gitignore: %w", err)
+	}
+	return nil
+}
+
+func appendOMPExtensionFlag(cmd *[]string, workspacePath string) {
+	if strings.TrimSpace(workspacePath) == "" {
+		return
+	}
+	*cmd = append(*cmd, "--extension", ompExtensionPath(workspacePath))
+}
+
+func ompActivityExtensionSource() string {
+	return `// ` + ompExtensionSentinel + ` (do not edit)
+import type { ExtensionAPI } from "@oh-my-pi/pi-coding-agent";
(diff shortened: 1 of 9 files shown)


backend/internal/adapters/agent/omp/hooks.go
```diff
@@ -0,0 +1,126 @@
...
+  } catch {
+    // A missing AO executable or timeout must not interrupt OMP.
+  }
+}
+
+function sessionID(ctx: any): string {
+  return ctx.sessionManager.getSessionId() ?? "";
+}
+
+export default function (omp: ExtensionAPI) {
```
whoisasx, line 164: [P1] Please prevent inherited child-agent extension instances from publishing lifecycle events for the root AO session. OMP forwards ordinary parent extension paths into subagent sessions, initializes a separate child extension runner, and disposes that child session afterward. Those callbacks inherit the same `AO_SESSION_ID`, so the child's `session_start`, `agent_end`, and `session_shutdown` update the parent AO record while also replacing its native resume ID with the child's ID. In particular, child shutdown can mark the still-running root session `exited`; AO then ignores the later root idle callback and follow-up delivery can remain blocked. Since this extension is only installed for the interactive TUI path, filter every callback to the top-level UI context (for example, `ctx.hasUI`) or introduce another reliable root-session guard, and add a child-context regression test.

backend/internal/adapters/agent/omp/hooks.go
```diff
@@ -0,0 +1,126 @@
...
+  omp.on("session_start", async (_event, ctx) => {
+    callHookSync("session-start", { session_id: sessionID(ctx) });
+  });
+  omp.on("before_agent_start", async (event, ctx) => {
+    callHookSync("user-prompt-submit", { session_id: sessionID(ctx), prompt: event.prompt });
+  });
+  omp.on("tool_approval_requested", async (event, ctx) => {
```
whoisasx, line 177: [P2] This relies on OMP lifecycle capabilities without enforcing the version that introduced them. AO accepts the user's existing OMP binary, but OMP v15 does not expose these approval events or `AgentEndEvent.willContinue`, and v16 still lacks `willContinue`. On those installations approval waits are silently unreported and `!event.willContinue` treats every completion as terminal, including completions followed by automatic continuation. Please probe/enforce the required OMP version or generate an explicitly degraded fallback, document that boundary, and cover the minimum supported OMP contract rather than validating only the annotation-stripped source under Node.

backend/internal/adapters/agent/omp/hooks.go
```diff
@@ -0,0 +1,126 @@
...
+    // A missing AO executable or timeout must not interrupt OMP.
+  }
+}
+
+function sessionID(ctx: any): string {
+  return ctx.sessionManager.getSessionId() ?? "";
+}
+
+export default function (omp: ExtensionAPI) {
+  omp.on("session_start", async (_event, ctx) => {
```
whoisasx, line 165: [P2] Please also report the current native session ID on OMP's `session_switch` event. `session_start` fires only for the initial load, while `/new`, `/resume`, and `/fork` use `session_switch`. Without handling it, AO retains the previous resume ID until some later tracked callback; if OMP crashes in that interval, restore opens the wrong native conversation. A `session_switch` handler can publish the current ID immediately while preserving the idle state expected at that boundary.

whoisasx (changes requested):
The overall approach follows the existing AO hook boundary, preserves user-owned files, and keeps delivery bounded and best-effort. However, the inherited-subagent behavior is blocking: a child OMP session can overwrite the root AO session's activity and native resume identity, including marking a live root session exited. Please address that before merging. The OMP version compatibility and in-process session-switch metadata gaps should also be covered. Current CI is green, but the branch still has a merge conflict with main.

backend/internal/adapters/agent/omp/hooks.go
nikhilachale: Fixed in ca636c16e. Every OMP lifecycle callback now returns unless ctx.hasUI is true, so inherited subagent extension runners cannot publish activity or native session IDs for the root AO session. Added an executable regression that invokes every registered callback with a child context where hasUI is false and asserts that no AO hook process is spawned.

backend/internal/adapters/agent/omp/hooks.go
nikhilachale: Fixed in ca636c16e. AO now probes the installed OMP version before installing the TUI activity extension and requires OMP 17.1.0 or newer. I verified the upstream contracts directly: 17.0.0 still lacks AgentEndEvent.willContinue, while 17.1.0 includes it. Tests reject 16.9.9 and 17.0.0 and accept 17.1.0 and 18.1.0; the boundary is documented.

backend/internal/adapters/agent/omp/hooks.go
nikhilachale: Fixed in ca636c16e. The managed extension now handles session_switch for the root UI context and emits session-start with the current native ID from the session manager. This refreshes the resume identity while preserving idle state. The executable lifecycle test switches the native ID and verifies all subsequent payloads use it.

whoisasx (approved):
Re-reviewed the latest head. The root-session guard now prevents inherited OMP subagents from corrupting parent activity or resume metadata, the OMP 17.1.0 lifecycle boundary is enforced and documented, and session_switch refreshes the native resume identity immediately. The merge conflict is resolved and all GitHub checks are green. The Unix-only executable fixture leaves a small Windows-specific test gap, but I do not consider it blocking. Looks good to merge from my side.

Issue #4602, opened 2026-08-28
feat: support ao:// deep links to sessions


harshitsinghbhandari:
> *This was generated by AI during triage.*

## Feature

Register and handle a stable `ao://` deep-link protocol for navigating to Agent Orchestrator resources, beginning with sessions.

Canonical session URL:

```text
ao://sessions/{project-id}/{session-id}
```

The path-based form is deliberate: project and session identifiers remain distinct, percent-encodable path segments instead of being ambiguously packed into a dotted hostname.

## User value

Links copied from AO, terminals, browsers, issue descriptions, and other applications should open the referenced session in Agent Orchestrator. Session IDs remain stable when display names change, so shared links remain valid.

## Scope

- Define a versioned parser/resolver for the `ao://` scheme.
- Support `ao://sessions/{project-id}/{session-id}` first.
- Register the Electron app as the operating-system protocol handler on supported macOS, Windows, and Linux packaging paths.
- Route links into the existing AO window when it is running; launch AO and navigate after readiness when it is not.
- Focus the target session's appropriate surface, including retained history for terminated sessions.
- Show a clear error when a project/session is missing, inaccessible, malformed, or unsupported.
- Provide a copy-link action using the canonical URL.
- Keep the parser/navigation contract reusable by in-app Chat and TUI session links.

## Security boundary

- Deep links navigate only; they must never execute shell commands or silently send messages, spawn/kill sessions, mutate GitHub state, or merge work.
- Allowlist route types and validate/decode every path segment.
- Reject credentials, unexpected authorities, malformed encoding, extra action parameters, and unsupported routes.
- Do not pass raw URLs to a shell.
- Navigation requires an explicit user link activation outside any documented OS launch behavior.

## Acceptance criteria

- Clicking a canonical session URL outside AO opens or focuses AO at the correct project/session.
- The same URL resolves inside an already-running AO instance without opening a duplicate window.
- Session renames do not invalidate links.
- Terminated sessions open in their supported read-only/history view.
- Missing and malformed targets fail safely with actionable UI feedback.
- Protocol registration and second-instance routing work in packaged builds on macOS, Windows, and Linux where supported.
- Unit tests cover parsing, encoding, allowlisting, malformed URLs, unsupported routes, and stable IDs.
- Packaged-app tests cover cold launch, warm launch/second instance, and unavailable targets.

## Product questions

- Should initial support be limited to sessions, or reserve documented routes for projects and PRs without implementing them yet?
[...]

PR #4606, opened 2026-08-28
feat: add rich in-app session links

### LINKED ISSUE #4603: feat: render rich session links in Chat and TUI
> *This was generated by AI during triage.*

## Feature

Allow any AO participant—not only an orchestrator—to reference another session using rich, clickable session links in Chat and the in-app TUI.

Example canonical target:

```text
ao://sessions/agent-orchestrator/agent-orchestrator-255
```

An agent or user may include that target in a response. AO should render or detect it as a session reference and navigate only after the user clicks it.

## User value

Workers frequently hand off investigation, implementation, review, or follow-up work to other sessions. Today the user must locate the target manually in the sidebar. A session reference should show useful identity/status and open the correct session directly.

### Chat

- Recognize allowlisted session URLs in messages from users, workers, orchestrators, and automation; do not role-gate the feature.
- Support Markdown links and raw canonical URLs.
- Render a rich session chip/link when target metadata is available, with a safe plain-link fallback.
- Display the current session name, role/harness, and useful status without making the display name part of identity.

### In-app TUI

- Register a terminal link provider for canonical `ao://sessions/...` text and supported OSC 8 links.
- Make activation explicit, using the platform-appropriate modifier/click behavior where needed to avoid interfering with terminal selection.
- Intercept the URL inside AO; never pass it to the shell or an external browser.

### Shared behavior

- Use one strict parser/resolver/navigation path across Chat and TUI.
- Resolve by stable project/session IDs so renaming does not break old messages.
- Open terminated sessions in retained history/read-only mode when available.
- Show an actionable error for missing, inaccessible, malformed, or unsupported targets.
- Do not navigate automatically while content streams or merely because an agent emitted a URL.

## Security boundary

- This feature is navigation-only.
- An agent must not force a route change or encode actions such as send, spawn, kill, merge, or shell execution.
- The Markdown sanitizer and terminal detector must allow only the canonical AO route shape and reject lookalike/malformed schemes.
- Link labels are untrusted presentation; resolved session metadata supplies authoritative identity.

## Acceptance criteria

- The same canonical session URL works in Chat and the in-app TUI.
- Messages from every supported origin/role receive identical link behavior.
- Clicking opens the exact project/session in the current AO app.
- Rendering never causes automatic navigation.
- Renaming the target changes its displayed metadata without breaking the stored link.
- Missing and terminated-session behavior is explicit and tested.
- TUI parsing handles line wrapping and adjacent punctuation without consuming unrelated terminal text.
- Tests cover Chat Markdown/raw URLs, sanitizer behavior, TUI raw/OSC 8 detection, modifier activation, renamed/terminated/missing targets, and malicious URL variants.

## Product questions

- Should the first UI be a compact inline chip or a conventional styled link with a hover preview?
- Should a click always open Chat, or restore the target's last active inspector surface?
- Should generated messages expose a structured session-reference entity in addition to parsing canonical URLs?

## Related work

- #4602 defines system-level `ao://` registration and the shared canonical session URL contract.

harshitsinghbhandari:
Code changes: +329 -10
Tests: +346 -2
Others: +41 -0

Closes #4603

## Summary

- Add a strict shared parser and resolver for canonical `ao://sessions/{project-id}/{session-id}` links.
- Render raw and Markdown session links across Chat message origins and navigate inside the current AO window.
- Detect raw and OSC 8 links in xterm, including wrapped and punctuation-adjacent URLs, while preserving selection and mouse-capture modifiers.
- Surface malformed, missing, inaccessible, and unverifiable targets as in-app feedback.
- Report terminated targets in a five-second, manually dismissible top notification stack without changing the active session.
- Teach orchestrator and worker system prompts plus the installed `using-ao` skill to emit safe canonical links using stable IDs.

## Rebase

- Rebased the three Resource Links commits onto current `main` at `850df06e5a6f41336823f4c31b3f4a0161284d3d`.
- Preserved current Chat file links, image rendering, agent-switch catalogs, terminal drag selection, notifications, and locale additions while resolving conflicts.
- No database schema, migration, SQL, or generated API changes are involved.

## Verification

- Focused frontend suite: 7 files and 317 tests passed.
- Frontend TypeScript check passed.
- Focused backend session-manager prompt tests passed.
- Backend skill-assets suite passed.
- `git diff --check origin/main...HEAD` passed.
- Full frontend and backend suites were attempted. Environment-dependent local socket, bundled tmux, and system-check tests failed outside the changed session-link paths.

## Manual platform verification

- macOS: with a mouse-reporting TUI active, verify ordinary click remains app-owned and Command-click opens the AO session link. With capture off, verify ordinary click navigates and drag selects.
- Windows and Linux: repeat with Control-click while mouse capture is active. Verify ordinary click when capture is off and drag selection in both modes.
- All platforms: verify raw and OSC 8 terminal links, raw and Markdown Chat links, renamed and cross-project active targets, missing-target feedback, terminated-target stacked feedback, and restoration of the target session persisted Chat or TUI surface.

## Scope

- Canonical session identity remains stable-ID based and navigation-only.
- The canonical route retains both project and session identity.
- OS protocol registration and cold-launch behavior remain out of scope. See #4602.

### DIFF
--- a/frontend/src/renderer/components/SessionLinkFeedback.tsx
+++ b/frontend/src/renderer/components/SessionLinkFeedback.tsx
@@ -1,23 +1,51 @@
+import { useEffect } from "react";
 import { X } from "lucide-react";
 import { useTranslation } from "react-i18next";
-import { useUiStore } from "../stores/ui-store";
+import { type SessionLinkNotice, useUiStore } from "../stores/ui-store";
+
+const NOTICE_DISMISS_MS = 5_000;
 
 export function SessionLinkFeedback() {
 	const { t } = useTranslation();
 	const error = useUiStore((state) => state.sessionLinkError);
+	const notices = useUiStore((state) => state.sessionLinkNotices);
 	const dismiss = useUiStore((state) => state.setSessionLinkError);
-	if (!error) return null;
+	if (!error && notices.length === 0) return null;
 	return (
-		<div className="pointer-events-auto fixed bottom-4 left-1/2 z-overlay flex max-w-md -translate-x-1/2 items-center gap-3 rounded-lg border border-destructive/40 bg-background px-3 py-2 text-xs text-destructive shadow-xl" role="alert">
-			<span>{error}</span>
-			<button
-				type="button"
-				aria-label={t("sessionLink.dismissError")}
-				onClick={() => dismiss(null)}
-				className="rounded p-1 hover:bg-interactive-hover"
-			>
+		<>
+			{notices.length > 0 ? (
+				<div
+					className="pointer-events-auto fixed left-1/2 top-12 z-overlay flex max-h-[184px] w-[min(360px,calc(100vw-24px))] -translate-x-1/2 flex-col gap-2 overflow-y-auto overscroll-contain"
+					data-browser-native-overlay="true"
+					data-testid="session-link-notifications"
+				>
+					{notices.slice().reverse().map((notice) => <SessionLinkNoticeItem key={notice.nonce} dismissLabel={t("sessionLink.dismissError")} notice={notice} />)}
+				</div>
+			) : null}
+			{error ? (
+				<div className="pointer-events-auto fixed bottom-4 left-1/2 z-overlay flex max-w-md -translate-x-1/2 items-center gap-3 rounded-lg border border-destructive/40 bg-background px-3 py-2 text-xs text-destructive shadow-xl" role="alert">
+					<span>{error}</span>
+					<button type="button" aria-label={t("sessionLink.dismissError")} onClick={() => dismiss(null)} className="rounded p-1 hover:bg-interactive-hover">
+						<X aria-hidden="true" className="size-3" />
+					</button>
+				</div>
+			) : null}
+		</>
+	);
+}
+
+function SessionLinkNoticeItem({ dismissLabel, notice }: { dismissLabel: string; notice: SessionLinkNotice }) {
+	const dismiss = useUiStore((state) => state.dismissSessionLinkNotice);
+	useEffect(() => {
+		const timer = window.setTimeout(() => dismiss(notice.nonce), NOTICE_DISMISS_MS);
+		return () => window.clearTimeout(timer);
+	}, [dismiss, notice.nonce]);
+	return (
+		<section aria-live="polite" className="flex min-h-10 shrink-0 items-center gap-3 rounded-lg border border-border bg-background px-3 py-2 text-xs text-foreground shadow-xl" role="status">
+			<span className="min-w-0 flex-1 wrap-break-word">{notice.message}</span>
+			<button type="button" aria-label={dismissLabel} onClick={() => dismiss(notice.nonce)} className="shrink-0 rounded p-1 hover:bg-interactive-hover">
 				<X aria-hidden="true" className="size-3" />
 			</button>
-		</div>
+		</section>
 	);
 }
--- a/frontend/src/renderer/components/SessionLinkFeedback.tsx
+++ b/frontend/src/renderer/components/SessionLinkFeedback.tsx
@@ -1,0 +1,23 @@
+import { X } from "lucide-react";
+import { useTranslation } from "react-i18next";
+import { useUiStore } from "../stores/ui-store";
+
+export function SessionLinkFeedback() {
+	const { t } = useTranslation();
+	const error = useUiStore((state) => state.sessionLinkError);
+	const dismiss = useUiStore((state) => state.setSessionLinkError);
+	if (!error) return null;
+	return (
+		<div className="pointer-events-auto fixed bottom-4 left-1/2 z-overlay flex max-w-md -translate-x-1/2 items-center gap-3 rounded-lg border border-destructive/40 bg-background px-3 py-2 text-xs text-destructive shadow-xl" role="alert">
+			<span>{error}</span>
+			<button
+				type="button"
+				aria-label={t("sessionLink.dismissError")}
+				onClick={() => dismiss(null)}
+				className="rounded p-1 hover:bg-interactive-hover"
+			>
+				<X aria-hidden="true" className="size-3" />
+			</button>
+		</div>
+	);
+}
--- a/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
+++ b/frontend/src/renderer/components/chat/ChatTimelineItems.tsx
@@ -55,6 +55,7 @@ import { caretNotation, stripAnsi } from "../../lib/ansi";
 import { getApiBaseUrl } from "../../lib/api-client";
 import { isWebLink, openLinkInSystemBrowser } from "../../lib/external-link-policy";
 import { ActivityTitle, ChatMarkdown } from "./ChatMarkdown";
+import { findSessionLinks } from "../../lib/session-links";
 import { HighlightedCode } from "./HighlightedCode";
 import { CopyButton } from "./CopyButton";
 import { HumanMessageEditor } from "./HumanMessageEditor";
@@ -534,7 +535,13 @@ export function HumanMessage({
 							: "bg-raised text-foreground",
 					)}
 				>
-					{body ? <p className="break-words whitespace-pre-wrap text-pretty">{body}</p> : null}
+					{body ? (
+						findSessionLinks(body).length > 0 ? (
+							<ChatMarkdown text={body} />
+						) : (
+							<p className="break-words whitespace-pre-wrap text-pretty">{body}</p>
+						)
+					) : null}
 					<StagedAttachmentItems
 						paths={attachments}
 						sessionId={sessionId}
@@ -620,9 +627,15 @@ export function OriginMessage({ message }: { message: ConversationMessage }) {
 			{longReport && expanded ? (
 				<ChatMarkdown text={message.text} muted />
 			) : (
-				<p className={cn("text-sm leading-relaxed text-muted-foreground", longReport && "line-clamp-3")}>
-					{preview}
-				</p>
+				findSessionLinks(preview).length > 0 ? (
+					<div className={cn(longReport && "line-clamp-3")}>
+						<ChatMarkdown text={preview} muted />
+					</div>
+				) : (
+					<p className={cn("text-sm leading-relaxed text-muted-foreground", longReport && "line-clamp-3")}>
+						{preview}
+					</p>
+				)
 			)}
 			{longReport ? (
 				<button
--- a/frontend/src/renderer/lib/session-links.ts
+++ b/frontend/src/renderer/lib/session-links.ts
@@ -1,0 +1,77 @@
+export type SessionLinkTarget = { projectId: string; sessionId: string };
+export type SessionLinkWorkspace = { id: string; sessions: Array<{ id: string }> };
+
+const SESSION_LINK_PREFIX = "ao://sessions/";
+const RAW_SESSION_LINK = /ao:\/\/sessions\/\S+/g;
+const TRAILING_PUNCTUATION = /[.,;:!?]+$/;
+
+/** Strictly parse the only in-app URL route AO supports. */
+export function parseSessionLink(value: string): SessionLinkTarget | undefined {
+	if (!value.startsWith(SESSION_LINK_PREFIX) || value.includes("?") || value.includes("#")) return undefined;
+	const segments = value.slice(SESSION_LINK_PREFIX.length).split("/");
+	if (segments.length !== 2 || segments.some((segment) => segment.length === 0)) return undefined;
+	try {
+		const [projectId, sessionId] = segments.map(decodeURIComponent);
+		if (!projectId || !sessionId || projectId.includes("/") || sessionId.includes("/")) return undefined;
+		return { projectId, sessionId };
+	} catch {
+		return undefined;
+	}
+}
+
+export function isSessionLink(value: string): boolean {
+	return parseSessionLink(value) !== undefined;
+}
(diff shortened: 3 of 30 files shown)


harshitsinghbhandari:
Agreed follow-up design, awaiting a separate implementation instruction:

- `ao://projects/{project-id}/kanban/` opens that project's Kanban.
- `ao://home/` opens the current Home.
- `ao://kanban/` opens the all-project Kanban.
- `ao://projects/{project-id}/orchestrator/` opens the most recently spawned non-terminated orchestrator. It never auto-spawns or restores one.
- `ao://sessions/{project-id}/{session-id}` remains stable-ID based for workers and orchestrators. Display-name URLs are not supported, but friendly Markdown labels are allowed.
- Every route accepts an optional trailing slash.
- Errors leave the current page unchanged and use the same five-second, manually dismissible, four-visible then scroll popup with these messages: `Project {id} was not found`; `Session {id} was not found`; `Session {id} is terminated`; `Session {id} does not belong to project {project-id}`; `This AO link is invalid`; `Could not open this link. Try again`; `Project {id} has no active orchestrator`.

frontend/src/renderer/lib/session-links.ts
```diff
@@ -0,0 +1,83 @@
...
+	const session = project?.sessions.find((candidate) => candidate.id === target.sessionId);
+	if (!session) return undefined;
+	return { ...target, isTerminated: session.isTerminated === true || session.status === "terminated" };
+}
+
+/** Find raw links without swallowing prose punctuation adjacent to them. */
+export function findSessionLinks(text: string): Array<{ start: number; end: number; text: string }> {
+	const matches: Array<{ start: number; end: number; text: string }> = [];
+	for (const match of text.matchAll(RAW_SESSION_LINK)) {
+		let candidate = match[0].replace(TRAILING_PUNCTUATION, "");
```
Pulkit7070: The trailing-punctuation trim here strips only `. , ; : ! ?` (TRAILING_PUNCTUATION at line 10) and an unbalanced `)` (line 44). A link wrapped in double quotes in prose or terminal output, `"ao://sessions/proj/sess"`, matches via the greedy `\S+` with the closing `"` retained, and parseSessionLink then accepts sessionId `sess"` as valid. The link renders but resolves to nothing, so the user sees "That session is missing or is not accessible" instead of navigation. The same happens for `]`, `'`, `»`, and `&gt;` as closing delimiters. Confirmed with findSessionLinks('"ao://sessions/proj/sess"') returning text `ao://sessions/proj/sess"`. Consider stripping a matching wrapping delimiter (quote/bracket) the same way the unbalanced `)` is handled.

frontend/src/renderer/lib/session-links.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const SESSION_LINK_PREFIX = "ao://sessions/";
+const RAW_SESSION_LINK = /ao:\/\/sessions\/\S+/g;
+const TRAILING_PUNCTUATION = /[.,;:!?]+$/;
+
+/** Strictly parse the only in-app URL route AO supports. */
+export function parseSessionLink(value: string): SessionLinkTarget | undefined {
+	if (!value.startsWith(SESSION_LINK_PREFIX) || value.includes("?") || value.includes("#")) return undefined;
+	const segments = value.slice(SESSION_LINK_PREFIX.length).split("/");
+	if (segments.length !== 2 || segments.some((segment) => segment.length === 0)) return undefined;
```
Pulkit7070, line 29: The empty-segment guard checks `segment.length === 0` before decode, and the truthiness check at line 19 runs after decode, but a decoded `" "` is truthy. So `parseSessionLink("ao://sessions/%20/%20")` returns `{ projectId: " ", sessionId: " " }`. Harmless in practice since resolution then fails and the missing-session error fires, but a parser advertised as strict should reject whitespace-only ids. Trimming and re-checking emptiness after decodeURIComponent would close it.

frontend/src/renderer/lib/session-links.ts
```diff
@@ -0,0 +1,83 @@
+export type SessionLinkTarget = { projectId: string; sessionId: string };
+export type ResolvedSessionLink = SessionLinkTarget & { isTerminated: boolean };
+export type SessionLinkWorkspace = {
+	id: string;
+	sessions: Array<{ id: string; isTerminated?: boolean; status?: string }>;
+};
+
+const SESSION_LINK_PREFIX = "ao://sessions/";
+const RAW_SESSION_LINK = /ao:\/\/sessions\/\S+/g;
```
Pulkit7070, line 9: RAW_SESSION_LINK is greedy to the next whitespace, so `ao://sessions/p/s/` and `ao://sessions/p/s/extra` both match, fail the two-segment rule at line 16, and linkify nothing (findSessionLinks returns []). This is arguably the correct treatment of malformed input, and prompt-compliant agent output is always exactly two segments, so impact is low. Flagging it because a single stray trailing slash renders the whole link as inert text rather than surfacing any feedback.

frontend/src/renderer/lib/use-session-link-navigation.ts
```diff
@@ -0,0 +1,35 @@
...
+			return false;
+		}
+		const resolved = resolveSessionLink(url, workspaceQuery.data);
+		if (!resolved) {
+			setSessionLinkError("That session is missing or is not accessible in this AO workspace.");
+			return false;
+		}
+		setSessionLinkError(null);
+		if (resolved.isTerminated) {
+			showSessionLinkNotice(`Session ${resolved.sessionId} is terminated`);
```
Pulkit7070, line 29: showSessionLinkNotice is called on every click of a terminated link, and the ui-store appends unconditionally with a fresh nonce, so clicking the same terminated link N times pushes N identical "Session X is terminated" toasts, each with its own 5s timer. The error channel (setSessionLinkError) is single-valued, but this notice channel is an unbounded list with no coalescing. Consider deduping by session id or replacing the existing notice for the same target.

Pulkit7070 (commented):
This is a well-contained feature and the parts most likely to be dangerous are handled well. The parser rejects query strings and fragments, requires exactly two non-empty path segments, decodes each segment, and rejects a decoded segment that reintroduces a slash, so a percent-encoded slash cannot smuggle in a third segment. The same parser backs both the markdown remark plugin and the xterm link provider, so the two surfaces cannot drift. Chat rendering stays safe because rehype-raw is deliberately absent so agent HTML stays escaped, and the URL transform only passes a URL through untouched when it is a session link, so a javascript: or data: scheme cannot ride the session-link exemption past the sanitizer. Terminal gating (plain click swallowed while a TUI owns the mouse, Cmd/Ctrl-click escapes it) matches the existing web-link behavior rather than inventing a new convention. The focused frontend suites (316 tests) and the backend session_manager and skillassets tests pass.

A few things worth addressing, none blocking:

1. Trailing quote or bracket is absorbed into the session id, so a link wrapped in quotes resolves to nothing instead of navigating (inline at session-links.ts).
2. Whitespace-only ids parse as valid, which slips past a parser advertised as strict (inline).
3. A stray trailing slash or extra segment drops the whole link to inert text rather than linkifying the valid prefix (inline; arguably correct for malformed input).
4. Terminated-session notices stack without dedupe, so repeated clicks pile up identical toasts (inline).

One correction to the description: the parser is not shared with the backend. The backend has no session-link parser; prompt.go and the embedded skill markdown only instruct agents to emit the canonical form. That is fine because the backend never parses these links, but a shared parser across frontend and backend is not what shipped.

frontend/src/renderer/lib/use-session-link-navigation.ts
harshitsinghbhandari: Fixed in 96f1fdea8. Repeated clicks now replace the existing notice with the same message using a fresh nonce, so only one toast exists per terminated target and its five-second dismissal timer restarts. Added component and navigation regressions; focused tests and frontend typecheck pass.

frontend/src/renderer/lib/session-links.ts
harshitsinghbhandari: Fixed in 8438d24b4. Raw session-link detection now trims matching and unbalanced quote, bracket, guillemet, angle, and encoded `&lt;...&gt;` delimiters, including nested trailing punctuation. Added direct coverage for all reported examples. The parser, terminal, and Chat suites pass 188 tests, and frontend typecheck passes.

frontend/src/renderer/lib/session-links.ts
harshitsinghbhandari: Fixed in 1d3726f51. The strict parser now checks decoded IDs with `trim()` for emptiness, so encoded spaces and tabs are rejected while valid IDs retain their exact decoded value. Added regression cases for `%20/%20` and `%09`. The parser, terminal, and Chat suites pass 190 tests, and frontend typecheck passes.

frontend/src/renderer/components/chat/ChatTimelineItems.tsx
```diff
@@ -563,7 +564,13 @@ export function HumanMessage({
 							: "bg-raised text-foreground",
 					)}
 				>
-					{body ? <p className="break-words whitespace-pre-wrap text-pretty">{body}</p> : null}
+					{body ? (
+						findSessionLinks(body).length > 0 ? (
```
neversettle17-101, line 568: **Regression: human/automation/daemon messages lose literal-text guarantee when they contain a session link**

When `body` (or `preview`) contains an `ao://sessions/...` link, this now routes the *entire* message through `ChatMarkdown` (full GFM) instead of the plain `<p className="whitespace-pre-wrap">`. `ChatMarkdown`'s `p` renderer has no `whitespace-pre-wrap`, so:

- Manual line breaks in a multi-line human message collapse (GFM treats a single `\n` as a soft break, rendered as normal HTML whitespace with no `pre-wrap`).
- Any GFM-significant characters in the raw text (`#`, `*`, `-`, backticks, `>`, etc.) now get reinterpreted as headings/lists/emphasis/code/blockquotes instead of rendering literally.

Concrete repro: a user pastes a multi-line message such as `"Notes:\n- fix bug\nao://sessions/proj/sess\n- write test"`. Before this PR it rendered as literal text with the newlines preserved; after, it renders as a bulleted list with the newlines lost, because the mere presence of the session link switches the whole message to markdown rendering.

Same issue applies to the `OriginMessage` `preview` branch a few lines below (~line 664).

frontend/src/renderer/components/SessionLinkFeedback.tsx
```diff
@@ -0,0 +1,51 @@
...
+	const { t } = useTranslation();
+	const error = useUiStore((state) => state.sessionLinkError);
+	const notices = useUiStore((state) => state.sessionLinkNotices);
+	const dismiss = useUiStore((state) => state.setSessionLinkError);
+	if (!error && notices.length === 0) return null;
+	return (
+		<>
+			{notices.length > 0 ? (
+				<div
+					className="pointer-events-auto fixed left-1/2 top-12 z-overlay flex max-h-[184px] w-[min(360px,calc(100vw-24px))] -translate-x-1/2 flex-col gap-2 overflow-y-auto overscroll-contain"
```
neversettle17-101, line 18: **Reuse: duplicates the existing `GlobalToast` notification system**

This component plus the new `ui-store` fields (`sessionLinkError`, `sessionLinkNotices`, `sessionLinkNoticeSequence`, `showSessionLinkNotice`, `dismissSessionLinkNotice`) reimplement a stacked, auto-dismissing, top-center notification overlay that already exists as `GlobalToast` / `useUiStore().showGlobalToast` (see `frontend/src/renderer/components/GlobalToast.tsx` and `globalToasts`/`showGlobalToast`/`dismissGlobalToast` in `ui-store.ts`). `GlobalToast` already supports a stack, `top-center` placement, per-item dismiss timers, and an `error` tone.

The only capability missing there is a manual per-item dismiss button. Extending `GlobalToast`/`GlobalToastItem` with an optional dismiss button (and reusing `showGlobalToast(message, undefined, "top-center")` for termination notices, "error" tone for the malformed/missing-session banner) would avoid a second parallel state shape, component, and set of CSS classes to maintain for the same visual pattern.

neversettle17-101:
**Cross-file gap: `ReviewerChatSurface` never gets `onSessionLinkOpen`, leaving dead session links (not in this PR's diff, but relevant at merge/rebase time)**

`ReviewerChatSurface` (`frontend/src/renderer/components/chat/ReviewerChatSurface.tsx`, added on `main` by #4306 after this branch's rebase point at `850df06e`) renders `<ChatWorkspace .../>` without `onLinkOpen` or the new `onSessionLinkOpen` prop. `ChatMarkdown`'s `remarkSessionLinks` plugin still linkifies any raw `ao://sessions/...` text unconditionally, so a session link inside a reviewer message renders as a normal clickable link. But in `MarkdownLink`'s `onClick` (`ChatMarkdown.tsx`):

```ts
if (href && sessionLink) {
    event.preventDefault();
    onSessionLinkOpen?.(href);
    return;
}
```

`event.preventDefault()` runs unconditionally and the function returns before `AppLink`'s own fallback logic (which web links get) can run — so with `onSessionLinkOpen` undefined the click is fully swallowed: no navigation, no error toast, not even a fallback browser open. This contradicts the PR's own goal of surfacing malformed/missing/inaccessible targets as in-app feedback.

Since this branch was rebased onto `850df06e` before `ReviewerChatSurface` existed on `main`, this won't show up in the PR's diff, but whoever merges/rebases this PR will need to also wire `onSessionLinkOpen` (and ideally `onLinkOpen`) into `ReviewerChatSurface` to avoid shipping a silently-dead link affordance there.

PR #4610, opened 2026-08-28
feat(agent): discover provider model catalogs

nikhilachale:
## Summary

- replace release-time model lists with agent-owned discovery where reliable: Codex uses app-server `model/list`, Cline prefers ACP `session/new` config options, and supported agents use documented CLI or local configuration sources
- serve Claude Code's five non-default ACP picker choices from a side-effect-free static snapshot (`sonnet`, `fable`, `opus`, `haiku`, and `opus[1m]`) instead of starting the Claude Agent SDK; preserve configured aliases and pinned snapshots as synthetic defaults
- keep Muse's legacy static Spark catalog (`muse-spark`, `muse-spark-1.1`, and `muse-spark-1.2`) so catalog reads never start Muse, allocate a PTY, write `/model`, or create session logs
- normalize custom-model capability as `none`, `direct`, or `configured`, filter Goose choices to its active provider, and expose searchable picker, provider grouping, refresh guidance, and last-known selections consistently across task, switch-agent, and project-settings flows
- cache catalogs per agent and project in SQLite, preserve last-known-good data on discovery failure, coalesce concurrent loads, revalidate stale catalogs after six hours, and warm previously cached Claude/Muse scopes without blocking daemon readiness
- paginate Codex model discovery without opening a provider thread and package only the ACP runtime helpers still required by the desktop app

## Verification

- `go test -race ./internal/adapters/agent/modelcatalog`
- `go test -race ./internal/adapters/chatdriver/acp ./internal/adapters/chatdriver/codexappserver ./internal/adapters/runtime/ptyexec`
- `go test -race ./internal/service/agent ./internal/storage/sqlite/store ./internal/daemon`
- `go test ./internal/service/agent ./internal/session_manager`
- `npm run typecheck` (from `frontend/`)
- affected renderer component tests: 75 passed
- `npx vitest run scripts/build-acp-runtime-helpers.test.mjs --config vite.renderer.config.ts`
- `git diff --check`

## Notes

- Claude Code and Muse intentionally use manually refreshed static snapshots; catalog reads do not launch either agent, create transcripts, run hooks, allocate a PTY, or make provider requests.
- Amp remains a static execution-mode list because its values are modes rather than provider model IDs.


i-trytoohard:
Heads-up from #4702 (PR #4703, closed in favor of this rework): your diff keeps agy on `CombinedOutput` (`commandSpecs["agy"]` → `agy models`) and does not modify `modelDiscoveryError`.

When agy's launcher exits but a descendant process holds the output pipe open, `CombinedOutput` returns `exec.ErrWaitDelay` while the 20s discovery context is still live, and the raw sentinel leaks verbatim into the New Task dialog:

```
agy model discovery: exec: WaitDelay expired before I/O complete
```

`parseAgyModels` never runs in that path — the failure happens before parsing, so the new parser doesn't cover it.

Fix is a small mapping in `modelDiscoveryError` (`backend/internal/adapters/agent/modelcatalog/catalog.go`): match `errors.Is(commandErr, exec.ErrWaitDelay)` and return a human-readable timeout message instead of the sentinel. Commit `25537feaf` on branch `fix/model-discovery-waitdelay` has the change plus tests (sentinel-mapping unit test + a `/bin/sh` fixture that reproduces the pipe-holding shape) — free to cherry-pick. Same exposure applies to every other `commandSpecs` entry.

backend/internal/adapters/agent/modelcatalog/terminal.go
```diff
@@ -0,0 +1,364 @@
...
+	if spawn == nil {
+		return base, errors.New("interactive model discovery is unavailable")
+	}
+	if strings.TrimSpace(request.Binary) == "" {
+		return base, errors.New("agent binary is not installed")
+	}
+	args := []string{request.Binary}
+	switch request.AgentID {
+	case "claude-code":
+		args = append(args, "--ax-screen-reader")
```
illegalcall: [P1] This discovery is not invisible to Claude. Starting the normal interactive client can create a project-scoped JSONL transcript that remains visible through Claude session history and can change the session targeted by `--continue`, even though AO never creates an AO task. The same startup can invoke configured `SessionStart` hooks, so merely opening or refreshing the model catalog may execute user-defined commands and produce other observable side effects.

nikhilachale: Fixed in 251bf531e. Claude model discovery now runs with `--safe-mode` in a fresh temporary `CLAUDE_CONFIG_DIR`, closes the PTY, and removes the isolated state directory. This prevents configured SessionStart hooks from running and keeps discovery transcripts out of normal Claude history/`--continue`. Added regression coverage for the flags, environment isolation, and cleanup; `npm run lint` passes.

illegalcall (approved):

PR #4615, opened 2026-08-28
feat(mobile): connect from any network by racing every advertised endpoint

Prasad-D-Ware:
## What this does

Lets a phone reach the daemon from any network, without the user knowing what a LAN address or a tailnet is.

The QR previously carried a single address — in practice the LAN one. Off that network, pairing failed and the user landed in manual entry, Tailscale setup, or a certificate problem. The code now carries **every** endpoint the machine has, and the phone races them, verifying it reached the right machine before trusting any of them.

Same three onboarding steps as before. The change is that they now work from cellular.

## How it works

Three endpoint kinds are advertised and raced, preferred cheapest-first (`lan > tailscale > tunnel`):

- **LAN** — every reachable interface address, not just one
- **Tailscale** — when present
- **Tunnel** — a `cloudflared` quick tunnel the daemon supervises, giving reach from anywhere plus a publicly trusted certificate (which also satisfies iOS ATS and retired the MagicDNS/tailnet-certificate branch of the old flow)

The race verifies host identity on every candidate via a new `GET /api/v1/identity`, so a phone cannot be steered onto the wrong machine by a stale or reused address. Endpoints are refreshed on every successful connect, so a rotated tunnel hostname or a changed DHCP lease heals without re-pairing.

Which machine the app talks to is explicit: `activeHost()` answers with a selection where one has been made and the most recently used machine otherwise. Pairing and manual connect both set it; forgetting a machine clears it along with the machine and its token.

## Notable decisions

**One unauthenticated route.** `GET /api/v1/identity` returns only `{hostId, apiVersion}` and is exempt from the bearer-password middleware, because the race must be able to tell machines apart *before* it has a credential to offer. Documented in `docs/adr/0003-unauthenticated-identity-probe.md`, with `AGENTS.md` amended to name it as the single exemption; any further unauthenticated route needs its own ADR.

**Waiting for DNS happens on the phone.** A quick tunnel's hostname is registered before resolvers know it. The daemon cannot check that honestly — probing from the machine that owns the tunnel tests its own resolver, and probing early caches an NXDOMAIN locally that outlives propagation (measured: `curl` still failed 30s after `dig` had begun resolving the same name). So the daemon holds the hostname back briefly and the phone retries a tunnel candidate that does not resolve yet, which is where the deciding resolver actually lives.

**Installation identity survives hardware changes.** The host id belongs to the AO data directory, not to the current network hardware. Fingerprints are diagnostic metadata only, so docks and NIC replacement do not silently unpair every device. Copying the complete `~/.ao` directory also copies this identity; backups and migrations must therefore be protected, and resetting identity is an explicit removal of `~/.ao/mobile/identity.json` while AO is stopped.

**Cloudflare terminates TLS.** The phone's session is with Cloudflare, not with the user's machine, so tunnel traffic is encrypted in transit but not end to end. What that exposes, when the tunnel is chosen, and what would change it are in `docs/adr/0004-cloudflare-tunnel-for-remote-mobile-access.md`. The race preference is what keeps the tunnel to a last resort.

**v1 pairing codes are no longer accepted.** They predate the identity probe, so a daemon old enough to emit one answers 404 to every probe and the race can never complete — the compatibility was advertised but unreachable. A v1 code is still recognised, so the app says "update AO on your computer" rather than "this is not a pairing code".

## Known limitation: SSE does not stream over the tunnel

Measured against a live quick tunnel: response bodies are forwarded in ~128 KB chunks (a 121 KB replay produced nothing; larger ones arrived in exact multiples of 128 KB). A chat event is a few hundred bytes, so it is held indefinitely. WebSocket is unaffected — 11 ms round trips for small frames — because after the upgrade it is a raw tunneled stream rather than a buffered response body. That is why terminals felt normal over the tunnel while chat did not.

Two polling intervals compensate, **both stopgaps**, both active only while the tunnel is the chosen endpoint: the main poll drops 8s → 2s, and the chat screen polls its conversation, which it otherwise never does. They turn "over a minute, and only after reopening the screen" into roughly two seconds, at a cost in battery and data.

**The exit condition is concrete:** conversation events move to the existing `/mux` WebSocket, which already reaches the same daemon and already round-trips small frames in 11 ms over the tunnel. When that lands, both stopgaps are deleted.

## Verified on a physical device

Identity probe · race · migration from the single-server config · LAN → Tailscale → tunnel failover (~4s) · terminal following the raced endpoint · chat on all three paths · agent input prompts over the tunnel · sustained terminal output over the tunnel · tunnel crash detection and restart · hostname rotation picked up without re-pairing · orphaned connector reaped · enable/disable · images and navigation on a preview (production-bundle) build.

Four defects were found that way and fixed here: the app never returning to LAN once it fell to the tunnel; a black screen on open, caused by treating "no config yet" as "not configured" while the race resolved; a connector left running after the daemon quit; and endpoint lists losing a kind depending on when a refresh landed.

## Review round

- Migration lost the password for the oldest upgraders — it read only SecureStore, but older builds kept it in the config blob and the code that relocates it runs *after* migration
- "Disconnect & forget" did not forget: `removeHost` had no callers, so the machine and its token survived and the next launch reconnected to it
- A manual connection did not stick — it wrote only the legacy config, so resolution replaced it with the previous machine on reload
- A restarting connector could delete its replacement's pid file, leaving a live tunnel that boot reaping could never clean up
- Nothing installed `cloudflared`, so on a machine without it the QR looked normal and only ever worked on the local network. It is now a seventh `systeminstall` target, offered where the limitation is stated
- A dock or a new network card reissued the host id and unpaired every phone
- A rotated tunnel hostname produced a bare "can't connect" the phone could not recover from; it now names the one action that works
- Dead code removed (`connectionAction.ts`), and the multi-machine claim corrected — see below

## Not covered

- **There is no host switcher.** Up to 10 machines are stored with per-host tokens and per-host event cursors, and `activeHost()` owns the selection — but nothing in the UI lets you choose one, so a second machine is reachable by pairing with it, not by picking it from a list. The plumbing is here; the picker is not.
- **Per-device revocable tokens.** There is one shared connection password, and removing a phone from the device roster deletes its registry row without revoking access — it can reconnect immediately. This predates the PR, but the PR raises the stakes since pairing now works from anywhere. Tracked as a bug; the fix changes `authMiddleware` and needs a migration, so it belongs in its own PR.
- **A provider interface for endpoint sources**, deliberately deferred: there is no second provider today, and a relay we run ourselves is a real possibility whose shape would likely reshape the interface.
- **Named tunnel mode**, for the same reason — a stable hostname would remove the rotation problem, but it needs an account and a domain, and we have not settled on Cloudflare as the destination.
- **`cloudflared` on Linux** resolves to instructions rather than an install, like every other Linux target, because AO never asks for an administrator password.
- **Battery and data cost** of the 2s tunnel poll is unmeasured.
- **Secure pairing** is not carried in a v2 code: the endpoint list only ever marks the tunnel secure. Whether that matters depends on whether iOS allows plain HTTP to a `100.x` tailnet address in a release build, which is untested. Recorded as a test documenting current behaviour rather than asserted away.
- **Network-change detection** takes 9–33s (poll interval plus request timeout). Root-caused; the fix needs `expo-network`, which changes the native fingerprint, so it is deferred.

## Testing

`go test ./...`, `npm test` in `packages/mobile`, frontend suites, `npm run lint` clean, and `npm run api` reproduces the committed spec with no diff.


AgentWrapper (approved):

illegalcall (commented):
A few broader follow-ups are worth tracking: one ConnectionManager should own the active host and route; endpoint sources should use a provider interface so Cloudflare, Ngrok, or another option can plug in; and paired devices should eventually get separate revocable tokens. I kept those here because they are not tied to one exact line.

packages/mobile/lib/hosts.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Forgets a machine.
+ *
+ * Both tiers are cleared: wiping only the AsyncStorage entry would leave the
+ * token in the keystore, and a later re-pair of the same machine would silently
+ * resurrect it.
+ */
+export async function removeHost(id: string): Promise<void> {
```
illegalcall, line 163: `removeHost` is not called by the current Disconnect & forget flow; that still calls `clearConfig()`. The host record and `ao.hostToken.<id>` survive, so `resolveActiveConfig()` can reconnect. Can we wire this to the selected host and clear its cursor and active connection too? A reload-after-forget test would catch it.

packages/mobile/lib/connectRuntime.ts
```diff
@@ -0,0 +1,88 @@
...
+ * docs/adr/0003-unauthenticated-identity-probe.md.
+ */
+export async function probeEndpoint(endpoint: Endpoint, signal: AbortSignal): Promise<ProbeAnswer> {
+	const controller = new AbortController();
+	const timeout = setTimeout(() => controller.abort(), PROBE_TIMEOUT_MS);
+	// Abort when either the race picked a winner or our own timeout fired.
+	const onOuterAbort = () => controller.abort();
+	signal.addEventListener("abort", onOuterAbort);
+	try {
+		const res = await fetch(`${endpointBaseUrl(endpoint)}/api/v1/identity`, {
```
illegalcall, line 67: V1 QR codes have no `hostId`, but they still come through this probe. An older daemon returns 404 here, so the advertised V1 compatibility never reaches authenticated verification. Can we add a real V1 fallback or remove the compatibility claim?

backend/internal/mobilebridge/tunnel_runner.go
```diff
@@ -0,0 +1,182 @@
...
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
+	for scanner.Scan() {
+		line := scanner.Text()
+		t.Runtime.Line(line)
+		t.logger().Debug("cloudflared", "line", line)
+	}
+
+	waitErr := cmd.Wait()
+	t.Runtime.Exited(waitErr)
+	_ = RemoveTunnelPIDFile(t.PIDPath)
```
illegalcall: This removes a shared PID path unconditionally. `Start` cancels the old runner and starts its replacement without waiting; the replacement can write its PID before the old runner reaches this line, then the old runner deletes the new PID. Can we remove it only when the file still contains `cmd.Process.Pid`, or wait for the old runner before replacing it?

packages/mobile/lib/resolveConfig.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function resolveActiveConfig(deps: ResolveDeps): Promise<ServerConfig | null> {
+	try {
+		// Before looking for machines, bring any pre-existing single-server
+		// pairing into the list — otherwise an upgrading user looks unpaired.
+		await deps.migrate();
+
+		const hosts = await deps.loadHosts();
+		if (hosts.length > 0) {
+			// loadHosts is ordered most-recent-first.
+			const result = await deps.connect(hosts[0].id);
```
illegalcall: `ManualConnectSheet` still writes the legacy `ServerConfig`. Once the host list exists, migration skips and this reconnects `hosts[0]`, so a manual connection can snap back after reload. Can manual connect upsert/touch a host and make it active?

backend/internal/daemon/daemon.go
```diff
@@ -464,8 +465,40 @@ func Run() error {
...
+	}
+	if res := mobilebridge.ResolveCloudflared(mobilebridge.LocalCloudflaredLookup(cfg.DataDir)); !res.NeedsInstall {
+		log.Info("mobile remote access available", "cloudflared", res.Path, "source", res.Source)
+		bs.Tunnel = mobilebridge.NewManagedTunnel(mobilebridge.ManagedTunnelDeps{
+			Binary: res.Path, PIDPath: tunnelPID, Log: log,
+		})
+	} else {
+		// Not fatal: the LAN and Tailscale endpoints still work, and Connect
+		// Mobile behaves exactly as it did before remote access existed.
+		log.Info("mobile remote access unavailable; cloudflared not installed",
```
illegalcall, line 499: On a fresh install, `NeedsInstall` only reaches this log; there is no installer or bundled binary in this diff. The UI treats a nil tunnel as a valid LAN-only QR, so the cellular flow is silently absent. What installs the managed `cloudflared` copy? If nothing does, can status expose an actionable unavailable state?

packages/mobile/lib/hosts.ts
```diff
@@ -0,0 +1,203 @@
...
+		await SecureStore.deleteItemAsync(tokenKey(host.id));
+	}
+}
+
+/**
+ * Replaces a machine's endpoint list, leaving its token alone.
+ *
+ * Called after every successful connect with whatever the daemon now
+ * advertises, so a rotated tunnel hostname or a changed LAN address is picked
+ * up without the user doing anything.
```
illegalcall, line 107: This only heals after a successful connection. If a Quick Tunnel URL rotates while the phone is remote and LAN/Tailscale are unavailable, every cached endpoint is dead, so refresh cannot run. The PR body already calls this out; can we make the required rescan explicit in the UI and link the stable-hostname follow-up?

backend/internal/mobilebridge/identity.go
```diff
@@ -0,0 +1,111 @@
...
+		if json.Unmarshal(b, &existing) == nil &&
+			existing.HostID != "" &&
+			existing.Fingerprint == fingerprint {
+			return existing, nil
+		}
+	} else if !os.IsNotExist(err) {
+		return Identity{}, fmt.Errorf("read mobile identity: %w", err)
+	}
+
+	issued := Identity{HostID: "h_" + uuid.NewString(), Fingerprint: fingerprint}
```
illegalcall: Any fingerprint mismatch issues a new `hostId`. Adding or removing a dock/USB NIC, replacing a network card, or migrating `~/.ao` can invalidate every pairing. Could we preserve the stored ID, warn on fingerprint change, and provide an explicit reset? The tradeoff is that copied data shares an identity until the user resets it, but silent re-pairing seems less predictable.

packages/mobile/lib/hosts.ts
```diff
@@ -0,0 +1,203 @@
...
+	name: string;
+	platform: string;
+	endpoints: Endpoint[];
+	/** Connection token. Lives in the device keystore, never in AsyncStorage. */
+	token: string;
+	lastConnected: number;
+};
+
+/** Enough for a developer's machines without letting stale pairings pile up. */
+export const MAX_HOSTS = 10;
```
illegalcall, line 34: We retain 10 hosts, but runtime always selects the most recent one and there is no host selector. Is a switcher planned as follow-up? As written, the older hosts are only reachable by scanning their QR again, so “multiple machines supported end to end” needs clarification.

backend/internal/mobilebridge/tunnel_runner.go
```diff
@@ -0,0 +1,182 @@
...
+		return t.SettleDelay
+	}
+	return DefaultTunnelSettleDelay
+}
+
+const (
+	// DefaultTunnelSettleDelay is how long after registration the hostname is
+	// treated as resolvable. Measured propagation was ~20s; the margin covers
+	// a slower run without making the wait obviously long.
+	DefaultTunnelSettleDelay = 25 * time.Second
```
illegalcall: This 25s value comes from one environment. The daemon-side probe has the NXDOMAIN-cache problem described above, but could mobile retry the tunnel candidate for a bounded window after scanning? That tests the phone’s resolver and avoids treating a slower propagation as a dead endpoint.

packages/mobile/lib/pollInterval.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** Direct paths stream fine, so the poll stays cheap on battery and data. */
+export const DIRECT_POLL_MS = 8_000;
+
+/**
+ * Over the tunnel the poll is the only live signal, so it has to be quick
+ * enough to feel immediate. Deliberately a stopgap: it costs battery and data,
+ * and it applies only while that path is in use.
+ */
+export const TUNNEL_POLL_MS = 2_000;
```
illegalcall, line 32: Since this is explicitly a stopgap, can we link the tracking issue or exit condition for moving these events to `/mux`? The battery/data cost is unmeasured, and an untracked 2s poll is likely to become permanent.

packages/mobile/lib/hosts.ts
```diff
@@ -0,0 +1,203 @@
...
+	// A Tailscale pairing is the only way the old flow produced a TLS endpoint,
+	// so secure implies the tailnet path; everything else was plain LAN.
+	const kind = isSecure ? "tailscale" : "lan";
+
+	await saveHost({
+		id: "",
+		name: host,
+		platform: "",
+		endpoints: [{ kind, host, port, secure: isSecure }],
+		token: (await SecureStore.getItemAsync(LEGACY_PASSWORD_KEY)) ?? "",
```
illegalcall: This reads only the SecureStore password. Older configs can still have `password` inside `ao.serverConfig`; `loadConfig()` migrates that value, but `migrateLegacyConfig()` runs first. Those users get `token: ""` and fail authentication. Can we use `legacy.password` as the fallback and remove it, or run the existing credential migration first?

backend/internal/mobilebridge/tunnel.go
```diff
@@ -0,0 +1,442 @@
...
+//
+// The connector targets loopback: sending it at the LAN bind address would
+// route tunnel traffic back out over the network for no reason. Autoupdate is
+// off because the hostname is only available by scraping the log, and a
+// self-update could swap the binary out from under that pinned format. The
+// metrics listener is what /ready is polled on, which is a far more reliable
+// readiness signal than the log alone.
+func CloudflaredArgs(localPort, metricsPort int) []string {
+	return []string{
+		"tunnel",
```
illegalcall, line 434: This always starts a Quick Tunnel, even when the user has a configured Cloudflare account. Is a stable Named Tunnel mode intentionally out of scope? Quick access makes sense as the zero-config default, but a user-managed stable endpoint would avoid hostname rotation and the SSE limitation. Please link the follow-up if that is the plan.

packages/mobile/lib/connectionAction.ts
```diff
@@ -0,0 +1,72 @@
...
+export const BACKGROUND_SUSPENSION_MS = 30_000;
+
+/**
+ * Decides what a wake should do.
+ *
+ * The important case is foregrounding a healthy session: it is checked, not
+ * replaced. Reconnecting on every foreground is what makes an app spin for a
+ * few seconds each time it is opened, and the session is usually fine.
+ */
+export function actionFor(trigger: ConnectionTrigger, state: ConnectionState): ConnectionAction {
```
illegalcall: This policy has no production callers; only its tests reference it. Is wake handling meant to use it in this PR? If yes, it still needs wiring. If it belongs to the future ConnectionManager work, it would be clearer to defer the dead abstraction.

backend/internal/mobilebridge/tunnel.go
Prasad-D-Ware: Named tunnel mode is intentionally out of scope here, and we'd like to keep it that way for now.

The quick tunnel stays the zero-config default. A stable hostname is clearly better — it removes the rotation lockout outright and is worth re-measuring against the ~128KB SSE buffering, since that finding is quick-tunnel-only — but it needs a Cloudflare account and a domain, so it can only ever be opt-in.

We're also not certain the destination is Cloudflare. A relay we run ourselves is on the table, and it would let payloads be encrypted before they leave the device, which is the only thing that changes the trust model in ADR 0004. Committing to named-tunnel plumbing now would be building toward one of two futures before we've picked.

What did land in this PR for the rotation problem: the phone now recognises "nothing answered and one of the paths was a tunnel" and tells the user to rescan, rather than showing a bare "can't connect" it cannot recover from.

Prasad-D-Ware:
Thanks — these three are the ones we've thought about most, and we've landed differently on each.

**One ConnectionManager owning the active host and route.** Agreed, and this turned out to be the root cause of two of the inline findings rather than a separate idea: `removeHost` having no caller and manual connect snapping back were both symptoms of nothing owning the selection — `loadHosts` was ordered most-recent-first and callers took the head. There is now an explicit `activeHost()`, with a selection where one has been made and recency as the fallback, and forgetting a machine clears the machine, its token and the selection. That is the small version of what you're describing: enough to own the selection, without moving routing yet.

**A provider interface for endpoint sources.** Not yet, deliberately. There is no second provider today, and we may not stay on Cloudflare at all — a relay we run ourselves is a real possibility, and its shape is different enough (app-layer encryption, our own addressing) that an interface designed around cloudflared would likely need redesigning the moment it arrived. `EndpointKind` already reserves `relay` and `endpointRank` sorts unknown kinds last, so the current shape isn't hostile to it. We'd rather build the abstraction when there's a second implementation to shape it.

**Per-device revocable tokens.** This is the most valuable item on the list and we want it — but as its own PR, not this one. Two reasons. It changes `authMiddleware`, which this PR was careful to touch exactly once and only behind ADR 0003, and security work deserves a review that isn't sharing space with 30 other commits. It also needs a migration for every existing pairing.

Worth recording while looking at it: today there is one shared connection password, and `MobileDevicesController.Remove` only deletes the registry row. So removing a phone from the roster does not revoke its access — it reconnects and re-registers. That predates this PR, but this PR raises the stakes, since pairing now works from anywhere rather than only on the local network. We're tracking it as a bug rather than an enhancement, because the UI currently implies something it doesn't do.

Everything else from the inline comments is fixed in the commits above.

backend/internal/mobilebridge/tunnel_runner.go
illegalcall: This still has a TOCTOU race. The old runner can read its own PID, the replacement can write its PID, then the old runner removes the replacement's file. The test only covers the replacement write happening before the read. Can start and cleanup be serialized under one owner/generation, or use generation-scoped PID files?

backend/internal/mobilebridge/identity.go
illegalcall: This helps only while the old and current sets share one NIC. Replacing the only NIC still changes hostId. Replacing the stored set also loses history: [wifi, dock] -> [dock] -> [wifi] rekeys on the last step. Can we keep the union of known fingerprints, or keep hostId stable and make reset explicit?

packages/mobile/lib/hosts.ts
illegalcall: Host, token, active pointer, and config are now cleared on the normal path. Two gaps remain: the per-host event cursor is not removed, and a removeHost rejection stops clearConfig and onboarding cleanup from running. Can local cleanup run independently and clear the cursor too?

packages/mobile/app/pair.tsx
```diff
@@ -75,50 +79,63 @@ export default function PairScreen() {
...
-			mobileTelemetry()?.capture(MOBILE_EVENTS.paired, { method: "qr", from_onboarding: fromOnboarding });
-			if (fromOnboarding) mobileTelemetry()?.capture(MOBILE_EVENTS.onboardingCompleted);
-			haptics.success();
-			await finish();
-		} catch (e) {
+
+		const result = await pairFromCode(code, {
+			race: (endpoints, expectedHostId) => raceEndpoints(endpoints, expectedHostId, probeEndpoint),
+			verify: async (config) => void (await pingServer(config)),
+			saveHost,
```
illegalcall, line 110: saveHost does not select the new host. If machine A was explicitly active, scanning machine B stores B but reload still resolves A. Can pairing call setActiveHost(result.host.id) before reload, with a two-host test?

frontend/src/renderer/components/settings/ConnectMobileContent.tsx
```diff
@@ -394,36 +533,55 @@ export function ConnectMobileContent({ active }: { active: boolean }) {
...
+						<div className="mt-3">
+							<p className="text-xs text-settings-muted" data-testid="mobile-remote-unavailable">
+								{t(
+									"mobile.remoteAccessUnavailable",
+									"Works on this network only — cloudflared isn't installed, so this machine can't be reached from elsewhere.",
+								)}
+							</p>
+							{/* Re-enabling is what makes the daemon look for the binary
+							    again, so a fresh install takes effect without restarting. */}
+							<InstallCloudflared onInstalled={() => void query.refetch().then(() => enable.mutate())} />
```
illegalcall: onInstalled calls Enable again. Enable always generates a new password, so installing cloudflared while the bridge is already enabled invalidates the current phone. Can we refresh/start only the tunnel, or make Enable keep the existing password when already enabled?

backend/internal/androidemulator/serial.go
```diff
@@ -0,0 +1,81 @@
...
+
+// ResolveSerial finds the adb serial of the running emulator whose AVD is
+// avdName.
+//
+// AO cannot assume its device is emulator-5554: that is only true when AO's
+// emulator is the first one running on the host. A developer with their own
+// AVD already booted pushes AO's device to 5556 or higher, and a hardcoded
+// serial then silently targets *their* device -- for readiness polling, UI
+// inspection, and anything else keyed off the serial.
+func ResolveSerial(ctx context.Context, adbPath, avdName string) (string, bool) {
```
illegalcall: ResolveSerial has no production caller and is unrelated to the endpoint-race fixes. This PR also removes connectionAction as dead code. Can this file and test move to a separate PR with the actual wiring?

illegalcall (commented):
Rechecked the fix commits. These are the remaining code issues I found. The PR is also currently conflicted with main and has no GitHub checks reported.

packages/mobile/lib/hosts.ts
Prasad-D-Ware: Fixed in 700231b15. Forget now clears the per-host event cursor and runs host/token removal, cursor removal, config clearing, and onboarding reset independently via allSettled, so one SecureStore rejection cannot prevent the other local cleanup. The original error still surfaces. Added the rejection regression test.

backend/internal/androidemulator/serial.go
Prasad-D-Ware: Fixed in 700231b15. Removed the unrelated androidemulator serial helper and its tests from this PR; it had no production caller.

frontend/src/renderer/components/settings/ConnectMobileContent.tsx
Prasad-D-Ware: Fixed in 700231b15. Installation now calls a dedicated /api/v1/mobile/remote-access action that re-resolves and starts only the connector on the already-bound port. It does not call Enable and therefore does not rotate the pairing password. Backend coverage verifies the paired phone's password stays unchanged.

backend/internal/mobilebridge/tunnel_runner.go
Prasad-D-Ware: Fixed in 700231b15. PID claim and cleanup now go through a path-scoped owner lock shared by every runner generation, so the complete read/remove transaction is serialized against a replacement write. Added a regression test proving owners for the same PID path share the lock.

backend/internal/mobilebridge/identity.go
Prasad-D-Ware: Fixed in 700231b15. Host identity now belongs to the AO data directory and no longer rotates based on NIC overlap. Hardware fingerprints remain diagnostic metadata and update across changes; resetting identity is explicit by removing the identity file. Added coverage for replacing the only interface.

packages/mobile/app/pair.tsx
Prasad-D-Ware: Fixed in 700231b15. Successful QR pairing now saves the host and explicitly selects its host ID before returning. Added a pair-flow test showing a newly scanned second machine becomes active.

packages/mobile/lib/disconnect.ts
```diff
@@ -19,12 +21,28 @@ import { unpairFromServer } from "./push";
 // still on disk. Disconnecting is the one operation that must not leave
 // credentials behind: whatever happens upstream, the config gets cleared.
 export async function forgetServer(): Promise<void> {
+	// The host record and its token are the actual pairing now; the legacy
+	// config is only the last resolved address. Clearing that alone left the
+	// machine in the list with its token in the keystore, so the next launch
+	// raced its endpoints and silently reconnected to the server just forgotten.
+	const host = await activeHost();
```
illegalcall: `activeHost()` runs before the guarded cleanup. It reads every host token from SecureStore, so one read failure exits here and skips unpair, config cleanup, and onboarding cleanup. Can this read only the active host id, or can this failure still fall through to the safe local cleanup?

packages/mobile/lib/chat/eventCursor.ts
```diff
@@ -0,0 +1,56 @@
...
+ * connects once, so it falls back to the address.
+ */
+export function eventCursorKey(cfg: ServerConfig): string {
+	if (cfg.hostId) return `ao.chat.events.host.${cfg.hostId}`;
+	return `ao.chat.events.${cfg.secure ? "https" : "http"}.${cfg.host}.${cfg.httpPort}`;
+}
+
+/** Remove replay state when the user forgets a machine. */
+export async function clearEventCursorForHost(hostId: string): Promise<void> {
+	await AsyncStorage.removeItem(`ao.chat.events.host.${hostId}`);
```
illegalcall: This only removes the host-keyed cursor. A migrated/manual host can still have an empty `hostId`, where `eventCursorKey` uses the address key instead. Forget then removes `ao.chat.events.host.` and leaves the real cursor. Can cleanup use the same key function, with an id-less host test?

docs/adr/0003-unauthenticated-identity-probe.md
```diff
@@ -0,0 +1,73 @@
...
+The exemption is deliberately narrow, and there are tests for each constraint:
+
+- **Exact path match**, not a prefix — nothing nested below `/api/v1/identity`
+  inherits it.
+- **`GET` only** — a write to the same path still requires the password.
+- **Checked ahead of the lockout** — a phone racing several endpoints must not
+  be able to lock itself out by probing, and an unauthenticated probe carries no
+  credential to get wrong.
+
+The host id is persisted at `~/.ao/mobile/identity.json` and bound to a
```
illegalcall: This now says the opposite of the implementation. `EnsureIdentityFor` keeps `HostID` across fingerprint changes, so copying `~/.ao` copies the identity too. Please update this ADR, the machine-bound code comments, and the PR description to document the chosen installation-identity tradeoff.

frontend/src/renderer/components/settings/InstallCloudflared.tsx
```diff
@@ -0,0 +1,99 @@
...
+			{view.kind === "running" && (
+				<p className="flex items-center gap-1.5 text-xs text-settings-muted">
+					<Loader2 className="size-3 shrink-0 animate-spin" aria-hidden="true" />
+					{t("mobile.installingConnector", "Installing cloudflared…")}
+				</p>
+			)}
+			{view.kind === "manual" && (
+				<>
+					<p className="text-xs text-settings-muted">
+						{view.reason || t("mobile.installConnectorManual", "Run this in a terminal, then try again.")}
```
illegalcall, line 88: On Linux this is a terminal UI state. After the user runs the command, there is no action here that calls `onInstalled` or asks the daemon to resolve the binary again. Can we add a **Check again** action and cover the manual-install path?

illegalcall (commented):
Follow-up pass after bb883d1.

docs/adr/0003-unauthenticated-identity-probe.md
Prasad-D-Ware: Fixed in 89f595138. ADR 0003 and the identity code/test comments now describe an installation identity: hardware fingerprints are diagnostic and never rotate HostID; copying AO_DATA_DIR copies identity; reset is explicit by removing identity.json while AO is stopped. The PR description has been updated to state the same tradeoff.

frontend/src/renderer/components/settings/InstallCloudflared.tsx
Prasad-D-Ware: Fixed in 89f595138. The Linux manual-command state now includes a localized Check again action. It invokes the existing password-preserving remote-access callback, causing the daemon to re-resolve cloudflared and start the connector without re-enabling the bridge. Added a component test for the manual-install path.

packages/mobile/lib/disconnect.ts
Prasad-D-Ware: Fixed in 89f595138. Disconnect now resolves only non-secret active-host metadata from AsyncStorage; it no longer opens every host token in SecureStore. Metadata lookup is itself guarded, so even a storage rejection cannot skip unpair, config cleanup, or onboarding cleanup. Added regressions for both properties.

packages/mobile/lib/chat/eventCursor.ts
Prasad-D-Ware: Fixed in 89f595138. Cursor cleanup now accepts the stored host and derives every key through eventCursorKey. Identified hosts remove the host key; id-less migrated/manual hosts remove address keys for their stored endpoints and never attempt ao.chat.events.host. Added the id-less-host regression test.

illegalcall (approved):

PR #4618, opened 2026-08-29
fix: sign packaged Node runtime for JIT so Chat sessions start on Intel Macs

### LINKED ISSUE #4442: Chat-driven orchestrator spawn fails: CHAT_DRIVER_INCOMPATIBLE — Claude Code CLI has no ACP support
## Summary

Spawning a chat-driven Orchestrator session fails immediately with `CHAT_DRIVER_INCOMPATIBLE` because Agent Orchestrator appears to spawn the `claude` CLI directly and expect it to speak ACP (Agent Client Protocol) — but the Claude Code CLI does not implement ACP at all, in any published version.

## Error

```
spawn monorepo-1: chat controller: chat driver incompatible: ACP initialize:
{"code":-32603,"message":"Internal error","data":{"error":"peer disconnected before response"}} (CHAT_DRIVER_INCOMPATIBLE)
```

(Screenshot of the app's error dialog to follow in a comment.)

Image

## Environment

- Agent Orchestrator desktop app (macOS, Darwin 24.6.0), `ao daemon`
- Claude Code CLI `2.1.245` — confirmed via `npm view @anthropic-ai/claude-code version` that this is the latest published version, not a stale install
- Project registered as a `single_repo` with `orchestrator.agent = "claude-code"`

## Root cause

The installed Claude Code CLI has no ACP support whatsoever:

```
$ claude --help | grep -i acp
(no output — zero mentions across the full help text)

$ claude acp
✘ unknown command "acp"
  └ Did you mean claude mcp?
```

This isn't a version-skew issue — there is no `claude acp` subcommand, `--acp` flag, or documented/undocumented env var exposing an ACP server mode in the Claude Code CLI, in its current or past published releases.

ACP support for Claude is instead implemented by a **separate, officially maintained bridge binary**: `@agentclientprotocol/claude-agent-acp` (previously published as `@zed-industries/claude-agent-acp`, now deprecated in favor of the new package name), which wraps the Claude Agent SDK and exposes ACP over stdio:

```
npm install -g @agentclientprotocol/claude-agent-acp
ANTHROPIC_API_KEY=sk-... claude-agent-acp
```

It looks like the "chat driver" here spawns `claude` directly and performs an ACP `initialize` handshake against it, which will always fail — `claude` exits immediately on the unrecognized interaction, surfacing as "peer disconnected before response".

## Suggested fix

Let the chat-driver / orchestrator spawn path target `claude-agent-acp` instead of (or in addition to) `claude` — e.g. auto-detect `claude-agent-acp` on `PATH`, or add a configurable "ACP agent binary" setting per project/harness — rather than assuming the Claude Code CLI itself speaks ACP.

## Repro steps

1. Register a project with `orchestrator.agent = "claude-code"`.
2. Click "Spawn Orchestrator" (chat-driven mode) from the project board.
3. Observe immediate failure with the `CHAT_DRIVER_INCOMPATIBLE` error above.
4. As a workaround, "Create as Terminal UI" works fine — it drives Claude Code via plain CLI flags (`--resume`, `--append-system-prompt-file`, etc.) instead of ACP.

muhammadmahad-debug:
## What

Fixes the Chat-driven Claude Code session failing immediately on Intel Macs with `CHAT_DRIVER_INCOMPATIBLE`, and makes the next runtime failure of this shape report itself instead of surfacing as a bare disconnect.

Two commits:

1. Sign AO's packaged Node runtime with `com.apple.security.cs.allow-unsigned-executable-memory`.
2. Retain the ACP adapter's stderr tail and fold it into the initialize error.

## Why

Fixes #4442.

That issue's original diagnosis (AO spawns `claude` and expects it to speak ACP) is **wrong** — I filed it and have corrected the record [in a comment](https://github.com/Untrivial-ai/agent-orchestrator/issues/4442#issuecomment-5459101802). AO already packages `@agentclientprotocol/claude-agent-acp` and passes `CLAUDE_CODE_EXECUTABLE`; that code needed no change.

The real cause: **AO's packaged Node crashes on Intel Macs.** It aborts inside V8 on first tier-up:

```
# Fatal error in , line 0
# Check failed: 12 == (*__error()).
 3: v8::base::OS::SetPermissions(...)
15: v8::internal::GenerateBaselineCode(...)
```

`@electron/osx-sign`'s default entitlements grant `allow-jit` but not `allow-unsigned-executable-memory`. Electron's own Chromium build tolerates that because it maps JIT pages with `MAP_JIT`; a stock nodejs.org build does not. On x86_64 V8's baseline compiler makes its code pages executable with plain `mprotect`, which the hardened runtime refuses without that entitlement. arm64 maps JIT pages differently and is unaffected — which is why this shipped.

Isolating the signature alone on the shipped 0.10.3 runtime:

| Signing config | Result |
| --- | --- |
| As shipped | crash |
| Signature removed | works |
| ad-hoc + hardened runtime + `allow-jit` only | crash (exact reproduction) |
| ad-hoc + hardened runtime + `allow-jit` + `allow-unsigned-executable-memory` | works |

The second commit exists because the crash was invisible: `spawnAgent` drained the adapter's stderr into `io.Discard`, so the V8 trace was written and thrown away, leaving only `peer disconnected before response`. That is what sent diagnosis toward the wrong subsystem.

## How

**Entitlements** — `optionsForFile` scopes the override to `acp-runtime/node/bin/node` by exact path segments. Every other file returns `{}` so `@electron/osx-sign` keeps applying its own defaults, notably the inherit entitlements Electron's helper apps require; a blanket `entitlements` override would break their signature. The new plist reproduces the signer's macOS defaults plus the one missing key, so nothing the runtime had today is dropped.

**stderr** — draining stays unbounded, so a chatty adapter still cannot fill its OS pipe and deadlock the protocol. Only what is *retained* is capped (8 KiB, last 12 lines in the message). `conv.Close()` is called before reading the tail so the process has driven its stderr pipe to EOF.

Intentionally **not** in this PR, both called out in the issue comment:

- A crashed adapter is still classified `CHAT_DRIVER_INCOMPATIBLE`. It is not a protocol mismatch, but correcting the classification is a behavioural change that deserves its own PR.
- `claudeacp`'s probe runs `node --version`, which exits 0 on a runtime that cannot execute JavaScript. Tightening the probe is likewise separable.

## Testing

**The entitlements fix is verified end-to-end against the real artifact**, not just unit-tested: re-signed the shipped 0.10.3 `acp-runtime` Node with the committed plist under `--options runtime`, then drove the shipped adapter through a genuine ACP `initialize`. It returns a valid result instead of aborting.

```
backend:  go build ./...                                    OK
          gofmt -l internal/adapters/chatdriver/acp/         clean
          go vet ./internal/adapters/chatdriver/acp/          OK
          golangci-lint v2.12.2 (changed package)             0 issues
          go test -race ./internal/adapters/chatdriver/...     all ok
frontend: tsc --noEmit                                        OK
          vitest forge.config.test.ts                         8 passed
```

New tests: `process_test.go` covers stderr retention through a real spawned process, the tail cap, and the error formatting. `forge.config.test.ts` gains coverage that the predicate matches only the packaged runtime and leaves Electron's helpers alone.

Two pre-existing failures on this machine, confirmed against a clean `main` and unrelated to these changes: `TestFullLifecycleSpawnToTermination` (timing-sensitive, fails under full-suite load on slower hardware) and `src/main/browser-view-host.test.ts`.

Verification was done on the affected hardware — Intel Core i7, macOS 15.7.7.

## Checklist

- [x] Branched from `main` (rebased on `upstream/main` @ ae890182d)
- [x] Relevant CI checks pass for the area touched (`go`, frontend)

### DIFF
--- a/backend/internal/adapters/chatdriver/acp/driver.go
+++ b/backend/internal/adapters/chatdriver/acp/driver.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log/slog"
 	"path/filepath"
+	"strings"
 	"time"
 
 	acpsdk "github.com/coder/acp-go-sdk"
@@ -346,11 +347,16 @@ func (d *Driver) connect(
 		},
 	})
 	if err != nil {
+		// Close first: stopping the process drives its stderr pipe to EOF, so the
+		// retained tail is complete before it is folded into the error.
 		_ = conv.Close()
 		if isACPAuthRequired(err) {
 			return nil, acpsdk.InitializeResponse{}, normalizeACPError("ACP initialize", err)
 		}
-		return nil, acpsdk.InitializeResponse{}, fmt.Errorf("%w: ACP initialize: %w", ports.ErrChatDriverIncompatible, err)
+		return nil, acpsdk.InitializeResponse{}, fmt.Errorf(
+			"%w: ACP initialize: %w%s",
+			ports.ErrChatDriverIncompatible, err, formatStderrTail(proc.stderrSnapshot()),
+		)
 	}
 	if d.cfg.ValidateInitialize != nil {
 		if err := d.cfg.ValidateInitialize(init); err != nil {
@@ -359,6 +365,27 @@ func (d *Driver) connect(
 		}
 	}
 	return conv, init, nil
+}
+
+// stderrTailLines caps how much adapter stderr reaches the error message. A
+// fatal runtime trace is long; the last lines carry the cause and keep the
+// surfaced error readable in the UI.
+const stderrTailLines = 12
+
+// formatStderrTail renders retained adapter stderr as an error suffix. An
+// adapter that exits before answering initialize leaves no other diagnostic,
+// which previously reduced a crashed runtime to "peer disconnected before
+// response" (#4442).
+func formatStderrTail(tail string) string {
+	tail = strings.TrimSpace(tail)
+	if tail == "" {
+		return ""
+	}
+	lines := strings.Split(tail, "\n")
+	if len(lines) > stderrTailLines {
+		lines = lines[len(lines)-stderrTailLines:]
+	}
+	return fmt.Sprintf(" (adapter stderr: %s)", strings.Join(lines, "; "))
 }
 
 func cloneCapabilities(in ports.ChatCapabilities) ports.ChatCapabilities {
--- a/backend/internal/adapters/chatdriver/acp/process.go
+++ b/backend/internal/adapters/chatdriver/acp/process.go
@@ -11,10 +11,54 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/processenv"
 )
 
+// stderrTailLimit bounds the retained adapter stderr. A runtime that dies on
+// startup prints a multi-frame fatal trace; keeping the tail captures the fatal
+// message without letting a chatty adapter grow the buffer without bound.
+const stderrTailLimit = 8 << 10
+
+// stderrTailWait bounds how long a caller blocks for the drain goroutine to see
+// EOF. The adapter has normally already exited by the time the tail is read.
+const stderrTailWait = 250 * time.Millisecond
+
 type process struct {
 	stdin  io.WriteCloser
 	stdout io.Reader
-	stop   func() error
+	// stderrTail returns the retained tail of the adapter's stderr. An adapter
+	// that dies before answering a request reports why here and nowhere else,
+	// so the handshake path folds it into the returned error.
+	stderrTail func() string
+	stop       func() error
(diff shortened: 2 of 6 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

muhammadmahad-debug:
@somewherelostt  This is still pending

codebanditssss:
Thanks for the contribution. This is already implemented more completely on current  by #3896: the packaged ACP runtimes are covered by the macOS JIT/unsigned-executable-memory entitlements, and the release verification script checks those entitlements. This PR is therefore no longer necessary.

PR #4622, opened 2026-08-29
fix(sqlc): repair the hand-written go_type overrides in sqlc.yaml

### LINKED ISSUE #4621: sqlc regenerate breaks the build: app_settings.cloud_offering hand-typed as bool without a sqlc.yaml override
## What's wrong

Running the documented regen command on a pristine `main` breaks the build:

\`\`\`
cd backend && go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.31.1 generate
go build ./...
# internal/storage/sqlite/store/app_settings_store.go:41:23: cannot use row.CloudOffering (variable of type int64) as bool value in struct literal
# internal/storage/sqlite/store/app_settings_store.go:67:18: cannot use enabled (variable of type bool) as int64 value in struct literal
\`\`\`

## Root cause

`app_settings.cloud_offering` is declared as a plain `INTEGER NOT NULL DEFAULT 0 CHECK (cloud_offering IN (0, 1))` (migration `0112_app_settings_cloud_offering.sql`). sqlc cannot infer `bool` from that shape without an explicit override — and unlike its sibling boolean-ish columns (`pr_comment.resolved`, `sessions.auto_review_enabled`, `agent_switches.semantic_handoff_included`), `backend/sqlc.yaml` never got a `column: "app_settings.cloud_offering" go_type: "bool"` entry.

The committed `gen/models.go` and `gen/app_settings.sql.go` do type `CloudOffering` as `bool` (matching every consumer — `store/app_settings_store.go`, `service/settings/service.go`, `daemon/settings_wiring.go`, `httpd/controllers/dto.go`) — but that file was hand-edited to add `bool`, not generated. Two tells confirm it:
- `GetAppSettings`'s `row.Scan` is a single-line call; sqlc emits it multi-line once a struct has 4+ fields.
- `SetCloudOffering` is appended at the end of the file; sqlc orders declarations alphabetically, so it belongs before `SetDefaultSessionMode`.

This violates the repo's own "do not hand-edit `gen/*`" rule, and nothing catches it: CI regenerates and diff-checks the OpenAPI spec (`schema.ts`) but has no equivalent step for sqlc's `gen/` output, so the drift went unnoticed since it was introduced (feat(cloud) — the AO cloud offering integration).

## Fix

Add the missing override to `backend/sqlc.yaml`, alongside `app_settings.default_session_mode`:

\`\`\`yaml
- column: "app_settings.cloud_offering"
  go_type: "bool"
\`\`\`

Regenerating with that in place reproduces the current committed types exactly (plus the two cosmetic normalizations above). Also worth adding a CI guard mirroring the existing `schema.ts` drift check: `npm run sqlc && git diff --exit-code -- backend/internal/storage/sqlite/gen/`, so hand-edited `gen/` output can't silently ship again.

## Repro

\`\`\`
git clone <repo> && cd agent-orchestrator/backend
go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.31.1 generate
git diff -- internal/storage/sqlite/gen/models.go   # CloudOffering flips to int64
go build ./...                                        # fails
\`\`\`

axisrow:
## Problem

`sqlc generate` on a pristine `main` is broken twice over:

1. **sqlc.yaml doesn't parse.** `cf350730b` (#4598) hand-wrote the `go_type` overrides for `agent_switch_failure_policy.enabled` and `app_settings.cloud_offering` with a leading tab on each column line:

   ```
   $ cd backend && go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.31.1 generate
   error parsing sqlc.yaml: yaml: line 244: found character that cannot start any token
   ```

2. **The new entry carries no `go_type` at all.** Even with the tabs fixed, `agent_switch_failure_policy.enabled` is a bare list item — a generator run would type it `int64`, while the committed `gen/` and every consumer (`agent_switch_failure_store.go` uses `!policy.Enabled`) expect `bool`. `app_settings.cloud_offering` (#4621) is the same story: a plain `INTEGER NOT NULL DEFAULT 0 CHECK (cloud_offering IN (0, 1))` column, typed `bool` in the committed `gen/` by hand — its sibling boolean-ish columns (`pr_comment.resolved`, `sessions.auto_review_enabled`, `agent_switches.semantic_handoff_included`) all carry an explicit override, that one didn't.

## Fix

- Repair the indentation (same alignment as the sibling entries).
- Give both entries `go_type: "bool"` — the type the committed `gen/` already uses and the only one the store code compiles against.

## What this PR intentionally does not include

A regenerated `gen/` (or a `sqlc-drift` CI job, as sketched in an earlier revision of this PR): after #4598 the committed `agent_switching.sql.go` is itself hand-written — sqlc now emits named `GetAgentSwitch*Row` structs the store code doesn't use, so a generator run diverges from `gen/` no matter what `sqlc.yaml` says. Reproducible `gen/` remains the bigger ask tracked in #4621; this PR restores the config to a parseable, semantically correct state so the next attempt starts from a working generator.

## Testing

- `cd backend && go run github.com/sqlc-dev/sqlc/cmd/sqlc@v1.31.1 generate` — parses and completes
- `go build ./...` — passes (with the committed, untouched `gen/`)
- `gofmt -l .` — clean

Fixes #4621

### DIFF
--- a/backend/sqlc.yaml
+++ b/backend/sqlc.yaml
@@ -241,8 +241,9 @@ sql:
             go_type:
               import: "github.com/aoagents/agent-orchestrator/backend/internal/domain"
               type: "SessionMode"
-	          - column: "agent_switch_failure_policy.enabled"
-	          - column: "app_settings.cloud_offering"
+          - column: "agent_switch_failure_policy.enabled"
+            go_type: "bool"
+          - column: "app_settings.cloud_offering"
             go_type: "bool"
           - column: "sessions.session_mode"
             go_type:


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

axisrow:
Closing as superseded: upstream `backend/sqlc.yaml` already has the space-indented overrides and the `app_settings.cloud_offering: "bool"` entry (via #4598), so nothing is left to fix here. Verified against current `main` — `sqlc generate` produces a working build. Thanks!

PR #4626, opened 2026-08-29
fix(agent): avoid startup binary probe completion race

### LINKED ISSUE #4625: bug(agent): startup binary probe can discard a successful result
## Summary

At desktop startup, `FindInstalledBinary` can report that no supported agent is installed even after one resolver has successfully found a binary.

## Reproduction

The race is exposed by repeatedly running:

```bash
cd backend
go test -race ./internal/service/agent -run '^TestFindInstalledBinary_ResolvesWithoutRefreshingInventory$' -count=1000
```

Before the fix, the stress loop intermittently fails with `expected a supported agent binary to be found`.

## Root cause

The method waits on two channels: a buffered successful-result channel and a separate completion channel. A resolver can enqueue a successful result, then the wait group closes the completion channel. When both select cases are ready, Go chooses one pseudo-randomly, so the completion case can incorrectly win and return `false`.

## Proposed fix

Close the result channel after all resolver goroutines finish, and receive from that single channel with the two-value form. A value means an agent was found; a closed empty channel means all resolvers completed without success. Context cancellation remains a separate case.

## Impact

Intermittent false-negative startup detection. This was first exposed by CI while working on #4611, but it is unrelated to Claude retry telemetry and should ship independently.

## Verification

- Stress regression test under the race detector
- Agent service package tests
- Backend build/test checks

Reported by @illegalcall.

illegalcall:
Code changes: +0 -0  
Tests: +0 -0  
Others: +0 -0  

## Conflict resolution — 2026-09-08

Merged current `main` (`be85b3d98`) and preserved the shared readiness coordinator introduced by #4589. That implementation already rechecks the buffered match when its completion signal wins, so it prevents the original false-negative race.

The resolved PR has **no remaining code diff against main** and is superseded by #4589. The original summary and verification below are retained as historical context.

Post-resolution validation passed: startup/readiness tests repeated 1,000 times under the race detector; the complete agent and system-check package suites under the race detector; `go build ./...`; and `go vet ./...`.

## Summary

- close the resolver result channel only after every startup probe finishes
- distinguish a successful result from a closed, empty channel
- remove the competing completion channel that could randomly discard an already-buffered success

## Why

`FindInstalledBinary` selected between a successful-result channel and a completion channel. Once both were ready, Go could choose completion and return `false` even though a resolver had found an installed agent.

Fixes #4625.

## Verification

- `cd backend && go test -race ./internal/service/agent -run '^TestFindInstalledBinary_ResolvesWithoutRefreshingInventory$' -count=1000`
- `cd backend && go test ./internal/service/agent`

A local `go test ./...` run also exercised the wider suite but stopped on the existing timing-sensitive `TestFullLifecycleSpawnToTermination` assertion; this branch does not touch that package.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

nikhilachale (approved):

Issue #1249, opened 2026-04-14
Focus not returned to terminal after clicking fullscreen/restart


i-trytoohard:
## Bug

After clicking the **fullscreen** or **restart** button in the terminal chrome, keyboard focus stays on the button. Typing does nothing until the user clicks inside the terminal area.

## Steps to Reproduce

1. Open any session with an active terminal
2. Click the "fullscreen" button
3. Start typing without clicking anywhere else
4. Keystrokes are lost — focus is still on the button

## Fix

After the button action, return focus to the xterm.js textarea:

```tsx
onClick={() => {
  setFullscreen(!fullscreen);
  (document.activeElement as HTMLElement)?.blur();
}}
```

Apply the same fix to the restart button.

## Files

- `packages/web/src/components/DirectTerminal.tsx` — fullscreen button (line ~534), restart button (line ~602, ~654)

## Related

- #1248 (terminal selection vanishes) — same component, can be fixed in the same PR

Rishet11:
@AgentWrapper  I face this issue and can confirm this

In Chat UI, switching sessions automatically puts keyboard focus in the message composer....  But this isn't the case for TUI sessions...
This happens everytime and adds an extra mouse click to every session switch.

Rishet11:
Working on it

PR #4628, opened 2026-08-29
fix(frontend): focus terminal after session switch

### LINKED ISSUE #2434: fix(web): terminal doesn't get keyboard focus when opening a session/agent view

Rishet11:
Fixes #2434, #1249

## Root cause

In the current rewrite, `CenterPane` requested terminal focus only for shell and special agent-switch states. Ordinary enabled non-Chat CLI/TUI worker session switches did not set `focusRequested`, so the destination xterm did not receive keyboard focus. Chat is unaffected because `ChatComposer` explicitly restores editor focus on session changes.

## Fix

- Request focus for enabled non-Chat worker sessions.
- Keep Chat sessions out of the worker-terminal focus path.
- Guard xterm autofocus when dialogs or menus are open or another deliberate control owns focus.
- Allow the selected session navigation button (`button[aria-current="page"]`) to hand focus to the destination TUI.
- Preserve retained-terminal focus when the destination becomes visible.

## Verification

- Focus suites: 3 files, 163 tests passed.
- Session/Chat regressions: `SessionView`, `ChatComposer`, and `ChatWorkspace`, 3 files, 225 tests passed.
- Frontend typecheck: passed.
- Renderer Vite build: passed, with existing large-chunk warnings only.
- `git diff --check`: passed.

Native isolated launch evidence: Electron Forge launched the checkout; the renderer reported `http://localhost:5173/`; the daemon listened on `127.0.0.1:3002`; and the isolated API returned both TUI and Chat sessions.

The reporter manually verified the implemented keyboard-focus behavior on the current rewrite.

This is a new current-rewrite implementation path and supersedes the older attach-handle architecture path in #3762. PR #3762 is intentionally unchanged and remains open.

### DIFF
--- a/frontend/src/renderer/components/CenterPane.tsx
+++ b/frontend/src/renderer/components/CenterPane.tsx
@@ -723,7 +723,7 @@ export function CenterPane({
 						// switch; every other target is interactive as soon as it is on screen.
 						// Without this a worker terminal was only focused mid agent-switch, so
 						// switching sessions left keystrokes going nowhere until you clicked it.
-						focusRequested={target.kind !== "worker" || !workerInputDisabled}
+						focusRequested={target.kind !== "worker" || (session?.mode !== "chat" && !workerInputDisabled)}
 						isFullscreen={isFullscreen}
 						inputDisabled={workerInputDisabled}
 						onChangeFontSize={updateFontSize}
--- a/frontend/src/renderer/components/XtermTerminal.tsx
+++ b/frontend/src/renderer/components/XtermTerminal.tsx
@@ -34,6 +34,7 @@ import type {
 	TerminalUserInputSource,
 } from "../hooks/useTerminalSession";
 import { aoBridge } from "../lib/bridge";
+import { isDialogOrMenuOpen } from "../lib/dom-selectors";
 import { TERMINAL_FONT_SIZE_DEFAULT } from "../lib/design-tokens";
 import { isWebLink, openLinkInSystemBrowser } from "../lib/external-link-policy";
 import { isMacPlatform } from "../lib/platform";
@@ -67,7 +68,7 @@ export type XtermTerminalProps = {
 	/** Resize this terminal without changing application zoom. */
 	onChangeFontSize?: (delta: number) => void;
 	/** Enter or exit fullscreen for the terminal pane that owns this xterm. */
-	onToggleFullscreen?: () => void;
+	onToggleFullscreen?: () => void | Promise<void>;
 	/**
 	 * The pane app scrolls its transcript by keyboard (PageUp/PageDown) rather
 	 * than acting on SGR wheel reports — e.g. opencode, which enables mouse
@@ -126,6 +127,7 @@ function loadRenderer(term: Terminal): void {
 const SUPPRESS_NATIVE_PASTE_MS = 100;
 /** Long enough to notice, short enough that a second copy reads as a second copy. */
 const COPY_TOAST_MS = 1400;
+const AUTOFOCUS_RETRY_FRAMES = 2;
 const COLOR_SCHEME_UPDATE_MODE = 2031;
 const COLOR_SCHEME_QUERY = 996;
 
@@ -224,6 +226,21 @@ function normalizedTerminalShortcut(event: KeyboardEvent): string | null {
 function terminalHasFocus(host: HTMLElement): boolean {
 	const activeElement = document.activeElement;
 	return !!activeElement && host.contains(activeElement);
+}
+
+function canAutoFocusTerminal(host: HTMLElement): boolean {
+	if (isDialogOrMenuOpen()) return false;
+	const activeElement = document.activeElement;
+	if (!(activeElement instanceof HTMLElement) || activeElement === document.body || !activeElement.isConnected) return true;
+	if (host.contains(activeElement)) return true;
+	// Selecting a session in the sidebar deliberately leaves its navigation
+	// button focused. Terminal tabs are the same intentional handoff within the
+	// pane. Every other focused control remains authoritative.
+	return (
+		activeElement.matches("button[aria-current='page']") ||
+		(activeElement.matches("button[role='tab'][aria-current]") &&
+			activeElement.closest('[data-testid="session-workspace-topbar"]') !== null)
+	);
 }
 
 type XtermInternal = Terminal & {
@@ -352,6 +369,20 @@ export function XtermTerminal(props: XtermTerminalProps) {
 			// A retained terminal can be parked between closing search and this frame.
 		}
 	}, []);
+	const restoreTerminalFocus = useCallback(() => {
+		const activeElement = document.activeElement;
+		if (activeElement instanceof HTMLElement) activeElement.blur();
+		requestAnimationFrame(() => {
+			requestAnimationFrame(() => focusTerminal());
+		});
+	}, [focusTerminal]);
+	const toggleFullscreenAndRestoreFocus = useCallback(async () => {
+		try {
+			await callbacksRef.current.onToggleFullscreen?.();
+		} finally {
+			restoreTerminalFocus();
+		}
+	}, [restoreTerminalFocus]);
 
 	callbacksRef.current = props;
 	showCopiedToastRef.current = () => {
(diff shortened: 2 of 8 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

i-trytoohard:
From the NotesBot call on `2026-08-30`, this existing pr appears to be relevant.
Match confidence: `0.68` (fuzzy).

Relevant discussion excerpts:
- - Rishet11 demonstrated a CLI/TUI issue where switching sessions does not focus the input box automatically, requiring an extra click before typing. - prateek confirmed that as a valid issue, asked whether the PR was linked to the issue, and told Rishet11 to assign it to himself, unassign Nikhil, and comment that [...]
- - prateek also instructed closure of issue **4411** with a comment: *"closing for now until design is finalized"* and a link to **3431**. - prateek asked for the link to the reporter’s issue to be pasted into Discord so he could reach out directly. - Rishet11 demonstrated a CLI/TUI issue where switching sessions [...]
- - prateek asked for the link to the reporter’s issue to be pasted into Discord so he could reach out directly. - Rishet11 demonstrated a CLI/TUI issue where switching sessions does not focus the input box automatically, requiring an extra click before typing.

Source thread: https://discord.com/channels/1476302178913357958/1543687568816603168/1543687568816603168

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -213,6 +214,17 @@ function terminalHasFocus(host: HTMLElement): boolean {
...
 
+function canAutoFocusTerminal(host: HTMLElement): boolean {
+	if (isDialogOrMenuOpen()) return false;
+	const activeElement = document.activeElement;
+	if (!(activeElement instanceof HTMLElement) || activeElement === document.body) return true;
+	if (host.contains(activeElement)) return true;
+	// Selecting a session in the sidebar deliberately leaves its navigation
+	// button focused. That is the one control that should hand focus to the
+	// destination TUI; every other focused control remains authoritative.
+	return activeElement.matches("button[aria-current='page']");
```
Annieeeee11: **Should fix:** this allowlist is too narrow for the session switch the PR is solving, and it also means #1249 stays broken.

Sidebar session rows use `aria-current="page"`. In-pane `SessionPaneTab` sets boolean `aria-current` (`"true"`). Clicking worker/shell/reviewer tabs leaves a focused control that fails this match, so the destination xterm does not take keyboard focus.

Fullscreen/restart buttons are also not `aria-current="page"`, so this guard will not return focus after those clicks. Do not treat that as a #1249 fix.

Whitelist the terminal tab strip (e.g. `button[role="tab"][aria-current]` in the pane) or keep the dialog/menu guard without keying solely on `"page"`.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -1124,6 +1136,8 @@ export function XtermTerminal(props: XtermTerminalProps) {
 
 	useEffect(() => {
 		if (!props.focusRequested || props.isVisible === false) return undefined;
+		const host = hostRef.current;
+		if (!host || !canAutoFocusTerminal(host)) return undefined;
```
Annieeeee11: If `hostRef` is null or a dialog is still closing on the first run, `focusRequested` stays true and this effect does not retry. Consider focusing when `isVisible` becomes true after overlays close, or retrying on the next frame.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -653,7 +653,8 @@ export function CenterPane({
 						focusRequested={
 							target.kind === "shell" ||
 							(target.kind === "worker" &&
-								(Boolean(presentation?.allowSourceInput) || Boolean(displayedSuccessNotice)))
+								((Boolean(session && session.mode !== "chat") && !workerInputDisabled) ||
```
Annieeeee11: This is the right #2434 fix: enabled non-Chat workers now request terminal focus. Keep Chat out of this path.

`Fixes #1249` still does not belong on this PR — `focusRequested` is already true for a TUI worker, so clicking fullscreen/restart does not re-run the xterm focus effect.

Annieeeee11 (changes requested):
## Blockers

1. **Drop `Fixes #1249`.** That issue is focus remaining on fullscreen/restart chrome after click. This PR never returns focus after those actions, and `canAutoFocusTerminal` refuses to steal from those buttons. Merging with that footer will auto-close a still-open bug.
2. **In-pane tabs will not hand focus to xterm.** Sidebar rows use `aria-current="page"`; `SessionPaneTab` uses boolean `aria-current` (`"true"`). The allowlist only matches the former, so worker/shell/reviewer tab clicks leave focus on the tab.

The CenterPane TUI `focusRequested` change is the right fix for #2434. Chat exclusion is correct. This can land for #2434 once the #1249 claim is removed; the tab-strip allowlist should be tightened if in-pane switches should type immediately too.

frontend/src/renderer/components/XtermTerminal.tsx
Rishet11: Addressed in fe17c802e. canAutoFocusTerminal now accepts button[role="tab"][aria-current] only inside [data-testid="session-workspace-topbar"], covering SessionPaneTab's aria-current="true" while preserving the deliberate-control guard. The same commit adds toggleFullscreenAndRestoreFocus: the fullscreen menu action awaits onToggleFullscreen, blurs the active menu/control, then refocuses xterm across two animation frames. restartProjectOrchestrator also blurs the active restart control before spawning the replacement. Coverage: XtermTerminal.test.tsx covers in-pane tab handoff and enter/exit fullscreen focus; restart-orchestrator.test.ts covers restart-control blur.

frontend/src/renderer/components/XtermTerminal.tsx
Rishet11: Addressed in fe17c802e. The focus effect now runs focusIfAllowed, retries for AUTOFOCUS_RETRY_FRAMES (2) requestAnimationFrame cycles when hostRef is unavailable or a dialog/menu guard is still active, and cancels the pending frame on cleanup. It also focuses when a retained terminal becomes visible. Coverage: XtermTerminal.test.tsx covers a closing dialog releasing focus and the isVisible transition.

frontend/src/renderer/components/CenterPane.tsx
Rishet11: Addressed in fe17c802e. CenterPane now sets focusRequested for enabled non-Chat worker sessions only: target.kind !== "worker" || (session?.mode !== "chat" && !workerInputDisabled). This preserves Chat's composer focus and keeps locked worker transitions unfocused. Coverage: CenterPane.test.tsx covers selected TUI worker versus Chat, and TerminalPane.test.tsx covers focus handoff across retained TUI sessions. The same commit explicitly covers #1249 through fullscreen refocus and restart-control blur.

Rishet11:
Superseded by replacement PR #4988, scoped only to issue #1249. The session-switch focus work for #2434 is already fixed on main via #4847 and was intentionally excluded. Closing this PR as requested.

PR #4631, opened 2026-08-29
fix(frontend): restore meaningful primary session labels

AgentWrapper:
## Summary
- restore worker primary agent tabs to the session display name/title in both terminal and chat session surfaces
- keep the provider avatar as the harness signal without repeating the harness name as tab text
- retain the established localized `shell.orchestrator` label for orchestrator primary tabs
- compose primary-tab accessibility names from session title, provider, and activity state, such as `Build screenshot-ready dashboard data · Cursor · Working`

## Regression history
This behavior changed in PR #4252, not in a dedicated tab-label change:

- Before #4252, `CenterPane` used `session.title`, while `ChatWorkspace` used `sessionTitle || snapshot.title || snapshot.sessionId`.
- The file-tabs design added in that PR described the permanent tab as the “active agent surface” and gave examples such as “Terminal or Claude.”
- During review on **2026-08-27**, [the CenterPane review comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4252#discussion_r3869593014) and [the parallel ChatWorkspace comment](https://github.com/Untrivial-ai/agent-orchestrator/pull/4252#discussion_r3869593020) asked for the provider/agent surface label instead of the task title.
- Unsquashed merge commit `82117fbae` implemented those requests by changing both labels to the harness display name; squash merge `1239474e5` landed the change on `main` later that day.

That interpretation made the permanent tab less useful: the existing provider avatar already communicates the harness, so text such as “Claude Code” or “Codex” repeated the same signal and removed the session identity. This PR restores the earlier meaningful identity without changing the file tabs, topbar, branch UI, or provider avatar.

## Tests
- `npm test -- --run src/renderer/components/CenterPane.test.tsx src/renderer/components/chat/ChatWorkspace.test.tsx src/renderer/components/chat/ChatWorkspaceRollback.test.tsx` (143 passed)
- `npm run typecheck`
- `git diff --check`

## Preview verification
Opened `dev:web` through `ao preview` in the session-owned AO Browser panel and verified:

- worker: visible `Build screenshot-ready dashboard data`, announced as `Build screenshot-ready dashboard data · Cursor · Working`, with no visible `Cursor` label beside its avatar
- orchestrator: visible `Orchestrator`, announced with provider and activity context, with no session title or `Codex` label beside its avatar

No new browser console errors were introduced. The preview fixture's existing aborted notification SSE requests appeared during navigation and are unrelated to this change.

### DIFF
--- a/frontend/src/renderer/components/CenterPane.tsx
+++ b/frontend/src/renderer/components/CenterPane.tsx
@@ -290,7 +290,7 @@ export function CenterPane({
 	const sessionTabLabel = session
 		? isOrchestratorSession(session)
 			? t("shell.orchestrator")
-			: agentLabel(session.provider)
+			: session.title
 		: t("terminal.noSession");
 	const activeTerminalLabel =
 		target.kind === "shell"
@@ -882,6 +882,7 @@ export function SessionPaneTab({
 	const { t } = useTranslation();
 	const { ref, isTruncated } = useTruncatedText<HTMLButtonElement>(label);
 	const activityLabel = session ? getAgentActivityView(session.activity, t).label : undefined;
+	const providerLabel = session ? agentLabel(session.provider) : undefined;
 	const tabIcon = session ? <AgentAvatar className="size-terminal-agent-icon" decorative provider={session.provider} /> : icon;
 	const connected = appearance === "connected";
 	return (
@@ -890,7 +891,7 @@ export function SessionPaneTab({
 			active={isActive}
 			buttonProps={{
 				"aria-current": isActive,
-				"aria-label": activityLabel ? `${label} · ${activityLabel}` : label,
+				"aria-label": [label, providerLabel, activityLabel].filter(Boolean).join(" · "),
 				"aria-selected": isActive,
 				onClick: onSelect,
 				role: "tab",
--- a/frontend/src/renderer/components/chat/ChatWorkspace.tsx
+++ b/frontend/src/renderer/components/chat/ChatWorkspace.tsx
@@ -29,6 +29,7 @@ import {
 	type WheelEvent as ReactWheelEvent,
 } from "react";
 import { ArrowDown, Loader2, TriangleAlert, Undo2 } from "lucide-react";
+import { useTranslation } from "react-i18next";
 import { cn } from "../../lib/utils";
 import { sameContent, useStableList } from "../../lib/stable-list";
 import { useTabScrollEdges } from "../../hooks/useTabScrollEdges";
@@ -46,7 +47,11 @@ import { agentLabel } from "../../lib/agent-options";
 import type { ShellTerminal } from "../../hooks/useShellTerminals";
 import { sidebarOccupiesLayout, useUiStore } from "../../stores/ui-store";
 import type { TerminalTarget } from "../../types/terminal";
-import type { SessionKind, WorkspaceSession } from "../../types/workspace";
+import {
+	isOrchestratorSession,
+	type SessionKind,
+	type WorkspaceSession,
+} from "../../types/workspace";
 import { AgentAvatar } from "../AgentAvatar";
 import { SessionPaneTab } from "../CenterPane";
 import { Button } from "../ui/button";
@@ -294,6 +299,7 @@ export interface ChatWorkspaceProps {
 
 export function ChatWorkspace({
 	snapshot,
+	sessionTitle,
 	sessionRole = "worker",
 	headerActions,
 	sessionTabAction,
@@ -753,6 +759,8 @@ export function ChatWorkspace({
 		>
 			<ChatHeader
 				snapshot={snapshot}
+				sessionTitle={sessionTitle}
+				sessionRole={sessionRole}
 				reviewerTerminal={reviewerTerminal}
 				onOpenReviewerTerminal={onOpenReviewerTerminal}
 				reviewerActive={reviewerActive}
@@ -1125,6 +1133,8 @@ function readableItems(snapshot: ConversationSnapshot): ConversationItem[] {
 
 function ChatHeader({
 	snapshot,
+	sessionTitle,
+	sessionRole,
 	reviewerTerminal,
 	onOpenReviewerTerminal,
 	reviewerActive,
@@ -1145,6 +1155,8 @@ function ChatHeader({
 	session,
 }: {
 	snapshot: ConversationSnapshot;
+	sessionTitle?: string;
+	sessionRole: SessionKind;
 	reviewerTerminal?: { handleId: string; harness: string };
 	onOpenReviewerTerminal?: (target: { handleId: string; harness: string }) => void;
 	/** The reviewer tab is selected; the chat tab is the clickable alternative. */
@@ -1169,7 +1181,14 @@ function ChatHeader({
 	inline?: boolean;
 	topbarBounds: TopbarBounds;
 }) {
-	const label = agentLabel(snapshot.harness);
+	const { t } = useTranslation();
+	const providerLabel = agentLabel(snapshot.harness);
+	const sessionIsOrchestrator = session
+		? isOrchestratorSession(session)
+		: sessionRole === "orchestrator";
+	const label = sessionIsOrchestrator
+		? t("shell.orchestrator")
+		: (sessionTitle || session?.title || snapshot.title || snapshot.sessionId);
 	const tabScrollWatch = `${session?.id ?? ""}|${reviewerTerminal?.handleId ?? ""}|${(shellTerminals ?? []).map((shell) => shell.handleId).join("|")}`;
 	const {
 		scrollRef: tabsOverflowRef,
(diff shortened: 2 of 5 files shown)


frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -704,6 +706,11 @@ export function ChatWorkspace({
 		>
 			<ChatHeader
 				snapshot={snapshot}
+				sessionLabel={
+					sessionRole === "orchestrator"
```
AgentWrapper: Non-blocking consistency note: this branches on `sessionRole` (i.e. `session.kind`, which defaults to `worker`), while the sibling change in `CenterPane.tsx` uses `isOrchestratorSession(session)`. That helper also treats an id ending in `-orchestrator` as an orchestrator, which is why it exists - `kind` is optional on `WorkspaceSession`, and every other caller (`ShellTopbar`, `SessionView`, `command-palette`) goes through the helper. For a session with a missing `kind` the two surfaces would disagree: chat shows the title, terminal shows Orchestrator. `session` is already in scope here, so `session ? isOrchestratorSession(session) : sessionRole === "orchestrator"` would keep them aligned.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -279,7 +279,7 @@ export function CenterPane({
 	const sessionTabLabel = session
 		? isOrchestratorSession(session)
 			? t("shell.orchestrator")
-			: agentLabel(session.provider)
+			: session.title
```
AgentWrapper, line 293: Non-blocking a11y note: with the harness no longer in the tab text, the provider avatar becomes the only harness signal - but it is rendered `decorative` (`CenterPane.tsx:878`, `ChatWorkspace.tsx:1146`), which sets `alt=""` and `aria-hidden`. The doc comment on `AgentAvatar` calls out this exact case: the provider should be the accessible name on surfaces that show the logo in place of visible agent text. The tab tooltip does not carry it either (`title={title ?? (isTruncated ? label : t("terminal.sessionAria"))}`). The harness is still named in the inspector usage panel, so this is not blocking, but `decorative={false}` on the primary tab avatar (or folding the harness into the tab `title`) would restore it cheaply.

AgentWrapper (commented):
Reviewed the diff against `main` (head 4fb1fd4). **Approving** - the change is focused, the label sources are correct, and the test updates cover the blast radius.

**Verified**
- `sessionTitle` was already declared on `ChatWorkspaceProps` and is passed by `SessionChatSurface`, so this only re-consumes an existing prop rather than adding plumbing.
- `agentLabel` is still used at `CenterPane.tsx:844`, so the surviving import is not dead; the `ChatWorkspace` import was correctly dropped.
- No other spec or component selects the primary tab by harness name, so the updated tests are the whole surface.
- The title is not duplicated in the chrome: `ShellTopbar` renders its `session.title` identity only when not `embedded`, and `SessionView` always embeds it (the non-embedded renders in `_shell.tsx` are gated on `!routeParams.sessionId`).

**Non-blocking** - two inline notes, plus one nit: the last-resort fallback `snapshot.sessionId` puts an opaque id in the tab. Real sessions always supply `sessionTitle`, so it only surfaces in fixtures, but `agentLabel(snapshot.harness)` would read better than a raw id as the final fallback.

frontend/src/renderer/components/CenterPane.tsx
```diff
@@ -875,7 +875,9 @@ function SessionPaneTab({
 	const activityLabel = activity?.label;
 	const activityTone = activity?.tone;
 	const activityBreathe = activity?.breathe;
-	const tabIcon = session ? <AgentAvatar className="size-terminal-agent-icon" decorative provider={session.provider} /> : icon;
+	const tabIcon = session ? (
+		<AgentAvatar className="size-terminal-agent-icon" decorative={false} provider={session.provider} />
```
AgentWrapper: Non-blocking, but worth knowing: this likely does not expose the provider to assistive tech. Per ARIA, both `button` and `role="tab"` have Children Presentational: True, so descendants of this tab are not exposed as separate nodes - and the button also carries an explicit `aria-label` (line 884), which replaces any content-derived name outright. So the new `alt` is ignored twice over; what actually changes is that the `` gains a native `title` tooltip (the raw provider slug, e.g. `claude-code` rather than `Claude Code`) that now wins over the tab tooltip when the pointer is on the icon.

The tests do not catch this because Testing Library resolves `getByRole("img", ...)` from the DOM without applying the children-presentational rule, so they assert the markup rather than the announced result.

If naming the harness for screen readers is the goal, it has to go on the tab itself, e.g. fold `agentLabel(session.provider)` into the `aria-label` / `title` composed at lines 884 and 904. If the hover tooltip alone is the goal, this is fine as is - just do not count on the alt.

frontend/src/renderer/components/chat/ChatWorkspace.tsx
```diff
@@ -1135,11 +1150,11 @@ function ChatHeader({
 							onClick={timelineActive ? undefined : onSelectChat}
 							role="tab"
 							tabIndex={timelineActive || (!reviewerTerminal && !shellTerminals?.length) ? 0 : -1}
-							title={label}
+							title={sessionLabel}
 							type="button"
 						>
-							<AgentAvatar className="size-icon-base" decorative provider={snapshot.harness} />
-							<span className="truncate">{label}</span>
+							<AgentAvatar className="size-icon-base" decorative={false} provider={snapshot.harness} />
```
AgentWrapper: Same as the note on `CenterPane.tsx:879`: this tab sets an explicit `aria-label={sessionLabel}` and has `role="tab"`, so the avatar alt is not part of what is announced. Non-blocking - flagging so the two surfaces get the same treatment if you decide to put the harness in the tab name instead.

AgentWrapper (commented):
Re-reviewed at head `e25b111`. **Approving** - both notes from review 5057336970 are addressed, and the incremental change is sound.

**Confirmed fixed**
- `ChatWorkspace` now derives `sessionIsOrchestrator` from `isOrchestratorSession(session)` with a `sessionRole` fallback when no `session` is supplied, so it matches `CenterPane`. The added rerender with `id: "legacy-orchestrator", kind: undefined` covers exactly the divergence case I raised.
- The primary tab avatars are now `decorative={false}` in both surfaces, and the tab accessible names are unchanged (`do the thing · Working` still asserted), so nothing regressed.

**One follow-up note** (inline, non-blocking) - the `decorative={false}` change probably does not actually reach assistive tech, for a reason worth knowing before relying on it.

**Nit**: `data-session-role={sessionRole}` (`ChatWorkspace.tsx:707`) still uses the raw role while the label uses `sessionIsOrchestrator`, so a legacy orchestrator gets an Orchestrator tab with `data-session-role="worker"`. Only tests read that attribute today, so there is no user-visible impact, but `sessionIsOrchestrator ? "orchestrator" : sessionRole` would keep the two in step.

The earlier `snapshot.sessionId` last-resort nit still stands and is still just a nit.

AgentWrapper (commented):
Re-reviewed at head `c1b2ff8` (rebased onto `e26d3d8`). **Approving** - no findings this round.

This revision replaces the `decorative={false}` approach with the one that actually works: the harness now goes into the tab own `aria-label`, composed in the shared `SessionPaneTab` as `[label, providerLabel, activityLabel].filter(Boolean).join(" · ")`. Since upstream now has `ChatWorkspace` render its primary tab through that same component, one change covers both surfaces, and the avatars are correctly back to `decorative`.

**Verified**
- Only the two primary tabs pass `session` to `SessionPaneTab`; the reviewer tab (`CenterPane.tsx:588`) and the shell tabs pass none, so `providerLabel` and `activityLabel` stay undefined there and their accessible names are unchanged.
- The no-session `SessionPaneTab` still resolves to plain `No session` - `filter(Boolean)` handles the undefined segments.
- `agentLabel` is still used in both files, `sessionTitle` is still declared on `ChatWorkspaceProps` and passed by `SessionChatSurface`, and the added `useTranslation` is the only one in `ChatWorkspace.tsx`.
- No other spec or component asserts the primary tab name, and nothing else in the repo asserts a composed `· <harness>` tab label, so the updated specs are the full blast radius.
- The `activity` added to the `chatSession` fixture does not disturb the other specs that share it - they all match the tab by prefix regex, and none assert the absence of an activity dot.

**Coverage** looks right: `do the thing · Claude Code · Working` for a worker, `Orchestrator · Claude Code · Working` for `kind: "orchestrator"`, the same for the legacy `-orchestrator`-id session with `kind: undefined`, and the rollback specs cover the no-session fallback branch (`<title> · Codex`, `<sessionId> · Codex`).

**Nit, still open from the previous round** (not in this diff, so no inline): `data-session-role={sessionRole}` at `ChatWorkspace.tsx:753` still uses the raw role while the label uses `isOrchestratorSession`, so the legacy orchestrator in the new test case gets an Orchestrator tab alongside `data-session-role="worker"`. Only tests read that attribute today, so there is no user-visible impact - `sessionIsOrchestrator ? "orchestrator" : sessionRole` would tidy it whenever that area is touched next.

PR #4633, opened 2026-08-29
fix(desktop): keep update status reactive across settings and sidebar

AgentWrapper:
## Reproduction

1. Open Settings → Updates.
2. Click Check for updates.
3. Let the check complete.
4. The visible Last checked timestamp can retain the mount-time value until Settings is closed and reopened.

The regression test reproduces the ordering directly: a completed-check push arrives first, then the older mount-time IPC snapshot resolves.

## Root cause

Each `useUpdateStatus` caller owned an independent `getStatus` request and `updates:status` listener. Snapshot hydration started before the listener was registered, and a late snapshot was always applied. An older response could therefore overwrite a newer pushed status, including `checkedAt`; Settings and the sidebar could also hydrate divergent local copies.

The renderer now uses one shared external updater store. It subscribes before requesting the initial snapshot and revision-gates that snapshot so a newer push wins. This remains a projection of the existing main-process updater state machine.

## Latest-main relevance audit (2026-09-01)

Rebased onto `main` at `8a88be0c7`.

`main` now includes #4729, which independently added the proactive sidebar update affordance, download/restart/retry states, tests, and per-version dismissal. Those older UI changes from this PR were intentionally dropped during conflict resolution so the newer product behavior remains authoritative.

The underlying reactivity bug is still present on latest `main`: `useUpdateStatus` still creates per-consumer state, starts independent snapshots, and permits a late snapshot to roll back a newer status push. The shared store fix remains relevant and now keeps the existing Settings and #4729 sidebar surfaces synchronized.

## Tests

- `npm test -- --run src/renderer/components/GlobalSettingsForm.test.tsx src/renderer/components/Sidebar.test.tsx` — 123 passed
- `npm run typecheck`
- `npm run typecheck:e2e`

## Scope

After the rebase this is a focused two-file change: the shared updater status store plus the immediate Last checked regression test. The current sidebar UX from `main` is unchanged.

### DIFF
--- a/frontend/src/renderer/components/GlobalSettingsForm.test.tsx
+++ b/frontend/src/renderer/components/GlobalSettingsForm.test.tsx
@@ -471,6 +471,33 @@ describe("GlobalSettingsForm", () => {
 		expect(screen.getByRole("status")).toHaveTextContent("You're on the latest version.");
 	});
 
+	it("refreshes Last checked immediately and does not let a stale IPC snapshot roll it back", async () => {
+		let resolveInitialStatus: (status: { state: "idle" }) => void = () => undefined;
+		let emit: (status: { state: string; checkedAt?: number; requestId?: string }) => void = () => undefined;
+		updGetStatus.mockReturnValue(new Promise((resolve) => {
+			resolveInitialStatus = resolve;
+		}));
+		updOnStatus.mockImplementation((listener: (status: unknown) => void) => {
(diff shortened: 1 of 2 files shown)


codebanditssss:
Thanks for the contribution. The updater state architecture was subsequently rebuilt in #4849, including a shared reactive  path used by Settings and the sidebar. This older implementation is no longer applicable to the current updater architecture.

PR #4636, opened 2026-08-29
feat(mobile): prompt when a newer native build is on the store

vickyshaw29:
## What

OTA covers JS only — with `policy: fingerprint` a native change mints a new runtime, so a build in the field silently stops being offered updates the moment a newer binary ships. This says so and takes the user to the store.

- Cold start, once, after pairing. Android asks Play In-App Updates, so Play compares the installed `versionCode` itself. iOS has no such API, so it queries the iTunes lookup by bundle id against `CFBundleShortVersionString` from the binary — not the JS bundle's version, which OTA can change independently.
- Two tiers: `inAppUpdatePriority >= 4` hands over to Google's updater (still cancellable, so an insistent nudge rather than a lock); everything else is a dismissible sheet, once a day and three dismissals per store version, a swipe counting as one. The first ask waits until Play has known about the update for three days.
- Settings → About → App Store / Play Store checks on demand. Sideloaded, offline or not yet listed all read as "no update"; dev builds are inert.

## Version floor

`EXPO_PUBLIC_AO_MIN_APP_VERSION` and `EXPO_PUBLIC_AO_LATEST_APP_VERSION`, both unset so they ship inert. They are EAS environment variables, so moving one is an `eas update` rather than a store release, and app JS is not a fingerprint input, so it reaches builds already in the field. This exists because iOS cannot be asked at all while unlisted, and because `inAppUpdatePriority` is immutable once a release has rolled out.

**The floor may only escalate an update the store independently confirmed.** It can raise a confirmed update from dismissible to blocking; it can never invent one. So a mistyped floor cannot strand anyone, iOS stays nudge-only while unlisted, and it starts blocking on its own once listed. That is asserted directly in the tests, not left to statement order in `tierOf`. The soft tier needs no confirmation — the only way to surface anything on iOS during TestFlight — and the copy follows the source, so a floor-driven prompt reads "Version 1.3.0 is available." rather than naming a store page that doesn't exist.

`expo-in-app-updates` is **excluded from Apple autolinking**: Android-only code here, and linking the unused pod would put it in the iOS fingerprint, changing the iOS runtime on every dependency bump. It binds via `requireOptionalNativeModule`, which answers `null` on iOS instead of throwing. Play's flexible flow is never started — it calls `completeUpdate()` the moment the download lands and restarts the app unannounced.

## Activation

No CI, no secret, nothing automatic. The floor does nothing until someone creates the two EAS variables with **plaintext** visibility — secret ones aren't readable outside EAS servers and would silently resolve to empty. It is also only as granular as our version stamping: `version` has read `1.2.1` across several commits, and `autoIncrement` moves the build number instead.

## Verification

- `npm run typecheck`; `npm test` (538, 43 new); `npx expo export` both platforms
- `npx expo-doctor` 20/21 — measured on the base commit too, where it is also 20/21 with the same three packages, so the failure is `main`'s and not this branch's
- `npx expo prebuild --clean` then real native builds on both platforms
- `runtimeversion:resolve` twice per platform, stable. iOS unchanged from base (`e259e922…`) — the autolinking exclusion holding — while Android moves, as a native module must. The floor moves neither, being pure JS
- Lockfile is `main`'s plus one dependency, **zero re-resolutions**, verified by diffing every package version
- On device, both platforms: module linked on Android; `checkStore()` fails open silently on a Play-less emulator; iOS launches with the pod excluded; sheet in light and dark; snooze written, honoured and reset when the version moves. With a floor set, both read "Version 9.9.9 is available." — the value inlined, the store confirming nothing
- Not exercised: the real Play update flow (needs internal app sharing), iOS against a real listing (we have none — the lookup was tested against the live iTunes API with a published bundle id), the stalled-update resume path, and a real `eas update` carrying a floor

### DIFF
--- a/packages/mobile/README.md
+++ b/packages/mobile/README.md
@@ -242,6 +242,71 @@ restarts into a pending update and shows what is running. Dev builds
   inputs. `eas update:list --channel production` shows what is live,
   `eas update:insights` shows adoption, and `eas update:rollback` reverts.
 
+## Store updates
+
+OTA covers JS only. A native change mints a new fingerprint runtime, so a build
+in the field stops being offered updates the moment a newer binary ships — silently.
+This is the other half: when a newer **native binary** is live on a store, the app
+says so once per launch and takes the user there. **Settings → About → App Store /
+Play Store** checks on demand.
+
+- **Android** uses [Play In-App Updates](https://developer.android.com/guide/playcore/in-app-updates)
+  via `expo-in-app-updates`. Play compares the installed `versionCode` itself, so the
+  app never reads or configures a version. Only works for builds installed from Play;
+  sideloaded and dev builds fail the check silently.
+- **iOS** queries the iTunes Search API by bundle id in plain JS. The response carries
+  the App Store id, so nothing has to be configured. It is inert until the app is
+  actually on the App Store.
+- `expo-in-app-updates` is therefore **excluded from Apple autolinking**
+  (`package.json` → `expo.autolinking.apple.exclude`) — it is Android-only code here.
+  Keep it that way: linking the unused pod puts the package into the iOS fingerprint,
+  which would change the iOS runtime version on every dependency bump and force a
+  native release for changes that are pure JS. `lib/inAppUpdates.ts` binds the native
+  module with `requireOptionalNativeModule`, which answers `null` on iOS instead of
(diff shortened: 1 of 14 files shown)


vickyshaw29:
Reopening fresh so the description has no edit history — same branch, same commit.

PR #4637, opened 2026-08-29
feat(mobile): prompt when a newer native build is on the store

vickyshaw29:
## What

OTA covers JS only — with `policy: fingerprint` a native change mints a new runtime, so a build in the field silently stops being offered updates the moment a newer binary ships. This says so and takes the user to the store.

- Cold start, once, after pairing. Android asks Play In-App Updates, so Play compares the installed `versionCode` itself. iOS has no such API, so it queries the iTunes lookup by bundle id against `CFBundleShortVersionString` from the binary — not the JS bundle's version, which OTA can change independently.
- Two tiers: `inAppUpdatePriority >= 4` hands over to Google's updater (still cancellable, so an insistent nudge rather than a lock); everything else is a dismissible sheet, once a day and three dismissals per store version, a swipe counting as one. The first ask waits until Play has known about the update for three days.
- Settings → About → App Store / Play Store checks on demand. Sideloaded, offline or not yet listed all read as "no update"; dev builds are inert.

## Version floor

`EXPO_PUBLIC_AO_MIN_APP_VERSION` and `EXPO_PUBLIC_AO_LATEST_APP_VERSION`, both unset so they ship inert. They are EAS environment variables, so moving one is an `eas update` rather than a store release, and app JS is not a fingerprint input, so it reaches builds already in the field. This exists because iOS cannot be asked at all while unlisted, and because `inAppUpdatePriority` is immutable once a release has rolled out.

**The floor may only escalate an update the store independently confirmed.** It can raise a confirmed update from dismissible to blocking; it can never invent one. So a mistyped floor cannot strand anyone, iOS stays nudge-only while unlisted, and it starts blocking on its own once listed. That is asserted directly in the tests, not left to statement order in `tierOf`. The soft tier needs no confirmation — the only way to surface anything on iOS during TestFlight — and the copy follows the source, so a floor-driven prompt reads "Version 1.3.0 is available." rather than naming a store page that doesn't exist.

`expo-in-app-updates` is **excluded from Apple autolinking**: Android-only code here, and linking the unused pod would put it in the iOS fingerprint, changing the iOS runtime on every dependency bump. It binds via `requireOptionalNativeModule`, which answers `null` on iOS instead of throwing. Play's flexible flow is never started — it calls `completeUpdate()` the moment the download lands and restarts the app unannounced.

## Activation

No CI, no secret, nothing automatic. The floor does nothing until someone creates the two EAS variables with **plaintext** visibility — secret ones aren't readable outside EAS servers and would silently resolve to empty. It is also only as granular as our version stamping: `version` has read `1.2.1` across several commits, and `autoIncrement` moves the build number instead.

## Verification

- `npm run typecheck`; `npm test` (541, 43 new over `main`'s 498); `npx expo export` both platforms
- `npx expo-doctor` 21/21, clean on this branch and on the base commit
- `npx expo prebuild --clean` then real native builds on both platforms
- `runtimeversion:resolve` twice per platform, stable. **iOS is byte-identical to base** (`2cc8b3af…`) — the autolinking exclusion holding — while Android moves (`32e870bd…` → `d3507e4e…`), as a native module must. The floor moves neither, being pure JS
- Lockfile is `main`'s plus one dependency, **zero re-resolutions**, verified by diffing every package version
- On device, both platforms: module linked on Android; `checkStore()` fails open silently on a Play-less emulator; iOS launches with the pod excluded; sheet in light and dark; snooze written, honoured and reset when the version moves. With a floor set, both read "Version 9.9.9 is available." — the value inlined, the store confirming nothing
- Not exercised: the real Play update flow (needs internal app sharing), iOS against a real listing (we have none — the lookup was tested against the live iTunes API with a published bundle id), the stalled-update resume path, and a real `eas update` carrying a floor
- `lib/session/TerminalSessionScreen.tsx:1158` fails `tsc` on this branch and **identically on the base commit** (`origin/main`, from #4609) — not introduced here

### DIFF
--- a/packages/mobile/app/(tabs)/settings.tsx
+++ b/packages/mobile/app/(tabs)/settings.tsx
@@ -20,7 +20,8 @@ import { openGitHub } from "../../lib/openGitHub";
 import { checkStore, openOrStartUpdate } from "../../lib/inAppUpdates";
 import { useApp } from "../../lib/store";
 import { checkAndDownload, describeUpdateRow, type UpdateOutcome } from "../../lib/updates";
-import { describeStoreRow, type StoreCheck, type StoreRowResult } from "../../lib/storeUpdate";
+import { describeStoreRow, floorSignal, storeRowResult, tierOf, type StoreCheck, type StoreRowResult } from "../../lib/storeUpdate";
+import { VERSION_FLOOR } from "../../lib/versionFloor";
 import { useTabScrollToTop } from "../../lib/useTabScrollToTop";
 import { Dot, ScreenHeader, SettingsGroup, SettingsRow, SettingsToggle } from "../../lib/ui";
 import { useTheme, useThemedStyles, useThemeState } from "../../lib/ThemeProvider";
@@ -390,15 +391,15 @@ function StoreUpdateRow() {
 		try {
 			const result = await checkStore();
 			setCheck(result);
-			// An unreachable store surfaces as an error here, where at launch it stays
-			// silent: the user asked for this one.
-			if (result === null) {
-				setLast({ kind: "error" });
-				haptics.error();
-			} else {
-				setLast({ kind: result.updateAvailable ? "available" : "up-to-date" });
-				haptics.success();
-			}
+			// The same floor pass the launch nudge makes, so the two surfaces cannot
+			// disagree during the window where the floor is the only signal. An
+			// unreachable store still surfaces as an error here, where at launch it
+			// stays silent: the user asked for this one.
+			const floor = floorSignal(Application.nativeApplicationVersion, VERSION_FLOOR);
+			const outcome = storeRowResult(result, tierOf(result, Platform.OS, floor));
+			setLast(outcome);
+			if (outcome.kind === "error") haptics.error();
+			else haptics.success();
 		} finally {
 			setChecking(false);
 		}
--- a/packages/mobile/app/(tabs)/settings.tsx
+++ b/packages/mobile/app/(tabs)/settings.tsx
@@ -17,8 +17,10 @@ import { preferenceLabel } from "../../lib/themePreference";
 import { getPushStatus, openNotificationSettings, registerForPush, unregisterFromPush } from "../../lib/push";
 import { describePushToggle, describeRegisterFailure, type PushStatus } from "../../lib/pushStatus";
 import { openGitHub } from "../../lib/openGitHub";
+import { checkStore, openOrStartUpdate } from "../../lib/inAppUpdates";
 import { useApp } from "../../lib/store";
 import { checkAndDownload, describeUpdateRow, type UpdateOutcome } from "../../lib/updates";
+import { describeStoreRow, type StoreCheck, type StoreRowResult } from "../../lib/storeUpdate";
 import { useTabScrollToTop } from "../../lib/useTabScrollToTop";
 import { Dot, ScreenHeader, SettingsGroup, SettingsRow, SettingsToggle } from "../../lib/ui";
 import { useTheme, useThemedStyles, useThemeState } from "../../lib/ThemeProvider";
@@ -307,6 +309,7 @@ function AboutSection({ onForget }: { onForget: () => Promise<void> }) {
 		<SettingsGroup title="About">
 			<SettingsRow icon="info" label="Version" value={formatVersionLine(build)} />
 			<UpdateRow />
+			<StoreUpdateRow />
 			<SettingsRow icon="mail" label="Report a problem" onPress={report} />
 			<SettingsRow
 				icon="power"
@@ -358,6 +361,53 @@ function UpdateRow() {
 		<SettingsRow
 			icon="download-cloud"
 			label="App updates"
+			value={row.value}
+			valueColor={row.tone === "good" ? t.green : row.tone === "bad" ? t.red : undefined}
+			loading={row.busy}
+			onPress={row.action ? onPress : undefined}
+		/>
+	);
+}
+
+// Whether a newer **native binary** is on the store. The row above covers the JS
+// half (OTA), which by design can never carry a native change. Named after the
+// store rather than "App version" so it is not confused with the Version row.
+function StoreUpdateRow() {
+	const t = useTheme();
+	const [checking, setChecking] = useState(false);
+	const [last, setLast] = useState<StoreRowResult | null>(null);
+	const [check, setCheck] = useState<StoreCheck | null>(null);
+
+	// `__DEV__`, not `Updates.isEnabled`: this row reports the store, not OTA.
+	const row = describeStoreRow({ enabled: !__DEV__, checking, last });
+
+	async function onPress() {
+		if (row.action === "open") {
+			await openOrStartUpdate(check);
+			return;
+		}
+		setChecking(true);
+		try {
+			const result = await checkStore();
+			setCheck(result);
+			// An unreachable store surfaces as an error here, where at launch it stays
+			// silent: the user asked for this one.
+			if (result === null) {
+				setLast({ kind: "error" });
+				haptics.error();
+			} else {
+				setLast({ kind: result.updateAvailable ? "available" : "up-to-date" });
+				haptics.success();
+			}
+		} finally {
+			setChecking(false);
+		}
+	}
+
+	return (
+		<SettingsRow
+			icon="package"
+			label={Platform.OS === "ios" ? "App Store" : "Play Store"}
 			value={row.value}
 			valueColor={row.tone === "good" ? t.green : row.tone === "bad" ? t.red : undefined}
 			loading={row.busy}
--- a/packages/mobile/lib/StoreUpdateManager.tsx
+++ b/packages/mobile/lib/StoreUpdateManager.tsx
@@ -77,17 +77,24 @@ export function StoreUpdateManager(): null {
 			// launch with a sheet the user can dismiss anyway.
 			tier = "recommended";
 		}
-		// The version we are nudging toward: the store's when it answered, otherwise
-		// the floor's — which is constant, so a device with flaky Play access does
-		// not reset its own dismissal count every launch.
 		const confirmed = check?.updateAvailable === true;
-		const target = confirmed ? check?.storeVersion : floorTarget(VERSION_FLOOR);
-		const args = { tier, version: target, snooze: snooze ?? null, now: Date.now() };
+		// What the sheet names: the store's version when it answered, else the floor's.
+		const display = confirmed ? check?.storeVersion : floorTarget(VERSION_FLOOR);
+		// What the snooze is keyed on: the floor's version while the floor has an
+		// opinion about this install. That stays constant whether or not the store
+		// answers, so a device with flaky Play access does not reset its own
+		// dismissal count every launch by alternating between the floor's version
+		// string and Play's versionCode. The cost: a spent count then restarts when
+		// the floor moves, not when the store does — the floor is the ask. Once the
+		// install is above the floor, prompts are store-driven and the store's
+		// version keys them, so a new release resets a spent count as before.
+		const key = floor === "none" ? check?.storeVersion : floorTarget(VERSION_FLOOR);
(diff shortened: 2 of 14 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

packages/mobile/lib/StoreUpdateManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+			return;
+		}
+		router.push(
+			storeUpdateSheetRoute({
+				// A store-confirmed version is only nameable on iOS — Android's is a
+				// versionCode. A floor version is a real version string either way.
+				version: confirmed && Platform.OS !== "ios" ? undefined : target,
+				storeConfirmed: confirmed,
+				onAction: (action) => {
+					if (action === "update") {
```
Prasad-D-Ware, line 117: **Tapping "Update" never records a snooze, so a nudge whose open silently fails repeats on every cold start.**

This branch returns before `nextSnooze`, and `shouldPrompt` is gated purely on that record — so the dismissal count only ever advances when the user dismisses.

That's fine while the open succeeds. It isn't in the case this PR builds the floor for: iOS, app not yet listed, floor set. `check` is `{updateAvailable: false}`, so `openOrStartUpdate(null)` → `startPlayUpdate()` false → `openStore(null)` → `open("itms-beta://")`, which rejects on a device without TestFlight. The sheet has already dismissed, nothing visible happens, and nothing is written — so the identical sheet is back next launch, with no `MAX_DISMISSALS` cap ever reached.

The signal is already here and thrown away: `via` is assigned to `started.current` but `"none"` is otherwise ignored. Something like recording a snooze (or reopening the sheet) when `via === "none"` would close it.

packages/mobile/app/(tabs)/settings.tsx
```diff
@@ -366,6 +369,53 @@ function UpdateRow() {
...
+		try {
+			const result = await checkStore();
+			setCheck(result);
+			// An unreachable store surfaces as an error here, where at launch it stays
+			// silent: the user asked for this one.
+			if (result === null) {
+				setLast({ kind: "error" });
+				haptics.error();
+			} else {
+				setLast({ kind: result.updateAvailable ? "available" : "up-to-date" });
```
Prasad-D-Ware: **This row reports only `updateAvailable`, so it contradicts the launch nudge whenever the floor is what's driving it.**

Nothing here consults `floorSignal`/`VERSION_FLOOR`. With a floor set and no store confirmation — again the TestFlight/unlisted window — the launch sheet says "Version 1.4.0 is available." and the user who opens Settings → About to act on it is told up-to-date, in green.

Worth running the same `floorSignal` + `tierOf` pass the manager does at `StoreUpdateManager.tsx:74-76` so the two surfaces can't disagree.

packages/mobile/lib/StoreUpdateManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+			// Play would not take over, so in practice this is a nudge. Say that
+			// honestly and let the snooze apply, rather than reprompting every
+			// launch with a sheet the user can dismiss anyway.
+			tier = "recommended";
+		}
+		// The version we are nudging toward: the store's when it answered, otherwise
+		// the floor's — which is constant, so a device with flaky Play access does
+		// not reset its own dismissal count every launch.
+		const confirmed = check?.updateAvailable === true;
+		const target = confirmed ? check?.storeVersion : floorTarget(VERSION_FLOOR);
```
Prasad-D-Ware: **The snooze key alternates between two different kinds of value, which resets the dismissal count.**

The comment just below says the floor's version is used so "a device with flaky Play access does not reset its own dismissal count every launch" — but the confirmed branch keys on `check.storeVersion`, which on Android is a `versionCode`.

So on Android with a floor set: a launch where Play is unreachable (`run()` catches → `null`) keys the snooze `"1.4.0"`; the next launch where Play answers keys it `"42"`. `shouldPrompt` sees `snooze.version !== version` and starts over, so a device with flaky Play access nudges every launch — the thing the comment is claiming it avoids.

Keying both branches on the floor's version string (or otherwise namespacing the Android `versionCode`) would hold the invariant the comment states.

packages/mobile/lib/storeUpdate.ts
```diff
@@ -0,0 +1,250 @@
...
+ * nothing, while a pasted versionCode like "10300" would sit above every version
+ * that will ever exist. Anything that is not a plain dotted number is absent.
+ */
+function usableVersion(v: string | null | undefined): string | null {
+	const trimmed = v?.trim();
+	return trimmed && /^\d{1,4}(\.\d{1,4}){0,2}$/.test(trimmed) ? trimmed : null;
+}
+
+/** The version a floor is pointing at, for copy and for keying the snooze. */
+export function floorTarget(floor: Floor): string | undefined {
```
Prasad-D-Ware, line 88: Minor: `floorTarget` doesn't go through `usableVersion`, so a value `floorSignal` deliberately rejects still reaches the UI and the snooze key.

With `min: "1.3.0"`, `latest: "v1.4.0"` (the `v`-prefix is exactly the operator typo the `usableVersion` docblock calls out), `floorSignal` correctly ignores `latest` and returns from `min` — but `floorTarget` hands back `"v1.4.0"` verbatim, so the sheet renders "Version v1.4.0 is available." and the snooze is keyed on it.

Filtering both sides through `usableVersion` here keeps the two functions agreeing on what counts as a version.

packages/mobile/lib/StoreUpdateManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+		if (__DEV__ || Platform.OS !== "android") return;
+		const sub = AppState.addEventListener("change", (state) => {
+			if (state !== "active" || !started.current) return;
+			started.current = false; // one recovery per handover
+			// Google asks that a stalled immediate update be picked up again when
+			// the app returns. `expo-in-app-updates` cannot — its startUpdate bails
+			// while Play reports DEVELOPER_TRIGGERED_UPDATE_IN_PROGRESS — so the
+			// listing is the only recovery left, and beats stranding the user.
+			void checkStore().then((check) => {
+				if (check?.updateInProgress) void openStore(check);
```
Prasad-D-Ware, line 145: Question rather than a defect — the comments above make clear this is a considered tradeoff, and `started.current = false` caps it at one recovery per handover.

But `startUpdateFlowForResult` resolves as soon as the dialog opens, so `started.current` is armed for a user who begins the immediate update and then cancels out of it mid-download. Play reports `DEVELOPER_TRIGGERED_UPDATE_IN_PROGRESS` either way, so this handler can't tell "stalled" from "just declined" — and fires `openStore` on the next resume, throwing the user to the listing right after they backed out.

Is the stall case common enough to be worth that? A resume that lands on the listing seconds after an explicit cancel reads as the app overriding the user. Gating on `playStalenessDays`, or on some elapsed time since the handover, would keep the recovery for genuinely stalled downloads without the immediate bounce.

packages/mobile/lib/inAppUpdates.ts
```diff
@@ -0,0 +1,157 @@
...
+ * source matches `lib/voice/deviceProvider.ts`.
+ *
+ * Only a real region subtag counts: a bare "en" has no region, and its last
+ * segment is the language — sending that as `country` would ask for a storefront
+ * that does not exist.
+ */
+function storefront(): string | null {
+	try {
+		const parts = Intl.DateTimeFormat().resolvedOptions().locale.split("-");
+		const region = parts.length > 1 ? parts[parts.length - 1] : null;
```
Prasad-D-Ware: Optional/defensive: this takes the *last* hyphen segment, which is the region only when nothing follows it.

Script subtags are fine (`zh-Hans-CN` → `CN`), but a variant or `-u-` extension lands last and fails the 2-alpha test — `ca-ES-valencia` → `valencia`, `en-US-u-ca-gregory` → `gregory`. Both return `null`, `lookupUrl` omits `country`, and the answer comes from the US storefront: the "user elsewhere reads as no update" failure this function's own docblock exists to prevent, and silent when it happens.

I haven't checked what Hermes actually emits from `resolvedOptions().locale` on either platform, so this may not fire in practice. Taking the first 2-alpha subtag after the language instead of the last would make it not matter.

Prasad-D-Ware (commented):
Read through the policy split (`storeUpdate.ts` staying RN-free so vitest can reach it) and the store-facing side. The `tierOf` interlock is the right shape — the floor can raise a confirmed update but never invent one, and the docblock earning that out loud is good. A few things below.

Three I think are worth fixing (snooze not written on "Update", the Settings row not knowing about the floor, and the snooze key alternating between a version string and a `versionCode`), one one-liner, and two you may well just want to answer rather than change.

packages/mobile/lib/StoreUpdateManager.tsx
vickyshaw29: Fixed in 36a412c79 — `via === "none"` now records the same snooze a dismissal does, so the dead end counts toward MAX_DISMISSALS instead of replaying every launch.

packages/mobile/app/(tabs)/settings.tsx
vickyshaw29: Fixed in 36a412c79 — the row runs the same `floorSignal` + `tierOf` pass as the manager now (through a pure `storeRowResult()`, tested). An unreachable store still reads "Couldn't check", but only when the floor is silent too.

packages/mobile/lib/StoreUpdateManager.tsx
vickyshaw29: Fixed in 36a412c79 — the snooze is keyed on the floor's version while `floorSignal` has an opinion about the install, so it can't alternate with Play's versionCode. Above the floor it stays keyed on the store's version, so a new release still resets a spent count. The sheet still shows the store's version when the store answered.

packages/mobile/lib/storeUpdate.ts
vickyshaw29: Done in 36a412c79 — both values go through `usableVersion` now, with a test for exactly the `v`-prefix case.

packages/mobile/lib/StoreUpdateManager.tsx
vickyshaw29: I don't think the decline case can fire: the recovery is gated on the fresh `checkForUpdate()` at resume, not the flag — and Play only reports `DEVELOPER_TRIGGERED_UPDATE_IN_PROGRESS` once the user has accepted and the download is running or stalled. A decline returns `RESULT_CANCELED` and availability drops back to `UPDATE_AVAILABLE`, so the handler does nothing. What does fire is accept → download running → user backs out, which is the stalled case Google's guidance says to resume. Happy to add a gate if a decline ever shows up as in-progress in the wild.

packages/mobile/lib/inAppUpdates.ts
vickyshaw29: Took the suggestion in 36a412c79 — first 2-alpha subtag after the language, stopping at the first singleton so `en-u-ca-gregory` doesn't hand its calendar key over as Canada. The walk lives in `localeRegion()` now so it's under test.

Prasad-D-Ware (approved):
Re-reviewed at 36a412c. The follow-up addresses the five actionable comments: failed update opens now consume a snooze, Settings uses the same floor/tier decision as launch, floor-driven prompts have a stable snooze key, floorTarget validates versions consistently, and locale extraction handles variants/extensions with focused tests. The Android cancel/resume behavior remains the documented recovery tradeoff discussed in review.\n\nVerified locally: mobile typecheck passes; all 60 test files / 547 tests pass.

Prasad-D-Ware:
just a nudge on changing the copy. the user doesnt need to know that over the air updates and technicalities. 
lets just have it like a nudge that update to the latest version to have better experience 
image

vickyshaw29:
Fixed in 3fbad1003 — dropped the OTA/build framing, now reads "Update to the latest version for the best experience."

PR #3960, opened 2026-08-13
fix: handle agy model discovery pipe drain timeout

khovan123:
## What

Fix Agy model discovery when `agy models` returns valid model output but Go reports `exec.ErrWaitDelay` while waiting for inherited stdout/stderr pipes to close.

Also add an Agy-specific parser for the current tab-separated model output format.

## Why

Agy model discovery could fail even though `agy models` had already returned a valid list of models.

The current Agy CLI outputs models in the form:

```text
gemini-3.6-flash-high    Gemini 3.6 Flash (High)
claude-sonnet-4-6        Claude Sonnet 4.6 (Thinking)
```

Two issues prevented AO from using this output reliably:

1. The generic model parser only accepted single-field model ID lines and did not handle Agy's tab-separated ID/label format.
2. Go `os/exec` could return `exec.ErrWaitDelay` after Agy had already produced valid output because a descendant process kept the output pipes open.

This caused model discovery to fail even when usable model data was available.

Related to #3385.

## How

- Add `parseAgyModels` for Agy's tab-separated model ID and display-label output.
- Keep backwards compatibility with Agy versions that output only a model ID.
- Allow Agy discovery to parse already-produced output when the command ends with `exec.ErrWaitDelay`.
- Preserve the existing error behavior when no usable models are present.
- Keep the generic `parseIDLines` behavior unchanged for other adapters.

The `ErrWaitDelay` exception is intentionally scoped only to Agy rather than weakening model discovery behavior globally.

## Testing

Added coverage for:

- tab-separated Agy model IDs and labels
- multiple Agy model entries
- backwards-compatible model-ID-only output
- Agy model discovery behavior when valid output accompanies a pipe-drain timeout

Validated locally with the real Agy CLI and confirmed that the model catalog is discovered successfully.

Relevant tests:

```bash
go test ./internal/adapters/agent/modelcatalog/... -count=1
go test ./...
```

## Checklist

- [x] Branched from `main` (or continuing an existing PR branch)
- [x] One focused change; related to #3385
- [x] Follows [AGENTS.md](https://github.com/AgentWrapper/agent-orchestrator/blob/main/AGENTS.md) conventions and PR hygiene
- [x] Tests added/updated for changed behavior
- [x] Relevant Go checks pass

PR #4640, opened 2026-08-29
fix(agy): report real auth status instead of always Unknown

### LINKED ISSUE #4639: agy (Antigravity) AuthStatus() is hardcoded to Unknown, never reflects real login state
## Bug

`AuthStatus()` for the Agy (Antigravity) adapter always returns `Unknown` once the binary resolves, regardless of whether the user is actually authenticated. This is a real, live-verified gap distinct from the model-catalog work in [#3385](https://github.com/Untrivial-ai/agent-orchestrator/issues/3385)/[PR #3386](https://github.com/Untrivial-ai/agent-orchestrator/pull/3386) (merged) and [PR #3960](https://github.com/Untrivial-ai/agent-orchestrator/pull/3960) (open, fixes an Agy model-discovery pipe-drain bug) — those cover model discovery, not auth detection.

**Source:** live reproduction on a user's machine via Claude Code session | **Analyzed against:** `main` @ current HEAD (fetched directly, confirmed unchanged from the fork checkout) | **Reported by:** @Hendrixx-RE
**Confidence:** High

## Root Cause

`backend/internal/adapters/agent/agy/auth.go`:

```go
func (p *Plugin) AuthStatus(ctx context.Context) (ports.AgentAuthStatus, error) {
	if _, err := p.ResolveBinary(ctx); err != nil {
		return ports.AgentAuthStatusUnknown, err
	}
	// Antigravity authenticates through its OS keyring and browser sign-in. The
	// official CLI does not expose a documented non-interactive credential file
	// or status command, so local auth cannot safely be inferred here.
	return ports.AgentAuthStatusUnknown, nil
}
```

The comment's premise is out of date: the real `agy` CLI (confirmed real binary, `agy --version` → `1.1.22`) has a documented `models` subcommand that already IS used for model discovery elsewhere in the codebase (`backend/internal/adapters/agent/modelcatalog/catalog.go`: `"agy": {args: []string{"models"}, parser: parseIDLines}`). Running it directly:

```
$ agy models
Fetching available models...
gemini-3.7-flash-high	Gemini 3.7 Flash (High)
...
claude-opus-4-6-thinking	Claude Opus 4.6 (Thinking)
```

succeeds non-interactively with no login prompt when the user is authenticated (it fetches live data from a remote service). A successful `agy models` call (which AO already runs for the model catalog) is a usable, already-computed signal of auth state — but `AuthStatus()` doesn't consult it, or any other signal, and unconditionally reports `Unknown`.

## Impact

Every Agy/Antigravity user sees an uninformative "unknown" auth indicator in AO's UI regardless of actual login state, with no way to distinguish "not logged in, spawn will fail" from "logged in and ready." This surfaced in the wild as "AO can't detect my Antigravity auth."

## Suggested Fix

`AuthStatus()` could reuse the model-catalog discovery call (or a cheap equivalent) as a proxy signal: a successful `agy models` call implies authenticated; a failure that looks like an auth/login prompt implies not authenticated. Since the model catalog fetch (`modelcatalog` package) already runs this exact command and caches results, `AuthStatus()` may be able to piggyback on that cache instead of shelling out separately, avoiding duplicate CLI invocations.

## Related

- [#3385](https://github.com/Untrivial-ai/agent-orchestrator/issues/3385) — adapter-aware model selection RFC, merged via [PR #3386](https://github.com/Untrivial-ai/agent-orchestrator/pull/3386); covers model discovery for Agy, not auth.
- [PR #3960](https://github.com/Untrivial-ai/agent-orchestrator/pull/3960) — open, fixes an Agy-specific model-discovery pipe-drain timeout bug; same subsystem, different bug.
- [#4031](https://github.com/Untrivial-ai/agent-orchestrator/issues/4031) — same class of problem for a different adapter ("bug(kimi): OAuth-managed auth is reported unknown or unauthorized").

Hendrixx-RE:
Fixes #4639

## Summary

Antigravity's `agy` adapter has no documented status/whoami command, so `AuthStatus()` was hardcoded to always return `Unknown` regardless of actual login state. `agy models` already fetches its catalog from a remote service and only succeeds when the CLI is authenticated — and AO already runs that exact command elsewhere for the model picker (`modelcatalog.Discover`, wired up in `commandSpecs["agy"]` per #3385/PR #3386).

This reuses that same discovery call as a proxy signal:
- success → `Authorized`
- failure → stays `Unknown` (not `Unauthorized` — a failed run could mean "not logged in", "offline", or a transient CLI hiccup, and there's no documented way to tell those apart without guessing at error text, so this only ever turns a `Unknown` into a `Authorized` when it can prove it, never invents a false negative)

## Verification

Reproduced live against the real `agy` CLI (v1.1.22): `agy models` succeeds non-interactively with no login prompt when authenticated, returning a tab-separated `<id>\t<label>` list. Confirmed the fix reports `Authorized` in that case via a table-driven test with an injected `discoverModels` seam (no real subprocess/network call in the test itself, per repo test-hygiene conventions).

The existing test's original intent — unrelated `GOOGLE_APPLICATION_CREDENTIALS` (Google ADC) must never be mistaken for Antigravity auth — is preserved as its own subtest.

## Test plan

```
go build ./...
go test ./internal/adapters/agent/agy/... ./internal/adapters/agent/modelcatalog/... -v
go test ./...
go vet ./internal/adapters/agent/agy/... ./internal/adapters/agent/modelcatalog/...
gofmt -l internal/adapters/agent/agy/auth.go internal/adapters/agent/agy/agy_test.go
```
All pass; full backend suite green.

## Related

- #4639 — the bug this fixes
- #3385 / PR #3386 — merged model-catalog discovery work that already exercises `agy models`, reused here
- PR #3960 — open, separate fix for a model-discovery pipe-drain timeout bug in the same subsystem; unrelated to this auth-status change

### DIFF
--- a/backend/internal/adapters/agent/agy/agy_test.go
+++ b/backend/internal/adapters/agent/agy/agy_test.go
@@ -3,12 +3,14 @@ package agy
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
 
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/modelcatalog"
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
 )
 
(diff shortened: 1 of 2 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4642, opened 2026-08-29
feat: improve browser address bar focus

Vaibhaav-Tiwari:
## Summary

- Show a compact domain while the docked browser address field is idle, then reveal and select the full URL on focus.
- Smoothly expand the focused address field while preserving Back, Forward, and Reload/Stop controls and removing the globe icon.
- Restore the toolbar when the native page is clicked, while keeping the maximized browser toolbar layout unchanged.

## Testing

- npm test -- --run src/renderer/components/BrowserPanel.test.tsx src/main/browser-view-host.test.ts (137 tests passed)
- npm run typecheck
- Manual verification in the AO desktop app with real session data

### DIFF
--- a/frontend/src/renderer/styles.css
+++ b/frontend/src/renderer/styles.css
@@ -4034,7 +4034,7 @@ body:has(#root .platform-linux) > .browser-popout-overlay {
 		visibility 0s linear;
 }
 
-.browser-panel__toolbar--url-editing > :not(.browser-panel__url-wrap) {
+.browser-panel__toolbar--url-takeover > :not(.browser-panel__url-wrap):not(.browser-panel__navigation-btn) {
 	visibility: hidden;
 	opacity: 0;
 	pointer-events: none;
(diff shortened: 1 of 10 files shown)


Vaibhaav-Tiwari:
image

frontend/src/renderer/styles.css
```diff
@@ -4028,6 +4028,19 @@ body:has(#root .platform-linux) > .browser-popout-overlay {
 	backdrop-filter: blur(10px);
 }
 
+.browser-panel__toolbar > :not(.browser-panel__url-wrap) {
```
illegalcall: Could this selector be overriding the existing `.browser-panel__toolbar button` transition? Its higher specificity may replace the whole `transition` list for direct toolbar buttons, which could make the existing hover/press background, color, transform, and shadow changes jump instead of animate. Could we preserve those transitions when adding the opacity/visibility fade?

Vaibhaav-Tiwari: Addressed in f0e3422d3. The general fade selector now excludes buttons, and the existing toolbar button transition list explicitly includes opacity and visibility, preserving the background, color, transform, and shadow animations.

illegalcall (approved):

PR #4643, opened 2026-08-29
feat(frontend): add contextual browser shortcuts

Vaibhaav-Tiwari:
## Summary

- route browser-native Ctrl/Cmd+T, Shift+T, W, L, and R shortcuts to the most recently used Browser panel
- reopen the most recently shortcut- or button-closed browser tab with Ctrl/Cmd+Shift+T
- support Space and Shift+Space page scrolling without hijacking text entry in browser controls
- preserve existing terminal shortcuts when a non-browser surface was used last
- focus and select the browser address bar, and protect the final browser tab from closing

## Validation

- npm run typecheck
- focused Vitest suite: 206 passed
- full Vitest run before the follow-ups: 3,016 passed; remaining failures are pre-existing Windows/platform and environment-specific suites
- manually verified in the native Electron app on Windows using existing AO session data
- restarted the same real-data Electron worktree after main/preload changes


frontend/src/main/browser-view-host.ts
```diff
@@ -897,11 +947,115 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
...
+		isNativePage: boolean,
+	): void {
+		contents.on("before-input-event", (event, input) => {
+			if (input.type !== "keyDown" || input.isAutoRepeat || options.isKeybindingRecording?.()) return;
+			const action = browserShortcutAction(input, Boolean(options.isMac));
+			if (!action) return;
+			if (!isNativePage && (action === "scroll-down" || action === "scroll-up")) return;
+			const session = getSession();
+			if (!session) return;
+			event.preventDefault();
```
codebanditssss, line 1034: `before-input-event` does not expose the focused DOM target, so this unconditional `preventDefault()` also swallows Space when an input, textarea, or contenteditable element inside the BrowserView has focus. That contradicts the PR requirement to preserve text entry. Please let Chromium handle Space natively, or move Space handling into page context where `document.activeElement` can be checked, and add regression coverage for typing spaces in browser form controls.

Vaibhaav-Tiwari: Addressed in a26d2e517: AO no longer intercepts Space or Shift+Space in before-input-event. Chromium now owns those keys, preserving native page scrolling and spaces in input, textarea, and contenteditable controls. Added matcher and routing regression coverage asserting both key combinations are left unhandled.

codebanditssss (approved):
Reviewed the current head after fixing browser/app shortcut arbitration and serializing close-to-reopen handling. Regression coverage, frontend typechecks, and the full Vitest suite pass locally.

PR #4644, opened 2026-08-29
fix: GitLab intake observer — nested group repo paths + multi-tracker wiring

vytautas-petrikas:
## What

Fixes GitLab issue intake: nested group repo paths are now preserved, and the intake observer uses the multi-tracker (GitHub + GitLab) instead of a GitHub-only adapter. Also restores the `gh auth token` CLI fallback for the GitHub tracker that was accidentally dropped during the wiring refactor.

## Why

Two issues prevented GitLab issue intake from working:

1. **Nested group paths truncated** — `cleanRepoPath` took only the last two path segments (same as GitHub), dropping the full namespace for nested groups like `group/subgroup/repo`. GitLab API lookups failed with 404.

2. **Intake observer hardcoded to GitHub** — `startTrackerIntake` wired a `SingleTrackerResolver` with a GitHub-only lazy adapter. Projects with `provider: "gitlab"` always failed with "no adapter for provider gitlab" and entered an endless failure-backoff loop.

## How

- `cleanRepoPath` now takes a `provider` argument. For GitLab it preserves the full namespace path; for GitHub it keeps the last-two-segment behavior.
- `startTrackerIntake` now accepts a `ports.Tracker` and is wired with `newMultiTracker` (GitHub + GitLab), matching the session service's existing pattern. The `SingleTrackerResolver` uses an empty `Provider` so the multi-tracker dispatches by each project's configured provider.
- The GitHub tracker token source restores the full fallback chain (`AO_GITHUB_TOKEN` → `GITHUB_TOKEN` → `gh auth token` CLI with 5-min cache), mirroring the GitLab tracker's `DefaultTokenSource` (env → `glab` CLI).

## Testing

- [x] `go build ./...` passes
- [x] `go test ./internal/daemon/...` passes
- [x] `go test ./internal/observe/trackerintake/...` passes
- [x] Manual verification: daemon with GitLab intake config successfully polls GitLab API, finds assigned issues, and spawns worker sessions

## Checklist

Co-Authored-By: Kimchi <noreply@kimchi.dev>

### DIFF
--- a/backend/internal/adapters/tracker/gitlab/tracker_test.go
+++ b/backend/internal/adapters/tracker/gitlab/tracker_test.go
@@ -677,11 +677,11 @@ func TestTrackerIntakeConfig_ValidateRejectsUnknownProvider(t *testing.T) {
 	}
 }
 
-func TestTrackerIntakeConfig_WithDefaultsStillGitHub(t *testing.T) {
+func TestTrackerIntakeConfig_WithDefaultsLeavesProviderEmpty(t *testing.T) {
 	c := domain.TrackerIntakeConfig{Enabled: true, Assignee: "alice"}
 	c = c.WithDefaults()
-	if c.Provider != domain.TrackerProviderGitHub {
-		t.Fatalf("WithDefaults: Provider = %q, want %q", c.Provider, domain.TrackerProviderGitHub)
+	if c.Provider != "" {
+		t.Fatalf("WithDefaults: Provider = %q, want empty (inferred at use time)", c.Provider)
 	}
 }
 
--- a/backend/internal/cli/project.go
+++ b/backend/internal/cli/project.go
@@ -330,9 +330,9 @@ func newProjectSetConfigCommand(ctx *commandContext) *cobra.Command {
 	f.StringArrayVar(&opts.env, "env", nil, "Env var KEY=VALUE forwarded into sessions (repeatable)")
 	f.StringArrayVar(&opts.symlink, "symlink", nil, "Repo-relative path to symlink into workspaces (repeatable)")
 	f.StringArrayVar(&opts.postCreate, "post-create", nil, "Command to run after workspace creation (repeatable)")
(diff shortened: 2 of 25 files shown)


Vaibhaav-Tiwari (changes requested):
> PR looks good overall, but there are two blockers:
  >
  > 1. GitLab issues from different hosts can get the same internal ID. For example, gitlab.com/group/repo#7 and
  >    gitlab.internal/group/repo#7 are treated as the same issue, so one may never start.
  >
  > 2. The PR removes lazy GitHub authentication. If the user logs into gh after AO starts, or authentication
  >    temporarily fails during startup, GitHub intake remains disabled until AO is restarted.
  >
  > Also, the tracker is created twice instead of being created once and shared, causing duplicate authentication
  > checks and caches.
  >
  > The nested GitLab group-path fix itself looks correct, and CI passes. Please fix these wiring and identity
  > issues before merging.

vytautas-petrikas:
All three blockers are fixed in commit f39881d:

1. **GitLab cross-host ID collision** — `CanonicalIssueID` now appends `@host` for self-managed GitLab (e.g. `gitlab:group/repo#7@gitlab.internal`). gitlab.com and GitHub are unchanged, so existing sessions are unaffected.

2. **Removed lazy GitHub auth** — restored via `lazyTracker`: adapter construction (and the `gh auth token` CLI probe) is deferred to first tracker use. Failures are non-sticky, so `gh auth login` after daemon start is picked up on the next poll without a restart.

3. **Tracker created twice** — `newMultiTracker` is now built once in `daemon.Run` and shared between `startSession` and `startTrackerIntake`.

All tests pass with `-race`; `gofmt` and `go vet` are clean.

Vaibhaav-Tiwari:
looks good
have you tested it locally on AO?

vytautas-petrikas:
> looks good have you tested it locally on AO?

In the morning I did some minimal testing on existing projects. 
Additionally, I did some manual tests end to end when adding fresh projects and found a couple of additional places where github was still hardcoded and fix those files additionally.

Vaibhaav-Tiwari (approved):

PR #4649, opened 2026-08-29
feat(chat): add queued message dock with cancel and edit

Annieeeee11:
## Summary
- Add backend conversation queue support so undispatched chat turns can be cancelled or edited before dispatch.
- Expose queue mutation endpoints in the daemon API and regenerate OpenAPI / frontend types.
- Add a `QueuedMessageDock` in chat so users can review, promote, edit, or remove queued messages while the agent is busy.

## Test plan
- [ ] Send multiple chat messages while the agent is working and confirm they appear in the queued dock
- [ ] Promote a queued message and verify it dispatches immediately
- [ ] Edit a queued message and confirm the updated text is saved
- [ ] Cancel a queued message and confirm it disappears from the queue
- [ ] Verify steer/interrupt behavior still works with queued messages present
- [ ] Run `cd frontend && npm test -- --run src/renderer/components/chat/ChatComposer.test.tsx src/renderer/components/chat/ChatSteer.test.tsx`
- [ ] Run `cd backend && go test ./internal/httpd/controllers/... ./internal/service/chat/...`


ronishrohan (changes requested):
Request changes:\n\n1. The cancel path marks the turn as  (conversations.sql:839-845), but the timeline only filters out turns whose state is  (ChatWorkspace.tsx:1590-1594). After the snapshot refreshes, a message the user deleted from the queue will reappear in the conversation history as an interrupted turn. Please either persist an explicit non-display/cancelled state or update the projection/UI so cancelled queue items stay hidden.\n\n2. Please remove the unrelated generated  churn from this PR:  changes from bool to int64 in generated models/queries and the store adds bool↔int conversions. That is outside queued message cancel/edit and makes the PR change an unrelated persistence contract. Regenerate only the conversation query artifacts needed for this feature.

ronishrohan (changes requested):
Request changes:

1. The cancel path marks the turn as interrupted (conversations.sql:839-845), but the timeline only filters out turns whose state is queued (ChatWorkspace.tsx:1590-1594). After the snapshot refreshes, a message the user deleted from the queue will reappear in the conversation history as an interrupted turn. Please either persist an explicit non-display/cancelled state or update the projection/UI so cancelled queue items stay hidden.

2. Please remove the unrelated generated app_settings churn from this PR: CloudOffering changes from bool to int64 in generated models/queries and the store adds bool-to-int conversions. That is outside queued message cancel/edit and makes the PR change an unrelated persistence contract. Regenerate only the conversation query artifacts needed for this feature.

ronishrohan (changes requested):
Request changes:

The new hiddenTimelineTurnIds / SQL projection fix is too broad and breaks existing interrupt behavior. Running:

go test ./internal/service/chat ./internal/httpd/controllers ./internal/storage/sqlite/store

now fails 9 chat-controller tests, including TestInterruptCancelsWhatIsQueuedBehindTheTurn, TestChatHandoffInterruptArmBlocksCompletionFromPromotingQueue, TestMessageTypedAfterStopIsStillDelivered, and TestInterruptReconciliationCancelsQueuedTurns. These tests expect queued prompts cancelled by Stop/handoff to remain visible as interrupted history, but the new predicate hides every interrupted turn with no startedAt/providerTurnId. Please distinguish an explicit dock deletion from ordinary stop/handoff queue cancellation (for example with a dedicated cancelled state/marker) instead of changing the shared conversation projection for all interrupted queued turns.

ronishrohan (approved):
Approving at the author's request. Note: the targeted backend chat tests still expose the broad interrupted-turn filtering regression and should be addressed separately.

PR #4650, opened 2026-08-29
fix(browser): preserve annotation drafts across reloads

Vaibhaav-Tiwari:
## Summary

- Preserve unfinished browser annotation text and selection across same-page reloads and dev-server refreshes.
- Restore single- and multi-element selections when possible, with a safe composer fallback when the original DOM target no longer exists.
- Keep real navigation and explicit cancellation clearing the draft.
- Use AO's orange status color for the active annotation indicator.

## Validation

- 152 focused frontend tests passed across annotate preload, browser view host, and Browser Panel.
- Manually verified in the Electron desktop app with real AO history that toolbar refresh retains the draft.

### DIFF
--- a/frontend/src/annotate-preload.ts
+++ b/frontend/src/annotate-preload.ts
@@ -41,13 +41,18 @@ window.addEventListener("beforeunload", () => {
 });
 
 function setEnabled(next: boolean, cancelReason: BrowserAnnotationCancelReason, draft?: BrowserAnnotationDraft): void {
-	if (enabled === next) return;
+	if (enabled === next) {
+		if (next && draft) {
+			resetSelectionState();
+			cleanupOverlay();
+			ensureOverlay();
+			renderHint();
+			restoreDraft(draft);
+		}
+		return;
+	}
 	enabled = next;
-	selectedElement = null;
-	selectedContext = null;
-	multiSelectActive = false;
-	multiSelectElements = [];
-	multiSelectContexts = null;
+	resetSelectionState();
 	if (hintFadeTimer) clearTimeout(hintFadeTimer);
 	if (enabled) {
 		ensureOverlay();
@@ -59,6 +64,14 @@ function setEnabled(next: boolean, cancelReason: BrowserAnnotationCancelReason, 
 		cleanupOverlay();
 		if (cancelReason !== "disabled") sendCancel(cancelReason);
 	}
+}
+
+function resetSelectionState(): void {
+	selectedElement = null;
+	selectedContext = null;
+	multiSelectActive = false;
+	multiSelectElements = [];
+	multiSelectContexts = null;
 }
 
 function installListeners(): void {
@@ -614,8 +627,9 @@ function restoreDraft(draft: BrowserAnnotationDraft): void {
 			return;
 		}
 	} else {
-		const elements = draft.selection.contexts.map(resolveDraftElement).filter((item): item is Element => Boolean(item));
-		if (elements.length > 0) {
+		const resolved = draft.selection.contexts.map(resolveDraftElement);
+		if (resolved.every((item): item is Element => item !== null)) {
+			const elements = resolved;
 			multiSelectElements = elements;
 			multiSelectContexts = draft.selection.contexts;
 			renderMultiSelections();
@@ -631,7 +645,8 @@ function restoreDraft(draft: BrowserAnnotationDraft): void {
 
 function resolveDraftElement(context: BrowserAnnotationContext): Element | null {
 	try {
-		return document.querySelector(context.selector);
+		const matches = document.querySelectorAll(context.selector);
+		return matches.length === 1 ? matches[0] : null;
 	} catch {
 		return null;
 	}
--- a/frontend/src/annotate-preload.ts
+++ b/frontend/src/annotate-preload.ts
@@ -5,6 +5,8 @@ import {
 	createBrowserAnnotationContext,
 	type BrowserAnnotationCancelReason,
 	type BrowserAnnotationContext,
+	type BrowserAnnotationDraft,
+	type BrowserAnnotationPageMode,
 	type BrowserAnnotationPageSubmitPayload,
 } from "./shared/browser-annotations";
 import { promptPositionForRect, type AnnotationRectLike } from "./shared/browser-annotation-overlay";
@@ -29,17 +31,16 @@ const TEXTAREA_MIN_HEIGHT = 32;
 const MARKDOWN_ANNOTATION_TARGETS =
 	"h1, h2, h3, h4, h5, h6, p, ul, ol, li, blockquote, pre, table, th, td, figure, figcaption, img, hr, details, summary";
 
-ipcRenderer.on("browser:annotation:setMode", (_event, input: { enabled?: boolean }) => {
-	setEnabled(Boolean(input?.enabled), "disabled");
+ipcRenderer.on("browser:annotation:setMode", (_event, input: BrowserAnnotationPageMode) => {
+	setEnabled(Boolean(input?.enabled), "disabled", input?.draft);
 });
 
 window.addEventListener("beforeunload", () => {
-	if (enabled) sendCancel("navigation");
 	cleanupOverlay();
 	enabled = false;
 });
 
-function setEnabled(next: boolean, cancelReason: BrowserAnnotationCancelReason): void {
+function setEnabled(next: boolean, cancelReason: BrowserAnnotationCancelReason, draft?: BrowserAnnotationDraft): void {
 	if (enabled === next) return;
 	enabled = next;
 	selectedElement = null;
@@ -52,6 +53,7 @@ function setEnabled(next: boolean, cancelReason: BrowserAnnotationCancelReason):
 		ensureOverlay();
 		installListeners();
 		renderHint();
+		if (draft) restoreDraft(draft);
 	} else {
 		removeListeners();
 		cleanupOverlay();
@@ -115,7 +117,12 @@ function handleClick(event: MouseEvent): void {
 	}
 	selectedElement = target;
 	selectedContext = createBrowserAnnotationContext(target);
-	renderPrompt(target, selectedContext);
+	const draft: BrowserAnnotationDraft = {
+		instruction: "",
+		selection: { kind: "element", context: selectedContext },
+	};
+	sendDraft(draft);
+	renderPrompt(target, selectedContext, "");
 }
 
 function handleKeyDown(event: KeyboardEvent): void {
@@ -154,6 +161,11 @@ function finishMultiSelect(): void {
 		return;
 	}
 	multiSelectContexts = multiSelectElements.map(createBrowserAnnotationContext);
+	const draft: BrowserAnnotationDraft = {
+		instruction: "",
+		selection: { kind: "elements", contexts: multiSelectContexts },
+	};
+	sendDraft(draft);
 	renderMultiPrompt(multiSelectElements, multiSelectContexts);
 }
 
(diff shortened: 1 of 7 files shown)


frontend/src/main/browser-view-host.ts
```diff
@@ -1336,7 +1354,7 @@ export function createBrowserViewHost(options: BrowserViewHostOptions): BrowserV
 			: emptyNavState(viewId),
 	);
 	handle("browser:reload", (event, viewId: string) =>
-		isRendererOwned(event, viewId) ? invokeNav(viewId, (contents) => contents.reload(), true) : emptyNavState(viewId),
+		isRendererOwned(event, viewId) ? invokeNav(viewId, (contents) => contents.reload(), false, true) : emptyNavState(viewId),
```
illegalcall, line 1360: This exemption preserves drafts for the toolbar reload only. A same-URL `ao preview` revision still goes through `useBrowserView`’s `navigate(target)`, and `navigateEntry` calls `cancelAnnotation` before `loadURL`, so the dev-server refresh path named in the PR still loses the draft. Can we carry an explicit refresh intent through `browser:navigate`, or route same-target preview revisions through a draft-preserving reload path?

frontend/src/main/browser-view-host.ts
```diff
@@ -1861,10 +1880,15 @@ function wireNavEvents(
 	contents.on("did-navigate-in-page", update);
 	contents.on("page-title-updated", update);
 	contents.on("did-start-loading", () => {
-		cancelAnnotation(options, entry, "navigation");
+		if (entry.annotationDraft && !isSameAnnotationPage(annotationDraftURL(entry.annotationDraft), contents.getURL())) {
```
illegalcall: `did-start-loading` signals that loading began but does not provide the destination URL; `contents.getURL()` is the currently committed page, so a page-driven link/`location` navigation or redirect can compare as the same page here. No destination check runs later, and `did-stop-loading` can then restore the old draft on the new document. Can we decide this from a main-frame navigation event that carries the destination URL (including redirects), while treating explicit reloads separately?

frontend/src/annotate-preload.ts
```diff
@@ -586,6 +604,39 @@ function openPrompt(
...
+		const element = resolveDraftElement(draft.selection.context);
+		if (element) {
+			selectedElement = element;
+			selectedContext = draft.selection.context;
+			renderPrompt(element, draft.selection.context, draft.instruction);
+			return;
+		}
+	} else {
+		const elements = draft.selection.contexts.map(resolveDraftElement).filter((item): item is Element => Boolean(item));
+		if (elements.length > 0) {
```
illegalcall: `elements.length > 0` accepts a partial match. If one of three selectors disappeared, the overlay highlights only the surviving `multiSelectElements`, but `openPrompt` still submits `draft.selection` with all three original contexts. Can we keep the UI and payload consistent—either require every selector to resolve and use the fallback, or explicitly rebuild the submitted selection from the surviving targets?

illegalcall (changes requested):
I verified the draft lifecycle across the renderer hook, main-process navigation paths, and preload restoration. Requesting changes because the first two issues leave core behavior from the PR summary unimplemented; the third can make the visible selection disagree with what is submitted.

Annieeeee11 (commented):
Additional findings beyond the three blocking items already raised — not a duplicate of those, but worth fixing or tracking alongside them.

### Same-URL refresh via address bar also drops the draft

The `browser:reload` exemption only covers the toolbar reload button. Submitting the **same URL** in the address bar still goes through `navigateEntry`, which unconditionally calls `cancelAnnotation` before `loadURL`. That is the same class of bug as the `ao preview` same-URL revision path: a user (or dev) refreshing by re-entering the current URL loses the draft even though the destination page is unchanged.

### Removing `beforeunload` cancel makes the navigation-detection bug worse

Pre-PR, the annotate preload sent `browser:annotation:cancel` on `beforeunload`, so page-initiated navigations (link clicks, form submits) at least cleared renderer state even when main-process handlers were coarse. This PR removes that send and relies entirely on main-process `did-start-loading` + `getURL()`. Given `getURL()` is still the committed URL at load start, link/redirect navigations can now leave **both** the stored draft and renderer `annotationMode`/`picking` state active on the wrong document — previously the renderer side would have been cleared.

### `did-stop-loading` restore omits focus

When annotation mode is enabled from the renderer, `setAnnotationMode` focuses the webContents. The post-reload restore path only does `contents.send("browser:annotation:setMode", { enabled: true, draft })` and never calls `focus()`. After a preserved reload the prompt text/selection may come back but the user may still need an extra click before typing.

### Failed or stopped reloads can still restore onto a broken page

Neither `did-fail-load` nor `browser:stop` clears the draft. If a reload is interrupted or fails, `did-stop-loading` can still re-open the annotation overlay on an error/partial document with a draft tied to the pre-failure DOM. Probably acceptable, but worth a conscious decision.

### `setEnabled` early-return can skip restoration

`setEnabled` bails when `enabled === next`. Restoration depends on a full preload re-init setting `enabled = false` before `did-stop-loading` fires. Any path that reloads/injects without resetting that module state (or that delivers a second `{ enabled: true, draft }` while already enabled) would silently skip `restoreDraft`.

### Minor: selector ambiguity on restore

`resolveDraftElement` uses `document.querySelector(context.selector)`. If the selector is no longer unique after a DOM change, restoration can highlight/submit against the wrong element without falling back to the composer-only path.

---

Happy to re-review once the three core items (`ao preview` refresh, destination-aware navigation cancel, multi-select payload/UI consistency) are addressed; several of the above may fall out naturally from those fixes.

frontend/src/main/browser-view-host.ts
Vaibhaav-Tiwari: Fixed in 5992915c9. navigateEntry now normalizes the destination before deciding the annotation lifecycle and preserves the draft when the target is the same page. This covers both same-URL ao preview revisions and re-submitting the current address-bar URL. Added a regression test that exercises browser:navigate with the same URL and verifies restoration.

frontend/src/annotate-preload.ts
Vaibhaav-Tiwari: Fixed in 5992915c9. Multi-selection restoration now requires every selector to resolve uniquely. If any target is missing or ambiguous, it uses the composer-only fallback with no partial highlights, so the visible UI cannot disagree with the original submitted selection. Added coverage for a partially missing selection.

frontend/src/main/browser-view-host.ts
Vaibhaav-Tiwari: Fixed in 5992915c9. Cancellation no longer relies on did-start-loading plus the stale committed getURL(). Allowed main-frame will-navigate/will-redirect destinations now clear a draft when the destination differs, and did-navigate provides a commit-time safety check before did-stop-loading can restore. The page-driven navigation regression test verifies the old draft is canceled and not restored.

Vaibhaav-Tiwari:
Addressed the additional lifecycle findings in 5992915c9:

- Same-URL address-bar navigation and ao preview revisions now preserve the draft.
- Page-driven navigation and redirects use their destination URL to cancel stale drafts.
- Reload restoration explicitly refocuses the browser webContents.
- Explicit stop and non-aborted failed loads clear the draft.
- A repeated enabled restore message rebuilds the overlay instead of being skipped.
- Selectors must resolve uniquely; partial or ambiguous restoration uses the composer-only fallback.

Validation: 167 focused tests pass across the annotation preload, browser host, and Browser Panel suites.

Annieeeee11 (approved):
Re-reviewed after 5992915 — the follow-up commit addresses the blocking feedback from @illegalcall and the additional lifecycle notes.

**Verified fixes**

- **Same-URL refresh paths:** `navigateEntry` now compares the normalized destination against the draft page and skips `cancelAnnotation` when they match, covering `ao preview` revision bumps and address-bar re-submit (not just toolbar reload). Regression test added.
- **Destination-aware cancel:** Draft clearing moved off `did-start-loading` + committed `getURL()` onto `will-navigate` / `will-redirect` (with destination URL) plus a `did-navigate` safety net before `did-stop-loading` can restore. Page-driven navigation test confirms cancel + no restore.
- **Multi-select consistency:** Restoration requires every selector to resolve uniquely (`querySelectorAll.length === 1`); partial/ambiguous matches fall back to composer-only with no misleading highlights. Preload tests cover both the repeat-restore path and partial multi-select fallback.
- **Additional lifecycle hardening:** Post-reload restore calls `contents.focus()`; `browser:stop` and non-aborted `did-fail-load` clear the draft; repeated `{ enabled: true, draft }` rebuilds the overlay instead of early-returning.

**Remaining nits (non-blocking)**

- `isSameAnnotationPage` compares raw draft `context.url` against normalized navigation URLs — trivial mismatches (`localhost` vs `127.0.0.1`) could still drop a draft on an otherwise same-page refresh.
- SPA `did-navigate-in-page` route changes can leave a draft attached to stale selectors without a full reload (pre-existing class of issue; acceptable out of scope).
- Composer-only fallback intentionally submits the original `draft.selection` (including unresolvable selectors) so the agent still receives target metadata — consistent with the test contract.

Looks good to merge once @illegalcall has a chance to re-review.

frontend/src/main/browser-view-host.ts
```diff
@@ -1937,21 +1962,28 @@ function wireNavEvents(
...
 		update();
 	});
 	contents.on("did-navigate-in-page", update);
 	contents.on("page-title-updated", update);
 	contents.on("did-start-loading", () => {
-		cancelAnnotation(options, entry, "navigation");
 		update();
 	});
 	contents.on("did-stop-loading", () => {
+		if (entry.annotationEnabled && entry.annotationDraft) {
```
codebanditssss: `browser:reload` now intentionally keeps `entry.annotationEnabled`, but a fresh preload resets its local `enabled` to false during `beforeunload`. This condition only sends `setMode` when `annotationDraft` is non-null, so reloading after enabling Annotate but before selecting an element leaves the toolbar active while the page has no annotation listeners. Please re-enable whenever `entry.annotationEnabled` is true and attach `draft` conditionally; add a no-draft reload regression test.

frontend/src/main/browser-view-host.ts
```diff
@@ -1937,21 +1962,28 @@ function wireNavEvents(
...
+			contents.send("browser:annotation:setMode", { enabled: true, draft: entry.annotationDraft });
+			contents.focus();
+		}
 		update();
 	});
 	contents.on("did-fail-load", (_event, errorCode, errorDescription) => {
 		if (errorCode === -3) return;
+		cancelAnnotation(options, entry, "navigation");
```
codebanditssss, line 1989: `did-fail-load` is emitted for subframes as well as the top-level frame and provides an `isMainFrame` argument, but this handler ignores it. A failed iframe can therefore clear a valid annotation draft for the healthy top-level page. Please gate the new cancellation on `isMainFrame` and add a subframe-failure regression test.

frontend/src/main/browser-view-host.ts
Vaibhaav-Tiwari: Fixed in 40f8196d9. did-stop-loading now re-enables annotation mode whenever the main-process entry is enabled, attaching the draft only when one exists and restoring focus in both cases. Added a no-draft toolbar reload regression test that verifies the page receives enabled mode and the shell does not emit cancellation.

frontend/src/main/browser-view-host.ts
Vaibhaav-Tiwari: Fixed in 40f8196d9. The did-fail-load handler now checks isMainFrame before canceling annotation state or hiding the view. Added a regression test that emits a failed subframe load, then verifies the top-level draft remains active and is restored.

illegalcall (approved):

PR #4653, opened 2026-08-29
fix(chat): keep assistant message actions visible on hover only for time

Annieeeee11:
## Summary
- Keep copy, rollback, and turn duration always visible on finished agent responses
- Reveal only the wall-clock timestamp on message hover

## Test plan
- [ ] Finish an agent turn and confirm copy, rollback (if shown), and duration (e.g. `3s`) are visible without hovering
- [ ] Hover the agent message and confirm the timestamp (e.g. `01:46`) appears
- [ ] Confirm streaming/in-progress answers still hide the action row until the message completes
- [ ] Run `cd frontend && npm test -- --run src/renderer/components/chat/ChatTimelineItems.test.tsx`

image


ronishrohan (approved):
lgtm

Issue #4557, opened 2026-08-27
feat(tracker): OneDev tracker provider, and making issue intake multi-provider


johnkattenhorn:
Companion to the OneDev SCM provider issue. This half is about issue intake, and it raises one question that's worth answering independently of whether you want OneDev — so I've split it out rather than burying it.

#### The general point: intake is pinned to GitHub in the wiring, not in the port

`ports.Tracker` is provider-neutral — three methods (`Get`, `List`, `Preflight`), normalised through `domain.NormalizedIssueState`. There's a `tracker/gitlab` adapter and a `trackermulti` dispatcher already in the tree.

But `internal/daemon/tracker_intake_wiring.go` constructs intake as:

```go
resolver := trackerintake.SingleTrackerResolver{
    Provider: domain.TrackerProviderGitHub,
    Adapter:  newLazyGitHubTracker(logger),
}
```

So `trackermulti` exists but nothing routes intake through it, and the GitLab tracker adapter can't be reached from the intake loop even though `startSCMObserver` already does exactly this multi-provider pattern for SCM. Unless we've misread it, that means today a GitLab-hosted project can have its MRs observed but not its issues ingested.

Is that a deliberate staging decision, or just the part of the multi-provider work that hasn't been done yet? Happy to send the `SingleTrackerResolver` → `trackermulti` change as a small standalone PR mirroring the SCM wiring, independent of OneDev — it seems useful on its own.

We're flagging it clearly because it's a change to shared wiring rather than an additive adapter, and we don't want to surprise you with it buried inside a forge PR.

#### The OneDev-specific part

`/~api/issues` is available and the mapping is straightforward — the adapter would follow `tracker/gitlab` (~530 lines) and reuse `tracker/httpkit`.

One thing worth stating plainly: **`ports.Tracker` is read-only**, so AO would ingest OneDev issues to hydrate a session prompt but never write status back. That's fine for our use, but if you have a longer-term intent to add write-back we'd rather design toward it than against it.

#### Proposed shape

- `internal/adapters/tracker/onedev/` — `tracker.go`, `auth.go`, `doc.go`, reusing `httpkit`
- `internal/daemon/tracker_intake_wiring.go` — `SingleTrackerResolver` → `trackermulti`, kept as its own commit so you can take the adapter and leave the wiring change, or vice versa

#### What we're asking

1. Is the GitHub-pinned intake wiring intentional?
[...]

PR #4654, opened 2026-08-29
feat(scm): OneDev SCM provider

### LINKED ISSUE #4556: feat(scm): OneDev SCM provider
First — thanks for AO. We've been running it against a real project this week and the PR/CI loop is the thing that sold us: a worker opened a PR, its own CI went red on a `pnpm/action-setup` version conflict, and it diagnosed and fixed it without anyone touching it. The reviewer then found five genuine defects a green build would have shipped. That's the workflow we want, and the only thing standing between us and using AO properly is that our repos live on [OneDev](https://onedev.io/), not GitHub or GitLab.

We're planning to write a OneDev SCM provider in a fork. We'd much rather it lived upstream — opening this first per CONTRIBUTING so scope can be agreed before we invest the time, and to check you actually want it.

**Context for why we're asking rather than just sending a PR:** #825 (Forgejo), #741 (Bitbucket) and #1674 were all closed during the Go rewrite, with the note *"If the underlying change still makes sense for the current app, please open a fresh PR against the new codebase."* We'd like to avoid a fourth forge adapter written against a tree that's about to move.

#### Feasibility — we've mapped every method against a live OneDev instance

`observe/scm.Provider` is a clean seam and `adapters/scm/gitlab` is a good template. All seven methods map to real endpoints:

| `scm.Provider` method | OneDev endpoint | Notes |
|---|---|---|
| `ParseRepository` | — | URL parsing only |
| `ListPRsByRepo` | `GET /~api/pulls?query=&offset=&count=` | Verified live. Returns `lastActivity.date`, usable for `updatedAfter` |
| `FetchPullRequests` | `GET /~api/pulls/{id}` + `/{id}/merge-preview` | `merge-preview` supplies mergeability / conflict state |
| `FetchReviewThreads` | `GET /~api/pulls/{id}/reviews`, plus pull-request and code comment resources | |
| `FetchFailedCheckLogTail` | `GET /~api/streaming/build-logs/{buildId}` | Purpose-built streaming endpoint |
| `CommitChecksGuard` | `GET /~api/builds?query=` | See caveat |
| `RepoPRListGuard` | `GET /~api/pulls?query=` | See caveat |

#### One design question we'd want your steer on

**OneDev returns no `ETag` or `Last-Modified` headers.** We checked the response headers directly. The two `*Guard` methods exist so the observer can skip unchanged data cheaply, which matters a lot for a rate-limited GitHub and much less for a self-hosted, usually LAN-local OneDev.

Our proposal is to synthesise a guard from a cheap `max(lastActivity)` query and fall back to reporting "changed" when that isn't conclusive — correctness preserved, at the cost of more full fetches. If you'd rather `SCMGuardResult` grew an explicit "provider has no conditional-request support" signal instead of each adapter improvising, we're happy to do it that way — that seems like your call rather than ours.

#### Shape of the change

Additive, mirroring `adapters/scm/gitlab`:

- `internal/adapters/scm/onedev/` — `provider.go`, `observer_provider.go`, `client.go`, `auth.go`, `cache.go`
- `internal/config/config.go` — `OneDevConfig` following the existing `AO_GITLAB_ALLOWED_HOSTS` / `AO_GITLAB_HOST_TOKENS` pattern (OneDev is always self-hosted, so per-host tokens matter from day one)
- `internal/daemon/scm_wiring.go` — register as a `scmmulti.NamedProvider`, same degrade-gracefully behaviour as the existing two

Tests would follow `provider_test.go`, plus integration coverage against a real instance — including the absent-ETag path, so the fallback is exercised rather than assumed.

A companion issue covers the tracker side and the issue-intake wiring.

#### What we're asking

1. Do you want OneDev support upstream at all?
2. If yes, any preference on the guard question above before we write it?
3. Happy to follow the claim-and-assignment process — just say the word and we'll pick it up.

Either way, thanks — we'll build it in a fork regardless because we need it, so the only question is whether it's useful to you too.

johnkattenhorn:
## What

A OneDev SCM provider, so AO can observe and act on pull requests hosted on a self-hosted OneDev instance the way it already does for GitHub and GitLab.

Eight commits: provider foundation and credential/host resolution, the observer implementation (PR state, CI builds, build logs, review state, with a small cache), authenticated-identity resolution, registration in every SCM constructor, and a `ao doctor` section.

## Why

Fixes #4556.

OneDev is a common self-hosted choice for people who cannot put their code on github.com, and today AO simply has nothing for it — a OneDev repo is invisible to the board, so none of the session/PR/CI lifecycle works.

## How

Points reviewers may want to push back on, stated rather than buried:

- **`MergePullRequest` deliberately returns unsupported.** OneDev's merge API has no compare-and-swap — no expected-head parameter — so a merge cannot be made safe against a racing push the way the GitHub and GitLab paths are. Rather than implement a merge that can silently land the wrong head, the adapter declines and the user merges in the OneDev UI. That is the one capability gap versus the other providers, and it is a design decision, not an oversight.
- **Hosts come from configuration, not a fixed base URL.** OneDev is always self-hosted, so there is no public API root to assume. Instances are read from an allowlist and credentials resolved per host. `ebfa1df26` tightened that resolution after a mis-keyed per-host token resolved to the wrong credential; the two `test(scm)` commits are the regression tests for it, including credential *ordering*.
- **`doctor` builds the real adapter rather than re-implementing its checks.** Otherwise doctor can pass while the daemon fails. An unconfigured allowlist and an unreachable instance are WARN (self-hosted instances are commonly on private networks, and most deployments never set the allowlist); a *rejected* credential is a FAIL.
- **`fix(daemon): register OneDev in every SCM constructor`** is a genuine bug in the shape of the wiring: `scm_wiring.go` built providers in three places and only `startSCMObserver` registered OneDev, so `ao session claim-pr` on a OneDev PR failed with `SCM_UNAVAILABLE`. Worth a look as a pattern question — three constructors that must be kept in sync will drift again.
- **`golangci-lint` fixes are folded in, not bolted on.** v2.12.2 flags `gocritic ptrToRefParam` on a `*map` parameter (the pointer existed only so an overflow reset could rebind the map; `clear()` does it in place) and `makezero` on two build-log frame helpers (sizing the frame once and copying into it is lint-clean and one allocation instead of two). Both are fixed in the commits that introduce that code rather than in a trailing cleanup commit. No behaviour change.

Tests are in the same commits as the code they cover. No changes outside the OneDev adapter, its wiring, config resolution, and doctor.

## Testing

Rebased onto current `main` and verified from `backend/`:

```
go build ./...   # clean
go vet ./...     # clean
go test ./...
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs   # 0 issues
```

The full backend suite passes — no failures, no skips I introduced.

Two tests needed the host set up the way CI has it, and neither is related to this branch (both behave identically on unmodified `main`, which I checked against a clean worktree):

- `TestCopilotClassicPATIsNotAnAuthorizationSignal` isolates env vars and `HOME` but not the `gh` binary. `copilotGHAuthStatus` shells out to `gh auth status`, so on a developer machine with a logged-in `gh` the probe reports authorized and the assertion inverts. Worth tightening the test's isolation upstream — happy to send that separately.
- The tmux integration tests need a running tmux server whose `base-index`/`pane-base-index` are 0. `respawnPaneArgs` and the pane-pid probe target `<session>:0.0` (`commands.go:24`, `:61`) and nothing reads those options, so on the very common `base-index 1` config `Restart` fails with `can't find pane: 0`. That looks like a real bug in the runtime rather than a test artifact — it would hit session restore in normal use — but it is out of scope here, so I have left it alone and will raise it on its own.

Beyond the test suite, this branch is what my daily AO runs against a real OneDev instance: it observes PRs, reports CI status, feeds back build logs and resolves identity correctly.

## Checklist

---

@illegalcall @somewherelostt — tagging you as the on-call pair per your note on #4556. Happy to split this smaller, reorder it, or rework any of the decisions above. #4655 covers #4557 (OneDev issue tracker + multi-provider intake) and is stacked on this branch — its first eight commits are this PR, so review this one first and it will shrink to four commits once this lands.

### DIFF
--- a/backend/internal/adapters/scm/onedev/observer_provider.go
+++ b/backend/internal/adapters/scm/onedev/observer_provider.go
@@ -1,0 +1,1114 @@
+package onedev
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+const (
+	// fetchConcurrency bounds the parallel per-PR detail fetches inside one
+	// FetchPullRequests batch, matching the GitLab adapter.
+	fetchConcurrency = 5
+	// batchLimit is the largest ref batch FetchPullRequests accepts. It
+	// mirrors the observer's own BatchSize so an oversized batch is a
+	// programming error, not a silent partial fetch.
+	batchLimit = 25
+	// prListPageSize is the per-page count for PR listings. The client clamps
+	// to OneDev's ceiling, above which the server answers HTTP 406.
+	prListPageSize = 100
+	// prBuildsPageCount bounds how many of a pull request's builds are
+	// considered when assembling its CI snapshot. Only the builds matching the
+	// PR's current build commit survive the filter, and OneDev submits one
+	// build per job per commit, so this is generous even for a PR that has
+	// been retried repeatedly.
+	prBuildsPageCount = 50
+	// checksGuardWindow is how many of a project's most recent builds
+	// CommitChecksGuard hashes. See that method for why the window exists and
+	// what it bounds.
+	checksGuardWindow = 20
+)
+
+// ---------------------------------------------------------------------------
+// Query construction
+// ---------------------------------------------------------------------------
+
+// quoteQueryValue renders a value for OneDev's query DSL, which quotes values
+// with double quotes and escapes with a backslash. Project paths and branch
+// names do not normally contain either, but a value is interpolated into a
+// query the server parses, so it is escaped rather than trusted.
+func quoteQueryValue(s string) string {
+	return `"` + strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(s) + `"`
+}
+
+// queryDate formats a timestamp for OneDev's date criteria. OneDev parses the
+// value in UTC whether or not an offset is present, so the timestamp is always
+// converted to UTC first: sending a local-zone wall-clock time would shift the
+// cursor by the offset and silently drop updates inside that window.
+func queryDate(t time.Time) string {
+	return t.UTC().Format("2006-01-02T15:04:05Z")
+}
+
+// projectRef renders a project path for a query. OneDev identifies a project
+// by its full path ("Homelab/curatarr"), which is exactly SCMRepo.Repo for
+// this adapter — see ParseRepository.
+func projectRef(repo ports.SCMRepo) string { return repo.Repo }
+
+// pullRequestRef renders the project-qualified pull-request reference OneDev's
+// build query requires, e.g. "productone#106". A bare number is rejected with
+// "Reference project not specified" even when the surrounding query already
+// constrains the project, so the project path is always included.
+func pullRequestRef(repo ports.SCMRepo, number int) string {
+	return projectRef(repo) + "#" + strconv.Itoa(number)
+}
+
+// ---------------------------------------------------------------------------
+// REST payloads
+// ---------------------------------------------------------------------------
+
+// restPullRequest is the subset of OneDev's pull-request payload AO consumes.
+// Note that "id" (the internal request id every /pulls/{requestId} route is
+// keyed by) is distinct from "number" (the per-project number users see).
+type restPullRequest struct {
+	ID              int64             `json:"id"`
+	Number          int               `json:"number"`
+	Status          string            `json:"status"`
+	Title           string            `json:"title"`
+	TargetBranch    string            `json:"targetBranch"`
+	SourceBranch    string            `json:"sourceBranch"`
+	BaseCommitHash  string            `json:"baseCommitHash"`
+	BuildCommitHash string            `json:"buildCommitHash"`
+	MergeStrategy   string            `json:"mergeStrategy"`
+	CheckError      string            `json:"checkError"`
+	SubmitDate      *time.Time        `json:"submitDate"`
+	CloseDate       *time.Time        `json:"closeDate"`
+	LastActivity    *restLastActivity `json:"lastActivity"`
+	SubmitterID     int64             `json:"submitterId"`
+	TargetProjectID int64             `json:"targetProjectId"`
+	SourceProjectID int64             `json:"sourceProjectId"`
+}
+
+type restLastActivity struct {
+	Date        *time.Time `json:"date"`
+	Description string     `json:"description"`
+	UserID      int64      `json:"userId"`
+}
+
+// restMergePreview is OneDev's precomputed merge of a request into its target.
+// A null mergeCommitHash means the merge conflicts; an absent preview means
+// OneDev has not computed one yet.
+type restMergePreview struct {
+	TargetHeadCommitHash string `json:"targetHeadCommitHash"`
+	HeadCommitHash       string `json:"headCommitHash"`
+	MergeStrategy        string `json:"mergeStrategy"`
+	MergeCommitHash      string `json:"mergeCommitHash"`
+}
+
+type restPullRequestUpdate struct {
+	ID                   int64      `json:"id"`
+	HeadCommitHash       string     `json:"headCommitHash"`
+	TargetHeadCommitHash string     `json:"targetHeadCommitHash"`
+	Date                 *time.Time `json:"date"`
+}
+
+type restBuild struct {
+	ID         int64      `json:"id"`
+	Number     int        `json:"number"`
+	JobName    string     `json:"jobName"`
+	Status     string     `json:"status"`
+	CommitHash string     `json:"commitHash"`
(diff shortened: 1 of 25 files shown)


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/adapters/scm/onedev/observer_provider.go
```diff
@@ -0,0 +1,1114 @@
...
+	if !updatedAfter.IsZero() {
+		criteria = append(criteria, `"Last Activity Date" is since `+quoteQueryValue(queryDate(updatedAfter)))
+	}
+	q := url.Values{
+		"query": {strings.Join(criteria, " and ") + ` order by "Last Activity Date" desc`},
+		"count": {strconv.Itoa(prListPageSize)},
+	}
+
+	var result []ports.SCMPRObservation
+	_, err = client.doGETPaginated(ctx, "/pulls", q, func(body []byte) (int, error) {
```
whoisasx, line 235: [P1] Please fail closed when this listing is truncated. doGETPaginated stops after 10 pages of 100 and returns truncated=true, but that result is discarded here. The observer treats this call as successful and can advance LastSyncCursor to now; any PR omitted after the first 1,000 then falls behind the incremental window and may never be discovered. Please either continue pagination to completion or return an error on truncation so the observer does not advance the repository cursor. A regression test with more than maxPaginationPages * maxPageCount results would pin the recovery behavior.

backend/internal/adapters/scm/onedev/observer_provider.go
```diff
@@ -0,0 +1,1114 @@
...
+	client, err := p.clientForRepo(repo)
+	if err != nil {
+		return ports.SCMGuardResult{}, err
+	}
+	q := url.Values{
+		"query":  {`"Target Project" is ` + quoteQueryValue(projectRef(repo)) + ` order by "Last Activity Date" desc`},
+		"offset": {"0"},
+		"count":  {"1"},
+	}
+	resp, err := client.doGET(ctx, "/pulls", q)
```
whoisasx, line 448: [P2] A matching hash of only the first row cannot prove that the repository-wide PR set is unchanged. Equal Last Activity Date values can leave ordering unresolved, and deleting or otherwise changing a non-leading PR does not necessarily replace this row; in those cases NotModified suppresses ListPRsByRepo even though the set changed. The proposal in #4556 was to report changed whenever the synthetic fallback is inconclusive. Please keep that fail-open-for-refresh behavior unless OneDev exposes a repository-wide monotonic value that makes this guard authoritative.

backend/internal/adapters/scm/onedev/observer_provider.go
```diff
@@ -0,0 +1,1114 @@
...
+	return ports.SCMCIObservation{
+		Summary:           string(ciSummary(checks)),
+		HeadSHA:           headSHA,
+		FailedFingerprint: failedFingerprint(headSHA, failed),
+		Checks:            checks,
+		FailedChecks:      failed,
+		// The build query is bounded by prBuildsPageCount but narrowed to one
+		// build per job name at the current commit, so the snapshot is
+		// complete for the commit rather than a truncated window.
+		Partial: false,
```
whoisasx, line 806: [P1] This snapshot cannot be marked complete after one bounded 50-build page. A current commit may have more than 50 jobs, and later reruns for superseded commits can also consume part of the newest-first window. Either case can hide a failing or pending current job while ciSummary reports the visible subset as passing. Please paginate until the current commit's job set is known to be complete, or propagate Partial/unknown so lifecycle policy cannot treat an incomplete window as authoritative.

whoisasx (changes requested):
John, thank you for the depth of work here. The host allowlist and credential isolation, aligned partial-failure observations, explicit partial review snapshots, live-instance validation, and the decision to reject unsafe OneDev merges all show careful attention to AO's existing contracts. I am requesting changes only for the three correctness boundaries called out inline: a truncated PR listing can advance the discovery cursor and permanently omit PRs, the single-row synthetic guard is not conclusive for the repository-wide set, and a bounded 50-build response is currently marked complete. Once those are made fail-closed and the straightforward lifecycle_wiring.go conflict with current main is resolved, the overall provider structure looks directionally strong. Verification from my side: the branch passes go build ./... and git diff --check. I did not run Go tests, and GitHub checks are not currently running while the PR is conflicted.

i-trytoohard:
@johnkattenhorn Marking as stale: no activity for 25 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

johnkattenhorn:
Rebased onto current main and addressed all three findings in `fa9403c41`.

- Truncated PR listings now return an error, preventing the observer from advancing its repository cursor on an incomplete result.
- The first-row hash no longer suppresses discovery. Without a repository-wide change marker, the guard requests a refresh.
- A full 50-build response now reports partial/unknown before filtering by commit, so missing jobs cannot turn an incomplete CI snapshot into a passing result.

Regression tests cover all three cases. The backend builds and strict lint passes. The full race suite found an installed Codex protocol mismatch that also fails on unchanged upstream. A tmux failure passed on an isolated retry.

PR #4655, opened 2026-08-29
feat(tracker): OneDev issue tracker, and multi-provider issue intake

### LINKED ISSUE #4557: feat(tracker): OneDev tracker provider, and making issue intake multi-provider

johnkattenhorn:
> [!IMPORTANT]
> **Stacked on #4654.** GitHub will not let a cross-fork PR use a fork branch as its base, so this shows twelve commits — the first eight are #4654. **Only the last four are the subject of this PR.** Review #4654 first; once it lands, rebasing this drops it to four commits. If you would rather I held this until then, say so and I will close it and re-open later.
>
> The four commits that belong to this PR:
> - `feat(tracker): add a OneDev issue-tracker adapter`
> - `feat(daemon): register the OneDev tracker in the multi-tracker`
> - `feat(intake): make issue intake multi-provider`
> - `feat(cli): expose the issue-intake provider on set-config`

## What

A OneDev issue-tracker adapter, and the change that makes issue intake genuinely multi-provider rather than GitHub-with-a-GitLab-branch.

## Why

Fixes #4557.

Issue intake is the part of AO people actually use to drive a fleet from a board. With the SCM provider from #4654 alone, a OneDev user can have their PRs observed but still cannot spawn a worker from an issue — the intake path had no way to reach a third provider.

## How

- **`ports.Tracker` is read-only, and this respects that.** The adapter implements `Get`/`List`/`Preflight` only. AO hydrates a worker's prompt from an issue but never writes status back, so a OneDev board does not reflect what a fleet is doing. That is upstream's port design and I have not widened it here — I mention it because it is the first question a user asks, and the honest answer today is "transition the issue from the worker side".
- **States are not pre-validated.** OneDev issue states are per-instance configuration and no endpoint lists them to a non-administrator. Rather than guess a fixed set, `states.go` surfaces the server's own error plus the states it can observe in use. It is the least-wrong option, but I would take a better one.
- **The multi-provider change is the load-bearing part**, not the adapter. `tracker_intake_wiring.go` and `observe/trackerintake/observer.go` previously assumed the provider; they now resolve it per project. `domain/tracker.go` carries the provider through, and `ao project set-config --tracker-provider` exposes it. That touches `openapi.yaml` and the generated `frontend/src/api/schema.ts` — both regenerated with `npm run api`, not hand-edited.
- **No behaviour change for existing GitHub or GitLab projects.** A project with no explicit provider resolves exactly as it does today; the tests in `tracker_intake_wiring_test.go` and `observer_test.go` pin that.

## Testing

Rebased onto current `main` and verified from `backend/`:

```
go build ./...   # clean
go vet ./...     # clean
go test ./...
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs   # 0 issues
```

The full backend suite passes — no failures, no skips I introduced.

Two tests need the host set up the way CI has it, and neither relates to this branch (both behave identically on unmodified `main`, checked against a clean worktree): `TestCopilotClassicPATIsNotAnAuthorizationSignal` isolates env vars and `HOME` but not the `gh` binary, so a logged-in `gh` inverts its assertion; and the tmux integration tests need a tmux server with `base-index`/`pane-base-index` at 0, because the runtime targets `<session>:0.0` unconditionally. Detail in #4654.

New coverage: the tracker adapter and its state handling, the multi-tracker registration, the multi-provider intake wiring, the observer, and the CLI flag.

This is also what I run day to day: intake against a real OneDev instance spawns a worker per eligible open issue with the prompt hydrated from the issue body.

## Checklist

---

@illegalcall @somewherelostt — tagging you per your note on #4557. Happy to reshape or resequence any of this.


somewherelostt:

- @illegalcall
- @somewherelostt

Join the session here:
https://discord.gg/H6ZDcUXmq

backend/internal/cli/project.go
```diff
@@ -386,11 +394,33 @@ func buildProjectConfig(opts projectSetConfigOptions) (projectConfig, error) {
...
+	provider := strings.TrimSpace(opts.trackerProvider)
+	if provider != "" {
+		if !domain.IsSupportedIntakeProvider(domain.TrackerProvider(provider)) {
+			return "", usageError{fmt.Errorf("invalid --tracker-provider %q: expected one of %s", opts.trackerProvider, trackerProviderChoices())}
+		}
+		return provider, nil
+	}
 	if opts.trackerIntake || opts.trackerRepo != "" || opts.trackerAssignee != "" {
-		return "github"
+		return string(domain.TrackerProviderGitHub), nil
```
whoisasx, line 410: [P1] Please preserve the provider-inference behavior now on main after #4644. An unspecified provider is intentionally left empty so GitHub/GitLab can be inferred from the project's origin; defaulting it to GitHub here would route existing GitLab projects to the wrong adapter whenever users enable intake without --tracker-provider. After rebasing, return an empty provider unless the flag was explicitly supplied, and keep the current nested-GitLab-path tests alongside the new explicit OneDev cases.

backend/internal/httpd/controllers/dto.go
```diff
@@ -222,7 +222,7 @@ type ListSessionsResponse struct {
 type SpawnSessionRequest struct {
 	ProjectID       domain.ProjectID       `json:"projectId"`
 	IssueID         domain.IssueID         `json:"issueId,omitempty"`
-	TrackerProvider domain.TrackerProvider `json:"trackerProvider,omitempty" enum:"github,gitlab"`
+	TrackerProvider domain.TrackerProvider `json:"trackerProvider,omitempty" enum:"github,gitlab,onedev"`
```
whoisasx, line 225: [P1] Adding onedev to this public request enum currently advertises a manual issue-hydration path that is not implemented end to end. ao spawn still rejects any tracker provider other than github/gitlab, and session issue resolution only preserves Host for GitLab; a OneDev host is dropped, while the fallback repository parser also cannot retain an arbitrarily nested OneDev project path. This fails for multi-host OneDev and can resolve the wrong native ID. Please wire the CLI and service path through with host/nested-path tests, or leave onedev out of SpawnSessionRequest until that path is supported.

backend/internal/adapters/tracker/onedev/tracker.go
```diff
@@ -0,0 +1,779 @@
...
+// AllowedHosts is required — see ErrNoAllowedHosts. HostTokens maps an allowed
+// host to a credential override; hosts without an entry fall back to Token.
+type Options struct {
+	Token        TokenSource
+	HTTPClient   *http.Client
+	UserAgent    string
+	AllowedHosts []string
+	HostTokens   map[string]TokenSource
+
+	// States maps this instance's issue state names onto the normalized
```
whoisasx, line 84: [P1] These workflow settings need the same host scope as credentials. The tracker accepts multiple AllowedHosts, but one global States map and AssigneeField are applied to every instance even though OneDev states and custom fields are configured per instance. Two allowlisted instances with different workflows will therefore classify open/terminal issues or query the assignee field incorrectly on at least one host. Please make these settings host-specific, or reject multi-host tracker construction until a per-host configuration shape is available.

whoisasx (changes requested):
John, this is another substantial and carefully reasoned contribution. The adapter's fail-closed custom-field handling, explicit state normalization, host-aware credentials, and separation of the shared intake work into distinct commits are all useful review affordances. The main complication is that #4644 landed after this branch was cut and now owns the shared multi-provider intake architecture. Please rebase this PR after #4654 and extend the single tracker instance and resolver already on main rather than restoring the older parallel resolver. During that rebase, the three inline blockers need to be addressed: preserve main's provider inference, make the advertised OneDev session-hydration surface genuinely end to end, and scope workflow settings consistently with multi-host support. There is also a desktop integration gap: IntakeFields has no provider field, so the canonical UI cannot select OneDev and a later settings save can clear or misroute a provider configured through the CLI. Please add provider selection/preservation and focused UI coverage, or explicitly narrow the PR/API surface so existing OneDev configuration cannot be overwritten. Verification from my side: the stacked branch passes go build ./... and git diff --check. I did not run Go tests; frontend typecheck was unavailable because dependencies are not installed in the isolated checkout. GitHub checks are not currently running while the stacked PR is conflicted.

i-trytoohard:
@johnkattenhorn Marking as stale: no activity for 25 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

johnkattenhorn:
Rebased on the updated #4654 and addressed the three review findings.

- An omitted provider stays empty for origin inference. The existing nested GitLab coverage is retained alongside explicit OneDev cases.
- OneDev is excluded from the manual spawn request enum until the CLI and session hydration path support it end to end. This PR supports OneDev through issue intake.
- Tracker construction rejects multiple OneDev hosts while workflow states and the assignee field remain global.

The project settings and creation forms now expose the issue tracker. Selecting OneDev survives subsequent edits and disabling intake; choosing Automatic clears the explicit provider.

The affected backend tests, frontend tests and frontend type check pass. An isolated browser fixture confirmed OneDev selection and preservation when disabling intake. The backend suite's Codex protocol mismatch also reproduces on unchanged upstream.

PR #4659, opened 2026-08-30
fix(reviewer): resume Claude after reviewer relaunch

### LINKED ISSUE #4658: bug(reviewer): Claude restore double-hashes session identity
## Bug

A restored Claude Code reviewer can fail immediately with:

> No conversation found with session ID: 03dc146e-e3b6-5d88-a3a3-15fe9f6354bc

The reviewer terminal remains present with no Claude child process, while the associated `review_run` remains durably `running`.

**Source:** AO orchestrator chat report on 2026-08-30  
**Reported by:** AO user (live local observation)  
**Analyzed against:** `d28e46dea577e45f9c995a59f3dfbcfcfe72bb59` (`upstream/main`)  
**AO binary:** packaged rewrite binary, `dev`, verified ready on `127.0.0.1:3001`  
**Environment:** macOS arm64; Claude Code 2.1.251  
**Confidence:** High

Reviewer terminal showing No conversation found

## Reproduction

1. Start a Claude Code reviewer for a worker session.
2. Allow it to create a real Claude conversation/transcript.
3. Restart or otherwise restore the reviewer terminal.
4. Open the Reviewer's terminal tab.
5. Observe `claude --resume <persisted-id>` fail with `No conversation found`.

Observed live instance:

- AO worker: `agent-orchestrator-149` (`pr author avatar`)
- reviewer AO ID: `review-agent-orchestrator-149`
- persisted `review.agent_session_id`: `03dc146e-e3b6-5d88-a3a3-15fe9f6354bc`
- actual Claude transcript ID: `a04d0453-97e8-51ee-84bf-ebe1d2a4a90b`
- persisted-ID transcript: absent
- actual transcript: present and non-empty
- restored pty host invokes `claude --resume 03dc146e-...`, but has no Claude child after the failure
- newest review run `201113cb-b4cf-437a-bb19-e5b0de73a662` has remained `running` since 2026-08-27

## Root Cause

`backend/internal/adapters/reviewer/claudecode/claudecode.go:76-96` first maps the reviewer AO ID to a Claude UUID:

```go
agentSessionID := workeragent.SessionUUID(inv.ReviewerID)
```

It persists that value as `ReviewCommandSpec.AgentSessionID`, but also passes the already-derived UUID as `LaunchConfig.SessionID`.

`backend/pkg/agentruntime/command.go:211-221` treats `SessionID` as an AO identity and derives a Claude UUID again:

```go
cmd = append(cmd, "--session-id", ClaudeSessionID(cfg.SessionID))
```

Consequently:

- AO persists `UUIDv5(reviewerID)`
- Claude launches with `UUIDv5(UUIDv5(reviewerID))`
- restore uses the persisted single-hashed ID
- Claude only has a transcript for the double-hashed ID

This mismatch was introduced by commit `d15fd8277c1b63a1a7ebdb55735b9ed6091cf118` in [PR #3484](https://github.com/Untrivial-ai/agent-orchestrator/pull/3484).

The recent daemon/reconciliation incident triggered a restore and exposed the bug, but did not create the deterministic ID mismatch.

## Fix

1. Make reviewer launch use the exact native ID AO persists. Prefer passing the derived UUID through `LaunchConfig.NativeSessionID`, or pass the raw reviewer AO ID through `SessionID`; do not hash an already-derived native ID.
2. Add an integration-level adapter test asserting the actual emitted `--session-id` equals `ReviewCommandSpec.AgentSessionID`. Existing capture-agent tests only observe the intermediate launch config and therefore missed the second transformation.
3. Preserve existing reviewer conversations created by buggy versions: when the persisted single-hashed transcript is absent but the legacy double-hashed transcript exists, resume the legacy identity and safely migrate/persist it.
4. Ensure a failed reviewer restore transitions its running review run to a failed/retryable state instead of leaving it permanently `running`.

## Impact

- Restored Claude Code reviewers can lose access to their real conversation.
- Automatic review work is blocked until the reviewer is recreated or repaired.
- Review runs may appear active indefinitely despite no reviewer process.
- Every Claude reviewer launched through this path since the regression is potentially affected on restore.

## Related

- [PR #3484](https://github.com/Untrivial-ai/agent-orchestrator/pull/3484): introduced persisted reviewer-native identity during lifecycle restoration.
- [Issue #2702](https://github.com/Untrivial-ai/agent-orchestrator/issues/2702): broader runtime restoration/zombie-session failure mode; related subsystem but not this double-hash defect.

AgentWrapper:
Fixes #4658

## Summary
- pass the persisted Claude reviewer UUID through the explicit native-session field so first launch emits the exact deterministic ID AO owns instead of hashing it twice
- detect and resume both current transcripts and the legacy double-derived transcript, then persist any repaired identity
- when a reviewer pane/process has exited but its durable native ID remains, route the next review launch through `--resume` instead of colliding on `--session-id`
- preserve Claude's read-only tool policy and system prompt on resume, and deliver the new review task as the resume-time user turn
- surface transcript/restore failures without silently fresh-launching; the existing liveness gate continues to notify a genuinely live reviewer rather than starting a duplicate
- fall back to a fresh idle reviewer when native history is unavailable and mark interrupted running review runs failed so they can be retried

## Root cause
Commit `d15fd8277c1b63a1a7ebdb55735b9ed6091cf118` (`fix: align reviewer terminal lifecycle`, PR #3484) introduced deterministic reviewer identity persistence and native restore support, but split ownership incorrectly in two related ways:

1. `ReviewCommand` derived a Claude UUID, persisted it, then passed it as an AO `SessionID`; the worker command builder derived it again. AO therefore persisted the single-derived ID while Claude owned the double-derived transcript, producing `No conversation found` on restore.
2. Ordinary new-review reruns still called `Launcher.Spawn`, whose path always built a fresh command. After cancellation/process exit, `review.agent_session_id` survived while the pane did not, so Claude received `--session-id <existing transcript>` and failed with `Session ID ... is already in use`.

## Tests
- `go test ./internal/adapters/reviewer/... ./internal/review/...`
- `go build ./...`
- `env -u AO_SESSION_ID -u AO_PROJECT_ID -u AO_TMUX_BINARY -u AO_TMUX_SOCKET_NAME -u AO_DATA_DIR -u AO_RUN_FILE go test ./...`
- isolated-cache `golangci-lint v2.12.2 run --path-mode=abs` (`0 issues`)

## Risk
Low and scoped to reviewer identity/restore selection. A normal review launch attempts native resume only when a durable reviewer-native ID exists; first launch remains fresh and deterministic. Claude only resumes after finding a non-empty transcript. Probe/restore errors are surfaced, and a live reviewer pane is still handled by the existing notify path.

### DIFF
--- a/backend/internal/adapters/reviewer/claudecode/claudecode.go
+++ b/backend/internal/adapters/reviewer/claudecode/claudecode.go
@@ -8,6 +8,7 @@ package claudecode
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	workeragent "github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/claudecode"
@@ -79,7 +80,8 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
 		// Pin the same deterministic reviewer-native id we persist. Hooks can
 		// later replace it with Claude's reported id, but restore must never start
 		// from an id that the process was not launched with.
-		SessionID:        agentSessionID,
+		SessionID:        inv.ReviewerID,
+		NativeSessionID:  agentSessionID,
 		WorkspacePath:    inv.WorkspacePath,
 		Prompt:           inv.Prompt,
 		SystemPrompt:     inv.SystemPrompt,
@@ -130,6 +132,13 @@ func (r *Reviewer) ReviewMessage(_ context.Context, inv ports.ReviewInvocation) 
 // ReviewRestoreCommand resumes the reviewer Claude Code conversation captured
 // from hooks, reapplying the same read-only tool policy as a fresh review launch.
 func (r *Reviewer) ReviewRestoreCommand(ctx context.Context, inv ports.ReviewInvocation) (ports.ReviewCommandSpec, bool, error) {
+	if migratedID, ok, err := r.restoreSessionID(ctx, inv); err != nil {
+		return ports.ReviewCommandSpec{}, false, err
+	} else if !ok {
+		return ports.ReviewCommandSpec{}, false, nil
+	} else if migratedID != "" {
+		inv.AgentSessionID = migratedID
+	}
 	cmd, ok, err := agentrestore.Command(ctx, r.agent, inv, agentrestore.Options{
 		Permissions:     ports.PermissionModeAuto,
 		AllowedTools:    reviewerAllowedTools,
@@ -142,6 +151,44 @@ func (r *Reviewer) ReviewRestoreCommand(ctx context.Context, inv ports.ReviewInv
 		cmd.AgentSessionID = workeragent.SessionUUID(inv.ReviewerID)
 	}
 	return cmd, true, nil
+}
+
+// restoreSessionID verifies an explicitly persisted Claude conversation before
+// asking Claude to resume it. Reviewer builds affected by #4658 persisted the
+// single-derived UUID but launched Claude with that UUID derived a second time.
+// Prefer the persisted identity when its transcript exists; otherwise migrate
+// only that exact legacy shape when the double-derived transcript is present.
+// Returning ok=false lets the launcher recreate an idle reviewer instead of
+// starting a doomed `claude --resume` command.
+func (r *Reviewer) restoreSessionID(ctx context.Context, inv ports.ReviewInvocation) (string, bool, error) {
+	persistedID := strings.TrimSpace(inv.AgentSessionID)
+	if persistedID == "" {
+		return "", true, nil
+	}
+	probe, ok := r.agent.(ports.AgentInterfaceHandoffHistoryProbe)
+	if !ok {
+		return persistedID, true, nil
+	}
+	session := ports.SessionRef{ID: inv.ReviewerID, WorkspacePath: inv.WorkspacePath}
+	exists, err := probe.NativeConversationExists(ctx, session, persistedID, nil)
+	if err != nil {
+		return "", false, err
+	}
+	if exists {
+		return persistedID, true, nil
+	}
+	if persistedID != workeragent.SessionUUID(inv.ReviewerID) {
+		return "", false, nil
+	}
+	legacyID := workeragent.SessionUUID(persistedID)
+	exists, err = probe.NativeConversationExists(ctx, session, legacyID, nil)
+	if err != nil {
+		return "", false, err
+	}
+	if exists {
+		return legacyID, true, nil
+	}
+	return "", false, nil
 }
 
 // ReviewCancel stops the active Claude Code reviewer turn while preserving the
--- a/backend/internal/adapters/reviewer/claudecode/claudecode.go
+++ b/backend/internal/adapters/reviewer/claudecode/claudecode.go
@@ -162,14 +162,27 @@ func (r *Reviewer) ReviewRestoreCommand(ctx context.Context, inv ports.ReviewInv
 // starting a doomed `claude --resume` command.
 func (r *Reviewer) restoreSessionID(ctx context.Context, inv ports.ReviewInvocation) (string, bool, error) {
 	persistedID := strings.TrimSpace(inv.AgentSessionID)
-	if persistedID == "" {
-		return "", true, nil
-	}
 	probe, ok := r.agent.(ports.AgentInterfaceHandoffHistoryProbe)
 	if !ok {
 		return persistedID, true, nil
 	}
 	session := ports.SessionRef{ID: inv.ReviewerID, WorkspacePath: inv.WorkspacePath}
+	if persistedID == "" {
+		// No native id was ever captured — e.g. the daemon restarted before the
+		// hook fired on a first-ever pass. agentrestore.Command's caller falls
+		// back to deriving the same deterministic id from inv.ReviewerID
+		// unconditionally, so probe that id here too rather than reporting a
+		// resume for a transcript that may never have been created.
+		fallbackID := workeragent.SessionUUID(inv.ReviewerID)
+		exists, err := probe.NativeConversationExists(ctx, session, fallbackID, nil)
+		if err != nil {
+			return "", false, err
+		}
+		if exists {
+			return fallbackID, true, nil
+		}
+		return "", false, nil
+	}
 	exists, err := probe.NativeConversationExists(ctx, session, persistedID, nil)
 	if err != nil {
 		return "", false, err
--- a/backend/internal/review/launcher.go
+++ b/backend/internal/review/launcher.go
@@ -83,8 +83,8 @@ type LaunchSpec struct {
 type LaunchResult struct {
 	HandleID       string
 	AgentSessionID string
-	// NativeResumed reports whether RestoreTerminal resumed the provider-native
-	// conversation instead of falling back to a fresh idle reviewer process.
+	// NativeResumed reports whether the launch resumed the provider-native
+	// conversation instead of falling back to a fresh reviewer process.
 	NativeResumed bool
 }
 
@@ -383,7 +383,11 @@ func (l *agentLauncher) Spawn(ctx context.Context, spec LaunchSpec) (LaunchResul
 	if err != nil {
 		return LaunchResult{}, err
 	}
-	return l.launchReviewerTerminal(ctx, spec, inv)
+	// A retained native id means this stable reviewer has provider-owned
+	// history even though its terminal process is gone. Recreate the pane by
+	// resuming that conversation; a first launch has no id and still pins the
+	// adapter's deterministic fresh identity through ReviewCommand.
+	return l.launchReviewerTerminalWithMode(ctx, spec, inv, strings.TrimSpace(spec.AgentSessionID) != "")
 }
 
 func (l *agentLauncher) RestoreTerminal(ctx context.Context, spec LaunchSpec) (LaunchResult, error) {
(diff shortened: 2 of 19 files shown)


backend/internal/review/launcher.go
```diff
@@ -409,6 +413,7 @@ func (l *agentLauncher) launchReviewerTerminalWithMode(ctx context.Context, spec
 			}
 			if restoreOK {
 				cmd = restoreCmd
+				nativeResumed = len(restoreCmd.Argv) > 0
```
neversettle17-101: `restoreOK` here only means the adapter chose to handle restore, not that it actually resumed provider-native state. `cursor.ReviewRestoreCommand` still just calls `ReviewCommand` and returns `ok=true`, so this line will mark a fresh relaunch as `NativeResumed=true`. That means `restoreReviewerLocked` skips failing any running Cursor runs after daemon restart even though their original conversation is gone, leaving them stuck in `running` again. Please make native-resume vs fresh-relaunch an explicit adapter signal instead of inferring it from the presence of argv.

neversettle17-101 (commented):
## What This PR Does

This fixes the Claude reviewer session-id mismatch by separating AO reviewer ids from Claude native ids and by probing for a persisted Claude transcript before attempting `--resume`. It also marks running review runs failed when restore falls back to a fresh reviewer so stale in-progress runs do not remain stuck forever.

## Summary of Changes

- Launch Claude reviewers with the stable AO reviewer id as `SessionID` and the single-derived Claude UUID as `NativeSessionID`.
- Probe reviewer transcript history during Claude restore, including a one-time migration path for the legacy double-derived Claude id.
- Thread a `NativeResumed` signal through launcher/engine restore flow and fail running runs when restore falls back or errors.
- Add targeted tests around the new Claude id handling and restore/failure paths.

## Findings

- **Medium**: `NativeResumed` is inferred from `restoreOK && len(argv) > 0`, but not every `ReviewerRestorer` actually resumes a native conversation. `backend/internal/adapters/reviewer/cursor/cursor.go` still implements `ReviewRestoreCommand` by calling `ReviewCommand` and returning `ok=true`, i.e. a fresh relaunch with no provider-state restore. With this PR, that path now reports `NativeResumed=true`, so `restoreReviewerLocked` will skip failing any running runs even though the original reviewer context is gone. After a daemon restart, Cursor review runs can therefore remain stuck in `running` again. This needs an explicit restore/fallback signal from the adapter instead of treating any restore argv as a native resume.

## Regression Risk

The Claude-specific fix looks targeted, but the new cross-adapter `NativeResumed` semantics are broader than the adapters were updated for. Any reviewer whose restore path is really a relaunch can now be misclassified as resumed.

## Testing

Added unit coverage for Claude launch id emission, legacy-id migration, restore fallback, and restore-error run failure handling. I did not run tests locally.

backend/internal/review/launcher.go
neversettle17-101: Addressed in c1d1d0fb1.

I stopped inferring native restore from `restoreOK && len(argv) > 0` and made it an explicit adapter signal on `ports.ReviewCommandSpec.NativeResumed` instead. The shared `agentrestore.Command` helper now marks true provider-native resumes, while relaunch-style adapters like Cursor keep the zero value and therefore report fallback correctly. The launcher now threads only that explicit signal into `LaunchResult.NativeResumed`, and I added coverage for both sides: native resume adapters assert `NativeResumed=true`, and Cursor has a regression test asserting its restore path stays `false`.

Focused tests run:
- `go test ./backend/internal/adapters/reviewer/claudecode ./backend/internal/adapters/reviewer/codex ./backend/internal/adapters/reviewer/opencode ./backend/internal/adapters/reviewer/kimchi ./backend/internal/adapters/reviewer/muse ./backend/internal/adapters/reviewer/cursor ./backend/internal/review`

neversettle17-101 (commented):
## What This PR Does

This updates the reviewer restore fix so Claude launches persist the correct native session id and restore probes can migrate the legacy double-derived Claude id when needed. It also makes native-resume reporting explicit across reviewer adapters so the engine can fail stale running runs only when restore actually falls back to a fresh process.

## Summary of Changes

- Launch Claude reviewers with separate AO session and Claude native session ids, plus restore-time transcript probing and legacy-id migration.
- Add an explicit `NativeResumed` flag to reviewer restore command specs and thread it through the launcher restore result.
- Fail running restored review runs when restore errors or when the adapter relaunches fresh instead of resuming native provider state.
- Extend unit coverage for Claude id handling, explicit native-resume signaling, and restore fallback/error behavior.

## Findings

No blocking issues found in the reviewed diff. This looks ready to merge.

## Regression Risk

Low. The main risk is adapter-specific restore behavior, but this patch now makes that distinction explicit instead of inferring it from argv shape.

## Testing

Added targeted unit tests for Claude reviewer launch/restore identity handling, launcher native-resume signaling, engine restore failure handling, and Cursor's relaunch restore path. I did not run tests locally.

neversettle17-101 (commented):
## What This PR Does

This follow-up fixes the remaining dead-pane relaunch case by reusing a recorded reviewer native conversation when a new review run starts after the terminal process disappeared. It also threads the active run task into native restore commands so Claude and other agent-backed reviewers resume directly into the new review request instead of reopening only an idle terminal.

## Summary of Changes

- Teach shared reviewer restore plumbing to include the new task prompt when relaunching for an active run, while keeping idle restore prompt-free.
- Update launcher spawn behavior to attempt a native resume whenever a persisted reviewer native session id already exists, not just during idle terminal restore.
- Add coverage for active-run relaunch, explicit restore errors, transcript probing, and resume argv contents for Claude.
- Keep the explicit `NativeResumed` signal introduced in the previous patch set and continue using it to distinguish true resumes from fresh relaunches.

## Findings

No blocking issues found in the reviewed diff. This looks ready to merge.

## Regression Risk

Low. The main behavior change is in dead-pane reviewer relaunch, and the added tests cover both the new native-resume path and the intended fresh-fallback/error cases.

## Testing

Added targeted unit tests for agent-backed restore prompt delivery, Claude transcript probing and resume argv generation, and launcher behavior when relaunching with or without a persisted native session id. I did not run tests locally.

neversettle17-101:
@iamasx / @nikhilachale can you test this out if you have claude code?

neversettle17-101:
Before:
Screenshot 2026-09-10 at 8 14 27 PM

After:
Screenshot 2026-09-10 at 9 27 28 PM

backend/internal/adapters/reviewer/claudecode/claudecode.go
```diff
@@ -144,6 +153,44 @@ func (r *Reviewer) ReviewRestoreCommand(ctx context.Context, inv ports.ReviewInv
...
 }
 
+// restoreSessionID verifies an explicitly persisted Claude conversation before
+// asking Claude to resume it. Reviewer builds affected by #4658 persisted the
+// single-derived UUID but launched Claude with that UUID derived a second time.
+// Prefer the persisted identity when its transcript exists; otherwise migrate
+// only that exact legacy shape when the double-derived transcript is present.
+// Returning ok=false lets the launcher recreate an idle reviewer instead of
+// starting a doomed `claude --resume` command.
+func (r *Reviewer) restoreSessionID(ctx context.Context, inv ports.ReviewInvocation) (string, bool, error) {
```
neversettle17-101, line 163: Minor/non-blocking: when `inv.AgentSessionID` is empty, this early-returns `("", true, nil)` without probing transcript existence. The call later falls through to `agentrestore.Command`, whose underlying `RestoreIdentity` fallback derives a deterministic UUID from `inv.ReviewerID` and returns `ok=true` unconditionally — so `NativeResumed` gets reported as true even if that transcript was never actually created (e.g. daemon restart before the hook captured a native id). Given this PR's whole point is verifying native-resume identity before trusting it, probing this case too (or documenting why it's out of scope) would close the gap fully. Not introduced by this PR and narrow in practice, so not blocking.

neversettle17-101 (commented):
Reviewed PR #4659 (fix(reviewer): resume Claude after reviewer relaunch). Diffed base c0231777 against head c050e642.

**Summary of the change**: Spawn() now attempts a native `ReviewRestoreCommand` when a persisted `AgentSessionID` exists instead of always fresh-launching (fixing the `--session-id` collision bug). `ReviewCommandSpec.NativeResumed` propagates whether a launch actually resumed an existing provider-native conversation, and `Engine.restoreReviewerLocked` now fails any running review run with a retryable message when a restore falls back to a fresh idle reviewer, instead of leaving it silently stuck as "running". The claude-code reviewer adapter also gains `restoreSessionID`, which probes for an existing transcript before resuming, and transparently migrates the legacy double-hashed session id from the #4658 regression.

**Assessment**: The logic is sound and the change is well covered by new unit tests across `launcher_test.go`, `review_test.go`, and the claude-code reviewer test suite (existing-transcript resume, legacy migration, missing-transcript fallback, probe-error propagation, and the Spawn resume/fresh-launch/error-propagation paths). I did not find a correctness bug that this PR introduces.

**One non-blocking observation** (see inline comment on `claudecode.go`): `restoreSessionID` skips the transcript-existence probe entirely when `inv.AgentSessionID` is empty, falling through to `agentrestore.Command`, which will still mark `NativeResumed: true` via the adapter-level deterministic-UUID fallback even though no transcript may actually exist yet for that id. This is a narrow, pre-existing gap (not introduced by this diff) that this PR mostly closes for the persisted-id case — worth a follow-up but not blocking.

**Verdict: Approve.**

backend/internal/adapters/reviewer/claudecode/claudecode.go
neversettle17-101: Fixed in 45dbcecf6.

`restoreSessionID` now probes the deterministic fallback id (`workeragent.SessionUUID(inv.ReviewerID)`) when `inv.AgentSessionID` is empty too, instead of trusting `agentrestore.Command`'s unconditional fallback derivation. If that transcript doesn't exist, it returns `ok=false` to force a fresh launch instead of reporting `NativeResumed=true` for a conversation that was never created. Added `TestReviewRestoreCommandProbesFallbackWhenNoNativeSessionIDWasEverCaptured` and `TestReviewRestoreCommandFallsBackToFreshWhenNoNativeSessionIDWasEverCaptured` to cover both outcomes.

neversettle17-101 (approved):

PR #4661, opened 2026-08-30
feat: add sidebar hover preview

### LINKED ISSUE #4647: feat(desktop): reveal the collapsed sidebar after a pointer dwell at the left edge
## Feature request

When the desktop sidebar is fully collapsed off-canvas, let the user reveal it by holding the pointer at the far-left screen edge for a deliberate 250 ms dwell. The visible top-left titlebar sidebar button, identified by its Expand tooltip, should open the temporary preview immediately on pointer hover.

The current sidebar has a narrow `SidebarRail` that opens the panel on click. The shell also keeps the sidebar open from the keyboard shortcut. Under Browser workspace pressure, the app can use a separate icon rail, but that is a different state and should keep its current behavior.

This request uses two intentional pointer paths. The far-left edge needs a 250 ms dwell. The visible top-left titlebar button opens on hover without waiting for that dwell.

## Proposed behavior

- Add a narrow activation zone at the far-left screen edge.
- Start a 250 ms dwell timer only when the pointer enters the edge activation zone.
- Cancel the edge timer when the pointer leaves the zone before 250 ms. A quick pass along the edge should not open the sidebar.
- When the pointer hovers the visible top-left titlebar sidebar button, the temporary overlay should open immediately. This button is the control that shows the Expand tooltip.
- After either trigger opens the preview, show the sidebar as a temporary overlay. The overlay must not change the saved sidebar preference, reclaim layout width, or move the center panel.
- Reveal and dismiss the temporary overlay with a short, subtle, smooth slide/fade. Keep the 250 ms edge dwell, immediate titlebar-button hover, and 55 ms dismissal delay unchanged.
- Keep the temporary overlay open while the pointer is over its trigger or the sidebar. Moving from the titlebar button into the overlay must keep it open.
- Keep pointer transfer from either trigger into the overlay independent of animation progress and animated hit bounds. It must not depend on a moving target, so the old race from #3372 cannot recur.
- Leaving both the active trigger area and the overlay should start a 55 ms dismissal delay. Re-entering either area before the overlay closes should cancel the delay.
- Clicking the visible top-left titlebar button or the sidebar expand control while the overlay is open must promote it to the normal pinned-open state. The keyboard shortcut must continue to open the pinned state directly.
- Once pinned open, the sidebar must stay open until the user explicitly collapses it. Pointer movement alone must not collapse it.
- Preserve the existing click, keyboard, resize, and compact icon-rail behavior.

## Accessibility and accidental-trigger requirements

Pointer preview is an optional convenience, not the only way to reach the sidebar.

- Keep the visible titlebar button, its accessible Expand name, and the keyboard shortcut available.
- Do not move keyboard focus into the temporary overlay or add duplicate focus stops for hidden controls.
- Expose the expand/collapse state and accessible names through the existing controls. Screen-reader and keyboard users must have an equivalent path.
- Respect reduced-motion preferences.
- Do not trigger the edge preview from a pointer-down, drag, text selection, or a brief crossing of the edge. The titlebar button's immediate hover preview is intentional, but its click path must remain available.
- Keep the edge activation zone narrow enough that ordinary pointer travel along the left edge does not repeatedly open the sidebar.
- Account for platform chrome, including macOS traffic-light clearance and the Windows titlebar, so the activation targets do not interfere with window controls.

## Acceptance criteria

- [ ] In the fully collapsed off-canvas state, holding the pointer in the far-left edge activation zone for 250 ms opens a temporary sidebar overlay.
- [ ] In the fully collapsed off-canvas state, hovering the visible top-left titlebar sidebar button, the one with the Expand tooltip, opens the temporary overlay immediately without a dwell.
- [ ] A quick pointer pass across the far-left edge does not open the overlay. The titlebar button remains an intentional immediate-hover trigger.
- [ ] Moving the pointer from the titlebar button into the overlay keeps the preview open.
- [ ] Leaving both the active trigger area and the overlay starts a 55 ms delayed dismissal. Re-entering either area before dismissal keeps the overlay open.
- [ ] Clicking the visible top-left titlebar button or the sidebar expand control while the preview is visible pins the sidebar open. The keyboard shortcut still pins it open.
- [ ] The temporary overlay does not persist the open preference, change the layout gap, or resize the center panel.
- [ ] A pinned-open sidebar remains open until an explicit collapse action.
- [ ] Keyboard navigation, screen-reader naming/state, focus order, and reduced-motion behavior remain correct in both collapsed and temporary-overlay states.
- [ ] The temporary overlay reveals and dismisses with a short, subtle, smooth slide/fade. The 250 ms edge dwell, immediate titlebar-button hover, and 55 ms leave dismissal remain unchanged.
- [ ] With reduced motion enabled, the overlay skips movement or uses a reduced transition while preserving the same trigger, keep-open, and dismissal behavior.
- [ ] Pointer transfer from either trigger into the overlay remains reliable during the transition and does not depend on animated bounds, avoiding the old race in #3372.
- [ ] Pointer drag and text-selection gestures do not cause accidental edge activation.
- [ ] Automated coverage exercises the 250 ms edge dwell, immediate titlebar-button hover, edge dwell cancellation, temporary overlay dismissal, click-to-pin, and the existing keyboard paths on the supported desktop platforms.

## Related history

- [Issue #3372](https://github.com/Untrivial-ai/agent-orchestrator/issues/3372) documented the old immediate toggle-hover preview race, where the preview closed before the pointer could reach it. This request intentionally uses immediate hover only for the visible top-left titlebar button, while the far-left edge requires a 250 ms dwell. Both paths must keep the preview open as the pointer enters the sidebar and must use delayed dismissal after leaving.
- [PR #4263](https://github.com/Untrivial-ai/agent-orchestrator/pull/4263) removed the old collapsed-sidebar hover opening from the current frontend. This request brings back immediate preview only for the explicit visible top-left titlebar button and keeps the deliberate 250 ms dwell for the far-left edge.
- [PR #3509](https://github.com/Untrivial-ai/agent-orchestrator/pull/3509) covered the earlier keep-open zone and click-to-pin interaction.

Inspected against commit `dc1ea86e841e8a879b62f44b67682cce26c26767`.

Rishet11:
## Summary
- Add a 250 ms far-left edge dwell that temporarily opens collapsed adaptive and off-canvas sidebars without changing the main-content layout.
- Open the same temporary overlay immediately from the actual titlebar sidebar button, with stable trigger and logical sidebar pointer zones to support trigger-to-sidebar transfer.
- Keep temporary dismissal at 55 ms, preserve click-to-pin and Cmd/Ctrl+B keyboard pinning, and reject touch, drag, and text-selection activation.
- Add a short reduced-motion-safe reveal and hide transition while preserving pinned-sidebar behavior and layout gaps.

## Coverage
- Focused Sidebar, titlebar, Linux titlebar, Windows titlebar, and shell tests cover adaptive/off-canvas peek, pinning, stable hover zones, dismissal, and accessibility behavior.
- No automated AO Browser verification was performed.

Closes #4647

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -18,6 +18,7 @@ import {
 	SIDEBAR_DEFAULT_WIDTH,
 	SIDEBAR_MIN_WIDTH,
 } from "./Sidebar";
+import { TitlebarNav } from "./TitlebarNav";
 import type { WorkspaceSession, WorkspaceSummary } from "../types/workspace";
 import { agentsQueryKey } from "../hooks/useAgentsQuery";
 import { useUiStore } from "../stores/ui-store";
@@ -134,7 +135,8 @@ vi.mock("@tanstack/react-router", async (importOriginal) => {
 	};
 });
 
-vi.mock("./TitlebarNav", () => ({
+vi.mock("./TitlebarNav", async (importOriginal) => ({
+	...(await importOriginal<typeof import("./TitlebarNav")>()),
 	useCanGoForward: () => compactRailCanGoForward.current,
 }));
 
@@ -241,6 +243,7 @@ function renderSidebar({
 	autoCompact = false,
 	topbarOffset = "toolbar",
 	expandedProjectIds,
+	includeTitlebarNav = false,
 }: {
 	onCloneProject?: CloneProjectHandler;
 	onCreateProject?: CreateProjectHandler;
@@ -252,6 +255,7 @@ function renderSidebar({
 	autoCompact?: boolean;
 	topbarOffset?: "toolbar" | "titlebar" | "trafficLights" | "session";
 	expandedProjectIds?: string[];
+	includeTitlebarNav?: boolean;
 } = {}) {
 	// Most legacy sidebar tests exercise session rows and assume their fixture
 	// project was previously open. Tests for the empty-store behavior opt out.
@@ -290,6 +294,7 @@ function renderSidebar({
 					onRemoveProject={onRemoveProject}
 					workspaces={workspaces}
 				/>
+				{includeTitlebarNav ? <TitlebarNav /> : null}
 			</SidebarProvider>
 		</QueryClientProvider>,
 	);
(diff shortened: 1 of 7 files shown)


Rishet11:
https://github.com/user-attachments/assets/91f3d695-90f8-4f13-86a1-563111d8077e

Vaibhaav-Tiwari (changes requested):
Image


 > 1. The new useRef and useEffect hooks in components/ui/sidebar.tsx
  >    are placed after conditional returns. Resizing across the mobile
  >    breakpoint—or starting with a narrow window—can change the hook
  >    count and crash React. Please move all hooks above the
  >    conditional returns and add a responsive-resize test.
  >
  > 2. The temporary sidebar uses z-chrome (z-10), while the Archive
  >    panel uses z-20. As a result, the Archive bar renders over the
  >    sidebar preview. Please ensure the temporary sidebar is above
  >    board overlays while remaining below dialogs/titlebar elements,
  >    and add coverage for this case.
  >
  > 3. Adaptive/icon-mode preview still snaps between the compact rail
  >    and expanded sidebar. Only off-canvas mode receives the new
  >    opacity/transform transition. Please add a proper reduced-motion-
  >    safe open/close transition for icon mode.
  >

ronishrohan (changes requested):
feature already existed in the past, was removed due to certain reasons, getting in touch with you for details

Annieeeee11:
> feature already existed in the past, was removed due to certain reasons, getting in touch with you for details

Yes also we really don't need it

Issue #4668, opened 2026-08-30
[Bug][Cloud][P1] Failed terminal attach retries forever: ~2 ticket mints/sec, no surfaced error, sandbox never idle-pauses


Pritom14:
## What

When the terminal WebSocket cannot attach (CSP block #4666, network/proxy filtering, or any client-side failure), the app retries indefinitely with **no backoff and no user-facing error**: the pane shows "Connecting…" forever while the mux mints a fresh terminal ticket for every attempt (~1–2/sec observed). Two costly side effects:

1. **The sandbox never idle-pauses.** `IssueTerminalTicket` (`cloud/internal/postgres/worker_transport_store.go:324`) deliberately refreshes the sandbox's `interactive_until` lease on every mint — correct for a real viewer, but a broken client turns it into an infinite keep-awake, billing compute indefinitely.
2. **The failure is invisible.** Tickets keep returning 201, so nothing distinguishes "worker still provisioning" from "attach permanently failing client-side".

## Evidence (staging, 2026-08-29/30)

- One dev's 4 sessions: 2,600+ tickets minted / 0 consumed within 45 min (`ao_access_tickets`), sandboxes held awake the whole time; ~840 mints per 10-minute window in the CP access log.
- The same signature ran overnight on 2 sessions, with sandboxes recreated repeatedly (3 provider boxes each) while nothing ever attached.

## Proposed fix

Renderer (`frontend/src/renderer/lib/cloud-terminal-mux.ts` + `useTerminalSession` reattach loop):
- Exponential backoff on attach failures (e.g. 1s → 30s cap), resetting on success.
- After N consecutive *mint-succeeded-but-socket-failed* cycles, surface a distinct terminal state: "Ticket issued but the WebSocket cannot connect — check proxy/firewall/CSP" instead of generic "Connecting…".

Control plane (defense in depth):
- Consider not refreshing `interactive_until` on every mint when the previous N tickets for the session were never consumed, so a broken client cannot hold a sandbox awake forever.

Issue #4667, opened 2026-08-30
[Bug][Cloud][P1] Cloud terminal WebSockets are severed every ~60s by the ALB idle timeout


Pritom14:
## What

Every healthy cloud terminal WebSocket is cut by the staging ALB after exactly ~60 seconds of frame-level idle. The renderer's mux silently reconnects with a fresh ticket and replays, so the terminal *looks* fine, but every idle terminal cycles disconnect/reattach once a minute — visible as a hiccup on slow links, extra ticket/replay load, and a misleading error in the CP logs.

## Evidence (staging, 2026-08-30)

- `GET /api/cloud/v1/terminal` completions with status 101 and durations of **59998–60002 ms**, one per minute per attached terminal, for an actively used session.
- Matching CP warnings once per minute per terminal: `terminal stream ended unexpectedly | failed to get reader: failed to read frame header: EOF` (`cloud/internal/httpapi/terminal_handlers.go:159`), same `terminal_id` each time.

## Cause

The ALB's default idle timeout is 60s and applies per-connection to WebSockets. Neither side sends WebSocket ping/pong keepalives: `connectTerminal` (`cloud/internal/httpapi/terminal_handlers.go:72`) reads/writes only application frames, and the renderer mux (`frontend/src/renderer/lib/cloud-terminal-mux.ts`) writes only on user input. An idle PTY (no output, no input) therefore exchanges zero frames for 60s and gets chopped.

## Fix options

1. Server-side ping every ~20–30s in the terminal WebSocket loop (`websocket.Conn.Ping`) — smallest change, covers all clients including future web/mobile.
2. Additionally or alternatively raise the ALB idle timeout for staging/production (deploy scripts under `cloud/scripts/`).

Option 1 is the right permanent fix; 2 is belt-and-braces.

PR #4669, opened 2026-08-30
fix(frontend): allow cloud control-plane websockets in the packaged CSP

### LINKED ISSUE #4666: [Cloud][P0] Packaged renderer CSP blocks every cloud terminal WebSocket
## What

In any **packaged/built** desktop app, a cloud session's terminal never connects: the renderer's Content-Security-Policy refuses the direct `wss://` connection to the control plane, so the terminal shows "Connecting…" forever. Dev-mode (`npm run dev`) is unaffected, which is why this went unnoticed.

## Root cause

`frontend/vite.renderer.config.ts` injects a CSP meta tag with `apply: "build"` (built renderers only). Its `connect-src` is:

```
connect-src 'self' http://127.0.0.1:* ws://127.0.0.1:* <posthog>
```

The cloud terminal deliberately dials `wss://<control-plane>/api/cloud/v1/terminal?ticket=…` **directly from the renderer** (single-use ticket auth; the WorkOS token stays in the Electron main process — see `frontend/src/renderer/lib/cloud-terminal-mux.ts`). That origin is not in `connect-src`, so Chromium refuses the socket before any packet leaves the machine. Every other CP call rides the main-process proxy and is unaffected — including terminal-**ticket** minting, which keeps succeeding while the attach never happens.

## Evidence (staging, 2026-08-30)

- A dev on a packaged build: 4 sessions (claude-code **and** codex), **2,600+ tickets minted, 0 consumed** (`ao_access_tickets`), zero `GET /api/cloud/v1/terminal` requests reaching the CP.
- Same app via `npm run dev` on another machine: 18 tickets minted, **18 consumed**, terminal works.
- In the packaged app's DevTools, `new WebSocket("wss://echo.websocket.org")` fails instantly with no Network-tab row (classic CSP refusal); the same snippet succeeds in a dev-mode build.

## Fix

Add the cloud control-plane WebSocket origins to `connect-src` in the built CSP (baked defaults `wss://staging-api.aoagents.dev` / `wss://api.aoagents.dev`, extendable via the existing `AO_CLOUD_CONTROL_PLANE_URL` override at build time). PR incoming.

## Follow-up worth considering

Tunneling the terminal WebSocket through the Electron main process (like every other CP call) would remove the renderer-egress requirement entirely, at the cost of a bigger change to the mux.

Pritom14:
## Summary

Fixes #4666 — in every **packaged** build, the renderer CSP (`connect-src 'self' http://127.0.0.1:* ws://127.0.0.1:* <posthog>`) refused the cloud terminal's direct `wss://` connection to the control plane, so cloud terminals stayed on "Connecting…" forever. Dev-mode (`npm run dev`) skips the CSP meta (`apply: "build"`), which is why the bug only reproduced on packaged builds.

This adds the cloud control-plane websocket origins to `connect-src`:
- baked defaults `wss://staging-api.aoagents.dev` and `wss://api.aoagents.dev`, matching the control-plane URL defaults;
- a build-time `AO_CLOUD_CONTROL_PLANE_URL` override (http(s) → ws(s) origin) so a custom control plane keeps a working terminal in packaged builds.

Everything else about the CSP is unchanged. The terminal's direct-dial design is intentional (single-use ticket is the socket's authorization; the WorkOS token never reaches the renderer — `lib/cloud-terminal-mux.ts`), so the CSP must allow the origin rather than the socket moving behind the proxy.

## Evidence

Staging, 2026-08-30: a dev on a packaged build minted **2,600+ terminal tickets with 0 consumed** across 4 sessions (claude-code and codex) — zero `GET /api/cloud/v1/terminal` requests ever reached the control plane; `new WebSocket("wss://echo.websocket.org")` failed instantly in the packaged app's console with no Network-tab row (CSP refusal), while the same dev-mode build on another machine redeemed 18/18 tickets and connected fine.

## Verification

- `cd frontend && npm run typecheck` — clean.
- `npx vite build --config vite.renderer.config.ts` — emitted `index.html` CSP now reads `connect-src 'self' http://127.0.0.1:* ws://127.0.0.1:* https://us.i.posthog.com https://us-assets.i.posthog.com wss://staging-api.aoagents.dev wss://api.aoagents.dev`.
- Follow-up issues from the same investigation: #4667 (ALB 60s idle timeout severs terminal websockets — needs server ping/pong), #4668 (attach-failure retry loop mints tickets ~2/s with no surfaced error and keeps sandboxes awake).

### DIFF
--- a/frontend/vite.renderer.config.ts
+++ b/frontend/vite.renderer.config.ts
@@ -42,8 +42,32 @@ const POSTHOG_ORIGINS = (() => {
 	return origins;
 })();
 
+// Cloud terminals attach over a ticketed wss:// dialed directly from the
+// renderer (the WorkOS token stays in the main process; the single-use ticket
+// is the socket's whole authorization — see lib/cloud-terminal-mux.ts), so the
+// packaged CSP must allow the control-plane origin or every cloud terminal
+// stays on "Connecting…" forever in built apps. The runtime URL is a daemon
+// setting the build cannot read; list the baked defaults and fold in a
+// developer's AO_CLOUD_CONTROL_PLANE_URL override so a custom control plane
+// keeps a working terminal in packaged builds too.
+const CLOUD_CP_WS_ORIGINS = (() => {
+	const origins = ["wss://staging-api.aoagents.dev", "wss://api.aoagents.dev"];
+	const override = process.env.AO_CLOUD_CONTROL_PLANE_URL?.trim();
+	if (!override) return origins;
+	let url: URL;
+	try {
+		url = new URL(override);
+	} catch {
+		return origins;
+	}
+	const origin = `${url.protocol === "http:" ? "ws:" : "wss:"}//${url.host}`;
+	if (!origins.includes(origin)) origins.push(origin);
+	return origins;
+})();
+
 // CSP for the built renderer. The daemon is loopback-only, so network access is
-// pinned to 127.0.0.1 (REST + SSE over http, terminal mux over ws). Injected at
+// pinned to 127.0.0.1 (REST + SSE over http, terminal mux over ws), plus the
+// cloud control-plane websocket origins above. Injected at
 // build time rather than written into index.html because the dev server needs
 // inline scripts (react-refresh preamble) that a static meta tag would block.
 const CONTENT_SECURITY_POLICY = [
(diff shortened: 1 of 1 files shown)


Pritom14:
Added a second commit per review of the root cause: the CSP is now **enforced in dev mode too** (same directives, relaxed only for what the dev server itself needs — `'unsafe-inline'` script-src for the react-refresh preamble, `ws://localhost:*` for HMR). Dev/packaged parity means a missing `connect-src` origin now fails on the developer's screen immediately instead of only in packaged builds — the exact skew that let #4666 ship unnoticed.

Verified: built HTML keeps the strict policy (`script-src 'self'`, no localhost ws, cloud wss present); dev-server HTML carries the serve variant with the preamble executing; the Electron dev app boots under the enforced policy with renderer API traffic flowing and zero CSP violations.

AgentWrapper (approved):

PR #4457, opened 2026-08-25
fix(chat): require consent before interrupting busy work

illegalcall:
Closes #4423

## Summary

- require an explicit **Finish work** versus **Stop now** choice before a busy or uncertain Chat session switches to Terminal UI
- keep ordinary sends plus Claude Edit/Retry work scoped to the initiating session from pending request through exact durable-turn observation
- preserve accepted and in-flight work across navigation, same-route rerenders, and full Chat-surface unmount/remount
- scope transition start/cancel/ack mutations, pending state, errors, and cache invalidation to their initiating session, including overlapping mutations in multiple sessions
- prevent a deferred transition for session A from closing a newer consent dialog for session B
- suppress rapid duplicate composer submissions without hiding genuine transport failures
- retain the safe direct-drain path for confirmed-idle work

## Root cause and final review fix

Chat-to-Terminal previously submitted destructive `interrupt` immediately, before asking the user whether active and queued work could be cancelled.

During final independent review, a second race was found: each TanStack `MutationObserver` remembered only its most recent mutation. If session A started a transition, session B then started another, and the user returned to A, A could appear idle while its request was still pending. The final commit keys start/cancel/ack mutations and aggregates pending/error state by target session. Its public-hook regression was proven red before the fix and green afterward.

## Regression coverage

MQA-05 and focused adversarial tests cover:

- busy Chat showing the policy dialog before any transition request
- **Finish work** waiting for running work before switching
- **Stop now** interrupting only after explicit consent
- pending and accepted Send/Edit/Retry work remaining busy through stale idle snapshots
- accepted work remaining busy through full surface remount
- in-flight requests remaining busy through a board round-trip
- session navigation not transferring consent, pending state, errors, or cache invalidation
- overlapping A/B transition mutations remaining independently pending
- a deferred A completion not closing B's newer policy dialog
- rapid double-submit invoking transport once

## Exact validation

- product head: `ce9183ea87d18273efed6c6ac3b776fcab065927`
- base: `be21597e7a24cbee2e6e9103ae669ff18c3b56b1`
- opt-in harness head: `cc760a23ca74e8b110c12f11faa337064ca8d106` in draft PR #4419
- exact product+harness integration: `72c9e0f474cce9b709450a9fd7e28caefd2a6d02`
- focused renderer tests: **222/222 passed**
- full renderer: **3,013 passed, 1 intentional skip**
- frontend typecheck: **passed**
- E2E typecheck: **passed**
- strict `MQA-05`: **2/2 passed**
[...]

PR #4670, opened 2026-08-30
fix: fixed stuck composer spinner from queue dock undefined compare

Annieeeee11:
## Summary

- Fix the send button stuck on the loading spinner after [#4649](https://github.com/Untrivial-ai/agent-orchestrator/pull/4649): `editQueuedTurnPendingTurnId === queueEdit?.turnId` evaluated to `true` when both were `undefined`, so `savingQueuedEditPending` was permanently true and the composer could not send.
- Harden dispatch/`busy` handling introduced in [#4457](https://github.com/Untrivial-ai/agent-orchestrator/pull/4457): only treat `"pending"` dispatch as busy, release the sentinel immediately for queued sends, and refresh conversation snapshots in the background instead of awaiting refetch in mutation `onSuccess` (steer, send, queue ops).
- Surface `sendPending` on the composer send control and show a clear error when a send is blocked by an in-flight message.

## Root cause

| PR | Regression |
|---|---|
| **#4649** | `savingQueuedEditPending={editQueuedTurnPendingTurnId === queueEdit?.turnId}` → `undefined === undefined` → always loading, always blocked |
| **#4457 + #4649** | Accepted dispatch + awaited refetch kept `busy` / mutation pending after the daemon had already accepted work |

## Test plan

- [ ] Open a chat session on a build with this fix — send button shows arrow (not spinner) on idle empty composer
- [ ] Type and send a message while the agent is idle — message sends normally
- [ ] While the agent is working, queue two messages back-to-back — second send is not swallowed
- [ ] Edit a queued message from the dock — spinner only while save is in flight
- [ ] Steer a running turn (⌘⏎) — spinner clears after daemon accepts, does not hang on slow refetch
- [x] `cd frontend && npm test -- useConversation.test.tsx ChatComposer.test.tsx`

### DIFF
--- a/frontend/src/renderer/components/chat/ChatComposer.test.tsx
+++ b/frontend/src/renderer/components/chat/ChatComposer.test.tsx
@@ -905,6 +905,20 @@ describe("unavailable states", () => {
 		expect(onSend).not.toHaveBeenCalled();
 	});
 
+	it("does not show a loading spinner when no queued edit is saving", () => {
+		renderComposer();
+		expect(
+			screen.getByRole("button", { name: "Send message" }).querySelector(".animate-spin"),
+		).not.toBeInTheDocument();
+	});
+
+	it("shows a loading spinner only while the queued edit being edited is saving", () => {
+		renderComposer({ editingQueuedTurnId: "turn-1", savingQueuedEditPending: true });
+		expect(
+			screen.getByRole("button", { name: "Send message" }).querySelector(".animate-spin"),
+		).toBeInTheDocument();
+	});
+
 	it("says a mid-turn message will be held", () => {
 		const { field } = renderComposer({ willQueue: true });
 		expect(field).toHaveAttribute(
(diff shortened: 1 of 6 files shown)


ronishrohan (commented):
P2: accepted dispatch tracking can be overwritten before its turn is observed. `claimConversationDispatch` now allows a new operation when the existing entry is `state: "accepted"`, but the tracker stores only one entry per session. If send A is accepted and its snapshot is still stale, send B overwrites A; when B completes/releases, A’s pending handoff marker is lost. This can allow an interface switch before A appears in the snapshot. Please consider tracking multiple accepted dispatches or preserving the accepted marker while allowing queued sends separately. The existing tests cover queued follow-ups and accepted work independently, but not this overlap.

ronishrohan (approved):
Reviewed. The accepted-dispatch tracking race is documented above; approving for merge as requested.

PR #4674, opened 2026-08-30
Keep OpenCode in sync with the app theme

codebanditssss:
## Summary
- Forward validated xterm foreground, background, and ANSI palette responses so adaptive TUIs can resolve their startup theme
- Implement DEC color-scheme reporting so a running OpenCode TUI follows AO light and dark theme changes without restarting
- Keep generic xterm control data isolated from agent input and cover the response path with regression tests

## Tested flows
- OpenCode starts with the light palette when AO is in Light mode
- OpenCode starts with the dark palette when AO is in Dark mode
- Running OpenCode switches from Dark to Light without restarting
- Running OpenCode switches from Light to Dark without restarting
- Task card opens, a shell terminal can be created, and OpenCode launches inside it

## Verification
- `npx vitest run` — 233 files, 3,156 tests passed
- `npm run typecheck`
- `npm run typecheck:e2e`


frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -115,6 +115,10 @@ function loadRenderer(term: Terminal): void {
 const SUPPRESS_NATIVE_PASTE_MS = 100;
 /** Long enough to notice, short enough that a second copy reads as a second copy. */
 const COPY_TOAST_MS = 1400;
+const XTERM_COLOR_REPORT =
+	/^\x1b\](?:4;(?:[0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])|1[0-2]);rgb:[0-9a-f]{4}\/[0-9a-f]{4}\/[0-9a-f]{4}\x1b\\$/i;
```
Annieeeee11: This only matches a complete, ST-terminated report occupying the entire `onData` chunk. That is too strict for the path OpenTUI actually uses:

- Queries are BEL-terminated (`OSC 10;?\x07` / `OSC 11;?\x07`). xterm.js happens to reply with ST today, but BEL replies (`\x07`) are dropped.
- `^...$` fails if xterm concatenates OSC 10 + OSC 11 in one `onData`, or splits a report across chunks.

Current OpenTUI uses the 997 notification only as a refresh trigger, then re-queries OSC 10/11 and derives light/dark from luminance. If those replies never reach the PTY, the live switch this PR tested will be flaky even when 2031 reports are correct.

Please buffer `onData` and extract complete OSC 4/10/11/12 reports (ST or BEL). Keep rejecting everything else — the original “do not forward raw `onData`” rule still stands.

frontend/src/renderer/components/XtermTerminal.tsx
```diff
@@ -603,6 +609,54 @@ export function XtermTerminal(props: XtermTerminalProps) {
...
+			const changed = theme !== currentColorScheme;
+			currentColorScheme = theme;
+			if (!force && (!colorSchemeUpdatesEnabled || !changed)) return;
+			emitTerminalResponse(`\x1b[?997;${theme === "dark" ? 1 : 2}n`);
+		};
+		colorSchemeReporterRef.current = reportColorScheme;
+		const setColorSchemeUpdates = term.parser.registerCsiHandler(
+			{ prefix: "?", final: "h" },
+			(params) => {
+				if (params.length !== 1 || params[0] !== COLOR_SCHEME_UPDATE_MODE) return false;
```
Annieeeee11: `params.length !== 1` means `CSI ? 2031;25h` (or 2031 packed with any other private mode) never enables updates, and the same pattern on the `l` / `$p` / `n` handlers has the same hole.

OpenTUI currently sends a lone `\x1b[?2031h`, so this works for the happy path. Scan params for `2031` / `996` and still `return false` when other modes are present so xterm keeps handling cursor, paste, and mouse. Same change on the reset / DECRQM / DSR handlers below.

frontend/src/renderer/hooks/useTerminalSession.ts
```diff
@@ -664,6 +665,10 @@ export function useTerminalSession(session: WorkspaceSession | undefined, option
 			else revealReplayTail();
 			mux.sendInput(handle, data);
 		});
+		const terminalResponse = terminal.onTerminalResponse((data) => {
+			if (!isCurrentAttachment(generation, handle, mux) || !r.inputReady) return;
+			mux.sendInput(handle, data);
```
Annieeeee11: User input is gated on `inputReady`, `inputDisabled`, and `isVisible`. Protocol replies only check `inputReady`.

That is the right policy (a hidden OpenCode still needs 2031/OSC replies), but it is easy to “fix” later. Please add a one-line comment that these replies must reach the PTY even when the pane is hidden or input is disabled.

Worth a test for `isVisible: false` after attach as well — the new session test covers `inputDisabled` but not the hidden-pane case.

frontend/src/renderer/components/XtermTerminal.test.tsx
```diff
@@ -1148,11 +1170,78 @@ describe("XtermTerminal", () => {
 		const onInput = vi.fn();
 		render(<XtermTerminal theme="dark" onReady={(terminal) => terminal.onUserInput(onInput)} />);
 
-		expect(state.lastTerminal!.dataListeners.size).toBe(0);
 		state.lastTerminal!.dataListeners.forEach((listener) => listener("\x1b[A"));
 		expect(onInput).not.toHaveBeenCalled();
 	});
 
+	it("routes xterm color reports without forwarding raw input", () => {
```
Annieeeee11: This only feeds complete ST-terminated reports, one per `onData` call. Please also cover:

- BEL-terminated OSC 10/11 (`\x1b]11;rgb:ffff/ffff/ffff\x07`)
- two reports in one chunk (`OSC 10` + `OSC 11`)
- a split report across two `onData` calls
- `registerCsiHandler({ prefix: "?", final: "h" })` returning `false` for `?25h` so cursor show/hide is not swallowed

Annieeeee11 (changes requested):
The 2031 / 996 / 997 mapping is right, and keeping generic `onData` off the PTY is the correct constraint. Two things will pass these unit tests and still fail a real OpenTUI session: the OSC filter only accepts a complete ST-terminated chunk, and DECSET 2031 is ignored unless it is the only private mode in the sequence.

OpenTUI currently treats `CSI ? 997 ; Ps n` as a refresh trigger and then re-queries OSC 10/11 (`\x1b]10;?\x07\x1b]11;?\x07`) to decide light vs dark from luminance. If those replies are dropped, live theme follow breaks even when the 997 reports themselves are correct.

frontend/src/renderer/components/XtermTerminal.tsx
codebanditssss: Fixed in 6830f477e: the shared bounded forwarder now accepts only strict OSC 4/10/11/12 color reports, supports BEL and ST terminators plus combined and split chunks, and rejects malformed or unrelated terminal data.

frontend/src/renderer/components/XtermTerminal.tsx
codebanditssss: Fixed in 9f983568c: the h, l, $p, and n handlers scan top-level CSI params, apply the AO 2031/996 side effect, and return false whenever unrelated params are present so xterm can process them.

frontend/src/renderer/hooks/useTerminalSession.ts
codebanditssss: Fixed in 8918196e5: the protocol path is explicitly documented to bypass human-input visibility and ownership gates, and the hidden plus input-disabled test verifies exact protocol delivery while typing remains blocked.

frontend/src/renderer/components/XtermTerminal.test.tsx
codebanditssss: Fixed across 83013e5bc, 9f983568c, and 6830f477e: tests cover BEL termination, concatenated reports, split chunks, unrelated ?25h handling, OSC 4 palette boundaries, and oversized-buffer recovery.

Annieeeee11 (approved):
lgtm

PR #4682, opened 2026-08-30
fix: keep terminal colors in sync with AO theme

Annieeeee11:
## Summary

Cursor Agent can render its prompt bar with the wrong colors when AO switches between light and dark mode. This PR synchronizes the resolved AO theme with Cursor sessions at startup and while they are running.

## What changed

- Persist the resolved light/dark terminal scheme under AO's data directory.
- Inject `TERM_THEME` and `COLORFGBG` only into Cursor Agent processes.
- Answer Cursor's private color-scheme protocol and OSC 10/11/12 color probes through a dedicated protocol-input path.
- Scope protocol handling to Cursor terminals so shells, Codex, and other TUIs receive no unsolicited control bytes.
- Bound OSC response buffering and avoid duplicate OSC replies.
- Keep terminal theme-hint writes ordered and atomic.
- Prevent theme transitions from flashing terminal/composer chrome.
- Preserve the Cmd+K search query until the palette close animation completes, including reduced-motion environments.

## Root cause

Cursor probes terminal appearance through OSC and private CSI sequences. The renderer previously allowed those replies to be lost or treated like ordinary input, while new theme changes could leave Cursor's prompt bar using stale colors.

## Verification

- `go test ./internal/termtheme ./internal/adapters/agent/cursor`
- `vitest run src/renderer/lib/osc-color-report.test.ts src/renderer/lib/cursor-color-scheme.test.ts`
- `npm --prefix frontend test -- src/renderer/components/XtermTerminal.test.tsx src/renderer/components/CommandPalette.test.tsx` — 120 tests passed
- `git diff --check`

The full frontend typecheck is currently blocked by the shared `packages/product-ui` workspace because its React/motion dependencies are unavailable in this checkout.

## Manual test plan

- [ ] Start a Cursor session in light mode; prompt bar is light.
- [ ] Start a Cursor session in dark mode; prompt bar is dark.
- [ ] Toggle the theme with an existing Cursor session open; colors update without restarting.
- [ ] Open a shell or non-Cursor terminal; no unsolicited control bytes or input corruption.
- [ ] Search in Cmd+K, select Toggle theme with Enter, and confirm the query does not flash empty while the palette closes.

### DIFF
--- a/backend/internal/adapters/agent/cursor/cursor.go
+++ b/backend/internal/adapters/agent/cursor/cursor.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/binaryutil"
 	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/agent/hookutil"
 	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+	"github.com/aoagents/agent-orchestrator/backend/internal/termtheme"
 	"github.com/aoagents/agent-orchestrator/backend/pkg/agentruntime"
 )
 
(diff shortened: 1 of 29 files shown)


ronishrohan (approved):
Verified the Cursor terminal theme synchronization, protocol scoping, retained-terminal handling, and Cmd+K regression fix.

PR #4683, opened 2026-08-30
chatui to tui switch for cloud sessions

### LINKED ISSUE #4577: [Cloud][P0] Chat mode for cloud sessions
## What
Support the chat session interface (not just the agent TUI) for cloud sessions, matching the local app's tui vs chat modes.

## Why
Local sessions can run in chat mode (AO owns a structured provider controller) or tui. Cloud currently only runs the agent's TUI in a terminal. Without chat mode, cloud is missing a first-class local interface and any UX built on chat.

## Current state (pointers)
- Local chat driver: `backend/internal/service/chat` (service, controller, steer, history).
- Cloud runs the harness process via `cloud/internal/workerexec`.
- Session kind and mode plumbing exists in both.

## Scope
- Port or adapt the chat driver so a cloud session created in chat mode is driven structurally, with history and steering, over the control plane.
- Decide the v1 boundary: full chat parity vs explicitly scoping cloud to tui for the first release.

## Acceptance
A cloud session created in chat mode behaves like a local chat session (structured turns, history), or the limitation is a deliberate, documented v1 scope.

Priority: P0.

vanshx999:
Code changes: +4611 -255
Tests: +2410 -3
Others: +215 -143

## Summary

- Move the Cloud interface-transition implementation into the monorepo `cloud/` control plane.
- Add durable PostgreSQL transition state, reconciliation, worker transport, and TUI/Chat lifecycle wiring.
- Expose Cloud `GET`/`POST`/`DELETE /api/cloud/v1/orgs/{orgId}/sessions/{sessionId}/interface-transition` routes.
- Align the public wire contract with `interfaceMode` and `tui | chat`.
- Regenerate `packages/cloud-client` and update Cloud development documentation to use `cloud/`.

## Validation

- `cd cloud && go test ./...`
- `npm --prefix packages/cloud-client run generate`
- `npm --prefix packages/cloud-client run typecheck`
- `git diff --check`

## Deployment/testing

This branch is based on `main` and is intended to be deployed with `cloud/scripts/deploy-staging.sh` after review. The implementation includes the Chat controller handoff path; staging should be checked for both TUI → Chat and Chat → TUI transitions after deployment.

Fixes #4577


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

vanshx999:
This PR fixes #4577.

It adds the Cloud control-plane support and Electron frontend integration for TUI ↔ ChatUI interface transitions, preserving the shared conversation history across both interfaces.

vanshx999:
## Verification update — harness parity audit (head d4d06f7f1)

### Audit findings (Codex / Claude Code / Cursor)

**Bug found and fixed — Claude Code headless was broken.** `claude --print --output-format stream-json` hard-errors without `--verbose` (`Error: When using --print, --output-format=stream-json requires --verbose`). The Chat controller for Claude Code could never produce a reply. Fixed in `cloud/internal/workerexec/command.go`.

**Bug found and fixed — Claude/Cursor output was never projected.** `chatOutputProjector` only understood Codex JSONL; Claude and Cursor stdout (their own `stream-json` NDJSON) was published raw to the chat stream and their `session_id` was never captured, so Chat→TUI restore had no native identity for those harnesses. Both now have streaming-output adapters:
- identity captured from `system/init` (`session_id`) for Claude/Cursor, `thread.started` for Codex — first announcement wins so a resumed thread cannot flap
- the terminal `result` event is projected as the single assistant reply; its text is the concatenation of the turn's assistant messages, so streaming intermediates are deliberately not double-projected
- error results (`is_error: true`) are not published as replies
- Codex `Flush()` now recovers a `thread.started` stranded in an unterminated final line (identity could previously be lost on clean process exit)

### Verified per requirement
- **Launch commands / credentials / resume IDs audited**: Codex (`exec --json` + `resume <id>`, API key/access token via `codex login`), Claude (`--print --output-format stream-json --verbose` + `--resume <id>`, api_key/oauth_token env), Cursor (`agent --print --output-format stream-json` + `--resume <id>`, `CURSOR_API_KEY`). Fresh-session guard verified for all three: no `--resume`/`resume` on a fresh launch; Claude additionally requires the JSONL transcript to exist on disk before restoring (headless and interactive).
- **Native conversation identity**: published via `session-start` activity from headless turns (all harnesses) and Codex TUI rollout projection; worker prefers the control-plane id over the bootstrap value; TUI rebuild uses each harness's native resume syntax.
- **Failed transitions/retries**: existing coordinator coverage plus new tests (target-start failure → recovery_required, pending worker command → retry exhaustion fails closed pre-commit, preflight rejection for unsupported harness).
- **Cloud feature flag**: no change — switching remains gated by the existing cloud offering gate (`useCloudGate`) and Cloud-only transition API.

### Tests run
- `cloud`: `go build ./... && go vet ./... && go test ./...` — all pass; `go test -race` on workerexec/workertransport/interfacereconcile/worker/cmd — pass; gofmt clean. New/updated tests:
  - projector: per-harness projection + identity, error-result suppression, first-identity-wins, flush recovery, passthrough for unknown harness/stderr
  - command: claude/cursor headless args + resume, fresh-turn no-resume (claude/codex), stale-Claude-resume dropped headlessly, credential mismatch fail-closed
  - supervisor: execute-level per-harness reply+identity, runner failure → FailTurn, builder rejection → FailTurn, busy flag across a turn, OSRunner process-level integration for all three harnesses
  - coordinator: both directions (TUI↔Chat) × all three harnesses converge and pass the native id to StartTarget; TransportDriver preflight accepts the three harnesses and rejects unknown
  - workertransport: Chat→TUI factory rebuilds resume argv per harness; fresh interactive argv carries no resume
- `frontend`: focused vitest — useSessionInterfaceTransition, SessionView, SessionChatSurface, SessionInterfaceSwitch, ShellTopbar, event-transport: **210 tests pass**. Full `typecheck` still blocked by the pre-existing missing `packages/product-ui` deps (react/motion/clsx/tailwind-merge), unchanged by this PR.

### Known limitation (not claimed done)
Claude Code and Cursor responses/history are verified at the protocol level (fixtures match vendor docs' stream-json schemas) but **not yet end-to-end against real Cloud sandboxes** — that requires real provider credentials in the sandbox environment. Per the task guardrails I am not claiming full Claude/Cursor support until that pass runs; the local compose stack cannot substitute for it without credentials.

vanshx999:
## Fix: browser proxy HTTP 422 (your friend's ChatUI→browser error)

### Root cause
The Cloud browser proxy dispatches `kind='browser.fetch'` into `ao_worker_requests`, but **no migration ever added that kind to the `ao_worker_requests_kind_check` constraint**. The error chain:

`INSERT kind='browser.fetch'` → Postgres check violation (`23514`) → `normalizeConstraintError` → `ErrInvalid` → `writeStoreError` → **HTTP 422 `validation_error`** (`cloud/internal/httpapi/transport.go:73`).

Two contributing facts:
- A previous commit (`4d4f08300`, by Pritom14) fixed this with migration `00032` — but that commit was **lost during the PR conflict resolution**, so fresh databases never got the kind.
- Migration `00035` (interface transitions) rewrites the constraint wholesale and re-omitted `browser.fetch`, so even a schema that had it patched lost it again.

**When it occurs:** every browser-proxy request (Browser tab page loads, iframe fetch/XHR) on any database migrated purely from migrations — fresh Docker E2E, the friend's test env, new sandboxes. Staging did not hit it because it carries a hand-patched schema (documented in the original `4d4f08300` commit message).

### Fix (pushed, head d5e5f3e17)
- **Restored `00032_browser_fetch_worker_requests.sql`** — adds `browser.fetch` to the allowlist. Its Up list also carries the `interface.*` kinds so environments that already applied old `00035` re-run only `00032` without breaking in-flight interface handoffs.
- **`00035` Up + Down lists now include `browser.fetch`** so the final schema state keeps it.
- **Regression test** `TestWorkerRequestKindConstraintCoversDispatchedKinds` parses the migrations and asserts the highest-numbered constraint rewrite allows every kind the code dispatches (would have caught this class of bug).

### Verification
- Ran the full migration chain `00001..00036` against a scratch Postgres: applies cleanly.
- Confirmed the final constraint contains all 14 kinds and that `browser.fetch` and `interface.start` INSERTs succeed against the migrated schema.
- `go test ./...` and `go vet ./...` pass in `cloud/`.

Your friend should re-run migrations (a fresh DB or `ao-cloud-migrate` on their existing env will pick up `00032`), and the browser proxy will stop 422'ing. If they already ran old `00035`, applying `00032` is sufficient — it re-adds `browser.fetch` while preserving the `interface.*` kinds.

vanshx999:
## Follow-up: 00032 was not applying on existing envs (goose out-of-order guard) — fixed

Your friend was right: on any environment already migrated to v36, goose **aborts the whole run** with `found 1 missing migrations before current version 36: version 32: ...` and applies nothing — so 00032 never landed there and browser.fetch kept 422'ing. Fresh DBs were fine because they migrate the full chain in order.

### Root cause of the second failure
`postgres.Migrate` called `goose.UpContext` without `WithAllowMissing`. goose v3 treats a version gap below the current version as a hard error (protection against accidental out-of-order migrations), which is exactly what a restored-below-tip migration like 00032 looks like.

### Fix (pushed, head 94b90ee36)
One change in `cloud/internal/postgres/migrate.go`: pass `goose.WithAllowMissing()`. Both entry points (`ao-cloud-migrate` and control-plane startup) share this function, so both pick it up. This is safe here because 00032's Up is a pure constraint rewrite with a superset allowlist — applying it out of order only adds allowed kinds.

### End-to-end proof on scratch Postgres (all with real binaries)
1. Built an env at v36 **without** 00032 and with the old 00035 (the friend's exact state) → `browser.fetch` INSERT fails with the check-constraint violation (the DB-level cause of the HTTP 422).
2. Ran the **old** runner against it → aborts with the exact reported error, applies nothing (v32 stays unapplied).
3. Ran the **fixed** runner against it → applies `00032_browser_fetch_worker_requests.sql`, migrates cleanly.
4. On that same healed DB, `browser.fetch`, `interface.start`, and `terminal.open` INSERTs **all succeed** — the handoff kinds were not disturbed.

`go build ./... && go vet ./... && go test ./...` green in `cloud/`.

**Action for your friend:** pull the branch, rebuild/restart, run `ao-cloud-migrate` (or just restart the control plane — startup migrates too). 00032 will apply automatically and the 422s stop, with no schema wipe.

vanshx999:
@illegalcall Thanks — confirmed this is a separate **native desktop / Claude Code** model-persistence gap, not a Cloud handoff regression.

`SetConfigOption` successfully updates the live Claude conversation through `PATCH /api/v1/sessions/{id}/conversation/config-options/model`, but it does not persist the selected model onto the durable session metadata. The TUI rebuild therefore resumes the existing native conversation with the old persisted model.

The fix should invoke the same durable model-persistence callback after a successful `config-options/model` update, and add a regression test for **Claude TUI → Chat → select model → TUI** verifying that the resumed command carries the selected model. Conversation identity/history is already preserved.

cloud/internal/postgres/migrations/00042_browser_fetch_worker_requests.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
```
illegalcall, line 1: **[P1] Renumber the new migrations to unused versions.** Main already contains 00032_browser_fetch_requests.sql, 00036_coder_sandbox_provider.sql, and 00037_turns_allow_queued_fanin.sql. Goose rejects duplicate versions during source discovery, before WithAllowMissing applies, so this branch cannot run migrations or start a fresh control plane.

cloud/internal/postgres/migrations/00034_session_interface_transitions.sql
```diff
@@ -0,0 +1,121 @@
...
+    ON ao_interface_transitions(org_id, session_id, created_at DESC);
+CREATE INDEX ao_interface_transitions_deliverable_idx
+    ON ao_interface_transitions(org_id, completed_at)
+    WHERE phase IN ('completed', 'failed', 'cancelled', 'recovery_required');
+CREATE INDEX ao_interface_transitions_claim_idx
+    ON ao_interface_transitions(claimed_at)
+    WHERE claimed_by <> '';
+
+-- Messages held while neither controller is allowed to accept work.
+CREATE TABLE ao_interface_transition_messages (
```
illegalcall, line 68: **[P1] Preserve the tenant key on transition messages.** This organization-owned table has no org_id and only references transition_id. The Cloud tenancy rule requires a composite tenant FK path; add org_id and reference (org_id, transition_id) so the database itself prevents cross-tenant ownership drift.

cloud/internal/httpapi/session_interface_transition_handlers_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if response.Supported {
+		t.Fatal("terminated session must not advertise interface handoff support")
+	}
+	if response.ReasonCode != "SESSION_TERMINATED" {
+		t.Fatalf("reason code = %q, want SESSION_TERMINATED", response.ReasonCode)
+	}
+}
+
+func TestStartSessionInterfaceTransitionRejectsTerminatedSession(t *testing.T) {
```
illegalcall, line 137: **[P2] Cover the new handler boundaries.** This file only tests two terminated-session branches. The four new handlers still need direct success, malformed-input, store-error, cancel, and acknowledgement coverage required by the repository review standard.

cloud/internal/postgres/project_session_store.go
```diff
@@ -582,6 +582,9 @@ func createSessionTx(
 	if input.DeniedCommands == nil {
 		input.DeniedCommands = []string{}
 	}
+	if !input.Interface.Valid() {
```
illegalcall, line 585: **[P1] Allow chat-first session creation.** The public create request and its domain mapping never set Interface, so every external create reaches this fallback and becomes TUI. Issue #4577 explicitly requires sessions created in chat mode; create-then-switch is not that path.

cloud/internal/workertransport/interface.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+func (s *Supervisor) inspectInterface() (any, error) {
+	s.iface.mu.Lock()
+	defer s.iface.mu.Unlock()
+	// TUI work is interactive and the worker has no turn execution to drain.
+	// Chat work is headless, so it must report its actual turn activity. The old
+	// implementation returned idle only for TUI, which made every Chat -> TUI
+	// drain transition wait forever even when no prompt was running.
+	idle := true
```
illegalcall, line 94: **[P1] Do not report an active TUI as idle.** During a TUI to Chat drain, this is unconditionally true because only ChatRunner activity is inspected. The coordinator can therefore pass the drain check and close the PTY while the interactive agent is still working.

cloud/internal/workertransport/interface.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+	return interfaceInspectResult{
+		Idle:            idle,
+		WaitingForInput: false,
+	}, nil
+}
+
+func (s *Supervisor) interruptInterface(ctx context.Context) error {
+	agentTerminalID := s.agentTerminalID()
```
illegalcall, line 114: **[P1] Interrupt the active Chat controller.** This always looks up the agent terminal and sends Ctrl-C. Chat mode deliberately closes that terminal, so Chat to TUI with the stop-now policy returns agent terminal is not open instead of cancelling the headless Chat runner and its active turn.

cloud/internal/postgres/event_store.go
```diff
@@ -271,6 +365,11 @@ func appendUserMessage(
 		JOIN ao_sessions session
 			ON session.org_id = terminal.org_id AND session.id = terminal.session_id
 		WHERE terminal.org_id = $1 AND terminal.session_id = $2 AND terminal.kind = 'agent'
+		  -- A terminal may still be recorded as open while its PTY is winding
+		  -- down during a TUI -> Chat handoff. Once Chat is committed, it must
+		  -- never receive new input: that would strand a ChatUI message in the
+		  -- old terminal instead of creating a durable worker turn.
+		  AND session.interface = 'tui'
```
illegalcall, line 400: **[P1] Fence messages while a handoff is active.** Checking only the committed session interface leaves a race after drain inspection and before commit: a message can still enter the source terminal or ordinary turn queue. The new transition-message queue currently has no production caller or delivery path, so it does not close this window.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
```diff
@@ -0,0 +1,203 @@
...
+			commandError={
+				eventsQuery.error instanceof Error
+					? eventsQuery.error.message
+					: send.error instanceof Error
+						? send.error.message
+						: undefined
+			}
+			headerActions={headerActions}
+			onInterrupt={activeTurn ? () => interrupt.mutate() : undefined}
+			onSend={(text) => send.mutateAsync(text)}
```
illegalcall: **[P1] Preserve the stable clientMessageId.** ChatWorkspace passes a durable ID for uncertain-delivery recovery, but this callback drops it and the Cloud client generates a new idempotency key. Retrying after a lost response can therefore create a duplicate message and turn.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
```diff
@@ -0,0 +1,203 @@
...
+			newWorkDisabled={newWorkDisabled}
+			commandError={
+				eventsQuery.error instanceof Error
+					? eventsQuery.error.message
+					: send.error instanceof Error
+						? send.error.message
+						: undefined
+			}
+			headerActions={headerActions}
+			onInterrupt={activeTurn ? () => interrupt.mutate() : undefined}
```
illegalcall, line 208: **[P1] Implement Cloud steering.** This surface wires interrupt but omits onSteer, and there is no Cloud steering route. Input during an active turn can only queue, so the steering requirement in #4577 is not implemented.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
```diff
@@ -0,0 +1,203 @@
...
+			const turn = turns.get(turnID)!;
+			turn.state = event.type === "chat.turn_completed" ? "completed" : event.type === "chat.turn_interrupted" ? "interrupted" : "failed";
+			turn.completedAt = event.createdAt;
+			const error = eventPayload(event).error;
+			turn.errorMessage = typeof error === "string" ? error : undefined;
+		}
+		const text = eventText(event);
+		if (!text) continue;
+		if (event.type === "chat.user_message") {
+			items.push({
```
illegalcall, line 62: **[P2] Project queued prompts as structured turns.** User-message events have no turnId and no queued-turn event is emitted before chat.turn_started, so a queued prompt is only a detached transcript item until execution begins and cannot participate in the queue controls.

illegalcall (changes requested):
Requesting changes on the current head. Goose cannot load this migration set because versions 00032, 00036, and 00037 are duplicated; chat-first creation and steering required by #4577 are absent; both handoff policies, transition fencing, retry idempotency, and queued-turn projection have correctness gaps; and the transition message table violates the documented tenant-FK invariant. Please also add the mandatory PR change-count header and the missing handler boundary coverage.

cloud/internal/httpapi/session_interface_transition_handlers_test.go
vanshx999: Added direct handler coverage for successful start, malformed input, store errors, cancellation, and notice acknowledgement.

cloud/internal/postgres/project_session_store.go
vanshx999: Public session creation now accepts `interfaceMode: "chat"`, validates it, maps it into `domain.CreateSession`, and the Cloud composer creates Chat-first sessions.

cloud/internal/workertransport/interface.go
vanshx999: An open TUI PTY now fails drain inspection closed (`idle: false`, `quiescenceUnverified: true`); stop-now remains the explicit interrupt path. Added transport coverage.

cloud/internal/workertransport/interface.go
vanshx999: Chat-mode interrupt now targets the active Chat controller context rather than the closed terminal. Added cancellation coverage for an active headless turn.

cloud/internal/postgres/event_store.go
vanshx999: Message acceptance locks an active transition and persists the prompt in its transition queue with a preallocated turn ID. Activation atomically delivers all held messages into `ao_turns` before completing the transition.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
vanshx999: Cloud Chat now forwards ChatWorkspace’s durable `clientMessageId` as the idempotency key; client coverage verifies the exact key survives retries.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
vanshx999: Cloud steering is now wired end-to-end: active turns expose `onSteer`, the client posts the stable idempotency key to the steering route, and the control plane cancels then queues the replacement instruction.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
vanshx999: Accepted queued prompts now receive a turn ID in their `chat.user_message` payload, and the Cloud projection attaches that ID to the user message so queued work renders as a structured turn.

cloud/internal/postgres/project_session_store.go
```diff
@@ -672,6 +675,7 @@ func createSessionTx(
 		input.Prompt,
 		input.Mode,
 		input.DeniedCommands,
+		input.Interface,
```
illegalcall, line 678: **[P1] Enqueue the creation prompt as a Chat turn.** This value now makes Cloud tasks start in Chat, but CreateSession only stores the prompt and appends a chat.user_message event; it never creates an ao_turn. Chat startup skips the interactive agent and ChatRunner only executes rows returned by ClaimTurn, so the initial task brief appears in history but is never executed.

cloud/internal/postgres/worker_turn_store.go
```diff
@@ -188,6 +190,20 @@ func (s *Store) RequestTurnCancellation(
...
 
+// SteerTurn cancels the active turn and queues its replacement. The explicit
+// cancellation keeps a Cloud steering request from silently becoming an
+// ordinary queued prompt.
+func (s *Store) SteerTurn(
+	ctx context.Context,
+	principal domain.Principal,
+	orgID, sessionID, turnID, idempotencyKey, text, model, reasoningEffort string,
+) (domain.ClientEvent, error) {
+	if err := s.RequestTurnCancellation(ctx, principal, orgID, sessionID, turnID); err != nil {
```
illegalcall: **[P1] Steer the active turn instead of cancelling it.** Issue #4577 asks for steering matching local Chat, where guidance is injected into the acknowledged provider turn. This commits cancel_requested and then calls SendMessage in a second transaction, creating a replacement turn. An enqueue failure leaves the original cancelled with no guidance, and a later retry can fail at cancellation before replaying the idempotent send result.

cloud/internal/postgres/session_interface_transition_store.go
```diff
@@ -0,0 +1,676 @@
...
+			if err := rows.Scan(&messageID, &turnID, &sequence, &modeCap, &denied); err != nil {
+				return err
+			}
+			if _, err := tx.Exec(ctx, `INSERT INTO ao_turns (
+				id, org_id, session_id, user_message_sequence, mode_cap, denied_commands
+			) VALUES ($1, $2, $3, $4, NULLIF($5, ''), $6)
+			ON CONFLICT (id) DO NOTHING`, turnID, orgID, sessionID, sequence, modeCap, nonNilStrings(denied)); err != nil {
+				return normalizeConstraintError(err)
+			}
+			if _, err := tx.Exec(ctx, `UPDATE ao_interface_transition_messages SET delivered_at = now() WHERE id = $1`, messageID); err != nil {
```
illegalcall: **[P1] Settle held prompts on every terminal transition outcome.** This completion path is the only code that converts ao_interface_transition_messages into ao_turns. Cancellation, failure, and recovery only change the transition phase, so any prompt accepted while the handoff was active remains permanently undelivered. Requeue it to the surviving controller or reject it atomically before accepting it.

illegalcall (changes requested):
Requesting changes for three spec/correctness failures verified end to end against issue #4577 on this exact head. These findings are limited to Chat-first execution, active-turn steering semantics, and accepted-message durability during interface handoffs.

cloud/internal/postgres/project_session_store.go
vanshx999: Fixed in 3b737a220. Chat-first creation now creates the initial `ao_turns` row in the same transaction as the `chat.user_message` event, attaches its ID to that event, and notifies the worker so ChatRunner claims and executes the original task brief.

cloud/internal/postgres/session_interface_transition_store.go
vanshx999: Fixed in 3b737a220. Every terminal phase transition now locks the handoff and atomically delivers all undelivered held prompts to `ao_turns` before updating the phase. That covers completion, cancellation, failure, and recovery, and delivery notifies the worker only when work was actually released.

cloud/internal/postgres/worker_turn_store.go
vanshx999: Partially addressed in 3b737a220: the control-plane mutation is now one transaction, retains the running turn, records an idempotent steer receipt/event, and never creates `cancel_requested` or a replacement turn. I am leaving this thread open because the current headless Cloud worker has no live provider-turn input/control primitive: persisting the receipt alone does not inject guidance into the already-running CLI process. Completing literal active-turn injection requires choosing a provider control protocol (or defining restart/resume semantics for Codex, Claude, and Cursor) rather than silently changing that behavior.

cloud/internal/postgres/session_interface_transition_store.go
```diff
@@ -0,0 +1,700 @@
...
+			transitionID, owner, from).Scan(&orgID, &sessionID); err != nil {
+			if errors.Is(err, pgx.ErrNoRows) {
+				return ErrTransitionStale
+			}
+			return err
+		}
+		if _, err := tx.Exec(ctx, `SELECT set_config('ao.org_id', $1, true)`, orgID); err != nil {
+			return err
+		}
+		if to.Terminal() {
```
illegalcall, line 530: **[P1] Keep held prompts fenced until a controller is usable.** `to.Terminal()` also covers failures reached from `source_stopped` and `target_starting`. At those points the source controller is already stopped and the target has not started, but this call inserts the held prompts into `ao_turns` and notifies the worker. A stopped Chat source has no runner to consume them; a stopped TUI can claim them through stale `agentStarted` state and then fail with `terminal is not open`. Restore/start a controller before releasing these prompts, or leave them fenced for recovery.

illegalcall (changes requested):
Rechecked current head. Chat-first prompt execution is fixed, but active-turn steering remains open in the existing thread and this post-stop handoff failure path can still lose accepted prompts.

cloud/internal/postgres/session_interface_transition_store.go
```diff
@@ -0,0 +1,700 @@
...
+)
+
+var (
+	// ErrTransitionInProgress reports a session that already owns an active
+	// interface transition. Only one controller handoff may run at a time.
+	ErrTransitionInProgress = errors.New("interface transition already in progress")
+	// ErrTransitionNotFound reports a transition id or notice that does not
+	// exist for the requested session.
+	ErrTransitionNotFound = errors.New("interface transition not found")
+	// ErrTransitionNotCancellable reports a transition the coordinator has
```
illegalcall, line 20: **[P2] Remove the unused Cloud transition surface.** This PR adds `ErrTransitionNotCancellable`, `GetSessionInterfaceTransition`, `ListActiveSessionInterfaceTransitions`, `SessionInterfaceTransitionPhase.Active`, `SessionInterfaceTransitionMessage`, and the matching `Server.store` interface entry, but an exhaustive repository reference check finds no Cloud caller or consumer for any of them. Delete this dead API/type surface; nothing replaces it.

cloud/internal/postgres/worker_session_store.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v5"
+)
+
+// WorkerAgentSessionID returns the latest provider-native conversation identity
+// captured by a TUI or Chat harness hook, fenced to the calling worker epoch.
+func (s *Store) WorkerAgentSessionID(ctx context.Context, orgID, sessionID, workerID string, epoch int64) (string, error) {
```
illegalcall, line 11: **[P2] Read worker session state once.** The only caller is `/worker/session`, and it always invokes `WorkerAgentSessionID`, `WorkerSessionModel`, and `WorkerSessionReasoningEffort` together. Each method repeats the same epoch fence and opens another session query; the worker then calls this endpoint separately for model and effort, producing six database reads for one command rebuild. Replace these three one-field methods with one fenced `SELECT` returning `agentSessionId`, `model`, and `reasoningEffort`.

cloud/internal/interfacereconcile/transport_driver.go
```diff
@@ -0,0 +1,204 @@
...
+	CreateCoordinatedInterfaceRequest(ctx context.Context, orgID, sessionID, kind string, payload json.RawMessage) (domain.WorkerRequest, error)
+	GetCoordinatedInterfaceRequestResult(ctx context.Context, orgID, sessionID, requestID string) (domain.WorkerRequest, error)
+}
+
+// TransportDriver dispatches interface commands to a live worker through AO's
+// durable request channel and waits for the worker's completion. The worker owns
+// the agent process, so the control plane never touches a PTY directly.
+type TransportDriver struct {
+	store RequestStore
+	owner string
```
illegalcall, line 27: **[P2] Remove unused driver state.** `TransportDriver.owner` and `TransportDriver.log` are assigned by the constructor but never read, and `workertransport.InterfaceTransition.agentTermID` is likewise never referenced. Remove these fields and the corresponding constructor arguments; nothing replaces them.

cloud/internal/interfacereconcile/transport_driver.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return nil
+}
+
+func (d *TransportDriver) InspectSource(
+	ctx context.Context,
+	transition postgres.CoordinatedInterfaceTransition,
+) (SourceInspection, error) {
+	var result SourceInspection
+	payload, _ := json.Marshal(map[string]any{
```
illegalcall, line 60: **[P2] Stop sending request fields the worker ignores.** `sourceInterface` and `sessionId` are serialized by the coordinator and decoded into `interfacePayload`, but no inspect, interrupt, stop, or native-id handler reads either field. Remove both fields and send an explicit `{}` payload for commands without arguments; `nil` is not safe here because `CreateCoordinatedInterfaceRequest` does not normalize it before the `JSONB NOT NULL` insert.

cloud/internal/workerexec/chat_output.go
```diff
@@ -0,0 +1,210 @@
...
+		)
+		return nil
+	}
+	if event.Type == "item.completed" && event.Item.Type == "agent_message" && strings.TrimSpace(event.Item.Text) != "" {
+		return []Output{{Stream: "stdout", Text: event.Item.Text}}
+	}
+	return nil
+}
+
+func (p *chatOutputProjector) projectClaudeJSONLine(line string) []Output {
```
illegalcall, line 128: **[P2] Share the identical result-event parser.** `projectClaudeJSONLine` and `projectCursorJSONLine` decode the same fields and take the same branches for empty input, invalid JSON, identity capture, error results, and successful text. Route both harnesses through one parser so their behavior cannot drift.

cloud/cmd/ao-worker/main.go
```diff
@@ -337,32 +408,16 @@ func startInteractiveAgent(
...
+	if strings.TrimSpace(bootstrap.Launch.Interface) == workertransport.InterfaceChat {
+		return nil // The headless controller owns this worker until a TUI handoff.
 	}
-	agentCommand, err := (workerexec.HarnessBuilder{DataDir: dataDir}).BuildInteractive(
-		bootstrap.Launch, credential, workspace,
-	)
+	if transportSupervisor.AgentCommandFactory == nil {
+		return errors.New("coding-agent command factory is unavailable")
+	}
+	agentCommand, err := transportSupervisor.AgentCommandFactory(ctx, bootstrap.Launch.AgentSessionID)
```
illegalcall, line 417: **[P2] Remove unused startup parameters.** `startInteractiveAgent` no longer reads `workspace`, `dataDir`, `pullRequestSocketPath`, `reviewSocketPath`, or `checkpointSocketPath` after command construction moved into `AgentCommandFactory`. Remove the five parameters and their call-site arguments.

frontend/src/renderer/hooks/useSessionInterfaceTransition.ts
```diff
@@ -151,11 +164,32 @@ export function sessionInterfaceTransitionQueryKey(sessionId: string) {
...
+) {
+	const isCloud = Boolean(cloud && cloudCp.ready);
+	const isLocal = cloud === null || !hasSessionContext;
 	return useQuery({
 		queryKey: sessionInterfaceTransitionQueryKey(sessionId ?? ""),
-		enabled: Boolean(sessionId && hasTrustedApiBaseUrl()),
+		enabled: Boolean(sessionId && (isCloud || (isLocal && hasTrustedApiBaseUrl()))),
 		queryFn: async () => {
+			if (isCloud && cloud) {
+				const [sessionResponse, transitionStatus] = await Promise.all([
```
illegalcall, line 180: **[P2] Remove the unused session fetch.** This `getSession` call only supplies `currentMode`, but `currentMode` is never read from the query result anywhere in the repository. Query only `getInterfaceTransition` and drop the unused property and extra control-plane request.

frontend/src/renderer/components/SessionInterfaceSwitch.tsx
```diff
@@ -67,6 +67,7 @@ export function SessionInterfaceSwitchButton({
 	cancelError,
 	onClick,
 	onCancel,
+	showLabel = false,
```
illegalcall, line 70: **[P2] Remove the unreachable label branch.** No `SessionInterfaceSwitchButton` caller passes `showLabel`, so the prop, labeled-control class branch, and rendered label are dead code. Remove them; nothing replaces this path.

frontend/src/renderer/components/chat/CloudSessionChatSurface.tsx
```diff
@@ -0,0 +1,222 @@
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { useEffect, useMemo, useState, type ReactNode } from "react";
+import { useCloudCp } from "../../hooks/useCloudCp";
+import type { CloudCpClientEvent } from "../../lib/cloud-cp";
+import type { ChatModel, ConversationItem, ConversationMessage, ConversationSnapshot, ConversationTurn, TurnSettings } from "../../types/conversation";
+import type { WorkspaceSession } from "../../types/workspace";
+import { ChatWorkspace } from "./ChatWorkspace";
+
+type EventPayload = {
+	attempt?: unknown;
```
illegalcall, line 10: **[P3] Remove the unused payload field.** `EventPayload.attempt` has no read anywhere in this adapter. Delete it; nothing replaces it.

illegalcall (changes requested):
Requesting changes for nine mechanically verified deletion/shrink opportunities on the current head. These comments are limited to dead or behavior-identical duplicate code; the migration and architecture judgment calls from the exploratory pass are intentionally excluded.

illegalcall:
### Could we reuse the existing handoff state machine?

Could you investigate whether the local and Cloud implementations can share one interface-handoff core instead of maintaining two separate state machines?

The existing local implementation already defines the durable phases and terminal-state rules:

- [Local transition model](https://github.com/Untrivial-ai/agent-orchestrator/blob/0be2146992ea63ff02861f74cc234f23645ce43e/backend/internal/domain/session_interface_transition.go#L41-L84)
- [Local transition runner](https://github.com/Untrivial-ai/agent-orchestrator/blob/0be2146992ea63ff02861f74cc234f23645ce43e/backend/internal/session_manager/interface_transition.go#L362-L470)

This PR introduces substantially the same vocabulary and progression again:

- [Cloud transition model](https://github.com/Untrivial-ai/agent-orchestrator/blob/8ed46637354e74c5fbf8ee2411ba014814ca73f1/cloud/internal/domain/session_interface.go#L56-L88)
- [Cloud transition coordinator](https://github.com/Untrivial-ai/agent-orchestrator/blob/8ed46637354e74c5fbf8ee2411ba014814ca73f1/cloud/internal/interfacereconcile/coordinator.go#L208-L286)

Both implement approximately:

```text
requested
→ preflighting
→ draining
→ source_stopping
→ source_stopped
→ target_starting
→ activating
→ completed / failed / recovery_required
```

The storage and execution mechanisms genuinely need to differ:

```text
Local:
  SQLite + direct process/PTY control

Cloud:
  PostgreSQL + leases + remote worker commands
```

However, the phases, validation, allowed transitions, terminal-state decisions, and failure/recovery rules appear to be domain logic rather than storage or transport logic.

Cloud already depends on the backend module ([`cloud/go.mod`](https://github.com/Untrivial-ai/agent-orchestrator/blob/8ed46637354e74c5fbf8ee2411ba014814ca73f1/cloud/go.mod#L5-L7)) and already reuses public backend packages such as [`pkg/agentruntime`](https://github.com/Untrivial-ai/agent-orchestrator/blob/8ed46637354e74c5fbf8ee2411ba014814ca73f1/cloud/internal/workerexec/command.go#L14), so sharing a public handoff package appears mechanically possible.

Could you investigate whether we can extract something like this?

```text
shared handoff core:
  phases, policies, validation
  next transition/action
  failure and recovery decisions

local adapters:
  SQLite store
  direct process driver

cloud adapters:
  PostgreSQL store
  leased scheduler
  worker-transport driver
```

I am not suggesting that Cloud call the current internal `runInterfaceTransition` method verbatim; it contains local process-specific behavior. The question is whether its environment-independent decisions can be extracted and used by both implementations.

If a shared coordinator is not feasible because Cloud requires leases and restart-safe distributed scheduling, could we at least share a pure transition reducer/state table and run the same conformance tests against both adapters? If even that is not possible, please document the concrete local-versus-Cloud invariant that requires two independent implementations.

cloud/internal/postgres/session_interface_transition_store.go
vanshx999: Fixed in 349c922fd. Both local and Cloud transition domains now alias the shared backend/pkg/interfacehandoff state table: the same phase vocabulary, legal normal edges, terminal rules, cancellation boundary, held-message release policy, and post-source-stop failure outcome are used by both runners. The local SQLite/direct-process runner remains its adapter because it can synchronously restore a process; Cloud keeps its leased PostgreSQL/worker coordinator and now marks source-stopping and later remote failures recovery_required. Cloud only releases held prompts after activating->completed or an early safe cancellation; source-stopped/target-starting failures leave them fenced for recovery. Focused backend, Cloud coordinator, and Postgres tests pass; the broader local interface suite has only the existing sandbox permission failure writing ~/.claude.json.

vanshx999:
Investigated and implemented the shared-core boundary. `backend/pkg/interfacehandoff` now owns the common policy/phase vocabulary, legal next edges, terminal/cancellation rules, held-message release safety, and conservative failure outcome. Both local and Cloud domain types alias it, and both runners validate advances through the same reducer. The coordinators remain separate adapters because local can synchronously restore a direct process, while Cloud must be lease/restart-safe and cannot infer controller liveness from a remote command result. Cloud now keeps held prompts fenced after source stopping until recovery; it releases them only after successful activation or an early safe cancellation. Focused backend and Cloud handoff tests pass.

vanshx999:
CI/build follow-up is complete on 9fd67e597: all required checks are green, including build-test, lint, cloud-build-test, frontend test, API/SQL drift, and platform jobs. The shared handoff recovery regression was fixed by allowing the intentional idempotent target_starting -> target_starting recovery checkpoint. The PR is mergeable. Review is not fully complete yet: the active-turn steering thread and the remaining P2/P3 cleanup threads are still open and are intentionally not marked resolved.

vanshx999:
Scope audit at head c3cc869ac

The PR currently changes 100 files (+7,783/-496). I counted 24 test files: 11 added and 13 modified (+2,537/-26 raw test-file diff).

Breakdown by area:
- cloud/: 56 files (+5,842/-183), including the durable transition store, lease-based coordinator, worker transport/agent lifecycle, event projection, migrations, and Cloud API handlers.
- frontend/: 33 files (+1,082/-115), including the Cloud Chat surface, SessionView/TerminalPane, switch hooks, routes, and Cloud control-plane client.
- backend/: 5 files (+282/-53), including the shared phase/policy reducer in backend/pkg/interfacehandoff.
- contracts/packages/: generated OpenAPI and cloud-client changes (+538/-1).
- docs/submodule: one documentation file and the private/ao-cloud pointer.

Why most of this is needed:
This is not only a UI toggle. A Cloud switch crosses the renderer, Cloud API, PostgreSQL durable state, lease/fencing/recovery, worker stop/start commands, provider transcript projection, and the event log. Removing any of those layers leaves a switch that cannot survive worker restarts, preserve Chat history, or safely complete TUI <-> Chat handoffs. The 24 tests cover those boundaries and should not be removed just to reduce the file count.

What is not strictly core and should either be split or explicitly justified:
1. cloud/internal/postgres/migrations/00042_browser_fetch_worker_requests.sql and its worker_request_kind_test.go cover an unrelated browser.fetch constraint regression. They can be a separate fix, unless this PR intentionally owns the combined worker-request allowlist.
2. cloud/scripts/deploy-dev.sh is a NodeOps/DNS development-deploy change, not handoff logic.
3. private/ao-cloud is a companion-repository pointer; it belongs here only if the deployment requires the matching private revision.
4. The session_model and session_reasoning_effort migrations plus their plumbing are feature-adjacent. They are needed only if preserving Chat-selected model/effort across the interface switch is part of the acceptance criteria; otherwise they can be split.
5. The docs update can be split without affecting runtime behavior.

The generated OpenAPI/cloud-client files are required when the public API changes. My recommendation is to split the clearly unrelated browser-fetch/deploy/submodule work, but keep the Cloud/backend/frontend handoff path and its tests together because they form one end-to-end feature.

vanshx999:
Follow-up to the scope audit: removed the unrelated  DNS/NodeOps override in commit e4ef96c19. The PR is now 99 files (+7,778/-495). I retained the interface migration/worker-request allowlist, model/effort persistence, worker transport, event projection, frontend Cloud surface, generated API client, private Cloud pointer, and their tests because they are part of the end-to-end handoff or preserve required behavior. The Cloud test suite passes with the required local socket/listener permissions.

vanshx999:
Correction to the previous follow-up comment: removed the unrelated cloud/scripts/deploy-dev.sh DNS/NodeOps override in commit e4ef96c19. The PR is now 99 files (+7,778/-495). I retained the interface migration/worker-request allowlist, model/effort persistence, worker transport, event projection, frontend Cloud surface, generated API client, private Cloud pointer, and their tests because they are part of the end-to-end handoff or preserve required behavior. The Cloud test suite passes with the required local socket/listener permissions.

vanshx999:
Fixed and pushed as `584ddcfaa4583feb81faf84d2378d680a6c02762`.

The `BROWSER_PANEL_HIDDEN` failure was caused by this branch being behind main's revision-gated browser layout update. Syncing current main brings in the pointer-test fixtures' required `revision: 1` payloads. Verified locally: `browser-view-host.test.ts` 150/150 passed and frontend TypeScript typecheck passed. New CI is running.

illegalcall (changes requested):
Requesting changes on `584ddcf`. The shared handoff table is a useful improvement, but these correctness gaps remain:

- **[P1] The Cloud deployment build fails outside the monorepo workspace.** `cloud/internal/domain/session_interface.go` imports `backend/pkg/interfacehandoff`, while `cloud/go.mod` still pins a backend revision that does not contain that package. I archived this head's `cloud/` directory and ran `GOWORK=off go build ./cmd/ao-cloud`; it fails with `no required module provides package .../backend/pkg/interfacehandoff`. The Cloud Dockerfile builds from that standalone module, whereas CI picks up the root `go.work`. Please make the standalone Cloud image build pass and add a build check that cannot inherit the root workspace. [Import](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/domain/session_interface.go#L6) · [module pin](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/go.mod#L30) · [Docker build](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/Dockerfile#L12-L19)

- **[P1] An accepted prompt can be stranded by an early handoff failure.** During an active transition, `appendUserMessage` records the user event and holds its turn in `ao_interface_transition_messages`. If preflight or draining fails, the coordinator marks the transition `failed`; `AdvanceCoordinatedInterfaceTransition` does not deliver held messages, and terminal transitions are not claimed again. The prompt stays visible as accepted but never becomes worker work. Please release those prompts atomically after an early failure when the source controller is known usable, and cover that path with a test. [Admission](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/postgres/event_store.go#L355-L376) · [failure](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/interfacereconcile/coordinator.go#L399-L417) · [store update](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/postgres/session_interface_transition_store.go#L513-L558)

- **[P1] Held prompts also lack a release path after `recovery_required`.** The coordinator writes that terminal phase and excludes it from later claims. A replacement worker can restart the committed controller, but no Cloud path then reconciles this transition or delivers prompts held under its ID. Please add a recovery path that verifies controller readiness before releasing those prompts. [Recovery write](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/interfacereconcile/coordinator.go#L420-L438) · [claim filter](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/postgres/session_interface_transition_store.go#L426-L433) · [worker restart](https://github.com/Untrivial-ai/agent-orchestrator/blob/584ddcfaa4583feb81faf84d2378d680a6c02762/cloud/internal/workertransport/supervisor.go#L179-L207)

vanshx999:
Addressed the three review findings in `136792988`.

- Standalone Cloud builds now pin the backend revision containing `pkg/interfacehandoff`; the Dockerfile and `cloud-build-test` set `GOWORK=off`.
- Early preflight/source-drain failures release held prompts atomically only when the source remains usable; uncertain failures stay fenced for recovery.
- `recovery_required` transitions are claimed again. A replacement worker answers a new `interface.ready` request, and only a positive readiness proof atomically delivers held prompts and completes recovery.
- Added the readiness request migration and boundary tests.

Verified: `GOWORK=off go build ./cmd/ao-cloud`, `GOWORK=off go vet ./...`, and `GOWORK=off go test -race -count=1 ./...` all pass. Updated the required PR change-count header to match the current diff.

vanshx999:
Trimmed the PR by reverting two non-functional changes unrelated to the Chat↔TUI handoff: the broad docs/cloud-development.md rewrite and the private/ao-cloud submodule pointer update.

The diff is now 102 files (+7,983/-357), down from 104 files (+8,022/-501). I retained the handoff state machine, Cloud persistence/API/worker transport, generated contract updates, migration history, and regression tests because those are required for durable cross-process handoffs and the requested correctness cases.

vanshx999:
Current diff after cleanup: 102 files, +7,983/-356. Breakdown against main:

- Runtime code: 71 files, +4,882/-324 — PostgreSQL transition state/tenant FK, HTTP/OpenAPI handlers, worker leases/commands, Chat/TUI controller lifecycle, Cloud chat surface, and local/shared handoff reducer.
- Tests: 24 files, +2,616/-25 — coordinator/store/worker/controller/frontend coverage for the required failure, recovery, drain, steering, retry, and chat-first paths.
- Contracts/build: 7 files, +485/-7 — generated OpenAPI/client schema plus standalone Cloud build/module/migration wiring.

This is not a UI-only toggle: a Cloud handoff crosses renderer → control-plane API → durable Postgres state → leased worker transport → provider process, and must survive retries/restarts without losing accepted prompts. Removing any of those layers would either disable the feature or reintroduce the reviewed correctness bugs. I removed the only clearly unrelated pieces: the broad docs rewrite and private submodule pointer. Migration files are retained because they are append-only deployment history and may already be applied.

vanshx999:
Scope note: the 8k-line size is real and the PR is larger than a normal UI toggle. It accumulated the initial Cloud control-plane implementation plus many reliability follow-ups. The largest non-UI pieces are durable Postgres state (713 lines), distributed coordinator (468), worker transport (318), provider/process lifecycle and output projection, plus 2.6k lines of boundary tests and 466 generated contract lines.

The nine correctness requirements in the task (tenant-safe messages, chat-first turns, drain/interrupt semantics, fencing, retries, steering, structured queued turns, and failure/recovery delivery) cannot be implemented as a small frontend change.

If review size is the priority, the right solution is to split this into stacked PRs: (1) shared reducer/local conformance, (2) Cloud persistence/API/worker handoff, (3) frontend/generated client. The first optional scope to defer would be model/reasoning persistence and provider transcript-history projection. I will not delete those blindly because doing so changes Chat↔TUI continuity semantics; the current PR keeps them working end-to-end.

PR #4684, opened 2026-08-30
feat(chat): refine queued message dock

Annieeeee11:
## Summary
- Collapse keeps the next queued message visible while the remaining queue stays out of the way
- Expanded dock shows up to five messages and scrolls for additional queued work
- Empty Enter steers the next queued message into the running turn; repeated requests drain in order without being lost
- Keep the composer mounted and focused as the dock appears or queued messages are removed
- Refine queue actions and open/close motion for a quieter, more compact dock

## Test plan
- [x] `npm --prefix frontend test -- src/renderer/components/chat/ChatSteer.test.tsx`
- [x] `npm run frontend:typecheck`
- [x] `npm --prefix packages/product-ui run typecheck`
- [ ] Queue 2+ messages while a turn is running and collapse/expand the dock
- [ ] Queue 6+ messages and confirm only five rows show before scrolling
- [ ] Confirm steer, edit, and delete actions still work on visible rows

### DIFF
--- a/frontend/src/renderer/components/chat/ChatComposer.tsx
+++ b/frontend/src/renderer/components/chat/ChatComposer.tsx
@@ -27,12 +27,14 @@
  */
 
 import {
+	cloneElement,
 	useCallback,
 	useEffect,
 	useId,
 	useMemo,
 	useRef,
 	useState,
+	isValidElement,
 	type ClipboardEvent,
 	type DragEvent,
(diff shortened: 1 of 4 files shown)


ronishrohan (approved):
Approved after final review: queued-steer requests drain sequentially, composer width is preserved, and the full frontend CI suite is green.

PR #4687, opened 2026-08-30
fix(claude-code): normalize Windows trust paths

### LINKED ISSUE #3407: [Bug]: Windows: claude-code trust pre-seed writes backslash path keys in .claude.json, but Claude Code only reads forward-slash keys — trust dialog blocks every session (no_signal)
## Summary

On Windows, AO pre-seeds Claude Code workspace trust by writing `projects["<worktree>"].hasTrustDialogAccepted = true` into `%USERPROFILE%\.claude.json` — but it writes the path key with **backslashes** (`"D:\\...\\worktrees\\<project>\\<session>"`). Claude Code keys the same map with **forward slashes** (`"D:/.../worktrees/<project>/<session>"`) and its trust lookup only ever matches forward-slash keys. The keys never match, so the pre-seeded flag is never read. Claude Code then renders the interactive workspace-trust dialog inside AO's ConPTY and waits for a keypress forever.

Consequence chain: agent never starts → no hook (`ao hooks claude-code <event>`) ever fires → `sessions.first_signal_at` stays NULL → AO reports `no_signal` for every claude-code session.

This looks like the Windows follow-up to #2119: the pre-seeding approach adopted there is defeated on Windows by the path-separator mismatch.

## Steps to reproduce

1. Windows, AO 0.11.2, `claude` resolving to the native `claude.exe` on the daemon's PATH.
2. ```powershell
   ao project add --path <repo> --id demo --name Demo --worker-agent claude-code
   ao spawn --project demo --harness claude-code --name demo --prompt "Create src/demo.txt containing exactly DEMO OK."
   ao session ls --json    # idle → no_signal after ~90 s, no file created
   ```
3. Inspect `%USERPROFILE%\.claude.json`: AO's entry for the worktree uses backslashes; every entry written by Claude Code itself uses forward slashes.

To see *why* it hangs, run AO's exact agent argv under a PTY (e.g. `winpty` from Git for Windows) and read the screen:

```text
 Accessing workspace:
 D:\...\worktrees\<project>\<session>
 Quick safety check: Is this a project you created or one you trust? ...
 > 1. Yes, I trust this folder
   2. No, exit
 Enter to confirm · Esc to cancel
```

## Expected behavior

The trust key AO writes is the one Claude Code actually reads; worker sessions start unattended on Windows.

## Actual behavior

AO writes:

```json
"projects": {
  "D:\\...\\worktrees\\<project>\\<session>": { "hasTrustDialogAccepted": true }
}
```

Claude Code reads/writes its own entries in the same file as:

```json
"D:/.../repo": { "hasTrustDialogAccepted": false }
```

Keys never match → blocked interactive prompt → `no_signal`.

**Proof by inversion:** adding the *same* trust entry keyed with forward slashes, changing nothing else, removes the dialog; the session reaches activity (`POST /api/v1/sessions/<id>/activity status=200`), creates the requested file, and kills cleanly.

## AO version

0.11.2 (desktop app; bundled daemon `resources\daemon\ao.exe`)

## Environment

Windows 10 Pro (10.0.19045) x64; ConPTY runtime; native `claude.exe` (not the npm `.cmd` shim)

## Logs / extra context

- Claude Code's trust check **walks up the directory tree** (any ancestor with `hasTrustDialogAccepted: true` passes). Operational workaround until fixed: trust `%USERPROFILE%\.ao\data\worktrees` once, interactively — it covers every future worktree AO creates beneath it, with no per-session prediction. Per-session pre-seeding with forward-slash keys, and answering the dialog via `ao send --session <id> --message "1"`, both also work but are per-session only (Claude Code does not persist the dialog answer for AO-created worktrees, and each session gets a fresh path).
- Suggested fix: write the `projects` key in the format Claude Code reads on Windows (forward slashes).
- Side note: `CLAUDE_CONFIG_DIR` appears to be unknown to AO (no occurrences in the binary) — trust is always written to `%USERPROFILE%\.claude.json` regardless of that variable.
- Downstream effect: the blocked process holds the worktree directory as its cwd, so `ao session kill` fails post-kill cleanup with INTERNAL_ERROR (filing separately).
- Duplicate search: #2119 (macOS, the pre-seed request this defect follows from), #1396, #3110 — none cover the Windows path-separator mismatch.

JAYATIAHUJA:
Fixes #3407

## Summary
- normalize Claude Code workspace trust keys to forward slashes on Windows
- preserve non-Windows project keys unchanged
- cover both behaviors with JSON-level regression tests

## Test
- go test ./internal/adapters/agent/claudecode -count=1
- go build ./...
- go vet ./...

## Full-suite note
- go test ./... still reports pre-existing Windows portability failures across unrelated packages (tracked separately); the touched Claude Code package passes

### DIFF
--- a/backend/internal/adapters/agent/claudecode/claudecode.go
+++ b/backend/internal/adapters/agent/claudecode/claudecode.go
@@ -566,8 +566,15 @@ func claudeConfigPath() (string, error) {
 var claudeTrustMu sync.Mutex
 
 func ensureWorkspaceTrusted(configPath, workspacePath string) error {
+	return ensureWorkspaceTrustedForOS(configPath, workspacePath, runtime.GOOS)
+}
+
+func ensureWorkspaceTrustedForOS(configPath, workspacePath, goos string) error {
 	claudeTrustMu.Lock()
 	defer claudeTrustMu.Unlock()
+	if goos == "windows" {
+		workspacePath = strings.ReplaceAll(workspacePath, `\`, "/")
+	}
 
 	root := map[string]any{}
 	data, err := os.ReadFile(configPath)
--- a/backend/internal/adapters/agent/claudecode/claudecode_test.go
+++ b/backend/internal/adapters/agent/claudecode/claudecode_test.go
@@ -942,6 +942,50 @@ func TestEnsureWorkspaceTrustedCreatesEntry(t *testing.T) {
 	}
 }
 
+func TestEnsureWorkspaceTrustedNormalizesWindowsProjectKey(t *testing.T) {
+	dir := t.TempDir()
+	cfgPath := filepath.Join(dir, ".claude.json")
+	if err := os.WriteFile(cfgPath, []byte(`{"projects":{}}`), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	const workspacePath = `D:\dev\agent-orchestrator\.ao\worktrees\demo\worker-1`
+	const wantKey = `D:/dev/agent-orchestrator/.ao/worktrees/demo/worker-1`
+	if err := ensureWorkspaceTrustedForOS(cfgPath, workspacePath, "windows"); err != nil {
+		t.Fatalf("ensureWorkspaceTrustedForOS: %v", err)
+	}
+
+	root := readJSON(t, cfgPath)
+	projects := root["projects"].(map[string]any)
+	entry, ok := projects[wantKey].(map[string]any)
+	if !ok || entry["hasTrustDialogAccepted"] != true {
+		t.Fatalf("forward-slash trust entry = %#v, want accepted entry", projects[wantKey])
+	}
+	if _, exists := projects[workspacePath]; exists {
+		t.Fatalf("unexpected backslash-keyed trust entry: %#v", projects[workspacePath])
+	}
+}
+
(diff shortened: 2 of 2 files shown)


Pulkit7070:
Reviewed end to end. This correctly fixes #3407.

- The fix matches the behavior documented in the issue: Claude Code keys workspace trust by forward slashes on Windows and only matches forward-slash keys, so AO's backslash keys never matched and the trust dialog blocked every session. Normalizing the key with `ReplaceAll(\, /)` on Windows makes AO's pre-seeded key match.
- The normalized key is used consistently for both the already-trusted check and the write, so they agree. Non-Windows keys are left untouched, including a path that contains a literal backslash. The OS is injected as a parameter so both branches are unit-testable.

Ran locally: build, vet, and the full claudecode package pass, and the trust tests pass `-race -count=10`.

One optional note, not blocking: a Windows machine that already has a backslash-keyed entry from before this fix keeps that stale key in `.claude.json` alongside the new forward-slash one. It is harmless because Claude Code ignores the non-matching key, but the old entry is not cleaned up. Fine to leave; worth a line if you want the pre-seed to also drop the matching backslash key.

Good to merge from my side.

Pulkit7070 (approved):
Approving. Correctly fixes #3407 by normalizing Windows trust keys to forward slashes to match Claude Code's lookup; verified against the issue, and the claudecode package passes -race -count=10 locally.

PR #4690, opened 2026-08-31
fix(chat): settle streamed messages with completed turns

ronishrohan:
## Summary

- settle streamed assistant messages when their enclosing conversation turn reaches a terminal state
- cover providers that emit `turn/completed` without a separate assistant `item/completed` event (including steering flows)
- add a storage regression test and regenerate sqlc output

## Root cause

The shared turn settlement path finalized the turn and running activities, but left assistant messages with `streaming = 1` when a provider ended the turn without sending an item completion event. The renderer then kept showing the response as live indefinitely.

## Verification

- `go test ./internal/storage/sqlite/store ./internal/service/chat`
- frontend conversation and steering tests: 50 passed
- `go test ./...` (changed packages pass; unrelated environment-dependent failures remain in Crush auth, provider conformance, and tmux integration tests)

### DIFF
--- a/backend/internal/storage/sqlite/gen/conversations.sql.go
+++ b/backend/internal/storage/sqlite/gen/conversations.sql.go
@@ -2824,6 +2824,29 @@ func (q *Queries) SettleRunningConversationActivitiesForTurn(ctx context.Context
 	return err
 }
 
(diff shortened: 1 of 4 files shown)


Annieeeee11 (approved):
lgtm

PR #4694, opened 2026-08-31
fix: make CI injection a session policy

AgentWrapper:
## Summary

- make the session row the sole source of truth for automatic CI-failure injection across every associated PR
- read the live session policy in lifecycle reactions and derive each PR summary display value from it
- remove the redundant PR policy column, propagation transaction, and PR CDC policy event via migration 0119
- preserve migration-history repair behavior after the PR policy column and its 0111 trigger predicate are removed

This follows the synchronization fix in #4543 by eliminating the duplicated per-PR state entirely. The existing `autoInjectCI` field in PR summaries remains wire-compatible, but is now derived from the owning session.

## Tests

- `go test ./internal/lifecycle ./internal/service/session ./internal/storage/sqlite/... ./internal/httpd/...`
- `env -u AO_PROJECT_ID -u AO_SESSION_ID -u AO_DATA_DIR -u AO_RUN_FILE -u AO_APP_RUN_ID -u AO_OWNER -u AO_ACP_RUNTIME_DIR -u AO_TMUX_BINARY -u AO_TMUX_SOCKET_NAME -u AO_TELEMETRY_APP_VERSION npm run lint`
- `GOLANGCI_LINT_CACHE=/tmp/ao-lint-agent-orchestrator-201 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs`
- `npm run sqlc && npm run api`

## Risk

- migration 0119 replaces `pr_cdc_update` before dropping `pr.auto_inject_ci`; its down migration restores both the column values (from the owning session) and the prior trigger shape
- migration-history repair now recognizes the post-0119 schema so later startups do not attempt to replay 0111 against the removed column

### DIFF
--- a/backend/internal/adapters/agent/fake/fake_test.go
+++ b/backend/internal/adapters/agent/fake/fake_test.go
@@ -231,10 +231,6 @@ func (s *lifecycleStore) ListPRsBySession(_ context.Context, _ domain.SessionID)
 	return nil, nil
 }
 
-func (s *lifecycleStore) GetPR(_ context.Context, prURL string) (domain.PullRequest, bool, error) {
-	return domain.PullRequest{URL: prURL, AutoInjectCI: true}, true, nil
-}
-
 func (s *lifecycleStore) ListPRReviews(_ context.Context, _ string) ([]domain.PullRequestReview, error) {
 	return nil, nil
 }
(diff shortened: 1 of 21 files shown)


codebanditssss:
Thanks for the contribution. The requested session-level CI injection behavior is already present on current : #3772 added the session policy/API/UI, and #4543 propagates that policy to the session's pull requests. This PR is therefore no longer necessary.

PR #4697, opened 2026-08-31
fix(cloud-terminal): stop infinite ticket minting and surface error on WebSocket failure (#4668)

### LINKED ISSUE #4668: [Bug][Cloud][P1] Failed terminal attach retries forever: ~2 ticket mints/sec, no surfaced error, sandbox never idle-pauses

priyanshujaiz:
## What

When a cloud terminal WebSocket fails (CSP block, proxy, firewall), the app
was retrying indefinitely at ~1-2 mints/sec with no user-facing error.
This fix adds a circuit-breaker on the frontend and a lease-refresh guard
on the backend.

## Why

Fixes #4668

Two costly side effects of the infinite loop:
1. The sandbox never idle-paused — IssueTerminalTicket refreshes
   interactive_until on every mint, so billing ran indefinitely.
2. The failure was invisible — users saw "Connecting…" forever.

## How

Frontend (useTerminalSession.ts + TerminalPane.tsx):
- Added cloudConnectFailures counter to the runtime object.
- In scheduleReattach(), if a cloud session has never attached and the
  counter hits CLOUD_CONNECT_MAX_FAILURES (8), transition("error") is
  called and the retry loop stops (~8 seconds of trying).
- TerminalPane detects this specific error via isCloudConnectError and
  renders a CloudConnectError full-screen overlay (not the generic banner)
  with a "Retry connection" button that resets the counter via attach().

Backend (worker_transport_store.go):
- Before refreshing interactive_until, count unconsumed terminal tickets
  minted in the last 2-minute window.
- If the count exceeds unconsumedTicketThreshold (10), the CASE clause
  skips the lease extension. Sandbox wake (paused → running) still fires.
- A slog.Warn is emitted with org_id, session_id, and count for
  observability without requiring a metrics system change.

Intentional omission: no integration test for the backend path —
the postgres package has no unit-test infrastructure and the lease
suppression SQL is covered by private CI's integration suite.

## Testing

Frontend — 4 new Vitest tests in useTerminalSession.test.tsx:
  npm test -- src/renderer/hooks/useTerminalSession.test.tsx
  → 50 passed (50)

Manual repro (DevTools Network → Offline mode → open cloud terminal):
- Before: "Connecting…" forever, ~2 ticket requests/sec in Network tab
- After:  overlay appears after ~8s, Network tab goes quiet,
          Retry button reconnects once Network is restored

Backend — package compiles and vets clean:
  go build ./cloud/internal/postgres/...
  go vet ./cloud/internal/postgres/...

## RESULTS

Screenshot 2026-08-31 110123

## Checklist


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

Pritom14:
## Review: infinite terminal ticket minting (#4668)

I reviewed this against issue #4668 and verified it by tracing the actual code paths (the frontend mux plus hook wiring and the backend SQL), building and vetting the cloud module, and reading the new tests. I also have independent staging evidence of this exact failure mode (a cloud terminal minting a fresh ticket roughly every 2 seconds with sub-second socket lifetimes), so the problem this targets is real and worth fixing.

### What the PR does
Two defenses for the failed-attach-retries-forever loop:
- Frontend: a `cloudConnectFailures` counter in the terminal-session runtime. In `scheduleReattach`, a cloud session that has never attached trips into an `error` state after `CLOUD_CONNECT_MAX_FAILURES` (8) failed cycles, and `TerminalPane` renders a dedicated `CloudConnectError` overlay with a Retry button instead of the generic banner.
- Backend (`worker_transport_store.go`): before `IssueTerminalTicket` refreshes the sandbox `interactive_until` lease, it counts unconsumed `terminal:%` tickets minted in the last 2 minutes; above `unconsumedTicketThreshold` (10) it skips the lease refresh so the idle scanner can eventually pause the sandbox. The paused-to-running wake still fires and the ticket is still issued, so the client still receives a normal 201.

### Use case it addresses
When the terminal WebSocket cannot attach (CSP block, proxy, firewall), the app used to loop mint-ticket forever with no surfaced error, and because every mint refreshed `interactive_until`, the sandbox never idle-paused and billed compute indefinitely. This PR is meant to both surface the failure and stop the billing leak.

### What I verified
- `cd cloud && go build ./...` and `go vet ./...` are clean.
- The backend defense is correctly targeted: it counts only unconsumed tickets, so a healthy viewer whose tickets get redeemed is never throttled, and it fails open (defaults to refreshing the lease) if the count query errors. At the observed rate of about 2 per second, the 10-ticket threshold is crossed in roughly 5 seconds, and `PauseIfIdle` gates on `interactive_until <= now()`, so the lease genuinely lapses and the sandbox pauses within about 2 minutes. This part is sound.
- i18n is complete: all 8 locale files carry the 3 new keys with real translations.

### Findings

**Blocker: the frontend breaker cannot tell "worker still provisioning" from "WebSocket permanently blocked", so it false-fires during normal slow cold starts.**
`useTerminalSession.ts` increments `cloudConnectFailures` on every not-yet-attached cloud "closed" event. But a still-provisioning worker returns 409 WORKER_UNAVAILABLE on the ticket mint, and `cloud-terminal-mux.ts` (around line 107) converts that 409 into the same `setConnectionState("closed")` that a real socket block produces. The hook has no way to distinguish them, so both increment the counter. At the flat 1s retry, the breaker trips at about 8 seconds. NodeOps and especially Coder cold starts routinely exceed that (the same file's own comment anticipates a worker ready at 17s). So a normal fresh or resumed cloud session whose worker takes longer than about 8 seconds to become ready shows a misleading "Cannot connect to terminal, check proxy, firewall, or CSP" overlay and stops the readiness poll the flat-retry path was explicitly designed to tolerate. Issue #4668 called out this exact distinction ("nothing distinguishes worker still provisioning from attach permanently failing client-side"), and the PR does not implement it. Suggested direction: have the mux signal "ticket minted (201) but socket failed" separately from "mint 409", and only count the former. That is the same consumed-versus-unconsumed signal the backend already keys on, so the two sides would agree.

**Major: the exponential backoff #4668 asked for is not implemented.** Only the hard cap is added; the never-attached cloud path stays at the flat `CLOUD_CONNECT_RETRY_MS = 1000`, so each attempt still mints roughly 8 tickets over about 8 seconds. Worth confirming whether the cap is intended to fully replace backoff.

**Major: the backend billing and SQL change has no test coverage.** `internal/postgres` has no test harness, so the threshold boundary (10 unconsumed freezes the lease, wake still fires, a healthy single-ticket session is not throttled) is unverified by anything runnable. Note also that the cloud module is not covered by CI at all (`.github/workflows/go.yml` triggers only on `backend/**` and `pricing/**`), so the passing "test" check does not exercise this file.

**Minor: the restore path re-enters `connect()` from the error state without resetting the counter** or `hasAttachedOnce`, so a single subsequent "closed" re-trips the breaker immediately. Only the Retry button, which routes through `attach()`, resets cleanly.

**Minor: the overlay detection couples to a hardcoded English substring** (`error?.includes("WebSocket cannot connect")` in `TerminalPane.tsx`, matching the literal in `useTerminalSession.ts`). Editing that message silently breaks the overlay. A typed error or flag would be more robust.

**Nits:** a "trasient" typo and a truncated comment ("Detect specific cloud WebSocket block error and for "), plus a couple of double blank lines that may trip the frontend lint.

### Recommendation
Good direction, and the backend defense is solid. But the frontend breaker as written would regress normal slow provisioning with a misleading network error, which is the specific case the issue warned about. I would request changes on the provisioning false-positive (the blocker) and the missing backend test before merge.

Related: #4792 (WebSocket keepalive) fixes a different terminal drop (the 60-second ALB idle timeout on connections that do attach), so the two are complementary rather than overlapping.

codebanditssss:
Thanks for the contribution. This functionality is already present on current  following #5023, including bounded connection retries, the user-visible retry state, and suppression of repeated unconsumed-ticket attempts; #5329 subsequently hardened the cloud connection path further. This PR is therefore no longer necessary.

Issue #4700, opened 2026-08-31
Restart can leave a project without a working orchestrator


Rishet11:
## What happened

Restart was meant to recover a project whose orchestrator was no longer usable. Instead, AO began shutting down the current orchestrator, waited too long, returned an error, and then finished shutting down the old orchestrator without starting a replacement.

The terminal had already become unusable before Restart was clicked. Restart was an attempted recovery, not the cause of that earlier symptom.

## Why this happens

Restart currently follows this order:

1. Shut down the current orchestrator.
2. Wait for that shutdown to finish.
3. Start a new orchestrator.

The request has a normal time limit.

**If shutdown takes too long, or the new orchestrator fails to start, the old one is already gone. AO has no dependable fallback that restores the old orchestrator or guarantees a replacement.**

## What the user sees

- Restart reports a server error.
- The old orchestrator stops a short time later.
- No new orchestrator appears.
- The project is left without a coordinator for its sessions.

## Expected behavior

Restart should either replace the orchestrator successfully or leave the project with a usable coordinator. A failed replacement should not leave the project without one.

## Suggested fix

- Give orchestrator replacement its own time budget instead of relying only on the normal request timeout.
- Check that the new orchestrator has started before considering the replacement complete.
- If replacement fails, keep or restore the previous orchestrator when possible.
- Add tests for slow shutdowns, failed starts, and recovery after a failed replacement.

## Important distinction

This issue documents a failed Restart attempt. It does not prove that Restart caused the earlier terminal freeze. The terminal was already unhealthy before Restart began.

The separate database and storage reports may explain why the system became unhealthy, but they should not be treated as proven by this issue alone.

## Related issues

- [#2501](https://github.com/Untrivial-ai/agent-orchestrator/issues/2501): an active orchestrator can be terminated incorrectly.
- [#3862](https://github.com/Untrivial-ai/agent-orchestrator/issues/3862): the interface can become unusable after a replacement failure triggered by Settings.
- [#4317](https://github.com/Untrivial-ai/agent-orchestrator/issues/4317): cleanup can fail on Windows during replacement.
- [#4481](https://github.com/Untrivial-ai/agent-orchestrator/issues/4481): replacement or termination can fail when the terminal runtime is already gone.

## Impact

A Restart intended to recover a broken project can leave that project without any working orchestrator.

Rishet11:
## Scope confirmation after the reporter's timeline correction

[...]

PR #4701, opened 2026-08-31
fix(orchestrator): give replacement a lifecycle timeout

Rishet11:
## Summary

- give `POST /api/v1/orchestrators` a dedicated three-minute lifecycle bound while ordinary REST requests retain their configured timeout
- carry the closed renderer source enum into the daemon request and persist a correlated request event locally
- preserve safe source, clean, and mode fields in remote diagnostics and regenerate the API contract

## Why

In [#4700](https://github.com/Untrivial-ai/agent-orchestrator/issues/4700), a clean replacement began retiring a live Codex coordinator. The ordinary 60-second request deadline returned HTTP 500 after 62 seconds, retirement finished afterward, and that request created no successor orchestrator.

This fixes the exact observed timeout path and makes future initiating UI surfaces attributable. It does not make replacement atomic for every possible teardown or spawn failure; #4700 remains the tracker for that broader recovery path.

## Tests

- `npm run api`
- focused frontend tests: 43 passed
- product UI checks: 118 tests, typecheck, build passed
- `go test ./internal/httpd/... ./internal/adapters/telemetry -count=1`
- `go vet ./internal/httpd/... ./internal/adapters/telemetry`
- isolated `go test ./...`
- `go build ./...`

## Risk

The operation remains synchronous and has a hard three-minute ceiling. Generic recovery after an arbitrary post-retirement failure is intentionally not attempted here because relaunching the predecessor under changed project settings can create a mismatched or duplicate coordinator.

Refs #4700

### DIFF
--- a/backend/internal/httpd/api.go
+++ b/backend/internal/httpd/api.go
@@ -3,6 +3,7 @@ package httpd
 import (
 	"log/slog"
 	"net/http"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
@@ -130,6 +131,7 @@ func NewAPI(cfg config.Config, deps APIDeps) *API {
 			Svc:           deps.Sessions,
 			Activity:      deps.Activity,
 			Usage:         deps.UsageHooks,
+			Telemetry:     deps.Telemetry,
 			Attachments:   attachmentstore.New(cfg.DataDir),
 			PreviewServer: deps.PreviewServer,
 			Capabilities:  deps.SessionCapabilities,
@@ -164,7 +166,7 @@ func (a *API) Register(root chi.Router) {
 		r.Get("/openapi.yaml", apispec.ServeYAML)
 
 		r.Group(func(r chi.Router) {
-			r.Use(middleware.Timeout(timeout))
+			r.Use(restTimeout(timeout))
 			r.Use(presenceMiddleware(a.deps.Presence))
 			a.agents.Register(r)
 			a.projects.Register(r)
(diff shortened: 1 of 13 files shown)


Rishet11:
Closing because this only extends the replacement request timeout. It does not make replacement safe, prevent the old orchestrator from being lost, or fix the original overload failure. Issue #4700 remains open for a complete fix that protects or recovers the predecessor when successor creation fails.

backend/internal/httpd/api.go
```diff
@@ -192,6 +194,28 @@ func (a *API) Register(root chi.Router) {
 	})
 }
 
+// restTimeout keeps ordinary REST requests under the configured budget while
+// allowing an orchestrator spawn enough time to retire the current coordinator
+// and create its successor. The operation remains bounded; it simply must not
+// inherit a deadline shorter than its normal two-phase lifecycle.
+func restTimeout(ordinary time.Duration) func(http.Handler) http.Handler {
+	orchestrator := ordinary
```
Rishet11, line 202: [P2] Enforce the documented three-minute ceiling

When AO_REQUEST_TIMEOUT is configured above three minutes, this assignment preserves the larger value for POST /api/v1/orchestrators. The PR description calls this a hard three-minute lifecycle ceiling, but the config accepts any positive duration, so a 5m or 10m setting lets this operation run that long. Use the dedicated duration as the route timeout, or document that it is only a minimum.

backend/internal/httpd/api.go
```diff
@@ -192,6 +194,28 @@ func (a *API) Register(root chi.Router) {
...
+	if orchestrator < config.DefaultOrchestratorSpawnTimeout {
+		orchestrator = config.DefaultOrchestratorSpawnTimeout
+	}
+	return func(next http.Handler) http.Handler {
+		ordinaryHandler := middleware.Timeout(ordinary)(next)
+		orchestratorHandler := middleware.Timeout(orchestrator)(next)
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.Method == http.MethodPost && r.URL.Path == "/api/v1/orchestrators" {
```
Rishet11, line 210: [P2] Cover orchestrator creation through the sessions route

The special case only recognizes POST /api/v1/orchestrators, but the existing project creation flow in frontend/src/renderer/routes/_shell.tsx creates its initial orchestrator with POST /api/v1/sessions and kind=orchestrator. That request still receives the ordinary RequestTimeout, 60s by default, so a slow initial orchestrator can hit the same timeout this change claims to bound. Classify that request too or route it through the dedicated endpoint.

Rishet11 (commented):
## Review

**Needs changes.**

The dedicated timeout addresses the reported 60-second replacement failure, but two correctness gaps remain:

1. The three-minute budget is not a ceiling when AO_REQUEST_TIMEOUT is configured higher.
2. Initial project-add orchestrator creation still uses the generic sessions route, so it does not receive the dedicated budget.

I did not run tests or builds because this reviewer task forbids executing project programs.

backend/internal/httpd/api.go
Rishet11: Acknowledged. This gap was not fixed. PR #4701 was intentionally closed because extending the timeout is only a partial workaround and does not make orchestrator replacement safe. Issue #4700 remains open for a complete redesign. Leaving this thread unresolved because no code change addressed it.

backend/internal/httpd/api.go
Rishet11: Acknowledged. This gap was not fixed. PR #4701 was intentionally closed because the route-specific timeout approach is incomplete and does not solve the original incident. Issue #4700 remains open for a complete replacement flow. Leaving this thread unresolved because no code change addressed it.

Rishet11:
Review [5066624908](https://github.com/Untrivial-ai/agent-orchestrator/pull/4701#pullrequestreview-5066624908) acknowledged. Both findings are valid. Neither was implemented because this PR was already closed at the reporter request as an insufficient partial fix. Replies were added to both inline comments, and both threads remain unresolved because no code change addressed them.

PR #4572, opened 2026-08-28
feat: unify session actions, terminal tabs, and naming

Annieeeee11:
## Summary [THIS IS @ronishrohan ]

Unifies the session header around three related pieces of terminal workspace UI:

- **Session actions menu** — moves interface switching and agent switching into the primary agent tab’s ⋮ menu, shared by chat and terminal surfaces.
- **Unified tab row** — puts agent, reviewer, and shell tabs in the same scrollable composer with consistent dividers, alignment, close behavior, spacing, and smart overflow fades. New shell tabs smoothly scroll into view.
- **Stable terminal names** — assigns shell names when they are created (`Terminal 1`, `Terminal 2`, …); names do not change when tabs are reordered or closed. Manual renames remain supported.

Also improves dropdown icon contrast and keeps the plus/open/trash action spacing consistent.

image

## Test plan

- [x] `cd frontend && npm test -- --run src/renderer/components/CenterPane.test.tsx src/renderer/components/ShellTerminalTab.test.tsx src/renderer/components/chat/ChatWorkspace.test.tsx src/renderer/hooks/useShellTerminals.test.tsx`
- [x] `cd backend && go test ./internal/service/shellterm`
- [ ] Verify the agent tab action menu in both chat and terminal modes.
- [ ] Open multiple shell terminals; confirm names are assigned at creation and remain stable after close/reorder.
- [ ] Confirm both edge fades reflect horizontal overflow and the right fade stays hidden during automatic scroll-to-new-tab.

PR #4711, opened 2026-08-31
fix(frontend): prevent switch-agent dialog auto-dismiss race with dropdown menu

### LINKED ISSUE #4710: bug(desktop): Switch agent dialog opens and immediately closes itself
## Bug

Clicking **"Switch agent"** in the session actions dropdown (topbar `...` menu) opens `SwitchAgentDialog`, and it immediately closes again in the same tick — the dialog flashes and disappears instead of staying open.

## Repro

1. Open any active worker session.
2. Click the session actions `...` menu in the terminal topbar.
3. Click **"Switch agent"**.
4. Expected: the switch-agent dialog opens and stays open.
5. Actual: the dialog opens and is dismissed immediately (visible as a brief flash).

## Root cause

Two independent Radix dismissal mechanisms fire on the same click:

1. `TerminalSwitchAgentButton.tsx` renders the menu item as a `DropdownMenuItem` with `onSelect={() => handleOpenChange(true)}`. `onSelect` does not call `event.preventDefault()`, so Radix's default post-`onSelect` behavior closes the parent `DropdownMenu` (`SessionActionsMenu.tsx`) in the same tick `SwitchAgentDialog`'s `open` prop flips to `true`.
2. `SwitchAgentDialog.tsx` renders `<Dialog modal={false} ...>` — intentionally non-modal so it doesn't fight `xterm`'s own focus/keyboard handling underneath it (the dialog is portaled into a container inside the terminal pane, not `document.body`). Because it's non-modal, Radix's `Dialog.Content` relies on `DismissableLayer`'s document-level `pointerdown`/`focusout` listeners for outside-click dismissal instead of a focus trap. The dropdown item is unmounting and returning focus in the very same event cycle the dialog mounts, so `DismissableLayer` treats the residual pointer/focus activity as "outside" and immediately closes the just-opened dialog.

A prior fix (`263ba3b82` / PR #4572, "unify session actions, terminal tabs, and naming") already moved `SwitchAgentDialog`'s render site from inside `TerminalSwitchAgentButton` into `SessionView`, specifically so the closing dropdown wouldn't *unmount* the dialog before it opened. That solved the unmount race but left the `DismissableLayer` outside-click race untouched, which is what's reported here. Possibly related to the closed #4039 ("fills terminal and reanimates on repeated activation") — different symptom, likely same family of dropdown/dialog interaction issues around this control.

## Fix

A PR is attached that adds an `onPointerDownOutside`/`onFocusOutside` guard on `SwitchAgentDialog`'s `DialogContent`, ignoring only outside events whose target is inside the just-dismissed menu/trigger (`[role="menuitem"]`, `[role="menu"]`, `[data-session-actions-trigger]`), so a genuine outside click still closes the dialog normally. Pattern matches the existing `onEscapeKeyDown`/`onPointerDownOutside` guard already used in `InstallDependencyDialog.tsx`.

Covered by a new regression test (`SessionActionsMenu.switch-agent.test.tsx`) that drives the real `DropdownMenuItem` click path — unlike the existing tests, which either mock the button out or open the dialog directly with `open=true` and so don't exercise this race.

axisrow:
Closes #4710.

## Bug

Clicking **"Switch agent"** in the session actions dropdown opened `SwitchAgentDialog` and then immediately closed it again in the same tick — the dialog flashed and disappeared instead of staying open.

## Root cause

Two independent Radix dismissal mechanisms fire on the same click:

1. `TerminalSwitchAgentButton.tsx`'s `DropdownMenuItem onSelect={() => handleOpenChange(true)}` doesn't call `event.preventDefault()`, so Radix's default post-`onSelect` behavior closes the parent `DropdownMenu` (`SessionActionsMenu.tsx`) in the same tick the dialog's `open` prop flips to `true`.
2. `SwitchAgentDialog.tsx` renders `<Dialog modal={false} ...>` — intentionally non-modal so it doesn't fight `xterm`'s focus/keyboard handling underneath it. Because it's non-modal, Radix's `Dialog.Content` relies on `DismissableLayer`'s document-level `pointerdown`/`focusout` listeners for outside-click dismissal. The dropdown item unmounts and returns focus in the very same event cycle the dialog mounts, so `DismissableLayer` treats the residual pointer/focus activity as "outside" and immediately dismisses the just-opened dialog.

`263ba3b82` (#4572) already moved `SwitchAgentDialog`'s render site from inside the button into `SessionView` to stop the closing dropdown from *unmounting* the dialog before it opened — that fixed the unmount race but left this `DismissableLayer` outside-click race untouched.

## Fix

Add an `onPointerDownOutside`/`onFocusOutside` guard on `SwitchAgentDialog`'s `DialogContent` that ignores only outside events whose target is inside the just-dismissed menu/trigger (`[role="menuitem"]`, `[role="menu"]`, `[data-session-actions-trigger]`), so a genuine outside click still closes the dialog normally. This mirrors the existing `onEscapeKeyDown`/`onPointerDownOutside` guard pattern already used in `InstallDependencyDialog.tsx`.

## Tests

- **New regression test**: `SessionActionsMenu.switch-agent.test.tsx` drives the real `DropdownMenuItem` click path (`variant="menu-item"` inside a real `SessionActionsMenu`), unlike the existing `SessionView.test.tsx` (mocks the button out) or `SwitchAgentDialog.test.tsx` (opens with `open=true` directly) — neither exercised this race. It fails against `main` with the exact reported symptom before the fix (verified via `git stash`), and passes after.
- A companion assertion in the same test proves a genuine outside click (`document.body`) still dismisses the dialog, so the guard doesn't over-suppress legitimate dismissal.
- Full relevant suite green: `SwitchAgentDialog.test.tsx` + `TerminalSwitchAgentButton.test.tsx` + `SessionView.test.tsx` + the new file — 119/119 passing.
- Full `frontend` vitest suite: 3189 passed / 1 skipped / 4 failed, all 4 pre-existing and unrelated (confirmed via `git stash` — they fail identically with and without this change; they're a known machine-locale `toLocaleString` issue in `AgentModelCombobox`/`ChatTimelineItems`/`ChatWorkspace`).

## Manual verification

- Verified in the real Electron desktop app (dev build on this branch, pointed at real session data) — clicking "Switch agent" from the session actions menu now opens the dialog and it stays open and interactive; a genuine click outside still closes it.
- Verified again against a full self-build of the packaged desktop app + Go daemon (not just the dev harness): this branch merged onto a freshly upstream-synced fork main, `.app` repackaged and daemon restarted, then re-tested live against real session data. Confirmed by the reporting user directly in the rebuilt production app: "все работает, я подтверждаю" ("everything works, confirmed") — the dialog opens and stays open, no more flash-close.

### DIFF
--- a/frontend/src/renderer/components/SwitchAgentDialog.tsx
+++ b/frontend/src/renderer/components/SwitchAgentDialog.tsx
@@ -1,6 +1,6 @@
 import { useQueryClient } from "@tanstack/react-query";
 import { LoaderCircle, Repeat2, TriangleAlert, X } from "lucide-react";
-import { type FormEvent, useEffect, useRef, useState } from "react";
+import { type FormEvent, useEffect, useLayoutEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import {
 	agentSwitchesQueryKey,
@@ -33,6 +33,7 @@ import {
 	DialogTitle,
 } from "./ui/dialog";
 import { Tooltip, TooltipContent, TooltipTrigger } from "./ui/tooltip";
+import { onMenuTeardownComplete } from "./ui/menu-focus";
 
 export const SWITCH_AGENT_OPTIONS = [
 	{ value: "claude-code", label: "Claude Code" },
@@ -57,32 +58,63 @@ function isFromDismissedMenuTrigger(target: EventTarget | null): boolean {
 	return Boolean(target.closest('[role="menuitem"], [role="menu"], [data-session-actions-trigger]'));
 }
 
-// The exemption only covers the opening interaction (the closing dropdown's
-// teardown in the ticks right after mount). After it settles, the menu and
-// trigger are ordinary outside elements again: suppressing outside events
-// from them for the dialog's whole lifetime would swallow later actions-menu
-// interactions and leave keyboard focus stranded outside the non-modal
-// dialog. Two animation frames are enough for the dropdown's same-tick
-// dismissal to have run; anything after that is a genuine outside event.
+// Longest teardown this can still be covering: Radix keeps a closing menu
+// mounted for the 100ms `animate-popover-out` exit, and its FocusScope defers
+// the focus restore one more tick after that. Safety net only — for opens with
+// no menu behind them (the toolbar icon button, an auto-open from a switch
+// error), where no teardown event will ever arrive.
+const OPENING_RACE_FALLBACK_MS = 300;
+
+// The exemption covers the opening interaction only, and only when the dialog
+// was genuinely opened from a menu: it is armed while the caret still sits on
+// the clicked menu item, and stays armed until that menu reports its teardown
+// as complete — content unmounted after the exit animation, deferred focus
+// restore dispatched. Radix keeps a closing menu mounted through its whole
+// exit animation, and only when that ends does its FocusScope restore focus to
+// the trigger, so any time-based window either expires too early (the real
+// renderer) or suppresses too long. After the teardown settles, the menu and
+// trigger are ordinary outside elements again: suppressing outside events from
+// them for the dialog's whole lifetime would swallow later actions-menu
+// interactions and leave keyboard focus stranded outside the non-modal dialog.
 function useSuppressOpeningRace(open: boolean) {
 	const suppressRef = useRef(false);
-	useEffect(() => {
-		if (!open) {
+	const armedMenuRef = useRef<Element | null>(null);
+	// Layout effect on purpose: the dialog's own FocusScope claims the caret
+	// from a passive effect, so by the time an ordinary effect ran, the clicked
+	// menu item would no longer be focused and the open could no longer be
+	// traced back to a menu.
+	useLayoutEffect(() => {
+		const disarm = () => {
 			suppressRef.current = false;
+			armedMenuRef.current = null;
+		};
+		if (!open) {
+			disarm();
 			return;
 		}
+		const active = document.activeElement;
+		if (!(active instanceof Element)) return;
+		const armed = active.closest('[role="menu"]');
+		if (!armed || !active.closest('[role="menuitem"], [role="menu"]')) return;
+		armedMenuRef.current = armed;
 		suppressRef.current = true;
-		let frames = 0;
-		let rafId = 0;
-		const tick = () => {
-			if (++frames >= 2) {
-				suppressRef.current = false;
-				return;
-			}
-			rafId = requestAnimationFrame(tick);
+		const unsubscribe = onMenuTeardownComplete(({ menu }) => {
+			const current = armedMenuRef.current;
+			if (current === null) return;
+			// Either the menu that opened this dialog finished its teardown, or it
+			// was replaced by another menu the user opened over the dialog and the
+			// armed one is already gone. Both mean the opening interaction is over.
+			if (menu !== current && current.isConnected) return;
+			// Microtask, not synchronous: an unprevented restore focuses the
+			// trigger inside this same macrotask, right after the dispatch, and
+			// that focusin is exactly what the guard exists to swallow.
+			queueMicrotask(disarm);
+		});
+		const fallback = window.setTimeout(disarm, OPENING_RACE_FALLBACK_MS);
+		return () => {
+			unsubscribe();
+			window.clearTimeout(fallback);
 		};
-		rafId = requestAnimationFrame(tick);
-		return () => cancelAnimationFrame(rafId);
 	}, [open]);
 	return suppressRef;
 }
--- a/frontend/src/renderer/components/SwitchAgentDialog.tsx
+++ b/frontend/src/renderer/components/SwitchAgentDialog.tsx
@@ -45,6 +45,18 @@ export function canSwitchAgentHarness(value: string): value is SwitchAgentHarnes
 	return SWITCH_AGENT_OPTIONS.some((option) => option.value === value);
 }
 
+// SwitchAgentDialog is opened from a DropdownMenuItem ("Switch agent" in the
+// session actions menu). Radix closes that dropdown on the same click that
+// opens this dialog; since the dialog is non-modal (see below), its
+// DismissableLayer would otherwise treat the dropdown's residual pointer/
+// focus activity as an outside interaction and dismiss the dialog right
+// after it opens. Ignore only outside events that originate from the
+// just-dismissed menu/trigger so a genuine outside click still closes it.
+function isFromDismissedMenuTrigger(target: EventTarget | null): boolean {
+	if (!(target instanceof Element)) return false;
+	return Boolean(target.closest('[role="menuitem"], [role="menu"], [data-session-actions-trigger]'));
+}
+
 function SwitchTargetPicker({
 	currentHarness,
 	disabled,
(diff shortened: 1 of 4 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/components/SwitchAgentDialog.tsx
```diff
@@ -44,6 +44,18 @@ export function canSwitchAgentHarness(value: string): value is SwitchAgentHarnes
...
 }
 
+// SwitchAgentDialog is opened from a DropdownMenuItem ("Switch agent" in the
+// session actions menu). Radix closes that dropdown on the same click that
+// opens this dialog; since the dialog is non-modal (see below), its
+// DismissableLayer would otherwise treat the dropdown's residual pointer/
+// focus activity as an outside interaction and dismiss the dialog right
+// after it opens. Ignore only outside events that originate from the
+// just-dismissed menu/trigger so a genuine outside click still closes it.
+function isFromDismissedMenuTrigger(target: EventTarget | null): boolean {
```
whoisasx, line 56: [P2] Please make this dismissal exception one-shot. This predicate remains active for the dialog’s entire lifetime and broadly exempts every menu, menu item, and session-actions trigger. Since the top-bar trigger sits outside the dialog portal and scrim, a later genuine click on it is also ignored by the dialog, allowing the actions menu (and potentially another action/dialog) to open over the switch-agent dialog. The focus guard can likewise leave keyboard focus on that external trigger because preventing the custom focus-out dismissal does not move focus back into this non-modal dialog. Please scope the exception to the original opening cycle, or defer opening until the dropdown has completed closing and restoring focus. Add coverage that focus lands inside the opened dialog and that a later click on Session actions dismisses it instead of layering the menu over it.

whoisasx (changes requested):
The PR correctly identifies the flash-close race, stays narrowly scoped, and adds a useful regression harness that exercises the real dropdown-menu path. All GitHub checks are green and the branch is mergeable. I am holding approval on one interaction issue: the selector-based dismissal exemption remains active for the entire dialog lifetime, so later actions-menu interactions can be suppressed or layered over the dialog, and keyboard focus can remain outside the non-modal dialog. Please scope the exception to the opening interaction and extend the regression coverage for focus placement and a later trigger click. With that addressed, this should be ready.

axisrow:
Addressed in 7cbe0c4a9: the dismissal exemption is now armed only for the opening interaction and disarmed two animation frames later (once the dismissed dropdown's teardown has run), instead of staying active for the dialog's lifetime. Added regression coverage for both cases you asked about — a later click on the session-actions trigger and a later outside focus change both dismiss the dialog (the focus case failed red against the previous lifetime exemption). All checks green.

axisrow:
Follow-up in ca01a3f3d: added the missing focus coverage — driving the real menu-item click path and asserting that after the opening focus race settles, `document.activeElement` lives inside the opened non-modal dialog and is not the outside session-actions trigger. It passes as-is on the rebased base (the menu focus handoff in the DropdownMenu wrapper keeps the dialog's FocusScope focus in place), so this is coverage-only — no production code touched. The branch is also rebased onto current main.

axisrow:
@whoisasx friendly bump

axisrow:
@whoisasx friendly bump

frontend/src/renderer/components/SwitchAgentDialog.tsx
```diff
@@ -45,6 +45,48 @@ export function canSwitchAgentHarness(value: string): value is SwitchAgentHarnes
...
+	if (!(target instanceof Element)) return false;
+	return Boolean(target.closest('[role="menuitem"], [role="menu"], [data-session-actions-trigger]'));
+}
+
+// The exemption only covers the opening interaction (the closing dropdown's
+// teardown in the ticks right after mount). After it settles, the menu and
+// trigger are ordinary outside elements again: suppressing outside events
+// from them for the dialog's whole lifetime would swallow later actions-menu
+// interactions and leave keyboard focus stranded outside the non-modal
+// dialog. Two animation frames are enough for the dropdown's same-tick
```
whoisasx: [P1] Two animation frames do not cover the actual dropdown teardown. `DropdownMenuContent` uses the 100 ms `animate-popover-out` exit, and Radix Presence keeps the menu mounted until that animation ends; its FocusScope runs `onCloseAutoFocus` and restores focus to the external session-actions trigger only when it finally unmounts. At that point this ref has already been cleared after roughly 16–33 ms, so `onFocusOutside` can dismiss the dialog again in the real Electron renderer. JSDOM has no real CSS animation and therefore unmounts the menu immediately, which is why CI stays green. The two 150 ms tests can also false-pass because they do not assert that the dialog is still open after the wait, while the focus test checks after only two frames. Please coordinate with the dropdown close lifecycle or consume the specific delayed restoration event once, and assert that the dialog remains present after the 100 ms exit completes.

whoisasx (changes requested):
Thanks for addressing the previous lifetime-exemption feedback and adding the requested trigger and focus coverage. The direction is better, and all GitHub checks are green. One production-timing blocker remains: the new guard expires after two animation frames, but the dropdown is retained by Radix Presence through its 100 ms exit animation and restores focus only when that animated content unmounts. JSDOM skips that lifecycle, and the delayed tests do not first prove that the dialog survived the wait, so they can pass even if the original flash-close returns in Electron. Please tie suppression to the actual close/focus-restoration event and cover the post-animation state. Once that is addressed, this should be ready.

frontend/src/renderer/components/SwitchAgentDialog.tsx
axisrow: Reworked in 43f302080 + 0b95f7c5e: suppression is no longer time-based.

- `composeMenuCloseAutoFocus` — the code already sitting on the exact delayed restoration event — now publishes a one-shot "menu teardown complete" (`menu` node + `trigger`) after the FocusScope restore has been decided (allowed or prevented). The dialog subscribes and disarms via `queueMicrotask`, so an unprevented restore's `focusin` lands while still suppressed (restore runs synchronously after the dispatch in the same macrotask). A 300 ms fallback timer remains only for opens with no menu behind them (icon variant, switch-error auto-open), and arming is now gated on the caret still being on the clicked menu item at open (layout effect), so those paths arm nothing at all.
- The new regression test reproduces the real-Electron timeline: a scoped `getComputedStyle` stub makes Radix Presence retain the closing menu for the `animate-popover-out` exit (asserted: menu still mounted with `data-state="closed"`, focus already in the dialog), then `animationend` unmounts it and runs the deferred restore — and the test asserts the dialog survived that teardown with focus still inside. It fails against the previous two-frame guard.
- The "later trigger click" / "later focus move" tests now settle on the real teardown event instead of a 150 ms sleep, and assert the dialog is still present after the settle before interacting — no more false-passes. Unit tests for the signal added in `menu-focus.test.tsx`.

axisrow:
@whoisasx friendly bump

whoisasx (approved):
Thanks for iterating on the earlier review feedback. I re-reviewed exact head 0b95f7c5e86a2afc7b8e727c6beb66d8a442c12b, and both previous dismissal-lifecycle concerns are addressed: suppression is tied to the actual menu teardown and deferred focus restoration, the dialog retains focus through the exit animation, and later trigger clicks, outside clicks, and outside focus changes dismiss it normally. The branch merges cleanly with current main; 42 focused frontend tests and the frontend typecheck passed against that synthetic merge, and GitHub CI is green. No Go tests were run, per instruction. Approved from my side.

axisrow:
@whoisasx friendly bump

axisrow:
Merged clean — thanks @Rishet11 for the quick turnaround! 🙌

CI went 14/14 green on the final head, and the fix is already living happily in our daily build.

PR #4712, opened 2026-08-31
fix(terminal): resolve reboot reattach loop and add native auto-resume (#4641)

### LINKED ISSUE #4641: Orchestrator killed by tmux/host death has no path back to its own history — only recovery action is a cold, unlinked respawn
## Bug

When a project orchestrator's TUI runtime dies (host reboot, OS kill, tmux server loss), the only recovery actions reachable today discard its conversation history — even though that history is fully intact on disk (confirmed live: a 1.7MB Claude Code transcript survives untouched after the reboot) and the daemon already has a working resume mechanism (`RestoreWithMode` → `GetRestoreCommand` → `RestoreIdentity`) that's simply never invoked by the orchestrator-recovery flow. `SpawnOrchestrator(clean=true)` always spawns a brand-new session with no `AgentSessionID` carried forward from the retired one, and neither the topbar "Orchestrator" button nor `RestoreUnavailableDialog` ever attempt a real restore before discarding.

Basically when I swtiched to  the terminal to talk to the orchastrator  and restarted my pc,the tmux session is cleared  and with that my chat was also gone,so I couldnt go back to the original orchastrator chat

## Reproduction

1. Run a project orchestrator (TUI, Claude Code) with real conversation history.
2. Reboot the host without a graceful `POST /shutdown` (kills the tmux server backing the orchestrator's PTY).
3. On next boot the orchestrator is still `isTerminated: false` but permanently unattachable.
4. `Kill` / clean-respawn (`POST /api/v1/orchestrators {"clean":true}`) both 500 today — tracked in #4481/PR #4482.
5. Even once that's fixed: the only recovery affordances (topbar "Orchestrator" button, `RestoreUnavailableDialog`'s "create new orchestrator") call `spawnOrchestrator(..., clean=true)`, which always creates a new session/conversation id — never attempts `POST /sessions/{id}/restore` against the just-retired orchestrator.

## Acceptance Criteria

- [ ] After a dead-runtime orchestrator is retired via clean respawn, AO attempts to resume its native conversation (the existing `RestoreWithMode`/`GetRestoreCommand` mechanism) before falling back to a fresh spawn.
- [ ] If resume succeeds, the user lands back in the same conversation with full history intact.
- [ ] If resume genuinely fails, the user is explicitly told the respawn started cold (extending the existing `orchestratorRetireNotice` mechanism) instead of silently swapping in a blank coordinator.
- [ ] A "restore instead of replace" action is reachable from the actual orchestrator-recovery UI (topbar button / `RestoreUnavailableDialog`), not only from the Archived Sessions list.

## Related

- #4481 / PR #4482 — prerequisite mechanical fix (Kill/RetireForReplacement 500 on dead runtime).
- #2501 — same underlying gap (orchestrator respawn with no linkage/notification), different trigger (probe race vs. host-level runtime death).

Hendrixx-RE:
## Summary

Fixes #4641.

This PR addresses two related issues with terminal session persistence and lifecycle:

### 1. Bug Fix: Infinite reattach loop after PC reboot
- **Root Cause**: In `tmux.go`, `IsAlive` classified `serverNotRunningOutput` ("no server running") as `ports.ErrRuntimeUnavailable` (a retryable/transient error) rather than `(false, nil)`. Because it was treated as transient, `attachment.go` retried with linear backoff, but the frontend's 3-second open timeout tore down the socket and reconnected first, resetting the backend's retry counter. This resulted in two retry loops perpetually interrupting each other in an infinite "reattaching" state instead of reaching a clean "session ended" state.
- **Fix**: Reclassified "no server running" as permanent death `(false, nil)`, identical to the "session not found" case. `attachment.go` immediately marks the session as exited. (Darwin native PTY host already uses this classification via `isConnRefused`).
- **Tests**: Updated `TestIsAliveReportsNoServerAsPermanentDeath` and integration test expectations.

### 2. Feature: Auto-resume terminal-only agents into fresh session
- **Behavior**: When a user opens/interacts with a confirmed permanently dead terminal session (`state === "exited"`):
  1. Calls `POST /api/v1/sessions/{sessionId}/auto-resume-agent` (`AutoResumeAgentWithMode`), which checks if the agent adapter's `GetRestoreCommand` returns `ok=true` (supported for agents like Cline, Claude Code, Goose, etc. that persist native conversation IDs on disk).
  2. If `ok=true`: Launches a fresh terminal process with native resume argv, attaches transparently, and displays a one-line banner: `"Session expired after restart — resumed via <agent>'s native session recovery"`.
  3. If `ok=false`: Returns `SESSION_NO_NATIVE_RESUME` (mapped to HTTP 409), falling back cleanly to the existing "session ended" state with no silent relaunch.
  4. Gated strictly on definitively-dead terminal state, never on transient errors.

## Test Coverage
- `TestIsAliveReportsNoServerAsPermanentDeath` in `tmux_test.go`
- `TestAutoResumeAgent_SucceedsForNativeRestore` in `manager_test.go`
- `TestAutoResumeAgent_FailsWithErrNoNativeResumeWhenNotResumable` in `manager_test.go`
- `TestAutoResumeAgent_RequiresLiveExitedSession` in `manager_test.go`
- `autoResumeAgent` endpoint test in `sessions_test.go`
- Full backend suite passing (`go test ./...`)

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -525,11 +525,12 @@ func (r *Runtime) paneSessionIDs(ctx context.Context, id string) []int {
 
 // IsAlive reports whether the handle's session still exists via `tmux
 // has-session`. Exit 0 means alive. A non-zero exit with output naming this
-// session as missing is a definitive false, nil. A conclusively absent server
-// wraps ports.ErrRuntimeUnavailable so recovery may recreate it. A transient
-// connection or protocol/client failure wraps ErrRuntimeProbeInconclusive so
-// no caller can treat a possibly-live session as absent. Any other non-zero
-// exit is a plain probe error, which is likewise never per-session death.
+// session as missing OR a conclusively absent server is a definitive (false,
+// nil) — after a reboot the tmux server is gone and every session it hosted
+// is permanently dead. A transient connection or protocol/client failure wraps
+// ErrRuntimeProbeInconclusive so no caller can treat a possibly-live session
+// as absent. Any other non-zero exit is a plain probe error, which is likewise
+// never per-session death.
 func (r *Runtime) IsAlive(ctx context.Context, handle ports.RuntimeHandle) (bool, error) {
 	id, err := handleID(handle)
 	if err != nil {
@@ -539,12 +540,8 @@ func (r *Runtime) IsAlive(ctx context.Context, handle ports.RuntimeHandle) (bool
 	if err != nil {
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) {
-			if sessionMissingOutput(string(out)) {
+			if sessionMissingOutput(string(out)) || serverNotRunningOutput(string(out)) {
 				return false, nil
-			}
-			if serverNotRunningOutput(string(out)) {
-				return false, fmt.Errorf("tmux runtime: probe session %s: %w: %s",
-					id, ports.ErrRuntimeUnavailable, strings.TrimSpace(string(out)))
 			}
 			if transientServerFailureOutput(string(out)) {
 				return false, fmt.Errorf("tmux runtime: probe session %s: %w: %s",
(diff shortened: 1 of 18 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

PR #4713, opened 2026-08-31
fix: disable chat minimap interaction when inspector is open

Annieeeee11:
## Summary
- Disable the chat conversation minimap entirely while the right inspector panel is open, not just hide the turn markers.
- Add a `minimapEnabled` gate so the scroll rail is hidden, non-interactive, and removed from the accessibility tree when the inspector is expanded.
- Clear hover state and guard pointer/wheel/keyboard handlers so the chat edge no longer steals interaction from the open panel.

## Test plan
- [ ] Open a chat session with a long conversation history
- [ ] With the right inspector panel **open**, hover/click/drag on the right edge of chat and confirm nothing scrolls or jumps between turns
- [ ] Collapse the inspector panel and confirm the minimap dashes appear on hover and turn navigation works again
- [ ] Run `npx vitest run src/renderer/components/chat/ChatWorkspace.test.tsx -t "conversation minimap"`

### DIFF
--- a/frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
+++ b/frontend/src/renderer/components/chat/ChatWorkspace.test.tsx
@@ -1025,15 +1025,13 @@ describe("ChatWorkspace timeline", () => {
 		expect(scrollbar).toHaveAttribute("aria-valuenow", "100");
 	});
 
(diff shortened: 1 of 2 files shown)


ronishrohan (approved):
Everything is fine, LGTM

PR #4714, opened 2026-08-31
Fix/chat minimap disable when inspector open

Annieeeee11:
## Follow-up

- Clear `drag.current` when the inspector opens so an in-progress minimap drag cannot linger after the rail is disabled.
- Add a toggle test: minimap works with inspector closed → disabled when opened → works again after collapse.

## Test plan
- [ ] Run `npx vitest run src/renderer/components/chat/ChatWorkspace.test.tsx -t "conversation minimap"`
- [ ] Confirm all 4 minimap tests pass


ronishrohan (approved):
Everything good, LGTM

PR #4715, opened 2026-08-31
feat: add drag-to-reorder for queued chat messages

Annieeeee11:
## Summary
- Add POST /api/v1/sessions/{sessionId}/conversation/queue/reorder to persist queued turn order by permuting existing requested_at values (no migration)
- Add drag-to-reorder in the queue dock with @dnd-kit, a right-side grip handle, and optimistic conversation cache updates for smooth drops
- Suppress hover Steer briefly after drag/drop to prevent a visible Steer blink

## Checklist

- [ ] Queue 3 messages while a turn is running, expand the dock, and drag rows by the grip handle
- [ ] Confirm order persists after drop and the next-up item dispatches in the new order
- [ ] Confirm no full-queue flash or Steer blink after drop
- [ ] cd backend && go test ./internal/storage/sqlite/store/... ./internal/httpd/controllers/...
- [ ] cd frontend && npx vitest run src/renderer/components/chat/ChatSteer.test.tsx

https://github.com/user-attachments/assets/bc82428a-9be8-48e3-9848-cd40b3dc1962


ronishrohan (commented):
P2 / Test gap

The new drag path is not actually tested. The frontend test only asserts that drag handles render (`frontend/src/renderer/components/chat/ChatSteer.test.tsx:230-233`); it does not simulate a drag, verify the FIFO order sent to `onReorderQueuedTurns`, or cover rollback on API failure. The backend route test only covers a successful request.

Please add coverage for:
- dragging first/last rows and asserting the submitted order;
- failed reorder restoring the visual order and showing the error;
- empty/invalid backend order returning 400;
- a queued turn becoming unavailable during reorder returning 409.

ronishrohan:
https://github.com/user-attachments/assets/9c8d0400-7fa9-4f6b-9bf8-aff7f0ff595e

@Annieeeee11 take a look at how session rows on the left handle drag and what limits they have (since both are in dynamic parent containers), 

another thing is when something is being dragged make sure its children stay static, for example some stuff only shows up on hover and when you drag the mouse might leave the container sometimes when you drag too quick, the projects sidebar fixes this too, so check that

ronishrohan (changes requested):
Request changes: the merge commit accidentally deletes all `ao-qr-*` animation definitions from `frontend/src/renderer/styles.css`, while `frontend/src/renderer/components/settings/PairingQr.tsx:75-95` still uses those classes. This breaks the Connect Mobile waiting QR scramble/drift/settling behavior and can leave the overlapping decoy QR layers unreadable.

Please restore the deleted QR keyframes and class rules from `origin/main`, then rerun the Pairing QR tests and a focused visual check of the Connect Mobile flow.

ronishrohan (approved):
Approved. The QR animation regression remains documented in the earlier review comment and should be addressed before merge.

PR #4717, opened 2026-08-31
fix(lifecycle): carry the resolved model through mergeMetadata

### LINKED ISSUE #4716: [Bug]: `ao spawn --model X` records an empty model — mergeMetadata drops Metadata.Model
### Summary

`ao spawn --model <model>` succeeds, but the session records no model at all. The `sessions.model` column is written as the empty string and `GET /api/v1/sessions` omits the field, so nothing can report which model a session resolved to.

The agent itself launches correctly — this is a record/read-back defect, not a launch defect.

### Steps to reproduce

```bash
export AO_DATA_DIR=/tmp/ao-model AO_RUN_FILE=/tmp/ao-model/running.json AO_PORT=3095
mkdir -p "$AO_DATA_DIR"
cd backend && go run ./cmd/ao daemon
```

In a second shell, register any git repo with a resolvable default branch and spawn with an explicit model:

```bash
cd backend
go run ./cmd/ao project add --path /path/to/a/git/repo --id demo --name demo
go run ./cmd/ao spawn --project demo --harness claude-code --skip-agent-check \
  --mode tui --name m1 --model sonnet

curl -s http://127.0.0.1:3095/api/v1/sessions | grep -o '"model":"[^"]*"'   # no output
sqlite3 "$AO_DATA_DIR/ao.db" "SELECT id, quote(model) FROM sessions;"
```

### Expected behavior

The resolved model is recorded and read back:

```
demo-1|'sonnet'
"model":"sonnet"
```

`domain.SessionMetadata.Model` documents exactly this — *"the agent model this session resolved to at spawn time, including any per-spawn --model override"* — and the schema carries a dedicated `model TEXT NOT NULL DEFAULT ''` column for it.

### Actual behavior

```
demo-1|''
```

and the API response has no `model` key.

### Root cause

`session_manager` builds the spawn metadata with `Model: resolvedModelForMetadata(...)` and hands it to `MarkSpawned`, which merges it into the existing row:

```go
rec.Metadata = mergeMetadata(rec.Metadata, metadata)
```

`seedRecord` never sets `Model`, so the base side is empty. `mergeMetadata` (`backend/internal/lifecycle/manager.go`) copies seventeen fields by an explicit allowlist and `Model` is not among them, so the empty base value wins and the resolved model is discarded.

This is the failure mode the package already documents. `TestMarkSpawnedPersistsChatControllerFacts` carries this comment:

> mergeMetadata is an explicit allowlist, so a field added to SessionMetadata without a line here is silently dropped on every spawn and restore. That happened to the chat resume handle: the provider still held the conversation, but AO forgot its id, so no restart could ever resume it — and nothing failed loudly, the column was just empty.

`Model` is the same omission one field over.

### Scope

Only two things read `Metadata.Model` — the session DTO (`httpd/controllers/sessions.go`) and the `sessions.model` column (`storage/sqlite/store/session_store.go`). No relaunch or resume path consumes it, so no session has been running the wrong model; the value is simply never recorded.

### AO version

Built from source at `dbd87a21e` (current `main`).

### Environment

Linux 7.1.8-arch1-3, x86_64, Go 1.25.12, tmux 3.7c.

### Logs / extra context

Found by running the daemon against an isolated data dir and cross-referencing the observed behaviour with the code path. Happy to open a PR — the fix is one line in the copy-if-non-empty group, plus a regression test.

yashksaini-coder:
## What

Adds the missing `Model` line to `mergeMetadata`, so a spawn's resolved model is recorded instead of discarded.

Fixes #4716.

## Why

`ao spawn --model X` persists an empty model. `sessions.model` is written as `''` and `GET /api/v1/sessions` omits the field, so nothing can report which model a session resolved to.

`session_manager` builds the spawn metadata with `Model: resolvedModelForMetadata(...)` and hands it to `MarkSpawned`, which does `rec.Metadata = mergeMetadata(rec.Metadata, metadata)`. `seedRecord` never sets `Model`, so the base side is empty; `mergeMetadata` copies seventeen fields by explicit allowlist and `Model` is not one of them, so the empty base wins.

This is the hazard the package already warns about. `TestMarkSpawnedPersistsChatControllerFacts` says it plainly:

> mergeMetadata is an explicit allowlist, so a field added to SessionMetadata without a line here is silently dropped on every spawn and restore. That happened to the chat resume handle ... nothing failed loudly, the column was just empty.

`Model` is the same omission one field over.

## How

One line, in the `set` (copy-if-non-empty) group rather than an unconditional assignment: a relaunch that resolves no explicit model must leave the recorded value alone. That is the opposite of the neighbouring `ControllerGeneration`, which is assigned precisely because rotation is the point — the test covers both directions.

**Scope — deliberately narrow.** This is a record/read-back fix, not a launch fix. The agent already started with the correct model; `session_manager` passes the adapter its own resolved config, and `Metadata.Model` exists for the API view. Its only readers are the session DTO (`httpd/controllers/sessions.go`) and the `sessions.model` column (`storage/sqlite/store/session_store.go`), so no session has been running the wrong model — AO just forgot which one it picked.

I have not audited the rest of `SessionMetadata` for further allowlist gaps in this PR. If you would like that swept as a follow-up, say so and I will do it separately rather than widen this one.

## Testing

`TestMarkSpawnedPersistsResolvedModel` asserts both halves: the resolved model reaches storage, and a later relaunch resolving none leaves it intact. It is committed **before** the fix so the failure is in the history:

```
manager_test.go:3711: model = "", want "sonnet"; a spawn's resolved model must survive the merge
```

Verified against a real daemon on an isolated data dir, before and after:

```
ao spawn --project demo --harness claude-code --model sonnet

before   sessions.model = ''          GET /api/v1/sessions -> model absent
after    sessions.model = 'sonnet'    GET /api/v1/sessions -> model='sonnet'
```

`go test -race ./internal/lifecycle/...` passes. Full backend suite: 161 packages pass. Three failures on my machine are pre-existing and unrelated to this change, all reproducing on a clean `main` checkout — `TestCopilotClassicPATIsNotAnAuthorizationSignal` (fails whenever a classic GitHub PAT is in the environment) and `TestRuntimeIntegrationSupervisedExitKeepsInteractiveShell` / `TestAttachmentStreamsRealTmuxPane` (both `base-index`, #4222).

## Checklist

- [x] Branched from `main`
- [x] Tests added for user-visible behavior

---

https://github.com/user-attachments/assets/c83bac8e-dd27-41ee-b378-3c0c8e16c0fd


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

illegalcall (approved):

PR #4719, opened 2026-08-31
perf: reduce renderer interaction overhead

ronishrohan:
## Summary
- isolate renderer subscriptions and sidebar drag updates to avoid broad React work
- remove forced composer suggestion layouts and hover-transition paint churn
- make shell terminal creation and tab switching feel immediate
- unify sidebar context-menu styling with dropdown menus

## Validation
- `npm run typecheck`
- focused renderer tests (149 passing)
- `npm test` is blocked by unrelated landing-site test setup failures (Android download tests and markdown-twin generation)

### DIFF
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -600,6 +600,7 @@ export function TerminalPane({
 	inputDisabled,
 	focusRequested,
 }: TerminalPaneProps) {
+	const { t } = useTranslation();
 	const terminalTarget =
 		requestedTerminalTarget &&
 		terminalTargetBelongsToSession(requestedTerminalTarget, session?.id)
@@ -668,7 +669,13 @@ export function TerminalPane({
 	// rather than surfacing a separate loading state or attaching xterm to a
 	// handle that cannot exist yet.
 	if (isOptimisticShell) {
-		return <div aria-label="Terminal" className="terminal-surface h-full" data-testid="optimistic-terminal" />;
+		return (
+			<div
+				aria-label={t("terminal.shellAria")}
+				className="terminal-surface h-full"
+				data-testid="optimistic-terminal"
+			/>
+		);
 	}
 
 	const props = {
--- a/frontend/src/renderer/components/TerminalPane.tsx
+++ b/frontend/src/renderer/components/TerminalPane.tsx
@@ -70,7 +70,7 @@ type TerminalCacheDescriptor = {
 
 type CachedTerminalEntry = TerminalCacheDescriptor & {
 	activationId: number;
-	activationPhase: "parked" | "preparing" | "ready" | "revealed" | "visible";
+	activationPhase: "parked" | "visible";
 	container: HTMLDivElement;
 	discardOnDeactivate?: boolean;
 	props: TerminalPaneProps;
@@ -169,7 +169,6 @@ function setTerminalPhase(
 	entry.activationPhase = phase;
 	entry.container.dataset.terminalActivationPhase = phase;
 	const interactive = phase === "visible";
-	const rendered = phase === "revealed" || interactive;
 	entry.container.inert = !interactive;
 	if (interactive) {
 		entry.container.removeAttribute("aria-hidden");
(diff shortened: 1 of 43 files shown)


somewherelostt:

- @neversettle17-101
- @illegalcall

Join the session here:
https://discord.gg/H6ZDcUXmq

frontend/src/renderer/styles.css
```diff
@@ -4719,3 +4719,23 @@ html[data-native-browser-composition="true"] .browser-popout-backdrop {
...
+:where(button, a[href], [role="button"], [role="menuitem"], [role="option"]):active {
+	transition-property: transform, scale !important;
+}
+
+:where(button, a[href], [role="button"], [role="menuitem"], [role="option"]) :where(
+		svg,
+		[class*="transition"]
+	) {
+	transition-property: none !important;
+}
```
Annieeeee11: **Blocker:** this is a global interaction change, not a local perf tweak.

`:where(button, a[href], [role="button"], [role="menuitem"], [role="option"])` plus `!important` turns off hover/opacity/color transitions for every action in the app. The descendant rule also matches `[class*="transition"]`, so it disables the folder/chevron `transition-[opacity,transform]` this PR just added on sidebar rows.

Please drop the global override. If hover paint is the problem, cut `transition-[background,color]` on the specific hot controls (sidebar rows, tab close, composer buttons) instead of a design-system-wide `!important` hammer.

frontend/src/renderer/components/TerminalPane.tsx
```diff
@@ -700,6 +605,11 @@ export function TerminalPane({
 		terminalTargetBelongsToSession(requestedTerminalTarget, session?.id)
 			? requestedTerminalTarget
 			: ({ kind: "worker" } satisfies TerminalTarget);
+	const shellTerminals = useShellTerminals().data ?? [];
+	const isOptimisticShell = Boolean(
+		terminalTarget.kind === "shell" &&
+			shellTerminals.some((shell) => shell.handleId === terminalTarget.handleId && shell.optimistic),
+	);
```
Annieeeee11: **Blocker:** this can miss the optimistic window and attach xterm to `pending-shell:…`.

`isOptimisticShell` is true only while the shell-terminals cache still has this `handleId` with `optimistic: true`. On success, `useOpenShellTerminal` replaces that row with the real handle (`useShellTerminals.ts` `onSuccess`) *before* `SessionView.addShellTerminal` updates `terminalTarget` off the pending id.

Query cache updates are an external store. One render can therefore see:

- `terminalTarget.handleId === "pending-shell:…"`
- cache row already swapped to the real handle

`isOptimisticShell` is then false, and we attach to a handle that does not exist.

Treat `pending-shell:` as optimistic regardless of cache shape (or keep the pending row until `terminalTarget` has moved). Also worth putting the selected-handle swap in the same update as the list replacement so the tab and pane cannot diverge.

frontend/src/renderer/components/chat/ComposerSuggestMenu.tsx
```diff
@@ -80,7 +88,7 @@ export function ComposerSuggestMenu({
 		const observer = new ResizeObserver(updateScrollIndicators);
 		observer.observe(node);
 		return () => observer.disconnect();
-	}, [items, updateScrollIndicators]);
+	}, [kind, updateScrollIndicators]);
```
Annieeeee11: **Blocker:** filtering can leave the highlighted row and scroll fades stale.

The index-0 skip is the right idea for typed `/` and `@` (avoid `scrollIntoView` on every character). Removing `items` from these effects goes too far:

- If `highlighted` stays at the same index while results re-rank, this effect does not run, so a keyboard selection can sit off-screen.
- The indicator effect now depends on `kind` instead of `items`. Filtering usually changes content height without changing `kind` or the list box size, so the top/bottom fades can be wrong until a later resize.

Keep the index-0 early return, but restore `items` (or an identity/length that changes when the result set changes) on both effects.

Annieeeee11 (changes requested):
## Blockers

Three issues that should be fixed before merge:

1. **Global `transition-property: none !important`** — design-system / interaction change, and it also disables the folder/chevron CSS this PR added.
2. **Optimistic shell attach race** — cache replacement can drop `optimistic: true` before `terminalTarget` switches off `pending-shell:…`.
3. **Composer suggest scroll** — `items` was removed from effect deps, so filtered `/` and `@` lists can leave the highlight and fade indicators stale.

The subscription/`select` split and sidebar drag isolation look good; these three are the ones I would not land as-is.

Annieeeee11 (approved):
lgtm

PR #4482, opened 2026-08-26
fix(session): make kill idempotent when runtime is already gone

joeythepepe:
## What

Make kill, resume-agent, and orchestrator replacement succeed when the TUI runtime is already gone (missing tmux, stale private socket, missing default socket, inconclusive legacy probe, failed native terminate).

Fixes [#4481](https://github.com/Untrivial-ai/agent-orchestrator/issues/4481).

## Why

After a daemon restart, a project orchestrator (and other TUI sessions) whose tmux server is gone cannot be talked to, resumed, closed, or replaced:

- `POST /sessions/{id}/send` → 409 `AGENT_EXITED`
- `POST /sessions/{id}/resume-agent` → 500 `INTERNAL_ERROR` (fingerprint `f05be741f335dfad`)
- `POST /sessions/{id}/kill` → 500 `INTERNAL_ERROR` (fingerprint `aad9192f2c617588`)
- Spawn orchestrator without `clean` returns the zombie; restart (`clean=true`) hits the same Destroy 500.

Live repro: `ao-killfix-2` (grok TUI orchestrator, handle `ao-killfix-2`, no `ptyhost-v1:` prefix). Private socket `/tmp/tmux-501/ao` had no server; default socket was `error connecting ... (No such file or directory)`. `socketForSession` classified that as `ErrRuntimeProbeInconclusive`, so resume/kill/replace all 500ed.

## How

- tmux `socketForSession`: a missing legacy default socket (`error connecting ... no such file`) is both-sockets-gone, same as `no server running`. `IsAlive` then reports `ErrRuntimeUnavailable` instead of inconclusive. Transient `connection refused` is unchanged.
- tmux `Destroy` treats `ErrRuntimeProbeInconclusive` / `ErrRuntimeUnavailable` as already-gone.
- `Kill` and `RetireForReplacement` log and continue on native-terminate failure and already-gone runtime Destroy, then still mark terminated.
- User-initiated `restartRuntime` (resume-agent) treats those runtime sentinels as "no existing runtime" and creates a replacement. Boot `reconcileLive` still refuses to recreate on `ErrRuntimeProbeInconclusive`.

## Testing

```bash
cd backend && env -u AO_TMUX_SOCKET_NAME go test ./internal/session_manager ./internal/adapters/runtime/tmux
```

Added/updated:

- `TestIsAliveReportsPrivateServerGoneAndLegacySocketAbsentAsUnavailable`
[...]

PR #4721, opened 2026-08-31
fix: resume dead orchestrator's conversation before cold respawn

### LINKED ISSUE #4641: Orchestrator killed by tmux/host death has no path back to its own history — only recovery action is a cold, unlinked respawn

Hendrixx-RE:
## Summary
- `attachment.go`: `ErrRuntimeUnavailable` (e.g. tmux's "no server running" after a reboot) is a definitive, permanent signal, but the reattach loop routed it through the same failures/backoff counter as an ordinary transient probe error. That raced against the frontend's ~3s open timeout — which tears down and reopens the attachment on every timeout, resetting the counter to zero — producing the infinite "reattaching" loop reported in this issue. It now reports a clean exit on the very first probe.
- `service.go`: `SpawnOrchestrator(clean=true)` always spawned a brand-new, unlinked session, discarding a retired orchestrator's conversation even though its history is fully intact on disk and the daemon already has a working resume mechanism (`RestoreWithMode` -> `restoreArgv` -> `GetRestoreCommand`). It now attempts that resume for a single retired orchestrator before falling back to a fresh spawn, and sends an explicit notice into the new session when resume genuinely fails, instead of silently swapping in a blank coordinator.
- **`tmux.go` (root cause, found via live reproduction while testing this branch):** on a real host reboot, `/tmp` is commonly tmpfs, so the *entire* per-user tmux socket directory disappears, not just one socket file. tmux then reports `error connecting to <path> (No such file or directory)` instead of `no server running`. `IsAlive` only recognized the latter as `ErrRuntimeUnavailable`; the reboot message fell through to the broader transient `error connecting` match and was reported as `ErrRuntimeProbeInconclusive` instead. Since an inconclusive probe is deliberately never treated as proof of death (to avoid regressing #3475), this left `reconcileLivePass` **permanently unable to ever recognize a session as dead after a real reboot** — it just stays `idle` forever with no live runtime and no automatic resume attempt, regardless of harness. This is very likely the actual, dominant cause of "sessions never come back after a restart" on Linux.
- No frontend changes needed: the topbar "Orchestrator" button and `RestoreUnavailableDialog` both already call `SpawnOrchestrator(clean=true)`, so the resume path is reachable from the existing recovery UI.

basically linux tmpfs is ram based so its sessions were getting cleared everytime I fixed it

Fixes #4641.

## Test plan
- [x] `go test ./backend/internal/terminal/...` — `TestAttachmentExitsImmediatelyOnRuntimeUnavailable` pins the reattach-loop fix
- [x] `go test ./backend/internal/service/session/...` — `TestSpawnOrchestratorCleanResumesRetiredOrchestrator`, `TestSpawnOrchestratorCleanFallsBackWhenResumeFails`, updated `TestSpawnOrchestratorCleanRetireNoticeIsBranchNeutral`
- [x] `go test ./backend/internal/adapters/runtime/tmux/... -tags integration` — `TestIsAliveReportsMissingSocketDirectoryAsRuntimeUnavailable` pins the reboot-message classification fix; updated `TestIsAliveReportsErrorConnectingAsProbeInconclusive` to a genuinely-ambiguous message so the inconclusive path keeps real coverage
- [x] `go build ./...`
- [x] Manually reproduced against a live snapshot of real post-reboot session state (dev daemon, `AO_DATA_DIR` copy): before this fix, reconcile logged the reboot error as inconclusive and never touched the session again; after it, the same sessions are classified `ErrRuntimeUnavailable` and reconcileLive proceeds into its relaunch-in-place path

### DIFF
--- a/backend/internal/adapters/runtime/tmux/tmux.go
+++ b/backend/internal/adapters/runtime/tmux/tmux.go
@@ -530,6 +530,19 @@ func (r *Runtime) paneSessionIDs(ctx context.Context, id string) []int {
 // connection or protocol/client failure wraps ErrRuntimeProbeInconclusive so
 // no caller can treat a possibly-live session as absent. Any other non-zero
 // exit is a plain probe error, which is likewise never per-session death.
+//
+// migrationSocketAbsentOutput ("error connecting ... no such file or
+// directory") is treated as equally definitive as serverNotRunningOutput
+// here, not merely transient: this is the exact message tmux prints after a
+// real host reboot on a tmpfs-backed /tmp, where the whole per-user socket
+// directory (not just one socket file) is gone. By the time control reaches
+// here, runForSession has already resolved any private/legacy-socket
+// migration ambiguity (#3475's concern) for this specific session, so this
+// output describes the one socket that was actually, authoritatively
+// checked — it is not a raw client-side dial hiccup. Before this fix, this
+// message fell through to the transient/inconclusive branch below, which
+// left reconcileLivePass permanently unable to detect the session as dead
+// and revive it after a reboot (#4641).
 func (r *Runtime) IsAlive(ctx context.Context, handle ports.RuntimeHandle) (bool, error) {
 	id, err := handleID(handle)
 	if err != nil {
@@ -542,7 +555,7 @@ func (r *Runtime) IsAlive(ctx context.Context, handle ports.RuntimeHandle) (bool
 			if sessionMissingOutput(string(out)) {
 				return false, nil
 			}
-			if serverNotRunningOutput(string(out)) {
+			if serverNotRunningOutput(string(out)) || migrationSocketAbsentOutput(string(out)) {
 				return false, fmt.Errorf("tmux runtime: probe session %s: %w: %s",
 					id, ports.ErrRuntimeUnavailable, strings.TrimSpace(string(out)))
 			}
--- a/backend/internal/adapters/runtime/tmux/tmux_test.go
+++ b/backend/internal/adapters/runtime/tmux/tmux_test.go
@@ -1148,9 +1148,39 @@ func TestIsAliveReportsNoServerAsRuntimeUnavailable(t *testing.T) {
 	}
 }
 
-func TestIsAliveReportsErrorConnectingAsProbeInconclusive(t *testing.T) {
+// A host reboot on a tmpfs-backed /tmp wipes the entire per-user tmux socket
+// directory, not just the socket file inside it — tmux reports that as
+// "error connecting ... no such file or directory" rather than "no server
+// running". Both mean the same thing: no server exists for this user at
+// all. Misclassifying this as merely inconclusive left reconcileLivePass
+// permanently unable to detect a rebooted session as dead and revive it,
+// which is the root cause behind #4641's "sessions never come back after a
+// restart" reports (confirmed live: this is exactly the message a real
+// daemon logs after a reboot). Socket-migration ambiguity (#3475) does not
+// apply here: runForSession has already resolved which single socket is
+// authoritative for this session before IsAlive inspects the output.
+func TestIsAliveReportsMissingSocketDirectoryAsRuntimeUnavailable(t *testing.T) {
 	r, fr := newTestRuntime(0)
 	fr.outputs = [][]byte{[]byte("error connecting to /tmp/tmux-1000/default (No such file or directory)")}
+	fr.err = &exec.ExitError{}
+
+	alive, err := r.IsAlive(context.Background(), ports.RuntimeHandle{ID: "sess-1"})
+	if !errors.Is(err, ports.ErrRuntimeUnavailable) {
+		t.Fatalf("IsAlive err = %v, want ports.ErrRuntimeUnavailable", err)
+	}
+	if alive {
+		t.Fatal("alive = true, want false")
+	}
+}
+
+// A dial failure that is NOT "no such file or directory" (the socket
+// directory exists but the connection itself failed, e.g. a permission
+// problem or a race against the server's own startup) stays genuinely
+// ambiguous: it says nothing about whether a server exists, unlike the
+// missing-directory case above.
+func TestIsAliveReportsErrorConnectingAsProbeInconclusive(t *testing.T) {
+	r, fr := newTestRuntime(0)
+	fr.outputs = [][]byte{[]byte("error connecting to /tmp/tmux-1000/default (Permission denied)")}
 	fr.err = &exec.ExitError{}
 
 	alive, err := r.IsAlive(context.Background(), ports.RuntimeHandle{ID: "sess-1"})
--- a/backend/internal/service/session/service.go
+++ b/backend/internal/service/session/service.go
@@ -473,7 +473,9 @@ func (s *Service) SpawnOrchestrator(
 // It reports ok=false for any failure — an unresumable adapter, a workspace
 // restore error, or a resumed session that fails the same replacement
 // verification a fresh spawn would — so the caller can fall back to a cold
-// spawn without leaking a half-restored session.
+// spawn without leaking a half-restored session. If failure occurs after
+// RestoreWithMode has already relaunched the runtime, it retires that
+// just-resumed runtime so the fallback spawn starts from a clean slate.
 func (s *Service) resumeRetiredOrchestrator(ctx context.Context, project domain.ProjectRecord, id domain.SessionID) (domain.Session, bool) {
 	res, err := s.manager.RestoreWithMode(ctx, id)
 	if err != nil {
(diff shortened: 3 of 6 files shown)


codebanditssss (changes requested):
### Critical Risk

A successful restore can leak a second active orchestrator when post-restore validation fails. In `backend/internal/service/session/service.go:484-495`, `RestoreWithMode` has already relaunched the retired session, but failures from `toSession` or `verifyOrchestratorReplacement` return `ok=false` without retiring that resumed runtime. The caller then falls through to `Spawn`, leaving both the restored orchestrator and the cold replacement active—the exact “half-restored session” leak this helper says it prevents.

### Other Gaps

- The current tests cover restore failure before relaunch and successful restore, but not successful relaunch followed by read-back or verification failure.

### Fix

1. Retire/clean up the resumed session before returning `ok=false` from either post-restore failure path.
2. Add tests where restore succeeds but read-back or replacement verification fails, asserting only the cold replacement remains active.

### Verdict

codebanditssss (changes requested):
### critical risk

`resumeRetiredOrchestrator` treats every `RestoreWithMode` error as safe to fall back to a fresh spawn (`backend/internal/service/session/service.go:480-485`). that is not safe for the manager contract: `RestoreWithMode` can recreate the terminated session's worktree and then fail later in `relaunchSession` (agent preparation, binary validation, runtime creation, or prompt delivery). the manager deliberately leaves the terminated row and restored worktree available after those failures.

this path has already retired the old orchestrator, so a partial restore can put the canonical orchestrator branch back on the old session's worktree. the following fresh spawn uses that same canonical branch and can fail with “branch is already checked out elsewhere”, while the partial worktree, restore markers, or attachments remain. the promised cold fallback therefore does not reliably produce any coordinator.

please make the restore-failure path clean up every resource materialized by `RestoreWithMode` before falling back (a terminated row is a no-op to `RetireForReplacement`, so that call alone cannot solve it), or move the resume/fallback operation behind an atomic manager-level recovery primitive. add a test that materializes the worktree, fails during relaunch, and proves the fresh spawn succeeds with no stale worktree/runtime/attachment left behind.

### other gaps

- this pr carries the same absent-tmux-socket classification as pr #4859, but uses the existing `migrationSocketAbsentOutput` helper whose own contract says it is only for deciding whether to inspect the legacy socket and must not be treated as per-session death. rebase onto the current main line and coordinate one implementation before merging so the two prs do not leave duplicate or conflicting liveness semantics.
- the branch is based on an older main commit (`5cb411a...` while current main is `9d2a9ef...`), so the final tests need to run after that rebase.

### fix

1. clean up or atomically roll back all workspace/runtime state created by a failed `RestoreWithMode` before starting the new session.
2. add a regression test for a post-workspace restore failure followed by the cold-spawn path.
3. rebase and remove or reconcile the duplicate tmux change with pr #4859, then rerun the full checks on the rebased head.

### verdict

changes requested. the targeted tests pass on the current head, but the restore-error fallback can still strand the canonical branch and prevent recovery, so this pr is not ready to approve or merge.

codebanditssss (changes requested):
### critical risk

the reboot recovery still has a failure before resume is reached. `backend/internal/service/session/service.go:495-498` calls `RetireForReplacement`, and `backend/internal/session_manager/manager.go:1991-1995` returns any `runtime.Destroy` error. current tmux destruction treats a missing session and some server failures as idempotent, but it does not classify the tmux >=3.4 absent-socket text (`error connecting ... no such file or directory`) as success. after a host reboot, clean replacement can therefore fail at retirement and never call `RestoreWithMode`. pr #4482 contains the related idempotence work but is still open.

the restore rollback is also incomplete. `backend/internal/session_manager/manager.go:3263-3275` returns without a rollback closure if a later workspace repo restore, row activation, or attachment materialization fails after earlier worktrees were created. the cold fallback can then collide with the partially restored branch. additionally, `manager.go:3283-3293` re-upserts prior rows but does not delete newly added child rows when prior markers are non-empty.

finally, `backend/internal/service/session/service.go:570-581` logs cleanup failures after a post-restore read-back or verification error and still starts a fresh replacement. that can leave the resumed runtime/worktree active beside the new coordinator.

### other gaps

- add a regression for absent-socket tmux destruction and for partial workspace restore/attachment failure.
- the branch is eight commits behind current main and should be rebased before final validation.
- retirement captures dirty work but deletes the restore markers, so the resumed conversation may not regain captured uncommitted changes.

### fix

make retirement idempotent for the current absent-socket error, roll back every resource materialized before any restore-stage error, remove newly created worktree rows during rollback, and stop or surface the replacement failure when cleanup cannot complete. then rebase and rerun the full checks.

### verdict

changes requested. the targeted tests pass, but the reboot path and failure rollback are not yet safe to merge.

backend/internal/session_manager/manager.go
```diff
@@ -3210,39 +3243,88 @@ func (m *Manager) restoreSessionWorkspace(ctx context.Context, project domain.Pr
 			Path:          rec.Metadata.WorkspacePath,
 		})
 		if err != nil {
-			return ports.WorkspaceInfo{}, err
+			return ports.WorkspaceInfo{}, nil, err
 		}
 		if err := m.restoreAttachments(ctx, rec.ID, ws); err != nil {
-			return ports.WorkspaceInfo{}, fmt.Errorf("restore attachments: %w", err)
+			return ports.WorkspaceInfo{}, nil, fmt.Errorf("restore attachments: %w", err)
```
codebanditssss, line 3249: this error path returns a nil rollback even though `m.workspace.Restore` has already created `ws`. `restoreAttachments` can fail after materializing files or adding the exclude rule, so `RestoreWithMode` returns an error and `resumeRetiredOrchestrator` treats it as safe to cold-spawn. the recreated worktree still owns the retired branch, so the fallback can collide with it or strand it. please retain a rollback as soon as `Restore` succeeds, run it before returning attachment errors, and add coverage for both attachment failure points.

backend/internal/session_manager/manager.go
```diff
@@ -3197,7 +3221,16 @@ func (m *Manager) markSessionWorktreesActive(ctx context.Context, rows []domain.
 	return nil
 }
 
-func (m *Manager) restoreSessionWorkspace(ctx context.Context, project domain.ProjectRecord, rec domain.SessionRecord) (ports.WorkspaceInfo, error) {
+// restoreWorkspaceRollback undoes exactly what restoreSessionWorkspace
+// materialized: it force-removes any worktree(s) restoreSessionWorkspace
+// recreated on disk and restores the session_worktrees row(s) to the state
+// they were in before the call (deleting them if none existed before). It is
+// a no-op-safe best effort — failures are logged, never returned — because it
```
codebanditssss, line 3228: making rollback best-effort while discarding its errors still allows an unsafe fallback. `RestoreWithMode` returns the original relaunch error even when `ForceDestroy`, `DeleteSessionWorktrees`, or row restoration fails, and `resumeRetiredOrchestrator` then starts a new coordinator. that can leave the terminated session's branch/worktrees (or active rows) in place beside the replacement. please propagate rollback failure or otherwise block cold-spawn fallback until cleanup is confirmed, with a regression test for a cleanup failure.

codebanditssss (changes requested):
the targeted session, session-manager, tmux, and terminal suites pass, including race mode. these resume cleanup paths still need to be fixed before this can be merged:

i-trytoohard:
@Hendrixx-RE friendly nudge: changes were requested on this 5d ago and it's waiting on you. If it's dead, say the word and we close it.

Just reply here.

PR #4722, opened 2026-08-31
feat(codex): add accounts and global switching

whoisasx:
## Summary

- Stacked only on `codex/phase-1-shared-harness-readiness`.
- Adds a secure AO vault for multiple Codex accounts and makes the selected account the device-global Codex account.
- Ordinary AO Codex launches use the normal effective `$CODEX_HOME` or `~/.codex`; AO does not create a private runtime or copy native histories.
- Reconciles and safely imports an eligible file-backed device login, and supports interactive inline Add account / Sign in again flows.
- Switching atomically installs the selected credential and restarts only AO-owned running Codex controllers, preserving AO session IDs, worktrees, interface modes, and native Codex thread IDs.
- External Codex clients may report an account-changed error and require a manual restart or resume after a switch.
- Retains structured plan, capacity, activity, reset-credit, logout, and signed-out inactive account deletion flows.

## Architecture and safety

- Treats `auth.json` as opaque secret bytes: it is never parsed, logged, returned by the API, or persisted in SQLite.
- Device-global switching is enabled only when the effective Codex credential store is safely file-backed.
- Serializes account mutations and uses a durable immediate stop / credential activation / exact resume transaction with recovery.
- Keeps account actions on the primary loopback listener, enforces the configured renderer-origin boundary, and excludes them from the optional LAN listener.
- Uses `activeAccountId` as the sole authority for the active account.
- Includes one additive migration: `0123_codex_account_management.sql`.

## Explicit omissions

- No per-task account selection.
- No assisted or automatic switching.
- No AO-private Codex runtime, configuration projection, profile binding, history backfill, or native-history copying.
- No new continuation session, handoff, archive, or worktree transfer during switching.
- No account rename, reorder, or in-place replacement.
- No token/JWT parsing, credential logging, billing scraping, or raw provider-payload persistence.

## Stack

This PR preserves the Phase 1 readiness authority and does not include the discarded Phase 2–6 profile architecture. After Phase 1 merges, it can be retargeted/rebased onto `main`.

The branch history was rebuilt into five focused commits from the current Phase 1 tip after sanitation and review.


backend/internal/session_manager/codex_account_switch.go
```diff
@@ -0,0 +1,1206 @@
...
+					errs = append(errs, restartErr)
+				}
+			} else {
+				at := m.clock()
+				item.RestartState, item.RestartedAt, item.ErrorCode = "restarted", &at, ""
+			}
+			if err := m.persistCodexSwitchSession(ctx, store, switchID, *item, item.StopState, previousRestart); err != nil {
+				return errors.Join(append(errs, err)...)
+			}
+			if len(errs) > 0 {
```
nikhilachale: **[P1] Continue restarting the remaining sessions after one failure.** Returning here leaves every later session unattempted (the query orders them by session ID), so recovery can repeatedly retry the same first failure while later sessions remain pending. There is a related cross-session problem at line 863: the shared `errs` slice prevents the current restart whenever any earlier item/reviewer failed. Please use a per-item error, persist that item, continue through the full snapshot, and return the aggregate only after all sessions have been attempted. Add a regression test with at least two sessions where the first restart fails and the second succeeds.

backend/internal/session_manager/codex_account_switch.go
```diff
@@ -0,0 +1,1206 @@
...
+				}
+				if readErr != nil {
+					errs = append(errs, readErr)
+				}
+			} else if identityErr := validateCodexSwitchWorkerIdentity(rec, *item); identityErr != nil {
+				errs = append(errs, identityErr)
+			}
+			var restartErr error
+			if !adopted && len(errs) == 0 {
+				forceFresh, requireNativeHistory := codexAccountSwitchRestartPolicy(*item)
```
nikhilachale, line 871: **[P1] Make active Chat restart settlement reliable.** The switch deliberately interrupts and terminates an in-flight Chat controller, then immediately resumes it with `RequireNativeHistory=true`. `readNativeHistory` gives the provider only 45 seconds to expose a settled checkpoint. I reproduced the unchanged path with an active Chat worker: it stayed exited for about 50 seconds, entered `recovery_required`, and then succeeded immediately on manual retry. Please settle/wait at the stop boundary or add a safe automatic bounded retry, and cover an account switch during an in-flight Chat turn.

frontend/src/renderer/hooks/codex-accounts-state.ts
```diff
@@ -0,0 +1,134 @@
...
+	tone: "muted" | "warning" | "error";
+	busy: boolean;
+	canRecover: boolean;
+};
+
+export function codexSwitchDisplay(switchState: CodexAccountSwitch): CodexSwitchDisplay {
+	const failureKey = switchState.failureCode ? codexAccountReasonKey(switchState.failureCode) : null;
+	const failureKnown = failureKey !== "settings.codexAccounts.reason.unknown";
+	const phase = switchState.phase;
+	const terminal = phase === "completed" || phase === "failed" || phase === "rollback_required" || phase === "recovery_required";
```
nikhilachale: **[P2] Do not treat `recovery_required` as unfenced/idle UI state.** Marking it terminal makes `busy=false`, and `CodexAccountsSection` uses `busy` to re-enable switch, login, logout, reset, and delete controls even though the backend deliberately retains the global account-switch fence. Those operations can only fail while recovery is outstanding. Please separate progress/spinner state from mutation-fence state and keep conflicting mutations disabled until the switch completes or fails.

frontend/src/renderer/components/SessionView.tsx
```diff
@@ -1479,6 +1491,20 @@ export function SessionView({ sessionId }: SessionViewProps) {
...
+			{codexAccountSwitchBlocksSession ? (
+				<div
+					aria-live="assertive"
+					className="absolute inset-0 z-50 grid place-items-center bg-background/80 p-6 backdrop-blur-sm"
+					data-testid="codex-account-switch-blocker"
+					role="status"
+				>
+					<div className="flex max-w-sm flex-col items-center gap-3 rounded-xl border border-border bg-card px-6 py-5 text-center shadow-lg">
+						{codexAccountSwitchPresentation?.busy ? <LoaderCircle className="size-5 animate-spin text-passive" aria-label={t(codexAccountSwitchPresentation.key)} /> : null}
+						<p className="text-sm font-medium">{t("settings.codexAccounts.switchingSessions")}</p>
```
nikhilachale: **[P2] Provide recovery from the session blocker.** `recovery_required` still satisfies `codexAccountSwitchBlocksSession`, but this overlay keeps saying the account is switching and offers no retry or Settings link. The affected worker is therefore hidden behind a blocker, and the user must discover the recovery button elsewhere. Render the recovery state explicitly and provide the retry action (or direct navigation to the Codex account recovery control) here.

nikhilachale (changes requested):
Requesting changes. The account-switch restart path still has two functional correctness problems: a failure can prevent later sessions from being restarted, and an active Chat worker can hit the native-history settlement timeout and require manual recovery. The recovery UI also exposes conflicting mutations and blocks the affected session without an action. Full Go tests and frontend typecheck pass, but the current tests do not cover these failure scenarios.

frontend/src/renderer/hooks/codex-accounts-state.ts
whoisasx: Fixed in c9dad719a. Switch presentation now separates spinner activity from mutation fencing. Recovery/rollback states stop the spinner but keep Add, Switch, login, logout, reset, and delete mutations disabled until the durable switch completes or fails; regression coverage verifies the recovery fence.

backend/internal/session_manager/codex_account_switch.go
whoisasx: Fixed in c9dad719a. An in-flight Chat restart that returns the typed unsettled-history error now receives one bounded automatic retry using the same native thread ID and reserved controller generation. Added coverage for interrupting and switching an active Chat session.

frontend/src/renderer/components/SessionView.tsx
whoisasx: Fixed in c9dad719a. The session blocker now renders the recovery state explicitly and exposes Retry failed sessions directly, wired to the existing exact switch-recovery operation with pending and error feedback. Added SessionView interaction coverage.

backend/internal/session_manager/codex_account_switch.go
whoisasx: Fixed in c9dad719a. Restart processing now uses per-session worker/reviewer errors, persists each outcome, continues through the complete durable snapshot, and aggregates errors only after every session is attempted. Added a regression where the first TUI restart fails and the second succeeds.

nikhilachale (approved):

whoisasx:
Following the RCA in [issue #4908](https://github.com/Untrivial-ai/agent-orchestrator/issues/4908), there is one rare startup/recovery edge case that this PR appears to have missed.

### Missed sequence

1. The daemon starts and `WarmCodexAccounts()` begins the first device-global Codex account bootstrap.
2. That first attempt fails transiently—for example, resolving/launching `codex app-server`, its `initialize` handshake, `account/read`, the 10-second timeout, catalog cleanup, or the account-state store fails.
3. `codexAccountManager.bootstrap()` stores the error inside `bootstrapOnce.Do(...)` and closes `bootstrapDone`.
4. The underlying condition recovers while the same daemon remains alive.
5. The user retries ordinary Codex task creation.
6. `WaitCodexAccountBootstrap()` returns the cached failure forever; `sync.Once` prevents a second bootstrap attempt. Every Codex controller admission then returns `503 CODEX_ACCOUNT_MANAGEMENT_UNAVAILABLE` until the daemon is restarted.

This is why it is easy to miss in Electron development: the happy path normally succeeds on the first attempt using a working Codex binary and the established device account in `~/.codex`. Dev AO state is isolated under `~/.ao/dev`, but the device-global Codex home is still the normal Codex home. Frequent daemon restarts can also clear the poisoned in-memory `sync.Once` state before it is noticed.

The missing regression test is specifically **fail once, recover, and retry without restarting**. A fake account factory should fail its first `Open`/`Read`, succeed on the second, and verify that a later launch can complete. With the current implementation, the second factory call never happens.

This is not caused by the account-switch endpoint URI. Ordinary task creation reaches the shared controller-admission gate before launch; that gate waits for account bootstrap independently of `POST /api/v1/agents/codex/account-switches`.

One additional diagnostic problem: the underlying bootstrap error is replaced with the generic “Codex account setup did not complete” response and is not logged at the bootstrap boundary. Therefore, the reporter's exact initial transient trigger cannot be recovered from the screenshot/503 alone.

Suggested direction: replace the terminal `sync.Once` failure with a serialized/single-flight bootstrap state that permits a bounded retry after retryable failures, while preserving fail-closed behavior for permanent credential-safety failures and logging a sanitized classified cause.

Analyzed against `feef57bcba40d2e485dec53e9ea138db7277be22`.

PR #4723, opened 2026-08-31
feat: add cross-platform Neovim editor handoff

### LINKED ISSUE #4607: Add Neovim to the supported "Open in Editor" targets
## What

Add Neovim as a supported target in the "Open in Editor" handoff feature (`frontend/src/main/editor-handoff.ts`, `frontend/src/shared/editor-handoff.ts`).

## Current state

`EDITOR_IDS` in `frontend/src/shared/editor-handoff.ts` lists 20 supported editors (Cursor, VS Code, VS Code Insiders, VSCodium, Windsurf, Zed, Trae, Kiro, Positron, Sublime Text, and the JetBrains family), each mapped in `frontend/src/main/editor-handoff.ts` to a detection command and macOS app name. Neovim isn't in either list, so it never appears in the picker even when installed.

## Ask

- Add `"neovim"` to `EDITOR_IDS`.
- Add a matching entry in `frontend/src/main/editor-handoff.ts` (name `"Neovim"`, detection command `nvim`).
- Neovim has no GUI app bundle on macOS, so it likely needs to launch inside a terminal (similar to how the existing terminal open-target is handled) rather than a `macApps` entry — worth deciding during implementation whether it opens via a terminal wrapper or is skipped on macOS unless a terminal-launch path exists.

## Why

Neovim is a common editor for engineers working in this kind of terminal-first tool; right now there's no way to jump from an AO session's worktree straight into it the way you can with VS Code or the JetBrains IDEs.

Pulkit7070:
## What & why

Add Neovim as an installed "Open in Editor" target and launch it inside a native terminal on macOS, Windows, and Linux. This gives terminal-first users the same one-click workspace handoff already available for GUI editors.

Closes #4607.

## Changes

- Detect nvim on PATH and expose Neovim only when both the editor and a usable terminal launch path are available.
- Launch Neovim through macOS Terminal, Windows Command Prompt, or one of six common Linux terminal emulators.
- Preserve existing GUI editor and terminal fallback behavior while making path handling platform-correct in simulated OS tests.
- Quote executable paths safely and preserve literal Windows workspace paths by opening `.` from the workspace process directory.
- Audit the editor list; Vim, Emacs, and Helix remain intentionally outside this issue.

## How to test

1. Run: cd frontend && npx vitest run src/main/editor-handoff.test.ts
2. Run: cd frontend && npm run typecheck
3. Install Neovim, start AO, open a session, and choose Open → Neovim.
4. Confirm macOS opens Terminal with nvim at the session workspace.
5. Confirm the cross-platform desktop artifact workflow builds darwin-arm64, darwin-x64, win32-x64, and linux-x64.

## Screenshots / demo

No layout changes. Neovim appears in the existing Open menu when detected.

## Risks & notes

- Linux support depends on one of: x-terminal-emulator, gnome-terminal, konsole, xfce4-terminal, kitty, or alacritty.
- Windows uses cmd.exe with AutoRun disabled and delayed expansion off; the terminal remains open after Neovim exits.
- No storage, daemon API, migration, or network behavior changes.


Vaibhaav-Tiwari (changes requested):
Neovim won’t open on Windows because the command quotes are escaped incorrectly.

  macOS and Linux look good. The code is otherwise clean, and the focused tests pass.

also resolve the merge conflicts

frontend/src/main/editor-handoff.ts
```diff
@@ -133,13 +139,68 @@ function resolveOnPath(
...
+				const script = `tell application "Terminal"\nactivate\ndo script ${appleScriptString(commandLine)}\nend tell`;
+				return ["-e", script];
+			},
+		};
+	}
+	if (deps.platform === "win32") {
+		return {
+			command: deps.env.ComSpec || deps.env.COMSPEC || "cmd.exe",
+			argsForWorkspace: (workspacePath) => {
+				const commandLine = [editorCommand, workspacePath].map(windowsCommandArg).join(" ");
```
Rishet11: Non-blocking edge case: `cmd.exe` expands `%NAME%` sequences even inside double quotes, so a literal workspace segment such as `%TEMP%` reaches Neovim as a different path. AO-generated worktree names make this unlikely, so I would not hold the PR for it. A future hardening change could avoid parsing the path through a command string or otherwise preserve literal percent signs.

Rishet11 (commented):
One non-blocking Windows path edge case noted inline.

frontend/src/main/editor-handoff.ts
Pulkit7070: Addressed in 5e3a3e89a: Windows now launches Neovim against `.` while `launch()` sets the real workspace as `cwd`, so `%NAME%` sequences never pass through `cmd.exe` expansion. The regression test uses a `%TEMP%` workspace segment.

Pulkit7070:
Addressed the requested Windows fixes and merged the latest `main` into this branch. The blocking `cmd.exe` quoting issue is fixed, and commit 5e3a3e89a also preserves literal `%NAME%` workspace segments by opening `.` from the workspace `cwd`. All focused and complete local frontend validations pass; remote checks are running. @Vaibhaav-Tiwari, please re-review when ready.

Pulkit7070:
@Vaibhaav-Tiwari — all threads addressed and the branch is updated, please re-review.
- Windows `cmd` quoting (blocking): fixed in `8fd96f6f` via `windowsVerbatimArguments` for direct `cmd.exe` launches; covered by the prequoted-string cases in `launch-command.test.ts`.
- `%TEMP%` expansion (Rishet11's inline): fixed in `5e3a3e89a`, Windows launches `nvim` against `.` with the real workspace as `cwd`, so `%NAME%` segments never pass through `cmd.exe` expansion; covered by the `%TEMP%` regression test.
- Merged latest `origin/main` (clean, no conflicts) and re-ran the focused suites: 30/30 green (`editor-handoff` + `launch-command`).

Rishet11 (approved):
Reviewed current head 1355cd462 after the main merge. No blocking findings. The Windows handoff preserves the prequoted cmd.exe command string and opens . from the workspace cwd, so literal percent sequences in workspace paths are not expanded. The macOS and Linux launch arguments also preserve path boundaries. Focused tests passed, frontend typecheck passed, and all GitHub checks are green.

Rishet11 (commented):
**Verdict: Approved.**

I found no blocking issues in head `1355cd462`. The Neovim target is gated on the executable and a supported terminal, the macOS, Linux, and Windows launch paths preserve workspace argument boundaries, and the resolved-command refactor preserves existing GUI editor and terminal behavior. The GitHub check runs on this head are green.

Vaibhaav-Tiwari (approved):
lgtm

Image

PR #4727, opened 2026-08-31
feat(runtime): implement native Linux PTY host support

### LINKED ISSUE #4728: [Feature]: Porting the already existing PTY terminal implementation to linux
### Problem

The terminal implementation using tmux is clunky at best

### Proposed direction

Port the terminal implementation to PTY for linux just like for mac

### Alternatives considered

_No response_

### Before submitting

Hendrixx-RE:
Fixes #4728

## Summary

Implements native Linux PTY host support, mirroring the macOS (`darwin`) implementation.

### Changes
1. **Linux PTY Spawner (`spawn_linux.go`)**: Launches `ao pty-host <sessionID> <cwd> <argv...>` detached (`Setsid: true`), streaming stdout for `READY:<pid> <port>` and writing stderr to a temporary log file for startup diagnostics.
2. **Linux ConPTY Conn (`host_conpty_linux.go`)**: Implements `newConPTY` via `github.com/creack/pty` (`pty.StartWithSize`), providing full `ptyConn` interface methods (`Read`, `Write`, `Resize`, `Close`, `Done`, `PID`, `ExitCode`) with process-group teardown via `syscall.Kill(-pgid, SIGTERM)` followed by graceful fallback to `SIGKILL`.
3. **Build Tags (`spawn_host.go`, `host_conpty_other.go`, `spawn_other.go`)**: Widened `spawn_host.go` to include `linux`, and narrowed fallback stubs in `host_conpty_other.go` and `spawn_other.go` to exclude `linux`.
4. **Hybrid Runtime Extraction (`hybrid.go`, `darwin.go`, `linux.go`, `runtimeselect.go`)**: Extracted OS-agnostic hybrid runtime routing between direct native PTY host (`ptyhost-v1:`) and legacy tmux fallback into `hybridRuntime`. Updated `darwin.go` and created `linux.go` to use `hybridRuntime`, wired into `runtimeselect.go`'s default case.
5. **Tests**: Added `host_conpty_linux_test.go` and `runtimeselect/linux_test.go`, verifying end-to-end PTY spawning, sizing, process-group cleanup, exit code handling, and fallback behavior.

## Verification
- Unit & integration tests in `conpty` and `runtimeselect` pass: `go test -v ./internal/adapters/runtime/conpty/... ./internal/adapters/runtime/runtimeselect/...`
- Race detector passed: `go test -race ./internal/adapters/runtime/conpty/... ./internal/adapters/runtime/runtimeselect/...`
- Full clean backend test suite passed: `go test ./...`
- Cross-platform compilation verified for `linux`, `darwin`, and `windows`.


codebanditssss (changes requested):
### Critical Risk

Linux PTY teardown can report success while background descendants remain alive. In `backend/internal/adapters/runtime/conpty/host_conpty_linux.go:84-100`, `doneC` tracks only the immediate PTY leader, so if that process exits after launching a background job, `Close` skips signaling entirely. While the leader is alive, `Kill(-pgid, ...)` reaches only that process group—not other job-control groups in the same PTY session. The host can therefore exit and `Runtime.Destroy` can return success while a dev server or agent still owns ports or the worktree. The new test keeps the leader alive with `wait` and its child in the same process group, so it does not cover either failure mode.

### Other Gaps

- CI has not executed: Go, gitleaks, and CLI E2E are awaiting maintainer approval.

### Fix

1. Reap every process in the Linux PTY session (SID equal to the initial child PID) on every `Close`, including when `doneC` is already closed.
2. Apply TERM → grace period → KILL to the whole session.
3. Add tests where the leader exits leaving a background child and where a descendant uses a different PGID in the same session.

### Verdict

Hendrixx-RE:
Check it out again @codebanditssss  I have made the following changes:

In host_conpty_linux.go:                                                              
      - Because creack/pty.StartWithSize sets Setsid: true, the initial process starts    
      as a session leader whose leaderPID establishes the session ID (sid).               
      - Implemented /proc inspection (linuxFindSessionProcesses) to scan active           
      processes matching the session (sid == leaderPID), the initial process group        
      (pgrp == leaderPID), or transitive descendants via PPID tree walking (excluding     
      zombies).
      - Updated Close() to always teardown all processes and process groups in the
      session regardless of whether doneC is already closed.
      - Implemented two-phase termination (SIGTERM first with grace period, followed by
      SIGKILL if any session process remains alive).
-In host_conpty_linux_test.go, added tests covering both failure modes:
      - host_conpty_linux_test.go:164-200: verifies an orphaned background task is
      reaped when the leader exits early.
      - host_conpty_linux_test.go:202-242: verifies descendants in separate process
      groups (spawned via bash set -m job control) are properly signaled and reaped.
      
I think this should cover it

codebanditssss (changes requested):
### Critical Risk

The original descendant-cleanup issue is resolved, but the replacement can kill an unrelated process after PID reuse. The detached host intentionally remains available after its PTY child/session exits; a later `Close` reuses the original child PID as `sid`. In `backend/internal/adapters/runtime/conpty/host_conpty_linux.go:197-204`, `killLinuxSession` sends signals to both `-sid` and `sid` before validating current `/proc` ownership. Once the original session is empty, Linux may recycle that PID or PGID, so destroying an old retained session can SIGTERM/SIGKILL an unrelated process or group. The subsequent scan cannot prevent this because signaling has already happened, and `linuxFindSessionProcesses` seeds the stale numeric ID.

### Other Gaps

- Tests cover live orphaned/session descendants but not delayed `Close` after the original session is fully gone and its PID identity can be reused.

### Fix

1. Capture a stable process identity at launch and validate it before signaling any PID/PGID.
2. Discover and validate current session members before sending TERM/KILL; do not unconditionally signal the stale leader PID or PGID.
3. Add a regression test for fully exited session followed by delayed close/PID-identity mismatch.

### Verdict

codebanditssss (changes requested):
### Critical Risk

The unconditional stale-PID signals are removed, but identity capture still fails open. In `backend/internal/adapters/runtime/conpty/host_conpty_linux.go:56-63`, any failure reading the launched leader's `/proc/<pid>/stat` silently leaves `leaderStartTime == 0`. Lines 190-207 then treat zero as “accept any identity,” so a later `Close` can signal processes/groups selected only by a recycled numeric SID. This is particularly reachable when a short-lived leader exits before its start time is captured, and it reopens the unrelated-process termination risk.

### Other Gaps

- The new tests cover a nonzero start-time mismatch and delayed close after exit, but not identity-capture failure/the zero-value path.

### Fix

1. Require a nonzero stable identity at launch, or make teardown fail closed and send no signals when identity is unavailable.
2. Add a regression test proving `leaderStartTime == 0` cannot signal any PID or process group.

### Verdict

codebanditssss (changes requested):
### Critical Risk

The zero-identity path is now fixed, but the TERM→KILL escalation reintroduces a PID-reuse race. In `backend/internal/adapters/runtime/conpty/host_conpty_linux.go:105-129`, `Close` snapshots `{pid,startTime}` once, sends TERM, then waits and classifies survivors using bare numeric `pidAlive(pid)`. If a session process exits during the 250 ms grace period and its PID is recycled while another process keeps the wait active, the unrelated replacement is added to `aliveProcs` and receives SIGKILL from the stale snapshot. There is also a smaller scan-to-first-signal TOCTOU window.

### Other Gaps

- A one-time snapshot misses descendants or new process groups created during the TERM grace period → teardown can still report success while late children survive.
- Existing tests do not exercise identity change between TERM and KILL.

### Fix

1. Re-read and validate each process's current `startTime` immediately before every signal.
2. Wait on `{pid,startTime}`, not numeric PID alone, and rediscover validated session members before SIGKILL escalation.
3. Add a regression test for PID identity changing during the grace period.

### Verdict

codebanditssss (changes requested):
### Critical Risk

The PID-identity protections are now correct, but rediscovery is skipped when every originally snapshotted process exits after TERM. In `backend/internal/adapters/runtime/conpty/host_conpty_linux.go:106-117`, the second `linuxFindSessionProcesses` call only runs when `waitForProcIdentitiesExit(procs, ...)` returns false. If an original process handles TERM by launching a late same-session child and then exits, the wait returns true, escalation is skipped, and that child survives while `Close` reports success.

### Other Gaps

- The new grace-period test exercises identity helpers with a preconstructed stale PID; it does not drive `Close` through a late-child transition.
- There is no final rescan after SIGKILL → a process created during escalation can also escape teardown.

### Fix

1. Rediscover validated session members unconditionally after the TERM grace phase, regardless of whether original identities remain.
2. Use a bounded rescan/escalation loop or final rescan after SIGKILL until the validated session is empty.
3. Add a `Close` regression where a TERM handler spawns a same-session child and exits.

### Verdict

codebanditssss (approved):
Re-reviewed at a46ce8df. The late-child teardown gap is resolved with unconditional post-TERM rediscovery, bounded identity-validated SIGKILL sweeps, and Close-level regression coverage. Earlier process-group, zero-identity, and PID-reuse protections remain intact, and all required checks pass.

PR #4624, opened 2026-08-29
feat(cloud): add Coder sandbox provider

AgentWrapper:
## Summary

- add a Coder implementation of the provider-neutral sandbox lifecycle
- create one deterministic Coder workspace per AO session and map Coder build/agent health onto AO states
- bootstrap and repair `ao-worker` through the authenticated Coder agent PTY WebSocket; use the buffered PTY backend so bootstrap output and exit status are reliable
- wire Coder into hosted configuration, provider resolution, worker-token TTLs, staging deployment, and production promotion
- inject only the selected sandbox provider secrets into ECS task definitions and preserve the staging-selected provider during promotion
- document customer access, API, connectivity, template, sudo, and harness requirements

## Verification

- `cd cloud && go vet ./...`
- `cd cloud && go test -race ./...`
- `python3 cloud/scripts/tests/test_deployment.py` (17 tests)
- `bash -n cloud/scripts/deploy-staging.sh cloud/scripts/promote-production.sh`
- live Coder CE lifecycle on AWS: create workspace -> connected agent -> passwordless-sudo worker bootstrap/upload/launch -> stop -> start -> delete (passed in 63.09s)
- staging task definition `:64` launched two Coder-configured control-plane tasks and both became ALB-healthy; another operator explicitly restored NodeOps task definition `:63` 119 seconds later, so the persistent staging/product E2E awaits a coordinated deployment window
- confirmed the `ao-integration` Coder user owns zero workspaces after testing

## Deployment and security notes

- the worker runtime now uses Debian 13 so the earlier OpenSSL and util-linux findings are fixed
- the ECR gate still rejects every new High/Critical finding; the explicit list contains current Debian Trixie no-DSA findings and is shared by staging and production promotion
- production was not modified

## Scope / follow-ups

- this slice configures one deployment-wide Coder connection and approved template; per-AO-organization Coder connections remain a follow-up
- the pilot bootstrap requires passwordless sudo; production should prefer a template-baked worker user/directories or a narrowly scoped install helper
- a full coding-agent task needs a Coder template containing the selected harness plus the corresponding user/provider credentials; the current pilot template validates workspace/terminal/worker lifecycle but does not contain those harnesses
## Bug fixes found during product testing

- Fixes #4730: cloud **New terminal** now opens a workspace shell through the control plane instead of calling the local daemon; cloud shell tabs route to `kind: workspace`, and bounded concurrent workspace shells are supported.

PR #4736, opened 2026-09-01
fix(cloud): await Coder bootstrap result

### LINKED ISSUE #4734: bug(cloud): Coder bootstrap reports success before install and worker launch complete
## Bug

Coder BootstrapWorker returns success after the archive upload acknowledgement rather than after extraction, installation, and AO worker launch succeed.

## Root cause

bootstrapThroughPTY returns after receiving __AO_UPLOAD_DONE__. It does not await the existing final __AO_BOOTSTRAP_OK__ or __AO_BOOTSTRAP_FAILED__ result. Closing at that point can also leave streamPTYOutput blocked on an unconsumed unbuffered channel.

## Required fix

- Keep the buffered PTY connection alive after upload completion.
- Await explicit bootstrap success/failure or process/transport termination.
- Propagate extraction, sudo/user creation, binary installation, and worker-launch failures to the reconciler.
- Ensure all stream goroutines terminate on success, failure, cancellation, and disconnect.
- Add tests for success, post-upload command failure, transport closure, and cancellation without goroutine leaks.

## Acceptance

BootstrapWorker returns nil only after __AO_BOOTSTRAP_OK__. A post-upload __AO_BOOTSTRAP_FAILED__ reaches reconciliation as an error. No output goroutine remains blocked when the method returns.

This is a merge blocker discovered in the independent review of #4624 and must be merged into the #4624 branch before that PR lands.

AgentWrapper:
## Summary

- keep the buffered Coder PTY open after `__AO_UPLOAD_DONE__` and return success only after explicit `__AO_BOOTSTRAP_OK__`
- propagate final `__AO_BOOTSTRAP_FAILED__` and PTY process/transport termination as bootstrap errors
- cancel, close, and join the PTY output producer on success, failure, cancellation, and disconnect so it cannot block after return
- preserve the existing framed archive upload and credential-handling behavior
- extend the credential-gated live lifecycle to prove a deterministic post-upload install failure is surfaced before the valid bootstrap path

Closes #4734.

This is a merge blocker for #4624 and is intentionally stacked on `ao/agent-orchestrator-211/coder-sandbox-provider`.

## Verification

- `cd cloud && go test ./internal/sandbox/coder -run 'TestBootstrapWorkerStreamsArchiveWithoutSecretsInURL|TestBootstrapWorkerPropagatesPostUploadFailure|TestBootstrapThroughPTYRejectsPrematureClose|TestBootstrapThroughPTYCancellationStopsOutput|TestStreamPTYOutputCancellationDoesNotBlock' -count=20`
- `cd cloud && go test -race ./internal/sandbox/coder`
- `cd cloud && go test ./...`
- `cd cloud && go test -race ./...`
- `cd cloud && go vet ./...`
- credentialed staging Coder: `cd cloud && go test ./internal/sandbox/coder -run '^TestLiveLifecycle$' -count=1 -v` — passed in 118.46s, including expected `/proc/ao-worker` post-upload failure, valid bootstrap, stop/start, and delete
- post-test authenticated verification: disposable workspace absent from the owner's active workspace list and direct lookup returned HTTP 410
- `git diff --check`

## Scope and risk

- no persistence paths or lifecycle/reconciler semantics changed
- no shared workspace or staging ECS service changed
- no deployment performed
- credential values are omitted from this PR

### DIFF
--- a/cloud/internal/sandbox/coder/client.go
+++ b/cloud/internal/sandbox/coder/client.go
@@ -316,7 +316,7 @@ func (c *Client) BootstrapWorker(ctx context.Context, id sandbox.ID, bootstrap s
 	query.Set("height", "40")
 	query.Set("command", command)
 	// Bootstrap is a short-lived, non-interactive command. The buffered backend
-	// lets the command finish after AO disconnects once the upload is accepted.
+	// preserves the final result after the upload while AO keeps the PTY open.
 	query.Set("backend_type", "buffered")
 	ptyURL.RawQuery = query.Encode()
 	var lastErr error
@@ -352,12 +352,16 @@ func (c *Client) bootstrapThroughPTY(ctx context.Context, ptyURL *url.URL, encod
 		}
 		return fmt.Errorf("coder: open workspace PTY: %w", err)
 	}
-	defer conn.CloseNow()
-	netConn := websocket.NetConn(context.Background(), conn, websocket.MessageBinary)
-	defer netConn.Close()
+	streamCtx, stopStream := context.WithCancel(ctx)
+	netConn := websocket.NetConn(streamCtx, conn, websocket.MessageBinary)
+	output, outputDone := streamPTYOutput(streamCtx, netConn)
+	defer func() {
+		stopStream()
+		_ = conn.CloseNow()
+		<-outputDone
+	}()
 
 	encoder := json.NewEncoder(netConn)
-	output := streamPTYOutput(netConn)
 	// Coder's reconnecting PTY writes each decoded Data field to the OS PTY but
 	// does not retry a short write. Frame the archive as canonical terminal lines
 	// with sequence and length metadata. The receiver acknowledges only a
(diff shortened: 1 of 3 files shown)


AgentWrapper (commented):
Verdict: APPROVE

Blockers: none.

The Coder PTY bootstrap state machine now has the required ordering: archive frames are acknowledged, __AO_UPLOAD_DONE__ only transitions into the result-wait phase, and nil is returned only after __AO_BOOTSTRAP_OK__. __AO_BOOTSTRAP_FAILED__, EOF/process termination, transport errors, and context cancellation remain errors. A fresh reconnect UUID is used for every outer retry, so buffered output from an earlier attempt cannot satisfy a later attempt.

Cleanup is structurally guaranteed: every connected attempt cancels the stream context, closes the websocket, and waits for outputDone before returning. This covers success, final failure, cancellation, and disconnect, including a producer blocked on its unbuffered output channel.

The tests do regress the prior false-success bug. TestBootstrapWorkerPropagatesPostUploadFailure sends __AO_UPLOAD_DONE__ and only then __AO_BOOTSTRAP_FAILED__; the base implementation would return nil at the upload marker. The cancellation test also holds the PTY open after upload and would reject the old early nil return. Happy-path OK and premature-close behavior are separately covered.

Output handling is acceptable for this fix: terminal C0 controls such as CR/ESC are removed, failure text is bounded to the final 1024 bytes, and the credential-bearing archive is sent with terminal echo disabled rather than included in the URL or result output.

Follow-ups, non-blocking: explicit protocol failures still participate in the pre-existing three-attempt outer retry loop. That preserves current retry semantics and each attempt is isolated, but terminal-versus-transient classification could be refined later if repeated deterministic installs prove costly. sanitizePTYOutput could also be hardened separately to remove DEL/C1 Unicode controls; no exploitable source for those controls is introduced by this patch.

Exact checks:
- go test ./internal/sandbox/coder -run TestBootstrapWorkerStreamsArchiveWithoutSecretsInURL|TestBootstrapWorkerPropagatesPostUploadFailure|TestBootstrapThroughPTYRejectsPrematureClose|TestBootstrapThroughPTYCancellationStopsOutput|TestStreamPTYOutputCancellationDoesNotBlock -count=20 — PASS (40.351s)
- go test -race ./internal/sandbox/coder -count=1 — PASS (3.905s)
- git diff --check origin/ao/agent-orchestrator-211/coder-sandbox-provider...642f0c8e9a0110ffc9608d58d696b32873aeebb9 — PASS
- GitHub gitleaks scan — PASS

Credentialed evidence inspected, not rerun: the PR records TestLiveLifecycle passing in 118.46s, including the expected post-upload /proc/ao-worker install failure, a subsequent valid bootstrap, stop/start, delete, and authenticated cleanup verification returning HTTP 410. The live_test.go change matches that claimed sequence.

PR #4745, opened 2026-09-01
fix(cloud): coalesce pending terminal input

AgentWrapper:
## What

Coalesce consecutive, unclaimed terminal input into the pending durable worker request at the PostgreSQL queue boundary.

## Why

Cloud terminal input currently amplifies every xterm `onData` event into one `ao_worker_requests` row followed by a serial worker `ClaimTransport` and `CompleteTransport` pair. For N input events, the saturated path therefore requires N durable work items and 2N worker/control-plane HTTP operations. At 75-100 ms per worker HTTP operation, a 20-character burst has a 3-4 second serial drain-time floor, before database and PTY work.

The focused database-backed test sends 100 one-byte events before claim. Before this change that represents 100 worker requests / 200 worker HTTP operations (and exceeds the outstanding-request cap unless the worker drains concurrently); after this change it is 1 request / 2 worker HTTP operations: a 99% reduction for the backlogged burst.

## How

- Serialize terminal enqueue mutations per session and append bytes only to the newest `pending` `terminal.input` request.
- Do not add a frontend timer: the first input is still enqueued immediately, and coalescing activates only when transport has fallen behind.
- Treat resize as an ordering boundary and never mutate a `claimed` request.
- Keep idempotency-keyed input in separate requests so reconnect/deduplication semantics remain intact.
- Preserve raw byte concatenation, including control bytes and partial/complete UTF-8 sequences, and cap coalesced payloads at the existing 16 KiB worker limit.
- Refresh the 15-second request TTL when appending so new bytes never inherit a nearly expired deadline.

Trade-off: each WebSocket input still performs one small PostgreSQL transaction (an update instead of another insert while backlogged). This intentionally targets the measured serial worker HTTP amplification without changing the public protocol, worker API, provider adapter, frontend timing, or delivery fencing.

## Testing

- `cd cloud && go build ./...`
- `cd cloud && go test ./... -count=1`
- `cd cloud && go test -race ./... -count=1`
- `cd cloud && go vet ./...`
- PostgreSQL 17 integration: `AO_CLOUD_TEST_DATABASE_URL=... go test ./internal/postgres -run TestQueueTerminalInputCoalescesOnlyThePendingTail -count=1`
  - Verifies 100 one-byte inputs claim as one ordered request.
  - Verifies resize splits batches.
  - Verifies no extra worker request remains.
- Unit coverage verifies Ctrl-C, escape sequences, NUL, UTF-8, carriage return, idempotency boundaries, terminal boundaries, empty input, and the 16 KiB cap.

No staging or production deployment was performed.

## Checklist

### DIFF
--- a/cloud/internal/postgres/worker_transport_store.go
+++ b/cloud/internal/postgres/worker_transport_store.go
@@ -83,17 +83,19 @@ func createWorkerRequest(
 	if kind == "workspace.write" && effectiveMode(mode, modeCap) == "read-only" {
 		return domain.WorkerRequest{}, ErrWorkspaceReadOnly
 	}
+	terminalRequest := strings.HasPrefix(kind, "terminal.")
 	var outstanding int
 	if err := tx.QueryRow(ctx,
 		`SELECT count(*) FROM ao_worker_requests
 		WHERE org_id = $1 AND session_id = $2
-		  AND status IN ('pending', 'claimed') AND expires_at > now()`,
-		orgID, sessionID,
+		  AND status IN ('pending', 'claimed')
+		  AND expires_at > CASE WHEN $3 THEN statement_timestamp() ELSE now() END`,
+		orgID, sessionID, terminalRequest,
 	).Scan(&outstanding); err != nil {
 		return domain.WorkerRequest{}, err
 	}
 	limit := maxOutstandingWorkspaceRequests
-	if strings.HasPrefix(kind, "terminal.") {
+	if terminalRequest {
 		limit = maxOutstandingWorkerRequests
 	}
 	if outstanding >= limit {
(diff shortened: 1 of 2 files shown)


AgentWrapper:
Addressed the ordering/TTL review in 67bde0371:

- Terminal request `created_at`, `updated_at`, expiry creation, expiry refresh, and expiry checks now use `statement_timestamp()` after the session lock. Non-terminal workspace request timing remains unchanged.
- Added a deterministic PostgreSQL regression where transaction A starts 1.5s before B, B serializes a resize first, and the worker must claim B before A input.
- The regression also checks that A retains at least 14s of the 15s TTL after delayed enqueue.

Re-ran Cloud build, all Go tests, race tests, vet, and the PostgreSQL 17 integration tests. No deployment performed.

i-trytoohard:
@AgentWrapper Marking as stale: no activity for 22 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4746, opened 2026-09-01
fix: avoid xterm viewport callback after StrictMode disposal

AgentWrapper:
## Summary

- eliminate a development-only xterm lifecycle pageerror during React StrictMode cleanup
- invalidate the AO terminal handle immediately, then defer physical xterm disposal one task so xterm 5's queued viewport-open callback can finish
- cover the exact disposal ordering with a deterministic regression test

## Diagnosis and scope

The live cloud terminal path is healthy: deeper inspection found 752 xterm buffer lines and a visibly rendered Claude prompt including AO_CODER_WORKER_OK. The cached dimensions exception is therefore not evidence of a blank-pane or replay failure.

The exception itself is reproducible and narrowly understood. xterm 5 schedules an uncancellable zero-delay Viewport.syncScrollArea callback from Terminal.open. React StrictMode cleans up its first development mount in the same task. Immediate disposal releases the renderer before that callback runs, so it reads RenderService.dimensions after disposal and raises a pageerror.

This PR only removes that lifecycle pageerror. It does not change cloud ticketing, replay coalescing, attachment ordering, or local/cloud mux behavior, and it should not be treated as a product-blocker fix.

## Exact verification run

- npm test -- XtermTerminal.test.tsx TerminalPane.test.tsx useTerminalSession.test.tsx --run (167 passed)
- npm run frontend:typecheck (passed)
- lifecycle regression failed before the production change: the fake xterm open timer observed prior disposal; after the change, queued open work runs first and disposal still occurs exactly once
- PR CI passed: frontend test, renderer-smoke, and gitleaks

## Native visual-verification follow-up

A second real Electron instance was intentionally not launched because agent-orchestrator-211 owns the shared AO dev Chromium profile and ports 5173/3002 for active user hands-on testing. This PR did not stop or modify that instance, its preview, its ports, or its profile, and did not launch against the production/shared AO_DATA_DIR.

Follow-up requested: after agent-orchestrator-211's user test is explicitly finished and the shared dev resources are released, rerun this branch in the real Electron app and confirm terminal mount/navigation produces no xterm dimensions pageerror.

No staging or production deployment was performed.

### DIFF
--- a/frontend/src/renderer/components/XtermTerminal.test.tsx
+++ b/frontend/src/renderer/components/XtermTerminal.test.tsx
@@ -313,6 +313,24 @@ describe("XtermTerminal", () => {
 		}
 	});
 
+	it("settles a pending write when terminal cleanup wins the race", () => {
+		let terminal: AttachableTerminal | undefined;
+		const view = render(
+			<XtermTerminal
+				theme="dark"
+				onReady={(ready) => {
+					terminal = ready;
+				}}
+			/>,
+		);
+		const done = vi.fn();
+
+		view.unmount();
+		terminal!.write(new Uint8Array([0x61]), done);
+
+		expect(done).toHaveBeenCalledOnce();
+	});
+
 	it("preserves the agent TUI palette without contrast remapping", () => {
 		render(<XtermTerminal theme="dark" />);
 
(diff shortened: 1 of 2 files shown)


codebanditssss:
Thanks for the contribution. This exact StrictMode/xterm lifecycle fix is already present on current  following #5105, which defers terminal disposal until the queued viewport initialization work drains and includes the regression coverage. This PR is therefore no longer necessary.

PR #4748, opened 2026-09-01
fix(lifecycle): preserve Chat and TUI status across desktop restarts

### LINKED ISSUE #4458: bug(update): tmux namespace changes falsely exit sessions and spawn duplicate controllers

illegalcall:
Fixes #4458

## Outcome

This replaces the original narrow tmux compatibility patch with a complete restart-continuity fix for both Chat and TUI sessions.

Existing sessions recover automatically when the desktop app starts. There is no import command, migration script, or user action required.

## Why this regressed

The visible failure came from several changes interacting:

- #4330 exposed the daemon API before startup reconciliation finished, so the UI could render stale durable `Exited` facts.
- #3711 treated every Chat stream closure—including a planned desktop shutdown—as terminal and persisted `Exited`.
- #4393 moved TUI sessions to a deterministic private tmux socket; #4465 later removed that runtime without retaining recovery for the private-socket cohort.
- #4071 limited startup recovery to four workers, explaining why larger boards appeared to repair themselves in waves or after the later reaper pass.

tmux was therefore one cause of legacy TUI recovery failures, but it was not the common cause of the false `Exited` state seen by both Chat and TUI.

### Startup visibility

- Gate session/state API routes until startup reconciliation succeeds, so the renderer never observes a partially recovered board.
- Keep startup unready when ownership is ambiguous or a probe is inconclusive; do not mutate the board or launch a replacement.

### Chat continuity

- Treat desktop shutdown as a planned Chat detach rather than conversation termination.
- Reattach Chat sessions to their persistent host and exact provider conversation.
- Recover complete Codex history through paginated `thread/turns/list`, with compatibility fallback for older `thread/read` servers.

### TUI continuity

- Recover modern PTY protocol-v3 and shipped protocol-v2 hosts using exact runtime/process-incarnation evidence.
- Preserve inconclusive PID semantics on Windows, tolerate Linux's exact post-update ` (deleted)` executable suffix, and give cold Windows legacy identity proof a dedicated bounded deadline.
- Discover historical private, named, and default tmux namespaces and canonicalize a proven owner automatically.
- Persist canonical tmux routing with the server PID, native session ID, pane ID, owner session, and launch ID.
- Fence tmux reads, writes, attach, destroy, and recovery to the proven object; fully canonical handles never fall back to broad name rediscovery.

## Safety properties

- Failed or unknown probes are never treated as proof that a session died.
- Ambiguous ownership leaves durable state and live runtimes untouched.
- Existing identities are upgraded in place only after ownership is proven.
- Same-name runtime replacements are rejected rather than adopted or mutated.
- Recovery is automatic; historical sessions are not copied, imported, or recreated.

## Packaged desktop E2E

The real packaged Electron app exercises six simultaneous cohorts:

1. Chat
2. native PTY protocol-v3 TUI
3. shipped PTY protocol-v2 TUI
4. historical private-socket tmux TUI
5. named-socket tmux TUI
6. default-socket tmux TUI

Across five app launches it verifies:

- first and second cold restarts
- immediate daemon-owner handoff
- startup API readiness and zero visible `Exited` state
- exact provider/runtime identity preservation
- real terminal input/output through every recovered PTY and tmux route
- survival beyond the reaper interval
- fail-closed behavior under deliberately ambiguous ownership

Final exact-HEAD packaged result: `1 passed (1.5m)` on `9e0f32e45`.

## Verification

- `go test ./...` and pinned golangci-lint v2.12.2 with an isolated cache — 0 issues
- focused adapter, lifecycle, reaper, session-manager, and storage tests
- race tests for Chat, PTY, runtime selection, and tmux recovery
- Linux and Windows ConPTY cross-builds; Windows vet
- Linux PTY arm64/amd64 race stress for the CI fixtures
- packaged desktop restart-continuity Playwright E2E


illegalcall:
## Packaged desktop E2E video evidence (`9e0f32e45`)

Fresh isolated run: **`1 passed (1.6m)`**. Both clips are renderer-only captures from the packaged Electron app; they do not capture the desktop or real AO data.

### 1. Cold restart continuity — Chat + every TUI generation

https://github.com/user-attachments/assets/9731dda5-9b36-42a3-ba82-d3b6d0a2bf0a

This starts at a real cold launch, passes through the startup recovery gate, and shows all six surviving cohorts: Chat, native PTY v3, PTY v2, historical-private tmux, named tmux, and default tmux. Their recovered statuses never render as `Exited`, and the clip then shows real terminal round trips against the preserved workloads.

The E2E's continuous DOM audit saw **zero visible `Exited` mutations**; its concurrent API observer also saw no false terminal state through startup and a periodic reaper interval. Runtime/session identities were asserted unchanged.

_The “No orchestrator is running for this project” banner refers only to the optional project-level orchestrator; it is unrelated to these six agent sessions._

### 2. Ambiguous legacy ownership fails closed

https://github.com/user-attachments/assets/09567799-6678-42d9-821b-c2f4b58dd4c3

This deliberately creates a second exact legacy owner. AO remains behind the recovery cover, reports `startup_recovery_failed`, serves sessions as `503`, exposes no stale Kanban cards, and leaves both runtimes plus the durable row untouched.

Tested from exact PR head: `9e0f32e45a7998d6a3765adbefedc04f4b3311b2`.

illegalcall:
## Exact-base vs exact-head desktop restart A/B

This is the differential visual proof that the earlier after-only recording was missing.

https://github.com/user-attachments/assets/5fba0da3-06a2-4a39-9177-112428aa14ea

### Exact problem being fixed

Restarting the AO desktop app does **not** stop the underlying Chat/PTY/tmux workloads. The broken behavior is that the base build publishes its listener and lets the board read durable `Exited` facts **before startup recovery has reconciled those facts with the still-live workloads**. That creates a false `Exited` frame; some sessions repair later, while historical runtime cohorts can remain wrong.

The PR changes startup into a barrier: state-bearing API/UI reads stay gated while Chat and TUI ownership is recovered, live activity is reconciled with exact workload identity, and only then does the first usable board render. It also prevents a planned Chat-controller detach from publishing a false terminal activity signal, fixing one source of stale Chat state as well as the startup exposure.

### Controlled, identical scenario

Both packaged runs use a fresh isolated AO data directory and the same six real live cohorts:

- Chat
- native PTY
- released protocol-v2 PTY
- historical private-socket tmux
- named-socket tmux
- default-socket tmux

The recording first shows the healthy board. The harness then creates the affected precondition deterministically by writing stale `Exited` activity through AO's production activity endpoint while leaving every workload alive, quits **only** the desktop app, and relaunches it. This is intentionally controlled: it tests whether stale durable state can leak to users before recovery, independent of timing luck.

### Observed result

| Revision | Startup/API observation | First stable board | Workload proof |
|---|---|---|---|
| **Before — exact PR base `f16f31c3759e36b0c4ce82be14d59587d03bed25`** | `0` recovery gates; HTTP `200` exposed all six sessions as `Exited` | Five TUI cards visibly remained `Exited`; Chat had repaired asynchronously by the held board frame | Chat controller plus both PTY hosts and all three tmux workloads were still alive |
| **After — exact PR head `9e0f32e45a7998d6a3765adbefedc04f4b3311b2`** | `58` sampled `startup_recovery_in_progress` gates, then `0` invalid `200` responses | `0` `Exited` cards and `0` DOM `Exited` sightings | All six identities survived; five terminal input/output round trips passed; state stayed correct through a reaper interval |

So the before recording proves the actual contradiction: AO says `Exited` while the exact workloads are alive. The after recording proves the PR does not merely repaint the cards later—it prevents the false state from being observable at either the API or UI boundary.

Both recordings capture only AO's renderer and synthetic fixture names. No normal AO data or other desktop content was recorded.

PR #4752, opened 2026-09-01
fix(cli): make screenshot JSON honor paths

### LINKED ISSUE #4707: bug(cli): browser screenshot ignores path with --json and floods output with base64
## Summary

`ao browser screenshot <path> --json` accepts the positional path but silently ignores it. It writes no PNG and prints the entire image as base64 JSON, which can flood and truncate agent tool output.

**Reported by:** external AO user; report relayed by the project operator (public handle/source URL not supplied)  
**Observed:** 2026-08-31  
**Analyzed against:** `main` at `ab17197b0d1bdc8d4d9849ef46675e067fc77119`  
**Confidence:** High; current CLI implementation has this exact branch.

## Reproduction

```sh
ao browser screenshot /tmp/ao-browser-test.png --json
test -f /tmp/ao-browser-test.png
```

Observed:

- command returns a very large `result.data` base64 field;
- the requested file does not exist.

The user report measured roughly 46,000 tokens before truncation.

## Root cause and contract conflict

In `backend/internal/cli/browser.go`, screenshot handling calls the browser action first and then immediately serializes the full response when `jsonOutput` is true. Positional-path handling exists only in the non-JSON branch.

The bundled browser guide currently documents base64 for `--json`, but the Cobra contract says `screenshot [path]` and describes the command as capturing to a PNG file. Accepting a path and silently discarding it is unsafe and surprising even if the behavior was documented elsewhere.

## Acceptance criteria

- When a path is supplied, write the PNG in both text and JSON modes.
- JSON output for a file write returns compact metadata: resolved path, byte size, width, and height.
- Inline image data requires an explicit opt-in such as `--base64`; it is never the accidental result of combining a path with `--json`.
- Preserve overwrite protections and path-validation behavior.
- Add CLI tests for absolute/relative paths, `--json`, file bytes, compact output, overwrite refusal, and explicit inline mode.
- Update the bundled browser guide and help text so they agree.

aniketshukla1:
## What

- write screenshots to the requested or generated path in both text and JSON modes
- return compact `path`, `size`, `width`, and `height` metadata after JSON file writes
- require explicit `--base64 --json` for inline image data and reject ambiguous path/flag combinations before contacting the daemon
- document both screenshot modes and add coverage for absolute and relative paths, file bytes, compact JSON, overwrite refusal, validation, and help text

## Why

`ao browser screenshot <path> --json` accepted the path but skipped file creation and emitted the full base64 image response. This could flood and truncate agent output while silently ignoring the requested destination.

Fixes #4707.

## Testing

- `go test ./internal/cli`
- `go test ./internal/skillassets`
- `go vet ./internal/cli`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs ./internal/cli/...` (`0 issues`)

A repository-wide `npm run lint` attempt reached and passed the modified CLI and skill-asset packages, but its test stage was blocked locally by an unrelated missing-tmux prerequisite in `internal/session_manager`; one unrelated Pi timing test also failed under the parallel run and passed immediately when rerun in isolation.

### DIFF
--- a/backend/internal/cli/browser.go
+++ b/backend/internal/cli/browser.go
@@ -36,6 +36,13 @@ type browserCommandResponseDTO struct {
 	Result    map[string]any `json:"result"`
 }
 
+type browserScreenshotFileResult struct {
+	Path   string `json:"path"`
+	Size   int64  `json:"size"`
+	Width  int    `json:"width"`
+	Height int    `json:"height"`
+}
+
 const browserCapabilityHeader = "X-AO-Browser-Capability"
 const maxBrowserWaitMillis = 55_000
 const (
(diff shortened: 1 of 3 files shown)


Vaibhaav-Tiwari:
reviewing

Vaibhaav-Tiwari (approved):
lgtm

PR #4761, opened 2026-09-01
feat: update copy now AO Mobile is live on both stores

Prasad-D-Ware:
AO Mobile is live on the App Store and Google Play, so every surface that
described a beta needed to change. Most of this is subtraction: the pre-launch
copy existed to explain multi-step workarounds that no longer exist.

**34 files, +674 / −1133.**

## Landing

The download page had two install funnels to explain — TestFlight (install
Apple's app, accept an invite) and an Android closed test (join a Google Group,
opt in, then install). Both are now one tap on a public listing, so the
walkthroughs are deleted rather than reworded, along with their phone-specific
sheets: a sheet whose job was to say "tap this other button" is pure friction
once the button goes straight to the store.

One shared `StoreQRDialog` replaces all five components. On a phone the badge
links directly to the listing; on a desktop it opens a QR, because the app
cannot install on the machine the visitor is reading this on. That handoff is
the only part of the old flow worth keeping.

The badges are Apple's and Google's own artwork, which their identity
guidelines require — the hand-drawn recreation this started with violated both
while also losing the recognition that makes a badge worth using. Their wording
differs ("Download on the" vs "Get it on") because each guideline fixes its
own; that mismatch is deliberate, not an oversight.

## Desktop — Connect Mobile

The iOS side still described a TestFlight beta. Pointing it at the App Store
made it identical to the Android side, which let the two branches collapse into
one and set off a chain of simplifications:

- **The platform dropdown is gone.** It never reached the pairing QR — the
  payload is built with `platform: ""`, and that field carries the desktop's OS,
  not the phone's — so all it selected was which of two store links to show.
  Step one now names both, each with its own hover QR.
- **The LAN step is gone.** `qrIsReady` withholds the code until the tunnel
  resolves, precisely so a scannable code always carries a tunnel endpoint, and
  the phone races every advertised endpoint. "Put your phone on the same Wi-Fi"
  contradicted a guarantee the panel already enforces, and the case where it is
  true (no cloudflared) has its own notice. Tailscale keeps its step — there,
  installing it really is a prerequisite.
- **`mobile.description` loses "over your LAN"** for the same reason, with no
  replacement qualifier: "from anywhere" would be wrong whenever the connector
  is missing.

All eight locales updated, mirroring each one's existing Google Play phrasing.
Store names stay untranslated as brand marks; the storefront name survives in
the aria labels for screen readers.

## Mobile app

The push-registration failure named an internal cause the person holding the
phone can do nothing with ("this build has no push entitlement") and sent them
to TestFlight. It now names the fix. `openStore`'s `itms-beta://` fallback
points at the real listing.

## Needs a decision before merge

**The connection-method picker (LAN/Tailscale) is commented out, not deleted**,
in three linked places so it restores as one unit. This also takes the
**secure-pairing toggle** off-screen, which iPhone-over-Tailscale requires —
iOS refuses cleartext to `100.x` addresses, and the TLS MagicDNS host is
deliberately not in the QR. LAN and tunnel paths are unaffected. The seven
tests that reached Tailscale through that picker are skipped with a matching
note rather than rewritten to bypass the UI, so they restore alongside it.

## Verification

- `tsc --noEmit` clean
- 15 suites, 99 passed / 7 skipped; mobile lib 70 passed
- Rendered on iPhone 13 and Pixel 7 viewports: no horizontal overflow, correct
  per-device badge behaviour (own store = direct link, other store = QR)

## Noted separately

On a phone the **cookie consent banner covers the store badges** on first load
— it buries the primary CTA on the page people arrive at to download the app.
Pre-existing and out of scope here, but worth its own fix.


ronishrohan (changes requested):
Blocking: the Connect Mobile connection-method picker is commented out in `frontend/src/renderer/components/settings/ConnectMobileContent.tsx:406-424`, while `mode` is pinned/reset to `"lan"` at lines 220-251. That makes the Tailscale flow and secure-pairing toggle unreachable in the shipped UI, and seven Tailscale/secure-pairing tests are skipped for exactly this reason (`ConnectMobileModal.test.tsx:132-144, 188-320`). Please either restore the picker and its coverage or explicitly remove/deprecate the Tailscale path in the same change with product approval. I did not approve or merge this PR.

ronishrohan (approved):
Approved per request. The Tailscale picker limitation remains documented in the prior review.

Prasad-D-Ware:
Addressed in a929dea3b. The picker is restored and all seven skipped tests are back — the suite now runs 108 passed / 0 skipped.

Rather than restoring the secure-pairing toggle alongside it, I removed it and made TLS unconditional for Tailscale, with product approval. The reasoning: it was never a real choice. iOS refuses cleartext connections to a `100.x` address, and the TLS MagicDNS host is deliberately kept out of the pairing QR, so a Tailscale pairing with the toggle off works on Android and fails on iPhone with nothing on either side to explain why. Selecting Tailscale now enables it automatically and the row reports state instead of offering a switch.

Enabling is attempted once per visit to Tailscale mode, guarded by a ref: a tailnet with no certificates rejects every attempt, and retrying on each status poll would hammer the daemon. The existing reason text already surfaces that failure, and switching modes away and back is the retry.

Test coverage for the new behaviour, in `ConnectMobileModal.test.tsx`:
- `turns secure pairing on by itself when Tailscale is selected` — asserts the POST fires and no switch is rendered
- `does not retry enabling secure pairing when it is unavailable`
- `surfaces an error when enabling secure pairing fails` — replaces the old toggle-driven version

Also rebased onto `d820a03c6`. One conflict in `ConnectMobileContent.tsx` from the tooltip refactor (#4558), resolved by dropping the `TooltipProvider` wrapper to match the new convention.

Note for CI: `ChatTimelineItems.test.tsx` and `ChatWorkspace.test.tsx` each fail one date-labelling test on this branch. Both reproduce on a clean checkout of `untrivial/main` at `d820a03c6` with none of this branch's code present, so they are pre-existing and unrelated.

Prasad-D-Ware:
Update on the connection picker, since this reverses part of what I said in my last comment.

Per product decision, the LAN/Tailscale picker is commented out again in 020397d95. The reasoning is the one the review thread was circling: a v2 pairing code carries every advertised endpoint — LAN, Tailscale and tunnel — and the phone races them (`mobilebridge.Endpoints`), so the picker never changed what a scan produces. It only ever selected which address the manual-entry rows displayed.

**The functional gap from the first review is closed, not re-introduced.** The blocking concern was that hiding the picker made secure pairing unreachable. TLS no longer keys off the picker at all — it is enabled whenever the machine has a tailnet address, independent of any UI mode. A hidden control should not take a security property with it, so:

- secure pairing is enabled automatically wherever Tailscale exists, with or without the picker
- the certificate-failure reason text is keyed the same way, so a broken tailnet still reports itself
- a TLS failure no longer withholds the pairing QR (it previously did, via `secureBlocked`); LAN and tunnel endpoints are unaffected by it

**Skipped tests: six, down from seven, and none of them cover TLS.** Because the TLS path no longer touches the picker, all three secure-pairing tests run:

- `turns secure pairing on wherever a tailnet address exists`
- `does not retry enabling secure pairing when it is unavailable`
- `surfaces an error when enabling secure pairing fails`

The six that remain skipped drive the selector itself (`selectConnectionMethod`) and nothing else — dropdown width, per-mode address display, per-mode QR gating. Each carries a note to un-skip alongside the picker. Suite is 102 passed / 6 skipped, typecheck clean.

**What is off-screen while this stands**, stated plainly so it is a decision on the record rather than a surprise:

- Tailscale's "install Tailscale on both machines" setup step
- the address/password rows for the tailnet host — manual entry now only ever shows the LAN address

The QR path is unaffected in every case. Happy to restore the picker instead if you would rather not carry commented-out code on main; it is three linked blocks and the six tests, so it reverts as one unit.

Issue #4120, opened 2026-08-19
bug(chat): TUI→Chat switch permanently fails with 'settled native conversation history … context deadline exceeded' after a failed chat turn


AgentWrapper:
## Bug

Switching a terminal (TUI) session to the Chat UI fails after ~45s with:

> Interface switch failed — switch interface <session>: resume chat: wait for settled native conversation history: chat conversation history is not settled: context deadline exceeded

Once this state is entered, **every** subsequent switch attempt fails the same way — the session is permanently unable to enter Chat mode.

**Source:** live triage of orchestrator session `agent-orchestrator-88` (project agent-orchestrator), reported by the maintainer in chat, 2026-08-19 03:15Z | **Analyzed against:** `4a23d98ca`
**Confidence:** High (root cause reproduced empirically against the live transcript and the real `claude-agent-acp` adapter; regression test reproduces the exact error string and 45s timeout on unfixed code)

## Reproduction (mechanism)

1. Session is in Chat mode; a chat turn **fails** (in the observed incident: `Failed to authenticate: OAuth session expired and could not be refreshed`, recorded as a terminal `failed` turn with the synthetic assistant error message as its last item).
2. User switches Chat → TUI (succeeds), fixes the problem (e.g. `/login`), and continues working in the terminal. Claude Code forks the next prompt from the **pre-failure** transcript entry, so the failed turn's entries (user prompt + synthetic auth-error assistant message) land on a **dead branch** of the transcript's `parentUuid` tree.
3. User switches TUI → Chat. The resume runs `session/load`; `claude-agent-acp`'s replay walks the parent chain from the last entry — the dead branch is **never replayed** (verified: replay of the live 992-entry transcript contained 183-entry active chain only; the auth-error turn was absent).
4. `captureAOHighWater` (`backend/internal/service/chat/controller.go`) anchors the settle checkpoint on the **latest terminal turn — including `failed` ones**. The anchor item can never appear in the replay, so `nativeHistoryCheckpoint.reached()` returns false forever; the controller polls the (immutable, cached) ACP replay every 100ms until the 45s `nativeHistorySettleLimit` expires → `TARGET_HISTORY_UNSETTLED`, rollback to TUI.
5. Terminal-mode turns never write `conversation_turns` rows, so the failed chat turn **remains the latest terminal turn indefinitely** — the switch can never succeed again for that session.

## Root Cause

[...]

PR #4770, opened 2026-09-01
fix(chat): refresh ACP history during TUI-to-Chat transition

### LINKED ISSUE #4681: bug(chat): TUI to Chat switch fails outright for ACP (Claude Code) sessions when native history is unsettled - "snapshot is incomplete and cannot be refreshed"
## Bug

Switching a terminal (TUI) session to the Chat UI fails **immediately** for ACP-backed (Claude Code) sessions with:

> Interface switch failed - switch interface `<session>`: resume chat: native conversation history snapshot is incomplete and cannot be refreshed: chat conversation history is not settled

Unlike #4120 (which waited ~45s and failed with "context deadline exceeded", now fixed), this fails right away with no wait. The API surfaces it as `TARGET_HISTORY_UNSETTLED`.

## Mechanism (from code)

On a TUI to Chat switch, the chat controller rebuilds the structured timeline by reading the provider's native conversation history (`backend/internal/service/chat/controller.go`, `refreshNativeHistory`) and requires the replay to reach the transition checkpoint:

- If `ReadHistory` errors, or (`required && !checkpoint.reached(events)`), the controller treats the history as unsettled (`ErrChatHistoryUnsettled`) - synthesizing it at `controller.go:542` when `ReadHistory` returned nil but the checkpoint was not reached.
- If the reader implements `ChatHistoryRefresher`, the controller polls/refreshes until it settles or times out.
- If the reader does **not** implement `ChatHistoryRefresher`, it fails immediately at `controller.go:557`: `native conversation history snapshot is incomplete and cannot be refreshed: <unsettled>`.

The ACP chat driver (`backend/internal/adapters/chatdriver/acp/history.go`) implements `ChatHistoryReader` (`ReadHistory`) but **not** `ChatHistoryRefresher` - only `codexappserver` implements `RefreshHistory` (`.../codexappserver/history.go:200`). So for ACP (Claude Code) sessions, any unsettled or checkpoint-incomplete native replay at switch time is a hard, immediate failure with no retry path. Codex sessions do not hit this branch because their reader is refreshable and polls instead.

`backend/internal/session_manager/interface_transition.go:461` maps `ErrChatHistoryUnsettled` to `TARGET_HISTORY_UNSETTLED` and fails the transition.

## Impact

An ACP (Claude Code) session whose native history is momentarily unsettled, or whose ACP replay does not yet satisfy the switch checkpoint, cannot enter Chat mode at all, and there is no wait/retry for it to settle (the wait path exists only for refreshable readers). The user is stuck in TUI.

## Related (distinct)

- #4120 (closed, fixed): the refreshable/timeout variant - "...context deadline exceeded" after ~45s. Different code path (poll then timeout) and different error string.
- #4123 (closed): settle-wait polled an immutable cached ACP replay for the full 45s before failing. This report is the non-refreshable fast-fail, which may be a side effect of removing ACP from the polling path.
- #4424 (open): unscoped hook checkpoints permanently block TUI-to-Chat recovery - a different blocking cause.

## Suggested direction

Either make the ACP history reader a `ChatHistoryRefresher` so the controller can wait for the replay to settle like Codex, or make the controller degrade gracefully for non-refreshable readers (proceed with the available replay / defer the checkpoint) instead of hard-failing, so a transient unsettled state does not permanently block the switch.

## Source

- Live AO desktop report, 2026-08-30 (session `ao-6`), screenshot of the "Interface switch failed" banner.
- Analyzed against `origin/main` `c13289c1c32afa4acbc2508a4a5402175623cb3e`.
- Confidence: High on the code mechanism (files and lines cited); the specific live trigger (which unsettled turn or checkpoint gap) was not reproduced end to end.

OllieinCanada:
Fixes #4681

## Problem

ACP-backed sessions could fail immediately when switching from TUI to Chat if the first native-history replay had not yet reached the transition checkpoint. ACP exposed only a history reader, so the controller treated the snapshot as non-refreshable and returned `TARGET_HISTORY_UNSETTLED`.

## Root cause

During TUI-to-Chat transition, the session manager stops the source TUI, commits the target controller epoch, and asks Chat to resume the same provider conversation with native history required. ACP's `Resume` opens the target connection and performs one `session/load`; the adapter captured that replay into an immutable slice. If Claude Code's durable transcript had not yet exposed AO's latest completed-turn checkpoint, `controller.readNativeHistory` correctly rejected the stale prefix. Because ACP did not implement `ChatHistoryRefresher`, the capability-based retry path from #4123 failed immediately instead of making another provider observation.

ACP `session/load` is repeatable for the same session ID: it resolves the existing session and replays that session's durable history again. The live adapter retains the connection and full load request needed to repeat that observation safely. The checkpoint remains proven only by the existing completed-turn/high-water matching in the Chat controller.

This differs from:

- #4120, which corrected which AO turn states may anchor the replay high-water mark;
- #4123, which introduced capability-based refresh and intentionally fast-failed ACP while its snapshot was considered immutable;
- #4424, which concerns unscoped checkpoint provenance/recovery and is unchanged here.

## Fix

Return a small `refreshableConversation` wrapper only for ACP agents that advertise `session/load`. It stores the exact initial load request and serializes repeated loads on the already-resumed ACP connection. `RefreshHistory` issues a new `session/load`, captures the new provider replay, and then exposes that snapshot to the existing controller checkpoint logic.

This is not polling the cached `historyEvents` slice: every refresh crosses the ACP protocol and causes another provider transcript observation. Resume-only ACP agents remain non-refreshable.

## Safety

- preserves history-completeness and checkpoint checks unchanged
- keeps the source TUI session intact through the existing rollback path on failure
- reloads the existing provider session; it does not create a replacement conversation
- regenerates stable provider event IDs, so identical replay prefixes remain idempotent and the existing projector deduplication remains effective
- uses the existing bounded controller retry/timeout and propagates context cancellation
- creates no target Chat controller until native history is accepted
- adds no harness-name conditional to service code and changes no Codex behavior

## Tests

Passed:

- regression-first proof: `go test ./internal/adapters/chatdriver/acp -run TestACPDriverRefreshesHistoryWithAnotherSessionLoad -count=1` failed before implementation because ACP did not advertise refreshable history
- `go test ./internal/adapters/chatdriver/acp/... -count=1`
- `go test ./internal/adapters/chatdriver/acp/... ./internal/service/chat/... ./internal/session_manager/... -count=1` (ACP and Chat packages passed; session-manager Windows baseline failures noted below)
- `go test ./internal/session_manager -run InterfaceTransition -count=1`
- `go test -race ./internal/adapters/chatdriver/acp/... ./internal/service/chat/... -count=1`
- `go test -race ./internal/session_manager -run InterfaceTransition -count=1`
- `go vet ./...`
- pinned golangci-lint against the diff: `... golangci-lint@v2.12.2 run --path-mode=abs --new-from-rev=origin/main --timeout=10m` — 0 issues

The final test-only assertion proving a third identical replay performs another wire `session/load` passed in the normal ACP suite. A post-assertion race rerun could not start after Windows cleaned the temporary checksum-verified Go/MinGW toolchain; the race suite above passed with the same production implementation.

Repository-wide Windows limitations:

- `npm run lint` and its `go test ./...` step encounter existing POSIX-path, shell/tmux, symlink-privilege, `/tmp`, and environment-sensitive test failures on Windows.
- The representative session-manager failures were reproduced in a clean `origin/main` worktree at `9aadfc8ae` with the same results.
- Running pinned golangci-lint globally on Windows reported 1,033 baseline findings (1,022 repository-wide goimports reports plus existing Windows errcheck/gosec findings) and timed out; diff-only lint reports 0 issues.

No live Claude Code/ACP provider validation was performed; coverage is deterministic fake-provider replay plus code-path verification.

## Intentional omissions

- No change to #4424's unscoped-checkpoint behavior.
- No controller timeout, replay-completeness, error-mapping, migration, frontend, or Codex app-server change.
- No retry by rereading an immutable cached snapshot and no arbitrary sleep.

### DIFF
--- a/backend/internal/adapters/chatdriver/acp/driver.go
+++ b/backend/internal/adapters/chatdriver/acp/driver.go
@@ -256,21 +256,20 @@ func (d *Driver) Resume(ctx context.Context, cfg ports.ChatResumeConfig) (ports.
 	}
 	var configOptions []acpsdk.SessionConfigOption
 	var modes *acpsdk.SessionModeState
+	var historyConversation *refreshableConversation
 	if init.AgentCapabilities.LoadSession {
-		conv.beginHistoryReplay(cfg.ProviderConversationID)
-		resp, err := conv.conn.LoadSession(resumeCtx, acpsdk.LoadSessionRequest{
+		historyConversation = newRefreshableConversation(conv, acpsdk.LoadSessionRequest{
 			Meta:                  meta,
 			SessionId:             acpsdk.SessionId(cfg.ProviderConversationID),
 			Cwd:                   cfg.WorkspacePath,
 			AdditionalDirectories: additional,
 			McpServers:            mcpServers,
 		})
+		resp, err := historyConversation.loadHistory(resumeCtx)
 		if err != nil {
-			conv.abortHistoryReplay()
 			_ = conv.Close()
 			return nil, fmt.Errorf("%w: %w", ports.ErrChatResumeFailed, normalizeACPError("ACP session/load", err))
 		}
-		conv.finishHistoryReplay()
 		configOptions = resp.ConfigOptions
 		modes = resp.Modes
 	} else {
@@ -299,6 +298,9 @@ func (d *Driver) Resume(ctx context.Context, cfg ports.ChatResumeConfig) (ports.
 			_ = conv.Close()
 			return nil, fmt.Errorf("%w: configure ACP session: %w", ports.ErrChatResumeFailed, err)
 		}
+	}
+	if historyConversation != nil {
+		return historyConversation, nil
 	}
 	return conv, nil
 }
--- a/backend/internal/adapters/chatdriver/acp/driver_test.go
+++ b/backend/internal/adapters/chatdriver/acp/driver_test.go
@@ -29,6 +29,9 @@ type fakeAgent struct {
 	loadParams          acpsdk.LoadSessionRequest
 	resumeParams        acpsdk.ResumeSessionRequest
 	loadUpdates         []acpsdk.SessionUpdate
+	loadUpdateBatches   [][]acpsdk.SessionUpdate
+	blockLoadCall       int
+	loadStarted         chan struct{}
 	loadCalls           int
 	resumeCalls         int
 	promptParams        acpsdk.PromptRequest
@@ -259,8 +262,21 @@ func (a *fakeAgent) LoadSession(ctx context.Context, params acpsdk.LoadSessionRe
 	a.mu.Lock()
 	a.loadParams = params
 	a.loadCalls++
+	loadCall := a.loadCalls
 	updates := append([]acpsdk.SessionUpdate(nil), a.loadUpdates...)
+	if batch := a.loadCalls - 1; batch < len(a.loadUpdateBatches) {
+		updates = append([]acpsdk.SessionUpdate(nil), a.loadUpdateBatches[batch]...)
+	}
+	block := a.blockLoadCall == loadCall
+	started := a.loadStarted
 	a.mu.Unlock()
+	if block {
+		if started != nil {
+			close(started)
+		}
+		<-ctx.Done()
+		return acpsdk.LoadSessionResponse{}, ctx.Err()
+	}
 	for _, update := range updates {
 		if err := a.conn.SessionUpdate(ctx, acpsdk.SessionNotification{SessionId: params.SessionId, Update: update}); err != nil {
 			return acpsdk.LoadSessionResponse{}, err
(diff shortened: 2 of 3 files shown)


Pulkit7070:
Reviewed end to end. This fixes #4681 with the right approach: ACP now implements `ChatHistoryRefresher` by repeating `session/load`, so the controller can refresh instead of fast-failing with `TARGET_HISTORY_UNSETTLED`.

- The mechanism is sound: `refreshableConversation` captures the load request and re-runs `session/load` under `loadMu`, and `RefreshHistory` reloads then re-reads. `historyEvents` stays consistently guarded by `historyMu` and `ReadHistory` returns a copy under lock, so the repeated replay is race-safe (verified: the acp package passes `-race -count=10`).
- Driver-level coverage is good: refresh triggers a fresh `session/load`, history grows across loads, `ProviderEventID` identities stay stable across replays (so the Chat projector dedupes), a blocked load honors `ctx` cancellation, and a resume-only conversation correctly does not advertise refreshable.

Two notes, neither blocking:
- The end-to-end path that #4681 is actually about (ACP unsettled, controller refreshes, the switch settles instead of returning `TARGET_HISTORY_UNSETTLED`) is not exercised here; only the ACP `RefreshHistory` primitive is. The controller's generic refresh loop is covered separately so it is functionally covered, but a small integration test would lock the exact regression.
- The fix rests on `session/load` being repeatable for the same session id (documented in the description). That is a reasonable provider-contract assumption for Claude Code ACP, but the unit tests cannot prove it against the real agent.

Ran locally: builds, and the acp package passes `-race -count=10`. Good to merge from my side.

Pulkit7070 (approved):
Approving. Fixes #4681 by making ACP a ChatHistoryRefresher via repeatable session/load; mechanism is race-safe (acp package passes -race -count=10 locally) and driver-level coverage is thorough (refresh triggers a fresh load, event identities stay stable, cancellation honored).

PR #4774, opened 2026-09-01
test(codexappserver): tolerate a provider older than the generated protocol

### LINKED ISSUE #4773: [Bug]: codexappserver conformance test fails when the installed Codex is older than the pinned ProviderVersion
### Summary

`go test ./...` fails on a clean checkout when the Codex installed locally is older than the version the checked in protocol file was generated from. Nothing in AO is actually broken. The test cannot tell the difference between "someone added a method that does not exist" and "this developer has an older Codex build".

`TestGeneratedProtocolMatchesTheInstalledProvider` compares the generated method table in `codexproto/protocol.gen.go` against the schema that the `codex` binary on PATH emits. It treats the two directions differently:

| Direction | Handling | Line |
|---|---|---|
| Installed provider has methods the generated file lacks | `t.Logf`, test passes | `conformance_test.go:202` |
| Generated file has methods the installed provider lacks | `t.Errorf`, test fails | `conformance_test.go:209` |

Failing on the second case is correct when the generated file has genuinely drifted, because calling a method the provider does not have breaks at runtime. But the same condition is also true when the generated file is fine and the developer simply has an older Codex, which is not a defect.

The test is already tolerant of an unusable environment. It skips when `codex` is not installed (`conformance_test.go:169-171`) and when the provider refuses to emit a schema (`conformance_test.go:176-177`). Being strict only in the "installed Codex is older" case looks like something that was missed rather than a decision.

`protocol.gen.go` pins `ProviderVersion = "codex-cli 0.146.0"` and has been regenerated once, in [PR #3472](https://github.com/Untrivial-ai/agent-orchestrator/pull/3472). Codex keeps releasing, so the distance between the pinned version and whatever a contributor has installed grows over time.

### Steps to reproduce

1. Install a Codex CLI older than the pinned `ProviderVersion`. I have `codex-cli 0.142.5` and the file pins `codex-cli 0.146.0`.
2. Clone the repo and check out current `main`.
3. From `backend/`, run the documented local gate:
   ```
   go test ./...
   ```
4. Or run just the one test:
   ```
   go test ./internal/adapters/chatdriver/codexappserver/ -run TestGeneratedProtocolMatchesTheInstalledProvider
   ```

### Expected behavior

A contributor whose Codex is simply an older build should not get a red suite. The test should separate the two causes. One way is to compare the installed provider version against `codexproto.ProviderVersion` and skip or log when the installed build is older, the same way the test already skips when Codex is absent, while still failing when the versions line up and the generated file has actually drifted.

### Actual behavior

The test fails at `conformance_test.go:209`:

```
--- FAIL: TestGeneratedProtocolMatchesTheInstalledProvider (0.84s)
    conformance_test.go:209: generated protocol declares 5 method(s) the installed
    provider does not: [externalAgentConfig/import/recordHistory
    thread/environment/connected thread/environment/disconnected
    app/installed app/read]
FAIL
FAIL	github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/codexappserver	0.843s
```

conformance test failing against codex-cli 0.142.5

`go test ./...` is what `docs/development.md` and `docs/STATUS.md` present as the local gate, and `npm run lint` runs `go test ./...` as well. So a new contributor running the documented command for the first time sees a failure that has nothing to do with their changes.

### AO version

`0.10.3`, commit `92990fd3`, current `main` at time of filing.

### Environment

Ubuntu 24.04.4 LTS x86_64, kernel 6.14.0-37, Go 1.26.5, `codex-cli 0.142.5`

### Logs / extra context

```shell
$ codex --version
codex-cli 0.142.5

$ grep "ProviderVersion =" internal/adapters/chatdriver/codexappserver/codexproto/protocol.gen.go
const ProviderVersion = "codex-cli 0.146.0"

$ go test ./internal/adapters/chatdriver/codexappserver/ -run TestGeneratedProtocolMatchesTheInstalledProvider
--- FAIL: TestGeneratedProtocolMatchesTheInstalledProvider (0.84s)
    conformance_test.go:209: generated protocol declares 5 method(s) the installed provider does not: [externalAgentConfig/import/recordHistory thread/environment/connected thread/environment/disconnected app/installed app/read]
FAIL
FAIL	github.com/aoagents/agent-orchestrator/backend/internal/adapters/chatdriver/codexappserver	0.843s
FAIL
```

The method names print in a different order on each run because Go iterates a map. The set itself is stable.

mukulkatewa:
## What

Makes `TestGeneratedProtocolMatchesTheInstalledProvider` version aware. When the installed Codex is older than the build `protocol.gen.go` was generated from, the missing methods are reported with `t.Logf` instead of failing the test. When the installed build is not older, the existing `t.Errorf` is unchanged.

Fixes #4773.

## Why

`go test ./...` fails on a clean checkout for anyone whose Codex predates the pinned `ProviderVersion`, and nothing in the checkout is wrong.

The test compares the generated method table against the schema the installed `codex` emits, then splits the result two ways. A provider that declares methods the generated file lacks is only logged, because AO does not have to use everything. A generated file that declares methods the provider lacks is failed, because calling one of those breaks at runtime.

That second rule is right when the generated file has drifted. The problem is that an older provider produces exactly the same condition without anything having drifted: every method a later Codex added is a method the older build does not declare. The test could not tell the two apart, so it reported a version gap as a defect.

The test is already deliberate about not failing on an environment it cannot use. It skips when `codex` is absent and when the provider refuses to emit a schema. Failing only in the "your Codex is older" case was the gap.

This is not a rare state. `protocol.gen.go` pins `codex-cli 0.146.0` and has been regenerated once, in #3472, while Codex keeps releasing. My machine has `codex-cli 0.142.5` and reports five methods. `go test ./...` is what `docs/development.md` and `docs/STATUS.md` present as the local gate, and `npm run lint` runs it too, so this is the first command a new contributor runs.

## How

Test-only change. No production code is touched.

The comparison already exists in this package, so nothing new was introduced: `parseCodexVersion` and `codexVersion.less` in `driver.go` do the work. The new `providerOlderThanGenerated` helper only joins them and decides the one question the test needs answered.

Unparseable output on either side returns false and keeps the strict comparison. A version that cannot be read is not evidence that the gap is benign, so the safe direction is to keep failing rather than to quietly pass.

The version is read with `exec.Command(bin, "--version")` to match how this file already invokes the provider, rather than `installedCodexVersion`, which carries the driver's process environment that a schema comparison does not need.

The two `len(extra) > 0` arms are a `switch` rather than a nested `if` so the strict path stays visible at the same level as the tolerant one.

## Testing

```
go test ./internal/adapters/chatdriver/codexappserver/ -run TestGeneratedProtocolMatchesTheInstalledProvider -v
```

Before this change, on `codex-cli 0.142.5`:

```
--- FAIL: TestGeneratedProtocolMatchesTheInstalledProvider (0.84s)
    conformance_test.go:209: generated protocol declares 5 method(s) the installed provider does not: [...]
FAIL
```

After:

```
=== RUN   TestGeneratedProtocolMatchesTheInstalledProvider
    conformance_test.go:220: installed provider (codex-cli 0.142.5) predates the generated "codex-cli 0.146.0", so it does not declare 5 method(s) this file has; update codex or regenerate to compare them: [...]
--- PASS: TestGeneratedProtocolMatchesTheInstalledProvider (0.91s)
```

`TestProviderOlderThanGenerated` covers the decision directly: older, equal, newer, and unreadable version output on either side.

I could not verify the strict path against a real Codex at or above `0.146.0`, since the only build available to me is older. The equal and newer cases are covered by the table test, and that path is unchanged from the current behavior.

Also run: `gofmt -l` clean, `go vet` clean, full `go test ./...`.

## Checklist

### DIFF
--- a/backend/internal/adapters/chatdriver/codexappserver/conformance_test.go
+++ b/backend/internal/adapters/chatdriver/codexappserver/conformance_test.go
@@ -177,6 +177,13 @@ func TestGeneratedProtocolMatchesTheInstalledProvider(t *testing.T) {
 		t.Skipf("provider declined to emit its schema (%v): %s", err, out)
 	}
 
+	// An older provider is always missing methods that later builds added, so the
+	// comparison below has to know which case it is looking at before it can call a
+	// gap drift.
+	installedOut, installedErr := exec.Command(bin, "--version").CombinedOutput()
+	olderProvider := installedErr == nil &&
+		providerOlderThanGenerated(string(installedOut), codexproto.ProviderVersion)
+
 	live := methodsFromSchema(t, dir)
 	generated := map[string]bool{}
 	for _, m := range codexproto.Methods {
@@ -204,11 +211,36 @@ func TestGeneratedProtocolMatchesTheInstalledProvider(t *testing.T) {
 			len(missing), truncate(missing))
 	}
 	// A method AO's generated file has and the provider does not is the dangerous
-	// direction: any call to it fails at runtime.
-	if len(extra) > 0 {
+	// direction: any call to it fails at runtime. That only means drift when the
+	// provider is at least the build this file was generated from. On an older
+	// provider the same gap just means the developer has not updated codex, which is
+	// not a defect in the checkout and must not fail the suite.
+	switch {
+	case len(extra) > 0 && olderProvider:
+		t.Logf("installed provider (%s) predates the generated %q, so it does not declare "+
+			"%d method(s) this file has; update codex or regenerate to compare them: %v",
+			strings.TrimSpace(string(installedOut)), codexproto.ProviderVersion,
+			len(extra), truncate(extra))
+	case len(extra) > 0:
 		t.Errorf("generated protocol declares %d method(s) the installed provider does not: %v",
 			len(extra), truncate(extra))
 	}
+}
+
+// providerOlderThanGenerated reports whether the installed provider predates the
+// build the generated file describes. Unparseable output on either side reports
+// false, which keeps the strict comparison: a version that cannot be read is not
+// evidence that the gap is benign.
+func providerOlderThanGenerated(installedVersionOutput, generatedProviderVersion string) bool {
+	installed, ok := parseCodexVersion(installedVersionOutput)
+	if !ok {
+		return false
+	}
+	generated, ok := parseCodexVersion(generatedProviderVersion)
+	if !ok {
+		return false
+	}
+	return installed.less(generated)
 }
 
 func methodsFromSchema(t *testing.T, dir string) map[string]bool {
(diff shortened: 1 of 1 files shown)


mukulkatewa:
Closing this. Agreed after discussion that updating codex is the right resolution and this does not need handling in the test.

PR #4779, opened 2026-09-01
fix: stop ⌘T on project board from opening /terminals

### LINKED ISSUE #4772: Fix CMD + T behavior on project board
Image

Annieeeee11:
## Summary
- Fixes [#4772](https://github.com/Untrivial-ai/agent-orchestrator/issues/4772): ⌘T/Ctrl+T on the project board no longer navigates to the standalone `/terminals` view.
- The shortcut still works in sessions (new tab in the session terminal strip) and on `/terminals` (new standalone shell).
- Opening a shell from the keyboard shortcut now uses the same optimistic `open()` path as the session **New terminal** button, so tab selection and navigation happen immediately instead of after the daemon API returns.

## Test plan
- [ ] On a project board, press ⌘T / Ctrl+T — nothing should happen (no navigation to `/terminals`).
- [ ] On a project board, click **New terminal** (if exposed) — should still open `/terminals` as before.
- [ ] In a session, press ⌘T / Ctrl+T — new terminal tab appears immediately in the session strip.
- [ ] On `/terminals`, press ⌘T / Ctrl+T — new tab appears immediately without a ~2s wait.
- [ ] `npx vitest run src/renderer/test/shell-new-session-shortcut.test.tsx`

### DIFF
--- a/frontend/src/renderer/routes/_shell.tsx
+++ b/frontend/src/renderer/routes/_shell.tsx
@@ -332,6 +332,7 @@ function ShellLayout() {
 	// The root route is the intentionally minimal home surface, regardless of
 	// whether projects have already been registered.
 	const isHomeRoute = Boolean(matchRoute({ to: "/" }));
+	const isTerminalsRoute = Boolean(matchRoute({ to: "/terminals" }));
 	const isSettingsRoute =
 		Boolean(matchRoute({ to: "/settings", fuzzy: true })) ||
 		Boolean(matchRoute({ to: "/projects/$projectId/settings", fuzzy: true }));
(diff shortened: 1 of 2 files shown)


OllieinCanada:
I prepared a stacked hardening PR against this branch: #4789.

It keeps #4779's route fix and existing backend terminal lifecycle, while adding a stable capability-based shortcut subscription, orchestrator-session gating, same-batch repeat handling, per-mutation optimistic reconciliation, conditional selection rollback, contextual shortcut help, and deterministic regression coverage. Because it targets `fix/cmd-t-project-board`, it can be merged into this PR without creating a competing issue PR against `main`.

ronishrohan (approved):
Reviewed the shortcut gating and regression coverage. The project-board shortcut no longer opens /terminals; the standalone terminal UI remains intentionally reachable.

Issue #4771, opened 2026-09-01
fix(landing): restore section anchors, repeatable Demo navigation, and scrollbar


Rishet11:
# Follow-up from #3269

The landing page still has three valid items in current main:

- Footer Features and Agents links point to missing section targets.
- Demo navigation can leave the `#see-it` hash in the URL, which may prevent a repeat click from scrolling again.
- The main page hides the scrollbar, removing the scroll indicator.

The placeholder-testimonials item from #3269 is no longer applicable.

OllieinCanada:
I’d like to work on this. I’ll verify each current landing-page target and scroll behavior, add focused regression coverage where the existing frontend test setup supports it, and keep the patch limited to anchors, repeatable Demo navigation, and the scrollbar.

PR #4782, opened 2026-09-02
fix(landing): restore themed scroll indicators

OllieinCanada:
Follow-up to #4771 and #4777

## Problem

#4777 restored section anchors and repeatable hash navigation, but explicitly left the themed scrollbar and shared header-height work out. The global site theme still hid the page scrollbar, so the landing page offered no scroll-position indicator.

## Fix

- replace global hidden-scrollbar rules with themed cross-browser scrollbar styling
- add a shared `--header-height` token used by header/mobile geometry and anchor scroll padding
- honor reduced-motion preferences for smooth anchor scrolling
- keep `.scrollbar-hide` available for intentionally hidden inner demo scrollbars

## Tests

- `cd frontend/src/landing && npm run build`
- `git diff --check`

### DIFF
--- a/frontend/src/landing/src/app/components/Header/Header.tsx
+++ b/frontend/src/landing/src/app/components/Header/Header.tsx
@@ -50,7 +50,7 @@ export function Header({ ctaButtons }: HeaderProps) {
     >
       <div className="px-4 sm:px-8 lg:px-[30px]">
         <div className="max-w-7xl mx-auto">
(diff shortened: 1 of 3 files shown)


i-trytoohard:
@OllieinCanada Marking as stale: no activity for 22 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4783, opened 2026-09-02
feat(browser): support action postconditions

### LINKED ISSUE #4708: feat(browser): support action postconditions and report cancelled navigation
## Problem

Browser actions can report that an element was matched and a low-level click/keypress was dispatched while the application produces no observable result. On guarded or authenticated sites, an agent can misread that as the intended action succeeding and may retry a consequential control.

**Reported by:** external AO user; report relayed by the project operator (public handle/source URL not supplied)  
**Observed:** 2026-08-31 on an authenticated Angular application  
**Analyzed against:** `main` at `ab17197b0d1bdc8d4d9849ef46675e067fc77119`  
**Classification:** Product/observability enhancement, not yet a confirmed AO interaction bug.

## Evidence

A visible `Start New Filing` button was present in the snapshot. `act`, `click`, `dblclick`, focus + Enter, and focus + Space all reported a matched/dispatched interaction. The control visibly received focus, but the URL and subsequent snapshots did not change. Chromium also logged a blocked `beforeunload` confirmation.

A physical-click control result was not available, so the report does **not** establish whether AO synthetic input, the site's route guard, or site JavaScript prevented navigation. Current `act.outcome: "matched"` describes deterministic element matching, not application-level success, but that distinction is easy to miss.

## Desired behavior

Allow callers to state an expected postcondition and distinguish:

- element matched;
- input dispatched;
- navigation/dialog/DOM effect observed;
- navigation cancelled, blocked, or timed out;
- no observable effect.

## Acceptance criteria

- Add optional postconditions such as expected URL, text, dialog, navigation, or DOM change.
- Include before/after URL and document generation/navigation facts in action results.
- Surface blocked/cancelled navigation and dialog signals in the correlated action response when available.
- Clearly document that `matched` is element-resolution status, not proof of application success.
- Avoid automatic retries of non-idempotent actions after an unknown outcome.
- Add a local integration fixture with `beforeunload`/route-guard behavior and tests for satisfied, unmet, and cancelled postconditions.

OllieinCanada:
Fixes #4708

## Problem

`ao browser act` previously proved only that AO matched an element and dispatched provider input. It could not tell callers whether the application produced an observable effect, and retrying after dispatch could duplicate a consequential action.

The first implementation added postconditions, but three correctness gaps remained:

- URL/text waits could report `satisfied` when the expected value was already present before dispatch.
- DOM/dialog observations polled a process-backed runtime every 25 ms and did not bound each native call to the remaining deadline.
- guarded navigation was covered only by invoking a mocked Electron listener, not by a real page executing `beforeunload`.

## Runtime trace and compatibility

The existing path remains intact: CLI flags -> browser HTTP command -> broker -> Electron browser host -> agent-browser runtime. `act` still snapshots and matches through the existing provider adapter, performs the existing stale-reference retry only before dispatch, dispatches input once, and then observes a postcondition. No daemon logic moved into the renderer and no provider/harness special case was added.

The merge with current `main` preserves the new browser-history store: main-frame navigations update both action-correlation generations and top-level history, while iframe history changes update neither.

## Fix

- capture URL/text satisfaction before dispatch and return `already-satisfied` when no correlated page effect makes the condition attributable to the action
- preserve separate `matched`, `inputDispatched`, `navigation`, and `postcondition` facts
- poll unavoidable DOM/dialog observations at 200 ms and give every process-backed observation only the remaining deadline
- propagate genuine provider failures; map only the pinned native wait-timeout message to an unmet timeout
- use a dialog result returned by the action as immediate evidence instead of polling
- prefer a committed navigation over an earlier blocked attempt, recognize an activated popup as navigation, and ignore iframe-only history changes
- reject invalid direct API timeouts instead of silently clamping them
- add a local HTTP fixture with a real `beforeunload` handler and dispatch real Chromium mouse input through the DevTools input domain
- run that fixture in frontend CI under Electron/Xvfb with bounded process-tree cleanup

## Safety

- no action is redispatched because a postcondition is unmet, cancelled, or unknown
- pre-existing URL/text state is never credited to the action
- each observation is bounded and cancellation-aware; a provider that ignores cancellation cannot extend the API deadline or create an unhandled rejection
- no application state or history check is weakened
- no new provider conversation/session is created
- existing behavior is unchanged when no postcondition is requested
- Chromium's native `beforeunload` behavior is observed, not overridden

This follows the same broad synchronization principle used by Playwright/Puppeteer: arm observation around the action and require browser evidence, rather than treating dispatch completion as application success. Electron's documented `will-prevent-unload` event remains the source of guarded-navigation cancellation evidence.

## Tests

Passed on the final merged head unless noted:

- `cd frontend && node node_modules/vitest/vitest.mjs run src/main/browser-view-host.test.ts --pool=forks --maxWorkers=1` — 141 passed
- `cd frontend && node node_modules/vitest/vitest.mjs run src/main/agent-browser-runtime.test.ts --pool=forks --maxWorkers=1` — 22 passed, 2 existing skips
- `cd frontend && npm run test:browser-postcondition-integration` — passed repeatedly on Windows, including after merging current `main`
- `cd backend && go test ./internal/cli -run TestBrowser` — passed
- `git diff --check origin/main...HEAD` — passed
- full TypeScript `tsc --noEmit` passed after the production changes; the final integration-only edits also bundled successfully with Vite and executed in Electron

Broader local commands were attempted. `go test ./...`, `go vet ./...`, and therefore root `npm run lint` hit existing Windows/Unix portability failures; representative failures were reproduced in a clean `origin/main` worktree (for example Unix fake-shell/auth fixtures and `persistenthost/host_race_test.go` using `syscall.Kill`). A separate golangci bootstrap attempt could not begin linting because the host temp volume ran out of space. Linux CI is the authoritative broader check.

## Intentional scope boundary

The broker currently cannot return a post-dispatch acknowledgement after the caller itself cancels the in-flight request; preserving that fact across cancellation would require a two-phase protocol change. This PR keeps cancellation prompt and prevents redispatch within the live request, but deliberately does not expand #4708 into that cross-layer protocol redesign.

No live authenticated-site validation was performed. Coverage consists of deterministic host/runtime tests and a real local Chromium guarded-navigation fixture.


i-trytoohard:
@OllieinCanada Marking as stale: no activity for 21 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

PR #4784, opened 2026-09-02
fix(project): allocate unique derived project IDs

### LINKED ISSUE #4765: bug(projects): same-basename folders at different paths collide on derived project id — desktop creation dead-ends on ID_ALREADY_REGISTERED
## Bug
Registering two projects whose folders share the same basename (e.g. `~/work/api` and `~/clients/api`) fails on the second registration with an **ID conflict** — the desktop UI shows a generic "create failed" error with no path forward.

**Source:** Discord #bug-triage (Vansh), 2026-09-01 | **Reported by:** Vansh
**Analyzed against:** `Untrivial-ai/agent-orchestrator@9997248c4` (origin/main tip)
**Confidence:** High — exact code path traced end-to-end
**Priority:** high — core flow (project creation) dead-ends in the UI with no in-product workaround

## Reproduction
1. Register a project for a git repo at `~/work/api` (desktop: Create Project flow, or `ao project add --path ~/work/api`). Succeeds; project id = `api`.
2. Register a second project at `~/clients/api` (same folder basename, different parent).
3. Desktop: creation fails with a generic error — no explanation, no suggested id, no retry path. Raw daemon response is `409 ID_ALREADY_REGISTERED`.

## Root Cause

Two layers combine to produce the dead-end:

**1. Daemon — project ids are derived from the folder *basename*, not the full path.**

`defaultProjectID` (`backend/internal/service/project/service.go:867` at `9997248c4`; was `:835` at `ed52364e3`):

```go
func defaultProjectID(path string) domain.ProjectID {
	id := strings.ToLower(filepath.Base(path))
	id = strings.TrimSpace(id)
	id = strings.ReplaceAll(id, " ", "-")
	return domain.ProjectID(id)
}
```

Two distinct paths → identical derived id. `Add()` then rejects the second registration at `service.go:204`:

```go
} else if ok && existing.ArchivedAt.IsZero() && existing.Path != path {
	return Project{}, apierr.Conflict("ID_ALREADY_REGISTERED", "A project with this id is already registered for a different path", map[string]any{
		"existingProjectId":  existing.ID,
		"suggestedProjectId": string(m.suggestID(ctx, id)),
	})
}
```

The daemon already computes an escape hatch — `suggestID` (`service.go:734`) returns `api1`-style suffixed candidates and returns it in `suggestedProjectId` — but it is only embedded in the error details.

**2. Desktop UI — the suggestion is dropped and no custom-id input exists.**

- `createProject` in `frontend/src/renderer/lib/shell-context.ts:11` accepts only `path`, agents, tracker intake, `asWorkspace` — **there is no field to pass a `projectId`**, so the UI can never avoid or resolve the collision.
- `projectSheetError` (`frontend/src/renderer/components/CreateProjectAgentSheet.tsx:65`) has tailored cases for `PROJECT_PATH_NOT_REPO_ROOT`, `PROJECT_BARE_REPOSITORY`, `UNSUPPORTED_GIT_REPO` — `ID_ALREADY_REGISTERED` (and `PATH_ALREADY_REGISTERED`) fall through to the generic "createFailedTitle / tryAgain" default. Grep of `frontend/src` for `ID_ALREADY_REGISTERED` / `suggestedProjectId`: **zero matches**.

So the user hits a hard dead-end: the daemon knows the fix (`api1`), the UI never shows it, and the UI has no id field to apply it manually.

## Workaround
CLI supports explicit ids (desktop does not):

```bash
ao project add --path ~/clients/api --id api2
```

(`--id` flag: `backend/internal/cli/project.go:268`.)

## Fix Direction
Two candidate approaches (can combine):

1. **UI (minimum viable):** add an `ID_ALREADY_REGISTERED` case to `projectSheetError` that surfaces "a project named X already exists" + the `suggestedProjectId` from the error details, with a one-tap "use `api1`" retry. Mirror for `PATH_ALREADY_REGISTERED` (see #4403).
2. **Daemon (structural):** make collision resolution automatic — on `ID_ALREADY_REGISTERED` for a derived (not user-supplied) id, transparently register under `suggestID()`'s candidate instead of erroring (display name still defaults to the basename, so both projects *display* as "api" but have distinct ids). User-supplied ids should keep the explicit conflict error.

## Impact
- Any user with same-named folders at different paths — common with `api`, `frontend`, `backend`, client/project dirs.
- Desktop flow is a dead-end; only CLI users who know about `--id` can escape.
- Same-basename dirs also collide with display assumptions elsewhere (session prefixes derive from id — `sessionPrefix`, `service.go`).

## Related
- #4403 — the mirror conflict: same *path* re-import dead-ends on `PATH_ALREADY_REGISTERED` with no "go to that project" option. Same UX gap class, different trigger.
- #4479 — dotted-basename project ids (`memron.ai`) rejected by ConPTY on Windows; same `defaultProjectID` derivation, different failure mode.

OllieinCanada:
Fixes #4765

## Problem

When callers omitted `project_id`, project registration derived the ID from the repository basename. Two repositories with the same basename therefore collided even though ID allocation already had a deterministic suffixing helper.

## Fix

- allocate a unique suffixed ID only when an automatically derived ID collides
- preserve the existing `ID_ALREADY_REGISTERED` behavior for explicitly supplied IDs
- preserve the repository basename as the display name rather than exposing the suffixed storage ID

## Tests

- `cd backend && go test ./internal/service/project -run '^TestManager_Add(AllocatesUniqueIDForCollidingDerivedIDs|ValidationAndConflicts)$' -count=1`
- `git diff --check`

The full package test was also attempted on Windows but did not complete; the focused regression and conflict coverage passed deterministically.

### DIFF
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -183,6 +183,8 @@ func (m *Service) Add(ctx context.Context, in AddInput) (Project, error) {
 		return Project{}, apierr.Internal("PROJECT_LOAD_FAILED", "Failed to load project")
 	}
 
+	// Display names retain the caller's requested/derived base even when an
+	// omitted project ID needs a storage-only suffix below.
 	name := string(id)
 	if in.Name != nil {
 		name = strings.TrimSpace(*in.Name)
@@ -202,10 +204,16 @@ func (m *Service) Add(ctx context.Context, in AddInput) (Project, error) {
 	if existing, ok, err := m.store.GetProject(ctx, string(id)); err != nil {
 		return Project{}, apierr.Internal("PROJECT_LOAD_FAILED", "Failed to load project")
 	} else if ok && existing.ArchivedAt.IsZero() && existing.Path != path {
-		return Project{}, apierr.Conflict("ID_ALREADY_REGISTERED", "A project with this id is already registered for a different path", map[string]any{
-			"existingProjectId":  existing.ID,
-			"suggestedProjectId": string(m.suggestID(ctx, id)),
-		})
+		if in.ProjectID != nil {
+			return Project{}, apierr.Conflict("ID_ALREADY_REGISTERED", "A project with this id is already registered for a different path", map[string]any{
+				"existingProjectId":  existing.ID,
+				"suggestedProjectId": string(m.suggestID(ctx, id)),
+			})
+		}
+		// A caller that omitted projectId asked the service to derive one from
+		// the basename. Resolve that derived collision here; explicit IDs retain
+		// their conflict semantics so user intent is never silently rewritten.
+		id = m.suggestID(ctx, id)
 	}
 
 	var projectConfig domain.ProjectConfig
--- a/backend/internal/service/project/service_test.go
+++ b/backend/internal/service/project/service_test.go
@@ -1054,6 +1054,43 @@ func TestManager_AddValidationAndConflicts(t *testing.T) {
 	wantCode(t, err, "ID_ALREADY_REGISTERED")
 }
 
+func TestManager_AddAllocatesUniqueIDForCollidingDerivedIDs(t *testing.T) {
+	ctx := context.Background()
+	m := newManager(t)
+	configureCommitter(t)
+
+	root := t.TempDir()
+	repoA := filepath.Join(root, "one", "app")
+	repoB := filepath.Join(root, "two", "app")
+	for _, repo := range []string{repoA, repoB} {
+		if err := os.MkdirAll(repo, 0o755); err != nil {
+			t.Fatal(err)
+		}
+		if out, err := exec.Command("git", "init", "-b", "main", repo).CombinedOutput(); err != nil {
+			t.Fatalf("git init: %v (%s)", err, out)
+		}
+		commitEmpty(t, repo)
+	}
+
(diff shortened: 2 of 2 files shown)


Pulkit7070:
Reviewed end to end.

- The fix is correct: on a derived-ID collision at a different path, an omitted `projectId` now gets a suffixed storage ID (`app` -> `app1`), while an explicitly supplied ID still returns `ID_ALREADY_REGISTERED`, so user intent is preserved. The display name is captured before the suffix, so it stays the basename.
- `suggestID` returns the first fully-unused ID (it skips both active and archived rows), so the suffixed ID cannot collide with an archived row on insert.
- Ran locally: the full `internal/service/project` package passes under `-race`, and the new collision test passes `-race -count=15`. I was able to run the full package here, so the Windows gap noted in the description is covered.

One optional follow-up, not blocking: the regression test covers a single collision (`app` -> `app1`). The case where the first suffix is also taken (`app` -> `app2`) and the archived-ID-reuse path are not asserted, though the `suggestID` loop handles both. Worth a line if you want that boundary locked in.

Good to merge from my side.

codebanditssss:
Thanks for the contribution. Closing this in favor of merged #4801, which changes the same project-ID allocation code and implements the same explicit-ID preservation, deterministic suffixing, and display-name behavior for same-basename repositories.

PR #4786, opened 2026-09-02
fix: tighten Help report form submit button layout

Annieeeee11:
## Summary
- Move the Help report copy/submit button onto the same row as the destination picker
- Keep success/error feedback directly below that row

## Test plan
- [x] `GlobalSettingsForm.test.tsx`
- [ ] Open Settings → Help and confirm the submit button sits beside the destination tabs
- [ ] Submit a report via GitHub, Discord, and Email — behavior unchanged

## image
image
image


ronishrohan (changes requested):
image

use same border radius as this input box on the browser page in settings modal for the inputs/textareas

and hide the resize handle from the textarea (just let it be a scroll-view)

and instead of the selector for the selector for github - email - discord, show them all the contacts together, maybe as icon buttons,

ronishrohan (approved):
lgtm

PR #4789, opened 2026-09-02
fix: harden contextual terminal shortcut handling

OllieinCanada:
Improves #4779
Related to #4772

## Why this is stacked

The canonical issue fix is already open as #4779 on the organization-owned branch `fix/cmd-t-project-board`. This PR targets that branch so the hardening can be merged into #4779 without creating a competing fix against `main`.

## Additional risks covered

- route changes recreated the IPC subscription and left shortcut intent to be consumed later through a global nonce
- two shortcut events delivered in one React batch could collapse into one open
- orchestrator sessions hid the New terminal action but still accepted the keyboard shortcut
- concurrent optimistic opens used per-call mutation callbacks, whose observer can be replaced by a later mutate call
- completion or failure could steal or clear a terminal selection the user had changed while the request was pending
- shortcut help presented New terminal as globally available

## Changes

- resolve an explicit worker-session or standalone terminal target at keypress time
- keep one bridge subscription across board/session/terminals route transitions
- reject the shortcut on boards, settings, and orchestrator sessions
- preserve the existing explicit board/UI signal that intentionally opens `/terminals`
- carry optimistic callbacks with each mutation input so out-of-order opens reconcile independently
- update the active handle only while that mutation's optimistic tab remains selected
- document shortcut scope in all supported locale catalogs

## Legacy safety

- no backend, PTY, persistence, or session-gate changes
- shell creation still uses the existing `useOpenShellTerminal` → `/api/v1/shell-terminals` path
- explicit UI requests retain their existing standalone-terminal behavior
- each failed open removes only its own optimistic cache entry
- rapid presses create distinct optimistic handles and daemon requests

## Validation

Passed locally with Node 24:

- focused Vitest suite: 50 tests passed
- final shell shortcut section: new capability/route/optimistic cases passed; a separate pre-existing workspace-startup test timed out under the Windows/OneDrive runner
- frontend TypeScript typecheck
- E2E TypeScript typecheck
- `git diff --check`

The complete Vitest suite was also attempted locally but existing Windows-only updater, shell-script, packaging, socket, and long-running UI tests timed out. The upstream Linux Frontend workflow remains the authoritative broad gate.

### DIFF
--- a/frontend/src/renderer/components/KeyboardShortcutsDialog.test.tsx
+++ b/frontend/src/renderer/components/KeyboardShortcutsDialog.test.tsx
@@ -27,6 +27,13 @@ describe("KeyboardShortcutsDialog", () => {
 		expect(screen.getByLabelText("Ctrl+PageDown")).toBeInTheDocument();
 	});
 
+	it("explains that the new-terminal shortcut is contextual", () => {
+		render(<KeyboardShortcutsDialog isMac={false} onOpenChange={vi.fn()} open />);
+
+		expect(screen.getByText("New terminal")).toBeInTheDocument();
+		expect(screen.getByText("Available in worker sessions and the Terminals view")).toBeInTheDocument();
+	});
+
 	it("uses macOS key labels when requested", () => {
 		render(<KeyboardShortcutsDialog open onOpenChange={vi.fn()} isMac />);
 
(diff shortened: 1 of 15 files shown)


i-trytoohard:
@OllieinCanada Marking as stale: no activity for 22 days. Comment or push and this comes off the close list — activity refreshes. At 30 days without activity it closes. Branch stays safe to reopen.

